PR #4640, opened 2021-04-13
Refactor several distributions

ricardoV94:
Couple more of distributions refactored to `v4`. This should cover all the ones that are to be implemented on Aesara, unless I missed something.

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -66,6 +66,7 @@ from pymc3.distributions.dist_math import (
 from pymc3.distributions.distribution import Continuous
 from pymc3.distributions.special import log_i0
 from pymc3.math import invlogit, log1mexp, log1pexp, logdiffexp, logit
+from pymc3.util import UNSET
 
 __all__ = [
     "Uniform",
@@ -111,10 +112,6 @@ class UnitContinuous(Continuous):
     """Base class for continuous distributions on [0,1]"""
 
 
-class BoundedContinuous(Continuous):
-    """Base class for bounded continuous distributions"""
-
-
 class CircularContinuous(Continuous):
     """Base class for circular continuous distributions"""
 
@@ -129,31 +126,36 @@ def unit_cont_transform(op):
     return transforms.logodds
 
 
-@logp_transform.register(BoundedContinuous)
-def bounded_cont_transform(op):
-    def transform_params(rv_var):
-        _, _, _, *args = rv_var.owner.inputs
-        if hasattr(op, "bound_args"):
-            # Use explicit argument position for lower and upper bound
-            # TODO: Enforce this for all bounded rvs?
-            lower = args[op.bound_args[0]]
-            upper = args[op.bound_args[1]]
-        else:
-            # Assume first argument is lower bound and second is upper bound
-            # Will fail if number of arguments is different than 2
-            lower, upper = args
-
-        print(lower.eval(), upper.eval())
-        lower = at.as_tensor_variable(lower) if lower is not None else None
-        upper = at.as_tensor_variable(upper) if upper is not None else None
-        return lower, upper
-
-    return transforms.interval(transform_params)
-
-
 @logp_transform.register(CircularContinuous)
 def circ_cont_transform(op):
     return transforms.circular
+
+
+class BoundedContinuous(Continuous):
+    """Base class for bounded continuous distributions"""
+
+    transform_args = None
+
+    def __new__(cls, *args, **kwargs):
+        transform = kwargs.get("transform", UNSET)
+        if transform is UNSET:
+            kwargs["transform"] = cls.default_transform()
+        return super().__new__(cls, *args, **kwargs)
+
+    @classmethod
+    def default_transform(cls):
+        if cls.transform_args is None:
+            raise ValueError(f"Must specify transform args for {cls.__name__} bounded distribution")
+
+        def transform_params(rv_var):
+            _, _, _, *args = rv_var.owner.inputs
+            lower = args[cls.transform_args[0]]
+            upper = args[cls.transform_args[1]]
+            lower = at.as_tensor_variable(lower) if lower is not None else None
+            upper = at.as_tensor_variable(upper) if upper is not None else None
+            return lower, upper
+
+        return transforms.interval(transform_params)
 
 
 def assert_negative_support(var, label, distname, value=-1e-6):
@@ -242,6 +244,7 @@ class Uniform(BoundedContinuous):
         Upper limit.
     """
     rv_op = uniform
+    transform_args = [0, 1]  # Lower, Upper
 
     @classmethod
     def dist(cls, lower=0, upper=1, **kwargs):
@@ -3335,7 +3338,7 @@ class Triangular(BoundedContinuous):
     """
 
     rv_op = triangular
-    rv_op.bound_args = [0, 2]  # TODO: Find less hacking solution?
+    transform_args = [0, 2]  # lower, upper
 
     @classmethod
     def dist(cls, lower=0, upper=1, c=0.5, *args, **kwargs):
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -26,6 +26,7 @@ import numpy as np
 from aesara.assert_op import Assert
 from aesara.tensor.random.basic import (
     BetaRV,
+    WeibullRV,
     cauchy,
     exponential,
     gamma,
@@ -33,9 +34,13 @@ from aesara.tensor.random.basic import (
     halfcauchy,
     halfnormal,
     invgamma,
+    logistic,
+    lognormal,
     normal,
     pareto,
+    triangular,
     uniform,
+    vonmises,
 )
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.var import TensorVariable
@@ -110,6 +115,10 @@ class BoundedContinuous(Continuous):
     """Base class for bounded continuous distributions"""
 
 
+class CircularContinuous(Continuous):
+    """Base class for circular continuous distributions"""
+
+
 @logp_transform.register(PositiveContinuous)
 def pos_cont_transform(op):
     return transforms.log
@@ -123,12 +132,28 @@ def unit_cont_transform(op):
 @logp_transform.register(BoundedContinuous)
 def bounded_cont_transform(op):
     def transform_params(rv_var):
-        _, _, _, lower, upper = rv_var.owner.inputs
+        _, _, _, *args = rv_var.owner.inputs
+        if hasattr(op, "bound_args"):
+            # Use explicit argument position for lower and upper bound
+            # TODO: Enforce this for all bounded rvs?
+            lower = args[op.bound_args[0]]
+            upper = args[op.bound_args[1]]
+        else:
+            # Assume first argument is lower bound and second is upper bound
+            # Will fail if number of arguments is different than 2
+            lower, upper = args
+
+        print(lower.eval(), upper.eval())
         lower = at.as_tensor_variable(lower) if lower is not None else None
         upper = at.as_tensor_variable(upper) if upper is not None else None
         return lower, upper
 
     return transforms.interval(transform_params)
+
+
+@logp_transform.register(CircularContinuous)
+def circ_cont_transform(op):
+    return transforms.circular
 
 
 def assert_negative_support(var, label, distname, value=-1e-6):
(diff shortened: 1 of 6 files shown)


pymc3/distributions/continuous.py
```diff
@@ -3393,45 +3339,18 @@ class Triangular(BoundedContinuous):
...
     """
 
-    def __init__(self, lower=0, upper=1, c=0.5, *args, **kwargs):
-        self.median = self.mean = self.c = c = at.as_tensor_variable(floatX(c))
-        self.lower = lower = at.as_tensor_variable(floatX(lower))
-        self.upper = upper = at.as_tensor_variable(floatX(upper))
-
-        super().__init__(lower=lower, upper=upper, *args, **kwargs)
+    rv_op = triangular
+    rv_op.bound_args = [0, 2]  # TODO: Find less hacking solution?
```
ricardoV94: @brandonwillard Is there a more elegant solution? The problem was that `BoundedContinuous` is looking for the `lower/upper` inputs based on order, and in this case it's the first and last inputs that it should use.

pymc3/distributions/continuous.py
```diff
@@ -123,14 +132,30 @@ def unit_cont_transform(op):
...
+            # Use explicit argument position for lower and upper bound
+            # TODO: Enforce this for all bounded rvs?
+            lower = args[op.bound_args[0]]
+            upper = args[op.bound_args[1]]
+        else:
+            # Assume first argument is lower bound and second is upper bound
+            # Will fail if number of arguments is different than 2
+            lower, upper = args
+
+        print(lower.eval(), upper.eval())
```
brandonwillard:
```suggestion
```

brandonwillard (changes requested):

pymc3/distributions/continuous.py
```diff
@@ -2651,6 +2643,17 @@ def __init__(self, nu, *args, **kwargs):
         super().__init__(alpha=nu / 2.0, beta=0.5, *args, **kwargs)
 
 
+class WeibullBetaRV(WeibullRV):
+    ndims_params = [0, 0]
+
+    @classmethod
+    def rng_fn(cls, rng, alpha, beta, size):
+        return beta * rng.weibull(alpha, size=size)
```
brandonwillard, line 2653: Instead of creating a class that's just a constant times an existing `RandomVariable`, let's just create the product in `Distribution.dist`.

pymc3/distributions/continuous.py
brandonwillard: Yeah, we don't want to add properties to objects like this.  As a matter of fact, `Op`s are likely to use `__slots__` in the near future, making this an error.

ricardoV94:
@brandonwillard What do you think of my last commit as a solution for the BoundContinuous issue? 

I moved the `transform_logp` inside `pm.Distribution.__new__()` instead of being inside `model.create_value_var()`. This allows to pass class attributes as optional arguments to `transform_logp` without cluttering the rv_op. 

BTW I see this was before in the metaclass: https://github.com/pymc-devs/pymc3/blob/1e4163339f963d9728d0160e74d594f685d76b92/pymc3/distributions/distribution.py#L112-L120

Should we move it there or remove those commented lines?

pymc3/distributions/continuous.py
ricardoV94: How will we pass the beta argument to the logp/logcdf?

brandonwillard: Since this is just a scaled random variable, we would need to write the `logpt` dispatch that formulates the log-likelihood for a scale random variable.

ricardoV94: That goes towards the idea of having that automatic logp inference for continuous transformations.

It's a great functionality to add but I think it's deserves it's own PR. Some work has to be done to figure out how it would fit with the automatic variable registration and automatic transformations inside the model object.

I am happy to either remove this distribution from the PR or add a clear TODO comment above it for now.

brandonwillard: I'm guessing it can be done in a single commit, so it might as well go here&mdash;or any other PR that needs it.

pymc3/distributions/continuous.py
```diff
@@ -110,27 +115,41 @@ class BoundedContinuous(Continuous):
...
-def bounded_cont_transform(op):
+def bounded_cont_transform(op, transform_args):
+    if transform_args is None:
+        raise ValueError(f"Must specify transform args for distribution of {op}")
+
     def transform_params(rv_var):
-        _, _, _, lower, upper = rv_var.owner.inputs
+        _, _, _, *args = rv_var.owner.inputs
+        lower = args[transform_args[0]]
+        upper = args[transform_args[1]]
```
brandonwillard: You could just override `Distribution.__new__` and create the transform (see [`Dirchlet.__new__`](https://github.com/pymc-devs/pymc3/blob/v4/pymc3/distributions/multivariate.py#L390)).

brandonwillard (changes requested):
We don't want to start adding variables at the class level unnecessarily; instead, each class that wants to use an `interval` transform should just create its own in `Distribution.__new__`.  That approach is class-level enough to accomplish the same thing, and it makes the transform-setup logic a little less spread out.

pymc3/distributions/continuous.py
ricardoV94: In that case do we want to keep the bounded continuous class for anything? Maybe to raise an error if transform is UNSET?

pymc3/distributions/continuous.py
ricardoV94: Will take care of here: https://github.com/pymc-devs/pymc3/pull/4653

pymc3/distributions/distribution.py
```diff
@@ -109,13 +109,6 @@ def logcdf(op, var, rvs_to_values, *dist_params, **kwargs):
                     value_var = rvs_to_values.get(var, var)
                     return class_logcdf(value_var, *dist_params, **kwargs)
 
-            # class_transform = clsdict.get("transform")
```
twiecki, line 112: Should just remove this.

pymc3/tests/test_distributions_random.py
```diff
@@ -888,14 +998,14 @@ def ref_rand(size, mu, sigma, nu):
 
         pymc3_random(pm.ExGaussian, {"mu": R, "sigma": Rplus, "nu": Rplus}, ref_rand=ref_rand)
 
-    @pytest.mark.xfail(reason="This distribution has not been refactored for v4")
+    @pytest.mark.skip(reason="This test is covered by Aesara")
```
twiecki: Shouldn't we just remove it then?

pymc3/distributions/distribution.py
ricardoV94: Done

pymc3/tests/test_distributions_random.py
ricardoV94: Done

twiecki (dismissed):

twiecki:
```

=================================== FAILURES ===================================
_______________ TestMatchesScipy.test_beta_binomial_distribution _______________

self = <pymc3.tests.test_distributions.TestMatchesScipy object at 0x12affbeb0>

    @pytest.mark.xfail(condition=(aesara.config.floatX == "float32"), reason="Fails on float32")
    def test_beta_binomial_distribution(self):
>       self.checkd(
            BetaBinomial,
            Nat,
            {"alpha": Rplus, "beta": Rplus, "n": NatSmall},
        )

pymc3/tests/test_distributions.py:1501: 

self = <pymc3.tests.test_distributions.TestMatchesScipy object at 0x12affbeb0>
distfam = <class 'pymc3.distributions.discrete.BetaBinomial'>
valuedomain = <pymc3.tests.test_distributions.Domain object at 0x12511d880>
vardomains = {'alpha': <pymc3.tests.test_distributions.Domain object at 0x12560c7f0>, 'beta': <pymc3.tests.test_distributions.Domain object at 0x12560c7f0>, 'n': <pymc3.tests.test_distributions.Domain object at 0x12511dfd0>}
checks = (<bound method TestMatchesScipy.check_int_to_1 of <pymc3.tests.test_distributions.TestMatchesScipy object at 0x12affbeb0>>,)
extra_args = {}

    def checkd(self, distfam, valuedomain, vardomains, checks=None, extra_args=None):
        if checks is None:
            checks = (self.check_int_to_1,)
    
        if extra_args is None:
            extra_args = {}
        m = build_model(distfam, valuedomain, vardomains, extra_args=extra_args)
        for check in checks:
>           check(m, m.named_vars["value"], valuedomain, vardomains)
E           AttributeError: 'tuple' object has no attribute 'named_vars'

pymc3/tests/test_distributions.py:863: AttributeError
=============================== warnings summary ===============================
```

ricardoV94:
Yeah, it's using this `checkd` test that has not been refactored for V4. Just marked as xfail for now. Will fix later when refactoring the ZeroInflated distributions that make use of it as well.

ricardoV94:
I added a skipif for the betabinomial test in case we are using scipy < 1.4.0 (which happens in the pytest workflow). The tests should still be covered in the arviz-compat workflow so it should be fine for now.

pymc3/distributions/continuous.py
```diff
@@ -120,15 +126,36 @@ def unit_cont_transform(op):
...
-        return lower, upper
+@logp_transform.register(CircularContinuous)
+def circ_cont_transform(op):
+    return transforms.circular
+
+
+class BoundedContinuous(Continuous):
+    """Base class for bounded continuous distributions"""
+
+    transform_args = None
```
brandonwillard: Since this field is very specific to upper and lower bounds, its name should reflect that.  Currently, it seems like this field specifies which arguments are to be transformed, which is very misleading.  Also, the values in this field are argument _indices_, so that should be made clearer somehow (e.g. `upper_lower_bound_indices`, `trans_ul_bound_indices`, etc.).

pymc3/distributions/continuous.py
```diff
@@ -2646,6 +2641,18 @@ def __init__(self, nu, *args, **kwargs):
         super().__init__(alpha=nu / 2.0, beta=0.5, *args, **kwargs)
 
 
+# TODO: Remove this once logpt for multiplication is working!
```
brandonwillard, line 2647: Let's create an issue specifically for this before/after merging.

brandonwillard (changes requested):
Aside from renaming that transform field, it looks like `test_distributions_random` could be introducing a significant amount of testing redundancy.  Regardless, we can merge this once the transform field name is updated.

pymc3/distributions/continuous.py
ricardoV94: Renamed to `bound_args_indices`

pymc3/distributions/continuous.py
ricardoV94: Done: https://github.com/pymc-devs/pymc3/issues/4683

pymc3/distributions/continuous.py
```diff
@@ -3388,45 +3341,18 @@ class Triangular(BoundedContinuous):
...
-        self.lower = lower = at.as_tensor_variable(floatX(lower))
-        self.upper = upper = at.as_tensor_variable(floatX(upper))
-
-        super().__init__(lower=lower, upper=upper, *args, **kwargs)
-
-    def random(self, point=None, size=None):
-        """
-        Draw random values from Triangular distribution.
+    rv_op = triangular
+    bound_args_indices = [0, 2]  # lower, upper
```
twiecki: Why `2`? And shouldn't these be tuples instead of lists?

pymc3/distributions/continuous.py
```diff
@@ -3810,39 +3733,15 @@ class Logistic(Continuous):
...
-        #     stats.logistic.rvs, loc=mu, scale=s, dist_shape=self.shape, size=size
-        # )
+    @classmethod
+    def dist(cls, mu=0.0, s=1.0, *args, **kwargs):
+        mu = at.as_tensor_variable(floatX(mu))
+        s = at.as_tensor_variable(floatX(s))
+        return super().dist([mu, s], *args, **kwargs)
 
-    def logp(self, value):
+    def logp(value, mu, s):
```
twiecki, line 3744: Why don't we need `self` here?

twiecki (dismissed):

pymc3/distributions/continuous.py
ricardoV94: Changed to tuples. 

The arguments of the triangular RandomOp are in this order `(lower, mode, upper)`, and so we need to specify that the first and last are the ones relevant for the transformation, hence `(0, 2)` (zero-based index)

pymc3/distributions/continuous.py
ricardoV94: Because these are used only as a sort of static-method by the `_logp` dispatcher here: https://github.com/pymc-devs/pymc3/blob/6c247638c506e4b3d6eff86ff24c5e98d34ae055/pymc3/distributions/distribution.py#L96-L102

We are not really using distribution instances anymore IIUC.

twiecki: What a curious pattern, thanks for explaining.

twiecki:
This gets us a big step closer to merging `v4`, thanks @ricardoV94!

PR #4641, opened 2021-04-13
Use optimized graph for JAX conversion

brandonwillard:
This PR uses the Aesara optimized graph for JAX sampling.  See the discussion [here](https://gist.github.com/twiecki/a77104299535b64b58953de3c84df56f#gistcomment-3703154).


twiecki:
This is only for the TFP sampler, not the numpyro one.

twiecki:
In fact, we can probably remove the TFP one.

Issue #4530, opened 2021-03-11
Consistent API for user friendly distribution customization


ricardoV94:
The following common "customized distributions" are missing / faulty / only implemented for unobserved variables. It would be nice to have a standard API for when all of these get implemented for both unobserved and observed distributions.

* **Mixture** (not yet in V4, but otherwise working, similar API via `.dist` also present in the LKJCholeskyCov)
*  **Sort** (Behaves a bit like the ordered transform but also works in ancestral sampling? #4213, see @junpenglao thread https://twitter.com/junpenglao/status/1263927315194576898) 
* **Truncate** (Bound works for unobserved only as it doesn't respect the change in the normalization constant, #1864)
* **Censor** (Similar to truncation but values stay "sticked" at the edges, #1864)
* **Shift, Scale** (is there a verb for affine?) (Can be done with verbose deterministics for unobserved variables, not available for observed, #4507, #1924?)
* **Other?**

This issue is intended to discuss a possible API going forward when implementing these custom distributions. I will illustrate with the example of a user who wants to create an **observable** shifted exponential distribution as in #4507. I will call this helper method `pm.Shift` but probably something like `pm.Affine` that allows both shifting and scaling would be better.

```python
# We want to shift by a random normal variable
a = pm.Normal('a', 0, 1)

# Adapt Deterministic syntax. Requires naming the raw variable and probably 
# incompatible with current RV registration logic
x = pm.Shift('x', pm.Exponential('x_raw', 1), shift=a, observed=data)

# Adapt syntax used in pm.Mixture. Requires the non intuitive .dist() call
x = pm.Shift('x', pm.Exponential.dist(1), shift=a, observed=data)

# Adapt syntax from pm.Bound. Not very intuitive. Also distribution and transformation 
# parameters are separated from class Names. Not obvious whether data is expected
# to be on the original or shifted scale (should be on the shifted scale)
x = pm.Shift(pm.Exponential, shift=a)('x', 1, observed=data)

# Add a generic "modify" argument
x = pm.Exponential('x', 1, modify=pm.Shift(shift=a), observed=data)

# Add all necessary arguments to all distributions
# sort = bool
# lower / upper = float (for truncation / censoring)
# shift / scale = float 
x = pm.Exponential('x', 1, shift=a, observed=data)

# Add new methods to RVs and (optionally) separate initialization from conditioning
# to remove ambiguity as to whether the data should be on the shifted scale (should be)
[...]

Issue #4507, opened 2021-03-07
Feature request: Add offset parameter to Exponential class (akin to mu in Laplace class)


jtstanley:
Could we include an additional parameter to the `Exponential` class that allows for the whole distribution to be shifted? This would be equivalent to the `mu` parameter in the `Laplace` class. Given the similarity of the two distributions (meaningfully differing only by their support), most of the changes could simply be copied from the relevant lines in `Laplace`.

A usecase was provided by `@harcel` on the discourse. 
See: https://discourse.pymc.io/t/exponential-likelihood-at-different-location/6919

ricardoV94:
This could potentially be useful for more distributions other than the exponential

jtstanley:
I second @ricardoV94's comment. For example, `HalfNormal`, `HalfCauchy`, `Gamma`, `ChiSquared`.

harcel:
Indeed I was the one suggesting this. I have not yet contributed before, but I would like to, and I think I am capable of fixing this issue (which isn't critical and there probably isn't any hurry, right?). I have found the guidelines for contributing, no issues there, but I would like to have some(-one as) back-up available for noob questions if I commit myself to doing this (imposter and all that...).

harcel:
Also, it seems like location parameters are named according to their standard statistical symbol (often mu, sometimes alpha) and never 'loc', like what is sometimes seen in scipy.stats. The Exponential and the distributions mentioned by @jtstanley above don't typically come with a location parameter (but I don't see a reason why not to allow that, and my discourse post showed a use case for when it could be helpful), so I guess we are free to use whatever symbol we like for it. Some unity would be good, so maybe alpha (given that mu is often reserved for the mean) is a good idea. When alpha is taken, 'loc' could be a good alternative.
Changing the location shifts the average, the logp and the logcdf by alpha and leaves the variance unchanged. The code seems pretty easy to adapt.

ricardoV94:
I liked the argument shift in the discourse. Makes it clear what it means, and since it's doesn't have a conventional greek symbol I don't think calling it alpha or something else would be very useful.

One thing that could be considered is having a generic class `pm.Shift` similar to `pm.Bound` that wraps around a distribution and gives the shifted version:

```python
shifted_exponential = pm.Shift(pm.Exponential, shift=5)
likelihood = shifted_exponential('likelihood', lam=2, observed=x)
```
[...]

PR #4653, opened 2021-04-20
Implement logp for `add` and `mul` ops involving random variables

ricardoV94:
This PR attempts to implement automatic logp for addition and multiplication of "unregistered" random variables (i.e., those defined using `.dist()`)

It contributes towards https://github.com/pymc-devs/pymc3/issues/4530 and could provide a general solution to user requests such as in https://github.com/pymc-devs/pymc3/issues/4507

```python
with pm.Model() as m:
    loc = pm.Normal('loc', 100, 50)
    scale = pm.HalfNormal('scale', 15, transform=None)  # Not working with transformed variables yet
    
    x = loc + pm.Normal.dist(0, 1) * scale
    m.register_rv(x, 'x')  # X works almost as a standard distribution, the next step would be to allow it to be observed

m.logp({'loc':100, 'scale':15, 'x':95})  # array(-11.94734738)


with pm.Model() as m:
    loc = pm.Normal('loc', 100, 50)
    scale = pm.HalfNormal('scale', 15, transform=None)
    x = pm.Normal('x', loc, scale)

m.logp({'loc':100, 'scale':15, 'x':95})  # array(-11.94734738)
   
```

Getting a lot of warnings of this type as well:
```
UserWarning: Variable ... cannot be replaced; it isn't in the FunctionGraph
```

Very far away down this road we could even attempt doing automatic convolution of two unregistered rvs for the expression 
```
x = pm.Normal.dist(mu, sigma) + pm.Normal.dist(mu, sigma)
```

***

### DIFF
--- a/pymc3/distributions/logp.py
+++ b/pymc3/distributions/logp.py
@@ -25,6 +25,8 @@ from aesara.graph.basic import Constant, clone, graph_inputs, io_toposort
 from aesara.graph.fg import FunctionGraph
 from aesara.graph.op import Op, compute_test_value
 from aesara.graph.type import CType
+from aesara.scalar.basic import Add, Mul
+from aesara.tensor.elemwise import Elemwise
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.random.opt import local_subtensor_rv_lift
 from aesara.tensor.subtensor import (
@@ -37,7 +39,12 @@ from aesara.tensor.subtensor import (
 )
 from aesara.tensor.var import TensorVariable
 
-from pymc3.aesaraf import extract_rv_and_value_vars, floatX, rvs_to_value_vars
+from pymc3.aesaraf import (
+    extract_rv_and_value_vars,
+    floatX,
+    rvs_to_value_vars,
+    walk_model,
+)
 
 
 @singledispatch
@@ -260,6 +267,130 @@ def _logp(
     """
     value_var = rvs_to_values.get(var, var)
     return at.zeros_like(value_var)
+    # raise NotImplementedError(f"Logp cannot be computed for op {op}")
+
+
+@_logp.register(Elemwise)
+def logp_elemwise(op, *args, **kwargs):
+    if hasattr(op, "scalar_op"):
+        return _logp(op.scalar_op, *args, **kwargs)
+    raise NotImplementedError
+
+
+# TODO: Implement DimShuffle logp?
+# @_logp.register(DimShuffle)
+# def logp_dimshuffle(op, var, *args, **kwargs):
+#     if var.owner and len(var.owner.inputs) == 1:
+#         inp = var.owner.inputs[0]
+#         if inp.owner and hasattr(inp.owner, 'op'):
+#             return _logp(inp.owner.op, inp, *args, **kwargs)
+#     raise NotImplementedError
+
+
+def find_rv_branch(inputs):
+    """
+    Helper function to find which input branch(es) contain unregistered random variables
+    """
+    rv_branch = []
+    no_rv_branch = []
+
+    for inp in inputs:
+        res_ancestors = list(walk_model((inp,), walk_past_rvs=True))
+        # unregistered variables do not contain a value_var tag
+        res_unregistered_ancestors = [
+            v
+            for v in res_ancestors
+            if v.owner
+            and isinstance(v.owner.op, RandomVariable)
+            and not getattr(v.tag, "value_var", False)
+        ]
+        if res_unregistered_ancestors:
+            rv_branch.append(inp)
+        else:
+            no_rv_branch.append(inp)
+
+    return rv_branch, no_rv_branch
+
+
+@_logp.register(Add)
+def add_logp(op, var, rvs_to_values, *add_inputs, **kwargs):
+
+    if len(add_inputs) != 2:
+        raise ValueError(f"Expected 2 inputs but got: {len(add_inputs)}")
+
+    rv, loc = find_rv_branch(add_inputs)
+
+    if len(rv) != 1:
+        raise NotImplementedError(
+            f"Logp of addition requires one branch with an unregistered RandomVariable but got {len(rv)}"
+        )
+
+    rv = rv[0]
+    rv_value = rvs_to_values.get(rv, getattr(rv.tag, "value_var", rv))
+    loc = loc[0]
+    loc_value = rvs_to_values.get(loc, getattr(loc.tag, "value_var", loc))
+
+    new_rvs_to_values = rvs_to_values.copy()
+    new_rvs_to_values[rv] = rv_value
+
+    logp_rv = logpt(rv, new_rvs_to_values, **kwargs)
+    fgraph = FunctionGraph(
+        [i for i in graph_inputs((logp_rv,)) if not isinstance(i, Constant)],
+        [logp_rv],
+        clone=False,
+    )
+
+    var_value = rvs_to_values.get(var, var)
+
+    fgraph.add_input(loc_value)
+    fgraph.add_input(var_value)
+    fgraph.replace(rv_value, var_value - loc_value)
+
+    logp_rv.name = f"__logp_{var.name}"
+
+    return logp_rv
+
+
+@_logp.register(Mul)
+def mul_logp(op, var, rvs_to_values, *mul_inputs, **kwargs):
+
+    if len(mul_inputs) != 2:
+        raise ValueError(f"Expected 2 inputs but got: {len(mul_inputs)}")
+
+    rv, scale = find_rv_branch(mul_inputs)
+
+    if len(rv) != 1:
+        raise NotImplementedError(
+            f"Logp of product requires one branch with an unregistered RandomVariable but got {len(rv)}"
+        )
+
+    rv = rv[0]
+    rv_value = rvs_to_values.get(rv, getattr(rv.tag, "value_var", rv))
+    scale = scale[0]
+    scale_value = rvs_to_values.get(scale, getattr(scale.tag, "value_var", scale))
+
+    new_rvs_to_values = rvs_to_values.copy()
+    new_rvs_to_values[rv] = rv_value
+
+    logp_rv = logpt(rv, new_rvs_to_values, **kwargs)
+    fgraph = FunctionGraph(
+        [i for i in graph_inputs((logp_rv,)) if not isinstance(i, Constant)],
+        [logp_rv],
+        clone=False,
+    )
+
+    var_value = rvs_to_values.get(var, var)
+
+    fgraph.add_input(scale_value)
+    fgraph.add_input(var_value)
+    # TODO: This is not correct for discrete variables
+    # TODO: Undefined behavior for scale = 0
+    fgraph.replace(rv_value, var_value / scale_value)
+
+    logp_rv = fgraph.outputs[0] - at.log(at.abs_(scale_value))
+    logp_rv.name = f"__logp_{var.name}"
+
+    return logp_rv
 
 
 def convert_indices(indices, entry):
--- a/pymc3/distributions/logp.py
+++ b/pymc3/distributions/logp.py
@@ -318,37 +318,38 @@ def add_logp(op, var, rvs_to_values, *add_inputs, **kwargs):
     if len(add_inputs) != 2:
         raise ValueError(f"Expected 2 inputs but got: {len(add_inputs)}")
 
-    rv, loc = find_rv_branch(add_inputs)
+    base_rv, loc = find_rv_branch(add_inputs)
 
-    if len(rv) != 1:
+    if len(base_rv) != 1:
         raise NotImplementedError(
-            f"Logp of addition requires one branch with an unregistered RandomVariable but got {len(rv)}"
+            f"Logp of addition requires one branch with an unregistered RandomVariable but got {len(base_rv)}"
         )
 
-    rv = rv[0]
-    rv_value = rvs_to_values.get(rv, getattr(rv.tag, "value_var", rv))
+    var_value = rvs_to_values.get(var, var)
     loc = loc[0]
-    loc_value = rvs_to_values.get(loc, getattr(loc.tag, "value_var", loc))
+    base_rv = base_rv[0]
+    base_value = base_rv.type()
 
-    new_rvs_to_values = rvs_to_values.copy()
-    new_rvs_to_values[rv] = rv_value
-
-    logp_rv = logpt(rv, new_rvs_to_values, **kwargs)
+    logp_base_rv = logpt(base_rv, {base_rv: base_value}, **kwargs)
     fgraph = FunctionGraph(
-        [i for i in graph_inputs((logp_rv,)) if not isinstance(i, Constant)],
-        [logp_rv],
+        [i for i in graph_inputs((logp_base_rv,)) if not isinstance(i, Constant)],
+        [logp_base_rv],
         clone=False,
     )
+    fgraph.replace(base_value, var_value - loc, import_missing=True)
+    logp_add_rv = fgraph.outputs[0]
 
-    var_value = rvs_to_values.get(var, var)
+    # Replace rvs in graph
+    # TODO: This shouldn't be here
+    (logp_add_rv,), _ = rvs_to_value_vars(
+        (logp_add_rv,),
+        apply_transforms=True,  # Change this
+        initial_replacements=None,
+    )
 
-    fgraph.add_input(loc_value)
-    fgraph.add_input(var_value)
-    fgraph.replace(rv_value, var_value - loc_value)
+    logp_add_rv.name = f"__logp_{var.name}"
 
-    logp_rv.name = f"__logp_{var.name}"
-
-    return logp_rv
+    return logp_add_rv
 
 
 @_logp.register(Mul)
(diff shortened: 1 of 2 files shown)


brandonwillard:
> Getting a lot of warnings of this type as well:
> 
> ```
> UserWarning: Variable ... cannot be replaced; it isn't in the FunctionGraph
> ```

You can ignore those; they're not errors&mdash;although they might help explain why/when certain graph related operations aren't working the way one expects.

pymc3/distributions/logp.py
```diff
@@ -260,6 +267,130 @@ def _logp(
...
+            if v.owner
+            and isinstance(v.owner.op, RandomVariable)
+            and not getattr(v.tag, "value_var", False)
+        ]
+        if res_unregistered_ancestors:
+            rv_branch.append(inp)
+        else:
+            no_rv_branch.append(inp)
+
+    return rv_branch, no_rv_branch
```
brandonwillard: Walking the `Add` inputs might be too much.  We need to do our graph walking with `logpt`/`_logp`; otherwise, we'll miss nested "changes of variables" (e.g. `1 + Y * 2`).

Also, why does this treat "registered" `RandomVariable`s differently from "unregistered" ones?

pymc3/distributions/logp.py
```diff
@@ -260,6 +267,130 @@ def _logp(
...
+
+
+@_logp.register(Add)
+def add_logp(op, var, rvs_to_values, *add_inputs, **kwargs):
+
+    if len(add_inputs) != 2:
+        raise ValueError(f"Expected 2 inputs but got: {len(add_inputs)}")
+
+    rv, loc = find_rv_branch(add_inputs)
```
brandonwillard: This part does seem to be the/a challenge: separate the `Add` arguments into ones that are effectively constant, and ones that are random.

brandonwillard (commented):
I think we need to clarify exactly how we're approaching log-likelihood graph construction.  This has nothing to do with these changes, but the process in general, because there might be a much more straightforward way for us to organize and implement everything.

In the simplest case, we need to construct conditional log-likelihood graphs.  This is easy, and what `logpt` does when it's given a `RandomVariable` for `var`.

Extensions like the one in this PR construct log-likelihood graphs when `var` **isn't** a `RandomVariable`, but, instead, is a _graph_ that potentially contains `RandomVariable`s.

In this scenario, with no other information, one of the only reasonable log-likelihood graphs one could construct is a joint log-likelihood for all the `RandomVariable`s in the graph of `var`.

Due to this, the current `logpt` interface is confusing.  

My first thought is to separate the two and have `logpt` only ever compute a conditional log-likelihood (i.e. `var` must be a `RandomVariable`), and create a distinct entry point for the computation of a joint/total log-likelihood for an arbitrary graph.  Let's call it `total_logpt`.  In this case, `_logp` would only ever dispatch on `RandomVariable` types, which makes things simpler.

Once that is done, some steps in PyMC3&mdash;perhaps exclusively within `Model`&mdash;will need to be changed so that `total_logpt(observed_RVs, ...)` is used instead of `sum([logpt(rv, ...) for rv in observed_RVs])`; otherwise, the current missing data features will break.  This change is pretty simple, though.

Now, where does that leave us?  Under `total_logpt`, we can easily extract all the `RandomVariable`s and get their log-likelihood graphs with `logpt` (and sum those).  More importantly, when one of the observed variable arguments is not simply a `RandomVariable`, we can look at the graph as a whole&mdash;in isolation&mdash;and perform operations on it.

For instance, if we get an observed variable with the form `a + Y * b = g(Y) = Z`, we will know the underlying density for `Y` and be able to match the transform `g` with a set of known/implemented inverses to obtain `g_inv` (or not).  Now, we have everything we need to produce the log-likelihood of `Z`!

This is a much more scalable and clear cut approach than the one I've currently hacked together.  I don't want to send you down a rabbit hole because of that, so let's make sure the above approach makes sense and, if it does, let's start implementing that right away.


N.B.: The approach above also accounts for the entire `Transform`s framework, so we'll need to address that afterward.

pymc3/distributions/logp.py
ricardoV94: > Walking the `Add` inputs might be too much.  We need to do our graph walking with `logpt`/`_logp`; otherwise, we'll miss nested "changes of variables" (e.g. `1 + Y * 2`).
> 

I am actually walking it with logpt inside add and mul. This is just to decide which branch (if any) is worth walking over.

And the nested Y = a + x * b seems to be working fine


> Also, why does this treat "registered" `RandomVariable`s differently from "unregistered" ones?

This was a hackish way of deciding which branch needs to be walked to find the logpt of a new rv and which can be treated (correctly) as a constant.

The thing about registered and unregistered rvs is exactly that and the goal was to not double count an independent variable that was created upstream and instead treat it as a constant into this logpt.

```python
with pm.Model() as m:
  # registered rv which will be in the model logpt 
  # and can be used as a constant downstream
  a = pm.Normal('a', 0, 1)

  # unregistered rv, can be used to build 
  # new "arbitrary" logp expressions
  b = pm.Normal.dist(0, 1)

  # such as this
  y = a + b * 5
```

In the add logp it would detect the branch on the right `b * 5` is where there is a potential base source for a logp, and that the one on the left will just "shift" it.

pymc3/distributions/logp.py
ricardoV94: That's what the helper function above tries to do. Curious why you think it doesn't make sense. Are you thinking of more complicated examples than the add mul I had in mind here?

twiecki:
This looks amazing!

I wonder about the API, using the model object in this way is fairly inconsistent with how we do it. Something like
```Python
with pm.Model() as m:
    loc = pm.Normal('loc', 100, 50)
    scale = pm.HalfNormal('scale', 15, transform=None)  # Not working with transformed variables yet
    
    x = loc + pm.Normal.dist(0, 1) * scale
    pm.RandomVariable('x', x)  # X works almost as a standard distribution, the next step would be to allow it to be observed
```

would be more in-line with the rest of the API. Not sure about the right name yet, however.

ricardoV94:
Yeah definitely not the register thing (I just used it because that's how we do it in the backed and it's already working)

Also it's possible to add specific helper functions such as `pm.Afffine(pm.Normal.dist(0, 1), loc=a, scale=b)` that work a bit more like our old `pm.Mixture`.

For the general case (ie just put an expression and hope that PyMC can figure it out) something like `pm.RandomVariable` sounds rather nice actually. Other suggestions involved `pm.Deterministic` (which could now be observed) but that sounds weird for random variables or a `pm.Observed()` but maybe that's also too specific sounding?

twiecki:
Out of the 3 I prefer `pm.RandomVariable`, but curious what others think.

Under this framework, how would a Gaussian Random Walk look like?

`pm.Normal.dist().cumsum()`? Could we somehow turn this into a dist itself so that it could be used just like the current `pm.GaussianRandomWalk`?

The `AffineTransform` is also really neat.

ricardoV94:
I hadn't thought about the random walk case, but that also seems it could fit within this WIP "framework" rather nicely. My understanding is that a general random walk is just an x0 + innovations right? 

```python
y = pm.Flat.dist() + pm.Normal.dist(sigma=2).cumsum()
y = pm.RandomVariable('y',  y, observed=data)
```
Seems to make sense indeed. This could be implemented via a `logp` for the `cumsum` op (and allowing `add` to work in this case).

(Actually in V3, I don't know whether the init point is integrated out in the RandomWalk distribution or modeled as an explicit separate RV...)

In any case the current framework still needs to be figured out, but I think this kind of functionality would be a nice target to aim for. We can then always have helper functions that only have to add the equivalent expression behind the scenes if that's more intuitive for some users.

```python
y = pm.RandomWalk('y', init=pm.Flat.dist(), dist=pm.Normal.dist(sigma=2), observed=data)
```
Which would return the following equivalent expression from above. 

I will add this example to the discussion in: #4530  

PS: Instead of `pm.RandomVariable`, something like `pm.Variable` or even `pm.Var` might be more pleasant to write.

twiecki:
`pm.Var` is nice.

brandonwillard:
> We can then always have helper functions that only have to add the equivalent expression behind the scenes if that's more intuitive for some users.

That's precisely the plan.

pymc3/distributions/logp.py
```diff
@@ -260,6 +267,132 @@ def _logp(
     """
     value_var = rvs_to_values.get(var, var)
     return at.zeros_like(value_var)
+    # raise NotImplementedError(f"Logp cannot be computed for op {op}")
+
+
+@_logp.register(Elemwise)
+def logp_elemwise(op, *args, **kwargs):
+    if hasattr(op, "scalar_op"):
```
brandonwillard: This should always be present.

pymc3/distributions/logp.py
```diff
@@ -260,6 +267,132 @@ def _logp(
     """
     value_var = rvs_to_values.get(var, var)
     return at.zeros_like(value_var)
+    # raise NotImplementedError(f"Logp cannot be computed for op {op}")
+
+
+@_logp.register(Elemwise)
+def logp_elemwise(op, *args, **kwargs):
```
brandonwillard:
```suggestion
def elemwise_logp(op, *args, **kwargs):
```

pymc3/distributions/logp.py
```diff
@@ -260,6 +267,132 @@ def _logp(
...
+
+    var_value = rvs_to_values.get(var, var)
+    loc = loc[0]
+    base_rv = base_rv[0]
+    base_value = base_rv.type()
+
+    logp_base_rv = logpt(base_rv, {base_rv: base_value}, **kwargs)
+    fgraph = FunctionGraph(
+        [i for i in graph_inputs((logp_base_rv,)) if not isinstance(i, Constant)],
+        [logp_base_rv],
```
brandonwillard:
```suggestion
        outputs=[logp_base_rv],
```

brandonwillard (changes requested):
`add_logp` and `mul_logp` share a lot of the same code; can you create one function that handles both cases?

pymc3/distributions/logp.py
ricardoV94: Removed the check

pymc3/distributions/logp.py
ricardoV94: Changed

pymc3/distributions/logp.py
ricardoV94: Changed

ricardoV94:
I combined the two logps. For the `mul` jacobian correction, I check if the `base_rv` is of `dtype` `float`, to decide whether to add it or not. Not sure whether our discrete logps will deal well with rounding on the `values` or not. An alternative might be to raise `NotImplemented` if the `base_rv` `dtype` is `int`.

ricardoV94:
More important, how should we deal with size issues (and `int/float` casting)? Should this fall under the responsibility of the `_logp` dispatcher?

This snippet:
```python
with pm.Model() as m:
    scale = pm.HalfNormal('scale', 15, size=2, transform=None)
    x = pm.Normal.dist(0, 1) * scale
    m.register_rv(x, 'x')
m.logp({'scale': [1, 1], 'x': [0, 0]})
```
raises the following `TypeError`
```python
File "/home/ricardo/Documents/Projects/pymc3/pymc3/distributions/logp.py", line 345, in linear_logp
    fgraph.replace(base_value, var_value / constant, import_missing=True)
  File "/home/ricardo/Documents/Projects/pymc3-venv/lib/python3.8/site-packages/aesara/graph/fg.py", line 515, in replace
    new_var = var.type.filter_variable(new_var, allow_convert=True)
  File "/home/ricardo/Documents/Projects/pymc3-venv/lib/python3.8/site-packages/aesara/tensor/type.py", line 258, in filter_variable
    raise TypeError(
TypeError: Cannot convert Type TensorType(float64, vector) (of Variable Elemwise{true_div,no_inplace}.0) into Type TensorType(float64, (True,)). You can try to manually convert Elemwise{true_div,no_inplace}.0 into a TensorType(float64, (True,)).
```

brandonwillard:
> This snippet:
> 
> ```python
> with pm.Model() as m:
>     scale = pm.HalfNormal('scale', 15, size=2, transform=None)
>     x = pm.Normal.dist(0, 1) * scale
>     m.register_rv(x, 'x')
> m.logp({'scale': [1, 1], 'x': [0, 0]})
> ```
> 
> raises the following `TypeError`

~I just tried this after rebasing onto `v4`, and it didn't throw the `TypeError`, so this might've been due to https://github.com/pymc-devs/aesara/issues/390.~   No, I _am_ able to reproduce it.

pymc3/distributions/logp.py
```diff
@@ -260,6 +267,86 @@ def _logp(
...
+
+    base_rv = base_rv[0]
+    constant = constant[0]
+    var_value = rvs_to_values.get(var, var)
+
+    # Get logp of base_rv
+    base_value = base_rv.type()
+    logp_base_rv = logpt(base_rv, {base_rv: base_value}, **kwargs)
+    fgraph = FunctionGraph(
+        [i for i in graph_inputs((logp_base_rv,)) if not isinstance(i, Constant)],
```
brandonwillard:
```suggestion
```
Leaving `input`s unspecified and setting `outputs=...` should automatically compute the inputs in exactly the same way.

pymc3/distributions/logp.py
```diff
@@ -260,6 +267,86 @@ def _logp(
...
+    elif isinstance(op, Mul):
+        fgraph.replace(base_value, var_value / constant, import_missing=True)
+        logp_linear_rv = fgraph.outputs[0]
+        if "float" in base_rv.dtype:
+            logp_linear_rv -= at.log(at.abs_(constant))
+
+    # Replace rvs in graph
+    (logp_linear_rv,), _ = rvs_to_value_vars(
+        (logp_linear_rv,),
+        apply_transforms=kwargs.get("transformed", True),
```
brandonwillard: If this is a locally-used keyword, we should make it explicit in the function signature.  It will need to be used explicitly in the call to `logpt`, as well.

pymc3/distributions/logp.py
```diff
@@ -260,6 +267,86 @@ def _logp(
...
+#             return _logp(inp.owner.op, inp, *args, **kwargs)
+#     raise NotImplementedError
+
+
+@_logp.register(Add)
+@_logp.register(Mul)
+def linear_logp(op, var, rvs_to_values, *linear_inputs, **kwargs):
+
+    if len(linear_inputs) != 2:
+        raise ValueError(f"Expected 2 inputs but got: {len(linear_inputs)}")
```
brandonwillard, line 302: This is an implementation limitation and not a misspecification of any kind, so, if we're not going to support more than two arguments for `Add` and/or `Mul`, we should probably make this a `NotImplementedError`.

brandonwillard (changes requested):
> This snippet:
> 
> ```python
> with pm.Model() as m:
>     scale = pm.HalfNormal('scale', 15, size=2, transform=None)
>     x = pm.Normal.dist(0, 1) * scale
>     m.register_rv(x, 'x')
> m.logp({'scale': [1, 1], 'x': [0, 0]})
> ```
> 
> raises the following `TypeError`

This is the entire graph of the `var` argument to `linear_logp` in the above case:
```python
Elemwise{mul,no_inplace} [id A] 'x'   
 |InplaceDimShuffle{x} [id B] ''   
 | |normal_rv.1 [id C] ''   
 |   |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F8BD66C5640>) [id D]
 |   |TensorConstant{[]} [id E]
 |   |TensorConstant{11} [id F]
 |   |TensorConstant{0} [id G]
 |   |TensorConstant{1.0} [id H]
 |halfnormal_rv.1 [id I] 'scale'   
   |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F8BD66C5040>) [id J]
   |TensorConstant{(1,) of 2} [id K]
   |TensorConstant{11} [id L]
   |TensorConstant{0.0} [id M]
   |TensorConstant{15.0} [id N]
```

You're replacing the node labeled `B` with a non-broadcastable vector (one that actually has a length equal to two).  The `DimShuffle` is explicitly stating that the output has a single broadcastable dimension, so the replacement definitely shouldn't work.

First off, it looks like the `fgraph` doesn't contain the actual log-likelihood for the `normal` term.  I see that you have a commented-out `DimShuffle` `_logp` implementation.  Instead of doing that, lift the `DimShuffle` `Op` to get a `RandomVariable` with new, `DimShuffle`d parameters.  You can use `aesara.tensor.random.opt.local_dimshuffle_rv_lift` to do that.  Just remember that the result is an entirely new `RandomVariable` that has no connection with the original, unlifted one.

NB: Ultimately, what we need is a generalized set of rewrites that "push" `RandomVariable`s down so that they're closer to the/a output nodes of a graph.  I have a simple version of something like that [here](https://github.com/AmpersandTV/pymc3-hmm/blob/651c34a2bb5865d07206023b865c217fc93d6bc6/pymc3_hmm/distributions.py#L86), where it's used to semi-manually derive log-likelihoods for switching distributions.

Second, notice that you're using the relation `Z = X / Y`, where `X` is essentially a scalar random variable and `Y` is a (conditionally) constant vector.  This means that you're replacing a scalar with a vector&mdash;although Aesara normalizes the dimensions, it doesn't actually **broadcast** both arguments to `Mul`, so you're still faced with a type of dimension/shape mismatch. 

You can probably avoid all this by using `X = Z / Y` as the value variable in the call to `logpt`.  A `Z` can be created from the `TensorType` of `var` itself, since it's effectively `Z`.  That approach will remove the need to replace anything in the graph.  Plus, you could even create/specify a transform for the value variable and have all the Jacobian stuff taken care of in `logpt`.  This would simplify the implementation a lot.

pymc3/distributions/logp.py
ricardoV94: Aren't `Add/Mul` binary input ops by definition? I put this just as a sanity check.

pymc3/distributions/logp.py
ricardoV94: This was just temporarily. I had the gut feeling this replacement shouldn't be inside this method to begin with, but I don't have a better place to put it. Feels like things would be a bit more clean if we separated the rewrite in two steps:
1. Introduce all logp terms in the graph (keeping the rvs as inputs)
2. Replace all rvs by value_vars (and only now apply transforms)

My confidence that this would be helpful is not much though :P

pymc3/distributions/logp.py
ricardoV94: Didn't know, thanks

pymc3/distributions/logp.py
```diff
@@ -258,8 +265,80 @@ def _logp(
     The default assumes that the log-likelihood of a term is a zero.
 
     """
-    value_var = rvs_to_values.get(var, var)
-    return at.zeros_like(value_var)
+    # value_var = rvs_to_values.get(var, var)
+    # return at.zeros_like(value_var)
+    raise NotImplementedError(f"Logp cannot be computed for op {op}")
```
ricardoV94, line 271: This is temporary, just to catch errors with DimShuffle ops. It causes the `test_logpt_subtensor` to fail

pymc3/distributions/logp.py
brandonwillard: The scalar `Op`s might be, but the `Elemwise` aren't necessarily.

twiecki:
Why did you close this?

ricardoV94:
It still needs quite some work before it becomes more functional (specially concerning non-scalar random variable sizes and dimshuffle operations), and it will be easier to figure out/ implement in https://github.com/aesara-devs/aeppl and then backport to PyMC3 with the new rv-to-logp framework that is being implemented there. 

Also I was going through my PRs that were still pointing to V4 and cleaning them up.

Issue #4652, opened 2021-04-20
Column vector breaks `observed`


michaelosthege:
While working on #4625 I ran into a failure that is demoed by the following example:

```python
with pm.Model() as model:
    pm.Normal("x1", mu=0, sd=1, observed=np.random.normal(size=(3, 4)))
    model.logp()
    pm.Normal("x2", mu=0, sd=1, observed=np.random.normal(size=(3, 1)))
    model.logp()
```

Essentially passing a column vector breaks the model, whereas a matrix is fine.

<details><summary>Traceback</summary>

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-428a24d8e8bb> in <module>
----> 1 test_observed_with_column_vector()

<ipython-input-3-7765b483c18e> in test_observed_with_column_vector()
      4         model.logp()
      5         pm.Normal("x2", mu=0, sd=1, observed=np.random.normal(size=(3, 1)))
----> 6         model.logp()

e:\source\repos\pymc3\pymc3\model.py in logp(self)
    264     def logp(self):
    265         """Compiled log probability density function"""
--> 266         return self.model.fn(self.logpt)
    267 
    268     @property

e:\source\repos\pymc3\pymc3\model.py in logpt(self)
    712         with self:
    713             factors = [logpt_sum(var, self.rvs_to_values.get(var, None)) for var in self.free_RVs]
--> 714             factors += [logpt_sum(obs, obs.tag.observations) for obs in self.observed_RVs]
    715 
    716             # Convert random variables into their log-likelihood inputs and

e:\source\repos\pymc3\pymc3\model.py in <listcomp>(.0)
    712         with self:
    713             factors = [logpt_sum(var, self.rvs_to_values.get(var, None)) for var in self.free_RVs]
--> 714             factors += [logpt_sum(obs, obs.tag.observations) for obs in self.observed_RVs]
    715 
    716             # Convert random variables into their log-likelihood inputs and

e:\source\repos\pymc3\pymc3\distributions\logp.py in logpt_sum(*args, **kwargs)
    366     if only the sum of the logp values is needed.
    367     """
--> 368     return logpt(*args, sum=True, **kwargs)

e:\source\repos\pymc3\pymc3\distributions\logp.py in logpt(var, rv_values, jacobian, scaling, transformed, cdf, sum, **kwargs)
    216     replacements.update({rv_var: rv_value, rv_value_var: rv_value})
    217 
[...]

PR #4625, opened 2021-04-08
Implement flexible shape/dims/size API

michaelosthege:
## Terminology
+ _implied_ dimensions are those that a RV takes without specifying any of `shape,dims,size,observed`.
+ The `size` of an RV are all dimensions _in addition to_ (preceeding) the _implied_ dimensions.

## Key Takeaways
+ `shape` can be numeric/symbolic and supports `Ellipsis` in the last position. Without `Ellipsis` a `SpecifyShape` `Op` is added automatically.
+ `dims` can be `None` and are defined at first mention. `Ellipsis` is also supported. In combination with `pm.Data` the dimensions specified through `dims` are re-sizeable.
+ `size` does care about implied dimensions in any way and just adds new dimensions to the RV. Internally `size` is inferred from `shape` and `dims`.
+ No more resizing in `Model.make_obs_var` because it leaves code in `Distribution.__new__` not knowing the output shape.

## ToDo
+ [x] `Distribution.dist()` and `Distribution.__new__` are similar in use (`dims` and `observed` only through `__new__`)
+ [x] Inputs are checked and informative errors are raised where possible
+ [x] Release notes were updated
+ [x] Some tests were edited to use `size`/`shape` appropriately -- some tests illegally used `size` to specify size _and implied dimensions_.

## ToDo (optional)
+ [ ] Ideally, it should be possible to pass `dims` in combination with `size`. We should just require a single source of truth (dim names must be undefined for dimensions that `size` integers apply to).

PR #4654, opened 2021-04-20
Add a test for observed data with a broadcast dimension

brandonwillard:
This PR adds a test for the issue described in https://github.com/pymc-devs/pymc3/issues/4652.

Do not merge this; it's only for external confirmation of the issue.


michaelosthege:
A similar test was added as XFAIL in #4625 and is changed to be OS-conditionally XFAIL in #4667.
I guess we can close this?

Issue #4665, opened 2021-04-24
Enabling vector autoregressions


ckrapu:
The `AR` distribution appears to be nearly complete for usage as a true vector autoregression parameterized by `p` cross-series coefficients, each of shape `(d,d)`. The main change that has to be enacted is to use a dot product instead of elementwise multiplication [here](https://github.com/pymc-devs/pymc3/blob/d7172c0a1a76301031d1b3b411d00643c416a0c4/pymc3/distributions/timeseries.py#L161). However, I am unable to determine the role of the `constant` argument and why it necessitates the calculation of `eps = value[self.p :] - self.rho[0] - x` where, under the AR / VAR model, `eps` is assumed to have a diagonal Normal distribution.

ricardoV94:
There is some code example of Vector Autoregression here: https://www.pymc-labs.io/blog-posts/bayesian-vector-autoregression/

>  However, I am unable to determine the role of the constant argument and why it necessitates the calculation of `eps = value[self.p :] - self.rho[0] - x`

The constant is just an intercept term: `y = rho[0] + x`, where `x` are the convolved values and lagged coefficients from `rho[1:]`. The AR was refactored to V4 in #5734

PR #4666, opened 2021-04-24
VAR update

ckrapu:
This PR modifies the `AR` distribution to accommodate cross-series coefficients as desired in a vector autoregressive model. More information can be found at https://discourse.pymc.io/t/how-to-implement-vector-autoregression-with-interdependencies/4287/2 and #4665. This is marked as a WIP as I would like some advice on how to handle the `constant` argument for the `AR` distribution.

Depending on the dimensions of `rho`, the `logp` function will either use an elementwise operation or a matrix multiplication for the independent and dependent time series cases, respectively. However, models which assume 2 or more dimensions indexing independent time series will no longer behave the same. In this case, the recommended approach is to reshape these two dimensions into a single one so that a `rho` tensor with shape `[lags, d, d]` can be interpreted as using multiple `d` x `d` cross-series coefficients.

There is a `constant` argument for the `AR` distribution which I am unable to decipher and have not been able to address the case in which `constant` is used.
No new tests or doc modifications have been added yet.

### DIFF
--- a/pymc3/distributions/timeseries.py
+++ b/pymc3/distributions/timeseries.py
@@ -97,7 +97,9 @@ class AR(distribution.Continuous):
     Parameters
     ----------
     rho: tensor
-        Tensor of autoregressive coefficients. The first dimension is the p lag.
+        Tensor of autoregressive coefficients. The first dimension is the p lag. For allowing
+        cross-series dependencies, rho must have 3 dimensions. For example, a rho of shape (2,3,3)
+        indicates that 2 lags are to be used for 3 series, each assumed to be dependent upon the other 2.
     sigma: float
         Standard deviation of innovation (sigma > 0). (only required if tau is not specified)
     tau: float
--- a/pymc3/distributions/timeseries.py
+++ b/pymc3/distributions/timeseries.py
@@ -156,18 +156,39 @@ class AR(distribution.Continuous):
         -------
         TensorVariable
         """
+
+        if self.rho.ndim > 2:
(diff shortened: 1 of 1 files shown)


mjhajharia:
from what i understand, only example notebooks and implementation details are left right?

mjhajharia:
are there existing notebooks that need to be updated or do we need to create new ones? @ckrapu i can take this up if you can tell me how far stuff is done and what would you have liked to do further :)

pymc3/distributions/timeseries.py
```diff
@@ -156,18 +158,39 @@ def logp(self, value):
...
+            # for each time lag
+            bcast_fn = at.dot
+            if self.constant:
+                raise NotImplementedError(
+                    "Cross-series lags not currently supported with constant=True"
+                )
+        else:
+            bcast_fn = at.mul
+
         if self.constant:
```
ckrapu, line 173: @almostmeenal The reason I paused on this PR was because I couldn't quite figure out exactly what the `self.contant` option was implementing. I essentially put in what looked to me like the broadcasted equivalent of it, but I'd like to figure it out some more.

ckrapu:
Also, you are definitely welcome to take over this PR. For my part, I am going to try to dig up the reason for that formulation and I'll leave the answer here if I find it.

mjhajharia:
mentioning #5291, so we remember this PR once the new AR dist is made

ricardoV94:
Closing this as it went stale. Feel free to open a new PR on top of the newer V4 code ;)

Issue #4662, opened 2021-04-24
Restore pre-4625 way how the `size` kwarg works


michaelosthege:
Due to a (big) misunderstanding, my #4625 PR broke the way how `size` parametrization works in `v4`.

The misunderstanding was specifically about the notion of implied dimensions and what it means for univariate vs. multivariate RVs:

⚠ The pre-#4625 notion was that `size` is in addition to the _support dimensions_.

⚠ The post-#4625 notion was that `size` is in addition to what's implied by parameters other than `size`/`shape`/`dims`.

The difference is subtle and maybe best explained with the following example:

```python
# v4 pre-4625 (45cb4ebf36500e502481bdced6980dd9e630acca)
MvNormal.dist(
    cov=eye(7), mu=ones(7), size=(2,3)    # MvNormal is multivariate with ndim_support=1,
).eval().shape == (2, 3, 7)               # therefore (7,) is implied/required and does not count into `size`.
Normal.dist(
    mu=[1,2,3], size=(2,3)
).eval().shape == (2,3)                   # Normal is univariate, so `mu` does not count as a support dim

# v4 post-4625 (e9f2e9616394275ccf7587a4818fe21251d51328)
MvNormal.dist(
    cov=eye(7), mu=ones(7), size=(2,3)
).eval().shape == (2, 3, 7)
Normal.dist(
    mu=[1,2,3], size=(2,3)
).eval().shape == (2, 3, 3)         # the last dimension of length 3 was implied by mu and does not count into `size`
```

With the changes from #4625 the outcome from specifying `shape=(1,2,3, ...)` and `size=(1,2,3)` is identical.

After some discussion about the advantages/disadvantages of either API flavor, we decided to go back to the pre-4625 flavor where `size` is essentially `shape` but without support dimensions.
This is also the way how numpy handles dimensionality of multivariate distributions:
```
np.random.mvnormal(
    cov=np.eye(7), mean=np.ones(7), size=(2, 3)
).shape == (2, 3, 7)
```

The flexibility added by #4625, namely the ability to _not_ specify dimensions that are implied from RV support or parameters, will continue to work through the `shape` with Ellipsis API:
```python
mu = aesara.shared([1, 2, 3])

rv = Normal.dist(
    mu=mu, shape=(7, 5, ...)   # only the additional dimensions are specified explicitly
)
assert rv.eval().shape == (7, 5, 3)

# Now change the parameter-implied dimensions:
mu.set_value([1, 2, 3, 4])
[...]

PR #4667, opened 2021-04-24
Revert `size` kwarg behaviour and make it work with `Ellipsis`

michaelosthege:
See #4662 for more information.

I will comment key changes in the diff.

### DIFF
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -26,7 +26,6 @@ from aesara.graph.basic import Apply
 from aesara.graph.op import Op
 from aesara.tensor.nlinalg import det, eigh, matrix_inverse, trace
 from aesara.tensor.random.basic import MultinomialRV, dirichlet, multivariate_normal
-from aesara.tensor.random.utils import broadcast_params
 from aesara.tensor.slinalg import (
     Cholesky,
     Solve,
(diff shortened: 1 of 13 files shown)


pymc3/tests/test_transforms.py
```diff
@@ -274,11 +274,11 @@ def test_chain_jacob_det():
 
 
 class TestElementWiseLogp(SeededTest):
-    def build_model(self, distfam, params, *, size=None, shape=None, transform=None, testval=None):
+    def build_model(self, distfam, params, *, size=None, transform=None, testval=None):
```
michaelosthege, line 277: All changes in this file are now back to pre-#4625

pymc3/distributions/multivariate.py
```diff
@@ -427,27 +426,6 @@ def _distr_parameters_for_repr(self):
         return ["a"]
 
 
-class MultinomialRV(MultinomialRV):
-    """Aesara's `MultinomialRV` doesn't broadcast; this one does."""
```
michaelosthege, line 431: The Aesara `MultinomialRV` can do broadcasting since v2.0.4.

pymc3/tests/test_distributions.py
```diff
@@ -1986,29 +1986,33 @@ def test_multinomial_mode(self, p, n):
...
         with Model() as model:
             m = Multinomial("m", n=n, p=p, size=size)
-
-        assert m.eval().shape == size + p.shape
+        # The support has length 4 in all test parametrizations!
+        # Broadcasting of the `p` parameter does not affect the ndim_supp
+        # of the Op, hence the broadcasted p must be included in `size`.
+        support_shape = (p.shape[-1],)
+        assert support_shape == (4,)
+        assert m.eval().shape == size + support_shape
```
michaelosthege, line 2015: The parametrization `sizes` were changed such that the support shape `(4,)` does not occur in the `size` which should _exclude_ the support. This way the test parametrization is slightly less confusing.

The broadcasting of the `MultinomialRV` does not change the fact that it has `ndim_supp == 1`.

michaelosthege:
The broadcasting of multivariate RVs is *very* weird already at the Aesara level:
```
aesara.tensor.random.basic.multivariate_normal(
    mean=np.ones((3, 9)), cov=np.eye(9), size=(3,)
).eval().shape
>>> expected: (3, 9)
>>> actual: (3, 3, 9)
```
The above parametrization apparently gave it a 2-dimensional support? That's a problem, because the `ndim_supp` attribute belongs to the `Op` and is still at `1`.
The other interpretation - keeping support fixed at 1 - is that it decided to replicate twice? Also doesn't make a lot of sense to me.


I noticed this when I tried to refactor the `Distribution.dist()` code such that `size` is eagerly passed instead of creating `rv_native` with `size=None`.
I attempted this refactor because it seemed more elegant, but in the resize-based implementation seems to be way more robust.

brandonwillard:
This can be rebased onto #4693 and, instead, re-introduce a non-breaking version of the `shape` keyword.

twiecki:
Closing in favor of #4696.

PR #4672, opened 2021-04-30
Fix bug in pseudo_likelihood computation

aloctavodia:
log_pseudolikelihood computations were wrong as the indexes were not properly resampled.


ricardoV94:
Should we add a test?

aloctavodia:
I do not have a very good test for checking that the pseudo_likelihoods values are good, I am not sure about which distribution they should follow.  I could check something like the variance is not too big.

aloctavodia:
@ricardoV94 what do you think of the test I added?

junpenglao (approved):

ricardoV94:
It's not obvious for me what the test actually does. Would it have failed before the new patch?

aloctavodia:
Yes, it would have failed as the pseudo-likelihood values would have had a wider distribution.

ricardoV94 (approved):
I see, LGTM

Issue #4670, opened 2021-04-29
Fix numpy deprecation warnings


twiecki:
```../../../../../usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/scalar/basic.py:2416
  /usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/scalar/basic.py:2416: DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    self.ctor = getattr(np, o_type.dtype)

../../../../../usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/scipy/special/orthogonal.py:81
../../../../../usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/scipy/special/orthogonal.py:81
  /usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/scipy/special/orthogonal.py:81: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    from numpy import (exp, inf, pi, sqrt, floor, sin, cos, around, int,

../../../../../usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/scipy/linalg/__init__.py:212
  /usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/scipy/linalg/__init__.py:212: DeprecationWarning: The module numpy.dual is deprecated.  Instead of using dual, use the functions directly from numpy or scipy.
    from numpy.dual import register_func

../../../../../usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/tensor/basic.py:265: 59 warnings
pymc3/tests/test_idata_conversion.py: 3983 warnings
[...]

PR #4673, opened 2021-04-30
Fixing np.bool Depreciation warnings: 3 files

larryshamalama:
Addressing issue [4670](https://github.com/pymc-devs/pymc3/issues/4670)

Not sure where the test warnings come from. I changed "np.bool" to "np.bool_" in 3 files as the warning suggests.


twiecki:
Thanks for taking this on! The warning says to just replace it with `bool`, rather than `np.bool_`.

larryshamalama:
Oh, I see, I assumed that numpy data types were preferred because of the other key/value pairs in `stats_dtypes`. Is `bool` preferred over `np.bool_`?

twiecki:
Seems like `np.bool` was just an alias to `bool` anyway, so let's just stick with that.

larryshamalama:
Fixed! (+ other files not mentioned in the error shown in the original issue)

michaelosthege (approved):

twiecki (approved):

twiecki:
Thanks @larryshamalama and congrats on your first PR! 🥳

larryshamalama:
Thank you very much! 🙂

PR #4675, opened 2021-05-03
BART with non-gaussian likelihoods

aloctavodia:
This allows to use other likelihoods than the Gaussian one. Adds a new argument `inv_link` to allow the user to specify the proper inverse link function, defaults to identity.  This also introduce an `scale` argument that allows to control the proposal of the leaf values. For the Gaussian likelihoods (and probably other unbounded likelihoods like the Student T) the default scale is ok, for other model like the binomial it may need to be adjusted for better results. It seems LOO can be use for this purpose. 

A simple example

```python
from scipy.special import expit

with pm.Model() as model:
    μ = pm.BART('μ', X, Y, m=50, scale=0.25, inv_link=expit)
    y = pm.Bernoulli('y', μ, observed=Y)
    trace = pm.sample(1000, return_inferencedata=True)
```

test_01b

### DIFF
--- a/pymc3/distributions/bart.py
+++ b/pymc3/distributions/bart.py
@@ -24,9 +24,20 @@ __all__ = ["BART"]
 
 class BaseBART(NoDistribution):
     def __init__(
-        self, X, Y, m=200, alpha=0.25, split_prior=None, scale=None, inv_link=None, *args, **kwargs
+        self,
+        X,
+        Y,
+        m=200,
+        alpha=0.25,
+        split_prior=None,
+        scale=None,
+        inv_link=None,
+        jitter=False,
+        *args,
+        **kwargs,
     ):
 
+        self.jitter = jitter
         self.X, self.Y, self.missing_data = self.preprocess_XY(X, Y)
 
         super().__init__(shape=X.shape[0], dtype="float64", testval=self.Y.mean(), *args, **kwargs)
@@ -79,7 +90,8 @@ class BaseBART(NoDistribution):
         if isinstance(X, (Series, DataFrame)):
             X = X.to_numpy()
         missing_data = np.any(np.isnan(X))
-        X = np.random.normal(X, np.nanstd(X, 0) / 100000)
+        if self.jitter:
+            X = np.random.normal(X, np.nanstd(X, 0) / 100000)
         Y = Y.astype(float)
         return X, Y, missing_data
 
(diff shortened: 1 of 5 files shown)


junpenglao:
Add a small test?

junpenglao (approved):
LGTM

pymc3/distributions/bart.py
```diff
@@ -257,6 +278,24 @@ def rvs(self):
                     return i
 
 
+class NormalSampler:
+    def __init__(self, sigma):
+        self.size = 5000
```
canyon289, line 283: Why 5000?

pymc3/step_methods/pgbart.py
```diff
@@ -86,12 +86,13 @@ def __init__(self, vars=None, num_particles=10, max_stages=5000, chunk="auto", m
...
     def astep(self, _):
         bart = self.bart
+        inv_link = bart.inv_link
         num_observations = bart.num_observations
         variable_inclusion = np.zeros(bart.num_variates, dtype="int")
 
         # For the tunning phase we restrict max_stages to a low number, otherwise it is almost sure
         # we will reach max_stages given that our first set of m trees is not good at all.
-        # Can set max_stages as a function of the number of variables/dimensions?
+        # Can set max_stages as a function of the number of variables/dimensions? XXX
```
canyon289, line 95: Whats XXX?

pymc3/tests/test_bart.py
```diff
@@ -0,0 +1,113 @@
...
+    idx_data_points = np.array(range(bart.num_observations), dtype="int32")
+    available_splitting_rules = bart.get_available_splitting_rules(
+        idx_data_points, idx_split_variable
+    )
+    assert available_splitting_rules.size == 1
+    np.testing.assert_almost_equal(available_splitting_rules, np.array([-3.3]), 1)
+
+
+def test_model():
+    np.random.seed(212480)
```
canyon289, line 94: Im not sure about this but should random seeds be generated with numpys rng thing?

canyon289:
Small comments, otherwise im good

canyon289 (approved):

pymc3/step_methods/pgbart.py
aloctavodia: A note to myself to check this later, I introduced the restriction during the early development. Now it seems it is not necessary anymore , but I did not have the time to tested thoroughly.

pymc3/distributions/bart.py
aloctavodia: larger cache size does not seem to provide any speed up.

pymc3/tests/test_bart.py
aloctavodia: Also not sue, i think this is something we do in many of our tests, so I will defer until we decide on an overall action.

PR #4677, opened 2021-05-05
glm module deprecation warning

### LINKED ISSUE #4680: Deprecate GLM submodule in v3
We decided to remove the glm submodule in `v4` in favor of `bambi`, so we need to raise a deprecation warning in `v3` redirecting users to `bambi`.

aloctavodia:
As discussed this adds a deprecation warning to the glm module in favor of Bambi. This closes #4680

### DIFF
--- a/pymc3/glm/linear.py
+++ b/pymc3/glm/linear.py
@@ -11,6 +11,7 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+import logging
 
 import numpy as np
 import theano.tensor as tt
@@ -20,6 +21,7 @@ from pymc3.glm import families
 from pymc3.glm.utils import any_to_tensor_and_labels
 from pymc3.model import Deterministic, Model
 
+_log = logging.getLogger("pymc3")
 __all__ = ["LinearComponent", "GLM"]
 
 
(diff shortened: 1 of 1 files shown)


ricardoV94:
Should we announce deprecation will happen forcibly on V4? That way we don't have to bother refactoring it at all.

pymc3/glm/linear.py
```diff
@@ -216,6 +218,10 @@ def from_formula(
             See `patsy.dmatrix` and `patsy.EvalEnvironment` for details.
         Other arguments are documented in the constructor.
         """
+        _log.warning(
+            "The glm module will be deprecated soon.\nWe recommend to instead use Bambi "
+            + "https://bambinos.github.io/bambi/"
+        )
```
OriolAbril: Should't this be a `FutureWarning`?

OriolAbril (commented):
Adding to Ricardo's point, if GLM needs any change at all to continue working with v4 we should probably remove it directly instead of doing some minor changes that will only be around between 4.0 and 4.1 so we can dedicate our efforts to actual v4 and bambi

pymc3/glm/linear.py
aloctavodia: I think this is ok.

aloctavodia:
If this is going to need refactoring to work under v4 I agree we should remove it for v4 and notify it accordingly.

ricardoV94 (approved):

Issue #4683, opened 2021-05-11
Replace custom RandomVariable for WeibullBetaRV


ricardoV94:
Reminder issue to remove the RandomVariable for WeibullBetaRV (introduced in #4640) once we have the automatic multiplication logp working (e.g., #4653)

brandonwillard:
We might be able to use `aesara.compile.builders.OpFromGraph` in the interim.  `OpFromGraph` will create a new `Op` to which we can assign a `_logp` specialization.  This removes the need to create a `RandomVariable` `Op` for cases in which operations that are already covered by existing `Op`s are applied to existing `RandomVariable`s.

ricardoV94:
Same for the Wald which is just the numpy offering + optional shifting. #4711

Issue #4511, opened 2021-03-08
Mixture should not allow mixing of discrete and continuous distributions


ricardoV94:
Right now the Mixture class allows to mix distributions of any kind, leading to improper logp evaluations where probability densities are mixed with probability masses.

```python
norm = pm.Normal.dist(0, 1)
pois = pm.Poisson.dist(1)
mix = pm.Mixture.dist([0.5, 0.5], comp_dists=[norm, pois], shape=1)

# Observed integer could only have come from Poisson component, so logp should be:
np.log(0.5 * np.exp(pois.logp(0))).eval()  # -1.69

# But it's higher:
mix.logp(0).eval()  # -0.958

# Observed float could only have come from Normal component, so logp should be:
np.log(0.5 * np.exp(norm.logp(1/3))).eval() # -1.66

# But it's higher:
mix.logp(1/3).eval()  # -0.93

# This is not a problem when evaluating points where the domains do not overlap:
mix.logp(-1).eval()  # -2.11
np.log(0.5 * np.exp(norm.logp(-1))).eval()  # -2.11
```

Questions related to the mixture of continuous and discrete distributions crop up now and then on the Discourse:
* https://discourse.pymc.io/t/sampling-from-a-learned-mixture-of-zeros-and-lognormal/3671
* https://discourse.pymc.io/t/zero-inflated-student-t/6170
* https://discourse.pymc.io/t/zero-inflated-normal/6857
* https://discourse.pymc.io/t/mixture-of-continuous-and-discrete-logp/2392

The STAN forum has an informative discussion on this:
* https://discourse.mc-stan.org/t/zero-inflated-models-in-stan/19496

***

It seems to me like this is an area where we could easily nudge users in the right direction, by raising an informative value error.

ricardoV94:
https://github.com/pymc-devs/pymc3/issues/3582 raises the same concern (applied to the proposal distributions)

brandonwillard:
This won't be worth addressing in `v3`, because `v4` will dramatically change the implementation of mixtures, and mixed datatype issues like this can be addressed much more naturally and directly in Aesara.

larryshamalama:
We can perhaps substitute `all_discrete` (which should be removed) by some other similar check in `dist`. I can create a draft PR on this

ricardoV94:
> We can perhaps substitute `all_discrete` (which should be removed) by some other similar check in `dist`. I can create a draft PR on this

[...]

PR #2246, opened 2017-05-31
WIP: Convert zero-inflated distributions to Mixture subclasses

fonnesbeck:
Replaces #2231, which I messed up.

PR #1459, opened 2016-10-18
WIP Implement ZeroInflatedPoisson as a subclass of Mixture

AustinRochford:
Putting this up for feedback, as it is not quite as straightforward as I would have liked.

Due to #1449 and its fix #1452, we can't use `ConstantDist` for the zero component, since it no longer broadcasts the way we need it to in this instance.

More specifically, since #1452, if any of the elements of `pm.ConstantDist.dist(0).logp(value)` are non-zero, the result will be `-np.inf`.  The `Zero` hack included below restores the pre-#1452 broadcasting behavior by bypassing `bound` and using `tt.switch` directly.  Using this hack, `pm.Zero.dist().logp(value)` will be an array whose entries are zero when the corresponding entry in `value` is zero and `-np.inf` when the corresponding entry in `value` is non-zero.  The use of `logsumexp` in `Mixture` will gracefully handle these `-np.inf`s, as desired.

My question for other is if we want to use this `Zero` hack to implement `ZeroInflated`\* distributions as subclasses of `Mixture` or leave them as-is.  Obviously we do not need to export `Zero` to the global namespace.

If we decide to go forward with this approach, I will port `ZeroInflatedNegativeBinomial` as well and fix any test that this might break.

PR #4684, opened 2021-05-11
Refactor missing discrete dists

ricardoV94:
This PR refactors all missing univariate discrete distributions.

1. I decided to refactor the `Constant` distribution after some debate on Slack as to whether this should be deprecated or not. Its direct use might be pretty problematic without a custom step algorithm, but could still be useful indirectly, such as in mixture distributions. I made the dtype of the RV be `floatX` to accept non-integer outputs (even though this distribution should be treated as a pmf, see https://github.com/pymc-devs/pymc3/issues/4511 for why this might matter)

2. I created separate RandomOps for the ZeroInflated* distributions for now. Similarly to #4683, this should only be done temporarily until we have Mixture Distributions working on V4 (it seems such approach was already attempted in the past #2246, #1459)
 
***


pymc3/distributions/discrete.py
```diff
@@ -1177,6 +1161,22 @@ def logp(value, p):
...
+    name = "constant"
+    ndim_supp = 0
+    ndims_params = [0]
+    dtype = "floatX"  # Should be treated as a discrete variable!
+    _print_name = ("Constant", "\\operatorname{Constant}")
+
+    @classmethod
+    def rng_fn(cls, rng, c, size=None):
+        ones = np.ones_like(c)
+        return rng.randint(ones, ones + 1, size=size) * c
```
ricardoV94: There must be a less hacking way to get a filled array given `c` and `size`, that works as other random generators would. `np.full(size, c)` doesn't seem to cut it, because it works with shape and not with size.

twiecki: Isn't shape and size the same thing here?

twiecki (dismissed):
This looks great!

twiecki:
Thanks for taking this on @ricardoV94. Separately, when I tried V4 it seems like other continuous distributions are also missing, like `StudentT`?

pymc3/distributions/discrete.py
ricardoV94: No because `c` could be a vector and `size` could be None for instance

ricardoV94:
Yes they are missing.

Only distributions that have a `rv_op` at the top have been refactored. StudentT hasn't yet.

I won't have more time this week, so if someone wants to open a PR for the missing continuous distributions it would be great. The logic should be pretty similar to what is done in this PR with the custom RandomOps

pymc3/distributions/discrete.py
```diff
@@ -1335,18 +1302,22 @@ def logp(self, value):
         -------
         TensorVariable
         """
-        psi = self.psi
-        theta = self.theta
 
         logp_val = at.switch(
             at.gt(value, 0),
-            at.log(psi) + self.pois.logp(value),
+            at.log(psi) + Poisson.logp(value, theta),
```
brandonwillard: Those `logp` functions inside the `Distribution` classes are just a convenience for declaring dispatch functions, so we don't want to rely on them in any way.  Instead, call the underlying `_logp` directly.

pymc3/distributions/discrete.py
```diff
@@ -1361,16 +1332,31 @@ def logcdf(self, value):
         -------
         TensorVariable
         """
-        psi = self.psi
 
         return bound(
-            logaddexp(at.log1p(-psi), at.log(psi) + self.pois.logcdf(value)),
+            logaddexp(at.log1p(-psi), at.log(psi) + Poisson.logcdf(value, theta)),
```
brandonwillard: See my comment above about using `_logp` instead.

pymc3/distributions/discrete.py
```diff
@@ -1467,19 +1432,24 @@ def logp(self, value):
...
         TensorVariable
         """
-        psi = self.psi
-        p = self.p
-        n = self.n
 
         logp_val = at.switch(
             at.gt(value, 0),
-            at.log(psi) + self.bin.logp(value),
+            at.log(psi) + Binomial.logp(value, n, p),
```
brandonwillard: See my comment above about using `_logp` instead.

pymc3/distributions/discrete.py
```diff
@@ -1493,22 +1463,39 @@ def logcdf(self, value):
...
         if np.ndim(value):
             raise TypeError(
                 f"ZeroInflatedBinomial.logcdf expects a scalar value but received a {np.ndim(value)}-dimensional object."
             )
 
-        psi = self.psi
-
         return bound(
-            logaddexp(at.log1p(-psi), at.log(psi) + self.bin.logcdf(value)),
+            logaddexp(at.log1p(-psi), at.log(psi) + Binomial.logcdf(value, n, p)),
```
brandonwillard: See my comment above about using `_logp` instead.

pymc3/distributions/discrete.py
```diff
@@ -1635,20 +1589,22 @@ def logp(self, value):
...
-        mu = self.mu
-        psi = self.psi
 
-        logp_other = at.log(psi) + self.nb.logp(value)
-        logp_0 = logaddexp(
-            at.log1p(-psi), at.log(psi) + alpha * (at.log(alpha) - at.log(alpha + mu))
+        return bound(
+            at.switch(
+                at.gt(value, 0),
+                at.log(psi) + NegativeBinomial.logp(value, n, p),
```
brandonwillard: See my comment above about using `_logp` instead.

pymc3/distributions/discrete.py
```diff
@@ -1667,13 +1623,14 @@ def logcdf(self, value):
             raise TypeError(
                 f"ZeroInflatedNegativeBinomial.logcdf expects a scalar value but received a {np.ndim(value)}-dimensional object."
             )
-        psi = self.psi
 
         return bound(
-            logaddexp(at.log1p(-psi), at.log(psi) + self.nb.logcdf(value)),
+            logaddexp(at.log1p(-psi), at.log(psi) + NegativeBinomial.logcdf(value, n, p)),
```
brandonwillard: See my comment above about using `_logp` instead.

brandonwillard (changes requested):
Looks good, but we really shouldn't rely on `Distribution.logp` functions, since those aren't officially part of the `Distribution` API&mdash;they're just a means of conveniently specifying dispatches.

pymc3/distributions/discrete.py
ricardoV94: Okay, actually I think the only problem was `size=None`. I added a check for that condition...

ricardoV94:
> Looks good, but we really shouldn't rely on `Distribution.logp` functions, since those aren't officially part of the `Distribution` API—they're just a means of conveniently specifying dispatches.

@brandonwillard I am not sure what you have in mind. This looks wrong, right?

```python
# at.log(psi) + Poisson.logp(value, theta)
at.log(psi) + _logp(poisson, value, {value: value}, theta)
```

I also tried this, but it fails following a `change_rv_size` operation:

```python
at.log(psi) + logpt(Poisson.dist(theta), value)
```

brandonwillard:
> @brandonwillard I am not sure what you have in mind. This looks wrong, right?
> 
> ```python
> # at.log(psi) + Poisson.logp(value, theta),
> at.log(psi) + _logp(poisson, value, {value: value}, theta),
> ```

No, that looks like the correct approach.

ricardoV94:
Could you give some intuition why this is better than the original approach?

brandonwillard:
> Could you give some intuition why this is better than the original approach?

As I stated in my original review comments, these `Distribution.logp` functions are not a part of the `Distribution` API and they are neither designed nor intended to be used in this way.  Their availability is questionable and subject to arbitrary change, so we don't want to use them outside of their intended purpose.

More importantly, we do *not* want to rely on the `Distribution` types in general, and especially not for log-likelihoods; they're currently just a legacy transition artifact.  This is clearly demonstrated by the lack of `Distribution` in the log-likelihood code itself.  

**Simply put, this approach unnecessarily adds `Distribution`s into the log-likelihood framework.**

If this sort of use case needs to be simplified, it can be done in better ways that are based only on the underlying dispatch.

brandonwillard (approved):

twiecki:
🥳

Issue #4686, opened 2021-05-12
Port remaining distributions to v4


twiecki:
There are still various continuous distributions missing for `v4` that need to get ported.
Here is an example of how to do this port for where `aesara` does not provide a `RandomVariable` (which is all of the currently missing ones): https://github.com/pymc-devs/pymc3/blob/v4/pymc3/distributions/continuous.py#L2648

Here is a list of the missing continuous ones:
- [x] `Flat` #4723 
- [x] `HalfFlat` #4723
- [x] `TruncatedNormal` #4711
- [x] `Wald` #4711
- [x] `Kumaraswamy`  #4706
- [x] `Laplace` #4691
- [x] `AsymmetricLaplace` #4746
- [x] `StudentT` #4694
- [x] `Pareto` https://github.com/pymc-devs/pymc3/pull/4691
- [x] `ChiSquared` #4695 
- [x] `HalfStudentT` #4746
- [x] `ExGaussian`  #4746
- [x] `SkewNormal` #4705
- [x] `Rice` #4705
- [x] `LogitNormal`  #4703
- [x] `Interpolated` #4746
- [x] `Moyal`  #4704
- [x] `DensityDist` #5026 

And the missing multivariate ones:
- [x] `MvStudentT`  #4731 
- [x] `DirichletMultinomial` #4758
- [x] `Wishart` https://github.com/pymc-devs/pymc3/pull/4777
- [x] `LKJCorr` #5382 
- [x] `LKJCholeskyCov` #5382
- [x] `MatrixNormal` #4777 
- [x] `KroneckerNormal` #4774
- [x] `CAR` #4596

Would be very important to add these to get an alpha out.

See @ricardoV94's comment below for a step-by-step guide on how to port a distribution.

ricardoV94:
The refactored distributions have this line at the beginning of the class `rv_op = x` and generally don't have a random method and the `logp/logcdf` start with `value` instead of `self`. That should be enough to identify them :p

twiecki:
@ricardoV94 Thanks, edited the above. If you have time, perhaps you could write a short list of steps of what to do to port them. (Unless you want to take these on yourself :)).

ricardoV94:
Also missing the following multivariate distributions:
* `MvStudentT`
* `DirichletMultinomial`
* `Wishart`
* `LKJCorr`
* `MatrixNormal`
* `KroneckerNormal`
[...]

PR #4694, opened 2021-05-14
Refactor Student T Distribution

themrzmaster:
This PR closes subset of #4686.

I am not sure if the parametrization is correct

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -1811,6 +1811,19 @@ class Lognormal(PositiveContinuous):
         )
 
 
+class StudentTRV(RandomVariable):
+    name = "studentt"
+    ndim_supp = 0
+    ndims_params = [0, 0]
+    dtype = "floatX"
+    _print_name = ("StudentT", "\\operatorname{StudentT}")
+
+    @classmethod
+    def rng_fn(cls, rng, nu, mu, lam, size=None):
+        return stats.t.rvs(nu, mu, lam ** -0.5, size=size, random_state=rng)
+
+studentt = StudentTRV()
+
 class StudentT(Continuous):
     r"""
     Student's T log-likelihood.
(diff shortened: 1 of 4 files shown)


pymc3/distributions/continuous.py
```diff
@@ -1811,6 +1811,19 @@ def logcdf(value, mu, sigma):
         )
 
 
+class StudentTRV(RandomVariable):
+    name = "studentt"
+    ndim_supp = 0
+    ndims_params = [0, 0]
```
ricardoV94: It should be 3 zeros here (one for each argument that the tstudent takes)

twiecki:
Thanks for taking this on @themrzmaster, we'll try to get some eyes on this soon.

pymc3/distributions/continuous.py
```diff
@@ -1874,45 +1887,28 @@ class StudentT(Continuous):
...
         if sd is not None:
             sigma = sd
-        self.nu = nu = at.as_tensor_variable(floatX(nu))
+        nu = at.as_tensor_variable(floatX(nu))
         lam, sigma = get_tau_sigma(tau=lam, sigma=sigma)
-        self.lam = lam = at.as_tensor_variable(lam)
-        self.sigma = self.sd = sigma = at.as_tensor_variable(sigma)
-        self.mean = self.median = self.mode = self.mu = mu = at.as_tensor_variable(mu)
+        lam = at.as_tensor_variable(lam)
+        sd = sigma = at.as_tensor_variable(sigma)
```
ricardoV94: We don't need to convert `sd` and `sigma` to tensor as it will not be used after this

pymc3/distributions/continuous.py
```diff
@@ -1874,45 +1887,28 @@ class StudentT(Continuous):
...
+        median = mode = mu = at.as_tensor_variable(mu)
```
ricardoV94: You can remove the `median = ...` line for now

pymc3/distributions/continuous.py
```diff
@@ -1926,11 +1922,7 @@ def logp(self, value):
         -------
         TensorVariable
         """
-        nu = self.nu
-        mu = self.mu
-        lam = self.lam
-        sigma = self.sigma
-
+        lam, sigma = get_tau_sigma(tau=lam, sigma=None)
```
ricardoV94, line 1925: We don't need to get the `sigma` again just to check if it's positive in the bound. The `lam` should have the same sign as `sigma`.

pymc3/distributions/continuous.py
```diff
@@ -1944,7 +1936,7 @@ def logp(self, value):
     def _distr_parameters_for_repr(self):
```
ricardoV94: This method can be removed

pymc3/tests/test_distributions_random.py
```diff
@@ -461,6 +455,18 @@ class TestGumbel(BaseTestDistribution):
...
 
+class TestStudentT(BaseTestDistribution):
+    pymc_dist = pm.StudentT
+    pymc_dist_params = {"nu": 5.0, "mu": 0.0, "lam": 1.0}
+    expected_rv_op_params = {"nu": 5.0, "mu": 0.0, "lam": 1.0}
+    reference_dist_params = {"df": 5.0, "loc": 0.0, "scale": 1.0}
+    reference_dist = seeded_scipy_distribution_builder("t")
+    tests_to_run = [
+        "check_pymc_params_match_rv_op",
+        "check_pymc_draws_match_reference"
```
ricardoV94: Can you add a "check_rv_size"?

ricardoV94 (changes requested):
Thanks for opening this PR!
I left some comments. Let me know if you have any questions :)

themrzmaster:
Thanks @ricardoV94 ! I'll probably get more refactors on the weekend.

pymc3/tests/test_distributions_random.py
```diff
@@ -461,9 +461,11 @@ class TestStudentT(BaseTestDistribution):
     expected_rv_op_params = {"nu": 5.0, "mu": 0.0, "lam": 1.0}
     reference_dist_params = {"df": 5.0, "loc": 0.0, "scale": 1.0}
     reference_dist = seeded_scipy_distribution_builder("t")
+    size = 15
```
ricardoV94: No need to specify `size=15`, it's already the default

ricardoV94:
Also it seems this test can be re-enabled: https://github.com/pymc-devs/pymc3/blob/1118940dd5238a724be167fcc58295c49e0f10aa/pymc3/tests/test_posteriors.py#L79-L87

This one seems to wrongly mention the StudentT, but I don't think it actually depends on the LKJCholeskyCov, not the StudentT:
https://github.com/pymc-devs/pymc3/blob/1118940dd5238a724be167fcc58295c49e0f10aa/pymc3/tests/test_posteriors.py#L101-L107

ricardoV94:
I think that's it. Let's see if the tests pass :crossed_fingers:

ricardoV94 (dismissed):

ricardoV94:
@themrzmaster I went ahead and changed the default parametrization of the op to be in terms of `sigma` instead of `lam` so that we can call `scipy` directly under the hood. 

Let me know if it looks good for you and if there's anything else you want to change!

If you want to make changes, don't forget to first pull from here.

themrzmaster:
LGTM! Thanks

ricardoV94 (approved):

ricardoV94:
@themrzmaster Great work! Looking forward to your next PR :rocket:

PR #4615, opened 2021-04-04
Refactor DiscreteWeibull

ricardoV94:
Refactoring DiscreteWeibull as a test for how to refactor other distributions whose random ops are not implemented in aesara.

PR #4695, opened 2021-05-14
Refactoring the ChiSquared distribution

larryshamalama:
This PR further addresses the ongoing issue [4686](https://github.com/pymc-devs/pymc3/issues/4686)

I don't fully understand the process of creating tests, but I try to follow the previously mentioned [template](https://github.com/pymc-devs/pymc3/pull/4615/files). If my edits are good (please let me know if I need to correct anything), my questions are:

- How can I know when to use `seeded_numpy_distribution_builder` or `seeded_scipy_distribution_builder` vs. creating my own function?
- What tests should be run in the list `tests_to_run`?

If there are other PRs or documentation that addresses my questions, I am happy to be pointed to them.

Thanks!


twiecki:
Thanks for taking this on @larryshamalama!

ricardoV94:
@larryshamalama Thanks for taking on this one.

Unfortunately it seems the chisquare RandomVariable should go on our [aesara library](https://github.com/pymc-devs/aesara) , as it is offered by [numpy](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.chisquare.html#numpy.random.Generator.chisquare). Would you like to open a PR there? It would look similar to this one: https://github.com/pymc-devs/aesara/pull/363/files, but including just the `ChiSquareRV` that you created here.

Then we can come back to this PR and import the op from `aesara`. We might still need other changes related to the fact we are inheriting from the `Gamma`, but we can discuss that after.

larryshamalama:
Thanks so much for the suggestions @ricardoV94, I can do just that and then edit this PR accordingly.

larryshamalama:
Using aesara's op now. Still unsure if I am properly inheriting from `Gamma`

Edit: jobs seem to be failing...

pymc3/distributions/continuous.py
```diff
@@ -2548,7 +2549,7 @@ def logcdf(value, alpha, beta):
         )
 
 
-class ChiSquared(Gamma):
```
ricardoV94, line 2551: We don't have to change the name of the PyMC3 distribution (unless that's what we want of course).

pymc3/distributions/continuous.py
```diff
@@ -2586,9 +2587,11 @@ class ChiSquared(Gamma):
...
         Degrees of freedom (nu > 0).
     """
+    rv_op = chisquare
 
-    def __init__(self, nu, *args, **kwargs):
-        self.nu = nu = at.as_tensor_variable(floatX(nu))
+    @classmethod
+    def dist(cls, nu, *args, **kwargs):
+        nu = at.as_tensor_variable(floatX(nu))
         super().__init__(alpha=nu / 2.0, beta=0.5, *args, **kwargs)
```
ricardoV94: This is the tricky part. The new chisquare op only expects 1 input (`nu`), but then we can't simply use the gamma logp by inheritance as we were doing in V3. 

I think the easiest solution is to implement `logp` and `logcdf` methods that call those from Gamma with the `Gamma.logp` and `Gamma.logcdf` methods. You can find an example of this in https://github.com/pymc-devs/pymc3/blob/79245ce631550af54be398cea73e990f06a50047/pymc3/distributions/discrete.py#L723.

With this, this distribution shouldn't inherit from Gamma anymore and the return at the end of `dist()` should be like the other distributions and pass `[nu]` only.

ricardoV94:
Even after fixing the errors, the tests will still fail until a new version of Aesara is released. You should test that it passes locally for the time being

ricardoV94:
The latest Aesara version was released, you should be able to install it (version `2.0.9`) and make use of the chisquare function now.

ricardoV94:
Hi @larryshamalama, any progress on this PR?

larryshamalama:
Hi @ricardoV94, so sorry for the delay, this PR sat on my back burner due to organizing an upcoming conference. Let me revisit this today.

ricardoV94:
There is no rush, just wanted to know what was the status :)

pymc3/distributions/continuous.py
```diff
@@ -2586,11 +2587,46 @@ class ChiSquared(Gamma):
...
+        ----------
+        value: numeric or np.ndarray or `TensorVariable`
+            Value(s) for which log CDF is calculated. If the log CDF for
+            multiple values are desired the values must be provided in a numpy
+            array or `TensorVariable`.
+        Returns
+        -------
+        TensorVariable
+        """
+        return Gamma.logcdf(value, nu/2, 2)
```
larryshamalama: After looking much into the other functions used, is the solution really this simple? I'm having some doubts

pymc3/distributions/continuous.py
larryshamalama: I used `ChiSquare` since that's what I used in Aesara (r.f. [PR 414](https://github.com/pymc-devs/aesara/pull/414)), which was based off of the spelling in NumPy and SciPy. What do you think?

larryshamalama:
> Even after fixing the errors, the tests will still fail until a new version of Aesara is released. You should test that it passes locally for the time being

Embarrassing question... do you mind explaining how to check this? I'm new to testing in general and was not able to successfully run some test functions by copy pasting into a notebook (see [here](https://github.com/larryshamalama/pymc3-playground/blob/master/aesara-pr414.ipynb)). Do you have any suggestions or tips?

pymc3/distributions/continuous.py
```diff
@@ -2586,11 +2587,46 @@ class ChiSquared(Gamma):
...
+        ----------
+        value: numeric
+            Value(s) for which log-probability is calculated. If the log probabilities for multiple
+            values are desired the values must be provided in a numpy array or Aesara tensor
+
+        Returns
+        -------
+        TensorVariable
+        """
+        return Gamma.logpdf(value, nu/2, 2)
```
ricardoV94: This should be called logp

pymc3/tests/test_distributions_random.py
```diff
@@ -277,10 +277,17 @@ class TestAsymmetricLaplace(BaseTestCases.BaseTestCase):
     params = {"kappa": 1.0, "b": 1.0, "mu": 0.0}
 
 
-@pytest.mark.xfail(reason="This distribution has not been refactored for v4")
-class TestChiSquared(BaseTestCases.BaseTestCase):
-    distribution = pm.ChiSquared
+class TestChiSquare(BaseTestCases.BaseTestCase):
```
ricardoV94: This should inherit from the new `BaseTestDistribution`

ricardoV94:
> Embarrassing question... do you mind explaining how to check this? I'm new to testing in general and was not able to successfully run some test functions by copy pasting into a notebook

There is nothing embarrassing about it. First of all, the tests should now work here as we have already updated the aesara version we are using.

In any case it's very important to get comfortable running tests locally. Ideally you would run your tests from the terminal by calling `pytest path/to/test file.py -k testfunction/testclass` You should read the documentation [here](https://docs.pytest.org/en/6.2.x/getting-started.html) or search for some tutorials.

Alternatively you can use some IDEs like pycharm that allow you to run tests from the interface.

Copying tests to a separate notebook should only be a last resort thing (if ever advisable) because it's much more difficult to setup, and not easy to use debugging tools that you'll need when you can't easily figure out why a test is failing.

ricardoV94:
You should also install `pre-commit` locally if you haven't yet. One of the failing tests is just the code style check because you didn't include a space between the division operator.

You can read our wiki to learn how to install and run it: https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style

This way it will catch this sort of errors automatically (and fix them if it can) when you try to make a commit, saving you some time.

pymc3/distributions/continuous.py
ricardoV94: I don't have a strong preference but if we change the name we should include the old class distribution just to raise a deprecation warning saying that the name has changed.

larryshamalama:
so I spent some time looking into testing and the library... My newly created tests are failing but I have no idea how to fix them. Checking `TestChiSquare` using `pytest test_distributions_random.py -k TestChiSquare` gives me to following error:

`============================================================================== short test summary info ==============================================================================
FAILED test_distributions_random.py::TestChiSquare::test_distribution - TypeError: object.__init__() takes exactly one argument (the instance to initialize)
`

I checked my ChiSquare class in `continuous.py`, but I feel a bit stuck. Do you have any thoughts or pointers that you want to share?

pymc3/distributions/continuous.py
```diff
@@ -2548,7 +2549,7 @@ def logcdf(value, alpha, beta):
         )
 
 
-class ChiSquared(Gamma):
+class ChiSquare(Gamma):
```
ricardoV94: Shouldn't inherit from Gamma anymore. Should be just `class ChiSquare:`

pymc3/distributions/continuous.py
```diff
@@ -2586,10 +2587,45 @@ class ChiSquared(Gamma):
...
     """
+    rv_op = chisquare
 
-    def __init__(self, nu, *args, **kwargs):
-        self.nu = nu = at.as_tensor_variable(floatX(nu))
-        super().__init__(alpha=nu / 2.0, beta=0.5, *args, **kwargs)
+    @classmethod
+    def dist(cls, nu, *args, **kwargs):
+        nu = at.as_tensor_variable(floatX(nu))
+        return super().__init__([nu / 2.0, 0.5], *args, **kwargs)
```
ricardoV94: The first argument should just be `[nu]`

pymc3/tests/test_distributions.py
```diff
@@ -1030,14 +1030,19 @@ def test_half_normal(self):
...
         )
 
-    @pytest.mark.xfail(reason="Distribution not refactored yet")
-    def test_chi_squared(self):
+    def test_chisquared(self):
         self.check_logp(
             ChiSquared,
             Rplus,
-            {"nu": Rplusdunif},
+            {"nu": Rplus},
```
ricardoV94, line 1037: I see now, I think the previous domain was more appropriate as it was positive + discrete.

pymc3/distributions/continuous.py
```diff
@@ -2586,10 +2587,45 @@ class ChiSquared(Gamma):
     nu: int
```
ricardoV94: the old documentation is wrong, nu can be a float

larryshamalama: fixed! thanks for pointing this out

ricardoV94 (approved):
Looks good to me. Let's see if the tests pass :)

twiecki (approved):

twiecki:
Thanks @larryshamalama!

ricardoV94:
Was this merged into `v4`?

twiecki:
Damn, my bad. @ricardoV94 can you fix that?

ricardoV94:
I don't think I would have realized it in time either. How can it be fixed, other than opening a new PR with the same content?

larryshamalama:
Oh no, was this supposed to be merged into another branch?

PR #4693, opened 2021-05-13
Revert breaking `size` changes

brandonwillard:
This PR reverts the `size`-related breaking changes and `Model`/`Distribution` design issues introduced by #4625.

Closes #4662.

Issue #4656, opened 2021-04-21
Allow naming of dims when parametrizing with size


michaelosthege:
## Description
The current implementation allows users to register/name/introduce dimensions to the model by specifying `dims` on RVs that have implied dimensions:

```python
with pm.Model():
    pm.Normal("n", mu=[1,2,3], dims=("town",))
```

But the `shape`/`dims`/`size` kwargs are currently implemented as exclusive, so the following scenario is currently not supported:
```python
with pm.Model():
    pm.Normal("n", mu=[1,2,3], size=7, dims=("day_of_week", "town"))
```

There's already a test for this scenario: `test_shape_handling.py::TestShapeDimsSize::test_data_defined_size_dimension_can_register_dimname`

Technically there should be nothing preventing us from supporting this use case. We'll just need to require the newly introduced dim name to be previously unknown, so we can rule out potential shape conflicts.

ricardoV94:
@michaelosthege Is this one done?

michaelosthege:
No, the corresponding test is still marked as XFAIL on `main`:

https://github.com/pymc-devs/pymc3/blob/0923d255a45773b7a92bbe36cb11ad5d946a3122/pymc3/tests/test_shape_handling.py#L315-L324

On a high level this is what needs to be changed:
* Permit simultaneously passing `dims and shape` or `dims and size`
* Check that the dimensionality of the created `rv_out` matches all expectations
* Think about ways how this could break & add corresponding warnings/tests:
  E.g. if elements in `shape`/`size` don't match the (symbolic) dimension lengths that are already registered with the model.
  In these cases we probably want to substitute the entries in `size`/`shape` such that `dims` takes priority over `size`/`shape`. (Ignoring the numeric entries that would restrict model flexibility.)

If nobody else picks this up I'll do it, but there are other important things on the ToDo-list too: Getting the `gp` submodule back online, for example.

ricardoV94:
I actually closed it by pure accident! Sorry for that

LukeLB:
Hi guys. I've been looking at trying to put together a PR for this issue. All I've done so far is remove the if statement in the Distrubution parent class that disallows dual passing of size/shape and dims (lines 216-223 in distrubution.py), then written a new test using some of the model spec code from #5181, this test passes. However, as @michaelosthege mentions above there is already a test for this case,
```
    @pytest.mark.xfail(reason="Simultaneous use of size and dims is not implemented")
[...]

Issue #4650, opened 2021-04-19
Add type annotations to `convert_observed_data`


michaelosthege:
We use this function to typecast a zoo of different objects. When using it, it is completely unclear what you'll get back..

ricardoV94:
The name should also be changed as it is no longer pandas specific (eg it has special functionality for masked arrays)

MarcoGorelli:
https://github.com/pymc-devs/pymc3/blob/7e35cddd7e54e6e6f51fc1aa3ed3df76016aa00a/pymc3/aesaraf.py#L89-L90

What inputs could end up there, other than pandas objects?

michaelosthege:
Someone could implement their own type with a `to_numpy` function I guess? At least there are no checks that prevent that from happening.

Probably more of a theoretical possibility though, but `hasattr` leaves that door open.

MarcoGorelli:
True, but then you're assuming that their implementation of `isnull` returns either a numpy array or that it returns an object with `to_numpy`

FWIW I think just checking
```
if isinstance(data, (pd.Series, pd.DataFrame)):
    ...
elif isinstance(data, pd.Index):
    ...
```
could lead to fewer surprises

michaelosthege:
If you promise that pandas won't remove the `to_numpy` and `isnull` methods in the future that's fine ;)

And it could help `mypy` so there's that.

michaelosthege:
The function is being renamed by #5743.

fonnesbeck:
It appears that `convert_data` (the new name) now has a return type.

michaelosthege:
closed by https://github.com/pymc-devs/pymc/pull/7299 and https://github.com/pymc-devs/pymc/pull/7334

PR #4696, opened 2021-05-14
Implement backwards-compatible shape and dims with Ellipsis support

twiecki:
Here is what I did:
Create a diff between #4693 and #4667 and stage that diff onto a branch off of #4693 (no conflicts!). I then ran `git commit -p` and only selected those that seemed necessary here. Mostly there were some changes to tests that changed `size` to `shape` I did not commit.

Remember: Before merging still need to change author to @michaelosthege.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -12,8 +12,9 @@
 - The `CAR` distribution has been added to allow for use of conditional autoregressions which often are used in spatial and network models.
 - The dimensionality of model variables can now be parametrized through either of `shape`, `dims` or `size` (see [#4696](https://github.com/pymc-devs/pymc3/pull/4696)):
   - With `shape` the length of dimensions must be given numerically or as scalar Aesara `Variables`. Numeric entries in `shape` restrict the model variable to the exact length and re-sizing is no longer possible.
-  - `dims` keeps model variables re-sizeable (for example through `pm.Data`) and leads to well defined coordinates in `InferenceData` objects. An `Ellipsis` (`...`) in the last position of `dims` can be used as short-hand notation for implied dimensions.
+  - `dims` keeps model variables re-sizeable (for example through `pm.Data`) and leads to well defined coordinates in `InferenceData` objects.
   - The `size` kwarg behaves like it does in Aesara/NumPy. For univariate RVs it is the same as `shape`, but for multivariate RVs it depends on how the RV implements broadcasting to dimensionality greater than `RVOp.ndim_supp`.
+  - An `Ellipsis` (`...`) in the last position of `shape` or `dims` can be used as short-hand notation for implied dimensions.
 - Add `logcdf` method to Kumaraswamy distribution (see [#4706](https://github.com/pymc-devs/pymc3/pull/4706)).
 - ...
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -10,6 +10,10 @@
 
 ### New Features
 - The `CAR` distribution has been added to allow for use of conditional autoregressions which often are used in spatial and network models.
+- The dimensionality of model variables can now be parametrized through either of `shape`, `dims` or `size` (see [#4696](https://github.com/pymc-devs/pymc3/pull/4696)):
+  - With `shape` the length of dimensions must be given numerically or as scalar Aesara `Variables`. Numeric entries in `shape` restrict the model variable to the exact length and re-sizing is no longer possible.
+  - `dims` keeps model variables re-sizeable (for example through `pm.Data`) and leads to well defined coordinates in `InferenceData` objects. An `Ellipsis` (`...`) in the last position of `dims` can be used as short-hand notation for implied dimensions.
+  - The `size` kwarg behaves like it does in Aesara/NumPy. For univariate RVs it is the same as `shape`, but for multivariate RVs it depends on how the RV implements broadcasting to dimensionality greater than `RVOp.ndim_supp`.
 - Add `logcdf` method to Kumaraswamy distribution (see [#4706](https://github.com/pymc-devs/pymc3/pull/4706)).
 - ...
 
--- a/pymc3/aesaraf.py
+++ b/pymc3/aesaraf.py
@@ -46,10 +46,12 @@ from aesara.graph.op import Op, compute_test_value
 from aesara.sandbox.rng_mrg import MRG_RandomStream as RandomStream
 from aesara.tensor.elemwise import Elemwise
 from aesara.tensor.random.op import RandomVariable
+from aesara.tensor.shape import SpecifyShape
 from aesara.tensor.sharedvar import SharedVariable
 from aesara.tensor.subtensor import AdvancedIncSubtensor, AdvancedIncSubtensor1
 from aesara.tensor.var import TensorVariable
 
+from pymc3.exceptions import ShapeError
 from pymc3.vartypes import continuous_types, int_types, isgenerator, typefilter
 
 PotentialShapeType = Union[
@@ -153,6 +155,16 @@ def change_rv_size(
         Expand the existing size by `new_size`.
 
     """
+    # Check the dimensionality of the `new_size` kwarg
+    new_size_ndim = np.ndim(new_size)
+    if new_size_ndim > 1:
+        raise ShapeError("The `new_size` must be ≤1-dimensional.", actual=new_size_ndim)
+    elif new_size_ndim == 0:
+        new_size = (new_size,)
+
+    # Extract the RV node that is to be resized, together with its inputs, name and tag
+    if isinstance(rv_var.owner.op, SpecifyShape):
+        rv_var = rv_var.owner.inputs[0]
     rv_node = rv_var.owner
     rng, size, dtype, *dist_params = rv_node.inputs
     name = rv_var.name
@@ -161,10 +173,10 @@ def change_rv_size(
     if expand:
         if rv_node.op.ndim_supp == 0 and at.get_vector_length(size) == 0:
             size = rv_node.op._infer_shape(size, dist_params)
-        new_size = tuple(at.atleast_1d(new_size)) + tuple(size)
+        new_size = tuple(new_size) + tuple(size)
 
-    # Make sure the new size is a tensor. This helps to not unnecessarily pick
-    # up a `Cast` in some cases
+    # Make sure the new size is a tensor. This dtype-aware conversion helps
+    # to not unnecessarily pick up a `Cast` in some cases (see #4652).
     new_size = at.as_tensor(new_size, ndim=1, dtype="int64")
 
     new_rv_node = rv_node.op.make_node(rng, new_size, dtype, *dist_params)
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -19,21 +19,29 @@ import types
 import warnings
 
 from abc import ABCMeta
-from typing import Optional, Sequence, Tuple, Union
+from typing import Optional
 
 import aesara
 import aesara.tensor as at
 import dill
-import numpy as np
 
-from aesara.graph.basic import Variable
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.random.var import RandomStateSharedVariable
-from aesara.tensor.var import TensorVariable
 
-from pymc3.aesaraf import change_rv_size, pandas_to_array
+from pymc3.aesaraf import change_rv_size
 from pymc3.distributions import _logcdf, _logp
-from pymc3.exceptions import ShapeError, ShapeWarning
+from pymc3.distributions.shape_utils import (
+    Dims,
+    Shape,
+    Size,
+    convert_dims,
+    convert_shape,
+    convert_size,
+    find_size,
+    maybe_resize,
+    resize_from_dims,
+    resize_from_observed,
+)
 from pymc3.util import UNSET, get_repr_for_variable
 from pymc3.vartypes import string_types
 
@@ -50,20 +58,6 @@ vectorized_ppc = contextvars.ContextVar(
 )  # type: contextvars.ContextVar[Optional[Callable]]
 
 PLATFORM = sys.platform
-
-# User-provided can be lazily specified as scalars
-Shape = Union[int, TensorVariable, Sequence[Union[int, TensorVariable, type(Ellipsis)]]]
-Dims = Union[str, Sequence[Union[str, None, type(Ellipsis)]]]
-Size = Union[int, TensorVariable, Sequence[Union[int, TensorVariable]]]
-
-# After conversion to vectors
-WeakShape = Union[TensorVariable, Tuple[Union[int, TensorVariable, type(Ellipsis)], ...]]
-WeakDims = Tuple[Union[str, None, type(Ellipsis)], ...]
-
-# After Ellipsis were substituted
-StrongShape = Union[TensorVariable, Tuple[Union[int, TensorVariable], ...]]
-StrongDims = Sequence[Union[str, None]]
-StrongSize = Union[TensorVariable, Tuple[Union[int, TensorVariable], ...]]
 
 
 class _Unpickling:
@@ -118,135 +112,6 @@ class DistributionMeta(ABCMeta):
             new_cls.register(rv_type)
 
         return new_cls
-
-
-def _convert_dims(dims: Dims) -> Optional[WeakDims]:
-    """ Process a user-provided dims variable into None or a valid dims tuple. """
-    if dims is None:
-        return None
-
-    if isinstance(dims, str):
-        dims = (dims,)
-    elif isinstance(dims, (list, tuple)):
-        dims = tuple(dims)
-    else:
-        raise ValueError(f"The `dims` parameter must be a tuple, str or list. Actual: {type(dims)}")
-
-    if any(d == Ellipsis for d in dims[:-1]):
-        raise ValueError(f"Ellipsis in `dims` may only appear in the last position. Actual: {dims}")
-
-    return dims
-
-
-def _convert_shape(shape: Shape) -> Optional[WeakShape]:
-    """ Process a user-provided shape variable into None or a valid shape object. """
-    if shape is None:
-        return None
-
-    if isinstance(shape, int) or (isinstance(shape, TensorVariable) and shape.ndim == 0):
-        shape = (shape,)
-    elif isinstance(shape, (list, tuple)):
-        shape = tuple(shape)
-    else:
-        raise ValueError(
-            f"The `shape` parameter must be a tuple, TensorVariable, int or list. Actual: {type(shape)}"
-        )
-
-    if isinstance(shape, tuple) and any(s == Ellipsis for s in shape[:-1]):
-        raise ValueError(
-            f"Ellipsis in `shape` may only appear in the last position. Actual: {shape}"
-        )
-
-    return shape
-
-
-def _convert_size(size: Size) -> Optional[StrongSize]:
-    """ Process a user-provided size variable into None or a valid size object. """
-    if size is None:
-        return None
-
-    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
-        size = (size,)
-    elif isinstance(size, (list, tuple)):
-        size = tuple(size)
-    else:
-        raise ValueError(
-            f"The `size` parameter must be a tuple, TensorVariable, int or list. Actual: {type(size)}"
-        )
-
-    if isinstance(size, tuple) and Ellipsis in size:
-        raise ValueError(f"The `size` parameter cannot contain an Ellipsis. Actual: {size}")
-
-    return size
-
-
-def _resize_from_dims(
-    dims: WeakDims, ndim_implied: int, model
-) -> Tuple[int, StrongSize, StrongDims]:
-    """Determines a potential resize shape from a `dims` tuple.
-
-    Parameters
-    ----------
-    dims : array-like
-        A vector of dimension names, None or Ellipsis.
-    ndim_implied : int
-        Number of RV dimensions that were implied from its inputs alone.
-    model : pm.Model
-        The current model on stack.
-
-    Returns
-    -------
-    ndim_resize : int
-        Number of dimensions that should be added through resizing.
-    resize_shape : array-like
-        The shape of the new dimensions.
-    """
-    if Ellipsis in dims:
-        # Auto-complete the dims tuple to the full length.
-        # We don't have a way to know the names of implied
-        # dimensions, so they will be `None`.
-        dims = (*dims[:-1], *[None] * ndim_implied)
-
-    ndim_resize = len(dims) - ndim_implied
-
-    # All resize dims must be known already (numerically or symbolically).
-    unknowndim_resize_dims = set(dims[:ndim_resize]) - set(model.dim_lengths)
-    if unknowndim_resize_dims:
-        raise KeyError(
-            f"Dimensions {unknowndim_resize_dims} are unknown to the model and cannot be used to specify a `size`."
-        )
-
-    # The numeric/symbolic resize tuple can be created using model.RV_dim_lengths
-    resize_shape = tuple(model.dim_lengths[dname] for dname in dims[:ndim_resize])
-    return ndim_resize, resize_shape, dims
-
-
-def _resize_from_observed(
-    observed, ndim_implied: int
-) -> Tuple[int, StrongSize, Union[np.ndarray, Variable]]:
-    """Determines a potential resize shape from observations.
-
-    Parameters
-    ----------
-    observed : scalar, array-like
-        The value of the `observed` kwarg to the RV creation.
-    ndim_implied : int
-        Number of RV dimensions that were implied from its inputs alone.
-
-    Returns
-    -------
-    ndim_resize : int
-        Number of dimensions that should be added through resizing.
-    resize_shape : array-like
-        The shape of the new dimensions.
-    observed : scalar, array-like
-        Observations as numpy array or `Variable`.
-    """
-    if not hasattr(observed, "shape"):
-        observed = pandas_to_array(observed)
-    ndim_resize = observed.ndim - ndim_implied
-    resize_shape = tuple(observed.shape[d] for d in range(ndim_resize))
-    return ndim_resize, resize_shape, observed
 
 
 class Distribution(metaclass=DistributionMeta):
@@ -335,7 +200,7 @@ class Distribution(metaclass=DistributionMeta):
             raise ValueError(
                 f"Passing both `dims` ({dims}) and `size` ({kwargs['size']}) is not supported!"
             )
-        dims = _convert_dims(dims)
+        dims = convert_dims(dims)
 
         # Create the RV without specifying testval, because the testval may have a shape
         # that only matches after replicating with a size implied by dims (see below).
@@ -346,9 +211,9 @@ class Distribution(metaclass=DistributionMeta):
         # `dims` are only available with this API, because `.dist()` can be used
         # without a modelcontext and dims are not tracked at the Aesara level.
         if dims is not None:
-            ndim_resize, resize_shape, dims = _resize_from_dims(dims, ndim_actual, model)
+            ndim_resize, resize_shape, dims = resize_from_dims(dims, ndim_actual, model)
         elif observed is not None:
-            ndim_resize, resize_shape, observed = _resize_from_observed(observed, ndim_actual)
+            ndim_resize, resize_shape, observed = resize_from_observed(observed, ndim_actual)
 
         if resize_shape:
             # A batch size was specified through `dims`, or implied by `observed`.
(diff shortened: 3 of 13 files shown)


pymc3/tests/test_distributions_random.py
```diff
@@ -188,10 +188,9 @@ def get_random_variable(self, shape, with_vector_params=False, name=None):
         @staticmethod
         def sample_random_variable(random_variable, size):
             """ Draws samples from a RandomVariable using its .random() method. """
-            if size is None:
-                return random_variable.eval()
-            else:
-                return change_rv_size(random_variable, size, expand=True).eval()
+            if size:
```
twiecki, line 191:
```suggestion
            if size is None:
```

twiecki, line 191: Can we make this explicit?

michaelosthege: With the new robustness in `change_rv_size` this is just fine.

RELEASE-NOTES.md
```diff
@@ -9,6 +9,10 @@
 
 ### New Features
 - The `CAR` distribution has been added to allow for use of conditional autoregressions which often are used in spatial and network models.
+- The dimensionality of model variables can now be parametrized through either of `shape`, `dims` or `size` (see [#4696](https://github.com/pymc-devs/pymc3/pull/4696)):
```
ricardoV94, line 13: A short doc with examples about these would be great to have in my opinion

michaelosthege: Where such a doc should live?
There are tons of other changes on `v4` that we should describe as well. Also the Readme section is essentially empty so far.
Let's write these docs after merging into master/main.

michaelosthege: For concrete examples about the shape/dims/size API please look at `test_shape_handling.py`

ricardoV94: Oh I didn't mean it should be a blocker at all!

pymc3/distributions/distribution.py
```diff
@@ -134,40 +234,165 @@ def __new__(cls, name, *args, **kwargs):
...
+            The created RV.
+        """
+        if "dims" in kwargs:
+            raise NotImplementedError("The use of a `.dist(dims=...)` API is not supported.")
+
+        shape, _, size = _validate_shape_dims_size(shape=shape, size=size)
+        ndim_supp = cls.rv_op.ndim_supp
+
+        if shape is not None:
+            ndim_expected = len(tuple(shape))
```
twiecki: Don't we know that `shape` is a tuple because of `_validate_shape_dims_size`?

twiecki:
I think this looks great!

pymc3/distributions/distribution.py
michaelosthege: I will go over these conversions again. There's still some room for optimization.

twiecki:
Just changed base to `v4`, not sure why we're getting conflicts there as #4693 was merged into v4?

michaelosthege:
After rebasing I'm now pushing this branch commit by commit, so we get CI results on all of them.

pymc3/tests/test_shape_handling.py
```diff
@@ -219,3 +220,21 @@ def test_sample_generate_values(fixture_model, fixture_sizes):
         prior = pm.sample_prior_predictive(samples=fixture_sizes)
         for rv in RVs:
             assert prior[rv.name].shape == size + tuple(rv.distribution.shape)
+
+
+def test_observed_with_column_vector():
```
ricardoV94: Maybe add a comment with the link to the original issue?

michaelosthege: I'm about to change that test again, because it turned out that #4700 did in fact not fix #4652.
Specifically, `int32`-typed `shapes` still break the RV broadcasting, but this was not covered by the tests.

I will mark these tests as XFAIL, linking to https://github.com/pymc-devs/aesara/issues/390

pymc3/distributions/distribution.py
```diff
@@ -134,40 +234,165 @@ def __new__(cls, name, *args, **kwargs):
...
-
         if not isinstance(name, string_types):
             raise TypeError(f"Name needs to be a string but got: {name}")
 
-        data = kwargs.pop("observed", None)
+        if rng is None:
+            rng = model.default_rng
 
-        total_size = kwargs.pop("total_size", None)
+        _, dims, _ = _validate_shape_dims_size(dims=dims)
```
twiecki: Can we move all this functionality into a separate function?

pymc3/distributions/distribution.py
```diff
@@ -134,40 +234,165 @@ def __new__(cls, name, *args, **kwargs):
...
+            Test value to be attached to the output RV.
+            Must match its shape exactly.
 
-        rv_var = cls.rv_op(*dist_params, **kwargs)
+        Returns
+        -------
+        rv : RandomVariable
+            The created RV.
+        """
+        if "dims" in kwargs:
```
twiecki, line 270: Same here, seems like this logic can be separated.

pymc3/distributions/distribution.py
michaelosthege: There would be a lot of kwargs to pass. Maybe it could help with testing, but I don't have much hope that it would improve readability a lot.

michaelosthege: After all that's just a refactoring.
First I'd like to improve at the level of how the `shape`/`size` variables are casted, sliced and their length measured. We can avoid unnecessary Subtensor/MakeVector operations this way.

michaelosthege:
Last time we made the mistake of not reviewing the tests.
So please: Don't just focus at `distribution.py`. I know it can--and will--be refactored, but it is way more important that we make sure that all the important ways how this API is used are properly covered in `test_shape_handling.py`.

pymc3/distributions/distribution.py
michaelosthege: I was able to simplify the branching logic quite a bit.
With how it is now, I don't think we gain anything by separating it out further, because only two lines above and two lines below would remain.

pymc3/distributions/distribution.py
michaelosthege: Here I did separate out the logic. The type hints are probably not 100 % water tight, but they should be a good basis for doing something with `mypy` in the future.

pymc3/distributions/distribution.py
```diff
@@ -115,13 +127,180 @@ def logcdf(op, var, rvs_to_values, *dist_params, **kwargs):
         return new_cls
 
 
+def _valid_ellipsis_position(items: Union[None, Shape, Dims, Size]) -> bool:
```
twiecki: I wonder if with all these changes we should places these into a `shape.py`.

twiecki: Then, the extra logic below in `__new__ ` and `dist()` could go in there too and everything would be nice and tidy.

michaelosthege: Let's not get distracted by "hiding everything under the carpet". There is actually very little extra logic left in these methods and extracting it out does not simplify anything.
Adding more and more layers of function calls doesn't solve any of our actual problems: A lot of these data types around here are way to promiscuous.

Unless someone points out exactly which block of code should be extracted and why, I have more useful things on my to do list.

twiecki:
Fair enough, I'll give it a try and we can see what it looks like.

pymc3/distributions/distribution.py
```diff
@@ -134,40 +313,127 @@ def __new__(cls, name, *args, **kwargs):
...
+            # A batch size was specified through `dims`, or implied by `observed`.
+            rv_out = change_rv_size(rv_var=rv_out, new_size=resize_shape, expand=True)
 
-        transform = kwargs.pop("transform", UNSET)
+        if dims is not None:
+            # Register named but previously unknown implied dimensions,
+            # such that they can be used for further downstream.
+            for di, dname in enumerate(dims[ndim_resize:]):
+                if not dname in model.dim_lengths:
+                    model.add_coord(dname, values=None, length=rv_out.shape[ndim_resize + di])
```
ricardoV94: I understand that it is a bit arbitrary, but it feels  like this would be better placed in `model.register_rv`, no?

pymc3/tests/test_shape_handling.py
```diff
@@ -219,3 +222,228 @@ def test_sample_generate_values(fixture_model, fixture_sizes):
...
+
+            with pytest.warns(None):
+                if parametrization == "implicit":
+                    rv = pm.Normal("rv", mu=mu).shape == param_shape
+                else:
+                    expected_shape = batch_shape + param_shape
+                    if parametrization == "shape":
+                        rv = pm.Normal("rv", mu=mu, shape=batch_shape + param_shape)
+                        assert rv.eval().shape == expected_shape
+                    # elif parametrization == "shape...":
```
ricardoV94: Should this be removed?

michaelosthege: No, the shape-Ellipsis is coming, even if just on a follow up PR. It doesn't hurt to have these things commented out in the meantime.

pymc3/distributions/distribution.py
michaelosthege: I split the conversion functions and made the type hints more precise.

After conversion `shape` can be `None`, `tuple` or `TensorVariable`. The conversion leaves `TensorVariable` shapes as they are, because conversion to tuple would create unnecessary Subtensor+MakeVector operations when the `shape` is used for RV creation later on.

pymc3/distributions/distribution.py
michaelosthege: These functions are now split and tested individually.

pymc3/tests/test_shape_handling.py
michaelosthege: @ricardoV94 I just added the shape-Ellipsis support!

pymc3/aesaraf.py
```diff
@@ -146,6 +148,19 @@ def change_rv_size(
         Expand the existing size by `new_size`.
 
     """
+    if new_size is None:
+        new_size = ()
+
+    # Check the dimensionality of the `new_size` kwarg
+    new_size_ndim = new_size.ndim if isinstance(new_size, Variable) else np.ndim(new_size)
```
ricardoV94, line 155: Very unimportant, but do we prefer to be explicit instead of just using `np.ndim` for both?

ricardoV94: Would `np.ndim` work for both types here? (not saying it's better...)

pymc3/distributions/distribution.py
```diff
@@ -115,13 +128,187 @@ def logcdf(op, var, rvs_to_values, *dist_params, **kwargs):
...
+            Most prominently: ``shape`` and ``size``
+
+        Returns
+        -------
+        rv : RandomVariable
+            The created RV, registered in the Model.
+        """
+
         try:
             from pymc3.model import Model
```
ricardoV94, line 170: Not related to this PR, but why is this import in here?

pymc3/distributions/distribution.py
```diff
@@ -134,40 +321,147 @@ def __new__(cls, name, *args, **kwargs):
...
+                create_size = tuple(shape)[:ndim_batch]
+        elif size is not None:
+            ndim_expected = ndim_supp + len(tuple(size))
+            ndim_batch = ndim_expected - ndim_supp
+            create_size = size
+
+        # Create the RV with a `size` right away.
+        # This is not necessarily the final result.
+        rv_out = cls.rv_op(*dist_params, size=create_size, **kwargs)
+        ndim_actual = rv_out.ndim
```
ricardoV94: Should this be called `ndim_implied` as it is in the `__new__` method, or is it actually used differently?

pymc3/distributions/distribution.py
```diff
@@ -134,40 +321,147 @@ def __new__(cls, name, *args, **kwargs):
...
+        ndim_implied = rv_out.ndim
+        resize_shape = None
 
-        if "shape" in kwargs:
-            raise DeprecationWarning("The `shape` keyword is deprecated; use `size`.")
+        # `dims` are only available with this API, because `.dist()` can be used
+        # without a modelcontext and dims are not tracked at the Aesara level.
+        if dims is not None:
+            ndim_resize, resize_shape, dims = _resize_from_dims(dims, ndim_implied, model)
+        elif observed is not None:
```
ricardoV94, line 215: Should a variable allow both `dims` and `observed`, and if so, are we checking whether they are compatible anywhere?  Do we even need to?

pymc3/model.py
```diff
@@ -981,7 +981,7 @@ def add_coord(
             )
         if name in self.coords:
             if not values.equals(self.coords[name]):
-                raise ValueError("Duplicate and incompatiple coordinate: %s." % name)
+                raise ValueError(f"Duplicate and incompatiple coordinate: {name}.")
```
ricardoV94: Typo: `incompatiBle`

pymc3/model.py
```diff
@@ -1019,14 +1019,18 @@ def set_data(
...
             New coordinate values for dimensions of the shared variable.
-            Must be provided for all named dimensions that change in length.
+            Must be provided for all named dimensions that change in length
+            and already have coordinate values.
         """
         shared_object = self[name]
         if not isinstance(shared_object, SharedVariable):
             raise TypeError(
-                f"The variable `{name}` must be defined as `pymc3.Data` inside the model to allow updating. "
+                f"The variable `{name}` must be a `SharedVariable` (e.g. `pymc3.Data`) allow updating. "
```
ricardoV94: typo: missing `TO` allow updating

pymc3/model.py
```diff
@@ -1019,14 +1019,18 @@ def set_data(
...
                 f"The current type is: {type(shared_object)}"
             )
+
+        if isinstance(values, list):
+            values = np.array(values)
```
ricardoV94, line 1074: Should this go into `pandas_to_array`?

pymc3/tests/test_data_container.py
```diff
@@ -283,6 +304,38 @@ def test_explicit_coords(self):
...
         assert pmodel.dim_lengths["columns"].eval() == 7
 
+    def test_symbolic_coords(self):
+        """
+        In v4 dimensions can be created without passing coordinate values.
+        Their lengths are then automatically linked to the corresponding Tensor dimension.
+        """
+        with pm.Model() as pmodel:
+            intensity = pm.Data("intensity", np.ones((2, 3)), dims=("row", "column"))
+            assert "row" in pmodel.dim_lengths
```
ricardoV94, line 314: Nitpick, but this would fail in the `isinstance` assertion below anyway no?

pymc3/model.py
```diff
@@ -1114,7 +1127,10 @@ def register_rv(self, rv_var, name, data=None, total_size=None, dims=None, trans
...
                 and data.owner is not None
             ):
-                raise TypeError("Observed data cannot consist of symbolic variables.")
+                raise TypeError(
+                    "Variables that depend on other nodes cannot be used for observed data."
+                    f"The data variable was: {data}"
+                )
 
             # `rv_var` is potentially a new variable (e.g. the original
             # variable could have its size changed to match the data, or be a
```
ricardoV94: > (e.g. the original variable could have its size changed to match the data,)

This is no longer true right?

pymc3/tests/test_shape_handling.py
```diff
@@ -219,3 +226,234 @@ def test_sample_generate_values(fixture_model, fixture_sizes):
...
+                    else:
+                        raise NotImplementedError("Invalid test case parametrization.")
+
+    def test_define_dims_on_the_fly(self):
+        with pm.Model() as pmodel:
+            agedata = aesara.shared(np.array([10, 20, 30]))
+
+            # Associate the "patient" dim with an implied dimension
+            age = pm.Normal("age", agedata, dims=("patient",))
+            assert "patient" in pmodel.dim_lengths
```
ricardoV94, line 302: Nitpick: first assert is not needed

pymc3/tests/test_shape_handling.py
```diff
@@ -219,3 +226,234 @@ def test_sample_generate_values(fixture_model, fixture_sizes):
...
+            effect = pm.Normal("effect", 0, dims=("patient",))
+            assert effect.ndim == 1
+            assert effect.eval().shape == (3,)
+
+            # Now change the length of the implied dimension
+            agedata.set_value([1, 2, 3, 4])
+            # The change should propagate all the way through
+            assert effect.eval().shape == (4,)
+
+    @pytest.mark.xfail(reason="Simultaneous use of size and dims is not implemented")
```
ricardoV94, line 315: Is this a future TODO?

pymc3/tests/test_shape_handling.py
```diff
@@ -219,3 +226,234 @@ def test_sample_generate_values(fixture_model, fixture_sizes):
...
+            assert y.eval().shape == (3, 2)
+            assert z.eval().shape == (3, 2)
+
+    @pytest.mark.xfail(reason="https://github.com/pymc-devs/aesara/issues/390")
+    def test_size32_doesnt_break_broadcasting():
+        size32 = at.constant([1, 10], dtype="int32")
+        rv = pm.Normal.dist(0, 1, size=size32)
+        assert rv.broadcastable == (True, False)
+
+    def test_observed_with_column_vector(self):
```
ricardoV94: Add comment linking to motivating GH issue?

pymc3/tests/test_shape_handling.py
```diff
@@ -219,3 +226,234 @@ def test_sample_generate_values(fixture_model, fixture_sizes):
         prior = pm.sample_prior_predictive(samples=fixture_sizes)
         for rv in RVs:
             assert prior[rv.name].shape == size + tuple(rv.distribution.shape)
+
+
+class TestShapeDimsSize:
```
ricardoV94, line 229: Since most of the shape issues occur with multivariate variables, should we add a similar test using `MvNormal`?

pymc3/distributions/distribution.py
OriolAbril: Not sure about the checking but we definitely should allow both and we already have several examples in the docs doing so too

pymc3/aesaraf.py
michaelosthege: Yes, looks like it works.
```suggestion
    new_size_ndim = np.ndim(new_size)
```

pymc3/distributions/distribution.py
michaelosthege: My guess would be to prevent a circular import.
If it works that'd be great, because we could use `Model` in some type hints.

But it's out of scope for this PR.

pymc3/distributions/distribution.py
michaelosthege: I'd rather rename the other one to `ndim_actual`, because both can include extra dimensions that are not implied by distribution support or parameters.

pymc3/distributions/distribution.py
michaelosthege: The changes from this PR make a few more restrictions on the compatibility of `observed`. Specifically the RV is no longer automatically resized based on `observed`, because that could conflict with a dimensionality parametrization through either of the other kwargs.

pymc3/model.py
michaelosthege:
```suggestion
                raise ValueError(f"Duplicate and incompatible coordinate: {name}.")
```

pymc3/model.py
michaelosthege:
```suggestion
                f"The variable `{name}` must be a `SharedVariable` (e.g. `pymc3.Data`) to allow updating. "
```

pymc3/model.py
michaelosthege, line 1074: Not sure why this is even there. Let's collect review edits and apply them in a batch. When it breaks we can amend that later.
```suggestion
```

pymc3/model.py
michaelosthege:
```suggestion
```

pymc3/tests/test_data_container.py
michaelosthege: Yes, but it doesn't hurt either I guess

pymc3/tests/test_shape_handling.py
michaelosthege: Yes, and probabily not even a big one. There's already a ticket for it: https://github.com/pymc-devs/pymc3/issues/4656

pymc3/tests/test_shape_handling.py
michaelosthege: I find these parametrized tests a little hard to read. The `MvNormal`-related weirdness should be covered by `test_mvnormal_shape_size_difference` further below already, don't you think?

pymc3/distributions/distribution.py
ricardoV94: Sounds good

pymc3/distributions/distribution.py
ricardoV94: I asked because of that `elif`, if there are `dims`, `_resize_from_observed` won't be called. Is this safe?

pymc3/tests/test_shape_handling.py
ricardoV94: They are definitely hard to read, but it seems this has two advantages over the `mvtest` below:
1. It uses the `non-dist` methods
2. It tests dims as well

pymc3/distributions/distribution.py
ricardoV94: Also do we have new tests for the "new" restrictions on observed?

twiecki:
I think this is starting to shape up nicely. Although I have to admit that I'm still quite concerned about the added complexity for something that *feels* like it should be simple (but provably turns out not to be). Usually I would not be too concerned about complexity, especially if it's well documented, commented, and tested like here. However, this is core code and `shape` is something that has caused us tremendous amounts of work in the past because we did not get it right the first time, so I think for developer generations to come it is important to get this right this time around.

I actually had a thought that I'd love to get your input on @michaelosthege. Currently a lot of the complexity comes from the fact that we are trying to convert `shape` to `size`, and that's really hard because of all the edge cases, then throw `dims`, `Ellipses`, and `observed` into the mix and here we are. My idea currently only pertains to `shape` and the simplest implementation to get that.

What if instead of doing the `shape`->`size` conversion, could we just create the RV with the specified `shape` and be done with it? I think we might be able to do this by just broadcasting the `dist_params` that are inputted to the specified `shape` and everything follows from there. 

We should be able to do that quite easily with `at.broadcast_to()`. There is still an issue with `ndims` but I found https://github.com/pymc-devs/aesara/blob/master/aesara/tensor/random/utils.py#L53 that seems to take care of that case.

Not thinking about `dims` or `Ellipses` for now, I think that should be quite straight forward and get us reasonably close with just a few lines (famous last words ;)).

From there, I think we'd be in a good spot where we can just use the `dims`->`shape` translation that you already implemented here in a separate PR.

Anyway, I know this thing is testing all of our patience to the maximum but thinking about our future selves and new contributors we owe it to us/them to think through various options here and pick the best path forward.

OriolAbril:
> Not thinking about dims or Ellipses for now, I think that should be quite straight forward and get us reasonably close with just a few lines (famous last words ;)).
>
> From there, I think we'd be in a good spot where we can just use the dims->shape translation that you already implemented here in a separate PR.

I am strongly against keeping the `dims->shape` approach. There was missunderstanding with what _implied_ dimensions/shape meant in the previous PR, but all the changes and improvements on how to define dims moving away from a dims->shape paradigm were extremely powerful. And I think that taking dims into account from more or less the beginning was part of the reason this was possible. Moreover, `shape` without ellipsis is mostly redundant given we already have `size` so it is probably true it will be relatively easy to implement, but it feels like tricking ourselves at the solitaire game (not sure if this is an English expression, we say that in Catalan). 

IMO, the main feature added by https://github.com/pymc-devs/pymc3/pull/4625 was awesome dim handling and allowing to use named dims without requiring to define all coordinate values beforehand and would therefore focus on `dims` and even remove `shape` completely if implementing it requires too much work.

I'm sorry I haven't had time to review the PR in detail yet, but I don't think I'll be able to this week either with gsoc and outreachy onboarding :confused:

michaelosthege:
@twiecki If you think it's easy go ahead and do it.
I don't think it's gonna work. There are distributions (e.g. `Categorical`) where the shape of input parameters has nothing to do with the output shape.

As you said, this PR "has shaped up nicely" and it does **exactly** what we were asking for API-wise.
Also it adds new features (symbolic, shape, size, dims handling, short-hand notation), finally gives us a consistent dimensionality API and **restores backwards compatibility**.

> From there, I think we'd be in a good spot where we can just use the `dims`->`shape` translation that you already implemented here in a separate PR.
> 
> Anyway, I know this thing is testing all of our patience to the maximum but thinking about our future selves and new contributors we owe it to us/them to think through various options here and pick the best path forward.

Let me be very clear:
This is the **third** PR I spent many hours on to get this right.

Every. Single. One. Of. Them. Was consistent, well-tested and had a clean git history.
The only thing that was "wrong" about the previous ones was that we had misunderstandings about how `size` was supposed to be interpreted.

My motivation about trying yet another approach because "it's too complicated" is **zero**.
There is not going to be a fourth PR about this from my side.

Oh and by the way, this is all the "complexity" were talking about:

<details><summary>After dropping type validation, sanity checks and docstrings</summary>

```python
def _resize_from_dims(
    dims: WeakDims, ndim_implied: int, model
) -> Tuple[int, StrongSize, StrongDims]:
    if Ellipsis in dims:
        # Auto-complete the dims tuple to the full length.
        # We don't have a way to know the names of implied
        # dimensions, so they will be `None`.
        dims = (*dims[:-1], *[None] * ndim_implied)

    ndim_resize = len(dims) - ndim_implied

    # The numeric/symbolic resize tuple can be created using model.RV_dim_lengths
    resize_shape = tuple(model.dim_lengths[dname] for dname in dims[:ndim_resize])
    return ndim_resize, resize_shape, dims


def _resize_from_observed(
    observed, ndim_implied: int
) -> Tuple[int, StrongSize, Union[np.ndarray, Variable]]:
    if not hasattr(observed, "shape"):
        observed = pandas_to_array(observed)
    ndim_resize = observed.ndim - ndim_implied
    resize_shape = tuple(observed.shape[d] for d in range(ndim_resize))
    return ndim_resize, resize_shape, observed


# Distribution.__new__
        
        # Create the RV without specifying testval, because the testval may have a shape
        # that only matches after replicating with a size implied by dims (see below).
        rv_out = cls.dist(*args, rng=rng, testval=None, **kwargs)
        ndim_implied = rv_out.ndim
        resize_shape = None

        # `dims` are only available with this API, because `.dist()` can be used
        # without a modelcontext and dims are not tracked at the Aesara level.
        if dims is not None:
            ndim_resize, resize_shape, dims = _resize_from_dims(dims, ndim_implied, model)
        elif observed is not None:
            ndim_resize, resize_shape, observed = _resize_from_observed(observed, ndim_implied)

        if resize_shape:
            # A batch size was specified through `dims`, or implied by `observed`.
            rv_out = change_rv_size(rv_var=rv_out, new_size=resize_shape, expand=True)


# Distribution.dist

        ndim_supp = cls.rv_op.ndim_supp
        ndim_expected = None
        ndim_batch = None
        create_size = None

        if shape is not None:
            if Ellipsis in shape:
                # Ellipsis short-hands all implied dimensions. Therefore
                # we don't know how many dimensions to expect.
                ndim_expected = ndim_batch = None
                # Create the RV with its implied shape and resize later
                create_size = None
            else:
                ndim_expected = len(tuple(shape))
                ndim_batch = ndim_expected - ndim_supp
                create_size = tuple(shape)[:ndim_batch]
        elif size is not None:
            ndim_expected = ndim_supp + len(tuple(size))
            ndim_batch = ndim_expected - ndim_supp
            create_size = size

        # Create the RV with a `size` right away.
        # This is not necessarily the final result.
        ndim_actual = rv_out.ndim
        ndims_unexpected = ndim_actual != ndim_expected

        if shape is not None and ndims_unexpected:
            if Ellipsis in shape:
                # Resize and we're done!
                rv_out = change_rv_size(rv_var=rv_out, new_size=shape[:-1], expand=True)
            else:
                # Recreate the RV without passing `size` to created it with just the implied dimensions.
                rv_out = cls.rv_op(*dist_params, size=None, **kwargs)

                # Now resize by any remaining "extra" dimensions that were not implied from support and parameters
                if rv_out.ndim < ndim_expected:
                    expand_shape = shape[: ndim_expected - rv_out.ndim]
                    rv_out = change_rv_size(rv_var=rv_out, new_size=expand_shape, expand=True)
```
</details>

<details><summary>Without the comments and type hints</summary>

```python
def _resize_from_dims(dims, ndim_implied, model):
    if Ellipsis in dims:
        dims = (*dims[:-1], *[None] * ndim_implied)

    ndim_resize = len(dims) - ndim_implied

    resize_shape = tuple(model.dim_lengths[dname] for dname in dims[:ndim_resize])
    return ndim_resize, resize_shape, dims


def _resize_from_observed(observed, ndim_implied):
    if not hasattr(observed, "shape"):
        observed = pandas_to_array(observed)
    ndim_resize = observed.ndim - ndim_implied
    resize_shape = tuple(observed.shape[d] for d in range(ndim_resize))
    return ndim_resize, resize_shape, observed


# Distribution.__new__
        
        rv_out = cls.dist(*args, rng=rng, testval=None, **kwargs)
        ndim_implied = rv_out.ndim
        resize_shape = None

        if dims is not None:
            ndim_resize, resize_shape, dims = _resize_from_dims(dims, ndim_implied, model)
        elif observed is not None:
            ndim_resize, resize_shape, observed = _resize_from_observed(observed, ndim_implied)

        if resize_shape:
            rv_out = change_rv_size(rv_var=rv_out, new_size=resize_shape, expand=True)


# Distribution.dist

        ndim_supp = cls.rv_op.ndim_supp
        ndim_expected = None
        ndim_batch = None
        create_size = None

        if shape is not None:
            if Ellipsis in shape:
                ndim_expected = ndim_batch = None
                create_size = None
            else:
                ndim_expected = len(tuple(shape))
                ndim_batch = ndim_expected - ndim_supp
                create_size = tuple(shape)[:ndim_batch]
        elif size is not None:
            ndim_expected = ndim_supp + len(tuple(size))
            ndim_batch = ndim_expected - ndim_supp
            create_size = size

        ndim_actual = rv_out.ndim
        ndims_unexpected = ndim_actual != ndim_expected

        if shape is not None and ndims_unexpected:
            if Ellipsis in shape:
                rv_out = change_rv_size(rv_var=rv_out, new_size=shape[:-1], expand=True)
            else:
                rv_out = cls.rv_op(*dist_params, size=None, **kwargs)

                if rv_out.ndim < ndim_expected:
                    expand_shape = shape[: ndim_expected - rv_out.ndim]
                    rv_out = change_rv_size(rv_var=rv_out, new_size=expand_shape, expand=True)
```
</details>

I'm going to rebase this branch now and incorporate the review feedback. Take it or leave it.

twiecki:
@michaelosthege Oh yeah, I didn't mean to imply you should do that. You definitely put in Herculian effort here and it's much appreciated. I can give that a try and if it doesn't work or we prefer this approach we'll just go this route. I just think it is important we really nail this shape stuff this time around, we'll all be grateful down the line.

pymc3/aesaraf.py
```diff
@@ -146,6 +148,19 @@ def change_rv_size(
         Expand the existing size by `new_size`.
 
     """
+    if new_size is None:
```
twiecki: I think with https://github.com/pymc-devs/pymc3/commit/0226047e36350f91b4562b6d6647e37806436589 we can probably drop this, I put this in as a short-cut but now with the fixed tests we should probably not keep supporting this.

pymc3/model.py
michaelosthege: Removing this part breaks `TestShapeDimsSize.tet_can_resize_data_defined_size` where a list of lists is used: `pmodel.set_data("x", [[1, 2], [3, 4], [5, 6]])`

I'm going to skip this and instead defer to #4650

pymc3/distributions/distribution.py
michaelosthege: This is a matter of having a single source of truth -- in this case the `dims` kwarg takes priority over the `observed`, because it belongs to the family of dimensionality kwargs. `Model.make_obs_var` raises an error if the dimensionality does not match:
```
if data.ndim != rv_var.ndim:
      raise ShapeError(
          "Dimensionality of data and RV don't match.", actual=data.ndim, expected=rv_var.ndim
      )
```

pymc3/tests/test_shape_handling.py
michaelosthege: Let's defer this to another PR.
Also we'll probably cover this in the examples/update guide too.

michaelosthege (dismissed):

OriolAbril (dismissed):

michaelosthege:
@twiecki it's been almost a week now. Are you still working on #4708 ?
I think we should merge this PR.
cc @ricardoV94 @Spaak

OriolAbril:
+1 on merging, further improvement and simplification can come in follow-up PRs, I don't think there is discussion in terms of functionality anymore so future PRs should not have any visible effect.

twiecki:
Let me try and finish up the shape.py PR soon.

michaelosthege:
As we decided in [todays lab meeting](https://github.com/pymc-devs/pymc3/wiki/Project-Meeting-Minutes#friday-june-4-2021) is rebased this branch so we can merge it.

There were quite a few git conflicts, as anticipated, but it looks like nothing broke.

pymc3/distributions/distribution.py
```diff
@@ -126,51 +178,87 @@ def __new__(cls, name, *args, **kwargs):
...
+        if dims is not None and "size" in kwargs:
+            raise ValueError(
+                f"Passing both `dims` ({dims}) and `size` ({kwargs['size']}) is not supported!"
             )
+        dims = convert_dims(dims)
 
-        initval = kwargs.pop("initval", testval)
+        # Create the RV without specifying testval, because the testval may have a shape
+        # that only matches after replicating with a size implied by dims (see below).
+        rv_out = cls.dist(*args, rng=rng, testval=None, **kwargs)
```
OriolAbril:
```suggestion
        # Create the RV without specifying initval, because the initval may have a shape
        # that only matches after replicating with a size implied by dims (see below).
        rv_out = cls.dist(*args, rng=rng, initval=None, **kwargs)
```

twiecki:
OK good to merge from my end!

OriolAbril (dismissed):

twiecki:
Thanks to everyone who helped get this over the finish line, especially the herculean efforts of @michaelosthege! Also thanks for helpful review from @brandonwillard, @ricardoV94, @OriolAbril and everyone else.

PR #4698, opened 2021-05-14
Fix `_check_start_shape` and `BinaryMetropolis.astep`

ricardoV94:
This PR fixes two small issues in the methods mentioned in the PR title.

Two tests are now failing because they end in a call to `az.summary`, is this known to still be broken?

### DIFF
--- a/pymc3/tests/test_examples.py
+++ b/pymc3/tests/test_examples.py
@@ -194,7 +194,7 @@ def build_disaster_model(masked=False):
 
 
 @pytest.mark.xfail(
-    reason="_check_start_shape fails with start dictionary"
+    reason="Arviz summary fails"
     # condition=(aesara.config.floatX == "float32"), reason="Fails on float32"
 )
 class TestDisasterModel(SeededTest):
@@ -222,7 +222,6 @@ class TestDisasterModel(SeededTest):
             az.summary(tr)
 
 
-@pytest.mark.xfail(reason="_check_start_shape fails with start dictionary")
 class TestLatentOccupancy(SeededTest):
     """
     From the PyMC example list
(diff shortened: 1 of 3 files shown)


pymc3/tests/test_examples.py
```diff
@@ -278,7 +277,7 @@ def test_run(self):
                 "theta": np.array(5, dtype="f"),
             }
             step_one = pm.Metropolis([model["theta_interval__"], model["psi_logodds__"]])
-            step_two = pm.BinaryMetropolis([model.z])
+            step_two = pm.BinaryMetropolis([model.rvs_to_values[model["z"]]])
```
ricardoV94, line 280: Do we need to pass the `value_var` explicitly, or it works as well with the `rv_var` (i.e., `model["z"]`)? 

The test passes but I didn't have the patience to check if it affected sampling.

The `ArrayStep` docs are ambiguous in this regard:
```
    Blocked step method that is generalized to accept vectors of variables.

    Parameters
    ----------
    vars: list
        List of variables for sampler.
```

brandonwillard: The value variables are given to the step methods.  See [here](https://github.com/pymc-devs/pymc3/blob/v4/pymc3/sampling.py#L198).

brandonwillard (approved):

pymc3/tests/test_examples.py
ricardoV94: I meant when specifying manually the steps as in this test

brandonwillard: In general, step methods require value variables as their `vars` argument/field, so it looks like your manual setup for that step method in the test is correct.

ricardoV94: Thanks, that was exactly my hunch.

Should we add some logic where we check if a variable manually assigned to a stepper contains a `value_var` and if so use that instead?

That would make manual assignments less cumbersome and error prone. What do you think?

brandonwillard: We only need to state what kind of variables they should be in a docstring.  Aside from that, we could perhaps check that the `vars` are actually value variables in the step method's `model`.

twiecki:
Thanks @ricardoV94!

PR #4700, opened 2021-05-15
Fix obs broadcast mismatch

### LINKED ISSUE #4652: Column vector breaks `observed`

brandonwillard:
This PR casts `new_shape` values in `pymc3.aesaraf.change_rv_size` before calling `RandomVariable.make_node`, which prevents unnecessary `Cast`s in the resulting `RandomVariable`'s `size` parameter when `new_shape` is a constant.

As a follow-up and/or alternative, `RandomVariable.make_node` can be updated in Aesara and accomplish the same thing.

Closes #4652.

### DIFF
--- a/pymc3/aesaraf.py
+++ b/pymc3/aesaraf.py
@@ -156,6 +156,10 @@ def change_rv_size(
             size = rv_node.op._infer_shape(size, dist_params)
         new_size = tuple(np.atleast_1d(new_size)) + tuple(size)
 
+    # Make sure the new size is int64 so that it doesn't unnecessarily pick
+    # up a `Cast` in some cases
+    new_size = at.as_tensor(new_size, ndim=1, dtype="int64")
+
     new_rv_node = rv_node.op.make_node(rng, new_size, dtype, *dist_params)
     rv_var = new_rv_node.outputs[-1]
     rv_var.name = name
--- a/pymc3/aesaraf.py
+++ b/pymc3/aesaraf.py
@@ -156,7 +156,7 @@ def change_rv_size(
             size = rv_node.op._infer_shape(size, dist_params)
         new_size = tuple(np.atleast_1d(new_size)) + tuple(size)
 
-    # Make sure the new size is int64 so that it doesn't unnecessarily pick
+    # Make sure the new size is a tensor. This helps to not unnecessarily pick
     # up a `Cast` in some cases
     new_size = at.as_tensor(new_size, ndim=1, dtype="int64")
 
--- a/pymc3/model.py
+++ b/pymc3/model.py
@@ -1116,8 +1116,6 @@ class Model(Factor, WithMemoization, metaclass=ContextMeta):
             ):
                 raise TypeError("Observed data cannot consist of symbolic variables.")
 
-            data = pandas_to_array(data)
-
             # `rv_var` is potentially a new variable (e.g. the original
             # variable could have its size changed to match the data, or be a
             # new graph that accounts for missing data)
--- a/pymc3/tests/test_aesaraf.py
+++ b/pymc3/tests/test_aesaraf.py
@@ -23,7 +23,7 @@ import pandas as pd
 import pytest
 import scipy.sparse as sps
 
-from aesara.graph.basic import Variable, ancestors
+from aesara.graph.basic import Constant, Variable, ancestors
 from aesara.tensor.random.basic import normal, uniform
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.subtensor import AdvancedIncSubtensor, AdvancedIncSubtensor1
@@ -66,6 +66,15 @@ def test_change_rv_size():
     rv_newer = change_rv_size(rv_new, new_size=(4,), expand=True)
     assert rv_newer.ndim == 3
     assert rv_newer.eval().shape == (4, 3, 2)
+
+    # Make sure we avoid introducing a `Cast` by converting the new size before
+    # constructing the new `RandomVariable`
+    rv = normal(0, 1)
+    new_size = np.array([4, 3], dtype="int32")
+    rv_newer = change_rv_size(rv, new_size=new_size, expand=False)
+    assert rv_newer.ndim == 2
+    assert isinstance(rv_newer.owner.inputs[1], Constant)
+    assert rv_newer.eval().shape == (4, 3)
 
 
 class TestBroadcasting:
--- a/pymc3/tests/test_distributions_random.py
+++ b/pymc3/tests/test_distributions_random.py
@@ -379,13 +379,9 @@ class BaseTestDistribution(SeededTest):
         sizes_to_check = self.sizes_to_check or [None, (), 1, (1,), 5, (4, 5), (2, 4, 2)]
         sizes_expected = self.sizes_expected or [(), (), (1,), (1,), (5,), (4, 5), (2, 4, 2)]
         for size, expected in zip(sizes_to_check, sizes_expected):
-            actual = change_rv_size(self.pymc_rv, size).eval().shape
+            pymc_rv = self.pymc_dist.dist(**self.pymc_dist_params, size=size)
+            actual = tuple(pymc_rv.shape.eval())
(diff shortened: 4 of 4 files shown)


pymc3/aesaraf.py
```diff
@@ -156,6 +156,10 @@ def change_rv_size(
             size = rv_node.op._infer_shape(size, dist_params)
         new_size = tuple(np.atleast_1d(new_size)) + tuple(size)
 
+    # Make sure the new size is int64 so that it doesn't unnecessarily pick
+    # up a `Cast` in some cases
+    new_size = at.as_tensor(new_size, ndim=1, dtype="int64")
```
michaelosthege: If the `new_size` is already a tensor variable, this line does not make sure it becomes `int64`:
```
at.as_tensor(at.constant([1,2], dtype="int32"), dtype="int64").dtype
# > "int32"
```
However it still makes #4652 less likely to occur.

```suggestion
    # Make sure the new size is a tensor. This helps to not unnecessarily pick
    # up a `Cast` in some cases
    new_size = at.as_tensor(new_size, ndim=1, dtype="int64")
```

michaelosthege (changes requested):

michaelosthege (approved):

pymc3/model.py
```diff
@@ -1116,8 +1116,6 @@ def register_rv(self, rv_var, name, data=None, total_size=None, dims=None, trans
             ):
                 raise TypeError("Observed data cannot consist of symbolic variables.")
 
-            data = pandas_to_array(data)
```
twiecki, line 1119: Won't this cause problems if a `DataFrame` is passed?

twiecki (approved):

pymc3/aesaraf.py
```diff
@@ -156,6 +156,10 @@ def change_rv_size(
             size = rv_node.op._infer_shape(size, dist_params)
         new_size = tuple(np.atleast_1d(new_size)) + tuple(size)
 
+    # Make sure the new size is a tensor. This helps to not unnecessarily pick
```
twiecki, line 159:
```suggestion
    # Make sure the new size is an int64 tensor. This helps to not unnecessarily pick
```

twiecki: Should we also link to the issue here?

pymc3/model.py
michaelosthege: One of the first things in `make_obs_var` is another call to `pandas_to_array`, that's why this one can go away

pymc3/aesaraf.py
michaelosthege: It's discoverable through the commits.
But on #4696 we'll make a few more robustness changes to `change_rv_size`. I can link the issue when rebasing the `reintro_shape` branch.

brandonwillard:
Unfortunately, the comment change introduced by commit https://github.com/pymc-devs/pymc3/pull/4700/commits/1716082e1fe6bd9c521a31a671017943afad94ac is actually *more* misleading than the original comment, because it makes a false statement instead of one that simply lacks context.

The `Cast` isn't avoided because `new_size` is a `Variable`&mdash;or any subclass thereof.  For instance, passing `new_size=at.as_tensor(np.array(..., dtype="int32"))` to `change_rv_size` will still result in a `Cast`ed `size` parameter in the resulting `RandomVariable` node.

The `Cast` is avoided only when `new_size` **isn't** a `Variable`, and is an `ndarray`, a primitive Python numeric type, a `tuple` or `list` of numeric types, etc.  In other words, the exact opposite of the new comment is true.

Here's the clarified form of the original comment:
```python
    # Make sure numeric/NumPy `new_size` values are converted to int64, so that
    # the `RandomVariable`'s new `size` doesn't unnecessarily pick up a `Cast`.
```

PR #4703, opened 2021-05-16
Refactor LogitNormal

themrzmaster:
partially helps #4686

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -3726,6 +3726,18 @@ class Logistic(Continuous):
             0 < s,
         )
 
+class LogitNormalRV(RandomVariable):
+    name = "logit_normal"
+    ndim_supp = 0
+    ndims_params = [0, 0]
+    dtype = "floatX"
+    _print_name = ("logitNormal", "\\operatorname{logitNormal}")
+
+    @classmethod
+    def rng_fn(cls, rng, mu, sigma, size=None):
+        return invlogit(stats.norm.rvs(loc=mu, scale=sigma, size=size, random_state=rng))
+
+logit_normal = LogitNormalRV()
 
 class LogitNormal(UnitContinuous):
     r"""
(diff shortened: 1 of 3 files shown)


pymc3/tests/test_distributions_random.py
```diff
@@ -513,6 +509,22 @@ class TestNormal(BaseTestDistribution):
         "check_pymc_draws_match_reference",
         "check_rv_size",
     ]
+class TestLogitNormal(BaseTestDistribution):
+    def logit_normal_rng_fn(self, rng, size, loc, scale):
+        return expit(st.norm.rvs(loc=loc, scale=scale, size=size, random_state=rng))
```
themrzmaster, line 514: is this the best way to test? I am having a problem here

pymc3/distributions/continuous.py
```diff
@@ -3726,6 +3726,18 @@ def logcdf(value, mu, s):
...
+class LogitNormalRV(RandomVariable):
+    name = "logit_normal"
+    ndim_supp = 0
+    ndims_params = [0, 0]
+    dtype = "floatX"
+    _print_name = ("logitNormal", "\\operatorname{logitNormal}")
+
+    @classmethod
+    def rng_fn(cls, rng, mu, sigma, size=None):
+        return invlogit(stats.norm.rvs(loc=mu, scale=sigma, size=size, random_state=rng))
```
ricardoV94: Should use `scipy.expit`. I think `invlogit` is the aesara symbolic counterpart. Here the output it shouldn't be symbolic anymore

pymc3/tests/test_distributions_random.py
ricardoV94: What problem are you having? Maybe it's related to my comment above?

ricardoV94:
@themrzmaster Can you add a test for the tau parametrization in test_distributions_random? There are some examples with the normal that you can copy from.

ricardoV94 (approved):
LGTM. Thank you for opening the PR!

PR #4705, opened 2021-05-16
Refactor Rice and Skew Normal distribution

farhanreynaldo:
This PR closes a subset of https://github.com/pymc-devs/pymc3/issues/4686, specifically Rice and Skew Normal distribution.

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -3465,8 +3465,8 @@ class RiceRV(RandomVariable):
     _print_name = ("Rice", "\\operatorname{Rice}")
 
     @classmethod
-    def rng_fn(cls, rng, nu, sigma, size=None):
-        return stats.rice.rvs(b=nu / sigma, scale=sigma, size=size, random_state=rng)
+    def rng_fn(cls, rng, b, sigma, size=None):
+        return stats.rice.rvs(b=b, scale=sigma, size=size, random_state=rng)
 
 
 rice = RiceRV()
@@ -3539,10 +3539,10 @@ class Rice(PositiveContinuous):
             sigma = sd
 
         nu, b, sigma = cls.get_nu_b(nu, b, sigma)
-        nu = at.as_tensor_variable(floatX(nu))
+        b = at.as_tensor_variable(floatX(b))
         sigma = at.as_tensor_variable(floatX(sigma))
 
-        return super().dist([nu, sigma], *args, **kwargs)
+        return super().dist([b, sigma], *args, **kwargs)
 
     @classmethod
     def get_nu_b(cls, nu, b, sigma):
@@ -3556,7 +3556,7 @@ class Rice(PositiveContinuous):
             return nu, b, sigma
         raise ValueError("Rice distribution must specify either nu" " or b.")
 
-    def logp(value, nu, sigma):
+    def logp(value, b, sigma):
         """
         Calculate log-probability of Rice distribution at specified value.
 
@@ -3570,12 +3570,10 @@ class Rice(PositiveContinuous):
         -------
         TensorVariable
         """
-        b = nu / sigma
         x = value / sigma
         return bound(
             at.log(x * at.exp((-(x - b) * (x - b)) / 2) * i0e(x * b) / sigma),
             sigma >= 0,
-            nu >= 0,
             value > 0,
         )
 
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -3193,7 +3193,7 @@ class SkewNormal(Continuous):
 
         return super().dist([mu, sigma, alpha], *args, **kwargs)
 
-    def logp(value, alpha, mu, sigma, tau):
+    def logp(value, alpha, mu, sigma):
         """
         Calculate log-probability of SkewNormal distribution at specified value.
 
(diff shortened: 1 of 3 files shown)


pymc3/tests/test_distributions_random.py
```diff
@@ -498,6 +492,33 @@ class TestMoyal(BaseTestDistribution):
...
+    reference_dist_params = {"loc": 0.0, "scale": 1.0, "a": 5.0}
+    reference_dist = seeded_scipy_distribution_builder("skewnorm")
+    tests_to_run = [
+        "check_pymc_params_match_rv_op",
+        "check_pymc_draws_match_reference",
+        "check_rv_size",
+    ]
+
+
+class TestRice(BaseTestDistribution):
```
ricardoV94, line 532: We should add an additional test with the `b` parametrization. There are some examples already of testing alternative parametrizations. For that the `params_match_rv_op` test is enough.

pymc3/tests/test_distributions_random.py
```diff
@@ -498,6 +492,33 @@ class TestMoyal(BaseTestDistribution):
     ]
 
 
+class TestSkewNormal(BaseTestDistribution):
```
ricardoV94, line 511: Similarly, we should add a test for the tau parametrization

ricardoV94 (commented):
Looks great. I left 2 small requests. Thanks for  opening the PR!

pymc3/tests/test_distributions_random.py
farhanreynaldo: Sure, I am going to follow the Normal distribution test for tau and sd.

pymc3/tests/test_distributions_random.py
farhanreynaldo: Idem.

pymc3/distributions/continuous.py
```diff
@@ -3471,6 +3454,21 @@ def logcdf(
...
+class RiceRV(RandomVariable):
+    name = "rice"
+    ndim_supp = 0
+    ndims_params = [0, 0]
+    dtype = "floatX"
+    _print_name = ("Rice", "\\operatorname{Rice}")
+
+    @classmethod
+    def rng_fn(cls, rng, nu, sigma, size=None):
+        return stats.rice.rvs(b=nu / sigma, scale=sigma, size=size, random_state=rng)
```
ricardoV94: Since the `logp` needs `b` anyway, maybe it makes sense to make the default parametrization in terms of `b` and `sigma` so that we can at least avoid the division `nu/sigma` in the scipy call. What do you think?

ricardoV94: Actually it seems we wouldn't even need `nu` in the `logp` altogether.

farhanreynaldo: I agree because the scipy function use `b` as the default, but should we adjust the documentation accordingly? Since it assumes that `b` is an alternative to the `nu`, rather than being the default parameter to pass.

ricardoV94: No need to change the documentation. It doesn't affect the behavior users see

pymc3/tests/test_distributions_random.py
```diff
@@ -498,6 +492,48 @@ class TestMoyal(BaseTestDistribution):
...
+    tests_to_run = [
+        "check_pymc_params_match_rv_op",
+        "check_pymc_draws_match_reference",
+        "check_rv_size",
+    ]
+
+
+class TestSkewNormalTau(BaseTestDistribution):
+    pymc_dist = pm.SkewNormal
+    pymc_dist_params = {"mu": 0.0, "tau": 1.0, "alpha": 5.0}
```
ricardoV94: Just realized sigma will be the same as tau (1). We should choose a test value where tau and sigma differ

farhanreynaldo: Indeed, I just noticed that. I might as well use the `get_tau_sigma` function.

pymc3/distributions/continuous.py
```diff
@@ -3157,51 +3172,25 @@ class SkewNormal(Continuous):
...
-        """
-        # mu, tau, _, alpha = draw_values(
-        #     [self.mu, self.tau, self.sigma, self.alpha], point=point, size=size
-        # )
-        # return generate_samples(
-        #     stats.skewnorm.rvs, a=alpha, loc=mu, scale=tau ** -0.5, dist_shape=self.shape, size=size
-        # )
-
-    def logp(self, value):
+    def logp(value, alpha, mu, sigma, tau):
```
ricardoV94: A test is falling because logp shouldn't expect tau in the signature. Only the arguments passed to super before.

ricardoV94:
I had to add an `xfail` to the rice `logp` test as some combinations failed on `float32`. I'll merge after the tests pass.

Great work!

farhanreynaldo:
> I had to add an `xfail` to the rice `logp` test as some combinations failed on `float32`. I'll merge after the tests pass.
> 
> Great work!

Thank you for your help and guidance!

ricardoV94 (dismissed):
Great work! Thanks :rocket:

ricardoV94:
A rebase was needed :D

ricardoV94 (approved):

Issue #4661, opened 2021-04-24
Flaky test `test_deterministic_of_observed_modified_interface`


ricardoV94:
```
FAILED pymc3/tests/test_sampling.py::TestSamplePPC::test_deterministic_of_observed_modified_interface
```

This test failed here: https://github.com/pymc-devs/pymc3/runs/2423911879

And I think I have seen it before as well. We should check if it's a newly introduced bug or just needs seeding.

ricardoV94:
Failed again here: https://github.com/pymc-devs/pymc3/pull/4646/checks?check_run_id=2512819254

ricardoV94:
Appeared again in https://github.com/pymc-devs/pymc3/runs/2597322712

ricardoV94:
Tentatively closed by #4707

PR #4706, opened 2021-05-17
refactor kumaraswamy

farhanreynaldo:
This PR closes a subset of https://github.com/pymc-devs/pymc3/issues/4686, which is Kumaraswamy distribution. But, I am not quite sure why the `check_pymc_draws_match_reference` test failed.

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -1375,7 +1375,7 @@ class Kumaraswamy(UnitContinuous):
         -------
         TensorVariable
         """
-        logcdf = at.log1p(-((1 - value ** a) ** b))
+        logcdf = log1mexp(-(b * at.log1p(-(value ** a))))
         return bound(at.switch(value < 1, logcdf, 0), value >= 0, a > 0, b > 0)
 
 
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -1359,6 +1359,24 @@ class Kumaraswamy(UnitContinuous):
 
         return bound(logp, value >= 0, value <= 1, a > 0, b > 0)
 
+    def logcdf(value, a, b):
+        r"""
+        Compute the log of cumulative distribution function for the Kumaraswamy distribution
+        at the specified value.
+
+        Parameters
+        ----------
+        value: numeric or np.ndarray or aesara.tensor
+            Value(s) for which log CDF is calculated. If the log CDF for
+            multiple values are desired the values must be provided in a numpy
+            array or Aesara tensor.
+
+        Returns
+        -------
+        TensorVariable
+        """
+        return bound(at.log1p(-((1 - value ** a) ** b)), value >= 0, value <= 1)
+
 
 class Exponential(PositiveContinuous):
     r"""
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -1275,7 +1275,7 @@ class KumaraswamyRV(RandomVariable):
     name = "kumaraswamy"
     ndim_supp = 0
     ndims_params = [0, 0]
-    dtype = "int64"
+    dtype = "floatX"
     _print_name = ("Kumaraswamy", "\\operatorname{Kumaraswamy}")
 
     @classmethod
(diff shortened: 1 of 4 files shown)


pymc3/distributions/continuous.py
```diff
@@ -1271,6 +1271,22 @@ def logcdf(value, alpha, beta):
         )
 
 
+class KumaraswamyRV(RandomVariable):
+    name = "kumaraswamy"
+    ndim_supp = 0
+    ndims_params = [0, 0]
+    dtype = "int64"
```
ricardoV94: dtype should be floatx

ricardoV94:
Completely optional for this PR but would you be interested in implementing the `logcdf` method for this distribution?

Gauging from wikipedia it has a very simple form: https://en.m.wikipedia.org/wiki/Kumaraswamy_distribution

ricardoV94:
Failing test is unrelated to this PR. Documented in #4661

farhanreynaldo:
> Completely optional for this PR but would you be interested in implementing the `logcdf` method for this distribution?
> 
> Gauging from wikipedia it has a very simple form: https://en.m.wikipedia.org/wiki/Kumaraswamy_distribution

My math is a little bit rusty, but does the following `logcdf` correct?
```
-b * at.log(1 - value ** a)
```
Because when I check against the following code, it doesn't match.
```
at.log(1 - (1 - value ** a) ** b)
```

ricardoV94:
You are missing the first `1 - ...`. If you want to remove the exponent you would need `log(1 - exp(b * log( 1 - value ** a))` but perhaps the outer log will suffice numerically. We can replace `log(1 - x)` by `log1p(-x)` which is a bit more stable.

That is `log1p(- (1 - value ** a) ** b)`

farhanreynaldo:
> You are missing the first `1 - ...`. If you want to remove the exponent you would need `log(1 - exp(b * log( 1 - value ** a))` but perhaps the outer log will suffice numerically. We can replace `log(1 - x)` by `log1p(-x)` which is a bit more stable.
> 
> That is `log1p(- (1 - value ** a) ** b)`

I totally missed that!  😩 Thank you for clarifying the equation.

I should also add the following test to `test_kumaraswamy`, right?
```python
def test_kumaraswamy(self):
        ...
        def scipy_log_cdf(value, a, b):
            return np.log1p(-(1 - value ** a) ** b)

        self.check_logp(Kumaraswamy, Unit, {"a": Rplus, "b": Rplus}, scipy_log_cdf)

**Comment by ricardoV94** [comment]
It should be `check_logcdf`, but other than that yes :)

File: pymc3/distributions/continuous.py [hunk 2]
```diff
@@ -1359,6 +1359,24 @@ def logp(value, a, b):
...
+        value: numeric or np.ndarray or aesara.tensor
+            Value(s) for which log CDF is calculated. If the log CDF for
+            multiple values are desired the values must be provided in a numpy
+            array or Aesara tensor.
+
+        Returns
+        -------
+        TensorVariable
+        """
+        return bound(at.log1p(-((1 - value ** a) ** b)), value >= 0, value <= 1)
```
ricardoV94: Value above one should give a logcdf of 0, so we need a switch: `at.switch(value < 1, normal expression, 0)`, and to remove `value <= 1` as a bound condition. We should still keep the `a > 0` and `b > 0` conditions as in the `logp`

ricardoV94:
Looks to be on the right track. I left you some comments. In the meantime you can add a release note mentioning the `logcdf` method was added to this distribution in the "New features": https://github.com/pymc-devs/pymc3/blob/v4/RELEASE-NOTES.md

pymc3/distributions/continuous.py
farhanreynaldo: does `tt` equal to `at`? Because I assume `tt` is Theano tensor and it is not available.

ricardoV94: Yes of course. It was a mistake, I am still used to writing with theano's abbreviation :p

Fixed the comment

ricardoV94:
The failing test seems to be a precision issue. This means we may want to break the exponent as in your first suggestion. I'll push some changes in a second.

ricardoV94 (approved):
I think it's ready! Thanks for the PR

farhanreynaldo:
> I think it's ready! Thanks for the PR

Thank you so much!

PR #4709, opened 2021-05-21
Robust generalized bart inference

aloctavodia:
This improve inference for non-gaussian-like likelihoods. Also small refactor. This remove the previously introduce scale parameter as is not longer necessary to tune it.  And now the `inv_link` parameter is string or a tuple of functions.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -7,7 +7,7 @@
 + Fix bug in the computation of the log pseudolikelihood values (SMC-ABC). (see [#4672](https://github.com/pymc-devs/pymc3/pull/4672)).
 
 ### New Features
-+ BART with non-gaussian likelihoods (see [#4675](https://github.com/pymc-devs/pymc3/pull/4675)).
++ BART with non-gaussian likelihoods (see [#4675](https://github.com/pymc-devs/pymc3/pull/4675) and [#4709](https://github.com/pymc-devs/pymc3/pull/4709)).
 
 ## PyMC3 3.11.2 (14 March 2021)
 
--- a/pymc3/distributions/bart.py
+++ b/pymc3/distributions/bart.py
@@ -15,6 +15,7 @@
 import numpy as np
 
 from pandas import DataFrame, Series
+from scipy.special import expit  # pylint: disable=unused-import
 
 from pymc3.distributions.distribution import NoDistribution
 from pymc3.distributions.tree import LeafNode, SplitNode, Tree
@@ -30,7 +31,6 @@ class BaseBART(NoDistribution):
         m=200,
         alpha=0.25,
         split_prior=None,
-        scale=None,
         inv_link=None,
         jitter=False,
         *args,
@@ -63,22 +63,30 @@ class BaseBART(NoDistribution):
             )
         self.m = m
         self.alpha = alpha
-        self.y_std = Y.std()
-
-        if scale is None:
-            self.leaf_scale = NormalSampler(sigma=None)
-        elif isinstance(scale, (int, float)):
-            self.leaf_scale = NormalSampler(sigma=Y.std() / self.m ** scale)
 
         if inv_link is None:
-            self.inv_link = lambda x: x
+            self.inv_link = self.link = lambda x: x
+            self.init_mean = self.Y.mean()
+            self.Y_un = self.Y
+        elif isinstance(inv_link, str):
+            if inv_link == "logistic":
+                self.inv_link = expit
+                self.link = lambda x: (x - 0.5) * 10
+            elif inv_link == "exp":
+                self.inv_link = np.exp
+                self.link = np.log
+                self.Y[self.Y == 0] += 0.0001
         else:
-            self.inv_link = inv_link
+            self.inv_link, self.link = inv_link
+
+        self.init_mean = self.link(self.Y.mean())
+        self.Y_un = self.link(self.Y)
 
         self.num_observations = X.shape[0]
         self.num_variates = X.shape[1]
         self.available_predictors = list(range(self.num_variates))
         self.ssv = SampleSplittingVariable(split_prior, self.num_variates)
+        self.initial_value_leaf_nodes = self.init_mean / self.m
         self.trees = self.init_list_of_trees()
         self.all_trees = []
         self.mean = fast_mean()
@@ -96,7 +104,7 @@ class BaseBART(NoDistribution):
         return X, Y, missing_data
 
     def init_list_of_trees(self):
-        initial_value_leaf_nodes = self.Y.mean() / self.m
+        initial_value_leaf_nodes = self.initial_value_leaf_nodes
         initial_idx_data_points_leaf_nodes = np.array(range(self.num_observations), dtype="int32")
         list_of_trees = []
         for i in range(self.m):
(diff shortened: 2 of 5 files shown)


pymc3/distributions/bart.py
```diff
@@ -63,22 +63,30 @@ def __init__(
...
 
         if inv_link is None:
-            self.inv_link = lambda x: x
+            self.inv_link = self.link = lambda x: x
+            self.init_mean = self.Y.mean()
+            self.Y_un = self.Y
+        elif isinstance(inv_link, str):
+            if inv_link == "logistic":
+                self.inv_link = expit
+                self.link = lambda x: (x - 0.5) * 10
```
ricardoV94, line 74: Shouldn't the link be the inverse of the expit?

pymc3/distributions/bart.py
```diff
@@ -63,22 +63,30 @@ def __init__(
...
+            elif inv_link == "exp":
+                self.inv_link = np.exp
+                self.link = np.log
+                self.Y[self.Y == 0] += 0.0001
```
ricardoV94, line 78: Should this use np.finfo.eps or something of the sort?

ricardoV94 (commented):
I am not familiar with the algorithm so apologies if I missed the main point. I just highlighted two things that seemed strange.

This was by no means a thorough review :)

pymc3/distributions/bart.py
aloctavodia: I will try. Although to avoid exactly zero value should be OK, np.finfo.eps maybe be too extreme.

pymc3/distributions/bart.py
aloctavodia: In principle you are right, but that will return -inf, inf. And also the purpose of self.link is just to allow the sampler to propose reasonable values for the leaf nodes. So this approximation seems to work ok.

aloctavodia: btw, no need for apologies and thanks for the comments. Actually my first attempt was to try something less ad-hoc than this.

ricardoV94:
I see, maybe a comment line mentioning that the link is just a rough approximation (and that this suffices) would be helpful to avoid confusion in the future.

In that case I also don't think any of the suggested changes are necessary

pymc3/distributions/bart.py
```diff
@@ -15,6 +15,7 @@
 import numpy as np
 
 from pandas import DataFrame, Series
+from scipy.special import expit  # pylint: disable=unused-import
```
ricardoV94: Why does this trigger pylint?

pymc3/distributions/bart.py
```diff
@@ -63,22 +63,30 @@ def __init__(
...
-
-        if scale is None:
-            self.leaf_scale = NormalSampler(sigma=None)
-        elif isinstance(scale, (int, float)):
-            self.leaf_scale = NormalSampler(sigma=Y.std() / self.m ** scale)
 
         if inv_link is None:
-            self.inv_link = lambda x: x
+            self.inv_link = self.link = lambda x: x
+        elif isinstance(inv_link, str):
```
ricardoV94, line 69: Perhaps raise a ValueError if the string is not valid? Or, more generally, catch a None link/invlink at the end?

ricardoV94 (commented):
Two more comments. Otherwise LGTM, but keep in mind I am by no means acquainted with the internal logic

pymc3/distributions/bart.py
aloctavodia: It doesn't, my mistake.

RELEASE-NOTES.md
```diff
@@ -7,7 +7,7 @@
 + Fix bug in the computation of the log pseudolikelihood values (SMC-ABC). (see [#4672](https://github.com/pymc-devs/pymc3/pull/4672)).
 
 ### New Features
-+ BART with non-gaussian likelihoods (see [#4675](https://github.com/pymc-devs/pymc3/pull/4675)).
++ BART with non-gaussian likelihoods (see [#4675](https://github.com/pymc-devs/pymc3/pull/4675) and [#4709](https://github.com/pymc-devs/pymc3/pull/4709)).
```
ricardoV94, line 10: Did you remove the gaussian likelihood or that old functionality is still present?

aloctavodia: That is still present, you can now have other likelihoods too. Actually "non-Gaussian" is probably not a very good description, as likelihoods like Student-T were already possible. The problem was more about the likelihood being bounded vs unbounded.

ricardoV94 (approved):

jlevy44:
Thanks for this PR, excited to see it merged to master soon. 

Testing right now on one of my datasets:
```
coords = {"group_id": np.arange(n_groups), "obs_id": np.arange(n_obs)}
with pm.Model(coords=coords) as bart_model:
    μ = pm.BART('μ', X, y, m=100, inv_link='logistic')

with bart_model:
    trace = pm.sample(cores = 2,  nuts={'target_accept':0.99} )
```

Right now I'm getting the following error:
```
DisconnectedInputError                    Traceback (most recent call last)
<ipython-input-18-855eef732e1d> in <module>
      1 with bart_model:
----> 2     trace = pm.sample(cores = 2,  nuts={'target_accept':0.99} )

/anaconda2/envs/py37/lib/python3.7/site-packages/pymc3/sampling.py in sample(draws, step, init, n_init, start, trace, chain_idx, chains, cores, tune, progressbar, model, random_seed, discard_tuned_samples, compute_convergence_checks, callback, jitter_max_retries, return_inferencedata, idata_kwargs, mp_ctx, pickle_backend, **kwargs)
    513             step = assign_step_methods(model, step, step_kwargs=kwargs)
    514     else:
--> 515         step = assign_step_methods(model, step, step_kwargs=kwargs)
    516 
    517     if isinstance(step, list):

/anaconda2/envs/py37/lib/python3.7/site-packages/pymc3/sampling.py in assign_step_methods(model, step, methods, step_kwargs)
    203             if has_gradient:
    204                 try:
--> 205                     tg.grad(model.logpt, var)
    206                 except (AttributeError, NotImplementedError, tg.NullTypeGradError):
    207                     has_gradient = False

/anaconda2/envs/py37/lib/python3.7/site-packages/theano/gradient.py in grad(cost, wrt, consider_constant, disconnected_inputs, add_names, known_grads, return_disconnected, null_gradients)
    622     for elem in wrt:
    623         if elem not in var_to_app_to_idx and elem is not cost and elem not in grad_dict:
--> 624             handle_disconnected(elem)
    625             grad_dict[elem] = disconnected_type()
    626 

/anaconda2/envs/py37/lib/python3.7/site-packages/theano/gradient.py in handle_disconnected(var)
    608         elif disconnected_inputs == "raise":
    609             message = utils.get_variable_trace_string(var)
--> 610             raise DisconnectedInputError(message)
    611         else:
    612             raise ValueError(

DisconnectedInputError: 
```

I removed the additional inputs and am still getting the error. 

I was also wondering how to incorporate random effects under the link specification? It seems that the conditional means cannot be augmented with a randomly drawn intercept prior to calculation of the likelihood, which seems problematic for binary classification tasks with repeated measures. Other examples for continuous outcomes run the particle gibbs step, then HMC/NUTS for drawing the random intercept. I would assume this can't be done here, since you need to apply the link first to compute the likelihood

aloctavodia:
To run BART, you need to define a likelihood, this is necessary to evaluate tree proposals in BART. Currently the BART implementation is limited because the values of μ are sampled by fitting to the residuals of "y". So even if you have something like μ+(a+b*X) in your model the values of μ will still be conditional on "y" so the term (a+b*X) will be "invisible" to BART. We have plans to refactor BART as part of the change introduced with V4 (or after that).

jlevy44:
Sounds good, thanks!

Issue #4567, opened 2021-03-24
Remove all uses of `*.tag.test_value` and replace with a new initial value framework


brandonwillard:
We need to set up a framework for setting and getting initial values for a model; one that doesn't rely on Aesara's test values.

In many cases, this may be as simple as renaming parts of the existing API (e.g. `Model.test_point` to `Model.initial_point`, `Model.check_test_point` to `Model.check_initial_point`, change the `testval` `Distribution` argument to `initial_value`, etc.), but, in some cases, we'll need to add new variables and logic.  

Specifically, it seems like we'll need to store the initial values in the `Model` object and, in multiple places, obtain those values from a `Model` object.

farhanreynaldo:
I would like to take on this issue. Should I do the pull request on `v4` branch? Would also really appreciate feedback and your help on this issue

brandonwillard:
You might need to wait until #4577 goes through, because it looks like I'll need to implement some of this for other reasons.

farhanreynaldo:
Sure thing, I will start working when it's merged.

ricardoV94:
For whomever takes this one, the `Flat` and `HalfFlat` are examples of pymc3 "distributions" that require special care because they cannot be sampled from directly, and as such an initial value has to be chosen some other way.

ricardoV94:
@brandonwillard can we close this one?

twiecki:
Sounds like this is implemented now.

PR #4711, opened 2021-05-23
Port Truncated Normal and Wald Distributions to V4

matteo-pallini:
Port Truncated Normal and Wald to V4 as per #4686 guidelines

Still need to do/check the followings:

- TruncatedNormal
+ [x] Need to investigate why `pymc3.tests.test_model.TestValueGradFunction.test_aesara_switch_broadcast_edge_cases_2` is failing
+ [x] Is it fine to rewrite a new RV? according to the issue there should be one already, but I couldn't find it 
+ [x] Is it fine to pass `transform` as argument in `dist` ?
+ [x] Is `_defaultval` deprecated? I haven't been able to find any use of it

- Wald
+ [x] Refactor as per [guidelines](https://github.com/pymc-devs/pymc3/issues/4686#issuecomment-839541040)
+ [ ] Investigate why `pymc3.tests.test_distributions_random.TestWaldAlpha` is failing.

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -46,7 +46,7 @@ from aesara.tensor.random.basic import (
     vonmises,
 )
 from aesara.tensor.random.op import RandomVariable
-from aesara.tensor.var import TensorVariable
+from aesara.tensor.var import TensorConstant, TensorVariable
 from scipy import stats
 from scipy.interpolate import InterpolatedUnivariateSpline
 from scipy.special import expit
@@ -162,8 +162,16 @@ class BoundedContinuous(Continuous):
             if cls.bound_args_indices[1] is not None:
                 upper = args[cls.bound_args_indices[1]]
 
-            lower = at.as_tensor_variable(lower) if lower is not None else None
-            upper = at.as_tensor_variable(upper) if upper is not None else None
+            lower = (
+                None
+                if lower is None or (isinstance(lower, TensorConstant) and lower.value == -np.inf)
+                else at.as_tensor_variable(lower)
+            )
+            upper = (
+                None
+                if upper is None or (isinstance(upper, TensorConstant) and upper.value == np.inf)
+                else at.as_tensor_variable(upper)
+            )
 
             return lower, upper
 
@@ -564,10 +572,18 @@ class TruncatedNormalRV(RandomVariable):
     ndim_supp = 0
     ndims_params = [0, 0, 0, 0]
     dtype = "floatX"
-    _print_name = ("dTruncatedNormal", "\\operatorname{dTruncatedNormal}")
+    _print_name = ("TruncatedNormal", "\\operatorname{TruncatedNormal}")
 
     @classmethod
-    def rng_fn(cls, rng, mu, sigma, lower, upper, size):
+    def rng_fn(
+        cls,
+        rng: np.random.RandomState,
+        mu: Union[np.ndarray, float],
+        sigma: Union[np.ndarray, float],
+        lower: Union[np.ndarray, float],
+        upper: Union[np.ndarray, float],
+        size: Optional[Union[List[int], int]],
+    ) -> np.ndarray:
         vals = stats.truncnorm.rvs(
             a=(lower - mu) / sigma,
             b=(upper - mu) / sigma,
@@ -661,33 +677,53 @@ class TruncatedNormal(BoundedContinuous):
     @classmethod
     def dist(
         cls,
-        mu=None,
-        sigma=None,
-        tau=None,
-        sd=None,
-        lower=None,
-        upper=None,
-        transform="auto",
+        mu: Optional[Union[float, np.ndarray]] = None,
+        sigma: Optional[Union[float, np.ndarray]] = None,
+        tau: Optional[Union[float, np.ndarray]] = None,
+        sd: Optional[Union[float, np.ndarray]] = None,
+        lower: Optional[Union[float, np.ndarray]] = None,
+        upper: Optional[Union[float, np.ndarray]] = None,
+        transform: str = "auto",
         *args,
         **kwargs,
-    ):
-        mu, sigma = _truncated_normal_prepare_mu_sigma_params(mu, sigma, tau, sd)
-        lower, lower_check, upper, upper_check = _truncated_normal_prepare_lower_and_upper(
-            lower, upper
-        )
+    ) -> RandomVariable:
+        sigma = sd if sd is not None else sigma
+        tau, sigma = get_tau_sigma(tau=tau, sigma=sigma)
+        sigma = at.as_tensor_variable(sigma)
+        tau = at.as_tensor_variable(tau)
+        mu = at.as_tensor_variable(floatX(mu))
+        assert_negative_support(sigma, "sigma", "TruncatedNormal")
+        assert_negative_support(tau, "tau", "TruncatedNormal")
 
-        if lower_check is None and upper_check is None:
-            _defaultval = mu
-        elif lower_check is None and upper_check is not None:
-            _defaultval = upper - 1.0
-        elif lower_check is not None and upper_check is None:
-            _defaultval = lower + 1.0
+        if lower is None and upper is None:
+            initval = mu
+        elif lower is None and upper is not None:
+            initval = upper - 1.0
+        elif lower is not None and upper is None:
+            initval = lower + 1.0
         else:
-            _defaultval = (lower + upper) / 2
+            initval = (lower + upper) / 2
 
-        return super().dist([mu, sigma, lower, upper], **kwargs)
+        lower = (
+            at.as_tensor_variable(floatX(lower))
+            if lower is not None
+            else at.as_tensor_variable(-np.inf)
+        )
+        upper = (
+            at.as_tensor_variable(floatX(upper))
+            if upper is not None
+            else at.as_tensor_variable(np.inf)
+        )
+        res = super().dist([mu, sigma, lower, upper], testval=initval, **kwargs)
+        return res
 
-    def logp(value, mu, sigma, lower, upper):
+    def logp(
+        value,
+        mu: Union[float, np.ndarray, TensorVariable],
+        sigma: Union[float, np.ndarray, TensorVariable],
+        lower: Union[float, np.ndarray, TensorVariable],
+        upper: Union[float, np.ndarray, TensorVariable],
+    ) -> RandomVariable:
         """
         Calculate log-probability of TruncatedNormal distribution at specified value.
 
@@ -701,67 +737,32 @@ class TruncatedNormal(BoundedContinuous):
         -------
         TensorVariable
         """
-        mu, sigma = _truncated_normal_prepare_mu_sigma_params(mu, sigma, tau=None, sd=None)
-        print(lower.eval())
-        lower, lower_check, upper, upper_check = _truncated_normal_prepare_lower_and_upper(
-            lower, upper
-        )
-        print(lower.eval())
-        norm = _truncated_normal_normalization(mu, sigma, lower, upper)
+        unbounded_lower = isinstance(lower, TensorConstant) and lower.value == -np.inf
+        unbounded_upper = isinstance(upper, TensorConstant) and upper.value == np.inf
+
+        if not unbounded_lower and not unbounded_upper:
+            lcdf_a = normal_lcdf(mu, sigma, lower)
+            lcdf_b = normal_lcdf(mu, sigma, upper)
+            lsf_a = normal_lccdf(mu, sigma, lower)
+            lsf_b = normal_lccdf(mu, sigma, upper)
+            norm = at.switch(lower > 0, logdiffexp(lsf_a, lsf_b), logdiffexp(lcdf_b, lcdf_a))
+        elif not unbounded_lower:
+            norm = normal_lccdf(mu, sigma, lower)
+        elif not unbounded_upper:
+            norm = normal_lcdf(mu, sigma, upper)
+        else:
+            norm = 0.0
+
         logp = Normal.logp(value, mu=mu, sigma=sigma) - norm
         bounds = [sigma > 0]
-        if lower_check is not None:
+        if lower is not None:
             bounds.append(value >= lower)
-        if upper_check is not None:
+        if upper is not None:
             bounds.append(value <= upper)
-        print([e.eval() for e in bounds])
         return bound(logp, *bounds)
 
     def _distr_parameters_for_repr(self):
         return ["mu", "sigma", "lower", "upper"]
-
-
-def _truncated_normal_prepare_mu_sigma_params(mu, sigma, tau, sd):
-    sigma = sd if sd is not None else sigma
-    tau, sigma = get_tau_sigma(tau=tau, sigma=sigma)
-    sigma = at.as_tensor_variable(sigma)
-    tau = at.as_tensor_variable(tau)
-    mu = at.as_tensor_variable(floatX(mu))
-    assert_negative_support(sigma, "sigma", "TruncatedNormal")
-    assert_negative_support(tau, "tau", "TruncatedNormal")
-    return mu, sigma
-
-
-def _truncated_normal_normalization(mu, sigma, lower, upper):
-    lower, lower_check, upper, upper_check = _truncated_normal_prepare_lower_and_upper(lower, upper)
-    if lower_check is None and upper_check is None:
-        return 0.0
-
-    if lower_check is not None and upper_check is not None:
-        lcdf_a = normal_lcdf(mu, sigma, lower)
-        lcdf_b = normal_lcdf(mu, sigma, upper)
-        lsf_a = normal_lccdf(mu, sigma, lower)
-        lsf_b = normal_lccdf(mu, sigma, upper)
-        return at.switch(lower > 0, logdiffexp(lsf_a, lsf_b), logdiffexp(lcdf_b, lcdf_a))
-    if lower_check is not None:
-        return normal_lccdf(mu, sigma, lower)
-    else:
-        return normal_lcdf(mu, sigma, upper)
-
-
-def _truncated_normal_prepare_lower_and_upper(lower, upper):
-    lower_check = at.as_tensor_variable(floatX(lower)) if lower is not None else lower
-    upper_check = at.as_tensor_variable(floatX(upper)) if upper is not None else upper
-
-    lower = (
-        at.as_tensor_variable(floatX(lower))
-        if lower is not None
-        else at.as_tensor_variable(-np.inf)
-    )
-    upper = (
-        at.as_tensor_variable(floatX(upper)) if upper is not None else at.as_tensor_variable(np.inf)
-    )
-    return lower, lower_check, upper, upper_check
 
 
 class HalfNormal(PositiveContinuous):
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -891,7 +891,7 @@ class WaldRV(RandomVariable):
 
     @classmethod
     def rng_fn(cls, rng, mu, lam, alpha, size):
-        return getattr(np.random.RandomState, cls.name)(rng, mu, lam, size) + alpha
+        return rng.wald(mu, lam, size=size) + alpha
 
 
 wald = WaldRV()
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -585,7 +585,7 @@ class TruncatedNormalRV(RandomVariable):
         upper: Union[np.ndarray, float],
         size: Optional[Union[List[int], int]],
     ) -> np.ndarray:
-        vals = stats.truncnorm.rvs(
+        return stats.truncnorm.rvs(
             a=(lower - mu) / sigma,
             b=(upper - mu) / sigma,
             loc=mu,
@@ -593,7 +593,6 @@ class TruncatedNormalRV(RandomVariable):
             size=size,
             random_state=rng,
         )
-        return vals
 
 
 truncated_normal = TruncatedNormalRV()
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -162,18 +162,17 @@ class BoundedContinuous(Continuous):
             if cls.bound_args_indices[1] is not None:
                 upper = args[cls.bound_args_indices[1]]
 
-            lower = (
-                at.as_tensor_variable(lower)
-                if lower is not None
-                and (isinstance(lower, TensorConstant) and not np.all(lower.value == -np.inf))
-                else None
-            )
-            upper = (
-                at.as_tensor_variable(upper)
-                if upper is not None
-                and (isinstance(upper, TensorConstant) and not np.all(upper.value == np.inf))
-                else None
-            )
+            if lower is not None:
+                if isinstance(lower, TensorConstant) and np.all(lower.value == -np.inf):
+                    lower = None
+                else:
+                    lower = at.as_tensor_variable(lower)
+
+            if upper is not None:
+                if isinstance(upper, TensorConstant) and np.all(upper.value == np.inf):
+                    upper = None
+                else:
+                    upper = at.as_tensor_variable(upper)
 
             return lower, upper
 
@@ -691,37 +690,24 @@ class TruncatedNormal(BoundedContinuous):
     ) -> RandomVariable:
         sigma = sd if sd is not None else sigma
         tau, sigma = get_tau_sigma(tau=tau, sigma=sigma)
-        single_param_length = 0 if not sigma.shape else sigma.shape[0]
         sigma = at.as_tensor_variable(sigma)
         tau = at.as_tensor_variable(tau)
         mu = at.as_tensor_variable(floatX(mu))
         assert_negative_support(sigma, "sigma", "TruncatedNormal")
         assert_negative_support(tau, "tau", "TruncatedNormal")
 
-        if lower is None and upper is None:
-            initval = mu
-        elif lower is None and upper is not None:
-            initval = upper - 1.0
-        elif lower is not None and upper is None:
-            initval = lower + 1.0
-        else:
-            initval = (lower + upper) / 2
+        # if lower is None and upper is None:
+        #     initval = mu
+        # elif lower is None and upper is not None:
+        #     initval = upper - 1.0
+        # elif lower is not None and upper is None:
+        #     initval = lower + 1.0
+        # else:
+        #     initval = (lower + upper) / 2
 
-        def _handle_missing_bound(bound_value, expected_input_length, replacement_value):
-            if bound_value is not None:
-                return at.as_tensor_variable(floatX(bound_value))
-            elif expected_input_length == 0:
-                return at.as_tensor_variable(floatX(replacement_value))
-            else:
-                return at.as_tensor_variable(
-                    floatX(np.repeat(replacement_value, expected_input_length))
-                )
-
-        lower = _handle_missing_bound(lower, single_param_length, -np.inf)
-        upper = _handle_missing_bound(upper, single_param_length, np.inf)
-
-        res = super().dist([mu, sigma, lower, upper], testval=initval, **kwargs)
-        return res
+        lower = at.as_tensor_variable(floatX(lower)) if lower is not None else at.constant(-np.inf)
+        upper = at.as_tensor_variable(floatX(upper)) if upper is not None else at.constant(np.inf)
+        return super().dist([mu, sigma, lower, upper], **kwargs)
 
     def logp(
         value,
@@ -743,8 +729,8 @@ class TruncatedNormal(BoundedContinuous):
         -------
         TensorVariable
         """
-        unbounded_lower = isinstance(lower, TensorConstant) and lower.value == -np.inf
-        unbounded_upper = isinstance(upper, TensorConstant) and upper.value == np.inf
+        unbounded_lower = isinstance(lower, TensorConstant) and np.all(lower.value == -np.inf)
+        unbounded_upper = isinstance(upper, TensorConstant) and np.all(upper.value == np.inf)
 
         if not unbounded_lower and not unbounded_upper:
             lcdf_a = normal_lcdf(mu, sigma, lower)
@@ -760,15 +746,14 @@ class TruncatedNormal(BoundedContinuous):
             norm = 0.0
 
         logp = Normal.logp(value, mu=mu, sigma=sigma) - norm
-        bounds = [sigma > 0]
-        if lower is not None:
+        bounds = []
+        if not unbounded_lower:
             bounds.append(value >= lower)
-        if upper is not None:
+        if not unbounded_upper:
             bounds.append(value <= upper)
+        if not unbounded_lower and not unbounded_upper:
+            bounds.append(lower <= upper)
         return bound(logp, *bounds)
-
-    def _distr_parameters_for_repr(self):
-        return ["mu", "sigma", "lower", "upper"]
 
 
 class HalfNormal(PositiveContinuous):
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -905,24 +905,8 @@ class WaldRV(RandomVariable):
     _print_name = ("Wald", "\\operatorname{Wald}")
 
     @classmethod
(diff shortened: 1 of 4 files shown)


ricardoV94:
> Is it fine to rewrite a new RV? according to the issue there should be one already, but I couldn't find it

I might have confused it with the truncexpon

pymc3/distributions/continuous.py
```diff
@@ -723,42 +686,69 @@ def logp(self, value):
...
-        if self.lower_check is not None:
-            return normal_lccdf(mu, sigma, self.lower)
-        else:
-            return normal_lcdf(mu, sigma, self.upper)
-
     def _distr_parameters_for_repr(self):
         return ["mu", "sigma", "lower", "upper"]
 
 
+def _truncated_normal_prepare_mu_sigma_params(mu, sigma, tau, sd):
```
ricardoV94: I think this would be better directly in the dist classmethod. This sort of initialization logic is what that method is exactly for, so it's okay to have it there.

It shouldn't be needed again in logp.

pymc3/distributions/continuous.py
```diff
@@ -723,42 +686,69 @@ def logp(self, value):
...
+    tau, sigma = get_tau_sigma(tau=tau, sigma=sigma)
+    sigma = at.as_tensor_variable(sigma)
+    tau = at.as_tensor_variable(tau)
+    mu = at.as_tensor_variable(floatX(mu))
+    assert_negative_support(sigma, "sigma", "TruncatedNormal")
+    assert_negative_support(tau, "tau", "TruncatedNormal")
+    return mu, sigma
+
+
+def _truncated_normal_normalization(mu, sigma, lower, upper):
```
ricardoV94: This can go directly in the logp method I think. Is it used anywhere else? Or at least a classmethod to keep things more organized

pymc3/distributions/continuous.py
```diff
@@ -723,42 +686,69 @@ def logp(self, value):
...
         bounds = [sigma > 0]
-        if self.lower_check is not None:
-            bounds.append(value >= self.lower)
-        if self.upper_check is not None:
-            bounds.append(value <= self.upper)
+        if lower_check is not None:
+            bounds.append(value >= lower)
+        if upper_check is not None:
+            bounds.append(value <= upper)
+        print([e.eval() for e in bounds])
```
ricardoV94: print statement

pymc3/distributions/continuous.py
```diff
@@ -723,42 +686,69 @@ def logp(self, value):
...
-
-        norm = self._normalization()
-        logp = Normal.dist(mu=mu, sigma=sigma).logp(value) - norm
-
+        mu, sigma = _truncated_normal_prepare_mu_sigma_params(mu, sigma, tau=None, sd=None)
+        print(lower.eval())
+        lower, lower_check, upper, upper_check = _truncated_normal_prepare_lower_and_upper(
+            lower, upper
+        )
+        print(lower.eval())
```
ricardoV94: print statement

pymc3/distributions/continuous.py
matteo-pallini: Ops, thanks

ricardoV94:
> Is `_defaultval` deprecated? I haven't been able to find any use of it

Kind of, we still need to refactor the testval/ initialization point logic for V4 as discussed in #4567

pymc3/distributions/continuous.py
matteo-pallini: Was it considered to add to the pre-commit checks also a check for print statements?

pymc3/distributions/continuous.py
```diff
@@ -723,42 +686,69 @@ def logp(self, value):
...
+
+
+def _truncated_normal_prepare_lower_and_upper(lower, upper):
+    lower_check = at.as_tensor_variable(floatX(lower)) if lower is not None else lower
+    upper_check = at.as_tensor_variable(floatX(upper)) if upper is not None else upper
+
+    lower = (
+        at.as_tensor_variable(floatX(lower))
+        if lower is not None
+        else at.as_tensor_variable(-np.inf)
```
ricardoV94: Are these `infs` used anywhere?

pymc3/distributions/continuous.py
ricardoV94: @MarcoGorelli suggested a way this could be done on the slack, following your suggestion. Would either one of you be interested in implementing this?

MarcoGorelli: I'm gonna be pretty busy in the next two weeks, but this script should work as a local hook:

```python
import ast
import sys

class Visitor(ast.NodeVisitor):
    def __init__(self, file):
        self.file = file
    def visit_Call(self, node: ast.Call) -> None:
        if isinstance(node.func, ast.Name) and node.func.id == 'print':
            sys.stdout.write(f'{self.file}:{node.lineno}:{node.col_offset} found print statement\n')
            sys.exit(1)

if __name__ == '__main__':
    for file in sys.argv[1:]:
        with open(file) as fd:
            content = fd.read()
        tree = ast.parse(content)
        visitor = Visitor(file)
        visitor.visit(tree)
```

@drabbit17 if you wanted to submit this as a separate PR, I'll review it

ricardoV94:
> Is it find to pass transform as argument in dist as per TruncatedNormal?

I am not sure what you mean. The BoundedContinuous class takes care of default transforms. It is possible for a user to specify another transform which will overwrite the default one. The `TruncatedNormal.dist()` initialization, on the other hand, will not do anything with transforms, they only matter for normal distributions intialized within a model.

I am not sure which of these you are referring to.

ricardoV94:
> Need to investigate why pymc3.tests.test_model.TestValueGradFunction.test_aesara_switch_broadcast_edge_cases_2 is failing

That test probably needs to be slightly refactored for V4. The dlogp call might need to be tweaked. What error are you seeing?

matteo-pallini:
`I am not sure what you mean. The BoundedContinuous ...`
Sorry, the bullet point wasn't really helpful. I wrote it as a reference for myself and didn't consider how cryptic it was. I was referring to the 2nd case you mentioned. 

I thought that the `transform` argument if passed to `dist` could have reached the `rv_op` call in `Distribution.dist` through `**kwargs` and that the logic to handle `transform` had been moved to aesara. But actually that would have not worked python-wise. So, I guess that the `transform` logic (and argument) will be simply removed with the refactoring.

`The dlogp call might need to be tweaked. What error are you seeing?`
`m.dlogp([mu])({"mu": 0})` is consistently 0. I need to familiarize with the internals of `Model` and `logp`/`dlogp` before being able to tweak the test appropriately. Thanks for letting me know that changing the test is an option worth considering

pymc3/distributions/continuous.py
matteo-pallini: Happy to give it a stab

pymc3/distributions/continuous.py
matteo-pallini: I guess they would be used when either `lower` or `upper` is None. There were a couple of tests for the cases in which either `upper` or `lower` were None. ie `TestTruncatedNormalLower` and `TestTruncatedNormalUpper`

ricardoV94: It seems to me they are not used, because the logp changes based on this and doesn't include the `None` extremes anyway. I am not even sure we need to separate between `lower` and `lower_check`.

ricardoV94:
> I thought that the transform argument if passed to dist could have reached the rv_op call in Distribution.dist through **kwargs and that the logic to handle transform had been moved to aesara. But actually that would have not worked python-wise. So, I guess that the transform logic (and argument) will be simply removed with the refactoring.

Only the first argument (the list of parameters) and `size/shape` are ever passed to the rv_op. The other kwargs are intercepted in `Distribution.__new__()` to be used there or forwarded to `Model.register_rv()`

ricardoV94:
@drabbit17 I pushed a tiny change for the failing test. The issue was that we were passing the RandomVariable to the dlogp function instead of the logp "value" variable. It is an expected V3->V4 refactoring

ricardoV94:
Hi @drabbit17, any progress on this PR?

We merged `V4` into `main` so you will have to redirect the target of this PR.

Let me know if you need any help.

matteo-pallini:
> We merged V4 into main so you will have to redirect the target of this PR.

Thanks for the update and congrats!

> Hi @drabbit17, any progress on this PR?

Sorry for the lack of progress here. During the last 2-3 weeks work left me with little/no mental bandwidth. I should be able to pick it up again on the weekend. I would like to, but please feel free to re-assign the issue to someone else in case I am being a blocker, or someone else is keen to pick it earlier than that.

ricardoV94:
@drabbit17 There is no rush, just wanted to check what was your status.

Also, do I understand correctly from the PR title that you intended to refactor the `Wald` distribution?

matteo-pallini:
> you intended to refactor the Wald distribution?

Yes

pymc3/distributions/continuous.py
```diff
@@ -632,101 +655,55 @@ class TruncatedNormal(BoundedContinuous):
...
+        if lower is None and upper is None:
+            initval = mu
+        elif lower is None and upper is not None:
+            initval = upper - 1.0
+        elif lower is not None and upper is None:
+            initval = lower + 1.0
+        else:
+            initval = (lower + upper) / 2
+
+        lower = (
```
ricardoV94: We shouldn't transform away from `None` unless it is needed, otherwise the branch logic in the logp (and also in the interval transform) is in vain, as it relies on a `None` check. Was this giving you problems?

In the `rng_fn` we can add extra logic to check for `None` and convert to `infs` if needed. If anything, in the `dist` method we could check for `np.inf` and convert those to `None`.

matteo-pallini: Sorry for that. I still have a partial understanding of the whole picture. Haven't had a chance yet to go through the sampling/inference part. So, I still don't know how `logp` is actually used (ie my guess from you comment is that it takes as arguments the dist_params that have already been processed by `dist`, ie the RandomVariable node inputs).

If the `None` are not handled `make_node` in `RandomVariable` will fail, as it's expecting objects that could be converted to tensor variables. So, I don't think that moving that check to `rng_fn` would solve the problem
https://github.com/aesara-devs/aesara/blob/3acdd78e7a5d27e70cd8de3fc00bd5f010c65ee3/aesara/tensor/random/op.py#L360

Stack trace
```
self = <pymc3.tests.test_distributions_random.TestTruncatedNormalLowerTau object at 0x7fb50c50a3d0>

    def test_distribution(self):
        self.validate_tests_list()
>       self._instantiate_pymc_rv()

pymc3/tests/test_distributions_random.py:288: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pymc3/tests/test_distributions_random.py:298: in _instantiate_pymc_rv
    self.pymc_rv = self.pymc_dist.dist(
pymc3/distributions/continuous.py:709: in dist
    res = super().dist([mu, sigma, lower, upper], testval=initval, **kwargs)
pymc3/distributions/distribution.py:285: in dist
../../miniconda3/envs/pymc3-dev-py38/lib/python3.8/site-packages/aesara/tensor/random/op.py:323: in __call__
    res = super().__call__(rng, size, dtype, *args, **kwargs)
../../miniconda3/envs/pymc3-dev-py38/lib/python3.8/site-packages/aesara/graph/op.py:271: in __call__
    node = self.make_node(*inputs, **kwargs)
../../miniconda3/envs/pymc3-dev-py38/lib/python3.8/site-packages/aesara/tensor/random/op.py:360: in make_node
    dist_params = tuple(
../../miniconda3/envs/pymc3-dev-py38/lib/python3.8/site-packages/aesara/tensor/random/op.py:361: in <genexpr>
    as_tensor_variable(p) if not isinstance(p, Variable) else p
../../miniconda3/envs/pymc3-dev-py38/lib/python3.8/site-packages/aesara/tensor/__init__.py:41: in as_tensor_variable
    return _as_tensor_variable(x, name, ndim, **kwargs)
../../miniconda3/envs/pymc3-dev-py38/lib/python3.8/functools.py:875: in wrapper
```

matteo-pallini: I wonder whether checking for `inf` and `-inf` rather than `None` in `logp` and in the interval transform could be a feasible approach

ricardoV94: That's spot on. I didn't see it immediately.

One ugly solution is to have 3 versions of the distribution (each with its own transform and logp / random methods) but I would rather not.

Another is to wrap the infinity arguments in an `at.constant` which we can safely query (I think) when calling the logp method (first checking if it's a constant and then if it's +- infinity). 

For the interval transform we can define the different variations at creation time in the `new` or `dist` classmethod (not sure which).

ricardoV94: Seems like you got ahead of me!

ricardoV94: > I wonder whether checking for `inf` and `-inf` rather than `None` in `logp` and in the interval transform could be a feasible approach

Yeah we can but only within an `at.constant` otherwise we would need symbolic branching with `at.switch` instead of the python `if else`. Since the point is to return a more optimized graph I would go with the constant + python branching.

pymc3/tests/test_distributions_random.py
```diff
@@ -525,6 +500,147 @@ def seeded_kumaraswamy_rng_fn(self):
...
+    pymc_dist = pm.Wald
+    reference_dist = seeded_wald_rng_fn
+    tests_to_run = [
+        "check_pymc_params_match_rv_op",
+        "check_pymc_draws_match_reference",
+        "check_rv_size",
+    ]
+
+
+class TestWaldPureScipy(BaseTestDistribution):
```
matteo-pallini: I would expect this to pass, but it doesn't. I wonder whether it fails because somehow the random sampler seeding is not returning the expected results when we have more than one distribution being sampled in `rng_fn`. In this case both `rng.normal(size=size)` and `rng.uniform(size=size)`

pymc3/distributions/continuous.py
ricardoV94: Do we have tests that cover this specific `+-inf` behavior?

pymc3/tests/test_distributions_random.py
ricardoV94: If the implementation is not the same, it's not necessarily the case they would match. Maybe the order the rng methods are called in also matters? By the way, is there a reason why we can't simply delegate to the `scipy` method in the new `RandomVariable`?

pymc3/distributions/continuous.py
matteo-pallini: `TestTruncatedNormalLowerTau` and `TestTruncatedNormalUpperTau`

ricardoV94: I mean testing the interval and the logp is actually doing what it should (returning different expressions based on whether one or both of the inputs are `None`)

pymc3/distributions/continuous.py
```diff
@@ -162,8 +162,16 @@ def transform_params(rv_var):
             if cls.bound_args_indices[1] is not None:
                 upper = args[cls.bound_args_indices[1]]
 
-            lower = at.as_tensor_variable(lower) if lower is not None else None
-            upper = at.as_tensor_variable(upper) if upper is not None else None
+            lower = (
+                None
+                if lower is None or (isinstance(lower, TensorConstant) and lower.value == -np.inf)
```
ricardoV94: I think this will raise an error if you get a non scalar constant in other contexts

matteo-pallini: Good point! I added a test for that, which is currently failing `pymc3.tests.test_distributions.TestBoundedContinuous.test_missing_upper_bound_array` still need to fix it though, I should be able to do it later today. I guess it will be a matter of having a check for scalar and one for arrays. btw I should also create an array of np.infs within `dist` when the other params are arrays.

pymc3/distributions/continuous.py
matteo-pallini: I added `pymc3.tests.test_distributions.TestBoundedContinuous` to test the interval. I haven't been able to test `BoundedContinuous` directly, so I had to do so through a `TruncatedNormal`

I also added a slightly more convoluted test (which seems sensible) for the interval behavior `pymc3.tests.test_distributions.TestBoundedContinuous.test_missing_upper_bound_with_richer_context` which is failing. I need to investigate why

I added a couple of tests for the logp with unbounded intervals `pymc3.tests.test_distributions.TestMatchesScipy.test_truncated_normal`.

Let me know if you think it makes sense to add more

pymc3/distributions/continuous.py
ricardoV94: By the way, I have been working on something related for Censored distributions, and this is how I wrote the "constant inf check" so far: https://github.com/aesara-devs/aeppl/blob/46fac1347d129e0a6c7391c87bbae456f51aa009/aeppl/truncation.py#L36 (related to https://github.com/aesara-devs/aeppl/pull/22)

I am not very happy with the solution. I am still thinking whether it makes more sense to subclass (in my case the Censored class), in your case the Truncated Normal, into left-truncated, right-truncated, and truncated, depending on the bounds.

pymc3/tests/test_distributions.py
```diff
@@ -2707,6 +2723,65 @@ def test_bound():
         BoundPoissonPositionalArgs = Bound(Poisson, upper=6)("x", 2.0)
 
 
+class TestBoundedContinuous:
```
ricardoV94, line 2732: Tests look like they go in the right direction. I'll check them more carefully later, now I just wanted to mention a nitpick: I would pass always the missing lower / upper explicitly as `None`, to make them easier to read.

pymc3/distributions/continuous.py
matteo-pallini: Thanks for sharing the link, I will give it a look later.

Possibly it may just be as simple as doing as per below:
```python
lower = (
      at.as_tensor_variable(lower)
      if lower is not None
      and (isinstance(lower, TensorConstant) and -np.inf not in lower.value)
      else None
  )
```

I pushed an updated (and not too ugly) approach that is passing the test with arrays. I need to check it more thoroughly though

ricardoV94: Can't read that code in my mind right now, but I am thinking of the case where you have lower = [-np.inf, 0], i.e., when one of the replications of the variable is lower bounded and the other is not, in which case the logp and transform should treat the variable as being lowerbounded. Not sure if your snippet would do the right thing.

pymc3/tests/test_distributions_random.py
matteo-pallini: > Maybe the order the rng methods are called in also matters

yeah I think that in that case the order of the rng mattered. I think that the Scipy implementation of the Wald distribution doesn't match the PyMC one though. The Numpy one seems to be right though, so I replace the custom sampling with the numpy one

https://numpy.org/doc/stable/reference/random/generated/numpy.random.wald.html

pymc3/distributions/continuous.py
matteo-pallini: Good point, I only realized it after having shared the snippet. The approach with `np.all` is pretty neat! thanks for sharing it

matteo-pallini:
One test still failing, everything else should be close to be finalised (as far as I can tell)

pymc3/distributions/continuous.py
```diff
@@ -632,99 +673,63 @@ class TruncatedNormal(BoundedContinuous):
...
-        if self.lower_check is None and self.upper_check is None:
-            self._defaultval = mu
-        elif self.lower_check is None and self.upper_check is not None:
-            self._defaultval = self.upper - 1.0
-        elif self.lower_check is not None and self.upper_check is None:
-            self._defaultval = self.lower + 1.0
-        else:
-            self._defaultval = (self.lower + self.upper) / 2
-
+        single_param_length = 0 if not sigma.shape else sigma.shape[0]
```
ricardoV94: The shape could be determined by `mu` or even the other `bound` as well.

Also, the shape may be symbolic (if you have a symbolic input). I am not sure if `np.repeat` is "overloaded" by Aesara if the shape is symbolic, but there should be an `at.repeat` equivalent that you can safely use.

pymc3/distributions/continuous.py
```diff
@@ -900,6 +897,21 @@ def _distr_parameters_for_repr(self):
...
+class WaldRV(RandomVariable):
+    name = "wald"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0]
+    dtype = "floatX"
+    _print_name = ("Wald", "\\operatorname{Wald}")
+
+    @classmethod
+    def rng_fn(cls, rng, mu, lam, alpha, size):
+        return getattr(np.random.RandomState, cls.name)(rng, mu, lam, size) + alpha
```
ricardoV94:
```suggestion
        rng.wald(mu, lam, size=size) + alpha
```
Does this work?

pymc3/distributions/continuous.py
matteo-pallini: I played a bit with different options and I cannot find a decent one supporting symbolic shapes.

If use a python if else I will be able to return a scalar when scalars as passed as inputs and an array otherwise. However I won't be able to use the symbolic shape (unless I eval it). eg:
```python 
def _handle_missing_bound(bound_value, expected_input_length, replacement_value): 
	if bound_value is not None: 
    	    return at.as_tensor_variable(floatX(bound_value)) 
	elif expected_input_length == 0: 
    	    return at.as_tensor_variable(floatX(replacement_value)) 
	else: 
    	    return at.repeat(at.as_tensor_variable(replacement_value), at.as_tensor_variable(expected_input_length))

```

If on the other hand I use at,switch I end up always having an array as output (an empy one when mu_shape is empty, ie the input is a scalar)
```python
switch(at.gt(mu_shape.size, at.as_tensor_variable(0)), at.repeat(at.as_tensor_variable(floatX(replacement_value)), mu_shape.size), replacement_value)
```

Alternatively I can use at.repeat directly, but again I end up always having an array as output
```python
repeat = at.switch(at.gt(mu_shape.size, at.as_tensor_variable(0)), mu_shape.shape[0], 1)
bound_value = at.repeat(at.as_tensor_variable(floatX(replacement_value)), repeat).eval()
```
None of the cases above seems a serious solution to me, at the same time I cannot think a decent symbolic approach supporting both scalars and arrays. I have fairly little Theano knowledge, so there may be an obvious solution to this

btw I am also going to add a check in `test_distributions_random` checking that the parametrization is as expected when symbolic inputs are passed to `dist`

pymc3/distributions/continuous.py
```diff
@@ -632,99 +673,63 @@ class TruncatedNormal(BoundedContinuous):
...
-        """
-        Draw random values from TruncatedNormal distribution.
+        if lower is None and upper is None:
+            initval = mu
+        elif lower is None and upper is not None:
+            initval = upper - 1.0
+        elif lower is not None and upper is None:
+            initval = lower + 1.0
+        else:
+            initval = (lower + upper) / 2
```
ricardoV94: This can be commented out for now

pymc3/distributions/continuous.py
```diff
@@ -738,38 +743,30 @@ def logp(self, value):
         -------
         TensorVariable
         """
-        mu = self.mu
-        sigma = self.sigma
-
-        norm = self._normalization()
-        logp = Normal.dist(mu=mu, sigma=sigma).logp(value) - norm
+        unbounded_lower = isinstance(lower, TensorConstant) and lower.value == -np.inf
+        unbounded_upper = isinstance(upper, TensorConstant) and upper.value == np.inf
```
ricardoV94: Needs the same logic as above with the `np.all`

pymc3/distributions/continuous.py
ricardoV94: I might have missed / forgot it, but why can't we pass a simple constant `-np.inf`. The RandomVariable should be able to broadcast it to a vector when needed.

ricardoV94:
I pushed some changes, I think the logic works fine with a scalar `-+np.inf` instead of `lower` / `upper`. I also rebased on `main` and fixed some merge conflicts that emerged. Don't forget to rebase from here before pushing new changes!

pymc3/distributions/continuous.py
```diff
@@ -738,40 +729,31 @@ def logp(self, value):
...
+            norm = 0.0
+
+        logp = Normal.logp(value, mu=mu, sigma=sigma) - norm
+        bounds = []
+        if not unbounded_lower:
+            bounds.append(value >= lower)
+        if not unbounded_upper:
+            bounds.append(value <= upper)
+        if not unbounded_lower and not unbounded_upper:
+            bounds.append(lower <= upper)
```
ricardoV94, line 755: This logic can be placed in the `if else` branches above, no point in double checking

pymc3/tests/test_distributions.py
```diff
@@ -2714,6 +2725,104 @@ def test_bound():
...
+            lower_interval,
+            upper_interval,
+        ) = self.get_dist_params_and_interval_bounds(model, bounded_rv_name)
+
+        assert lower.value == -1
+        assert np.array_equal(upper.value, [np.inf, np.inf])
+        assert lower_interval.value == -1
+        assert upper_interval is None
+
+    def test_hierarchical(self):
```
ricardoV94, line 2810: I think we can do away with this test, doesn't seem to offer anything new

pymc3/tests/test_distributions.py
```diff
@@ -2714,6 +2725,104 @@ def test_bound():
...
+            lower_interval,
+            upper_interval,
+        ) = self.get_dist_params_and_interval_bounds(model, bounded_rv_name)
+
+        assert lower.value == -1
+        assert np.array_equal(upper.value, [np.inf, np.inf])
+        assert lower_interval.value == -1
+        assert upper_interval is None
+
+    def test_rich_context(self):
```
ricardoV94: I think we can remove this one. Does not seem to test anything extra

ricardoV94:
Seeing a weird warning in the failing jobs: https://github.com/pymc-devs/pymc3/pull/4711/checks?check_run_id=2950387511#step:7:2342

>  /usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192 from C header, got 216 from PyObject


Which seems to be related to the failure just before: https://github.com/pymc-devs/pymc3/pull/4711/checks?check_run_id=2951980417#step:7:2332

```python
          if self.ndim != data.ndim:
            raise TypeError(
>               f"Wrong number of dimensions: expected {self.ndim},"
                f" got {data.ndim} with shape {data.shape}."
            )
E           TypeError: Wrong number of dimensions: expected 0, got 1 with shape (1,).
```

pymc3/distributions/continuous.py
matteo-pallini: Thanks for stepping in and improving it.

> The RandomVariable should be able to broadcast it to a vector when needed.

Oh! I didn't know about it and I thought it was necessary to handle it in`.dist`. That's the reason why I wanted to use `np.repeat` /`at.repeat`

pymc3/tests/test_distributions.py
matteo-pallini: The error trace you shared below is coming from that test, Originally I left it because I was expecting it to pass and it didn't, so I wanted to investigate why. Judging by [this thread](https://stackoverflow.com/questions/40845304/runtimewarning-numpy-dtype-size-changed-may-indicate-binary-incompatibility) I would say that the warning is unrelated.

The error seems to be coming from `pymc3.model.Model.set_initval`. When running the Truncated with lower=None there is a mismatch in the number of dimensions between the `rv_value_var` (which is a scalar) and the `initval` (which is an array). The second is generated by `initval_fn(), but its ndim is due to `transform.forward(rv_var, value)`, I think. So, it may be possible that the interval is returning a wrong value due to the refactoring. But I haven't been able to replicate the issue with simpler tests, so I may be simply wrong. I may have written the test incorrectly, or there may be a bug. I think it's worth keeping the test at least until we cannot make it pass. For now I removed it though

ricardoV94: I also saw that thread about not being harmful... but it's suspicious that the warning appears just after the failure, which comes from an Aesara check that ndims did not change and the warning does not appear in the other successful runs of the same test.

ricardoV94:
@drabbit17 check my [last commit](https://github.com/pymc-devs/pymc3/pull/4711/commits/26a442d5cab641ab0b82e673434eedc0e04b461e). I removed some tests (or `tests_to_run`) that felt unnecessary. Let me know if you disagree. Otherwise I think this PR is ready to merge.

I tested creating a bunch of TruncatedNormals with random sizes and lower/ upper parameters in a single model and I did not find any issues with the `initval` like the one we were getting in that `rich_context` model. So I am pretty confident that it was not an issue on our side.

matteo-pallini:
> Otherwise I think this PR is ready to merge

Agreed, I did a very small change. sorry for dragging the PR for so long and thanks for the support (it would have been way faster for you to simply do the whole thing yourself :-) ).

> So I am pretty confident that it was not an issue on our side.

our side as this PR or PyMC?

I have been trying to replicate the test failure for `pymc3/tests/test_distributions_random.py::TestNestedRandom::test_TruncatedNormal` locally but I haven't been able to, can you? (the same goes for the `rich_context` test). Are the github tests being ran through a docker container? if so, is it possible to download it?

ricardoV94:
> Agreed, I did a very small change. sorry for dragging the PR for so long and thanks for the support (it would have been way faster for you to simply do the whole thing yourself :-) ).

I disagree. The dynamic interval thing is something that we needed to figure out, and will be used in other places as well. It was really great that you dived in and started figuring it out.

> > So I am pretty confident that it was not an issue on our side.
> 
> our side as this PR or PyMC?

Both. I am pretty confident it was an issue with incompatible numpy / aesara binaries, that emerged on that specific environment.

> I have been trying to replicate the test failure for `pymc3/tests/test_distributions_random.py::TestNestedRandom::test_TruncatedNormal` locally but I haven't been able to, can you? (the same goes for the `rich_context` test). Are the github tests being ran through a docker container? if so, is it possible to download it?

The `TestNestedRandom` seems like a weird caching issue, because we recently made tests marked as xfail strict (the tests fail if the test pass) and in this case it seems to still be surprised that the tests are passing. https://github.com/pymc-devs/pymc3/pull/4711/checks?check_run_id=2973646757#step:8:682

I can't find a explicit error message in the logs related to the "failing tests" (@michaelosthege any ideas?)

michaelosthege:
I added the `mark.xfail` on `test_TruncatedNormal` because the test used a nonexistent API, but the distribution was not yet refactored.
This PR refactored the distribution, so now the test is passing.
👉 Just remove the xfail.

michaelosthege:
Rebased and removed that `mark.xfail`. Should go green now, but please check if my rebase didn't mess up anything :)

ricardoV94:
> I added the `mark.xfail` on `test_TruncatedNormal` because the test used a nonexistent API, but the distribution was not yet refactored.
> This PR refactored the distribution, so now the test is passing.
> 👉 Just remove the xfail.


I was convinced that I had removed it xD

michaelosthege:
@ricardoV94 the rebase was tricky too. You probably ended up with another tests `mark.xfail` earlier.

Both test failures seem to be related to the known problem with the `find_MAP` #4771.
Merge away?

ricardoV94:
> @ricardoV94 the rebase was tricky too. You probably ended up with another tests `mark.xfail` earlier.
> 
> Both test failures seem to be related to the known problem with the `find_MAP` #4771.
> Merge away?

The MLE one will be adjusted in #4833. ~~Let me check the other one quickly~~ Yeah the second one looks another MAP issue. Does not seem related to any changes in this PR

ricardoV94 (approved):

ricardoV94:
Great work @drabbit17 This was a fun one to crack. Looking forward to your next PR :)

Issue #4676, opened 2021-05-03
Small Dataset pulled when you call linear_training_data()


drachimera:
This is a request from the DEVS on the discourse channel:

On the homepage there is this function:
X, y = linear_training_data()
when you copy and paste it just to see if you have a working install, it bombs. Perhaps we could put a very small test dataset in the install so a function like this works out of the box?

When I was installing and trying to see if the install worked, the example didn't work because there is no data when you call linear_training_data().  

This issue would resolve that so that if someone copies and pastes the first example they see, it works if they have a working install.

Full discussion is here: https://discourse.pymc.io/t/attributeerror-module-arviz-has-no-attribute-geweke/6818/27

farhanreynaldo:
Hi, what if we create simple fake data for the `linear_training_data()`? I would suggest something like this:

```python
import numpy as np

np.random.seed(12)

def linear_training_data():
    X = np.random.normal(size=100)
    y = 1.2 * X + np.random.normal(size=100)
    return X, y
```

But I am not quite sure where to put this function. Or, we might as well change the homepage data snippet with existing PyMC3 examples, such as [milk data](https://github.com/pymc-devs/pymc-examples/blob/main/examples/data/milk.csv).

michaelosthege:
closed by #4713

PR #4713, opened 2021-05-24
Add linear training data to documentation

robertdefilippi:
Hello,

Was looking for a simple PR to start contributing to the project, and I hope I did everything correctly. If something is not clear or I did something wrong, please let me know. I would like to be able to contribute correctly to this project.

Original issue is: #4676 

This will add function to generate linear data.

None. This should only add functionality to the docs, to allow them to quickly get up and running without having to import any data.
#4676
None.
All run 👍 
Updated the core docs to reflect the addition of the new function to generate the linear data.
I'm not sure what to say here, as it is my first contribution to the project.

Thanks!

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -21,7 +21,7 @@
 
     import pymc3 as pm
 
-    X, y = pm._linear_training_data()
+    X, y = pm.math._linear_training_data()
     with pm.Model() as linear_model:
         weights = pm.Normal("weights", mu=0, sigma=1)
         noise = pm.Gamma("noise", alpha=2, beta=1)
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -22,6 +22,7 @@
     import pymc3 as pm
 
     X, y = pm.math._linear_training_data()
+
     with pm.Model() as linear_model:
         weights = pm.Normal("weights", mu=0, sigma=1)
         noise = pm.Gamma("noise", alpha=2, beta=1)
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -21,7 +21,7 @@
 
     import pymc3 as pm
 
-    X, y = linear_training_data()
+    X, y = pm._linear_training_data()
     with pm.Model() as linear_model:
         weights = pm.Normal("weights", mu=0, sigma=1)
         noise = pm.Gamma("noise", alpha=2, beta=1)
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -21,7 +21,8 @@
 
     import pymc3 as pm
 
-    X, y = pm.math._linear_training_data()
+    X = np.random.normal(size=100)
+    y = np.random.normal(X) * 1.2
 
     with pm.Model() as linear_model:
         weights = pm.Normal("weights", mu=0, sigma=1)
--- a/pymc3/math.py
+++ b/pymc3/math.py
@@ -489,8 +489,8 @@ def _linear_training_data():
         Array of output values
 
     """
-    np.random.seed(1234)
-    X = np.random.normal(size=100)
-    y = 1.2 * X + np.random.normal(size=100)
+    rng = np.random.default_rng(1234)
+    X = rng.standard_normal(100)
+    y = 1.2 * X + rng.standard_normal(100)
 
     return X, y
--- a/pymc3/math.py
+++ b/pymc3/math.py
@@ -476,12 +476,9 @@ def block_diagonal(matrices, sparse=False, format="csr"):
     return BlockDiagonalMatrix(sparse=sparse, format=format)(*matrices)
 
 
-def _linear_training_data(size=100):
+def _linear_training_data():
     r"""Create a simple linear model training dataset
-
-    Parameters
-    ----------
-    size: int (default 100) number of observations
+    for use in the documentation
 
     Returns
     -------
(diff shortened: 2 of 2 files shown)


pymc3/math.py
```diff
@@ -474,3 +474,26 @@ def block_diagonal(matrices, sparse=False, format="csr"):
     if len(matrices) == 1:  # graph optimization
         return matrices[0]
     return BlockDiagonalMatrix(sparse=sparse, format=format)(*matrices)
+
+
+def _linear_training_data(size=100):
```
ricardoV94: No need to have size as an argument, this function should only be used for that example. Also I would mention in the docstring that this is exactly the purpose of the function.

ricardoV94:
Thanks for opening the PR. Indeed it seems useful to have the example snippet run if copied directly. 

Alternatively, our example could have an explicit `numpy` import and include the generating code in it.

robertdefilippi:
> Thanks for opening the PR. Indeed it seems useful to have the example snippet run if copied directly.
> 
> Alternatively, our example could have an explicit `numpy` import and include the generating code in it.

No worries! I wanted to get into contributing to open source, and this seemed like a good first little project. I really enjoy working with `pymc3` and want to see how I can give back in a small way.

I've updated the function with your suggestions and I hope that works 😄

pymc3/math.py
```diff
@@ -474,3 +474,23 @@ def block_diagonal(matrices, sparse=False, format="csr"):
...
+    Returns
+    -------
+    X: np.ndarray()
+        Array of input values
+
+    y: np.ndarray()
+        Array of output values
+
+    """
+    np.random.seed(1234)
```
OriolAbril: this should use the new generator, otherwise it will globally set the seed of numpy.random behind the scenes and users won't get random behaviour when they expect it to. https://numpy.org/doc/stable/reference/random/index.html#quick-start

robertdefilippi: Done! I had no idea that was update. Thanks for letting me know

michaelosthege:
I just rebased your branch on the new `main`. This should fix the CI.

michaelosthege (changes requested):
No idea why that one test in the CI failed, but regarding the changes here: Instead of adding an unrelated & trivial function to `pm.math` we can just add the two lines for creating the linear dataset right into the docstring, no?
```
X = np.random.normal(size=100)
y = 1.2 * np.random.normal(X)
```

ricardoV94:
Only inelegance of that is that a numpy import would be needed in the example as well

docs/source/index.rst
```diff
@@ -21,7 +21,8 @@
 
     import pymc3 as pm
 
-    X, y = linear_training_data()
+    X, y = pm.math._linear_training_data()
+
```
michaelosthege: I don't think we need to import numpy. Whoever makes it to this point can infer that by themselves.
```suggestion
    X = np.random.normal(size=100)
    y = np.random.normal(X) * 1.2

```

pymc3/math.py
```diff
@@ -474,3 +474,23 @@ def block_diagonal(matrices, sparse=False, format="csr"):
...
+
+    y: np.ndarray()
+        Array of output values
+
+    """
+    rng = np.random.default_rng(1234)
+    X = rng.standard_normal(100)
+    y = 1.2 * X + rng.standard_normal(100)
+
+    return X, y
```
michaelosthege:
```suggestion
```

michaelosthege (changes requested):
@robertdefilippi just as an explanation, since you mentioned that you want to start contributing to open source:
There's nothing wrong with the function you added, but as maintainers we have an interest to keep the codebase as concise as possible. More code == more things to maintain.
While we do have a submodule `pm.data` to fetch example datasets, those datasets are much bigger than the one needed for this docstring example.

michaelosthege (approved):

docs/source/index.rst
```diff
@@ -21,7 +21,9 @@
 
     import pymc3 as pm
 
-    X, y = linear_training_data()
+    X = np.random.normal(size=100)
```
ricardoV94, line 24: Should we also add `import numpy as np`, or leave it just like this?

michaelosthege: I'd leave it out, because whoever ends up most probably knows what `np` is already

michaelosthege: Also you can't make code suggestions on unchanged lines, so someone would have to checkout out the branch and do it locally.
I'm too lazy for that.

robertdefilippi:
> @robertdefilippi just as an explanation, since you mentioned that you want to start contributing to open source:
> There's nothing wrong with the function you added, but as maintainers we have an interest to keep the codebase as concise as possible. More code == more things to maintain.
> While we do have a submodule `pm.data` to fetch example datasets, those datasets are much bigger than the one needed for this docstring example.

Hey @michaelosthege that's perfectly understandable. I thought adding this change would be a simple change, but it seems to have added a bunch of additional overhead to a lot of people. Maybe it would be best that this PR is just closed, and I can try my hand at another part of the project. I would not want this small addition to burden any part of the project.

michaelosthege:
No worries. I merged it anyways. There are many more [constructions sites](https://github.com/pymc-devs/pymc3/issues?q=is%3Aopen+is%3Aissue+label%3A%22beginner+friendly%22) 😅

PR #4723, opened 2021-05-26
Refactor Flat and HalfFlat distributions

ricardoV94:
Replaced `ValueError` by `NotImplementedError`. 

Also cleaned some tests that had wrong `pytest.marks` or were redundant.

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -308,31 +308,36 @@ class Uniform(BoundedContinuous):
         )
 
 
+class FlatRV(RandomVariable):
+    name = "flat"
+    ndim_supp = 0
+    ndims_params = []
+    dtype = "floatX"
+    _print_name = ("Flat", "\\operatorname{Flat}")
+
+    @classmethod
+    def rng_fn(cls, rng, size):
+        raise NotImplementedError("Cannot sample from flat variable")
+
+
+flat = FlatRV()
+
+
 class Flat(Continuous):
     """
     Uninformative log-likelihood that returns 0 regardless of
     the passed value.
     """
 
-    def __init__(self, *args, **kwargs):
-        self._default = 0
-        super().__init__(defaults=("_default",), *args, **kwargs)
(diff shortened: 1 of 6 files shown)


pymc3/distributions/continuous.py
```diff
@@ -308,31 +308,36 @@ def logcdf(value, lower, upper):
...
 
-        Raises
-        -------
-        ValueError
-        """
-        raise ValueError("Cannot sample from Flat distribution")
+    @classmethod
+    def dist(cls, *, size=None, testval=None, **kwargs):
+        if testval is None:
+            testval = np.full(size, floatX(0.0))
```
brandonwillard, line 337: I'm assuming these test values were necessary because, when `aesara.config.compute_test_value` is enabled, `RandomVariable.perform` is called in order to generate a test value automatically, and it will fail.

In general, test value-related logic like this needs to be conditioned on `aesara.config.compute_test_value != "off"`; otherwise, we'll end up doing unnecessary work.  This case might be different, though.

brandonwillard (approved):
I still question the purpose/place of these `Flat` distributions (and, now, `RandomVariable`s); however, this PR looks fine.

At some point, we need to revisit these and determine whether or not they're still needed or relevant given the new framework (or a potentially newer one).

pymc3/distributions/continuous.py
ricardoV94: The `Flat` testval is critical (in the codebase) in a couple of places (e.g., for getting the model.initial_point for sampling).

brandonwillard: Just keep in mind that `v4` cannot actually *require* test values, and anywhere that currently does is a place we need to refactor.  

Aside from that, it should always be possible to enable test values and have them work throughout `v4`, it's just not *mandatory* or automatically enabled at any point within PyMC.

ricardoV94: Definitely, I am adding a note about this in #4567

PR #4726, opened 2021-05-27
Update version string and add warning on import.

twiecki:


pymc3/__init__.py
```diff
@@ -27,6 +27,10 @@
         handler = logging.StreamHandler()
         _log.addHandler(handler)
 
+_log.info(
+    "You are running the development version of PyMC3 which currently still lacks key features."
```
ricardoV94: Perhaps add something along the lines of:

> The latest stable release is version `3.11.2`.

brandonwillard:
Side note: have we considered setting up `versioneer`?

https://github.com/pymc-devs/pymc3/issues/4727

brandonwillard (approved):
These changes look fine, but I don't know what the failures are about, yet.

brandonwillard (changes requested):
Ah, looks like [the error](https://github.com/pymc-devs/pymc3/pull/4726/checks?check_run_id=2683532131#step:7:311) is due to the version string parsing in Arviz.

I created an issue: https://github.com/arviz-devs/arviz/issues/1708.

pymc3/__init__.py
```diff
@@ -13,7 +13,7 @@
 #   limitations under the License.
 
 # pylint: disable=wildcard-import
-__version__ = "3.11.1"
+__version__ = "4.0.dev0"
```
twiecki:
```suggestion
__version__ = "4.0"
```

PR #4718, opened 2021-05-25
Fix `change_rv_size` bug

brandonwillard:
This PR updates Aesara to 2.0.9 and fixes a potential `change_rv_size` bug.

PR #4729, opened 2021-05-28
Use an incrementally updated RNG state in `Model`

### LINKED ISSUE #4728: Make sure `Model` RVs are distinct via their RNGs
Currently, all the Aesara `RandomVariable` `Op`s are being converted into in-place `Op`s (i.e. when a graph containing one of them is compiled and evaluated, the underlying shared variable RNG state is updated in place).  This can be a mild convenience&mdash;and a potential efficiency gain&mdash;but it's not a good use of the `RandomVariable` API, because it can lead to confusion.

For example,
```python
import numpy as np

import aesara
import aesara.tensor as at

import pymc3 as pm


with pm.Model() as test_model:
    X_rv = pm.Normal("x")
    Y_rv = pm.Normal("y")


test_fn = test_model.fn(Y_rv + 2 * X_rv)

aesara.dprint(test_fn.f)
# Elemwise{Composite{(i0 + (i1 * i0))}}[(0, 0)] [id A] ''   1
#  |normal_rv.1 [id B] 'y'   0
#  | |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F1A343DE160>) [id C]
#  | |TensorConstant{[]} [id D]
#  | |TensorConstant{11} [id E]
#  | |TensorConstant{0} [id F]
#  | |TensorConstant{1.0} [id G]
#  |TensorConstant{2.0} [id H]
# RandomStateSharedVariable(<RandomState(MT19937) at 0x7F1A343DE160>) [id C]
```
The compiled graph contains only **one** `RandomVariable`.

The merge optimizations removed the second `RandomVariable` because it was identical to the first, and we need only produce the same samples once.

If we want to make sure that Aesara knows these two `RandomVariable`s are distinct terms, we can provide a distinct RNG state for each:
```python
with pm.Model() as test_model:
    X_rv = pm.Normal("x")

    rng = aesara.shared(np.random.RandomState(2023532), borrow=True)
    # or
    # rng = X_rv.owner.outputs[0]

    Y_rv = pm.Normal("y", rng=rng)

test_fn = test_model.fn(Y_rv + 2 * X_rv)

aesara.dprint(test_fn.f)
# Elemwise{Composite{(i0 + (i1 * i2))}}[(0, 0)] [id A] ''   2
#  |normal_rv.1 [id B] 'y'   1
#  | |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F1A343DE490>) [id C]
#  | |TensorConstant{[]} [id D]
#  | |TensorConstant{11} [id E]
#  | |TensorConstant{0} [id F]
#  | |TensorConstant{1.0} [id G]
#  |TensorConstant{2.0} [id H]
#  |normal_rv.1 [id I] 'x'   0
#    |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F1A343DE5A0>) [id J]
#    |TensorConstant{[]} [id D]
#    |TensorConstant{11} [id E]
#    |TensorConstant{0} [id F]
#    |TensorConstant{1.0} [id G]
# RandomStateSharedVariable(<RandomState(MT19937) at 0x7F1A343DE5A0>) [id J]
```

In other words, we shouldn't rely on in-place `RandonVariable` `Op`s, and we should use explicitly updated `RandomState`s.  This doesn't prevent us from using in-place updates for efficiency, especially since there's already an optimization that converts non-in-place `RandomVariable`s to in-place ones.

Finally, we can automatically make sure that each `RandomVariable` created within a `Model` context is distinct by simply updating `Model.default_rng` after each `RandomVariable` is created in `Distribution.__new__` (i.e. after [this step](https://github.com/pymc-devs/pymc3/blob/v4/pymc3/distributions/distribution.py#L156)).

brandonwillard:
This PR updates the role of `Model.default_rng` by adding a `Model.next_rng` that is incrementally updated when variables are registered.

Under this change, `Model.default_rng` still serves as a mutable starting point for RNG seeding, while also (softly) guaranteeing that newly added/registered random variables&mdash;that don't specify their own `rng` parameters&mdash;will be unique.

This PR also updates how initial values are specified, stored, and generated.  Since computing initial values may require sampling, these updates were needed in order to prevent initial values from unnecessarily affecting the RNG states in confusing ways.

Closes #4728

### DIFF
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -277,14 +277,14 @@ class NoDistribution(Distribution):
         self,
         shape,
         dtype,
-        testval=None,
+        initval=None,
         defaults=(),
         parent_dist=None,
         *args,
         **kwargs,
     ):
         super().__init__(
-            shape=shape, dtype=dtype, testval=testval, defaults=defaults, *args, **kwargs
+            shape=shape, dtype=dtype, initval=initval, defaults=defaults, *args, **kwargs
         )
         self.parent_dist = parent_dist
 
@@ -342,7 +342,7 @@ class DensityDist(Distribution):
         logp,
         shape=(),
         dtype=None,
-        testval=0,
+        initval=0,
         random=None,
         wrap_random_with_dist_shape=True,
         check_shape_in_random=True,
@@ -363,8 +363,8 @@ class DensityDist(Distribution):
             a value here.
         dtype: None, str (Optional)
             The dtype of the distribution.
-        testval: number or array (Optional)
-            The ``testval`` of the RV's tensor that follow the ``DensityDist``
+        initval: number or array (Optional)
+            The ``initval`` of the RV's tensor that follow the ``DensityDist``
             distribution.
         args, kwargs: (Optional)
             These are passed to the parent class' ``__init__``.
@@ -400,7 +400,7 @@ class DensityDist(Distribution):
         """
         if dtype is None:
             dtype = aesara.config.floatX
-        super().__init__(shape, dtype, testval, *args, **kwargs)
+        super().__init__(shape, dtype, initval, *args, **kwargs)
         self.logp = logp
         if type(self.logp) == types.MethodType:
             if PLATFORM != "linux":
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -19,12 +19,12 @@ import types
 import warnings
 
 from abc import ABCMeta
-from copy import copy
 from typing import TYPE_CHECKING
 
 import dill
 
 from aesara.tensor.random.op import RandomVariable
+from aesara.tensor.random.var import RandomStateSharedVariable
 
 from pymc3.distributions import _logcdf, _logp
 
@@ -77,14 +77,6 @@ class DistributionMeta(ABCMeta):
         rv_type = None
 
         if isinstance(rv_op, RandomVariable):
-            if not rv_op.inplace:
-                # TODO: This is a temporary work-around.
-                # Remove this once we know what we want regarding RNG states
-                # and their propagation.
-                rv_op = copy(rv_op)
-                rv_op.inplace = True
-                clsdict["rv_op"] = rv_op
-
             rv_type = type(rv_op)
 
         new_cls = super().__new__(cls, name, bases, clsdict)
@@ -158,14 +150,30 @@ class Distribution(metaclass=DistributionMeta):
         return model.register_rv(rv_out, name, data, total_size, dims=dims, transform=transform)
 
     @classmethod
-    def dist(cls, dist_params, **kwargs):
+    def dist(cls, dist_params, rng=None, **kwargs):
 
         testval = kwargs.pop("testval", None)
 
-        rv_var = cls.rv_op(*dist_params, **kwargs)
+        rv_var = cls.rv_op(*dist_params, rng=rng, **kwargs)
 
         if testval is not None:
             rv_var.tag.test_value = testval
+
+        if (
+            rv_var.owner
+            and isinstance(rv_var.owner.op, RandomVariable)
+            and isinstance(rng, RandomStateSharedVariable)
+            and not getattr(rng, "default_update", None)
+        ):
+            # This tells `aesara.function` that the shared RNG variable
+            # is mutable, which--in turn--tells the `FunctionGraph`
+            # `Supervisor` feature to allow in-place updates on the variable.
+            # Without it, the `RandomVariable`s could not be optimized to allow
+            # in-place RNG updates, forcing all sample results from compiled
+            # functions to be the same on repeated evaluations.
+            new_rng = rv_var.owner.outputs[0]
+            rv_var.update = (rng, new_rng)
+            rng.default_update = new_rng
 
         return rv_var
 
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -143,21 +143,44 @@ class Distribution(metaclass=DistributionMeta):
         if "shape" in kwargs:
             raise DeprecationWarning("The `shape` keyword is deprecated; use `size`.")
 
+        testval = kwargs.pop("testval", None)
+
+        if testval is not None:
+            warnings.warn(
+                "The `testval` argument is deprecated; use `initval`.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
+        initval = kwargs.pop("initval", testval)
+
         transform = kwargs.pop("transform", UNSET)
 
         rv_out = cls.dist(*args, rng=rng, **kwargs)
 
-        return model.register_rv(rv_out, name, data, total_size, dims=dims, transform=transform)
+        if testval is not None:
+            rv_out.tag.test_value = testval
+
+        return model.register_rv(
+            rv_out, name, data, total_size, dims=dims, transform=transform, initval=initval
+        )
 
     @classmethod
     def dist(cls, dist_params, rng=None, **kwargs):
 
         testval = kwargs.pop("testval", None)
 
-        rv_var = cls.rv_op(*dist_params, rng=rng, **kwargs)
-
         if testval is not None:
-            rv_var.tag.test_value = testval
+            warnings.warn(
+                "The `testval` argument is deprecated. "
+                "Use `initval` to set initial values for a `Model`; "
+                "otherwise, set test values on Aesara parameters explicitly "
+                "when attempting to use Aesara's test value debugging features.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
+        rv_var = cls.rv_op(*dist_params, rng=rng, **kwargs)
 
         if (
             rv_var.owner
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -137,7 +137,7 @@ class Distribution(metaclass=DistributionMeta):
         rng = kwargs.pop("rng", None)
 
         if rng is None:
-            rng = model.default_rng
+            rng = model.next_rng()
 
         if not isinstance(name, string_types):
             raise TypeError(f"Name needs to be a string but got: {name}")
--- a/pymc3/model.py
+++ b/pymc3/model.py
@@ -49,6 +49,7 @@ from pandas import Series
 
 from pymc3.aesaraf import (
     change_rv_size,
+    compile_rv_inplace,
     gradient,
     hessian,
     inputvars,
@@ -455,7 +456,7 @@ class ValueGradFunction:
 
         inputs = grad_vars
 
-        self._aesara_function = aesara.function(inputs, outputs, givens=givens, **kwargs)
+        self._aesara_function = compile_rv_inplace(inputs, outputs, givens=givens, **kwargs)
 
     def set_weights(self, values):
         if values.shape != (self._n_costs - 1,):
@@ -1378,7 +1379,7 @@ class Model(Factor, WithMemoization, metaclass=ContextMeta):
         Compiled Aesara function
         """
         with self:
-            return aesara.function(
+            return compile_rv_inplace(
                 self.value_vars,
                 outs,
                 allow_input_downcast=True,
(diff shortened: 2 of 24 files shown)


ricardoV94:
I guess this also affects variables created via `.dist()`? We need to keep that in mind until we decide what to do with that API.

ricardoV94:
Are successive `rng`s guaranteed to not repeat or just nearly so?

brandonwillard:
> I guess this also affects variables created via `.dist()`?

Howso?

ricardoV94:
NVM, I thought they were also given the model rng

michaelosthege:
If we keep the uncoordinated merge of #4718 it forces me to rebase the #4696 branch--unless someone else steps in to help out.
Possibly more because of emotional than technical reasons I am not at all looking forward to do this rebase.

Call me overreacting, but out of respect for a fellow core contributor please don't merge anything into `v4` until we have discussed how to proceed as a team.

brandonwillard:
> Call me overreacting, but out of respect for a fellow core contributor please don't merge anything into `v4` until we have discussed how to proceed as a team.

Out of respect for the core developers and anyone who might be interested in using `v4`, I would hope that the core developers would prioritize self-contained bug fixes like this and #4718 over large, non-critical features such as #4696.

Regardless, I won't merge this until others weigh in.

twiecki:
I don't remember that we've locked branches (including `master`) before for other PRs, so this would be something new. We can certainly discuss whether this is a good idea, I personally don't think it is as it would slow us down tremendously if a single PR that we have difficulty merging brings development to a complete halt.

We do have tools to resolve this, namely a rebase, and I don't see how it would be a particularly difficult one in this case. Plus, we will have to rebase #4696 anyway. I'm not at all saying that's on you @michaelosthege as you said before that the PR is done from your side, and that's totally fine.

twiecki:
@brandonwillard So changing the `rng` is only to make `RV`s unique? I guess I'm a bit confused why they aren't unique to begin with and if we shouldn't attack the problem there?

brandonwillard:
> @brandonwillard So changing the `rng` is only to make `RV`s unique? I guess I'm a bit confused why they aren't unique to begin with and if we shouldn't attack the problem there?

No, not exclusively; changing the `rng` for each variable is also generally more consistent (e.g. it doesn't force one to use in-place `Op`s and it provides at least the same level of configurability/control).

Otherwise, the uniqueness in question is consistent and fundamental to general graph representations, and also how Aesara is designed: if the same `Op` is applied to the same inputs, the resulting nodes and their outputs are equivalent.  This is the basis of "merge" optimizations, which prevent graphs from computing the same things more than once.

We could make `RandomVariable` nodes unique in other ways (e.g. by adding new inputs that could be used to uniquely identify them), but the points above would remain, and, because the `rng` solution is the design-intended approach&mdash;and addresses other issues as well&mdash;there's no need to invent anything new (that doesn't entail other improvements, of course).


Aside from all that, this lack of uniqueness can result in real **bugs**, because a compiled function that conflates equivalent `RandomVariable`s&mdash;like the `x` and `y` in the example for #4728&mdash;will drop one of the arguments (definitely when `on_unused_input="ignore"` is set), so a user attempting to provide such a function with distinct values for each variable will find that the results are not correct.

pymc3/tests/test_distributions.py
```diff
@@ -2949,3 +2949,16 @@ def func(x):
     import pickle
 
     pickle.loads(pickle.dumps(y))
+
+
+def test_distinct_rvs():
```
ricardoV94, line 2954: This test does not seem to fail in my local V4 branch

ricardoV94 (dismissed):

brandonwillard:
If we changed this so that it worked more like `RandomStream` (i.e. generate a new `RandomState` for each value of `model.next_rng`), we could avoid producing very inter-connected graphs.  This might be the better approach.

Update: it's definitely the better approach; I'll rebase this with the changes soon.

brandonwillard:
The changes I just pushed allow `Model` to generate new RNG states like `RandomStream` does, as mentioned previously.

The `Model` object will keep a list of the RNG states it has generated, in case we want to allow easier model re-seeding.

Also, the forced in-place `RandomVariable` `Op` creation has been removed and replaced with the intended optimization-based approach.  In other words, when `aesara.function` is used with a mode that includes the `RandomVariable` in-place optimization step, it will convert each `RandomVariable` `Op` in the graph into its in-place equivalent, and allow evaluations of a compiled function to update its graph's RNG states (i.e. produce new samples each time).

The standard `FAST_RUN` mode includes this optimization, but, to be safe, I've created a `pymc3.aesaraf.compile_rv_inplace` function that will add this particular in-place optimization to an existing mode.  PyMC will use this helper function wherever it needs to compile a graph, ensuring that the result produces distinct samples.

pymc3/model.py
```diff
@@ -931,6 +954,20 @@ def cont_vars(self):
         """All the continuous variables in the model"""
         return list(typefilter(self.value_vars, continuous_types))
 
+    def next_rng(self) -> RandomStateSharedVariable:
+        """Generate a new ``RandomStateSharedVariable``.
+
+        The new ``RandomStateSharedVariable`` is also added to
+        ``Model.rng_seq``.
+        """
+        new_seed = self.rng_seeder.randint(2 ** 30, dtype=np.int64)
```
ricardoV94, line 967: Would the original problem still happen (even if very rarely) if a RV has the same inputs and the same seed by chance?

Even if very rare should we check it's not on the seed list to make sure no seeds are duplicated? Or it doesn't matter?

brandonwillard:
The current failure is pretty straightforward: `Model.initial_point` uses `Variable.eval`, which compiles a function in `FAST_RUN` mode and alters the RNG state in-place.  Since `Model.initial_point` is implicitly called by [`pm.load_trace`](https://github.com/pymc-devs/pymc3/pull/4729/files#diff-415c0fe5b2ff370d5dcb6e4123609a9adc353fdc9059eee09bdb57b6e19893a2R278), the subsequent `pm.sample_posterior_predictive` call is starting at a different RNG state than the earlier one, and that causes the results to differ.

It's a pretty straightforward fix (i.e. don't use `Variable.eval`), but it's also a good time to clean up `initial_point` a little bit, so I'll do that as a means of fixing this.

pymc3/model.py
brandonwillard: With this approach, in the rare case that the same seed is produced, it should still result in a distinct shared state object and, thus, not compare equal.  That's probably worth confirming on the side, though, but not as a test.

ricardoV94: Got it

pymc3/tests/test_distributions.py
```diff
@@ -2944,3 +2967,25 @@ def func(x):
...
```
ricardoV94, line 2972: Does this make more sense in test model?

ricardoV94 (commented):
Should we test that rvs can still be merged when they are in fact repeated in the graph expression?

pymc3/model.py
brandonwillard: This example demonstrates how the same RNG states can be used by distinct shared RNG objects and preserve the identities of their corresponding `RandomVariable`s:
```python
import numpy as np

import aesara
import aesara.tensor as at


rng_state = aesara.shared(
    np.random.RandomState(np.random.MT19937(np.random.SeedSequence(1234)))
)

rng_state2 = aesara.shared(
    np.random.RandomState(np.random.MT19937(np.random.SeedSequence(1234)))
)


x = at.random.normal(rng=rng_state)
y = at.random.normal(rng=rng_state)
z = at.random.normal(rng=rng_state2)


fn = aesara.function([], x + y + z, mode="FAST_RUN")

aesara.dprint(fn.maker.fgraph)
# Elemwise{Add}[(0, 2)] [id A] ''   2
#  |normal_rv{0, (0, 0), floatX, False}.1 [id B] ''   1
#  | |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F6CD2960E20>) [id C]
#  | |TensorConstant{[]} [id D]
#  | |TensorConstant{11} [id E]
#  | |TensorConstant{0.0} [id F]
#  | |TensorConstant{1.0} [id G]
#  |normal_rv{0, (0, 0), floatX, False}.1 [id B] ''   1
#  |normal_rv{0, (0, 0), floatX, False}.1 [id H] ''   0
#    |RandomStateSharedVariable(<RandomState(MT19937) at 0x7F6CD27BC050>) [id I]
#    |TensorConstant{[]} [id D]
#    |TensorConstant{11} [id E]
#    |TensorConstant{0.0} [id F]
#    |TensorConstant{1.0} [id G]

# Let's confirm that these all produce the same thing, because
# they have the same RNGs...
fn2 = aesara.function([], [x, y, z], mode="FAST_COMPILE")

fn2()
# [array(-0.24304841), array(-0.24304841), array(-0.24304841)]
```

pymc3/tests/test_distributions.py
brandonwillard: Not so much; the relevant logic is actually implemented in `Distribution.dist`.

brandonwillard:
> Should we test that rvs can still be merged when they are in fact repeated in the graph expression?

No, that's very much Aesara's responsibility.

pymc3/model.py
ricardoV94: That's exactly what I was wondering. Great. 

And something like this wouldn't be useful having on Aesara test suite either?

ricardoV94 (dismissed):
LGTM

junpenglao:
`testval` -> `initval` is a breaking change that is not backward compatible, are we ok without a deprecation cycle here?

michaelosthege:
> `testval` -> `initval` is a breaking change that is not backward compatible, are we ok without a deprecation cycle here?

I agree that we should reconsider this breaking change. The value is still assigned to `.tag.test_value` and "`initval`" may confuse users that it has something to do with MCMC initialization.
If we decide to deprecate it, there should be a warning in the `.dist` API, similar to the warning added to the `__new__` API.

michaelosthege (dismissed):
(see comment about breaking change above)

pymc3/model.py
brandonwillard: No, not really.

brandonwillard:
> `testval` -> `initval` is a breaking change that is not backward compatible, are we ok without a deprecation cycle here?

What's a breaking change?  A deprecation _warning_ was added, but `testval` is still usable for now.

Regardless, folks, don't forget that this is **a new version**, and that there's no practical benefit to preserving a misnomer like `testval`.

pymc3/distributions/distribution.py
```diff
@@ -151,21 +143,48 @@ def __new__(cls, name, *args, **kwargs):
         if "shape" in kwargs:
             raise DeprecationWarning("The `shape` keyword is deprecated; use `size`.")
 
-        transform = kwargs.pop("transform", UNSET)
-
-        rv_out = cls.dist(*args, rng=rng, **kwargs)
+        testval = kwargs.pop("testval", None)
 
-        return model.register_rv(rv_out, name, data, total_size, dims=dims, transform=transform)
+        if testval:
```
twiecki:
```suggestion
        if testval is None:
```

twiecki:
I think the deprecation warning we have here is fine, but should definitely add a note in the release-notes on this change.

pymc3/distributions/distribution.py
```diff
@@ -151,21 +143,48 @@ def __new__(cls, name, *args, **kwargs):
         if "shape" in kwargs:
             raise DeprecationWarning("The `shape` keyword is deprecated; use `size`.")
 
-        transform = kwargs.pop("transform", UNSET)
-
-        rv_out = cls.dist(*args, rng=rng, **kwargs)
+        testval = kwargs.pop("testval", None)
 
-        return model.register_rv(rv_out, name, data, total_size, dims=dims, transform=transform)
+        if testval is None:
```
ricardoV94: Not `if testval is not None`?

brandonwillard: Already caught that one.

twiecki (dismissed):

brandonwillard:
> The value is still assigned to `.tag.test_value` and "`initval`" may confuse users that it has something to do with MCMC initialization.

The former confirms that this is _not_ a breaking change, and the latter wouldn't be a confusion, because `initval` **does** have something to do with MCMC initialization, just like the old test values.

twiecki (approved):

Issue #4552, opened 2021-03-17
Implement shape and dims for v4


twiecki:
@brandonwillard is proposing to remove the current `shape` kwarg and replace it with `size` which works more similar to the way `numpy` handles it. The difference is that `size` only specifies the batch dimension, while the event dimension is taken from the distribution. For univariate distributions, there is no difference, but for multivariate, there would be.

```
coords=dict(town=ABC, year=[2020, 2021])
X = MvNormal(cov=eye(3), shape=(2, 3))
Y = MvNormal(cov=eye(3), dims=("town", "year"))
Z = MvNormal(cov=eye(3) , size=2)
```

All three would lead to the same shaped `(2, 3)` MvNormal.

From @brandonwillard: "in v4, all the broadcasting and shape calculations for the arguments are already worked out so we always know what the shape of a single sample is under the given parameters. size just says that you want size-many more of those samples. as a matter of fact, size can be symbolic in v4!"

The big downside is that we're breaking backwards compatibility in a major way. Shapes are some of the hardest things to get correctly and I'm very concerned that users might shy away from `v4` because of that.

For now, @brandonwillard @michaelosthege and I agree that definitely we should start with keeping `shape` for compatibility. The question is whether we should ever remove it completely, which is where I'm having concerns. After talking @junpenglao who also shared these concerns.

ricardoV94:
Would specifying shape just be a redundant way of specifying size (+ some informative message if the shape didn't make sense)?

michaelosthege:
`shape` and `dims` are fully redundant to each other.

`shape` and `size` are only directly comparable _given_ the other parameters such as `cov=eye(3)`.

_Because_ `shape ≡ dims` and I think we all agree that `dims` was a great improvement of our API, I don't think we can/need/should make the internal logic go away.
However, we might still want to nudge everybody away from it.

One idea that circulated in our chat discussion was to slightly change what `shape` does in order to make it a new _feature_:

If distribution parameters are symbolic, **it's not immediately clear how the RV looks like**:
```
x = MvNormal(cov=cov_prior)
```
Specifying `shape` makes the model code easier to read, but makes it very rigid w.r.t. out-of-sample prediction:
```python
x = MvNormal(cov=cov_prior, shape=(3,))   # can no longer be broadcasted into a different size
```
It's the same as
```python
x = MvNormal(cov=cov_prior)
aesara.tensor.specify_shape(x, (3,))    # this Op needs to become a function output to be effective!
```

[...]

PR #4731, opened 2021-05-30
Make MvStudentT distribution v4 compatible

Sayam753:

+ [X] what are the (breaking) changes that this PR makes?
Making MvStudentT distribution v4 compatible as per https://github.com/pymc-devs/pymc3/issues/4686
+ [X] important background, or details about the implementation
Follows the instructions as per this https://github.com/pymc-devs/pymc3/issues/4686#issuecomment-839541040
+ [X] are the changes—especially new features—covered by tests and docstrings?
    - [X] Enabled pymc3/tests/test_distributions.py::TestMatchesScipy::test_mvt
    - [X] Enabled pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_mv_t
    - [X] pymc3/tests/test_distributions.py::TestBugfixes::test_issue_3051 
    - [X] But need to change `rng_fn` code to use a similar implementation as in master.
+ [X] [linting/style checks have been run](https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style)
+ [X] [consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples)

### DIFF
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -316,8 +316,8 @@ class MvStudentT(Continuous):
 
     Parameters
     ----------
-    nu: int
-        Degrees of freedom.
+    nu: float
+        Degrees of freedom, should be a positive scalar.
     Sigma: matrix
         Covariance matrix. Use `cov` in new code.
     mu: array
@@ -339,8 +339,8 @@ class MvStudentT(Continuous):
             if cov is not None:
                 raise ValueError("Specify only one of cov and Sigma")
             cov = Sigma
-        nu = at.as_tensor_variable(nu)
-        mu = at.as_tensor_variable(mu)
+        nu = at.as_tensor_variable(floatX(nu))
+        mu = at.as_tensor_variable(floatX(mu))
         cov = quaddist_matrix(cov, chol, tau, lower)
         return super().dist([nu, mu, cov], **kwargs)
 
@@ -361,7 +361,7 @@ class MvStudentT(Continuous):
         quaddist, logdet, ok = quaddist_parse(value, mu, cov)
         k = floatX(value.shape[-1])
 
-        norm = gammaln((nu + k) / 2.0) - gammaln(nu / 2.0) - 0.5 * k * floatX(np.log(nu * np.pi))
+        norm = gammaln((nu + k) / 2.0) - gammaln(nu / 2.0) - 0.5 * k * at.log(nu * np.pi)
         inner = -(nu + k) / 2.0 * at.log1p(quaddist / nu)
         return bound(norm + inner - logdet, ok)
 
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -27,6 +27,7 @@ from aesara.graph.op import Op
 from aesara.tensor import gammaln
 from aesara.tensor.nlinalg import det, eigh, matrix_inverse, trace
 from aesara.tensor.random.basic import MultinomialRV, dirichlet, multivariate_normal
+from aesara.tensor.random.op import RandomVariable, default_shape_from_params
 from aesara.tensor.random.utils import broadcast_params
 from aesara.tensor.slinalg import (
     Cholesky,
@@ -248,6 +249,66 @@ class MvNormal(Continuous):
         return ["mu", "cov"]
 
 
+def safe_multivariate_t(nu, mu, cov, size=None, rng=None):
+    res = np.atleast_1d(
+        stats.multivariate_t(loc=mu, shape=cov, df=nu, allow_singular=True).rvs(
+            size=size, random_state=rng
+        )
+    )
+
+    if size is not None:
+        res = res.reshape(list(size) + [-1])
+
+    return res
+
+
+class MvStudentTRV(RandomVariable):
+    name = "multivariate_studentt"
+    ndim_supp = 1
+    ndims_params = [0, 1, 2]
+    dtype = "floatX"
+    _print_name = ("MvStudentT", "\\operatorname{MvStudentT}")
+
+    def __call__(self, nu, mu=None, cov=None, size=None, **kwargs):
+
+        dtype = aesara.config.floatX if self.dtype == "floatX" else self.dtype
+
+        if mu is None:
+            mu = np.array([0.0], dtype=dtype)
+        if cov is None:
+            cov = np.array([[1.0]], dtype=dtype)
+        return super().__call__(nu, mu, cov, size=size, **kwargs)
+
+    def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
+        return default_shape_from_params(self.ndim_supp, dist_params, rep_param_idx, param_shapes)
+
+    @classmethod
+    def rng_fn(cls, rng, nu, mu, cov, size):
+
+        if mu.ndim > 1 or cov.ndim > 2:
+            # Neither SciPy nor NumPy implement parameter broadcasting for
+            # multivariate normals (or many other multivariate distributions),
+            # so we have implement a quick and dirty one here
+            mu, cov = broadcast_params([mu, cov], cls.ndims_params[1:])
+            size = tuple(size or ())
+
+            if size:
+                mu = np.broadcast_to(mu, size + mu.shape)
+                cov = np.broadcast_to(cov, size + cov.shape)
+
+            res = np.empty(mu.shape)
+            for idx in np.ndindex(mu.shape[:-1]):
+                m = mu[idx]
+                c = cov[idx]
+                res[idx] = safe_multivariate_t(nu, m, c, rng=rng)
+            return res
+        else:
+            return safe_multivariate_t(nu, mu, cov, size=size, rng=rng)
+
+
+mv_studentt = MvStudentTRV()
+
+
 class MvStudentT(Continuous):
     r"""
     Multivariate Student-T log-likelihood.
@@ -288,55 +349,20 @@ class MvStudentT(Continuous):
     lower: bool, default=True
         Whether the cholesky fatcor is given as a lower triangular matrix.
     """
+    rv_op = mv_studentt
 
-    def __init__(
-        self, nu, Sigma=None, mu=None, cov=None, tau=None, chol=None, lower=True, *args, **kwargs
-    ):
+    @classmethod
+    def dist(cls, nu, Sigma=None, mu=None, cov=None, tau=None, chol=None, lower=True, **kwargs):
         if Sigma is not None:
             if cov is not None:
                 raise ValueError("Specify only one of cov and Sigma")
             cov = Sigma
-        super().__init__(mu=mu, cov=cov, tau=tau, chol=chol, lower=lower, *args, **kwargs)
-        self.nu = nu = at.as_tensor_variable(nu)
-        self.mean = self.median = self.mode = self.mu = self.mu
+        nu = at.as_tensor_variable(nu)
+        mu = at.as_tensor_variable(mu)
+        cov = quaddist_matrix(cov, chol, tau, lower)
+        return super().dist([nu, mu, cov], **kwargs)
 
-    def random(self, point=None, size=None):
-        """
-        Draw random values from Multivariate Student's T distribution.
-
-        Parameters
-        ----------
-        point: dict, optional
-            Dict of variable values on which random values are to be
-            conditioned (uses default point if not specified).
-        size: int, optional
-            Desired size of random sample (returns one sample if not
-            specified).
-
-        Returns
-        -------
-        array
-        """
-        # with _DrawValuesContext():
-        #     nu, mu = draw_values([self.nu, self.mu], point=point, size=size)
-        #     if self._cov_type == "cov":
-        #         (cov,) = draw_values([self.cov], point=point, size=size)
-        #         dist = MvNormal.dist(mu=np.zeros_like(mu), cov=cov, shape=self.shape)
-        #     elif self._cov_type == "tau":
-        #         (tau,) = draw_values([self.tau], point=point, size=size)
-        #         dist = MvNormal.dist(mu=np.zeros_like(mu), tau=tau, shape=self.shape)
-        #     else:
-        #         (chol,) = draw_values([self.chol_cov], point=point, size=size)
-        #         dist = MvNormal.dist(mu=np.zeros_like(mu), chol=chol, shape=self.shape)
-        #
-        #     samples = dist.random(point, size)
-        #
-        # chi2_samples = np.random.chisquare(nu, size)
-        # # Add distribution shape to chi2 samples
-        # chi2_samples = chi2_samples.reshape(chi2_samples.shape + (1,) * len(self.shape))
-        # return (samples / np.sqrt(chi2_samples / nu)) + mu
-
-    def logp(value, nu, cov):
+    def logp(value, nu, mu, cov):
         """
         Calculate log-probability of Multivariate Student's T distribution
         at specified value.
@@ -350,7 +376,7 @@ class MvStudentT(Continuous):
         -------
         TensorVariable
         """
-        quaddist, logdet, ok = quaddist_parse(value, nu, cov)
+        quaddist, logdet, ok = quaddist_parse(value, mu, cov)
         k = floatX(value.shape[-1])
 
         norm = gammaln((nu + k) / 2.0) - gammaln(nu / 2.0) - 0.5 * k * floatX(np.log(nu * np.pi))
@@ -358,7 +384,7 @@ class MvStudentT(Continuous):
         return bound(norm + inner - logdet, ok)
 
     def _distr_parameters_for_repr(self):
-        return ["mu", "nu", "cov"]
+        return ["nu", "mu", "cov"]
 
 
 class Dirichlet(Continuous):
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -42,7 +42,7 @@ import pymc3 as pm
 
 from pymc3.aesaraf import floatX, intX
 from pymc3.distributions import transforms
-from pymc3.distributions.continuous import ChiSquared, Normal
+from pymc3.distributions.continuous import ChiSquared, Normal, assert_negative_support
 from pymc3.distributions.dist_math import bound, factln, logpow, multigammaln
 from pymc3.distributions.distribution import Continuous, Discrete
 from pymc3.math import kron_diag, kron_dot, kron_solve_lower, kronecker
@@ -342,6 +342,7 @@ class MvStudentT(Continuous):
         nu = at.as_tensor_variable(floatX(nu))
         mu = at.as_tensor_variable(floatX(mu))
         cov = quaddist_matrix(cov, chol, tau, lower)
+        assert_negative_support(nu, "nu", "MvStudentT")
         return super().dist([nu, mu, cov], **kwargs)
 
     def logp(value, nu, mu, cov):
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -279,9 +279,7 @@ class MvStudentTRV(RandomVariable):
         # Add distribution shape to chi2 samples
         chi2_samples = chi2_samples.reshape(chi2_samples.shape + (1,) * len(mu.shape))
 
-        mv_samples = pm.MvNormal.dist(
-            mu=np.zeros_like(mu), cov=cov, size=size, rng=aesara.shared(rng)
-        ).eval()
+        mv_samples = multivariate_normal.rng_fn(rng=rng, mean=np.zeros_like(mu), cov=cov, size=size)
 
         size = tuple(size or ())
         if size:
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -275,7 +275,7 @@ class MvStudentTRV(RandomVariable):
         # Don't reassign broadcasted cov, since MvNormal expects two dimensional cov only.
         mu, _ = broadcast_params([mu, cov], cls.ndims_params[1:])
 
-        chi2_samples = rng.chisquare(nu, size=size)
+        chi2_samples = np.sqrt(rng.chisquare(nu, size=size) / nu)
         # Add distribution shape to chi2 samples
         chi2_samples = chi2_samples.reshape(chi2_samples.shape + (1,) * len(mu.shape))
 
@@ -287,7 +287,7 @@ class MvStudentTRV(RandomVariable):
         if size:
             mu = np.broadcast_to(mu, size + mu.shape)
 
-        return (mv_samples / np.sqrt(chi2_samples / nu)) + mu
+        return (mv_samples / chi2_samples) + mu
 
 
 mv_studentt = MvStudentTRV()
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -249,19 +249,6 @@ class MvNormal(Continuous):
         return ["mu", "cov"]
 
 
-def safe_multivariate_t(nu, mu, cov, size=None, rng=None):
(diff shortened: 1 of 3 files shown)


Sayam753:
Looking at the failed [pipeline](https://github.com/pymc-devs/pymc3/pull/4731/checks?check_run_id=2704501178), it seems like it uses cached `scipy==1.2.1`. But doing a `pip install -e .` on v4, gives me `scipy==1.6.3`. 

I am using `scipy.stats.multivariate_t` for drawing random samples which was added in `scipy==1.6.0`. I think its better to not use a function from a newer version, and revert to using a similar implementation as in master -

https://github.com/pymc-devs/pymc3/blob/d7172c0a1a76301031d1b3b411d00643c416a0c4/pymc3/distributions/multivariate.py#L406-L409

ricardoV94:
Why not use the newer scipy version?

Sayam753:
> Why not use the newer scipy version?

I think it may lead to changing the requirements file and our pipeline cache. If I already had `scipy<1.6.0` cached locally, doing `pip install -e .` would have installed that earlier version of scipy (which will still be compatible with pymc3 requirements of `scipy>=1.2.0`)

If using newer scipy versions is feasible with other dependencies' requirements, I am happy to make the required changes :)

pymc3/tests/test_distributions_random.py
```diff
@@ -1380,7 +1380,6 @@ def ref_rand_evd(size, mu, evds, sigma):
                 model_args=evd_args,
             )
 
-    @pytest.mark.xfail(reason="This distribution has not been refactored for v4")
```
ricardoV94, line 1405: The idea is for these old tests that use `pymc3_random` to be replaced by the new, less wasteful, `BaseTestDistribution` tests. Once you confirm this passes locally, you can go ahead and replace it.

pymc3/tests/test_distributions.py
```diff
@@ -2909,9 +2909,12 @@ class TestBugfixes:
         "dist_cls,kwargs", [(MvNormal, dict(mu=0)), (MvStudentT, dict(mu=0, nu=2))]
     )
     @pytest.mark.parametrize("dims", [1, 2, 4])
-    @pytest.mark.xfail(reason="Distribution not refactored yet")
     def test_issue_3051(self, dims, dist_cls, kwargs):
-        d = dist_cls.dist(**kwargs, cov=np.eye(dims), size=(dims,))
+        mu = np.repeat(kwargs["mu"], dims)
```
ricardoV94: Not very important but this would also work:

```python
    @pytest.mark.parametrize(
        "dist_cls,kwargs", [(MvNormal, dict()), (MvStudentT, dict(nu=2))]
    )
    @pytest.mark.parametrize("dims", [1, 2, 4])
    def test_issue_3051(self, dims, dist_cls, kwargs):
        mu = np.repeat(0, dims)
        d = dist_cls.dist(mu=mu, cov=np.eye(dims), **kwargs, size=20)
```

Sayam753:
`scipy.stats.multivariate_t` has weird behavior. Its `rvs` method works only when `size` parameter is strictly 1 dimensional.

```python
>>> import numpy as np
>>> import scipy
>>> scipy.__version__
'1.6.3'
>>> from scipy.stats import multivariate_normal, multivariate_t
>>> mean, cov = np.zeros(2), np.eye(2)
>>> multivariate_normal.rvs(mean=mean, cov=cov, size=None).shape
(2,)
>>> multivariate_t.rvs(loc=mean, shape=cov, size=None)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 4148, in rvs
    samples = loc + z / np.sqrt(x)[:, None]
IndexError: invalid index to scalar variable.
>>> multivariate_normal.rvs(mean=mean, cov=cov, size=(10, 3)).shape
(10, 3, 2)
>>> multivariate_t.rvs(loc=mean, shape=cov, size=(10, 3))
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 4148, in rvs
    samples = loc + z / np.sqrt(x)[:, None]
ValueError: operands could not be broadcast together with shapes (10,3,2) (10,1,3) 
>>> multivariate_t.rvs(loc=mean, shape=cov, size=(3))
array([[ 1.54502738, -0.23866557],
       [-0.79040924, -1.72035808],
       [-0.28060545, -0.69889876]])
```
This is the reason why many pipelines failed. 
 
Now, I want to try out implementing `rng_fn` method similar to what we had in v3.

Sayam753:
I am on latest main branch@ 4de95c3e8e65d8897ee48017f735418e8181c416 and trying to understand how new API for shapes work.

A small snippet -
```python
>>> import pymc3 as pm
You are running the v4 development version of PyMC3 which currently still lacks key features. You probably want to use the stable v3 instead which you can either install via conda or find on the v3 GitHub branch: https://github.com/pymc-devs/pymc3/tree/v3
>>> import numpy as np
>>> dist = pm.MvNormal.dist(mu=np.zeros(1), cov=np.eye(3), shape=3)
>>> dist.shape.eval()
array([1])
```

I am finding it hard to understand why the shape is [1]
Shouldn't it be [3]? 

And some fancy errors during `dist.eval()`

<details>
<summary>Traceback</summary>


```python
>>> dist.eval()
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/compile/function/types.py", line 977, in __call__
    if output_subset is None
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/graph/op.py", line 495, in rval
    r = p(n, [x[0] for x in i], o)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/tensor/random/op.py", line 413, in perform
    smpl_val = self.rng_fn(rng, *(args + [size]))
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/tensor/random/basic.py", line 288, in rng_fn
    return safe_multivariate_normal(mean, cov, size=size, rng=rng)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/tensor/random/basic.py", line 239, in safe_multivariate_normal
    stats.multivariate_normal(mean=mean, cov=cov, allow_singular=True).rvs(
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 369, in __call__
    seed=seed)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 744, in __init__
    None, mean, cov)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 409, in _process_parameters
    cov.shape = (1, 1)
ValueError: cannot reshape array of size 9 into shape (1,1)

During handling of the above exception, another exception occurred:

  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/graph/basic.py", line 555, in eval
    rval = self._fn_cache[inputs](*args)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/compile/function/types.py", line 993, in __call__
    storage_map=getattr(self.fn, "storage_map", None),
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/link/utils.py", line 521, in raise_with_op
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/compile/function/types.py", line 977, in __call__
    if output_subset is None
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/graph/op.py", line 495, in rval
    r = p(n, [x[0] for x in i], o)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/tensor/random/op.py", line 413, in perform
    smpl_val = self.rng_fn(rng, *(args + [size]))
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/tensor/random/basic.py", line 288, in rng_fn
    return safe_multivariate_normal(mean, cov, size=size, rng=rng)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/tensor/random/basic.py", line 239, in safe_multivariate_normal
    stats.multivariate_normal(mean=mean, cov=cov, allow_singular=True).rvs(
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 369, in __call__
    seed=seed)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 744, in __init__
    None, mean, cov)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 409, in _process_parameters
    cov.shape = (1, 1)
ValueError: cannot reshape array of size 9 into shape (1,1)
Apply node that caused the error: multivariate_normal_rv(RandomStateSharedVariable(<RandomState(MT19937) at 0x7FB57E736380>), TensorConstant{[]}, TensorConstant{11}, TensorConstant{(1,) of 0.0}, TensorConstant{[[1. 0. 0...0. 0. 1.]]})
Toposort index: 0
Inputs types: [RandomStateType, TensorType(int64, vector), TensorType(int64, scalar), TensorType(float64, (True,)), TensorType(float64, matrix)]
Inputs shapes: ['No shapes', (0,), (), (1,), (3, 3)]
Inputs strides: ['No strides', (8,), (), (8,), (24, 8)]
Inputs values: [RandomState(MT19937) at 0x7FB57E736380, array([], dtype=int64), array(11), array([0.]), 'not shown']
Outputs clients: [[], ['output']]

Backtrace when the node is created (use Aesara flag traceback__limit=N to make it longer):
  File "/Users/sayam/Desktop/pymc/pymc3/pymc3/distributions/multivariate.py", line 226, in dist
    return super().dist([mu, cov], **kwargs)
  File "/Users/sayam/Desktop/pymc/pymc3/pymc3/distributions/distribution.py", line 285, in dist

HINT: Use the Aesara flag `exception_verbosity=high` for a debug print-out and storage map footprint of this Apply node.
```

</details>

cc @michaelosthege

michaelosthege:
>```
> >>> dist = pm.MvNormal.dist(mu=np.zeros(1), cov=np.eye(3), shape=3)
> >>> dist.shape.eval()
> array([1])
> ```
> 
> I am finding it hard to understand why the shape is [1]
> Shouldn't it be [3]?

This parametrization is incoherent. However, this is a scenario that motivates automatically adding a `SpecifyShape` `Op` which I had implemented but then took out, because I couldn't come up with a test case to trigger it.

Your `shape` kwarg has no effect, because the last `aesara.tensor.random.basic.multivariate_normal.ndim_supp` dimensions are shaved off.
Internally it becomes:
```python
>>> import aesara
>>> aesara.__version__
'2.0.8'
>>> import aesara.tensor as at
>>> import aesara.tensor.random.basic as ar
>>> import numpy as np
>>> ar.multivariate_normal(mean=np.zeros(1), cov=np.eye(3)).shape.eval()
array([1], dtype=int64)
```

The `mean`/`mu` parameter to the MvNormal should include the support dimensions. So for example:
```python
>>> ar.multivariate_normal(mean=[0,0,0], cov=np.eye(3)).shape.eval()
array([3], dtype=int64)
```

One could argue that the Aesara implementation shouldn't have downcasted `cov` just because `mean` had shape `(1,)`. At least not without printing a warning.
cc @brandonwillard

Sayam753:
> The mean/mu parameter to the MvNormal should include the support dimensions

I guess this is due to [rep_param_idx](https://github.com/aesara-devs/aesara/blob/main/aesara/tensor/random/op.py#L27) parameter used to infer distribution shape in aesara.

> One could argue that the Aesara implementation shouldn't have downcasted cov just because mean had shape (1,). At least not without printing a warning.

How hard would it be to loop over all distribution parameters to infer distribution shape, instead of relying on one parameter?

brandonwillard:
> However, this is a scenario that motivates automatically adding a `SpecifyShape` `Op` which I had implemented but then took out, because I couldn't come up with a test case to trigger it.

`SpecifyShape` wouldn't help at all.


> The `mean`/`mu` parameter to the MvNormal should include the support dimensions. So for example:
> 
> ```python
> >>> ar.multivariate_normal(mean=[0,0,0], cov=np.eye(3)).shape.eval()
> array([3], dtype=int64)
> ```

This has nothing to do with "including the support dimensions"; the actual issue is that the parameters specify **conflicting lengths/sizes** across the relevant dimensions.  Your follow-up example isn't including anything extra, it's only fixing the conflicting dimensions by making the length of the mean match the lengths of the covariance dimensions.

Consider what `[0, 0, 0]` actually is: a **one-dimensional vector of length three**.  It's not a three-dimensional array.

I think a lot of confusion regarding `shape`, `size`, etc., might be stemming from this fundamental misconception regarding dimensions and their sizes/lengths.

> One could argue that the Aesara implementation shouldn't have downcasted `cov` just because `mean` had shape `(1,)`. At least not without printing a warning.

What "downcasting"?  The dtype looks correct.

The result itself appears to be due to a missing check in the shape inference.  That's all.

If you want to understand why, start by computing the result itself, and not just its shape.  You'll see the underlying error that the shape inference logic missed:
```python
>>> ar.multivariate_normal(mean=np.zeros(1), cov=np.eye(3)).eval()
...
ValueError: cannot reshape array of size 9 into shape (1,1)
```

This is the same error that appears in the original problem.  In other words, the `shape` parameter isn't even relevant, nor is the `MvNormal` wrapper.

As usual, it's always good to see what the underlying NumPy function would do, since these `Op`s are modeling/wrapping those functions:
```python
>>> np.random.multivariate_normal(mean=np.zeros(1), cov=np.eye(3))
ValueError: mean and cov must have same length
```

As expected, NumPy fails as well; however, it gives a much better error message.


@Sayam753, feel free to open an issue in Aesara to add that missing check to the shape inference steps; otherwise, your example specifies a mathematically invalid distribution, because the mean and covariance are incompatible.  Use `mu=np.zeros(3)` if you want a multivariate normal that produces vectors of length three instead.

michaelosthege:
> SpecifyShape wouldn't help at all.

IIRC I once had a version where `SpecifyShape` worked and made sense (when working on #4552). I don't remeber the details or what changed though.
Brandon is right: The `ValueError: operands could not be broadcast together` is raised before a `SpecifyShape` can get involved.

> What "downcasting"? The dtype looks correct.

Sorry, I mean down_sizing_. (The bug in the shape inference.)

> This has nothing to do with "including the support dimensions"; the actual issue is that the parameters specify conflicting lengths/sizes across the relevant dimensions.

Please excuse that my language wasn't exactly on point. The phrase "including the support dimensions" is more from the perspective of wanting to imply more dimensions from the RV parameters, which wasn't the intention in the first place.

Anyways. @Sayam753 let us know if you need assistance with the remaining ToDo for this PR.

Sayam753:
Thanks @michaelosthege and @brandonwillard for clearing the doubts above.

A little interesting behaviour of `MvNormal.logp` as it works even when mean and cov are not having the same length -

```python
>>> import numpy as np
>>> import pymc3 as pm
You are running the v4 development version of PyMC3 which currently still lacks key features. You probably want to use the stable v3 instead which you can either install via conda or find on the v3 GitHub branch: https://github.com/pymc-devs/pymc3/tree/v3
>>> mu, cov = np.ones(1), np.eye(3)
>>> pm.MvNormal.logp(value=np.random.randn(20, 3), mu=mu, cov=cov).eval()
array([ -3.67340545,  -7.19130309,  -3.23923742,  -5.15713797,
       -17.29246223, -10.32441825,  -4.2939353 ,  -5.66928884,
        -4.05121295,  -4.58167586,  -5.50445437,  -4.20362596,
        -6.12579132,  -8.63937152,  -4.656641  ,  -9.63474273,
        -4.35076017,  -3.33788195,  -3.50893476,  -4.77016247])
```

ricardoV94:
Calling the `logp` method directly does not go through the random variable op, so no shape or size checks would be done (it's just a normal aesara function call). As long as the function broadcasts well you wouldn't see any error.

Sayam753:
I am trying to understand how [`logpt`](https://github.com/pymc-devs/pymc3/blob/main/pymc3/distributions/logp.py#L112) works under the hood. This will help in writing a better solution for the test case [pymc3/tests/test_distributions.py::TestBugfixes::test_issue_3051](https://github.com/pymc-devs/pymc3/blob/main/pymc3/tests/test_distributions.py#L2945-L2953)

My question is - why cannot I use `filter_variable` method to broadcast a vector of shape (3) and a matrix of shape (20, 3) in aesara?

```python
>>> import numpy as np
>>> import aesara.tensor as at
>>> import pymc3 as pm
You are running the v4 development version of PyMC3 which currently still lacks key features. You probably want to use the stable v3 instead which you can either install via conda or find on the v3 GitHub branch: https://github.com/pymc-devs/pymc3/tree/v3
>>> mu, cov = np.ones(3), np.eye(3)
>>> dist = pm.MvNormal.dist(mu=mu, cov=cov)
>>> X = np.random.normal(size=(20, 3))

>>> X = at.as_tensor(X)
>>> dist.type, dist.dtype, dist.ndim, dist.shape.eval()
(TensorType(float64, vector), 'float64', 1, array([3]))
>>> X.type, X.dtype, X.ndim, X.shape.eval()
(TensorType(float64, matrix), 'float64', 2, array([20,  3]))

>>> dist.type.filter_variable(X)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/aesara/tensor/type.py", line 261, in filter_variable
    f"Cannot convert Type {other.type} "
TypeError: Cannot convert Type TensorType(float64, matrix) (of Variable TensorConstant{[[ 0.30673..15832267]]}) into Type TensorType(float64, vector). You can try to manually convert TensorConstant{[[ 0.30673..15832267]]} into a TensorType(float64, vector).

```

cc @brandonwillard

Sayam753:
The pipelines are green. So, the PR is ready for another round of review.

pymc3/distributions/multivariate.py
```diff
@@ -248,6 +249,50 @@ def _distr_parameters_for_repr(self):
...
+    def rng_fn(cls, rng, nu, mu, cov, size):
+
+        # Don't reassign broadcasted cov, since MvNormal expects two dimensional cov only.
+        mu, _ = broadcast_params([mu, cov], cls.ndims_params[1:])
+
+        chi2_samples = np.sqrt(rng.chisquare(nu, size=size) / nu)
+        # Add distribution shape to chi2 samples
+        chi2_samples = chi2_samples.reshape(chi2_samples.shape + (1,) * len(mu.shape))
+
+        mv_samples = pm.MvNormal.dist(
```
ricardoV94: This seems odd. Can't we call scipy/numpy MvNormal directly? Or just the rng classmethod of aesara's MvNormal op?

My understanding is that random ops should not depend on pymc distributions at all.

ricardoV94 (changes requested):

pymc3/distributions/multivariate.py
Sayam753: > Can't we call scipy/numpy MvNormal directly

Unfortunately, scipy and numpy `multivariate_normal` generators cannot work when `mu` takes two-dimensional shape -

```python
>>> import numpy as np
>>> import scipy
>>> np.__version__, scipy.__version__
('1.20.3', '1.6.3')
>>> mu, cov = np.ones((10, 3)), np.eye(3)
>>> np.random.multivariate_normal(mean=mu, cov=cov)
  File "mtrand.pyx", line 4071, in numpy.random.mtrand.RandomState.multivariate_normal
ValueError: mean must be 1 dimensional
>>> from scipy.stats import multivariate_normal
>>> multivariate_normal.rvs(mean=mu, cov=cov)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 663, in rvs
    dim, mean, cov = self._process_parameters(None, mean, cov)
  File "/Users/sayam/miniconda3/envs/pymc_dev/lib/python3.7/site-packages/scipy/stats/_multivariate.py", line 413, in _process_parameters
    dim)
ValueError: Array 'mean' must be a vector of length 30.
>>> 
```

Aesara's [multivariate_normal RV](https://github.com/aesara-devs/aesara/blob/3acdd78e7a5d27e70cd8de3fc00bd5f010c65ee3/aesara/tensor/random/basic.py#L270-L288) handles all the cases. So that's why I am using `pm.MvNormal.dist` to access underneath implementation.

> just the rng classmethod of aesara's MvNormal op

Let me try this out

ricardoV94:
In the meantime, were there no other tests marked as xfail because they relied on this distribution?

Sayam753:
I think all the MvStudentT related tests in `test_distributions.py` and `test_distributions_random.py` have been refactored/added in this PR.

Thinking more about the test cases, `MvStudentTRandomWalk`'s rng method must depend on `MvStudentT` rng method. AFAIK the RandomWalk distributions have not been refactored yet and including them here will be an overkill for this PR.

pymc3/distributions/multivariate.py
```diff
@@ -288,55 +331,20 @@ class MvStudentT(Continuous):
...
-        #         dist = MvNormal.dist(mu=np.zeros_like(mu), chol=chol, shape=self.shape)
-        #
-        #     samples = dist.random(point, size)
-        #
-        # chi2_samples = np.random.chisquare(nu, size)
-        # # Add distribution shape to chi2 samples
-        # chi2_samples = chi2_samples.reshape(chi2_samples.shape + (1,) * len(self.shape))
-        # return (samples / np.sqrt(chi2_samples / nu)) + mu
+        nu = at.as_tensor_variable(nu)
+        mu = at.as_tensor_variable(mu)
```
```suggestion
        nu = at.as_tensor_variable(floatX(nu))
        mu = at.as_tensor_variable(floatX(mu))
```

The docstrings should indicate a positive float is expected `nu > 0`

pymc3/distributions/multivariate.py
```diff
@@ -350,15 +358,15 @@ def logp(value, nu, cov):
         -------
         TensorVariable
         """
-        quaddist, logdet, ok = quaddist_parse(value, nu, cov)
+        quaddist, logdet, ok = quaddist_parse(value, mu, cov)
         k = floatX(value.shape[-1])
 
         norm = gammaln((nu + k) / 2.0) - gammaln(nu / 2.0) - 0.5 * k * floatX(np.log(nu * np.pi))
```
```suggestion
        norm = gammaln((nu + k) / 2.0) - gammaln(nu / 2.0) - 0.5 * k * at.log(nu * np.pi)
```

ricardoV94: `nu` will be converted to a floatX in `dist` above, so it should be safe to remove this extra call

ricardoV94 (commented):
I left some minor suggestions. Otherwise it looks good to me. Great work :)

pymc3/distributions/multivariate.py
Sayam753: Do we have something along the lines of [assert_negative_support](https://github.com/pymc-devs/pymc3/blob/v3/pymc3/distributions/continuous.py#L120) in v4 as well? for ensuring the positive nature of nu?

ricardoV94: We have, it's being used in some of the univariate dists I think.

 I am not sure about it's usefulness but you can add it if you want.

ricardoV94 (approved):
LGTM!

ricardoV94:
Thanks @Sayam753, great work.

Sayam753:
Thanks @ricardoV94 for a thorough review. I do want to follow up for a better fix for pymc3/tests/test_distributions.py::TestBugfixes::test_issue_3051 and will open a related issue :)

PR #4732, opened 2021-06-01
Remove unnecessary gammaln and psi from pymc3.distributions.special

brandonwillard:
This PR updates `pymc3.distributions.special` so that it's compatible with the changes introduced in https://github.com/pymc-devs/aesara/pull/445.  It also finishes the unused test for `multigammaln`.


ricardoV94 (approved):
LGTM, It feels like this file could be merged with `dist_math.py`, but no need to do it in this PR.

PR #4519, opened 2021-03-09
Replace slow dist_math::incomplete_beta with aesara op

ricardoV94:
This PR replaces the `incomplete_beta` method in dist_math by an aesara op `betainc` that wraps the `scipy.special.betainc` method.

This provides a large increase in speed (and graph compilation) even without providing the c_code implementation directly (this can always be added later). It also allows the evaluation of the logcdf for multiple values in `Beta`, `StudentT`, `Binomial`, `NegativeBinomial` (and zero inflated variants). I removed the previous limitation warning when trying to evaluate these methods for multiple values. These changes are critical for any practical implementation of Truncated Versions of these distributions.

Gradients were adapted from the STAN implementation, and require the use of two extra "pseudo aesara ops" that simply wrap the equivalent python functions. This does not allow for further graph optimizations, but I doubt aesara could do much with the previous scan implementation either. Let me know if you think there is a better solution.

Depending on what your PR does, here are a few things you might want to address in the description:
+ [x] what are the (breaking) changes that this PR makes? dist_math::incomplete_beta was deprecated in favor of the new betainc op
+ [x] important background, or details about the implementation
+ [x] are the changes—especially new features—covered by tests and docstrings?
+ [x] [linting/style checks have been run](https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style)
+ [x] [consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples)
+ [x] right before it's ready to merge, mention the PR in the RELEASE-NOTES.md

Closes #4420

PR #4734, opened 2021-06-03
Speedup logcdf tests

ricardoV94:
The logcdf tests were running very slow due to constantly rebuilding the logcdf function. 

Also fixed a couple of failing tests on float32 (which were difficult to identify before via because setting `n_samples=-1` took ages to run.

### DIFF
--- a/pymc3/tests/test_distributions.py
+++ b/pymc3/tests/test_distributions.py
@@ -1137,13 +1137,17 @@ class TestMatchesScipy:
             {"alpha": Rplus, "beta": Rplus},
             lambda value, alpha, beta: sp.beta.logpdf(value, alpha, beta),
         )
-        self.check_logp(Beta, Unit, {"mu": Unit, "sigma": Rplus}, beta_mu_sigma)
+        self.check_logp(
+            Beta,
+            Unit,
+            {"mu": Unit, "sigma": Rplus},
+            beta_mu_sigma,
+        )
         self.check_logcdf(
             Beta,
             Unit,
             {"alpha": Rplus, "beta": Rplus},
             lambda value, alpha, beta: sp.beta.logcdf(value, alpha, beta),
-            n_samples=10,
             decimal=select_by_precision(float64=5, float32=3),
         )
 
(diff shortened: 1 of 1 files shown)


ricardoV94:
I am getting a strange issue with some of the logcdf methods which seems to be behind the failing tests:

Setting the values to constants gives the correct values
```python
with pm.Model() as model:
    value = pm.NegativeBinomial('value', mu=0.9, alpha=0.1)

logcdf = logpt(model['value'], cdf=True)
fn1 = model.fastfn(logcdf)

print(fn1({'value': 1.0}))  # -0.14408081390569186
```

But using shared variables does not
```python
with pm.Model() as model:
    mu = aesara.shared(np.asarray(0.9))
    alpha = aesara.shared(np.asarray(0.1))
    value = pm.NegativeBinomial('value', mu=mu, alpha=alpha)

logcdf = logpt(model['value'], cdf=True)
fn2 = model.fastfn(logcdf)

mu.set_value(0.9)
alpha.set_value(0.1)
print(fn2({'value': 1.0}))  # -0.08777554398474162
```

~~This doesn't happen with all distributions, and unfortunately those that do have a crazy long graph to be able to compare easily side by side :/~~

Found the culprint: it's the incomplete_beta function:

```python
incomplete_beta(0.1, 2.0, 0.1).eval()   # array(0.86581778)

alpha = aesara.shared(0.1, 'alpha')
beta = aesara.shared(2.0, 'beta')
incomplete_beta(alpha, beta, 0.1).eval()   # array(0.91596645)

import scipy.special
scipy.special.betainc(0.1, 2.0, 0.1)   # 0.8658177758494668
``` 

All failing tests (expect for the HyperGeometric, which is a different issue) rely on the incomplete_beta. 

Maybe it's time I finish #4519

ricardoV94:
Also found some issues with the recent `initval` changes, as it doesn't respect the parents initvals.

This snippet often leads to a `ValueError`:

```python
with pm.Model() as model:
    mu = pm.Normal('mu', initval=100)
    alpha = pm.HalfNormal('alpha', initval=100, transform=None)
    value = pm.NegativeBinomial('value', mu=mu, alpha=alpha)

model.initial_values
# {mu: array(100., dtype=float32),
#  alpha: array(100., dtype=float32),
#  value: array(1)}
```

When it doesn't fail, initval is still far from the expected ~ 100:
```python
pm.NegativeBinomial.dist(mu=100, alpha=100).eval()
# array(98)
```

ricardoV94:
The `HyperGeometric` test is failing because of this issue in Aesara: https://github.com/pymc-devs/aesara/issues/461

ricardoV94:
Closing in favor of #4736

PR #4736, opened 2021-06-04
Replace incomplete_beta with betainc and speedup logcdf tests

### LINKED ISSUE #4420: dist_math.incomplete_beta is very slow
I notice this when running test_distributions.py. Any distribution `logcdf` method that uses `incomplete_beta` takes ages to run. 

Right now it is used in the `logcdf` methods of  `Beta`, `StudentT`, `Binomial`, `NegativeBinomial`, and `ZeroInflated` versions of the last two distributions.

Is there a reason why this cannot be implemented as a Theano C Op? Scipy has C implementation here, which seems to be doing exactly the same as our algorithm in `dist_math.py`:
https://github.com/scipy/scipy/blob/master/scipy/special/cephes/incbet.c

In addition, this would probably make it trivial to implement a vectorized version for tensors, allowing the logcdf methods of the distributions above to evaluate more than one value at a time.

### LINKED ISSUE #4467: Remove hacks to avoid previous aesara gammainc(c) limitations
**This is a reminder issue**

The gamma functions in aesara were changed to work with invalid values (returning nan) instead of doing a c-level assertion (https://github.com/pymc-devs/aesara/pull/290). The following lines can be changed / removed once pymc3 is paired with the next version of aesara:

Distributions:

https://github.com/pymc-devs/pymc3/blob/89916ad724ae88ebb4c1af3d65919b87b318f180/pymc3/distributions/continuous.py#L2631-L2634

https://github.com/pymc-devs/pymc3/blob/89916ad724ae88ebb4c1af3d65919b87b318f180/pymc3/distributions/continuous.py#L2805-L2808

Unittests:

https://github.com/pymc-devs/pymc3/blob/e40304b98503907f620d67d4a9ff6e42c63171d1/pymc3/tests/test_distributions.py#L1231-L1240

https://github.com/pymc-devs/pymc3/blob/e40304b98503907f620d67d4a9ff6e42c63171d1/pymc3/tests/test_distributions.py#L1253-L1262

ricardoV94:
# Implement Betainc

This PR replaces the incomplete_beta method in dist_math by an aesara op `betainc` that wraps the `scipy.special.betainc` method. Supersedes #4519; Closes #4420.

This provides a large increase in speed (and graph compilation) even without providing the c_code implementation directly (this can always be added later). I did some comparisons [here](https://github.com/ricardoV94/derivatives_betainc/blob/master/comparison_aesara.ipynb). On my machine, the compilation time of a simple input/output function for the `incomplete_beta` was around 10 seconds, and it's pretty much instantaneous for the `betainc` as would be expected. Run time is ~2 orders of magnitude faster for the main op (even though it's calling scipy via Python), and ~4-5 orders of magnitude faster for the gradient.

It also allows the evaluation of the logcdf for multiple values in Beta, StudentT, Binomial, NegativeBinomial (and zero inflated variants). I removed the previous limitation warning when trying to evaluate these methods for multiple values. These changes are critical for any practical implementation of Truncated Versions of these distributions.

The derivative algorithm was adapted from [this paper](https://core.ac.uk/download/pdf/6303153.pdf) and requires the use of two extra "pseudo aesara ops" that simply wrap the equivalent python function. This does not allow for further graph optimizations, but I doubt aesara could do much with the previous scan implementation either. I did some extensive comparison with the STAN implementation, and these derivatives seem to be both faster and more accurate than theirs, although it's not easy to find a good numerical reference. Some discussion can be found in https://github.com/stan-dev/math/issues/2417

The one downside is the constant warning that the op has no `c-implementation` whenever the logp/logcdf makes use of it.

# Speedup logcdf tests

The logcdf tests are refactored to avoid recreating the logcdf graph for every parameter / value evaluation. This supersedes #4734. The old `incomplete_beta` was found to be defective when attempting #4734, and so I decided to merge the two PRs into this one instead.

This allows to remove most of the custom `n_samples` limitations, and makes the test suite run considerably faster.

A temporary work-around for the `HyperGeometric` tests was added, related to an optimization issue on Aesara side https://github.com/pymc-devs/aesara/issues/461

Finally, closes #4467

### DIFF
--- a/pymc3/distributions/dist_math.py
+++ b/pymc3/distributions/dist_math.py
@@ -17,18 +17,18 @@ Created on Mar 7, 2011
 
 @author: johnsalvatier
 """
+import warnings
+
 import aesara
 import aesara.tensor as at
 import numpy as np
 import scipy.linalg
 import scipy.stats
 
-from aesara import scan
 from aesara.compile.builders import OpFromGraph
 from aesara.graph.basic import Apply
 from aesara.graph.op import Op
 from aesara.scalar import UnaryScalarOp, upgrade_to_float_no_complex
-from aesara.scan import until
 from aesara.tensor import gammaln
 from aesara.tensor.elemwise import Elemwise
 from aesara.tensor.slinalg import Cholesky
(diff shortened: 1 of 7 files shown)


ricardoV94:
Failing test was related to the minimum scipy version 1.2.0 not including `betabinom`, and one of tests now calls it, incidentally, to draw the `initval`. This is somewhat related to #4737

I raised the requirement to 1.4.1, is this okay? There were some concerns about raising it to 1.6.0, because Colab is still on 1.4.1, but this should be fine in that regard.

ricardoV94:
Looking at the pytest workflow, this PR seems to shave about 20-25 minutes from ~1h05m to 40m

brandonwillard (changes requested):
This is the wrong place for an Aesara implementation of a `scipy.special` `Op`.  Such `Op`s should be submitted as a PR to Aesara, so they can go alongside all the other `scipy.special` `Op`s; otherwise, we'll have more of the same unnecessary overlap in testing and maintenance that we've been trying to avoid.

Regardless, this is a very exciting update!

ricardoV94:
I can open a PR for the betainc in Aesara. When I opened the PR for the first time nobody mentioned that interest :)

pymc3/distributions/dist_math.py
```diff
@@ -634,3 +470,224 @@ def clipped_beta_rvs(a, b, size=None, random_state=None, dtype="float64"):
...
+    dBm2, dBm1 = 0, 0
+
+    f = (q * x) / (p * (1 - x))
+    K = np.exp(p * np.log(x) + (q - 1) * np.log1p(-x) - np.log(p) - scipy.special.betaln(p, q))
+    if wrtp:
+        dK = np.log(x) - 1 / p + scipy.special.digamma(p + q) - scipy.special.digamma(p)
+    else:
+        dK = np.log1p(-x) + scipy.special.digamma(p + q) - scipy.special.digamma(q)
+
+    for n in range(1, max_iters + 1):
```
junpenglao: Is the plain python for loop going to be slow here, or is the function is meant to run in python/numpy/scipy

brandonwillard:
> I can open a PR for the betainc in Aesara. When I opened the PR for the first time nobody mentioned that interest :)

Sorry, I've been too busy to keep up on PRs, but, if you request a review, I'll usually get to it by the weekend.

pymc3/distributions/dist_math.py
ricardoV94: Main Python loop for now. It's not really possible to vectorize as it is an iterative loop. It's still about ~4-5 orders of magnitude faster than the old `C` derivatives obtained by autodiff from the scan implementation of the `incomplete_beta`.

It's also very simple code to port to `C` or to `Numba` but I didn't want to do it just yet.

I have some speed comparisons here https://www.github.com/ricardoV94/derivatives_betainc/tree/master/comparison_aesara.ipynb

ricardoV94:
> > I can open a PR for the betainc in Aesara. When I opened the PR for the first time nobody mentioned that interest :)
> 
> Sorry, I've been too busy to keep up on PRs, but, if you request a review, I'll usually get to it by the weekend. 

I didn't mean anyone was at fault, there is no rush on my side. I'll open it there soon. 😅

ricardoV94:
Closing this one and opening a new one, since many things have changed

PR #4742, opened 2021-06-05
Rename references to the PyMC3 master branch to main

michaelosthege:
This also renames the conditions in GitHub action workflow definitions.

Let's hope off this works out with the CI.

### DIFF
--- a/docs/source/api/bounds.rst
+++ b/docs/source/api/bounds.rst
@@ -66,7 +66,7 @@ Caveats
 
 * Bounds cannot be given to variables that are ``observed``.  To model
   truncated data, use a :func:`~pymc3.model.Potential` in combination with a cumulative
-  probability function.  See `this example notebook <https://github.com/pymc-devs/pymc-examples/blob/main/examples/survival_analysis/weibull_aft.ipynb>`_.
+  probability function.  See `this example notebook <https://docs.pymc.io/pymc-examples/examples/survival_analysis/weibull_aft.html>`_.
 
 * The automatic transformation applied to continuous distributions results in
   an unnormalized probability distribution.  This doesn't effect inference
--- a/docs/source/api/bounds.rst
+++ b/docs/source/api/bounds.rst
@@ -66,7 +66,7 @@ Caveats
 
 * Bounds cannot be given to variables that are ``observed``.  To model
   truncated data, use a :func:`~pymc3.model.Potential` in combination with a cumulative
-  probability function.  See `this example notebook <https://github.com/pymc-devs/pymc3/blob/master/docs/source/notebooks/weibull_aft.ipynb>`_.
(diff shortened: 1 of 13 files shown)


twiecki (approved):

docs/source/api/bounds.rst
```diff
@@ -66,7 +66,7 @@ Caveats
 
 * Bounds cannot be given to variables that are ``observed``.  To model
   truncated data, use a :func:`~pymc3.model.Potential` in combination with a cumulative
-  probability function.  See `this example notebook <https://github.com/pymc-devs/pymc3/blob/master/docs/source/notebooks/weibull_aft.ipynb>`_.
+  probability function.  See `this example notebook <https://github.com/pymc-devs/pymc-examples/blob/main/examples/survival_analysis/weibull_aft.ipynb>`_.
```
OriolAbril: shouldn't this point to the website instead of github?

docs/source/developer_guide.rst
```diff
@@ -712,15 +712,15 @@ dictionary with the free\_RVs being sampled now has a new value (if
 accepted, see
 `here <https://github.com/pymc-devs/pymc3/blob/6d07591962a6c135640a3c31903eba66b34e71d8/pymc3/step_methods/compound.py#L27>`__
 and
-`here <https://github.com/pymc-devs/pymc3/blob/master/pymc3/step_methods/compound.py#L41>`__).
+`here <https://github.com/pymc-devs/pymc3/blob/main/pymc3/step_methods/compound.py#L41>`__).
```
OriolAbril: I'm not sure this is pointing to the right line, and it will have to be updated from time to time. I think ideally this would point to the documentation and the documentation would have the source code (so you can link to methods independently of the line they are in, for example https://arviz-devs.github.io/arviz/_modules/arviz/stats/stats.html#hdi.

OriolAbril (approved):
I think the comments should probably not be addressed here but taken into account at some point, maybe ported to an issue? cc @martinacantaro

docs/source/api/bounds.rst
michaelosthege:
```suggestion
  probability function.  See `this example notebook <https://docs.pymc.io/pymc-examples/examples/survival_analysis/weibull_aft.html>`_.
```

docs/source/developer_guide.rst
michaelosthege: The entire file has just 72 lines and the compound step docs are linked a few lines blow.
I guess we can just remove the line number.
```suggestion
`here <https://github.com/pymc-devs/pymc3/blob/main/pymc3/step_methods/compound.py>`__).
```

michaelosthege:
@OriolAbril I changed those two links and squash-merged. Didn't want to waste another hour of CI-compute energy ;)

PR #4741, opened 2021-06-05
Fix numpy bool deprecation warnings for v3

michaelosthege:
Cherry-picked the changes by @Nr90 from #4678.

PR #4745, opened 2021-06-06
Bump version to v3.11.3

canyon289:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,7 +11,7 @@ algorithms. Its flexibility and extensibility make it applicable to a
 large suite of problems.
 
 Check out the `getting started guide <http://docs.pymc.io/notebooks/getting_started>`__,  or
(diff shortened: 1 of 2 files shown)


michaelosthege (dismissed):

michaelosthege:
@canyon289 the same [compilation error](https://github.com/pymc-devs/pymc3/pull/4745/checks?check_run_id=2758429339#step:7:2663) appeared on #4741 

So it's probably an update to a so far unidentified dependency..

michaelosthege:
@canyon289 I reduced the installation instructions to a minimal set of dependencies. With the new `environment.yml` from the guide also BLAS stuff works and the tests that failed in the CI passes locally.
The trick was to include both `numba` and `scipy` in the dependencies at the same time.
Maybe someone can take this updated guide to fix the CI for the Windows tests?

twiecki:
What about release notes?

PR #4746, opened 2021-06-06
Refactored continuous distributions to v4 (AsymmetricLaplace, HalfStudentT, ExGaussian, Interpolated)

kc611:
Linked Issue: #4686 

This PR refactors the following distributions:

- AsymmetricLaplace
- HalfStudentT
-  ExGaussian
-  Interpolated

(Random Note: It's strange how git is showing that many file changes in the test file `pymc3/tests/test_distributions.py` when only difference I made is of few lines ?)

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -1587,6 +1587,26 @@ class Laplace(Continuous):
         )
 
 
+class AsymmetricLaplaceRV(RandomVariable):
+    name = "asymmetriclaplace"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0]
+    dtype = "floatX"
+    _print_name = ("AsymmetricLaplace", "\\operatorname{AsymmetricLaplace}")
+
+    @classmethod
+    def rng_fn(cls, rng, b, kappa, mu, size=None):
+        u = rng.uniform(size=size)
+        switch = kappa ** 2 / (1 + kappa ** 2)
+        non_positive_x = mu + kappa * np.log(u * (1 / switch)) / b
+        positive_x = mu - np.log((1 - u) * (1 + kappa ** 2)) / (kappa * b)
+        draws = non_positive_x * (u <= switch) + positive_x * (u > switch)
+        return draws
+
+
+asymmetriclaplace = AsymmetricLaplaceRV()
+
+
 class AsymmetricLaplace(Continuous):
     r"""
     Asymmetric-Laplace log-likelihood.
@@ -1622,49 +1642,23 @@ class AsymmetricLaplace(Continuous):
     --------
     `Reference <https://en.wikipedia.org/wiki/Asymmetric_Laplace_distribution>`_
     """
+    rv_op = asymmetriclaplace
 
-    def __init__(self, b, kappa, mu=0, *args, **kwargs):
-        self.b = at.as_tensor_variable(floatX(b))
-        self.kappa = at.as_tensor_variable(floatX(kappa))
-        self.mu = mu = at.as_tensor_variable(floatX(mu))
+    @classmethod
+    def dist(cls, b, kappa, mu=0, *args, **kwargs):
+        b = at.as_tensor_variable(floatX(b))
+        kappa = at.as_tensor_variable(floatX(kappa))
+        mu = mu = at.as_tensor_variable(floatX(mu))
 
-        self.mean = self.mu - (self.kappa - 1 / self.kappa) / b
-        self.variance = (1 + self.kappa ** 4) / (self.kappa ** 2 * self.b ** 2)
+        # mean = mu - (kappa - 1 / kappa) / b
+        # variance = (1 + kappa ** 4) / (kappa ** 2 * b ** 2)
 
         assert_negative_support(kappa, "kappa", "AsymmetricLaplace")
         assert_negative_support(b, "b", "AsymmetricLaplace")
 
-        super().__init__(*args, **kwargs)
+        return super().dist([b, kappa, mu], *args, **kwargs)
 
-    def _random(self, b, kappa, mu, size=None):
-        u = np.random.uniform(size=size)
-        switch = kappa ** 2 / (1 + kappa ** 2)
-        non_positive_x = mu + kappa * np.log(u * (1 / switch)) / b
-        positive_x = mu - np.log((1 - u) * (1 + kappa ** 2)) / (kappa * b)
-        draws = non_positive_x * (u <= switch) + positive_x * (u > switch)
-        return draws
-
-    def random(self, point=None, size=None):
-        """
-        Draw random samples from this distribution, using the inverse CDF method.
-
-        Parameters
-        ----------
-        point: dict, optional
-            Dict of variable values on which random values are to be
-            conditioned (uses default point if not specified).
-        size:int, optional
-            Desired size of random sample (returns one sample if not
-            specified).
-
-        Returns
-        -------
-        array
-        """
-        # b, kappa, mu = draw_values([self.b, self.kappa, self.mu], point=point, size=size)
-        # return generate_samples(self._random, b, kappa, mu, dist_shape=self.shape, size=size)
-
-    def logp(self, value):
+    def logp(value, b, kappa, mu):
         """
         Calculate log-probability of Asymmetric-Laplace distribution at specified value.
 
@@ -1678,12 +1672,12 @@ class AsymmetricLaplace(Continuous):
         -------
         TensorVariable
         """
-        value = value - self.mu
+        value = value - mu
         return bound(
-            at.log(self.b / (self.kappa + (self.kappa ** -1)))
-            + (-value * self.b * at.sgn(value) * (self.kappa ** at.sgn(value))),
-            0 < self.b,
-            0 < self.kappa,
+            at.log(b / (kappa + (kappa ** -1)))
+            + (-value * b * at.sgn(value) * (kappa ** at.sgn(value))),
+            0 < b,
+            0 < kappa,
         )
 
 
@@ -2759,6 +2753,21 @@ class Weibull(PositiveContinuous):
         )
 
 
+class HalfStudentTRV(RandomVariable):
+    name = "halfstudentt"
+    ndim_supp = 0
+    ndims_params = [0, 0]
+    dtype = "floatX"
+    _print_name = ("HalfStudentT", "\\operatorname{HalfStudentT}")
+
+    @classmethod
+    def rng_fn(cls, rng, nu, sigma, size=None):
+        return np.abs(stats.t.rvs(nu, sigma, size=size, random_state=rng))
+
+
+halfstudentt = HalfStudentTRV()
+
+
 class HalfStudentT(PositiveContinuous):
     r"""
     Half Student's T log-likelihood
@@ -2816,46 +2825,29 @@ class HalfStudentT(PositiveContinuous):
         with pm.Model():
             x = pm.HalfStudentT('x', lam=4, nu=10)
     """
+    rv_op = halfstudentt
 
-    def __init__(self, nu=1, sigma=None, lam=None, sd=None, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    @classmethod
+    def dist(cls, nu=1, sigma=None, lam=None, sd=None, *args, **kwargs):
+
         if sd is not None:
             sigma = sd
 
-        self.mode = at.as_tensor_variable(0)
+        nu = at.as_tensor_variable(floatX(nu))
         lam, sigma = get_tau_sigma(lam, sigma)
-        self.median = at.as_tensor_variable(sigma)
-        self.sigma = self.sd = at.as_tensor_variable(sigma)
-        self.lam = at.as_tensor_variable(lam)
-        self.nu = nu = at.as_tensor_variable(floatX(nu))
+        sigma = at.as_tensor_variable(sigma)
 
-        assert_negative_support(sigma, "sigma", "HalfStudentT")
-        assert_negative_support(lam, "lam", "HalfStudentT")
+        # mode = at.as_tensor_variable(0)
+        # median = at.as_tensor_variable(sigma)
+        # sd = at.as_tensor_variable(sigma)
+
         assert_negative_support(nu, "nu", "HalfStudentT")
+        assert_negative_support(lam, "lam", "HalfStudentT")
+        assert_negative_support(sigma, "sigma", "HalfStudentT")
 
-    def random(self, point=None, size=None):
-        """
-        Draw random values from HalfStudentT distribution.
+        return super().dist([nu, sigma], *args, **kwargs)
 
-        Parameters
-        ----------
-        point: dict, optional
-            Dict of variable values on which random values are to be
-            conditioned (uses default point if not specified).
-        size: int, optional
-            Desired size of random sample (returns one sample if not
-            specified).
-
-        Returns
-        -------
-        array
-        """
-        # nu, sigma = draw_values([self.nu, self.sigma], point=point, size=size)
-        # return np.abs(
-        #     generate_samples(stats.t.rvs, nu, loc=0, scale=sigma, dist_shape=self.shape, size=size)
-        # )
-
-    def logp(self, value):
+    def logp(value, nu, sigma):
         """
         Calculate log-probability of HalfStudentT distribution at specified value.
 
@@ -2869,9 +2861,8 @@ class HalfStudentT(PositiveContinuous):
         -------
         TensorVariable
         """
-        nu = self.nu
-        sigma = self.sigma
-        lam = self.lam
+
+        lam, sigma = get_tau_sigma(None, sigma)
 
         return bound(
             at.log(2)
@@ -2887,6 +2878,21 @@ class HalfStudentT(PositiveContinuous):
 
     def _distr_parameters_for_repr(self):
         return ["nu", "lam"]
+
+
+class ExGaussianRV(RandomVariable):
+    name = "exgaussian"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0]
+    dtype = "floatX"
+    _print_name = ("ExGaussian", "\\operatorname{ExGaussian}")
+
+    @classmethod
+    def rng_fn(cls, rng, mu, sigma, nu, size=None):
+        return rng.normal(mu, sigma, size=size) + rng.exponential(scale=nu, size=size)
+
+
+exgaussian = ExGaussianRV()
 
 
 class ExGaussian(Continuous):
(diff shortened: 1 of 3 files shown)


pymc3/distributions/continuous.py
```diff
@@ -2723,6 +2717,21 @@ def logcdf(value, alpha, beta):
...
 
+class HalfStudentTRV(RandomVariable):
+    name = "halfstudentt"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0]
+    dtype = "floatX"
+    _print_name = ("HalfStudentT", "\\operatorname{HalfStudentT}")
+
+    @classmethod
+    def rng_fn(cls, rng, nu, sigma, lam, sd, size=None):
```
ricardoV94: `lam` and `sd` are not being used, they should be removed, no?

pymc3/distributions/continuous.py
```diff
@@ -2780,46 +2789,28 @@ class HalfStudentT(PositiveContinuous):
...
-
-        Returns
-        -------
-        array
-        """
-        # nu, sigma = draw_values([self.nu, self.sigma], point=point, size=size)
-        # return np.abs(
-        #     generate_samples(stats.t.rvs, nu, loc=0, scale=sigma, dist_shape=self.shape, size=size)
-        # )
+        return super().dist([nu, sigma, lam, sd], *args, **kwargs)
```
ricardoV94: Only one default parametrization should go into `super().dist`, probably `[nu, sigma]`. You can check the already refactored t-student for an example.

pymc3/distributions/continuous.py
```diff
@@ -2780,46 +2789,28 @@ class HalfStudentT(PositiveContinuous):
...
 
-    def logp(self, value):
+    def logp(value, nu, sigma, lam, sd):
```
ricardoV94: Similarly, the `logp` should receive only one default parametrization (the same you define in the `dist` classmethod.

pymc3/distributions/continuous.py
```diff
@@ -2853,6 +2841,21 @@ def _distr_parameters_for_repr(self):
...
 
+class ExGaussianRV(RandomVariable):
+    name = "exgaussian"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0]
+    dtype = "floatX"
+    _print_name = ("ExGaussian", "\\operatorname{ExGaussian}")
+
+    @classmethod
+    def rng_fn(cls, rng, mu, sigma, nu, sd, size=None):
```
ricardoV94: Same here, `sd` should not be in here

pymc3/distributions/continuous.py
```diff
@@ -3836,27 +3835,27 @@ class Interpolated(BoundedContinuous):
...
+
+        lower = at.as_tensor_variable(x_points[0])
+        upper = at.as_tensor_variable(x_points[-1])
 
         interp = InterpolatedUnivariateSpline(x_points, pdf_points, k=1, ext="zeros")
         Z = interp.integral(x_points[0], x_points[-1])
 
-        self.Z = at.as_tensor_variable(Z)
+        Z = at.as_tensor_variable(Z)
         self.interp_op = SplineWrapper(interp)
```
ricardoV94: the `interp_op` will not be reached from the`logp`, you will have to define it in there.

pymc3/distributions/continuous.py
```diff
@@ -3902,7 +3882,7 @@ def logp(self, value):
         -------
         TensorVariable
         """
-        return at.log(self.interp_op(value) / self.Z)
+        return at.log(self.interp_op(value) / Z)
```
ricardoV94: A bit confusingly, we are not using `self` instances, the logp has to work as a standard function dependent only on it's inputs. I am referring to the `self.interp_op`

ricardoV94 (changes requested):
Thanks for opening the PR! I marked a couple of things than need to be changed. Let me know if you have any questions :)

Also, we will need the test in `test_distributions_random.py` to be refactored / added, and any tests that may be marked as `xfail` because of these distributions anywhere else on the test module.

pymc3/tests/test_distributions.py
```diff
@@ -1307,207 +1306,206 @@ def test_laplace_asymmetric(self):
...
-            {"alpha": R, "beta": Rplusbig},
-            lambda value, alpha, beta: sp.cauchy.logpdf(value, alpha, beta),
-        )
-        self.check_logcdf(
-            Cauchy,
-            R,
-            {"alpha": R, "beta": Rplusbig},
-            lambda value, alpha, beta: sp.cauchy.logcdf(value, alpha, beta),
-        )
+        def test_lognormal(self):
```
ricardoV94: This new indentation is (probably) wrong

pymc3/tests/test_distributions.py
```diff
@@ -2511,100 +2508,100 @@ def test_ex_gaussian_cdf_outside_edges(self):
...
+            self.check_logcdf(
+                Moyal,
+                R,
+                {"mu": R, "sigma": Rplusbig},
+                lambda value, mu, sigma: floatX(sp.moyal.logcdf(value, mu, sigma)),
+            )
 
     @pytest.mark.xfail(condition=(aesara.config.floatX == "float32"), reason="Fails on float32")
-    @pytest.mark.xfail(reason="Distribution not refactored yet")
+    @pytest.mark.xfail(reason="Test not yet refactored for new version")
```
ricardoV94: This will need to be refactored for this PR, otherwise we have no test that covers the `logp` method of the `Interpolated`, or do we?

ricardoV94:
> (Random Note: It's strange how git is showing that many file changes in the test file `pymc3/tests/test_distributions.py` when only difference I made is of few lines ?)

Could it be some auto-formatting on your end? I can try to check locally to see if it is just a GitHub confusion or if you actually indented those lines.

pymc3/tests/test_distributions.py
kc611: Yea, I plan on doing exactly that in this PR. It seems some problem with the arguments not being passed correctly.

kc611:
> Could it be some auto-formatting on your end? I can try to check locally to see if it is just a GitHub confusion or if you actually indented those lines.

Please do. The only changes it shows on my side are the few changed `x_fail` lines.

Edit: Never mind, Looks like VS Code doesn't take into account the indentation changes as "real" changes in git. Fixed it now.

pymc3/distributions/continuous.py
```diff
@@ -2844,7 +2832,6 @@ def logp(self, value):
             - 0.5 * at.log(nu * np.pi * sigma ** 2)
             - (nu + 1.0) / 2.0 * at.log1p(value ** 2 / (nu * sigma ** 2)),
             sigma > 0,
-            lam > 0,
```
kc611: I'm a bit skeptical about this particular parametrization, since that required removing this condition over here.(I don't know if this particular condition is important or not, though)

pymc3/distributions/continuous.py
```diff
@@ -2780,46 +2789,28 @@ class HalfStudentT(PositiveContinuous):
...
         if sd is not None:
             sigma = sd
 
-        self.mode = at.as_tensor_variable(0)
         lam, sigma = get_tau_sigma(lam, sigma)
-        self.median = at.as_tensor_variable(sigma)
-        self.sigma = self.sd = at.as_tensor_variable(sigma)
-        self.lam = at.as_tensor_variable(lam)
-        self.nu = nu = at.as_tensor_variable(floatX(nu))
+        mode = at.as_tensor_variable(0)
```
ricardoV94: everything that is not going into `dist` can be commented for the time being (and doesn't need the `as_tensor_variable` call either. This is related to sampler starting point logic that has not been refactored.

pymc3/distributions/continuous.py
ricardoV94: The lam sigma is a bit weird in that one can be converted from the other, even if the original is negative. You can obtain the lam again from sigma by  using the `get_tau_sigma` inside the `logp`. This way you can keep the lam check in the `bound`, but probably we should just open a separate issue to fix `get_tau_sigma`

pymc3/distributions/continuous.py
```diff
@@ -1621,49 +1641,23 @@ class AsymmetricLaplace(Continuous):
...
-        self.mu = mu = at.as_tensor_variable(floatX(mu))
+    @classmethod
+    def dist(cls, b, kappa, mu=0, *args, **kwargs):
+        b = at.as_tensor_variable(floatX(b))
+        kappa = at.as_tensor_variable(floatX(kappa))
+        mu = mu = at.as_tensor_variable(floatX(mu))
 
-        self.mean = self.mu - (self.kappa - 1 / self.kappa) / b
-        self.variance = (1 + self.kappa ** 4) / (self.kappa ** 2 * self.b ** 2)
+        mean = mu - (kappa - 1 / kappa) / b
```
ricardoV94: mean and variance can be commented out for the same reason as below. They are not being used by the current codebase

pymc3/distributions/continuous.py
```diff
@@ -2918,49 +2920,26 @@ class ExGaussian(Continuous):
...
         if sd is not None:
             sigma = sd
 
-        self.mu = mu = at.as_tensor_variable(floatX(mu))
-        self.sigma = self.sd = sigma = at.as_tensor_variable(floatX(sigma))
-        self.nu = nu = at.as_tensor_variable(floatX(nu))
-        self.mean = mu + nu
-        self.variance = (sigma ** 2) + (nu ** 2)
+        mu = mu = at.as_tensor_variable(floatX(mu))
+        ssigma = sd = sigma = at.as_tensor_variable(floatX(sigma))
```
ricardoV94: typo `ssigma`

pymc3/distributions/continuous.py
```diff
@@ -2918,49 +2920,26 @@ class ExGaussian(Continuous):
...
 
         if sd is not None:
             sigma = sd
 
-        self.mu = mu = at.as_tensor_variable(floatX(mu))
-        self.sigma = self.sd = sigma = at.as_tensor_variable(floatX(sigma))
-        self.nu = nu = at.as_tensor_variable(floatX(nu))
-        self.mean = mu + nu
-        self.variance = (sigma ** 2) + (nu ** 2)
+        mu = mu = at.as_tensor_variable(floatX(mu))
```
ricardoV94: leftover typo: `mu = mu`

pymc3/distributions/continuous.py
```diff
@@ -3786,6 +3759,31 @@ def logp(value, mu, sigma):
         )
 
 
+class InterpolatedRV(RandomVariable):
+    name = "interpolated"
+    ndim_supp = 0
+    ndims_params = [0, 0]
```
ricardoV94, line 3765: seems like the rng_fn needs four arguments, so `ndims_params = [0, 0, 0, 0]`

pymc3/distributions/continuous.py
```diff
@@ -3786,6 +3759,31 @@ def logp(value, mu, sigma):
...
+    dtype = "floatX"
+    _print_name = ("Interpolated", "\\operatorname{Interpolated}")
+
+    @classmethod
+    def rng_fn(cls, rng, p, pdf, cdf, x, size=None):
```
ricardoV94: Something is off with the `rng_fn` signature. `p` seems to be overwritten so it shouldn't be there, and if you want to pass the `cdf` already specified it should also be in the `dist` classmethod bellow and `ndim_params` shoud be 3 elements long.

pymc3/distributions/continuous.py
ricardoV94: I think `ndims_params` cannot be scalar (i.e., `0`), right? In that case it should be `ndims_params = [1, 1]`

pymc3/distributions/continuous.py
```diff
@@ -3836,28 +3834,24 @@ class Interpolated(BoundedContinuous):
...
+    rv_op = interpolated
 
-        super().__init__(lower=lower, upper=upper, *args, **kwargs)
+    @classmethod
+    def dist(cls, x_points, pdf_points, *args, **kwargs):
+        lower = at.as_tensor_variable(x_points[0])
+        upper = at.as_tensor_variable(x_points[-1])
 
-        interp = InterpolatedUnivariateSpline(x_points, pdf_points, k=1, ext="zeros")
         Z = interp.integral(x_points[0], x_points[-1])
```
ricardoV94, line 3889: `Z` is not being passed to `super().dist()` and as such won't reach the `logp`. It seems it is not necessary for the random draws so it's probably better to also specify it inside the `logp` from the `x_points` input.

pymc3/distributions/continuous.py
kc611: Ah, okay got it.

pymc3/distributions/continuous.py
```diff
@@ -2723,6 +2717,21 @@ def logcdf(value, alpha, beta):
         )
 
 
+class HalfStudentTRV(RandomVariable):
+    name = "halfstudentt"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0]
```
ricardoV94: `ndims_params` should be `[0, 0]`, i.e., the same length as the number of parameters

pymc3/distributions/continuous.py
```diff
@@ -2833,9 +2824,9 @@ def logp(self, value):
         -------
         TensorVariable
         """
-        nu = self.nu
-        sigma = self.sigma
-        lam = self.lam
+
+        lam, sigma = get_tau_sigma(None, sigma)
+        assert_negative_support(lam, "lam", "HalfStudentT")
```
ricardoV94: No need to add the assert here, it should just be in the `dist` classmethod

pymc3/distributions/continuous.py
```diff
@@ -3786,6 +3765,39 @@ def logp(value, mu, sigma):
...
+    ndim_supp = 0
+    ndims_params = [1, 1]
+    dtype = "floatX"
+    _print_name = ("Interpolated", "\\operatorname{Interpolated}")
+
+    @classmethod
+    def rng_fn(cls, rng, x, pdf, size=None):
+        interp = InterpolatedUnivariateSpline(x, pdf, k=1, ext="zeros")
+
+        Z = interp.integral(x[0], x[-1])
```
ricardoV94: Inside `rng_fn` there shouldn't be any symbolic operations. I didn't realize you need `Z` here as well.

You can either use scipy equivalents or, if easier, have everything that's needed for both `logp` and the `rng_fn` defined once during the `dist classmethod` and passed to `super` as additional inputs (and expected in both the `logp` and `rng_fn`. This way, the `rng_fn` will receive the non-symbolic arguments as inputs during normal sampling, while the `logp` method will receive the `symbolic` ones. 

Apologies if this sounds very confusing :D

kc611: I think I should probably refactor the test first, just so that I'll have an intuition of how this entire `Interpolated` thing should work.

ricardoV94: Yeah, you can also check it out in V3, and see how it works there (and in the test). That may help, since this is a very non-standard one

pymc3/distributions/continuous.py
```diff
@@ -3786,6 +3765,42 @@ def logp(value, mu, sigma):
         )
 
 
+def _interpolated_argcdf(p, pdf, cdf, x):
+    index = np.searchsorted(cdf, p) - 1
+    slope = (pdf[index + 1] - pdf[index]) / (x[index + 1] - x[index])
```
kc611, line 3806: It looks like we cannot directly access the index of `pdf` or `x` like this. (The `index` variable is an `Elemwise` during the initial graph construction and hence this fails.). What class/function will be a good substitute for this?

ricardoV94: I think the problem is that you are using symbolic operations inside the rng_fn. Everything there should be non-symbolic AFAIK

kc611: That indeed is the problem. What I wanted to know was what is the non-symbolic substitute for such operation?

 (If there isn't I'll try your suggestion from https://github.com/pymc-devs/pymc3/pull/4746#discussion_r646490859 )

ricardoV94: I think the symbolic ops used in here are all wrapping standard scipy function: https://github.com/pymc-devs/pymc3/blob/25eaa715018ed7add3ba1289b6ad068636e29e5e/pymc3/distributions/dist_math.py#L318

kc611: I tried fiddling around with `SplineWrapper` but I don't think that'll be of much use here since we are using different methods of `InterpolatedUnivariateSpline` like the `.integral()` and the `.antiderivate()` ( But can we actually use `SplineWrapper` like that ? )

Instead as you mentioned in https://github.com/pymc-devs/pymc3/pull/4746#discussion_r646490859 I just kept the non-symbolic operations in `dist` and forwarded `cdf_points` as an extra argument.

michaelosthege (approved):
LGTM
@ricardoV94 you make the call

pymc3/tests/test_distributions_random.py
```diff
@@ -1506,11 +1502,16 @@ def test_interpolated(self):
                 def ref_rand(size):
                     return st.norm.rvs(loc=mu, scale=sigma, size=size)
 
+                from pymc3.distributions.continuous import interpolated
```
ricardoV94: should the import be at the top?

pymc3/tests/test_distributions_random.py
```diff
@@ -270,7 +270,6 @@ class TestWald(BaseTestCases.BaseTestCase):
     params = {"mu": 1.0, "lam": 1.0, "alpha": 0.0}
 
 
-@pytest.mark.xfail(reason="This distribution has not been refactored for v4")
 class TestAsymmetricLaplace(BaseTestCases.BaseTestCase):
```
ricardoV94: These tests that rely on `BaseTestCases.BaseTestCase` should be removed and replaced with tests using the new `BaseTestDistribution`

ricardoV94 (changes requested):
Looks great so far. The random tests still need to be refactored to use the `BaseTestDistribution`

pymc3/tests/test_distributions_random.py
```diff
@@ -1497,7 +1494,6 @@ def ref_rand(size, mu, sigma):
 
         pymc3_random(pm.Moyal, {"mu": R, "sigma": Rplus}, ref_rand=ref_rand)
 
-    @pytest.mark.xfail(reason="This distribution has not been refactored for v4")
     @pytest.mark.xfail(condition=(aesara.config.floatX == "float32"), reason="Fails on float32")
     def test_interpolated(self):
```
ricardoV94: This one should be moved to it's own class. The reason for this is that after refactoring all distributions, the `TestScalarParameterSamples` where this one is will disappear, but this distribution is special enough that the test should be kept (i.e., we are not simply drawing from scipy / numpy distributions).

pymc3/tests/test_distributions_random.py
```diff
@@ -1257,7 +1256,6 @@ def ref_rand(size, mu, lam, alpha):
             ref_rand=ref_rand,
         )
 
-    @pytest.mark.xfail(reason="This distribution has not been refactored for v4")
```
ricardoV94, line 1260: These tests that are using a `ref_rand` should be removed as well, unless there is a good reason to test them against a reference. If we are just copying the same scipy/numpy expression we use in our `random Ops` the default tests in `BaseTestDistribution` should be used instead.

pymc3/tests/test_distributions_random.py
kc611: I just moved it to it's own class and named it `TestInterpolated`

ricardoV94 (commented):
Looks good. The last thing I would suggest is to add a `BaseTestDistribution` for the `Interpolated` just with the `check_rv_size` test enabled. Just to make sure `size` is working properly.

Otherwise, I think this PR is pretty much spot on. Specially wth the thorny `Interpolated`.

kc611:
Thanks @ricardoV94 for all your suggestions, (and especially for being patient with me in the start of this PR with all those silly mistakes. I guess my coffee didn't work quite right that particular day.)

Edit: Also (just a heads up) I'll be working on refactoring `KroneckerNormal` from the multivariate distributions next.

ricardoV94:
I am afraid the `Interpolated` is still problematic. Right now it works fine if you pass non-symbolic arguments, but it fails if the inputs are symbolic. This snippet raises an enigmatic `ValueError` from within the `scipy.interpolate` method:

```python
x_points = np.linspace(0, 10)
pdf_points = st.norm.pdf(x_points, loc=5, scale=0.1)
ae_x_points = at.as_tensor_variable(x_points)
ae_pdf_points = at.as_tensor_variable(pdf_points)
interp = pm.Interpolated.dist(ae_x_points, ae_pdf_points, size=(2, 4))
# ValueError: diff requires input that is at least one dimensional
```

`V3` also had the same limitation.

It seems a proper fix would require integrating the non-symbolic `InterpolatedUnivariateSpline` inside the symbolic `SplineWrapper` (or as a separate symbolic Op). The `logp` method making use of `x_points.data` is also quite hackish. For instance, it wouldn't work with a generic `at.arange` input.

~~If we can't be bothered at the moment, perhaps we should at least mention this limitation explicitly in the docstring and add an informative `TypeError` in `dist` if the input is symbolic. What do you think?~~

Edit: I see this line is in the docstrings:
```
Both parameters ``x_points`` and values ``pdf_points`` are not variables, but
plain array-like objects, so they are constant and cannot be sampled.
```

ricardoV94:
I pushed some changes that hopefully clarify a bit how the `Interpolated` is expected to work internally, ~~and also placed the contents of the `_argcdf` method directly inside the `rng_fn`.~~

I am okay with merging as is, and perhaps open a GH issue to discuss whether we want to make this distribution accept symbolic elements in the future.

ricardoV94 (approved):
Great work @kc611 and thanks for your patience. Let me know if you also agree with merging it, given my last comments.

kc611:
Sure, you can go ahead and merge this, and yes, we should probably open an issue for making a proper fix for integrating the non-symbolic `InterpolatedUnivariateSpline` inside the symbolic `SplineWrapper` (or as a separate symbolic Op). Otherwise we'll be stuck with this hackish approach for `Interpolated` which I'm sure we don't want for `v4`.

ricardoV94:
> Sure, you can go ahead and merge this, and yes, we should probably open an issue for making a proper fix for integrating the non-symbolic `InterpolatedUnivariateSpline` inside the symbolic `SplineWrapper` (or as a separate symbolic Op). Otherwise we'll be stuck with this hackish approach for `Interpolated` which I'm sure we don't want for `v4`. 

Do you mind creating that issue?

I'll wait until Monday to merge the PR in case anyone else wants to have a look.

PR #4758, opened 2021-06-08
Port Dirichlet Multinomial to v4

AlexAndorra:
See #4686 

- [x] Create a new `RV op` by inheriting from aesara `RandomVariable`, porting the code in the old `random` method to the `rng_fn` `classmethod` of the new `RandomVariable`. 
- [x] Add the new (instance) of the `RandomVariable` to the `rv_op` field in the respective PyMC3 distribution.
- [x] Port any initialization logic in the old `__init__` to the new `dist` `classmethod`
- [x] Refactor `logp` and `logcdf` to expect arguments in this order: `value, arg1, arg2, ... argn`
- [x]  Re-enable tests in `test_distributions.py`
- [x] Create new test in `test_distributions_random.py` and remove old `BaseTestCase` in that file.
- [x] Re-enable other tests that depended on this distribution in other test files.
- [x] Run linting / style checks

### DIFF
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -396,7 +396,6 @@ class Dirichlet(Continuous):
     a: array
         Concentration parameters (a > 0).
     """
-
     rv_op = dirichlet
 
     def __new__(cls, name, *args, **kwargs):
@@ -504,11 +503,6 @@ class Multinomial(Discrete):
         n = at.as_tensor_variable(n)
         p = at.as_tensor_variable(p)
 
-        # mean = n * p
-        # mode = at.cast(at.round(mean), "int32")
-        # diff = n - at.sum(mode, axis=-1, keepdims=True)
-        # inc_bool_arr = at.abs_(diff) > 0
-        # mode = at.inc_subtensor(mode[inc_bool_arr.nonzero()], diff[inc_bool_arr.nonzero()])
         return super().dist([n, p], *args, **kwargs)
 
     def logp(value, n, p):
@@ -518,7 +512,7 @@ class Multinomial(Discrete):
 
         Parameters
         ----------
-        x: numeric
+        value: numeric
             Value for which log-probability is calculated.
 
         Returns
@@ -534,6 +528,46 @@ class Multinomial(Discrete):
             at.all(at.ge(n, 0)),
             broadcast_conditions=False,
         )
+
+
+class DirichletMultinomialRV(RandomVariable):
+    name = "dirichlet_multinomial"
+    ndim_supp = 1
+    ndims_params = [0, 1]
+    dtype = "int64"
+    _print_name = ("DirichletMN", "\\operatorname{DirichletMN}")
+
+    def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
+        return default_shape_from_params(self.ndim_supp, dist_params, rep_param_idx, param_shapes)
+
+    @classmethod
+    def rng_fn(cls, rng, n, a, size):
+
+        if n.ndim > 0 or a.ndim > 1:
+            n, a = broadcast_params([n, a], cls.ndims_params)
+            size = tuple(size or ())
+
+            if size:
+                n = np.broadcast_to(n, size + n.shape)
+                a = np.broadcast_to(a, size + a.shape)
+
+            res = np.empty(a.shape)
+            for idx in np.ndindex(a.shape[:-1]):
+                p = rng.dirichlet(a[idx])
+                res[idx] = rng.multinomial(n[idx], p)
+            return res
+        else:
+            # n is a scalar, a is a 1d array
+            p = rng.dirichlet(a, size=size)  # (size, a.shape)
+
+            res = np.empty(p.shape)
+            for idx in np.ndindex(p.shape[:-1]):
+                res[idx] = rng.multinomial(n, p[idx])
+
+            return res
+
+
+dirichlet_multinomial = DirichletMultinomialRV()
 
 
 class DirichletMultinomial(Discrete):
@@ -569,92 +603,16 @@ class DirichletMultinomial(Discrete):
         Describes shape of distribution. For example if n=array([5, 10]), and
         a=array([1, 1, 1]), shape should be (2, 3).
     """
+    rv_op = dirichlet_multinomial
 
-    def __init__(self, n, a, shape, *args, **kwargs):
-
-        super().__init__(shape=shape, defaults=("_defaultval",), *args, **kwargs)
-
+    @classmethod
+    def dist(cls, n, a, *args, **kwargs):
         n = intX(n)
         a = floatX(a)
-        if len(self.shape) > 1:
-            self.n = at.shape_padright(n)
-            self.a = at.as_tensor_variable(a) if a.ndim > 1 else at.shape_padleft(a)
-        else:
-            # n is a scalar, p is a 1d array
-            self.n = at.as_tensor_variable(n)
-            self.a = at.as_tensor_variable(a)
 
-        p = self.a / self.a.sum(-1, keepdims=True)
+        return super().dist([n, a], **kwargs)
 
-        self.mean = self.n * p
-        # Mode is only an approximation. Exact computation requires a complex
-        # iterative algorithm as described in https://doi.org/10.1016/j.spl.2009.09.013
-        mode = at.cast(at.round(self.mean), "int32")
-        diff = self.n - at.sum(mode, axis=-1, keepdims=True)
-        inc_bool_arr = at.abs_(diff) > 0
-        mode = at.inc_subtensor(mode[inc_bool_arr.nonzero()], diff[inc_bool_arr.nonzero()])
-        self._defaultval = mode
-
-    def _random(self, n, a, size=None):
-        # numpy will cast dirichlet and multinomial samples to float64 by default
-        original_dtype = a.dtype
-
-        # Thanks to the default shape handling done in generate_values, the last
-        # axis of n is a dummy axis that allows it to broadcast well with `a`
-        n = np.broadcast_to(n, size)
-        a = np.broadcast_to(a, size)
-        n = n[..., 0]
-
-        # np.random.multinomial needs `n` to be a scalar int and `a` a
-        # sequence so we semi flatten them and iterate over them
-        n_ = n.reshape([-1])
-        a_ = a.reshape([-1, a.shape[-1]])
-        p_ = np.array([np.random.dirichlet(aa) for aa in a_])
-        samples = np.array([np.random.multinomial(nn, pp) for nn, pp in zip(n_, p_)])
-        samples = samples.reshape(a.shape)
-
-        # We cast back to the original dtype
-        return samples.astype(original_dtype)
-
-    def random(self, point=None, size=None):
-        """
-        Draw random values from Dirichlet-Multinomial distribution.
-
-        Parameters
-        ----------
-        point: dict, optional
-            Dict of variable values on which random values are to be
-            conditioned (uses default point if not specified).
-        size: int, optional
-            Desired size of random sample (returns one sample if not
-            specified).
-
-        Returns
-        -------
-        array
-        """
-        # n, a = draw_values([self.n, self.a], point=point, size=size)
-        # samples = generate_samples(
-        #     self._random,
-        #     n,
-        #     a,
-        #     dist_shape=self.shape,
-        #     size=size,
-        # )
-        #
-        # # If distribution is initialized with .dist(), valid init shape is not asserted.
-        # # Under normal use in a model context valid init shape is asserted at start.
-        # expected_shape = to_tuple(size) + to_tuple(self.shape)
-        # sample_shape = tuple(samples.shape)
-        # if sample_shape != expected_shape:
-        #     raise ShapeError(
-        #         f"Expected sample shape was {expected_shape} but got {sample_shape}. "
-        #         "This may reflect an invalid initialization shape."
-        #     )
-        #
-        # return samples
-
-    def logp(self, value):
+    def logp(value, n, a):
         """
         Calculate log-probability of DirichletMultinomial distribution
         at specified value.
@@ -668,13 +626,16 @@ class DirichletMultinomial(Discrete):
         -------
         TensorVariable
         """
-        a = self.a
-        n = self.n
-        sum_a = a.sum(axis=-1, keepdims=True)
+        if value.ndim >= 1:
+            n = at.shape_padright(n)
+            if a.ndim > 1:
+                a = at.shape_padleft(a)
 
+        sum_a = a.sum(axis=-1, keepdims=True)
         const = (gammaln(n + 1) + gammaln(sum_a)) - gammaln(n + sum_a)
         series = gammaln(value + a) - (gammaln(value + 1) + gammaln(a))
         result = const + series.sum(axis=-1, keepdims=True)
+
         # Bounds checking to confirm parameters and data meet all constraints
         # and that each observation value_i sums to n_i.
         return bound(
@@ -855,7 +816,7 @@ class Wishart(Continuous):
 
 
 def WishartBartlett(name, S, nu, is_cholesky=False, return_cholesky=False, initval=None):
-    R"""
+    r"""
     Bartlett decomposition of the Wishart distribution. As the Wishart
     distribution requires the matrix to be symmetric positive semi-definite
     it is impossible for MCMC to ever propose acceptable matrices.
--- a/pymc3/tests/test_distributions.py
+++ b/pymc3/tests/test_distributions.py
@@ -2272,6 +2272,16 @@ class TestMatchesScipy:
         sample = dist.eval()
         assert_allclose(sample, np.stack([vals, vals], axis=0))
 
+    def test_multinomial_zero_probs(self):
+        # test multinomial accepts 0 probabilities / observations:
+        value = aesara.shared(np.array([0, 0, 100], dtype=int))
+        logp = pm.Multinomial.logp(value=value, n=100, p=at.constant([0.0, 0.0, 1.0]))
+        logp_fn = aesara.function(inputs=[], outputs=logp)
+        assert logp_fn() >= 0
+
+        value.set_value(np.array([50, 50, 0], dtype=int))
+        assert np.isneginf(logp_fn())
+
     @pytest.mark.parametrize("n", [2, 3])
     def test_dirichlet_multinomial(self, n):
         self.check_logp(
--- a/pymc3/tests/test_distributions.py
+++ b/pymc3/tests/test_distributions.py
@@ -2273,7 +2273,6 @@ class TestMatchesScipy:
         assert_allclose(sample, np.stack([vals, vals], axis=0))
 
     @pytest.mark.parametrize("n", [2, 3])
-    @pytest.mark.xfail(reason="Distribution not refactored yet")
     def test_dirichlet_multinomial(self, n):
         self.check_logp(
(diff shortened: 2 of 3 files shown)


pymc3/distributions/multivariate.py
```diff
@@ -557,92 +590,26 @@ class DirichletMultinomial(Discrete):
...
-        n_ = n.reshape([-1])
-        a_ = a.reshape([-1, a.shape[-1]])
-        p_ = np.array([np.random.dirichlet(aa) for aa in a_])
-        samples = np.array([np.random.multinomial(nn, pp) for nn, pp in zip(n_, p_)])
-        samples = samples.reshape(a.shape)
+        p = a / a.sum(-1, keepdims=True)
 
-        # We cast back to the original dtype
-        return samples.astype(original_dtype)
+        return super().dist([n, a, p, shape], **kwargs)
```
ricardoV94: The first argument to `super().dist` should include the parameters that the `rv_op` is defined in terms of, in this case just `[n, a]`.

pymc3/distributions/multivariate.py
```diff
@@ -557,92 +590,26 @@ class DirichletMultinomial(Discrete):
...
         a=array([1, 1, 1]), shape should be (2, 3).
     """
+    rv_op = dirichlet_multinomial
 
-    def __init__(self, n, a, shape, *args, **kwargs):
-
-        super().__init__(shape=shape, defaults=("_defaultval",), *args, **kwargs)
-
+    @classmethod
+    def dist(cls, n, a, shape, *args, **kwargs):
```
ricardoV94: We have to be careful with shape. A user can define `size` or `shape` or `dims`, and the rv_ops should be defined in terms of `size` alone.  Therefore, any broadcasting operations should rely on `size` and be contained (if needed) inside the `rng_fn` and/or `logp` methods.

ricardoV94 (commented):
Thanks for opening up the PR! I left some comments in case it helps figuring things out. Multivariate distributions are a bit of PITA :/

pymc3/distributions/multivariate.py
ricardoV94: This other WIP PR might provide some clarification: https://github.com/pymc-devs/pymc3/pull/4731/files, or perhaps even better, the already refactored [Multinomial](https://github.com/pymc-devs/pymc3/blob/1be3ad705a3995262ff669151ff643d470d8f17d/pymc3/distributions/multivariate.py#L430) and [Dirichlet](https://github.com/aesara-devs/aesara/blob/3a3adaee3b8ceccff6d0208996f117fcd5e5eb6f/aesara/tensor/random/basic.py#L294)

ricardoV94: There are also some broadcast utils in here: https://github.com/aesara-devs/aesara/blob/main/aesara/tensor/random/utils.py

AlexAndorra: Yeah that didn't work anyway -- can't instantiate an RV with `shape`. I got rid of all this shape handling, to see how this goes for now 🤷‍♂️

pymc3/distributions/multivariate.py
```diff
@@ -524,6 +518,38 @@ def logp(value, n, p):
         )
 
 
+class DirichletMultinomialRV(RandomVariable):
+    name = "dirichlet_multinomial"
+    ndim_supp = 1
+    ndims_params = [0, 1]
```
AlexAndorra, line 536: Not sure about this

pymc3/distributions/multivariate.py
```diff
@@ -524,6 +518,38 @@ def logp(value, n, p):
...
+
+            res = np.empty(a.shape)
+            for idx in np.ndindex(a.shape[:-1]):
+                p = rng.dirichlet(a[idx])
+                res[idx] = rng.multinomial(n[idx], p)
+            return res
+        else:
+            # n is a scalar, a is a 1d array
+            p = rng.dirichlet(a)
+            return rng.multinomial(n, p, size=size)
```
AlexAndorra: Does this look better @ricardoV94 ?

pymc3/distributions/multivariate.py
ricardoV94: Seems fine, but you might need to copy these lines to have the op even initialize: https://github.com/aesara-devs/aesara/blob/3a3adaee3b8ceccff6d0208996f117fcd5e5eb6f/aesara/tensor/random/basic.py#L518-L521

After copying that I can run the following snippet successfully:

```python
from pymc3.distributions.multivariate import dirichlet_multinomial
dirichlet_multinomial(5, [1, 1, 1, 1])
dirichlet_multinomial(5, [1, 1, 1, 1]).eval()
# array([2, 2, 1, 0])
```

pymc3/distributions/multivariate.py
ricardoV94: I don't know where exactly is the problem (if there's any) but the sampling seems off:

```python
rng = aesara.shared(np.random.RandomState(123), borrow=True)
x = pm.DirichletMultinomial.dist(5, np.ones(4)/100, size=10, rng=rng)
x.eval()
array([[0, 0, 0, 5],
       [0, 0, 0, 5],
       [0, 0, 0, 5],
       [0, 0, 0, 5],
       [0, 0, 0, 5],
       [0, 0, 0, 5],
       [0, 0, 0, 5],
       [0, 0, 0, 5],
       [0, 0, 0, 5],
       [0, 0, 0, 5]])
```
I would expect more variation over rows. Here is what comes out from the V3 version
```python
pm.DirichletMultinomial.dist(n=5, a=np.ones(4)/100, shape=(4,)).random(size=10) 
array([[0., 5., 0., 0.],
       [0., 0., 0., 5.],
       [0., 0., 5., 0.],
       [0., 5., 0., 0.],
       [0., 5., 0., 0.],
       [0., 0., 5., 0.],
       [0., 0., 5., 0.],
       [1., 0., 0., 4.],
       [5., 0., 0., 0.],
       [0., 0., 5., 0.]])
```

pymc3/distributions/multivariate.py
```diff
@@ -524,6 +518,38 @@ def logp(value, n, p):
...
+                a = np.broadcast_to(a, size + a.shape)
+
+            res = np.empty(a.shape)
+            for idx in np.ndindex(a.shape[:-1]):
+                p = rng.dirichlet(a[idx])
+                res[idx] = rng.multinomial(n[idx], p)
+            return res
+        else:
+            # n is a scalar, a is a 1d array
+            p = rng.dirichlet(a)
```
ricardoV94: Here you probably need to pass size into `rng.dirichlet` to obtain multiple samples of `a` when size is present

pymc3/distributions/multivariate.py
AlexAndorra: Oooh thanks, that's a sneaky one! That was driving me crazy. It works for me too now 🎉

pymc3/distributions/multivariate.py
AlexAndorra: Thanks, that was very helpful! I fixed this implementation and it seems to work well now:
```python
rng = aesara.shared(np.random.RandomState(123), borrow=True)

# 1d case 
x = pm.DirichletMultinomial.dist(5, np.ones(4)/100, size=10, rng=rng)
x.eval()
array([[0, 0, 0, 5],
       [5, 0, 0, 0],
       [0, 0, 0, 5],
       [5, 0, 0, 0],
       [0, 0, 0, 5],
       [0, 0, 5, 0],
       [0, 0, 5, 0],
       [0, 5, 0, 0],
       [5, 0, 0, 0],
       [5, 0, 0, 0]])

# multidim case
x = pm.DirichletMultinomial.dist(np.array([5, 10]), np.array([np.ones(4)/100, np.array([1, 1, 1, 100])]), size=10, rng=rng)
x.eval()
array([[[ 4,  0,  1,  0],
        [ 0,  1,  0,  9]],

       [[ 0,  0,  5,  0],
        [ 0,  0,  0, 10]],

       [[ 0,  0,  0,  5],
        [ 0,  1,  0,  9]],

       [[ 0,  0,  0,  5],
        [ 1,  0,  0,  9]],

       [[ 0,  0,  0,  5],
        [ 0,  0,  0, 10]],

       [[ 0,  0,  0,  5],
        [ 0,  0,  0, 10]],

       [[ 0,  0,  0,  5],
        [ 0,  0,  0, 10]],

       [[ 5,  0,  0,  0],
        [ 0,  0,  0, 10]],

       [[ 0,  5,  0,  0],
        [ 0,  0,  0, 10]],

       [[ 0,  0,  0,  5],
        [ 1,  0,  1,  8]]])

File: pymc3/distributions/multivariate.py (hunk 5 above)
Comment (AlexAndorra) [reply pymc3/distributions/multivariate.py]: Thanks, fixed

File: pymc3/tests/test_distributions.py [hunk 6]
```diff
@@ -2256,14 +2255,13 @@ def test_dirichlet_multinomial(self, n):
...
     def test_dirichlet_multinomial_matches_beta_binomial(self):
         a, b, n = 2, 1, 5
         ns = np.arange(n + 1)
-        ns_dm = np.vstack((ns, n - ns)).T  # covert ns=1 to ns_dm=[1, 4], for all ns...
-        bb_logp = logpt(pm.BetaBinomial.dist(n=n, alpha=a, beta=b), ns).tag.test_value
+        ns_dm = np.vstack((ns, n - ns)).T  # convert ns=1 to ns_dm=[1, 4], for all ns...
+        bb_logp = logpt(pm.BetaBinomial.dist(n=n, alpha=a, beta=b, size=2), ns).tag.test_value
         dm_logp = logpt(
-            pm.DirichletMultinomial.dist(n=n, a=[a, b], size=(1, 2)), ns_dm
+            pm.DirichletMultinomial.dist(n=n, a=[a, b], size=2), ns_dm
```
AlexAndorra: - If I don't put a `size` of at least 2, I get a `TypeError: Cannot convert Type TensorType(int64, vector) (of Variable TensorConstant{[0 1 2 3 4 5]}) into Type TensorType(int64, scalar). You can try to manually convert TensorConstant{[0 1 2 3 4 5]} into a TensorType(int64, scalar).` Is that expected?
- When I do use a `size` > 1, I get a `AttributeError: 'ValidatingScratchpad' object has no attribute 'test_value'`, because the variable is `None`, which makes sense with this `UserWarning` that appears in the console when calling `logpt`: `/Users/alex_andorra/tptm_alex/pymc3/pymc3/aesaraf.py:357: UserWarning: No value variable found for dirichlet_multinomial_rv.out; the random variable will not be replaced.`. I don't understand why that happens though? 🤔

michaelosthege: I am unfamiliar with these distributions, but if `size=1` vs. `size=2` makes a difference, could it be that there's some unexpected downsizing of `(1,)` → `()` involved?

AlexAndorra: From what I understand, `logpt` is expecting that the first and second args have the same type. What I don't understand is why the `UserWarning` appears in the first place: the second arg is `rv_values` and is informed by `ns` and `ns_dm`. 
I'm missing something and probably need @ricardoV94 or @brandonwillard knowledge of Aesara

ricardoV94: You don't need to worry about the `UserWarning` at least. It's just that the logpt is tailored to be used within a Model object. The "proper" steps to call it would be:
```python
bb = pm.BetaBinomial.dist(n=n, alpha=a, beta=b, size=2)
bb_value = bb.type()  # This will be the variable that corresponds to the values of the bb in logp
bb.tag.value_var = bb_value  # the missing step that crates the warning
bb_logp = logpt(bb, {bb: bb_value}).eval({bb_value: ns})
# Can also just do the following, although that's not it's main goal:
bb_logp = logpt(bb, ns).eval()
```
All of these steps happen behind the scenes when you create the distribution inside a `Model`.

I think @brandonwillard will put an option to suppress these warnings when calling logpt directly

AlexAndorra: Oh ok, your comment seems very useful thanks! I'll try that out ASAP and keep you posted 😉 
I still have 3 tests failing 🤞

AlexAndorra: Thanks @ricardoV94, this is updated now

pymc3/tests/test_distributions.py
```diff
@@ -2302,56 +2306,52 @@ def test_dirichlet_multinomial_vec(self):
...
         a = np.array([0.2, 0.3, 0.5])
         ns = np.array([10, 11])
 
         with Model() as model:
-            DirichletMultinomial("m", n=ns, a=a, size=vals.shape)
+            pm.DirichletMultinomial("m", n=ns, a=a)
 
         assert_almost_equal(
             sum([dirichlet_multinomial_logpmf(val, n, a) for val, n in zip(vals, ns)]),
             model.fastlogp({"m": vals}),
```
AlexAndorra, line 2354: I don't understand why yet, but `model.fastlogp({"m": vals})` is evaluating to `-inf` (whereas `model.m.eval({model.m: vals})` works: `array([[2, 4, 4], [4, 3, 4]])`), which of course makes this test fail.
Interestingly, `test_dirichlet_multinomial_vec` (the test above) does pass 🤔 
Does this ring any bell @ricardoV94 ?

ricardoV94: The first eval is a logp computation while the second is giving a random draw, no? I'll try and check it locally.

AlexAndorra: Yeah but why is the logp `-inf` when evaluated on `vals` (`model.fastlogp({"m": vals})`)? I don't see a problem with `vals`

ricardoV94: Probably one of the bounds in the `logp` is being triggered. You can try to disable one at a time and find the culprit.

AlexAndorra: Mmmh ok, sneaky! I'll take a look at the logp and keep you posted 😉

AlexAndorra:
This is ready for review and merge @ricardoV94 -- thanks a lot for your help 🍾 
The rebase on `main` was actually a very bad idea -- that picked up a lot of commits and changes that are not mine; sorry about that 🤦‍♂️ I'll merge `main` for the next PR...

ricardoV94:
> This is ready for review and merge @ricardoV94 -- thanks a lot for your help champagne
> The rebase on `main` was actually a very bad idea -- that picked up a lot of commits and changes that are not mine; sorry about that man_facepalming I'll merge `main` for the next PR...

I think you will have to make another PR or clean the git history to fix the commit issues

michaelosthege:
Looks like you rebased with some weird settings that preserved the chronological order of commits?
I hope you can copy/paste your changes onto a new branch?

brandonwillard:
> The rebase on `main` was actually a very bad idea -- that picked up a lot of commits and changes that are not mine; sorry about that  I'll merge `main` for the next PR...

It looks like this branch **wasn't** rebased onto `main`; that's the problem.

Also, there's no need to create a new branch/PR, especially since that could easily involve much more work.  If you rebase onto `main` from the branch's current state, it will stop at the commits that conflict and you can decide whether or not to keep them.  Given the duplicate commits in this latter part of this branch, it looks like you can check out the changes labeled "ours" and drop the most recent duplicate commits.

brandonwillard:
I just performed the rebase locally and checked out only "our" changes on each conflict; the result has an empty diff when compared with this upstream branch (i.e. the aforementioned rebase onto `main` leaves the code in the same state it's currently in, while also fixing the bad history).

I can push this rebase if you'd like.

AlexAndorra:
Ah yes that'd be great, thanks Brandon!

AlexAndorra:
Ah yes that'd be great, thanks @brandonwillard ! Tell me if I need to do anything else

AlexAndorra:
Magic 🧙‍♂️ Thanks @brandonwillard !
I also fixed a last failing test (`test_batch_dirichlet_multinomial`) so this should _really_ be ready for review and merge now @ricardoV94 🍾

pymc3/distributions/multivariate.py
```diff
@@ -1433,12 +1398,7 @@ def logp(self, x):
         result = _lkj_normalizing_constant(eta, n)
         result += (eta - 1.0) * at.log(det(X))
         return bound(
-            result,
-            X >= -1,
-            X <= 1,
-            matrix_pos_def(X),
-            eta > 0,
-            broadcast_conditions=False,
+            result, X >= -1, X <= 1, matrix_pos_def(X), eta > 0, broadcast_conditions=False
```
```suggestion
            result, 
            X >= -1, 
            X <= 1, 
            matrix_pos_def(X), 
            eta > 0, 
            broadcast_conditions=False,
```
The comma after the last item keeps the pre-commit happy with multilines

pymc3/tests/test_distributions.py
```diff
@@ -166,10 +166,7 @@ def __mul__(self, other):
             )
         except TypeError:
             return Domain(
-                [v * other for v in self.vals],
-                self.dtype,
-                (self.lower, self.upper),
-                self.shape,
+                [v * other for v in self.vals], self.dtype, (self.lower, self.upper), self.shape
```
ricardoV94:
```suggestion
                [v * other for v in self.vals],
                self.dtype,
                (self.lower, self.upper),
                self.shape,
```

ricardoV94:
Your pre-commit seems to have changed a lot of lines that were not related to the PR, which makes it difficult to review the changes. 

Having an explicit comma after the last item in a multi-line call / tuple prevents it from auto-collapsing everything into a single line.

AlexAndorra:
WHy do you want to prevent it from being on a single line if it's passing black?

michaelosthege:
> WHy do you want to prevent it from being on a single line if it's passing black?

These lines are likely to cause git conflicts, because the same files are edited in other PRs as well. I don't understand why this happened in the first place though.

Apart from that (and the git conflict which looks reasonably easy to fix) how far away from the finish line is this PR ?

ricardoV94:
> WHy do you want to prevent it from being on a single line if it's passing black?

Besides what @michaelosthege said, it's an issue that this was an accidental change. The next person could come around and introduce the multiline fomat again (which also passes black) and we would go back and forth over time without need, just complicating reviews and the git history.

For example most of the same tests in those files keep the multiline format. If we want one style over the other it should be enforced automatically (and consistently). 

Also, I think these multilines are easier to read than the collapsed format, since most don't use kwargs.

AlexAndorra:
>  I don't understand why this happened in the first place though.

Yeah me neither, I didn't change anything manually -- formatting was done automatically by pre-commit locally.

> Apart from that, how far away from the finish line is this PR ?

It's done @michaelosthege. Once tests pass and conflicts are resolved, it can be merged.

> Also, I think these multilines are easier to read than the collapsed format, since most don't use kwargs.

I agree with all that. I just don't know how these changes appeared in the first place. If they were introduced by pre-commit, it seems to me there is a low chance they'll be changed again in a further PR.

ricardoV94:
> I agree with all that. I just don't know how these changes appeared in the first place. If they were introduced by pre-commit, it seems to me there is a low chance they'll be changed again in a further PR.

I don't know either. But looking at the black github, many issues seem related to this type of trailing commas...

ricardoV94:
I cleaned up the history and force-pushed here. Checking if the tests pass or if I screwed up something

AlexAndorra:
Wasn't the history already cleaned by Brandon?
In any case, all tests are green now, so feel free to review and merge 🥳

michaelosthege (approved):

ricardoV94:
> Wasn't the history already cleaned by Brandon?

I reverted the accidental formatting issues

ricardoV94:
Thanks @AlexAndorra this one was tough! :)

PR #4760, opened 2021-06-09
Move content of distributions.special into distributions.dist_math

### LINKED ISSUE #4756: Move content of distributions.special into distributions.dist_math
There are currently only two functions in [that file](https://www.github.com/pymc-devs/pymc3/tree/main/pymc3%2Fdistributions%2Fspecial.py) and they seem to be in nature quite similar to some in the larger [dist_math file](https://www.github.com/pymc-devs/pymc3/tree/main/pymc3%2Fdistributions%2Fdist_math.py). Also there is a single test in `test_special_functions.py` that could be moved to `test_dist_math.py`

themrzmaster:
closes #4756 
@ricardoV94


ricardoV94:
The failing tests seem to be unrelated:
* `pymc3/tests/test_ode.py::TestDiffEqModel::test_vector_ode_2_param` might just need to specify the initval of the model distributions to ensure the initial point `logp` is finite (similar to this PR https://github.com/pymc-devs/pymc3/pull/4707/files, except `testval` is now called `initval`).
* The `t-test` logcdf is flaky at the moment (it has failed in other PRs). For the time being feel free to comment out this section with a reminder `TODO` to be reenabled, linking to a new GH Issue: 
https://github.com/pymc-devs/pymc3/blob/4de95c3e8e65d8897ee48017f735418e8181c416/pymc3/tests/test_distributions.py#L1359-L1365

I am coming back to fix it in this PR #4736

pymc3/distributions/dist_math.py
```diff
@@ -30,11 +30,11 @@
 from aesara.scalar import UnaryScalarOp, upgrade_to_float_no_complex
 from aesara.scan import until
 from aesara.tensor.elemwise import Elemwise
+from aesara.tensor.math import gammaln
```
ricardoV94: I don't think we need to import this separately anymore. It's enough to use `at.gammaln` directly in the expression inside `multigammaln`

ricardoV94: For the `distributions/multivariate.py` where it is used a lot you can have a `from aesara.tensor import gammaln` directly there.

pymc3/tests/test_dist_math.py
```diff
@@ -16,7 +16,9 @@
 import numpy as np
 import numpy.testing as npt
 import pytest
+import scipy.special as ss
```
ricardoV94: I would change `as ss` to `as sp`, which I think is a more conventional alias that we use in the rest of the library. I know this is how it was before, but seems more appropriate.

ricardoV94 (commented):
I left a couple of minor comments. Let me know if you agree. Otherwise looks fine :)

pymc3/tests/test_distributions.py
```diff
@@ -1356,13 +1356,16 @@ def test_t(self):
             lambda value, nu, mu, lam: sp.t.logcdf(value, nu, mu, lam ** -0.5),
             n_samples=10,  # relies on slow incomplete beta
         )
+        # TODO: reenable when PR #4736 is merged
+        """
```
michaelosthege, line 1360: A reminder issue would be good. Or @ricardoV94 you rebase the other PR after we merge it and then you can enable it right there.

ricardoV94 (approved):
Looks good!

pymc3/tests/test_dist_math.py
```diff
@@ -16,7 +16,7 @@
 import numpy as np
 import numpy.testing as npt
 import pytest
-import scipy.special as st
+import scipy.special as sp
```
michaelosthege: There's just one occurrence of this module in this file.
IMHO it would be more readable to just `import scipy.special` without any aliasing.

michaelosthege (approved):

pymc3/tests/test_distributions.py
ricardoV94: I'll do that. Linking it here for now #4736

michaelosthege (approved):

ricardoV94:
Thanks @themrzmaster!

PR #4761, opened 2021-06-10
Switch to pydata sphinx theme

OriolAbril:
This PR changes the theme to pydata sphinx theme and also fixes many long standing issues
with the documentation like the search function not working, the non-existance of tables of contents
or adding links to the source code. Some of these fixes are thanks to the theme change,
others are somewhat independent.

Pros of the PR:
* We don't have to maintain a custom theme with lots of css, pure html and javascript.
  This is a particularly important pro because we were not maintaining the theme, so it
  progressively broke over time.
* Automatic tables of contents! The left sidebar has the section level table of contents, the right
  sidebar the page level table of contents. Moreover, the left table of contents is foldable,
  so with the right reorganization each notebook will have a notebook level table of contents and
  a section (examples or tutorials) level toc which is foldable by sections.
* Search bar works again
* Links to source code from API docs (theme independent)
* Looks better IMO

Cons/TODOs:
* We have to restructure the documentation a bit
  * pymc-examples -> I think we should move away from javascript and use pure rST/MyST with
    [sphinx_panels](https://sphinx-panels.readthedocs.io/en/latest/#card-layout)
  * learn page: more or less the same
  * A lot of the pure html will need to be updated because we are changing the bootstrap major
    version docs are built on -> this is also a pro because couples with cool extension like sphinx
    panels means we won't have to care about bootstrap anymore, everything will be updated by these
    extensions


OriolAbril:
I have converted the learn.rst file (which was only raw html) to myst and kept most of its appereance intact thanks to sphinx-panels extension. The new file is shorter in both number and lenght of lines and easier to read. 

<details><summary>Here is how it looks:</summary>

image
image
image

</details>

OriolAbril:
I think we can merge like this and fix the pages that still need some work between now and the first 4.0 release, in a follow-up or multiple follow-up PRs.

ricardoV94 (approved):

PR #4682, opened 2021-05-10
Remove GLM submodule in v4

farhanreynaldo:
This PR closes https://github.com/pymc-devs/pymc3/issues/4681, which remove GLM submodule and its test, in favor to Bambi.

PR #4764, opened 2021-06-11
Multioutput gp.Latent

bwengals:
This PR adds support for multi-output GPs, or matrix instead of just vector `y` inputs.   There have been a few requests for this on discourse, most recently [this one](https://discourse.pymc.io/t/the-predict-function-of-gp/7554/5).

Not sure yet, but I think that for some GP implementations, vector `y` will need have shape `(n, 1)` instead of `(n, )`, in the same way the `X` inputs do, which would be a breaking change (if there's no way to broadcast it automatically in aesara?).

At this point, the very basics of `gp.Marginal` work with matrix `Y`, but none of the other implementations do.  

- Would it be better to have separate PRs to add this functionality to each type of GP, or to have one large PR that addresses all implementations at the same time?
- Should this point towards `main` or `v3` or `v4`?


fonnesbeck:
`v4` has been merged, so definitely for `main`

Nothing wrong with doing marginal GP first, and then latent in a separate PR, but this isnt core functionality, so it can certainly wait if you'd prefer.

pymc3/glm/families.py
```diff
@@ -36,7 +36,7 @@ def __call__(self, x):
 
 
 identity = Identity()
-logit = at.nnet.sigmoid
+logit = at.sigmoid
```
twiecki: Confused why this file is here after https://github.com/pymc-devs/pymc3/pull/4682.

pymc3/gp/gp.py
```diff
@@ -15,7 +15,7 @@
 import functools
 import warnings
 
-import aesara.tensor as at
+import aesara.tensor as aet
```
twiecki: `at` is correct.

bwengals: gotcha, happy to flip it back

pymc3/glm/families.py
bwengals: weird, I'll double check where exactly my main is at

bwengals:
@fonnesbeck sounds good.  I think I'll try and keep them separate then if that's alright so the PRs are smaller (fixed the PR title).

michaelosthege:
Looking through some old PRs... @bwengals do you want to get back to this one?
The GP module needs a some refactoring (#5053) as part of the [4.0.0-beta2 milestone](https://github.com/pymc-devs/pymc/milestone/25) and any help there would be greatly appreciated!

bwengals:
@michaelosthege thanks for the nudge.  Yes, I'll prioritize #5035 and this for 4.0.  It's important to get this in

ricardoV94:
@bwengals Is this PR still relevant / being pursued?

bwengals:
Ah thanks for the nudge @ricardoV94, yes I need to finish this up.  Can we keep it open?  It shouldn't be too much work either I just need to do it

ricardoV94:
> Ah thanks for the nudge @ricardoV94, yes I need to finish this up. Can we keep it open? It shouldn't be too much work either I just need to do it

Of course :)

bwengals:
Gonna close, @danhphan is working on improvements to this

Issue #4490, opened 2021-02-27
Non-reproducible random draws with `sample_prior_predictive` and set random seed


alcrene:
If I have a code file with the following:

```python
# non_reproducible.py
import pymc3 as pm
import theano.tensor as tt

with pm.Model() as model:
    pm.Normal('a')
    pm.Normal('b')
    pm.Normal('c')
    pm.Normal('d')

pp = pm.sample_prior_predictive(samples=1, model=model,
                                random_seed=100)
print({k:pp[k] for k in sorted(pp)})
```
and run it multiple times, I get different results. (In the example below, values for `a` and `c` are swapped.)
```bash
python non_reproducible.py
# {'c': array([-1.74976547]), 'b': array([0.3426804]), 'a': array([1.1530358]), 'd': array([-0.25243604])}
python non_reproducible.py
# {'a': array([-1.74976547]), 'b': array([0.3426804]), 'c': array([1.1530358]), 'd': array([-0.25243604])}
```
In my opinion this defeats the whole purpose of being able to set the random seed.
(Note: it is essential to run the code in different sessions python sessions to see the effect.)

The reason is that the function [uses `set` to get a unique set of variable names](https://github.com/pymc-devs/pymc3/blob/47b36586676a0af6505f376afff2193ab5a5ed84/pymc3/sampling.py#L1935), and the random draws depend on the  (undefined) ordering of the returned set.

A simple fix might be to sort the variables names before sampling, as I've done [here](https://github.com/alcrene/pymc3/commit/c0753dbb0a7de6d0af0711be0223c80ecd7a5432). This has the side-effect of making the result independent of the order in which variables are specified, which I think is nice.

I'm not proposing my fix as a PR because figuring out and testing consistency with other functions goes beyond my current needs. For example, `sample_posterior_predictive` is (I think) reproducible given a random seed, because it just takes the list of variable names. But it likely _does_ depend on the order in which variables are specified.

## Versions and main components

* PyMC3 Version: master ([27 Feb 2021](https://github.com/pymc-devs/pymc3/tree/47b36586676a0af6505f376afff2193ab5a5ed84))
* Theano Version: Theano-PyMC 1.1.2
* Python Version: 3.8.6
* Operating system: OpenSUSE LEAP 15.2
* How did you install PyMC3: pip  (dependencies with conda)

ricardoV94:
Thanks for your detailed input. It is really helpful. 

[...]

Issue #4553, opened 2021-03-19
Do not force lower precision in intX


ferrine:
## Description of your problem

**Please provide a minimal, self-contained, and reproducible example.**
See implementation here
https://github.com/pymc-devs/pymc3/blob/c1efb7ad51e34e6bd8b292ef226a7e191adbdb82/pymc3/theanof.py#L92
for float32 precision, it converts values to int16
for float64 precision, it converts values to int32
But the implementation of Binomial distributions
https://github.com/pymc-devs/pymc3/blob/c1efb7ad51e34e6bd8b292ef226a7e191adbdb82/pymc3/distributions/multivariate.py#L714
https://github.com/pymc-devs/pymc3/blob/c1efb7ad51e34e6bd8b292ef226a7e191adbdb82/pymc3/distributions/discrete.py#L103
Does not assume a float is converted to int, thus converting ints in this way may overflow (I ran exactly into this issue)

**Please provide the full traceback.**
The referenced code is enough to spot the problem

**Please provide any additional information below.**
The issue is also relevant to pymc v4

## Versions and main components

* PyMC3 Version: 3.11.2
* Aesara Version: None
* Python Version: 3.8.9
* Operating system: Archlinux
* How did you install PyMC3: pip

ricardoV94:
>  Does not assume a float is converted to int, thus converting ints in this way may overflow (I ran exactly into this issue)

What do you mean by does not assume?

ferrine:
It converts presumably int parameters but takes float precision in account

ferrine:
The correct behaviour should convert int parameters to RV.dtype that user has control over

ricardoV94:
I think @lucianopaz introduced these methods. 

He will probably have a better idea of the rationale, but IIRC it had to do with ensuring int * float operations downstream worked predictably?

https://github.com/pymc-devs/pymc3/issues/4279#issuecomment-737044970

ricardoV94:
But are you saying there is no point in converting to intX at all in these specific distributions? 

If that's what you mean, I agree.

lucianopaz:
@ricardoV94, just to give some context, this was introduced [here](https://github.com/pymc-devs/pymc3/pull/3300) and it was intended to solve problems that could happen when a value they should be a float was an int, and viceversa.

[...]

PR #4769, opened 2021-06-14
Bring back `SMC` and allow `prior_predictive_sampling` to return transformed values

ricardoV94:
SMC was broken after the refactoring because it starts with a `prior_predictive_sampling` call to set up the particles positions, expecting it to return also transformed values. I extended `prior_predictive` to return transformed values if (and only if) transformed variables are explicitly passed in the optional `var_names` argument. For reference, in `v3` transformed variables were returned by default. If anyone has a strong opinion about the old default let me know!

Tests were added for this as well as for the now stale issue #4490 

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -13,11 +13,13 @@ jobs:
         floatx: [float32, float64]
         test-subset:
         # Tests are split into multiple jobs to accelerate the CI.
+        # Different jobs should be organized to take approximately the same
+        # time to complete (and not be prohibitely slow)
         #
         # How this works:
         # 1st block: Only passes --ignore parameters to pytest.
         # → pytest will run all test_*.py files that are NOT ignored.
-        # Other blocks: Only pass paths to test files.
+        # Subsequent blocks: Only pass paths to test files.
         # → pytest will run only these files
         #
         # Any test that was not ignored runs in the first job.
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -30,7 +30,6 @@ jobs:
             --ignore=pymc3/tests/test_modelcontext.py
             --ignore=pymc3/tests/test_parallel_sampling.py
             --ignore=pymc3/tests/test_profile.py
-            --ignore=pymc3/tests/test_smc.py
             --ignore=pymc3/tests/test_step.py
             --ignore=pymc3/tests/test_tuning.py
             --ignore=pymc3/tests/test_types.py
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -7,7 +7,8 @@
 - The GLM submodule has been removed, please use [Bambi](https://bambinos.github.io/bambi/) instead.
 - The `Distribution` keyword argument `testval` has been deprecated in favor of `initval`.
 - `pm.sample` now returns results as `InferenceData` instead of `MultiTrace` by default (see [#4744](https://github.com/pymc-devs/pymc3/pull/4744)).
-- ...
+- `pm.sample_prior_predictive` no longer returns transformed variable values by default. Pass them by name in `var_names` if you want to obtain these draws (see [4769](https://github.com/pymc-devs/pymc3/pull/4769)).
+...
 
 ### New Features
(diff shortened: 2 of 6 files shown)


junpenglao (approved):
LGTM

michaelosthege (approved):

michaelosthege (changes requested):
These failing tests look systematic. Some kind of dytpe problem..

ricardoV94:
> These failing tests look systematic. Some kind of dytpe problem..

Definitely. One of the SMC tests is failing in float32 when there is a discrete variable:

https://github.com/pymc-devs/pymc3/blob/57e4f5a177c98a928c3e590800d1a17af4237b50/pymc3/tests/test_smc.py#L67-L72

 It happens in the `join_nonshared_inputs` which concatenates all the raveled variables, leading to an upcast to `float64` when there are both discrete and continuous variables (since discretes are `int64`).

https://github.com/pymc-devs/pymc3/blob/f7d460212c0539e6c7a7ae394a3f2de6416068c7/pymc3/aesaraf.py#L604-L608

The problem then is that the input can no longer be of type `float32 / floatX`. I can make the test pass by just wrapping `joined` in a `pm.aesaraf.floatX`, confirming the problem lies there. I don't think this is a great fix though...

This could be a problem in other areas of the codebase that make use of this function. I've seen it in `metropolis.py`, `mlda.py` and `pgbart.py`.

Edit: Possibly related to https://github.com/pymc-devs/pymc3/issues/4553

Edit2: Link to the failing test: https://github.com/pymc-devs/pymc3/runs/2821726372?check_suite_focus=true#step:7:599
(I canceled the last workflow, I had just pushed a rebase and some minor comments, so the tests would still fail)

ricardoV94:
Also, more in general, within SMC we are treating discrete variables as continuous (e.g., in the proposal). Are we comfortable with this @aloctavodia? I know that the `logp` methods deal fine with float inputs, but it still feels strange to feed them non-rounded values.

ricardoV94:
I temporarily disabled `trust_input` to pass the tests. I tried a bunch of things (mostly with `clone_replace` + `casting`) to allow the `prior_logp_func` function to accept `float32` inputs  in models with discrete variables, without success. 

Edit: I restricted this change to when absolutely needed, and issued an informative `UserWarning` in those cases.

junpenglao (approved):

michaelosthege (approved):

aloctavodia:
Sorry for being late to the party. Yes I am conformable with treating discrete variables as continuous when proposing new values.

OriolAbril:
I saw this checking the release notes in:

> pm.sample_prior_predictive no longer returns transformed variable values by default. Pass them by name in var_names if you want to obtain these draws (see 4769).

The converter to InferenceData ignores transformed values by default, so I find the phrasing is a bit misleading and potentially troublesome. We should probably add an argument to the converter to include transformed variables into the inferencedata otherwise we'll need to keep the dict return and the capabilities of the function will depend on the output chosen

ricardoV94:
The plan is to revert this, see https://github.com/pymc-devs/pymc/issues/5076

This is no longer needed, as we can use the model.initial_point to get transformed prior predictive samples when they are needed for our samplers

OriolAbril:
I think I should probably go over issues and PRs in both pymc and arviz and make a serious cleanup of integration with InferenceData, but I don't think I'll have time for a while. We still have arguments that were more workarounds than actual arguments/fixes and should be removed as they are generally useless now (i.e. density_dist_obs in to_inferende_data, keep_size in sample_posterior_predictive), the transforms presence is also annoying: https://github.com/arviz-devs/arviz/issues/1509, https://github.com/arviz-devs/arviz/issues/230, and in general we can simplify the converter quite a bit now that it lives in the pymc codebase and should not need complicated logic to work with multiple pymc versions. I think we could also make pointwise log likeihood storage and posterior predictive sampling work with dask (as in my experience it is common that the model/posterior fits in memory but there are many observations and ll and pp do not, and arviz does support working with dask backed arrays, the main limitation right now is creating those dask backed arrays).

Maybe other improvements are also relaatively low hanging fruit?

PR #4770, opened 2021-06-14
Reenable more tests

ricardoV94:
Some tests were mistakenly marked as `--ignore`. I have re-enabeld them. We should fix the outdated comments in `pytest.yml`. @michaelosthege feel free to push to this PR


ricardoV94:
Also found that `find_MAP` is failing quite often after the `V4` refactoring. Most issues seem to boil down to gradients of discrete vars and/or models with both discrete and continuous unobserved variables. The failing tests are marked with an `xfail` for the time being.

michaelosthege (approved):
@ricardoV94 I updated the comments. Take a look to check if they now explain how this works.

Issue #4771, opened 2021-06-14
`find_MAP` is failing in `V4`


ricardoV94:
While working on #4770 I found that `find_MAP` is failing quite often after the `V4` refactoring.

Most issues seem related to gradients of discrete vars and/or models with both discrete and continuous unobserved variables. The failing tests in [here](https://github.com/pymc-devs/pymc3/blob/722341f0f7bda1f96671fe2d532e79b768b0db1c/pymc3/tests/test_starting.py#L49-L100) are marked with an `xfail` for the time being.

michaelosthege:
Could it be a problem with the initialization too? (crossref #4752)
I found "find_MAP" to be quite brittle when the initialization is bad. Not only because optimization is hard, but also because the automatic convergence detection often makes it give up after a handful of iterations.

ricardoV94:
I have the impression it's more than that. But it was some time since I had a look

twiecki:
@michaelosthege is this fixed now?

ricardoV94:
This is not fixed. The tests marked with xfail above are still failing in main

michaelosthege:
Yeah we need the `get_moment` implementations on the most important distributions plus the switch of the default strategy.

I opened #5087, but currently don't have any time to continue working on it for the near future. Anybody please feel free to take over.

ricardoV94:
I double checked, the issue is not (just) the starting points. We are getting Type/Shape Errors somewhere inside the scipy optimization library

ricardoV94:
After #5320 there is only one test that is failing: https://github.com/pymc-devs/pymc/blob/6191c61d0f913c2c5f18effd3074548f9d266186/pymc/tests/test_starting.py#L49

Some shape error comes from deep inside the scipy optimization routine. Only the first of the two calls to `find_MAP` seems to be raising, suggesting it is something to do with the exclusion of the discrete variable, which does not happen in the second call:

https://github.com/pymc-devs/pymc/blob/6191c61d0f913c2c5f18effd3074548f9d266186/pymc/tests/test_starting.py#L62-L63

PR #4773, opened 2021-06-15
Create Ordered Multinomial distribution

AlexAndorra:
This PR enables the ordered logistic constraint on multinomial observed data (i.e aggregated by trial).
Currently, `OrderedLogistic` only accepts data in a disaggregated format (i.e `Categorical` observed data).

I also added the option of having the inferred probabilities in the trace (not only the cutpoints, which are hard to interpret anyway). As I'm guessing that's mostly what people care about when using this likelihood, the option is `True` by default (it's only useful to disable it when memory usage is an issue).

This probably needs a few tests (inspired by those of `OrderedLogistic`), but here is already a proof of concept:
```python
true_cum_p = np.array([0.1, 0.15, 0.25, 0.50, 0.65, 0.90, 1.0])
true_p = np.hstack([true_cum_p[0], true_cum_p[1:] - true_cum_p[:-1]])
fake_elections = np.random.multinomial(n=1_000, pvals=true_p, size=60)

with pm.Model() as m:
    cutpoints = pm.Normal(
        "cutpoints",
        mu=np.arange(6) - 2.5,
        sigma=1.5,
        initval=np.arange(6) - 2.5,
        transform=pm.distributions.transforms.ordered,
    )

    pm.OrderedMultinomial(
        "results",
        eta=0.0,
        cutpoints=cutpoints,
        n=fake_elections.sum(1),
        observed=fake_elections,
    )

    trace = pm.sample()
```
Which does recover the true probabilities:
complete_p

+ [x] No breaking changes
+ [x] New tests
+ [x] Run linting / style checks
+ [x] Mention in `RELEASE-NOTES.md`


pymc3/distributions/multivariate.py
```diff
@@ -687,6 +688,32 @@ def _distr_parameters_for_repr(self):
...
+        p_cum = at.concatenate(
+            [
+                at.zeros_like(at.shape_padright(pa[..., 0])),
+                pa,
+                at.ones_like(at.shape_padright(pa[..., 0])),
+            ],
+            axis=-1,
+        )
+        if compute_p:
+            p = pm.Deterministic("complete_p", p_cum[..., 1:] - p_cum[..., :-1])
```
ricardoV94: This should probably check if it's being called inside a model context, otherwise it will raise a perhaps not very informative error. I.e., when calling `OrderedMultinomial.dist(...)` outside of a `Model` context.

pymc3/distributions/multivariate.py
```diff
@@ -687,6 +688,32 @@ def _distr_parameters_for_repr(self):
...
 
 
+class OrderedMultinomial(Multinomial):
+    rv_op = multinomial
+
+    @classmethod
+    def dist(cls, eta, cutpoints, n, compute_p=True, *args, **kwargs):
+        eta = at.as_tensor_variable(floatX(eta))
+        cutpoints = at.as_tensor_variable(cutpoints)
+        n = at.as_tensor_variable(n)
```
```suggestion
        n = at.as_tensor_variable(intX(n))
```

pymc3/distributions/multivariate.py
```diff
@@ -687,6 +688,32 @@ def _distr_parameters_for_repr(self):
         return ["n", "a"]
 
 
+class OrderedMultinomial(Multinomial):
+    rv_op = multinomial
```
ricardoV94, line 695: Not sure if this is even needed. I see I put it in the other `Ordered` when I refactored them but it might be unnecessary

ricardoV94:
> I also added the option of having the inferred probabilities in the trace (not only the cutpoints, which are hard to interpret anyway). 

This would also be nice for the other `Ordered` distributions no?

pymc3/distributions/multivariate.py
ricardoV94: Also name should extend the user defined name. Otherwise you'll get an error when using two distributions inside a model due to the repeated Deterministic name.

AlexAndorra: Good points! I handled the model context stuff, but how do I get the user-defined name in v4?

pymc3/distributions/multivariate.py
AlexAndorra: Done

pymc3/distributions/multivariate.py
AlexAndorra: Apparently it is: I just tried without it and it throws an `AttributeError: 'NoneType' object has no attribute 'ndim_supp'`

pymc3/distributions/multivariate.py
ricardoV94: Good question actually

AlexAndorra: I've tried a few things but can't figure it out 🤔

ricardoV94: Then it's in `__new__` I guess

AlexAndorra:
> This would also be nice for the other Ordered distributions no?

Just added that and addressed your other comments 🥳 
Only the name issue and new tests remaining -- will handle this later

pymc3/distributions/multivariate.py
AlexAndorra: You mean `cls.__new__().name` ?

ricardoV94:
Maybe it's easier to keep the two things separate. Have a helper class that creates the raw Ordered distribution and then, optionally, creates the deterministic p. You can retrieve `p` from the inputs that go into the variable returned from the raw Ordered distribution. Something along the lines of:
```python
class _OrderedMultinomial(Multinomial):
  # standard implementation without the helper Deterministic

class OrderedMultinomial:
  def __new__(cls, name, *args, save_p=True, **kwargs):
    out_rv = _OrderedMultinomial(name, *args, **kwargs)
    if save_p:
      pm.Deterministic(f'{name}_p', out_rv.owner.inputs[4])  # I think this is the p vector in the multinomial
    return out_rv

  @classmethod
  def dist(cls, *args, **kwargs):
    return _OrderedMultinomial.dist(*args, **kwargs)
```
No need to even check if it's inside the model context, that's done by the default `_OrderedMultinomial`.

ricardoV94:
~~Ignore the previous idea. I think you can overwrite `__new__` to save name in `self.name` that's probably simpler.~~ No, doesn't work either. The previous comment goes back to being my best idea :p

AlexAndorra:
Agreed, I was thinking the same -- that's actually how I did it for `LKJCholeskyCov` back in the days

AlexAndorra:
`out_rv.owner.inputs[4])` does seem to work @ricardoV94 ! Small question though: how did you know that the p vector was in the 4th position? 😅

ricardoV94:
> `out_rv.owner.inputs[4])` does seem to work @ricardoV94 ! Small question though: how did you know that the p vector was in the 4th position? sweat_smile

The RVs inputs are always of the form `(RandomStateSharedVariable, size, dtype, param1, param2 ..., paramN)`

```python
import aesara.tensor as at
x = at.random.multinomial(5, np.ones(3)/3, size=2)
print(x.owner.inputs)
```
```
[RandomStateSharedVariable(<RandomState(MT19937) at 0x7F125671EC40>),
 TensorConstant{(1,) of 2},
 TensorConstant{4},
 TensorConstant{5},
 TensorConstant{(3,) of 0...3333333333}]
```

AlexAndorra:
> The RVs inputs are always of the form (RandomStateSharedVariable, size, dtype, param1, param2 ..., paramN)

Ooooh, ok, thanks! What does a `dtype` of ` TensorConstant{4}` mean though? 🤔

ricardoV94:
> > The RVs inputs are always of the form (RandomStateSharedVariable, size, dtype, param1, param2 ..., paramN)
> 
> Ooooh, ok, thanks! What does a `dtype` of ` TensorConstant{4}` mean though? thinking

It's a numerical code for different dtypes ("float32", "float64", "int32", and so on). I don't know where they are defined though. 

4 is probably for "int64" since that's the default dtype of discrete RVs

AlexAndorra:
Ah ok, I wouldn't have guessed that 😅 
So the only thing left for this PR is adding a few tests. I think I need to write the logpdf for that but don't know how it works yet for this distribution

ricardoV94:
> So the only thing left for this PR is adding a few tests. I think I need to write the logpdf for that but don't know how it works yet for this distribution

The logp shouldn't have to be tested as it's just the `multinomial`. The tests should just make sure the parameters that go into it are the correct ones. Have a look at the `check_pymc_params_match_rv_op` tests in `test_random.py` ([guide](https://github.com/pymc-devs/pymc3/blob/54be453f9bff0fe73430ae8f9d26732c3816e6bb/docs/source/distribution_implementation.md#3-adding-tests-for-the-new-randomvariable)), I think it does just what is needed.

Plus some tests for the helper Deterministic

pymc3/distributions/discrete.py
```diff
@@ -1721,18 +1727,30 @@ def dist(cls, eta, cutpoints, *args, **kwargs):
         )
         p = p_cum[..., 1:] - p_cum[..., :-1]
 
-        return super().dist(p, **kwargs)
+        return super().dist(p, *args, **kwargs)
 
 
-class OrderedProbit(Categorical):
+class OrderedLogistic:
```
ricardoV94: It's probably better to move the docstrings to the user-facing class, or they won't find them.

AlexAndorra: Indeed. That's done

ricardoV94 (dismissed):
LGTM!

ricardoV94:
The failing tests in `distributions_random.py` are due to the test looking for the rv_op which doesn't exist in the wrapper class. Probably just need to overwrite the relevant part in the inherited test subclasses.

AlexAndorra:
I just added back the rv_op in the wrapper class. Tests pass locally. Is that a good solution?

ricardoV94:
> Is that a good solution?

Actually I think testing the wrapped distribution might the best. Just test the _Ordered* classes instead.

This shouldn't require any changes to the wrapper classes

pymc3/tests/test_distributions_random.py
```diff
@@ -1451,6 +1451,21 @@ class TestOrderedProbit(BaseTestDistribution):
     ]
 
 
+class TestOrderedMultinomial(BaseTestDistribution):
+    pymc_dist = pm.OrderedMultinomial
```
ricardoV94:
```suggestion
    pymc_dist = pm._OrderedMultinomial
```

Or an explicitly imported `_OrderedMultinomial` if it's not available at the module level. Same thing for the other ones

AlexAndorra: Done

pymc3/distributions/multivariate.py
```diff
@@ -690,6 +687,123 @@ def _distr_parameters_for_repr(self):
...
+                n=fake_elections.sum(1),
+                observed=fake_elections,
+            )
+
+            trace = pm.sample()
+
+        # Plot the results
+        arviz.plot_posterior(trace_12_4, var_names=["complete_p"], ref_val=list(true_p));
+    """
+    rv_op = multinomial
```
```suggestion
```

This should now be fine to leave out (also for the others)

AlexAndorra: Of course, I just forgot to delete them 🤦‍♂️

AlexAndorra:
All tests are green 🥳 
Just added a mention in `RELEASE-NOTES.md`

pymc3/distributions/multivariate.py
```diff
@@ -112,17 +113,13 @@ def quaddist_parse(value, mu, cov, mat_type="cov"):
...
     delta = value - mu
-
-    if mat_type == "cov":
-        # Use this when Theano#5908 is released.
-        # return MvNormalLogp()(self.cov, delta)
-        chol_cov = cholesky(cov)
+    # Use this when Theano#5908 is released.
+    # return MvNormalLogp()(self.cov, delta)
+    chol_cov = cholesky(cov)
+    if mat_type == "cov" or mat_type != "tau":
```
```suggestion
    if mat_type != "tau":
```
Just a minor suggestion, feel free to merge if you want

ricardoV94 (approved):
Looks good! Left a minor suggestion that can be ignored.

ricardoV94:
The failing test comes from #4771. It can be ignored.

~~Finally, do we need to add an entry in the docs for the new distribution or is it automatic?~~ I added it

AlexAndorra:
Merging 🍾 Thanks for your help and reviews @ricardoV94 !

Issue #4488, opened 2021-02-26
Imputed variables are not auto-transformed


npschafer:
I’m looking to do imputation of partially observed random variables with positive support. Attempting to do so results in sampling values outside the support of the distribution and causes various downstream steps to fail, e.g., finding the MAP and variational inference.
```python
import pymc3 as pm
from numpy.ma import masked_values

MASK_VALUE = -999.
observed_values = np.array([1., MASK_VALUE], dtype=float)
with pm.Model() as model:
    pm.Gamma("partially_observed", alpha=3., beta=1/0.1, observed=masked_values(observed_values, value=MASK_VALUE))
    map_estimate = pm.find_MAP()
print(map_estimate)
print(model.test_point)
```

```python
{'partially_observed_missing': array([-0.7])}
{'partially_observed_missing': array([0.3])}
```

I originally raised this on the PyMC3 Discourse (https://discourse.pymc.io/t/imputation-for-partially-observed-random-variables-with-positive-support/6844) and @ckrapu suggested that I create an issue.

## Versions and main components

* PyMC3 Version: '3.9.3'
* Theano Version: '1.0.5' (not sure if this is the correct version as things are somewhat confusing with having Theano, Theano-PyMC, and Aesara all co-existing)
* Python Version: 3.6.12
* Operating system: Linux
* How did you install PyMC3: pip

Thanks for all the great work and for your help with resolving the issue!

ricardoV94:
I can replicate.

This is a very unusual example in that there is nothing to infer other than the missing value (other gamma parameters are fixed). If you add an unobserved parameter it seems things work fine:

```python
with pm.Model() as model: 
    alpha = pm.Gamma('alpha', mu=3, sigma=.001) 
    pm.Gamma("partially_observed", alpha=alpha, beta=1/0.1, observed=masked_values(observed_values, value=MASK_VALUE)) 
    print(pm.find_MAP())
```
```
{'alpha_log__': array(1.09861256), 'partially_observed_missing': array([0.19999973]), 'alpha': array(3.00000081)}
```                               

If you just model this as an unobserved distribution it's also fine:
```python
with pm.Model() as model: 
     pm.Gamma('not_observed', alpha=3, beta=1/0.1)
     print(pm.find_MAP())
```
```
{'not_observed_log__': array(-1.60943705), 'not_observed': array(0.20000017)}
```

Looks like a real edge case. 

Did you have a more realistic example where this issue also appeared?

[...]

PR #4774, opened 2021-06-16
[WIP] Porting kroneckernormal distribution to v4

kc611:
This PR refactors the `KroneckerNormal` distribution to be compatible with `v4`.

### DIFF
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -17,6 +17,8 @@
 
 import warnings
 
+from functools import reduce
+
 import aesara
 import aesara.tensor as at
 import numpy as np
@@ -45,7 +47,7 @@ from pymc3.distributions import transforms
 from pymc3.distributions.continuous import ChiSquared, Normal, assert_negative_support
 from pymc3.distributions.dist_math import bound, factln, logpow, multigammaln
 from pymc3.distributions.distribution import Continuous, Discrete
-from pymc3.math import kron_diag, kron_dot, kron_solve_lower, kronecker
+from pymc3.math import kron_diag, kron_dot
 
 __all__ = [
     "MvNormal",
@@ -1702,6 +1704,32 @@ class MatrixNormal(Continuous):
         return ["mu", "row" + mapping[self._rowcov_type], "col" + mapping[self._colcov_type]]
 
 
+class KroneckerNormalRV(RandomVariable):
+    name = "kroneckernormal"
+    ndim_supp = 2
+    ndims_params = [1, 0, 2]
+    dtype = "floatX"
+    _print_name = ("KroneckerNormal", "\\operatorname{KroneckerNormal}")
+
+    def _shape_from_params(self, dist_params, rep_param_idx=0, param_shapes=None):
+        return default_shape_from_params(1, dist_params, rep_param_idx, param_shapes)
+
+    def rng_fn(self, rng, mu, sigma, *covs, size=None):
+        size = size if size else covs[-1]
+        covs = covs[:-1] if covs[-1] == size else covs
+
+        cov = reduce(linalg.kron, covs)
+
+        if sigma:
+            cov = cov + sigma ** 2 * np.eye(cov.shape[0])
+
+        x = multivariate_normal.rng_fn(rng=rng, mean=mu, cov=cov, size=size)
+        return x
+
+
+kroneckernormal = KroneckerNormalRV()
+
+
 class KroneckerNormal(Continuous):
     r"""
     Multivariate normal log-likelihood with Kronecker-structured covariance.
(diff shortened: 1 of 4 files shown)


pymc3/distributions/multivariate.py
```diff
@@ -1781,119 +1797,87 @@ class KroneckerNormal(Continuous):
...
+                cov_i = at.dot(Q, at.dot(at.diag(eig), Q.T))
+                covs.append(cov_i)
+            cov = kronecker(*covs)
+            if is_noisy:
+                cov = cov + sigma ** 2 * at.identity_like(cov)
 
-    def random(self, point=None, size=None):
+        mu = at.as_tensor_variable(mu)
+
+        chols = at.as_tensor_variable(chols)
```
kc611: @ricardoV94 The `chols` parameter here is a list of either `Cholesky` Ops or a pre-defined list of arrays. How do we turn this into the `TensorConstant` variant ?

pymc3/distributions/multivariate.py
```diff
@@ -1902,39 +1886,39 @@ def _quaddist(self, value):
...
-        delta = value - self.mu
-        if self._isEVD:
-            sqrt_quad = kron_dot(self.QTs, delta.T)
-            sqrt_quad = sqrt_quad / at.sqrt(self.eigs[:, None])
-            logdet = at.sum(at.log(self.eigs))
+        delta = value - mu
+
+        # Setup EVD if applicable
+        if eigh_iterable.data.shape != (0,):
+            eigs_sep, Qs = zip(*eigh_iterable.data)  # Unzip
```
kc611: Again a few hacks were required over here (using `.shape` for checking an empty `TensorVariable` and `.data` for accessing the data inside)

pymc3/distributions/multivariate.py
```diff
@@ -1902,39 +1886,39 @@ def _quaddist(self, value):
...
+            if is_noisy:
+                eigs += sigma ** 2
+            N = eigs.shape[0]
+
+            sqrt_quad = kron_dot(QTs, delta.T)
+            sqrt_quad = sqrt_quad / at.sqrt(eigs[:, None])
+            logdet = at.sum(at.log(eigs))
         else:
-            sqrt_quad = kron_solve_lower(self.chols, delta.T)
+            sizes = at.as_tensor_variable([chol.shape[0] for chol in chols])
```
kc611: This is the part where the `chols` variable is required in `logp` method.

pymc3/distributions/multivariate.py
```diff
@@ -1693,6 +1694,21 @@ def _distr_parameters_for_repr(self):
         return ["mu", "row" + mapping[self._rowcov_type], "col" + mapping[self._colcov_type]]
 
 
+class KroneckerNormalRV(RandomVariable):
+    name = "kroneckernormal"
+    ndim_supp = 1
+    ndims_params = [1, 2, 1, 1]
+    dtype = "floatX"
+    _print_name = ("KroneckerNormal", "\\operatorname{Dir}")
```
```suggestion
    _print_name = ("KroneckerNormal", "\\operatorname{KroneckerNormal}")
```

pymc3/distributions/multivariate.py
ricardoV94: I don't think we can make them Constants, it seems like everything can be symbolic here. What is the problem with `at.as_tensor_variable`?

pymc3/distributions/multivariate.py
ricardoV94: Can this be more easily computed in `.dist`? 

Edit: Nevermind it requires delta

pymc3/distributions/multivariate.py
kc611: The list gets converted into a `np.array` with `dtype` as `object` and `TensorVariable` doesn't support that.

pymc3/distributions/multivariate.py
```diff
@@ -1790,119 +1805,83 @@ class KroneckerNormal(Continuous):
...
+                eigh_iterable = eigh_map
+                cov = cov + sigma ** 2 * at.identity_like(cov)
             else:
                 # Otherwise use cholesky as usual
-                self.chols = list(map(self.cholesky, self.covs))
-                self.chol_diags = list(map(at.diag, self.chols))
-                self.sizes = at.as_tensor_variable([chol.shape[0] for chol in self.chols])
-                self.N = at.prod(self.sizes)
+                chols = list(map(cholesky, covs))
+                chols = [_chol.eval() for _chol in chols]
```
kc611: I don't know how practical it is to use an `eval()` in a `dist`. But this is one way to get the output's of the cholesky matrices from the `Op`.

pymc3/distributions/multivariate.py
```diff
@@ -1790,119 +1805,83 @@ class KroneckerNormal(Continuous):
...
+            if is_noisy:
+                cov = cov + sigma ** 2 * at.identity_like(cov)
 
-    def random(self, point=None, size=None):
+        mu = at.as_tensor_variable(mu)
+
+        chols = at.as_tensor_variable(chols)
+
+        # mean = median = mode = mu
+        return super().dist([mu, cov, chols, eigh_iterable], **kwargs)
```
kc611: @brandonwillard we discussed the possibility of using a parametrization of just `super().dist([mu, *cov], **kwargs)`. That will be fine for `rng_fn` but the extra args required for `logp` must be passed if we are to support `evd` there.

pymc3/tests/test_distributions.py
```diff
@@ -1944,7 +1944,6 @@ def test_matrixnormal(self, n):
     @pytest.mark.parametrize("n", [2, 3])
     @pytest.mark.parametrize("m", [3])
     @pytest.mark.parametrize("sigma", [None, 1.0])
-    @pytest.mark.xfail(reason="Distribution not refactored yet")
     def test_kroneckernormal(self, n, m, sigma):
         np.random.seed(5)
         N = n * m
```
kc611, line 1949: @ricardoV94 mentioned that we aren't supporting broadcasting means of multivariate distributions anymore. 

To me, this seems to be the case in the current test configuration. (Different `n` and `m` generate `covs` and `chols` matrices of different dimensions, and the domain is a vector of their product) But I'd like someone to confirm weather the current parameters of the test are still valid or weather they should be changed.

ricardoV94: We are supporting "some" cases, but not as many as in `v3`. There was a discussion about this in the port of the `MultivariateT` here: https://github.com/pymc-devs/pymc3/pull/4731

kc611: Thanks for the tip, I think I'm facing a case in `MatrixNormal` in #4777, where like `MVStudentT` the sample output is not scalar but matrix dependent on the parameters.

pymc3/distributions/multivariate.py
ricardoV94: Another, not very pretty option might be to use a specific `_logp` dispatcher, which extracts the extra non-numeric variables from the `tag` and passes them on to the `logp` function

pymc3/tests/test_distributions.py
brandonwillard: > We are supporting "some" cases, but not as many as in `v3`. There was a discussion about this in the port of the `MultivariateT` here: #4731

How do you mean?  We should be able to support _all_ well-defined forms of broadcasting, and with a lot less effort than it took in `v3`.

ricardoV94: I guess it falls under the not well-defined category. Some cases like `pm.MvNormal.dist(0, np.eye(3), shape=(3,3))`, which would work in `V3` but not in `V4`.

brandonwillard: > I guess it falls under the not well-defined category. Some cases like `pm.MvNormal.dist(0, np.eye(3), shape=(3,3))`, which would work in `V3` but not in `V4`.

Yes, that is a mathematically misspecified distribution, and the problem isn't with broadcasting&mdash;the problem is that `v3` was incorrectly changing the user input to something that might make sense (emphasis on the "might"), instead of raising an error like it should have.

pymc3/distributions/multivariate.py
```diff
@@ -1911,39 +1867,26 @@ def _quaddist(self, value):
...
+        chols = list(map(cholesky, covs))
+        chols = [_chol.eval() for _chol in chols]
+        sizes = at.as_tensor_variable([chol.shape[0] for chol in chols])
+        N = at.prod(sizes)
+        chol_diags = list(map(at.diag, chols))
+        sqrt_quad = kron_solve_lower(chols, delta.T)
+        logdet = 0
+        for chol_size, chol_diag in zip(sizes, chol_diags):
+            logchol = at.log(chol_diag) * N / chol_size
+            logdet += at.sum(2 * logchol)
```
kc611: So for now I removed the Eigen vector decomposition from here, and this works for all three cases of `cov`, `chol` and `evd`. However this leaves us unable to use `sigma` i.e. Gaussian white noise, over here in `logp`. 

Regardless can someone verify if this implementation is mathematically correct for the `evd` case ? (The test pass, however, I'm not too sure if this approach is mathematically correct for that case.)

pymc3/tests/test_distributions.py
kc611: I used `*covs` as @brandonwillard suggested and that handles the differently shaped matrices in the arguments. So I think we're good over here.

pymc3/tests/test_distributions.py
```diff
@@ -1989,32 +1988,32 @@ def test_kroneckernormal(self, n, m, sigma):
...
-        )
-        self.check_logp(
-            KroneckerNormal,
-            dom,
-            std_args,
-            kron_normal_logpdf_evd,
-            extra_args=evd_args,
-            scipy_args=evd_args,
-        )
+        # dom = Domain([np.random.randn(2, N) * 0.1], edges=(None, None), shape=(2, N))
```
kc611, line 1991: This is again related to how `logp` expects the shape of `value` variable to be. Does it derive it using `params` ? If so it'll require override similar to `_shape_from_params` but for `logp`

For instance in these tests it expects the shape of `value` to be 1 dimensional. But receives a two dimensional array that we provide from here.

pymc3/tests/test_distributions.py
```diff
@@ -1943,8 +1943,7 @@ def test_matrixnormal(self, n):
 
     @pytest.mark.parametrize("n", [2, 3])
     @pytest.mark.parametrize("m", [3])
-    @pytest.mark.parametrize("sigma", [None, 1.0])
-    @pytest.mark.xfail(reason="Distribution not refactored yet")
+    @pytest.mark.parametrize("sigma", [None])
```
kc611, line 1946: Removed to see if anything else fails.

kc611:
@ricardoV94 The current failure is the test from https://github.com/pymc-devs/pymc3/issues/4488 which is something that you worked upon. The difference between the values is nearly negligible, though you might wanna have a look at it again if it's something that needs to be very precise.

ricardoV94:
> @ricardoV94 The current failure is the test from #4488 which is something that you worked upon. The difference between the values is nearly negligible, though you might wanna have a look at it again if it's something that needs to be very precise.

Maybe try to increase the tolerance to `1e-4`

pymc3/distributions/multivariate.py
brandonwillard: Yeah, we almost never want to use `eval`/`aesara.function` outside of a place where it's intended to be used (e.g. posterior predictive sampling).

Regardless, why would we want/need to use non-symbolic values here?

brandonwillard (changes requested):

pymc3/distributions/multivariate.py
kc611: The exact place this comment tags right now i.e. `dist` is `eval()` free. The `.eval()` is currently required over in `rng_fn` since the `kronecker()` method was returning `TensorVariable` (I'm working on replacing that with `scipy.linalg.kron` so we can count on `eval()` being removed from there too) The other place `.eval()` is in is `logp` where it is used to infer shapes (Theoretically that can also be removed with a bit of effort.)

Overall with the current implementation (which is a whole lot of omitted code) I wouldn't worry about `eval()` as much as the inability of supporting `sigma` which us the bigger issue with this implementation.

pymc3/tests/test_distributions_random.py
```diff
@@ -1348,6 +1344,35 @@ def dist(cls, **kwargs):
...
+    reference_dist_params = {"mu": mu, "covs": covs, "sigma": sigma}
+    sizes_to_check = [None, (), 1, (1,), 5, (4, 5), (2, 4, 2)]
+    sizes_expected = [(N,), (N,), (1, N), (1, N), (5, N), (4, 5, N), (2, 4, 2, N)]
+
+    reference_dist = lambda self: functools.partial(
+        self.kronecker_rng_fn, rng=self.get_random_state()
+    )
+    tests_to_run = [
+        "check_pymc_draws_match_reference",
+        "check_rv_size",
```
ricardoV94, line 1372:
```suggestion
        "check_pymc_params_match_rv_op",
        "check_pymc_draws_match_reference",
        "check_rv_size",
```

Any reason not to test the `params_match`?

ricardoV94 (commented):
Look great, I added just a tiny suggestion

pymc3/tests/test_distributions_random.py
kc611: Oh, that fails because the way we're handling the `cov` parameter internally. (as `*cov`) . So to that test it appears as a list of `TensorVariable` instead of a single `TensorVariable`

ricardoV94: I see. Do you think there is any point in implementing an alternative test for this logic or given things are working on the logp / random side, it should be fine?

pymc3/distributions/multivariate.py
brandonwillard: We simply cannot evaluate things like this unless we place very strong and unnatural restrictions on _all_ of the terms involved.  More specifically, the graphs in `chols` must have no symbolic inputs; otherwise, this will fail.  If the tests are passing with this in place, then our testing isn't sufficient, because it's only ever testing `Constant`/`SharedVariable` inputs.

brandonwillard (dismissed):

brandonwillard (approved):

pymc3/tests/test_distributions_random.py
kc611: It should be fine, I think.

pymc3/distributions/multivariate.py
```diff
@@ -1790,119 +1818,49 @@ class KroneckerNormal(Continuous):
...
-        self._setup(covs, chols, evds, sigma)
-        super().__init__(*args, **kwargs)
-        self.mu = at.as_tensor_variable(mu)
-        self.mean = self.median = self.mode = self.mu
+    @classmethod
+    def dist(cls, mu, covs=None, chols=None, evds=None, sigma=None, *args, **kwargs):
 
-    def _setup(self, covs, chols, evds, sigma):
-        self.cholesky = Cholesky(lower=True, on_error="raise")
+        cholesky = Cholesky(lower=True, on_error="raise")
```
ricardoV94: This line seems to not be doing anything.
```suggestion
```

pymc3/distributions/multivariate.py
```diff
@@ -1790,119 +1818,49 @@ class KroneckerNormal(Continuous):
...
-        self._setup_random()
-        self.mv_params["shape"] = self.shape
-        dist = MvNormal.dist(**self.mv_params)
-        return dist.random(point, size)
-
-    def _quaddist(self, value):
-        """Computes the quadratic (x-mu)^T @ K^-1 @ (x-mu) and log(det(K))"""
+        # Computes the quadratic (x-mu)^T @ K^-1 @ (x-mu) and log(det(K))
         if value.ndim > 2 or value.ndim == 0:
             raise ValueError("Invalid dimension for value: %s" % value.ndim)
```
```suggestion
            raise ValueError(f"Invalid dimension for value: {value.ndim}")
```

pymc3/distributions/multivariate.py
```diff
@@ -1911,39 +1869,29 @@ def _quaddist(self, value):
...
-            Value for which log-probability is calculated.
-
-        Returns
-        -------
-        TensorVariable
-        """
-        quad, logdet = self._quaddist(value)
-        return -(quad + logdet + self.N * at.log(2 * np.pi)) / 2.0
+        a = -(quad + logdet + N * at.log(2 * np.pi)) / 2.0
+        return at.as_tensor_variable(a)
```
```suggestion
        return a
```
I don't think we need the `as_tensor_variable`

ricardoV94 (commented):
Look great, although I am not familiar with the distribution so I will have to trust the tests on this one. I left a couple of minor suggestions.

ricardoV94 (approved):
Checking if the tests pass one last time before merging. 

Great work @kc611!

Issue #4518, opened 2021-03-09
Developing random method for conditional autoregression


ckrapu:
Support for conditional autoregressions AKA Gaussian Markov random fields was added in PR [#4504](https://github.com/pymc-devs/pymc3/pull/4504) but the `.random` method was left not implemented as there was not an obvious choice of algorithms that would scale well as a function of the field dimension. A bit of discussion on this can be seen at issue #3689. There may exist efficient methods for doing this which could be implemented, however.

To be more explicit, the core difficulty is sampling `x ~ CAR(mu, tau, alpha, W)` where the standard way to do this is by taking a unit isotropic Gaussian vector `z` and solving `z = Lx` where `L` is the Cholesky decomposition of the CAR-structured precision matrix. This operation has cubic complexity in the dimension of `x` and thus is a nonstarter for CAR/GMRF vectors with more than a few hundred sites.

ckrapu:
A potentially viable method is described in the paper [Fast sampling of Gaussian Markov random fields](https://rss.onlinelibrary.wiley.com/doi/pdf/10.1111/1467-9868.00288) by Havard Rue. The basic recipe is as follows:

1. Sample hyperparameters `tau`, `alpha` for the CAR scale and autocorrelation, respectively
2. Compute a permutation of the rows of adjacency matrix `W` which minimizes the matrix bandwidth ([Scipy method referenced here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.reverse_cuthill_mckee.html))
3.  Calculate the Cholesky decomposition `L` of the precision matrix `Q = tau*(D - alpha * W)` using efficient routines for banded matrices ([Scipy function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cholesky_banded.html))
4.  Solve the equation `z = Lx` for `x` which is relatively quick since `L` is triangular. ([Scipy function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cho_solve_banded.html#scipy.linalg.cho_solve_banded))

Note that step 2 doesn't need to be done for each sample - it only needs to be done once for each value of `W`.

brandonwillard:
`Distribution.random` will be removed shortly during the move to `v4`, so, if anyone is looking to do this, it should be done in the context of an Aesara `RandomVariable` `Op`.

aerubanov:
[...]

PR #4775, opened 2021-06-16
Generalized poisson distribution

ally-lee:
In this PR, I am adding an implementation of the Generalized Poisson distribution.

In the standard Poisson distribution, the mean is always equal to the variance. The Generalized Poisson has a dispersion parameter `-1 <= lambda <= 1` that controls the variance. When `lambda < 0`, the variance is less than the mean. When `lambda > 0`, the variance is greater than the mean. When `lambda = 0`, the Generalized Poisson reduces to the standard Poisson. You can read more about the distribution in this [blog post](https://towardsdatascience.com/generalized-poisson-regression-for-real-world-datasets-d1ff32607d79).

The Generalized Poisson likelihood was very useful for a recent project where we were modeling count data because we wanted that flexibility in dispersion. If you're interested in checking out our work, it's available [here](https://arxiv.org/abs/2104.09327).

I defined the Generalized Poisson as a custom distribution in PyMC3 for my project. I will also submit a PR to the `pymc-examples` repo with a notebook outlining how I defined a custom distribution.


pymc3/distributions/discrete.py
```diff
@@ -603,6 +603,89 @@ def logcdf(value, mu):
...
+        """
+        log_prob = bound(np.log(theta) + logpow(theta + lam * value, value - 1)
+                         - (theta + lam * value) - factln(value),
+                         theta >= 0,
+                         -1 <= lam, -theta/4 <= lam, lam <= 1,
+                         value >= 0)
+        # Return zero when value > m, where m is the largest pos int for which theta + m * lam > 0 (when lam < 0)
+        return at.switch(theta + value * lam <= 0, 0, log_prob)
+
+    def random(theta, lam, size=None):
```
ally-lee: This generator function uses the Inversion Algorithm defined in [this paper](https://www.tandfonline.com/doi/abs/10.1080/01966324.1997.10737439?journalCode=umms20).

I see that the distributions now have a `logcdf` function instead of a `random` function. If it seems like I'm on the right track so far, I'll look into defining a `logcdf` function.

ricardoV94:
Hi @ally-lee thank you for opening the PR. It seems that your distribution implementation still mixes a bit of the old PyMC logic. 

There is a big change going on that requires distributions to be implemented in a slightly different manner. You can have a look here for a quick guide of how to translate distributions to the new version: https://github.com/pymc-devs/pymc3/issues/4686#issuecomment-839541040

And here about how to create these new RandomVariable ops that implement the random method: https://github.com/pymc-devs/pymc3/issues/4518#issuecomment-806236360

pymc3/distributions/discrete.py
ricardoV94: The logcdf method is not instead of the random method. See my comment on the main thread. 

The `logcdf` is also not necessary to implement (but it's a nice plus if you can provide it)

pymc3/distributions/discrete.py
```diff
@@ -603,6 +603,89 @@ def logcdf(value, mu):
...
+    lam: float
+        Dispersion parameter (max(-1, -theta/4) <= lam <= 1).
+
+    Notes
+    -----
+    When lam = 0, the Generalized Poisson reduces to the standard Poisson with mu = theta.
+    When lam < 0, the mean is greater than the variance (underdispersion).
+    When lam > 0, the mean is less than the variance (overdispersion).
+    """
+    rv_op = genpoisson
```
ricardoV94: This will have to be implemented

pymc3/distributions/discrete.py
```diff
@@ -603,6 +603,89 @@ def logcdf(value, mu):
...
+        -------
+        TensorVariable
+        """
+        log_prob = bound(np.log(theta) + logpow(theta + lam * value, value - 1)
+                         - (theta + lam * value) - factln(value),
+                         theta >= 0,
+                         -1 <= lam, -theta/4 <= lam, lam <= 1,
+                         value >= 0)
+        # Return zero when value > m, where m is the largest pos int for which theta + m * lam > 0 (when lam < 0)
+        return at.switch(theta + value * lam <= 0, 0, log_prob)
```
ricardoV94: Did you mean `0` or `-inf` here?

ricardoV94 (changes requested):

ally-lee:
Hi @ricardoV94 thank you for this feedback! I will take a stab at making these updates.

ricardoV94:
There is also a current PR with a developer facing guide on adding new distributions: https://github.com/pymc-devs/pymc3/pull/4783

It seems your new notebook in pymc-examples is more user-focused, which we also need. I think the two new resources could work very well together.

Feel free to ask any questions about the new changes, if you have any.

pymc3/distributions/discrete.py
ally-lee: I will double check my math. Are you referring to the first or second `0`?

pymc3/distributions/discrete.py
```diff
@@ -603,6 +603,98 @@ def logcdf(value, mu):
...
 
+class GeneralizedPoissonRV(RandomVariable):
+    name = "generalized_poisson"
+    ndim_supp = 0
+    ndims_params = [0, 0]
+    dtype = "int64"
+    _print_name = ("GeneralizedPoisson", "\\operatorname{GeneralizedPoisson}")
+
+    @classmethod
+    def rng_fn(cls, rng, theta, lam, size):
```
ally-lee: Is there an expected shape for these parameters? This implementation is assuming that `theta` is an array and `lam` should be a scalar, so the method will return an array of random samples, one for each value of `theta`.

ricardoV94: The minimum expected shape is the one you define in ndims_supp, which as a `[0,0]` means two scalar values. This sounds correct for this univariate distribution.

Can you tweak the logic slightly to allow for two scalar arguments as well as the vector case?

It also seems like you generating code could be vectorized (instead of looping through the `i`) for efficiency

pymc3/distributions/discrete.py
ally-lee: Ah you are so right! The prob is `0`, so the logprob should be `-inf`

OriolAbril:
> There is also a current PR with a developer facing guide on adding new distributions: #4783

I wanted to echo https://github.com/pymc-devs/pymc3/pull/4783#discussion_r654362128 to make sure the new distribution is shown in the docs when the time comes and we make a release. In this case it should go somewhere in https://github.com/pymc-devs/pymc3/blob/main/docs/source/api/distributions/continuous.rst

pymc3/distributions/discrete.py
ally-lee: Updated the logic so that both arguments can be either scalars or vectors, and vectorized the generating code.

pymc3/distributions/discrete.py
```diff
@@ -603,6 +604,105 @@ def logcdf(value, mu):
...
+                         -1 <= lam, -theta/4 <= lam, lam <= 1,
+                         value >= 0)
+        # Probability is 1 when theta and value are both 0.
+        # Probability is 0 when value > m, where m is the largest positive integer for which
+        # theta + m * lam > 0 (when lam < 0).
+        return at.switch(at.eq(theta, 0) * at.eq(value, 0),
+                         0,
+                         at.switch(theta + value * lam <= 0,
+                                   -np.inf,
+                                   log_prob))
```
ally-lee: I am taking the PMF from [Consul & Famoye (1992)](https://www.tandfonline.com/doi/pdf/10.1080/03610929208830766):
Screen Shot 2021-06-20 at 7 30 13 PM

pymc3/distributions/discrete.py
```diff
@@ -603,6 +604,105 @@ def logcdf(value, mu):
...
+        Returns
+        -------
+        TensorVariable
+        """
+        log_prob = bound(np.log(theta) + logpow(theta + lam * value, value - 1)
+                         - (theta + lam * value) - factln(value),
+                         theta >= 0,
+                         -1 <= lam, -theta/4 <= lam, lam <= 1,
+                         value >= 0)
+        # Probability is 1 when theta and value are both 0.
```
ricardoV94: It says above in the docstrings that theta must be `>0`. We usually prefer to keep the logp formula simple and to not include limiting cases. 

Can we just enforce the parameter domain to be strictly positive or is it important to include this special case?

pymc3/distributions/discrete.py
```diff
@@ -603,6 +604,105 @@ def logcdf(value, mu):
...
+        # Probability is 0 when value > m, where m is the largest positive integer for which
+        # theta + m * lam > 0 (when lam < 0).
+        return at.switch(at.eq(theta, 0) * at.eq(value, 0),
+                         0,
+                         at.switch(theta + value * lam <= 0,
```
ricardoV94: This switch that leads to `-inf` can be replaced by an expression on the `bound` above

pymc3/distributions/discrete.py
ricardoV94: The reference should be added to the docstrings in that case.

pymc3/distributions/discrete.py
ally-lee: You are right, the bound should actually be `theta > 0` so we can remove this special case.

pymc3/distributions/discrete.py
ally-lee: Can you show what that would look like?

ricardoV94: You add the reverse condition, if it's not true, it will lead to `-inf`:
```python
bound(expr, cond1, cond2, (theta + value * lam) > 0,)

File: pymc3/distributions/discrete.py [hunk 8]
```diff
@@ -603,6 +604,107 @@ def logcdf(value, mu):
...
+        TensorVariable
+        """
+        log_prob = bound(np.log(theta) + logpow(theta + lam * value, value - 1)
+                         - (theta + lam * value) - factln(value),
+                         theta > 0,
+                         -1 <= lam, -theta/4 <= lam, lam <= 1,
+                         value >= 0)
+        # Probability is 0 when value > m, where m is the largest positive integer for which
+        # theta + m * lam > 0 (when lam < 0).
+        return at.switch(theta + value * lam <= 0, -np.inf, log_prob)
```
ricardoV94: I tried to simplify the logp expression. What do you think? (Please check if I didn't add any mistakes)

```suggestion
        theta_lam_value = theta + lam * value
        log_prob = (
            np.log(theta)
            + logpow(theta_lam_value, value - 1) 
            - theta_lam_value - factln(value)
        )
        
        # Probability is 0 when value > m, where m is the largest positive integer for which
        # theta + m * lam > 0 (when lam < 0).
        log_prob = at.switch(theta_lam_value <= 0, -np.inf, log_prob)
        
        return bound(
            log_prob,
            value >= 0,
            theta > 0,
            abs(lam) <= 1, 
            -theta/4 <= lam, 
        )
```

ricardoV94:
We will also need tests. Let me know if you need help

pymc3/distributions/discrete.py
ally-lee: Looks good to me!

ally-lee:
Thank you for being patient with me! I probably won't have time to work on this again until 2 weeks from now. If this needs to go out soon and you're able to write the tests without my help, feel free to take it over. If not, I'm happy to work on them myself!

twiecki:
@ally-lee Any progress on this or should we close it?

ally-lee:
Hi, I believe the only thing left is to write tests. If anyone else wants to work on them, that would be awesome! I'm sorry I won't have time to work on this anymore, so if no one can complete this, I'm ok with closing it out.

mjhajharia:
> Hi, I believe the only thing left is to write tests. If anyone else wants to work on them, that would be awesome! I'm sorry I won't have time to work on this anymore, so if no one can complete this, I'm ok with closing it out.

thanks for the contribution! I'll rebase and add tests etc

mjhajharia:
@twiecki i tried adding tests, but an equivalent doesn't exist in scipy or any popular library, there's one in statsmodels but that's a regression model, and there's this one implementation with a test here in julia https://github.com/grero/StatUtils.jl/blob/200994b3143b692499f11028ea4e5c4c749463b0/test/runtests.jl#L79, but i couldn't find any reliable place to test, with v4 there's already lots of work + this is missing moments and logcdf so maybe we can just tag this inactive and let it be till we have a better solution or more time, because I think the rest of the PR is very nicely done

twiecki:
@mjhajharia Can't we write a test against scipy's Poisson and set lambda=0? "When lambda = 0, the Generalized Poisson reduces to the standard Poisson." That way we have at least that part covered.

OriolAbril:
This looks like the perfect candidate for https://github.com/pymc-devs/pymc-experimental where IIUC, extensive testing wouldn't be required (or maybe even no testing could be enough). cc @fonnesbeck

mjhajharia:
> @mjhajharia Can't we write a test against scipy's Poisson and set lambda=0? "When lambda = 0, the Generalized Poisson reduces to the standard Poisson." That way we have at least that part covered.

yes we can!

twiecki:
@mjhajharia I think that would be good enough for us here.

ricardoV94:
I suggest we move this PR to https://github.com/pymc-devs/pymc-experimental.

If nobody objects I'll close it soon

twiecki:
Actually I think with the test for the poisson this is good for pymc proper.

ricardoV94:
> Actually I think with the test for the poisson this is good for pymc proper.

Then I think we should have some tests with `lambda != 0`

twiecki:
What test could we do?

ricardoV94:
> What test could we do?

Old school comparison with an independent implementation? I think we do that for the ExGaussian for example. Someone just evaluated a bunch of points from the reference R library and checked that our implementation matches

pymc/distributions/discrete.py
```diff
@@ -644,6 +645,121 @@ def logcdf(value, mu):
...
+    dtype = "int64"
+    _print_name = ("GeneralizedPoisson", "\\operatorname{GeneralizedPoisson}")
+
+    @classmethod
+    def rng_fn(cls, rng, theta, lam, size):
+        U = rng.uniform(size=size)
+        omega = np.exp(-lam)
+        S = np.exp(-theta)
+        P = np.copy(S)
+        vrandom = np.vectorize(random)
```
ricardoV94: This call to `np.vectorize` should be done just once at module level

pymc/distributions/discrete.py
```diff
@@ -644,6 +645,121 @@ def logcdf(value, mu):
...
+class GeneralizedPoissonRV(RandomVariable):
+    name = "generalized_poisson"
+    ndim_supp = 0
+    ndims_params = [0, 0]
+    dtype = "int64"
+    _print_name = ("GeneralizedPoisson", "\\operatorname{GeneralizedPoisson}")
+
+    @classmethod
+    def rng_fn(cls, rng, theta, lam, size):
+        U = rng.uniform(size=size)
```
ricardoV94: size has to be inferred from broadcasted shapes of theta and lam when not explicitly defined

pymc/tests/test_distributions.py
```diff
@@ -1769,6 +1769,30 @@ def test_poisson(self):
...
+            GeneralizedPoisson,
+            Nat,
+            {"theta": Rplus, "lam": Domain([0.0, 0.0, 0.0])},
+            lambda value, mu: sp.poisson.logcdf(value, mu),
+            skip_paramdomain_outside_edge_test=True,
+        )
+        self.check_selfconsistency_discrete_logcdf(
+            GeneralizedPoisson,
+            Nat,
+            {"theta": Rplus, "lam": Domain([0.0, 0.0, 0.0])},
```
```suggestion
            {"theta": Rplus, "lam": Domain([0.0, 0.0, 0.0])},
```
We can test different lam here. This is a good way to test our logp/logcdf are at least consistent in the cases where the distribution is not a Poisson

pymc/tests/test_distributions_random.py
```diff
@@ -1035,6 +1035,13 @@ class TestPoisson(BaseTestDistribution):
     tests_to_run = ["check_pymc_params_match_rv_op"]
 
 
+class TestGeneralizedPoisson(BaseTestDistribution):
+    pymc_dist = pm.GeneralizedPoisson
+    pymc_dist_params = {"theta": 4.0, "lam": 0.0}
+    expected_rv_op_params = {"theta": 4.0, "lam": 0.0}
+    tests_to_run = ["check_pymc_params_match_rv_op"]
```
ricardoV94: We should test sizes as well

ricardoV94:
I pushed several changes, namely:
1. Manually vectorize random method
  This is critical for large variables with large mu, otherwise we would be doing the same slow step by step CDF calculation for every draw in the same variable.
1. Improve numerical stability of random method by computing CDF on the log scale
  The vanilla method would start failing for large mu due to rounding issues
1. Added more tests

According to the reference paper, we should probably use a mixed algorithm to take draws, depending on whether lambda is positive or negative. This might alleviate the need for computing the CDF on the log-scale.

I would still propose to move this distribution to pymc-experimental, but in the meantime we need #5689 and #5688 merged.

@ally-lee If you have the chance, I would appreciate if you could review/comment on the changes I have made.

ricardoV94:
We have a working version of this. Will add it in a new PR, perhaps in pymc-experimental

reddigari:
Any update on the possibility of getting this distribution into pymc or pymc-experimental? Works beautifully.

ricardoV94:
Yeah I'll try to add it next week to pymc-experimental. We found some issues when trying to improve the random method so never came around doing it. But it's definitely usable already

jaharvey8:
Did this ever get added to pymc-experimental? I didn't see it but I might be looking in the wrong place.

jaharvey8:
Awesome, thank you! I had just copied the code and created a new class within my script so this will be nice to have added.

PR #4776, opened 2021-06-17
Added ZeroSumNormal Distribution

kc611:
This PR adds a `ZeroSumNormal` distribution along with a `zerosum` transform. The original code for this is was written by @aseyboldt (probably).

 This is a case of `Normal` distribution which constrains the sum of the means to be zero along the given axes.

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -65,6 +65,7 @@ __all__ = [
     "Lognormal",
     "ChiSquared",
     "HalfNormal",
+    "ZeroSumNormal",
     "Wald",
     "Pareto",
     "InverseGamma",
@@ -922,6 +923,67 @@ class HalfNormal(PositiveContinuous):
             0 <= value,
             0 < sigma,
         )
+
+
+class ZeroSumNormal(Continuous):
+    def __init__(self, sigma=1, zerosum_dims=None, zerosum_axes=None, **kwargs):
+        shape = kwargs.get("shape", ())
+        dims = kwargs.get("dims", None)
+        if isinstance(shape, int):
+            shape = (shape,)
+
+        if isinstance(dims, str):
+            dims = (dims,)
+
+        self.mu = self.median = self.mode = tt.zeros(shape)
+        self.sigma = tt.as_tensor_variable(sigma)
+
+        if zerosum_dims is None and zerosum_axes is None:
+            if shape:
+                zerosum_axes = (-1,)
+            else:
+                zerosum_axes = ()
+
+        if isinstance(zerosum_axes, int):
+            zerosum_axes = (zerosum_axes,)
+
+        if isinstance(zerosum_dims, str):
+            zerosum_dims = (zerosum_dims,)
+
+        if zerosum_axes is not None and zerosum_dims is not None:
+            raise ValueError("Only one of zerosum_axes and zerosum_dims can be specified.")
+
+        if zerosum_dims is not None:
+            if dims is None:
+                raise ValueError("zerosum_dims can only be used with the dims kwargs.")
+            zerosum_axes = []
+            for dim in zerosum_dims:
+                zerosum_axes.append(dims.index(dim))
+        self.zerosum_axes = [a if a >= 0 else len(shape) + a for a in zerosum_axes]
+
+        if "transform" not in kwargs or kwargs["transform"] == None:
+            kwargs["transform"] = transforms.ZeroSumTransform(zerosum_axes)
+
+        super().__init__(**kwargs)
+
+    def logp(self, value):
+        return Normal.dist(sigma=self.sigma).logp(value)
+
+    def _random(self, scale, size):
+        samples = stats.norm.rvs(loc=0, scale=scale, size=size)
+        for axis in self.zerosum_axes:
+            samples -= np.mean(samples, axis=axis, keepdims=True)
+        return samples
+
+    def random(self, point=None, size=None):
+        (sigma,) = draw_values([self.sigma], point=point, size=size)
+        return generate_samples(self._random, scale=sigma, dist_shape=self.shape, size=size)
+
+    def _distr_parameters_for_repr(self):
+        return ["sigma"]
+
+    def logcdf(self, value):
+        raise NotImplementedError()
 
 
 class Wald(PositiveContinuous):
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -926,19 +926,37 @@ class HalfNormal(PositiveContinuous):
 
 
 class ZeroSumNormal(Continuous):
-    def __init__(self, sigma=1, zerosum_dims=None, zerosum_axes=None, **kwargs):
-        shape = kwargs.get("shape", ())
+    def __new__(cls, name, *args, **kwargs):
+        zerosum_axes = kwargs.get("zerosum_axes", None)
+        zerosum_dims = kwargs.get("zerosum_dims", None)
         dims = kwargs.get("dims", None)
-        if isinstance(shape, int):
-            shape = (shape,)
 
+        if isinstance(zerosum_dims, str):
+            zerosum_dims = (zerosum_dims,)
         if isinstance(dims, str):
             dims = (dims,)
+
+        if zerosum_dims is not None:
+            if dims is None:
+                raise ValueError("zerosum_dims can only be used with the dims kwargs.")
+            if zerosum_axes is not None:
+                raise ValueError("Only one of zerosum_axes and zerosum_dims can be specified.")
+            zerosum_axes = []
+            for dim in zerosum_dims:
+                zerosum_axes.append(dims.index(dim))
+            kwargs["zerosum_axes"] = zerosum_axes
+
+        return super().__new__(cls, name, *args, **kwargs)
+
+    def __init__(self, sigma=1, zerosum_axes=None, zerosum_dims=None, **kwargs):
+        shape = kwargs.get("shape", ())
+        if isinstance(shape, int):
+            shape = (shape,)
 
         self.mu = self.median = self.mode = tt.zeros(shape)
         self.sigma = tt.as_tensor_variable(sigma)
 
-        if zerosum_dims is None and zerosum_axes is None:
+        if zerosum_axes is None:
             if shape:
                 zerosum_axes = (-1,)
             else:
@@ -947,21 +965,9 @@ class ZeroSumNormal(Continuous):
         if isinstance(zerosum_axes, int):
             zerosum_axes = (zerosum_axes,)
 
-        if isinstance(zerosum_dims, str):
-            zerosum_dims = (zerosum_dims,)
-
-        if zerosum_axes is not None and zerosum_dims is not None:
-            raise ValueError("Only one of zerosum_axes and zerosum_dims can be specified.")
-
-        if zerosum_dims is not None:
-            if dims is None:
-                raise ValueError("zerosum_dims can only be used with the dims kwargs.")
-            zerosum_axes = []
-            for dim in zerosum_dims:
-                zerosum_axes.append(dims.index(dim))
         self.zerosum_axes = [a if a >= 0 else len(shape) + a for a in zerosum_axes]
 
-        if "transform" not in kwargs or kwargs["transform"] == None:
+        if "transform" not in kwargs or kwargs["transform"] is None:
             kwargs["transform"] = transforms.ZeroSumTransform(zerosum_axes)
 
         super().__init__(**kwargs)
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -98,7 +98,7 @@ class Distribution:
             raise TypeError("observed needs to be data but got: {}".format(type(data)))
         total_size = kwargs.pop("total_size", None)
 
-        dims = kwargs.pop("dims", None)
+        dims = kwargs["dims"] if "dims" in kwargs else None
         has_shape = "shape" in kwargs
         shape = kwargs.pop("shape", None)
         if dims is not None:
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -98,7 +98,7 @@ class Distribution:
             raise TypeError("observed needs to be data but got: {}".format(type(data)))
         total_size = kwargs.pop("total_size", None)
 
-        dims = kwargs["dims"] if "dims" in kwargs else None
+        dims = kwargs.pop("dims", None)
         has_shape = "shape" in kwargs
         shape = kwargs.pop("shape", None)
         if dims is not None:
--- a/pymc3/distributions/transforms.py
+++ b/pymc3/distributions/transforms.py
@@ -14,6 +14,8 @@
 
 import warnings
 
+from typing import List
+
 import numpy as np
 import theano.tensor as tt
 
@@ -565,3 +567,78 @@ class Chain(Transform):
             else:
                 det += det_
         return det
+
+
+def _extend_axis(array, axis):
+    n = array.shape[axis] + 1
+    sum_vals = array.sum(axis, keepdims=True)
+    norm = sum_vals / (np.sqrt(n) + n)
+    fill_val = norm - sum_vals / np.sqrt(n)
+
+    out = tt.concatenate([array, fill_val], axis=axis)
+    return out - norm
+
+
+def _extend_axis_rev(array, axis):
+    if axis < 0:
+        axis = axis % array.ndim
+    assert axis >= 0 and axis < array.ndim
+
+    n = array.shape[axis]
+    last = tt.take(array, [-1], axis=axis)
+
+    sum_vals = -last * np.sqrt(n)
+    norm = sum_vals / (np.sqrt(n) + n)
+    slice_before = (slice(None, None),) * axis
+    return array[slice_before + (slice(None, -1),)] + norm
+
+
+def _extend_axis_val(array, axis):
+    n = array.shape[axis] + 1
+    sum_vals = array.sum(axis, keepdims=True)
+    norm = sum_vals / (np.sqrt(n) + n)
+    fill_val = norm - sum_vals / np.sqrt(n)
+
+    out = np.concatenate([array, fill_val], axis=axis)
+    return out - norm
+
+
+def _extend_axis_rev_val(array, axis):
+    n = array.shape[axis]
+    last = np.take(array, [-1], axis=axis)
+
+    sum_vals = -last * np.sqrt(n)
+    norm = sum_vals / (np.sqrt(n) + n)
+    slice_before = (slice(None, None),) * len(array.shape[:axis])
+    return array[slice_before + (slice(None, -1),)] + norm
+
+
+class ZeroSumTransform(Transform):
+    name = "zerosum"
+
+    _zerosum_axes: List[int]
+
+    def __init__(self, zerosum_axes):
+        self._zerosum_axes = zerosum_axes
+
+    def forward(self, x):
+        for axis in self._zerosum_axes:
+            x = _extend_axis_rev(x, axis=axis)
+        return floatX(x)
+
+    def forward_val(self, x, point):
+        for axis in self._zerosum_axes:
+            x = _extend_axis_rev_val(x, axis=axis)
+        return x
+
+    def backward(self, z):
+        z = tt.as_tensor_variable(z)
+        for axis in self._zerosum_axes:
+            z = _extend_axis(z, axis=axis)
+        return floatX(z)
+
+    def jacobian_det(self, x):
+        return tt.constant(0.0)
+
+
+zerosum = ZeroSumTransform
--- a/pymc3/tests/test_distributions_random.py
+++ b/pymc3/tests/test_distributions_random.py
@@ -343,6 +343,11 @@ class TestHalfNormal(BaseTestCases.BaseTestCase):
     params = {"tau": 1.0}
 
 
+class TestZeroSumNormal(BaseTestCases.BaseTestCase):
+    distribution = pm.ZeroSumNormal
+    params = {"sigma": 1.0}
+
+
 class TestUniform(BaseTestCases.BaseTestCase):
     distribution = pm.Uniform
     params = {"lower": 0.0, "upper": 1.0}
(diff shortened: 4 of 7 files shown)


pymc3/distributions/continuous.py
```diff
@@ -924,6 +925,74 @@ def logcdf(self, value):
...
+                zerosum_axes.append(dims.index(dim))
+        self.zerosum_axes = [a if a >= 0 else len(shape) + a for a in zerosum_axes]
+        self._degrees_of_freedom = np.prod(
+            [s if axis not in zerosum_axes else s - 1 for axis, s in enumerate(shape)]
+        )
+        self._full_size = np.prod(shape)
+        self._rescaling = np.sqrt(self._full_size / self._degrees_of_freedom)
+
+        # if "transform" not in kwargs or kwargs["transform"]==None:
+        #     kwargs["transform"]=transforms.ZeroSumTransform(zerosum_axes)
```
kc611: I'm not sure if this transform is required every single time. If so we'll just have to un-comment these.

pymc3/tests/test_transforms.py
```diff
@@ -242,6 +242,15 @@ def test_chain():
     close_to_logical(np.diff(vals) >= 0, True, tol)
 
 
+def test_zerosum():
+    zerosum_axes = [0]
+    zerosum_transf = tr.ZeroSumTransform(zerosum_axes)
+
+    check_transform(
+        tr.zerosum(zerosum_axes), Simplex(2), constructor=tt.dvector, test=np.random.random(10)
```
kc611: Not sure if this is the right way to test this transform.

twiecki:
This is great, thanks @kc611. Definitely needs a line in the release-notes.

pymc3/distributions/continuous.py
AlexAndorra: I think @lucianopaz can give you a hint here

lucianopaz: Yes, you always have to add the transform. Without it, `pm.sample` wouldn't have any guarantees that the samples would sum to zero.

pymc3/distributions/continuous.py
```diff
@@ -924,6 +925,74 @@ def logcdf(self, value):
...
+                raise ValueError("zerosum_dims can only be used with the dims kwargs.")
+            zerosum_axes = []
+            for dim in zerosum_dims:
+                zerosum_axes.append(dims.index(dim))
+        self.zerosum_axes = [a if a >= 0 else len(shape) + a for a in zerosum_axes]
+        self._degrees_of_freedom = np.prod(
+            [s if axis not in zerosum_axes else s - 1 for axis, s in enumerate(shape)]
+        )
+        self._full_size = np.prod(shape)
+        self._rescaling = np.sqrt(self._full_size / self._degrees_of_freedom)
```
aseyboldt: I'm not so sure if rescaling really is the right thing here. We can define this anyway we like of course, but I think I prefer to have it unscaled, so just `ZeroSumNormal = MVN(0, I - 1/n * J)` where J is a matrix of all zeros.
Where does this rescaling come from?

pymc3/distributions/continuous.py
aseyboldt: Test for None with `is None`.
We always need this.

pymc3/distributions/transforms.py
```diff
@@ -565,3 +565,81 @@ def jacobian_det(self, y):
             else:
                 det += det_
         return det
+
+
+def extend_axis(array, axis):
```
aseyboldt: Those functions should be private.

pymc3/distributions/transforms.py
```diff
@@ -565,3 +565,81 @@ def jacobian_det(self, y):
...
+    n = array.shape[axis]
+    last = np.take(array, [-1], axis=axis)
+
+    sum_vals = -last * np.sqrt(n)
+    norm = sum_vals / (np.sqrt(n) + n)
+    slice_before = (slice(None, None),) * len(array.shape[:axis])
+    return array[slice_before + (slice(None, -1),)] + norm
+
+
+from typing import List
```
aseyboldt: Move to top of file

pymc3/tests/test_distributions_random.py
```diff
@@ -622,6 +627,24 @@ def ref_rand(size, tau):
...
+            )
+            _full_size = np.prod(shape)
+            sigma_factor = np.sqrt(_full_size / _degrees_of_freedom)
+
+            samples = st.norm.rvs(loc=0, scale=sigma * sigma_factor, size=size)
+            for axis in zerosum_axes:
+                samples -= np.mean(samples, axis=axis, keepdims=True)
+            return samples
+
+        pymc3_random(pm.HalfNormal, {"sigma": Rplus}, ref_rand=ref_rand)
```
aseyboldt: Shouldn't this be `pm.ZeroSumNormal`?
Maybe it would be better to draw the samples using `stats.multivariate_normal(cov=sigma ** 2 * (np.eye(n) - np.ones(n)/n))`. That way we check for the correct covariance.

pymc3/distributions/continuous.py
```diff
@@ -924,6 +925,74 @@ def logcdf(self, value):
...
+
+        # if "transform" not in kwargs or kwargs["transform"]==None:
+        #     kwargs["transform"]=transforms.ZeroSumTransform(zerosum_axes)
+
+        super().__init__(**kwargs)
+
+    def logp(self, value):
+        return Normal.dist(sigma=self.sigma / self._rescaling).logp(value)
```
aseyboldt: I think this would be cleaner if we could define the logp on the transformed space. I don't think we implemented a way of doing this yet, did we?

aseyboldt (changes requested):
Thanks for making the PR!
I added a couple of comments.
We should also add some docs, and include a definition of what distribution this actually has (sigma **2 * (I - 1/n J), where I = eye(n) and J = ones(n).

I think we should have tests that use the zerosum_dims and zerosum_axes parameters.

I'm also not perfectly happy with those two names, can't think of anything better though. If anyone else has an idea...

Since I wrote the original code, could you add me as a co-author? (https://github.blog/2018-01-29-commit-together-with-co-authors/)

twiecki:
CC @tomicapretto this will be super neat for Bambi where we don't have to define a base-condition and do relative coding for the others.

pymc3/tests/test_distributions_random.py
kc611: Thanks for the tip, did just that.

pymc3/distributions/continuous.py
kc611: Alright I removed all the rescaling related architecture.

twiecki:
@kc611 Any progress on this? seems like the tests are not passing yet.

kc611:
Sorry for late response, and yeah this is a shapes problem, the `ZeroSumNormal` is adding one extra dimension to the output shape. cc @aseyboldt Is this (extra dimension in result) intended (the case in which the tests would need a change) or do we need an `np.squeeze` somewhere ?

AlexAndorra:
Just an idea in passing: this distribution should also be added in v4 -- would be a shame to not have it in the new version, especially as it's very useful!
Note that I don't want to add to your to-dos @kc611. Do it if you feel like it, otherwise it can be done by someone else in another PR

pymc3/distributions/continuous.py
```diff
@@ -924,6 +925,67 @@ def logcdf(self, value):
...
+        return Normal.dist(sigma=self.sigma).logp(value)
+
+    def _random(self, scale, size):
+        samples = stats.norm.rvs(loc=0, scale=scale, size=size)
+        for axis in self.zerosum_axes:
+            samples -= np.mean(samples, axis=axis, keepdims=True)
+        return samples
+
+    def random(self, point=None, size=None):
+        sigma = draw_values([self.sigma], point=point, size=size)
```
lucianopaz:
```suggestion
        sigma, = draw_values([self.sigma], point=point, size=size)
```

draw_values returns a list of samples for the distribution parameters. We weren’t unpacking that list to get the value of `sigma`, and this ended up adding an extra dimension to the samples.

pymc3/distributions/continuous.py
```diff
@@ -924,6 +925,67 @@ def logcdf(self, value):
...
+                zerosum_axes.append(dims.index(dim))
+        self.zerosum_axes = [a if a >= 0 else len(shape) + a for a in zerosum_axes]
+
+        if "transform" not in kwargs or kwargs["transform"] == None:
+            kwargs["transform"] = transforms.ZeroSumTransform(zerosum_axes)
+
+        super().__init__(**kwargs)
+
+    def logp(self, value):
+        return Normal.dist(sigma=self.sigma).logp(value)
```
lucianopaz, line 976: If we don’t add the scaling of sigma here, our `random` method will be inconsistent with the logp

lucianopaz (changes requested):

pymc3/distributions/continuous.py
kc611: How so ? Isn't `random` also drawing samples using `self.sigma` ?

twiecki:
@kc611 Where are we on this?

zoj613:
The name of the class suggests that the samples themselves are to sum to zero, not the means as explained in the PR description. Would a name like `ZeroMeanSumNormal/MeanZeroSumNormal` not be an alternative in this case?

twiecki:
Or just `MeanZeroNormal`?

kc611:
It seemed like a simple type filtering issue. But thanks for reminding. Anyway put it on testing right now.

What should we rename he distribution to ? ( For now I've let it be `ZeroSumNormal` )

drbenvincent:
FYI this is being showcased in a PR to `PyMC3-examples` here https://github.com/pymc-devs/pymc-examples/pull/210
In that example I noticed graphviz comes out like this
Screenshot 2021-08-16 at 18 34 01
Not mission critical, but it would be pretty neat if it displayed as a specific `\beta \sim \text{ZeroSumNormal}` distribution node 😎
Please do let me know if I've not got the attributions/credits right.

pymc3/distributions/distribution.py
```diff
@@ -98,7 +98,7 @@ def __new__(cls, name, *args, **kwargs):
             raise TypeError("observed needs to be data but got: {}".format(type(data)))
         total_size = kwargs.pop("total_size", None)
 
-        dims = kwargs.pop("dims", None)
+        dims = kwargs["dims"] if "dims" in kwargs else None
```
aseyboldt: This is a breaking change to the `pm.Distribution` API. Right now you can define new distributions that do not accept a dims kwargs, but with this they have to accept it. Maybe we can live with this, but at the very least this should be mentioned in the release notes somewhere.

pymc3/distributions/continuous.py
aseyboldt: This logp is somewhat strange still. From the math side this should be pm.MvNormal with cov=I - J / n where J is a matrix of all 1s. We don't want to write it like this though, because we don't want to do matrix factorization, and pm.MvNormal doesn't work if an eigenvalues is 0.
It would be great, if instead we could define the logp simply in the transformed space. This would imply changes to TransformedDistribution though.

aseyboldt (changes requested):

aseyboldt:
@zoj613 The description is a bit strange with the "which constrains the sum of the means to be zero". It really constrains the sum (and equivalently mean) to zero.

aseyboldt:
@drbenvincent The strange graphviz output is because the other PR is using an old version of the ZeroSumNormal that is implemented as a deterministic.

aseyboldt:
Oh, and I think we have to check if the axes length is 1 or 0 and handles those cases correctly, or throw a good error.

zoj613:
Im wondering how Algorithm 2 presented in [this paper](https://projecteuclid.org/journals/bayesian-analysis/volume-12/issue-4/Fast-Simulation-of-Hyperplane-Truncated-Multivariate-Normal-Distributions/10.1214/17-BA1052.full) would compare to the current implementation....or if it would be useful?
The `G` matrix would be a `n by 1` vector of 1's in this case, and I think the covariance would be a diagonal matrix with entries equal to the variance of the samples. The algorithm would likely reduce to just scalar calculations for the given `G`. Any opinions on this?

twiecki:
@kc611 Any progress on this?

pymc3/distributions/continuous.py
```diff
@@ -924,6 +925,67 @@ def logcdf(self, value):
...
+            dims = (dims,)
+
+        self.mu = self.median = self.mode = tt.zeros(shape)
+        self.sigma = tt.as_tensor_variable(sigma)
+
+        if zerosum_dims is None and zerosum_axes is None:
+            if shape:
+                zerosum_axes = (-1,)
+            else:
+                zerosum_axes = ()
```
lucianopaz, line 963: I think it makes no sense to have a `ZeroSumNormal` when `shape=()` or `None`. In that case, the RV should also be exactly equal to zero. I think that we should test if `shape is None or len(shape) == 0` and raise a `ValueError` in that case. Something that says, `ZeroSumNormal` is defined only for RVs that are not scalar.

pymc3/distributions/distribution.py
AlexAndorra: I reverted that change, as we're now using `__new__` to enable the `zerosum_dims` kwarg

pymc3/distributions/continuous.py
```diff
@@ -924,6 +925,73 @@ def logcdf(self, value):
...
+            if zerosum_axes is not None:
+                raise ValueError("Only one of zerosum_axes and zerosum_dims can be specified.")
+            zerosum_axes = []
+            for dim in zerosum_dims:
+                zerosum_axes.append(dims.index(dim))
+            kwargs["zerosum_axes"] = zerosum_axes
+
+        return super().__new__(cls, name, *args, **kwargs)
+
+    def __init__(self, sigma=1, zerosum_axes=None, zerosum_dims=None, **kwargs):
```
AlexAndorra, line 951: Note that `zerosum_dims` is not used in `__init__`, but if I don't put it here, it doesn't seem to be passed on to `__new__`: `TypeError: __init__() got an unexpected keyword argument 'zerosum_dims'`
Not sure we can do it otherwise though. If someone has a better idea, I'm all ears

AlexAndorra:
@aseyboldt and I have implemented the `zerosum_dims` kwarg (with the caveat I put in my message just above; not sure how big a deal it is.

IIUC, only two things are blocking merge for now:
1. Adding a test of this `zerosum_dims` feature
2. Deciding on the parametrization we wanna use for `sigma`

We'll be looking at those during a hackathon this Friday

twiecki:
Why even add this to v3 and not v4 directly?

AlexAndorra:
The best reason I've found is sunk-cost fallacy 😅 
TBH I agree: let's only add that in v4. That means no duplicated efforts and it nudges people towards v4

aseyboldt:
About the meaning of the (scalar) `sigma` parameter, and whether that should be the std of the underlying normal, or the std of the marginals (cc @lucianopaz):

I think we have basically two options about what `sigma` could refer to:

- The marginal std, so `pm.ZeroSumNormal(sigma=1).random(1000).std() == 1`. We get `C = sigma ** 2 * (I - I / n) / (1 - 1/n)` as covariance.
- The sqrt of the non-zero eigenvalues of the covariance matrix. This would imply `pm.ZeroSumNormal(sigma=1).random().std() == sqrt(1 - 1 / n))`. We get `C = sigma**2 * (I - I / n)` as covariance.

The obvious downside of the second option would be that it might be confusing to users, as the marginal std is not what people might expect. But I think there are reasons for the second option, and I think I'd prefer that one. It will be very confusing if we decide to change it later however, so we should give it some thought.

Let's assume we use the ZeroSumNormal in a regression (all sigma values in terms of the second option):

```python
with pm.Model() as model:
    intercept = pm.Normal("intercept", sd=10)
    group_effect = pm.Normal("group_effect", sd=1, dims="group")
    
    mu = (
        intercept
        + group_effect[group_idx]
    )
```

We can rewrite this as this (without changing the model, only the parametrization is different:
```python
with pm.Model() as model:
    intercept = pm.Normal("intercept", sd=10)

    mean_sigma = at.sqrt(1 ** 2 / len(groups))
    group_effect_mean = pm.Normal("group_effect_mean", sigma=mean_sigma)
    group_effect_diff = pm.ZeroSumNormal("group_effect_diff", sigma=1, dims="group")
    
    group_effect = pm.Deterministic("group_effect", group_effect_mean + group_effect_diff)
    
    mu = (
        intercept
        + group_effect_mean
        + group_effect_diff[group_idx]
    )
```

Now we use the fact that the sum of two normal dists is a normal again (the intercept + group_effect_mean):

```python
with pm.Model() as model:
    mean_sigma = at.sqrt(1 ** 2 / len(groups))
    intercept_plus_mean = pm.Normal("intercept", sd=tt.sqrt(10 ** 2 + mean_sigma ** 2))

    group_effect_diff = pm.ZeroSumNormal("group_effect_diff", sigma=1, dims="group")

    mu = (
        intercept_plus_mean
        + group_effect_diff[group_idx]
    )
```

So we turned our model that assumed that `group_effect ~ normal(0, 1)` into a model using the zerosumnormal, but interestingly the sigma value is still 1, independent of the number of groups. If we had used the first option for sigma, we would have to set sigma to something that depends on the number of groups.

AlexAndorra:
Thanks for this detailed explanation @aseyboldt ! I remember that discussion from a late-night train commute in Lisbon 😅 
So, IIUC:
- The plus of the first option is that it's easy to understand and parametrize: the resulting `ZeroSumNormal`'s marginals have std = 1. The downside is that to make sure of that, `ZeroSumNormal`'s sigma's prior has to depend on the number of groups (which may in turn hurt users' understanding of the parametrization).
- The plus of the second option is that the sigma prior doesn't depend on the number of groups, but the interpretation of the marginals (the sqrt of the non-zero eigenvalues of the covariance matrix) is... not self-evident, to put it gently.

Is my summary mistaken in any way?

pymc3/distributions/continuous.py
aseyboldt: I think the zerosum_dims is probably still in kwargs from line 949? We could just remove there.

aseyboldt:
I think that's a good summary :-)

A related question is what we would want to happen if sigma is not a scalar. Maybe the safest thing is to just not allow that for now.

AlexAndorra:
That would mean a different sigma for each group?

twiecki:
@AlexAndorra Any plans to revive this?

AlexAndorra:
Yeah yeah, it's still on my radar but this week is very busy work-wise. Hope to start next week. In any case, it's not gonna be a short PR

pymc3/distributions/continuous.py
```diff
@@ -924,6 +925,73 @@ def logcdf(self, value):
...
+        if zerosum_axes is None:
+            if shape:
+                zerosum_axes = (-1,)
+            else:
+                zerosum_axes = ()
+
+        if isinstance(zerosum_axes, int):
+            zerosum_axes = (zerosum_axes,)
+
+        self.zerosum_axes = [a if a >= 0 else len(shape) + a for a in zerosum_axes]
```
lucianopaz, line 968: Enforcing positive axis here leads to problems when you draw samples from the prior predictive. It's better to replace this line with this
```suggestion
        self.zerosum_axes = [a if a < 0 else a - len(shape) for a in zerosum_axes]
```

lucianopaz (changes requested):
I'm picking up some slack here because I started looking into porting the ZeroSumNormal into v4 and I found a small bug with prior predictive sampling.
We should add a test for the random method that asserts that we get `allclose` results to 0 on the axes that we want to sum to zero

pymc3/distributions/continuous.py
lucianopaz: We can avoid the `MvNormal.logp` problem if we force `self.sigma` to be a scalar or to have a single element in all the `zerosum_axes`. In this case, all the directions in the zerosum manifold are uncorrelated and have equal variance. This means that we can use the `Normal.logp` as long as we also include a `bound` condition that guarantees that we are on the zerosum manifold. You can do this by using this logp:

```suggestion
        zerosums = [tt.all(tt.abs_(tt.mean(x, axis=axis)) <= 1e-9) for axis in self.zerosum_axes]
        return bound(
            pm.Normal.dist(sigma=self.sigma).logp(x),
            tt.all(self.sigma > 0),
            broadcast_conditions=False,
            *zerosums,
        )
```

lucianopaz (changes requested):
I also wanted to suggest how to change the `logp`

pymc3/distributions/continuous.py
lucianopaz: I came across [this wiki section](https://en.wikipedia.org/wiki/Multivariate_normal_distribution#Degenerate_case) that talks about the degenerate MvNormal case (which is what we have with the ZeroSumNormal). We could use that formula as the expected logp value and test if the `logp` that we are using in the distribution matches it. The expected logp would look something like this:

```python
def pseudo_log_det(A, tol=1e-13):
    v, w = np.linalg.eigh(A)
    return np.sum(np.log(np.where(np.abs(v) >= tol, v, 1)), axis=-1)

def logp(value, sigma):
    n = value.shape[-1]
    cov = np.asarray(sigma)[..., None, None]**2 * (np.eye(n) - np.ones((n, n)) / n)
    psdet = 0.5 *  pseudo_log_det(2 * np.pi * cov)
    exp = 0.5 * (value[..., None, :] @ np.linalg.pinv(cov) @ value[..., None])[..., 0, 0]
    return np.where(np.abs(np.sum(value, axis=-1)) < 1e-9, -psdet - exp, -np.inf)
```

lucianopaz: I ran a few tests with the logp and it looks like the `logp` that we are using in this PR, doesn't match what one would expect from a degenerate multivariate normal distribution. In my comment above, I posted what a degenerate MvNormal `logp` looks like. For this particular problem, where we know that we have only one eigenvector with zero eigenvalue, we can re-write the logp as:

```python
def logp(value, sigma):
    n = value.shape[-1]
    cov = np.asarray(sigma)[..., None, None]**2 * (np.eye(n) - np.ones((n, n)) / n)
    v, w = np.linalg.eigh(cov)
    psdet =  0.5 * (np.sum(np.log(v[..., 1:])) + (n - 1) * np.log(2 * np.pi))
    cov_pinv = w[:, 1:] @ np.diag(1 / v[1:]) @ w[:, 1:].T
    exp = 0.5 * (value[..., None, :] @ cov_pinv @ value[..., None])[..., 0, 0]
    return np.where(np.abs(np.sum(value, axis=-1)) < 1e-9, -psdet - exp, -np.inf)
```

This is different from the `logp` that we are currently using in this PR. The difference is in the normalization constant:
`psdet =  0.5 * (np.sum(np.log(v[..., 1:])) + (n - 1) * np.log(2 * np.pi))`. In particular, since, all eigenvalues `v` except the first one are the same and are equal to `sigma**2`, `psdet = (n - 1) * (0.5 * np.log(2 * np.pi) + np.log(np.sigma))`. Whereas, with the assumed `pm.Normal.dist(sigma=self.sigma).logp(x)` the normalization factor we are getting is:

```python
psdet = n * (0.5 * np.log(2 * np.pi) + np.log(np.sigma))
```

This means that we have to multiply the `logp` that we are using by (n-1)/n (in the case where only one axis sums to zero) to get the correct log probability density. I'll check what happens when more than one axes has to zerosum.

ricardoV94:
With V4 closer and closer to being out, this PR makes less sense. I'll close it soon if nobody objects.

Possibly, we should add the ZeroSumNormal to pymc-experimental anyway.

twiecki:
Agreed.

PR #4596, opened 2021-03-31
CAR random variables

aerubanov:
close #4518 
I Changed CAR distribution class to compatible with ```v4``` ```RandomVariable``` and implemented CAR random variable class. But now I need some advises for further improvements. 

 @brandonwillard, I have few questions about random variable for CAR:
- if some parameter of random variable have param that may be scalar or vector (like ```mu```, ```tau``` and ```alpa```), what I should specify in ```ndims_params```? If I pass scalar instead vector  an error occurred.
- Should I somehow handle size param in ```rng_fn``` method?
- Where I should store intermediate variables (like ```D``` and ```lam```) that need to be calculated only once.  Initially they were calculated in the ```__init__``` method of distribution class and stored as distribution class atribute.

@ckrapu, Do you have any ideas for testing the implementation of the algorithm?

PR #701, opened 2015-04-27
Add Wishart with Bartlett decomposition.

twiecki:
Closes https://github.com/pymc-devs/pymc3/issues/538.

This adds the Wishart distribution that can be sampled from because it's using matrix decomposition to sample from an unconstrained space. 
However, it's a bit odd as it adds RVs itself to the model. I think it's pretty convenient though. @jsalvatier @fonnesbeck thoughts on adding this?

PR #4777, opened 2021-06-17
Refactored Wishart and MatrixNormal distribution

kc611:
This PR refactors the `Wishart` and `MatrixNormal` multivariate distributions.

### DIFF
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -47,6 +47,7 @@ from pymc3.distributions import transforms
 from pymc3.distributions.continuous import ChiSquared, Normal, assert_negative_support
 from pymc3.distributions.dist_math import bound, factln, logpow, multigammaln
 from pymc3.distributions.distribution import Continuous, Discrete
+from pymc3.distributions.shape_utils import broadcast_dist_samples_to, to_tuple
 from pymc3.math import kron_diag, kron_dot
 
 __all__ = [
@@ -739,6 +740,26 @@ class PosDefMatrix(Op):
 matrix_pos_def = PosDefMatrix()
 
 
+class WishartRV(RandomVariable):
+    name = "wishart"
+    ndim_supp = 2
+    ndims_params = [0, 2]
+    dtype = "floatX"
+    _print_name = ("Wishart", "\\operatorname{Wishart}")
+
+    def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
+        # The shape of second parameter `V` defines the shape of the output.
+        return dist_params[1].shape
+
+    @classmethod
+    def rng_fn(cls, rng, nu, V, size=None):
+        size = size if size else 1  # Default size for Scipy's wishart.rvs is 1
+        return stats.wishart.rvs(np.int(nu), V, size=size, random_state=rng)
+
+
+wishart = WishartRV()
+
+
 class Wishart(Continuous):
     r"""
     Wishart log-likelihood.
@@ -775,9 +796,13 @@ class Wishart(Continuous):
     This distribution is unusable in a PyMC3 model. You should instead
     use LKJCholeskyCov or LKJCorr.
     """
+    rv_op = wishart
 
-    def __init__(self, nu, V, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    @classmethod
+    def dist(cls, nu, V, *args, **kwargs):
+        nu = at.as_tensor_variable(intX(nu))
+        V = at.as_tensor_variable(floatX(V))
+
         warnings.warn(
             "The Wishart distribution can currently not be used "
             "for MCMC sampling. The probability of sampling a "
@@ -787,34 +812,13 @@ class Wishart(Continuous):
             "https://github.com/pymc-devs/pymc3/issues/538.",
             UserWarning,
         )
-        self.nu = nu = at.as_tensor_variable(nu)
-        self.p = p = at.as_tensor_variable(V.shape[0])
-        self.V = V = at.as_tensor_variable(V)
-        self.mean = nu * V
-        self.mode = at.switch(at.ge(nu, p + 1), (nu - p - 1) * V, np.nan)
 
-    def random(self, point=None, size=None):
-        """
-        Draw random values from Wishart distribution.
+        # mean = nu * V
+        # p = V.shape[0]
+        # mode = at.switch(at.ge(nu, p + 1), (nu - p - 1) * V, np.nan)
+        return super().dist([nu, V], *args, **kwargs)
 
-        Parameters
-        ----------
-        point: dict, optional
-            Dict of variable values on which random values are to be
-            conditioned (uses default point if not specified).
-        size: int, optional
-            Desired size of random sample (returns one sample if not
-            specified).
-
-        Returns
-        -------
-        array
-        """
-        # nu, V = draw_values([self.nu, self.V], point=point, size=size)
-        # size = 1 if size is None else size
-        # return generate_samples(stats.wishart.rvs, nu.item(), V, broadcast_shape=(size,))
-
-    def logp(self, X):
+    def logp(X, nu, V):
         """
         Calculate log-probability of Wishart distribution
         at specified value.
@@ -828,9 +832,8 @@ class Wishart(Continuous):
         -------
         TensorVariable
         """
-        nu = self.nu
-        p = self.p
-        V = self.V
+
+        p = V.shape[0]
 
         IVI = det(V)
         IXI = det(X)
@@ -1445,6 +1448,36 @@ class LKJCorr(Continuous):
         return ["eta", "n"]
 
 
+class MatrixNormalRV(RandomVariable):
+    name = "matrixnormal"
+    ndim_supp = 2
+    ndims_params = [2, 2, 2]
+    dtype = "floatX"
+    _print_name = ("MatrixNormal", "\\operatorname{MatrixNormal}")
+
+    @classmethod
+    def rng_fn(cls, rng, mu, rowchol, colchol, size=None):
+
+        size = to_tuple(size)
+        dist_shape = to_tuple([rowchol.shape[0], colchol.shape[0]])
+        output_shape = size + dist_shape
+
+        # Broadcasting all parameters
+        (mu,) = broadcast_dist_samples_to(to_shape=output_shape, samples=[mu], size=size)
+        rowchol = np.broadcast_to(rowchol, shape=size + rowchol.shape[-2:])
+
+        colchol = np.broadcast_to(colchol, shape=size + colchol.shape[-2:])
+        colchol = np.swapaxes(colchol, -1, -2)  # Take transpose
+
+        standard_normal = rng.standard_normal(output_shape)
+        samples = mu + np.matmul(rowchol, np.matmul(standard_normal, colchol))
+
+        return samples
+
+
+matrixnormal = MatrixNormalRV()
+
+
 class MatrixNormal(Continuous):
     r"""
     Matrix-valued normal log-likelihood.
@@ -1533,153 +1566,63 @@ class MatrixNormal(Continuous):
             vals = pm.MatrixNormal('vals', mu=mu, colchol=colchol, rowcov=rowcov,
                                    observed=data, shape=(m, n))
     """
+    rv_op = matrixnormal
 
-    def __init__(
-        self,
-        mu=0,
+    @classmethod
+    def dist(
+        cls,
+        mu,
         rowcov=None,
         rowchol=None,
-        rowtau=None,
         colcov=None,
         colchol=None,
-        coltau=None,
         shape=None,
         *args,
         **kwargs,
     ):
-        self._setup_matrices(colcov, colchol, coltau, rowcov, rowchol, rowtau)
-        if shape is None:
-            raise TypeError("shape is a required argument")
-        assert len(shape) == 2, "shape must have length 2: mxn"
-        self.shape = shape
-        super().__init__(shape=shape, *args, **kwargs)
-        self.mu = at.as_tensor_variable(mu)
-        self.mean = self.median = self.mode = self.mu
-        self.solve_lower = solve_lower_triangular
-        self.solve_upper = solve_upper_triangular
 
-    def _setup_matrices(self, colcov, colchol, coltau, rowcov, rowchol, rowtau):
         cholesky = Cholesky(lower=True, on_error="raise")
 
-        # Among-row matrices
-        if len([i for i in [rowtau, rowcov, rowchol] if i is not None]) != 1:
+        if mu.ndim == 1:
             raise ValueError(
-                "Incompatible parameterization. "
-                "Specify exactly one of rowtau, rowcov, "
-                "or rowchol."
+                "1x1 Matrix was provided. Please use Normal distribution for such cases."
+            )
+
+        # Among-row matrices
+        if len([i for i in [rowcov, rowchol] if i is not None]) != 1:
+            raise ValueError(
+                "Incompatible parameterization. Specify exactly one of rowcov, or rowchol."
             )
         if rowcov is not None:
-            self.m = rowcov.shape[0]
-            self._rowcov_type = "cov"
-            rowcov = at.as_tensor_variable(rowcov)
             if rowcov.ndim != 2:
                 raise ValueError("rowcov must be two dimensional.")
-            self.rowchol_cov = cholesky(rowcov)
-            self.rowcov = rowcov
-        elif rowtau is not None:
-            raise ValueError("rowtau not supported at this time")
-            self.m = rowtau.shape[0]
-            self._rowcov_type = "tau"
-            rowtau = at.as_tensor_variable(rowtau)
-            if rowtau.ndim != 2:
-                raise ValueError("rowtau must be two dimensional.")
-            self.rowchol_tau = cholesky(rowtau)
-            self.rowtau = rowtau
+            rowchol_cov = cholesky(rowcov)
         else:
-            self.m = rowchol.shape[0]
-            self._rowcov_type = "chol"
             if rowchol.ndim != 2:
                 raise ValueError("rowchol must be two dimensional.")
-            self.rowchol_cov = at.as_tensor_variable(rowchol)
+            rowchol_cov = at.as_tensor_variable(rowchol)
 
         # Among-column matrices
-        if len([i for i in [coltau, colcov, colchol] if i is not None]) != 1:
+        if len([i for i in [colcov, colchol] if i is not None]) != 1:
             raise ValueError(
-                "Incompatible parameterization. "
-                "Specify exactly one of coltau, colcov, "
-                "or colchol."
+                "Incompatible parameterization. Specify exactly one of colcov, or colchol."
             )
         if colcov is not None:
-            self.n = colcov.shape[0]
-            self._colcov_type = "cov"
             colcov = at.as_tensor_variable(colcov)
             if colcov.ndim != 2:
                 raise ValueError("colcov must be two dimensional.")
-            self.colchol_cov = cholesky(colcov)
-            self.colcov = colcov
-        elif coltau is not None:
-            raise ValueError("coltau not supported at this time")
-            self.n = coltau.shape[0]
-            self._colcov_type = "tau"
-            coltau = at.as_tensor_variable(coltau)
-            if coltau.ndim != 2:
-                raise ValueError("coltau must be two dimensional.")
-            self.colchol_tau = cholesky(coltau)
-            self.coltau = coltau
+            colchol_cov = cholesky(colcov)
         else:
-            self.n = colchol.shape[0]
-            self._colcov_type = "chol"
             if colchol.ndim != 2:
                 raise ValueError("colchol must be two dimensional.")
-            self.colchol_cov = at.as_tensor_variable(colchol)
+            colchol_cov = at.as_tensor_variable(colchol)
 
-    def random(self, point=None, size=None):
-        """
-        Draw random values from Matrix-valued Normal distribution.
+        mu = at.as_tensor_variable(floatX(mu))
+        # mean = median = mode = mu
 
-        Parameters
-        ----------
-        point: dict, optional
-            Dict of variable values on which random values are to be
-            conditioned (uses default point if not specified).
-        size: int, optional
-            Desired size of random sample (returns one sample if not
-            specified).
+        return super().dist([mu, rowchol_cov, colchol_cov], **kwargs)
 
-        Returns
-        -------
-        array
-        """
-        # mu, colchol, rowchol = draw_values(
-        #     [self.mu, self.colchol_cov, self.rowchol_cov], point=point, size=size
-        # )
-        # size = to_tuple(size)
-        # dist_shape = to_tuple(self.shape)
-        # output_shape = size + dist_shape
-        #
-        # # Broadcasting all parameters
-        # (mu,) = broadcast_dist_samples_to(to_shape=output_shape, samples=[mu], size=size)
-        # rowchol = np.broadcast_to(rowchol, shape=size + rowchol.shape[-2:])
-        #
-        # colchol = np.broadcast_to(colchol, shape=size + colchol.shape[-2:])
-        # colchol = np.swapaxes(colchol, -1, -2)  # Take transpose
-        #
-        # standard_normal = np.random.standard_normal(output_shape)
-        # samples = mu + np.matmul(rowchol, np.matmul(standard_normal, colchol))
-        # return samples
-
-    def _trquaddist(self, value):
-        """Compute Tr[colcov^-1 @ (x - mu).T @ rowcov^-1 @ (x - mu)] and
-        the logdet of colcov and rowcov."""
-
-        delta = value - self.mu
-        rowchol_cov = self.rowchol_cov
-        colchol_cov = self.colchol_cov
-
-        # Find exponent piece by piece
-        right_quaddist = self.solve_lower(rowchol_cov, delta)
-        quaddist = at.nlinalg.matrix_dot(right_quaddist.T, right_quaddist)
-        quaddist = self.solve_lower(colchol_cov, quaddist)
-        quaddist = self.solve_upper(colchol_cov.T, quaddist)
-        trquaddist = at.nlinalg.trace(quaddist)
-
-        coldiag = at.diag(colchol_cov)
-        rowdiag = at.diag(rowchol_cov)
-        half_collogdet = at.sum(at.log(coldiag))  # logdet(M) = 2*Tr(log(L))
-        half_rowlogdet = at.sum(at.log(rowdiag))  # Using Cholesky: M = L L^T
-        return trquaddist, half_collogdet, half_rowlogdet
-
-    def logp(self, value):
+    def logp(value, mu, rowchol, colchol):
         """
         Calculate log-probability of Matrix-valued Normal distribution
         at specified value.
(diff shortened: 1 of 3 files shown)


pymc3/distributions/multivariate.py
```diff
@@ -1531,175 +1560,96 @@ class MatrixNormal(Continuous):
...
+    def dist(
+        cls,
+        mu,
         rowcov=None,
         rowchol=None,
-        rowtau=None,
         colcov=None,
         colchol=None,
-        coltau=None,
         shape=None,
```
kc611, line 1579: The `rowtau` and `choltau` arguments are not supported right now. Maybe there was a plan for supporting those (but was maybe abandoned midway?).

pymc3/tests/test_distributions.py
```diff
@@ -1893,8 +1893,7 @@ def test_mvnormal_init_fail(self):
             with pytest.raises(ValueError):
                 x = MvNormal("x", mu=np.zeros(3), cov=np.eye(3), tau=np.eye(3), size=3)
 
-    @pytest.mark.parametrize("n", [1, 2, 3])
-    @pytest.mark.xfail(reason="Distribution not refactored yet")
+    @pytest.mark.parametrize("n", [2, 3])
```
kc611, line 1896: This fails for n=1 , maybe the dimensions are being downcasted to a scalar when matrix is shaped `1x1`.

pymc3/tests/test_distributions_random.py
```diff
@@ -1306,6 +1303,56 @@ class TestOrderedProbit(BaseTestDistribution):
...
+    reference_dist = lambda self: functools.partial(
+        self.matrixnormal_rng_fn, rng=self.get_random_state()
+    )
+    sizes_to_check = [None, (1), (2, 3)]
+    sizes_expected = [(3,), (1, 3), (2, 3, 3)]
+
+    tests_to_run = [
+        "check_rv_size",
+        "check_pymc_params_match_rv_op",
+        "check_pymc_draws_match_reference",
```
kc611, line 1352: This test fails, strangely the first element matches everytime. @brandonwillard is this related to https://github.com/aesara-devs/aesara/issues/209 ? There seems to a similar operation involved here.

pymc3/tests/test_distributions_random.py
```diff
@@ -1306,6 +1304,71 @@ class TestOrderedProbit(BaseTestDistribution):
...
+        (mu,) = broadcast_dist_samples_to(to_shape=output_shape, samples=[mu], size=size)
+        rowchol = np.broadcast_to(rowchol, shape=size + rowchol.shape[-2:])
+
+        colchol = np.broadcast_to(colchol, shape=size + colchol.shape[-2:])
+        colchol = np.swapaxes(colchol, -1, -2)  # Take transpose
+
+        standard_normal = rng.standard_normal(output_shape)
+        samples = mu + np.matmul(rowchol, np.matmul(standard_normal, colchol))
+
+        return samples
```
kc611: Kind-of counter-intuitive to use this function here as reference, since it's basically same as the internal one, but it looks like `st.matrix_normal.rvs(mean=mu, rowcov=rowcov, colcov=colcov, size=size)` used previously resulted in draws that are differently valued than our implementation and our previous testing just checked if the values came from similar distributions. (See previous commit https://github.com/pymc-devs/pymc3/pull/4777/commits/4ee1360158885556c7801b97bc166ae2a40c37bd).

pymc3/distributions/multivariate.py
brandonwillard: They weren't supported in `v3`?

kc611: No, passing either of `rowtau` or `choltau` raised an error. (It's in the code right below this)

pymc3/tests/test_distributions_random.py
ricardoV94: Does this help? https://github.com/pymc-devs/pymc3/blob/fe115b857ae42e84d1f359153fd1f7efb2c327ae/docs/source/distribution_implementation.md#notes-on-check_pymcs_draws_match_reference-test

ricardoV94: In this case, it may be better to use something along the lines of the `pymc3_random` test and compare with the Scipy version. Maybe compare dimension by dimension since it is a multivariate distribution.

kc611: In that case we can add a seperate class which does this. (Similar to how we did with `Interpolated` distribution )

pymc3/tests/test_distributions.py
brandonwillard: If that `n` is being used to determine the number of output dimensions, then `n=1` might not make sense for a matrix-valued distribution like this.

pymc3/tests/test_distributions_random.py
ricardoV94: You can also add it in that class to keep things organized, such as this: https://github.com/pymc-devs/pymc3/blob/f4343ca7ba112f843cf256378e88ff80a25342a6/pymc3/tests/test_distributions_random.py#L385-L390

pymc3/distributions/multivariate.py
```diff
@@ -737,6 +738,21 @@ def __str__(self):
 matrix_pos_def = PosDefMatrix()
 
 
+class WishartRV(RandomVariable):
+    name = "wishart"
+    ndim_supp = 0
```
brandonwillard: Shouldn't this be `ndim_supp = 2` (i.e. the Wishart distribution is over matrices)?

pymc3/distributions/multivariate.py
```diff
@@ -785,34 +802,15 @@ def __init__(self, nu, V, *args, **kwargs):
...
-        Parameters
-        ----------
-        point: dict, optional
-            Dict of variable values on which random values are to be
-            conditioned (uses default point if not specified).
-        size: int, optional
-            Desired size of random sample (returns one sample if not
-            specified).
+        nu = at.as_tensor_variable(nu)
+        V = at.as_tensor_variable(V)
```
```suggestion
        nu = at.as_tensor_variable(intX(nu))
        V = at.as_tensor_variable(floatX(V))
```

pymc3/tests/test_distributions.py
kc611: I raised an error if such a matrix is passed.

pymc3/distributions/multivariate.py
```diff
@@ -1531,175 +1558,101 @@ class MatrixNormal(Continuous):
...
         else:
-            self.n = colchol.shape[0]
-            self._colcov_type = "chol"
             if colchol.ndim != 2:
                 raise ValueError("colchol must be two dimensional.")
-            self.colchol_cov = at.as_tensor_variable(colchol)
+            colchol_cov = at.as_tensor_variable(colchol)
 
-    def random(self, point=None, size=None):
+        mu = at.as_tensor_variable(mu)
```
```suggestion
        mu = at.as_tensor_variable(floatX(mu))
```

pymc3/tests/test_distributions.py
```diff
@@ -1907,6 +1906,8 @@ def test_matrixnormal(self, n):
                 "colcov": PdMatrix(n) * mat_scale,
             },
             matrix_normal_logpdf_cov,
+            extra_args={"size": n},
+            decimal=select_by_precision(float64=5, float32=5),
```
ricardoV94: Do the tests pass if aesara `floatX` is set to "float32"? Usually the default precision we use for float32 is `3`. Your change is increasing the precision compared to what we were testing before.

ricardoV94: Also make sure you test all combinations locally by setting `n_samples=-1` (and test both in float32 and float64)

pymc3/distributions/multivariate.py
kc611: It somehow fails if we set `ndim_supp = 2`. Strangely, the samples returned by `rng_fn` are actually 2-dimensional. Any ideas why this might be happening.

pymc3/tests/test_distributions.py
kc611: I set the precision to 2 decimals for `float32` just to be sure.

ricardoV94: You should actually test it locally. 2 might not be necessary or it might not be enough.

ricardoV94: You can check the notes 3 and 4 in here: https://github.com/pymc-devs/pymc3/blob/fe115b857ae42e84d1f359153fd1f7efb2c327ae/docs/source/distribution_implementation.md#4-adding-tests-for-the-logp--logcdf-methods

pymc3/distributions/multivariate.py
ricardoV94: You might need to override `_shape_from_params` if the RV is failing to guess the shape with the default method. This is needed here for example: https://github.com/aesara-devs/aesara/blob/3a3adaee3b8ceccff6d0208996f117fcd5e5eb6f/aesara/tensor/random/basic.py#L518-L521

What error are you seeing?

kc611: > What error are you seeing?

<details>
<summary>Here's the full traceback</summary>
<br>





```
________________________________________________________ TestWishart.test_distribution ________________________________________________________

self = <pymc3.tests.test_distributions_random.TestWishart object at 0x7f53e0b50810>

    def test_distribution(self):
        self.validate_tests_list()
>       self._instantiate_pymc_rv()

pymc3/tests/test_distributions_random.py:300: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pymc3/tests/test_distributions_random.py:311: in _instantiate_pymc_rv
    **params, size=self.size, rng=aesara.shared(self.get_random_state(reset=True))
pymc3/distributions/multivariate.py:816: in dist
    return super().dist([nu, V], *args, **kwargs)
pymc3/distributions/distribution.py:285: in dist
../../../miniconda3/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/tensor/random/op.py:323: in __call__
    res = super().__call__(rng, size, dtype, *args, **kwargs)
../../../miniconda3/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/graph/op.py:269: in __call__
    node = self.make_node(*inputs, **kwargs)
../../../miniconda3/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/tensor/random/op.py:370: in make_node
    bcast = self.compute_bcast(dist_params, size)
../../../miniconda3/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/configparser.py:49: in res
    return f(*args, **kwargs)
../../../miniconda3/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/tensor/random/op.py:287: in compute_bcast
    shape = self._infer_shape(size, dist_params)
../../../miniconda3/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/tensor/random/op.py:257: in _infer_shape
    param_shapes=param_shapes,
pymc3/distributions/multivariate.py:750: in _shape_from_params
    self.ndim_supp, dist_params, rep_param_idx, param_shapes
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

ndim_supp = 2, dist_params = (TensorConstant{4}, TensorConstant{[[1. 0. 0...0. 0. 1.]]}), rep_param_idx = 2
param_shapes = [(), (Subtensor{int64}.0, Subtensor{int64}.0)]

    def default_shape_from_params(
        ndim_supp, dist_params, rep_param_idx=0, param_shapes=None
    ):
        """Infer the dimensions for the output of a `RandomVariable`.
    
        This is a function that derives a random variable's support
        shape/dimensions from one of its parameters.
    
        XXX: It's not always possible to determine a random variable's support
        shape from its parameters, so this function has fundamentally limited
        applicability and must be replaced by custom logic in such cases.
    
        XXX: This function is not expected to handle `ndim_supp = 0` (i.e.
        scalars), since that is already definitively handled in the `Op` that
        calls this.
    
        TODO: Consider using `aesara.compile.ops.shape_i` alongside `ShapeFeature`.
    
        Parameters
        ----------
        ndim_supp: int
            Total number of dimensions for a single draw of the random variable
            (e.g. a multivariate normal draw is 1D, so `ndim_supp = 1`).
        dist_params: list of `aesara.graph.basic.Variable`
            The distribution parameters.
        param_shapes: list of tuple of `ScalarVariable` (optional)
            Symbolic shapes for each distribution parameter.  These will
            be used in place of distribution parameter-generated shapes.
        rep_param_idx: int (optional)
            The index of the distribution parameter to use as a reference
            In other words, a parameter in `dist_param` with a shape corresponding
            to the support's shape.
            The default is the first parameter (i.e. the value 0).
    
        Results
        -------
        out: a tuple representing the support shape for a distribution with the
        given `dist_params`.
    
        """
        if ndim_supp <= 0:
            raise ValueError("ndim_supp must be greater than 0")
        if param_shapes is not None:
>           ref_param = param_shapes[rep_param_idx]
E           IndexError: list index out of range

../../../miniconda3/envs/pymc3-dev-py37/lib/python3.7/site-packages/aesara/tensor/random/op.py:69: IndexError
```
</details>

And overriding `_shape_from_params` didn't work. You might also wanna take a look at the test first, just to see if the parameters are valid.

ricardoV94: It's definitely looks like you need to override that method. I didn't mean to imply that it should be the same overwrite, just that it needs one.

brandonwillard: > It somehow fails if we set `ndim_supp = 2`. Strangely, the samples returned by `rng_fn` are actually 2-dimensional. Any ideas why this might be happening?

`ndim_supp` is used for shape inference and a few other related things, but not necessarily for critical `RandomVariable.perform` operations, so it's not surprising that this value would be incorrect and still produce valid samples.

Regardless, we need the `ndim_supp` values to be accurate.  We might also need to add some more sanity checks at the symbolic level using that property (i.e. when/if we can verify that the output dimensions match `ndim_supp`, we probably should).

For example, we might be missing a sanity check like `outputs[1].ndim >= self.ndim_supp` in `RandomVariable.make_node`.  It would be better if we could somehow guarantee that condition from the construction of `outputs`, but given how subclasses can override some of the steps involved, such a simple check might be worthwhile.

brandonwillard: Otherwise, @ricardoV94 is correct; it looks like this needs a custom `_shape_from_params` that says the second parameter, `V`, determines the output shape.  I think the default assumes that the first parameter does, which is clearly wrong in this case.

brandonwillard (changes requested):

pymc3/distributions/multivariate.py
brandonwillard: OK, then we don't need to add that functionality here/now.

pymc3/tests/test_distributions.py
kc611: You were right, the passed all the way upto 4 decimal points in `float32` mode. And started showing failures at 5 decimal points. I left it at our default of 3 cause it may show flaky failures at 4 decimal points.

pymc3/distributions/multivariate.py
kc611: The current approach should cover the correct `ndim_supp` value as well as the returned shape correctly. I simply returned the shape of `V` from the `param_shapes` list.

pymc3/tests/test_distributions_random.py
```diff
@@ -1306,23 +1306,29 @@ class TestOrderedProbit(BaseTestDistribution):
...
+        (2, 4, 2, 3, 3),
+    ]
 
     reference_dist = lambda self: functools.partial(
         self.wishart_rng_fn, rng=self.get_random_state()
     )
     tests_to_run = [
         "check_rv_size",
         "check_pymc_params_match_rv_op",
-        "check_pymc_draws_match_reference",
```
ricardoV94: Why did you remove that test?

kc611: The test was failing with `AssertionError` with different numeric values, and this is flaky (see e4d10d0a636761339cea20a3d8eb822acbfd4b67 for a passing instance) and considering https://github.com/pymc-devs/pymc3/blob/7be2025c4091e450f9e9dd050eb0dca08995c83f/pymc3/tests/test_distributions_random.py#L1620 I though it would be better to remove it and just check for shapes. The test for `Wishart` in `test_distributions.py` is also commented out giving the reason https://github.com/pymc-devs/pymc3/blob/7be2025c4091e450f9e9dd050eb0dca08995c83f/pymc3/tests/test_distributions.py#L2043-L2047

(I don't know how applicable this is in the current scenario though)

ricardoV94: This is about the random method, not the `logp`/`logdf` as in the second test you linked to. If the scipy method is not good for testing, it is also not good enough to be used inside the `rng_fn`. Unless I am missing something.

kc611: Alright I re-added it. It showed a flaky failure (only once) on my system, if it arises again, we can investigate into this further.

Maybe it could've been caused by typecasting `nu` as `np.int(nu)` in `Wishart`'s `rng_fn`. I added a similar function in the test.

pymc3/tests/test_distributions_random.py
```diff
@@ -1306,6 +1305,65 @@ class TestOrderedProbit(BaseTestDistribution):
...
+
+    pymc_dist_params = {"mu": mu, "rowcov": row_cov, "colcov": col_cov}
+    expected_rv_op_params = {"mu": mu, "rowcov": row_cov, "colcov": col_cov}
+
+    sizes_to_check = [None, (1), (2, 3)]
+    sizes_expected = [(3,), (1, 3), (2, 3, 3)]
+
+    tests_to_run = ["check_rv_size", "check_pymc_params_match_rv_op", "test_matrix_normal"]
+
+    def test_matrix_normal(self):
```
kc611, line 1406: Should we remove this test / replace this test with `rng_fn` from MatrixNormal ?

ricardoV94: I think the current one is better, as it provides a more independent check that our `rng_fn` is doing the correct thing. The standard test is used when we are just relying on a standard implementation in `rng_fn` (i.e., just calling scipy / numpy, without many steps in it). Is the test passing?

kc611: > Is the test passing?

It's flaky. Currently fails (in the current run of the Github jobs) but passed before.

pymc3/tests/test_distributions_random.py
```diff
@@ -1306,6 +1305,65 @@ class TestOrderedProbit(BaseTestDistribution):
...
+    sizes_to_check = [None, (), 1, (1,), 5, (4, 5), (2, 4, 2)]
+    sizes_expected = [
+        (3, 3),
+        (3, 3),
+        (1, 3, 3),
+        (1, 3, 3),
+        (5, 3, 3),
+        (4, 5, 3, 3),
+        (2, 4, 2, 3, 3),
+    ]
```
brandonwillard: What extra logic (e.g. implemented in `WishartRV`) are all these sizes testing?  

Remember, we don't want the test time to grow larger without proportionally increasing the actual test coverage (and that coverage **doesn't** include NumPy/SciPy's code).  This test module is easily one of the most egregious violators of that principal, so we should always be a little more critical of additions to it.

kc611: Should that be checking only for edge cases like `None` and `()` ? I just took the `sizes_to_check` list from the original function in `BaseTestDistribution` where the sizes are being checked by default. (And constructed the corresponding expected list.)

ricardoV94: To echo @kc611, this is not actually testing more than what the default test does, just setting the specific parameters of the size test, which usually have to be adapted for multivariate distributions.

pymc3/tests/test_distributions_random.py
ricardoV94: You might need to check dimension by dimension. The default test flattens everything and that might not be so comparable. This issue also cropped in here: https://github.com/pymc-devs/pymc3/pull/4596#issuecomment-822589620

(I might also be completely off)

kc611: Looks like that could work, I added it for now.

pymc3/tests/test_distributions_random.py
```diff
@@ -1352,16 +1350,40 @@ class TestMatrixNormal(BaseTestDistribution):
...
+            check = pm.sample_prior_predictive(n_fails)
+
+        ref_smp = ref_rand(size[0], mu=np.random.random((3, 3)), rowcov=np.eye(3), colcov=np.eye(3))
+
+        p, f = delta, n_fails
+        while p <= delta and f > 0:
+            matrixnormal_smp = check["mvnormal"][f - 1, :, :]
+            curr_ref_smp = ref_smp[f - 1, :, :]
+
+            p = min(
```
ricardoV94: Not sure about the `min` here.

pymc3/tests/test_distributions_random.py
```diff
@@ -1352,16 +1350,40 @@ class TestMatrixNormal(BaseTestDistribution):
     tests_to_run = ["check_rv_size", "check_pymc_params_match_rv_op", "test_matrix_normal"]
 
     def test_matrix_normal(self):
+        delta = 0.05  # limit for KS p-value
+        n_fails = 100  # Allows the KS fails a certain number of times
```
ricardoV94: 100 `n_fails` also seems extreme.

kc611: I set it to the default of 10.

pymc3/tests/test_distributions_random.py
kc611: We can take an mean maybe ?

ricardoV94: Nevermind, `min` is the strongest test, I thought about it the wrong way. nfails can be a bit bigger than 10 but 100 seemed too much.

pymc3/distributions/multivariate.py
```diff
@@ -737,6 +738,25 @@ def __str__(self):
...
+class WishartRV(RandomVariable):
+    name = "wishart"
+    ndim_supp = 2
+    ndims_params = [0, 2]
+    dtype = "floatX"
+    _print_name = ("Wishart", "\\operatorname{Wishart}")
+
+    def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
+        # The shape of second parameter `V` defines the shape of the output.
+        return param_shapes[1]
```
ricardoV94: Does this make sense like this @brandonwillard, I am not very familiar with the signature.

ricardoV94 (commented):
LGTM otherwise. Is anything else missing?

pymc3/distributions/multivariate.py
kc611: We can also use `dist_params[1].shape`. Just needs to be same as shape of second argument `V`.

pymc3/tests/test_distributions_random.py
```diff
@@ -1306,6 +1303,91 @@ class TestOrderedProbit(BaseTestDistribution):
...
+            check = pm.sample_prior_predictive(n_fails)
+
+        ref_smp = ref_rand(size[0], mu=np.random.random((3, 3)), rowcov=np.eye(3), colcov=np.eye(3))
+
+        p, f = delta, n_fails
+        while p <= delta and f > 0:
+            matrixnormal_smp = check["mvnormal"][f - 1, :, :]
+            curr_ref_smp = ref_smp[f - 1, :, :]
+
+            p = np.mean(
```
ricardoV94: Sorry for the earlier confusion, but I think the minimum is the right metric to use. If the 10 attempts are not enough you can increase those a bit, but we should still use the min.

pymc3/tests/test_distributions.py
```diff
@@ -2040,7 +2045,6 @@ def test_AR1(self, n):
         self.check_logp(AR1, Vector(R, n), {"k": Unit, "tau_e": Rplus}, AR1_logpdf)
 
     @pytest.mark.parametrize("n", [2, 3])
-    @pytest.mark.xfail(reason="Distribution not refactored yet")
     def test_wishart(self, n):
```
michaelosthege, line 2048: Over in #4824 I noticed that this test XPASS, but removing the XFAIL also doesn't make much sense, since the test has no active content.
Can you bring this test back online?

ricardoV94: That tests requires the `checkd` which has not been refactored, and it seems the original test was no good either. I think the best is to just remove it.

kc611: Updated it to the standard `check_logp` test.

pymc3/distributions/multivariate.py
```diff
@@ -775,9 +796,10 @@ class Wishart(Continuous):
     This distribution is unusable in a PyMC3 model. You should instead
     use LKJCholeskyCov or LKJCorr.
     """
+    rv_op = wishart
 
-    def __init__(self, nu, V, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    @classmethod
+    def dist(cls, nu, V, *args, **kwargs):
         warnings.warn(
```
michaelosthege, line 806: Looks like @twiecki fixed the distribution in https://github.com/pymc-devs/pymc3/pull/701 ?
Does that mean we can remove the warning, or at least "downgrade" the warning into a recommendation to use LKJ dists instead?

**Disclaimer:** I have no clue what either distributions are 😅

pymc3/distributions/multivariate.py
```diff
@@ -1533,175 +1565,101 @@ class MatrixNormal(Continuous):
...
-        self.solve_lower = solve_lower_triangular
-        self.solve_upper = solve_upper_triangular
-
-    def _setup_matrices(self, colcov, colchol, coltau, rowcov, rowchol, rowtau):
+
         cholesky = Cholesky(lower=True, on_error="raise")
 
+        if mu.ndim == 1:
+            raise ValueError(
+                "1x1 Matrix was provided, " " Please use Normal distribution " "for such cases."
```
michaelosthege:
```suggestion
                "1x1 Matrix was provided. Please use Normal distribution for such cases."
```

pymc3/distributions/multivariate.py
```diff
@@ -1533,175 +1565,101 @@ class MatrixNormal(Continuous):
...
+            )
+
         # Among-row matrices
-        if len([i for i in [rowtau, rowcov, rowchol] if i is not None]) != 1:
+        if len([i for i in [rowcov, rowchol] if i is not None]) != 1:
             raise ValueError(
-                "Incompatible parameterization. "
-                "Specify exactly one of rowtau, rowcov, "
-                "or rowchol."
+                "Incompatible parameterization. " "Specify exactly one of rowcov, " "or rowchol."
```
michaelosthege:
```suggestion
                "Incompatible parameterization. Specify exactly one of rowcov, or rowchol."
```

pymc3/distributions/multivariate.py
```diff
@@ -1533,175 +1565,101 @@ class MatrixNormal(Continuous):
...
+            rowchol_cov = at.as_tensor_variable(rowchol)
 
         # Among-column matrices
-        if len([i for i in [coltau, colcov, colchol] if i is not None]) != 1:
+        if len([i for i in [colcov, colchol] if i is not None]) != 1:
             raise ValueError(
-                "Incompatible parameterization. "
-                "Specify exactly one of coltau, colcov, "
-                "or colchol."
+                "Incompatible parameterization. " "Specify exactly one of colcov, " "or colchol."
```
michaelosthege:
```suggestion
                "Incompatible parameterization. Specify exactly one of colcov, or colchol."
```

pymc3/tests/test_distributions_random.py
michaelosthege: I agree to both: This implementations is very much in line with the defaults.
The the `BaseTestDistribution` for sure takes a sledgehammer to break a nut, but that problem is out of scope of this PR.

michaelosthege (commented):
@kc611 take a look at my comments and tell us when you see this as ready to merge :)

pymc3/tests/test_distributions_random.py
brandonwillard: Pushing admittedly poor defaults forward is a great way to get into a bad situation, and an even better way to get deeper into existing ones.  

Also, the answer is not one big PR that removes all the testing redundancies; given the knowledge and effort required for that, it's quite unrealistic.  Plus, that won't prevent the problem from reappearing again and again after such a heroic PR.

At the very least, we need to stop adding to the testing redundancies on a *case-by-case basis*, and we can do that by looking for it during reviews; otherwise, we're just hoping that contributors will do that for us, and that's not how things have turned out.


Regarding these exact tests, if they're only targeting `WishartRV.rng_fn`, it looks like we only need a test for an empty and non-empty `size` (and perhaps one other), because that's the only relevant condition in the code added by this PR.  Every other value besides the first two appears to be testing the code in `RandomVariable` and `scipy.stats.wishart.rvs`.

kc611: I think there's no harm in not testing the `[ (), (1,), 5,  (2, 4, 2)]` cases if we are testing for `[None, 1, (4, 5)]` cases since they're pretty much the same thing with different values. (Updated accordingly)

pymc3/distributions/multivariate.py
kc611: I think #701 adds the `WishartBartlett` as a solution for that, I updated the warning to recommend use of  `WishartBartlett` instead of `LKJCorr` or `LKJCholeskyCov`.

cc @twiecki Is this correct ?

twiecki: No, LKJCorr should be used. `Wishart` itself is basically unusable, so if a user really wants that distribution, they can only use `WishartBartlett`, but they should always use either `LKJCorr` or `LKJCholeskyCov`.

ricardoV94: Let's revert the warning to how it was before? Other than this PR looks pretty ready.

Also, should this warn be in `__new__`? Does it appear twice during a normal initialization inside a `pm.Model()`?

ricardoV94 (approved):
LGTM, let's wait for the tests. Great work, these two were a tough group to crack

pymc3/distributions/multivariate.py
michaelosthege: I don't think it would appear twice. And implementing a `__new__` just because of this warning seems a little too much.
Let's just rephrase to mention the entire recommendation from above and then go for it.

michaelosthege: Or like this. (I changed as I was writing the above.)

michaelosthege (approved):

PR #4779, opened 2021-06-17
Create dedicated environment YML for Windows

michaelosthege:
The install installation guide for Windows was recently updated after it was discovered that the combination of numba+scipy dependencies leads to the correct installation of BLAS dependencies.

This PR creates a conda `environment.yml` specifically for Windows. That's because the `m2w64-toolchain` dependency is Windows-only and conda does not yet support flags to make dependencies OS-specific.

The GitHub Action for Windows was changed to use the new environment file.
Maybe this fixes the CI issues for the `v3` branch.

### DIFF
--- a/conda-envs/windows-environment-dev-py38.yml
+++ b/conda-envs/windows-environment-dev-py38.yml
@@ -1,0 +1,26 @@
+name: pymc3-dev-py38
+channels:
+- conda-forge
+- defaults
+dependencies:
+ # base dependencies (see install guide for Windows)
+- h5py>=2.7
+- libpython
+- mkl-service
+- m2w64-toolchain
+- numba
+- pip
+- python=3.8
+- python-graphviz
+- scipy
+# Extra stuff for dev, testing and docs build
+- ipython>=7.16
+- nbsphinx>=0.4
+- numpydoc>=0.9
+- pre-commit>=2.8.0
+- pytest-cov>=2.5
+- pytest>=3.0
+- recommonmark>=0.4
+- sphinx-autobuild>=0.7
+- sphinx>=1.5
+- watermark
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -1,7 +1,6 @@
 # This file is auto-generated by scripts/generate_pip_deps_from_conda.py, do not modify.
 # See that file for comments about the need/usage of each dependency.
 
-h5py>=2.7
```
MarcoGorelli: why remove this?

MarcoGorelli: 🤔 the check might need updating here, as h5py is in the other conda environments but not in the windows one

---

or just remove requirements-dev.txt and require pymc devs to use conda for development, that might be simpler 😄

michaelosthege: I though this was already a dependency in ArviZ, but it looks like that's not the case.
I will undo that change,

rpgoldman:
I can't really review this, @twiecki because I don't have access to a Windows box.  And I like it that way! ;-)

michaelosthege:
Thanks @MarcoGorelli I put the h5py dependency back in.

@rpgoldman I had hoped that this PR kinda reviews itself by fixing the CI pipeline that is broken on all other current v3 PRs.

If we can merge it asap I'll rebase the other PRs to unblock the road to 3.11.3 @twiecki

conda-envs/windows-environment-dev-py38.yml
```diff
@@ -0,0 +1,26 @@
+name: pymc3-dev-py38
```
canyon289, line 1: Same comment as the other PR. Just add a docstring explaining the nuances of "what does windows do different"

canyon289 (approved):

conda-envs/windows-environment-dev-py38.yml
michaelosthege: Deferred to #4786

Issue #4517, opened 2021-03-09
Combine Windows and Pytest workflows using YAML anchors


michaelosthege:
There is considerable duplication of YAML code between the `pytest.yml` and `windows.yml` workflows.

With #4512 the splitting of tests across jobs is cleaned up. Exactly this new matrix definition should be refactored into a YAML anchor such that the Windows job from `windows.yml` can be migrated into `pytest.yml` and also run all the tests.

I would add the "beginner friendly" label, but I find YAML so unbelievably hard to work with that it wouldn't be fair.

PR #4785, opened 2021-06-19
Create dedicated environment YML for Windows & reenable tests

michaelosthege:
This is based on a cherry-pick of the commit from #4779 which fixes the Windows step in the CI by using a more robust conda environment file.

I also re-enabled the Windows job in the CI, but did not yet add more test files.

This is the first step towards #4517


scripts/generate_pip_deps_from_conda.py
```diff
@@ -117,7 +127,7 @@ def main(conda_fname, pip_fname):
         f"# This file is auto-generated by scripts/generate_pip_deps_from_conda.py, "
         "do not modify.\n# See that file for comments about the need/usage of each dependency.\n\n"
     )
-    pip_content = header + "\n".join(pip_deps) + "\n"
+    pip_content = header + "\n".join(sorted(pip_deps)) + "\n"
```
canyon289, line 130: Why does it have to be sorted?

conda-envs/windows-environment-dev-py38.yml
```diff
@@ -0,0 +1,37 @@
+name: pymc3-dev-py38
```
canyon289, line 1: Is there any particular reason the window dev environment needs to be different? If so add a comment up top so whoever reads this next is aware what those are. 

I havet used windows in many years now so even I dont know, which is where the comment will help

canyon289 (approved):

scripts/generate_pip_deps_from_conda.py
michaelosthege: In the Windows env file I split the dependencies by absolutely necessary vs. dev/docs related. The function here then took them in the same order, but `requirements-dev.txt` is sorted by another step in the precommit..

conda-envs/windows-environment-dev-py38.yml
michaelosthege: The `m2w64-toolchain` dependency is Windows-only, but conda doesn't support OS-conditional dependencies in these files..

I figured it was easier to maintain & diagnose two separate files than to find the set of dependencies that magically makes the conda dependency resolver find the correct result on all platforms.

canyon289: I would add this explanation to the file

michaelosthege: Deferred to #4786

PR #4787, opened 2021-06-19
Run CI jobs on v3 branch too

michaelosthege:
The workflows ran on PRs, but not on the `v3` after merging..
I'm not sure how GitHub handles this. Maybe this is a change we have to make on the `main` branch?


.github/workflows/arviz_compat.yml
```diff
@@ -3,7 +3,7 @@ name: arviz-compatibility
 on:
   pull_request:
   push:
-    branches: [master]
+    branches: [v3]
```
MarcoGorelli, line 6: why not both? see https://github.com/pandas-dev/pandas/blob/master/.github/workflows/database.yml for reference

michaelosthege: From what it did so far, my impression was that the workflow definition is specific to the branch where it's defined.
This PR is for `v3` which won't go back into `main` anyway.

But my understanding of these things is limited, so please elaborate if I'm wrong.

MarcoGorelli: ah OK, yeah I missed that - looks good to me then

PR #4783, opened 2021-06-18
Add V4 distribution implementation developer guide

ricardoV94:
This PR adds a developer guide on how to implement (and test) new distributions for PyMC3 `v4`. 

Somethings are still marked as `TODO` and feedback is much appreciated. 

The way distributions are implemented (and tested) will probably evolve over time, and hopefully this document will be easy to update as those things happen.

Issue #4781, opened 2021-06-18
Mixtures not refactored for V4


larryshamalama:
The following snipet and error uses `v4`:

``` python
import numpy as np
import pandas as pd
import pymc3 as pm

from aesara import tensor as at

old_faithful_df = pd.read_csv(pm.get_data("old_faithful.csv"))
waiting_times = old_faithful_df["waiting"]
waiting_times = ((waiting_times - waiting_times.mean())/waiting_times.std()).values

def stick_breaking(betas):
    '''
    betas is a K-vector of iid draws from a Beta distribution
    '''
    sticks = tt.concatenate(
        [
            [1],
            (1 - betas[:-1])
        ]
    )
    
    return tt.mul(betas, tt.cumprod(sticks))

with pm.Model() as model:
    alpha = pm.Gamma(name="alpha", alpha=1, beta=1)
    v = pm.Beta(name="v", alpha=1, beta=alpha, shape=(K,)) # beta=alpha kinda confusing here
    
    w = pm.Deterministic(name="w", var=stick_breaking(v))
    mu = pm.Normal(name="mu", mu=0, sigma=5)
    sigma = pm.InverseGamma(name="sigma", alpha=1, beta=1, shape=(K,))
    obs = pm.NormalMixture(name="theta", w=w, mu=mu, tau=1/sigma, observed=waiting_times)
```

yields

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-9959a48ef76f> in <module>
      6     mu = pm.Normal(name="mu", mu=0, sigma=5)
      7     sigma = pm.InverseGamma(name="sigma", alpha=1, beta=1, shape=(K,))
----> 8     obs = pm.NormalMixture(name="theta", w=w, mu=mu, tau=1/sigma, observed=waiting_times)

~/anaconda3/envs/pymc3-dev-py39/lib/python3.9/site-packages/pymc3/distributions/distribution.py in __new__(cls, name, rng, dims, initval, observed, total_size, transform, *args, **kwargs)
    205         # Create the RV without specifying initval, because the initval may have a shape
    206         # that only matches after replicating with a size implied by dims (see below).
--> 207         rv_out = cls.dist(*args, rng=rng, initval=None, **kwargs)
    208         ndim_actual = rv_out.ndim
    209         resize_shape = None

TypeError: dist() missing 1 required positional argument: 'dist_params'

---------------------------------------------------------------------------
[...]

PR #4795, opened 2021-06-21
Refactoring mixture distributions

larryshamalama:
Addressing issue [4781](https://github.com/pymc-devs/pymc3/issues/4781) that seeks to refactor Mixture distributions.

Here is my ongoing WIP. It may not be great, and I would need some pointers. I am also having trouble managing development on different branches, so irrelevant commits may appear here too.

### Questions

- Do I need to create a MixtureRV op?
- My understanding of Distribution classes is still hazy. How do I set `comp_dists` as an instance attribute without `__init__`?

Guidelines on Distribution refactoring available [here](https://github.com/pymc-devs/pymc3/pull/4783/files) for any readers of this thread.


larryshamalama:
Back to the drawing board

PR #4797, opened 2021-06-21
add glossary example as orphan page

OriolAbril:
cc @martinacantaro @ricardoV94

  * Creates a glossary page so that we can already start adding terms and definitions there.
    Choosing where within the docs it will be will come later. If we separate building
    of examples and docs, I think we should still keep a single glossary in the main docs, we can
    then link to term from the notebooks too using intersphinx. The main/only drawback is that
    notebooks that need to add new terms to the glossary need an auxiliary PR to pymc3 repo.
  * We could also do that the other way around, glossary in pymc-examples, link from pymc3 docs with
    intersphinx.
  * Another somewhat too ambitious alternative would be to host a Bayesian glossary (single web, non
    versioned) so that we can then link there from pymc3, arviz, bambi docs... I think @canyon289
    has wanted something like that for a long time

### DIFF
--- a/docs/source/developer_guide_implementing_distribution.md
+++ b/docs/source/developer_guide_implementing_distribution.md
@@ -4,7 +4,7 @@ This guide provides an overview on how to implement a distribution for version 4
 It is designed for developers who wish to add a new distribution to the library.
 Users will not be aware of all this complexity and should instead make use of helper methods such as (TODO).
 
-PyMC3 {`class`}`~pymc3.distributions.Distribution` build on top of Aesara's {`class`}`~aesara.tensor.random.op.RandomVariable`, and implement `logp` and `logcdf` methods as well as other initialization and validation helpers, most notably `shape/dims`, alternative parametrizations, and default `transforms`.
+PyMC3 {class}`~pymc3.distributions.Distribution` build on top of Aesara's {class}`~aesara.tensor.random.op.RandomVariable`, and implement `logp` and `logcdf` methods as well as other initialization and validation helpers, most notably `shape/dims`, alternative parametrizations, and default `transforms`.
 
 Here is a summary check-list of the steps needed to implement a new distribution.
 Each section will be expanded below:
@@ -19,7 +19,7 @@ This guide does not attempt to explain the rationale behind the `Distributions` 
 
 ## 1. Creating a new `RandomVariable` `Op`
 
-{`class`}`~aesara.tensor.random.op.RandomVariable` are responsible for implementing the random sampling methods, which in version 3 of PyMC3 used to be one of the standard `Distribution` methods, alongside `logp` and `logcdf`.
+{class}`~aesara.tensor.random.op.RandomVariable` are responsible for implementing the random sampling methods, which in version 3 of PyMC3 used to be one of the standard `Distribution` methods, alongside `logp` and `logcdf`.
 The `RandomVariable` is also responsible for parameter broadcasting and shape inference.
 
 Before creating a new `RandomVariable` make sure that it is not offered in the [Numpy library](https://numpy.org/doc/stable/reference/random/generator.html#distributions).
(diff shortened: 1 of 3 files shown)


ricardoV94:
Can't we keep two glossaries, one for pymc-examples and the other for pymc3 docs? 

Edit: I see, we only need a PR to pymc3 if a new glossary item is needed? That sounds reasonable

OriolAbril:
Yes, I think two glossaries would be confusing and end up being more work. notebooks will be able to link to any term in the glossary without any problem, and users clicking will go to the main pymc3 docs. Side note, in practice I am not sure this will have much of an effect though, we could for example keep the exact same navbar in both docs, so that visually everything looks basically the same, in pymc3 docs, the examples and tutorials would be external links, in pymc-examples the other way around

What is not possible (from any of the two) is linking to terms that are not yet defined, so if new notebooks need to add terms to the glossary, they need an extra PR to pymc3 repo (or viceversa, maybe the main docs are more stable and it makes more sense to host the glossary in pymc-examples which is where more new terms will be added :thinking: ) to add the term so that then it can be referenced in the notebook

ricardoV94:
Yes I think the pymc3 docs is a better place for the glossary. 

Edit: I see you actually said the opposite haha :D I will refrain from misreading you again in this PR

OriolAbril:
> Yes I think the pymc3 docs is a better place for the glossary.
>
> Edit: I see you actually said the opposite haha :D I will refrain from misreading you again in this PR

I actually don't know where it would be better!

ricardoV94:
Do you want me to push here the terms that will be used in #4783?

michaelosthege:
@OriolAbril what shall we do with this PR ?

Sayam753:
> Another somewhat too ambitious alternative would be to host a Bayesian glossary (single web, non
versioned) so that we can then link there from pymc3, arviz, bambi docs... I think @canyon289
has wanted something like that for a long time

Creating a dedicated repository in PyMC just for glossary seems like a good option to me. 
And then we can decide either adding it as a submodule or just linking it from the documentation.

OriolAbril:
I think we can start trying out the glossary here (thus merge this) and if we see that having a glossary as an independent webpage or having it rendered via pymc-examples works better then move the page. 

It isn't yet ready to be merged though. I can start fixing some syntax from the distributions dev guide which is so far the only page that has terms so that someone else can then add the definitions. 

@Sayam753 do we already have PR doc preview for v4?

Sayam753:
> @Sayam753 do we already have PR doc preview for v4?

Yes, preview - https://pymc3-fork.readthedocs.io/en/latest/

But it is on my fork. I'll roll out a PR later this week, adding support of RTD on main. I guess it will be just a yml file.

docs/source/developer_guide_implementing_distribution.md
```diff
@@ -84,12 +84,12 @@ blah = BlahRV()
 
 Some important things to keep in mind:
 
-1. Everything inside the `rng_fn` method is pure Python code (as are the inputs) and should not make use of other `Aesara` symbolic ops. The random method should make use of the `rng` which is a Numpy    {`class`}`~numpy.random.RandomState`, so that samples are reproducible.
+1. Everything inside the `rng_fn` method is pure Python code (as are the inputs) and should not make use of other `Aesara` symbolic ops. The random method should make use of the `rng` which is a Numpy    {class}`~numpy.random.RandomState`, so that samples are reproducible.
```
OriolAbril, line 88: just to double check, v4 uses RandomState or Generator?

docs/source/developer_guide_implementing_distribution.md
```diff
@@ -280,7 +280,7 @@ The latter kind of test (if warranted) can be performed with the aid of `pymc3_r
 This kind of test only makes sense if there is a good independent generator reference (i.e., not just the same composition of numpy / scipy python calls that is done inside `rng_fn`).
 
 Finally, when your `rng_fn` is doing something more than just calling a `numpy` or `scipy` method, you will need to setup an equivalent seeded function with which to compare for the exact draws (instead of relying on `seeded_[scipy|numpy]_distribution_builder`).
-You can find an example in {`class`}`~pymc3.tests.test_distributions_random.TestWeibull`, whose `rng_fn` returns `beta * np.random.weibull(alpha, size=size)`.
+You can find an example in {class}`~pymc3.tests.test_distributions_random.TestWeibull`, whose `rng_fn` returns `beta * np.random.weibull(alpha, size=size)`.
```
OriolAbril, line 284: now the syntax for these links is correct, but tests are not part of the documentation, so they don't render as links as their destination doesn't exist.

docs/source/glossary.md
```diff
@@ -0,0 +1,19 @@
...
+  to an external reference.
+
+Second term
+  Definitions can have as many lines as desired, and should be written in markdown. Definitions
+  can contain any markdown formatting for MyST to parse, this includes basic markdown like **bold**
+  as well as MyST directives and roles like {fa}`fort awesome,style=fab`
+Functional Programming
+  Functional programming is writing pure functions
+Dispatching
+  Choosing which implementation of an operation is used
```
OriolAbril: @ricardoV94 or anyone else who feels like it, please add the right definitions, relevant links... I have left the two example terms and defintions I added originally to show a bit what is possible.

docs/source/developer_guide_implementing_distribution.md
ricardoV94: Both can be used. By default RVs use / create RandomStates

docs/source/glossary.md
```diff
@@ -0,0 +1,19 @@
...
+  Terms are defined within this glossary directive. The term id is defined as the non
+  indented line, and can be text alone (like {term}`second term`) or also include a link
+  to an external reference.
+
+Second term
+  Definitions can have as many lines as desired, and should be written in markdown. Definitions
+  can contain any markdown formatting for MyST to parse, this includes basic markdown like **bold**
+  as well as MyST directives and roles like {fa}`fort awesome,style=fab`
+Functional Programming
+  Functional programming is writing pure functions
```
ricardoV94:
```suggestion
  Functional programming is a programming style that prefers the use of basic functions with explicit and distinct inputs and outputs.
This contrasts with functions or methods that depend on variables that are not explicitly passed as an input (such as accessing `self.variable` inside a method) or that alter the inputs or other state variables in-place, instead of returning new distinct variables as outputs.
```

docs/source/glossary.md
ricardoV94:
```suggestion
  Choosing which function or method implementation to use based on the type of the input variables (usually just the first variable). For some examples, see Python's documentation for the [singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) decorator.
```

michaelosthege (approved):

PR #4802, opened 2021-06-24
Refactor pm.Simulator (WIP)

ricardoV94:
This is a proposal on how to `pm.Simulator` could be refactored into `V4`. It provides a helper method to create a SimulatorRV which tries to behave as a typical RV for prior and posterior predictive sampling, and whose logp is `distance(epsilon, sum_stat(value), sim_stat(sim_rv))` where `sim_rv` is just another reinstantiation of the original `SimulatorRV`. 

The advantages is that we don't need special logic for the `ABC` kernel in `sample_smc` as it works just like a normal variable. It can also be used in some conventional samplers (at least it seems to work with Metropolis MCMC). It does away with the limitation of having a single `pm.Simulator` and so on.

Two other options would be to:

2. Separate the `Simulator` object from a `pm.Distance` likelihood term which may be more transparent / easier to think for the user. The mixing of the two doesn't make a lot of sense if we think of `Simulator` as a true RV (for example if we use the mean summary statistic, then the logp shape is completely different than the samples shape). It may also make it easier to manipulate parameters like the `epsilon` by the sampler if we separate the two. Perhaps distance should even be completely the responsibility of the sampler kernel.

3. Keep most of the unique logic from `v3`, which keeps (for better or worse) `pm.Simulator` and `ABC` on their very specific corner in the library.

CC: @aloctavodia @junpenglao

### DIFF
--- a/pymc3/distributions/simulator.py
+++ b/pymc3/distributions/simulator.py
@@ -14,12 +14,15 @@
 
 import logging
 
+import aesara
 import aesara.tensor as at
 import numpy as np
 
+from aesara.graph.op import Op
 from aesara.tensor.random.op import RandomVariable
 from scipy.spatial import cKDTree
 
+from pymc3.aesaraf import floatX
 from pymc3.distributions.distribution import NoDistribution
 from pymc3.distributions.logp import _logp
 
@@ -84,36 +87,38 @@ class Simulator(NoDistribution):
         **kwargs,
     ):
 
-        if distance == "gaussian":
-            distance = gaussian
-        elif distance == "laplace":
-            distance = laplace
-        elif distance == "kullback_leibler":
-            raise NotImplementedError("KL not refactored yet")
-            # TODO: Wrap KL in aesara OP
-            # distance = KullbackLiebler(observed)
-            # if sum_stat != "identity":
-            #     _log.info(f"Automatically setting sum_stat to identity as expected by {distance}")
-            #     sum_stat = "identity"
-        elif hasattr(distance, "__call__"):
-            # TODO: Wrap (optionally) non symbolic distance in Aesara OP
-            distance = distance
-        else:
-            raise ValueError(f"The distance metric {distance} is not implemented")
+        if not isinstance(distance, Op):
+            if distance == "gaussian":
+                distance = gaussian
+            elif distance == "laplace":
+                distance = laplace
+            elif distance == "kullback_leibler":
+                raise NotImplementedError("KL not refactored yet")
+                # TODO: Wrap KL in aesara OP
+                # distance = KullbackLiebler(observed)
+                # if sum_stat != "identity":
+                #     _log.info(f"Automatically setting sum_stat to identity as expected by {distance}")
+                #     sum_stat = "identity"
+            elif callable(distance):
+                distance = create_distance_op_from_fn(distance)
+            else:
+                raise ValueError(f"The distance metric {distance} is not implemented")
 
-        if sum_stat == "identity":
-            sum_stat = identity
-        elif sum_stat == "sort":
-            sum_stat = at.sort
-        elif sum_stat == "mean":
-            sum_stat = at.mean
-        elif sum_stat == "median":
-            sum_stat = at.median
-        elif hasattr(sum_stat, "__call__"):
-            # TODO: Wrap (optionally) non symbolic sum_stat in Aesara OP
-            sum_stat = sum_stat
-        else:
-            raise ValueError(f"The summary statistics {sum_stat} is not implemented")
+        if not isinstance(sum_stat, Op):
+            if sum_stat == "identity":
+                sum_stat = identity
+            elif sum_stat == "sort":
+                sum_stat = at.sort
+            elif sum_stat == "mean":
+                sum_stat = at.mean
+            elif sum_stat == "median":
+                sum_stat = at.median
+            elif callable(sum_stat):
+                sum_stat = create_sum_stat_op_from_fn(sum_stat)
+            else:
+                raise ValueError(f"The summary statistics {sum_stat} is not implemented")
+
+        epsilon = at.as_tensor_variable(floatX(epsilon))
 
         if params is None:
             params = []
@@ -162,7 +167,6 @@ class Simulator(NoDistribution):
         sim_op = sim_rv.owner.op
         sim_data = at.as_tensor_variable(sim_op.make_node(*sim_rv.owner.inputs))
         sim_data.name = "sim_data"
-
         return distance(epsilon, sum_stat(value), sum_stat(sim_data))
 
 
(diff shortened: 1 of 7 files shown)


pymc3/distributions/simulator.py
```diff
@@ -54,86 +68,102 @@ class Simulator(NoDistribution):
...
+            distance = gaussian
         elif distance == "laplace":
-            self.distance = laplace
+            distance = laplace
         elif distance == "kullback_leibler":
-            self.distance = KullbackLiebler(observed)
-            if sum_stat != "identity":
-                _log.info(f"Automatically setting sum_stat to identity as expected by {distance}")
-                sum_stat = "identity"
+            raise NotImplementedError("KL not refactored yet")
```
michaelosthege: Would be great to link a Github issue right here.

pymc3/distributions/simulator.py
```diff
@@ -54,86 +68,102 @@ class Simulator(NoDistribution):
...
+                # sum_stat=sum_stat,
+                # epsilon=epsilon,
+            ),
+        )()
+
+        # Register custom logp
+        rv_type = type(sim_op)
+
+        @_logp.register(rv_type)
+        def logp(op, sim_rv, rvs_to_values, *sim_params, **kwargs):
```
michaelosthege, line 169: locally defined functions often cause problems with pickling.

Also it looks like this `logp` uses variables from the `__new__` scope. Doesn't this, in combination with the register lead to problems when having more than one `Simulator`?

pymc3/smc/smc.py
```diff
@@ -114,36 +113,36 @@ def setup_kernel(self):
...
+        #         self.model,
+        #         self.var_info,
+        #         self.variables,
+        #         distance,
+        #         sum_stat,
+        #         self.draws,
+        #         self.save_sim_data,
+        #         self.save_log_pseudolikelihood,
+        #     )
+        # elif self.kernel == "metropolis":
```
michaelosthege: ToDo: remove commented code

pymc3/tests/test_smc.py
```diff
@@ -128,7 +130,8 @@ def quantiles(x):
             return np.quantile(x, [0.25, 0.5, 0.75])
 
         def abs_diff(eps, obs_data, sim_data):
-            return np.mean(np.abs((obs_data - sim_data) / eps))
+            at.mean(at.abs_((obs_data - sim_data) / eps))
+            # return np.mean(np.abs((obs_data - sim_data) / eps))
```
michaelosthege: ToDo: looks like an unintentional change to not return the result?

pymc3/tests/test_smc.py
```diff
@@ -144,7 +147,7 @@ def abs_diff(eps, obs_data, sim_data):
             )
 
         with pm.Model() as self.SMABC_potential:
-            a = pm.Normal("a", mu=0, sigma=1)
+            a = pm.Normal("a", mu=0, sigma=1, initval=0.5)
```
michaelosthege, line 225: Why was the initval added? Maybe this indicates an underlying problem?

pymc3/tests/test_smc.py
```diff
@@ -153,36 +156,36 @@ def abs_diff(eps, obs_data, sim_data):
 
     def test_one_gaussian(self):
         with self.SMABC_test:
-            trace = pm.sample_smc(draws=1000, kernel="ABC")
+            trace = pm.sample_smc(draws=1000)
```
michaelosthege: Oh, does `sample_smc` still return a `MultiTrace`? @OriolAbril

ricardoV94:
@michaelosthege thanks for the pointers. This was not meant for implementation review yet, just proof of concept but some of the points you raise are already useful.

pymc3/distributions/simulator.py
ricardoV94: I am not sure about pickling. Depends on the point at which the function is used (after obtaining the logp graph, the function is not needed anymore). 

The registration part should be fine because I am creating a new `type` which has a unique name. I also have a test for multiple Simulators with different methods and it seems to be fine.

pymc3/tests/test_smc.py
ricardoV94: Yes it still does. I am not sure how SMC should be mapped to inferencedata. It doesn't follow the traditional chains x samples. Maybe @aloctavodia  has some ideas.

pymc3/distributions/simulator.py
ricardoV94: I can also try to add a layer of indirection when creating the simulator so that I can attach what are now the local variables to the `tag`, in which case a single logp function / dispatcher would be enough.

pymc3/tests/test_smc.py
aloctavodia: At least for the moment we should treat SMC output as we do with MCMC. So we should run more than one instance of the algorithm as a different "chain" (even if the samples are not markov chains). If we do this many diagnostics still work (at least qualitatively the same as with MCMC, @ricardoV94 I think I share with you some preliminary results about this.). Currently there is some information stored in "report" like the like the log_likelihood and the sequence of betas that we could store in  the `sample_stats` group,

pymc3/distributions/simulator.py
```diff
@@ -54,86 +68,102 @@ class Simulator(NoDistribution):
...
+            sum_stat = at.sort
         elif sum_stat == "mean":
-            self.sum_stat = np.mean
+            sum_stat = at.mean
         elif sum_stat == "median":
-            self.sum_stat = np.median
+            sum_stat = at.median
         elif hasattr(sum_stat, "__call__"):
-            self.sum_stat = sum_stat
+            # TODO: Wrap (optionally) non symbolic sum_stat in Aesara OP
```
aloctavodia: Having this is very important, so we do not force the practitioners to use Aesara

aloctavodia (commented):
I really like where this is going!

I am not sure about  separating the `Simulator` from the  `Distance`, they both together define a pseudolikelihood. So it seems reasonable to have them as parts of  the same object. For the same reason making the distance part of the sampling method is weird. Actually. I think the first version of SMC-ABC was like that. We can of course discuss the benefits of trying to automatically tune epsilon, nut I think having to define a single value of epsilon is actually a good feature in practice as allows to control the level of accuracy and the cost of the simulation with minimal hand tuning.

ricardoV94:
@michaelosthege was totally right about the pickling issues. I didn't spot them in the first iteration because the default in `sample_smc` was `parallel=False`. 

~~In contrast to `SMC`, the vanilla `Metropolis` stepper works fine in multiprocessing because the logp graph is compiled before the forking step, whereas `SMC` tries to pickle the entire model to be sent to each process.~~ only on fork multiprocess

I temporarily set the ABC tests to all be singleprocess.

pymc3/smc/sample_smc.py
```diff
@@ -156,6 +146,29 @@ def sample_smc(
         %282007%29133:7%28816%29>`__
     """
 
+    if kernel is not None:
```
aloctavodia, line 149: How are we going to select if the kernel is metropolis, independent Metropolis-Hasting, HMC, etc?

michaelosthege:
> @michaelosthege was totally right about the pickling issues. I didn't spot them in the first iteration because the default in `sample_smc` was `parallel=False`.
> 
> ~In contrast to `SMC`, the vanilla `Metropolis` stepper works fine in multiprocessing because the logp graph is compiled before the forking step, whereas `SMC` tries to pickle the entire model to be sent to each process.~ only on fork multiprocess
> 
> I temporarily set the ABC tests to all be singleprocess.

Might be a good idea to add include the `test_smc.py` in the Windows tests then.

pymc3/smc/sample_smc.py
ricardoV94: Haven't thought about that yet. We might use the kernel argument for that, in which case I'll remove the DeprecationWarning. This is just temporarily.

ricardoV94:
The pickling issue is proving non-trivial to overcome. 

It is caused by the dynamic RV class creation (as well as user `Op`s created at runtime with the new helpers).
The `logp` dispatching does not seem to be an issue, but in any case I standardized that in my last commitl.

I tried another approach that would require users to explicit subclass from the SimulatorRV instead of creating one dynamically, and these can be pickled / used in multiprocess. However as soon as the classes are defined inside a function (such as the `setup_class` in `TestSMCABC` it fails to pickle again). This method also requires more effort from the users.

The old `pm.DensityDist` faced similar issues and required some workarounds to deal with the `logp` method. I am not sure how/if this approach can be adapted to `V4` to use with `RandomVariables` (`DensityDist` hasn't been refactored and might be deprecated altogether). I fiddled with it a bit and did not succeed.

Any suggestions?

PR #4806, opened 2021-06-26
Fix various warnings in tests for v3.11.3

maresb:
**Thank your for opening a PR!**

Depending on what your PR does, here are a few things you might want to address in the description:
+ [X] what are the (breaking) changes that this PR makes?
+ [X] important background, or details about the implementation
+ [X] are the changes—especially new features—covered by tests and docstrings?
+ [X] [linting/style checks have been run](https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style)
+ [X] [consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples)
+ [X] right before it's ready to merge, mention the PR in the RELEASE-NOTES.md

* There are no breaking changes, I hope.
* Background is that I was annoyed by superfluous warnings which show up during the test suite and wanted to clean them up.
* There are no new features, so no additional tests/docstrings
* I ran pre-commit.
* No example notebooks updated.
* I'm not sure this deserves mention in the release notes...

PR #4808, opened 2021-06-27
Empty commit to run CI tests

maresb:
Test my claim that `pymc3/tests/test_distributions_random.py` fails on the head of branch `v3.11.3`.


maresb:
Attn: @ricardoV94, for #4806

ricardoV94:
The discrete random tests are definitely failing

ricardoV94:
It is due to scipy `1.7.0`. For instance `test_distributions_random.py::TestScalarParameterSamples::test_poisson` passes locally with scipy `1.6.3` but fails with `1.7.0`

It comes from a `ValueError` in `st.chisquare`

```python
            if diff_gt_tol:
                       f"{relative_diff}")
```

Edit: It's documented here: https://github.com/scipy/scipy/issues/13362

PR #4782, opened 2021-06-18
(Ongoing) Development of DP submodule

larryshamalama:
Here is the beginning of my GSoC project in which I implement a submodule for Dirichlet Processes (DP) and some of its extensions (tbd!). For the weeks to come, I intend to keep this as a [WIP] as I work with @fonnesbeck and @AustinRochford.

PR #4809, opened 2021-06-27
[WIP] Dirichlet Process GSoC

larryshamalama:
Recreating this PR (previously #4782) because I had trouble managing my Git workflow.

I follow closely the `gp` submodule construction in that I create a `DPBase` class and will have other DP extensions build upon `DirichletProcess(DPBase)`.

Things are not yet working, but this is more to show my ongoing work to @AustinRochford  and @fonnesbeck (for now).


larryshamalama:
I ended up with some merge conflicts and wound up with all other commits on top of one and another... is this normal? Will also update with a to-do list for the coming days shortly.

pymc3/dp/dp.py
```diff
@@ -0,0 +1,170 @@
...
+
+        dirac = at.sum(at.lt(Xnew, self.Xs), axis=1)
+
+        empirical_cdf = at.lt(self.base_dist + dirac, Xnew) # P(X \leq x) since X is a distribution
+        linear_comb = at.sum(at.mul(self.weights, empirical_cdf), axis=1)
+        posterior_dp = pm.Deterministic(
+            name=name,
+            var=linear_comb,
+        )
+
```
larryshamalama, line 212: Printing out the shapes, I see that `posterior_dp` has a shape of `[4]`, which is what I want. But things don't work under `pm.sample`... Here is a peak at (a fragment of) the error code

```python
Xs shape:  [1 3]
Xnew shape:  (4, 1)
addition shape:  [ 1 20]
empirical_cdf shape:  [20  4]
weights shape:  [20]
posterior_dp shape:  [4]
Auto-assigning NUTS sampler...
Initializing NUTS using jitter+adapt_diag...
Sequential sampling (1 chains in 1 job)
NUTS: [G0, sticks]
ERROR (aesara.graph.opt): Optimization failure due to: constant_folding-----------------------------| 0.00% [0/2000 00:00<00:00 Sampling chain 0, 0 divergences]
ERROR (aesara.graph.opt): node: Elemwise{lt,no_inplace}(TensorConstant{[-3.  -1. .. 0.5  3.2]}, TensorConstant{[-1  0  1]})
ERROR (aesara.graph.opt): TRACEBACK:
ERROR (aesara.graph.opt): Traceback (most recent call last):
  File "/Users/larryshamalama/anaconda3/envs/pymc3-dev-py39/lib/python3.9/site-packages/aesara/graph/opt.py", line 2014, in process_node
    replacements = lopt.transform(fgraph, node)
  File "/Users/larryshamalama/anaconda3/envs/pymc3-dev-py39/lib/python3.9/site-packages/aesara/graph/opt.py", line 1187, in transform
    return self.fn(*args, **kwargs)
  File "/Users/larryshamalama/anaconda3/envs/pymc3-dev-py39/lib/python3.9/site-packages/aesara/tensor/basic_opt.py", line 4356, in constant_folding
    required = thunk()
  File "/Users/larryshamalama/anaconda3/envs/pymc3-dev-py39/lib/python3.9/site-packages/aesara/graph/op.py", line 631, in rval
    thunk()
  File "/Users/larryshamalama/anaconda3/envs/pymc3-dev-py39/lib/python3.9/site-packages/aesara/link/c/basic.py", line 1772, in __call__
ValueError: Input dimension mismatch. (input[0].shape[0] = 4, input[1].shape[0] = 3)
```

and the error goes on for a bit.

pymc3/dp/util.py
```diff
@@ -0,0 +1,88 @@
...
+    rv_op = stickbreakingweights
+
+    @classmethod
+    def dist(cls, alpha, *args, **kwargs):
+        alpha = at.as_tensor_variable(alpha)
+
+        assert_negative_support(alpha, "alpha", "StickBreakingWeights")
+
+        return super().dist([alpha], **kwargs)
+
```
larryshamalama, line 70: Gives an `IndexError: tuple index out of range` error

pymc3/dp/dp.py
```diff
@@ -0,0 +1,170 @@
...
+# from pymc3.distributions.dist_math import bound
+# from pymc3.distributions.distribution import Distribution
+# from pymc3.dp.util import StickBreakingWeights
+
+__all__ = [
+    "DirichletProcess",
+]
+
+
+class DPBase:
```
larryshamalama: @AustinRochford  @fonnesbeck What are your thoughts on the `DPBase` class as created here? This was largely inspired by the `Base` [class](https://github.com/pymc-devs/pymc3/blob/main/pymc3/gp/gp.py#L46) for Gaussian Processes, so I'm wondering if I should "redesign" how I approach DPs in PyMC3.

Edit: One additional comment behind my decision making is that a general construction for random variables in PyMC3 seems to build upon the `Distribution` class. (Here's my hazy thought process) Because a Dirichlet Process is a stochastic process and whose realizations are distributions, I decided to create its own "base" class rather than inherit from `Distribution`.

larryshamalama (commented):
Need feedback on `DPBase`

pymc3/dp/dp.py
larryshamalama: Solved, but I would need to rethink what we want to "come out" from a DP instance following our meeting today.

pymc3/dp/dp.py
```diff
@@ -0,0 +1,297 @@
...
+                "instead."
+            )
+
+        if not isinstance(base_dist, aesara.graph.basic.Variable):
+            raise ValueError(
+                "base_dist must be of type pymc3.distributions.Distribution"
+                f"but got {type(base_dist)} instead."
+            )
+
+        if K is not None:
```
Sayam753, line 118: Should the checks for `K` come before creating `alpha` parameter?

pymc3/dp/dp.py
```diff
@@ -0,0 +1,297 @@
...
+                    assert K.is_integer()
+            except (AssertionError, AttributeError):
+                raise AttributeError(
+                    "K need to be an int if specified."
+                )
+
+            try:
+                base_dist_shape = base_dist.shape.eval()
+
+                assert len(base_dist_shape) == 1
```
Sayam753, line 130: I dont know much literature, but is there any restriction that the base distribution has to be univariate? I am coming from [Pyro's Dirichet Process tutorial](http://pyro.ai/examples/dirichlet_process_mixture.html) and I think there they use MvNormal distribution as the base.

pymc3/dp/dp.py
```diff
@@ -0,0 +1,297 @@
...
+    #     )
+
+    sticks = at.concatenate(
+        [
+            [1],
+            at.cumprod(1 - betas[:-1]),
+        ]
+    )
+
+    return betas * sticks
```
Sayam753, line 64: Does `betas * sticks` be equal to 1?

The thing I could not understand is -
Lets say I set K=4, then lets represent betas as `b1, b2, b3, b4`.
With the code snippet in this function - 

```
betas * sticks = b1 * 1 + b2 * (1-b1) + b3 * (1-b1) * (1-b2) + b4 * (1-b1) * (1-b2) * (1-b3)
```

Aren't we missing `(1-b1) * (1-b2) * (1-b3) * (1-b4)` to make it one?

pymc3/dp/dp.py
```diff
@@ -0,0 +1,297 @@
...
+        if isinstance(alpha, (int, float)):
+            self.alpha = np.tile(alpha, reps=(K,))
+        elif isinstance(alpha, (np.ndarray, aesara.graph.basic.Variable)):
+            self.alpha = alpha
+        else:
+            raise ValueError(
+                "alpha parameter must be of type float, numpy.ndarray or "
+                f"pymc3.distributions.Distribution, but got {type(alpha)}"
+                "instead."
+            )
```
Sayam753, line 110: I think we can simply do

```
betas = pm.Beta("betas", 1., self.alpha, size=K)
```

And broadcasting semantics are taken care automagically.

Sayam753 (commented):
I am starting to learn about Dirichlet Processes. This PR is awesome to understand things better in PyMC3's context. 
Just some minor comments below -

larryshamalama:
Hey @Sayam753 ! Admittedly I haven't checked this PR in quite some time, I've been spending more time better understanding Dirichlet Processes. Thanks for your comments and I'll reply to them shortly!

pymc3/dp/dp.py
larryshamalama: You are correct! In the case of K = 4, we should effectively append (1 - b1) * ... * (1 - b4) to the list of weights we wish to output

pymc3/dp/dp.py
larryshamalama: I haven't thought of this aspect too much... (was more focused on better understanding DPs lately) Are there arguments to support which check should come first? If I remember correctly, there is some relationship between the lower bound of `K` and `alpha` or something (need to double-check this)

pymc3/dp/dp.py
larryshamalama: Hmm I believe not, I probably put this temporarily and expect to remove it later

twiecki:
@larryshamalama Is this PR still active or should we close it?

larryshamalama:
Yes we can close it :'( There are many more intermediate steps before addressing the API design of a DP submodule

PR #4814, opened 2021-06-28
SMC return inferencedata and perform convergence checks

ricardoV94:
This PR makes `sample_smc` return InferenceData and run convergence checks by default.

@aloctavodia I am not familiar at all with `InferenceData` objects so let me know if I am doing something wrong or not returning as much information as we could / should. 

You mentioned in https://github.com/pymc-devs/pymc3/pull/4802#discussion_r659476595 that we could store the `betas` and the `log_likelihood` in `sample_stats`. However, the `log_likelihood` appears magically in the returned idata, is this one correct? Should we manually add ours instead, since it is already pre-computed. Or did you mean the `log_marginal_likelihood`?

These are the variables stored previously in `trace.report`:
```python
trace.report._n_draws = draws
trace.report._n_tune = 0
trace.report.log_marginal_likelihood = np.array(log_marginal_likelihoods)
trace.report.log_pseudolikelihood = log_pseudolikelihood
trace.report.betas = betas
trace.report.accept_ratios = accept_ratios
trace.report.nsteps = nsteps
trace.report._t_sampling = time.time() - t1
```

Also I am running the same convergence checks as in the normal `pm.sample`, should we do this or instead, implement specific checks for `SMC` (or do nothing at all)?

### DIFF
--- a/pymc3/smc/sample_smc.py
+++ b/pymc3/smc/sample_smc.py
@@ -21,7 +21,11 @@ from collections.abc import Iterable
 
 import numpy as np
 
-from pymc3.backends.arviz import to_inference_data
+from arviz import InferenceData
+
+import pymc3
+
+from pymc3.backends.arviz import dict_to_dataset, to_inference_data
 from pymc3.backends.base import MultiTrace
 from pymc3.model import modelcontext
 from pymc3.parallel_sampling import _cpu_count
@@ -32,6 +36,7 @@ def sample_smc(
     draws=2000,
     kernel="metropolis",
     n_steps=25,
+    *,
     start=None,
     tune_steps=True,
     p_acc_rate=0.85,
@@ -45,6 +50,7 @@ def sample_smc(
     cores=None,
     compute_convergence_checks=True,
     return_inferencedata=True,
+    idata_kwargs=None,
 ):
     r"""
     Sequential Monte Carlo based sampling.
@@ -100,6 +106,8 @@ def sample_smc(
     return_inferencedata : bool, default=True
         Whether to return the trace as an :class:`arviz:arviz.InferenceData` (True) object or a `MultiTrace` (False)
         Defaults to ``True``.
+    idata_kwargs : dict, optional
+        Keyword arguments for :func:`pymc3.to_inference_data`
     Notes
     -----
     SMC works by moving through successive stages. At each stage the inverse temperature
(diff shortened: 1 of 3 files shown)


ricardoV94:
Apologies for the reviewers spam, just thought that all of you might have a better understanding of the InferenceData than me :D

pymc3/smc/sample_smc.py
```diff
@@ -42,6 +43,8 @@ def sample_smc(
     parallel=False,
     chains=None,
     cores=None,
+    compute_convergence_checks=True,
+    return_inferencedata=True,
```
michaelosthege, line 52: Maybe add a `, *,` to force kwargs?

pymc3/smc/sample_smc.py
```diff
@@ -223,10 +231,25 @@ def sample_smc(
     trace.report.nsteps = nsteps
     trace.report._t_sampling = time.time() - t1
 
+    if compute_convergence_checks or return_inferencedata:
+        ikwargs = dict(model=model)
+        idata = to_inference_data(trace, **ikwargs)
```
michaelosthege, line 278: For `pm.sample` we have a kwarg to pass additional parameters such as `log_likelhood=False` for the conversion. This is important for some models that are expensive to evaluate. Without this the progress bar can appear to get stuck at 100 % while the conversion re-evaluates the model at every iteration, potentially taking as long as the original MCMC.

ricardoV94: Yeah, actually in our case we will just use the pre-computed `log_likelihood` since we have it from the sampler. I just have to figure out how to initialize the idata with it, so it doesn't try to recompute it. Are there other optional arguments that the user might want to pass besides this?

pymc3/smc/sample_smc.py
```diff
@@ -223,10 +231,25 @@ def sample_smc(
...
+                stacklevel=2,
+            )
+        else:
+            trace.report._run_convergence_checks(idata, model)
+    trace.report._log_summary()
+
+    posterior = idata if return_inferencedata else trace
     if save_sim_data:
-        return trace, {modelcontext(model).observed_RVs[0].name: np.array(sim_data)}
+        return posterior, {modelcontext(model).observed_RVs[0].name: np.array(sim_data)}
```
aloctavodia, line 295: We should save `sim_data` as posterior predictive samples. The motivation of having `save_sim_data` is that if the simulator is very expensive, we can save resources by saving the simulated data instead of recomputing it

ricardoV94: Okay I'll change it here for the time being, even though I still don't know if we will be able to keep these samples after #4802

aloctavodia:
Just for the record and summarizing our previous conversation. 

* We should keep diagnostics as we have preliminary simulations showing they are useful (even when they were not designed with SMC in mind).

* For ABC (but not "plain SMC") we should overwrite the log-likelihood with the values from "log_pseudolikelihood" (I wonder if with the changes introduced in #4802, this is still necessary... maybe not!)

*  We should store in `sample_stats`. the `log_marginal_likelihood`,  `betas`, `accept_ratios` and `nsteps`.  The  `_t_sampling` should be automatically added to the attributes, the same goes for `_n_tune` (even when this does not make sense for SMC).

ricardoV94:
>  * We should store in `sample_stats`. the `log_marginal_likelihood`,  `betas`, `accept_ratios` and `nsteps`.  The  `_t_sampling` should be automatically added to the attributes, the same goes for `_n_tune` (even when this does not make sense for SMC).

I am reading here https://arviz-devs.github.io/arviz/schema/schema.html?highlight=sample_stats#sample-stats, and it seems `sample_stats` expects one measure per posterior sample, but most of those `SMC` measures do not work like this. Should I just ignore the text and save them there anyway?

pymc3/smc/sample_smc.py
ricardoV94: Realized there is no point in changing it here, because I have no way of testing it. Will address in #4802

ricardoV94:
Pushed changes to include the sampler stats as well. Feels pretty hacky, because I have to deal with the case where chains have different numbers of draws. Let me know if you have a better suggestion

aloctavodia:
I think that sample_stats should be more general, but that's on the ArviZ side. 

Different chains should have the same number of draws.

ricardoV94:
> I think that sample_stats should be more general, but that's on the ArviZ side. 
> 
> Different chains should have the same number of draws. 

What I was calling draws here are the "stages". Sorry for the confusion. Those can vary between chains

michaelosthege (approved):
LGTM

ricardoV94:
The current `sample_stats` dimensions are "chain" x "draw" by default. Should I rename them to "chain" x "stage", or it's not worth the trouble?

aloctavodia:
I think is not worth the trouble at this point.

aloctavodia (approved):

Issue #4800, opened 2021-06-22
Future of pm.Bound?


ricardoV94:
I think @kc611 plans to refactor `pm.Bound` to `V4` as part of his GSOC project. 

Some previous issues have been raised in https://github.com/pymc-devs/pymc3/issues/4643 and https://github.com/pymc-devs/pymc3/issues/1864

I am of the opinion that `pm.Bound` should be changed (if not deprecated altogether) to simply add a custom interval transform (for continuous distributions) and a `pm.Potential` with a `-inf` term for values outside of the bounds (more important for the discrete distributions which cannot be transformed). Importantly `pm.Bound` should not be allowed to sample from in prior/posterior predictive, due to the limitations explained in #4643. This happens because "bounded" variables are not always conditionally independent from other variables (and vice-versa).

Adding the explicit `pm.Potential` would raise a warning during predictive sampling, but I feel it wouldn't be clear enough for users where it's coming from, and in what situations it can be safely ignored (i.e., when bounded variables have constant input parameters). Alternatively, we could also raise an explicit UserWarning when a bounded variable is initialized with variable inputs.

A more useful effort would be to add Truncated variables which can be obtained from the `logcdf` methods that exist for nearly all univariate variables, and which can be directly (if not very efficiently) sampled from. They can also be "observed". Truncated variables are not exactly the same as Bounded variables, but (often) they achieve the same goal and are easier to reason about (citation needed)

Issue #4515, opened 2021-03-09
Write v4 upgrade guide


twiecki:
In anticipation of the big v4 release we should start to write a document walking people through the upgrade process. Specifically, this document should contain:
* Information about the Theano->Aesara switch
* Instructions and code on how to remove any former Theano installation and install Aesara
* A script to guide in the renaming of imports from Theano to Aesara for a code-base as well as NBs (I think @brandonwillard has some code for this already).
* [Please add more]

@MarcoGorelli is that something you can help with?

MarcoGorelli:
Hey @twiecki - which part did you want help with?

ricardoV94:
We should mention somewhere some of the benefits / future developments that are facilitated by this change, so that it doesn't look like we are just trying to reinvent the wheel (i.e., get people excited for the change and not simply bothered)

OriolAbril:
This will be also great to help outreachy interns learn how to update all the other notebooks to take advantage of all the improvements. 

Maybe we can start by updating the getting started notebook https://github.com/pymc-devs/pymc-examples/issues/31 and from there start gathering and writing down all the changes that are needed to update the code.

brandonwillard:
Here are the basics:
* `theano` is now `aesara`
* `theano.gof` is now `aesara.graph` and, more specifically, `theano.gof.graph` is `aesara.graph.basic`
* config options previously under `.`-delimited "sections", like `theano.config.gcc`, are now replaced with double-underscores (e.g. `theano.config.gcc.cxxflags` is now `aesara.config.gcc__cxxflags`)
* most internal classes and functions are no longer exported at the package level (e.g. `theano.Variable` no longer works; use `aesara.graph.basic.Variable` instead)

twiecki:
@MarcoGorelli Ideally the conversion script, is that something you are familiar with?

MarcoGorelli:
I think you could subclass `ast.NodeVisitor` and override some methods like `visit_ImportFrom` and `visit_Name` ([example](https://github.com/MarcoGorelli/absolufy-imports/blob/46719e1aba642ee7af6716dc30d4bfc8d59d8959/absolufy_imports.py)) - I'm starting a new job next week though and likely won't have capacity for this in the near future though, sorry

brandonwillard:
I wrote a small template/example for `Distribution`-to-`RandomVariable` conversions [here](https://github.com/pymc-devs/pymc3/issues/4518#issuecomment-806236360).

[...]

PR #2277, opened 2017-06-06
Allow arrays in pm.Bound

aseyboldt:
`pm.Bound` doesn't support array valued bounds at the moment. This is fixed in this PR. I also refactored the `Bounded.__init__` function a bit.

PR #4815, opened 2021-06-28
Refactored Bound for v4

kc611:
This PR refactors the `Bound` distribution for `v4`

### DIFF
--- a/pymc3/distributions/bound.py
+++ b/pymc3/distributions/bound.py
@@ -12,42 +12,43 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-from numbers import Real
-
-import aesara.tensor as at
 import numpy as np
 
-from pymc3.aesaraf import floatX
-from pymc3.distributions import transforms
+from aesara.tensor import as_tensor_variable
+from aesara.tensor.random.op import RandomVariable
+from aesara.tensor.var import TensorVariable
+
+from pymc3.aesaraf import floatX, intX
+from pymc3.distributions import _logp
+from pymc3.distributions.continuous import BoundedContinuous
 from pymc3.distributions.dist_math import bound
-from pymc3.distributions.distribution import Continuous, Discrete, Distribution
+from pymc3.distributions.distribution import Continuous, Discrete
+from pymc3.distributions.shape_utils import to_tuple
+from pymc3.model import modelcontext
 
 __all__ = ["Bound"]
 
 
-class _Bounded(Distribution):
-    def __init__(self, distribution, lower, upper, default, *args, **kwargs):
-        self.lower = lower
-        self.upper = upper
-        self._wrapped = distribution.dist(*args, **kwargs)
+class BoundRV(RandomVariable):
+    name = "bound"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0]
+    dtype = "floatX"
+    _print_name = ("Bound", "\\operatorname{Bound}")
 
-        if default is None:
-            defaults = self._wrapped.defaults
-            for name in defaults:
-                setattr(self, name, getattr(self._wrapped, name))
-        else:
-            defaults = ("_default",)
-            self._default = default
+    @classmethod
+    def rng_fn(cls, rng, distribution, lower, upper, size):
+        raise NotImplementedError("Cannot sample from a bounded variable")
 
-        super().__init__(
-            shape=self._wrapped.shape,
-            dtype=self._wrapped.dtype,
-            initval=self._wrapped.initval,
-            defaults=defaults,
-            transform=self._wrapped.transform,
-        )
 
-    def logp(self, value):
+boundrv = BoundRV()
+
+
+class _ContinuousBounded(BoundedContinuous):
+    rv_op = boundrv
+    bound_args_indices = [1, 2]
+
+    def logp(value, distribution, lower, upper):
         """
         Calculate log-probability of Bounded distribution at specified value.
 
@@ -55,155 +56,59 @@ class _Bounded(Distribution):
         ----------
         value: numeric
             Value for which log-probability is calculated.
+        distribution: TensorVariable
+            Distribution which is being bounded
+        lower: numeric
+            Lower bound for the distribution being bounded.
+        upper: numeric
+            Upper bound for the distribution being bounded.
 
         Returns
         -------
         TensorVariable
         """
-        logp = self._wrapped.logp(value)
-        bounds = []
-        if self.lower is not None:
-            bounds.append(value >= self.lower)
-        if self.upper is not None:
-            bounds.append(value <= self.upper)
-        if len(bounds) > 0:
-            return bound(logp, *bounds)
-        else:
-            return logp
+        logp = _logp(distribution.owner.op, value, {}, *distribution.owner.inputs[3:])
+        return bound(logp, (value >= lower), (value <= upper))
 
-    def _random(self, lower, upper, point=None, size=None):
-        lower = np.asarray(lower)
-        upper = np.asarray(upper)
-        if lower.size > 1 or upper.size > 1:
-            raise ValueError(
-                "Drawing samples from distributions with array-valued bounds is not supported."
-            )
-        total_size = np.prod(size).astype(np.int)
-        samples = []
-        s = 0
-        while s < total_size:
-            sample = np.atleast_1d(self._wrapped.random(point=point, size=total_size)).flatten()
 
-            select = sample[np.logical_and(sample >= lower, sample <= upper)]
-            samples.append(select)
-            s += len(select)
-        if size is not None:
-            return np.reshape(np.concatenate(samples)[:total_size], size)
-        else:
-            return samples[0]
+class DiscreteBoundRV(BoundRV):
+    name = "discrete_bound"
+    dtype = "int64"
 
-    def random(self, point=None, size=None):
+
+discrete_boundrv = DiscreteBoundRV()
+
+
+class _DiscreteBounded(Discrete):
+    rv_op = discrete_boundrv
+
+    def __new__(cls, *args, **kwargs):
+        transform = kwargs.get("transform", None)
+        if transform is not None:
+            raise ValueError("Cannot transform discrete variable.")
+        return super().__new__(cls, *args, **kwargs)
+
+    def logp(value, distribution, lower, upper):
         """
-        Draw random values from Bounded distribution.
+        Calculate log-probability of Bounded distribution at specified value.
 
         Parameters
         ----------
-        point: dict, optional
-            Dict of variable values on which random values are to be
-            conditioned (uses default point if not specified).
-        size: int, optional
-            Desired size of random sample (returns one sample if not
-            specified).
+        value: numeric
+            Value for which log-probability is calculated.
+        distribution: TensorVariable
+            Distribution which is being bounded
+        lower: numeric
+            Lower bound for the distribution being bounded.
+        upper: numeric
+            Upper bound for the distribution being bounded.
 
         Returns
         -------
-        array
+        TensorVariable
         """
-        # if self.lower is None and self.upper is None:
-        #     return self._wrapped.random(point=point, size=size)
-        # elif self.lower is not None and self.upper is not None:
-        #     lower, upper = draw_values([self.lower, self.upper], point=point, size=size)
-        #     return generate_samples(
-        #         self._random,
-        #         lower,
-        #         upper,
-        #         dist_shape=self.shape,
-        #         size=size,
-        #         not_broadcast_kwargs={"point": point},
-        #     )
-        # elif self.lower is not None:
-        #     lower = draw_values([self.lower], point=point, size=size)
-        #     return generate_samples(
-        #         self._random,
-        #         lower,
-        #         np.inf,
-        #         dist_shape=self.shape,
-        #         size=size,
-        #         not_broadcast_kwargs={"point": point},
-        #     )
-        # else:
-        #     upper = draw_values([self.upper], point=point, size=size)
-        #     return generate_samples(
-        #         self._random,
-        #         -np.inf,
-        #         upper,
-        #         dist_shape=self.shape,
-        #         size=size,
-        #         not_broadcast_kwargs={"point": point},
-        #     )
-        pass
-
-
-class _DiscreteBounded(_Bounded, Discrete):
-    def __init__(self, distribution, lower, upper, transform="infer", *args, **kwargs):
-        if transform == "infer":
-            transform = None
-        if transform is not None:
-            raise ValueError("Can not transform discrete variable.")
-
-        if lower is None and upper is None:
-            default = None
-        elif lower is not None and upper is not None:
-            default = (lower + upper) // 2
-        if upper is not None:
-            default = upper - 1
-        if lower is not None:
-            default = lower + 1
-
-        super().__init__(distribution, lower, upper, default, *args, transform=transform, **kwargs)
-
-
-class _ContinuousBounded(_Bounded, Continuous):
-    r"""
-    An upper, lower or upper+lower bounded distribution
-
-    Parameters
-    ----------
-    distribution: pymc3 distribution
-        Distribution to be transformed into a bounded distribution
-    lower: float (optional)
-        Lower bound of the distribution, set to -inf to disable.
-    upper: float (optional)
-        Upper bound of the distribibution, set to inf to disable.
-    tranform: 'infer' or object
-        If 'infer', infers the right transform to apply from the supplied bounds.
-        If transform object, has to supply .forward() and .backward() methods.
-        See pymc3.distributions.transforms for more information.
-    """
-
-    def __init__(self, distribution, lower, upper, transform="infer", *args, **kwargs):
-        if lower is not None:
-            lower = at.as_tensor_variable(floatX(lower))
-        if upper is not None:
-            upper = at.as_tensor_variable(floatX(upper))
-
-        if transform == "infer":
-            if lower is None and upper is None:
-                transform = None
-                default = None
-            elif lower is not None and upper is not None:
-                transform = transforms.interval(lower, upper)
-                default = 0.5 * (lower + upper)
-            elif upper is not None:
-                transform = transforms.upperbound(upper)
-                default = upper - 1
-            else:
-                transform = transforms.lowerbound(lower)
-                default = lower + 1
-        else:
-            default = None
-
-        super().__init__(distribution, lower, upper, default, *args, transform=transform, **kwargs)
+        logp = _logp(distribution.owner.op, value, {}, *distribution.owner.inputs[3:])
+        return bound(logp, (value >= lower), (value <= upper))
 
 
 class Bound:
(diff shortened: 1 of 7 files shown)


pymc3/distributions/bound.py
```diff
@@ -263,47 +75,61 @@ class Bound:
...
-
-    def __call__(self, name, *args, **kwargs):
-        if "observed" in kwargs:
-            raise ValueError(
-                "Observed Bound distributions are not supported. "
-                "If you want to model truncated data "
-                "you can use a pm.Potential in combination "
-                "with the cumulative probability function."
-            )
+        # if "observed" in kwargs:
```
kc611: Maybe we can apply a `Potential` internally?

pymc3/distributions/bound.py
```diff
@@ -263,47 +75,61 @@ class Bound:
...
+                if lower is not None and upper is not None:
+                    transform = transforms.interval(lower, upper)
+                    default = 0.5 * (lower + upper)
+                elif upper is not None:
+                    transform = transforms.upperbound(upper)
+                else:
+                    transform = transforms.lowerbound(lower)
+
+        elif issubclass(distribution, Discrete):
+            raise ValueError("Can not transform discrete variable.")
```
kc611: Not sure if `interval` is the right thing for `Discrete` distributions.

pymc3/distributions/bound.py
```diff
@@ -14,218 +14,30 @@
...
+    name = "bound"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0]
+    dtype = "floatX"
+    _print_name = ("Bound", "\\operatorname{Bound}")
 
-        super().__init__(distribution, lower, upper, default, *args, transform=transform, **kwargs)
+    @classmethod
+    def rng_fn(cls, rng, distribution, lower, upper, size):
+        raise NotImplementedError("Cannot sample from a bounded variable")
```
kc611, line 41: @ricardoV94, This is what you described in https://github.com/pymc-devs/pymc3/issues/4800 , right ?

ricardoV94: Yes. But it means you need to compute the `initval` instead, otherwise the model will raise an error when a bounded variable is created.

pymc3/distributions/bound.py
ricardoV94: What do you mean?

pymc3/distributions/bound.py
ricardoV94: What do you mean?

pymc3/distributions/bound.py
```diff
@@ -263,47 +75,61 @@ class Bound:
...
-            return _ContinuousBounded(
-                name, self.distribution, self.lower, self.upper, transform, *args, **kwargs
-            )
-        elif issubclass(self.distribution, Discrete):
-            return _DiscreteBounded(
-                name, self.distribution, self.lower, self.upper, transform, *args, **kwargs
-            )
+        transform = kwargs.pop("transform", None)
+        if issubclass(distribution, Continuous):
+            if transform == None:
```
ricardoV94, line 95:
```suggestion
            if transform is None:
```

pymc3/distributions/bound.py
```diff
@@ -263,47 +75,61 @@ class Bound:
...
+        TensorVariable
+        """
+        logp = _logp(distribution.owner.op, {}, value, *distribution.owner.inputs[3:])
+        bounds = []
+        if lower is not None:
+            bounds.append(value >= lower)
+        if upper is not None:
+            bounds.append(value <= upper)
+        if len(bounds) > 0:
+            return bound(logp, *bounds)
```
kc611: How is the `bounds` function different from `interval` ?

pymc3/distributions/bound.py
kc611: In `v3` the implementation avoided use of `interval` transform upon `Discrete` (maybe because it uses `sigmoid` that would lead to values not in the allowed values of the discrete distribution) so I'm not sure about applying the same upon `Discrete`. What could be a good approach here ?

pymc3/distributions/bound.py
kc611: Nevermind, I confused it with something else.

pymc3/distributions/bound.py
ricardoV94: I think you should follow the original, transform only continuous variables and not transform discrete ones. The bounding of discrete variables is achieved by the `bound` call in the `logp`

pymc3/distributions/bound.py
ricardoV94: You mean `bound` from `distributions.dist_math`? It simply wraps a switch around the `logp` expression, which returns `-inf` if any of the bounds is not respected. In this case it's basically `at.switch((value < lower) or (value > upper), -np.inf, logp)`

pymc3/distributions/bound.py
```diff
@@ -263,47 +75,61 @@ class Bound:
...
+            raise ValueError("Can not transform discrete variable.")
         else:
             raise ValueError("Distribution is neither continuous nor discrete.")
 
-    def dist(self, *args, **kwargs):
-        if issubclass(self.distribution, Continuous):
-            return _ContinuousBounded.dist(
-                self.distribution, self.lower, self.upper, *args, **kwargs
-            )
+        if transform is not None:
```
ricardoV94: transform has a flag value of `UNSET` by default. If transform is `None`, the user is requesting no transform explicitly. https://github.com/pymc-devs/pymc3/blob/ce88e7392c172da78b2bd0ea86b9de10b5824124/pymc3/util.py#L30

This comment applies above, where transform first appears

pymc3/distributions/bound.py
```diff
@@ -263,47 +75,61 @@ class Bound:
...
+            kwargs["transform"] = transform
+        return super.dist([distribution, lower, upper], **kwargs)
```
ricardoV94: Is distribution a `class` or a `rv_out`? If it's the former this will fail when the argument is passed to the `RandomVariable` no?

pymc3/distributions/bound.py
ricardoV94: These transforms have changed since this code was last touched, see here https://github.com/pymc-devs/pymc3/issues/4515#issuecomment-821250583

pymc3/distributions/bound.py
kc611: It's supposed to be an `rv_out`. Maybe we should raise if a class is passed. (Or we could instantiate it?)

ricardoV94: If it's a rv_out I don't think you can use the `isinstance` class checks above, but the RV op dtype can be used as a proxy. 

If it's a Class, I don't think you can instantiate it because you would need to know the arguments, which the user has not provided yet. 

Anyway I think it makes more sense to work with a rv_out, as is done in the rest of the library (we will just need to raise a useful error because it will be a breaking change from V3). But I wonder if you'll have these intermediate rvs bothering you when creating the logp graph...

kc611: Oh, this is one thing that needs discussion,

Do we keep the previous `pm.Bound` interface of `call` upon a distribution class and then passing it the arguments ? Or do we make the distribution (just the `Distribution.dist`) separately in the Model and then `Bound` it ? 

There is also a middle ground where we can pass the class and arguments of the distribution to `Bound` and it gets combined in the `dist` method of `Bound`.

ricardoV94: Yeah it needs discussion. Personally I prefer the `.dist` approach as that is much more coherent with the rest of the library.

On the other hand, pm.Bound is being refactored mostly for backwards compatibility because a lot of people have got used to it...

pymc3/distributions/bound.py
```diff
@@ -75,40 +81,64 @@ class Bound(Distribution):
...
+
+    def __new__(cls, *args, **kwargs):
+        transform = kwargs.get("transform", UNSET)
+        if transform is UNSET:
+            kwargs["transform"] = cls.default_transform(kwargs)
+        return super().__new__(cls, *args, **kwargs)
+
     @classmethod
-    def dist(self, distribution, lower=None, upper=None):
+    def default_transform(cls, kwargs_dict):
```
ricardoV94, line 93: This might work out of the box if you inherit from `BoundedContinuous`, which applies this logic by default. You just need to tell what are the indexes of the arguments that specify "lower" and "upper", which in this case is (1, 2). There are some examples in `distributions/continuous.py`. In either case you still need to override transform to be `None` for discrete variables.

pymc3/distributions/bound.py
```diff
@@ -75,40 +81,64 @@ class Bound(Distribution):
                     'par2', mu=0.0, sigma=1.0, initval=1.0)
     """
 
+    rv_op = boundrv
+
+    def __new__(cls, *args, **kwargs):
+        transform = kwargs.get("transform", UNSET)
```
ricardoV94: should not be applied to `Discrete` variables.

pymc3/distributions/bound.py
```diff
@@ -285,25 +121,39 @@ def __call__(self, name, *args, **kwargs):
...
+                initval = np.full(size, floatX(0))
+
+        lower = lower if lower else -np.inf
+        upper = upper if upper else np.inf
+        res = super().dist([distribution, lower, upper], size=size, **kwargs)
+        res.tag.test_value = initval
+
+        return res
+
+    def logp(value, distribution, lower, upper):
```
kc611: Not sure why but the value of `value` changes during the `logp` calculation hence the test failures.

twiecki:
@kc611 How close do you think this is to merge? If there's still a ways to go, we might want to focus on the aeppl integration instead where we can get mixture almost for free. CC @brandonwillard

kc611:
I'd say we're halfway through over here. But if we're bringing in `aeppl` https://github.com/aesara-devs/aeppl/pull/22 PR of `CensoredRV` can be a much better replacement for `Bound` logic. 

cc @ricardoV94

ricardoV94:
`CensoredRV` is not a good replacement. We would need something like `TruncatedRV` but that is still some time away

pymc3/distributions/bound.py
```diff
@@ -263,20 +81,39 @@ class Bound:
                     'par2', mu=0.0, sigma=1.0, initval=1.0)
     """
 
-    def __init__(self, distribution, lower=None, upper=None):
+    rv_op = boundrv
+
+    def __new__(cls, *args, **kwargs):
+        transform = kwargs.get("transform", UNSET)
+        base_rv_type = args[1].owner.op.dtype
+        if transform is UNSET and base_rv_type not in ["intX", "int32", "int64"]:
```
ricardoV94: Actually it seems that because of the Distribution Metaclass magic we can check if a random `Op` is of type `Continuous`:

```python
isinstance(at.random.normal().owner.op, pm.Continuous)  # True
```

ricardoV94: You should probably keep the old Exception if a transform is provided for a Continuous distribution

```python
 raise ValueError("Cannot transform a discrete variable")
```

pymc3/distributions/bound.py
```diff
@@ -263,20 +81,39 @@ class Bound:
...
+            kwargs["transform"] = cls.default_transform(kwargs)
+        return super().__new__(cls, *args, **kwargs)
+
+    @classmethod
+    def default_transform(cls, kwargs_dict):
```
ricardoV94: This seems to duplicate a lot of the logic already implemented in:
https://github.com/pymc-devs/pymc3/blob/9d90c891de5a2f491c51f8311b387f051e27538a/pymc3/distributions/continuous.py#L150

We can try to inherit from it in the current Class or make the `default_transform` in `BoundedContinuous` a `staticmethod` that can be called from other places. 

Also this`Bound` class might still have to be split in `Discrete` and `Continuous` classes, which I think is necessary for some samplers to decide how to deal with the variables. In that case `Bound` can be just a wrapper that returns / instantiates the right type of Variable.

pymc3/distributions/bound.py
```diff
@@ -14,40 +14,44 @@
...
 
-class _Bounded(Distribution):
-    def __init__(self, distribution, lower, upper, default, *args, **kwargs):
-        self.lower = lower
-        self.upper = upper
-        self._wrapped = distribution.dist(*args, **kwargs)
+class BoundRV(RandomVariable):
+    name = "bound"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0]
```
ricardoV94, line 35: Not sure about the `ndims_params`. In fact the distribution is characterized by the `original params + [0, 0]`. Also the `dtype` should be the same as the original... But I don't know what is the best approach here, since obviously you won't know in advance which distribution is being bounded

pymc3/distributions/bound.py
```diff
@@ -60,153 +64,39 @@ def logp(self, value):
...
+
+        return bound(logp, *bounds)
+
+
+class _ContinuousBounded(BoundedContinuous):
+    rv_op = boundrv
+
+    def __new__(cls, *args, **kwargs):
+        cls.bound_args_indices = [1, 2]
+        return super().__new__(cls, *args, **kwargs)
```
ricardoV94: Does this work?

```suggestion
    rv_op = boundrv
    cls.bound_args_indices = [-2, -1]
```

kc611: Yes , that's probably a better way to do it.

pymc3/distributions/bound.py
kc611: I thought it wouldn't matter in this case since we are never actually drawing values of that shape in the first place. `original ndims + [0, 0]` will probably be a good idea to implement here if that's not the case.

pymc3/distributions/bound.py
```diff
@@ -60,153 +64,36 @@ def logp(self, value):
...
-            return samples[0]
+        logp = _logp(distribution.owner.op, value, {}, *distribution.owner.inputs[3:])
+        bounds = [(value >= lower), (value <= upper)]
+
+        return bound(logp, *bounds)
+
+
+class _ContinuousBounded(BoundedContinuous):
+    rv_op = boundrv
+    cls.bound_args_indices = [-2, -1]
```
ricardoV94: My bad...
```suggestion
    bound_args_indices = [-2, -1]
```

pymc3/tests/test_distributions.py
```diff
@@ -19,22 +19,6 @@
 import aesara.tensor as at
 import numpy as np
 import numpy.random as nr
-
```
ricardoV94: I guess this removed section was a mistake?

kc611: Oh `pre-commit` was complaining about it not being used, I though it was some sort of residual code. What does it do anyway ?

pymc3/distributions/bound.py
```diff
@@ -60,153 +60,40 @@ def logp(self, value):
...
+        return bound(logp, *bounds)
 
-    def random(self, point=None, size=None):
+
+class _ContinuousBounded(BoundedContinuous):
+
+    bound_args_indices = [1, 2]
+
+    def __new__(cls, *args, **kwargs):
+        cls.rv_op = BoundRV("bound", 0, [args[1][0].ndim, 0, 0], args[1][0].dtype)
```
kc611: For some reason doing this breaks `logp` generation.

pymc3/distributions/bound.py
```diff
@@ -266,25 +184,37 @@ def __call__(self, name, *args, **kwargs):
...
+            MvNormalRV,
+            MvStudentTRV,
+        )
 
-        elif issubclass(self.distribution, Discrete):
-            return _DiscreteBounded.dist(self.distribution, self.lower, self.upper, *args, **kwargs)
-        else:
-            raise ValueError("Distribution is neither continuous nor discrete.")
+        if isinstance(distribution.owner.op, mv_var_list):
+            raise NotImplementedError("Bounding of MultiVariate RVs is not yet supported.")
```
kc611, line 200: There should be a better way of doing this rather than a long list of Multivariate RVs

pymc3/distributions/bound.py
```diff
@@ -266,25 +187,32 @@ def __call__(self, name, *args, **kwargs):
...
             )
-        else:
-            raise ValueError("Distribution is neither continuous nor discrete.")
 
-    def dist(self, *args, **kwargs):
-        if issubclass(self.distribution, Continuous):
-            return _ContinuousBounded.dist(
-                self.distribution, self.lower, self.upper, *args, **kwargs
-            )
+        if distribution.owner.ndim_supp != 0:
```
ricardoV94: Does this fix it?
```suggestion
        if distribution.owner.op.ndim_supp != 0:
```

pymc3/distributions/bound.py
```diff
@@ -232,32 +132,53 @@ class Bound:
...
+            negative_normal = pm.Bound(normal_dist, upper=0.0)
+
     """
 
-    def __init__(self, distribution, lower=None, upper=None):
-        if isinstance(lower, Real) and lower == -np.inf:
-            lower = None
-        if isinstance(upper, Real) and upper == np.inf:
-            upper = None
+    def __new__(cls, name, distribution, lower=None, upper=None, size=None, **kwargs):
```
ricardoV94: If we are in a `modelcontext` we should check here that the `distribution` is not in `model` (which would mean the user did not use `.dist()` and the original variable would be accounted on the model logp as well. 

~~In addition we can add the check that it is a TensorVariable, if not, the user is probably trying to work with the old API~~ I see you already did this. Then the first error message can also go there

pymc3/distributions/bound.py
```diff
@@ -232,32 +133,70 @@ class Bound:
...
-        self.lower = lower
-        self.upper = upper
+        lower, upper, initval = cls._set_values(lower, upper, size, initval)
+
+        if size is not None:
+            kwargs["size"] = size
 
-    def __call__(self, name, *args, **kwargs):
+        if isinstance(distribution.owner.op, Continuous):
+            # TODO: Get actual rv names here, can be done using kwargs
```
ricardoV94: What was this TODO about?

pymc3/distributions/bound.py
```diff
@@ -232,32 +133,70 @@ class Bound:
     .. code-block:: python
 
         with pm.Model():
-            NegativeNormal = pm.Bound(pm.Normal, upper=0.0)
```
twiecki, line 235: Do we not support this use-case anymore?

ricardoV94: No, you have to pass a .dist(). There is a very informative error message with a code example if users do that.

This was done to simplify the implementation and also be more in line with the rest of the API (e.g when you specify distributions for Mixtures, LKJCorr priors, RWs?). Also in the future, Truncated and Censored distributions would work the same way.

~~I checked the original PRs for Bound and this was actually the initial intended API, but they couldn't make it work.~~ It was not, but it came up a couple of times, e.g: https://github.com/pymc-devs/pymc3/pull/2277#discussion_r124477153

pymc3/tests/test_distributions.py
```diff
@@ -2669,74 +2642,140 @@ def ref_pdf(value):
...
-    dist = ArrayNormal.dist(mu=0, sigma=1, size=2)
-    logp_dist = logp(dist, [0.5, 3.5]).eval({lower: lower.tag.test_value})
-    assert_equal(logp_dist, -np.array([np.inf, np.inf]))
-    assert dist.transform is not None
+class TestBound:
+    """Tests for pm.Bound distribution"""
 
-    with Model():
-        a = ArrayNormal("c", size=2)
-        assert_equal(a.tag.test_value, np.array([2, 2.5]))
```
ricardoV94: We need tests that cover the case where lower / upper is a vector, which were covered in these `ArrayNormal` checks

pymc3/distributions/bound.py
```diff
@@ -14,196 +14,97 @@
...
+boundrv = BoundRV()
+
+
+class _DiscreteBounded(Discrete):
+    rv_op = boundrv
+
+    def __new__(cls, *args, **kwargs):
+        transform = kwargs.get("transform", None)
+        if transform is not None:
+            raise ValueError("Cannot transform discrete variable.")
```
ricardoV94: Codecov is suggesting some important error branches that are not covered by the new tests

pymc3/distributions/bound.py
```diff
@@ -232,32 +133,70 @@ class Bound:
...
+
+        lower, upper, initval = cls._set_values(lower, upper, size, initval)
+
+        if isinstance(distribution.owner.op, Continuous):
+            # TODO: Get actual rv names here, can be done using kwargs
+            res = _ContinuousBounded.dist(
+                [distribution, lower, upper], size=size, initval=initval, **kwargs
+            )
+        else:
+            res = _DiscreteBounded.dist(
```
ricardoV94, line 213: This branch should be tested

ricardoV94 (commented):
Left some comments, We should wrap the `lower / upper` with `at.as_tensor_variable([floatX|intX])`, probably in the `.dist` method of the private classes `_[Discrete|Continuous]Bounded`, to not interfere with the `cls.set_values`. 

Also after #4867 that logic can probably be simplified, as we will be able to just return symbolic initvals on demand

twiecki (approved):

twiecki:
Thanks @kc611 and @ricardoV94!

PR #4823, opened 2021-06-29
Combine coverage across test matrix

michaelosthege:
The coverage badge in the readme shows only 69 %.. It's probably because the coverage stats are not combined across test jobs.

Even better would be to add a separate upload job that depends on both `ubuntu` and `windows` jobs. This way we could merge coverage across platforms.

Does anybody here speak GitHub Action YAML ?

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -125,7 +125,7 @@ jobs:
       - name: Run tests
         run: |
           conda activate pymc3-dev-py37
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
no objections here, looks like pandas does `--cov-append` too

https://github.com/pandas-dev/pandas/blob/12513c4cdb14fe70ec7226e12bdea70faccdc2cc/ci/run_tests.sh#L12-L14

twiecki (approved):

PR #4824, opened 2021-06-30
Fail CI on XPASS and remove XFAIL marks on tests that succeed

### LINKED ISSUE #4743: Make XPASSing tests fail the CI so we remove XFAIL marks
On the way to the `v4.0.0` release we're striving to get our test suite back ✔.

We currently have a lot of tests that are already passing again, but are still marked as XFAIL.

Example: https://github.com/pymc-devs/pymc3/runs/2754411365#step:7:33

https://github.com/pymc-devs/pymc3/commit/70f2afa9384cf13aa9a81e4228411a4ed4203fe7 was supposed to make XPASSing tests fail the CI so we detect them and remove the XFAIL.
But it looks like this setting does not have an effect.

michaelosthege:
Closes #4743

This may lead to quite a number of CI failures during refactoring, but this way we can solidify the refactored API faster.

My local runs of the test suites indicated that we have at least 72 tests that are XPASSing... https://github.com/conda-forge/pymc3-feedstock/issues/69#issuecomment-869166748

The first commit/CI run will fail spectacularly. I'll fowllow it up with a commit that removes the XFAIL marks on all the tests that XPASS.

### DIFF
--- a/pymc3/tests/test_distributions.py
+++ b/pymc3/tests/test_distributions.py
@@ -1274,7 +1274,6 @@ class TestMatchesScipy:
             n_samples=10,
         )
 
-    @pytest.mark.xfail(reason="Distribution not refactored yet")
     @pytest.mark.parametrize(
         "mu, p, alpha, n, expected",
         [
@@ -1580,7 +1579,6 @@ class TestMatchesScipy:
             lambda value, alpha, beta, n: sp.betabinom.logpmf(value, a=alpha, b=beta, n=n),
         )
 
-    @pytest.mark.xfail(condition=(aesara.config.floatX == "float32"), reason="Fails on float32")
     @pytest.mark.skipif(
         condition=(SCIPY_VERSION < parse("1.4.0")), reason="betabinom is new in Scipy 1.4.0"
     )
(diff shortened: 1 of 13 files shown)


pymc3/tests/test_distributions.py
```diff
@@ -2573,10 +2571,6 @@ def test_logitnormal(self):
             decimal=select_by_precision(float64=6, float32=1),
         )
 
-    @pytest.mark.xfail(
```
ricardoV94: These logp/logcdf tests probably have to stay as `xfail`. The problem is that only a subset of parameter / values is tested on each run and the failing cases don't always show up.

michaelosthege: But that doesn't work in combination with the strict flag.
Which parameter values are the problematic ones?
Or what's the exception and then we'll catch that specifically.

michaelosthege: I undid the removal of the xfail mark, but added an Exception raised on purpose to prevent XPASS.

ricardoV94:
What is the plan here? I imagine this is just a temporary switch until we refactor everything that's left? Or we don't want to support `xfails` going into the future at all? If it's the first, we should open a reminder issue to reinstate the old behavior.

Also, I wouldn't be surprised if those tests based on NUTS sampling and alike are not properly seeded, and that they may fail now and then. Unlike the logp/logcdf checks, this is just a hunch.

Other than that it LGTM

michaelosthege:
We can absolutely continue to have XFAILs. These changes are just to require that if a tests is marked XFAIL, then it must not pass.
For refactoring and bugfixes I expect this to help not only near-term, but also long-term, because it reduces the number of zombie-tests that are left with `mark.xfail` decorators just because one did not notice them passing after some bugfix.

The only downside I see is that we can't `mark.xfail` flaky tests just because they fail every now and then. But doing that effectively makes them to zombies anyway, so we can just as well raise the exception on purpose. (Like I did here.)

ricardoV94 (approved):

PR #4825, opened 2021-07-01
Refactoring Mixture RV for v4

kc611:
This PR provides a `Mixture` refactor for `v4`. It takes an approach similar to https://github.com/pymc-devs/pymc3/pull/4802 and creates an `MixtureRV` 'on-the-fly', (not sure how good/bad this approach would be for `Mixture` though).

### DIFF
--- a/pymc3/distributions/mixture.py
+++ b/pymc3/distributions/mixture.py
@@ -12,30 +12,43 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-from collections.abc import Iterable
-
-import aesara
 import aesara.tensor as at
 import numpy as np
 
-from pymc3.aesaraf import _conversion_map, take_along_axis
+from aesara.tensor.random.op import RandomVariable
+
+from pymc3.aesaraf import take_along_axis
+from pymc3.distributions import _logp
 from pymc3.distributions.continuous import Normal, get_tau_sigma
 from pymc3.distributions.dist_math import bound
-from pymc3.distributions.distribution import Discrete, Distribution
+from pymc3.distributions.distribution import Distribution
 from pymc3.distributions.shape_utils import to_tuple
 from pymc3.math import logsumexp
 
 __all__ = ["Mixture", "NormalMixture", "MixtureSameFamily"]
 
 
-def all_discrete(comp_dists):
-    """
-    Determine if all distributions in comp_dists are discrete
-    """
-    if isinstance(comp_dists, Distribution):
-        return isinstance(comp_dists, Discrete)
-    else:
-        return all(isinstance(comp_dist, Discrete) for comp_dist in comp_dists)
+class MixtureRV(RandomVariable):
+    def __init__(self, w, comp_dist):
+        self.name = "mixture"
+        self.comp_dist = comp_dist
+        self.w = w
+        self.ndim_supp = 0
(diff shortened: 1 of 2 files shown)


pymc3/distributions/mixture.py
```diff
@@ -12,30 +12,75 @@
...
+        orig_size = size
+        # Flatten the size to a one dimensional one
+        size = np.prod(size)
+
+        comp_dist = self.comp_dist
+        w = self.w
+
+        if isinstance(w, (list, tuple)):
+            w = np.array(w)
+        # Take care of case when w is a distribution
```
kc611: Not sure how `Mixture` works when `w` is a distribution. Are the component weights drawn from the random of `w` ?

pymc3/tests/test_mixture.py
```diff
@@ -583,3 +584,17 @@ def test_broadcasting_in_shape(self):
...
         assert prior["mix"].shape == (self.n_samples, 1000)
+
+
+def test_mixture():
+    with pm.Model() as model:
+        components = [pm.Poisson.dist(mu=3), pm.Poisson.dist(mu=3)]
+
+        # w = pm.Dirichlet("w", a=np.array([1, 1]))  # two mixture component weights.
+
+        like = pm.Mixture("like", w=[0.5, 0.5], comp_dists=components)
```
kc611: Only tested for this way of 'mixing' as of now.

pymc3/distributions/mixture.py
ricardoV94: `w` is never a distribution, but an explicit parameter, I think

Sayam753: > `w` is never a distribution

It can be Dirichlet distribution/stickbreaking over beta distribution as well.

pymc3/distributions/mixture.py
```diff
@@ -12,30 +12,75 @@
...
+        _samples = []
+        for dist in comp_dist:
+            curr_op = dist.owner.op
+            curr_inputs = dist.owner.inputs
+
+            if size is not None:
+                curr_inputs[1] = at.as_tensor_variable(size)
+
+            curr_op_random = at.as_tensor_variable(curr_op.make_node(*curr_inputs))
+            curr_op_random = curr_op_random.eval()
```
ricardoV94: I am not very optimistic about this approach, but in any case, you can try to call the `dist.rng_fn` directly to avoid creating and evaluating symbolic arguments in here.

pymc3/distributions/mixture.py
ricardoV94: But it's an explicit parameter, no? Not a `.dist()` kind of thing (unlike the `comp_dists`):

```python
w = pm.Dirichlet("w", [1, 1, 1])
y = pm.Mixture("y", w, ...)  # w is an explicit parameter
```

twiecki:
@kc611 How close do you think this is to merge? If there's still a ways to go, we might want to focus on the `aeppl` integration instead where we can get `mixture` almost for free. CC @brandonwillard

kc611:
Yeah well, this is still very incomplete I'd say, I initially opened this cause I thought t'll take a long time for `aeppl` to come in. (which is obviously not the case) If everyone's okay with it I'll just go ahead and close this one. We can start with the new Mixture once https://github.com/aesara-devs/aeppl/pull/19 is merged.

PR #4827, opened 2021-07-01
Fix flaky ODE test by setting an initval

michaelosthege:
The test did not fail because of the reason that was written into the `mark.xfail`. Instead it was a numerical problem with the model test point evaluation being unstable on float32 because of randomly drawn `sigma` initial values.


ricardoV94:
In these cases we can also seed the model, that will ensure the initval draws are deterministic.

ricardoV94 (approved):

Issue #4828, opened 2021-07-01
Reenable test_types.py and make it run faster


michaelosthege:
The `test_types.py` test is currently disabled, but this stuff is important.

Also it can probably be accelerated by something like `tune=10, cores=1`.

PR #4830, opened 2021-07-01
Reenable tests that were mostly good already

michaelosthege:
This PR includes more test files in the CI actions.
Only three remain excluded:
+ `test_types.py` where I wasn't familiar enough with what's going on internally (see #4828)
+ `test_mixture.py`
+ `test_variational_inference.py`

I also split up the test jobs based on their runtime. This was before, labeled by the name of the first test in each job.
grafik

The `test_distributions.py` is by far the largest single one. We should definitely split it into two or three files and run them separately!!

### DIFF
--- a/pymc3/tests/test_parallel_sampling.py
+++ b/pymc3/tests/test_parallel_sampling.py
@@ -89,39 +89,51 @@ def test_remote_pipe_closed():
             pm.sample(step=step, mp_ctx="spawn", tune=2, draws=2, cores=2, chains=2)
 
 
+@pytest.mark.xfail(
+    reason="Possibly the same issue described in https://github.com/pymc-devs/pymc3/pull/4701"
+)
 def test_abort():
     with pm.Model() as model:
         a = pm.Normal("a", shape=1)
         pm.HalfNormal("b")
         step1 = pm.NUTS([a])
-        step2 = pm.Metropolis([model.b_log__])
+        step2 = pm.Metropolis([model["b_log__"]])
 
     step = pm.CompoundStep([step1, step2])
 
-    ctx = multiprocessing.get_context()
-    proc = ps.ProcessAdapter(
-        10,
-        10,
-        step,
-        chain=3,
-        seed=1,
-        mp_ctx=ctx,
-        start={"a": 1.0, "b_log__": 2.0},
-        step_method_pickled=None,
-        pickle_backend="pickle",
-    )
-    proc.start()
-    proc.write_next()
-    proc.abort()
-    proc.join()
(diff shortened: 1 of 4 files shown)


ricardoV94:
The test_distributions.py will be significantly faster after https://github.com/pymc-devs/pymc3/pull/4736 

I'll try to split the speedup part into a separate PR

pymc3/tests/test_parallel_sampling.py
```diff
@@ -176,30 +192,36 @@ def func(x):
         pm.sample(draws=10, tune=10, step=pm.Metropolis(), cores=2, mp_ctx="spawn")
 
 
+@pytest.mark.xfail(reason="DensityDist was not yet refactored for v4")
 def test_spawn_densitydist_bound_method():
     with pm.Model() as model:
         mu = pm.Normal("mu", 0, 1)
         normal_dist = pm.Normal.dist(mu, 1)
-        obs = pm.DensityDist("density_dist", normal_dist.logp, observed=np.random.randn(100))
+        logp = lambda x: pm.logp(normal_dist, x, transformed=False)
```
ricardoV94, line 200: `pm.logp` doesn't exist. did you mean `logpt`? In any case, I would leave these DensityDist unchanged, and just add the xfail mark

pymc3/tests/test_profile.py
```diff
@@ -23,7 +23,7 @@ def test_profile_model(self):
         assert self.model.profile(self.model.logpt).fct_call_time > 0
 
     def test_profile_variable(self):
-        assert self.model.profile(self.model.value_vars[0].logpt).fct_call_time > 0
+        assert self.model.profile(self.model.value_vars[0]).fct_call_time > 0
```
ricardoV94: This one does not seem equivalent...

```suggestion
        rv = self.model.basic_RVs[0]
        self.model.profile(pm.logpt(rv, self.model.rvs_to_values[rv])).fct_call_time    
```

michaelosthege: It returns the `TensorVariable` which is then profiled, no?
Can you suggest a change? As you can see I'm still confused by these logpt things..

ricardoV94: I added a suggestion in my first comment, but it has to be tested locally, as I might have screwed up the names while copy pasting from the REPL

ricardoV94 (approved):

ricardoV94:
This failing test seems pretty close to the expected result https://github.com/pymc-devs/pymc3/pull/4830/checks?check_run_id=2971381082#step:7:199

But if you keep finding issues it might be related to #4771

PR #4833, opened 2021-07-02
Add user-friendly logp and logcdf methods

ricardoV94:
This PR adds a `logp` helper similar to the `logcdf` that already existed to facilitate the extraction of `logp` and `logcdf` expressions, similar to how the `.logp` and `.logcdf` methods worked in `v3`:

```python
import pymc3 as pm

# v3:
logp_norm = pm.Normal.dist(0, 1).logp(5.0)
locdf_norm = pm.Normal.dist(0, 1).logcdf(5.0)

# v4:
logp_norm = pm.logp(pm.Normal.dist(0, 1), 5.0)
logcdf_norm = pm.logcdf(pm.Normal.dist(0, 1), 5.0)
```

The file `distributions/logp.py` was renamed to `distributions/logprob.py` to avoid the the name conflict.

### DIFF
--- a/pymc3/distributions/logprob.py
+++ b/pymc3/distributions/logprob.py
@@ -342,9 +342,24 @@ def subtensor_logp(op, var, rvs_to_values, indexed_rv_var, *indices, **kwargs):
     return logp_var
 
 
-def logcdf(*args, **kwargs):
+def logp(var, rv_values, **kwargs):
+    """Create a log-probability graph."""
+
+    # Attach the value_var to the tag of var when it does not have one
+    if not hasattr(var.tag, "value_var"):
+        if isinstance(rv_values, Mapping):
+            value_var = rv_values[var]
+        else:
+            value_var = rv_values
+        var.tag.value_var = at.as_tensor_variable(value_var, dtype=var.dtype)
+
+    return logpt(var, rv_values, **kwargs)
+
+
+def logcdf(var, rv_values, **kwargs):
     """Create a log-CDF graph."""
-    return logpt(*args, cdf=True, **kwargs)
+
+    return logp(var, rv_values, cdf=True, **kwargs)
 
 
 @singledispatch
--- a/pymc3/tests/test_distributions.py
+++ b/pymc3/tests/test_distributions.py
@@ -36,7 +36,7 @@ from scipy.special import erf, logit
 
 import pymc3 as pm
 
-from pymc3.aesaraf import change_rv_size, floatX
+from pymc3.aesaraf import change_rv_size, floatX, intX
 from pymc3.distributions import (
     AR1,
     CAR,
@@ -101,7 +101,7 @@ from pymc3.distributions import (
     ZeroInflatedPoisson,
     continuous,
     logcdf,
-    logpt,
+    logp,
     logpt_sum,
 )
 from pymc3.math import kronecker, logsumexp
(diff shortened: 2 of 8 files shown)


pymc3/distributions/logprob.py
```diff
@@ -342,9 +342,24 @@ def subtensor_logp(op, var, rvs_to_values, indexed_rv_var, *indices, **kwargs):
     return logp_var
 
 
-def logcdf(*args, **kwargs):
+def logp(var, rv_values, **kwargs):
+    """Create a log-probability graph."""
+
+    # Attach the value_var to the tag of var when it does not have one
+    if not hasattr(var.tag, "value_var"):
```
ricardoV94, line 349: This is needed to avoid a warning issued in `aesaraf::rvs_to_value_vars` function.

In main:

```python
import pymc3 as pm
x = pm.Normal.dist(0, 1, size=2)
pm.logcdf(x, [0, 1]).eval()
```
```python
UserWarning: No value variable found for normal_rv{0, (0, 0), floatX, False}.out; the random variable will not be replaced.
```

pymc3/tests/test_distributions_random.py
```diff
@@ -1472,12 +1472,6 @@ def ref_rand(size, mu, sigma, upper):
             pm.TruncatedNormal, {"mu": R, "sigma": Rplusbig, "upper": Rplusbig}, ref_rand=ref_rand
         )
 
-    def test_skew_normal(self):
-        def ref_rand(size, alpha, mu, sigma):
-            return st.skewnorm.rvs(size=size, a=alpha, loc=mu, scale=sigma)
-
-        pymc3_random(pm.SkewNormal, {"mu": R, "sigma": Rplus, "alpha": R}, ref_rand=ref_rand)
-
```
michaelosthege, line 1668: why did you remove it?

michaelosthege (approved):
I like it!

That reduces one breaking `v3`→`v4` change to something less painful:
```python
my_rv = pm.Normal.dist(0, 2)

# on v3:
my_rv.logp(1.5).eval()

# on v4.
pm.logp(my_rv, 1.5).eval()
```

pymc3/tests/test_distributions_random.py
ricardoV94: It's one of those old random tests we have been replacing with the BaseTestDistribution class. Should have been removed when we refactored it but I must have missed it.

Also, I checked that it was passing locally before removing it. It did fail in this PR but it's not surprising since it is a non-seeded statistical test

brandonwillard (commented):
Generally speaking, don't do too much with respect to the likelihood calculations in PyMC, because we're ultimately going to set PyMC up to use `aeppl`.  Since this looks more like a PyMC-specific interface change, that might not be an issue, though.

ricardoV94:
> Generally speaking, don't do too much with respect to the likelihood calculations in PyMC, because we're ultimately going to set PyMC up to use `aeppl`.  Since this looks more like a PyMC-specific interface change, that might not be an issue, though.

Yeah, this one is actually mimicking a bit what we have in aeppl just to extract a logp/logcdf of a vanilla RV

pymc3/tests/test_distributions.py
```diff
@@ -2398,13 +2389,10 @@ def test_batch_dirichlet_multinomial(self):
         dist = DirichletMultinomial.dist(n=n, a=a)
 
         # Logp should be approx -9.98004998e-06
-        value = at.tensor3(dtype="int32")
-        value.tag.test_value = np.zeros_like(vals, dtype="int32")
-        logp = logpt(dist, value)
-        f = aesara.function(inputs=[value], outputs=logp)
-        expected_logp = np.full(shape=f(vals).shape, fill_value=-9.98004998e-06)
+        dist_logp = logp(dist, vals).eval()
+        expected_logp = np.full_like(dist_lopg, fill_value=-9.98004998e-06)
```
michaelosthege:
```suggestion
        expected_logp = np.full_like(dist_logp, fill_value=-9.98004998e-06)
```

michaelosthege (commented):
bad luck

pymc3/tests/test_distributions.py
ricardoV94: Yah :)

michaelosthege (approved):

PR #4836, opened 2021-07-03
Fix for multinomial random test

### LINKED ISSUE #4810: `test_distributions_random` failing due to change in `st.chisquare` behavior
The tests in `test_distributions_random` which rely on `pymc3_random_discrete` started failing in the `v3` branch due to changes in the behavior of `st.chisquare`. I don't think any test is making use of this helper in the `main` branch after the `v4` refactoring, so we are not seeing failures related to that here.

For instance, in the `v3` branch, `test_distributions_random.py::TestScalarParameterSamples::test_poisson` passes locally with scipy `1.6.3` but fails with `1.7.0`

It raises a `ValueError` in `st.chisquare`:

```python
            if diff_gt_tol:
                       f"{relative_diff}")
```
This issue / breaking change was added in https://github.com/scipy/scipy/pull/13193 and mentioned in https://github.com/scipy/scipy/issues/13362 and https://github.com/scipy/scipy/issues/14298

_Originally posted by @ricardoV94 in https://github.com/pymc-devs/pymc3/pull/4808#issuecomment-869192843_

fonnesbeck:
Fixes failure that was triggered by addition of `st.chisquare` denominator test in scipy. I assume we want this merged to v3.11.3.

Closes #4810

### DIFF
--- a/pymc3/__init__.py
+++ b/pymc3/__init__.py
@@ -13,7 +13,7 @@
 #   limitations under the License.
 
 # pylint: disable=wildcard-import
-__version__ = "3.11.2"
+__version__ = "3.11.3"
 
 import logging
 import multiprocessing as mp
--- a/pymc3/tests/test_distributions_random.py
+++ b/pymc3/tests/test_distributions_random.py
@@ -109,9 +109,10 @@ def pymc3_random_discrete(
             e = np.atleast_1d(e).flatten()
             observed = dict(zip(*np.unique(o, return_counts=True)))
             expected = dict(zip(*np.unique(e, return_counts=True)))
-            for e in expected.keys():
-                expected[e] = (observed.get(e, 0), expected[e])
-            k = np.array([v for v in expected.values()])
+            obs_exp = {
+                i: (observed.get(i, 0), expected.get(i, 0)) for i in set().union(observed, expected)
+            }
+            k = np.array([v for v in obs_exp.values()])
             if np.all(k[:, 0] == k[:, 1]):
                 p = 1.0
             else:


michaelosthege:
We _could_ merge into `v3.11.3`, but it'd better to go into `v3`. The `v3.11.3` branch was created just for bumping the version number and shouldn't have been named like that in the first place.

Sayam753:
> The v3.11.3 branch was created just for bumping the version number and shouldn't have been named like that in the first place.

Hi. When working on a release, there is no information about branch names on [Release Checklist](https://github.com/pymc-devs/pymc3/wiki/Release-Checklist). Can you add some info about correct branch names there, to provide more clarity? 

Edit - I realize this is out of context for this PR. But could not resist sharing the ideas. :)

michaelosthege:
> Hi. When working on a release, there is no information about branch names on [Release Checklist](https://github.com/pymc-devs/pymc3/wiki/Release-Checklist). Can you add some info about correct branch names there, to provide more clarity?

Done.

The problem really is that all the core contributors can push new branches to the original repo. They accumulate and in some cases cause confusion because of their naming.
Every now and then we have to go to the [branches](https://github.com/pymc-devs/pymc3/branches) page and delte some.

Sayam753:
> The problem really is that all the core contributors can push new branches to the original repo. They accumulate and in some cases cause confusion because of their naming.

If this has to be avoided, I hope GitHub has some settings to do so. But the more important question is - what are the differences in pushing the changes directly to PyMC3 against developing everything on the fork?

michaelosthege:
Not sure about technical differences, but one thing is that maintainers 2 years from now will wonder what's the matter with these old branches and if they can be deleted. (I for example don't want to delete other peoples work.)
Also users might click the "branches" dropdown and confuse them for official branchs because of how they're named.
It's not super important, and I fear there are no GitHub settings to prevent it, but just something to keep in mind.

fonnesbeck:
We should probably have an official branch naming convention specified in the docs to help mitigate this.

michaelosthege:
@fonnesbeck looks like the pre-commit and the fix is not yetworking. Should we instead mark them as xfail for the time being?

michaelosthege:
Rebased onto `v3` and applied the pre-commit. The tests are still :x:, but now there'll be less noise in the CI logs.

ricardoV94:
These were the tests that were failing before this PR [see here](https://github.com/pymc-devs/pymc3/runs/2929551347#step:8:4124):

```
FAILED pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_beta_binomial
FAILED pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_poisson
FAILED pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_negative_binomial
FAILED pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_geometric
FAILED pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_categorical_random[4]
FAILED pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_dirichlet_multinomial
FAILED pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_multinomial
```

These two are still failing [see here](https://github.com/pymc-devs/pymc3/pull/4836/checks?check_run_id=2983355743#step:7:2832): 
```
FAILED pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_negative_binomial
FAILED pymc3/tests/test_distributions_random.py::TestScalarParameterSamples::test_geometric
```

You seem to have fixed 5/7 of the failing cases!

fonnesbeck:
Interesting: for negative binomial with mu=20 and p=0.5 and for geometric with p=0.001 we never get a p-value that is non-zero.

fonnesbeck:
The rebase messed up the PR on my end. I'm going to create a new one.

PR #4838, opened 2021-07-03
Refactor CI workflow to get codecov reports back

michaelosthege:
Trying to get the codecov comments back.

I'm not optimistic about this first try, but we can copyedit from Aesara: https://github.com/aesara-devs/aesara/blob/main/.github/workflows/test.yml

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,4 +2,3 @@
 omit =
     # exclude tests files from coverage calculation
     pymc3/tests/test_*.py
-    pymc3/examples/*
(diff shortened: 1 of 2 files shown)


michaelosthege:
No idea why the coverage scores are fluctuating around. But how about merging this to see what happens?
And then we can decide to switch to a multi-stage workflow like in the Aesara repo later on.
(I absolutely hate editing YAML and the GitHub action for that Aesara workflow looks terrifying.)

ricardoV94 (approved):

PR #4847, opened 2021-07-08
Adds quadratic approximation

rasmusbergpalm:
The quadratic approximation is an extremely fast way to obtain an estimate of the posterior. It only works if the posterior is unimodal and approximately symmetric.

The implementation finds the quadratic approximation and then samples from it to return an arviz.InferenceData object. Finding the exact (approximate) posterior and then sampling from it might seem counter intuitive, but it's done to be compatible with the rest of the codebase and Arviz functionality. The exact (approximate) posterior is also returned as a scipy.stats.multivariate_normal distribution.

### DIFF
--- a/pymc3/quadratic_approximation.py
+++ b/pymc3/quadratic_approximation.py
@@ -26,7 +26,7 @@ from pymc3.tuning import find_MAP, find_hessian
 
 
 def quadratic_approximation(vars, n_chains=2, n_samples=10_000):
-    """ Finds the quadratic approximation to the posterior, also known as the laplace approximation.
+    """ Finds the quadratic approximation to the posterior, also known as the Laplace approximation.
 
     NOTE: The quadratic approximation only works well for unimodal and roughly symmetrical posteriors of continuous variables.
     The usual MCMC convergence and mixing statistics (e.g. R-hat, ESS) will NOT tell you anything about how well this approximation fits your actual (unknown) posterior, indeed they'll always be extremely nice since all "chains" are sampling from exactly the same distribution, the posterior quadratic approximation.
--- a/pymc3/quadratic_approximation.py
+++ b/pymc3/quadratic_approximation.py
@@ -1,0 +1,68 @@
+#   Copyright 2020 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""Functions for Quadratic Approximation."""
+
+import numpy as np
+import scipy
+import arviz as az
+
+__all__ = [
+    "quadratic_approximation"
+]
+
+from pymc3.tuning import find_MAP, find_hessian
+
+
+def quadratic_approximation(vars, n_chains=2, n_samples=10_000):
+    """ Finds the quadratic approximation to the posterior, also known as the laplace approximation.
+
+    NOTE: The quadratic approximation only works well for unimodal and roughly symmetrical posteriors of continuous variables.
+    The usual MCMC convergence and mixing statistics (e.g. R-hat, ESS) will NOT tell you anything about how well this approximation fits your actual (unknown) posterior, indeed they'll always be extremely nice since all "chains" are sampling from exactly the same distribution, the posterior quadratic approximation.
+    Use at your own risk.
+
+    See Chapter 4 of "Bayesian Data Analysis" 3rd edition for background.
+
+    Returns an arviz.InferenceData object for compatibility by sampling from the approximated quadratic posterior. Note these are NOT MCMC samples, so the notion of chains is meaningless, and is only included for downstream compatibility with Arviz.
+
+    Also returns the exact posterior approximation as a scipy.stats.multivariate_normal distribution.
+
+    Parameters
+    ----------
+    vars: list
+        List of variables to approximate the posterior for.
+    n_chains: int
+        How many chains to simulate.
+    n_samples: int
+        How many samples to sample from the approximate posterior for each chain.
+
+    Returns
+    -------
+    (arviz.InferenceData, scipy.stats.multivariate_normal):
+        InferenceData with samples from the approximate posterior, multivariate normal posterior approximation
+
+    """
+    map = find_MAP(vars=vars)
+    H = find_hessian(map, vars=vars)
+    cov = np.linalg.inv(H)
+    mean = np.concatenate([np.atleast_1d(map[v.name]) for v in vars])
+    posterior = scipy.stats.multivariate_normal(mean=mean, cov=cov)
+    draws = posterior.rvs((n_chains, n_samples))
+    samples = {}
+    i = 0
+    for v in vars:
+        var_size = map[v.name].size
+        samples[v.name] = draws[:, :, i:i + var_size].squeeze()
+        i += var_size
+    return az.convert_to_inference_data(samples), posterior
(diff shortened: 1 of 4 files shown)


pymc3/quadratic_approximation.py
```diff
@@ -0,0 +1,68 @@
+#   Copyright 2020 The PyMC Developers
```
twiecki:
```suggestion
#   Copyright 2021 The PyMC Developers
```

pymc3/quadratic_approximation.py
```diff
@@ -0,0 +1,68 @@
...
+
+__all__ = [
+    "quadratic_approximation"
+]
+
+from pymc3.tuning import find_MAP, find_hessian
+
+
+def quadratic_approximation(vars, n_chains=2, n_samples=10_000):
+    """ Finds the quadratic approximation to the posterior, also known as the laplace approximation.
```
twiecki:
```suggestion
    """ Finds the quadratic approximation to the posterior, also known as the Laplace approximation.
```

pymc3/quadratic_approximation.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Returns an arviz.InferenceData object for compatibility by sampling from the approximated quadratic posterior. Note these are NOT MCMC samples, so the notion of chains is meaningless, and is only included for downstream compatibility with Arviz.
+
+    Also returns the exact posterior approximation as a scipy.stats.multivariate_normal distribution.
+
+    Parameters
+    ----------
+    vars: list
+        List of variables to approximate the posterior for.
+    n_chains: int
```
twiecki: I don't think we need this kwarg and should fix it to 1.

pymc3/quadratic_approximation.py
```diff
@@ -0,0 +1,68 @@
...
+        How many chains to simulate.
+    n_samples: int
+        How many samples to sample from the approximate posterior for each chain.
+
+    Returns
+    -------
+    (arviz.InferenceData, scipy.stats.multivariate_normal):
```
twiecki: Also list these items individually, you can look at other doc strings for examples for multiple return elements.

twiecki:
This looks pretty good. Needs a line in the release notes. I wonder if we should make this available through `sample`?

Really though this probably fits into the `VI` submodule, but that one is a bit more complex. Curious what @ferrine thinks.

ricardoV94:
I suggest adding a test with multivariate variables and / or variables with different dimensions (larger than 1 dimension). Not sure how well that mean concatenation will work in those cases.

ricardoV94:
> This looks pretty good. Needs a line in the release notes. I wonder if we should make this available through `sample`?

I don't think that would be the best.

But we could perhaps add a general `approx` module where we put stuff like this and MAP (which doesn't make much sense as "tuning" anymore).

I am planning to add a "grid approximation" function that works a lot like this quadratic in spirit (deterministic evaluation + samples to returns an inferencedata), and would place it there as well.

pymc3/quadratic_approximation.py
```diff
@@ -0,0 +1,68 @@
...
+        InferenceData with samples from the approximate posterior, multivariate normal posterior approximation
+
+    """
+    map = find_MAP(vars=vars)
```
ricardoV94, line 53: Are there any arguments to `find_MAP` that we might want to allow the user to pass? I.e. `map = find_MAP(vars=vars, **map_kwargs)`? Same for the `Hessian`. If not, that's fine (even better)

rasmusbergpalm:
> I suggest adding a test with multivariate variables and / or variables with different dimensions (larger than 1 dimension). Not sure how well that mean concatenation will work in those cases.

That's why I made `test_hdi_contains_parameters_in_linear_regression` which has weights with shape=2 :)

pymc3/quadratic_approximation.py
rasmusbergpalm: Unfortunately, Arviz will complain if there aren't at least 2 chains.

pymc3/quadratic_approximation.py
rasmusbergpalm: I couldn't find any examples. I tried a new format. Please provide a reference if it's not correct :)

pymc3/quadratic_approximation.py
rasmusbergpalm: Hmm. I think I'd say no. It's a leaky abstraction. If someone raises an issue where they need to pass args to find_MAP we can think about how to best do it at that time. That's my 2 cents.

rasmusbergpalm:
I've adressed the PR comments as best as I could, ran linting and added a line in release notes. Please take another look @twiecki 

I will add an example notebook once merged.

rasmusbergpalm:
> > This looks pretty good. Needs a line in the release notes. I wonder if we should make this available through `sample`?
> 
> I don't think that would be the best.
> 
> But we could perhaps add a general `approx` module where we put stuff like this and MAP (which doesn't make much sense as "tuning" anymore).
> 
> I am planning to add a "grid approximation" function that works a lot like this quadratic in spirit (deterministic evaluation + samples to returns an inferencedata), and would place it there as well.

Agreed it doesn't fit well in sample. A humble suggestion might be to make an inference module and place the various types of posterior inference there, e.g. MCMC, VI, this, grid approx., etc.

fonnesbeck:
I like the `approx` submodule idea best. If we ever get around to adding INLA it could live there as well.

pymc3/quadratic_approximation.py
ColCarroll: can you share an example? I think you'll get trouble for multivariate distributions, where arviz will interpret the leading dimension as the chains, but I would solve that as follows:

```python
data = {'a': np.ones(10), 'b': np.ones((100, 3))}

az.convert_to_inference_data(data) # warning, misinterprets `b` as having 100 chains

az.convert_to_inference_data({k: v[np.newaxis, ...] for k, v in data.items()}) # Good, explicitly sets number of chains to 1.
```

But maybe I'm misunderstanding the issue!

I'm against including chains since it gives the impression that you might want to run diagnostics on the quality of the returned samples.

pymc3/quadratic_approximation.py
```diff
@@ -0,0 +1,68 @@
...
+    H = find_hessian(map, vars=vars)
+    cov = np.linalg.inv(H)
+    mean = np.concatenate([np.atleast_1d(map[v.name]) for v in vars])
+    posterior = scipy.stats.multivariate_normal(mean=mean, cov=cov)
```
ColCarroll, line 57: Does it make sense to just return this scipy object (or `pm.MvNormal.dist(mean, cov)`)? I think it has more facility than samples from it.

ColCarroll:
(sorry, submitted a review late, but excited to see this!)

pymc3/quadratic_approximation.py
ricardoV94: Not in V4. It would be just an Aesara random variable

ColCarroll: (missed that posterior was the second return object, carry on!)

pymc3/quadratic_approximation.py
rasmusbergpalm: image

I guess that's not too bad. All the plots seem to work. I just saw that warning and figured nothing would work. I'll remove the chains. I fully agree. The less impression we can give that running diagnostics the better. A little warning and a NaN might actually be good :)

rasmusbergpalm:
@ColCarroll I removed the chains. PTAL :)

rasmusbergpalm:
I've spent the entire day trying to reproduce the test errors and have failed. I simply cannot create a working development environment. 

First I spent countless hours fixing C import `stdio.h` errors, upgrading macOS, installing xcode, etc.
Then I spent a lot of time with "Couldn't find Library System" when compiling Aesara, following the contributing.MD installation guide (using conda)
Then I tried using a venv and pip which is the closest I've got which gave me 
```
>           _lbfgsb.setulb(m, x, low_bnd, upper_bnd, nbd, f, g, factr,
                           pgtol, wa, iwa, task, iprint, csave, lsave,
                           isave, dsave, maxls)
E           ValueError: unexpected array size: new_size=2, got array with arr_size=0
```
Some bug deep in the FORTRAN code of LBFGS, which is not the error in CI.
Then I tried the docker approach, but running `. ./../../scripts/test.sh` gives `unknown command pytest`.
Installing pytest and re-running gives `E   ModuleNotFoundError: No module named 'aesara'`.

At this point I've given up. The code works, as you can see here https://colab.research.google.com/drive/1DTe7QchyW-wpbUmulzY27lBTRFQXzhm0?usp=sharing (at least in PyMC3 3.11.2)

twiecki:
Closing due to inactivity, feel free to reopen.

Issue #4845, opened 2021-07-08
MatrixNormal complains about 1x1 matrix even when multidimensional means are provided


Spaak:
The following worked fine in v3:

```python
n = 3
with Model():
    matrix_normal = MatrixNormal(
        "mat_normal",
        mu=np.random.normal(size=n),
        rowcov=np.eye(n),
        colchol=np.linalg.cholesky(np.eye(n)),
        size=(n, n),
    )
```

but now gives the following:

```
Traceback (most recent call last):
  File "/home/eelke/python/playground_pymc3_v4_strings.py", line 260, in <module>
    matrix_normal = MatrixNormal(
  File "/home/eelke/repos/pymc3/pymc3/distributions/distribution.py", line 208, in __new__
    rv_out = cls.dist(*args, rng=rng, initval=None, **kwargs)
  File "/home/eelke/repos/pymc3/pymc3/distributions/multivariate.py", line 1661, in dist
    raise ValueError(
ValueError: 1x1 Matrix was provided. Please use Normal distribution for such cases.
```

I ran into this when updating the tests in `test_distributions.py::TestStrAndLatexRepr`. I think either the check here is wrong: https://github.com/pymc-devs/pymc3/blob/7e464e59bcb0adb28df94f379b3e8d4af12bd4d1/pymc3/distributions/multivariate.py#L1656-L1659

or perhaps the input shape of `mu` is more constrained now in v4 than it was in v3 (when a 1d vector was fine). In any case the check for `mu.ndim == 1` is not sufficient to conclude that a 1x1 matrix was passed; that can also hold for a 1d vector.

## Versions and main components

* PyMC3 Version: `main` (at a061106c4555837de64c197d7217dfadfa4bfbbe)
* Aesara/Theano Version: aesara-devs/aesara@89cfea013a6963ef28c5f18640d2972bda59e3ae (HEAD, tag: rel-2.0.12)
* Python Version: 3.8.2
* Operating system: Ubuntu 20.04.2 via WSL2 (Win10)
* How did you install PyMC3: git

ricardoV94:
@kc611 did the refactoring for V4, any insights?

kc611:
Ah yes, the check's definitely wrong in such cases. I also tried removing it, doesn't work, seems like internally `MatrixNormal` doesn't like being given dimensions other than `(m x n)` types. 

[...]

PR #4848, opened 2021-07-08
Fixed MatrixNormal shape handling

kc611:
Linked Issue: #4845 

So it turns out that the shape handling was a bit off in `MatrixNormal` as it didn't support `mu` broadcasting, also it'll be better to just remove the `1x1` error since it won't make much sense if we have it supported already.

### DIFF
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -1525,6 +1525,10 @@ class MatrixNormalRV(RandomVariable):
     dtype = "floatX"
     _print_name = ("MatrixNormal", "\\operatorname{MatrixNormal}")
 
+    def _infer_shape(self, size, dist_params, param_shapes=None):
+        shape = tuple(size) + tuple(dist_params[0].shape)
+        return shape
+
     @classmethod
     def rng_fn(cls, rng, mu, rowchol, colchol, size=None):
 
@@ -1596,7 +1600,7 @@ class MatrixNormal(Continuous):
         n = colcov.shape[0]
         mu = np.zeros((m, n))
         vals = pm.MatrixNormal('vals', mu=mu, colcov=colcov,
-                               rowcov=rowcov, shape=(m, n))
+                               rowcov=rowcov)
 
     Above, the ith row in vals has a variance that is scaled by 4^i.
     Alternatively, row or column cholesky matrices could be substituted for
@@ -1634,7 +1638,7 @@ class MatrixNormal(Continuous):
             rowcov = at.diag([scale**(2*i) for i in range(m)])
 
             vals = pm.MatrixNormal('vals', mu=mu, colchol=colchol, rowcov=rowcov,
-                                   observed=data, shape=(m, n))
+                                   observed=data)
     """
     rv_op = matrixnormal
 
@@ -1646,16 +1650,22 @@ class MatrixNormal(Continuous):
         rowchol=None,
         colcov=None,
         colchol=None,
-        shape=None,
         *args,
         **kwargs,
     ):
 
         cholesky = Cholesky(lower=True, on_error="raise")
 
-        if mu.ndim == 1:
-            raise ValueError(
-                "1x1 Matrix was provided. Please use Normal distribution for such cases."
+        if kwargs.get("size", None) is not None:
+            raise NotImplementedError("MatrixNormal doesn't support size argument")
+
+        if "shape" in kwargs:
+            kwargs.pop("shape")
+            warnings.warn(
+                "Passing shape to MatrixNormal is now depricated."
+                "MatrixNormal automatically derives the shape"
+                "from row and column matrix dimensions.",
+                DeprecationWarning,
             )
 
         # Among-row matrices
@@ -1687,6 +1697,11 @@ class MatrixNormal(Continuous):
                 raise ValueError("colchol must be two dimensional.")
             colchol_cov = at.as_tensor_variable(colchol)
 
+        dist_shape = (rowchol_cov.shape[0], colchol_cov.shape[0])
+
+        # Broadcasting mu
+        mu = at.extra_ops.broadcast_to(a=mu, shape=dist_shape)
+
         mu = at.as_tensor_variable(floatX(mu))
         # mean = median = mode = mu
 
@@ -1706,6 +1721,9 @@ class MatrixNormal(Continuous):
         -------
         TensorVariable
         """
+
+        if value.ndim != 2:
+            raise ValueError("Value must be two dimensional.")
 
         # Compute Tr[colcov^-1 @ (x - mu).T @ rowcov^-1 @ (x - mu)] and
         # the logdet of colcov and rowcov.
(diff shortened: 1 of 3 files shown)


kc611:
So the reason I've drafted this is :

The `logp` for `MatrixNormal` can only support a `value` of `ndim`s 2 as of now (see failing tests as they try to set an initial `value` with a 4 dimensional variable in `logp`), do we want it to support higher `ndim`s or should we restrict it similar to how `KroneckerNormal` does.

https://github.com/pymc-devs/pymc3/blob/7e464e59bcb0adb28df94f379b3e8d4af12bd4d1/pymc3/distributions/multivariate.py#L1892-L1893

 Though I don't know how this restriction works or wether it's even supposed to be there. (It was present in `v3`'s `KroneckerNormal`)

michaelosthege:
If a restriction is associated with a clear error message asking for contributions, I'd go with that rather than to invest disproportionate much dev time in implementing a feature without knowing if there's even any demand for it.

But the changes you made look rather simple - could it be that the test is just failing because of a simple mixup?

kc611:
I think in this case it'll be better to give it at-least 1 extra dimensional support. Cause similar cases are observed in `MvNormal` (and on further introspection `KroneckerNormal` actually does the same). They actually perform operations on 1-D values and provide support for a extra `size` dimension. In `MatrixNormal` we have 2-dimensional values so with only 2-dim support `MatrixNormal` won't work for any `size` we pass. (And that is why specifically the tests with `size` argument fail.) 

That said, is it possible to somehow 'iterate' over an axis of `TensorVariable` (`value` in this instance) to apply a given set of operations on last two dimensions ?

ricardoV94:
> That said, is it possible to somehow 'iterate' over an axis of `TensorVariable` (`value` in this instance) to apply a given set of operations on last two dimensions ? 

I don't know what exactly you have in mind but axis can be tuples such as `np.sum(arr, axis=(-2, -1))` 

Or `numpy.apply_over_axes`, but that might not exist in aesara (yet)

kc611:
The exact thing I wanted to do is basically apply the operations in the current`MatrixNormal` logp method (which contain some `Op`s explicitly designed for 2-dimensional(matrix) kind of values ) over an third dimensional axis (which will represent the `size`) parameter. So it'll be like iteration in which we apply the current operations in all values over the `size` axis.

The `MvNormal` (whose values are one dimension lower than that of `MatrixNormal`) does a similar thing in its logp by having that `onedim` variable which allows it to process one dimension higher. (The size dimension) The problem here is, unlike `MatrixNormal` we can't just add in an extra dimension (or can we?) , since the `Op`s here explicitly check for 2 dimensional values and I'm pretty sure they do it for good reasons. 

Basically I want the `TensorVariable` equivalent of:

```
for _value in value:
      do_the_thing(_value)
```

pymc3/distributions/multivariate.py
```diff
@@ -1646,16 +1650,21 @@ def dist(
...
         **kwargs,
     ):
 
         cholesky = Cholesky(lower=True, on_error="raise")
 
-        if mu.ndim == 1:
-            raise ValueError(
-                "1x1 Matrix was provided. Please use Normal distribution for such cases."
+        if "size" in kwargs:
+            raise NotImplementedError("MatrixNormal doesn't support size argument")
```
kc611, line 1660: Turns out `size` was never supported in the first place in `MatrixNormal` in `v3`. (Thanks to @ricardoV94 for pointing that out.)

ricardoV94: Users may also pass shape (and invalid shapes) or "dims", so this check might not be enough. Can a ndim check be done inside the logp method instead? And what about the RV, does it work properly for non-default sizes?

ricardoV94: Also just to be clear 2d shapes (rows x cols) were allowed, but not "replication" shapes. I am not sure if that maps to any non-`None` size because I don't really grasp what size means for multivariate distributions.

kc611: I added a warning for `shape` since if only `(m x n)` shape is allowed we can directly derive `m` and `n` from parameters. 

> And what about the RV, does it work properly for non-default sizes?

What do you mean by non-default sizes?

ricardoV94: If you used the RV directly with some size like (2, 3, 3), would it do a sensible thing in the `rng_fn`? Or should it also raise there?

kc611: The RV can handle sizes like that because the main operation over there i.e. `np.matmul` will treat the random output (from `rng.standard_normal` ) as stack of matrices (which is what we intend) so I don't think we need to raise there.

ricardoV94: I think size is one of the default arguments for distributions. Perhaps it's better to simply check it is `None`

ricardoV94:
So the issue in #4845 was just the wrong specification of `size`?

We need tests that cover those new warnings / errors.

pymc3/distributions/multivariate.py
```diff
@@ -1646,16 +1650,21 @@ def dist(
...
         *args,
         **kwargs,
     ):
 
         cholesky = Cholesky(lower=True, on_error="raise")
 
-        if mu.ndim == 1:
-            raise ValueError(
-                "1x1 Matrix was provided. Please use Normal distribution for such cases."
+        if "size" in kwargs and kwargs["size"] is not None:
```
```suggestion
        if kwargs.get("size", None) is not None:
```

pymc3/distributions/multivariate.py
```diff
@@ -1707,6 +1721,9 @@ def logp(value, mu, rowchol, colchol):
         TensorVariable
         """
 
+        if value.ndim != 2:
+            raise ValueError("Value must be two dimensional.")
```
ricardoV94, line 1726: Need a test that hits this branch

pymc3/distributions/multivariate.py
```diff
@@ -1646,16 +1650,21 @@ def dist(
...
 
         cholesky = Cholesky(lower=True, on_error="raise")
 
-        if mu.ndim == 1:
-            raise ValueError(
-                "1x1 Matrix was provided. Please use Normal distribution for such cases."
+        if "size" in kwargs and kwargs["size"] is not None:
+            raise NotImplementedError("MatrixNormal doesn't support size argument")
+
+        if "shape" in kwargs:
```
ricardoV94, line 1662: Should add a test for this warning. And it should be a DeprecationWarning

pymc3/distributions/multivariate.py
```diff
@@ -1646,16 +1650,22 @@ def dist(
...
-        if mu.ndim == 1:
-            raise ValueError(
-                "1x1 Matrix was provided. Please use Normal distribution for such cases."
+        if kwargs.get("size", None) is not None:
+            raise NotImplementedError("MatrixNormal doesn't support size argument")
+
+        if "shape" in kwargs:
+            kwargs.pop("shape")
+            warnings.warn(
+                "Passing shape to MatrixNormal is now depricated."
```
```suggestion
                "The shape argument in MatrixNormal is deprecated and will be ignored."
```

ricardoV94 (approved):

michaelosthege:
Wow, approved 20 days ago. Can we merge this?

kc611:
Yeah, the work here is done as far as I know.

PR #4857, opened 2021-07-12
Replace `incomplete_beta` with `at.betainc` and speedup/clean logcdf tests

### LINKED ISSUE #4420: dist_math.incomplete_beta is very slow

### LINKED ISSUE #4467: Remove hacks to avoid previous aesara gammainc(c) limitations

ricardoV94:
# Implement Betainc

This PR deprecates the `incomplete_beta` method in `dist_math.py` in favor of the new `aesara.tensor.betainc` which wraps the `scipy.special.betainc` method. Supersedes #4736; Closes #4420.

This change provides a large increase in speed (and graph compilation) for distributions that made use of the `incomplete_beta` and/or it's gradients. In addition, the logcdf of Beta, StudentT, Binomial, NegativeBinomial (and zero inflated variants) can now evaluate multiple values. 

The downside is the constant warning that the op has no `c-implementation` whenever the logp/logcdf makes use of it. This limitation is tracked in https://github.com/aesara-devs/aesara/issues/267 and https://github.com/aesara-devs/aesara/issues/514 

# Speedup logcdf tests

The logcdf tests were refactored to avoid recreating the logcdf graph for every parameter / value evaluation. This supersedes #4734. The old `incomplete_beta` was found to be defective when attempting #4734, and so I decided to merge the two PRs into this one instead.

This allows to remove most of the custom `n_samples` limitations, and makes the test suite run considerably faster.

# Other changes
1. The work-around for the `HyperGeometric` tests caused by https://github.com/aesara-devs/aesara/issues/461 was removed.
1. The work-around in the `Gamma` and `InverseGamma` logcdf methods was removed (closes #4467)


ricardoV94:
Comparing these two workflows for the `test_distributions.py`: 
1. [Other PR](https://github.com/pymc-devs/pymc3/runs/3042805582?check_suite_focus=true) - 48m
2. [This PR](https://github.com/pymc-devs/pymc3/runs/3045311114?check_suite_focus=true) - 15m

RELEASE-NOTES.md
```diff
@@ -3,12 +3,13 @@
 ## PyMC3 vNext (4.0.0)
 ### Breaking Changes
 - ⚠ Theano-PyMC has been replaced with Aesara, so all external references to `theano`, `tt`, and `pymc3.theanof` need to be replaced with `aesara`, `at`, and `pymc3.aesaraf` (see [4471](https://github.com/pymc-devs/pymc3/pull/4471)).
+- ⚠ PyMC3 now requires Scipy version `>= 1.4.1` (see [4736](https://github.com/pymc-devs/pymc3/pull/4736)).
```
ricardoV94, line 6:
```suggestion
- ⚠ PyMC3 now requires Scipy version `>= 1.4.1` (see [4736](https://github.com/pymc-devs/pymc3/pull/4857)).
```

pymc3/distributions/continuous.py
```diff
@@ -1246,23 +1245,19 @@ def logcdf(value, alpha, beta):
 
         Parameters
         ----------
-        value: numeric
-            Value(s) for which log CDF is calculated.
+        value: numeric or np.ndarray or aesara.tensor
+            Value(s) for which log CDF is calculated. If the log CDF for multiple
+            values are desired the values must be provided in a numpy array or Aesara tensor.
```
brandonwillard, line 1250: Might as well keep this parameter description simple and put all the type-based information in type annotations.

pymc3/tests/test_distributions.py
```diff
@@ -751,25 +738,30 @@ def check_logcdf(
...
                 value = params.pop("value")
-                with Model() as m:
-                    dist = pymc3_dist("y", **params)
+                # Update shared parameter variables in pymc3_logcdf function
+                for param_name, param_value in params.items():
+                    param_vars[param_name].set_value(param_value)
+                pymc3_eval = pymc3_logcdf({"value": value})
+
                 params["value"] = value  # for displaying in err_msg
-                with aesara.config.change_flags(on_opt_error="raise", mode=Mode("py")):
```
brandonwillard, line 766: Are we setting `on_opt_error="raise"` elsewhere?  This is a good option to have set globally.

Also, are we sure we want to compile all these tests via C?  The Python-based graphs can sometimes be about as quick as the C-compiled graphs + their compilation time.

brandonwillard (commented):
Minor comments; otherwise, looks fine&mdash;although it is missing coverage for the deprecation warning.

pymc3/distributions/continuous.py
ricardoV94: Yeah, these docstrings could benefit from an overhaul (or be removed altogether). I simply changed them to be equal to those of the remaining distributions.

pymc3/tests/test_distributions.py
ricardoV94: > Are we setting on_opt_error="raise" elsewhere? This is a good option to have set globally.

I don't think we are. You mean set it up for all the tests?

> Also, are we sure we want to compile all these tests via C? The Python-based graphs can sometimes be about as quick as the C-compiled graphs + their compilation time.

I would say it's better to test them with the default back-end, which for the time being is `C`

ricardoV94:
> Minor comments; otherwise, looks fine—although it is missing coverage for the deprecation warning.

Added a test to cover the deprecation warning

pymc3/distributions/continuous.py
ricardoV94: Opened an issue here: https://github.com/pymc-devs/pymc3/issues/4859

michaelosthege (changes requested):
I agree we should have coverage on the deprecation warnings.

And the minimum Aesara version also needs to be changed in `/conda-envs/*.yml`.

ricardoV94:
> I agree we should have coverage on the deprecation warnings.

I already added that in my last force-push.

> And the minimum Aesara version also needs to be changed in `/conda-envs/*.yml`.

It doesn't use the `requirements.txt`?

michaelosthege:
> It doesn't use the `requirements.txt`?

There are Aesara lines in there and those env files are used for the test pipelines..

ricardoV94:
Changed the conda-envs

michaelosthege (approved):

Issue #4856, opened 2021-07-12
Switch from dill to cloudpickle


twiecki:
Other major packages (like airflow) are moving to cloudpickle: https://github.com/apache/airflow/issues/7870

It also seems that cloudpickle as some incompatibilities with dill (as per @michaelosthege) during installation.

Spaak:
So far the default serialization method is `pickle`, with an optional `pickle_backend` parameter to `pm.sample` which can be set to `"dill"` (required in certain cases, e.g. custom `DensityDist`). Do we want to keep this as a customizable parameter? Or should we always use `cloudpickle`? (I think at the very least we want to set `cloudpickle` to be the default if we make the effort of switching.) I think simply always using `cloudpickle` (and removing the `pickle_backend` parameter) would be fine, I don't see any scenario when someone would want to explicitly fall back to using `pickle`. (But me not seeing it is not enough to conclude there cannot be one ;) )

twiecki:
Agree with your logic, seems like we can just use that as a default.

Spaak:
Closed via #4858

PR #4858, opened 2021-07-13
switch from pickle/dill to cloudpickle

Spaak:
Fixing #4856 and (hopefully) preventing future issues with serialization, parallel sampling, etc. This PR also removes the `pickle_backend` parameter from `pm.sample`.

@twiecki @michaelosthege

### DIFF
--- a/pymc3/tests/test_minibatches.py
+++ b/pymc3/tests/test_minibatches.py
@@ -13,9 +13,9 @@
 #   limitations under the License.
 
 import itertools
-import pickle
 
 import aesara
+import cloudpickle
 import numpy as np
 import pytest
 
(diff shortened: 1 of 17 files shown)


pymc3/tests/test_minibatches.py
```diff
@@ -132,10 +132,10 @@ def gen():
 
     def test_pickling(self, datagen):
         gen = generator(datagen)
-        pickle.loads(pickle.dumps(gen))
+        cloudpickle.loads(cloudpickle.dumps(gen))
         bad_gen = generator(integers())
-        with pytest.raises(Exception):
-            pickle.dumps(bad_gen)
+        with pytest.raises(TypeError):
```
michaelosthege, line 137: Maybe add a `match="..."`?

Spaak: I had that actually (in [this commit](https://github.com/pymc-devs/pymc3/commit/4740bb21a08e24c6b1275959229ffaf1f4a0908a#diff-bb64d53388ed0aaa4fc7eae24a0b54708b78ec32e547ebfb7c5bf36b5646f68eR137)), but had to take it out because apparently the error message is different on different platforms (specifically: "cannot pickle" versus "can't pickle" 🤦). Didn't want to fail on that, obviously.

michaelosthege: Ah okay. And according to the CI logs it's _very_ different.
Then let's go without it.

michaelosthege (approved):

twiecki:
Great stuff!

PR #4860, opened 2021-07-13
Remove custom math functions

### LINKED ISSUE #4747: Remove explicit optimization from `math.logsumexp`
Aesara already performs one pass of this optimization (which will be exactly equivalent to the pymc3 one after https://github.com/pymc-devs/aesara/pull/465). As things stand, the final graph contains two passes of the same optimization, because the optimized output still contains the same form as the original unoptimized graph `log(sum(exp(x))) -> x_max + << log(sum(exp(x - x_max))) >> `

https://github.com/pymc-devs/pymc3/blob/c62100c99df9a74194db89863776f155a95c076a/pymc3/math.py#L189-L194

```python
import aesara
import aesara.tensor as at
from pymc3.math import logsumexp

x = at.vector('x')
y = logsumexp(x)
f = aesara.function([x], y)
aesara.dprint(f)
Elemwise{Composite{(i0 + log(i1) + i2)}}[(0, 0)] [id A] ''   9
 |InplaceDimShuffle{x} [id B] ''   5
 | |Reduce{maximum}{0} [id C] 'max'   4
 |   |Elemwise{sub,no_inplace} [id D] ''   3
 |     |x [id E]
 |     |Elemwise{Composite{Switch(IsInf(i0), i1, i0)}}[(0, 0)] [id F] ''   2
 |       |InplaceDimShuffle{x} [id G] ''   1
 |       | |Reduce{maximum}{0} [id H] 'max'   0
 |       |   |x [id E]
 |       |TensorConstant{(1,) of 0} [id I]
 |InplaceDimShuffle{x} [id J] ''   8
 | |Sum{acc_dtype=float64} [id K] ''   7
 |   |Elemwise{Composite{exp((i0 - i1))}}[(0, 0)] [id L] ''   6
 |     |Elemwise{sub,no_inplace} [id D] ''   3
 |     |InplaceDimShuffle{x} [id B] ''   5
 |Elemwise{Composite{Switch(IsInf(i0), i1, i0)}}[(0, 0)] [id F] ''   2
```
```python
y = at.log(at.sum(at.exp(x), axis=None, keepdims=True))
f = aesara.function([x], y)
aesara.dprint(f)
Elemwise{Composite{(i0 + log(i1))}}[(0, 0)] [id A] ''   5
 |InplaceDimShuffle{x} [id B] ''   1
 | |Reduce{maximum}{0} [id C] 'max'   0
 |   |x [id D]
 |InplaceDimShuffle{x} [id E] ''   4
   |Sum{acc_dtype=float64} [id F] ''   3
     |Elemwise{Composite{exp((i0 - i1))}} [id G] ''   2
       |x [id D]
       |InplaceDimShuffle{x} [id B] ''   1
```

ricardoV94:
This PR removes the custom `logaddexp`, `logsumexp`, and `log1pexp` functions and deprecates `log1mexp` in favor of Aesara implementations.

A deprecation warning is added for `log1mexp` (and the numpy counterpart) because the Aesara function, more logically, expects a negative input.

Also fixed a bug in `log1mexp_numpy` when the input was an integer.

Closes #4747

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -30,6 +30,7 @@
 - The length of `dims` in the model is now tracked symbolically through `Model.dim_lengths` (see [#4625](https://github.com/pymc-devs/pymc3/pull/4625)).
 - We now include `cloudpickle` as a required dependency, and no longer depend on `dill` (see [#4858](https://github.com/pymc-devs/pymc3/pull/4858)).
 - The `incomplete_beta` function in `pymc3.distributions.dist_math` was replaced by `aesara.tensor.betainc` (see [4857](https://github.com/pymc-devs/pymc3/pull/4857)).
+- `math.log1mexp` and `math.log1mexp_numpy` will expect negative inputs in the future. A `FutureWarning` is now raised unless `negative_input=True` is set (see [#4860](https://github.com/pymc-devs/pymc3/pull/4860)).
 - ...
 
 ## PyMC3 3.11.2 (14 March 2021)
--- a/pymc3/math.py
+++ b/pymc3/math.py
@@ -13,6 +13,7 @@
 #   limitations under the License.
 
 import sys
+import warnings
 
 from functools import partial, reduce
 
@@ -50,6 +51,9 @@ from aesara.tensor import (
     gt,
     le,
     log,
+    log1pexp,
+    logaddexp,
+    logsumexp,
     lt,
     maximum,
     minimum,
@@ -186,27 +190,14 @@ def tround(*args, **kwargs):
     return at.round(*args, **kwargs)
 
 
-def logsumexp(x, axis=None, keepdims=True):
-    # Adapted from https://github.com/Theano/Theano/issues/1563
-    x_max = at.max(x, axis=axis, keepdims=True)
-    x_max = at.switch(at.isinf(x_max), 0, x_max)
-    res = at.log(at.sum(at.exp(x - x_max), axis=axis, keepdims=True)) + x_max
-    return res if keepdims else res.squeeze()
-
-
-def logaddexp(a, b):
-    diff = b - a
-    return at.switch(diff > 0, b + at.log1p(at.exp(-diff)), a + at.log1p(at.exp(diff)))
-
-
 def logdiffexp(a, b):
     """log(exp(a) - exp(b))"""
-    return a + log1mexp(a - b)
+    return a + at.log1mexp(b - a)
 
 
 def logdiffexp_numpy(a, b):
     """log(exp(a) - exp(b))"""
-    return a + log1mexp_numpy(a - b)
+    return a + log1mexp_numpy(b - a, negative_input=True)
 
 
 def invlogit(x, eps=sys.float_info.epsilon):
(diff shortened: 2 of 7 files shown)


RELEASE-NOTES.md
```diff
@@ -26,6 +26,8 @@
 - Logp method of `Uniform` and `DiscreteUniform` no longer depends on `pymc3.distributions.dist_math.bound` for proper evaluation (see [#4541](https://github.com/pymc-devs/pymc3/pull/4541)).
 - `Model.RV_dims` and `Model.coords` are now read-only properties. To modify the `coords` dictionary use `Model.add_coord`. Also `dims` or coordinate values that are `None` will be auto-completed (see [#4625](https://github.com/pymc-devs/pymc3/pull/4625)).
 - The length of `dims` in the model is now tracked symbolically through `Model.dim_lengths` (see [#4625](https://github.com/pymc-devs/pymc3/pull/4625)).
+- `math.log1mexp` now raises a `DeprecationWarning`. The alternative `aesara.tensor.log1mexp` should be used instead. Note that the Aesara function expects negative inputs unlike the pymc3 one (see [#4860](https://github.com/pymc-devs/pymc3/pull/4860)).
```
```suggestion
- `math.log1mexp` now raises a `DeprecationWarning`. The alternative `aesara.tensor.log1mexp` should be used instead. Note that the Aesara function expects negative inputs unlike the PyMC3 one (see [#4860](https://github.com/pymc-devs/pymc3/pull/4860)).
```

pymc3/math.py
```diff
@@ -246,21 +229,37 @@ def log1mexp(x):
        "Accurately computing `\log(1-\exp(- \mid a \mid))` Assessed by the Rmpfr package"
 
     """
-    return at.switch(at.lt(x, 0.6931471805599453), at.log(-at.expm1(-x)), at.log1p(-at.exp(-x)))
+    warnings.warn(
+        "pymc3.math.log1mexp is deprecated in favor of aesara.tensor.log1mexp.\n "
```
twiecki: This is a bit confusing because there will still be a `pm.math.log1mexp` that can be used, from the user perspective, just the sign of the input changed. This is actually quite tricky, because we probably want to change the sign here too to make them compatible. As this is a new major version, I think the best way would be to raise a warning and say that they need to invert the input as the interface changed.

twiecki: Or we also add a `negative_input` flag here and deprecate that way.

pymc3/math.py
```diff
@@ -186,27 +190,14 @@ def tround(*args, **kwargs):
     return at.round(*args, **kwargs)
 
 
-def logsumexp(x, axis=None, keepdims=True):
```
twiecki, line 189: Are there tests for this we can remove?

pymc3/math.py
ricardoV94: I was thinking that we would remove it completely for one version and then bring it back, but yes, maybe that's more sensible.

pymc3/math.py
ricardoV94: Yeah, I guess we can remove the old tests

pymc3/math.py
ricardoV94: Updated to follow the same logic as the `numpy` version. Later on we should deprecate the keyword argument and finally remove the function...

pymc3/math.py
ricardoV94: Removed

twiecki (approved):

twiecki (approved):

Issue #4752, opened 2021-06-07
Reimplement old start value logic


ricardoV94:
After the initial V4 refactoring, the model initial point is now specified by random samples from the prior distribution. In, comparison, in V3 we were using central moments (mean, median, mode) to specify the starting point for most (some?) samplers. @junpenglao and @ColCarroll argue that the changes might be suboptimal.

If anyone with more background knowledge could clarify how the starting values were being used (and from where) that might help figuring out a plan to reimplement them (if that's indeed what we want to do).

brandonwillard:
That should be straightforward enough to do, but I'm admittedly curious about this strict deterministic starting point idea.  

@junpenglao and @ColCarroll, is this a concern for something outside of initializing MCMC chains, because it seems odd that one would want to fix the starting points of a sample chain based on the general form of a model (and its observations), especially when/if RNG seeding is involved, because that would ultimately fix the sample chain across multiple runs.

ricardoV94:
There is also the initial random jittering on top of the fixed initial point, which I guess would break the deterministic sampling over multiple runs. But that's where my knowledge ends

brandonwillard:
Before I forget, we still need to figure out exactly how we want to start multiple chains.  The way the initial point sampling is currently designed, we should be able to produce multiple valid starting points using the same code.

brandonwillard:
> There is also the initial random jittering on top of the fixed initial point, which I guess would break the deterministic sampling over multiple runs. But that's where my knowledge ends

Looks like you anticipated my follow-up!

Yeah, I think this all boils down to the when and where of initialization for multi-chain sampling, which has yet to be fully resolved.

PR #4867, opened 2021-07-17
Restore support for default initvals

michaelosthege:
This builds on top of #4913 to implement an API for setting default initial values.

### Changes
+ `initval=UNSET` in the signature of `Distribution.__new__` to distinguish between the user not passing the kwarg vs. asking for random initial values with `initval=None`.
+ `Distribution.pick_initval` classmethod that distributions can override to implement their own initval creation logic (as suggested by @ricardoV94 ).

### DIFF
--- a/pymc3/distributions/continuous.py
+++ b/pymc3/distributions/continuous.py
@@ -87,7 +87,7 @@ from pymc3.distributions.dist_math import (
 )
 from pymc3.distributions.distribution import Continuous
 from pymc3.math import logdiffexp, logit
-from pymc3.util import UNSET
+from pymc3.util import UNSET, select_initval
 
 __all__ = [
     "Uniform",
@@ -366,6 +366,18 @@ class Flat(Continuous):
         res.tag.test_value = np.full(size, floatX(0.0))
         return res
 
+    @classmethod
+    def pick_initval(cls, *, size=None, initval=UNSET, **kwargs) -> np.ndarray:
+        initval = select_initval(
+            candidate=initval,
+            default=np.full(size, floatX(0.0)),
+        )
+        if initval is None:
+            raise NotImplementedError(
+                "The `Flat` distribution does not support random initval sampling (`initval=None`)."
+            )
+        return initval
+
     def logp(value):
         """
         Calculate log-probability of Flat distribution at specified value.
@@ -427,6 +439,18 @@ class HalfFlat(PositiveContinuous):
         res = super().dist([], size=size, **kwargs)
         res.tag.test_value = np.full(size, floatX(1.0))
         return res
+
+    @classmethod
+    def pick_initval(cls, *, size=None, initval=UNSET, **kwargs) -> np.ndarray:
+        initval = select_initval(
+            candidate=initval,
+            default=np.full(size, floatX(1.0)),
+        )
+        if initval is None:
+            raise NotImplementedError(
+                "The `HalfFlat` distribution does not support random initval sampling (`initval=None`)."
+            )
+        return initval
 
     def logp(value):
         """
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -19,11 +19,13 @@ import types
 import warnings
 
 from abc import ABCMeta
-from typing import Optional
+from typing import Optional, Union
 
 import aesara
 import aesara.tensor as at
+import numpy as np
 
+from aesara.graph.basic import Constant, Variable
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.random.var import RandomStateSharedVariable
 
@@ -42,7 +44,7 @@ from pymc3.distributions.shape_utils import (
     resize_from_observed,
 )
 from pymc3.printing import str_for_dist
-from pymc3.util import UNSET
+from pymc3.util import UNSET, select_initval
 from pymc3.vartypes import string_types
 
 __all__ = [
@@ -126,7 +128,7 @@ class Distribution(metaclass=DistributionMeta):
         *args,
         rng=None,
         dims: Optional[Dims] = None,
-        initval=None,
+        initval=UNSET,
         observed=None,
         total_size=None,
         transform=UNSET,
@@ -149,6 +151,9 @@ class Distribution(metaclass=DistributionMeta):
         initval : optional
             Test value to be attached to the output RV.
             Must match its shape exactly.
+            If set to `None`, an initial value will be drawn randomly.
+            With a value of `UNSET`, or not passing `initval` the `cls` distribution
+            may provide a default initial value at its own discretion.
         observed : optional
             Observed data to be passed when registering the random variable in the model.
             See ``Model.register_rv``.
@@ -219,6 +224,10 @@ class Distribution(metaclass=DistributionMeta):
             # A batch size was specified through `dims`, or implied by `observed`.
             rv_out = change_rv_size(rv_var=rv_out, new_size=resize_shape, expand=True)
 
+        initval = select_initval(
+            candidate=initval,
+            default=cls.pick_initval(*args, rng=rng, **kwargs),
+        )
         rv_out = model.register_rv(
             rv_out,
             name,
@@ -326,6 +335,14 @@ class Distribution(metaclass=DistributionMeta):
             rng.default_update = new_rng
 
         return rv_out
+
+    @classmethod
+    def pick_initval(cls, *args, **kwargs) -> Union[int, float, np.ndarray, Constant, Variable]:
+        """Fallback method for creating an initial value for a random variable.
+
+        Parameters are the same as for the `.dist()` method.
+        """
+        return None
 
 
 class NoDistribution(Distribution):
--- a/pymc3/model.py
+++ b/pymc3/model.py
@@ -944,15 +944,13 @@ class Model(Factor, WithMemoization, metaclass=ContextMeta):
         return list(typefilter(self.value_vars, continuous_types))
 
     def set_initval(self, rv_var, initval):
-        if initval is not None:
+        if not isinstance(initval, (type(None), TensorVariable)):
             initval = rv_var.type.filter(initval)
-
-        test_value = getattr(rv_var.tag, "test_value", None)
 
         rv_value_var = self.rvs_to_values[rv_var]
         transform = getattr(rv_value_var.tag, "transform", None)
 
-        if initval is None or transform:
+        if transform or isinstance(initval, (type(None), TensorVariable)):
             # Sample/evaluate this using the existing initial values, and
             # with the least effect on the RNGs involved (i.e. no in-placing)
             from aesara.compile.mode import Mode, get_mode
@@ -981,17 +979,7 @@ class Model(Factor, WithMemoization, metaclass=ContextMeta):
             initval_fn = aesara.function(
                 [], rv_var, mode=mode, givens=givens, on_unused_input="ignore"
             )
-            try:
-                initval = initval_fn()
-            except NotImplementedError as ex:
-                if "Cannot sample from" in ex.args[0]:
-                    # The RV does not have a random number generator.
-                    # Our last chance is to take the test_value.
-                    # Note that this is a workaround for Flat and HalfFlat
-                    # until an initval default mechanism is implemented (#4752).
-                    initval = test_value
-                else:
-                    raise
+            initval = initval_fn()
 
         self.initial_values[rv_value_var] = initval
 
--- a/pymc3/tests/test_initvals.py
+++ b/pymc3/tests/test_initvals.py
@@ -11,9 +11,68 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+import aesara
+import aesara.tensor as at
+import aesara.tensor.random.basic as atr
+import numpy as np
+import pandas as pd
 import pytest
 
+from aesara.graph.basic import Variable
+
 import pymc3 as pm
+
+from pymc3.util import UNSET, select_initval
+
+
+def test_util_select_initval():
+    # The candidate is preferred, particularly if the default is invalid
+    assert select_initval(4, default=None) == 4
+    assert select_initval([1, 2], default=None) == [1, 2]
+    assert select_initval(None, default=3) is None
+    assert select_initval(UNSET, default=pd.Series([1, 2, 3])) is UNSET
+
+    # The default is preferred if the candidate is UNSET or invalid
+    assert select_initval(UNSET, default=3) == 3
+    assert select_initval(UNSET, default=None) is None
+    assert select_initval(pd.Series([1, 2, 3]), default=None) is None
+    assert isinstance(select_initval(UNSET, default=at.scalar()), Variable)
+    assert isinstance(select_initval(at.scalar(), default=3), Variable)
+
+    # None is the fallback if both are invalid
+    assert select_initval(pd.Series([1, 2, 3]), default="not good") is None
+    pass
+
+
+class NormalWithoutInitval(pm.Distribution):
+    """A distribution that does not specify a default initial value."""
+
+    rv_op = atr.normal
+
+    @classmethod
+    def dist(cls, mu=0, sigma=None, **kwargs):
+        mu = at.as_tensor(pm.floatX(mu))
+        sigma = at.as_tensor(pm.floatX(sigma))
+        return super().dist([mu, sigma], **kwargs)
+
+
+class UniformWithInitval(pm.distributions.continuous.BoundedContinuous):
+    """
+    A distribution that defaults the initial value.
+    """
+
+    rv_op = atr.uniform
+    bound_args_indices = (0, 1)  # Lower, Upper
+
+    @classmethod
+    def dist(cls, lower=0, upper=1, initval=UNSET, **kwargs):
+        lower = at.as_tensor_variable(pm.floatX(lower))
+        upper = at.as_tensor_variable(pm.floatX(upper))
+        return super().dist([lower, upper], **kwargs)
+
+    @classmethod
+    def pick_initval(cls, lower, upper, **kwargs):
+        return (lower + upper) / 2
 
 
 def transform_fwd(rv, expected_untransformed):
@@ -22,6 +81,9 @@ def transform_fwd(rv, expected_untransformed):
 
 class TestInitvalAssignment:
     def test_dist_warnings_and_errors(self):
+        rv = UniformWithInitval.dist(1, 2)
+        assert not hasattr(rv.tag, "test_value")
+
         with pytest.warns(DeprecationWarning, match="argument is deprecated and has no effect"):
             rv = pm.Exponential.dist(lam=1, testval=0.5)
         assert not hasattr(rv.tag, "test_value")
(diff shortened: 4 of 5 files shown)


pymc3/util.py
```diff
@@ -24,7 +24,20 @@
...
+    the argument at all."""
+
+    def __str__(self):
+        return "not set"
+
+    def __repr__(self):
+        return str(self)
+
+
+UNSET = _UnsetPlaceholder()
```
brandonwillard: What is the point of this?  Does a simple sentinel value really need a `__str__`, `__repr__`, and/or a custom type?

pymc3/model.py
```diff
@@ -913,16 +912,26 @@ def independent_vars(self):
...
         return Point(list(self.initial_values.items()), model=self)
 
+    @property
+    def initial_values(self) -> Dict[TensorVariable, np.ndarray]:
+        """Maps transformed variables to initial values.
+
+        ⚠ The keys are NOT the objects returned by, `pm.Normal(...)`.
+        """
+        return self._initial_values
+
```
brandonwillard: Why is this being purposefully exposed?  Do we expect that users will need access to this particular map?  

Remember, when you start exposing things like this, it's more likely that code outside of the PyMC3 internals will start to use/rely on it, and the internal initial value handling doesn't look mature enough to risk that.

brandonwillard (commented):
I understand that this is a WIP, but what's the basic idea (e.g. what is the mechanism for setting default initial values, how will it work, etc.)?

pymc3/util.py
michaelosthege: Not having a concise str representation makes tracebacks, error messages and warnings much longer and more confusing.
I added this right at the start when working on this branch, because it was so annoying..

brandonwillard: Can you give an example of where/when this arises?

pymc3/model.py
michaelosthege: The dictionary was exposed already, I just made it a property so I could attach type hints and docstring.

After working with `Model.initial_values` and `Model.initial_point` for a few hours now I must agree that we shouldn't expose them. (We can make it a "private" property too.)
I was very confused by the keys in these dictionaries not having anything to do with the RVs returned at the user level API.
Accordingly the values in there are in terms of transformed initvals too..

michaelosthege:
> I understand that this is a WIP, but what's the basic idea (e.g. what is the mechanism for setting default initial values, how will it work, etc.)?

I'm close to pushing another two commits. I'm primarily adding a lot of testing to nail down what even is the expected behavior.
The mechanism I'm implementing uses `UNSET` and `None` to distinguish two initval behaviors:
+ `UNSET` → The distribution may implement a default initval.
+ `None` → This means "use a random draw"

My intention there is to have the implementation of choosing a default initval in the distributions (where we had it before), while allowing users to disable that too.

pymc3/model.py
brandonwillard: > The dictionary was exposed already, I just made it a property so I could attach type hints and docstring.

Making it a property with type hints and a docstring is the type of "exposing" that I'm talking about.  If we build more stuff around these things, then they tend to become harder to refactor and more likely to be used (as-is) elsewhere.

pymc3/util.py
michaelosthege: The pytest traceback logs incoming variables:
`cls = <class 'pymc3.tests.test_initvals.NormalWithoutInitval'>, dist_params = [TensorConstant{1.0}, TensorConstant{2.0}], shape = None, size = None, initval = UNSET
kwargs = {'rng': RandomStateSharedVariable(<RandomState(MT19937) at 0x2145DA58640>)}
`
vs.
`
cls = <class 'pymc3.tests.test_initvals.NormalWithoutInitval'>, dist_params = [TensorConstant{1.0}, TensorConstant{2.0}], shape = None, size = None
initval = <pymc3.util._UnsetPlaceholder object at 0x000002017CC29130>, kwargs = {'rng': RandomStateSharedVariable(<RandomState(MT19937) at 0x2017EA98640>)}
`

I needed the type anyways to do `isinstance` checks.

**EDIT:** I have pending changes to repr as "UNSET" and change the `isinstance` checks in favor of `initval is UNSET`. Then there's no strong need for the custom class, but the str/repr is still nice to have. I have no strong opinion there.

pymc3/model.py
michaelosthege: We can make them private, but that's not in the scope of this PR.
We should then also discuss if both of `Model.initial_values`, `Model.initial_point` should be private.

Again, I'm all in favor of making them private, but to my understanding that's a matter of naming and not of docstrings and type hints, which we should have also on internal things for own good.

pymc3/distributions/continuous.py
```diff
@@ -361,8 +361,8 @@ class Flat(Continuous):
     rv_op = flat
 
     @classmethod
-    def dist(cls, *, size=None, initval=None, **kwargs):
-        if initval is None:
+    def dist(cls, *, size=None, initval=UNSET, **kwargs):
+        if initval == UNSET or initval is None:
```
michaelosthege: These two distributions have a different behavior because they cannot be sampled from.
For consistency we might want to consider raising an error on `initval=None` instead?

pymc3/distributions/distribution.py
```diff
@@ -293,6 +307,13 @@ def dist(
         # Create the RV with a `size` right away.
         # This is not necessarily the final result.
         rv_out = cls.rv_op(*dist_params, size=create_size, **kwargs)
+        if initval is None and hasattr(rv_out.tag, "test_value"):
+            warnings.warn(
+                f"An initval=None was specified, but the {cls.rv_op.__class__.__name__} Op assigned a test value anyways."
+                f" This unexpected behaviour and should be investigated!",
+                RuntimeWarning,
+            )
```
michaelosthege: In a future where we are fully compatible with randomly initvals and move away from `test_value` we don't want things like that to go unnoticed, right?

pymc3/distributions/distribution.py
```diff
@@ -322,6 +343,9 @@ def dist(
             rv_out.update = (rng, new_rng)
             rng.default_update = new_rng
 
+        if initval != UNSET and initval is not None:
+            rv_out.tag.test_value = initval
```
michaelosthege: If there's a user-specified `initval` there's a good chance that's because of numerical reasons.. The `__new__` implementation sets `test_value` too.

pymc3/distributions/distribution.py
```diff
@@ -219,9 +227,15 @@ def __new__(
...
             rv_out = change_rv_size(rv_var=rv_out, new_size=resize_shape, expand=True)
 
-        if initval is not None:
+        if initval is not UNSET and initval is not None:
             # Assigning the testval earlier causes trouble because the RV may not be created with the final shape already.
             rv_out.tag.test_value = initval
+        elif initval is None and hasattr(rv_out.tag, "test_value"):
+            warnings.warn(
+                f"An initval=None was specified, but the {cls.__name__} distribution assigned a test value anyways."
+                f" This unexpected behaviour and should be investigated!",
```
ricardoV94: I would rather raise an error then.

pymc3/tests/test_initvals.py
```diff
@@ -0,0 +1,206 @@
...
+    # The candidate is preferred if it's None or numeric, or the default is invalid
+    assert select_initval(4, default=None) == 4
+    assert select_initval(None, default=3) is None
+    with pytest.warns(None):
+        # No warning, because defaults are often symbolic (e.g. the mode of a distribution).
+        assert select_initval(UNSET, default=at.scalar()) is UNSET
+
+    # The default is preferred if the candidate is UNSET or invalid
+    assert select_initval(UNSET, default=3) == 3
+    assert select_initval(UNSET, default=None) == None
```
ricardoV94:
```suggestion
    assert select_initval(UNSET, default=None) is None
```

ricardoV94 (commented):
It seems reasonable but I am not sure we need a separate helper function just to choose between `None`, `Unset`, and something else. Also I don't really see the need for type checking, it seems things would fail pretty quickly without it (e.g. with the tag.test_value assignment), but that's just a personal opinion.

pymc3/distributions/continuous.py
ricardoV94: The error will be raised anyway during sampling of the initial point. I don't see anyone setting a Flat initial value to None and not understanding where the error is coming from

pymc3/distributions/distribution.py
michaelosthege: Are we sure that there aren't any Ops doing this legally?

ricardoV94: I am not sure, but if having a testval automatically set by the Op changes the behavior in PyMC3 wouldn't it be better to find out ASAP (ie tests failing / users opening an issue)?

michaelosthege: Your're right, we should fail faster.
Let's add it as another commit at the end after making the tests pass first. There are still some unexplained failures =/

ricardoV94:
What about having a new distrobution classmethod that takes the RV inputs and returns an initval?

Then during the super `__new__` / `dist` we just call it whenever the initval is set to `UNSET`. I think this would be safer than policing future PRs in order to respect the API by calling `select_initval`

The default would obviously be `None`

```python
class Distribution:

    ...

    @classmethod
    def initval(cls, *args):
      return None

```

```python

class Normal(Continuous):

    ...

    @classmethod
    def initval(cls, mu, sigma):
        return mu
```

Edit: Or a dispatched method like `logp` and `logcdf` if the initval is not needed at creation time

michaelosthege:
> [...] I am not sure we need a separate helper function just to choose between None, Unset, and something else.

Even though it's in the git history like that, I didn't write the helper function first: I ran into a few subtle traps when implementing it without the helper function. For example `initval == UNSET` is a bad idea, because of elementwise array comparisons.. Also one often can't use the default, because it's symbolic.. Asking future devs to do that correctly every time felt nasty..

michaelosthege:
> 
> 
> What about having a new classmethod that takes the RV inputs and returns an initval?
> 
> Then during super `__new__` / `dist` we just call it whenever the initval is set to `UNSET`. I think this would be safer than policing future PRs in order to respect the API by calling `select_initval`

In `v3` most of the initvals were computed from some kind of input transformations that were done in `.dist()` anways. Moving that into a separate method could lead to quite a bit of code duplication, no?
W.r.t. the policing: That's done in `__new__`/`dist` and I wrote quite a few tests for that too.

ricardoV94:
> In `v3` most of the initvals were computed from some kind of input transformations that were done in `.dist()` anways. Moving that into a separate method could lead to quite a bit of code duplication, no?


I think that was more out of habit than due to a conscious decision to do it that way. One advantage of having a separate method is for instance that the inputs will already be symbolic. This avoids bugs when the initval logic was based on pure numpy operations that could fail with symbolic inputs. 

Also many times it's just a question of returning one of the parameters. I don't think it makes the code more complicated, if anything it seems like a more clean separation of responsibility

Edit: it's not intuitive at all that `.dist` should be responsible for implement thr initval. Having an explicit classmethod seems much more obvious. Don't you think so?

michaelosthege:
> 
> 
> > In `v3` most of the initvals were computed from some kind of input transformations that were done in `.dist()` anways. Moving that into a separate method could lead to quite a bit of code duplication, no?
> 
> I think that was more out of habit than due to a conscious decision to do it that way. One advantage of having a separate method is for instance that the inputs will already be symbolic. This avoids bugs when the initval logic was based on pure numpy operations that could fail with symbolic inputs.
> 
> Also many times it's just a question of returning one of the parameters. I don't think it makes the code more complicated, if anything it seems like a more clean separation of responsibility

Okay fair enough. One thing though: Right now we don't support symbolic initvals, because the `test_value` needs to be numeric.

A switch to a classmethod doesn't make a difference for the `UNSET` / `None` logic, right?

ricardoV94:
> One thing though: Right now we don't support symbolic initvals, because the `test_value` needs to be numeric.

Hmm. How did it work before? For example in the case where the initval is just the `mu` of the `Normal`, which itself comes from a parent distribution.

> A switch to a classmethod doesn't make a difference for the `UNSET` / `None` logic, right?

No, but then you don't really need the helper function I think. You just call it in `__new__`, `dist` whenever you get an `UNSET` (but if you really see an advantage in the extra type checks that's fine, again it's just an aesthetic opinion)

conda-envs/environment-dev-py37.yml
```diff
@@ -3,7 +3,7 @@ channels:
 - conda-forge
 - defaults
 dependencies:
-- aesara>=2.1.0
+- aesara==2.1.0
```
twiecki: I think with https://github.com/pymc-devs/pymc3/pull/4869 we don't need this anymore.

pymc3/distributions/continuous.py
michaelosthege: I implemented an error in the `make_initval` method to make it easier to spot. Feel free to review/comment the updated code again :)

ricardoV94 (commented):
Only two minor questions / suggestions:

1. I would rename `make_initval` to `[choose|pick]_initval` or simply`initval`. 
2. Perhaps it would make more sense to move the new calls inside `model.set_initval`? Since that already handles the prior-sampling case (It seemed to me they aren't needed before). To achieve that we can simply dispatch the `make_initval` to the underlying RV, just like logp / logcdf does.

pymc3/distributions/distribution.py
michaelosthege: The `Flat` distributions assign a test value and I wouldn't be surprised if bad things happened if we were to remove that (because they can't draw random samples).
How should we proceed w.r.t. this warning then?

ricardoV94: Either remove the warning or add an exception for the distributions that must assign a test value. 

Neither option sounds good to me, but maybe removing the warning is the best? What are the problems associated with having a test value?

michaelosthege:
@ricardoV94 what's the matter with this test? https://github.com/pymc-devs/pymc3/pull/4867/checks?check_run_id=3150614327#step:7:448

I restarted it already, but it's still flaky??

ricardoV94:
> I restarted it already, but it's still flaky??

Yeah, it was not tweaked, but let's just remove it.

pymc3/distributions/distribution.py
```diff
@@ -293,6 +316,13 @@ def dist(
         # Create the RV with a `size` right away.
         # This is not necessarily the final result.
         rv_out = cls.rv_op(*dist_params, size=create_size, **kwargs)
+        if hasattr(rv_out.tag, "test_value"):
+            warnings.warn(
+                f"The {cls.rv_op.__class__.__name__} Op assigned a test value automatically."
+                f" This unexpected behaviour and should be investigated!",
```
twiecki:
```suggestion
                f" This is unexpected behavior and should be investigated!",
```

twiecki:
What's the use-case for this, when do we need it?

michaelosthege:
> What's the use-case for this, when do we need it?

It most likely fixes #4771 or at least opens a route to fix it. (E.g. by implementing a default initval for the StudentT, or other long-tailed distributions that are likely to initialize with numerically unfavorable draws).

Setting an initval can be a good idea on, for example, long random walks that would have a numerically unstable prior probability if initialized randomly (I had this recently).

ricardoV94:
> It most likely fixes #4771 

Have you tried to re-enable the tests with the old initvals of those distributions? I am still a bit skeptical that was the only issue but happy to be wrong.

ricardoV94:
> What's the use-case for this, when do we need it?

It's explained in this issue: https://github.com/pymc-devs/pymc3/issues/4752

But it was not completely argued for...

michaelosthege:
> Have you tried to re-enable the tests with the old initvals of those distributions? I am still a bit skeptical that was the only issue but happy to be wrong.

I haven't since I did not - and still don't - have the time. I'd be happy if we don't stale-out this PR.

My gut feeling is that `find_MAP` fails when extreme values are drawn at initialization. IIRC there was a StudentT involved in one of those tests. But this PR does not add default initval for StudentT; it just implements a mechanism to make that easy to do in a follow-up.

michaelosthege:
How are we feeling about this?

ricardoV94:
> How are we feeling about this?

I think it would be more clean to call `pick_initval` inside the `model.set_initval` function, so that all the `initival` logic is found in one place.

I am not sure what the new `test_val` check achieves.

Otherwise LGTM

michaelosthege:
> I think it would be more clean to call pick_initval inside the model.set_initval function, so that all the initival logic is found in one place.

Oh sorry, I forgot to mention: I tried this, but `Model.register_rv` and `Model.set_initval` are fed a `TensorVariable` and thereby doesn't have a handle on the corresponding `Distribution` class - thereby not having a handle towards the `pick_initval` classmethod.

> I am not sure what the new test_val check achieves.

The intent is to assist migration by pointing out distributions that still follow the old patterns.
But we can take those warnings out if you like.

ricardoV94:
> Oh sorry, I forgot to mention: I tried this, but `Model.register_rv` and `Model.set_initval` are fed a `TensorVariable` and thereby doesn't have a handle on the corresponding `Distribution` class - thereby not having a handle towards the `pick_initval` classmethod.
> 

We can get the class `pick_initval` via the owner.op of the TensorVariable, similar to how we get the logp / logcdf

michaelosthege:
I rebased this and reconsidered the [dispatching approach](https://github.com/pymc-devs/pymc3/pull/4867#issuecomment-888230480), but concluded that it's not worth the additional complexity. While logp and logcdf are related to the RV, the `initval` just concerns the PyMC3 model, so I don't think we should dispatch it onto the Aesara Op.

Please take a final look so we can (squash?) merge and get this over with.

ricardoV94:
How does your implementation cope with hierarchical models? If it's not based on the `tag.test_value` trick, the initvals will be symbolic expressions.

```python
with pm.Model() as m:
  x = pm.Normal("x", 0, 1)
  y = pm.Normal("y", x, 1)
```

Assuming the `pick_initval` of `pm.Normal` is just `mu`, how would the `y` initval be calculated? The only way I see it (without relying on `tag.test_value`) is to compile the expression and evaluate it, similarly to how it's done in `model.set_initval` for forward samples.

ricardoV94:
By the way @kc611 is exploring the alternative solution with dispatching. So he might have some ideas as well.

michaelosthege:
> How does your implementation cope with hierarchical models? If it's not based on the `tag.test_value` trick, the initvals will be symbolic expressions.
> 
> ```python
> with pm.Model() as m:
>   x = pm.Normal("x", 0, 1)
>   y = pm.Normal("y", x, 1)
> ```
> 
> Assuming the `pick_initval` of `pm.Normal` is just `mu`, how would the `y` initval be calculated? The only way I see it (without relying on `tag.test_value`) is to compile the expression and evaluate it, similarly to how it's done in `model.set_initval` for forward samples.

I'm just forwarding them to `model.set_initval`, so whatever it did with symbolic initvals before this PR still applies the same way.
I'm not on the laptop right now, but if I remember correctly there already was code to compile and evaluate symbolic initvals.

ricardoV94:
> I'm just forwarding them to `model.set_initval`, so whatever it did with symbolic initvals before this PR still applies the same way.
> I'm not on the laptop right now, but if I remember correctly there already was code to compile and evaluate symbolic initvals. 

Looking at `model.set_initval` it seems that when given a non-`None` initval it only checks if it should be transformed. It wouldn't do anything for the `Normal` for example

https://github.com/pymc-devs/pymc3/blob/c9e460f29396130ff625bf84a99943c935ac5be9/pymc3/model.py#L951

And even when transforming initvals, it seems to expect them to be a Constant ~~Nvm that's for the previous initvals~~

michaelosthege:
> Looking at `model.set_initval` it seems that when given a non-`None` initval it only checks if it should be transformed. It wouldn't do anything for the `Normal` for example

Improving the `model.set_initval` method is out of scope of this PR.
I edited just one line in there to stop it from accessing `test_values`, but that doesn't change its behavor for scenarios where `initval` is a default OR user-provided `TensorVariable`.

Please open a new Issue if it misbehaves.

ricardoV94:
> Improving the `model.set_initval` method is out of scope of this PR.

I don't see how that can be out-of-scope for a PR that is refactoring how initvals work.
 
> Please open a new Issue if it misbehaves.

I am pretty sure it will misbehave in the example I just gave.

If the new implementation is not designed to handle a simple realistic case, how can it be evaluated?

ricardoV94:
```python
import pymc3 as pm
import aesara.tensor as at
import aesara.tensor.random as atr

class NormalWithInitval(pm.distributions.Continuous):
    """
    A distribution that defaults the initial value.
    """

    rv_op = atr.normal

    @classmethod
    def dist(cls, mu=0, sigma=1, **kwargs):
        mu = at.as_tensor_variable(pm.floatX(mu))
        sigma = at.as_tensor_variable(pm.floatX(sigma))
        return super().dist([mu, sigma], **kwargs)

    @classmethod
    def pick_initval(cls, mu, sigma, **kwargs):
        return mu

with pm.Model() as m:
    x = NormalWithInitval('x', 0, 1)
    y = NormalWithInitval('y', x, 1)  # raises TypeError
```

michaelosthege:
Your example also raises the TypeError on `main`. I have therefore opened #4911.

It was also not too difficult to fix, even though that was not the goal of this PR.

michaelosthege:
superseded by #4983

Issue #4831, opened 2021-07-02
Refactor or deprecate `DensityDist`


ricardoV94:
We need to decide whether we want to deprecate `DensityDist` and if so, provide at least documentation on how to add simple logp terms to a model or how to quickly implement a custom Distribution.

If we want to refactor it, it is probably good time to get rid of the whole dictionary as observed thing and force the distribution parameters to go into the signature as normal arguments in other distributions

```python
# V3 allows this
def my_logp(value, param1, param2):
    return value * param1 - param2

param1 = pm.Normal('param1', 0, 1)
param2 = pm.Normal('param2', 0, 1)
y = pm.DensityDist('y', my_logp, observed=dict(value=data, param1=param1, param2=param2))
```

```python
# V4 should work only like this (if not deprecated)
y = pm.DensityDist('y', my_logp, param1, param2, obeserved=data)

ricardoV94:
CC @lucianopaz, if you want to assign this to yourself. Seems like we are pretty close to having a working solution

OriolAbril:
looks related to https://github.com/pymc-devs/pymc3/issues/4534, and I also wanted to leave a link to https://discourse.pymc.io/t/using-a-random-variable-as-observed/7184/5 so we are aware of all the (in my opinion strange and redundant) things DensityDist allows, expecially when taking a dictionary as `observed` (which I also think is the only Distribution that does). 

As I have commented a few times and started changing in https://github.com/pymc-devs/pymc3/pull/4433, I think it's a bad idea to allow `FreeRVs` to be values in the dictionary passed as observed. If this is a feature that is really important to DensityDist, I would suggest doing this `givens` extra argument or following @ricardoV94 suggestion.

PR #4365, opened 2020-12-21
Pass a string to NDArray(name=...) and raise NotImplemented on named models in SMC

basnijholt:
Currently, `name` is a `Model`.

**Thank your for opening a PR!**

Depending on what your PR does, here are a few things you might want to address in the description:
+ [x] what are the (breaking) changes that this PR makes? → see thread & release notes
+ [x] important background, or details about the implementation → see thread
+ [x] are the changes—especially new features—covered by tests and docstrings?
+ [ ] <s>consider adding/updating relevant example notebooks</s>
+ [x] right before it's ready to merge, mention the PR in the RELEASE-NOTES.md

PR #4877, opened 2021-07-23
Refactor pm.Simulator (2nd attempt)

ricardoV94:
The previous attempt at this in #4802 revealed big issues between pickling and dynamically created classes (which subsist even after #4858). The alternative presented in this PR is a bit more cumbersome from the user-side but at least avoids these issues. 

Here is a minimal example demonstrating the new API:

```python
import numpy as np
import pymc3 as pm

data = np.random.normal(0, 1, size=10)

def my_simulator_fn(rng, loc, scale, size):
    return rng.normal(loc, scale, size=size)

class MySimulatorRV(pm.SimulatorRV):
    ndim_supp = 0
    ndims_params = [0, 0, 0]
    fn = my_simulator_fn
    distance = "gaussian"
    sum_stat = "sort"

my_simulator = MySimulatorRV()

with pm.Model() as m:
    simulator = pm.Simulator("sim", my_simulator, 0, 1, epsilon=1.0, observed=data)
```

If anyone has better suggestions about how to avoid pickling issues when creating dynamic classes please let me know! For example the classes used in these tests had to be defined outside of `TestSimulator.setup_class`.

### DIFF
--- a/pymc3/aesaraf.py
+++ b/pymc3/aesaraf.py
@@ -350,15 +350,26 @@ def rvs_to_value_vars(
 
     """
 
+    # Avoid circular dependency
+    from pymc3.distributions.simulator import SimulatorRV
+
     def transform_replacements(var, replacements):
         rv_var, rv_value_var = extract_rv_and_value_vars(var)
 
         if rv_value_var is None:
-            warnings.warn(
-                f"No value variable found for {rv_var}; "
-                "the random variable will not be replaced."
-            )
-            return []
+            # If RandomVariable does not have a value_var and corresponds to
+            # a SimulatorRV, we allow further replacements in upstream graph
+            if isinstance(rv_var.owner.op, SimulatorRV):
+                # First 3 inputs are just rng, dtype, and size, which don't
+                # need to be replaced.
+                return var.owner.inputs[3:]
+
+            else:
+                warnings.warn(
+                    f"No value variable found for {rv_var}; "
+                    "the random variable will not be replaced."
+                )
+                return []
 
         transform = getattr(rv_value_var.tag, "transform", None)
 
--- a/pymc3/distributions/simulator.py
+++ b/pymc3/distributions/simulator.py
@@ -13,113 +13,25 @@
 #   limitations under the License.
 
 import logging
+import warnings
 
+import aesara
+import aesara.tensor as at
 import numpy as np
 
+from aesara.graph.op import Apply, Op
+from aesara.graph.utils import MetaType
+from aesara.tensor.random.op import RandomVariable
+from aesara.tensor.var import TensorVariable
 from scipy.spatial import cKDTree
 
+from pymc3.aesaraf import floatX
 from pymc3.distributions.distribution import NoDistribution
+from pymc3.distributions.logprob import _logp
 
-__all__ = ["Simulator"]
+__all__ = ["Simulator", "SimulatorRV"]
 
 _log = logging.getLogger("pymc3")
-
-
-class Simulator(NoDistribution):
-    r"""
-    Define a simulator, from a Python function, to be used in ABC methods.
-
-    Parameters
-    ----------
-    function: callable
-        Python function defined by the user.
-    params: list
-        Parameters passed to function.
-    distance: str or callable
-        Distance functions. Available options are "gaussian" (default), "laplacian",
-        "kullback_leibler" or a user defined function that takes epsilon, the summary statistics of
-        observed_data and the summary statistics of simulated_data as input.
-        ``gaussian`` :math: `-0.5 \left(\left(\frac{xo - xs}{\epsilon}\right)^2\right)`
-        ``laplace`` :math: `{\left(\frac{|xo - xs|}{\epsilon}\right)}`
-        ``kullback_leibler`` `:math: d \sum(-\log(\frac{\nu_d} {\rho_d}) / \epsilon) + log_r`
-        gaussian + ``sum_stat="sort"`` is equivalent to the 1D 2-wasserstein distance
-        laplace + ``sum_stat="sort"`` is equivalent to the the 1D 1-wasserstein distance
-    sum_stat: str or callable
-        Summary statistics. Available options are ``indentity``, ``sort``, ``mean``, ``median``.
-        If a callable is based it should return a number or a 1d numpy array.
-    epsilon: float or array
-        Scaling parameter for the distance functions. It should be a float or an array of the
-        same size of the output of ``sum_stat``.
-    *args and **kwargs:
-        Arguments and keywords arguments that the function takes.
-    """
-
-    def __init__(
-        self,
-        function,
-        *args,
-        params=None,
-        distance="gaussian",
-        sum_stat="identity",
-        epsilon=1,
-        **kwargs,
-    ):
-        self.function = function
-        self.params = params
-        observed = self.data
-        self.epsilon = epsilon
-
-        if distance == "gaussian":
-            self.distance = gaussian
-        elif distance == "laplace":
-            self.distance = laplace
-        elif distance == "kullback_leibler":
-            self.distance = KullbackLiebler(observed)
-            if sum_stat != "identity":
-                _log.info(f"Automatically setting sum_stat to identity as expected by {distance}")
-                sum_stat = "identity"
-        elif hasattr(distance, "__call__"):
-            self.distance = distance
-        else:
-            raise ValueError(f"The distance metric {distance} is not implemented")
-
-        if sum_stat == "identity":
-            self.sum_stat = identity
-        elif sum_stat == "sort":
-            self.sum_stat = np.sort
-        elif sum_stat == "mean":
-            self.sum_stat = np.mean
-        elif sum_stat == "median":
-            self.sum_stat = np.median
-        elif hasattr(sum_stat, "__call__"):
-            self.sum_stat = sum_stat
-        else:
-            raise ValueError(f"The summary statistics {sum_stat} is not implemented")
-
-        super().__init__(shape=np.prod(observed.shape), dtype=observed.dtype, *args, **kwargs)
-
-    def random(self, point=None, size=None):
-        """
-        Draw random values from Simulator.
-
-        Parameters
-        ----------
-        point: dict, optional
-            Dict of variable values on which random values are to be conditioned (uses default
-            point if not specified).
-        size: int, optional
-            Desired size of random sample (returns one sample if not specified).
-
-        Returns
-        -------
-        array
-        """
-        # size = to_tuple(size)
-        # params = draw_values([*self.params], point=point, size=size)
-        # if len(size) == 0:
-        #     return self.function(*params)
-        # else:
-        #     return np.array([self.function(*params) for _ in range(size[0])])
 
 
 def identity(x):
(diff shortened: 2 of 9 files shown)


pymc3/aesaraf.py
```diff
@@ -350,15 +350,24 @@ def rvs_to_value_vars(
...
         rv_var, rv_value_var = extract_rv_and_value_vars(var)
 
         if rv_value_var is None:
-            warnings.warn(
-                f"No value variable found for {rv_var}; "
-                "the random variable will not be replaced."
-            )
-            return []
+            # If orphan RandomVariable corresponds to a SimulatorRV, we allow
+            # for further replacements in upstream graph
```
michaelosthege: Why do you consider `rv_var` an "orphan" RandomVariable?
I guess this function could use a docstring and maybe also a comment on line 357 to explain what it's doing on a high level.

pymc3/aesaraf.py
```diff
@@ -350,15 +350,24 @@ def rvs_to_value_vars(
...
+            if isinstance(rv_var.owner.op, SimulatorRV):
+                return var.owner.inputs[3:]
```
michaelosthege, line 365: Why `[3:]` ? What are the first three elements and why are they skipped?

A type hint on the return value would also help to understand.

pymc3/distributions/simulator.py
```diff
@@ -14,127 +14,314 @@
...
+
+        def my_simulator_fn(rng, loc, scale, size):
+            return rng.normal(loc, scale, size=size)
+
+        class MySimulatorRV(pm.SimulatorRV):
+            ndim_supp = 0
+            ndims_params = [0, 0, 0]
+            fn = my_simulator_fn
+            distance = "gaussian"
+            sum_stat = "sort"
```
michaelosthege, line 108: Shouldn't the name be implemented by a subclass too?

pymc3/distributions/simulator.py
```diff
@@ -14,127 +14,314 @@
...
+            elif callable(sum_stat):
+                sum_stat = create_sum_stat_op_from_fn(sum_stat)
+            else:
+                raise ValueError(f"The summary statistic {sum_stat} is not implemented")
+
+        cls.distance = distance
+        cls.sum_stat = sum_stat
+        return super().__new__(
+            cls,
+        )
```
michaelosthege:
```suggestion
        return super().__new__(cls)
```

pymc3/tests/test_smc.py
```diff
@@ -270,28 +468,38 @@ def test_automatic_use_of_sort(self):
...
         with pm.Model(name="NamedModel"):
             a = pm.Normal("a", mu=0, sigma=1)
             b = pm.HalfNormal("b", sigma=1)
-            c = pm.Potential("c", pm.math.switch(a > 0, 0, -np.inf))
-            s = pm.Simulator(
-                "s", normal_sim, params=(a, b), sum_stat="sort", epsilon=1, observed=self.data
-            )
+            s = pm.Simulator("s", NormalSimRV1(), a, b, observed=self.data)
+
+            # TODO: Why is this?
```
michaelosthege, line 484: The exception message links to my answer here: https://github.com/pymc-devs/pymc3/pull/4365#issuecomment-761221146

If you read it you can probably tell if that's outdated with your latest changes or not.

pymc3/aesaraf.py
ricardoV94: Because it doesn't have a value var associated with it. I'll make the docstrings more explicit

pymc3/aesaraf.py
ricardoV94: The first 3 arguments of a RV are the `rng`, `dtype` and `size`. They don't need to be replaced by value vars. I'll add a comment

pymc3/distributions/simulator.py
ricardoV94: It can be implemented optionally, but I did not want to make it a requirement if it's not needed

ricardoV94:
I decided to leave epsilon as an explicit argument to make it easier to manipulate (e.g., if we want our samplers to tune it), but can be a class attribute like `distance` and `sum_stat` if we don't think that will be ever used.

I can see it as being confusing like this...

junpenglao:
Instead of subclassing and initializing the subclass, would something like:
```python
my_simulator = pm.SimulatorRV(
    ndim_supp = 0
    ndims_params = [0, 0, 0]
    fn = my_simulator_fn
    distance = "gaussian"
    sum_stat = "sort"
)
```
works?

michaelosthege:
> 
> 
> Instead of subclassing and initializing the subclass, would something like:
> 
> ```python
> my_simulator = pm.SimulatorRV(
>     ndim_supp = 0
>     ndims_params = [0, 0, 0]
>     fn = my_simulator_fn
>     distance = "gaussian"
>     sum_stat = "sort"
> )
> ```
> 
> works?

The thing needs to be a class, so `pm.make_simulator()` would have to create a class inside and return that. But these kinds of class definitions often cause problems with pickling.
But I agree that the API looks much nicer. @ricardoV94 what do you think?

junpenglao:
> The thing needs to be a class, so pm.make_simulator() would have to create a class inside and return that. But these kinds of class definitions often cause problems with pickling.

I am assuming that as long as `pm.make_simulator()` is called OUTSIDE of the with context we should be able to pickle it - @ricardoV94 do we have some minimal reproducible example that is easy to play with?

ricardoV94:
> > The thing needs to be a class, so pm.make_simulator() would have to create a class inside and return that. But these kinds of class definitions often cause problems with pickling.
> 
> I am assuming that as long as `pm.make_simulator()` is called OUTSIDE of the with context we should be able to pickle it - @ricardoV94 do we have some minimal reproducible example that is easy to play with?

Here is a minimal gist that implements the current API and has some checks: https://gist.github.com/ricardoV94/b632085b20be716b87fd146609168090

And here is another gist with a previous iteration on these ideas (using pickle instead of the more flexible cloudpickle that we are now using in PyMC3): https://gist.github.com/ricardoV94/2bb59a2ac18a29f501f5511c9671ebbc

pymc3/smc/sample_smc.py
```diff
@@ -157,6 +147,29 @@ def sample_smc(
         %282007%29133:7%28816%29>`__
     """
 
+    if kernel is not None:
```
aloctavodia: This should not be deprecated, we still want to have kernels.

pymc3/smc/sample_smc.py
```diff
@@ -157,6 +147,29 @@ def sample_smc(
...
+        warnings.warn(
+            "The argument kernel in sample_smc has been deprecated. It is no "
+            "longer needed to specify it.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+    if save_sim_data is not None:
```
aloctavodia, line 158: We still want this, as doing `pm.sample_posterior_predictive` could be potentially too expensive.

pymc3/smc/sample_smc.py
```diff
@@ -313,23 +309,16 @@ def sample_smc(
             trace.report._run_convergence_checks(idata, model)
     trace.report._log_summary()
 
-    posterior = idata if return_inferencedata else trace
-    if save_sim_data:
-        return posterior, {modelcontext(model).observed_RVs[0].name: np.array(sim_data)}
-    else:
-        return posterior
+    return idata if return_inferencedata else trace
```
aloctavodia, line 318: when is the trace going to be deprecated? maybe we could only return inferencedata

pymc3/distributions/simulator.py
```diff
@@ -14,127 +14,319 @@
...
+    @classmethod
+    def _distance(cls, epsilon, value, simulated_value):
+        return cls.distance(epsilon, value, simulated_value)
+
+    @classmethod
+    def _sum_stat(cls, value):
+        return cls.sum_stat(value)
+
+
 class Simulator(NoDistribution):
```
aloctavodia, line 200: nitpick, if we are going to have a `Simulator` and a `SimulatorRV`, maybe the first one should be  renamed to `PseudoLikelihood`, `SimulatedLikelihood` `AbcLikelihood` or something similar. One counterargument to this proposal, is that if we are going to distinguish between simulator and pseudolikelihood, the distance and summary statistics should be part of the later not the former.

aloctavodia (commented):
a few nitpicks

pymc3/smc/sample_smc.py
ricardoV94: We no longer have access to the simulated data in the `logp` graph

pymc3/smc/sample_smc.py
ricardoV94: I would only deprecate it when it is deprecated in `pm.sample`

pymc3/distributions/simulator.py
ricardoV94: Yeah, that was one of the ideas, the `SimulatorRV` would simply be concerned with the random draws and the `Pseudolikelihood` would take care of the `logp` factor. 

The downside is that we don't yet know how to create a "dynamic" `logp` using the optional user defined `sum_stat` and `distance` functions. It means we might need to have users subclass not only the `SimulatorRV` but also the `Pseudolikelihood` if they want to use non-default functions. If someone figures out #4831, then we could simply copy their strategy.

Defining the functions in the `SimulatorRV` was just an ugly hack to avoid forcing users to create two new classes...

pymc3/smc/sample_smc.py
ricardoV94: Yeah I know. We should just deprecate the "keywords" for the time being.

pymc3/aesaraf.py
```diff
@@ -350,15 +350,26 @@ def rvs_to_value_vars(
...
+            if isinstance(rv_var.owner.op, SimulatorRV):
+                # First 3 inputs are just rng, dtype, and size, which don't
+                # need to be replaced.
+                return var.owner.inputs[3:]
+
+            else:
+                warnings.warn(
+                    f"No value variable found for {rv_var}; "
+                    "the random variable will not be replaced."
+                )
```
michaelosthege, line 371: What conclusion should a user make from from this warning?
Is it serious? If so we should raise. Otherwise maybe just `_log.warn()`?

lucianopaz:
> > Instead of subclassing and initializing the subclass, would something like:
> > ```python
> > my_simulator = pm.SimulatorRV(
> >     ndim_supp = 0
> >     ndims_params = [0, 0, 0]
> >     fn = my_simulator_fn
> >     distance = "gaussian"
> >     sum_stat = "sort"
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > works?
> 
> The thing needs to be a class, so `pm.make_simulator()` would have to create a class inside and return that. But these kinds of class definitions often cause problems with pickling.
> But I agree that the API looks much nicer. @ricardoV94 what do you think?

Could this be made to work with pickle if we defined `SimulatorRV` as a metaclass or provide a `__reduce__` method to it? Something like what's [done here](https://codereview.stackexchange.com/a/259797)?

pymc3/aesaraf.py
ricardoV94: We should probably raise

PR #4886, opened 2021-07-28
Fix tuning of CompoundStep with multiple chains in serial

msibaev:
When sampling a model using Compound step in serial the first chain runs with tuning, but for subsequent chains individual sampler has `tune` parameter set to `False`. 

The problem seems to be in `sampling._iter_sample()` [here](https://github.com/pymc-devs/pymc3/blob/main/pymc3/sampling.py#L1016-L1018).  The `tune` flag is set, but CompoundStep class does not have that variable so it doesn't get propagated to individual samplers. It does have `stop_tuning` which gets triggered on the first chain and stops tuning for all subsequent chains.

Now, there is also `reset_tuning` which in `BaseHMC` sets `tune=True` [here](https://github.com/pymc-devs/pymc3/blob/v3/pymc3/step_methods/hmc/base_hmc.py#L227). However, `Metropolis` resets all tuning parameters other than value of `tune` itself. A more consistent approach might be to store the original value of `tune` and use that in `reset_tuning`. Then the line in `sampling._iter_sample()` where `tune` is set could be removed.

But, I'm not sure how this would work with all the other samplers.

Instead, to keep the changes minimal I added a `tune` property to `CompoundStep`. The setter will set any `tune` variables in individual steppers if it exists. And the getter checks if any of the steppers has tuning on.

This was good enough for my purposes, but I'm still not sure whether it solves all issues. In `_iter_sample` [here](https://github.com/pymc-devs/pymc3/blob/main/pymc3/sampling.py#L1023-L1024) `step.iter_count` is set to 0, if it exists. That variable does not exist in `CompoundStep`, but does exist in `BaseHMC`.  

There are two ways we can solve this

1. Add `iter_count` as property to `CompoundStep` same as I did with `tune`
2. Update `reset_tuning` to set `iter_count = 0` and possibly remove that line from `_iter_sample`

I think option 2 would be a cleaner solution. If we go that route than `tune` should be set to original value when resetting as well. I think I'll implement that, unless there are other suggestions.


msibaev:
Some samplers have a `tune` variable and no `reset_tuning` method. That's why [`_iter_sample()`](https://github.com/pymc-devs/pymc3/blob/main/pymc3/sampling.py#L1016-L1018)  cannot rely on calling `reset_tuning()` only.

__Note__, `DEMetropolis`  only accepts `tune` values in `[None, "scaling", "lambda]"`, but `_iter_sample` sets it to `bool`. As a result, tuning won't be performed during a step, see [here](https://github.com/pymc-devs/pymc3/blob/main/pymc3/step_methods/metropolis.py#L734-L737). This has been fixed in `DEMetropolisZ` by introducing [`tune_type`](https://github.com/pymc-devs/pymc3/blob/main/pymc3/step_methods/metropolis.py#L864-L867) parameter and making `tune` boolean. This fix should be copied in a separate PR

msibaev:
Variable `iter_count` is used by `NUTS` to effect [max tree depth during tuning](https://github.com/pymc-devs/pymc3/blob/main/pymc3/step_methods/hmc/nuts.py#L175-L176) and by `base_hmc` in warning messages [here](https://github.com/pymc-devs/pymc3/blob/main/pymc3/step_methods/hmc/base_hmc.py#L163-L168) and [here](https://github.com/pymc-devs/pymc3/blob/main/pymc3/step_methods/hmc/base_hmc.py#L205-L214).

It can be set to 0 in `reset_tuning` as well. Although, it seems what `_iter_sample` needs is a general `step.reset()` method which brings it back to the original configuration. However, for most samplers `reset()` will simply call `reset_tuning()`, except for `NUTS` which would also set `iter_count=0`. It's not worth adding extra layer of complexity for only one step sampler.

Funny enough `BaseHMC` has a `reset` method, which is called by `reset_tuning`. Exactly the opposite mechanism that I described above. Anyway, that's the natural place to set `iter_count=0`.

michaelosthege:
Hi @msibaev ,
Sorry about the delay. I just got the chance to read your descriptions and look at the changes. I believe the history was such that in the beginning there was only a boolean tune parameter, but the typing of that interface diffused away over time.
When I implemented `DEMetropolis` the boolean setting was no longer enough, because the sampler had to reset some things when switching in/out the tuning phase. If I remember correctly that's when I introduced the `reset_tuning` method.

At the `_iter_sample` level we can't deal with all the specifics of the different samplers, so yes we should unify that to one interface and get rid of the `hasattr() ` things.

msibaev:
Hi @michaelosthege

I don't get much time to look at these things nowadays, sorry for the delay. 

Thanks for the context of `rest_tuning()`.  I guess there should be some more standardization across all samplers in that respect, which would allow `_iter_sample()` to be cleaned up a bit more. But, that probably goes beyond the scope of this PR. My main goal was just to fix tuning when multiple chains are running in serial. It's an edge case, but an important one for windows machines where multiprocessing can be buggy.

Is there anything else that you think should be done here?

pymc3/tests/test_step.py
```diff
@@ -677,6 +677,22 @@ def test_blocked(self):
                 assert not isinstance(sampler_instance, CompoundStep)
                 assert isinstance(sampler_instance, sampler)
 
+    def test_tune_parameter_in_sequential_chain_sampling(self):
+        __, model = simple_2model_continuous()
```
michaelosthege: Typically we use single underscores. Not sure if the pre-commit may complain about it...

pymc3/tests/test_step.py
```diff
@@ -677,6 +677,22 @@ def test_blocked(self):
...
+                assert isinstance(sampler_instance, CompoundStep)
+                for __ in range(2):
+                    sampling = _iter_sample(draws=3, step=sampler_instance, tune=1)
+                    tune_sampler = []
+                    tune_method = []
+                    for __ in sampling:
+                        tune_sampler.append(sampler_instance.tune)
+                        tune_method.append([method.tune for method in sampler_instance.methods])
+                    assert [True, False, False] == tune_sampler
+                    assert [[True, True], [False, False], [False, False]] == tune_method
```
michaelosthege, line 720: The name of the test implies that this should be about sequential chains. So I guess at least one more chain should be "sampled", after calling `reset_tuning`?

pymc3/step_methods/compound.py
```diff
@@ -72,6 +72,17 @@ def reset_tuning(self):
...
+    def tune(self):
+        do_tune = any(method.tune for method in self.methods if hasattr(method, "tune"))
+        return do_tune
+
+    @tune.setter
+    def tune(self, value):
+        for method in self.methods:
+            if hasattr(method, "tune"):
+                method.tune = value
+
```
michaelosthege, line 85: The `reset_tuning` method above already does a very similar thing. I would prefer if we didn't add a `tune` property here, but instead add/implement the `reset_tuning` method on the steppers that currently don't reset their `tune` setting themselves.

With small modifications (see below) the `test_tune_parameter_in_sequential_chain_sampling` test you wrote seems to be well-suited to verify that behavior for the samplers that currently don't reset correctly.

pymc3/sampling.py
```diff
@@ -1002,8 +1002,6 @@ def _iter_sample(
             stats = None
             diverging = False
 
-            if i == 0 and hasattr(step, "iter_count"):
-                step.iter_count = 0
```
michaelosthege, line 1022: I like this!

michaelosthege (changes requested):
hI @msibaev, thank's for picking it up again.

I think we should go for the long-term solution right away. It seems to me that it could be realized with equally few changes.

Another thing we'll need to is a mention of this PR and the API change with respect to tuning reset in the `RELEASE-NOTES.md`.

pymc3/step_methods/metropolis.py
```diff
@@ -94,6 +98,28 @@ def __call__(self, num_draws=None):
...
 
 
+class SettingNotFoundInAttribute(BaseException):
+    pass
+
+
+class SettingsResetter:
+    """Stores a copy of initial settings so they can be reset on call"""
+
+    initial_settings: dict[str, Any]
```
michaelosthege:
```suggestion
    initial_settings: Dict[str, Any]
```
To fix the docs build.

michaelosthege:
Hi @msibaev , I see you're working on this branch again. Cool!
Let us know when it's ready for review

pymc3/tests/test_step.py
msibaev: Thanks, I'll use single underscore to be consistent.

ricardoV94:
Would be good to ooen an issue for the problem this PR was trying to address, as it's been stale for a while

ricardoV94:
Closing this due to lack of activity and the fact it is targeting V3

PR #4887, opened 2021-07-28
Integration of `aeppl` with PyMC

kc611:
This PR refactors PyMC's `_logp` framework to be able to use `aeppl`'s `joint_logprob` framework. 

The changes this PR aims to make:

- [x] Registering the Log-probability functions of PyMC's distributions on `_logprob` instead of `_logp` so that `joint_logprob` can use them internally.
- [x] Separating `log-cdf` logic handling from `logp` logic (currently both are being handled in the same `logpt` function)
- [x] Removing redundant `logp` from Distribution classes (since `_logprob` registration present in `aeppl, for e.g. `NormalRV`)
- [x] Calling `factorized_joint_logprob` inside `logpt` function. and then return the appropriate RVs (since `joint_logprob` tends to return all of the RV's `logp` present in the graph)
- [x] Switching from `pymc`'s transforms to `aeppl`'s transforms.
  - [x] Refactor `transforms.py` and remove transforms already present in `aeppl`
  - [x] Updating transform related logic to match `aeppl`'s. 
  - [x] Passing transforms to `joint_logprob` correctly
  - [x] Refactoring `test_transforms.py`
- [x] Investigate test failures.
  - [x] test_distributions failures. 
  - [x] `test_transforms.py` test failures. 
  - [x] `test_logprob` failures.


pymc3/distributions/logprob.py
```diff
@@ -259,7 +265,14 @@ def _logp(
 
     """
     value_var = rvs_to_values.get(var, var)
-    return at.zeros_like(value_var)
+    _values = {k: at.as_tensor_variable(e) for k, e in rvs_to_values.items()}
+    logp_fn = joint_logprob(var, _values, **kwargs)
```
kc611: This way the `Op`s with no registered logp's in PyMC can 'fall back' to their aeppl implementation and can be replaced with appropriate RVs (For instance the current Mixture will be a `Subtensor` later replaced by `MixtureRV` and also the upcoming `Scan` which will be replaced by `ScanRV` )

Is this a good entry point for `joint_logprob`? cc @brandonwillard

pymc3/distributions/logprob.py
```diff
@@ -283,63 +296,64 @@ def indices_from_subtensor(idx_list, indices):
...
 
 
-@_logp.register(IncSubtensor)
-@_logp.register(AdvancedIncSubtensor)
-@_logp.register(AdvancedIncSubtensor1)
-def incsubtensor_logp(op, var, rvs_to_values, indexed_rv_var, rv_values, *indices, **kwargs):
+# @_logp.register(IncSubtensor)
+# @_logp.register(AdvancedIncSubtensor)
+# @_logp.register(AdvancedIncSubtensor1)
+# def incsubtensor_logp(op, var, rvs_to_values, indexed_rv_var, rv_values, *indices, **kwargs):
```
kc611, line 302: These needed to be removed (temporarily) since Mixture will being a `Subtensor` will be dispatched here.  We need to take care of these cases i.e when the `Subtensor`s aren't `Mixture`s they should fall back to here. Is there a way we could do that ?

pymc3/distributions/mixture.py
```diff
@@ -112,451 +109,15 @@ class Mixture(Distribution):
...
-        # mixed_samples = np.reshape(mixed_samples, (-1, w.shape[-1]))
-        # # Select the samples from the mixture
-        # samples = np.array([mixed[choice] for choice, mixed in zip(w_samples, mixed_samples)])
-        # # Reshape the samples to the correct output shape
-        # if size is None:
-        #     samples = np.reshape(samples, dist_shape)
-        # else:
-        #     samples = np.reshape(samples, size + dist_shape)
-        # return samples
+        return at.stack(comp_dists)[w]
```
kc611: Here we'll need to take care of the `size` argument passed to the Mixture distribution itself.

pymc3/distributions/logprob.py
brandonwillard: Something like that could work for now, but keep on mind that we'll probably want to entirely remove this module at some point after the integration is established.

ricardoV94:
I would guess the "Mixture" distribution in PyMC3 would only have to be concerned with returnig the right random graph, from which the logprob can be derived with the `aeppl`. 

I am not sure how feasible it would be to plug the `aeppl.joint_logprob` for individual variables, as opposed to using it for the entire model.

kc611:
One thing we can do over here is :

https://github.com/pymc-devs/pymc3/blob/819f045ad36d1d8b18651528384972dc2bea8213/pymc3/distributions/distribution.py#L97

Registering the `logp`s to `_logprob` from aeppl instead of `_logp`. This will allow us to make Mixture's of RV's not present in `aeppl` but are present in PyMC (Which is something we'd obviously want to support. ) The problem however comes with the RV's present in both such as Normal and Uniform. They'll get registered twice.

pymc3/distributions/logprob.py
```diff
@@ -302,10 +257,13 @@ def incsubtensor_logp(op, var, rvs_to_values, indexed_rv_var, rv_values, *indice
     return logp_var
 
 
-@_logp.register(Subtensor)
-@_logp.register(AdvancedSubtensor)
-@_logp.register(AdvancedSubtensor1)
-def subtensor_logp(op, var, rvs_to_values, indexed_rv_var, *indices, **kwargs):
+@_logprob.register(Subtensor)
```
kc611: For some reason these aren't being registered onto `_logprob`.

brandonwillard: We shouldn't need these anymore, right?  The functionality should already be covered by AePPL.

pymc3/distributions/logprob.py
```diff
@@ -141,8 +129,89 @@ def logpt(
...
+        for node in io_toposort(graph_inputs((logp_var,)), (logp_var,)):
+            compute_test_value(node)
+
+    if var.name is not None:
+        logp_var.name = "__logp_%s" % var.name
+
+    return logp_var
+
+
+def logcdfpt(
```
kc611: Separated the `logcdf` and `logp` frameworks since down the line we can expect changes to `logpt` function when `joint_logprob` supports the transforms being passed to it. (And the difference between the logic for handling one thing or other will increase)

pymc3/distributions/logprob.py
```diff
@@ -141,8 +129,89 @@ def logpt(
...
+        rv_values = (
+            {var: at.as_tensor_variable(rv_values).astype(var.type)}
+            if rv_values is not None
+            else {}
+        )
+
+    tmp_rv_values = {}
+    for node in io_toposort(graph_inputs((var,)), (var,)):
+        if isinstance(node.op, RandomVariable):
+            tmp_rv_values[node.out] = node.out
```
kc611: This was required since `joint_logprob` will simply ignore the RandomVariables which are not in the `rvs_to_values` list. Plus the transforms are not fully supported hence we still have to do the replacement logic below. (L-160 to L-167)

pymc3/distributions/logprob.py
```diff
@@ -141,8 +130,78 @@ def logpt(
...
+                curr_var.tag, "observations", getattr(curr_var.tag, "value_var", curr_var)
+            )
+            rv_value = rv_values.get(curr_var, rv_value_var)
+            tmp_rv_values[curr_var] = rv_value
+
+            if hasattr(rv_value_var.tag, "transform"):
+                transform_map[curr_var] = rv_value_var.tag.transform
+
+    transform_opt = TransformValuesOpt(transform_map)
+    logp_var = joint_logprob(var, tmp_rv_values, extra_rewrites=transform_opt, **kwargs)
```
kc611: Made a call to `joint_logprob` using transforms mapping.

requirements-dev.txt
```diff
@@ -1,6 +1,7 @@
 # This file is auto-generated by scripts/generate_pip_deps_from_conda.py, do not modify.
 # See that file for comments about the need/usage of each dependency.
 
+aeppl>=0.0.8
```
twiecki: You need to add this to the conda test and dev env, as this is auto-generated.

pymc3/distributions/continuous.py
```diff
@@ -765,7 +728,7 @@ def logp(
         else:
             norm = 0.0
 
-        logp = Normal.logp(value, mu=mu, sigma=sigma) - norm
+        logp = _logprob(normal, [value], *(0, 0, 0, mu, sigma)) - norm
```
ricardoV94:
```suggestion
        logp = _logprob(normal, (value,), None, None, None, mu, sigma) - norm
```
Seems less error prone, same for others below.

pymc3/distributions/continuous.py
```diff
@@ -1040,6 +980,9 @@ def get_mu_lam_phi(
             "mu and lam, mu and phi, or lam and phi."
         )
 
+    def _distr_parameters_for_repr(self):
+        return ["mu", "lam", "alpha"]
+
```
ricardoV94:
```suggestion
```
I might be wrong but I don't think we are using this anymore

pymc3/distributions/distribution.py
```diff
@@ -394,6 +396,13 @@ def _distr_parameters_for_repr(self):
         return []
 
 
+@_logprob.register(NoDistribution)
```
ricardoV94: Not sure about this one. Why did you add it?

I would guess an exception might be better

pymc3/distributions/logprob.py
```diff
@@ -141,8 +130,88 @@ def logpt(
...
+    tmp_rv_values = {}
+    transform_map = {}
+    for node in io_toposort(graph_inputs((var,)), (var,)):
+        if isinstance(node.op, RandomVariable):
+            curr_var = node.out
+            rv_value_var = getattr(
+                curr_var.tag, "observations", getattr(curr_var.tag, "value_var", curr_var)
+            )
+            rv_value = rv_values.get(curr_var, rv_value_var)
+            tmp_rv_values[curr_var] = rv_value
```
ricardoV94:
```suggestion
            tmp_rvs_to_values[curr_var] = rv_value
            tmp_values_to_rvs[rv_value] = curr_var
```
Then you don't need the list of keys thing in the loop below

pymc3/distributions/logprob.py
```diff
@@ -141,8 +130,88 @@ def logpt(
...
+
+
+def logcdfpt(
+    var: TensorVariable,
+    rv_values: Optional[Union[TensorVariable, Dict[TensorVariable, TensorVariable]]] = None,
+    *,
+    jacobian: bool = True,
+    scaling: bool = True,
+    transformed: bool = True,
+    sum: bool = False,
```
ricardoV94:
```suggestion
    scaling: bool = True,
    sum: bool = False,
```

I don't think those 2 ever make sense for cdfs

pymc3/tests/test_model.py
```diff
@@ -26,6 +26,7 @@
 import pytest
 import scipy.sparse as sps
 
+from aeppl import transforms as atr
```
ricardoV94:
```suggestion
from aeppl import transforms
```

pymc3/distributions/transforms.py
```diff
@@ -14,388 +14,126 @@
...
 __all__ = [
-    "Transform",
-    "stick_breaking",
-    "logodds",
-    "interval",
-    "log_exp_m1",
-    "ordered",
-    "log",
-    "sum_to_1",
-    "circular",
```
ricardoV94: We probably want to make the aeppl transforms available here for PyMC3 users

pymc3/distributions/distribution.py
kc611: Ah okay, I added that because the `BART` class wasn't dispatching according to `NoDistribution`s `logp` so had to add that separately like this.

conda-envs/environment-dev-py37.yml
```diff
@@ -3,6 +3,7 @@ channels:
 - conda-forge
 - defaults
 dependencies:
+- aeppl>=0.0.8
```
michaelosthege: Let's set upper bounds on the minor version number and conciously update instead of getting surprised by breaking changes.

pymc3/distributions/bound.py
```diff
@@ -166,6 +166,7 @@ def __new__(
                 raise ValueError("Given dims do not exist in model coordinates.")
 
         lower, upper, initval = cls._set_values(lower, upper, size, shape, initval)
+        distribution.tag.ignore_logprob = True
```
michaelosthege: Why? Can we add a comment?

pymc3/distributions/continuous.py
```diff
@@ -2005,7 +1855,7 @@ class Pareto(BoundedContinuous):
         Scale parameter (m > 0).
     """
     rv_op = pareto
-    bound_args_indices = (1, None)  # lower-bounded by `m`
+    bound_args_indices = (4, None)  # lower-bounded by `m`
```
michaelosthege: Is this on purpose?

pymc3/distributions/continuous.py
```diff
@@ -3294,7 +2976,7 @@ class Triangular(BoundedContinuous):
     """
 
     rv_op = triangular
-    bound_args_indices = (0, 2)  # lower, upper
+    bound_args_indices = (3, 5)  # lower, upper
```
michaelosthege: Short explanation maybe? (same as above)

pymc3/distributions/discrete.py
```diff
@@ -1303,7 +1304,7 @@ def logp(value, psi, theta):
 
         logp_val = at.switch(
             at.gt(value, 0),
-            at.log(psi) + _logp(poisson, value, {}, theta),
+            at.log(psi) + _logprob(poisson, [value], *[0, 0, 0, theta]),
```
michaelosthege: TODO replace 0 with None

pymc3/distributions/distribution.py
```diff
@@ -96,10 +97,11 @@ def _random(*args, **kwargs):
...
             if class_logp:
 
-                @_logp.register(rv_type)
-                def logp(op, var, rvs_to_values, *dist_params, **kwargs):
-                    value_var = rvs_to_values.get(var, var)
-                    return class_logp(value_var, *dist_params, **kwargs)
+                @_logprob.register(rv_type)
+                def logp(op, value_var, *dist_params, **kwargs):
+                    _dist_params = dist_params[3:]
+                    value_var = value_var[0]
```
michaelosthege: If `value_var` is not the `value_var`, it should not be called `value_var` in the first place.
(On one side of the `=` it should say something else than `value_var`.)

pymc3/distributions/distribution.py
```diff
@@ -376,7 +378,7 @@ def __getattr__(self, name):
             raise AttributeError("'NoDistribution' has no attribute '%s'" % name)
         return getattr(self.parent_dist, name)
 
-    def logp(self, x):
+    def logp(x, *inputs):
```
michaelosthege: The existence of this method is not in line with actual distributions where this stuff is done through `pm.logp`.
Can it be dispatched?
**EDIT: ** It looks like it _is_ dispatched below?

Also the docstring does not account for all parameters.

pymc3/distributions/distribution.py
```diff
@@ -394,6 +396,13 @@ def _distr_parameters_for_repr(self):
         return []
 
 
+@_logprob.register(NoDistribution)
+def logp(op, value_var, *dist_params, **kwargs):
+    _dist_params = dist_params[3:]
+    value_var = value_var[0]
```
michaelosthege: Same thing with the `value_var` variable name. (mypy would also complain about it, because overwriting the variable name changes the type...)

pymc3/distributions/logprob.py
```diff
@@ -141,8 +130,86 @@ def logpt(
...
+        )
+
+    # Recompute test values for the changes introduced by the replacements
+    # above.
+    if config.compute_test_value != "off":
+        for node in io_toposort(graph_inputs((logp_var,)), (logp_var,)):
+            compute_test_value(node)
+
+    if var.name is not None:
+        logp_var.name = "__logp_%s" % var.name
```
michaelosthege:
```suggestion
    # Recompute test values for the changes introduced by the replacements above.
    if config.compute_test_value != "off":
        for node in io_toposort(graph_inputs((logp_var,)), (logp_var,)):
            compute_test_value(node)

    if var.name is not None:
        logp_var.name = f"__logp_{var.name}"
```

pymc3/tests/test_distributions_random.py
```diff
@@ -1614,23 +1615,21 @@ def test_errors(self):
...
+        #         "matnormal",
+        #         mu=np.random.random((3, 3)),
+        #         rowcov=np.eye(3),
+        #         colcov=np.eye(3),
+        #     )
+        #     with pytest.raises(ValueError, match=msg):
+        #         logpt(
+        #             matrixnormal,
+        #             aesara.tensor.ones((3, 3, 3))
+        #         )
```
michaelosthege: remove?

twiecki:
> Once the aeppl integration is done, refactor Mixture to return a graph in a form expected by joint_logprob i.e. using at.stack

I think we can push this to a future PR, as `Mixture` hasn't been refactored to begin with.

twiecki:
https://github.com/aesara-devs/aeppl/pull/68#pullrequestreview-763353534 is merged now.

kc611:
I'm currently using this branch for local development: https://github.com/kc611/aeppl/tree/keep_transformed_values
Which is an rebased derivative of https://github.com/aesara-devs/aeppl/pull/67
along with a fix for https://github.com/aesara-devs/aeppl/issues/65

So we'd need those two things to go through before this can be implemented.

pymc/tests/test_distributions.py
```diff
@@ -2673,21 +2678,21 @@ def test_continuous(self):
...
         assert logpt(InfBoundedNormal, 11).eval() != -np.inf
 
-        assert logpt(LowerNormalTransform, -1).eval() != -np.inf
+        # TODO: Investigate evaluation issue in aeppl over here.
+        # This passed since passed value to aeppl is simply a TensorConstant
+        value = at.as_tensor_variable(-1)
+        assert logpt(LowerNormalTransform, {LowerNormalTransform: value}).eval() != -np.inf
+        # This fails because passed value to aeppl is a graph
+        # (which eventually valuates to a TensorConstant)
         assert logpt(UpperNormalTransform, 1).eval() != -np.inf
```
kc611: Is this an `aeppl` constraint ?

pymc/tests/test_transforms.py
```diff
@@ -300,15 +306,18 @@ def check_transform_elementwise_logp(self, model):
     def check_vectortransform_elementwise_logp(self, model, vect_opt=0):
         x = model.free_RVs[0]
         x0 = x.tag.value_var
-        assert (x.ndim - 1) == logpt(x).ndim
+        # TODO: For some reason the ndim relations
+        # dont hold up here. But final log-probablity
+        # values are what we expected.
+        # assert (x.ndim - 1) == logpt(x, sum=False).ndim
```
kc611, line 307: This might need further investigation. The tests do pass for the value checks below but not for these dimension checks.

kc611:
Marked ready for review since now the tests do pass on this particular branch https://github.com/pymc-devs/pymc3/pull/4887#issuecomment-927109122

pymc/distributions/bart.py
```diff
@@ -138,3 +140,24 @@ def __new__(
     @classmethod
     def dist(cls, *params, **kwargs):
         return super().dist(params, **kwargs)
+
+    def logp(x, *inputs):
+        """Calculate log probability.
+
+        Parameters
+        ----------
+        x: numeric
```
michaelosthege: Don't all other logps take symbolic x too?

pymc/distributions/logprob.py
```diff
@@ -141,8 +141,109 @@ def logpt(
         A scaling term to apply to the generated log-likelihood graph.
     transformed
         Apply transforms.
-    cdf
-        Return the log cumulative distribution.
+
```
michaelosthege, line 158: The new kwarg is missing from the docstring

pymc/distributions/logprob.py
```diff
@@ -141,8 +141,109 @@ def logpt(
...
+    for node in io_toposort(graph_inputs((var,)), (var,)):
+        if isinstance(node.op, RandomVariable):
+            curr_var = node.out
+            rv_value_var = getattr(
+                curr_var.tag, "observations", getattr(curr_var.tag, "value_var", curr_var)
+            )
+            rv_value = rv_values.get(curr_var, rv_value_var)
+            tmp_rvs_to_values[curr_var] = rv_value
+            # if hasattr(rv_value_var.tag, "transform") and transformed:
+            #     transform_map[rv_value] = rv_value_var.tag.transform
```
michaelosthege: ToDo

pymc/distributions/logprob.py
```diff
@@ -141,8 +141,109 @@ def logpt(
...
+    )
+
+    logp_var_dict = {}
+    for value_var, _logp in temp_logp_var_dict.items():
+        if value_var in rv_values.values():
+            logp_var_dict[value_var] = _logp
+
+    if not logp_var_dict:
+        logp_var = None
+
```
michaelosthege: It would be great to have comments explaining what these next code blocks are about. (Also for blocks above.)

pymc/distributions/logprob.py
```diff
@@ -141,8 +141,109 @@ def logpt(
...
+        else:
+            logp_var = at.add(*logp_var_dict.values())
+
+    if scaling:
+        logp_var *= _get_scaling(
+            getattr(var.tag, "total_size", None), rv_value.shape, rv_value.ndim
+        )
+
+    # Recompute test values for the changes introduced by the replacements
+    # above.
```
michaelosthege:
```suggestion
    # Recompute test values for the changes introduced by the replacements above.
```

pymc/distributions/logprob.py
```diff
@@ -141,8 +141,109 @@ def logpt(
...
+    if config.compute_test_value != "off":
+        for node in io_toposort(graph_inputs((logp_var,)), (logp_var,)):
+            compute_test_value(node)
+
+    if var.name is not None:
+        logp_var.name = "__logp_%s" % var.name
```
michaelosthege:
```suggestion
        logp_var.name = f"__logp_{var.name}"
```

pymc/distributions/logprob.py
```diff
@@ -141,8 +141,109 @@ def logpt(
...
+    Parameters
+    ==========
+    var
+        The `RandomVariable` output that determines the log-likelihood graph.
+    rv_values
+        A variable, or ``dict`` of variables, that represents the value of
+        `var` in its log-likelihood.  If no `rv_value` is provided,
+        ``var.tag.value_var`` will be checked and, when available, used.
+    jacobian
+        Whether or not to include the Jacobian term.
```
michaelosthege: To include it where?

Also is this kwarg even used somewhere?

pymc/distributions/simulator.py
```diff
@@ -217,14 +217,12 @@ def __new__(
...
 
-        # @_logp.register(rv_type)
-        @_logp.register(SimulatorRV)
-        def logp(op, sim_rv, rvs_to_values, *sim_params, **kwargs):
-            value_var = rvs_to_values.get(sim_rv, sim_rv)
-            return cls.logp(
-                value_var,
-                sim_rv,
-            )
+        # @_logprob.register(rv_type)
```
michaelosthege: ToDo

pymc/distributions/transforms.py
```diff
@@ -35,371 +39,128 @@
...
+
+
+stick_breaking = StickBreaking()
+logodds = LogOddsTransform()
+interval = IntervalTransform
+log_exp_m1 = LogExpM1()
+ordered = Ordered()
+log = LogTransform()
+sum_to_1 = SumTo1()
+circular = CircularTransform()
```
michaelosthege, line 166: Maybe add an alias, or a placeholder class `Transform` for printing migration instructions?

pymc/tests/test_distributions.py
```diff
@@ -1154,6 +1154,9 @@ def test_wald_logp_custom_points(self, value, mu, lam, phi, alpha, logp):
         decimals = select_by_precision(float64=6, float32=1)
         assert_almost_equal(model.fastlogp(pt), logp, decimal=decimals, err_msg=str(pt))
 
+    @pytest.mark.xfail(
+        reason="Fails because mu and sigma values are being picked randomly from domains"
+    )
```
michaelosthege, line 1179: From the text here I don't understand what this means.

Can you open a corresponding issue?

pymc/tests/test_distributions.py
```diff
@@ -2673,21 +2678,21 @@ def test_continuous(self):
         assert logpt(InfBoundedNormal, 0).eval() != -np.inf
         assert logpt(InfBoundedNormal, 11).eval() != -np.inf
 
-        assert logpt(LowerNormalTransform, -1).eval() != -np.inf
+        # TODO: Investigate evaluation issue in aeppl over here.
+        # This passed since passed value to aeppl is simply a TensorConstant
```
michaelosthege: ToDo (convert to issue?)

pymc/tests/test_distributions_random.py
```diff
@@ -1614,23 +1615,22 @@ def test_errors(self):
...
+        #         "matnormal",
+        #         mu=np.random.random((3, 3)),
+        #         rowcov=np.eye(3),
+        #         colcov=np.eye(3),
+        #     )
+        #     with pytest.raises(ValueError, match=msg):
+        #         logpt(
+        #             matrixnormal,
+        #             aesara.tensor.ones((3, 3, 3))
+        #         )
```
michaelosthege: Another ToDo

pymc/tests/test_missing.py
```diff
@@ -98,7 +98,7 @@ def test_interval_missing_observations():
         assert "theta1_observed_interval__" in model.named_vars
         assert "theta1_missing_interval__" in model.named_vars
         assert isinstance(
-            model.rvs_to_values[model.named_vars["theta1_observed"]].tag.transform, Interval
+            model.rvs_to_values[model.named_vars["theta1_observed"]].tag.transform, interval
```
michaelosthege, line 100: `interval` is a type? That's unexpected from a lowercase name =/

pymc/tests/test_transforms.py
```diff
@@ -220,7 +226,7 @@ def transform_params(x, z=a, y=b):
         close_to_logical(vals < b, True, tol)
 
 
-@pytest.mark.skipif(aesara.config.floatX == "float32", reason="Test fails on 32 bit")
+@pytest.mark.xfail(reason="This test produces infinite values using aeppl")
```
michaelosthege, line 224: Does that mean the circular transform is broken?

pymc/distributions/logprob.py
kc611: Specifically which new `kwarg` are you talking about ?

pymc/distributions/logprob.py
kc611: `jacobian` is being passed onto `factorized_joint_logprob` below. The functionality was recently added into `aeppl`.

pymc/tests/test_distributions.py
kc611: There's already an issue running regarding this: https://github.com/pymc-devs/pymc3/issues/5023

pymc/tests/test_distributions.py
kc611: Maybe. Otherwise the better way to do this would probably be to use a corresponding value variable while constructing the log probability graph and then passing the actual values onto the `.eval()`. Rather than directly passing those value.

pymc/tests/test_missing.py
kc611: It already is the case, no ?

https://github.com/pymc-devs/pymc3/blob/8ae296bdf3ee58850824c9997293803d2df958ec/pymc/distributions/transforms.py#L224

It's just that the actual name of the class went from `Interval` to `IntervalTransform`

pymc/distributions/simulator.py
kc611: Oh this one can just be removed. I didn't know what the significance of this line https://github.com/pymc-devs/pymc3/blob/8ae296bdf3ee58850824c9997293803d2df958ec/pymc/distributions/simulator.py#L220 was. I just changed it to it's new corresponding form.

pymc/distributions/bart.py
kc611: Yes they do, I'll change this doc-string accordingly.

pymc/distributions/logprob.py
kc611: This particular case was because `factorized_joint_logprob` returns an empty dictionary when there are no logp graph where-as PyMC expects it to be `None`. Hence this shortcut. Regardless you're right, I should probably add such information as comments.

pymc/tests/test_transforms.py
kc611: This was discussed during the call. We should probably open an issue regarding this.

pymc/distributions/logprob.py
```diff
@@ -141,8 +141,108 @@ def logpt(
...
+            )
+            rv_value = rv_values.get(curr_var, rv_value_var)
+            tmp_rvs_to_values[curr_var] = rv_value
+            if hasattr(rv_value_var.tag, "transform") and transformed:
+                transform_map[rv_value] = rv_value_var.tag.transform
+        # The condition below is a hackish way of excluding the value variable for the
+        # RV being indexed in case of Advanced Indexing of RVs. It gets added by the
+        # logic above but aeppl does not expect us to include it in the dictionary of
+        # {RV:values} given to it.
+        if isinstance(node.op, subtensor_types):
```
kc611, line 218: CC @ricardoV94

pymc/distributions/logprob.py
michaelosthege: The sum kwarg

kc611:
@ricardoV94 has a suggestion over here to rename the `logpt` function to something like `joint_logpt` since now it supports generation of joint log likelihood graphs of a list of variables.

pymc/distributions/bart.py
```diff
@@ -138,3 +140,24 @@ def __new__(
...
+        Parameters
+        ----------
+        x: numeric, TensorVariable
+            Value for which log-probability is calculated.
+
+        Returns
+        -------
+        TensorVariable
+        """
+        return at.zeros_like(x)
```
ricardoV94: Can this raise a NotImplementedError?

twiecki:
Ping @kc611, what's still missing here?

kc611:
I just rebased the changes on latest main branch. We still need https://github.com/aesara-devs/aeppl/pull/67 for this to run but other than that most code based changes are done.

cc @ricardoV94 Do you see any issue with this particular implementation with respect to the latest changes you're making in `aeppl`. (The `Subtensor` handling for instance)

ricardoV94:
> I just rebased the changes on latest main branch. We still need [aesara-devs/aeppl#67](https://github.com/aesara-devs/aeppl/pull/67) for this to run but other than that most code based changes are done.
> 
> cc @ricardoV94 Do you see any issue with this particular implementation with respect to the latest changes you're making in `aeppl`. (The `Subtensor` handling for instance)

I don't think so. We don't manually handle observed values, and that's where they were being used in the PyMC codebase I think. Did you run the tests on that branch after my changes to the incSubtensor?

pymc/distributions/logprob.py
ricardoV94: Can you explain a bit of what you are doing here? Where there duplicated variables in graphs with IncSubtensors?

pymc/distributions/transforms.py
ricardoV94: Also add an alias to the hold `stick_breaking`, with a deprecation warning (will have to subclass Simplex I think)

pymc/tests/test_distributions.py
ricardoV94:
```suggestion
    @pytest.mark.xfail(
        reason="Fails due to brittle parametrization triggering numpy domain check in RandomVariable, see #5023"
    )
```

pymc/tests/test_distributions.py
```diff
@@ -2066,7 +2069,7 @@ def test_lkj(self, x, eta, n, lp):
     def test_dirichlet(self, n):
         self.check_logp(Dirichlet, Simplex(n), {"a": Vector(Rplus, n)}, dirichlet_logpdf)
 
-    @pytest.mark.parametrize("dist_shape", [1, (2, 1), (1, 2), (2, 4, 3)])
+    @pytest.mark.parametrize("dist_shape", [(1, 2), (2, 4, 3)])
```
ricardoV94, line 2095: Can we revert these changes?

pymc/tests/test_distributions.py
```diff
@@ -1161,7 +1164,7 @@ def test_beta_logp(self):
         self.check_logp(
             Beta,
             Unit,
-            {"mu": Unit, "sigma": Rplus},
+            {"mu": Domain([0, 0.5, 1]), "sigma": Domain([0, 0.1, 1])},
```
ricardoV94: No point in changing things if it's still an x-fail

pymc/tests/test_model.py
```diff
@@ -323,11 +323,11 @@ def test_aesara_switch_broadcast_edge_cases_1(self):
         # Known issue 1: https://github.com/pymc-devs/pymc/issues/4389
         data = pm.floatX(np.zeros(10))
         with pm.Model() as m:
-            p = pm.Beta("p", 1, 1)
+            p = pm.Beta("p", 1, 1, transform=None)
```
ricardoV94: Let's revert this one

pymc/tests/test_transforms.py
ricardoV94: Let's open an issue to track this one

pymc/tests/test_transforms.py
```diff
@@ -266,7 +266,7 @@ def test_chain_vector_transform():
     check_vector_transform(chain_tranf, UnitSortedVector(3))
 
 
-@pytest.mark.xfail(condition=(aesara.config.floatX == "float32"), reason="Fails on float32")
+@pytest.mark.xfail(reason="Fails due to precision issue. Values just close to expected.")
```
ricardoV94, line 270: Same here

pymc/tests/test_transforms.py
ricardoV94: Let's open an issue for tracking this

pymc/tests/test_distributions.py
ricardoV94: I think so. For instance constants may have the same hash (no idea if they do), which would lead to issues in our dict output. Anyway this seems fine.

pymc/tests/test_transforms.py
kc611: This failure happens with current `aeppl` version. We should probably open up an issue for this. The output values are losing precision somewhere along the chain transform. (It's a values deviation of about `10^-6` from expected)

pymc/tests/test_distributions.py
kc611: No, essentially we don't have the shortcut which returned the same value in the `Stickbreaking`/`Simplex` transform in `aeppl` anymore for a particular case. And I don't think we'd should be doing that until we absolutely need to.  

https://github.com/pymc-devs/pymc/blob/a8b475425f3cfe226043242cf0551eaeaf224cb2/pymc/distributions/transforms.py#L289-L292

pymc/tests/test_transforms.py
ricardoV94: This is expected for transformations whose inverse is not completely undone by the aesara rewrites. Do open an issue. I wanted to investigate it, but didn't have the time yet

kc611:
Any reason why windows tests aren't picking up changes in `conda-env` files ? (They don't seem to be installing `aeppl` at all)

twiecki:
> Any reason why windows tests aren't picking up changes in conda-env files ? (They don't seem to be installing aeppl at all)

For tests we now have conda envs with a test in the name (instead of dev). Those need to be changed too.

twiecki:
https://github.com/pymc-devs/pymc/blob/main/conda-envs/windows-environment-test-py38.yml

pymc/distributions/bart.py
kc611: The tests end up raising that error. (During sampling from a model containing `BART`). Here is an exampe stack trace

```
pymc/sampling.py:500: in sample
    step = assign_step_methods(model, step, step_kwargs=kwargs)
pymc/sampling.py:219: in assign_step_methods
    return instantiate_steppers(model, steps, selected_steps, step_kwargs)
pymc/sampling.py:143: in instantiate_steppers
    step = step_class(vars=vars, model=model, **args)
pymc/bart/pgbart.py:184: in __init__
    self.likelihood_logp = logp(initial_values, [model.datalogpt], vars, shared)
pymc/model.py:825: in datalogpt
    obs_factors = logpt(self.observed_RVs, obs_values)
pymc/distributions/logprob.py:228: in logpt
    tmp_rvs_to_values, extra_rewrites=transform_opt, use_jacobian=jacobian, **kwargs
/home/kc611/anaconda3/envs/pymcv4/lib/python3.7/site-packages/aeppl/joint_logprob.py:190: in factorized_joint_logprob
    **kwargs,
/home/kc611/anaconda3/envs/pymcv4/lib/python3.7/functools.py:840: in wrapper
pymc/bart/bart.py:181: in logp
    return BART.logp(value_var, *_dist_params)
pymc/bart/bart.py:163: NotImplementedError()
```

Is this preventable ?

kc611:
Note that most of the test failures in Ubuntu's `float32` systems are actually optimization errors and not code failures. Everything that fails with a `AttributeError: 'numpy.bool_' object has no attribute 'type'` actually runs but fails a certain optimization along the way. (And `pytest` detects it as a code failure.) 

There's already an issue for that: https://github.com/aesara-devs/aesara/issues/616

The only concerning thing that remains is `pymc/tests/test_posteriors.py::TestSliceUniform` on `float32` systems which actually fails with `AssertionError`s. So far I haven't pinned down the exact issue with that test.

twiecki:
@kc611 now that https://github.com/aesara-devs/aesara/issues/616 is fixed, is this done? Maybe we mark that slice test with xfail and move on?

kc611:
Yeah sure.

kc611:
Looks like one of the subclass's tests of `pymc/tests/test_posteriors.py::TestSliceUniform` passes while others don't. I can't put the `xfail` decorator in subclass's test since there are other class's tests being derived from it. Is there anyway we can temporarily relax the strict `xpass` condition for that particular test ? 

Edit: Got around it by temporarily failing the test for that particular class, the failure can be removed once the entire test is fixed. A bit crude method butsince the logic is actually passing on other systems such as float64 linux and windows systems any logical errors arising from other changes/other PRs will be caught there so I don't see any issues with failing it like this temporarily for a specific subsystem. 

Any other solutions are also welcome.

twiecki:
@kc611 Tests are passing 🥳 can we merge?

pymc/sampling_jax.py
```diff
@@ -83,6 +84,18 @@ def perform(self, node, inputs, outputs):
         raise NotImplementedError()
 
 
+@jax_funcify.register(Assert)
+def jax_funcify_Assert(op, **kwargs):
```
ricardoV94, line 31: Not a blocker, but shouldn't this go into Aesara?

pymc/tests/test_distributions.py
```diff
@@ -945,17 +946,16 @@ def test_triangular(self):
...
             assert np.all(logp(valid_dist, np.array([-1, 2])).eval() == -np.inf)
-            assert np.all(logcdf(valid_dist, np.array([-1, 2])).eval() == [-np.inf, 0])
+            assert np.all(logcdf(valid_dist, np.array([-1, 2]), sum=False).eval() == [-np.inf, 0])
 
-        # Custom logp / logcdf check for invalid parameters
+        # Custom logcdf check for invalid parameters.
+        # Invalid logp checks for triangular are being done in aeppl
         invalid_dist = Triangular.dist(lower=1, upper=0, c=0.1)
         with aesara.config.change_flags(mode=Mode("py")):
-            assert logp(invalid_dist, 0.5).eval() == -np.inf
```
ricardoV94, line 979: What happened to these conditions?

ricardoV94:
Just a reminder this PR should include a test for https://github.com/pymc-devs/pymc/issues/5007

I can push something later if you are too busy now @kc611

pymc/tests/test_distributions.py
kc611: Giving invalid parameters to a distribution now raises an `AsserionError` from `aeppl` (From the `assert` statements in each `logp` so this will no longer give infinite values). Plus the invalid values are being tested in `aeppl` so I didn't see a point adding such a test in PyMC.

pymc/sampling_jax.py
kc611: That is supposed to be just a temporary pass-through for `Assert` Op since Jax doesn't allow asserts in it's code. Though you're right we should be adding this functionality (maybe with a good workaround) in Aesara, I'll open an issue for this.

pymc/tests/test_logprob.py
```diff
@@ -199,19 +174,29 @@ def test_logp_helper():
...
+
+
+def test_logprob_nochange_inplace():
+    with Model() as model:
+        a = Normal("a")
+        c = Uniform("c", lower=a - 1, upper=1)
+
+    d = c.owner.inputs[3]
+    model.logpt
+    assert d == c.owner.inputs[3]
```
kc611: @ricardoV94 is this is right way to test that issue ?

ricardoV94: ~I think that's good, it fails on main locally!~

I would add a comment referencing the relevant Github issue for context

ricardoV94: Nevermind, it's not that as it passes on main

ricardoV94: I pushed a test that passes now and fails on main

ricardoV94:
Pushed a test that works and rebased from main. Apologies if I messed up anything @kc611!

conda-envs/environment-dev-py37.yml
```diff
@@ -4,6 +4,7 @@ channels:
 - conda-forge
 - defaults
 dependencies:
+- aeppl>=0.0.13
```
OriolAbril, line 7: Disclaimer: I skimmed the discussion and have not reviewed the code

What does this dependency mean for the users? Will using aeppl functions/methods be necessary? Or will it be handled under the hood and advanced users will be able to use it if they want?

This is a minor-medium concern for the beta release, but aeppl seems to be very little documented. I was unable to find the _link_ to it's docs anywhere. I also did check and saw there was a `gh-pages` branch so I went to https://aesara-devs.github.io/aeppl/ which has some api docs. We need to work on aeppl docs if we expect pymc **users** to work with it directly before releasing stable 4.0

ricardoV94: Users don't need to know about aeppl, it will be used by us developers to create more complex distributions.

Indeed we need documentation building on aeppl, we have a PR opened for that, but we welcome help. All our methods and functions are quite well documented on the other hand

twiecki:
Can we merge?

ricardoV94 (approved):

ricardoV94:
Pulled the trigger, 🤞

Great work @kc611

PR #4896, opened 2021-08-01
Add black to requirements

canyon289:
Black is required for, useful, dev


canyon289:
I guess theres some other magic going on so ill just leave this be

Issue #4819, opened 2021-06-28
Iterating a `MultiTrace` only yields samples from chain 0


helmutsimon:
In the code below, I run a simple example model (the details of the model don't seem relevant to the problem). There should be 2000 draws, so why does the array `xvars`  have only 1000? In contrast, tracetab.trace_to_dataframe results in a dataframe with 2000 rows, as I would expect. My system details are:
Numpy  1.16.3
PyMC3  3.11.0
Theano  1.1.0
sys.version_info(major=3, minor=7, micro=3, releaselevel='final', serial=0)

```
import sys
import numpy as np
import pymc3
from pymc3.distributions.discrete import Poisson, Categorical
from pymc3.sampling import sample
from pymc3 import Model
from pymc3.backends import tracetab
import theano


def run_MCMC(n, obs, draws):
    obs = np.array(obs)
    with Model() as model: 
        
        pn = np.ones(n) / n
        x = Categorical('x', pn, testval=0)
        sfs_obs = Poisson('sfs_obs', mu=x + 1, observed=obs) 

    with model:
        trace = sample(draws, tune=0, return_inferencedata=False) 
    return trace

print('Numpy ', np.__version__)
print('PyMC3 ', pymc3.__version__)
print('Theano ', theano.__version__)
print(sys.version_info)
obs = 3
draws = 1000
n = 10
trace = run_MCMC(n, obs, draws)
xvars = [t['x'] for t in trace]
print(len(xvars))
print(xvars[:10])
trace_df = tracetab.trace_to_dataframe(trace)
print(trace_df.shape)
```

Output is:
Numpy  1.16.3
PyMC3  3.11.0
Theano  1.1.0
sys.version_info(major=3, minor=7, micro=3, releaselevel='final', serial=0)
Multiprocess sampling (2 chains in 2 jobs)
CategoricalGibbsMetropolis: [x]

 100.00% [2000/2000 00:02<00:00 Sampling 2 chains, 0 divergences]
Sampling 2 chains for 0 tune and 1_000 draw iterations (0 + 2_000 draws total) took 11 seconds.
1000
[3, 6, 7, 0, 7, 4, 1, 1, 1, 1]
(2000, 1)

michaelosthege:
I don't know by heart what an iterator like `for t in trace` gives you, but it looks like the `np.shape(xvar)` might be `(1000, 2)`?

[...]

PR #4900, opened 2021-08-03
Raise Not_Implemented_Error for iterating over t in trace (Issue #4819)

helmutsimon:
MultiTrace does not implement an `__iter__` method, but for some reason you can still do for `t in trace`.  This iteration yields iteration over the last chain `chain[-1]` only. Since`MultiTrace` will ultimately be deprecated in favour of `InferenceData`, this change is resolved by raising a `NotImplementedError`.

See issue #4819.

### DIFF
--- a/pymc3/backends/base.py
+++ b/pymc3/backends/base.py
@@ -541,8 +541,7 @@ class MultiTrace:
             If a chain is not given, the highest chain number is used.
         """
         if chain is None:
-            chain = self.chains[-1]
-        return self._straces[chain].point(idx)
+            raise NotImplementedError
 
     def points(self, chains=None):
         """Return an iterator over all or some of the sample points


pymc3/backends/base.py
```diff
@@ -541,8 +541,7 @@ def point(self, idx, chain=None):
             If a chain is not given, the highest chain number is used.
         """
         if chain is None:
-            chain = self.chains[-1]
```
ricardoV94, line 544: This looks like it will break a lot of things internally... it removes all the functionality of the `point` method.

More generally, by the looks of it, it was actually a deliberate design choice to iterate over the last chain in a MultiTrace. Therefore, we cannot just make it illegal without providing an alternative interface / making changes in the methods that depend on this functionality.

Just for fun I'll let the tests run to see how quickly it fails.

ricardoV94 (changes requested):

ricardoV94:
> MultiTrace does not implement an `__iter__` method, but for some reason you can still do for `t in trace`.

I think this work because we have a `__getittem__` that accepts integers and a `__len__`. Otherwise `__iter__` would probably take precedence (and we could try add a `NotImplementedError` there explicity)

helmutsimon:
Sorry about that. I hope the update is better.

ricardoV94:
@helmutsimon did you push the update to Github? I think we are still seeing your first commit here

helmutsimon:
The revised changes seem to have been successfully pushed to the feature branch of  [my fork of the repository](https://github.com/helmutsimon/pymc3/blob/iter_issue_4819/pymc3/backends/base.py). My understanding is that github should automatically update the PR? Should I just cancel this PR and raise a new one?

MarcoGorelli:
> My understanding is that github should automatically update the PR? Should I just cancel this PR and raise a new one?

not sure what's happened here, normally GitHub would let you click on "from helmutsimon:iter_issue_4819", but that doesn't seem to be allowed here - I think I'd suggest the oldest trick in tech (turn it off and on again, i.e. close and open a new PR)

PR #4902, opened 2021-08-03
Condition xfails on Windows after bug fix affected linux only

michaelosthege:
A fix for https://github.com/aesara-devs/aesara/issues/390 was rolled out with Aesara 2.1.3 but apparently it did not fix the problem on Windows.

The pipeline on `main` started failing right after the release, while the commits on PyMC3 `main` were just typo fixes:
grafik

Unfortunately I don't have much time to spare, so this PR just limits the scope of the XFAIL marks to make the CI pass again.

CC @brandonwillard in case you have an idea while the issue may still persist.

### DIFF
--- a/pymc3/tests/test_ode.py
+++ b/pymc3/tests/test_ode.py
@@ -173,7 +173,7 @@ class TestSensitivityInitialCondition:
         np.testing.assert_array_equal(np.ravel(model5_sens_ic), model5._sens_ic)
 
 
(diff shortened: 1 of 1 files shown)


canyon289 (approved):

canyon289:
Sorry closed by accident. Reopened

ricardoV94 (approved):
Do we have an open issue to track this?

PR #4903, opened 2021-08-04
Refactor `pm.Simulator`

ricardoV94:
Third attempt at refactoring `pm.Simulator`.

Apparently `cloudpickle` overcomes the limitations that we were facing in previous iterations of this PR in #4802 and #4877

### DIFF
--- a/pymc3/aesaraf.py
+++ b/pymc3/aesaraf.py
@@ -350,15 +350,24 @@ def rvs_to_value_vars(
 
     """
 
+    # Avoid circular dependency
+    from pymc3.distributions import NoDistribution
+
     def transform_replacements(var, replacements):
         rv_var, rv_value_var = extract_rv_and_value_vars(var)
 
         if rv_value_var is None:
-            warnings.warn(
-                f"No value variable found for {rv_var}; "
-                "the random variable will not be replaced."
-            )
-            return []
+            # If RandomVariable does not have a value_var and corresponds to
+            # a NoDistribution, we allow further replacements in upstream graph
+            if isinstance(rv_var.owner.op, NoDistribution):
+                return rv_var.owner.inputs
+
+            else:
+                warnings.warn(
+                    f"No value variable found for {rv_var}; "
+                    "the random variable will not be replaced."
+                )
+                return []
 
         transform = getattr(rv_value_var.tag, "transform", None)
 
--- a/pymc3/aesaraf.py
+++ b/pymc3/aesaraf.py
@@ -895,6 +895,22 @@ def compile_rv_inplace(inputs, outputs, mode=None, **kwargs):
     Using this function ensures that compiled functions containing random
     variables will produce new samples on each call.
     """
+
+    # Avoid circular dependency
+    from pymc3.distributions import NoDistribution
+
+    # Set the default update of a NoDistribution RNG so that it is automatically
+    # updated after every function call
+    output_to_list = outputs if isinstance(outputs, list) else [outputs]
+    for rv in (
+        node
+        for node in walk_model(output_to_list, walk_past_rvs=True)
+        if node.owner and isinstance(node.owner.op, NoDistribution)
+    ):
+        rng = rv.owner.inputs[0]
+        if not hasattr(rng, "default_update"):
+            rng.default_update = rv.owner.outputs[0]
+
     mode = get_mode(mode)
     opt_qry = mode.provided_optimizer.including("random_make_inplace")
     mode = Mode(linker=mode.linker, optimizer=opt_qry)
--- a/pymc3/distributions/simulator.py
+++ b/pymc3/distributions/simulator.py
@@ -14,112 +14,248 @@
 
 import logging
 
+import aesara
+import aesara.tensor as at
 import numpy as np
 
+from aesara.graph.op import Apply, Op
+from aesara.tensor.random.op import RandomVariable
+from aesara.tensor.var import TensorVariable
 from scipy.spatial import cKDTree
 
+from pymc3.aesaraf import floatX
 from pymc3.distributions.distribution import NoDistribution
+from pymc3.distributions.logprob import _logp
 
 __all__ = ["Simulator"]
 
 _log = logging.getLogger("pymc3")
 
 
+class SimulatorRV(RandomVariable):
+    """
+    Base class for SimulatorRVs
+
+    This should be subclassed when defining custom Simulator objects.
+    """
+
+    name = "SimulatorRV"
+    ndim_supp = None
+    ndims_params = None
+    dtype = "floatX"
+    _print_name = ("Simulator", "\\operatorname{Simulator}")
+
+    fn = None
+    _distance = None
+    _sum_stat = None
+    epsilon = None
+
+    @classmethod
+    def rng_fn(cls, *args, **kwargs):
+        return cls.fn(*args, **kwargs)
+
+    @classmethod
+    def distance(cls, *args, **kwargs):
+        return cls._distance(*args, **kwargs)
+
+    @classmethod
+    def sum_stat(cls, *args, **kwargs):
+        return cls._sum_stat(*args, **kwargs)
+
+
 class Simulator(NoDistribution):
     r"""
-    Define a simulator, from a Python function, to be used in ABC methods.
+    Simulator distribution, used for Approximate Bayesian Inference (ABC)
+    with Sequential Monte Carlo (SMC) sampling via ``pm.sample_smc``.
+
+    Simulator distributions have a stochastic pseudo-loglikelihood defined by
+    a distance metric between the observed and simulated data, and tweaked
+    by a hyper-parameter ``epsilon``.
 
     Parameters
     ----------
-    function: callable
-        Python function defined by the user.
+    fn: callable
+        Python random simulator function. Should expect the following signature
+        ``(rng, arg1, arg2, ... argn, size)``, where rng is a ``numpy.random.RandomStream()``
+        and ``size`` defines the size of the desired sample.
     params: list
-        Parameters passed to function.
-    distance: str or callable
-        Distance functions. Available options are "gaussian" (default), "laplacian",
-        "kullback_leibler" or a user defined function that takes epsilon, the summary statistics of
-        observed_data and the summary statistics of simulated_data as input.
-        ``gaussian`` :math: `-0.5 \left(\left(\frac{xo - xs}{\epsilon}\right)^2\right)`
-        ``laplace`` :math: `{\left(\frac{|xo - xs|}{\epsilon}\right)}`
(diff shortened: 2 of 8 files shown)


pymc3/distributions/simulator.py
```diff
@@ -14,112 +14,238 @@
...
+        )()
+
+        # The logp function is registered to the more general SimulatorRV,
+        # in order to avoid issues with multiprocessing / pickling
+
+        # rv_type = type(sim_op)
+        # NoDistribution.register(rv_type)
+        NoDistribution.register(SimulatorRV)
+
+        # @_logp.register(rv_type)
```
ricardoV94, line 220: @lucianopaz, I cannot register it on the new `rv_type`, or I get a weird failure on multiprocessing, where the logp defaults to the standard `at.zeros_like(value)`. 

This causes the first Simulator test to fail, complaining that the input variables are not used in the compiled function (which is true because the default `logp` is based only on `value`)

 The strange thing, is that when I run it on Jupyter, it fails in the first time the cell is run but passes the second time.

pymc3/distributions/simulator.py
```diff
@@ -14,112 +14,238 @@
...
+            f"Simulator_{name}",
+            (SimulatorRV,),
+            dict(
+                name="Simulator",
+                ndim_supp=ndim_supp,
+                ndims_params=ndims_params,
+                dtype=dtype,
+                inplace=False,
+                fn=fn,
+                _distance=distance,
```
ricardoV94, line 207: Ideally I wouldn't have to pass `distance`, `sum_stat` or `epsilon` via the `SimulatorRV`, but the dispatching is not working very well yet, see comment below

ricardoV94:
I still have an issue with the logp function of models with Simulator variables. The upstream input RandomVariables persist in the final compiled function, even though they don't contribute at all to the logp term: https://gist.github.com/ricardoV94/11e0bba211b746d369252a51773cf429

They seem to persist in the graph of the `sim_value_rng` despite being properly replaced in the graph of `sim_value`

This might by why the Simulator is performing slightly worse than in V3.

@brandonwillard do you have any idea?

brandonwillard:
> I still have an issue with the logp function of models with Simulator variables. The upstream input RandomVariables persist in the final compiled function, even though their output is completely disconnected: https://gist.github.com/ricardoV94/11e0bba211b746d369252a51773cf429
> 
> They seem to persist in the graph of the `sim_value_rng` even though they are properly replaced in the graph of `sim_value`
> 
> This might by why the Simulator is performing slightly worse than in V3.
> 
> @brandonwillard do you have any idea?

It looks like it's adding the shared RNG states implicitly created by `a`, `b`, and `s` as outputs to the compiled `m.logpt` graph, and I believe this has to do with shared variable updates.  Those RNG state inputs have a `default_update` attribute (e.g. `a.owner.inputs[0].default_update`) that's probably being picked up by `aesara.function` and adding them to the compiled function's outputs.

In other words, `aesara.function` could be automatically adding things to its set of `updates` (i.e. shared variables that it needs to update), and those updates require samples from `a` and `b` in order to update the shared RNG state outputs.

If you add `no_default_updates=True` to the `aesara.function` call, you probably won't see those extra outputs; however, we should think about the repercussions of doing that, and perhaps adjust something in Aesara, if need be.

ricardoV94:
> It looks like it's adding the shared RNG states implicitly created by `a`, `b`, and `s` as outputs to the compiled `m.logpt` graph, and I believe this has to do with shared variable updates. Those RNG state inputs have a `default_update` attribute (e.g. `a.owner.inputs[0].default_update`) that's probably being picked up by `aesara.function` and adding them to the compiled function's outputs.

I updated my gist where I try to manually recreate this sort of shared rngs with default update and then replace one of the inputs, and it seems to not have the same issue: https://gist.github.com/ricardoV94/11e0bba211b746d369252a51773cf429

> If you add `no_default_updates=True` to the `aesara.function` call, you probably won't see those extra outputs; however, we should think about the repercussions of doing that, and perhaps adjust something in Aesara, if need be.

That indeed removes them from the graph, but also removes the automatic updates for the inserted RandomVariable, so it's not what I need.

ricardoV94:
@brandonwillard, if I add the RandomVariable to replacements in `pymc3.aesaraf.rvs_to_value_vars`, the compiled logp function no longer has the unecessary upstream RVs:

```python
def transform_replacements(var, replacements):
    rv_var, rv_value_var = extract_rv_and_value_vars(var)

    if rv_value_var is None:
        # If RandomVariable does not have a value_var and corresponds to
        # a SimulatorRV, we allow further replacements in upstream graph
        if isinstance(rv_var.owner.op, SimulatorRV):
            replacements[rv_var] = rv_var  # << ADDED LINE
            # First 3 inputs are just rng, dtype, and size, which don't
            # need to be replaced.
            return rv_var.owner.inputs[3:]
```

```python
import numpy as np
import aesara
import pymc3 as pm

def normal_sim(rng, a, b, size):
    return rng.normal(a, b, size=size)

data = np.random.normal(loc=0, scale=1, size=10)

with pm.Model() as m:
    a = pm.Normal("a", mu=0, sigma=1)
    b = pm.HalfNormal("b", sigma=1)
    s = pm.Simulator("s", normal_sim, a, b, observed=5)

a.owner.outputs[0].name = "a_rng"
b.owner.outputs[0].name = "b_rng"
s.owner.outputs[0].name = "s_rng"

f = aesara.function([a.tag.value_var, b.tag.value_var], m.logpt)
aesara.dprint(f)
```
```
Sum{acc_dtype=float64} [id A] '__logp'   6
 |MakeVector{dtype='float64'} [id B] ''   5
   |Elemwise{Composite{(i0 * (i1 + (-sqr(i2))))}} [id C] '__logp_a'   1
   | |TensorConstant{0.5} [id D]
   | |TensorConstant{-1.8378770664093453} [id E]
   | |a [id F]
   |Elemwise{Composite{(Switch(Cast{int8}(GE(i0, i1)), (i2 + (i3 * sqr(i0))), i4) + i5)}}[(0, 0)] [id G] '__logp_b'   3
   | |Elemwise{exp,no_inplace} [id H] ''   0
   | | |b_log__ [id I]
   | |TensorConstant{0.0} [id J]
   | |TensorConstant{-0.2257913526447274} [id K]
   | |TensorConstant{-0.5} [id L]
   | |TensorConstant{-inf} [id M]
   | |b_log__ [id I]
   |Elemwise{Composite{(i0 * sqr((i1 - i2)))}}[(0, 2)] [id N] '__logp_s'   4
     |TensorConstant{-0.5} [id L]
     |TensorConstant{5.0} [id O]
     |Simulator_rv{0, (0, 0), floatX, True}.1 [id P] 'sim_value'   2
       |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F383DD42120>) [id Q]
       |TensorConstant{[]} [id R]
       |TensorConstant{11} [id S]
       |a [id F]
       |Elemwise{exp,no_inplace} [id H] ''   0
Simulator_rv{0, (0, 0), floatX, True}.0 [id P] ''   2
```

However, `pymc3.aesaraf.join_nonshared_inputs` stops working for this type of graphs, probably due to a conflict between the `default_update` and `aesara.clone_replace` which is called inside (although it did work before with the more inefficient graph):

```python
(new_datalogpt,), inarray = pm.aesaraf.join_nonshared_inputs(m.initial_point, [m.logpt], m.value_vars, {})
new_f = aesara.function([inarray], new_datalogpt)
aesara.dprint(new_f)
```
```
MissingInputError: Input 0 (b_log__) of the graph (indices start from 0), used to compute Elemwise{exp,no_inplace}(b_log__), was not provided and not given a value. Use the Aesara flag exception_verbosity='high', for more information on this error.
```

Two steps forward, one step backward...

pymc3/aesaraf.py
```diff
@@ -882,6 +893,19 @@ def compile_rv_inplace(inputs, outputs, mode=None, **kwargs):
...
+    # it is automatically updated after every function call
+    output_to_list = outputs if isinstance(outputs, list) else [outputs]
+    for rv in (
+        node
+        for node in walk_model(output_to_list, walk_past_rvs=True)
+        if node.owner and isinstance(node.owner.op, RandomVariable)
+    ):
+        rng = rv.owner.inputs[0]
+        if not hasattr(rng, "default_update"):
+            rng.default_update = rv.owner.outputs[0]
```
ricardoV94, line 912: This is a WIP hack that seems to fix the issues I was finding in this PR, until I get to the bottom of the problem.

In the meantime and just to be sure, is it a bad idea to set the `default_updates` here before compilation. It requires a full graph transversal, but also means we could remove this code here: https://github.com/pymc-devs/pymc3/blob/c913709d02deab1307df89245782630d8f25d953/pymc3/distributions/distribution.py#L307-L322

CC @brandonwillard

ricardoV94:
I added a hack to delay the specification of `default_update` for the shared RandomState variables until the function is actually compiled. This avoids the issue where old RVs remain in the logp graph, upstream of the `Simulator` rng output.

Now this might not be something we want to do just for the sake of the `Simulator` in which case I can create a specialized `compile_rv_inplace`. However, this means that the `Simulator` will only work with `SMC`, since other samplers will keep calling the old `compile_rv_inplace`.

For the good news, the slowdown I was observing relative to V3 is gone!

***

Edit: This gist shows the limitation of `aesaraf.rvs_to_value_vars` when the model is supposed to contain a RandomVariable. Put simply the RVs in the `default_update` graphs are not being replaced by value vars: https://gist.github.com/ricardoV94/11e0bba211b746d369252a51773cf429

ricardoV94:
I created a (semi-)minimal example of the problem where I'm stuck at: https://gist.github.com/ricardoV94/08badc6c501652559900fdcb276e7e50

I need to figure out what is needed for `aesaraf.join_nonshared_inputs` to cope with RV models where the shared RNGs have a `default_update` set.

This is pretty much the last thing blocking the Simulator refactor

pymc3/distributions/simulator.py
```diff
@@ -14,112 +14,244 @@
...
+
 class Simulator(NoDistribution):
     r"""
-    Define a simulator, from a Python function, to be used in ABC methods.
+    Simulator pseudo-distribution, used for Approximate Bayesian Inference (ABC)
+    with Sequential Monte Carlo (SMC) sampling via ``pm.sample_smc``.
+
+    Simulator distributions have a stochastic pseudo-loglikelihood defined by
+    a distance metric between the observed data and simulated data, and tweaked
+    by a hyper-parameter ``epsilon``.
```
aloctavodia:
```suggestion
    Simulator distribution, used for Approximate Bayesian Inference (ABC)
    with Sequential Monte Carlo (SMC) sampling via ``pm.sample_smc``.

    Simulator distributions have a stochastic pseudo-loglikelihood defined by
    a distance metric between the observed and simulated data, and tweaked
    by a hyper-parameter ``epsilon``.
```

aloctavodia (approved):

PR #4906, opened 2021-08-05
arma RV and arma dist [WIP]

mjhajharia:
This pull request adds the ARMA distribution to Time series. 

So everything is done. I have tested the dist function, the generation is working fine because i have tested it with statsmodels arma.fit, the logp method has some undetected bug, because the parameter estimation isn't right. I really appreciate any help or feedback


pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,76 @@ def logp(self, value):
...
+        else:
+            phi = Normal.dist(0.0, 2.0, size=order[0])
+        if theta is not None:
+            theta = at.as_tensor_variable(floatX(theta))
+        else:
+            theta = Normal.dist(0.0, 2.0, size=order[1])
+        return super().dist([phi, theta], size=size, **kwargs)
+
+    def logp(Y, phi, theta, loc, scale):
+        P=2
```
mjhajharia: p,q values were only given when i was testing one of the loops

mjhajharia:
hi! to anyone reviewing hold on, i haven't pushed an important commit with many changes

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,76 @@ def logp(self, value):
...
+        Y_lagged = shared(np.zeros((T,P)))
+        for i in range(P):
+            Y_lagged = at.set_subtensor(Y_lagged[(i+1):,i], Y[:-(i+1)])
+
+        mu = at.fill(Y, loc) + at.dot(Y_lagged, at.transpose(phi))
+        epsilon = at.sub(Y,mu)
+
+        epsilon_lagged = shared(np.zeros((T,Q)))
+        for j in range(Q):
+            epsilon_lagged = at.set_subtensor(epsilon_lagged[j:-1,j], epsilon[:-(j+1)])
```
OriolAbril: indexing here is wrong 

```suggestion
            epsilon_lagged = at.set_subtensor(epsilon_lagged[(j+1):,j], epsilon[:-(j+1)])
```

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,76 @@ def logp(self, value):
...
+            theta = at.as_tensor_variable(floatX(theta))
+        else:
+            theta = Normal.dist(0.0, 2.0, size=order[1])
+        return super().dist([phi, theta], size=size, **kwargs)
+
+    def logp(Y, phi, theta, loc, scale):
+        T = Y.shape[0].eval()
+        P = phi.shape[0].eval()
+        Q = theta.shape[0].eval()
+        Y_lagged = shared(np.zeros((T,P)))
```
OriolAbril: now that it works and is commited, I'd try removing the `eval()`s and using aesara for the zeros matrix too, see also comment below

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,76 @@ def logp(self, value):
...
+        for i in range(P):
+            Y_lagged = at.set_subtensor(Y_lagged[(i+1):,i], Y[:-(i+1)])
+
+        mu = at.fill(Y, loc) + at.dot(Y_lagged, at.transpose(phi))
+        epsilon = at.sub(Y,mu)
+
+        epsilon_lagged = shared(np.zeros((T,Q)))
```
OriolAbril: here too numpy->aesara 2/3

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,76 @@ def logp(self, value):
...
+        for j in range(Q):
+            epsilon_lagged = at.set_subtensor(epsilon_lagged[j:-1,j], epsilon[:-(j+1)])
+
+        mu = mu + at.dot(epsilon_lagged, at.transpose(theta))
+
+        epsilon = at.sub(Y,mu)
+        
+        logp = at.sum(pm.logp(Normal.dist(mu=loc, sigma=scale,shape=int(T)), epsilon))
```
OriolAbril: here `T` as a symbolic aesara (no eval) should work, so no `int`, and in v4, shape is only present for backward compatibility reasons, all the code in the library should use `size`

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,75 @@ def logp(self, value):
         return at.sum(innov_like) + at.sum(init_like)
 
 
+class ARMARV(RandomVariable):
+    name = "ARMA"
+    ndim_supp = 0
```
michaelosthege, line 189: The output of this Op will be a vector, right? Then the ndim_supp should be be 1, because no matter what it will have at least that one dimension.

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,75 @@ def logp(self, value):
...
+    ndims_params = [1, 1, 0, 0]
+    dtype = "floatX"
+    _print_name = ("ARMA", "\\operatorname{ARMA}")
+
+    def __call__(self, phi=[], theta=[], loc=[0.0], scale=[1.0], size=100, **kwargs) -> TensorVariable:
```
michaelosthege: Lists (or dicts) as default kwargs can be dangerous, because they are mutable. Popular alternatives are tuples or None.

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,75 @@ def logp(self, value):
...
+
+    @classmethod
+    def dist(cls, P=None, Q=None, phi=None, theta=None, loc=0.0, scale=1.0,size=100, **kwargs):
+        if (phi is None and P is None) or (theta is None and Q is None):
+            raise ValueError('Specify (phi or P) and (theta or Q)')
+
+        if phi is not None:
+            phi = at.as_tensor_variable(floatX(phi))
+        if phi is None:
+            phi = Uniform("phi",-0.99, 0.99,shape=P)
```
michaelosthege: This 0.99 looks a bit arbitrary. Should it maybe be configurable?

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,75 @@ def logp(self, value):
...
+        return signal.lfilter(theta, phi, x, axis=0)
+
+
+ARMArv = ARMARV()
+
+class ARMA(Continuous):
+    rv_op = ARMArv
+
+    @classmethod
+    def dist(cls, P=None, Q=None, phi=None, theta=None, loc=0.0, scale=1.0,size=100, **kwargs):
```
michaelosthege: Are the symbols Q, P, phi, theta commonly used in the literature for this distribution? Are there maybe alternatives that encode the purpose in the name?

Also make sure to describe these parameters in the docstring without assuming familiarity with ARMA.

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,75 @@ def logp(self, value):
...
+            theta = Uniform("theta",-0.99, 0.99, size=Q)
+        return super().dist([phi, theta], loc=loc, scale=scale,size=size, **kwargs)
+    
+    def logp(Y,phi=None, theta=None, loc=None, scale=None):
+        T = Y.shape[0]
+        P = phi.shape[0]
+        Q = theta.shape[0]
+        Y_lagged = at.zeros(shape=(T,P))
+        for i in range(P.eval()):
+            Y_lagged = at.set_subtensor(Y_lagged[(i+1):,i], Y[:-(i+1)])
```
michaelosthege: When you add tests, make sure to test with 0- and 1-lengths inputs. Slicing from the end often breaks in such cases.

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,75 @@ def logp(self, value):
...
+            phi: np.ndarray,
+            theta: np.ndarray,
+            loc: np.ndarray,
+            scale: np.ndarray,
+            size: Tuple[int, ...],
+    ) -> np.ndarray:
+        phi = np.r_[1, -phi]
+        theta = np.r_[1, theta]
+        x = rng.normal(loc=loc, scale=scale, size=size)
+        return signal.lfilter(theta, phi, x, axis=0)
```
michaelosthege: Maybe a high level comment what the lfilter does?

pymc3/distributions/timeseries.py
canyon289: I have the same question, should we just use one set of args and have a docstring or do we want to support both?

For example in the PyMC3 beta distribution we support both `alpha` and `beta` params as well as `mu,` `sigma` but not sure if thats a good idea or not?

Not saying we need a change, just asking the same question

mjhajharia: so there's two sides to this, there's a reason i'm supporting both ways of parametrizing. say a user wants to generate ARMA(2,2) without knowing phi or theta, then P and Q are useful. But if they know phi and theta, and want to actually generate a specific ARMA (2,2) where say phi = [0.25,0.56] and theta = [-0.76,0.87] so both the ways might be useful, and we should leave it upto the user to decide what they want? and phi and theta are commonly used in the literature but, to avoid any risks whatsoever, I'll name the parameters as follows:

P - number(order) of autogressive lags
Q - number(order) of moving-average lags

phi (renamed as AR) - tensor of autoregressive coefficients
theta(renamed as MA) - tensor of moving-average coefficients

I think it would make sense to do this in AR(p) in timeseries.py as well?

ofcourse i might be wrong, let me know if this sounds reasonable?

michaelosthege: Sounds reasonable to support both parametrizations, but I would still consider to name them `lags_autoregressive`, `lags_movingaverage` (`coeff_autoregressive`, `coeff_movingaverage`) or similar. Otherwise consider lowercase letters - I think some linters might complain about uppercase letters.

Also a `, *,` to require kwargs might be a good idea.
One might argue, that it's a little verbose for the model definition, but the time series distribution is probably quite important for the users model, so it should be okay to be more verbose?

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,84 @@ def logp(self, value):
...
+
+    @classmethod
+    def dist(cls, P=None, Q=None, phi=None, theta=None, loc=None, scale=None, *args, **kwargs):
+        if (phi is None and P is None) or (theta is None and Q is None):
+            raise ValueError('Specify (phi or P) and (theta or Q)')
+
+        if phi is not None:
+            phi = at.as_tensor_variable(floatX(phi))
+        if phi is None:
+            phi = Uniform("phi", -0.99, 0.99, shape=P)
```
fonnesbeck, line 231: These (phi and theta) should just be `if ... else` switches.

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,84 @@ def logp(self, value):
...
+            scale: np.ndarray,
+            size: Tuple[int, ...],
+    ) -> np.ndarray:
+
+        phi = np.r_[1, -phi]
+        theta = np.r_[1, theta]
+        if size:
+            x = rng.normal(loc=loc, scale=scale, size=size)
+        else:
+            x = rng.normal(loc=loc, scale=scale, size=100)
```
canyon289, line 213: Why 100? Is there some significance to this number?

pymc3/distributions/timeseries.py
```diff
@@ -176,6 +184,84 @@ def logp(self, value):
...
+
+    def logp(Y, phi=None, theta=None, loc=None, scale=None):
+        T = Y.shape[0]
+        P = phi.shape[0]
+        Q = theta.shape[0]
+        Y_lagged = at.zeros(shape=(T, P))
+        for i in range(P.eval()):
+            Y_lagged = at.set_subtensor(Y_lagged[(i + 1):, i], Y[:-(i + 1)])
+
+        mu = at.fill(Y, loc) + at.dot(Y_lagged, at.transpose(phi))
```
canyon289, line 250: Comments here explaining the AR part and the MA part would be helpful

pymc3/distributions/timeseries.py
mjhajharia: will add all this, thanks a ton @michaelosthege

pymc3/distributions/timeseries.py
mjhajharia: made size compulsory arg and removed 100

pymc3/distributions/timeseries.py
mjhajharia: thanks, will do

PR #4910, opened 2021-08-07
add explicit keyword arg for broadcast_conditions

### LINKED ISSUE #4892: Add explicit keyword arg for broadcast_conditions in dist_math.bound
Add a keyword arg in this method to match docstring and aid in discovery
https://github.com/pymc-devs/pymc3/blob/main/pymc3/distributions/dist_math.py#L61

neelriyer:
**What are the (breaking) changes that this PR makes?**
Resolves #4892.

**Important background, or details about the implementation**
Doesn't really change that much. This is my first contribution to this repository. Let me know what you think.

**Are the changes—especially new features—covered by tests and docstrings?**
Yes they are.

### DIFF
--- a/pymc3/distributions/dist_math.py
+++ b/pymc3/distributions/dist_math.py
@@ -46,7 +46,7 @@ _beta_clip_values = {
 }
 
 
-def bound(logp, *conditions, broadcast_conditions, **kwargs):
+def bound(logp, *conditions, broadcast_conditions=True, **kwargs):
     """
     Bounds a log probability density with several conditions.
     When conditions are not met, the logp values are replaced by -inf.
--- a/pymc3/distributions/dist_math.py
+++ b/pymc3/distributions/dist_math.py
@@ -46,7 +46,7 @@ _beta_clip_values = {
 }
 
 
-def bound(logp, *conditions, **kwargs):
+def bound(logp, *conditions, broadcast_conditions, **kwargs):
     """
     Bounds a log probability density with several conditions.
     When conditions are not met, the logp values are replaced by -inf.
@@ -81,8 +81,6 @@ def bound(logp, *conditions, **kwargs):
             return logp
     except TypeError:  # No model found
         pass
-
-    broadcast_conditions = kwargs.get("broadcast_conditions", True)
 
     if broadcast_conditions:
         alltrue = alltrue_elemwise
--- a/pymc3/distributions/dist_math.py
+++ b/pymc3/distributions/dist_math.py
@@ -46,7 +46,7 @@ _beta_clip_values = {
 }
 
 
-def bound(logp, *conditions, broadcast_conditions=True, **kwargs):
+def bound(logp, *conditions, broadcast_conditions=True, model=None):
     """
     Bounds a log probability density with several conditions.
     When conditions are not met, the logp values are replaced by -inf.
@@ -59,6 +59,7 @@ def bound(logp, *conditions, broadcast_conditions=True, **kwargs):
     logp: float
     *conditions: booleans
     broadcast_conditions: bool (optional, default=True)
+    model: Model (optional, default=None)
         If True, conditions are broadcasted and applied element-wise to each value in logp.
         If False, conditions are collapsed via at.all(). As a consequence the entire logp
         array is either replaced by -inf or unchanged.
@@ -73,14 +74,12 @@ def bound(logp, *conditions, broadcast_conditions=True, **kwargs):
     """
 
     # If called inside a model context, see if bounds check is disabled
-    try:
+    if model!=None:
         from pymc3.model import modelcontext
 
-        model = modelcontext(kwargs.get("model"))
+        model = modelcontext(model)
         if not model.check_bounds:
             return logp
-    except TypeError:  # No model found
-        pass
 
     if broadcast_conditions:
         alltrue = alltrue_elemwise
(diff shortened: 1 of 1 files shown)


ricardoV94:
Thanks @spiyer99 

I wonder if we need to keep the kwargs at all. 

Seems they are used only to check for a model keyword, but I don't think it's used anywhere in the codebase (perhaps in the tests). Meaning we could just always call `modelcontext(None)`

pymc3/distributions/dist_math.py
```diff
@@ -46,7 +46,7 @@
 }
 
 
-def bound(logp, *conditions, **kwargs):
+def bound(logp, *conditions, broadcast_conditions, **kwargs):
```
ricardoV94:
```suggestion
def bound(logp, *conditions, broadcast_conditions=True, **kwargs):
```
That was the old default

ricardoV94 (changes requested):

pymc3/distributions/dist_math.py
neelriyer: Yeah good point I'll add that in

neelriyer:
Hey @ricardoV94 sorry I'm a bit confused. What do I need to do to get this merged? I've added your change [here](https://github.com/pymc-devs/pymc3/pull/4910/commits/ff4c62fe15776e86765a60a751b9ad78a0668db0).

ricardoV94:
> Hey @ricardoV94 sorry I'm a bit confused. What do I need to do to get this merged? I've added your change [here](https://github.com/pymc-devs/pymc3/pull/4910/commits/ff4c62fe15776e86765a60a751b9ad78a0668db0).

I suggest you check if the `kwargs` argument can be removed altogether. See my comment above. Otherwise the PR looks fine

pymc3/distributions/dist_math.py
```diff
@@ -59,6 +59,7 @@ def bound(logp, *conditions, **kwargs):
     logp: float
     *conditions: booleans
     broadcast_conditions: bool (optional, default=True)
+    model: Model (optional, default=None)
```
ricardoV94: Is the `model` kwarg used anywhere? If not, I would remove it altogether. If it crops up somewhere it's probably in the tests, and that's not a good enough reason to add / maintain the extra keyword.

pymc3/distributions/dist_math.py
```diff
@@ -73,16 +74,12 @@ def bound(logp, *conditions, **kwargs):
     """
 
     # If called inside a model context, see if bounds check is disabled
-    try:
+    if model!=None:
         from pymc3.model import modelcontext
 
-        model = modelcontext(kwargs.get("model"))
+        model = modelcontext(model)
```
michaelosthege: pylint will complain about the `!= None` comparison. Always use `is None` or `is not None` instead of `==`/`!=`.

After already establishing that there is a model, the `modelcontext` line can be removed. See [`def modelcontext`](https://github.com/pymc-devs/pymc3/blob/main/pymc3/model.py#L252-L264).

But what should happen when this function is called inside a modelcontext without passing the model object?
By doing
```
try:
    model = modelcontext(model)
except TypeError:
    # no model found
```
you could make that a little more explicit.

neelriyer: Thanks @michaelosthege. I added these changes into the latest commit.

pymc3/distributions/dist_math.py
```diff
@@ -46,7 +46,7 @@
 }
 
 
-def bound(logp, *conditions, **kwargs):
+def bound(logp, *conditions, broadcast_conditions=True, model=None):
```
ricardoV94:
```suggestion
def bound(logp, *conditions, broadcast_conditions=True):
```

pymc3/distributions/dist_math.py
ricardoV94:
```suggestion
```

pymc3/distributions/dist_math.py
```diff
@@ -73,16 +74,15 @@ def bound(logp, *conditions, **kwargs):
...
-    broadcast_conditions = kwargs.get("broadcast_conditions", True)
+    if model is not None:
+        try:
+            from pymc3.model import modelcontext
+
+            model = modelcontext(model)
+            if not model.check_bounds:
+                return logp
+        except TypeError:
+            pass # no model found
```
ricardoV94:
```suggestion
    try:
        from pymc3.model import modelcontext

        model = modelcontext(None)
        if not model.check_bounds:
            return logp
    except TypeError:
        pass # no model found
```

ricardoV94 (commented):
This is what I was suggesting. Does it work?

pymc3/distributions/dist_math.py
neelriyer: It works for me. Will commit these changes to the branch. Let me know what you think.

spiyer99:
> This is what I was suggesting. Does it work?

It works for me. I added it to the branch. Let me know what you think.

pymc3/distributions/dist_math.py
ricardoV94: Cool, let me know when you do

neelriyer: @ricardoV94 just committed the changes. Thanks for your help :thumbsup:

ricardoV94 (approved):

ricardoV94:
Thanks @spiyer99!

PR #4913, opened 2021-08-08
Remove initval from `dist()` API and add docstrings, type hints, tests

michaelosthege:
This PR makes `initial_values` a property with docstring and type hints. It also adds some more type hints and docstrings to related functions and properties.

For the `UNSET` constant I added its own type, because inspecting signatures with `UNSET` kwargs (for example `pm.Normal`) results in confusing output, as demonstrated by this example:

```
>>> UNSET = object()
>>>
>>> def say_hello(name=UNSET):
...     print("Hi!" if name is UNSET else f"Hello {name}.")
...
>>>
>>> help(say_hello)
Help on function say_hello in module __main__:

say_hello(name=<object object at 0x0000018BD4226150>)
#                👆
```

With the new repr it just becomes
```
Help on function say_hello in module __main__:

say_hello(name=UNSET)
```

+ [x] ~what are the (breaking) changes that this PR makes?~ None
+ [x] ~[consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples)~
+ [x] ~right before it's ready to merge, mention the PR in the RELEASE-NOTES.md~ not relevant enough

### DIFF
--- a/pymc3/model.py
+++ b/pymc3/model.py
@@ -39,7 +39,6 @@ import numpy as np
 import scipy.sparse as sps
 
 from aesara.compile.sharedvalue import SharedVariable
-from aesara.gradient import grad
 from aesara.graph.basic import Constant, Variable, graph_inputs
 from aesara.graph.fg import FunctionGraph
 from aesara.tensor.random.opt import local_subtensor_rv_lift
@@ -446,7 +445,7 @@ class ValueGradFunction:
             givens.append((var, shared))
 
         if compute_grads:
-            grads = grad(cost, grad_vars, disconnected_inputs="ignore")
+            grads = aesara.grad(cost, grad_vars, disconnected_inputs="ignore")
             for grad_wrt, var in zip(grads, grad_vars):
                 grad_wrt.name = f"{var.name}_grad"
             outputs = [cost] + grads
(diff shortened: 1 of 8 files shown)


pymc3/model.py
```diff
@@ -972,7 +981,18 @@ def initval_to_rvval(value_var, value):
...
-            initval = initval_fn()
+            try:
+                initval = initval_fn()
+            except NotImplementedError as ex:
+                if "Cannot sample from" in ex.args[0]:
+                    # The RV does not have a random number generator.
+                    # Our last chance is to take the test_value.
+                    # Note that this is a workaround for Flat and HalfFlat
+                    # until an initval default mechanism is implemented (#4752).
+                    initval = test_value
```
ricardoV94, line 992:
```suggestion
                if "Cannot sample from" in ex.args[0]:
                    # The RV does not have a random number generator.
                    # Our last chance is to take the test_value.
                    # Note that this is a workaround for Flat and HalfFlat
                    # until an initval default mechanism is implemented (#4752).
                    initval = test_value
                else:
                    raise
```

Don't commit it, I think the else is not aligned!

ricardoV94 (approved):

PR #4914, opened 2021-08-09
Bring back BART to V4 and make it more general

aloctavodia:
This PR brings BART to V4, including some of the features in V3 (unreleased) plus a couple of changes not in V3 in order to make BART more general. For example It should now be possible to create semi-parametric models like:

```python
with pm.Model() as model:
    μ = pm.BART("μ", X, Y)
    α = pm.Normal("α", 0, 1)
    β = pm.Normal("β", 0, 1)
    σ = pm.HalfNormal('σ', 1)
    y = pm.Normal("y", μ+α+β*X, σ, observed=Y)
```

There was a broadcasting problem when mixing NUTS and PGBART, that was actually a bug in [base_hmc](https://github.com/pymc-devs/pymc3/pull/4914/files#diff-f95827a70f0ac1a6ce0c09711ddaf46fe6694ef45fda7f88079994865c3fb4beR101-R103). It was finally fixed in #4952

### DIFF
--- a/pymc3/distributions/bart.py
+++ b/pymc3/distributions/bart.py
@@ -40,11 +40,11 @@ class BARTRV(RandomVariable):
         return super().__new__(cls)
 
     @classmethod
-    def rng_fn(cls, X_new=None, *args, **kwargs):
+    def rng_fn(cls, rng, X_new=None, *args, **kwargs):
         all_trees = cls.all_trees
         if all_trees:
             pred = 0
-            idx = np.random.randint(len(all_trees))
+            idx = rng.randint(len(all_trees))
             trees = all_trees[idx]
             if X_new is None:
                 for tree in trees:
@@ -111,7 +111,7 @@ class BART(NoDistribution):
 
         if split_prior is None:
             split_prior = np.ones(X.shape[1])
-        cls.all_trees = []
+        cls.all_trees = all_trees
 
         bart_op = type(
             f"BART_{name}",
--- a/pymc3/distributions/bart.py
+++ b/pymc3/distributions/bart.py
@@ -70,7 +70,7 @@ bart = BARTRV()
 
 class BART(NoDistribution):
     """
-    BART distribution.
+    Bayesian Additive Regression Tree distribution.
 
     Distribution representing a sum over trees
 
@@ -85,13 +85,13 @@ class BART(NoDistribution):
     alpha : float
         Control the prior probability over the depth of the trees. Even when it can takes values in
         the interval (0, 1), it is recommended to be in the interval (0, 0.5].
-    k : int
+    k : float
         Scale parameter for the values of the leaf nodes. Defaults to 2. Recomended to be between 1
         and 3.
     split_prior : array-like
-        Each element of split_prior should be in the [0, 1] interval and the elements should sum
-        to 1. Otherwise they will be normalized.
-        Defaults to None, i.e. all covariate have the same prior probability to be selected.
+        Each element of split_prior should be in the [0, 1] interval and the elements should sum to
+        1. Otherwise they will be normalized.
+        Defaults to None, i.e. all covariates have the same prior probability to be selected.
     """
 
     def __new__(
@@ -109,29 +109,10 @@ class BART(NoDistribution):
         **kwargs,
     ):
 
-        if X.ndim != 2:
-            raise ValueError("The covariate matrix X must have two dimensions")
-
-        if Y.ndim != 1:
-            raise ValueError("The response vector Y must have one dimension")
-        if X.shape[0] != Y.shape[0]:
-            raise ValueError(
-                "The covariate matrix X and the response vector Y must have the same number of elements"
-            )
-        if not isinstance(m, int):
-            raise ValueError("The number of trees m type must be an integer")
-        if m < 1:
-            raise ValueError("The number of trees m must be greater than zero")
-
-        if alpha <= 0 or 1 <= alpha:
-            raise ValueError(
-                "The value for the alpha parameter for the tree structure "
-                "must be in the interval (0, 1)"
-            )
-
         if split_prior is None:
             split_prior = np.ones(X.shape[1])
         cls.all_trees = []
+
         bart_op = type(
             f"BART_{name}",
             (BARTRV,),
--- a/pymc3/distributions/bart.py
+++ b/pymc3/distributions/bart.py
@@ -15,8 +15,8 @@
 import numpy as np
 
 from aesara.tensor.random.op import RandomVariable, default_shape_from_params
-from pymc3.distributions.distribution import NoDistribution
 
+from pymc3.distributions.distribution import NoDistribution
 
 __all__ = ["BART"]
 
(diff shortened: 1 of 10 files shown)


pymc3/step_methods/pgbart.py
```diff
@@ -275,6 +327,210 @@ def set_particle_to_step(self, t):
...
+        for i in range(count):
+            suma += a[i]
+        return suma / count
+
+    return mean
+
+
+def discrete_uniform_sampler(upper_value):
+    """Draw from the uniform distribution with bounds [0, upper_value)."""
+    return int(np.random.random() * upper_value)
```
ricardoV94, line 499: https://numpy.org/doc/stable/reference/random/generated/numpy.random.randint.html

Seems to do this

aloctavodia: yes, but faster. I can add a comment to the docstring.

pymc3/distributions/bart.py
```diff
@@ -14,271 +14,119 @@
...
+        return super().__new__(cls)
+
+    @classmethod
+    def rng_fn(cls, rng=np.random.default_rng(), X_new=None, *args, **kwargs):
+        all_trees = cls.all_trees
+        if all_trees:
+            # this should be rng.integers() but when sampling from the prior/posterior predictive
+            # I get 'numpy.random.mtrand.RandomState' object has no attribute 'integers'
+            # So I guess those functions need to be updated
+            idx = np.random.randint(len(all_trees))
```
ricardoV94:
```suggestion
            idx = rng.randint(len(all_trees))
```

pymc3/distributions/bart.py
```diff
@@ -14,271 +14,119 @@
...
+    all_trees = None
+
+    def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
+        return default_shape_from_params(self.ndim_supp, dist_params, rep_param_idx, param_shapes)
+
+    def __new__(cls, *args, **kwargs):
+        return super().__new__(cls)
+
+    @classmethod
+    def rng_fn(cls, rng=np.random.default_rng(), X_new=None, *args, **kwargs):
```
ricardoV94: The size argument is not allowed?

pymc3/sampling.py
```diff
@@ -232,13 +233,14 @@ def _print_step_hierarchy(s: Step, level=0) -> None:
...
-def all_continuous(vars):
+def all_continuous(vars, model):
     """Check that vars not include discrete variables or BART variables, excepting observed RVs."""
 
     vars_ = [var for var in vars if not (var.owner and hasattr(var.tag, "observations"))]
+
     if any(
         [
-            (var.dtype in discrete_types or (var.owner and isinstance(var.owner.op, pm.BART)))
+            (var.dtype in discrete_types or isinstance(model.values_to_rvs[var].owner.op, BARTRV))
```
ricardoV94:
```suggestion
            (var.dtype in discrete_types or isinstance(model.values_to_rvs[var].owner.op, NoDistribution))
```
?

pymc3/step_methods/hmc/nuts.py
```diff
@@ -196,8 +196,9 @@ def _hamiltonian_step(self, start, p0, step_size):
     @staticmethod
     def competence(var, has_grad):
         """Check how appropriate this class is for sampling a random variable."""
+
         dist = getattr(var.owner, "op", None)
-        if var.dtype in continuous_types and has_grad and not isinstance(dist, BART):
+        if var.dtype in continuous_types and has_grad and not isinstance(dist, BARTRV):
```
ricardoV94, line 203:
```suggestion
        if var.dtype in continuous_types and has_grad and not isinstance(dist, NoDistribution):
```

aloctavodia:
This is ready to merge @ricardoV94 @twiecki?

pymc3/distributions/bart.py
```diff
@@ -14,271 +14,127 @@
...
+    ndims_params = [2, 1, 0, 0, 0, 1]
+    dtype = "floatX"
+    _print_name = ("BART", "\\operatorname{BART}")
+    all_trees = None
+
+    def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
+        return default_shape_from_params(self.ndim_supp, dist_params, rep_param_idx, param_shapes)
+
+    @classmethod
+    def rng_fn(cls, rng=np.random.default_rng(), *args, **kwargs):
```
ricardoV94, line 40:
```suggestion
    def rng_fn(cls, rng=np.random.default_rng(), *args, size=None, X_new=None, **kwargs):
```
Not sure if it's better, just suggesting

pymc3/distributions/bart.py
```diff
@@ -14,271 +14,127 @@
...
+        size = kwargs.pop("size", None)
+        X_new = kwargs.pop("X_new", None)
```
ricardoV94, line 42:
```suggestion
    def rng_fn(cls, rng=np.random.default_rng(), *args, size=None, X_rew=None, **kwargs):
```

Not sure if it's better, just suggesting it

ricardoV94 (commented):
Looks good, but it seems some important branches in the `rng_fn` are not reached by the tests: https://github.com/pymc-devs/pymc3/pull/4914/files#annotation_1841363883

twiecki (approved):

pymc3/step_methods/pgbart.py
```diff
@@ -56,223 +61,460 @@ class PGBART(ArrayStepShared):
...
+    m,
+    normal,
+    mu_std,
+):
+    current_node = tree.get_node(index_leaf_node)
+
+    index_selected_predictor = ssv.rvs()
+    selected_predictor = available_predictors[index_selected_predictor]
+    available_splitting_values = X[current_node.idx_data_points, selected_predictor]
+    if missing_data:
```
ricardoV94, line 413: Should we add a small test for missing data?

pymc3/step_methods/pgbart.py
```diff
@@ -56,223 +61,460 @@ class PGBART(ArrayStepShared):
...
+        self.alpha = self.bart.alpha
+        self.k = self.bart.k
+        self.split_prior = self.bart.split_prior
+        if self.split_prior is None:
+            self.split_prior = np.ones(self.X.shape[1])
+
+        self.init_mean = self.Y.mean()
+        # if data is binary
+        Y_unique = np.unique(self.Y)
+        if Y_unique.size == 2 and np.all(Y_unique == [0, 1]):
```
ricardoV94, line 82: Is `self.Y` or `np.unique` sorted? Otherwise `[1, 0]` would be missed. 

Also, what about just checking if all unique values are discrete? That would also work for count data as you mention in the comment below...?

In either case, a test for discrete data would probably be good.

aloctavodia: It is sorted. 

My comment about count data is that we may need to check for count data and have some other rule to compute the std. I have some tests saying that this may not be actually necessary, still the comment is there as a reminder until I have more solid evidence.

PR #4919, opened 2021-08-12
SMC kernel refactor

ricardoV94:
This PR refactors the SMC base kernel to facilitate implementing alternative kernels.

Feedback of any kind is more than welcome at this stage!

### DIFF
--- a/pymc3/smc/sample_smc.py
+++ b/pymc3/smc/sample_smc.py
@@ -36,24 +36,22 @@ from pymc3.smc.smc import SMC
 
 def sample_smc(
     draws=2000,
-    kernel="metropolis",
+    kernel=None,
     n_steps=25,
     *,
     start=None,
     tune_steps=True,
     p_acc_rate=0.85,
     threshold=0.5,
-    save_sim_data=False,
-    save_log_pseudolikelihood=True,
     model=None,
     random_seed=-1,
-    parallel=None,
     chains=None,
     cores=None,
     compute_convergence_checks=True,
     return_inferencedata=True,
     idata_kwargs=None,
     progressbar=True,
+    **kwargs,
 ):
     r"""
     Sequential Monte Carlo based sampling.
@@ -63,9 +61,6 @@ def sample_smc(
     draws: int
         The number of samples to draw from the posterior (i.e. last stage). And also the number of
         independent chains. Defaults to 2000.
-    kernel: str
-        Kernel method for the SMC sampler. Available option are ``metropolis`` (default) and `ABC`.
-        Use `ABC` for likelihood free inference together with a ``pm.Simulator``.
     n_steps: int
         The number of steps of each Markov Chain. If ``tune_steps == True`` ``n_steps`` will be used
         for the first stage and for the others it will be determined automatically based on the
@@ -83,13 +78,6 @@ def sample_smc(
         Determines the change of beta from stage to stage, i.e.indirectly the number of stages,
         the higher the value of `threshold` the higher the number of stages. Defaults to 0.5.
         It should be between 0 and 1.
-    save_sim_data : bool
-        Whether or not to save the simulated data. This parameter only works with the ABC kernel.
-        The stored data corresponds to a samples from the posterior predictive distribution.
-    save_log_pseudolikelihood : bool
-        Whether or not to save the log pseudolikelihood values. This parameter only works with the
-        ABC kernel. The stored data can be used to compute LOO or WAIC values. Computing LOO/WAIC
-        values from log pseudolikelihood values is experimental.
     model: Model (optional if in ``with`` context)).
     random_seed: int
         random seed
@@ -157,6 +145,31 @@ def sample_smc(
         %282007%29133:7%28816%29>`__
     """
 
+    if isinstance(kernel, str) and kernel.lower() in ("abc", "metropolis"):
+        warnings.warn(
+            f'The kernel string argument "{kernel}" in sample_smc has been deprecated. '
+            f"It is no longer needed to distinguish between `abc` and `metropolis`",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    if kwargs.pop("save_sim_data", None) is not None:
+        warnings.warn(
+            "save_sim_data has been deprecated. Use pm.sample_posterior_predictive "
+            "to obtain the same type of samples.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    if kwargs.pop("save_log_pseudolikelihood", None) is not None:
+        warnings.warn(
+            "save_log_pseudolikelihood has been deprecated. This information is "
+            "now saved as log_likelihood in models with Simulator distributions.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    parallel = kwargs.pop("parallel", None)
     if parallel is not None:
         warnings.warn(
             "The argument parallel is deprecated, use the argument cores instead.",
@@ -199,22 +212,13 @@ def sample_smc(
     if not isinstance(random_seed, Iterable):
         raise TypeError("Invalid value for `random_seed`. Must be tuple, list or int")
 
-    if kernel.lower() == "abc":
-        if len(model.observed_RVs) != 1:
-            warnings.warn("SMC-ABC only works properly with models with one observed variable")
-        if model.potentials:
-            _log.info("Potentials will be added to the prior term")
-
     params = (
         draws,
-        kernel,
         n_steps,
         start,
         tune_steps,
         p_acc_rate,
         threshold,
-        save_sim_data,
-        save_log_pseudolikelihood,
         model,
     )
 
@@ -245,9 +249,7 @@ def sample_smc(
 
     (
         traces,
-        sim_data,
         log_marginal_likelihoods,
-        log_pseudolikelihood,
         betas,
         accept_ratios,
         nsteps,
@@ -263,7 +265,6 @@ def sample_smc(
         trace.report._n_draws = draws
         trace.report._n_tune = _n_tune
         trace.report.log_marginal_likelihood = log_marginal_likelihoods
-        trace.report.log_pseudolikelihood = log_pseudolikelihood
         trace.report.betas = betas
         trace.report.accept_ratios = accept_ratios
         trace.report.nsteps = nsteps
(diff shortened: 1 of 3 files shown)


pymc3/smc/sample_smc.py
```diff
@@ -110,6 +83,23 @@ def sample_smc(
         Keyword arguments for :func:`pymc3.to_inference_data`
     progressbar : bool, optional default=True
         Whether or not to display a progress bar in the command line.
+    **kernel_kwargs: keyword arguments passed to the SMC kernel.
+        The default IMH kernel takes the following keywords:
+            threshold: float
+                Determines the change of beta from stage to stage, i.e.indirectly the number of stages,
```
aloctavodia:
```suggestion
                Determines the change of beta from stage to stage, i.e. indirectly the number of stages,
```

pymc3/smc/sample_smc.py
```diff
@@ -313,113 +316,88 @@ def sample_smc(
...
-    betas = []
-    accept_ratios = []
-    nsteps = []
 
     if progressbar:
         progressbar.comment = f"{getattr(progressbar, 'base_comment', '')} Stage: 0 Beta: 0"
         progressbar.update_bar(getattr(progressbar, "offset", 0) + 0)
 
-    smc.initialize_population()
+    smc._setup_kernel()
```
aloctavodia: can we use a different name? maybe `_initialize_kernel`?

pymc3/smc/sample_smc.py
```diff
@@ -313,113 +316,88 @@ def sample_smc(
...
+
+        smc.update_beta_and_weights()
+
         if progressbar:
             progressbar.comment = (
                 f"{getattr(progressbar, 'base_comment', '')} Stage: {stage} Beta: {smc.beta:.3f}"
             )
             progressbar.update_bar(getattr(progressbar, "offset", 0) + int(smc.beta * 100))
-        smc.update_proposal()
+
```
aloctavodia, line 374: To update the proposal we should either use the weighted tempered posterior `np.cov(self.population, aweights=self.weights`) (i.e. `self.population` not yet resampled), or the resampled posterior `np.cov(self.population, aweights=None`.  But here both are mixed and we are computing the variance weighting the already resampled particles.
In practice i think the better idea is to use the  weighted tempered posterior and not the resampled posterior.  We should do this to compute both the covariance and the mean of the multivariate normal proposal

pymc3/smc/smc.py
```diff
@@ -35,134 +37,193 @@
...
-    """Sequential Monte Carlo with Independent Metropolis-Hastings and ABC kernels."""
+class SMC_KERNEL(ABC):
+    """Base class for the Sequential Monte Carlo kernels
+
+    To creat a new SMC kernel you should subclass from this.
+
+    Before sampling, the following methods are called once in order:
+
+        initialize_population
+            Choose initial population of SMC particles. Should return a dictonary
```
aloctavodia:
```suggestion
            Choose initial population of SMC particles. Should return a dictionary
```

pymc3/smc/smc.py
```diff
@@ -35,134 +37,193 @@
...
+            with {var.name : numpy array of size (draws, var.size)}. Defaults
+            to sampling from the prior predictive model. This method is only called
```
aloctavodia:
```suggestion
            to sampling from the prior distribution. This method is only called
```

pymc3/smc/smc.py
```diff
@@ -35,134 +37,193 @@
...
+            model variables shape and size as {var.name : (var.shape, var.size)
+
+            The functions self.prior_logp_func and self.likelihood_logp_func are
+            created in this step. These expect a 1D numpy array with the summed
+            sizes of each raveled model variable (in the order specified in
+            model.inial_point).
+
+            Finally, this method computes the initial prior and posterior log
+            likelihoods for the initial particles, and saves them in self.prior_logp
+            and self.likelihood_logp.
```
aloctavodia:
```suggestion
            Finally, this method computes the log prior and  log
            likelihood for the initial particles, and saves them in self.prior_logp
            and self.likelihood_logp.
```

pymc3/smc/smc.py
```diff
@@ -35,134 +37,193 @@
...
+
+            This method should not be modified.
+
+        setup_kernel: optional
+            May include any logic that should be performed once before sampling
```
aloctavodia:
```suggestion
            May include any logic that should be performed before sampling
```

pymc3/smc/smc.py
```diff
@@ -35,134 +37,193 @@
...
+            starts.
+
+    During each sampling stage the following methods are called in order:
+
+        update_beta_and_weights: default
+            The inverse temperature self.beta is updated based on the self.likelihood_logp
+            and `threshold` parameter
+
+            The importance self.weights of each particle are computed from the newly
+            selected inverse temperature
```
aloctavodia:
```suggestion
            The importance self.weights of each particle are computed from the old and newly
            selected inverse temperature
```

pymc3/smc/smc.py
```diff
@@ -35,134 +37,193 @@
...
+        pass
 
-    def update_weights_beta(self):
-        """Calculate the next inverse temperature (beta).
+    def update_beta_and_weights(self):
+        """Calculate the next inverse temperature (beta)
 
-        The importance weights based on current beta and tempered likelihood and updates the
-        marginal likelihood estimate.
+        The importance weights based on current beta and tempered likelihood and
```
aloctavodia:
```suggestion
        The importance weights based on two sucesive tempered likelihoods (i.e. two successive values of beta) and
```

aloctavodia: Or something like this...

pymc3/smc/smc.py
```diff
@@ -184,109 +245,180 @@ def update_weights_beta(self):
...
 
     def resample(self):
-        """Resample particles based on importance weights."""
+        """Resample particles based on importance weights"""
         resampling_indexes = np.random.choice(
             np.arange(self.draws), size=self.draws, p=self.weights
         )
 
-        self.posterior = self.posterior[resampling_indexes]
+        self.population = self.population[resampling_indexes]
```
aloctavodia: isn't `posterior` more clear than population? We already use prior and likelihood.

pymc3/smc/smc.py
```diff
@@ -184,109 +245,180 @@ def update_weights_beta(self):
...
         self.prior_logp = self.prior_logp[resampling_indexes]
         self.likelihood_logp = self.likelihood_logp[resampling_indexes]
-        self.posterior_logp = self.prior_logp + self.likelihood_logp * self.beta
-        if self.save_sim_data:
-            self.sim_data = self.sim_data[resampling_indexes]
+        self.tempered_posterior_logp = self.prior_logp + self.likelihood_logp * self.beta
```
aloctavodia, line 270: do we need to clarify that this is the tempered posterior? If that's the case then my previous comments should be to use `self.tempered_posterior` instead of `self.population`.

pymc3/smc/sample_smc.py
ricardoV94: Definitely

pymc3/smc/sample_smc.py
ricardoV94: Oh, I didn't notice I changed the behavior. You are totally right!

ricardoV94: It should be enough to move the `smc.tune()` before the `smc.resample()` right?

pymc3/smc/smc.py
ricardoV94: My reasoning was that it isn't the posterior until the very last iteration. Whereas prior and likelihood are indeed true throughout the entire sampling

pymc3/smc/smc.py
ricardoV94: We don't need to... let's discuss on the meeting :)

pymc3/smc/sample_smc.py
```diff
@@ -110,6 +83,23 @@ def sample_smc(
...
+                The number of steps of each Markov Chain. If ``tune_steps == True`` ``n_steps`` will be used
+                for the first stage and for the others it will be determined automatically based on the
+                acceptance rate and `p_acc_rate`, the max number of steps is ``n_steps``.
+            tune_steps: bool
+                Whether to compute the number of steps automatically or not. Defaults to True
+            p_acc_rate: float
+                Used to compute ``n_steps`` when ``tune_steps == True``. The higher the value of
+                ``p_acc_rate`` the higher the number of steps computed automatically. Defaults to 0.85.
+                It should be between 0 and 1.
+        For other kernels one should check their specific docstrings for details on specific keword arguments.
```
ricardoV94:
```suggestion
        Keyword arguments for other kernels should be checked in the respective docstrings.
```

pymc3/smc/smc.py
```diff
@@ -184,109 +245,180 @@ def update_weights_beta(self):
...
+        self.proposed = self.proposed = self.draws * self.n_steps
+        self.proposal_dist = None
+        self.acc_rate = None
+        self.first_iteration = True
 
     def tune(self):
-        """Tune n_steps based on the acceptance rate."""
-        if self.tune_steps:
+        # Tune n_steps based on the acceptance rate
+        if self.tune_steps and not self.first_iteration:
```
ricardoV94: Add comment emphasizing we skip in first iteration

pymc3/smc/smc.py
```diff
@@ -184,109 +245,180 @@ def update_weights_beta(self):
...
+        self.first_iteration = False
 
-        self.proposed = self.draws * self.n_steps
+        # Update MVNormal proposal based on the covariance matrix from tempered posterior
+        cov = np.cov(self.population, ddof=0, aweights=self.weights, rowvar=0)
+        cov = np.atleast_2d(cov)
+        cov += 1e-6 * np.eye(cov.shape[0])
+        if np.isnan(cov).any() or np.isinf(cov).any():
+            raise ValueError('Sample covariances not valid! Likely "draws" is too small!')
+        self.proposal_dist = multivariate_normal(self.population.mean(axis=0), cov)
```
ricardoV94: Compute weighted mean instead

junpenglao (approved):

pymc3/smc/smc.py
```diff
@@ -35,134 +37,193 @@
...
         self.random_seed = random_seed
-        self.chain = chain
 
         self.model = modelcontext(model)
 
         if self.random_seed != -1:
             np.random.seed(self.random_seed)
 
+        self.var_info = {}
+        self.population = None
```
ricardoV94:
```suggestion
        self.tempered_posterior = None
```

aloctavodia (approved):

PR #4921, opened 2021-08-13
small edits in documentation and messages

stestoll:
This PR fixes a couple of typos and adds a few small improvements in the documentation.

### DIFF
--- a/pymc3/step_methods/hmc/hmc.py
+++ b/pymc3/step_methods/hmc/hmc.py
@@ -131,13 +131,13 @@ class HamiltonianMC(BaseHMC):
             div_info = DivergenceInfo("Integration failed.", e, last, None)
         else:
             if not np.isfinite(state.energy):
-                div_info = DivergenceInfo("Divergence encountered: infinite energy.", None, last, state)
+                div_info = DivergenceInfo("Divergence encountered, bad energy.", None, last, state)
             energy_change = start.energy - state.energy
             if np.isnan(energy_change):
                 energy_change = -np.inf
             if np.abs(energy_change) > self.Emax:
                 div_info = DivergenceInfo(
-                    f"Divergence encountered: energy change larger than {self.Emax}.", None, last, state
+                    f"Divergence encountered, energy change larger than {self.Emax}.", None, last, state
                 )
 
         accept_stat = min(1, np.exp(energy_change))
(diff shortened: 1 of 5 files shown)


pymc3/step_methods/hmc/hmc.py
```diff
@@ -127,13 +131,13 @@ def _hamiltonian_step(self, start, p0, step_size):
             div_info = DivergenceInfo("Integration failed.", e, last, None)
         else:
             if not np.isfinite(state.energy):
-                div_info = DivergenceInfo("Divergence encountered, bad energy.", None, last, state)
+                div_info = DivergenceInfo("Divergence encountered: infinite energy.", None, last, state)
```
ricardoV94: Although this is a better description, we might want to stick with `bad` as it is easier for users to find out the countless related discourse posts in search engines

ricardoV94 (commented):
Looks great. I just left a comment about the `bad` -> `infinite` change.

pymc3/step_methods/hmc/hmc.py
stestoll: That's a fair point. I changed it because one of my undergrads was confused by the error message, and explaining them that the energy came out infinite cleared it up. But admittedly, that's n=1 user evidence...

stestoll: I reverted the change.

ricardoV94 (approved):
Thanks, the changes look good

ricardoV94:
You seem to have a linting issue here: https://github.com/pymc-devs/pymc3/pull/4921/checks?check_run_id=3319910721#step:4:90

You can install `pre-commit` locally to format the files before pushing them here: https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style

stestoll:
Oops. Sorry, I am new to this. Will fix.

stestoll:
Why does Codecov report a coverage change in files that are unchanged by this PR? Is this just noise from non-deterministic tests?

ricardoV94:
> Why does Codecov report a coverage change in files that are unchanged by this PR? Is this just noise from non-deterministic tests?

The Codecov is impatient and starts reporting changes when some tests are still running. The final codecov should be fine though.

Otherwise, it might be a stochastic test that did not trigger a specific branch or something

ricardoV94:
Thank @stestoll

PR #4923, opened 2021-08-13
adjust definition of energy change in HMC

stestoll:
This small PR adjusts the definition of `energy_change` in HMC such that an energy increase in a proposed step gives `energy_change>0`, and an energy decrease gives `energy_change<0`.

This brings it in line with the definitions of energy change in `nuts.py`[here](https://github.com/pymc-devs/pymc3/blob/6d2aa5ddebed01d81c2ab66b9d4bd02194f82508/pymc3/step_methods/hmc/nuts.py#L337) and [here](https://github.com/pymc-devs/pymc3/blob/6d2aa5ddebed01d81c2ab66b9d4bd02194f82508/pymc3/step_methods/hmc/nuts.py#L420).

### DIFF
--- a/pymc3/step_methods/hmc/hmc.py
+++ b/pymc3/step_methods/hmc/hmc.py
@@ -119,7 +119,7 @@ class HamiltonianMC(BaseHMC):
         n_steps = max(1, int(self.path_length / step_size))
         n_steps = min(self.max_steps, n_steps)
 
-        energy_change = -np.inf
+        energy_change = np.inf
         state = start
         last = state
         div_info = None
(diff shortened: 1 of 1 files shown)


stestoll:
Looks like one of the test runners failed to start. Is there a way to retrigger it?

ricardoV94:
> Looks like one of the test runners failed to start. Is there a way to retrigger it? 

I don't think we need to worry about it. 

We should have coverage from the other runs (or from one of the two flost32/float64 modes)

ColCarroll (approved):
Looks good to me!

ColCarroll:
Thanks, @stestoll!

PR #4928, opened 2021-08-13
Change "Lognormal" to "LogNormal" (camel case)

### LINKED ISSUE #4885: `Lognormal` distribution is not in camel case, unlike all the other distributions
All of the distribution names with two or more parts use camel case (i.e. `DiscreteUniform`, `ZeroInflatedPoisson`, etc).

The exception however is the [`Lognormal`](https://docs.pymc.io/api/distributions/continuous.html#pymc3.distributions.continuous.Lognormal) distribution. 

I propose that this distribution be changed to `LogNormal`.

mjhajharia:
Changes class name (actually creates an alias) and adds a deprecation warning.

fixes #4885


twiecki:
Needs a line in the release notes and should probably  have a `Lognormal` that raises a deprecation error pointing users to `LogNormal`.

pymc3/distributions/continuous.py
```diff
@@ -1776,8 +1776,15 @@ def logp(self, value):
             0 < self.kappa,
         )
 
+class LogNormal(LogNormal):
```
mjhajharia: @twiecki made the alias, but stupidly named the deprecated one LogNormal too, will correct it

fonnesbeck:
Do we really want deprecation warning at all for this? Seems like output noise for something rather harmless that will disappear naturally over time.

fonnesbeck:
Do we really want deprecation warning at all for this? Seems like output noise for something rather harmless that will disappear naturally over time.

mjhajharia:
> Do we really want deprecation warning at all for this? Seems like output noise for something rather harmless that will disappear naturally over time.

yeah I agree, also since its already in conformity with the rest of the pymc nomenclature it will assimilate rather fast. and too many warnings encourage users to silence warnings which is risky

pymc3/distributions/continuous.py
```diff
@@ -1776,8 +1776,15 @@ def logp(self, value):
             0 < self.kappa,
         )
 
+class Lognormal(LogNormal):
+    """deprecated, use LogNormal instead!"""
+    def __init__(self, *args, **kwargs):
+        warnings.warn(
+          "The class name LogNormal is deprecated, use LogNormal instead!",
```
twiecki:
```suggestion
          "The class name Lognormal is deprecated, use LogNormal instead.",
```

mjhajharia:
so for now what i'm gonna do is have no warnings for lognormal and then once we make a decision on using or not using deprecated library as @michaelosthege mentioned for pymc3 in general we can change it, does that sound good?

RELEASE-NOTES.md
```diff
@@ -13,7 +13,7 @@
 ### Maintenance
 - Remove float128 dtype support (see [#4514](https://github.com/pymc-devs/pymc3/pull/4514)).
 - Logp method of `Uniform` and `DiscreteUniform` no longer depends on `pymc3.distributions.dist_math.bound` for proper evaluation (see [#4541](https://github.com/pymc-devs/pymc3/pull/4541)).
-- ...
+- Change name of `Lognormal` distribution to `LogNormal` to synchronize Camel Case usage for distribution names.
```
```suggestion
- Change name of `Lognormal` distribution to `LogNormal` to harmonize CamelCase usage for distribution names.
```

pymc3/distributions/continuous.py
```diff
@@ -1776,8 +1776,15 @@ def logp(self, value):
             0 < self.kappa,
         )
 
+class Lognormal(LogNormal):
+    """deprecated, use LogNormal instead!"""
+    def __init__(self, *args, **kwargs):
+#         warnings.warn(
```
twiecki: I agree that we should just silently allow this for backwards compat. But then the comment can just be removed.

twiecki:
WTH? 
```
ImportError while loading conftest '/home/runner/work/pymc3/pymc3/pymc3/tests/conftest.py'.
pymc3/__init__.py:41: in <module>
    from pymc3 import gp, ode, sampling
pymc3/gp/__init__.py:15: in <module>
    from pymc3.gp import cov, mean, util
pymc3/gp/util.py:24: in <module>
    solve_lower = Solve(A_structure="lower_triangular")
E   TypeError: __init__() got an unexpected keyword argument 'A_structure'
Error: Process completed with exit code 4.
```

pymc3/distributions/continuous.py
mjhajharia: im not sure of how this gets printed in the documentation, but if it does wouldnt the comment make sense for users to to see this in the documentation?

twiecki: The docstring, yes, but I mean the commented warning.

twiecki (approved):

twiecki:
Thanks @mjhajharia !

Sayam753:
Why didn't we merge it to `main` branch? @twiecki

twiecki:
Oh, that's what I wanted to do but didn't see it didn't point there.

mjhajharia:
> Oh, that's what I wanted to do but didn't see it didn't point there.

uhh oh probably because I made the changes on the fix_capitalization branch instead of a local branch in the fork, my bad, do you want me to commit again or can you change it

twiecki:
Reopened as https://github.com/pymc-devs/pymc3/pull/4944.

PR #3006, opened 2018-06-05
re-implement robust invlogit

junpenglao:
rerun LDA ADVI notebook and close #3005
Also makes the error report in VI meanfield more robust

Issue #1147, opened 2016-05-26
ADVI returns NaN with many features


parsing-science:
Hi,

I've been testing out the new ADVI feature the last couple of days. If I only have a small number of features, it seems to produce good values, but if I increase the number of features, I keep getting NaNs. 

Here's a basic logistic regression based on one of your examples: 

```
def invlogit(x):
    return np.exp(x) / (1 + np.exp(x))

def tinvlogit(x):
    return T.exp(x) / (1 + T.exp(x))

# Set up basic parameters
num_features = 7
n = 100000

# Choose random values for the actual alpha and betas
alpha_a = random.normal(size=1)

betas_a = random.normal(size = num_features)

# Create fake predictor data
predictors = random.normal(size=(n, num_features))

# Calculate the outcomes
outcomes = random.binomial(1, invlogit(alpha_a + np.sum(betas_a[None, :] * predictors, 1)))

model = Model()

with model:
    alpha = Normal(b'alpha', mu=0, tau=2.**-2, shape=(1))
    betas = Normal(b'betas', mu=0, tau=2. ** -2, shape=(1, num_features))

    p = tinvlogit(alpha+sum(betas * predictors, 1))

    o = Bernoulli(b'o', p, observed=outcomes)

with model:
    mu, sds, elbo = variational.advi(n=10000)

```

This is the output I get:

```
Iteration 0 [0%]: ELBO = -183969.24
Iteration 1000 [10%]: ELBO = nan
Iteration 2000 [20%]: ELBO = nan
Iteration 3000 [30%]: ELBO = nan
Iteration 4000 [40%]: ELBO = nan
Iteration 5000 [50%]: ELBO = nan
Iteration 6000 [60%]: ELBO = nan
Iteration 7000 [70%]: ELBO = nan
Iteration 8000 [80%]: ELBO = nan
Iteration 9000 [90%]: ELBO = nan
Finished [100%]: ELBO = nan
```

As long as the number of features is fewer than 7, it seems to work ok, but as soon as I go up to 7 or more features, I only get NaNs back. 

Any idea what could be happening? It seems very strange that just going from 6 to 7 features could produce this behavior.

fonnesbeck:
Thanks for reporting this. Actually, I get `nan` lower bounds with 6 features as well.

fonnesbeck:
I can push it a little larger using `T.dot` instead of summing, but it eventually produces `nan`s as well. I suspect we are running into numerical issues.

[...]

PR #4935, opened 2021-08-16
Reintroduce `logit_p` Bernoulli parametrization and deprecate `eps` in `math.invlogit`

ricardoV94:
The function `math.invlogit` had an artificial truncation point to avoid numerical issues that occurred in the past with the bernoulli:
https://github.com/pymc-devs/pymc3/issues/1147
https://github.com/pymc-devs/pymc3/pull/3006

However, there has been some progress on `Aesara` for making the invlogit (and related rewrites) more numerically stable:
https://github.com/aesara-devs/aesara/pull/473
https://github.com/aesara-devs/aesara/pull/531
https://github.com/aesara-devs/aesara/pull/262
https://github.com/aesara-devs/aesara/pull/32

And this no longer seems necessary. After this PR:
```python
import aesara 
import aesara.tensor as at
import pymc3 as pm 

logit_p = at.scalar("logit_p")
value = at.lscalar("value")
logp = pm.logp(pm.Bernoulli.dist(p=pm.math.invlogit(logit_p)), value)
fn = aesara.function([logit_p, value], logp)  

fn(-745, 0)  # array(-5.e-324)
fn(-746, 0)  # array(-0.)  # overflows here as before

fn(745, 1)  # array(-5.e-324)
fn(746, 1)  # array(-0.)  # overflows here, new!!!

fn(1000, 0)  # array(-1000.)  # never underflows

fn(-1000, 1)  # array(-1000.)  # never underflows
```

In v3:
```python
import theano 
import theano.tensor as tt
import pymc3 as pm 

logit_p = tt.scalar("logit_p")                                                     
value = tt.lscalar("value")                                            
logp = pm.Bernoulli.dist(p=pm.math.invlogit(logit_p)).logp(value)            
fn = theano.function([logit_p, value], logp)   

fn(-745, 0)  # array(-5.e-324)
fn(-746, 0)  # array(-0.)  # same overflow point

fn(1000, 1)  # array(-2.22044605e-16)  # never overflows
fn(37, 1)  # array(-2.22044605e-16)  # but value would be flat from >= 37

fn(1000, 0)  # array(-1000.)  # never underflows either

fn(-1000, 1)  # array(-36.04365339)  # never underflows either
fn(-69, 1)  # array(-36.04365339)  # but value would be flat from ~<=-69
```

Regarding the old `logit_p` logp implementation, I ran some benchmarks and there is some small loss in speed from forcing it to be computed on the natural scale, but this is rather small. On the other hand forcing a natural probability to be computed on the logit scale would have been considerably slower: https://gist.github.com/ricardoV94/65a334968472deec65ffc285ae2a809f

### DIFF
--- a/pymc3/distributions/discrete.py
+++ b/pymc3/distributions/discrete.py
@@ -333,8 +333,15 @@ class Bernoulli(Discrete):
 
     @classmethod
     def dist(cls, p=None, logit_p=None, *args, **kwargs):
+        if p is not None and logit_p is not None:
+            raise ValueError("Incompatible parametrization. Can't specify both p and logit_p.")
+        elif p is None and logit_p is None:
+            raise ValueError("Incompatible parametrization. Must specify either p or logit_p.")
+
+        if logit_p is not None:
+            p = at.sigmoid(logit_p)
+
         p = at.as_tensor_variable(floatX(p))
-        # mode = at.cast(tround(p), "int8")
         return super().dist([p], **kwargs)
 
     def logp(value, p):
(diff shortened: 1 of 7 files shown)


twiecki (approved):

twiecki:
This is really cool!

PR #4947, opened 2021-08-18
ARMA RV and ARIMA distribution [WIP]

mjhajharia:
`ARIMA` distribution and Random Variable.

A few notes:

- I think two forms of parametrization are not useful, I remember @canyon289 's comment on this from the old PR (made a new one because my fork was messed up), so `p` and `q` aren't used in `AR` as well, and most importantly, in Bayesian estimation, priors are almost always necessary and we shouldn't use default priors for users in my opinion, also two forms of parametrization might make this model more confusing for beginners as it already has many params. Additionally, for generating an ARMA process it makes no sense to give p, q and get a random phi, theta. So users can use ARMA.dist to generate and ARMA(..` observed=Y`) to estimate phi and theta. 

- Additionally, I have taken @fonnesbeck 's suggestion and made `size` a necessary parameter, because it's essentially very important for time series distributions, so either `size` has to be provided by the user or is inferred from the `observed` array.

- @michaelosthege and @canyon289 had some great suggestions on the previous PR(#4906) and I'll make most of those changes - I'm going with phi and theta instead of `ar_coeff` and `ma_coeff`, because it would be difficult to represent the math in the docstring with `ar_coeff`, so phi makes sense here. also, it's very standard, the wikipedia page and almost all popular literature uses phi and theta. and we can give a simple example in documentation itself so I think we'll be okay. If the example in the comments of this PR looks ok then we can add that itself. It's the simplest way to generate and estimate `ARMA`.

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -18,7 +18,8 @@ from aesara.tensor.var import TensorVariable
 from aesara.tensor.random.op import RandomVariable
 import numpy as np
 
-from scipy import stats
+from aesara import scan
+from scipy import stats, signal
 from typing import List, Tuple
 
 import pymc as pymc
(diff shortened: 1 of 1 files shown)


mjhajharia:
@canyon289 i'm closing this and we can just continue working on the arma branch, once we have something ready we can reopen the PR?

PR #4949, opened 2021-08-18
Reduce sigfigs for acceptance probability warning

stestoll:
This small PR adjusts the NUTS sampler warning that reports an acceptance probability not close enough to the target acceptance probability set in `target_accept`. It reduces the number of sigfigs that are shown from >10 to 4.

Showing more than 4 sigfigs doesn't make sense, given the approximate nature of this comparison.

Somewhat related is the [old discussion on Discourse](https://discourse.pymc.io/t/warning-when-nuts-probability-is-greater-than-acceptance-level/594) about this warning.

Example:

Before:
```
The acceptance probability does not match the target. It is 0.94542317739, but should be close to 0.8.
```

After:
```
The acceptance probability does not match the target. It is 0.9454, but should be close to 0.8.
```

### DIFF
--- a/pymc3/step_methods/step_sizes.py
+++ b/pymc3/step_methods/step_sizes.py
@@ -67,15 +67,15 @@ class DualAverageAdaptation:
         mean_accept = np.mean(accept)
         target_accept = self._target
         # Try to find a reasonable interval for acceptable acceptance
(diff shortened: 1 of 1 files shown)


twiecki (approved):

twiecki:
Thanks @stestoll!

PR #4952, opened 2021-08-19
Fix bug in NUTS variable assignment

aloctavodia:
This is a bug I found while working on bringing BART to V4 #4914. NUTS should use the variables passed to the sampler and not get them from the model. @ricardoV94 recommend having this fix in its own PR and separated from the BART PR.

### DIFF
--- a/pymc3/tests/test_parallel_sampling.py
+++ b/pymc3/tests/test_parallel_sampling.py
@@ -87,7 +87,7 @@ def test_remote_pipe_closed():
             pm.sample(step=step, mp_ctx="spawn", tune=2, draws=2, cores=2, chains=2)
 
 
-@pytest.mark.xfail(reason="Unclear")
+@pytest.mark.skip(reason="Unclear")
 @pytest.mark.parametrize("mp_start_method", ["spawn", "fork"])
 def test_abort(mp_start_method):
     with pm.Model() as model:
--- a/pymc3/tests/test_parallel_sampling.py
+++ b/pymc3/tests/test_parallel_sampling.py
@@ -13,9 +13,11 @@
 #   limitations under the License.
 import multiprocessing
 import os
+import platform
 
 import aesara
 import aesara.tensor as at
+import cloudpickle
 import numpy as np
 import pytest
 
@@ -86,7 +88,8 @@ def test_remote_pipe_closed():
 
 
 @pytest.mark.xfail(reason="Unclear")
-def test_abort():
+@pytest.mark.parametrize("mp_start_method", ["spawn", "fork"])
+def test_abort(mp_start_method):
     with pm.Model() as model:
         a = pm.Normal("a", shape=1)
         b = pm.HalfNormal("b")
@@ -95,8 +98,16 @@ def test_abort():
 
     step = pm.CompoundStep([step1, step2])
 
+    # on Windows we cannot fork
+    if platform.system() == "Windows" and mp_start_method == "fork":
+        return
+    if mp_start_method == "spawn":
+        step_method_pickled = cloudpickle.dumps(step, protocol=-1)
+    else:
+        step_method_pickled = None
+
     for abort in [False, True]:
-        ctx = multiprocessing.get_context()
+        ctx = multiprocessing.get_context(mp_start_method)
         proc = ps.ProcessAdapter(
             10,
             10,
@@ -105,7 +116,7 @@ def test_abort():
             seed=1,
             mp_ctx=ctx,
             start={"a": floatX(np.array([1.0])), "b_log__": floatX(np.array(2.0))},
-            step_method_pickled=None,
+            step_method_pickled=step_method_pickled,
         )
         proc.start()
         while True:
@@ -118,7 +129,8 @@ def test_abort():
         proc.join()
 
 
-def test_explicit_sample():
+@pytest.mark.parametrize("mp_start_method", ["spawn", "fork"])
+def test_explicit_sample(mp_start_method):
     with pm.Model() as model:
         a = pm.Normal("a", shape=1)
         b = pm.HalfNormal("b")
@@ -127,7 +139,15 @@ def test_explicit_sample():
 
     step = pm.CompoundStep([step1, step2])
 
-    ctx = multiprocessing.get_context()
+    # on Windows we cannot fork
+    if platform.system() == "Windows" and mp_start_method == "fork":
+        return
+    if mp_start_method == "spawn":
+        step_method_pickled = cloudpickle.dumps(step, protocol=-1)
+    else:
+        step_method_pickled = None
+
+    ctx = multiprocessing.get_context(mp_start_method)
     proc = ps.ProcessAdapter(
         10,
         10,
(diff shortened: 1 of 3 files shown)


ricardoV94:
We should add a minimal test. This seems to trigger it locally:

```python
import pymc3 as pm
with pm.Model() as m:
    x = pm.Normal("x", 0, 1)
    y = pm.Normal("y", 0, 1)
    pm.sample(step=pm.Metropolis([m.rvs_to_values[x]]), chains=1, tune=10, draws=10)
```

pymc3/tests/test_parallel_sampling.py
```diff
@@ -119,7 +123,8 @@ def test_abort():
 
 
 @pytest.mark.xfail(
-    reason="Possibly the same issue described in https://github.com/pymc-devs/pymc3/pull/4701"
+    condition=(IS_WINDOWS or IS_FLOAT32),
```
ricardoV94: Your fix might have affected what was actually wrong with this test, since the sampler is using NUTS for only one of two continuous variables.

pymc3/tests/test_parallel_sampling.py
```diff
@@ -154,7 +159,8 @@ def test_explicit_sample():
 
 
 @pytest.mark.xfail(
-    reason="Possibly the same issue described in https://github.com/pymc-devs/pymc3/pull/4701"
+    condition=(IS_FLOAT32),
```
ricardoV94: Same here

pymc3/tests/test_parallel_sampling.py
aloctavodia: Yeah, that's why I assumed we do not need a new test.

ricardoV94: I see, you're right. But it should not depend on windows or float32 though

ricardoV94: Let me push a fix, I think it's just a question of converting the start to the right float type

pymc3/tests/test_parallel_sampling.py
```diff
@@ -87,15 +85,13 @@ def test_remote_pipe_closed():
             pm.sample(step=step, mp_ctx="spawn", tune=2, draws=2, cores=2, chains=2)
 
 
-@pytest.mark.xfail(
-    reason="Possibly the same issue described in https://github.com/pymc-devs/pymc3/pull/4701"
-)
+@pytest.mark.xfail(reason="Unclear")
```
ricardoV94: Let's open a separate issue to investigate this one. The other two tests seem to pass after the changes I added

ricardoV94:
Getting this error on the Windows run it `test_parallel_sampling.py::test_explicit_sample`: `AttributeError: 'functools.partial' object has no attribute '__name__'`

https://github.com/pymc-devs/pymc3/pull/4952/checks?check_run_id=3368890696#step:7:75

@Spaak, @michaelosthege do you happen to have any idea? These tests are all related to multiprocessing...

Spaak:
> Getting this error on the Windows run it `test_parallel_sampling.py::test_explicit_sample`: `AttributeError: 'functools.partial' object has no attribute '__name__'`
> 
> https://github.com/pymc-devs/pymc3/pull/4952/checks?check_run_id=3368890696#step:7:75
> 
> @Spaak, @michaelosthege do you happen to have any idea? These tests are all related to multiprocessing...

Yes this is definitely caused by pickling (probably a step method), required when multiprocessing uses spawn (which is the default on Windows; POSIX OSes use fork). Default `pickle` chokes on `functools.partial` objects, which is why we switched to `cloudpickle` (#4858). Could be I missed one/a few instances of pickling that still uses default `pickle`? I can have a look later today.

ricardoV94:
> Could be I missed one/a few instances of pickling that still uses default `pickle`? I can have a look later today.

You might have missed it because the tests were marked as xfail for other reasons (the one addressed in this PR)

pymc3/tests/test_parallel_sampling.py
aloctavodia: Yeah, those fails are unrelated to the fix in this PR.

Spaak:
> You might have missed it because the test was marked as an x-fail for other reasons (the one addressed in this PR)

Yes that's probably what happened. I'll work on a fix now, which I'll file as a separate PR (rather than push commits here) to keep the commit history on main clean (I don't yet know how much code I'll have to touch).

ricardoV94:
> > You might have missed it because the test was marked as an x-fail for other reasons (the one addressed in this PR)
> 
> Yes that's probably what happened. I'll work on a fix now, which I'll file as a separate PR (rather than push commits here) to keep the commit history on main clean (I don't yet know how much code I'll have to touch).

Sure, but if you find it easier, feel free to force push a clean history to this PR. 

The bugfix that is revelant to this PR is really just those two lines that @aloctavodia changed in ` pymc3/step_methods/hmc/base_hmc.py ` in the first commit

Spaak:
Apologies; `pre-commit` has made me lazy and I forgot I was working in an environment that I hadn't installed `pre-commit` in yet. (So I pushed a commit here that failed pre-commit CI.) I'll force-push the same changes (but now properly blacked/isorted etc.) here in a minute. (So no separate PR; it turned out to be a minor change.)

ricardoV94:
@Spaak any chances something similar was needed for this other failing test?

https://github.com/pymc-devs/pymc3/blob/e6ded51174f13ba0033c3cf51c6f3ed0ddc7c3b6/pymc3/tests/test_parallel_sampling.py#L88-L118

Spaak:
@ricardoV94 I did make the same changes to `test_abort`, but it's still failing for another reason, so I left the `xfail`.

ricardoV94:
> @ricardoV94 I did make the same changes to `test_abort`, but it's still failing for another reason, so I left the `xfail`.

Didn't notice. Thanks. 

One of the changed test seems to have failed though

pymc3/tests/test_parallel_sampling.py
```diff
@@ -97,6 +98,9 @@ def test_abort(mp_start_method):
 
     step = pm.CompoundStep([step1, step2])
 
+    # on Windows we cannot fork
+    if platform.system() == "Windows":
+        mp_start_method = "spawn"
```
ricardoV94:
```suggestion
    if platform.system() == "Windows" and mp_start_method == "fork":
        return
```

Spaak: Made the change (I agree it's better) and of course now the test started passing on Windows/fork (return before the error had a chance to happen), resulting in CI complaining because of strict xfail settings...

ricardoV94: Makes sense

ricardoV94 (approved):

ricardoV94:
Oh well... that `test_abort` is still giving problems. Going to simply `skip` it until we figure it out. I did not change anything in my force push: https://github.com/pymc-devs/pymc3/compare/e721e11df92604d448601e97bbd0ad7677bd7735..b504f0c416c91c4f6b3334110b4e2703a1aaaa92

Spaak:
Hmm after your force push the `test_abort` is xpassing: https://github.com/pymc-devs/pymc3/pull/4952/checks?check_run_id=3370627851 which was still xfail before. Did you rebase onto main or so causing it now to succeed..?

ricardoV94:
It fails locally for me (regardless of floatX), so I think it is just a weird fluke

Spaak:
> Going to simply skip it until we figure it out.

Yep good idea

ricardoV94:
Thanks @aloctavodia and @Spaak

PR #4955, opened 2021-08-19
Raise SamplingError if model doesn't have free RVs

stestoll:
This small PR changes `sample()` such that it checks for free RVs in the model immediately (before checking start point), and raises a `SamplingError` instead of a `ValueError` if the model doesn't have any free RVs.

I think raising a `SamplingError` is better than a `ValueError`, since nothing value-wise is wrong with the model itself, just sampling is not possible. Also, `SamplingError` is more specific than `ValueError`.


michaelosthege (approved):
Looks good. The message is also easier to understand.

canyon289:
LGTM

PR #4971, opened 2021-08-29
(Truncated) StickBreakingWeights for Dirichlet Processes

larryshamalama:
Here is my PR on adding stick-breaking weights which would be useful for Dirichlet Processes. For the time being, I am labelling this as a WIP and I have yet to write tests for this distribution.

I am able to call `rng_fn` on the instance `stickbreakingweights` (see output [here](https://github.com/larryshamalama/pymc3-playground/blob/95b6d6f9cb9b47322d9d0c005af2668a72cc0e81/notebooks/progress/stickbreakingweights.ipynb)*), but I obtain `IndexError: list index out of range` when calling `stickbreakingweights(2).eval()`. @ricardoV94 mentioned that this should be fixable via `_shape_from_params` (or `_infer_shape` according to bullet point 3 in section 1 [here](https://github.com/pymc-devs/pymc3/blob/main/docs/source/developer_guide_implementing_distribution.md)), but I don't understand these methods enough to solve my problem. Some comments are included as a code review.

Any help or pointers would be appreciated. Thanks!

CC: @ricardoV94 @AustinRochford @fonnesbeck

*Edited to reflect the version of the notebook when writing this message

### DIFF
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -2107,15 +2107,15 @@ class StickBreakingWeightsRV(RandomVariable):
     dtype = "floatX"
     _print_name = ("StickBreakingWeights", "\\operatorname{StickBreakingWeights}")
 
-    def __call__(self, alpha, size=None, **kwargs):
-        return super().__call__(alpha, size=size, **kwargs)
+    def __call__(self, alpha, K, size=None, **kwargs):
+        return super().__call__(alpha, K, size=size, **kwargs)
 
     def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
         return default_shape_from_params(self.ndim_supp, dist_params, rep_param_idx, param_shapes)
 
     @classmethod
-    def rng_fn(cls, rng, alpha, size):
-        size = tuple(size or ())
+    def rng_fn(cls, rng, alpha, K, size):
+        size = tuple(size or ()) + (K,)
 
         betas = rng.beta(1, alpha, size=size)
         sticks = np.concatenate(
@@ -2142,14 +2142,17 @@ class StickBreakingWeights(Continuous):
     rv_op = stickbreakingweights
 
     @classmethod
-    def dist(cls, alpha, *args, **kwargs):
+    def dist(cls, alpha, K, *args, **kwargs):
         alpha = at.as_tensor_variable(alpha)
+        K = at.as_tensor_variable(K)
 
         assert_negative_support(alpha, "alpha", "StickBreakingWeights")
+        assert_negative_support(K, "K", "StickBreakingWeights")
 
-        return super().dist([alpha], **kwargs)
+        return super().dist([alpha, K], **kwargs)
 
-    def logp(value, alpha):
+    def logp(value, alpha, K):
+        # K does not affect the log-likelihood value to my understanding...
         return bound(
             at.sum(pm.Beta.logp(value, 1, alpha)),
             alpha > 0,
--- a/pymc3/distributions/multivariate.py
+++ b/pymc3/distributions/multivariate.py
@@ -43,7 +43,7 @@ import pymc3 as pm
 
 from pymc3.aesaraf import floatX, intX
 from pymc3.distributions import transforms
-from pymc3.distributions.continuous import ChiSquared, Normal, assert_negative_support
+from pymc3.distributions.continuous import ChiSquared, Normal, assert_negative_support, UnitContinuous
 from pymc3.distributions.dist_math import bound, factln, logpow, multigammaln
 from pymc3.distributions.distribution import Continuous, Discrete
 from pymc3.distributions.shape_utils import broadcast_dist_samples_to, to_tuple
@@ -2103,15 +2103,15 @@ class CAR(Continuous):
 class StickBreakingWeightsRV(RandomVariable):
     name = "stick_breaking_weights"
     ndim_supp = 1
-    ndims_params = [0]
+    ndims_params = [0, 0]
     dtype = "floatX"
     _print_name = ("StickBreakingWeights", "\\operatorname{StickBreakingWeights}")
 
-    def __call__(self, alpha, K, size=None, **kwargs):
+    def __call__(self, alpha=1.0, K=10, size=None, **kwargs):
         return super().__call__(alpha, K, size=size, **kwargs)
 
     def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
-        return default_shape_from_params(self.ndim_supp, dist_params, rep_param_idx, param_shapes)
+        return [dist_params[1] + 1,]
 
     @classmethod
     def rng_fn(cls, rng, alpha, K, size):
@@ -2138,12 +2138,25 @@ class StickBreakingWeightsRV(RandomVariable):
 stickbreakingweights = StickBreakingWeightsRV()
 
 
-class StickBreakingWeights(Continuous):
+class StickBreakingWeights(UnitContinuous):
+    r"""
+    Weights obtained by a (truncated) stick-breaking process
+
+    References
+    ----------
+    ..  Ishwaran, H. and James, L.F., 2001. Gibbs sampling methods for
+        stick-breaking priors. Journal of the American Statistical Association,
+        96(453), pp.161-173.
+    """
     rv_op = stickbreakingweights
+
+    def __new__(cls, name, *args, **kwargs):
+        kwargs.setdefault("transform", transform.stick_breaking)
+        return super().__new__(cls, name, *args, **kwargs)
 
     @classmethod
     def dist(cls, alpha, K, *args, **kwargs):
-        alpha = at.as_tensor_variable(alpha)
+        alpha = at.as_tensor_variable(floatX(alpha))
         K = at.as_tensor_variable(K)
 
         assert_negative_support(alpha, "alpha", "StickBreakingWeights")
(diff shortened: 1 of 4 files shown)


pymc3/distributions/multivariate.py
```diff
@@ -2098,3 +2098,62 @@ def logp(value, mu, W, alpha, tau):
...
+    ndim_supp = 1
+    ndims_params = [0]
+    dtype = "floatX"
+    _print_name = ("StickBreakingWeights", "\\operatorname{StickBreakingWeights}")
+
+    def __call__(self, alpha, size=None, **kwargs):
+        return super().__call__(alpha, size=size, **kwargs)
+
+    def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
+        return default_shape_from_params(self.ndim_supp, dist_params, rep_param_idx, param_shapes)
```
larryshamalama: This was temporarily taken from [`DirichletMultinomialRV`](https://github.com/pymc-devs/pymc3/blob/7e464e59bcb0adb28df94f379b3e8d4af12bd4d1/pymc3/distributions/multivariate.py#L533-L534), but it still isn't correct.

pymc3/distributions/multivariate.py
```diff
@@ -2098,3 +2098,62 @@ def logp(value, mu, W, alpha, tau):
             at.all(alpha >= -1),
             tau > 0,
         )
+
+
+class StickBreakingWeightsRV(RandomVariable):
+    name = "stick_breaking_weights"
+    ndim_supp = 1
+    ndims_params = [0]
```
larryshamalama: Following and revisiting the discussion [here](https://github.com/larryshamalama/pymc3-playground/pull/1) with @Sayam753 , my current guess is that `ndim_supp = 1` and `ndims_params = [0]` because `alpha` should be a scalar and the sampled weights consist of a 1D vector. Still not 100% sure of this.

pymc3/distributions/multivariate.py
```diff
@@ -2098,3 +2098,62 @@ def logp(value, mu, W, alpha, tau):
...
+        )
+
+        return weights
+
+
+stickbreakingweights = StickBreakingWeightsRV()
+
+
+class StickBreakingWeights(Continuous):
+    rv_op = stickbreakingweights
```
larryshamalama, line 2154: This class was included in the PR for completeness, but I have yet to test it given that `stickbreakingweights` is not working

larryshamalama (commented):
Some first comments about my code

pymc3/distributions/multivariate.py
ricardoV94: Seems correct

pymc3/distributions/multivariate.py
ricardoV94: `rep_param_idx=1` is incorrect since your distribution only has 1 argument (zero based indexing). Isn't the shape of your distribution given by `return dist_params`?

pymc3/distributions/multivariate.py
```diff
@@ -2098,3 +2098,62 @@ def logp(value, mu, W, alpha, tau):
...
+
+class StickBreakingWeightsRV(RandomVariable):
+    name = "stick_breaking_weights"
+    ndim_supp = 1
+    ndims_params = [0]
+    dtype = "floatX"
+    _print_name = ("StickBreakingWeights", "\\operatorname{StickBreakingWeights}")
+
+    def __call__(self, alpha, size=None, **kwargs):
+        return super().__call__(alpha, size=size, **kwargs)
```
```suggestion
```

I think you only need to define call if you want to provide default parameters

pymc3/distributions/multivariate.py
```diff
@@ -2098,3 +2098,62 @@ def logp(value, mu, W, alpha, tau):
...
+        return super().dist([alpha], **kwargs)
+
+    def logp(value, alpha):
+        return bound(
+            at.sum(pm.Beta.logp(value, 1, alpha)),
+            alpha > 0,
+        )
+
+    def _distr_parameters_for_repr(self):
+        return ["alpha"]
```
```suggestion
```

This is not used anymore

pymc3/distributions/multivariate.py
```diff
@@ -2098,3 +2098,62 @@ def logp(value, mu, W, alpha, tau):
...
+
+
+stickbreakingweights = StickBreakingWeightsRV()
+
+
+class StickBreakingWeights(Continuous):
+    rv_op = stickbreakingweights
+
+    @classmethod
+    def dist(cls, alpha, *args, **kwargs):
```
```suggestion
    def dist(cls, alpha, **kwargs):
```

pymc3/distributions/multivariate.py
```diff
@@ -2098,3 +2098,62 @@ def logp(value, mu, W, alpha, tau):
...
+        alpha = at.as_tensor_variable(alpha)
```
```suggestion
        alpha = at.as_tensor_variable(floatX(alpha))
```

pymc3/distributions/multivariate.py
ricardoV94: That is, if `alpha=2`, the shape is `2` right?

larryshamalama: Hmm perhaps, but frankly I don't fully understand what `_shape_from_params` is trying to achieve and the difference between `ndim_supp` and the output shape of `RandomVariable`s.

I added a parameter `K` that determines the number of sticks to break and the length of the sampled vector.

Using `return dist_params` works but it seems to require `alpha` to be discrete as yields the following error displayed in the notebook [here](https://github.com/larryshamalama/pymc3-playground/blob/master/notebooks/progress/stickbreakingweights.ipynb):

```{python}
~/anaconda3/envs/pymc3-dev-py38/lib/python3.8/site-packages/aesara/tensor/basic_opt.py in on_import(self, fgraph, node, reason)
   1334                 # keep it this way. See #266 for a better long-term fix.
   1335                 if getattr(d, "dtype", "int64") != "int64":
-> 1336                     assert d.dtype in discrete_dtypes, (node, d.dtype)
   1337                     assert str(d.dtype) != "uint64", node
   1338                     new_shape += sh[len(new_shape) : i + 1]

AssertionError: (stick_breaking_weights_rv{1, (0, 0), floatX, False}(RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7FDE47636120>), TensorConstant{[]}, TensorConstant{11}, TensorConstant{2.0}, TensorConstant{5}), 'float32')
```

ricardoV94: `ndim_supp` provides information about the minimum number of dimensions of the basic distribution. For example a beta distribution is a vector.

But this, by itself, is not sufficient to know the output shape of a specific beta variable (we know it's a vector, but what size?).

`_shape_from_params` does this. Given the parameters and/or their shapes, what is the output shape?

For example, the beta shape is fully determined by the shape of the alpha parameter, so the method could return  `alpha.shape`.

By default, `_shape_from_params` tries to make an educated guess, but it does not always work, and that's when you need to completely override the default method (or, sometimes, just nudge it slightly as in the MultinomialRV case)

larryshamalama: Thanks for the explanation above!

Not sure if this is reflected as default in the code, I added a commit that adds `K` as a parameter.

(Thinking out loud) Using a scalar `alpha`, the output vector would have a shape of `(K + 1,)`. Then, would the shape be `alpha.shape + (K + 1,)` or something like this? I'm not exactly sure how we would write this in Aesara. Would we still have `return dist_params` in `_shape_from_params`?

Here is the mathematical formula of what I am trying to generate (w1, ..., wK, wK+1) where wK+1 = 1 - w1 - ... - wK:

Screen Shot 2021-08-30 at 1 34 21 AM

larryshamalama:
Here is a summary of the current progress. See [notebook here](https://github.com/larryshamalama/pymc3-playground/blob/master/notebooks/progress/stickbreakingweights.ipynb):

`StickBreakingWeightsRV`:
• `rng_fn` is working as expected
• `_shape_from_params` returns `[dist_params[1] + 1,]` because `K` dictates the shape of a single draw, I’m not 100% certain of this (mainly because of the +1)
`StickBreakingWeights`:
• Not sure about the `logp` method
• Should this class inherit from `Continuous` or `UnitContinuous`? `Dirichlet` inherits from `Continuous` and has a default transform `transforms.stick_breaking`, which I currently use for this class, but I’m also not sure
• Sampling not working as we want to, so there’s an error somewhere… The current error that I’m getting is: `ValueError: Mass matrix contains zeros on the diagonal. The derivative of RV 'sbw_stickbreaking__'.ravel()[[ 2  5 11 16 18 19]] is zero.`

larryshamalama:
> Here is a summary of the current progress. See [notebook here](https://github.com/larryshamalama/pymc3-playground/blob/master/notebooks/progress/stickbreakingweights.ipynb):
> 
> `StickBreakingWeightsRV`:
> • `rng_fn` is working as expected
> • `_shape_from_params` returns `[dist_params[1] + 1,]` because `K` dictates the shape of a single draw, I’m not 100% certain of this (mainly because of the +1)
> `StickBreakingWeights`:
> • Not sure about the `logp` method
> • Should this class inherit from `Continuous` or `UnitContinuous`? `Dirichlet` inherits from `Continuous` and has a default transform `transforms.stick_breaking`, which I currently use for this class, but I’m also not sure
> • Sampling not working as we want to, so there’s an error somewhere… The current error that I’m getting is: `ValueError: Mass matrix contains zeros on the diagonal. The derivative of RV 'sbw_stickbreaking__'.ravel()[[ 2 5 11 16 18 19]] is zero.`

Answering my own questions:

- We cannot simply "unstick-break" the weights and plug them back into a Beta `logp` method. We need to use the Generalized Dirichlet distribution (which can further be implemented in the same file). See [wiki page](https://en.wikipedia.org/wiki/Generalized_Dirichlet_distribution) and equation 10 on page 4, section 3.1 in [Ishwaran and James (2001)](http://people.ee.duke.edu/~lcarin/Yuting3.3.06.pdf).
- The distribution inherits from `Continuous` as `UnitContinuous` seems to be mainly for univariate distributions. The default transform is the same for `Dirichlet`, which is `transforms.stick_breaking`.

larryshamalama:
Quick check-in with this PR: the temporary Distribution tests that I have tried need to be revisited. I will have a discussion about the process of this PR (and broadly the DP submodule) and will hopefully have refined tests to be included here.

larryshamalama:
While this is still ongoing work, following the great renaming to PyMC and me accidentally wiping my Documents folder recently, I will close this PR and reopen this in the (hopefully very) near future

PR #4974, opened 2021-08-31
SMC: add n_steps tuning and use non-isometric proposal distribution in MH kernel 

aloctavodia:
This also updates the covariance matrix between stages, this step is very important for the kernel to work properly,

### DIFF
--- a/pymc3/smc/smc.py
+++ b/pymc3/smc/smc.py
@@ -352,7 +352,7 @@ class IMH(SMC_KERNEL):
         self.p_acc_rate = p_acc_rate
 
         self.max_steps = n_steps
-        self.proposed = self.proposed = self.draws * self.n_steps
+        self.proposed = self.draws * self.n_steps
         self.proposal_dist = None
         self.acc_rate = None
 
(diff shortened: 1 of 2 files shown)


twiecki (approved):

twiecki:
@aloctavodia Want to add a small test?

pymc3/smc/smc.py
```diff
@@ -462,23 +472,31 @@ def resample(self):
...
         ac_ = np.empty((self.n_steps, self.draws))
 
         log_R = np.log(np.random.rand(self.n_steps, self.draws))
         for n_step in range(self.n_steps):
-            proposal = floatX(
-                self.tempered_posterior
-                + self.proposal_dist.rvs(size=self.draws) * self.proposal_scales[:, None]
-            )
+            proposal = self.proposal_dist.rvs(size=self.draws)
+            proposal = proposal.reshape(len(proposal), -1)
```
ricardoV94: Why is the reshape necessary?

aloctavodia: The test passed for IMH kernel and not for MH and when I check the dimensions were OK for IMH  but not MH. Also I find another issue with the MH kernel, but I will add it in a separated PR as is not related to the tuning of the steps and explains why in my simulations I was getting such poor results with the MH kernel.

aloctavodia: I removed the reshaping I am now using the MultivariateProposal from pm.step_methods.metropolis as we used to do and because of that change I decided to also include in this PR the update of the covariance matrix and the multivariate proposal at each stage, that we used to do but we missed when we added the MH kernel back to SMC.

ricardoV94 (approved):

Issue #4959, opened 2021-08-22
`sample_prior_predictive` does not work with Aesara v.2.1.3


larryshamalama:
Reopening this [issue](https://github.com/aesara-devs/aesara/issues/568); the text below is copy pasted from the latter link. As @ricardoV94 suggests, the problem may be caused by this [PR](https://github.com/aesara-devs/aesara/pull/556).

`pm.sample_prior_predictive` doesn't seem to work when I upgrade Aesara to the latest version (2.1.3), but it works if I use Aesara v.2.0.12. Any ideas why this may be the case? See the example below and notebooks [here](https://github.com/larryshamalama/pymc3-playground/blob/master/notebooks/shortcomings/sample_prior_aesara_working.ipynb) and [here](https://github.com/larryshamalama/pymc3-playground/blob/master/notebooks/shortcomings/sample_prior_aesara_not_working.ipynb)

```{python}
with pm.Model() as model:
    norm = pm.Normal("normal-dist", mu=0., sigma=5.)
    
    prior = pm.sample_prior_predictive()
```

yields

```{python}
TypeError: ('Constants not allowed in param list', TensorConstant{[]})
```

larryshamalama:
In [pymc3/sampling.py](https://github.com/pymc-devs/pymc3/blob/main/pymc3/sampling.py#L2004), printing `inputs` in `inputs = [i for i in inputvars(vars_to_sample) if not isinstance(i, SharedVariable)]` yields:

`[TensorConstant{[]}, TensorConstant{11}, TensorConstant{[]}, TensorConstant{11}, TensorConstant{0.0}, TensorConstant{1.0}, TensorConstant{5.0}]`

in to the example above. It seems that it is passed into `compile_rv_inplace(inputs, ...)` which does not accept Constants as per [here](https://github.com/aesara-devs/aesara/blob/main/aesara/compile/function/pfunc.py#L538).

PR #4975, opened 2021-09-01
Filter out constants from `pm.sample_prior_predictive` inputs

larryshamalama:
This PR addressed issue #4959, which may be due to this [PR](https://github.com/aesara-devs/aesara/pull/556).

In [sampling.py](https://github.com/pymc-devs/pymc3/blob/main/pymc3/sampling.py#L2004), `inputs` can be a list of `TensorConstant`s, which is not permitted in [`compile_rv_function`](https://github.com/pymc-devs/pymc3/blob/main/pymc3/aesaraf.py#L879), more specifically in [`pfunc`](https://github.com/aesara-devs/aesara/blob/main/aesara/compile/function/pfunc.py#L538).

A naive fix to the example mentioned in the issue is to exclude `TensorConstant`s on top of `SharedVariable`s in `inputs`. Would this have other implications that I am not aware of?

### DIFF
--- a/pymc3/sampling.py
+++ b/pymc3/sampling.py
@@ -31,6 +31,7 @@ import numpy as np
 import xarray
 
 from aesara.compile.mode import Mode
+from aesara.graph.basic import Constant
 from aesara.tensor.sharedvar import SharedVariable
 from arviz import InferenceData
 from fastprogress.fastprogress import progress_bar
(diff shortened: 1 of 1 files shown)


larryshamalama:
The failing tests suggest that my approach is incorrect...

ricardoV94 (approved):
Nothing naive about the fix, but the commit message should be more "to the point"

larryshamalama:
Gotcha, thanks for the suggestion

twiecki:
Thanks @larryshamalama!

PR #4977, opened 2021-09-01
Extract initval evaluation into its own method

michaelosthege:
This is a refactoring in preparation of changing how and when initial values are numerified.

This PR does not change any behavior. It just extracts a code block into a private method and adds unit tests for it that are more targeted than our current tests for it.

We may want to add more test cases to cover chained initial values too. (Suggestions welcome!)


twiecki (approved):

twiecki:
Thanks @michaelosthege!

PR #4978, opened 2021-09-02
Removed temporary pin of Aesara version

kc611:
Fixes #4961

PR #4979, opened 2021-09-03
Updated incorrect docstring and removed redundant logic in sampling.py

kc611:
Minor fixes based upon: https://github.com/pymc-devs/pymc3/pull/4912#discussion_r701208830 and https://github.com/pymc-devs/pymc3/pull/4978#issuecomment-911814142

### DIFF
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -339,17 +339,15 @@ class Distribution(metaclass=DistributionMeta):
 
 @singledispatch
 def _get_moment(op, rv, size, *rv_inputs) -> TensorVariable:
-    """Fallback method for creating an initial value for a random variable.
-
-    Parameters are the same as for the `.dist()` method.
-    """
     return None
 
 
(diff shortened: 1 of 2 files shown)


pymc3/distributions/distribution.py
```diff
@@ -339,17 +339,15 @@ def dist(
...
-
-    Parameters are the same as for the `.dist()` method.
-    """
     return None
 
 
 def get_moment(rv: TensorVariable) -> TensorVariable:
-    """Fallback method for creating an initial value for a random variable.
+    """Method for specification of an stable starting point/value
+    for sampling of a distribution.
```
michaelosthege:
```suggestion
    """Method for choosing a representative point/value
    that can be used to start optimization or MCMC sampling.
```

kc611: Applied in latest commit :+1:

ricardoV94 (approved):
LGTM

PR #4982, opened 2021-09-04
Prepare initval API for method-based lazy evaluation

michaelosthege:
This PR refactors the `Model.initial_point` property into a `get_initial_point` method that can takes a `recompute` parameter.
This kwarg is going to be relevant for the implementation of lazy initval evaluation, since it allows us to re-draw initial values from the prior.

### DIFF
--- a/pymc3/model.py
+++ b/pymc3/model.py
@@ -913,6 +913,16 @@ class Model(Factor, WithMemoization, metaclass=ContextMeta):
         return inputvars(self.unobserved_RVs)
 
     @property
+    def disc_vars(self):
+        """All the discrete variables in the model"""
+        return list(typefilter(self.value_vars, discrete_types))
+
+    @property
+    def cont_vars(self):
+        """All the continuous variables in the model"""
+        return list(typefilter(self.value_vars, continuous_types))
+
+    @property
     def test_point(self) -> Dict[str, np.ndarray]:
         """Deprecated alias for `Model.initial_point`."""
         warnings.warn(
@@ -934,16 +944,6 @@ class Model(Factor, WithMemoization, metaclass=ContextMeta):
         For a name-based dictionary use the `initial_point` property.
         """
         return self._initial_values
-
-    @property
-    def disc_vars(self):
-        """All the discrete variables in the model"""
-        return list(typefilter(self.value_vars, discrete_types))
-
-    @property
-    def cont_vars(self):
-        """All the continuous variables in the model"""
-        return list(typefilter(self.value_vars, continuous_types))
 
     def set_initval(self, rv_var, initval):
         if initval is not None:
--- a/pymc3/model.py
+++ b/pymc3/model.py
@@ -650,6 +650,7 @@ class Model(Factor, WithMemoization, metaclass=ContextMeta):
         # The sequence of model-generated RNGs
         self.rng_seq = []
         self._initial_values = {}
+        self._initial_point_cache = {}
 
         if self.parent is not None:
             self.named_vars = treedict(parent=self.parent.named_vars)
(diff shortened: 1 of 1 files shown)


ricardoV94:
Why not stick with the less verbose `initial_point`? That is new from V3 anyway, so I think it's fine to change the behavior still.

michaelosthege:
You mean for the name of the method?
I included the imperative to make it clear that there is an action involved, and that it is not a property.

My understanding is that it's a good practice to include a verb in method or function names, no?

michaelosthege:
Also I didn't want to break any code that is using the property.

ricardoV94:
Doesn't seem like it would break anything with the default `recompute=False`, no?

michaelosthege:
> Doesn't seem like it would break anything with the default `recompute=False`, no?

Exactly.
I'm branching off this branch to continue with the lazy evaluation. Just wanted to keep these commits - particularly the first of the two - out of the upcoming PR.

I'll be afk soon, but will continue later today. If I run into problems with the toposorted lazy evaluation I'll ping you on Slack

pymc3/model.py
```diff
@@ -912,18 +912,49 @@ def independent_vars(self):
...
+            "`Model.test_point` has been deprecated. Use `Model.get_initial_point()` instead.",
             DeprecationWarning,
         )
-        return self.initial_point
+        return self.get_initial_point(recompute=False)
 
     @property
     def initial_point(self) -> Dict[str, np.ndarray]:
-        """Maps names of variables to initial values."""
+        """Deprecated alias for `Model.get_initial_point()`."""
```
twiecki: I agree with @ricardoV94, I don't think we need to deprecate or could just keep with this property, no?

michaelosthege: Methods and functions should have a verb in their name.
My understanding is that this is quite a universal rule of thumb across all languages.

We could keep the property around without the warning if we want that alias to exist though.

ricardoV94: Actually found this discussion recently on twitter: https://twitter.com/raymondh/status/1428457757674508289

Personally, I don't think the `get` adds anything. What else could `initial_point` mean? Certainly it must return something since it has no verb, and the thing it returns is right in the name. On the other hand, many people in that thread disagree with not using `get`

michaelosthege: > What else could initial_point mean?

Exactly! When I read `initial_point`, I expect that to be the initial point. Not a callable!

Particulary because the old API `test_point` which _is_ the test point. And there are also properties like `initial_values` which _are_ the dict already.
All other method names in the `Model` (except `point_logps`) have either a verb, or a pattern like `name_of(...)`.

[Here's a good article](https://dmitripavlutin.com/coding-like-shakespeare-practical-function-naming-conventions/).

ricardoV94: Then what about calling the new method `(re)compute_initial_point` without any boolean flag, and keep the `initial_point` attribute without deprecation warning? I guess this is the reason why you didn't want to remove the `initial_point` altogether?

michaelosthege: So the method `recompute_initial_point` always recomputes and the `initial_point` property gives the cached version and recomputes only when necessary?

I see two re-compute stategies though. So we might still need a `all: bool` flag that `inital_point` sets to `False` if it has to trigger a recompute.
* Re-compute all.
* Re-compute only those that are missing from the cache.

ricardoV94: > So the method `recompute_initial_point` always recomputes and the `initial_point` property gives the cached version and recomputes only when necessary?

Yes

> I see two re-compute stategies though. So we might still need a `all: bool` flag that `inital_point` sets to `False` if it has to trigger a recompute.
> 
>     * Re-compute all.
> 
>     * Re-compute only those that are missing from the cache.

What is the point of the second? That is something you have not implemented in this PR yet, or have you?

michaelosthege: > What is the point of the second? That is something you have not implemented in this PR yet, or have you?

Correct, that's not implemented right now, but it would be rather easy by skipping all vars that are already in the cache.

Doing this would add more consistency for situations where variables are added _after_ `find_MAP`/`sample` were already called. This is commonly seen with gaussian processes, where the conditional is added after the sampling.
It would also save a bit of compute - scaling with the model size..

michaelosthege: I renamed it and changed the caching as discussed above.
Now the `Point` function is called less often..

ricardoV94 (approved):

twiecki (approved):

Issue #4846, opened 2021-07-08
Consistent API for choosing model RVs / value vars


ricardoV94:
We have these three types of variables: RandomVariables --- Value vars  --- Transformed value vars

```python
with pm.Model() as m:
    x = pm.Normal('x')
    y = pm.HalfNormal('y')
```
`model["x"]` returns the `x` RV
`model["y"]` returns the `y` RV
`model["y_log__"]` returns the `y` transformed value var.

`model.rvs_to_values[model["x"]]` returns the `x` value var
`model.rvs_to_values[model["y"]]` returns the `y` transformed value var

The automatic transforms were always a source of confusion, and I am afraid the new changes will make things a bit more confusing still. We should try to be consistent in what is expected as input in PyMC sampling methods. e.g., when:
1. assigning variables to step methods (see #4769)
1. passing a starting point  dictionary (see #4689)
1. choosing which variables to include prior/posterior predictive sampling (see #4701)
1. Probably more...

## Proposal

I suggest that calling `model["varname"]` should always return the corresponding RV. All PyMC3 functions that need a list of variables as inputs should expect an RV and obtain the respective (transformed) value vars if needed. In line with this:
* Calling `model["y_log__"]` should fail.
* When defining a starting dict, the untransformed value of `y` should be expected and automatically transformed to `y_log__` if needed
* Step methods should expect a list of RVs as input and retrieve the respective value vars, so that `step = pm.Metropolis([model["x"], model["y"]])` would work.
* Any methods that accept variable names instead of the actual RVs / value vars should expect these to refer to the name of the RVs. 

Defining everything in terms of value vars makes less sense to me. For example they are not used in prior/predictive sampling. Users are also less likely to understand what they mean/ do.

The current approach makes even less sense, and results in a lot of duck typing / guessing on the backend to figure out what the user is trying to select.

brandonwillard:
> We have these three types of variables: RandomVariables --- Value vars --- Transformed value vars

Transformed variables are not a type of variable in the same sense that random variables and their value variables are; instead, they're value variables without explicit random variables.  This is due to the old design of PyMC3, which exposes transformed variables at the user-level and makes them an explicit part of the API.

> I suggest that calling `model["varname"]` should always return the corresponding RV.

In general, it would probably be better to have the user only ever deal with random variables, and transforms should only ever be used internally.

> Defining everything in terms of value vars makes less sense to me.

What do you mean by "Defining everything in terms of value vars"?

[...]

PR #4983, opened 2021-09-04
Implement and switch to lazy initval evaluation framework

### LINKED ISSUE #4484: Invalid logic in sample prior predictive for transformed variables
Prior predictive sampling draws values from the untransformed variables and then attempts to recreate the transformed values via `transformation.forward_val(untransformed_values)`

However, this logic is not sufficient to recover the correct values when there are stochastic bounds (e.g., `pm.Uniform('x', lower=0, upper=y)`, because we also need to know what the values of the stochastic bounds were in each sample (i.e., the `y`).

People don't usually care about transformed variables, but this is critical for example in `sample_smc()` as it creates it's particles from an initial prior_predictive call.

Hopefully this will be gone in V4.0, but I thought best to document it just in case (and in the meanwhile).

Minimal reproducible example:
```python
np.random.seed(1)
with pm.Model() as m:
    y = pm.Uniform('y', 0, 1)
    x = pm.Uniform('x', 0, y)

print(np.mean(np.isnan(prior['x_interval__'])))
```
```
/home/ricardo/Documents/Projects/pymc3/pymc3/distributions/transforms.py:294: RuntimeWarning: invalid value encountered in log
  return floatX(np.log(x - a) - np.log(b - x))

0.314
```
The 31.4% `NANs` correspond to cases where `a` and `b` were either above or below x due to random sampling in the block below. The `NANs` are the most obvious symptom, but all values are in fact incorrect:

https://github.com/pymc-devs/pymc3/blob/32b5c941c8f43b485afd3da1ae5544c3c50b9e77/pymc3/distributions/transforms.py#L289-L294

Which is called at the end of `prior_predictive_sampling`, without providing a `point` or any other contextual info:

https://github.com/pymc-devs/pymc3/blob/37ca5ea9b25aa00ed4d460f0fb417712d125d248/pymc3/sampling.py#L1941-L1957

### LINKED ISSUE #4924: Initval refactoring
# Proposal

**Create a `model.compile_initial_point_fn` which creates a compiled aesara function that computes a (transformed) initial point for each RV in the model simultaneously.**

This method takes as input a dictionary with optional user choices concerning initvals, whose values might be either numerical values, symbolic expressions (which can only depend on upstream RVs / Deterministics / or shared variables), or a string of `["random", "moment"]`, which means a random draw should be taken from the RV or a fixed moment should be extracted from the RV (using something like #4912).

These choices would be saved in a model dictionary such as `model._user_initval_choices` as new variables are defined.

Finally `model.initial_point` would look something like this:
```python
def initial_point(self, recompute=False):
    if recompute or self.last_computed_initval is None:
        initial_point = self.compile_initial_point_fn(self._user_initval_choices)():
        ... # convert to dict format
        self.last_computed_initval = initial_point_dict
    return self.last_computed_initval
```
The main goal is to decouple the model definition from the sampling phases, addressing issues like #4918 

A `recompute` flag is used to ensure a new initial_point is not recomputed uselessly, as many places in the codebase call this property frequently. Alternatively we can keep the old property and add a `recompute_inital_point()` that changes the `self.last_computed_initval` 

## Other benefits

This would allow us to simplify the slightly redundant and potentially defective `model.update_start_vals` (mentioned in https://github.com/pymc-devs/pymc3/issues/4484#issuecomment-788715869): https://github.com/pymc-devs/pymc3/blob/6a75744b31f3ec015856ac6ea374fe12be8cc156/pymc3/model.py#L1548-L1556

to something like:
```python
def update_start_vals(self, new_initval_choices):
    initval_choices = deepcopy.copy(self._user_initval_choices)
    initval_choices.update(new_initval_choices)
    start_vals = model.compile_initial_point_fn(initval_choices)()
    ... # convert to dict
    return start_vals_dict
```

*** 
We could also revert some ugly changes to `prior_predictive_sampling` introduced by yours truly in https://github.com/pymc-devs/pymc3/pull/4769/commits/687f044bfcf37f58a410ea21509174d7a3ea5fc0:

https://github.com/pymc-devs/pymc3/blob/6d2aa5ddebed01d81c2ab66b9d4bd02194f82508/pymc3/sampling.py#L1983-L1999

This hack resulted from a different difficulty in obtaining multiple transformed `initial_points` (to kick-start the SMC sampler). All that would be needed after the proposed changes would be something like the following:

```python
random_initvals = {var: "random" for var in model._user_initval_choices.keys()}
initial_point_fn = model.compile_initial_point_fn(random_initvals)
initvals = zip(*(initial_point_fn() for i in range(samples)))
...  # convert to dict format
```

And the changes in https://github.com/pymc-devs/pymc3/pull/4769/commits/687f044bfcf37f58a410ea21509174d7a3ea5fc0 could be removed altogether.

michaelosthege:
## Changes
+ **The initial values are now evaluated lazily, all at once.** `Model.initial_values` and `Distribution(initval=...)` can now take `None`, `ndarray`, `Variable`, `"moment"` or `"prior"`.
+ `Model.initial_values` are now managed by RV instead of value var tensor
+ `init_nuts` signature changes
+ `pm.sample()` start kwarg change of name and meaning
+ `Model.update_start_vals` was removed (with informative error message)
+ All new initial point logic was moved to it's own module at `initial_point.py`

## ToDo
+ [x]  Add additional tests for the new re-size flexibility that the lazy evaluation allows for. (`test_initvals.py::TestInitvalEvaluation::test_initval_resizing`)
+ [x] <s>Mention changes in the RELEASE-NOTES.md</s> Doing this in the HackMD document

Closes #4924 
Closes #4484


michaelosthege:
Could the remaining test failure `pymc3/tests/test_ndarray_backend.py::TestSaveLoad::test_sample_posterior_predictive` be related to seeds/rng of the initial values?

pymc3/model.py
```diff
@@ -912,51 +913,90 @@ def independent_vars(self):
...
+                        rv_var, initval, test_value, transform, given=numeric_initvals
+                    )
 
-    @property
-    def disc_vars(self):
-        """All the discrete variables in the model"""
-        return list(typefilter(self.value_vars, discrete_types))
+            # Cache the evaluation results for next time.
+            self._cached_numeric_initvals = numeric_initvals
+        return Point(list(numeric_initvals.items()), model=self)
```
ricardoV94: Were we creating a `Point` object every time we accessed `initial_point`before?

michaelosthege: Yeah.

I considered to cache the point, but the set comparison further up is a little easier with the tensors as keys instead of the names.

michaelosthege:
I added two more tests - dependent initvals and resizing works.

But something is odd with the random state. This is from the failing `test_ndarray_backend.py::TestSaveLoad::test_sample_posterior_predictive`:
image


That backend is deprecated and the test doesn't seem to be very targeted. It looks like the `MultiTrace` constructor accesses `Model.initial_point` and this random states used for the initval evaluation actually do interfer with the models `rng_seeder`.
@ricardoV94 can you take a look at the evaluation method?

Also what exactly do we want in terms of rng for initial value evaluation? Should it be independent, optionally with their own `rng`, or should it re-use the models rng?

pymc3/model.py
michaelosthege: Caching was improved in #4982

aseyboldt:
<s>The remaining test failures seem to be a result of #5007, but hopefully that would be fixed by #4887, so maybe we wait for that to be merged?</s>

__UPDATE:__ I marked them as XFAIL to unblock this PR.

pymc3/distributions/discrete.py
```diff
@@ -394,6 +394,10 @@ def logcdf(value, p):
             p <= 1,
         )
 
+    def get_moment(value, size, p):
+        p = at.full(size, p)
+        return at.switch(p < 0.5, np.int64(0), 1)
```
ricardoV94:
```suggestion
        return at.switch(p < 0.5, np.int64(0), np.int64(1))
```

pymc3/distributions/distribution.py
```diff
@@ -339,7 +341,7 @@ def dist(
 
 @singledispatch
 def _get_moment(op, rv, size, *rv_inputs) -> TensorVariable:
-    return None
+    raise NotImplementedError(f"For random variable {rv} of type {op} has no get_moment.")
```
ricardoV94:
```suggestion
    raise NotImplementedError(f"random variable {rv} of type {op} has no get_moment.")
```

pymc3/model.py
```diff
@@ -935,50 +943,214 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
+        if jitter_rvs is None:
+            jitter_rvs = set()
+
+        initial_values = []
+        initial_values_transformed = []
+
+        for variable in self.free_RVs:
+            strategy = self.initial_values.get(variable, None)
+            if strategy is None:
+                strategy = default_strategy
```
ricardoV94:
```suggestion
            strategy = self.initial_values.get(variable, default_strategy)
```

pymc3/model.py
```diff
@@ -935,50 +943,214 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
+
+            if transform is not None:
+                value = transform.backward(variable, value)
+
+            initial_values.append(value)
+
+        # Copy the outputs in a way that we still know what corresponds to the free_RVS
+        all_inputs = []
+        for var in self.free_RVs:
+            all_inputs.extend(aesara.graph.graph_inputs([var]))
```
ricardoV94: Is this being used anywhere?

pymc3/model.py
```diff
@@ -935,50 +943,214 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
 
+        Parameters
+        ----------
+        jitter_rvs : set
+            The set (or list or tuple) of random variables for which a U(-1, +1) jitter should be
+            added to the initial value. Only available for variables that have a transform or real-valued support.
+        default_strategy : str
+            Which of { "moment", "prior" } to prefer if the initval setting for an RV is None.
+        return_transformed : bool
+            Switches between returning the dictionary based on RV vars or RV value vars as keys.
```
ricardoV94: This docstring is not correct, I think

pymc3/model.py
```diff
@@ -935,50 +943,214 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
+        graph = aesara.graph.FunctionGraph(outputs=initial_values, clone=False)
+
+        rng_nodes = find_rng_nodes(graph.outputs)
+        new_rng = [aesara.shared(np.random.RandomState(np.random.PCG64())) for _ in rng_nodes]
+        graph.replace_all(zip(rng_nodes, new_rng), import_missing=True)
+        func = compile_rv_inplace(inputs=[], outputs=graph.outputs)
+
+        varnames = []
+        for var in self.free_RVs:
+            transform = getattr(var.tag.value_var.tag, "transform", None)
```
ricardoV94:
```suggestion
            transform = getattr(self.rvs_to_values[var].tag, "transform", None)
```

ricardoV94:
Good time to update the brittle `model.update_start_vals`

https://github.com/pymc-devs/pymc3/blob/4f8ad5d2772c8fd98b403d7c1d141e8f8394ba84/pymc3/model.py#L1594

There's more context here (with a different pseudo-API): https://github.com/pymc-devs/pymc3/issues/4924#issue-970271432

pymc3/model.py
michaelosthege: So far we put `None` in the `self.initial_values` too (to indicate that no choice was made).

pymc3/model.py
michaelosthege:
```suggestion
        return_transformed : bool
            If `true` the returned variables will correspond to transformed initial values.
```

pymc3/model.py
ricardoV94: What do we do with that info?

pymc3/distributions/distribution.py
michaelosthege: applied in https://github.com/pymc-devs/pymc3/pull/5025

pymc3/distributions/discrete.py
michaelosthege: https://github.com/pymc-devs/pymc3/pull/5025

pymc/model.py
```diff
@@ -920,50 +928,228 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
+        all_inputs = []
+        for var in self.free_RVs:
+            all_inputs.extend(aesara.graph.graph_inputs([var]))
+
+        all_outputs = []
+        all_outputs.extend(self.free_RVs)
+        all_outputs.extend(initial_values)
+        all_outputs.extend(initial_values_transformed)
+
+        copy_graph = aesara.graph.FunctionGraph(outputs=all_outputs, clone=True)
```
ricardoV94: We should see if we can use the `memo` functionality as done in aeppl here: https://github.com/aesara-devs/aeppl/blob/d24eee80a7448c48b55a8ec41aec150d1dd9d6a7/aeppl/joint_logprob.py#L82-L103

pymc/model.py
```diff
@@ -920,50 +928,228 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
+
+        def find_rng_nodes(variables):
+            def expand(node):
+                if not hasattr(node, "owner") or node.owner is None:
+                    return []
+                return node.owner.inputs
+
+            rng_nodes = []
+
+            for node in aesara.graph.basic.walk(variables, expand):
```
ricardoV94: We can use `graph_inputs` here instead of `walk`

pymc3/model.py
michaelosthege: We fall back to the `default_strategy`.
It's a way of enforcing the default strategy and generally more robust, even if we decide to not store `None`-valued initvals, which I would advise against in comparison to a dict that has entries for all model variables.

pymc3/distributions/discrete.py
ricardoV94: Why not use the version that will be used in that PR already?

pymc3/model.py
ricardoV94: My suggestion does exactly the same, but in a single line, that's all

michaelosthege: No it doesn't: If there's a `None`-valued entry in `initial_values`, the `get` method will return the `None`, not the `default_strategy`

ricardoV94: Maybe I am looking at an outdated version, but when I commented there was an immediate follow-up, which ends up in exactly the same result:
```python
strategy = self.initial_values.get(variable, None)
if strategy is None:
   strategy = default_strategy
```

pymc3/distributions/discrete.py
michaelosthege: I rebased onto `main` which now includes those changes already.

pymc/model.py
```diff
@@ -920,50 +928,216 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
+        initial_values_transformed = []
+
+        for variable in self.free_RVs:
+            if variable in overrides:
+                strategy = overrides[variable]
+            else:
+                strategy = self.initial_values.get(variable, None)
+
+            if strategy is None:
+                strategy = default_strategy
```
michaelosthege: @ricardoV94 let's continue here. This is the current version.

pymc3/model.py
michaelosthege: see new thread

pymc/model.py
michaelosthege: Not sure how we'd do that here; also not sure if it's worth the added complexity.
If you think it's worthwhile, can we defer that to a follow up?

michaelosthege:
I believe that one test failure is just flaky in the float32 setting.
Please give your review and I'm sure there'll be another (force) push anyway.

ricardoV94 (commented):
Looks great so far.

A couple of ideas/suggestions/concerns:

1. Should we move the initval logic to a separate file? 
2. What about breaking the `start` dict functionality in `pm.sample`, to exclusively expect RVs as keys and initial_values / strategies on the constrained space? Thinking about moving towards a consistent API where users don't have to juggle between RVs, Value vars, Transformed Value vars, Strings, etc (as discussed in #4846)
3. Are we sure the sampler methods are not doing expensive `model.initial_point` calls, which now would result in recompiling and drawing from a graph every time the attribute is accessed? I didn't look carefully, so this might very well be an unfounded concern.

michaelosthege:
> Should we move the initval logic to a separate file?

Just did that 👍

> What about breaking the start dict functionality in pm.sample, to exclusively expect RVs as keys and initial_values / strategies on the constrained space? Thinking about moving towards a consistent API where users don't have to juggle between RVs, Value vars, Transformed Value vars, Strings, etc (as discussed in #4846)

For `4.0.0` we should definitely keep the `str`-compatibility, but I'm open to to reconsider that API for `>= 4.1.0`.
Personally I'm not unconditionally a fan of requiring `TensorVariable` keys, because it may end up in quite a lot of boilerplate user code:
```
pm.sample(start={
    pmodel["intercept"] : 12,
    pmodel["slope"]: [3, 5],
})

# vs

pm.sample(start={
    "intercept" : 12,
    "slope": [3, 5],
})
```
Maybe let's create a [Discussion](https://github.com/pymc-devs/pymc/discussions) and to consider also other (upcoming) properties like the transforms dictionary?

> Are we sure the sampler methods are not doing expensive model.initial_point calls, which now would result in recompiling and drawing from a graph every time the attribute is accessed?

There are calls to `model.initial_point` mostly in backend and step method initialization. However I think a lot of that can be refactored by creating a dedicated `Model.get_rv_shapes(transformed: bool) -> Dict[TensorVariable, Tuple[int, ...]]` method (see #5031).
I see that as an important refactor, but out-of-scope of this PR.

pymc/model.py
ricardoV94: Yes, this is quite a hackish thing we did just to be able to get a reference between the old and new variables. Just because we didn't know about the memo thing

ricardoV94:
> > Should we move the initval logic to a separate file?
> 
> Just did that +1
> 

Should we move even more? We can pass the model to most of these functions instead of relying on self internally.

michaelosthege:
> Should we move even more? We can pass the model to most of these functions instead of relying on self internally.

I don't like that approach, because it makes type annotations hard (cyclic imports) and does not simplify anything. The dependency on the model object is still there and it has no advantages for testing either.
Furthermore the access to such functions is then again less discoverable.

ricardoV94:
Sure, is just that model.py is rather large as is. 

In that case I would revert the last commit. Just for those two (+ the one todo) functions, it doesn't seem worth it to separate things that are supposed to work together across two files.

My preference would still be to put everything that's not a property in the separate file.

michaelosthege:
Actually I prefer the split-up since these helper functions are needed by code in `approximations.py` and `starting.py`.

Generally, since we're nitpicking about moving code around, does that mean we're happy with the behavioral and algorithmic aspects of this PR? (3rd opinion by @lucianopaz or @aseyboldt or @twiecki please)

ricardoV94:
> Personally I'm not unconditionally a fan of requiring TensorVariable keys, because it may end up in quite a lot of boilerplate user code:

Well, in most cases you have the handle of the TensorVariables already. Also you can always call your model something shorter.

pymc/tests/test_initvals.py
```diff
@@ -38,59 +43,87 @@ def test_new_warnings(self):
...
+        with pm.Model() as pmodel:
+            A = pm.Flat("A", initval="moment")
+            B = pm.HalfFlat("B", initval="moment")
+            C = pm.Normal("C", mu=A + B, initval="moment")
+            fn = pmodel.make_initial_point_fn(jitter_rvs={B}, return_transformed=True)
+            iv = fn(0)
+        # Moment of the Flat is 0
+        assert iv["A"] == 0
+        # Moment of the HalfFlat is 1, but HalfFlat is log-transformed by default
+        # so the transformed initial value with jitter will be
```
twiecki: will be what?

twiecki:
OK, I think this moves us closer to our goal and I like that we put the jitter logic into the initial choice rather than into the sampler. I'm not a huge fan of allowing to pick a point from the prior as I don't think it's necessary but probably not a big burden to support it here. So I think we should move ahead as it unblocks several other things.

pymc/tests/test_initvals.py
michaelosthege:
```suggestion
        # so the transformed initial value with jitter will be zero plus a jitter between [-1, 1].
```

twiecki:
I agree with @ricardoV94 that moving more into the initial value submodule is a good idea and that `model.py` is overburdened. I also agree that it's enough to allow indexing by the TensorVariable directly, a user has that handy from the model spec.

michaelosthege:
> I agree with @ricardoV94 that moving more into the initial value submodule is a good idea and that `model.py` is overburdened.

I agree that it's overburdened, but note that `Model` is by far not the only thing in there. Changing OOP classmethods into functional API because the file is "too big" however does not sound appealing.

Also note that @ricardoV94 reversed his opinion about the moving. (After convincing me that it's a good idea to put helper functions there.)

> I also agree that it's enough to allow indexing by the TensorVariable directly, a user has that handy from the model spec.

So you want to make another unanounced breaking change of the `start` kwarg API by requiring TensorVariables as keys?

Instead how about this:
1. We keep it as it for the `4.0.0` release
2. You do another Twitter poll, asking users which of the three options they prefer:
  * **A:** `str` (untransformed or transformed) variable names
  * **B:** RV tensors or transformed RV tensors (which we don't have yet)
  * **C:** support for both A and B

twiecki:
> Also note that @ricardoV94 reversed his opinion about the moving. (After convincing me that it's a good idea to put helper functions there.)

Alright, then.

> So you want to make another unanounced breaking change of the start kwarg API by requiring TensorVariables as keys?

Well if not with this release, then when? I don't think it's a big issue because very few people are using this. That's also why I don't think a twitter poll will work as it would have to describe a feature that maybe <5% of users even know or care about. So making our lives easier here is a win.

michaelosthege:
> Alright, then.

...so? How do we proceed?

> Well if not with this release, then when? I don't think it's a big issue because very few people are using this. That's also why I don't think a twitter poll will work as it would have to describe a feature that maybe <5% of users even know or care about. So making our lives easier here is a win.

I though we all agreed that we want to have as few unexpected breaking changes as possible, so the transition is smooth?

Also we're taking about the `pm.sample(start=...)` and `pm.find_MAP(start=...)` kwarg. In my experience this is a _very_ frequently used API !
On discourse we have a ton of threads containing `start=`.

Also since we've removed the option to pass `len(trace) > 0` traces to `pm.sample()`, the go-to way to continue sampling from an existing trace is `pm.sample(start=trace[-1])`, or the equivalent expression from an `InferenceData`. That's a `str`-keyed dictionary.

ricardoV94:
I still think we should move more initvals things to its own module. 

My second preference would be to not move anything at all if we are only moving the two(?) methods as in the current PR. 

***

One sidenote about transforms that make me not like the string thing so much. In V4 we are actually free to change/define the transforms of a variable at any point. There is nothing about the RVs or value variables that is "transformed" until a logp is called. This means that using things like `"x_log__"` as keys might not make much sense. 

Anyway I think RV based in general is much more intuitive. Users already use it to define the models. Nobody seems to want to write `pm.Normal("y”, mu="x")`. Why would they want it/ expect it here (other than accidental history)? 

This discussion however does not need to be had here. Let's open a discussion topic.

ricardoV94:
In general I rather build a consistent API over ensuring backwards compatibility. We are not Microsoft xD. 

Also, didn't this PR change the `start` keyword argument as well?

michaelosthege:
Fair point about changing transforms after the fact, but I'm not going to support breaking the `start` API.
If discussion and Twitter poll say so we can require strings for the `initval` kwarg. Nevertheless I see abosolutely no benefit in not supporting untransformed RV names, because it's literally 1 line of library code that's greatly improving UX  in countless scenarios.

This PR is now a month old, which is about 4-5x more than what I consider a healthy target turnaround time.
I need to get this off my radar ASAP.
Do we want to merge with or without that last commit @ricardoV94 @twiecki ?

twiecki:
Definitely with the last commit. Well, I think you should respect the majority rule here but if you insist I'm OK moving ahead.

michaelosthege:
I insist on not spending additional dev time to make an unannounced breaking change that's not even technically necessary (the implementation is there).

If you insist on removing `str`-key support (in favor of `TensorVariable`) from the `initval` kwarg please tell me how to deal with transformed initial values for which we do not yet have a `TensorVariable` representation?

ricardoV94:
> If you insist on removing `str`-key support (in favor of `TensorVariable`) from the `initval` kwarg please tell me how to deal with transformed initial values for which we do not yet have a `TensorVariable` representation?

What do you mean? Defining the initval directly on the unconstrained space? That would not be possible (the same way you can't do it via the distribution initval kwarg). Or did you mean something else?

michaelosthege:
> > If you insist on removing `str`-key support (in favor of `TensorVariable`) from the `initval` kwarg please tell me how to deal with transformed initial values for which we do not yet have a `TensorVariable` representation?
> 
> What do you mean? Defining the initval directly on the unconstrained space? That would not be possible (the same way you can't do it via the distribution initval kwarg). Or did you mean something else?

So far it was **required** to specify the entries for the `start` dict in terms of transformed name keys and transformed values (unconstrained space).

We can change that API for v4.1 after putting a deprecation warning on the backwards compatible ways.
(My favorite would be to allow for constrained space names and tensor as keys, but to disallow the unconstrained space parametrizations.)
So this route:
* 4.0.0 start kwarg works as before, initvals kwarg gives deprecation warning on unconstrained space entries
* 4.1.0 start is removed and initvals kwarg only takes constrained space parametrizations with Union[Str, TensorVariable] keys

pymc/model.py
ricardoV94: I checked again, and the code does not change much if we use the `memo`, so not worth the bother.

ricardoV94:
Honestly, I think we changed so many things already that it's better to do all the breaking changes for 4.0.0, and add it to the V4 upgrade guide. In any case, I am okay with your final proposal (but would still vote to do it immediately for 4.0.0).

The last point of contention for me is whether we put the new initval logic in a separate file. For me, this feels a lot like specialized code that deserves its own file. The fact that you decided to create a separate test module, suggests you may feel the same. If anything, the new code would make more sense in `sampling.py` than in `model.py`.

michaelosthege:
@ricardoV94 @twiecki I moved more code into the `initvalues.py`. Please check it.
Let's do deprecation of accepting unconstrained RV names as keys in `initvals` in a follow-up since that work can be parallelized with many other things and likely goes together with migrating a ton of tests from unconstrained-name-keyed `start` to RV-tensor-keyed `initvals`.

twiecki (approved):

pymc/model.py
```diff
@@ -920,111 +927,112 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
-        rv_value_var = self.rvs_to_values[rv_var]
-        transform = getattr(rv_value_var.tag, "transform", None)
-
-        if initval is None or transform:
-            initval = self._eval_initval(rv_var, initval, test_value, transform)
-
-        self.initial_values[rv_value_var] = initval
-
-    def _eval_initval(
+    def make_initial_point_fn(
```
ricardoV94, line 944: This function can be merged with the make_expression one, no?

pymc/model.py
```diff
@@ -920,111 +927,112 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
+            jitter_rvs=jitter_rvs,
+            default_strategy=default_strategy,
+            return_transformed=return_transformed,
+        )
+        graph = FunctionGraph(outputs=initial_values, clone=False)
+
+        rng_nodes = find_rng_nodes(graph.outputs)
+        new_rng = [aesara.shared(np.random.RandomState(np.random.PCG64())) for _ in rng_nodes]
+        graph.replace_all(zip(rng_nodes, new_rng), import_missing=True)
+        func = compile_rv_inplace(inputs=[], outputs=graph.outputs)
```
ricardoV94: Should we use fast compile mode? Or keep the rewrites but compile to python still.

pymc/model.py
```diff
@@ -920,111 +927,112 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
         """Maps free variable names to transformed, numeric initial values."""
-        if set(self._initial_point_cache) != {get_var_name(k) for k in self.initial_values}:
-            return self.recompute_initial_point()
-        return self._initial_point_cache
+        seed = self.rng_seeder.randint(2 ** 30, dtype=np.int64)
+        return self.recompute_initial_point(seed)
 
-    def recompute_initial_point(self) -> Dict[str, np.ndarray]:
-        """Recomputes numeric initial values for all free model variables.
+    def recompute_initial_point(self, seed) -> Dict[str, np.ndarray]:
```
ricardoV94, line 933: I forgot, what's the point of compute and recompute methods, if we are not caching the results?

pymc/model.py
ricardoV94: Should this be merged with the make_expression function?

michaelosthege: Both functions are rather complex on their own already and the split gives us a more fine-grained test surface.

I don't want to rule out further refactoring in follow-ups though. In many ways this PR is just the first step into a new era of initvals and I find it hard to predict what changes we'll make when refactoring `init_nuts`.

pymc/model.py
michaelosthege: With `mode=Mode("py")` the tests run a little faster than with `mode=FAST_COMPILE`.
I'll push this change now...

ricardoV94: Strange, is the failing JAX test possibly related?

michaelosthege: Multiple tests from the pytest workflows also failed.
I just switched to `FAST_COMPILE` let's see if it's faster or slower than the successful CI run of the commit before (if it works).

pymc/sampling.py
```diff
@@ -2124,17 +2131,33 @@ def init_nuts(
...
     cb = [
         pm.callbacks.CheckParametersConvergence(tolerance=1e-2, diff="absolute"),
         pm.callbacks.CheckParametersConvergence(tolerance=1e-2, diff="relative"),
     ]
 
-    # TODO: Consider `initvals` for selecting the starting point.
+    initial_points = get_initial_points(
+        model=model,
+        initvals=initvals,
+        rvs_to_jitter={
```
ricardoV94: This should depend on the init method (e.g., "jitter+adapt_diag"). Also we should integrate with the `init_jitter` method which retries to sample if the jittering of constrained variables leads to -inf logp

We should not jitter variables if the method is not supposed to induce jittering

pymc/tests/test_model.py
```diff
@@ -532,68 +533,6 @@ def test_point_logps():
     assert "a" in logp_vals.keys()
 
 
-class TestUpdateStartVals(SeededTest):
```
ricardoV94, line 535: Should we port these tests to the new framework? We still want to make sure that we can update the start vals (not though the old method), and get the same results

ricardoV94 (commented):
Left two comments, and also codecov complains some branches are not being tested

pymc/sampling.py
michaelosthege: Yeah it also needs new API on step methods such that they can be queried.

Let's do it in the PR that takes care of init jitter. It is hard to predict what exactly the signature will be.

pymc/tests/test_model.py
michaelosthege: The test was about an advanced inplace updating of those ducts. That's something we should not do to begin with..

But there is a test where the initial_values dict is modified and initial point re-evaluated.

pymc/sampling.py
ricardoV94: I am not sure about waiting for another PR. We are breaking/changing the behavior of both jittering and non-jittering init methods.

pymc/sampling.py
```diff
@@ -2124,17 +2131,33 @@ def init_nuts(
...
+            rv
+            for rv in model.free_RVs
+            if hasattr(rv.tag, "transform")
+            # TODO: Also jitter untransformed RVs as long as they have a real-valued support
+        },
+        seeds=seeds,
+    )
```
ricardoV94:
```suggestion    
    initial_points = get_initial_points(
        model=model,
        initvals=initvals,
        rvs_to_jitter={},
        seeds=seeds,
    )
```

And then re-introduce the jitter again in another PR?

pymc/distributions/continuous.py
```diff
@@ -365,10 +365,13 @@ class Flat(Continuous):
 
     rv_op = flat
 
+    def __new__(cls, *args, **kwargs):
+        kwargs["initval"] = kwargs.get("initval", "moment")
```
lucianopaz:
```suggestion
        kwargs.setdefault("initval", "moment")
```

Small nitpick, but this could be done with [setdefault](https://docs.python.org/3/library/stdtypes.html#dict.setdefault)

pymc/distributions/continuous.py
```diff
@@ -430,10 +433,13 @@ class HalfFlat(PositiveContinuous):
 
     rv_op = halfflat
 
+    def __new__(cls, *args, **kwargs):
+        kwargs["initval"] = kwargs.get("initval", "moment")
```
lucianopaz:
```suggestion
        kwargs.setdefault("initval", "moment")
```

Same as for the `Flat` distribution

pymc/model.py
```diff
@@ -920,111 +927,114 @@ def test_point(self) -> Dict[str, np.ndarray]:
     @property
     def initial_point(self) -> Dict[str, np.ndarray]:
         """Maps free variable names to transformed, numeric initial values."""
-        if set(self._initial_point_cache) != {get_var_name(k) for k in self.initial_values}:
-            return self.recompute_initial_point()
-        return self._initial_point_cache
+        seed = self.rng_seeder.randint(2 ** 30, dtype=np.int64)
```
lucianopaz: I'm confused with this. Two successive calls to `initial_point` will return different points because the `seed` is randomly selected between calls. If `initial_point` is a model property, I would assume that it should be the same between calls, unless I somehow change the state of the model before trying to get the attribute again.
Why was the logic around the initial point cache dropped?

If we indeed want to leave the `initial_point` as something that returns random points, I suggest that we make it a regular method (not a model property), and add a `seed=None` argument to it. If `seed` is `None`, then we can draw a random seed as what is done now, but if the user passes a `seed`, the resulting initial point will always be the same between calls.

pymc/model.py
```diff
@@ -920,111 +927,114 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
         """
-        return self._initial_values
 
-    def set_initval(self, rv_var, initval):
-        if initval is not None:
-            initval = rv_var.type.filter(initval)
+        def find_rng_nodes(variables):
+            return [
+                node
+                for node in aesara.graph.basic.graph_inputs(graph.outputs)
```
lucianopaz:
```suggestion
                for node in aesara.graph.basic.graph_inputs(variables)
```

pymc/model.py
```diff
@@ -920,111 +927,114 @@ def test_point(self) -> Dict[str, np.ndarray]:
...
+            jitter_rvs=jitter_rvs,
+            default_strategy=default_strategy,
+            return_transformed=return_transformed,
+        )
+        graph = FunctionGraph(outputs=initial_values, clone=False)
 
-        rv_value_var = self.rvs_to_values[rv_var]
-        transform = getattr(rv_value_var.tag, "transform", None)
+        rng_nodes = find_rng_nodes(graph.outputs)
+        new_rng = [aesara.shared(np.random.RandomState(np.random.PCG64())) for _ in rng_nodes]
```
lucianopaz: Can't we already use a seeded sequence here? That would remove the need for defining a `make_seeded_function` lower in this function.

pymc/sampling.py
```diff
@@ -2124,17 +2131,33 @@ def init_nuts(
     if init == "auto":
         init = "jitter+adapt_diag"
 
-    _log.info(f"Initializing NUTS using {init}...")
+    if seeds is None:
+        seeds = model.rng_seeder.randint(2 ** 30, dtype=np.int64, size=chains)
+    if not isinstance(seeds, (list, tuple, np.ndarray)):
+        raise ValueError(f"The `seeds` must be array-like. Got {type(seeds)} instead.")
```
lucianopaz, line 2135: Maybe we can add a `   # pragma: no cover` to exclude this exception from code coverage. It seems silly to write a test exclusively for it.

pymc/sampling.py
```diff
@@ -2124,17 +2131,33 @@ def init_nuts(
...
+    if len(seeds) != chains:
+        raise ValueError(
```
lucianopaz, line 2137: Maybe we can add a `   # pragma: no cover` to exclude this exception from code coverage. It seems silly to write a test exclusively for it.

pymc/sampling.py
lucianopaz: I agree with @ricardoV94 here. We should try to cover it in this PR

pymc/initvalues.py
```diff
@@ -0,0 +1,212 @@
...
+def convert_start_dict_to_initval_dict(
+    model, start: StartDict
+) -> Dict[TensorVariable, Optional[Union[np.ndarray, Variable, str]]]:
+    """Helper function for converting a user-provided start dict with str keys of (transformed) variable names
+    to a dict mapping the RV tensors to untransformed initvals.
+    """
+    initvals = {}
+    for key, initval in start.items():
+        if is_transformed_name(key):
+            rv = model[get_untransformed_name(key)]
```
lucianopaz: I feel that it's important to write a test for this function when a user passes values in the natural space of a transformed variable. For example

```python
with pm.Model():

convert_start_dict_to_initval_dict({"sigma": 4})
```

pymc/initvalues.py
```diff
@@ -0,0 +1,212 @@
...
+    rvs_to_jitter: Set[TensorVariable],
+    seeds: Sequence[int],
+) -> PointType:
+    """Evaluates initial points for multiple chains.
+
+    Parameters
+    ----------
+    initvals : optional, list or dict
+        Initial value strategy overrides that should take precedence over the defaults from the model.
+        Can be None or dict will apply to all chains.
```
lucianopaz: Should this line in the docstring be removed? I find it very confusing, but it seems like the following line explains the same thing but much better.

pymc/initvalues.py
```diff
@@ -0,0 +1,212 @@
...
+            model.make_initial_point_fn(
+                jitter_rvs=rvs_to_jitter,
+                return_transformed=True,
+                overrides=convert_start_dict_to_initval_dict(model, chain_start or None),
+            )
+            for chain_start in initvals
+        ]
+    else:
+        raise ValueError(
+            f"Number of start dicts ({len(initvals)}) does not match the number of seeds ({len(seeds)})."
```
lucianopaz: Should we test this? If not, we can always add a `# pragma: no cover` directive

pymc/initvalues.py
```diff
@@ -0,0 +1,212 @@
...
+    return start_points
+
+
+def make_initial_point_expression(
+    *,
+    free_rvs: Sequence[TensorVariable],
+    rvs_to_values: Dict[TensorVariable, TensorVariable],
+    initval_strategies: Dict[TensorVariable, Optional[Union[np.ndarray, Variable, str]]],
+    jitter_rvs: Set[TensorVariable] = None,
+    default_strategy: str = "prior",
```
lucianopaz: Shouldn't the default strategy be moment as in v3?

pymc/initvalues.py
```diff
@@ -0,0 +1,212 @@
...
+            value = at.as_tensor(strategy, dtype=variable.dtype).astype(variable.dtype)
+
+        transform = getattr(rvs_to_values[variable].tag, "transform", None)
+
+        if transform is not None:
+            value = transform.forward(variable, value)
+
+        if variable in jitter_rvs:
+            jitter = at.random.uniform(at.zeros_like(variable) - 1, at.zeros_like(variable) + 1)
+            jitter.name = "jitter"
```
lucianopaz:
```suggestion
            jitter.name = variable.name + "_jitter"
```

Maybe we could prefix the jitter with the variable's name?

pymc/initvalues.py
```diff
@@ -0,0 +1,212 @@
...
+    for i in range(n_variables):
+        outputs = [initial_values_clone[i], initial_values_transformed_clone[i]]
+        graph = FunctionGraph(outputs=outputs, clone=False)
+        graph.replace_all(zip(free_rvs_clone[:i], initial_values), import_missing=True)
+        initial_values.append(graph.outputs[0])
+        initial_values_transformed.append(graph.outputs[1])
+
+    if return_transformed:
+        return initial_values_transformed
+    return initial_values
```
lucianopaz: We should definitely add a test for `return_transformed=False`.

pymc/tests/test_initvals.py
```diff
@@ -38,59 +43,87 @@ def test_new_warnings(self):
...
+            B = pm.Uniform("B", lower=L, upper=2, initval=1.25)
+            ip = pmodel.recompute_initial_point(seed=0)
+            assert ip["L_interval__"] == 0
+            assert ip["B_interval__"] == 0
+
+            # Modify initval of L and re-evaluate
+            pmodel.initial_values[L] = 0.9
+            ip = pmodel.recompute_initial_point(seed=0)
+            assert ip["B_interval__"] < 0
         pass
```
lucianopaz: We should delete the `pass`, no?

pymc/tests/test_initvals.py
```diff
@@ -38,59 +43,87 @@ def test_new_warnings(self):
...
+            data = aesara.shared(np.arange(4))
+            rv = pm.Uniform("u", lower=data, upper=10, initval="prior")
+
+            ip = pmodel.recompute_initial_point(seed=0)
+            assert np.shape(ip["u_interval__"]) == (4,)
+
+            data.set_value(np.arange(5))
+            ip = pmodel.recompute_initial_point(seed=0)
+            assert np.shape(ip["u_interval__"]) == (5,)
         pass
```
lucianopaz: We should remove the `pass`

pymc/tests/test_initvals.py
```diff
@@ -38,59 +43,87 @@ def test_new_warnings(self):
...
+        assert iv["A"] == 0
+        # Moment of the HalfFlat is 1, but HalfFlat is log-transformed by default
+        # so the transformed initial value with jitter will be zero plus a jitter between [-1, 1].
+        b_transformed = iv["B_log__"]
+        b_untransformed = transform_back(B, b_transformed)
+        assert b_transformed != 0
+        assert -1 < b_transformed < 1
+        # C is centered on 0 + untransformed initval of B
+        assert iv["C"] == np.array(0 + b_untransformed, dtype=aesara.config.floatX)
+        pass
```
lucianopaz:
```suggestion
```

pymc/tests/test_initvals.py
```diff
@@ -38,59 +43,87 @@ def test_new_warnings(self):
...
+                return_transformed=True,
+                overrides={
+                    A: at.as_tensor(2, dtype=int),
+                    C: 5,
+                },
+            )
+            iv = fn(0)
+        assert iv["A"] == 2
+        assert iv["C"] == 5
         pass
```
lucianopaz: We should remove `pass`

lucianopaz (dismissed):
Sorry it took me so long to review. This looks really great. I left a bunch of comments in the middle

pymc/initvalues.py
michaelosthege: We can't do that until we implemented moments for a ton of RVs. (out of scope of this PR)

michaelosthege:
Thanks for the additional comments @lucianopaz @ricardoV94 .
I have run out of energy on this PR, so I'll have to defer the vast majority of them to follow-ups. I hope you understand

pymc/sampling.py
michaelosthege: Before these changes, we had a wild mix of when the point returned by `init_nuts` was used or not. At least these changes make that consistent.
But I am not familiar with `init_nuts` and the 9 (!) different init strategies it implements. I don't know how decide between using the `init` result or the `get_initial_points` result, when to apply jitter and when not to.
This `init_nuts`/jittering needs input from, for example @aseyboldt who is familiar with the math behind those initializations.

I am not qualified to make the necessary decisions here and IMHO this should not block the PR from moving forward, since it at least makes the choice of initial points more consistent to begin with.

pymc/model.py
michaelosthege: The method you propose is `Model.recompute_initial_point(seed)`. I will change it to default `seed=None` and draw the seed there.

The `test_point` and `initial_point` properties continue to exist for compatibility (they're used all across the codebase).
I'll add the same `DeprecationWarning` we have for `test_point` also to `initial_point`. It will be annoying as hell though.

pymc/initvalues.py
ricardoV94: moment should just default to random/prior if not implemented

ricardoV94:
I don't think we should merge the PR if there are valid concerns with its current state. I don't mind taking over if you don't have the time/patience right now

pymc/model.py
michaelosthege: I was not involved when this code was written and can't quite follow what it does.
@ricardoV94 can you help us out here?

To me it looks like we can at least un-indent the `make_seeded_function` since it's only called once anyway?
Does the code inside the `def inner` replace the rngs in the graph of the already-compiled function and that's why at first it must be built with placeholder rngs?

pymc/initvalues.py
michaelosthege: > moment should just default to random/prior if not implemented

Moment defaults to `NotImplementedError`. Anything else is like asking for trouble, don't you think? I mean if you do `pm.get_moment()` and it gives you a random prior draw that's like... super unexpected and hard to find out what happened

ricardoV94: Ok then when we call it we can have a try except block. 

I don't think we had moments for all dists in the past so forcing it would be a bit too restrictive (or did we)

michaelosthege: We had "_defaultval" for a lot of dists and I don't know how we handled the others.

My thinking about the default so far was this:
1. This PR implements the framework that makes it even possible
2. Next, we implement `get_moment` for a ton of distributions.
3. Next, we set `default_strategy="moment"` and for all distributions that don't have `get_moment`, we put a `kwargs.setdefault("initval", "prior")`, or `initval="prior"` into their signatures.

Then at least all behavior will be defined. Also users who put `initval="moment"` on a dist that doesn't have it _will_ get the `NotImplementedError`.

pymc/sampling.py
michaelosthege: I like the test more than these black magic comments..

pymc/initvalues.py
michaelosthege: I don't understand what your example is testing there that's not already covered by the existing tests?

Other than that, the `convert_start_dict_to_initval_dict` will be looked at in the follow-up PR that deprecates passing of transformed initvals.

pymc/tests/test_initvals.py
michaelosthege: I like to put `pass` at the end of tests to mark the end of the code block both visually and for the IDE. (Comments after the last "real" line don't collapse correctly.)
Like a `return` statement it makes clear that the code block ended, in contrast to just leaving it hanging at the 3rd or 4th indentation level. (I always wonder if code got deleted accidentally.)

If anybody strongly objects feel free to push the branch. Though I'd like to point out that both `pylint` and `black` are fine with it.

michaelosthege:
As you can see I just pushed several changes that dealt with suggestions you made.
At this point please feel free to take over, @ricardoV94.

Ping me when this thing is merged. Until then I'm unsubscribing notifications from this topic.

michaelosthege:
I just rebased this onto latest `main` to resolve conflicts.
IMHO we should merge once green and create an issue for `init_nuts` refactoring.

pymc/model.py
ricardoV94: > To me it looks like we can at least un-indent the `make_seeded_function` since it's only called once anyway?
> Does the code inside the `def inner` replace the rngs in the graph of the already-compiled function and that's why at first it must be built with placeholder rngs?

Yes.

This way seems more flexible in that you don't need to recompile the function if you want to recompute with different seeds. This is useful already in `init_nuts` and will be useful for defining the initial_points in SMC as well

pymc/initvalues.py
ricardoV94: We probably(?) already have a test that covers this (or did codecov complain?) since it was old functionality. But should double-check

ricardoV94: I modified the `test_respects_overrides` to cover this situation. We will want to deprecate this, but for now it's tested

pymc/initvalues.py
```diff
@@ -0,0 +1,320 @@
+#   Copyright 2021 The PyMC Developers
```
twiecki: should we name this `init_values.py`?

twiecki:
LGTM. What happens when an init point is -inf?

ricardoV94:
> LGTM. What happens when an init point is -inf?

What do you mean? The model logp at the initial point? If so, same as before

pymc/initvalues.py
ricardoV94: Or `initial_point` or just `point`

michaelosthege:
Both pre-commit and docs worked on master. But that should be easy to fix(?)

pymc/initvalues.py
ricardoV94: Renamed it to `initial_point.py`

ricardoV94:
> Both pre-commit and docs worked on master. But that should be easy to fix(?)

Rebased after #5070. Should be fixed now

pymc/initial_point.py
```diff
@@ -0,0 +1,320 @@
...
+    ----------
+    step : BlockedStep or CompoundStep
+        One or many step methods that were assigned model variables.
+
+    Returns
+    -------
+    rvs_to_jitter : set
+        The random variables for which jitter should be added.
+    """
+    # TODO: implement this
```
michaelosthege, line 67: Created an issue for this: https://github.com/pymc-devs/pymc/issues/5077

pymc/initial_point.py
```diff
@@ -0,0 +1,320 @@
...
+            node
+            for node in graph_inputs(variables)
+            if isinstance(
+                node,
+                (
+                    at.random.var.RandomStateSharedVariable,
+                    at.random.var.RandomGeneratorSharedVariable,
+                ),
+            )
+        ]
```
michaelosthege, line 161: Can be extracted

michaelosthege (commented):
I can't approve since I'm the original author.

That one nitpick comment I made shouldn't hold us back from anything

twiecki:
🥳

PR #4984, opened 2021-09-05
[CU-a73n08, Docs] Glossary - Dispersion

CloudChaoszero:
# Description
* Add "Dispersion"-related definitions in PyMC glossary
* [Optional]: Creating template for adding terms into Glossary

[CU-5t5y0p Glossary](https://app.clickup.com/t/5t5y0p)

PR #4987, opened 2021-09-07
CU-5t5y0p Adds to glossary prior, posterior and related definitions

olgadk7:
Adds Prior, Posterior (+ Likelihood, Bayesian model, and Bayesian data analysis) definitions to docs/source/glossary.md

### DIFF
--- a/docs/source/glossary.md
+++ b/docs/source/glossary.md
@@ -21,17 +21,17 @@ Dispatching
 Bayesian model
   A Bayesian model is a composite of variables and distributional definitions for these variables. Fundamentally, models are mappings of one set of variables through a probability distribution onto another set of variables. These models define the ways values of some variables can arise, given values of other variables.
   
-Bayesian data analysis
-  Once we have defined the statistical model, Bayesian data analysis logically processes the data to produce inference. Probability theory is used to represent plausibility, whether in reference to countable events in the world, or theoretical constructs like parameters.
+Bayesian inference
+  Once we have defined the statistical model, Bayesian inference processes the data and model to produce a posterior distribution. that is a joint distribution of all parameters in the model. This distribution is used to represent plausibility, and is the logical consequence of the model and data.
   
 Likelihood
   The probability of the data, or the relative number of ways that a value of a parameter can produce the data, is called a likelihood. It is derived by enumerating all the possible data sequences that could have happened and then eliminating those sequences inconsistent with the data.
   
 Prior
-  Bayesian statistics allow us to include all information we have about the structure of the problem into the model. We do this via assuming prior distributions of the model’s variables, or the priors. Priors represent the plausibility of each possible value of the parameters before accounting for the data. Priors multiplied by likelihood produces the posterior. 
+  Bayesian statistics allow us, in principle, to include all information we have about the structure of the problem into the model. We can do this via assuming prior distributions of the model’s parameters. Priors represent the plausibility of the value of the parameters before accounting for the data. Priors multiplied by likelihood produce the posterior. 
   
 Posterior
-  The outcome of a Bayesian model is the posterior probability, the updated plausibility of any specific target of inference after the model has seen the data. It produces the posterior distribution, which is a ranking of the relative plausibilities of every possible combination of parameter values.
+  The outcome of a Bayesian model is the posterior distribution, which describes the relative plausibilities of every possible combination of parameter values. We can think the posterior as the updated priors after the model has seen the data.
 
   Once your model produces a posterior distribution, you need to summarize and interpret it. Common questions for the posterior include: intervals of defined boundaries, intervals of defined probability mass, point estimates. When the posterior distribution is very similar to the prior, the available data does not contain much information about a parameter of interest.
 
--- a/docs/source/glossary.md
+++ b/docs/source/glossary.md
@@ -17,22 +17,45 @@ Functional Programming
   This contrasts with functions or methods that depend on variables that are not explicitly passed as an input (such as accessing `self.variable` inside a method) or that alter the inputs or other state variables in-place, instead of returning new distinct variables as outputs.
 Dispatching
   Choosing which function or method implementation to use based on the type of the input variables (usually just the first variable). For some examples, see Python's documentation for the [singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) decorator.
-  
-Bayesian model
-  A Bayesian model is a composite of variables and distributional definitions for these variables. Fundamentally, models are mappings of one set of variables through a probability distribution onto another set of variables. These models define the ways values of some variables can arise, given values of other variables.
-  
-Bayesian inference
-  Once we have defined the statistical model, Bayesian inference processes the data and model to produce a posterior distribution. that is a joint distribution of all parameters in the model. This distribution is used to represent plausibility, and is the logical consequence of the model and data.
-  
-Likelihood
-  The probability of the data, or the relative number of ways that a value of a parameter can produce the data, is called a likelihood. It is derived by enumerating all the possible data sequences that could have happened and then eliminating those sequences inconsistent with the data.
-  
-Prior
-  Bayesian statistics allow us, in principle, to include all information we have about the structure of the problem into the model. We can do this via assuming prior distributions of the model’s parameters. Priors represent the plausibility of the value of the parameters before accounting for the data. Priors multiplied by likelihood produce the posterior. 
-  
-Posterior
-  The outcome of a Bayesian model is the posterior distribution, which describes the relative plausibilities of every possible combination of parameter values. We can think the posterior as the updated priors after the model has seen the data.
-
-  Once your model produces a posterior distribution, you need to summarize and interpret it. Common questions for the posterior include: intervals of defined boundaries, intervals of defined probability mass, point estimates. When the posterior distribution is very similar to the prior, the available data does not contain much information about a parameter of interest.
 
 :::::
+
+```{glossary}
+{term}`Bayesian Workflow`
+  Bayesian workflow is the overall iterative procedure towards model refinement. It often includes the two related tasks of inference and the exploratory analysis of models.
+  - For a compact overview, see Bayesian statistics and modelling by van de Schoot, R., Depaoli, S., King, R. et al in Nat Rev Methods     - Primers 1, 1 (2021). https://doi.org/10.1038/s43586-020-00001-2
+  - For an in-depth overview, see Bayesian Workflow by Andrew Gelman, Aki Vehtari, Daniel Simpson, Charles C. Margossian, Bob Carpenter, Yuling Yao, Lauren Kennedy, Jonah Gabry, Paul-Christian Bürkner, Martin Modrák
+  - For an exercise-based material, see Think Bayes 2e: Bayesian Statistics Made Simple by Allen B. Downey 
+  - For an upcoming textbook that uses PyMC3, Tensorflow Probability, and ArviZ libraries, see Bayesian Modeling and Computation by Osvaldo A. Martin, Ravin Kumar, Junpeng Lao
+
+Bayesian inference
+  Once we have defined the statistical model, Bayesian inference processes the data and model to produce a posterior distribution. That is a joint distribution of all parameters in the model. This distribution is used to represent plausibility, and is the logical consequence of the model and data.
+
+Bayesian model
+  A Bayesian model is a composite of variables and distributional definitions for these variables. Fundamentally, it tells you all the ways that the observed data could have been produced.
+  
+Prior
+  Bayesian statistics allow us, in principle, to include all information we have about the structure of the problem into the model. We can do this via assuming prior distributions of the model’s parameters. Priors represent the plausibility of the value of the parameters before accounting for the data. Priors multiplied by likelihood produce the posterior.
+  
+  Priors’ informativeness can fall anywhere on the complete uncertainty to relative certainty continuum. An informative prior might encode known restrictions on the possible range of values of that parameter.
+  
+  To understand the implications of a prior, as well as the model itself, we can simulate predictions from the model, using only the prior distribution instead of the posterior distribution - a process sometimes referred to as prior predictive simulation.
+  
+  - For an in-depth guide to priors, consider Statistical Rethinking 2nd Edition By Richard McElreath, especially chapters 2.3
+
+Likelihood
+  There are many perspectives on likelihood, but conceptually we can think about it as the relative number of ways the model could have produced the data; in other words, the probability of the data, given the parameters.
+  
+  - For an in-depth unfolding of the concept, refer to Statistical Rethinking 2nd Edition By Richard McElreath, particularly chapter 2.
+  - For the problem-based material, see Think Bayes 2e: Bayesian Statistics Made Simple by Allen B. Downey 
+  - For univariate, continuous scenarios, see the calibr8 paper: Bayesian calibration, process modeling and uncertainty quantification in biotechnology by Laura Marie Helleckes,  Michael Osthege, Wolfgang Wiechert, Eric von Lieres, Marco Oldiges 
+  
+Posterior
+  The outcome of a Bayesian model is the posterior distribution, which describes the relative plausibilities of every possible combination of parameter values. We can think of the posterior as the updated priors after the model has seen the data.
+  
+  When the posterior is obtained using numerical methods we first need to check how adequately the model fits to data. By sampling from the posterior distribution we can simulate the observations, or the implied predictions of the model. This posterior predictive distribution can then be compared to the observed data, the process known as the posterior predictive check.
+
+  Once you are satisfied with the model, posterior distribution can be summarized and interpreted. Common questions for the posterior include: intervals of defined boundaries, intervals of defined probability mass, and point estimates. When the posterior is very similar to the prior, the available data does not contain much information about a parameter of interest.
+  
+  - For more on generating and interpreting the posterior samples, see Statistical Rethinking 2nd Edition By Richard McElreath, chapter 3.
+```
--- a/docs/source/glossary.md
+++ b/docs/source/glossary.md
@@ -17,4 +17,22 @@ Functional Programming
   This contrasts with functions or methods that depend on variables that are not explicitly passed as an input (such as accessing `self.variable` inside a method) or that alter the inputs or other state variables in-place, instead of returning new distinct variables as outputs.
 Dispatching
   Choosing which function or method implementation to use based on the type of the input variables (usually just the first variable). For some examples, see Python's documentation for the [singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) decorator.
+  
+Bayesian model
+  A Bayesian model is a composite of variables and distributional definitions for these variables. Fundamentally, models are mappings of one set of variables through a probability distribution onto another set of variables. These models define the ways values of some variables can arise, given values of other variables.
+  
+Bayesian data analysis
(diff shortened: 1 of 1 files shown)


OriolAbril (commented):
Same general comment as in https://github.com/pymc-devs/pymc3/pull/4984#pullrequestreview-748430155, I haven't had time to go over the definitions in detail yet

docs/source/glossary.md
```diff
@@ -17,4 +17,22 @@ Functional Programming
...
 Dispatching
   Choosing which function or method implementation to use based on the type of the input variables (usually just the first variable). For some examples, see Python's documentation for the [singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) decorator.
+  
+Bayesian model
+  A Bayesian model is a composite of variables and distributional definitions for these variables. Fundamentally, models are mappings of one set of variables through a probability distribution onto another set of variables. These models define the ways values of some variables can arise, given values of other variables.
+  
+Bayesian data analysis
+  Once we have defined the statistical model, Bayesian data analysis logically processes the data to produce inference. Probability theory is used to represent plausibility, whether in reference to countable events in the world, or theoretical constructs like parameters.
```
michaelosthege: A popular book to link would be "Think Bayes".

Also https://www.nature.com/articles/s43586-020-00001-2 is very comprehensive about the overall workflow.

docs/source/glossary.md
```diff
@@ -17,4 +17,22 @@ Functional Programming
...
+  
+Likelihood
+  The probability of the data, or the relative number of ways that a value of a parameter can produce the data, is called a likelihood. It is derived by enumerating all the possible data sequences that could have happened and then eliminating those sequences inconsistent with the data.
```
michaelosthege: The likelihood is sooo important, but there are also many different perspectives.
Best case we could link to both discrete and continuous, univariate and multivariate likelihoods in How-To guides and literature.
I haven't read it, but "Regression and other stories" probably has a chapter about likelihoods.

One resource for likelihood in univariate, continuous scenarios could be our [`calibr8` paper](https://www.biorxiv.org/content/10.1101/2021.06.30.450546v2). I hate to advertise my own work here, but we kinda wrote that thing because the topic wasn't covered from that perspective before.

itsguneetsingh:
For some reason I am not able to make a PR to patch1 branch of @olgadk7 repo. So I suppose I should just make a suggestion for the definitions here. Is that ok @michaelosthege ?

olgadk7:
Will be adding sources for the definitions in the next couple of days finally, thank you.

docs/source/glossary.md
aloctavodia:
```suggestion
Bayesian inference
  Once we have defined the statistical model, Bayesian inference processes the data and model to produce a posterior distribution. that is a joint distribution of all parameters in the model. This distribution is used to represent plausibility, and is the logical consequence of the model and data.
```

aloctavodia: I think we should restrict Bayesian inference to the computation of posteriors and Bayesian data analysis or Bayesian workflow to the overall iterative modelling procedure. This is in line with https://arxiv.org/abs/2011.01808 and our upcoming book https://www.routledge.com/Bayesian-Modeling-and-Computation-in-Python/Martin-Kumar-Lao/p/book/9780367894368 In this book and also in ArviZ we divide the Bayesian Workflow in two related tasks Bayesian Inference and Exploratory analysis of Bayesian models, but we acknowledge that this last term is not widely accepted.

docs/source/glossary.md
aloctavodia: I will clarify that is the probability  of the data given the parameters p(D| \theta) , if we average over the prior then we get the marginal likelihood p(D).

I agree with @michaelosthege  maybe we can say "conceptually we can think about it as  enumerating all the possible data sequences that could have happened and then eliminating those sequences inconsistent with the data." so we do not try to provide an strict definition but more a convenient way to think about it?...

docs/source/glossary.md
```diff
@@ -17,4 +17,22 @@ Functional Programming
...
+  
+Prior
+  Bayesian statistics allow us to include all information we have about the structure of the problem into the model. We do this via assuming prior distributions of the model’s variables, or the priors. Priors represent the plausibility of each possible value of the parameters before accounting for the data. Priors multiplied by likelihood produces the posterior.
```
aloctavodia:
```suggestion
  Bayesian statistics allow us, in principle, to include all information we have about the structure of the problem into the model. We can do this via assuming prior distributions of the model’s parameters. Priors represent the plausibility of the value of the parameters before accounting for the data. Priors multiplied by likelihood produce the posterior. 
```

docs/source/glossary.md
```diff
@@ -17,4 +17,22 @@ Functional Programming
...
+  
+Likelihood
+  The probability of the data, or the relative number of ways that a value of a parameter can produce the data, is called a likelihood. It is derived by enumerating all the possible data sequences that could have happened and then eliminating those sequences inconsistent with the data.
+  
+Prior
+  Bayesian statistics allow us to include all information we have about the structure of the problem into the model. We do this via assuming prior distributions of the model’s variables, or the priors. Priors represent the plausibility of each possible value of the parameters before accounting for the data. Priors multiplied by likelihood produces the posterior. 
+  
+Posterior
+  The outcome of a Bayesian model is the posterior probability, the updated plausibility of any specific target of inference after the model has seen the data. It produces the posterior distribution, which is a ranking of the relative plausibilities of every possible combination of parameter values.
```
aloctavodia:
```suggestion
Posterior
  The outcome of a Bayesian model is the posterior distribution, which describes the relative plausibilities of every possible combination of parameter values. We can think the posterior as the updated priors after the model has seen the data.

File: docs/source/glossary.md [hunk 5]
```diff
@@ -17,4 +17,22 @@ Functional Programming
...
+
+  Once your model produces a posterior distribution, you need to summarize and interpret it. Common questions for the posterior include: intervals of defined boundaries, intervals of defined probability mass, point estimates. When the posterior distribution is very similar to the prior, the available data does not contain much information about a parameter of interest.
+
```
aloctavodia, line 55: We could mention here that when the posterior is obtained using numerical methods we first need to run diagnostic to detect problems.

docs/source/glossary.md
```diff
@@ -17,4 +17,22 @@ Functional Programming
   This contrasts with functions or methods that depend on variables that are not explicitly passed as an input (such as accessing `self.variable` inside a method) or that alter the inputs or other state variables in-place, instead of returning new distinct variables as outputs.
 Dispatching
   Choosing which function or method implementation to use based on the type of the input variables (usually just the first variable). For some examples, see Python's documentation for the [singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) decorator.
+  
+Bayesian model
+  A Bayesian model is a composite of variables and distributional definitions for these variables. Fundamentally, models are mappings of one set of variables through a probability distribution onto another set of variables. These models define the ways values of some variables can arise, given values of other variables.
```
martinacantaro: Include reference to Statistical Rethinking since there are textual quotes from the book in this definition. As a general rule, all textual quotes should be attributed.

martinacantaro:
Task linked: [CU-5t5y0p Glossary](https://app.clickup.com/t/5t5y0p)

docs/source/glossary.md
olgadk7: Added, thank you.

docs/source/glossary.md
olgadk7: Thank you, re-worded (and minimized).

docs/source/glossary.md
```diff
@@ -17,4 +17,45 @@ Functional Programming
   This contrasts with functions or methods that depend on variables that are not explicitly passed as an input (such as accessing `self.variable` inside a method) or that alter the inputs or other state variables in-place, instead of returning new distinct variables as outputs.
 Dispatching
   Choosing which function or method implementation to use based on the type of the input variables (usually just the first variable). For some examples, see Python's documentation for the [singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) decorator.
+
 :::::
+
+```{glossary}
```
OriolAbril:
```suggestion

```

the glossary is defined above with the `:::::{glossary}` thingy, all term definitions should be within the colon fences (added suggestion below too to close it). This looks right in the website but it's actually creating a 2nd glossary which would give problems when linking from other parts of the docs to here. Sphinx does support multiple glossaries and they can have special labels (and should but it's not required) which is why it builds instead of erroring out, but I don't think we should have 2 glossaries.

Also, bit of a side note, should we update the `:::::{glossary}` thing to be 

```
:::::{glossary}
:sorted:
```

so terms are alphabetically sorted? reference https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-glossary

docs/source/glossary.md
```diff
@@ -17,4 +17,45 @@ Functional Programming
...
+{term}`Bayesian Workflow`
```
OriolAbril: This links to itself. The unindented terms in the glossary are added by sphinx as cross-reference targets. You can then use ``{term}`glossary term` `` to link to that term from anywhere in the docs (even definitions of other terms or of itself)

docs/source/glossary.md
```diff
@@ -17,4 +17,45 @@ Functional Programming
...
+  - For a compact overview, see Bayesian statistics and modelling by van de Schoot, R., Depaoli, S., King, R. et al in Nat Rev Methods     - Primers 1, 1 (2021). https://doi.org/10.1038/s43586-020-00001-2
+  - For an in-depth overview, see Bayesian Workflow by Andrew Gelman, Aki Vehtari, Daniel Simpson, Charles C. Margossian, Bob Carpenter, Yuling Yao, Lauren Kennedy, Jonah Gabry, Paul-Christian Bürkner, Martin Modrák
+  - For an exercise-based material, see Think Bayes 2e: Bayesian Statistics Made Simple by Allen B. Downey 
+  - For an upcoming textbook that uses PyMC3, Tensorflow Probability, and ArviZ libraries, see Bayesian Modeling and Computation by Osvaldo A. Martin, Ravin Kumar, Junpeng Lao
+
+Bayesian inference
+  Once we have defined the statistical model, Bayesian inference processes the data and model to produce a posterior distribution. That is a joint distribution of all parameters in the model. This distribution is used to represent plausibility, and is the logical consequence of the model and data.
```
OriolAbril:
```suggestion
  Once we have defined the statistical model, Bayesian inference processes the data and model to produce a {term}`posterior` distribution. That is a joint distribution of all parameters in the model. This distribution is used to represent plausibility, and is the logical consequence of the model and data.
```

this will make posterior be clickable and point to the definition of the posterior term

docs/source/glossary.md
```diff
@@ -17,4 +17,45 @@ Functional Programming
...
+  
+Posterior
+  The outcome of a Bayesian model is the posterior distribution, which describes the relative plausibilities of every possible combination of parameter values. We can think of the posterior as the updated priors after the model has seen the data.
+  
+  When the posterior is obtained using numerical methods we first need to check how adequately the model fits to data. By sampling from the posterior distribution we can simulate the observations, or the implied predictions of the model. This posterior predictive distribution can then be compared to the observed data, the process known as the posterior predictive check.
+
+  Once you are satisfied with the model, posterior distribution can be summarized and interpreted. Common questions for the posterior include: intervals of defined boundaries, intervals of defined probability mass, and point estimates. When the posterior is very similar to the prior, the available data does not contain much information about a parameter of interest.
+  
+  - For more on generating and interpreting the posterior samples, see Statistical Rethinking 2nd Edition By Richard McElreath, chapter 3.
+```
```
OriolAbril:
```suggestion
:::::
```

docs/source/glossary.md
martinacantaro: Agree with sorting alphabetically!

martinacantaro:
> For some reason I am not able to make a PR to patch1 branch of @olgadk7 repo. So I suppose I should just make a suggestion for the definitions here. Is that ok @michaelosthege ?

HI @Guneetconvent2002 ! Let me know if you need anything!

docs/source/glossary.md
olgadk7: Right, noted. I guess I tried the backquotes as it was the format to be followed linked in the instructions (and because the initial glossary.md file with colon fence isn't rendering right on github (though it does on readthedocs, which is what's important)). Curious though, where does the colon fence format comes from (looked up [Add a Glossary](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/glossary.html) instructions but it's not there).

OriolAbril (approved):
I think the definitions look great already, and we can always continue improving them iteratively. 

I see two options now. @olgadk7 if you have time today or tomorrow, it would be great to add the `:sorted:` thing right below the start of the glossary directive, if you have time also add some extra cross-references between terms (i.e. the prior definition references he likelihood and the posterior and could be linked, things like that). Otherwise I think we should merge now and take care of sorting in a follow-up PR

olgadk7:
@OriolAbril yep done, thank you.

docs/source/glossary.md
OriolAbril: The colon fence format comes from MyST. It supports both backtick and colon fences if configured: https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#code-fences-using-colons. I prefer the colons for text heavy directives so even with a simple markdown renderer, most of the formatting looks good already. Specific myst things will still look as text and generally bad, but cursive, code, links... will already look good

michaelosthege (approved):

Guneetconvent2002:
```suggestion
[Equidispersion](https://www.researchgate.net/publication/321375217_Extended_Poisson_INAR1_processes_with_equidispersion_underdispersion_and_overdispersion)
  If in a Poisson distribution if the variance equals the mean of the distribution, it is reffered to as equidispersion.
```

docs/source/glossary.md
```diff
@@ -16,5 +17,43 @@ Functional Programming
...
+  
+Posterior
+  The outcome of a Bayesian model is the posterior distribution, which describes the relative plausibilities of every possible combination of parameter values. We can think of the posterior as the updated {term}`priors` after the model has seen the data.
+  
+  When the posterior is obtained using numerical methods we first need to check how adequately the model fits to data. By sampling from the posterior distribution we can simulate the observations, or the implied predictions of the model. This posterior predictive distribution can then be compared to the observed data, the process known as the posterior predictive check.
+
+  Once you are satisfied with the model, posterior distribution can be summarized and interpreted. Common questions for the posterior include: intervals of defined boundaries, intervals of defined probability mass, and point estimates. When the posterior is very similar to the prior, the available data does not contain much information about a parameter of interest.
+  
+  - For more on generating and interpreting the posterior samples, see Statistical Rethinking 2nd Edition By Richard McElreath, chapter 3.
 :::::
```
Guneetconvent2002, line 59: [Equidispersion](https://www.researchgate.net/publication/321375217_Extended_Poisson_INAR1_processes_with_equidispersion_underdispersion_and_overdispersion)
  If in a Poisson distribution if the variance equals the mean of the distribution, it is reffered to as equidispersion.

[Generalized Poisson PMF](https://www.jstor.org/stable/1267389)
  A generalization of the {term}`Poisson distribution`, with two parameters X1, and X2, is obtained as a limiting form of the {term}`generalized negative binomial distribution`. The variance of the distribution is greater than, equal to or smaller than the mean according as X2 is positive, zero or negative. For formula and more detail, visit the link in the title.

[Bayes' theorem](https://en.wikipedia.org/wiki/Bayes%27_theorem)
  Describes the probability of an event, based on prior knowledge of conditions that might be related to the event. For example, if the risk of developing health problems is known to increase with age, Bayes' theorem allows the risk to an individual of a known age to be assessed more accurately (by conditioning it on their age) than simply assuming that the individual is typical of the population as a whole.
  > Formula:
  > $$
  > P(A|B) = \frac{P(B|A) P(A}{Pr(B)}
  > $$
  > Where A and B are events and P(B) != 0
  

[Markov Chain](https://en.wikipedia.org/wiki/Markov_chain)
  A Markov chain or Markov process is a stochastic model describing a sequence of possible events in which the probability of each event depends only on the state attained in the previous event.

[Markov Chain Monte Carlo](https://en.wikipedia.org/wiki/Markov_chain_Monte_Carlo)
[MCMC]
  Markov chain Monte Carlo (MCMC) methods comprise a class of algorithms for sampling from a probability distribution. By constructing a {term}`Markov Chain` that has the desired distribution as its equilibrium distribution, one can obtain a sample of the desired distribution by recording states from the chain.  Various algorithms exist for constructing chains, including the Metropolis–Hastings algorithm.

docs/source/glossary.md
```diff
@@ -16,5 +17,43 @@ Functional Programming
...
+  
+Posterior
+  The outcome of a Bayesian model is the posterior distribution, which describes the relative plausibilities of every possible combination of parameter values. We can think of the posterior as the updated {term}`priors` after the model has seen the data.
+  
+  When the posterior is obtained using numerical methods we first need to check how adequately the model fits to data. By sampling from the posterior distribution we can simulate the observations, or the implied predictions of the model. This posterior predictive distribution can then be compared to the observed data, the process known as the posterior predictive check.
+
+  Once you are satisfied with the model, posterior distribution can be summarized and interpreted. Common questions for the posterior include: intervals of defined boundaries, intervals of defined probability mass, and point estimates. When the posterior is very similar to the prior, the available data does not contain much information about a parameter of interest.
+  
+  - For more on generating and interpreting the posterior samples, see Statistical Rethinking 2nd Edition By Richard McElreath, chapter 3.
```
itsguneetsingh, line 58: [Equidispersion](https://www.researchgate.net/publication/321375217_Extended_Poisson_INAR1_processes_with_equidispersion_underdispersion_and_overdispersion)
  If in a Poisson distribution if the variance equals the mean of the distribution, it is reffered to as equidispersion.

[Generalized Poisson PMF](https://www.jstor.org/stable/1267389)
  A generalization of the {term}`Poisson distribution`, with two parameters X1, and X2, is obtained as a limiting form of the {term}`generalized negative binomial distribution`. The variance of the distribution is greater than, equal to or smaller than the mean according as X2 is positive, zero or negative. For formula and more detail, visit the link in the title.

[Bayes' theorem](https://en.wikipedia.org/wiki/Bayes%27_theorem)
  Describes the probability of an event, based on prior knowledge of conditions that might be related to the event. For example, if the risk of developing health problems is known to increase with age, Bayes' theorem allows the risk to an individual of a known age to be assessed more accurately (by conditioning it on their age) than simply assuming that the individual is typical of the population as a whole.
  > Formula:
  > $$
  > P(A|B) = \frac{P(B|A) P(A}{Pr(B)}
  > $$
  > Where A and B are events and P(B) != 0
  

[Markov Chain](https://en.wikipedia.org/wiki/Markov_chain)
  A Markov chain or Markov process is a stochastic model describing a sequence of possible events in which the probability of each event depends only on the state attained in the previous event.

[Markov Chain Monte Carlo](https://en.wikipedia.org/wiki/Markov_chain_Monte_Carlo)
[MCMC]
  Markov chain Monte Carlo (MCMC) methods comprise a class of algorithms for sampling from a probability distribution. By constructing a {term}`Markov Chain` that has the desired distribution as its equilibrium distribution, one can obtain a sample of the desired distribution by recording states from the chain.  Various algorithms exist for constructing chains, including the Metropolis–Hastings algorithm.

itsguneetsingh (commented):
I couldn't see the suggest changes button so I've made a review for adding some more definitions.  I think that these definitions are fine so if someone could have a look at this and @olgadk7 if you could then add these to the file

PR #4996, opened 2021-09-15
Fix Flat/HalfFlat initial value dimensionality

michaelosthege:
Just adding the MRE as unit tests.

Anybody please feel free to pick this up.


michaelosthege:
Superseded by #4983 which is going to avoid the issue alltogether.

PR #5004, opened 2021-09-20
Improve tuning by skipping the first samples + add new experimental tuning method

aseyboldt:
Tuning the mass matrix starts right away in the current implementation, but in most models during the first couple of samples we are only moving to the typical set, so we do not get information about the posterior variance at all. In the worst case we learn a mass matrix that doesn't match the posterior at all, so that sampling the the first adaptation window will be very slow (you can see this a slowdown of sampling after step 100). Usually, we will recover from this, but it seems to be better to just skip those samples during adaptation in the first place.

In an example model by @ricardoV94 we can see this behavior clearly when we look at the distance of the currently used mass matrix to the final mass matrix:

image

This PR also contains an experimental tuning implementation using gradients and samples that can be enabled by `init="jitter+adapt_diag_grad"`. During tests on a few models this seems to be more stable than the only sample based tuning system we use right now, but there are also a few cases where it performs worse. For posteriors that are normal it should converge to the same mass matrix as our current implementation (and much faster), but for non-normal posteriors the result can differ. Unfortunately I don't know of any other way to tell which is better other than trying it on a large number of models.

An example notebook can be found here:
https://gist.github.com/aseyboldt/7897fbddacacaa0c86efc917afe9ce3f

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -23,6 +23,8 @@
 - The `OrderedMultinomial` distribution has been added for use on ordinal data which are _aggregated_ by trial, like multinomial observations, whereas `OrderedLogistic` only accepts ordinal data in a _disaggregated_ format, like categorical
   observations (see [#4773](https://github.com/pymc-devs/pymc3/pull/4773)).
 - The `Polya-Gamma` distribution has been added (see [#4531](https://github.com/pymc-devs/pymc3/pull/4531)). To make use of this distribution, the [`polyagamma>=1.3.1`](https://pypi.org/project/polyagamma/) library must be installed and available in the user's environment.
+- A small change to the default mass matrix tuning method jitter+adapt_diag improves performance early on during tuning for some models. [#5004](https://github.com/pymc-devs/pymc3/pull/5004)
+- New experimental mass matrix tuning method jitter+adapt_diag_grad. [#5004](https://github.com/pymc-devs/pymc3/pull/5004)
 - ...
 
 ### Maintenance
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -23,7 +23,7 @@
 - The `OrderedMultinomial` distribution has been added for use on ordinal data which are _aggregated_ by trial, like multinomial observations, whereas `OrderedLogistic` only accepts ordinal data in a _disaggregated_ format, like categorical
   observations (see [#4773](https://github.com/pymc-devs/pymc3/pull/4773)).
 - The `Polya-Gamma` distribution has been added (see [#4531](https://github.com/pymc-devs/pymc3/pull/4531)). To make use of this distribution, the [`polyagamma>=1.3.1`](https://pypi.org/project/polyagamma/) library must be installed and available in the user's environment.
-- A small change to the default mass matrix tuning method jitter+adapt_diag improves performance early on during tuning for some models. [#5004](https://github.com/pymc-devs/pymc3/pull/5004)
+- A small change to the mass matrix tuning methods jitter+adapt_diag (the default) and adapt_diag improves performance early on during tuning for some models. [#5004](https://github.com/pymc-devs/pymc3/pull/5004)
 - New experimental mass matrix tuning method jitter+adapt_diag_grad. [#5004](https://github.com/pymc-devs/pymc3/pull/5004)
 - ...
 
--- a/pymc3/sampling.py
+++ b/pymc3/sampling.py
@@ -516,6 +516,7 @@ def sample(
                 random_seed=random_seed,
                 progressbar=progressbar,
                 jitter_max_retries=jitter_max_retries,
+                tune=tune,
                 **kwargs,
             )
             if start is None:
@@ -2078,6 +2079,7 @@ def init_nuts(
     random_seed=None,
     progressbar=True,
     jitter_max_retries=10,
+    tune=None,
     **kwargs,
 ):
     """Set up the mass matrix initialization for NUTS.
@@ -2174,6 +2176,24 @@ def init_nuts(
         var = np.ones_like(mean)
         n = len(var)
         potential = quadpotential.QuadPotentialDiagAdapt(n, mean, var, 10)
+    elif init == "jitter+adapt_diag_grad":
+        start = _init_jitter(model, model.initial_point, chains, jitter_max_retries)
+        mean = np.mean([DictToArrayBijection.map(vals).data for vals in start], axis=0)
+        var = np.ones_like(mean)
+        n = len(var)
+
+        if tune is not None and tune > 200:
+            stop_adaptation = tune - 50
+        else:
+            stop_adaptation = None
+
+        potential = quadpotential.QuadPotentialDiagAdaptExp(
+            n,
+            mean,
+            alpha=0.02,
+            use_grads=True,
+            stop_adaptation=stop_adaptation,
+        )
     elif init == "advi+adapt_diag_grad":
         approx: pm.MeanField = pm.fit(
             random_seed=random_seed,
--- a/pymc3/sampling.py
+++ b/pymc3/sampling.py
@@ -287,25 +287,7 @@ def sample(
         by default. See ``discard_tuned_samples``.
     init : str
         Initialization method to use for auto-assigned NUTS samplers.
-
-        * auto: Choose a default initialization method automatically.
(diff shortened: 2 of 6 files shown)


pymc3/step_methods/hmc/quadpotential.py
```diff
@@ -342,6 +360,8 @@ def __init__(
 
     def add_sample(self, x, weight):
         x = np.asarray(x)
+        if weight != 1:
+            raise ValueError("weight is unused and broken")
```
twiecki:
```suggestion
            raise ValueError("Setting weight != 1 is not supported.")
```

Or maybe we should just remove it all-together.

pymc3/tests/test_sampling.py
```diff
@@ -87,7 +87,7 @@ def test_sample(self):
 
     def test_sample_init(self):
         with self.model:
-            for init in ("advi", "advi_map", "map"):
+            for init in ("advi", "advi_map", "map", "jitter+adapt_diag_grad"):
```
twiecki: Should we add all the others here too?

twiecki:
Needs a line in the release-notes.

pymc3/tests/test_sampling.py
aseyboldt: done

pymc3/step_methods/hmc/quadpotential.py
aseyboldt: done

rlouf:
I noticed something similar when debugging `blackjax`'s warmup. This is great ! It should also be useful for `aehmc`.

twiecki:
```

181
=================================== FAILURES ===================================
182
_________________________ TestSample.test_sample_init __________________________
183

184
self = <pymc3.tests.test_sampling.TestSample object at 0x7fe49ec976d0>
185

186
    def test_sample_init(self):
187
        with self.model:
188
            for init in (
189
                "advi",
190
                "advi_map",
191
                "map",
192
                "adapt_diag",
193
                "jitter+adapt_diag",
194
                "jitter+adapt_diag_grad",
195
                "advi+adapt_diag_grad",
196
                "advi+adapt_diag",
197
                "adapt_full",
198
                "jitter+adapt_full",
199
            ):
200
                pm.sample(
201
                    init=init,
202
                    tune=120,
203
                    n_init=1000,
204
                    draws=50,
205
>                   random_seed=self.random_seed,
206
                )
207

208
pymc3/tests/test_sampling.py:107: 
209
210
pymc3/sampling.py:617: in sample
211
    trace = _sample_many(**sample_args)
212
pymc3/sampling.py:752: in _sample_many
213
    **kwargs,
214
pymc3/sampling.py:889: in _sample
215
    for it, (strace, diverging) in enumerate(sampling):
216
/usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/fastprogress/fastprogress.py:47: in __iter__
217
    raise e
218
/usr/share/miniconda/envs/pymc3-dev-py37/lib/python3.7/site-packages/fastprogress/fastprogress.py:41: in __iter__
219
    for i,o in enumerate(self.gen):
220
pymc3/sampling.py:1043: in _iter_sample
221
    point, stats = step.step(point)
222
pymc3/step_methods/arraystep.py:286: in step
223
    return super().step(point)
224
pymc3/step_methods/arraystep.py:208: in step
225
    step_res = self.astep(q)
226
pymc3/step_methods/hmc/base_hmc.py:189: in astep
227
    self.potential.update(hmc_step.end.q, hmc_step.end.q_grad, self.tune)
228
229

230
self = <pymc3.step_methods.hmc.quadpotential.QuadPotentialDiagAdaptExp object at 0x7fe49ed46e50>
231
sample = array([-0.5442348 , -0.46783447], dtype=float32)
232
grad = array([-2.0224946, -2.121815 ], dtype=float32), tune = True
233

234
    def update(self, sample, grad, tune):
235
>       if tune and self._n_samples < self._stop_adaptation:
236
E       TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

RELEASE-NOTES.md
```diff
@@ -23,6 +23,8 @@
 - The `OrderedMultinomial` distribution has been added for use on ordinal data which are _aggregated_ by trial, like multinomial observations, whereas `OrderedLogistic` only accepts ordinal data in a _disaggregated_ format, like categorical
   observations (see [#4773](https://github.com/pymc-devs/pymc3/pull/4773)).
 - The `Polya-Gamma` distribution has been added (see [#4531](https://github.com/pymc-devs/pymc3/pull/4531)). To make use of this distribution, the [`polyagamma>=1.3.1`](https://pypi.org/project/polyagamma/) library must be installed and available in the user's environment.
+- A small change to the default mass matrix tuning method jitter+adapt_diag improves performance early on during tuning for some models. [#5004](https://github.com/pymc-devs/pymc3/pull/5004)
```
```suggestion
- A small change to the default mass matrix tuning methods using adapt_diag improves performance early on during tuning for some models. [#5004](https://github.com/pymc-devs/pymc3/pull/5004)
```

pymc3/step_methods/hmc/quadpotential.py
```diff
@@ -222,12 +234,18 @@ def _update_from_weightvar(self, weightvar):
 
     def update(self, sample, grad, tune):
         """Inform the potential about a new sample during tuning."""
+        if self._store_mass_matrix_trace:
+            self._mass_trace.append(self._stds.copy())
```
ricardoV94, line 264: Perhaps remove this altogether

pymc3/step_methods/hmc/quadpotential.py
```diff
@@ -154,6 +154,10 @@ def __init__(
         adaptation_window=101,
         adaptation_window_multiplier=1,
         dtype=None,
+        discard_window=50,
```
ricardoV94, line 157: Docstrings for the new variables at least?

pymc3/sampling.py
```diff
@@ -2174,6 +2176,24 @@ def init_nuts(
         var = np.ones_like(mean)
         n = len(var)
         potential = quadpotential.QuadPotentialDiagAdapt(n, mean, var, 10)
+    elif init == "jitter+adapt_diag_grad":
+        start = _init_jitter(model, model.initial_point, chains, jitter_max_retries)
+        mean = np.mean([DictToArrayBijection.map(vals).data for vals in start], axis=0)
+        var = np.ones_like(mean)
+        n = len(var)
+
+        if tune is not None and tune > 200:
```
ricardoV94: use `tune > 250`

pymc3/sampling.py
```diff
@@ -2174,6 +2179,24 @@ def init_nuts(
...
```
ricardoV94:
```suggestion
        if tune is not None and tune > 250:
```

pymc3/step_methods/hmc/quadpotential.py
```diff
@@ -342,6 +360,8 @@ def __init__(
 
     def add_sample(self, x, weight):
         x = np.asarray(x)
+        if weight != 1:
+            raise ValueError("Setting weight != 1 is not supported.")
```
ricardoV94:
```suggestion
    def add_sample(self, x, weight=None):
        if weight is not None:
            warning.warn(
                "Setting weight is no longer supported and and will raise an error in the future.",
                DeprecationWarning,
            )
        x = np.asarray(x)
```

pymc3/tests/test_sampling.py
```diff
@@ -87,10 +87,21 @@ def test_sample(self):
...
         with self.model:
-            for init in ("advi", "advi_map", "map"):
+            for init in (
+                "advi",
+                "advi_map",
+                "map",
+                "adapt_diag",
+                "jitter+adapt_diag",
+                "jitter+adapt_diag_grad",
+                "advi+adapt_diag_grad",
```
ricardoV94: Let's remove this one

ricardoV94 (changes requested):
Left a couple of comments

pymc3/step_methods/hmc/quadpotential.py
aseyboldt: I think a hard break is fine here. This really was internal, unused and wrong

ricardoV94: Then I would suggest removing the weight argument altogether

ricardoV94 (approved):

ricardoV94:
Thanks @aseyboldt

PR #5008, opened 2021-09-21
deprecated_args decorator for sphinx formatting

mjhajharia:
this PR is for doing a draft version of adding auto-generate docs for deprecated arguments.


mjhajharia:
here's an example
image

mjhajharia:
[ Things to note: from Oriol ] -> not sure that would work. There could be colons within the parameter description so option 1 would need extra work regarding indentation or something like this (example https://arviz-devs.github.io/arviz/api/generated/arviz.loo.html scale parameter). It also has to accept both name: type or name : type (which is actually the numpydoc compliant formatting) because we have both in the codebase. option two should work but needs a couple caveats: section names should be preceded by an empty line and the number of underlines should be the same as the characters of the section, so the ------- should take something like between 4 or 15 characters, not completely sure they need to be dashes but I do think we always use dashes. The main problem is option 1 which might become too complicated to handle by ourselves with regex or other simple things

pymc/deprecator.py
```diff
@@ -0,0 +1,97 @@
...
+
+def deprecator(reason=None, version=None,action='deprecate', deprecated_args=None, docs=True):
+    """
+    PyMC deprecation helper. This decorator emits deprecated warnings.
+    """    
+    if deprecated_args is not None:
+        deprecated_args = set(deprecated_args.split())
+    cause = reason
+    if cause is not None and version is not None:
+        reason = ", since version " + version + " " + "("+cause+")"
```
canyon289: Dont use + to concat strings. Use f strings or .format. This + version is way harder to read and maintain

pymc/deprecator.py
```diff
@@ -0,0 +1,97 @@
...
+        reason = "("+ str(cause) + ")"
+
+    if cause is None and version is not None :
+        reason = ", since version " + version
+    #this function is an edited version of the source code taken from the library Deprecated
+    #https://github.com/tantale/deprecated (MIT License)
+    def sphinxformatter(version, reason):
+        fmtsphinx = ".. deprecated:: {version}" if version else ".. deprecated::"
+        sphinxtext = [fmtsphinx.format(directive = "deprecated", version=version)]
+        width = 2 ** 16
```
canyon289: Comment magic numbers explaining why they are the way they are

pymc/deprecator.py
```diff
@@ -0,0 +1,97 @@
...
+                        initial_indent="   ",
+                        subsequent_indent="   ",
+                    ).splitlines()
+                )
+            else:
+                sphinxtext.append("")
+        return sphinxtext
+
+
+    def decorator(func):
```
canyon289: Add docstring please

pymc/deprecator.py
```diff
@@ -0,0 +1,97 @@
...
+        if deprecated_args is None:
+            if inspect.isclass(func):
+                fmt = "Class {name} is deprecated{reason}."
+            else:
+                fmt = "Function or method {name} is deprecated{reason}."
+        else:
+            fmt = "Parameter(s) {name} deprecated{reason}"
+            
+        if docs==True:
```
canyon289: `if docs is True`

Im not so picky about this but some style guides are very picky. Ill leave it up to you if you want to change it

https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html

pymc/deprecator.py
```diff
@@ -0,0 +1,97 @@
...
+            docstring = textwrap.dedent(func.__doc__ or "")
+            if docstring:
+                docstring = re.sub(r"\n+$", "", docstring, flags=re.DOTALL) + "\n\n"
+            else:
+                docstring = "\n"
+
+            sphinxtext = sphinxformatter(version, cause)
+            if deprecated_args is None:
+                docstring += "".join("{}\n".format(line) for line in sphinxtext)
```
canyon289: Try to stick to one string formatting method if you can. Unfortunately the whole python "one obvious way" thing is a bit faulty here since python provides many ways to do to string concat, but within a file try and stick to one pattern.

pymc/deprecator.py
```diff
@@ -0,0 +1,97 @@
...
+            else:
+                for deprecated_arg in deprecated_args:
+                    doc=docstring.split(f'\n{deprecated_arg}:')[1]
+                    nextitem = re.search("\\n{1}\\w+:{1}", doc)
```
canyon289: I would provide docstrings here. Trying to guess what regex does from the code is hard

pymc/deprecator.py
mjhajharia: so when a para is given as reason. the existing line breaks will be used in the texts. but in case a string and not a paragraph is given then we'd need some defined line width and this renders very differently depending on the screen size because sphinx doesnt hard code the max width anymore afaik. these are the numbers the deprecated library used and its being used so i think this is safe to go?

mjhajharia: thanks for this comment. i had taken this from deprecated library (i saw that they had some descriptions and i figured since it was being used widely this might work) turns out. sphinx doesnt even have a fixed max width anymore. the default width for textwrap is 70. and i tested giving no width with lines and paragraphs both. works totally fine. so we dont need the unnecessary specification. im glad you asked this and i looked more into it

mjhajharia:
closing in favour of https://github.com/tantale/deprecated/pull/51

PR #5010, opened 2021-09-22
Monkeypatch instructions for migrating away from the old rv.logp API

michaelosthege:
In `v3` the logp was a method on the RV value variables returned from distributions, but it is no longer available with `v4`.

This PR monkeypatches a method that raises an `AttributeError` with migration instructions instead of the non-informative default one.

### DIFF
--- a/pymc3/distributions/distribution.py
+++ b/pymc3/distributions/distribution.py
@@ -123,6 +123,13 @@ class DistributionMeta(ABCMeta):
         return new_cls
 
 
+def _make_nice_attr_error(oldcode: str, newcode: str):
+    def fn(*args, **kwargs):
+        raise AttributeError(f"The `{oldcode}` method was removed. Instead use `{newcode}`.`")
+
+    return fn
+
+
 class Distribution(metaclass=DistributionMeta):
     """Statistical distribution"""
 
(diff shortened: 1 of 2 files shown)


pymc3/distributions/distribution.py
```diff
@@ -244,6 +248,7 @@ def __new__(
             functools.partial(str_for_dist, formatting="latex"), rv_out
         )
 
+        rv_out.logp = _nice_logp_attribute_error
```
ricardoV94: I don't think you need this one?

pymc3/tests/test_distributions.py
```diff
@@ -3262,3 +3262,12 @@ def test_distinct_rvs():
         pp_samples_2 = pm.sample_prior_predictive(samples=2)
 
     assert np.array_equal(pp_samples["y"], pp_samples_2["y"])
+
+
+def test_logp_gives_migration_instructions():
+    with pm.Model(coords={"year": [2019, 2021, 2022]}):
```
ricardoV94, line 3284: Why coords in this test?

pymc3/distributions/distribution.py
michaelosthege: We need it because of the resizing that can happen a few lines above.
That's why in the test case I'm adding a dim that triggers a resize _after_ the `.dist()` line.

ricardoV94: Maybe mention that in the test?

ricardoV94:
What about logcdf and random?

ricardoV94 (approved):

ricardoV94:
Thanks @michaelosthege

PR #5011, opened 2021-09-22
BART: clamp first particle to old full tree

aloctavodia:
We no longer keep track of the full history of a particle/tree, we just keep the old tree (so there is still a chance to do not accept a move/new tree). The effect on the function approximation is negligible, but instead the sampler is faster and also the code becomes simpler.

### DIFF
--- a/pymc/step_methods/pgbart.py
+++ b/pymc/step_methods/pgbart.py
@@ -108,9 +108,9 @@ class PGBART(ArrayStepShared):
 
         if self.chunk == "auto":
             self.chunk = max(1, int(self.m * 0.1))
-        self.num_particles = num_particles
         self.log_num_particles = np.log(num_particles)
         self.indices = list(range(1, num_particles))
+        self.len_indices = len(self.indices)
         self.max_stages = max_stages
 
         shared = make_shared_replacements(initial_values, vars, model)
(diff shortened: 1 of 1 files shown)


pymc/step_methods/pgbart.py
```diff
@@ -274,6 +259,19 @@ def init_particles(self, tree_id):
 
         return np.array(particles)
 
+    def update_weight(self, particle):
+        """
+        Update the weight of a particle
+
+        Since the prior is used as the proposal,the weights are updated additively as the ratio of
+        the new and old log_likelihoods.
+        """
```
michaelosthege: Can you add a type hint to the signature?
```suggestion
    def update_weight(self, particle):
        """
        Update the weight of a particle

        Since the prior is used as the proposal, the weights are updated additively as the ratio of
        the new and old log-likelihoods.
        """
```

pymc/step_methods/pgbart.py
```diff
@@ -166,22 +164,14 @@ def astep(self, q: RaveledVars) -> Tuple[RaveledVars, List[Dict[str, Any]]]:
...
-                # Update weights. Since the prior is used as the proposal,the weights
-                # are updated additively as the ratio of the new and old log_likelihoods
-                for clp, p in zip(compute_logp, particles):
-                    if clp:  # Compute the likelihood when p has changed from the previous iteration
-                        new_likelihood = self.likelihood_logp(
-                            sum_trees_output_noi + p.tree.predict_output()
-                        )
-                        p.log_weight += new_likelihood - p.old_likelihood_logp
-                        p.old_likelihood_logp = new_likelihood
+                    if clp:  # update weights only if p has changed from the previous iteration
```
michaelosthege: Comments on their own line typically give prettier layouts and git diffs.
Maybe a more informative variable name than `clp` could make the comment redundant?
```suggestion
                    if clp:
                        # update weights only if p has changed from the previous iteration
```

michaelosthege (approved):
I have just a few nitpicks related to code style.
Since I'm not familiar with the sampler, I can't tell if you're breaking it, so take my approval with a grain of salt ;)

pymc/step_methods/pgbart.py
```diff
@@ -32,14 +32,67 @@
...
+
 class PGBART(ArrayStepShared):
     """
     Particle Gibss BART sampling step
 
     Parameters
     ----------
     vars: list
-        List of value variables for sampler
+        List of variables for sampler
```
ricardoV94: From skimming the `__init__` method it seems the step method expects value variables, no?
In particular these lines:

```python
        value_bart = inputvars(vars)[0]
        self.bart = model.values_to_rvs[value_bart].owner.op
```

ricardoV94: That's why I changed the docstrings in https://github.com/pymc-devs/pymc3/commit/46209f6a779ab38effb7f70accd7297dee7b2f13

aloctavodia: You are right, I introduced this change inadvertently

Issue #4993, opened 2021-09-15
HalfFlat initval dimensionality is broken


michaelosthege:
## Description of your problem
Initial values for non-scalar HalfFlat (and probably Flat too) have a shape problem.

**Please provide a minimal, self-contained, and reproducible example.**
```python
with pm.Model(coords={
    "city": np.array(["Paris", "Berlin", "Venice"], dtype=str),
}) as pmodel:
    pm.HalfFlat("hf_scalar")
    pm.HalfFlat("hf_shape", shape=(2,))
    pm.HalfFlat("hf_dims", dims=("city",))
```

**Please provide the full traceback.**

<details><summary>Complete error traceback</summary>

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-21fc589ae251> in <module>
      3 }) as pmodel:
      4     pm.HalfFlat("hf_scalar")
----> 5     pm.HalfFlat("hf_shape", shape=(2,))
      6     pm.HalfFlat("hf_dims", dims=("city",))

c:\users\osthege\repos\pymc3-v4\pymc3\distributions\distribution.py in __new__(cls, name, rng, dims, initval, observed, total_size, transform, *args, **kwargs)
    214         # Create the RV without dims information, because that's not something tracked at the Aesara level.
    215         # If necessary we'll later replicate to a different size implied by already known dims.
--> 216         rv_out = cls.dist(*args, rng=rng, **kwargs)
    217         ndim_actual = rv_out.ndim
    218         resize_shape = None

c:\users\osthege\repos\pymc3-v4\pymc3\distributions\continuous.py in dist(cls, size, **kwargs)
    429     def dist(cls, *, size=None, **kwargs):
    430         res = super().dist([], size=size, **kwargs)
--> 431         res.tag.test_value = np.full(size, floatX(1.0))
    432         return res
    433 

~\AppData\Local\Continuum\miniconda3\envs\pm3v4\lib\site-packages\aesara\graph\utils.py in __setattr__(self, attr, obj)
    265 
    266         if getattr(self, "attr", None) == attr:
--> 267             obj = self.attr_filter(obj)
    268 
    269         return object.__setattr__(self, attr, obj)

~\AppData\Local\Continuum\miniconda3\envs\pm3v4\lib\site-packages\aesara\tensor\type.py in filter(self, data, strict, allow_downcast)
    201 
    202         if self.ndim != data.ndim:
--> 203             raise TypeError(
[...]

PR #5025, opened 2021-09-24
More moments

michaelosthege:
This adds get_moment implementations and tests them more thoroughly.

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -336,6 +336,11 @@ class Uniform(BoundedContinuous):
             ),
         )
 
+    def get_moment(value, size, lower, upper):
+        lower = at.full(size, lower, dtype=aesara.config.floatX)
+        upper = at.full(size, upper, dtype=aesara.config.floatX)
+        return (lower + upper) / 2
+
 
 class FlatRV(RandomVariable):
     name = "flat"
@@ -366,7 +371,7 @@ class Flat(Continuous):
         res.tag.test_value = np.full(size, floatX(0.0))
         return res
 
-    def get_moment(rv, size, *rv_inputs) -> np.ndarray:
+    def get_moment(rv, size, *rv_inputs):
         return at.zeros(size, dtype=aesara.config.floatX)
 
     def logp(value):
@@ -431,7 +436,7 @@ class HalfFlat(PositiveContinuous):
         res.tag.test_value = np.full(size, floatX(1.0))
         return res
 
-    def get_moment(value_var, size, *rv_inputs) -> np.ndarray:
+    def get_moment(value_var, size, *rv_inputs):
         return at.ones(size, dtype=aesara.config.floatX)
 
     def logp(value):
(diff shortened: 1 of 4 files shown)


pymc/tests/test_initvals.py
```diff
@@ -103,3 +109,33 @@ def test_basic(self):
         assert np.all(get_moment(rv).eval() == np.zeros((2, 4)))
         rv = pm.HalfFlat.dist(size=(2, 4))
         assert np.all(get_moment(rv).eval() == np.ones((2, 4)))
+
+    @pytest.mark.xfail(reason="Test values are still used for initvals.")
```
twiecki, line 113: Isn't that what this PR is about?

michaelosthege: No this PR is just to get some commits out of the upcoming initval framework PR, which is still blocked by `aeppl` integration.

These tests originally come from #4996 and are  regression tests for #4993 which needs the new initval framework to get fixed.

pymc/distributions/continuous.py
```diff
@@ -336,6 +336,11 @@ def logcdf(value, lower, upper):
             ),
         )
 
+    def get_moment(value, size, lower, upper):
+        lower = at.full(size, lower, dtype=aesara.config.floatX)
```
lucianopaz, line 340: Shouldn’t the dtype be taken from the op instead of aesara’s config?

pymc/distributions/discrete.py
```diff
@@ -394,6 +394,10 @@ def logcdf(value, p):
             p <= 1,
         )
 
+    def get_moment(value, size, p):
+        p = at.full(size, p)
+        return at.switch(p < 0.5, np.int64(0), np.int64(1))
```
lucianopaz: Maybe instead of forcing int64, we could rely on the op’s dtype?

pymc/distributions/continuous.py
michaelosthege:
```suggestion
        lower = at.full(size, lower, dtype=value.owner.op.inputs[2])
```

pymc/distributions/discrete.py
michaelosthege:
```suggestion
        return at.switch(p < 0.5, at.zeros_like(value), at.ones_like(value))
```

pymc/distributions/continuous.py
```diff
@@ -336,6 +336,11 @@ def logcdf(value, lower, upper):
             ),
         )
 
+    def get_moment(value, size, lower, upper):
+        lower = at.full(size, lower, dtype=value.owner.op.inputs[2])
+        upper = at.full(size, upper, dtype=aesara.config.floatX)
```
twiecki, line 341: why other dtype?

pymc/distributions/continuous.py
```diff
@@ -336,6 +336,11 @@ def logcdf(value, lower, upper):
             ),
         )
 
+    def get_moment(value, size, lower, upper):
+        lower = at.full(size, lower, dtype=value.owner.op.inputs[2])
```
twiecki:
```suggestion
        lower = at.full(size, lower, dtype=aesara.config.floatX)
```

twiecki (approved):

pymc/distributions/continuous.py
lucianopaz: Because if the user passes a dtype to the distribution, using floatX might ignore it.

lucianopaz: And we don’t want that. If the user doesn’t specify anything, the ops dtype will already be floatX

PR #5026, opened 2021-09-24
Refactor DensityDist into v4

### LINKED ISSUE #4831: Refactor or deprecate `DensityDist`

lucianopaz:
Closes #4831

This PR ports `DensityDist` into v4. It creates a `DensityDistRV` class and `DensityDist` class in a similar way as is done by the `Simulator` class. When a `DensityDist` is created, it registers the `logp`, `logcdf` and `get_moment` functions for the associated Op type. The `DensityDistRV` is used to overload the `rng_fn` from the base class to be able to run the user supplied `random` callable with the appropriate signature.

+ [X] what are the (breaking) changes that this PR makes?

The signature of `DensityDist` is completely different than in v3:

- It now takes as positional arguments, the distribution parameters. In v3, these were passed in through a dictionary that was supplied as `observed`. The associated `MultiObservedRV` is no longer used or needed and can be removed in a following PR. 
- `logp` is now an optional keyword only argument instead of the first positional argument. If users try to pass it as the first positional argument, they will get an exception that tells them that they are using the old v3 API and should change.
- The signature of `logp` is now different, it takes in the distribution parameters as positional arguments. Something similar happens for `random`.
- All the distribution methods: `logp`, `random`, `logcdf` and `get_moment` are optional. If they are not passed during construction, the RV will be built, but a `NotImplementedError` will be raised when you try to do something like `pm.sample`, `pm.sample_prior_predictive` or `get_moment`.

+ [X] important background, or details about the implementation
+ [X] are the changes—especially new features—covered by tests and docstrings?

The docstrings were updated and new tests have been added for the different optional methods (except `logcdf`) which behaves in the same way as `logp`. If necessary, it's possible to add tests for `logcdf` too.

+ [X] [linting/style checks have been run](https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -10,6 +10,9 @@
 - `pm.sample` now returns results as `InferenceData` instead of `MultiTrace` by default (see [#4744](https://github.com/pymc-devs/pymc/pull/4744)).
 - `pm.sample_prior_predictive` no longer returns transformed variable values by default. Pass them by name in `var_names` if you want to obtain these draws (see [4769](https://github.com/pymc-devs/pymc/pull/4769)).
 - ⚠ `pm.Bound` interface no longer accepts a callable class as argument, instead it requires an instantiated distribution (created via the `.dist()` API) to be passed as an argument. In addition, Bound no longer returns a class instance but works as a normal PyMC distribution. Finally, it is no longer possible to do predictive random sampling from Bounded variables. Please, consult the new documentation for details on how to use Bounded variables (see [4815](https://github.com/pymc-devs/pymc/pull/4815)).
+- `pm.DensityDist` no longer accepts the `logp` as its first position argument. It is now an optional keyword argument. If you pass a callable as the first positional argument, a `TypeError` will be raised (see [5026](https://github.com/pymc-devs/pymc3/pull/5026)).
+- `pm.DensityDist` now accepts distribution parameters as positional arguments. Passing them as a dictionary in the `observed` keyword argument is no longer supported and will raise an error (see [5026](https://github.com/pymc-devs/pymc3/pull/5026)).
+- The signature of the `logp` and `random` functions that can be passed into a `pm.DensityDist` has been changed (see [5026](https://github.com/pymc-devs/pymc3/pull/5026)).
 - ...
 
 ### New Features
@@ -25,6 +28,8 @@
 - The `Polya-Gamma` distribution has been added (see [#4531](https://github.com/pymc-devs/pymc/pull/4531)). To make use of this distribution, the [`polyagamma>=1.3.1`](https://pypi.org/project/polyagamma/) library must be installed and available in the user's environment.
 - A small change to the mass matrix tuning methods jitter+adapt_diag (the default) and adapt_diag improves performance early on during tuning for some models. [#5004](https://github.com/pymc-devs/pymc/pull/5004)
 - New experimental mass matrix tuning method jitter+adapt_diag_grad. [#5004](https://github.com/pymc-devs/pymc/pull/5004)
+- `pm.DensityDist` can now accept an optional `logcdf` keyword argument to pass in a function to compute the cummulative density function of the distribution (see [5026](https://github.com/pymc-devs/pymc3/pull/5026)).
+- `pm.DensityDist` can now accept an optional `get_moment` keyword argument to pass in a function to compute the moment of the distribution (see [5026](https://github.com/pymc-devs/pymc3/pull/5026)).
 - ...
 
 ### Maintenance
--- a/docs/source/Probability_Distributions.rst
+++ b/docs/source/Probability_Distributions.rst
@@ -58,16 +58,16 @@ An exponential survival function, where :math:`c=0` denotes failure (or non-surv
     f(c, t) = \left\{ \begin{array}{l} \exp(-\lambda t), \text{if c=1} \\
                \lambda \exp(-\lambda t), \text{if c=0}  \end{array} \right.
 
-Such a function can be implemented as a PyMC distribution by writing a function that specifies the log-probability, then passing that function as an argument to the ``DensityDist`` function, which creates an instance of a PyMC distribution with the custom function as its log-probability.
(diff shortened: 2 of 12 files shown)


RELEASE-NOTES.md
```diff
@@ -10,6 +10,10 @@
...
+- `pm.DensityDist` no longer accepts the `logp` as its first position argument. It is now an optional keyword argument. If you pass a callable as the first positional argument, a `TypeError` will be raised (see [5026](https://github.com/pymc-devs/pymc3/pull/5026)).
+- `pm.DensityDist` now accepts distribution parameters as positional arguments. Passing them as a dictionary in the `observed` keyword argument is no longer supported and will raise an error (see [5026](https://github.com/pymc-devs/pymc3/pull/5026)).
+- The signature of the `logp` and `random` functions that can be passed into a `pm.DensityDist` has been changed (see [5026](https://github.com/pymc-devs/pymc3/pull/5026)).
+>>>>>>> Refactor DensityDist into v4
```
```suggestion
```

docs/source/Probability_Distributions.rst
```diff
@@ -58,16 +58,16 @@ An exponential survival function, where :math:`c=0` denotes failure (or non-surv
     f(c, t) = \left\{ \begin{array}{l} \exp(-\lambda t), \text{if c=1} \\
                \lambda \exp(-\lambda t), \text{if c=0}  \end{array} \right.
 
-Such a function can be implemented as a PyMC distribution by writing a function that specifies the log-probability, then passing that function as an argument to the ``DensityDist`` function, which creates an instance of a PyMC distribution with the custom function as its log-probability.
+Such a function can be implemented as a PyMC distribution by writing a function that specifies the log-probability, then passing that function as a keyword argument to the ``DensityDist`` function, which creates an instance of a PyMC3 distribution with the custom function as its log-probability.
```
michaelosthege: PyMC3 - > PyMC

docs/source/Probability_Distributions.rst
```diff
@@ -58,16 +58,16 @@ An exponential survival function, where :math:`c=0` denotes failure (or non-surv
...
 
 For the exponential survival function, this is:
 
 ::
 
-    def logp(failure, value):
-        return (failure * log(λ) - λ * value).sum()
+    def logp(value, t, λ):
```
michaelosthege: Earlier this year I put a sigma character in a signature, because I thought it was neat.

Two months later I took it back out, because it was so annoying to type Greek letters outside of Jupyter notebooks.

"lam" is the alternative we typically use in other places.

pymc/distributions/distribution.py
```diff
@@ -387,96 +390,241 @@ class NoDistribution(Distribution):
...
-            The dtype of the distribution.
-        initval: number or array (Optional)
-            The ``initval`` of the RV's tensor that follow the ``DensityDist``
-            distribution.
-        args, kwargs: (Optional)
-            These are passed to the parent class' ``__init__``.
+        name : str
+        dist_params : Tuple
+            A sequence of the distribution's parameter. These will be converted into
+            aesara tensors internally. These parameters could be other ``RandomVariable``
```
michaelosthege: Aesara capitalization

pymc/distributions/distribution.py
```diff
@@ -387,96 +390,241 @@ class NoDistribution(Distribution):
...
-                )
-        self.rand = random
-        self.wrap_random_with_dist_shape = wrap_random_with_dist_shape
-        self.check_shape_in_random = check_shape_in_random
+        else:
+            dtype = cls.rv_op.dtype
+        ndim_supp = cls.rv_op.ndim_supp
+        if not hasattr(output.tag, "test_value"):
+            size = to_tuple(kwargs.get("size", None)) + (1,) * ndim_supp
+            output.tag.test_value = np.zeros(size, dtype)
```
michaelosthege, line 603: The new initval framework will no longer fall back to the test_value. Furthermore this assignment doesn't have a lot of dimensionality flexibility, for example with symbolic size.

My recommendation is to take out the assignment and also the hack below.

pymc/tests/test_moment.py
```diff
@@ -0,0 +1,38 @@
...
+
+from aesara import tensor as at
+
+import pymc as pm
+
+from pymc.distributions.distribution import get_moment
+from pymc.distributions.shape_utils import to_tuple
+
+
+@pytest.mark.parametrize("size", [None, (), (2,), (3, 2)], ids=str)
```
michaelosthege: IIRC the None to () conversion happens in the `Distribution.__new__` and doesn't need to be tested here.

pymc/distributions/distribution.py
lucianopaz: Yes, but the new initial value framework has not been finished, so I’ve used the same hack that Flat and HalfFlat use [to get their initial values](https://github.com/pymc-devs/pymc3/blob/main/pymc/model.py#L1017-L1027). I don’t think that this PR should have to wait for the new initial values framework to merge it.
In any case, over the new framework is in place, we could change the default initialisation used by density dist.

docs/source/Probability_Distributions.rst
lucianopaz: Thanks for the comments @michaelosthege, I’ll edit the code as soon as possible.

michaelosthege (approved):
I like it 👍

twiecki (approved):

twiecki:
Thanks @lucianopaz!

PR #5027, opened 2021-09-26
Rename and refactor start dict in sampling

michaelosthege:
This PR is in preparation of switching to the new initval framework.

To make debugging and reviewing a little easier, I added type hints and moved the `start`-related code closer together.

## API/Behavior changes:
+ [x] A few lines related to the now unsupported use of length-zero traces as `pm.sample` inputs were removed.
+ [x] The result from `init_nuts` is now _always_ used as the initial/starting point, whereas before it was only used unless a `start` dict was manually specified. `init_nuts` itself combines the automatically determined initial point with the user-provided `initvals` such that `initvals` take priority.
+ [x] The `pm.sample(start=...)` kwarg was renamed to `initvals`, to reflect that it takes the same keys/values/signature as `model.initial_values` or the corresponding `Distribution.__new__(initval=...)` kwarg.
+ [x] `start` kwargs of lower-level sampling functions are now required to be numeric & complete. `model.update_start_vals` is no longer applied by lower-level functions.
+ [x] Moving forward only `pm.sample(initvals=...)` and `init_nuts(initvals=...)` take the fully-flexibly `initval`-style dictionary of potentially incomplete (and soon also symbolic, "prior", or "moment" valued) initval strategies.
+ [x] Checks of non-inf/nan initial points and corresponding shapes now run under all circumstances.

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -23,7 +23,7 @@ import warnings
 
 from collections import defaultdict
 from copy import copy, deepcopy
-from typing import Any, Dict, Iterable, List, Optional, Set, Union, cast
+from typing import Dict, Iterable, List, Optional, Sequence, Set, Union, cast
 
 import aesara.gradient as tg
 import cloudpickle
@@ -253,7 +253,7 @@ def sample(
     step=None,
     init="auto",
     n_init=200_000,
-    start=None,
+    start: Optional[Union[PointType, Sequence[Optional[PointType]]]] = None,
     trace: Optional[Union[BaseTrace, List[str]]] = None,
     chain_idx=0,
     chains=None,
@@ -689,7 +689,7 @@ def _sample_many(
     draws,
     chain: int,
     chains: int,
-    start: list,
+    start: Sequence[PointType],
     random_seed: list,
     step,
     callback=None,
@@ -746,7 +746,7 @@ def _sample_population(
     draws: int,
     chain: int,
     chains: int,
-    start,
+    start: Sequence[PointType],
     random_seed,
     step,
     tune,
@@ -809,7 +809,7 @@ def _sample(
     chain: int,
     progressbar: bool,
     random_seed,
-    start,
+    start: PointType,
     draws: int,
     step=None,
     trace: Optional[Union[BaseTrace, List[str]]] = None,
@@ -875,7 +875,7 @@ def _sample(
 def iter_sample(
     draws: int,
     step,
-    start: Optional[Dict[Any, Any]] = None,
+    start: PointType,
     trace=None,
     chain=0,
     tune: Optional[int] = None,
@@ -895,8 +895,7 @@ def iter_sample(
     step : function
         Step function
     start : dict
-        Starting point in parameter space (or partial point). Defaults to trace.point(-1)) if
-        there is a trace provided and model.initial_point if not (defaults to empty dict)
+        Starting point in parameter space (or partial point).
     trace : backend or list
         This should be a backend instance, or a list of variables to track.
         If None or a list of variables, the NDArray backend is used.
@@ -935,7 +934,7 @@ def iter_sample(
 def _iter_sample(
     draws,
     step,
-    start=None,
+    start: Optional[PointType],
     trace: Optional[Union[BaseTrace, List[str]]] = None,
     chain=0,
     tune=None,
@@ -951,8 +950,8 @@ def _iter_sample(
         The number of samples to draw
     step : function
         Step function
-    start : dict, optional
-        Starting point in parameter space (or partial point). Defaults to model.initial_point if not (defaults to empty dict)
+    start : dict
+        Starting point in parameter space (or partial point).
     trace : backend or list
         This should be a backend instance, or a list of variables to track.
         If None or a list of variables, the NDArray backend is used.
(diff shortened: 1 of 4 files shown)


michaelosthege:
@aseyboldt the only remaining test failure is limited to float32. The test was written by you 4 years ago and I can't tell why it failed:
https://github.com/pymc-devs/pymc3/pull/5027/checks?check_run_id=3720168022#step:7:2486

I was able to establish that the initial point that failes the `_check_start_shape()` was `{'a': array([0., 0.], dtype=float32), 'c': array([0., 0.], dtype=float32)}`.

michaelosthege:
I XFAILed the test, since the error appears in a branch (start shape checking) that was not triggered before.

pymc/sampling.py
```diff
@@ -655,31 +656,41 @@ def sample(
...
+                else:
+                    # if model var has a specified shape
+                    if var_shape.size > 0:
+                        e += "\nExpected shape {} for var " "'{}', got scalar {}".format(
+                            tuple(var_shape), var.name, start[var.name]
+                        )
+        except NotImplementedError as ex:
+            if ex.args[0].startswith("Cannot sample"):
+                _log.warning(
+                    "Unable to check start shape of {var} because the RV does not implement random sampling."
```
ricardoV94: Unrelated to this PR, but in theory we should be able to infer shapes without sampling, even for variables that cannot be sampled from...

ricardoV94 (commented):
Needs a release note

pymc/sampling.py
michaelosthege: Totally. I opened a ticket #5031

michaelosthege:
I'm collecting all release notes in the Hackmd document. Also there will be another update on this API with the next PR.

If that's fine with you, I will update the release notes in the next PR?

pymc/sampling.py
```diff
@@ -291,11 +291,10 @@ def sample(
         users.
     n_init : int
         Number of iterations of initializer. Only works for 'ADVI' init methods.
-    start : dict, or array of dict
-        Starting point in parameter space (or partial point)
-        Defaults to ``trace.point(-1))`` if there is a trace provided and model.initial_point if not
-        (defaults to empty dict). Initialization methods for NUTS (see ``init`` keyword) can
-        overwrite the default.
+    initvals : optional, dict, array of dict
+        Dict or list of dicts with initial values to used instead of the defaults from `Model.initial_values`.
```
aloctavodia:
```suggestion
        Dict or list of dicts with initial values to use instead of the defaults from `Model.initial_values`.
```

pymc/sampling.py
```diff
@@ -417,31 +416,27 @@ def sample(
             mean     sd  hdi_3%  hdi_97%
         p  0.609  0.047   0.528    0.699
     """
+    if "start" in kwargs:
+        if initvals is not None:
+            raise ValueError("Passing both `start` and `initvals` is not supported.")
+        warnings.warn(
+            "The `start` kwarg was renamed to `initvals`. Please check the docstring.",
+            DeprecationWarning,
```
aloctavodia: This should be a FutureWarning as this warning is intended for end users.

pymc/sampling.py
aloctavodia:
```suggestion
                   f"Unable to check start shape of {var} because the RV does not implement random sampling."
```

pymc/sampling.py
```diff
@@ -2084,6 +2085,9 @@ def init_nuts(
 
     chains : int
         Number of jobs to start.
+    initvals : optional, dict or list of dicts
+        Dict or list of dicts with initial values to used instead of the defaults from `Model.initial_values`.
```
aloctavodia:
```suggestion
        Dict or list of dicts with initial values to use instead of the defaults from `Model.initial_values`.
```

pymc/sampling.py
```diff
@@ -951,8 +961,9 @@ def _iter_sample(
         The number of samples to draw
     step : function
         Step function
-    start : dict, optional
-        Starting point in parameter space (or partial point). Defaults to model.initial_point if not (defaults to empty dict)
+    start : dict
+        Starting point in parameter space (or partial point).
+        Must be complete.
```
aloctavodia: Not entirely sure, but maybe we could be more explicit that "complete" (here and in the 2 other instances were "complete" is used in this PR)?

pymc/sampling.py
```diff
@@ -978,19 +989,14 @@ def _iter_sample(
...
 
-    model.update_start_vals(start, model.initial_point)
-
     try:
         step = CompoundStep(step)
     except TypeError:
         pass
 
-    point = Point(start, model=model, filter_model_vars=True)
+    point = start
```
aloctavodia, line 999: better to simply use "start" instead of renaming the variable.

michaelosthege: I decided against that because the "point" variable is overwritten all over again while iterating the draws. So it's no longer the "start" after the first iteration and I wanted to avoid confusion because of lines like `strace.record(start, stats)`.

michaelosthege (commented):
Thanks @aloctavodia ! With the exception of one item I implemented your suggestions :)

Please let me know if we can move forward with this PR.

aloctavodia (approved):

PR #5040, opened 2021-10-04
Deprecated

mjhajharia:
this pr is made to branch out efforts in #5008 which will be a draft for figuring out the auto formatting sphinx with respect to deprecated parameters. this PR does the following:
- deprecator that deprecated functions or classes (lets you choose to emit or not emit warnings) 
- tests for deprecator
- adds deprecated in docstring for methods, functions and classes
- additionally you can deprecate arguments as well. to be able to document argument deprecation im doing #5008 

things to do:

- [x] add a docstring for what the parameters in the deprecator function do
- [ ] test existing deprecated class with this decorator
- [ ] make a table of all the deprecated functions in documentation

### DIFF
--- a/pymc/tests/test_util.py
+++ b/pymc/tests/test_util.py
@@ -20,7 +20,7 @@ from cachetools import cached
 import pymc as pm
 
 from pymc.distributions.transforms import Transform
-from pymc.util import UNSET, hash_key, hashable, locally_cachedmethod
+from pymc.util import UNSET, deprecator, hash_key, hashable, locally_cachedmethod
 
 
 class TestTransformName:
@@ -136,3 +136,44 @@ def test_unset_repr(capsys):
     help(fn)
     captured = capsys.readouterr()
     assert "a=UNSET" in captured.out
+
+
+class TestDeprecatorDecorator:
+    def test_warning_msg():
+        version = "1.2.3"
+        reason = "Good reason"
+        deprecated_args = "x"
+
+        @deprecator(version=version, reason=reason, deprecated_args="x")
+        def foo(x, y):
+            pass
+
+        with pytest.warns(DeprecationWarning, match=version) as warnversion:
+            foo()
+        with pytest.warns(DeprecationWarning, match=deprecated_args) as warnwarnargs:
+            foo()
+        with pytest.warns(DeprecationWarning, match=reason) as warnreason:
+            foo()
+
+        assert len(warnversion[0]) == 1
+        assert len(warnwarnargs[0]) == 1
+        assert len(warnreason[0]) == 1
+
+    def test_ignore_action():
+        @deprecator(reason="some reason", version="1", action="ignore")
+        def foo():
+            pass
+
+        with pytest.warns(None) as warns:
+            foo()
+        assert len(warns) == 0
+
+    def test_sphinx_docstring():
+        @deprecator(version="1.0", reason="some reason")
+        def basicfunc(x, y):
+            """
+            something
+            """
+            return x + y
+
+        assert basicfunc.__doc__ == "\nsomething\n\n.. deprecated:: 1.0\n   some reason\n"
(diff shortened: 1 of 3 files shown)


pymc/tests/test_util.py
```diff
@@ -136,3 +136,44 @@ def fn(a=UNSET):
     help(fn)
     captured = capsys.readouterr()
     assert "a=UNSET" in captured.out
+
+# tests for deprecator
```
michaelosthege: Instead of a comment, better group these tests in a `TestDeprecatorDecorator` class

pymc/util.py
```diff
@@ -345,3 +348,90 @@ def cf(self):
         return cf
 
     return cachedmethod(self_cache_fn(f.__name__), key=hash_key)(f)
+
+
+def deprecator(reason=None, version=None,action='deprecate', deprecated_args=None, docs=True):
+    """
+    PyMC deprecation helper. This decorator emits deprecated warnings.
```
michaelosthege, line 355: Please add docstrings about the parameters. If applicable you can also link the GitHub/Stackoverflow pages from where you got larger code snippets right in the docstring.

pymc/util.py
```diff
@@ -345,3 +348,90 @@ def cf(self):
...
+                docstring = re.sub(r"\n+$", "", docstring, flags=re.DOTALL) + "\n\n"
+            else:
+                docstring = "\n"
+
+            sphinxtext = sphinxformatter(version, cause)
+            if deprecated_args is None:
+                for line in sphinxtext: 
+                    docstring += f'{line}\n' 
+            else:
+            #if deprecated arguments are given, docstring is not modified
```
michaelosthege: Indentation. _Maybe_ black will correct it, but if not please indent manually

michaelosthege (changes requested):
@mjhajharia you'll definitely need to run `pre-commit run --all` here. It will make a ton of formatting changes.

This PR doesn't need to migrate a ton of existing deprecation warnings (in fact it's probably easier to cherry-pick it to `v3` if it doesn't).
But maybe you'll want to take _any_ existing deprecation warning and replace it with the decorator so you can check the docs build if the docstring stuff worked.

pymc/tests/test_util.py
```diff
@@ -136,3 +136,44 @@ def fn(a=UNSET):
     help(fn)
     captured = capsys.readouterr()
     assert "a=UNSET" in captured.out
+
+
+class TestDeprecatorDecorator:
+    def test_warning_msg():
```
michaelosthege, line 142: On this and the other test methods you need to include a `self` in the method arguments.
I guess that's also why according to codecov these lines weren't covered.

pymc/util.py
```diff
@@ -345,3 +348,108 @@ def cf(self):
...
+def deprecator(reason=None, version=None, warn=True, deprecated_args=None, docs=True):
+    """
+    PyMC deprecation helper. This decorator emits deprecated warnings.
+
+    Parameters
+    ----------
+    reason: string
+        Brief description of reason/motive behind deprecation or additional notes.
+    version: float
+        PyMC version in which the method or class was deprecated.
```
michaelosthege, line 362: IIRC one of these 3rd party libraries takes two version arguments. It is something like
* `deprecated_in : str` Library version when the deprecation was added.
* `removed_in : str` Version number when it stops working.

When the class/function/method/kwarg is then used with a library version that is >= the `removed_in`, the decorator could automatically raise an exception.
This way we can't bump the version without following through on deprecations, or at least modifying the upper bound.

However, if we do this, we need to consider the case of working on the `main` branch where we typically are already at a version > than the latest release. But we could just append a `-dev` there and remove it before a release.

michaelosthege (commented):
I didn't look at the generated docs, but maybe these comments already help

mjhajharia:
closing in favour of https://github.com/tantale/deprecated/pull/51

PR #5044, opened 2021-10-05
BART: add linear response, increase number of trees fitted per step

aloctavodia:
This adds the option to use a linear fit to compute the values of the leaf_nodes. A new argument "response" control if the leaf_node are computed using the mean value (as previously), a linear fit or a mix. I have set "mix" by default. At in preliminary test it seems to provide a better general performance (close to just using linear). This also increase the number of trees fitted per step, increasing the computational cost, but improving diagnostics. This also includes other small fixes.

### DIFF
--- a/pymc/step_methods/pgbart.py
+++ b/pymc/step_methods/pgbart.py
@@ -101,7 +101,7 @@ class PGBART(ArrayStepShared):
         Number of particles for the conditional SMC sampler. Defaults to 10
     max_stages : int
         Maximum number of iterations of the conditional SMC sampler. Defaults to 100.
-    batch = int
+    batch : int
         Number of trees fitted per step. Defaults to  "auto", which is the 10% of the `m` trees
         during tuning and 20% after tuning.
     model: PyMC Model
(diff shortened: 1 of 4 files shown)


pymc/step_methods/pgbart.py
```diff
@@ -97,11 +101,17 @@ class PGBART(ArrayStepShared):
         Number of particles for the conditional SMC sampler. Defaults to 10
     max_stages : int
         Maximum number of iterations of the conditional SMC sampler. Defaults to 100.
-    chunk = int
-        Number of trees fitted per step. Defaults to  "auto", which is the 10% of the `m` trees.
+    batch = int
```
twiecki:
```suggestion
    batch : int
```

pymc/step_methods/pgbart.py
```diff
@@ -97,11 +101,17 @@ class PGBART(ArrayStepShared):
...
+        Number of trees fitted per step. Defaults to  "auto", which is the 10% of the `m` trees
+        during tuning and 20% after tuning.
     model: PyMC Model
         Optional model for sampling step. Defaults to None (taken from context).
 
+    Note
+    ----
+    This sampler is inspired on [Lakshminarayanan2015] Particle Gibbs sampler, but introduces
```
twiecki:
```suggestion
    This sampler is inspired by the [Lakshminarayanan2015] Particle Gibbs sampler, but introduces
```

pymc/step_methods/pgbart.py
```diff
@@ -97,11 +101,17 @@ class PGBART(ArrayStepShared):
...
+    several changes. The changes will be proeperly documented soon.
```
twiecki:
```suggestion
    several changes. The changes will be properly documented soon.
```

twiecki (approved):

PR #5047, opened 2021-10-06
GLM glossary term

mjhajharia:
this PR adds GLM to glossary.md

### DIFF
--- a/docs/source/glossary.md
+++ b/docs/source/glossary.md
@@ -59,4 +59,11 @@ Posterior
 
   - For more on generating and interpreting the posterior samples, see Statistical Rethinking 2nd Edition By Richard McElreath, chapter 3.
 
+Generalized Linear Model(GLM)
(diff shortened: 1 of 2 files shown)


docs/source/glossary.md
```diff
@@ -56,4 +56,12 @@ Posterior
   Once you are satisfied with the model, posterior distribution can be summarized and interpreted. Common questions for the posterior include: intervals of defined boundaries, intervals of defined probability mass, and point estimates. When the posterior is very similar to the prior, the available data does not contain much information about a parameter of interest.
 
   - For more on generating and interpreting the posterior samples, see Statistical Rethinking 2nd Edition By Richard McElreath, chapter 3.
+
+Generalized Linear Model(GLM)
```
OriolAbril: I would write generalized linear models and glm in different lines instead. Whatever we use as name in the glossary is what we have to use to reference it. Either or
```
{term}`glm`
{term}`generalized linear models`
```

is fine, but I'd rather not have to write both name and acronym every time.

OriolAbril (commented):
side note, we should also choose glm or generalized linear models to use in tags: https://pymc-examples.readthedocs.io/en/latest/blog/tag/glm.html, https://pymc-examples.readthedocs.io/en/latest/blog/tag/generalized-linear-model.html

and I guess we could add links from glossary definitions to the correspondent tag in pymc-examples?

OriolAbril (approved):

PR #5057, opened 2021-10-08
add 404 page to docs

OriolAbril:
This should add a 404 page to the documentation so when users follow a non existing link
within the docs.pymc.io domain they'll go to this page instead of the rtd maze page.

Follows what I did for ArviZ, if you go to a non existing page like https://arviz-devs.github.io/arviz/generated/arviz.plot_autocorr.html
you see a somewhat informative page.


OriolAbril:
You can preview the 404 page from any wrong url with the base corresponding to this PR, for example https://pymc--5057.org.readthedocs.build/en/5057/api/inexistent.html

twiecki:
This is great. I'm wondering if we could also redirect any links to e.g. docs.pymc.io/notebook to docs.pymc.io/en-stable/notebook? This way all our links will continue to work.

twiecki:
I think we should first try to auto-direct actually, *a lot* of those links will never be updated (e.g. if they are in a book) and still keeping them working is pretty easy by just redirecting to stable.

OriolAbril:
> I'm wondering if we could also redirect any links to e.g. docs.pymc.io/notebook to docs.pymc.io/en-stable/notebook? This way all our links will continue to work.
> I think we should first try to auto-direct actually, a lot of those links will never be updated (e.g. if they are in a book) and still keeping them working is pretty easy by just redirecting to stable.

redirecting would be great _too_, both options should be perfectly compatible with one another, and the 404 page will probably be quite useful with the 4.0 release because there the change won't only be this language plus version extension to the url, most pages will be reorganized or renamed. We will also use sphinx-design in v4 docs which is the pip only dependency that is somehow breaking pre-commit so this is also something that needs to be adressed at some point, why not now to get this page up.

I have no idea if this redirection is possible nor how to do it, and I can't look into it either in the near-mid future either. We should probably open an issue for that or even ask for help on twitter. The only comment regarding the redirection is that it should **not** add `/en/stable/` because we'll again all the links with the 4.0 release, nor stable nor latest are permanent links, they change where they point to with our releases. As we want these links to work and be somewhat frozen, they should redirect to `/en/v3.11.4` instead.

twiecki:
OK this turned out to be trivial: https://docs.readthedocs.io/en/stable/user-defined-redirects.html should work now.

twiecki:
```
- hook id: conda-env-sort
68
- exit code: 1
69

70
71
  File "/home/runner/work/pymc/pymc/scripts/sort_conda_envs.py", line 19, in <module>
72
    doc["dependencies"].sort()
73
TypeError: '<' not supported between instances of 'dict' and 'str'
74

75
Generate pip dependency from conda.......................................Passed
76
No relative imports......................................................Passed
77
Error: The process '/opt/hostedtoolcache/Python/3.9.7/x64/bin/pre-commit' failed with exit code 1
Post Run actions/checkout@v2

```

OriolAbril:
@Sayam753 @michaelosthege @MarcoGorelli maybe, ideas on how to fix pre-commit?

michaelosthege:
@OriolAbril I'm not familiar with `yaml` but it looks like you'll have to go in manually (maybe with an interactive debugger) and replace the sorting with something that recursively sorts hierarchies too.
https://github.com/pymc-devs/pymc/blob/main/scripts/sort_conda_envs.py

MarcoGorelli:
Can you try using the conda env sorter from https://github.com/MarcoGorelli/madforhooks ? IIRC it handles this - else I'll take a look next week

Sayam753:
The failing pre-commit hooks changed the files in place. So, @OriolAbril to fix them all -
```
pre-commit run --all-files
```
And push?

OriolAbril:
I will try that, but the environments and requirement checks are broken or didn't work from the start. I also did some clean-up in #5060 and not only the version requirements were different on different files, but also nbsphinx was listed as requirement in some places but not everywhere. Or it doesn't run in all cases where one of the modified files is changed

twiecki:
Thanks @OriolAbril!

Issue #4884, opened 2021-07-26
Split BART into its own submodule


canyon289:
## Description of your problem
BART is currently contained in the `distributions/` submodule which doenst quite made sense. 
Currently BART is part of `distributions/` in three modules 
https://github.com/pymc-devs/pymc3/blob/main/pymc3/distributions/bart.py  
https://github.com/pymc-devs/pymc3/blob/main/pymc3/distributions/tree.py
plus the PG sampler inside step method

Refactor request is to make BART its own submodule like GP and ODE are since BART is more of a "model architecture" than a distribution

PR #5058, opened 2021-10-08
move BART to its own module

aloctavodia:
Solves #4884. I do not agree with the characterization of BART as a "model architecture", I think is better to think about it as an stochastic process (or at least as something that approximated that). But GPs have the same description and they are also separated in its own module. Also this separation makes sense when in the near future we (I) add some utilities to work with BART.


michaelosthege (approved):
Structure. I like it

PR #5065, opened 2021-10-09
Add `Model.eval_rv_shapes` for fast shape evaluation

michaelosthege:
This PR adds a `Model.eval_rv_shapes` method that can be used to evaluate the shapes of all free untransformed and transformed RVs at the same time.

Consequently the `_check_start_shapes` function that's applied at the start of every sampling became much simpler.

Even though the new shape checks only evaluate the `.shape` of tensors, the `TestNutsCheckTrace.test_linalg` still errored from a finite-check.
Apparently the `at.slinalg.solve` Op also applies the finite-check in shape inference (on purpose? cc @brandonwillard).

+ [x] what are the (breaking) changes that this PR makes? → no breaking changes 🎉
+ [x] important background, or details about the implementation → above
+ [x] are the changes—especially new features—covered by tests and docstrings? → ✔️ 

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -62,7 +62,14 @@ from pymc.distributions import logp_transform, logpt, logpt_sum
 from pymc.distributions.transforms import Transform
 from pymc.exceptions import ImputationWarning, SamplingError, ShapeError
 from pymc.math import flatten_list
-from pymc.util import UNSET, WithMemoization, get_var_name, treedict, treelist
+from pymc.util import (
+    UNSET,
+    WithMemoization,
+    get_transformed_name,
+    get_var_name,
+    treedict,
+    treelist,
+)
 from pymc.vartypes import continuous_types, discrete_types, typefilter
 
 __all__ = [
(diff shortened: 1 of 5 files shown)


pymc/model.py
```diff
@@ -1603,6 +1610,33 @@ def update_start_vals(self, a: Dict[str, np.ndarray], b: Dict[str, np.ndarray]):
...
+            names.append(rv.name)
+            outputs.append(rv.shape)
+        f = aesara.function(
+            inputs=[],
+            outputs=outputs,
+            givens=[(obs, obs.tag.observations) for obs in self.observed_RVs],
+            mode=aesara.compile.mode.FAST_COMPILE,
+            on_unused_input="ignore",
+        )
+        return {name: tuple(shape) for name, shape in zip(names, f())}
```
twiecki, line 1638: Why use names here and not the RV objects?

michaelosthege: Would if I could... We don't have tensors for transforms.

twiecki (approved):

PR #5067, opened 2021-10-10
Fixed Interpolated Distribution's interval transform initialization

kc611:
Linked Issue: https://github.com/pymc-devs/pymc/issues/5048

### DIFF
--- a/pymc/tests/test_distributions.py
+++ b/pymc/tests/test_distributions.py
@@ -20,6 +20,8 @@ import aesara.tensor as at
 import numpy as np
 import numpy.random as nr
 
+from pymc.util import UNSET
+
 try:
     from polyagamma import polyagamma_cdf, polyagamma_pdf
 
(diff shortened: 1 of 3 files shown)


ricardoV94 (changes requested):
Looks good, but we should add a regression test

pymc/tests/test_distributions.py
```diff
@@ -2676,6 +2676,13 @@ def ref_pdf(value):
 
                 self.check_logp(TestedInterpolated, R, {}, ref_pdf)
 
+    def test_interpolated_inits(self):
+        # Issue: https://github.com/pymc-devs/pymc/issues/5048
+        x_points = np.linspace(0, 10, 100000)
+        pdf_points = sp.norm.pdf(x_points, loc=1, scale=1)
+        with pm.Model() as m:
+            x = pm.Interpolated("interpolated", x_points, pdf_points)
```
ricardoV94: We should assert something, even though this was failing before the PR by itself. Perhaps eval the logp outside of the range and check it's not -inf?

ricardoV94 (changes requested):

ricardoV94 (approved):

ricardoV94:
@kc611 I updated the PR. Let me know if you see any problems

kc611 (commented):
Looks Good. Back then I was confused as of what tests to add. This looks pretty reasonable.

PR #5062, opened 2021-10-09
Fix coords in constant_data issue #5046

michaelosthege:
I got a minimum example to reproduce the problem.

@OriolAbril it looks like ArviZ `0.11.4` also has a bug there. This is the traceback when passing to `dict_to_dataset` from ArviZ:

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  

shape = (1, 17)
coords = {'backwards': <xarray.IndexVariable 'backwards' (backwards: 17)>
array([16, 15, 14, 13, 12, 11, 10,  9,  8,  7,  6,  5...: <xarray.IndexVariable 'draw' (draw: 17)>
array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16])}
dims = ['chain', 'draw', 'backwards']

    def _infer_coords_and_dims(
        shape, coords, dims
    ) -> "Tuple[Dict[Any, Variable], Tuple[Hashable, ...]]":
        """All the logic for creating a new DataArray"""

        if (
            coords is not None
            and not utils.is_dict_like(coords)
            and len(coords) != len(shape)
        ):
            raise ValueError(
                f"coords is not dict-like, but it has {len(coords)} items, "
                f"which does not match the {len(shape)} dimensions of the "
                "data"
            )

        if isinstance(dims, str):
            dims = (dims,)

        if dims is None:
            dims = [f"dim_{n}" for n in range(len(shape))]
            if coords is not None and len(coords) == len(shape):
                # try to infer dimensions from coords
                if utils.is_dict_like(coords):
                    # deprecated in GH993, removed in GH1539
                    raise ValueError(
                        "inferring DataArray dimensions from "
                        "dictionary like ``coords`` is no longer "
                        "supported. Use an explicit list of "
                        "``dims`` instead."
                    )
                for n, (dim, coord) in enumerate(zip(dims, coords)):
                    coord = as_variable(coord, name=dims[n]).to_index_variable()
                    dims[n] = coord.name
[...]

PR #5060, opened 2021-10-08
Remove workaround in InferenceData conversion; require ArviZ >=0.11.4

michaelosthege:
This PR removes a workaround in the InferenceData conversion and at the same time fixes a bug that was present in the workaround.

Depending on what your PR does, here are a few things you might want to address in the description:
+ [x] what are the (breaking) changes that this PR makes? → possibly ArviZ API changes
+ [x] important background, or details about the implementation → 
+ [x] are the changes—especially new features—covered by tests and docstrings? → yup
+ [x] [linting/style checks have been run](https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style) → aye

PR #4759, opened 2021-06-08
Move from nbsphinx to myst-nb

OriolAbril:
My plan is to first make this change, which will work once the widget issue in
https://github.com/pymc-devs/pymc-examples/issues/83 is fixed, @chiral-carbon will
send a PR for this at some point tomorrow. Moving from nbsphinx to myst-nb will allow us
to use all sphinx features from within markdown. Therefore, if we write a glossary,
we'll be able to link to terms defined there with `` {term}`RV` `` from within any markdown cell
in notebooks, we'll be able to link to respective api docs with ``
{func}`pymc3.to_inference_data` `` or `` {class}`pymc3.Normal` ``, to define anchors with
`(anchor_name)=` and reference them with `` {ref}`anchor_name` `` and so on.

After that, I will also change the theme to pydata-sphinx-theme, like pandas and ArviZ. Note that
both changes are independent and we could do only one without the other, which is why I'm doing two
separate PRs. I think it would be great to change the appearance of the docs with 4.0, but
it would still be good to have myst already in v3 so we can start using it in pymc-examples without
breaking anything. We can then cherry pick this one onto v4 but not port the theme one.

cc @martinacantaro @ricardoV94 

+ [x] what are the (breaking) changes that this PR makes? **None**
+ [x] important background, or details about the implementation **see above**
+ [x] [linting/style checks have been run](https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style)
+ [ ] right before it's ready to merge, mention the PR in the RELEASE-NOTES.md

PR #5070, opened 2021-10-12
Fix dependencies in requirements-dev.txt

Sayam753:
The requirements-dev were updated in PR - https://github.com/pymc-devs/pymc/pull/4932. After the merge, RTD could not build docs because of missing dependencies. So, this PR reverts the changes in requirements-dev.txt.

I am not sure of the differences between `environment-dev-py37.yml` and `environment-test-py37.yml`. But `sphinx_design` and `sphinx-notfound-page` are not present in environment-test* yml files. So, is there a way to generate these yml files from requirements-dev.txt?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,7 @@ repos:
     - id: pip-from-conda
       additional_dependencies: [pyyaml]
       entry: python scripts/generate_pip_deps_from_conda.py
-      files: ^conda-envs/
+      files: ^conda-envs/environment-dev-py.+.yml$
       language: python
       name: Generate pip dependency from conda
     - id: no-relative-imports
--- a/conda-envs/environment-dev-py37.yml
+++ b/conda-envs/environment-dev-py37.yml
@@ -1,3 +1,4 @@
+# "dev" conda envs are to be used by devs in setting their local environments
 name: pymc-dev-py37
 channels:
 - conda-forge
@@ -8,7 +9,7 @@ dependencies:
 - cachetools>=4.2.1
 - cloudpickle
 - fastprogress>=0.2.0
-- h5py
+- h5py>=2.7
 - ipython
 - myst-nb
 - numpy>=1.15
@@ -20,13 +21,11 @@ dependencies:
 - pytest>=3.0
 - python-graphviz
 - python=3.7
-- recommonmark>=0.4
 - scipy>1.4.1
 - sphinx-autobuild>=0.7
 - sphinx-notfound-page
 - sphinx-panels
 - sphinx>=1.5
 - typing-extensions
-- watermark
 - pip:
   - sphinx-design
--- a/conda-envs/environment-dev-py38.yml
+++ b/conda-envs/environment-dev-py38.yml
@@ -1,3 +1,4 @@
+# "dev" conda envs are to be used by devs in setting their local environments
 name: pymc-dev-py38
 channels:
 - conda-forge
@@ -8,7 +9,7 @@ dependencies:
 - cachetools>=4.2.1
 - cloudpickle
 - fastprogress>=0.2.0
-- h5py
+- h5py>=2.7
 - ipython
 - myst-nb
 - numpy>=1.15.0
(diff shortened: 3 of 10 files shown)


Sayam753:
Turns out that the requirements-dev.txt is generated from conda env files in the pre-commit.

OriolAbril:
(we should probably end up moving that to an issue and not block this PR, but I wanted to start the
discussion here)

I want to take this opportunity and rethink all these conda environments we have. I understand they are helpful and convenient to create our dev environments and they are also used for ci. However, at least in my case, the extra work and burden on updating 4 (8 now) nearly equivalent files with a pre-commit check that seems to only partially work (i.e. pandas has a lower bound in 3.7 and 3.8 but not in 3.9 nor in requirements-dev.txt) is clearly larger than the benefit I get. Moreover, issues with these files, installation and pre-commit blocked #5060. #5062 was only needed because of #5060 being blocked.

I think there are several issues at hand regarding all this.

* No single source of truth. From what I understand, the pre-commit script somehow takes all conda env files and from them generates the pip file. However, it seems like conda files don't need to be equivalent between them (see for example from  #5057 how nbsphinx was on some conda files but not all of them, as well as several versions not matching between files).
* Overly bloated and non-specific dependency files. We have 2/3 (still not completely understand the new test file but it seems like a step in the right direction)  types of requirements files, library requirements (user facing), dev and test ones. But we use those files to do **much** more than 3 types of things. We generate local envs for devs, we set up CI envs in github acions for testing, we set up the env on rtd to build documentation... So here are some things that we are doing and I think we should avoid as they also make our ci more brittle and prone to conda finding version conflicts:
  * We are installing sphinx and sphinx extensions everywhere, even on github actions
  * We are installing pytest on rtd
  * We are installing libraries like watermark or sphinx-autobuild also everywhere, but these are only used locally! watermark is only needed when writing/running example notebooks but not to build the docs that contain such notebooks. sphinx-autobuild is only needed to get an auto-updating preview of the docs _locally_.
* Most of us (or maybe even all of us) have no idea about what some of these dependencies are for. I for example have no idea what `cachetools` is for. I think that this is both issue and a consequence of the previous two, but I think it's important to put that on the table, because in my opinion, unless we make an effort in explaining all that in the developer guide and keep it up to date it will only get worse.

Ideas for possible solutions:
* Being more strict about CI. I have the feeling that some PRs are blocked when only pre-commit fails when others are green-lighted with failing tests and/or rtd job. I haven't written much tests myself here and I imagine testing distributions is inherently hard, but I think we need some way of getting rid of flaky tests or grouping them in a "potentially flaky" job, any idea here welcome. I think that having flaky tests makes ourselves open to the possibility of "maybe this is a flaky test too, I don't recall changing anything that should affect this test" or similar reasonings
* Being more strict about reviews. In a similar manner, I have a feeling that some PRs are merged without approving reviews or with approving reviews even though there were still open questions in the PR comments.
* Split requirements into task related chunks. This by design eliminates the 1 line to dev environment/single file to serve all, so we'll need to install 2-4 files, I'm also not sure how that would fit with all the pre-commit scripts but I'm sure we'll figure it out, maybe even write a helper script to keep creating dev envs a single line operation. Here are some ideas:
  * requirements.txt -> requirements like we have now for pymc
  * requirements-optional.txt -> optional requirements for pymc, we could also update `setup.py` so that running `pip install pymc[all]` installs both requiremetns and requirements-optional
  * requirements-test.txt -> requirements for testing
  * requirements-docs.txt -> requirements for building the docs
  * requirements-write.txt -> requirements for writing example notebooks, see https://github.com/pymc-devs/pymc-examples/blob/main/requirements-write.txt
  * others?
* **Documenting everything**. Explain on the developer guide/contributing guide when should each file be used plus maybe also extra recommendations for specific local env only tools (doc related examples of that would be sphinx-autobuild or sphobjinv libraries)

Sayam753:
+1 on revisiting how we bundle requirements.

OriolAbril (approved):
I have removed unnecessary dependencies from the ci only files to simplify conda's job and hopefully speed up a bit the env creation process

I have also removed recommonmark and nbsphinx from everywhere they appeared. We don't use any of the two anymore, nor we have done since #4759. Please don't add them again, this has to be the 3rd or 4th PR where I have removed them from the requirements.

PR #5073, opened 2021-10-13
Make forward sampling functions return `InferenceData`

AlexAndorra:
This PR adds the possibility to return foward samples directly as an `InferenceData` objects, and sets this behavior as the default. These objects then just need to be extended to the original trace (which is already an `InferenceData` object by default in 4.0), which simplifies the workflow.

I implemented this for `sample_prior_predictive` and you can already review -- I would like to get your approval on the logic before going to `sample_posterior_predictive` and `sample_posterior_predictive_w`, as the implementation will be very similar.

+ [x] Refactor `sample_prior_predictive`
+ [x] Refactor `sample_posterior_predictive`
+ [x] Refactor `sample_posterior_predictive_w`
+ [x] Breaking changes: forward sampling functions now return an `InferenceData` object by default.
+ [x] Mention in the RELEASE-NOTES.md

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -271,7 +271,7 @@ def sample(
     idata_kwargs: dict = None,
     mp_ctx=None,
     **kwargs,
-):
+) -> Union[InferenceData, MultiTrace]:
     r"""Draw samples from the posterior using the given step methods.
 
     Multiple step methods are supported via compound step methods.
@@ -338,7 +338,7 @@ def sample(
         init methods.
     return_inferencedata : bool, default=True
         Whether to return the trace as an :class:`arviz:arviz.InferenceData` (True) object or a `MultiTrace` (False)
-        Defaults to `False`, but we'll switch to `True` in an upcoming release.
+        Defaults to `True`.
     idata_kwargs : dict, optional
         Keyword arguments for :func:`pymc.to_inference_data`
     mp_ctx : multiprocessing.context.BaseContent
@@ -1893,7 +1893,9 @@ def sample_prior_predictive(
     var_names: Optional[Iterable[str]] = None,
     random_seed=None,
     mode: Optional[Union[str, Mode]] = None,
-) -> Dict[str, np.ndarray]:
+    return_inferencedata=None,
+    idata_kwargs: dict = None,
+) -> Union[InferenceData, Dict[str, np.ndarray]]:
     """Generate samples from the prior predictive distribution.
 
     Parameters
@@ -1909,14 +1911,21 @@ def sample_prior_predictive(
         Seed for the random number generator.
     mode:
         The mode used by ``aesara.function`` to compile the graph.
+    return_inferencedata : bool, default=True
+        Whether to return an :class:`arviz:arviz.InferenceData` (True) object or a dictionary (False).
+        Defaults to `True`.
+    idata_kwargs : dict, optional
+        Keyword arguments for :func:`pymc.to_inference_data`
 
     Returns
     -------
-    dict
-        Dictionary with variable names as keys. The values are numpy arrays of prior
-        samples.
+    arviz.InferenceData or Dict
+        An ArviZ ``InferenceData`` object containing the prior and prior predictive samples (default),
+        or a dictionary with variable names as keys and samples as numpy arrays.
     """
     model = modelcontext(model)
+    if return_inferencedata is None:
+        return_inferencedata = True
 
     if model.potentials:
         warnings.warn(
@@ -1980,7 +1989,15 @@ def sample_prior_predictive(
     for var_name in vars_:
         if var_name in data:
             prior[var_name] = data[var_name]
-    return prior
+
+    if not return_inferencedata:
+        return prior
+
+    ikwargs = dict(model=model)
+    if idata_kwargs:
+        ikwargs.update(idata_kwargs)
+
+    return pm.to_inference_data(prior=prior, **ikwargs)
 
 
 def _init_jitter(model, point, chains, jitter_max_retries):
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -267,7 +267,7 @@ def sample(
     callback=None,
     jitter_max_retries=10,
     *,
-    return_inferencedata=None,
+    return_inferencedata=True,
     idata_kwargs: dict = None,
     mp_ctx=None,
     **kwargs,
@@ -336,7 +336,7 @@ def sample(
         Maximum number of repeated attempts (per chain) at creating an initial matrix with uniform jitter
         that yields a finite probability. This applies to ``jitter+adapt_diag`` and ``jitter+adapt_full``
         init methods.
-    return_inferencedata : bool, default=True
+    return_inferencedata : bool
         Whether to return the trace as an :class:`arviz:arviz.InferenceData` (True) object or a `MultiTrace` (False)
         Defaults to `True`.
     idata_kwargs : dict, optional
@@ -449,9 +449,6 @@ def sample(
 
     if not isinstance(random_seed, abc.Iterable):
         raise TypeError("Invalid value for `random_seed`. Must be tuple, list or int")
-
-    if return_inferencedata is None:
-        return_inferencedata = True
 
     if not discard_tuned_samples and not return_inferencedata:
         warnings.warn(
@@ -1893,7 +1890,7 @@ def sample_prior_predictive(
     var_names: Optional[Iterable[str]] = None,
     random_seed=None,
     mode: Optional[Union[str, Mode]] = None,
-    return_inferencedata=None,
+    return_inferencedata=True,
     idata_kwargs: dict = None,
 ) -> Union[InferenceData, Dict[str, np.ndarray]]:
     """Generate samples from the prior predictive distribution.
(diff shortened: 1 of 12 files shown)


pymc/sampling.py
```diff
@@ -1893,7 +1893,9 @@ def sample_prior_predictive(
     var_names: Optional[Iterable[str]] = None,
     random_seed=None,
     mode: Optional[Union[str, Mode]] = None,
-) -> Dict[str, np.ndarray]:
+    return_inferencedata=None,
```
ricardoV94: Why not?

```suggestion
    return_inferencedata=True,
```

pymc/sampling.py
```diff
@@ -1909,14 +1911,21 @@ def sample_prior_predictive(
         Seed for the random number generator.
     mode:
         The mode used by ``aesara.function`` to compile the graph.
+    return_inferencedata : bool, default=True
+        Whether to return an :class:`arviz:arviz.InferenceData` (True) object or a dictionary (False).
+        Defaults to `True`.
+    idata_kwargs : dict, optional
+        Keyword arguments for :func:`pymc.to_inference_data`
```
aloctavodia:
```suggestion
    return_inferencedata : bool
        Whether to return an :class:`arviz:arviz.InferenceData` (True) object or a dictionary (False).
        Defaults to True.
    idata_kwargs : dict, optional
        Keyword arguments for :func:`pymc.to_inference_data`
```

pymc/sampling.py
AlexAndorra: I don't know why, but that's how it's done in `pm.sample`. If there is no particular reason, I agree we should switch to `return_inferencedata=True`, even in `sample`

AlexAndorra: I switched to this, it doesn't seem to be a problem

AlexAndorra:
This is all done and ready for (final?) review 🥳

AlexAndorra:
This is weird: all the tests for `sample_smc` are failing... but I didn't change anything in there 🤔  
Any idea why @aloctavodia ?

AlexAndorra:
I managed to fix all the tests, except those related to SMC

ricardoV94:
> I managed to fix all the tests, except those related to SMC

SMC uses prior predictive sampling internally. It should be a simple matter of setting return_inferencedata to False there

ricardoV94:
You have a lot of unrelated changes due to your pre-commit. This seems to always happen with your setup :P (and you should revert them...)

AlexAndorra:
> SMC uses prior predictive sampling internally.

Aaaaah, that's why!

> You have a lot of unrelated changes due to your pre-commit

I didn't see any unrelated changes -- this is just Black doing its thing through pre-commit. And I did change quite a lot of lines (mainly tests; we use forward sampling a lot 😅 ). 
Note that I deleted `test_normal_scalar_idata`, this is not a typo. It's because I just switched `test_normal_scalar` to use idata by default

pymc/tests/test_distributions_random.py
```diff
@@ -102,13 +102,7 @@ def pymc_random(
...
 
 def pymc_random_discrete(
-    dist,
-    paramdomains,
-    valuedomain=Domain([0]),
-    ref_rand=None,
-    size=100000,
-    alpha=0.05,
-    fails=20,
+    dist, paramdomains, valuedomain=Domain([0]), ref_rand=None, size=100000, alpha=0.05, fails=20
```
michaelosthege: @AlexAndorra looks like you're rolling with an outdated pre-commit configuration.
Maybe a `pre-commit uninstall` + `pre-commit install` is enough to get you updated?

ricardoV94:
Look at `test_distributions_random.py`, it's full of black changes in lines that were not modified

pymc/tests/test_sampling.py
```diff
@@ -987,9 +994,9 @@ def test_transformed(self):
             y = pm.Binomial("y", n=at_bats, p=thetas, observed=hits)
             gen = pm.sample_prior_predictive(draws)
 
-        assert gen["phi"].shape == (draws,)
-        assert gen["y"].shape == (draws, n)
-        assert "thetas" in gen
+        assert gen.prior["phi"].shape == (1, draws)
```
michaelosthege, line 1022: What's that additional prior dimension? `"chain"` doesn't make much sense for prior predictive..

michaelosthege (commented):
Good move to add this capability!
However, I'd prefer if we didn't add an unnecessary dimension to the prior.

With these changes, what's the recommended workflow if I want to combine the MCMC, prior and posterior traces into the same `InferenceData`? Is there a `idata.concat()` or `idata.swallow()` method or something like that?

AlexAndorra:
> Look at test_distributions_random.py, it's full of black changes in lines that were not modified

@ricardoV94: Yes, I saw them, it's just Black. But I think I know what's happening: PyCharm is probably running a Black from another venv than my pymc env! Seems hard to revert only those changes though, isn't it? Maybe easier in another PR?

pymc/tests/test_distributions_random.py
AlexAndorra: I don't think so: my pymc venv is brand new and I pulled from `main`. I _think_ the issue is PyCharm is running a Black from another environment than my pymc env. Is there an easy way to revert those changes though (some of them are not in dedicated commits)?

pymc/tests/test_sampling.py
AlexAndorra: I think it's necessary for the next step: merging the prior pred object automagically with the trace and the post pred samples: `xarray` needs this dimension to match everything, as the two other objects have it (speaking under the control of @aloctavodia and @OriolAbril )

AlexAndorra:
> With these changes, what's the recommended workflow if I want to combine the MCMC, prior and posterior traces into the same InferenceData?

@michaelosthege: I use that:
```python
with model:
    idata = pm.sample()
    idata.extend(pm.sample_prior_predictive())
    idata.extend(pm.sample_posterior_predictive(idata))
```
There may be something even better, but that already works like a (py)charm 🤩

ricardoV94:
> > Look at test_distributions_random.py, it's full of black changes in lines that were not modified
> 
> @ricardoV94: Yes, I saw them, it's just Black. But I think I know what's happening: PyCharm is probably running a Black from another venv than my pymc env! Seems hard to revert only those changes though, isn't it? Maybe easier in another PR?

It seems the bad changes are (all) in the `test_distributions_random.py`? So maybe you can revert those and redo your intended changes?

pymc/tests/test_distributions_random.py
michaelosthege: I don't know any automated way to revert them.
Next time try to stage selected ranges/lines contiously, then it's easier to spot this before committing?

AlexAndorra: This is weird: now I'm _sure_ that the Black ran on this file is from pre-commitn and it still gives me diffs compared to `main`. Doesn't that mean that there was a problem with this even before my PR?

AlexAndorra:
Ooh damn, I thought the problem was coming from
```
[tool.black]
line-length = 100
```
in `pyproject.toml`, but it really doesn't look like it 😅 --> Reverting

AlexAndorra:
> It seems the bad changes are (all) in the test_distributions_random.py? So maybe you can revert those and redo your intended changes?

Ok, I just did that and _think_ I fixed all the tests @ricardoV94 and @michaelosthege. So, if tests pass, everything should be ready to merge 🤞

pymc/tests/test_missing.py
```diff
@@ -27,23 +27,22 @@
...
-    [ma.masked_values([1, 2, -1, 4, -1], value=-1), pd.DataFrame([1, 2, np.nan, 4, np.nan])],
+    "data", [ma.masked_values([1, 2, -1, 4, -1], value=-1), pd.DataFrame([1, 2, np.nan, 4, np.nan])]
 )
 def test_missing(data):
 
     with Model() as model:
         x = Normal("x", 1, 1)
         with pytest.warns(ImputationWarning):
-            y = Normal("y", x, 1, observed=data)
+            _ = Normal("y", x, 1, observed=data)
```
ricardoV94, line 37: Why this change?

ricardoV94 (approved):
Looks good, let's see if the CI passes :D

michaelosthege (approved):
The diff looks fine now.

I'm still not a fan of having an unnecessary `"chain"` dimension in prior predictive. InferenceData can handle `constant_data` without "chain"/"draw" dims just fine. But if that's a matter of what `InferenceData` supports, we should discuss that separately.

pymc/tests/test_missing.py
AlexAndorra: Ha ha, just my own neurosis when I see a named but unused variable 😅

AlexAndorra:
CI passes, CI passes 🍾 (the failing docs are expected)

Agreed @michaelosthege. If this makes you feel any better, this is already the current behavior: a `chain` dim is added to the prior pred object when calling `arviz.from_pymc3` 🤷‍♂️

michaelosthege:
> If this makes you feel any better, this is already the current behavior: a chain dim is added to the prior pred object when calling arviz.from_pymc3 🤷‍♂️

😬

OriolAbril:
To add a bit of extra info to Alex's comment:

> ```
> with model:
>     idata = pm.sample()
>     idata.extend(pm.sample_prior_predictive())
>     idata.extend(pm.sample_posterior_predictive(idata))
> ```

The sampling and extends are "commutative". That is, this would also work

```
with model:
    idata = pm.sample_prior_predictive()
    idata.extend(pm.sample())
    idata.extend(pm.sample_posterior_predictive(idata))
```

PR #5083, opened 2021-10-17
Add backports module and backport logp syntax

### LINKED ISSUE #5052: Backport `pm.logp` and related functional syntax to `v3`
To reduce the rate of users running into problems like #5048, we should backport the `pm.logp`, `pm.logcdf` etc syntax to `v3`.

While it will of course not have all the features available in `v4` (for example doing `pm.logp()` on things other than what's returned by `Distribution()`), the availability of the new syntax allow us to replace all uses of `rv.logp(x)` with forwards-compatible `pm.logp(rv, x)` (like #5051).

The actual implementation may be as simple as:
```python
def logp(rv, x):
    return rv.logp(x)
```

* PyMC3 Version: `v3.11.x`

michaelosthege:
I did not backport the other related functions (e.g. logcdf), but the backport of `pm.logp()` alone should cover >90 % of the use cases.

Closes #5052

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -13,6 +13,7 @@ jobs:
         floatx: [float32, float64]
         test-subset:
           - |
+            --ignore=pymc3/tests/test_backports.py
             --ignore=pymc3/tests/test_dist_math.py
             --ignore=pymc3/tests/test_distribution_defaults.py
             --ignore=pymc3/tests/test_distributions.py
@@ -42,6 +43,7 @@ jobs:
             pymc3/tests/test_shared.py
             pymc3/tests/test_smc.py
           - |
+            pymc3/tests/test_backports.py
             pymc3/tests/test_examples.py
             pymc3/tests/test_mixture.py
             pymc3/tests/test_ode.py
--- a/pymc3/backports.py
+++ b/pymc3/backports.py
@@ -28,12 +28,12 @@ def logp(
     """
     Calculate log-probability of a distribution at specified value.
 
-    This function is a backported poor-mans version of PyMC >=4.0 like capabilities.
+    This function is a limited functionality backported version of PyMC >=4.0 like capabilities.
 
     Parameters
     ----------
     value : numeric
-        Value(s) for which log-probability is calculated. If the log probabilities for multiple
+        Value(s) for which log-probability is calculated. If the log-probabilities for multiple
         values are desired the values must be provided in a numpy array or theano tensor
 
     Returns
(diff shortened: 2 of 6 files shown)


michaelosthege:
The compilation error in that test appeared on another PR a few days ago already.
Looks like a compilation dependency is slowly falling apart..

.github/workflows/pytest.yml
```diff
@@ -13,6 +13,7 @@ jobs:
         floatx: [float32, float64]
         test-subset:
           - |
+            --ignore=pymc3/tests/test_backports.py
```
canyon289, line 16: Just curious why do we need to ignore it here? is it due to CI time?

pymc3/backports.py
```diff
@@ -0,0 +1,43 @@
...
+from pymc3.model import Factor
+
+
+def logp(
+    rv: Union[Factor, Distribution], value: Union[TensorVariable, np.ndarray]
+) -> Union[TensorVariable, np.ndarray]:
+    """
+    Calculate log-probability of a distribution at specified value.
+
+    This function is a backported poor-mans version of PyMC >=4.0 like capabilities.
```
canyon289:
```suggestion
    This function is a limited functionality backported version of PyMC >=4.0 like capabilities.
```
Replacing with a more neutral word

pymc3/backports.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Parameters
+    ----------
+    value : numeric
+        Value(s) for which log-probability is calculated. If the log probabilities for multiple
```
canyon289: If we write log-probability with the em dash so we also write log-probabilities?

canyon289 (dismissed):
Assuming CI failures are unrelated per comment

pymc3/backports.py
michaelosthege:
```suggestion
        Value(s) for which log-probability is calculated. If the log-probabilities for multiple
```

.github/workflows/pytest.yml
michaelosthege: The first job in this matrix is the fall-back which runs all test files that aren't explicitly mentioned by the others.
It's complemented by the `check_all_tests_are_covered.py` pre-commit script checking that there's also no test file that's run by another job but **not** ignored here.

Pro-tip: On `main` you can run `python scripts/check_all_tests_are_covered.py` locally and it shows you which OS/floatX combination each test file is running in.

twiecki (approved):

canyon289:
Thank you @michaelosthege

PR #5084, opened 2021-10-17
Add dims to *Ordered probs variables when appropriate

AustinRochford:
Add `dims` to `_probs` RV for ordered distributions when appropriate.


ColCarroll (approved):
If there's a small repro test you wanted to add, that'd be great, but this looks fine as is.

mjhajharia (approved):

PR #5085, opened 2021-10-18
Added Generalized Extreme Value distribution

ccaprani:
This PR adds the GEV distribution to Pymc v4. An example notebook is also submitted as PR to pymc-examples: https://github.com/pymc-devs/pymc-examples/pull/241

Docstrings, an example, and tests are added; code is formatted with black.

Considering the complexity of the bounds for this distribution, a careful review of the tensor algebra used in the `logpdf` and `log_cdf` functions would be very welcome.

### DIFF
--- a/pymc/tests/test_distributions.py
+++ b/pymc/tests/test_distributions.py
@@ -75,6 +75,7 @@ from pymc.distributions import (
     Flat,
     Gamma,
     Geometric,
+    GenExtreme,
     Gumbel,
     HalfCauchy,
     HalfFlat,
@@ -2542,6 +2543,20 @@ class TestMatchesScipy:
             R,
             {"mu": R, "beta": Rplusbig},
             lambda value, mu, beta: sp.gumbel_r.logcdf(value, loc=mu, scale=beta),
+        )
+
+    def test_genextreme(self):
+        self.check_logp(
+            GenExtreme,
+            R,
+            {"mu": R, "sigma": Rplus, "xi": Domain([-1, 1])},
+            lambda value, mu, sigma, xi: sp.genextreme.logpdf(value, c=-xi, loc=mu, scale=sigma),
+        )
+        self.check_logcdf(
+            GenExtreme,
+            R,
+            {"mu": R, "sigma": Rplus, "xi": Domain([-1, 1])},
+            lambda value, mu, sigma, xi: sp.genextreme.logcdf(value, c=-xi, loc=mu, scale=sigma),
         )
 
     def test_logistic(self):
--- a/pymc/tests/test_distributions.py
+++ b/pymc/tests/test_distributions.py
@@ -2551,14 +2551,12 @@ class TestMatchesScipy:
             R,
             {"mu": R, "sigma": Rplus, "xi": Domain([-1, -1, -0.5, 0, 0.5, 1, 1])},
             lambda value, mu, sigma, xi: sp.genextreme.logpdf(value, c=-xi, loc=mu, scale=sigma),
-            n_samples=-1,
         )
         self.check_logcdf(
             GenExtreme,
             R,
             {"mu": R, "sigma": Rplus, "xi": Domain([-1, -1, -0.5, 0, 0.5, 1, 1])},
             lambda value, mu, sigma, xi: sp.genextreme.logcdf(value, c=-xi, loc=mu, scale=sigma),
-            n_samples=-1,
         )
 
     def test_logistic(self):
--- a/pymc/tests/test_distributions.py
+++ b/pymc/tests/test_distributions.py
@@ -2549,14 +2549,18 @@ class TestMatchesScipy:
         self.check_logp(
             GenExtreme,
             R,
-            {"mu": R, "sigma": Rplus, "xi": Domain([-1, 1])},
-            lambda value, mu, sigma, xi: sp.genextreme.logpdf(value, c=-xi, loc=mu, scale=sigma),
+            {"mu": R, "sigma": Rplus, "xi": Domain([-1, -1, -0.5, 0, 0.5, 1, 1])},
+            lambda value, mu, sigma, xi: sp.genextreme.logpdf(
+                value, c=-xi, loc=mu, scale=sigma, n_samples=-1
+            ),
         )
         self.check_logcdf(
             GenExtreme,
             R,
-            {"mu": R, "sigma": Rplus, "xi": Domain([-1, 1])},
-            lambda value, mu, sigma, xi: sp.genextreme.logcdf(value, c=-xi, loc=mu, scale=sigma),
+            {"mu": R, "sigma": Rplus, "xi": Domain([-1, -1, -0.5, 0, 0.5, 1, 1])},
+            lambda value, mu, sigma, xi: sp.genextreme.logcdf(
+                value, c=-xi, loc=mu, scale=sigma, n_samples=-1
+            ),
         )
 
     def test_logistic(self):
--- a/pymc/tests/test_distributions.py
+++ b/pymc/tests/test_distributions.py
@@ -2550,17 +2550,15 @@ class TestMatchesScipy:
             GenExtreme,
             R,
             {"mu": R, "sigma": Rplus, "xi": Domain([-1, -1, -0.5, 0, 0.5, 1, 1])},
-            lambda value, mu, sigma, xi: sp.genextreme.logpdf(
-                value, c=-xi, loc=mu, scale=sigma, n_samples=-1
-            ),
+            lambda value, mu, sigma, xi: sp.genextreme.logpdf(value, c=-xi, loc=mu, scale=sigma),
+            n_samples=-1,
         )
         self.check_logcdf(
             GenExtreme,
             R,
             {"mu": R, "sigma": Rplus, "xi": Domain([-1, -1, -0.5, 0, 0.5, 1, 1])},
-            lambda value, mu, sigma, xi: sp.genextreme.logcdf(
-                value, c=-xi, loc=mu, scale=sigma, n_samples=-1
-            ),
+            lambda value, mu, sigma, xi: sp.genextreme.logcdf(value, c=-xi, loc=mu, scale=sigma),
+            n_samples=-1,
         )
 
     def test_logistic(self):
--- a/docs/source/api/distributions/continuous.rst
+++ b/docs/source/api/distributions/continuous.rst
@@ -37,6 +37,7 @@ Continuous
    LogitNormal
    Interpolated
    PolyaGamma
+   GenExtreme
 
 .. automodule:: pymc.distributions.continuous
    :members:
--- a/pymc/distributions/__init__.py
+++ b/pymc/distributions/__init__.py
@@ -32,6 +32,7 @@ from pymc.distributions.continuous import (
     Exponential,
     Flat,
     Gamma,
+    GenExtreme,
     Gumbel,
     HalfCauchy,
     HalfFlat,
@@ -198,4 +199,5 @@ __all__ = [
     "logcdf",
     "_logcdf",
     "logpt_sum",
+    "GenExtreme",
 ]
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -123,6 +123,7 @@ __all__ = [
     "Moyal",
     "AsymmetricLaplace",
     "PolyaGamma",
+    "GenExtreme",
 ]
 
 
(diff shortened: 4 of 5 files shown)


pymc/tests/test_distributions.py
```diff
@@ -2544,6 +2545,20 @@ def test_gumbel(self):
             lambda value, mu, beta: sp.gumbel_r.logcdf(value, loc=mu, scale=beta),
         )
 
+    def test_genextreme(self):
+        self.check_logp(
+            GenExtreme,
+            R,
+            {"mu": R, "sigma": Rplus, "xi": Domain([-1, 1])},
```
ricardoV94: This will only test the default `xi=0`. `Domain` strips out the two edges. If you wan to test `xi in [-1, 1]` you can write `Domain([-1, -1, 1, 1])`. Also, shouldn't the `xi=0` be tested? If your intention was to test multiple values in the `[-1, 1]` range you should specify more values. Domain should specify an exact grid of values with which to test the `logp/logcdf` (which is probably not intuitive from the name :P)

Also make sure to add `n_samples=-1` locally to make sure all combination are tested and are passing. And do test locally with `float32`, which is usually a source of failure in these tests. You can do this by setting `aesaara.config.floatX = "float32"` at the top of the module, immediately after importing `aesara`

pymc/tests/test_distributions.py
```diff
@@ -2544,6 +2545,20 @@ def test_gumbel(self):
...
+            lambda value, mu, sigma, xi: sp.genextreme.logpdf(value, c=-xi, loc=mu, scale=sigma),
+        )
+        self.check_logcdf(
+            GenExtreme,
+            R,
+            {"mu": R, "sigma": Rplus, "xi": Domain([-1, 1])},
```
ricardoV94: Same concerns for `Domain` of `xi`

ricardoV94 (changes requested):
Thanks for opening the PR. I left a couple of comments below. The logp/logcd expression seems reasonable, if `xi==0` is a valid parameter value (and not just a limiting case, such as with `sigma`).

You seem to be missing tests for the RandomVariable, which should go into `test_distributions_random.py`. Here is a guide that explains how to write these tests and a bit more: https://github.com/pymc-devs/pymc/blob/main/docs/source/developer_guide_implementing_distribution.md

You can also consider adding a `moment` to be used as a stable initialization point. This has not been documented in that guide yet, but you can find some examples for the `normal` distribution here: https://github.com/pymc-devs/pymc/blob/598dd9de2b818a58480071720a9f3da63177be89/pymc/distributions/continuous.py#L602-L603

pymc/tests/test_distributions.py
ccaprani: I'm not quite following setting `aesara.config.floatX = "float32"` at the top of the module? I guess you mean in my notebook perhaps? Or should that be in the tests?

ricardoV94: In the tests, just locally to make sure it passes on all combinations of parameters in both float64 and float32. Often people only test it in the default float64 and don't notice failures in float32. Our our runs we test only a random subset of 100 combinations by default, which sometimes does not reveal the failing cases immediately. Hence why I suggest you try locally with `n_samples=-1` and both float32 and float64.

ccaprani:
These are such great comments @ricardoV94 - thank you very much indeed. 

The parameter `xi` should indeed be tested at `xi = 0` (it's not a bound), so I've added that as well. It should return the same as the Gumbel distribution in that case.

The guide was super helpful, and I should not have missed the bit about adding the tests to `test_distributions_random.py` - thanks again for pointing this out, and my other misunderstandings.

I've made the changes and in running the new improved tests I am having trouble with the `at.switch` in `logp` - it does not seem to be properly recognizing `at.isclose(xi,0)` - is this something I'm doing, or a known problem? The [docs](https://aesara.readthedocs.io/en/latest/library/tensor/basic.html?highlight=isclose#aesara.tensor.isclose) are pretty clear.

ricardoV94:
> I've made the changes and in running the new improved tests I am having trouble with the `at.switch` in `logp` - it does not seem to be properly recognizing `at.isclose(xi,0)` - is this something I'm doing, or a known problem? The [docs](https://aesara.readthedocs.io/en/latest/library/tensor/basic.html?highlight=isclose#aesara.tensor.isclose) are pretty clear.

What problem are you seeing? This works for me:

```python
import numpy as np
import aesara.tensor as at
xi = at.vector('xi')
out = at.isclose(0, xi)
out.eval({xi: np.array([-1e-5, -1e-9, 0, 1e-9, 1e-5])})
# array([False,  True,  True,  True, False])
```

pymc/tests/test_distributions.py
```diff
@@ -2544,6 +2545,24 @@ def test_gumbel(self):
             lambda value, mu, beta: sp.gumbel_r.logcdf(value, loc=mu, scale=beta),
         )
 
+    def test_genextreme(self):
+        self.check_logp(
+            GenExtreme,
+            R,
+            {"mu": R, "sigma": Rplus, "xi": Domain([-1, -1, -0.5, 0, 0.5, 1, 1])},
+            lambda value, mu, sigma, xi: sp.genextreme.logpdf(
+                value, c=-xi, loc=mu, scale=sigma, n_samples=-1
```
ricardoV94: The `n_samples=-1` should be used only for local testing. We choose to limit the number of permutations tested in our end to keep runtimes manageable.

ccaprani: Oops! - thanks, amended this now.

ccaprani:
Thanks again @ricardoV94 - I found a missing minus sign that has helped things along nicely! 

This is a numerically sensitive beast. Here's a short code for my own testing, and the aesara and scipy warnings for certain parameter combinations...I don't think there is much to be done about it though? I could play around with the `rtol` and `atol` for `at.isclose`, but I'm not sure we want to be make decisions on behalf of users?

```python
import pymc as pm
import aesara.tensor as at
from pymc.aesaraf import floatX
import scipy.stats.distributions as sp
import numpy as np
from itertools import product

μ_l = [-1, 0, 1]
σ_l = [0.1, 1, 10]
ξ_l = [-0.9, -0.5, 0, 0.5, 0.9]
x = np.linspace(-2,5)

for μ,σ,ξ in product(μ_l, σ_l, ξ_l):
    print(f"Running: {μ=}, {σ=}, {ξ=}")
    mu = at.as_tensor_variable(floatX(μ))
    sigma = at.as_tensor_variable(floatX(σ))
    xi = at.as_tensor_variable(floatX(ξ))
    logp_pm = pm.GenExtreme.logp(x,mu,sigma,xi).eval()
    logp_sp = sp.genextreme.logpdf(x, c=-ξ, loc=μ, scale=σ)
    if not np.allclose(logp_pm,logp_sp):
        print(f"Fail on {μ=}, {σ=}, {ξ=}")

```

gives these warnings at certain parameter combination:

```bash
Running: μ=-1, σ=0.1, ξ=0.5

/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/aesara/tensor/math_opt.py:1054: RuntimeWarning: divide by zero encountered in true_divide
  v = reduce(np.multiply, num, one) / reduce(np.multiply, denum, one)
/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/aesara/tensor/math_opt.py:1054: RuntimeWarning: divide by zero encountered in double_scalars
  v = reduce(np.multiply, num, one) / reduce(np.multiply, denum, one)

Running: μ=-1, σ=1, ξ=-0.5

/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/scipy/stats/_continuous_distns.py:2492: RuntimeWarning: invalid value encountered in subtract
  -pex2+logpex2-logex2)
/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/aesara/tensor/math_opt.py:1054: RuntimeWarning: divide by zero encountered in true_divide
  v = reduce(np.multiply, num, one) / reduce(np.multiply, denum, one)
/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/aesara/tensor/math_opt.py:1054: RuntimeWarning: divide by zero encountered in double_scalars
  v = reduce(np.multiply, num, one) / reduce(np.multiply, denum, one)


Running: μ=0, σ=1, ξ=0.5

/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/scipy/stats/_continuous_distns.py:2492: RuntimeWarning: invalid value encountered in add
  -pex2+logpex2-logex2)

Running: μ=0, σ=1, ξ=0.9
```

pymc/tests/test_distributions.py
ccaprani: Thanks. Understood now, and I have checked this, and yields the same results as float64.

ricardoV94:
We can lower the threshold of the `allclose` test. We do this for many distributions with the `select_by_precision` that you might have seen in `test_distributions.py`. Otherwise can you choose parameters that do not lead to very extreme logp and see if that helps?

By the way, are those failing `allclose` due to `-inf/inf/nan`?

Edit: regarding the rtol/atol in the logp/logcdf. If the distribution is indeed sensitive to those we can provide them as optional parameters to the users

ccaprani:
I've done quite a bit more exploration on this, and it's not really a problem with the `allclose`, but with evaluations of `at.log1p(0)` - the same happens in Scipy, as you can see from some output below.

This is actually tricky to pin down - sometimes the warnings are emitted, other times not. Nevertheless, the `bound` function kicks in and returns sensible results at the limits `-inf`. So, even though scipy issues warnings for the same problem, should we attempt to avoid these warnings by only computing when inside the support? In other words, a much more nested logp function?

I think at this point, for performance, I'd rather leave as is, and see if these edge case warnings cause issues in the wild or not.

```bash
Running: μ=-1, σ=0.1, ξ=0.5

/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/aesara/tensor/math_opt.py:1054: RuntimeWarning: divide by zero encountered in true_divide
  v = reduce(np.multiply, num, one) / reduce(np.multiply, denum, one)
/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/aesara/tensor/math_opt.py:1054: RuntimeWarning: divide by zero encountered in float_scalars
  v = reduce(np.multiply, num, one) / reduce(np.multiply, denum, one)

Running: μ=-1, σ=1, ξ=-0.5

/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/scipy/stats/_continuous_distns.py:2492: RuntimeWarning: invalid value encountered in subtract
  -pex2+logpex2-logex2)
/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/aesara/tensor/math_opt.py:1054: RuntimeWarning: divide by zero encountered in true_divide
  v = reduce(np.multiply, num, one) / reduce(np.multiply, denum, one)
/home/ccaprani/anaconda3/envs/pymc4/lib/python3.9/site-packages/aesara/tensor/math_opt.py:1054: RuntimeWarning: divide by zero encountered in float_scalars
  v = reduce(np.multiply, num, one) / reduce(np.multiply, denum, one)

```

twiecki:
@ricardoV94 what should we do here?

ricardoV94:
> @ricardoV94 what should we do here?

From https://github.com/pymc-devs/pymc-examples/pull/241, it seems this distribution leads to many sampling issues, arising from the fact that the support domain changes from [-inf, 0], to [-inf, inf] to [0, inf] given the parameters. 

I am not sure whether a custom transform that dynamically converts between lower-bounded, unbounded and upper bounded with a switch logic could fix some of these issues (and whether it is valid in the context of MCMC sampling). 

Without an appropriate transformation this distribution seems to be of little practical use, and we might end up with a situation like one of the Cholesky distributions where we have a warning not to use it...

Perhaps we should include it only as a pymc-example?

twiecki:
That makes sense to me.

ccaprani:
@ricardoV94 @twiecki there are a number of niche distributions already including in pymc, and in the modelling of extremes, this distribution is the only way to determine the extremal tail behaviour directly from the data, i.e. whether Gumbel, Weibull, or Frechet. Without the GEV nested model, there is no Bayesian way to compare between them (in the frequentist setting, people compare the sum of log-likelihoods of fits). While it probably does have a limited user-base, as you can see from Discourse, a number of users have implemented various versions of it in the past, and a benchmark pymc implementation is needed. As for the cases where warnings are emitted; as above, scipy does this too, and so users will be expecting it. If including a warning is required, then that's much better than not having it at all. Thanks!

ricardoV94:
> @ricardoV94 @twiecki there are a number of niche distributions already including in pymc, and in the modelling of extremes, this distribution is the only way to determine the extremal tail behaviour directly from the data, i.e. whether Gumbel, Weibull, or Frechet. Without the GEV nested model, there is no Bayesian way to compare between them (in the frequentist setting, people compare the sum of log-likelihoods of fits). 

There are probably other solutions to address this in a Bayesian conext, including (marginalized?) mixture models and/or model comparison like leave-one-out

> While it probably does have a limited user-base, as you can see from Discourse, a number of users have implemented various versions of it in the past, and a benchmark pymc implementation is needed. As for the cases where warnings are emitted; as above, scipy does this too, and so users will be expecting it. If including a warning is required, then that's much better than not having it at all. Thanks!

It's important to note the different needs and goals of PyMC and SciPy. SciPy does not need to worry about sampling properties of the distribution since it's job is merely to provide random samples, and point evaluations. In contrast, PyMC tends to be concerned with best-practices in the context of Bayesian modelling and MCMC sampling, and sometimes in this context not having a feature may be better than having it at all. 

As I suggested, we could and should have a notebook that we can point users to, which explains how to implement such distribution and warns about the limitations, since this seems to be a popular model overall. 

It is also important to note the effort of adding a distribution to the codebase can be very small compared to the effort of maintaining such distribution in the long term + address related users issues.

***

I am not saying we cannot add this distribution of course. I just expressed my opinion and gave suggestion of work that may be necessary to make this distribution more useful / appropriate within the PyMC framework (i.e., explore whether a specialized transformation can alleviate some of the problems observed).

It goes without saying that I really appreciate time and effort taken here @ccaprani. We should perhaps hear from other developers/ users if you still believe this is a worthwhile addition as is.

ccaprani:
> It is also important to note the effort of adding a distribution to the codebase can be very small compared to the effort of maintaining such distribution in the long term + address related users issues.

I really do understand that, and while I am active on discourse, especially on this distribution, I know that relying on one person is not ideal. I *think* other maintainers will emerge.

>  gave suggestion of work that may be necessary to make this distribution more useful / appropriate within the PyMC framework (i.e., explore whether a specialized transformation can alleviate some of the problems observed).

I'm actually hopeful that bringing this important distribution (in extremes at least) to the wider bayesian community might open up some opportunities like this. I noted in the notebook about a Box-Cox transformation attempt, but this isn't good numerically.

> As I suggested, we could and should have a notebook that we can point users to, which explains how to implement such distribution and warns about the limitations, since this seems to be a popular model overall.

This would certainly be a good second-best, and at least a 'meeting point' on bayesian GEV inference.

> It goes without saying that I really appreciate time and effort taken here @ccaprani. We should perhaps hear from other developers/ users if you still believe this is a worthwhile addition as is.

Sounds good; and as ever all of the pymc devs' time is massively appreciated. This contribution - whether notebook only or not - is only a tiny bit in comparison...thank you!

ccaprani:
@ricardoV94 I've been looking around the literature a bit more on the divergences problem for this distribution. Most of the work uses just MH and so the sampler is not yelling at them as loudly as HMC, for which there are few works. [This one](https://doi.org/10.1080/03610918.2016.1152365) though limits the shape parameter away from heavy tails to deal with the problem, and also uses a log scaling on the scale parameter (mostly so they can assume normal priors). Whereas [this early paper](https://doi.org/10.2307/1403426 ) suggests using the knowledge of the physical problem as reparametrized priors, which will usually lead to reasonable parameter estimations, just like the example in the notebook. 

So if you think it would improve things, I can add more to the notebook (and doc strings) that the behaviour of this distribution does require meaningful priors for the problem domain, be it hydrology, finance, engineering, and so on... Thanks!

ricardoV94:
> @ricardoV94 I've been looking around the literature a bit more on the divergences problem for this distribution. Most of the work uses just MH and so the sampler is not yelling at them as loudly as HMC, for which there are few works. [This one](https://doi.org/10.1080/03610918.2016.1152365) though limits the shape parameter away from heavy tails to deal with the problem, and also uses a log scaling on the scale parameter (mostly so they can assume normal priors). Whereas [this early paper](https://doi.org/10.2307/1403426) suggests using the knowledge of the physical problem as reparametrized priors, which will usually lead to reasonable parameter estimations, just like the example in the notebook.
> 
> So if you think it would improve things, I can add more to the notebook (and doc strings) that the behaviour of this distribution does require meaningful priors for the problem domain, be it hydrology, finance, engineering, and so on... Thanks!

That sounds great. 

@fonnesbeck suggested we may want to add a PyMC module (or separate package) for more untested/experimental stuff, where we would be less worried about adding things like this distribution or more exotic samplers. Anyway, for the time being feel free to update this PR unless someone else wants to veto it.

ricardoV94:
@ccaprani We have created a new repository for more experimental features at https://github.com/pymc-devs/pymc-experimental

Your contribution would be more than welcome over there if you are interested. The pymc-examples could then be updated to import it from there.

Let us know if you want to pick this up and thanks again for all the work already done here!

ccaprani:
@ricardoV94 sounds like a good plan. I presume the proposed `pmx.Distribution` can be mixed with the usual pymc distributions in the same model context?

What are the logistics then? I see that the [PR](https://github.com/pymc-devs/pymc-experimental/pull/3) in the pymc-experiemental repo is being pulled from pymc-main - how do we get this PR over there without needing to fully replicate pymc itself over there (including documentation). I'm probably missing something obvious - any guidance appreciated!

ricardoV94:
> @ricardoV94 sounds like a good plan. I presume the proposed `pmx.Distribution` can be mixed with the usual pymc distributions in the same model context?

Yes, that's the idea

> What are the logistics then? I see that the [PR](https://github.com/pymc-devs/pymc-experimental/pull/3) in the pymc-experiemental repo is being pulled from pymc-main - how do we get this PR over there without needing to fully replicate pymc itself over there (including documentation). I'm probably missing something obvious - any guidance appreciated!

I have no idea what is going on with that PR. I would imagine you would fork the `pymc-experimental` repo and implement your distribution there under an analogous directory `pymc-experimental/distributions/continuous.py` (and make it accessible from the root)

Users can then:
```python
import pymc as pm
import pymc_experimental as pmx
 
with pm.Model():
  x = pmx.GenExtreme(...)
```

ccaprani:
That's wonderful! And the documentation relating to the distribution would go where then?

ricardoV94:
> That's wonderful! And the documentation relating to the distribution would go where then?

Right now just the docstrings. We still have to setup a documentation page for that repository

ccaprani:
This PR is moved to `pymc-experiemental` as https://github.com/pymc-devs/pymc-experimental/pull/4

PR #5086, opened 2021-10-18
Raise ValueError when Domain has no values

ricardoV94:
Small update to make the `Domain` test helper a bit more user friendly.

Situation appeared in #5085


ricardoV94:
This seems to have caught some parameterizations that were not actually being tested in the ordered dists

twiecki (approved):

Issue #5078, opened 2021-10-14
Bring back distribution moments


ricardoV94:
With #4983 and #5087 finished, it's a good time to bring back moments for our distributions for more stable starting points.

With this we can then do the switch in #5009

We should also update the distribution developer guide to mention the implementation of moments: https://github.com/pymc-devs/pymc/blob/main/docs/source/developer_guide_implementing_distribution.md

***
# How to help?

1. This PR should give a template on how to implement and test new moments for distributions: https://github.com/pymc-devs/pymc/pull/5087/files
2. In most cases we should be able to copy the moments we were using in the [V3 branch](https://github.com/pymc-devs/pymc/tree/v3/pymc3/distributions). For example here is what we were doing for the `Beta`
    https://github.com/pymc-devs/pymc/blob/efbaccee94159de0a15b04baf84b1250047f132a/pymc3/distributions/continuous.py#L1235
  2.1 We used to have multiple moments for some distributions such as `mean`, `median`, `mode`. We only support one moment now, and probably the "higher-order" one is the most useful (that is `mean` > `median` > `mode`)... You might need to truncate the moment if you are dealing with a discrete distribution.
  2.2 We left some of these moments commented out inside the distribution `dist` classmethod. Make sure they are removed when you implement them! https://github.com/pymc-devs/pymc/blob/8f3636daf7d9946f6eca4717f3bb0c6d77d9c6e9/pymc/distributions/continuous.py#L538-L539
3. We have to be careful with `size != None` and broadcasting properly when when some parameters that are not used in the `moment` may nevertheless inform about the shape of the distribution. E.g. `pm.Normal.dist(mu=0, sigma=np.arange(1, 6))` returns a moment of `[mu, mu, mu, mu, mu]`. Again #5087 should give some template to think about this https://github.com/pymc-devs/pymc/blob/8f3636daf7d9946f6eca4717f3bb0c6d77d9c6e9/pymc/distributions/continuous.py#L546-L550
  3.1 In the case where you have to manually broadcast the parameters with each other it's important to add test conditions that would fail if you were not to do that. A straightforward way to do this is to make the used parameter a scalar, the unused one(s) a vector (one at a time) and size `None`
 
4. Just to keep things uniform*ish*, please add the `get_moment` immediately below the `dist` classmethod and before `logp` or `logcdf` methods.
5. New tests have to be added in `test_distributions_moments.py`. Make sure to test different combinations of size and broadcasting to cover the cases mentioned in point 3.
[...]

PR #5087, opened 2021-10-18
Add tests for distributions moments

michaelosthege:
Just checking the CI already.

Do we want to systematically test the moments?
Any ideas how to do that without lots of copy-paste?

Related to #5078

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -85,6 +85,7 @@ from pymc.distributions.dist_math import (
     zvalue,
 )
 from pymc.distributions.distribution import Continuous
+from pymc.distributions.shape_utils import rv_size_is_none
 from pymc.math import logdiffexp, logit
 from pymc.util import UNSET
 
@@ -290,6 +291,13 @@ class Uniform(BoundedContinuous):
         upper = at.as_tensor_variable(floatX(upper))
         return super().dist([lower, upper], **kwargs)
 
+    def get_moment(rv, size, lower, upper):
+        lower, upper = at.broadcast_arrays(lower, upper)
+        moment = (lower + upper) / 2
+        if not rv_size_is_none(size):
+            moment = at.full(size, moment)
+        return moment
+
     def logcdf(value, lower, upper):
         """
         Compute the log of the cumulative distribution function for Uniform distribution
@@ -314,11 +322,6 @@ class Uniform(BoundedContinuous):
                 0,
             ),
         )
-
-    def get_moment(value, size, lower, upper):
-        lower = at.full(size, lower, dtype=aesara.config.floatX)
-        upper = at.full(size, upper, dtype=aesara.config.floatX)
-        return (lower + upper) / 2
 
 
 class FlatRV(RandomVariable):
@@ -353,8 +356,8 @@ class Flat(Continuous):
         res = super().dist([], size=size, **kwargs)
         return res
 
-    def get_moment(rv, size, *rv_inputs):
-        return at.zeros(size, dtype=aesara.config.floatX)
+    def get_moment(rv, size):
+        return at.zeros(size)
 
     def logp(value):
         """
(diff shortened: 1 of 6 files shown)


ricardoV94:
> Do we want to systematically test the moments?

Definitely

> Any ideas how to do that without lots of copy-paste?

None yet. We should probably test that it respects the RV size, while also working with broadcasted params.

pymc/distributions/continuous.py
```diff
@@ -789,6 +780,20 @@ def logp(
             bounds.append(lower <= upper)
         return bound(logp, *bounds)
 
+    def get_moment(value_var, size, mu, sigma, lower, upper):
+        initval = None
+        has_lower = at.isinf(lower).eval()
```
ricardoV94: This can probably work with a nested switch that does not require knowing lower/upper at call time.

Otherwise you can check if it's a constant and retrieve the value. If it's not a constant the `eval` will fail with missing input

pymc/tests/test_distributions_moments.py
```diff
@@ -0,0 +1,140 @@
...
+        (np.arange(5), (2, 5), np.full((2, 5), np.arange(5))),
+    ],
+)
+def test_halfnormal_moment(sigma, size, expected):
+    with Model() as model:
+        HalfNormal("x", sigma=sigma, size=size)
+    assert_moment_is_expected(model, expected)
+
+
+@pytest.mark.skip(reason="aeppl interval transform fails when both edges are None")
```
ricardoV94, line 116: https://github.com/aesara-devs/aeppl/pull/82

pymc/distributions/distribution.py
```diff
@@ -371,7 +370,7 @@ def get_moment(rv: TensorVariable) -> TensorVariable:
     for which the value is to be derived.
     """
     size = rv.owner.inputs[1]
-    return _get_moment(rv.owner.op, rv, size, *rv.owner.inputs[3:])
+    return _get_moment(rv.owner.op, rv, size, *rv.owner.inputs[3:]).astype(aesara.config.floatX)
```
michaelosthege: Good placement!

pymc/distributions/shape_utils.py
```diff
@@ -674,3 +675,8 @@ def maybe_resize(
         )
 
     return rv_out
+
+
+def rv_size_is_none(size: Variable) -> bool:
+    """Check wether an rv size is None (ie., at.Constant([]))"""
+    return isinstance(size, Constant) and size.data.size == 0
```
michaelosthege, line 682: It's perfectly fine to do `pm.get_moment(..., size=())`, right? Or even with `size=[]`?
Maybe this utility function should account for `size` being `()`, `[]` or `None` too?

ricardoV94: No, the size comes from the RV and is always converted to an empty string when it's None IIRC

michaelosthege (commented):
I can't approve it, even though @ricardoV94 did 98 % of the changes.

pymc/distributions/distribution.py
ricardoV94: Actually just realized it should convert to the RVs dtype which maybe be int64

twiecki (approved):

twiecki:
Do we still need more moments?

ricardoV94:
> Do we still need more moments?

Yes, see #5078

pymc/distributions/continuous.py
```diff
@@ -828,6 +850,12 @@ def dist(cls, sigma=None, tau=None, sd=None, *args, **kwargs):
 
         return super().dist([0.0, sigma], **kwargs)
 
+    def get_moment(rv, size, loc, sigma):
+        moment = loc + sigma
```
patel-zeel, line 854: I have a doubt here.

Current implementation: loc + sigma
[Wikipedia version](https://en.wikipedia.org/wiki/Half-normal_distribution): loc + \sqrt{\frac{2}{\pi}}sigma

ricardoV94: I think you are right, we should probably be using that. Do you want to open a PR to fix it?

patel-zeel: Sure, happily :)

ricardoV94: A bit more involved, we should probably also use the truncated normal mean as the moment. We might even be able to simplify the switch statement, if the normal logcdf behaves well with +- infinity values: https://en.wikipedia.org/wiki/Truncated_normal_distribution

PR #5091, opened 2021-10-21
BART: add partial dependence plots and individual conditional expectation plots

aloctavodia:
This add functions to compute partial dependence plots and individual conditional expectation plots. It also move the in-sample and out-of-sample function predictions to the function `predict`, also part of `bart.utils`. After this PR is merged I will add an example of  how to use these function and BART in general. 

This is an example of a partial dependence plot, where BART is used to predict the number of rented bikes given 4 covariates.

index


twiecki:
Can you post an example plot of what it looks like?

Also, have you considered spawning this out into a separate package `pymc-bart`?

aloctavodia:
I have not considered a separated package. But not opposed to that. Most of the logic for BART is now on its own module. The only exception is the code related to the automatic assignment of PGBART sampler to BART variables.

ricardoV94:
Maybe we can harbor these more specialized modules until they grow stable/popular and then spin them off to their own packages?

aloctavodia:
Maybe that's a good strategy. One spin off library could be a more machine-learning/Bayesian non-paramateric kind of library and not just a BART library.

Additionally one interesting feature of our BART implementation is precisely that is PPL friendly.  There is still a lot of room for improvement and exploration here. Both at the implementation level and at the level of the  types of model we can build and their uses.

aloctavodia:
@twiecki @ricardoV94 can we merge this?

ricardoV94:
@aloctavodia do you want to add a release note about the new features?

ricardoV94 (approved):

PR #5092, opened 2021-10-21
Refactor sampling_jax

ricardoV94:
sampling_jax is no longer wrapped in a symbolic Op, making sampling more performant. It was insanely slow for the newer versions of JAX, perhaps due to omnistaging not being optional anymore, but I couldn't quite figure it out.

SharedVariables are now replaced by constants, and a ValueError is raised if any contains a default_update (this was not handled before either AFAIK).

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -15,12 +15,10 @@ import numpy as np
 import pandas as pd
 
 from aesara.compile import SharedVariable
-from aesara.graph.basic import Apply, Constant, clone, graph_inputs
+from aesara.graph.basic import clone_replace, graph_inputs
 from aesara.graph.fg import FunctionGraph
-from aesara.graph.op import Op
 from aesara.graph.opt import MergeOptimizer
 from aesara.link.jax.dispatch import jax_funcify
-from aesara.tensor.type import TensorType
 
 from pymc import modelcontext
 from pymc.aesaraf import compile_rv_inplace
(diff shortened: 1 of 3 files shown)


pymc/sampling_jax.py
```diff
@@ -165,72 +59,101 @@ def sample_numpyro_nuts(
...
-        draws=draws,
-        tune=tune,
-        chains=chains,
-        seed=map_seed,
-        progress_bar=progress_bar,
-    )(*init_state_batched_at)
+    logpt = replace_shared_variables([model.logpt])[0]
+    logpt_fgraph = FunctionGraph(outputs=[logpt], clone=False)
+    MergeOptimizer().optimize(logpt_fgraph)
+    logp_fn = jax_funcify(logpt_fgraph)
```
twiecki, line 76: neat!

twiecki (approved):

ricardoV94:
@michaelosthege are we ignoring the jaxtests in the codecov? The patch of 0.00% diff hit cannot be correct

ricardoV94:
Also we don't need to install tensorflow_probability anymore right, @twiecki ?

https://github.com/pymc-devs/pymc/blob/8ae296bdf3ee58850824c9997293803d2df958ec/.github/workflows/jaxtests.yml#L61

ricardoV94:
Also we don't need to install tensorflow_probability anymore right, @twiecki ?

https://github.com/pymc-devs/pymc/blob/8ae296bdf3ee58850824c9997293803d2df958ec/.github/workflows/jaxtests.yml#L61

twiecki:
> Also we don't need to install tensorflow_probability anymore right, @twiecki ?

Correct.

twiecki:
Line in the release-notes, then we can merge.

ricardoV94:
> Line in the release-notes, then we can merge.

We haven't changed the external behavior (other than the logging messages)

michaelosthege:
> @michaelosthege are we ignoring the jaxtests in the codecov? The patch of 0.00% diff hit cannot be correct

The codecov upload step is part of the ubuntu job within the pytest workflow. The jax tests are not running in that configuration.

Ideally coverage should be combined across all workflows floatX and OS settings, and that's probably also possible. If this was GitLab CI I might have some ideas, but this GitHub Actions YAML with "actions" confuses me a lot.

PR #5094, opened 2021-10-21
Reintroduce test_types.py

### LINKED ISSUE #4828: Reenable test_types.py and make it run faster

ricardoV94:
Closes #4828


michaelosthege (approved):
maybe run it on all three operating systems?

twiecki (approved):

PR #5096, opened 2021-10-22
Optimize sample_numpyro_nuts graph more thoroughly

ricardoV94:
This makes sure the graph we send to NumPyro is properly optimized. Follow up to #5092 

Still have to add tests...

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -4,6 +4,11 @@ import re
 import sys
 import warnings
 
+from typing import Callable, List
+
+from aesara.graph import optimize_graph
+from aesara.tensor import TensorVariable
+
 xla_flags = os.getenv("XLA_FLAGS", "").lstrip("--")
 xla_flags = re.sub(r"xla_force_host_platform_device_count=.+\s", "", xla_flags).split()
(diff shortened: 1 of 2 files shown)


pymc/sampling_jax.py
```diff
@@ -49,6 +50,45 @@ def replace_shared_variables(graph):
...
+    """Compile model.logpt into an optimized jax function"""
+
+    logpt = replace_shared_variables([model.logpt])[0]
+
+    logpt_fgraph = FunctionGraph(outputs=[logpt], clone=False)
+    # We need to add a Supervisor to the fgraph to be able to run the
+    # JAX sequential optimizer without warnings. We made sure there
+    # are no mutable input variables, so we only need to check for
+    # "destroyers"
+    logpt_fgraph.attach_feature(
```
twiecki: Looks scary, something we should change in aesara?

ricardoV94: That's the reason why this is still a draft. 

@brandonwillard gave me another suggestion, but I didn't look at it yet.

ricardoV94: Okay this is now done more to the point!

twiecki (approved):

PR #5098, opened 2021-10-24
Pass kwargs from `pm.Data` to `aesara.shared`

### LINKED ISSUE #4967: `pm.Data` breaks broadcasting of different shapes

Typically when you add shapes that are mismatched, broadcasting is expected.

```pycon
# Numpy example of expected broadcasting
>>> np.array([[1, 2]]) + np.array([[1], [2]])
array([[2, 3],
       [3, 4]])
```

This works as expected when inputting a numpy array, but when fed through via `pm.Data`, the broadcasting does not work as expected, and raises an exception on the mismatched shapes.

```python
"""This works!"""
import numpy as np
import pymc3 as pm

raw_data = np.array([[1], [2], [3]])

with pm.Model() as model:
    val = pm.Normal("val", mu=0, sd=10, shape=(1, 2))
    pm.Deterministic("add", var=val + raw_data)

```

```python
"""This does not work!"""
import numpy as np
import pymc3 as pm

raw_data = np.array([[1], [2], [3]])

with pm.Model() as model:
    val = pm.Normal("val", mu=0, sd=10, shape=(1, 2))
    data = pm.Data("data", raw_data)  # New line
    pm.Deterministic("add", var=val + data)

```

```pytb
<ipython-input-231-ccd5d08fd0d4> in <module>
      4     val = pm.Normal("val", mu=0, sd=10, shape=(1, 2))
      5     data = pm.Data("data", raw_data)
----> 6     pm.Deterministic("add", var=val + data)

~/.cache/pypoetry/virtualenvs/time-series-demand-5LaNRWUQ-py3.9/lib/python3.9/site-packages/theano/tensor/var.py in __add__(self, other)
     99     def __add__(self, other):
    100         try:
--> 101             return theano.tensor.basic.add(self, other)
    102         # We should catch the minimum number of exception here.
    103         # Otherwise this will convert error when Theano flags

~/.cache/pypoetry/virtualenvs/time-series-demand-5LaNRWUQ-py3.9/lib/python3.9/site-packages/theano/graph/op.py in __call__(self, *inputs, **kwargs)
    251 
    252         if config.compute_test_value != "off":
--> 253             compute_test_value(node)
    254 
    255         if self.default_output is not None:

~/.cache/pypoetry/virtualenvs/time-series-demand-5LaNRWUQ-py3.9/lib/python3.9/site-packages/theano/graph/op.py in compute_test_value(node)
    128     thunk.outputs = [storage_map[v] for v in node.outputs]
    129 
--> 130     required = thunk()
    131     assert not required  # We provided all inputs
    132 

~/.cache/pypoetry/virtualenvs/time-series-demand-5LaNRWUQ-py3.9/lib/python3.9/site-packages/theano/graph/op.py in rval()
    604 
    605         def rval():
--> 606             thunk()
    607             for o in node.outputs:
    608                 compute_map[o][0] = True

~/.cache/pypoetry/virtualenvs/time-series-demand-5LaNRWUQ-py3.9/lib/python3.9/site-packages/theano/link/c/basic.py in __call__(self)
   1769                 print(self.error_storage, file=sys.stderr)
   1770                 raise
-> 1771             raise exc_value.with_traceback(exc_trace)
   1772 
   1773 

ValueError: Input dimension mis-match. (input[0].shape[1] = 2, input[1].shape[1] = 1)
```

</details>

* PyMC3 Version: 3.11.2
* Aesara/Theano Version: 1.1.2
* Python Version: 3.9.5
* Operating system: CentOS Stream 8
* How did you install PyMC3: (conda/pip) pip via poetry

paw-lu:
Closes #4967

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -482,6 +482,8 @@ class Data:
     export_index_as_coords: bool, optional, default=False
         If True, the `Data` container will try to infer what the coordinates should be
         if there is an index in `value`.
+    **kwargs: dict, optional
+        Extra arguments passed to :func:`aesara.shared`.
 
     Examples
     --------
@@ -512,7 +514,15 @@ class Data:
     https://docs.pymc.io/notebooks/data_container.html
     """
 
-    def __new__(self, name, value, *, dims=None, export_index_as_coords=False):
+    def __new__(
+        self,
+        name,
+        value,
+        *,
+        dims=None,
+        export_index_as_coords=False,
+        **kwargs,
+    ):
         if isinstance(value, list):
             value = np.array(value)
 
@@ -528,7 +538,7 @@ class Data:
 
         # `pandas_to_array` takes care of parameter `value` and
         # transforms it to something digestible for pymc
-        shared_object = aesara.shared(pandas_to_array(value), name)
+        shared_object = aesara.shared(pandas_to_array(value), name, **kwargs)
 
         if isinstance(dims, str):
             dims = (dims,)
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -464,8 +464,8 @@ def align_minibatches(batches=None):
 
 
 class Data:
-    """Data container class that wraps the Aesara ``SharedVariable`` class
-    and lets the model be aware of its inputs and outputs.
+    """Data container class that wraps :func:`aesara.shared` and lets
+    the model be aware of its inputs and outputs.
 
     Parameters
     ----------
@@ -478,7 +478,7 @@ class Data:
         random variables). Use this when `value` is a pandas Series or DataFrame. The
         `dims` will then be the name of the Series / DataFrame's columns. See ArviZ
         documentation for more information about dimensions and coordinates:
-        https://arviz-devs.github.io/arviz/notebooks/Introduction.html
+        :ref:`arviz:quickstart`.
     export_index_as_coords: bool, optional, default=False
         If True, the `Data` container will try to infer what the coordinates should be
         if there is an index in `value`.
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -88,6 +88,7 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
 - New features for BART:
   -  Added linear response, increased number of trees fitted per step [5044](https://github.com/pymc-devs/pymc3/pull/5044).
   -  Added partial dependence plots and individual conditional expectation plots [5091](https://github.com/pymc-devs/pymc3/pull/5091).
+- `pm.Data` now passes additional kwargs to `aesara.shared`. [#5098](https://github.com/pymc-devs/pymc/pull/5098)
 - ...
 
 
--- a/pymc/tests/test_data_container.py
+++ b/pymc/tests/test_data_container.py
@@ -366,6 +366,19 @@ class TestData(SeededTest):
         assert "columns" in pmodel.coords
         assert pmodel.RV_dims == {"observations": ("rows", "columns")}
 
+    def test_data_kwargs(self):
+        strict_value = True
+        allow_downcast_value = False
+        with pm.Model():
+            data = pm.Data(
+                "data",
+                value=[[1.0], [2.0], [3.0]],
+                strict=strict_value,
+                allow_downcast=allow_downcast_value,
+            )
+        assert data.container.strict is strict_value
+        assert data.container.allow_downcast is allow_downcast_value
+
 
 def test_data_naming():
     """


pymc/data.py
```diff
@@ -482,6 +482,7 @@ class Data:
     export_index_as_coords: bool, optional, default=False
         If True, the `Data` container will try to infer what the coordinates should be
         if there is an index in `value`.
+    kwargs: kwargs passed to Aesara ``SharedVariable``.
```
michaelosthege:
```suggestion
    kwargs:
        Forwarded to Aesara ``SharedVariable``.
```

michaelosthege (approved):
Looks good!

Also note https://github.com/pymc-devs/pymc/issues/5105 which is slightly related.

pymc/data.py
paw-lu: I have zero problems with this suggestion, but isn't it out of line with other kwargs docstrings in the repo, specifically the newline?

- https://github.com/pymc-devs/pymc/blob/9e7e8aac516b3c9e357cc0a1cefc5286acc0d969/pymc/step_methods/hmc/base_hmc.py#L86
- https://github.com/pymc-devs/pymc/blob/9e7e8aac516b3c9e357cc0a1cefc5286acc0d969/pymc/step_methods/hmc/nuts.py#L162
- https://github.com/pymc-devs/pymc/blob/9e7e8aac516b3c9e357cc0a1cefc5286acc0d969/pymc/step_methods/hmc/hmc.py#L110

michaelosthege: @OriolAbril can you clarify how docstrings for **kwargs should be formatted?

OriolAbril: Docstrings should follow the numpydoc style: https://numpydoc.readthedocs.io/en/latest/format.html#sections (see the last example of point 4 about parameters, copied below too). Ignore what is currently done as many docstrings do not follow numpmydoc for a number of reasons. 

```
**kwargs : dict, optional
    Extra arguments to `metric`: refer to each metric documentation for a
    list of all possible arguments.
```

Moreover, in this case, as the kwargs are passed down to another pymc object, that object should be written as a sphinx cross-reference so that it is rendered as a clickable link pointing to the right place. I believe in this case that would be:

```
Passed to :class:`pymc.BaseHMC`
```

but please do check the doc preview from the readthedocs check of the PR

twiecki: @paw-lu Are you going to address this so we can merge? Also needs a rebase.

paw-lu: Yes!
I'll attempt to make time for this tomorrow @twiecki.

paw-lu: Rebased and fixed.

Some deviations:

- No space between arguemnt name and colon to remain consistent with rest of docstring
- Actual object is an Aesara object—not PyMC—so Sphinx cross-reference not used.

paw-lu: Screenshot of rendered Sphinx docs for convenience.

Screen Shot 2021-11-04 at 6 47 07 PM

pymc/data.py
```diff
@@ -482,6 +482,8 @@ class Data:
     export_index_as_coords: bool, optional, default=False
```
OriolAbril, line 482: fix the link right above, it should not be html but `` :ref:`arviz:quickstart` ``

pymc/data.py
```diff
@@ -482,6 +482,8 @@ class Data:
...
         If True, the `Data` container will try to infer what the coordinates should be
         if there is an index in `value`.
+    **kwargs: dict, optional
+        Extra arguments passed to Aesara ``SharedVariable``.
```
OriolAbril: `SharedVariable` should be a clickable link so users can easily find what values are accepted and what are not without needing to first get to aesara docs, then finding sharedvariable within it.

Taking a look at the code, it should be `` :func:`aesara.shared` ``

OriolAbril (changes requested):

pymc/data.py
paw-lu: Oh I totally forgot about [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html)! Good catch. I'll also apply this change to the top level `SharedVariable` reference. Apprecaited

OriolAbril: No problem, it is particularly tricky because you can't really trust the rest of the docstring to be following numpydoc. Hopefully we'll get there little by little

pymc/data.py
paw-lu: Addressed via 56c45fa

pymc/data.py
paw-lu: - Rebased via 509a6e3 to address args reference
- Added 56c45fa for top level doc reference

OriolAbril (approved):
approved from a doc side

twiecki (approved):

twiecki:
Thanks @paw-lu!

PR #5106, opened 2021-10-27
Fix deprecation warning for save/load_trace

aloctavodia:
`FutureWarnings` are intended for users and `DeprecationWarnings` for developers. I am not sure but I think by default jupyter notebook will show the former but not the later. Just in case I added an additional message in case a user attempts to use `save_trace` for saving InferenceData.


canyon289 (approved):

twiecki:
How long has this been there? Isn't it time we just remove them?

aloctavodia:
Not sure, but highly likely that users have not seen the message.

twiecki:
Seems like it was added 2 years ago: https://github.com/pymc-devs/pymc/pull/3902

twiecki:
https://github.com/pymc-devs/pymc/issues/5108

PR #5109, opened 2021-10-28
replace deprecationwarning with futurewarning

aloctavodia:
`DeprecationWarning` is intended for other developers and `FutureWarning` for users. When in doubt is probably safer to go with `FutureWarning`.


twiecki:
Again, I think most of these should be deleted, let's open an issue for those.

PR #5110, opened 2021-10-28
Fix minor typos

canyon289:


canyon289:
Noticing theres other small fixes needed here. Marking WIP till I can get to them all later

canyon289:
Alright were good now

michaelosthege:
I must say I'm not a fan of not mentioning version numbers and just going with the "PyMC" vs. "PyMC3" distinction.

A sentence such as
> This is effectively the PyMC replacement for `Distribution.random`.

could easily confuse someone who is not aware that there once was a "PyMC3" and that the lack of the 3 indicates that `pymc >=4` is meant.

Maybe we can instead try to mention `v3` explicitly like this?

> This is effectively the replacement of the old `Distribution.random` API from v3.x.

aloctavodia:
I agree with @michaelosthege we should not make a distinction between PyMC3 and PyMC, there is only one project/software/package/tool and that is PyMC. If we want to contrast features from different versions we should explicitly use the version numbers like 3.x vs 4.x or <4.0, etc

canyon289:
Ok! Will update

twiecki (approved):

ricardoV94:
@canyon289 Can we merge this or are there any other changes you had in mind?

canyon289:
Ah I see we merged this, was going to update based on discussion in our lab meeting. Noting that within our community we still are unclear about how to refer to the distinctions of PyMC

Between subsets of us we still use

* PyMC vs PyMC3
* PyMC 4.x vs PyMC 3.x (and drop PyMC3 altogether)
* PyMC V4 vs PyMC V3 (and drop PyMC3 altogether)

PR #5129, opened 2021-11-03
Remove deprecated functions

### LINKED ISSUE #5124: Remove deprecated functions
Throughout the code-base there are quite a few things we deprecated a while ago and should remove for 4.0. Here are a few: https://github.com/pymc-devs/pymc/pull/5109

But in general, searching for `DeprecationWarning` or `FutureWarning` should bring those up.

farhanreynaldo:
This PR closes https://github.com/pymc-devs/pymc/issues/5124, to remove deprecated functions.

Should we also remove an `if statement` with a `FutureWarning` in it? For example this line:

```python
def invlogit(x, eps=None):
    """The inverse of the logit function, 1 / (1 + exp(-x))."""
    if eps is not None:
        warnings.warn(
            "pymc.math.invlogit no longer supports the ``eps`` argument and it will be ignored.",
            FutureWarning,
            stacklevel=2,
        )
    return at.sigmoid(x)
```

If that's the case, I think we need to remove the related argument(s).

### DIFF
--- a/pymc/backends/ndarray.py
+++ b/pymc/backends/ndarray.py
@@ -16,10 +16,7 @@
 
 Store sampling values in memory as a NumPy array.
 """
-import json
-import os
-import shutil
-import warnings
+
 
 from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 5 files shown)


farhanreynaldo:
I am also not quite sure whether to drop a class, like the `SerializeNDArray` and function like `trace_to_dataframe`.

ricardoV94:
@farhanreynaldo You have to be careful to make sure we are deprecating things that are actually old. Most of the features that you are removing in this PR were only added in V4, which was not yet released for users. As a rule of thumb I guess we should only remove things that have been marked as to-be deprecated prior to v3.11?

ricardoV94:
This shows the repository state when version 3.10 was released: https://github.com/pymc-devs/pymc/tree/v3.10.0

ricardoV94:
~~Unrelated to this PR, that invlogit should not be a FutureWarning as the behavior is already changed...~~ This seems to be the correct convention according to https://docs.python.org/3/library/warnings.html

farhanreynaldo:
> @farhanreynaldo You have to be careful to make sure we are deprecating things that are actually old. Most of the features that you are removing in this PR were only added in V4, which was not yet released for users. As a rule of thumb I guess we should only remove things that have been marked as to-be deprecated prior to v3.11?

Are the `marked as to-be deprecated prior to v3.11` indicated by `DeprecationWarning` in v3.10? After a quick look at the link you attached, the only function applicable with the criteria is `dataset_to_point_dict`, and the rest should be reverted. It seems like both `SerializeNDArray` and `trace_to_dataframe` fit the criteria.

ricardoV94:
> Are the `marked as to-be deprecated prior to v3.11` indicated by `DeprecationWarning` in v3.10? 

Yes they should be. Either with DeprecationWarning or FutureWarning. Otherwise it means our warnings have not been in the codebase long enough to justify removing them already. 

We can discuss if 3.10 was released too long ago, and whether we might want to check against 3.11.x, but that's not up to me :)

twiecki:
I think 3.11 is fine.

farhanreynaldo:
I have reverted the previous changes and removed deprecated functions based on `v3.11.0`. I also found that `plot_posterior_predictive_glm` should be deprecated, but are we going to delete the file containing said function and related tests altogether?

Since I also removed the related tests, I tried to run the following line but an error occurred: ```pytest --cov=pymc pymc/tests```

```
ImportError while loading conftest '/Users/farhanreynaldo/Documents/pymc3/pymc/tests/conftest.py'.
pymc/__init__.py:61: in <module>
    _check_install_compatibilitites()
pymc/__init__.py:52: in _check_install_compatibilitites
    + f" the legacy version PyMC3 {pymc3.__version__} installed."
E   AttributeError: module 'pymc3' has no attribute '__version__'```

ricardoV94:
> I have reverted the previous changes and removed deprecated functions based on `v3.11.0`. I also found that `plot_posterior_predictive_glm` should be deprecated, but are we going to delete the file containing said function and related tests altogether?
> 
> Since I also removed the related tests, I tried to run the following line but an error occurred: `pytest --cov=pymc pymc/tests`
> 
> ```
> ImportError while loading conftest '/Users/farhanreynaldo/Documents/pymc3/pymc/tests/conftest.py'.
> pymc/__init__.py:61: in <module>
>     _check_install_compatibilitites()
> pymc/__init__.py:52: in _check_install_compatibilitites
>     + f" the legacy version PyMC3 {pymc3.__version__} installed."
> E   AttributeError: module 'pymc3' has no attribute '__version__'```
> ```

That means that you probably have an empty pymc3 folder somewhere in the packages directory of your environment, and instead of raising some sort of ImportError Python imports pymc3 as an empty NameSpace.

twiecki:
@farhanreynaldo A few changes show up here that shouldn't, so you probably messed up a rebase. One approach is to start from a clean `main` and cherry-pick your commits.

farhanreynaldo:
> @farhanreynaldo A few changes show up here that shouldn't, so you probably messed up a rebase. One approach is to start from a clean `main` and cherry-pick your commits.

I am sorry I messed up during the rebase. I did that because there is 1 error test that I think is unrelated to this PR. 

To start a clean slate as you suggested, should I create a new branch from `main` (e.g. `remove-deprecated-functions-clean`, cherry-pick the commits, and `git push origin remove-deprecated-functions`? I want to make sure that I do the steps correctly.

twiecki:
Yes but you will need to force push (or name the new branch differently).

twiecki (approved):

PR #5131, opened 2021-11-03
Use Random Generator in SMC instead of global seeding

### LINKED ISSUE #4938: Do not use global random seeding in SMC sampling
We are creating new random seeds in `sample_smc` here:

https://github.com/pymc-devs/pymc3/blob/2257c34ccf59a5e500c7faf1ca3d480244e7848f/pymc3/smc/sample_smc.py#L207-L216

And then setting a global seed state in the `SMC` kernel here:
https://github.com/pymc-devs/pymc3/blob/2257c34ccf59a5e500c7faf1ca3d480244e7848f/pymc3/smc/smc.py#L155-L156

Instead we should propagate the rng RandomStream/Generator to the methods that take random draws and use it directly

ricardoV94:
Closes #4938 

Other changes:
1. Use model.recompute_initial_point instead of deprecated model.initial_point.
2. Update Metropolis Proposal classes to accept a Numpy Random Generator in `__call__`

### DIFF
--- a/pymc/smc/sample_smc.py
+++ b/pymc/smc/sample_smc.py
@@ -42,7 +42,7 @@ def sample_smc(
     *,
     start=None,
     model=None,
-    random_seed=-1,
+    random_seed=None,
     chains=None,
     cores=None,
     compute_convergence_checks=True,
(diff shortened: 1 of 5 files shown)


twiecki (approved):

aloctavodia (approved):
LGTM!

PR #5132, opened 2021-11-03
Update computation of proposal logp ratio in SMC IMH kernel

ricardoV94:
In a conversation with @aloctavodia we realized that the computation of the acceptance ratio was slightly off for the IMH kernel. @aloctavodia wrote down the equations on paper so we should be pretty confident this is correct :D

### DIFF
--- a/pymc/smc/smc.py
+++ b/pymc/smc/smc.py
@@ -379,31 +379,33 @@ class IMH(SMC_KERNEL):
     def mutate(self):
         """Independent Metropolis-Hastings perturbation."""
         ac_ = np.empty((self.n_steps, self.draws))
-
-        cov = self.proposal_dist.cov
         log_R = np.log(self.rng.random((self.n_steps, self.draws)))
+
+        # The proposal is independent from the current point.
+        # We have to take that into account to compute the Metropolis-Hastings acceptance
+        # We first compute the logp of proposing a transition to the current points.
+        # This variable is updated at the end of the loop with the entries from the accepted
+        # transitions, which is equivalent to recomputing it in every iteration of the loop.
+        backward_logp = self.proposal_dist.logpdf(self.tempered_posterior)
         for n_step in range(self.n_steps):
-            # The proposal is independent from the current point.
-            # We have to take that into account to compute the Metropolis-Hastings acceptance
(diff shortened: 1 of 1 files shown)


pymc/smc/smc.py
```diff
@@ -379,31 +379,31 @@ def tune(self):
...
         """Independent Metropolis-Hastings perturbation."""
         ac_ = np.empty((self.n_steps, self.draws))
-
-        cov = self.proposal_dist.cov
         log_R = np.log(self.rng.random((self.n_steps, self.draws)))
+
+        # The proposal is independent from the current point.
+        # We have to take that into account to compute the Metropolis-Hastings acceptance
+        # We first compute the logp of proposing a transition to the current points.
+        backward_logp = self.proposal_dist.logpdf(self.tempered_posterior)
```
aloctavodia, line 389: This should be inside the loop, self.tempered_posterior is updated at each n_step

ricardoV94: I update this variable inplace at the end of the loop, with the logp of the accepted transitions

aloctavodia: Ohh I see now!

aloctavodia (approved):

pymc/smc/smc.py
```diff
@@ -379,31 +379,31 @@ def tune(self):
...
-                (proposal_logp + backward) - (self.tempered_posterior_logp + forward)
+                (proposal_logp + backward_logp) - (self.tempered_posterior_logp + forward_logp)
             )
+
             ac_[n_step] = accepted
             self.tempered_posterior[accepted] = proposal[accepted]
             self.tempered_posterior_logp[accepted] = proposal_logp[accepted]
             self.prior_logp[accepted] = pl[accepted]
             self.likelihood_logp[accepted] = ll[accepted]
+            backward_logp[accepted] = forward_logp[accepted]
```
junpenglao, line 408: So this is the same as running `backward_logp = self.proposal_dist.logpdf(self.tempered_posterior)` here again right?

ricardoV94: Yes, just avoiding the recomputation because we already have all the values in memory

ricardoV94: Added a comment at the top clarifying this, as it was brought up twice already :)

PR #5133, opened 2021-11-03
Do not compare arrays with strings in make_initial_point_expression

ricardoV94:
This gets rid of numpy FutureWarnings when we set the initval to an array:

```python
import numpy as np
import pymc as pm

with pm.Model() as m:
  x = pm.Normal('x', size=10, initval=np.arange(10))
m.recompute_initial_point()
```

### DIFF
--- a/pymc/initial_point.py
+++ b/pymc/initial_point.py
@@ -269,10 +269,15 @@ def make_initial_point_expression(
         if strategy is None:
             strategy = default_strategy
 
-        if strategy == "moment":
-            value = get_moment(variable)
-        elif strategy == "prior":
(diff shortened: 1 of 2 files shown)


michaelosthege (approved):
Test failures are unrelated (flaky tests?).

I remember that you (?) mentioned this problem somewhere, but it looks like there's no corresponding issue?

Anyway, your changes fix it so that's good.

ricardoV94:
The failures are coming from #5131 because the Metropolis proposals are no longer following a global seeding, but several tests in `test_step` are based on this assumption.

Issue #4904, opened 2021-08-05
Scalar ODEs on Windows failing because of a casting & broadcastability issue that was fixed only for Linux


michaelosthege:
## Description of your problem
Three ODE tests - all testing with a scalar ODE - still run into broadcastability problems.
The related Aesara ticket https://github.com/aesara-devs/aesara/issues/390 was closed with a fix that did fix the aforementioned tests for non-Windows OSes.

But the tests kept failing for Windows, so apparently the issue still persists under some rare circumstances.

Also see https://github.com/pymc-devs/pymc3/pull/4902 which edited the `@pytest.mark.xfail` conditions accordingly.

## Versions and main components

* PyMC3 Version: `main`
* Aesara Version: `2.1.3`
* Operating system: Windows

ricardoV94:
Figured out the source of difference in behavior. Aesara constant integer lists have dtype int32 in Windows but dtype int64 in Linux. 

```python
import aesara.tensor as at

# Changes from int64 in Linux to int32 in Windows
print(at.constant([1]).dtype)
```

This leads to an extra casting in the size input when creating RVs in Windows compared to Linux.
```python
# Will be a constant in Linux but a SpecifyShape of a cast in Windows
print(at.random.normal(size=[1]).owner.inputs[1])
```

Which might lead to further issues down the road due to things like https://github.com/aesara-devs/aesara/issues/692

PR #5140, opened 2021-11-05
Change initial_point default strategy from "prior" to "moment"

### LINKED ISSUE #5009: Change default initial_point method to "moment"
We set "prior" as default in #4983, but we actually want "moment". This makes the PR easier, because most distributions don't define `get_moment` yet, but as soon as that becomes available we should switch the default over.

ricardoV94:
This PR adds a check that we have moments for all distributions and changes the default initial point strategy from `"prior"` to `"moment"`

Also added a check in the default test moments that the logp of the provided moment is finite.

Closes #5009 
Related to #5078 

We also need to decide whether we default to a draw from the prior if a distribution does not have a `get_moment` implemented or if we just fail in that case (which is what is happening now). It would be trivial to allow this with a try/except block

### DIFF
--- a/pymc/tests/test_distributions_moments.py
+++ b/pymc/tests/test_distributions_moments.py
@@ -8,7 +8,6 @@ from scipy import special
 
 import pymc as pm
 
-from pymc import Simulator
 from pymc.distributions import (
     AsymmetricLaplace,
     Bernoulli,
@@ -50,6 +49,7 @@ from pymc.distributions import (
     Poisson,
     PolyaGamma,
     Rice,
+    Simulator,
     SkewNormal,
     StudentT,
     Triangular,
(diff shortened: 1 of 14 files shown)


ricardoV94:
Getting some weird parallel sampling [test failures](https://github.com/pymc-devs/pymc/runs/4263439376?check_suite_focus=true) only on Windows. Perhaps related to https://github.com/pymc-devs/pymc/issues/4904?

ricardoV94:
I am inclined to defaulting back to a prior sample if a distribution does not have a moment, perhaps with a UserWarning like

> Moment not defined for distribution X, defaulting to a draw from the prior. This can lead to difficulties during tuning. You can manually define an initval or implement a `get_moment` dispatched function for this distribution.

This would allow users to more gradually implement their own distributions. We still have a check added in this PR that all our standard distributions have moments implemented.

michaelosthege:
I like the fallback plus warning idea. Maybe include a link to the relevant docs section on implementing custom distributions?

ricardoV94:
We have very few moments missing, and it was consensus that this should no longer block the beta release. 

I have added the known missing moments to `test_all_distributions_have_moments`. This test will fail whenever those moments get implemented, making sure our coverage stays up to date.

ricardoV94:
Still have those surprising failing tests on Windows alone...

michaelosthege:
> Still have those surprising failing tests on Windows alone...

To me it looks like the same problem as #4904 - broadcasts are not correctly propagated through Cast operations.

ricardoV94:
Windows issues are related to a mix of:

1. https://github.com/aesara-devs/aesara/issues/691,
2. https://github.com/aesara-devs/aesara/issues/692

ricardoV94:
Tests are passing with the latest Aesara version

pymc/tests/test_distributions_moments.py
```diff
@@ -62,13 +62,70 @@
...
+    # Assert that they do not have moments yet
+    assert missing_moments.intersection(not_implemented) == not_implemented
+
+    missing_moments -= not_implemented
+
+    if missing_moments:
+        raise NotImplementedError(
+            "The following distributions do not have a moment method:\n"
+            + "\n".join(sorted(map(str, list(missing_moments))))
+        )
```
michaelosthege: can we make the set operations a little easier to comprehend?

```suggestion
    unexpected_implemented = not_implemented - missing_moments
    unexpected_not_implemented = missing_moments - not_implemented
    
    if unexpected_implemented:
        raise Exception(
            f"Distributions {unexpected_implemented}" have a `get_moment` implemented."
            " That's good! But this test must be updated."
        )
        
    if unexpected_not_implemented:
        raise NotImplementedError(
            "Unexpected by this test, some distributions do not have a `get_moment` implementation:\n"
            + "\n".join(sorted(map(str, list(unexpected_not_implemented))))
        )
```

I didn't test this and my syntax probably doesn't pass the pre-commit. So please apply locally..

ricardoV94: Updated!

michaelosthege (approved):

PR #5141, opened 2021-11-05
Add sigma to OrderedProbit fixes #5103

### LINKED ISSUE #5103: pm.OrderedProbit bug
So the new `pm.OrderedProbit` looks great. I've been trying to work through some examples and I've noticed a mismatch between the docstring and the working of the function. In the docstring it states that you can provide a `sigma` kwarg (i.e. "The standard deviation of probit function"). However when trying to do a simple example, this errors out. Example...

```python
import pymc3 as pm
import seaborn as sns

samples = pm.OrderedProbit.dist(cutpoints=[-2, 0, 1, 2], eta=0, sigma=1).random(size=500)
sns.countplot(x=samples, color='C0');
```

but it errors with
```
TypeError: __init__() got an unexpected keyword argument 'sigma'
```

So basically the `pm.OrderedProbit` doesn't yet do what it states in the docstring. Taking a look at the code, it looks like the function does not have `sigma` as a kwarg and doesn't do anything with `sigma`.

Version details...
```
Python implementation: CPython
Python version       : 3.9.7
IPython version      : 7.28.0

theano: 1.1.2
xarray: 0.19.0

pandas    : 1.3.4
theano    : 1.1.2
arviz     : 0.11.4
pymc3     : 3.11.2
seaborn   : 0.11.2
matplotlib: 3.4.3
numpy     : 1.21.3
```

zaxtax:
This is to fix OrderedProbit to actually use the sigma argument.

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -1761,7 +1761,7 @@ class _OrderedProbit(Categorical):
     rv_op = categorical
 
     @classmethod
-    def dist(cls, eta, cutpoints, sigma, *args, **kwargs):
+    def dist(cls, eta, cutpoints, sigma = 1, *args, **kwargs):
         eta = at.as_tensor_variable(floatX(eta))
         cutpoints = at.as_tensor_variable(cutpoints)
 
@@ -1816,7 +1816,7 @@ class OrderedProbit:
         The length K - 1 array of cutpoints which break :math:`\eta` into
         ranges. Do not explicitly set the first and last elements of
         :math:`c` to negative and positive infinity.
-    sigma: float
+    sigma: float, default 1.0
          The standard deviation of probit function.
     compute_p: boolean, default True
         Whether to compute and store in the trace the inferred probabilities of each categories,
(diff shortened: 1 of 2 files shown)


pymc/distributions/discrete.py
```diff
@@ -1761,16 +1761,16 @@ class _OrderedProbit(Categorical):
     rv_op = categorical
 
     @classmethod
-    def dist(cls, eta, cutpoints, *args, **kwargs):
+    def dist(cls, eta, cutpoints, sigma, *args, **kwargs):
```
```suggestion
    def dist(cls, eta, cutpoints, sigma=1, *args, **kwargs):
```
To make it optional

zaxtax: Yep, also going to update that and the tests.

zaxtax:
Should be good to go.

pymc/distributions/discrete.py
```diff
@@ -1816,12 +1816,12 @@ class OrderedProbit:
         The length K - 1 array of cutpoints which break :math:`\eta` into
         ranges. Do not explicitly set the first and last elements of
         :math:`c` to negative and positive infinity.
+    sigma: float, default 1.0
+         The standard deviation of probit function.
```
twiecki:
```suggestion
         Standard deviation of the probit function.
```

twiecki (approved):

twiecki:
Thanks @zaxtax, good to see you here!

Issue #4426, opened 2021-01-20
Run test suite with new JAX backend


twiecki:
This will enable us to identify any gaps we might have.

michaelosthege:
We have a JAX test pipeline for quite a while now. Closing in favor of more targeted issues

PR #5142, opened 2021-11-05
Accomodate alternative Aesara backends

ricardoV94:
This PR tries to disentangle some aspects of the codebase to be more agnostic to Aesara's execution backend, so that we can easily sample with NUMBA or JAX compiled functions as well as the default C backend, by simply changing the `aesara.config.mode` flag.

This now works fine on my machine:

```python
import aesara
aesara.config.mode = "JAX"  # Default is "Mode"

#%%
import pymc as pm
import numpy as np

#%%
rng = np.random.default_rng(seed=42)
prior_idxs = np.repeat(np.arange(4), 25).astype(int)
beta_prior_true = np.abs(rng.normal(size=4))
beta_true = rng.normal(beta_prior_true[prior_idxs])
x = rng.uniform(-5, 5, size=(1000, 100))
y = np.dot(beta_true, x.T)

#%%
with pm.Model(rng_seeder=123) as m:
    beta_prior = pm.HalfNormal('beta_prior', shape=4)
    beta = pm.Normal('beta', beta_prior[prior_idxs], shape=100)
    pm.Normal('y', pm.math.dot(beta, x.T), observed=y)

#%%
pm.model_to_graphviz(m)

#%%
m.point_logps(m.recompute_initial_point(2))

#%%
with m:
    # Model's logp_dlog are compiled to JAX, with potential speedups when running on the GPU
    trace = pm.sample()
```

### DIFF
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -16,6 +16,7 @@ import warnings
 from collections import defaultdict, deque
 from typing import Dict, Iterator, NewType, Optional, Set
 
+from aesara import function
 from aesara.compile.sharedvalue import SharedVariable
 from aesara.graph.basic import walk
 from aesara.tensor.random.op import RandomVariable
@@ -159,6 +160,9 @@ class ModelGraph:
 
         graph.node(var_name.replace(":", "&"), **kwargs)
 
+    def _eval(self, var):
+        return function([], var, mode="FAST_COMPILE")()
+
     def get_plates(self):
         """Rough but surprisingly accurate plate detection.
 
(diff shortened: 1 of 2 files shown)


pymc/model_graph.py
```diff
@@ -159,6 +160,9 @@ def _make_node(self, var_name, graph, *, formatting: str = "plain"):
 
         graph.node(var_name.replace(":", "&"), **kwargs)
 
+    def _eval(self, var):
+        return function([], var, mode="FAST_COMPILE")()
```
ricardoV94, line 164: Also has the benefit that it executes faster...

twiecki:
Curious why that's what fixes it? We couldn't change backends like that before?

ricardoV94:
> Curious why that's what fixes it? We couldn't change backends like that before?

This fixes some parts I noticed were failing. There are probably more, hence why this is just a draft.

1. The `model_to_graphviz` called `.eval()` on the model `RandomVariable`s to obain the shapes, but we don't have implementations of those `Op`s for the NUMBA/JAX backends, so I force it to use the Python backend.
2. `point_logps` saves results as a pandas series and then `check_start_vals` calls `np.isnan` on these. However if you are using the JAX backend the returned ouputs are in `DeviceArray`s and this will fail.

ricardoV94:
Related to #4426

twiecki:
@ricardoV94 What's missing here?

ricardoV94:
I guess we want to test the suite (once) as mentioned in #4426

twiecki:
@ricardoV94 We can do that in a separate PR.

ricardoV94:
> @ricardoV94 We can do that in a separate PR.

Sure

PR #5147, opened 2021-11-06
Add back some missing moments

michaeloriordan:
Add moments and tests for the below distributions as part of https://github.com/pymc-devs/pymc/issues/5078

- pymc.distributions.continuous.Kumaraswamy
- pymc.distributions.continuous.Exponential
- pymc.distributions.continuous.Laplace
- pymc.distributions.continuous.StudentT
- pymc.distributions.continuous.Cauchy

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1313,6 +1313,12 @@ class Kumaraswamy(UnitContinuous):
 
         return super().dist([a, b], *args, **kwargs)
 
+    def get_moment(rv, size, a, b):
+        mean = at.exp(at.log(b) + at.gammaln(1 + 1 / a) + at.gammaln(b) - at.gammaln(1 + 1 / a + b))
+        if not rv_size_is_none(size):
+            mean = at.full(size, mean)
+        return mean
+
     def logp(value, a, b):
         """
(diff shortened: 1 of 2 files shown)


pymc/distributions/continuous.py
```diff
@@ -1399,6 +1405,12 @@ def dist(cls, lam, *args, **kwargs):
         # Aesara exponential op is parametrized in terms of mu (1/lam)
         return super().dist([at.inv(lam)], **kwargs)
 
+    def get_moment(rv, size, lam):
```
ricardoV94: We are converting lam to mu in the `dist` class method, so `get_moment` should expect `mu` already and you don't need to invert it (the same happens with the logcdf below). 

That's why your tests are failing for the exponential.

ricardoV94 (changes requested):

pymc/distributions/continuous.py
michaeloriordan: Ah that makes sense, thanks! I'll make that change.

ricardoV94 (approved):
Looks good to me. Waiting to see if the tests pass

ricardoV94:
Thanks once again @michaeloriordan!

PR #5148, opened 2021-11-06
HalfCauchy, Gamma, Weibull and LogNormal moments

lucianopaz:
Add moments and tests for the below distributions as part of #5078

- pymc.distributions.continuous.HalfCauchy
- pymc.distributions.continuous.Gamma
- pymc.distributions.continuous.Weibull
- pymc.distributions.continuous.LogNormal

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2106,6 +2106,12 @@ class HalfCauchy(PositiveContinuous):
         assert_negative_support(beta, "beta", "HalfCauchy")
         return super().dist([0.0, beta], **kwargs)
 
+    def get_moment(rv, size, loc, beta):
+        mean = beta
+        if not rv_size_is_none(size):
+            mean = at.full(size, mean)
+        return mean
+
     def logcdf(value, loc, beta):
         """
         Compute the log of the cumulative distribution function for HalfCauchy distribution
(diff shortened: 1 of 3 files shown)


pymc/distributions/continuous.py
```diff
@@ -2100,6 +2106,12 @@ def dist(cls, beta, *args, **kwargs):
         assert_negative_support(beta, "beta", "HalfCauchy")
         return super().dist([0.0, beta], **kwargs)
 
+    def get_moment(rv, size, loc, beta):
+        mean, _ = at.broadcast_arrays(at.as_tensor(0.0, dtype=rv.dtype), beta)
```
ricardoV94: Does something like this also work?

```suggestion
        mean = at.zeros_like(beta)
```

ricardoV94: Is zero a good point to start sampling from this distribution?

ricardoV94:
@lucianopaz seems like pre-commit is failing with import sorting

pymc/distributions/continuous.py
ricardoV94: Looking at V3 I think the median (beta) takes precedence over the mode (0), so we were probably using the former to initialize halfcauchys

lucianopaz: Oh, I'll use the median instead then. Maybe we should list the order of preference for the moments in #5078

lucianopaz:
> @lucianopaz seems like pre-commit is failing with import sorting

Yeah I messed up the pre commit while rebasing the PR. But now everything should be fixed

twiecki (approved):

twiecki:
👍

PR #5150, opened 2021-11-07
Binomial and Poisson Moment

farhanreynaldo:
Add moments and tests for the below distributions as part of https://github.com/pymc-devs/pymc/issues/5078:

* pymc.distributions.continuous.Binomial
* pymc.distributions.Discrete.Poisson

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -114,11 +114,10 @@ class Binomial(Discrete):
     def dist(cls, n, p, *args, **kwargs):
         n = at.as_tensor_variable(intX(n))
         p = at.as_tensor_variable(floatX(p))
-        # mode = at.cast(tround(n * p), self.dtype)
         return super().dist([n, p], **kwargs)
 
     def get_moment(rv, size, n, p):
-        mean = n * p
+        mean = at.round(n * p)
         if not rv_size_is_none(size):
             mean = at.full(size, mean)
         return mean
@@ -573,10 +572,10 @@ class Poisson(Discrete):
     @classmethod
     def dist(cls, mu, *args, **kwargs):
         mu = at.as_tensor_variable(floatX(mu))
-        # mode = intX(at.floor(mu))
         return super().dist([mu], *args, **kwargs)
 
     def get_moment(rv, size, mu):
+        mu = at.floor(mu)
         if not rv_size_is_none(size):
             mu = at.full(size, mu)
         return mu
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -117,6 +117,12 @@ class Binomial(Discrete):
         # mode = at.cast(tround(n * p), self.dtype)
         return super().dist([n, p], **kwargs)
 
+    def get_moment(rv, size, n, p):
+        mean = n * p
+        if not rv_size_is_none(size):
+            mean = at.full(size, mean)
+        return mean
+
     def logp(value, n, p):
         r"""
         Calculate log-probability of Binomial distribution at specified value.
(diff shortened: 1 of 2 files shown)


twiecki:
@farhanreynaldo Need to resolve conflicts.

farhanreynaldo:
> @farhanreynaldo Need to resolve conflicts.

This is my first time having a conflict. Are the following steps correct?

```
git fetch upstream
git rebase upstream/main
```

and then accept both changes and push the resolved conflicts?
```
git commit -m "resolve conflicts"
git push origin <branch-name>
```

twiecki:
I think during rebasing you just git add, and then you force push.

pymc/distributions/discrete.py
```diff
@@ -570,6 +576,11 @@ def dist(cls, mu, *args, **kwargs):
         # mode = intX(at.floor(mu))
```
ricardoV94: Seems like we have a leftover from before. Do you mind removing the commented out code?
```suggestion
```

ricardoV94: Also note that we used the floor of mu, we should do the same in the new `get_moment`

pymc/distributions/discrete.py
```diff
@@ -117,6 +117,12 @@ def dist(cls, n, p, *args, **kwargs):
         # mode = at.cast(tround(n * p), self.dtype)
```
```suggestion
```

Same here, we should remove the old commented out code, and use `at.round` in the new `get_moment`

ricardoV94 (changes requested):
Left a couple of suggestions @farhanreynaldo. Thanks for the PR!

pymc/distributions/discrete.py
farhanreynaldo: Is the `IntX` necessary or should I just include the `at.floor`?

pymc/distributions/discrete.py
farhanreynaldo: Is the `at.cast` also necessary here?

ricardoV94: Not anymore. That's done automatically by the function that calls get_moment

pymc/distributions/discrete.py
ricardoV94: Not needed either I think

pymc/tests/test_distributions_moments.py
```diff
@@ -318,11 +326,44 @@ def test_gamma_moment(alpha, beta, size, expected):
...
+@pytest.mark.parametrize(
+    "mu, size, expected",
+    [
+        (2, None, 2),
+        (2, 5, np.full(5, 2)),
+        (np.arange(1, 5), None, np.arange(1, 5)),
+        (np.arange(1, 5), (2, 4), np.full((2, 4), np.arange(1, 5))),
+    ],
+)
+def test_poisson_moment(mu, size, expected):
```
ricardoV94, line 366: Some of the test combinations should include non integer values for mu to check the flooring is working as expected

pymc/tests/test_distributions_moments.py
```diff
@@ -318,11 +326,44 @@ def test_gamma_moment(alpha, beta, size, expected):
...
+@pytest.mark.parametrize(
+    "n, p, size, expected",
+    [
+        (10, 0.5, None, 5),
+        (10, 0.5, 5, np.full(5, 5)),
+        (10, np.arange(1, 6) / 10, None, np.arange(1, 6)),
+        (10, np.arange(1, 6) / 10, (2, 5), np.full((2, 5), np.arange(1, 6))),
+    ],
+)
+def test_binomial_moment(n, p, size, expected):
```
ricardoV94, line 351: Some of the test combinations should have a `n*p` that is different when rounded and not rounded to make sure it is working as expected

ricardoV94 (commented):
Some suggestions to make the tests more robust

ricardoV94 (approved):

PR #5152, opened 2021-11-08
Add HalfStudentT moment

azihna:
Add moments and tests for the below distribution as part of #5078

pymc.distributions.continuous.HalfStudentT

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2635,11 +2635,10 @@ class HalfStudentT(PositiveContinuous):
         return super().dist([nu, sigma], *args, **kwargs)
     
     def get_moment(rv, size, nu, sigma):
-        median = at.as_tensor_variable(sigma)
         sigma, _ = at.broadcast_arrays(sigma, nu)
         if not rv_size_is_none(size):
-            median = at.full(size, median)
-        return median
+            sigma = at.full(size, sigma)
+        return sigma
 
     def logp(value, nu, sigma):
         """
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2633,7 +2633,7 @@ class HalfStudentT(PositiveContinuous):
         assert_negative_support(sigma, "sigma", "HalfStudentT")
 
         return super().dist([nu, sigma], *args, **kwargs)
-    
+
     def get_moment(rv, size, nu, sigma):
         sigma, _ = at.broadcast_arrays(sigma, nu)
         if not rv_size_is_none(size):
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2637,6 +2637,12 @@ class HalfStudentT(PositiveContinuous):
         assert_negative_support(sigma, "sigma", "HalfStudentT")
 
         return super().dist([nu, sigma], *args, **kwargs)
+    
+    def get_moment(rv, size, nu, sigma):
+        median = at.as_tensor_varialbe(sigma)
+        if not rv_size_is_none(size):
+            median = at.full(size, median)
+        return median
 
     def logp(value, nu, sigma):
         """
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2635,8 +2635,8 @@ class HalfStudentT(PositiveContinuous):
         return super().dist([nu, sigma], *args, **kwargs)
     
     def get_moment(rv, size, nu, sigma):
-        lam, sigma = get_tau_sigma(None, sigma)
         median = at.as_tensor_variable(sigma)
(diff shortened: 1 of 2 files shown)


pymc/distributions/continuous.py
```diff
@@ -2628,15 +2628,18 @@ def dist(cls, nu=1, sigma=None, lam=None, sd=None, *args, **kwargs):
...
-        # sd = at.as_tensor_variable(sigma)
-
         assert_negative_support(nu, "nu", "HalfStudentT")
         assert_negative_support(lam, "lam", "HalfStudentT")
         assert_negative_support(sigma, "sigma", "HalfStudentT")
 
         return super().dist([nu, sigma], *args, **kwargs)
+    
+    def get_moment(rv, size, nu, sigma):
+        lam, sigma = get_tau_sigma(None, sigma)
```
ricardoV94: This line is not necessary as sigma is already given

ricardoV94: And you need to make sure sigma is broadcasted to the shape of nu, with `sigma, _ = at.broadcast_arrays(sigma, nu)` Otherwise you might return the wrong shape for the moment (e.g., when `sigma=0` and `nu=[1,2,3,4,5]`)

ricardoV94 (changes requested):
Thanks for opening the PR. I left some comments

pymc/tests/test_distributions_moments.py
```diff
@@ -127,6 +128,21 @@ def test_halfnormal_moment(sigma, size, expected):
     assert_moment_is_expected(model, expected)
 
 
+@pytest.mark.parametrize(
+    "nu, sigma, size, expected",
+    [
+        (1, 1, None, 1),
+        (1, 1, 5, np.ones(5)),
+        (1, np.arange(5), None, np.arange(5)),
```
ricardoV94: You should add a test where nu is a vector and sigma is a scalar. That would have caught the issue I mentioned above. 

Something like this:
```suggestion
        (1, np.arange(5), None, np.arange(5)),
        (np.arange(5), 1, None, np.full(5, 1)),
```

azihna: Thanks for the review @ricardoV94, added a new commit with both suggestions.

pymc/distributions/continuous.py
azihna: Thanks for the tip, change applied.

pymc/distributions/continuous.py
```diff
@@ -2628,15 +2628,18 @@ def dist(cls, nu=1, sigma=None, lam=None, sd=None, *args, **kwargs):
...
         assert_negative_support(sigma, "sigma", "HalfStudentT")
 
         return super().dist([nu, sigma], *args, **kwargs)
+    
+    def get_moment(rv, size, nu, sigma):
+        median = at.as_tensor_variable(sigma)
+        sigma, _ = at.broadcast_arrays(sigma, nu)
+        if not rv_size_is_none(size):
+            median = at.full(size, median)
+        return median
```
ricardoV94: I think you want something like this:

```suggestion
        sigma, _ = at.broadcast_arrays(sigma, nu)
        if not rv_size_is_none(size):
            sigma = at.full(size, sigma)
        return sigma
```

ricardoV94 (changes requested):
There we still some issues in the get_moment. Let me know if the suggestion makes sense

ricardoV94 (approved):
Looks good, waiting to see if tests pass

ricardoV94:
Pre-commit was failing due to sorting of imports. There is a guide here on how to setup locally to make sure it's working before pushing: https://github.com/pymc-devs/pymc/wiki/Python-Code-Style

pymc/tests/test_distributions_moments.py
```diff
@@ -127,6 +128,22 @@ def test_halfnormal_moment(sigma, size, expected):
...
+        (1, np.arange(5), (2, 5), np.full((2, 5), np.arange(5))),
+        (np.arange(1, 6), 1, 5, np.full(5, 1)),
```
ricardoV94: This one does not really test the broadcast because size is not None. Maybe something like this?


```suggestion
        (np.arange(1, 6), 1, None, np.full(5, 1)),
```

And you can remove one of the previous 2 tests, we don't need 5 in total.

ricardoV94 (changes requested):

pymc/tests/test_distributions_moments.py
azihna: Sure! Thanks for your patience!

ricardoV94: I thank you :)

ricardoV94 (approved):

ricardoV94:
Thanks!

PR #5158, opened 2021-11-08
Added moment for geometric distribution along with test for issue #5078

sagartomar:

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -819,6 +819,12 @@ class Geometric(Discrete):
         p = at.as_tensor_variable(floatX(p))
         return super().dist([p], *args, **kwargs)
 
+    def get_moment(rv, size, p):
+        mean = at.round(1.0 / p)
+        if not rv_size_is_none(size):
+            mean = at.full(size, mean)
+        return mean
+
     def logp(value, p):
         r"""
         Calculate log-probability of Geometric distribution at specified value.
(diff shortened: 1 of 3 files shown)


pymc/distributions/discrete.py
```diff
@@ -813,6 +813,12 @@ def dist(cls, p, *args, **kwargs):
         p = at.as_tensor_variable(floatX(p))
         return super().dist([p], *args, **kwargs)
 
+    def get_moment(rv, size, p):
```
ricardoV94, line 822: We were using the mode before (always return 1 regardless of `p`) 

Not sure if we should change the default to be the rounded mean as in this PR. Do you have a good reason for it?

sagartomar: It was mentioned in #5078 that "the "higher-order" one is the most useful (that is mean > median > mode)" that's why I went with mean. Apart from that I don't really have any good reason. Let me know if I should change it back to mode.

ricardoV94: Yeah in general the mean is better but not necessarily so for discrete distributions. The geometric has the additional issue that it can blow up quite easily. 

Maybe someone else will be able to ping in here...

ricardoV94: If nobody does, my gut feeling would be to take the conservative choice and keep things as they were in V3 (that is using the mode)

ricardoV94: Changed my mind, let's use this and see if anybody complains :D. You still need to solve the merge conflicts that cropped in due to other merged PRs

ricardoV94:
It seems like you picked some commits that are not part of your PR. You will probably need to rebase to fix it.

ricardoV94 (approved):
I rebased your PR and added a couple of unrelated commits that I had in the backlog.

ricardoV94:
Thanks @sagartomar just merged the PR! Looking forward to your next one :)

sagartomar:
Thanks @ricardoV94

PR #5159, opened 2021-11-08
Add DensityDist moment

lucianopaz:
Adds a default moment of 0 for the `DensityDist` distribution. This allows the `DensityDist` to be used by `pm.sample` even if the users don't explicitly pass a `get_moment` function to it.

This PR contributes to #5078

- pymc.distributions.distribution.DensityDist


ricardoV94:
This won't work for multivariate distributions... 

Should we raise a NotImplementedError in those cases. We can inspect the RV to know the number of dimensions.

Also we should port these tests to the present module: https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_moment.py ?

lucianopaz:
> This won't work for multivariate distributions...
> 
> Should we raise a NotImplementedError in those cases. We can inspect the RV to know the number of dimensions.

I hadn’t thought of that. How can we know the shape of the support dimensions? We could potentially return a tensor with the support dimensions as broadcastable. Would that then be broadcasted correctly at the later sampling stage?

> Also we should port these tests to the present module: https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_moment.py ?

Yes, and maybe also remove that module entirely

lucianopaz:
By the way, this should be very similar to what we’ll want to do for the Simulator distribution, so once we agree on how we handle multivariate distributions, I can port it to the simulator

ricardoV94:
> I hadn’t thought of that. How can we know the shape of the support dimensions? We could potentially return a tensor with the support dimensions as broadcastable. Would that then be broadcasted correctly at the later sampling stage?

We can access `ndim_supp` and `ndims_params` class attributes from the RV Op, but that only gives the minimum dimensions, and we have no way of knowing how they map to each other. 

We can also just return something like `at.zeros_like(rv)` which should work in case the user implemented the random function. But in that case defaulting to a sample from the prior might actually be more sensible.

> Yes, and maybe also remove that module entirely

Sounds good.

ricardoV94:
> By the way, this should be very similar to what we’ll want to do for the Simulator distribution, so once we agree on how we handle multivariate distributions, I can port it to the simulator

Yeah I think so too. My only question is whether for the Simulator a draw from the random function may actually be the preferred starting point. We haven't really explored how well Simulators behave in non SMC samplers... In SMC it doesn't matter because we always start sampling form the prior and not from moments.

@aloctavodia any opinion on this?

lucianopaz:
@ricardoV94, I wrote a very small fallback option that uses the size of a random draw to inform the size of the moment, in the case of multivariate `DensityDist`. If the users haven't supplied the `random` method, then a `TypeError` is raised. Let me know what you think about this solution. If you think it's ok, I'll rebase the PR.

ricardoV94:
> @ricardoV94, I wrote a very small fallback option that uses the size of a random draw to inform the size of the moment, in the case of multivariate `DensityDist`. If the users haven't supplied the `random` method, then a `TypeError` is raised. Let me know what you think about this solution. If you think it's ok, I'll rebase the PR.

Sounds good. I'll check later more carefully for any bugs/issues but you can rebase in the meantime.

pymc/distributions/distribution.py
```diff
@@ -615,3 +618,18 @@ def func(*args, **kwargs):
         raise NotImplementedError(message)
 
     return func
+
+
+def default_get_moment(rv, size, *rv_inputs, rv_name=None, has_fallback=False):
+    moment = at.zeros(size, dtype=rv.dtype)
+    if moment.ndim == rv.ndim:
+        return moment
+    if moment.ndim != rv.ndim:
```
ricardoV94:
```suggestion
    else:
```

pymc/tests/test_distributions_moments.py
```diff
@@ -434,3 +440,83 @@ def test_logistic_moment(mu, s, size, expected):
...
+def test_density_dist_moment(get_moment, size, expected):
+    if get_moment == "custom_moment":
+        get_moment = lambda rv, size, *rv_inputs: 5 * at.ones(size, dtype=rv.dtype)
+    with Model() as model:
+        DensityDist("x", get_moment=get_moment, size=size)
+    assert_moment_is_expected(model, expected)
+
+
+@pytest.mark.parametrize("size", [(), (2,), (3, 2)], ids=str)
+def test_density_dist_moment_scalar(size):
```
ricardoV94:
```suggestion
def test_density_dist_custom_moment_univariate(size):
```

Just a more descriptive test name

pymc/tests/test_distributions_moments.py
```diff
@@ -434,3 +440,83 @@ def test_logistic_moment(mu, s, size, expected):
...
+    with pm.Model():
+        mu = pm.Normal("mu")
+        a = pm.DensityDist("a", mu, get_moment=moment, size=size)
+    evaled_moment = get_moment(a).eval({mu: mu_val})
+    assert evaled_moment.shape == to_tuple(size)
+    assert np.all(evaled_moment == mu_val)
+
+
+@pytest.mark.parametrize("size", [(), (2,), (3, 2)], ids=str)
+def test_density_dist_moment_multivariate(size):
```
ricardoV94:
```suggestion
def test_density_dist_custom_moment_multivariate(size):
```

pymc/tests/test_distributions_moments.py
```diff
@@ -434,3 +440,83 @@ def test_logistic_moment(mu, s, size, expected):
...
+@pytest.mark.parametrize(
+    "get_moment, size, expected",
+    [
+        (None, None, 0.0),
+        (None, 5, np.zeros(5)),
+        ("custom_moment", None, 5),
+        ("custom_moment", (2, 5), np.full((2, 5), 5)),
+    ],
+)
+def test_density_dist_moment(get_moment, size, expected):
```
ricardoV94:
```suggestion
def test_density_dist_default_moment_univariate(get_moment, size, expected):
```

And I think you can remove the "custom_moment" checks as they are already covered in the tests below

pymc/distributions/distribution.py
```diff
@@ -615,3 +618,18 @@ def func(*args, **kwargs):
         raise NotImplementedError(message)
 
     return func
+
+
+def default_get_moment(rv, size, *rv_inputs, rv_name=None, has_fallback=False):
+    moment = at.zeros(size, dtype=rv.dtype)
+    if moment.ndim == rv.ndim:
```
ricardoV94: Mulling it over, I am not sure we can safely assume we got the right `size` by simply checking the `ndim` matches. 

The shape may be determined by the `rv_inputs` and not the size:

```python
pm.DensityDist('x', np.arange(5), logp=lambda value, rv_input1: ..., size=None)
```

We can perhaps use the `rv._infer_shape` method, which seems to do just that: 
https://github.com/aesara-devs/aesara/blob/d9fd640e04164759ac752333b94b294a0a5c5117/aesara/tensor/random/op.py#L162

I would still be weary of relying on that for multivariate variables because that often requires overriding for correct multivariate RandomVariable shape inference

ricardoV94 (changes requested):

pymc/distributions/distribution.py
lucianopaz: I guess this depends on whether something like `pm.Normal(..., mu=np.arange(3), size=())` will work, and broadcast into something that has shape `(3,)`. From what I understand, numpy and scipy forbid this, and always force the `size` to match the shape induced by the distribution parameters.

```python
>>> np.random.default_rng().normal(loc=np.arange(3), scale=1, size=())

----> 1 np.random.default_rng().normal(loc=np.arange(3), scale=1, size=())

_generator.pyx in numpy.random._generator.Generator.normal()

_common.pyx in numpy.random._common.cont()

_common.pyx in numpy.random._common.cont_broadcast_2()

_common.pyx in numpy.random._common.validate_output_shape()

ValueError: Output size () is not compatible with broadcast dimensions of inputs (3,).
```

ricardoV94: Unfortunately Aesara behaves differently than Numpy/Scipy, since those are quite restrictive: https://github.com/aesara-devs/aesara/discussions/412

ricardoV94 (approved):
Looks great @lucianopaz, just need to resolve the conflicts with the main branch

lucianopaz:
> Looks great @lucianopaz, just need to resolve the conflicts with the main branch

Damn, I can't keep up with all the merged PRs

ricardoV94:
Thanks @lucianopaz!

PR #5161, opened 2021-11-09
Add Pareto and Wald moment

azihna:
Add moments and tests for the below distribution as part of #5078
* pymc.distributions.continuous.Pareto
* pymc.distributions.continuous.Wald

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -995,6 +995,12 @@ class Wald(PositiveContinuous):
 
         return super().dist([mu, lam, alpha], **kwargs)
 
+    def get_moment(rv, size, mu, lam, alpha):
+        mu, _, _ = at.broadcast_arrays(mu, lam, alpha)
+        if not rv_size_is_none(size):
+            mu = at.full(size, mu)
+        return mu
+
     @staticmethod
     def get_mu_lam_phi(
         mu: Optional[float], lam: Optional[float], phi: Optional[float]
@@ -1943,8 +1949,11 @@ class Pareto(BoundedContinuous):
 
         return super().dist([alpha, m], **kwargs)
 
-    def _distr_parameters_for_repr(self):
-        return ["alpha", "m"]
+    def get_moment(rv, size, alpha, m):
+        median = m * 2 ** (1 / alpha)
+        if not rv_size_is_none(size):
+            median = at.full(size, median)
+        return median
 
     def logcdf(
         value: Union[float, np.ndarray, TensorVariable],
--- a/pymc/tests/test_distributions_moments.py
+++ b/pymc/tests/test_distributions_moments.py
@@ -26,10 +26,12 @@ from pymc.distributions import (
     LogNormal,
     NegativeBinomial,
     Normal,
+    Pareto,
     Poisson,
     StudentT,
     TruncatedNormal,
     Uniform,
+    Wald,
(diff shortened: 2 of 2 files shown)


pymc/distributions/continuous.py
```diff
@@ -995,6 +995,11 @@ def dist(
 
         return super().dist([mu, lam, alpha], **kwargs)
 
+    def get_moment(rv, size, mu, lam, phi):
```
ricardoV94: You will need to broadcast mu to the shape of lam and phi, and add test conditions that would fail if you were not to do that. For instance by making mu a scalar, and lam or phi vectors (and size None).


```suggestion
    def get_moment(rv, size, mu, lam, phi):
        mu, _, _, = at.broadcast_arrays(mu, lam, phi)
```

ricardoV94 (changes requested):
Thanks for opening that PR. I left some comments below

pymc/distributions/continuous.py
```diff
@@ -1943,6 +1948,12 @@ def dist(
 
         return super().dist([alpha, m], **kwargs)
 
+    def get_moment(rv, size, alpha, m):
+        mean = alpha * m / (alpha - 1)
```
ricardoV94: Hmm this moment isn't great when alpha <= 1. It was also not great in V3!

Perhaps we should use the median or switch between the mean and the median when alpha <= 1?

azihna: Thanks for this! I had that question on my mind as well! I think the switch might be confusing when using it. I'd much rather do the switching to median sounds better. What do you think?

ricardoV94: Which one do you mean? The two option I see are:
1. Use median regardless of alpha
2. Switch between mean and median based on alpha

I think I like option 1 better but perhaps there's a good argument for 2 (or 3 if you have any)

azihna: I also meant that it should be option 1. I'll change the implementation as such. Thanks!

pymc/distributions/continuous.py
azihna: I added broadcasting, for the tests, would it be fine to go over 4 tests or should I try to blend it into the existing ones?

ricardoV94: It's fine to go over 4 if they are testing different things, specially with a distribution that has 3 parameters

azihna: I added 3 more tests to check different parameters.

ricardoV94: The tests do not cover the cases where `at.broadcast_arrays` is needed. You can confirm this by commenting out the line in `get_moment` and check that all tests still pass. You need to choose a combination of parameters that would fail without this line.

The following seem to do the trick:
```python
(1, None, np.ones(5), None, np.full(5, 1)),
(3, np.full(5, 2), None, None, np.full(5, 3)),
```

Besides this, it should be enough to include 2/3 tests more. More than 5 is probably too much.

Let me know if you have any questions

azihna: Got it. I thought changing it to vector would do the trick. Thanks for the guidance. Updated the tests as recommended.

ricardoV94 (approved):
Thanks @azihna I just rebased on main and removed some other obsolete code. Should be good to merge if tests pass

PR #5165, opened 2021-11-09
Add ExGaussian moment

karm-patel:
Add moment and tests for the below distribution as part of #5078

pymc.distributions.continuous.ExGaussian


pymc/distributions/continuous.py
```diff
@@ -2771,15 +2771,18 @@ def dist(cls, mu=0.0, sigma=None, nu=None, sd=None, *args, **kwargs):
...
-        # variance = (sigma ** 2) + (nu ** 2)
-
         assert_negative_support(sigma, "sigma", "ExGaussian")
         assert_negative_support(nu, "nu", "ExGaussian")
 
         return super().dist([mu, sigma, nu], *args, **kwargs)
 
+    def get_moment(rv, size, mu, sigma, nu):
+        moment, _ = at.broadcast_arrays(sigma, nu)
+        moment = mu + nu
```
ricardoV94: This is probably what you want. Otherwise your first line is not doing anything as you overwriting moment after

```suggestion
        mu, nu, _ = at.broadcast_arrays(mu, nu, sigma)
        moment = mu + nu
```

You should also change one of the tests, making nu a vector and mu, sigma scalars and size is `None` to test the case where that first line is really needed

ricardoV94 (changes requested):
Thanks for opening the PR! I left some comments below

pymc/distributions/continuous.py
karm-patel: Thanks for the suggestion, I have corrected `at.broadcast_arrays()` and added one test case for the same. But I am not sure about the test case if it is proper or not. Can you please check that?

ricardoV94: The tests do not seem to cover the case(s) where `at.broadcast_arrays` would be necessary. You can check this by commenting out that line in `get_moment` and verifying that all tests still pass.

twiecki:
@karm216 Thanks -- some conflicts left to resolve.

pymc/distributions/continuous.py
karm-patel: Finally, I tried to make a test case that cover the `at.broadcast_arrays()` line, please look into that!

karm-patel:
Hi @ricardoV94, Can you please check if everything is fine?

ricardoV94 (approved):
Looks great. Waiting to see if tests pass

karm-patel:
Thank you for accepting PR :)

PR #5166, opened 2021-11-10
Refactor logpt calls to aeppl

ricardoV94:
This snippet would raise an unnecessary warning before this PR:

```python
import pymc as pm
with pm.Model() as m:
    x = pm.Uniform('x')
m.logpt   
```

```
warnings.warn("No value variables provided the logp will be an empty graph")
```

CC @kc611

### DIFF
--- a/pymc/distributions/logprob.py
+++ b/pymc/distributions/logprob.py
@@ -195,33 +195,26 @@ def logpt(
                 getattr(_var.tag, "total_size", None), rv_value_var.shape, rv_value_var.ndim
             )
 
-    # Unlike aeppl, PyMC's logpt is expected to plug in the values variables to corresponding
-    # RVs automatically unless the values are explicity set to None. Hence we iterate through
-    # the graph to find RVs and construct a new RVs to values dictionary.
+    # Aeppl needs all rv-values pairs, not just that of the requested var.
+    # Hence we iterate through the graph to collect them.
     tmp_rvs_to_values = rv_values.copy()
     transform_map = {}
     for node in io_toposort(graph_inputs(var), var):
-        if isinstance(node.op, RandomVariable):
-            curr_var = node.out
+        try:
+            curr_vars = [node.default_output()]
+        except ValueError:
+            curr_vars = node.outputs
+        for curr_var in curr_vars:
             rv_value_var = getattr(
-                curr_var.tag, "observations", getattr(curr_var.tag, "value_var", curr_var)
+                curr_var.tag, "observations", getattr(curr_var.tag, "value_var", None)
             )
+            if rv_value_var is None:
+                continue
             rv_value = rv_values.get(curr_var, rv_value_var)
             tmp_rvs_to_values[curr_var] = rv_value
             # Along with value variables we also check for transforms if any.
             if hasattr(rv_value_var.tag, "transform") and transformed:
                 transform_map[rv_value] = rv_value_var.tag.transform
-        # The condition below is a hackish way of excluding the value variable for the
-        # RV being indexed in case of Advanced Indexing of RVs. It gets added by the
-        # logic above but aeppl does not expect us to include it in the dictionary of
-        # {RV:values} given to it.
-        if isinstance(node.op, subtensor_types):
-            curr_var = node.out
-            if (
-                curr_var in tmp_rvs_to_values.keys()
-                and curr_var.owner.inputs[0] in tmp_rvs_to_values.keys()
-            ):
-                tmp_rvs_to_values.pop(curr_var.owner.inputs[0])
 
     transform_opt = TransformValuesOpt(transform_map)
     temp_logp_var_dict = factorized_joint_logprob(
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -714,54 +714,24 @@ class Model(Factor, WithMemoization, metaclass=ContextMeta):
                     raise ValueError(f"Can only compute the gradient of continuous types: {var}")
 
         if tempered:
-            with self:
-                # Convert random variables into their log-likelihood inputs and
-                # apply their transforms, if any
-                potentials, _ = rvs_to_value_vars(self.potentials, apply_transforms=True)
-
-                free_RVs_logp = at.sum(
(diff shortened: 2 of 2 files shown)


pymc/distributions/logprob.py
```diff
@@ -215,6 +215,7 @@ def logpt(
         # RV being indexed in case of Advanced Indexing of RVs. It gets added by the
         # logic above but aeppl does not expect us to include it in the dictionary of
         # {RV:values} given to it.
+        # TODO: This should no longer be needed!
         if isinstance(node.op, subtensor_types):
```
twiecki: Why not remove it?

ricardoV94: I didn't have time to properly check it.

kc611: The issue with removing that is it'll give out unused input errors while constructing the graph down the line in `aeppl.factorized_logprob`. Which technically we could turn off but that' cause problems on user side (They wouldn't know of unused inputs  if we do it implicitly)

It does work if we allow unused inputs, But the above stated is the only reason I decided to keep it in the end.

Also with current way of building the value variable dictionary (By iterating over the graph nodes using `toposort`). I don't think we can avoid adding that RV value in the first place unless we put in a similar `Subtensor` check for it's parent node.

Probably the best way to deal with this issue long term will be to avoid the use of `tag.value_var` and switching to `value_var` dictionary as the only way to provide value variables. (This approach will also remove a whole lot of extra logic from `logpt`)

ricardoV94: @kc611 Can you provide a MWE? It's hard to understand from the comment alone what is the purpose of that branch and whether it is still valid. For instance we no longer have the `{var: None}` thing going on in aeppl but the comment above seems to talk about it.

kc611:
```python
import aesara  
import aesara.tensor as at  
import numpy as np
import pymc as pm

size = 5
mu_base = np.arange(size)

mu = np.stack([mu_base, -mu_base])
sigma = 0.001
A_rv = pm.Normal.dist(mu, sigma)
A_rv.name = "A"

p = 0.5
I_rv = pm.Bernoulli.dist(p, size=size)
I_rv.name = "I"

A_idx = A_rv[I_rv]

A_idx_value_var = A_idx.type()
A_idx_value_var.name = "A_idx_value"

I_value_var = I_rv.type()
I_value_var.name = "I_value"

A_idx_logp = pm.logpt(A_idx, {A_idx: A_idx_value_var, I_rv: I_value_var}, sum=False)

# If you comment out the subtensor branch the following line will not run
logp_vals_fn = aesara.function([A_idx_value_var, I_value_var], A_idx_logp)

# But this will
logp_vals_fn = aesara.function([A_idx_value_var, I_value_var], A_idx_logp, on_unused_input='ignore')

aesara.dprint(logp_vals_fn)
# Graph without the subtensor branch in logpt 
# Elemwise{Composite{Switch(Cast{int8}((GE(i0, i1) * LE(i0, i2))), i3, i4)}} [id A] 'I_value_logprob'   0
#  |I_value [id B]
#  |TensorConstant{(1,) of 0} [id C]
#  |TensorConstant{(1,) of 1} [id D]
#  |TensorConstant{(1,) of -0..1805599453} [id E]
#  |TensorConstant{(1,) of -inf} [id F]

# Graph with the subtensor branch in logpt
# Elemwise{Add}[(0, 1)] [id A] ''   8
#  |Elemwise{Composite{Switch(Cast{int8}((GE(i0, i1) * LE(i0, i2))), i3, i4)}} [id B] ''   3
#  | |InplaceDimShuffle{x,0} [id C] ''   0
#  | | |I_value [id D]
#  | |TensorConstant{(1, 1) of 0} [id E]
#  | |TensorConstant{(1, 1) of 1} [id F]
#  | |TensorConstant{(1, 1) of ..1805599453} [id G]
#  | |TensorConstant{(1, 1) of -inf} [id H]
#  |Assert{msg='sigma > 0'} [id I] 'A_idx_value_logprob'   7
#    |Elemwise{Composite{((i0 + (i1 * sqr(((i2 - i3) / i4)))) - log(i4))}}[(0, 4)] [id J] ''   5
#    | |TensorConstant{(1, 1) of ..5332046727} [id K]
#    | |TensorConstant{(1, 1) of -0.5} [id L]
#    | |A_idx_value [id M]
#    | |AdvancedSubtensor1 [id N] ''   2
#    | | |TensorConstant{[[ 0  1  2..-2 -3 -4]]} [id O]
#    | | |I_value [id D]
#    | |AdvancedSubtensor1 [id P] ''   1
#    |   |TensorConstant{(2, 5) of 0.001} [id Q]
#    |   |I_value [id D]
#    |All [id R] ''   6
#      |Elemwise{gt,no_inplace} [id S] ''   4
#        |AdvancedSubtensor1 [id P] ''   1
#        |TensorConstant{(1, 1) of 0.0} [id T]
```

kc611: This MWE is derived from the `test_logprob.py` file so there are more instances of subtensors over there which show similar behavior.

pymc/model.py
```diff
@@ -718,50 +718,53 @@ def logp_dlogp_function(self, grad_vars=None, tempered=False, **kwargs):
...
-        rv_factors = logpt(self.free_RVs, rv_values)
+        if rv_values:
+            logp_var += logpt(self.free_RVs, rv_values)
 
         obs_values = {}
         for obs in self.observed_RVs:
             obs_values[obs] = obs.tag.observations
-        obs_factors = logpt(self.observed_RVs, obs_values)
+        if obs_values:
+            logp_var += logpt(self.observed_RVs, obs_values)
```
ricardoV94: @kc611 Do you remember if there was a reason why we need two separate calls to `logpt` and why can't we just pass `rv_values` + `obs_values` once?

kc611: I guess in some cases there were overlapping keys in those value dictionaries.

ricardoV94: That's not an issue for dictionaries, right?

kc611 (approved):
LGTM!

PR #5169, opened 2021-11-10
Add SymbolicDistribution and Censored distributions

ricardoV94:
With https://github.com/aesara-devs/aeppl/pull/22 we can now compute the logprob of censored Distributions.

This PR adds those to PyMC:

```python
import pymc as pm

with pm.Model(rng_seeder=2021) as m:
    x = pm.Normal.dist()
    y = pm.Censored('y', x, -.1, .1)

print([float(y.eval()) for _ in range(5)])
# [0.1, 0.01990278315435156, 0.1, -0.1, -0.1]

y_val = m.rvs_to_values[y]
for y_test_val in [-.2, -.1, -.09, 0, 0.09, .1, .2]:
    print(y_test_val, m.logp({y_val: y_test_val}))
# -0.2 -inf
# -0.1 -0.7761545942065803
# -0.09 -0.9229885332046727
# 0 -0.9189385332046727
# 0.09 -0.9229885332046727
# 0.1 -0.7761545942065803
# 0.2 -inf
```

A large chunk of this PR concerns how to accommodate aeppl derived distributions with PyMC API seamlessly. To do this I started by duplicating `pymc.Distribution` into `pymc.DerivedDistribution` and changing things as necessary. Hopefully we will be able to use the same class in the end or at least refactor large chunks that are shared... 

This may not be the best way... Suggestions are welcome

What works:
- [x] Defining size/shape and observed
- [x] Evaluating model logp
- [x] Prior predictive sampling
- [x] Sampling
- [x] Posterior predictive sampling

TODO:
- [x] Transforms
- [x] Dims
- [ ] String representation
- [x] Integration with `shape_utils.maybe_resize`
- [x] fix log_likelihood in InferenceDataConverter
- [x] get_moment
- [x] tests
- [x] DOCSTRINGS
- [x] Open issue in pymc-examples to update examples involving censoring

### DIFF
--- a/pymc/distributions/censored.py
+++ b/pymc/distributions/censored.py
@@ -1,0 +1,146 @@
+#   Copyright 2020 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+import aesara.tensor as at
+import numpy as np
+
+from aesara.scalar import Clip
+from aesara.tensor import TensorVariable
+from aesara.tensor.random.op import RandomVariable
+
+from pymc.distributions.distribution import SymbolicDistribution, _get_moment
+from pymc.util import check_dist_not_registered
+
+
+class Censored(SymbolicDistribution):
+    r"""
+    Censored distribution
+
+    The pdf of a censored distribution is
+
+    .. math::
+
+        \begin{cases}
+            0 & \text{for } x < lower, \\
+            \text{CDF}(lower, dist) & \text{for } x = lower, \\
+            \text{PDF}(x, dist) & \text{for } lower < x < upper, \\
+            1-\text{CDF}(upper, dist) & \text {for} x = upper, \\
+            0 & \text{for } x > upper,
+        \end{cases}
+
+
+    Parameters
+    ----------
+    dist: PyMC unnamed distribution
+        PyMC distribution created via the `.dist()` API, which will be censored. This
+        distribution must be univariate and have a logcdf method implemented.
+    lower: float or None
+        Lower (left) censoring point. If `None` the distribution will not be left censored
+    upper: float or None
+        Upper (right) censoring point. If `None`, the distribution will not be right censored.
+
+
+    Examples
+    --------
+    .. code-block:: python
+
+        with pm.Model():
+            normal_dist = pm.Normal.dist(mu=0.0, sigma=1.0)
+            censored_normal = pm.Censored("censored_normal", normal_dist, lower=-1, upper=1)
+    """
+
+    @classmethod
+    def dist(cls, dist, lower, upper, **kwargs):
+        if not isinstance(dist, TensorVariable) or not isinstance(dist.owner.op, RandomVariable):
+            raise ValueError(
+                f"Censoring dist must be a distribution created via the `.dist()` API, got {type(dist)}"
+            )
+        if dist.owner.op.ndim_supp > 0:
+            raise NotImplementedError(
+                "Censoring of multivariate distributions has not been implemented yet"
+            )
+        check_dist_not_registered(dist)
+        return super().dist([dist, lower, upper], **kwargs)
+
+    @classmethod
+    def rv_op(cls, dist, lower=None, upper=None, size=None, rngs=None):
+        if lower is None:
+            lower = at.constant(-np.inf)
+        if upper is None:
+            upper = at.constant(np.inf)
+
+        # Censoring is achieved by clipping the base distribution between lower and upper
+        rv_out = at.clip(dist, lower, upper)
+
+        # Reference nodes to facilitate identification in other classmethods, without
+        # worring about possible dimshuffles
+        rv_out.tag.dist = dist
+        rv_out.tag.lower = lower
+        rv_out.tag.upper = upper
+
+        if size is not None:
+            rv_out = cls.change_size(rv_out, size)
+        if rngs is not None:
+            rv_out = cls.change_rngs(rv_out, rngs)
+
+        return rv_out
+
+    @classmethod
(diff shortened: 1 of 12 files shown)


pymc/distributions/censored.py
```diff
@@ -0,0 +1,99 @@
...
+        # lower / upper may have broadcasted the distribution, in which case we get a Dimshuffle Op
+        if isinstance(dist.owner.op, DimShuffle):
+            dist_node = dist_node.inputs[0].owner
+        olg_rng, size, dtype, *dist_params = dist_node.inputs
+        new_dist = dist_node.op.make_node(new_rng, size, dtype, *dist_params).default_output()
+        return cls.rv_op(new_dist, lower, upper)
+
+    @classmethod
+    def graph_rvs(cls, dist, *bounds):
+        return (dist,)
```
ricardoV94: Check for DimShuffle here

pymc/distributions/distribution.py
```diff
@@ -354,11 +355,241 @@ def dist(
...
+        dims: Optional[Dims] = None,
+        initval=None,
+        observed=None,
+        total_size=None,
+        transform=UNSET,
+        **kwargs,
+    ) -> TensorVariable:
+        """Adds a TensorVariable corresponding to a PyMC derived distribution to the current model.
+
+        Note that all remaining kwargs must be compatible with ``.dist()``
```
brandonwillard: A description/definition of a "derived distribution" might be necessary here.

brandonwillard (dismissed):

twiecki:
@ricardoV94 This is becoming a blocker for time series, how much is left to do here?

ricardoV94:
I am getting back to this one this week!

ricardoV94:
In the meantime, suggestions for the base class name? `DerivedDistribution` was my first suggestion because `Aeppl` derives the logprob terms for these graphs, but there is not much "derivation" going on in PyMC. These are just symbolic graphs that somewhere have a RandomVariable as a node / input. Other ideas:
1. `CompositeDistribution`
2. `SymbolicDistribution`
3. ... ?

twiecki:
`CustomDistribution`? Otherwise I like `SymbolicDistribution`.

brandonwillard:
What purpose are these `*Distribution` classes serving?  Remember, the original `Distribution` class in v4 was just an all static method-based stub/placeholder for (temporary) backwards compatibility.

ricardoV94:
They serve the same goal the "standard" Distribution class is doing right now: basically manage all non distribution kwargs: shape/dims, transformed, observed; and behind the scenes magic things like value variables and rngs

brandonwillard:
> They serve the same goal the "standard" Distribution class is doing right now: basically manage all non distribution kwargs: shape/dims, transformed, observed; and behind the scenes magic things like value variables and rngs

The base `Distribution` class was only ever supposed to make the hand-off to the `Model` class, which would then handle all the generic tasks (e.g. value variables, RNGs, shape/dims, etc.), and the rest was just for backward compatibility (e.g. the now completely unnecessary `Disribution.dist` interface).  Notice how instances of these classes are never made, and that only static/class/type-level functions are used.

By extending `Distribution` and adding more logic to those classes, it could become considerably more difficult to unwind those temporary backward compatibility-only choices, and they'll eventually become a permanent and confusing part of v4's design.

This isn't the time or place to start addressing all that, but these are the kinds of design considerations that need to go alongside changes/additions to the relevant areas of code (i.e. `Distribution`-related code).

ricardoV94:
> By extending Distribution and adding more logic to those classes, it could become considerably more difficult to unwind those temporary backward compatibility-only choices, and they'll eventually become a permanent and confusing part of v4's design.

Yeah, we need ideas of how to get rid of the base distribution if that's desirable, but that is not the goal of this PR.

I created a bit of a replica of that base class (and did not modify or extended it) so hopefully whatever works for standard distributions will also work for these ones and won't be made more difficult because of the changes introduced in this PR.

pymc/distributions/distribution.py
```diff
@@ -354,11 +355,266 @@ def dist(
...
+            Performs input validation and converts optional alternative parametrizations
+            to a canonical parametrization. It should call `super().dist()` and pass
+            a list with the default parameters as the first and only non keyword argument,
+            followed by other keyword arguments like size and rngs.
+        cls.rv_op
+            Returns a TensorVariable that represents the symbolic distribution
+            parametrized by a default set of parameters and a size and rngs arguments
+        cls.ndim_supp
+            Returns the support of the symbolic distribution, given the default
+            parameters.This may not always be constant, for instance if the symbolic
```
twiecki:
```suggestion
            parameters. This may not always be constant, for instance if the symbolic
```

pymc/distributions/distribution.py
```diff
@@ -354,11 +355,266 @@ def dist(
...
+            See ``Model.register_rv``.
+        **kwargs
+            Keyword arguments that will be forwarded to ``.dist()``.
+            Most prominently: ``shape`` and ``size``
+
+        Returns
+        -------
+        var : TensorVariable
+            The created variable, registered in the Model.
+        """
```
twiecki, line 434: Doc string is super clear, a short example would really drive the point home.

twiecki: Alternatively can point to `Censored`.

ricardoV94: Added a reference to Censored

twiecki (approved):

ricardoV94:
@brandonwillard do you have other pointers for this PR? It is currently blocked by your old requested changes concerning docstrings

brandonwillard (commented):
> @brandonwillard do you have other pointers for this PR? It is currently blocked by your old requested changes concerning docstrings

Sorry, I dismissed the old review; otherwise, I don't have any more input for this PR right now.

ghost:
Thanks @ricardoV94! Can't wait to try these out.

pymc/distributions/distribution.py
```diff
@@ -354,11 +355,268 @@ def dist(
         return rv_out
 
 
+class SymbolicDistribution:
```
twiecki, line 358: Will those show up in the docs?

twiecki: We also didn't mention it in the release notes.

ricardoV94: They show in the docs (you can check the RTD job in this PR). I didn't see the SymbolicDist as being user facing, that's why it's not mentioned in the release notes.

twiecki: I agree, but it's such an important new addition that I think it deserves some more attention.

PR #5173, opened 2021-11-11
Add MvStudentT and MatrixNormal moment

patel-zeel:
Add moments and tests for the below distributions as part of #5078

* ```pymc.distributions.multivariate.MvStudentT```
* ```pymc.distributions.multivariate.MatrixNormal```


pymc/distributions/multivariate.py
```diff
@@ -343,6 +347,14 @@ def dist(cls, nu, Sigma=None, mu=None, cov=None, tau=None, chol=None, lower=True
         assert_negative_support(nu, "nu", "MvStudentT")
         return super().dist([nu, mu, cov], **kwargs)
 
+    def get_moment(rv, size, nu, mu, cov):
+        moment = mu
+        if not rv_size_is_none(size):
+            if isinstance(size, int):
+                size = (size,)
+            moment = at.full((*size, mu.size), moment)
+        return moment
```
Sayam753: The logic here can be simplified by exactly following how `get_moment` method is added for MvNormal distribution in this PR -  https://github.com/pymc-devs/pymc/pull/5171/files.

pymc/tests/test_distributions_moments.py
```diff
@@ -612,3 +614,37 @@ def test_discrete_uniform_moment(lower, upper, size, expected):
...
+rand2d = np.random.rand(2, 3)
+
+
+@pytest.mark.parametrize(
+    "mu, rowchol, colchol, expected",
+    [
+        (np.ones((1, 1)), np.eye(1), np.eye(1), np.ones((1, 1))),
+        (rand2d, np.eye(2), np.eye(3), rand2d),
+    ],
+)
```
Sayam753: size argument is missing here.

Sayam753 (changes requested):

pymc/distributions/multivariate.py
```diff
@@ -1707,6 +1719,14 @@ def dist(
 
         return super().dist([mu, rowchol_cov, colchol_cov], **kwargs)
 
+    def get_moment(rv, size, mu, rowchol, colchol):
+        moment = mu
+        if not rv_size_is_none(size):
+            if isinstance(size, int):
+                size = (size,)
+            moment = at.full((*size, mu.size), moment)
+        return moment
```
Sayam753: `mu` also needs to be broadcasted to match `rowchol` and `cholchol` dimensions. We can reuse the logic to broadcast `mu` [implemented in `MatrixNormalRV.rng_fn` method](https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/multivariate.py#L1548-L1553).

pymc/tests/test_distributions_moments.py
patel-zeel: @Sayam753 distribution does not accept size argument. Is it safe to remove the following restriction?
```python
if kwargs.get("size", None) is not None:
    raise NotImplementedError("MatrixNormal doesn't support size argument")
```

pymc/distributions/multivariate.py
patel-zeel: Yes, you are right @Sayam753. Applied this change.

pymc/distributions/multivariate.py
patel-zeel: Fixed.

pymc/distributions/multivariate.py
```diff
@@ -1656,8 +1667,8 @@ def dist(
 
         cholesky = Cholesky(lower=True, on_error="raise")
 
-        if kwargs.get("size", None) is not None:
-            raise NotImplementedError("MatrixNormal doesn't support size argument")
+        # if kwargs.get("size", None) is not None:
```
ricardoV94: We can't to this change. If I remember correctly either our logp or rng_fn do not (yet) accomodate non-empty sizes. That's why that check was there.

On the other hand, that also means that your `get_moment` does not need to worry about it for the time being (perhaps we can just raise a ValieError of size is not None for now).

We should probably move this restriction to the RV itself, but that's a separate matter. CC @kc611, he might correct me here or give more context

ricardoV94 (changes requested):

pymc/distributions/multivariate.py
patel-zeel: Done.

pymc/tests/test_distributions_moments.py
```diff
@@ -764,3 +766,48 @@ def test_moyal_moment(mu, sigma, size, expected):
...
+rand2d = np.random.rand(2, 3)
+
+
+@pytest.mark.parametrize(
+    "nu, mu, cov, size, expected",
+    [
+        (2, np.ones(1), np.eye(1), None, np.ones(1)),
+        (2, rand1d, np.eye(2), None, rand1d),
+        (2, rand1d, np.eye(2), 2, np.full((2, 2), rand1d)),
+        (2, rand2d, np.eye(3), 2, np.full((2, 2, 3), rand2d)),
```
ricardoV94: This test combination seems to be failing, because the moment is ignoring broadcasting induced by `nu`

```suggestion
        (2, rand2d, np.eye(3), 2, np.full((2, 2, 3), rand2d)),
        (np.array([3, 4]), np.ones(2), np.eye(2), None, np.ones((2, 2))),
```

ricardoV94 (changes requested):
Found an issue with the MvStudentT moment

pymc/tests/test_distributions_moments.py
patel-zeel: Tried to fix it. I suppose `nu` and `size` are not going well together. Is the following test correct? if not why?

`(np.array([3, 4]), np.ones(2), np.eye(2), 2, np.ones((2, 2, 2))),`

ricardoV94: @Sayam753 Can you confirm this is the expected behavior with nu broadcasting?

Sayam753: I am first time seeing `nu` not being a scaler. And this is not so straight-forward -

In the case -
```
(np.array([3, 4]), np.ones(2), np.eye(2), 2, np.ones((2, 2, 2)))
size = 2
nu = np.array([3, 4])
mean = np.ones(2)
```

Due to the dimensions of `nu`, we form two independent (batched) MvStudentT distribution. Each random sample from this distribution will be of shape (2, 2). First dimension is implied by `nu` and second by `cov.shape[1]`
Then `size` comes into play and we thus have the shape (2, 2, 2).

<hr>

For the case,

```
size = 2
nu = np.array([3, 4])
mean = np.ones((2, 2))
```

The resulting shape would still be (2, 2, 2) not (2, 2, 2, 2). This is because `mean.shape[:-1]` and `nu.shape` broadcasts with each other. This case is not handled with the current implementation of `get_moment` method.

Sayam753: Some pseudo code to fix the same would be -
```
event_shape = cov.shape[-1:]
batch_shape = broadcast_shapes(nu.shape, mu.shape[:-1])
output_shape = size + batch_shape + event_shape
at.fill(output_shape, mu)
```

patel-zeel: Thanks for the suggestions @Sayam753 and @ricardoV94. I was trying these tests but `result` and `expected` are not matching for these tests. 
For example, if test is:
`(np.array([3, 4]), np.ones(2), np.eye(2), 2, np.ones((2, 2, 2)))`
then in `assert_moment_is_expected`:
```python
result = fn(0)["x"]
result.shape = (2,2)
expected.shape = (2,2,2)
```

I suspect that to enable `nu` broadcasting behaviour, something needs to be changed in the distribution because as per the docs, `MvStudentT` is expecting `nu` as a `float`.
```
Parameters
    ----------
    nu: float
        Degrees of freedom, should be a positive scalar.
    Sigma: matrix
        Covariance matrix. Use `cov` in new code.
    mu: array
        Vector of means.
    cov: matrix
        The covariance matrix.
```

patel-zeel: Thanks for the suggestions @Sayam753 and @ricardoV94. I was trying these tests but `random_draw` and `expected` are not matching for these tests. 
For example, if test is:
`(np.array([3, 4]), np.ones(2), np.eye(2), 2, np.ones((2, 2, 2)))`
then in `assert_moment_is_expected`:
```python
result = fn(0)["x"]
random_draw = model["x"].eval()
result.shape = (2,2,2)
expected.shape = (2,2,2)
random_draw.shape = (2,2)
```

I suspect that to enable `nu` broadcasting behaviour, something needs to be changed in the distribution because as per the docs, `MvStudentT` is expecting `nu` as a `float`.
```
Parameters
    ----------
    nu: float
        Degrees of freedom, should be a positive scalar.
    Sigma: matrix
        Covariance matrix. Use `cov` in new code.
    mu: array
        Vector of means.
    cov: matrix
        The covariance matrix.
```

Sayam753: Yes. Also as per `ndim_params` parameter of MvStudentT here -

https://github.com/pymc-devs/pymc/blob/90423a9c7444b2b5ea2b692921d2ea657d1db087/pymc/distributions/multivariate.py#L264

`nu` parameter is expected to be a scaler. But if there any extra dimensions in `nu`, it should imply batch dimensions and `rng_fn` method should work out of the box. I do not know how aesara handles broadcasting logic, so cc @ricardoV94 or @brandonwillard.

ricardoV94: > it should imply batch dimensions and rng_fn method should work out of the box

@Sayam753, does this mean the `rng_fn` is working out of the box? We can tweak the shape inference if that's not giving the right results. I am not sure, but I think Aesara will not broadcast any parameters before calling `rng_fn`, broadcasting is only done internally when trying to infer the shape of the RV, but I would need to double check.

If neither the `rng_fn` nor the Aesara shape inference are working for vector `nu`, we can also force it to be a scalar in `make_node` and raise otherwise. 

Also we would need to check if the `logp`/`logcdf` behave properly with vector `nu`, so maybe forcing it to be a scalar is the safest course of action for now? We can open an issue to investigate things can/are working properly with vector `nu`

Sayam753: > Does this mean the rng_fn is working out of the box?

Yes. As per @patel-zeel comment two messages above, shape from random samples does not match the shape obtained from initial moments. 

> so maybe forcing it to be a scalar is the safest course of action for now?

Agree. @patel-zeel please revert the `nu` broadcasting logic. And lets only deal with `mu` for moments this time.

ricardoV94: > > Does this mean the rng_fn is working out of the box?
> 
> Yes. As per @patel-zeel comment two messages above, shape from random samples does not match the shape obtained from initial moments.

Just to be clear, I don't think we should limit the MvStudentT just because the `get_moment` is not giving the right shape. We could limit it on the grounds that `rv.eval().shape` disagrees with `rv.shape.eval()` or that our `logp`/`logcdf` do not work properly with vector `nu`.

ricardoV94: The shape inference does seem to fail:

```python
x = pm.MvStudentT.dist(nu=[3, 4], mu=np.ones(2), cov=np.eye(2), size=2)
tuple(x.shape.eval()) == tuple(x.eval().shape)  # False
```

ricardoV94:
So we don't want to support vector `nu` (we should open an issue to make sure we enforce that in MvStudentT dist or RV), but what about a 3D sigma? 

Is that allowed by the RV? ~I think we have tests for this case in the MvNormal moment.~ No we don't

ricardoV94 (approved):

pymc/tests/test_distributions_moments.py
patel-zeel: Thanks for the suggestions @ricardoV94 and @Sayam753. I have disabled `nu` broadcasting for now. I hope I am covering everything else with the current test cases.

patel-zeel:
That was an interesting thought @ricardoV94. But I think `cov` is enforced to be 2d in `quaddist_matrix`. Thanks for approving the PR :)

ricardoV94:
> That was an interesting thought @ricardoV94. But I think `cov` is enforced to be 2d in `quaddist_matrix`. Thanks for approving the PR :)

Yeah, we might want to revisit that limitation another time. Thanks a lot for your help. I'll leave some time in case @Sayam753 wants to have a look before merging

patel-zeel:
Sure. No problem at all. It was a good learning experience for me.

Sayam753 (approved):
LGTM 🚀

PR #5174, opened 2021-11-11
Adding Dirichlet moment and tests

larryshamalama:
This pull request adds `get_moment` to the `Dirichlet` class and corresponding tests.

Regarding tests, my third test fails as I am unsure if I am understanding broadcasting correctly. Hoping to add at least a fourth test upon passing the third one

CC: @ricardoV94

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -46,7 +46,11 @@ from pymc.distributions import transforms
 from pymc.distributions.continuous import ChiSquared, Normal, assert_negative_support
 from pymc.distributions.dist_math import bound, factln, logpow, multigammaln
 from pymc.distributions.distribution import Continuous, Discrete
-from pymc.distributions.shape_utils import broadcast_dist_samples_to, rv_size_is_none, to_tuple
+from pymc.distributions.shape_utils import (
+    broadcast_dist_samples_to,
+    rv_size_is_none,
+    to_tuple,
+)
 from pymc.math import kron_diag, kron_dot
 
 __all__ = [
@@ -407,9 +411,11 @@ class Dirichlet(Continuous):
 
     def get_moment(rv, size, a):
         norm_constant = at.sum(a, axis=-1)[..., None]
-        moment = a/norm_constant
+        moment = a / norm_constant
         if not rv_size_is_none(size):
-            return at.full(size, moment)
+            if isinstance(size, int):
+                size = (size,)
+            moment = at.full((*size, *a.shape), moment)
         return moment
 
     def logp(value, a):
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -46,7 +46,7 @@ from pymc.distributions import transforms
 from pymc.distributions.continuous import ChiSquared, Normal, assert_negative_support
 from pymc.distributions.dist_math import bound, factln, logpow, multigammaln
 from pymc.distributions.distribution import Continuous, Discrete
-from pymc.distributions.shape_utils import broadcast_dist_samples_to, to_tuple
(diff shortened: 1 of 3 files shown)


pymc/tests/test_distributions_moments.py
```diff
@@ -611,4 +612,28 @@ def test_hyper_geometric_moment(N, k, n, size, expected):
...
+        (
+            np.array([[1, 2, 3], [5, 6, 7]]),
+            None,
+            np.array([[1, 2, 3], [5, 6, 7]])/np.array([6, 18])[..., np.newaxis],
+        ),
+        (
+            np.full(shape=np.array([7, 3]), fill_value=np.array([13, 17, 19])),
+            (11, 5,),
+            np.broadcast_to([13, 17, 19], shape=[11, 5, 7, 3]),
+        ),
```
larryshamalama, line 646: Hi @ricardoV94 , the third test here fails with error `ValueError: could not broadcast input array from shape (7,3) into shape (11,5)`. Is this expected?  If I put shape=`(11, 5, 7, 3)` instead of `(11, 5)`, my output shape is `(11, 5, 7, 3, 7, 3)` and I don't know how to write this test correctly

larryshamalama (commented):
Comment on third moment test for `pm.Dirichlet`

pymc/tests/test_distributions_moments.py
ricardoV94: I think that means you cannot simply use these lines to manage size information. That logic is only that straightforward for univariate distributions

```
if not rv_size_is_none(size):
    return at.full(size, moment)
```

larryshamalama: Got it, thanks for the pointer. I believe that I was able to fix this

pymc/tests/test_distributions_moments.py
```diff
@@ -611,4 +612,41 @@ def test_hyper_geometric_moment(N, k, n, size, expected):
...
+                11,
+                5,
+            ),
+            np.broadcast_to([13, 17, 19], shape=[11, 5, 7, 3]) / 49,
+        ),
+    ],
+)
+def test_dirichlet_moment(a, size, expected):
+    with Model() as model:
+        Dirichlet("x", a=a, size=size)
```
ricardoV94, line 651: Looks great Larry! Thanks a lot

ricardoV94 (approved):

ricardoV94:
Looks great Larry! Thanks a lot

ricardoV94:
Would be great to have your help for some of the other multivariate distributions

pymc/distributions/multivariate.py
```diff
@@ -405,6 +409,15 @@ def dist(cls, a, **kwargs):
 
         return super().dist([a], **kwargs)
 
+    def get_moment(rv, size, a):
+        norm_constant = at.sum(a, axis=-1)[..., None]
+        moment = a / norm_constant
+        if not rv_size_is_none(size):
+            if isinstance(size, int):
```
ricardoV94, line 416: Actually we may need to revisit this. When size is a number it is still inside a TensorConstant. Not sure what `isinstance(size, int)` will do. Maybe we should check for `size.ndim`?

Size may also be a symbolic shape. So the same reasoning applies.

larryshamalama: > Looks great Larry! Thanks a lot

As always, thanks to you for all your help.

Should we change `isinstance(size, int)` to `if isinstance(size, int) or size.ndim > 0`? Would that be a correct approach?

ricardoV94: I checked again and I think that checks will never evaluate true. Even when size is an integer, it gets converted to a `TensorConstant([size])`:

```python
size = x.owner.inputs[1]
isinstance(size, int)  # False
size.data  # array([2])
```

If I remove the `isinstance` block, all tests still pass

PR #5175, opened 2021-11-11
adds beta-binomial mean and test cases

morganstrom:
Note that the get_moment method returns an integer value (rounding the mean down to the nearest integer) in this implementation. Probably due to multiplying an integer with floats.. 

This is made clear in two of the test cases, where np.floor is applied to the expected output. This is similar to the implementation of the Binomial mean (also a integer multiplied with a float, and where the test cases are also rounded to integer values) so I figured it would be OK. Let me know otherwise!

### DIFF
--- a/pymc/tests/test_distributions_moments.py
+++ b/pymc/tests/test_distributions_moments.py
@@ -6,6 +6,7 @@ from scipy import special
 from pymc.distributions import (
     Bernoulli,
     Beta,
+    BetaBinomial,
     Binomial,
     Cauchy,
     ChiSquared,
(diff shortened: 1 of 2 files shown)


pymc/tests/test_distributions_moments.py
```diff
@@ -208,6 +209,21 @@ def test_beta_moment(alpha, beta, size, expected):
     assert_moment_is_expected(model, expected)
 
 
+@pytest.mark.parametrize(
+    "n, alpha, beta, size, expected",
+    [
+        (10, 1, 1, None, 5),
+        (10, 1, 1, 5, np.full(5, 5)),
+        (10, 1, np.arange(1, 6), None, np.floor(10 / np.arange(2, 7))),
+        (10, 1, np.arange(1, 6), (2, 5), np.full((2, 5), np.floor(10 / np.arange(2, 7)))),
```
ricardoV94: Why is the floor needed in the expected results?

ricardoV94 (commented):
Thanks for opening the PR. I have a question below

pymc/tests/test_distributions_moments.py
morganstrom: I mentioned this in the first comment of the pull request - the get_moment method returns an integer for some reason, while the actual mean should be a float (according to the scipy.stats.betabinom). I got the impression from other pymc distributions (binomial for example) that get_moment should return an integer anyway.. let me know if this was incorrect!

ricardoV94: I see. I forgot the binomial part of the beta-binomial xD. Then we should add a floor or round (whichever makes more sense) as an explicit operation in the `get_moment` as well

morganstrom: Noted!

ricardoV94:
Seems like you picked some of the main commits by accident during your merge. Can you try to rebase to fix it?

morganstrom:
I don't know how that happened but I think I fixed it now!

ricardoV94 (approved):

ricardoV94:
Waiting to see if tests pass

PR #5176, opened 2021-11-12
add categorical moment

farhanreynaldo:
Add moments and tests for the below distributions as part of https://github.com/pymc-devs/pymc/issues/5078:

* pymc.distributions.discrete.Categorical

### DIFF
--- a/pymc/tests/test_distributions_moments.py
+++ b/pymc/tests/test_distributions_moments.py
@@ -618,13 +618,13 @@ def test_discrete_uniform_moment(lower, upper, size, expected):
 @pytest.mark.parametrize(
     "p, size, expected",
     [
-        (np.arange(0.1, 0.4, 0.1), None, 3),
-        (np.arange(0.1, 0.4, 0.1), 5, np.full(5, 3)),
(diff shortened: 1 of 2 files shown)


pymc/tests/test_distributions_moments.py
```diff
@@ -612,3 +613,22 @@ def test_discrete_uniform_moment(lower, upper, size, expected):
     with Model() as model:
         DiscreteUniform("x", lower=lower, upper=upper, size=size)
     assert_moment_is_expected(model, expected)
+
+
+@pytest.mark.parametrize(
+    "p, size, expected",
+    [
+        (np.arange(0.1, 0.4, 0.1), None, 3),
```
ricardoV94: Looks good but can we perhaps have some of the tests have the moment not be the last element?

Also for readability defining the p explicitly such as `np.array([.6, .1, .3])` seems better.

farhanreynaldo: Ah sure, I agree it is not really readable.

ricardoV94 (approved):

ricardoV94:
Looks good, just checking if the tests pass

PR #5178, opened 2021-11-12
Avoid unnecessary double clone_replace in `smc._logp_forw`

ricardoV94:
This is a tiny optimization. The first time I wrote this in https://github.com/pymc-devs/pymc/pull/4898 I introduced two clone_replaces, but a single one is enough.


aloctavodia (approved):
LGTM

Issue #5100, opened 2021-10-25
`sample_numpyro_nuts` does not compute `pm.Deterministic` variables


ricardoV94:
We have simply not implemented such feature...

PR #5182, opened 2021-11-14
Adding Deterministic for sampling_numpyro

zaxtax:
This addresses #5100 and passes tests. My only concern is that this is likely not the fastest way to get deterministic variables as iterating per sample and per chain in pure python takes away much of the speed of this method.

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -168,10 +168,10 @@ def sample_numpyro_nuts(
 
     for i in range(draws):
         for c in range(chains):
-            draw = dict(
-                (value_var.name, raw_samples[c, i])
+            draw = {
+                value_var.name: raw_samples[c, i]
                 for value_var, raw_samples in zip(model.value_vars, raw_mcmc_samples)
-            )
+            }
             sample = fn(draw)
             for vi, v in enumerate(vars_to_sample):
                 mcmc_samples[v.name].append(sample[vi])
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -116,7 +116,6 @@ def sample_numpyro_nuts(
     vars_to_sample = list(get_default_varnames(var_names, include_transformed=keep_untransformed))
     inputs = [model.rvs_to_values[i] for i in model.free_RVs]
 
-
     tic1 = pd.Timestamp.now()
     print("Compiling...", file=sys.stdout)
 
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -27,6 +27,7 @@ from aesara.link.jax.dispatch import jax_funcify
 
 from pymc import Model, modelcontext
 from pymc.aesaraf import compile_rv_inplace
+from pymc.util import get_default_varnames
 
 warnings.warn("This module is experimental.")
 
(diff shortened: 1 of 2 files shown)


twiecki:
Thanks @zaxtax! Yeah, the looping is pretty unfortunate and I think will be too slow. Wasn't there a PR that already added that? I vaguely remember something about it.

twiecki:
https://github.com/pymc-devs/pymc/pull/4427

zaxtax:
Good call!

pymc/tests/test_sampling_jax.py
```diff
@@ -44,6 +44,23 @@ def test_transform_samples():
...
+    obs_at = aesara.shared(obs, borrow=True, name="obs")
+    with pm.Model() as model:
+        a = pm.Normal("a", 0, 1)
+        b = pm.Deterministic("b", a / 2.0)
+        c = pm.Normal("c", a, sigma=1.0, observed=obs_at)
+
+        trace = sample_numpyro_nuts(chains=2, random_seed=1322, keep_untransformed=True)
+
+    assert 8 < trace.posterior["a"].mean() < 11
+    assert 4 < trace.posterior["b"].mean() < 6
```
ricardoV94: This can be a much more strict: 
```suggestion
    assert np.allclose(trace.posterior["b"].values, trace.posterior["a"].value / 2)
```

pymc/tests/test_sampling_jax.py
```diff
@@ -44,6 +44,23 @@ def test_transform_samples():
...
 
+def test_deterministic_samples():
+    aesara.config.on_opt_error = "raise"
+    np.random.seed(13244)
+
+    obs = np.random.normal(10, 2, size=100)
+    obs_at = aesara.shared(obs, borrow=True, name="obs")
+    with pm.Model() as model:
+        a = pm.Normal("a", 0, 1)
+        b = pm.Deterministic("b", a / 2.0)
```
ricardoV94, line 55: Let's (also) add a deterministic of a transformed variable? That's often where we find problems.

ricardoV94:
I left two comments about the tests. I agree with @twiecki that we probably want to find a better solution, if there is one.

pymc/sampling_jax.py
```diff
@@ -116,6 +123,7 @@ def sample_numpyro_nuts(
     init_state_batched = jax.tree_map(lambda x: np.repeat(x[None, ...], chains, axis=0), init_state)
 
     logp_fn = get_jaxified_logp(model)
+    fn = model.fastfn(vars_to_sample)
```
ricardoV94: Can you give a more informative name and/or a comment that explains what the purpose of this `fn` will be?

pymc/tests/test_sampling_jax.py
zaxtax: I believe `a` does get transformed here. But maybe we can add a check around that to this test.

ricardoV94: By transform, I mean a variable like `s = pm.Lognormal()`, which has a `logTransform` automatically under the hood. I am not referring to the `pm.Deterministic` transform

zaxtax: Sorry I forgot that for this test I draw `a` from a Normal and not a Uniform which has a transform under the hood. I'll add something here.

zaxtax:
I think I've addressed all concerns.

pymc/tests/test_sampling_jax.py
ricardoV94: No problem! The test looks great now

zaxtax:
Is there any way to make the code coverage failure go away? The lines it's complaining out are definitely run by the tests.

ricardoV94:
The results of `test_sampling_jax` are not added to the coverage report. You can ignore the "failure"

twiecki:
Thanks!

PR #5185, opened 2021-11-15
Add InverseGamma moment

morganstrom:
Simplifies calculation using aesara.tensor.switch statement


pymc/distributions/continuous.py
```diff
@@ -2326,23 +2326,17 @@ def dist(cls, alpha=None, beta=None, mu=None, sigma=None, sd=None, *args, **kwar
...
         assert_negative_support(alpha, "alpha", "InverseGamma")
         assert_negative_support(beta, "beta", "InverseGamma")
 
         return super().dist([alpha, beta], **kwargs)
 
+    def get_moment(rv, size, alpha, beta):
+        m = beta / (alpha - 1.0)
+        if not rv_size_is_none(size):
+            m = at.full(size, m)
+        return at.switch(alpha > 1, m, np.inf)
```
ricardoV94: We should either use the mode as the moment or switch between the mode and the mean, so that we don't have to return an infinite value. The goal of `get_moment` is to obtain a value that is always okay as a starting point for that distribution.

morganstrom: Okay, I think I fixed it now! Let me know if I need to make further changes

pymc/tests/test_distributions_moments.py
```diff
@@ -385,6 +386,26 @@ def test_gamma_moment(alpha, beta, size, expected):
...
+        (1, 1, None, 0.5),
+        (5, 1, 5, np.full(5, 1 / (5 - 1))),
+        (np.arange(2, 7), 1, None, 1 / np.arange(1, 6)),
+        (np.arange(2, 7), 1, 5, 1 / np.arange(1, 6)),
+        (2, np.arange(1, 6), None, np.arange(1, 6)),
+        (np.arange(2, 7), np.arange(1, 6), 5, np.ones(5)),
+        (np.arange(1, 6), 1, None, np.array([0.5, 1, 1 / 2, 1 / 3, 1 / 4])),
+    ],
+)
+def test_inverse_gamma_moment(alpha, beta, size, expected):
```
ricardoV94, line 410: Can we reduce a bit the number of tests?

morganstrom: Better now?

morganstrom:
I did `git fetch upstream` and `git rebase upstream/main`, as recommended here: https://github.com/pymc-devs/pymc/blob/35f9966f15e76d8ba304914be69e4115fdcfd4ef/CONTRIBUTING.md 

Is that why there are so many commits in this pull request suddenly? 

Don't know how to solve it easily, should I just create a brand new branch and add my changes there perhaps?

PR #5170, opened 2021-11-11
Fix `make_initial_point_expression` recursion bug

ricardoV94:
Fixes #5168

PR #5186, opened 2021-11-15
Fix `rvs_to_value_vars` inplace update bug

### LINKED ISSUE #5172: Calling `rvs_to_value_vars` modifies `model.free_RVs`
This error was found while working on #5170:

```python
with pm.Model() as pmodel:
    hyper = pm.LogNormal("hyper", mu=0)
    group = pm.LogNormal("group", mu=at.log(hyper))

    # We add potentials or deterministics that are not in topological order
    pm.Potential('group_pot', group)
    pm.Potential("hyper_pot", hyper)

    before = len(FunctionGraph(outputs=pmodel.free_RVs).toposort())

    # This call will change the model free_RVs in place!
    # rvs_to_value_vars(pmodel.deterministics, apply_transforms=True)
    rvs_to_value_vars(pmodel.potentials, apply_transforms=True)

    after = len(FunctionGraph(outputs=pmodel.free_RVs).toposort())

    assert before == after
```
```
  File "/home/ricardo/Documents/Projects/pymc3-venv/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3418, in run_code
  File "<ipython-input-9-66cc9d4cc5c8>", line 37, in <module>
    assert before == after
AssertionError
```

The same happens with the `Deterministics`, and can be checked by flipping the commented lines and commenting those referring to `Potentials`

* PyMC/PyMC3 Version: `main`
* Aesara/Theano Version: `2.2.6`

ricardoV94:
Fixes #5172

### DIFF
--- a/pymc/tests/test_aesaraf.py
+++ b/pymc/tests/test_aesaraf.py
@@ -23,7 +23,7 @@ import pandas as pd
 import pytest
 import scipy.sparse as sps
 
-from aesara.graph.basic import Constant, Variable, ancestors
+from aesara.graph.basic import Constant, Variable, ancestors, equal_computations
 from aesara.tensor.random.basic import normal, uniform
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.subtensor import AdvancedIncSubtensor, AdvancedIncSubtensor1
@@ -529,3 +529,24 @@ def test_rvs_to_value_vars():
     assert a_value_var in res_ancestors
     assert b_value_var in res_ancestors
     assert c_value_var in res_ancestors
+
+
+def test_rvs_to_value_vars_nested():
+    # Test that calling rvs_to_value_vars in models with nested transformations
+    # does not change the original rvs in place. See issue #5172
+    with pm.Model() as m:
+        one = pm.LogNormal("one", mu=0)
+        two = pm.LogNormal("two", mu=at.log(one))
+
+        # We add potentials or deterministics that are not in topological order
+        pm.Potential("two_pot", two)
+        pm.Potential("one_pot", one)
+
+        before = aesara.clone_replace(m.free_RVs)
+
+        # This call would change the model free_RVs in place in #5172
+        res, _ = rvs_to_value_vars(m.potentials, apply_transforms=True)
+
+        after = aesara.clone_replace(m.free_RVs)
+
+        assert equal_computations(before, after)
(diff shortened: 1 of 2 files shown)


pymc/tests/test_aesaraf.py
```diff
@@ -529,3 +529,23 @@ def test_rvs_to_value_vars():
     assert a_value_var in res_ancestors
     assert b_value_var in res_ancestors
     assert c_value_var in res_ancestors
+
+
+def test_rvs_to_value_vars_nested_transform_bug():
```
brandonwillard: It's better to name a test after the things being tested, instead of a reference to a previous issue/bug that likely won't hold up over time.  In general, a developer seeing this in the future shouldn't need to know/read old issues in order to understand what's being tested; this applies to test names, as well.

pymc/tests/test_aesaraf.py
```diff
@@ -529,3 +529,23 @@ def test_rvs_to_value_vars():
...
+        pm.Potential("one_pot", one)
+
+        before = aesara.clone_replace(m.free_RVs)
+
+        # This call would change the model free_RVs in place
+        res, _ = rvs_to_value_vars(m.potentials, apply_transforms=True)
+
+        after = aesara.clone_replace(m.free_RVs)
+
+        assert equal_computations(before, after)
```
michaelosthege, line 552: I appreciate testing a reduced, more focused scenario, but it would be more reassuring to also test the full scenario that was initially observed.
That is doing `m.recompute_initial_point()` before and after a call to `pm.sample(chains=1, tune=1, draws=1)`.

It looks like you now understood much more of what's going on under the hood than I do, but when thinking and chatting about this issue last week we were also concerned about needing to maintain the `free_RVs` identity.
For example because they're keys in dicts like `m.initial_values`, or used as args to step methods. As far as I can tell the assert above checks only computation, not the identity?

ricardoV94: The identity was not changed. It was the inputs of (some of) free.RVs that were changed. This checks their computations are the same, which is a more strict check than what we were doing with the toposort (just checking if the number of nodes in the graph are the same). 

I can also check that the set of `graph_inputs` is exactly the same. That's what `recompute_initial_point` was complaining about.

Regarding the more complex test, I did check it locally and it's fixed as well. I prefer to have an example that is easier to understand. The case you found had a lot of moving parts that did not matter to raise the error. In the end it was just the nested LogTransform out of topological order, similar to #5170

ricardoV94: Actually it seems `equal_computations` also makes sure the input variables are the same.

michaelosthege (approved):

PR #5187, opened 2021-11-15
Remove logcdfpt and simplify logp and logcdf helpers

ricardoV94:
These methods will only retrieve the RV logp and logcdf terms, and no longer attempt to replace any potential RVs in their arguments.

This also lifts the constraint that the RV whose logp or logcdf term is requested must have the right dimensions, as long as the respective expression broadcasts properly. Since these terms will not be used as inputs to anything else, it seems safe to allow this.

Therefore it is now possible to do the following, more in line with V3:
```python
x = Normal.dist(0, 1)
x_logp = logp(x, [0, 1])
```

**However the following is now also possible:**
```python
x = pm.Normal.dist()
y = pm.Normal.dist(x)
pm.logp(y, 0.5).eval()
```
And the logp term evaluation will depend on a stochastic value of `x` (or it's first draw). You could also do this before the PR, but would at least get warning:
```
UserWarning: Found a random variable that was neither among the observations nor the conditioned variables
```

This would have failed in V3, because `pm.Normal.dist` would be a `pymc3.distributions.continuous.Normal` which could NOT be used as input to other distributions.

To have a logprob of `y` that depends on an input for `x` the user should have done the following:
```python
x = at.scalar('x')
y = pm.Normal.dist(x)
pm.logp(y, 0.5).eval({x: 0})
```

During normal usage in a model context, where a logp or logcdf call only affect the model through a Potential this is not an issue because the RVs in the graph of `y` would be replaced by the respective value variable, via a call to `rvs_to_value_vars` that happens at the end of `model.potentiallogpt`:

```python
with pm.Model() as m:
  x = pm.Normal('x')
  y = pm.Normal.dist(x)
  pm.Potential('y_logprob', pm.logp(y, 0.5))
```

### DIFF
--- a/pymc/distributions/__init__.py
+++ b/pymc/distributions/__init__.py
@@ -13,10 +13,8 @@
 #   limitations under the License.
 
 from pymc.distributions.logprob import (  # isort:skip
-    _logcdf,
     logcdf,
     logp,
-    logcdfpt,
     logp_transform,
     logpt,
     logpt_sum,
@@ -195,6 +193,5 @@ __all__ = [
     "logp",
     "logp_transform",
     "logcdf",
-    "_logcdf",
     "logpt_sum",
 ]
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -14,7 +14,6 @@
 import aesara.tensor as at
 import numpy as np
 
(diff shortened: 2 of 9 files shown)


twiecki:
This looks great!

michaelosthege (approved):

Issue #5121, opened 2021-10-31
Include `observed_data` and `sample_stats` in inferencedata returned from `sample_numpyro_nuts`


GMCobraz:
Dear PyMC developers,

I notice the trace output difference between ordinary sampling and jax_sampling.

I notice that with return_inferencedata=True in ordinary sampling, I will get the following in trace:
posterior
log_likelihood
sample_stats
observed_data

**But in jax_sampling, I only get posterior, which I cannot proceed with plot_ppc as I do not have observed_data.**
May I know how to resolve this issue? Or is there any setup I can do to resolve this issue?

thanks

Below is my reproducible code:
import warnings

import arviz as az
import matplotlib.pyplot as plt
import numpy as np
import pandas as pd
import pymc3 as pm
import pymc3.sampling_jax
import numpyro
import theano
numpyro.util.set_platform('cpu')

print(f"Running on PyMC3 v{pm.__version__}")

import numpy as np
import pandas as pd
import datetime as dt
#from pandas_datareader import data
import matplotlib.pyplot as plt

%matplotlib inline

returns = pd.read_csv(pm.get_data("SP500.csv"), index_col="Date")
returns["change"] = np.log(returns["Close"]).diff()
returns = returns.dropna()
n=len(returns)
returns.head()

# Wiggins
def WigginsDGP(volatility_mu, volatility_theta, volatility_sigma):

    wiggins = []
    volatility = [volatility_mu]
    for _ in range(n):
        v = volatility[-1] + volatility_theta * (volatility_mu - volatility[-1]) \
            + np.random.normal(0., volatility_sigma)

        volatility.append(v)
        r = np.random.normal(0., np.exp(v))

        wiggins.append(r)

    return np.array(wiggins)

def sde(x, theta, mu, sigma):
    return theta * (mu - x), sigma

%%time
import pymc3.distributions.timeseries as ts
with pm.Model() as wiggins_model:

    volatility_theta = pm.Uniform('volatility_theta', lower=0., upper=1., testval=0.5)
    volatility_mu = pm.Normal('volatility_mu', mu=-5., sd=.1, testval=-5)
    volatility_sigma = pm.Uniform('volatility_sigma', lower=0.001, upper=0.09, testval=0.05)

    #sde = lambda x, theta, mu, sigma: (theta * (mu - x), sigma)
    
    volatility = ts.EulerMaruyama('volatility',
                                  1.0,
                                  sde,
[...]

PR #4489, opened 2021-02-26
Port InferenceData conversion code to pymc3 codebase

OriolAbril:
I have started porting the `arviz.from_pymc3` converter to the pymc3 codebase.
This should first allow to simplify significantly the code (which has already happened quite a lot
and probably can be further simplified) and to allow having an ArviZ independent test suite.

closes https://github.com/arviz-devs/arviz/issues/1278, closes https://github.com/arviz-devs/arviz/issues/939 and closes https://github.com/arviz-devs/arviz/issues/1470

I think we can also solve the issue in https://github.com/arviz-devs/arviz/issues/1224 by vectorizing and being fast enough a progressbar makes no sense anymore.

There surely is a lot of room for improvement as I am not yet too familiar with all the v4 changes.

Depending on what your PR does, here are a few things you might want to address in the description:
+ [x] what are the (breaking) changes that this PR makes? It should not break anything
+ [x] important background, or details about the implementation: see description above
+ [ ] are the changes—especially new features—covered by tests and docstrings?
+ [x] [linting/style checks have been run](https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style)
+ [ ] [consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples)
+ [ ] right before it's ready to merge, mention the PR in the RELEASE-NOTES.md

PR #5189, opened 2021-11-16
Adding log_likelihood, observed_data, and sample_stats to numpyro sampler

zaxtax:
This adds more fields to the trace object returned from `sample_numpyro_nuts` addressing some of the concerns in #5121

### DIFF
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -25,7 +25,7 @@ import pymc
 
 from pymc.aesaraf import extract_obs_data
 from pymc.distributions import logpt
-from pymc.model import modelcontext, Model
+from pymc.model import Model, modelcontext
 from pymc.util import get_default_varnames
 
 if TYPE_CHECKING:
@@ -216,7 +216,6 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
 
         self.density_dist_obs = density_dist_obs
         self.observations = find_observations(self.model)
-
 
     def split_trace(self) -> Tuple[Union[None, "MultiTrace"], Union[None, "MultiTrace"]]:
         """Split MultiTrace object into posterior and warmup.
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -25,13 +25,14 @@ import pymc
 
 from pymc.aesaraf import extract_obs_data
 from pymc.distributions import logpt
-from pymc.model import Model, modelcontext
+from pymc.model import modelcontext
 from pymc.util import get_default_varnames
 
 if TYPE_CHECKING:
     from typing import Set  # pylint: disable=ungrouped-imports
 
     from pymc.backends.base import MultiTrace  # pylint: disable=invalid-name
+    from pymc.model import Model
 
 ___all__ = [""]
 
@@ -41,7 +42,7 @@ _log = logging.getLogger("pymc")
 Var = Any  # pylint: disable=invalid-name
 
 
-def find_observations(model: Model) -> Optional[Dict[str, Var]]:
+def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
     """If there are observations available, return them as a dictionary."""
     if model is None:
         return None
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -25,14 +25,13 @@ import pymc
 
 from pymc.aesaraf import extract_obs_data
 from pymc.distributions import logpt
-from pymc.model import modelcontext
+from pymc.model import modelcontext, Model
 from pymc.util import get_default_varnames
 
 if TYPE_CHECKING:
     from typing import Set  # pylint: disable=ungrouped-imports
 
     from pymc.backends.base import MultiTrace  # pylint: disable=invalid-name
-    from pymc.model import Model
 
 ___all__ = [""]
 
@@ -40,6 +39,26 @@ _log = logging.getLogger("pymc")
 
 # random variable object ...
 Var = Any  # pylint: disable=invalid-name
+
+
+def find_observations(model: Model) -> Optional[Dict[str, Var]]:
+    """If there are observations available, return them as a dictionary."""
+    if model is None:
+        return None
+
+    observations = {}
+    for obs in model.observed_RVs:
+        aux_obs = getattr(obs.tag, "observations", None)
+        if aux_obs is not None:
+            try:
+                obs_data = extract_obs_data(aux_obs)
+                observations[obs.name] = obs_data
+            except TypeError:
+                warnings.warn(f"Could not extract data from symbolic observation {obs}")
+        else:
+            warnings.warn(f"No data for observation {obs}")
+
+    return observations
 
 
 class _DefaultTrace:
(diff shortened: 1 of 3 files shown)


pymc/sampling_jax.py
```diff
@@ -94,6 +94,45 @@ def logp_fn_wrap(x):
     return logp_fn_wrap
 
 
+# Adopted from pm.to_inference_data
+def find_observations(model: Model) -> Dict[str, Any]:
```
ricardoV94: Can we refactor the function in the backends.arviz so that we don't duplicate the code inside the library.l? It seems that the original function can be taken out of the class where it exists or made into static method that expects a model as the only argument like here.

Then we can simply import it here.

ricardoV94 (changes requested):
Looks great! Just left a comment below to avoid recreating the same code

pymc/sampling_jax.py
zaxtax: Yep, just refactored `find_observations` so we just import it

pymc/backends/arviz.py
```diff
@@ -42,6 +42,26 @@
 Var = Any  # pylint: disable=invalid-name
 
 
+def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
```
ricardoV94, line 45: Shouldn't that type hint be something like:
```suggestion
def find_observations(model: Optional[Model]) -> Optional[Dict[str, Var]]:
```

Where `Model` is found in `pymc.model`

zaxtax: `Optional["Model"]` is the convention used in the rest of the file.

zaxtax:
I don't understand why I am getting these errors. All the tests pass on my local machine. Tried it with python 3.7, 3.8, and 3.9

ricardoV94 (approved):

ricardoV94:
@zaxtax Do you want to do something more or can we merge?

pymc/sampling_jax.py
```diff
@@ -95,6 +97,39 @@ def logp_fn_wrap(x):
...
+            continue
+        name = rename_key.get(stat, stat)
+        value = value.copy()
+        data[name] = value
+        if stat == "num_steps":
+            data["tree_depth"] = np.log2(value).astype(int) + 1
+    return data
+
+
+def _get_log_likelihood(model, samples):
```
ricardoV94, line 121: Can you add a test that shows this is working? Perhaps compare with the output of the implementation in backends/arviz.py?

ricardoV94 (changes requested):

pymc/sampling_jax.py
zaxtax: Yep! Will do. Just got too excited that I have this working!

ricardoV94: Yeah it's awesome!

zaxtax: The test isn't perfect but it does use a transformed variable and I feel is fairly tight.

pymc/tests/test_sampling_jax.py
```diff
@@ -61,6 +62,24 @@ def test_deterministic_samples():
...
 
+def test_get_log_log_likelihood():
+    obs = np.random.normal(10, 2, size=100)
+    obs_at = aesara.shared(obs, borrow=True, name="obs")
+    with pm.Model() as model:
+        a = pm.Normal("a", 0, 2)
+        sigma = pm.HalfNormal("sigma")
+        b = pm.Normal("b", a, sigma=sigma, observed=obs_at)
+
+        trace = pm.sample(chains=1, random_seed=1322)
```
ricardoV94: We can run a smaller sampling with multiple chains like `(tune=10, draws=10, chains=2)`. That should be as good of a test but also covers multiple chains

zaxtax: Sounds good.

pymc/sampling_jax.py
ricardoV94: It seems good enough

pymc/tests/test_sampling_jax.py
```diff
@@ -61,6 +62,24 @@ def test_deterministic_samples():
     assert np.allclose(trace.posterior["b"].values, trace.posterior["a"].values / 2)
 
 
+def test_get_log_log_likelihood():
```
ricardoV94:
```suggestion
def test_get_log_likelihood():
```

ricardoV94 (approved):

ricardoV94:
Thanks @zaxtax. Awesome work!

pymc/sampling_jax.py
```diff
@@ -95,6 +97,39 @@ def logp_fn_wrap(x):
...
+    "Compute log-likelihood for all observations"
+    data = {}
+    for v in model.observed_RVs:
+        logp_v = replace_shared_variables([logpt(v)])
+        fgraph = FunctionGraph(model.value_vars, logp_v, clone=False)
+        jax_fn = jax_funcify(fgraph)
+        result = jax.vmap(jax.vmap(jax_fn))(*samples)[0]
```
ricardoV94, line 128: Out of curiosity, would we expect any benefits to jit_compiling this outer `vmap`?

OriolAbril: Would it be possible to use a similar approach with Aesara directly? 

Here we only loop over observed variables in order to get the pointwise log likelihood. We had some discussion about this in https://github.com/pymc-devs/pymc/pull/4489 but ended up keeping the 3 nested loops over variables, chains and draws.

OriolAbril:
Thanks! We should document that while posterior, log_likelihood, sample_stats and observed_data groups will be created, all coords and dims are ignored unlike with the "regular" backend. 

Is sample numpyro in the docs already? Should it be?

ricardoV94:
> Thanks! We should document that while posterior, log_likelihood, sample_stats and observed_data groups will be created, all coords and dims are ignored unlike with the "regular" backend.

I guess we could also retrieve these, no?

> Is sample numpyro in the docs already? Should it be?

If it's not yet, it should!

pymc/sampling_jax.py
ricardoV94: It should be possible, but requires a Aesara Scan, and at least for small models this was not faster than python looping when I checked it. Here is a Notebook that documents some things I tried: https://gist.github.com/ricardoV94/6089a8c46a0e19665f01c79ea04e1cb2

It might be faster if using shared variables...

zaxtax: No idea. I think the easiest thing to do is just benchmark it. I don't even call `optimize_graph` on either the graph in this function or the main sample routine.

When I run the model in the unit test with the change

`result = jax.vmap(jax.vmap(jax_fn))(*samples)[0]` to 
`result = jax.jit(jax.vmap(jax.vmap(jax_fn)))(*samples)[0]`

I don't really get a speed-up until there are millions of samples.

zaxtax:
@OriolAbril I think it would take me about the same effort to document the discrepancy as just do the correct thing with coords and dims.

Out of curiosity, there used to be a jax sampler based on TFP. Did that just silently get dropped?

pymc/sampling_jax.py
ricardoV94: > I  don't even call `optimize_graph` on either the graph in this function or the main sample routine

We should definitely call `optimize_graph`, otherwise the computed logps may not correspond to the ones used during sampling. For instance we have many optimizations that improve numerically stability, so you might get underflows to `-inf` for some of the posterior samples (which would never have been accepted by NUTS) which could screw up things downstream.

OriolAbril:
> @OriolAbril I think it would take me about the same effort to document the discrepancy as just do the correct thing with coords and dims.

Thanks! I assume the code is already written actually, either here in the `to_inference_data` function or in io_numpyro in ArviZ it should be a matter of combining the pieces together.

ricardoV94:
> Out of curiosity, there used to be a jax sampler based on TFP. Did that just silently get dropped?

Yes

pymc/sampling_jax.py
OriolAbril: > It should be possible, but requires a Aesara Scan, and at least for small models this was not faster than python looping when I checked it.

Then it's probably not worth it. I was under the impression it would be possible to vectorize/broadcast the operation from the conversations in #4489 and in slack.

ricardoV94: It must be possible, since the vmap above works just fine. I just have no idea how they do it xD, or how/if you could do it in Aesara. I also wonder whether the vmap works for more complicated models with multivariate distributions and the like

zaxtax: Alright. I'm going to make a separate PR for some of this other stuff.

OriolAbril: Cool, feel free to tag me if you want me to review, I am not watching PRs. I can already say I won't be able to help with the vectorized log_likelihood thing, I tried and I lost much more time with that than what would have been healthy. I should be able to help with coords and dims though

Issue #5191, opened 2021-11-16
Test suite failing in latest version of aeppl


ricardoV94:
I have opened an issue in https://github.com/aesara-devs/aeppl/issues/92

It is still unclear what is going on, but seems related to transforms

brandonwillard:
`pymc/tests/test_dist_math.py::test_multinomial_bound` might be failing due to a change in a `Shape` sub-graph.  Printing `aesara.dprint(modelA.logpt)`, the change between AePPL versions appears to take the node
```python
 |   |     | |     | |Shape [id DB] ''
 |   |     | |     | | |Elemwise{sub,no_inplace} [id DC] ''
 |   |     | |     | |   |Subtensor{:int64:} [id DD] ''
 |   |     | |     | |   | |Elemwise{log,no_inplace} [id DE] ''
 |   |     | |     | |   | | |Elemwise{true_div,no_inplace} [id N] 'p_simplex___simplex'
 |   |     | |     | |   | |ScalarConstant{-1} [id DF]
 |   |     | |     | |   |Elemwise{true_div,no_inplace} [id DG] ''
 |   |     | |     | |     |InplaceDimShuffle{x} [id DH] ''
 |   |     | |     | |     | |Sum{axis=[0], acc_dtype=float64} [id DI] ''
 |   |     | |     | |     |   |Elemwise{log,no_inplace} [id DE] ''
 |   |     | |     | |     |InplaceDimShuffle{x} [id DJ] ''
 |   |     | |     | |       |Subtensor{int64} [id DK] ''
 |   |     | |     | |         |Shape [id DL] ''
 |   |     | |     | |         | |Elemwise{true_div,no_inplace} [id N] 'p_simplex___simplex'
 |   |     | |     | |         |ScalarConstant{-1} [id DM]
```
and changes it to
```python
 |   |     | |     | |Shape [id DC] ''
 |   |     | |     | | |p_simplex__ [id T]
```

Here are the full graph print-outs for reference:
<details><summary>old graph</summary>
<p>

```python
Elemwise{add,no_inplace} [id A] '__logp'
 |Sum{acc_dtype=float64} [id B] ''
 | |Elemwise{mul,no_inplace} [id C] ''
 |   |Elemwise{add,no_inplace} [id D] 'p_simplex___logprob'
 |   | |Elemwise{switch,no_inplace} [id E] ''
 |   | | |All [id F] ''
 |   | | | |MakeVector{dtype='bool'} [id G] ''
[...]

PR #5193, opened 2021-11-17
Adding polya gamma moment and corresponding tests

jlopezarriaza:
Adding polya gamma moment and corresponding tests

### DIFF
--- a/pymc/tests/test_distributions_moments.py
+++ b/pymc/tests/test_distributions_moments.py
@@ -41,6 +41,7 @@ from pymc.distributions import (
     Normal,
     Pareto,
     Poisson,
+    PolyaGamma,
     Rice,
     SkewNormal,
     StudentT,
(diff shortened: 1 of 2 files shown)


jlopezarriaza:
I'm not 100% what's causing the checks to fail

ricardoV94:
> I'm not 100% what's causing the checks to fail

#5191

pymc/tests/test_distributions_moments.py
```diff
@@ -898,3 +899,39 @@ def test_rice_moment(nu, sigma, size, expected):
...
+                0.25,
+                0.46211715726000974,
+                0.5711956169668236,
+                0.6034321690965776,
+                0.6025172375473855,
+            ),
+        ),
+    ],
+)
+def test_polyagamma_moment(h, z, size, expected):
```
ricardoV94, line 1035: Can you add a test where size is not None?

ricardoV94 (changes requested):

pymc/tests/test_distributions_moments.py
jlopezarriaza: Done, overlooked that.

ricardoV94: No problem and thanks!

ricardoV94 (approved):

ricardoV94:
Thanks @jlopezarriaza

PR #5200, opened 2021-11-17
Adding `StickBreakingWeights` distribution class

larryshamalama:
This pull request adds a class for truncated stick-breaking weights which are vital in the construction of Dirichlet Processes, which I hope to work on soon. See section 1.1 in [Ishwaran and James (2001)](http://people.ee.duke.edu/~lcarin/Yuting3.3.06.pdf). My previous attempt as a PR is #4971 .

Some tests are failing, will have to resolve them first.

CC: @ricardoV94 @AustinRochford @fonnesbeck

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2237,6 +2237,36 @@ stickbreakingweights = StickBreakingWeightsRV()
 
 
 class StickBreakingWeights(Continuous):
+    r"""
+    Likelihood of truncated stick-breaking weights. The weights are generated
+
+    .. math:
+
+
+        f(\mathbf{x}|\alpha, K) = 
+
+    ========  ===============================================
+    Support   :math:`x_i \in (0, 1)` for :math:`i \in \{1, \ldots, K+1\}`
+              such that :math:`\sum x_i = 1`
+    Mean      :math:`\dfrac{a_i}{\sum a_i}`
+    Variance  :math:`\dfrac{a_i - \sum a_0}{a_0^2 (a_0 + 1)}`
+              where :math:`a_0 = \sum a_i`
+    ========  ===============================================
+
+    Parameters
+    ----------
+    alpha: float
+        Concentration parameters (alpha > 0).
+    K: int
+        The number of "sticks" to break off from an initial one-unit stick. The length
+        of categories is K + 1, where the last weight is one minus the sum of all the first sticks.
+
+    References
+    ----------
+    .. [1] Ishwaran James
+
+    .. [2] Peter Mueller
+    """
     rv_op = stickbreakingweights
 
     def __new__(cls, name, *args, **kwargs):
@@ -2277,6 +2307,19 @@ class StickBreakingWeights(Continuous):
         return moment
 
     def logp(value, alpha, K):
+        """
+        Calculate log-probability of the distribution induced from the stick-breaking process
+        at specified value.
+
+        Parameters
+        ----------
+        value: numeric
+            Value for which log-probability is calculated.
+
+        Returns
+        -------
+        TensorVariable
+        """
         logp = -at.sum(
             at.log(
                 at.cumsum(
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -67,6 +67,7 @@ __all__ = [
     "MatrixNormal",
     "KroneckerNormal",
     "CAR",
+    "StickBreakingWeights",
 ]
 
 # Step methods and advi do not catch LinAlgErrors at the
@@ -2166,4 +2167,91 @@ class CAR(Continuous):
             alpha >= -1,
             tau > 0,
             msg="-1 <= alpha <= 1, tau > 0",
+        )
+
+
+class StickBreakingWeightsRV(RandomVariable):
+    name = "stick_breaking_weights"
+    ndim_supp = 1
+    ndims_params = [0]
+    dtype = "floatX"
+    _print_name = ("StickBreakingWeights", "\\operatorname{StickBreakingWeights}")
+    
+    def __call__(self, alpha=1., size=None, **kwargs):
+        return super().__call__(alpha, size=size, **kwargs)
+
+    def _infer_shape(self, size, dist_params, param_shapes=None):
+        if isinstance(size, int):
+            return size
+        else:
+            return tuple(size[:-1]) + tuple(size[-1] + 1,)
+    
+    @classmethod
+    def rng_fn(cls, rng, alpha, size):
+        if size is None:
+            raise ValueError("size cannot be None because K is inferred from size")
+            
+        if isinstance(size, int):
+            size = (size,)
+            
+        betas = rng.beta(1, alpha, size=size) # adapt this to vector alpha
+        
+        sticks = np.concatenate(
+            (
+                np.ones(shape=(size[:-1] + (1,))),
+                np.cumprod(1 - betas[..., :-1], axis=-1),
+            ),
+            axis=-1,
+        )
+        
+        weights = sticks * betas
+        weights = np.concatenate(
+            (
+                weights,
+                1 - weights.sum(axis=-1)[..., np.newaxis]
+            ),
+            axis=-1,
+        )
+
+        return weights
+    
+
+stickbreakingweights = StickBreakingWeightsRV()
+
+
+class StickBreakingWeights(Continuous):
+    rv_op = stickbreakingweights
+    
+    def __new__(cls, name, *args, **kwargs):
+        kwargs.setdefault("transform", transforms.stick_breaking)
+        return super().__new__(cls, name, *args, **kwargs)
+
+    @classmethod
+    def dist(cls, alpha, *args, **kwargs):
+        alpha = at.as_tensor_variable(floatX(alpha))
+
+        assert_negative_support(alpha, "alpha", "StickBreakingWeights")
+
+        return super().dist([alpha], **kwargs)
+
+    def logp(value, alpha):
+        K = floatX(value.shape[-1])
+        
+        logp = -at.sum(
+            at.log(
+                at.cumsum(
+                    value[..., ::-1],
+                    axis=-1,
+                )
+            ),
+            axis=-1,
+        )
+        logp -= -K * betaln(1, alpha)
+        logp += alpha * at.log(value[..., -1]) # check this
+
+        return bound(
+            logp,
+            alpha > 0,
+            at.all(value >= 0),
+            at.all(value <= 1),
         )
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2183,8 +2183,20 @@ class StickBreakingWeightsRV(RandomVariable):
     dtype = "floatX"
     _print_name = ("StickBreakingWeights", "\\operatorname{StickBreakingWeights}")
 
-    def __call__(self, alpha=1.0, K=10, size=None, **kwargs):
-        return super().__call__(alpha, K, size=size, **kwargs)
+    def make_node(self, rng, size, dtype, alpha, K):
+
+        alpha = at.as_tensor_variable(alpha)
+        K = at.as_tensor_variable(intX(K))
+
+        # if at.lt(K, 0):
+        #     print(at.lt(K, 0).eval())
+        #     print(K.eval() < 0)
+        #     raise ValueError("K needs to be positive.")
+
+        if alpha.ndim > 0:
+            raise ValueError("The concentration parameter needs to be a scalar.")
+
+        return super().make_node(rng, size, dtype, alpha, K)
 
     def _infer_shape(self, size, dist_params, param_shapes=None):
         alpha, K = dist_params
@@ -2195,12 +2207,6 @@ class StickBreakingWeightsRV(RandomVariable):
 
     @classmethod
     def rng_fn(cls, rng, alpha, K, size):
-        if K < 0:
-            raise ValueError(f"K needs to be a positive integer, but K = {K}.")
-
-        if np.ndim(alpha) > 0:
-            raise ValueError("The concentration parameter needs to be a scalar.")
-
         if size is None:
             size = (K,)
         elif isinstance(size, int):
@@ -2208,7 +2214,7 @@ class StickBreakingWeightsRV(RandomVariable):
         else:
             size = tuple(size) + (K,)
 
-        betas = rng.beta(1, alpha, size=size)  # adapt this to vector alpha
+        betas = rng.beta(1, alpha, size=size)
 
         sticks = np.concatenate(
             (
@@ -2281,13 +2287,26 @@ class StickBreakingWeights(Continuous):
             axis=-1,
         )
         logp += -(K - 1) * betaln(1, alpha)
-        logp += alpha * at.log(value[..., -1])  # check this
+        logp += alpha * at.log(value[..., -1])
+
+        logp = at.switch(
+            at.or_(
+                at.any(
+                    at.and_(at.le(value, 0), at.ge(value, 1)),
+                    axis=-1,
+                ),
+                at.or_(
+                    at.bitwise_not(at.allclose(value.sum(-1), 1)),
+                    at.neq(value.shape[-1], K + 1),
+                ),
+            ),
+            -np.inf,
+            logp,
+        )
 
         return check_parameters(
             logp,
             alpha > 0,
             K > 0,
-            at.all(value >= 0),
-            at.all(value <= 1),
-            at.allclose(value.sum(axis=-1), 1),
+            msg="alpha > 0, K > 0, 0 <= value <= 1, sum(value) == 1",
         )
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2173,26 +2173,34 @@ class CAR(Continuous):
 class StickBreakingWeightsRV(RandomVariable):
     name = "stick_breaking_weights"
     ndim_supp = 1
-    ndims_params = [0]
+    ndims_params = [0, 0]
     dtype = "floatX"
     _print_name = ("StickBreakingWeights", "\\operatorname{StickBreakingWeights}")
     
-    def __call__(self, alpha=1., size=None, **kwargs):
-        return super().__call__(alpha, size=size, **kwargs)
+    def __call__(self, alpha=1., K=10, size=None, **kwargs):
+        return super().__call__(alpha, K, size=size, **kwargs)
 
     def _infer_shape(self, size, dist_params, param_shapes=None):
-        if isinstance(size, int):
-            return size
-        else:
-            return tuple(size[:-1]) + tuple(size[-1] + 1,)
+        alpha, K = dist_params
+        return [K + 1,]
     
     @classmethod
-    def rng_fn(cls, rng, alpha, size):
+    def rng_fn(cls, rng, alpha, K, size):
+        if K < 0:
+            raise ValueError("K needs to be a positive integer.")
+        
+        if np.ndim(alpha) > 0:
+            raise ValueError("The concentration parameter needs to be a scalar.")
+
         if size is None:
             raise ValueError("size cannot be None because K is inferred from size")
             
-        if isinstance(size, int):
-            size = (size,)
+        if size is None:
+            size = (K,)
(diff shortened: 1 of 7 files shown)


pymc/distributions/multivariate.py
```diff
@@ -2136,3 +2137,117 @@ def logp(value, mu, W, alpha, tau):
...
+    def logp(value, alpha, K):
+        if at.lt(K, 0).eval():
+            raise ValueError("K needs to be a positive integer.")
+
+        if at.lt(alpha, 0).eval():
+            raise ValueError("alpha needs to be a positive.")
+
+        if not at.allclose(value.sum(axis=-1), 1).eval():
+            raise ValueError("weights need to sum to 1.")
+
```
larryshamalama, line 2317: Generally, I am using `.eval()` for some checks here, but I don't know if this is generalizable especially that inputs can be `np.ndarray` or Aesara tensors. Is there a better to do these checks?

pymc/distributions/multivariate.py
```diff
@@ -2136,3 +2137,117 @@ def logp(value, mu, W, alpha, tau):
...
+    def logp(value, alpha, K):
+        if at.lt(K, 0).eval():
+            raise ValueError("K needs to be a positive integer.")
+
+        if at.lt(alpha, 0).eval():
+            raise ValueError("alpha needs to be a positive.")
+
+        if not at.allclose(value.sum(axis=-1), 1).eval():
+            raise ValueError("weights need to sum to 1.")
```
ricardoV94: These should just be conditions in bound, and not `.eval` preemptively (you would need to know the inputs to be able to evaluate them)

pymc/tests/test_distributions.py
```diff
@@ -2136,6 +2137,42 @@ def test_dirichlet_2D(self):
...
+    def test_stickbreakingweights(self, value, alpha, K, logp):
+        with Model() as model:
+            StickBreakingWeights("sbw", alpha=alpha, K=K)
+        pt = {"sbw": value}
+        assert_almost_equal(
+            model.fastlogp(pt),
+            logp,
+            decimal=select_by_precision(float64=6, float32=2),
+            err_msg=str(pt),
+        )
```
larryshamalama, line 2151: I drew inspiration `test_ex_gaussian` because there is no scipy or numpy logp equivalent. Here is the error that I'm getting with this test.

Screen Shot 2021-11-17 at 5 42 05 PM

pymc/tests/test_distributions_random.py
```diff
@@ -1175,6 +1175,26 @@ class TestDirichlet(BaseTestDistribution):
...
+            20,
+        ),
+        (11, 5, 20),
+        (3, 13, 5, 20),
+    ]
+    tests_to_run = [
+        "check_pymc_params_match_rv_op",
+        "check_rv_size",
+    ]
+
```
larryshamalama, line 1199: This test is passing!

pymc/tests/test_distributions_moments.py
```diff
@@ -897,4 +898,12 @@ def test_asymmetriclaplace_moment(b, kappa, mu, size, expected):
 def test_rice_moment(nu, sigma, size, expected):
     with Model() as model:
         Rice("x", nu=nu, sigma=sigma, size=size)
+
+
+@pytest.mark.parametrize(
+    "alpha, K, size, expected", [(3, 11, None, (3 / 4) ** np.arange(12) * 1 / 3)]
+)
+def test_stickbreakingweights_moment(alpha, K, size, expected):
```
larryshamalama, line 1117: Here, I am obtaining the following error and I am not sure where `result` comes from. Does it represent a random draw? My understanding is that, for other moment tests, `result` is not a random draw. I obtain the same error with the same numbers every time I try to run this test.

Screen Shot 2021-11-17 at 5 44 19 PM

larryshamalama (commented):
Two sets of tests not passing locally in `test_distributions.py` and `test_distributions_moments.py`. The test in `test_distributions_random.py` seems okay (from my perspective)

pymc/distributions/multivariate.py
larryshamalama: Okay, will add these to `bound`, thanks! I just finished writing some comments in a code review but am much slower than your response speed 😅

pymc/tests/test_distributions_moments.py
ricardoV94: `result` is the `moment`. It seems to deviate slightly from the expected moment. How did you pre-compute it? Are you sure it matches the `get_moment` computation?

larryshamalama: Fixed, my `get_moment` and unit tests were wrong

pymc/tests/test_distributions.py
```diff
@@ -2112,6 +2113,32 @@ def test_dirichlet(self, n):
...
+            (np.tile(1, 13) / 13, 2, 12, 13.286898065112881),
+            (np.array([0.001] * 10 + [0.99]), 0.1, 10, -20.66907735582068),
+            (np.append(0.5 ** np.arange(1, 20), 0.5 ** 20), 5, 19, 92.59518981534681),
+        ],
+    )
+    def test_stickbreakingweights(self, value, alpha, K, logp):
+        with Model() as model:
+            StickBreakingWeights("sbw", alpha=alpha, K=K)
+        pt = {"sbw": value}
+        assert_almost_equal(
```
larryshamalama, line 2146: I am getting a `TypeError: Unknown input or state: sbw.` error here... Am I supposed to name things different or replace `pt` by `pt = {"sbw_simplex__": value}`? If I do that, then I obtained `AssertionError` but I'm not sure if that's the correct approach...

larryshamalama (commented):
Question about `test_distributions.py`

pymc/tests/test_distributions.py
ricardoV94: You have to disable transform, with `transform=None`, or else your reference logp has to account for the Jacobian term as well.

larryshamalama: This works well, thanks!

pymc/distributions/multivariate.py
larryshamalama: Quick comment that I replace `bound` with `check_parameters` following a recent merged PR (to my understanding)

pymc/tests/test_distributions.py
larryshamalama: Solved by using `transform=None`, which is okay when we're just testing the logp method

pymc/distributions/multivariate.py
```diff
@@ -2153,3 +2160,120 @@ def logp(value, mu, W, alpha, tau):
...
+
+
+class StickBreakingWeightsRV(RandomVariable):
+    name = "stick_breaking_weights"
+    ndim_supp = 1
+    ndims_params = [0, 0]
+    dtype = "floatX"
+    _print_name = ("StickBreakingWeights", "\\operatorname{StickBreakingWeights}")
+
+    def __call__(self, alpha=1.0, K=10, size=None, **kwargs):
```
ricardoV94: Are you sure you want default parameters? Any reason to think these are particularly popular?

pymc/distributions/multivariate.py
```diff
@@ -2153,3 +2160,120 @@ def logp(value, mu, W, alpha, tau):
...
+    def _infer_shape(self, size, dist_params, param_shapes=None):
+        alpha, K = dist_params
+
+        size = tuple(size)
+
+        return size + (K + 1,)
+
+    @classmethod
+    def rng_fn(cls, rng, alpha, K, size):
+        if K < 0:
```
ricardoV94, line 2208: I would remove the "integer" from the message, it makes it less clear the problem is the being negative

pymc/distributions/multivariate.py
```diff
@@ -2153,3 +2160,120 @@ def logp(value, mu, W, alpha, tau):
...
+    @classmethod
+    def dist(cls, alpha, K, *args, **kwargs):
+        alpha = at.as_tensor_variable(floatX(alpha))
+        K = at.as_tensor_variable(intX(K))
+
+        if alpha.ndim > 0:
+            raise ValueError("alpha must be a scalar.")
+
+        if K.ndim > 0:
+            raise ValueError("K must be a scalar.")
```
ricardoV94: If you move these checks to make_node in the RV you don't need to repeat them in the rng_fn because you already know that the RV cannot be created with the wrong input types

pymc/distributions/multivariate.py
```diff
@@ -2153,3 +2160,120 @@ def logp(value, mu, W, alpha, tau):
...
+        logp += -(K - 1) * betaln(1, alpha)
+        logp += alpha * at.log(value[..., -1])  # check this
+
+        return check_parameters(
+            logp,
+            alpha > 0,
+            K > 0,
+            at.all(value >= 0),
+            at.all(value <= 1),
+            at.allclose(value.sum(axis=-1), 1),
```
ricardoV94: Value checks should now be done outside of check parameters with an explicit switch statement that returns -inf if they are not met.

Also to be more specific you can add an `axis=-1` to the `at.all(value...)` to only switch the logp of the "invalid values"

You can see the multinomial distribution for an example.

pymc/distributions/multivariate.py
```diff
@@ -2153,3 +2160,120 @@ def logp(value, mu, W, alpha, tau):
...
+            at.log(
+                at.cumsum(
+                    value[..., ::-1],
+                    axis=-1,
+                )
+            ),
+            axis=-1,
+        )
+        logp += -(K - 1) * betaln(1, alpha)
+        logp += alpha * at.log(value[..., -1])  # check this
```
ricardoV94: Should that comment be removed?

pymc/distributions/multivariate.py
```diff
@@ -2153,3 +2160,120 @@ def logp(value, mu, W, alpha, tau):
...
+
+        return check_parameters(
```
ricardoV94, line 2349: You can add a `msg` kwargs that is shown when the checks are invalid. It also makes the debug_print easier to read. 

There are some examples in most other distributions

pymc/tests/test_distributions.py
```diff
@@ -2185,6 +2186,32 @@ def test_dirichlet_invalid(self):
...
+            model.fastlogp(pt),
+            logp,
+            decimal=select_by_precision(float64=6, float32=2),
+            err_msg=str(pt),
+        )
+
+    @pytest.mark.parametrize("n", [2, 3])
+    def test_multinomial(self, n):
+        self.check_logp(
+            Multinomial, Vector(Nat, n), {"p": Simplex(n), "n": Nat}, multinomial_logpdf
```
ricardoV94: This change seems unrelated. If you add a comma after the last function argument, pre-commit won't try to change those lines

pymc/distributions/multivariate.py
ricardoV94, line 2349: We need tests for the logp parameter checks and multivariate logp `value`s.

You can check `test_dirichlet_multinomial_vectorized` and `test_dirichlet_multinomial_invalid` for examples of both.

If you add a test that uses `check_logp` it will test for invalid scalar parameters `alpha` and `k` automatically. You would only need manual tests for the `value`

pymc/tests/test_distributions_random.py
```diff
@@ -1163,6 +1163,26 @@ class TestDirichlet(BaseTestDistribution):
     ]
 
 
+class TestStickBreakingWeights(BaseTestDistribution):
```
ricardoV94, line 1179: Is there any obvious statistical property you could easily test to be more confident the rng_fn is returning valid samples?

The values must some to 1 right? Maybe other things as well?

ricardoV94 (commented):
Looks great. Left some comments bellow, mostly concerning tests. Glad to see this PR entering the polishing phase!

pymc/distributions/multivariate.py
larryshamalama: Hmm no... they were somewhat arbitrary and perhaps there just to serve as a reminder about the functionality of `__call__`. Would you recommend removing this method?

pymc/distributions/multivariate.py
larryshamalama: Yes!

pymc/tests/test_distributions_random.py
larryshamalama: Yes values must sum to one. Should I also write a test that checks `0 <= value <= 1`?

pymc/tests/test_distributions.py
larryshamalama: Oh this might have been dangling from me copy pasting things for visual references and/or mix-ups when rebasing... I will delete this bit

larryshamalama (commented):
December 20 work in progress

pymc/distributions/multivariate.py
```diff
@@ -2160,3 +2167,139 @@ def logp(value, mu, W, alpha, tau):
...
+    def make_node(self, rng, size, dtype, alpha, K):
+
+        alpha = at.as_tensor_variable(alpha)
+        K = at.as_tensor_variable(intX(K))
+
+        # if at.lt(K, 0):
+        #     print(at.lt(K, 0).eval())
+        #     print(K.eval() < 0)
+        #     raise ValueError("K needs to be positive.")
+
```
larryshamalama: Screen Shot 2021-12-21 at 12 50 29 AM

This check in `make_node` is commented out for now because I keep mistakenly raising a `ValueError`

pymc/distributions/multivariate.py
```diff
@@ -2160,3 +2167,139 @@ def logp(value, mu, W, alpha, tau):
...
+            at.or_(
+                at.any(
+                    at.and_(at.le(value, 0), at.ge(value, 1)),
+                    axis=-1,
+                ),
+                at.or_(
+                    at.bitwise_not(at.allclose(value.sum(-1), 1)),
+                    at.neq(value.shape[-1], K + 1),
+                ),
+            ),
```
larryshamalama, line 2344: There are a lot of `at.or_`... is there a way to combine multiple elements on which we want to do the "or" operation? Would `scan` help us here? (I know nothing about `scan`)

pymc/tests/test_distributions.py
```diff
@@ -2185,6 +2186,34 @@ def test_dirichlet_invalid(self):
...
+        )
+
+    def test_stickbreakingweights_invalid(self):
+        sbw = pm.StickBreakingWeights.dist(3.0, 3)
+        sbw_wrong_K = pm.StickBreakingWeights.dist(3.0, 7)
+        assert pm.logp(sbw, np.array([0.4, 0.3, 0.2, 0.15])).eval() == -np.inf
+        assert pm.logp(sbw, np.array([1.1, 0.3, 0.2, 0.1])).eval() == -np.inf
+        assert pm.logp(sbw, np.array([0.4, 0.3, 0.2, -0.1])).eval() == -np.inf
+        assert pm.logp(sbw_wrong_K, np.array([0.4, 0.3, 0.2, 0.1])).eval() == -np.inf
+
```
larryshamalama, line 2160: Here are tests for invalid logp values akin to `pm.Multinomial`

pymc/tests/test_distributions_random.py
```diff
@@ -1176,6 +1176,40 @@ class TestDirichlet(BaseTestDistribution):
...
+            alpha=3.5,
+            K=19,
+            size=(2, 3, 5),
+            rng=default_rng,
+        )
+
+        assert np.allclose(draws.sum(-1).eval(), 1)
+        assert np.all(draws.eval() >= 0)
+        assert np.all(draws.eval() <= 1)
+
```
larryshamalama, line 1212: Another new test here

larryshamalama (commented):
Two new sets of tests + question in `make_node`

pymc/distributions/multivariate.py
```diff
@@ -2160,3 +2167,182 @@ def logp(value, mu, W, alpha, tau):
...
+            -np.inf,
+            logp,
+        )
+
+        return check_parameters(
+            logp,
+            alpha > 0,
+            K > 0,
+            msg="alpha > 0, K > 0, 0 <= value <= 1, sum(value) == 1",
```
ricardoV94: The value checks shouldn't be part of the message since they happen outside of parameter checks

pymc/distributions/multivariate.py
ricardoV94: This approach looks reasonable

pymc/distributions/multivariate.py
larryshamalama: Temporarily I have:

```{Python}
if K.eval() < 0:
            raise ValueError("K needs to be positive.")
```

pymc/distributions/multivariate.py
```diff
@@ -2167,3 +2174,183 @@ def logp(value, mu, W, alpha, tau):
...
+    dtype = "floatX"
+    _print_name = ("StickBreakingWeights", "\\operatorname{StickBreakingWeights}")
+
+    def make_node(self, rng, size, dtype, alpha, K):
+
+        alpha = at.as_tensor_variable(alpha)
+        K = at.as_tensor_variable(intX(K))
+
+        if K.eval() < 0:
+            raise ValueError("K needs to be positive.")
```
larryshamalama: If I write `if at.lt(K, 0)`, it often seems to incorrectly evaluate to True... Is there a better way to write this `if` statement?

pymc/tests/test_distributions.py
```diff
@@ -2123,6 +2124,34 @@ def test_dirichlet_invalid(self):
...
+    @pytest.mark.parametrize(
+        "value,alpha,K,logp",
+        [
+            (np.array([5, 4, 3, 2, 1]) / 15, 0.5, 4, 2.2057773112876893),
+            (np.tile(1, 13) / 13, 2, 12, 13.286898065112881),
+            (np.array([0.001] * 10 + [0.99]), 0.1, 10, -20.66907735582068),
+            (np.append(0.5 ** np.arange(1, 20), 0.5 ** 20), 5, 19, 92.59518981534681),
+        ],
+    )
+    def test_stickbreakingweights(self, value, alpha, K, logp):
```
```suggestion
    def test_stickbreakingweights_logp(self, value, alpha, K, logp):
```

ricardoV94: A test with 2d value might be useful

pymc/distributions/multivariate.py
ricardoV94: Here you can't raise a ValueError yet because the values are possibly still symbolic.  Such error should be in the `rng_fn` where it is already concrete.

In contrast you can raise erros for ndims, because those are fixed.

ricardoV94 (commented):
Looks great! Left some comments, let me know if anything is confusing

pymc/tests/test_distributions.py
larryshamalama: Added!

pymc/tests/test_distributions_random.py
```diff
@@ -1176,6 +1176,41 @@ class TestDirichlet(BaseTestDistribution):
...
+        draws = pm.StickBreakingWeights.dist(
+            alpha=3.5,
+            K=19,
+            size=(2, 3, 5),
+            rng=default_rng,
+        )
+
+        assert np.allclose(draws.sum(-1).eval(), 1)
+        assert np.all(draws.eval() >= 0)
+        assert np.all(draws.eval() <= 1)
```
ricardoV94: No need to call eval 3 times, you can store the results in a variable and then do the 3 checks on them

larryshamalama: Are you saying to do something like: `assert all([bool, bool2, bool3])`

ricardoV94:
```suggestion
        draws = draws.eval()
        assert np.allclose(draws.sum(-1), 1)
        assert np.all(draws >= 0)
        assert np.all(draws <= 1)
```

larryshamalama: Oops, got it!

larryshamalama: Thanks! Will commit this asap

ricardoV94 (approved):
LGTM. Awesome work @larryshamalama

larryshamalama:
As always, thanks so much for all your suggestions and help @ricardoV94

twiecki:
Would be great to have an example NB in pymc-examples that shows how to use this.

PR #5201, opened 2021-11-18
Add Multinomial moments

morganstrom:
NOTE that the size argument has a different behavior with this implementation compared to previously! Now, new dimensions are appended to the output tensor, as specified by size. Previously, no new dimensions were added to the mode if the size argument matched the shape of the mode tensor.. 

Also, the mode is rounded to the nearest integer in this implementation. The new tests are different from the previous ones - instead of making sure that the values of the last dimension summed to 1 (which isn't guaranteed with rounding), now each element of the output tensor are tested. 


ricardoV94:
> NOTE that the size argument has a different behavior with this implementation compared to previously!

Before we didn't have `size`, only `shape`. Indeed `size` behaves differently and is now what we use internally. The refactoring of the old tests can (but don't need to) keep using the `shape` definition, so that they are exactly the same as before.  Any new tests should use `size`.

pymc/tests/test_distributions_moments.py
```diff
@@ -1036,3 +1037,28 @@ def test_polyagamma_moment(h, z, size, expected):
     with Model() as model:
         PolyaGamma("x", h=h, z=z, size=size)
     assert_moment_is_expected(model, expected)
+
+
+@pytest.mark.parametrize(
+    "p, n, size, expected",
+    [
+        (np.array([0.25, 0.25, 0.25, 0.25]), 1, None, np.repeat(0, 4)),
```
ricardoV94: This is a bad moment, because it has a logp of `-np.inf`. We always need the moment to add up to `n`. IIRC there is no simple closed formula solution for the mode of the multinomial, but you should be able to use our hacky approximation from V3.

ricardoV94 (changes requested):

pymc/tests/test_distributions_moments.py
morganstrom: This? https://github.com/pymc-devs/pymc/blob/c332f8c957bb2ebaf346832f440e5ee94b65026a/pymc3/distributions/multivariate.py#L583

ricardoV94: Yeah I think that's it. Basically it's a hack to make sure the values always add up to `n`, and not just in the case where they would all be zero

morganstrom: Cool, I pretty much copied the code but made some changes for the cases when either `p` has more then 1 dimension or when `n` is an array rather than a scalar. Added tests for the corner cases I could think of. Once this PR is finalized, the same code can be copied to the DirichletMultinomial distribution!

ricardoV94 (approved):
LGTM! Just need to fix the conflicts with main

ricardoV94:
In the meantime feel free to start a PR on the DirichletMultinomial :)

ricardoV94:
You got a pre-commit formatting complaint

morganstrom:
> You got a pre-commit formatting complaint

Darn! How do I run the same tests locally before commiting? Can't find instructions here: https://github.com/pymc-devs/pymc/blob/35f9966f15e76d8ba304914be69e4115fdcfd4ef/CONTRIBUTING.md

ricardoV94:
> > You got a pre-commit formatting complaint
> 
> Darn! How do I run the same tests locally before commiting? Can't find instructions here: https://github.com/pymc-devs/pymc/blob/35f9966f15e76d8ba304914be69e4115fdcfd4ef/CONTRIBUTING.md

They were recently moved to: https://github.com/pymc-devs/pymc/blob/main/docs/source/contributing/python_style.md

Issue #5088, opened 2021-10-19
Issue page for `xfail`s added during `aeppl` integration.


kc611:
The `XFail`s added during `aeppl` integration #4887 :

- [x] `pymc/tests/test_idata_conversion.py::TestDataPyMC::test_multivariate_observations` : Already fixed. Xfail decorator can be removed on next release of Aesara.
- [x] `pymc/tests/test_transforms.py::TestElementWiseLogp::test_beta_ordered` : Already fixed. Xfail decorator can be removed on next release of Aesara.
- [x] `pymc/tests/test_sampling.py::TestSamplePPC::test_deterministic_of_observed`: Failing on  specific systems similar to those above. i.e. Float32 Linux systems. This was caused by the new Assertions introduced by aeppl and discussed in https://github.com/aesara-devs/aeppl/issues/84. #5203 Fixes this by making the prior more lenient.
- [x] `pymc/tests/test_posteriors.py::TestSliceUniform` : Failing on  specific systems similar to those above. i.e. Float32 Linux systems. Fixed in #5203 
- [x] `pymc/tests/test_transforms.py::test_interval_near_boundary` : This test produces infinite values using aeppl. Fixed in #5203 
- [ ] `pymc/tests/test_transforms.py::test_chain_jacob_det` : Fails due to precision issue. Values just close to expected. (Deviation of order 10^-6).
  - The log_jac_det of the LogOdds transform is slightly different than the previous implementation and less comparable to the one given by `actual_ljd` function. This difference could be due to either a lower or higher numerical precision of the new log_jac_det, and it is unclear which one it is without access to an evaluation with known error bounds.

We can discuss these over here.

twiecki:
@brandonwillard will cut a new aesara release so that should help with some of these.

twiecki:
@kc611 we have a new aesara release: https://github.com/aesara-devs/aesara/releases/tag/rel-2.2.3 (thanks @brandonwillard !)

ricardoV94:
Closing this given given no complaints/ low priority/ need for investigation

PR #5203, opened 2021-11-18
Unpin aeppl version

### LINKED ISSUE #5191: Test suite failing in latest version of aeppl

ricardoV94:
Closes #5191

The new version also fixes some test regressions in #5088


ricardoV94:
~Toggled as a draft while testing xfails that are passing locally but not failing here on Github with the strict-xfail.~ It was a case of global seeding differing between local testing and when running the whole suite.

kc611 (approved):
Looks like the issue was related to `tags` and has been fixed already.

PR #5207, opened 2021-11-18
Adding coords, dims, and some optimizations to numpyro-based sampler

zaxtax:
This is further work to make the output of `sample_numpyro_nuts` indistinguishable from the default sampler output.

Any feedback on how to test coords and dims from @OriolAbril and @ricardoV94 is more than welcome!

Also interested in what the inference library field should be for this xarray? Should I use numpyro or pymc?

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -124,8 +124,9 @@ def _get_log_likelihood(model, samples):
     for v in model.observed_RVs:
         logp_v = replace_shared_variables([logpt(v)])
         fgraph = FunctionGraph(model.value_vars, logp_v, clone=False)
+        optimize_graph(fgraph, include=["fast_run"], exclude=["cxx_only", "BlasOpt"])
         jax_fn = jax_funcify(fgraph)
-        result = jax.vmap(jax.vmap(jax_fn))(*samples)[0]
+        result = jax.jit(jax.vmap(jax.vmap(jax_fn)))(*samples)[0]
         data[v.name] = result
     return data
 
@@ -149,6 +150,20 @@ def sample_numpyro_nuts(
         var_names = model.unobserved_value_vars
 
     vars_to_sample = list(get_default_varnames(var_names, include_transformed=keep_untransformed))
+
+    coords = {
+        cname: np.array(cvals) if isinstance(cvals, tuple) else cvals
+        for cname, cvals in model.coords.items()
+        if cvals is not None
+    }
+
+    if hasattr(model, "RV_dims"):
+        dims = {
+            var_name: [dim for dim in dims if dim is not None]
+            for var_name, dims in model.RV_dims.items()
+        }
+    else:
+        dims = {}
 
     tic1 = pd.Timestamp.now()
     print("Compiling...", file=sys.stdout)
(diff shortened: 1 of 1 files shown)


ricardoV94 (approved):

ricardoV94:
> Also interested in what the inference library field should be for this xarray? Should I use numpyro or pymc?

I guess Numpyro?

twiecki:
Or maybe `pymc-numpyro`?

zaxtax:
I think it's good to merge either way. But it might be worth somewhere else discussing whether it should be `pymc`, `numpyro`, or `pymc-numypro` and if it's `pymc-numpyro` what version to put there. The version of `pymc` or the version of `numpyro` or some concatenation of those.

twiecki:
This is great, thanks @zaxtax!

What are your experiences with this sampler?

twiecki:
We've also been meaning to add the blackjax sampler, any interest in working on this?

zaxtax:
> This is great, thanks @zaxtax!
> 
> What are your experiences with this sampler?

The numpyro sampler has in my experience been consistently the most performant sampler. If you can use it it's always worth it to do so. I'm curious what the plan is with integrating jax-based samplers. Is the goal to have blackjax, tfp, and numpyro? Or just include blackjax?

twiecki:
Good to know. Probably just blackjax, tfp nuts didn't work well.

zaxtax:
Alright. In that case there needs to be some work on blackjax since it's noticeably slower than numpyro. I don't see any reason there should be any difference between them, but it would be silly to drop numpyro support until blackjax is roughly as performant.

ricardoV94:
> Alright. In that case there needs to be some work on blackjax since it's noticeably slower than numpyro. I don't see any reason there should be any difference between them, but it would be silly to drop numpyro support until blackjax is roughly as performant.

We are not planning to drop numpyro, the idea was to have both options, if they both work. 

What kind of model are you basing your performance comparisons upon?

zaxtax:
I was using non-centered eight schools, following the examples in:

https://github.com/blackjax-devs/blackjax/blob/main/notebooks/use_with_numpyro.ipynb
https://docs.pymc.io/en/v3/pymc-examples/examples/diagnostics_and_criticism/Diagnosing_biased_Inference_with_Divergences.html

twiecki:
> Alright. In that case there needs to be some work on blackjax since it's noticeably slower than numpyro.

That's good to know. CC @rlouf

rlouf:
We're aware of performance issues with the warmup, which should be solved with a design change that @junpenglao is working on. Sampling itself should take about the same time.

rlouf:
@ricardoV94 suggested I
open a PR with a blackjax sampler for PyMC, which I will do once the `AeHMC` NUTS sampler is finalized.

twiecki:
That'd be awesome @rlouf.

zaxtax:
> We're aware of performance issues with the warmup, which should be solved with a design change that @junpenglao is working on. Sampling itself should take about the same time.

That's really neat. What's the best way I can help?

twiecki:
@zaxtax I think you can go ahead an start on adding blackjax similar to how numpyro is in here while @rlouf finishes up `aehmc`.

PR #5208, opened 2021-11-19
Add moment for Simulator distribution

ricardoV94:
#5078 

I went with a simple heuristic: the mean of 10 draws. 

This is all very experimental, because we haven't really investigated how unobserved Simulators work and much less how they work outside of SMC where the moment would be used.

@aloctavodia what do you think?


aloctavodia (approved):
That seems to be a reasonable heuristic

twiecki (approved):

PR #5211, opened 2021-11-19
add moment for BART distribution

aloctavodia:
Following discussion in #5078

### DIFF
--- a/pymc/bart/bart.py
+++ b/pymc/bart/bart.py
@@ -19,7 +19,8 @@ from aeppl.logprob import _logprob
 from aesara.tensor.random.op import RandomVariable, default_shape_from_params
 from pandas import DataFrame, Series
 
-from pymc.distributions.distribution import NoDistribution
+from pymc.distributions.distribution import NoDistribution, _get_moment
+from pymc.distributions.shape_utils import rv_size_is_none
 
 __all__ = ["BART"]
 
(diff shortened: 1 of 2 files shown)


pymc/bart/bart.py
```diff
@@ -132,6 +137,13 @@ def logp(x, *inputs):
         """
         return at.zeros_like(x)
 
+    @classmethod
+    def get_moment(cls, rv, size, *rv_inputs):
```
ricardoV94, line 140: Could BART shape be determined by the shape of the rv_inputs? Or does the mean already integrate this info?

pymc/bart/bart.py
```diff
@@ -132,6 +137,13 @@ def logp(x, *inputs):
...
+        mean = rv.Y.mean()
+        if not rv_size_is_none(size):
```
ricardoV94: Can you add a condition that hits this branch?

pymc/bart/bart.py
aloctavodia: The shape of BART is always the shape of observed data Y.

ricardoV94: And do you do anything with the size variable or is that branch below unnecessary?

aloctavodia: everything is a lie :-) Initially I tough about having a size argument for future extension. But probably that will not be very useful.

ricardoV94 (approved):

PR #5212, opened 2021-11-20
n and p parametrization on Zero Inflated Negative Binomial

farhanreynaldo:
This PR added alternative parametrization (`n` and `p`) on Zero Inflated Negative Binomial based on this issue https://github.com/pymc-devs/pymc/issues/5196.

### DIFF
--- a/pymc/tests/test_distributions_random.py
+++ b/pymc/tests/test_distributions_random.py
@@ -1502,6 +1502,40 @@ class TestZeroInflatedNegativeBinomial(BaseTestDistribution):
     ]
 
 
+class TestZeroInflatedNegativeBinomial(BaseTestDistribution):
+    def zero_inflated_negbinomial_rng_fn(
+        self, size, psi, n, p, negbinomial_rng_fct, random_rng_fct
+    ):
+        return negbinomial_rng_fct(n, p, size=size) * (random_rng_fct(size=size) < psi)
+
+    def seeded_zero_inflated_negbinomial_rng_fn(self):
+        negbinomial_rng_fct = functools.partial(
+            getattr(np.random.RandomState, "negative_binomial"),
+            self.get_random_state(),
+        )
+
+        random_rng_fct = functools.partial(
+            getattr(np.random.RandomState, "random"), self.get_random_state()
+        )
+
+        return functools.partial(
+            self.zero_inflated_negbinomial_rng_fn,
+            negbinomial_rng_fct=negbinomial_rng_fct,
+            random_rng_fct=random_rng_fct,
+        )
+
+    pymc_dist = pm.ZeroInflatedNegativeBinomial
+    pymc_dist_params = {"psi": 0.9, "n": 12, "p": 0.7}
+    expected_rv_op_params = {"psi": 0.9, "n": 12, "p": 0.7}
+    reference_dist_params = {"psi": 0.9, "n": 12, "p": 0.7}
+    reference_dist = seeded_zero_inflated_negbinomial_rng_fn
+    tests_to_run = [
+        "check_pymc_params_match_rv_op",
+        "check_pymc_draws_match_reference",
+        "check_rv_size",
+    ]
+
+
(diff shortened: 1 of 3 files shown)


pymc/tests/test_distributions_random.py
```diff
@@ -1502,6 +1502,40 @@ def seeded_zero_inflated_negbinomial_rng_fn(self):
     ]
 
 
+class TestZeroInflatedNegativeBinomial(BaseTestDistribution):
```
ricardoV94, line 1505: This test can be much smaller. Check other tests for alternative parametrizations in this module. Like `NormalTau`

ricardoV94 (changes requested):
Just did a quick skim. Looks great so far

pymc/tests/test_distributions_random.py
farhanreynaldo: After a quick glance on `NormalTau`, we only need to test on `check_pymc_params_match_rv_op`?

ricardoV94: Yes

ricardoV94 (approved):

ricardoV94:
Some of the tests seen to be failing.

farhanreynaldo:
> Some of the tests seen to be failing.

This line is failing:

```
self.check_logcdf(
    ZeroInflatedNegativeBinomial,
    Nat,
    {"psi": Unit, "p": Unit, "n": NatSmall},
    lambda value, psi, p, n: np.log(
        (1 - psi) + psi * sp.nbinom.cdf(value, n, p)
    ),
)
```

Did I misspecify the `{"psi": Unit, "p": Unit, "n": NatSmall}` this one?

ricardoV94:
From the trace it seems the distribution is not respecting the invalid negative `n` parameter: https://github.com/pymc-devs/pymc/runs/4276276491?check_suite_focus=true#step:8:429

ricardoV94:
Which is strange because it passes with the mu/alpha parametrization

ricardoV94:
Seems like we have a small bug in the logcdf. We have to add a `0 < n` bound condition to the logcdf here: https://github.com/pymc-devs/pymc/blob/dc92865f906e20d6ae9970f49671ae0accf9779a/pymc/distributions/discrete.py#L1710

The case `n <= 0` was not being tested independently before apparently. Perhaps because it led to a negative `p` (when converting from `mu`) which was enough to activate the bound conditions.

farhanreynaldo:
> Seems like we have a small bug in the logcdf. We have to add a `0 < n` bound condition to the logcdf here:
> 
> https://github.com/pymc-devs/pymc/blob/dc92865f906e20d6ae9970f49671ae0accf9779a/pymc/distributions/discrete.py#L1710
> 
> The case `n<0` was not being tested independently before apparently. Perhaps because it led to a negative `p` (when converting from `mu`) which was enough to activate the bound conditions.

I am wondering, so in order to test the `n` and `p` parametrization, the `logcdf` bound (i.e. `n<0`) should be written explicitly. Why can't we rely on the implicit case where negative `n` would yield negative `p`, which activate the `0 < p` bounds?

ricardoV94:
> I am wondering, so in order to test the `n` and `p` parametrization, the `logcdf` bound (i.e. `n<0`) should be written explicitly. Why can't we rely on the implicit case where negative `n` would yield negative `p`, which activate the `0 < p` bounds?

We are not computing `p` from `n` because the new parametrization is already given in terms of `p`.

ricardoV94:
Thanks @farhanreynaldo!

farhanreynaldo:
> > I am wondering, so in order to test the `n` and `p` parametrization, the `logcdf` bound (i.e. `n<0`) should be written explicitly. Why can't we rely on the implicit case where negative `n` would yield negative `p`, which activate the `0 < p` bounds?
> 
> We are not computing `p` from `n` because the new parametrization is already given in terms of `p`.

Ah right.

Issue #5214, opened 2021-11-20
Enforce MvStudentT nu to be a scalar variable


ricardoV94:
#5173 revealed a bit of an edge case in the MvStudentT with broadcasting when `nu` is a vector. Since we haven't tested whether the RV, logp, and logcdf, are doing the right thing in that case, we should assert that `nu` is a scalar in the `make_node` method of the respective RandomVariable.

We can always remove this restriction if we confirm things are working consistently with the rules of broadcasting.

sagartomar:
I would like to work on this. It looks like we need to override the make_node method in MvStudentTRV class and need to check if nu is scalar and if the check passes then call the super class' make_node method.

twiecki:
Great, go for it @sagartomar :+1:.

PR #5220, opened 2021-11-22
Add moments for CAR distribution

zoj613:
This adds moments for the CAR distribution discussed in https://github.com/pymc-devs/pymc/issues/5078#issuecomment-974662268

### DIFF
--- a/pymc/tests/test_distributions_moments.py
+++ b/pymc/tests/test_distributions_moments.py
@@ -9,6 +9,7 @@ from scipy import special
 import pymc as pm
 
 from pymc.distributions import (
+    CAR,
     AsymmetricLaplace,
     Bernoulli,
     Beta,
@@ -109,7 +110,6 @@ def test_all_distributions_have_moments():
     # Distributions that have been refactored but don't yet have moments
     not_implemented |= {
         dist_module.discrete.DiscreteWeibull,
-        dist_module.multivariate.CAR,
         dist_module.multivariate.DirichletMultinomial,
         dist_module.multivariate.Wishart,
     }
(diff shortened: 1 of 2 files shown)


zoj613:
The tests are based on 
https://github.com/pymc-devs/pymc/blob/b8522dc7408e29af4ec7f3b175b7b2991cb7eedb/pymc/tests/test_distributions_random.py#L2369-L2381

@ckrapu @aerubanov @ricardoV94 Could you please take a look at this when you can.

zoj613:
Not sure why the coverage tanked here. I did not make any changes to `parallel_sampling.py`.

pymc/tests/test_distributions_moments.py
```diff
@@ -817,6 +818,34 @@ def test_mv_normal_moment(mu, cov, size, expected):
...
+            np.full((4, 5, 2, 4), [[3.0, 5, 2, 1], [1, 4, 0.5, 9]]),
+        ),
+    ],
+)
+def test_car_moment(mu, size, expected):
+    W = np.array(
+        [[0.0, 1.0, 1.0, 0.0], [1.0, 0.0, 0.0, 1.0], [1.0, 0.0, 0.0, 1.0], [0.0, 1.0, 1.0, 0.0]]
+    )
+    tau = 2
+    alpha = 0.5
```
ricardoV94, line 957: Are tau and alpha constrained to be scalars?

zoj613: just noticed the class docstring allows them to be arrays. I just don't see it being tested anywhere in 
https://github.com/pymc-devs/pymc/blob/99ec0ff93362a6d6eba90cbad49a2338c353eccd/pymc/tests/test_distributions.py#L3162-L3242

and

https://github.com/pymc-devs/pymc/blob/99ec0ff93362a6d6eba90cbad49a2338c353eccd/pymc/tests/test_distributions_random.py#L2368-L2405

which is what this PR's tests are based on. I also don't remember any literature that treats tau is anything more than a scalar.

ricardoV94: If the random/logp methods work with non-scalar inputs for those parameters, we need `get_moment` to do the same. If they are not really supposed to work with those being non-scalar we don't need to, but we might have to open a new issue similar to https://github.com/pymc-devs/pymc/issues/5214

michaelosthege: #5214 was fixed in #5241 - the changeset was rather simple. @zoj613 do you want to add the enforcement of scalar `tau` and `alpha`, or add them as vectors to the test cases?

mjhajharia:
@zoj613 any updates on this?!

mjhajharia:
@zoj613 any updates on this?!

pymc/tests/test_distributions_moments.py
zoj613: According to https://github.com/pymc-devs/pymc/blob/b6f76e54485dcd376acbf188fdc9262ac40456c2/pymc/distributions/multivariate.py#L1999

it looks like tau and alpha are both scalars. The docstring seems wrong about allowing array input for those parameters.

ricardoV94: `ndims_params` indicates the minimum/base dimensions, not whether they can't be of higher dimensionality. Idelly we wouldn't impose any limitations, but the rng_fn/logp methods might not broadcast/handle higer values than the minimum case properly

zoj613: @ckrapu is the `tau` parameter meant to handle array input? I have not seen that case in literature. I also do not see that case being tested in the implementation unittests. I'm asking because it appears you submitted the PR for the distribution. Would you be able to clarify if we need to account for non-scalar tau?

zoj613:
> @zoj613 any updates on this?!

The tests pass, however there is still ambiguity about allowing non-scalar input for the `tau` parameter. The docstring suggests array input as acceptable, but that case is not tested anywhere in the codebase.

twiecki:
@zoj613 Can you add a test for that case?

pymc/tests/test_distributions_moments.py
ricardoV94: @zoj613 I think it's safe to go ahead and add that explicit limitation for the time being. We can always lift it later (and add tests). Do you want to include that in this PR?

zoj613: @ricardoV94  I think it's best that array tau be handled as a separate PR. I believe there is no good motivation to account for array tau given its lack of application in literature. IIRC the paper linked in the doc-string only mentions tau as an array in the case an MCAR model, which I believe is different from what is implemented by CAR.

ricardoV94: Lack of application in the literature shouldn't be a reason to restrict it, but instead whether we handling it properly (and have tests to give us confidence we are). I agree it can be done in a separate PR, would you be interested in doing that? I'll go ahead and merge this one after one last review

ricardoV94 (approved):

Issue #5048, opened 2021-10-07
Interpolated class missing `bound_args_indices`


FabiJa:
Hi,
a have a problem with the ``Interpolated`` class in the development branch. As a starting point I tried the example from the [source](https://github.com/pymc-devs/pymc/blob/328a8eb29e64bc55ac45bae0651a410523912b74/pymc/distributions/continuous.py#L3831):

```
import matplotlib.pyplot as plt
import numpy as np
import pymc as pm
import arviz as az
from scipy.stats import gamma
plt.style.use('arviz-darkgrid')
rv = gamma(1.99)
x = np.linspace(rv.ppf(0.01),rv.ppf(0.99), 1000)
points = np.linspace(x[0], x[-1], 50)
pdf = rv.pdf(points)
interpolated = pm.Interpolated.dist(points, pdf)
fig, ax = plt.subplots(1, 1)
ax.plot(x, rv.pdf(x), 'C0', linestyle = '--',  label='Original Gamma pdf',alpha=0.8,lw=2)
ax.plot(points, pdf, color='black', marker='o', label='Lattice Points',alpha=0.5,linestyle='')
ax.plot(x, np.exp(interpolated.logp(x).eval()),'C1',label='Interpolated pdf',alpha=0.8,lw=3)
r = interpolated.random(size=1000)
ax.hist(r, density=True, alpha=0.4,align ='mid',color='grey')
ax.legend(loc='best', frameon=False)
plt.show()
```

I get the following error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_5967/2908023609.py in <module>
     13 ax.plot(x, rv.pdf(x), 'C0', linestyle = '--',  label='Original Gamma pdf',alpha=0.8,lw=2)
     14 ax.plot(points, pdf, color='black', marker='o', label='Lattice Points',alpha=0.5,linestyle='')
---> 15 ax.plot(x, np.exp(interpolated.logp(x).eval()),'C1',label='Interpolated pdf',alpha=0.8,lw=3)
     16 r = interpolated.random(size=1000)
     17 ax.hist(r, density=True, alpha=0.4,align ='mid',color='grey')

~/anaconda3/envs/pymc3_jax/lib/python3.9/site-packages/pymc/distributions/distribution.py in fn(*args, **kwargs)
    128 def _make_nice_attr_error(oldcode: str, newcode: str):
    129     def fn(*args, **kwargs):
[...]

PR #5222, opened 2021-11-23
Adding Interpolated moment

larryshamalama:
This PR adds moment and a test for `pm.Interpolated` (see issue #5078). However, this PR requires issue #5048 to be addressed first (see corresponding WIP PR #5067).

Currently, the tests yield the following error:

```{python}
ValueError: Must specify bound_args_indices for Interpolated bounded distribution
```

This PR serves as a placeholder to track my progress.

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -3689,13 +3689,8 @@ class Interpolated(BoundedContinuous):
     def get_moment(rv, size, x_points, pdf_points, cdf_points):
         # cdf_points argument is unused
         # moment = at.as_tensor(0.)
-        print(x_points)
-        print("")
-        for pdf in pdf_points.eval():
-            print(pdf)
-        print("")
         moment = at.sum(at.mul(x_points, pdf_points))
-        print(moment.eval())
+
         if not rv_size_is_none(size):
             moment = at.full(size, moment)
 
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -3686,7 +3686,9 @@ class Interpolated(BoundedContinuous):
 
         return super().dist([x_points, pdf_points, cdf_points], **kwargs)
 
-    def get_moment(rv, size, x_points, pdf_points):
+    def get_moment(rv, size, x_points, pdf_points, cdf_points):
+        # cdf_points argument is unused
+        # moment = at.as_tensor(0.)
         moment = at.mul(x_points, pdf_points)
         if not rv_size_is_none(size):
             moment = at.full(size, moment)
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -3689,7 +3689,7 @@ class Interpolated(BoundedContinuous):
     def get_moment(rv, size, x_points, pdf_points, cdf_points):
         # cdf_points argument is unused
         # moment = at.as_tensor(0.)
-        moment = at.mul(x_points, pdf_points)
+        moment = at.sum(at.mul(x_points, pdf_points))
         if not rv_size_is_none(size):
             moment = at.full(size, moment)
 
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -3688,7 +3688,6 @@ class Interpolated(BoundedContinuous):
 
     def get_moment(rv, size, x_points, pdf_points, cdf_points):
         # cdf_points argument is unused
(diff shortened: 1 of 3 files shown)


pymc/distributions/continuous.py
```diff
@@ -3669,6 +3669,13 @@ def dist(cls, x_points, pdf_points, *args, **kwargs):
 
         return super().dist([x_points, pdf_points, cdf_points], **kwargs)
 
+    def get_moment(rv, size, x_points, pdf_points):
+        moment = at.mul(x_points, pdf_points)
+        if not rv_size_is_none(size):
+            moment = at.full(size, moment)
+
```
larryshamalama, line 3695: Are `x_points` and `pdf_points` 1D vectors? [This line](https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/continuous.py#L3658) makes me think that they are

ricardoV94: Yes they have to be. We should assert that in the RV.make_node method

ricardoV94:
@larryshamalama Can you try to rebase your PR on the latest version of main to see if the tests pass now that #5067 was merged?

larryshamalama:
I will do this today!

ricardoV94:
Seems like you are giving invalid values for the spline algorithm that is used inside the Interpolated: https://github.com/pymc-devs/pymc/runs/4374823157?check_suite_focus=true#step:7:742

larryshamalama:
I fixed the increasing `x` issue, I just needed values in `x_points` to be increasing.

However, I am encountering another issue. I don't have full understanding of the `dist` method but I am getting the following error: 

Screen Shot 2021-12-03 at 4 10 01 PM

`rv_inputs` is something like `[x_points, pdf_points, cdf_points]` although only `x_points` and `pdf_points` are given as argument in the distribution. Is this related to the signature of the `dist` method of `Interpolated`?

Reference: https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/continuous.py#L3687

ricardoV94:
For conveniente the interpolated RV actually uses 3 variables, x, pdf, cdf, but we create the last for the user in `.dist`. We could do with just 2 but this just proved easier. 

Anyway the get_moment should expect those 3 in the signature even if it doesn't make use of them.

pymc/tests/test_distributions_moments.py
```diff
@@ -800,6 +801,26 @@ def test_categorical_moment(p, size, expected):
     assert_moment_is_expected(model, expected)
 
 
+@pytest.mark.parametrize(
+    "x_points, pdf_points, size, expected",
+    [
+        (np.array([-1, 1]), np.array([0.4, 0.6]), None, 0.2),
+        (np.array([-4, -1, 3, 9, 19]), np.array([0.1 , 0.15, 0.2 , 0.25, 0.3 ]), None, 8),
+        # (np.array([-22, -4, 0, 8, 13]), np.tile(1 / 5, 5), (5, 3), -np.ones((5, 3))),
```
larryshamalama: The `result` from the moment test is always 1.54589372 despite their dot product being 8. Any thoughts on why I am getting the same, constant error? This may be related to the error that I am encountering in `StickBreakingWeights`

larryshamalama: Solved, the answer to this is that `Interpolated` "connects" the $$(x, f(x))$$ through interpolation. As such, we need to divide the pdf values by `Z`.

larryshamalama:
Given that the pdf is normalized, we do not need that `sum(pdf_points)` equals 1. Should I add another test that has `sum(pdf_points) != 1`?

ricardoV94 (approved):
Looks great, thanks @larryshamalama

pymc/distributions/continuous.py
```diff
@@ -3686,6 +3686,16 @@ def dist(cls, x_points, pdf_points, *args, **kwargs):
 
         return super().dist([x_points, pdf_points, cdf_points], **kwargs)
 
+    def get_moment(rv, size, x_points, pdf_points, cdf_points):
+        # cdf_points argument is unused
+        # moment = at.as_tensor(0.)
```
ricardoV94:
```suggestion
```

ricardoV94 (changes requested):

ricardoV94 (approved):

larryshamalama:
Thanks to you @ricardoV94 for all your help as always

PR #5223, opened 2021-11-24
BART: further changes in sampler 

aloctavodia:
This introduces several changes in the pgbart sampler. The main motivation was to improve convergence  while keeping a good accuracy.
* Increase the number of trees fitted per step from 20% to 100%
* Reset the sampler after certain number of steps. This prevents the sampler to get stuck in a local minima,
* Particles are no longer re-weighted, we just grow n particles as deep as possible (which is actually not very deep). 

From my tests, using `response="linear"` is less beneficial than before. Nevertheless I decide to keep this feature and runs mores test to determine if we should keep it.

**A note** in case someone wants more info:
Given the other changes introduced in this PR the benefit from re-weighting particles seems negligible. This PR makes BART slower, but accuracy and convergence are good.

Future PRs should explore ways to accelerate BART. Something trivial is finding some intermediate value between 20% to 100% of the trees fitted per step. Additionally there is room to find better hyper-parameters during tuning like the standard deviation for the leaf node values, or when to reset the sampler or the number of trees fitted per step. Also now that there is no particle reweighting, the trees can be build in a more efficient way in one single pass. Another route could be reusing particles.

Also given that the fitted trees are not very deep, we should maybe move away from particles sampler into something like [this](https://projecteuclid.org/journals/bayesian-analysis/volume-11/issue-3/Efficient-MetropolisHastings-Proposal-Mechanisms-for-Bayesian-Regression-Tree-Models/10.1214/16-BA999.full), or maybe a mix. To clarify we use a prior proposed by [Rockova](https://arxiv.org/abs/1810.00787) favouring much shallower trees than the original by Chipman. We should call it Bayesian Additive Regression Bushes (BARB) :-) The particle Gibbs sampler shows good performance with deep trees (something we do not have) and/or high dimensional data. So far my tests for "high-dimensional" data has included only examples were many of the variables are actually unrelated to the response variable and we have a mechanism to focus the sampling on the important variables, so for these cases the effective dimensionality is actually reduced.

Just one example 

Old 
old

New 
new

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -95,7 +95,7 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
 - New features for BART:
   - Added linear response, increased number of trees fitted per step [5044](https://github.com/pymc-devs/pymc3/pull/5044).
   - Added partial dependence plots and individual conditional expectation plots [5091](https://github.com/pymc-devs/pymc3/pull/5091).
-  - Modify how particle weights are computed. This improves accuracy of the modeled function (see [5177](https://github.com/pymc-devs/pymc3/pull/5177)).
+  - Modify PGBART sampler. Particles are not longer reweighted and the trees are reset from time to time to avoid getting trap in a local mnima. This improves accuracy of the modeled function and improves convergence (see [5223](https://github.com/pymc-devs/pymc3/pull/5223)).
 - `pm.Data` now passes additional kwargs to `aesara.shared`. [#5098](https://github.com/pymc-devs/pymc/pull/5098)
 - ...
 
--- a/pymc/bart/pgbart.py
+++ b/pymc/bart/pgbart.py
@@ -14,9 +14,6 @@
 
 import logging
 
-from copy import copy
-from typing import Any, Dict, List, Tuple
-
 import aesara
 import numpy as np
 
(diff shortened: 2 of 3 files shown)


junpenglao (approved):

RELEASE-NOTES.md
```diff
@@ -95,7 +95,7 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
 - New features for BART:
   - Added linear response, increased number of trees fitted per step [5044](https://github.com/pymc-devs/pymc3/pull/5044).
   - Added partial dependence plots and individual conditional expectation plots [5091](https://github.com/pymc-devs/pymc3/pull/5091).
-  - Modify how particle weights are computed. This improves accuracy of the modeled function (see [5177](https://github.com/pymc-devs/pymc3/pull/5177)).
+  - Modify PGBART sampler. Particles are not longer reweighted and the trees are reset from time to time to avoid getting trap in a local mnima. This improves accuracy of the modeled function and improves convergence (see [5223](https://github.com/pymc-devs/pymc3/pull/5223)).
```
twiecki, line 98:
```suggestion
  - Modify PGBART sampler. Particles are not longer reweighted and the trees are reset from time to time to avoid getting trap in a local minima. This improves accuracy of the modeled function and improves convergence (see [5223](https://github.com/pymc-devs/pymc3/pull/5223)).
```

aloctavodia:
Closing in favour of  #5229. Sorry for the noise!

PR #5225, opened 2021-11-25
Adds tests and mode for dirichlet multinomial distribution

morganstrom:
The code is almost identical to that for the multinomial distribution (PR #5201)

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -453,9 +453,7 @@ class Dirichlet(Continuous):
         norm_constant = at.sum(a, axis=-1)[..., None]
         moment = a / norm_constant
         if not rv_size_is_none(size):
-            if isinstance(size, int):
-                size = (size,)
-            moment = at.full((*size, *a.shape), moment)
+            moment = at.full(at.concatenate([size, a.shape]), moment)
         return moment
 
     def logp(value, a):
@@ -683,6 +681,19 @@ class DirichletMultinomial(Discrete):
         a = floatX(a)
 
         return super().dist([n, a], **kwargs)
+
+    def get_moment(rv, size, n, a):
+        p = a / at.sum(a, axis=-1)
+        mode = at.round(n * p)
+        diff = n - at.sum(mode, axis=-1, keepdims=True)
+        inc_bool_arr = at.abs_(diff) > 0
+        mode = at.inc_subtensor(mode[inc_bool_arr.nonzero()], diff[inc_bool_arr.nonzero()])
+        # Reshape mode according to base shape (ignoring size)
+        mode = at.reshape(mode, rv.shape[size.size :])
+        if not rv_size_is_none(size):
+            output_size = at.concatenate([size, mode.shape])
+            mode = at.full(output_size, mode)
+        return mode
 
     def logp(value, n, a):
         """
--- a/pymc/tests/test_distributions_moments.py
+++ b/pymc/tests/test_distributions_moments.py
@@ -21,6 +21,7 @@ from pymc.distributions import (
     Constant,
     DensityDist,
     Dirichlet,
+    DirichletMultinomial,
     DiscreteUniform,
     DiscreteWeibull,
     ExGaussian,
@@ -112,7 +113,6 @@ def test_all_distributions_have_moments():
 
     # Distributions that have been refactored but don't yet have moments
     not_implemented |= {
-        dist_module.multivariate.DirichletMultinomial,
         dist_module.multivariate.Wishart,
     }
 
@@ -797,10 +797,7 @@ def test_discrete_weibull_moment(q, beta, size, expected):
         ),
         (
             np.full(shape=np.array([7, 3]), fill_value=np.array([13, 17, 19])),
-            (
-                11,
-                5,
-            ),
+            (11, 5),
             np.broadcast_to([13, 17, 19], shape=[11, 5, 7, 3]) / 49,
         ),
     ],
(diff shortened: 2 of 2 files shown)


pymc/distributions/multivariate.py
```diff
@@ -632,6 +628,22 @@ def dist(cls, n, a, *args, **kwargs):
 
         return super().dist([n, a], **kwargs)
 
+    def get_moment(rv, size, n, a):
+        p = a / at.sum(a, axis=-1, keepdims=True)
+        if a.ndim > 1:
+            n = at.shape_padright(n)
+        if (a.ndim == 1) & (n.ndim > 0):
```
```suggestion
        if (a.ndim == 1) and (n.ndim > 0):
```

ricardoV94 (commented):
I am not sure the shape handling is completely correct with the V4 version of the DM. You may want to check how the `rng_fn` is handling broadcasting. It might suffice to do this line: 

https://github.com/pymc-devs/pymc/blob/9d4691c4ceaa18dd50081db32f38fd1310359d48/pymc/distributions/multivariate.py#L582

This may also apply to the Multinomial distribution:
https://github.com/pymc-devs/pymc/blob/9d4691c4ceaa18dd50081db32f38fd1310359d48/pymc/distributions/multivariate.py#L468

morganstrom:
> I am not sure the shape handling is completely correct with the V4 version of the DM. You may want to check how the `rng_fn` is handling broadcasting. It might suffice to do this line:
> 
> https://github.com/pymc-devs/pymc/blob/9d4691c4ceaa18dd50081db32f38fd1310359d48/pymc/distributions/multivariate.py#L582
> 
> This may also apply to the Multinomial distribution:
> 
> https://github.com/pymc-devs/pymc/blob/9d4691c4ceaa18dd50081db32f38fd1310359d48/pymc/distributions/multivariate.py#L468

Thanks, I wasn't aware of this function! I'm getting incompatible shapes when I try to use it however.. The code you refer to uses a constant (`cls.ndims_params`) that is set to [0, 1], which doesn't seem to work here.. 

https://github.com/pymc-devs/pymc/blob/9d4691c4ceaa18dd50081db32f38fd1310359d48/pymc/distributions/multivariate.py#L571

I think the issue might be that dimension 0 of `n` (in case of an array) is required to match dimension 1 of `a` - this is why we're using `at.shape_padright` for `n` and `at.shape_padleft` for `a` :  

https://github.com/pymc-devs/pymc/blob/9d4691c4ceaa18dd50081db32f38fd1310359d48/pymc/distributions/multivariate.py#L630

ricardoV94:
> I think the issue might be that dimension 0 of `n` (in case of an array) is required to match dimension 1 of `a` - this is why we're using `at.shape_padright` for `n` and `at.shape_padleft` for `a` :
> 
> https://github.com/pymc-devs/pymc/blob/9d4691c4ceaa18dd50081db32f38fd1310359d48/pymc/distributions/multivariate.py#L630

That sounds like an old constraint that is no longer needed in V4. This works just fine on the random side, we might just need to check it also works on the logp:

```python
y = pm.DirichletMultinomial.dist(n=np.arange(1, 1+2*4).reshape(2, 4), a=np.ones(3))
y.eval()
```
```
array([[[1, 0, 0],
        [0, 2, 0],
        [0, 0, 3],
        [3, 0, 1]],
       [[1, 2, 2],
        [0, 5, 1],
        [0, 3, 4],
        [2, 0, 6]]])
```

ricardoV94:
> > I think the issue might be that dimension 0 of `n` (in case of an array) is required to match dimension 1 of `a` - this is why we're using `at.shape_padright` for `n` and `at.shape_padleft` for `a` :
> > https://github.com/pymc-devs/pymc/blob/9d4691c4ceaa18dd50081db32f38fd1310359d48/pymc/distributions/multivariate.py#L630
> 
> That sounds like an old constraint that is no longer needed in V4. This works just fine on the random side, we might just need to check it also works on the logp:
> 
> ```python
> y = pm.DirichletMultinomial.dist(n=np.arange(1, 1+2*4).reshape(2, 4), a=np.ones(3))
> y.eval()
> ```
> 
> ```
> array([[[1, 0, 0],
>         [0, 2, 0],
>         [0, 0, 3],
>         [3, 0, 1]],
>        [[1, 2, 2],
>         [0, 5, 1],
>         [0, 3, 4],
>         [2, 0, 6]]])
> ```

ricardoV94:
I am lifting the constraint of `n` in #5234

ricardoV94:
@morganstrom in #5234 I removed the old constraint on the dimensionality of `n` and `p`. The `get_moment` should share the same flexibility, and we have to do the same for the `Multinoimal`. Do you want to incorporate those updates in this PR?

twiecki:
@morganstrom Any update on this?

morganstrom:
I'm sorry, have been meaning to work on this but haven't had the time or
energy. Perhaps somebody else can pick it up instead?

Den fre 7 jan. 2022 10:58Thomas Wiecki ***@***.***> skrev:

pymc/distributions/multivariate.py
```diff
@@ -629,12 +629,7 @@ def dist(cls, n, a, *args, **kwargs):
         return super().dist([n, a], **kwargs)
 
     def get_moment(rv, size, n, a):
-        p = a / at.sum(a, axis=-1, keepdims=True)
-        if a.ndim > 1:
-            n = at.shape_padright(n)
-        if (a.ndim == 1) & (n.ndim > 0):
-            n = at.shape_padright(n)
-            p = at.shape_padleft(p)
+        p = a / at.sum(a, axis=-1)
```
morganstrom, line 686: @ricardoV94 Here I'm trying to mirror the changes you made in #5234

pymc/tests/test_distributions_moments.py
```diff
@@ -1171,28 +1171,10 @@ def normal_sim(rng, mu, sigma, size):
...
-            np.array([1, 10]),
-            None,
-            np.array([[1, 0, 0, 0], [2, 3, 3, 2]]),
-        ),
-        (
-            np.array([[25, 25, 25, 25], [26, 26, 26, 22]]),
-            np.array([1, 10]),
-            2,
-            np.full((2, 2, 4), [[1, 0, 0, 0], [2, 3, 3, 2]]),
+            np.array([[1, 0, 0, 0], [2, 3, 3, 2]]),     # Dim: 2 x 4
```
morganstrom: I'm getting an error here: `TypeError: Cannot convert Type TensorType(int64, matrix) (of Variable Elemwise{Cast{int64}}.0) into Type TensorType(int64, (False, True, False))`

ricardoV94: It seems moment is returning a 3d tensor?

ricardoV94: You can try to call `get_moment` method directly and evaluate the returned graph, in order to facilitate debugging.

morganstrom: When I do what you suggest (in a debug console), I get a 2d array - confusing!
```
get_moment(model["x"]).eval()
array([[1, 0, 0, 0],
       [2, 3, 3, 2]])
```

ricardoV94: Are you using the latest Aesara version? There was a bug with shapes/casting, that was fixed recently.

ricardoV94: Okay, so in that example the distribution shape is actually `(2, 1, 4)`, but the returned moment is `(2, 4)`. 
```python
pm.DirichletMultinomial.dist(
    a=np.array([[26, 26, 26, 22]]),  # Dim: 1 x 4
    n=np.array([[1], [10]]),  # Dim: 2 x 1
).eval().shape  # (2, 1, 4)
```

ricardoV94:
@morganstrom I rebased from main, so make sure to pull from here before doing further work

pymc/tests/test_distributions_moments.py
ricardoV94: I pushed some changes that I think make the mode consistent with the RV shape. Does it look sensible to you?

This was a great edge case you stumbled upon. We could very well have missed it

ricardoV94 (approved):

PR #5226, opened 2021-11-27
Deprecate those kwargs in `v3` that will break in `v4`

michaelosthege:
The deprecation warnings are working, but the Sphinx build is not.

Also, @mjhajharia please reconsider if you want to contribute your changes back to the original project. They have a tickets about this feature (5 7 and 8 in https://github.com/tantale/deprecated/issues) already.
Also the original [`deprecated` project is getting >2 million downloads per week](https://pepy.tech/project/deprecated), so it would be a more sustainable dependency.


OriolAbril:
The sphinx issues look like indentation errors, could you share the output of `Distribution.__init__.__doc__` to see what the modified yet raw docstring looks like?

michaelosthege:
> The sphinx issues look like indentation errors, could you share the output of `Distribution.__init__.__doc__` to see what the modified yet raw docstring looks like?

Full output with all the line breaks..
```
(pm3v3) E:\Source\Repos\pymc3-v3>python -c "import pymc3;print(pymc3.Distribution.__init__.__doc__)"
Creates a PyMC distribution object.

        Parameters
        ----------
        shape : tuple
            Output shape of the RV.
            Forwarded to the Theano TensorType of this RV.
        dtype
            Forwarded to the Theano TensorType of this RV.
        initval : np.ndarray
            Initial value for this RV.
            In PyMC 4.0.0 this will no longer assign test values to the tensors.
        defaults : tuple
        transform : pm.Transform
        broadcastable : tuple
            Forwarded to the Theano TensorType of this RV.
        dims : tuple
            Ignored.
        testval : np.ndarray
            The old way of specifying initial values assigning test-values.

            .. admonition:: Deprecated
      :class: warning

      Parameter testval deprecated since 3.11.5
              (replaced by `initval` in PyMC 4.0.0)



(pm3v3) E:\Source\Repos\pymc3-v3>
```

michaelosthege:
Looks like a pre-commit script is messing with the environment YAML files in ways that break their syntax.

OriolAbril:
Might be the same pre-commit issue as in https://github.com/pymc-devs/pymc/pull/5057?

michaelosthege:
> Might be the same pre-commit issue as in #5057?

Thanks for the pointer!
I think my last commit fixed the YAML syntax, but the pre-commit needs the same fix you did for `main`. Do you remember what made the difference? Was it just that one line in the pre-commit yaml?

Looks like the docs theme on `v3` is generally not rendering the deprecation warnings like the new theme does?

mjhajharia:
> Looks like the docs theme on `v3` is generally not rendering the deprecation warnings like the new theme does?

yeah these don't look that good, actually every sphinx theme renders these sections differently, @OriolAbril can we customize the v3 theme to make the deprecation warnings more prominent or similar to the latest version

OriolAbril:
> Do you remember what made the difference? Was it just that one line in the pre-commit yaml?

It was stopping to use our own script and using madforhooks one, all important changes were in the pre-commit yaml, but were multipe lines, in addition, after the change there are scripts not used anymore that can be removed but keeping them won't break anything.

> @OriolAbril can we customize the v3 theme to make the deprecation warnings more prominent or similar to the latest version

We can try adding https://github.com/pydata/pydata-sphinx-theme/blob/master/src/pydata_sphinx_theme/assets/styles/_admonitions.scss or a subset of it to https://github.com/pymc-devs/pymc/blob/v3/docs/source/_static/main.css see if it works. I did not know about that at the time but this is yet another pro to add to the list of reasons to not use the old v3 custom sphinx theme.

michaelosthege:
@OriolAbril I fixed the pre-commit and CI. Please take a look :)

@mjhajharia any news on the PR to `deprecated`?

mjhajharia:
nope no updates, i think we can go ahead with this now, i'm pretty available now so if there's more bugs let me know and i'll fix them and i'm making tests soon as well

mjhajharia:
@OriolAbril try adding this content, I think that would work https://github.com/pydata/pydata-sphinx-theme/blob/102f741ae7568f3cbea28eae496169120c33c5ff/src/pydata_sphinx_theme/assets/styles/_admonitions.scss

OriolAbril:
> @OriolAbril try adding this content, I think that would work https://github.com/pydata/pydata-sphinx-theme/blob/102f741ae7568f3cbea28eae496169120c33c5ff/src/pydata_sphinx_theme/assets/styles/_admonitions.scss

This is what I did originally, but scss and css seem to be quite different things, and adding that code in our main.css did not work, only the inital lines were taken into account. I don't know how to add scss directly to sphinx either, only css and js. I have now tried copying from the css via the html style inspector in mozilla, if it works great, if it doesn't I'll leave the admonition ugly, we can make a follow up PR if someone is interested

OriolAbril (dismissed):

michaelosthege:
@OriolAbril I just squashed your commits. Looks like that invalidated your approval..

OriolAbril (approved):

mjhajharia:
this looks very nice, thanks @OriolAbril and @michaelosthege!!

PR #5229, opened 2021-11-30
BART: improve sampling

aloctavodia:
This has essentially the same effect as the changes proposed in #5223 but avoiding the horrible heuristic of resetting the sampler. Instead, one the particle is copied from the previous tree and the leaf values are resampled taking into account the current sum of trees. This PR also increases the number of particles but decreases the number of tress fitted per step. The overall effect is a slower sampler that we currently have but much faster than the change proposed in #5223. Still hold the observation that proposing better trees should lead to decrease the number of necessary particles while keeping the accuracy high and the sampler running smoothly. That is something I will explore in a future PR.


junpenglao (approved):

junpenglao:
Dont forget the release note :)

PR #5231, opened 2021-11-30
Function to optimize prior under constraints

AlexAndorra:
**Ready for review** 🍾 

This PR introduces the `pm.find_optim_prior` function, that finds the optimal prior parameters of a distribution under some constraints.

For instance, sometimes, while building a model, we think "**I need a Gamma that has 95% probability mass between 0.1 and 0.4**". It may take a long time to find an appropriate combination of parameters that give a decent Gamma.
`pm.find_optim_prior` basically answers the question: "**What are the `alpha` and `beta` parameters that fit this constraint?**".

Notes:
-  `pm.find_optim_prior` can be used with **any PyMC distribution for which the log-cdf function is implemented**.
- For now, it **only works for two-parameter distributions**, as there are exactly two constraints (a lower and an upper bound).
- When possible, it uses Aesara and gradients of the PyMC log CDF for more efficient computations. Otherwise, it falls back to Scipy's default.

+ [x] Docstring
+ [x] Tests
+ [x] Mentioned in `RELEASE-NOTES.md`


pymc/find_optim_prior.py
```diff
@@ -0,0 +1,96 @@
...
+        optimal parameters of `pm_dist` fitting the interval constraint.
+    mass: float, default to 0.95
+        Share of the probability mass we want between ``lower`` and ``upper``.
+        Defaults to 95%.
+
+    Returns
+    -------
+    The optimized distribution parameters as a numpy array.
+    """
+    if len(pm_dist.rv_op.ndims_params) != 2:
```
ricardoV94: I still don't get it why you wouldn't be able to optimize a distribution with a single parameter.  There is always one parameter value that minimizes the optimization error. 

We can issue a warning if the final approximation error is too large.

For distributions with more than 3 parameters we could allow users to fix some of the parameters so that there are only 2 to optimize? For instance say we want a StudenT mu and sigma and we fix nu to a constant.

ricardoV94 (commented):
Looks cool, left some suggestions below

ferrine:
I would do the API differently to make that less verbose and elegant. Ideally, something like this:
```python
intercept = pm.Normal.fit("intercept", lower=0.1, upper=0.9)
# That could (in future) be extended as
intercept = pm.Normal.fit("intercept", data=input_data)
```

pymc/find_optim_prior.py
```diff
@@ -0,0 +1,96 @@
...
+import pymc as pm
+
+__all__ = ["find_optim_prior"]
+
+
+def find_optim_prior(
+    pm_dist: pm.Distribution,
+    lower: float,
+    upper: float,
+    init_guess: List[float],
```
ricardoV94: I think this should be a dictionary, sometimes order of dist parameters is not the most intuitive, and some distributions do require keyword arguments because of alternative parametrizations.

Also then you can return a dictionary that you use to initialize the distribution with `pm.Normal(name, **results)`

ricardoV94:
> I would do the API differently to make that less verbose and elegant. Ideally, something like this:
> ```python
> intercept = pm.Normal.fit("intercept", lower=0.1, upper=0.9)
> That could (in future) be extended as
> intercept = pm.Normal.fit("intercept", data=input_data)
> ```

I don't agree. You still need to provide initial parameter guesses, lower/upper can be distribution parameters, and you might want to specify that some parameters should be fixed, as I mentioned above. All that is more difficult if you mix optimization goals with distribution initialization. 

Also are you suggesting this would automatically add a new random variable to the model? I think this suggestion can create a lot of ambiguity, similar to how the `.dist` often confuses people. Specially because you need to play nice with all the additional keywords like dims, size, observed, transform and so on...

ricardoV94 (commented):
Should this go into `util.py`? That's not a very discoverable file name but this is also a very small helper function.

pymc/find_optim_prior.py
AlexAndorra: > We can issue a warning if the final approximation error is too large.

Yeah, something like a warning that's triggered when there is +- 2% difference between the expected and the actual mass? E.g you want 95% of the mass between 1 and 2, but you got 97% (or 93%) of the mass between those bounds. That's super arbitrary of course

> For distributions with more than 3 parameters we could allow users to fix some of the parameters so that there are only 2 to optimize? For instance say we want a StudenT mu and sigma and we fix nu to a constant.

Actually, why not implement the case for 3 parameters? IIUC we just have to add another constraint, which I suggest should be the median. For distributions with more than 3 parameters, then we can allow users to fix some of the parameters, as you suggested

pymc/find_optim_prior.py
AlexAndorra: Good point, will do

pymc/find_optim_prior.py
ricardoV94: We don't have closed form median expression for distributions. In general we could allow for arbitrary constraints but they might not be so intuitive.

I have no idea what the median of a half-normal distribution looks like for instance.

AlexAndorra:
> I would do the API differently to make that less verbose and elegant. Ideally, something like this

@ferrine I agree with @ricardoV94 on this, at least for a first version, where we can try and see how people use that feature and what is missing (or not).

> Should this go into util.py? That's not a very discoverable file name but this is also a very small helper function.

@ricardoV94 I actually put that into `util.py` originally, but got into issues when trying to make `find_optim_prior` accessible at the `pm` level 🤷‍♂️ It's not something I'm a master at, so if someone knows how to do that, I'm down to put it back into `util.py`

pymc/find_optim_prior.py
AlexAndorra: I do. But I don't want to tell you.
Kidding aside, I imagine it's the same for the mean 🤔 So let's go with the two parameters with fix params when there are more

ricardoV94:
You can import it explicitly in `init.py` I think.

pymc/find_optim_prior.py
AlexAndorra: Actually, just testing this with `Exponential`, the optimization seems to work but actually doesn't: if you ask for 95% of the mass between 0.1 and 0.4, you do get a "solution", but then you take random samples from the optimized prior and get 95% of the probability mass between... 0.04 and 4.8 😆

ricardoV94: I wouldn't think of optimizing the exponential with a lower bound other than zero. 

Also I guess the final alpha was also off? So it would show up with the user warning no?

AlexAndorra: Mmmh, I think in the one-parameter case we could fix the lower bound at 0 and then just optimize based on the upper bound (that implies that the one-parameter distributions with a PyMC logcdf are the ones truncated at 0, like Exponential, HalfNormal, etc.).
I can also see that being useful is you only wnat to optimize one of two parameters, e.g I need a sigma so that Normal(1, sigma) has 95% prob mass between 0 and 2.
Does that make sense, and is it even sensible mathematically?

ricardoV94:
I suggest that instead of minimizing two distances: `logcdf(lower) - np.log(alpha/2)` and `logcdf(upper) - np.log(1-alpha/2)`, we just minimize `logdiffexp(logcdf(upper), logcdf(lower)) - np.log(mass)`, which among other things allows one to set the lower point to zero/a value where at the edge of the distribution support.

@aseyboldt do you see any loss in functionality by doing things like this?

pymc/tests/test_util.py
```diff
@@ -174,9 +172,36 @@ def test_find_optim_prior():
...
+            init_guess={"mu": 5, "sigma": 2, "nu": 7},
+        )
+    assert len(record) == 0
+
+    # Exponential without warning
+    with pytest.warns(None) as record:
+        opt_params = pm.find_optim_prior(
+            pm.Exponential, lower=0, upper=1, mass=MASS, init_guess={"lam": 1}
+        )
+    assert len(record) == 0
```
ricardoV94: I would add these to all that are expected to work.

ricardoV94 (commented):
Perhaps add a test with a discrete distribution?

pymc/tests/test_util.py
AlexAndorra: Done

AlexAndorra:
> Perhaps add a test with a discrete distribution?

Done @ricardoV94. I think we can merge if tests pass 🍾

AlexAndorra:
Ah no: I still need to migrate this to utils 🤦‍♂️

AlexAndorra:
I fixed the tests @ricardoV94 , let's see if they pass now.

Also, migrating this to `util.py` doesn't work and raises an `AttributeError: partially initialized module 'pymc' has no attribute 'Distribution' (most likely due to a circular import)` when doing `import pymc`. This is raised by the type hint `pm_dist: pm.Distribution` in `find_optim_prior`, but I really don't understand what the problem is. 
So I suggest keeping this as is -- or I can rename this file `func_utils.py`, to make clear that it could contain other helper functions in the future

ricardoV94:
> I fixed the tests @ricardoV94 , let's see if they pass now.
> 
> Also, migrating this to `util.py` doesn't work and raises an `AttributeError: partially initialized module 'pymc' has no attribute 'Distribution' (most likely due to a circular import)` when doing `import pymc`. This is raised by the type hint `pm_dist: pm.Distribution` in `find_optim_prior`, but I really don't understand what the problem is. So I suggest keeping this as is -- or I can rename this file `func_utils.py`, to make clear that it could contain other helper functions in the future

That usually means you are importing something from another module which is itself importing something from the original module.

AlexAndorra:
It took me some time to figure it out, but I think the issue was that scipy expects float64 but aesara was giving it float32.
The only way I know to do that is to add `aesara.config.floatX = "float64"` in the function code. Tell me if there is better way. Otherwise, we can merge if tests pass @ricardoV94

AlexAndorra:
The only test failing now is [a strange](https://github.com/pymc-devs/pymc/runs/4523115436?check_suite_focus=true#step:7:887) `aeppl.logprob.ParameterValueError: nu > 0, sigma > 0`, which doesn't happen locally 🤔 
Rings any bell?

ricardoV94:
> The only test failing now is [a strange](https://github.com/pymc-devs/pymc/runs/4523115436?check_suite_focus=true#step:7:887) `aeppl.logprob.ParameterValueError: nu > 0, sigma > 0`, which doesn't happen locally 🤔 
> Rings any bell?

If you update to the latest aeppl version you should get the error locally as well. You can use `aesaraf.compile_pymc` which replaces those asserts by a switch

AlexAndorra:
Oh come, just _one_ failing test 😅 Once again, I cannot reproduce it locally though. @ricardoV94 do the tests pass locally for you too?

ricardoV94:
> Oh come, just _one_ failing test 😅 Once again, I cannot reproduce it locally though. @ricardoV94 do the tests pass locally for you too?

Are you testing in float32 and the same aesara/aeppl versions as in the requirements?

AlexAndorra:
I was not... I don't know what's worse: covid or float32 🤔

AlexAndorra:
Ok, the problem is that, in float32, the Poisson doesn't have enough flexibility to get away from the initial guess. I seem to remember it's an issue we noticed in development @ricardoV94 ?

In float64, `pm.find_optim_prior(pm.Poisson, 0, 10, {"mu": 0.5}, mass=0.95)` will yield `{'mu': 3.0016}` without warning, but in float32, the `UserWarning` will be raised even with an initial guess of 2.5 or 3.5

I suggest moving that float32 test to the expected warnings.

AlexAndorra:
Yes, tests are passing, that's a Christmas miracle 🥳 
@ricardoV94 if you agree with my last changes, we can merge! And note that if we do, it'll be an exact round number of commits (i.e 40) -- coincidence...?

pymc/tests/test_util.py
```diff
@@ -142,3 +143,138 @@ def fn(a=UNSET):
...
+            upper=1,
+            mass=0.95,
+            init_guess={"mu": 5, "cov": np.asarray([[1, 0.2], [0.2, 1]])},
+        )
+
+
+distribution = pm.Poisson
+lower = 0
+upper = 10
+if aesara.config.floatX == "float64":
```
ricardoV94: This is probably due to the discrete nature of the poisson. Can't we pick a range of values where the cdf is more "continuous", like those using high values of mu?

AlexAndorra: Yeah that's exactly why, but I hoped you wouldn't pick up on that 😜 I tried a bunch of values but nothing was really working. I'll give it another try ASAP, because it's not as if today is a special day 🤷‍♂️

ricardoV94: Haha I'm sorry but you are on the naughty list. Poisson with high mu becomes something more normal-ish, no? I would make the lower bound not be zero anymore.

AlexAndorra: Yeah that's what I'm trying right now. 
Also, this function doesn't work with `pm.Exponential` right now (because the gradient is broken); I think we should use the scipy default when the user provides that distribution. As Harry Potter would say: agreed?

AlexAndorra: Honestly it's a huge pain to find a working test in float32: even `pm.find_optim_prior(pm.Poisson, 93, 107, {"mu": 100}, mass=0.5)` will raise a warning because we get 51% of the mass in there instead of 50%!
Note that this is only a problem for automated tests here, not for a user, who would probably be happy with that result.

I see two options:
- We merge as is
- We introduce a new kwarg that can relax the tolerance of 1% difference between desired and actual mass that we currently have. And we of course relax that assumption for the Poisson tests

ricardoV94: Yeah lets treat the Poisson separately. Perhaps put it in it the same test and choose a single mass range that works for both float64 and 32? you already found those right?

AlexAndorra: I think that's what I did in the bit of code we're currently commenting on. Or do you mean something else?

ricardoV94: I was thinking something that does not need different logic for float64 and float32

AlexAndorra: Ah ok, then I found one combination that works for both floats, but only with mass=0.95. The lower the mass to optimize, the harder

AlexAndorra: Updated

AlexAndorra:
All green @ricardoV94 🥳 Let's make that Xmas miracle happen!

pymc/func_utils.py
```diff
@@ -0,0 +1,142 @@
...
+import numpy as np
+
+from scipy import optimize
+
+import pymc as pm
+
+__all__ = ["find_optim_prior"]
+
+
+def find_optim_prior(
```
ricardoV94: I think functionality wise this is great, but I wonder about the name. 

Why optimal? What about something like `find_prior_in_range` or `find_constrained_prior` or better yet something else.

Optimal has a connotation that such priors are special, but it seems the `optim` comes from the mere fact that we are using an optimization procedure to find them.

AlexAndorra: Actually the "optim" stands for "optimized", so I should rename the function to `find_optimized_prior` if that sounds good to you.
Otherwise, I also like `find_constrained_prior`. `better_yet_something_else` does have a nice ring to it, but TBH is way too general 😜

ricardoV94: Hmm I see. They are indeed "optimized". Not sure which of the alternative names is more discoverable/intuitive. I am more inclined to "constrained" but cant really say why xD

pymc/func_utils.py
```diff
@@ -0,0 +1,142 @@
...
+            f"You cannot use `find_optim_prior` with {distribution} -- it doesn't have a logcdf "
+            "method yet.\nOpen an issue or, even better, a pull request on PyMC repo if you really "
+            "need it."
+        )
+
+    cdf_error = (pm.math.exp(logcdf_upper) - pm.math.exp(logcdf_lower)) - mass
+    cdf_error_fn = pm.aesaraf.compile_pymc([dist_params], cdf_error, allow_input_downcast=True)
+
+    # PyMC Exponential gradient is failing miserably, need to figure out why
+    if distribution == pm.Exponential:
```
ricardoV94: Can we open an issue? 

Also, not sure about the specific logic for the Exponential. It may be there are many other broken dists but we didn't test them. Can't we just have the open issue for now?

pymc/func_utils.py
AlexAndorra: I'll go with "constrained"

pymc/func_utils.py
AlexAndorra: I'll open the issue. If I remove this logic though, this function will not work with Exponential -- is this what you want?
I think this is a shame as the Exponential is widely used and the optimization works with Scipy's default Jacobian -- it's just less efficient because we can't use PyMC's gradient

ricardoV94: Hopefully we will fix the underlying issue before too late

AlexAndorra: But why not allow using the use of Exponential in the meantime?

pymc/func_utils.py
AlexAndorra: Done

pymc/func_utils.py
AlexAndorra: [Opened the issue](https://github.com/pymc-devs/pymc/issues/5289) and removed the logic

RELEASE-NOTES.md
```diff
@@ -114,7 +114,8 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
   - Modify how particle weights are computed. This improves accuracy of the modeled function (see [5177](https://github.com/pymc-devs/pymc3/pull/5177)).
   - Improve sampling, increase default number of particles [5229](https://github.com/pymc-devs/pymc3/pull/5229).
 - `pm.Data` now passes additional kwargs to `aesara.shared`. [#5098](https://github.com/pymc-devs/pymc/pull/5098)
-- ...
+- The new `pm.find_constrained_prior` function can be used to find optimized prior parameters of a distribution under some
```
twiecki, line 116: Still says "optimized"

pymc/tests/test_util.py
```diff
@@ -142,3 +142,128 @@ def fn(a=UNSET):
...
+
+# test Poisson separately -- hard to optimize precisely when float32
+@pytest.mark.parametrize(
+    "lower, upper, init_guess",
+    [
+        (1, 15, {"mu": 10}),
+        (19, 41, {"mu": 30}),
+    ],
+)
+def test_optim_prior_poisson(lower, upper, init_guess):
```
twiecki: `optim`

pymc/tests/test_util.py
```diff
@@ -142,3 +142,128 @@ def fn(a=UNSET):
...
+
+@pytest.mark.parametrize(
+    "distribution, lower, upper, init_guess, fixed_params",
+    [
+        (pm.Gamma, 0.1, 0.4, {"alpha": 1}, {"beta": 10}),
+        (pm.Exponential, 0.1, 1, {"lam": 1}, {}),
+        (pm.Binomial, 0, 2, {"p": 0.8}, {"n": 10}),
+    ],
+)
+def test_find_optim_prior_error_too_large(distribution, lower, upper, init_guess, fixed_params):
```
twiecki: these and other test functions still have `optim`

twiecki (changes requested):

RELEASE-NOTES.md
AlexAndorra: Well here I think it's appropriate: the function really returns "optimized prior parameters under some constraints" -- it doesn't say that they are perfect, or the best, or optimal

pymc/tests/test_util.py
AlexAndorra: Good catch, done

pymc/tests/test_util.py
AlexAndorra: Done

twiecki (approved):

pymc/func_utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+    cdf_error = (pm.math.exp(logcdf_upper) - pm.math.exp(logcdf_lower)) - mass
+    cdf_error_fn = pm.aesaraf.compile_pymc([dist_params], cdf_error, allow_input_downcast=True)
+
+    try:
+        aesara_jac = pm.gradient(cdf_error, [dist_params])
+        jac = pm.aesaraf.compile_pymc([dist_params], aesara_jac, allow_input_downcast=True)
+    # when PyMC cannot compute the gradient
+    # TODO: use specific gradient, not implemented exception
+    except Exception:
```
ricardoV94: I think the exception we want here is the one found in `aesara.gradient.NullTypeGradError` as well as `NotImplementedError`
```suggestion
    except (NotImplementedError, NullTyppeGradError):
```

AlexAndorra: The exception that's thrown is a `TypeError` by the Aesara `grad` method (line 501) in `aesara/gradient.py`:
```python
if cost is not None and cost.ndim != 0:
>           raise TypeError("cost must be a scalar.")
```
I added that error in the Except clause and tests pass locally. `aesara.gradient.NullTypeGradError` and `NotImplementedError` don't seem to be raised but I kept them in case they are by other cases we may have forgotten

ricardoV94: We shouldn't catch that TypeError. That means we produced a wrong input to aesara grad. 

The other two exceptions are the ones that (should) appear when a grad is not implemented for an Op.

ricardoV94 (changes requested):
See my reply above. We shouldn't catch the TypeError

pymc/func_utils.py
AlexAndorra: That means we should make these two exceptions appear then, shouldn't we? Because they are not raised right now -- only the TypeError is raised
(here is to my first GH comment of the year 🥂 )

ricardoV94: > That means we should make these two exceptions appear then, shouldn't we? Because they are not raised right now -- only the TypeError is raised
> (here is to my first GH comment of the year 🥂 )

Those two exceptions mean there is no grad implemented for some Op in the cdf which can very well happen and its a good reason to silently default to the scipy approximation. The TypeError, on the other hand, should not be catched, as that means we did something wrong. 

In fact there was a point during this PR when it was always silently defaulting to the scipy approximation because we were passing two values to grad and suppressing the TypeError.

AlexAndorra: Ooooh ok, I see, thanks. So you're saying that the problem we have with Exponential here is because our code is wrong, not only because the grad is failing in PyMC?

ricardoV94: I think the exponential is a different issue altogether. Or are you getting a TypeError during compilation for the exponential tests?

AlexAndorra: Ah no it's for the Poisson that I get the TypeError. It's because we don't have a gradient for this one, right? If yes, I would expect one of the two other errors to be raised (or, here, silently default to Scipy)

ricardoV94: ~Okay, so that probably means there is some bug in an Op used by the Poisson logcdf. It should not raise a TypeError.~

ricardoV94: Okay, so it seems the Poisson logcdf is not returning a scalar variable, and that was behind the TypeError you were seeing. Good reason to not catch it!

ricardoV94:
```python
import pymc as pm
pm.logcdf(pm.Poisson.dist(1), 1).eval()
# array([-0.30685282])
```

ricardoV94: #5307 should fix it!

ricardoV94: I wonder if using the symbolic gradient gets rid of the numerical issues we were seeing in float32

AlexAndorra: Lol, such a silly error. 
> I wonder if using the symbolic gradient gets rid of the numerical issues we were seeing in float32

That sounds promising! How would you do that?

ricardoV94: > Lol, such a silly error.
> 
> > I wonder if using the symbolic gradient gets rid of the numerical issues we were seeing in float32
> 
> That sounds promising! How would you do that?

Check if the Poisson test passes with the original precision of 1e-5 instead of the 1e-2

ricardoV94: That seems to be the case. Locally it is passing in float64/float32 and 1e-5 precision, so we don't need to have the separate test just for the Poisson anymore.

AlexAndorra: Ha ha damn, I just pushed without that change. Tests are indeed passing locally. Gonna refactor the tests and push again

AlexAndorra: Hot damn, tests are passing locally 🔥 Pushed!
Why does the symbolic gradien help so much with numerical errors?

ricardoV94: Because the logcdf uses gammaincc whose gradient is notoriously tricky. We somewhat recently added a numerically stable(r) implementation to Aesara

ricardoV94: https://github.com/aesara-devs/aesara/pull/513

ricardoV94 (approved):
We should move the new tests to a file `test_func_utils.py` if `func_utils.py` is where the functions are going to live.

AlexAndorra:
Good point, just did it

pymc/func_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    fixed_params: Dict[str, float], Optional, default None
+        Only used when `pm_dist` has at least three parameters.
+        Dictionary of fixed parameters, so that there are only 2 to optimize.
+        For instance, for a StudenT, you fix nu to a constant and get the optimized
+        mu and sigma.
+
+    Returns
+    -------
+    The optimized distribution parameters as a dictionary with the parameters'
+    name as key and the optimized value as value.
```
ricardoV94, line 70: Follow up PR (not to be sadistic with this one): we should add a code example in the docstrings

AlexAndorra: I'm not preaching for my choir here, but I actually should add that here. **Don't merge in the meantime**. Will ping when done

AlexAndorra: Done @ricardoV94 -- we can merge once tests pass

PR #5232, opened 2021-12-01
Vonmises moments

Domenico89:
Add moments and tests for the below distributions as part of #5078:
-pymc.distributions.continuous.VonMises

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2924,6 +2924,12 @@ class VonMises(CircularContinuous):
         assert_negative_support(kappa, "kappa", "VonMises")
         return super().dist([mu, kappa], *args, **kwargs)
 
+    def get_moment(rv, size, mu, kappa):
+        mu, _ = at.broadcast_arrays(mu, kappa)
+        if not rv_size_is_none(size):
+            mu = at.full(size, mu)
+        return mu
+
 
 class SkewNormalRV(RandomVariable):
     name = "skewnormal"
--- a/pymc/tests/test_distributions_moments.py
+++ b/pymc/tests/test_distributions_moments.py
@@ -54,6 +54,7 @@ from pymc.distributions import (
     Triangular,
     TruncatedNormal,
     Uniform,
+    VonMises,
     Wald,
     Weibull,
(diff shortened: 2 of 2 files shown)


Domenico89:
The curren tests should work fine, but if I use np.arange(5) as mu in line 446 ,f or example, I get as a result array([ 0.,  1., 2.,  3., -2.28318531]), which is not what I would expect. I couldn't figure out where -2.2831853 is coming from. Other strange numbers appear when I run the test for np.arange(k) with k>4.

ricardoV94:
> The curren tests should work fine, but if I use np.arange(5) as mu in line 446 ,f or example, I get as a result array([ 0., 1., 2., 3., -2.28318531]), which is not what I would expect. I couldn't figure out where -2.2831853 is coming from. Other strange numbers appear when I run the test for np.arange(k) with k>4.

Seems like it is just mirroring around `+-np.pi`? Are you testing the moment with a transformed distribution?

ricardoV94:
All tests are passing except for a pre-commit issue. I added a commit fix it with the bot, so make sure to merge from here if you want to do anything with this branch.

I guess you were seeing the effects of the CircularTransform?
```python
with pm.Model() as m:
  x = pm.Vonmises("x", np.arange(4))  
  y = pm.Vonmises("y", np.arange(4), transform=None)
```

The moment of `x` will by default be on the Circular scale, whereas that of `y` will not. We are testing with the second case, when we call this line:

https://github.com/pymc-devs/pymc/blob/1d337ca57d0fa348e67b73c05e7c678220c31612/pymc/tests/test_distributions_moments.py#L90

ricardoV94 (approved):

ricardoV94:
Thanks @Domenico89, everything seemed to be working in our tests here.

Let me know if you find issues that need to be fixed or if my hunch above was correct.

Domenico89:
> Thanks @Domenico89, everything seemed to be working in our tests here.
> 
> Let me know if you find issues that need to be fixed or if my hunch above was correct.


ah, ok. Sorry if took this long, I am new with the base code. Thanks for the hunch!

PR #5233, opened 2021-12-01
Add rewrites to replace or remove Aeppl CheckParameterValue Ops

### LINKED ISSUE #4429: Interaction between logcdf and new check_bounds flag
The new check_bounds flag can be used to skip the parameter and value checks done by the `bound` method of most logps (as intended), but it also affects the logcdf methods of distributions evaluated within a model context. 

Unlike most logps, however, the logcdf is usually always defined over the real range, even if the respective pdf is only defined/discussed within a finite/non-real domain. Thus, it might make sense to have the cdf give the right result (for valid parameters) even when check_bounds is False.

```python
In [4]: with pm.Model(check_bounds=False) as m: 
   ...:     dist = pm.Bernoulli.dist(p=.1) 
   ...:     print(dist.logcdf(-2).eval()) 
   ...:                                                                         
-0.10536051565782631
```

This would require changing most of the logcdf methods to apply the bound on distribution parameters, but not on the input values, as is done now. For instance, in the Bernoulli example:

https://github.com/pymc-devs/pymc3/blob/89916ad724ae88ebb4c1af3d65919b87b318f180/pymc3/distributions/discrete.py#L473-L482

The function would be changed to:
```python
return bound(
    tt.switch(
        tt.le(value, 0),
        -np.inf,
        tt.switch(
            tt.lt(value, 1),
            tt.log1p(-p),
            0,
        ),
    ),
    0 <= p,
    p <= 1,
)
```

So that the logcdf does not return the wrong result when check_bounds is False due to the input value (but can still do due to invalid parameters). I think this may make the check_bounds flag behavior more consistent/intuitive. I would also extend the docstrings of the check_bounds flag to explain the logcdf implications.

This might become more relevant if, and when, we implement a generic Truncated and Censored class as discussed in #1864.

If people agree this makes sense, I am happy to do a PR for it.

### LINKED ISSUE #5204: Create a GH Action to update pinned `aeppl` version every time there's a new release
Every now and then we need to pin/unpin Aesara and aeppl versions, because something breaks (#5194 #5203 #4963).

Instead of that manual work can't we just keep the version pinned and have a bot open a new PR every time there's a new release?

Then we don't break main, but can swiftly upgrade if there are no blockers.

@twiecki @MarcoGorelli you strike me as the GitHub robot tamers around here. Can you help out?

### LINKED ISSUE #5205: Sampling issues due to Aeppl logp assertions
As discussed in https://github.com/aesara-devs/aeppl/issues/84, Aeppl introduced explicit Aesara asserts in the logp graphs of several distributions. Until a rewrite is implemented ~there or~ here, we can expect to see some `AssertionErrors` crashing `pm.sampling`. 

For instance, one of the test regression documented in #5088 was precisely because of this.

ricardoV94:
This PR adds rewrites to replace or remove Aeppl CheckParameterValue Ops in logprob expressions.

Closes #5205
Closes #5204
Closes #4429 

* These rewrites are included by default when calling pymc.aesaraf.compile_pymc, which was previously named compile_rv_inplace
* pymc.distributions.dist_math.bound was renamed to check_parameters and now returns an expression wrapped in the CheckParameterValue Op
* The Model.check_bounds flag now only affects graphs at compilation time when pymc.aesaraf.compile_pymc is called within a Model context and that flag is set to False.
* Most logp/logcdf methods were changed so that the bounding of the value variable is done separately from the parameter checks (missing this in a couple of multivariate distributions). This is in line with how aeppl defines logprob graphs, and to a lesser extent how scipy does things as well. It also provides a solution to #4429
* Added explicit tests for univariate value and parameter bounds in check_logp, similar to what was already done in check_logcdf, as well as some specialized tests for non-scalar parameters/values in multivariate distributions.

The new `dist_math.check_parameters` always compresses the conditions to a scalar since that's required by `CheckParameterValue`. That means that when evaluating a logp vector, either all results get replaced by `-np.inf` or none do. It no longer switches only those that corresponded to invalid parameters/values.

I don't think it makes sense to generate logprob graphs differently than Aeppl does. If we want to keep the old format we should then reintroduce our own versions of `logp`/`logcdf` to replace those that are defined in Aeppl (many common distributions are defined there), or else we will have a mix of expressions that follow distinct logics.


pymc/aesaraf.py
```diff
@@ -899,11 +901,66 @@ def take_along_axis(arr, indices, axis=0):
...
+        "local_remove_check_parameter",
+        local_remove_check_parameter,
+        use_db_name_as_tag=False,
+    )
+
+    aesara.compile.optdb[database].register(
+        "local_check_parameter_to_ninf_switch",
+        local_check_parameter_to_ninf_switch,
+        use_db_name_as_tag=False,
+    )
```
brandonwillard: Why add these to all those databases?

Also, don't forget that you can set a priority that affects the order in which they're run.

ricardoV94: I just copied what I've seen elsewhere.

Can you give some recommendation on what is the minimum database(s) where these should be registered, as well as what the priority should be?

ricardoV94:
The failing dirichlet_multinomial tests should be solved by #5234

pymc/aesaraf.py
brandonwillard: You should be able to put those in just the pass named `"useless"` and have them removed before the other passes.  Check out `aesara.compile.mode` to see the passes and their ordering.

Otherwise, I would think that removing those checks should be a user-configurable option, and, if so, you might not want to include them by default.  Instead, an option could be set that includes those rewrites when graphs are compiled by PyMC (e.g. by constructing a `Mode` object for `aesara.function` that uses `includes=...`).

ricardoV94: > Otherwise, I would think that removing those checks should be a user-configurable option, and, if so, you might not want to include them by default. Instead, an option could be set that includes those rewrites when graphs are compiled by PyMC (e.g. by constructing a Mode object for aesara.function that uses includes=...).

The rewrites are not running by default due to the `use_db_name_as_tag=False` option. I am manually including them in `compile_pymc` helper function only. Otherwise the new tests that expect the specific Exception would fail.

ricardoV94: > You should be able to put those in just the pass named "useless" and have them removed before the other passes. Check out aesara.compile.mode to see the passes and their ordering.

For some reason putting it in "useless" alone is not doing anything, but putting them in "canonicalize" seems to do the job.

brandonwillard: Looks like we need to refactor that `OptimizationDatabase` code; the whole `use_db_name_as_tag` thing, and the special way it works with only one DB implementation (i.e. `EquilibriumDB`), is not good.  Just the fact that there's a note about a specific subclass in the base class is a bad sign.

Anyway, `"useless"` doesn't use the same kind of underlying `OptimizationDatabase` types as `"canonicalize"`, so I'm guessing that the reason the latter works and the former doesn't is related to that.

twiecki (approved):

PR #5234, opened 2021-12-02
Tweak DirichletMultinomial logp and refactor some multivariate logp tests

ricardoV94:
Couple of tweaks/fixes to the logp /docstrings of Multinomial and DirichletMultinomial

Summary:
1. Remove unnecessary constraints on `n` and `a` in `DirichletMultinomial`, related to discussion in #5225. This also solves a minor issue where the logp was returning a 1D vector for the base case instead of just a scalar.
2. Remove similar old reference about constraints on `n` and `p` in the docstrings of the `Multinomial`
3. Simplify vectorized logp tests for the two distributions as well as the `Dirichlet`

Also:
4. Remove couple old/redundant random tests


ricardoV94:
I don't get why we would automatically normalize the vector p in the Multinomial, but I'm happy to revert those changes if they cause too much trouble. It hides the same problems that were raised in this old issue for the Categorical: https://github.com/pymc-devs/pymc/issues/2082#issuecomment-297243611

AlexAndorra:
Agreed, the implicit normalization is sneaky and can cause issues down the road

ricardoV94:
If we changed this, it would make sense to also not normalize things in the Categorical by default...

ricardoV94:
I reverted the changes to the Multinomial normalization, I'll open a discussion on this for the Multinomial and the Categorical. In the meantime this can hopefully be merged.

pymc/distributions/multivariate.py
```diff
@@ -507,13 +507,11 @@ class Multinomial(Discrete):
...
     Parameters
     ----------
-    n: int or array
-        Number of trials (n > 0). If n is an array its shape must be (N,) with
-        N = p.shape[0]
-    p: one- or two-dimensional array
-        Probability of each one of the different outcomes. Elements must
-        be non-negative and sum to 1 along the last axis. They will be
-        automatically rescaled otherwise.
+    n: int
```
AlexAndorra, line 511: But `n`can be a vector too, can't it? E.g if the number of trials vary by observation

pymc/distributions/multivariate.py
```diff
@@ -626,17 +624,12 @@ class DirichletMultinomial(Discrete):
...
     Parameters
     ----------
-    n : int or array
-        Total counts in each replicate. If n is an array its shape must be (N,)
-        with N = a.shape[0]
-
-    a : one- or two-dimensional array
-        Dirichlet parameter. Elements must be strictly positive.
-        The number of categories is given by the length of the last axis.
+    n : int
```
AlexAndorra, line 629: Same remark as for `Multinomial`

AlexAndorra (commented):
Wow, those tests are much simpler now 🤩 
LGTM @ricardoV94 -- I just had a question about the `n` parameter. Once that's settled we can merge

pymc/distributions/multivariate.py
ricardoV94: Yeah n can be a vector or a matrix, anything as long as dimensions broadcast properly. I put the int to indicate the base case, but perhaps the best is to remove any information of shape and only mention the meaning of the parameters?

AlexAndorra: Why not `int, vector, matrix` ? That's more explicit

ricardoV94: I don't find that very helpful. Other multivariate distributions use an ambiguous "array" and most univariate distributions say only "int" or "float" even though they also support vectors and tensors of any shape.

ricardoV94:
@AlexAndorra I added a commit that harmonizes the docstrings of the 3 distributions. The emphasis is on the dtype of each parameter and the special meaning of the last axis of `a` and `p`, which determines the number of categories.

I rather not make it explicitly that `n` can be multidimensional, and let users expect this to work as it does in other distributions. Otherwise I think it creates more confusion to have, `p: int, vector, matrix, ...`, since these can be of arbitrary dimensionality as long as they broadcast well with `a`.

AlexAndorra (approved):
Yeah, that looks better now, thanks @ricardoV94

PR #5241, opened 2021-12-05
Added check that nu must be a scalar in MvStudentTRV

sagartomar:
This should fix issue #5214

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -263,7 +263,7 @@ class MvStudentTRV(RandomVariable):
     _print_name = ("MvStudentT", "\\operatorname{MvStudentT}")
 
     def make_node(self, rng, size, dtype, nu, mu, cov):
-        nu = at.as_tensor_variable(floatX(nu))
+        nu = at.as_tensor_variable(nu)
         if not nu.ndim == 0:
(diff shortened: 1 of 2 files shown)


pymc/distributions/multivariate.py
```diff
@@ -265,6 +265,13 @@ class MvStudentTRV(RandomVariable):
     dtype = "floatX"
     _print_name = ("MvStudentT", "\\operatorname{MvStudentT}")
 
+    def make_node(self, rng, size, dtype, nu, mu, cov):
+        nu = at.as_tensor_variable(floatX(nu))
```
```suggestion
        nu = at.as_tensor_variable(nu)
```
I think we don't need the `floatX` here, that is something that belongs to PyMC distributions.

ricardoV94 (changes requested):
Looks great. I left a suggestion below, and we should add a test that covers this branch.

Perhaps in `test_distributions_random` you can add something along the lines of the following one:

https://github.com/pymc-devs/pymc/blob/95bd5e59f940cfbd4076dd21d93b0e2aec270add/pymc/tests/test_distributions_random.py#L1622-L1632

michaelosthege (approved):
Looks good!
I like the `re.escape(msg)` it always takes me ages to find the right escape syntax..

ricardoV94 (approved):
Looks great

PR #5244, opened 2021-12-07
documenting get_moment

### LINKED ISSUE #5210: Document `get_moment`
With #5140 the `get_moment` function will be a requirement for all standard distributions and we have to document it properly in a place that users can find easily as well as in this [dev guide](https://github.com/pymc-devs/pymc/blob/77f2c9844bcfaf451648289e9e30ec7475dfcce1/docs/source/developer_guide_implementing_distribution.md).

amyoshino:
This PR updates the file `developer_guide_implementing_distribution.md` documenting the `get_moment()` method based on the issue #5210 

@ricardoV94 I included your instructions from #5078 in the docs where I think they fit better in. Let me know what you think or if I missed anything.

Closes #5210

### DIFF
--- a/docs/source/contributing/developer_guide_implementing_distribution.md
+++ b/docs/source/contributing/developer_guide_implementing_distribution.md
@@ -118,7 +118,7 @@ After implementing the new `RandomVariable` `Op`, it's time to make use of it in
 PyMC 4.x works in a very {term}`functional <Functional Programming>` way, and the `distribution` classes are there mostly to facilitate porting the `PyMC3` v3.x code to the new `PyMC` v4.x version, add PyMC API features and keep related methods organized together.
 In practice, they take care of:
 
-1. Linking ({term}`Dispatching`) a rv_op class with the corresponding logp and logcdf methods.
+1. Linking ({term}`Dispatching`) a rv_op class with the corresponding `get_moment`, `logp` and `logcdf` methods.
 1. Defining a standard transformation (for continuous distributions) that converts a bounded variable domain (e.g., positive line) to an unbounded domain (i.e., the real line), which many samplers prefer.
 1. Validating the parametrization of a distribution and converting non-symbolic inputs (i.e., numeric literals or numpy arrays) to symbolic variables.
 1. Converting multiple alternative parametrizations to the standard parametrization that the `RandomVariable` is defined in terms of.
@@ -152,6 +152,14 @@ class Blah(PositiveContinuous):
         # The first value-only argument should be a list of the parameters that
         # the rv_op needs in order to be instantiated
         return super().dist([param1, param2], **kwargs)
+
+    # get_moment returns the stable moment from which to start sampling for the
+    # distribution given the implicit `rv`, `size` and `param1` ... `paramN`
+    def get_moment(rv, size, param1, param2):
+        moment, _ = at.broadcast_arrays(param1, param2)
+        if not rv_size_is_none(size):
+            moment = at.full(size, moment)
+        return moment
 
     # Logp returns a symbolic expression for the logp evaluation of the variable
     # given the `value` of the variable and the parameters `param1` ... `paramN`
@@ -188,6 +196,8 @@ Some notes:
    overriding `__new__`.
 1. As mentioned above, `PyMC` v4.x works in a very {term}`functional <Functional Programming>` way, and all the information that is needed in the `logp` and `logcdf` methods is expected to be "carried" via the `RandomVariable` inputs. You may pass numerical arguments that are not strictly needed for the `rng_fn` method but are used in the `logp` and `logcdf` methods. Just keep in mind whether this affects the correct shape inference behavior of the `RandomVariable`. If specialized non-numeric information is needed you might need to define your custom`_logp` and `_logcdf` {term}`Dispatching` functions, but this should be done as a last resort.
 1. The `logcdf` method is not a requirement, but it's a nice plus!
+1. Currently only one moment is supported in the `get_moment` method, and probably the "higher-order" one is the most useful (that is `mean` > `median` > `mode`)... You might need to truncate the moment if you are dealing with a discrete distribution.
+1. When creating the `get_moment` method, we have to be careful with `size != None` and broadcast properly when some parameters that are not used in the moment may nevertheless inform about the shape of the distribution. E.g. `pm.Normal.dist(mu=0, sigma=np.arange(1, 6))` returns a moment of `[mu, mu, mu, mu, mu]`.
 
 For a quick check that things are working you can try the following:
 
@@ -201,6 +211,10 @@ blah = pm.Blah.dist([0, 0], [1, 2])
 # Test that the returned blah_op is still working fine
 blah.eval()
 # array([0.62778803, 1.95165513])
+
+# Test the get_moment
+pm.distributions.distribution.get_moment(blah).eval()
+# array([0.5, 1. ])
 
 # Test the logp
 pm.logp(blah, [1.5, 1.5]).eval()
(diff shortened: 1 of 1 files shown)


docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -202,6 +212,10 @@ blah = pm.Blah.dist([0, 0], [1, 2])
 blah.eval()
 # array([0.62778803, 1.95165513])
 
+# Test the get_moment
+pm.distributions.distribution.get_moment(blah).eval()
```
ricardoV94:
```suggestion
from pm.distributions.distribution import get_moment
get_moment(blah).eval()
```

docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -350,8 +364,56 @@ def test_blah_logcdf(self):
...
+1. Moments return `expected` values 
+1. Moments have the expected size and shape
+
+```python
+
+import pytest
+from pymc.distributions import Blah
+from pymc.initial_point import make_initial_point_fn
+
+def assert_moment_is_expected(model, expected):
```
ricardoV94: We shouldn't copy the body of `assert_moment_is_expected`, just mention it exists, what it does and and how it should be used (which you show below)

ricardoV94 (commented):
Looks great! I left some suggestions below

docs/source/contributing/developer_guide_implementing_distribution.md
amyoshino: Thanks for the review! That makes sense if eventually the function `assert_moment_is_expected` changes in the test file we don't have to track the changes and update the documentation. I will work on these changes a commit them soon.

amyoshino:
I changed and committed a new file based on your suggestions. 
I re-read it all and found two places I forgot to mention the `get_method` method in the objectives of this document, those were added as well. 

Again, thanks for reviewing!

docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -202,6 +213,10 @@ blah = pm.Blah.dist([0, 0], [1, 2])
 blah.eval()
 # array([0.62778803, 1.95165513])
 
+# Test the get_moment
+get_moment(blah).eval()
+# array([0.5, 1. ])
```
ricardoV94: These would be `array([0., 0.])` with the pseudo code above no?

amyoshino: This is what I'm getting from `get_moment()` for `blah = pm.Uniform.dist([0, 0], [1, 2])` 

I did some tests and I think this is the right output. Considering that the first argument is location and second is scale (or lower and upper), then first interval is `[0, 1]` and second interval is `[0, 2]`. With the formula `mean = 0.5 * [a, b]`, then we have `[0.5, 1.0]`. I compared with scipy implementation and got the same result. Let me now if I am mistaken here :)

```python
import pymc as pm
from pymc.distributions.distribution import get_moment
from scipy.stats import uniform 

# PyMC implementation
blah = pm.Uniform.dist([0, 0], [1, 2])
get_moment(blah).eval()
>> array([0.5, 1. ])

# spicy implementation
uniform.mean([0, 0], [1, 2])
>> array([0.5, 1. ])
```

I see that there is an error in `pre-commit / pre-commit (pull_request)` that is related with `Trim Trailing Whitespace`. Is this something I did wrong? Let me know how I can fix it.

Thank you,

amyoshino:
I think I was able to solve the error in `pre-commit / pre-commit (pull_request)` that is related with `Trim Trailing Whitespace`. I ran the pre-commit locally and it looks like passing now. 

My sublime text wasn't automatically trimming trailing whitespace on save. 

image

docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -4,15 +4,15 @@ This guide provides an overview on how to implement a distribution for version 4
 It is designed for developers who wish to add a new distribution to the library.
 Users will not be aware of all this complexity and should instead make use of helper methods such as (TODO).
 
-PyMC {class}`~pymc.distributions.Distribution` build on top of Aesara's {class}`~aesara.tensor.random.op.RandomVariable`, and implement `logp` and `logcdf` methods as well as other initialization and validation helpers, most notably `shape/dims`, alternative parametrizations, and default `transforms`.
+PyMC {class}`~pymc.distributions.Distribution` build on top of Aesara's {class}`~aesara.tensor.random.op.RandomVariable`, and implement `logp`, `logcdf` and `get_moment` methods as well as other initialization and validation helpers, most notably `shape/dims`, alternative parametrizations, and default `transforms`.
```
michaelosthege:
```suggestion
PyMC {class}`~pymc.distributions.Distribution` builds on top of Aesara's {class}`~aesara.tensor.random.op.RandomVariable`, and implements `logp`, `logcdf` and `get_moment` methods as well as other initialization and validation helpers.
Most notably `shape/dims` kwargs, alternative parametrizations, and default `transforms`.
```

docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -153,6 +153,14 @@ class Blah(PositiveContinuous):
         # the rv_op needs in order to be instantiated
         return super().dist([param1, param2], **kwargs)
 
+    # get_moment returns the stable moment from which to start sampling for the
+    # distribution given the implicit `rv`, `size` and `param1` ... `paramN`
```
michaelosthege:
```suggestion
    # get_moment returns a symbolic expression for the stable moment from which to start sampling
    # the variable, given the implicit `rv`, `size` and `param1` ... `paramN`
```

docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -188,12 +196,15 @@ Some notes:
...
 1. The `logcdf` method is not a requirement, but it's a nice plus!
+1. Currently only one moment is supported in the `get_moment` method, and probably the "higher-order" one is the most useful (that is `mean` > `median` > `mode`)... You might need to truncate the moment if you are dealing with a discrete distribution.
+1. When creating the `get_moment` method, we have to be careful with `size != None` and broadcast properly when some parameters that are not used in the moment may nevertheless inform about the shape of the distribution. E.g. `pm.Normal.dist(mu=0, sigma=np.arange(1, 6))` returns a moment of `[mu, mu, mu, mu, mu]`.
 
 For a quick check that things are working you can try the following:
 
 ```python
 
 import pymc as pm
+from pm.distributions.distribution import get_moment
```
michaelosthege:
```suggestion
from pymc.distributions.distribution import get_moment
```

docs/source/contributing/developer_guide_implementing_distribution.md
michaelosthege: Your comments & code here are correct. I think Ricardo referred to the example _implementation_ above.

The test case further down also assumes that `Blah == Normal`, so maybe make this example also with a normal get_moment?

docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -350,8 +365,40 @@ def test_blah_logcdf(self):
 
 ```
 
+## 5. Adding tests for the `get_moment` method
+
+Tests for the `get_moment` method are contained in `pymc/tests/test_distributions_moments.py`, and make use of the function `assert_moment_is_expected`
+which checks if:
+1. Moments return `expected` values
```
michaelosthege:
```suggestion
1. Moments return the `expected` values
```
For less confusability with the statistical term _expected values_.

docs/source/contributing/developer_guide_implementing_distribution.md
amyoshino: Thanks for the clarification! Yes, I agree it makes sense to keep it uniform by changing this example from `Uniform` to `Normal`. I will commit the changes below.

amyoshino:
@michaelosthege Thank you very much for the comments and suggestions! I have accepted them all and added a commit that changes the examples we have discussed from `Uniform` to `Normal`. 
I hope the parameters I chose are ok. Let me know if something else is needed. 

Thank you,

ricardoV94 (approved):

amyoshino:
Thank you @ricardoV94 @michaelosthege for your reviews and guidance. It was fun working on this PR, the first one we never forget :)

PR #5245, opened 2021-12-07
Add model.logp_elemwiset

### LINKED ISSUE #5230: Perform single call to `aeppl.factorized_joint_logprob` when calling `model.logpt`
For `model.logpt` we are calling `model.datalogpt` and `model.varlogpt`, each of which calls `aeppl.factorized_joint_logprob`. It should be possible to do a single call to aeppl with all the free and observed model variables, which would save quite a lot of repeated operations on the aeppl side. For some reason this was not working when we tried it in #5166 and #4887.

https://github.com/pymc-devs/pymc/blob/9d4691c4ceaa18dd50081db32f38fd1310359d48/pymc/model.py#L734

### LINKED ISSUE #5236: inferencedata.log_likelihood is summing observations
When talking with @lucianopaz I realized we completely broke log_likelihood computation in V4.

```python
import pymc as pm
with pm.Model() as m:
    y = pm.Normal("y")
    x = pm.Normal("x", y, 1, observed=[5, 2])    
    idata = pm.sample(tune=5, draws=5, chains=2)
print(idata.log_likelihood['x'].values.shape)
# (2, 5, 1)
```

Whereas in V3:
```python
import pymc3 as pm
with pm.Model() as m:
    y = pm.Normal("y")
    x = pm.Normal("x", y, 1, observed=[5, 2])    
    idata = pm.sample(tune=5, draws=5, chains=2, return_inferencedata=True)
print(idata.log_likelihood['x'].values.shape)
# (2, 5, 2)
```

This happened because the default `model.logpt` now returns the `summed` logp by default whereas before it returned the vectorized logp by default. The change was done in https://github.com/pymc-devs/pymc/commit/0a172c87e39ee64bf5101a5887281ad6548e6ea4

Although that is a more sane default, we have to reintroduce an easy helper `logp_elemwiset` (I think this is pretty much broken right now as well) which calls `logpt` with `sum=False`. 

Also in this case we might want to just return the logprob terms as the dictionary items that are returned by `aeppl.factorized_joint_lopgrob` and let the end-user decide how he wants to combine them. These keys contain `{value variable: logp term}`. The default of calling `at.add` on all variables when `sum=False` is seldom useful (that's why we switched the default), due to potential unwanted broadcasting across variables with different dimensions.

One extra advantage of returning the dictionary items is that we don't need to create nearly duplicated graphs for each observed variable when computing the log-likelihood here:

https://github.com/pymc-devs/pymc/blob/fe2d101bb27e05b889eafda7e54b07e05250faee/pymc/backends/arviz.py#L268

We can request it for any number of observed variables at the same time, and then simply compile a function that has each variable logp term as an output, but otherwise shares the common nodes, saving on compilation, computation and memory footprint, when a model has more than one observed variable. 

For instance, this nested loop would no longer be needed:

https://github.com/pymc-devs/pymc/blob/fe2d101bb27e05b889eafda7e54b07e05250faee/pymc/backends/arviz.py#L276-L282

CC @OriolAbril

ricardoV94:
Closes #5236 
Closes #5230

### DIFF
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -24,7 +24,6 @@ from arviz.data.base import CoordSpec, DimSpec, dict_to_dataset, requires
 import pymc
 
 from pymc.aesaraf import extract_obs_data
-from pymc.distributions import logpt
 from pymc.model import modelcontext
 from pymc.util import get_default_varnames
 
@@ -264,11 +263,15 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
         if self.model is None:
             return None
 
+        # TODO: We no longer need one function per observed variable
         if self.log_likelihood is True:
-            cached = [(var, self.model.fn(logpt(var))) for var in self.model.observed_RVs]
+            cached = [
+                (var, self.model.fn(self.model.logp_elemwiset(var)[0]))
+                for var in self.model.observed_RVs
+            ]
         else:
             cached = [
-                (var, self.model.fn(logpt(var)))
+                (var, self.model.fn(self.model.logp_elemwiset(var)[0]))
                 for var in self.model.observed_RVs
                 if var.name in self.log_likelihood
             ]
--- a/pymc/distributions/logprob.py
+++ b/pymc/distributions/logprob.py
@@ -24,10 +24,8 @@ from aeppl import factorized_joint_logprob
 from aeppl.logprob import logcdf as logcdf_aeppl
 from aeppl.logprob import logprob as logp_aeppl
 from aeppl.transforms import TransformValuesOpt
-from aesara import config
 from aesara.graph.basic import graph_inputs, io_toposort
-from aesara.graph.op import Op, compute_test_value
-from aesara.tensor.random.op import RandomVariable
+from aesara.graph.op import Op
 from aesara.tensor.subtensor import (
     AdvancedIncSubtensor,
     AdvancedIncSubtensor1,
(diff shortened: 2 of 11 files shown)


pymc/distributions/logprob.py
```diff
@@ -156,45 +154,36 @@ def logpt(
...
         for _var in var:
-            if isinstance(_var.owner.op, RandomVariable):
-                rv_value_var = getattr(
-                    _var.tag, "observations", getattr(_var.tag, "value_var", _var)
-                )
-                rv_values = {_var: rv_value_var}
-    elif not isinstance(rv_values, Mapping):
-        # Else if we're given a single value and a single variable we assume a mapping among them.
-        rv_values = (
-            {var[0]: at.as_tensor_variable(rv_values).astype(var[0].type)} if len(var) == 1 else {}
```
ricardoV94, line 184: Size was always 1, because we automatically wrapped it in a list above

pymc/distributions/logprob.py
```diff
@@ -224,40 +213,23 @@ def logpt(
...
 
     # aeppl returns the logpt for every single value term we provided to it. This includes
-    # the extra values we plugged in above so we need to filter those out.
+    # the extra values we plugged in above so we filter those we actually wanted in the
+    # same order they were given in.
     logp_var_dict = {}
-    for value_var, _logp in temp_logp_var_dict.items():
-        if value_var in rv_values.values():
-            logp_var_dict[value_var] = _logp
+    for value_var in rv_values.values():
```
ricardoV94, line 233: Iterating over the expected value_vars will make sure we are not returning less terms than expected.

pymc/distributions/logprob.py
```diff
@@ -224,40 +213,23 @@ def logpt(
...
-            else:
-                logp_var = at.add(*logp_var_dict.values())
+    if scaling:
+        for value_var in logp_var_dict.keys():
+            if value_var in rv_scalings:
+                logp_var_dict[value_var] *= rv_scalings[value_var]
 
-        # Recompute test values for the changes introduced by the replacements
-        # above.
-        if config.compute_test_value != "off":
```
ricardoV94, line 258: Aeppl does this in factorized_joint_logprob

pymc/model.py
```diff
@@ -728,10 +727,74 @@ def logp_dlogp_function(self, grad_vars=None, tempered=False, **kwargs):
...
+        -------
+        Elemwise logp terms for ecah requested variable, in the same order of input.
+        """
+        if vars is None:
+            vars = self.free_RVs + self.observed_RVs + self.potentials
+        elif not isinstance(vars, (list, tuple)):
+            vars = [vars]
+
+        # We need to separate random variables from potential terms, and remember their
+        # original so that we can merge them together in the right order at the end
```
```suggestion
        # original order so that we can merge them together in the right order at the end
```

pymc/backends/arviz.py
```diff
@@ -264,11 +263,18 @@ def _extract_log_likelihood(self, trace):
         if self.model is None:
             return None
 
+        # TODO: The missing portion of partially observed variables no longer appears in the
+        #  logp graph of model.observed_RVs, nor in varg.tag.observations, altough this is
+        #  expected by `self.log_likelihood_vals_point`
+        # TODO: We no longer need one function per observed variable
```
OriolAbril, line 266: should we open issues to keep these todos in mind?

pymc/tests/test_idata_conversion.py
```diff
@@ -637,6 +636,14 @@ def test_issue_5043_autoconvert_coord_values(self):
             )
             assert isinstance(converter.coords["city"], pd.MultiIndex)
 
+    def test_elemwise_log_likelihood(self):
```
OriolAbril: I would include the check on log_likelihood shape into _multiple_ other tests so we check multiple scenarios like multiple observed variables, multivariate observed variables...

`test_to_idata` for example uses the `check_multiple_attrs` helper so it checks for presence or non-presence of keys as groups and of values as variables. But it then has an extra test on posterior predictive to make sure the reshaping is working correctly. We could add the test for log likelihood shape here to cover the same scenario as this new test.

Then it would probably be good to add checks in `test_missing_data_model`, `test_mv_missing_data_model`, `test_multiple_observed_rv` and `test_multivariate_observations`.

pymc/backends/arviz.py
ricardoV94: Definitely, something like #5247 and something for the missing obs

pymc/tests/test_idata_conversion.py
ricardoV94: Yeah I can do that, then we don't even need this test

pymc/backends/arviz.py
OriolAbril: missed that, thanks! I might add a note that it is beginner friendly provided you know Aesara?

ricardoV94: For the single function you don't need to know Aesara. The only reason I didn't do it in this PR is because I am not sure how to feed the multiple evaluated values to InferenceData xD

OriolAbril: then it's easy provided you know ArviZ it seems :sweat_smile:. 

If this generates a single function but that function doesn't vectorize over chain and draw dimensions, it would require retructuring all the log likelihood extraction. It is currently structured as 3 nested loops: variable, chain, draw, with the loop over variable being the outermost one. Not sure if this helps.

ricardoV94: Yeah that's exactly it. Now you can get rid of the outermost loop per variable. And I agree it's mostly a InferenceData question.

It's not a very pressing issue anyway. Most models only have one observation so it won't matter for those.

twiecki: Can we then link to the GH issues here?

ricardoV94: Added reminder issues #5255 and #5256

pymc/tests/test_idata_conversion.py
ricardoV94: I added log_likelihood shape checks to pre-existing tests

OriolAbril (approved):
I only checked the arviz and test parts which look good, I can't really review the other files

twiecki (approved):

PR #5249, opened 2021-12-08
Remove windows xfails

### LINKED ISSUE #4904: Scalar ODEs on Windows failing because of a casting & broadcastability issue that was fixed only for Linux

ricardoV94:
Tests were passing locally on my machine, just a sanity check that they still fail here

It seems they are passing. Somehow the strict XFAIL was not triggered?

Closes #4904

### DIFF
--- a/pymc/tests/test_ode.py
+++ b/pymc/tests/test_ode.py
@@ -173,7 +173,6 @@ class TestSensitivityInitialCondition:
         np.testing.assert_array_equal(np.ravel(model5_sens_ic), model5._sens_ic)
 
 
-@pytest.mark.xfail(condition=IS_WINDOWS, reason="https://github.com/pymc-devs/aesara/issues/390")
 def test_logp_scalar_ode():
     """Test the computation of the log probability for these models"""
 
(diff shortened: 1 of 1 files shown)


michaelosthege (approved):
In pytest they have spaces around the = sign: https://github.com/pytest-dev/pytest/blob/main/pyproject.toml#L22

ricardoV94:
Surprisingly enough the strict was raising when I tested locally one at a time

michaelosthege:
I saw a discussion in the pytest issues about changes to the meaning/scope of the --strict option. There might be a change in behavior that crept in with a newer pytest version a d becomes effective because we're running specific test files?

PR #5262, opened 2021-12-14
fix link to v3

mjhajharia:
we're using the v3 tag instead of stable for readthedocs (stable tag links to a documentation with missing images etc)


docs/source/404.md
```diff
@@ -32,7 +32,7 @@ modified the base url in order to support multi version docs
 :::{grid-item}
 :::
 :::{grid-item}
-```{button-link} https://docs.pymc.io/en/stable/
+```{button-link} https://docs.pymc.io/en/v3/
```
OriolAbril:
```suggestion
```{button-link} https://docs.pymc.io/en/{{ version_slug }}/
```

OriolAbril (changes requested):
I would not do that. I agree it makes sense to change something, but this only solves the broken images issue and for a short time (once a new 3.11 bugfix release is out or 4.0 is released we'll want to go back to stable and latest urls only). 

I would do two things:
* Use jinja substitutions via myst to have this link point to the same version the 404 originates from. See https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#substitutions-with-jinja2 for the changes needed to conf.py to activate the feature and add the `rtd_version` variable as value for the `version_slug` key in myst substitutions (or keep the key name also rtd_version
* add a 404 page to v3 pointing to v3 (there is no myst there so needs to be hardcoded). The page there can also be much uglier

OriolAbril (approved):

PR #5266, opened 2021-12-16
api docs template

OriolAbril:
I have modified the API docs for the distributions subsection so it can serve as
a template to update the rest of the modules (both present and missing, I haven't checked
if there are important functions missing from the api docs).

Key points:
* make sure that objects are documented at the path we want users to import them from.
  Users will do `pymc.Normal`, not `pymc.distributions.continuous.Normal` so the class
  should be documented as `pymc.Normal`
* Use autosummary with `:toctree:` option instead of automodule. This is more personal preference,
  but I think it looks better if each object has its own page. It is still easy to navigate between
  objects because they are all listed in the left sidebar.
* Make sure docs are working properly. The transform module is not rendering right,
  not really know why but sphinx doesn't like it and refuses to see `pymc.transforms` as a module,
  any ideas?


docs/source/api.rst
```diff
@@ -15,11 +15,34 @@ API Reference
...
+"exploratory analysis of Bayesian models".
+
+* Functions from the `arviz.plots` module are available through ``pymc.<function>`` or ``pymc.plots.<function>``,
+but for their API documentation please refer to the :ref:`ArviZ documentation <arviz:plot_api>`.
+
+* Functions from the `arviz.stats` module are available through ``pymc.<function>`` or ``pymc.stats.<function>``,
+but for their API documentation please refer to the :ref:`ArviZ documentation <arviz:stats_api>`.
+
+ArviZ is a dependency of PyMC and so, in addition to the locations described above,
+importing arviz and using ``arviz.<function>`` will also work without any extra installation.
```
michaelosthege:
```suggestion
importing ArviZ and using ``arviz.<function>`` will also work without any extra installation.
```

docs/source/api.rst
```diff
@@ -15,11 +15,34 @@ API Reference
...
+
+Generalized Linear Models (GLMs)
+--------------------------------
+
+Generalized Linear Models are delegated to the
+`Bambi <https://bambinos.github.io/bambi>`_.
+library, a high-level Bayesian model-building
+interface built on top of the PyMC.
```
michaelosthege:
```suggestion
interface built on top of PyMC.
```

michaelosthege (approved):

PR #5268, opened 2021-12-16
update posterior predictive sampling and notebook

### LINKED ISSUE #4647: Set keep_size=True in v4
By default, the output of `sample_posterior_predictive` and `fast_sample_posterior_predictive` returns an array with the chain and draw dimensions flattened. This makes conversion to `InferenceData` more complicated. I propose changing the default to `keep_size=True`, simplify the code in `to_inferencedata` and do not support conversion to inferencedata without `keep_dims`. Users would still be able to get the flattened samples, but to use "on their own". cc @rpgoldman

### LINKED ISSUE #4863: Return prior & posterior predictive samples as xarray object
As of now, sampling returns an [xarray](http://xarray.pydata.org/en/stable/) object, but when doing posterior predictive checks the posterior predictive samples are _not_ returned as xarray objects. Perhaps this could be changed for increased consistency?

OriolAbril:
- some updates to posterior predictive sampling and conversion
- update posterior predictive notebook

notebook was executed on this branch, not on the released beta though


twiecki (approved):

fonnesbeck:
What is the `(posterior_predictive)=` header for? Is it still needed?

OriolAbril:
> What is the (posterior_predictive)= header for? Is it still needed?

This is needed to add a target pointing to the title so we can then reference the notebook with sphinx cross-references. 
It is always better to use them instead of using urls because then links are more robust and will work even if we move the
notebook to a different location for example, however, now that we have versioned docs we can't use urls at all anywhere. Otherwise you are browsing 4.1 docs, click on a link and without warning start browsing 3.11 docs or 4.3

pymc/sampling.py
```diff
@@ -1557,21 +1559,25 @@ def sample_posterior_predictive(
...
-        sample per posterior sample, that is, the number of draws times the number of chains. It
-        is not recommended to modify this value; when modified, some chains may not be represented
-        in the posterior predictive sample.
+        sample per posterior sample, that is, the number of draws times the number of chains.
+
+        It is not recommended to modify this value; when modified, some chains may not be
+        represented in the posterior predictive sample. Instead, in cases when generating
+        posterior predictive samples is too expensive to do it once per posterior sample,
+        the recommended approach is to thin the ``trace`` argument
+        before passing it to ``sample_posterior_predictive``. In such cases it
```
michaelosthege, line 1568: A short example would be great. I know it's a one-liner, but I can't remember how...

pymc/sampling.py
```diff
@@ -1671,6 +1692,10 @@ def sample_posterior_predictive(
     vars_to_sample = list(get_default_varnames(vars_, include_transformed=False))
 
     if not vars_to_sample:
+        if return_inferencedata and not extend:
+            return None
+        elif return_inferencedata and extend:
+            return trace
```
michaelosthege:
```suggestion
        if return_inferencedata and not extend_inferencedata:
            return None
        elif return_inferencedata and extend_inferencedata:
            return trace
```

michaelosthege (changes requested):

michaelosthege (approved):

pymc/backends/arviz.py
```diff
@@ -396,34 +376,35 @@ def log_likelihood_to_xarray(self):
...
             else:
                 data[k] = np.expand_dims(ary, 0)
-                # pylint: disable=line-too-long
-                _log.warning(
-                    "posterior predictive variable %s's shape not compatible with number of chains and draws. "
-                    "This can mean that some draws or even whole chains are not represented.",
-                    k,
-                )
+                warning_vars.append(k)
+        warnings.warn(
```
thomasaarholt, line 389: @OriolAbril I think you intended for this warning to only be printed if `warnings_vars` was not empty. Is this correct? I am currently always getting the output:
```python
/Users/thomas/repo/pymc/pymc/backends/arviz.py:389: UserWarning: The shape of variables  in posterior_predictive group is not compatible with number of chains and draws. The automatic dimension naming might not have worked. This can also mean that some draws or even whole chains are not represented.
```

thomasaarholt: To be extra clear: The warning prints a space in place of `{', '.join(warning_vars)}`.

michaelosthege: Yeah, sorry I also missed it in my review.
It needs to be indented into an `if warning_vars`.
Thanks for reporting this! Do you want to open a PR? Otherwise I'll do it in the morning

PR #5271, opened 2021-12-19
tighten test_sample_after_set_data bound

austereantelope:
Hi,

The test `test_sample_after_set_data` in `test_data_container.py` has an assertion bound (`np.testing.assert_allclose(new_y, pp_trace.posterior_predictive["obs"].mean(("chain", "draw")), atol=1e-1)`) that is too loose. This means potential bug in the code could still pass the original test.

To quantify this I conducted some experiments where I generated multiple mutations of the source code under test and ran each mutant and the original code 100 times to build a distribution of their outputs. I used KS-test to find mutants that produced a different distribution from the original and use those mutants as a proxy for bugs that could be introduced. In the graph below I show the distribution of both the original code and also the mutants with a different distribution.

<p align="center">

</p>

Here we see that the bound of `0.1` is too loose since the original distribution (in orange) is much less than `0.1`. Furthermore in this graph we can observe that there are many mutants (proxy for bugs) that are below the bound as well and that is undesirable since the test should aim to catch potential bugs in code. I quantify the "bug detection" of this assertion by varying the bound in a trade-off graph below.

<p align="center">

</p>

In this graph, I plot the mutant catch rate (ratio of mutant outputs that fail the test) and the original pass rate (ratio of original output that pass the test). The original bound of `0.1` (red dotted line) has a catch rate of 0.09.

To improve this test, I propose to tighten the bound to `0.015` (the blue dotted line). The new bound has a catch rate of 0.21 (+0.12 increase compare to original) while still has >99 % pass rate (test is not flaky, I ran the updated test 500 times and observed >99 % pass rate). I think this is a good balance between improving the bug-detection ability of the test while keep the flakiness of the test low.

Do you guys think this makes sense? Please let me know if this looks good or if you have any other suggestions or questions. 

My Environment:
```
python=3.7.11
numpy=1.21.4
```

my pymc Experiment SHA:
`19e67f371d4641fd2f9ad7c8a7887361bdaa53d6`

### DIFF
--- a/pymc/tests/test_data_container.py
+++ b/pymc/tests/test_data_container.py
@@ -122,7 +122,7 @@ class TestData(SeededTest):
 
         assert pp_trace.posterior_predictive["obs"].shape == (1, 1000, 3)
         np.testing.assert_allclose(
-            new_y, pp_trace.posterior_predictive["obs"].mean(("chain", "draw")), atol=1e-1
+            new_y, pp_trace.posterior_predictive["obs"].mean(("chain", "draw")), atol=0.015
         )
 
     def test_shared_data_as_index(self):


ricardoV94:
@austereantelope, thanks for your analysis.

After checking that test, I am not convinced your analysis is the relevant one. The point of that test is to check that `pm.set_data` is working correctly (that the model now contains the new data instead of the original), and that the posterior predictive changes following this. It is not to test the statistical precision of the posterior predictive samples.

The relevant counterfactual case is that `set_data` does not do anything, and for that the test tolerance is probably more than enough since the initial data has no overlap with the updated data.

Let me know if I missed something obvious.

Edit: This kind of analysis would make sense for a test that is designed to assess posterior predictive precision, of which we have some here using statistical tests https://github.com/pymc-devs/pymc/blob/c00971710b167186b599db90ea52d1f4605b9ef8/pymc/tests/test_sampling.py#L477

austereantelope:
@ricardoV94 Thank you for the prompt response. 

I see your point that this test is to check if `pm.set_data` is working correctly. However we should also expect that `set_data` doesn't _reduce_ the precision of the predicted samples. It could be possible that a bug in `set_data` lowers the output precision after inputting new data and by tightening the bound we can catch potential cases like that.

Let me know if this makes sense or not

ricardoV94:
> However we should also expect that `set_data` doesn't _reduce_ the precision of the predicted samples. It could be possible that a bug in `set_data` lowers the output precision after inputting new data and by tightening the bound we can catch potential cases like that.

I think you need a causal model of why something like that could happen. Otherwise there are many other tests you could apply the same logic and many other things that we are not testing. 

What if the `set_data` fails for higher value ranges but not small values, what if there is a bug that affects the relative tolerance and not the absolute one. What if calling set_data before the first sampling behaves differently, or if setting it twice to the same value works as if nothing was done. What if set_data renders all the sample values constant?

These are silly examples, but the point is that we need a causal model of how things are expected to fail. set_data is just a fancy wrapper around Aesara shared variables,  and the tests should mostly be focused on whether this wrapping is working as intended. Tests about possible bugs with shared variables would not even go into the PyMC library. Similarly we are not testing if the basic Numpy functions we use in our codebase work as expected. We just trust they do.

I'll merge this PR because there is no good reason not to and you already did a lot of work thinking about this. My point is to try to direct any future investigations of this kind into directions where they are likely to have a bigger impact.

ricardoV94 (approved):

austereantelope:
@ricardoV94 Thanks for the very detailed feedback and the merge! I agree with what you said about the main focus of the test. In our analysis we generate mutations on the source code (only on the lines covered by the test and source files in the project). We observed that certain mutations do in fact decrease the precision of this test. While the mutantions are simple changes in the source code (ex. changing > to < ), they act as a proxy for potential bugs in the source code and in a way they can be used a causal model for how things can fail in this test. Again I understand that the point of the test might not be to check the high precision.

Thanks again for all your comments and I will do some more experiements on the tests that you pointed out before that are designed to assess the predicition precision.

PR #5272, opened 2021-12-19
dependence plot better representation for discrete/categorical variables

aloctavodia:
Now 
pdp_discrete

Before
pdp_non_discrete


michaelosthege (approved):
LGTM

PR #5273, opened 2021-12-20
Get rid of FutureWarnings related to Model.initial_point

### LINKED ISSUE #5247: Compile single function in model.point_logps
After #5245 we can use the new method to request a function that includes the `logp` terms for each variable separately, instead of compiling and evaluating it for each RV which is very likely much slower.

ricardoV94:
Also closes #5247


michaelosthege (approved):
Thanks!

PR #5274, opened 2021-12-21
Refactor of Sequential Monte Carlo internals

ciguaran:
+ This is my first PR ever to Pymc.
+ There should not be any breaking changes. This PR modularizes a little bit SMC internals.
+ This is an initial refactor, that adds documentation and makes it easier for a new reader to find where the actual SMC sampling happens, vs what checks/stats are generated.
+ Maybe some reviewer can answer the question in the comment line 86 from test_smc.

### DIFF
--- a/pymc/smc/runners.py
+++ b/pymc/smc/runners.py
@@ -1,6 +1,9 @@
 import multiprocessing as mp
+
 from itertools import repeat
+
 import cloudpickle
+
 from fastprogress.fastprogress import progress_bar
 
 
@@ -44,8 +47,5 @@ def run_chains_sequential(chains, progressbar, to_run, params, random_seed, kern
     for chain in range(chains):
         pbar.offset = 100 * chain
         pbar.base_comment = f"Chain: {chain + 1}/{chains}"
-        results.append(
-            to_run(*params, random_seed[chain], chain, pbar, **kernel_kwargs)
-        )
+        results.append(to_run(*params, random_seed[chain], chain, pbar, **kernel_kwargs))
     return results
-
--- a/pymc/smc/runners.py
+++ b/pymc/smc/runners.py
@@ -29,17 +29,6 @@ def run_chains_parallel(chains, progressbar, to_run, params, random_seed, kernel
     return results
 
 
-def _starmap_with_kwargs(pool, fn, args_iter, kwargs_iter):
-    # Helper function to allow kwargs with Pool.starmap
-    # Copied from https://stackoverflow.com/a/53173433/13311693
-    args_for_starmap = zip(repeat(fn), args_iter, kwargs_iter)
-    return pool.starmap(_apply_args_and_kwargs, args_for_starmap)
-
-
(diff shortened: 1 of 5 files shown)


ciguaran:
@aloctavodia FYI

pymc/tests/test_smc.py
```diff
@@ -80,8 +83,9 @@ def test_sample(self):
         initial_rng_state = np.random.get_state()
         with self.SMC_test:
             mtrace = pm.sample_smc(draws=self.samples, return_inferencedata=False)
-        assert_random_state_equal(initial_rng_state, np.random.get_state())
-
+        assert_random_state_equal(
+            initial_rng_state, np.random.get_state()
+        )  # TODO: why this? maybe to verify that nothing was sampled?
```
ricardoV94, line 88: The PR where this test was added provides some context: https://github.com/pymc-devs/pymc/pull/5131

Since it was not obvious why it was there, a comment before the assert explaining the purpose of the check could be helpful.

pymc/smc/runners.py
```diff
@@ -0,0 +1,51 @@
+import multiprocessing as mp
```
ricardoV94, line 1: I am not sure this needs to be in its own separate file. The original file is pretty concise

ricardoV94:
@ciguaran Thanks for opening the PR. I left some comments above.

Since the purpose of this PR is just to clean things out a bit, it would be useful if the commit history is also cleaned up, so that we don't mix changes that have different purposes (like tests doc-strings vs internal refactoring). There is a guide here that might provide some help: https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes

pymc/smc/runners.py
ciguaran: Right I didn't extracted due to size, but because it feels that running parallel vs sequential is kind of independent of SMC, and that it could be reused somewhere else. LMK if its enough reason to be outside or I move it again in.

ciguaran:
> @ciguaran Thanks for opening the PR. I left some comments above.
> 
> Since the purpose of this PR is just to clean things out a bit, it would be useful if the commit history is also cleaned up, so that we don't mix changes that have different purposes (like tests doc-strings vs internal refactoring). There is a guide here that might provide some help: https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes

cool, do you prefer me to squash commits or create a new PR?

ricardoV94:
> cool, do you prefer me to squash commits or create a new PR?

You can try the first, it's a good practice. If it fails miserably you can always open a new PR

pymc/smc/runners.py
ricardoV94: If a similar need emerges on another sampler we can refactor it then. Doing it now seems like some premature refactoring. 

Also it wouldn't make sense to host it inside the SMC module

ciguaran:
Closing to fix commit messages easily, please look at https://github.com/pymc-devs/pymc/pull/5281 instead

PR #5279, opened 2021-12-22
Add softmax to math

### LINKED ISSUE #4226: Suggestion: Add tt.nnet.softmax to pm.math?
Since the invlogit is already available from pm.math, maybe it would make sense to add the softmax as well, since it is the generalization of the invlogit for more than 2 variables. 

Personally, it's one of the most common functions I use, and many times the sole reason I have to import theano.tensor.

What do you think?

ricardoV94:
This PR adds softmax and log_softmax wrappers to the new Aesara Ops which accept the axis argument. The wrappers are there just to suppress the deprecation warnings issue on the Aesara side.

Closes #4226

### DIFF
--- a/pymc/math.py
+++ b/pymc/math.py
@@ -211,6 +211,22 @@ def invlogit(x, eps=None):
     return at.sigmoid(x)
 
 
+def softmax(x, axis=None):
+    # Ignore vector case UserWarning issued by Aesara. This can be removed once Aesara
+    # drops that warning
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", UserWarning)
+        return at.nnet.softmax(x, axis=axis)
+
+
+def log_softmax(x, axis=None):
+    # Ignore vector case UserWarning issued by Aesara. This can be removed once Aesara
+    # drops that warning
+    with warnings.catch_warnings():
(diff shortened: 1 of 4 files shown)


twiecki (approved):

ricardoV94:
Some scary looking errors on windows, most likely triggered by the new Aesara version: https://github.com/pymc-devs/pymc/runs/4605536352?check_suite_focus=true#step:7:21

https://github.com/aesara-devs/aesara/pull/701 would be my first guess

michaelosthege (changes requested):
This PR is blocked by https://github.com/aesara-devs/aesara/issues/707 and will be delayed until  we can upgrade to an Aesara release that includes a fix.

pymc/math.py
```diff
@@ -211,6 +211,22 @@ def invlogit(x, eps=None):
     return at.sigmoid(x)
 
 
+def softmax(x, axis=None):
+    # Ignore vector case UserWarning issued by Aesara. This can be removed once Aesara
+    # drops that warning
+    with warnings.catch_warnings():
```
ricardoV94, line 217: I checked locally that this does not screw up UserWarnings elsewhere

michaelosthege (approved):

twiecki (approved):

PR #5286, opened 2021-12-24
Fix UserWarning always being printed

thomasaarholt:
I noticed that #5268 introduced a small bug that always printed a warning. Now fixed. (@michaelosthege)
I'd be happy to run linting, but the linting/style checks link below from the PR template takes me to a blank page?

### DIFF
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -386,12 +386,13 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
             else:
                 data[k] = np.expand_dims(ary, 0)
                 warning_vars.append(k)
-        warnings.warn(
-            f"The shape of variables {', '.join(warning_vars)} in {kind} group is not compatible "
-            "with number of chains and draws. The automatic dimension naming might not have worked. "
-            "This can also mean that some draws or even whole chains are not represented.",
-            UserWarning,
-        )
+        if warning_vars:
+            warnings.warn(
+                f"The shape of variables {', '.join(warning_vars)} in {kind} group is not compatible "
+                "with number of chains and draws. The automatic dimension naming might not have worked. "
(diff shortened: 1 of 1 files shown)


michaelosthege (approved):
Thanks!
We had the instructions on a wiki page. Can't find it right now either 🤔
Essentially:
```
pip install pre-commit

# To run it automatically on every commit
pre-commit install

# Or to run it manually 
pre-commit run --all
```

The pre-commit was satisfied with your changes.

ricardoV94:
> Thanks!
> We had the instructions on a wiki page. Can't find it right now either 🤔
> Essentially:
> ```
> pip install pre-commit
> 
> # To run it automatically on every commit
> pre-commit install
> 
> # Or to run it manually 
> pre-commit run --all
> ```
> 
> The pre-commit was satisfied with your changes. 

The link still has pymc3, the new wiki page uses pymc and directs to the docs. We have to update the new issue template

https://github.com/pymc-devs/pymc/wiki/Python-Code-Style

thomasaarholt:
...what is going on with coverage? 😅

michaelosthege:
> ...what is going on with coverage? 😅

Our count of finished jobs needed before the first coverage estimate is posted is not correct.
It is set to post a coverage update before all jobs are done. That's because if the count is too high it would never get posted. Not sure, but probably this could be reconfigured to post it exactly when all jobs are done.

The codecov comment should update itself as more jobs complete. There is still some noise in the final result though.

OriolAbril:
I created a placeholder page for the old name so links continue to work, but we also moved everything to the documentation and shpuld default to that unless there is a compelling reason to use wikis. I updated the wiki but it seems it was renamed right before these changes so the I updated a page where no link was pointing to

All those devdoc style pages should use links to latest (instead of stable or specific versions) and will then be automatically updated with every commit

PR #5294, opened 2021-12-30
Add or fix type hints related to backends

michaelosthege:
Also raise a `ValueError` in `_choose_chains` instead of returning an incompatible type.


canyon289 (approved):
This is great, makes things clear, especially where basetrace vs multitrace is needed

Issue #4441, opened 2021-01-25
Imputation does not work in combination with pm.Data


michaelosthege:
## Description
Even after https://github.com/pymc-devs/pymc3/pull/4439 imputations don't work in combination with `pm.Data`.
This is because `pm.Data` creates a `SharedVariable` that currently does not support a `np.ma.MaskedArray`.

Almost identical to the example from #4437:
```python
data = numpy.array([
    [1,2,3],
    [4,5,float("nan")],
    [7,8,9],
])
print(data)
with pymc3.Model():
    pymc3.Normal(
        "L",
        mu=pymc3.Normal("x", shape=data.shape),
        sd=10,
        observed=pm.Data("D", data),
        shape=data.shape
    )
    pymc3.sample()
```

**Please provide the full traceback.**
```python
SamplingError                             Traceback (most recent call last)
<ipython-input-32-e837139c32a4> in <module>
     13         shape=data.shape
     14     )
---> 15     pymc3.sample()

c:\users\osthege\repos\pymc3-dev\pymc3\sampling.py in sample(draws, step, init, n_init, start, trace, chain_idx, chains, cores, tune, progressbar, model, random_seed, discard_tuned_samples, compute_convergence_checks, callback, jitter_max_retries, return_inferencedata, idata_kwargs, mp_ctx, pickle_backend, **kwargs)
    425     model = modelcontext(model)
    426     if start is None:
--> 427         check_start_vals(model.test_point, model)
    428     else:
    429         if isinstance(start, dict):

c:\users\osthege\repos\pymc3-dev\pymc3\util.py in check_start_vals(start, model)
    236                 "Initial evaluation of model at starting point failed!\n"
    237                 "Starting values:\n{}\n\n"
--> 238                 "Initial evaluation results:\n{}".format(elem, str(initial_eval))
    239             )
    240 

SamplingError: Initial evaluation of model at starting point failed!
Starting values:
{'x': array([[0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.]])}

Initial evaluation results:
x   -8.27
L     NaN
Name: Log-probability of test_point, dtype: float64
```

## Versions and main components

* PyMC3 Version: master (with #4439)
* Theano Version: 1.1.0

ricardoV94:
Is this something that might be problematic if automatically inputed (setting new Data with more/less missing values than the initial data)?

michaelosthege:
The whole implementation for switching the data in an existing model is broken.
[...]

PR #5295, opened 2021-12-30
Option to create non-shared `pm.Data`

### LINKED ISSUE #5105: Tweak `pm.Data` to create or auto-replace by `TensorConstant`
### Background
`pm.Data` is often used to track data in the model, making it end up in the `InferenceData` and so on.

Right now this always creates shared variables, but in the vast majority of applications one doesn't actually use the shared variable features.

Generally there are three (?) flavors of data variables:
* constant (could be a `TensorConstant` instead)
* shared with constant shape
* shared with flexible shape (constant ndim)

AFAIK Aesara does not distinguish between shared variables that can, or cannot be resized.

### Motivation
But the difference between `TensorConstant` and `SharedVariable` is a thing, and sometimes also costly.
For example see #3818

### Potential solution
* Additional kwargs to `pm.Data` through which one can constrain the flexibility (creating `TensorConstant` instead).
* Automatically replacing all `SharedVariable`s in the graph with `TensorConstant` before starting MCMC.

* PyMC/PyMC3 Version: `main`

michaelosthege:
Until now `pm.Data` always created a `SharedVariable`, even though most uses of `pm.Data` are not because it's mutable, but because it registers the variable in the model.

With `pm.Data(..., mutable=False/True)`, this PR adds the option to create non-shared data variables (`TensorConstant`) that possibly improve performance (via optimizations available for `TensorConstant`) and can also be used in places where a `SharedVariable` is not supported.

I also refactored `pm.Data` from a class into a function. I'm not 100 % sure why it was a class in the first place, but if I understand it correctly our use of `__new__` actually violates its intended signature, since we're not returning an instance of the correct type(?).

+ [x] what are the (breaking) changes that this PR makes? The `pm.Data.get_coord` staticmethod was refactored to `pm.data.determine_coords`, but this is just internal API.
+ [x] important background, or details about the implementation. See above.
+ [x] are the changes—especially new features—covered by tests and docstrings? Yes.
+ [consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples) 👉 Do we want to add a deprecation warning when `mutable` is not specified? We could also make `mutable=None` by default and warn that the default will change in the future?

Closes #5105

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -113,7 +113,10 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
   - Added partial dependence plots and individual conditional expectation plots [5091](https://github.com/pymc-devs/pymc3/pull/5091).
   - Modify how particle weights are computed. This improves accuracy of the modeled function (see [5177](https://github.com/pymc-devs/pymc3/pull/5177)).
   - Improve sampling, increase default number of particles [5229](https://github.com/pymc-devs/pymc3/pull/5229).
-- `pm.Data` now passes additional kwargs to `aesara.shared`. [#5098](https://github.com/pymc-devs/pymc/pull/5098)
+- New features for `pm.Data` containers:
+  - With `pm.Data(..., mutable=True/False)`, or by using `pm.MutableData` vs. `pm.ConstantData` one can now create `TensorConstant` data variables. They can be more performant and compatible in situtations where a variable doesn't need to be changed via `pm.set_data()`. See [#5295](https://github.com/pymc-devs/pymc/pull/5295).
+  - New named dimensions can be introduced to the model via `pm.Data(..., dims=...)`. For mutable data variables (see above) the lengths of these dimensions are symbolic, so they can be re-sized via `pm.set_data()`.
+  - `pm.Data` now passes additional kwargs to `aesara.shared`/`at.as_tensor`. [#5098](https://github.com/pymc-devs/pymc/pull/5098).
 - ...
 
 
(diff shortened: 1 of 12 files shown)


michaelosthege:
@ricardoV94 these changes could help for #4441 - I didn't test it, but with a (float) `pm.ConstantData/TensorConstant` having NaN values one could get at least the dims/coords/InferenceData.constant_data functionality also for imputation.

ricardoV94:
> @ricardoV94 these changes could help for #4441 - I didn't test it, but with a (float) `pm.ConstantData/TensorConstant` having NaN values one could get at least the dims/coords/InferenceData.constant_data functionality also for imputation.

Yes it should be possible, just need to consider the TensorConstant case in there

ricardoV94 (approved):

twiecki:
In `pm.set_data()` can we check if it was unmutable and then give an informative error?

Also, should the default really be mutable=True?

michaelosthege:
> In `pm.set_data()` can we check if it was unmutable and then give an informative error?

I'll make the error message more instructive 👍

> Also, should the default really be mutable=True?

No it shouldn't, but we can't flip that switch at the same time as we're introducing it. That's why I'd first default to showing `pm.ConstantData`/`pm.MutableData`in model graphs (see Slack).
Next step would be a `FutureWarning`.

Should I add the `FutureWarning` already? (When `mutable is None`..)

twiecki:
Sure, that all sounds good.

twiecki (changes requested):
Then we still need the check in `pm.set_data()` if it was immutable.

michaelosthege:
> Then we still need the check in `pm.set_data()` if it was immutable.

It was there already, line 1132 in model.py

twiecki (approved):

twiecki:
:+1:

PR #5296, opened 2021-12-30
Add type hints for rng_fn return value

canyon289:
If I did this would it be be right?

Copying from https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/continuous.py#L602

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -3008,7 +3008,9 @@ class SkewNormalRV(RandomVariable):
 
     @classmethod
     def rng_fn(cls, rng, mu, sigma, alpha, size=None) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


michaelosthege:
Depending on the parametrization, most of these can also return `float`.

We could wrap the returns in `np.asarray(...)`?

canyon289:
Good point, should I proceed with wrapping all the return values then?

michaelosthege:
> Good point, should I proceed with wrapping all the return values then?

Yeah, let's see if it breaks anything.

On a side note it might be time to start running mypy. At least in a way such that we can see our score in the CI.

ricardoV94 (approved):

michaelosthege (approved):

PR #5307, opened 2022-01-03
Fix Poisson logcdf

ricardoV94:
This showed up in #5231, the evalution of a scalar logcdf was returning a 1D result, because of a silly comma :D

Didn't add tests, because we are not checking this for any other distribution...

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -635,12 +635,10 @@ class Poisson(Discrete):
         safe_mu = at.switch(at.lt(mu, 0), 0, mu)
         safe_value = at.switch(at.lt(value, 0), 0, value)
 
-        res = (
-            at.switch(
-                at.lt(value, 0),
-                -np.inf,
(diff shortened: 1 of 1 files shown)


pymc/distributions/discrete.py
```diff
@@ -635,12 +635,10 @@ def logcdf(value, mu):
         safe_mu = at.switch(at.lt(mu, 0), 0, mu)
         safe_value = at.switch(at.lt(value, 0), 0, value)
 
-        res = (
-            at.switch(
-                at.lt(value, 0),
-                -np.inf,
-                at.log(at.gammaincc(safe_value + 1, safe_mu)),
-            ),
+        res = at.switch(
```
canyon289, line 638: Do we need a test, or does it not need to be updated?

ricardoV94: I mentioned above, I didn't add a test because we are not doing any test like this for any of the other logcdf methods

AlexAndorra (approved):
Thanks @ricardoV94

PR #5314, opened 2022-01-05
execute notebooks when building docs

OriolAbril:
This will ensure that the core notebooks in this repo are correctly versioned along with
the rest of the documentation. As they are only a handful of small and fast notebooks
it doesn't take long, but we should keep in mind to not add long/slow notebooks here
and keep them on pymc-examples.

I think we should wait to merge this until we have merged the updated notebooks in #5369


OriolAbril:
This should fix itself once we unpin/upgrade aesara and/or pin numpy

OriolAbril:
Aesara doesn't seem to work on readthedocs, even with the numpy pin, now it is installed and imported but can't compile models: https://pymc--5314.org.readthedocs.build/en/5314/learn/examples/pymc_overview.html#model-specification. We'll need to use something like: https://github.com/dfm/rtds-action which I don't have time to set up right now, unless someone can try this nowish we should move it to beta3 milestone and release beta2

OriolAbril:
Auto execution working now: https://pymc--5314.org.readthedocs.build/en/5314/learn/examples/GLM_linear.html is now running on beta2 and bambi 0.7

twiecki (approved):

PR #5315, opened 2022-01-06
Add initial distribution parameter types

canyon289:
Per comment https://github.com/pymc-devs/pymc/pull/5298#discussion_r778537160

Let me know if I got it right @ricardoV94

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -85,7 +85,7 @@ from pymc.distributions.dist_math import (
     normal_lcdf,
     zvalue,
 )
-from pymc.distributions.distribution import Continuous
+from pymc.distributions.distribution import DIST_PARAMETER_TYPES, Continuous
 from pymc.distributions.shape_utils import rv_size_is_none
 from pymc.math import invlogit, logdiffexp, logit
 from pymc.util import UNSET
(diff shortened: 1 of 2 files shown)


pymc/distributions/continuous.py
```diff
@@ -594,7 +597,7 @@ class TruncatedNormalRV(RandomVariable):
     def rng_fn(
         cls,
         rng: np.random.RandomState,
-        mu: Union[np.ndarray, float],
+        mu: DIST_PARAMETER_TYPES,
```
ricardoV94: In the `rng_fn` the previous type hints were correct, it's in the `.dist` classmethod in the distribution below that the `TensorVariable` also shows up as a possibility

ricardoV94 (changes requested):
Just to not be mistakenly merged in the meantime

pymc/distributions/continuous.py
```diff
@@ -692,12 +692,12 @@ class TruncatedNormal(BoundedContinuous):
     @classmethod
     def dist(
         cls,
-        mu: Optional[Union[float, np.ndarray]] = None,
-        sigma: Optional[Union[float, np.ndarray]] = None,
-        tau: Optional[Union[float, np.ndarray]] = None,
-        sd: Optional[Union[float, np.ndarray]] = None,
-        lower: Optional[Union[float, np.ndarray]] = None,
-        upper: Optional[Union[float, np.ndarray]] = None,
+        mu: Optional[DIST_PARAMETER_TYPES] = None,
```
canyon289, line 695: Thanks for previous comment @ricardoV94. Hows it looking now?

One question, are dist parameter always optional? If so I can move the optional flag to the definition

ricardoV94: No they are not always optional. Most times they are required actually

canyon289: Hows it look now?

ricardoV94: Good!

canyon289: 🎉  can we merge? Ill create an issue ticket for data umbrella sprint too have others add to the other parameters.

Thanks Ricardo

ricardoV94 (approved):

Issue #5033, opened 2021-09-29
Remove value variable information from `RV.tag`


ricardoV94:
As we did with test/initial values we should remove the value variable transformation from the `.tag` if possible. Probably saving it in a `model.value_transforms` dictionary that can be used directly in `aeppl`.

ricardoV94:
Same probably applies for observations and scaling information that is added to the tag

ricardoV94:
And ultimately to `value` variables themselves

PR #5320, opened 2022-01-07
Refactor `Factor` properties

### LINKED ISSUE #5318: Deprecate LoosePointFunc and make FastPointFunc the default
### Discussed in https://github.com/pymc-devs/pymc/discussions/5237

<sup>Originally posted by **ricardoV94** December  3, 2021</sup>
It seems that the difference is that `LoosePointFunc` allows you to evaluate a  `logp`/`dlogp`/ what have you with args and kwargs, instead of just passing a dictionary, but at the cost of being slower because it has to wrap them in a dictionary (Point) before calling logp.

```python
import pymc as pm

with pm.Model() as m:
    x = pm.Normal('x')
    y = pm.Normal('y', 0, 1, observed=0)

# Uses `LoosePointFunc` under the hood
m.logp({'x': 0})
m.logp(x=0)
m.logp([('x', 0)])  # Couldn't think of any other way of using `args`

# Uses `FastPointFunc` under the hood
m.fastlogp({'x': 0})
m.fastlogp(x=0) # Fails
m.fastlogp([('x', 0)]) # Fails
```

https://github.com/pymc-devs/pymc/blob/f6f1a8e807a0719a07fdb17ed79af9adce065814/pymc/model.py#L1722-L1766</div>

Since both attributes do the costly job of compiling a new logp function, we could also deprecate the `fastlogp` name and rename it  to something like `compile_logp`, similar to `compute_initial_point`, which gives a better intuition that you should probably cache the output of this, and not call it in a loop for instance

ricardoV94:
This PR deprecates most of the Factor properties, and gives the following default methods:

* compile_logp(self, vars=None, jacobian=True, sum=True):
* compile_dlogp(self, vars=None, jacobian=True):
* compile_d2logp(self, vars=None, jacobian=True):
 
The prefix `compile` is there to make it obvious this is a costly operation, and that the results should be cached once and reused afterwards. Also this makes it easier to deprecate the old properties, which couldn't easily be changed into methods.

It also deprecates `LoosePointFunc` and renames `FastPointFunc` to `PointFunc` which is now the default. This closes #5318.

Finally this PR also renames `pm.logpt` to `pm.joint_logpt`, which is now the main bridge between `PyMC` and `Aeppl`. The reason for this is that `pm.logpt` has diverged quite a lot from what it was doing in `V3`, and because of that it has a lot of ugly internal logic that should be handled by the `Model` object instead (involving trying to find value variables, transforms and scaling in the tags of the variables). The rename will hopefully help break the dependency that's there for backwards compat and facilitate refactoring in the future, such as #5033.

***
~~It would be nice to find a nice way to transition from `Model.logpt` being a property -> method that can accept `vars`, `jacobian` and `sum` as an argument. Then we can remove the more verbose `logp_elemwiset`, which is temporarily playing this more flexible role, in service of all other methods and properties~~ Decided to break away immediately following discussion in #5333


pymc/distributions/logprob.py
```diff
@@ -118,7 +118,7 @@ def _get_scaling(total_size, shape, ndim):
 )
 
 
-def logpt(
+def joint_logpt(
```
michaelosthege, line 120: Can you do a PR with the same rename on `v3`?
The warning in `3.11.5` points users to the function you're renaming here..

pymc/model.py
```diff
@@ -279,57 +280,124 @@ class Factor:
...
+        return self.model.compile_fn(self.dlogpt2(vars=vars, jacobian=jacobian))
+
     @property
     def logp(self):
         """Compiled log probability density function"""
+        warnings.warn(
+            "`logp` property has been deprecated, use `compile_logp` instead",
+            FutureWarning,
+        )
         return self.model.fn(self.logpt)
```
michaelosthege: Maybe you did this where possible already, but can't more of the properties/methods that now have a warning follow their own instruction and switch to the other API ?

pymc/model.py
```diff
@@ -1461,7 +1561,7 @@ def __getitem__(self, key):
             except KeyError:
                 raise e
 
-    def makefn(self, outs, mode=None, *args, **kwargs):
+    def makefn(self, outs, mode=None, **kwargs):
```
```suggestion
    def makefn(self, outs: Sequence[Variable], mode=None, **kwargs) -> Callable([Sequence[np.ndarray]], Sequence[np.ndarray]):
```
Is it something like this?

pymc/sampling_jax.py
```diff
@@ -28,7 +28,6 @@
 
 from pymc import Model, modelcontext
 from pymc.backends.arviz import find_observations
-from pymc.distributions import logpt
```
michaelosthege, line 31: Was this line unused?

pymc/tuning/starting.py
```diff
@@ -113,18 +116,15 @@ def find_MAP(
...
         # This might be needed for calls to `dlogp_func`
         # start_map_info = tuple((v.name, v.shape, v.dtype) for v in vars)
-
-        def dlogp_func(x):
-            return DictToArrayBijection.mapf(model.fastdlogp_nojac(vars))(
-                RaveledVars(x, x0.point_map_info)
-            )
-
+        compiled_dlogp_func = DictToArrayBijection.mapf(model.compile_dlogp(rvs, jacobian=False))
+        dlogp_func = lambda x: compiled_dlogp_func(RaveledVars(x, x0.point_map_info))
```
michaelosthege, line 127: 👍

pymc/model.py
```diff
@@ -788,6 +856,54 @@ def logp_elemwiset(
...
+                else:
+                    raise ValueError(
+                        f"Requested variable {var} not found among the model variables"
+                    )
+
+        cost = self.logpt if not jacobian else self.logp_nojact
+        return hessian(cost, value_vars)
+
     @property
     def logpt(self):
```
```suggestion
    def logpt(self) -> Variable:
```
I like the refactor. Adding more type hints will make it easier for beginners like me who can't remember which property is a variable and which one is a function...

michaelosthege (commented):
Sorry about that nitpick, but given how much code calls these properties/methods, could you add some type hints?
Even incomplete type hints here can give a lot of mypy coverage, right?

pymc/distributions/logprob.py
ricardoV94: For joint_logpt? That one has no good equivalent in V3. Here we are a advising people to use a new V4 model method (logp_elemwise)

joint_logpt should really be only called internally

pymc/model.py
ricardoV94: I thought that would be more bug prone and difficult to review.

pymc/model.py
ricardoV94: I'll have to check. thanks for the suggestion

pymc/sampling_jax.py
ricardoV94: Let me check again

pymc/model.py
```diff
@@ -788,6 +856,54 @@ def logp_elemwiset(
...
+            for i, var in enumerate(vars):
+                value_var = self.rvs_to_values.get(var)
+                if value_var is not None:
+                    value_vars.append(value_var)
+                else:
+                    raise ValueError(
+                        f"Requested variable {var} not found among the model variables"
+                    )
+
+        cost = self.logpt if not jacobian else self.logp_nojact
```
```suggestion
        cost = self.logpt if jacobian else self.logp_nojact
```

pymc/model.py
```diff
@@ -279,57 +280,124 @@ class Factor:
...
+            factors = at.sum([at.sum(factor) for factor in factors])
+        return self.model.compile_fn(factors)
+
+    def compile_dlogp(self, vars=None, jacobian=True):
+        """Compiled log probability density gradient function"""
+        return self.model.compile_fn(self.dlogpt(vars=vars, jacobian=jacobian))
+
+    def compile_d2logp(self, vars=None, jacobian=True):
+        """Compiled log probability density hessian function"""
+        return self.model.compile_fn(self.dlogpt2(vars=vars, jacobian=jacobian))
```
```suggestion
        return self.model.compile_fn(self.d2logpt(vars=vars, jacobian=jacobian))
```

pymc/sampling_jax.py
ricardoV94: Seems like it was

ricardoV94:
Marking as a draft to address #5333 in this PR

pymc/model.py
```diff
@@ -279,57 +281,137 @@ class Factor:
...
+    def compile_d2logp(
+        self,
+        vars: Optional[Union[Variable, Sequence[Variable]]] = None,
+        jacobian: bool = True,
+    ):
+        """Compiled log probability density hessian function"""
+        return self.model.compile_fn(self.d2logpt(vars=vars, jacobian=jacobian))
+
     @property
     def logp(self):
```
twiecki: So yeah, I would just get rid of all of these.

ricardoV94:
Band-aids are now being removed quickly! :adhesive_bandage:

RELEASE-NOTES.md
```diff
@@ -39,9 +39,13 @@ All of the above apply to:
 - `pm.sample_prior_predictive`, `pm.sample_posterior_predictive` and `pm.sample_posterior_predictive_w` now return an `InferenceData` object by default, instead of a dictionary (see [#5073](https://github.com/pymc-devs/pymc/pull/5073)).
 - `pm.sample_prior_predictive` no longer returns transformed variable values by default. Pass them by name in `var_names` if you want to obtain these draws (see [4769](https://github.com/pymc-devs/pymc/pull/4769)).
 - `pm.sample(trace=...)` no longer accepts `MultiTrace` or `len(.) > 0` traces ([see 5019#](https://github.com/pymc-devs/pymc/pull/5019)).
+- `logpt`, `logpt_sum`, `logp_elemwiset` and `nojac` variations were removed. Use `Model.logpt(jacobian=jacobian, sum=sum)` instead.
```
```suggestion
- `logpt`, `logpt_sum`, `logp_elemwiset` and `nojac` variations were removed. Use `Model.logpt(jacobian=True/False, sum=True/False)` instead.
```

pymc/model.py
```diff
@@ -1830,22 +1755,6 @@ def __call__(self, state):
         return self.f(**state)
 
 
-class LoosePointFunc:
```
twiecki, line 1833: it's so great that these confusing things finally can go.

RELEASE-NOTES.md
ricardoV94: Done

twiecki:
This is shaping up to be a really nice refactor of some quite obscure code.

ricardoV94:
> This is shaping up to be a really nice refactor of some quite obscure code.

Hopefully one will also be less lost when doing autocomplete on a Model to find a method xD

pymc/model_graph.py
```diff
@@ -91,6 +91,7 @@ def _filter_parents(self, var, parents) -> Set[VarName]:
 
     def get_parents(self, var: TensorVariable) -> Set[VarName]:
         """Get the named nodes that are direct inputs to the var"""
+        # TODO: Update these lines, variables no longer have a `logpt` attribute
```
michaelosthege, line 94: Is this broken now?

(Update: This code wasn't working before either.)

pymc/model.py
```diff
@@ -731,24 +645,53 @@ def logp_dlogp_function(self, grad_vars=None, tempered=False, **kwargs):
...
         return ValueGradFunction(costs, grad_vars, extra_vars_and_values, **kwargs)
 
-    def logp_elemwiset(
+    def compile_logp(
+        self,
+        vars: Optional[Union[Variable, Sequence[Variable]]] = None,
+        jacobian: bool = True,
+        sum: bool = True,
+    ):
+        """Compiled log probability density function"""
```
michaelosthege: This needs a more detailed docstring.

pymc/model.py
```diff
@@ -731,24 +645,53 @@ def logp_dlogp_function(self, grad_vars=None, tempered=False, **kwargs):
...
+    ):
+        """Compiled log probability density gradient function"""
+        return self.model.compile_fn(self.dlogpt(vars=vars, jacobian=jacobian))
+
+    def compile_d2logp(
+        self,
+        vars: Optional[Union[Variable, Sequence[Variable]]] = None,
+        jacobian: bool = True,
+    ):
+        """Compiled log probability density hessian function"""
```
michaelosthege: This one too

pymc/model.py
```diff
@@ -785,90 +728,101 @@ def logp_elemwiset(
...
 
-    @property
-    def logp_nojact(self):
-        """Aesara scalar of log-probability of the model but without the jacobian
-        if transformed Random Variable is presented.
+    def dlogpt(
+        self,
+        vars: Optional[Union[Variable, Sequence[Variable]]] = None,
+        jacobian: bool = True,
+    ) -> Variable:
```
michaelosthege, line 779: Needs a docstring

pymc/model.py
```diff
@@ -785,90 +728,101 @@ def logp_elemwiset(
...
+        cost = self.logpt(jacobian=jacobian)
+        return gradient(cost, value_vars)
 
-        if self.name:
-            logp_var.name = f"__logp_nojac_{self.name}"
+    def d2logpt(
+        self,
+        vars: Optional[Union[Variable, Sequence[Variable]]] = None,
+        jacobian: bool = True,
+    ) -> Variable:
```
michaelosthege, line 817: Also here; the name is not self-explanatory.

pymc/tests/test_shared.py
```diff
@@ -26,7 +27,10 @@ def test_deterministic(self):
             data_values = np.array([0.5, 0.4, 5, 2])
             X = aesara.shared(np.asarray(data_values, dtype=aesara.config.floatX), borrow=True)
             pm.Normal("y", 0, 1, observed=X)
-            model.logp(model.recompute_initial_point())
+            # TODO: This should assert something
```
michaelosthege:
```suggestion
```

RELEASE-NOTES.md
```diff
@@ -39,9 +39,13 @@ All of the above apply to:
 - `pm.sample_prior_predictive`, `pm.sample_posterior_predictive` and `pm.sample_posterior_predictive_w` now return an `InferenceData` object by default, instead of a dictionary (see [#5073](https://github.com/pymc-devs/pymc/pull/5073)).
 - `pm.sample_prior_predictive` no longer returns transformed variable values by default. Pass them by name in `var_names` if you want to obtain these draws (see [4769](https://github.com/pymc-devs/pymc/pull/4769)).
 - `pm.sample(trace=...)` no longer accepts `MultiTrace` or `len(.) > 0` traces ([see 5019#](https://github.com/pymc-devs/pymc/pull/5019)).
+- `logpt`, `logpt_sum`, `logp_elemwiset` and `nojac` variations were removed. Use `Model.logpt(jacobian=True/False, sum=True/False)` instead.
```
michaelosthege, line 42:
```suggestion
- `logpt`, `logpt_sum`, `logp_elemwiset` and `nojac` property variations were removed. Use `Model.logpt(jacobian=True/False, sum=True/False)` instead.
```

pymc/model.py
```diff
@@ -731,24 +645,53 @@ def logp_dlogp_function(self, grad_vars=None, tempered=False, **kwargs):
...
+    ):
+        """Compiled log probability density function"""
+        return self.model.compile_fn(self.logpt(vars=vars, jacobian=jacobian, sum=sum))
+
+    def compile_dlogp(
+        self,
+        vars: Optional[Union[Variable, Sequence[Variable]]] = None,
+        jacobian: bool = True,
+    ):
+        """Compiled log probability density gradient function"""
```
michaelosthege: And here

michaelosthege (changes requested):

michaelosthege:
I'm open to having docstring link to docstrings of the other properties/methods.
We can probably make the necessary changes all as suggestions and have one commit for them?

pymc/model_graph.py
ricardoV94: It was already

ricardoV94:
> I'm open to having docstring link to docstrings of the other properties/methods. We can probably make the necessary changes all as suggestions and have one commit for them?

Yeah I think the `compile_*` is better of just directing to the respective methods for the meaning of the variables. Feel free to make changes and force-push to this branch! Won't have time to work on this until next week

michaelosthege (approved):

PR #5325, opened 2022-01-08
Re-enable passing `dims` alongside `shape` or `size`

michaelosthege:
This PR makes the following possible (which was supported in `v3` already).

```python
with pm.Model():
    pm.Normal("n", size=3, dims="town")
```

### Changes in this PR
+ Dropping the `ndim_resize` item from the tuple returned by `resize_from_shape` and `resize_from_dims`, because it was simply the length of the second item and not used anywhere.
+ Removing the `raise ValueError` blocks that denied passing `dims` alongside `shape` or `size`.
+ Refactoring/addition of tests to validate that passing `shape+dims` or `size+dims` also works when combined with `Ellipsis`.

+ [x] what are the (breaking) changes that this PR makes? → it undoes a breaking change
+ [x] ~[consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples)~
+ [x] ~right before it's ready to merge, mention the PR in the RELEASE-NOTES.md~ not needed, because this feature worked in v3 and users most probably assume already that passing dims in addition or size or shape is valid.

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -213,16 +213,6 @@ class Distribution(metaclass=DistributionMeta):
         if rng is None:
             rng = model.next_rng()
 
-        if dims is not None and "shape" in kwargs:
-            raise ValueError(
-                f"Passing both `dims` ({dims}) and `shape` ({kwargs['shape']}) is not supported!"
-            )
-        if dims is not None and "size" in kwargs:
-            raise ValueError(
-                f"Passing both `dims` ({dims}) and `size` ({kwargs['size']}) is not supported!"
-            )
-        dims = convert_dims(dims)
-
         # Create the RV without dims information, because that's not something tracked at the Aesara level.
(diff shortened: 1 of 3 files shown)


ricardoV94 (approved):
Question: what happens if dims and observed mismatch?

michaelosthege:
> Question: what happens if dims and observed mismatch?

The same as when dims and shape, or dims and size or dims and the actual shape mismatch: You get shape errors in MCMC or upon InferenceData conversion.

It would be great to make this more robust for the user.
For example by validating dim lengths in variable registration, or right before sampling.
I'm constantly running into these kinds of shape problems myself, but I'm also afraid of the performance hit from doing all those `.eval()`s.

We could also consider to change the signature of `Model.dim_lengths: Dict[str, Variable]` to `Model.dim_lengths: Dict[str, Union[Variable, int]]` and store ints for dims that don't change. (e.g. when they were registered from a `TensorConstant`.)
That would make it a lot faster to perform the aforementioned checks.

PR #5347, opened 2022-01-13
Add helper wrapper aound Interval transform

ricardoV94:

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -38,6 +38,7 @@ All of the above apply to:
 - `pm.sample_prior_predictive`, `pm.sample_posterior_predictive` and `pm.sample_posterior_predictive_w` now return an `InferenceData` object by default, instead of a dictionary (see [#5073](https://github.com/pymc-devs/pymc/pull/5073)).
 - `pm.sample_prior_predictive` no longer returns transformed variable values by default. Pass them by name in `var_names` if you want to obtain these draws (see [4769](https://github.com/pymc-devs/pymc/pull/4769)).
 - `pm.sample(trace=...)` no longer accepts `MultiTrace` or `len(.) > 0` traces ([see 5019#](https://github.com/pymc-devs/pymc/pull/5019)).
+- `transforms` module is no longer accessible ta the root level. It is accessible at `pymc.distributions.transforms` (see[#5347](https://github.com/pymc-devs/pymc/pull/5347)).
 - `logpt`, `logpt_sum`, `logp_elemwiset` and `nojac` variations were removed. Use `Model.logpt(jacobian=True/False, sum=True/False)` instead.
 - `dlogp_nojact` and `d2logp_nojact` were removed. Use `Model.dlogpt` and `d2logpt` with `jacobian=False` instead.
 - `logp`, `dlogp`, and `d2logp` and `nojac` variations were removed. Use `Model.compile_logp`, `compile_dlgop` and `compile_d2logp` with `jacobian` keyword instead.
--- a/docs/source/api/distributions/transforms.rst
+++ b/docs/source/api/distributions/transforms.rst
@@ -15,12 +15,25 @@ Transform instances are the entities that should be used in the
 
     simplex
     logodds
-    interval
     log_exp_m1
     ordered
     log
     sum_to_1
     circular
+
+
+Specific Transform Classes
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. autosummary::
+   :toctree: generated
+
+    CholeskyCovPacked
+    Interval
+    LogExpM1
+    Ordered
+    SumTo1
+
 
 Transform Composition Classes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -29,14 +42,3 @@ Transform Composition Classes
    :toctree: generated
 
     Chain
-    CholeskyCovPacked
-
-Specific Transform Classes
-~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-.. autosummary::
-   :toctree: generated
-
-    LogExpM1
-    Ordered
-    SumTo1
--- a/docs/source/api/distributions/transforms.rst
+++ b/docs/source/api/distributions/transforms.rst
@@ -2,7 +2,7 @@
 Transformations
 ***************
 
-.. currentmodule:: pymc.transforms
+.. currentmodule:: pymc.distributions.transforms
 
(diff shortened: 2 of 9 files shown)


ricardoV94:
The doc trick does not seem to work: https://pymc--5347.org.readthedocs.build/en/5347/api/distributions/transforms.html

ricardoV94:
CC @OriolAbril guru :genie:

OriolAbril:
There are two issues here. The main one is that we are trying to use `pymc.transforms`. This is what happens on my computer:

```python
In [1]: import pymc.transforms as transforms
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-dd33c3c1e58a> in <module>
----> 1 import pymc.transforms as transforms

ModuleNotFoundError: No module named 'pymc.transforms'

In [2]: from pymc import transforms

In [3]: transforms.Chain
Out[3]: pymc.distributions.transforms.Chain

In [4]: import pymc.transforms
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-4-2bd367763092> in <module>
----> 1 import pymc.transforms

ModuleNotFoundError: No module named 'pymc.transforms'
```

I have no idea how this happens nor how to fix it.

sphinx uses `import pymc.transforms` to then get the docstrings form the onbjects listed in https://github.com/pymc-devs/pymc/blob/main/docs/source/api/distributions/transforms.rst. As it fails, only the index table at https://pymc--5347.org.readthedocs.build/en/5347/api/distributions/transforms.html is generated, but unlike what happens with continuous distributions for example, you can't click on them to get to the specific page with the whole docstring.

The second much less important issue is that even if that were fixed, docstrings have some constraints because we use numpydoc to parse them, so directives right in the extended description might not work as expected. I'd suggest using sections
defined by numpydoc instead of raw directives, iirc, the warning one is parsed and formatted as an admonition, but the notes are not. Either way, numpydoc also defines the order in which every section appears in the rendered docs independently of the order in the docstring I belive, so if the notes end up at the bottom and we want the tip to appear at the top we can look into it.

OriolAbril:
forgot some links: https://numpydoc.readthedocs.io/en/latest/format.html#warnings and https://numpydoc.readthedocs.io/en/latest/format.html#notes

ricardoV94:
@OriolAbril I fixed the module path and the elements are now clickable. It still seems like sphinx ignores the custom `__doc__`: 

https://pymc--5347.org.readthedocs.build/en/5347/api/distributions/generated/pymc.distributions.transforms.Interval.html#pymc.distributions.transforms.Interval

Would it be better to add specific documentation directly in the `rst` (shivers) file?

OriolAbril:
It is still being imported in init https://github.com/ricardoV94/pymc3/blob/transform_issue/pymc/__init__.py#L49 to expose pymc.transforms to users, and even though it is now documented as pymc.distributions.transforms, the examples (and cross references) use pymc.transforms instead. I find this very confusing. Imo, If we want users to use pymc.transforms we should document there and try and fix that import issue. The transforms file seems quite standalone, maybe it could be moved outside the distributions folder? (even then we can document it inside distributions in the api,  those groupings are conceptual now, most things are imported and documented as pymc.object) 

The ignoring of the __doc__ is very confusing. The rest are working,  my guess is that somehow doing this trick on classes instead of functions is not possible or has to be done in some other way. And the worse thing is it has a docstring somehow. Where does the `alias of aeppl.transforms.intervaltransform` come from? It even uses the correct syntax for sphinx cross references?!?!?!!!! 

> Would it be better to add specific documentation directly in the rst (shivers) file?

We can try that yeah, but then the docstring won't be available from ides which is far from ideal.

OriolAbril:
had an idea that might fix this? (coming from someone with little understanding of imports)

We now do 

https://github.com/pymc-devs/pymc/blob/eb5177a5bb97a437870727906fc52acf29dc53d2/pymc/__init__.py#L54

in the "global" init, but we also do 

https://github.com/pymc-devs/pymc/blob/eb5177a5bb97a437870727906fc52acf29dc53d2/pymc/__init__.py#L53

and all distributions are available in the pymc namespace. So maybe doing the `from pymc.distributions import transforms` in the `distributions/__init__.py` and adding `transforms` to the `__all__` variable will work if we then keep only the `from distributions import *` in the global init

(assuming we do want to keep the transforms inside the distributions folder)

ricardoV94:
@OriolAbril sounds like it could work

ricardoV94:
@OriolAbril does my last commit do what you were suggesting?

ricardoV94:
Docs not good yet: https://pymc--5347.org.readthedocs.build/en/5347/api/distributions/generated/pymc.transforms.Interval.html

OriolAbril:
That was my idea but transforms continues to not be a proper module:

```
In [1]: import pymc.transforms as transforms
ModuleNotFoundError                       Traceback (most recent call last)
Input In [1], in <module>
----> 1 import pymc.transforms as transforms

ModuleNotFoundError: No module named 'pymc.transforms'

In [2]: from pymc import transforms

In [3]:
```

OriolAbril:
I tried updating the transforms.rst file to change `currentmodule:: pymc.transforms` by `pymc` only, then list elements as `transforms.simplex`... instead of only `simplex`. This partially tricks sphinx, it generates pages for each object, but is still unable to access their `__doc__` attribute and the page ends up empty and not properly linked:

imatge

I think we can merge for now and at some point fix the import issue so that both `import` and `from ... import` work. If that is fixed, the docs will render correctly without any change to the docs themselves.

ricardoV94:
@OriolAbril would it help if `transforms` was a proper module itself? With a directory and `__init__.py`?

OriolAbril:
I am quite sure it will. Sphinx works as long as it is able to import the objects. If both imports from https://github.com/pymc-devs/pymc/pull/5347#issuecomment-1071034498 work then sphinx will work, I think making it a proper module will make the two import ways work, but here is where I am not completely sure.

ricardoV94:
I'll give it a try

OriolAbril (approved):
Looks good. My only comment is this probably deserves a mention in the release notes

OriolAbril:
Docs are working now: https://pymc--5347.org.readthedocs.build/en/5347/api/distributions/transforms.html. There are some rendering issues though.

ricardoV94:
@OriolAbril I think formatting is fixed now?

OriolAbril:
yep

ricardoV94:
Thanks a lot for the help @OriolAbril. I realized now I should have added you as a co-author :/

PR #5349, opened 2022-01-13
Disable sampling of transformed variables from prior_predictive 

### LINKED ISSUE #5076: Disable sampling of transformed variables from `prior_predictive`
This was added to facilitate sampling from the prior as a starting point for SMC, but with the new functions added in #4983 this is no longer needed.

https://github.com/pymc-devs/pymc/blob/6d2aa5ddebed01d81c2ab66b9d4bd02194f82508/pymc3/sampling.py#L1983-L1999

markvrma:
**What are the (breaking) changes that this PR makes?**

removed the ```if``` statement that appends ```rv_var.name``` to ```names``` and ```rv_var``` to ```vars_to_sample```

resolves #5076

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -2062,12 +2062,6 @@ def sample_prior_predictive(
         names.append(name)
         vars_to_sample.append(transformed_rv_var)
 
-        # If the user asked for the transformed variable in var_names, but not the
-        # original RV, we add it manually here
-        if rv_var.name not in names:
-            names.append(rv_var.name)
-            vars_to_sample.append(rv_var)
-
     inputs = [i for i in inputvars(vars_to_sample) if not isinstance(i, (Constant, SharedVariable))]
 
     sampler_fn = compile_pymc(


ricardoV94:
Hi @markvrma thanks for opening the PR! These lines should also be updated: https://github.com/pymc-devs/pymc/blob/a823043cc69e848fd37c42100505967f3c745d76/pymc/sampling.py#L2003-L2006

And more importantly, SMC is relying on prior predictive to get a bunch of transformed samples here: https://github.com/pymc-devs/pymc/blob/e6fc2ec100dafacca4f0b9eab40661174707afcd/pymc/smc/smc.py#L179-L184

That could will have to be refactored to make use of this function: https://github.com/pymc-devs/pymc/blob/d84c71884762fb925eb7272b9d28e879399e326a/pymc/initial_point.py#L130 and calling it a bunch of times

pymc/sampling.py
```diff
@@ -2062,12 +2062,6 @@ def sample_prior_predictive(
         names.append(name)
         vars_to_sample.append(transformed_rv_var)
 
-        # If the user asked for the transformed variable in var_names, but not the
```
ricardoV94, line 2065: The lines above are also related to this logic

markvrma:
Hi! would it be right declare a copy of `vars_to_sample` before  `missing_names`, and then use the copy in `inputvars()` and use `vars_to_sample` in `compile_pymc`

``` py  
   
    # Making a copy of vars_to_sample
    vars_to_sample_without_tranformed_var = vars_to_sample
    
    # Any variables from var_names that are missing must be transformed variables.
    # Misspelled variables would have raised a KeyError above.
    missing_names = vars_.difference(names)
    for name in missing_names:
        transformed_value_var = model[name]
        rv_var = model.values_to_rvs[transformed_value_var]
        transform = transformed_value_var.tag.transform
        transformed_rv_var = transform.forward(rv_var, *rv_var.owner.inputs)

        names.append(name)
        vars_to_sample.append(transformed_rv_var)

        # If the user asked for the transformed variable in var_names, but not the
        # original RV, we add it manually here
        if rv_var.name not in names:
            names.append(rv_var.name)
            vars_to_sample.append(rv_var)
    
    inputs = [i for i in inputvars(vars_to_sample_without_tranformed_var) if not isinstance(i, (Constant, SharedVariable))]

    sampler_fn = compile_pymc(
        inputs, vars_to_sample, allow_input_downcast=True, accept_inplace=True, mode=mode
    )
```

ricardoV94:
Closing this PR due to inactivity. Feel free to reopen a new one anytime!

PR #5348, opened 2022-01-13
Bump for 4.0.0b2

michaelosthege:
I fixed two links in the readme that were broken.

⚠ The links in the readme point to a mix between v3 and v4 documentation, but for the ones I changed I didn't find good v3 targets.

PR #5351, opened 2022-01-13
Fix link & re-run notebooks

michaelosthege:
Follow-up to #5348


OriolAbril (approved):
Thanks :heart: everything looks great

PR #5363, opened 2022-01-17
Add additional compiler flag to work around GCC bug.

jedludlow:
Work around compiler bug in GCC < 8.4 related to structured exception
handling registers on Windows.
See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65782 for details.

See #4937 for additional discussion.


twiecki (approved):

twiecki:
Thanks @jedludlow!

PR #5364, opened 2022-01-17
Test conditional workflow runs

OriolAbril:
Also use the opportunity for some cleanup of infrastructure related files


OriolAbril:
Tried using the info from: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore but doesn't seem to work

OriolAbril:
If someone can look into the actions thing that would be great, otherwise I'll remove those files from the PR and keep only the removals of unused files

michaelosthege (approved):

michaelosthege:
For the CI runs here, your changes did not have an effect. Maybe because it need to have been merged first?

OriolAbril:
Let's try if merging works, good idea

PR #5368, opened 2022-01-18
Add polyagamma to requirements-dev.txt

canyon289:
Local tests fail if this isn't installed, which makes development confusing until the dev figures this out


OriolAbril:
@canyon289 you should modify the conda envs with `dev` in their name. Then pre-commit automatically writes/updates requirements.txt. It doesn't work the other way around

twiecki (approved):

Issue #5331, opened 2022-01-10
Remove automatic normalization in Multinomial and Categorical


ricardoV94:
### Discussed in https://github.com/pymc-devs/pymc/discussions/5246

<div type='discussions-op-text'>

<sup>Originally posted by **ricardoV94** December  8, 2021</sup>
This cropped up in #5234 

Should we stop doing automatic normalization of the p parameter? This can hide very wrong inputs such as 
```python
import pymc as pm

with pm.Model() as m:
  x = pm.Multinomial('x', n=5, p=[-2, -2, -2, -2])

m.logp({x: [3, 2, 0, 0]})
# array(-4.62888666)
```

If we decide to keep the automatic normalization, we can at least remove some checks in the logp definition, since they cannot be triggered in this case.</div>

***

I think this caused some problems e.g. if you a user specifies `[.3, .3, .3]` where things *almost* line up.

_Originally posted by @twiecki in https://github.com/pymc-devs/pymc/discussions/5246#discussioncomment-1897914_

***

Whenever the user creates a Multinomial / Categorical distribution with concrete (numpy) values we check if they are valid and if not we normalize them but also issue a UserWarning along the lines of:

```
p values sum up to 0.9, instead of 1.0. They will be automatically rescaled. You can rescale them directly to get rid of this warning.
```

In the logp or whenever we have symbolic inputs we don't do any invisible normalization, and let it evaluate to -inf as with invalid parameters in other distributions.

I think this covers most of the user cases and does not have big backwards compat issues.

_Originally posted by @ricardoV94 in https://github.com/pymc-devs/pymc/discussions/5246#discussioncomment-1935518_

LukeLB:
@ricardoV94 I'm happy to take this. Just to make sure I understand the ask, the normalisation step in `pm.Multinomial()` takes place using this line of code in the dist function (line 534 in multivariate.py):
`
p = p / at.sum(p, axis=-1, keepdims=True)
`
and the same code is used in the logp function within `pm.Categorical()` (line 1250 in discrete.py). Is this correct? For consistency should the normalisation step be moved to the dist function in `pm.Categorical()`?

ricardoV94:
Yes the normalization in both should be done only in the `dist` and only of the inputs are lists or numpy arrays

ricardoV94:
I've assigned the issue to you. Thanks for volunteering!

LukeLB:
[...]

PR #5370, opened 2022-01-18
 Remove automatic normalization in Multinomial and Categorical #5331 

LukeLB:
This PR removes the silent normalisation of p-values passed to a distribution. Instead, a UserWarning is raised when p-values do not sum to 1.0 and then normlisation is done. Examples are highlighted below:
```python
with pm.Model() as m:
    x = pm.Multinomial('x', n=5, p=[1,1,1,1,1])
```
> UserWarning: p values sum up to [5], instead of 1.0. They will be automatically rescaled. You can rescale them directly to get rid of this warning.
```python
with pm.Model() as m:
    x = pm.Categorical('x', p=[2,2,2,2,2])
```
> UserWarning: p values sum up to [10], instead of 1.0. They will be automatically rescaled. You can rescale them directly to get rid of this warning.

In addition, after discussion with @ricardoV94 negative p-values now raise a ValueError:

```python
with pm.Model() as model:
    x = pm.Multinomial("x", n = 5, p=[-1, 1, 1])
```
> ValueError: Negative probabilities are not valid

```python
with pm.Model() as model:
    x = pm.Categorical("x", p=[-1, 1, 1])
```
> ValueError: Negative probabilities are not valid

### Changes in this PR
- UserWarning is now raised when p-values passed which do not sum to 1.0. (Both Categorical and Multinomial)
- Normalisation of p-values in Categorical moved out of logp() to dist(). (Categorical only)
- ValueError is now raised when a negative p-value is passed. (Both Categorical and Multinomial)
- Changes to tests_distributions.py:
   - test_categorical_valid_p() and test_multinomial_invalid() now test whether a ValueError is raised using negative p-values during variable creation.
   - Two new tests have been written to check that a UserWarning is raised when using p-values that don't sum to 1.0. 

+ [ ] [consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples) I'm unsure whether this relevant for a PR like this?

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -145,6 +145,7 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
 - `math.log1mexp` and `math.log1mexp_numpy` will expect negative inputs in the future. A `FutureWarning` is now raised unless `negative_input=True` is set (see [#4860](https://github.com/pymc-devs/pymc/pull/4860)).
 - Changed name of `Lognormal` distribution to `LogNormal` to harmonize CamelCase usage for distribution names.
 - Attempt to iterate over MultiTrace will raise NotImplementedError.
+- Removed silent normalisation of `p` parameters in Categorical and Multinomial distributions (see [#5370](https://github.com/pymc-devs/pymc/pull/5370)).
 - ...
 
 
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -11,6 +11,8 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+import warnings
+
 import aesara.tensor as at
 import numpy as np
 
@@ -1233,7 +1235,16 @@ class Categorical(Discrete):
 
     @classmethod
     def dist(cls, p, **kwargs):
-
+        if isinstance(p, np.ndarray) or isinstance(p, list):
+            if (np.asarray(p) < 0).any():
+                raise ValueError(f"Negative `p` parameters are not valid, got: {p}")
+            p_sum = np.sum([p], axis=-1)
+            if not np.all(np.isclose(p_sum, 1.0)):
+                warnings.warn(
+                    f"`p` parameters sum to {p_sum}, instead of 1.0. They will be automatically rescaled. You can rescale them directly to get rid of this warning.",
+                    UserWarning,
+                )
+                p = p / at.sum(p, axis=-1, keepdims=True)
         p = at.as_tensor_variable(floatX(p))
         return super().dist([p], **kwargs)
 
@@ -1256,7 +1267,6 @@ class Categorical(Discrete):
         """
         k = at.shape(p)[-1]
         p_ = p
-        p = p_ / at.sum(p_, axis=-1, keepdims=True)
         value_clip = at.clip(value, 0, k - 1)
 
         if p.ndim > 1:
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -537,10 +537,18 @@ class Multinomial(Discrete):
 
     @classmethod
     def dist(cls, n, p, *args, **kwargs):
-        p = p / at.sum(p, axis=-1, keepdims=True)
+        if isinstance(p, np.ndarray) or isinstance(p, list):
+            if (np.asarray(p) < 0).any():
+                raise ValueError(f"Negative `p` parameters are not valid, got: {p}")
+            p_sum = np.sum([p], axis=-1)
+            if not np.all(np.isclose(p_sum, 1.0)):
+                warnings.warn(
+                    f"`p` parameters sum up to {p_sum}, instead of 1.0. They will be automatically rescaled. You can rescale them directly to get rid of this warning.",
+                    UserWarning,
+                )
+                p = p / at.sum(p, axis=-1, keepdims=True)
         n = at.as_tensor_variable(n)
         p = at.as_tensor_variable(p)
-
         return super().dist([n, p], *args, **kwargs)
 
     def get_moment(rv, size, n, p):
@@ -582,7 +590,7 @@ class Multinomial(Discrete):
         return check_parameters(
             res,
             p <= 1,
-            at.eq(at.sum(p, axis=-1), 1),
+            at.isclose(at.sum(p, axis=-1), 1),
             at.ge(n, 0),
             msg="p <= 1, sum(p) = 1, n >= 0",
         )
--- a/pymc/tests/test_distributions.py
+++ b/pymc/tests/test_distributions.py
@@ -2189,19 +2189,43 @@ class TestMatchesScipy:
             lambda value, n, p: scipy.stats.multinomial.logpmf(value, n, p),
         )
 
-    def test_multinomial_invalid(self):
-        # Test non-scalar invalid parameters/values
-        value = np.array([[1, 2, 2], [4, 0, 1]])
-
-        invalid_dist = Multinomial.dist(n=5, p=[-1, 1, 1], size=2)
-        # TODO: Multinomial normalizes p, so it is impossible to trigger p checks
-        # with pytest.raises(ParameterValueError):
-        with does_not_raise():
-            pm.logp(invalid_dist, value).eval()
-
-        value[1] -= 1
+    def test_multinomial_invalid_value(self):
+        # Test passing non-scalar invalid parameters/values to an otherwise valid Multinomial,
+        # evaluates to -inf
+        value = np.array([[1, 2, 2], [3, -1, 0]])
         valid_dist = Multinomial.dist(n=5, p=np.ones(3) / 3)
         assert np.all(np.isfinite(pm.logp(valid_dist, value).eval()) == np.array([True, False]))
+
+    def test_multinomial_negative_p(self):
+        # test passing a list/numpy with negative p raises an immediate error
+        with pytest.raises(ValueError, match="[-1, 1, 1]"):
+            with Model() as model:
+                x = Multinomial("x", n=5, p=[-1, 1, 1])
+
+    def test_multinomial_p_not_normalized(self):
+        # test UserWarning is raised for p vals that sum to more than 1
+        # and normaliation is triggered
+        with pytest.warns(UserWarning, match="[5]"):
+            with pm.Model() as m:
+                x = pm.Multinomial("x", n=5, p=[1, 1, 1, 1, 1])
+        # test stored p-vals have been normalised
+        assert np.isclose(m.x.owner.inputs[4].sum().eval(), 1.0)
+
+    def test_multinomial_negative_p_symbolic(self):
+        # Passing symbolic negative p does not raise an immediate error, but evaluating
+        # logp raises a ParameterValueError
+        with pytest.raises(ParameterValueError):
+            value = np.array([[1, 1, 1]])
+            invalid_dist = pm.Multinomial.dist(n=1, p=at.as_tensor_variable([-1, 0.5, 0.5]))
+            pm.logp(invalid_dist, value).eval()
+
+    def test_multinomial_p_not_normalized_symbolic(self):
+        # Passing symbolic p that do not add up to on does not raise any warning, but evaluating
+        # logp raises a ParameterValueError
+        with pytest.raises(ParameterValueError):
+            value = np.array([[1, 1, 1]])
+            invalid_dist = pm.Multinomial.dist(n=1, p=at.as_tensor_variable([1, 0.5, 0.5]))
+            pm.logp(invalid_dist, value).eval()
 
     @pytest.mark.parametrize("n", [(10), ([10, 11]), ([[5, 6], [10, 11]])])
     @pytest.mark.parametrize(
@@ -2317,12 +2341,22 @@ class TestMatchesScipy:
             np.array([-1, -1, 0, 0]),
         ],
     )
-    def test_categorical_valid_p(self, p):
-        with Model():
-            x = Categorical("x", p=p)
+    def test_categorical_negative_p(self, p):
+        with pytest.raises(ValueError, match=f"{p}"):
+            with Model():
+                x = Categorical("x", p=p)
 
-            with pytest.raises(ParameterValueError):
-                logp(x, 2).eval()
+    def test_categorical_negative_p_symbolic(self):
+        with pytest.raises(ParameterValueError):
+            value = np.array([[1, 1, 1]])
+            invalid_dist = pm.Categorical.dist(p=at.as_tensor_variable([-1, 0.5, 0.5]))
+            pm.logp(invalid_dist, value).eval()
+
+    def test_categorical_p_not_normalized_symbolic(self):
+        with pytest.raises(ParameterValueError):
+            value = np.array([[1, 1, 1]])
+            invalid_dist = pm.Categorical.dist(p=at.as_tensor_variable([2, 2, 2]))
+            pm.logp(invalid_dist, value).eval()
 
     @pytest.mark.parametrize("n", [2, 3, 4])
     def test_categorical(self, n):
@@ -2332,6 +2366,14 @@ class TestMatchesScipy:
             {"p": Simplex(n)},
             lambda value, p: categorical_logpdf(value, p),
         )
+
+    def test_categorical_p_not_normalized(self):
+        # test UserWarning is raised for p vals that sum to more than 1
+        # and normaliation is triggered
+        with pytest.warns(UserWarning, match="[5]"):
+            with pm.Model() as m:
+                x = pm.Categorical("x", p=[1, 1, 1, 1, 1])
+        assert np.isclose(m.x.owner.inputs[3].sum().eval(), 1.0)
 
     @pytest.mark.parametrize("n", [2, 3, 4])
     def test_orderedlogistic(self, n):
--- a/pymc/tests/test_idata_conversion.py
+++ b/pymc/tests/test_idata_conversion.py
@@ -531,6 +531,7 @@ class TestDataPyMC:
         data = np.random.multinomial(20, [0.2, 0.3, 0.5], size=20)
         with pm.Model(coords=coords):
             p = pm.Beta("p", 1, 1, size=(3,))
+            p = p / p.sum()
             pm.Multinomial("y", 20, p, dims=("experiment", "direction"), observed=data)
             idata = pm.sample(draws=50, chains=2, tune=100, return_inferencedata=True)
         test_dict = {


pymc/distributions/discrete.py
```diff
@@ -1224,7 +1226,16 @@ class Categorical(Discrete):
 
     @classmethod
     def dist(cls, p, **kwargs):
-
+        if isinstance(p, np.ndarray) or isinstance(p, list):
+            if (np.array(p) < 0).any():
+                raise ValueError("Negative probabilities are not valid")
+            p_sum = np.sum([p], axis=-1)
+            if (p_sum != 1.0).any():
```
ricardoV94: We probably have to use something more soft due to floating point errors:

```suggestion
            if not np.all(np.isclose(p_sum, 1.0)): 
```

pymc/distributions/multivariate.py
```diff
@@ -530,10 +530,18 @@ class Multinomial(Discrete):
 
     @classmethod
     def dist(cls, n, p, *args, **kwargs):
-        p = p / at.sum(p, axis=-1, keepdims=True)
+        if isinstance(p, np.ndarray) or isinstance(p, list):
+            if (np.array(p) < 0).any():
+                raise ValueError("Negative probabilities are not valid")
+            p_sum = np.sum([p], axis=-1)
+            if (p_sum != 1.0).any():
```
ricardoV94: Same here

ricardoV94: Also in the `multinomial.logp` we should use `at.isclose` instead of `at.eq` for the sum check. That is probably why one of the tests is failing

LukeLB: I have made these changes now.

ricardoV94:
The failing test seems to be due to an invalid `p` parameter here: 
https://github.com/pymc-devs/pymc/blob/f12b1fe04e4c50d4060803ad32dfa9c158cdf073/pymc/tests/test_idata_conversion.py#L590-L591

That should either have a `Dirichlet` prior or a `p = p / p.sum()` before going into the `pm.Multinomal`

ricardoV94:
While at it, this PR should include a release note about the changed behavior in https://github.com/pymc-devs/pymc/blob/f12b1fe04e4c50d4060803ad32dfa9c158cdf073/RELEASE-NOTES.md?plain=1

LukeLB:
Re-comitted with your suggestions. @ricardoV94 it's not clear to me why use of the beta distrubution led to a failing test? Its support is [0,1] so it won't produce negative values....

ricardoV94:
> Re-comitted with your suggestions. @ricardoV94 it's not clear to me why use of the beta distrubution led to a failing test? Its support is [0,1] so it won't produce negative values....

Yes, but the three independent betas don't add up to 1, which is the other requirement

pymc/tests/test_distributions.py
```diff
@@ -2219,12 +2219,12 @@ def test_multinomial(self, n):
...
         value = np.array([[1, 2, 2], [4, 0, 1]])
-
-        invalid_dist = Multinomial.dist(n=5, p=[-1, 1, 1], size=2)
-        # TODO: Multinomial normalizes p, so it is impossible to trigger p checks
-        # with pytest.raises(ParameterValueError):
-        with does_not_raise():
-            pm.logp(invalid_dist, value).eval()
+        
+        # test new behaviour, ValueError is now riased for negative probabilities prior to
+        # normalisation.
```
```suggestion
        # ValueError is raised for negative probabilities during initialization
```

ricardoV94:
I think it's better to separate clearly the different expected behaviors across different tests:
1. Passing a list/numpy with negative `p` raises an immediate error
2. Passing a list/numpy `p` that do not add up to one raises a warning and triggers normalization
3. Passing symbolic negative `p` like `at.as_tensor_variable([-1, 0.5, 0.5])` does not raise an immediate error, but evaluating the `logp` raises a `ParameterValueError`
4. Passing symbolic `p` that do not add up to on does not raise any warning, but evaluating `logp` raises a `ParameterValueError`
5. Passing an invalid `value` to an otherwise valid Multinomial, evaluates to `-inf` (it does not raise anything), as is being checked in here https://github.com/pymc-devs/pymc/blob/7fd960ce50395be551f700198e5f0e44f2888ea1/pymc/tests/test_distributions.py#L2229-L2231

LukeLB:
Yeah I agree! I'll get to writing these tests. Before I do, I have question about suggestion 4:
>     4. Passing symbolic `p` that do not add up to on does not raise any warning, but evaluating `logp` raises a `ParameterValueError`
What do you mean by symbolic here? Is this like the `pm.Beta()` that you highlighted in the failing test previously?

ricardoV94:
> Yeah I agree! I'll get to writing these tests. Before I do, I have question about suggestion 4:
> 
> > ```
> > 4. Passing symbolic `p` that do not add up to on does not raise any warning, but evaluating `logp` raises a `ParameterValueError`
> > ```
> 
> What do you mean by symbolic here? Is this like the `pm.Beta()` that you highlighted in the failing test previously?

Yes. More directly you can simply use `at.as_tensor_variable([-1, 0.5, 0.5])` to make an otherwise numpy/list input into a `symbolic` input for the purposes of testing

pymc/tests/test_distributions.py
```diff
@@ -2216,20 +2216,45 @@ def test_multinomial(self, n):
             lambda value, n, p: scipy.stats.multinomial.logpmf(value, n, p),
         )
 
-    def test_multinomial_invalid(self):
-        # Test non-scalar invalid parameters/values
+    def test_multinomial_invalid_value(self):
+        # Test passing non-scalar invalid parameters/values to an otherwise valid Multinomial, 
+        # evaluates to -inf
```
ricardoV94: Since we are only using the value once now, we can immediately create the final one (instead of doing `value[1] -= 1` after creation

pymc/tests/test_distributions.py
```diff
@@ -2216,20 +2216,45 @@ def test_multinomial(self, n):
...
-        # with pytest.raises(ParameterValueError):
-        with does_not_raise():
-            pm.logp(invalid_dist, value).eval()
-
         value[1] -= 1
         valid_dist = Multinomial.dist(n=5, p=np.ones(3) / 3)
         assert np.all(np.isfinite(pm.logp(valid_dist, value).eval()) == np.array([True, False]))
-
+    
+    def test_multinomial_negativep_value_error(self):
```
```suggestion
    def test_multinomial_negative_p(self):
```

pymc/tests/test_distributions.py
```diff
@@ -2216,20 +2216,45 @@ def test_multinomial(self, n):
...
+        # test passing a list/numpy with negative p raises an immediate error
+        with pytest.raises(ValueError):
```
ricardoV94: You can add a `match` string with the expected error message in `pytest.raises` for a more strict test.

pymc/tests/test_distributions.py
```diff
@@ -2216,20 +2216,45 @@ def test_multinomial(self, n):
...
+            with Model() as model:
+                x = Multinomial("x", n = 5, p=[-1, 1, 1])
+    
+    def test_multinomial_pnormalisation_warning(self):
```
```suggestion
    def test_multinomial_p_not_normalized(self):
```

pymc/tests/test_distributions.py
```diff
@@ -2216,20 +2216,45 @@ def test_multinomial(self, n):
...
+        # test UserWarning is raised for p vals that sum to more than 1
+        # and normaliation is triggered 
+        with pytest.warns(UserWarning):
+            with pm.Model() as m:
+                x = pm.Multinomial('x', n=5, p=[1,1,1,1,1])
+        # test stored p-vals have been normalised 
+        assert (m.x.owner.inputs[4].eval() == [1,1,1,1,1]/np.sum([1,1,1,1,1])).all()
```
```suggestion
        assert np.isclose(m.x.owner.inputs[4].sum().eval(), 1.0)
```

pymc/tests/test_distributions.py
```diff
@@ -2216,20 +2216,45 @@ def test_multinomial(self, n):
...
+        
+    def test_multinomial_negative_symbolicp_logperror(self):
```
```suggestion
    def test_multinomial_negative_p_symbolic(self):
```

pymc/tests/test_distributions.py
```diff
@@ -2216,20 +2216,45 @@ def test_multinomial(self, n):
...
+        # Passing symbolic negative p does not raise an immediate error, but evaluating 
+        # logp raises a ParameterValueError
+        with pytest.raises(ParameterValueError):  
+            value = np.array([[1, 1, 1]])
+            invalid_dist = pm.Multinomial.dist(n=1, p=at.as_tensor_variable([-1, 0.5, 0.5]))
+            pm.logp(invalid_dist, value).eval()
+            
+    def test_multinomial_nonnorm_symbolicp_logperror(self):
```
```suggestion
    def test_multinomial_p_not_normalized_symbolic(self):
```

pymc/distributions/multivariate.py
```diff
@@ -530,10 +530,18 @@ class Multinomial(Discrete):
 
     @classmethod
     def dist(cls, n, p, *args, **kwargs):
-        p = p / at.sum(p, axis=-1, keepdims=True)
+        if isinstance(p, np.ndarray) or isinstance(p, list):
+            if (np.array(p) < 0).any():
+                raise ValueError("Negative probabilities are not valid")
```
```suggestion
                raise ValueError(f"Negative `p` parameters are not valid, got: {p}")
```

pymc/distributions/multivariate.py
```diff
@@ -530,10 +530,18 @@ class Multinomial(Discrete):
...
+            p_sum = np.sum([p], axis=-1)
+            if not np.all(np.isclose(p_sum, 1.0)):  
+                warnings.warn(
+                    f"p values sum up to {p_sum}, instead of 1.0. They will be automatically rescaled. You can rescale them directly to get rid of this warning.",
```
```suggestion
                    f"`p` parameters sum up to {p_sum}, instead of 1.0. They will be automatically rescaled. You can rescale them directly to get rid of this warning.",
```

pymc/distributions/discrete.py
```diff
@@ -1224,7 +1226,16 @@ class Categorical(Discrete):
...
     @classmethod
     def dist(cls, p, **kwargs):
-
+        if isinstance(p, np.ndarray) or isinstance(p, list):
+            if (np.array(p) < 0).any():
+                raise ValueError("Negative probabilities are not valid")
+            p_sum = np.sum([p], axis=-1)
+            if not np.all(np.isclose(p_sum, 1.0)): 
+                warnings.warn(
+                    f"p values sum up to {p_sum}, instead of 1.0. They will be automatically rescaled. You can rescale them directly to get rid of this warning.",
```
```suggestion
                    f"`p` parameters sum to {p_sum}, instead of 1.0. They will be automatically rescaled. You can rescale them directly to get rid of this warning.",
```

RELEASE-NOTES.md
```diff
@@ -132,6 +132,7 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
 - `math.log1mexp` and `math.log1mexp_numpy` will expect negative inputs in the future. A `FutureWarning` is now raised unless `negative_input=True` is set (see [#4860](https://github.com/pymc-devs/pymc/pull/4860)).
 - Changed name of `Lognormal` distribution to `LogNormal` to harmonize CamelCase usage for distribution names.
 - Attempt to iterate over MultiTrace will raise NotImplementedError.
+- Removed silent normalisation of p-values passed to Categorical and Multinomial distribution. Instead, a UserWarning is raised when p-values do not sum to 1.0 and then normalisation is done. Normalisation in Categorical now takes place in `dist` rather than `logp`. In addition, negative p-values now raise ValueError.
```
```suggestion
- Removed silent normalisation of `p` parameters in Categorical and Multinomial distributions (see [#5370](https://github.com/pymc-devs/pymc/pull/5370)).
```

pymc/distributions/discrete.py
```diff
@@ -1224,7 +1226,16 @@ class Categorical(Discrete):
 
     @classmethod
     def dist(cls, p, **kwargs):
-
+        if isinstance(p, np.ndarray) or isinstance(p, list):
+            if (np.array(p) < 0).any():
+                raise ValueError("Negative probabilities are not valid")
```
```suggestion
                raise ValueError(f"Negative `p` parameters are not valid, got: {p}")
```

pymc/distributions/discrete.py
```diff
@@ -1224,7 +1226,16 @@ class Categorical(Discrete):
 
     @classmethod
     def dist(cls, p, **kwargs):
-
+        if isinstance(p, np.ndarray) or isinstance(p, list):
+            if (np.array(p) < 0).any():
```
```suggestion
            if (np.asarray(p) < 0).any():
```
This is slightly more efficient if the input is already a numpy array

pymc/distributions/multivariate.py
```diff
@@ -530,10 +530,18 @@ class Multinomial(Discrete):
 
     @classmethod
     def dist(cls, n, p, *args, **kwargs):
-        p = p / at.sum(p, axis=-1, keepdims=True)
+        if isinstance(p, np.ndarray) or isinstance(p, list):
+            if (np.array(p) < 0).any():
```
```suggestion
            if (np.asarray(p) < 0).any():
```

pymc/tests/test_distributions.py
```diff
@@ -2344,12 +2369,11 @@ def test_categorical_bounds(self):
             np.array([-1, -1, 0, 0]),
         ],
     )
-    def test_categorical_valid_p(self, p):
-        with Model():
-            x = Categorical("x", p=p)
-
-            with pytest.raises(ParameterValueError):
-                logp(x, 2).eval()
+    def test_categorical_invalid_p(self, p):
```
```suggestion
    def test_categorical_negative_p(self, p):
```

pymc/tests/test_distributions.py
```diff
@@ -2359,7 +2383,15 @@ def test_categorical(self, n):
             {"p": Simplex(n)},
             lambda value, p: categorical_logpdf(value, p),
         )
-
+        
+    def test_categorical_pnormalisation_warning(self):
```
```suggestion
    def test_categorical_p_not_normalized(self):
```

ricardoV94 (changes requested):
Looks great. I left some minor naming/testing suggestions

LukeLB (commented):
Looks great thanks for making the changes!

ricardoV94:
> Looks great thanks for making the changes!

These are only suggestions. You'll have to commit them (this can be easily done via Github) for them to be "applied".

LukeLB:
> > Looks great thanks for making the changes!
> 
> These are only suggestions. You'll have to commit them (this can be easily done via Github) for them to be "applied".

Appologies, I thought reviewing them would commit them. Will commit them now!

pymc/tests/test_distributions.py
LukeLB: Added `match` string to tests expecting UserWarning and ValueError.

LukeLB:
Ah! Sorry it looks like there were some commit suggestions that github had hidden. I'll need to commit these.

LukeLB:
I think this might be another floating point error, I have now changed to:
```python
assert np.isclose(m.x.owner.inputs[3].sum().eval(), 1.0)
```
in the failing test. test_distrubution.py passed fine on my machine not sure why it fails here.

ricardoV94:
You should also check locally with float32, often that's why tests fail in here but not locally. You can do this by adding these lines at the very top of the test file

```python
import aesara
aesara.config.floatX = "float32"
```

pymc/tests/test_distributions.py
```diff
@@ -2216,20 +2216,44 @@ def test_multinomial(self, n):
...
+        
+    def test_multinomial_negative_symbolicp_logperror(self):
+        # Passing symbolic negative p does not raise an immediate error, but evaluating 
+        # logp raises a ParameterValueError
+        with pytest.raises(ParameterValueError):  
+            value = np.array([[1, 1, 1]])
+            invalid_dist = pm.Multinomial.dist(n=1, p=at.as_tensor_variable([-1, 0.5, 0.5]))
+            pm.logp(invalid_dist, value).eval()
+            
+    def test_multinomial_nonnorm_symbolicp_logperror(self):
```
ricardoV94: Any reason for this type of test naming? Seems strange to use a mix of underscore and concatenated words. That's why I suggested the previous names.

LukeLB: No not really, I thought I had accepted your changes to these names. I'll go through and change to the correct style.

ricardoV94: Ah... you might have reverted them by accident when you did your last commit. Perhaps you forgot to first pull from your remote before doing new commits?

LukeLB: I think that was it, I had a merge conflict on those tests and fixed the underlying code but not the naming style. My bad!

LukeLB:
> You should also check locally with float32, often that's why tests fail in here but not locally. You can do this by adding these lines at the very top of the test file
> 
> ```python
> import aesara
> aesara.config.floatX = "float32"
> ```

I've ran test_distrubution.py locally using this and they all pass now. Thanks for the heads up!

pymc/tests/test_distributions.py
```diff
@@ -2216,20 +2216,44 @@ def test_multinomial(self, n):
...
+        
+    def test_multinomial_negative_p_symbolic(self):
+        # Passing symbolic negative p does not raise an immediate error, but evaluating 
+        # logp raises a ParameterValueError
+        with pytest.raises(ParameterValueError):  
+            value = np.array([[1, 1, 1]])
+            invalid_dist = pm.Multinomial.dist(n=1, p=at.as_tensor_variable([-1, 0.5, 0.5]))
+            pm.logp(invalid_dist, value).eval()
+            
+    def test_multinomial_nonnorm_p_symbolic(self):
```
ricardoV94: This one seems to have escaped

```suggestion
    def test_multinomial_p_not_normalized_symbolic(self):
```

pymc/tests/test_distributions.py
```diff
@@ -2344,12 +2368,11 @@ def test_categorical_bounds(self):
             np.array([-1, -1, 0, 0]),
         ],
     )
-    def test_categorical_valid_p(self, p):
-        with Model():
-            x = Categorical("x", p=p)
-
-            with pytest.raises(ParameterValueError):
-                logp(x, 2).eval()
+    def test_categorical_negative_p(self, p):
```
ricardoV94, line 2344: Is the symbolic version equivalent already covered by other tests?

LukeLB: I don't think so. I'll write the equivlent tests for Categorical to cover this.

LukeLB:
So in the future to stop the pre-commit failure I should run
```
pre-commit run --all-files
```
prior to commiting? Do I need to run this locally and commit for this PR? Or has this been run already?

ricardoV94:
Easiest way is to install pre-commit, so it will run everytime you try to create a commit. https://docs.pymc.io/en/latest/contributing/python_style.html

ricardoV94:
> Do I need to run this locally and commit for this PR

Yes, you still need to do that

ricardoV94 (approved):
Looks good! Checking if the tests pass 🤞

LukeLB:
Woooooooooooo

LukeLB:
Thank you for all the help @ricardoV94! Even though it's been a small issue I feel like I've learnt a lot.

ricardoV94:
@LukeLB just merged it. Congrats for your first contributon! Looking forward to your next one :D

PR #5376, opened 2022-01-20
Restructure test jobs

ricardoV94:
Trying to balance the runtime across jobs more evenly

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -38,7 +38,6 @@ jobs:
         # → pytest will run only these files
           - |
             --ignore=pymc/tests/test_distributions_timeseries.py
-            --ignore=pymc/tests/test_initial_point.py
             --ignore=pymc/tests/test_mixture.py
             --ignore=pymc/tests/test_model_graph.py
             --ignore=pymc/tests/test_modelcontext.py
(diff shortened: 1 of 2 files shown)


.github/workflows/pytest.yml
```diff
@@ -64,9 +63,11 @@ jobs:
             --ignore=pymc/tests/test_distributions.py
             --ignore=pymc/tests/test_distributions_random.py
             --ignore=pymc/tests/test_idata_conversion.py
+            --ignore=pymc/tests/test_smc.py
+            --ignore=pymc/tests/test_bart.py
+            --ignore=pymc/tests/test_missing.py
 
           - |
-            pymc/tests/test_initial_point.py
```
ricardoV94, line 69: This test module is to short to justify placing it manually (everything takes ~10 seconds locally)

ricardoV94:
Seems to be working. All jobs in `pytest` are now `~ <= 20` minutes. I will follow up with a PR that speeds up the quite slow `test_step.py` module which should further balance things out.

As expected joining the two fast jobs in `arviz_compat` makes no difference, because the other job is still quite slower than the two combined.

twiecki (approved):

PR #5381, opened 2022-01-22
Add additional compiler flag to work around GCC bug.

jedludlow:
Work around compiler bug in GCC < 8.4 related to structured exception handling registers on Windows.
See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65782 for details.

Addresses compiler failure issues as described in #4937. This change is applied to the v3 branch.

This fix was tested on an Intel Xeon W-2125 processor where previous compile errors were observed with GCC 5.3.0. This fix resolved the compile errors.

### DIFF
--- a/pymc3/__init__.py
+++ b/pymc3/__init__.py
@@ -82,10 +82,16 @@ def __set_compiler_flags():
     # Workarounds for Theano compiler problems on various platforms
     current = theano.config.gcc__cxxflags
     augmented = f"{current} -Wno-c++11-narrowing"
+
     # Work around compiler bug in GCC < 8.4 related to structured exception
     # handling registers on Windows.
     # See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65782 for details.
-    augmented = f"{augmented} -fno-asynchronous-unwind-tables"
+    # First disable C++ exception handling altogether since it's not needed
+    # for the C extensions that we generate.
+    augmented = f"{augmented} -fno-exceptions"
+    # Now disable the generation of stack unwinding tables.
+    augmented = f"{augmented} -fno-unwind-tables -fno-asynchronous-unwind-tables"
+
     theano.config.gcc__cxxflags = augmented
 
 
--- a/pymc3/__init__.py
+++ b/pymc3/__init__.py
@@ -81,7 +81,12 @@ def _check_backend_version():
 def __set_compiler_flags():
     # Workarounds for Theano compiler problems on various platforms
     current = theano.config.gcc__cxxflags
-    theano.config.gcc__cxxflags = f"{current} -Wno-c++11-narrowing"
+    augmented = f"{current} -Wno-c++11-narrowing"
+    # Work around compiler bug in GCC < 8.4 related to structured exception
+    # handling registers on Windows.
+    # See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65782 for details.
+    augmented = f"{augmented} -fno-asynchronous-unwind-tables"
+    theano.config.gcc__cxxflags = augmented
 
 
 def _hotfix_theano_printing():


twiecki:
Ugh, seems like that introduces problems elsewhere? 
```
E               Exception: ("Compilation failed (return status=1): C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\ccF25RbP.o: In function `__struct_compiled_op_m08938589cd8f97605f089a439e04454bcb25ecee867cab1070a489f1eba08c2e_executor':\r. C:/Users/runneradmin/AppData/Local/Theano/compiledir_Windows-10-10.0.20348-SP0-Intel64_Family_6_Model_85_Stepping_4_GenuineIntel-3.8.12-64/tmpbj_hc544/mod.cpp:565: undefined reference to `__gxx_personality_sj0'\r. C:/Users/runneradmin/AppData/Local/Theano/compiledir_Windows-10-10.0.20348-SP0-Intel64_Family_6_Model_85_Stepping_4_GenuineIntel-3.8.12-64/tmpbj_hc544/mod.cpp:565: undefined reference to `_Unwind_SjLj_Register'\r. C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\ccF25RbP.o: In function `run':\r. C:/Users/runneradmin/AppData/Local/Theano/compiledir_Windows-10-10.0.20348-SP0-Intel64_Family_6_Model_85_Stepping_4_GenuineIntel-3.8.12-64/tmpbj_hc544/mod.cpp:537: undefined reference to `_Unwind_SjLj_Unregister'\r. C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\ccF25RbP.o: In function `__gnu_cxx::new_allocator<std::pair<int, int> >::deallocate(std::pair<int, int>*, unsigned long long)':\r. C:/Miniconda/envs/pymc3-dev-py38/Library/mingw-w64/include/c++/5.3.0/ext/new_allocator.h:110: undefined reference to `_Unwind_SjLj_Resume'\r. C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\ccF25RbP.o:mod.cpp:(.data+0xc0): undefined reference to `__gxx_personality_sj0'\r. collect2.exe: error: ld returned 1 exit status\r. ", 'FunctionGraph(gammaln(<TensorType(int64, matrix)>))')
1871

1872
C:\Miniconda\envs\pymc3-dev-py38\lib\site-packages\theano\link\c\cmodule.py:2546: Exception
```

Do those tests pass locally for you?

jedludlow:
No, the tests fail locally for me, as well. I didn't even consider running them before because #5363 went through on the main branch without a hitch. ~~In retrospect, it looks like many of these Windows-specific checks are not currently running in the continuous integration testing for main, so #5363 may need additional attention, as well.~~ The continuous integration testing for main includes these same tests, and they all pass on that branch which uses `aesara` instead of `theano`.

This looks like a linker failure, where one object file is expecting to find functions in another but not succeeding. I suspect what's happening is that one call to the complier is using the `-fno-asynchronous-unwind-tables` flag, but some other call from a different place in the code is not. That would produce one object file with the feature enabled and another without, and the linker would be unable to resolve the references between them. I will have to look more closely later today.

This PR proposes a complier flag change only inside of `theano` compiler commands. Is there anywhere else inside of `pymc3` itself that is generating compiler commands, perhaps related to the sampler where these tests are failing?

twiecki:
Hm, I don't know where that could happen. You could rule that out perhaps by setting the flag in .theanorc.

jedludlow:
@twiecki I spent a fair bit of time investigating why this is failing, and I wrote up my findings in https://github.com/pymc-devs/pymc/issues/4937#issuecomment-1021352178. In short, this PR can be made to pass by disabling more of the C++ exception handling mechanisms. See the comment for more details and the option for solving this that I came up with. I'd probably vote for just disabling C++ exception handling altogether because it's not clear how it helps in this application. If you'd like me to update the PR with this option, let me know.

twiecki:
Thanks!

Issue #5383, opened 2022-01-24
Increase support for batched multivariate distributions


ricardoV94:
The `logp` of several multivariate distributions does not work (or is not tested) for arbitrarily batched dimensions. Some cases I could confirm include:

- [x] MvNormal - #6897 
- [x] MvStudentT #6897 
- [ ] KroneckerNormal
- [ ] MatrixNormal
- [x] LKJCorr
- [ ] LKJCholeskyCov
- [x] StickBreakingWeights (alpha) #6042
- [x] Wishart (may be fine to ignore) #8246 
- [ ] CARRV (maybe better to leave this one as is)

Reproducible code:
```python
size = (4, 3)
pm.logp(pm.MvNormal.dist(mu=np.ones(2), cov=np.eye(2), size=size), np.ones((*size, 2)))
# ValueError: Invalid dimension for value: 3
pm.logp(pm.MvStudentT.dist(nu=3, mu=np.ones(2), cov=np.eye(2), size=size), np.ones((*size, 2)))
# ValueError: Invalid dimension for value: 3
```

Distributions that already support (and have tests for) arbitrary shapes
- [x] Dirichlet
- [x] Multinomial
- [x] DirichletMultinomial - see https://github.com/pymc-devs/pymc/pull/5234

purna135:
I'd like to work on this, could you please point me in the right direction?

Sayam753:
I think one can start by exploring the implementation of Random Variables in v4. @ricardoV94 did a great job of [explaining design steps here](https://github.com/pymc-devs/pymc/issues/4686#issuecomment-839541040).

- Then, gather all Ops from Aesara that are used in the implementations of multivariate distributions.
- Generalize the Ops to work beyond 2d inputs.
- Make sure the updated Ops work readily / Make use of Aesara's broadcasting capabilities to integrate the updated Ops in multivariate distributions.
- Update any/all test cases related to these Ops and multivariate distributions.
- Bonus here would be to see how mixture distributions will play with generalized multivariate distributions.

ricardoV94:
>     * Bonus here would be to see how mixture distributions will play with generalized multivariate distributions.

They should work fine out of the box. We have tests for batched Dirichlet for instance. Only limitation is whether the distribution accepts batched values in the logp

Sayam753:
[...]

PR #5386, opened 2022-01-25
Allow for scalar or size 1 `mu` in `MvNormal` and `MvStudentT`

ricardoV94:
Just for user friendliness, as this was accepted in V3


pymc/distributions/multivariate.py
```diff
@@ -233,6 +233,9 @@ class MvNormal(Continuous):
     def dist(cls, mu, cov=None, tau=None, chol=None, lower=True, **kwargs):
         mu = at.as_tensor_variable(mu)
         cov = quaddist_matrix(cov, chol, tau, lower)
+        # Aesara is stricter about the shape of mu, than PyMC used to be
+        if mu.ndim < 2:
+            mu = at.broadcast_to(mu, cov.shape[0])
```
Sayam753: AFAIK, `mu` will be broadcasted to shape (10, 3) in the case below in v3.

```python
>>> pm.MvNormal.dist(mu=np.ones((10, 1)), cov=np.eye(3))
```

More generally, `mu.shape` and `cov.shape[:-1]` should be broadcasted.
Is this case handled here as well?

ricardoV94: Thanks, let me think about it

ricardoV94: @Sayam753 I pushed an update which broadcasts `mu` to `cov[..., -1]`

ricardoV94:
Failing test is flaky: https://github.com/pymc-devs/pymc/issues/5411

Sayam753 (approved):
LGTM 🚀 

We may need to revisit here after https://github.com/pymc-devs/pymc/issues/5383 gets solved, because if `cov` parameter supports batch dimensions, it may require broadcasting against `mu`.

ricardoV94:
Thanks for reviewing @Sayam753 

Are you interested in picking some of #5383?

Sayam753:
Yeah. Let's do it.

PR #5389, opened 2022-01-25
API backports and deprecation warnings

michaelosthege:
This PR makes some simple backports to reduce API breakage in the `v3`→`v4` transition.


RELEASE-NOTES.md
```diff
@@ -5,6 +5,7 @@
 + The `pm.logp(rv, x)` syntax is now available and recommended to make your model code `v4`-ready. Note that this backport is just an alias and much less capable than what's available with `pymc >=4` (see [#5083](https://github.com/pymc-devs/pymc/pulls/5083)).
 + The `pm.Distribution(testval=...)` kwarg was deprecated and will be replaced by `pm.Distribution(initval=...)`in `pymc >=4` (see [#5226](https://github.com/pymc-devs/pymc/pulls/5226)).
 + The `pm.sample(start=...)` kwarg was deprecated and will be replaced by `pm.sample(initvals=...)`in `pymc >=4` (see [#5226](https://github.com/pymc-devs/pymc/pulls/5226)).
++ `pm.Lognormal` is now available as an alias for `pm.Lognormal` (see [#5389](https://github.com/pymc-devs/pymc/pull/5389)).
```
OriolAbril, line 8: Capitalization is wrong, I think the first one should be LogNormal

conda-envs/environment-dev-py37.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
 - mkl-service
 - nbsphinx>=0.4
 - numpy=1.15
-- numpydoc>=0.9
+- numpydoc<1.2
```
OriolAbril, line 12: Would using numpydoc! =1.2.0 allow installation of 1.2.1? The PR fix was merged today, the patch release should hopefully not take long, and with that we could forget about the pin completely

OriolAbril (commented):
Might have missed it but I don't see the new name in the docs, it should be added to the api/distributions/continuous file

conda-envs/environment-dev-py37.yml
michaelosthege: While they labeled it `1.2.0` on GitHub, the PyPI release is just `1.2`, so I'm not sure if `!=1.2.0` would work.
Also we have the `<1.2` on `main` already.
But I'm now "watching" the numpydoc repo and will get a notification 🔔 when they do the next release.

michaelosthege:
@OriolAbril can we merge this?

OriolAbril (approved):

PR #5399, opened 2022-01-26
Added blas to conda dev envs

fonnesbeck:
Running 4.0.0b2 on Linux generates a warning about missing blas header. It even appears on [the current overview notebook on the rendered docs](https://docs.pymc.io/en/latest/learn/examples/pymc_overview.html).

image

Installing blas from conda resolves the issue for me, so I'm adding it to the dev conda envs.

### DIFF
--- a/conda-envs/environment-dev-py37.yml
+++ b/conda-envs/environment-dev-py37.yml
@@ -7,6 +7,7 @@ dependencies:
 - aeppl=0.0.18
 - aesara=2.3.6
 - arviz>=0.11.4
+- blas
 - cachetools>=4.2.1
 - cloudpickle
 - fastprogress>=0.2.0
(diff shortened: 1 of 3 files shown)


twiecki:
I think this should be fixed at the aesara level, I recently removed all the compiler references here and added them there (including blas) to have separation of concerns. PyMC does not compile anything, only aesara does.

ghost:
I agree that it should not be a dependency, but you want it in your conda env (at least if you are running Linux--not sure about other platforms).

twiecki:
I would like to understand why it's not pulled in through aesara, it's included here: https://github.com/conda-forge/aesara-feedstock/blob/master/recipe/meta.yaml#L25

does `libblas` get installed, just not `blas`?

twiecki:
Ping @fonnesbeck .

fonnesbeck:
Good question. Just blas, but the warning goes away. Probably means that it uses the available binary rather than trying to compile (?). Not sure how else the warning would be fixed.

twiecki:
OK, I think we should add `blas` to aesara then.

twiecki:
OK, I added it to the aesara feedstock which is the appropriate place.

PR #5403, opened 2022-01-27
Disable C++ exception handling

jedludlow:
Update GCC compiler options to disable C++ exception handling altogether.
Also disable the creation of stack unwinding tables since these will not be
required without exceptions enabled.

This should be considered an interim solution to the compiler bugs on Windows until a better C++ compiler situation can be found. It should make it much simpler for new Windows users to get a working PyMC by allowing `conda` to install the MinGW C++ compiler.

Addresses #4937, specifically https://github.com/pymc-devs/pymc/issues/4937#issuecomment-1021352178, for the main branch.

### DIFF
--- a/pymc/__init__.py
+++ b/pymc/__init__.py
@@ -34,10 +34,16 @@ def __set_compiler_flags():
 
     current = aesara.config.gcc__cxxflags
     augmented = f"{current} -Wno-c++11-narrowing"
+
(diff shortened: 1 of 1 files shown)


twiecki (approved):

twiecki:
Thanks @jedludlow!

PR #5414, opened 2022-01-27
fix: check the output type of ode_func

mirko-m:
This PR addresses the issue discussed here https://discourse.pymc.io/t/return-value-for-2n-dimensional-ode-system/4141?u=mirko-m .

When estimating the parameters of a system of ODEs we need to define a a function called `ode_func` which returns the right hand side of the ODE system. It is not clear what the return type of `ode_func` is. In some of the examples it is a list, but for some ODE systems the user may prefer to return a  `TensorVariable` or a `numpy.ndarray`. The changes in this PR check for the return type of `ode_func` and then convert to a `TensorVariable`.

I am not sure if using `at.shared()` is the best method for casting a`numpy.ndarray`  to a `TensorVariable` .

I do not have a test for the changes I made, but it should be possible to adapt one of the existing examples. Let me know if you need me to do this.

### DIFF
--- a/pymc/ode/utils.py
+++ b/pymc/ode/utils.py
@@ -108,7 +108,9 @@ def augment_system(ode_func, n_states, n_theta):
     else:
         # Stack the results of the ode_func into a single tensor variable
         if not isinstance(yhat, (list, tuple)):
-            yhat = (yhat,)
+            raise TypeError(
+                f"Unexpected type, {type(yhat)}, returned by ode_func. TensorVariable, list or tuple is expected."
+            )
         t_yhat = at.stack(yhat, axis=0)
     if t_yhat.ndim > 1:
         raise ValueError(
--- a/pymc/ode/utils.py
+++ b/pymc/ode/utils.py
@@ -110,7 +110,9 @@ def augment_system(ode_func, n_states, n_theta):
         else:
             t_yhat = yhat
     elif isinstance(yhat, np.ndarray):
-        t_yhat = at.shared(yhat)
+        msg = f'Invalid Output type for odefunc: {type(yhat)}.\n'
+        msg += 'Valid output types are list, tuple, or at.TensorVariable.'
+        raise TypeError(msg)
     else:
         # Stack the results of the ode_func into a single tensor variable
         if not isinstance(yhat, (list, tuple)):
--- a/pymc/ode/utils.py
+++ b/pymc/ode/utils.py
@@ -107,11 +107,15 @@ def augment_system(ode_func, n_states, n_theta):
         if yhat.ndim == 0:
             # Convert yhat from scalar to array
             t_yhat = at.stack((yhat,), axis=0)
-        else:
+        elif yhat.ndim == 1:
             t_yhat = yhat
+        else:
+            raise ValueError(
+                f"The odefunc returned a {yhat.ndim}-dimensional tensor, but 0 or 1 dimensions were expected."
+            )
     elif isinstance(yhat, np.ndarray):
-        msg = f'Invalid Output type for odefunc: {type(yhat)}.\n'
-        msg += 'Valid output types are list, tuple, or at.TensorVariable.'
+        msg = f"Invalid Output type for odefunc: {type(yhat)}.\n"
+        msg += "Valid output types are list, tuple, or at.TensorVariable."
         raise TypeError(msg)
     else:
         # Stack the results of the ode_func into a single tensor variable
--- a/pymc/ode/utils.py
+++ b/pymc/ode/utils.py
@@ -105,15 +105,15 @@ def augment_system(ode_func, n_states, n_theta):
     yhat = ode_func(t_y, t_t, t_p[n_states:])
     if isinstance(yhat, at.TensorVariable):
         t_yhat = at.atleast_1d(yhat)
-        if t_yhat.ndim > 1:
-            raise ValueError(
-                f"The odefunc returned a {yhat.ndim}-dimensional tensor, but 0 or 1 dimensions were expected."
-            )
     else:
         # Stack the results of the ode_func into a single tensor variable
         if not isinstance(yhat, (list, tuple)):
             yhat = (yhat,)
         t_yhat = at.stack(yhat, axis=0)
+    if t_yhat.ndim > 1:
+        raise ValueError(
+            f"The odefunc returned a {yhat.ndim}-dimensional tensor, but 0 or 1 dimensions were expected."
+        )
 
     # Now compute gradients
     J = at.jacobian(t_yhat, t_y)
--- a/pymc/ode/utils.py
+++ b/pymc/ode/utils.py
@@ -103,10 +103,15 @@ def augment_system(ode_func, n_states, n_theta):
 
     # Get symbolic representation of the ODEs by passing tensors for y, t and theta
     yhat = ode_func(t_y, t_t, t_p[n_states:])
-    # Stack the results of the ode_func into a single tensor variable
-    if not isinstance(yhat, (list, tuple)):
-        yhat = (yhat,)
-    t_yhat = at.stack(yhat, axis=0)
+    if isinstance(yhat, at.TensorVariable):
+        t_yhat = yhat
+    elif isinstance(yhat, np.ndarray):
+        t_yhat = at.shared(yhat)
+    else:
+        # Stack the results of the ode_func into a single tensor variable
+        if not isinstance(yhat, (list, tuple)):
+            yhat = (yhat,)
+        t_yhat = at.stack(yhat, axis=0)
 
     # Now compute gradients
     J = at.jacobian(t_yhat, t_y)
--- a/pymc/ode/utils.py
+++ b/pymc/ode/utils.py
@@ -112,7 +112,7 @@ def augment_system(ode_func, n_states, n_theta):
         t_yhat = at.stack(yhat, axis=0)
     if t_yhat.ndim > 1:
         raise ValueError(
-            f"The odefunc returned a {yhat.ndim}-dimensional tensor, but 0 or 1 dimensions were expected."
+            f"The odefunc returned a {t_yhat.ndim}-dimensional tensor, but 0 or 1 dimensions were expected."
         )
 
     # Now compute gradients
--- a/pymc/ode/utils.py
+++ b/pymc/ode/utils.py
@@ -113,10 +113,6 @@ def augment_system(ode_func, n_states, n_theta):
             raise ValueError(
                 f"The odefunc returned a {yhat.ndim}-dimensional tensor, but 0 or 1 dimensions were expected."
             )
-    elif isinstance(yhat, np.ndarray):
-        msg = f"Invalid Output type for odefunc: {type(yhat)}.\n"
-        msg += "Valid output types are list, tuple, or at.TensorVariable."
-        raise TypeError(msg)
     else:
         # Stack the results of the ode_func into a single tensor variable
         if not isinstance(yhat, (list, tuple)):
--- a/pymc/ode/utils.py
+++ b/pymc/ode/utils.py
@@ -104,7 +104,11 @@ def augment_system(ode_func, n_states, n_theta):
     # Get symbolic representation of the ODEs by passing tensors for y, t and theta
     yhat = ode_func(t_y, t_t, t_p[n_states:])
     if isinstance(yhat, at.TensorVariable):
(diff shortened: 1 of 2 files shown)


canyon289:
@mirko-m thanks for making a contribution. As you noted please make a test. I'll defer your question about `at.shared` to one of the inhouse Aesara experts!

ricardoV94:
I am not familiar with ODE module but I don't think `np.ndarray` is a ever a valid output since the lines immediately below are trying to extract gradients from it.

mirko-m:
Thanks for your replies. My changes actually broke some of the existing tests. I am trying to debug this, but in my development environment the `import pymc` statement takes a lot of time to execute (several minutes). This makes the debugging process very slow? Do you happen to know a fix or workaround for this?

I installed the `main` branch on Ubuntu-20.04 running under WSL using
```bash
pip install -r requirements-dev.txt
pip install -e .
```

mirko-m:
I made the following changes:

- Fix the case where `ode_func` returns a scalar `TensorVariable`.
- Add a test for the case where `ode_func` returns a vector `TensorVariable`.
- Raise a `TypeError` when `ode_func` returns a np.ndarray.
- Add a test checking that the `TypeError` is raised.

Do we need to add more tests for the case where the output of` `ode_func` is a `TensorVariable`? And do we want to put them in a separate class or leave them in `TestSensitivityInitialCondition`?

pymc/ode/utils.py
```diff
@@ -103,10 +103,21 @@ def augment_system(ode_func, n_states, n_theta):
...
-    # Stack the results of the ode_func into a single tensor variable
-    if not isinstance(yhat, (list, tuple)):
-        yhat = (yhat,)
-    t_yhat = at.stack(yhat, axis=0)
+    if isinstance(yhat, at.TensorVariable):
+        if yhat.ndim == 0:
+            # Convert yhat from scalar to array
+            t_yhat = at.stack((yhat,), axis=0)
+        else:
+            t_yhat = yhat
```
michaelosthege:
```suggestion
        elif yhat.ndim == 2:
            t_yhat = yhat
        else:
            raise ValueError(f"The odefunc returned a {yhat.ndim}-dimensional tensor, but 0 or 1 dimensions were expected.")
```

michaelosthege (commented):
Looks good. I'd just like to cover this additional edge case with an informative error.

michaelosthege:
FYI: I fixed the pre-commit.
If all tests go ✔ this should be good to go.

pymc/ode/utils.py
```diff
@@ -103,10 +103,25 @@ def augment_system(ode_func, n_states, n_theta):
...
+        if yhat.ndim == 0:
+            # Convert yhat from scalar to array
+            t_yhat = at.stack((yhat,), axis=0)
+        elif yhat.ndim == 1:
+            t_yhat = yhat
+        else:
+            raise ValueError(
+                f"The odefunc returned a {yhat.ndim}-dimensional tensor, but 0 or 1 dimensions were expected."
+            )
+    elif isinstance(yhat, np.ndarray):
```
ricardoV94: This check is a bit arbitrary, no? Was the original issue related to this?

Otherwise why aren't we checking if the tuple/list are not composed of ndarrays as well?

michaelosthege: On a second though, yes.
Maybe simply doing `at.atleast_1d(yhat)` and then checking `ndim==1`?
Do we even need to anticipate the `ndarray` case?

mirko-m: I think it is possible that `ode_func` returns an `ndarray` . This is for example what was tried here https://discourse.pymc.io/t/return-value-for-2n-dimensional-ode-system/4141/3 . It looks like in this case `np.concatenate` threw an error, but maybe there are other operations which will cast a `TensorVariable` to an `ndarray` without throwing an error.

ricardoV94: That's not possible

mirko-m: I removed the check for `np.ndarray` and I made the code more concise by using `at.atleast_1d` as suggested. Please let me know if there is anything else that needs to be done before this PR is good to go.

michaelosthege: Thanks! We're working to fix the CI on main and once that's done we can rebase your branch so we can have a green CI before merging

pymc/ode/utils.py
```diff
@@ -103,10 +103,17 @@ def augment_system(ode_func, n_states, n_theta):
 
     # Get symbolic representation of the ODEs by passing tensors for y, t and theta
     yhat = ode_func(t_y, t_t, t_p[n_states:])
-    # Stack the results of the ode_func into a single tensor variable
-    if not isinstance(yhat, (list, tuple)):
-        yhat = (yhat,)
-    t_yhat = at.stack(yhat, axis=0)
+    if isinstance(yhat, at.TensorVariable):
+        t_yhat = at.atleast_1d(yhat)
+        if t_yhat.ndim > 1:
```
ricardoV94: Does this check also make sense in the `else` branch? If so, you could move it below the `if/else` statements.

ricardoV94: Also, can you add a test for it?

mirko-m: I moved the check so that it also applies to the else branch.

I also created two tests which check that the `ValueError` is thrown.

pymc/tests/test_ode.py
```diff
@@ -267,6 +267,22 @@ def test_number_of_params(self):
...
+    def test_tensor_shape(self):
+        with pytest.raises(ValueError):
+            DifferentialEquation(func=self.system_2d_tensor, t0=0, times=self.times, n_states=1, n_theta=0)
+
+    def system_2d_list(y, t, p):
+        s = np.exp(-t) - p[0] * y[0]
+        return [[s,s], [s,s]]
+
+    def test_list_shape(self):
+        with pytest.raises(ValueError):
```
michaelosthege:
```suggestion
        with pytest.raises(ValueError, match="returned a 2-dimensional tensor"):
```

pymc/tests/test_ode.py
```diff
@@ -267,6 +267,22 @@ def test_number_of_params(self):
         with pytest.raises(ValueError):
             DifferentialEquation(func=self.system, t0=0, times=self.times, n_states=1, n_theta=0)
 
+    def system_2d_tensor(y, t, p):
+        s = np.exp(-t) - p[0] * y[0]
+        return at.stack((s, s, s, s)).reshape((2,2))
+
+    def test_tensor_shape(self):
+        with pytest.raises(ValueError):
```
michaelosthege:
```suggestion
        with pytest.raises(ValueError, match="returned a 2-dimensional tensor"):
```

michaelosthege (commented):
Let's see what the CI tests say 👍

mirko-m:
After accepting the suggested changes it looks to me like the tests that I wrote are catching the wrong ValueError. I will fix that.

mirko-m:
I added two more commits. The first commit fixes the new tests that I had written to check that when `yhat` is 2-dimensional a `ValueError` is thrown (`TestErrors.test_tensor_shape` and `TestErrors.test_list_shape`).

The second commit adds the `match` keyword from `pytest.raises` to the other tests inside the `TestErrors` class. I think that this is useful, but I am not sure if you want to merge it as part of this pull request or a separate pull request, because perhaps it it off topic. That is why I created a second commit for this change, so that you can easily ignore it.

When fixing the new tests (`TestErrors.test_tensor_shape` and `TestErrors.test_list_shape`) I noticed that `TestErrors.system`, `TestErrors.system_2d_tensor` and `TestErrors.system_2d_list` are not really instance methods of the class and therefore cannot be called as e.g. `self.system` . Instead in order to get the new tests to work I had to make the calls as `TestErrors.system_2d_tensor` and `TestErrors.system_2d_list`, respectively. For the other tests that are part of the `TestErrors` class this does not matter so much, because the errors that `pytest` expects are raised before `TestErrors.system` is called, but I changed the calls from `self.system` to `TestErrors.system` anyways because I think that this the correct way of calling it. Let me know if you disagree or if there is a better way of calling these methods/functions.

pymc/tests/test_ode.py
```diff
@@ -218,35 +237,57 @@ def system(y, t, p):
...
+    def system_2d_list(y, t, p):
+        s0 = np.exp(-t) - p[0] * y[0]
+        s1 = np.exp(-t) - p[0] * y[1]
+        s2 = np.exp(-t) - p[0] * y[2]
+        s3 = np.exp(-t) - p[0] * y[3]
+        return [[s0,s1], [s2,s3]]
+
+    def test_list_shape(self):
+        with pytest.raises(ValueError, match="returned a 2-dimensional tensor"):
+            DifferentialEquation(func=TestErrors.system_2d_list, t0=0, times=self.times, n_states=4, n_theta=1)
```
michaelosthege: If the `system_*` functions are just used inside one test each, better define them inside the test methods.

Otherwise decorate them with `@staticmethod` and continue to reference them like `TestErrors.system_*`

mirko-m: The `system_2d_*` functions are only used for a single test, so I moved the definition inside those tests.

The function called `system` is used in multiple tests and I therefore decorated it with `@staticmethod`. This means that on line 238 I need to change the call to `func=system.__func__` because `system` is not callable. If instead I try to use `func=TestErrors.system` I get a `NameError`, because `TestErrors` is not defined.

pymc/tests/test_ode.py
```diff
@@ -209,44 +228,85 @@ def system_1(y, t, p):
...
     """Test running model for a scalar ODE with 1 parameter"""
 
+    @staticmethod
     def system(y, t, p):
         return np.exp(-t) - p[0] * y[0]
 
     times = np.arange(0, 9)
 
-    ode_model = DifferentialEquation(func=system, t0=0, times=times, n_states=1, n_theta=1)
+    ode_model = DifferentialEquation(func=system.__func__, t0=0, times=times, n_states=1, n_theta=1)
```
michaelosthege: This would be typical for a `setup_method`:

```python
def setup_method(self, method):
    def system(y, t, p):
        return np.exp(-t) - p[0] * y[0]
    self.times = np.arange(0, 9)
    self.ode_model = DifferentialEquation(func=system, t0=0, times=self.times, n_states=1, n_theta=1)
```
Also see https://docs.pytest.org/en/latest/how-to/xunit_setup.html

michaelosthege (commented):
Hi @mirko-m,
sorry that this is taking so long.
I think you'll need to rebase onto the main branch, otherwise the CI pipeline will fail, because of installation dependencies.
Let me know if you need help with it

pymc/tests/test_ode.py
mirko-m: I implemented the setup method and I will try to rebase onto main next.

mirko-m: Looks like the rebase fixed the CI pipeline. Let me know what you think.

pymc/ode/utils.py
```diff
@@ -103,10 +103,17 @@ def augment_system(ode_func, n_states, n_theta):
...
-    # Stack the results of the ode_func into a single tensor variable
-    if not isinstance(yhat, (list, tuple)):
-        yhat = (yhat,)
-    t_yhat = at.stack(yhat, axis=0)
+    if isinstance(yhat, at.TensorVariable):
+        t_yhat = at.atleast_1d(yhat)
+    else:
+        # Stack the results of the ode_func into a single tensor variable
+        if not isinstance(yhat, (list, tuple)):
+            yhat = (yhat,)
```
michaelosthege: The CI says this is not covered by the tests?
I think this line could become an exception: Unexpected type returned by the ODE function.
And you could test it by returning a dictionary or set of ODEs instead of a tensor/list/tuple.

FYI you can generate the coverage report locally by passing `--cov=pymc --cov-report=term-missing`. Needs "pip install pytest-cov" and can be combined with running just a specific test file.

I'll approve the CI run as soon as I notice a commit. Let's get this merged on Saturday, because I want to do another typing related PR this weekend. I can also take over from here if you want to

mirko-m: I was away over the weekend, but I made the changes that you suggested just now. Hopefully the CI still works.

michaelosthege (approved):

michaelosthege:
Thanks, @mirko-m !

mirko-m:
Thanks for guiding me through the process and all your suggestions.

PR #5417, opened 2022-01-28
Fix convert_size failure with 1d symbolic vectors

### LINKED ISSUE #5394: `convert_size` wrongly assumes symbolic sizes have to be scalars
```python
import pymc as pm
import aesara.tensor as at

s = at.scalar('s')
size = at.stack([s, s])
x = at.random.normal(size=size)
x.eval({s: 2})
# array([[-0.66285345,  0.97341598],
#        [ 0.05158132,  2.03399059]])

pm.Normal.dist(size=size)
# ValueError: The `size` parameter must be a tuple, TensorVariable, int or list. Actual: <class 'aesara.tensor.var.TensorVariable'>
```

The error comes from this check here:
https://github.com/pymc-devs/pymc/blob/915592235ea6920689f89bece48e2d679e8ed4f1/pymc/distributions/shape_utils.py#L475

CC @michaelosthege

5hv5hvnk:
```convert_size```  wrongly assumes symbolic sizes have to be scalars
```
import pymc as pm
import aesara.tensor as at

s = at.scalar('s')
size = at.stack([s, s])
x = at.random.normal(size=size)
x.eval({s: 2})
# array([[-0.66285345,  0.97341598],
#        [ 0.05158132,  2.03399059]])

pm.Normal.dist(size=size)
# ValueError: The `size` parameter must be a tuple, TensorVariable, int or list. Actual: <class 'aesara.tensor.var.TensorVariable'>
```

fixed by changing code in 
https://github.com/pymc-devs/pymc/blob/915592235ea6920689f89bece48e2d679e8ed4f1/pymc/distributions/shape_utils.py#L475

by accepting ```size.ndim == 1``` as well as ```size.ndim == 0```

Closes #5394

### DIFF
--- a/pymc/distributions/shape_utils.py
+++ b/pymc/distributions/shape_utils.py
@@ -450,8 +450,10 @@ def convert_shape(shape: Shape) -> Optional[WeakShape]:
     if shape is None:
         return None
 
-    if isinstance(shape, int) or (isinstance(shape, TensorVariable) and shape.ndim == 0):
-        shape = (shape,)
+    if isinstance(shape, int):
+        shape = shape
+    if isinstance(shape, TensorVariable) and shape.ndim <= 0:
+        shape = tuple(shape)
     elif isinstance(shape, (list, tuple)):
         shape = tuple(shape)
     else:
@@ -473,7 +475,7 @@ def convert_size(size: Size) -> Optional[StrongSize]:
         return None
     if isinstance(size, int):
         size = size
-    elif isinstance(size,TensorVariable):
+    elif isinstance(size, TensorVariable) and size.ndim <= 1:
         size = tuple(size)
     elif isinstance(size, (list, tuple)):
         size = tuple(size)
--- a/pymc/distributions/shape_utils.py
+++ b/pymc/distributions/shape_utils.py
@@ -451,7 +451,7 @@ def convert_shape(shape: Shape) -> Optional[WeakShape]:
         return None
     if isinstance(shape, int) or (isinstance(size, TensorVariable) and shape.ndim == 0):
         shape = (shape,)
-    if isinstance(shape, TensorVariable) and shape.ndim == 1:
+    elif isinstance(shape, TensorVariable) and shape.ndim == 1:
         shape = tuple(shape)
     elif isinstance(shape, (list, tuple)):
         shape = tuple(shape)
@@ -459,12 +459,10 @@ def convert_shape(shape: Shape) -> Optional[WeakShape]:
         raise ValueError(
             f"The `shape` parameter must be a tuple, TensorVariable, int or list. Actual: {type(shape)}"
         )
-
     if isinstance(shape, tuple) and any(s == Ellipsis for s in shape[:-1]):
         raise ValueError(
             f"Ellipsis in `shape` may only appear in the last position. Actual: {shape}"
         )
-
     return shape
 
 
@@ -472,7 +470,7 @@ def convert_size(size: Size) -> Optional[StrongSize]:
     """Process a user-provided size variable into None or a valid size object."""
     if size is None:
         return None
-    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
+    elif isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
         size = (size,)
     elif isinstance(size, TensorVariable) and size.ndim == 1:
         size = tuple(size)
--- a/pymc/distributions/shape_utils.py
+++ b/pymc/distributions/shape_utils.py
@@ -472,8 +472,8 @@ def convert_size(size: Size) -> Optional[StrongSize]:
     if size is None:
         return None
 
-    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
-        size = (size,)
+    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim >= 1):
+        size = tuple(size,)
     elif isinstance(size, (list, tuple)):
         size = tuple(size)
     else:
--- a/pymc/distributions/shape_utils.py
+++ b/pymc/distributions/shape_utils.py
@@ -449,10 +449,9 @@ def convert_shape(shape: Shape) -> Optional[WeakShape]:
     """Process a user-provided shape variable into None or a valid shape object."""
     if shape is None:
         return None
-
-    if isinstance(shape, int):
-        shape = shape
-    if isinstance(shape, TensorVariable) and shape.ndim <= 0:
+    if isinstance(shape, int) or (isinstance(size, TensorVariable) and shape.ndim == 0):
+        shape = (shape,)
+    if isinstance(shape, TensorVariable) and shape.ndim == 1:
         shape = tuple(shape)
     elif isinstance(shape, (list, tuple)):
         shape = tuple(shape)
@@ -473,9 +472,9 @@ def convert_size(size: Size) -> Optional[StrongSize]:
     """Process a user-provided size variable into None or a valid size object."""
     if size is None:
         return None
-    if isinstance(size, int):
-        size = size
-    elif isinstance(size, TensorVariable) and size.ndim <= 1:
+    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
+        size = (size,)
+    elif isinstance(size, TensorVariable) and size.ndim == 1:
         size = tuple(size)
     elif isinstance(size, (list, tuple)):
         size = tuple(size)
--- a/pymc/distributions/shape_utils.py
+++ b/pymc/distributions/shape_utils.py
@@ -471,19 +471,18 @@ def convert_size(size: Size) -> Optional[StrongSize]:
     """Process a user-provided size variable into None or a valid size object."""
     if size is None:
         return None
-
-    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim==1 or size.ndim==0):
-        size = tuple(size,)
+    if isinstance(size, int):
+        size = size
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by michaelosthege — CHANGES_REQUESTED** [review]
Please add unit tests that can detect the original issue, ideally testing the function with all different kinds of inputs that should be supported according to its signature.
Also you should edit the description of the PR to explain what this is about: what was the problem with the original code and how do your changes fix it?

File: pymc/distributions/shape_utils.py [hunk 1]
```diff
@@ -472,8 +472,8 @@ def convert_size(size: Size) -> Optional[StrongSize]:
     if size is None:
         return None
 
-    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
-        size = (size,)
+    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim >= 1):
```
ricardoV94: I think size can never have `ndim >= 1`, so this check doesn't make sense. We need to make sure logic works for the two possible cases (ndim=0 and ndim=1)

ricardoV94 (changes requested):
Thanks for opening the PR. I've left a comment above

pymc/distributions/shape_utils.py
5hv5hvnk: @ricardoV94 made a new edit making sure it only accepts ndim=0 and ndim=1 please have a look

5hv5hvnk:
@ricardoV94  I made a new edit making sure it only accepts ```ndim=0 and ndim=1``` please have a look

ricardoV94:
@5hv5hvnk We need tests showing that this is no longer an issue

5hv5hvnk:
@ricardoV94 how do I make test cases and show that they are working? As of now when I tested few different things on the code that was in the initial issue itself and they work fine.

ricardoV94:
Hi @5hv5hvnk, apologies for the delay.

You can add a new test case in this function: https://github.com/pymc-devs/pymc/blob/915592235ea6920689f89bece48e2d679e8ed4f1/pymc/tests/test_shape_handling.py#L463-L468

Something like this might be enough:
```python
x = at.constant(5)
assert convert_size(x) is size

x = at.constant([5, 5])
assert convert_size(x) is size
```

pymc/distributions/shape_utils.py
```diff
@@ -472,8 +472,8 @@ def convert_size(size: Size) -> Optional[StrongSize]:
     if size is None:
         return None
 
-    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
-        size = (size,)
+    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim==1 or size.ndim==0):
+        size = tuple(size,)
     elif isinstance(size, (list, tuple)):
         size = tuple(size)
     else:
```
ricardoV94: Does this work? I don't think we have to do anything if it is already a TensorVariable. We could assert it has `ndim < 2`. but that's about it.

```suggestion
    if isinstance(size, int):
        size = tuple(size,)
    elif isinstance(size, (list, tuple)):
        size = tuple(size)
    elif not isinstance(TensorVariable):
```

ricardoV94 (changes requested):

ricardoV94:
@5hv5hvnk This error also seems to be present in `convert_shape`: https://github.com/pymc-devs/pymc/blob/90948a3fe3f4417f2f0cda2d916a59302d50c60b/pymc/distributions/shape_utils.py#L475-L476

A similar fix should work there

pymc/distributions/shape_utils.py
```diff
@@ -471,19 +471,18 @@ def convert_size(size: Size) -> Optional[StrongSize]:
     """Process a user-provided size variable into None or a valid size object."""
     if size is None:
         return None
-
-    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
-        size = (size,)
+    if isinstance(size, int):
+        size = size
+    elif isinstance(size,TensorVariable):
```
michaelosthege: Please check for the ndim too (like a few lines above). For ndim>1 we still want to run into the exception case.

michaelosthege (changes requested):
Please run the pre-commit locally before pushing l:
`pip install pre-commit`

`pre-commit run --all`

pymc/distributions/shape_utils.py
```diff
@@ -471,19 +473,18 @@ def convert_size(size: Size) -> Optional[StrongSize]:
     """Process a user-provided size variable into None or a valid size object."""
     if size is None:
         return None
-
-    if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
-        size = (size,)
+    if isinstance(size, int):
+        size = size
+    elif isinstance(size, TensorVariable) and size.ndim <= 1:
+        size = tuple(size)
```
michaelosthege: `tuple(size)` does't work when `size.ndim == 0`.
The 0 and 1 ndim must be separate cases

ricardoV94: does `to_tuple` work?

5hv5hvnk: something like this would be sufficient ?
```
if isinstance(shape, int) or (isinstance(size,TensorVariable) and shape.ndim==0):
    shape = shape
if isinstance(shape, TensorVariable) and shape.ndim == 1:
    shape = tuple(shape)
```

michaelosthege: Just keep the code that was there before and add another elif case:
```
if isinstance(size, int) or (isinstance(size, TensorVariable) and size.ndim == 0):
    size = (size,)
elif isinstance(size, TensorVariable) and size.ndim == 1:
    size = tuple(size)
```
And the same further up in `convert_shape`

5hv5hvnk: on using 
```
pre-commit run --all
```
all the tests are passed, however it shows build failed when I commit the changes how can I resolve this?

pymc/distributions/shape_utils.py
```diff
@@ -449,9 +449,10 @@ def convert_shape(shape: Shape) -> Optional[WeakShape]:
     """Process a user-provided shape variable into None or a valid shape object."""
     if shape is None:
         return None
-
-    if isinstance(shape, int) or (isinstance(shape, TensorVariable) and shape.ndim == 0):
+    if isinstance(shape, int) or (isinstance(size, TensorVariable) and shape.ndim == 0):
```
canyon289: I would add parentheses around the or statement conditionsm its not strictly needed but it makes the code intent  much more obvious

michaelosthege: I don't mind the parantheses, but in the 2nd isinstance you changed shape to size.
This line should have no change at all

pymc/distributions/shape_utils.py
```diff
@@ -449,9 +449,10 @@ def convert_shape(shape: Shape) -> Optional[WeakShape]:
...
         shape = (shape,)
+    if isinstance(shape, TensorVariable) and shape.ndim == 1:
```
michaelosthege: This should be an `elif`

michaelosthege (changes requested):
I canceled the CI workflows because the pre-commit was failing.
Please run pre-commit before pushing to reduce the CI workload.

canyon289:
> I canceled the CI workflows because the pre-commit was failing. Please run pre-commit before pushing to reduce the CI workload.

Noted, and thank you

pymc/distributions/shape_utils.py
canyon289: Precommit is not running tests, but other things like linters and code format checks. The output should tell you what the fix is, or automatically add the changes to git staging. Feel free to post a screenshot of the output and we can more specifically let you know what to do!

5hv5hvnk: image
this shows pre commit works fine, yet the commit fails to build.
How can I fix this?

canyon289: Im not sure what you mean by build. This indicates the commit should have taken place. What happens when you try and push it to github?

michaelosthege: I think your branch could be outdated. The [pre-commit in the CI failed in the type checking step](https://github.com/pymc-devs/pymc/runs/5528156814?check_suite_focus=true#step:4:64) but that step did not run in your screenshot.

Do the following to update your branch:
1. Make sure that you don't have uncommitted changes (commit, stash or revert them if needed)
2. `git checkout main`
3. `git pull`
4. `git checkout my-feature` to go back to the branch of this PR
5. `git rebase main` to rebase the PR branch onto main. Since GitHub doesn't show any merge conflicts this step should work without problems.

Then run `pre-commit` again. This time it should reproduce the failure from the CI.

michaelosthege:
The diff looks good.
I triggered the CI tests, but I'm afraid they might fail because the branch is outdated (similar to why the docs build failed).

5hv5hvnk:
> The diff looks good.
> I triggered the CI tests, but I'm afraid they might fail because the branch is outdated (similar to why the docs build failed).
Do I have to clone the repo again and do changes there in order to update the branch or is there something else I can do?

twiecki:
@5hv5hvnk You need to pull main and then rebase your branch and force-push: https://stackoverflow.com/questions/8965781/update-an-outdated-branch-against-master-in-a-git-repo

5hv5hvnk:
> @5hv5hvnk You need to pull main and then rebase your branch and force-push: https://stackoverflow.com/questions/8965781/update-an-outdated-branch-against-master-in-a-git-repo

image

This shows my branch is up to the date I still used push --force to push it to the main branch

twiecki:
@5hv5hvnk This is your fork, you can add the pymc-devs repo as a remote and then pull to main from there and then rebase.

Issue #5216, opened 2021-11-21
allow `OrderedProbit` and `OrderedLogit` to take vector inputs


drbenvincent:
At the moment, as far as I can tell, `OrderedProbit` and `OrderedLogit` break if you try to provide vector inputs for `eta` and `sigma` parameters.

For example, if you have data like this
<img width="478" alt="Screenshot 2021-11-21 at 13 21 51" src="https://user-images.githubusercontent.com/6765047/142763568-3dc35b32-39ed-4bb6-9387-0a13375e003e.png">
where your data is encoded as a vector of `y` values alongside a vector of `grp_idx`, then it looks like you cannot do this...

## Approach 1 - better but doesn't work
```python
with pm.Model() as model1:
    theta = constrainedUniform(K)
    mu = pm.Normal("mu", mu=K/2, sigma=K, shape=2)
    sigma = pm.HalfNormal("sigma", 1, shape=2)
    pm.OrderedProbit("y_obs", 
                     cutpoints=theta, 
                     eta=mu[grp_idx], 
                     sigma=sigma[grp_idx], 
                     observed=y)
```
We have a common set of cutpoints `theta`, but each of the groups has different `eta` and `sigma`. We try to pass in a vector of `eta` and `sigma` values using advanced indexing. But this does not work 🙁

## Approach 2 - split into separate likelihood terms
If we cannot pass in vectors to `eta` or `sigma`, then the solution that works is to encode the data differently so that you now have observations from one group as `y1` and observations from another group as `y2`

```python
with pm.Model() as model2:
    theta = constrainedUniform(K)
    mu = pm.Normal("mu", mu=K/2, sigma=K, shape=2)
    sigma = pm.HalfNormal("sigma", 1, shape=2)
    pm.OrderedProbit("y1_obs", 
                     cutpoints=theta, 
                     eta=mu[0], 
                     sigma=sigma[0], 
                     observed=y1)
    pm.OrderedProbit("y2_obs", 
                     cutpoints=theta, 
                     eta=mu[1], 
                     sigma=sigma[1], 
                     observed=y2)
```
This approach _does_ work, but makes model specification cumbersome, and doesn't scale well with more groups.

Not crucial for this issue, but the `constrainedUniform` is a proposed new distribution, outlined in pymc-devs/pymc-experimental#32.

danhphan:
[...]

PR #5418, opened 2022-01-29
Allow OrderedProbit distribution to take vector inputs

danhphan:
This PR allows the `OrderedProbit` distribution to take vector inputs using advanced indexing in #5216

- Fix the shape of the `sigma` parameter
- Add test_vector_inputs for `OrderedProbit` in test_distributions_random.py

Hi @ricardoV94, let me know if it needs to update. Thanks


ricardoV94 (commented):
We can probably simplify this test quite a lot. We don't need to actually create a whole dataset with pandas, just pass vector inputs and then check that the categorical variable that is returned has the right inputs (shape and values)

```python
categorical = pm.OrderedProbit.dist(
  cutpoints=np.array(...)
  eta=np.array(...),
  sigma=np.array(...),
)

p = categorical.owner.inputs[3].eval()
# Assert p is what we expected (shape and values)
```

We can also go one step further and check if it is also working with matrix inputs, or combinations of scalar and vector inputs

danhphan:
Hi @ricardoV94 , somethings like this?

```
def test_vector_inputs(self):
    """ 
    This test checks when providing vector inputs for `eta` and `sigma` parameters using advanced indexing.
    """
    categorical = pm.OrderedProbit.dist(
        eta=0,
        cutpoints=np.array([-2.0, 0, 2.0]),
        sigma=1.0,
        )
    p = categorical.owner.inputs[3].eval()
    assert p.shape == (4,)

    categorical = pm.OrderedProbit.dist(
        eta=np.array([1.0, 2.0, 3.0, 4.0, 5.0]),
        cutpoints=np.array([-2.0, 0, 2.0]),
        sigma=1,
        )
    p = categorical.owner.inputs[3].eval()
    assert p.shape == (5, 4)

    categorical = pm.OrderedProbit.dist(
        eta=np.array([1.0, 2.0, 3.0, 4.0, 5.0]),
        cutpoints=np.array([-2.0, 0, 2.0]),
        sigma=np.array([1.0, 2.0, 3.0, 4.0, 5.0]),
        )
    p = categorical.owner.inputs[3].eval()
    assert p.shape == (5, 4)
```
Besides, how do we know that `.owner.inputs[3]` will be `p`?

ricardoV94:
> Besides, how do we know that .owner.inputs[3] will be p?

That's just the position of the `p` input in the returned `CategoricalRV` variable. It's always the same position.

I think the new tests are great. I just wonder what happens when the input has more dimensions (such as 2D cutpoints, or 2D eta/sigma). Do we get sensible `p`? Or does it just fail altogether?

If it works it would be nice to test it as well. Otherwise we can:
1. Try to make it work
2. Add an explicit check in the `dist` with an informative `NotImplementedError` for invalid input dimensions

ricardoV94:
By the way, I like to use `np.vectorize` when testing that a function is broadcasting as expected. In your case something like this:

```python
# Create vectorized function from base case (scalar eta, scalar sigma, and vector cutpoints)
eta = at.scalar('eta')
sigma = at.scalar('sigma')
cutpoints = at.vector('cutpoints')

probits = eta - cutpoints
left = normal_lccdf(0, sigma, probits[0])
middle = log_diff_normal_cdf(0, sigma, probits[:-1], probits[1:])
right = normal_lcdf(0, sigma, probits[-1])
p = at.exp(at.concatenate([[left], middle, [right]]))

base_p = aesara.function([eta, sigma, cutpoints], ret)
vec_p = np.vectorize(base_p, signature="(),(),(n)->(m)")
```

You can ten use this `vec_p` to test arbitrary (valid) shapes of inputs.
```python
assert vec_p(eta=0, sigma=1, cutpoints=[-2, 0, 2]) == (4,)
assert vec_p(eta=np.zeros((5, 2)), sigma=np.ones((2, 5, 2)), cutpoints=[[-2, 0, 2], [-2, 0, 2]]).shape == (2, 5, 2, 4)
```

Of couse, even better, we can test not only the shapes but that all the values are close to the expected, with `numpy.testing.assert_array_almost_equal`

danhphan:
Hi yes, let's also test 2D case. 

I will prefer to use `OrderedProbit.dist` as this is the goal of the test function. Besides, it may be better if the test is not too complicated :), we also seem not test other distribtions this way (np.vectorize)?

I think we should not repeat ourself by rewrite the following code.
```
probits = eta - cutpoints
left = normal_lccdf(0, sigma, probits[0])
middle = log_diff_normal_cdf(0, sigma, probits[:-1], probits[1:])
right = normal_lcdf(0, sigma, probits[-1])
p = at.exp(at.concatenate([[left], middle, [right]]))
```

Thanks mate 👍

ricardoV94:
@danhphan the idea was not to skip using `OrderedProbit.dist`, but call it and compare the values of the `p` obtained from `p = categorical.owner.inputs[3].eval()` with those from the `np.vectorize`d function. 

This would not only test the shape of `p` (which you were already testing by hand) but also that the values of `p` are what is expected (e.g., that there was no weird mixing of the axis).

For precedence, here is a test that uses a `np.vectorize`d function for purposes of testing a logp:

https://github.com/pymc-devs/pymc/blob/1a35a3dfc0f125dfa1515e3c8d3575c9f17a3074/pymc/tests/test_distributions.py#L471-L476

https://github.com/pymc-devs/pymc/blob/1a35a3dfc0f125dfa1515e3c8d3575c9f17a3074/pymc/tests/test_distributions.py#L2169

Edit: I can be convinced that it is an overkill, and testing shapes is enough because there is no other way the values could have broadcasted. Just wanted to clarify what I was suggesting :)

ricardoV94:
By the way, do you want to address the same limitation of `OrderedLogit` in this PR?

danhphan:
Hi @ricardoV94, thanks for the information.

On the `_dirichlet_logpdf(value, a)` function, what I guess is that it is used to check if a pymc's function can produce similar results as in `scipy`. But should we also need to write a test for `_dirichlet_logpdf(value, a)` itself to make sure it work as expected?

FYI, I tested the vector inputs for `OrderedLogit` and it works fine. Since the error stems from the shape of `sigma` parameter which only `_OrderedProbit` has, but not in `OrderedLogit`. So we do not fix anything in `OrderedLogit` distribution.

To be safe, I will add a test to check different shapes of `cutpoints` and `eta` parameters for `OrderedLogit`. 

Cheers.

ricardoV94:
> But should we also need to write a test for _dirichlet_logpdf(value, a) itself to make sure it work as expected?

I see what you mean. We have tests before were we check our pymc distribution (what we care about) matches scipy reference directly in the basic situation (no vectorization).

Given this, we can then use `_dirichlet_logpdf` just to test that vectorization is working as expected. It would be very unlikely that our distribution would both match the scipy in the basic case and the vectorized helper if there was a bug in the latter.

Similarly here, you can see that we have a `check_rv_params_match_pymc_params` test where we check that one specific set of `cutpoints`, `sigma`, and `eta`  get converted to specific expected `p`. It would be unlikely that both this check and the vectorized one would work if there was a bug in the latter.

In a sense we don't need to test out helpers because of this type of test triangulation.

***

> To be safe, I will add a test to check different shapes of cutpoints and eta parameters for OrderedLogit.

Definitely useful!

danhphan:
Hi @ricardoV94, I have simplify the `test_shape_inputs` to check scalar, vector, and matrix inputs for `_OrderedProbit`. Also added a similar test for `_OrderedLogistic`. 

Let's me know if they need to update anythings.

Thanks mate :+1:

twiecki (approved):

pymc/tests/test_distributions_random.py
```diff
@@ -1688,6 +1688,26 @@ class TestOrderedLogistic(BaseTestDistributionRandom):
...
+    @pytest.mark.parametrize(
+        "eta, cutpoints, expected",
+        [
+            (0, [-2.0, 0, 2.0], (4,)),
+            ([-1], [-2.0, 0, 2.0], (1, 4)),
+            ([1.0, -2.0], [-1.0, 0, 1.0], (2, 4)),
+            ([[1.0, -1.0, 0.0], [-1.0, 3.0, 5.0]], [-2.0, 0, 1.0], (2, 3, 4)),
+        ],
+    )
+    def test_shape_inputs(self, eta, cutpoints, expected):
```
ricardoV94, line 1700: We should also test 2d cutpoints

ricardoV94, line 1702: We should also test 2d cutpoints

pymc/tests/test_distributions_random.py
```diff
@@ -1698,6 +1718,34 @@ class TestOrderedProbit(BaseTestDistributionRandom):
...
+        [
+            (0, [-2.0, 0, 2.0], 1.0, (4,)),
+            ([-1], [-2.0, 0, 2.0], [2.0], (1, 4)),
+            ([1.0, -2.0], [-1.0, 0, 1.0], 1.0, (2, 4)),
+            ([1.0, -2.0, 3.0], [-2.0, 0, 2.0], [-1.0, -2.0, 5.0], (3, 4)),
+            ([[1.0, -1.0, 0.0], [-1.0, 3.0, 5.0]], [-2.0, 0, 1.0], [-1.0, -2.0, 5.0], (2, 3, 4)),
+            (
+                [[1.0, -2.0, 3.0], [1.0, 2.0, -4.0]],
+                [-2.0, 0, 1.0],
+                [[0.0, 2.0, -4.0], [-1.0, 1.0, 3.0]],
```
ricardoV94: A lot of these sigma are negative. We should test with valid sigma values (> 0)

ricardoV94 (changes requested):
Looks great, just some small suggestions

pymc/tests/test_distributions_random.py
ricardoV94: Also test with 2d cutpoints missing

danhphan: Hi @ricardoV94, I will change the `sigma` to all positive. 

Also, `cutpoints` should be always 1 dimension (to my understanding) as it represents (n-1) cut points of a categorical feature with n categories. I am not sure if there is any cases that needs 2d `cutpoints`.

ricardoV94: Our distributions, when possible, can always be "batched". That means we can arbitrarily increase the dimensionality of the distribution by adding parameters with more dimensions. The last axes represent the parameters for each "atomic" distribution in the batch

For example the Categorical distribution is happy to take 2D, 3D, ... ND dimensional probability parameters, as long as they add up to 1 over the last axis

```python
pm.Categorical.dist(
  np.full((4, 2, 3), [
    [0., .1, .9],
    [.9, .1, 0]
  ])
).eval()
```

The same should apply here if possible. See this issue where we are pursuing this for all multivariate distributions: https://github.com/pymc-devs/pymc/issues/5383

The reason why this is useful is vectorization. Specifying a (3, 3) shaped distribution with different cutpoints can be much more efficient than specifying 3 times a (3,) shaped distribution with the different cutpoints.

ricardoV94: This was exactly how this issue started by the way, just with the batching across sigma and eta, and fixed cutpoints. But it could have been the other way around.

danhphan: Hi yes, the `batch` dimension totally make sense. My initial thought was that for dealing with a large data set, `batch_size` should be managed in `pm.Data` (similar to DataLoader in pytorch). Although I have not checked `pm.Data` yet :)

Anyways, I will check the case of 2d cutpoints as well.

ricardoV94: Maybe now my crazy example from before makes more sense?

```python
OrderedProbit.dist(
  eta=np.zeros((5, 2)), 
  sigma=np.ones((2, 5, 2)), 
  cutpoints=[[-2, 0, 2], [-2, 0, 2]]
)  #shape should be (2, 5, 2, 4)
```

danhphan: Hi, it makes sense. I have added the tests for 2d `cutpoints` and positve `sigma`.

Already run `push`, but not sure why it has not updated in this PR:
https://github.com/danhphan/pymc/commit/9b311bf8b4d8fae93db2a085fb5be1e8c730c570

ricardoV94 (approved):
Looks great! Thanks for your help @danhphan 

Looking forward to your next PR :D

danhphan:
Hi @ricardoV94 many thanks for your support :100: 

Cheer! :wine_glass: :wine_glass:

twiecki:
Congrats on number 2 @danhphan!

PR #5420, opened 2022-01-29
Cleanup trace initialization and type hints

michaelosthege:
This PR refactors the code and type hints in `sampling.py` that are related to trace backend initialization and related type hints.

I found several type hints that were too loose or even wrong.
* In multiple places `tune` was advertised as `int | None` but the code inside assumed it to be `int`.
* Similarly, multiple `trace` kwargs were advertised as `BaseTrace | MultiTrace` but the code inside called methods that are only available with `BaseTrace`.
* After tightening the input types, I fixed the return type hints in a similar fashion.

Finally, this tightening enabled a consolidation of the trace initialization (`_choose_backend()` + `strace.setup()`) such that now all of `_iter_sample`, `_mp_sample`, `_prepare_iter_population` make the same call to `_init_trace`.

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -32,6 +32,7 @@ from aesara.tensor.elemwise import Elemwise
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.random.var import RandomStateSharedVariable
 from aesara.tensor.var import TensorVariable
+from typing_extensions import TypeAlias
 
 from pymc.aesaraf import change_rv_size
 from pymc.distributions.shape_utils import (
@@ -61,7 +62,7 @@ __all__ = [
     "NoDistribution",
 ]
 
-DIST_PARAMETER_TYPES = Union[np.ndarray, int, float, TensorVariable]
+DIST_PARAMETER_TYPES: TypeAlias = Union[np.ndarray, int, float, TensorVariable]
 
 vectorized_ppc = contextvars.ContextVar(
     "vectorized_ppc", default=None
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -46,6 +46,7 @@ from aesara.graph.basic import Constant, Variable
 from aesara.tensor.sharedvar import SharedVariable
 from arviz import InferenceData
 from fastprogress.fastprogress import progress_bar
+from typing_extensions import TypeAlias
 
 import pymc as pm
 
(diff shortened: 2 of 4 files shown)


michaelosthege:
The changes reduce the number of errors found by `mypy -p pymc` from 362 to 266.

https://www.diffchecker.com/aHofAXhG

pymc/distributions/distribution.py
```diff
@@ -61,7 +62,7 @@
     "NoDistribution",
 ]
 
-DIST_PARAMETER_TYPES = Union[np.ndarray, int, float, TensorVariable]
+DIST_PARAMETER_TYPES: TypeAlias = Union[np.ndarray, int, float, TensorVariable]
```
canyon289, line 65: Im just curiouys, what does TypeAlias do?

pymc/sampling.py
```diff
@@ -1393,6 +1397,27 @@ def _choose_backend(trace: Optional[Union[BaseTrace, List[str]]], **kwds) -> Bac
     return NDArray(vars=trace, **kwds)
 
 
+def _init_trace(
```
canyon289, line 1400: Does this need a test?

pymc/distributions/distribution.py
michaelosthege: This is a type hint itself. The mypy errors and documentation told me that aliases should be annotated like this.

pymc/sampling.py
michaelosthege: It's covered extensively by the existing tests, though I agree that after extraction it is now easier to test it separately.
But my plan is to refactor this into using McBackend. This `_init_trace()` will then become `run.init_chain()` and then I'd also invest in the tests.

pymc/sampling.py
```diff
@@ -1217,7 +1220,7 @@ def _prepare_iter_population(
     step,
     start: Sequence[PointType],
     parallelize: bool,
-    tune=None,
+    tune: int,
```
twiecki, line 1223: Why remove the optional?

michaelosthege: 1. The (only) call to `_prepare_iter_population` is in a context where tune is already `int`.
2. It's passed down into `_iter_population` which assumed `tune: int`.

...because its an internal function called in exactly one place, it's safe to make it non-optional. The alternative would be to pick an arbitrary default, which we do already in `pm.sample([tune=1000])`.

PR #5431, opened 2022-01-31
Add optional idata_kwargs to sample_numpyro_nuts

### LINKED ISSUE #5379: Accept `idata_kwargs` in `sampling_jax`
Right now we can't pass `idata_kwargs` to `sampling_jax` (most commonly used to set `loglikelihood=False`, so that it does not compute it at the end of sampling). We should accommodate this, to bring the two APIs closer together

ricardoV94:
Skip log_likelihood computation if `idata_kwargs["log_likelihood"]=False`

Closes #5379

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -6,8 +6,6 @@ import warnings
 
 from typing import Callable, List, Optional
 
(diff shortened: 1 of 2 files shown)


lucianopaz (approved):
LGTM

PR #5434, opened 2022-01-31
Updated ExpQuad docstring

### LINKED ISSUE #4871: minor pymc3.gp.cov.Exponential documentation confusion
Currently, the [doc](https://docs.pymc.io/api/gp/cov.html#pymc3.gp.cov.Exponential) for the `ls` parameter in `pymc3.gp.cov.Exponential` states it is involved as as a square: 
https://github.com/pymc-devs/pymc3/blob/bc2c24a3fda645e97e180b7e45d045340476e0b3/pymc3/gp/cov.py#L446

However, the implementation seems to imply it is is not squared in this formula:
```python
import numpy as np
import pandas as pd
from pymc3.gp.cov import Exponential
import plotnine as p9

ls = 100
grid = np.arange(2000)

fn = Exponential(1, ls=ls)

def Exp(l, x):
    return np.exp(-np.abs(x)/(2*l))

def Exp2(l, x):
    return np.exp(-np.abs(x)/(2*l**2))

df0 = pd.DataFrame({
    'function':'pymc3.gp.cov.Exponential(ls)(x)',
    'x':grid,
    'cov':fn(np.zeros((1, 1)), grid[:, None]).eval().ravel(),
})
df1 = pd.DataFrame({
    'function':'np.exp(-np.abs(x)/(2*ls))',
    'x':grid,
    'cov':Exp(ls, grid),
})
df2 = pd.DataFrame({
    'function':'np.exp(-np.abs(x)/(2*ls**2))',
    'x':grid,
    'cov':Exp2(ls, grid),
})
lcorrdf = pd.concat([df0, df1, df2], axis=0)
pl = (
    p9.ggplot(lcorrdf)
    + p9.aes('x', 'cov', color='function')
    + p9.geom_line(alpha=.5)
    + p9.ggtitle(f'ls = 500')
    + p9.theme(figure_size=(6, 4))
)
pl
```
image

fonnesbeck:
Removed square on lengthscale parameter in docstring.

Closes #4871

### DIFF
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -445,7 +445,7 @@ class ExpQuad(Stationary):
 
     .. math::
 
-       k(x, x') = \mathrm{exp}\left[ -\frac{(x - x')^2}{2 \ell} \right]
+       k(x, x') = \mathrm{exp}\left[ -\frac{(x - x')^2}{2 \ell^2} \right]
     """
 
     def full(self, X, Xs=None):
@@ -526,7 +526,7 @@ class Exponential(Stationary):
 
     .. math::
 
-       k(x, x') = \mathrm{exp}\left[ -\frac{||x - x'||}{2\ell^2} \right]
+       k(x, x') = \mathrm{exp}\left[ -\frac{||x - x'||}{2\ell} \right]
     """
 
     def full(self, X, Xs=None):
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -445,7 +445,7 @@ class ExpQuad(Stationary):
 
     .. math::
 
-       k(x, x') = \mathrm{exp}\left[ -\frac{(x - x')^2}{2 \ell^2} \right]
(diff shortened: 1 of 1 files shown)


michaelosthege (commented):
In the [kernel cookbook](https://www.cs.toronto.edu/~duvenaud/cookbook/) they write `2 \ell^2`.
Same in the [`sklearn.gaussian_process.kernels.RBF` documentation](https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.kernels.RBF.html).

Is our implementation special?
If so we should probably elaborate why our docs differ from these other sources.

michaelosthege:
The issue #4871 was about the `Exponential` kernel but you changed it in the `ExpQuad` ?

ghost:
Ah, yes. I was going by the line number cited in the issue, which points to the ExpQuad docs. Will fix.

michaelosthege (approved):
👍

PR #5438, opened 2022-02-01
Refactor Mixture distribution for V4

### LINKED ISSUE #4781: Mixtures not refactored for V4

ricardoV94:
This PR is an attempt to refactor (Marginalized) Mixture distributions for V4

- [x] Refactor Mixture with iterable components
- [x] Refactor Mixture with single component
- [x] Refactor NormalMixture
- [x] Rebase from main and handle new meaning of size for multivariate dists
- [x] Reenable TestMixtureVsLatent
- [x] Deprecate MixtureSameFamily (keep tests)
- [x] Add more design decision context to refactor commit
- [x] Add moment and tests
- [x] Allow for Nested Mixtures / Other symbolic distributions as components (reminder #5533)

## Changes
There are two big changes in how Mixture works compared to V3:
1. The support dimensionality of the Mixture no longer depends on the size of the weights, but on the support dimensionality of the components. This means one cannot use a batched scalar distribution (like Normal) as if it was a vector distribution, which was supposedly allowed by the docstrings. However, while the old random method respected this, the logp method did not (the logp would be indifferent to mixing of values across this "fake" vector components), so this is not clearly a regression.
```python
# This mixture now assumes interchangeability across the 5 values of each component, 
# regardless of the dimensionality of the weights
mix = pm.Mixture.dist(w=[0.5, 0.5], comp_dists=pm.Normal.dist([-10, 10], size=(5, 2)))
```
I am exploring adding a keyword argument to override the support dimensionality in a follow-up PR. I managed to do this for the random method, but haven't had time yet to figure out what needs to be changed in the logp method.

2. Nested Mixtures or Mixtures using non pure RandomVariables (such as Censored variables) are not yet possible. This will require creating a dispatch to retrieve and manipulate these variables (i.e., check ndim_supp, resize), in the same way that's being done with the component RandomVariables. This should be straightforward, since these had to be implemented as methods of the SymbolicDistribution anyway...

Closes #4781


ricardoV94:
@Sayam753 could you give some context on what was the idea with `MixtureSameFamily`?

Now that we have meta information about the components (ndim_supp, ndims_params), it is perhaps no longer needed?

Sayam753:
> could you give some context on what was the idea with MixtureSameFamily?

`MixtureSameFamily` distribution helps to create mixture distribution for multivariate distributions.

The legacy `Mixture` distribution assumes that the mixture components are present in the last dimension of a distribution. This creates a problem for multivariate distributions, because in this case, the last dimension corresponds to events, not the mixture components.

`MixtureSameFamily` takes an input an integer asking which axis to consider for mixture components. And it reduces that axis during logp computations.

@lucianopaz is the magician behind this distribution.

ricardoV94:
If I understand correctly, I think we can infer the correct mixture axis from the meta information we have now. 

That's my TODO point above. So far, I have just been hacking it by trial and error. 

Do you think there is inherent ambiguity still?  Even considering ndim_supp, ndims_params, shape_from_params... and whatever else methods we have to reason about shapes of RandomVariables?

pymc/distributions/mixture.py
```diff
@@ -112,454 +132,178 @@ class Mixture(Distribution):
...
-    @comp_dists.setter
-    def comp_dists(self, comp_dists):
-        self._comp_dists = comp_dists
-        if isinstance(comp_dists, Distribution):
-            self._comp_dist_shapes = to_tuple(comp_dists.shape)
-            self._broadcast_shape = self._comp_dist_shapes
-            self.comp_is_distribution = True
+        stacked_components_ = at.stack(components_type, axis=-1)
+        if weights.ndim < stacked_components_.ndim:
+            weights_ = at.shape_padaxis(weights_type, axis=weights_type.ndim - 1)
```
ricardoV94: This is very likely completely wrong. I just trial and errored to get the tests to pass. From what @Sayam753 said in https://github.com/pymc-devs/pymc/pull/5438#issuecomment-1030590138, it should probably be based on the support dimensionality of the components, not on the dimensionality of the weights

pymc/distributions/mixture.py
```diff
@@ -112,454 +132,178 @@ class Mixture(Distribution):
...
+        axis=-1,
+    )
 
-        Returns
-        -------
-        TensorVariable
-        """
-        w = self.w
+    # TODO: Is one padding always enough?
+    if weights.ndim < components_logp.ndim:
```
ricardoV94: Gotta figure out how weights should be padded in the logp

pymc/distributions/mixture.py
```diff
@@ -112,454 +132,178 @@ class Mixture(Distribution):
...
-        # else:
-        #     samples = np.reshape(samples, size + dist_shape)
-        # return samples
 
-    def _distr_parameters_for_repr(self):
-        return []
+@_get_moment.register(MarginalMixtureRV)
+def get_moment_marginal_mixture(op, rv, rng, weights, *components):
+    moment_components = at.stack([get_moment(component) for component in components], axis=-1)
+    return at.sum(weights * moment_components, axis=-1)
```
ricardoV94: Needs padding of weights as well

pymc/distributions/mixture.py
```diff
@@ -112,454 +132,178 @@ class Mixture(Distribution):
...
+            component.tag.ignore_logprob = True
+
+        if size is not None:
+            mix_out = cls.change_size(mix_out, size)
+
+        return mix_out
+
+    @classmethod
+    def ndim_supp(cls, rng, weights, *components):
+        # We already checked that all components have the same dimensionality
```
larryshamalama: I think that the signature of this method should be `ndim_supp(cls, weights, *components)` since an `rng` doesn't seem to be provided [here](https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/distribution.py#L587)

ricardoV94: Good catch, I think you're right!

pymc/distributions/mixture.py
```diff
@@ -112,454 +132,178 @@ class Mixture(Distribution):
...
-        else:
-            return at.squeeze(
-                at.stack([comp_dist.logp(value) for comp_dist in comp_dists], axis=-1)
-            )
+    @classmethod
+    def graph_rvs(cls, rv):
+        # We return rv, which is a pseudo RandomVariable, that contains a choices RV
+        # in its inner graph. We want super().dist() to generate components + 1 rngs for
+        # us, and it will do so based on how many elements we return here
+        return (*rv.tag.components, rv)
```
ricardoV94, line 365: Perhaps better to return Aesara `NoneConst` instead of `rv` to avoid confusion down the road

ricardoV94:
Figured out the weights shape padding/broadcasting thanks tho @lucianopaz!

pymc/distributions/mixture.py
```diff
@@ -112,454 +132,211 @@ class Mixture(Distribution):
...
-                defaults.append("mode")
-        except (AttributeError, ValueError, IndexError):
-            pass
+        # Append (ndim_supp + stack) dimensions to the right of mix_indexes
+        mix_indexes_padded = at.shape_padright(mix_indexes_, component_ndim_supp + 1)
 
-        super().__init__(shape, dtype, defaults=defaults, *args, **kwargs)
+        # Append  missing dimensions (if any) to the left of stacked_components
+        stacked_components_ = at.stack(components_, axis=-1)
+        stacked_components_padded_ = at.shape_padleft(
```
ricardoV94: I think this is just the `weights_ndim_reps`

pymc/distributions/mixture.py
```diff
@@ -112,454 +132,211 @@ class Mixture(Distribution):
...
-                comp_mode_logps = comp_dists.logp(comp_dists.mode)
-            else:
-                comp_mode_logps = at.stack([cd.logp(cd.mode) for cd in comp_dists])
+            # Create new rng for the choices internal RV
+            mix_indexes_rng = aesara.shared(np.random.default_rng())
+
+        # Extract support and replication ndims from components and weights
+        component_ndim_supp = components[0].owner.op.ndim_supp
+        component_ndim_reps = components[0].ndim - component_ndim_supp
+        weights_ndim_reps = max(0, weights.ndim - component_ndim_reps - 1)
```
ricardoV94: We could instead change the size of the `Mixture` so that there are never batched dimensions coming exclusively from the weights. Perhaps issue a UserWarning before doing this...

pymc/distributions/mixture.py
```diff
@@ -112,454 +132,211 @@ class Mixture(Distribution):
...
-
-        try:
-            if isinstance(comp_dists, Distribution):
-                comp_mode_logps = comp_dists.logp(comp_dists.mode)
-            else:
-                comp_mode_logps = at.stack([cd.logp(cd.mode) for cd in comp_dists])
+            # Create new rng for the choices internal RV
+            mix_indexes_rng = aesara.shared(np.random.default_rng())
+
+        # Extract support and replication ndims from components and weights
```
ricardoV94, line 222: Rename `replication/reps` to `batched`

pymc/tests/test_mixture.py
```diff
@@ -627,53 +632,27 @@ def test_normal_mixture_nd(self, nd, ncomp):
...
+            obs1 = Mixture("obs", w=ws, comp_dists=comp_dist, observed=observed)
 
         with Model() as model2:
-            # Expected to fail if comp_shape is not provided,
-            # nd is multidim and it does not broadcast with ncomp. If by chance
-            # it does broadcast, an error is raised if the mixture is given
-            # observed data.
-            # Furthermore, the Mixture will also raise errors when the observed
-            # data is multidimensional but it does not broadcast well with
-            # comp_dists.
```
ricardoV94: This no longer seems to be a problem. @lucianopaz can you confirm that if the current tests pass, this is indeed fine?

ricardoV94:
Except for the moment, everything seems ready for review. Thanks to both @lucianopaz and @larryshamalama for your inputs an help!

pymc/distributions/mixture.py
```diff
@@ -112,457 +125,249 @@ class Mixture(Distribution):
...
+
+        single_component = len(components) == 1
+
+        # Extract support and replication ndims from components and weights
+        ndim_supp = components[0].owner.op.ndim_supp
+        ndim_batch = components[0].ndim - ndim_supp
+        if single_component:
+            # One dimension is taken by the mixture axis in the single component case
+            ndim_batch -= 1
+        weights_ndim_batch = max(0, weights.ndim - ndim_batch - 1)
```
larryshamalama, line 231: A bit confused with everything going on with dimensions 😅 What is the range of `weights_ndim_batch` here? Is it either 0 or 1 (based on the `if` statement)?

pymc/distributions/mixture.py
```diff
@@ -112,457 +125,249 @@ class Mixture(Distribution):
...
+        # Draw mixture indexes
+        mix_indexes_ = at.random.categorical(weights_broadcasted_, rng=mix_indexes_rng_)
+
+        # Append (ndim_supp + stack) dimensions to the right of mix_indexes
+        mix_indexes_padded = at.shape_padright(mix_indexes_, ndim_supp + 1)
+
+        # Index components and squeeze mixture dimension
+        if single_component:
+            mix_out_ = at.take_along_axis(components_[0], mix_indexes_padded, axis=-ndim_supp - 1)
+            # There is a Aeasara bug in squeeze with negative axis
```
larryshamalama: After some thought, this sounds right!

pymc/distributions/mixture.py
```diff
@@ -112,457 +125,249 @@ class Mixture(Distribution):
...
+
+@_logprob.register(MarginalMixtureRV)
+def marginal_mixture_logprob(op, values, rng, weights, *components, **kwargs):
+    (value,) = values
+
+    # single component
+    if len(components) == 1:
+        # Need to broadcast value across mixture axis
+        mix_axis = -components[0].owner.op.ndim_supp - 1
+        components_logp = logp(components[0], at.expand_dims(value, mix_axis))
```
larryshamalama, line 382: Looks good!

larryshamalama (commented):
Just left a single question, looks great :)

pymc/distributions/mixture.py
ricardoV94: `weights_ndim_batch` is the extra dimensions besides what is implied by the components dimensions. It can be arbitrarily large.
```python
# In this example, weights_ndim_batch = 4
pm.Mixture(w=(np.full((4, 4, 4, 4, 4, 2), comp_dists=pm.Normal.dist(size=(4, 2)))`
```

ricardoV94: I added a comment to explain better in the code

ricardoV94:
@OriolAbril Any idea why the `tensor_like` here does not have a link to the glossary? https://pymc--5438.org.readthedocs.build/en/5438/api/distributions/generated/pymc.NormalMixture.html

OriolAbril:
There should be spaces both before and after the colon that separates parameter name and type

ricardoV94:
Moments are also ready thanks to @larryshamalama. This is ready for review and merge :rocket:

pymc/distributions/mixture.py
```diff
@@ -53,516 +72,357 @@ class Mixture(Distribution):
...
-                    type(comp_dists)
-                    if not isinstance(comp_dists, Iterable)
-                    else [type(c) for c in comp_dists]
+
+        # Check that components are not associated with a registered variable in the model
+        components_ndim = set()
+        components_ndim_supp = set()
+        for dist in comp_dists:
+            # TODO: Allow these to not be a RandomVariable as long as we can call `ndim_supp` on them
+            #  and resize them
```
larryshamalama, line 183: Just wanting to clarify (again?) for my understanding: what else can `dist` be? For instance, is this where we want to be more flexible in allowing things like nested mixtures and censored distributions? E.g. something like [line 208 below](https://github.com/pymc-devs/pymc/pull/5438/files#diff-a46f0cbffb99eea1c2c7c51bfae351225ae8f38ad9b825d0b4687a7456bc6288R208)

larryshamalama (commented):
Looks great 🙂 Thanks for leading the effort @ricardoV94

pymc/distributions/mixture.py
ricardoV94: Yeah exactly!

twiecki (approved):

Issue #5448, opened 2022-02-06
Remove pin of SciPy 1.8.0 once we run Aesara 2.4.0


michaelosthege:
## Description of your problem
 There's a new SciPy release 1.8.0 and they refactored their public/private API.

https://github.com/scipy/scipy/releases/tag/v1.8.0

Aesara has some import incompatibilities and this is causing our CIs to fail.

Related Aesara issue: https://github.com/aesara-devs/aesara/issues/810

The incompatibility will most probably affect `v3` too, so we'll want to backport a pin of the upper SciPy version.

## Versions and main components
* PyMC/PyMC3 Version: `<=3.11.4`, `<=4.0.0b2`
* Aesara/Theano Version: all `<=2.3.8`

ricardoV94:
The temporary pin can be removed once we update our Aesara dependency (pending #5446)

PR #5449, opened 2022-02-06
Pin SciPy to `<1.8.0`

michaelosthege:
See tracking issue #5448.


OriolAbril:
rtd is still failing, _from trying to install numpy==1.19_ which I can only assume is a strange requirement of some testing library. We might need to split test/docs... requirements for this to work (and therefore address https://github.com/pymc-devs/pymc/issues/5074)

michaelosthege:
Running `code --diff .github\workflows\pytest.yml .github\workflows\jaxtests.yml` and `code --diff .github\workflows\pytest.yml .github\workflows\arviz_compat.yml` it looks like the environments between those workflows are quite different:
* `pytest.yml` vs. `jaxtests.yml`/`arviz-compat.yml`
* `ubuntu-18-04` vs. `ubuntu-latest`/`maxos-latest`
* `environment-dev-py37.yml` vs. `environment-dev-py39.yml`
* no extras vs. `pip install "numpyro>=0.8.0"`/`pip install git+git://github.com/arviz-devs/arviz.git`

OriolAbril:
I guess there are significant differences between ubuntu 18 and latest (should be 20 I think). But also with python 3.7! IIRC, the latest versions of numpy/scipy... don't support it anymore, there was a NEP about this but don' t have the link at hand, I'll try to come back and help a bit more tonight.

twiecki:
This is probably fine as a hot fix, but technically this is not a pymc issue but an aesara one and that's where it should be fixed (e.g. by pinning the version there), rather than here.

twiecki:
Just seeing it's already fixed there: https://github.com/aesara-devs/aesara/pull/812 so shouldn't we just wait for a new aesara release and pin to that?

twiecki:
If you pin `aesara=2.4.0` we shouldn't need to pin anything else here.

twiecki:
Ah, that's what you did in [498dc43](https://github.com/pymc-devs/pymc/pull/5449/commits/498dc435c5c65e23a1497387f5f3a88fdbdf8e61). That actually looks right, not sure why numpy even gets compiled in the arviz installation?

ricardoV94:
~~Just quick check to see if problems come from the recent change to mamba...~~

It wasn't that

conda-envs/environment-dev-py38.yml
```diff
@@ -22,8 +22,7 @@ dependencies:
 - pytest-cov>=2.5
 - pytest>=3.0
 - python-graphviz
-- python=3.8
```
ricardoV94, line 25: Why was the python version dependency removed?

twiecki: Yeah that's not right.

ricardoV94:
Okay installation seems to have worked now!

ricardoV94:
Also weirdly... arviz compat and jax workflows are using the "dev" environments, and not the "test" ones... Why is that?

And some environments pin the pandas version but not others...

ricardoV94:
Reverting the Aesara upgrade, because that fails due to https://github.com/pymc-devs/pymc/issues/5446

twiecki:
> Also weirdly... arviz compat and jax workflows are using the "dev" environments, and not the "test" ones... Why is that?

It shouldn't, oversight.

ricardoV94:
The `pip-from-conda` hook does not seem to handle double sided version limits...

ricardoV94:
Ok, so `scipy` appears in the list of packages that should be excluded from requirements... not sure why this is:

https://github.com/pymc-devs/pymc/blob/b54c8af5c4a43b1a46b6597ebae4ddec1b5b22ca/scripts/generate_pip_deps_from_conda.py#L53

ricardoV94:
Docs failure is related to Aeppl dependency. Will open a separate PR to update it

twiecki (approved):

PR #5452, opened 2022-02-09
Community gov update

cluhmann:
This PR updates the governance document to reflect recent discussions regarding the mission and naming of PyMC's community team as requested by @OriolAbril.  As with other recent updates to the document, the diagram illustrating the organization of the various teams (docs/community_diagram.png) does not reflect the current update.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -664,7 +664,7 @@ Team:
 * Documentation team members are given permissions to [pymc-examples](https://github.com/pymc-devs/pymc-examples)
   and [resources](https://github.com/pymc-devs/resources) repositories.
 
-In addition, Council members are given admin rights to all repositories within
+In addition, Council members are given administrative rights to all repositories within
 the [pymc-devs](https://github.com/pymc-devs) organization.
 
 #### Discourse
@@ -709,7 +709,7 @@ provide examples on how to assign permissions:
   * Added to all private communication channels
   * Write permissions on pymc-examples and resources repositories, triage permissions
     to pymc and pymc-experimental repositories
-  * Admin access to ReadTheDocs accounts
+  * Administrative access to ReadTheDocs accounts
 
 - Dolors, core contributor, dev and community teams
   * Added to the Community PyMC_team and given "leader" trust level,
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -640,7 +640,7 @@ core contributor membership sections.
 All tasks related to the project that are not specifically listed in the
 description of a dedicated team are the responsibility of the PyMC team
 as a whole. At the time of writing, this includes but is not limited to:
-enforcing this governance  and the [PyMC code of conduct](https://github.com/pymc-devs/pymc/blob/main/CODE_OF_CONDUCT.md), developing project-related grants, soliciting of project sponsorships, decisions regarding the allocation of project-related funds, planning of the project roadmap, and triaging of GitHub issues.
+enforcing this governance and the [PyMC code of conduct](https://github.com/pymc-devs/pymc/blob/main/CODE_OF_CONDUCT.md), developing project-related grants, soliciting of project sponsorships, decisions regarding the allocation of project-related funds, planning of the project roadmap, or triaging of GitHub issues.
 
 ### Team structure in practice
 This section describes how members of the PyMC team are given
@@ -672,10 +672,11 @@ Similar to the above section, Discourse permissions are also mapped to the commu
 and the two contributor roles.
 
 Role:
-- Recurring Contributors are added to the [PyMC_team](https://discourse.pymc.io/g/PyMC_devs)
-  group and are given the "leader" trust level.
-- Core Contributors are given [moderator permissions](https://discourse.pymc.io/g/moderators)
-  if possible.
+- Recurring Contributors are given no special permissions
+- Core Contributors are added to the [PyMC_core](https://discourse.pymc.io/g/PyMC_core)
+  group independently of the teams they are part of.
+  Core Contributors in the community team are also added to the
+  [Community Team](https://discourse.pymc.io/g/Community_Team) group.
 
 #### Accounts and services ownership and administration
 The PyMC Project also has accounts and hosts services on several platforms
@@ -694,7 +695,7 @@ provide examples on how to assign permissions:
 <details><summary>See permission examples</summary>
 
 - Arnau, recurrent contributor, community team
-  * Added to the Community PyMC_team and given "leader" trust level
+  * No permissions on Discourse
   * Added to all private communication channels
   * Triage permissions on the pymcon, pymc-data-umbrella, and other "event" repositories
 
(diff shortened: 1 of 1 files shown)


twiecki (approved):

GOVERNANCE.md
```diff
@@ -630,10 +630,10 @@ write all the documentation for the PyMC library.
...
-### Discourse team
-The focus of the Discourse team is managing and energizing the PyMC Discourse.
+### Community team
+The focus of the Community team is activities intended to nuture, energize, and grow the community of PyMC users and contributors.
 
-For current members of the discourse team, refer to the recurrent and
+For current members of the community team, refer to the recurrent and
 core contributor membership sections.
 
 ### "No-team" tasks
```
OriolAbril, line 639: "no-team" tasks need to be updated to avoid listing community team things.

GOVERNANCE.md
```diff
@@ -664,11 +664,11 @@ Team:
 * Documentation team members are given permissions to [pymc-examples](https://github.com/pymc-devs/pymc-examples)
```
OriolAbril, line 663: Community team members now also get github permissions to pymcon and pymc-data-umbrella repos (plus other repos we might create for other events and initiatives)

GOVERNANCE.md
```diff
@@ -664,11 +664,11 @@ Team:
...
-In addition, Council members are given admin rights to all repositories within
+In addition, Council members are given administrative rights to all repositories within
 the [pymc-devs](https://github.com/pymc-devs) organization.
 
 #### Discourse
-Similarly to the above section, Discourse permissions are also mapped to the discourse team
+Similarly to the above section, Discourse permissions are also mapped to the community team
 and the two contributor roles.
 
 Role:
```
OriolAbril, line 674: Do we all agree with these roles? 

I don't know how Discourse works and I'm the one who proposed that so we'd have somehting to start the discussion. IMO the process should be as much as possible a "take what we have or want and write it down here" not really write down something here and then change everything else. cc @junpenglao 

What I will do though after a grace period is making sure we are doing what we have written here; and if that were an issue restart the discussion, but it would be very nice to have the discussion "beforehand"

OriolAbril (changes requested):

GOVERNANCE.md
cluhmann: I wasn't sure about the allocation of the no-team tasks.  Given the discussion of the team's mission, I guess these makes sense?

No-team tasks: fundraising and issue triaging
Community: running PyMC related events like PyMCon or sprints, outreach, or presence on social networks.

But then it seemed like the majority of the no-team tasks were being shifted to the community team and we had discussed how things like PyMCon would almost certainly involve many people/teams (social media activity certainly will be).

OriolAbril: Sounds good. I think it's important to clear expectations on both what teams are supposed and _not supposed_ to do. I also added a clarification on the docs team description for example; its existence doesn't mean only members of the docs team will write documentation or that only members of the docs team will update notebooks to v4. The docs team goal is to maintain the documentation tooling and infrastructure and see if it can be improved, review the docs to see if there are missing parts... but then its everyone that uses said tooling and documents the library. 

I think something similar should happen with pymcon/sprints, and it did already with the sprint. The community team should probably take care of organizing and coordinating the event, like Meenal, Reshama, Beryl and I did for the sprint, but then everyone who can should help with reviewing talk proposals, giving talks, and other tasks "prepared" by the organization team. More or less the same way Ricardo and Austin gave webinars for the sprint and Sayam, Ricardo and Thomas are volunteering for the sprint itself. I wouldn't consider those part of "organizing" even if they are also critical and necessary for the event to suceed.

Does that make sense? 

side note, if you think it would be good to still keep a longer list of "no-team" example tasks I have some ideas: enforcing the governance and code of conduct are also very clear "no team" tasks, also funding allocation and roadmap planning, fundraising can be divided between grant applications and reaching out to potential sponsors too.

cluhmann: I have added more info about both the community team activities/responsibilities and tried to limit the intended scope (like you did for the doc team).  I added additional no-team activities as you suggested.

GOVERNANCE.md
cluhmann: I changed this and the permission examples.  I think I hit got it correct, but some of the examples are a bit complicated.

GOVERNANCE.md
cluhmann: I think this is roughly the way it works currently.  But I agree that getting everyone on the same page now is a good idea.  I know that there has been some confusion about what permission the various discourse roles/levels have.

junpenglao: We should limit the moderator and admins to community team member if possible.

GOVERNANCE.md
OriolAbril: Thanks, now we should probably review the team members. i.e. Meenal is leading the sprint and taking care of pymc twitter so should definitely be on the team with the goals defined now.

GOVERNANCE.md
OriolAbril: @junpenglao I don't understand the comment, should that be changed? Right now core contributors on the community team and council members should be on the pool of rotating moderator/admin persissions, should it be only community core contributors?

GOVERNANCE.md
```diff
@@ -679,24 +679,24 @@ Role:
...
 #### Permission examples
 This section lists some imaginary contributors with their teams and roles to
 provide examples on how to assign permissions:
 
 <details><summary>See permission examples</summary>
 
-- Arnau, recurrent contributor, discourse team
-  * Added to the Discourse PyMC_team and given "leader" trust level
+- Arnau, recurrent contributor, community team
+  * Added to the Community PyMC_team and given "leader" trust level
```
OriolAbril: This was a reference to the first point in discourse roles about adding members to the `PyMC_team` (or devs or something else, again, don't really know what I'm talking about) on Discourse. The roles look wrong on that end which is probably why this doesn't make much sense either.

GOVERNANCE.md
```diff
@@ -705,21 +705,21 @@ provide examples on how to assign permissions:
...
-    to pymc-examples and resources repositories
+  * Write permissions on pymc, pymc-experimental, pymcon, pymc-data-umbrella, and other "event" repositories, triage permissions
+    on pymc-examples and resources repositories
 
 - Eudald, core contributor, no dedicated team membership
-  * Added to the Discourse PyMC_team and given "leader" trust level
+  * Added to the Community PyMC_team and given "leader" trust level
   * Added to all private communication channels
   * Triage permissions on all repositories
   * Access to pymc_devs twitter account as they are the main manager
```
OriolAbril:
```suggestion
  * Access to pymc gmail, drive and grant application platforms (i.e. CZI slack) as they are the main grant writing coordinator
```
Updated that so we keep a no-team member here with special and important permissions.

GOVERNANCE.md
cluhmann: Agreed.

GOVERNANCE.md
cluhmann: Yeah, the "PyMC_team" thing was confusing to me.

michaelosthege (approved):
Looks generally good to me.
Can you push it over the finish line so we have one fewer open PR in the list?

GOVERNANCE.md
canyon289: We can realign discourse teams as needed per this doc

OriolAbril: That is the idea, but we need to define something here that we agree on and makes sense. I have no idea how discourse teams work and wrote the current proposal :sweat_smile:. I wasn't sure it made much sense when I wrote it and I take @cluhmann struggling to understand what it meant as a confirmation that the doc needs updating and clarifying before we can go back on Discourse and make sure we follow what is defined here

cluhmann: Just to explicate some of the imprecision here, take our friend Arnau who is a recurrent contributor on the community team.  Questions:

- Do we add Arnau to the "the Community PyMC_team"? If so, what is that?  Is this just supposed to be "Added to the Community team"?  How does one become "a recurrent contributor on the community team" without being "added to the community team"?  In other words, are these two separate steps or is this just a statement that, as a recurrent contributor, Arnau is necessarily listed as being part of the community team?
- Do we give Arnau "leader" trust level on discourse?  Not moderator?  Administrator?  Trust level X?

OriolAbril: IMO being a contributor (of any type) and getting the premissions associated with that type are one and the same, a necessary and sufficient condition. If we write that core contributors of the dev team get write permissions to pymc repo, there should be no core dev contributors without nor people with write permissions who are not core dev contributors (or other role that gives those permissions like could be council member). 

I don't really know what "leader" trust level entails, but as a general rule, recurrent contributors should not have administrative permissions, that should be restricted to core contributors. If "leader" trust level is somewhat similar to triage permissions on github I'd vote for that.

cluhmann: I would agree.  Just trying to highlight some of what (I think) we're looking for input/consensus on.

twiecki (approved):

OriolAbril (approved):

canyon289:
Looks great, merging and we can fix up anything in a follow up PR

PR #5455, opened 2022-02-09
Record sampling time for NumPyro sampler

AustinRochford:
Record sampling time when using BlackJax

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -262,6 +262,7 @@ def sample_numpyro_nuts(
         sample_stats=_sample_stats_to_xarray(pmap_numpyro),
         coords=coords,
         dims=dims,
+        attrs={"sampling_time": (tic3 - tic2).total_seconds()},
         **idata_kwargs,
     )
 


pymc/sampling_jax.py
```diff
@@ -264,5 +264,6 @@ def sample_numpyro_nuts(
         dims=dims,
         **idata_kwargs,
     )
+    az_trace.sample_stats.attrs["sampling_time"] = (tic3 - tic2).total_seconds()
```
fonnesbeck: Do we want to call it "sampling_time_posterior" so that we know which sampling it is referring to? I suppose we don't care about the sampling time of prior/posterior checks, so probably fine like this.

fonnesbeck (approved):

pymc/sampling_jax.py
AustinRochford: I chose this because it is the same as the default PyMC samplers.

michaelosthege: Compare [this line](https://github.com/pymc-devs/pymc/blob/main/pymc/backends/arviz.py#L155). Attributes should be provided via the `attrs` kwarg.
```suggestion
        attrs={
            "sampling_time": (tic3 - tic2).total_seconds(),
        },
        **idata_kwargs,
    )
    az_trace.sample_stats.attrs["sampling_time"] = (tic3 - tic2).total_seconds()
```
<sup>This suggestion does not come with a `black` guarantee.</sup>

michaelosthege (changes requested):

pymc/sampling_jax.py
AustinRochford: Changed!

michaelosthege (approved):

PR #5456, opened 2022-02-09
Move the NumPyro sampling time to the sample_stats group

AustinRochford:
`sampling_time` should be an attribute of the `sample_stats` group, not the top-level `InferenceData` object.

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -262,7 +262,7 @@ def sample_numpyro_nuts(
         sample_stats=_sample_stats_to_xarray(pmap_numpyro),
         coords=coords,
         dims=dims,
-        attrs={"sampling_time": (tic3 - tic2).total_seconds()},
+        sample_stats_attrs={"sampling_time": (tic3 - tic2).total_seconds()},
         **idata_kwargs,
     )
 


michaelosthege:
Isn't it usually in the `idata.posterior.attrs`?
I'm sorry I suggested the wrong solution there. The converter code is a little diffuse.

michaelosthege:
Looks like it normally ends up in both groups:
```
>>> idata.sample_stats.attrs
{'created_at': '2022-02-09T19:46:16.342747', 'arviz_version': '0.11.4', 'inference_library': 'pymc', 'inference_library_version': '4.0.0b2', 'sampling_time': 20.66627860069275, 'tuning_steps': 1000}
>>> idata.posterior.attrs
{'created_at': '2022-02-09T19:46:16.340767', 'arviz_version': '0.11.4', 'inference_library': 'pymc', 'inference_library_version': '4.0.0b2', 'sampling_time': 20.66627860069275, 'tuning_steps': 1000}
>>>
```

And also when passing it just as `attrs`:
```
>> arviz.from_dict(dict(x=[[1,2,3]]), sample_stats=dict(tune=[[True]]), attrs=dict(sampling_time=3))                
Inference data with groups:
        > posterior
        > sample_stats
>>> arviz.from_dict(dict(x=[[1,2,3]]), sample_stats=dict(tune=[[True]]), attrs=dict(sampling_time=3)).sample_stats.attrs
{'created_at': '2022-02-09T19:49:24.113039', 'arviz_version': '0.11.4', 'sampling_time': 3}
```
But for `sample_stats_attrs` I get an error:
```
>>> arviz.from_dict(dict(x=[[1,2,3]]), sample_stats=dict(tune=[[True]]), sample_stats_attrs=dict(sampling_time=3)).sample_stats.attrs 
TypeError: from_dict() got an unexpected keyword argument 'sample_stats_attrs'
```
@AustinRochford is this different for you locally?
(Note that the line is not covered by tests.)

AustinRochford:
No worries, I should have tested more thoroughly.  Works for me:

image

michaelosthege:
Strange. The env I tested the above command with was created today.
But on my other PC it works.

@aloctavodia @OriolAbril where do you want this information to end up?
With `pm.sample` it goes into all groups, but it looks like you introduced these group-wise kwargs to split things up?

Also note that when `sample_stats` is an empty dict or not provided, the entire group and thereby also the `attrs` are missing from the result.

OriolAbril:
InferenceData level attrs are still unreleased feature IIRC (like with pymc, installing from github shows the same version as latest release), and aren't yet well integrated with read/write functions. I would avoid them for now. Between setting attributes of the posterior dataset of the sample_stats dataset I think both are fine and would probably add those to both like with other libraries. Does the numpyro one use `from_dict` to generate the InferenceData?

ricardoV94:
@AustinRochford Any progress on this one?

michaelosthege:
The respective lines now look different on `main`.
`az.from_dict` is used and there was also an ArviZ release in the meantime.

I assume we can consider this as "done" or "no longer relevant".

Issue #5459, opened 2022-02-10
Update docstrings for better rendering on website and numpydoc compliance


OriolAbril:
### Overview
We have to review and update all the docstrings of objects part of the public API (that is, that appear on the [API reference](https://www.pymc.io/projects/docs/en/latest/api.html) section of the website). Why? 

There are two main reasons:
* Some functions are not documented correctly. i.e. pymc.Uniform didn't document that lower and upper bounds are optional and default to 0 and 1 respectively
* Not following numpydoc results in non-ideal rendering. i.e. if there are no spaces around the colon that separates parameter name and type numpydoc+sphinx can't render it correctly and the type ends up being shown as bold text instead of being a link to the documentation of that object (which is particularly useful for functions taking or returning external objects such as ArviZ InferenceData or aesara TensorVariable).

**Current view of the parameter section of pymc.Uniform**

![imatge](https://user-images.githubusercontent.com/23738400/174354131-83796711-5641-4efd-a293-c585e32787c1.png)

**Before doing the changes described below**

![imatge](https://user-images.githubusercontent.com/23738400/153495274-5b63c32d-c59d-4a23-a241-62d4bdd7118b.png)

### Details
A detailed tutorial on how to go about these updates is available at the [PyMC-Data Umbrella sprint website](https://pymc-data-umbrella.xyz/en/latest/sprint/docstring_tutorial.html) (the page also has a video recorded by @reshamas showing an example PR).

### How to choose what page to work on?
Look for pages that don't have hyperlinks in _all_ their parameters. In case of doubt, don't hesitate to ask in this issue for pages to work on.

ravichikkam:
I am working on the docstrings for **weibull distribution**. 

EDIT (by @OriolAbril ): I moved the updated list to the initial comment so its easier to find for new people getting to this issue

jmloyola:
Working on `pm.Data`.

cuchoi:
Is this where the `ConstantData` and `MutableData` docs should appear? https://docs.pymc.io/en/v3/api/data.html. I can't find them.

Edit: There in v4 here https://docs.pymc.io/en/latest/api/data.html

chritter:
Working on [`pymc.Exponential`](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Exponential.html#pymc-exponential)

cuchoi:
Going to work on `pymc.HyperGeometric`

sagartomar:
[...]

PR #5461, opened 2022-02-10
some small improvements to get everything ready for the sprint

OriolAbril:
+ [x] ~what are the (breaking) changes that this PR makes?~ None
  - This will make things easier for people working on #5459

### DIFF
--- a/docs/source/contributing/jupyter_style.md
+++ b/docs/source/contributing/jupyter_style.md
@@ -47,13 +47,13 @@ This guide does not teach nor cover MyST extensively, only gives some opinionate
 ### Variable names
 
 * Above all, stay consistent with variable names within the notebook. Notebooks using multiple names for the same variable will not be merged.
-
 * Use meaningful variable names wherever possible. Our users come from different backgrounds and not everyone is familiar with the same naming conventions.
-
+  - Annotate dimensions too. Notebooks are published to be read, so even if the shape is derived
+    from the inputs or you don't like to use named dims and don't use them in your presonal
+    code, notebooks must use dims, even if annotating and not setting the shape.
+    It makes the code easier to follow, especially for newcomers.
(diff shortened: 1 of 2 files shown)


docs/source/contributing/jupyter_style.md
```diff
@@ -186,14 +186,23 @@ The syntax is as follows:
...
+The {ref}`MyST target <myst:syntax/targets>` (the `(notebook_name)=` bit)
 is important to ease referencing and linking notebooks between each other.
+It must be notebook specific, for example its file name.
 
 Tags can be anything, but we ask you to try to use [existing tags](https://github.com/pymc-devs/pymc/wiki/Categories-and-Tags-for-PyMC-Examples)
 to avoid the tag list from getting too long.
 
-Each notebook should have a single category indicating the level of the notebook.
-Choose a category from [existing categories](https://github.com/pymc-devs/pymc/wiki/Categories-and-Tags-for-PyMC-Examples#categories).
+Each notebook should have a one or two categories indicating:
```
canyon289, line 196: I think we should make it two so its not ambiguous , also so its very clear what category a document is in

canyon289 (approved):

docs/source/contributing/jupyter_style.md
OriolAbril: The problem is that most notebooks are a mix of different types, so forcing the two categories already is a huge effort. My idea was to use that for new notebooks, for notebooks that already match a type and for notebooks we want to split/combine/restructure

canyon289: My suggestion is force two moving forward and we can fix the older ones over time. This way at least at some point in the future all notebooks will have both labels

OriolAbril: sounds good, I have updated the proposal template in https://github.com/pymc-devs/pymc-examples/pull/281. I'd still keep it like this here for now because >95% of the changes that we have to do (and 99.9% of the changes done by newcomers) are updating existent notebooks, not writing new ones

PR #5472, opened 2022-02-15
Change tests included in arviz_compat workflow

### LINKED ISSUE #5410: Consider removing `test_distributions.py` from arviz-compatibility workflow
This job usually takes more than 45-50 minutes in MacOS, while all others jobs (Ubuntu/Windows/MacOs) take less than 30 minutes, so we end up in fact always waiting for this one. This divergence is pretty clear here:

image

Blue: `arviz-compatibility.yml`
Orange: `pytest.yml` (includes all tests in `arviz-compatibility.yml` and more)

Obtained from https://share.streamlit.io/koaning/gitlit/main

We could consider replacing this test by something else, perhaps a more `CI-like` suite of tests as opposed to this number crunching one, which is already covered in `pytest.yml` (Ubuntu 18.04 only).

ricardoV94:
The macOS machines in Github actions seem incredibly slow (see #5410), so I am proposing to remove the computationally heavier tests in favor of something more integration-like (which should also be more relevant for the "arviz_compat" part).

Currently this is a draft to check what the new Runtime might look like.

***

Chose a set of tests that are more related to sampling + arviz, while dropping the slow test_distributions and test_distributions_random that have nothing to do with arviz

Closes #5410

### DIFF
--- a/.github/workflows/arviz_compat.yml
+++ b/.github/workflows/arviz_compat.yml
@@ -20,11 +20,18 @@ jobs:
         os: [ubuntu-latest, macos-latest]
         floatx: [float64]
         test-subset:
-          - pymc/tests/test_distributions.py
+          - |
+            pymc/tests/test_parallel_sampling.py
+            pymc/tests/test_posteriors.py
+            pymc/tests/test_sampling.py
 
           - |
-            pymc/tests/test_distributions_random.py
-            pymc/tests/test_sampling.py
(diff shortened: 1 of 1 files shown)


ricardoV94:
Runtime is now below 25 minutes for the slowest arviz_compat workflow, down from the previous ~50 minutes. If people are happy with the included tests we can merge

michaelosthege (approved):

PR #5474, opened 2022-02-15
Upgrade Aesara version pin and unpin SciPy upper limit

### LINKED ISSUE #5446: Refactor Multivariate distributions due to new meaning of size
The next Aesara release will include https://github.com/aesara-devs/aesara/pull/446 which altered the meaning of `size` for multivariate distributions, to be more consistent with that of univariate distributions.

Size will now specify the number of replications + dimensions implied by batched parameters. It is equivalent to `shape[:-ndim.supp]`. We will likely need to update our multivariate distributions and/or some tests (e.g. in [test_distributions_random.py](https://github.com/pymc-devs/pymc/blob/1005d20b3c12d9b9a424c069f6a0f9962d73c41d/pymc/tests/test_distributions_random.py))

These changes should also be incorporated in https://github.com/pymc-devs/pymc/pull/5437

### LINKED ISSUE #5448: Remove pin of SciPy 1.8.0 once we run Aesara 2.4.0

michaelosthege:
I'm opening the PR so the CI tells me which tests break.

Closes #5446
Closes #5448

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -19,7 +19,7 @@ import warnings
 
 from abc import ABCMeta
 from functools import singledispatch
-from typing import Callable, Iterable, Optional, Sequence, Tuple, Union
+from typing import Callable, Iterable, Optional, Sequence, Tuple, Union, cast
 
 import aesara
 import numpy as np
@@ -45,7 +45,6 @@ from pymc.distributions.shape_utils import (
     convert_shape,
     convert_size,
     find_size,
-    maybe_resize,
     resize_from_dims,
     resize_from_observed,
 )
@@ -353,17 +352,11 @@ class Distribution(metaclass=DistributionMeta):
         # Create the RV with a `size` right away.
         # This is not necessarily the final result.
         rv_out = cls.rv_op(*dist_params, size=create_size, **kwargs)
-        rv_out = maybe_resize(
-            rv_out,
-            cls.rv_op,
-            dist_params,
-            ndim_expected,
-            ndim_batch,
-            ndim_supp,
-            shape,
-            size,
-            **kwargs,
-        )
+
+        # Replicate dimensions may be prepended via a shape with Ellipsis as the last element:
+        if shape is not None and Ellipsis in shape:
+            replicate_shape = cast(StrongShape, shape[:-1])
+            rv_out = change_rv_size(rv_var=rv_out, new_size=replicate_shape, expand=True)
 
         rng = kwargs.pop("rng", None)
         if (
@@ -589,18 +582,11 @@ class SymbolicDistribution:
         # Create the RV with a `size` right away.
         # This is not necessarily the final result.
         graph = cls.rv_op(*dist_params, size=create_size, **kwargs)
-        graph = maybe_resize(
-            graph,
-            cls.rv_op,
-            dist_params,
-            ndim_expected,
-            ndim_batch,
-            ndim_supp,
-            shape,
-            size,
-            change_rv_size_fn=cls.change_size,
-            **kwargs,
-        )
+
+        # Replicate dimensions may be prepended via a shape with Ellipsis as the last element:
+        if shape is not None and Ellipsis in shape:
+            replicate_shape = cast(StrongShape, shape[:-1])
+            graph = cls.change_size(rv=graph, new_size=replicate_shape, expand=True)
 
         rngs = kwargs.pop("rngs", None)
         if rngs is not None:
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -905,7 +905,7 @@ class WishartRV(RandomVariable):
 
     def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
         # The shape of second parameter `V` defines the shape of the output.
-        return dist_params[1].shape
+        return dist_params[1].shape[-2:]
 
     @classmethod
     def rng_fn(cls, rng, nu, V, size):
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -247,7 +247,7 @@ class MvNormal(Continuous):
     def get_moment(rv, size, mu, cov):
         moment = mu
         if not rv_size_is_none(size):
-            moment_size = at.concatenate([size, mu.shape])
+            moment_size = at.concatenate([size, [mu.shape[-1]]])
             moment = at.full(moment_size, mu)
         return moment
 
@@ -301,18 +301,13 @@ class MvStudentTRV(RandomVariable):
     @classmethod
     def rng_fn(cls, rng, nu, mu, cov, size):
 
-        # Don't reassign broadcasted cov, since MvNormal expects two dimensional cov only.
-        mu, _ = broadcast_params([mu, cov], cls.ndims_params[1:])
-
-        chi2_samples = np.sqrt(rng.chisquare(nu, size=size) / nu)
-        # Add distribution shape to chi2 samples
-        chi2_samples = chi2_samples.reshape(chi2_samples.shape + (1,) * len(mu.shape))
-
         mv_samples = multivariate_normal.rng_fn(rng=rng, mean=np.zeros_like(mu), cov=cov, size=size)
 
-        size = tuple(size or ())
+        # Take chi2 draws and add an axis of length 1 to the right for correct broadcasting below
+        chi2_samples = np.sqrt(rng.chisquare(nu, size=size) / nu)[..., None]
+
         if size:
-            mu = np.broadcast_to(mu, size + mu.shape)
+            mu = np.broadcast_to(mu, size + (mu.shape[-1],))
 
         return (mv_samples / chi2_samples) + mu
 
@@ -379,7 +374,7 @@ class MvStudentT(Continuous):
     def get_moment(rv, size, nu, mu, cov):
         moment = mu
         if not rv_size_is_none(size):
-            moment_size = at.concatenate([size, moment.shape])
+            moment_size = at.concatenate([size, [mu.shape[-1]]])
             moment = at.full(moment_size, moment)
         return moment
 
@@ -453,7 +448,7 @@ class Dirichlet(Continuous):
         norm_constant = at.sum(a, axis=-1)[..., None]
         moment = a / norm_constant
         if not rv_size_is_none(size):
-            moment = at.full(at.concatenate([size, a.shape]), moment)
+            moment = at.full(at.concatenate([size, [a.shape[-1]]]), moment)
         return moment
 
     def logp(value, a):
@@ -497,8 +492,8 @@ class MultinomialRV(MultinomialRV):
             size = tuple(size or ())
 
             if size:
-                n = np.broadcast_to(n, size + n.shape)
-                p = np.broadcast_to(p, size + p.shape)
+                n = np.broadcast_to(n, size)
+                p = np.broadcast_to(p, size + (p.shape[-1],))
 
             res = np.empty(p.shape)
             for idx in np.ndindex(p.shape[:-1]):
@@ -571,7 +566,7 @@ class Multinomial(Discrete):
         inc_bool_arr = at.abs_(diff) > 0
         mode = at.inc_subtensor(mode[inc_bool_arr.nonzero()], diff[inc_bool_arr.nonzero()])
         if not rv_size_is_none(size):
-            output_size = at.concatenate([size, p.shape])
+            output_size = at.concatenate([size, [p.shape[-1]]])
             mode = at.full(output_size, mode)
         return mode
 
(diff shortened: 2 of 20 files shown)


pymc/distributions/multivariate.py
```diff
@@ -247,8 +247,9 @@ def dist(cls, mu, cov=None, tau=None, chol=None, lower=True, **kwargs):
     def get_moment(rv, size, mu, cov):
         moment = mu
         if not rv_size_is_none(size):
-            moment_size = at.concatenate([size, mu.shape])
-            moment = at.full(moment_size, mu)
+            # MvNormal has ndim_supp == 1. Only the last entry is needed.
+            shape = at.concatenate([size, [mu.shape[-1]]])
+            moment = at.full(shape, mu)
```
michaelosthege: Not only the `get_moment`, but also the `rng_fn` need to get updated. This applies to at least the `MvNormal`, `MvStudentT`, `Multinomial` and `DirichletMultinomial`.

I'm not familiar enough with these distributions, so @Sayam753 @ricardoV94 @twiecki please check.

pymc/distributions/shape_utils.py
```diff
@@ -661,7 +662,9 @@ def maybe_resize(
     if shape is not None and ndims_unexpected:
         if Ellipsis in shape:
             # Resize and we're done!
-            rv_out = change_rv_size_fn(rv_var=rv_out, new_size=shape[:-1])
+            # Into added dims 👇 plus implied dims 👇 but excluding support 👇 dims.
+            new_size = (*shape[:-1], *tuple(rv_out.shape[:ndim_actual - ndim_supp]))
+            rv_out = change_rv_size_fn(rv_var=rv_out, new_size=new_size)
```
michaelosthege: The `else:` below may no longer be relevant, but we should first add type hints to the signature, because when I tried to un-indent the `ShapeError` it turned out that some of the `int` parameters are actually `Optional[int]`.

pymc/tests/test_distributions_moments.py
```diff
@@ -161,7 +161,8 @@ def assert_moment_is_expected(model, expected, check_finite_logp=True):
     except NotImplementedError:
         random_draw = moment
 
-    assert moment.shape == expected.shape == random_draw.shape
+    assert moment.shape == expected.shape
+    assert expected.shape == random_draw.shape
```
michaelosthege, line 165: I split the `a == b == c` into two asserts to make the errors easier to understand.

pymc/tests/test_distributions_moments.py
```diff
@@ -1010,8 +1011,8 @@ def test_moyal_moment(mu, sigma, size, expected):
         (2, rand1d, np.eye(2), 2, np.full((2, 2), rand1d)),
         (2, rand1d, np.eye(2), (2, 5), np.full((2, 5, 2), rand1d)),
         (2, rand2d, np.eye(3), None, rand2d),
-        (2, rand2d, np.eye(3), 2, np.full((2, 2, 3), rand2d)),
-        (2, rand2d, np.eye(3), (2, 5), np.full((2, 5, 2, 3), rand2d)),
+        (2, rand2d, np.eye(3), 2, np.full((2, 3), rand2d)),
+        (2, rand2d, np.eye(3), (2, 5, 2), np.full((2, 5, 2, 3), rand2d)),
```
michaelosthege: These two used to assert the inconsistent interpretation where size was concatenated in front of all other dimensions for multivariate RVs.

pymc/tests/test_shape_handling.py
```diff
@@ -323,7 +323,7 @@ def test_simultaneous_size_and_dims(self, with_dims_ellipsis):
             assert "ddata" in pmodel.dim_lengths
 
             # Size does not include support dims, so this test must use a dist with support dims.
-            kwargs = dict(name="y", size=2, mu=at.ones((3, 4)), cov=at.eye(4))
+            kwargs = dict(name="y", size=(2, 3), mu=at.ones((3, 4)), cov=at.eye(4))
```
michaelosthege, line 325: The `mu` creates an implied dimension which is now required to be included in `size`.

michaelosthege (commented):
I ran only the tests that were red, so maybe I broke something. (We'll see.)

But at least the following tests are still failing, because the RV implementations need to be fixed:
```
FAILED pymc/tests/test_distributions.py::TestMatchesScipy::test_dirichlet_vectorized[2-a2] - ValueError: input operand has more dimensions than allowed by the axis remapping
FAILED pymc/tests/test_distributions.py::TestMatchesScipy::test_dirichlet_vectorized[size1-a2] - ValueError: shape mismatch: objects cannot be broadcast to a single shape
FAILED pymc/tests/test_distributions.py::TestMatchesScipy::test_dirichlet_vectorized[size2-a1] - ValueError: shape mismatch: objects cannot be broadcast to a single shape
FAILED pymc/tests/test_distributions.py::TestMatchesScipy::test_dirichlet_vectorized[size2-a2] - ValueError: shape mismatch: objects cannot be broadcast to a single shape
FAILED pymc/tests/test_distributions_moments.py::test_dirichlet_moment[a2-7-expected2] - TypeError: Cannot convert Type TensorType(float64, (None, None, None)) (of Variable Elemwise{true_div,no_inplace}.0) ...
FAILED pymc/tests/test_distributions_moments.py::test_dirichlet_moment[a3-size3-expected3] - TypeError: Cannot convert Type TensorType(float64, (None, None, None, None)) (of Variable Elemwise{true_div,no_in...
FAILED pymc/tests/test_distributions_moments.py::test_mv_normal_moment[mu7-cov7-size7-expected7] - ValueError: could not broadcast input array from shape (2,2) into shape (4,5,2)
FAILED pymc/tests/test_distributions_moments.py::test_mvstudentt_moment[2-mu5-cov5-2-expected5] - assert (2, 3) == (2, 2, 3)
FAILED pymc/tests/test_distributions_moments.py::test_mvstudentt_moment[2-mu6-cov6-size6-expected6] - ValueError: operands could not be broadcast together with shapes (2,5,2,3) (2,5,2,1,1)
FAILED pymc/tests/test_distributions_moments.py::test_multinomial_moment[p6-n6-2-expected6] - TypeError: Cannot convert Type TensorType(int64, (None, None, None)) (of Variable Elemwise{Cast{int64}}.0) into ...
FAILED pymc/tests/test_distributions_random.py::TestDirichletMultinomial_1D_n_2D_a::test_distribution - assert (1, 2, 4) == (1, 4)
```

For MvNormal and MvStudentT I think this is just the `rng_fn` (and `get_moment`).

-------------------
I gotta jump now, so anybody please feel free to take over.

OriolAbril:
You should also check https://pymc--5474.org.readthedocs.build/en/5474/learn/examples/pymc_overview.html#model-specification to make sure not only the installation finishes but that it is actually working and pymc can be imported and used (same url will work as new commits come in)

pymc/tests/test_shape_handling.py
```diff
@@ -438,13 +438,12 @@ def test_mvnormal_shape_size_difference(self):
...
             assert tuple(rv.shape.eval()) == (5, 4, 3)
 
-        # When using `size` the API behaves like Aesara/NumPy
-        with pytest.warns(
-            ShapeWarning,
-            match=r"You may have expected a \(2\+1\)-dimensional RV, but the resulting RV will be 5-dimensional",
-        ):
+        # The support dimensions are excluded from size. This is a difference w.r.t. NumPy
+        # and from 2021-05 through 2022-02 the behavior here matched NumPy but warned about
+        # potentially unexpected results.
```
ricardoV94: I find this comment confusing. The support dimensions being excluded from size is the official Numpy API, see size explanation for multivariate normal: https://numpy.org/doc/stable/reference/random/generated/numpy.random.multivariate_normal.html

I would also remove the from x to y behavior...

michaelosthege: I left it as an explanation for why there's a `pytest.warns(None)` below. But I can take it out

ricardoV94: Ah... I think I would take it out yes

pymc/distributions/multivariate.py
```diff
@@ -301,18 +301,13 @@ def _shape_from_params(self, dist_params, rep_param_idx=1, param_shapes=None):
...
-
-        chi2_samples = np.sqrt(rng.chisquare(nu, size=size) / nu)
-        # Add distribution shape to chi2 samples
-        chi2_samples = chi2_samples.reshape(chi2_samples.shape + (1,) * len(mu.shape))
-
         mv_samples = multivariate_normal.rng_fn(rng=rng, mean=np.zeros_like(mu), cov=cov, size=size)
 
-        size = tuple(size or ())
+        # Take chi2 draws and add an axis of length 1 to the right for correct broadcasting below
+        chi2_samples = np.sqrt(rng.chisquare(nu, size=size) / nu)[..., None]
```
ricardoV94, line 307: @Sayam753 can you have a second look if the new Multivariate rng_fn looks correct?

ricardoV94:
Tests are passing, and I doubled check the following suspicious (initially) non-failing distributions:
- [x] MatrixNormal
- [x] KroneckerNormal
- [x] CAR

pymc/distributions/shape_utils.py
michaelosthege: @ricardoV94 it looks like this whole block in `shape_utils.py` is no longer covered by any tests. It is [covered on `main` (line 661)](https://app.codecov.io/gh/pymc-devs/pymc/blob/main/pymc/distributions/shape_utils.py).
Can we remove it, or is there an edge case that could still trigger these branches?

grafik

ricardoV94: >  is there an edge case that could still trigger these branches?

None that I know of

pymc/util.py
```diff
@@ -14,7 +14,7 @@
 
 import functools
 
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List, Tuple, Union, cast
```
michaelosthege, line 17: This was a missing import I noticed when running `mypy -p pymc`

pymc/distributions/shape_utils.py
```diff
@@ -565,26 +569,30 @@ def resize_from_observed(
...
+) -> Tuple[Optional[StrongSize], Optional[int], Optional[int], int]:
     """Determines the size keyword argument for creating a Distribution.
 
     Parameters
     ----------
-    shape : tuple
+    shape
         A tuple specifying the final shape of a distribution
-    size : tuple
+    size
```
michaelosthege, line 578: These hints were inaccurate and I didn't want to add the fully expanded meaning of `Optional[WeakShape]` etc.

michaelosthege (commented):
With the last commit `mypy -p pymc` has zero complaints about `shape_utils.py` :)

pymc/distributions/shape_utils.py
michaelosthege: Looks like I messed up the ShapeWarning. I'll take a look at it in the morning.

pymc/distributions/distribution.py
```diff
@@ -353,17 +352,11 @@ def dist(
...
-            ndim_expected,
-            ndim_batch,
-            ndim_supp,
-            shape,
-            size,
-            **kwargs,
-        )
+
+        # Replicate dimensions may be prepended via a shape with Ellipsis as the last element:
+        if shape is not None and Ellipsis in shape:
```
ricardoV94, line 357: do we check anywhere that the ellipsis is in the last position?

michaelosthege: Yes, there are functions further above that do this.
Also that's the difference between `WeakShape` and `StrongShape` so mypy also tells us about errors.
Actually I ran mypy before pushing and it was satisfied.

If only I had run pre-commit the right way. It's still messing with the environment YAML files every time - we have some kind of line ending problem in the hook. So it's always failing for me and sometimes I overlook black problems because of that.

pymc/distributions/distribution.py
```diff
@@ -589,18 +582,11 @@ def dist(
...
-            shape,
-            size,
-            change_rv_size_fn=cls.change_size,
-            **kwargs,
-        )
+
+        # Replicate dimensions may be prepended via a shape with Ellipsis as the last element:
+        if shape is not None and Ellipsis in shape:
+            replicate_shape = cast(StrongShape, shape[:-1])
+            graph = cls.change_size(rv_var=graph, new_size=replicate_shape, expand=True)
```
ricardoV94: I don't think I've implemented the expand kwarg in censored... but should

```suggestion
            graph = cls.change_size(rv_var=graph, new_size=replicate_shape)
```

ricardoV94: https://github.com/pymc-devs/pymc/blob/a3bab7ddc0f4612ad039a7d2755f22fa6c8c0a72/pymc/distributions/censored.py#L103

michaelosthege: is it gonna work at all then?

ricardoV94: right now it's working if not "expanding". We can put a `NotImplementedError` if expand is `True` in `Censored.change_size` for the time being

michaelosthege: You wanna do that commit?
I gotta go run t the supermarket before the vegetables are gone 🙂

ricardoV94: Fixed

ricardoV94 (approved):

PR #5476, opened 2022-02-15
Generalize multinomial moment to arbitrary dimensions

### LINKED ISSUE #5393: Generalize multinomial moment to arbitrary dimensions
The multinomial distribution is not restricted to 2D, but it's moment assumes this is the case. This should be similar to https://github.com/pymc-devs/pymc/pull/5225

markvrma:
Resolves #5393 


pymc/distributions/multivariate.py
```diff
@@ -565,13 +565,14 @@ def dist(cls, n, p, *args, **kwargs):
     def get_moment(rv, size, n, p):
         if p.ndim > 1:
             n = at.shape_padright(n)
-        if (p.ndim == 1) & (n.ndim > 0):
+        if (p.ndim == 1) and (n.ndim > 0):
             n = at.shape_padright(n)
             p = at.shape_padleft(p)
```
ricardoV94: I think we shouldn't need any these shape_padright/left anymore

ricardoV94 (changes requested):
Hi @markvrma thanks for opening a PR!

Besides the comment below, we also need to add test conditions to make sure the moment is indeed correct for higher dimensions. This PR is a good template: https://github.com/pymc-devs/pymc/pull/5225/files

pymc/distributions/multivariate.py
markvrma: I tried removing:
```py
 if p.ndim > 1:
            n = at.shape_padright(n)
 if (p.ndim == 1) and (n.ndim > 0):
            n = at.shape_padright(n)
            p = at.shape_padleft(p)
```

but existing tests fail in `test_distributions_moments.py`

```
FAILED test_distributions_moments.py::test_multinomial_moment[p4-n4-None-expected4]
FAILED test_distributions_moments.py::test_multinomial_moment[p5-n5-None-expected5]
FAILED test_distributions_moments.py::test_multinomial_moment[p6-n6-2-expected6]
```

markvrma:
Added new test for higher dimensions.

I couldn't remove:
```py
 if p.ndim > 1:
            n = at.shape_padright(n)
 if (p.ndim == 1) and (n.ndim > 0):
            n = at.shape_padright(n)
            p = at.shape_padleft(p)
```

as existing tests fail in `test_distributions_moments.py`.

ricardoV94:
> Added new test for higher dimensions.
> 
> I couldn't remove:
> 
> ```python
>  if p.ndim > 1:
>             n = at.shape_padright(n)
>  if (p.ndim == 1) and (n.ndim > 0):
>             n = at.shape_padright(n)
>             p = at.shape_padleft(p)
> ```
> 
> as existing tests fail in `test_distributions_moments.py`.

We might need something else, instead of just removing these lines. Or perhaps those lines also give the right result for higher dimensions that are now supported, but I would be surprised if that was the case, since they seem to have specialized logic for <= 2D parameters. 

I would brute-force with local tests where we try different combinations of `p` and `n` with shapes ranging from scalar to 3D and see if this function returns the correct output. One way might be to compare with a `np.vectorized` version of `get_moment` that works for the base case (1d vector of `p` and scalar `n`). There are some examples where we use this approach for testing the `logp` here: 

https://github.com/pymc-devs/pymc/blob/c892317aba0663df0d250c1813b2ca2fe3460774/pymc/tests/test_distributions.py#L2228-L2251

and here:

https://github.com/pymc-devs/pymc/blob/c892317aba0663df0d250c1813b2ca2fe3460774/pymc/tests/test_distributions.py#L335-L347

https://github.com/pymc-devs/pymc/blob/c892317aba0663df0d250c1813b2ca2fe3460774/pymc/tests/test_distributions.py#L2297-L2319

ricardoV94:
Also did you figure out what exactly is failing? Looking at the DirichletMultinomial mode it seems like it should be exactly the same, except that here we already have the `p`, and can skip the first step of `p = a / at.sum(a, axis=-1)` in https://github.com/pymc-devs/pymc/pull/5225/files Unless there's an error in the implementation of the DirichletMultinomial moment as well

ricardoV94:
@markvrma It seems than it was enough to add one dimension to the right of `n`. The old and new tests were quite helpful to figure this out, and also show that the `DirichletMultinomial` moment would fail for some configurations.

Thanks for the help!

ricardoV94 (approved):

PR #5485, opened 2022-02-18
Fix numpydoc in pm.Data, pm.MutableData, pm.ConstantData

jmloyola:
Addresses #5459.

This PR:
- Add space between colon in parameters
- Fix link to Model.set_data function in documentation
- Fix docstring parameters type
- Fix link to example notebook
- Move information to extended description
- Add punctuation to parameter description
- Move parameter description to respect parameters order
- Add double ticks for code references in docstrings
- Link function in pymc
- Add default behaviour of optional parameters
- Add reference to pm.Data in pm.ConstantData and pm.MutableData

The `value` parameter still references pandas objects (`pd.Series`, `pd.Dataframe`).
How should we handle them?

On the other hand, I think the [reference notebook](https://docs.pymc.io/projects/examples/en/latest/pymc3_howto/data_container.html) remains to be translated to the new version. I don't know if we should leave a comment to remember to change it once it is updated or not.

#DataUmbrellaPyMCSprint

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -577,7 +577,7 @@ def Data(
     ----------
     name : str
         The name for this variable.
-    value : array_like or pd.Series, pd.Dataframe
+    value : array_like or pandas.Series, pandas.Dataframe
         A value to associate with this variable.
     dims : str or tuple of str, optional
         Dimension names of the random variables (as opposed to the shapes of these
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -614,7 +614,7 @@ def Data(
     :func:`pymc.Model.set_data`.
 
     For more information, take a look at this example notebook
-    https://docs.pymc.io/notebooks/data_container.html
+    https://docs.pymc.io/projects/examples/en/latest/pymc3_howto/data_container.html
     """
     if isinstance(value, list):
         value = np.array(value)
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -571,8 +571,7 @@ def Data(
     To set the value of the data container variable, check out
     :func:`pymc.Model.set_data`.
 
-    For more information, take a look at this example notebook
-    https://docs.pymc.io/projects/examples/en/latest/pymc3_howto/data_container.html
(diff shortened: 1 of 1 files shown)


pymc/data.py
```diff
@@ -562,35 +564,47 @@ def Data(
...
-    in value and shape, but NOT in dimensionality using ``pm.set_data()``.
+    is registered as a :class:`~aesara.compile.sharedvalue.SharedVariable`,
+    enabling it to be altered in value and shape, but NOT in dimensionality using
+    :func:`pymc.set_data`.
+
+    To set the value of the data container variable, check out
+    :func:`pymc.Model.set_data`.
+
+    For more information, take a look at this example notebook
+    https://docs.pymc.io/projects/examples/en/latest/pymc3_howto/data_container.html
```
OriolAbril: We should not use html links to refer to pymc (or other python libraries commonly used like arviz, aesara, numpy...). This should be done with intersphinx:

```
:ref:`nb:data_container`
```

Which will show the title of the notebook as link text, so it's probably best to update the text something like "read the <link here> notebook"

pymc/data.py
```diff
@@ -562,35 +564,47 @@ def Data(
...
-        A value to associate with this variable
-    mutable : bool, optional
-        Switches between creating a ``SharedVariable`` (``mutable=True``, default)
-        vs. creating a ``TensorConstant`` (``mutable=False``).
-        Consider using ``pm.ConstantData`` or ``pm.MutableData`` as less verbose
-        alternatives to ``pm.Data(..., mutable=...)``.
-    dims: {str, tuple of str}, optional, default=None
+    name : str
+        The name for this variable.
+    value : list or ndarray or pd.Series, pd.Dataframe
```
OriolAbril:
```suggestion
    value : array_like or pd.Series, pd.Dataframe
```

OriolAbril (changes requested):

OriolAbril (approved):

PR #5487, opened 2022-02-18
Updated pymc.Normal docstring

sagartomar:

For issue #5459 

#DataUmbrellaPyMCSprint

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -525,12 +525,12 @@ class Normal(Continuous):
 
     Parameters
     ----------
-    mu : float or array_like or TensorVariable, default 0
+    mu : tensor_like of float, default 0
         Mean.
-    sigma : float or array_like or TensorVariable, optional
+    sigma : tensor_like of float, optional
         Standard deviation (sigma > 0) (only required if tau is not specified).
         Defaults to 1 if neither sigma nor tau is specified.
-    tau : float or array_like or TensorVariable, optional
+    tau : tensor_like of float, optional
         Precision (tau > 0) (only required if sigma is not specified).
         Defaults to 1 if neither sigma nor tau is specified.
 
@@ -576,12 +576,12 @@ class Normal(Continuous):
 
         Parameters
         ----------
-        value : float or ndarray or TensorVariable
+        value : tensor_like of float
             Value(s) for which log CDF is calculated. If the log CDF for multiple
             values are desired the values must be provided in a numpy array or `TensorVariable`.
-        mu : float or array_like or TensorVariable
-            Mean. Defaults to 0.
-        sigma : float or array_like or TensorVariable
+        mu : tensor_like of float
+            Mean.
+        sigma : tensor_like of float
             Standard deviation (sigma > 0).
 
         Returns
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -525,12 +525,14 @@ class Normal(Continuous):
 
     Parameters
     ----------
-    mu : float
-        Mean. Defaults to 0.
-    sigma : float
+    mu : float or array_like or TensorVariable, default 0
+        Mean.
+    sigma : float or array_like or TensorVariable, optional
         Standard deviation (sigma > 0) (only required if tau is not specified).
-    tau : float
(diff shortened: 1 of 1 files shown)


pymc/distributions/continuous.py
```diff
@@ -524,11 +525,11 @@ class Normal(Continuous):
 
     Parameters
     ----------
-    mu: float
-        Mean.
-    sigma: float
+    mu : float
+        Mean. Defaults to 0.
```
OriolAbril: the default should go in the above line here as it's of the documented type and fixed. reference https://pymc-data-umbrella.xyz/en/latest/sprint/docstring_tutorial.html#parameters

pymc/distributions/continuous.py
```diff
@@ -573,9 +574,13 @@ def logcdf(value, mu, sigma):
...
         Parameters
         ----------
-        value: numeric or np.ndarray or `TensorVariable`
+        value : numeric or ndarray or TensorVariable
             Value(s) for which log CDF is calculated. If the log CDF for multiple
             values are desired the values must be provided in a numpy array or `TensorVariable`.
+        mu : float
+            Mean. Defaults to 0.
+        sigma : float
+            Standard deviation (sigma > 0).
```
Sayam753: I think `mu` and `sigma` parameters can be numpy array/tensor variables as well.

pymc/distributions/continuous.py
```diff
@@ -524,11 +525,11 @@ class Normal(Continuous):
 
     Parameters
     ----------
-    mu: float
-        Mean.
-    sigma: float
+    mu : float
+        Mean. Defaults to 0.
+    sigma : float
```
```suggestion
    sigma : float, optional
```

here the default is not immediate as it depends on the value of tau too, so it should use optional only with the default on the description. In this case the description should mention that when no value is passed for tau nor sigma, sigma defaults to one

pymc/distributions/continuous.py
```diff
@@ -524,11 +525,11 @@ class Normal(Continuous):
...
         Standard deviation (sigma > 0) (only required if tau is not specified).
-    tau: float
+    tau : float
```
OriolAbril: same, optional here

pymc/distributions/continuous.py
```diff
@@ -524,11 +525,11 @@ class Normal(Continuous):
 
     Parameters
     ----------
-    mu: float
-        Mean.
-    sigma: float
+    mu : float
```
OriolAbril: I think the type should be `numeric, array_like or TensorVariable` cc @Sayam753 @twiecki @michaelosthege

pymc/distributions/continuous.py
ricardoV94: I think we only mention the base case

ricardoV94: numpy uses float or array_like of floats: https://numpy.org/doc/stable/reference/random/generated/numpy.random.normal.html

OriolAbril: I have a similar comment below, missed this one. I'd propose something like `float, array_like or TensorVariable` not sure if we need to add `of float` to both array and tensor types :thinking:

sagartomar: `float, array_like or TensorVariable` looks good to me. Should I update the pull request with this?

OriolAbril: Use `tensor_like of float`. We will update the glossary to have a definition for the type like numpy does with array_like.

pymc/distributions/continuous.py
```diff
@@ -524,12 +525,14 @@ class Normal(Continuous):
...
+    mu : tensor_like of float, default 0
         Mean.
-    sigma: float
+    sigma : tensor_like of float, optional
         Standard deviation (sigma > 0) (only required if tau is not specified).
-    tau: float
+        Defaults to 1 if neither sigma nor tau is specified.
+    tau : tensor_like of float, optional
         Precision (tau > 0) (only required if sigma is not specified).
+        Defaults to 1 if neither sigma nor tau is specified.
```
```suggestion
```

only one of tau or sigma are used, I think it's better to leave the explanation as no value for tau or sigma -> sigma=1 (tau=1 is equivalent to sigma=1 but that might still be confusing for many pymc users)

OriolAbril (approved):

Issue #5308, opened 2022-01-03
Refactor old Distribution base class


ricardoV94:
PyMC distribution classes are weird objects that hold RandomVariables, logp, logcdf and moment methods together (basically doing runtime dispatching) and manage most of the non-RandomVariable kwargs that users are familiar with (observed, transformed, size/dims) and behind the scenes actions like registration in the model.

This exists mostly for backwards compatibility with V3 and ease of developer refactoring, but the current result is far from pretty.

We need to figure out a more elegant/permanent architecture now that many things that existed to accommodate V3 limitations no longer hold.

## Distribution
`Distribution` is currently performing the following tasks:

https://github.com/pymc-devs/pymc/blob/75ea2a80cb27773e93b7b207043077953940e6ff/pymc/distributions/distribution.py#L135

1. Input validation:
    1. Raising `FutureWarnings` for `testval` kwarg
    1. Raising TypeError when distribution is initialized outside of a Model context
    1. Raising TypeError when name is not given to a distribution
    1. Raising ValueError when more than one of dims/shape/size is given
1. Convert alternative parametrizations to standard parametrization (e.g, `tau` -> `sigma`). This is done by the `.dist` methods.
1. Add informative attribute errors for deprecated `logp`, `logcdf`, `random` methods
1. Resize the final RV based on observed, shape, dims or size
1. Provides the `.dist()` API to create an unnamed RV that is not registered in the model. This type of variables is necessary for use in Potentials and other distribution factories that use RVs as building blocks such as Bound and Censored distributions, as well as Mixtures and Timeseries once they get refactored for V4

## DistributionMeta
In addition we have a `DistributionMeta` that does the following:

https://github.com/pymc-devs/pymc/blob/75ea2a80cb27773e93b7b207043077953940e6ff/pymc/distributions/distribution.py#L70

1. Dispatch the `logp`, `logcdf`, `moment`, `default_transform` methods defined in the old PyMC distributions to apply to the respective `rv_op`
1. Register the `rv_op` type as subclass of the old style PyMC distribution, so that V3 Discrete/Continuous subclass checks still work?

```python
isinstance(pm.Normal.dist().owner.op, pm.distributions.Continuous)  # True
```

[...]

PR #5488, opened 2022-02-18
Updates to pymc.Hypergeometric docstrings

cuchoi:
As part of the work of #5459:

- Added the bounds to the parameters N, k, and n in the parameters section of pymc.Hypergeometric.
- Added docstrings for parameters "good" and "bad" in `logp` and `logcdf`. These parameters are different than the ones used in `__init__`, which is odd to me. They transform the `N` and `k` parameters to `bad` (which is `N - k`) and `good` (which is `k`), but this transformation could happen inside `logp` and `logcdf`. Would it make sense to always use N, k? 

#DataUmbrellaPyMCSprint

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -971,11 +971,11 @@ class HyperGeometric(Discrete):
 
     Parameters
     ----------
-    N : integer
+    N : integer, array_like or TensorVariable
         Total size of the population (N > 0)
-    k : integer
+    k : integer, array_like or TensorVariable
         Number of successful individuals in the population (0 <= k <= N)
-    n : integer
+    n : integer, array_like or TensorVariable
         Number of samples drawn from the population (0 <= n <= N)
     """
 
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -972,11 +972,11 @@ class HyperGeometric(Discrete):
     Parameters
     ----------
     N : integer
-        Total size of the population
+        Total size of the population (N > 0)
     k : integer
-        Number of successful individuals in the population
+        Number of successful individuals in the population (0 <= k <= N)
     n : integer
-        Number of samples drawn from the population
+        Number of samples drawn from the population (0 <= n <= N)
     """
 
     rv_op = hypergeometric
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -1004,6 +1004,10 @@ class HyperGeometric(Discrete):
         value : numeric
             Value(s) for which log-probability is calculated. If the log probabilities for multiple
             values are desired the values must be provided in a numpy array or Aesara tensor
+        good : integer
+            Number of successful individuals in the population. Alias for parameter :math:`k`.
+        bad : integer
+            Number of unsuccessful individuals in the population. Alias for :math:`N-k`.
 
         Returns
         -------
@@ -1042,8 +1046,14 @@ class HyperGeometric(Discrete):
 
         Parameters
         ----------
-        value: numeric
+        value : numeric
             Value for which log CDF is calculated.
+        good : integer
+            Number of successful individuals in the population. Alias for parameter :math:`k`.
+        bad : integer
+            Number of unsuccessful individuals in the population. Alias for :math:`N-k`.
+        n : integer
+            Number of samples drawn from the population (0 <= n <= N)
 
         Returns
         -------
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -971,11 +971,11 @@ class HyperGeometric(Discrete):
 
     Parameters
     ----------
-    N : integer, array_like or TensorVariable
+    N : tensor_like of integer
         Total size of the population (N > 0)
-    k : integer, array_like or TensorVariable
+    k : tensor_like of integer
         Number of successful individuals in the population (0 <= k <= N)
-    n : integer, array_like or TensorVariable
+    n : tensor_like of integer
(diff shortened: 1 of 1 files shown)


pymc/distributions/discrete.py
```diff
@@ -972,11 +972,11 @@ class HyperGeometric(Discrete):
     Parameters
     ----------
     N : integer
```
OriolAbril: are scalars the only valid type for all 3 arguments? cc @ricardoV94 ? (no idea who knows, feel free to tag someone else)

OriolAbril:
@cuchoi did you mean to close the PR? maybe open another not from `main`?

pymc/distributions/discrete.py
cuchoi: They should be only integers: https://en.wikipedia.org/wiki/Hypergeometric_distribution

OriolAbril: but maybe an array of integers is also supported? :thinking:

cuchoi:
I closed the PR while adding more changes to it to avoid it being reviewed before adding them. 

Should I change the PR to be from a branch of my forked repo?

pymc/distributions/discrete.py
cuchoi: logcdf does say:
```
        # logcdf can only handle scalar values at the moment
        if np.ndim(value):
            raise TypeError(
                f"HyperGeometric.logcdf expects a scalar value but received a {np.ndim(value)}-dimensional object."
            )
```

cuchoi: But maybe logp supports it?

ricardoV94: Arrays/ tensors should always assumed to be supported unless otherwise specified (as in the logcdf) or some multivariate distributions

cuchoi:
Added docstrings for parameters "good" and "bad" in logp and logcdf. These parameters are different than the ones used in __init__, which is odd to me. They transform the N and k parameters to bad (which is N - k) and good (which is k), but this transformation could happen inside logp and logcdf. Would it make sense to always use N, k?

pymc/distributions/discrete.py
OriolAbril: Then I'd use something like 

```
integer, array_like or TensorVariable
scalar, array_like or TensorVariable 
integer, array_like of integer or TensorVariable of integer
```

The second option would explain in the description that only integers are allowed (like it does with the >0 constraint)

cuchoi: It does support numpy arrays. Will add it to the docstrings.

Running
```
In [5]: import pymc as pm
   ...: import arviz
   ...: import numpy as np
   ...: with pm.Model() as model:
   ...:     hyp = pm.HyperGeometric('hyp', np.array([50, 50]), np.array([1, 30]), np.array([30, 30]
   ...: ))
   ...:     samples = pm.sample()
  ```
  
  Get me this 
image

Comment (cuchoi) [reply pymc/distributions/discrete.py]: Can I use a different format to specify the array types? Example `integer, array_like[integer] or TensorVariable[integer]`?

Comment (OriolAbril) [reply pymc/distributions/discrete.py]: no, numpydoc formatting is very clear in the `array of type` format

Comment (OriolAbril) [reply pymc/distributions/discrete.py]: We decided to use `tensor_like of integer` and we'll add its definition on the glossary so we have a type that covers scalars, arrays, aesara tensorvariable and randomvariable

Comment (cuchoi) [reply pymc/distributions/discrete.py]: Noted! Picked `integer, array_like or TensorVariable`

Comment (OriolAbril) [reply pymc/distributions/discrete.py]: Sorry for the late review. We decided to go with `tensor_like of integer` which will get rendered as a link to our glossary where we have an explanation on what inputs are accepted for arguments documented like this: https://docs.pymc.io/en/latest/glossary.html#term-tensor_like

File: pymc/distributions/discrete.py [hunk 2]
```diff
@@ -971,12 +971,12 @@ class HyperGeometric(Discrete):
...
-        Total size of the population
-    k : integer
-        Number of successful individuals in the population
-    n : integer
-        Number of samples drawn from the population
+    N : integer, array_like or TensorVariable
+        Total size of the population (N > 0)
+    k : integer, array_like or TensorVariable
+        Number of successful individuals in the population (0 <= k <= N)
+    n : integer, array_like or TensorVariable
```
OriolAbril:
```suggestion
    N : tensor_like of integer
        Total size of the population (N > 0)
    k : tensor_like of integer
        Number of successful individuals in the population (0 <= k <= N)
    n : tensor_like of integer
```

OriolAbril (approved):

pymc/distributions/discrete.py
```diff
@@ -1004,6 +1004,10 @@ def logp(value, good, bad, n):
         value : numeric
             Value(s) for which log-probability is calculated. If the log probabilities for multiple
             values are desired the values must be provided in a numpy array or Aesara tensor
+        good : integer, array_like or TensorVariable
+            Number of successful individuals in the population. Alias for parameter :math:`k`.
+        bad : integer, array_like or TensorVariable
+            Number of unsuccessful individuals in the population. Alias for :math:`N-k`.
```
cuchoi, line 1010: Should we also use "tensor_like of integer" here @OriolAbril ?

OriolAbril: We are in the process of removing the distribution methods from the documentation and leaving only the `dist` one. The Distribution class is still being refactored, see #5308 for some more details (I don't really understand what is going on very well) and those methods might be removed completely, and because they should not be used themselves _already_ but called via `pm.logp()` or `pm.logcdf`

I think it's best to leave them be for now.

PR #5490, opened 2022-02-18
adding nightly builds

### LINKED ISSUE #5144: Add support for nightly builds
Seems like that's quite straight-forward: https://github.com/aesara-devs/aesara/pull/656

theorashid:
This PR closes #5144. It adds support for building a "nightly" release, pushing it to the pymc-nightly PyPI package.

I have taken it mostly from [aesara-devs/aesara/pull/656](https://github.com/aesara-devs/aesara/pull/656), copying the relevant nightly.yml and setup.py files.

If the environment variable BUILD_PYMC_NIGHTLY is present, the name of the package in setup.py gets renamed and the VERSION variable is changed to return the current base version with the date appended.

Differences between aesara PR and questions:
- this is "pymc" not "aesara"
- pymc has it's own `get_version()` rather than using `versioneer`. The aesara versioneer code uses `versions["version"] = versions["version"].split("+")[0] + suffix` and I just use `+ suffix`, which in my environment gives `VERSION="4.0.0b2.dev20220218"`
- I added a test to `nightly.yml` for the nightly install. Aesara doesn't have this. No idea what `${GITHUB_REF:11}` means
- Are the secrets. correct?
- Am I okay using `python -m build --sdist` rather than `twine wheel`

There's no way of testing this... so check carefully

Cheers

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,10 +1,11 @@
-name: Nightly
+name: nightly
+
 on:
   schedule:
     - cron: "0 0 * * *"
 
 jobs:
-  build_and_publish:
+  build-and-publish-nightly:
     name: Build source distribution
     runs-on: ubuntu-latest
     steps:
@@ -28,7 +29,7 @@ jobs:
           user: __token__
           password: ${{ secrets.nightly_pypi_secret }}
   test-install-job:
-    needs: release-job
+    needs: build-and-publish-nightly
     runs-on: ubuntu-latest
     steps:
     - name: Set up Python
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,42 @@
+name: Nightly
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  build_and_publish:
+    name: Build source distribution
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip
+          python -m pip install build
+      - name: Build the sdist
+        run: python -m build --sdist .
+        env:
+          BUILD_PYMC_NIGHTLY: true
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@v1.5.0
+        with:
+          user: __token__
+          password: ${{ secrets.nightly_pypi_secret }}
+  test-install-job:
+    needs: release-job
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - name: Give PyPI a chance to update the index
+      run: sleep 240
+    - name: Install from PyPI
+      run: |
+        pip install pymc-nightly==${GITHUB_REF:11}
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -40,4 +40,4 @@ jobs:
       run: sleep 240
     - name: Install from PyPI
       run: |
-        pip install pymc-nightly==${GITHUB_REF:11}
+        pip install pymc-nightly==${GITHUB_REF:11}.dev$(date +"%Y%m%d")
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ AUTHOR = "PyMC Developers"
 AUTHOR_EMAIL = "pymc.devs@gmail.com"
 URL = "http://github.com/pymc-devs/pymc"
 LICENSE = "Apache License, Version 2.0"
+NIGHLTY = "BUILD_PYMC_NIGHTLY" in os.environ
 
 classifiers = [
     "Development Status :: 5 - Production/Stable",
@@ -81,12 +82,10 @@ def get_version(nightly_build=False):
     raise RuntimeError(f"Unable to find version in {version_file}.")
 
 
-nightly = "BUILD_PYMC_NIGHTLY" in os.environ
-
 if __name__ == "__main__":
     setup(
-        name=get_distname(nightly),
-        version=get_version(nightly),
+        name=get_distname(NIGHLTY),
+        version=get_version(NIGHLTY),
         maintainer=AUTHOR,
         maintainer_email=AUTHOR_EMAIL,
         description=DESCRIPTION,
--- a/setup.py
+++ b/setup.py
@@ -16,11 +16,11 @@ import os
 import re
 
 from codecs import open
+from datetime import datetime, timezone
 from os.path import dirname, join, realpath
 
 from setuptools import find_packages, setup
 
-DISTNAME = "pymc"
 DESCRIPTION = "Probabilistic Programming in Python: Bayesian Modeling and Probabilistic Machine Learning with Aesara"
 AUTHOR = "PyMC Developers"
 AUTHOR_EMAIL = "pymc.devs@gmail.com"
@@ -55,9 +55,17 @@ with open(REQUIREMENTS_FILE) as f:
 test_reqs = ["pytest", "pytest-cov"]
 
 
+def get_distname(nightly_build=False):
+    distname = "pymc"
+    if nightly_build:
+        distname = "{}{}".format(distname, "-nightly")
+
+    return distname
+
+
 def get_version(nightly_build=False):
-    VERSIONFILE = join("pymc", "__init__.py")
-    lines = open(VERSIONFILE).readlines()
+    version_file = join("pymc", "__init__.py")
+    lines = open(version_file).readlines()
     version_regex = r"^__version__ = ['\"]([^'\"]*)['\"]"
     for line in lines:
         mo = re.search(version_regex, line, re.M)
@@ -65,27 +73,20 @@ def get_version(nightly_build=False):
             version = mo.group(1)
 
             if nightly_build:
-                from datetime import datetime, timezone
-
                 suffix = datetime.now(timezone.utc).strftime(r".dev%Y%m%d")
-                version += suffix
+                version = f"{version}{suffix}"
 
             return version
 
-    raise RuntimeError(f"Unable to find version in {VERSIONFILE}.")
+    raise RuntimeError(f"Unable to find version in {version_file}.")
 
 
-if "BUILD_PYMC_NIGHTLY" in os.environ:
-    nightly = True
-    DISTNAME += "-nightly"
-    VERSION = get_version(nightly_build=True)
-else:
-    VERSION = get_version()
+nightly = "BUILD_PYMC_NIGHTLY" in os.environ
 
 if __name__ == "__main__":
     setup(
-        name=DISTNAME,
-        version=VERSION,
+        name=get_distname(nightly),
+        version=get_version(nightly),
         maintainer=AUTHOR,
         maintainer_email=AUTHOR_EMAIL,
         description=DESCRIPTION,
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -65,10 +66,26 @@ def get_version():
     raise RuntimeError(f"Unable to find version in {VERSIONFILE}.")
 
 
+VERSION = get_version()
```
canyon289: Can the function `get_version()`  be updated to take `BUILD_PYMC_NIGHTLY` as a parameter? Its  odd to have this secondary function that gets defined under an if block and a global thats overwritten

canyon289 (changes requested):

setup.py
theorashid: I agree. That was code I lazily copied across from aesara. I've improved this but not sure if I've exactly interpreted what you wanted

setup.py
```diff
@@ -54,21 +55,37 @@
 test_reqs = ["pytest", "pytest-cov"]
 
 
-def get_version():
+def get_version(nightly_build=False):
     VERSIONFILE = join("pymc", "__init__.py")
```
canyon289: `VERSIONFILE` should not be in caps since its not global

setup.py
```diff
@@ -54,21 +55,37 @@
...
     lines = open(VERSIONFILE).readlines()
     version_regex = r"^__version__ = ['\"]([^'\"]*)['\"]"
     for line in lines:
         mo = re.search(version_regex, line, re.M)
         if mo:
-            return mo.group(1)
+            version = mo.group(1)
+
+            if nightly_build:
+                from datetime import datetime, timezone
```
canyon289: imports should be at top of tile

setup.py
```diff
@@ -54,21 +55,37 @@
...
+
+                suffix = datetime.now(timezone.utc).strftime(r".dev%Y%m%d")
+                version += suffix
```
canyon289: use string interpolation, either `.format` or fstring not +=

setup.py
```diff
@@ -54,21 +55,37 @@
...
+
+            return version
+
     raise RuntimeError(f"Unable to find version in {VERSIONFILE}.")
 
 
+if "BUILD_PYMC_NIGHTLY" in os.environ:
```
canyon289: This whole logic block should be able to be replaced by `os.getenv` whose value is passed straight to `get_version`

https://docs.python.org/3/library/os.html#os.getenv

setup.py
```diff
@@ -54,21 +55,37 @@
...
+    nightly = True
+    DISTNAME += "-nightly"
```
canyon289: if dist name needs to be mutated with environmental variables then there should be a a similar function to `get_version`

canyon289:
@theorashid definitely a step in the right direction! Thank you for making the updates, I left a few more comments

setup.py
theorashid: Could you explain what you mean by this? `os.getenv` returns the value of the environment variable key if it exists, or default if it doesn’t, rather than True/False, so I don't see how `os.getenv("BUILD_PYMC_NIGHTLY")` can be passed directly into `get_version()`

setup.py
theorashid: I've done this now. Out of interest, what's wrong with `+=`? `version = "{}{}".format(version, suffix)` does not seem more succinct/readable

setup.py
canyon289: My apologies youre right. this can be done just fine with the if block you have above. My suggestion then is to move this entire block of logic to `get_version` and also consolidate the distname in there as well.

The reason for this is right now to understand where a version and dist name is coming from one needs to trace in four (and a half) places
1. Github secrets setting
2. The yml file
3. the `get_version` function
4.  This conditional in the global scope

In OSS the code that builds the package, while extremely important, often is not well understood by most people and is not tested, so its good to make it as simple as possible. That way when someone needs to change it its very clear and precise whats happening. Hope that helps!

I also know this may be a bit more scope than you were intending! I hope you dont mind

theorashid: I've used two separate functions. Before editing the setup.py, we had `version=get_version()`, so this maintains one function for name and version

canyon289 (approved):

canyon289:
Thanks @theorashid these changes look great. Now all we need is someone with pypi permissions and whatnot to fix up that side. Do you know if someone on the team has covered that or would it help if I asked around?

setup.py
```diff
@@ -54,21 +55,38 @@
...
-    raise RuntimeError(f"Unable to find version in {VERSIONFILE}.")
+            version = mo.group(1)
+
+            if nightly_build:
+                suffix = datetime.now(timezone.utc).strftime(r".dev%Y%m%d")
+                version = f"{version}{suffix}"
+
+            return version
+
+    raise RuntimeError(f"Unable to find version in {version_file}.")
```
canyon289, line 82: nit: Move this up top with the other globals on line 23 or so and capitalize it to make it clear it is a global

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip
+          python -m pip install build
+      - name: Build the sdist
+        run: python -m build --sdist .
+        env:
+          BUILD_PYMC_NIGHTLY: true
```
canyon289, line 25: The one lingering thing I'm worried about is that someone could see this one yaml line and think if they change it to false the nightly build will stop, when in fact they wont because what's being checked in the script is the presence of an environmental var, not that its True or False. Any suggestions on how to mitigate this issue?

It'll be good to get nightly issues so I dont think this should block this PR is no solution is evident, just want to note it

setup.py
theorashid: I'm confused. What would you like me to move up.?`version` and `version_file` are now within the function so they aren't global

.github/workflows/nightly.yml
theorashid: I am unsure on this and it is beyond my experience with workflows. I copied this from https://github.com/aesara-devs/aesara/pull/656. Perhaps @dfm, who wrote the aesara PR, might be able to shed some light on this?

theorashid:
> Thanks @theorashid these changes look great. Now all we need is someone with pypi permissions and whatnot to fix up that side. Do you know if someone on the team has covered that or would it help if I asked around?

Cheers. I haven't asked anyone to cover that so if you could ask around, that would be great

setup.py
canyon289: I'm sorry I meant this line with the `nightly` global var

https://github.com/pymc-devs/pymc/pull/5490/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R84

.github/workflows/nightly.yml
canyon289: No problem, this as is is already a great contribution. This can be followed up later! :)

canyon289:
> > Thanks @theorashid these changes look great. Now all we need is someone with pypi permissions and whatnot to fix up that side. Do you know if someone on the team has covered that or would it help if I asked around?
> 
> Cheers. I haven't asked anyone to cover that so if you could ask around, that would be great


@fonnesbeck @twiecki @michaelosthege could one of you three setup pypi and the github repo secrets to be enable nightly publishing per this pr? Listing you three based on current pypi ownership


image

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - name: Give PyPI a chance to update the index
+      run: sleep 240
+    - name: Install from PyPI
+      run: |
+        pip install pymc-nightly==${GITHUB_REF:11}
```
michaelosthege: `${GITHUB_REF:11}` is a substring starting at the 11th position of the [`GITHUB_REF` environment variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables). Quoting the docs:
> The branch or tag ref that triggered the workflow run. For branches this is the format refs/heads/<branch_name>, for tags it is refs/tags/<tag_name>, and for pull requests it is refs/pull/<pr_number>/merge. This variable is only set if a branch or tag is available for the event type. For example, refs/heads/feature-branch-1.

For example, with a tag `v1.2.3` the `GITHUB_REF` is `refs/heads/v1.2.3` and `${GITHUB_REF:11}` evaluates to `1.2.3`.

So you'll have to do something else to get the nightly build version here.

theorashid: An example of the version would be 4.0.0b2.dev20220221. I've made an attempt following [this post](https://stackoverflow.com/questions/60942067/get-current-date-and-time-in-github-workflows), but it's difficult to test .yml. I skipped the step to create a date variable because we don't need to reuse the date elsewhere in the workflow and the extra step makes it less readable – although I haven't tested whether it is _essential_ to create a date variable for `pip install pymc-nightly==${GITHUB_REF:11}.dev$(date +"%Y%m%d")` to work

twiecki:
Seems like I don't have admin:
image

theorashid:
> @fonnesbeck @twiecki @michaelosthege could one of you three setup pypi and the github repo secrets to be enable nightly publishing per this pr? Listing you three based on current pypi ownership

@fonnesbeck @michaelosthege are either of you able to setup the secrets so I can get this merged?

.github/workflows/nightly.yml
michaelosthege: The current version won't work, because the `GITHUB_REF` only includes the version number because it's a manually created Tag.
You'll need an environment variable of the current PyMC version, for example by extracting it from `__init__.py` with grep.

Does this have implications for _when_ we should bump our version number? Currently we do that right before a release.
Does the 4.0.0b2 version number even need to be included here?

theorashid: In the edited setup.py, I generate the pymc-nightly version using `suffix = datetime.now(timezone.utc).strftime(r".dev%Y%m%d")` and `version = f"{version}{suffix}"`.

The code showing on this thread is outdated, and I have updated to `pymc-nightly==${GITHUB_REF:11}.dev$(date +"%Y%m%d")`, which I hope translates to (e.g.) `pymc-nightly==4.0.0b2.dev20220221`.

I would say the version number is needed here. If we just ran `pip install pymc-nightly`, we cannot be sure we are downloading the _latest_ version of pymc-nightly.

To clarify, is the version number from `GITHUB_REF` 4.0.0 or 4.0.0b2?  If it's a manually created tag, all the nightly builds in between will still start with this version number, so as long as `GITHUB_REF` is updated each release, this should be fine

canyon289:
@theorashid pinged a couple of folks again. Apologies on the delay

fonnesbeck:
I can set this up.

fonnesbeck:
There's an existing `PYPI_TOKEN_PYMC` secret that has upload permissions. Let me know if this is not sufficient.

michaelosthege:
> There's an existing `PYPI_TOKEN_PYMC` secret that has upload permissions. Let me know if this is not sufficient.

The current token was created by me recently and is only valid for uploading "pymc".
Last time I checked the changes in this PR it looked like the upload was configured to publish a "pymc-nightly" package, which will need it's own token.
First upload one manually. Whoever does this becomes the owner of that new project, so I suggest Chris should do it. Once that project is uploaded you can create a token specific to it.

fonnesbeck:
OK will run it on TestPyPI first just to make sure it works.

fonnesbeck:
[Here's the test](https://test.pypi.org/project/pymc-nightly/4.0.0b2.dev20220302/). Look okay to everyone?

fonnesbeck:
Uploaded successfully: https://pypi.org/project/pymc-nightly/4.0.0b2.dev20220303/

Created an upload token as well. Should be all set.

twiecki:
Thanks @theorashid!

PR #5491, opened 2022-02-19
update Exponential numpydoc

chritter:
Addresses #5459. 

Ensures `pm.Exponential` class descriptions and docstrings are conform with numpy doc validation.

Note we introduced restructured text syntax for a link (`aeppl.logprob <https://github.com/aesara-devs/aeppl/blob/main/aeppl/logprob.py/>`_.) which I would propose to adopt for all code. What do you think?
 

#DataUmbrellaPyMCSprint

Working with @jmloyola.


pymc/distributions/continuous.py
```diff
@@ -1487,7 +1487,7 @@ def logcdf(value, mu):
 
         Parameters
         ----------
-        value: numeric or np.ndarray or aesara.tensor
+        value : numeric or ndarray or TensorVariable
```
```suggestion
        value : tensor_like of float
```

pymc/distributions/continuous.py
```diff
@@ -1456,12 +1457,11 @@ class Exponential(PositiveContinuous):
...
     Notes
     -----
-    Logp calculation is defined in aeppl.logprob
-    https://github.com/aesara-devs/aeppl/blob/main/aeppl/logprob.py
+    Logp calculation is defined in `aeppl.logprob <https://github.com/aesara-devs/aeppl/blob/main/aeppl/logprob.py/>`_.
 
     Parameters
     ----------
-    lam: float
+    lam : float
```
OriolAbril: we need to decide on how to document input types to be clear that scalars, arrays and aesara tensors (at least) are valid

pymc/distributions/continuous.py
```diff
@@ -1456,12 +1457,11 @@ class Exponential(PositiveContinuous):
 
     Notes
     -----
-    Logp calculation is defined in aeppl.logprob
-    https://github.com/aesara-devs/aeppl/blob/main/aeppl/logprob.py
+    Logp calculation is defined in `aeppl.logprob <https://github.com/aesara-devs/aeppl/blob/main/aeppl/logprob.py/>`_.
```
OriolAbril, line 1471: This is an improvement but ideally that would use sphinx cross references cc @ricardoV94. Are there plans to document the functions in logprob module? 

related also to https://github.com/pymc-devs/pymc/issues/5329

pymc/distributions/continuous.py
OriolAbril: We have decided to use `tensor_like of float` (or of integer when the parameter should be integer). can you update it?

OriolAbril (approved):

PR #5500, opened 2022-02-19
Updated docstrings for pymc.Beta

sagartomar:

For issue #5459

#DataUmbrellaPyMCSprint

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1205,10 +1205,10 @@ class Beta(UnitContinuous):
 
     Parameters
     ----------
-    alpha : tensor_like of float
-        ``alpha`` > 0.
-    beta : tensor_like of float
-        ``beta`` > 0.
+    alpha : tensor_like of float, optional
+        ``alpha`` > 0. If not specified, then calculated using ``mu`` and ``sigma``.
+    beta : tensor_like of float, optional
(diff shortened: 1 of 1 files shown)


pymc/distributions/continuous.py
```diff
@@ -1204,14 +1205,14 @@ class Beta(UnitContinuous):
...
     ----------
-    alpha: float
-        alpha > 0.
-    beta: float
-        beta > 0.
-    mu: float
-        Alternative mean (0 < mu < 1).
-    sigma: float
-        Alternative standard deviation (0 < sigma < sqrt(mu * (1 - mu))).
+    alpha : tensor_like of float
```
OriolAbril: alpha and beta are also optional I think, but I don't know what the defaults are @ricardoV94

sagartomar: If we don't specify alpha, beta then they will be calculated using mu, sigma. Does it make sense to write the equation for computing alpha, beta from mu, sigma in the docstrings or will that be too long?

OriolAbril: the formula is already there in the extended summary (at the bottom)

sagartomar: is it okay then to just change it to: `alpha : tensor_like of float, optional` or should we add something in the description saying that if not specifed then calculated from mu and sigma?

OriolAbril (approved):

PR #5503, opened 2022-02-19
MNT Removes pandas dependence in data module

thomasjpfan:
Towards https://github.com/pymc-devs/pymc/issues/5469

This PR uses ducktyping instead of directly checking for type in `determine_coords`.

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -25,7 +25,6 @@ from typing import Any, Dict, List, Optional, Sequence, Union
 import aesara
 import aesara.tensor as at
 import numpy as np
-import pandas as pd
 
 from aesara.compile.sharedvalue import SharedVariable
 from aesara.graph.basic import Apply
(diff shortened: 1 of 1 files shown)


pymc/data.py
```diff
@@ -501,7 +500,7 @@ def determine_coords(model, value, dims: Optional[Sequence[str]] = None) -> Dict
         for size, dim in zip(value.shape, dims):
             coord = model.coords.get(dim, None)
             if coord is None:
-                coords[dim] = pd.RangeIndex(size, name=dim)
+                coords[dim] = range(size)
```
ricardoV94, line 503: not sure, but maybe it needs to be converted to a list?

thomasjpfan: I think it's okay either way. In the end, the coords get turned into a tuple:

https://github.com/pymc-devs/pymc/blob/fb7e041fd08ad9c23e06c5ded2cac70a2ee0f9a5/pymc/model.py#L1105 

If it's the same, I slightly prefer `range` since we do not need to create the whole list here. (Also `pd.RangeIndex` stores a `range` internally)

ricardoV94: sounds good

michaelosthege (approved):

PR #5504, opened 2022-02-19
Find instances of rv.logp(x) in PyMC and correct to pm.logp(rv,x)

LucyJimenez:
xref https://github.com/pymc-devs/pymc/issues/5051
On this PR `rv.logp(x)` was replaced by `pm.logp(rv,x)`.

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -289,7 +289,7 @@ class Distribution(metaclass=DistributionMeta):
             functools.partial(str_for_dist, formatting="latex"), rv_out
         )
 
-        rv_out.logp = _make_nice_attr_error("rv.logp(x)", "pm.logp(rv, x)")
+        rv_out.logp = _make_nice_attr_error("pm.logp(rv,x)", "pm.logp(rv, x)")
         rv_out.logcdf = _make_nice_attr_error("rv.logcdf(x)", "pm.logcdf(rv, x)")
         rv_out.random = _make_nice_attr_error("rv.random()", "rv.eval()")
         return rv_out
@@ -382,7 +382,7 @@ class Distribution(metaclass=DistributionMeta):
             rv_out.update = (rng, new_rng)
             rng.default_update = new_rng
 
-        rv_out.logp = _make_nice_attr_error("rv.logp(x)", "pm.logp(rv, x)")
+        rv_out.logp = _make_nice_attr_error("pm.logp(rv,x)", "pm.logp(rv, x)")
         rv_out.logcdf = _make_nice_attr_error("rv.logcdf(x)", "pm.logcdf(rv, x)")
         rv_out.random = _make_nice_attr_error("rv.random()", "rv.eval()")
         return rv_out


OriolAbril:
Thanks for the PR! I have asked someone more familiar with the issue to review. I think that those instances should not be updated as it looks like they are used to write error messages or deprecation warnings for people trying to use the old way so they get something informative saying something along the lines of "rv.logp is deprecated, use pm.logp(rv) instead"

ricardoV94 (changes requested):
Yes these lines shouldn't be changed. The original issue referred to possible uses in the documentation. Apologies if that was not clear.

LucyJimenez:
Thanks @OriolAbril and @ricardoV94 for the explanations. I have been checking all the files and found no matches with `rv.logp(x)`, is it possible that I am missing something?

ricardoV94:
> Thanks @OriolAbril and @ricardoV94 for the explanations. I have been checking all the files and found no matches with `rv.logp(x)`, is it possible that I am missing something?

Have you looked for the general pattern? The `rv.logp(x)` is just a template, not the actual code that may or not exist.

Things might also be spread across different lines:
```python
with pm.Model() as m:
  # may be called something
  # completely different like x or mu
  rv = pm.Normal("rv")

# value can be anything
# like a numpy array or a number
rv.logp(value)  # old API
pm.logp(rv, value)  # new API
```

Basically if you see something like `rv.logp(value)`, it's the old API and needs to be updated to something like `pm.logp(rv, value)`,

It's also possible that there are no uses left in the documentation, and we can just close the issue.

Searching for `.logp` should be enough to find all occurrences of both old and new uses.

ricardoV94:
For example this guide is still outdated: https://github.com/pymc-devs/pymc/blob/99ec0ff93362a6d6eba90cbad49a2338c353eccd/docs/source/guides/Probability_Distributions.rst

But it's not a simple fix either...

LucyJimenez:
@ricardoV94 thank you for all the resources. I'm going to check it more carefully and getting back with questions ;)

OriolAbril:
@LucyJimenez I checked for instances of logp, logcdf and random _method_ usage and found ocurrences in only 3 files: the developer guide, `PyMC_and_Aesara` one and `Probability_Distributions` that was mentioned by @ricardoV94 above. All 3 explain PyMC distributions, how they are structured into classes, it's methods and the class internals. All three are outdated and refer to version 3 and only the last two are excluded from the docs right now.

It doesn't make much sense to update the code only in these files, both code an explanations need to be updated and that can only be done by someone who understands the new v4 internals (not yet documented anywhere). We should have closed the issue when we updated the notebooks inside the "Learning" section in https://github.com/pymc-devs/pymc/pull/5269 instead of labeling it as recommended issue for the sprint, sorry about that. I'll go close it now. 

You'll have to find another issue to work on. Don't hesitate to tag me on Data Umbrella discord (`@Oriol Abril Pla` there) if you want to have a quick chat/call to find a new issue and get started on it.

LucyJimenez:
@OriolAbril Thank you for clarifying the general situation. Considering your comment, I will leave this PR free and look for you on discord.

OriolAbril:
Perfect, I'll try to check Discord a couple times a day to make sure I don't miss your messages.

PR #5505, opened 2022-02-20
Update docstring for model.point_logps

### LINKED ISSUE #4634: Update docstring for model.point_logps
The argument `point_logps` method is underdocumented. It doesnt indicate that both arguments are optional and what the defaults are

https://github.com/pymc-devs/pymc3/blob/v4/pymc3/model.py#L1364

purna135:
Addresses #4634

Documented both arguments in point_logps are optional and what the defaults are.

#DataUmbrellaPyMCSprint

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1691,10 +1691,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         Parameters
         ----------
-        point: Point, default 'None'
+        point: Point, optional
             Point to be evaluated.  If ``None``, then ``model.initial_point``
             is used.
-        round_vals: int, default '2'
+        round_vals: int, default 2
             Number of decimals to round log-probabilities.
 
         Returns
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1691,11 +1691,11 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         Parameters
         ----------
-        point: Point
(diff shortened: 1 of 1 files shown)


pymc/model.py
```diff
@@ -1691,11 +1691,11 @@ def point_logps(self, point=None, round_vals=2):
 
         Parameters
         ----------
-        point: Point
+        point: Point, default 'None'
```
```suggestion
        point: Point, optional
```

The default isn't really None as explained in the description, None is used as a placeholder to indicate that the attribute `model.initial_point` should be used. In such cases optional instead of default should be used. Reference https://pymc-data-umbrella.xyz/en/latest/sprint/docstring_tutorial.html#parameters

pymc/model.py
```diff
@@ -1691,11 +1691,11 @@ def point_logps(self, point=None, round_vals=2):
...
             Point to be evaluated.  If ``None``, then ``model.initial_point``
             is used.
-        round_vals: int
-            Number of decimals to round log-probabilities
+        round_vals: int, default '2'
```
```suggestion
        round_vals: int, default 2
```

The default is the _integer_ 2, not the _string_ '2'. The ticks can be confusing, especially since the type is indicated as int

purna135:
Thanks @OriolAbril, updated the docstring as per your suggestions.

OriolAbril (approved):

PR #5511, opened 2022-02-21
Remove or in test suite

canyon289:
Makes things less confusing

### DIFF
--- a/pymc/tests/test_distributions_random.py
+++ b/pymc/tests/test_distributions_random.py
@@ -428,9 +428,9 @@ class TestFlat(BaseTestDistributionRandom):
     ]
 
     def check_rv_inferred_size(self):
-        self.sizes_to_check = [None, (), 1, (1,), 5, (4, 5), (2, 4, 2)]
-        self.sizes_expected = [(), (), (1,), (1,), (5,), (4, 5), (2, 4, 2)]
-        for size, expected in zip(self.sizes_to_check, self.sizes_expected):
+        sizes_to_check = [None, (), 1, (1,), 5, (4, 5), (2, 4, 2)]
+        sizes_expected = [(), (), (1,), (1,), (5,), (4, 5), (2, 4, 2)]
+        for size, expected in zip(sizes_to_check, sizes_expected):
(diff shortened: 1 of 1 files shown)


pymc/tests/test_distributions_random.py
```diff
@@ -428,9 +428,9 @@ class TestFlat(BaseTestDistributionRandom):
     ]
 
     def check_rv_inferred_size(self):
-        sizes_to_check = self.sizes_to_check or [None, (), 1, (1,), 5, (4, 5), (2, 4, 2)]
```
twiecki, line 431: But this will override `self.sizes_to_check` which the previous code didn't do, is that safe?

canyon289: That's of point of the or statement isnt it? So if it gets overridden already through a conditional why not just override it explicitly in this function?

The or statement makes the test much more confusing to read, the tests pass without it and so far no one has mentioned a reason why its needed

canyon289: Ok I see what youre saying. If its the assignment on self that`s concerning I'll change the PR to purely work off of local variables. All I care about is removing the `or` which doesn't add anything, just makes it harder to trace this test and understand what its doing.

In that case we can change it to the following
```
        sizes_to_check = [None, (), 1, (1,), 5, (4, 5), (2, 4, 2)]
        sizes_expected = [(), (), (1,), (1,), (5,), (4, 5), (2, 4, 2)]
        for size, expected in zip(sizes_to_check, sizes_expected):
            pymc_rv = self.pymc_dist.dist(**self.pymc_dist_params, size=size)
            expected_symbolic = tuple(pymc_rv.shape.eval())
            assert expected_symbolic == expected
```

ricardoV94: This is changing the original behavior, and ignoring when a test uses custom sizes_to_check

ricardoV94: yeah that looks good

canyon289:
done

ricardoV94 (approved):

twiecki (approved):

PR #5516, opened 2022-02-23
update reported warnings after sampling

aloctavodia:
I am proposing to simplify the messages we provide to users and update the recommendation to match the state-of-the-art methods and practices.

* Only report a problem with rhat if it is higher than 1.01.
* Only report a problem with ESS if is less than 100 per chain
* Report when user are using 1 or less than 4 chains

I think we currently provide too many levels of warnings, for example with the ESS, the only thing a user should care about is if they have enough effective samples to compute something. So messages like "The number of effective samples is smaller than x% " should only be valuable to someone comparing the efficiency of sampling, but not someone worried about computing stuff with posteriors. Regarding this point with Oriol and others we are working on bringing new tools to ArviZ that will make easier for the user to know if they have enough samples to compute something to a given accuracy and if not how many more samples should then get.

### DIFF
--- a/pymc/backends/report.py
+++ b/pymc/backends/report.py
@@ -126,7 +126,7 @@ class SamplerReport:
 
         elif idata.posterior.sizes["chain"] < 4:
             msg = (
-                "We recommend running at least 4 chains for proper computation of "
+                "We recommend running at least 4 chains for robust computation of "
                 "convergence diagnostics"
             )
             warn = SamplerWarning(WarningType.BAD_PARAMS, msg, "info")
@@ -151,7 +151,8 @@ class SamplerReport:
         if rhat_max > 1.01:
             msg = (
                 "The rhat statistic is larger than 1.01 for some "
-                "parameters. This might indicate problems during sampling."
+                "parameters. This indicates problems during sampling. "
+                "See https://arxiv.org/abs/1903.08008 for details"
             )
             warn = SamplerWarning(WarningType.CONVERGENCE, msg, "info", extra=rhat)
             warnings.append(warn)
@@ -161,7 +162,8 @@ class SamplerReport:
         if eff_per_chain < 100:
             msg = (
                 "The effective sample size per chain is smaller than 100 for some parameters. "
-                " A higher number is needed for reliable rhat and ess computation"
+                " A higher number is needed for reliable rhat and ess computation. "
+                "See https://arxiv.org/abs/1903.08008 for details"
             )
             warn = SamplerWarning(WarningType.CONVERGENCE, msg, "error", extra=ess)
             warnings.append(warn)
--- a/pymc/backends/report.py
+++ b/pymc/backends/report.py
@@ -124,6 +124,15 @@ class SamplerReport:
             self._add_warnings([warn])
             return
 
+        elif idata.posterior.sizes["chain"] < 4:
+            msg = (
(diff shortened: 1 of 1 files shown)


pymc/backends/report.py
```diff
@@ -139,44 +148,23 @@ def _run_convergence_checks(self, idata: arviz.InferenceData, model):
...
-            msg = "The rhat statistic is larger than 1.2 for some " "parameters."
-            warn = SamplerWarning(WarningType.CONVERGENCE, msg, "warn", extra=rhat)
-            warnings.append(warn)
-        elif rhat_max > 1.05:
-            msg = (
-                "The rhat statistic is larger than 1.05 for some "
-                "parameters. This indicates slight problems during "
-                "sampling."
+                "The rhat statistic is larger than 1.01 for some "
+                "parameters. This might indicate problems during sampling."
```
OriolAbril: are there cases with large rhat that have no problems during sampling? I think the might can be confusing and overly optimistic, but I am biased towards strong wording given how I have seen disastrous diagnostics be ignored very often

pymc/backends/report.py
```diff
@@ -124,6 +124,15 @@ def _run_convergence_checks(self, idata: arviz.InferenceData, model):
             self._add_warnings([warn])
             return
 
+        elif idata.posterior.sizes["chain"] < 4:
+            msg = (
+                "We recommend running at least 4 chains for proper computation of "
```
ColCarroll:
```suggestion
                "We recommend running at least 4 chains for reliable computation of "
```

ColCarroll: or "robust"?

pymc/backends/report.py
ColCarroll:
```suggestion
                "parameters. This might indicate problems during sampling. See https://arxiv.org/abs/1903.08008"
```

ColCarroll: My hope with these is to also give users a followup action to take, so a citation might help -- what do you think? 

With respect to "are there cases with large rhat that have no problems during sampling?", i did two quick experiments using a poorly behaved distribution (Cauchy), and independent samples. My rough impression is that in this case, it would be quite rare to get an rhat over 1.01, but this is a super sketchy experiment in a github comment, so who knows:
1. Simulate drawing 4 chains of 400 draws from a Cauchy(1), and check the max r-hat from doing that 1,000 times:
image
2. What if we had shorter chains? Here is a plot of the rhat after `n` draws, where `n` goes from 30 to 400 total draws:
image

this one changes a lot depending on which draws you get, so here are 6 such experiments on the same plot: image

aloctavodia: I swear that I was internally struggling with using or not that "might" :-) It was an attempt to soften that hard threshold. But agree that strong language is a better idea, users are still free to ignore warnings, but at least now they will feel a little bit more guilty, haha. 

I agree @ColCarroll that we should provide actions. We have been talking with Oriol, Aki and others about having a centralized website that we can point users to. This site could have "levels of information" so the first level could be a direct action, the second level could provide more nuanced recommendations and/or a better explanations of why the diagnostics work and a third level could be a paper.

Regarding your experiments for short chains at least too short the ESS will be low, so the r-hat is not trustworthy.

aloctavodia:
@OriolAbril @ColCarroll what do you think about this?

OriolAbril (approved):

PR #5517, opened 2022-02-24
Fix minor typo

saiwing-yeung:
Looks like the docstring for `sample_prior_predictive()` has a minor typo.

"posterior predictive" -> "prior predictive"


ColCarroll (approved):
thank you for the catch!

PR #5518, opened 2022-02-24
Raises error when dims match the variable name

### LINKED ISSUE #5309: `pm.Data` silently missing from `InferenceData` when dims match the variable name
Xarray, and thereby `InferenceData` don't support labeling dimensions with the same name as the variable.

```python
with pm.Model() as pmodel:
    pm.Data("city", numpy.arange(5), dims="cities") # ✔ ends up in .constant_data as expected

    pm.Data("time", numpy.arange(5), dims="time")   # ❌ silently missing
    pm.Normal("a")                                  # ❌ errors in InferenceData conversion
    idata = pm.sample(chains=1)
```

Only variable "a" caused an error - but unfortunately only after MCMC.
```
MissingDimensionsError: 'a' has more than 1-dimension and the same name as one of its dimensions ('chain', 'draw', 'a').
xarray disallows such variables because they conflict with the coordinates used to label dimensions.
```

Since dims are passed at creation of the variables, we should raise that error in `Model.add_random_variable`. That will raise the error much earlier, and work for variables and data.

https://github.com/pymc-devs/pymc/blob/14c4f39122ae313e25b2698195067ec44b69b26a/pymc/model.py#L1416-L1429

* PyMC/PyMC3 Version: latest

kimbochen:
Fixes #5309 , where a data variable misses when its name collides with its dimensions.

- Added an assertion that checks if all dimension labels are different from the variable name.
- Added a test that checks if an error is raised when a variable name collides with its dimension labels.

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1441,6 +1441,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
             if isinstance(dims, str):
                 dims = (dims,)
             assert all(dim in self.coords or dim is None for dim in dims)
+            assert all(
+                var.name != dim for dim in dims
+            ), f"Variable `{var.name}` has the same name as its dimension label."
             self._RV_dims[var.name] = dims
 
         self.named_vars[var.name] = var
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1444,7 +1444,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
                 raise ValueError(f"Dimension {dim} is not specified in `coords`.")
             if any(var.name == dim for dim in dims):
                 raise ValueError(f"Variable `{var.name}` has the same name as its dimension label.")
-
             self._RV_dims[var.name] = dims
 
         self.named_vars[var.name] = var
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1440,16 +1440,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
         if dims is not None:
             if isinstance(dims, str):
                 dims = (dims,)
-
-            for dim in dims:
-                if dim in self.coords:
-                    raise ValueError(f"Dimension {dim} is already specified in `coords`.")
-                if dim is None:
-                    raise ValueError("Empty dimension name.")
-                if var.name != dim:
-                    raise ValueError(
-                        f"Variable `{var.name}` has the same name as its dimension label."
-                    )
+            if any(dim not in self.coords and dim is not None for dim in dims):
+                raise ValueError(f"Dimension {dim} is not specified in `coords`.")
+            if any(var.name == dim for dim in dims):
+                raise ValueError(f"Variable `{var.name}` has the same name as its dimension label.")
 
             self._RV_dims[var.name] = dims
 
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1440,10 +1440,17 @@ class Model(WithMemoization, metaclass=ContextMeta):
         if dims is not None:
             if isinstance(dims, str):
                 dims = (dims,)
-            assert all(dim in self.coords or dim is None for dim in dims)
-            assert all(
-                var.name != dim for dim in dims
-            ), f"Variable `{var.name}` has the same name as its dimension label."
+
+            for dim in dims:
+                if dim in self.coords:
+                    raise ValueError(f"Dimension {dim} is already specified in `coords`.")
+                if dim is None:
+                    raise ValueError("Empty dimension name.")
+                if var.name != dim:
+                    raise ValueError(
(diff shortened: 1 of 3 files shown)


pymc/tests/test_idata_conversion.py
```diff
@@ -602,6 +602,15 @@ def test_issue_5043_autoconvert_coord_values(self):
             )
             assert isinstance(converter.coords["city"], pd.MultiIndex)
 
+    @pytest.mark.xfail(
```
OriolAbril: The test should use a `pytest.raises` context and not be marked as xfail. See https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions

pymc/model.py
```diff
@@ -1441,6 +1441,9 @@ def add_random_variable(self, var, dims: Optional[Tuple[Union[str, None], ...]]
             if isinstance(dims, str):
                 dims = (dims,)
             assert all(dim in self.coords or dim is None for dim in dims)
+            assert all(
+                var.name != dim for dim in dims
+            ), f"Variable `{var.name}` has the same name as its dimension label."
```
michaelosthege: Since asserts can be turned off globally, and because of coverage calculation, the lines above should be an if statement raising an exception such as ValueError.
I'm sorry to nitpick about this, since clearly you were just following the style of the line above.. Could you change both into an if/raise?

pymc/tests/test_idata_conversion.py
```diff
@@ -602,6 +602,15 @@ def test_issue_5043_autoconvert_coord_values(self):
...
+        reason="Xarray doesn't support labeling dimensions with the same name as the variable"
+    )
+    def test_variable_dimension_name_collision(self):
+        with pm.Model() as pmodel:
+            pm.ConstantData("time", np.ones(9), dims="time")
+            pm.Normal("a")
+            idata = pm.sample(chains=1)
```
michaelosthege: The above two lines are unreachable. You could also call model.register_rv directly

kimbochen: Could you be more specific about which 2 lines are unreachable?
I also looked up the documentation of model.register_rv, but I couldn't figure out how to use it.
Would you mind giving me some additional pointers?

kimbochen:
Thank you @OriolAbril and @michaelosthege for the detailed review!

- I replaced xfail with pytest.raise, which asserts that a ValueError to be raised.
- In `model.py`, I replaced the two assert statements with if / raise statements.

As for the tests, I had trouble using model.register_rv, so I haven't changed the test yet.

michaelosthege:
@kimbochen The change in `model.py` looks good!
Our coverage tool things that it's not yet covered, but it actually is. The coverage just didn't receive the upload because some of the tests failed.

image

So your change actually found bugs in these four tests! 🥳
```
ERROR pymc/tests/test_model_graph.py::TestModelWithDims::test_inputs - ValueE...
ERROR pymc/tests/test_model_graph.py::TestModelWithDims::test_compute_graph
ERROR pymc/tests/test_model_graph.py::TestModelWithDims::test_plates - ValueE...
ERROR pymc/tests/test_model_graph.py::TestModelWithDims::test_graphviz - Valu...
```
Can you fid them by renaming the variables in the test case?

> Could you be more specific about which 2 lines are unreachable?
> I also looked up the documentation of model.register_rv, but I couldn't figure out how to use it.
> Would you mind giving me some additional pointers?
```python
with pm.Model() as pmodel:
    pm.ConstantData("time", np.ones(9), dims="time") # 👈 should raise an error
    pm.Normal("a") # 👈 making this line unreachable because of the error above
```
I think you can use `register_rv` like this:
```python
with pm.Model() as pmodel:
    var = at.as_tensor([1, 2, 3])
    pmodel.register_rv(var, name="mytestvar", dims=("time",))
```

Oh and please update the test to assert the message: `with pytest.raises(ValueError, match="same name as its dimension")`.
To make sure it's not passing because of _any_ ValueError.

kimbochen:
@michaelosthege Thank you for the guidance!
I'm glad that my contribution managed to find some bugs.

- I updated the test according to your suggestion: https://github.com/pymc-devs/pymc/blob/9316eb88033a8f989f6db73d69bfe1d366ed03ce/pymc/tests/test_idata_conversion.py#L607-L610
- As for the bugs in the tests, the 4 errors seem to be triggered by the name collision of the `time` variable: https://github.com/pymc-devs/pymc/blob/a3bab7ddc0f4612ad039a7d2755f22fa6c8c0a72/pymc/tests/test_model_graph.py#L107
  - The name of `time` variable is changed to `time`.
  - The following mentions of the variable name are all changed accordingly.

michaelosthege (approved):
Yes, that's looking good. Thanks @kimbochen !

PR #5519, opened 2022-02-24
Group potential and likelihood terms in the same cost in `logp_dlogp_function`

ricardoV94:
Here is a PR that nobody asked for. 

The reason for this change? This makes it consistent with how `SMC` and `VI` handle tempered posteriors.

Also, it t seems tempered `logp_dlogp_function` is not being used anywhere in the library.


ricardoV94:
@aloctavodia Shall we go ahead with this one for the time being? At least, it makes it internally consistency with SMC and VI

aloctavodia (approved):

PR #5520, opened 2022-02-24
Enable users to pass extra options to numpyro.infer.NUTS in sampling_jax

lucianopaz:
This is a relatively small PR that only enables users to pass extra options to the NUTS step method when performing `sampling_jax`.

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -151,6 +151,7 @@ def sample_numpyro_nuts(
     keep_untransformed=False,
     chain_method="parallel",
     idata_kwargs=None,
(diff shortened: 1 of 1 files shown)


lucianopaz:
I don't understand why pre commit is complaining about the `docs/source/contributing/pr_tutorial.md` file, which I haven't touched at all.

ricardoV94 (approved):

twiecki (approved):

twiecki:
Fixed the pre-commit issue on main.

ricardoV94:
Actually this should interact with the other nuts hard coded kwargs no? @lucianopaz

PR #5526, opened 2022-02-25
Jitter initial points in sample_numpyro_nuts

ricardoV94:


twiecki:
Should we add this to blackjax too?

ricardoV94:
> Should we add this to blackjax too?

That one is not merged yet

twiecki (approved):

pymc/sampling_jax.py
```diff
@@ -6,6 +6,8 @@
 
 from typing import Callable, List, Optional
 
+from pymc.sampling import _init_jitter
```
lucianopaz, line 9: I think that this is ok for now, but in the long run I think that `sampling` should be a directory. It could house the code for the initial point selection (e.g. `_init_jitter`), and sampling specific to our python samplers, future aehmc, numpyro, blackjax or whatever comes along.

lucianopaz (commented):
Looks good to me @ricardoV94. I'll wait until the tests pass before I approve the PR.

This problem got me thinking that in the long run, we should consider making `sampling` a directory. We could house the inference samplers in different files for each backend that we use, we could also split the utility init_jitter code into a module of its own, and we could also take our forward samplers out of `sampling.py` itself.

ricardoV94:
> Looks good to me @ricardoV94. I'll wait until the tests pass before I approve the PR.
> 
> This problem got me thinking that in the long run, we should consider making `sampling` a directory. We could house the inference samplers in different files for each backend that we use, we could also split the utility init_jitter code into a module of its own, and we could also take our forward samplers out of `sampling.py` itself.

Yeah. I think what you are saying makes perfect sense in the context of this GSOC project: https://github.com/pymc-devs/pymc/wiki/GSoC-2022-projects#integrate-jax-samplers-into-standard-api

PR #5529, opened 2022-02-28
SMC: Use correlation to autotune n_steps

aloctavodia:
Instead of pre computing the number of steps from the previous acceptance rate this PR proposes to use the autocorrelation and a threshold value to decide when to stop sampling.

### DIFF
--- a/pymc/smc/smc.py
+++ b/pymc/smc/smc.py
@@ -337,41 +337,21 @@ class SMC_KERNEL(ABC):
 class IMH(SMC_KERNEL):
     """Independent Metropolis-Hastings SMC kernel"""
 
-    def __init__(self, *args, n_steps=25, tune_steps=True, p_acc_rate=0.85, **kwargs):
+    def __init__(self, *args, correlation_threshold=0.01, **kwargs):
         """
         Parameters
         ----------
-        n_steps: int
-            The number of steps of each Markov Chain. If ``tune_steps == True`` ``n_steps`` will be used
-            for the first stage and for the others it will be determined automatically based on the
-            acceptance rate and `p_acc_rate`, the max number of steps is ``n_steps``.
-        tune_steps: bool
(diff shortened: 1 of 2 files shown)


pymc/smc/smc.py
```diff
@@ -384,34 +364,43 @@ def tune(self):
...
                 (proposal_logp + backward_logp) - (self.tempered_posterior_logp + forward_logp)
             )
 
-            ac_[n_step] = accepted
-            self.tempered_posterior[accepted] = proposal[accepted]
-            self.tempered_posterior_logp[accepted] = proposal_logp[accepted]
-            self.prior_logp[accepted] = pl[accepted]
-            self.likelihood_logp[accepted] = ll[accepted]
-            backward_logp[accepted] = forward_logp[accepted]
+            tmp = np.copy(self.tempered_posterior)
```
ricardoV94: why do we need to copy the tempered posterior?

aloctavodia: in case we reject the autocorrelation condition on line 394

ricardoV94: Why not just do
```python
pearson_r = corr.get(proposal[accepted])
```

Or do you think the repeated indexing is slower than the copy operation?

aloctavodia: In that case you will be evaluating only the new (accepted) positions, but you need to evaluate both the new and old, i.e. those that you did not change because you rejected the proposal.

ricardoV94: You are right. Alternatively why not break after updating the values? Does it matter if we stop after correlation becomes smaller than threshold, instead of just before (as now)? We already did the work of resampling anyway

aloctavodia: practicality beats purity. I like it!

ricardoV94 (approved):

PR #5532, opened 2022-03-01
Bart variable importance

aloctavodia:
This adds a new helper function to help interpret variable importance from BART.

Short version:

from the top plot the most important variable are "hour", and then "temperature", "humidity" and finally "workingday".
from the bottom plot a model with just the two most important components ("hour" and "temperature") are enough to generate predictions that are on average indistinguishable from a model with more variables.

index

Long version:
The top plot is commonly used in BART literature and is representing the relative importance of covariates. This relative importance is simply computed as how many times covariates are used in the underlying trees. Our implementation automatically adjust the prior probability of covariates to be included, reducing the chance of spurious variables having high importance even when the number of trees, m, is increased.

The bottom plot is done by computing predictions of a reference model, with all, p, covariates against models with 1 to p covariates. That is, a model with only the most important covariate, the model with the two most important covariates, etc. The , predictions are compared using the Pearson correlation coefficient.  Notice that this comparison is actually an approximation. The model is fitted only once with all covariates and to estimate the prediction with less covariates we simply transverse the trees ignoring covariates (or if you wish by pruning branches). Given that BART trees are very shallow, this seems like a good approximation.


cluhmann:
>The predictions are compared using the Pearson correlation coefficient.

Is there any reason that the comparison is relative (Pearson's _r_) rather than absolute (e.g., delta-logp)?

aloctavodia:
The comparison in terms of Pearson's coefficient is simple to both implement and interpret. And in my test it seems to behave adequately. Nevertheless I don't rule out other metrics and in fact I am preparing experiments on that matter.

pymc/bart/utils.py
```diff
@@ -296,3 +299,68 @@ def plot_dependence(
 
     fig.text(-0.05, 0.5, Y_label, va="center", rotation="vertical", fontsize=15)
     return axes
+
+
+def variable_importance(idata, labels=None, figsize=None, samples=100, random_seed=None):
```
fonnesbeck: Since its returning a plot, perhaps it should be called `plot_variable_importance` or similar?

pymc/bart/utils.py
```diff
@@ -21,6 +22,8 @@ def predict(idata, rng, X_new=None, size=None):
         A new covariate matrix. Use it to obtain out-of-sample predictions
     size: int or tuple
         Number of samples.
+    excluded: list
+        id of the variables to exclude when computing predictions
```
fonnesbeck: By "id", do we mean variable names?

aloctavodia: no, indexes. fixed

aloctavodia:
closing in favor of pymc-devs/pymc-experimental#14

PR #5535, opened 2022-03-03
Fix seeding bug in `sample_numpyro_nuts` when more than one chain was sampled

### LINKED ISSUE #5528: numpyro sampling fails with more than one chain
Hi all,

it looks like the latest version of pymc may have broken numpyro sampling when more than one chain is used? Here's a minimal example:

```python
import pymc as pm
from pymc.sampling_jax import sample_numpyro_nuts

with pm.Model() as m:

    test = pm.Normal('test', mu=0, sigma=1)

with m:
    result_jax = sample_numpyro_nuts()
```

Produces:

```python
Compiling...
Compilation time =  0:00:00.754562
Sampling...

Input In [5], in <module>
      1 with m:
----> 2     result_jax = sample_numpyro_nuts()

File ~/projects/pymc_v4/pymc/pymc/sampling_jax.py:258, in sample_numpyro_nuts(draws, tune, chains, target_accept, random_seed, initvals, model, var_names, progress_bar, keep_untransformed, chain_method, idata_kwargs, nuts_kwargs)
    254 print("Compilation time = ", tic2 - tic1, file=sys.stdout)
    256 print("Sampling...", file=sys.stdout)
--> 258 map_seed = jax.random.PRNGKey(random_seed)
    259 if chains > 1:
    260     map_seed = jax.random.split(map_seed, chains)

File ~/miniconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/jax/_src/random.py:119, in PRNGKey(seed)
    105 """Create a pseudo-random number generator (PRNG) key given an integer seed.
    106 
    107 The resulting key carries the default PRNG implementation, as
   (...)
    116 
    117 """
    118 impl = default_prng_impl()
--> 119 key = prng.seed_with_impl(impl, seed)
    120 return _return_prng_keys(True, key)

File ~/miniconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/jax/_src/prng.py:222, in seed_with_impl(impl, seed)
    221 def seed_with_impl(impl: PRNGImpl, seed: int) -> PRNGKeyArray:
--> 222   return PRNGKeyArray(impl, impl.seed(seed))

File ~/miniconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/jax/_src/prng.py:256, in threefry_seed(seed)
    254   seed_arr = jnp.asarray(seed)
    255 if seed_arr.shape:
--> 256   raise TypeError(f"PRNG key seed must be a scalar; got {seed!r}.")
    257 if not np.issubdtype(seed_arr.dtype, np.integer):
    258   raise TypeError(f"PRNG key seed must be an integer; got {seed!r}")

TypeError: PRNG key seed must be a scalar; got array([ 970387577,  294435933, 1058679839,  301875541]).
```

This is with commit hash `a3bab7ddc0f4612ad039a7d2755f22fa6c8c0a72`.

ricardoV94:
Also adds several type-hints

Closes #5528

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -4,8 +4,9 @@ import re
 import sys
 import warnings
 
-from typing import Callable, List, Optional
+from typing import Callable, Dict, List, Optional, Sequence, Union
 
+from pymc.initial_point import StartDict
 from pymc.sampling import _init_jitter
 
 xla_flags = os.getenv("XLA_FLAGS", "")
@@ -130,7 +131,7 @@ def _sample_stats_to_xarray(posterior):
     return data
 
 
-def _get_log_likelihood(model, samples):
+def _get_log_likelihood(model: Model, samples) -> Dict:
     """Compute log-likelihood for all observations"""
     data = {}
     for v in model.observed_RVs:
@@ -142,8 +143,13 @@ def _get_log_likelihood(model, samples):
 
 
 def _get_batched_jittered_initial_points(
-    model, chains, initvals, random_seed, jitter=True, jitter_max_retries=10
-):
+    model: Model,
+    chains: int,
+    initvals: Optional[Union[StartDict, Sequence[Optional[StartDict]]]],
+    random_seed: int,
+    jitter: bool = True,
+    jitter_max_retries: int = 10,
+) -> Union[np.ndarray, List[np.ndarray]]:
     """Get jittered initial point in format expected by NumPyro MCMC kernel
 
     Returns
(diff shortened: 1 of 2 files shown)


twiecki:
I don't see where the actual fix is?

pymc/sampling_jax.py
```diff
@@ -213,9 +216,7 @@ def sample_numpyro_nuts(
         dims = {}
 
     if random_seed is None:
-        random_seed = model.rng_seeder.randint(
-            2**30, dtype=np.int64, size=chains if chains > 1 else None
-        )
+        random_seed = model.rng_seeder.randint(2**30, dtype=np.int64)
```
ricardoV94, line 219: @twiecki The fix is here. Only one seed is created instead of one per `chain`

twiecki: I see, but then all chains have the same seed, right?

ricardoV94: No, they are split below and also in the jitter method

ricardoV94: The new tests also check that the samples differ across chains

twiecki: :+1:

twiecki (approved):

PR #5540, opened 2022-03-04
Replacing `assert_op` by `raise_op`

larryshamalama:
In light of a warning that I encountered earlier, `aesara.assert_op` seems to be depreciated and replaced by `raise_op`.

From [here](https://github.com/aesara-devs/aesara/blob/main/aesara/assert_op.py)

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -25,9 +25,9 @@ import aesara.tensor as at
 import numpy as np
 
 from aeppl.logprob import _logprob, logcdf
-from aesara.assert_op import Assert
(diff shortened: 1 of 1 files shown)


ricardoV94 (approved):
Is this the only place where we are using assert?

larryshamalama:
`Assert` is used in two other files, but imported from `aesara.raise_op`

See [here](https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/multivariate.py#L29) and [here](https://github.com/pymc-devs/pymc/blob/main/pymc/sampling_jax.py#L27)

PR #5541, opened 2022-03-04
Infer GRW steps from shape

canyon289:
@ricardoV94 Thanks for adding this commit to the grw branch.

I've moved it to another PR solely for reasons of my own learning. and not because I disagree with the changes or think they shouldn't be merged.. Once the GRW pr is good enough as is I'd like to go through and "relearn" everything about it so I can make sure I understood that scope. I'll then do the same for this logic too! 

Hopefully that makes sense


canyon289:
Ill come back this this one now that GRW is merged
https://github.com/pymc-devs/pymc/pull/5298

PR #5543, opened 2022-03-04
Doc change the default value of pymc.draw from 500 to 1.

### LINKED ISSUE #5542: Doc change the default value of pymc.draw from 500 to 1.
image

In [pymc.draw](https://docs.pymc.io/en/latest/api/generated/pymc.draw.html#pymc.draw), the function takes the dafault value of draws=1 but in docstring it's mentioned the default value as 500.

purna135:
Closes #5542

Changed the default value of draws from 500 to 1.

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -2128,7 +2128,7 @@ def draw(
     vars
         A variable or a list of variables for which to draw samples.
     draws : int
-        Number of samples needed to draw. Detaults to 500.
+        Number of samples needed to draw. Detaults to 1.
     mode
         The mode used by ``aesara.function`` to compile the graph.
     **kwargs
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -2125,18 +2125,18 @@ def draw(
 
     Parameters
     ----------
-    vars
+    vars : tensor_like or iterable of tensor_like
         A variable or a list of variables for which to draw samples.
-    draws : int
-        Number of samples needed to draw. Detaults to 1.
-    mode
-        The mode used by ``aesara.function`` to compile the graph.
-    **kwargs
+    draws : int, default 1
+        Number of samples needed to draw.
+    mode : str or `Mode` instance, optional
+        The mode used by :func:``aesara.function`` to compile the graph.
+    **kwargs : dict, optional
         Keyword arguments for :func:`pymc.aesara.compile_pymc`
 
     Returns
     -------
-    List[np.ndarray]
+    out : list of ndarrays
         A list of numpy arrays.
 
     Examples
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -2125,18 +2125,18 @@ def draw(
 
     Parameters
     ----------
-    vars : tensor_like or iterable of tensor_like
+    vars : Variable or iterable of Variable
         A variable or a list of variables for which to draw samples.
     draws : int, default 1
         Number of samples needed to draw.
-    mode : str or `Mode` instance, optional
-        The mode used by :func:``aesara.function`` to compile the graph.
+    mode : str or aesara.compile.mode.Mode, optional
+        The mode used by :func:`aesara.function` to compile the graph.
     **kwargs : dict, optional
-        Keyword arguments for :func:`pymc.aesara.compile_pymc`
+        Keyword arguments for :func:`pymc.aesara.compile_pymc`.
 
     Returns
     -------
-    out : list of ndarrays
+    list of ndarray
         A list of numpy arrays.
 
     Examples


ricardoV94 (approved):

OriolAbril:
do you want to update the rest of the docstring following #5459? Both yes or no are fine, they just influence the scope of the review

purna135:
Yes @OriolAbril, I'm happy to update the remainder of the docstring, but I'm not sure how to set the proper type for `vars`, what is default for `mode` as it's a optional.
Could you kindly guide me how to find the default value of a variable? 
I will work on all the methords present in [website](https://docs.pymc.io/en/latest/api.html) following https://github.com/pymc-devs/pymc/issues/5459.

pymc/sampling.py
```diff
@@ -2128,7 +2128,7 @@ def draw(
     vars
```
OriolAbril:
```suggestion
    vars : tensor_like or iterable of tensor_like
```

vars is missing the type but it is required, so it should have no optional nor default after the type.

pymc/sampling.py
```diff
@@ -2128,7 +2128,7 @@ def draw(
...
         A variable or a list of variables for which to draw samples.
     draws : int
```
OriolAbril:
```suggestion
    draws : int, default 1
```

draws is optional, so it should have either optional or default. As here the default is of the documented type and a specific value, using default instead of optional is preferred, as explained in the third point of https://pymc-data-umbrella.xyz/en/latest/sprint/docstring_tutorial.html#parameters. The default would then be removed from the description below

pymc/sampling.py
```diff
@@ -2128,7 +2128,7 @@ def draw(
...
-        Number of samples needed to draw. Detaults to 500.
+        Number of samples needed to draw. Detaults to 1.
     mode
```
OriolAbril: this is missing the type and the `, optional`

pymc/sampling.py
```diff
@@ -2128,7 +2128,7 @@ def draw(
...
         The mode used by ``aesara.function`` to compile the graph.
     **kwargs
```
OriolAbril:
```suggestion
    **kwargs : dict, optional
```

OriolAbril (commented):
the return section also needs some updates following https://pymc-data-umbrella.xyz/en/latest/sprint/docstring_tutorial.html#returns-and-yields (or what is the same https://numpydoc.readthedocs.io/en/latest/format.html#returns)

pymc/sampling.py
purna135: Thank you, I will follow the same : )

pymc/sampling.py
```diff
@@ -2125,18 +2125,18 @@ def draw(
...
+    mode : str or `Mode` instance, optional
+        The mode used by :func:``aesara.function`` to compile the graph.
+    **kwargs : dict, optional
         Keyword arguments for :func:`pymc.aesara.compile_pymc`
 
     Returns
     -------
-    List[np.ndarray]
+    out : list of ndarrays
         A list of numpy arrays.
```
purna135: I'm not sure if this return statement is correct; please have a look.

pymc/sampling.py
```diff
@@ -2125,18 +2125,18 @@ def draw(
 
     Parameters
     ----------
-    vars
+    vars : tensor_like or iterable of tensor_like
```
OriolAbril: I made a mistake here before, tensor_like includes things that can be converted to aesara tensorvariable like lists or numpy arrays, that is not the case here. 

```suggestion
    vars : Variable or iterable of Variable
```

pymc/sampling.py
```diff
@@ -2125,18 +2125,18 @@ def draw(
...
         A variable or a list of variables for which to draw samples.
-    draws : int
-        Number of samples needed to draw. Detaults to 500.
-    mode
-        The mode used by ``aesara.function`` to compile the graph.
-    **kwargs
+    draws : int, default 1
+        Number of samples needed to draw.
+    mode : str or `Mode` instance, optional
```
OriolAbril: there should be no backticks in type descriptions. I am not sure if `Mode` is a common input type, if it is, this should be `Mode` without backticks nor instance and we'll add a numpydoc alias. If it is not, it should be `aesara.compile.mode.Mode` again no backticks nor instance.

pymc/sampling.py
```diff
@@ -2125,18 +2125,18 @@ def draw(
...
+        The mode used by :func:``aesara.function`` to compile the graph.
+    **kwargs : dict, optional
         Keyword arguments for :func:`pymc.aesara.compile_pymc`
 
     Returns
     -------
-    List[np.ndarray]
+    out : list of ndarrays
```
OriolAbril:
```suggestion
    list of ndarray
```
types should always be used in singular version, and in numpydoc, if there is only one output it should not be named, only the type info should be present.

pymc/sampling.py
```diff
@@ -2125,18 +2125,18 @@ def draw(
...
         A variable or a list of variables for which to draw samples.
-    draws : int
-        Number of samples needed to draw. Detaults to 500.
-    mode
-        The mode used by ``aesara.function`` to compile the graph.
-    **kwargs
+    draws : int, default 1
+        Number of samples needed to draw.
+    mode : str or `Mode` instance, optional
+        The mode used by :func:``aesara.function`` to compile the graph.
```
OriolAbril:
```suggestion
        The mode used by :func:`aesara.function` to compile the graph.
```

purna135:
Thank you very much, @OriolAbril; I'm learning a lot from you 😊😃

pymc/sampling.py
```diff
@@ -2125,18 +2125,18 @@ def draw(
...
-    draws : int
-        Number of samples needed to draw. Detaults to 500.
-    mode
-        The mode used by ``aesara.function`` to compile the graph.
-    **kwargs
-        Keyword arguments for :func:`pymc.aesara.compile_pymc`
+    draws : int, default 1
+        Number of samples needed to draw.
+    mode : str or aesara.compile.mode.Mode, optional
+        The mode used by :func:`aesara.function` to compile the graph.
```
purna135, line 2133: is this fine now?

OriolAbril: I think so but we need to wait for the readthedocs preview to be generated and check that links are correctly generated.

OriolAbril: These two links are working, the Variable and compile_pymc ones need extra work on a different pr to work, they are fine like this for now

OriolAbril (approved):

pymc/sampling.py
purna135: Thank You : )

PR #5546, opened 2022-03-04
Update pymc.TruncatedNormal docstring

purna135:
Related to #5459

This will update the [pymc.TruncatedNormal](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.TruncatedNormal.html) docstring.


canyon289 (approved):

canyon289:
cc @OriolAbril in case youd like to review

pymc/distributions/continuous.py
```diff
@@ -673,14 +673,14 @@ class TruncatedNormal(BoundedContinuous):
...
         Mean.
-    sigma: float
+    sigma : tensor_like of float, optional
         Standard deviation (sigma > 0).
-    lower: float (optional)
-        Left bound.
-    upper: float (optional)
-        Right bound.
+    lower : tensor_like of float, optional
+        Left bound. Defaults to 0.
```
OriolAbril: I think the defaults are wrong (same for upper). I'd bet the default is no bound or infinity or somehting of the sort

pymc/distributions/continuous.py
```diff
@@ -673,14 +673,14 @@ class TruncatedNormal(BoundedContinuous):
 
     Parameters
     ----------
-    mu: float
+    mu : tensor_like of float, optional
         Mean.
```
OriolAbril, line 678: This is missing the default, like sigma

OriolAbril (changes requested):
There are parameters missing. If you look at the `dist` method you'll see that `tau`, `sd` and `transform` are also accepted. I am not up to date with v4 so I don't know if maybe some of these are deprecated now (which should also be documented via the decorator) their defaults... Can someone weigh in on both deprecations and defaults?

pymc/distributions/continuous.py
purna135: should I remove the default values?

pymc/distributions/continuous.py
purna135: I was unable to find out the default value for mu and sigma.
could you please guide me, how to identify these dafult value so that it can help me for other methords.

pymc/distributions/continuous.py
purna135: I am unable to find out the default value for mu and sigma.
could you please guide me, how to identify these dafult value so that it can help me for other methords.

pymc/distributions/continuous.py
purna135: should I remove the default values?

pymc/distributions/continuous.py
OriolAbril: First check the description in the docstring, sometimes the defaults are incorrectly explained in the extended summary section. If they are not you should read the `.dist` method and follow the variables defined there and/or function calls to see what the defaults are. Here are some examples, I hope they are illustrative and you can follow the pattern for the rest.

#### mu

https://github.com/pymc-devs/pymc/blob/4223b6e0600232b1074c1959337a1bea43b570b9/pymc/distributions/continuous.py#L720

it is defined straight away here, so it seems like it is not optional actually. @ricardoV94 is this intended? If the value in the function call were 0 instead of None then you would use the `default 0` instead of optional for example, however, if some function is called to get the default, it probably means you should use optional and explain the default (or combinations of defaults depending on other parameters) in the description

#### lower

https://github.com/pymc-devs/pymc/blob/4223b6e0600232b1074c1959337a1bea43b570b9/pymc/distributions/continuous.py#L724

the default in the function call is None, and when it is None, what is actually used is - infinity. When a value is given, that value is used. So lower should be `lower : tensor_like of float, default -numpy.inf` (and check in the readthedocs preview that it gets rendered correctly).

pymc/distributions/continuous.py
OriolAbril: no, you can review your own PR and ask questions about the parameters or defaults you don't see clear

pymc/distributions/continuous.py
ricardoV94: Looks like the default of mu should indeed be zero, not None

ricardoV94: transform as a positional argument is outdated an should be removed from the `dist` signature

purna135:
Hi @OriolAbril, could you please verify if the doc for `sigma` and `tau` is correct.
also I noticed there is no description for `sd` in [pymc.Normal](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Normal.html#pymc-normal)
https://github.com/pymc-devs/pymc/blob/4223b6e0600232b1074c1959337a1bea43b570b9/pymc/distributions/continuous.py#L549

and [pymc.TruncatedNormal.logp](https://docs.pymc.io/en/latest/api/distributions/generated/classmethods/pymc.TruncatedNormal.logp.html#pymc.TruncatedNormal.logp) takes `mu`, `sigma`, `lower`, `upper` as parameter but these are not mentioned in parameter list, should we add these parameter also?

image

purna135:
default `-numpy.inf` is not rendering properly should we use `numpy.NINF` ?

OriolAbril:
> also I noticed there is no description for sd in [pymc.Normal](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Normal.html#pymc-normal)

maybe it is deprecated? I don't know why there would be two equivalent arguments :thinking: let's see if someone else knows.

> [pymc.TruncatedNormal.logp](https://docs.pymc.io/en/latest/api/distributions/generated/classmethods/pymc.TruncatedNormal.logp.html#pymc.TruncatedNormal.logp) takes mu, sigma, lower, upper as parameter but these are not mentioned in parameter list, should we add these parameter also?

Ignore all the methods (in distributions) that are not `.dist`. They should not be used directly, there is no need to document them. We are looking into removing them from the docs, but it's a bit tricky.

> default -numpy.inf is not rendering properly should we use numpy.NINF ?

Try `- numpy.inf`, maybe the space fixes the rendering. The uppercase feels a bit strange personally, kind of breaks the symmetry.

OriolAbril:
When both sd and sigma are present, sd should not only not be documented but removed completely, in general. Therefore, here sd should be removed and consequently should not be documented

purna135:
> When both sd and sigma are present, sd should not only not be documented but removed completely, in general. Therefore, here sd should be removed and consequently should not be documented

Hi @OriolAbril, I reoved `sd` completly, please have a look if this is the right way to do so.

OriolAbril (approved):

ricardoV94:
> When both sd and sigma are present, sd should not only not be documented but removed completely, in general. Therefore, here sd should be removed and consequently should not be documented

I think the sd has been debated repeatedly and folks didn't want to remove it last time it was brought up

ricardoV94:
> > When both sd and sigma are present, sd should not only not be documented but removed completely, in general. Therefore, here sd should be removed and consequently should not be documented
> 
> I think the sd has been debated repeatedly and folks didn't want to remove it last time it was brought up

Seems like people are happy to remove them now. We should probably do it in a separate PR and for all the distributions that have the sigma/sd duplication at once

purna135:
> Seems like people are happy to remove them now. We should probably do it in a separate PR and for all the distributions that have the sigma/sd duplication at once

I am updating all the docstrings, can I remove `sd` in this PR only ?

purna135:
https://pymc--5546.org.readthedocs.build/en/5546/api/distributions/generated/pymc.TruncatedNormal.html#pymc.TruncatedNormal
image

-numpy.inf is still not rendering properly, what to do ?

ricardoV94:
> I am updating all the docstrings, can I remove `sd` in this PR only ?

I think it's better to leave that for a separate PR

purna135:
Sure, should I undo the removal of sd in this PR or will I avoid it from the next PR?

ricardoV94:
> Sure, should I undo the removal of sd in this PR or will I avoid it from the next PR?

Yup

OriolAbril:
@purna135 can you undo the sd change here so all sd arguments are removed in #5583? We'll then merge the PR

purna135:
Sure

purna135:
Hi @OriolAbril, it's done now, please have a look

OriolAbril (approved):

PR #5552, opened 2022-03-06
Update pymc.StudentT docsting

purna135:
Related to #5459 

This will update the [pymc.StudentT](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.StudentT.html) docsting


purna135:
`def dist(cls, nu, mu=0, lam=None, sigma=None, sd=None, *args, **kwargs)`

dist took `sd` parameter but this is not listed in parameter list, should we add `sd` to parameter list ?

OriolAbril (commented):
close figs and remove sd argument

purna135:
In my previous PR https://github.com/pymc-devs/pymc/pull/5550 which is already merged, I missed the `:context: close-figs` so corrected in this PR

OriolAbril (approved):

PR #5553, opened 2022-03-06
Update pymc.Cauchy docsting

purna135:
Related to #5459 

This will update the docstring of [pymc.Cauchy](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Cauchy.html)


OriolAbril (commented):
only close figs context needed

OriolAbril (approved):

PR #5555, opened 2022-03-06
Update pymc.HalfStudentT docsting

purna135:
Related to #5459 

This will update the docsting of [pymc.HalfStudentT](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.HalfStudentT.html)


michaelosthege:
@purna135 please don't open multiple PRs for making similar changes. Every PR causes at least 2 CI runs of the full test suite. This is a LOT of compute and energy resources.

purna135:
I am sorry, @michaelosthege, I was unaware of this. The same questions came to mind for me as well, so I inquired in the discussion and am waiting for an answer.

So, can I edit all of the docstrings in a file and submit a single PR?

michaelosthege:
Yes, you can combine these things in the same PR. As multiple commits (you could cherrypick them from the other branches) or as one commit if all changes are logically the same. For example example if they are all about fixing the math syntax.
In PyMC we're not that strict, but some open source projects have strong requirements for logically separated commits. For example one commit "Fix punctuation in docstrings" and another "Repair whitespace in docstring type hints".

The rule of thumb I use is to make separate PRs if, for example one is a refactoring/fix that does not need much discussion and the other is something like a new feature where I expect a longer review/discussion process.

Hope this helps you for the future :)

purna135:
Thank you @michaelosthege, I will follow the same : )

OriolAbril (commented):
here changes needed are removal of sd argument and addition of close-figs

OriolAbril:
Also for new PRs, if you have new PRs in mind, now that you are familiar with the contributing process you should modify the docstrings for 4-6 distributions at the same time, which will be easier for you to manage as you won't have 6 branches between which to switch and faster for us to review.

purna135:
> Also for new PRs, if you have new PRs in mind, now that you are familiar with the contributing process you should modify the docstrings for 4-6 distributions at the same time, which will be easier for you to manage as you won't have 6 branches between which to switch and faster for us to review.

Sure @OriolAbril, Thanks 
After these PRs, I will update all the documents in a batch of 5-6 update per PR.

OriolAbril (approved):

PR #5562, opened 2022-03-07
Initiate v4.0.0b3 release

fonnesbeck:
Version bump for v4.0.0b3

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -129,6 +129,9 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
 - `softmax` and `log_softmax` functions added to `math` module (see [#5279](https://github.com/pymc-devs/pymc/pull/5279)).
 - ...
 
+### Maintenance
+- Variational inference has been ported to v4.
+
 ## Documentation
 - Switched to the [pydata-sphinx-theme](https://pydata-sphinx-theme.readthedocs.io/en/latest/)
 - Updated our documentation tooling to use [MyST](https://myst-parser.readthedocs.io/en/latest/), [MyST-NB](https://myst-nb.readthedocs.io/en/latest/), sphinx-design, notfound.extension,
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -7,10 +7,10 @@ Instead update the vNext section until 4.0.0 is out.
 ⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠⚠
 -->
 
-## PyMC vNext (4.0.0b1 → 4.0.0b2 → 4.0.0b3 → 4.0.0)
+## PyMC vNext (4.0.0b1 → 4.0.0b2 → 4.0.0b3 → 4.0.0b4 → 4.0.0)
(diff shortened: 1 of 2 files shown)


michaelosthege (approved):
We should prominently mention the change of interpretation for the size kwarg with multivariate distributions.
And that VI is available again.

ghost:
@michaelosthege are those not in the release notes, or are you just talking about the text associated with the release on the releases page?

twiecki:
I think we forgot to add them, VI is still in the list of unported.

fonnesbeck:
The size kwarg change is already in the release notes.

RELEASE-NOTES.md
```diff
@@ -129,6 +129,9 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
 - `softmax` and `log_softmax` functions added to `math` module (see [#5279](https://github.com/pymc-devs/pymc/pull/5279)).
 - ...
 
+### Maintenance
```
twiecki: Upon reconsideration, we probably don't need to mention it at all because it doesn't add new functionality, we just need re move it from the list of not-working things.

ghost: Probably fine to leave it under maintenance. Users likely will not notice something missing from the non-working list but will notice it here, and VI is a pretty big feature to have working again.

ricardoV94:
Cholesky distributions also still show up in the not yet working, but they are!

RELEASE-NOTES.md
ricardoV94: Oops, I pushed a commit that removed this change. I do agree with @twiecki, this changelist should reflect the difference between current and v3, and as such it does not make sense to say something is still working...?

ricardoV94:
Actually... several algorithms of VI are not yet working... so maybe we should replace the old notice by something like:
```
### Not-yet working features
...
- Some Variational inference algorithms (see PR [#4582](https://github.com/pymc-devs/pymc/pull/4582))
...
```

ghost:
Which are not working, and do we plan on making them work or are they gone forever?

twiecki:
Gone forever, most likely.

ricardoV94:
Normalizing flows: https://github.com/pymc-devs/pymc/blob/ef5f91b41d733c64cf3ea785ec2760caae5d1bd4/pymc/variational/approximations.py#L363

ricardoV94:
And these: https://github.com/pymc-devs/pymc/blob/e987950b359edfe4a49c92ee67c57983fe0fc5d9/pymc/variational/opvi.py#L959-L962

ricardoV94:
And, finally, this one, which seems minor: https://github.com/pymc-devs/pymc/blob/e987950b359edfe4a49c92ee67c57983fe0fc5d9/pymc/variational/approximations.py#L274-L277

RELEASE-NOTES.md
```diff
@@ -129,6 +127,7 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
 - `softmax` and `log_softmax` functions added to `math` module (see [#5279](https://github.com/pymc-devs/pymc/pull/5279)).
 - ...
 
+
```
ricardoV94, line 130: My bad: 
```suggestion
```

ghost:
OK, should we leave these with NotImplemented exceptions? If Thomas is correct in anticipating they won't be revived any time soon, perhaps they should be removed from the code base, given that this is a major release.

twiecki:
I think we should remove it, yes.

twiecki:
But we can merge this.

PR #5564, opened 2022-03-07
Fix nightly upload pypi error (again)

theorashid:
This is for https://github.com/pymc-devs/pymc/issues/5548. Simple change of secret for the nightly uploads. Should do the job @twiecki @fonnesbeck

(That closed PR was because I created a branch from an old branch rather than main, so it had old commits. This should be clean.)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -27,7 +27,7 @@ jobs:
         uses: pypa/gh-action-pypi-publish@v1.5.0
         with:
           user: __token__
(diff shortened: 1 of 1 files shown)


twiecki:
Thanks @theorashid , let's see :)

theorashid:
One final comment: whilst pymc v4 is under beta, is it easier for users to pip install pymc-nightly rather than cloning the repo?

twiecki:
Yes, definitely.

PR #5566, opened 2022-03-08
Removes BART

aloctavodia:
This removes BART from the main PyMC codebase in favor of moving it to [pymc-experimental.](https://github.com/pymc-devs/pymc-experimental)


ricardoV94:
We need to remove the test file from the pytest workflow:

https://github.com/pymc-devs/pymc/blob/afe210acc8b90fef718eb4d28da641996c278c91/.github/workflows/pytest.yml#L80

https://github.com/pymc-devs/pymc/blob/afe210acc8b90fef718eb4d28da641996c278c91/.github/workflows/pytest.yml#L66

ricardoV94:
What about the competence? Is there a level above `Ideal` that we can give to BART variables?

ricardoV94:
Can you give more context what changes with this commit? https://github.com/pymc-devs/pymc/commit/159e7bfe059ff9c866095c555346ae83e7ef7805

ricardoV94 (commented):
Looks good,

Can you add a test for the step methods change, that shows you can add a step method externally?

Also can you split the bart changes from the step/methods assignment into 2 distinct commits (or PRs)?

aloctavodia:
What should I do to split the changes into 2 distinct commits (or PRs)?

ricardoV94:
> What should I do to split the changes into 2 distinct commits (or PRs)?

git interactive rebase -> reorder + squash all the commits into two commits: 1) remove bart and 2) step method changes

ricardoV94:
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

pymc/sampling.py
```diff
@@ -490,29 +465,12 @@ def sample(
...
-                **kwargs,
-            )
-        except (AttributeError, NotImplementedError, tg.NullTypeGradError):
-            # gradient computation failed
-            _log.info("Initializing NUTS failed. Falling back to elementwise auto-assignment.")
-            _log.debug("Exception in init nuts", exc_info=True)
-            step = assign_step_methods(model, step, step_kwargs=kwargs)
-    else:
-        step = assign_step_methods(model, step, step_kwargs=kwargs)
+<<<<<<< HEAD
```
ricardoV94: Some conflict lines here

ricardoV94 (changes requested):
Looks great.

Can you change the commit messages to have a more descriptive title/messages? You don't need to include the titles of the squashed commits unless you want to. This can be done with interactive rebase + reword command

aloctavodia:
I was confused with how to change the messages. Now I think I got it right.

pymc/sampling.py
```diff
@@ -490,29 +465,7 @@ def sample(
...
-                **kwargs,
-            )
-        except (AttributeError, NotImplementedError, tg.NullTypeGradError):
-            # gradient computation failed
-            _log.info("Initializing NUTS failed. Falling back to elementwise auto-assignment.")
-            _log.debug("Exception in init nuts", exc_info=True)
-            step = assign_step_methods(model, step, step_kwargs=kwargs)
-    else:
-        step = assign_step_methods(model, step, step_kwargs=kwargs)
+    step = assign_step_methods(model, step, methods=pm.STEP_METHODS, step_kwargs=kwargs)
```
ricardoV94, line 468: Does it still add all variables to NUTS when that's possible? If so, any idea why we had this two step strategy before?

aloctavodia: Besides the tests, I run a few models manually and assignment seems to be working properly both when only NUTS is used or when we have more than one sampler.

ricardoV94 (approved):
Looks great. Award for best branch name! Just checking what is going on with read the docs failure

ricardoV94:
I merged it but just realized we need to mention it in the release notes @aloctavodia

PR #5572, opened 2022-03-10
Stack array stats if possible

aloctavodia:
BART (that now lives in pymc-experimental) stores 2 stats which at each point are arrays. Variable importance and bart_trees. For example variable importance is a vector with the position encoding a given covariate and an integer encoding the number of times that covariate was used in the trees. 

We used to have this to stack those stats https://github.com/pymc-devs/pymc/pull/5566/files#diff-7b3470589153c7c90a2c05fc86402f9acf65555200d90f01b51a47ccef592628L637-L654 but that code is BART-specific, so here I am proposing as alternative a more general solution.

### DIFF
--- a/pymc/backends/base.py
+++ b/pymc/backends/base.py
@@ -204,10 +204,18 @@ class BaseTrace(ABC):
         vals = np.stack(
             [self._get_sampler_stats(stat_name, i, burn, thin) for i in sampler_idxs], axis=-1
         )
+
         if vals.shape[-1] == 1:
-            return vals[..., 0]
-        else:
-            return vals
+            vals = vals[..., 0]
+
+        if vals.dtype == np.dtype(object):
+            try:
+                vals = np.vstack(vals)
+            except ValueError:
+                # Most likely due to non-identical shapes. Just stick with the object-array.
+                pass
+
+        return vals
 
     def _get_sampler_stats(self, stat_name, sampler_idx, burn, thin):
         """Get sampler statistics."""
--- a/pymc/step_methods/mlda.py
+++ b/pymc/step_methods/mlda.py
@@ -946,28 +946,22 @@ def extract_Q_estimate(trace, levels):
     MLDA with variance reduction has been used for sampling.
     """
 
-    Q_0_raw = trace.get_sampler_stats("Q_0")
-    # total number of base level samples from all iterations
-    total_base_level_samples = sum(it.shape[0] for it in Q_0_raw)
-    Q_0 = np.concatenate(Q_0_raw).reshape((1, total_base_level_samples))
+    Q_0_raw = trace.get_sampler_stats("Q_0").squeeze()
+    Q_0 = np.concatenate(Q_0_raw)[None, ::]
     ess_Q_0 = az.ess(np.array(Q_0, np.float64))
     Q_0_var = Q_0.var() / ess_Q_0
 
     Q_diff_means = []
     Q_diff_vars = []
     for l in range(1, levels):
-        Q_diff_raw = trace.get_sampler_stats(f"Q_{l}_{l-1}")
-        # total number of samples from all iterations
-        total_level_samples = sum(it.shape[0] for it in Q_diff_raw)
-        Q_diff = np.concatenate(Q_diff_raw).reshape((1, total_level_samples))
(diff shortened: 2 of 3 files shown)


OriolAbril (commented):
Code looks good but I don't really know how it will integrate with the rest of the library, hope these two questions help a bit.

Do we have tests with compound steps similar to https://github.com/pymc-devs/pymc/issues/4602, does this modify this?

Would it be helpful to try and provide stepper defined dims or coords?

aloctavodia:
This is not directly related to compound steps, but to samplers/steps that have stats that are arrays and not floats/integer/boolean etc. So far it seems that the only  samplers with that kind of stats are from PGBART and MLDA.

pymc/backends/arviz.py
```diff
@@ -165,6 +165,17 @@ def __init__(
...
                     )
             self.ntune = len(self.trace) - self.ndraws
+
+            for strace in self.trace._straces.values():
+                if strace._stats is not None:
+                    for stat in strace._stats:
+                        for k, v in stat.items():
+                            if isinstance(v, np.ndarray) and isinstance(v[0], np.ndarray):
+                                if self.nchains > 1:
+                                    stat[k] = np.vstack(v)
```
michaelosthege: Here I have a hard time understanding what the code does.
Mostly because the things are being modified inplace.

**Over which dimension are you stacking things and why?** Isn't this just replacing arrays by vstacked versions of themself? I mean the line that's not test-covered.

pymc/backends/arviz.py
```diff
@@ -165,6 +165,17 @@ def __init__(
                         UserWarning,
                     )
             self.ntune = len(self.trace) - self.ndraws
+
+            for strace in self.trace._straces.values():
+                if strace._stats is not None:
+                    for stat in strace._stats:
```
michaelosthege: `BaseTrace` does not have a `_stats` attribute !

It's just the `NDArray` that has it.
https://github.com/pymc-devs/pymc/blob/44c5495aec159b96d01b20c125c2f4b582bbd331/pymc/backends/base.py#L40-L84

michaelosthege (changes requested):
> Would it be helpful to try and provide stepper defined dims or coords?

Generally yes. Alongside shape information.

@aloctavodia why do you want to stack these variables at all?
They should be yielded in the correct type right away. We don't have a stats-postprocessing API and I'd prefer to undertstand what you're trying to achieve first.

pymc/backends/arviz.py
aloctavodia: Currently if the stat is an array, we get and array of arrays, instead of an array of the correct shape. Thus the solution proposed here is to stack the array of arrays.

Happy to fix this with a different approach, but not sure what that approach should be.

michaelosthege: Probably something here: https://github.com/pymc-devs/pymc/blob/main/pymc/backends/base.py#L204-L206

This code relies on the standard `BaseTrace` API and in that retrieval mechanism it should be okay to attempt a stacking & fall back to an array of arrays if the stacking doesn't work.

pymc/backends/base.py
```diff
@@ -204,10 +204,14 @@ def get_sampler_stats(self, stat_name, sampler_idx=None, burn=0, thin=1):
...
         )
+
         if vals.shape[-1] == 1:
-            return vals[..., 0]
-        else:
-            return vals
+            vals = vals[..., 0]
+
+        if isinstance(vals, np.ndarray) and isinstance(vals[0], np.ndarray):
+            vals = np.vstack(vals)
```
michaelosthege: Due to the `vals.shape` above it's safe to assume `isinstance(vals, np.ndarray)`.

The `isinstance(vals[0], np.ndarray)` is also true for any 2D array. We'll have to check the dtype instead.

```suggestion
        if vals.dtype == np.dtype(object):
            try:
                vals = np.vstack(vals)
            except ValueError:
                # Most likely due to non-identical shapes.
                # Just stick with the object-array.
                pass
```

michaelosthege (commented):
I didn't check why the CI failed, but I'm quite certain about the `isinstance` checks I commented above.

michaelosthege (approved):

PR #5577, opened 2022-03-11
Return an empty string for empty model representation

### LINKED ISSUE #5568: Representation for empty model raises ValueError
```python
import pymc as pm
pm.Model().str_repr()
```

```python
  File "/home/ricardo/miniconda3/envs/roche_preclinical_hd/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3444, in run_code
  File "<ipython-input-5-fa2b838ad5d3>", line 1, in <module>
    pm.Model().str_repr()
  File "/home/ricardo/Documents/Projects/pymc/pymc/printing.py", line 82, in str_for_model
    maxlen = str(max(len(x) for x in names))
ValueError: max() arg is an empty sequence
```

sagartomar:
… instead of raising an error

Should fix #5568 

### DIFF
--- a/pymc/printing.py
+++ b/pymc/printing.py
@@ -58,6 +58,9 @@ def str_for_dist(rv: TensorVariable, formatting: str = "plain", include_params: 
 def str_for_model(model: Model, formatting: str = "plain", include_params: bool = True) -> str:
     """Make a human-readable string representation of Model, listing all random variables
     and their distributions, optionally including parameter values."""
+    if len(model.unobserved_RVs) + len(model.observed_RVs) + len(model.potentials) == 0:
+        return ""
+
     all_rv = itertools.chain(model.unobserved_RVs, model.observed_RVs, model.potentials)
 
     rv_reprs = [rv.str_repr(formatting=formatting, include_params=include_params) for rv in all_rv]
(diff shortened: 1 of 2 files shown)


pymc/printing.py
```diff
@@ -58,6 +58,9 @@ def str_for_dist(rv: TensorVariable, formatting: str = "plain", include_params:
 def str_for_model(model: Model, formatting: str = "plain", include_params: bool = True) -> str:
     """Make a human-readable string representation of Model, listing all random variables
     and their distributions, optionally including parameter values."""
+    if len(model.unobserved_RVs) + len(model.observed_RVs) + len(model.potentials) == 0:
```
ricardoV94: Instead of adding this verbose `if statement` you can check if `rv_reprs` below is an empty list before going into the the `if "latex"... else` statement.

ricardoV94 (changes requested):
Thanks for opening the PR, I left a suggestion below. Also we should add a test to make sure the original error no longer happens

pymc/printing.py
sagartomar: Thanks for the feedback. I have one question. There doesn't seem to be any existing test for this function so I am not sure in which file to write the test. Should I create a new test_printing file for just this one test or is there any existing file where I can add it?

ricardoV94: Good question. That relates to #5579, in the meantime you can perhaps add it to `test_model.py`?

ricardoV94 (approved):

ricardoV94:
Thanks @sagartomar!

PR #5584, opened 2022-03-11
Replace zero inflated distributions with mixtures

ricardoV94:
Required the following extra changes:

* Implement logcdf for Mixtures
* Allow Constant distribution to have a varying dtype

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -74,6 +74,7 @@ All of the above apply to:
   - The function `replace_with_values` function has been added to `gp.utils`.
   - `MarginalSparse` has been renamed `MarginalApprox`.
   - Removed `MixtureSameFamily`. `Mixture` is now capable of handling batched multivariate components (see [#5438](https://github.com/pymc-devs/pymc/pull/5438)).
+  - `ZeroInflatedPoisson` `theta` parameter was renamed to `mu` (see [#5584](https://github.com/pymc-devs/pymc/pull/5584)).
 - ...
 
 ### Expected breaks
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -46,6 +46,7 @@ from pymc.distributions.distribution import Discrete
 from pymc.distributions.logprob import logcdf, logp
 from pymc.distributions.shape_utils import rv_size_is_none
 from pymc.math import sigmoid
+from pymc.vartypes import continuous_types
 
 __all__ = [
     "Binomial",
@@ -1315,8 +1316,11 @@ class ConstantRV(RandomVariable):
     name = "constant"
     ndim_supp = 0
     ndims_params = [0]
-    dtype = "floatX"  # Should be treated as a discrete variable!
     _print_name = ("Constant", "\\operatorname{Constant}")
+
+    def make_node(self, rng, size, dtype, c):
+        c = at.as_tensor_variable(c)
+        return super().make_node(rng, size, c.dtype, c)
 
     @classmethod
     def rng_fn(cls, rng, c, size=None):
@@ -1334,15 +1338,19 @@ class Constant(Discrete):
 
     Parameters
     ----------
-    value: float or int
-        Constant parameter.
+    c: float or int
+        Constant parameter. The dtype of `c` determines the dtype of the distribution.
+        This can affect which sampler is assigned to Constant variables, or variables
+        that use Constant, such as Mixtures.
     """
 
     rv_op = constant
 
     @classmethod
     def dist(cls, c, *args, **kwargs):
-        c = at.as_tensor_variable(floatX(c))
+        c = at.as_tensor_variable(c)
+        if c.dtype in continuous_types:
+            c = floatX(c)
         return super().dist([c], **kwargs)
 
     def get_moment(rv, size, c):
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -43,7 +43,8 @@ from pymc.distributions.dist_math import (
     normal_lcdf,
 )
 from pymc.distributions.distribution import Discrete
-from pymc.distributions.logprob import logcdf, logp
+from pymc.distributions.logprob import logp
+from pymc.distributions.mixture import Mixture
 from pymc.distributions.shape_utils import rv_size_is_none
 from pymc.math import sigmoid
 from pymc.vartypes import continuous_types
@@ -1386,22 +1387,24 @@ class Constant(Discrete):
         )
 
 
-class ZeroInflatedPoissonRV(RandomVariable):
-    name = "zero_inflated_poisson"
-    ndim_supp = 0
-    ndims_params = [0, 0]
-    dtype = "int64"
-    _print_name = ("ZeroInflatedPois", "\\operatorname{ZeroInflatedPois}")
+def _zero_inflated_mixture(*, name, nonzero_p, nonzero_dist, **kwargs):
+    """Helper function to create a zero-inflated mixture
 
-    @classmethod
-    def rng_fn(cls, rng, psi, lam, size):
-        return rng.poisson(lam, size=size) * (rng.random(size=size) < psi)
+    If name is `None`, this function returns an unregistered variable
+    """
+    nonzero_p = at.as_tensor_variable(floatX(nonzero_p))
+    weights = at.stack([1 - nonzero_p, nonzero_p], axis=-1)
+    comp_dists = [
+        Constant.dist(0),
+        nonzero_dist,
+    ]
+    if name is not None:
+        return Mixture(name, weights, comp_dists, **kwargs)
+    else:
+        return Mixture.dist(weights, comp_dists, **kwargs)
 
 
-zero_inflated_poisson = ZeroInflatedPoissonRV()
-
-
-class ZeroInflatedPoisson(Discrete):
+class ZeroInflatedPoisson:
     R"""
     Zero-inflated Poisson log-likelihood.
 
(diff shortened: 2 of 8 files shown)


twiecki:
Nice refactor.

pymc/distributions/discrete.py
```diff
@@ -1370,23 +1377,28 @@ def logp(value, c):
...
+    nonzero_p = at.as_tensor_variable(floatX(nonzero_p))
+    weights = at.stack([1 - nonzero_p, nonzero_p], axis=-1)
+    comp_dists = [
+        Constant.dist(0),
+        nonzero_dist,
+    ]
+    if name is not None:
+        return Mixture(name, weights, comp_dists, **kwargs)
+    else:
+        return Mixture.dist(weights, comp_dists, **kwargs)
```
larryshamalama, line 1405: Is `name = None` here an indicator on how the user is using the API?

pymc/distributions/discrete.py
```diff
@@ -1370,23 +1377,28 @@ def logp(value, c):
...
+        )
 
 
-zero_inflated_poisson = ZeroInflatedPoissonRV()
+def _zero_inflated_mixture(*, name, nonzero_p, nonzero_dist, **kwargs):
+    nonzero_p = at.as_tensor_variable(floatX(nonzero_p))
+    weights = at.stack([1 - nonzero_p, nonzero_p], axis=-1)
+    comp_dists = [
+        Constant.dist(0),
+        nonzero_dist,
```
larryshamalama, line 1399: Does `nonzero_dist` here have to be `Discrete`?

pymc/distributions/discrete.py
```diff
@@ -1342,7 +1347,9 @@ class Constant(Discrete):
 
     @classmethod
     def dist(cls, c, *args, **kwargs):
-        c = at.as_tensor_variable(floatX(c))
+        c = at.as_tensor_variable(c)
+        if c.dtype in continuous_types:
+            c = floatX(c)
```
larryshamalama, line 1354: Just for my own curiosity, in what situations would it help that we allow varying `dtype` here?

larryshamalama (commented):
Overall looks great! Left some questions so I can better understand myself 😅 (and maybe for others reading this PR!)

An overarching question that I have is what benefits do we get from positing a Mixture-type rv_op for zero-inflated distributions? My understanding was that Mixture distributions inherit from `SymbolicDistribution` as a way to initialize a different rv_op depending on how many and what `comp_dists` are provided. However, for zero-inflated distributions, there will always be "two" components: the zero Constant distribution and Poisson/binomial/negative binomial. Perhaps the answer is simple but I was just wondering...

pymc/distributions/discrete.py
canyon289: I have same question and same guess, Whatever the answer is here please add it as a docstring, itll make the code easier to understand later

pymc/distributions/mixture.py
```diff
@@ -404,6 +403,38 @@ def marginal_mixture_logprob(op, values, rng, weights, *components, **kwargs):
...
+    else:
+        components_logcdf = at.stack(
+            [logcdf(component, value) for component in components],
+            axis=-1,
+        )
+
+    mix_logcdf = at.logsumexp(at.log(weights) + components_logcdf, axis=-1)
+
+    # Squeeze stack dimension
+    # There is a Aeasara bug in squeeze with negative axis
```
canyon289:
```suggestion
    # There is a Aesara bug in squeeze with negative axis
```

pymc/distributions/discrete.py
ricardoV94: I think your question might be related to https://github.com/pymc-devs/pymc/issues/4511

Still need to make sure we are not combining discrete and continuous distributions in mixtures. In the cases here they are both supposed to be discrete yes

pymc/distributions/discrete.py
ricardoV94: PyMC uses the dtype of the variables to decide what sampler to use. We want discrete samplers to be used with the zero inflated mixtures, so Constant must be allowed to be discrete. 

However we might want Constant to also take float values in other circumstances so I left that as an option.

ricardoV94:
> An overarching question that I have is what benefits do we get from positing a Mixture-type rv_op for zero-inflated distributions

The biggest advantage is less code for us to maintain. As Thomas hinted above this is just a refactor.

Reusing a building block also means any bug fixes there propagate everywhere quickly. For instance, the old RandomVariable Ops actually had a slight bug in that they did not broadcast psi with the other parameters when size was not specified. So `pm.ZIP.dist(0.5, [200, 300])` Could never draw mixed zeros and non zeros, even though it should have. 

Instead of having to fix this everywhere, it's now enough to do it (and test it) in one place as I did here: https://github.com/pymc-devs/pymc/pull/5584/commits/78f1786d30a4784d0f541d19bc60acfafe27090b

pymc/distributions/discrete.py
ricardoV94: The name=None was an internal convention to distinguish between the two calls with as little repetition as possible. 

This is not meant to be a user facing function, hence the leading underscore. 

If you have better ideas let me know, I couldn't think of anything better. I'll add a comment explaining it more explicitly

pymc/distributions/discrete.py
twiecki: Maybe add a comment about that.

pymc/distributions/discrete.py
```diff
@@ -1437,97 +1449,19 @@ class ZeroInflatedPoisson(Discrete):
...
-
     @classmethod
-    def rng_fn(cls, rng, psi, n, p, size):
-        return rng.binomial(n=n, p=p, size=size) * (rng.random(size=size) < psi)
-
-
-zero_inflated_binomial = ZeroInflatedBinomialRV()
+    def dist(cls, psi, theta, **kwargs):
+        return _zero_inflated_mixture(
+            name=None, nonzero_p=psi, nonzero_dist=Poisson.dist(mu=theta), **kwargs
```
twiecki: Should we rename `theta` to `mu` to be consistent?

pymc/distributions/mixture.py
twiecki: Maybe link to the issue so that we know when we can remove it.

twiecki:
What about the integrated versions? I guess we get those automatically once it's added to `aeppl`?

ricardoV94:
> What about the integrated versions? I guess we get those automatically once it's added to `aeppl`?

What do you mean integrated?

twiecki:
> What do you mean integrated?

Upon thinking about it I don't think it makes sense here.

pymc/distributions/discrete.py
ricardoV94: I think so

pymc/distributions/discrete.py
ricardoV94: Added to the docstrings

pymc/distributions/discrete.py
ricardoV94: Added docstrings

pymc/distributions/mixture.py
ricardoV94: Done

pymc/distributions/discrete.py
ricardoV94: Done

pymc/distributions/discrete.py
ricardoV94: Did you have another thing in mind?

pymc/distributions/discrete.py
```diff
@@ -1396,8 +1414,8 @@ class ZeroInflatedPoisson(Discrete):
 
     .. math::
 
-        f(x \mid \psi, \theta) = \left\{ \begin{array}{l}
-            (1-\psi) + \psi e^{-\theta}, \text{if } x = 0 \\
+        f(x \mid \psi, \mu) = \left\{ \begin{array}{l}
```
twiecki, line 1417: For completeness should add this to release notes. And maybe even then we can just remove the depreciation warning.

ricardoV94: Yeah forgot the release notes

RELEASE-NOTES.md
```diff
@@ -74,6 +74,7 @@ All of the above apply to:
   - The function `replace_with_values` function has been added to `gp.utils`.
   - `MarginalSparse` has been renamed `MarginalApprox`.
   - Removed `MixtureSameFamily`. `Mixture` is now capable of handling batched multivariate components (see [#5438](https://github.com/pymc-devs/pymc/pull/5438)).
+  - `ZeroInflatedPoisson``theta` parameter was renamed to `mu` (see [#5584](https://github.com/pymc-devs/pymc/pull/5584)).
```
```suggestion
  - `ZeroInflatedPoisson` `theta` parameter was renamed to `mu` (see [#5584](https://github.com/pymc-devs/pymc/pull/5584)).
```

pymc/distributions/discrete.py
larryshamalama: This is exactly what I had in mind! I was wondering if it would help to have a check that `nonzero_dist` is discrete, but, as you mentioned above, it may not be necessary since `_zero_inflated_mixture` is not a user-facing function.

larryshamalama:
> The biggest advantage is less code for us to maintain. As Thomas hinted above this is just a refactor.
> 
> Reusing a building block also means any bug fixes there propagate everywhere quickly. For instance, the old RandomVariable Ops actually had a slight bug in that they did not broadcast psi with the other parameters when size was not specified. So `pm.ZIP.dist(0.5, [200, 300])` Could never draw mixed zeros and non zeros, even though it should have.
> 
> Instead of having to fix this everywhere, it's now enough to do it (and test it) in one place as I did here: [78f1786](https://github.com/pymc-devs/pymc/commit/78f1786d30a4784d0f541d19bc60acfafe27090b)

Okay I see, thanks for the clarification! Makes a lot of sense

pymc/distributions/discrete.py
ricardoV94: Yes, and I want to add an explicit check in the Mixture class in a follow up PR

twiecki (approved):

PR #5589, opened 2022-03-12
Fix typing errors in `sampling`, `smc`, `hmc` and `nuts`

michaelosthege:
In some places I renamed local variables to have fewer typing conflicts.

In other places, I simplified if/else branching based on stronger typing.

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1443,8 +1443,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
         if dims is not None:
             if isinstance(dims, str):
                 dims = (dims,)
-            if any(dim not in self.coords and dim is not None for dim in dims):
-                raise ValueError(f"Dimension {dim} is not specified in `coords`.")
+            for dim in dims:
+                if dim not in self.coords and dim is not None:
+                    raise ValueError(f"Dimension {dim} is not specified in `coords`.")
             if any(var.name == dim for dim in dims):
                 raise ValueError(f"Variable `{var.name}` has the same name as its dimension label.")
             self._RV_dims[var.name] = dims
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -749,11 +749,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
                         f"Requested variable {var} not found among the model variables"
                     )
 
-        rv_logps = []
+        rv_logps: List[TensorVariable] = []
         if rv_values:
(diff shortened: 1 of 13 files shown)


pymc/model.py
```diff
@@ -749,11 +749,10 @@ def logpt(
                         f"Requested variable {var} not found among the model variables"
                     )
 
-        rv_logps = []
+        rv_logps: List[TensorVariable] = []
         if rv_values:
             rv_logps = joint_logpt(list(rv_values.keys()), rv_values, sum=False, jacobian=jacobian)
-            if not isinstance(rv_logps, list):
-                rv_logps = [rv_logps]
+            assert isinstance(rv_logps, list)
```
ricardoV94, line 755: I think this change is wrong. `joint_logpt` returns a single variable when there is only one rv

michaelosthege: Oh. I thought it was safe to assume that when `sum=False` it returns a list.
I'll edit it accordingly.


Functions with dynamic return types are making it really hard to get types right.
Please do not add any more of them.

ricardoV94: Trust me I didn't want to add it

ricardoV94: Do we still need that assert or can mypy figure it out now?

michaelosthege: Mypy can not figure it out without either an assert, isinstance check or cast.

As I said. Functions with dynamic types are a typing nightmare.

ricardoV94: I am just uneasy with an assert. Have you seen any discussion about cast vs asserts for typing?

michaelosthege: At runtime the cast becomes a dummy function call not doing anything.
The assert takes a tiny bit longer, but with the added benefit of raising an error when the type is not actually as expected.
For that reason, and because this code is not time critical at the microsecond level, I prefer the asserts.
Note that the assert I added detected my incorrect assumption, giving me the same information as your initial comment.

pymc/step_methods/arraystep.py
```diff
@@ -50,7 +50,7 @@ class Competence(IntEnum):
 class BlockedStep(ABC):
 
     generates_stats = False
-    stats_dtypes: List[Dict[str, np.dtype]] = []
+    stats_dtypes: List[Dict[str, type]] = []
```
michaelosthege, line 53: FYI @aloctavodia here I made it less strict, because also all the other step methods had things like `np.float64` in the dict but those weren't classified as dtypes either.

twiecki (approved):

PR #5594, opened 2022-03-14
Binomial distribution allows n=0

### LINKED ISSUE #5593: logp of Binomial distribution raises error when n is zero

Following the [discourse post](https://discourse.pymc.io/t/how-does-pymc3-handle-n-0-in-binomial-discrete-likelihood-function/7900/5), the Binomial distribution works when `n=0` but the `logp` has the constraint that `n>0`.

```python
import pymc as pm
pm.logp(pm.Binomial.dist(n=0, p=0.5), 0).eval()
```
raises
```
aeppl.logprob.ParameterValueError: n > 0, 0 <= p <= 1
```

</details>

`n=0` cases are important, for example in mortality modelling, when the populaiton for a subgroup is 0 but we still want to estimate the probability of death. So we should allow `n=0` to work.

Note n=0 are common in other count distributions, so BetaBinomial, Poisson (observed counts=0), Negative Binomial, etc will need to be allow this too

* PyMC Version: 4.0.0b2

theorashid:
This closes #5593. I added four `=` so hopefully it's as easy as that.

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -144,7 +144,7 @@ class Binomial(Discrete):
             binomln(n, value) + logpow(p, value) + logpow(1 - p, n - value),
         )
 
-        return check_parameters(res, 0 < n, 0 <= p, p <= 1, msg="n > 0, 0 <= p <= 1")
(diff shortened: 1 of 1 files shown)


ricardoV94 (approved):

ricardoV94:
Thanks @theorashid!

PR #5595, opened 2022-03-14
Update docstring in ChiSquared, HalfNormal, Wald, Pareto, ExGaussion and VonMiese distribution

purna135:
Related to https://github.com/pymc-devs/pymc/issues/5459

This will update the docsting of

[pymc.ChiSquared](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.ChiSquared.html)
[pymc.HalfNormal](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.HalfNormal.html)
[pymc.Wald](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Wald.html)
[pymc.Pareto](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Pareto.html)
[pymc.ExGaussian](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.ExGaussian.html)
[pymc.VonMises](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.VonMises.html)

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -822,9 +822,10 @@ class HalfNormal(PositiveContinuous):
        the standard deviation/precision of the unfolded normal distribution, for
        the standard deviation of the half-normal distribution, see below. For
        the half-normal, they are just two parameterisation :math:`\sigma^2
-       \equiv \frac{1}{\tau}` of a scale parameter
+       \equiv \frac{1}{\tau}` of a scale parameter.
 
     .. plot::
+        :context: close-figs
 
         import matplotlib.pyplot as plt
         import numpy as np
@@ -848,10 +849,12 @@ class HalfNormal(PositiveContinuous):
 
     Parameters
     ----------
-    sigma: float
+    sigma : tensor_like of float, optional
         Scale parameter :math:`sigma` (``sigma`` > 0) (only required if ``tau`` is not specified).
-    tau: float
+        Defaults to 1.
+    tau : tensor_like of float, optional
         Precision :math:`tau` (tau > 0) (only required if sigma is not specified).
+        Defaults to 1.
 
     Examples
     --------
@@ -890,7 +893,7 @@ class HalfNormal(PositiveContinuous):
 
         Parameters
         ----------
-        value: numeric or np.ndarray or aesara.tensor
+        value : tensor_like of float
             Value(s) for which log CDF is calculated. If the log CDF for multiple
             values are desired the values must be provided in a numpy array or Aesara tensor.
 
@@ -942,6 +945,7 @@ class Wald(PositiveContinuous):
            \right\}
 
     .. plot::
+        :context: close-figs
 
         import matplotlib.pyplot as plt
         import numpy as np
@@ -974,13 +978,13 @@ class Wald(PositiveContinuous):
 
     Parameters
     ----------
-    mu: float, optional
+    mu : tensor_like of float, optional
         Mean of the distribution (mu > 0).
-    lam: float, optional
+    lam : tensor_like of float, optional
         Relative precision (lam > 0).
-    phi: float, optional
+    phi : tensor_like of float, optional
         Alternative shape parameter (phi > 0).
-    alpha: float, optional
+    alpha : tensor_like of float, default 0
         Shift/location parameter (alpha >= 0).
 
     Notes
@@ -1067,14 +1071,14 @@ class Wald(PositiveContinuous):
 
         Parameters
         ----------
-        value: numeric
+        value : tensor_like of float
             Value(s) for which log-probability is calculated. If the log probabilities for multiple
-            values are desired the values must be provided in a numpy array or Aesara tensor
-        mu: float or TensorVariable
+            values are desired the values must be provided in a numpy array or Aesara tensor.
+        mu : tensor_like of float
             Mean of the distribution (mu > 0).
-        lam: float or TensorVariable
+        lam : tensor_like of float
             Relative precision (lam > 0).
-        alpha: float or TensorVariable
+        alpha : tensor_like of float
             Shift/location parameter (alpha >= 0).
 
         Returns
@@ -1112,14 +1116,14 @@ class Wald(PositiveContinuous):
 
         Parameters
         ----------
-        value: numeric or np.ndarray or aesara.tensor
+        value : tensor_like of float
             Value(s) for which log CDF is calculated. If the log CDF for multiple
             values are desired the values must be provided in a numpy array or Aesara tensor.
-        mu: float or TensorVariable
+        mu : tensor_like of float
             Mean of the distribution (mu > 0).
-        lam: float or TensorVariable
+        lam : tensor_like of float
             Relative precision (lam > 0).
-        alpha: float or TensorVariable
+        alpha : tensor_like of float
             Shift/location parameter (alpha >= 0).
 
         Returns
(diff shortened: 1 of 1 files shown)


purna135:
Here `no_assert` has no documentation and also not used in function.
Is this ok or we can remove this ?

https://github.com/pymc-devs/pymc/blob/8cb71a5b1104166127e89a25d80c34774e182712/pymc/distributions/continuous.py#L2035-L2037

pymc/distributions/continuous.py
```diff
@@ -2561,10 +2567,11 @@ def logcdf(value, nu):
 
         Parameters
         ----------
-        value: numeric or np.ndarray or `TensorVariable`
+        value : tensor_like of float
             Value(s) for which log CDF is calculated. If the log CDF for
             multiple values are desired the values must be provided in a numpy
             array or `TensorVariable`.
```
purna135: Hi @OriolAbril, should we replace `TensorVariable` in the description with :class:\`aesara.tensor.TensorVariable`

purna135:
I think I did a mistake by commiting all changes at once, I should commit one docstring change at once so that it will be easy to review the code.

pymc/distributions/continuous.py
purna135: I noticed in [pymc.Uniform](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Uniform.html)
the value is mentioned as `numeric or ndarray or TensorVariable` but in this PR I am using `tensor_like of float` which one is right ?

https://github.com/pymc-devs/pymc/blob/7cc570a867fa8a4f8018d91c0aab1cf2aab64258/pymc/distributions/continuous.py#L319-L322

OriolAbril:
> Here no_assert has no documentation and also not used in function. Is this ok or we can remove this ?

 No idea what this no_assert is, and there seem to be others around. My intuition is that this is some leftover from v3 and can/should be removed. However, someone else like @ricardoV94 should confirm this and if that is the case it should be removed in a different PR, like you did for sd. 

For this specific PR, wait until knowing if this should be there or not. If it should be there then it has to be documented. If it should not be there, it doesn't need to be documented but should not be removed yet.

> I think I did a mistake by commiting all changes at once, I should commit one docstring change at once so that it will be easy to review the code.

There is no need to split into commits, at least in my case I always review all commits at the same time, not one by one.

pymc/distributions/continuous.py
OriolAbril: This should be `tensor_like`. I haven't had time to update the guide. While reviewing PRs and while chatting with Ricardo during the 2nd sprint session we realized it would be better to define this `tensor_like` concept in our glossary instead of using numeric, ndarray or TensorVariable. tensor_like is both more informative (it's very easy to say tensor_like of float but otherwise it would be something like float, ndarray of float or TensorVariable of float) and more correct as any type that can be interpreted as an aesara tensor variable is valid.

pymc/distributions/continuous.py
```diff
@@ -848,10 +849,12 @@ class HalfNormal(PositiveContinuous):
 
     Parameters
     ----------
-    sigma: float
+    sigma : tensor_like of float, optional
         Scale parameter :math:`sigma` (``sigma`` > 0) (only required if ``tau`` is not specified).
```
```suggestion
        Scale parameter :math:`\sigma` (``sigma`` > 0) (only required if ``tau`` is not specified).
```

this wasn't rendering correctly, it should match the notation used in the extended description

pymc/distributions/continuous.py
```diff
@@ -848,10 +849,12 @@ class HalfNormal(PositiveContinuous):
...
-    tau: float
+        Defaults to 1.
+    tau : tensor_like of float, optional
         Precision :math:`tau` (tau > 0) (only required if sigma is not specified).
```
```suggestion
        Precision :math:`\tau` (tau > 0) (only required if sigma is not specified).
```

pymc/distributions/continuous.py
```diff
@@ -2023,9 +2028,9 @@ class Pareto(BoundedContinuous):
 
     Parameters
     ----------
-    alpha: float
+    alpha : tensor_like of float, optional
```
OriolAbril: They don't look optional from the code, I don't think calling `Pareto()` works as None should not be interpreted correctly as a float.

pymc/distributions/continuous.py
```diff
@@ -2864,11 +2872,11 @@ class ExGaussian(Continuous):
...
     Parameters
     ----------
-    mu: float
+    mu : tensor_like of float, default 0
         Mean of the normal distribution.
-    sigma: float
+    sigma : tensor_like of float, optional
         Standard deviation of the normal distribution (sigma > 0).
-    nu: float
+    nu : tensor_like of float, optional
```
OriolAbril: These don't look optional either even if being set to None in the call signature

pymc/distributions/continuous.py
```diff
@@ -3018,9 +3027,9 @@ class VonMises(CircularContinuous):
 
     Parameters
     ----------
-    mu: float
+    mu : tensor_like of float, default 0
         Mean.
-    kappa: float
+    kappa : tensor_like of float, optional
```
OriolAbril: same comment here about kappa not being really optional. I think the None in the call signature (also in the other functions) is used so that both options below work:

```
VonMises("name", 2, 1)
VonMises("name", kappa=3)
```
but as you see, kappa is not really optional and needs to be given a value.

ricardoV94:
> Here `no_assert` has no documentation and also not used in function.
> Is this ok or we can remove this ?
> 
> https://github.com/pymc-devs/pymc/blob/8cb71a5b1104166127e89a25d80c34774e182712/pymc/distributions/continuous.py#L2035-L2037

That's related to https://github.com/pymc-devs/pymc/issues/5162

Since nobody objected I think it's time to remove the assert_negative_support and this related no_assert.

As @OriolAbril mentioned that should be done in a separate PR

pymc/distributions/continuous.py
purna135: Thanks @OriolAbril please give me the permission to work on the below methords which are mentioned in https://github.com/pymc-devs/pymc/issues/5459
image

The PRs for these functions are merged but I think that need some corrections. I will do the necessary changes in my upcoming PR for this issue.

OriolAbril: Feel free to make a 2nd review, some will benefit from it because they were merged before we wrote the tensor_like definition for example and don't use it. But I think many (or even most) are already correct (if only the logp or logcdf need changes consider them correct)

purna135:
> As @OriolAbril mentioned that should be done in a separate PR

Sure @ricardoV94

pymc/distributions/continuous.py
purna135: is this for only `alpha`  or for both `alpha` and `m` ?

pymc/distributions/continuous.py
purna135: got it Thanks : )

pymc/distributions/continuous.py
purna135: > Hi @OriolAbril, should we replace `TensorVariable` in the description with :class:`aesara.tensor.TensorVariable`

what about this ?

purna135: > what about this ?

I think we should use "Aesara tensor" since all other functions have logp the description like this only.

```
value : tensor_like of float
            values are desired the values must be provided in a numpy array or Aesara tensor.
```

OriolAbril: Aesara tensor sounds good

pymc/distributions/continuous.py
OriolAbril: From the code I think both, you can test the behaviour. See if `Pareto("name", alpha=1)` works and if `Pareto("name", m=1)` works. If the parameters were both optional, both should work

purna135: ok thanks

OriolAbril (approved):

purna135:
I accidentally deleted this branch.

PR #5596, opened 2022-03-14
Update docstring of SkewNormal, Triangular, Gumbel, Logistic, LogitNormal

purna135:
Related to https://github.com/pymc-devs/pymc/issues/5459

This will update the docsting of

[pymc.SkewNormal](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.SkewNormal.html)
[pymc.Triangular](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Triangular.html)
[pymc.Gumbel](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Gumbel.html)
[pymc.Logistic](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.Logistic.html)
[pymc.LogitNormal](https://docs.pymc.io/en/latest/api/distributions/generated/pymc.LogitNormal.html)

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -3530,6 +3530,7 @@ class Logistic(Continuous):
            \frac{\exp\left(-\frac{x - \mu}{s}\right)}{s \left(1 + \exp\left(-\frac{x - \mu}{s}\right)\right)^2}
 
     .. plot::
+        :context: close-figs
 
         import matplotlib.pyplot as plt
         import numpy as np
@@ -3556,9 +3557,9 @@ class Logistic(Continuous):
 
     Parameters
     ----------
-    mu: float
+    mu : tensor_like of float, default 0
         Mean.
-    s: float
+    s : tensor_like of float, default 1
         Scale (s > 0).
     """
 
@@ -3583,7 +3584,7 @@ class Logistic(Continuous):
 
         Parameters
         ----------
-        value: numeric or np.ndarray or aesara.tensor
+        value : tensor_like of float
             Value(s) for which log CDF is calculated. If the log CDF for multiple
             values are desired the values must be provided in a numpy array or Aesara tensor.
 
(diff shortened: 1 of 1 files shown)


pymc/distributions/continuous.py
```diff
@@ -3242,7 +3246,7 @@ def logcdf(value, lower, c, upper):
 
         Parameters
         ----------
-        value: numeric or np.ndarray or aesara.tensor
+        value : tensor_like of float
             Value(s) for which log CDF is calculated. If the log CDF for multiple
             values are desired the values must be provided in a numpy array or Aesara tensor.
```
purna135, line 3251: Every `logcdf( )` and `logp( )` functions are taking more than one parameter but we have the documentation for only one do we need to add doc for others as well?

OriolAbril: no, changes to logp and logcdf methods should be minimal and without adding new content as they should not be used directly but via the `pymc.logp` and related methods.

pymc/distributions/continuous.py
```diff
@@ -3069,6 +3069,7 @@ class SkewNormal(Continuous):
        2 \Phi((x-\mu)\sqrt{\tau}\alpha) \phi(x,\mu,\tau)
```
OriolAbril, line 3069: There is an extra space above in the "The pdf" sentence which makes it render as a quote instead of normal text: https://pymc--5596.org.readthedocs.build/en/5596/api/distributions/generated/pymc.SkewNormal.html

pymc/distributions/continuous.py
```diff
@@ -3315,9 +3320,9 @@ class Gumbel(Continuous):
 
     Parameters
     ----------
-    mu: float
+    mu : tensor_like of float, optional
         Location parameter.
-    beta: float
+    beta : tensor_like of float, optional
```
OriolAbril: These do not seem optional

OriolAbril (approved):

PR #5608, opened 2022-03-17
Particles representation for Sequential Monte Carlo.

ciguaran:
Implements a class that represents particles and its operations within Sequential Monte Carlo.
Removes a little of code duplication (_tempered_posterior_p)

+ There shouldn't be any breaking changes.
+ This PR happens in the context of extending SMC. Possibly Particles will have a subclass implemented in a follow up PR that holds Points and Array representations in a consistent way.
+ There are new tests for the new class.
+ No need to change notebooks since this is an internal class. But it will help us extend existing samplers.


ricardoV94:
@ciguaran Can you explain the motivation for this refactor?

ciguaran:
@ricardoV94  Although SMC works with an array to represent particles (the old tempered_posterior variable) there is indeed a lot of information moving around to actually operate with it, since, in order to construct a trace, it needs to use a biyection between a flattened array representation and a Point representation (with variable shapes, names and types). 
So the concept of particle is hidden there because a particle is not just an array, it has more semantics than that. Indeed that semantics is closer to the representation used in `step` method of Gradient shared step. This PR is a stone in the bridge between the way SMC operates and the richer representation that is used for other steps, helping the integration of step methods within SMC.

ricardoV94:
Thanks for your reply. I have a lengthy reply below. Just to be clear, my reply is not because I am against such PR or refactor, but just because I don't think there is yet enough context to evaluate it. 

Unless a PR is fixing a bug or doing an obvious improvement, it should usually be preceded with a Github issue or discussion that lays out the motivation for new code changes, and gives time for people to assess, ask question or disagree with it. 

This makes reviews easier and can save wasted work writing code that ends up being rejected. Because you haven't opened such issue, the discussion has to be taken in this PR. 

> @ricardoV94 Although SMC works with an array to represent particles (the old tempered_posterior variable) there is indeed a lot of information moving around to actually operate with it, since, in order to construct a trace, it needs to use a biyection between a flattened array representation and a Point representation (with variable shapes, names and types).

That's a single mapping, but your new class does much more than that. Why are the other functionalities needed? 

Also, can't we just pass the mapping and the array together to whomever needs it? My understanding is that other step methods are usually responsible for creating their own mapping, no?

Isn't the SMC kernel a good enough place to store this meta-information, since it will be coordinating calls to other methods anyway?

> So the concept of particle is hidden there because a particle is not just an array, it has more semantics than that. Indeed that semantics is closer to the representation used in `step` method of Gradient shared step. 

[ Just a note that these other steppers may not always doing the right thing, see for instance: https://github.com/pymc-devs/pymc/issues/4609 ]

> This PR is a stone in the bridge between the way SMC operates and the richer representation that is used for other steps, helping the integration of step methods within SMC.

I assume you mean something like being able to call other step methods like HMC from within SMC? 

In general, it's difficult to judge the merit of a refactor like this in isolation. If it makes future extensions easier, can it be done at the same time as those? We might find out a different type of structure makes more sense when we actually begin making use of it.

pymc/smc/smc.py
```diff
@@ -570,3 +547,86 @@ def _logp_forw(point, out_vars, in_vars, shared):
...
+
+class Particles:
+    """
+    List of draws from variables (particles)
+    belonging to a model that can be modified
+    in place as if they were an array.
+    """
+
+    @classmethod
+    def build(cls, draws: int, initial_point, init_rnd, model, dual_representation=False):
```
aloctavodia: what is dual_representation? I think is not used

pymc/smc/smc.py
```diff
@@ -570,3 +547,86 @@ def _logp_forw(point, out_vars, in_vars, shared):
     f = compile_pymc([inarray0], out_list[0])
     f.trust_input = True
     return f
+
+
+class Particles:
+    """
+    List of draws from variables (particles)
```
aloctavodia: the line limit is 100

pymc/smc/smc.py
```diff
@@ -570,3 +547,86 @@ def _logp_forw(point, out_vars, in_vars, shared):
...
+        if indexes is None:
+            if not len(values) == len(self):
+                raise ValueError(
+                    "If values size doesn't match particle size, then indexes need to be specified"
+                )
+            indexes = [True] * len(self)
+
+        elif not len(indexes) == len(self):
+            raise ValueError(
+                f"Indexes length {len(indexes)} does not match particles size {len(self)}"
```
aloctavodia: Do we need to raise these errors? This is something a user should never get

pymc/smc/smc.py
```diff
@@ -570,3 +547,86 @@ def _logp_forw(point, out_vars, in_vars, shared):
     f = compile_pymc([inarray0], out_list[0])
     f.trust_input = True
     return f
+
+
+class Particles:
+    """
+    List of draws from variables (particles)
+    belonging to a model that can be modified
+    in place as if they were an array.
```
aloctavodia:
```suggestion
    in place as if it were an array.
```

pymc/smc/smc.py
```diff
@@ -570,3 +547,86 @@ def _logp_forw(point, out_vars, in_vars, shared):
...
+        elif not len(indexes) == len(self):
+            raise ValueError(
+                f"Indexes length {len(indexes)} does not match particles size {len(self)}"
+            )
+
+        self.as_array[indexes] = values
+
+    def to_trace(self, chain):
+        """
+        Builds a trace object out self,
```
aloctavodia: not clear to me

```suggestion
        Builds a trace object from itself,
```

aloctavodia (changes requested):
I think this PR helps to organize the code and thus it has value in itself and  is not introducing any overhead on the pymc-devs side. Thus, I am totally fine about accepting it before discussing the HMC kernel.

ricardoV94:
What do your mean overhead on the pymc-devs side?

pymc/smc/smc.py
```diff
@@ -570,3 +547,86 @@ def _logp_forw(point, out_vars, in_vars, shared):
...
+        """
+        Consistently updates array and points
+        representations. If no index is selected, modify all
+        """
+        if indexes is None:
+            if not len(values) == len(self):
+                raise ValueError(
+                    "If values size doesn't match particle size, then indexes need to be specified"
+                )
+            indexes = [True] * len(self)
```
ricardoV94: Why not just slice?  Creating a list seems wasteful

pymc/smc/smc.py
ricardoV94: Why is this not part of the `__init__` method?

pymc/smc/smc.py
```diff
@@ -311,27 +301,16 @@ def _posterior_to_trace(self, chain=0) -> NDArray:
...
-                # Round discrete variable samples. The rounded values were the ones
-                # actually used in the logp evaluations (see logp_forw)
-                var = self.model[varname]
-                if var.dtype in discrete_types:
-                    var_samples = np.round(var_samples).astype(var.dtype)
-                value.append(var_samples.reshape(shape))
-                size += new_size
-            strace.record(point={k: v for k, v in zip(varnames, value)})
-        return strace
+    def _tempered_posterior_p(self, samples):
```
ricardoV94: I am not sure this function is a good abstraction. It's not clear from the name that it will return 3 objects (perhaps add type hints?)

Also the internal names don't make sense out of their old context (e.g., samples -> prop). Probably better `for sample in samples`

Since it is its own function, it seems fine to give lp and pp longer names.

ricardoV94: Perhaps?

```suggestion
    def _tempered_logps(self, samples):
```

pymc/smc/smc.py
```diff
@@ -570,3 +547,86 @@ def _logp_forw(point, out_vars, in_vars, shared):
...
+    def resample(self, resampling_indexes):
+        self.as_array = self.as_array[resampling_indexes]
+
+    def __len__(self):
+        return self._draws
+
+    def __getitem__(self, item):
+        return self.as_array[item]
+
+    def set(self, indexes, values):
```
ricardoV94: Why not use `__setittem__`?

Also can we just rely on NumPy to take care of the shape checks?

pymc/smc/smc.py
ciguaran: Removed the index construction altogether

pymc/smc/smc.py
ciguaran: Removing this error handling, it was needed in another context, before I extracted this as a separate PR, we can revisit if we need them again

pymc/smc/smc.py
ciguaran: This was github's yesterday's outage that didn't show my latest changes, it was removed.

ciguaran: Because by using a factory we are decoupling details of a very particular way of building an object, from the object itself. For example, Particles shouldn't need an initial point to be built, it's just one way of doing it, by splitting this two things we can build Particles with the dependencies it needs, and no circunstancial objects.

ricardoV94: I don't get it yet. Do you have a case in mind where you will make use of this split?

pymc/smc/smc.py
```diff
@@ -570,3 +546,69 @@ def _logp_forw(point, out_vars, in_vars, shared):
...
+    def __len__(self):
+        return self._draws
+
+    def __getitem__(self, item):
+        return self.as_array[item]
+
+    def __setitem__(self, key, value):
+        """
+        Consistently updates array and points representations.
+        """
```
ricardoV94, line 587:
```suggestion
```
Doesn't seem necessary anymore

pymc/smc/smc.py
```diff
@@ -570,3 +546,69 @@ def _logp_forw(point, out_vars, in_vars, shared):
...
+        self.as_array[key] = value
+
+    def to_trace(self, chain):
+        """
+        Builds a trace object from self,
+        rounding discrete variable samples.
```
ricardoV94, line 593:
```suggestion
        Builds a trace object from self, rounding discrete variable samples.
```

twiecki:
@ciguaran Did you mean to close this one?

ciguaran:
Hi yes, I did it on purpose!

twiecki:
@ciguaran Why? :)

Issue #5575, opened 2022-03-10
Flaky test `test_mixture.py::TestMixture::test_list_univariate_components_deterministic_weights`


michaelosthege:
I think this is new, probably introduced by 620b11d3bf86bcd9121eed2d57016993b5cbe044.

Saw it happen in https://github.com/pymc-devs/pymc/runs/5497901632?check_suite_focus=true#step:7:903 (belongs to #5572) and https://github.com/pymc-devs/pymc/runs/5488262408?check_suite_focus=true#step:7:990 (beloging to #5571).

cc @ricardoV94

ricardoV94:
I guess with float32 and the low sigma the probability of identical draws is not that small. I am getting fed up with these float32 tests by the day xD

https://github.com/pymc-devs/pymc/blob/c3c76c5024aab209524788eeef16e1bfe2c26105/pymc/tests/test_mixture.py#L192-L193

ricardoV94:
It should be sufficient to increase the sigma slightly. This issue and related PR by @LukeLB are a good example of how to make sure flaky tests are fixed properly: https://github.com/pymc-devs/pymc/issues/5409

ricardoV94:
We are no longer testing in float32 so this should not be a problem anymore (I think)

PR #5615, opened 2022-03-18
Do not set RNG updates inplace in compile_pymc

ricardoV94:
Just a small maintenance change. No point in doing changes inplace if we don't have to


michaelosthege (approved):
That one failing test looks flaky.

ricardoV94:
Yeah it's the one in #5575

PR #5617, opened 2022-03-18
[DO NOT MERGE] Test that ci triggers fix works also on prs from forks

OriolAbril:


michaelosthege:
I see you're still experimenting?

If you're right, I'll have to fix this in at least 3 other repos 😓

It would be nice if we didn't need to duplicate the `paths:` filter.

OriolAbril:
do you want me try another change in another commit? (I left for a bit to cook)

I think it is clear no github actions are triggered here. All actions should have run for the last commit, and pre-commit should have run for the second to last one.

michaelosthege:
Can you try something like this?
```
on: [push, pull_request]
  paths:
```
If this was XML the IDE could tell us, but with YAML I have no idea if the above is gonna work.

OriolAbril:
I forgot the --no-verify and we have a yaml check that failed with these changes so it probably won't work, but pushing anyway to check

OriolAbril:
github also complains: https://github.com/pymc-devs/pymc/actions/runs/2006002679

michaelosthege:
Sounds like we should duplicate the paths filter then.
I'll do that in the other branch tomorrow.

PR #5619, opened 2022-03-19
Refine trigger filter for both PRs and pushes

michaelosthege:
Previously only one of the trigger conditions had the filter.
This is a follow-up to #5364.

Unfortunately, the changes can't be tested on this PR because of how GitHub determines the diff:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#git-diff-comparisons

### DIFF
--- a/.github/workflows/arviz_compat.yml
+++ b/.github/workflows/arviz_compat.yml
@@ -2,16 +2,22 @@ name: arviz-compatibility
 
 on:
   pull_request:
+    paths:
+      - ".github/workflows/*.yml"
+      - "pymc/**.py"
+      - "*.py"
+      - "conda-envs/**"
+      - "codecov.yml"
+      - "scripts/*.sh"
   push:
     branches: [main]
     paths:
-      - ".github/workflows/*"
-      - "pymc/**"
-      - "setup.py"
-      - "pyproject.toml"
-      - "buildosx"
+      - ".github/workflows/*.yml"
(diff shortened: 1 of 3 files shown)


.github/workflows/arviz_compat.yml
```diff
@@ -2,16 +2,22 @@ name: arviz-compatibility
...
+      - "codecov.yml"
+      - "scripts/*.sh"
   push:
     branches: [main]
     paths:
-      - ".github/workflows/*"
-      - "pymc/**"
-      - "setup.py"
-      - "pyproject.toml"
-      - "buildosx"
```
michaelosthege, line 12: The `pyproject.toml` is only relevant for the tools running in pre-commit.

The `buildosx` does not seem to have a purpose. It's 8 years old, was never changed and is not mentioned anywhere in the codebase.

.github/workflows/arviz_compat.yml
```diff
@@ -2,16 +2,22 @@ name: arviz-compatibility
...
+      - ".github/workflows/*.yml"
+      - "pymc/**.py"
+      - "*.py"
       - "conda-envs/**"
       - "codecov.yml"
+      - "scripts/*.sh"
```
michaelosthege, line 20: These scripts are involved in creating the test environments.

OriolAbril (approved):
looks good, thanks!

PR #5622, opened 2022-03-19
Bring back sampler argument target_accept 

### LINKED ISSUE #4291: Implicit behavior in pm.sample() when init=None
When `sample()` is initiated with `init=None`, pymc3 does not try to call the general `init_nuts()` function, and instead silently defaults to the elementwise auto-assignment with `assign_step_methods()`. 

This behavior however is not explicitly mentioned in the documentation, where it is suggested that the init argument should always be a string with one of the listed options. Should we: 

1) enforce this constraint, similarly to what is already done in `init_nuts()`, by raising TypeError and ValueError when an invalid init argument is passed?
2) document and allow this behavior,  either by adding None as an explicit option in the documentation or adding a new string argument `init="elementwise"` that would work as None currently does.
3) do nothing...

I would be inclined to 1). If users want to avoid the default auto-assignment of the Nuts sampler, the documentation is quite clear that they should pass the desired step methods via the argument `step`.

### LINKED ISSUE #5620: Sampler argument target_accept not working in 4.0.0b4
```python
with pm.Model() as model:
    norm = pm.Normal("norm")
    pm.sample(target_accept=0.9, tune=2000)
```

yields

```python
Input In [2], in <module>
      1 with pm.Model() as model:
      2     norm = pm.Normal("norm")
----> 3     pm.sample(target_accept=0.9, tune=2000)

File ~/opt/anaconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/pymc/sampling.py:470, in sample(draws, step, init, n_init, initvals, trace, chain_idx, chains, cores, tune, progressbar, model, random_seed, discard_tuned_samples, compute_convergence_checks, callback, jitter_max_retries, return_inferencedata, idata_kwargs, mp_ctx, **kwargs)
    467 draws += tune
    469 initial_points = None
--> 470 step = assign_step_methods(model, step, methods=pm.STEP_METHODS, step_kwargs=kwargs)
    472 if isinstance(step, list):
    473     step = CompoundStep(step)

File ~/opt/anaconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/pymc/sampling.py:219, in assign_step_methods(model, step, methods, step_kwargs)
    211         selected = max(
    212             methods,
    213             key=lambda method, var=rv_var, has_gradient=has_gradient: method._competence(
    214                 var, has_gradient
    215             ),
    216         )
    217         selected_steps[selected].append(var)
--> 219 return instantiate_steppers(model, steps, selected_steps, step_kwargs)

File ~/opt/anaconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/pymc/sampling.py:142, in instantiate_steppers(model, steps, selected_steps, step_kwargs)
    140 unused_args = set(step_kwargs).difference(used_keys)
    141 if unused_args:
--> 142     raise ValueError("Unused step method arguments: %s" % unused_args)
    144 if len(steps) == 1:
    145     return steps[0]

ValueError: Unused step method arguments: {'target_accept'}
```

with PyMC version 4.0.0b4, aesara version 2.5.1 and aeppl version 0.0.27. However, there is no error when using PyMC version 4.0.0b3 (and aesara version 2.4.0, aeppl version 0.0.26). See notebooks [here (b3)](https://github.com/larryshamalama/pymc3-playground/blob/master/notebooks/shortcomings/target-accept-vb3.ipynb) and [here (b4)](https://github.com/larryshamalama/pymc3-playground/blob/master/notebooks/shortcomings/target-accept-vb4.ipynb)

First raised by @OriolAbril

aloctavodia:
fixes #5620

Edit:
This also fixes #4291 by enforcing `init` to be a string

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -279,18 +279,17 @@ def sample(
         The number of samples to draw. Defaults to 1000. The number of tuned samples are discarded
         by default. See ``discard_tuned_samples``.
     init : str
-        Initialization method to use for auto-assigned NUTS samplers.
-        See `pm.init_nuts` for a list of all options.
+        Initialization method to use for auto-assigned NUTS samplers. See `pm.init_nuts` for a list
+        of all options. This argument is ignored when manually passing the NUTS step method.
     step : function or iterable of functions
         A step function or collection of functions. If there are variables without step methods,
-        step methods for those variables will be assigned automatically.  By default the NUTS step
-        method will be used, if appropriate to the model; this is a good default for beginning
-        users.
+        step methods for those variables will be assigned automatically. By default the NUTS step
+        method will be used, if appropriate to the model.
     n_init : int
         Number of iterations of initializer. Only works for 'ADVI' init methods.
     initvals : optional, dict, array of dict
-        Dict or list of dicts with initial value strategies to use instead of the defaults from `Model.initial_values`.
-        The keys should be names of transformed random variables.
+        Dict or list of dicts with initial value strategies to use instead of the defaults from
+        `Model.initial_values`. The keys should be names of transformed random variables.
         Initialization methods for NUTS (see ``init`` keyword) can overwrite the default.
     trace : backend or list
         This should be a backend instance, or a list of variables to track.
@@ -317,8 +316,8 @@ def sample(
     model : Model (optional if in ``with`` context)
         Model to sample from. The model needs to have free random variables.
     random_seed : int or list of ints
-        Random seed(s) used by the sampling steps.  A list is accepted if
-        ``cores`` is greater than one.
+        Random seed(s) used by the sampling steps. A list is accepted if ``cores`` is greater than
+        one.
     discard_tuned_samples : bool
         Whether to discard posterior samples of the tune interval.
     compute_convergence_checks : bool, default=True
@@ -330,17 +329,17 @@ def sample(
         is drawn from.
         Sampling can be interrupted by throwing a ``KeyboardInterrupt`` in the callback.
     jitter_max_retries : int
-        Maximum number of repeated attempts (per chain) at creating an initial matrix with uniform jitter
-        that yields a finite probability. This applies to ``jitter+adapt_diag`` and ``jitter+adapt_full``
-        init methods.
+        Maximum number of repeated attempts (per chain) at creating an initial matrix with uniform
+        jitter that yields a finite probability. This applies to ``jitter+adapt_diag`` and
+        ``jitter+adapt_full`` init methods.
     return_inferencedata : bool
-        Whether to return the trace as an :class:`arviz:arviz.InferenceData` (True) object or a `MultiTrace` (False)
-        Defaults to `True`.
+        Whether to return the trace as an :class:`arviz:arviz.InferenceData` (True) object or a
+        `MultiTrace` (False). Defaults to `True`.
     idata_kwargs : dict, optional
         Keyword arguments for :func:`pymc.to_inference_data`
     mp_ctx : multiprocessing.context.BaseContent
-        A multiprocessing context for parallel sampling. See multiprocessing
-        documentation for details.
+        A multiprocessing context for parallel sampling.
+        See multiprocessing documentation for details.
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


pymc/sampling.py
```diff
@@ -471,6 +464,20 @@ def sample(
 
     if isinstance(step, list):
         step = CompoundStep(step)
+    elif isinstance(step, NUTS):
```
ricardoV94: What happens with compound steps that include NUTS? Aren't we initializing things the same way then?

ricardoV94:
Can we add a regression test. Perhaps a short sampling and the  checking the sample_stats target_accept?

pymc/sampling.py
aloctavodia: We never did, this is related to my comment on slack. I can try to address that, but I preferred to first make this change that revert us to what we "always" did  (plus being able to have external samplers like BART) so we can make a quick release.

aloctavodia:
We have a test with the argument `init=None` here https://github.com/pymc-devs/pymc/blob/c3ffec89f058854e727e6ffc24379eee3154b623/pymc/tests/test_quadpotential.py#L153 but we have a condition here https://github.com/pymc-devs/pymc/blob/849a64d7c580e201ea027ac473fa906d9dd9b682/pymc/sampling.py#L2262-L2263 restricting init to be a string (which agrees with the docstrings)

ricardoV94:
I had opened an issue about that long time ago: https://github.com/pymc-devs/pymc/issues/4291

I would say we raise if it's not a string

pymc/sampling.py
```diff
@@ -352,37 +352,28 @@ def sample(
...
-       2. ``transit_p`` to BinaryGibbsMetropolis: binary_gibbs_metropolis={'transit_p':.7}
-
-       Note that available names are:
-
-        ``nuts``, ``hmc``, ``metropolis``, ``binary_metropolis``,
-        ``binary_gibbs_metropolis``, ``categorical_gibbs_metropolis``,
-        ``DEMetropolis``, ``DEMetropolisZ``, ``slice``
-
-    B. If you manually declare the ``step_method``\ s, within the ``step``
+    Alternativelly, if you manually declare the ``step_method``\ s, within the ``step``
```
michaelosthege:
```suggestion
    Alternatively, if you manually declare the ``step_method``\ s, within the ``step``
```

michaelosthege (changes requested):
I would recommend to bring back this entire section: https://github.com/pymc-devs/pymc/commit/2dd4c8c5401121b3e48fffd558838fe2d547a4c2#diff-7b3470589153c7c90a2c05fc86402f9acf65555200d90f01b51a47ccef592628L484-L506

The reason is mostly that before attempting to do a `at.grad` there's no way to know if the model is differentiable. The old `try/catch` logic took care of that.

pymc/tests/test_sampling.py
```diff
@@ -103,8 +103,8 @@ def test_sample(self):
     def test_sample_init(self):
         with self.model:
             for init in (
-                "advi",
-                "advi_map",
+                # "advi",
```
ricardoV94: Why are these commented out?

aloctavodia: they are still not working in V4. This used to pass because init_nuts was removed and before that because of the try-except.

twiecki: That's a bug then as VI should work again.

aloctavodia: ohh you are right it seems the bug was introduced here https://github.com/pymc-devs/pymc/issues/4819 and is failing because we are doing https://github.com/pymc-devs/pymc/blob/0985bd47092107a004cc3bd7d93a210669b037fe/pymc/sampling.py#L2353 I will fix it after lunch :-)

pymc/sampling.py
```diff
@@ -466,11 +458,37 @@ def sample(
 
     draws += tune
 
+    nuts_init = True
```
ricardoV94: Perhaps?
```suggestion
    auto_nuts_init = True
```

ricardoV94 (commented):
We should merge this soon, but is there a test that ensures we are doing the automatic init?

ricardoV94 (approved):

ricardoV94:
We should release immediately after this gets in. @aloctavodia Did you confirm we don't do any useful NUTS init when we get a default CompoundStep? If so, we should open an issue for that

aloctavodia:
Agreed, and yes we should open an issue.

ricardoV94:
> Agreed, and yes we should open an issue.

Do you mind doing that, as you have more context atm?

aloctavodia:
Yes, sure!

pymc/sampling.py
```diff
@@ -466,11 +458,37 @@ def sample(
...
+            chains=chains,
+            n_init=n_init,
+            model=model,
+            seeds=random_seed,
+            progressbar=progressbar,
+            jitter_max_retries=jitter_max_retries,
+            tune=tune,
+            initvals=initvals,
+            **kwargs,
+        )
```
michaelosthege, line 491: I'm not confident about this. As I commented before, the `try/catch` block caught not-implemented gradients, __which we're probably not testing__.

Also I'm sceptical about the pattern of having `isinstance(step, NUTS)` and `if "nuts" in kwargs`. It reads like a hack to account for some specific cases, and it's not as general as the previous implementation.

You're the ones who've been looking into this much more than me, so feel free to overrule this.

ricardoV94: The `assign_step_methods` checks for gradients as well. If that does not do a good enough job of testing if NUTS is appropriate then it should be treated as a bug and be fixed there

michaelosthege: Okay, you're right. Also we're testing it: https://app.codecov.io/gh/pymc-devs/pymc/blob/18346acc797c32ccfa2d61ebc8f368ef03bc216c/pymc/sampling.py

michaelosthege (approved):

Issue #5041, opened 2021-10-05
Adjust pre-commit step in CI to run every time, but fast fail if red


michaelosthege:
In a glossary PR I merged yesterday the docs step ran, but the pre-commit step did not.
We can easily run the pre-commit also for PRs of new contributors.

Also we should probably fast-fail the CI if the pre-commit is red.

michaelosthege:
Looks like this is really hard when pre-commit and tests are separate GitHub Actions workflows: https://github.com/pre-commit-ci/issues/issues/106

ricardoV94:
Could we have the pre-commit run in each job? Would that make things easier?

michaelosthege:
It looks like we're getting a switch to `mamba` with #5387, but to solve this issue, I'd prefer to pursue this:

> [...] combine pre-commit.yml, arviz_compat.yml and pytest.yml into the same workflow.
> Then we could fail-fast all jobs when the pre-commit doesn't pass (#5041)

quoted [my comment in #5375](https://github.com/pymc-devs/pymc/issues/5375#issuecomment-1017650388)

That'll be less steps in total and cancel/rerun would be a single click.

michaelosthege:
I'm not sure how to interpret the docs about the `fail-fast` setting. (Asked a question in https://github.com/github/docs/issues/16401)

The trigger filters are another hurdle: They apply to the workflow but we want to run pre-commit every time. So if we include a pre-commit job in the `tests.yml` workflow, we'll have to keep the other pre-commit workflow around too.

All in all not elegant either way.

michaelosthege:
Update: the `fail-fast` only applies to the same matrix. So we'd have to restructure everything into the same matrix, or make all jobs depend on the `pre-commit` job (delays all jobs by a few minutes).

ricardoV94:
Second option doesn't sound so bad

PR #5623, opened 2022-03-19
Combine test workflows into one

### LINKED ISSUE #5475: Combine windows workflows
We have 4 Windows workflows with a combined test runtime of \~10 minutes, but we are spending about 4x5~20 minutes just building the conda environments

If we can combine the tests in 1/2 workflows we could save some compute time without prolonging the total runtimes

michaelosthege:
This PR combines `pytest`, `arviz-compat` and `jaxtests` workflows into one (big) `tests` workflow.

# Motivations
* Cancelling all workflows is now 3x less clicking. (Only core devs have the necesary permissions to cancel workflows.)
* GitHub Actions workflows can't depend on each other, but if we want to fail fast on ❌ `pre-commit` (#5041), having all jobs in one workflows will make it a lot easier.
* Tracking workflow runtime via https://share.streamlit.io/koaning/gitlit/main will be easier after they're all in one.

# Changes
* ✔ Removed the `buildosx` build script. Chris confirmed via Slack that it can be removed.
* ✔ `arviz-compat` and `jaxtests` were deleted - their jobs were moved into `tests` (previously `pytest`.
* ✔ Upgraded the tests from `ubuntu 18.04` to `20.04` which [is the most recent LTS release](https://wiki.ubuntu.com/Releases).
* ✔ Redistributed tests between jobs to reduce the runtime (closes #5475)


.github/workflows/tests.yml
```diff
@@ -249,3 +248,148 @@ jobs:
...
+      - name: Run tests
+        run: |
+          python -m pytest -vv --cov=pymc --cov-report=xml --cov-report term --durations=50 $TEST_SUBSET
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v2
+        with:
+          env_vars: TEST_SUBSET
+          name: JAX tests - ${{ matrix.os }} ${{ matrix.floatx }}
+          fail_ci_if_error: false
+  arviz:
```
ricardoV94: Can we give this a better name? Perhaps "latest" or "latest_dependency"? It's not that much focused on arviz (at least not anymore)

ricardoV94: Or perhaps just get rid of it and just have a windows-macos job?

.github/workflows/tests.yml
```diff
@@ -33,7 +32,7 @@ jobs:
   ubuntu:
     strategy:
       matrix:
-        os: [ubuntu-18.04]
+        os: [ubuntu-latest]
```
ricardoV94: Could also be LTS supported version? Right now this is 20.4 and in April will become 22.4 until 2024

michaelosthege: In principle yes, but we use `-latest` for the other two OSes.

...okay, so Aesara uses `ubuntu-latest` then I'll change this to `ubuntu-20.04` so we get higher coverage overall

ricardoV94:
Looking at the runtime there are 4 or 5 workflows that took ~6 minutes to run, one of which only spent 45 seconds running actual tests, and the rest was setting up the environment

https://github.com/pymc-devs/pymc/runs/5612971573?check_suite_focus=true

I would merge those short workflows with some of the others that take like 11 minutes. We still have some workflows that take ~30 minutes, so we will not loose anything by merging the fast ones.

michaelosthege:
I aimed for 15-20 minutes per job. Only the `test_distribution.py` still takes much longer. But I don't want to split it up in this PR.

I was about to comment final success, but then I noticed that the Windows jobs were only running the first item in each subset 😮 And that was already the case on `main` 😲
I home my latest push fixes it, but in any case I'll probably have to do more balancing.

Some stats compared to 849a64d7c580e201ea027ac473fa906d9dd9b682:
* Number of test jobs reduced from 23 to 22
* Total runtime reduced from __6h38m__ in  to __6h22m__

ricardoV94 (approved):
LGTM,

This also explains why the windows xfails were not failing with strict XPASS in the past

ricardoV94:
Failing test seems to be https://github.com/pymc-devs/pymc/issues/5575

I'll try to fix it this week. Shouldn't be a blocker here

michaelosthege:
I noticed that the `test_distribution_moments.py` takes 12 minutes, which is much longer than I would have expected.

My last force-push moved mixture tests between the ubuntu groups to shave the longest job runtime down.

I'm going to rebase-merge since there's no need to run the CI again.

Issue #5539, opened 2022-03-03
Add a private API page to the documentation


OriolAbril:
We can set up an private API page inside the contributing section. If you can comment a few objects that should be there I can get it started so adding more is adding the objects to the list and follow the pattern.

cc @ricardoV94

soma2000-lang:
@OriolAbril  I want to work on this issue 
Can you how to start

OriolAbril:
I can set this up later today if I have time, after that you would be able to add all the relevant objects to that section. However, as I mentioned in the issue description, I don't know what should be listed there and there is little to be done without knowing this.

@ricardoV94 maybe we could start by moving some functions/classes from the public api docs to the private ones? Are all objects listed in https://docs.pymc.io/en/latest/api/aesaraf.html user facing? And https://docs.pymc.io/en/latest/api/shape_utils.html?

Issue #5346, opened 2022-01-13
Developer guide needs to be updated for V4


ricardoV94:
First step is figuring out what is no longer correct. Re-running the code snippets will likely reveal most of the things that are working differently

twiecki:
Hiding it with https://github.com/pymc-devs/pymc/pull/5850 for now until it's updated. It's not critical for the release.

PR #5625, opened 2022-03-19
start private API section

OriolAbril:
Starts a private api section and fills an example page.
Still not sure what should be documented there and the goal is not to add more things,
only to show how it is done so other people can take over.
Related to #5539

+ [x] what are the (breaking) changes that this PR makes? **None**
+ [x] important background, or details about the implementation: see above

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ tags
 # Sphinx
 _build
 docs/jupyter_execute
+docs/.jupyter_cache
 docs/**/generated/*
 
 # Merge tool
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,9 @@ help:
 clean:
 	rm -rf $(BUILDDIR)/*
 	rm -rf $(SOURCEDIR)/api/**/generated
+	rm -rf $(SOURCEDIR)/api/**/classmethods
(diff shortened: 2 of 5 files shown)


michaelosthege:
Maybe move `ARCHITECTURE.md` from the root level into the new section?

Also, linking this because it's related: https://github.com/pymc-devs/pymc/wiki/PyMC3-Internals-(for-devs)

OriolAbril:
> Maybe move ARCHITECTURE.md from the root level into the new section?

I'll always agree to moving files from random markdown files in github to the documentation. I see both more related to the developer guide (and therefore to #5346) than to this page that would be only API docs so that we can reference private functions in the developer guide, in the architecture... as links and get to their api page with nice docstring, the [source] section. The hope would be that it would help both developers and new contributors.

michaelosthege (approved):

michaelosthege:
@OriolAbril shall we merge this one before we start getting git conflicts?

Only maintainers could push to this branch. So having this on `main` will make it easier for everybody to contribute.

OriolAbril:
Sounds good, I didn't merge because I was focused on other things, merging now

PR #5627, opened 2022-03-19
preserve chain and draw dims in posterior predictive

OriolAbril:
+ [x] what are the (breaking) changes that this PR makes? None that I can think of
  - This is mostly relevant when thinning the posterior to generate posterior predictive samples
    for only a subset of it (if pp sampling is expensive for example), but can also
    be an issue if someone uses non standard coordinate values.
  - It depends on https://github.com/arviz-devs/arviz/pull/2001 and an ArviZ release before the
    specific test can pass. Should I use a skipif to request arviz>0.11.4 so it can be merged
    before that? (not sure when next ArviZ release will be)
+ [x] [consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples): Updating the golf putting notebook that samples posterior predictive on a thinned posterior only.

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -1627,13 +1627,13 @@ def sample_posterior_predictive(
     if "coords" not in idata_kwargs:
         idata_kwargs["coords"] = {}
     if isinstance(trace, InferenceData):
-        idata_kwargs["coords"].setdefault("draw", trace.posterior.draw)
-        idata_kwargs["coords"].setdefault("chain", trace.posterior.chain)
+        idata_kwargs["coords"].setdefault("draw", trace["posterior"]["draw"])
+        idata_kwargs["coords"].setdefault("chain", trace["posterior"]["chain"])
         _trace = dataset_to_point_list(trace["posterior"])
         nchain, len_trace = chains_and_samples(trace)
     elif isinstance(trace, xarray.Dataset):
-        idata_kwargs["coords"].setdefault("draw", trace.draw)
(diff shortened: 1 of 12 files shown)


OriolAbril:
This should be ready to merge now that arviz 0.12 is out. The specific tests that need arviz>=0.12 are passing like all other tests and only pre-commit is failing due to the black issue mentioned in slack.

OriolAbril:
@michaelosthege I rebased after merging the fix ci PR but mypy seems to be broken again

pymc/sampling.py
```diff
@@ -1620,10 +1620,20 @@ def sample_posterior_predictive(
...
     nchain: int
+    if idata_kwargs is None:
+        idata_kwargs = {}
+    else:
+        idata_kwargs = idata_kwargs.copy()
+    if "coords" not in idata_kwargs:
+        idata_kwargs["coords"] = {}
     if isinstance(trace, InferenceData):
+        idata_kwargs["coords"].setdefault("draw", trace.posterior.draw)
+        idata_kwargs["coords"].setdefault("chain", trace.posterior.chain)
```
michaelosthege: Mypy isn't broken - you just introduced two lines that according to the type information from ArviZ don't work.

Replace `trace.posterior` with `trace["posterior"]` in the two lines above

OriolAbril: Oohh, sorry. Now I see what Ravin meant in slack. These two lines prompted over 350 lines of mypy warnings and errors that start with aesaraf related things

michaelosthege (approved):

PR #5628, opened 2022-03-20
Make StudentT/AsymmetricLaplace .dist() signatures consistent

### LINKED ISSUE #5602: Reorder sigma and lam parameter in StudentT distribution
It's a bit surprising that lam appears before sigma, when defining a StudentT with positional arguments.

https://github.com/pymc-devs/pymc/blob/09bddde70b33a54cd8aa494bb15d8d11462805ec/pymc/distributions/continuous.py#L1917

The docstrings suggest sigma comes before, and this is actually the case for the HalfStudentT. In the Normal, sigma also shows up before tau. I think we should reverse them so that sigma takes precedence here as well.

This would warrant a user warning about the changed behavior, which would need a dummy new variable like `_sigma=None` so that we can detect a positional only argument was used.

This emerged recently in discourse: https://discourse.pymc.io/t/pymc3-produces-different-results-than-stan-numpyro/9030/5

cluhmann:
This PR closes https://github.com/pymc-devs/pymc/issues/5602.  Breaking changes:

- StudentT dist signature changed to match Normal, HalfStudentT, and others
- StudentT now requires either `sigma` or `lam` as  kwarg to alert uses to change
- ~StudentT's `nu` argument now defaults to 1.0 (now matches HalfStudentT)~
- HalfStudentT's `nu` argument now has no default (now matches StudentT)
- AsymmetricLaplace dist signature changed to match Laplace, SkewNormal, and StudentT
- AsymmetricLaplace's `mu` argument no longer has a default value (now matches Laplace)


pymc/distributions/continuous.py
```diff
@@ -1901,7 +1901,7 @@ class StudentT(Continuous):
     rv_op = studentt
 
     @classmethod
-    def dist(cls, nu, mu=0, lam=None, sigma=None, *args, **kwargs):
+    def dist(cls, nu=1, mu=0, sigma=None, lam=None, *args, **kwargs):
```
ricardoV94: I would instead remove the default nu from the HalfStudentT, I don't see a good reason to provide a default.

ricardoV94: I think this distribution is critical enough that we need to do this sigma lam change non silently. The easiest way I can think of is to temporarily force it to be a keyword argument, and later on we can drop that requirement again:

```suggestion
    def dist(cls, nu, mu=0, *, sigma=None, lam=None, **kwargs):
```

cluhmann: Makes sense.  Incidentally, I completely missed the fact that sigma/lam have default values of 1.0 because those defaults are hidden in the `get_tau_sigma()`.  I had assumed that function resolved the ambiguity in the 2 specifications, but didn't realize that it also supplied default values.

cluhmann: I think this bumps into @michaelosthege's [comment](https://github.com/pymc-devs/pymc/issues/5602#issuecomment-1071238984) about why we are accepting positional arguments at all.  Given that most of the tests are passing (which surprised me), it doesn't seem like we are expecting users to rely on this behavior.

michaelosthege: Yeah, apart from small "let me just demo this" code snippets, I have not seen non-explicit distr parametrization in the wild.
I don't think we'd break a lot of user code by requiring kwargs. But we'd certainly create a lot of damage by silently breaking positional parametrization.

cluhmann: > I have not seen non-explicit distr parametrization in the wild.

I have seen them in code posted to discourse before.

ricardoV94:
Will need release-notes as well

pymc/distributions/continuous.py
```diff
@@ -1662,7 +1662,7 @@ class AsymmetricLaplace(Continuous):
     rv_op = asymmetriclaplace
 
     @classmethod
-    def dist(cls, b, kappa, mu=0, *args, **kwargs):
+    def dist(cls, kappa, mu, b, *args, **kwargs):
```
ricardoV94, line 1663: Looking at the Laplace, should it be this? 
```suggestion
    def dist(cls, mu, b kappa, *args, **kwargs):
```

Also need to reorder the parameters in the docstrings to reflect the new order

cluhmann: As I said in the original issue, SkewNormal and StudentT both have their non-normality/non-symmetry arguments first (before location and scale), so it was unclear why `kappa` would be different.

ricardoV94: Sounds good. Should still reorder the docstring parameters

RELEASE-NOTES.md
```diff
@@ -27,6 +27,10 @@ Also check out the [milestones](https://github.com/pymc-devs/pymc/milestones) fo
 
 All of the above apply to:
 
+- `pm.StudentT` now now requires either `sigma` or `lam` as kwarg
+- `pm.StudentT` now requires `nu` to be specified (no longer defaults to 1.0)
+- `pm.AsymmetricLaplace` positional arguments re-ordered
+- `pm.AsymmetricLaplace` now requires `mu` to be specified (no longer defaults to 1.0)
```
```suggestion
- `pm.AsymmetricLaplace` now requires `mu` to be specified (no longer defaults to 0.0)
```

cluhmann:
Should the signatures of `get_moment()` and `logp()` match the signature of `dist()`?  I had assumed that the former were used mostly internally, but looking again they seem to be reasonably public.  Should have asked earlier.

ricardoV94:
> Should the signatures of `get_moment()` and `logp()` match the signature of `dist()`? I had assumed that the former were used mostly internally, but looking again they seem to be reasonably public. Should have asked earlier.

No, those should be considered internal, and the order is fixed by the RandomVariable

pymc/distributions/continuous.py
cluhmann: Just checked and this matches scipy's [asymmetric Laplace signature](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.laplace_asymmetric.html).  The [Half]StudentT signature (now) matches [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.t.html) as well (no default value for `nu`).

ricardoV94:
One HalfStudentT test seems to be failing. It probably relied on old parameter order or default, and has to be updated: https://github.com/pymc-devs/pymc/runs/5632254063?check_suite_focus=true

RELEASE-NOTES.md
```diff
@@ -27,6 +27,10 @@ Also check out the [milestones](https://github.com/pymc-devs/pymc/milestones) fo
 
 All of the above apply to:
 
+- `pm.StudentT` now now requires either `sigma` or `lam` as kwarg
```
ricardoV94: We usually mention the PR with a link for easier book keeping.

pymc/distributions/continuous.py
OriolAbril: Hence the need to document the defaults in the docstring!

OriolAbril: somewhat related, there are also several cases where `param=None` doesn't mean that parameter is optional. I think this only makes sense when combining positional and keywork args, for example things like this: https://github.com/pymc-devs/pymc/pull/5595#discussion_r826996478

OriolAbril:
> No, those should be considered internal, and the order is fixed by the RandomVariable

And these methods should be removed from docs. The only reason they are not is the method listing in classes is automatic. With @ricardoV94 we tried a template to list only (but always) the dist method, but some distributions are actually functions/wrappers that don't have a dist and we weren't able to figure out conditionals in autodoc templates, not sure if we have an issue for that yet though

pymc/distributions/continuous.py
```diff
@@ -1644,12 +1644,12 @@ class AsymmetricLaplace(Continuous):
...
-    kappa : tensor_like of float
-        Symmetry parameter (kappa > 0).
-    mu : tensor_like of float, default 0
-        Location parameter.
+    kappa: float
+        Symmetry parameter (kappa > 0)
+    mu: float
+        Location parameter
+    b: float
+        Scale parameter (b > 0)
```
OriolAbril: There have been some issues with the merge, type should still indicate `tensor_like`, the rest seems to have all updated content.

cluhmann: Yeah, I managed to mess that up.

pymc/distributions/continuous.py
```diff
@@ -1644,12 +1644,12 @@ class AsymmetricLaplace(Continuous):
 
     Parameters
     ----------
-    b : tensor_like of float
-        Scale parameter (b > 0).
     kappa : tensor_like of float
         Symmetry parameter (kappa > 0).
     mu : tensor_like of float, default 0
```
```suggestion
    mu : tensor_like of float
```

mu no longer defaults to 0

RELEASE-NOTES.md
```diff
@@ -26,6 +26,10 @@ Also check out the [milestones](https://github.com/pymc-devs/pymc/milestones) fo
 
 All of the above apply to:
 
+- `pm.StudentT` now now requires either `sigma` or `lam` as kwarg [#5628](https://github.com/pymc-devs/pymc/pull/5628)
+- `pm.StudentT` now requires `nu` to be specified (no longer defaults to 1)
+- `pm.AsymmetricLaplace` positional arguments re-ordered
+- `pm.AsymmetricLaplace` now requires `mu` to be specified (no longer defaults to 0)
```
```suggestion
Signature and default parameters changed for several distributions (see [#5628](https://github.com/pymc-devs/pymc/pull/5628)):
    - `pm.StudentT` now requires either `sigma` or `lam` as kwarg
    - `pm.StudentT` now requires `nu` to be specified (no longer defaults to 1)
    - `pm.AsymmetricLaplace` positional arguments re-ordered
    - `pm.AsymmetricLaplace` now requires `mu` to be specified (no longer defaults to 0)
```
Also removes a double `now`

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
         exclude: ^requirements-dev\.txt$
     -   id: trailing-whitespace
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v0.940
+  rev: v0.941
```
ricardoV94: This change shouldn't be in this PR. You might need to rebase from main

michaelosthege (approved):

PR #5629, opened 2022-03-20
Adding check to prevent mixing of discrete and continuous distributions

### LINKED ISSUE #4511: Mixture should not allow mixing of discrete and continuous distributions

larryshamalama:
Closes #4511. `all_discrete` in `mixture.py` has also been removed

Is there a test that should be added?

### DIFF
--- a/pymc/distributions/mixture.py
+++ b/pymc/distributions/mixture.py
@@ -177,7 +177,7 @@ class Mixture(SymbolicDistribution):
             if not (
                 all(comp_dist.dtype in continuous_types for comp_dist in comp_dists)
                 or all(comp_dist.dtype in discrete_types for comp_dist in comp_dists)
-                ):
+            ):
                 # Determine if the distributions in comp_dists are all discrete or continuous
                 raise ValueError(
                     "All distributions in comp_dists must be either discrete or continuous.\n"
--- a/pymc/distributions/mixture.py
+++ b/pymc/distributions/mixture.py
@@ -174,7 +174,6 @@ class Mixture(SymbolicDistribution):
                 all(comp_dist.dtype in continuous_types for comp_dist in comp_dists)
                 or all(comp_dist.dtype in discrete_types for comp_dist in comp_dists)
             ):
-                # Determine if the distributions in comp_dists are all discrete or continuous
                 raise ValueError(
                     "All distributions in comp_dists must be either discrete or continuous.\n"
                     "See the following issue for more information: https://github.com/pymc-devs/pymc/issues/4511."
--- a/pymc/distributions/mixture.py
+++ b/pymc/distributions/mixture.py
@@ -43,21 +43,6 @@ from pymc.vartypes import continuous_types, discrete_types
 __all__ = ["Mixture", "NormalMixture"]
 
 
-def all_same_type(comp_dists):
-    """
-    Determine if the distributions in comp_dists are all discrete or continuous
-    """
-
-    if len(comp_dists) == 1:
-        # should return True for either check_discrete=True or False
-        return comp_dists[0] in continuous_types | discrete_types
-    else:
-        all_continuous = all([comp_dist in continuous_types for comp_dist in comp_dists])
-        all_discrete = all([comp_dist in discrete_types for comp_dist in comp_dists])
-
-        return all_continuous or all_discrete
-
-
 class MarginalMixtureRV(OpFromGraph):
     """A placeholder used to specify a log-likelihood for a mixture sub-graph."""
 
@@ -185,11 +170,16 @@ class Mixture(SymbolicDistribution):
                 UserWarning,
             )
 
-        if not all_same_type(comp_dists):
-            raise ValueError(
-                "All distributions in comp_dists must be either discrete or continuous.\n"
-                "See the following issue for more information: https://github.com/pymc-devs/pymc/issues/4511."
-            )
+        if len(comp_dists) > 1:
+            all_continuous = all([comp_dist.dtype in continuous_types for comp_dist in comp_dists])
+            all_discrete = all([comp_dist.dtype in discrete_types for comp_dist in comp_dists])
+
+            if not (all_continuous or all_discrete):
+                # Determine if the distributions in comp_dists are all discrete or continuous
+                raise ValueError(
+                    "All distributions in comp_dists must be either discrete or continuous.\n"
+                    "See the following issue for more information: https://github.com/pymc-devs/pymc/issues/4511."
+                )
 
         # Check that components are not associated with a registered variable in the model
         components_ndim_supp = set()
--- a/pymc/distributions/mixture.py
+++ b/pymc/distributions/mixture.py
@@ -30,8 +30,6 @@ from pymc.distributions import transforms
 from pymc.distributions.continuous import Normal, get_tau_sigma
 from pymc.distributions.dist_math import check_parameters
 from pymc.distributions.distribution import (
-    Discrete,
-    Distribution,
     SymbolicDistribution,
     _moment,
     moment,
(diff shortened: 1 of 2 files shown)


pymc/distributions/mixture.py
```diff
@@ -27,28 +27,31 @@
...
     """
-    if isinstance(comp_dists, Distribution):
-        return isinstance(comp_dists, Discrete)
+
+    if len(comp_dists) == 1:
+        # should return True for either check_discrete=True or False
+        return comp_dists[0] in continuous_types | discrete_types
     else:
-        return all(isinstance(comp_dist, Discrete) for comp_dist in comp_dists)
+        all_continuous = all([comp_dist in continuous_types for comp_dist in comp_dists])
```
ricardoV94: shouldn't it be based on the comp_dist.dtype?

pymc/distributions/mixture.py
```diff
@@ -27,28 +27,31 @@
...
-def all_discrete(comp_dists):
+def all_same_type(comp_dists):
     """
-    Determine if all distributions in comp_dists are discrete
+    Determine if the distributions in comp_dists are all discrete or continuous
     """
-    if isinstance(comp_dists, Distribution):
-        return isinstance(comp_dists, Discrete)
+
+    if len(comp_dists) == 1:
```
ricardoV94: Can't we just return True immediately?

pymc/distributions/mixture.py
```diff
@@ -27,28 +27,31 @@
...
 from pymc.distributions.shape_utils import to_tuple
 from pymc.util import check_dist_not_registered
-from pymc.vartypes import discrete_types
+from pymc.vartypes import continuous_types, discrete_types
 
 __all__ = ["Mixture", "NormalMixture"]
 
 
-def all_discrete(comp_dists):
+def all_same_type(comp_dists):
```
ricardoV94: maybe we don't need this as a separate function?

ricardoV94:
> Is there a test that should be added?

Always!

pymc/distributions/mixture.py
larryshamalama: ...yes, late night brain farts 😅

pymc/distributions/mixture.py
larryshamalama: I guess so. Are there cases where this could be an issue?

pymc/distributions/mixture.py
larryshamalama: Do mean that we should just have the check in the `.dist()` method?

pymc/distributions/mixture.py
ricardoV94: Can't think of any

pymc/distributions/mixture.py
```diff
@@ -176,6 +164,17 @@ def dist(cls, w, comp_dists, **kwargs):
                 UserWarning,
             )
 
+        if len(comp_dists) > 1:
+            all_continuous = all([comp_dist.dtype in continuous_types for comp_dist in comp_dists])
+            all_discrete = all([comp_dist.dtype in discrete_types for comp_dist in comp_dists])
```
ricardoV94: Micro-optimization, we don't need to wrap them in a list if we only care about the boolean. Also if you put them in an `or` `if` statement it will not need to evaluate the second expression if the first is already `True`

```suggestion
            all_continuous = all(comp_dist.dtype in continuous_types for comp_dist in comp_dists)
            all_discrete = all(comp_dist.dtype in discrete_types for comp_dist in comp_dists)
```

larryshamalama: I agree with the list wrapping suggestion. Fixed!

However, would using `if`/`or` for cases where we don't have to evaluate the second expression be very beneficial in terms of code optimization? My first instinct is that defining it this way is better for readability purposes, but I don't know what would the best choice here

larryshamalama:
> > Is there a test that should be added?
> 
> Always!

Can you share a quick pointer on how to catch errors?

pymc/distributions/mixture.py
ricardoV94: In general, there is no reason to compute something we don't need afterwards. In this case we only need an if / else check so we should not compute everything. I think this is still pretty readable:

```python
 if not (
  all(comp_dist.dtype in continuous_types for comp_dist in comp_dists)
  or all(comp_dist.dtype in discrete_types for comp_dist in comp_dists)
):
  raise ValueError(...)

**Comment by ricardoV94** [comment]
> > > Is there a test that should be added?
> > 
> > 
> > Always!
> 
> Can you share a quick pointer on how to catch errors?

Usually with `with pytest.raises(ExpectedError, match="expected error string"): ...`

There are a bunch of examples in the codebase, here is just a random one: https://github.com/pymc-devs/pymc/blob/8b063f9c3b344753acdf91416cf2bb3c4533f80a/pymc/tests/test_aesaraf.py#L60-L61

File: pymc/distributions/mixture.py [hunk 5]
```diff
@@ -182,6 +166,20 @@ def dist(cls, w, comp_dists, **kwargs):
...
 
+        if len(comp_dists) > 1:
+            all_continuous = all(comp_dist.dtype in continuous_types for comp_dist in comp_dists)
+            all_discrete = all(comp_dist.dtype in discrete_types for comp_dist in comp_dists)
+
+            if not (
+                all(comp_dist.dtype in continuous_types for comp_dist in comp_dists)
+                or all(comp_dist.dtype in discrete_types for comp_dist in comp_dists)
+            ):
+                # Determine if the distributions in comp_dists are all discrete or continuous
```
ricardoV94: The error message is clear enough

```suggestion
```

pymc/tests/test_mixture.py
```diff
@@ -721,6 +722,31 @@ def test_broadcast_components(self, comp_dists, expected_shape):
...
+                Binomial.dist(n=10, p=0.5),
+                Normal.dist(),
+            ),
+            (
+                Categorical.dist(np.broadcast_to(1 / 3, (5, 2, 3))),
+                MvNormal.dist(np.ones(3), np.eye(3), shape=(5, 2)),
+            ),
+        ],
+    )
+    def test_preventing_mixing_cont_and_discrete(self, comp_dists):
```
ricardoV94: One test is probably enough here

ricardoV94 (commented):
Looks great. I left some minor suggestions

pymc/distributions/mixture.py
larryshamalama: Just removed it

pymc/tests/test_mixture.py
larryshamalama: Should I just include it in the function rather than use `pytest.mark.parametrize`?

ricardoV94: Yup

ricardoV94 (approved):

ricardoV94:
Awesome, thanks!

PR #5630, opened 2022-03-20
Run float32 jobs separately

### LINKED ISSUE #5626: Split floatX tests into their own CI job
Currently our CI has a `floatX` job matrix on Ubuntu and Windows, running each set of tests with both float32 and float64.

But there are just a handful of tests where floatX actually makes a difference.
Only these tests should run with a floatX job matrix, thereby dramatically reducing total CI runtime.

The following screenshot is an incomplete list of candidates:

image

michaelosthege:
This change will save a lot of CI resources.

For tests that should run with either floatX setting, `aesara.change_flags` can be used as seen in `test_types.py`.

Initially I wanted to parametrize the float32 job across Ubuntu & Windows, but the different shells require different syntax in the YAML. The best way to unify that would probably to outsource the `pytest` commands and `test-subset` into a non-YAML file, but I don't have time for that right now.

CI time reduced to ~3h45m 🎉

Closes #5626


ricardoV94 (approved):
I pushed a change that should fix the SMC tests

michaelosthege:
The coverage check for 6af5104 was good and in the last commit 5677646 I don't see how it could affect `sampling_jax.py` coverage.

Let's ship it :)

PR #5631, opened 2022-03-20
Remove SGMCMC and fix flaky mypy results

### LINKED ISSUE #5580: Remove SGMCMC step method
As discussed in #5508

michaelosthege:
I upgraded Aesara and aeppl in the pre-commit environment. Some changes in Aesara's type hints mandated updates here.

For `sampling_jax.py` the `run_mypy.py` script will no longer treat it as an error if it unexpectedly passes. This caused some confusion before.

Similarly, I removed `sgmcmc.py` which was the second file with flaky mypy results.

Closes #5580


pymc/distributions/shape_utils.py
```diff
@@ -618,4 +618,4 @@ def find_size(
 
 def rv_size_is_none(size: Variable) -> bool:
     """Check wether an rv size is None (ie., at.Constant([]))"""
-    return size.type.shape == (0,)
+    return size.type.shape == (0,) # type: ignore [attr-defined]
```
michaelosthege: Ideally this should be fixed upstream, but I didn't want to put `shape_utils.py` onto the list of non-passing files because for `shape_utils.py` it's quite valuable to enforce mypy checks.

.pre-commit-config.yaml
```diff
@@ -26,8 +26,8 @@ repos:
       - types-filelock
       - types-setuptools
       - arviz
-      - aesara==2.4.0
-      - aeppl==0.0.26
```
ricardoV94, line 30: Oh great, even more places for the dependencies xD

ricardoV94 (approved):

.pre-commit-config.yaml
michaelosthege: You can't imagine how delighted I was to learn that the greatness of YAML required me to add the dependencies in yet another place! `</sarcasm>`

PR #5633, opened 2022-03-20
Conditionally depend on pandas in tests

### LINKED ISSUE #5469: Remove Pandas dependency
Pandas is quite a large dependency that is not truly critical to PyMC. AFAIK we use it in two contexts (there could be more):

1. Format the output of `model.point_logps` (old `model.check_test_point`) by converting it into a Pandas Series. We can do just fine with a dictionary here
2. Do some data wrangling / detect NAN when a distribution is "observed". We can probably keep supporting this with an optional pandas import (try/except).

thomasjpfan:
Closes https://github.com/pymc-devs/pymc/issues/5469

This PR updates the tests to conditionally dependent on pandas. With this PR the only code that uses pandas are maintenance, benchmark scripts, and two notebooks:

- scripts/check_all_tests_are_covered.py
- scripts/run_mypy.py
- benchmarks/benchmarks/benchmarks.py
- docs/source/learn/examples/GLM_linear.ipynb
- docs/source/learn/examples/pymc_overview.ipynb

### DIFF
--- a/pymc/tests/test_aesaraf.py
+++ b/pymc/tests/test_aesaraf.py
@@ -19,7 +19,6 @@ import aesara.tensor as at
 import numpy as np
 import numpy.ma as ma
 import numpy.testing as npt
-import pandas as pd
 import pytest
 import scipy.sparse as sps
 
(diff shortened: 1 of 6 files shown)


michaelosthege:
Do you want to remove pandas from the `requirements.txt` then?
Since it's only used in the tests, maybe that would have been enough. But the changes look good - just one file where you forgot this `importskip` thing

ricardoV94:
@thomasjpfan May I ask what does CLN in your PR title stand for?

michaelosthege (approved):

thomasjpfan:
> May I ask what does CLN in your PR title stand for?

It means "Clean". I like to tag my commits so they are easier to track down in the future.

That being said, I removed it from the PR title.

michaelosthege:
Thanks, @thomasjpfan !

Lets hope this makes the life of conda/mamba dependency resolves a little less complicated.

Issue #5616, opened 2022-03-18
Latex representation broken for models with SymbolicDistributions


ricardoV94:
```python
with pm.Model() as model:
    x = pm.ZeroInflatedPoisson("x", 0.5, 5)
print(model)
```

```
  File "/home/ricardo/miniconda3/envs/roche_preclinical_hd/lib/python3.9/site-packages/IPython/core/formatters.py", line 224, in catch_format_error
    r = method(self, *args, **kwargs)
  File "/home/ricardo/miniconda3/envs/roche_preclinical_hd/lib/python3.9/site-packages/IPython/core/formatters.py", line 702, in __call__
    printer.pretty(obj)
  File "/home/ricardo/miniconda3/envs/roche_preclinical_hd/lib/python3.9/site-packages/IPython/lib/pretty.py", line 377, in pretty
    return self.type_pprinters[cls](obj, self, cycle)
  File "/home/ricardo/miniconda3/envs/roche_preclinical_hd/lib/python3.9/site-packages/pymc/printing.py", line 185, in _default_repr_pretty
    output = obj.str_repr()
  File "/home/ricardo/miniconda3/envs/roche_preclinical_hd/lib/python3.9/site-packages/pymc/printing.py", line 83, in str_for_model
    names = [s[: s.index("~") - 1] for s in rv_reprs]
  File "/home/ricardo/miniconda3/envs/roche_preclinical_hd/lib/python3.9/site-packages/pymc/printing.py", line 83, in <listcomp>
    names = [s[: s.index("~") - 1] for s in rv_reprs]
ValueError: substring not found
```

larryshamalama:
I can have a look into this

larryshamalama:
Seems like there are two things here that can be fixed: the [graphviz](https://github.com/pymc-devs/pymc/blob/main/pymc/model_graph.py#L155) and [LaTeX printing](https://github.com/pymc-devs/pymc/blob/main/pymc/printing.py#L93) representation of the model with `SymbolicDistributions`.

In general, the approach seems to first call `rv.owner.op`. Because each `SymbolicDistribution` defines `rv_op`s in their own way, i.e. not instances of `RandomVariable`, just looking at `rv.owner.op` probably won't work, to my understanding. For mixtures and zero-inflated distributions, `rv_op.owner.op` yields is an instance of `MarginalMixtureRV` and, for censored distributions, the owner op is a result from applying the Clip function.

[...]

PR #5634, opened 2022-03-21
Variable selection for graphviz visualization via kwarg

### LINKED ISSUE #5527: Variable selection for `pm.model_to_graphviz`

I have a big big model and would like to make `pm.model_to_graphviz` plots of only some of the variables inside.
For example after adding `gp.conditional`s and then only wanting to plot the nodes my new variable depends on.

My model also includes some `ConstantData` variables I use to store indexing information.
They aren't relevant for model understanding and I'd like to hide them.

## Proposal
The `pm.model_to_graphviz` function could take additional kwargs to customize the plot:
* `var_names` (or `vars`?) to plot only certain variables and their dependencies
* `show_disconnected_data: bool` to hide `ConstantData`/`MutableData` nodes that don't contribute to (selected) model variables

## Implementation
Variable selection should be straightforward since it's already in the constructor, just not accessible via a kwarg:

https://github.com/pymc-devs/pymc/blob/a3bab7ddc0f4612ad039a7d2755f22fa6c8c0a72/pymc/model_graph.py#L33-L36

larryshamalama:
Fixes #5527. Some elements are hackish and currently need more work.

+ [x] Do not plot all descendants when specifying an intermediate rv unless the variable is observed (plot observed node)
+ [x] Add `var_names` as kwarg
+ [x] Rename previous `var_names` class instance by `_all_var_names`

### DIFF
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -102,11 +102,27 @@ class ModelGraph:
         parents = self._get_ancestors(var, func)
         return self._filter_parents(var, parents)
 
-    def make_compute_graph(self) -> Dict[str, Set[VarName]]:
+    def vars_to_plot(self, selected_vars=None):
+        if selected_vars is None:
+            return self.var_names
+        else:
+            all_selected_ancestors = set()
+
+            for selected_var_name in selected_vars:
+                var_ancestors = self._get_ancestors(self.model[selected_var_name], self.model[selected_var_name])
+                all_selected_ancestors = all_selected_ancestors.union(var_ancestors)
+
+            vars_on_graph = all_selected_ancestors.union(set(self.model[selected_var_name] for selected_var_name in selected_vars))
+            vars_on_graph = {var.name for var in vars_on_graph}
+
+            # ordering of self.var_names is important
+            return [var_name for var_name in self.var_names if var_name in vars_on_graph]
+
+    def make_compute_graph(self, selected_vars=None) -> Dict[str, Set[VarName]]:
         """Get map of var_name -> set(input var names) for the model"""
         input_map = defaultdict(set)  # type: Dict[str, Set[VarName]]
 
-        for var_name in self.var_names:
+        for var_name in self.vars_to_plot(selected_vars):
             var = self.model[var_name]
             key = var_name
             val = self.get_parents(var)
@@ -120,6 +136,7 @@ class ModelGraph:
                         input_map[obs_name] = input_map[obs_name].union({var_name})
                 except AttributeError:
                     pass
+
         return input_map
 
     def _make_node(self, var_name, graph, *, formatting: str = "plain"):
@@ -168,7 +185,7 @@ class ModelGraph:
     def _eval(self, var):
         return function([], var, mode="FAST_COMPILE")()
 
-    def get_plates(self):
+    def get_plates(self, selected_vars=None):
         """Rough but surprisingly accurate plate detection.
 
         Just groups by the shape of the underlying distribution.  Will be wrong
@@ -179,7 +196,8 @@ class ModelGraph:
         dict: str -> set[str]
         """
         plates = defaultdict(set)
-        for var_name in self.var_names:
+
+        for var_name in self.vars_to_plot(selected_vars):
             v = self.model[var_name]
             if var_name in self.model.RV_dims:
                 plate_label = " x ".join(
@@ -189,9 +207,10 @@ class ModelGraph:
             else:
                 plate_label = " x ".join(map(str, self._eval(v.shape)))
             plates[plate_label].add(var_name)
+
         return plates
 
-    def make_graph(self, formatting: str = "plain"):
+    def make_graph(self, selected_vars=None, formatting: str = "plain"):
         """Make graphviz Digraph of PyMC model
 
         Returns
@@ -207,7 +226,7 @@ class ModelGraph:
                 "\tconda install -c conda-forge python-graphviz"
             )
         graph = graphviz.Digraph(self.model.name)
-        for plate_label, var_names in self.get_plates().items():
+        for plate_label, var_names in self.get_plates(selected_vars).items():
             if plate_label:
                 # must be preceded by 'cluster' to get a box around it
                 with graph.subgraph(name="cluster" + plate_label) as sub:
@@ -219,13 +238,18 @@ class ModelGraph:
                 for var_name in var_names:
                     self._make_node(var_name, graph, formatting=formatting)
 
-        for key, values in self.make_compute_graph().items():
-            for value in values:
-                graph.edge(value.replace(":", "&"), key.replace(":", "&"))
+        for child, parents in self.make_compute_graph(selected_vars=selected_vars).items():
+            # parents is a set of rv names that preceed child rv nodes
+            for parent in parents:
+                try:
+                    graph.edge(parent.replace(":", "&"), child.replace(":", "&"))
+                except AttributeError:
+                    print(child, parent)
+
         return graph
 
 
-def model_to_graphviz(model=None, *, formatting: str = "plain"):
+def model_to_graphviz(model=None, *, selected_vars=None, formatting: str = "plain"):
     """Produce a graphviz Digraph from a PyMC model.
 
     Requires graphviz, which may be installed most easily with
@@ -275,4 +299,4 @@ def model_to_graphviz(model=None, *, formatting: str = "plain"):
             "Formattings other than 'plain' are currently not supported.", UserWarning, stacklevel=2
         )
     model = pm.modelcontext(model)
-    return ModelGraph(model).make_graph(formatting=formatting)
+    return ModelGraph(model).make_graph(selected_vars=selected_vars, formatting=formatting)
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -118,7 +118,7 @@ class ModelGraph:
                     if model_var.tag.observations == self.model[var_name]:
                         selected_names.add(model_var.name)
 
-        selected_ancestors = set(
+        selected_ancestors = g(
             filter(
                 lambda rv: rv.name in self._all_var_names,
                 list(ancestors([self.model[var_name] for var_name in selected_names])),
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -14,7 +14,7 @@
 import warnings
 
 from collections import defaultdict, deque
-from typing import Dict, Iterator, List, NewType, Optional, Set
+from typing import Dict, Iterable, Iterator, List, NewType, Optional, Set
 
 from aesara import function
 from aesara.compile.sharedvalue import SharedVariable
@@ -108,7 +108,8 @@ class ModelGraph:
 
         selected_names = set(var_names)
 
-        for var_name in selected_names:
+        # .copy() because sets cannot change in size during iteration
+        for var_name in selected_names.copy():
             if var_name not in self._all_var_names:
                 raise ValueError(f"{var_name} is not in this model.")
 
@@ -124,7 +125,7 @@ class ModelGraph:
             )
         )
 
-        for var in selected_ancestors:
+        for var in selected_ancestors.copy():
             if hasattr(var.tag, "observations"):
                 selected_ancestors.add(var.tag.observations)
 
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -13,15 +13,15 @@
 #   limitations under the License.
 import warnings
 
-from collections import defaultdict, deque
-from typing import Dict, Iterable, Iterator, List, NewType, Optional, Set
+from collections import defaultdict
+from typing import Dict, Iterable, List, NewType, Optional, Set
 
 from aesara import function
 from aesara.compile.sharedvalue import SharedVariable
 from aesara.graph import Apply
 from aesara.graph.basic import ancestors, walk
 from aesara.tensor.random.op import RandomVariable
-from aesara.tensor.var import TensorConstant, TensorVariable
+from aesara.tensor.var import TensorConstant
 
 import pymc as pm
 
@@ -47,7 +47,7 @@ class ModelGraph:
                 return reversed(x.owner.inputs)
             return []
 
-        parents = [x.name for x in walk(nodes=var.owner.inputs, expand=_expand) if x.name]
+        parents = [get_var_name(x) for x in walk(nodes=var.owner.inputs, expand=_expand) if x.name]
 
         return parents
 
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -118,7 +118,7 @@ class ModelGraph:
                     if model_var.tag.observations == self.model[var_name]:
                         selected_names.add(model_var.name)
 
-        selected_ancestors = g(
+        selected_ancestors = set(
             filter(
                 lambda rv: rv.name in self._all_var_names,
                 list(ancestors([self.model[var_name] for var_name in selected_names])),
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -21,7 +21,7 @@ from aesara.compile.sharedvalue import SharedVariable
 from aesara.graph import Apply
 from aesara.graph.basic import ancestors, walk
 from aesara.tensor.random.op import RandomVariable
-from aesara.tensor.var import TensorConstant
+from aesara.tensor.var import TensorConstant, TensorVariable
 
 import pymc as pm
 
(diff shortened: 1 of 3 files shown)


pymc/model_graph.py
```diff
@@ -219,13 +238,18 @@ def make_graph(self, formatting: str = "plain"):
...
-        for key, values in self.make_compute_graph().items():
-            for value in values:
-                graph.edge(value.replace(":", "&"), key.replace(":", "&"))
+        for child, parents in self.make_compute_graph(selected_vars=selected_vars).items():
+            # parents is a set of rv names that preceed child rv nodes
+            for parent in parents:
+                try:
+                    graph.edge(parent.replace(":", "&"), child.replace(":", "&"))
+                except AttributeError:
+                    print(child, parent)
```
canyon289: Why print? Can this be a logging statement?

larryshamalama: Oops, this was left over from when I was debugging 😅 It shouldn't be there

larryshamalama:
@michaelosthege Can you have a look at these changes on a model or two? Happy to hear comments. It seems to work on my model in [this notebook](https://github.com/larryshamalama/pymc3-playground/blob/master/notebooks/shortcomings/graphviz-filter-out-vars.ipynb) but we can't see the figures on GitHub

pymc/model_graph.py
```diff
@@ -219,13 +242,15 @@ def make_graph(self, formatting: str = "plain"):
...
+        for child, parents in self.make_compute_graph(selected_vars=selected_vars).items():
+            # parents is a set of rv names that preceed child rv nodes
+            for parent in parents:
+                graph.edge(parent.replace(":", "&"), child.replace(":", "&"))
+
         return graph
 
 
-def model_to_graphviz(model=None, *, formatting: str = "plain"):
+def model_to_graphviz(model=None, *, selected_vars=None, formatting: str = "plain"):
```
michaelosthege: Let's align it with ArviZ: `var_names`

The ArviZ documentation describes it as:
```
var_names: list[str], optional
    List of variables to plot (defaults to None, which results in all variables plotted).
    Prefix the variables by ~ when you want to exclude them from the plot.
```

michaelosthege (commented):
It's going into the right direction.
Here's a minimal example that unveils two problems:
1. Selecting nodes causes some of them to change formatting
2. Nodes of observed data have an inverted dependency - this could be tricky.. One can get the desired result by selecting the observed RV, but because of the reversed arrow direction it's counterintuitive.

```python
import pymc as pm
from IPython.display import display

with pm.Model() as pmodel2:
    a = pm.Normal("a")
    b = pm.Normal("b")
    pm.Normal("c", a * b)
    intermediate = pm.Deterministic("intermediate", a + b)
    pred = pm.Deterministic("pred", intermediate * 3)

    obs = pm.ConstantData("obs", 1.75)

    pm.Normal("L", mu=1 + 0.5 * pred, observed=obs)

display(pm.model_to_graphviz(pmodel2))
display(pm.model_to_graphviz(pmodel2, selected_vars=["L"]))
display(pm.model_to_graphviz(pmodel2, selected_vars=["obs"]))
```

grafik

pymc/model_graph.py
```diff
@@ -102,11 +102,31 @@ def get_parents(self, var: TensorVariable) -> Set[VarName]:
...
+
+            # ordering of self.var_names is important
+            return [var_name for var_name in self.var_names if var_name in vars_on_graph]
+
+    def make_compute_graph(self, selected_vars=None) -> Dict[str, Set[VarName]]:
         """Get map of var_name -> set(input var names) for the model"""
         input_map = defaultdict(set)  # type: Dict[str, Set[VarName]]
 
-        for var_name in self.var_names:
+        for var_name in self.vars_to_plot(selected_vars):
```
michaelosthege: Commenting here because the diff above has all these codecov warnings...

In addition to the minimal example, I tested it with my big model and the result was missing a lot of variables.
The ancestor determination in `vars_to_plot()` appears to miss a lot of nodes.

The old code started from `self.var_names` which appears to be the complete list. Maybe the way how it's determined in `__init__` is relevant?

__EDIT__: I extended the example to demo this

pymc/model_graph.py
larryshamalama: I temporarily use `selected_vars` because `var_names` happens to be an instance variable. Would it be an issue or bad practice to use `var_names` as an argument for the function?

larryshamalama:
Thanks for the pointers and example. Will look into it.

Do you know approximately when the code has been written? My guess is that `_get_ancestors` was written sometime during v3 and doesn't work well anymore

michaelosthege:
> Thanks for the pointers and example. Will look into it.
> 
> Do you know approximately when the code has been written? My guess is that `_get_ancestors` was written sometime during v3 and doesn't work well anymore

You can check the git blame: https://github.com/pymc-devs/pymc/blame/main/pymc/model_graph.py

But generally the code works. Also for my biggest model all variables are shown

larryshamalama:
I believe that the plots now display desired graphs in your example above. Can you verify?

I am using `ancestors` from [`aesara.graph.basic.py`](https://github.com/aesara-devs/aesara/blob/main/aesara/graph/basic.py#L711) and there seems to many tools in that file that can be beneficial for this file. `get_parents` in `ModelGraph` still seems to work but I'm not sure about `_get_ancestors`, which is why I am using the Aesara one.

pymc/model_graph.py
```diff
@@ -102,11 +102,42 @@ def get_parents(self, var: TensorVariable) -> Set[VarName]:
...
-    def make_compute_graph(self) -> Dict[str, Set[VarName]]:
+    def vars_to_plot(self, selected_vars=None):
+        if selected_vars is None:
+            return self.var_names
+        else:
+            selected_vars = set(selected_vars)
+
+            for var_name in selected_vars:
+                if var_name not in self.var_names:
+                    raise ValueError(f"{var_name} is not part of the defined model.")
```
michaelosthege:
```suggestion
                    raise ValueError(f"{var_name} is not in this model.")
```

pymc/model_graph.py
```diff
@@ -102,11 +102,42 @@ def get_parents(self, var: TensorVariable) -> Set[VarName]:
         parents = self._get_ancestors(var, func)
         return self._filter_parents(var, parents)
 
-    def make_compute_graph(self) -> Dict[str, Set[VarName]]:
+    def vars_to_plot(self, selected_vars=None):
+        if selected_vars is None:
+            return self.var_names
+        else:
+            selected_vars = set(selected_vars)
```
michaelosthege: because of the return in the line above you can do
```suggestion
        selected_vars = set(selected_vars)
```
Ideally give it a new name, because it has a different type now: `sel_names: Set[str]`

pymc/model_graph.py
michaelosthege: Could also be done with sets: `outside_names = sel_names - set(self.var_names)`

pymc/model_graph.py
```diff
@@ -102,11 +102,42 @@ def get_parents(self, var: TensorVariable) -> Set[VarName]:
...
+
+            for var_name in selected_vars:
+                if var_name not in self.var_names:
+                    raise ValueError(f"{var_name} is not part of the defined model.")
+
+                for model_var in self.var_list:
+                    if hasattr(model_var.tag, "observations"):
+                        if model_var.tag.observations == self.model[var_name]:
+                            selected_vars = selected_vars.union({model_var.name})
```
michaelosthege:
```suggestion
                            selected_vars.add(model_var.name)
```

pymc/model_graph.py
```diff
@@ -102,11 +102,42 @@ def get_parents(self, var: TensorVariable) -> Set[VarName]:
...
+            selected_ancestors = set(
+                filter(
+                    lambda rv: rv.name in self.var_names,
+                    list(ancestors([self.model[var_name] for var_name in selected_vars])),
+                )
+            )
+
+            for var in selected_ancestors:
+                if hasattr(var.tag, "observations"):
+                    selected_ancestors = selected_ancestors.union({var.tag.observations})
```
michaelosthege:
```suggestion
                    selected_ancestors.add(var.tag.observations)
```

pymc/model_graph.py
```diff
@@ -102,11 +102,42 @@ def get_parents(self, var: TensorVariable) -> Set[VarName]:
...
+
+            selected_ancestors = [var.name for var in selected_ancestors]
+
+            # ordering of self.var_names is important
+            return selected_ancestors
```
michaelosthege: Is is sorted anywhere?
Does the return value need to be sorted?

And you can return the listcomprehension without creating another local variable

pymc/model_graph.py
```diff
@@ -102,11 +102,42 @@ def get_parents(self, var: TensorVariable) -> Set[VarName]:
         parents = self._get_ancestors(var, func)
         return self._filter_parents(var, parents)
 
-    def make_compute_graph(self) -> Dict[str, Set[VarName]]:
+    def vars_to_plot(self, selected_vars=None):
```
michaelosthege:
```suggestion
    def vars_to_plot(self, var_names: Optional[Iterable[str]]=None) -> List[str]:
```
Reading the code I was confused because `selected_vars` are the _names_, right?

pymc/model_graph.py
michaelosthege: I don't see a problem with using it as the argument.
You could even rename the `self.var_names` to `self._all_var_names` or so. They should have been private to begin with..

michaelosthege (commented):
> I believe that the plots now display desired graphs in your example above. Can you verify?

Yes, it's looking good for my big model too! Tomorrow morning I can then finally plot my model and print it on an A3 sheet of paper for the meeting on thursday 😃 So thank you for getting it to work 💯

Maybe a bit unexpected that selecting "L" also shows "obs", but it's probably a good idea to do that. It's a feature.

I added some type hint suggestions, and renaming of variables. My suggestions are incomplete, but I think you get the idea.

You can add tests to https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_model_graph.py. These tests work by comparing the dict representation of the grpah/plates with the expected outcome.

larryshamalama:
Thanks so much for the suggestions! I'll address them one by one. Just wanted you to have the figures for your meeting and work on polishing the code more slowly

pymc/model_graph.py
larryshamalama: Done!

pymc/model_graph.py
larryshamalama: Not sure about sorted, would have to revisit why I wrote this comment again

larryshamalama:
Reminder for myself to add a test

larryshamalama:
The solution currently is very ad-hoc as I do not understand `_get_ancestors` and `get_parents`. However, the code seems to work. Things to consider for next for this PR:

- Remove variables from model with ~
- Add a test to catch error when variable is not in the model

Things not sure if I should look into more
- refactor `get_parents` so it aligns with [`aesara.graph.basic.ancestors`](https://github.com/aesara-devs/aesara/blob/main/aesara/graph/basic.py#L711)

@michaelosthege can I get your input?

larryshamalama:
Quick comment that I did not forget about this PR, just busy with many other things. There's ongoing work with `networkx` integration, but I need to read up more on that.

larryshamalama:
I'm revisiting this PR and would like to push it to the finish line after a long hiatus 😅 In summary, this PR:

- adds the possibility to specify a subgraph to plot;
- adds a corresponding test for the above;
- refactors `get_parents` and removes (dangerously...) code that is no longer in use.

I'm hoping that all tests pass. Happy to hear any comments about this progress as it would be worthwhile to address issues #5616 and #5766 shortly afterwards. I imagine that I should add type hints to the methods?

michaelosthege:
Type hints are always good 👍

But also make sure to explain the new `var_names` kwarg in the docstring of `model_to_graphviz`.

michaelosthege (commented):
@larryshamalama we'll only need that docstinrg, then this is good to go!

larryshamalama:
Sounds good! I'll fix this later today on the train 😅

larryshamalama:
I only added the docstring to `pm.model_to_graphviz` since it seemed to be the only user-facing function. I also added type hints and, instead of `str`, I used [`VarName`](https://github.com/pymc-devs/pymc/blob/main/pymc/model_graph.py#L29).

On another note, I refactored a lot of the code since, with `v4`, many of it seemed irrelevant. The hope is that nothing breaks as I rely on functions from`aesara.graph.basic`, but I have a bit of uncertainty there. I'm happy to look into this file if anything else breaks in the future.

pymc/model_graph.py
```diff
@@ -32,85 +33,64 @@
...
+
+        # ordering of self._all_var_names is important
+        return [var.name for var in selected_ancestors]
+
+    def make_compute_graph(
+        self, var_names: Optional[Iterable[VarName]] = None
+    ) -> Dict[VarName, Set[VarName]]:
         """Get map of var_name -> set(input var names) for the model"""
-        input_map = defaultdict(set)  # type: Dict[str, Set[VarName]]
+        input_map = defaultdict(set)  # type: Dict[VarName, Set[VarName]]
```
michaelosthege:
```suggestion
        input_map: Dict[VarName, Set[VarName]] = defaultdict(set)
```

pymc/model_graph.py
```diff
@@ -168,18 +149,19 @@ def _make_node(self, var_name, graph, *, formatting: str = "plain"):
     def _eval(self, var):
         return function([], var, mode="FAST_COMPILE")()
 
-    def get_plates(self):
+    def get_plates(self, var_names: Optional[Iterable[VarName]] = None):
```
michaelosthege:
```suggestion
    def get_plates(
        self, var_names: Optional[Iterable[VarName]] = None
    ) -> Dict[str, Set[VarName]]:
```

pymc/model_graph.py
```diff
@@ -189,9 +171,10 @@ def get_plates(self):
             else:
                 plate_label = " x ".join(map(str, self._eval(v.shape)))
             plates[plate_label].add(var_name)
+
         return plates
 
-    def make_graph(self, formatting: str = "plain"):
+    def make_graph(self, var_names: Optional[Iterable[VarName]] = None, formatting: str = "plain"):
```
michaelosthege, line 177:
```suggestion
    def make_graph(
        self, var_names: Optional[Iterable[VarName]] = None, formatting: str = "plain"
    ) -> graphviz.Digraph:
```

pymc/model_graph.py
```diff
@@ -207,25 +190,29 @@ def make_graph(self, formatting: str = "plain"):
...
+            for parent in parents:
+                graph.edge(parent.replace(":", "&"), child.replace(":", "&"))
+
         return graph
 
 
-def model_to_graphviz(model=None, *, formatting: str = "plain"):
+def model_to_graphviz(
+    model=None, *, var_names: Optional[Iterable[VarName]] = None, formatting: str = "plain"
+):
```
michaelosthege, line 215:
```suggestion
) -> graphviz.Digraph:
```

pymc/model_graph.py
```diff
@@ -168,18 +149,19 @@ def _make_node(self, var_name, graph, *, formatting: str = "plain"):
...
+    def get_plates(self, var_names: Optional[Iterable[VarName]] = None):
         """Rough but surprisingly accurate plate detection.
 
         Just groups by the shape of the underlying distribution.  Will be wrong
         if there are two plates with the same shape.
 
         Returns
         -------
-        dict: str -> set[str]
+        dict: VarName -> set(VarName)
```
michaelosthege:
```suggestion
        dict
            Maps plate labels to the set of ``VarName``s inside the plate.
```

michaelosthege (commented):
I just have a few typing changes, otherwise this looks great 👍

I'll apply my suggestions & fix the pre-commit if needed... ⏳

michaelosthege (approved):

michaelosthege:
Thanks @larryshamalama !

I'm sure this will prove to be a very useful feature for the still underappreciated `pm.model_to_graphviz`

larryshamalama:
Thanks to you for your suggestions too :)

pymc/model_graph.py
```diff
@@ -32,85 +33,64 @@
...
+                lambda rv: rv.name in self._all_var_names,
+                list(ancestors([self.model[var_name] for var_name in selected_names])),
+            )
+        )
+
+        for var in selected_ancestors.copy():
+            if hasattr(var.tag, "observations"):
+                selected_ancestors.add(var.tag.observations)
+
+        # ordering of self._all_var_names is important
```
ricardoV94, line 81: I know this has been merged, but what's with this comment? I don't think this will be deterministically ordered since the input comes from a set.

larryshamalama: I'm not sure anymore... it's been a while since I wrote this. I guess we can remove it

Issue #3582, opened 2019-08-08
Mixture of discrete and contiuous distributions


aseyboldt:
Our samplers won't work correctly (except perhaps ABC) for variables that are neither discrete nor continuous. We should not allow the creation of such variables or we should throw and error when sampling from the model.
```
with pm.Model() as model:
    zero = pm.Constant.dist(DELTA)
    w = pm.Dirichlet('w', a=np.array([1,1]))

    # Should throw an error
    pm.Mixture("log_dist", w=w, comp_dists=[zero, nonzero])

    # TODO is this working correctly?
    pm.Mixture("log_obs", w=w, comp_dists=[zero, nonzero], observed=samples)
```
See
https://discourse.pymc.io/t/sampling-from-a-learned-mixture-of-zeros-and-lognormal/3671

junpenglao:
Yeah in general I dont think the mixture class should support components that have different support - this invalidate the mixture distribution assumption -  but I am not sure what is the best way to enforce this, we can check the domain if the comp_dist are pymc3 distributions, but if it is user supply density function there is no way to know.

PR #5636, opened 2022-03-21
Implement default transform for Mixtures

ricardoV94:
This PR implements default transforms for Mixture distributions. It will use the default transform of the underlying components, provided all the components use the same default transform.

This should make constrained latent Mixtures much more stable during NUTS sampling

Other changes:
* Rename `[_]get_moment` to `[_]moment` 
* Rename `_get_default_transform` to `_default_transform`
* Use dispatch for all default transforms instead of overriding `__new__`, this allows Mixtures to ask for the default transform of its components


ricardoV94:
~~Something seems wrong~~ Looks good:

```python
import pymc as pm
import arviz as az

with pm.Model() as m:
    halfnorm = pm.HalfNormal("halfnorm")
    comp_dists = [pm.HalfNormal.dist(), pm.HalfNormal.dist()]
    mix_transf = pm.Mixture("mix_transf", w=[0.5, 0.5], comp_dists=comp_dists)
    # mix = pm.Mixture("mix", w=[0.5, 0.5], comp_dists=comp_dists, transform=None)
    
    trace = pm.sample()
    
print(f"divergences={trace.sample_stats.diverging.sum().values}")
az.summary(trace)
```
```
divergences=19

             mean     sd  hdi_3%  hdi_97%  ...  mcse_sd  ess_bulk  ess_tail  r_hat
halfnorm    0.807  0.600     0.0    1.865  ...    0.012     981.0     920.0    1.0
mix_transf  0.847  0.607     0.0    1.954  ...    0.012    1108.0     914.0    1.0
```

And without transform:
```
divergences=2166

           mean     sd  hdi_3%  hdi_97%  ...  mcse_sd  ess_bulk  ess_tail  r_hat
mix       0.787  0.612   0.000    1.864  ...    0.017     672.0    1114.0   1.01
halfnorm  0.787  0.592   0.005    1.867  ...    0.013     781.0     784.0   1.01
```

ricardoV94:
Okay seems to be fixed now:

```python
             mean     sd  hdi_3%  hdi_97%  ...  mcse_sd  ess_bulk  ess_tail  r_hat
mix         0.819  0.605   0.001    1.898  ...    0.019     510.0     878.0   1.01
halfnorm    0.828  0.608   0.003    1.949  ...    0.013     663.0     572.0   1.01
mix_transf  0.803  0.612   0.001    1.933  ...    0.013     889.0     848.0   1.00
```

larryshamalama:
Haven't forgotten about this. I will have a look tomorrow evening NA time!

larryshamalama:
Haven't forgotten about this. I will have a look tomorrow evening NA time!

pymc/distributions/distribution.py
```diff
@@ -618,24 +618,24 @@ def dist(
 
 
 @singledispatch
-def _get_moment(op, rv, *rv_inputs) -> TensorVariable:
-    raise NotImplementedError(f"Variable {rv} of type {op} has no get_moment implementation.")
+def _moment(op, rv, *rv_inputs) -> TensorVariable:
+    raise NotImplementedError(f"Variable {rv} of type {op} has no moment implementation.")
```
larryshamalama, line 622: Just so I understand well, why exactly is it better/necessary to rename `_get_moment` to `_moment` and `get_moment` to `moment`? I feel like it's related to dispatching but I'm not familiar with this concept

pymc/tests/test_mixture.py
```diff
@@ -19,11 +19,15 @@
 import pytest
 import scipy.stats as st
 
+from aeppl.transforms import IntervalTransform, LogTransform
+from aeppl.transforms import Simplex as SimplexTransform
```
larryshamalama, line 23: Minor comment: Should we just rename it to `SimplexTransform`?

larryshamalama: Just browsing the aeppl repo, I saw that you did in a recent commit. Upon bumping the aeppl dependency to a prospective release, this line should be changed as well

pymc/distributions/mixture.py
```diff
@@ -438,25 +444,86 @@ def marginal_mixture_logcdf(op, value, rng, weights, *components, **kwargs):
...
+    transforms.SumTo1,
+)
+
+
+@_default_transform.register(MarginalMixtureRV)
+def marginal_mixture_default_transform(op, rv):
+    rng, weights, *components = rv.owner.inputs
+
+    default_transforms = [
+        _default_transform(component.owner.op, component) for component in components
```
larryshamalama, line 502: Right now this is okay because we don't allow nested mixtures or components to be censored rvs, is that correct? Would it be a good idea to add a check or warning or comment about this? (for ourselves and users?)

pymc/distributions/mixture.py
```diff
@@ -438,25 +444,86 @@ def marginal_mixture_logcdf(op, value, rng, weights, *components, **kwargs):
...
+    ]
+
+    # If there are more than one type of default transforms, we do not apply any
+    if len({type(transform) for transform in default_transforms}) != 1:
```
larryshamalama, line 506: What happens if components don't have the same support? My understanding from this is that we do not perform any variable transformation.

I guess a follow-up question is if it is discouraged to have components with different supports. If so, would it be good to raise a warning?

larryshamalama (commented):
Looks great and left some questions as I was reading, as usual 😅

pymc/distributions/distribution.py
ricardoV94: It's not better, just a different name. The `get` feels a bit awkward in hindsight. Would be like having the `logp` and `logcdf` functions be called `get_logp`, `get_logcdf`...

pymc/tests/test_mixture.py
ricardoV94: Yup. We will be reminded because it will simply fail when that happens.

pymc/distributions/mixture.py
ricardoV94: Default transforms should work with those types of distributions as well. No need to treat them differently.

ricardoV94: What was your concern though?

pymc/distributions/mixture.py
ricardoV94: If they don't have the same default transform we don't try to apply any, which is how things worked so far.

Note that distributions may have the same support, but have different default transforms (say logodds vs Interval(0, 1)) and we still won't apply any transform. That's just for simplicity.

Whether people should use distributions with different supports... I think it depends on what the goal is. I don't know if we should raise a warning... Good question

lucianopaz: It is valid to mix components with different support. Some zero inflated distributions mix a Dirac delta at zero with a distribution that doesn’t include zero in its support. Imagine a zero inflated half normal, if we apply the half normal’s default log transform to 0, we get -inf, which would then cause problems down the line.

pymc/tests/test_mixture.py
```diff
@@ -19,11 +19,15 @@
...
 from aesara import tensor as at
+from aesara.tensor import TensorVariable
 from aesara.tensor.random.op import RandomVariable
 from numpy.testing import assert_allclose
 from scipy.special import logsumexp
 
+from pymc import Triangular
```
lucianopaz: Shouldn’t this import be under from distributions instead?

lucianopaz (approved):
This looks great @ricardoV94! I wonder if some mixture models might see some efficiency improvements by using the transforms

pymc/distributions/mixture.py
lucianopaz: Maybe we could emit a warning saying that the mixture has components with different support, but it should be a new dedicated warming type, so that it’s easy to catch and silence

ricardoV94: > It is valid to mix components with different support. Some zero inflated distributions mix a Dirac delta at zero with a distribution that doesn’t include zero in its support. Imagine a zero inflated half normal, if we apply the half normal’s default log transform to 0, we get -inf, which would then cause problems down the line.

That example which mixes discrete and continuous components should actually not be allowed until we properly implement such logp (#4511), and even then they would only work for observed variables not unobserved as they would need a specialized sampler (#3582), so the question of transforms does not apply.

Otherwise, 100% discrete or 100% continuous mixtures with different support should sample alright. Ideally (for continuous), the user would manually apply a transform that unconstrained the combined support of all the components. We simply won't try such thing here. I think I'll add the warning you mentioned

pymc/tests/test_mixture.py
ricardoV94: Yes, thanks

ricardoV94:
> This looks great @ricardoV94! I wonder if some mixture models might see some efficiency improvements by using the transforms

I think so. You can see my second comment with a HalfNormal mixture. The number of divergences went down quite a lot and it was now similar to what you would obtain when sampling a vanilla HalfNormal on its own.

lucianopaz:
> > This looks great @ricardoV94! I wonder if some mixture models might see some efficiency improvements by using the transforms
> 
> I think so. You can see my second comment with a HalfNormal mixture. The number of divergences went down quite a lot and it was now similar to what you would obtain when sampling a vanilla HalfNormal on its own.

I hadn’t noticed the number of divergences! This is a great change then! One question though, what happens with an ordered transform? Could one potentially chain transforms together? Like `Log` and then `Ordered`?

ricardoV94:
> I hadn’t noticed the number of divergences! This is a great change then! One question though, what happens with an ordered transform? Could one potentially chain transforms together? Like `Log` and then `Ordered`?

Yes, users can pass any custom transform. This is only used when transform is `UNSET` (by default). More generally, you might be talking abuot #5404

PR #5640, opened 2022-03-22
Refine Interval docstrings and fix typo

ricardoV94:


pymc/distributions/transforms.py
```diff
@@ -191,20 +191,20 @@ class Interval(IntervalTransform):
     Parameters
     ----------
     lower : int, float, or None
-        Lower bound of the interval transform. Must be a constant value. If ``None``, the
-        interval is not bounded below.
+        Lower bound of the interval transform. Must be a constant finite value. If
+        ``None``, the interval is not bounded below.
```
OriolAbril: suggestion:

```
lower : int or float, optional
    Lower bound of the interval transform. Must be a constant finite value.
    By default (``lower=None``) the interval is not bounded below.
```

as the call signature is not immediate to see for everyone I suggest something like this with `optional` instead of `or None` to better indicate it doesn't need to be given a value explicitly.

pymc/distributions/transforms.py
```diff
@@ -191,20 +191,20 @@ class Interval(IntervalTransform):
...
     upper : int, float or None
-        Upper bound of the interval transfrom. Must be a finite value. If ``None``, the
-        interval is not bounded above.
+        Upper bound of the interval transform. Must be a constant finite value. If
+        ``None``, the interval is not bounded above.
     bounds_fn : callable
```
OriolAbril:
```suggestion
    bounds_fn : callable, optional
```

pymc/distributions/transforms.py
ricardoV94: My only concern is to mention that they can't both be `None` or it will raise

OriolAbril: Can't they both be none if bounds_fn is provided? We could add a not on the description or a notes section that at least one of the arguments is required

ricardoV94: Yes they can be both None if the function is provided. The function itself cannot return both None either

OriolAbril: I think it is clear/straightforward from both lower/upper saying the default is not bounded. If none are provided and no function either then there is no bound to apply, why are you using the transformation? but we could definitely extend a bit on that in the extended summary or notes section.

ricardoV94: Yeah, I think you're right. I am being over-concerned here

OriolAbril (approved):

Issue #5581, opened 2022-03-11
Deprecate remaining VI methods


ricardoV94:
#4582 left some old VI methods marked as NotImplemented

Normalizing flows: https://github.com/pymc-devs/pymc/blob/ef5f91b41d733c64cf3ea785ec2760caae5d1bd4/pymc/variational/approximations.py#L363

And these: https://github.com/pymc-devs/pymc/blob/e987950b359edfe4a49c92ee67c57983fe0fc5d9/pymc/variational/opvi.py#L959-L962

We should decide whether we want to refactor or remove them

ricardoV94:
CC @ferrine, @fonnesbeck, @twiecki

twiecki:
My vote is on removing them, don't think anyone uses those. If they do, they'll probably complain. We could ask on discourse too.

fonnesbeck:
I'm pretty sure OVPI can go, but normalizing flows are something we would want. They are a pretty  useful and are available in most VI implementations these days (e.g. pyro, TF). That said, they can always be added back if anyone ever wants to implement them.

ferrine:
I'd vote for removal, agree with @fonnesbeck. This functionality did not prove to be super useful.

fonnesbeck:
@ferrine do you think it will be straightforward to get normalizing flows ported? Were they fully functional in v3?

ferrine:
they were, but that time I struggled with shape inference for parameter creation. In v3 I relied on test values for NF, now test values are less strict and it broke the internals. To fix this issue new mechanism has to be implemented to initialize normalizing flows.

ferrine:
To summarize there are 3 blocks of features

- [AEVB](https://arxiv.org/abs/1312.6114) that is all `self.local` occasions. It is rarely used, and not PyMC focus
- batched block sparse inference. All `self.batched` occasions. Was used to improve UX of AEVB
- Normalizing flows. Normalizing flows API is nice but stays very far from PyMC focus. I'm ok to just delete it

ricardoV94:
So shall we delete everything and open an issue to re-implement normalizing flows in the future (either here or in pymc-experimental)?

twiecki:
That sounds good to me.

fonnesbeck:
Yes, seems sensible, assuming "everything" is all methods with NotImplementedInference exceptions.

purna135:
I would like to work on this.

twiecki:
[...]

PR #5642, opened 2022-03-22
Deprecate remaining VI methods

purna135:
Addressing #5581

Replaced remaining NotImplementedInference methods with deprecation warning.


purna135:
I need some help with the next step; is adding a deprecation warning the right way to go?

ghost:
I think the idea was to just remove them (even though the title in the issue says "deprecate"). The methods are not implemented in v4, so there is really nothing to deprecate.

purna135:
Could someone please help me remove the Normalizing flows?
I'm not sure if I should remove the entire NormalizingFlowGroup class or just a few lines of code.

https://github.com/pymc-devs/pymc/blob/ef5f91b41d733c64cf3ea785ec2760caae5d1bd4/pymc/variational/approximations.py#L362-L364

canyon289:
@ferrine would you like to take a look here?

pymc/variational/approximations.py
```diff
@@ -270,14 +272,14 @@ def _new_initial(self, size, deterministic, more_replacements=None):
                 at.repeat(self.mean.reshape((1, -1)), size, -1),
                 self.histogram[self.randidx(size)],
             )
+        elif deterministic:
+            warnings.warn(
+                "Deterministic sampling from a Histogram is deprecated and will be removed soon.",
```
ferrine: Is it indeed broken or it was a false alarm?

ferrine: We can enable the test for that and check

pymc/variational/opvi.py
```diff
@@ -957,9 +957,9 @@ def __init_group__(self, group):
         if not group:
             raise GroupError("Got empty group")
         if self.local:
```
ferrine: I think the context of the issue was to entirely delete the code and `local` kwarg, feel free to just delete large portions of the code.

pymc/variational/opvi.py
```diff
@@ -957,9 +957,9 @@ def __init_group__(self, group):
...
-            raise NotImplementedInference("Local inferene aka AEVB is not supported in v4")
+            warnings.warn("Local inferene aka AEVB is deprecated.", DeprecationWarning)
         if self.batched:
```
ferrine: Same for `batched` kwarg and all its mentions. Just delete without hesitation

ferrine (commented):
As for normalizing flows I expect something like `rm pymc/variational/flows.py` as a good first step

fonnesbeck:
Is this enough to go on @purna135 ?

purna135:
I'll work on these suggestions and let you know if I need anything else.
Thank you very much.

michaelosthege:
Hi @purna135, this PR is one of the final steps towards `v4.0.0`. Could you pick it up again and push it over the finish line?

purna135:
Hi @michaelosthege sorry for the delay; I was preoccupied with my exam, which has now been done.
I will complete this RP in few days.

pymc/variational/opvi.py
```diff
@@ -957,9 +957,9 @@ def __init_group__(self, group):
         if not group:
             raise GroupError("Got empty group")
         if self.local:
-            raise NotImplementedInference("Local inferene aka AEVB is not supported in v4")
+            warnings.warn("Local inferene aka AEVB is deprecated.", DeprecationWarning)
```
twiecki: We should just delete this code, rather than deprecate.

purna135:
Hi @twiecki and @ferrine could you please have a look if this is the right way to go?

ferrine (approved):
Looks good to me, can be merged once tests pass

purna135:
Thank you very much, @ferrine. Could you kindly advise me on how to pass these failed tests?
In the doc string, I realised I left some references to `batched` and `local`, which I will remove in my next commit.

ferrine:
@purna135, the tests diff was collapsed. I think I missed something here. I have to go through it once more

pymc/tests/test_variational_inference.py
```diff
@@ -98,89 +87,6 @@ def three_var_model():
...
-            },
-        ),
-        (
-            pytest.raises(TypeError, match="No approximation is specified"),
-            {MeanFieldGroup: ["one", "two"]},
-        ),
-        (not_raises(), {MeanFieldGroup: ["one"], FullRankGroup: ["two", "three"]}),
-    ],
-)
-@ignore_not_implemented_inference
```
ferrine, line 125: the `ignore_not_implemented_inference` hook was used to ignore tests that fail due to the not supported functionality. 

`NotImplementedInference` is an explicit error not to confuse failures with other "bugs". Most of the tests that are deleted so far are valid tests. 

The commit that deletes all these tests should be reverted. Or, if not easy, you can use the original file that is present in master.

To summarise. 

1. The only obsolete thing in the test file is `ignore_not_implemented_inference`. 
2. Once it is removed, all tests should pass and no `NotImplementedInference` is raised.
3. The only tests to remove are that fail because of the removed functionality.

ferrine (changes requested):
I missed the collapsed diff of tests. Most of the deleted tests are valid. I've left a comment above with the explanation.

ferrine:
We did a quick video chat with @purna135 to avoid further confusion. Additional thing to revisit was the documentation. That missed my eye as well.

purna135:
Thank you so much for your time and help @ferrine : )

purna135:
Hello, @ferrine. I deleted several flow-related tests and refactored others, but there are still 10 tests that I am unable to refactor and have commented so that other tests can be checked. Could you please assist me in refactoring the tests I commented? 
Many thanks

Issue #5586, opened 2022-03-12
observed_data missing from InferenceData when data is cast


ricardoV94:
A user on discourse reported that `observed_data` is missing from the InferenceData returned from the jax samplers in `sampling_jax.py`. We should check if that is still the case and fix it.

https://discourse.pymc.io/t/pymc-v4-not-returning-observed-data-in-inferencedata-object/8916

zaxtax:
I just did a cursory check and `observed_data` is in the `InferenceData` object returned. Happy to add a unit test for this, but more interested in reproducing the user's error.

fbarfi:
Have been using sampling_jax with no problem -- observed_data always in inference data object.

ricardoV94:
Thanks @zaxtax and @fbarfi

I'll close the issue as it seems it was already solved

vethno:
I apologize for the delay in uploading the minimal example to PyMC Discourse. I am still having the issue.

https://discourse.pymc.io/t/pymc-v4-not-returning-observed-data-in-inferencedata-object/8916/3

Per the post, it works with some distributions, but when I use a mixture distribution like `pm.NegativeBinomial()` or `pm.ZeroInflatedNegativeBinomial()`

zaxtax:
Yep, I am able to reproduce this. Looking into it now.

zaxtax:
The problem actually shows up if you use regular `pm.sample` so it's likely something on the distribution as @OriolAbril suggested.

zaxtax:
It seems to be related to an interaction between `pm.Data` and `pm.NegativeBinomial`

The following contains `observed_data`:

````python
with pm.Model() as model:
     t = pm.Gamma("t", 1, 1)
     y = pm.NegativeBinomial('y', mu=t, alpha=0.5, observed=5)
     trace = pm.sample(tune=20, draws=40)
````

While this does not:

````python
with pm.Model() as model:
     c = pm.Data("c", 5)
     t = pm.Gamma("t", 1, 1)
     y = pm.NegativeBinomial('y', mu=t, alpha=0.5, observed=c)
     trace = pm.sample(tune=20, draws=40)
````

ricardoV94:
Does that problem with `pm.Data`, happen with other distributions?

zaxtax:
@ricardoV94 yep. I have replicated it with other discrete distributions like `Bernoulli` and `Binomial`

ricardoV94:
[...]

PR #5646, opened 2022-03-24
Recognize cast data in InferenceData

zaxtax:
This is adds a cast check to `extracting_obs_data` fixing #5586

### DIFF
--- a/pymc/tests/test_sampling.py
+++ b/pymc/tests/test_sampling.py
@@ -1078,7 +1078,7 @@ class TestSamplePriorPredictive(SeededTest):
             o = pm.Deterministic("o", obs)
             gen1 = pm.sample_prior_predictive(draws)
 
-        assert gen1.prior["y"].shape == (1, draws, n1)
+        assert gen1.prior_predictive["y"].shape == (1, draws, n1)
         assert gen1.prior["o"].shape == (1, draws, n1)
 
         n2 = 20
@@ -1086,7 +1086,7 @@ class TestSamplePriorPredictive(SeededTest):
         with m:
             gen2 = pm.sample_prior_predictive(draws)
 
-        assert gen2.prior["y"].shape == (1, draws, n2)
+        assert gen2.prior_predictive["y"].shape == (1, draws, n2)
         assert gen2.prior["o"].shape == (1, draws, n2)
 
     def test_density_dist(self):
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -47,6 +47,7 @@ from aesara.graph.basic import (
 from aesara.graph.fg import FunctionGraph
 from aesara.graph.op import Op, compute_test_value
 from aesara.sandbox.rng_mrg import MRG_RandomStream as RandomStream
+from aesara.scalar.basic import Cast
 from aesara.tensor.elemwise import Elemwise
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.shape import SpecifyShape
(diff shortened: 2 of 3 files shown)


zaxtax:
I think some of these failing tests might have bugs. For example, should observed variables be in "prior" as they are assumed by some tests. Or into "observed_data" like happens after this change?

ricardoV94:
> I think some of these failing tests might have bugs. For example, should observed variables be in "prior" as they are assumed by some tests. Or into "observed_data" like happens after this change?

Can you link to the specific tests that are failing?

zaxtax:
> > I think some of these failing tests might have bugs. For example, should observed variables be in "prior" as they are assumed by some tests. Or into "observed_data" like happens after this change?
> 
> Can you link to the specific tests that are failing?

pymc/tests/test_sampling.py::TestSamplePriorPredictive::test_shared
https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_sampling.py#L1070

pymc/tests/test_step.py::TestStepMethods::test_step_continuous
https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_step.py#L91

I have no idea why the second test fails. The first one does since gen1.prior doesn't have have the observation y after this commit. That instead is in observed_data or prior_predictive. This test fails even if all discrete variables in the model are made continuous.

ricardoV94:
Thanks @zaxtax 

It seems that the first test just needs to be tweaked for the new (correct) expectation.

For the second test, are you sure it is not just flaky? If not, can you narrow down which of the steppers is failing?

zaxtax:
> Thanks @zaxtax
> 
> It seems that the first test just needs to be tweaked for the new (correct) expectation.
> 
> For the second test, are you sure it is not just flaky? If not, can you narrow down which of the steppers is failing?

I think it's just flakey. I can make it fail less often if use more tuning steps. I don't know what it's suppose to test other than that all the step methods work.

ricardoV94:
> I think it's just flakey. I can make it fail less often if use more tuning steps. I don't know what it's suppose to test other than that all the step methods work.

We can just open an issue about the newly observed flakiness and we can ignore it for this PR. Either it is something that already happened or something subtle changed and should be addressed in a separate PR anyway. There is no reason why these changes here should affect sampling, as that happens before any inferencedata conversion. 

By the way do you see it failing in specific step methods? And do you mind opening that issue?

zaxtax:
> > I think it's just flakey. I can make it fail less often if use more tuning steps. I don't know what it's suppose to test other than that all the step methods work.
> 
> We can just open an issue about the newly observed flakiness and we can ignore it for this PR. Either it is something that already happened or something subtle changed and should be addressed in a separate PR anyway. There is no reason why these changes here should affect sampling, as that happens before any inferencedata conversion.
> 
> By the way do you see it failing in specific step methods? And do you mind opening that issue?

I'm going to make the test a bit more robust and open another PR for the observed flakiness.

pymc/tests/test_sampling.py
```diff
@@ -1078,15 +1078,15 @@ def test_shared(self):
             o = pm.Deterministic("o", obs)
             gen1 = pm.sample_prior_predictive(draws)
 
-        assert gen1.prior["y"].shape == (1, draws, n1)
+        assert gen1.prior_predictive["y"].shape == (1, draws, n1)
```
ricardoV94, line 1081: Can you merge this with the first commit?

pymc/tests/test_step.py
```diff
@@ -117,11 +117,11 @@ def test_step_continuous(self):
         for step in steps:
             idata = sample(
                 0,
-                tune=8000,
+                tune=12000,
```
ricardoV94: Let's not include this change in this PR. This test is quite lengthy already, and it's better to let it fail now and then to motivate someone to properly fix it, instead of monkey-patching it every time it fails in unrelated PRs. 

Otherwise your PR is ready to merge

zaxtax: Alright, I dropped those commits for now.

zaxtax:
@ricardoV94 ooo and it passes all the checks (for now)

ricardoV94 (approved):
Awesome, thanks for the help @zaxtax!

PR #5649, opened 2022-03-24
added default value for inplace

TimOliverMaier:
Just adding one line allowing for updating traces with out-of-sample predictions using pm.sample_psoterior_predictive()

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -1788,6 +1788,7 @@ def sample_posterior_predictive(
     if predictions:
         if extend_inferencedata:
             ikwargs.setdefault("idata_orig", trace)
+            ikwargs.setdefault("inplace", True)
         return pm.predictions_to_inference_data(ppc_trace, **ikwargs)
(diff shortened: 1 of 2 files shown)


OriolAbril:
Thanks, the change looks good. Could you add a test in https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_idata_conversion.py (or add an assert to an existing one) to make sure we can't go back to the buggy behaviour without tests failing?

TimOliverMaier:
Sure. Looking into it ...

TimOliverMaier:
I added a test function. Maybe I shouldn't have rebased. The PR seems a little confusing now. I contributed 2 commits. The test function can be found in 65cc846 ("added test for idata extension")

OriolAbril (approved):
Managed to fix the rebase and extended a bit the test to check for both groups and variables. It looks like we have an issue with CI right now and it will probably not pass, but the change and test look good, thanks!

PR #5651, opened 2022-03-26
Disable sampling of transformed variables from prior_predictive

purna135:
Addressing #5076

Deleted the code that was added to facilitate sampling from the prior as a starting point for SMC.


purna135:
Hello, @ricardoV94. I'm still having trouble with `make_initial_point_fn`. I saw your previous comment but was unable to put it into action. I need your assistance with this.

PR #5655, opened 2022-03-28
Add admonition about unobserved continuous censored distributions

ricardoV94:


ricardoV94:
Actually we could do that ourselves automatically...

pymc/distributions/censored.py
```diff
@@ -49,6 +49,16 @@ class Censored(SymbolicDistribution):
     upper: float or None
         Upper (right) censoring point. If `None`, the distribution will not be right censored.
 
+    Notes
```
OriolAbril: I think this should be a [warnings](https://numpydoc.readthedocs.io/en/latest/format.html#warnings) section with raw text and no admonition. 

It will look like: https://arviz-devs.github.io/arviz/api/generated/arviz.reloo.html (that also has a notes section)

pymc/distributions/censored.py
```diff
@@ -49,6 +49,16 @@ class Censored(SymbolicDistribution):
...
+    -----
+
+        .. warning:: Continuous censored distributions should only be used as likelihoods.
+            Continuous censored distributions are a form of discrete-continuous mixture
+            and as such cannot be sampled properly without a custom step sampler.
+            If you wish to sample such a distribution, you can add the latent uncensored
+            distribution to the model and then wrap it in a {class}`~pymc.Deterministic`
+            {func}`~pymc.math.clip`.
```
michaelosthege:
```suggestion
    Warnings
    --------
    Continuous censored distributions should only be used as likelihoods.
    Continuous censored distributions are a form of discrete-continuous mixture
    and as such cannot be sampled properly without a custom step sampler.
    If you wish to sample such a distribution, you can add the latent uncensored
    distribution to the model and then wrap it in a :class:`~pymc.Deterministic`
    :func:`~pymc.math.clip`.
```

michaelosthege:
@OriolAbril can you check my suggested edit here?

It looks like this could be an easy merge

OriolAbril (approved):

ricardoV94:
Thanks for the help!

Issue #5659, opened 2022-03-28
`pm.distributions.transforms.ordered` fails on >1 dimension


drbenvincent:
Applying the `pm.distributions.transforms.ordered` fails when applied to more than one dimension

## Minimum working example
```python
import pymc as pm
import numpy as np

COORDS = {'question': np.arange(10), 
          'thresholds': np.arange(4)}

with pm.Model(coords=COORDS) as model:
    kappa = pm.Normal('kappa', 
                      mu=[-3, -1,  1,  2], 
                      sigma=1, 
                      dims=['question', 'thresholds'],
                      transform=pm.distributions.transforms.ordered)

model.point_logps()
```

## Traceback
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/opt/miniconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/aesara/compile/function/types.py in __call__(self, *args, **kwargs)
    963             outputs = (
--> 964                 self.fn()
    965                 if output_subset is None

ValueError: Input dimension mismatch. (input[2].shape[1] = 4, input[4].shape[1] = 10)

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
/var/folders/3d/p29f9frn6d3c3gnfq2bxbd1c0000gn/T/ipykernel_31128/1701730019.py in <module>
----> 1 model.point_logps()

~/opt/miniconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/pymc/model.py in point_logps(self, point, round_vals)
   1726             for factor, factor_logp in zip(
   1727                 factors,
-> 1728                 self.compile_fn(factor_logps_fn)(point),
   1729             )
   1730         }

~/opt/miniconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/pymc/model.py in __call__(self, state)
   1825 
   1826     def __call__(self, state):
-> 1827         return self.f(**state)
   1828 
   1829 

~/opt/miniconda3/envs/pymc-dev-py39/lib/python3.9/site-packages/aesara/compile/function/types.py in __call__(self, *args, **kwargs)
    975                 if hasattr(self.fn, "thunks"):
    976                     thunk = self.fn.thunks[self.fn.position_of_error]
--> 977                 raise_with_op(
    978                     self.maker.fgraph,
    979                     node=self.fn.nodes[self.fn.position_of_error],

[...]

PR #5660, opened 2022-03-28
Fix `pm.distributions.transforms.ordered` fails on >1 dimension

purna135:
Addressing #5659

Kept the original dims in the log_jac_det by changing
```
     return at.sum(value[..., 1:], axis=-1)
```
to 
```
    return at.sum(value[..., 1:], axis=-1, keepdims=True)
```

### DIFF
--- a/pymc/distributions/transforms.py
+++ b/pymc/distributions/transforms.py
@@ -82,7 +82,7 @@ class Ordered(RVTransform):
         return y
 
     def log_jac_det(self, value, *inputs):
-        return at.sum(value[..., 1:], axis=-1)
+        return at.sum(value[..., 1:], axis=-1, keepdims=True)
 
 
 class SumTo1(RVTransform):
@@ -102,7 +102,7 @@ class SumTo1(RVTransform):
 
     def log_jac_det(self, value, *inputs):
         y = at.zeros(value.shape)
-        return at.sum(y, axis=-1)
+        return at.sum(y, axis=-1, keepdims=True)
 
 
 class CholeskyCovPacked(RVTransform):
--- a/pymc/tests/test_transforms.py
+++ b/pymc/tests/test_transforms.py
@@ -557,3 +557,19 @@ def test_interval_transform_raises():
         tr.Interval(at.constant(5) + 1, None)
 
     assert tr.Interval(at.constant(5), None)
+
+
+def test_transforms_ordered():
+    COORDS = {"question": np.arange(10), "thresholds": np.arange(4)}
+
+    with pm.Model(coords=COORDS) as model:
+        kappa = pm.Normal(
+            "kappa",
+            mu=[-3, -1, 1, 2],
+            sigma=1,
+            dims=["question", "thresholds"],
+            transform=pm.distributions.transforms.ordered,
+        )
+
+    log_prob = model.point_logps()
+    np.testing.assert_allclose(list(log_prob.values()), np.array([18.69]))
--- a/pymc/tests/test_transforms.py
+++ b/pymc/tests/test_transforms.py
@@ -560,16 +560,28 @@ def test_interval_transform_raises():
 
 
 def test_transforms_ordered():
-    COORDS = {"question": np.arange(10), "thresholds": np.arange(4)}
-
-    with pm.Model(coords=COORDS) as model:
-        kappa = pm.Normal(
-            "kappa",
+    with pm.Model() as model:
+        pm.Normal(
+            "x",
             mu=[-3, -1, 1, 2],
             sigma=1,
-            dims=["question", "thresholds"],
+            size=(10, 4),
             transform=pm.distributions.transforms.ordered,
         )
-
+        
     log_prob = model.point_logps()
     np.testing.assert_allclose(list(log_prob.values()), np.array([18.69]))
+
+
+def test_transforms_sumto1():
+    with pm.Model() as model:
+        pm.Normal(
+            "x",
+            mu=[-3, -1, 1, 2],
+            sigma=1,
+            size=(10, 4),
+            transform=pm.distributions.transforms.sum_to_1,
+        )
+        
+    log_prob = model.point_logps()
+    np.testing.assert_allclose(list(log_prob.values()), np.array([-56.76]))
--- a/pymc/tests/test_transforms.py
+++ b/pymc/tests/test_transforms.py
@@ -568,7 +568,7 @@ def test_transforms_ordered():
             size=(10, 4),
             transform=pm.distributions.transforms.ordered,
         )
-        
+
     log_prob = model.point_logps()
     np.testing.assert_allclose(list(log_prob.values()), np.array([18.69]))
 
@@ -582,6 +582,6 @@ def test_transforms_sumto1():
             size=(10, 4),
             transform=pm.distributions.transforms.sum_to_1,
         )
-        
+
     log_prob = model.point_logps()
     np.testing.assert_allclose(list(log_prob.values()), np.array([-56.76]))


purna135:
Hello @ricardoV94, after making these changes I ran the code mentioned in the issue, and it ran without error this time.
However, I am unsure how to select the appropriate test for this; could you please assist me with this?

ricardoV94:
You can just add the original issue example as a new test to `test_transforms.py` and assert the result of `point_logps` is what's expected.

purna135:
test has been added, please have a look.

pymc/tests/test_transforms.py
```diff
@@ -557,3 +557,19 @@ def test_interval_transform_raises():
...
+    COORDS = {"question": np.arange(10), "thresholds": np.arange(4)}
+
+    with pm.Model(coords=COORDS) as model:
+        kappa = pm.Normal(
+            "kappa",
+            mu=[-3, -1, 1, 2],
+            sigma=1,
+            dims=["question", "thresholds"],
+            transform=pm.distributions.transforms.ordered,
+        )
```
ricardoV94: We don't need to use coords for this example.

```suggestion
def test_transforms_ordered():

    with pm.Model() as model:
        kappa = pm.Normal(
            "kappa",
            mu=[-3, -1, 1, 2],
            sigma=1,
            size=(10, 4),
            transform=pm.distributions.transforms.ordered,
        )
```

Also should test with the other transform that you have changed.

ricardoV94 (changes requested):
Looks good. Left a suggestion and request

pymc/tests/test_transforms.py
purna135: Done : )
but what about the other tests which are falling ?

drbenvincent:
Thanks for the contribution so far @purna135 ! I have my eye on this, and it looks like there are just two failing tests before this can be merged?

purna135:
Thank you so much, @drbenvincent; I was hoping for a response.
Yes, some tests are failing, but I can't figure out how to fix them; could you please assist me?

drbenvincent:
> Thank you so much, @drbenvincent; I was hoping for a response.
> Yes, some tests are failing, but I can't figure out how to fix them; could you please assist me?

I would gladly, if I could, but so far my contributions have been to `pymc-examples`, and not to `pymc` itself. So I do not have enough experience. Maybe @ricardoV94 could give some hints when he has time?

ricardoV94:
@purna135 Have you tried to investigate why those tests are failing locally?

purna135:
Yes, I ran the test locally, but I couldn't figure out what was causing it.
I assumed that because we changed `log_jac_det`, the expectation of the following line changed as well. 
`jacob_det = transform.log_jac_det(test_array_transf, *x.owner.inputs)`

As a result, I think we should revise the `assert` statement.

ricardoV94:
You can print the results before and after the changes to be sure what's changed. Then we have to figure out if we introduced a bug or if the old results were wrong (probably the former)

purna135:
Is this `jacob_det`  the one I should print to check the results?

purna135:
Hi @ricardoV94 I'm still waiting for a response and am unsure how to proceed. Could you please assist me?

pymc/distributions/transforms.py
```diff
@@ -82,7 +82,7 @@ def forward(self, value, *inputs):
         return y
 
     def log_jac_det(self, value, *inputs):
-        return at.sum(value[..., 1:], axis=-1)
+        return at.sum(value[..., 1:], axis=-1, keepdims=True)
```
Sayam753, line 85: Setting `keepdims=True` will keep the last dimension while computing logprob for multivariate distributions. This behaviour can be a lot confusing to interpret when adding transforms for multivariate distributions.

Example:
```python
from pymc.distributions.logprob import joint_logpt

with pm.Model() as model:
    mv = pm.MvNormal(
        "mv",
        mu=[-3, -1, 1, 2],
        cov=np.eye(4),
        initval=np.random.randn(4),
    )

print(joint_logpt(mv, sum=False)[0].ndim)  # will output 0
```

After adding Ordered transform
```python
from pymc.distributions.logprob import joint_logpt

with pm.Model() as model:
    mv = pm.MvNormal(
        "mv",
        mu=[-3, -1, 1, 2],
        cov=np.eye(4),
        initval=np.random.randn(4),
        transform=pm.distributions.transforms.ordered,
    )

print(joint_logpt(mv, sum=False)[0].ndim)  # will output 1
```

I think there can be a simple check here: https://github.com/aesara-devs/aeppl/blob/65d25c39fa771809b4d992c9323f59e78f282cd5/aeppl/transforms.py#L372-L388

to squeeze out the last dimension in case when transforms are added for multivariate distributions.

cc @ricardoV94

ricardoV94: We might need to tell the transform what is the support dimensionality of the RV. Can be an internal variable like `n` in the transform that's used by the LKJCholeskyCov

Sayam753: > We might need to tell the transform what is the support dimensionality of the RV.

If I get it right, we need to somehow pass `op.ndim_supp` to transforms to decide whether or not to keep last dimension. Seems like a simpler solution. 

Also, I think [SimplexTransform defined in AePPL](https://github.com/aesara-devs/aeppl/blob/main/aeppl/transforms.py#L289) needs to account for this change.

ricardoV94: We don't need to pass it to all transforms, we can instead initialize this transform with that information. Like we initialize Interval or LKJ transform with the specific information they need. This would follow a similar API and be less disruptive

In this case instead of having a single ordered already initialized, the user would initialize it (or us behind the scenes, like pm.interval does). It can be a kwarg that by default is 0 (for univariate RVs)

pseudo-code:
```python
class OrderedTransform(TransformRV):

  def __init__(self, ndim_supp):
    self.ndim_supp = ndim_supp

    # use self.ndim_supp to return the right shape

def ordered(ndim_supp=0):
    return OrderedTransform(ndim_supp)
```

ricardoV94: Is there anything wrong with the Simplex? That seems to be vector specific transform, I don't think it would be used in other cases.

We can think of it as being a VectorSimplex (i.e., a Simplex for RVs with ndim_supp=1), which is what it is used for in PyMC. If we wanted to use it with other ndim_supp we could make it more flexible...

Sayam753: Passing the required info to transforms during initialization indeed makes sense. @purna135 can you take up adding the proposed change in this PR itself?

> Is there anything wrong with the Simplex?

I noticed a test in `test_transforms.py` that applies Simplex transform to a batched univariate distribution. 

https://github.com/pymc-devs/pymc/blob/bae5087621323af3cb3cc95cf8d673da21ab4023/pymc/tests/test_transforms.py#L509-L526

This prompted me to make sure that even Simplex should not reduce the last dimension in case when applied to univariate distributions.

purna135: Thank you, @Sayam753. I'll work on these suggested changes and let you know if I need any assistance.

drbenvincent:
Hi. Just dropping a message here to see if anyone is able to push this forward? I have a vested interest in this fix, but am not familiar enough to help contribute efficiently.

TimOliverMaier:
Hey! @purna135, is there any news on this? I would also be interested in this fix. I tried to add the suggestions into the code, but I am still having failing tests and tbh a lack of understanding, as well. I would be happy to help here, but I can't accomplish it by myself, too.

TimOliverMaier:
I've been advancing on this locally. There are two remaining issues, as far as I see:
1.  There are some remaining failing tests that are caused by the changing dimensionality of the jacobian determinant <- Some 
     Transforms keep the last dimension, others do not. Keeping the other transforms as they are will make some further cases 
     in `test_transform.check_vectortransform_elementwise_logp()` necessary.
2.  Sampling of ordered multivariate distributions fails with `ValueError` :
     `"array must not contain infs or NaNs\nApply node that caused the error: SolveTriangular..."` 
     However, this is also the case on the main branch for me. 
What is the best way to collaborate on this PR? @purna135

purna135:
Hello, @TimOliverMaier. Thank you so much for addressing this issue and sorry for the delay in responding; I was preoccupied with other issues and didn't have time to solve this one.
It would be greatly appreciated if you could open a new PR from your local branch and I would close this one.

TimOliverMaier:
Hey, I opened a new PR. #6255 It is still work in progress, though.

ricardoV94:
Fixed in #6255

PR #5664, opened 2022-03-28
Replace point vmap with point map

### LINKED ISSUE #5656: Jax sampling OOM when computing stats and transformed
When I have a relatively big model, sampling goes just fine but when the method computes transformed vars I get OOM

It happens in
* [likelihood](https://github.com/pymc-devs/pymc/blob/main/pymc/sampling_jax.py#L141)
* [transformed](https://github.com/pymc-devs/pymc/blob/main/pymc/sampling_jax.py#L344)
* same with [blackjax](https://github.com/pymc-devs/pymc/blob/main/pymc/sampling_jax.py#L528)

I propose to change `vmap(vmap(fn))` to `vmap(map(fn))` since it is not so memory demanding

* PyMC/PyMC3 Version: 
* Aesara/Theano Version:
* Python Version:
* Operating system:

ferrine:
Resolves #5656

I've changed `jax.vmap(jax_fn)` with `partial(jax.lax.map, jax_fn)`. The reason is that `vmap` does the computation all at once. Sometimes it causes OOM. `jax.lax.map` is more memory friendly and, basically, does the computation in a loop.

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -149,13 +149,13 @@ def _sample_stats_to_xarray(posterior):
     return data
 
 
-def _get_log_likelihood(model: Model, samples) -> Dict:
+def _get_log_likelihood(model: Model, samples, map_fn=jax.vmap) -> Dict:
     """Compute log-likelihood for all observations"""
     data = {}
     for v in model.observed_RVs:
         v_elemwise_logpt = model.logpt(v, sum=False)
         jax_fn = get_jaxified_graph(inputs=model.value_vars, outputs=v_elemwise_logpt)
-        result = jax.jit(jax.vmap(smap(jax_fn)))(*samples)[0]
+        result = jax.jit(jax.vmap(map_fn(jax_fn)))(*samples)[0]
         data[v.name] = result
     return data
 
@@ -243,6 +243,7 @@ def sample_blackjax_nuts(
     var_names=None,
     keep_untransformed=False,
     chain_method="parallel",
+    postprocessing_method="vectorized",
     idata_kwargs=None,
 ):
     """
@@ -272,6 +273,8 @@ def sample_blackjax_nuts(
         Include untransformed variables in the posterior samples. Defaults to False.
     chain_method : str, default "parallel"
         Specify how samples should be drawn. The choices include "parallel", and "vectorized".
+    postprocessing_method : str, default "vectorized"
+        Specify how postprocessing should be computed. The choices include "sequential", and "vectorized".
     idata_kwargs : dict, optional
         Keyword arguments for :func:`arviz.from_dict`. It also accepts a boolean as value
         for the ``log_likelihood`` key to indicate that the pointwise log likelihood should
@@ -347,6 +350,14 @@ def sample_blackjax_nuts(
         raise ValueError(
             "Only supporting the following methods to draw chains:" ' "parallel" or "vectorized"'
         )
+    if postprocessing_method == "vectorized":
+        post_map_fn = jax.vmap
+    elif postprocessing_method == "sequential":
+        post_map_fn = smap
+    else:
+        raise ValueError(
+            "Only supporting the following methods to draw chains:" ' "sequential" or "vectorized"'
+        )
 
     states, _ = map_fn(get_posterior_samples)(keys, init_params)
     raw_mcmc_samples = states.position
@@ -358,7 +369,7 @@ def sample_blackjax_nuts(
     mcmc_samples = {}
     for v in vars_to_sample:
         jax_fn = get_jaxified_graph(inputs=model.value_vars, outputs=[v])
-        result = jax.jit(jax.vmap(smap(jax_fn)))(*raw_mcmc_samples)[0]
+        result = jax.jit(jax.vmap(post_map_fn(jax_fn)))(*raw_mcmc_samples)[0]
         mcmc_samples[v.name] = result
 
     tic4 = datetime.now()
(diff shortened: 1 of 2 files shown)


zaxtax:
The `jax.vmap` is there on purpose since it speeds up postprocessing, which can become a bottleneck otherwise. Perhaps we should add an `sequential` argument there similarly to the way numpyro does this during sampling?

https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/mcmc.py#L602=

pymc/sampling_jax.py
```diff
@@ -272,6 +273,8 @@ def sample_blackjax_nuts(
         Include untransformed variables in the posterior samples. Defaults to False.
     chain_method : str, default "parallel"
         Specify how samples should be drawn. The choices include "parallel", and "vectorized".
+    postprocessing_method : str, default "vectorized"
```
ricardoV94: A bit repetitive, but this is missing from `sample_numpyro_nuts` docstrings

pymc/sampling_jax.py
```diff
@@ -255,6 +273,8 @@ def sample_blackjax_nuts(
...
```
zaxtax: Does it make sense for `chain_method` and `postprocessing_method` to be different arguments? I'm inclined to think you made the correct choice here.

zaxtax (approved):
Looks good!

ricardoV94:
@ferrine were you seeing problems because of the log-likelihood or also transforms? The log-likelihood often leads to memory issues, but can already be disabled with `idata_kwargs`

ferrine:
> @ferrine were you seeing problems because of the log-likelihood or also transforms? The log-likelihood often leads to memory issues, but can already be disabled with `idata_kwargs`

I have troubles with transformed variables

pymc/sampling_jax.py
ferrine: I have a big model with low dimensional state space -> samples under vectorized quickly, but the transformed space is large and the "getting transformed variables" step gets me OOM.

ferrine:
the proposed method did not solve my issue, I'm inclined to admit that its more a modelling problem than the package problem. What would solve the problem is a map with a CPU buffer for output, but I am not sure it should be here. But this is the only way so get the transformed variables out of the model

ferrine:
@zaxtax, @ricardoV94 do you think we should provide a user with an opportunity to sample a really large model with the fallback to a CPU buffer in case transformed space gives OOM?

ricardoV94:
@ferrine I think that's a bit too much control at this point. Users already have some room with setting the JAX device backend, the chain vectorization method and whether or not to compute transforms and the likelihood. 

The code to go from a trace with untransformed variables to transformed variables is not too bad if someone really needs a custom implementation (so that they can sample in GPU but post-process in CPU as in your case)

zaxtax:
> the proposed method did not solve my issue, I'm inclined to admit that its more a modelling problem than the package problem. What would solve the problem is a map with a CPU buffer for output, but I am not sure it should be here. But this is the only way so get the transformed variables out of the model

I'm sorry this didn't solve your problem. I think this might be too custom to implement directly. Do you think a design where users can pass kwags to jax is enough to make a custom implementation not too onerous? `jax.jit` supports `backend` and `device` arguments that could be handy in general.

ferrine:
> > the proposed method did not solve my issue, I'm inclined to admit that its more a modelling problem than the package problem. What would solve the problem is a map with a CPU buffer for output, but I am not sure it should be here. But this is the only way so get the transformed variables out of the model
> 
> I'm sorry this didn't solve your problem. I think this might be too custom to implement directly. Do you think a design where users can pass kwags to jax is enough to make a custom implementation not too onerous? `jax.jit` supports `backend` and `device` arguments that could be handy in general.

Thanks, `jit(fn, backend=postprocessing_backend)` solved the OOM problem. Pushed the changes

zaxtax:
> > > the proposed method did not solve my issue, I'm inclined to admit that its more a modelling problem than the package problem. What would solve the problem is a map with a CPU buffer for output, but I am not sure it should be here. But this is the only way so get the transformed variables out of the model
> > 
> > 
> > I'm sorry this didn't solve your problem. I think this might be too custom to implement directly. Do you think a design where users can pass kwags to jax is enough to make a custom implementation not too onerous? `jax.jit` supports `backend` and `device` arguments that could be handy in general.
> 
> Thanks, `jit(fn, backend=postprocessing_backend)` solved the OOM problem. Pushed the changes

Nice!

ferrine:
@ricardoV94 @zaxtax merge?

zaxtax:
@ferrine I don't have privs to merge, but looks good to me.

pymc/sampling_jax.py
```diff
@@ -255,6 +256,8 @@ def sample_blackjax_nuts(
         Include untransformed variables in the posterior samples. Defaults to False.
     chain_method : str, default "parallel"
         Specify how samples should be drawn. The choices include "parallel", and "vectorized".
+    postprocessing_backend : Optional[str]
```
ricardoV94: I think the numpydoc type hints we are using are something like this:

```suggestion
    postprocessing_backend : str, optional
```

Same for the other function

pymc/tests/test_sampling_jax.py
```diff
@@ -145,15 +156,18 @@ def test_get_jaxified_logp():
         dict(log_likelihood=False),
     ],
 )
-def test_idata_kwargs(sampler, idata_kwargs):
+@pytest.mark.parametrize("postprocessing_backend", [None, "cpu"])
+def test_idata_kwargs(sampler, idata_kwargs, postprocessing_backend):
     with pm.Model() as m:
         x = pm.Normal("x")
+        z = pm.Normal("z")
```
ricardoV94, line 163: Does the new variable add anything?

ferrine: While refactoring I had a false positive test without it. Then I refactored back but think this use case is why not

ricardoV94 (approved):

PR #5667, opened 2022-03-29
Fix default update bug

### LINKED ISSUE #5653: Prior predictive incorrect randomness behaviour
## Random is weird

Sampling from prior in pymc4 I get weird results. Samples seem to be very correlated

```python
with pm.Model(coords=dict(time=range(150), i=range(5)), rng_seeder=42):
    rotated = pm.Normal("latent", dims=("time", "i"))
    t = pm.sample_prior_predictive(30)
plt.figure(figsize=(16, 4))
[plt.plot(t.prior.latent.sel(chain=0, i=1, draw=i).values, color="b", alpha=0.5) for i in [2, 7]];
```
image

* PyMC/PyMC3 Version: '4.0.0b5'
* Aesara/Theano Version: '2.5.1'
* Python Version: 3.10
* Operating system: Linux
* How did you install PyMC/PyMC3: (conda/pip) conda

ricardoV94:
Two changes:

1. No longer set `rng.default_update` when creating RVs via PyMC distributions
2. Just to be on the safe-side, update the `rng.default_update` in `change_rv_size`, when it followed the old "traditional" update rule

Closes #5653

### DIFF
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -142,7 +142,7 @@ def pandas_to_array(data):
 
 
 def change_rv_size(
-    rv_var: TensorVariable,
+    rv: TensorVariable,
     new_size: PotentialShapeType,
     expand: Optional[bool] = False,
 ) -> TensorVariable:
@@ -150,8 +150,8 @@ def change_rv_size(
 
     Parameters
     ==========
-    rv_var
-        The `RandomVariable` output.
+    rv
+        The old `RandomVariable` output.
     new_size
         The new size.
     expand:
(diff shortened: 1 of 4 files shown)


pymc/distributions/continuous.py
```diff
@@ -365,6 +365,10 @@ def __new__(cls, *args, **kwargs):
     @classmethod
     def dist(cls, *, size=None, **kwargs):
         res = super().dist([], size=size, **kwargs)
+        rng = res.owner.inputs[0]
+        # We set a dummy default_update that does not require sampling
+        if not hasattr(rng, "default_update"):
+            rng.default_update = rng
```
ricardoV94: I am not quite sure why this was not a problem with the previous `default_update`. But if I do not include these lines, this test fails: https://github.com/pymc-devs/pymc/blob/5b31ec71389c56789578b110b111a08d230c712e/pymc/tests/test_sampling.py#L684

ricardoV94: I see now, updates are always executed, even if they don't involve variables in the graph, whereas default_updates are not (obviously)

michaelosthege (approved):
LGTM

PR #5671, opened 2022-03-30
Bump version for b6 release

ricardoV94:
We fixed this critical bug: https://github.com/pymc-devs/pymc/issues/5653 which affected prior and posterior predictive for variables whose size was defined from `dims` or `observed` alone.

We should merge #5670 first


ricardoV94:
macos failing test seems just a bit flaky, the windows one has a known issue here: https://github.com/pymc-devs/pymc/issues/5665

michaelosthege (approved):

twiecki:
Should we update aesara and aeppl dependencies while we're at it?

michaelosthege:
> Should we update aesara and aeppl dependencies while we're at it?

Let's do the bump on another PR. If it goes green we can merge it before releasing. Otherwise after

PR #5688, opened 2022-04-04
Test tweaks

ricardoV94:
This PR cleans/fixes a couple of tests. 

The biggest change is the refactor of the test helpers in `TestMatchesScipy`, so that they can be used across different modules (see https://github.com/pymc-devs/pymc/discussions/5579), as well as in pymc-experimental.


michaelosthege (approved):
Looks good to me. If you want me to look closer, please comment it inline with a mention.

PR #5689, opened 2022-04-04
Misc changes

ricardoV94:
This PR is a minor spinoff of #4775


pymc/math.py
```diff
@@ -358,6 +358,22 @@ def log1mexp_numpy(x, *, negative_input=False):
     return out
 
 
+def log1pexp_numpy(x):
+    """Return log(1 + exp(x)).
+
+    This function is numerically more stable than the naive approach.
+    """
+    return np.where(
+        x < -37.0,
```
twiecki: where is that algorithm from?

ricardoV94: Copied from Aesara, there's a reference in the `at.log1pexp`. I will add it.

Also, this should not really be public, I'll take it from all.

ricardoV94:
Realized there was no point in implementing the `log1pexp_numpy` since the `np.where` will evaluate all the branches, precluding any speed benefits

twiecki (approved):

PR #5690, opened 2022-04-05
Add GRW automatic steps from shape

canyon289:
Copies of intent of this commit https://github.com/pymc-devs/pymc/pull/5541/commits/47571a7179ecab07d65829ea832938a9ca90d3f2 from this PR https://github.com/pymc-devs/pymc/pull/5541

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -18,6 +18,7 @@ import aesara.tensor as at
 import numpy as np
 
 from aesara import scan
+from aesara.raise_op import Assert
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.random.utils import normalize_size_param
 
@@ -167,8 +168,26 @@ class GaussianRandomWalk(distribution.Continuous):
 
         mu = at.as_tensor_variable(floatX(mu))
         sigma = at.as_tensor_variable(floatX(sigma))
+
+        # Check if shape contains information about number of steps
+        steps_from_shape = None
+        shape = kwargs.get("shape", None)
+        if shape is not None:
+            shape = to_tuple(shape)
+            if shape[-1] is not ...:
+                steps_from_shape = shape[-1] - 1
+
         if steps is None:
-            raise ValueError("Must specify steps parameter")
+            if steps_from_shape is not None:
+                steps = steps_from_shape
+            else:
+                raise ValueError("Must specify steps or shape parameter")
+        elif steps_from_shape is not None:
+            # Assert that steps and shape are consistent
+            steps = Assert(msg="Steps do not match last shape dimension")(
+                steps, at.eq(steps, steps_from_shape)
+            )
+
         steps = at.as_tensor_variable(intX(steps))
 
         # If no scalar distribution is passed then initialize with a Normal of same mu and sigma
--- a/pymc/tests/test_distributions_timeseries.py
+++ b/pymc/tests/test_distributions_timeseries.py
@@ -32,102 +32,100 @@ from pymc.tests.helpers import select_by_precision
 from pymc.tests.test_distributions_random import BaseTestDistributionRandom
 
 
-class TestGaussianRandomWalkRandom(BaseTestDistributionRandom):
-    # Override default size for test class
-    size = None
+class TestGaussianRandomWalk:
+    class TestGaussianRandomWalkRandom(BaseTestDistributionRandom):
+        # Override default size for test class
+        size = None
 
-    pymc_dist = pm.GaussianRandomWalk
-    pymc_dist_params = {"mu": 1.0, "sigma": 2, "init": pm.Constant.dist(0), "steps": 4}
-    expected_rv_op_params = {"mu": 1.0, "sigma": 2, "init": pm.Constant.dist(0), "steps": 4}
+        pymc_dist = pm.GaussianRandomWalk
+        pymc_dist_params = {"mu": 1.0, "sigma": 2, "init": pm.Constant.dist(0), "steps": 4}
+        expected_rv_op_params = {"mu": 1.0, "sigma": 2, "init": pm.Constant.dist(0), "steps": 4}
 
-    checks_to_run = [
-        "check_pymc_params_match_rv_op",
-        "check_rv_inferred_size",
-    ]
+        checks_to_run = [
+            "check_pymc_params_match_rv_op",
+            "check_rv_inferred_size",
+        ]
 
-    def check_rv_inferred_size(self):
-        steps = self.pymc_dist_params["steps"]
-        sizes_to_check = [None, (), 1, (1,)]
-        sizes_expected = [(steps + 1,), (steps + 1,), (1, steps + 1), (1, steps + 1)]
+        def check_rv_inferred_size(self):
+            steps = self.pymc_dist_params["steps"]
+            sizes_to_check = [None, (), 1, (1,)]
+            sizes_expected = [(steps + 1,), (steps + 1,), (1, steps + 1), (1, steps + 1)]
 
-        for size, expected in zip(sizes_to_check, sizes_expected):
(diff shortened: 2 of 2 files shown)


pymc/distributions/timeseries.py
```diff
@@ -157,15 +157,27 @@ def dist(
...
+            steps = to_tuple(shape)[-1] - 1
+        elif steps is not None:
+            pass
+        else:
+            raise ValueError("Must specify steps or shape parameter")
+
         steps = at.as_tensor_variable(intX(steps))
 
-        shape = kwargs.get("shape", None)
+        # TODO Ask what this does exactly
```
canyon289:
```suggestion
        # Determine the size of the init distribution

```

pymc/distributions/timeseries.py
```diff
@@ -157,15 +157,28 @@ def dist(
 
         mu = at.as_tensor_variable(floatX(mu))
         sigma = at.as_tensor_variable(floatX(sigma))
-        if steps is None:
-            raise ValueError("Must specify steps parameter")
+        shape = kwargs.get("shape", None)
+
+        # Get number of steps directly or from shape
+        if steps is not None and shape is not None:
+            raise ValueError("Specify only one of steps or shape parameter")
```
ricardoV94: This ValueError is wrong, shape contains more information than steps, including batched dimensions as in `shape=(4, 2)`

pymc/tests/test_distributions_timeseries.py
```diff
@@ -87,10 +87,12 @@ def test_gaussian_random_walk_init_dist_shape(init):
...
     assert tuple(grw.owner.inputs[-2].shape.eval()) == (5,)
 
-    grw = pm.GaussianRandomWalk.dist(mu=0, sigma=1, steps=1, init=init, shape=1)
+    # Test that if steps are none, they are inferred from shape
+    grw = pm.GaussianRandomWalk.dist(mu=0, sigma=1, steps=None, init=init, shape=1)
     assert tuple(grw.owner.inputs[-2].shape.eval()) == ()
 
-    grw = pm.GaussianRandomWalk.dist(mu=0, sigma=1, steps=1, init=init, shape=(5, 1))
+    # Test that if steps are none, they are inferred from shape
+    grw = pm.GaussianRandomWalk.dist(mu=0, sigma=1, steps=None, init=init, shape=(5, 1))
```
ricardoV94: These should be valid, related to ValueError check

ricardoV94 (changes requested):

ricardoV94:
Looks good, except the new `ValueError` is too restrictive

pymc/distributions/timeseries.py
canyon289: I agree that shape contains more information. What I'm catching in this error is if someone specifies both, for example
What happens if someone specifies something like `steps=1, shape=(2,5)` Which one is correct?

Or are you saying that this value error should just check if someone specifies a mismatch for example
`steps=4, shape=(2,5)` should pass?

ricardoV94: You can add an assert to the step argument if you're worried about that.

```python
steps = Assert("steps do not match last shape dimension")(shape[-1], at.eq(shape[-1], steps))
```

But I don't think it's a big problem to just overwrite it silently to be honest.

canyon289: Asserts shouldnt be used for production enforcement so an exception of some sort is the correct here. The reason for this is asserts can be disabled at runtime meaning this check would go away.

https://stackoverflow.com/questions/1273211/disable-assertions-in-python

I dont think we should silently override it. It can be confusing for a user who doesnt have the indepth knowledge of internals like we do

I can update the value error to check for mismatches if thats the only thing were concerned about

ricardoV94: That assert is a Symbolic exception, not a Python assert. The problem is that you cannot check what you want to check when the user calls the class, because both shapes and steps can be symbolic.

Also Aesara asserts can raise specific types of Errors, including `ValueError`: https://github.com/aesara-devs/aeppl/blob/0624b6b9b3301a81a74f5cf98c775c3f65b2f285/aeppl/logprob.py#L15-L29

canyon289: Oh man you are absolutely right. I keep forgetting these inputs can be symbolic and not actually determined till later. Thank you for pointing this out. Ill make the change in the next 96 hours

ricardoV94: Ugh... just realized that `shape` may have an `ellipsis`, which we didn't account for when trying to resize the `init` distribution... Allowing shape/ellipsis in `dist` open all sort of problems for composite distributions like this

michaelosthege:
@canyon289 can you push this PR over the finish line?

pymc/tests/test_distributions_timeseries.py
canyon289: Theyre valid! thats why theyre in the passing test function

pymc/distributions/timeseries.py
canyon289: Updated to what I think is right

pymc/tests/test_distributions_timeseries.py
ricardoV94: I meant the original were valid, before this PR

ricardoV94:
I am no longer sure about this. There are a couple of more special cases that would fail or redundant information would be ignored

1. Ellipsis in the last shape entry
2. Dims instead of shape
3. Observed instead of shape

The first would fail with the new logic and has to be fixed. The others are unfortunate that we don't support as well.

pymc/distributions/timeseries.py
```diff
@@ -157,15 +158,30 @@ def dist(
 
         mu = at.as_tensor_variable(floatX(mu))
         sigma = at.as_tensor_variable(floatX(sigma))
-        if steps is None:
-            raise ValueError("Must specify steps parameter")
+        shape = kwargs.get("shape", None)
+
+        # Get number of steps directly or from shape
+        if steps is not None and shape is not None:
+            steps = Assert("steps do not match last shape dimension")(
```
ricardoV94: Would need a test where this assert fails

ricardoV94 (changes requested):

pymc/distributions/timeseries.py
canyon289: That test is already included?

canyon289: Nevermind its not

canyon289: Ok it is now

canyon289:
How we feeling about this one?

canyon289:
CI is not feeling good about it :sob: 

I see the change is due to the recently added ellipsis changes, I wont be able to get to this for a while, a couple of weeks for sure, and then maybe after as well. If someone wants to take this over go for it, if not ill be back eventually

ricardoV94:
@canyon289 I force pushed some changes. Let me know if you spot any issues

canyon289:
Ill review in next 48 hours. Thanks Ricardo

pymc/distributions/timeseries.py
```diff
@@ -167,8 +168,26 @@ def dist(
...
         mu = at.as_tensor_variable(floatX(mu))
         sigma = at.as_tensor_variable(floatX(sigma))
+
+        # Check if shape contains information about number of steps
+        steps_from_shape = None
+        shape = kwargs.get("shape", None)
+        if shape is not None:
+            shape = to_tuple(shape)
+            if shape[-1] is not ...:
+                steps_from_shape = shape[-1] - 1
```
michaelosthege, line 178: I think this could break for `shape=tuple()`

...which is too low-dimensional to begin with.

michaelosthege (approved):
Looks good to me. I commented one detail, but go ahead however you like

canyon289:
Ah sorry for closing by accident

ricardoV94 (approved):

ricardoV94:
Merging this now as I will reuse this logic for the AR

PR #5698, opened 2022-04-07
Allow passing `compile_pymc` kwargs to sampling functions

ricardoV94:
This is needed for sampling of `RandomVariables` within `Scan`s among other things

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -43,7 +43,6 @@ import cloudpickle
 import numpy as np
 import xarray
 
-from aesara.compile.mode import Mode
 from aesara.graph.basic import Constant, Variable
 from aesara.tensor import TensorVariable
 from aesara.tensor.sharedvar import SharedVariable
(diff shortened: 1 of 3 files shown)


pymc/sampling.py
```diff
@@ -1746,13 +1746,16 @@ def sample_posterior_predictive(
     if size is not None:
         vars_to_sample = [change_rv_size(v, size, expand=True) for v in vars_to_sample]
 
+    if aesara_function_kwargs is None:
```
twiecki: Shouldn't this be `compile_kwargs`?

pymc/sampling.py
```diff
@@ -1599,6 +1597,8 @@ def sample_posterior_predictive(
     idata_kwargs : dict, optional
         Keyword arguments for :func:`pymc.to_inference_data` if ``predictions=False`` or to
         :func:`pymc.predictions_to_inference_data` otherwise.
+    compile_kwargs: dict, optional
```
twiecki, line 1599: Or this be `aesara_function_kwargs`

pymc/sampling.py
ricardoV94: yeah, typo when I changed the name

pymc/sampling.py
```diff
@@ -1746,13 +1745,16 @@ def sample_posterior_predictive(
     if size is not None:
         vars_to_sample = [change_rv_size(v, size, expand=True) for v in vars_to_sample]
 
+    if compile_kwargs is None:
```
twiecki, line 1748: Seems like the tests didn't catch that.

ricardoV94: What didn't they catch? There were a bunch of failing tests. Is there an error still?

twiecki: I didn't check, but did the test fail before you fixed the wrong name or did you add a new one that would have caught it?

ricardoV94: The tests (new and old) failed. I just didn't rerun all of them locally after doing the variable name change.

twiecki (approved):

PR #5699, opened 2022-04-08
Ignore variables out of idata as inputs in sample_posterior_predictive

### LINKED ISSUE #5695: New variables not sampled in posterior predictive when using InferenceData

I'm trying to do some predictions and forecasts that involve GPs, and deterministics. I run inference on an initial model, but then for the predictions I have to add new RVs and deterministics that are conditioned on the new time points. This doesn't seem to be possible at the moment.

```python
with pm.Model() as m:
    a = pm.Normal("a")
    b = pm.Normal("b", a, 0.1, observed=np.random.randn(10)* 0.1 + 3)
    ida = pm.sample()
# You sample on one model, then add deterministics on another and try to draw ppc samples
with m:
    d = pm.Deterministic("d", m["a"] - 4)
    c = pm.Normal("c", d, size=5)
    ppc = pm.sample_posterior_predictive(ida, var_names="c")
```

```python
KeyError                                  Traceback (most recent call last)
/tmp/ipykernel_14672/2975257511.py in <module>
      2     d = pm.Deterministic("d", m["a"] - 4)
      3     c = pm.Normal("c", d, size=5)
----> 4     ppc = pm.sample_posterior_predictive(ida, var_names="c")
      5 ppc

~/repos/pymc3/pymc/sampling.py in sample_posterior_predictive(trace, samples, model, var_names, size, keep_size, random_seed, progressbar, mode, return_inferencedata, extend_inferencedata, predictions, idata_kwargs)
   1769                 param = _trace[idx % len_trace]
   1770 
-> 1771             values = sampler_fn(*(param[n] for n in input_names))
   1772 
   1773             for k, v in zip(vars_, values):

~/repos/pymc3/pymc/sampling.py in <genexpr>(.0)
   1769                 param = _trace[idx % len_trace]
   1770 
-> 1771             values = sampler_fn(*(param[n] for n in input_names))
   1772 
   1773             for k, v in zip(vars_, values):

KeyError: 'd'
```

</details>

* PyMC/PyMC3 Version: 4.0.0b6
* Aesara/Theano Version: 2.5.1
* Python Version:
* Operating system:

lucianopaz:
Closes #5695 

Before you proceed, please make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).

I'm not sure if it's worth to mention in the release notes, since this was expected to work well before and it indeed worked with `MultiTrace` objects

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -1720,27 +1720,22 @@ def sample_posterior_predictive(
 
     inputs: Sequence[TensorVariable]
     input_names: Sequence[str]
-    if not hasattr(_trace, "varnames"):
-        inputs_and_names = [
-            (rv, rv.name)
-            for rv in walk_model(vars_to_sample, walk_past_rvs=True)
-            if rv not in vars_to_sample
-            and rv in model.named_vars.values()
-            and not isinstance(rv, (Constant, SharedVariable))
-        ]
-        if inputs_and_names:
-            inputs, input_names = zip(*inputs_and_names)
-        else:
-            inputs, input_names = [], []
+    if not isinstance(_trace, MultiTrace):
+        names_in_trace = list(_trace[0])
     else:
-        assert isinstance(_trace, MultiTrace)
-        output_names = [v.name for v in vars_to_sample if v.name is not None]
-        input_names = [
-            n
-            for n in _trace.varnames
-            if n not in output_names and not isinstance(model[n], (Constant, SharedVariable))
-        ]
-        inputs = [model[n] for n in input_names]
+        names_in_trace = _trace.varnames
+    inputs_and_names = [
+        (rv, rv.name)
+        for rv in walk_model(vars_to_sample, walk_past_rvs=True)
+        if rv not in vars_to_sample
+        and rv in model.named_vars.values()
+        and not isinstance(rv, (Constant, SharedVariable))
+        and rv.name in names_in_trace
+    ]
+    if inputs_and_names:
+        inputs, input_names = zip(*inputs_and_names)
+    else:
+        inputs, input_names = [], []
 
     if size is not None:
         vars_to_sample = [change_rv_size(v, size, expand=True) for v in vars_to_sample]
--- a/pymc/tests/test_sampling.py
+++ b/pymc/tests/test_sampling.py
@@ -24,6 +24,7 @@ import numpy as np
 import numpy.testing as npt
 import pytest
 import scipy.special
+import xarray as xr
 
 from aesara import Mode, shared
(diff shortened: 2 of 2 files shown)


pymc/tests/test_sampling.py
```diff
@@ -865,6 +866,27 @@ def test_idata_extension(self):
         fails = check_multiple_attrs(test_dict, idata)
         assert not fails
 
+    def test_deterministics_out_of_idata(self):
```
ricardoV94: Does this mean that deterministics now work in inferencedata but not multitrace?

ricardoV94: Ah the idata only contains `a`

ricardoV94: Can we parametrize this test for both idata/multitrace in any case?

pymc/sampling.py
```diff
@@ -1727,6 +1727,7 @@ def sample_posterior_predictive(
             if rv not in vars_to_sample
             and rv in model.named_vars.values()
             and not isinstance(rv, (Constant, SharedVariable))
+            and rv.name in _trace[0]  # Only accept tensors that have values in the trace points
```
ricardoV94: This and the else block are pretty confusing, and it looks like they should be doing the same. 

Can't we just save the list of variables for which we have known values (regardless of whether it is in a multitrace or idata) and then use the same logic to pick the input `RV`s and their names

pymc/tests/test_sampling.py
lucianopaz: They worked before with MultiTrace, but didn't work with InferenceData. This PR fixed that

ricardoV94: I know that, but it seems useful to test both cases to prevent a future regression when using MultiTrace. By the looks of it, this bug could have happened equally easily the other way around.

pymc/sampling.py
lucianopaz: Yes, we could do that too. I didn't want to change the code base too much though, and this filtering was all that was necessary to fix the problem with this particular posterior predictive sampling

ricardoV94: Cleaning it up would help maintaining things down the road. Do you mind doing it in this PR or do you prefer opening a separate issue?

lucianopaz: No problem, I’ll do it here

lucianopaz:
@ricardoV94 the code cov failure makes no sense. Feel free to review and we can try to get it merged regardless

ricardoV94 (approved):
Looks good!

PR #5684, opened 2022-04-02
some updates to how docs look

OriolAbril:
Some updates to how docs look, and addressing some of the todos in the learning page, arguably the most important page in the docs. Closes #5425 

The preview is available at: https://pymc--5684.org.readthedocs.build/en/5684/index.html

@martinacantaro I tried to continue your GSoD work, I hope I haven't botched it too much

@drbenvincent and @reshamas I think this will address some of your comments

PR #5700, opened 2022-04-08
Remove tidelift.

twiecki:


michaelosthege (approved):

OriolAbril (commented):
Looks good, but it might be better to merge #5684 first to avoid having to rebase there. #5684 removes the whole about folder instead of removing only the tidelift section and updating the toctree

Issue #5533, opened 2022-03-01
Reimplement nested Mixtures


ricardoV94:
Follow up from #5438 

Nested Mixtures or Mixtures using non pure RandomVariables (such as Censored variables) are not yet possible. This will require creating a dispatch to retrieve and manipulate these variables (i.e., check ndim_supp, resize), in the same way that's being done with the component RandomVariables. This should be straightforward, since these had to be implemented as methods of the SymbolicDistribution anyway...

PR #5702, opened 2022-04-08
Create dispatch functions for resize_dist and ndim_supp_dist

ricardoV94:
This allows us to better manipulate SymbolicDistributions and get closer to #5533

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -44,6 +44,7 @@ from pymc.distributions.shape_utils import (
     convert_shape,
     convert_size,
     find_size,
+    ndim_supp_dist,
     resize_from_dims,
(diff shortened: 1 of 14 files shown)


pymc/distributions/distribution.py
```diff
@@ -559,8 +564,11 @@ def dist(
         shape = convert_shape(shape)
         size = convert_size(size)
 
+        # Create a temporary dist to obtain the ndim_supp
+        ndim_supp = ndim_supp_dist(cls.rv_op(*dist_params, size=size))
```
ricardoV94, line 568: This is not great, but can't think of a less wasteful way

ricardoV94:
Closed in favor of #6072

PR #5708, opened 2022-04-11
improve total_size

ferrine:
Improving total_size API for observed variables to implement histogram trick

```python
values = np.random.randn(10)
counts = np.random.randint(1, 100, size=10)
```

<table>
  <tr>
    <td>Before</td><td>After</td>
  </tr>
  <tr>
   <td>
<pre>
with pm.Model():
    a = pm.Potential("obs", pm.logp(pm.Normal.dist(), values) * counts)
</pre>
   </td>
<td>
<pre>
with pm.Model():
    pm.Normal("obs", observed=values, total_size=counts)
</pre>
</td>
  </tr>
</table>
It also adds a warning

```
/home/ferres/dev/pymc3/pymc/distributions/logprob.py:104: UserWarning: Using array as scaling that is experimental
warnings.warn("Using array as scaling that is experimental", UserWarning)
```

### DIFF
--- a/pymc/distributions/logprob.py
+++ b/pymc/distributions/logprob.py
@@ -103,7 +103,7 @@ def _get_scaling(total_size: Optional[Union[int, Sequence[int]]], shape, ndim: i
         coef = at.prod(coefs)
     elif isinstance(total_size, (np.ndarray, at.TensorVariable)):
         coef = total_size
-        warnings.warn("Using array as scaling that is experimental", UserWarning)
+        warnings.warn("Using an array for total_size is experimental.", UserWarning)
     else:
         raise TypeError(
             "Unrecognized `total_size` type, expected int or list of ints, got %r" % total_size
--- a/pymc/distributions/logprob.py
+++ b/pymc/distributions/logprob.py
@@ -12,6 +12,8 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+import warnings
+
 from collections.abc import Mapping
 from typing import Dict, List, Optional, Sequence, Union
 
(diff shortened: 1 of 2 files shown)


pymc/distributions/logprob.py
```diff
@@ -99,6 +101,9 @@ def _get_scaling(total_size: Optional[Union[int, Sequence[int]]], shape, ndim: i
         end_coef = [floatX(t) / floatX(shp_end[i]) for i, t in enumerate(end) if t is not None]
         coefs = begin_coef + end_coef
         coef = at.prod(coefs)
+    elif isinstance(total_size, (np.ndarray, at.TensorVariable)):
+        coef = total_size
+        warnings.warn("Using array as scaling that is experimental", UserWarning)
```
twiecki:
```suggestion
        warnings.warn("Using an array for total_size is experimental.", UserWarning)
```

ferrine:
My only concern is that total size has double purpose. passing list or tuple, sets total size for minibatching as `shape[i]/total[i]`, passing array acts as multiplier

ricardoV94:
Why not make a specialized distribution:

```python
class WeightedNormalRV(NormalRV):
  name = "weighted_normal"
  ndim_params = [0, 0, 0]
  
  @classmethod
  rng_fn(cls, *args, **kwargs):
    # Filter out weight input
    return super().rng_fn(*args, **kwargs)

weighted_normal = WeightedNormalRV()

class WeightedNormal(Continuous):
  rv_op = weighted_normal

  @classmethod
  def dist(cls, mu, sigma, weights, **kwargs):
    return super().dist([mu, sigma, weights], **kwargs)

  def logp(value, mu, sigma, weights):
    return pm.logp(pm.Normal.dist(mu, sigma), value) * weights)
```

Or, instead, a distribution factory like `Mixture`/`Censored`?

ricardoV94:
I do think that the `Potential` is the right object here, because the weighted distribution is no longer a `Normal`, and users shouldn't use it for prior/posterior predictive, for instance.

ferrine:
Interesting. I did not think I could do this way

ferrine:
Maybe I put a helper function into https://github.com/pymc-devs/pymc-experimental?

ricardoV94:
> Maybe I put a helper function into https://github.com/pymc-devs/pymc-experimental?

Yeah, I think that's better

ferrine:
moved to https://github.com/pymc-devs/pymc-experimental/pull/38

PR #5711, opened 2022-04-13
Deterministic group order of Approximations

zcabaal:
In the `__init__` of the `Approximation` class, there is an issue with `list(set(...))` makes the order of elements not deterministic, and it varies between runs. This means runs are not reproducible even when fixing random seeds.

This PR fixes the issue, and make sure that the order of variables in the group is the same is the order is `model.free_RVs`

### DIFF
--- a/pymc/variational/opvi.py
+++ b/pymc/variational/opvi.py
@@ -1337,6 +1337,7 @@ class Approximation(WithMemoization):
                     raise GroupError("Found duplicates in groups")
                 seen.update(g.group)
                 self.groups.append(g)
+        # List iteration to preserve order for reproducibility between runs
         unseen_free_RVs = [var for var in model.free_RVs if var not in seen]
         if unseen_free_RVs:
             if rest is None:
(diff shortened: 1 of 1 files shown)


ricardoV94 (approved):

canyon289 (approved):

pymc/variational/opvi.py
```diff
@@ -1337,11 +1337,12 @@ def __init__(self, groups, model=None):
                     raise GroupError("Found duplicates in groups")
                 seen.update(g.group)
                 self.groups.append(g)
-        if set(model.free_RVs) - seen:
+        unseen_free_RVs = [var for var in model.free_RVs if var not in seen]
```
canyon289, line 1341:
```suggestion
        # List iteration to preserve order for reproducibility between runs
        unseen_free_RVs = [var for var in model.free_RVs if var not in seen]
```

ricardoV94:
Thanks for opening the PR @zcabaal

Issue #5722, opened 2022-04-21
InferenceData from prior predictions without constant data 


TimOliverMaier:
## constant_data missing after pm.sample_prior_predictive()

In some use-cases one might want to sample and plot prior predictions before sampling the posterior. For
this purpose, it would be nice to have "constant_data" group in the az.InferenceData object, that
pm.sample_prior_predictive() returns.

**Please provide a minimal, self-contained, and reproducible example.**
```python
import pymc as pm
import numpy as np

x = np.arange(-10,10)
y = 2*x+1.4+np.random.normal(0,0.3,x.size)

with pm.Model() as model:
    x = pm.MutableData("x",x)
    y = pm.MutableData("y",y)
    a = pm.Normal("a",mu=1,sigma=2)
    b = pm.Normal("b",mu=0,sigma=2)

    mu = pm.Deterministic("mu",var=a*x+b)
    e = pm.Exponential("e",lam=10)

    obs = pm.Normal("obs",mu=mu,sigma=e,observed=y)

with model:
    prior_pred = pm.sample_prior_predictive(model=model)

```

In the example above `prior_pred` has following groups:
+prior
+prior_predictive
+observed_data

## Versions and main components

* PyMC/PyMC3 Version: d322f7 (current main branch v4)
* Aesara/Theano Version: e5ebf2 (current main branch)
* Python Version: 3.10.4
* Operating system: Ubuntu 20.4
* How did you install PyMC/PyMC3: (conda/pip) pip -e

PR #5723, opened 2022-04-21
Prior predictions constant data

TimOliverMaier:
This is my suggestion for issue  #5722
With this minor changes, pm.sample_prior_predictive() returns an InferenceData object, that
contains the `constant_data` group, useful for e.g. plotting purposes of prior predictions before sampling.

### DIFF
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -468,7 +468,7 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
             default_dims=[],
         )
 
-    @requires(["trace", "predictions","prior_predictions"])
+    @requires(["trace", "predictions", "prior_predictions"])
     @requires("model")
     def constant_data_to_xarray(self):
         """Convert constant data to xarray."""
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -169,6 +169,10 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
             self.nchains = self.ndraws = 0
 
         self.prior = prior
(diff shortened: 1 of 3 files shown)


TimOliverMaier:
I don't know why the test is failing on windows. The failing test `test_step_continuous` passes on my system (Ubuntu 20.04). Can somebody tell me if the test is failing because of my additions or if there is a problem with the test?

pymc/backends/arviz.py
```diff
@@ -464,7 +468,7 @@ def observed_data_to_xarray(self):
             default_dims=[],
         )
 
-    @requires(["trace", "predictions"])
+    @requires(["trace", "predictions", "prior_predictions"])
```
OriolAbril: I think this prior predictions attribute is not needed, generating the constant data group here should be enough. IIUC, removing this line completely and leaving everything else untouched should be enough

pymc/tests/test_sampling.py
```diff
@@ -1042,6 +1042,28 @@ def point_list_arg_bug_fixture() -> Tuple[pm.Model, pm.backends.base.MultiTrace]
 
 
 class TestSamplePriorPredictive(SeededTest):
+    def test_idata_output(self):
```
OriolAbril: I think this test should be in idata_conversion. And I would try to avoid having a new test but instead checking for constant data in existing tests where it is currently missing

TimOliverMaier:
Yes, you are right. Thank you for the review. I pushed the changes.

OriolAbril (approved):

OriolAbril:
Thanks @TimOliverMaier !

PR #5724, opened 2022-04-23
Update pr_tutorial.md

juanitorduz:
Make the python version consistent on the PR tutorial, which seems to be `py38`.

### DIFF
--- a/docs/source/contributing/pr_tutorial.md
+++ b/docs/source/contributing/pr_tutorial.md
@@ -34,18 +34,18 @@ The preferred workflow for contributing to PyMC is to fork the [GitHub repositor
    :::{tab-item} Linux/MacOS
 
    ```bash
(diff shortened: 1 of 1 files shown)


canyon289:
I hesitate to change this to be a version number down. Where is the other source that is at 3.8? Can that be upversioned?

canyon289:
Apologies closed by accident but question remains

juanitorduz:
Sorry for the lack of description. The conflict IMO is with line 48 (see below), where the python version is 3.8

image

canyon289:
Ah great catch. lets change the ones below then to go to python 3.9. What would that take?

juanitorduz:
Suggested change added in https://github.com/pymc-devs/pymc/pull/5724/commits/9f5f7fc03a8f0f2c68a1faeec347e6b9394b98f6 :)

twiecki (approved):

PR #5738, opened 2022-05-03
Misc updates

ricardoV94:
* Remove uses of tag.ignore_logprob which is no longer recognized by Aeppl, and would result in a UserWarning
* Retrieve RNG updates for non RandomVariables at compile time

### DIFF
--- a/pymc/distributions/logprob.py
+++ b/pymc/distributions/logprob.py
@@ -20,6 +20,7 @@ import aesara.tensor as at
 import numpy as np
 
 from aeppl import factorized_joint_logprob
+from aeppl.abstract import assign_custom_measurable_outputs
 from aeppl.logprob import logcdf as logcdf_aeppl
 from aeppl.logprob import logprob as logp_aeppl
 from aeppl.transforms import TransformValuesOpt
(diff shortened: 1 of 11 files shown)


pymc/distributions/logprob.py
```diff
@@ -221,7 +222,11 @@ def joint_logpt(
 
     transform_opt = TransformValuesOpt(transform_map)
     temp_logp_var_dict = factorized_joint_logprob(
-        tmp_rvs_to_values, extra_rewrites=transform_opt, use_jacobian=jacobian, **kwargs
+        tmp_rvs_to_values,
+        extra_rewrites=transform_opt,
+        use_jacobian=jacobian,
+        warn_missing_rvs=False,
```
ricardoV94, line 228: Instead of warning, it will fail which is a more strict behavior

ricardoV94:
This test keeps failing: test_sample_posterior_predictive_after_set_data_with_coords across PRs

twiecki (approved):
Skimmed it.

ricardoV94:
Had missed one warning for the dist used in the GaussianRandomWalk

PR #5742, opened 2022-05-04
Add GaussianRandomWalk moment and fix broadcasting bug

ricardoV94:
This PR adds a moment to the GRW distribution and fixes bugs in the RandomVariable and logp methods when trying to broadcast mu/sigma across the time dimension

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -26,6 +26,7 @@ from pymc.aesaraf import change_rv_size, floatX, intX
 from pymc.distributions import distribution, multivariate
 from pymc.distributions.continuous import Flat, Normal, get_tau_sigma
 from pymc.distributions.dist_math import check_parameters
+from pymc.distributions.distribution import moment
 from pymc.distributions.logprob import ignore_logprob, logp
 from pymc.distributions.shape_utils import rv_size_is_none, to_tuple
 from pymc.util import check_dist_not_registered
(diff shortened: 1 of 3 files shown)


pymc/distributions/timeseries.py
```diff
@@ -214,27 +215,14 @@ def logp(
...
     ) -> at.TensorVariable:
-        """Calculate log-probability of Gaussian Random Walk distribution at specified value.
-
-        Parameters
-        ----------
-        value: at.Variable,
-        mu: at.Variable,
-        sigma: at.Variable,
-        init: at.Variable, Not used
-        steps: at.Variable, Not used
```
ricardoV94, line 229: These do not follow the style used in our docs and are wrong, because init and steps are both used. In any case, the logp function should be hidden from the users...

pymc/tests/test_distributions_timeseries.py
```diff
@@ -142,6 +143,28 @@ def test_gaussian_random_walk_init_dist_logp(self, init):
...
+                10,
+                (4, 2),
+                np.full((4, 2, 11), np.vstack((np.arange(11), -np.arange(11)))),
+            ),
+        ],
+    )
+    def test_moment(self, mu, sigma, init, steps, size, expected):
+        with Model() as model:
+            GaussianRandomWalk("x", mu=mu, sigma=sigma, init=init, steps=steps, size=size)
+        assert_moment_is_expected(model, expected)
```
ricardoV94, line 166: These tests cover the changes to the logp and random methods because `assert_moment_is_expected` calls both

twiecki (approved):

pymc/distributions/timeseries.py
canyon289: Even for functions and methods hidden from users we should write docstrings for developers when itll help. This one isnt all that useful so deleting is fine. Just want to note developers are a form of users too!

canyon289 (approved):

PR #5743, opened 2022-05-05
Obtain timeseries step information from dims or observed

ricardoV94:
Follow up to #5690 so that steps can also be retrieved from observed and dims when possible.

Other changes:
* Renamed `pandas_to_array` to `convert_observed_data`
* Made AR `steps` more consistent with expectation / GRW

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -13,7 +13,7 @@
 #   limitations under the License.
 import warnings
 
-from typing import Optional, Tuple, Union
+from typing import Any, Optional, Tuple, Union
 
 import aesara
 import aesara.tensor as at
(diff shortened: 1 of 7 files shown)


ricardoV94:
We really need to sort out what's going with https://github.com/pymc-devs/pymc/issues/5739, it's failing nearly every time

pymc/distributions/timeseries.py
```diff
@@ -245,6 +274,7 @@ def dist(
                 and init.owner.op.ndim_supp == 0
             ):
                 raise TypeError("init must be a univariate distribution variable")
+            check_dist_not_registered(init)
```
michaelosthege, line 277: Can you add a comment what this line does?

michaelosthege (approved):
Looks really good!
I particularly like the test cases.

pymc/distributions/timeseries.py
ricardoV94: I think the function self documents itself well enough: https://github.com/pymc-devs/pymc/blob/02897d159de9eec12951595cdc5869791f2e7f93/pymc/util.py#L337-L351

PR #5747, opened 2022-05-05
Refactor get_tau_sigma to handle lists

### LINKED ISSUE #5745: TypeError when passing list of values as sigma
`get_tau_sigma` used in many of our distributions fails with a `TypeError` when passing a list of values. We should probably fix this for better user experience:

```python
import pymc as pm:
pm.Normal.dist(sigma=[1, 2])
```
```
TypeError: unsupported operand type(s) for ** or pow(): 'list' and 'float'
```

fonnesbeck:
Non-`Variable` arguments converted to arrays before bounds checking. Added simple test.

Closes #5745

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -227,7 +227,8 @@ def get_tau_sigma(tau=None, sigma=None):
                 sigma_ = check_parameters(sigma, sigma > 0, msg="sigma > 0")
             else:
                 sigma_ = np.asarray(sigma)
-                assert np.all(sigma_ > 0), "sigma must be positive"
+                if np.any(sigma_ <= 0):
+                    raise ValueError("sigma must be positive")
             tau = sigma_**-2.0
 
     else:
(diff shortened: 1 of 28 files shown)


pymc/distributions/continuous.py
```diff
@@ -237,9 +237,8 @@ def get_tau_sigma(tau=None, sigma=None):
             if isinstance(tau, Variable):
                 tau_ = check_parameters(tau, tau > 0, msg="tau > 0")
             else:
-                assert np.all(np.asarray(tau) > 0)
-                tau_ = tau
-
+                tau_ = np.asarray(tau)
+                assert np.all(tau_ > 0), "tau must be positive"
```
ricardoV94: While we are here, we should replace the asserts by a ValueError

ricardoV94 (approved):

ghost:
Stochastic test failure?

fonnesbeck:
Hmm, bad rebase.

PR #5759, opened 2022-05-13
Add a method to compile an aesara forward sampling function

### LINKED ISSUE #5302: Deterministics not resampled in posterior predictive

I noticed that when I want to use `pm.set_data()`, if my model has a `pm.Deterministic` transformation that gets passed into the likelihood, I will get shape issues. 

Without the use of `pm.Deterministic` in my model, the shapes of posterior predictive samples, when I pass in new data of length 1, are of shape `(chain, draw, 1)`.

With the use of `pm.Deterministic` in my model, the shapes of posterior predictive samples, when I pass in new data, are of shape `(chain, draw, n)`, where `n` == "length of data that I used to fit the model".

I believe this is a bug with `pm.Deterministic`s interacting with `pm.set_data`, but I'm not quite sure how to pinpoint the problem.

I found it much easier to provide a reprex as a Jupyter notebook available here: https://gist.github.com/ericmjl/e82b40805805d9feeaff9f39182ff32c

No traceback necessary.

* PyMC/PyMC3 Version: 4.0.0b1
* Aesara/Theano Version: 2.3.2
* Python Version: 3.9.9
* Operating system: Ubuntu Linux
* How did you install PyMC/PyMC3: conda

lucianopaz:
Closes #5302

This PR adds a low level method to compile the aesara function responsible for generating forward samples. It introduces the notion of **volatile** variables in a graph, which allows it to determine which values it should take from an inferred posterior during forward sampling. Volatile variables are variables whose values could change between a run of `pm.sample` and future calls to `pm.sample_posterior_predictive`. These are:

- Variables in the outputs list (the ones that should actually be returned by the compiled function)
- SharedVariable instances
- RandomVariable instances that have volatile parameters
- RandomVariables that are not in the ``vars_in_trace`` list (this means that they don't have values in the inferred posterior, so they should be drawn from their a priori assumed distribution)
- Variables in the `givens_dict`, because setting values through `givens` is considered disruptive with respect to the values the variable could have taken during `pm.sample`.
- Any other type of variable that depends on volatile variables.

Tasks that remain before this PR is ready to merge:

- [x] Use `compile_forward_sampling_function` in `sample_prior_predictive`
~~- [ ] Use `compile_forward_sampling_function` in `sample_posterior_predictive_w`~~

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -43,8 +43,14 @@ import cloudpickle
 import numpy as np
 import xarray
 
-from aesara.graph.basic import Constant, Variable
-from aesara.tensor import TensorVariable
+from aesara import tensor as at
+from aesara.graph.basic import Apply, Constant, Variable, general_toposort, walk
+from aesara.graph.fg import FunctionGraph
+from aesara.tensor.random.op import RandomVariable
+from aesara.tensor.random.var import (
+    RandomGeneratorSharedVariable,
+    RandomStateSharedVariable,
+)
 from aesara.tensor.sharedvar import SharedVariable
 from arviz import InferenceData
 from fastprogress.fastprogress import progress_bar
@@ -52,7 +58,7 @@ from typing_extensions import TypeAlias
 
 import pymc as pm
 
-from pymc.aesaraf import change_rv_size, compile_pymc, inputvars, walk_model
+from pymc.aesaraf import change_rv_size, compile_pymc
 from pymc.backends.arviz import _DefaultTrace
 from pymc.backends.base import BaseTrace, MultiTrace
 from pymc.backends.ndarray import NDArray
@@ -75,6 +81,7 @@ from pymc.util import (
     get_default_varnames,
     get_untransformed_name,
     is_transformed_name,
+    point_wrapper,
 )
 from pymc.vartypes import discrete_types
 
@@ -83,6 +90,7 @@ sys.setrecursionlimit(10000)
 __all__ = [
     "sample",
     "iter_sample",
+    "compile_forward_sampling_function",
     "sample_posterior_predictive",
     "sample_posterior_predictive_w",
     "init_nuts",
@@ -1534,6 +1542,147 @@ def stop_tuning(step):
     return step
 
 
+def get_vars_in_point_list(trace, model):
+    """Get the list of Variable instances in the model that have values stored in the trace."""
+    if not isinstance(trace, MultiTrace):
+        names_in_trace = list(trace[0])
+    else:
+        names_in_trace = trace.varnames
+    vars_in_trace = [model[v] for v in names_in_trace]
+    return vars_in_trace
+
+
+def compile_forward_sampling_function(
+    outputs: List[Variable],
+    vars_in_trace: List[Variable],
+    basic_rvs: Optional[List[Variable]] = None,
+    givens_dict: Optional[Dict[Variable, Any]] = None,
+    **kwargs,
+) -> Callable[..., Union[np.ndarray, List[np.ndarray]]]:
+    """Compile a function to draw samples, conditioned on the values of some variables.
+
+    The goal of this function is to walk the aesara computational graph from the list
+    of output nodes down to the root nodes, and then compile a function that will produce
+    values for these output nodes. The compiled function will take as inputs the subset of
+    variables in the ``vars_in_trace`` that are deemed to not be **volatile**.
+
+    Volatile variables are variables whose values could change between runs of the
+    compiled function or after inference has been run. These variables are:
+
+    - Variables in the outputs list
+    - ``SharedVariable`` instances that are not ``RandomStateSharedVariable`` or ``RandomGeneratorSharedVariable``
+    - Basic RVs that are not in the ``vars_in_trace`` list
+    - Variables that are keys in the ``givens_dict``
+    - Variables that have volatile inputs
+
+    Where by basic RVs we mean ``Variable`` instances produced by a ``RandomVariable`` ``Op``
+    that are in the ``basic_rvs`` list.
+
+    Concretely, this function can be used to compile a function to sample from the
+    posterior predictive distribution of a model that has variables that are conditioned
+    on ``MutableData`` instances. The variables that depend on the mutable data will be
+    considered volatile, and as such, they wont be included as inputs into the compiled function.
+    This means that if they have values stored in the posterior, these values will be ignored
+    and new values will be computed (in the case of deterministics and potentials) or sampled
+    (in the case of random variables).
+
+    This function also enables a way to impute values for any variable in the computational
+    graph that produces the desired outputs: the ``givens_dict``. This dictionary can be used
+    to set the ``givens`` argument of the aesara function compilation. This will essentially
+    replace a node in the computational graph with any other expression that has the same
+    type as the desired node. Passing variables in the givens_dict is considered an intervention
+    that might lead to different variable values from those that could have been seen during
+    inference, as such, **any variable that is passed in the ``givens_dict`` will be considered
+    volatile**.
+
+    Parameters
+    ----------
+    outputs : List[aesara.graph.basic.Variable]
+        The list of variables that will be returned by the compiled function
+    vars_in_trace : List[aesara.graph.basic.Variable]
+        The list of variables that are assumed to have values stored in the trace
+    basic_rvs : Optional[List[aesara.graph.basic.Variable]]
+        A list of random variables that are defined in the model. This list (which could be the
+        output of ``model.basic_RVs``) should have a reference to the variables that should
+        be considered as random variable instances. This includes variables that have
+        a ``RandomVariable`` owner op, but also unpure random variables like Mixtures, or
+        Censored distributions. If ``None``, only pure random variables will be considered
+        as potential random variables.
+    givens_dict : Optional[Dict[aesara.graph.basic.Variable, Any]]
+        A dictionary that maps tensor variables to the values that should be used to replace them
+        in the compiled function. The types of the key and value should match or an error will be
+        raised during compilation.
+    """
+    if givens_dict is None:
+        givens_dict = {}
+
+    if basic_rvs is None:
+        basic_rvs = []
+
+    # We need a function graph to walk the clients and propagate the volatile property
+    fg = FunctionGraph(outputs=outputs, clone=False)
+
+    # Walk the graph from inputs to outputs and tag the volatile variables
+    nodes: List[Variable] = general_toposort(
+        fg.outputs, deps=lambda x: x.owner.inputs if x.owner else []
+    )
+    volatile_nodes: Set[Any] = set()
+    for node in nodes:
+        if (
+            node in fg.outputs
+            or node in givens_dict
+            or (  # SharedVariables, except RandomState/Generators
+                isinstance(node, SharedVariable)
+                and not isinstance(node, (RandomStateSharedVariable, RandomGeneratorSharedVariable))
+            )
+            or (  # Basic RVs that are not in the trace
+                node.owner
+                and isinstance(node.owner.op, RandomVariable)
+                and node in basic_rvs
+                and node not in vars_in_trace
+            )
+            or (  # Variables that have any volatile input
+                node.owner and any(inp in volatile_nodes for inp in node.owner.inputs)
+            )
+        ):
+            volatile_nodes.add(node)
+
+    # Collect the function inputs by walking the graph from the outputs. Inputs will be:
+    # 1. Random variables that are not volatile
+    # 2. Variables that have no owner and are not constant or shared
+    inputs = []
+
+    def expand(node):
+        if (
+            (
+                node.owner is None and not isinstance(node, (Constant, SharedVariable))
+            )  # Variables without owners that are not constant or shared
+            or node in vars_in_trace  # Variables in the trace
+        ) and node not in volatile_nodes:
+            # This test will include variables without owners, and that are not constant
+            # or shared, because these nodes will never be considered volatile
+            inputs.append(node)
+        if node.owner:
+            return node.owner.inputs
+
+    # walk produces a generator, so we have to actually exhaust the generator in a list to walk
+    # the entire graph
+    list(walk(fg.outputs, expand))
+
+    # Populate the givens list
+    givens = [
+        (
+            node,
+            value
+            if isinstance(value, (Variable, Apply))
+            else at.constant(value, dtype=getattr(node, "dtype", None), name=node.name),
+        )
+        for node, value in givens_dict.items()
+    ]
+
+    return compile_pymc(inputs, fg.outputs, givens=givens, on_unused_input="ignore", **kwargs)
+
+
 def sample_posterior_predictive(
     trace,
     samples: Optional[int] = None,
@@ -1718,38 +1867,23 @@ def sample_posterior_predictive(
             return trace
         return {}
 
-    inputs: Sequence[TensorVariable]
-    input_names: Sequence[str]
-    if not isinstance(_trace, MultiTrace):
-        names_in_trace = list(_trace[0])
-    else:
-        names_in_trace = _trace.varnames
-    inputs_and_names = [
-        (rv, rv.name)
-        for rv in walk_model(vars_to_sample, walk_past_rvs=True)
-        if rv not in vars_to_sample
-        and rv in model.named_vars.values()
-        and not isinstance(rv, (Constant, SharedVariable))
-        and rv.name in names_in_trace
-    ]
-    if inputs_and_names:
-        inputs, input_names = zip(*inputs_and_names)
-    else:
-        inputs, input_names = [], []
-
     if size is not None:
         vars_to_sample = [change_rv_size(v, size, expand=True) for v in vars_to_sample]
+    vars_in_trace = get_vars_in_point_list(_trace, model)
 
     if compile_kwargs is None:
         compile_kwargs = {}
+    compile_kwargs.setdefault("allow_input_downcast", True)
+    compile_kwargs.setdefault("accept_inplace", True)
 
-    sampler_fn = compile_pymc(
-        inputs,
-        vars_to_sample,
-        allow_input_downcast=True,
-        accept_inplace=True,
-        on_unused_input="ignore",
-        **compile_kwargs,
+    sampler_fn = point_wrapper(
+        compile_forward_sampling_function(
+            outputs=vars_to_sample,
+            vars_in_trace=vars_in_trace,
+            basic_rvs=model.basic_RVs,
+            givens_dict=None,
+            **compile_kwargs,
+        )
     )
 
     ppc_trace_t = _DefaultTrace(samples)
@@ -1775,7 +1909,7 @@ def sample_posterior_predictive(
             else:
                 param = _trace[idx % len_trace]
 
-            values = sampler_fn(*(param[n] for n in input_names))
+            values = sampler_fn(**param)
 
             for k, v in zip(vars_, values):
                 ppc_trace_t.insert(k.name, v, idx)
@@ -2063,16 +2197,16 @@ def sample_prior_predictive(
             names.append(rv_var.name)
             vars_to_sample.append(rv_var)
 
-    inputs = [i for i in inputvars(vars_to_sample) if not isinstance(i, (Constant, SharedVariable))]
-
     if compile_kwargs is None:
         compile_kwargs = {}
+    compile_kwargs.setdefault("allow_input_downcast", True)
+    compile_kwargs.setdefault("accept_inplace", True)
 
-    sampler_fn = compile_pymc(
-        inputs,
+    sampler_fn = compile_forward_sampling_function(
         vars_to_sample,
-        allow_input_downcast=True,
-        accept_inplace=True,
+        vars_in_trace=[],
+        basic_rvs=model.basic_RVs,
+        givens_dict=None,
         **compile_kwargs,
     )
 
--- a/pymc/tests/test_sampling.py
+++ b/pymc/tests/test_sampling.py
@@ -27,6 +27,7 @@ import scipy.special
 import xarray as xr
 
 from aesara import Mode, shared
+from aesara.compile import SharedVariable
 from arviz import InferenceData
 from arviz import from_dict as az_from_dict
 from arviz.tests.helpers import check_multiple_attrs
@@ -38,6 +39,7 @@ from pymc.aesaraf import compile_pymc
 from pymc.backends.base import MultiTrace
 from pymc.backends.ndarray import NDArray
 from pymc.exceptions import IncorrectArgumentsError, SamplingError
+from pymc.sampling import compile_forward_sampling_function
 from pymc.tests.helpers import SeededTest, fast_unstable_sampling_mode
 from pymc.tests.models import simple_init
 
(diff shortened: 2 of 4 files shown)


ricardoV94:
`sample_posterior_predictive_w` is not refactored yet so we don't need to incorporate it yet

pymc/aesaraf.py
```diff
@@ -1015,3 +1019,144 @@ def compile_pymc(
...
+            value = givens_dict[node]
+            if not isinstance(value, (Apply, Variable)):
+                # If the value is not a tensor, it is wrapped as a Constant with the node's name
+                value = at.constant(
+                    givens_dict[node],
+                    dtype=node.dtype,
+                    name=node.name,
+                )
+            givens.append((node, value))
+            registered_givens.add(node)
```
ricardoV94: This does not seem like it needs to be in this loop, it has nothing to do with volatility inference

pymc/aesaraf.py
```diff
@@ -1015,3 +1019,144 @@ def compile_pymc(
...
+                    clients = node.outputs
+                else:
+                    raise
+            for client in clients:
+                if (
+                    isinstance(client, aesara.graph.basic.Variable)
+                    and client.owner is not None
+                    and isinstance(client.owner.op, at.random.basic.RandomVariable)
+                    and any(rv_param in volatile_nodes for rv_param in client.owner.inputs[3:])
+                ):
```
ricardoV94: size which is in `inputs[1]`, may be dynamic (i.e., set by a shared variable or another volatile node)

pymc/aesaraf.py
```diff
@@ -1015,3 +1019,144 @@ def compile_pymc(
...
+                clients = [c for c, _ in fg.clients[node]]
+            except KeyError:
+                # Some Apply nodes don't show up in the clients of a FunctionGraph
+                # so we take the Apply node's outputs instead
+                if isinstance(node, aesara.graph.basic.Apply):
+                    clients = node.outputs
+                else:
+                    raise
+            for client in clients:
+                if (
```
ricardoV94: Can't we reduce these two `if` statements to simply checking if any of the node inputs is volatile?

pymc/aesaraf.py
```diff
@@ -1015,3 +1019,144 @@ def compile_pymc(
...
+
+    # Collect the function inputs by walking the graph from the outputs. Inputs will be:
+    # 1. Random variables that are not volatile
+    # 2. Variables that have no owner and are not constant or shared
+    inputs = []
+
+    def expand(node):
+        if isinstance(node, Apply):
+            # Apply nodes are never inputs
+            return node.get_parents()
```
ricardoV94: Why are we including the `Apply`s in the expansion at all?

pymc/util.py
```diff
@@ -349,3 +350,25 @@ def check_dist_not_registered(dist, model=None):
                 f"You should use an unregistered (unnamed) distribution created via "
                 f"the `.dist()` API instead, such as:\n`dist=pm.Normal.dist(0, 1)`"
             )
+
+
+def point_wrapper(core_function):
```
ricardoV94, line 355: Can we reuse `Point` and `PointFunc`?

https://github.com/pymc-devs/pymc/blob/9a905faa8799f3d4e666603feb7e0e5ba8a8007f/pymc/model.py#L1791-L1822

pymc/aesaraf.py
```diff
@@ -1015,3 +1019,144 @@ def compile_pymc(
         **kwargs,
     )
     return aesara_function
+
+
+def compile_forward_sampling_function(
```
ricardoV94: I think this should be in `sampling.py`

pymc/util.py
ricardoV94: Or improve those, as it seems the main goal is similar

pymc/tests/test_aesaraf.py
```diff
@@ -704,3 +706,126 @@ def update(self, node):
...
+            pm.Normal("obs", nested_mu, sigma, observed=y)
+
+        f = compile_forward_sampling_function(
+            outputs=model.observed_RVs,
+            vars_in_trace=[nested_mu, sigma],  # mu isn't in the trace and will be deemed volatile
+        )
+        assert {i.name for i in self.get_function_inputs(f)} == {"sigma"}
+        assert {i.name for i in self.get_function_roots(f)} == {"sigma"}
+
+        aesara.dprint(model.observed_RVs[0])
```
ricardoV94:
```suggestion
```

pymc/tests/test_aesaraf.py
```diff
@@ -704,3 +706,126 @@ def update(self, node):
...
+            "x",
+            "p",
+            "category",
+            "beta",
+            "sigma",
+        }
+
+    def test_volatile_parameters(self):
+        rng_state = np.random.RandomState(42)
+        rng = np.random.default_rng(42)
```
ricardoV94: No need for these seeds as we are not sampling anywhere in these test

pymc/aesaraf.py
lucianopaz: Ok, so it should be enough to only look at `inputs[1:]`, right? Because the first input will always be volatile, since it's a shared random stream.

pymc/aesaraf.py
lucianopaz: Yeah, I'll take it out

pymc/tests/test_aesaraf.py
```diff
@@ -704,3 +706,126 @@ def update(self, node):
         fn = compile_pymc(inputs=[], outputs=dummy_x)
         assert fn() == 2.0
         assert fn() == 3.0
+
+
+class TestCompileForwardSampler:
```
ricardoV94: We need tests with variables created by `SymbolicDistributions` like `pm.Censored` and `pm.Mixture`, which do not return pure `RandomVariables` but should be treated as such for the sake of forward sampling.

pymc/aesaraf.py
lucianopaz: What do you mean?

pymc/aesaraf.py
ricardoV94: Any preference between `forward_sampling` or `ancestral_sampling`?

pymc/util.py
lucianopaz: I think that the approach I used for the inputs could be copied into `PointFunc` so that it can also handle `*args`, but the main goal of the `point_wrapper` is to only take the values from a `point` which are valid inputs to the core function and ignore the rest

pymc/aesaraf.py
lucianopaz: Why?

pymc/tests/test_aesaraf.py
lucianopaz: Oh damn

pymc/tests/test_aesaraf.py
lucianopaz: Ok

pymc/aesaraf.py
ricardoV94: I think we might want to just ignore Shared RandomState/Generator Variables altogether. Some RandomVariable-like `Op`s created with `OpFromGraph` will include those as the last input instead of the first (e.g., the AR distribution has to do this because of a Scan limitation)

pymc/aesaraf.py
lucianopaz: Forward sampling all the way. Ancestral sampling seems like some black magic horoscope thing

pymc/aesaraf.py
ricardoV94: I don't get why we are getting `Apply` nodes in the expansion function. Is this something that `node.get_parents()` does? What about expanding with `return node.owner.inputs` instead?

pymc/aesaraf.py
ricardoV94: It's implementing a specific sampling algorithm, it's not just a Aesara helper

pymc/util.py
ricardoV94: Isn't that what `Point(filter_vars=True)`, is trying to do as well, except focused only on `value_vars`?

ricardoV94: But the point was indeed if we can combine the functionality, instead of keeping two separate things that are kind of trying to do the same

pymc/aesaraf.py
lucianopaz: Not necessarily. The random variables will always have at least 1 volatile input, the random stream shared generator. That's why I had to take special care when they are defined in the trace and don't have other volatile inputs. But maybe I didn't understand your question

pymc/aesaraf.py
lucianopaz: How should I approach that then? Filter the inputs that aren't random stream instances?

pymc/aesaraf.py
lucianopaz: I though that `get_parents` just gave the `owner` if the node was a `Variable` or the `inputs` if the node was an `Apply`

pymc/aesaraf.py
ricardoV94: I think so

pymc/aesaraf.py
ricardoV94: Seems like we should just ignore RandomState inputs everywhere?

pymc/util.py
lucianopaz: The `point_wrapper` is model agnostic, it just cares about the calling function's signature. The `Point` is agnostic of the calling function and only cares about the variables defined in a model. These two things are different, and I'm not sure how would be a simple way to merge them

pymc/aesaraf.py
ricardoV94: Yes I think that's what it does, but why do we need the owner of the Variable? That will always be the end of the graph in that branch

pymc/tests/test_aesaraf.py
lucianopaz: What should we test for here? Have a latent mixture that is volatile or not?

pymc/util.py
ricardoV94: I see Point used only in 1-2 places in the codebase, so it sounds like we can easily transfer the point logic to the `PointFunc` if that makes the two more equivalent.

ricardoV94: And one of the places used is in SMC for `prior_predictive_sampling` which would probably be addressed by the new changes implemented in the PR

pymc/tests/test_aesaraf.py
ricardoV94: I don't know what the test should be, but my concern is that Mixture and quite some other distributions take as inputs random variables that are not in the trace, but we don't want to resample Mixtures if they are in the trace, right?

This happens basically for every distribution that makes use of `ignore_logprob`: https://github.com/pymc-devs/pymc/blob/138a64f6f2bfedcd7940f8bca37518d3a72d9d18/pymc/distributions/logprob.py#L286

https://github.com/pymc-devs/pymc/blob/447120462eb2537ece5fa9548447a506a4d3ad28/pymc/distributions/mixture.py#L273

pymc/aesaraf.py
```diff
@@ -1015,3 +1019,130 @@ def compile_pymc(
...
+                inputs.append(node)
+        return node.get_parents()
+
+    # walk produces a generator, so we have to actually exhaust the generator in a list to walk
+    # the entire graph
+    list(walk(fg.outputs, expand))
+    givens = []
+
+    # Populate the givens list
+    givens = [(node, value) for node, value in givens_dict.items()]
```
ricardoV94: Can't we just pass givens to `compile_pymc` as one of the general kwargs?

pymc/tests/test_aesaraf.py
lucianopaz: This indeed is hard to fix. Is there a way for me to test if the RV was created by a symbolic distribution? Maybe I’ll have to add a new type like SymbolicRV and test against that

ricardoV94: Those RVs shouldn't be in model.basic_RVs so maybe you can use that

pymc/aesaraf.py
lucianopaz: What do you mean? Do you want to set `givens` as a keyword argument of `compile_pymc` itself? I think that we could, but we should do it in another PR

pymc/tests/test_aesaraf.py
lucianopaz: I didn't want to go check the `model` instance. I wanted to make the aesara variables have all the information necessary to make this work on their own, without having to resort to external book keeping. To make this happen, I had to go quite meta-python. I had to write a metaclass that wraps the `SymbolicDistribution.rv_op` classmethod. This was really tricky because classmethods are descriptors, so they can't be wrapped as easily as regular functions or methods. The good news is that I managed to figure out how to wrap the descriptor and add a simple boolean tag to the output variable of `rv_op`, so the code now checks if the clients are pure `RandomVariable`s or if they were created using a `SymbolicDistribution.rv_op`. We could actually use this new tag to fix the behavior point out in #5765.

pymc/aesaraf.py
lucianopaz: I don't think I followed you, don't we need to expand the full graph through the Variable owners?

lucianopaz:
@ricardoV94, this is ready for review now. We'll address the `sample_posterior_predictive_w` function in another PR

pymc/sampling.py
```diff
@@ -1718,38 +1877,22 @@ def sample_posterior_predictive(
...
-        and not isinstance(rv, (Constant, SharedVariable))
-        and rv.name in names_in_trace
-    ]
-    if inputs_and_names:
-        inputs, input_names = zip(*inputs_and_names)
-    else:
-        inputs, input_names = [], []
-
     if size is not None:
         vars_to_sample = [change_rv_size(v, size, expand=True) for v in vars_to_sample]
```
ricardoV94, line 1871: We should remove these lines, but doesn't need to happen in this PR

pymc/tests/test_sampling.py
```diff
@@ -1446,3 +1449,171 @@ def test_no_init_nuts_compound(caplog):
...
+        list(walk(function.maker.fgraph.outputs, expand))
+        return roots
+
+    @staticmethod
+    def get_function_inputs(function):
+        return {
+            i
+            for i, istore, *_ in function.indices
+            if istore is None and not isinstance(i.variable, SharedVariable)
+        }
```
ricardoV94: Does this work?

```suggestion
        return {
            i
            for i in function.maker.fgraph.inputs
            if not isinstance(i, SharedVariable)
        }
```

pymc/tests/test_sampling.py
```diff
@@ -1446,3 +1449,171 @@ def test_no_init_nuts_compound(caplog):
...
+            if (
+                isinstance(node, Variable)
+                and node.owner is None
+                and getattr(node, "name", None) is not None
+            ):
+                roots.append(node)
+            return node.get_parents()
+
+        list(walk(function.maker.fgraph.outputs, expand))
+        return roots
```
ricardoV94: Does something like this work?

```suggestion
    return [
        var 
        for var in aesara.graph.basic.graph_inputs(f.maker.fgraph.outputs))
        if var.name 
]
```

pymc/tests/test_sampling.py
```diff
@@ -1446,3 +1449,171 @@ def test_no_init_nuts_compound(caplog):
...
+    @staticmethod
+    def get_function_inputs(function):
+        return {
+            i
+            for i, istore, *_ in function.indices
+            if istore is None and not isinstance(i.variable, SharedVariable)
+        }
+
+    def test_linear_model(self):
+        rng = np.random.default_rng()
```
ricardoV94:
```suggestion
```

pymc/tests/test_sampling.py
```diff
@@ -1446,3 +1449,171 @@ def test_no_init_nuts_compound(caplog):
...
+            mu = pm.Deterministic("mu", alpha + beta * x)
+            sigma = pm.HalfNormal("sigma", 0.1)
+            obs = pm.Normal("obs", mu, sigma, observed=y)
+
+        f = compile_forward_sampling_function([obs], vars_in_trace=[alpha, beta, sigma, mu])
+        assert {i.name for i in self.get_function_inputs(f)} == {"alpha", "beta", "sigma", "mu"}
+        assert {i.name for i in self.get_function_roots(f)} == {"mu", "sigma"}
+
+    def test_nested_observed_model(self):
+        rng = np.random.default_rng()
```
ricardoV94:
```suggestion
```

pymc/tests/test_sampling.py
```diff
@@ -1446,3 +1449,171 @@ def test_no_init_nuts_compound(caplog):
...
+        assert {i.name for i in self.get_function_roots(f)} == {
+            "x",
+            "p",
+            "category",
+            "beta",
+            "sigma",
+        }
+
+    def test_volatile_parameters(self):
+        rng = np.random.default_rng()
```
ricardoV94:
```suggestion
```

pymc/tests/test_sampling.py
```diff
@@ -1446,3 +1449,171 @@ def test_no_init_nuts_compound(caplog):
...
+            vars_in_trace=[mu, nested_mu, sigma],
+            givens_dict={
+                mu: np.array(1.0)
+            },  # mu will be considered volatile because it's in givens
+        )
+        assert {i.name for i in self.get_function_inputs(f)} == {"sigma"}
+        assert {i.name for i in self.get_function_roots(f)} == {"mu", "sigma"}
+
+    def test_distributions_op_from_graph(self):
+        rng = np.random.default_rng()
```
ricardoV94:
```suggestion
```

pymc/tests/test_sampling.py
```diff
@@ -1446,3 +1449,171 @@ def test_no_init_nuts_compound(caplog):
...
+
+        f = compile_forward_sampling_function(
+            outputs=[obs],
+            vars_in_trace=[mu, w],
+        )
+        assert {i.name for i in self.get_function_inputs(f)} == {"w", "mu"}
+        assert {i.name for i in self.get_function_roots(f)} == {"w", "mu"}
+
+    def test_distributions_no_op_from_graph(self):
+        rng = np.random.default_rng()
```
ricardoV94:
```suggestion
```

pymc/util.py
```diff
@@ -349,3 +350,25 @@ def check_dist_not_registered(dist, model=None):
...
+    """Wrap an aesara compiled function to be able to ingest point dictionaries whilst
+    ignoring the keys that are not valid inputs to the core function.
+    """
+    ins = [
+        i.name
+        for i, istore, *_ in core_function.indices
+        if istore is None and not isinstance(i.variable, SharedVariable)
+    ]
+
+    def wrapped(*args, **kwargs):
```
ricardoV94: Do we need to allow for non keyword arguments?

pymc/tests/test_aesaraf.py
ricardoV94: This bookkeeping is precisely the one and only job of `Model`. My issue with adding tag information is that it makes graph manipulation using standard Aesara features more dangerous because we have to make sure to duplicate all the tag information that is important in the context of PyMC in all rewrites, so that other routines will work properly with the optimized graphs.

pymc/distributions/distribution.py
```diff
@@ -363,7 +363,46 @@ def dist(
         return rv_out
 
 
-class SymbolicDistribution:
+class RvOpTagger:
```
ricardoV94: I don't think this extra code complexity is more reasonable than just checking if a variable shows up in `model.basic_RVs`. The sampling / logp functions are all supposed to be created within a `Model` context anyway.

ricardoV94:
Looks good, just trying to push to simplify code as much as possible. Left some comments above @lucianopaz

pymc/distributions/distribution.py
ricardoV94: By the way, I am not saying we need to have access to the model in the new function, we can just pass `model.basic_RVs` as another argument, similar to how `outputs` comes from the model anyway

ricardoV94: It might also simplify some of the isinstance(...) checks in the `volatile` loop

pymc/sampling.py
```diff
@@ -1534,6 +1542,157 @@ def stop_tuning(step):
...
+            or (
+                node.owner is not None and isinstance(node.owner.op, RandomVariable)
+            )  # Random variables
+            or (node in vars_in_trace)  # Variables in the trace
+        ):
+            if node not in volatile_nodes:
+                # This test will include variables without owners, and that are not constant
+                # or shared, because these nodes will never be considered volatile
+                inputs.append(node)
+        return node.get_parents()
```
ricardoV94: This seems to pass all tests locally:
```suggestion
    def expand(node):
        if (
            (
                (
                    node.owner is None
                    and not isinstance(node, (Constant, SharedVariable))
                )  # Variables without owners that are not constant or shared
                or node in vars_in_trace  # Variables in the trace
            )
            and node not in volatile_nodes
        ):
                # This test will include variables without owners, and that are not constant
                # or shared, because these nodes will never be considered volatile
                inputs.append(node)

        if node.owner:
            return node.owner.inputs
```

pymc/sampling.py
```diff
@@ -1534,6 +1542,157 @@ def stop_tuning(step):
...
+
+    # walk produces a generator, so we have to actually exhaust the generator in a list to walk
+    # the entire graph
+    list(walk(fg.outputs, expand))
+    givens = []
```
ricardoV94:
```suggestion
```

pymc/sampling.py
```diff
@@ -1534,6 +1542,157 @@ def stop_tuning(step):
...
+                    client in vars_in_trace
+                    and isinstance(client, Variable)
+                    and client.owner is not None
+                    and (
+                        isinstance(client.owner.op, RandomVariable)
+                        or getattr(client.tag, "symbolic_distribution", False)
+                    )
+                ):
+                    # Other variables that aren't random variables in the trace are volatile
+                    volatile_nodes.add(client)
```
ricardoV94: This passes all the local tests:

```suggestion
    # Walk the graph from inputs to outputs and tag the volatile variables
    nodes: List[Union[Variable, Apply]] = general_toposort(
        fg.outputs, deps=lambda x: x.owner.inputs if x.owner else []
    )
    volatile_nodes: Set[Any] = set()
    for node in nodes:
        if isinstance(node, SharedVariable) or node in fg.outputs or node in givens_dict:
            volatile_nodes.add(node)
        is_volatile = node in volatile_nodes

        # Propagate the volatile property to the node's clients
        if is_volatile:
            clients = [
                out
                for c, _ in fg.clients[node] if c!= "output"
                for out in c.outputs
            ]
            for client in clients:
                if (
                    isinstance(client.owner.op, RandomVariable)
                    and any(
                        rv_param in volatile_nodes
                        for rv_param in client.owner.inputs
                        if not isinstance(
                            rv_param, (RandomStateSharedVariable, RandomGeneratorSharedVariable)
                        )
                    )
                ):
                    # Random variables that have volatile parameters are also volatile
                    volatile_nodes.add(client)
                elif not (
                    client in vars_in_trace
                    and (
                        isinstance(client.owner.op, RandomVariable)
                        or getattr(client.tag, "symbolic_distribution", False)
                    )
                ):
                    # Other variables that aren't random variables in the trace are volatile
                    volatile_nodes.add(client)
```

ricardoV94 (commented):
Tested some simplifications locally, do you spot any issues?

pymc/tests/test_sampling.py
```diff
@@ -1446,3 +1449,171 @@ def test_no_init_nuts_compound(caplog):
         b = pm.Poisson("b", 1)
         pm.sample(10, tune=10)
         assert "Initializing NUTS" not in caplog.text
+
+
+class TestCompileForwardSampler:
```
ricardoV94, line 1453: We don't seem to have any test for when an unobserved RV depends on MutableData. What is supposed to happen in that case?

```python
with pm.Model() as m:
  prior_mean = pm.MutableData("prior_means", np.arange(5))
  x = pm.Normal("x", mu=prior_means)
  y = pm.Normal("y", mu=x, observed=np.arange(5))
```

pymc/aesaraf.py
ricardoV94: I added a change suggestion below that explains why we don't need to care about applys in the graph expansion

pymc/aesaraf.py
ricardoV94: It was a question of just using givens directly, instead of this intermediate `givens_dict` variable, but I see the point of it now.

pymc/tests/test_sampling.py
lucianopaz: Yeah, that should work. I originally had tried to do something similar but used `fn` instead of `maker`, and that failed for some compile modes. I'll switch to your suggestion, which is cleaner.

pymc/tests/test_sampling.py
lucianopaz: The `x` random variable will have a volatile parameter, and will always be interpreted as a volatile variable. That case is covered by `test_volatile_parameters` where we make a parameter volatile or not depending on whether it's included in the trace or not.

pymc/distributions/distribution.py
lucianopaz: Your argument about the rewrites convinced me. I'll take `model.basic_RVs` and pass into the compilation function.

ricardoV94: This would also simplify the `is_volatile` loop I think

pymc/sampling.py
```diff
@@ -1534,6 +1542,166 @@ def stop_tuning(step):
...
+        (
+            node,
+            value
+            if isinstance(value, (Variable, Apply))
+            else at.constant(value, dtype=getattr(node, "dtype", None), name=node.name),
+        )
+        for node, value in givens_dict.items()
+    ]
+
+    return compile_pymc(inputs, fg.outputs, givens=givens, on_unused_input="ignore", **kwargs)
```
ricardoV94, line 1702: @lucianopaz what about this? It's very similar but it is more strict about the sources of `volatility`, and after that any node that has a volatile input is considered volatile, regardless of what it is.

The reason I went to look for an alternative is the case where `dist`s depend on other standard RVs. With this code I think the following model would work as expected:

```python
with pm.Model() as m:
  latent_mu = pm.Normal("latent_mu")
  base_dist = pm.Normal.dist(latent_mu, 1)
  llike = pm.Censored("llike", base_dist, lower=-1, upper=1, observed=data)
```

```suggestion
    if givens_dict is None:
        givens_dict = {}

    if basic_rvs is None:
        basic_rvs = []

    # We need a function graph to walk the clients and propagate the volatile property
    fg = FunctionGraph(outputs=outputs, clone=False)

    # Walk the graph from inputs to outputs and tag the volatile variables
    nodes: List[Variable] = general_toposort(
        fg.outputs, deps=lambda x: x.owner.inputs if x.owner else []
    )
    volatile_nodes: Set[Any] = set()
    for node in nodes:
        if (
            # SharedVariables, except RandomState/Generators
            (isinstance(node, SharedVariable) and not isinstance(node, (RandomStateSharedVariable, RandomGeneratorSharedVariable)))
            # Basic RVs that are not in the trace
            or (node.owner and isinstance(node.owner.op, RandomVariable) and node not in vars_in_trace and node in basic_rvs)
            # Variables that have any volatile input
            or (node.owner and any(inp in volatile_nodes for inp in node.owner.inputs))
            or node in fg.outputs
            or node in givens_dict
        ):
            volatile_nodes.add(node)

    # Collect the function inputs by walking the graph from the outputs. Inputs will be:
    # 1. Random variables that are not volatile
    # 2. Variables that have no owner and are not constant or shared
    inputs = []
    def expand(node):
        if (
            (
                node.owner is None and not isinstance(node, (Constant, SharedVariable))
            )  # Variables without owners that are not constant or shared
            or node in vars_in_trace  # Variables in the trace
        ) and node not in volatile_nodes:
            # This test will include variables without owners, and that are not constant
            # or shared, because these nodes will never be considered volatile
            inputs.append(node)
        if node.owner:
            return node.owner.inputs

    # walk produces a generator, so we have to actually exhaust the generator in a list to walk
    # the entire graph
    list(walk(fg.outputs, expand))

    # Populate the givens list
    givens = [
        (
            node,
            value
            if isinstance(value, (Variable, Apply))
            else at.constant(value, dtype=getattr(node, "dtype", None), name=node.name),
        )
        for node, value in givens_dict.items()
    ]

    return compile_pymc(inputs, fg.outputs, givens=givens, on_unused_input="ignore", **kwargs)
```

pymc/sampling.py
```diff
@@ -1534,6 +1542,166 @@ def stop_tuning(step):
...
+            for client in clients:
+                if is_a_random_variable(client) and any(
+                    rv_param in volatile_nodes
+                    for rv_param in distribution_not_dist_parameters(client)
+                ):
+                    # Random variables that have volatile parameters are also volatile
+                    volatile_nodes.add(client)
+                elif not (client in vars_in_trace and is_a_random_variable(client)):
+                    # Other variables that aren't random variables in the trace are volatile
+                    volatile_nodes.add(client)
```
ricardoV94: @lucianopaz what about this? It's very similar but it is more strict about the sources of `volatility`, and after that any node that has a volatile input is considered volatile, regardless of what it is.

The reason I went to look for an alternative is the case where `dist`s depend on other standard RVs. With this code I think the following model would work as expected:

```python
with pm.Model() as m:
  latent_mu = pm.Normal("latent_mu")
  base_dist = pm.Normal.dist(latent_mu, 1)
  censored = pm.Censored("censored", base_dist, lower=-1, upper=1)
```
That is, if `latent_mu` is in the trace `censored` will not be resampled by default, but if it is missing it will, as `base_dist` will then be a volatile input. Do we have a test for this case?

```suggestion
    # We need a function graph to walk the clients and propagate the volatile property
    fg = FunctionGraph(outputs=outputs, clone=False)

    # Walk the graph from inputs to outputs and tag the volatile variables
    nodes: List[Variable] = general_toposort(
        fg.outputs, deps=lambda x: x.owner.inputs if x.owner else []
    )
    volatile_nodes: Set[Any] = set()
    for node in nodes:
        if (
            # SharedVariables, except RandomState/Generators
            (isinstance(node, SharedVariable) and not isinstance(node, (RandomStateSharedVariable, RandomGeneratorSharedVariable)))
            # Basic RVs that are not in the trace
            or (node.owner and isinstance(node.owner.op, RandomVariable) and node not in vars_in_trace and node in basic_rvs)
            # Variables that have any volatile input
            or (node.owner and any(inp in volatile_nodes for inp in node.owner.inputs))
            or node in fg.outputs
            or node in givens_dict
        ):
            volatile_nodes.add(node)

```

pymc/tests/test_sampling.py
```diff
@@ -1446,3 +1448,207 @@ def test_no_init_nuts_compound(caplog):
...
+
+            obs = pm.Normal("obs", mu, 1, observed=np.ones((10, 3)))
+
+        f = compile_forward_sampling_function(
+            outputs=[obs],
+            vars_in_trace=[mu],
+            basic_rvs=model.basic_RVs,
+        )
+        assert {i.name for i in self.get_function_inputs(f)} == {"mu"}
+        assert {i.name for i in self.get_function_roots(f)} == {"mu"}
```
ricardoV94: Here is the test that covers the suggested changes above:

```suggestion
        with pm.Model() as model:
            latent_mu = pm.Normal("latent_mu", mu=np.arange(3))
            mu = pm.Censored("mu", pm.Normal.dist(latent_mu, sigma=1), lower=-1, upper=1)
            obs = pm.Normal("obs", mu, 1, observed=np.ones((10, 3)))

        f = compile_forward_sampling_function(
            outputs=[obs],
            vars_in_trace=[latent_mu, mu],
            basic_rvs=model.basic_RVs,
        )
        assert {i.name for i in self.get_function_inputs(f)} == {"latent_mu", "mu"}
        assert {i.name for i in self.get_function_roots(f)} == {"mu"}

        f = compile_forward_sampling_function(
            outputs=[obs],
            vars_in_trace=[mu],
            basic_rvs=model.basic_RVs,
        )
        assert {i.name for i in self.get_function_inputs(f)} == set()
        assert {i.name for i in self.get_function_roots(f)} == set()
```

pymc/util.py
ricardoV94: @lucianopaz ^ ?

lucianopaz: Not for this, at least. I'll take it out

ricardoV94 (approved):
Awesome @lucianopaz! This was actually one of the trickiest blockers for V4

Issue #5760, opened 2022-05-13
problem with set_data with custom coords


TimOliverMaier:
When I use custom coords for a model, resetting data in a shared variable fails.

**Please provide a minimal, self-contained, and reproducible example.**
```python
import pymc as pm
import numpy as np

coords = {"category":["A","B","C"],"data_point":list(range(1,101))}
x = np.random.normal([13,14,12],[2,1,3],size=(100,3))
x_new = x = np.random.normal([14,15,11],[2,1,3],size=(100,3))

with pm.Model(coords=coords) as model:
    x = pm.MutableData("x",x,dims=["data_point","country"])
    mu = pm.Normal("mu",mu=[12,12,12],sigma=2)
    sigma = pm.Exponential("sigma",lam=[1,1,1])
    obs = pm.Normal("obs",mu=mu,sigma=sigma,observed=x)
with model:
    pm.set_data({"x":x_new},coords=coords)
```

**Please provide the full traceback.**

<details><summary></summary>

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/home/tim/Master/prototyping/TutorialsAndArchive/test_stuff.ipynb Cell 1' in <cell line: 13>()
     [12](vscode-notebook-cell:/home/tim/Master/prototyping/TutorialsAndArchive/test_stuff.ipynb#ch0000000?line=11)     obs = pm.Normal("obs",mu=mu,sigma=sigma,observed=x)
     [13](vscode-notebook-cell:/home/tim/Master/prototyping/TutorialsAndArchive/test_stuff.ipynb#ch0000000?line=12) with model:
---> [14](vscode-notebook-cell:/home/tim/Master/prototyping/TutorialsAndArchive/test_stuff.ipynb#ch0000000?line=13)     pm.set_data({"x":x_new},coords=coords)

File ~/Master/prototyping/repos/pymc/pymc/model.py:1770, in set_data(new_data, model, coords)
   [1767](file:///home/tim/Master/prototyping/repos/pymc/pymc/model.py?line=1766) model = modelcontext(model)
   [1769](file:///home/tim/Master/prototyping/repos/pymc/pymc/model.py?line=1768) for variable_name, new_value in new_data.items():
-> [1770](file:///home/tim/Master/prototyping/repos/pymc/pymc/model.py?line=1769)     model.set_data(variable_name, new_value, coords=coords)

[...]

Issue #5751, opened 2022-05-06
Fix dim lengths created from `coords` or `ConstantData`


michaelosthege:
In [this comment](https://github.com/pymc-devs/pymc/issues/5613#issuecomment-1107752199) we proposed to make dimensions created from `coords` non-resizeable, because it would prevent users from running into issues like #5181.

## What needs to be done
In https://github.com/pymc-devs/pymc/blob/dd23c9008302fc184298076669d0a3a332aa59f0/pymc/model.py#L1070-L1076 the `set_coord` method could take another kwarg `resizable: bool=False` or `fixed: bool=True`.

Then we can actively turn off re-sizing of dimensions that are created from `ConstantData` or `coords`.

ricardoV94:
Let's remember to add #5181 as a test

LukeLB:
Hi guys just been looking at how this can be implemented. Can I check my understanding first... So this line:
```
self._dim_lengths[name] = length or aesara.shared(len(values))
```
Is causing the problem due to the `aesara.shared(len(values))` being resizeable. So the sugestion here is to create a new kwarg which we can use in an if statement, something like:
```
if fixed:
    self._dim_lengths[name] = length or aesara.tensor.constant(len(values))
else: 
    self._dim_lengths[name] = length or aesara.shared(len(values))
```

michaelosthege:
@LukeLB yes, that's the idea

canyon289:
@LukeLB Did you want to take a shot at fixing this, and if so when were you planning on doing so? I ask as this is holding up the Bambi migration to V4 so we want to get fixed soon here!

LukeLB:
@canyon289 I've got time today so I'll have a crack at it.

canyon289:
Sounds great! Looking forward to the PR and thank you. Let us know if you need any help.

LukeLB:
@canyon289 just put in a WIP-PR as I'm heading out and was having some problems with locally failing tests.

PR #5761, opened 2022-05-13
check for original coords

TimOliverMaier:
This is my quick fix for #5760  . When coords are set manually, `length_tensor` in `set_data` has no owner. That's why it is crashing. In my fix I only check if `original_coords` is not None, meaning they were set explicitly (as far as I understand). However, I am pretty unsure if that's the way to go. Especially since there were a couple of issues in the last time concering `coords`...

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -46,7 +46,7 @@ from aesara.graph.fg import FunctionGraph
 from aesara.tensor.random.opt import local_subtensor_rv_lift
 from aesara.tensor.random.var import RandomStateSharedVariable
 from aesara.tensor.sharedvar import ScalarSharedVariable
-from aesara.tensor.var import TensorVariable
+from aesara.tensor.var import TensorVariable, TensorConstant
 
 from pymc.aesaraf import (
     compile_pymc,
@@ -61,7 +61,7 @@ from pymc.data import GenTensorVariable, Minibatch
 from pymc.distributions import joint_logpt
 from pymc.distributions.logprob import _get_scaling
 from pymc.distributions.transforms import _default_transform
-from pymc.exceptions import ImputationWarning, SamplingError, ShapeError
+from pymc.exceptions import ImputationWarning, ShapeWarning, SamplingError, ShapeError
 from pymc.initial_point import make_initial_point_fn
 from pymc.math import flatten_list
 from pymc.util import (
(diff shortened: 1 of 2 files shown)


TimOliverMaier:
#5760

pymc/model.py
```diff
@@ -1179,16 +1179,17 @@ def set_data(
...
-            if not isinstance(length_belongs_to, SharedVariable) and length_changed:
-                raise ShapeError(
-                    f"Resizing dimension '{dname}' with values of length {new_length} would lead to incompatibilities, "
-                    f"because the dimension was initialized from '{length_belongs_to}' which is not a shared variable. "
-                    f"Check if the dimension was defined implicitly before the shared variable '{name}' was created, "
-                    f"for example by a model variable.",
-                    actual=new_length,
-                    expected=old_length,
-                )
+            if original_coords is None:
```
canyon289: Add a comment to explain why this was needed, otherwise looks good

canyon289:
Is there a way to write a test for this to ensure whatever bug is being fixed is fixed?

michaelosthege:
In light of https://github.com/pymc-devs/pymc/issues/5751 we'll have to do this slightly different.

Something like this:

1. `if length_changed`
    1. `if `length_tensor.owner.inputs[0].owner is None` → Warning: You're changing the shape of a shared variable in the '{dname}' dimension which was initialized from coords. Make sure to update the corresponding coords, otherwise you'll get shape issues.
    2. `if not isinstance(length_belongs_to, SharedVariable)` → ShapeError: You're trying to resize the '{dname}' dimension which is linked to {length_belongs_to} which can not be resized.

The i) case happens with the example from #5760.

The ii) case happens with something like this:
```python
with pm.Model():
    a = pm.Normal("a", mu=[1,2,3], dims="fixed")
    m = pm.MutableData("m", [[1,2,3]], dims=("one", "fixed"))

    # This is fine because the "fixed" dim is not resized
    pm.set_data({"m": [[1,2,3], [3,4,5]]})

    # Can't work because the "fixed" dimension is linked to a constant shape:
    # Note that the new data tries to change both dimensions
    pm.set_data({"m": [[1,2], [3,4]]})
```

@TimOliverMaier you could make that change here, ideally with a test case.

It is also relevant for https://github.com/pymc-devs/pymc/issues/5751 and AFAIK @LukeLB is working on that at the moment.

TimOliverMaier:
Thanks for the feedback I will look into this today. Also keeping an eye on #5751

TimOliverMaier:
I rearranged the if-statements a little to cluster those relevant for the `length_changed = True` case. Furthermore the code now test if `length_tensor` has an owner. If not, it assumes the coords were set customly and logs a warning if the length was changed. I looked into the PR from @LukeLB addressing issue #5751 and added an if statement checking if dimension length is stored in TensorConstant. 
Now I am looking for the best place for tests. Does anybody have a suggestion here?

TimOliverMaier:
I added a test only to assert that it is possible to set new data with identical dimensions.  I think additional tests should wait until a possible merge with #5763.

In `models` `set_data` stored coords differently from `add_coord`. I changed the line in `set_data` to store coords as tuple as well.

pymc/model.py
```diff
@@ -1179,23 +1179,48 @@ def set_data(
...
+                        f"to change the dimension length, 'fixed' in "
+                        f"'model.add_coord' must be passed False."
                     )
+                if length_tensor.owner is None:
+                    # This is the case if the dimension was initialized
+                    # from custom coords, but dimension length was not
+                    # stored in TensorConstant e.g by 'fixed' set to False
+
+                    warnings.warn(
+                        f"You're changing the shape of a shared variable "
```
```suggestion
                        f"You're changing the shape of a variable "
```

pymc/model.py
```diff
@@ -1179,23 +1179,48 @@ def set_data(
...
-                        f"The '{name}' variable already had {len(original_coords)} coord values defined for"
-                        f"its {dname} dimension. With the new values this dimension changes to length "
-                        f"{new_length}, so new coord values for the {dname} dimension are required."
+            if length_changed:
+                if isinstance(length_tensor, TensorConstant):
+                    raise ShapeError(
+                        f"Resizing dimension '{dname}' is impossible, because "
+                        f"a 'TensorConstant' stores its length. To be able "
+                        f"to change the dimension length, 'fixed' in "
+                        f"'model.add_coord' must be passed False."
```
```suggestion
                        f"'model.add_coord' must be set to `False`."
```

pymc/model.py
```diff
@@ -1179,23 +1179,48 @@ def set_data(
...
+                        ShapeWarning,
+                    )
+                else:
+                    length_belongs_to = length_tensor.owner.inputs[0].owner.inputs[0]
+                    if not isinstance(length_belongs_to, SharedVariable):
+                        raise ShapeError(
+                            f"Resizing dimension '{dname}' with values of length {new_length} would lead to incompatibilities, "
+                            f"because the dimension was initialized from '{length_belongs_to}' which is not a shared variable. "
+                            f"Check if the dimension was defined implicitly before the shared variable '{name}' was created, "
+                            f"for example by a model variable.",
```
```suggestion
                            f"for example by another model variable.",
```

pymc/model.py
```diff
@@ -1179,23 +1179,48 @@ def set_data(
...
+                            actual=new_length,
+                            expected=old_length,
+                        )
+                if original_coords is not None:
+                    if new_coords is None:
+                        raise ValueError(
+                            f"The '{name}' variable already had {len(original_coords)} coord values defined for"
```
```suggestion
                            f"The '{name}' variable already had {len(original_coords)} coord values defined for "
```

michaelosthege (approved):
I really like the informative messages!

Just a few rephrasing & typo changes... I'll apply them and merge it and then @LukeLB can close the coverage gaps after rebasing #5763.

Issue #5181, opened 2021-11-13
RV size is not correctly inferred from dims


lucianopaz:
## Description of your problem

While writing a hierarchical model, I ran into a shape problem. Minimally, what I want to do looks like this:

```python
>>> with pm.Model(coords={"feature": [1], "group": ["A", "B"]}):
...     x = pm.Normal("x", 0, 1, dims="feature")
...     y = pm.Normal("y", x[..., None], 1, dims=("feature", "group"))
... y.shape.eval()
WARNING (aesara.tensor.basic_opt): Failed to infer_shape from Op normal_rv{0, (0, 0), floatX, True}.
Input shapes: [None, (TensorConstant{0},), (), (<TensorType(int64, scalar)>, TensorConstant{1}), ()]
Exception encountered during infer_shape: <class 'Exception'>
Exception message: sharedvar.value does not exist anymore. Use sharedvar.get_value() or sharedvar.set_value() instead.
Traceback: Traceback (most recent call last):
  File "/home/lpaz/repos/aesara/aesara/tensor/basic_opt.py", line 867, in get_node_infer_shape
    o_shapes = shape_infer(
  File "/home/lpaz/repos/aesara/aesara/tensor/random/op.py", line 299, in infer_shape
    shape = self._infer_shape(size, dist_params, param_shapes=param_shapes)
  File "/home/lpaz/repos/aesara/aesara/tensor/random/op.py", line 196, in _infer_shape
    param_shapes = params_broadcast_shapes(
  File "/home/lpaz/repos/aesara/aesara/tensor/random/utils.py", line 47, in params_broadcast_shapes
    rev_extra_dims = [
  File "/home/lpaz/repos/aesara/aesara/tensor/random/utils.py", line 48, in <listcomp>
    max_bcast(a, b)
  File "/home/lpaz/repos/aesara/aesara/tensor/random/utils.py", line 41, in max_bcast
    if getattr(x, "value", x) == 1:
  File "/home/lpaz/repos/aesara/aesara/compile/sharedvalue.py", line 211, in _value_get
    raise Exception(
Exception: sharedvar.value does not exist anymore. Use sharedvar.get_value() or sharedvar.set_value() instead.
array([1, 1])
```

So the shape of `y` is inferred incorrectly, and this led to problems downstream in the full model. If instead of using `dims`, I use `size`, everything works fine.

```python
>>> with pm.Model(coords={"feature": [1], "group": ["A", "B"]}):
[...]

Issue #5739, opened 2022-05-04
Flaky `test_sample_posterior_predictive_after_set_data_with_coords`


ricardoV94:
This test has failed quite some times recently, last time in #5738

https://github.com/pymc-devs/pymc/runs/6286089630?check_suite_focus=true

We should investigate to see if there's a bug or we need less strict checks

PR #5763, opened 2022-05-14
Fix dim lengths created from coords or ConstantData #5751 

### LINKED ISSUE #5751: Fix dim lengths created from `coords` or `ConstantData`

LukeLB:
Closes #5751. Fixes dimensions created by add_coord by making dimension length a TensorConstant. I have also added a test for this using some code from #5181. 

I'm setting as WIP as @canyon289 suggested this PR is required for other work to become unblocked and I'm having problems with some failing tests (which could just be due to me running things locally with an incorrect set-up?). So it's worth seeing if they fail on here and then look at fixes, which will be faster than me trying to figure things out on my own!

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -20,7 +20,7 @@ import urllib.request
 import warnings
 
 from copy import copy
-from typing import Any, Dict, List, Optional, Sequence, Union, cast
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
 
 import aesara
 import aesara.tensor as at
@@ -466,9 +466,15 @@ def align_minibatches(batches=None):
                 rng.seed()
 
 
-def determine_coords(model, value, dims: Optional[Sequence[str]] = None) -> Dict[str, Sequence]:
+def determine_coords(
+    model,
+    value,
+    dims: Optional[Sequence[Optional[str]]] = None,
+    coords: Optional[Dict[str, Sequence]] = None,
+) -> Tuple[Dict[str, Sequence], Sequence[Optional[str]]]:
     """Determines coordinate values from data or the model (via ``dims``)."""
-    coords = {}
+    if coords is None:
+        coords = {}
 
     # If value is a df or a series, we interpret the index as coords:
     if hasattr(value, "index"):
@@ -499,10 +505,14 @@ def determine_coords(model, value, dims: Optional[Sequence[str]] = None) -> Dict
             )
         for size, dim in zip(value.shape, dims):
             coord = model.coords.get(dim, None)
-            if coord is None:
+            if coord is None and dim is not None:
                 coords[dim] = range(size)
 
-    return coords
+    if dims is None:
+        # TODO: Also determine dim names from the index
+        dims = [None] * np.ndim(value)
+
+    return coords, dims
 
 
 def ConstantData(
@@ -510,6 +520,7 @@ def ConstantData(
     value,
     *,
     dims: Optional[Sequence[str]] = None,
+    coords: Optional[Dict[str, Sequence]] = None,
     export_index_as_coords=False,
     **kwargs,
 ) -> TensorConstant:
@@ -522,6 +533,7 @@ def ConstantData(
         name,
         value,
         dims=dims,
+        coords=coords,
         export_index_as_coords=export_index_as_coords,
         mutable=False,
         **kwargs,
@@ -534,6 +546,7 @@ def MutableData(
     value,
     *,
     dims: Optional[Sequence[str]] = None,
+    coords: Optional[Dict[str, Sequence]] = None,
     export_index_as_coords=False,
     **kwargs,
 ) -> SharedVariable:
@@ -546,6 +559,7 @@ def MutableData(
         name,
         value,
         dims=dims,
+        coords=coords,
         export_index_as_coords=export_index_as_coords,
         mutable=True,
         **kwargs,
@@ -558,6 +572,7 @@ def Data(
     value,
     *,
     dims: Optional[Sequence[str]] = None,
+    coords: Optional[Dict[str, Sequence]] = None,
     export_index_as_coords=False,
     mutable: Optional[bool] = None,
     **kwargs,
@@ -588,9 +603,11 @@ def Data(
         :ref:`arviz:quickstart`.
         If this parameter is not specified, the random variables will not have dimension
         names.
+    coords : dict, optional
+        Coordinate values to set for new dimensions introduced by this ``Data`` variable.
     export_index_as_coords : bool, default=False
-        If True, the ``Data`` container will try to infer what the coordinates should be
-        if there is an index in ``value``.
+        If True, the ``Data`` container will try to infer what the coordinates
+        and dimension names should be if there is an index in ``value``.
     mutable : bool, optional
         Switches between creating a :class:`~aesara.compile.sharedvalue.SharedVariable`
         (``mutable=True``) vs. creating a :class:`~aesara.tensor.TensorConstant`
@@ -624,6 +641,9 @@ def Data(
     ...         model.set_data('data', data_vals)
     ...         idatas.append(pm.sample())
     """
+    if coords is None:
+        coords = {}
+
     if isinstance(value, list):
         value = np.array(value)
 
@@ -665,15 +685,27 @@ def Data(
             expected=x.ndim,
         )
 
-    coords = determine_coords(model, value, dims)
-
+    # Optionally infer coords and dims from the input value.
     if export_index_as_coords:
-        model.add_coords(coords)
-    elif dims:
+        coords, dims = determine_coords(model, value, dims)
+
+    if dims:
+        if not mutable:
+            # Use the dimension lengths from the before it was tensorified.
+            # These can still be tensors, but in many cases they are numeric.
+            xshape = np.shape(arr)
+        else:
+            xshape = x.shape
         # Register new dimension lengths
         for d, dname in enumerate(dims):
             if not dname in model.dim_lengths:
-                model.add_coord(dname, values=None, length=x.shape[d])
+                model.add_coord(
+                    name=dname,
+                    # Note: Coordinate values can't be taken from
+                    # the value, because it could be N-dimensional.
+                    values=coords.get(dname, None),
+                    length=xshape[d],
+                )
 
     model.add_random_variable(x, dims=dims)
 
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -704,6 +704,7 @@ def Data(
                     # Note: Coordinate values can't be taken from
                     # the value, because it could be N-dimensional.
                     values=coords.get(dname, None),
+                    mutable=mutable,
                     length=xshape[d],
                 )
 
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1071,6 +1071,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         self,
         name: str,
         values: Optional[Sequence] = None,
+        mutable: bool = False,
         *,
         length: Optional[Union[int, Variable]] = None,
     ):
@@ -1084,9 +1085,12 @@ class Model(WithMemoization, metaclass=ContextMeta):
         values : optional, array-like
             Coordinate values or ``None`` (for auto-numbering).
             If ``None`` is passed, a ``length`` must be specified.
+        mutable : bool
+            Whether the created dimension should be resizable.
+            Default is False.
         length : optional, scalar
             A scalar of the dimensions length.
-            Defaults to ``aesara.shared(len(values))``.
+            Defaults to ``aesara.tensor.constant(len(values))``.
         """
         if name in {"draw", "chain", "__sample__"}:
             raise ValueError(
@@ -1111,8 +1115,11 @@ class Model(WithMemoization, metaclass=ContextMeta):
             if not np.array_equal(values, self.coords[name]):
                 raise ValueError(f"Duplicate and incompatible coordinate: {name}.")
         else:
+            if mutable:
+                self._dim_lengths[name] = length or aesara.shared(len(values))
+            else:
+                self._dim_lengths[name] = length or aesara.tensor.constant(len(values))
             self._coords[name] = values
-            self._dim_lengths[name] = length or aesara.shared(len(values))
 
     def add_coords(
         self,
@@ -1192,9 +1199,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
                 if isinstance(length_tensor, TensorConstant):
                     raise ShapeError(
                         f"Resizing dimension '{dname}' is impossible, because "
-                        f"a 'TensorConstant' stores its length. To be able "
-                        f"to change the dimension length, 'fixed' in "
-                        f"'model.add_coord' must be set to `False`."
+                        "a 'TensorConstant' stores its length. To be able "
+                        "to change the dimension length, pass `mutable=True` when "
+                        "registering the dimension via `model.add_coord`, "
+                        "or define it via a `pm.MutableData` variable."
                     )
                 else:
                     length_belongs_to = length_tensor.owner.inputs[0].owner.inputs[0]
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1072,7 +1072,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         name: str,
         values: Optional[Sequence] = None,
         *,
-        length: Optional[Variable] = None,
+        length: Optional[Union[int, Variable]] = None,
     ):
         """Registers a dimension coordinate with the model.
 
@@ -1085,7 +1085,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
             Coordinate values or ``None`` (for auto-numbering).
             If ``None`` is passed, a ``length`` must be specified.
         length : optional, scalar
-            A symbolic scalar of the dimensions length.
+            A scalar of the dimensions length.
             Defaults to ``aesara.shared(len(values))``.
         """
         if name in {"draw", "chain", "__sample__"}:
@@ -1097,7 +1097,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
             raise ValueError(
                 f"Either `values` or `length` must be specified for the '{name}' dimension."
             )
-        if length is not None and not isinstance(length, Variable):
+        if isinstance(length, int):
+            length = at.constant(length)
+        elif length is not None and not isinstance(length, Variable):
             raise ValueError(
                 f"The `length` passed for the '{name}' coord must be an Aesara Variable or None."
             )
@@ -1116,7 +1118,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         self,
         coords: Dict[str, Optional[Sequence]],
         *,
-        lengths: Optional[Dict[str, Union[Variable, None]]] = None,
+        lengths: Optional[Dict[str, Optional[Union[int, Variable]]]] = None,
     ):
         """Vectorized version of ``Model.add_coord``."""
         if coords is None:
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -61,7 +61,7 @@ from pymc.data import GenTensorVariable, Minibatch
 from pymc.distributions import joint_logpt
 from pymc.distributions.logprob import _get_scaling
 from pymc.distributions.transforms import _default_transform
-from pymc.exceptions import ImputationWarning, SamplingError, ShapeError, ShapeWarning
+from pymc.exceptions import ImputationWarning, SamplingError, ShapeError
 from pymc.initial_point import make_initial_point_fn
 from pymc.math import flatten_list
 from pymc.util import (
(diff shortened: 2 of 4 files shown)


canyon289:
> it's worth seeing if they fail on here and then look at fixes, which will be faster than me trying to figure things out on my own

Great strategy. Thanks for taking such a principled approach to development

It seems that there type being changed somewhere. Might be as simple as updating this test.
https://github.com/pymc-devs/pymc/runs/6435716258?check_suite_focus=true#step:7:80

pymc/tests/test_model.py
```diff
@@ -702,6 +702,13 @@ def test_nested_model_coords():
     assert set(m2.RV_dims) < set(m1.RV_dims)
 
 
+def test_add_coord_fixed_dims():
+    with pm.Model(coords={"feature": [1], "group": ["A", "B"]}):
```
canyon289: Add a comment to this one since its not obvious from reading how fixed dims are being tested

LukeLB:
Failing test looks to be down to sampling error? 
```
test_sample_posterior_predictive_after_set_data_with_coord
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0.1
E       
E       Mismatched elements: 2 / 2 (100%)
E       Max absolute difference: 0.13669218
E       Max relative difference: 0.02331315
E        x: array([5, 6])
E        y: array([4.89599 , 5.863308]) 
```
And it's highlighted in an outstanding issue #5739.

pymc/model.py
```diff
@@ -1109,8 +1113,11 @@ def add_coord(
             if not np.array_equal(values, self.coords[name]):
                 raise ValueError(f"Duplicate and incompatible coordinate: {name}.")
         else:
+            if fixed:
+                self._dim_lengths[name] = length or aesara.tensor.constant(len(values))
+            else:
+                self._dim_lengths[name] = length or aesara.shared(len(values))
```
michaelosthege, line 1119: We should add a test to cover this

pymc/tests/test_model.py
```diff
@@ -702,6 +702,16 @@ def test_nested_model_coords():
...
 
 
+def test_add_coord_fixed_dims():
+    """
+    test that dims does not cause resizing of y. See #5181.
+    """
+    with pm.Model(coords={"feature": [1], "group": ["A", "B"]}):
+        x = pm.Normal("x", 0, 1, size=(1), dims="feature")
+        y = pm.Normal("y", x[..., None], 1, size=(1, 2), dims=("feature", "group"))
+    assert y.eval().shape == (1, 2)
```
michaelosthege:
```suggestion
        x = pm.Normal("x", 0, 1, dims="feature")
        y = pm.Normal("y", x[..., None], 1, dims=("feature", "group"))
    assert x.eval().shape == (1,)
    assert y.eval().shape == (1, 2)
```

LukeLB:
This test is failing because y is not the correct shape, `y.eval().shape == (1, 1)`. Both feature and group are TensorConstants so the addition to add_coord is doing as intented with respect to type changes. I'm not sure I understand why this is still happening.....
```{python}
def test_add_coord_fixed_dims():
    """
    test that dims does not cause resizing of y. See #5181.
    """
    with pm.Model(coords={"feature": [1], "group": ["A", "B"]}):
        x = pm.Normal("x", 0, 1, dims="feature")
        y = pm.Normal("y", x[..., None], 1, dims=("feature", "group"))
    assert x.eval().shape == (1,)
    assert y.eval().shape == (1, 2)
```

michaelosthege:
Maybe do more asserts?

```python
with pm.Model(coords={"feature": [1], "group": ["A", "B"]}):
    assert pmodel.dim_lengths["feature"].eval() == 1
    assert pmodel.dim_lengths["group"].eval() == 2
```

LukeLB:
My thinking for that test was to make sure that size and dims can be passed at the same time and we get the expected shape out which will not have been possible before this type change in `add_coord()`. So maybe its worth having two tests, the one you've suggested, and the other we can use the case you highlight https://github.com/pymc-devs/pymc/issues/5181#issuecomment-970573037 to test that size and dims can be passed at the same time.  What do you think?

michaelosthege:
When `size/shape` and `dims` are passed at the same time, the `size/shape` takes priority and `dims` are just tracked for labeling the dimensions in the final `xarray` objects.

That's why I suggested to _only_ pass dims, and actually it's a little concerning that the test doesn't pass.

Maybe call `add_coord` directly while passing `fixed`. For the variables depending on it there should be no difference between `aesara.shared()` or `at.constant()` dim lengths.

```python
with pm.Model() as pmodel:
    pmodel.add_coord("feature", [1], fixed=True/False)
    pmodel.add_coord("group", ["A", "B"], fixed=True/False)
    assert pmodel.dim_lengths["feature"].eval() == 1
    assert pmodel.dim_lengths["group"].eval() == 2
    x = pm.Normal("x", 0, 1, dims="feature")
    y = pm.Normal("y", x[..., None], 1, dims=("feature", "group"))
assert x.eval().shape == (1,)
assert y.eval().shape == (1, 2)
```

LukeLB:
When I try that all cases of fixed = True/False lead to y.eval().shape == (1, 1).

michaelosthege:
Oh, now I see why. The `x[..., None]` creates the RV as `(1, 1)` and because the dimensionality is as expected, there's no further resizing.

This works 👇 because then the RV is `(1,)` and repeated for the "group" dimension.

```python
    # ...
    x = pm.Normal("x", 0, 1, dims="feature")
    y = pm.Normal("y", x, 1, dims=("group", "feature"))
assert x.eval().shape == (1,)
assert y.eval().shape == (2, 1)
```

LukeLB:
I think I'm still misunderstanding something here. Wouldn't passing dims like that still be allowed prior to this PR? So I don't understand what it is testing with respect to this PR.

michaelosthege:
Yes, that's already allowed, but where this test comes in is in combination with the changes from #5761.

Actually, it's a little unfortunate that these changes are on separate branches.
Your test case should (latest after adding a `pm.set_data`) close the coverage gaps from that other PR.

I'm inclined to merge #5761 even with those coverage gaps (I'll review it really carefully right now).
After that you could just `git rebase origin/main`, add two lines doing `pm.set_data` and assert the `ShapeError` or `ShapeWarning` and then we should be good to go 🤔

michaelosthege:
@LukeLB the other PR was merged, so now you should be able to rebase and bring the coverage back up :)

LukeLB:
@michaelosthege OK I think I understand. I'll rebase and commit your suggested changes.

pymc/tests/test_model.py
```diff
@@ -702,6 +703,61 @@ def test_nested_model_coords():
...
+    """
+    with #5751 dimensions set by add_coord() are default TensorConstant type
+    and so have fixed dimensions. Trying to resize them will raise a ShapeError,
+    which this test checks for.
+    """
+    with pm.Model() as pmodel:
+        a = pm.Normal("a", mu=[1, 2, 3], dims="fixed")
+        m = pm.MutableData("m", [[1, 2, 3]], dims=("one", "fixed"))
+
+    with pytest.raises(ShapeError):
```
michaelosthege:
```suggestion
    with pytest.raises(ShapeError, match="was initialized from 'a'"):
```

pymc/tests/test_model.py
```diff
@@ -702,6 +703,61 @@ def test_nested_model_coords():
...
+    correct dimensions passed.
+    """
+    with pm.Model() as pmodel:
+        a = pm.Normal("a", mu=[1, 2, 3], dims="fixed")
+        m = pm.MutableData("m", [[1, 2, 3]], dims=("one", "fixed"))
+
+        # This is fine because the "fixed" dim is not resized
+        pm.set_data({"m": [[1, 2, 3], [3, 4, 5]]})
+
+
```
michaelosthege:
```suggestion
```
Let's include those two lines in the test above

pymc/tests/test_model.py
michaelosthege: Included from the test below, to reduce test runtime.
```suggestion
        m = pm.MutableData("m", [[1, 2, 3]], dims=("one", "fixed"))
        
        # This is fine because the "fixed" dim is not resized
        pm.set_data({"m": [[1, 2, 3], [3, 4, 5]]})

    with pytest.raises(ShapeError):
```

pymc/tests/test_model.py
```diff
@@ -702,6 +703,61 @@ def test_nested_model_coords():
...
+def test_add_coord_fixed_dims():
+    """
+    test that dims does not cause resizing of y. See #5181.
+    """
+    with pm.Model(coords={"feature": [1], "group": ["A", "B"]}):
+        x = pm.Normal("x", 0, 1, dims="feature")
+        y = pm.Normal("y", x, 1, dims=("group", "feature"))
+    assert x.eval().shape == (1,)
+    assert y.eval().shape == (2, 1)
```
michaelosthege, line 752: This is the test case that can be used to assert the `ShapeWarning`:

```python
with pytest.warns(ShapeWarning, match="was initialized from coords"):
    ...
```

michaelosthege (commented):
Can you add test cases for the `ShapeWarning` and `ValueError`?

With `match=""` one can make sure that the exception is the correct one. That's important because there are two different `ShapeError`s that one could run into..

We should aim to cover all code paths from the diff here https://github.com/pymc-devs/pymc/pull/5761/files

pymc/tests/test_model.py
LukeLB: This code doesn't raise a ShapeWarning. Should it?

LukeLB: Ah no worries, figured it out.

LukeLB: @michaelosthege I'm struggling to isolate the ShapeWarning, the code 
```python
with pm.Model() as pmodel:
    pmodel.add_coord("shared", [1,2,3], fixed=False)
    m = pm.MutableData("m", [1,2,3], dims=("shared"))

    pm.set_data({"m": [1,2,3,4]}) 
```
raises both ShapeWarning and a ValueError
```
ShapeWarning: You're changing the shape of a variable in the 'shared' dimension which was 
initialized from coords. Make sure to update the corresponding coords, otherwise you'll get shape issues.
  warnings.warn(

/var/folders/yw/hvmlf6sd7yv3040wn8str6540000gn/T/ipykernel_1859/831241587.py in <module>
      4     m = pm.MutableData("m", [1,2,3], dims=("shared"))
      5 
----> 6     pm.set_data({"m": [1,2,3,4]})

pymc/pymc/model.py in set_data(new_data, model, coords)
   1798 
   1799     for variable_name, new_value in new_data.items():
-> 1800         model.set_data(variable_name, new_value, coords=coords)
   1801 
   1802 

pymc/pymc/model.py in set_data(self, name, values, coords)
   1220                 if original_coords is not None:
   1221                     if new_coords is None:
-> 1222                         raise ValueError(
   1223                             f"The '{name}' variable already had {len(original_coords)} coord values defined for "
   1224                             f"its {dname} dimension. With the new values this dimension changes to length "

ValueError: The 'm' variable already had 3 coord values defined for its shared dimension. 
With the new values this dimension changes to length 4, so new coord values for the shared dimension are required.
```
I also tried the example from #5760,
```python
x = np.random.normal([13,14,12],[2,1,3],size=(100,3))
x_new = x = np.random.normal([14,15,11],[2,1,3],size=(100,3))

    model.add_coord("category", ["A","B","C"], fixed = False)
    model.add_coord("data_point", list(range(1,101)), fixed = False)
    x = pm.MutableData("x",x,dims=["data_point","country"])
    mu = pm.Normal("mu",mu=[12,12,12],sigma=2)
    sigma = pm.Exponential("sigma",lam=[1,1,1])
    obs = pm.Normal("obs",mu=mu,sigma=sigma,observed=x)
with model:
    pm.set_data({"x":x_new})
```
Which doesn't raise any errors or warnings. Not sure how to proceed here...

michaelosthege: The second example doesn't resize the `"data_point"` dimension, that's why it works without problems.

The traceback you posted above looks like `set_data` first warns, but then proceeds to raise a value error anyhow.

Obviously I overlooked this code path in #5761.
We should try to answer the following question:

__Is there a valid scenario where someone might want to resize a dimension that was initially created from `coords`, but can not provide new coordinate values in the same call to `set_data`?__

I feel like with the `ShapeWarning` we assume that there is such a scenario, so we should come up with a corresponding test.
But if in fact this scenario does not exist, we can just raise the `ValueError` directly.

LukeLB: @michaelosthege I would say we should raise the ValueError directly for this scenario. I can't imagine a situation wherein someone would both be using `add_coord()` directly and then not be able to use the coord they had previously passed to `add_coord()` in `set_data()`.

michaelosthege: Agreed. And let's make sure to have `match="..."` for all the error/warning asserts :)

pymc/tests/test_model.py
```diff
@@ -702,6 +703,78 @@ def test_nested_model_coords():
...
+    """
+    with pm.Model() as pmodel:
+        pmodel.add_coord("shared", [1, 2, 3], fixed=False)
+        m = pm.MutableData("m", [1, 2, 3], dims=("shared"))
+    with pytest.warns(ShapeWarning, match="was initialized from coords"):
+        with pmodel:
+            # correctly passes coords during resize and so warns user with ShapeWarning
+            pm.set_data({"m": [1, 2, 3, 4]}, coords={"shared": [1, 2, 3, 4]})
+
+
```
michaelosthege: The warning is now obsolete. In this test you were already passing new `coords` which the warning recommended to do..
```suggestion
```

pymc/tests/test_model.py
```diff
@@ -702,6 +703,78 @@ def test_nested_model_coords():
...
+def test_value_error_from_non_fixed_dims_no_coords():
+    """
+    test that resizing dims without passing coords raises a ValueError.
+    """
+    with pm.Model() as pmodel:
+        pmodel.add_coord("shared", [1, 2, 3], fixed=False)
+        m = pm.MutableData("m", [1, 2, 3], dims=("shared"))
+    with pytest.raises(ValueError, match="'m' variable already had 3"):
+        with pmodel:
```
michaelosthege: No need to close & open the modelcontext
```suggestion
        with pytest.raises(ValueError, match="'m' variable already had 3"):
```

michaelosthege (changes requested):
I can't make the change suggestion, but the whole block containing the `ShapeWarning` can be removed..

pymc/tests/test_model.py
LukeLB: My bad I should of read that Warning more carefully, I thought it was warning that there was a change in shape!

michaelosthege (approved):
Thanks @LukeLB !

This brings the coverage back up should lead to fewer shape problems long term.

pymc/tests/test_model.py
```diff
@@ -702,6 +703,64 @@ def test_nested_model_coords():
...
+
+def test_add_coord_fixed_dims():
+    """
+    test that dims does not cause resizing of y. See #5181.
+    """
+    with pm.Model(coords={"feature": [1], "group": ["A", "B"]}):
+        x = pm.Normal("x", 0, 1, dims="feature")
+        y = pm.Normal("y", x, 1, dims=("group", "feature"))
+    assert x.eval().shape == (1,)
+    assert y.eval().shape == (2, 1)
```
michaelosthege:
```suggestion
def test_coords_and_constantdata_create_fixed_dims():
    """
    When created from `pm.Model(coords=...)` or `pm.ConstantData`
    a dimension should be fixed in length.
    """
    with pm.Model(coords={"group": ["A", "B"]}) as m:
        x = pm.ConstantData("x", [0], dims="feature")
        y = pm.Normal("y", x, 1, dims=("group", "feature"))
    assert isinstance(m._dim_lengths["feature"], TensorConstant)
    assert isinstance(m._dim_lengths["group"], TensorConstant)
    assert x.eval().shape == (1,)
    assert y.eval().shape == (2, 1)
```

michaelosthege (changes requested):
ah damn, I just realized that we forgot to actually _use_ the new `fixed` kwarg.

I commented a change in one of the tests that should currently fail.
Let me know if you want me to take over - you put in a lot of effort already and I don't want to get demotivated by all this back and forth

LukeLB:
> ah damn, I just realized that we forgot to actually _use_ the new `fixed` kwarg.
> 
> I commented a change in one of the tests that should currently fail. Let me know if you want me to take over - you put in a lot of effort already and I don't want to get demotivated by all this back and forth

Not at all! The longer it goes the more I learn :) Thanks for the offer though.

pymc/tests/test_model.py
LukeLB: Why does the assert statement on line 759 fail? Its created from coord so should be a `TensorConstant` but is `Subtensor{int64}.0`.

LukeLB:
Why does the assert statement on line 758 fail? The `ConstantData()` function suggests that it returns a `TensorConstant`.

michaelosthege:
> Why does the assert statement on line 759 fail? Its created from coord so should be a `TensorConstant` but is `Subtensor{int64}.0`.

The `Subtensor{int64}.0` is because from `pm.ConstantData([1,2,3], dims="foo")` it's effectively doing this: `dim_lengths["foo"] = at.constant([1,2,3]).shape[0]`.

So what needs to be done is a change in the logic behind `pm.ConstantData` (or rather `pm.Data`) to determine the constant data's shape non-symbolically first, and then call `add_coord(..., fixed=True)`.


And I should add that `pm.MutableData([1,2,3], dims="bla")` becomes: `dim_lengths["bla"] = aesara.shared([1,2,3]).shape[0]`.

pymc/model.py
```diff
@@ -1084,9 +1085,12 @@ def add_coord(
         values : optional, array-like
             Coordinate values or ``None`` (for auto-numbering).
             If ``None`` is passed, a ``length`` must be specified.
+        fixed : bool
+            Whether dimensions created from coords should be fixed.
```
canyon289: FYI Fixed has could have a double meaning here, Fixed meaning immutable, or fixed meaning "fixed from broken. Suggest expanding the docstring. @michaelosthege let me know if this makes sense

```suggestion
            Whether dimensions created from coords should be fixed to an immutable value.
```

pymc/tests/test_model.py
```diff
@@ -702,6 +703,67 @@ def test_nested_model_coords():
     assert set(m2.RV_dims) < set(m1.RV_dims)
 
 
+def test_shape_error_from_fixed_dims():
+    """
+    with #5751 dimensions set by add_coord() are default TensorConstant type
+    and so have fixed dimensions. Trying to resize them will raise a ShapeError,
+    which this test checks for.
```
canyon289:
```suggestion
    which this what test checks for.
```

pymc/tests/test_model.py
```diff
@@ -702,6 +703,67 @@ def test_nested_model_coords():
     assert set(m2.RV_dims) < set(m1.RV_dims)
 
 
+def test_shape_error_from_fixed_dims():
+    """
+    with #5751 dimensions set by add_coord() are default TensorConstant type
```
canyon289: Nit: Change docstrings to match numpy style. Notably first line contains summary, and subsequent notes fit in a paragraph below.

https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

LukeLB:
OK so I think I have a solution which I have pushed but I'm running into a wall with a failing test which I can't seem to fix. Unfortunatley I won't be able to work on this till after the weekend so if there is a time pressure on this then it will be worth someone else picking it up.

michaelosthege:
Thanks @LukeLB, I will try to pick it up in the evening

canyon289:
@LukeLB this weekend is fine! You've made great progress on this and don't want to take away joy of getting it merged!

michaelosthege:
I rebased the branch and squashed the history into three logically separate commits:
1. ✔ Removing the ShapeWarning in favor of raising earlier
2. ✔ Adding optional `coords` to the `pm.Data` API, because we were taking coordinate values from the data variable, which does not work for >1 dimensions.
3. ✔ Addition of the `fixed` kwarg (the actual goal of this PR)

I will force-push them one by one so we can confirm that the first two commits don't create new problems.

michaelosthege:
@LukeLB @canyon289 @ricardoV94 please review and note the new changes I made in the `pm.Data` API

pymc/data.py
```diff
@@ -499,17 +505,22 @@ def determine_coords(model, value, dims: Optional[Sequence[str]] = None) -> Dict
             )
         for size, dim in zip(value.shape, dims):
             coord = model.coords.get(dim, None)
-            if coord is None:
+            if coord is None and dim is not None:
```
ricardoV94, line 508: I don't have the full context of this function but in case it's appropriate, can we also take advantage of static shape of Aesara\`s variables? Every entry in `tensorvariable.type.shape` should either be `None` or an integer for the size of that dimension.

pymc/data.py
```diff
@@ -665,15 +685,28 @@ def Data(
...
     if export_index_as_coords:
-        model.add_coords(coords)
-    elif dims:
+        coords, dims = determine_coords(model, value, dims)
+
+    if dims:
+        if not mutable:
+            # Use the dimension lengths from the before it was tensorified.
+            # These can still be tensors, but in many cases they are numeric.
+            xshape = np.shape(arr)
```
ricardoV94, line 696: doesn't matter much but `x.type.shape` should have the same info

pymc/data.py
```diff
@@ -665,15 +685,28 @@ def Data(
...
         # Register new dimension lengths
         for d, dname in enumerate(dims):
             if not dname in model.dim_lengths:
-                model.add_coord(dname, values=None, length=x.shape[d])
+                model.add_coord(
+                    name=dname,
+                    # Note: Coordinate values can't be taken from
+                    # the value, because it could be N-dimensional.
+                    values=coords.get(dname, None),
+                    fixed=not mutable,
```
ricardoV94: can't we use the same keyword so that we don't need to negate booleans?

ricardoV94 (approved):
Looks good to me, left a small question and a suggestion about using the same keyword internally (fixed vs mutable)

pymc/data.py
michaelosthege: The context of this function is to extract coords from data variales such as `pd.Series` or `pd.DataFrame` that have an index.
I extended the signature to also extract `dims`, because indexes can be named.

It would also be neat to consider `xarray` variables here, ebcause they are a 1:1 match for `pm.Data`.

The `and dim is not None` I added is because `dims` elements can be `None` and then the `coords[None]` doesn't make sense.

What you're talking about with the static shapes might be more relevant in the `pm.Data` function itself, where I created the local `xshape` variable.

pymc/data.py
canyon289: Id also prefer the same keyword throughout the codebase so its clear how userapi flows through the rest of the codebase

michaelosthege: Ja, you're right, of course.

Done.

michaelosthege (approved):
I feel 👍 about merging as long as the tests go ✔.

canyon289:
Me too!

canyon289:
Congrats @LukeLB! Thank you for leading this PR

LukeLB:
@canyon289 @michaelosthege @ricardoV94 thanks all!! It's been a pleasure :)

PR #5772, opened 2022-05-17
Remove size parameter from sample_posterior_predictive

### LINKED ISSUE #5771: Remove `size` option from `sample_posterior_predictive`
We should remove this option, as the call to `change_rv_size` will break any dependencies across resized variables.

https://github.com/pymc-devs/pymc/blob/85468b0cf9f174d9d0f8076ebe763f71dae8ec8f/pymc/sampling.py#L1740-L1741

ricardoV94:
Closes #5771 

Seems like this was mostly a leftover from V3 and is not really being used anywhere

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -58,7 +58,7 @@ from typing_extensions import TypeAlias
 
 import pymc as pm
 
-from pymc.aesaraf import change_rv_size, compile_pymc
+from pymc.aesaraf import compile_pymc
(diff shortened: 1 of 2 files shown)


ricardoV94:
Flaky test is back https://github.com/pymc-devs/pymc/issues/5739, opened a PR to fix it: https://github.com/pymc-devs/pymc/pull/5773

lucianopaz (approved):
LGTM

PR #5778, opened 2022-05-17
Add script for converting code from 3 to 4.

### LINKED ISSUE #4515: Write v4 upgrade guide

twiecki:
Closes https://github.com/pymc-devs/pymc/issues/4515

### DIFF
--- a/scripts/pymc3_to_4.sh
+++ b/scripts/pymc3_to_4.sh
@@ -2,6 +2,7 @@
 
 if [ -z "$1" ]
   then
+    echo "Script to automatically convert code (*.py and *.ipynb) from PyMC3 to 4.0. Use with care."
     echo "Usage: pymc3_to_4.sh <path>"
     exit 1
 fi
--- a/scripts/pymc3_to_4.sh
+++ b/scripts/pymc3_to_4.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+if [ -z "$1" ]
+  then
+    echo "Usage: pymc3_to_4.sh <path>"
+    exit 1
+fi
+
+declare -a replace_strings=(
+    "s/az.from_pymc3/pm.to_inference_data/g"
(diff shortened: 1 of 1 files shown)


ricardoV94:
I don't like this :)

scripts/pymc3_to_4.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+declare -a replace_strings=(
+    "s/az.from_pymc3/pm.to_inference_data/g"
+    "s/arviz.from_pymc3/pm.to_inference_data/g"
+    "s/pymc3/pymc/g"
+    "s/PyMC3/PyMC/g"
+    "s/from theano import tensor as tt/import aesara.tensor as at/g"
+    "s/import theano\.tensor as tt/import aesara.tensor as at/g"
+    "s/tt\./at./g"
```
lucianopaz, line 17: This is very dangerous. It will replace any `tt.` with `at.`, regardless of where in the line, `tt` is found. It should somehow have a lookbehind that checks if `tt` was not preceded by a letter, underscore or a dot. From the little I know of sed, I think that it doesn't support look around conditions ([see here](https://stackoverflow.com/questions/26110266/does-lookbehind-work-in-sed))

twiecki:
Alright then :).

canyon289:
@ricardoV94 just curious, do you not like this because its a bash script, or dont like the idea of an autoconverter?

ricardoV94:
> @ricardoV94 just curious, do you not like this because its a bash script, or dont like the idea of an autoconverter?

The second

PR #5779, opened 2022-05-18
Use unit normal as default init_dist in GaussianRandomWalk and AR

### LINKED ISSUE #5744: Improve default GaussianRandomWalk and AR init distributions
When refactoring the GaussianRandomWalk to V4, we set the `init` to default to `Normal(mu, sigma)`, which does not make much sense. `mu` is the drift parameter, and says nothing about the expected starting location of the series. `sigma` probably also makes little sense, as it assumes the uncertainty about the initial location has anything to do with the innovations noise. 

Possible alternatives:
* `init = Normal.dist(0, 1)`
* `init = Flat.dist()`, which was the default in V3, but it means that prior and posterior predictive will fail by default.

We should also rename `init` to `init_dist` to emphasize this parameter must de a distribution.

https://github.com/pymc-devs/pymc/blob/c76b9b9e74ee6f13993f255dd09d9290b882d6ab/pymc/distributions/timeseries.py#L239

A similar question arises in the AR, where we again reuse the innovation noise `sigma` for the uncertainty about the initial location of the series.

https://github.com/pymc-devs/pymc/blob/c76b9b9e74ee6f13993f255dd09d9290b882d6ab/pymc/distributions/timeseries.py#L429

ricardoV94:
Closes #5744

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -25,11 +25,14 @@ Also check out the [milestones](https://github.com/pymc-devs/pymc/milestones) fo
 
 All of the above apply to:
 
-Signature and default parameters changed for several distributions (see [#5628](https://github.com/pymc-devs/pymc/pull/5628)):
-    - `pm.StudentT` now requires either `sigma` or `lam` as kwarg
-    - `pm.StudentT` now requires `nu` to be specified (no longer defaults to 1)
-    - `pm.AsymmetricLaplace` positional arguments re-ordered
-    - `pm.AsymmetricLaplace` now requires `mu` to be specified (no longer defaults to 0)
+Signature and default parameters changed for several distributions:
+    - `pm.StudentT` now requires either `sigma` or `lam` as kwarg (see [#5628](https://github.com/pymc-devs/pymc/pull/5628))
+    - `pm.StudentT` now requires `nu` to be specified (no longer defaults to 1) (see [#5628](https://github.com/pymc-devs/pymc/pull/5628))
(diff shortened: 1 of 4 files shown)


ricardoV94:
Should we issue a warning that the default behavior changed from being a Flat in V3 to being a unit Normal?

Edit: Added it to the release notes at least

canyon289 (approved):

twiecki:
Didn't we default to `Flat`? I think that makes most sense. Or at least a Normal with high SD.

ricardoV94:
@twiecki The problem with flat is that you can't do prior and posterior predictive with it.

You couldn't do this before with AR ofc. 

The GRW did not respect the init dist and just assumed the initial point was zero, regardless of what the distribution was. That was just broken behavior imo.

We can make the default Flat, but users will perhaps be upset when their model can't be used for posterior predictive after sampling. I don't have a strong preference.

twiecki:
That's a good point. How about a SD of 10? I know it's arbitrary, but not more arbitrary than 1. And maybe add a note that it's important for the user to test.

twiecki:
Hm, but why can't we do posterior predictive with Flat? We still get a non-flat posterior for the initial dist?

ricardoV94:
> Hm, but why can't we do posterior predictive with Flat? We still get a non-flat posterior for the initial dist?

If the GRW is the likelihood, posterior predictive will try to resample it using forward sampling, which it can't because of the `Flat`. If it's a variable that is in the trace then it won't be an issue as it won't be resampled.

ricardoV94:
We can also just remove the default behavior and force users to always specify the `init_dist`. That's probably the best if there is no sane default.

PR #5786, opened 2022-05-19
add calendar to website

OriolAbril:
cc @cluhmann @reshamas

### DIFF
--- a/docs/source/community.md
+++ b/docs/source/community.md
@@ -38,6 +38,7 @@ community/how_to_ask_on_discourse
 :::{toctree}
 :caption: General resources
 
+community/calendar
 community/code_of_conduct
 Blog <https://www.pymc.io/blog.html>
 :::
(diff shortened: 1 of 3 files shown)


OriolAbril:
Preview available at https://pymc--5786.org.readthedocs.build/en/5786/community/calendar.html

cluhmann:
This looks good to me.  @reshamas had pointed me to sci-kit learn's public calendar ([here](https://blog.scikit-learn.org/calendar/)) and we might consider following that as a model.  So maybe we consider adding elements of the blurb they have above the calendar (particularly the relevance of the CoC to team events)?

reshamas:
@OriolAbril @cluhmann 
Yay!!! This looks *very* nice!  Congrats!

reshamas:
@cluhmann Wondering if you would prefer to use UTC instead of CET for default timezone.

P.S.  It is "scikit-learn", *not* "sci-kit" ;)

OriolAbril:
added some description, it would be nice to add more details, not necessarily now. I don't think we can expect users to read this and understand which of the available meetings is the one they should attend

cluhmann (approved):
Looks good to me.

twiecki (approved):

PR #5787, opened 2022-05-19
Remove rng seeder

### LINKED ISSUE #4301: Enhancement: Extend seeding option to a numpy RandomState/Generator
Hello,

I would like to discuss the possibility of adding the option of passing a [numpy RandomState or Generator](https://numpy.org/doc/stable/reference/random/index.html) as 'seed', in order to seed via a client maintained rng state machine.

SciPy allows the seed to be used like this:
> seed{None, int, RandomState, Generator}, optional
> This parameter defines the object to use for drawing random variates. If seed is None the RandomState singleton is used. If seed is an int, a new RandomState instance is used, seeded with seed. If seed is already a RandomState or Generator instance, then that object is used. Default is None.

as is mentioned [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.html#scipy.stats.rv_continuous).

This would have the following advantages:

- Thread-safe seeding
- Non-interference by other code (setting the seed via np.random.seed(seed) is vulnerable to outside manipulation and prone to create hard to detect bugs)
- Client decided RNG source if needed, but especially client maintained and manipulated if desired.

I am unfamiliar with the pymc3 code base and have only seen the 'seed' parameter used via `np.random.seed` in the few files I have searched, but it seemed like it could be generally made to support generators.

What are your thoughts on this?

Cheers,
Michael

### LINKED ISSUE #5733: Use SeedSequence spawning in Model.next_rng
Numpy docs advise using [SeedSequence](https://numpy.org/doc/stable/reference/random/bit_generators/generated/numpy.random.SeedSequence.html#numpy.random.SeedSequence) for spawning independent `bit_generators`: https://numpy.org/doc/stable/reference/random/parallel.html

Their method should be quite more robust to collisions than our naive `random.integers(2**30)` strategy, we use here and in a couple other places:

https://github.com/pymc-devs/pymc/blob/bae5087621323af3cb3cc95cf8d673da21ab4023/pymc/model.py#L1047

```python
import math

unique_seeds = 2**30
for n_seeds in (10, 100, 1_000, 10_000, 100_000):
    # birthday paradox probability: https://en.wikipedia.org/wiki/Birthday_problem
    p_collision = 1-math.perm(unique_seeds, n_seeds) / unique_seeds**n_seeds
    print(f"{n_seeds=}, {p_collision=}")

```
```
n_seeds=10, p_collision=4.1909515080540416e-08
n_seeds=100, p_collision=4.610036260510597e-06
n_seeds=1000, p_collision=0.0004650875835883195
n_seeds=10000, p_collision=0.04549425469529611
n_seeds=100000, p_collision=0.9905023499278603
```

Similar to https://github.com/aesara-devs/aesara/issues/936

### LINKED ISSUE #5784: NUTS initialized with ADVI is not reproducible across multiple calls
This happens because ADVI compiles a function from the model RandomVariables which depend on shared RandomState/Generator variables that are updated in place and not reset.

There is currently no easy way to reseed these shared variables. There used to be something like this at some point in the V4 refactor, but it's gone. Probably it will be reintroduced by whatever fixes #5733. 

This is also a limitation in prior and posterior predictive, as those functions can only be resampled with the same results if a whole `Model` is recreated with the original `rng_seeder`. This is "less of a problem" because those functions explicitly raise if a user tries to pass seeds, whereas these are still supported in `sampling`.

### LINKED ISSUE #5785: Consider removing `rng_seeder` user-interface from Model
Right now `draw`, `prior` and `posterior_predictive` can be seeded only from `rng_seeder` passed when creating a new `pm.Model`. This makes it convenient in that a user does not need to pass seeds to these functions individually to get seeded results, but it is less convenient in that there is an "invisible" dependency from the order and number of methods that are called.

In concrete, if you call `pm.draw` before `pm.sample_prior_predictive` you will get different results than if you do it in the opposite order. 
```python
import pymc as pm
with pm.Model(rng_seeder=3) as m:
    x = pm.Normal("x")
    print("draw: ", pm.draw(x))
    print("prior: ", pm.sample_prior_predictive(samples=1, return_inferencedata=False)["x"][0])
    
with pm.Model(rng_seeder=3) as m:
    x = pm.Normal("x")
    print("prior: ", pm.sample_prior_predictive(samples=1, return_inferencedata=False)["x"][0])
    print("draw: ", pm.draw(x))   
```
```python
draw:  1.1557538103058285
prior:  -1.2931919932524942
prior:  1.1557538103058285
draw:  -1.2931919932524942
```

This happens because shared RandomState/Generator variables for the model RVs are only set once when they are created (based on the rng_seeder) and never re-seeded. There is currently no easy mechanism to re-seed all variables although it seems like there was one at some point during the V4 refactor, probably when we were using a `RandomStream` under the hood.

If there was one mechanism (which will probably be reintroduced when #5733 gets fixed), users could in theory call `Model.seed(seed)` between calls to control/reset the seeding. It would look something like this:

```python
import pymc as pm
with pm.Model(rng_seeder=3) as m:
    x = pm.Normal("x")
    print("draw1: ", pm.draw(x))

    # Does not exist currently
    m.seed(3)
    print("prior1: ", pm.sample_prior_predictive(samples=1, return_inferencedata=False)["x"][0])
    
with pm.Model(rng_seeder=3) as m:
    x = pm.Normal("x")
    print("prior2: ", pm.sample_prior_predictive(samples=1, return_inferencedata=False)["x"][0])

    # Does not exist currently
    m.seed(3)
    print("draw2: ", pm.draw(x))    
```
```
draw:  1.1557538103058285
prior:  1.1557538103058285
prior:  1.1557538103058285
draw:  1.1557538103058285
```

But if users want to do this, it seems more naturally to allow them to pass a seed directly to these methods and then do the re-seeding internally for the user. It would look something like this:

```python
import pymc as pm
# rng_seeder would be removed to avoid confusion over which "way" is best
with pm.Model() as m:
    x = pm.Normal("x")
    # Neither of these methods accepts seeds currently
    print("draw1: ", pm.draw(x, seed=3))
    print("prior1: ", pm.sample_prior_predictive(samples=1, seed=3, return_inferencedata=False)["x"][0])
    
with pm.Model() as m:
    x = pm.Normal("x")
    # Neither of these methods accepts seeds currently
    print("prior2: ", pm.sample_prior_predictive(samples=1, seed=3, return_inferencedata=False)["x"][0])
    print("draw2: ", pm.draw(x, seed=3))
```
With the same results as in the previous example.

Right now only `sample` accepts seeds explicitly, but if that's missing it defaults to `rng_seeder`. IMHO it is confusing to allow two ways to seed things, and this has emerged a couple of times in pymc-examples (cc @OriolAbril). 

There is one tiny technical difference in that users can directly specify a seed per chain only via `sample`, but I think this is not enough reason to justify keeping two interfaces for that function. Also it's probably safer if users pass a single seed and we spawn multiple ones "smartly", to avoid the type of issues described in #5733

***
Creating this as an issue instead of discussion, because I think it should be addressed quickly to not create confusion as we release V4 / too much work re-running pymc-examples.

@pymc-devs/dev-core

ricardoV94:
This PR removes the `Model(rng_seeder=seed)` API as discussed in #5785 and instead makes use of the `random_seed` kwarg in all sampling functions. One big breaking change done on purpose (but not strictly needed) is that these functions will not care/respect user defined external global seeding.

```python
import pymc as pm
import numpy as np

with pm.Model() as m:
  x = pm.Normal("x")
  np.random.seed(0)  # DOES NOT MATTER!
  always_different = pm.sample()
  always_the_same = pm.sample(random_seed=0)
```

This is done so that default seeding quality can be "as good as possible", as suggested by NumPy best practice: https://numpy.org/doc/stable/reference/random/bit_generators/generated/numpy.random.SeedSequence.html#numpy.random.SeedSequence

```Python
# Global seeding is ignored by numpy default_rng, 
# and that's why our PyMC sampling functions now do the same:
import numpy as np

np.random.seed(0)
print(np.random.default_rng(None).random())  # 0.7166779009481196
np.random.seed(0)
print(np.random.default_rng(None).random())  # 0.6145552524375484
```
Is everyone okay with this change?

Also allowed passing RandomState or Generators to all sampling functions for user convenience.
```python
rng = np.random.default_rng()
with pm.Model() as m:
  x = pm.Normal("x")
  pm.sample(random_seed=rng)
```

Closes #5785 
Closes #5733
Closes #5784
Closes #4301

### DIFF
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -929,9 +929,28 @@ def reseed_rngs(
 
 
 def compile_pymc(
-    inputs, outputs, mode=None, **kwargs
+    inputs,
+    outputs,
+    random_seed: SeedSequenceSeed = None,
+    mode=None,
+    **kwargs,
 ) -> Callable[..., Union[np.ndarray, List[np.ndarray]]]:
     """Use ``aesara.function`` with specialized pymc rewrites always enabled.
+
+    This function also ensures shared RandomState/Generator used by RandomVariables
+    in the graph are updated across calls, to ensure independent draws.
+
+    Parameters
+    ----------
+    inputs: list of TensorVariables, optional
+        Inputs of the compiled Aesara function
+    outputs: list of TensorVariables, optional
+        Outputs of the compiled Aesara function
+    random_seed: int, array-like of int or SeedSequence, optional
+        Seed used to override any RandomState/Generator shared variables in the graph.
+        If not specified, the value of original shared variables will still be overwritten.
+    mode: optional
+        Aesara mode used to compile the function
 
     Included rewrites
     -----------------
@@ -952,7 +971,6 @@ def compile_pymc(
     """
     # Create an update mapping of RandomVariable's RNG so that it is automatically
     # updated after every function call
-    # TODO: This won't work for variables with InnerGraphs (Scan and OpFromGraph)
     rng_updates = {}
     output_to_list = outputs if isinstance(outputs, (list, tuple)) else [outputs]
     for random_var in (
@@ -966,10 +984,16 @@ def compile_pymc(
             rng = random_var.owner.inputs[0]
             if not hasattr(rng, "default_update"):
                 rng_updates[rng] = random_var.owner.outputs[0]
+            else:
+                rng_updates[rng] = rng.default_update
         else:
             update_fn = getattr(random_var.owner.op, "update", None)
             if update_fn is not None:
                 rng_updates.update(update_fn(random_var.owner))
+
+    # We always reseed random variables as this provides RNGs with no chances of collision
+    if rng_updates:
+        reseed_rngs(rng_updates.keys(), random_seed)
 
     # If called inside a model context, see if check_bounds flag is set to False
     try:
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -20,6 +20,7 @@ from typing import (
     Iterable,
     List,
     Optional,
+    Sequence,
     Set,
     Tuple,
     Union,
@@ -891,6 +892,40 @@ aesara.compile.optdb["canonicalize"].register(
     local_check_parameter_to_ninf_switch,
     use_db_name_as_tag=False,
 )
+
+
+def find_rng_nodes(variables: Iterable[TensorVariable]):
+    """Return RNG variables in a graph"""
+    return [
+        node
+        for node in graph_inputs(variables)
+        if isinstance(
+            node,
+            (
+                at.random.var.RandomStateSharedVariable,
+                at.random.var.RandomGeneratorSharedVariable,
+            ),
+        )
+    ]
+
+
+SeedSequenceSeed = Optional[Union[int, Sequence[int], np.ndarray, np.random.SeedSequence]]
+
+
+def reseed_rngs(
+    rngs: Sequence[SharedVariable],
+    seed: SeedSequenceSeed,
+) -> None:
+    """Create a new set of RandomState/Generator for each rng based on a seed"""
+    bit_generators = [
+        np.random.PCG64(sub_seed) for sub_seed in np.random.SeedSequence(seed).spawn(len(rngs))
+    ]
+    for rng, bit_generator in zip(rngs, bit_generators):
+        if isinstance(rng, at.random.var.RandomStateSharedVariable):
+            new_rng = np.random.RandomState(bit_generator)
+        else:
+            new_rng = np.random.Generator(bit_generator)
+        rng.set_value(new_rng, borrow=True)
 
 
 def compile_pymc(
--- a/pymc/distributions/censored.py
+++ b/pymc/distributions/censored.py
@@ -86,15 +86,11 @@ class Censored(SymbolicDistribution):
         return super().dist([dist, lower, upper], **kwargs)
 
     @classmethod
-    def num_rngs(cls, *args, **kwargs):
-        return 1
-
-    @classmethod
     def ndim_supp(cls, *dist_params):
         return 0
 
     @classmethod
-    def rv_op(cls, dist, lower=None, upper=None, size=None, rngs=None):
+    def rv_op(cls, dist, lower=None, upper=None, size=None):
 
         lower = at.constant(-np.inf) if lower is None else at.as_tensor_variable(lower)
         upper = at.constant(np.inf) if upper is None else at.as_tensor_variable(upper)
@@ -112,20 +108,7 @@ class Censored(SymbolicDistribution):
         rv_out.tag.lower = lower
         rv_out.tag.upper = upper
 
-        if rngs is not None:
-            rv_out = cls._change_rngs(rv_out, rngs)
-
         return rv_out
-
-    @classmethod
-    def _change_rngs(cls, rv, new_rngs):
-        (new_rng,) = new_rngs
-        dist_node = rv.tag.dist.owner
-        lower = rv.tag.lower
-        upper = rv.tag.upper
-        olg_rng, size, dtype, *dist_params = dist_node.inputs
-        new_dist = dist_node.op.make_node(new_rng, size, dtype, *dist_params).default_output()
-        return cls.rv_op(new_dist, lower, upper)
 
     @classmethod
     def change_size(cls, rv, new_size, expand=False):
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -104,6 +104,8 @@ ArrayLike: TypeAlias = Union[np.ndarray, List[float]]
 PointList: TypeAlias = List[PointType]
 Backend: TypeAlias = Union[BaseTrace, MultiTrace, NDArray]
 
+RandomSeed = Optional[Union[int, Sequence[int], np.ndarray]]
+
 _log = logging.getLogger("pymc")
 
 
@@ -437,14 +439,14 @@ def sample(
     if random_seed == -1:
         random_seed = None
     if chains == 1 and isinstance(random_seed, int):
-        random_seed = [random_seed]
+        random_seed_list = [random_seed]
 
     if random_seed is None or isinstance(random_seed, int):
         if random_seed is not None:
             np.random.seed(random_seed)
-        random_seed = [np.random.randint(2**30) for _ in range(chains)]
+        random_seed_list = [np.random.randint(2**30) for _ in range(chains)]
 
-    if not isinstance(random_seed, abc.Iterable):
+    if not isinstance(random_seed_list, abc.Iterable):
         raise TypeError("Invalid value for `random_seed`. Must be tuple, list or int")
 
     if not discard_tuned_samples and not return_inferencedata:
@@ -490,7 +492,7 @@ def sample(
             chains=chains,
             n_init=n_init,
             model=model,
-            seeds=random_seed,
+            seeds=random_seed_list,
             progressbar=progressbar,
             jitter_max_retries=jitter_max_retries,
             tune=tune,
@@ -506,7 +508,7 @@ def sample(
             jitter_rvs=filter_rvs_to_jitter(step),
             chains=chains,
         )
-        initial_points = [ipfn(seed) for ipfn, seed in zip(ipfns, random_seed)]
+        initial_points = [ipfn(seed) for ipfn, seed in zip(ipfns, random_seed_list)]
 
     # One final check that shapes and logps at the starting points are okay.
     for ip in initial_points:
@@ -523,7 +525,6 @@ def sample(
         "tune": tune,
         "progressbar": progressbar,
         "model": model,
-        "random_seed": random_seed,
         "cores": cores,
         "callback": callback,
         "discard_tuned_samples": discard_tuned_samples,
@@ -542,6 +543,19 @@ def sample(
     )
 
     parallel = cores > 1 and chains > 1 and not has_population_samplers
+    # At some point it was decided that PyMC should not set a global seed by default,
+    # unless the user specified a seed. This is a symptom of the fact that PyMC samplers
+    # are built around global seeding. This branch makes sure we maintain this unspoken
+    # rule. See https://github.com/pymc-devs/pymc/pull/1395.
+    if parallel:
+        # For parallel sampling we can pass the list of random seeds directly, as
+        # global seeding will only be called inside each process
+        sample_args["random_seed"] = random_seed_list
+    else:
+        # We pass None if the original random seed was None. The single core sampler
+        # methods will only set a global seed when it is not None.
+        sample_args["random_seed"] = random_seed if random_seed is None else random_seed_list
+
     t_start = time.time()
     if parallel:
         _log.info(f"Multiprocess sampling ({chains} chains in {cores} jobs)")
@@ -674,7 +688,7 @@ def _sample_many(
     chain: int,
     chains: int,
     start: Sequence[PointType],
-    random_seed: list,
+    random_seed: Optional[Sequence[RandomSeed]],
     step,
     callback=None,
     **kwargs,
@@ -691,7 +705,7 @@ def _sample_many(
         Total number of chains to sample.
     start: list
         Starting points for each chain
-    random_seed: list
+    random_seed: list of random seeds, optional
         A list of seeds, one for each chain
     step: function
         Step function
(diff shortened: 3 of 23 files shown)


michaelosthege:
@ricardoV94 I can't invest the time for a thorough review right now---the diff is a little big and seeding is a sensitive and tricky topic..

pymc/aesaraf.py
```diff
@@ -893,11 +897,58 @@ def local_check_parameter_to_ninf_switch(fgraph, node):
...
+                at.random.var.RandomGeneratorSharedVariable,
+            ),
+        )
+    ]
+
+
+def reseed_rngs(rngs: Iterable[SharedVariable], seed: "RandomSeed") -> None:
+    """Create a new set of RandomState/Generator for each rng based on a seed"""
+    bit_generators = [
+        np.random.PCG64(sub_seed) for sub_seed in np.random.SeedSequence(seed).spawn(len(rngs))
```
lucianopaz, line 921: Since we're at this, why don't we switch to [PCG-64 DXSM](https://numpy.org/doc/stable/reference/random/bit_generators/pcg64dxsm.html) by default. [Numpy says](https://numpy.org/doc/stable/reference/random/upgrading-pcg64.html#upgrading-pcg64) that the default PCG-64 has some statistical weaknesses in very parallel settings.

pymc/aesaraf.py
```diff
@@ -893,11 +897,58 @@ def local_check_parameter_to_ninf_switch(fgraph, node):
...
+    ]
+    for rng, bit_generator in zip(rngs, bit_generators):
+        if isinstance(rng, at.random.var.RandomStateSharedVariable):
+            new_rng = np.random.RandomState(bit_generator)
```
lucianopaz, line 925: This is more of an aesara question, will support for the `RandomState` eventually be removed?

pymc/aesaraf.py
```diff
@@ -893,11 +897,58 @@ def local_check_parameter_to_ninf_switch(fgraph, node):
...
+            node,
+            (
+                at.random.var.RandomStateSharedVariable,
+                at.random.var.RandomGeneratorSharedVariable,
+            ),
+        )
+    ]
+
+
+def reseed_rngs(rngs: Iterable[SharedVariable], seed: "RandomSeed") -> None:
```
lucianopaz: This comment came after looking at `compile_pymc`. When no random seed is provided, and `seed is None`. What do you want this to do exactly?

pymc/distributions/censored.py
```diff
@@ -112,21 +108,8 @@ def rv_op(cls, dist, lower=None, upper=None, size=None, rngs=None):
         rv_out.tag.lower = lower
         rv_out.tag.upper = upper
 
-        if rngs is not None:
-            rv_out = cls._change_rngs(rv_out, rngs)
-
         return rv_out
 
-    @classmethod
-    def _change_rngs(cls, rv, new_rngs):
```
lucianopaz, line 121: What did this have to do with seeding?

pymc/sampling.py
```diff
@@ -253,6 +255,38 @@ def all_continuous(vars):
...
+        random_state = tuple(random_state)
+    elif not isinstance(random_state, (list, tuple)):
+        raise ValueError(f"The `seeds` must be array-like. Got {type(random_state)} instead.")
+
+    if len(random_state) != chains:
+        raise ValueError(
+            f"Number of seeds ({len(random_state)}) does not match the number of chains ({chains})."
+        )
+
+    return random_state
```
lucianopaz: What's the point of this function? To get different values to feed into the `numpy.random.SeedSequence` as the seeder's entropy? If yes, that's not the best approach to get independent random streams.
The best approach is to have a single entropy value for the `SeedSequence`, and spawn as many seeders as necessary. The resulting `spawn_key` should ensure that you get independent streams. From the [numpy docs](https://numpy.org/doc/stable/reference/random/bit_generators/generated/numpy.random.SeedSequence.html#numpy.random.SeedSequence), it looks like we can feed a common entropy value to many `SeedSequence` instances but specify the `spawn_key`. So there must be a better way to do what we want, but I need to understand how this function is integrated to the rest of the `rngs` during sampling.

lucianopaz (commented):
I have a bunch of questions @ricardoV94. The aesaraf functions look good but I have some questions regarding sampling. How does the random seed go into the step methods at the moment? The idea of passing different "seeds" (the entropy parameter) to `SeedSequence` isn't the best way to ensure independent streams. The best way is to spawn as many seeders as necessary. These will all get the same entropy value, but they'll have different `spawn_keys`. We should aim to use this instead of different entropies somehow.

pymc/aesaraf.py
ricardoV94: I checked that but that bit_generator was not available in my NumPy version (1.20.3), and I did not want to raise the NumPy dependency (for users). 

I added an explicit test with information about the decision that will have to be done if NumPy switches their default. Until they do, I think we can keep as is: https://github.com/pymc-devs/pymc/blob/8a3cb40504e06de5a51ad4c38d84aa3771fbd3f8/pymc/tests/test_aesaraf.py#L515-L521

pymc/aesaraf.py
ricardoV94: I think so, but not anytime soon. They have been struggling to get `NUMBA` to work with Generators.

pymc/aesaraf.py
ricardoV94: When no seed is provided we still want to override the values. This is because by default when you initialize a Aesara RandomVariable without specifying a RandomState/Shared variable it will create one on the spot. This however will not have the nice properties of Seeding with a SeedSequence and there's a considerable high change of collision/ bad bit_generators.

pymc/distributions/censored.py
ricardoV94: SymbolicDistributions required the ability to calculate the number of RNGs and reseed them in their graphs, because PyMC models were creating these variable states eagerly (and not letting them update after creation easily). Now that reseeding is done at compilation time, there is no need for these methods.

pymc/sampling.py
ricardoV94: This function does two things and perhaps not very well yet. 

1. It allows users to pass RandomState/Generators via `random_seed`. Numpy SeedSequence does not accept these as inputs AFAICT, so we need to create integer(s) to seed the SeedSequence. Addresses https://github.com/pymc-devs/pymc/issues/4301 
2. It creates multiple seeds per chain, which unfortunately we need as our PyMC sampling step methods rely on global seeding.

There is a bit of a redundant case where a user specifies a single integer seed and there is only need for one. In that case we are needlessly creating a Generator just to draw another integer from it. This however should have no less quality than the original number. Two notes here:
1. If users pass `None` to any of the forward sampling functions, we don't do this conversion and allow the seeding to be done with it (best recommendation according to NumPy).
2. Seeds don't need to be integers, they can also be a list of integers. However PyMC looks at this for evidence that we want multiple seeds. I decided to keep this limitation for simplicity.

ricardoV94:
> I have a bunch of questions @ricardoV94. The aesaraf functions look good but I have some questions regarding sampling. How does the random seed go into the step methods at the moment? The idea of passing different "seeds" (the entropy parameter) to `SeedSequence` isn't the best way to ensure independent streams. The best way is to spawn as many seeders as necessary. These will all get the same entropy value, but they'll have different `spawn_keys`. We should aim to use this instead of different entropies somehow.

In forward sampling, we always go via a single `compile_pymc` and we do a single `SeedSequence(seed).spawn(len(rngs))`. This should be optimal.

In MCMC sampling, all our step methods rely on global seeding. We have to create different integer seeds for each chain. Ideally we would be passing Generators around instead, and we would probably spawn them (or their SeedSequences) before branching off. This however will require a larger refactoring. See https://github.com/pymc-devs/pymc/issues/5093

pymc/sampling.py
ricardoV94: I should add tests directly to this function so that the behavior is better covered/documented (and obviously better docs as well). But first, do you agree it makes sense to have this helper?

pymc/aesaraf.py
ricardoV94: I also thought about using a sentinel default value `UNSET`, and not do anything when a user explicitly passes `None`.

lucianopaz: Ok, it seems like a reasonable default to me

pymc/distributions/censored.py
lucianopaz: Nice

pymc/aesaraf.py
lucianopaz: The only alternative I could think of is to have some sort of global seeder and spawn from it by default, but having `None` as an initial entropy looks more consistent

lucianopaz: Another option that I just realized is that we could play around with the extra keyword arguments of `SeedSequence`:

```python
>>> np.random.SeedSequence(entropy=1).spawn(1)
SeedSequence(
    entropy=1,
    spawn_key=(1,),
)
>>> np.random.SeedSequence(entropy=1, n_children_spawned=1).spawn(1)[0]
SeedSequence(
    entropy=1,
    spawn_key=(1,),
)
>>> np.random.SeedSequence(entropy=1, spawn_key=[1])
SeedSequence(
    entropy=1,
    spawn_key=(1,),
)
>>> np.random.SeedSequence(entropy=1, spawn_key=[1]).spawn(1)[0]
SeedSequence(
    entropy=1,
    spawn_key=(1, 0),
)
```

It looks like there's a bunch of ways to get the exact same `SeedSequence`, and that if you use a `SeedSequence` that already has a `spawn_key`, the result is a new `SeedSequence` that can generate independent streams from its parent or siblings.

pymc/sampling.py
```diff
@@ -253,6 +255,38 @@ def all_continuous(vars):
...
+    random_state: RandomState,
+    chains: int,
+) -> Sequence[int]:
+    """Obtain new or validate specified integer seeds per chain"""
+
+    def _get_unique_seeds_per_chain(integers_fn):
+        seeds = []
+        while len(set(seeds)) != chains:
+            seeds = [int(seed) for seed in integers_fn(2**30, dtype=np.int64, size=chains)]
+        return seeds
```
lucianopaz, line 286: We could maybe avoid drawing as many integers as the total number of `chains` every time? Maybe we could use an order preserving set like entity? Something like a dictionary with no value corresponding to the keys?
```suggestion
        seeds = {}
        while len(set(seeds)) < chains:
            seeds.update(
                {int(seed): None for seed in integers_fn(2**30, dtype=np.int64, size=chains - len(seeds))}
            )
        return list(seeds)
```

pymc/sampling.py
lucianopaz: I see, so we aren't passing seeds or anything like that to the steppers! We should open an issue to change that and have some way of providing independent streams given by `SeedSequence` to the steppers. We could use something like `SeedSequence(entropy=some_integer, spawn_key=(chain_idx, stepper_idx))`, which would guarantee independent streams of random numbers for each chain and stepper, but at the same time be deterministic given the entropy value, which would play the role of a single `random_seed` given to the `sample` function.

Anyway, given what you said about stepper relying on global seeds, this approach looks ok.

ricardoV94: Alternatively we can also just passa single generator or even seedsequence to each stepper and let it create/branch generators as they please. There's a vague issue about this here, we should update with more details: https://github.com/pymc-devs/pymc/issues/5093

ricardoV94: @lucianopaz I updated the docstrings of this function, added a direct test and made it so nothing is done if an integer is passed and a single chain is requested

pymc/sampling.py
ricardoV94: I am not too concerned with this. With 10_000 chains we would still expect zero collisions 95% of the time, and with 1000 chains 99.95% of the time.

I would rather leave it in this simple form.  This logic will be entirely removed once we stop using global seeding for our PyMC samplers.

twiecki (approved):

lucianopaz (approved):
Looks very good @ricardoV94 !

lucianopaz:
@ricardoV94, I’ll let you merge since I’m not sure if you want to squash merge or rebase and preserve the history

ricardoV94:
> @ricardoV94, I’ll let you merge since I’m not sure if you want to squash merge or rebase and preserve the history

Always rebase with me :)

ricardoV94:
Thanks for reviewing!

PR #5788, opened 2022-05-20
Remove size from Distribution signatures

### LINKED ISSUE #5754: Remove size from distribution API
This was brought up in the last meeting, the question is whether there's any benefit in introducing size as a new parameter to our distributions. Size is simply `shape[:-ndim_supp]` so it's pretty easy for us to convert a user provided shape to size which Aesara RandomVariables use and will keep using internally.

We should decide this before we release V4. For some context on the distinction see this WIP PR https://github.com/pymc-devs/pymc/pull/5746

For a more involved example where we are converting from shape/dims/observed to size, take a look at these PRs https://github.com/pymc-devs/pymc/pull/5743 and https://github.com/pymc-devs/pymc/pull/5690

michaelosthege:
This PR implements how I propose to resolve #5754.

It kicks the argument from the signature, but continues to support it for internal use & to maintain API continuity with Aesara & aeppl.

I made the corresponding changes for `Flat`, `HalfFlat`, but I don't know how to go about `Bound` which has multiple other kwargs from the Distribution API in it's signature.
Also the GaussianRandomWalk has a `size` kwarg that I wasn't sure how to remove without having to starting thinking.

Anybody please feel free to take over the branch.

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -344,7 +344,7 @@ class FlatRV(RandomVariable):
     _print_name = ("Flat", "\\operatorname{Flat}")
 
     @classmethod
-    def rng_fn(cls, rng, size):
+    def rng_fn(cls, rng, *args):
         raise NotImplementedError("Cannot sample from flat variable")
 
 
@@ -364,8 +364,8 @@ class Flat(Continuous):
         return super().__new__(cls, *args, **kwargs)
 
     @classmethod
-    def dist(cls, *, size=None, **kwargs):
-        res = super().dist([], size=size, **kwargs)
+    def dist(cls, **kwargs):
+        res = super().dist([], **kwargs)
         return res
 
     def moment(rv, size):
@@ -415,7 +415,7 @@ class HalfFlatRV(RandomVariable):
     _print_name = ("HalfFlat", "\\operatorname{HalfFlat}")
 
     @classmethod
-    def rng_fn(cls, rng, size):
+    def rng_fn(cls, rng, *args):
         raise NotImplementedError("Cannot sample from half_flat variable")
 
 
@@ -432,8 +432,8 @@ class HalfFlat(PositiveContinuous):
         return super().__new__(cls, *args, **kwargs)
 
     @classmethod
-    def dist(cls, *, size=None, **kwargs):
-        res = super().dist([], size=size, **kwargs)
+    def dist(cls, **kwargs):
+        res = super().dist([], **kwargs)
         return res
 
     def moment(rv, size):
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -344,7 +344,7 @@ class FlatRV(RandomVariable):
     _print_name = ("Flat", "\\operatorname{Flat}")
 
     @classmethod
-    def rng_fn(cls, rng, *args):
+    def rng_fn(cls, rng, size):
         raise NotImplementedError("Cannot sample from flat variable")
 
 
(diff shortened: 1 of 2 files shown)


canyon289:
Thanks @michaelosthege for taking action! The code here looks good to me, will let others comment on the "philosophy of it"

twiecki:
Related to https://github.com/pymc-devs/pymc/issues/5789

pymc/distributions/continuous.py
```diff
@@ -344,7 +344,7 @@ class FlatRV(RandomVariable):
     _print_name = ("Flat", "\\operatorname{Flat}")
 
     @classmethod
-    def rng_fn(cls, rng, size):
+    def rng_fn(cls, rng, *args):
```
ricardoV94: I thought size might be passed as kwarg by Aesara, but then it would have failed (assuming we have coverage). In any case it can still exist in RandomVariables and rng_fn because those are always defined in terms of size

pymc/distributions/continuous.py
```diff
@@ -415,7 +415,7 @@ class HalfFlatRV(RandomVariable):
     _print_name = ("HalfFlat", "\\operatorname{HalfFlat}")
 
     @classmethod
-    def rng_fn(cls, rng, size):
+    def rng_fn(cls, rng, *args):
```
ricardoV94: same here

pymc/distributions/distribution.py
```diff
@@ -226,8 +226,8 @@ def __new__(
         transform : optional
             See ``Model.register_rv``.
         **kwargs
-            Keyword arguments that will be forwarded to ``.dist()``.
-            Most prominently: ``shape`` and ``size``
+            Keyword arguments that will be forwarded to ``.dist()`` or the Aesara RV Op.
+            Most prominently: ``shape`` for ``.dist()`` and ``size`` or ``dtype`` for the Op.
```
ricardoV94: This could be confusing, users cant specify both shape and size. Better to not mention at all?

michaelosthege:
```suggestion
            Most prominently: ``shape`` for ``.dist()`` or ``dtype`` for the Op.
```

pymc/distributions/continuous.py
michaelosthege: The superclass doesn't have it in the signatures and also many other distributions don't have it.

Generally we shouldn't override signatures - it angers the typechecking gods.

lucianopaz: I strongly disagree with this change. It's the `rng_fn` that aesara needs and we should use aesara's expected signature for it. I think that we should leave `size` here since a `RandomVariable` `Op` is low level enough as to not confuse beginners.

pymc/distributions/continuous.py
lucianopaz: Same coment as above. We should keep `size` here

pymc/distributions/continuous.py
michaelosthege: That's fine with me, ist just think we should be consistent.
I'll revert the changes in `rng_fn` signatures.
```suggestion
    def rng_fn(cls, rng, size):
```

pymc/distributions/continuous.py
michaelosthege:
```suggestion
    def rng_fn(cls, rng, size):
```

michaelosthege:
@lucianopaz @ricardoV94 @twiecki @canyon289 are we happy?

(Want to get the ticket off my mental ToDo list..)

ricardoV94:
I am okay. This does not really address the bigger discussion, but we can definitely merge.

ricardoV94 (approved):

canyon289:
yes, this does two very important thing which is remove the confusion for
users and subset the discussion to an internal one.

thanks much Michael for your patience with all this 🙏

PR #5793, opened 2022-05-23
Latex representation for SymbolicDistributions

### LINKED ISSUE #5616: Latex representation broken for models with SymbolicDistributions

larryshamalama:
Closes #5616.

This PR adds Latex presentation to `SymbolicDistribution`s which does not construct `rv_op`s like regular `Distribution`s, hence why the Latex representation is currently broken.

I would like some advice on how to carry this out. Dispatching was recommended, but the initial class which was called does not seem to be encoded in `TensorVariable`s for `SymbolicDistribution`s. I currently set it up so that a new `str_for_symbolic_dist` function would be called for `SymbolicDistribution`s for `_print_name`s which can be automatically initialized for such distributions (except for zero-inflated ones because they are created via `pm.Mixture`). Here are the internal changes that I came up with, but I'm happy to hear discussion about these points:

- `cls` is now passed in `_zero_inflated_mixture` so they don't come out as "Mixture"
- Akin to [`str_for_dist`](https://github.com/pymc-devs/pymc/blob/main/pymc/printing.py#L34), `str_for_symbolic_dist` is created in `printing.py` in which pattern matching can be used to determine how to print the corresponding distribution. I have yet to address this and perhaps this is where dispatching would be useful?
- `_print_name` attributes initialized for `SymbolicDistribution`s

Some printing references that have been provided: [AePPL printing](https://github.com/aesara-devs/aeppl/blob/main/aeppl/printing.py) and [symbolic-pymc printing](https://github.com/pymc-devs/symbolic-pymc/blob/84e8d612c714f502f8d188c1766498f4ff7beecf/symbolic_pymc/tensorflow/printing.py). Both of which use dispatching, but only the latter uses `singledispatch`.

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -42,7 +42,7 @@ from pymc.distributions.dist_math import (
     normal_lccdf,
     normal_lcdf,
 )
-from pymc.distributions.distribution import Discrete
+from pymc.distributions.distribution import Discrete, set_print_name
 from pymc.distributions.logprob import logp
 from pymc.distributions.mixture import Mixture
 from pymc.distributions.shape_utils import rv_size_is_none
@@ -1411,7 +1411,7 @@ class Constant(Discrete):
         )
 
 
-def _zero_inflated_mixture(*, name, nonzero_p, nonzero_dist, **kwargs):
+def _zero_inflated_mixture(*, cls, name, nonzero_p, nonzero_dist, **kwargs):
     """Helper function to create a zero-inflated mixture
 
     If name is `None`, this function returns an unregistered variable
@@ -1423,9 +1423,14 @@ def _zero_inflated_mixture(*, name, nonzero_p, nonzero_dist, **kwargs):
         nonzero_dist,
     ]
     if name is not None:
-        return Mixture(name, weights, comp_dists, **kwargs)
+        rv_out = Mixture(name, weights, comp_dists, **kwargs)
     else:
-        return Mixture.dist(weights, comp_dists, **kwargs)
+        rv_out = Mixture.dist(weights, comp_dists, **kwargs)
+
+    # overriding Mixture _print_name
+    set_print_name(cls, rv_out)
+
+    return rv_out
 
 
 class ZeroInflatedPoisson:
@@ -1481,13 +1486,13 @@ class ZeroInflatedPoisson:
 
     def __new__(cls, name, psi, mu, **kwargs):
         return _zero_inflated_mixture(
-            name=name, nonzero_p=psi, nonzero_dist=Poisson.dist(mu=mu), **kwargs
+            cls=cls, name=name, nonzero_p=psi, nonzero_dist=Poisson.dist(mu=mu), **kwargs
         )
 
     @classmethod
     def dist(cls, psi, mu, **kwargs):
         return _zero_inflated_mixture(
-            name=None, nonzero_p=psi, nonzero_dist=Poisson.dist(mu=mu), **kwargs
+            cls=cls, name=None, nonzero_p=psi, nonzero_dist=Poisson.dist(mu=mu), **kwargs
         )
 
 
@@ -1545,13 +1550,13 @@ class ZeroInflatedBinomial:
 
     def __new__(cls, name, psi, n, p, **kwargs):
         return _zero_inflated_mixture(
-            name=name, nonzero_p=psi, nonzero_dist=Binomial.dist(n=n, p=p), **kwargs
+            cls=cls, name=name, nonzero_p=psi, nonzero_dist=Binomial.dist(n=n, p=p), **kwargs
         )
 
     @classmethod
     def dist(cls, psi, n, p, **kwargs):
         return _zero_inflated_mixture(
-            name=None, nonzero_p=psi, nonzero_dist=Binomial.dist(n=n, p=p), **kwargs
+            cls=cls, name=None, nonzero_p=psi, nonzero_dist=Binomial.dist(n=n, p=p), **kwargs
         )
 
 
@@ -1638,6 +1643,7 @@ class ZeroInflatedNegativeBinomial:
 
     def __new__(cls, name, psi, mu=None, alpha=None, p=None, n=None, **kwargs):
         return _zero_inflated_mixture(
+            cls=cls,
             name=name,
             nonzero_p=psi,
             nonzero_dist=NegativeBinomial.dist(mu=mu, alpha=alpha, p=p, n=n),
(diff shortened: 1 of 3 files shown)


pymc/printing.py
```diff
@@ -55,6 +56,37 @@ def str_for_dist(rv: TensorVariable, formatting: str = "plain", include_params:
...
+
+    # code would look something like this:
+    # if "ZeroInflated" in rv.owner.op._print_name[0]:
+    #     return
+    # if "Mixture" in rv.owner.op._print_name[0]:
+    #     return
+
+    # below is copy-pasted from str_for_dist
+    if include_params:
+        # first 3 args are always (rng, size, dtype), rest is relevant for distribution
```
ricardoV94: This is not true for SymbolicDists. We might just want to exclude any inputs that are Shared RandomState/Generator variables.

pymc/distributions/distribution.py
```diff
@@ -500,15 +500,11 @@ def __new__(
...
-        # TODO: Refactor this
-        # add in pretty-printing support
-        rv_out.str_repr = lambda *args, **kwargs: name
-        rv_out._repr_latex_ = f"\\text{name}"
-        # rv_out.str_repr = types.MethodType(str_for_dist, rv_out)
-        # rv_out._repr_latex_ = types.MethodType(
-        #     functools.partial(str_for_dist, formatting="latex"), rv_out
-        # )
-
+        set_print_name(cls, rv_out)
```
ricardoV94, line 488: If we clone a variable (e.g., during graph rewrite) this will be lost right? It's not a big issue for now, but of that's the case we should add a comment stating so

pymc/distributions/discrete.py
```diff
@@ -1423,9 +1423,14 @@ def _zero_inflated_mixture(*, name, nonzero_p, nonzero_dist, **kwargs):
...
     ]
     if name is not None:
-        return Mixture(name, weights, comp_dists, **kwargs)
+        rv_out = Mixture(name, weights, comp_dists, **kwargs)
     else:
-        return Mixture.dist(weights, comp_dists, **kwargs)
+        rv_out = Mixture.dist(weights, comp_dists, **kwargs)
+
+    # overriding Mixture _print_name
+    set_print_name(cls, rv_out)
```
ricardoV94, line 1431: The alternative would be to be more clever about mixture print name and on the spot check:
1. If it has 2 components and one is a zero constant, call it "ZeroInflatedX", where "X" is the name of the nonzero component, like "ZeroInflatedBinomial"
2. If all the components follow the same distribution call it "XMixture" like "NormalMixture" or "GammaMixture"
3. If there are two different components, call it "X-YMixture" like "Gamma-ExponentialMixture"
4. Otherwise call it simply Mixture

This is where the dispatching idea becomes a bit more powerful. You can have an arbitrarily complex function that specializes on an `Op` and can look at its inputs at evaluation time to figure out a nice name.

The same type of logic could be used for Censored (dispatched on at.clip) to call it "CensoredX" like "CensoredNormal", or the future RandomWalk to call them "XRandomWalk" (perhaps with a special case for Normal, where we call it "Gaussian")

The dispatching itself only means we don't need to define the name at creation time (and others can overwrite it more easily). The more fundamental difference is that it uses a per-op function to decide what name to give to the distribution. 

The basecase would be the `RandomVariable` `Op` which does what was already done eagerly before this PR.

This is just an idea. Feel free to investigate something like this or leave it as a separate enhancement/feature request issue!

ricardoV94 (commented):
Left some comments below. We need some yummy tests!

larryshamalama:
Notes for myself for tomorrow or Wednesday:

- How to deal with `TensorConstants` or random variable constants (like in the `ZeroInflatedPoisson`)
- Single component mixture representation

And of course adding aforementioned yummy tests

pymc/printing.py
```diff
@@ -59,17 +59,49 @@ def str_for_dist(rv: TensorVariable, formatting: str = "plain", include_params:
...
+                # is a single component!
+                # (rng, weights, single_component)
+                pass
+
+        elif "Censored" in rv.owner.op._print_name[0]:
+            start_idx_para = 2
+        else:
+            raise ValueError(
+                "Latex printing not yet implemented for this SymbolicDistribution\n"
+                "Please update the str_for_symbolic_dist in pymc/printing.py file."
```
ricardoV94: We should not raise a ValueError (unless this is being caught up somewhere upstream). That's what originated this issue. We should have a sensible fallback like "Unkwon" or whatever

larryshamalama: Done!

larryshamalama:
Need to check:

- Censored distributions
- Mixtures with non-constant mixing probabilities
- Where `UnmeasurableConstantRV` comes from

larryshamalama:
Mixing weights for mixture components needs to be discussed because it would hard to delineate them from `nonzero_p` in zero-inflated distributions. Also, in `[p, 1 - p]`, `1 - p` appears as an `Elemwise`.

larryshamalama:
*Note: I just marked as ready for review, but it is not ready to be merged. It is missing tests and how to properly address mixture weights.

ricardoV94:
@twiecki this needs to be a blocker for V4, otherwise you get errors when creating models with Distributions in an interactive environment like jupyter or ipython

michaelosthege:
@larryshamalama @ricardoV94 how much work is left on this one? Can we include it in the 4.0.2 milestone?

larryshamalama:
To do it well, I think that there is still quite some work to be done. We wanted to do it the hackish way in preparation for v4 but we temporarily have [this even more hackish PR](https://github.com/pymc-devs/pymc/pull/5847) merged. I just saw that you added this to the v4.2.0 milestone and I think that's appropriate since I'd like to work on this slowly.

ricardoV94:
Closing this in favor of #6072

PR #5799, opened 2022-05-24
raise not implemented error for non refactored distributions and functions

### LINKED ISSUE #5676: Warn or raise explicitly for known broken features in V4
- [x] Raise a warning when sampling with a single chain that seeding is not deterministic #5378 - Fixed by #5377
- [x] Raise an error when trying to use not yet refactored posterior_predictive_w Closed by #5799 
- [x] Raise an error when trying to use not yet refactored timeseries distributions - Closed by #5799

ferrine:
Closes #5676

* The PR raises a `NotImplemented` error on attempt to create not yet refactored timeseries distributions or sample_posterior_predctive_w
* Also refactored tests to catch the correct error is passed to the user side

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -648,6 +648,13 @@ class GARCH11(distribution.Continuous):
         initial_vol >= 0, initial volatility, sigma_0
     """
 
+    def __new__(cls, *args, **kwargs):
+        raise NotImplementedError(f"{cls.__name__} has not yet been ported to PyMC 4.0.")
+
+    @classmethod
+    def dist(cls, *args, **kwargs):
+        raise NotImplementedError(f"{cls.__name__} has not yet been ported to PyMC 4.0.")
+
     def __init__(self, omega, alpha_1, beta_1, initial_vol, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
@@ -705,6 +712,13 @@ class EulerMaruyama(distribution.Continuous):
         parameters of the SDE, passed as ``*args`` to ``sde_fn``
     """
 
+    def __new__(cls, *args, **kwargs):
+        raise NotImplementedError(f"{cls.__name__} has not yet been ported to PyMC 4.0.")
+
+    @classmethod
+    def dist(cls, *args, **kwargs):
+        raise NotImplementedError(f"{cls.__name__} has not yet been ported to PyMC 4.0.")
+
     def __init__(self, dt, sde_fn, sde_pars, *args, **kwds):
         super().__init__(*args, **kwds)
         self.dt = dt = at.as_tensor_variable(dt)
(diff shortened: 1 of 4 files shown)


ricardoV94:
@ferrine  MvStudentTRandomWalk and  MvGaussianRandomWalk are also not refactored yet

pymc/distributions/timeseries.py
```diff
@@ -666,6 +666,20 @@ class GARCH11(distribution.Continuous):
...
+    def __new__(
+        cls,
+        name: str,
+        *args,
+        rng=None,
+        dims: Optional[Dims] = None,
+        initval=None,
+        observed=None,
+        total_size=None,
+        transform=...,
```
ricardoV94: Should be enough to have args and kwargs no?

```suggestion
```

pymc/distributions/timeseries.py
```diff
@@ -723,6 +737,20 @@ class EulerMaruyama(distribution.Continuous):
...
+        **kwargs,
+    ) -> RandomVariable:
+        raise NotImplementedError("EulerMaruyama is not yet refactored for pymc v4")
```
twiecki:
```suggestion
        raise NotImplementedError("EulerMaruyama has not yet been ported to PyMC 4.0.")
```

pymc/distributions/timeseries.py
```diff
@@ -666,6 +666,20 @@ class GARCH11(distribution.Continuous):
...
+        *args,
+        rng=None,
+        dims: Optional[Dims] = None,
+        initval=None,
+        observed=None,
+        total_size=None,
+        transform=...,
+        **kwargs,
+    ) -> RandomVariable:
+        raise NotImplementedError("GARCH11 is not yet refactored for pymc v4")
```
twiecki:
```suggestion
        raise NotImplementedError("GARCH11 has not yet been ported to PyMC 4.0.")
```

michaelosthege:
```suggestion
        raise NotImplementedError("GARCH11 has not yet been ported to PyMC >=4.0.")
```

pymc/distributions/timeseries.py
michaelosthege:
```suggestion
        raise NotImplementedError("EulerMaruyama has not yet been ported to PyMC >=4.0.")
```

michaelosthege (commented):
I would phrase the errors differently, but other than that I would say this PR only needs suggested changes to be applied and then we're good to go.

pymc/distributions/timeseries.py
ferrine: That was an autocomplete from vscode I relied on, I agree, generic kwargs are better

ferrine:
The things left so far as I read are the MvStudentTRandomWalk and MvGaussianRandomWalk

michaelosthege (approved):

twiecki (approved):

Issue #4807, opened 2021-06-27
Refactor `sample_posterior_predictive_w` for V4


ricardoV94:

ricardoV94:
Maybe we should deprecate this one instead?

aloctavodia:
And move the functionality to sample_posterior_predictive?

ricardoV94:
> And move the functionality to sample_posterior_predictive?

I was discussing deprecating it altogether (not saying we should, just asking).

aloctavodia:
I think we should keep the functionality either in this function or in `sample_posterior_predictive`

ricardoV94:
@aloctavodia would you be interested/available to refactor this one? Nobody volunteered in a long time now :)

twiecki:
@zaxtax I think expressed interest.

zaxtax:
I have something that works and passes unit tests, but is not efficient. Maybe I can open a WIP for it.

ricardoV94:
> I have something that works and passes unit tests, but is not efficient. Maybe I can open a WIP for it.

Definitely!

ricardoV94:
I suggest removing this feature due to low interest from the community

zaxtax:
@ricardoV94 I agree. I think this should be deprecated in favour of https://github.com/arviz-devs/arviz/pull/2147 when it is merged.

PR #5800, opened 2022-05-24
Refactor `sample_posterior_predictive_w` for V4

zaxtax:
This tries to address #4807 

Right now, it passes all unit tests, but doesn't use the fast sampling procedure of `sample_posterior_predictive`

The code is also not feature-complete.

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -2032,8 +2032,6 @@ def sample_posterior_predictive_w(
         weighted models (default), or a dictionary with variable names as keys, and samples as
         numpy arrays.
     """
-    raise NotImplementedError(f"sample_posterior_predictive_w has not yet been ported to PyMC 4.0.")
-
     if isinstance(traces[0], InferenceData):
         n_samples = [
             trace.posterior.sizes["chain"] * trace.posterior.sizes["draw"] for trace in traces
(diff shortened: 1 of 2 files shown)


ricardoV94:
fast posterior_predictive_sampling no longer exists (everything is fast now). Also some tweaks will be needed following #5787 which should be merged soon!

zaxtax:
Let me make those tweaks and push tonight.

pymc/sampling.py
```diff
@@ -1958,13 +1958,14 @@ def sample_posterior_predictive_w(
             # TODO sample_posterior_predictive_w is currently only work for model with
             # one observed.
             # XXX: This needs to be refactored
-            # ppc[var.name].append(draw_values([var], point=param, size=size[idx])[0])
-            raise NotImplementedError()
+            ppcl[var.name].append(draw([var])[0])
```
ricardoV94, line 2141: This won't be conditioned on the posterior samples. 

You will have to create a custom function similarly to what `sample_posterior_predictive` does. See recent changes in https://github.com/pymc-devs/pymc/pull/5759

ricardoV94 (changes requested):

pymc/sampling.py
ricardoV94: It might also be useful to compare what changed between V3 and V4 in `sample_posterior_predictive` because some of the things in `sample_posterior_predictive_w` might no longer be needed / make sense

zaxtax:
Closing in favour of #6254

PR #5804, opened 2022-05-25
Fix non-reparametrized GP prior shape problem and forward `dims`

### LINKED ISSUE #5803: Shape problem in GP priors when `reparameterize=False`
When passing `reparameterize=False` to a `pm.Latent` prior, the resulting variable changes its shape.

```python
with pm.Model() as pmodel:

    ls = pm.LogNormal("ls")
    scaling = pm.LogNormal("scaling")

    mean_func = pm.gp.mean.Zero()
    cov_func = scaling**2 * pm.gp.cov.ExpQuad(input_dim=1, ls=ls)
    gp = pm.gp.Latent(mean_func=mean_func, cov_func=cov_func)
    
    x = numpy.array([1,2,3])
    predA = gp.prior(
        "predA",
        x[:, None],
        size=3,
        reparameterize=True,
    )
    print(predA.shape.eval())
    # [3]

    predB = gp.prior(
        "predB",
        x[:, None],
        size=3,
        reparameterize=False,
    )
    print(predB.shape.eval())
    # [3 3]
```

* PyMC/PyMC3 Version: `main`

michaelosthege:
Fixes #5803 by using a straightforward link to the input shape - symbolic or not.

I added asserts into existing tests to safeguard against the original issue.

Also, I forwarded the `dims` from `kwargs` to the `Deterministic` variable that's created in the reparametrization.
This leads to prettier `pm.model_to_graphviz` outputs.

### DIFF
--- a/pymc/gp/gp.py
+++ b/pymc/gp/gp.py
@@ -27,7 +27,6 @@ from pymc.gp.util import (
     JITTER_DEFAULT,
     cholesky,
     conditioned_vars,
-    infer_size,
     replace_with_values,
     solve_lower,
     solve_upper,
@@ -131,7 +130,7 @@ class Latent(Base):
         mu = self.mean_func(X)
         cov = stabilize(self.cov_func(X), jitter)
         if reparameterize:
-            size = infer_size(X, kwargs.pop("size", None))
+            size = np.shape(X)[0]
             v = pm.Normal(name + "_rotated_", mu=0.0, sigma=1.0, size=size, **kwargs)
             f = pm.Deterministic(name, mu + cholesky(cov).dot(v))
         else:
@@ -278,7 +277,7 @@ class TP(Latent):
         mu = self.mean_func(X)
         cov = stabilize(self.cov_func(X), jitter)
         if reparameterize:
-            size = infer_size(X, kwargs.pop("size", None))
+            size = np.shape(X)[0]
             v = pm.StudentT(name + "_rotated_", mu=0.0, sigma=1.0, nu=self.nu, size=size, **kwargs)
             f = pm.Deterministic(name, mu + cholesky(cov).dot(v))
         else:
--- a/pymc/gp/gp.py
+++ b/pymc/gp/gp.py
@@ -132,7 +132,7 @@ class Latent(Base):
         if reparameterize:
             size = np.shape(X)[0]
             v = pm.Normal(name + "_rotated_", mu=0.0, sigma=1.0, size=size, **kwargs)
-            f = pm.Deterministic(name, mu + cholesky(cov).dot(v))
+            f = pm.Deterministic(name, mu + cholesky(cov).dot(v), dims=kwargs.get("dims", None))
         else:
             f = pm.MvNormal(name, mu=mu, cov=cov, **kwargs)
         return f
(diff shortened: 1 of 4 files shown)


pymc/gp/gp.py
```diff
@@ -131,9 +130,9 @@ def _build_prior(self, name, X, reparameterize=True, jitter=JITTER_DEFAULT, **kw
         mu = self.mean_func(X)
         cov = stabilize(self.cov_func(X), jitter)
         if reparameterize:
-            size = infer_size(X, kwargs.pop("size", None))
+            size = np.shape(X)[0]
             v = pm.Normal(name + "_rotated_", mu=0.0, sigma=1.0, size=size, **kwargs)
-            f = pm.Deterministic(name, mu + cholesky(cov).dot(v))
+            f = pm.Deterministic(name, mu + cholesky(cov).dot(v), dims=kwargs.pop("dims", None))
```
ricardoV94: Any reason to pop?
```suggestion
            f = pm.Deterministic(name, mu + cholesky(cov).dot(v), dims=kwargs.get("dims", None))
```

michaelosthege: No, but it also makes no difference.
Since I don't have the branch in front of me, could we go ahead with the current version?
Seems overkill to rebase, push and run the CI for this..

Unless you have another suggestion, of course

pymc/gp/util.py
```diff
@@ -88,29 +88,6 @@ def replace_with_values(vars_needed, replacements=None, model=None):
...
-    n_points: None or int
-        The number of rows of `X`.  If `None`, the number of rows of `X` is
-        calculated from `X` if possible.
-    """
-    if n_points is None:
-        try:
-            n_points = int(X.shape[0])
-        except TypeError:
-            raise TypeError("Cannot infer 'shape', provide as an argument")
-    return n_points
```
bwengals, line 111: If I recall, the purpose of this function was to catch errors where input `X` was a theano tensor or PyMC random variable, wherever shape was symbolic.  If so, the user would have to pass the shape in as an arg.  With the new Aesara, will `np.shape` work alright here, and make `infer_size` not necessary?

pymc/gp/gp.py
bwengals: I think its just `pop` will mutate kwargs and `get` wont.  I'm in favor of `get`, because `v` right above `f` should also be changed to also get passed `dims` if the user provides it.

bwengals (changes requested):
Thanks for adding this @michaelosthege, it really needs it.

pymc/gp/util.py
michaelosthege: `np.shape(tensor)` gives the `tensor.shape` symbolically and these days it's completely fine to have a symbolic `size` too

pymc/gp/gp.py
michaelosthege: The `Normal` already gets the `dims`, because if they're passed they will be in the `**kwargs`.
Mutating the kwargs won't matter, because the method ends in the line after.

But I can change it if it makes you feel better ;)

pymc/gp/util.py
bwengals: wow, amazing.

pymc/gp/gp.py
bwengals: Ah gotcha, if both `f` and `v` get dims then I feel great

bwengals (approved):

pymc/gp/gp.py
ricardoV94: There's no reason to pop, even if right now the function call ends here. It might be extended in the future and devs will have to investigate if there was any reason for the pop in the first place.

michaelosthege: I changed it to a `.get()` and added a commit to pin the blackjax version to fix the CI.

PR #5805, opened 2022-05-25
Limit model graph nodes to named model variables

michaelosthege:
Some nodes such as `NoneConst` in an Aesara graph are named, but aren't model variables
that the user is interested in.

Prior to this fix, the model graph would print nodes like this:

grafik


larryshamalama (approved):
Thanks for catching this. After a quick search, the `NoneConst` named variable comes from [this line](https://github.com/aesara-devs/aesara/blob/main/aesara/tensor/type_other.py#L145).

Issue #5781, opened 2022-05-18
JAX/NUMPYRO does not include constant_data


jbh1128d1:
## When I sample larger datasets using jax (which will only sample this without error), the constant_data is not included in the inference_data output.  See below.

**example.**
```python
coords = {
    "obs_id": [0,1,2,3,4],
}
with pm.Model(coords=coords) as rugby_model:
    item_idx = pm.Data('item_idx',[0,1,2,3,4], dims="obs_id", mutable=False)
    a = pm.Normal("a", 0.0, sigma=10.0, shape=5)

    theta = a[item_idx]
    sigma = pm.HalfCauchy("error", 0.5)

    y = pm.Normal("y", theta, sigma=sigma, observed=[3,2,6,8,4])

    idata= pm.sample()
    idata_jax = pymc.sampling_jax.sample_numpyro_nuts(tune=1000, chains = 4, target_accept=0.9)
```

![image](https://user-images.githubusercontent.com/32133556/169066810-36d676ce-5c42-466e-aef6-0c2c21848c04.png)

<details><summary>Complete error traceback</summary>

```python
[The complete error output here]
```

</details>

**Please provide any additional information below.**

## Versions and main components

* PyMC/PyMC3 Version: 4.0.0b6
* Aesara/Theano Version: 2.5.1
* Python Version: 3.7.12
* Operating system: Debian via Google Cloud Platform
* How did you install PyMC/PyMC3: pip install pymc --pre (https://github.com/pymc-devs/pymc/wiki/Installation-Guide-(Linux))

ricardoV94:
Jeesh how many InferenceData groups are there xD? I confirmed that this is still missing in main. Thanks for reporting

OriolAbril:
@ricardoV94 worst case scenario below :stuck_out_tongue_winking_eye: 

![imatge](https://user-images.githubusercontent.com/23738400/169106948-21819680-d1bc-437e-ae9c-66c23a15f721.png)

twiecki:
@ricardoV94 Can you add some more info on what needs to be done here?

ricardoV94:
@OriolAbril might have a better idea, I am not very familiar with the InferenceData creation from PyMC

OriolAbril:
[...]

PR #5807, opened 2022-05-26
Add constant_data into sample_numpyro_nuts()

danhphan:
This PR add constant_data into `pymc.sampling_jax.sample_numpyro_nuts()`. It addresses the issue #5781

cc @ricardoV94 :)

Let me know if it needs to be changed. Thank you.

### DIFF
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -43,11 +43,8 @@ _log = logging.getLogger("pymc")
 Var = Any  # pylint: disable=invalid-name
 
 
-def find_observations(model: Optional["Model"]) -> Dict[str, Var]:
+def find_observations(model: "Model") -> Dict[str, Var]:
     """If there are observations available, return them as a dictionary."""
-    if model is None:
-        return {}
-
     observations = {}
     for obs in model.observed_RVs:
         aux_obs = getattr(obs.tag, "observations", None)
@@ -63,12 +60,9 @@ def find_observations(model: Optional["Model"]) -> Dict[str, Var]:
     return observations
 
 
-def find_constants(model: Optional["Model"]) -> Dict[str, Var]:
+def find_constants(model: "Model") -> Dict[str, Var]:
     """If there are constants available, return them as a dictionary."""
     # The constant data vars must be either pm.Data or TensorConstant or SharedVariable
-    if model is None:
-        return {}
-
     def is_data(name, var, model) -> bool:
         observations = find_observations(model)
         return (
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -63,6 +63,22 @@ def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
     return observations
 
 
+def find_constants(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
+    """If there are constants available, return them as a dictionary."""
+    if model is None or not model.named_vars:
+        return None
+
+    constants = {}
+    for name, var in model.named_vars.items():
+        if isinstance(var, (Constant, SharedVariable)):
+            if hasattr(var, "data"):
+                var = var.data
+            elif hasattr(var, "get_value"):
+                var = var.get_value()
+            constants[name] = var
+    return constants
+
+
 class _DefaultTrace:
     """
     Utility for collecting samples into a dictionary.
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -43,10 +43,10 @@ _log = logging.getLogger("pymc")
 Var = Any  # pylint: disable=invalid-name
 
 
-def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
+def find_observations(model: Optional["Model"]) -> Dict[str, Var]:
     """If there are observations available, return them as a dictionary."""
     if model is None:
-        return None
+        return {}
 
     observations = {}
     for obs in model.observed_RVs:
@@ -63,20 +63,37 @@ def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
     return observations
 
 
-def find_constants(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
+def find_constants(model: Optional["Model"]) -> Dict[str, Var]:
     """If there are constants available, return them as a dictionary."""
-    if model is None or not model.named_vars:
-        return None
+    # The constant data vars must be either pm.Data or TensorConstant or SharedVariable
+    if model is None:
+        return {}
 
-    constants = {}
+    def is_data(name, var, model) -> bool:
+        observations = find_observations(model)
+        return (
+            var not in model.deterministics
+            and var not in model.observed_RVs
+            and var not in model.free_RVs
+            and var not in model.potentials
+            and var not in model.value_vars
+            and name not in observations
+            and isinstance(var, (Constant, SharedVariable))
+        )
+
+    # The assumption is that constants (like pm.Data) are named
+    # variables that aren't observed or free RVs, nor are they
+    # deterministics, and then we eliminate observations.
+    constant_data = {}
     for name, var in model.named_vars.items():
-        if isinstance(var, (Constant, SharedVariable)):
-            if hasattr(var, "data"):
-                var = var.data
-            elif hasattr(var, "get_value"):
+        if is_data(name, var, model):
+            if hasattr(var, "get_value"):
                 var = var.get_value()
-            constants[name] = var
-    return constants
+            elif hasattr(var, "data"):
+                var = var.data
+            constant_data[name] = var
+
+    return constant_data
 
 
 class _DefaultTrace:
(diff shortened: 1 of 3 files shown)


danhphan:
It seems that `blackjax` failed when running on `gpu`.

On my local machine, when run tests with jax flag on `cpu` as follow, all the tests in `test_sampling_jax.py` passed:
`JAX_PLATFORM_NAME=cpu pytest pymc/tests/test_sampling_jax.py`

However, when run with `gpu` flag, `numpyro` worked well, while `blackjax` failed:

`JAX_PLATFORM_NAME=gpu pytest pymc/tests/test_sampling_jax.py`
Outputs:
```
========================================================================= short test summary info =========================================================================
FAILED pymc/tests/test_sampling_jax.py::test_transform_samples[None-sample_blackjax_nuts] - ValueError: compiling computation that requires 2 logical devices, but only ...
FAILED pymc/tests/test_sampling_jax.py::test_transform_samples[cpu-sample_blackjax_nuts] - ValueError: compiling computation that requires 2 logical devices, but only 1...
FAILED pymc/tests/test_sampling_jax.py::test_deterministic_samples[sample_blackjax_nuts] - ValueError: compiling computation that requires 2 logical devices, but only 1...
FAILED pymc/tests/test_sampling_jax.py::test_seeding[2-None-sample_blackjax_nuts] - ValueError: compiling computation that requires 2 logical devices, but only 1 XLA de...
FAILED pymc/tests/test_sampling_jax.py::test_seeding[2-123-sample_blackjax_nuts] - ValueError: compiling computation that requires 2 logical devices, but only 1 XLA dev...
========================================================== 5 failed, 22 passed, 12 warnings in 74.31s (0:01:14) ===========================================================

```
It seems a down stream error of `blackjax`?
`ValueError: compiling computation that requires 2 logical devices, but only 1 XLA devices are available (num_replicas=2, num_partitions=1)`

danhphan:
Also, any ideas why the `pre-commit` failed at `mypy`? I'm still not able to figure it out  though :) Thanks

pymc/backends/arviz.py
```diff
@@ -63,6 +63,22 @@ def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
     return observations
 
 
+def find_constants(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
```
ricardoV94: I think you needs something more strict as in this method: https://github.com/pymc-devs/pymc/blob/162ad6c3d1cc08613ba61e2851e8de253878ebf9/pymc/backends/arviz.py#L468-L503

If possible we should refactor it to be a standalone function, so that we don't duplicate the code (assuming the same code can be used here)

ricardoV94:
If you rebase from main, the failing blackjax tests should go away

danhphan:
Hi yes, agree. I will refactor the codes.

danhphan:
Hopefully it is alright now :)

pymc/backends/arviz.py
```diff
@@ -63,6 +63,43 @@ def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
...
+    """If there are constants available, return them as a dictionary."""
+    # For constant data, we are concerned only with deterministics and
+    # data. The constant data vars must be either pm.Data
+    # (TensorConstant/SharedVariable) or pm.Deterministic
```
ricardoV94: Looks like we explicitly ignore deterministic 

```suggestion
    # The constant data vars must be either pm.Data
    # or (TensorConstant/SharedVariable)
```

pymc/backends/arviz.py
```diff
@@ -63,6 +63,43 @@ def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
...
+
+    if model is None:
+        return None
```
ricardoV94: Let's return the same type of variable to keep things simpler

```suggestion
        return {}
```

pymc/backends/arviz.py
```diff
@@ -63,6 +63,43 @@ def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
...
+
+    observations = find_observations(model)
```
ricardoV94: Similarly if we make this return an empty dict we don't need to worry about the `None` case below.

pymc/backends/arviz.py
ricardoV94: Actually we should just assume a model is always passed. The caller shouldn't call `find_constants(None)`

pymc/backends/arviz.py
```diff
@@ -467,41 +504,10 @@ def observed_data_to_xarray(self):
...
-        # I don't know how to find pm.Data, except that they are named
-        # variables that aren't observed or free RVs, nor are they
-        # deterministics, and then we eliminate observations.
-        for name, var in self.model.named_vars.items():
-            if is_data(name, var):
-                constant_data_vars[name] = var
-
-        if not constant_data_vars:
+        constant_data = find_constants(self.model)
+        if constant_data is None:
```
ricardoV94:
```suggestion
        if not constant_data:
```

ricardoV94 (changes requested):
Looks good. Left some suggestions to simplify / tidy up the original functions

danhphan:
Hi, it has been updated :)

danhphan:
Hi @ricardoV94 I think it is alright for this pr to be merged :) Thanks

pymc/backends/arviz.py
```diff
@@ -43,10 +43,10 @@
 Var = Any  # pylint: disable=invalid-name
 
 
-def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
+def find_observations(model: Optional["Model"]) -> Dict[str, Var]:
     """If there are observations available, return them as a dictionary."""
     if model is None:
-        return None
```
ricardoV94, line 49: We don't need the model is none cases. Whomever calls this function should always pass a model. There's no good reason why they would call the function with None

pymc/backends/arviz.py
```diff
@@ -63,6 +63,39 @@ def find_observations(model: Optional["Model"]) -> Optional[Dict[str, Var]]:
     return observations
 
 
+def find_constants(model: Optional["Model"]) -> Dict[str, Var]:
+    """If there are constants available, return them as a dictionary."""
+    # The constant data vars must be either pm.Data or TensorConstant or SharedVariable
+    if model is None:
```
ricardoV94: Same here

ricardoV94 (commented):
Looks good, but we need to test that constant_data is showing up in test_sampling_jax.py, probably in the same test that already checks for other InferenceData groups

pymc/backends/arviz.py
danhphan: Hi, I think it better than the previous codes now. Why we did not remove it in previous review [here](https://github.com/pymc-devs/pymc/commit/fe2d101bb27e05b889eafda7e54b07e05250faee#diff-d11fbc930fa0db01efa57955e9481ca2605187656bc12a0326a44d090094698cR47)? But it could be good to have some defensive codes here, since we not 100% sure people always call it with a model. Any way, this is an easy fix, and just a minor thing.

danhphan:
> Looks good, but we need to test that constant_data is showing up in test_sampling_jax.py, probably in the same test that already checks for other InferenceData groups

Hi, I cannot find where these codes though? Can you point it out?

I have already run these following codes and check to make sure the constant_data is shown in `idata_jax`.
```
import numpy as np
import pymc as pm
import aesara as ae
import aesara.tensor as at
from pymc.sampling_jax import sample_numpyro_nuts

coords = {
    "obs_id": [0, 1, 2, 3, 4],
}
with pm.Model(coords=coords) as rugby_model:
    item_idx = pm.Data("item_idx", [0, 1, 2, 3, 4], dims="obs_id", mutable=False)
    b = ae.shared(0.1)
    obs = np.random.normal(10, 2, size=100)
    c = ae.shared(obs, borrow=True, name="obs")
    a = pm.Normal("a", 0.0, sigma=10.0, shape=5)

    theta = a[item_idx]
    sigma = pm.HalfCauchy("error", 0.5)

    y = pm.Normal("y", theta, sigma=sigma, observed=[3, 2, 6, 8, 4])

    idata = pm.sample()
    idata_jax = sample_numpyro_nuts(tune=1000, chains=4, target_accept=0.9)
```

pymc/backends/arviz.py
ricardoV94: > Why we did not remove it in previous review here? 

We just missed it then.

> But it could be good to have some defensive codes here, since we not 100% sure people always call it with a model. 

I don't see why you would call this function if not with a model. It just doesn't make sense. And if someone does it that's probably a mistake and the best thing is to let it naturally crash.

Adding special cases like this can make it more difficult to spot errors instead of helping.

michaelosthege:
> Hi, I cannot find where these codes though? Can you point it out?

I don't see an existing test that checks `pm.Data` containers with jax sampling.
Best create a new one right below this one:

https://github.com/pymc-devs/pymc/blob/ab05d443bcb1d8edc767522040782b3e2c0bafde/pymc/tests/test_sampling_jax.py#L160

But please make sure to set something like `chain=1, tune=10, draws=20` so the test runs fast.

danhphan:
Yes, I will add the test for both `observations` and `constants`, as well as remove the checking `model is None` soon. 

Just need some more time to do that :) Thanks

michaelosthege:
@danhphan let us know if someone should take over the rest here. This is one of currently two open items in the 4.0.1 milestone..

ricardoV94 (approved):

michaelosthege (approved):

danhphan:
Hi @michaelosthege Sorry I'm quite busy with other more urgent stuffs right now, so please feel free to go ahead with other steps. I will not able to do it in early this week. If no-one updates this, I will try to spend sometime this weekend to complete it. Thanks you.

ricardoV94:
Thanks @danhphan and @michaelosthege!

PR #5808, opened 2022-05-26
Fallback to Aeppl when input does not have a dispatched logp

ricardoV94:
This (and any other simple / single variable Aeppl derivations) will now work:

```python
import pymc as pm
pm.logp(pm.math.exp(pm.Normal.dist()), 5)  # Equivalent to `pm.logp(pm.Lognormal(), 5)
```

Because of how Aeppl works, unlike the direct `logp`, the value has to be of the same type as the RV, meaning we can't do something like:

```python
pm.logp(pm.math.exp(pm.Normal.dist(), [1, 2])
```

But must instead do:

```python
pm.logp(pm.math.exp(pm.Normal.dist(size=2), [1, 2])
```


ricardoV94:
Somewhat related to https://github.com/pymc-devs/pymc/discussions/5801

ricardoV94:
Getting a strange failure in the MacOs run:

image

https://github.com/pymc-devs/pymc/runs/6608218080?check_suite_focus=true

michaelosthege:
> Getting a strange failure in the MacOs run:
> 
> image
> 
> https://github.com/pymc-devs/pymc/runs/6608218080?check_suite_focus=true

I had the same error in #5804 and it went away after re-running just that one job.

twiecki (approved):

michaelosthege (approved):
TIL one should `raise from`.

PR #5721, opened 2022-04-20
[DOC] PyMC, Aesara and  Aeppl intro notebook

juanitorduz:
This PR partially solves https://github.com/pymc-devs/pymc/issues/5538. 

I am trying to add comments and understand @ricardoV94's nice intro notebook. Nevertheless, I still have a long way to fully understand. Specifically, starting from section `Graph manipulation 102` things become less clear for me. Still, as suggested I am opening this PR to get feedback and iterate.

PR #5810, opened 2022-05-26
Update dev guide

### LINKED ISSUE #5346: Developer guide needs to be updated for V4

michaelosthege:
I went over the files under `/docs/source/contributing/*` except the ones that had an `:orphan:` line at the top.

I wrote the instructions on running the test suite and fixed typos in several places.

For the guide on implementing distributions I updated it to the best of my knowledge, but here @ricardoV94 should fact-check me.

Closes #5346

-----

Some remarks that generalize to other pages:
* I changed the formatting of "Project" and "`packages`" to be consistent w.r.t. capitalization and style:
  * NumPy or `numpy`
  * SciPy or `scipy`
  * PyMC or `pymc`
  * Aesara or `aesara`
* Instead of "PyMC `4.x`" we should write "PyMC `>=4.0.0`" so we don't have to update these lines with the next major release.
* cc @OriolAbril I would generally prefer a "one line per sentence" rule in the docs sources, because it makes the `git diff` a lot easier to work with. (Only double line-breaks become line breaks in the rendered output.)


OriolAbril:
> cc @OriolAbril I would generally prefer a "one line per sentence" rule in the docs sources, because it makes the git diff a lot easier to work with. (Only double line-breaks become line breaks in the rendered output.)

That is perfectly fine with me, I also prefer this style generally

> except the ones that had an :orphan: line at the top.
>
> Closes https://github.com/pymc-devs/pymc/issues/5346

To close the issue all pages should be updated, independently of having `:orphan:` or not. afaik, all pages are currently present in one of the multiple toctrees and show in the left sidebar, the orphan tags are inherited and left in the pages that still have not been touched yet. Most relevant on this and to close this issue is https://docs.pymc.io/en/latest/contributing/developer_guide.html.

docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -86,10 +86,10 @@ blah = BlahRV()
...
+1. `Aesara` tries to infer the output shape of the `RandomVariable` (given a user-specified size) by introspection of the `ndim_supp` and `ndim_params` attributes. However, the default method may not work for more complex distributions. In that case, custom `_supp_shape_from_params` (and less probably, `_infer_shape`) should also be implemented in the new `RandomVariable` class. One simple example is seen in the {class}`~pymc.distributions.multivariate.DirichletMultinomialRV` where it was necessary to specify the `rep_param_idx` so that the `default_supp_shape_from_params` helper method can do its job. In more complex cases, it may not suffice to use this default helper. This could happen for instance if the argument values determined the support shape of the distribution, as happens in the `~pymc.distribution.multivarite._LKJCholeskyCovRV`.
+1. It's okay to use the `rng_fn` `classmethods` of other Aesara, aeppl and PyMC `RandomVariables` inside the new `rng_fn`. For example if you are implementing a negative HalfNormal `RandomVariable`, your `rng_fn` can simply return `- halfnormal.rng_fn(rng, scale, size)`.
```
ricardoV94:
```suggestion
1. It's okay to use the `rng_fn` `classmethods` of other Aesara and PyMC `RandomVariables` inside the new `rng_fn`. For example if you are implementing a negative HalfNormal `RandomVariable`, your `rng_fn` can simply return `- halfnormal.rng_fn(rng, scale, size)`.
```

Aeppl doesn't implement any RVs

ricardoV94 (changes requested):
The developer distribution part looks good, just an issue with mentioning aeppl

michaelosthege:
@ricardoV94 I updated the beginning of the `developer_guide.rst` to something that's hopefully correct and more accessible to people without a strong math background (please check).

I stopped where it got more detailed on `logp`. Can you take over from there?

Large parts of that document are written in the "I" perspective and talk about things that are no longer relevant.

Generally, I would prefer to delete large sections instead of keeping things that are outdated or misleading.

We could also move some submodule-specific developer documentation such as the "Dynamic HMC" or "Variational Inference" sections into `README.md` files that are located in the corresponding subfolders.
Maybe that would help to keep them aligned with the implementation while keeping the docs focused on user API ?

docs/source/contributing/developer_guide.rst
```diff
@@ -7,146 +7,87 @@
 PyMC Developer Guide
```
OriolAbril: We should get rid of the orphan thing and the comment, and add a target right before the title so we can reference this page easily. In rST targets are

```
.. _developer_guide:
```

docs/source/contributing/developer_guide.rst
```diff
@@ -7,146 +7,87 @@
...
-functions. Instead, to access a stateless distribution, you need to call
-``pm.SomeDistribution.dist(...)`` or ``RV.dist`` *after* you initialized
-``RV`` in a model context (see
-https://docs.pymc.io/Probability\_Distributions.html#using-pymc-distributions-without-a-model).
+In PyMC, model variables are defined by calling probability distribution classes with parameters:
 
-With this distinction in mind, we can take a closer look at the
-stateless distribution part of pymc (see distribution api in :ref:`doc <api_distributions>`), which divided into:
+.. code:: python
+    x = Normal('x', 0, 1)
```
OriolAbril:
```suggestion
.. code-block::

    x = Normal('x', 0, 1)
```

Main comment is the space is needed, otherwise the code isn't shown in the rendered page.

I think the ideal scenario is using the code block directive. The default syntax highlight is python so unless we use code blocks in other languages it doesn't need to be specified. Not 100% sure why the code thing alone works as it should be a role (aka inline formatting) instead of a directive.

ref: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block and https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-code

docs/source/contributing/developer_guide.rst
```diff
@@ -7,146 +7,87 @@
...
+This is done inside the context of a ``pm.Model``, which intercepts some information, for example to capture known dimensions.
+The notation aligns with the typically used math notation:
 
-- Discrete
+.. math::
+    x \sim \text{Normal}(0, 1)
 
-- Multivariate
+A call to a :class:`~pymc.Distribution` constructor as shown above returns an Aesara :class:`~aesara.tensor.TensorVariable`, which is a symbolic representation of the model variable and the graph of inputs it depends on.
+Under the hood, the variables are created through the :method:`~pymc.Distribution.dist` API, which calls the :class:`~aesara.tensor.random.basic.RandomVariable` :class:`~aesara.graph.op.Op` corresponding to the distribution.
```
OriolAbril:
```suggestion
Under the hood, the variables are created through the :meth:`~pymc.Distribution.dist` API, which calls the :class:`~aesara.tensor.random.basic.RandomVariable` :class:`~aesara.graph.op.Op` corresponding to the distribution.
```

docs/source/contributing/developer_guide.rst
```diff
@@ -7,146 +7,87 @@
...
 
-- Mixture
+At a high level of abstraction, the idea behind ``RandomVariable`` ``Op``s is to create symbolic variables (``TensorVariable``s) that can be associated with the properties of a probability distribution.
```
OriolAbril: adding a letter after a closing set of two backticks without a space seems to break rendering

docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -1,6 +1,6 @@
 # Implementing a Distribution
```
OriolAbril: the import paths need to be fixed here in multiple places. i.e. `pymc.Distribution` instead of `pymc.distributions.Distribution`

docs/source/contributing/developer_guide_implementing_distribution.md
```diff
@@ -23,7 +23,7 @@ This guide does not attempt to explain the rationale behind the `Distributions`
 {class}`~aesara.tensor.random.op.RandomVariable` are responsible for implementing the random sampling methods, which in version 3 of PyMC used to be one of the standard `Distribution` methods, alongside `logp` and `logcdf`.
 The `RandomVariable` is also responsible for parameter broadcasting and shape inference.
 
-Before creating a new `RandomVariable` make sure that it is not offered in the {mod}`Numpy library <numpy.random>`.
+Before creating a new `RandomVariable` make sure that it is not already offered in the {mod}`NumPy library <numpy.random>`.
 If it is, it should be added to the {doc}`Aesara library <aesara>` first and then imported into the PyMC library.
```
OriolAbril:
```suggestion
If it is, it should be added to the {doc}`Aesara library <aesara:index>` first and then imported into the PyMC library.
```

OriolAbril (commented):
> We could also move some submodule-specific developer documentation such as the "Dynamic HMC" or "Variational Inference" sections into README.md files that are located in the corresponding subfolders.
Maybe that would help to keep them aligned with the implementation while keeping the docs focused on user API ?

I always advocate for having documentation in the docs. Pages outside the "contributing" navbar section have users as target audience, pages inside this section have developers and contributors as target audience. IMO, having the docs close or even in the same file has a negligible effect on keeping them updated when compared to making sure this is done as part of the review process. i.e. even docstrings that are as close as possible to the implementation still end up outdated quite often

cc @danhphan I think this should fix some of the issues you mentioned on the running tests page. Is there something else you think is missing?

docs/source/contributing/developer_guide.rst
michaelosthege: I converted the entire page into Markdown to make it easier to format..

docs/source/contributing/developer_guide.rst
michaelosthege: it works now that it's all in markdown

michaelosthege:
I formatted the dev guide as a Markdown file to make it easier to update.

I only dealt with the beginning, but I wont have the time to continue with the rest.
Therefore I would recommend to (squash) merge the changes I made so far and continue updating the dev guide in the coming days.

danhphan:
Hi @OriolAbril I think it looks good now. Thanks mate :)

twiecki:
Can we merge this?

ricardoV94:
The changes to the dev guide look good. We will probably want to merge this with the new document that is being added in #5721 but for now we can keep both.

ricardoV94:
Nevermind that file is still too outdated...: https://pymc--5810.org.readthedocs.build/en/5810/contributing/developer_guide.html?highlight=developer

twiecki:
Why is this not a NB?

twiecki:
Maybe we'll remove it for now?

ricardoV94:
> Maybe we'll remove it for now?

Let's keep the file around but not show it in the docs? Is this possible @OriolAbril 
I'll branch of this PR with the remaining commits, as those seem still useful.

OriolAbril:
> Let's keep the file around but not show it in the docs? Is this possible @OriolAbril

Yes, it is possible, we just need to add

```
---
orphan: true
---
```

at the top of the file and remove it from the toctree in https://github.com/pymc-devs/pymc/blob/main/docs/source/contributing/index.md which seems to have happened already

michaelosthege:
@OriolAbril can you fix the precommit here?

The passage is written to literally have the URLs in there, but the pre-commit doesn't allow it.

OriolAbril:
> @OriolAbril can you fix the precommit here?
>
> The passage is written to literally have the URLs in there, but the pre-commit doesn't allow it.

Updated the doc (that I copied over unmodified not long ago while fixing the website and navbar) and added the file to the pre-commit exclusion for the cross-reference task. Here the urls must be present, so it needs to be excluded from that check

michaelosthege:
Ok, so the page is now orphaned, and git conflicts were resolved.

Can we merge now? Because this PR adds new content on contributing pages that were previously filled with ToDo placeholders.

OriolAbril (approved):

PR #5814, opened 2022-05-28
Incorporating AePPL into `GaussianRandomWalk`

### LINKED ISSUE #5762: Incorporate AePPL into GaussianRandomWalk
A recent PR #5298 added GaussianRandomWalk distribution to v4. It would be possible to use AePPL to obtain the Aesara graph corresponding to desired `rv_op` and retrieve the correct log-probability graph. The use of AePPL here would be desirable as we will no longer need to implement a corresponding `RandomVariable` class nor implement the logp method. Moreover, it would generalize the random walk to make use of non-Gaussian distributions.

cc @ricardoV94

larryshamalama:
Closes #5762.

This PR removes `GaussianRandomWalkRV` altogether by defining an `rv_op` as a cumulative sum of distributions. The most recent version of AePPL, i.e. 0.0.31, is now able to retrieve the appropriate logp graph from a `CumOp`.

This is a WIP for now as I figure out how to use AePPL...

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -13,7 +13,7 @@
 #   limitations under the License.
 import warnings
 
-from typing import Any, Optional, Tuple, Union
+from typing import Any, Optional, Union
 
 import aesara
 import aesara.tensor as at
@@ -28,22 +28,15 @@ from aesara.graph.basic import Node
 from aesara.raise_op import Assert
 from aesara.tensor import TensorVariable
 from aesara.tensor.basic_opt import ShapeFeature, topo_constant_folding
+from aesara.tensor.extra_ops import CumOp
 from aesara.tensor.random.op import RandomVariable
-from aesara.tensor.random.utils import normalize_size_param
 
 from pymc.aesaraf import change_rv_size, convert_observed_data, floatX, intX
 from pymc.distributions import distribution, multivariate
 from pymc.distributions.continuous import Flat, Normal, get_tau_sigma
-from pymc.distributions.dist_math import check_parameters
 from pymc.distributions.distribution import SymbolicDistribution, _moment, moment
 from pymc.distributions.logprob import ignore_logprob, logp
-from pymc.distributions.shape_utils import (
-    Dims,
-    Shape,
-    convert_dims,
-    rv_size_is_none,
-    to_tuple,
-)
+from pymc.distributions.shape_utils import Dims, Shape, convert_dims, to_tuple
 from pymc.model import modelcontext
 from pymc.util import check_dist_not_registered
 
@@ -114,110 +107,8 @@ def get_steps(
     return inferred_steps
 
 
-class GaussianRandomWalkRV(RandomVariable):
-    """
-    GaussianRandomWalk Random Variable
-    """
-
-    name = "GaussianRandomWalk"
-    ndim_supp = 1
-    ndims_params = [0, 0, 0, 0]
-    dtype = "floatX"
-    _print_name = ("GaussianRandomWalk", "\\operatorname{GaussianRandomWalk}")
-
-    def make_node(self, rng, size, dtype, mu, sigma, init_dist, steps):
-        steps = at.as_tensor_variable(steps)
-        if not steps.ndim == 0 or not steps.dtype.startswith("int"):
-            raise ValueError("steps must be an integer scalar (ndim=0).")
-
-        mu = at.as_tensor_variable(mu)
-        sigma = at.as_tensor_variable(sigma)
-        init_dist = at.as_tensor_variable(init_dist)
-
-        # Resize init distribution
-        size = normalize_size_param(size)
-        # If not explicit, size is determined by the shapes of mu, sigma, and init
-        init_dist_size = (
-            size if not rv_size_is_none(size) else at.broadcast_shape(mu, sigma, init_dist)
-        )
-        init_dist = change_rv_size(init_dist, init_dist_size)
-
-        return super().make_node(rng, size, dtype, mu, sigma, init_dist, steps)
-
-    def _supp_shape_from_params(self, dist_params, reop_param_idx=0, param_shapes=None):
-        steps = dist_params[3]
-
-        return (steps + 1,)
-
-    @classmethod
-    def rng_fn(
-        cls,
-        rng: np.random.RandomState,
-        mu: Union[np.ndarray, float],
-        sigma: Union[np.ndarray, float],
-        init_dist: Union[np.ndarray, float],
-        steps: int,
-        size: Tuple[int],
-    ) -> np.ndarray:
-        """Gaussian Random Walk generator.
-
-        The init value is drawn from the Normal distribution with the same sigma as the
-        innovations.
-
-        Notes
-        -----
-        Currently does not support custom init distribution
-
-        Parameters
-        ----------
-        rng: np.random.RandomState
-           Numpy random number generator
-        mu: array_like of float
-           Random walk mean
-        sigma: array_like of float
-            Standard deviation of innovation (sigma > 0)
-        init_dist: array_like of float
-            Initialization value for GaussianRandomWalk
-        steps: int
-            Length of random walk, must be greater than 1. Returned array will be of size+1 to
-            account as first value is initial value
-        size: tuple of int
-            The number of Random Walk time series generated
-
-        Returns
-        -------
-        ndarray
-        """
-
-        if steps < 1:
-            raise ValueError("Steps must be greater than 0")
-
-        # If size is None then the returned series should be (*implied_dims, 1+steps)
-        if size is None:
-            # broadcast parameters with each other to find implied dims
-            bcast_shape = np.broadcast_shapes(
-                np.asarray(mu).shape,
-                np.asarray(sigma).shape,
-                np.asarray(init_dist).shape,
-            )
-            dist_shape = (*bcast_shape, int(steps))
-
-        # If size is None then the returned series should be (*size, 1+steps)
-        else:
-            dist_shape = (*size, int(steps))
-
-        # Add one dimension to the right, so that mu and sigma broadcast safely along
-        # the steps dimension
-        innovations = rng.normal(loc=mu[..., None], scale=sigma[..., None], size=dist_shape)
-        grw = np.concatenate([init_dist[..., None], innovations], axis=-1)
-        return np.cumsum(grw, axis=-1)
-
-
-gaussianrandomwalk = GaussianRandomWalkRV()
-
-
-class GaussianRandomWalk(distribution.Continuous):
-    r"""Random Walk with Normal innovations.
+class GaussianRandomWalk(SymbolicDistribution):
+    r"""Random Walk with Normal innovations
 
     Parameters
     ----------
@@ -235,8 +126,6 @@ class GaussianRandomWalk(distribution.Continuous):
         Number of steps in Gaussian Random Walk (steps > 0). Only needed if shape is not
         provided.
     """
-
-    rv_op = gaussianrandomwalk
 
     def __new__(cls, *args, steps=None, **kwargs):
         steps = get_steps(
@@ -269,6 +158,8 @@ class GaussianRandomWalk(distribution.Continuous):
                 FutureWarning,
             )
             init_dist = kwargs.pop("init")
+        if not steps.ndim == 0:
+            raise ValueError("steps must be an integer scalar (ndim=0).")
 
         # If no scalar distribution is passed then initialize with a Normal of same mu and sigma
         if init_dist is None:
@@ -293,37 +184,65 @@ class GaussianRandomWalk(distribution.Continuous):
 
         return super().dist([mu, sigma, init_dist, steps], **kwargs)
 
-    def moment(rv, size, mu, sigma, init_dist, steps):
-        grw_moment = at.zeros_like(rv)
-        grw_moment = at.set_subtensor(grw_moment[..., 0], moment(init_dist))
-        # Add one dimension to the right, so that mu broadcasts safely along the steps
-        # dimension
-        grw_moment = at.set_subtensor(grw_moment[..., 1:], mu[..., None])
-        return at.cumsum(grw_moment, axis=-1)
+    @classmethod
+    def ndim_supp(cls, *args):
+        return 1
 
-    def logp(
-        value: at.Variable,
-        mu: at.Variable,
-        sigma: at.Variable,
-        init_dist: at.Variable,
-        steps: at.Variable,
-    ) -> at.TensorVariable:
-        """Calculate log-probability of Gaussian Random Walk distribution at specified value."""
+    @classmethod
+    def rv_op(cls, mu, sigma, init_dist, steps, size=None):
+        # If not explicit, size is determined by the shapes of mu, sigma, and init
+        if size is not None:
+            # we have all the information regarding size from users or .dist()
+            init_size = size
+        else:
+            # we infer size from parameters
+            init_size = at.broadcast_shape(
+                mu,
+                sigma,
+                init_dist,
+            )
 
-        # Calculate initialization logp
-        init_logp = logp(init_dist, value[..., 0])
+        # TODO: extend for multivariate init
+        init_dist = change_rv_size(init_dist, init_size)
+        innovation_dist = Normal.dist(mu[..., None], sigma[..., None], size=(*init_size, steps))
+        rv_out = at.cumsum(at.concatenate([init_dist[..., None], innovation_dist], axis=-1), axis=-1)
 
-        # Make time series stationary around the mean value
-        stationary_series = value[..., 1:] - value[..., :-1]
-        # Add one dimension to the right, so that mu and sigma broadcast safely along
-        # the steps dimension
-        series_logp = logp(Normal.dist(mu[..., None], sigma[..., None]), stationary_series)
+        rv_out.tag.mu = mu
+        rv_out.tag.sigma = sigma
+        rv_out.tag.init_dist = init_dist
+        rv_out.tag.innovation_dist = innovation_dist
+        rv_out.tag.steps = steps
+        rv_out.tag.is_grw = True  # for moment dispatching
 
-        return check_parameters(
-            init_logp + series_logp.sum(axis=-1),
-            steps > 0,
-            msg="steps > 0",
+        return rv_out
+
+    @classmethod
+    def change_size(cls, rv, new_size, expand=False):
+        if expand:
+            new_size = at.concatenate([new_size, rv.shape[:-1]])
+
+        return cls.rv_op(
+            mu=rv.tag.mu,
+            sigma=rv.tag.sigma,
+            init_dist=rv.tag.init_dist,
+            steps=rv.tag.steps,
+            size=new_size,
         )
+
+
+@_moment.register(CumOp)
+def moment_grw(op, rv, dist_params):
+    """
+    This moment dispatch is currently only applicable for a GaussianRandomWalk.
+    TODO: Encapsulate GRW graph in an OpFromGraph so that we can dispatch
+     the moment directly on it
+    """
+    if not getattr(rv.tag, "is_grw", False):
+        raise NotImplementedError("Moment not implemented for `CumOp`")
+    init_dist = rv.tag.init_dist
+    innovation_dist = rv.tag.innovation_dist
+    grw_moment = at.concatenate([moment(init_dist)[..., None], moment(innovation_dist)], axis=-1)
+    return at.cumsum(grw_moment, axis=-1)
 
 
 class AutoRegressiveRV(OpFromGraph):
--- a/pymc/tests/test_distributions_moments.py
+++ b/pymc/tests/test_distributions_moments.py
@@ -28,6 +28,7 @@ from pymc.distributions import (
     Exponential,
     Flat,
     Gamma,
+    GaussianRandomWalk,
     Geometric,
     Gumbel,
     HalfCauchy,
@@ -1505,4 +1506,44 @@ def test_lkjcholeskycov_moment(n, eta, size, expected):
 def test_dirichlet_multinomial_moment(a, n, size, expected):
     with Model() as model:
         DirichletMultinomial("x", n=n, a=a, size=size)
+    assert_moment_is_expected(model, expected)
+
+
+even_numbers = np.insert(np.cumsum(np.tile(2, 10)), 0, 0)  # for test_gaussianrandomwalk below
+
+
+@pytest.mark.parametrize(
+    "mu, sigma, init_dist, steps, size, expected",
+    [
+        (0, 3, StudentT.dist(5), 10, None, np.zeros(11)),
+        (Normal.dist(2, 3), Gamma.dist(1, 1), StudentT.dist(5), 10, None, even_numbers),
+        (
+            Normal.dist(2, 3, size=(3, 5)),
+            Gamma.dist(1, 1),
+            StudentT.dist(5),
+            10,
+            None,
+            np.broadcast_to(even_numbers, shape=(3, 5, 11)),
+        ),
+        (
+            Normal.dist(2, 3, size=(3, 1)),
+            Gamma.dist(1, 1, size=(1, 5)),
+            StudentT.dist(5),
+            10,
+            (3, 5),
+            np.broadcast_to(even_numbers, shape=(3, 5, 11)),
+        ),
+        (
+            Normal.dist(2, 3),
+            Gamma.dist(1, 1),
+            StudentT.dist(5),
+            10,
+            (3, 5),
+            np.broadcast_to(even_numbers, shape=(3, 5, 11)),
+        ),
+    ],
+)
+def test_gaussianrandomwalk(mu, sigma, init_dist, steps, size, expected):
+    with Model() as model:
+        GaussianRandomWalk("x", mu=mu, sigma=sigma, init_dist=init_dist, steps=steps, size=size)
     assert_moment_is_expected(model, expected)
--- a/pymc/tests/test_distributions_timeseries.py
+++ b/pymc/tests/test_distributions_timeseries.py
@@ -21,7 +21,7 @@ from aesara.tensor import TensorVariable
 import pymc as pm
 
 from pymc.aesaraf import floatX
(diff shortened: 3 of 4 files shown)


twiecki:
Nice! However, this isn't the same thing, it's a different parameterization of the same thing. Like centered vs non-centered. So I don't think we should remove the previous one as people might be using it and then changing the parameterization might screw with the inference.

However, we could add it either as an option or a separate dist.

ricardoV94:
> Nice! However, this isn't the same thing, it's a different parameterization of the same thing. Like centered vs non-centered. So I don't think we should remove the previous one as people might be using it and then changing the parameterization might screw with the inference.
> 
> However, we could add it either as an option or a separate dist.

What do you mean? This should be completely equivalent to what we had before.

ricardoV94:
You might be thinking of the latent parametrizations but that was not what GRW did before (as that wouldn't have been a vanilla distribution)

Something like that would just require implementing a default (diff) transform which would definitely be helpful for random walks.

Alternatively it would be useful to allow different implementations between unobserved and observed variables. Either eagerly at creation time or, even better, lazily when compiling the logp, but the latter requires some more work (make sure we have the right deterministics).

twiecki:
Isn't this changing the previous parameterization: $x_t \sim \mathcal{N}(x_{t-1}, \sigma)$ to $x_t \sim x_{t-1} + \mathcal{N}(0, \sigma)$

ricardoV94:
> Isn't this changing the previous parameterization: xt∼N(xt−1,σ) to xt∼xt−1+N(0,σ)

Where would the parametrization be changed? 

The random generator method is exactly the same as before if you check what was going on in the old `rng_fn`. The logp is exactly the same as before (otherwise this test would fail): https://github.com/pymc-devs/pymc/blob/5703a9d24ca01da8b81363874e023e59cd18688d/pymc/tests/test_distributions.py#L2609-L2624

The  expression $x_{t-1}+N(0,σ)$ "does not have a logp", it always has to be converted to $N(x_{t-1},σ)$, which Aeppl does when it sees it or, in this case, when it sees $x \sim \text{cumsum}(N(0, σ))$

Now for NUTS sampling (when we have unobserved GRW) it would probably be better if we were sampling in the latent space $x_{\text{raw}} \sim N(0, σ)$ and only then did the deterministic mapping to $x = \text{cumsum}(x_{\text{raw}})$, but this was not done before nor is it now.

For that we would need to either: 1) Add a new type of transform or 2) Have a way to create different graphs depending on whether the variables are observed or not.

twiecki:
I see, so there is no change from this implementation to the previous in v4, but it is different compared to v3, right?

ricardoV94:
> I see, so there is no change from this implementation to the previous in v4, but it is different compared to v3, right?

No, V3 did the same as well. The logp is written a bit different, but it's equivalent to the manual logp we had here in V4 up to this PR:
https://github.com/pymc-devs/pymc/blob/ed74406735b2faf721e7ebfa156cc6828a5ae16e/pymc3/distributions/timeseries.py#L227-L246

And there was no special transformation either.

twiecki:
Oh right, I read you previous comment again. It's always been $x_t \sim N(x_{t-1}, \sigma)$, then this makes sense 👍

pymc/distributions/timeseries.py
```diff
@@ -292,37 +193,62 @@ def dist(cls, mu=0.0, sigma=1.0, *, init_dist=None, steps=None, **kwargs) -> at.
...
+        return rv_out
+
+
+    @classmethod
+    def change_size(cls, rv, new_size, expand=False):
+        pass
+
+
+@_moment.register(CumOp)
+def moment_cumop(op, rv, dist_params):
```
larryshamalama: The random graph is shown below as a comment in this moment function. Is it normal that, to retrieve distribution parameters, I have to go up the graph a bit? If this is alright, related tests will need to be adapted with respect to this.

larryshamalama (commented):
cc @ricardoV94 about a question regarding retrieving `dist_params` from the random graph

pymc/distributions/timeseries.py
ricardoV94: Yes this is fine but we might have to start being a bit more careful because other symbolic dists may also use cumop as the outer node.

Specially when we allow users to use arbitrary graphs as variables.

For now we should at least check that the graph inside the cumop has the structure we expect from the GRW and raise NotImplementedError otherwise. In particular check we have 1) a cumop with axis=-1, 2) an inner join of two variables, 3) the second of which has the Dimshuffle on the last dimension.

pymc/tests/test_distributions_timeseries.py
```diff
@@ -223,27 +207,27 @@ def test_gaussian_random_walk_init_dist_logp(self, init):
...
-                (4, 2),
-                np.full((4, 2, 11), np.vstack((np.arange(11), -np.arange(11)))),
-            ),
-        ],
-    )
-    def test_moment(self, mu, sigma, init_dist, steps, size, expected):
-        with Model() as model:
-            GaussianRandomWalk("x", mu=mu, sigma=sigma, init_dist=init_dist, steps=steps, size=size)
-        assert_moment_is_expected(model, expected)
+    # @pytest.mark.parametrize(
```
ricardoV94: Why does this test fail now?

larryshamalama: I forgot to uncomment it for the commit. The `init_dist` had to be fetched by going up the graph a bit and now the tests are passing :)

pymc/distributions/timeseries.py
larryshamalama: How can I check the axis on which an Op (here `CumOp` or `DimShuffle`) has been performed?

canyon289:
cc @nicospinu for GSOC reference

This PR will end up being a good reference for a side by side comparison of how time series work when implemented in PyMC and how they'll be implemented if more integrated with AePPL

larryshamalama:
To my knowledge, three sets of tests are failing and they are definitely worth looking into.

- (Static?) shape checks
- `pm.logp` yields component-wise log-probability rather than a sum, which seems to be the case in the previous API
- Inferring step sizes from shapes. Would this be possible with AePPL?

Would these issues be important to address? To my knowledge, some of these, especially static shape inference, would take substantial work to be included in Aesara. Happy to hear thoughts about this

Uncommenting the error pertaining to unexpected rv nodes [here](https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/logprob.py#L247-L252) still yields many errors, so this would be important to look into

ricardoV94:
> To my knowledge, three sets of tests are failing and they are definitely worth looking into.
> 
> - (Static?) shape checks

Can you expand?

> - `pm.logp` yields component-wise log-probability rather than a sum, which seems to be the case in the previous API

Yes, that behaves differently, but I don't think it poses a significant problem. We can just change the expectation of the tests.

> - Inferring step sizes from shapes. Would this be possible with AePPL?

That should work the same way as before. We do that in AR which is also a symbolic dist. It happens before any call to rv_op. It does not interfere or depend on Aeppl

larryshamalama:
> > To my knowledge, three sets of tests are failing and they are definitely worth looking into.
> > 
> > * (Static?) shape checks
> 
> Can you expand?

Some tests that checked for shape inference were failing. They seem to have gone away whether this is related or not to the implemented `change_size`. However, it is incorrectly implemented. Can you have a look at this? Notably, `test_shape_ellipsis` is failing. I will tag you in a review.

> 
> > * `pm.logp` yields component-wise log-probability rather than a sum, which seems to be the case in the previous API
> 
> Yes, that behaves differently, but I don't think it poses a significant problem. We can just change the expectation of the tests.

In the tests, I summed up the `logp` terms, for now.

pymc/distributions/timeseries.py
```diff
@@ -292,37 +187,80 @@ def dist(cls, mu=0.0, sigma=1.0, *, init_dist=None, steps=None, **kwargs) -> at.
...
+        rv_out = at.cumsum(at.concatenate([init_dist, innovation_dist], axis=-1), axis=-1)
+        rv_out.tag.mu = mu
+        rv_out.tag.sigma = sigma
+        rv_out.tag.init_dist = init_dist.owner.inputs[0]
+        rv_out.tag.steps = steps
+
+        return rv_out
+
+    @classmethod
+    def change_size(cls, rv, new_size, expand=False):
```
larryshamalama, line 220: Two tests are not performing as expected. I don't think that the logic with `expand` is working. In the following code:

```python
with pm.Model(coords={"batch": range(5), "steps": range(20)}):
    x = pm.GaussianRandomWalk("x", dims=("batch", "steps"))
print(x.shape.eval())
```
yields `array([ 5, 20, 20])`.

pymc/tests/test_distributions_timeseries.py
```diff
@@ -147,33 +130,37 @@ def test_gaussianrandomwalk_inference(self):
...
 
         grw = pm.GaussianRandomWalk.dist(mu=np.zeros((3, 1)), sigma=[1, 1], steps=1, init_dist=init)
-        assert tuple(grw.owner.inputs[-2].shape.eval()) == (3, 2)
+        assert tuple(get_init_dist(grw).shape.eval()) == (3, 2)
 
     def test_shape_ellipsis(self):
         grw = pm.GaussianRandomWalk.dist(
             mu=0, sigma=1, steps=5, init_dist=pm.Normal.dist(), shape=(3, ...)
         )
         assert tuple(grw.shape.eval()) == (3, 6)
```
larryshamalama, line 162: I am getting `grw.shape.eval() = (3, 6, 6)` with respect to the `change_size` implemented above

pymc/tests/test_distributions_timeseries.py
```diff
@@ -200,13 +187,13 @@ def test_inconsistent_steps_and_shape(self):
     def test_inferred_steps_from_dims(self):
         with pm.Model(coords={"batch": range(5), "steps": range(20)}):
             x = GaussianRandomWalk("x", dims=("batch", "steps"))
-        steps = x.owner.inputs[-1]
+        steps = x.owner.inputs[0].owner.inputs[2].owner.inputs[1][-1]
```
larryshamalama, line 190: Getting the steps parameter this way still passes the test, but the dimension is one too big to my understanding (again, c.f. the classmethod `change_size`)

larryshamalama (commented):
I believe that I'm getting closer, minus the implementation of `change_size`.

pymc/distributions/timeseries.py
```diff
@@ -235,9 +128,8 @@ class GaussianRandomWalk(distribution.Continuous):
         provided.
     """
 
-    rv_op = gaussianrandomwalk
-
     def __new__(cls, *args, steps=None, **kwargs):
+        # TODO: go lazy for now, and overwrite initval subsequently
```
ricardoV94: This is no longer relevant right?

```suggestion
```

pymc/distributions/timeseries.py
```diff
@@ -292,37 +187,80 @@ def dist(cls, mu=0.0, sigma=1.0, *, init_dist=None, steps=None, **kwargs) -> at.
...
+        rv_out.tag.sigma = sigma
+        rv_out.tag.init_dist = init_dist.owner.inputs[0]
+        rv_out.tag.steps = steps
+
+        return rv_out
+
+    @classmethod
+    def change_size(cls, rv, new_size, expand=False):
+        if expand:
+            new_size = at.concatenate([new_size, rv.shape])
```
ricardoV94: I think the problem is here, size should not include the last dimension.
```suggestion
            new_size = at.concatenate([new_size, rv.shape[:-1])
```

pymc/distributions/timeseries.py
ricardoV94: `CumOp` contains axis as part of the Op I think: `x.owner.op.axis`? `DimShuffle` Op as a `new_order` propertiy which is a tuple of dimensions order, with "x" where new dimensions were created. A new dimension on the left of a matrix would look like `("x", 1, 0)`

```python
import aesara.tensor as at

x = at.matrix("x")
y = x[None]
y.owner.op.new_order
```

pymc/distributions/timeseries.py
larryshamalama: It works :) The `else` statement had to be removed altogether too

ricardoV94: Makes sense!

pymc/distributions/timeseries.py
larryshamalama: With the moment dispatch, I believe so, right?

pymc/tests/test_distributions_timeseries.py
larryshamalama: Done

pymc/distributions/timeseries.py
larryshamalama: Done

pymc/distributions/timeseries.py
larryshamalama: I just added a check for `CumOp` to have axis = -1. Would there be more necessary graph checks? e.g. adding dimensions to the right of `mu` and `sigma` in `innovation_dist`

pymc/distributions/logprob.py
```diff
@@ -244,12 +244,12 @@ def joint_logpt(
             and not isinstance(node.owner.op, SimulatorRV)
         )
     ]
-    if unexpected_rv_nodes:
-        raise ValueError(
-            f"Random variables detected in the logp graph: {unexpected_rv_nodes}.\n"
-            "This can happen when DensityDist logp or Interval transform functions "
-            "reference nonlocal variables."
-        )
```
larryshamalama: This part probably needs discussion...

larryshamalama:
Converted this PR back to draft. [These lines](https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/logprob.py#L247-L252) probably warrant discussion as the implementation of the `GaussianRandomWalk` wouldn't work. Other things to be done:

- Moment tests
- Test to catch `NotImplementedError` in moment dispatching of `CumOp`
- Investigating `test_gaussianrandomwalk` in `test_distributions.py`

pymc/distributions/timeseries.py
larryshamalama: Yes that worked! Done

pymc/distributions/logprob.py
larryshamalama: Nodes appear because, when `size` is not specified, `init_dist` gets broadcasted with respected to `mu`, `sigma` and `init_dist`, as such, the latter being a r.v. will appear in the `logp` graph. The presence of this random variable raises this error.

pymc/distributions/timeseries.py
```diff
@@ -292,37 +186,81 @@ def dist(cls, mu=0.0, sigma=1.0, *, init_dist=None, steps=None, **kwargs) -> at.
...
+        init_dist = change_rv_size(init_dist, init_size)[..., None]
+
+        innovation_dist = Normal.dist(mu[..., None], sigma[..., None], size=(*init_size, steps))
+
+        # TODO: extend for multivariate init
+        rv_out = at.cumsum(at.concatenate([init_dist, innovation_dist], axis=-1), axis=-1)
+        rv_out.tag.mu = mu
+        rv_out.tag.sigma = sigma
+        rv_out.tag.init_dist = init_dist.owner.inputs[0]
+        rv_out.tag.steps = steps
```
larryshamalama, line 214: Add `tag` to be the class or class name. A string would be a good idea

pymc/distributions/timeseries.py
```diff
@@ -292,37 +186,81 @@ def dist(cls, mu=0.0, sigma=1.0, *, init_dist=None, steps=None, **kwargs) -> at.
...
+        |Join [id B]
+            |mu
+            |InplaceDimShuffle{x} [id D]
+            |   |init_dist
+            |       |...
+            |innovation_dist
+
+    This moment dispatch is currently only valid for a GaussianRandomWalk.
+    """
+    try:
```
ricardoV94: Let's go lazy for now, and add a tag to the rv_out like `rv_out.tag.is_grw` and rely on that.

larryshamalama (commented):
Some notes from the meeting on Friday, June 10, 2022

pymc/distributions/logprob.py
larryshamalama: The shape can be evaluated akin to what is done in the `AR` distribution below

https://github.com/pymc-devs/pymc/blob/729f79caa68551a9c86c47ac14e623518b099835/pymc/distributions/timeseries.py#L493-L499

pymc/distributions/timeseries.py
```diff
@@ -293,37 +197,86 @@ def dist(cls, mu=0.0, sigma=1.0, *, init_dist=None, steps=None, **kwargs) -> at.
...
+            init_size = size
+        else:
+            # we infer size from parameters
+            init_size = at.broadcast_shape(
+                mu_shape,
+                sigma_shape,
+                init_dist_shape,
+                arrays_are_shapes=True,
+            )
+
```
larryshamalama, line 204: I'm not so sure if `ValueError: Random variables detected in the logp graph` arises due to inference of distribution shapes. Here is my train of thought:

- I was able to retrieve shapes via `FunctionGraph`, but the same error still occurs.
- If we specify the `size` argument, the same error occurs as well.
- In my `test_moment`, I substituted the `init_dist` by a Student T distribution and normal rvs were still present in the logp graph

Would the issue be inherent to the use of AePPL and how `joint_logp` is written in `logprob.py`? It's been a while, I may have forgotten some important details...

larryshamalama (commented):
Revisiting this PR after a hiatus. I'm not sure if the presence of random variables in the logp graph is due to retrieving distribution/rv static shapes.

pymc/distributions/timeseries.py
ricardoV94: Hmm... interesting. We need to figure out when are these RVs included in the graph then, and if they correspond to other model variables why aren't they being replaced by their value variables in aeppl.

Can you share a small snippet that raises the error?

It might also help to create the GRW manually (just with aesara.tensor calls and not pymc) and give names to the intermediate variables (cumsum, join, etc ...). You can include your manual GRW RV in a model by calling `model.register_rv`

larryshamalama: Here is a notebook reproducing the issue: https://github.com/larryshamalama/aesara-playground/blob/main/notebooks/grw-rv-logp-graph.ipynb

Am I printing the appropriate logp graph in [5]?

pymc/distributions/logprob.py
```diff
@@ -244,9 +245,16 @@ def joint_logp(
     # Only SimulatorRVs are allowed
     from pymc.distributions.simulator import SimulatorRV
 
+    temp_logp_var_dict_values = list(
+        filter(
+            lambda node: not isinstance(node.owner.op, Join),
+            temp_logp_var_dict.values(),
+        )
+    )
+
```
larryshamalama: Waiting for tests to run. Would this be an appropriate fix? Would anything else break (e.g. nodes not having a `.owner.op` attribute)?

larryshamalama (commented):
Would this be the appropriate hackish fix? 😅 It seems that the presence of random variables in the logp graph is not an AePPL problem but rather PyMC-related

larryshamalama:
Still needs a test for moments

pymc/distributions/logprob.py
ricardoV94: Pretty hackish yeah xD. If we don't do constant folding of the shapes in the joinOp logp function defined in Aeppl we can just rewrite the joinOp logp function in PyMC and do it here.

larryshamalama: I guess the bigger question is whether the development of PyMC will diverge from AePPL...

larryshamalama:
Actually, I'm realizing that we need a dispatch for the `DimShuffle` op because `moment(mu[..., None])` is returning a `NotImplementedError`. @ricardoV94 Does this sound right? This would be useful for the following model:

```python
with pm.Model() as model:
    mu = pm.Normal("mu", 2, 3)
    sigma = pm.Gamma("sigma", 1, 1)
    grw = pm.GaussianRandomWalk(name="grw", mu=mu, sigma=sigma, init_dist=pm.StudentT.dist(5), steps=10)
```

ricardoV94:
> Actually, I'm realizing that we need a dispatch for the `DimShuffle` op because `moment(mu[..., None])` is returning a `NotImplementedError`. @ricardoV94 Does this sound right? This would be useful for the following model:
> 
> ```python
> with pm.Model() as model:
>     mu = pm.Normal("mu", 2, 3)
>     sigma = pm.Gamma("sigma", 1, 1)
>     grw = pm.GaussianRandomWalk(name="grw", mu=mu, sigma=sigma, init_dist=pm.StudentT.dist(5), steps=10)
> ```

We can handle it in the same dispatch moment function, just have an if branch for that. It's the same discussion we had before of whether we specialize or create more generalized moment functions.

pymc/distributions/logprob.py
ricardoV94: For now we can just overwrite it here and later remove it if the Aeppl one is changed.

larryshamalama:
> Actually, I'm realizing that we need a dispatch for the `DimShuffle` op because `moment(mu[..., None])` is returning a `NotImplementedError`. @ricardoV94 Does this sound right? This would be useful for the following model:
> 
> ```python
> with pm.Model() as model:
>     mu = pm.Normal("mu", 2, 3)
>     sigma = pm.Gamma("sigma", 1, 1)
>     grw = pm.GaussianRandomWalk(name="grw", mu=mu, sigma=sigma, init_dist=pm.StudentT.dist(5), steps=10)
> ```

I realized that I could just do `moment(mu)[..., None]` instead 😅.

I just added tests for moments. I believe that this PR is ready to be merged, but moment tests are only passing locally so far and will be with [PR 151 in AePPL](https://github.com/aesara-devs/aeppl/pull/151) merged.

larryshamalama:
With #5955 soon to be merged, I believe that this PR would be good to go? This PR adds a hackish tweak that we don't check for random variables in the ancestors of a `Join` Op due to [issue 149 in AePPL](https://github.com/aesara-devs/aeppl/issues/149), so we can perhaps think of a more sophisticated way of dealing with this later.

ricardoV94:
> With #5955 soon to be merged, I believe that this PR would be good to go? This PR adds a hackish tweak that we don't check for random variables in the ancestors of a `Join` Op due to [issue 149 in AePPL](https://github.com/aesara-devs/aeppl/issues/149), so we can perhaps think of a more sophisticated way of dealing with this later.

I don't think we should proceed with the hack at all. It's mixing logic that should be separated. We should override aeppl join logprob instead as we have more strict requirements than they do, and do the constant fold of the shapes there.

larryshamalama:
Oh, my bad, and sounds good. It was recommended by @brandonwillard to edit a graph rewrite or register another one in the [database](https://github.com/aesara-devs/aeppl/blob/main/aeppl/opt.py#L234-L250) for shape inference. `ShapeFeature` can also be added a listener on `FunctionGraph` calls if the former is not automatically included. This would be an AePPL-centred solution. An override within PyMC would also work, I suppose

pymc/tests/test_distributions_moments.py
```diff
@@ -1506,3 +1507,43 @@ def test_dirichlet_multinomial_moment(a, n, size, expected):
...
+            Normal.dist(2, 3),
+            Gamma.dist(1, 1),
+            StudentT.dist(5),
+            10,
+            (3, 5),
+            np.broadcast_to(even_numbers, shape=(3, 5, 11)),
+        ),
+    ],
+)
+def test_gaussianrandomwalk(mu, sigma, init_dist, steps, size, expected):
```
ricardoV94, line 1546: There is already a test of moments in `test_distribution_timeseries`

pymc/tests/test_distributions_timeseries.py
```diff
@@ -147,33 +130,37 @@ def test_gaussianrandomwalk_inference(self):
     @pytest.mark.parametrize("init", [None, pm.Normal.dist()])
     def test_gaussian_random_walk_init_dist_shape(self, init):
         """Test that init_dist is properly resized"""
+
+        def get_init_dist(grw_dist):
+            return grw_dist.owner.inputs[0].owner.inputs[1].owner.inputs[0]
+
         grw = pm.GaussianRandomWalk.dist(mu=0, sigma=1, steps=1, init_dist=init)
-        assert tuple(grw.owner.inputs[-2].shape.eval()) == ()
+        assert tuple(get_init_dist(grw).shape.eval()) == ()
```
ricardoV94, line 138: Use tag

pymc/tests/test_distributions_timeseries.py
```diff
@@ -147,33 +130,37 @@ def test_gaussianrandomwalk_inference(self):
...
-        assert tuple(grw.owner.inputs[-2].shape.eval()) == (3, 2)
+        assert tuple(get_init_dist(grw).shape.eval()) == (3, 2)
 
     def test_shape_ellipsis(self):
         grw = pm.GaussianRandomWalk.dist(
             mu=0, sigma=1, steps=5, init_dist=pm.Normal.dist(), shape=(3, ...)
         )
         assert tuple(grw.shape.eval()) == (3, 6)
-        assert tuple(grw.owner.inputs[-2].shape.eval()) == (3,)
+        assert tuple(grw.owner.inputs[0].owner.inputs[1].owner.inputs[0].shape.eval()) == (3,)
```
ricardoV94, line 163: use tag

pymc/tests/test_distributions_timeseries.py
```diff
@@ -219,30 +206,30 @@ def test_inferred_steps_from_observed(self):
...
-                Normal.dist(0),
+                StudentT.dist(5),
                 10,
                 (4, 2),
                 np.full((4, 2, 11), np.vstack((np.arange(11), -np.arange(11)))),
             ),
         ],
     )
-    def test_moment(self, mu, sigma, init_dist, steps, size, expected):
+    def test_moment(self, mu, sigma, init, steps, size, expected):
```
ricardoV94, line 230: No reason to rename `init_dist` to `init`

ricardoV94:
@larryshamalama do you feel like picking this up after #6072? I think that makes our lives easier here

larryshamalama:
Absolutely

ricardoV94:
Actually it seems like #6072 will have to include this one by necessity

ricardoV94:
Done in #6072

PR #5816, opened 2022-05-28
Fix bug in Slice sampler and speedup `test_step.py`

### LINKED ISSUE #5815: Slice sampler does not converge in simple MvNormal prior model
This example is taken from the step_tests here: https://github.com/pymc-devs/pymc/blob/4d2f3a8d5f99fc773bc0a1aad6d51e4f3f047698/pymc/tests/test_step.py#L89

```python
import pymc as pm
import numpy as np

mu = np.array([-0.1, 0.5, 1.1])
p = np.array([[2.0, 0, 0], [0.05, 0.1, 0], [1.0, -0.05, 5.5]])
tau = np.dot(p, p.T)

with pm.Model() as model:
    pm.MvNormal( "x",  mu, tau=tau,)

expected_mean = mu
expected_std = np.diag(np.linalg.inv(tau)) ** 0.5

with model:
    trace = pm.sample(
        tune=10_000,
        draws=10_000,
        chains=1,
        step=pm.Slice(),
        return_inferencedata=False,
    )

# Very bad, the second mean seems to be consistently biased
print(expected_mean, trace["x"].mean(axis=0))
# [-0.1  0.5  1.1] [-0.14368236  1.76355742  1.11952911]

# Quite good
print(expected_std, trace["x"].std(axis=0)) 
# [ 0.56672996 10.00041321  0.18181818] [0.56335647 9.88682831 0.18468393]
```

This was also the case in V3: https://colab.research.google.com/drive/1Pw7a-1J-bnb4dPsPvxwuWH6TdFENYb5m?usp=sharing

ricardoV94:
Closes #5815 

I changed the logic when creating the initial interval according to MacKay (2005, page 375) [[pdf](https://www.inference.org.uk/itprnn/book.pdf)].

image

### Before

```python
ql[i] = q[i] - nr.uniform(0, self.w[i])
qr[i] = q[i] + self.w[i]
```

### After
```python
r = nr.uniform()
ql[i] = q[i] - r * self.w[i]
qr[i] = q[i] + (1 - r) * self.w[i]
```

I think the previous was causing the proposals to be systematically biased towards higher values (right).

Also made sure that we set the boundaries of previous variables to the accepted values, which probably by mistake was only being done during tuning.

```
# Set qr and ql to the accepted points (they matter for subsequent iterations)
qr[i] = ql[i] = q[i]
```

If someone more familiar with these can confirm my changes to `Slice` make sense that would be great. 

The sampler now converges in the multivariate example in #5815, and does an increasingly better job with more samples.

***

Also did some other minor changes, see individual commits for more details.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -148,8 +148,7 @@ This includes API changes we did not warn about since at least `3.11.0` (2021-01
 - Added the low level `compile_forward_sampling_function` method to compile the aesara function responsible for generating forward samples (see [#5759](https://github.com/pymc-devs/pymc/pull/5759)).
 - ...
 
-
-## Documentation
+### Documentation
 - Switched to the [pydata-sphinx-theme](https://pydata-sphinx-theme.readthedocs.io/en/latest/)
 - Updated our documentation tooling to use [MyST](https://myst-parser.readthedocs.io/en/latest/), [MyST-NB](https://myst-nb.readthedocs.io/en/latest/), sphinx-design, notfound.extension,
   sphinx-copybutton and sphinx-remove-toctrees.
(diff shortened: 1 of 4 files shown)


ricardoV94:
It seems like this is how the sampler used to behave until commit https://github.com/pymc-devs/pymc/commit/b988ba99a084a18b2c907ef2ee4aeb1f4ab56160 changed `right_bound = left_bound + w` -> `right_bound = initial_point + w`

michaelosthege:
> It seems like this is how the sampler used to behave until commit [b988ba9](https://github.com/pymc-devs/pymc/commit/b988ba99a084a18b2c907ef2ee4aeb1f4ab56160) changed `right_bound = left_bound + w` -> `right_bound = initial_point + w`

Wow that means the Slice sampler was broken since version 3.2.

ricardoV94:
> Wow that means the Slice sampler was broken since version 3.2.


Should we backport the fix to V3?

michaelosthege:
> Should we backport the fix to V3?

I'm not sure if it's worth it, but an exception telling users to upgrade to v4 should be easy enough.

ricardoV94:
It's the same amount of work to add an exception or change the 2 lines

michaelosthege (approved):
LGTM

ricardoV94:
I should mention something in the Release Notes

ricardoV94:
Added Release note

PR #5817, opened 2022-05-28
More safety nets for resizing mutable dimensions

### LINKED ISSUE #5812: set_data cannot deal with nodes without owners

`set_data` seems to have a problem when the new coordinates are nodes without an owner.

```python
import numpy as np
import pymc as pm


with pm.Model() as m:
    m.add_coord("a", np.arange(10), mutable=True)
    pm.MutableData("x", np.zeros(10), dims="a")

m.set_data("x", np.ones(5), coords={"a": np.arange(10, 15)})
```

```python
<ipython-input-1-dc6231ed80dd> in <module>
      7     pm.MutableData("x", np.zeros(10), dims="a")
      8 
----> 9 m.set_data("x", np.ones(5), coords={"a": np.arange(10, 15)})

~/repos/pymc3/pymc/model.py in set_data(self, name, values, coords)
   1172                     )
   1173                 else:
-> 1174                     length_belongs_to = length_tensor.owner.inputs[0].owner.inputs[0]
   1175                     if not isinstance(length_belongs_to, SharedVariable):
   1176                         raise ShapeError(

AttributeError: 'NoneType' object has no attribute 'inputs'
```

</details>

In the traceback, `length_tensor` is `m.dim_lengths["a"]`. The type of this variable is `aesara.tensor.sharedvar.ScalarSharedVariable` and it has no owners. I don't know when the owner logic was introduced, but it doesn't cover the simple case where the new coordinate is an array. Maybe this was intended? In any case, this should be fixed before releasing v4

### Original code

Originally, I was running on an older version with respect to main. The code that I used was this

```python
import numpy as np
import pymc as pm


coords = {"a": np.arange(10)}
with pm.Model(coords=coords) as m:
    pm.MutableData("x", np.zeros(10), dims="a")

m.set_data("x", np.ones(5), coords={"a": np.arange(10, 15)})
```

<details><summary>Which now leads to the following traceback instead</summary>

```python
ShapeError                                Traceback (most recent call last)
<ipython-input-2-d7c5b13f43a6> in <module>
      7     pm.MutableData("x", np.zeros(10), dims="a")
      8 
----> 9 m.set_data("x", np.ones(5), coords={"a": np.arange(10, 15)})

~/repos/pymc3/pymc/model.py in set_data(self, name, values, coords)
   1164                         )
   1165                 if isinstance(length_tensor, TensorConstant):
-> 1166                     raise ShapeError(
   1167                         f"Resizing dimension '{dname}' is impossible, because "
   1168                         "a 'TensorConstant' stores its length. To be able "

ShapeError: Resizing dimension 'a' is impossible, because a 'TensorConstant' stores its length. To be able to change the dimension length, pass `mutable=True` when registering the dimension via `model.add_coord`, or define it via a `pm.MutableData` variable.

```
</details>

I think that it's rather strange not to be able to modify coordinates given via a constant tensor, but to be honest I didn't see when/where it was discussed.

* PyMC/PyMC3 Version: 4.0.0b6
* Aesara/Theano Version:
* Python Version:
* Operating system:

michaelosthege:
+ Adds a `Model.set_dim` method for resizing dimensions that were created by `add_coord(..., mutable=True)`.
+ Changes `Model.set_data` to anticipate that data-induced resizing can target dimensions that were created through `add_coord(..., mutable=True)` which are not symbolically linked to the data variables.

Closes #5812.

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1067,12 +1067,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
             raise ValueError(
                 f"Either `values` or `length` must be specified for the '{name}' dimension."
             )
-        if isinstance(length, int):
-            length = at.constant(length)
-        elif length is not None and not isinstance(length, Variable):
-            raise ValueError(
-                f"The `length` passed for the '{name}' coord must be an Aesara Variable or None."
-            )
         if values is not None:
             # Conversion to a tuple ensures that the coordinate values are immutable.
             # Also unlike numpy arrays the's tuple.index(...) which is handy to work with.
@@ -1080,12 +1074,19 @@ class Model(WithMemoization, metaclass=ContextMeta):
         if name in self.coords:
             if not np.array_equal(values, self.coords[name]):
                 raise ValueError(f"Duplicate and incompatible coordinate: {name}.")
-        else:
+        if length is not None and not isinstance(length, (int, Variable)):
+            raise ValueError(
+                f"The `length` passed for the '{name}' coord must be an int, Aesara Variable or None."
+            )
+        if length is None:
+            length = len(values)
+        if not isinstance(length, Variable):
             if mutable:
-                self._dim_lengths[name] = length or aesara.shared(len(values))
+                length = aesara.shared(length)
             else:
-                self._dim_lengths[name] = length or aesara.tensor.constant(len(values))
-            self._coords[name] = values
+                length = aesara.tensor.constant(length)
+        self._dim_lengths[name] = length
+        self._coords[name] = values
 
     def add_coords(
         self,
@@ -1100,6 +1101,36 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         for name, values in coords.items():
             self.add_coord(name, values, length=lengths.get(name, None))
+
+    def set_dim(self, name: str, new_length: int, coord_values: Optional[Sequence] = None):
+        """Update a mutable dimension.
+
+        Parameters
+        ----------
+        name
+            Name of the dimension.
+        new_length
+            New length of the dimension.
+        coord_values
+            Optional sequence of coordinate values.
+        """
+        if not isinstance(self.dim_lengths[name], ScalarSharedVariable):
+            raise ValueError(f"The dimension '{name}' is immutable.")
+        if coord_values is None and self.coords.get(name, None) is not None:
+            raise ValueError(
+                f"'{name}' has coord values. Pass `set_dim(..., coord_values=...)` to update them."
+            )
+        if coord_values is not None:
+            len_cvals = len(coord_values)
+            if len_cvals != new_length:
+                raise ShapeError(
+                    f"Length of new coordinate values does not match the new dimension length.",
+                    actual=len_cvals,
+                    expected=new_length,
+                )
+            self._coords[name] = tuple(coord_values)
+        self.dim_lengths[name].set_value(new_length)
+        return
 
     def set_data(
         self,
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1199,28 +1199,29 @@ class Model(WithMemoization, metaclass=ContextMeta):
                     # definitely lead to shape problems.
                     raise ShapeError(
                         f"Resizing dimension '{dname}' is impossible, because "
-                        "a 'TensorConstant' stores its length. To be able "
+                        "a `TensorConstant` stores its length. To be able "
                         "to change the dimension length, pass `mutable=True` when "
                         "registering the dimension via `model.add_coord`, "
                         "or define it via a `pm.MutableData` variable."
                     )
-                elif isinstance(length_tensor, ScalarSharedVariable):
-                    # The dimension is mutable, but was defined without being linked
-                    # to a shared variable. This is allowed, but slightly dangerous.
-                    warnings.warn(
-                        f"You are resizing a variable with dimension '{dname}' which was initialized"
-                        " as a mutable dimension and is not linked to the `MutableData` variable."
-                        " Remember to update the dimension length by calling "
-                        f"`Model.set_dim({dname}, new_length={new_length})` manually,"
-                        " preferably _before_ updating `MutableData` variables that use this dimension.",
-                        ShapeWarning,
-                        stacklevel=2,
-                    )
-                else:
-                    # The dimension was created from another model variable.
-                    # If that was a non-mutable variable, there will definitely be shape problems.
+                elif length_tensor.owner is not None:
+                    # The dimension was created from a model variable.
                     length_belongs_to = length_tensor.owner.inputs[0].owner.inputs[0]
-                    if not isinstance(length_belongs_to, SharedVariable):
+                    if length_belongs_to is shared_object:
+                        # No surprise it's changing.
+                        pass
+                    elif isinstance(length_belongs_to, SharedVariable):
+                        # The dimension is mutable through a SharedVariable other than the one being modified.
+                        # But the other variable was not yet re-sized! Warn the user to do that!
+                        warnings.warn(
+                            f"You are resizing a variable with dimension '{dname}' which was initialized "
+                            f"as a mutable dimension by another variable ('{length_belongs_to}')."
+                            " Remember to update that variable with the correct shape to avoid shape issues.",
+                            ShapeWarning,
+                            stacklevel=2,
+                        )
+                    else:
+                        # The dimension is immutable.
                         raise ShapeError(
                             f"Resizing dimension '{dname}' with values of length {new_length} would lead to incompatibilities, "
                             f"because the dimension was initialized from '{length_belongs_to}' which is not a shared variable. "
@@ -1230,8 +1231,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
                             expected=old_length,
                         )
                 if isinstance(length_tensor, ScalarSharedVariable):
-                    # Updating the shared variable resizes dependent nodes that use this dimension for their `size`.
-                    length_tensor.set_value(new_length)
+                    # The dimension is mutable, but was defined without being linked
+                    # to a shared variable. This is allowed, but a little less robust.
+                    self.set_dim(dname, new_length, coord_values=new_coords)
 
             if new_coords is not None:
                 # Update the registered coord values (also if they were None)
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -60,7 +60,7 @@ from pymc.data import GenTensorVariable, Minibatch
 from pymc.distributions import joint_logpt
 from pymc.distributions.logprob import _get_scaling
 from pymc.distributions.transforms import _default_transform
-from pymc.exceptions import ImputationWarning, SamplingError, ShapeError
+from pymc.exceptions import ImputationWarning, SamplingError, ShapeError, ShapeWarning
 from pymc.initial_point import make_initial_point_fn
 from pymc.math import flatten_list
 from pymc.util import (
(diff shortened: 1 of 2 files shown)


pymc/model.py
```diff
@@ -1163,14 +1194,31 @@ def set_data(
...
                         f"Resizing dimension '{dname}' is impossible, because "
                         "a 'TensorConstant' stores its length. To be able "
                         "to change the dimension length, pass `mutable=True` when "
                         "registering the dimension via `model.add_coord`, "
                         "or define it via a `pm.MutableData` variable."
                     )
+                elif isinstance(length_tensor, ScalarSharedVariable):
+                    # The dimension is mutable, but was defined without being linked
+                    # to a shared variable. This is allowed, but slightly dangerous.
+                    warnings.warn(
```
lucianopaz: I think that the key thing that is missing here is that users can also pass `coords` when they `set_data`. Ideally, we should use the supplied `coords` internally to call `set_dim` and then update the values of the `MutableData` instance. This warning should only be raised if the users haven't passed `coords` along to `set_data`.

pymc/tests/test_model.py
```diff
@@ -765,6 +765,57 @@ def test_add_coord_mutable_kwarg():
...
+    pmodel.dim_lengths["mdim"].set_value(3)
+    # Then change the data.
+    pmodel.set_data("mdata", [1, 2, 3])
+
+    # Now the other way around.
+    # Because the dimension doesn't depend on the data variable,
+    # a warning shoudl be emitted.
+    with pytest.warns(ShapeWarning, match="update the dimension length"):
+        pmodel.set_data("mdata", [1, 2, 3, 4])
+    pass
```
lucianopaz: We should add the test where `pmodel.set_data("mdata", [1, 2, 3, 4], coords={"mdim": range(4)})`.
Ideally, we want to catch the coords fed into `set_data` and call `set_dim` with them

lucianopaz (dismissed):
Thanks, @michaelosthege. I think that we need to use `coords` values that are supplied to `set_data` to call `set_dim` internally, instead of expecting the users to `set_dim` before they `set_data`.

pymc/model.py
michaelosthege: Hmm.. You're welcome to try, but I think this could become more complicated than the current way with the warning.
The `set_data` method doesn't actually care if the `ScalarSharedVariable` corresponds to a dimension in `Model.dim_lengths` or not!

Most important will be to add test cases that showcase the intended call order

lucianopaz: What would be the expected flow to change data and make predictions? Before we had to do:

```python
with pm.Model(coords={"A": range(10)}) as m:
    x = pm.MutableData("x", x_values, dims="A")
    y = pm.MutableData("y", y_values, dims="A")
    a = pm.Normal("a", 0, 1)
    b = pm.Normal("b", 0, 1)
    c = pm.HalfNormal("c", 1)
    obs = pm.Normal("obs", mu=a + b * x, sigma=c, observed=y)
    idata = pm.sample()

with m:
    pm.set_data({"x": np.linspace(-2, 3, 100), "y": np.full(100, np.nan)})
    ppc = pm.sample_posterior_predictive(idata)
```

I would like us to be able to do this:
```python
with pm.Model(coords={"A": range(10)}) as m:
    x = pm.Data("x", x_values, dims="A")
    y = pm.Data("y", y_values, dims="A")
    a = pm.Normal("a", 0, 1)
    b = pm.Normal("b", 0, 1)
    c = pm.HalfNormal("c", 1)
    obs = pm.Normal("obs", mu=a + b * x, sigma=c, observed=y)
    idata = pm.sample()

with m:
    pm.set_data({"x": np.linspace(-2, 3, 100), "y": np.full(100, np.nan)}, coords={"A": range(10, 110)})
    ppc = pm.sample_posterior_predictive(idata)
```

But due to the choice of default Mutable/Constant coords, we need to:

1. Manually `add_coord` for "A" and say it is mutable
2. Call `set_dim` before calling `set_data`

Or we have to create the coordinate values using `MutableData`, which I don't really know how to do.

AlexAndorra:
Thanks for the proposed fix @michaelosthege ! I agree with @lucianopaz that it'd be great if the user didn't have to set_dim before they set_data

michaelosthege:
> Thanks for the proposed fix @michaelosthege ! I agree with @lucianopaz that it'd be great if the user didn't have to set_dim before they set_data

Since #5763 (7ec106c4e57861a320fc4f6a76e147c01c26f707 to be exact) one can pass `coords` to `pm.Data` containers.

So with this, you'd create the `"A"` dimension as mutable.

```python
with pm.Model() as m:
    x = pm.MutableData("x", x_values, dims="A", coords={"A": range(10)})
    y = pm.MutableData("y", y_values, dims="A")
```

Then, if you add the `coords` to `set_data`, we could also call `set_dim` internally.

michaelosthege:
I'll add the test suggested above and look into adding the `coords` kwarg to `set_data`.

michaelosthege:
@lucianopaz please review the new test cases - they should govern which usage patterns are now supported.

The only thing where I'm hestiant is introducing mutable dims through the `pm.Model` constructor.
For that we could introduce a new kwarg, but I didn't want to make that decision already.
Options:
+ `pm.Model(coords_mutable=..., coords=...)`
+ `pm.Model(coords_mutable=..., coords_immutable=...)`
+ more?

Please continue on this branch as you see fit. Let's get it over the finish line today

ricardoV94:
This is a vague hand waving comment, so apologies for that. I feel the internal code to manage dims is become a bit too convoluted for its own good. Suggestion:

1. Have a `coords` and `mutable_coords` kwargs in `pm.Model`. 
2. Remove defining news dims / coords with variables
3. Downsides?

michaelosthege:
The idea of defining `coords` through `pm.Data` came from the `export_index_as_coords` kwarg.

Thinking this further, with #5796 we could do this:

```
population = xarray.DataArray(
    "population",
    [10, 200],
    dims="city",
    coords=dict(city=["Atown", "Bvil"]),
)

with pm.Model() as m:
    pop = pm.MutableData("population", population)
    assert "city" in m.coords # True
```

Note that this would work for multi-dimensional `xarray.DataArray` objects but not for `pd.DataFrame` with multi-index.

For wroking with GPs this would be really useful, because one has to carry data and grid coordinates around quite a lot..

-----

I agree that the code is a little convoluted - it would probably benefit from a little extraction refactoring.

But the downsides of removing `coords` from `pm.Data` would be that we'd have to break the `export_index_as_coords` feature and we couldn't do what the above example does.

If we add a kwarg to the `Model` signature, I would prefer `coords_mutable` because of how things are sorted by code completion tools.

michaelosthege:
Any decision regarding the default `(im)mutable` of `pm.Model(coords)` will build on top of the commits made in this branch.

Since the commits here don't affect the thing we're still in discussion about, can we go ahead and merge this in the interest of closing #5812 and increasing overall safety?

pymc/model.py
```diff
@@ -1101,6 +1102,36 @@ def add_coords(
         for name, values in coords.items():
             self.add_coord(name, values, length=lengths.get(name, None))
 
+    def set_dim(self, name: str, new_length: int, coord_values: Optional[Sequence] = None):
```
ricardoV94, line 1105:
```suggestion
    def resize_dim(self, name: str, new_length: int, coord_values: Optional[Sequence] = None):
```

ricardoV94 (commented):
Suggestion

pymc/model.py
```diff
@@ -1101,6 +1102,36 @@ def add_coords(
...
+        coord_values
+            Optional sequence of coordinate values.
+        """
+        if not isinstance(self.dim_lengths[name], ScalarSharedVariable):
+            raise ValueError(f"The dimension '{name}' is immutable.")
+        if self.coords.get(name, None) is not None and coord_values is None:
+            raise ValueError(
+                f"'{name}' has coord values. Pass `set_dim(..., coord_values=...)` to update them."
+            )
+        if coord_values is not None:
```
```suggestion
        if coord_values is None:
            if name in self.coords:
                raise ValueError(
                    f"'{name}' has coord values. Pass `set_dim(..., coord_values=...)` to update them."
            )
        else:
```

pymc/model.py
```diff
@@ -1163,16 +1194,34 @@ def set_data(
...
-                        "a 'TensorConstant' stores its length. To be able "
+                        "a `TensorConstant` stores its length. To be able "
                         "to change the dimension length, pass `mutable=True` when "
                         "registering the dimension via `model.add_coord`, "
                         "or define it via a `pm.MutableData` variable."
                     )
-                else:
+                elif length_tensor.owner is not None:
+                    # The dimension was created from a model variable.
                     length_belongs_to = length_tensor.owner.inputs[0].owner.inputs[0]
```
ricardoV94, line 1215: Can you give me some context on what this `length_belongs_to` does? I have been staring at it for a while now without success. Is it checking if it comes from a call to `variable.shape` or something? What is the graph above `length_tensor` that we are assuming?

pymc/model.py
```diff
@@ -1163,16 +1194,34 @@ def set_data(
...
+                        pass
+                    elif isinstance(length_belongs_to, SharedVariable):
+                        # The dimension is mutable through a SharedVariable other than the one being modified.
+                        # But the other variable was not yet re-sized! Warn the user to do that!
+                        warnings.warn(
+                            f"You are resizing a variable with dimension '{dname}' which was initialized "
+                            f"as a mutable dimension by another variable ('{length_belongs_to}')."
+                            " Remember to update that variable with the correct shape to avoid shape issues.",
+                            ShapeWarning,
+                            stacklevel=2,
```
ricardoV94, line 1227: Why this stacklevel? Isn't this the function we want to see the trace from?

ricardoV94 (commented):
Some suggestions and questions :)

pymc/model.py
michaelosthege: I named it `set_dim` to align with the related method `set_data` and because it's not always resizing the dimension, but can also be used to just change coords.

pymc/model.py
michaelosthege:
```suggestion
                    # Get a handle on the tensor from which this dimension length was
                    # obtained by doing subindexing on the shape as in `.shape[i]`.
                    # Needed to check if it was another shared variable.
                    length_belongs_to = length_tensor.owner.inputs[0].owner.inputs[0]
```

pymc/model.py
michaelosthege: This makes the trace point directly point to the line of user code - unless I'm misunderstanding how this works.

michaelosthege (commented):
@ricardoV94 @twiecki I won't have time to do any commits here until Monday.
As in "*I'm out, please take over the rest here and get this merged*".

pymc/model.py
ricardoV94: My understanding is that a custom stacklevel is only warranted when you are issuing the warning from a helper function and you don't want the trace to point to that helper function but to the line that called the helper function. Here we want to see the exact line where it is called no?

pymc/model.py
ricardoV94: Should we then check something like `isinstance(length_tensor.owner.op, Subtensor)` and `isinstance(length_tensor.owner.inputs[0].owner.op, Shape)` just to be on the safe side?

pymc/model.py
ricardoV94: Then perhaps we can make the docstrings a bit more in line with it?

twiecki:
What's missing here, can we merge?

pymc/model.py
ricardoV94: Nevermind, warnings don't show the fulltrace by default

ricardoV94 (approved):
Waiting for tests to pass to make sure I didn't break anything accidentally

PR #5823, opened 2022-05-30
Make Metropolis cope better with multiple dimensions

ricardoV94:
Metropolis now updates each dimension sequentially and tunes a proposal scale parameter per dimension

The performance of the Metropolis sampler with multiple dimensions was incredibly poor as highlighted in this old closed issue https://github.com/pymc-devs/pymc/issues/4223. After these changes, the performance should now be similar between the scalar and batched cases.

Some care had to be taken towards multivariate discrete distributions where we cannot safely update one dimension at a time (e.g., multinomial), because any intermediate change would result in -inf logp. We should probably implement a smarter sampler just for this distribution, but for the time being we avoid doing elemwise updates when multivariate discrete variables are assigned.

### DIFF
--- a/pymc/step_methods/metropolis.py
+++ b/pymc/step_methods/metropolis.py
@@ -168,10 +168,12 @@ class Metropolis(ArrayStepShared):
             vars = model.value_vars
         else:
             vars = [model.rvs_to_values.get(var, var) for var in vars]
+
         vars = pm.inputvars(vars)
 
+        initial_values_shape = [initial_values[v.name].shape for v in vars]
         if S is None:
-            S = np.ones(sum(initial_values[v.name].size for v in vars))
+            S = np.ones(int(sum(np.prod(ivs) for ivs in initial_values_shape)))
 
         if proposal_dist is not None:
             self.proposal_dist = proposal_dist(S)
@@ -186,7 +188,6 @@ class Metropolis(ArrayStepShared):
         self.tune = tune
         self.tune_interval = tune_interval
         self.steps_until_tune = tune_interval
-        self.accepted = 0
 
         # Determine type of variables
         self.discrete = np.concatenate(
@@ -195,11 +196,33 @@ class Metropolis(ArrayStepShared):
         self.any_discrete = self.discrete.any()
         self.all_discrete = self.discrete.all()
 
-        # remember initial settings before tuning so they can be reset
-        self._untuned_settings = dict(
-            scaling=self.scaling, steps_until_tune=tune_interval, accepted=self.accepted
+        # Metropolis will try to handle one batched dimension at a time This, however,
+        # is not safe for discrete multivariate distributions (looking at you Multinomial),
+        # due to high dependency among the support dimensions. For continuous multivariate
+        # distributions we assume they are being transformed in a way that makes each
+        # dimension semi-independent.
+        is_scalar = len(initial_values_shape) == 1 and initial_values_shape[0] == ()
+        self.elemwise_update = not (
+            is_scalar
+            or (
+                self.any_discrete
+                and max(getattr(model.values_to_rvs[var].owner.op, "ndim_supp", 1) for var in vars)
+                > 0
+            )
         )
+        if self.elemwise_update:
+            dims = int(sum(np.prod(ivs) for ivs in initial_values_shape))
+        else:
+            dims = 1
+        self.enum_dims = np.arange(dims, dtype=int)
+        self.accept_rate_iter = np.zeros(dims, dtype=float)
+        self.accepted_iter = np.zeros(dims, dtype=bool)
+        self.accepted_sum = np.zeros(dims, dtype=int)
 
+        # remember initial settings before tuning so they can be reset
+        self._untuned_settings = dict(scaling=self.scaling, steps_until_tune=tune_interval)
+
+        # TODO: This is not being used when compiling the logp function!
         self.mode = mode
 
         shared = pm.make_shared_replacements(initial_values, vars, model)
@@ -210,6 +233,7 @@ class Metropolis(ArrayStepShared):
         """Resets the tuned sampler parameters to their initial values."""
         for attr, initial_value in self._untuned_settings.items():
             setattr(self, attr, initial_value)
+        self.accepted_sum[:] = 0
         return
 
     def astep(self, q0: RaveledVars) -> Tuple[RaveledVars, List[Dict[str, Any]]]:
@@ -219,10 +243,10 @@ class Metropolis(ArrayStepShared):
 
         if not self.steps_until_tune and self.tune:
             # Tune scaling parameter
-            self.scaling = tune(self.scaling, self.accepted / float(self.tune_interval))
+            self.scaling = tune(self.scaling, self.accepted_sum / float(self.tune_interval))
             # Reset counter
             self.steps_until_tune = self.tune_interval
-            self.accepted = 0
+            self.accepted_sum[:] = 0
 
         delta = self.proposal_dist() * self.scaling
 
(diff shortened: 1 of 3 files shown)


pymc/step_methods/metropolis.py
```diff
@@ -275,26 +312,38 @@ def tune(scale, acc_rate):
     >0.95         x 10
 
     """
-    if acc_rate < 0.001:
+    return scale * np.where(
```
aloctavodia, line 315: While being here, what about using a continuous update like in SMC? https://github.com/pymc-devs/pymc/blob/de555439b36f8c67e6e541ec33fac1e4aadcde88/pymc/smc/smc.py#L474

aloctavodia (approved):

pymc/step_methods/metropolis.py
ricardoV94: Maybe I'll propose that in a separate PR?

ricardoV94: Or maybe here. Is there any downside that you see to switching?

aloctavodia: I don't think so.

pymc/tests/test_step.py
```diff
@@ -1314,8 +1342,9 @@ def perform(self, node, inputs, outputs):
                     (nchains, ndraws * nsub)
                 )
                 Q_2_1 = np.concatenate(trace.get_sampler_stats("Q_2_1")).reshape((nchains, ndraws))
-                assert Q_1_0.mean(axis=1) == 0.0
-                assert Q_2_1.mean(axis=1) == 0.0
+                # This used to be a scrict zero equality!
+                assert np.isclose(Q_1_0.mean(axis=1), 0.0, atol=1e-4)
+                assert np.isclose(Q_2_1.mean(axis=1), 0.0, atol=1e-4)
```
ricardoV94, line 1347: I don't know who is maintaining the MLDA step method, but I had to change this strict equality... Is that okay?

pymc/step_methods/metropolis.py
```diff
@@ -262,39 +299,10 @@ def competence(var, has_grad):
...
-        return scale * 10.0
-    elif acc_rate > 0.75:
-        # increase by double
-        return scale * 2.0
-    elif acc_rate > 0.5:
-        # increase by ten percent
-        return scale * 1.1
-
-    return scale
+    return np.exp(np.log(scale) + (acc_rate - 0.234))
```
ricardoV94: @aloctavodia Is this correct?

ricardoV94:
I am going to temporarily revert the last commit to see if the tests pass. The Metropolis related tests seem to be performing a bit poorer with the proposed alternative tune rule (or it could very well be just be a seeding thing)

pymc/tests/test_step.py
```diff
@@ -403,6 +403,40 @@ def test_tuning_reset(self):
...
+                shape=(
+                    2,
+                    20,
+                ),
+            ),
+            Dirichlet.dist(a=np.ones(3) * (np.arange(40) + 1)[:, None], shape=(40, 3)),
+            Dirichlet.dist(a=np.ones(3) * (np.arange(20) + 1)[:, None], shape=(2, 20, 3)),
+        ),
+    )
+    def test_elemwise_update(self, batched_dist):
```
ricardoV94, line 423: I run these new tests 10 times each, with the current new algorithm and the one proposed by @aloctavodia that we use in SMC. Things look sensible in both cases:

## 10 runs each with different seeds

### Tuning method based on switch statement:
```
1. (0D Bin) max_rhat(mean=1.005, std=0.005), min_ess=(mean=327.4, std=29.9)
2. (1D Bin) max_rhat(mean=1.031, std=0.007), min_ess=(mean=115.5, std=20.9)
3. (2D Bin) max_rhat(mean=1.021, std=0.005), min_ess=(mean=180.0, std=22.5)
4. (2D Dir) max_rhat(mean=1.041, std=0.010), min_ess=(mean=91.2, std=24.7)
5. (3D Dir) max_rhat(mean=1.037, std=0.008), min_ess=(mean=110.4, std=24.4)
```

### Tuning method based on distance from 0.234:
```
1. (0D Bin) max_rhat(mean=1.009, std=0.008), min_ess=(mean=231.3, std=45.6)  # Worse
2. (1D Bin) max_rhat(mean=1.026, std=0.008), min_ess=(mean=206.9, std=21.5)  # Better
3. (2D Bin) max_rhat(mean=1.026, std=0.005), min_ess=(mean=188.7, std=22.3)
4. (2D Dir) max_rhat(mean=1.043, std=0.012), min_ess=(mean=98.9, std=27.8)
5. (3D Dir) max_rhat(mean=1.039, std=0.009), min_ess=(mean=102.6, std=21.5)
```

pymc/step_methods/metropolis.py
aloctavodia: Yes, also the initial guess for the scale should be something like this https://github.com/pymc-devs/pymc/blob/5703a9d24ca01da8b81363874e023e59cd18688d/pymc/smc/smc.py#L461-L462 (for problem with a few dimensions is the same as starting from 1)

We can make the scaling changes in a different PR.

pymc/tests/test_step.py
ricardoV94: But the MLDA complains a lot about the SMC-like tuning... specially this test fails even if I increase the number of draws: https://github.com/ricardoV94/pymc/blob/3a84db6b91035a7c7f68633bdbb18c5f11efd46f/pymc/tests/test_step.py#L1184

Maybe I am just misunderstanding what that test is supposed to check

pymc/step_methods/metropolis.py
ricardoV94: Yeah, let's do it in a separate PR, also because of the MLDA complaining here

ricardoV94:
@aloctavodia I removed the new tune strategy. I'll open a new PR for that. Do you care to review again?

pymc/step_methods/metropolis.py
```diff
@@ -195,11 +196,33 @@ def __init__(
...
         self.all_discrete = self.discrete.all()
 
-        # remember initial settings before tuning so they can be reset
-        self._untuned_settings = dict(
-            scaling=self.scaling, steps_until_tune=tune_interval, accepted=self.accepted
+        # Metropolis will try to handle one batched dimension at a time This, however,
+        # is not safe for discrete multivariate distributions (looking at you Multinomial),
+        # due to high dependency among the support dimensions. For continuous multivariate
+        # distributions we assume they are being transformed in a way that makes each
+        # dimension semi-independent.
```
michaelosthege, line 203:
```suggestion
        # dimension semi-independent.
        # Consequently we'd like to tune the scaling based on
        # acceptance rate in each dimension independently.
```

pymc/step_methods/metropolis.py
```diff
@@ -195,11 +196,33 @@ def __init__(
         self.any_discrete = self.discrete.any()
         self.all_discrete = self.discrete.all()
 
-        # remember initial settings before tuning so they can be reset
-        self._untuned_settings = dict(
-            scaling=self.scaling, steps_until_tune=tune_interval, accepted=self.accepted
+        # Metropolis will try to handle one batched dimension at a time This, however,
```
michaelosthege, line 199:
```suggestion
        # Metropolis will try to handle one batched dimension at a time. This, however,
```

pymc/step_methods/metropolis.py
```diff
@@ -195,11 +196,33 @@ def __init__(
...
+        self.elemwise_update = not (
+            is_scalar
+            or (
+                self.any_discrete
+                and max(getattr(model.values_to_rvs[var].owner.op, "ndim_supp", 1) for var in vars)
+                > 0
+            )
         )
+        if self.elemwise_update:
+            dims = int(sum(np.prod(ivs) for ivs in initial_values_shape))
```
michaelosthege, line 214: 1. Should this be an internal attribute?
2. Could you come up with a more informative name?

Right now (reading the diff top to bottom) I'm confused because this smells like `CompoundStep`, but it's only about elementwise tuning, right?

pymc/step_methods/metropolis.py
```diff
@@ -237,23 +261,36 @@ def astep(self, q0: RaveledVars) -> Tuple[RaveledVars, List[Dict[str, Any]]]:
...
+            np.random.shuffle(self.enum_dims)
+            for i in self.enum_dims:
+                q_temp[i] = q[i]
+                accept_rate_i = self.delta_logp(q_temp, q0)
+                q_temp_, accepted_i = metrop_select(accept_rate_i, q_temp, q0)
+                q_temp[i] = q_temp_[i]
+                self.accept_rate_iter[i] = accept_rate_i
+                self.accepted_iter[i] = accepted_i
+                self.accepted_sum[i] += accepted_i
+            q = q_temp
```
michaelosthege, line 276: Is this doing what we usually do with `CompoundStep`?

If not, maybe explain what the if/else blocks do in a code comment

pymc/step_methods/metropolis.py
```diff
@@ -275,26 +312,38 @@ def tune(scale, acc_rate):
...
+                            # increase by ten percent
+                            1.1,
+                            # Do not change
+                            1.0,
+                        ),
+                    ),
+                ),
+            ),
+        ),
+    )
```
michaelosthege, line 346: Ugh, can't we do this in a `for` iteration with a `break`?

pymc/tests/test_step.py
michaelosthege: Looking at the standard deviations I'm not that impressed thb.
(Considering that the code is now more complicated..)

Or do you have another application that motivated this?

pymc/tests/test_step.py
michaelosthege: Nobody is actively maintaining it. Also, didn't we want to move it to `pymc-experimental`?

pymc/step_methods/metropolis.py
ricardoV94: CompoundStep assigns one variable per step, here we are sampling one dimension within a variable (or within multiple variables) semi-independently

pymc/tests/test_step.py
ricardoV94: @michaelosthege this is not the change from now vs before the PR, it's the change between this PR and an alternative procedure that @aloctavodia suggested during this PR.

Before this PR there were zero accepted transitions in all the new tests (except the first scalar case)

I updated the headings to make it obvious

pymc/step_methods/metropolis.py
ricardoV94: See my comment below. `CompoundStep` has nothing to do with this.

pymc/step_methods/metropolis.py
ricardoV94: We can probably index, that should be faster for large arrays because all branches of `np.where` are evaluated by default

pymc/tests/test_step.py
ricardoV94: I think it is still maintained as it was ported by the original team to V4

aloctavodia (approved):
LGTM!

PR #5824, opened 2022-05-30
Small improvements to early nuts behaviour

aseyboldt:
Step size adaptation:
Currently, we compute the acceptance rate of a draw only in that part of the trajectory that is accepted at the end. This way we discard some additional information about energy errors in the remaining discarded part. After this PR we compute the mean acceptance rate over all leapfrog steps we take.

Divergences:
Currently we count large derivations from the initial energy as a divergence regardless of the direction of the energy error.
This PR changes this so that only large energy errors that correspond to a low acceptance probability are considered a divergence. I think this helps with some model in the initial phase of sampling, were with the old behavior we stop promising trajectories that are *too good*, leading to worse performance.

Also a little bit of housekeeping:
This introduces an additional sampler statistic: `index_in_trajectory`. I don't see much of a use-case for this other than teaching and debugging, but it proved useful while comparing the sampler to other implementations.

I also introduced some array copies for the integrated momentum array. I thought I saw bugs while comparing sampler behavior to nuts-rs, but somehow I can't reproduce those now. Still, I think it's better to be safe.

### DIFF
--- a/pymc/step_methods/hmc/quadpotential.py
+++ b/pymc/step_methods/hmc/quadpotential.py
@@ -254,6 +254,7 @@ class QuadPotentialDiagAdapt(QuadPotential):
 
     def _update_from_weightvar(self, weightvar):
         weightvar.current_variance(out=self._var)
+        self._var = np.clip(self._var, 1e-12, 1e12)
         np.sqrt(self._var, out=self._stds)
         np.divide(1, self._stds, out=self._inv_stds)
         self._var_aesara.set_value(self._var)
--- a/pymc/step_methods/hmc/quadpotential.py
+++ b/pymc/step_methods/hmc/quadpotential.py
@@ -136,6 +136,9 @@ class QuadPotential:
     def reset(self):
         pass
 
+    def stats(self):
+        return {"largest_eigval": np.nan, "smallest_eigval": np.nan}
+
 
 def isquadpotential(value):
     """Check whether an object might be a QuadPotential object."""
--- a/pymc/step_methods/hmc/base_hmc.py
+++ b/pymc/step_methods/hmc/base_hmc.py
@@ -236,6 +236,7 @@ class BaseHMC(GradientSharedStep):
 
         stats.update(hmc_step.stats)
         stats.update(self.step_adapt.stats())
+        stats.update(self.potential.stats())
 
         return hmc_step.end.q, [stats]
 
--- a/pymc/step_methods/hmc/hmc.py
+++ b/pymc/step_methods/hmc/hmc.py
@@ -51,6 +51,8 @@ class HamiltonianMC(BaseHMC):
             "process_time_diff": np.float64,
             "perf_counter_diff": np.float64,
             "perf_counter_start": np.float64,
+            "largest_eigval": np.float64,
+            "smallest_eigval": np.float64,
         }
     ]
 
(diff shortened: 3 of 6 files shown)


pymc/step_methods/hmc/quadpotential.py
```diff
@@ -136,6 +136,9 @@ def raise_ok(self, map_info=None):
     def reset(self):
         pass
 
+    def stats(self):
+        return {"largest_eigval": np.nan, "smallest_eigval": np.nan}
```
ColCarroll, line 140: is this for the future?

ColCarroll (commented):
looks good! 

two questions:
1. is the change from `log_weighted_accept_sum` to `log_accept_sum` what is making the change in step size adaptation? I am having trouble seeing that.
2. these changes are careful enough that I might ask for, say, 10k samples from a random 100d gaussian to show that it still provides unbiased samples.

aseyboldt:
@ColCarroll 
> is the change from log_weighted_accept_sum to log_accept_sum what is making the change in step size adaptation? I am having trouble seeing that.

Yes. Maybe it is easier to see if you only look at the first commit.
Previously, we had information about the acceptance rate in the `Subtree` structure, now there is only the attribute `log_accept_sum` in the NutsTree class, that keeps track of the sum of acceptance rates, and a counter `n_proposals` that counts how many leapfrog steps we did.

> these changes are careful enough that I might ask for, say, 10k samples from a random 100d gaussian to show that it still provides unbiased samples.

That should also be in the tests already, (https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_posteriors.py), but you are right to ask, I'll also run that manually to make sure. (also we have to run the long-running test manually)

pymc/step_methods/hmc/quadpotential.py
aseyboldt: That got here my accident...
I can take it out, it is useful though for the new code in covadapt (which I think is pretty nice actually, and I'll try to get into pymc at some point).

ColCarroll: If you want to leave it in, I'm ok with it (so long as it has a comment!) I wonder what you think of algorithm 2 in https://proceedings.mlr.press/v151/hoffman22a/hoffman22a.pdf as a means of estimating scales?

aseyboldt: Thomas sent me a link to that paper just this morning. I also wonder what I'll think about algorithm 2, looks fascinating though. ;-)
If you like to compare ideas with what I did in covadapt, I just wrote a sketch of an intro in the readme: https://github.com/aseyboldt/covadapt
If you don't understand what I'm talking about over there, that's my fault. I'll try to improve it soon. :-)

aseyboldt:
Notebook with the promised verification of the posterior: https://gist.github.com/aseyboldt/43354eded52981340304d3aec94ced3c
I also added a few lines in the docstring about the new stats.

ricardoV94:
> Notebook with the promised verification of the posterior: https://gist.github.com/aseyboldt/43354eded52981340304d3aec94ced3c I also added a few lines in the docstring about the new stats.

Can we include that as slow test (which will not run by default)

ricardoV94:
@ColCarroll can we get a binary review from you :D?

ColCarroll (approved):
apologies for the wait, and thanks for the verification, @aseyboldt!

ricardoV94:
Thanks @aseyboldt and @ColCarroll!

ColCarroll:
I guess this is an enhancement of the NUTS algorithm, so it will change results for NUTS users (i.e., most users), we assume for the better. My impression (defer to adrian!) is that these should be fairly subtle or only matter for the most careful or difficult models, so it does *not* require (say) a minor version release (which I'd suggest doing if the old algorithm was arguably wrong), and can go out whenever the next release was planned anyways.

PR #5827, opened 2022-05-31
Supress warning SMC initialization

aloctavodia:
This is a little bit weird. Catching the warning makes the progressbar not progress. But it seems this is not related to the warning itself, manually removing the warning from `sample_prior_predictive` has the same effect. Additionally it works when core=1, i.e. not parallel sampling and it also works if we add a print statement.  So I guess is related to  something going on here https://github.com/pymc-devs/pymc/blob/5703a9d24ca01da8b81363874e023e59cd18688d/pymc/smc/sample_smc.py#L398-L417 @ricardoV94 any guess?

### DIFF
--- a/pymc/smc/smc.py
+++ b/pymc/smc/smc.py
@@ -14,6 +14,7 @@
 import abc
 
 from abc import ABC
+import sys
 from typing import Dict, cast
 import warnings
 
@@ -176,7 +177,7 @@ class SMC_KERNEL(ABC):
 
     def initialize_population(self) -> Dict[str, np.ndarray]:
         """Create an initial population from the prior distribution"""
-        print()
+        sys.stdout.write(" ")
         with warnings.catch_warnings():
             warnings.filterwarnings(
                 "ignore", category=UserWarning, message="The effect of Potentials"
--- a/pymc/smc/smc.py
+++ b/pymc/smc/smc.py
@@ -177,7 +177,7 @@ class SMC_KERNEL(ABC):
 
     def initialize_population(self) -> Dict[str, np.ndarray]:
         """Create an initial population from the prior distribution"""
-        sys.stdout.write(" ")
+        sys.stdout.write(" ")  # see issue #5828
         with warnings.catch_warnings():
             warnings.filterwarnings(
                 "ignore", category=UserWarning, message="The effect of Potentials"
--- a/pymc/smc/smc.py
+++ b/pymc/smc/smc.py
@@ -15,6 +15,7 @@ import abc
 
 from abc import ABC
 from typing import Dict, cast
+import warnings
 
 import aesara.tensor as at
 import numpy as np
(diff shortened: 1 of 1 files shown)


ricardoV94:
Might you be catching too many things with that context manager? Including whatever progressbar uses for piping the messages?

aloctavodia:
It is not the catching, removing the warning has the same effect. And works as intended when no parallel sampling.

ricardoV94:
> It is not the catching, removing the warning has the same effect. And works as intended when no parallel sampling.

What about a model that would not raise the warning?

aloctavodia:
Not working. Also I just checked that is not working inside VS code or the browser. And checked a couple of older versions of fastprogressbar as well.

ricardoV94:
I remember the code was pretty hacky / fragile and I could only see output in some contexts (e.g., terminal vs pycharm). So I am not super shocked

aloctavodia:
What about just leaving the "print()" for the moment and open an issue so we can back to this after the release? It is possible to locally disable the pre-commit so it does not complain about the print?

ricardoV94:
You can do `sys.stdout.write(msg)` I think, which will do the same as `print`

ricardoV94:
I just confirmed that I see the message in a vanilla bash terminal, but not in Jupyter Notebook. This definitely used to work in notebooks:

```python
import pymc as pm

with pm.Model() as m:
    x = pm.Normal("x")
    y = pm.Normal("y", x, observed=5)
    pm.sample_smc()
```

pymc/smc/smc.py
```diff
@@ -175,12 +177,18 @@ def __init__(
 
     def initialize_population(self) -> Dict[str, np.ndarray]:
         """Create an initial population from the prior distribution"""
-        result = sample_prior_predictive(
-            self.draws,
-            var_names=[v.name for v in self.model.unobserved_value_vars],
-            model=self.model,
-            return_inferencedata=False,
-        )
+        sys.stdout.write(" ")
```
ricardoV94: This makes it show up?

aloctavodia: yes. Actually it can be anywhere inside `smc.py`

ricardoV94: Haha, I have to test it locally!!! If it works, we should add a comment explaining why that is there

ricardoV94: It now works on Jupyter. Still does not work in PyCharm, but I think that is their problem

aloctavodia: Feel free to add more details to this issue https://github.com/pymc-devs/pymc/issues/5828

ricardoV94 (approved):
What a hack xD!!! LGTM otherwise

jucor:
Heads up that today this hack bit us, in an admittedly niche setup: 
- I'm calling `pymc` 
- within a python chunk 
- in a RStudio Rmarkdown document thanks to  the package [reticulate](https://rstudio.github.io/reticulate/articles/r_markdown.html). 
- within a conda environment (pfew)
- on Windows 10 😆  (The problem does not appear on my Mac machine, only on my colleague @Caroline Win10 machine)

So it is interactive *and* called from something else than a terminal or a jupyter notebook, and on a Windows machine to boot. 

And in that case, apparently `stdout` has been somewhere set to `None`, as I get the error `nonetype object has no attribute write` 😢 

I'm heartbroken 💔  and super tickled of trying to find out what's going on. I'll do more debugging next week (need to dig up another Windows machine for my own tests),  will report here.  Maybe I'll simply add a check that `stdout` isn't `None` before deploying the hack? (a hack on a hack 😆 ) 

In the meantime if you have any idea, they're very welcome!

jucor:
Yup, I can confirm that  the below patch fixes the issue with reticulate (which indeed seems to mangle somewhat stdout). At the moment I'm monkey-patching the package on the fly (🐷 ). Want me to open a proper PR?

```patch
--- a/kernels.py	2024-06-12 22:15:14.506303300 +0200
+++ b/kernels.py	2024-06-12 22:15:32.926642800 +0200
@@ -186,7 +186,8 @@
 
     def initialize_population(self) -> dict[str, np.ndarray]:
         """Create an initial population from the prior distribution"""
-        sys.stdout.write(" ")  # see issue #5828
+        if sys.stdout is not None:
+          sys.stdout.write(" ")  # see issue #5828
             warnings.filterwarnings(
                 "ignore", category=UserWarning, message="The effect of Potentials"
```

PR #5829, opened 2022-05-31
Change the default init_dist in timeseries and issue UserWarning

### LINKED ISSUE #5744: Improve default GaussianRandomWalk and AR init distributions

ricardoV94:
Sorry @AlexAndorra, I went ahead and opened a PR for this. Tests likely to fail for now :)

Closes #5744

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -227,7 +227,6 @@ class GaussianRandomWalk(distribution.Continuous):
         sigma > 0, innovation standard deviation, defaults to 1.0
     init_dist : unnamed distribution
         Univariate distribution of the initial value, created with the `.dist()` API.
-        Defaults to a unit Normal.
 
         .. warning:: init will be cloned, rendering them independent of the ones passed as input.
 
@@ -274,7 +273,12 @@ class GaussianRandomWalk(distribution.Continuous):
 
         # If no scalar distribution is passed then initialize with a Normal of same mu and sigma
         if init_dist is None:
-            init_dist = Normal.dist(0, 1)
+            warnings.warn(
+                "Initial distribution not specified, defaulting to `Normal.dist(0, 100)`."
+                "You can specify an init_dist manually to suppress this warning.",
+                UserWarning,
+            )
+            init_dist = Normal.dist(0, 100)
         else:
             if not (
                 isinstance(init_dist, at.TensorVariable)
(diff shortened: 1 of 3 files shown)


pymc/distributions/timeseries.py
```diff
@@ -170,16 +172,16 @@ def rng_fn(
...
         rng: np.random.RandomState
            Numpy random number generator
-        mu: array_like
+        mu: array_like of float
            Random walk mean
-        sigma: np.ndarray
+        sigma: array_like of float
             Standard deviation of innovation (sigma > 0)
-        init: float
+        init_dist: array_like of float
```
twiecki, line 179: float? Shouldn't this be a distribution?

pymc/distributions/timeseries.py
```diff
@@ -452,7 +465,13 @@ def dist(
...
                 )
         else:
-            init_dist = Normal.dist(0, 1, size=(*sigma.shape, ar_order))
+            warnings.warn(
+                "Initial distribution not specified, defaulting to "
+                "`Normal.dist(0, 100, shape=...)`. You can specify an init_dist "
+                "manually to suppress this warning.",
+                UserWarning,
+            )
+            init_dist = Normal.dist(0, 100, size=(*sigma.shape, ar_order))
```
twiecki: Should use shape.

pymc/distributions/timeseries.py
ricardoV94: In the `rng_fn` we get non-symbolic inputs already

ricardoV94: In the `rng_fn`, we get the non-symbolic values already

twiecki: Got it.

pymc/distributions/timeseries.py
ricardoV94: +1 for picking one or the other.

Changing it

twiecki (approved):

twiecki:
One down!

PR #5865, opened 2022-06-07
Fixing keyword args in top-level pm.compile_fn

dfm:
`Model.compile_fn` requires all parameters except `outs` to be keyword arguments so the top level `compile_fn` didn't work as written.

Before you proceed, please make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).

### DIFF
--- a/pymc/tests/test_model.py
+++ b/pymc/tests/test_model.py
@@ -975,8 +975,13 @@ def test_compile_fn():
 
     test_vals = np.array([0.0, -1.0])
     state = {"x": test_vals, "y_log__": test_vals}
-    result_expect = m.compile_fn([x, y])(state)
-    with m:
-        result_compute = pm.compile_fn([x, y])(state)
-    for a, b in zip(result_compute, result_expect):
-        np.testing.assert_allclose(a, b)
+
+    for target in [x, y]:
+        with m:
+            func = pm.compile_fn(target)
+            result_compute = func(state)
+
+        func = m.compile_fn(target)
+        result_expect = func(state)
+
+        np.testing.assert_allclose(result_compute, result_expect)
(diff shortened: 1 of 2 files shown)


twiecki:
Maybe add a test?

dfm:
This wasn't previously tested (hence why the bug existed!) so I'd say that that's beyond the scope of my PR, but if someone else wants to, that would be cool.

ricardoV94:
> This wasn't previously tested (hence why the bug existed!) so I'd say that that's beyond the scope of my PR, but if someone else wants to, that would be cool.

Yeah we should test, so we weren't calling this function anywhere? It doesn't need to be you if you don't want to ofc. Thanks for opening the PR

ricardoV94 (dismissed):
Thanks @dfm!

dfm:
Wait until we see if the test executes - I just wrote it on the GitHub UI :D

pymc/tests/test_model.py
```diff
@@ -966,3 +966,22 @@ def test_deterministic():
...
+
+def test_compile_fn():
+    with pm.Model() as m:
+        x = pm.Normal("x", 0, 1, size=2)
+        y = pm.LogNormal("y", 0, 1, size=2)
+
+    test_vals = np.array([0.0, -1.0])
+    state = {"x": test_vals, "y_log__": test_vals}
+
+    for target in [x, y]:
```
ricardoV94, line 979: I don't think that should be the case :P (that it only accepts one output). Did we introduce that constraint somewhere by accident?

dfm: I'm not sure - it's still not working with a strange error message!

dfm:
I'm not sure why my proposed test isn't working, but I'll close this PR and just open an issue and someone else can deal with it because I don't have time!

My proposed 5 character change fixes things in my case but YMMV :D

PR #5869, opened 2022-06-08
Moved wiki install guides into installation.md directly

### LINKED ISSUE #5868: Remove all install guides from wiki and move to source control
People can edit wiki silently which is an issue for install guides
https://twitter.com/pymc_devs/status/1534472124072591362

Moving to source control means they can't be edited without visibility and approval

fonnesbeck:
Closes #5868

### DIFF
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -1,11 +1,9 @@
 (installation)=
 # Installation
 
-To install PyMC, select the operating system where you want to perform the installation.
+We recommend using [Anaconda](https://www.anaconda.com/) (or [Miniforge](https://github.com/conda-forge/miniforge)) to install Python on your local machine, which allows for packages to be installed using its `conda` utility.
 
-## Linux
-
-We recommend using [Anaconda](https://www.anaconda.com/) (or [Miniforge](https://github.com/conda-forge/miniforge)) to install Python on MacOS, which allows for packages to be installed using its `conda` utility.
+Once you have installed one of the above, PyMC can be installed into a new conda environment as follows:
 
 ```console
 conda create -c conda-forge -n pymc_env pymc
@@ -13,7 +11,8 @@ conda activate pymc_env
 ```
 If you like, replace the name `pymc_env` with whatever environment name you prefer.
 
-### JAX sampling
+
+## JAX sampling
 
 If you wish to enable sampling using the JAX backend via NumPyro (experimental), the following should also be installed:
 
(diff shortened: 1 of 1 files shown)


twiecki:
We can just have one set of instructions across all OS.

OriolAbril (changes requested):
As part of the docs this will now be versioned and therefore should have no pymc3 content

fonnesbeck:
Except the PyMC3 install instructions aren't yet in a version, are they?

OriolAbril:
> Except the PyMC3 install instructions aren't yet in a version, are they?

3.11 docs points to the wiki for install instructions: https://www.pymc.io/projects/docs/en/v3.11.5/ (there is one clickable card per OS in the installation section). Killing the wiki is the issue with 3.x install instructions, I don't think adding them on 4.x docs will fix anything on that end.

docs/source/installation.md
```diff
@@ -1,13 +1,30 @@
...
+```console
+conda create -c conda-forge -n pymc_env pymc
+conda activate pymc_env
+```
+If you like, replace the name `pymc_env` with whatever environment name you prefer.
+
+
+## JAX sampling
+
+If you wish to enable sampling using the JAX backend via NumPyro (experimental), the following should also be installed:
```
OriolAbril: is only numpyro experimental but not blackjax? If so I'd change the order and put blackjax first

docs/source/installation.md
```diff
@@ -1,13 +1,30 @@
...
+
+```console
+pip install jax jaxlib numpyro
```
OriolAbril: I find it a bit confusing that jax and jaxlib are included here but not for blackjax. Both numpyro and blackjax depend on jax and jaxlib and will install both packages when installing them. I'd suggest doing the same in both cases (either with or without jax+jaxlib, that part I don't really care)

docs/source/installation.md
twiecki, line 17:
```suggestion
If you wish to enable sampling using the JAX backend via NumPyro or BlackJAX, the following should also be installed:
```

twiecki: I don't think it's that experimental anymore.

fonnesbeck: I assumed it was experimental until it was integrated into `pm.sample` ... happy to drop it though,

docs/source/installation.md
fonnesbeck: Actually, its the reverse! Numpyro will install jax and jaxlib by default, but blackjax will not. I will switch them.

twiecki (approved):

twiecki:
```

 Note that JAX is not directly supported on Windows systems at the moment.
-
Error: The process '/opt/hostedtoolcache/Python/3.10.4/x64/bin/pre-commit' failed with exit code 1

```

OriolAbril (approved):

PR #5874, opened 2022-06-09
explicit error about empirical approx

ferrine:
Currently the Empirical approximation is impossible initialize with inference data. This happens due to missing variables in the output. While histogram requires transformed variables (e.g. `a_log__`), inference data only has untransformed ones (e.g. `a`). This PR makes this error explicit with a recommendation to solve the issue.


michaelosthege:
I rebaes, fixed the pre-commit, made the error message more detailed and mentioned it in the release notes.

Also I think this can/should eventually be refactored: #5884

michaelosthege (approved):

ferrine:
Yes, I agree. Refactoring is a good next step

Issue #5323, opened 2022-01-07
Update Docker image to `v4`


michaelosthege:
As discussed in the meeting earlier today, it is now time to update the Docker image: https://hub.docker.com/u/pymc

# Short-term goals
* [x] Update the `Dockerfile` to a fresh base image (done in #5881)
* [ ] Create a GitHub Action to build & push the Docker image. (#5939)
* [x] Install dependencies to NOT get the `NumPy C-API` warning... (done in #5881)
* [x] Install graphviz/pydot such that `pm.model_to_graphviz()` doesn't raise errors. (done in #5881)
* (optional, b/c difficult) Install graphviz dependencies such that it can render to a file 👇

```python
with pymc.Model() as pmodel:
    pymc.Normal()
fp = pathlib.Path("model.pdf")
graph = pymc.model_to_graphviz(pmodel)
graph.render(format="pdf", filename=str(fp))
assert fp.exists()
```

# Simplifications/Constraints
* Python minor versions don't matter much these days, shall we build it for just `3.9` for now?
* We'll want a "latest" build and one per release? For `4.0.0b1` that'll be a manually built one, but for upcoming ones we can automate it?

Tagging @canyon289 @AustinRochford @bwengals who have relevant insight.

I can also contribute a (stripped-down) `environment.yml` to start from 👇

Known limitations:
* Suffers from `NumPy C-API` warning
* Graphviz/pydot install to the degree that `pm.model_to_graphviz()` doesn't raise, but rendering to fixes crashes.

```yaml
channels:
- defaults
- conda-forge
dependencies:
- conda-build
- libblas
- mkl-service
- numpy>=1.15.0
- pip
# To fix pydot/graphviz (?)
- pango
- pydot
# Limit graphviz upper version because of https://stackoverflow.com/a/69997168/4473230
- python-graphviz<=0.18
- scipy>=1.2.0
# To be discussed:
- jupyterlab
- statsmodels
- watermark
- pip:
  - matplotlib>=3.*
  - pandas>=1.*
  # PyMC installs more reliably with pip
  # as long as important dependencies are installed with conda already
  - git+https://github.com/pymc-devs/pymc@9d4691c4ceaa18dd50081db32f38fd1310359d48
  # To be discussed:
  - scikit-learn
  - openpyxl
```

When this issue is dealt with we can also close #3160 + #4386 as fixed and wontfix respectively.

ricardoV94:
The second issue looks wrong

[...]

PR #5881, opened 2022-06-12
Update dockerfile to PyMC v4

danhphan:
This PR updates the dockerfile to PyMC v4, which addresses some parts of the issue #5323
- Update the Dockerfile to a fresh base image for PyMC 4.0.0
- Fix NumPy `blas` warnings
- Run pm.model_to_graphviz(model) successfully

Todos:
- Update docs on how to build and run the docker image
- Create a GitHub Action to build & push the Docker image (need guidance) 

Thanks.


danhphan:
The docker guideline could be somethings like:

### Build docker image

```
clone git@github.com:pymc-devs/pymc.git
cd pymc
bash scripts/build_container.sh
```
This building process may take a while. After the script successful finished, there should be a docker image named `pymc` in your machine. 
### How to run
You can now run the `pymc` docker container from `jupyter notebook` or from `bash`:

```
# Run with jupyter notebook
bash scripts/run_container_bash.sh
# Run with bash
bash scripts/run_container_bash.sh
```


The current docker document was commented in [CONTRIBUTING.md](https://github.com/pymc-devs/pymc/blob/main/CONTRIBUTING.md?plain=1). Should we update the docker guideline here in the CONTRIBUTING.md file? Or should it be moved into the documentation website like other parts (pull-request guideline, pre-commit guideline, ...), and then link to that docs.

cc @ricardoV94 @michaelosthege @twiecki @OriolAbril 

Thank you.

scripts/Dockerfile
```diff
@@ -1,18 +1,55 @@
...
+WORKDIR /home/jovyan/workspace
+
+# For running from bash
+SHELL ["/bin/bash","-c"]
+RUN echo "conda activate pymc-dev-py39" >> ~/.bashrc && \
+    source ~/.bashrc
+
+# For running from jupyter notebook
+EXPOSE 8888
+CMD ["conda", "run", "--no-capture-output", "-n", "pymc-dev-py39", "jupyter","notebook","--ip=0.0.0.0","--port=8888","--no-browser","--allow-root"]
```
michaelosthege: Can `--allow-root` be removed, since the user was switched?

michaelosthege (commented):
Looks great!

The CI pipeline will be an important addition, but let's do that on a follow-up PR.

OriolAbril:
> The current docker document was commented in [CONTRIBUTING.md](https://github.com/pymc-devs/pymc/blob/main/CONTRIBUTING.md?plain=1). Should we update the docker guideline here in the CONTRIBUTING.md file? Or should it be moved into the documentation website like other parts (pull-request guideline, pre-commit guideline, ...), and then link to that docs.

It should be on the docs, not on CONTRIBUTING.md (it could have a link there though if we think it is relevant enough)

scripts/Dockerfile
danhphan: Yes agree, this makes sense. Let me remove it :)

scripts/Dockerfile
```diff
@@ -1,18 +1,55 @@
...
-ARG SRC_DIR=.
+LABEL name="pymc"
+LABEL description="Environment for PyMC version 4"
 
-MAINTAINER Austin Rochford <austin.rochford@gmail.com>
+ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 
-ADD $SRC_DIR /home/jovyan/
-RUN /bin/bash /home/jovyan/scripts/create_testenv.sh --global --no-setup
+USER root
```
canyon289: Containers should not run as root. Is there a reason this is needed?

https://theorangeone.net/posts/containers-as-root/

scripts/run_container_bash.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash
+
+SRC_DIR=${SRC_DIR:-`pwd`}
```
canyon289: Can this script be merged with run_container_jupyter.sh? Theyre so similar it seems itd be easier to maintain one, rather than two

canyon289: Actually can all three bash scripts be merged? One with a nice CLI would be better for both users and maintenance

canyon289:
Great stuff, thanks updating the dockerfile @danhphan!

scripts/run_container_bash.sh
danhphan: Yes, I have combine them into one script, and also update the Docker documents.

scripts/Dockerfile
danhphan: Hi agree, it has been removed :)

danhphan:
Hi, thank you all for the suggestions, I have:
- removed the root user and `--allow-root` from the Dockerfile,
- combined the three bash scripts into one, 
- updated the Docker documents in docs folder and also linked it in the [CONTRIBUTING.md](https://github.com/pymc-devs/pymc/blob/main/CONTRIBUTING.md?plain=1),
- removed obsolete scripts ([create_testenv.sh](https://github.com/pymc-devs/pymc/blob/main/scripts/create_testenv.sh) and [install_miniconda.sh](https://github.com/pymc-devs/pymc/blob/main/scripts/install_miniconda.sh)), but not sure if it is still used anywhere else?

docs/source/contributing/docker_container.md
```diff
@@ -0,0 +1,21 @@
+(pr_checklist)=
```
OriolAbril: This is defining a target to which we can then link from anywhere in the docs to generate a link that points here. The `pr_checklist` label is already in use so it can't be used and it should be something specific to this file

docs/source/contributing/docker_container.md
```diff
@@ -0,0 +1,21 @@
...
+
+We have provided a Dockerfile which helps for isolating build problems, and local development.
+Install [Docker](https://www.docker.com/) for your operating system, clone this repo, then
+run `./scripts/docker_container.sh build` to build a `pymc` docker image.
+
+After successfully building the docker image, you can start a local docker container called `pymc` from `bash` or from [`jupyter`](http://jupyter.org/) notebook server running on port 8888.
+
+```bash
+$  bash scripts/docker_container.sh bash # running the container with bash
+$  bash scripts/docker_container.sh bash # running the container on jupyter notebook
```
OriolAbril: I would remove the `$` so people can use the copy button that appears on code blocks directly. And both commands look the same, I think the 2nd one should use jupyter as argument or something of the sort.

docs/source/contributing/docker_container.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+The repo will be running the code from your local copy of `pymc`, so it is good for development. You may also use it to run the test suite, with
+
+```bash
+$  bash scripts/docker_container.sh bash # logon to the container
+$  cd ./pymc/tests
+$  . ./../../scripts/test.sh # takes a while!
```
OriolAbril: In addition to the dollar sign comment, do we use these `scripts/test.sh` in CI? Shouldn't the advise be to use pytest commands explicitly like we explain on https://www.pymc.io/projects/docs/en/stable/contributing/running_the_test_suite.html? We also recommend not running the whole test suite all at once unless necessary which that would seem to point against.

docs/source/contributing/docker_container.md
danhphan: Hi, got it.

docs/source/contributing/docker_container.md
danhphan: Yes, this content is from the old docs. I will remove them then. Also, not sure if the `scripts/test.sh` is used in any CI. If it is not used, we should also remove this script.

scripts/Dockerfile
```diff
@@ -1,18 +1,31 @@
...
+    mamba install -c conda-forge -y pymc" && \
+    conda clean --all -f -y
 
-ENV PYTHONPATH $PYTHONPATH:"$HOME"
+# Fix PkgResourcesDeprecationWarning
+RUN pip install --upgrade --user setuptools==58.3.0
+
+#Setup working forder
+RUN mkdir /home/jovyan/workspace
+WORKDIR /home/jovyan/workspace
```
bwengals: I can double check myself tomorrow, but I *think* the jupyter base image makes a directory `/home/jupyter/work` that you're supposed to use as a mount target.  If so, I think you could remove 20-22.  (also very small spelling nit: forder -> folder on line 20)

scripts/Dockerfile
```diff
@@ -1,18 +1,31 @@
-FROM jupyter/minimal-notebook
+FROM jupyter/base-notebook:python-3.9.12
```
bwengals, line 1: Good call on switching to `base-notebook`.  Any reason to fix the python version?  I think it might be preferable to have the container test build on "latest", just so no one needs to remember to bump versions.

scripts/Dockerfile
danhphan: Hi yes, I will check it as well :)

scripts/Dockerfile
danhphan: Yes, using `base-notebook` saves some space. The image with a specific python 3.9 version will make it consistent to the python version in [environment-dev-py39.yml](https://github.com/pymc-devs/pymc/blob/main/conda-envs/environment-dev-py39.yml), which will be the only dev environment left after https://github.com/pymc-devs/pymc/pull/5842 will be merged. This image is also updated regularly on Dockerhub I think. 

It also work with latest `base-notebook`. I have tested in before, however the python version will be 3.10 (if I am not wrong), and our tests seem not ready with that yet https://github.com/pymc-devs/pymc/issues/5209 

I am happy to change it to the `latest` `base-notebook` image if you guys prefer this option.

michaelosthege: we can also set up dependabot to automatically bump this version number when updates become available (I wanted open that PR once this one is merged)

bwengals: Those are both good reasons to keep the version fixed, sounds great to me.

twiecki:
Is this good to merge?

canyon289:
No, we're waiting on Bills comment about working directory

OriolAbril:
And my comments about docs still need to be addressed it looks like

scripts/Dockerfile
bwengals: Yup just build the container, everything works great and mounts correctly.  There is both `work` and `workspace` in there though.  It's mentioned very briefly in their docs here: https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#additional-runtime-configurations.   It'd be a little neater to remove `RUN mkdir /home/jovyan/workspace` and use `WORKDIR /home/jovyan/work` instead.

Also, just a heads up that `jovyan` *should* be replaced with `$NB_USER` in your Dockerfile, since that is [something that's configurable too](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#user-related-configurations).  I think it adds unnecessary complication to `scripts/docker_container.sh` though for this case, but might come in handy in other contexts.  Just want to mention if by chance you hadn't seen it!

scripts/docker_container.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# COMMAND can be either `build` or `bash` or `jupyter`
+if [[ $COMMAND = 'build' ]]; then
+   docker build \
+      -t ${CONTAINER_NAME} \
+      -f $SRC_DIR/scripts/Dockerfile $SRC_DIR
+
+elif [[ $COMMAND = 'bash' ]]; then
+   docker run -it -v $SRC_DIR:/home/jovyan/workspace --rm  --name ${CONTAINER_NAME} ${CONTAINER_NAME} bash
+else
```
bwengals, line 22: Might be nice to have an explicit elif for the `jupyter` option, not necessary though.  I also think it'd be nice to keep the bit at the end that [launches the notebook in the browser automatically](https://github.com/pymc-devs/pymc/blob/66215ace64b73f7c4ff2cc431ced9154e33da5e3/scripts/start_container.sh#L30-L32) but that's just a "nice to have" and could be a later PR.

bwengals:
I tried running it with all three options, `build`, `bash`, `jupyter`, everything works as expected!  Nicely done.  Other than the docs thing Oriol mentioned and the `workspace` -> `work` I think it's just about ready.

scripts/docker_container.sh
danhphan: Hi yes, the $COMMAND is set to `jupyter` as default in `COMMAND="${1:-jupyter}"`. So, after the docker image is built, we can run with `jupyter` as a default option.

danhphan:
Hi, I have updated the docs and replaced `workspace` with `work`. Hopefully it is alright now :) Thank you

michaelosthege (approved):
Looks good to me.
There's no build or run script for Windows though, but we can also add that later.

twiecki (approved):

danhphan:
Hi @michaelosthege , do you mean a script to create a Windows docker image? maybe we should not create a script for building a Windows-based docker image, since most docker image are linux-based, and most cloud systems use linux os.  In case of users that have Windows os with docker on their machine, they can also use the ubuntu image we have here to build and run the docker container. Thanks.

michaelosthege:
No I mean not a Windows Docker image, but a `docker_container.bat` with the same functionality as `docker_container.sh`.

But it's not critical since `docker_container.sh` is mostly a convenience thing.
The same can be achieved by modifying the commands below:

```bash
# Build the image, tagged as "pymc:local"
docker build -t pymc:local .

# Run in Jupyter notebook mode
docker run -it -p 8888:8888 -v %cd%:/home/jovyan/work --rm --name pymcJupyter pymc:local

# Run in interactive bash mode
docker run -it -v %cd%:/home/jovyan/work --rm --name pymcBash pymc:local bash
```

danhphan:
Yes, totally agree :)

PR #5882, opened 2022-06-12
Align advertised `Metropolis` `stats_dtypes` with changes from 1e7d91f

michaelosthege:
In https://github.com/michaelosthege/mcbackend/pull/27 I tried to upgrade to `4.0.0` but the CI failed.

The reason for the failure is an incorrect `stats_dtypes["accepted"]` which was not updated with the changes in 1e7d91f.

------

+ [x] No breaks

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,6 +1,10 @@
 # Release Notes
 
-## PyMC 4.0.0
+## PyMC 4.0.1 (vNext)
++ Fixed an incorrect entry in `pm.Metropolis.stats_dtypes` (see #5582).
++ ...
+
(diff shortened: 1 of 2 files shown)


ricardoV94:
Should we be testing this somewhere here?

ricardoV94 (approved):

michaelosthege:
> Should we be testing this somewhere here?

We should refactor the whole sampler stats API and test should be part of that..

ricardoV94:
> > Should we be testing this somewhere here?
> 
> We should refactor the whole sampler stats API and test should be part of that..

Can you open an issue for remembering to test this?

PR #5887, opened 2022-06-13
Cast rounded input replacement to original discrete dtype in sample_smc

ricardoV94:
Solves the problem described in https://discourse.pymc.io/t/model-works-with-nuts-prior-predictive-but-not-with-smc/9585/4

### DIFF
--- a/pymc/smc/smc.py
+++ b/pymc/smc/smc.py
@@ -567,7 +567,7 @@ def _logp_forw(point, out_vars, in_vars, shared):
             if in_var.dtype in discrete_types:
                 float_var = at.TensorType("floatX", in_var.broadcastable)(in_var.name)
                 new_in_vars.append(float_var)
-                replace_int_input[in_var] = at.round(float_var)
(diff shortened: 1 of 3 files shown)


pymc/smc/smc.py
```diff
@@ -567,7 +567,7 @@ def _logp_forw(point, out_vars, in_vars, shared):
             if in_var.dtype in discrete_types:
                 float_var = at.TensorType("floatX", in_var.broadcastable)(in_var.name)
                 new_in_vars.append(float_var)
-                replace_int_input[in_var] = at.round(float_var)
+                replace_int_input[in_var] = at.round(float_var).astype(in_var.dtype)
```
ricardoV94, line 570: We only had a test for unobserved bernoulli, but its logp does not care if the value is discrete or not. The categorical logp on the other hand requires it to be discrete as it does some indexing operation.

michaelosthege (approved):
LGTM

PR #5889, opened 2022-06-13
Compute some basic Slice sample stats

ricardoV94:


michaelosthege (approved):
LGTM

Eventually, we should check the dtypes when they're emitted, but the test you implemented is a start.

PR #5893, opened 2022-06-13
Remove self-directing arrow in observed nodes

### LINKED ISSUE #5892: Self-pointing arrow with observed nodes
```python
import pymc as pm
import numpy as np

data = np.random.normal(loc=5., scale=1.5, size=[1000,])

with pm.Model() as model:
    mu = pm.Normal(name="mu", mu=0., sigma=5.)
    y = pm.Normal(name="y", mu=mu, sigma=3., observed=data)
    
pm.model_to_graphviz(model)
```

yields:

Screen Shot 2022-06-13 at 11 03 44 PM

Notice the arrow from the `y` node to itself because `data` is a NumPy array and hence does not have a name.

larryshamalama:
Closes #5892.

One question in the code review and some tests in `tests/test_model_graph.py` relocated in the same file.

### DIFF
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -100,7 +100,7 @@ class ModelGraph:
             if hasattr(var.tag, "observations"):
                 try:
                     obs_name = var.tag.observations.name
-                    if obs_name:
+                    if obs_name and obs_name != var_name:
                         input_map[var_name] = input_map[var_name].difference({obs_name})
                         input_map[obs_name] = input_map[obs_name].union({var_name})
                 except AttributeError:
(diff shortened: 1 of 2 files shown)


pymc/model_graph.py
```diff
@@ -100,7 +100,7 @@ def make_compute_graph(
             if hasattr(var.tag, "observations"):
                 try:
                     obs_name = var.tag.observations.name
-                    if obs_name:
+                    if obs_name and obs_name != var_name:
```
larryshamalama, line 103: will `obs_name` ever be `None`?

michaelosthege: I don't know. I'd keep the check, also because at least the type hint on that `.name` will be `Optional[str]`, so removing the check could lead to `mypy` complaints.

michaelosthege (approved):

pymc/model_graph.py
larryshamalama: Sounds good 👍

PR #5903, opened 2022-06-17
Rename pm.Constant to pm.DiracDelta

### LINKED ISSUE #5716: Rename Constant to DiracDelta and add docstring warning about sampling it
A problem with `pm.Constant` recently cropped up here: https://discourse.pymc.io/t/typeerror-when-combining-constant-and-normal-distribution-in-function-within-deterministic-node/9219

`Constant` should only be used in special cases like Mixtures, not as a direct sampled variable.  We should make this point clear in the docstrings, mentioning constant python/numpy values can be used in a PyMC models and if traces are desired `pm.Deterministic` can be used. If the goal is to show up in graphviz, `pm.ConstantData` can be used.

We can probably raise a UserWarning in `pm.Constant.__new__` to the same end.

cluhmann:
Fixes #5716, renaming pm.Constant as pm.DiractDelta.  This is currently a basic find-and-replace in both the source and tests.  In the original issue, there was mention of adding a user warning about use of this distribution (i.e., sampling is hard).  However, pm.Constant had no `__new__` method, so I wasn't sure where to add it.  Happy to do so once I know where.


twiecki:
I think we should deprecate.

ricardoV94:
> I think we should deprecate.

What do you mean, have an alias with a warning for a while?

twiecki:
Yeah.

ricardoV94:
> Yeah.

I agree

ricardoV94:
> In the original issue, there was mention of adding a user warning about use of this distribution (i.e., sampling is hard). However, pm.Constant had no `__new__` method, so I wasn't sure where to add it. Happy to do so once I know where.

Let's not add it to the new renamed distribution, hopefully the name is good enough to dissuade people. We could add a warning message in docstrings.

For the deprecation, you will have to do something like:

```python
class Constant:
  def __new__(cls, *args, **kwargs):
    warnings.warn("bla bla blah", FutureWarning)
    return DiracDelta(*args, **kwargs)

  @classmethod
  def dist(cls, *args, **kwargs):
    warnings.warn("blah blah blah", FutureWarning)
    return DiracDelat.dist(*args, **kwargs)
```

pymc/distributions/discrete.py
```diff
@@ -1411,6 +1412,23 @@ def logcdf(value, c):
         )
 
 
+class Constant:
```
ricardoV94, line 1415: Can you add a test where you call `Constant` and confirm the warning is emitted and a DiracDelta is returned?

OriolAbril:
API docs at https://github.com/pymc-devs/pymc/blob/main/docs/source/api/distributions/discrete.rst also need to be updated

cluhmann:
> API docs at https://github.com/pymc-devs/pymc/blob/main/docs/source/api/distributions/discrete.rst also need to be updated

Should the now-deprecated pm.Constant be documented at all or should it be hidden to (further) discourage its use?

ricardoV94:
> > API docs at https://github.com/pymc-devs/pymc/blob/main/docs/source/api/distributions/discrete.rst also need to be updated
> 
> Should the now-deprecated pm.Constant be documented at all or should it be hidden to (further) discourage its use?

Hidden I think

pymc/tests/test_distributions.py
```diff
@@ -1729,9 +1730,15 @@ def test_poisson(self):
...
         )
 
+    def test_diracdeltadist(self):
+        check_logp(DiracDelta, I, {"c": I}, lambda value, c: np.log(c == value))
+        check_logcdf(DiracDelta, I, {"c": I}, lambda value, c: np.log(value >= c))
+
     def test_constantdist(self):
-        check_logp(Constant, I, {"c": I}, lambda value, c: np.log(c == value))
-        check_logcdf(Constant, I, {"c": I}, lambda value, c: np.log(value >= c))
+        with pytest.warns(FutureWarning):
```
```suggestion
        with pytest.warns(FutureWarning, match="..."):
```

ricardoV94: This test can be at the end of the file, it shouldn't be in this generic class that checks for logp/logcdf stuff

ricardoV94 (approved):
LGTM

michaelosthege:
Hehe, I'm on the train and was just reading through the updates here, wanting to review..
Suddenly the "open" changes to "merged". Okay :p

PR #5908, opened 2022-06-18
refactor sampling_jax postrocessing to avoid jit

ferrine:
In #5664 I enabled postprocessing to choose between cpu/gpu or the default device. That time I implemented that using `jax.jit(..., backend=device)` API. However, in some cases jitting is not desirable and causes errors. In this pull request I maintain the functionality but remove jitting.

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -6,6 +6,8 @@ import warnings
 from functools import partial
 from typing import Callable, Dict, List, Optional, Sequence, Union
 
+import aesara
+
 from pymc.initial_point import StartDict
 from pymc.sampling import RandomSeed, _get_seeds_per_chain, _init_jitter
 
@@ -136,9 +138,9 @@ def _get_log_likelihood(model: Model, samples, backend=None) -> Dict:
     """Compute log-likelihood for all observations"""
     data = {}
     for v in model.observed_RVs:
-        v_elemwise_logp = model.logp(v, sum=False)
-        jax_fn = get_jaxified_graph(inputs=model.value_vars, outputs=v_elemwise_logp)
-        result = jax.jit(jax.vmap(jax.vmap(jax_fn)), backend=backend)(*samples)[0]
+        v_elemwise_logpt = model.logpt(v, sum=False)
+        jax_fn = get_jaxified_graph(inputs=model.value_vars, outputs=v_elemwise_logpt)
+        result = jax.vmap(jax.vmap(jax_fn))(*jax.device_put(samples, jax.devices(backend)[0]))[0]
         data[v.name] = result
     return data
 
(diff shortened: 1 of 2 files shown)


ferrine:
How come?
https://github.com/pymc-devs/pymc/runs/6948826840?check_suite_focus=true#step:4:74

pymc/sampling_jax.py
```diff
@@ -136,9 +138,9 @@ def _get_log_likelihood(model: Model, samples, backend=None) -> Dict:
     """Compute log-likelihood for all observations"""
     data = {}
     for v in model.observed_RVs:
-        v_elemwise_logp = model.logp(v, sum=False)
-        jax_fn = get_jaxified_graph(inputs=model.value_vars, outputs=v_elemwise_logp)
-        result = jax.jit(jax.vmap(jax.vmap(jax_fn)), backend=backend)(*samples)[0]
+        v_elemwise_logpt = model.logpt(v, sum=False)
+        jax_fn = get_jaxified_graph(inputs=model.value_vars, outputs=v_elemwise_logpt)
+        result = jax.vmap(jax.vmap(jax_fn))(*jax.device_put(samples, jax.devices(backend)[0]))[0]
```
ricardoV94: loglikelihood should be fine with JIT, as non-elemwise logp has to be jittable for the model to sample in the first place

ferrine: it does not matter here imo

ricardoV94: It should still be faster with jit even if we call it only once according to folks on the slack

pymc/sampling_jax.py
```diff
@@ -134,13 +134,10 @@ def _sample_stats_to_xarray(posterior):
...
 def _get_log_likelihood(model: Model, samples, backend=None) -> Dict:
     """Compute log-likelihood for all observations"""
-    data = {}
-    for v in model.observed_RVs:
-        v_elemwise_logp = model.logp(v, sum=False)
-        jax_fn = get_jaxified_graph(inputs=model.value_vars, outputs=v_elemwise_logp)
-        result = jax.jit(jax.vmap(jax.vmap(jax_fn)), backend=backend)(*samples)[0]
-        data[v.name] = result
-    return data
+    elemwise_logpt = [model.logp(v, sum=False)[0] for v in model.observed_RVs]
```
ricardoV94:
```suggestion
    elemwise_logp = model.logp(model.observed_RVs, sum=False)
```

ricardoV94 (changes requested):
This does the same in a single Aeppl call, which can be much faster

pymc/sampling_jax.py
```diff
@@ -134,13 +134,10 @@ def _sample_stats_to_xarray(posterior):
...
+    jax_fn = get_jaxified_graph(inputs=model.value_vars, outputs=elemwise_logpt)
```
ricardoV94: We got rid of the `t` suffix just recently
```suggestion
    jax_fn = get_jaxified_graph(inputs=model.value_vars, outputs=elemwise_logp)
```

ricardoV94 (approved):

ricardoV94:
Going to merge as the JAX tests already passed. Thanks @ferrine!

PR #5909, opened 2022-06-19
Fix `compile_fn` bug and reduce return type confusion

### LINKED ISSUE #5867: Bug in pm.compile_fn

As discussed in #5865, `pm.compile_fn` is incorrect as implemented:

```python
import pymc as pm

with pm.Model() as model:
    x = pm.Normal("x")
    pm.compile_fn(x)
```

Fails with `TypeError` because `Model.compile_fn` expects `mode` as keyword argument:

```python

[<ipython-input-11-13dd2931074e>](https://localhost:8080/#) in <module>()
      5     x = pm.Normal("x")
      6 
----> 7     pm.compile_fn(x)

[/usr/local/lib/python3.7/dist-packages/pymc/model.py](https://localhost:8080/#) in compile_fn(outs, mode, point_fn, model, **kwargs)
   1826     """
   1827     model = modelcontext(model)
-> 1828     return model.compile_fn(outs, mode, point_fn=point_fn, **kwargs)
   1829 
   1830 

TypeError: compile_fn() takes 2 positional arguments but 3 positional arguments (and 1 keyword-only argument) were given
```

</details>

[The fix is simple](https://github.com/pymc-devs/pymc/pull/5865), but I don't know the code base enough to add a sensible test.

* PyMC/PyMC3 Version:
* Aesara/Theano Version:
* Python Version:
* Operating system:

michaelosthege:
This takes the bugfix from #5865 by @dfm albeit without the test case.

I looked into the test case and couldn't quite figure out why it's not working.
But if we were to test these module-level functions that are just forwarding to the `Model` methods, we should be doing that with mocks..

While I was at it, I also deprecated `Model.compile_fn(point_fn=True)` in favor of having `Model.compile_fn` vs. `Model.compile_point_fn`.
The motivation here is that switching the return type via a kwarg is really terrible from a typing perspective.

The `Model.compile_point_fn() -> PointFunc` signature on the other hand is less confusing.

## Incompatible Changes
+ None

## Compatible Changes
+ `Model.compile_fn(point_fn=True)` was deprecated in favor of a new `Model.compile_point_fn()` method

## Bugfixes
+ `pm.compile_fn` now correctly forwards the `mode` kwarg (closes #5867)
+ Docstring formatting
+ Type hints

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -48,6 +48,7 @@ from aesara.tensor.sharedvar import ScalarSharedVariable
 from aesara.tensor.var import TensorConstant, TensorVariable
 
 from pymc.aesaraf import (
+    PointFunc,
     compile_pymc,
     convert_observed_data,
     gradient,
@@ -640,7 +641,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         vars: Optional[Union[Variable, Sequence[Variable]]] = None,
         jacobian: bool = True,
         sum: bool = True,
-    ):
+    ) -> PointFunc:
         """Compiled log probability density function.
 
         Parameters
@@ -660,7 +661,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         self,
         vars: Optional[Union[Variable, Sequence[Variable]]] = None,
         jacobian: bool = True,
-    ):
+    ) -> PointFunc:
         """Compiled log probability density gradient function.
 
         Parameters
@@ -677,7 +678,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         self,
         vars: Optional[Union[Variable, Sequence[Variable]]] = None,
         jacobian: bool = True,
-    ):
+    ) -> PointFunc:
         """Compiled log probability density hessian function.
 
         Parameters
@@ -1597,7 +1598,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         mode=None,
         point_fn: bool = True,
         **kwargs,
-    ) -> Union["PointFunc", Callable[[Sequence[np.ndarray]], Sequence[np.ndarray]]]:
+    ) -> Union[PointFunc, Callable[[Sequence[np.ndarray]], Sequence[np.ndarray]]]:
         """Compiles an Aesara function
 
         Parameters
@@ -1911,16 +1912,6 @@ def Point(*args, filter_model_vars=False, **kwargs) -> Dict[str, np.ndarray]:
         for k, v in d.items()
         if not filter_model_vars or (get_var_name(k) in map(get_var_name, model.value_vars))
     }
-
-
-class PointFunc:
-    """Wraps so a function so it takes a dict of arguments instead of arguments."""
-
-    def __init__(self, f):
-        self.f = f
-
-    def __call__(self, state):
-        return self.f(**state)
 
 
 def Deterministic(name, var, model=None, dims=None, auto=False):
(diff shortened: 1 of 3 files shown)


ricardoV94 (commented):
I don't think the point_fn distinction is useful enough to warrant separate methods, and specially a more verbose default one.

I understand the typing concerns, but then again... this is Python.

pymc/model.py
```diff
@@ -425,6 +426,16 @@ def profile(self):
         return self._aesara_function.profile
 
 
+class PointFunc:
```
ricardoV94: We can just put it in utils or aesaraf instead of moving it to the top of the module. Suggesting this because I was planning to add a dispatch for aesara dprint method as well...

michaelosthege:
Do we need the `PointFunc` at all? I mean it's just a matter of `f(params)` vs. `f(**params) `, right?

I'd really prefer to reduce the type confusion, because this function is used in a lot of places. Having well defined types here will help typing throughout the entire codebase.

ricardoV94:
Yeah let's get rid of the silly wrapper. How would we go about doing it?

michaelosthege:
I moved `PointFunc` to `aesaraf.py`.

I also looked into removing `PointFunc` alltogether, but this would alter the return type of `compile_logp`, `compiled_dlogp` and `compile_d2logp`.
So I'd leave it like this..

pymc/model.py
```diff
@@ -1627,9 +1632,43 @@ def compile_fn(
             )
 
         if point_fn:
+            warnings.warn(
+                "Model.compile_fn(..., point_fn=True) is deprecated. Use `Model.compile_point_fn(...)` instead.",
+                DeprecationWarning,
+            )
```
ricardoV94: Why not start deprecating away from it altogether?

```suggestion
            warnings.warn(
                "Model.compile_fn(..., point_fn=True) is deprecated. In a future release it will no longer be possible to obtain a PointFunc.
                FutureWarning,
            )
```

And instead of creating the new verbose methods, let's just make sure we always call `point_fn = False` internally and adjust the calls that use these functions (which you don't have to do on your own of course!)

ricardoV94: Or if it's too fast we should start adding a `point_fn` kwarg to the user facing functions (compile_logp, and so on). This way we can issue a warn about the future breaking change and allow users to start setting it to False and have their code ready for when we finally remove that option :/

ricardoV94:
Regardless of what we decide we still need to have a test that calls these functions. The fact that the original PR test failed makes this even more important.

michaelosthege:
@ricardoV94 I neither have the time to start refactoring here, nor to figure out why the test Dan wrote ran into that error. I looked into it this morning, but couldn't figure out why.

The __only__ reason why I picked up the issue is that this is a blocker for `4.0.1` which I want to get released ASAP.
I see no reason to delay this fix, or to start a big refactoring at this time.

ricardoV94:
Then I suggest you limit the PR to the bugfix alone (and test)

michaelosthege:
That original test failed not because of the forwarding of the Mode kwarg. IMO we should instead just remove those module-level functions and stick with the methods that we're using anyway.

But it's Monday again and I have other priorities. I can't justify investing more time here.
If there's anybody who can get this test working without dumping an hour of valuable time it's you, @ricardoV94. I'll be happy to review that and learn.

ricardoV94 (approved):

michaelosthege (commented):
Thanks for adding the test! Now that I see the solution it does make sense that inputs need to be specified..

Let's merge when the test go green 👍

PR #5842, opened 2022-06-03
Remove some dev envs for older python versions.

twiecki:
…ove pandas from envs.

**Thank your for opening a PR!**

Before you proceed, please make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).

Depending on what your PR does, here are a few things you might want to address in the description:
+ [ ] what are the (breaking) changes that this PR makes?
+ [ ] important background, or details about the implementation
+ [ ] are the changes—especially new features—covered by tests and docstrings?
+ [ ] [consider adding/updating relevant example notebooks](https://github.com/pymc-devs/pymc-examples)
+ [ ] right before it's ready to merge, mention the PR in the RELEASE-NOTES.md

Issue #5074, opened 2021-10-13
Rethink bundling of requirements


OriolAbril:
I want to take this opportunity and rethink all these conda environments we have. I understand they are helpful and convenient to create our dev environments and they are also used for ci. However, at least in my case, the extra work and burden on updating 4 (8 now) nearly equivalent files with a pre-commit check that seems to only partially work (i.e. pandas has a lower bound in 3.7 and 3.8 but not in 3.9 nor in requirements-dev.txt) is clearly larger than the benefit I get. Moreover, issues with these files, installation and pre-commit blocked #5060. #5062 was only needed because of #5060 being blocked.

I think there are several issues at hand regarding all this.

* No single source of truth. From what I understand, the pre-commit script somehow takes all conda env files and from them generates the pip file. However, it seems like conda files don't need to be equivalent between them (see for example from  #5057 how nbsphinx was on some conda files but not all of them, as well as several versions not matching between files).
* Overly bloated and non-specific dependency files. We have 2/3 (still not completely understand the new test file but it seems like a step in the right direction)  types of requirements files, library requirements (user facing), dev and test ones. But we use those files to do **much** more than 3 types of things. We generate local envs for devs, we set up CI envs in github acions for testing, we set up the env on rtd to build documentation... So here are some things that we are doing and I think we should avoid as they also make our ci more brittle and prone to conda finding version conflicts:
  * We are installing sphinx and sphinx extensions everywhere, even on github actions
  * We are installing pytest on rtd
  * We are installing libraries like watermark or sphinx-autobuild also everywhere, but these are only used locally! watermark is only needed when writing/running example notebooks but not to build the docs that contain such notebooks. sphinx-autobuild is only needed to get an auto-updating preview of the docs _locally_.
* Most of us (or maybe even all of us) have no idea about what some of these dependencies are for. I for example have no idea what `cachetools` is for. I think that this is both issue and a consequence of the previous two, but I think it's important to put that on the table, because in my opinion, unless we make an effort in explaining all that in the developer guide and keep it up to date it will only get worse.

Ideas for possible solutions:
[...]

PR #5911, opened 2022-06-20
Get rid of python-version specific conda environments

ricardoV94:
This will probably fail spectacularly...

Related to #5074 
Supersedes #5842


.github/workflows/tests.yml
```diff
@@ -88,12 +88,12 @@ jobs:
...
         uses: actions/cache@v3
         env:
-          # Increase this value to reset cache if environment-test-py37.yml has not changed
+          # Increase this value to reset cache if environment-test.yml has not changed
           CACHE_NUMBER: 0
         with:
           path: ~/conda_pkgs_dir
           key: ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{
-            hashFiles('conda-envs/environment-test-py37.yml') }}
+            hashFiles('conda-envs/environment-test.yml') }}
```
ricardoV94: Probably safer to add the python version to the cache, but I am afraid devs will forget to update it when they change the `python-version` below. Any YML/BASH magicians, feel free to push something that uses variables... (or just reply in comment what I should do)

conda-envs/environment-dev.yml
```diff
@@ -30,6 +30,8 @@ dependencies:
 - pydata-sphinx-theme
 - pytest-cov>=2.5
 - pytest>=3.0
+- python-graphviz
+- scipy>=1.4.1
 - sphinx-copybutton
```
OriolAbril, line 32: We can also update the file and move sphinx-design to the conda section here instead of the pip one below. It is already on conda forge: https://anaconda.org/conda-forge/sphinx-design

ricardoV94:
Seems to be working

michaelosthege (approved):
Lots of deleted lines. I like it.

twiecki (approved):

Issue #4981, opened 2021-09-04
Remove unused `Model.__init__` kwargs without warnings


michaelosthege:
## Description of your problem

The two kwargs `model` and `aesara_config` of the `Model.__init__` are not accessed.

They don't have any effect, but there's also no warning or docstring about it.

## Versions and main components

* PyMC3 Version: `main`

itsguneetsingh:
I'm a little new here. Could maybe tell me where exactly I would find the Model.__init__ (like in which directory?)

Sayam753:
Yes. The related Model class is in `pymc3/model.py` and [here](https://github.com/pymc-devs/pymc3/blob/main/pymc3/model.py#L627-L635) are `__init__` lines.

itsguneetsingh:
I want to try this issue. could anyone please assign me this issue?

itsguneetsingh:
Thanks @Sayam753

Sayam753:
Awesome. Assigned you the issue.

itsguneetsingh:
If I have this right, I just have to delete "model" and "aesara_config" present on line no. 630 and 631. Is that it?

Sayam753:
Yes. Along with cleanup of the docstrings would work.

itsguneetsingh:
Hey, made some changes could you maybe have a look? https://github.com/pymc-devs/pymc3/pull/5013

PR #5915, opened 2022-06-20
Fix 'model' and 'aesara_config' kwargs for pm.Model

ferrine:
Fixes

```python
with pm.Model(aesara_config=dict(mode="JAX")):
    ...
```

and 

```python
with pm.Model(model=parent):
    ...
```

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -542,7 +542,11 @@ class Model(WithMemoization, metaclass=ContextMeta):
         name="",
         coords=None,
         check_bounds=True,
+        *,
+        aesara_config=None,
+        model=None,
     ):
+        del aesara_config, model  # used in __new__
         self.name = self._validate_name(name)
         self.check_bounds = check_bounds
 
--- a/pymc/tests/test_model.py
+++ b/pymc/tests/test_model.py
@@ -1014,3 +1014,20 @@ def test_compile_fn():
     result_expect = func(state)
 
     np.testing.assert_allclose(result_compute, result_expect)
+
+
+def test_model_aesara_config():
+    assert aesara.config.mode != "JAX"
+    with pm.Model(aesara_config=dict(mode="JAX")) as model:
+        assert aesara.config.mode == "JAX"
+    assert aesara.config.mode != "JAX"
+
+
+def test_model_parent_set_programmatically():
+    with pm.Model() as model:
+        x = pm.Normal("x")
+
+    with pm.Model(model=model):
+        y = pm.Normal("y")
+
+    assert "y" in model.named_vars


ricardoV94:
I don't think we are using that anywhere...? All functions that end up compiling stuff accept `mode`

ricardoV94:
> I don't think we are using that anywhere...? All functions that end up compiling stuff accept `mode`

Meaning we should just remove it if it is indeed never used.

ferrine:
> > I don't think we are using that anywhere...? All functions that end up compiling stuff accept `mode`
> 
> Meaning we should just remove it if it is indeed never used.

I tried to run VI using Jax and failed to do so (couple of more issues to be solved), I hoped this method would work for me as it was implemented in v3.

ricardoV94:
I just know it was removed because of the release notes. I don't know how it was used before and I don't remember seeing it being used anywhere in the library since I started working on it.

You can of course use JAX or NUMBA by setting the global aesara compile flag or passing `mode="JAX"` to `pm.sample` or whichever method you are using.

ricardoV94:
I was referring to this line: https://github.com/pymc-devs/pymc/blob/121a5ae0005d7e55cf613fbd8d943e1857e512a3/RELEASE-NOTES.md?plain=1#L125

But maybe it was just renamed, not removed? I see it is used in the ContextMeta

ferrine:
> I was referring to this line:
> 
> https://github.com/pymc-devs/pymc/blob/121a5ae0005d7e55cf613fbd8d943e1857e512a3/RELEASE-NOTES.md?plain=1#L125
> 
> But maybe it was just renamed, not removed? I see it is used in the ContextMeta

I failed to find an issue about this change. Neither removing model nor aesara/theano_config.

ferrine:
Oh, another query gave this: https://github.com/pymc-devs/pymc/issues/4981

ricardoV94:
> Oh, another query gave this: https://github.com/pymc-devs/pymc/issues/4981

Do you think that was a misunderstanding (ignoring the `__new__` path)? CC @michaelosthege

ferrine:
> > Oh, another query gave this: #4981
> 
> Do you think that was a misunderstanding (ignoring the **new** path)? CC @michaelosthege

The arguments might seem to be unused, but they are passed to `__new__`, if missing in `__init__` you get errors

ricardoV94:
> > > Oh, another query gave this: #4981
> > 
> > 
> > Do you think that was a misunderstanding (ignoring the **new** path)? CC @michaelosthege
> 
> The arguments might seem to be unused, but they are passed to `__new__`, if missing in `__init__` you get errors

Yeah, it seems to be the case, and your new tests clearly show it has an effect.

michaelosthege (changes requested):
Is `__new__` automatically called before `__init__`???

The `theano_config` kwarg was intentionally removed without replacement. It looks like it was forgotten to make changes in `__new__`.

We should
- either re-introduce the kwargs in `__init__` __while raising a `TypeError`__ with instructions to use Aesara config functions directly (like we do in tests)
- XOR remove it also from `__new__`

ferrine:
> Is `__new__` automatically called before `__init__`???

Yes, it is called before `__init__`. I'm fine to remove `aesara_config`, and keep the `model` kwarg

ricardoV94:
> > Is `__new__` automatically called before `__init__`???
> 
> Yes, it is called before `__init__`. I'm fine to remove `aesara_config`, and keep the `model` kwarg

+1 for less magic stuff behind the scenes

michaelosthege (approved):
I'm not a fan of keeping an API that is seldom used, but it looks like the removal was incomplete and this PR fixes it.

At some point we might want to remove `aesara_config` entirely, but that's out of scope for now.

So let's merge this in order to get `4.1.0` over the finish line?

PR #5919, opened 2022-06-22
Review docstrings checkmarcked as best practice

OriolAbril:
Minor fixes to docstrings marked as best practices already so they can serve as example in the coming open source working sessions.

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -231,7 +231,7 @@ class GaussianRandomWalk(distribution.Continuous):
 
         .. warning:: init will be cloned, rendering them independent of the ones passed as input.
 
-    steps : tensor_like of int, optional
+    steps : int, optional
         Number of steps in Gaussian Random Walk (steps > 0). Only needed if shape is not
         provided.
     """
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -217,20 +217,21 @@ gaussianrandomwalk = GaussianRandomWalkRV()
 
 
 class GaussianRandomWalk(distribution.Continuous):
-    r"""Random Walk with Normal innovations
+    r"""Random Walk with Normal innovations.
 
     Parameters
     ----------
-    mu : tensor_like of float
-        innovation drift, defaults to 0.0
-    sigma : tensor_like of float, optional
-        sigma > 0, innovation standard deviation, defaults to 1.0
-    init_dist : unnamed distribution
-        Univariate distribution of the initial value, created with the `.dist()` API.
+    mu : tensor_like of float, default 0
+        innovation drift
+    sigma : tensor_like of float, default 1
+        sigma > 0, innovation standard deviation.
+    init_dist : Distribution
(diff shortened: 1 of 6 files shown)


pymc/sampling.py
```diff
@@ -2277,14 +2277,14 @@ def draw(
 
     Parameters
     ----------
-    vars : Variable or iterable of Variable
+    vars : RandomVariable or iterable of RandomVariable
```
ricardoV94: Everything is TensorVariables in Aesara, the RandomVariables are TensorVariables whose Op is a RandomVariable, but they are not a type in itself

OriolAbril: But here only variables whose Op is RandomVariable is valid? Is that right? Or any TensorVariable would work?

If the RandomVariable Op is needed, we should then update the description or use a glossary definition as "type" like we do for `tensor_like` to explain this.

pymc/distributions/timeseries.py
```diff
@@ -217,20 +217,21 @@ def rng_fn(
...
+    sigma : tensor_like of float, default 1
+        sigma > 0, innovation standard deviation.
+    init_dist : Distribution
+        Unnamed univariate distribution of the initial value. Unnamed refers to distributions
+         created with the ``.dist()`` API.
 
         .. warning:: init will be cloned, rendering them independent of the ones passed as input.
 
-    steps : int, optional
+    steps : tensor_like of int, optional
```
OriolAbril: assuming this can be tensor_like from a quick look at the code as it uses `as_tensor_variable` like mu and sigma. Not 100% sure though.

ricardoV94: sounds correct, only thing is it needs to be a discrete scalar.

pymc/sampling.py
ricardoV94: It doesn't need to be RandomVariables, any graph that can be evaluated without inputs is fine. But the purpose is indeed RandomVariables...

OriolAbril: It looks like we should use TensorVariable, thanks for the review.

In case it helps with the context or update some docs, I was under the impression when writing this above that: all RandomVariables are TensorVariables, but not all TensorVariables are RandomVariables, and that this function didn't work on just any TensorVariable, only on this RandomVariables subset of TensorVariables.

ricardoV94: I understand the impression, all examples are like that. The simplest exception is any deterministic that depends on RVs like `at.exp(at.random.normal)`, which is a perfectly valid function output.

pymc/distributions/timeseries.py
OriolAbril:
```suggestion
    steps : int, optional
```

pymc/sampling.py
OriolAbril:
```suggestion
    vars : TensorVariable or iterable of TensorVariable
```

OriolAbril:
Should be ready to merge now. If something else still needed fixing please take over to get it merged before the open source working sessions

michaelosthege (approved):
Nice one!

PR #5928, opened 2022-06-25
Correct the order of `rvs` sent to `compile_dlogp` in `find_MAP`

### LINKED ISSUE #5923: Variables in logp and dlogp functions used in `find_MAP` are not always aligned
In this reproducible example, sampling from the model gives the correct
parameter estimates while `pymc.find_MAP()` does not.
(I just installed the latest pymc version using conda in a new environment on Ubuntu)

I am modeling weight as a function of height, here is some simulated data:
```python
import numpy as np
from scipy import stats
import pymc
n_individuals = 1000
alpha = 50
beta = 0.3
sigma = 5
height = np.random.uniform(130, 170, n_individuals)
mu = alpha + beta * (height-height.mean())
weight = stats.norm.rvs(loc=mu, scale=sigma, size=n_individuals)
```

When the prior for alpha is centered on the true value of alpha,
find_MAP() correctly estimates the values for alpha, beta and sigma:
```python
with pymc.Model() as model:
    alpha = pymc.Normal('alpha', mu=50, sigma=10)
    beta = pymc.Lognormal('beta', mu=0, sigma=1)
    sigma = pymc.Uniform('sigma', lower=0, upper=10)
    mu = alpha + beta * (height-height.mean())
    weight = pymc.Normal('height', mu=mu, sigma=sigma, observed=weight)
    mean_q = pymc.find_MAP()
print(mean_q)
```

However, when the true value of alpha differs from the prior mean,
beta and sigma are estimated to be 0:
```python
import numpy as np
from scipy import stats
import pymc
n_individuals = 1000
alpha = 30
beta = 0.3
sigma = 5
height = np.random.uniform(130, 170, n_individuals)
mu = alpha + beta * (height-height.mean())
weight = stats.norm.rvs(loc=mu, scale=sigma, size=n_individuals)
with pymc.Model() as model:
    alpha = pymc.Normal('alpha', mu=50, sigma=10)
    beta = pymc.Lognormal('beta', mu=0, sigma=1)
    sigma = pymc.Uniform('sigma', lower=0, upper=10)
    mu = alpha + beta * (height-height.mean())
    weight = pymc.Normal('height', mu=mu, sigma=sigma, observed=weight)
    mean_q = pymc.find_MAP()
print(mean_q)
```

Sampling from the model still returns the correct values for all three parameters:
```python
with model:
    trace = pymc.sample(1000, tune=1000)
pymc.summary(trace)
```
So I am wondering if this is a bug or if find_MAP() is supposed to fail in the second case.
If latter, I would have expected some error or warning being put out by the function.

quantheory:
This PR is a quick fix for the bad behavior of `find_MAP` in v4 (closes #5923).

When using a gradient-based method in `scipy.minimize`, we need the order of derivatives in the Jacobian to match the order of the variables in `x`. Since `x0.point_map_info` is used to define this order in the rest of the function, I just use it to set the order of variables in the `rvs` argument sent to `compile_dlogp`, which fixes the issue.

I've also added `test_find_MAP_issue_5923`, which fails for the old code and passes in this commit. (In fact, the original `test_find_MAP` test would also fail if we used a starting point slightly farther from the MAP.)

### DIFF
--- a/pymc/tuning/starting.py
+++ b/pymc/tuning/starting.py
@@ -110,9 +110,9 @@ def find_MAP(
     start = ipfn(seed)
     model.check_start_vals(start)
 
-    var_names = {var.name for var in vars}
+    vars_dict = {var.name: var for var in vars}
     x0 = DictToArrayBijection.map(
-        {var_name: value for var_name, value in start.items() if var_name in var_names}
+        {var_name: value for var_name, value in start.items() if var_name in vars_dict}
     )
 
     # TODO: If the mapping is fixed, we can simply create graphs for the
@@ -120,7 +120,7 @@ def find_MAP(
     compiled_logp_func = DictToArrayBijection.mapf(model.compile_logp(jacobian=False), start)
     logp_func = lambda x: compiled_logp_func(RaveledVars(x, x0.point_map_info))
 
-    rvs = [model.values_to_rvs[value] for value in vars]
+    rvs = [model.values_to_rvs[vars_dict[name]] for name, _, _ in x0.point_map_info]
     try:
         # This might be needed for calls to `dlogp_func`
         # start_map_info = tuple((v.name, v.shape, v.dtype) for v in vars)
--- a/pymc/tests/test_starting.py
+++ b/pymc/tests/test_starting.py
@@ -98,6 +98,32 @@ def test_find_MAP():
     close_to(map_est2["sigma"], 1, tol)
 
 
+def test_find_MAP_issue_5923():
+    # Test that gradient-based minimization works well regardless of the order
+    # of variables in `vars`, and even when starting a reasonable distance from
+    # the MAP.
+    tol = 2.0**-11  # 16 bit machine epsilon, a low bar
+    data = np.random.randn(100)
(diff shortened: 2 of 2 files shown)


pymc/tuning/starting.py
```diff
@@ -120,7 +120,8 @@ def find_MAP(
     compiled_logp_func = DictToArrayBijection.mapf(model.compile_logp(jacobian=False), start)
     logp_func = lambda x: compiled_logp_func(RaveledVars(x, x0.point_map_info))
 
-    rvs = [model.values_to_rvs[value] for value in vars]
+    vars_dict = {var.name: var for var in vars}
```
ricardoV94: we could replace the `var_names` set above with this dict so that we don't create two objects with similar purposes.

ricardoV94:
Looks perfect! I left a small refactoring suggestion above, let me know what you think.

quantheory:
Makes sense. I've made the change you suggested.

ricardoV94 (approved):

ricardoV94:
Thanks @quantheory!

PR #5931, opened 2022-06-26
pass names from RV dims to change_rv_size

### LINKED ISSUE #5669: Add variable names to size from dims
Would make dprint more readable

```python
import aesara
import pymc as pm

with pm.Model(coords=dict(cities=range(5))) as m:
  x = pm.Normal("x", dims=("cities"))

aesara.dprint(x)
```

```
normal_rv{0, (0, 0), floatX, False}.1 [id A] 'x'   
 |RandomStateSharedVariable(<RandomState(MT19937) at 0x7FF40F93C540>) [id B]
 |SpecifyShape [id C] ''   
 | |MakeVector{dtype='int64'} [id D] ''   
 | | |<TensorType(int64, ())> [id E]  <- Cryptic variable!
 | |TensorConstant{(1,) of 1} [id F]
 |TensorConstant{11} [id G]
 |TensorConstant{0} [id H]
 |TensorConstant{1.0} [id I]
```

Instead we could give it a name (doing it manually here for illustration):

```python
with pm.Model(coords=dict(cities=range(5))) as m:
  size = aesara.shared(len(m.coords), name="cities_dim")
  x = pm.Normal("x", size=(size,))

aesara.dprint(x)
```
```
normal_rv{0, (0, 0), floatX, False}.1 [id A] 'x'   
 |RandomStateSharedVariable(<RandomState(MT19937) at 0x7FF408D8E140>) [id B]
 |SpecifyShape [id C] ''   
 | |MakeVector{dtype='int64'} [id D] ''   
 | | |cities_dim [id E]  <- Readable variable name
 | |TensorConstant{(1,) of 1} [id F]
 |TensorConstant{11} [id G]
 |TensorConstant{0} [id H]
 |TensorConstant{1.0} [id I]
```

flo-schu:
@canyon289 to resolve #5669, I have made some small changes mainly to aesaraf.py to include the dim names in the new_size tensor (https://github.com/pymc-devs/pymc/commit/6585f9021a17a7f0c9c833bf4700fab240d2164b)

This solves the issue mainly and `test_model` runs without errors. There are a few things that probably need to be addressed.

- do dims passed to `change_rv_size` need to be tested more rigorously?
- is None a good default name for the broadcasting tensor dimension?
- should information if the broadcasting comes from observed or other processes be included in 
the name
- things that I'm not aware of :)

## changes of the PR
- adds argument new_size_dims to change_rv_size in aesaraf.py
- tests if dims is None and wraps in tuple if RV has no dim
- creates new_size_name by adding dim name and decorations
- passes new_size_name as name argument to at.as_tensor method

### DIFF
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -145,6 +145,7 @@ def convert_observed_data(data):
 def change_rv_size(
     rv: TensorVariable,
     new_size: PotentialShapeType,
+    new_size_dims: Optional[tuple] = (None,),
     expand: Optional[bool] = False,
 ) -> TensorVariable:
     """Change or expand the size of a `RandomVariable`.
@@ -155,6 +156,8 @@ def change_rv_size(
         The old `RandomVariable` output.
     new_size
         The new size.
+    new_size_dims
+        dim names of the new size vector
     expand:
         Expand the existing size by `new_size`.
 
@@ -165,6 +168,10 @@ def change_rv_size(
         raise ShapeError("The `new_size` must be ≤1-dimensional.", actual=new_size_ndim)
     elif new_size_ndim == 0:
         new_size = (new_size,)
+
+    # wrap None in tuple, if new_size_dims are None
+    if new_size_dims is None:
+        new_size_dims = (None,)
 
     # Extract the RV node that is to be resized, together with its inputs, name and tag
     assert rv.owner.op is not None
@@ -180,9 +187,13 @@ def change_rv_size(
         size = shape[: len(shape) - rv_node.op.ndim_supp]
         new_size = tuple(new_size) + tuple(size)
 
+    # create the name of the RV's resizing tensor
+    # TODO: add information where the dim is coming from (obseverd, prior, ...)
+    new_size_name = f"Broadcast to {new_size_dims[0]}_dim"
+
     # Make sure the new size is a tensor. This dtype-aware conversion helps
     # to not unnecessarily pick up a `Cast` in some cases (see #4652).
-    new_size = at.as_tensor(new_size, ndim=1, dtype="int64")
+    new_size = at.as_tensor(new_size, ndim=1, dtype="int64", name=new_size_name)
 
     new_rv_node = rv_node.op.make_node(rng, new_size, dtype, *dist_params)
     new_rv = new_rv_node.outputs[-1]
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -266,7 +266,9 @@ class Distribution(metaclass=DistributionMeta):
 
         if resize_shape:
             # A batch size was specified through `dims`, or implied by `observed`.
-            rv_out = change_rv_size(rv=rv_out, new_size=resize_shape, expand=True)
+            rv_out = change_rv_size(
+                rv=rv_out, new_size=resize_shape, new_size_dims=dims, expand=True
+            )
 
         rv_out = model.register_rv(
             rv_out,


pymc/aesaraf.py
```diff
@@ -180,9 +187,13 @@ def change_rv_size(
         size = shape[: len(shape) - rv_node.op.ndim_supp]
         new_size = tuple(new_size) + tuple(size)
 
+    # create the name of the RV's resizing tensor
+    # TODO: add information where the dim is coming from (obseverd, prior, ...)
```
canyon289, line 191: Is this todo going to be implemented in this PR or is this for later?

canyon289:
Thanks much for the pr!
Are you planning to add tests?

pymc/aesaraf.py
```diff
@@ -145,6 +145,7 @@ def convert_observed_data(data):
 def change_rv_size(
     rv: TensorVariable,
     new_size: PotentialShapeType,
+    new_size_dims: Optional[tuple] = (None,),
```
ricardoV94, line 148: The name info should not be handled by this function. I think somewhere in the model we create constants or shared variables for dim sizes and there is where we should set the names.

ricardoV94 (changes requested):

pymc/aesaraf.py
flo-schu: yes, I think it would be helpful. Just wanted to make sure I'm on the right track before I spend more time on it

pymc/aesaraf.py
flo-schu: Could you give me a hint roughly where this should be? Before or after the call to change_rv_size?

ricardoV94: Could be done here where the dims variables are created: https://github.com/pymc-devs/pymc/blob/be048a495b32a09c49ec8d3e8a95d4f5e4290d2a/pymc/model.py#L1146-L1149

Both constants and shared variables accept names during construction.

flo-schu: @ricardoV94, I looked into it and when I set the names in the suggested lines, I receive only the following output from dprint:

```
normal_rv{0, (0, 0), floatX, False}.1 [id A] 'x'
 |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F99BD4B6960>) [id B]
 |TensorConstant{(1,) of 5} [id C]
 |TensorConstant{11} [id D]
 |TensorConstant{0} [id E]
 |TensorConstant{1.0} [id F]
```

only after passing the name to new_size in aesaraf.py::change_rv_size, I get the change
https://github.com/pymc-devs/pymc/blob/be048a495b32a09c49ec8d3e8a95d4f5e4290d2a/pymc/aesaraf.py#L185

```python
new_size = at.as_tensor(new_size, ndim=1, dtype="int64", name=new_size[0].name)
```

```
normal_rv{0, (0, 0), floatX, False}.1 [id A] 'x'
 |RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7FAD2928A960>) [id B]
 |cities_dim{(1,) of 5} [id C]
 |TensorConstant{11} [id D]
 |TensorConstant{0} [id E]
 |TensorConstant{1.0} [id F]
```

If this is okay, I'll implement the changes in this PR

ricardoV94:
This PR should now be a bit more straightforward due to #6063 Dims will now always be used when shape is not provided.

pymc/aesaraf.py
ricardoV94: This PR added a name for the mutable case, but not the constant one: https://github.com/pymc-devs/pymc/pull/6147/commits/f45ca4a20c4868693769f14892bb9aab7a3e93fe, I think adding it to the constant is all that's needed.

flo-schu: Thanks, I will try to give it another go in this week

flo-schu: @ricardoV94, I'm dancing around this PR for quite a bit now, but I'm not finding the time to do it. In the beginning of next year I would have more time, but if it bothers you to have an open PR for such a long time or someone else takes over, I won't mind if I don't get to finish it.

ricardoV94: No worries. The original issue lost some of it's relevance when we switched from defaulting resizable dims to constant dims. It's still nice but not urgent by any means. Feel free to come back to it whenever you find the time.

covertg:
Hi folks, I was interested in this so I took a look with the latest codebase. As @flo-schu was finding, it's not sufficient to simply add the dim name to the call to `pytensor.tensor.constant` here https://github.com/pymc-devs/pymc/blob/e1d36cacb2cfb946f56872ee03e8678dc6ccf7f4/pymc/model.py#L1066-L1069

Somewhere along the RV creation call stack, the name of the Variable holding dim length gets lost. I believe this is actually happening in `pytensor.tensor.random.utils.normalize_size_parameter()`, here https://github.com/pymc-devs/pytensor/blob/a1a805c9119e986be9f3e08591daf3698cede0ef/pytensor/tensor/random/utils.py#L136

With shared variables, the call to `as_tensor_variable` returns a MakeVector, which preserves the original variable(s) holding dim lengths. But with constant variables, `as_tensor_variable` squashes the dim length(s) into one TensorConstant and wipes out the names.

Using the following model,
```python
with pm.Model() as m:
    mutable = True
    d2 = pm.Data("data1", range(5), dims="d1", mutable=mutable)
    d1 = pm.Data("data2", range(7), dims="d2", mutable=mutable)
    x = pm.Normal("x", dims=("d1", "d2"))
```
Before that line in normalize_size_parameter, `dprint(size)` shows
```
Subtensor{int64} [id A]
 |Shape [id B]
 | |data1 [id C]
 |ScalarConstant{0} [id D]
Subtensor{int64} [id E]
 |Shape [id F]
 | |data2 [id G]
 |ScalarConstant{0} [id H]
```
and after,
```
MakeVector{dtype='int64'} [id A]
 |Subtensor{int64} [id B]
 | |Shape [id C]
 | | |data1 [id D]
 | |ScalarConstant{0} [id E]
 |Subtensor{int64} [id F]
   |Shape [id G]
   | |data2 [id H]
   |ScalarConstant{0} [id I]
```
With pymc patched as described above to pass along name to `pytensor.tensor.constant` when creating dims, we can try the same model but with `mutable=False`. That line in normalize_size_parameter changes
```
d1{5} [id A]
d2{7} [id B]
```
to
```
TensorConstant{[5 7]} [id A]
```
So, what to do?
If this issue is still of interest, would be glad to work on it given some advice.

covertg:
N.B. the tensorconstant name gets disappeared even if there is only one dim, a la
```python
>>> pt.as_tensor_variable((pt.constant(5, name="disappears"),)).name
None
```
Because normalize_size_parameter will always receive a tuple.

A couple naive thoughts include:
- We could pre-empt pytensor's (re)creation of TensorConstant by always turning the dim length information into a more complex Variable, somewhere close to the below line. For example call `pt.join(create_size)` before the call to `rv_op` so that the dim length is always a MakeVector.
    - https://github.com/pymc-devs/pymc/blob/e1d36cacb2cfb946f56872ee03e8678dc6ccf7f4/pymc/distributions/distribution.py#L388-L389
- Or perhaps allow the TensorConstants to get squashed into one, but preserve name information by concatenating the strings of the original names if available. (would by a pytensor change)

ricardoV94:
@covertg thanks for investigating. This issue was more relevant before when all dims used SharedVariable, and hence names were preserved. With TensorConstant's PyTensor is very liberal and happy to return you objects that don't link to the original ones as you found out. I think we can simply close the PR and issue. It was a nice to have, but doesn't seem worth any further trouble at this point.

PR #5952, opened 2022-07-05
correct docstrings in Binomial Class

SangamSwadiK:
**What is this PR about?**
This PR updates docstrings of Binomial for numpydoc compliance
Issue: #5459 

**Checklist**

## Major / Breaking Changes
None

## Bugfixes / New features
None

## Docs / Maintenance
Updates parameters of Binomial Class for numpydoc compliance as per #5459 

#DataUmbrellaPyMCSprint2022  @reshamas

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -82,6 +82,7 @@ class Binomial(Discrete):
     .. math:: f(x \mid n, p) = \binom{n}{x} p^x (1-p)^{n-x}
 
     .. plot::
+        :context: close-figs
 
         import matplotlib.pyplot as plt
         import numpy as np
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -107,11 +107,11 @@ class Binomial(Discrete):
 
     Parameters
     ----------
-    n : int
+    n : tensor_like of int
         Number of Bernoulli trials (n >= 0).
-    p : float
+    p : tensor_like of float
         Probability of success in each trial (0 < p < 1).
-    logit_p : float
+    logit_p : tensor_like of float
         Alternative log odds for the probability of success.
     """
     rv_op = binomial


reshamas:
@OriolAbril @cluhmann @twiecki @symeneses 

Some general questions I have when reviewing this:
1) Should updates be made at the class level or function level?

2) If it's class, then for line 145, should this be updated too?
a) `value: numeric` ==> `value : tensor-like of float`
b) description needs a fullstop at the end: `Aesara tensor` ==> `Aesara tensor.`

```
        Parameters
        ----------
        value: numeric
            Value(s) for which log-probability is calculated. If the log probabilities for multiple
            values are desired the values must be provided in a numpy array or Aesara tensor
```

3) There is no default for `value`.  Is that correct?

4) Interesting that the parameter name is `value`.  Something like `x` or `random_value` seems more intuitive, but I supposed that's based from my textbook experience.

symeneses:
Hi @reshamas,
Regarding to your questions:

1. It should be to the class level including all methods which have already a docstring.
2. Yes, that should be also modified and also in line 169.
3. Correct, there is no default value.

OriolAbril:
Thanks for the PR @SangamSwadiK! It looks great!

The only change needed to merge the PR is adding the context arg to the `plot` directive as shown in the 2nd bullet point of the [section independent comments](https://pymc-data-umbrella.xyz/en/latest/sprint/tutorials/docstring_tutorial.html#section-independent-comments).

---

@reshamas and other reviewers

1. As @symeneses said, updates should always be at class and function level, _generally_ including methods (functions defined inside a class) too. I say generally including the methods because for distributions (which is the case here) methods should be ignored. The [docstring guide page](https://pymc-data-umbrella.xyz/en/latest/sprint/tutorials/docstring_tutorial.html#edit-the-docstring) says:

   > You have to review section by section to make sure everything is well documented. If you have chosen a class that is not a distribution (unlike this example where we are working on the pymc.Uniform distribution), you should review the docstrings of all the methods (only if they already exist though, no need to write missing docstrings). Otherwise, you should work only on the function or class docstring. We will therefore ignore the docstrings of the logcdf, get_moment and dist methods.

   The reason for this is that these methods should not be used, the `pm.logp` or `pm.logcdf` functions should be used instead and they will call these methods (which might not be methods either in the near future). See for example https://github.com/pymc-devs/pymc/issues/5308.
1. If it were a not-distribution class, this should also be updated. Here however it is not necessary.
1. Yes, there is not default for `value`. And here there are no defaults for `n` or `p` but there are distributions whose parameters do have default values (defined in `.dist` method) so in those cases `value` would continue to have a default argument but the parameters that came after would do even if they are not present in the call signature. This is also related to the strange use (or maybe better _non-use_) that should be done of those methods and also adds to the reasons for ignoring those methods, at least  for now.
1. I wasn't here when that was written so I can't speak to the reasons that were used when defining this originally. I think it might be helpful to note that distributions can be used to define either prior or log likelihood terms so the tensor used as `value` is not always random from a Bayesian perspective.

SangamSwadiK:
I've updated with close-figs context.
Thanks !

reshamas:
Thanks @OriolAbril for explanation on those points.

OriolAbril (approved):

cluhmann (approved):
Sorry for the delay.  I was just double-checking the type of `n`.  But everything looks good to me!

Issue #5954, opened 2022-07-06
`pymc.sampling_jax.sample_blackjax_nuts` crashes with `chains=1`


aussetg:
## Description of your problem

`pymc.sampling_jax.sample_blackjax_nuts` crashes with `chains=1`

**Please provide a minimal, self-contained, and reproducible example.**
```python
# %%
import numpy as np
import pandas as pd
import pymc as pm
import pymc.sampling_jax
import pymc.distributions.transforms
import aesara.tensor as at
import aesara
np.set_printoptions(2)

# %%
data = pd.DataFrame(np.random.randn(100, 23))

# %%
K = 15
n_f = len(data.columns)

# %%
with pm.Model() as model:
    alpha = pm.Gamma("alpha", 0.7, 1.0)
    w = pm.StickBreakingWeights("w", alpha=alpha, K=K-1)

    cov = np.identity(n_f)

    mu = pm.MvNormal(
        "mu",
        mu=data.mean(axis=0),
        cov=cov,
        shape=(K, n_f),
    )

    components = [
        pm.MvNormal.dist(mu=mu[i, :], cov=cov, shape=(n_f,))
        for i in range(K)
    ]

    obs = pm.Mixture(
        "obs",
        w=w,
        comp_dists=components,
        observed=data,
    )

# %%
with model:
    trace = pymc.sampling_jax.sample_blackjax_nuts(target_accept=0.9, chains=1)
```

**Please provide the full traceback.**

```python
Compiling...
Traceback (most recent call last):
  File "/usr/lib64/python3.9/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 2, in <module>
  File "/home/guillaume/.cache/pypoetry/virtualenvs/tabular-T_K5n0PG-py3.9/lib/python3.9/site-packages/pymc/sampling_jax.py", line 301, in sample_blackjax_nuts
    init_params = [np.stack(init_params)]
  File "<__array_function__ internals>", line 180, in stack
  File "/home/guillaume/.cache/pypoetry/virtualenvs/tabular-T_K5n0PG-py3.9/lib64/python3.9/site-packages/numpy/core/shape_base.py", line 426, in stack
    raise ValueError('all input arrays must have the same shape')
ValueError: all input arrays must have the same shape
```

## Versions and main components

* PyMC/PyMC3 Version: PyMC4.1.1
* Aesara/Theano Version: 2.7.4
* Python Version: 3.9
[...]

PR #5969, opened 2022-07-11
Provide a fix for `sample_blackjax_nuts` failing with `chains=1` with prior parameters of different shapes

bherwerth:
The PR addresses #5954, which reported that calling `sample_blackjax_nuts` failed with `chains=1` when prior parameters had different shapes. A fix is provided by:
* A change in [sampling_jax.py](https://github.com/pymc-devs/pymc/blob/ab40dcc194ac9a358c690f0532841235b6cde13b/pymc/sampling_jax.py)  so that `init_params` has the same structure for `chains=1` as for `chains>1`
* Modification of the test case in [test_sampling_jax.py](https://github.com/pymc-devs/pymc/blob/ab40dcc194ac9a358c690f0532841235b6cde13b/pymc/tests/test_sampling_jax.py) so that it covers parameters with different shapes.
...

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- Fix bug in `sampling_jax.sample_blackjax_nuts` failing when called for one chain and prior parameters of different shape

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/sampling_jax.py
+++ b/pymc/sampling_jax.py
@@ -298,8 +298,7 @@ def sample_blackjax_nuts(
     )
 
     if chains == 1:
-        init_params = [np.stack(init_params)]
-        init_params = [np.stack(init_state) for init_state in zip(*init_params)]
+        init_params = [np.stack(init_state) for init_state in zip(init_params)]
 
     logprob_fn = get_jaxified_logp(model)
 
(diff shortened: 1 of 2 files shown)


twiecki:
Can you add a test for the `chains=1` case?

bherwerth:
Hi @twiecki ,

The test for `chains=1` was already there before my changes:

https://github.com/pymc-devs/pymc/blob/1e52d43d9514d137f4680eafa9386be3dafa0ac3/pymc/tests/test_sampling_jax.py#L31-L40

However, it did not fail because the prior parameters of the test case are both of the same shape. I changed the sigma parameter in the test case to be of shape (2,), cf. the diff on `test_sampling_jax.py`. With this change of the test case, the test fails before my change to `sampling_jax.py` and it works afterwards (when I run the test locally).

Please let me know in case you had in mind something else.

On the code coverage report, I don't understand the coverage change on `base_hmc.py` because I did not edit that file:
image

twiecki (approved):

twiecki:
Thanks @bherwerth!

PR #5981, opened 2022-07-16
Constrain priors with symmetric mass distribution

lucianopaz:
**What is this PR about?**
The `find_constrained_prior` function was optimising the difference between the upper and lower cdf values so that it matched a particular mass. This can be very brittle since there are multiple lower and upper cdf values that produce exactly the same mass. For example, [0.01, 0.96] will have 0.95 mass but [0.025, 0.975] will have it too. This PR proposes to use a different optimisation target so that the prior will be constrained to have the desired mass, and that the lower and upper cdf bounds will be symmetric.

**Checklist**

## Bugfixes / New features
- find_constrained_prior now attempts to find the parameters that lead to the desired mass between the lower and upper bounds, but also forces the lower and upper cdf values to be symmetric.

### DIFF
--- a/pymc/func_utils.py
+++ b/pymc/func_utils.py
@@ -11,8 +11,6 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
-import warnings
-
 from typing import Callable, Dict, Optional, Union
 
 import aesara.tensor as aet
@@ -33,6 +31,8 @@ def find_constrained_prior(
     init_guess: Dict[str, float],
     mass: float = 0.95,
     fixed_params: Optional[Dict[str, float]] = None,
+    mass_below_lower: Optional[float] = None,
+    **kwargs,
 ) -> Dict[str, float]:
     """
     Find optimal parameters to get `mass` % of probability
@@ -64,6 +64,11 @@ def find_constrained_prior(
         Dictionary of fixed parameters, so that there are only 2 to optimize.
         For instance, for a StudentT, you fix nu to a constant and get the optimized
         mu and sigma.
+    mass_below_lower : float, optional, default None
+        The probability mass below the ``lower`` bound. If ``None``,
+        defaults to ``(1 - mass) / 2``, which implies that the probability
+        mass below the ``lower`` value will be equal to the probability
+        mass above the ``upper`` value.
 
     Returns
     -------
@@ -71,6 +76,11 @@ def find_constrained_prior(
         The optimized distribution parameters as a dictionary.
         Dictionary keys are the parameter names and
         dictionary values are the optimized parameter values.
+
+    Notes
+    -----
+    Optional keyword arguments can be passed to ``find_constrained_prior``. These will be
+    delivered to the underlying call to :external:py:func:`scipy.optimize.minimize`.
 
     Examples
     --------
@@ -96,11 +106,31 @@ def find_constrained_prior(
             init_guess={"mu": 5, "sigma": 2},
             fixed_params={"nu": 7},
         )
+
+    Under some circumstances, you might not want to have the same cummulative
+    probability below the ``lower`` threshold and above the ``upper`` threshold.
+    For example, you might want to constrain an Exponential distribution to
+    find the parameter that yields 90% of the mass below the ``upper`` bound,
+    and have zero mass below ``lower``. You can do that with the following call
+    to ``find_constrained_prior``
+
+    .. code-block:: python
+
+        opt_params = pm.find_constrained_prior(
+            pm.Exponential,
+            lower=0,
+            upper=3.,
+            mass=0.9,
+            init_guess={"lam": 1},
+            mass_below_lower=0,
+        )
     """
     assert 0.01 <= mass <= 0.99, (
         "This function optimizes the mass of the given distribution +/- "
         f"1%, so `mass` has to be between 0.01 and 0.99. You provided {mass}."
     )
+    if mass_below_lower is None:
+        mass_below_lower = (1 - mass) / 2
 
     # exit when any parameter is not scalar:
     if np.any(np.asarray(distribution.rv_op.ndims_params) != 0):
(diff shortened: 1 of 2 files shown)


ricardoV94 (commented):
I think the reason for this rule was for positive distributions, so that you can specify lower=0, upper=x? I don't remember exactly.

If this or something similar we might want to allow both symmetric and assymetric rules.

ricardoV94:
Can you share the flaky example? I would imagine that with the 3 constraints (lower, upper, and mass in between), most distributions of up tp 3 parameters would be uniquely constrained

lucianopaz:
> I think the reason for this rule was for positive distributions, so that you can specify lower=0, upper=x? I don't remember exactly.
> 
> If this or something similar we might want to allow both symmetric and assymetric rules.

But with the previous rule, the optimisation target has no way of knowing that you want lower=0 and upper=mass. So it will still be underdetermined.

lucianopaz:
> Can you share the flaky example? I would imagine that with the 3 constraints (lower, upper, and mass in between), most distributions of up tp 3 parameters would be uniquely constrained

```python
pm.find_constrained_prior(
    pm.Gamma,
    lower=20,
    upper=700,
    mass=0.95,
    init_guess=dict(alpha=5.29, beta=0.023),
)
```

If I’m not missing anything, I’m just constraining lower and upper. The mass being in the lower upper interval is a side effect.

ricardoV94:
> If I’m not missing anything, I’m just constraining lower and upper. The mass being in the lower upper interval is a side effect.

Lower and upper are constants and, given those, it's optimizing the mass in between. 

Okay, so the problem (as you wrote) is that you can just shift the distribution left and right and keep the interval mass between upper and lower constant right?

Should we add a second constraint for the mass to the left of the first point which defaults to `(1-mass)/2` (your PR) but can also be set to (eg, 0) if you want to set an absolute lower bound for positive variables? 

I have the feeling that with positive distributions I would often want to think about stuff in the range (0-x) but maybe it's just me, or it makes things unnecessarily complicated.

lucianopaz:
> Should we add a second constraint for the mass to the left of the first point which defaults to `(1-mass)/2` (your PR) but can also be set to (eg, 0) if you want to set an absolute lower bound for positive variables?

Yeah, I can add that.

> I have the feeling that with positive distributions I would often want to think about stuff in the range (0-x) but maybe it's just me, or it makes things unnecessarily complicated.

Maybe when you’re working with an exponential distribution the [0, mass] range is more natural, but not for the gamma or the lognormal distributions. Anyway, we can include the option to set the lower bound.

aloctavodia:
This is one of the reasons that in PreliZ we use the maximum entropy constraint. https://preliz.readthedocs.io/en/latest/api_reference.html#preliz.maxent. My plan is to have a usable version of PreliZ this year.

lucianopaz:
> This is one of the reasons that in PreliZ we use the maximum entropy constraint. https://preliz.readthedocs.io/en/latest/api_reference.html#preliz.maxent. My plan is to have a usable version of PreliZ this year.

Yeah, entropy is better behaved than the `exp(logcdf)` at some edge. @aloctavodia, would there be any benefit of implementing the entropy of the distributions? Maybe it could be useful for aeppl?

aloctavodia:
Not sure if having the entropy of the distributions will be of general use. At this point I can only think of maximum entropy priors. And in a few months I think we will be ready to deprecate find_constrained_prior in favor of PreliZ.

pymc/func_utils.py
```diff
@@ -64,6 +64,11 @@ def find_constrained_prior(
         Dictionary of fixed parameters, so that there are only 2 to optimize.
         For instance, for a StudentT, you fix nu to a constant and get the optimized
         mu and sigma.
+    mass_below_lower : float, optional, default None
+        The probability mass below the ``lower`` bound. If ``None``,
+        defaults to ``(1 - mass) / 2``, which implies that the probability
+        mass below the ``lower`` value will be the equal to the probability
```
AlexAndorra:
```suggestion
        mass below the ``lower`` value will be equal to the probability
```

pymc/func_utils.py
```diff
@@ -129,39 +141,39 @@ def find_constrained_prior(
...
-        pm.math.exp(pm.logcdf(opt_dist, upper)) - pm.math.exp(pm.logcdf(opt_dist, lower))
-    ).eval()
-    if (np.abs(mass_in_interval - mass)) > 0.01:
-        warnings.warn(
-            f"Final optimization has {(mass_in_interval if mass_in_interval.ndim < 1 else mass_in_interval[0])* 100:.0f}% of probability mass between "
-            f"{lower} and {upper} instead of the requested {mass * 100:.0f}%.\n"
-            "You may need to use a more flexible distribution, change the fixed parameters in the "
-            "`fixed_params` dictionary, or provide better initial guesses."
-        )
-
```
AlexAndorra, line 166: Are we sure we don't need this check anymore @lucianopaz ?

pymc/func_utils.py
```diff
@@ -64,6 +64,11 @@ def find_constrained_prior(
         Dictionary of fixed parameters, so that there are only 2 to optimize.
         For instance, for a StudentT, you fix nu to a constant and get the optimized
         mu and sigma.
+    mass_below_lower : float, optional, default None
```
AlexAndorra, line 67: I would add a small example of why that would be useful

AlexAndorra (changes requested):
Looks better indeed, thanks @lucianopaz ! Just had a few comments before merging

pymc/func_utils.py
lucianopaz: Yes, we don't need it anymore. Now that the mass condition is an optimization constraint, the optimization will fail if the condition is not met.

pymc/func_utils.py
```diff
@@ -96,11 +106,31 @@ def find_constrained_prior(
             init_guess={"mu": 5, "sigma": 2},
             fixed_params={"nu": 7},
         )
+
+    Under some circumstances, you might not want to have the same cummulative
```
ricardoV94:
```suggestion
    Under some circumstances, you might not want to have the same cumulative
```

ricardoV94 (approved):
Looks great!

AlexAndorra (approved):
All good now, thanks @lucianopaz ! I'll merge once tests pass

PR #5998, opened 2022-07-22
Updating docstrings of distributions

vitaliset:
# What is this PR about?
In response to issue #5459. Me and @pibieta updated docstrings for some distribution classes inside the multivariate.py module, namely:
- LKJCholeskyCov
- LKJCorr
- MatrixNormal
- KroneckerNormal
- CAR

## Attention points
- Inside the KroneckerNormal class, the docstring types of `covs`, `chols` and `evds` were left as `list of arrays` and `list of tuples` because the proper type is being developed as @OriolAbril told us.
- Inside the KroneckerNormal class the type of `sigma` should be `float` as far as I understand but we were not sure because the old docstring called it a `variable`.
- Inside the CAR class, there was a `Numpy matrix` input that we changed to `ndarray of int`. Does this make sense?
- Inside the MatrixNormal class there were some `nxn array` that we changed to `nxn tensor_like of float`. Does that make sense or we should put something else?

# Docs / Maintenance
updated docstrings to follow the numpydoc standard

#DataUmbrellaPyMCSprint
cc: @reshamas

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -1850,7 +1850,7 @@ class KroneckerNormal(Continuous):
     ========  ==========================
     Support   :math:`x \in \mathbb{R}^N`
     Mean      :math:`\mu`
-    Variance  :math:`K = \bigotimes K_i` + \sigma^2 I_N
+    Variance  :math:`K = \bigotimes K_i + \sigma^2 I_N`
     ========  ==========================
 
     Parameters
@@ -1869,7 +1869,7 @@ class KroneckerNormal(Continuous):
         :math:`K_i = Q_i \text{diag}(v_i) Q_i'`. For example::
 
             v_i, Q_i = at.nlinalg.eigh(K_i)
-    sigma : scalar, variable
+    sigma : scalar, optional
         Standard deviation of the Gaussian white noise.
 
     Examples
@@ -2103,17 +2103,18 @@ class CAR(Continuous):
     ----------
     mu : tensor_like of float
         Real-valued mean vector
-    W : ndarray of int
+    W : (M, M) tensor_like of int
         Symmetric adjacency matrix of 1s and 0s indicating
-        adjacency between elements.
+        adjacency between elements. If possible, *W* is converted
+        to a sparse matrix, falling back to a dense variable.
+        :func:`~aesara.sparse.basic.as_sparse_or_tensor_variable` is
+        used for this sparse or tensorvariable conversion.
     alpha : tensor_like of float
         Autoregression parameter taking values between -1 and 1. Values closer to 0 indicate weaker
         correlation and values closer to 1 indicate higher autocorrelation. For most use cases, the
-        support of alpha should be restricted to (0, 1)
+        support of alpha should be restricted to (0, 1).
     tau : tensor_like of float
         Positive precision variable controlling the scale of the underlying normal variates.
-    sparse : bool, default=False
-        Determines whether or not sparse computations are used
 
     References
     ----------
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -918,7 +918,7 @@ class Wishart(Continuous):
     ----------
     nu : tensor_like of int
         Degrees of freedom, > 0.
-    V : array_like
+    V : tensor_like of float
         p x p positive definite matrix.
 
     Notes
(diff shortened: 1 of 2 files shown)


pymc/distributions/multivariate.py
```diff
@@ -1855,22 +1855,21 @@ class KroneckerNormal(Continuous):
...
-    covs: list of arrays
+    covs : list of arrays
         The set of covariance matrices :math:`[K_1, K_2, ...]` to be
         Kroneckered in the order provided :math:`\bigotimes K_i`.
-    chols: list of arrays
+    chols : list of arrays
         The set of lower cholesky matrices :math:`[L_1, L_2, ...]` such that
         :math:`K_i = L_i L_i'`.
-    evds: list of tuples
+    evds : list of tuples
```
OriolAbril, line 1866: @Sayam753 @purna135 Is `KroneckerNormal` one of the distributions being updated as part of your GSoC work? Has it been or will it be?

pymc/distributions/multivariate.py
```diff
@@ -1649,20 +1649,20 @@ class MatrixNormal(Continuous):
 
     Parameters
     ----------
-    mu: array
+    mu : tensor_like of float
         Array of means. Must be broadcastable with the random variable X such
         that the shape of mu + X is (m,n).
-    rowcov: mxm array
+    rowcov : mxm tensor_like of float, optional
```
OriolAbril: Also, can you update this to be 

```suggestion
    rowcov : (M, M) tensor_like of float, optional
```

and the same for the other similar cases. I know this didn't appear in the guide, but doing that we'd be following matplotlib's convention instead of creating a new one. Ref: https://matplotlib.org/stable/devel/documenting_mpl.html#parameter-type-descriptions. It is also not part of the numpydoc description but it also seems to be the convention used by numpy itself, see https://numpy.org/doc/stable/reference/generated/numpy.linalg.eig.html for example

Note: similar comment for @Sayam753 and @purna135. Not sure if it should be `(M, M)` or `(..., M, M)`, also, if it is M,M but will be updated to ..., M, M please keep M, M for now but remember to update it once it is improved 🚀

pymc/distributions/multivariate.py
Sayam753: Hi @OriolAbril 
Yes. The plan is to work on multivariate distributions in PyMC post mid-evaluations.

OriolAbril: Thanks! @vitaliset  it is ok then to leave the types as is and count on Sayam and Purna to update them as part of the planned improvements.

pymc/distributions/multivariate.py
Sayam753: Only 2 dimensions are supported for rowcov currently i.e. (M, M). @purna135 will work on adding batch support  post GSoC mid-evaluations to make it `(..., M, M)`.

vitaliset: I'll keep it bidimensional in the documentation for now and @purna135 can change it when the feature gets added. :)

vitaliset:
Regarding the `W` parameter of CAR and the `sigma` parameter of KroneckerNormal, do you think we should do something @OriolAbril?

pymc/distributions/multivariate.py
```diff
@@ -2102,18 +2101,18 @@ class CAR(Continuous):
 
     Parameters
     ----------
-    mu: array
+    mu : tensor_like of float
         Real-valued mean vector
-    W: Numpy matrix
+    W : ndarray of int
         Symmetric adjacency matrix of 1s and 0s indicating
         adjacency between elements.
```
OriolAbril: I am not familiar at all with the distribution. I suspect the "matrix" was used to indicate "at least 2d". So it might be something like `mu : (..., M) tensor_like ...` and `W : (..., M, M) tensor_like ...`. IIRC, @junpenglao wrote an example notebook about CAR, so he probably is able to help solve this.

Note: similarly to the other cases, it might be `M` and `(M, M)` for now with plans to be updated to taking batched inputs

OriolAbril:
Also, if you look at the preview, the `M` and `N` are rendered in pink, which is not good. Can you add them to the [`numpydoc_xref_ignore`](https://github.com/pymc-devs/pymc/blob/main/docs/source/conf.py#L65) set? This will make sure they are shown in bold black font.

pymc/distributions/multivariate.py
ricardoV94: Several of our multivariate distributions only work with matrix inputs, so unless someone tests it, it should remain a part of the docs. 

2D tensor_like or something like that could be used.

One of this year GSOC projects is focused on allowing multivariate distributions to broadcast arbitrarily, but this one will probably not be worked on.

OriolAbril: The goal is to document the actual accepted input, the main issue is I have no idea what that is. Not when it comes to shape nor when it comes to accepted types.

The current `numpy matrix` is somewere between ambiguous, wrong and concerning. One might think it means the only accepted inputs are those of type [numpy.matrix](https://numpy.org/doc/stable/reference/generated/numpy.matrix.html) a type not recommended to use that might even get removed at some point, it might also mean 2d numpy arrays which also seems odd, it probably accepts aesara tensors too, and being a symetric matrix it isn't really any 2d array/tensor, hence the `(M, M) tensor_like` (or `(..., M, M) tensor_like` if it already supports batched inputs).

If in addition there are shape restrictions between multiple parameters, they should also be documented. Seeing as this looks similar to a multivariate normal, where there are these restrictions, I suggested updating `mu` too together with the changes to `W`.

twiecki (approved):

twiecki:
Thanks @vitaliset!

PR #5986, opened 2022-07-19
Fix `pm.Interpolated` moment

larryshamalama:
Closes #5959.

Previously, the `moment` method of `pm.Interpolated` was $\sum_{i=1}^N x_i f(x_i)$ given finite `x_points` and their density values `pdf_points`. However, this can lead to the aforementioned issue as the distances between points in `x_points` are not accounted.

The code in this first commit needs to be adjusted for cases where `x_points` is not strictly increasing. Another commit (or two) should follow to account for this.

PR #6010, opened 2022-07-25
Proposal: Readd 3.7

canyon289:
Right now just checking to see if this works. If it does will discuss with you all if we can readd 3.7 support

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -417,3 +417,108 @@ jobs:
           env_vars: TEST_SUBSET
           name: ${{ matrix.os }} ${{ matrix.floatx }}
           fail_ci_if_error: false
+  ubuntu_37:
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
+        floatx: [float64]
+        python-version: ["3.7"]
+        test-subset:
+          - |
+            pymc/tests/test_util.py
+            pymc/tests/test_logprob.py
+            pymc/tests/test_aesaraf.py
+            pymc/tests/test_math.py
+            pymc/tests/test_posdef_sym.py
+            pymc/tests/test_ndarray_backend.py
+            pymc/tests/test_hmc.py
+            pymc/tests/test_func_utils.py
+            pymc/tests/test_shape_handling.py
+            pymc/tests/test_starting.py
+            pymc/tests/test_mixture.py
+
+          - |
+            pymc/tests/test_distributions.py
+
+          - |
+            pymc/tests/test_tuning.py
+            pymc/tests/test_shared.py
+            pymc/tests/test_types.py
+            pymc/tests/test_distributions_moments.py
+
+          - |
+            pymc/tests/test_modelcontext.py
+            pymc/tests/test_dist_math.py
+            pymc/tests/test_minibatches.py
+            pymc/tests/test_pickling.py
+            pymc/tests/test_transforms.py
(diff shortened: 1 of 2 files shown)


michaelosthege:
Can we do this in a way that does increase our CI time by a third?

For example, we could replace the `3.10` jobs with `3.7` and do the `3.10` test on `macos`?

ricardoV94:
github `macos` machines are really slow, so you have to be careful which tests you run with it.

twiecki:
So most likely we won't keep updating the version that's running on colab (or do folks update packages there frequently? The fact that it's 3.7 makes me think not so). In that case, we now saw that it works on 3.7 so we could just release with marking 3.7 as supported, get it on colab, and then disable it for the next release. It's not clean but we're making an exception here anyway.

canyon289:
> So most likely we won't keep updating the version that's running on colab (or do folks update packages there frequently? The fact that it's 3.7 makes me think not so). In that case, we now saw that it works on 3.7 so we could just release with marking 3.7 as supported, get it on colab, and then disable it for the next release. It's not clean but we're making an exception here anyway.

This is my thought exactly. Get one PyMC v4 release out there that were 100% confident works in a colab 3.7 so there is a path for "zero install" Python/PyMC v4 environment. After that remove 3.7 again, as power users will be working in their own environments.

Would releasing this specific commit of PyMC be acceptable? Are there any other bugfixes or things we want to sneak in before we lock in this specific codebase? I'll post this question on slack too

michaelosthege:
> Are there any other bugfixes or things we want to sneak in before we lock in this specific codebase?

Possibly #6011 because it fixes a bug in the plate model visualization which is disproportionally more relevant in colab notebooks.

twiecki:
Cool, let's do it!

larryshamalama:
I just marked #5986 ready for review. Should we also get that in before the release?

.github/workflows/tests.yml
```diff
@@ -417,3 +417,108 @@ jobs:
           env_vars: TEST_SUBSET
           name: ${{ matrix.os }} ${{ matrix.floatx }}
           fail_ci_if_error: false
+  ubuntu_37:
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
+        floatx: [float64]
+        python-version: ["3.7"]
```
ricardoV94, line 425: You can just change to `python-version: ["3.x", "3.7"]` in all the jobs. If the goal is to remove support again immediately after, that will be make it an easier commit to review.

ricardoV94 (changes requested):

.github/workflows/tests.yml
ricardoV94: You could have* (too late)

canyon289: we dont want to do that. We want to ensure the release works for all versions of PyMC that we want to support which includes 3.8, 3.9, 3.10 etc

ricardoV94: I think I didn't explain myself. Where you see e.g. `python-version=["3.8"]` before, you could have changed to `python-version=["3.7", "3.8"]` and both versions would have been tested. You could have done this for all existing tests, instead of copying the entirety of the first section of the file to just swap 3.8 to 3.7.

That would have been a much smaller PR and easier to review and revert. That's all

canyon289: that would have been way easier 🤦 

TIL. Thanks @ricardoV94

PR #6011, opened 2022-07-25
Create arrow to observation nodes subject to arbitrary `dtype` casting in `pm.model_to_graphviz`

### LINKED ISSUE #5795: Graphviz does not properly show observed `ConstantData` when subjected to intermediate Ops
```python
with pm.Model() as model:
    data = pm.ConstantData("data", [0, 0, 1, 1], dtype=np.int64)
    p = pm.Beta("p", 1, 1)
    bern = pm.Bernoulli("response", p, observed=data)
    
pm.model_to_graphviz(model)
```

yields

image

In other words, there should be an arrow from `response` to `data`. This is because `model["data"].tag.observations` is in fact an `Elemwise` operation rather than `data` as defined under the context manager. I don't think that this was broken from the recently merged PR #5634.

larryshamalama:
Closes #5795.

As described in the issue, when casting data to another `dtype`, a casting operation is performed and ignored by the `pm.model_to_graphviz` functionality.

However, without investigating further, I believe that this PR covers the scope of #5795, but there is an [important example](https://github.com/pymc-devs/pymc/issues/5795#issuecomment-1184236854) of another problem which seems to be caused by a different issue. I hope to have a look at the latter this Wednesday.

CC @drbenvincent

### DIFF
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -99,11 +99,13 @@ class ModelGraph:
             parent_name = self.get_parent_names(var)
             input_map[var_name] = input_map[var_name].union(parent_name)
 
-            while True:
+            if hasattr(var.tag, "observations"):
+                obs_node = var.tag.observations
+
                 # loop created so that the elif block can go through this again
                 # and remove any intermediate ops, notably dtype casting, to observations
-                if hasattr(var.tag, "observations"):
-                    obs_node = var.tag.observations
+                while True:
+
                     obs_name = obs_node.name
                     if obs_name and obs_name != var_name:
                         input_map[var_name] = input_map[var_name].difference({obs_name})
@@ -114,16 +116,14 @@ class ModelGraph:
                         # see issue 5795: https://github.com/pymc-devs/pymc/issues/5795
                         obs_node.owner
                         and isinstance(obs_node.owner.op, Elemwise)
-                        and isinstance(obs_node.owner.inputs[0].owner.op.scalar_op, Cast)
+                        and isinstance(obs_node.owner.op.scalar_op, Cast)
                     ):
                         # go to the beginning of the loop
                         # this comment is going up the graph to retrieve the node, i.e.
                         # observations here, is subject to dtype casting
-                        var.tag.observations = obs_node.owner.inputs[0].owner.inputs[0]
+                        obs_node = obs_node.owner.inputs[0]
                     else:
                         break
-                else:
-                    break
 
         return input_map
 
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -100,6 +100,8 @@ class ModelGraph:
             input_map[var_name] = input_map[var_name].union(parent_name)
 
             while True:
+                # loop created so that the elif block can go through this again
+                # and remove any intermediate ops, notably dtype casting, to observations
                 if hasattr(var.tag, "observations"):
                     obs_node = var.tag.observations
                     obs_name = obs_node.name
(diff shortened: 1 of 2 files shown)


pymc/model_graph.py
```diff
@@ -100,19 +100,22 @@ def make_compute_graph(
...
+                            var.tag.observations = obs_node.owner.inputs[0].owner.inputs[0]
+                    else:
                         break
-                    elif isinstance(obs_node.owner.op, Elemwise) and isinstance(
-                        obs_node.owner.inputs[0].owner.op.scalar_op, Cast
-                    ):
-                        # go to the beginning of the loop
-                        var.tag.observations = obs_node.owner.inputs[0].owner.inputs[0]
-                else:
+                except AttributeError:
```
michaelosthege: Where is the `AttributeError` coming from?
Couldn't it be solved with a `hasattr` in the same way like above?

twiecki:
@larryshamalama Would be great to get this one over the finish line for the new release!

pymc/model_graph.py
larryshamalama: Great point, I believe that it should. For some data containers, `obs_node` would node have an `owner` attribute. I am checking the tests locally and pushing if all passes.

pymc/model_graph.py
```diff
@@ -97,14 +99,25 @@ def make_compute_graph(
             parent_name = self.get_parent_names(var)
             input_map[var_name] = input_map[var_name].union(parent_name)
 
-            if hasattr(var.tag, "observations"):
-                try:
-                    obs_name = var.tag.observations.name
+            while True:
+                if hasattr(var.tag, "observations"):
```
canyon289: Add comment explaining logic block

pymc/model_graph.py
```diff
@@ -97,14 +99,25 @@ def make_compute_graph(
...
+                    obs_node = var.tag.observations
+                    obs_name = obs_node.name
                     if obs_name and obs_name != var_name:
                         input_map[var_name] = input_map[var_name].difference({obs_name})
                         input_map[obs_name] = input_map[obs_name].union({var_name})
-                except AttributeError:
-                    pass
+                        break
+                    elif (
+                        obs_node.owner
```
canyon289, line 117: Add comment

pymc/tests/test_model_graph.py
```diff
@@ -236,6 +256,10 @@ class TestUnnamedObservedNodes(BaseModelGraphTest):
     model_func = model_unnamed_observed_node
 
 
+class TestObservationDtypeCasting(BaseModelGraphTest):
+    model_func = model_observation_dtype_casting
```
canyon289, line 260: Why not just add the function here into the class directly?

canyon289:
Thanks @larryshamalama for providing this. Before we merge can we rebase on main as well so it gets tested by our 3.7 test runners that just got added?

pymc/tests/test_model_graph.py
larryshamalama: I just naively followed the structure of previous tests based on `BaseModelGraphTest`...

canyon289: Thanks, I see that now that I expanded past diffs. Dont worry about it on this PR, ill just openly say I dont believe this is the right pattern, it very much feels like we should be using fixtures to reference model setup blocks as pytest provides many more functionality and guarantees when we use its architecture correctly

Thanks for the PR!

larryshamalama:
Need a moment to fix another bug, marking this as draft temporarily

Edit: fixed! With this PR merged, [@drbenvincent's notebook](https://www.pymc.io/projects/examples/en/latest/causal_inference/excess_deaths.html) should display the right graph

pymc/model_graph.py
larryshamalama: Done!

pymc/model_graph.py
larryshamalama: Done!

pymc/tests/test_model_graph.py
larryshamalama: I'm not super familiar with the alternative approach that you're proposing. Maybe it could be something to improve at a later date. In short, what do you mean by using fixtures and why does it allow better usage of pytest? Is [this page](https://docs.pytest.org/en/6.2.x/fixture.html) what you are referring to? I'm going to guess that the answer to this is beyond the scope of this PR and can warrant a whole discussion...

michaelosthege (approved):

Issue #6009, opened 2022-07-25
Strange behavior of `pm.sample()` with respect to NUTS-specific keyword arguments


mschmidt87:
## Description of your problem

I noticed a strange behavior of the `pm.sample()` function in terms of specifying arguments for the NUTS sampler. As explained in the docstring, it is possible to specify `target_accept` as a direct arguments to `pm.sample()`, but for `max_treedepth` (and other arguments), this is not possible. Now, this leads to an inconsistency in the interface where `target_accept` is ignored when specified directly while `max_treedepth` is specified in a dictionary. 
I find this quite confusing because it means that specifying an argument changes the behavior of another argument. 

This is a minimal reproducer with comments explaining my observations:

**Please provide a minimal, self-contained, and reproducible example.**
```python
import numpy as np
import pymc as pm


rng = np.random.RandomState(123)

with pm.Model() as model:
    mu = pm.Normal("mu", mu=0, sigma=1)
    obs = pm.Normal("obs", mu=mu, sigma=1, observed=rng.standard_normal(100))

# Arguments passed to the NUTS sampler internally:
# target_accept = 0.99

with model:
    pm.sample(target_accept=0.99)

# Arguments passed to the NUTS sampler internally:
# max_treedepth = 15

with model:
    pm.sample(nuts={"max_treedepth": 15})

# Arguments passed to the NUTS sampler internally:
# max_treedepth = 15
# That means target_accept is not passed to the sampler but just silently dropped.

with model:
    pm.sample(nuts={"max_treedepth": 15}, target_accept=0.99)


# Arguments passed to the NUTS sampler internally:
# max_treedepth = 15, target_accept = 0.99
# Here, everything works as expected.
with model:
    pm.sample(nuts={"max_treedepth": 15, "target_accept": 0.99})
```

**Please provide any additional information below.**

## Versions and main components

* PyMC/PyMC3 Version: 4.1.3
* Aesara/Theano Version: 2.7.7
* Python Version: 3.9.11
* Operating system: Mac OS
* How did you install PyMC/PyMC3: pip

michaelosthege:
The second case sounds like an unintended bug.

Looking at the code (https://github.com/pymc-devs/pymc/blob/73eac989fba7c74064fd1c18b5b68e84bdc80bd2/pymc/sampling.py#L530-L531) we should maybe deprecate supporting a dedicated `nuts` kwarg and instead pass everything through the `kwargs`?
[...]

PR #6018, opened 2022-07-29
Fix problems with specifying target_accept and nuts kwargs

mschmidt87:
This PR addresses #6009.

## Major / Breaking Changes
 
None

## Bugfixes / New features

- Fixes the problem that target_accept is overwritten if nuts kwargs are specified, i.e., this case
    `pm.sample(nuts={"max_treedepth": 15}, target_accept=0.99)`
- Raises an error when target_accept is specified twice; directly and in nuts kwargs, i.e. this case
    `pm.sample(nuts={"target_accept": 0.9}, target_accept=0.99)`

To this end, I slightly modified the code in `pm.sample()` such that
- We first check whether `nuts` is defined in `kwargs` and if yes, update that dictionary, and if not, create it. Before, the `setdefault()` method was used which meant that `target_accept` was just dropped if `nuts` was already defined in `kwargs`.
-  it is explicitly checked whether `target_accept` is defined twice.

## Docs / Maintenance
 I added two new test cases to the `test_step_args()` test.

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -479,7 +479,7 @@ def sample(
                 "`target_accept` was defined twice. Please specify it either as a direct keyword argument or in the `nuts` kwarg."
             )
         if "nuts" in kwargs:
-            kwargs["nuts"].update({"target_accept": kwargs.pop("target_accept")})
+            kwargs["nuts"]["target_accept"] = kwargs.pop("target_accept")
         else:
             kwargs = {"nuts": {"target_accept": kwargs.pop("target_accept")}}
 
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -474,7 +474,14 @@ def sample(
         )
         initvals = kwargs.pop("start")
     if "target_accept" in kwargs:
-        kwargs.setdefault("nuts", {"target_accept": kwargs.pop("target_accept")})
+        if "nuts" in kwargs and "target_accept" in kwargs["nuts"]:
+            raise ValueError(
+                "`target_accept` was defined twice. Please specify it either as a direct keyword argument or in the `nuts` kwarg."
+            )
+        if "nuts" in kwargs:
+            kwargs["nuts"].update({"target_accept": kwargs.pop("target_accept")})
+        else:
+            kwargs = {"nuts": {"target_accept": kwargs.pop("target_accept")}}
 
     model = modelcontext(model)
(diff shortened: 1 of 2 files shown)


pymc/sampling.py
```diff
@@ -474,7 +474,14 @@ def sample(
         )
         initvals = kwargs.pop("start")
     if "target_accept" in kwargs:
-        kwargs.setdefault("nuts", {"target_accept": kwargs.pop("target_accept")})
+        if "nuts" in kwargs and "target_accept" in kwargs["nuts"]:
+            raise ValueError(
+                "`target_accept` was defined twice. Please specify it either as a direct keyword argument or in the `nuts` kwarg."
+            )
+        if "nuts" in kwargs:
+            kwargs["nuts"].update({"target_accept": kwargs.pop("target_accept")})
```
ricardoV94: Is this enough?
```suggestion
            kwargs["nuts"]["target_accept"] = kwargs.pop("target_accept")
```

pymc/tests/test_sampling.py
```diff
@@ -1450,9 +1450,14 @@ def test_step_args():
         a = pm.Normal("a")
         idata0 = pm.sample(target_accept=0.5, random_seed=1410)
         idata1 = pm.sample(nuts={"target_accept": 0.5}, random_seed=1410 * 2)
+        idata2 = pm.sample(target_accept=0.5, nuts={"max_treedepth": 10}, random_seed=1410)
+
+        with pytest.raises(ValueError):
```
ricardoV94: You can match the error string for a more precise test
```suggestion
        with pytest.raises(ValueError, match="..."):
```

ricardoV94 (commented):
Looks good. I left two small suggestions, nothing major

pymc/sampling.py
mschmidt87: Yes, I think this is easier, thank you.

pymc/tests/test_sampling.py
mschmidt87: Thank you, I added a pattern here to improve the test.

mschmidt87:
> Looks good. I left two small suggestions, nothing major

Thank you for the review, I added your suggested changes.

ricardoV94 (approved):

ricardoV94:
Thanks @mschmidt87 !!

Issue #6020, opened 2022-07-29
Open all NUTS kwargs to user for sampling with Numpyro 


jhrcook:
## Description of the problem

In the current implementation of `pymc.sampling_jax.sample_numpyro_nuts()`, only some of the arguments for the Numpyro NUTS sampler are available to the user because a few are preset in the function:

```python
if nuts_kwargs is None:
    nuts_kwargs = {}
nuts_kernel = NUTS(
    potential_fn=logp_fn,
    target_accept_prob=target_accept,
    adapt_step_size=True,
    adapt_mass_matrix=True,
    dense_mass=False,
    **nuts_kwargs,
)
```

I think it would be useful to make all of these arguments available to the user through the PyMC interface. Is there a reason they aren’t? If not, would this be a PR I could make?

For clarity, here is a simple example of what currently happens if a PyMC user wants to set `dense_mass=True`:

```python
with pm.Model():
    y = pm.Normal("y", 0, 1)
    trace = pymc.sampling_jax.sample_numpyro_nuts(nuts_kwargs={"dense_mass": True})
```

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [24], in <cell line: 1>()
      1 with pm.Model():
      2     y = pm.Normal("y", 0, 1)
----> 3     trace = pymc.sampling_jax.sample_numpyro_nuts(nuts_kwargs={"dense_mass": True})

File /usr/local/Caskroom/miniconda/base/envs/speclet/lib/python3.10/site-packages/pymc/sampling_jax.py:487, in sample_numpyro_nuts(draws, tune, chains, target_accept, random_seed, initvals, model, var_names, progress_bar, keep_untransformed, chain_method, postprocessing_backend, idata_kwargs, nuts_kwargs)
    485 if nuts_kwargs is None:
    486     nuts_kwargs = {}
--> 487 nuts_kernel = NUTS(
    488     potential_fn=logp_fn,
    489     target_accept_prob=target_accept,
    490     adapt_step_size=True,
    491     adapt_mass_matrix=True,
    492     # dense_mass=False,
    493     **nuts_kwargs,
    494 )
    496 pmap_numpyro = MCMC(
    497     nuts_kernel,
    498     num_warmup=tune,
   (...)
    503     progress_bar=progress_bar,
    504 )
    506 tic2 = datetime.now()

TypeError: numpyro.infer.hmc.NUTS() got multiple values for keyword argument 'dense_mass'
```

[...]

PR #6021, opened 2022-07-31
Pass user-provided NUTS kwargs to Numpyro

jhrcook:
**What is this PR about?**

There are currently defaults for a few keyword arguments for the Numpyro NUTS initialization method that prevent the PyMC user for passing certain keyword arguments. This PR refactors the code to update the user-provided keyword arguments dictionary with the defaults if they weren't included.

Addresses issue #6020 

**Checklist**

## Bugfixes / New features
- Enable the user to pass any keyword arguments to the Numpyro NUTS initialization.

## Docs / Maintenance
- Updated docstring of `sample_numpyro_nuts` to include information for the `initvals` argument
- fixed some incorrect typehints of arguments in the `sample_numpyro_nuts` definition
- added a return typehint to `sample_numpyro_nuts`

### DIFF
--- a/pymc/tests/test_sampling_jax.py
+++ b/pymc/tests/test_sampling_jax.py
@@ -2,13 +2,14 @@ from typing import Any, Dict
 
 import aesara
 import aesara.tensor as at
-import arviz as az
 import jax
 import numpy as np
+import numpyro
 import pytest
 
 from aesara.compile import SharedVariable
 from aesara.graph import graph_inputs
+from numpyro.infer import MCMC, NUTS
 
 import pymc as pm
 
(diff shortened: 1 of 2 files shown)


twiecki:
@jhrcook Can you add a test?

jhrcook:
@twiecki  I've added tests for the keyword-argument updating function and another set of tests that run a short MCMC with some different NUTS keyword arguments, but I have limited ability to actually determine if `numpyro.NUTS` actually *used* the arguments. The only one I could think of was to set the `step_size` and confirm that it was in the result `trace.sample_stats`. I could try monkey-patching into the NUTS initializer method, but it could be tricky because the `sample_numpyro_nuts` imports `NUTS` internally. Let me know if you'd like me to try, though.

pymc/tests/test_sampling_jax.py
```diff
@@ -270,3 +275,47 @@ def test_seeding(chains, random_seed, sampler):
...
+@pytest.mark.parametrize(
+    "nuts_kwargs",
+    [
+        {"adapt_step_size": False},
+        {"adapt_mass_matrix": True},
+        {"dense_mass": True},
+        {"adapt_step_size": False, "adapt_mass_matrix": True, "dense_mass": True},
+        {"adapt_step_size": False, "step_size": 0.13},
+    ],
+)
```
ricardoV94: Let's try to monkey patch? If it's not possible I think we can survive without explicit tests. This one seems too inneficient.

jhrcook: I'll give it shot. I agree that the current method I used is inefficient and isn't really testing the desired behavior.

jhrcook:
I used a monkeypatch to intercept the `NUTS` object in the initialization of the Numpyro `MCMC` object when called in `pymc.sampling_jax.sample_numpyro_jax()`. I just made a `MockMCMC` that checks some of the properties of its `NUTS` sampler and asserts they are what I passed to `sample_numpyro_jax(..., nuts_kwargs={...})`.

I confirmed that the monkeypatch was intercepting the `MCMC.__init__()` by purposefully making the tests fail with incorrect assertions.

I have limited experience with using monkeypatches in tests, so let me know if you think there would be a better construction/format.

ricardoV94:
When using monkey patching there's a helper method, something like `assert_called_once_with(...)` where you can directly test your monkey patched object was called once with specific arguments and kwargs.

jhrcook:
Thank you @ricardoV94 for the help. I've replaced the monkey patching system with a rather simple (too simple?) mocking system. I used mocking to allow me to peer into when the `numpyro.infer.MCMC` object is made and then inspect the NUTS sampler it was passed. This puts all of the assertions in a single function and removed the need to make a new mock class.

ricardoV94:
> Thank you @ricardoV94 for the help. I've replaced the monkey patching system with a rather simple (too simple?) mocking system. I used mocking to allow me to peer into when the `numpyro.infer.MCMC` object is made and then inspect the NUTS sampler it was passed. This puts all of the assertions in a single function and removed the need to make a new mock class.

Looks perfect. The simpler the better!

ricardoV94 (approved):

twiecki:
Thanks @jhrcook!

Issue #4687, opened 2021-05-12
Improve workflow around saving and loading fitted models


twiecki:
After talking to users who are trying to deploy models, one thing that commonly comes up is that it's cumbersome to do so. One reason is that it's hard to fit a PyMC3 model, save it, and then load it in later for doing predictions, like scikit-learn allows to do.

Fortunately, there exist two packages that already have implemented this on top of PyMC3: 
* https://github.com/parsing-science/pymc3_models by @parsing-science
* https://github.com/pymc-learn/pymc-learn by @Emaasit 
where `pymc-learn` looks like a fork of `pymc3_models` with some extensions and more models.

I think these already do most of what we need here, but both of the packages seem abandoned. There are a few options I can see and would like to get opinions on:
1. We could fork one of these packages and continue to support them (probably `pymc-learn` because it has more).
2. We could just take the base-class (https://github.com/pymc-learn/pymc-learn/blob/master/pmlearn/base.py#L18)  and add it to PyMC3.
3. Take just the base-class (https://github.com/pymc-learn/pymc-learn/blob/master/pmlearn/base.py#L18)  and add it to another dedicated package.

I'm leaning towards the second option because I don't think users really need a model zoo but just a nice API for their own models to load and save+predict. It's also easier to maintain. I also think this should be core PyMC3 functionality and it's not a lot of code (hence not option 3).

The only downside to 2 I see is that it would add `scikit-learn` as a dependency because we inherit from `sklearn.base.BaseEstimator` (https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/base.py#L142). Mostly this adds repr features which can be nice, but maybe they are also not really required. We could either not inherit from `BaseEstimator` and save the dependency, or make the dependency optional.

ricardoV94:
Just in case this sounds relevant: one feature I often wish was there was a method to automatically load saved traces if they are available but (re)sample and save the new traces if they are not, or if the model or data has changed since the last time it was sampled.

Emaasit:
@twiecki I had taken a long hiatus to focus on personal mental health. 

This is a timely git issue because I am back and currently working on upgrading pymc-learn to the latest version of PyMC3.

twiecki:
@Emaasit Always an important thing to prioritize, good to have you back.

[...]

Issue #64, opened 2011-10-27
Docstring format


sczesla:
Dear all,

first of all, thanks for this great package, which I enjoy much.

Do you use a fixed docstring format to generate API documentation? If
I just did not find it, please give me a hint. If not, do you have any intention
to use, e.g., the numpydoc format in the future?

Best regards,
Stefan

fonnesbeck:
We endeavor to use the numpy docstring format, but I'm sure there are places we have been inconsistent.

PR #6023, opened 2022-08-03
Added a modelBuilder class for deployment

5hv5hvnk:
**What is this PR about?**
This PR resolves  issue #4687, By creating a model builder class to perform `fit` , `predict` , `save`, `load` and few more things. Having direct API calls would be useful for users to deploy their standard model and allow people to use them directly using the methods mentioned above.
...

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- ...


ferrine:
In pymc you can already do this

```python
class LinearRegression(pm.Model):
    def __init__(self, data: pd.DataFrame, *, sigma=1, sigma_obs=1, observed=None, **kwargs):
        super().__init__(**kwargs)
        # use self.name_for to avoid name collisions for coords
        self.add_coord(self.name_for("coefs"), data.columns)
        alpha = pm.Normal("alpha", sigma=sigma, dims=self.name_for("coefs"))
        intercept = pm.HalfCauchy("intercept", 1)
        eps = pm.HalfNormal("sigma", sigma_obs)
        y = pm.Normal("y", data.values @ alpha + intercept, observed=observed)
```

and then use the template model like this

```python
data = pd.DataFrame(np.random.randn(100, 10))
lr = LinearRegression(data)
lr1 = LinearRegression(data, name="new")
```
image

The base Model class already supports a lot of automation for name creation and model building blocks. You can extend the idea to provide methods to the Model class itself, so that it is possible to do like this:

```
trace = lr.sample()
ppc = lr.sample_poseterior_predictive(trace)
```

an additional check should pass `lr1.isroot is True` to avoid cases like

```python
with pm.Model() as model:
    lr1 =  LinearRegression(data, name="first")
    y1 = lr1.y
    lr2 =  LinearRegression(data, name="second")
    lr1.sample()
>>> TypeError("Can't call model.sample on nested models")
```

ferrine:
This discussion is also related https://github.com/pymc-devs/pymc/discussions/6024

fonnesbeck:
The PyMC convention is to use underscores for file names rather than camelCase, so I would rename `modelBuilder.py` to `model_builder.py`.

twiecki:
@ferrine This inherits from `Model` but adds more functionality. And current `Model` still lacks a bunch of other functionality we're adding here. Or are you suggesting to just beef up `Model` directly?

ricardoV94:
> @ferrine This inherits from `Model` but adds more functionality. And current `Model` still lacks a bunch of other functionality we're adding here. Or are you suggesting to just beef up `Model` directly?

I think inheriting makes sense. 

Could we do this in pymc-experimental?

twiecki:
> Could we do this in pymc-experimental?

Yes, that's probably the right place for now.

5hv5hvnk:
I have generated a PR on PyMC experimental [#64](https://github.com/pymc-devs/pymc-experimental/pull/64)

PR #6035, opened 2022-08-06
Updated pymc.simulator docstring (typos, defaults, type description)

daniel-saunders-phil:
**What is this PR about?**

fixing a number of issues with the pymc.simulator docstring.

...

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- ...  added defaults to each parameter, changed instances of Aesara Op to correct format Aesara_Op, fixed a typo

### DIFF
--- a/pymc/distributions/simulator.py
+++ b/pymc/distributions/simulator.py
@@ -66,7 +66,7 @@ class SimulatorRV(RandomVariable):
 class Simulator(NoDistribution):
     r"""
     Simulator distribution, used for Approximate Bayesian Inference (ABC)
-    with Sequential Monte Carlo (SMC) sampling via ``pm.sample_smc``.
+    with Sequential Monte Carlo (SMC) sampling via ``~pymc.sample_smc``.
 
     Simulator distributions have a stochastic pseudo-loglikelihood defined by
     a distance metric between the observed and simulated data, and tweaked
@@ -78,11 +78,17 @@ class Simulator(NoDistribution):
         Python random simulator function. Should expect the following signature
         ``(rng, arg1, arg2, ... argn, size)``, where rng is a ``numpy.random.RandomStream()``
         and ``size`` defines the size of the desired sample.
-    params : list
-        Parameters used by the Simulator random function. Parameters can also
-        be passed by order, in which case the keyword argument ``params`` is
-        ignored. Alternatively, each parameter can be passed by order after fn,
-        ``param1, param2, ..., paramN``
+
+    *unnamed_params : list of TensorVariable
+        Parameters used by the Simulator random function. Each parameter can be passed 
+        by order after fn, for example ``param1, param2, ..., paramN``. params can also 
+        be passed with keyword argument "params".
+    
+    params : list of TenssorVariable
+        Keyword form of ''unnamed_params''. 
+        One of unnamed_params or params must be provided. 
+        If passed both unnamed_params and params, an error is raised. 
+        
     distance : Aesara_Op, callable or str, default "gaussian"
         Distance function. Available options are ``"gaussian"``, ``"laplace"``,
         ``"kullback_leibler"`` or a user defined function (or Aesara_Op) that takes
@@ -102,14 +108,15 @@ class Simulator(NoDistribution):
         Summary statistic function. Available options are ``"identity"``,
         ``"sort"``, ``"mean"``, ``"median"``. If a callable (or Aesara_Op) is defined,
         it should return a 1d numpy array (or Aesara vector).
-    epsilon : float or array, default 1.0
+    epsilon : tensor_like of float, default 1.0
         Scaling parameter for the distance functions. It should be a float or
         an array of the same size of the output of ``sum_stat``.
     ndim_supp : int, default 0
         Number of dimensions of the SimulatorRV (0 for scalar, 1 for vector, etc.)
-    ndims_params : list[int], default 0 for each parameter.
+    ndims_params : list of int, optional 
         Number of minimum dimensions of each parameter of the RV. For example,
-        if the Simulator accepts two scalar inputs, it should be ``[0, 0]``.
+        if the Simulator accepts two scalar inputs, it should be ``[0, 0]``. 
+        Default to list of 0 with length equal to the number of parameters.
 
     Examples
     --------
@@ -156,7 +163,7 @@ class Simulator(NoDistribution):
             elif distance == "kullback_leibler":
                 raise NotImplementedError("KL not refactored yet")
                 # TODO: Wrap KL in aesara OP
-                # distance = KullbackLiebler(observed)
+                # distance = KullbackLeibler(observed)
                 # if sum_stat != "identity":
                 #     _log.info(f"Automatically setting sum_stat to identity as expected by {distance}")
                 #     sum_stat = "identity"
@@ -271,8 +278,8 @@ def laplace(epsilon, obs_data, sim_data):
     return -at.abs_((obs_data - sim_data) / epsilon)
 
 
-class KullbackLiebler:
-    """Approximate Kullback-Liebler."""
+class KullbackLeibler:
+    """Approximate Kullback-Leibler."""
 
     def __init__(self, obs_data):
         if obs_data.ndim == 1:
--- a/pymc/distributions/simulator.py
+++ b/pymc/distributions/simulator.py
@@ -83,9 +83,9 @@ class Simulator(NoDistribution):
         be passed by order, in which case the keyword argument ``params`` is
         ignored. Alternatively, each parameter can be passed by order after fn,
         ``param1, param2, ..., paramN``
-    distance : Aesara Op, callable or str
-        Distance function. Available options are ``"gaussian"`` (default), ``"laplace"``,
-        ``"kullback_leibler"`` or a user defined function (or Aesara Op) that takes
+    distance : Aesara_Op, callable or str, default "gaussian"
+        Distance function. Available options are ``"gaussian"``, ``"laplace"``,
+        ``"kullback_leibler"`` or a user defined function (or Aesara_Op) that takes
         ``epsilon``, the summary statistics of observed_data and the summary statistics
         of simulated_data as input.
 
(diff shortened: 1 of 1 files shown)


pymc/distributions/simulator.py
```diff
@@ -98,20 +98,18 @@ class Simulator(NoDistribution):
...
         ``distance="laplace"`` + ``sum_stat="sort"`` is equivalent to the the 1D 1-wasserstein distance
-    sum_stat : Aesara Op, callable or str
-        Summary statistic function. Available options are ``"indentity"`` (default),
-        ``"sort"``, ``"mean"``, ``"median"``. If a callable (or Aesara Op) is defined,
+    sum_stat : Aesara_Op, callable or str, default "identity"
+        Summary statistic function. Available options are ``"identity"``,
+        ``"sort"``, ``"mean"``, ``"median"``. If a callable (or Aesara_Op) is defined,
         it should return a 1d numpy array (or Aesara vector).
-    epsilon : float or array
+    epsilon : float or array, default 1.0
```
OriolAbril: I think this should be `tensor_like of float`

pymc/distributions/simulator.py
```diff
@@ -98,20 +98,18 @@ class Simulator(NoDistribution):
...
         Scaling parameter for the distance functions. It should be a float or
-        an array of the same size of the output of ``sum_stat``. Defaults to ``1.0``
-    ndim_supp : int
+        an array of the same size of the output of ``sum_stat``.
+    ndim_supp : int, default 0
         Number of dimensions of the SimulatorRV (0 for scalar, 1 for vector, etc.)
-        Defaults to ``0``.
-    ndims_params : list[int]
+    ndims_params : list[int], default 0 for each parameter.
```
OriolAbril:
```suggestion
    ndims_params : list of int, optional
```

As the default is a bit complex here, it is not a value we can add here but something that depends on other parameters, I would use optional and keep the explanation in the description

pymc/distributions/simulator.py
```diff
@@ -83,9 +83,9 @@ class Simulator(NoDistribution):
         be passed by order, in which case the keyword argument ``params`` is
         ignored. Alternatively, each parameter can be passed by order after fn,
         ``param1, param2, ..., paramN``
```
OriolAbril: Can't comment further up. It looks like `unnamed_params` is missing. I think it should be something like

```
*unnamed_params : list of TensorVariable
    description starts here
params : list of TensorVariable
    description extended here, plus mention that if both are present an error is thrown.
```

someone else will need to take over on the actual description though as I don't really know what to write. Is one preferred over the other, are there any potential behaviour differences? 

If they are both exactly the same, then the we can follow for example xarray's pattern where this situation is very common with kwargs, see for example: https://docs.xarray.dev/en/stable/generated/xarray.Dataset.rename.html

OriolAbril (commented):
Thanks!

It already looks great, and the docs look better and are more clear already. I left some more comments, plus this one here because it is on a line I wasn't able to comment on: the short description also has

```
``pm.sample_smc``
# should instead be
:func:`~pymc.sample_smc`
```

daniel-saunders-phil:
Hi Oriol, thanks for the feedback - these suggestions are really helpful. I'm a bit unsure about the procedure. To revise a pull request, do I make the changes on my branch and then resubmit the whole PR through the git command line instructions or is there a different procedure?

OriolAbril:
Make the changes locally and commit them to the branch you opened your PR from, `docstring_update` in this case. After that you only need to push the changes and this same PR will be updated automatically with the new commits.

This is the last section in https://pymc-data-umbrella.xyz/en/latest/sprint/tutorials/pr_tutorial.html#get-ready-for-reviews which has the specific commands and extends the overview I gave in the above paragraph.

pymc/distributions/simulator.py
daniel-saunders-phil: It looks like they should behave identically. If the keyword argument isn't passed, then params just gets defined as the unnamed_params and then params is used throughout the rest of the file. I went ahead and wrote a description similar to xarray.

daniel-saunders-phil:
I implemented the suggested changes and then pushed them. 

Also, one thing came up that wasn't strictly a docstring fix. I noticed the file uses inconsistent spellings of Kullback-Leiber and the class is incorrectly called KullbackL**ie**bler. I thought we might want to change all of them to the correct spelling. I searched through the rest of the pymc package and couldn't find any references to the incorrectly spelled class, so I don't think this should cause trouble for other functions. But I'm also happy to ignore this and just stick to docstrings for now.

reshamas:
@daniel-saunders-phil 
I think it's ok to fix the spelling of Kullback-Leiber.
I've been fixing spelling in the notebooks I have been updating, and it's been ok.

daniel-saunders-phil:
I’m not sure how to handle this pre-commit problem. I’m gonna try to figure it out for a little bit but I might have some questions in a couple of days.

reshamas:
@daniel-saunders-phil I am working on my 3rd notebook and, once again(!), stuck at the pre-commit stage. (+ possible git issue)

@OriolAbril 
Perhaps we should have a special office hours session on how to get through pre-commit.  ;)

cc:  @symeneses

OriolAbril:
@daniel-saunders-phil if you click on the "Details" button of the checks seciton you'll be able to see the checks logs. In this PR, the issue is with the "trim trailing whitespace": https://github.com/pymc-devs/pymc/runs/7715083643?check_suite_focus=true#step:4:64.

That means that there are lines that end up with a white character instead of finishing with the last letter/punctiation sign of the line. If you have followed the steps from https://pymc-data-umbrella.xyz/en/latest/sprint/tutorials/environment_setup.html to set up your environment, you can run `pre-commit run --all` on your command line (while inside the pymc-dev environment). pre-commit will fail, but it will also modify the files and remove those trailing whitespaces. You can then git add and git commit those changes and push, which should fix the CI here.

daniel-saunders-phil:
It looks like all tests are passing this time! after a little bit of tinkering, I managed to get pre-commit working. Thanks a lot, @OriolAbril and @reshamas !

pymc/distributions/simulator.py
```diff
@@ -66,7 +66,7 @@ def sum_stat(cls, *args, **kwargs):
 class Simulator(NoDistribution):
     r"""
     Simulator distribution, used for Approximate Bayesian Inference (ABC)
-    with Sequential Monte Carlo (SMC) sampling via ``pm.sample_smc``.
+    with Sequential Monte Carlo (SMC) sampling via ``~pymc.sample_smc``.
```
OriolAbril:
```suggestion
    with Sequential Monte Carlo (SMC) sampling via :func:`~pymc.sample_smc`.
```

it needs to use this syntax in order to be rendered as a link, not only use the correct import path of the function. Ref: https://pymc-data-umbrella.xyz/en/latest/about/contributing_to_documentation/docs_presentation.html#cross-referencing-roles

pymc/distributions/simulator.py
```diff
@@ -78,14 +78,20 @@ class Simulator(NoDistribution):
...
-    distance : Aesara Op, callable or str
-        Distance function. Available options are ``"gaussian"`` (default), ``"laplace"``,
-        ``"kullback_leibler"`` or a user defined function (or Aesara Op) that takes
+
+    *unnamed_params : list of TensorVariable
+        Parameters used by the Simulator random function. Each parameter can be passed
+        by order after fn, for example ``param1, param2, ..., paramN``. params can also
+        be passed with keyword argument "params".
+
+    params : list of TenssorVariable
```
OriolAbril:
```suggestion
    params : list of TensorVariable
```

pymc/distributions/simulator.py
```diff
@@ -78,14 +78,20 @@ class Simulator(NoDistribution):
...
         and ``size`` defines the size of the desired sample.
-    params : list
-        Parameters used by the Simulator random function. Parameters can also
-        be passed by order, in which case the keyword argument ``params`` is
-        ignored. Alternatively, each parameter can be passed by order after fn,
-        ``param1, param2, ..., paramN``
-    distance : Aesara Op, callable or str
-        Distance function. Available options are ``"gaussian"`` (default), ``"laplace"``,
-        ``"kullback_leibler"`` or a user defined function (or Aesara Op) that takes
+
```
OriolAbril:
```suggestion
```

pymc/distributions/simulator.py
```diff
@@ -78,14 +78,20 @@ class Simulator(NoDistribution):
...
+    *unnamed_params : list of TensorVariable
+        Parameters used by the Simulator random function. Each parameter can be passed
+        by order after fn, for example ``param1, param2, ..., paramN``. params can also
+        be passed with keyword argument "params".
+
```
OriolAbril:
```suggestion
```

pymc/distributions/simulator.py
```diff
@@ -78,14 +78,20 @@ class Simulator(NoDistribution):
...
+    params : list of TenssorVariable
+        Keyword form of ''unnamed_params''.
+        One of unnamed_params or params must be provided.
+        If passed both unnamed_params and params, an error is raised.
+
```
OriolAbril:
```suggestion
```

OriolAbril (approved):

michaelosthege (approved):

michaelosthege (changes requested):
Sorry for requesting changes this late, but I think we should keep an alias of the `KullbackLeibler` class:

```python
class KullbackLiebler(KullbackLeibler):
    def __init__(self,  *args, **kwargs):
        """Alias for KullbackLeibler."""
        warnings.warn("You are using `KullbackLiebler` which is a typo. Use `KullbackLeibler` instead.", DeprecationWarning)
        super(). __init__(*args, **kwargs)

OriolAbril:
I considered it too but but eventually decided on not doing it because 1) it is not included in our docs and therefore not part of the public api and 2) no tests broke because it has not been refactored to v4,  so it would be a breaking change on the private api with a major change version, from 3.11 to 4.x (whatever the version it gets updated).

We can still add the alias to be nice,  but then it should have a different type of warning. This is the description of deprecationwarning in https://docs.python.org/3/library/exceptions.html#DeprecationWarning :

> Base class for warnings about deprecated features when those warnings are intended for other Python developers.
>
> Ignored by the default warning filters, except in the __main__ module (PEP 565). Enabling the Python Development Mode shows this warning.

michaelosthege (approved):
ok, thanks Oriol! I didn't realize this wasn't public API.

reshamas:
Congrats @daniel-saunders-phil !

#DataUmbrellaPyMCSprint

PR #6042, opened 2022-08-09
Allow for batched `alpha` in `StickBreakingWeights`

purna135:
**What is this PR about?**
Addressing #5383
This enables `StickBreakingWeight`'s `alpha` to accept batched data (>2D), make the `infer_shape` work with batched data, and fix the `rng_fn` by broadcasting alpha to K.

**Checklist**

## Major / Breaking Changes
* ...

## Bugfixes / New features
- `StickBreakingWeights` now supports batched `alpha` parameters

## Docs / Maintenance
* ...

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2277,9 +2277,10 @@ class StickBreakingWeights(SimplexContinuous):
         return super().dist([alpha, K], **kwargs)
 
     def moment(rv, size, alpha, K):
+        alpha = alpha[..., np.newaxis]
         moment = (alpha / (1 + alpha)) ** at.arange(K)
         moment *= 1 / (1 + alpha)
-        moment = at.concatenate([moment, [(alpha / (1 + alpha)) ** K]], axis=-1)
+        moment = at.concatenate([moment, (alpha / (1 + alpha)) ** K], axis=-1)
         if not rv_size_is_none(size):
             moment_size = at.concatenate(
                 [
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2206,10 +2206,12 @@ class StickBreakingWeightsRV(RandomVariable):
         if K < 0:
             raise ValueError("K needs to be positive.")
 
-        size = to_tuple(size)
-        size = np.broadcast_shapes(alpha.shape, size) + (K,)
+        if size is None:
+            size = alpha.shape + (K,)
+            alpha = alpha[..., np.newaxis]
+        else:
+            size = size + (K,)
 
-        alpha = alpha[..., np.newaxis]
         betas = rng.beta(1, alpha, size=size)
 
         sticks = np.concatenate(
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2206,11 +2206,9 @@ class StickBreakingWeightsRV(RandomVariable):
         if K < 0:
             raise ValueError("K needs to be positive.")
 
-        if size is None:
-            size = alpha.shape + (K,)
-            alpha = alpha[..., np.newaxis]
-        else:
-            size = size + (K,)
+        size = size if size is not None else alpha.shape
+        size = size + (K,)
+        alpha = alpha[..., np.newaxis]
 
         betas = rng.beta(1, alpha, size=size)
 
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2192,9 +2192,6 @@ class StickBreakingWeightsRV(RandomVariable):
         alpha = at.as_tensor_variable(alpha)
         K = at.as_tensor_variable(intX(K))
 
-        if alpha.ndim > 0:
-            raise ValueError("The concentration parameter needs to be a scalar.")
-
         if K.ndim > 0:
             raise ValueError("K must be a scalar.")
 
@@ -2205,20 +2202,17 @@ class StickBreakingWeightsRV(RandomVariable):
 
         size = tuple(size)
 
-        return size + (K + 1,)
+        return size + tuple(alpha.shape) + (K + 1,)
 
     @classmethod
     def rng_fn(cls, rng, alpha, K, size):
         if K < 0:
             raise ValueError("K needs to be positive.")
 
-        if size is None:
-            size = (K,)
-        elif isinstance(size, int):
-            size = (size,) + (K,)
-        else:
-            size = tuple(size) + (K,)
+        distribution_shape = alpha.shape + (K,)
+        size = to_tuple(size) + distribution_shape
 
+        alpha = alpha[..., np.newaxis]
         betas = rng.beta(1, alpha, size=size)
 
         sticks = np.concatenate(
@@ -2262,7 +2256,7 @@ class StickBreakingWeights(SimplexContinuous):
 
     Parameters
     ----------
-    alpha : tensor_like of float
+    alpha: float or array_like of floats
         Concentration parameter (alpha > 0).
     K : tensor_like of int
         The number of "sticks" to break off from an initial one-unit stick. The length of the weight
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2197,20 +2197,17 @@ class StickBreakingWeightsRV(RandomVariable):
 
         return super().make_node(rng, size, dtype, alpha, K)
 
-    def _infer_shape(self, size, dist_params, param_shapes=None):
-        alpha, K = dist_params
-
-        size = tuple(size)
-
-        return size + tuple(alpha.shape) + (K + 1,)
+    def _supp_shape_from_params(self, dist_params, **kwargs):
+        K = dist_params[1]
+        return (K + 1,)
 
     @classmethod
     def rng_fn(cls, rng, alpha, K, size):
         if K < 0:
             raise ValueError("K needs to be positive.")
 
-        distribution_shape = alpha.shape + (K,)
-        size = to_tuple(size) + distribution_shape
+        size = to_tuple(size)
+        size = np.broadcast_shapes(alpha.shape, size) + (K,)
 
         alpha = alpha[..., np.newaxis]
         betas = rng.beta(1, alpha, size=size)
@@ -2256,7 +2253,7 @@ class StickBreakingWeights(SimplexContinuous):
 
     Parameters
     ----------
-    alpha: float or array_like of floats
+    alpha : tensor_like of float
         Concentration parameter (alpha > 0).
     K : tensor_like of int
         The number of "sticks" to break off from an initial one-unit stick. The length of the weight
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2206,7 +2206,7 @@ class StickBreakingWeightsRV(RandomVariable):
         if K < 0:
             raise ValueError("K needs to be positive.")
 
-        size = size if size is not None else alpha.shape
+        size = to_tuple(size) if size is not None else alpha.shape
         size = size + (K,)
         alpha = alpha[..., np.newaxis]
 
--- a/pymc/tests/test_distributions.py
+++ b/pymc/tests/test_distributions.py
@@ -24,6 +24,7 @@ from aeppl.logprob import ParameterValueError
 from aesara.tensor.random.utils import broadcast_params
 
 from pymc.distributions.continuous import get_tau_sigma
+from pymc.distributions.dist_math import betaln
 from pymc.util import UNSET
 
 try:
@@ -950,6 +951,38 @@ def test_hierarchical_obs_logp():
     ops = {a.owner.op for a in logp_ancestors if a.owner}
     assert len(ops) > 0
     assert not any(isinstance(o, RandomVariable) for o in ops)
+
+
+def _stickbreakingweights_logpdf(value, alpha, K):
+    logp = -at.sum(
+        at.log(
+            at.cumsum(
+                value[..., ::-1],
+                axis=-1,
+            )
+        ),
+        axis=-1,
+    )
+    logp += -K * betaln(1, alpha)
+    logp += alpha * at.log(value[..., -1])
+    logp = at.switch(
+        at.or_(
+            at.any(
+                at.and_(at.le(value, 0), at.ge(value, 1)),
+                axis=-1,
+            ),
+            at.or_(
+                at.bitwise_not(at.allclose(value.sum(-1), 1)),
+                at.neq(value.shape[-1], K + 1),
+            ),
+        ),
+        -np.inf,
+        logp,
+    )
+    return logp.eval()
+
+
+stickbreakingweights_logpdf = np.vectorize(_stickbreakingweights_logpdf, signature="(n),(),()->()")
 
 
 class TestMatchesScipy:
@@ -2291,24 +2324,6 @@ class TestMatchesScipy:
                 3,
                 np.array([1.29317672, 1.50126157]),
             ),
-            (
-                np.array([5, 4, 3, 2, 1]) / 15,
-                np.array([0.5, 1, 2], dtype="float64"),
-                4,
-                np.array([1.51263013, 2.93119375, 2.99573227]),
-            ),
-            (
-                np.array([5, 4, 3, 2, 1]) / 15,
-                np.arange(1, 10, dtype="float64").reshape(3, 3),
-                4,
-                np.array(
-                    [
-                        [2.93119375, 2.99573227, 1.9095425],
-                        [0.35222059, -1.4632554, -3.44201938],
-                        [-5.53346686, -7.70739149, -9.94430955],
-                    ]
-                ),
-            ),
         ],
     )
     def test_stickbreakingweights_logp(self, value, alpha, K, logp):
@@ -2329,6 +2344,28 @@ class TestMatchesScipy:
         assert pm.logp(sbw, np.array([1.1, 0.3, 0.2, 0.1])).eval() == -np.inf
         assert pm.logp(sbw, np.array([0.4, 0.3, 0.2, -0.1])).eval() == -np.inf
         assert pm.logp(sbw_wrong_K, np.array([0.4, 0.3, 0.2, 0.1])).eval() == -np.inf
+
+    @pytest.mark.parametrize(
+        "value, alpha, K",
+        [
+            (np.array([5, 4, 3, 2, 1]) / 15, [0.5, 1.0, 2.0], 19),
+            (
+                np.append(0.5 ** np.arange(1, 20), 0.5**20),
+                np.arange(1, 7, dtype="float64").reshape(2, 3),
+                4,
+            ),
+        ],
+    )
+    def test_stickbreakingweights_vectorized(self, value, alpha, K):
+        with Model():
+            sbw = StickBreakingWeights("sbw", alpha=alpha, K=K, transform=None)
+        pt = {"sbw": value}
+        assert_almost_equal(
+            pm.logp(sbw, value).eval(),
+            stickbreakingweights_logpdf(value, alpha, K),
+            decimal=select_by_precision(float64=6, float32=2),
+            err_msg=str(pt),
+        )
 
     @aesara.config.change_flags(compute_test_value="raise")
     def test_categorical_bounds(self):
--- a/pymc/tests/test_distributions.py
+++ b/pymc/tests/test_distributions.py
@@ -962,6 +962,15 @@ def _compile_stickbreakingweights_logpdf():
     return compile_pymc([_value, _alpha, _k], _logp)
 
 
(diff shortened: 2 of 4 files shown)


pymc/tests/test_distributions_random.py
```diff
@@ -1287,25 +1287,26 @@ class TestDirichletMultinomial_1D_n_2D_a(BaseTestDistributionRandom):
...
-    checks_to_run = [
-        "check_pymc_params_match_rv_op",
-        "check_rv_size",
-        "check_basic_properties",
+    parameters = [
+        (np.array(3.5), 19),
+        (np.array([1, 2, 3], dtype="float64"), 17),
+        (np.arange(1, 10, dtype="float64").reshape(3, 3), 15),
+        (np.arange(1, 25, dtype="float64").reshape(2, 3, 4), 5),
     ]
```
purna135: I need some assistance in writing a better test for testing multiple alphas. Can we use `@pytest.mark.parametrize` here?

pymc/tests/test_distributions_random.py
```diff
@@ -1287,25 +1287,26 @@ class TestDirichletMultinomial_1D_n_2D_a(BaseTestDistributionRandom):
...
+        for size in sizes_to_check:
+            sizes_expected.append(to_tuple(size) + alpha.shape + (K + 1,))
+
+        checks_to_run = [
+            "check_pymc_params_match_rv_op",
+            "check_rv_size",
+            "check_basic_properties",
+        ]
 
     def check_basic_properties(self):
```
purna135: Is there any wrong with using `@pytest.mark.parametrize` for `check_basic_properties` method?
I tried the below code but got the error: `check_basic_properties() missing 2 required positional arguments: 'alpha' and 'K'`

```python
@pytest.mark.parametrize(
  "alpha, K",
  [
      (np.array(1, dtype="float64"), 3),
      (np.array([1, 2, 3], dtype="float64"), 3),
      (np.arange(1, 10, dtype="float64").reshape(3, 3), 2),
      (np.arange(1, 25, dtype="float64").reshape(2, 3, 4), 2),
  ],
)
def check_basic_properties(self, alpha, K):
  default_rng = aesara.shared(np.random.default_rng(1234))
  draws = pm.StickBreakingWeights.dist(
      alpha=alpha,
      K=K,
      size=(2, 3, 5),
      rng=default_rng,
  ).eval()
```

pymc/distributions/multivariate.py
```diff
@@ -2205,20 +2202,17 @@ def _infer_shape(self, size, dist_params, param_shapes=None):
 
         size = tuple(size)
 
-        return size + (K + 1,)
+        return size + tuple(alpha.shape) + (K + 1,)
```
ricardoV94: This is incorrect. Shape is always `size + (K + 1,), if size is provided. If size is not provided, the shape of alpha determines it. 

However a simpler solution is to remove this method and implement only `supp_shape_from_params` which should always return `(k + 1)` Aesara will then automatically know how the dimensions change with batched alpha or size info

pymc/distributions/multivariate.py
```diff
@@ -2205,20 +2202,17 @@ def _infer_shape(self, size, dist_params, param_shapes=None):
...
             raise ValueError("K needs to be positive.")
 
-        if size is None:
-            size = (K,)
-        elif isinstance(size, int):
-            size = (size,) + (K,)
-        else:
-            size = tuple(size) + (K,)
+        distribution_shape = alpha.shape + (K,)
+        size = to_tuple(size) + distribution_shape
```
ricardoV94: Again, size should not simply be prepended to the shape of alpha. Size (when specified) must include (but can go beyond) any batched dimensions implied by alpha. This guide might make this point more clear: https://www.pymc.io/projects/docs/en/stable/learn/core_notebooks/dimensionality.html

pymc/distributions/multivariate.py
```diff
@@ -2262,7 +2256,7 @@ class StickBreakingWeights(SimplexContinuous):
 
     Parameters
     ----------
-    alpha : tensor_like of float
+    alpha: float or array_like of floats
```
ricardoV94: tensor_like is the correct docstring

ricardoV94 (changes requested):
Good start. I found some issues about the interpretation of size in conjugation with batched alphas

pymc/tests/test_distributions.py
```diff
@@ -2291,6 +2291,24 @@ def test_dirichlet_multinomial_vectorized(self, n, a, extra_size):
                 3,
                 np.array([1.29317672, 1.50126157]),
             ),
+            (
+                np.array([5, 4, 3, 2, 1]) / 15,
```
ricardoV94: These tests don't have a right shape for value. More importantly I don't think the hard coded logp values is very maintainable. Since this PR is all about allowing for batching of alpha I suggest you test it with a np.vectorize function of the base logp, and make it a separate test. You can find some examples in the tests for batched Dirichlet and Multinomial.

ricardoV94 (changes requested):

pymc/distributions/multivariate.py
```diff
@@ -2192,33 +2192,24 @@ def make_node(self, rng, size, dtype, alpha, K):
...
             raise ValueError("K needs to be positive.")
 
-        if size is None:
-            size = (K,)
-        elif isinstance(size, int):
-            size = (size,) + (K,)
-        else:
-            size = tuple(size) + (K,)
+        size = to_tuple(size)
+        size = np.broadcast_shapes(alpha.shape, size) + (K,)
```
ricardoV94: This is still a bit incorrect, alpha is not allowed to broadcast size. Simply put `size = size if size is not None else alpha.shape`

ricardoV94 (changes requested):

pymc/distributions/multivariate.py
purna135: Ok, Thank you

Sayam753: > alpha is not allowed to broadcast size

Hi @ricardoV94 

Considering the below snippet and assuming MvNormal works with batched covariance matrix, what is shape of resulting draws from MvNormal:

```python
mu = np.array([1, 2, 3])
cov = np.eye(3)
batched_cov = np.full((10, 3, 3), cov)
pm.MvNormal.dist(mu=mu, cov=cov, size=(5, 1)).eval().shape
```

Will it be (5, 1, 10, 3) or (5, 10, 3)?

I think the shape should be (5, 10, 3) as batches from alpha can broadcast with size.

ricardoV94: That's invalid. Size determines the number of samples. It can either be implied by the batches dimensions of the parameters, match them or go beyond.

It should behave like `np.full(value, shape)` (I don't remember the order of the parameters). You can't ask for `np.full(np.zeros((10, 3, 3), size=(5, 1))`

ricardoV94: Deleted by mistake. The example is invalid, it should fail. Size can be either `None`, or some tuple that ends in `(10,)`. It can't contradict the batched shape already implied by the covariance. 

Similar to `np.full`, where the shape can't broadcast to match the value, only the other way around. `np.full((5, 1), np.zeros((10,)))` also fails.

Sayam753: Got it. Thanks Ricardo. 

I think this behaviour (that size can't contradict the batched shape already implied) is a lot easier to work with, compared to PyMC v3.

pymc/distributions/multivariate.py
```diff
@@ -2192,32 +2192,25 @@ def make_node(self, rng, size, dtype, alpha, K):
...
     def rng_fn(cls, rng, alpha, K, size):
         if K < 0:
             raise ValueError("K needs to be positive.")
 
         if size is None:
-            size = (K,)
-        elif isinstance(size, int):
-            size = (size,) + (K,)
+            size = alpha.shape + (K,)
+            alpha = alpha[..., np.newaxis]
```
ricardoV94: Don't you need to expand alpha regardless of whether size is provided or not?

pymc/tests/test_distributions.py
purna135: How can I determine the right shape for value?

pymc/tests/test_distributions_random.py
```diff
@@ -1321,6 +1321,18 @@ def check_basic_properties(self):
         assert np.all(draws <= 1)
 
 
+class TestStickBreakingWeights_1D_alpha(BaseTestDistributionRandom):
+    pymc_dist = pm.StickBreakingWeights
+    pymc_dist_params = {"alpha": [1.0, 2.0, 3.0], "K": 19}
+    expected_rv_op_params = {"alpha": [1.0, 2.0, 3.0], "K": 19}
+    sizes_to_check = [None]
```
ricardoV94: What about a couple more?
```suggestion
    sizes_to_check = [None, (3,), (5, 3)]
```

pymc/tests/test_distributions.py
```diff
@@ -952,6 +953,38 @@ def test_hierarchical_obs_logp():
     assert not any(isinstance(o, RandomVariable) for o in ops)
 
 
+def _stickbreakingweights_logpdf(value, alpha, K):
```
purna135: Could you please provide a reference for a better way to calculate the logp here?

ricardoV94: You can replace this with:

```python
_value = at.vector()
_alpha = at.scalar()
_k = at.iscalar()
_logp = logp(StickBreakingWeights.dist(_alpha, _k), _value)
_stickbreakingweights_logpdf = compile_pymc([_value, _alpha, _k], _logp)
```

ricardoV94 (changes requested):

pymc/tests/test_distributions.py
purna135: hmm got it thank you

pymc/tests/test_distributions_random.py
purna135: Sure

pymc/tests/test_distributions.py
ricardoV94: You can take a random draw from the distribution. Value should have the same shape as the distribution, that is `(*alpha.shape, k+1)` or `(*size, k+1)`

Since you are already checking that the distribution has the right draw shapes, you can simply create a test value for the logp by taking a random draw from the distribution, instead of creating one manually. Thay why is enough to parametrize your logp tests by providing different alphas and sizes.

pymc/distributions/multivariate.py
purna135: Is this the case when alpha.shape == size[-1]? yes in that case we have to expand alpha.

ricardoV94: I think you need to expand always because you are feeding it into beta with (k+1) draws to the right and you want alpha to broadcast to those.

purna135: yes got it

pymc/tests/test_distributions.py
```diff
@@ -2312,6 +2313,37 @@ def test_stickbreakingweights_invalid(self):
...
+        [
+            (np.array([5, 4, 3, 2, 1]) / 15, [0.5, 1.0, 2.0], 4),
+            (
+                np.append(0.5 ** np.arange(1, 20), 0.5**20),
+                np.arange(1, 7, dtype="float64").reshape(2, 3),
+                19,
+            ),
+        ],
+    )
+    def test_stickbreakingweights_vectorized(self, value, alpha, K):
```
ricardoV94: You should define this logpdf function outside of the test_function, no point in compiling it everytime.

The best is probably inside a module-level pytest fixture so that it does not slowdown unrelated tests.

ricardoV94: value still has a wrong shape. Did you see my other comment about how you could fix it?

ricardoV94 (changes requested):

pymc/distributions/multivariate.py
```diff
@@ -2192,32 +2192,23 @@ def make_node(self, rng, size, dtype, alpha, K):
...
             raise ValueError("K needs to be positive.")
 
-        if size is None:
-            size = (K,)
-        elif isinstance(size, int):
-            size = (size,) + (K,)
-        else:
-            size = tuple(size) + (K,)
+        size = size if size is not None else alpha.shape
+        size = size + (K,)
```
larryshamalama, line 2210: @ricardoV94 Do we require `size` to be a tuple? If not, would we need `size = pm.distributions.shape_utils.to_tuple(size)`?

pymc/distributions/multivariate.py
```diff
@@ -2192,32 +2192,23 @@ def make_node(self, rng, size, dtype, alpha, K):
...
         if K < 0:
             raise ValueError("K needs to be positive.")
 
-        if size is None:
-            size = (K,)
-        elif isinstance(size, int):
-            size = (size,) + (K,)
-        else:
-            size = tuple(size) + (K,)
+        size = size if size is not None else alpha.shape
```
larryshamalama: In cases of `alpha = 3`, we probably need to wrap `alpha = np.array(array)`.

larryshamalama (commented):
Great work @purna135! :)

larryshamalama:
Would the [`moment`](https://github.com/pymc-devs/pymc/blob/d933c070352deb3b1a3261e59cf07de9a511f93c/pymc/distributions/multivariate.py#L2292) and [`logp`](https://github.com/pymc-devs/pymc/blob/d933c070352deb3b1a3261e59cf07de9a511f93c/pymc/distributions/multivariate.py#L2320) methods in the distribution class need appropriate broadcasting for vector-valued `alpha`s? I am just reading over this PR and might have missed previous discussions

pymc/tests/test_distributions.py
purna135: Yes I got the comment, will fix this in the next commit

purna135: > The best is probably inside a module-level pytest fixture so that it does not slowdown unrelated tests.

do you mean a function inside pymc\tests\test_distributions.py ?

ricardoV94: > > The best is probably inside a module-level pytest fixture so that it does not slowdown unrelated tests.
> 
> do you mean a function inside pymc\tests\test_distributions.py ?

Yes

pymc/distributions/multivariate.py
ricardoV94: `tuple(size)` like before should be fine

pymc/distributions/multivariate.py
ricardoV94: alpha will be converted to an array by Aesara automatically, so that should be safe

ricardoV94:
> Would the [`moment`](https://github.com/pymc-devs/pymc/blob/d933c070352deb3b1a3261e59cf07de9a511f93c/pymc/distributions/multivariate.py#L2292) and [`logp`](https://github.com/pymc-devs/pymc/blob/d933c070352deb3b1a3261e59cf07de9a511f93c/pymc/distributions/multivariate.py#L2320) methods in the distribution class need appropriate broadcasting for vector-valued `alpha`s? I am just reading over this PR and might have missed previous discussions

We have tests for batched alpha, but not moment (we should)

pymc/tests/test_distributions.py
```diff
@@ -2279,27 +2289,25 @@ def test_dirichlet_multinomial_vectorized(self, n, a, extra_size):
...
-            ),
+            (0.5, 4),
+            (2, 12),
+            (np.array([0.5, 1.0, 2.0]), 3),
+            (np.arange(1, 7, dtype="float64").reshape(2, 3), 5),
         ],
     )
-    def test_stickbreakingweights_logp(self, value, alpha, K, logp):
-        with Model() as model:
+    def test_stickbreakingweights_logp(self, alpha, K, _compile_stickbreakingweights_logpdf):
```
purna135: Is it acceptable to combine `test_stickbreakingweights_logp` and `test_stickbreakingweights_vectorized` in a single test?

pymc/tests/test_distributions.py
purna135: I now use distribution to generate value.

purna135:
> We have tests for batched alpha, but not moment (we should)

I need some assistance calculating `expected` in `test_stickbreakingweights_moment`.

pymc/tests/test_distributions.py
Sayam753: If `test_stickbreakingweights_vectorized` fails, it would point to shapes not being handled properly to form batches. So, lets keep them separate to have better isolation of tests.

purna135: Now I separate the test for batched alpha

pymc/tests/test_distributions.py
```diff
@@ -952,6 +953,24 @@ def test_hierarchical_obs_logp():
...
+    return compile_pymc([_value, _alpha, _k], _logp)
+
+
+def _stickbreakingweights_logpdf(value, alpha, k, _compile_stickbreakingweights_logpdf):
+    return _compile_stickbreakingweights_logpdf(value, alpha, k)
+
+
+stickbreakingweights_logpdf = np.vectorize(
+    _stickbreakingweights_logpdf, signature="(n),(),(),()->()"
+)
```
ricardoV94: I am not sure if this is the one that should have the `pytest.fixture`, it seems like it should be a function that returns the vectorized `stickbreakingweights_logpdf`. The idea being that we only compile the Aesara function if the test that requires it, is run. Maybe it works like this already, if `np.vectorize` is lazy?

Otherwise my gut feeling is that it should be something like:

```suggestion
@pytest.fixture(scope="module")
def stickbreakingweights_logpdf():
   
    _value = at.vector()
    _alpha = at.scalar()
    _k = at.iscalar()
    _logp = logp(StickBreakingWeights.dist(_alpha, _k), _value)
    core_fn = compile_pymc([_value, _alpha, _k], _logp)
    
    return np.vectorize(_stickbreakingweights_logpdf, signature="(n),(),()->()")
```

Also it seems like you had too many parameters in the signature of np.vectorize? I removed one of the parenthesis.

ricardoV94:
Left a comment above about the fixture. Also don't forget @larryshamalama remark above that we should test the `moment` function works for batched alpha as well. The existing tests are in here: https://github.com/pymc-devs/pymc/blob/7af102d40f7b64184cd0fa013ced8570772fc8eb/pymc/tests/test_distributions_moments.py#L1171-L1174

Should be enough to test with a vector of two alphas, maybe one of those that is already tested for single alpha (reusing the same k) and the other being an extreme value like alpha=1 or alpha=0 (if that's valid), which might have a very simple moment.

purna135:
Yes, I got the test for `moment` but I am not sure how the `expected` is calculated here. 
Is there any equation to determine the `expected`?

https://github.com/pymc-devs/pymc/blob/7af102d40f7b64184cd0fa013ced8570772fc8eb/pymc/tests/test_distributions_moments.py#L1147-L1151

ricardoV94:
> Yes, I got the test for `moment` but I am not sure how the `expected` is calculated here. Is there any equation to determine the `expected`?
> 
> https://github.com/pymc-devs/pymc/blob/7af102d40f7b64184cd0fa013ced8570772fc8eb/pymc/tests/test_distributions_moments.py#L1147-L1151

You can check what the moment is for two distinct single alphas, and it should be the same for a batched alpha that has those two values.

purna135:
Ok got it now, do I need to create a separate test for batched alpha as we did in `TestStickBreakingWeights_1D_alpha` ?

ricardoV94:
> Ok got it now, do I need to create a separate test for batched alpha as we did in `TestStickBreakingWeights_1D_alpha` ?

Nope, you can just add it as an extra condition in the existing tests. Moments is less sensitive than logp so we can keep it bundled together

ricardoV94 (approved):

ricardoV94:
Looks complete to me. The failing test is unrelated. Just asked if @larryshamalama could leave a review as well.

pymc/tests/test_distributions.py
```diff
@@ -2312,6 +2324,25 @@ def test_stickbreakingweights_invalid(self):
...
         assert pm.logp(sbw_wrong_K, np.array([0.4, 0.3, 0.2, 0.1])).eval() == -np.inf
 
+    @pytest.mark.parametrize(
+        "alpha,K",
+        [
+            (np.array([0.5, 1.0, 2.0]), 3),
+            (np.arange(1, 7, dtype="float64").reshape(2, 3), 5),
+        ],
+    )
+    def test_stickbreakingweights_vectorized(self, alpha, K, stickbreakingweights_logpdf):
```
larryshamalama, line 2334: Is `stickbreakingweights_logpdf` passed as an argument here via the fixture sharing decorator of the function?

larryshamalama (approved):
Looks great @purna135! Just one question for myself 😅

pymc/tests/test_distributions.py
ricardoV94: Yes, that's how fixtures work

PR #6046, opened 2022-08-10
Added networkx export functionality

jonititan:
Apologies for delay but finally got around to updating the latest version of model_graph.py to include networkx export function
as discussed in [5677](https://github.com/pymc-devs/pymc/pull/5677)
This creates a method for exporting the graph network structure to networkx.
This is useful as it allows direct conversion instead of exporting to graphviz, saving to dot file, importing the dotfile via pydot before finally creating the networkx graph.

This is a new method so there should be no breaking changes.
Docstrings have all been updated.

The same approach could be used in v4. I did it in v3 because that's what I'm currently using.

Editted
This was copy paste error from previous pull request.  This current pull request is for v4 main and was tested on v4.0.0

**What is this PR about?**
...

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- ...


ricardoV94:
Thanks for the follow up. 

It seems there are several unrelated whitespace changes. Could you remove them?

> The same approach could be used in v4. I did it in v3 because that's what I'm currently using.

The current PR is directed at main, which is V4.

jonititan:
I think that sorted the whitespace issues?
Sorry I actually realised just after submitting.

jonititan:
and yes this is directed at main v4 and was tested on v4.0.0

pymc/model_graph.py
```diff
@@ -125,14 +125,17 @@ def make_compute_graph(
 
         return input_map
 
-    def _make_node(self, var_name, graph, *, formatting: str = "plain"):
-        """Attaches the given variable to a graphviz Digraph"""
+    def _make_node(self, var_name, graph, *, nx=False, cluster=False, formatting: str = "plain"):
```
ColCarroll, line 128: `formatting` is unused here, right? (not from your change, just observing)

ColCarroll, line 128: it might be pleasant to update the signature to just take `graph_node_fn`, which is either `graph.add_node` or `graph.node`, since they both appear to do this update in place. then we do not need a specific `nx` argument here.

ColCarroll (commented):
i'm not familiar with networkx (and a little surprised it has such a similar api!), but this seems like a nice addition! A few asks, and feel free to push back if these are onerous:

1. Can you post a screenshot of the output of the sample code from `model_to_networkx` in a notebook?
2. can you add a test in `model_to_graphviz.py`? 
3. it looks like there are some edits (maybe from an autoformatter?) on unrelated lines. if it isn't terrible, can you remove those?

as a side conversation that maybe you already had -- does networkx know how to directly read in a dot file? would it be possible to have `model_to_networkx` be (something like) `nx.dot_to_nx(model_to_graphviz(model).source)`? that would be very attractive, since it would not require extra code or tests here!

ColCarroll:
(I tagged @ericmjl even though he's busy in case he has wisdom to share here, but no pressure!)

pymc/model_graph.py
jonititan: > `formatting` is unused here, right? (not from your change, just observing)

Yes not mine and I don't think its used.  I think it was because you can pass Tex math and things?

jonititan:
> i'm not familiar with networkx (and a little surprised it has such a similar api!), but this seems like a nice addition! A few asks, and feel free to push back if these are onerous:
> 
>     1. Can you post a screenshot of the output of the sample code from `model_to_networkx` in a notebook?
> 
>     2. can you add a test in `model_to_graphviz.py`?
> 
>     3. it looks like there are some edits (maybe from an autoformatter?) on unrelated lines. if it isn't terrible, can you remove those?
> 
> 
> as a side conversation that maybe you already had -- does networkx know how to directly read in a dot file? would it be possible to have `model_to_networkx` be (something like) `nx.dot_to_nx(model_to_graphviz(model).source)`? that would be very attractive, since it would not require extra code or tests here!

1. yes I'll do that shortly

2. Did you mean add it to this? [test_model_graph.py](https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_model_graph.py)


3. I think i fixed the whitespace but happy to do it if not.

 networkx plays just fine with dot files
https://networkx.org/documentation/stable/reference/drawing.html
There are also examples [here](https://networkx.org/documentation/latest/auto_examples/index.html#graphviz-drawing) on plotting using graphviz from networkx

It's just very annoying as I wanted to export directly and manipulate the network graph and add things directly from the pymc model without having to pass the graph to string and back as dot first.
I also ran into errors using that method where some links were not retained but this direct method I've added preserves all of that.
When I patched it locally back in April it was so I could plot my pymc model interactively and inspect all the node values, distributions, etc
Made it easier to show others.  Please see attached.
networkdiagram

it uses [d3graph](https://erdogant.github.io/d3graph/pages/html/Core%20Functionalities.html?highlight=networkx)

jonititan:
Here's a screenshot.  I saved my code as alt_model_graph.py and you can see where I import it.
The model is just a draft I was messing around with
actionscreenshot
.

jonititan:
Sorry just realised you meant the code in the docstring.  Wait one

jonititan:
Ok that throws an error I will fix and update the pull request.

jonititan:
Interestingly I fixed that and found a bug that I didn't understand.
In that example there is a self loop 
obs -> obs
However when iterating over the same data networkx is never passed a parent child pair of obs and obs to create that edge.
Turns out it was removed from main but isn't in the version I have installed yet(v4.0.0)
https://github.com/pymc-devs/pymc/commit/ce07e7066ff6e08c62280974f55ecbd5e15fa6f9

New update fixes the issue and matches behaviour of latest version(no self loop for observed)
screenshot2

jonititan:
OK I tested my fix and attributes aern't preserved for subgraphs correctly as i'm using compose to combine subgraphs and it overwrites attributes by default. https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.operators.binary.compose.html?highlight=compose
I've got to put the kids to bed but afterwards i'll think of a nice way to sort it that works for all attributes instead of just the ones I see currently.

jonititan:
Right i've sorted out the attribute behaviour.
networkx doesn't have a way to add attributes to a subgraph you either make them graph, node, or edge attributes.
I've made them node attributes and every node in a cluster can be identifield by its cluster attribute which storesthe cluster/plate name

pymc/model_graph.py
```diff
@@ -234,20 +242,100 @@ def make_graph(self, var_names: Optional[Iterable[VarName]] = None, formatting:
...
+            for parent in parents:
+                graphnetwork.add_edge(parent.replace(":", "&"), child.replace(":", "&"))
+        return graphnetwork
+
+
+def model_to_networkx(
+    model=None, *, var_names: Optional[Iterable[VarName]] = None, formatting: str = "plain"
+):
+    """Produce a networkx Digraph from a PyMC model.
+    Requires networkx, which may be installed most easily with
```
OriolAbril:
```suggestion
    """Produce a networkx Digraph from a PyMC model.
    
    Requires networkx, which may be installed most easily with
```

There must be a new line between the short summary and the extended description. Docstring formatting should follow [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html), I recommend you go over https://pymc-data-umbrella.xyz/en/latest/sprint/tutorials/docstring_tutorial.html#edit-the-docstring.

Moreover, the lack of empty lines between sections is breaking the rendering of the docstring. Docstrings are written in rst which is not markdown but it has many similarities like for example, ignoring breaklines and adding breaklines only when an empty line is present. 

* model_to_graphviz docstring:
  - With this PR: https://pymcio--6046.org.readthedocs.build/projects/docs/en/6046/api/generated/pymc.model_to_graphviz.html
  - on latest https://www.pymc.io/projects/docs/en/latest/api/generated/pymc.model_to_graphviz.html

The same way you need to add the spaces here, you need to _keep_ them on model_to_graphviz, your PR is currently deleting all the empty lines in that docstring (which is the one I linked to show the rendering issues). 

Also, you need to add this new function to the docs, you should list it in https://github.com/pymc-devs/pymc/blob/main/docs/source/api/model.rst right under model_to_graphviz

OriolAbril (changes requested):

pymc/model_graph.py
jonititan: Ok I've done both of those.
Thanks for highlighting

pymc/model_graph.py
jonititan: Perhaps that is a future improvement?

ColCarroll: yeah, looks like it -- this is low-hanging fruit for a future 'tidying' PR, but doesn't need any action here.

ColCarroll: sounds good to me!

ColCarroll:
> Did you mean add it to this? [test_model_graph.py](https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_model_graph.py)

Yes! apologies. A test or two would be super helpful here. 

Right now formatting is causing the test suite to fail -- the docstrings normally need a little more white space than you have here (probably running `pre-commit run --all-files` after installing https://pre-commit.com/ in your environment will fix it)

jonititan:
ok pre-commit has run successfully and i've committed those changes.

I'll have a think re tests

jonititan:
Right i've added a simple test in to test_model_graph.py.  I think i've understood how that test function file works correctly.
I extended SeededTest as used for the modelgraph tests but didn't add as many different tests.

How many test types are needed?

pymc/tests/test_model_graph.py
```diff
@@ -23,7 +23,60 @@
...
+    'obs': {'tau': {}, 'eta': {}, 'mu': {}},
+    'tau': {},
+    'mu': {}},
+    '_succ': {'eta': {'obs': {}},
+    'obs': {},
+    'tau': {'obs': {}},
+    'mu': {'obs': {}}}}
+
+    def test_networkx(self):
+        assert self.network_model == model_to_networkx(school_model()).__dict__
```
OriolAbril, line 78: you are not importing `model_to_networkx` on the import sections, the test will fail for now (thus waiting on aproving the github actions jobs)

pymc/model_graph.py
```diff
@@ -234,20 +248,133 @@ def make_graph(self, var_names: Optional[Iterable[VarName]] = None, formatting:
...
+    model=None,
+    *,
+    var_names: Optional[Iterable[VarName]] = None,
+    formatting: str = "plain",
 ):
     """Produce a graphviz Digraph from a PyMC model.
 
     Requires graphviz, which may be installed most easily with
         conda install -c conda-forge python-graphviz
-
```
OriolAbril: You are removing an empty line here, it should not be the case as it affects rendering. If you ckeck the preview you'll see the short description is now on its own line but the rest of the docstring is wrong. 

Lines 250, 259, 263 265, 268, 272, 274 and more are also being remove when they should not. You can take a look at the diff below here on github, they are all marked in red.

pymc/model_graph.py
```diff
@@ -234,20 +248,133 @@ def make_graph(self, var_names: Optional[Iterable[VarName]] = None, formatting:
...
+
+def model_to_networkx(
+    model=None,
+    *,
+    var_names: Optional[Iterable[VarName]] = None,
+    formatting: str = "plain",
+):
+    """Produce a networkx Digraph from a PyMC model.
+
+    Requires networkx, which may be installed most easily with
```
OriolAbril: see the comment about the graphviz docstring, you need to use similar spacing than it does. the `.. code-block` needs an empty line before and after to be rendered correctly, the section titles need spaces right above them...

OriolAbril (dismissed):
I assume the goal is for the function to be used as `pymc.model_to_networkx` (which is what is being documented) which won't work yet. For this to work you need to add it to https://github.com/pymc-devs/pymc/blob/main/pymc/__init__.py#L68

ColCarroll:
Test looks good -- it looks like it mostly verifies that this works on certain inputs, and I'm ok for now with that (this seems like a cool feature, and if it gets use that reveals bugs, we can fix those and add more tests). 

The linting is still broken, and I'll let @OriolAbril or @ricardoV94 approve/merge once that is fixed, but I approve of the functionality and the tests, for what that's worth!

pymc/tests/test_model_graph.py
jonititan: Sorry I had that in my local file but missed it when I updated it on GitHub.

I'm still new to CI so I'm not working integrated yet.

I'll fix.

It did run pre-commit so I I blame it this time not me.

I will try to fix the whitespace again.

jonititan: Ok I've done both of those.

Thank you

pymc/model_graph.py
jonititan: Right I think i've fixed that.  Thanks for bearing with me on this

jonititan:
I think i've solved all the linting issues now

ColCarroll:
:( thanks for bearing with it -- it looks like it is still a little mucked up. Sorry I can't offer more guidance - I haven't been contributing a ton lately!

jonititan:
It passed pre-commit though?
And it has the docstring spacing?

Or at least it passed on my pc looks like it's failing here.

I'll spend some time on my setup.
Maybe I'm introducing formatting issues when I ctrl+c ctrl+v
This time I'll make sure to pull direct and then push back to try to avoid those issues.

jonititan:
Well I pulled from main, carefully applied my changes, ran pre-commit only on the four files I changed, and then pushed to my patch-2 branch.


pymc\__init__.py
pymc\tests\test_model_graph.py
pymc\model_graph.py
pymc\docs\source\api\models.rst



Could you confirm the formatting is ok?
Pre-commit seems to still be failing on github but it passed locally on my machine so i don't know whats happening there.
Sorry for all this hassle.

jonititan:
Looks like everything passes in Github except pre-commit itself which gives an opaque error message.
`Error: The process '/opt/hostedtoolcache/Python/3.10.6/x64/bin/pre-commit' failed with exit code 1`

jonititan:
I took a guess it might be because networkx needs to be added to the requirements so i've done that and when it re runs hopefully that clears it all up.

jonititan:
Can we re-run the pre-commit action with debug enabled? Or does anyone have an idea what is causing it to fail?

ricardoV94:
I am not sure about adding a new dependency.    Can we make it optional? You can use `pytest.importorskip` for the tests

michaelosthege:
> Can we re-run the pre-commit action with debug enabled? Or does anyone have an idea what is causing it to fail?

So see the failure reason click the [Details](https://github.com/pymc-devs/pymc/runs/7808557176?check_suite_focus=true) of the failed ❌ `pre-commit` job.

It failed the `black` step which normalizes code style.
You have two options:
* Make the same changes it shows in the log by hand,
* or run the pre-commit locally (recommended).

First `pip install pre-commit` and second `pre-commit run --all` to manually run it.
This second step makes the necessary changes which you'll have to commit.
You can also run `pre-commit install` (vs. `pre-commit uninstall`) to set it up to automatically run every time you make/amend a commit.

jonititan:
I tried running it twice this time before committing. I push it to my path repo and am running the same tests on it there.  If that all works I will report back here for workflows to be retriggered.

jonititan:
Ok all tests passed on my patch repo so It should surely pass this time here.

jonititan:
> I assume the goal is for the function to be used as `pymc.model_to_networkx` (which is what is being documented) which won't work yet. For this to work you need to add it to https://github.com/pymc-devs/pymc/blob/main/pymc/__init__.py#L68

I've made that change you requested @OriolAbril could you confirm please?

pymc/model_graph.py
```diff
@@ -234,9 +239,133 @@ def make_graph(self, var_names: Optional[Iterable[VarName]] = None, formatting:
...
```
OriolAbril:
```suggestion
    Requires networkx, which may be installed most easily with::
    
```

to render this as literal block: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

pymc/model_graph.py
```diff
@@ -234,9 +239,133 @@ def make_graph(self, var_names: Optional[Iterable[VarName]] = None, formatting:
...
+        conda install networkx
+
+    Alternatively, you may install using pip with
+        pip install networkx
```
OriolAbril:
```suggestion
    Alternatively, you may install using pip with::
    
        pip install networkx
        
```

pymc/model_graph.py
```diff
@@ -234,9 +239,133 @@ def make_graph(self, var_names: Optional[Iterable[VarName]] = None, formatting:
...
+    See
+    https://networkx.org/documentation/stable/
+    for more information.
```
OriolAbril:
```suggestion
    See https://networkx.org/documentation/stable/ for more information.
```

pymc/model_graph.py
```diff
@@ -234,9 +239,133 @@ def make_graph(self, var_names: Optional[Iterable[VarName]] = None, formatting:
...
+
+    Parameters
+    ----------
+    model : pm.Model
```
OriolAbril:
```suggestion
    model : Model
```

sphinx doesn't understand `pm.Model`, it should use the alias we have set up `Model` (recommended) or `pymc.Model` so it is rendered as a link

pymc/model_graph.py
```diff
@@ -234,9 +239,133 @@ def make_graph(self, var_names: Optional[Iterable[VarName]] = None, formatting:
...
+        The model to plot. Not required when called from inside a modelcontext.
+    var_names : iterable of variable names, optional
```
OriolAbril:
```suggestion
    var_names : iterable of str, optional
```

OriolAbril (commented):
Docs look ok now
* model "module" api page: https://pymcio--6046.org.readthedocs.build/projects/docs/en/6046/api/model.html
* model_to_networkx rendered docstring: https://pymcio--6046.org.readthedocs.build/projects/docs/en/6046/api/generated/pymc.model_to_networkx.html#pymc.model_to_networkx

Left some extra comments to make them look a bit better. Doc-wise it can be merged now, I am removing the request changes.

jonititan:
Right i've fixed the trailing whitespace issue and it is [passing pre-commit tests again.](https://github.com/jonititan/pymc/runs/7838895866?check_suite_focus=true)

ColCarroll (approved):
this looks good to me! will merge once tests pass. thank you for bearing with the pre-commit rigamarole.

jonititan:
Great.
Thanks likewise for bearing with me while I learned on the go.

jonititan:
It's the same category as graphviz

Not required for typical workflows.

PR #6055, opened 2022-08-18
update the default value of jitter to JITTER_DEFAULT

danhphan:
This PR makes a minor change in gp models by setting the default value of `jitter` to `JITTER_DEFAULT`, instead of `0.0`. 

This could reduce the number of errors (like positive definite errors) when calling `conditional` functions.

I am not sure, but I'm not aware if there is any cases that the default jitter of `0.0` will be useful for end users. Please confirm if I'm wrong. Thank you.

### DIFF
--- a/pymc/gp/gp.py
+++ b/pymc/gp/gp.py
@@ -160,7 +160,7 @@ class Latent(Base):
             variable by the Cholesky factor of the covariance matrix.
         jitter: scalar
             A small correction added to the diagonal of positive semi-definite
-            covariance matrices to ensure numerical stability.  Default value is 1e-6.
+            covariance matrices to ensure numerical stability.
         **kwargs
             Extra keyword arguments that are passed to distribution constructor.
         """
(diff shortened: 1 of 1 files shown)


danhphan:
Hi, not sure why `pymc/tests/test_sampling_jax.py` not passed. It seems running alright at my local machine.

bwengals:
Reran the other build that failed, it passes now.  The readthedocs build is failing I think because of the same issue upstream with jax that has since been fixed.  Is there a way to rerun it to verify?  I don't see anything in this PR that might cause an issue with docs.  

It [looks possible](https://github.com/readthedocs/readthedocs.org/issues/6524) but maybe the feature is too new.  I may not see a rebuild button because I'm not an admin on RTD, maybe @OriolAbril can see a rebuild button, if you have a sec?

OriolAbril:
asked for a rebuild, if it doesn't work it probably needs a rebase

bwengals:
Thanks for running that @OriolAbril, looks like everythings green now!

bwengals (approved):
lgtm, and thanks for fixing this @danhphan this will prevent a lot of unnecessary errors for people trying to use this

danhphan:
Hi, thanks @bwengals and @OriolAbril :)

michaelosthege (approved):

danhphan:
Hi, thanks @michaelosthege

PR #6059, opened 2022-08-23
Upgrade to aesara=2.8.2 and aeppl=0.0.35

### LINKED ISSUE #6053: Adapt to new aesara API renaming optimize -> rewrite
https://github.com/aesara-devs/aesara/pull/1054

Armavica:
**What is this PR about?**
Adapting to the new aesara API, fixes #6053.

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Upgrade to aesara=2.8.2 and aeppl=0.0.35


twiecki:
Thanks for taking this on @Armavica!

Armavica:
Now the only tests that do not pass are the ones related to the random variable `WeibullRV` of `aesara` which might need fixing: https://github.com/aesara-devs/aesara/issues/1141

Armavica:
So apparently, as discussed in the linked aesara issue, the `WeibullRV` has been reparameterized to only keep the shape parameter: the scale parameter disappeared.

So I am not sure what to do. The two options that I see are:

* Keep the current PyMC parameterization and reparameterize for aesara on the fly. Pros: no breaking change and keep a consistent scale parameter among most continuous distributions. Cons: It might be rather hackish code-wise.
* Propagate the change of parameterization to PyMC, to make `Weibull` require only one parameter. Pros: one-to-one mapping with aesara. Cons: breaking change, inconsistencies between distributions with a scale parameters and those without.

What are your thoughts?

ricardoV94:
We already have a weibull with scale: https://github.com/pymc-devs/pymc/blob/f7315a37919afde9d77e7e4b1556b385ec5521f9/pymc/distributions/continuous.py#L2521

It seems we only need to implement the logp method (you can copy the one that was being used in aeppl until recently)

pymc/distributions/continuous.py
```diff
@@ -1464,7 +1463,7 @@ def dist(cls, lam, *args, **kwargs):
         lam = at.as_tensor_variable(floatX(lam))
 
         # Aesara exponential op is parametrized in terms of mu (1/lam)
-        return super().dist([at.inv(lam)], **kwargs)
+        return super().dist([at.reciprocal(lam)], **kwargs)
```
ricardoV94, line 1466: Why this change to reciprocal (and the others below)?

pymc/distributions/continuous.py
```diff
@@ -2518,14 +2517,38 @@ def logcdf(value, nu):
 
 
 # TODO: Remove this once logp for multiplication is working!
-class WeibullBetaRV(WeibullRV):
-    ndims_params = [0, 0]
+class WeibullBetaRV(RandomVariable):
+    name = "weibull"
+    ndim_supp = 0
+    ndims_params = [0]
```
ricardoV94: You need one per parameter:

```suggestion
    ndims_params = [0, 0]
```

pymc/distributions/continuous.py
```diff
@@ -2518,14 +2517,38 @@ def logcdf(value, nu):
...
+
+    def __call__(self, alpha, beta, size=None, **kwargs):
+        return super().__call__(alpha, beta, size=size, **kwargs)
 
     @classmethod
     def rng_fn(cls, rng, alpha, beta, size) -> np.ndarray:
         return np.asarray(beta * rng.weibull(alpha, size=size))
 
 
+@_logprob.register(WeibullBetaRV)
```
ricardoV94: For the time being we should keep it consistent and implement it as the fake `logp` method of the PyMC Weibull distribution class, as is done for other distributions in this module.

In that case you don't need to do the value unwrap thing either.

ricardoV94 (changes requested):

Armavica:
Thank you for your help. The tests are now all passing, so I think that the PR is ready for review.

It took me some time to figure out what I had to do for this Weibull distribution, and I still don't understand how everything works… So I hope that I did what you meant, but otherwise I would be happy to redo it differently.

pymc/distributions/continuous.py
Armavica: I had the impression that inv was deprecated in favor of reciprocal, did I misunderstand? see https://github.com/aesara-devs/aesara/pull/1054/commits/1a6fecf9b4cd47b59727e9f01d6981c17018b291 (actually it looks like `at.inv` doesn't exist anymore)

pymc/distributions/continuous.py
Armavica: Ok, thank you, I will fix that.

pymc/distributions/continuous.py
Armavica: Ok, I will try like you say.

Armavica:
I realized that I forgot to address some deprecations which only raised warnings and not hard errors. So I am going to pass this PR back into draft mode until this is done.

pymc/distributions/continuous.py
ricardoV94: Oh I missed that! You are absolutely correct.

ricardoV94 (approved):
Awesome work!

twiecki:
Very cool, and already in our new release: https://github.com/pymc-devs/pymc/releases/tag/v4.1.6

PR #6063, opened 2022-08-25
Fix broadcasting via observed and dims

### LINKED ISSUE #5993: Implied dimensions of RVs are not broadcasted by dims or observed
The dims argument is not respected when a random variable is created with parameters that broadcast to the desired output dimensions:

```python
coords = {
     "dim1": pd.RangeIndex(10),
     "dim2": pd.RangeIndex(7)
}

     mu = np.zeros((10, 1))
     x = pm.Normal("x", mu=mu, dims=("dim1", "dim2"))
     y = pm.Normal("y", mu=mu, dims=("dim1", "dim2"), shape=(10, 7))

x.eval().shape
# Returns (10, 1), when it should be (10, 7)

y.eval().shape
# Correctly returns (10, 7)
```

* PyMC/PyMC3 Version: latest release 4.1.3
* Aesara/Theano Version:
* Python Version:
* Operating system:
* How did you install PyMC/PyMC3: conda

ricardoV94:
Fixes #5993 

This PR also removes the support of ellipsis in dims and shape. This feature was not very advertised so hopefully this won't cause too much pain.

These tests illustrate the bug fix

```python
    def test_broadcast_by_dims(self):
        with pm.Model(coords={"broadcast_dim": range(3)}) as m:
            x = pm.Normal("x", mu=np.zeros((1,)), dims=("broadcast_dim",))
            assert x.eval().shape == (3,)

    def test_broadcast_by_observed(self):
        with pm.Model() as m:
            x = pm.Normal("x", mu=np.zeros((1,)), observed=np.zeros((3,)))
            assert x.eval().shape == (3,)
```

**Checklist**

## Major / Breaking Changes
- When shape is not provided, dims or observed are used to define the shape of a variable. If you want a variable that has dims or is observed to automatically resize when its inputs change, you must specify that explicitly via the shape argument. For example, `pm.Normal("likelihood", mu=mu, sigma=sigma, observed=data, dims="data", shape=mu.shape)`
- Remove support of Ellipsis (`...`) in shape and dims

## Bugfixes / New features
- Fix bug where distribution shape would not be broadcasted by dims or observed
- Allow specifying dims on the fly from observed
- Do not show shape-related dependencies of RandomVariables in `model_graph`

## Docs / Maintenance
- Update `_make_nice_attr_error` to suggest `pm.draw` instead of `.eval()`
- Add `_make_nice_attr_error` to `SymbolicDistribution`s

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -209,7 +209,8 @@ class Distribution(metaclass=DistributionMeta):
         rng : optional
             Random number generator to use with the RandomVariable.
         dims : tuple, optional
-            A tuple of dimension names known to the model.
+            A tuple of dimension names known to the model. When shape is not provided,
+            the shape of dims is used to define the shape of the variable.
         initval : optional
             Numeric or symbolic untransformed initial value of matching shape,
             or one of the following initial value strategies: "moment", "prior".
@@ -217,6 +218,8 @@ class Distribution(metaclass=DistributionMeta):
             or moment-based initial values in the transformed space.
         observed : optional
             Observed data to be passed when registering the random variable in the model.
+            When neither shape nor dims is provided, the shape of observed is used to
+            define the shape of the variable.
             See ``Model.register_rv``.
         total_size : float, optional
             See ``Model.register_rv``.
@@ -405,7 +408,8 @@ class SymbolicDistribution:
         name : str
             Name for the new model variable.
         dims : tuple, optional
-            A tuple of dimension names known to the model.
+            A tuple of dimension names known to the model. When shape is not provided,
+            the shape of dims is used to define the shape of the variable.
         initval : optional
             Numeric or symbolic untransformed initial value of matching shape,
             or one of the following initial value strategies: "moment", "prior".
@@ -413,6 +417,8 @@ class SymbolicDistribution:
             symbolic or moment-based initial values in the transformed space.
         observed : optional
             Observed data to be passed when registering the random variable in the model.
+            When neither shape nor dims is provided, the shape of observed is used to
+            define the shape of the variable.
             See ``Model.register_rv``.
         total_size : float, optional
             See ``Model.register_rv``.
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -33,18 +33,18 @@ from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.var import TensorVariable
 from typing_extensions import TypeAlias
 
-from pymc.aesaraf import change_rv_size
+from pymc.aesaraf import change_rv_size, convert_observed_data
 from pymc.distributions.shape_utils import (
     Dims,
     Shape,
     Size,
+    StrongDims,
     StrongShape,
     convert_dims,
     convert_shape,
     convert_size,
     find_size,
-    resize_from_dims,
-    resize_from_observed,
+    shape_from_dims,
 )
 from pymc.printing import str_for_dist, str_for_symbolic_dist
 from pymc.util import UNSET
@@ -152,29 +152,28 @@ def _make_nice_attr_error(oldcode: str, newcode: str):
 def _make_rv_and_resize_shape(
     *,
     cls,
-    dims: Optional[Dims],
+    dims: Optional[StrongDims],
     model,
     observed,
     args,
     **kwargs,
-) -> Tuple[Variable, Optional[Dims], Optional[Union[np.ndarray, Variable]], StrongShape]:
-    """Creates the RV and processes dims or observed to determine a resize shape."""
-    # Create the RV without dims information, because that's not something tracked at the Aesara level.
-    # If necessary we'll later replicate to a different size implied by already known dims.
-    rv_out = cls.dist(*args, **kwargs)
-    ndim_actual = rv_out.ndim
+) -> Tuple[Variable, StrongShape]:
+    """Creates the RV, possibly using dims or observed to determine a resize shape (if needed)."""
     resize_shape = None
+    size_or_shape = kwargs.get("size") or kwargs.get("shape")
 
-    # # `dims` are only available with this API, because `.dist()` can be used
-    # # without a modelcontext and dims are not tracked at the Aesara level.
-    dims = convert_dims(dims)
-    dims_can_resize = kwargs.get("shape", None) is None and kwargs.get("size", None) is None
-    if dims is not None:
-        if dims_can_resize:
-            resize_shape, dims = resize_from_dims(dims, ndim_actual, model)
-    elif observed is not None:
-        resize_shape, observed = resize_from_observed(observed, ndim_actual)
-    return rv_out, dims, observed, resize_shape
+    # Create the RV without dims or observed information
+    rv_out = cls.dist(*args, **kwargs)
+
+    # Preference is given to size or shape, if not provided we use dims and observed
+    # to resize the variable
+    if not size_or_shape:
+        if dims is not None:
+            resize_shape = shape_from_dims(dims, tuple(rv_out.shape), model)
+        elif observed is not None:
+            resize_shape = tuple(observed.shape)
+
+    return rv_out, resize_shape
 
 
 class Distribution(metaclass=DistributionMeta):
@@ -254,15 +253,20 @@ class Distribution(metaclass=DistributionMeta):
         if not isinstance(name, string_types):
             raise TypeError(f"Name needs to be a string but got: {name}")
 
-        # Create the RV and process dims and observed to determine
-        # a shape by which the created RV may need to be resized.
-        rv_out, dims, observed, resize_shape = _make_rv_and_resize_shape(
+        dims = convert_dims(dims)
+        if observed is not None:
+            observed = convert_observed_data(observed)
+
+        # Create the RV, possibly taking into consideration dims and observed to
+        # determine its shape
+        rv_out, resize_shape = _make_rv_and_resize_shape(
             cls=cls, dims=dims, model=model, observed=observed, args=args, **kwargs
         )
 
+        # A shape was specified only through `dims`, or implied by `observed`.
         if resize_shape:
-            # A batch size was specified through `dims`, or implied by `observed`.
-            rv_out = change_rv_size(rv=rv_out, new_size=resize_shape, expand=True)
+            resize_size = find_size(shape=resize_shape, size=None, ndim_supp=cls.rv_op.ndim_supp)
+            rv_out = change_rv_size(rv=rv_out, new_size=resize_size, expand=False)
 
         rv_out = model.register_rv(
             rv_out,
@@ -336,11 +340,7 @@ class Distribution(metaclass=DistributionMeta):
         shape = convert_shape(shape)
         size = convert_size(size)
 
-        create_size, ndim_expected, ndim_batch, ndim_supp = find_size(
-            shape=shape, size=size, ndim_supp=cls.rv_op.ndim_supp
-        )
-        # Create the RV with a `size` right away.
-        # This is not necessarily the final result.
+        create_size = find_size(shape=shape, size=size, ndim_supp=cls.rv_op.ndim_supp)
         rv_out = cls.rv_op(*dist_params, size=create_size, **kwargs)
 
         rv_out.logp = _make_nice_attr_error("rv.logp(x)", "pm.logp(rv, x)")
@@ -448,19 +448,20 @@ class SymbolicDistribution:
         if not isinstance(name, string_types):
             raise TypeError(f"Name needs to be a string but got: {name}")
 
-        # Create the RV and process dims and observed to determine
-        # a shape by which the created RV may need to be resized.
-        rv_out, dims, observed, resize_shape = _make_rv_and_resize_shape(
+        dims = convert_dims(dims)
+        if observed is not None:
+            observed = convert_observed_data(observed)
+
+        # Create the RV, possibly taking into consideration dims and observed to
+        # determine its shape
+        rv_out, resize_shape = _make_rv_and_resize_shape(
             cls=cls, dims=dims, model=model, observed=observed, args=args, **kwargs
         )
 
+        # A shape was specified only through `dims`, or implied by `observed`.
         if resize_shape:
-            # A batch size was specified through `dims`, or implied by `observed`.
-            rv_out = cls.change_size(
-                rv=rv_out,
-                new_size=resize_shape,
-                expand=True,
-            )
+            resize_size = find_size(shape=resize_shape, size=None, ndim_supp=rv_out.tag.ndim_supp)
+            rv_out = cls.change_size(rv=rv_out, new_size=resize_size, expand=False)
 
         rv_out = model.register_rv(
             rv_out,
@@ -529,18 +530,17 @@ class SymbolicDistribution:
         shape = convert_shape(shape)
         size = convert_size(size)
 
-        create_size, ndim_expected, ndim_batch, ndim_supp = find_size(
-            shape=shape, size=size, ndim_supp=cls.ndim_supp(*dist_params)
-        )
-        # Create the RV with a `size` right away.
-        # This is not necessarily the final result.
-        graph = cls.rv_op(*dist_params, size=create_size, **kwargs)
+        ndim_supp = cls.ndim_supp(*dist_params)
+        create_size = find_size(shape=shape, size=size, ndim_supp=ndim_supp)
+        rv_out = cls.rv_op(*dist_params, size=create_size, **kwargs)
+        # This is needed for resizing from dims in `__new__`
+        rv_out.tag.ndim_supp = ndim_supp
 
         # TODO: Create new attr error stating that these are not available for DerivedDistribution
         # rv_out.logp = _make_nice_attr_error("rv.logp(x)", "pm.logp(rv, x)")
         # rv_out.logcdf = _make_nice_attr_error("rv.logcdf(x)", "pm.logcdf(rv, x)")
         # rv_out.random = _make_nice_attr_error("rv.random()", "rv.eval()")
-        return graph
+        return rv_out
 
 
 @singledispatch
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -19,7 +19,7 @@ import warnings
 
 from abc import ABCMeta
 from functools import singledispatch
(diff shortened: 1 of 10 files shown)


pymc/tests/test_shape_handling.py
```diff
@@ -304,7 +324,7 @@ def test_can_resize_data_defined_size(self):
         with pm.Model() as pmodel:
             x = pm.MutableData("x", [[1, 2, 3, 4]], dims=("first", "second"))
             y = pm.Normal("y", mu=0, dims=("first", "second"))
-            z = pm.Normal("z", mu=y, observed=np.ones((1, 4)))
+            z = pm.Normal("z", mu=y, observed=np.ones((1, 4)), size=y.shape)
```
ricardoV94, line 342: This behavior was neat, but could cause bugs, because observed may broadcast the shape implied by the parameters. 

One must specify explicitly how the shape flows from the parameters as in this updated test.

pymc/distributions/distribution.py
```diff
@@ -150,35 +149,35 @@ def fn(*args, **kwargs):
...
+    size_or_shape = kwargs.get("size") or kwargs.get("shape")
+    dims = convert_dims(dims)
+
+    # Preference is given to size or shape. If not specified, we rely on dims and
+    # finally, observed, to determine the shape of the variable. Because dims can be
+    # specified on the fly, we need a two-step process where we first create the RV
+    # without dims information and then resize it.
+    # TODO: Consider removing dims-on-the-fly for simplicity sake.
+    if not size_or_shape and observed is not None:
+        kwargs["shape"] = shape_from_observed(observed)
```
twiecki: Does this have any effect?

ricardoV94: Yes, check the new `test_define_dims_on_the_fly_from_observed`

ricardoV94:
Started a discussion for removing dims on the fly: https://github.com/pymc-devs/pymc/discussions/6065

ricardoV94:
I seem to have broken some things according to the failing `test_density_scaling_with_generator`. It seems that we support generators as observed data... which means we can only inquiry about data once (and then we must pass it around). Is this something we really need?

Seems to be used by Minibatch, cc @ferrine

pymc/distributions/distribution.py
```diff
@@ -150,35 +149,34 @@ def fn(*args, **kwargs):
...
-        if dims_can_resize:
-            resize_shape, dims = resize_from_dims(dims, ndim_actual, model)
-        elif Ellipsis in dims:
-            # Replace ... with None entries to match the actual dimensionality.
-            dims = (*dims[:-1], *[None] * ndim_actual)[:ndim_actual]
-    elif observed is not None:
-        resize_shape, observed = resize_from_observed(observed, ndim_actual)
-    return rv_out, dims, observed, resize_shape
+
+    if not size_or_shape and dims is not None:
```
aseyboldt, line 175: Couldn't we use this spot to do some error checking? If a size or shape is specified and dims is also specified, then we could check if they are consistent, and throw an error if they are not. So we wouldn't have to say something "shape takes precedent", but just: "If you specify both they must say the same thing". Otherwise I think we would get an error when we try to export a trace to arviz, right? And we'd rather get the error at definition, than getting it after sampling? Or is there a check for this somewhere else already?

We could then add a test like this:

```python
    with pm.Model(coords={"a": [1, 2]}) as model:
        pm.Normal("y", shape=3, dims="a")
```

ricardoV94: The question about precedence in this comment/ section is not related to error checking but what inputs ultimately go into the RV size argument. The reason we give preference to `size`/`shape` is because it can be an arbitrary expression and so is more powerful in principle than dims.

This PR keeps the same level of no-inconsistency checking as before. The direct tests in `test_shape_handling` mention this in their comments. 

I would leave such discussion outside of this PR unless it feels very relevant.

aseyboldt: Sure, it isn't really related to the issue in the PR.

I'll open a separate PR about that, I think we can at least do some checks if the shape/size is an integer. Has been too long anyway since I did my last pymc PR. :-)

ricardoV94:
Tests are passing

twiecki (approved):

michaelosthege:
> Tests are passing

👍 I will review this tonight!

ricardoV94:
> > Tests are passing
> 
> +1 I will review this tonight!

I'll wait for it!

pymc/distributions/distribution.py
```diff
@@ -150,35 +149,34 @@ def fn(*args, **kwargs):
...
+    """Creates the RV, possibly using dims or observed to determine a resize shape (if needed)."""
+    resize_shape_from_dims = None
+    size_or_shape = kwargs.get("size") or kwargs.get("shape")
+
+    # Preference is given to size or shape. If not specified, we rely on dims and
+    # finally, observed, to determine the shape of the variable. Because dims can be
+    # specified on the fly, we need a two-step process where we first create the RV
+    # without dims information and then resize it.
+    if not size_or_shape and observed is not None:
+        kwargs = kwargs.copy()
```
michaelosthege: Because it's `**kwargs` in the signature, there's no need to `.copy()`.
```suggestion
```

pymc/distributions/shape_utils.py
```diff
@@ -490,85 +465,50 @@ def convert_size(size: Size) -> Optional[StrongSize]:
...
     if unknowndim_resize_dims:
         raise KeyError(
             f"Dimensions {unknowndim_resize_dims} are unknown to the model and cannot be used to specify a `size`."
         )
 
     # The numeric/symbolic resize tuple can be created using model.RV_dim_lengths
-    resize_shape: Tuple[Variable, ...] = tuple(
-        model.dim_lengths[dname] for dname in sdims[:ndim_resize]
+    return tuple(
+        model.dim_lengths[dname] if dname in model.dim_lengths else shape_implied[i]
```
michaelosthege, line 502:
```suggestion
        model.dim_lengths.get(dname, shape_implied[i])
```

pymc/tests/test_data_container.py
```diff
@@ -45,7 +45,7 @@ def test_sample(self):
         with pm.Model():
             x_shared = pm.MutableData("x_shared", x)
             b = pm.Normal("b", 0.0, 10.0)
-            pm.Normal("obs", b * x_shared, np.sqrt(1e-2), observed=y)
+            pm.Normal("obs", b * x_shared, np.sqrt(1e-2), observed=y, shape=x_shared.shape)
```
michaelosthege, line 48: This change is technically not needed, right?

Because the `mu` parameter here is already shape `(100,)` just like the `y`..

pymc/tests/test_sampling.py
```diff
@@ -1606,7 +1606,7 @@ def test_linear_model(self):
             beta = pm.Normal("beta", 0, 0.1)
             mu = pm.Deterministic("mu", alpha + beta * x)
             sigma = pm.HalfNormal("sigma", 0.1)
-            obs = pm.Normal("obs", mu, sigma, observed=y)
+            obs = pm.Normal("obs", mu, sigma, observed=y, shape=x.shape)
```
michaelosthege, line 1609: Similar to above: Why pass `shape`?

Oh, is this to keep `obs` resizeable?
(See comments here: https://github.com/pymc-devs/pymc/pull/6063/files#r955056055)

pymc/tests/test_shape_handling.py
```diff
@@ -249,69 +247,60 @@ def test_param_and_batch_shape_combos(
...
 
+    def test_simultaneous_dims_and_observed(self):
+        with pm.Model() as pmodel:
+            x = pm.ConstantData("x", [1, 2, 3], dims="ddata")
+            assert "ddata" in pmodel.dim_lengths
+
+            # Note: No checks are performed that observed and dims actually match.
+            y = pm.Normal("y", observed=[0, 0, 0], dims="ddata")
+            assert pmodel.RV_dims["y"] == ("ddata",)
+            assert y.eval().shape == (3,)
```
michaelosthege, line 302: Don't you want to broadcast in this test too?
```suggestion
            y = pm.Normal("y", observed=[[0, 0, 0]], dims=("dobs", "ddata"))
            assert pmodel.RV_dims["y"] == ("dobs", "ddata")
            assert y.eval().shape == (1, 3)
```

pymc/tests/test_shape_handling.py
michaelosthege: As far as I can tell, this `z` variable doesn't need a `size` kwarg, because the shape is already implied by `mu=y`.
And if `x` was resized, the `observed` shape wouldn't matter and there would be a shape error.

Or did I overlook something?
Because if feels very wrong that observed RVs need the `shape` but unobserved don't..

pymc/distributions/distribution.py
```diff
@@ -258,15 +256,21 @@ def __new__(
...
             raise TypeError(f"Name needs to be a string but got: {name}")
 
-        # Create the RV and process dims and observed to determine
-        # a shape by which the created RV may need to be resized.
-        rv_out, dims, observed, resize_shape = _make_rv_and_resize_shape(
+        dims = convert_dims(dims)
+        if observed is not None:
+            observed = convert_observed_data(observed)
+
+        # Create the RV, without taking into consideration `dims`
```
```suggestion
        # Create the RV, without taking `dims` into consideration
```

pymc/distributions/shape_utils.py
```diff
@@ -490,85 +465,50 @@ def convert_size(size: Size) -> Optional[StrongSize]:
...
     dims : tuple of (str or None)
-        Names or None for all dimensions after resizing.
+        Names or None for all RV dimensions.
     """
-    if Ellipsis in dims:
-        # Auto-complete the dims tuple to the full length.
-        # We don't have a way to know the names of implied
-        # dimensions, so they will be `None`.
-        dims = (*dims[:-1], *[None] * ndim_implied)
     sdims = cast(StrongDims, dims)
```
michaelosthege: Signature violation: The previous `dims: WeakDims` parameter was already a sequence, but with `dims: Dims` the new signature allows for `dims="bla"`.
This is not accounted for in the implementation; the `cast(StrongDims, dims)` is not safe yet.

pymc/distributions/distribution.py
```diff
@@ -150,35 +149,34 @@ def fn(*args, **kwargs):
...
-            resize_shape, dims = resize_from_dims(dims, ndim_actual, model)
-        elif Ellipsis in dims:
-            # Replace ... with None entries to match the actual dimensionality.
-            dims = (*dims[:-1], *[None] * ndim_actual)[:ndim_actual]
-    elif observed is not None:
-        resize_shape, observed = resize_from_observed(observed, ndim_actual)
-    return rv_out, dims, observed, resize_shape
+
+    if not size_or_shape and dims is not None:
+        resize_shape_from_dims = shape_from_dims(dims, tuple(rv_out.shape), model)
```
michaelosthege, line 176: This call has `dims: StrongDims` which means that `shape_from_dims` can be typed to require `dims: StrongDims`

pymc/distributions/distribution.py
```diff
@@ -460,19 +452,21 @@ def __new__(
...
         )
 
-        if resize_shape:
-            # A batch size was specified through `dims`, or implied by `observed`.
-            rv_out = cls.change_size(
-                rv=rv_out,
-                new_size=resize_shape,
-                expand=True,
+        # Resize variable based on `dims` information
+        if resize_shape_from_dims:
```
michaelosthege, line 470: Why is this not covered?

pymc/tests/test_shape_handling.py
ricardoV94: Observed variables don't need the shape argument, unless you want to tell PyMC something that it would otherwise not be able to know: i.e., it's shape is the same as another variable or one of it's inputs.

Otherwise you would incorrectly miss these cases:

```python
def test_broadcast_by_observed(self):
        with pm.Model() as m:
            y = pm.Normal("y")
            x = pm.Normal("x", mu=y, observed=np.zeros((3,)))
            assert x.eval().shape == (3,)
```
You can't just ignore the shape of observed. Notice that  all this is just a shortcut so the user doesn't have to specify shape when we have observed. And the correct thing is to just replace `shape=observed.shape`... unless the user gives us shape directly. 

We can't safely know where the source of the batched shape is coming from otherwise.

pymc/distributions/distribution.py
ricardoV94: Thanks 👍

pymc/distributions/shape_utils.py
ricardoV94: Can't do that because dims may extend beyond the implied shape and the fallback expression in `get` is eagerly evaluated. You would get an IndexError

pymc/tests/test_data_container.py
ricardoV94: It's needed for the same reason as above. Observed can broadcast implied dims and not the other way around. It could be that mu has shape 1.

ricardoV94: Only needed because we want to resize it indirectly through x later.

pymc/distributions/distribution.py
ricardoV94: We didn't have a case with SymbolicDistributions. I am removing this experimental Class in the follow up PR I already opened anyway, so I didn't want to add nearly duplicated tests just for this.

pymc/tests/test_sampling.py
ricardoV94: It's just so that we don't get an extra graph dependency and do not have to change the tests expectations. I could equally have kept the original graph and add the new expected variable in the checks below. This was faster.

pymc/tests/test_shape_handling.py
michaelosthege: I understand the example,  but in multiple of these changes in the tests I don't see why it's necessary.

Is this the behavioral change because you are resizing the RV to the observed shape even if it already matches?
And this unlinks it from the symbolic shape that is implied by the parameters?

I'm not sure if that's a great idea..  This is a non-obvious need to specify shape in order **to turn off** the smart behavior of broadcasting to observed shape..

It would be more consistent to **not** automatically broadcast to the observed shape, but require to specify shape/size in these cases.
This would also mean that the shape is inherited from parameters just like with unobserved variables.

pymc/distributions/distribution.py
michaelosthege: Ah I didn't notice that this was in `SymbolicDistribution`

pymc/tests/test_shape_handling.py
ricardoV94: > It would be more consistent to not automatically broadcast to the observed shape, but require to specify shape/size in these cases.
This would also mean that the shape is inherited from parameters just like with unobserved variables.

Observed is by definition the shape of the variable, wheres implied dimensions is not. I understand it's not as neat as what we had but the problem is that in Aesara we don't know if parameters are going to have to be broadcasted or not many times.

ricardoV94:
@michaelosthege Do you see a difference between broadcasting by dims and by observed? Because all the same changes would have been needed if the tests had used dims instead of observed and then assumed you could just resize a variable indirectly by resizing it's arguments.

pymc/tests/test_shape_handling.py
michaelosthege: Taken together with the other breaks I think we should do a major version bump then.
This started as a fix of a rare edge case,  but now it is a breaking change that effects potentially many lines of user code.

michaelosthege: Please edit the PR description to mention this new need for explicit shape/size!

michaelosthege:
> @michaelosthege Do you see a difference between broadcasting by dims and by observed? Because all the same changes would have been needed if the tests had used dims instead of observed and then assumed you could just resize a variable indirectly by resizing it's arguments.

I see you point, but I think for most people the intuition is that shape, size and dims are an input of the RV, whereas the observed is something that the RV points to.
That's also how we draw it in `pm.model_to_graphviz` and it resembles the direction of the generative process.

That's probably why I don't like to broadcast automatically based on `observed`. (I found that strange and complicated when I worked on these implementations in the past..)

ricardoV94:
The graphviz is a bit strange because we use 2 nodes (RV and Value) to represent what is done by a single one for unobserved RVs.

Anyway, I think:
1) The output of `test_broadcast_by_observed` would be very surprising before this PR. Also would be very surprising if we ignored extra dimensions implied by observed (which we don't)
2) We should make it clear (in the dimensionality notebook?) that in the generative graph, observed is used as a shortcut for `shape=observed.shape` unless dims or shape is provided. I hope this provides a good mental picture so users don't expect indirect resizing by default.
3) We should make it clear that the shape argument is more powerful than before. It can take arbitrary symbolic expressions. Hopefully it will become intuitive to specify `shape=mu.shape` or whatever as we start adding it to official examples.

I agree it's more cumbersome, but it does push towards being more explicit about shapes, which we are already doing anyway when we suggest users use dims.

I agree with making it a ~~minor~~ bigger (what's the name for intermediate?) release.

pymc/tests/test_shape_handling.py
ricardoV94: > Please edit the PR description to mention this new need for explicit shape/size!

Added this to major/breaking changes:
- When shape is not provided, dims or observed are used to define the shape of a variable. If you want a variable that has dims or is observed to automatically resize when its inputs change, you must specify that explicitly via the shape argument. For example, `pm.Normal("likelihood", mu=mu, sigma=sigma, observed=data, dims="data", shape=mu.shape)`

pymc/distributions/distribution.py
ricardoV94: done

pymc/distributions/shape_utils.py
ricardoV94: Removed the cast, now that the signature of `shape_from_dims` requires `StrongDims`

pymc/tests/test_shape_handling.py
ricardoV94: Not sure, is this not already covered by `test_define_dims_on_the_fly_from_observed`? If it is the same I would like to keep it separate, in case we decide to drop support for dims on the fly.

michaelosthege (approved):
I can't review super-thoroughly right now, but I trust that @ricardoV94 addressed my concerns.

I'm still not super happy about prioritizing `observed.shape` over the implied dimensionality, but we'll see how the users take it.
I know enough about PyMC so my own models don't blow up because of it ¯\_(ツ)_/¯

PR #6067, opened 2022-08-26
Clean up some warnings from the test suite

Armavica:
**What is this PR about?**

This PR removes 200 warnings from the 678 generated by the test suite. This de-clutters the output of the tests and makes it easier to spot if something is wrong. Also, but I am not sure if this is significant, the whole test suite ran 10% faster on my computer after the cleanup. I will check if it also makes the github runners faster.

I think that the ones that I fixed are the less controversial ones.

A good proportion if not the majority of the remaining warnings comes from a numpy deprecation in aesara which I will also address.

The rest are less easy to fix, or potential legit indications of problems (when I was not sure, I left them alone).

If there is interest, I can make a second pass later and try to tackle the rest.

I can breakup this huge commit into smaller thematic ones, let me know if this would be easier to review.

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Clean up some warnings generated by the test suite

### DIFF
--- a/pymc/tests/test_idata_conversion.py
+++ b/pymc/tests/test_idata_conversion.py
@@ -1,4 +1,5 @@
 # pylint: disable=no-member, invalid-name, redefined-outer-name, protected-access, too-many-public-methods
+import warnings
 
 from typing import Dict, Tuple
 
@@ -18,6 +19,7 @@ from pymc.backends.arviz import (
     predictions_to_inference_data,
     to_inference_data,
 )
+from pymc.exceptions import ImputationWarning
 
 
 @pytest.fixture(scope="module")
@@ -204,9 +206,13 @@ class TestDataPyMC:
 
     def test_posterior_predictive_warning(self, data, eight_schools_params, caplog):
         with data.model:
-            posterior_predictive = pm.sample_posterior_predictive(
-                data.obj, 370, return_inferencedata=False, keep_size=False
-            )
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore", ".*smaller than nchains times ndraws.*", UserWarning
+                )
+                posterior_predictive = pm.sample_posterior_predictive(
+                    data.obj, 370, return_inferencedata=False, keep_size=False
+                )
             with pytest.warns(UserWarning, match="shape of variables"):
                 inference_data = to_inference_data(
                     trace=data.obj,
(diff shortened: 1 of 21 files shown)


Armavica:
I am not sure why this test is failing on the runner, as it is passing on my computer.

canyon289:
Thank you for the PR. Its because this one test is too sensitive. Feel free to double the relative tolerance to make it less flaky

image

Armavica:
I see. Shouldn't that better be fixed by fixing the PRNG seed?

pymc/tests/test_idata_conversion.py
```diff
@@ -320,7 +332,9 @@ def test_missing_data_model(self):
         model = pm.Model()
         with model:
             x = pm.Normal("x", 1, 1)
-            y = pm.Normal("y", x, 1, observed=data)
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", ImputationWarning)
```
ricardoV94: Instead of ignoring ImputationWarnings, could we confirm they are happening?

pymc/tests/test_quadpotential.py
```diff
@@ -206,7 +209,9 @@ def test_full_adapt_sample_p(seed=4566):
     )
 
     n_samples = 1000
-    pot = quadpotential.QuadPotentialFullAdapt(2, np.zeros(2), m_inv, 1)
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", ".*experimental feature.*", UserWarning)
```
ricardoV94: I think we should also confirm this one is being issued

pymc/tests/test_sampling_jax.py
```diff
@@ -16,17 +16,19 @@
...
-    _numpyro_nuts_defaults,
-    _replace_shared_variables,
-    _update_numpyro_nuts_kwargs,
-    get_jaxified_graph,
-    get_jaxified_logp,
-    sample_blackjax_nuts,
-    sample_numpyro_nuts,
-)
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", ".*module is experimental.*", UserWarning)
```
ricardoV94: Also should be confirmed

pymc/tests/test_variational_inference.py
```diff
@@ -526,7 +526,9 @@ def init_(**kw):
 @pytest.fixture(scope="function")
 def inference(inference_spec, simple_model):
     with simple_model:
-        return inference_spec()
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", ".*experimental inference Operator.*", UserWarning)
```
ricardoV94: Same here

ricardoV94 (changes requested):
This is amazing. There are just some cases where I think we should assert the warning is being issued instead of ignoring it.

ricardoV94:
> I see. Shouldn't that better be fixed by fixing the PRNG seed?

I never saw this test fail so I think it's safe to just ignore and rerun.

ricardoV94:
Opened an issue for the failing test: https://github.com/pymc-devs/pymc/issues/6069

Armavica:
Thank you for your review! I pushed the requested changes.

ricardoV94 (approved):

PR #6068, opened 2022-08-26
Rename cov_func/cov to scale_func/scale for TP/MvStudentT

### LINKED ISSUE #2865: Cov arguments in TP and MvStudentT
I am migrating this discussion from [Discourse](https://discourse.pymc.io/t/marginal-student-t-process-and-misleading-parameters/867):

> ... It seems like `cov_func` and `cov` are misleading parameter names, since they aren’t actually the covariances of the distributions. Rather, they refer to the `Sigma` parameter of `MvStudentT`, which is related to the covariance by `nu * Sigma / (nu-2)`. Thus, [here](http://docs.pymc.io/notebooks/GP-TProcess.html#Samples-from-a-TP-prior), for example, it looks like apples and oranges are being compared when using the same `cov_func` in a `GP` and `TP`. Can these arguments be changed so that they really are covariances, or at least be better documented so that people know that they aren’t actually what they seem to be?

What is the correct course of action here? Better documentation or changing the functionality to align with the argument names?

fonnesbeck:
The scale matrix/function in the `MvStudentT` distribution and the `TP` have to date been called `cov` and `cov_func`, respectively. This is confusing because these are not covariance matrices/functions, but rather scale matrices/functions that only become covariances as a function of the `nu` parameter. This PR renames `cov` and `cov_func` to `scale` and `scale_func`, respectively, but still allows `cov` and `cov_func` for backward compatibility, with a deprecation warning.

Closes #2865 

**Checklist**

## Bugfixes / New features
- fixes #2865

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -368,13 +368,13 @@ class MvStudentT(Continuous):
 
     @classmethod
     def dist(cls, nu, Sigma=None, mu=None, scale=None, tau=None, chol=None, lower=True, **kwargs):
-        if kwargs.get('cov') is not None:
+        if kwargs.get("cov") is not None:
             warnings.warn(
                 "Use the scale argument to specify the scale matrix."
                 "cov will be removed in future versions.",
                 DeprecationWarning,
             )
-            scale = kwargs.get('cov')
+            scale = kwargs.get("cov")
         if Sigma is not None:
             if scale is not None:
                 raise ValueError("Specify only one of scale and Sigma")
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -352,42 +352,49 @@ class MvStudentT(Continuous):
     nu : tensor_like of float
         Degrees of freedom, should be a positive scalar.
     Sigma : tensor_like of float, optional
-        Covariance matrix. Use `cov` in new code.
+        Scale matrix. Use `scale` in new code.
     mu : tensor_like of float, optional
         Vector of means.
-    cov : tensor_like of float, optional
-        The covariance matrix.
+    scale : tensor_like of float, optional
+        The scale matrix.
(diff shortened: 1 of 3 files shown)


pymc/distributions/multivariate.py
```diff
@@ -352,42 +352,49 @@ class MvStudentT(Continuous):
...
-        The cholesky factor of the covariance matrix.
+        The cholesky factor of the scale matrix.
     lower : bool, default=True
         Whether the cholesky fatcor is given as a lower triangular matrix.
     """
     rv_op = mv_studentt
 
     @classmethod
-    def dist(cls, nu, Sigma=None, mu=None, cov=None, tau=None, chol=None, lower=True, **kwargs):
+    def dist(cls, nu, Sigma=None, mu=None, scale=None, tau=None, chol=None, lower=True, **kwargs):
```
canyon289, line 370: I know this isnt part of code change but just curious. Why is Sigma capitalized?

canyon289 (approved):

pymc/distributions/multivariate.py
fonnesbeck: Just a convention related to the fact that its a matrix rather than a scalar.

pymc/distributions/multivariate.py
```diff
@@ -352,42 +352,49 @@ class MvStudentT(Continuous):
...
+        if kwargs.get("cov") is not None:
+            warnings.warn(
+                "Use the scale argument to specify the scale matrix."
+                "cov will be removed in future versions.",
+                DeprecationWarning,
```
ricardoV94: I remember FutureWarning being the recommended one, but I am unsure.

```suggestion
                FutureWarning,
```

ricardoV94 (approved):

bwengals (approved):

PR #6072, opened 2022-08-26
Implement Symbolic RVs

### LINKED ISSUE #5533: Reimplement nested Mixtures

ricardoV94:
This PR wraps the graphs returned by the old `SymbolicDistribution.rv_op` in a `SymbolicRV` `OpFromGraph` `Op` so that they can be manipulated like vanilla `RandomVariable`s. The main benefit is that we can now:
1. Change their size, which is necessary for creating graphs with nested `SymbolicRV`s (nested mixtures, censored mixtures, AR with censored init_dist, you name it)
2.  Dispatch moment functions, which we noted was necessary in #5762

When these `OpFromGraph` don't have a specialized logprob dispatched, they can be inlined inside Aeppl, so that their logprob can be derived dynamically. This was one of the main reasons for creating `SymbolicDistribution`s in the first place.

Closes #5533 

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- Implemented general univariate RandomWalk distribution factory
- Distribution factories like Mixture, Censored, RandomWalk, and AR can now be composed together
- Reintroduced support for nested Mixtures

## Docs / Maintenance
- Implemented `SymbolicRV`s to encapsulate graphs from distribution factories
- Removed temporary `SymbolicDistribution` class in favor of `Distribution`
- `change_rv_size` was removed in favor of the more general `change_dist_size`
- `_LKJCholeskyCov` and `GaussianRandomWalk` are now `SymbolicRV`s


pymc/distributions/censored.py
```diff
@@ -100,26 +107,23 @@ def rv_op(cls, dist, lower=None, upper=None, size=None):
...
-        rv_out.tag.lower = lower
-        rv_out.tag.upper = upper
+        dist_, lower_, upper_ = dist.type(), lower.type(), upper.type()
+        censored_rv_ = at.clip(dist_, lower_, upper_)
 
-        return rv_out
+        return CensoredRV(
+            inputs=[dist_, lower_, upper_],
+            outputs=[censored_rv_],
+            ndim_supp=0,
```
lucianopaz, line 119: Is it necessary to restrict the ndim_supp only to scalar distributions?

pymc/distributions/censored.py
```diff
@@ -100,26 +107,23 @@ def rv_op(cls, dist, lower=None, upper=None, size=None):
...
-        dist = rv.tag.dist
-        lower = rv.tag.lower
-        upper = rv.tag.upper
+        dist, lower, upper = rv.owner.inputs
         new_dist = change_rv_size(dist, new_size, expand=expand)
         return cls.rv_op(new_dist, lower, upper)
 
 
-@_moment.register(Clip)
+@_moment.register(CensoredRV)
```
lucianopaz, line 131: This is so much cleaner now!

pymc/distributions/distribution.py
```diff
@@ -464,7 +468,7 @@ def __new__(
         # Resize variable based on `dims` information
         if resize_shape_from_dims:
             resize_size_from_dims = find_size(
-                shape=resize_shape_from_dims, size=None, ndim_supp=rv_out.tag.ndim_supp
+                shape=resize_shape_from_dims, size=None, ndim_supp=rv_out.owner.op.ndim_supp
```
lucianopaz: Why was this tag before??

pymc/distributions/distribution.py
```diff
@@ -542,15 +546,45 @@ def dist(
...
+    op = node.op
+    # If `Op` is not a `MeasurableVariable`, Aeppl must derive its logp from the inner graph
+    if not isinstance(op, MeasurableVariable):
+        return clone_replace(op.inner_outputs, {u: v for u, v in zip(op.inner_inputs, node.inputs)})
+
+
+# Registered before pre-canonicalization which happens at position=-10
+logprob_rewrites_db.register(
+    "inline_SymbolicRandomVariable",
+    in2out(inline_SymbolicRandomVariable),
```
lucianopaz: What does in2out do?

pymc/distributions/distribution.py
```diff
@@ -542,15 +546,45 @@ def dist(
...
+    "basic",
+    position=-20,
```
lucianopaz, line 439: I don’t really understand what the rewrite_db arguments mean. Could you explain these to me?

pymc/distributions/distribution.py
```diff
@@ -542,15 +546,45 @@ def dist(
...
 
 
+class SymbolicRandomVariable(OpFromGraph):
+    """TODO"""
+
+    ndim_supp = None
+
+    def __init__(self, *args, ndim_supp, **kwargs):
+        self.ndim_supp = ndim_supp
+        kwargs.setdefault("inline", True)
```
lucianopaz: What happens if inline is false?

pymc/distributions/distribution.py
```diff
@@ -861,3 +865,77 @@ def default_moment(rv, size, *rv_inputs, rv_name=None, has_fallback=False, ndim_
             f"Please provide a moment function when instantiating the {rv_name} "
             "random variable."
         )
+
+
+def change_dist_size(
```
lucianopaz: Shouldn’t this live on shape_utils?

pymc/distributions/distribution.py
```diff
@@ -861,3 +865,77 @@ def default_moment(rv, size, *rv_inputs, rv_name=None, has_fallback=False, ndim_
...
+    """Change or expand the size of a Distribution.
+
+    Parameters
+    ==========
+    dist
+        The old Distribution output.
+    new_size
+        The new size.
+    expand:
+        Expand the existing size by `new_size`.
```
lucianopaz: I can’t really understand the docstring for this. Could you add an example that shows what expands will do? I had to read the source code to understand it

pymc/distributions/distribution.py
```diff
@@ -861,3 +865,77 @@ def default_moment(rv, size, *rv_inputs, rv_name=None, has_fallback=False, ndim_
...
+
+    # Update "traditional" rng default_update, if that was set for old RV
+    default_update = getattr(rng, "default_update", None)
+    if default_update is not None and default_update is rv_node.outputs[0]:
+        rng.default_update = new_rv_node.outputs[0]
+
+    return new_rv
+
+
+@_change_dist_size.register(SpecifyShape)
```
lucianopaz: Should this also dispatch asserts and prints?

lucianopaz (commented):
Very nice work Ricardo! I skimmed through things and left a bunch of comments. Once you think this isn’ta draft anymore I’ll do a more thorough review of the tests. Did anything change in aeppl or aesara to enable this?

pymc/distributions/censored.py
ricardoV94: It's not, but for now we only have logcdf for scalar distributions, and aeppl and pymc check the `dist` is scalar so I just hard-coded it here.

pymc/distributions/distribution.py
ricardoV94: It was just a temporary work-around  when cleaning #6063

pymc/distributions/distribution.py
ricardoV94: It's the simplest (maybe?) form of applying a rewrite, it just traverses a graph from inputs to outputs and tries to apply the rewrite in nodes that match the `tracks` Op

pymc/distributions/distribution.py
ricardoV94: I can add kwargs to make it more readable. It's just name, rewrite, tag (used for manually including or excluding rewrites), and order in which rewrite is applied from all the rewrites in the database

pymc/distributions/distribution.py
ricardoV94: When inline is False the inner code of an OpFromGraph remains as a black-box, so compilation does not mix (or optimize) across the outer and inner nodes. This only matters for the generative/random functions, because in the logp functions, the whole node is replaced by the logp expression instead.

pymc/distributions/distribution.py
ricardoV94: Probably

pymc/distributions/distribution.py
ricardoV94: Sure

pymc/distributions/distribution.py
ricardoV94: I never used prints so I don't know. 

I am not sure about asserts, what if they are related to shapes of inputs/outputs? We can't easily know what they are about. If a variable requires Asserts to work properly, I would suggest they are added in the inner graph of a SymbolicRV. The custom `_change_dist_size` dispatch function for that SymbolicRV will then know how to modify/recreate the asserts needed in the resized variable.

pymc/distributions/distribution.py
ricardoV94: I should also add a note that the original RV should not be used anymore, as the returned one might use the same RandomState/Generator variable as the original one.

ricardoV94:
> Once you think this isn’t a draft anymore I’ll do a more thorough review of the tests. 

I'll re-request your review then.

> Did anything change in aeppl or aesara to enable this?

Nope. It's all PyMC insfrastructure.

pymc/distributions/distribution.py
```diff
@@ -150,35 +155,34 @@ def fn(*args, **kwargs):
...
-        resize_shape, observed = resize_from_observed(observed, ndim_actual)
-    return rv_out, dims, observed, resize_shape
+
+    if not size_or_shape and dims is not None:
+        resize_shape_from_dims = shape_from_dims(dims, tuple(rv_out.shape), model)
+
+    return rv_out, resize_shape_from_dims
 
 
 class Distribution(metaclass=DistributionMeta):
```
larryshamalama, line 232: Unrelated to this PR: what does the class attribute `rv_class` do?

pymc/distributions/distribution.py
```diff
@@ -343,223 +350,49 @@ def dist(
...
+@node_rewriter([SymbolicRandomVariable])
+def inline_SymbolicRandomVariable(fgraph, node):
+    """
+    This optimization expands the internal graph of a SymbolicRV when obtaining logp
+    from Aeppl, in the case it does not have a specialized logp associated with it.
+    """
+    op = node.op
+    # If `Op` is not a `MeasurableVariable`, Aeppl must derive its logp from the inner graph
+    if not isinstance(op, MeasurableVariable):
+        return clone_replace(op.inner_outputs, {u: v for u, v in zip(op.inner_inputs, node.inputs)})
```
larryshamalama, line 431: Two questions here:

1. My understanding is that, if `op` is an instance of a `MeasurableVariable`, it will (and should) have a `logp` dispatched. Will it be of any value of have an `else` statement to check that this is effectively the case in `_logprob.registry`?
2. What does the `clone_replace` here do? Does it emulate what `OpFromGraph` would do, i.e. allowing the log-probability can extract the `logp` of outputs? Can you provide an example of such a case?

pymc/distributions/mixture.py
```diff
@@ -189,7 +195,7 @@ def dist(cls, w, comp_dists, **kwargs):
             # TODO: Allow these to not be a RandomVariable as long as we can call `ndim_supp` on them
             #  and resize them
             if not isinstance(dist, TensorVariable) or not isinstance(
-                dist.owner.op, RandomVariable
+                dist.owner.op, (RandomVariable, SymbolicRandomVariable)
```
larryshamalama, line 194: Thinking out loud: do we get any identifiability problems with nested mixtures? How does posterior sampling perform when a mixture component is a mixture itself? My guess is that things will work out well. Basically, given that a mixture of mixtures is a mixture itself, I'm wondering if there are any computational pitfalls with this approach, i.e. not canonicalizing a graph of nested mixtures.

Perhaps it would a `node_rewriter` of nested mixtures into a single `MarginalMixtureRV` would be an interesting avenue to pursue later...

pymc/printing.py
```diff
@@ -31,40 +35,62 @@
 ]
```
larryshamalama, line 37: I'm guessing that this would close #5793 😅

pymc/tests/test_sampling.py
```diff
@@ -1606,7 +1606,7 @@ def test_linear_model(self):
             beta = pm.Normal("beta", 0, 0.1)
             mu = pm.Deterministic("mu", alpha + beta * x)
             sigma = pm.HalfNormal("sigma", 0.1)
-            obs = pm.Normal("obs", mu, sigma, observed=y)
+            obs = pm.Normal("obs", mu, sigma, observed=y, shape=x.shape)
```
larryshamalama: Not super familiar with shape inference. Why is it necessary to provide a `shape` argument here?

larryshamalama (commented):
This is amazing @ricardoV94! 🙂 I feel like having an explicit `SymbolicRandomVariable` class would make the code structure of symbolic distributions much more clear. 

Would #5931 need to be first merged before this PR?

pymc/distributions/distribution.py
ricardoV94: Nothing?

pymc/distributions/distribution.py
ricardoV94: Actually cannot use kwargs for this one.

pymc/distributions/distribution.py
ricardoV94: Removed it

pymc/distributions/distribution.py
ricardoV94: Done

pymc/distributions/distribution.py
ricardoV94: Updated the docstrings, what do you think?

pymc/distributions/distribution.py
ricardoV94: 1. We can
2. This rewrite just does the same as the one in Aesara when OpFromGraph is inlined. This basically includes the inner graph in the original one, so that instead of seeing `CensoredRV`, Aeppl will see the inner `Clip(...)`. That way we can use Aeppl to derive it's logp, even though in PyMC we handle it inside a SymbolicRV
 
https://github.com/aesara-devs/aesara/blob/63f52536a09b13cc55481f071ccb4fc59c9740f4/aesara/compile/builders.py#L932

pymc/distributions/mixture.py
ricardoV94: AFAIK there is nothing special about nested mixtures that is not present in flat mixtures. The same switching issues can emerge in both cases if the components are equivalent a priori.

pymc/printing.py
ricardoV94: Yeah, that's another benefit of using SymbolicRVs

pymc/tests/test_sampling.py
ricardoV94: This is part of the previous PR #6063. You can read the discussion there.

pymc/distributions/distribution.py
ricardoV94: Added a check and test

ricardoV94:
> Would https://github.com/pymc-devs/pymc/pull/5931 need to be first merged before this PR?

I don't think so

pymc/distributions/distribution.py
ricardoV94: Ended up removing the else branch, otherwise it fails because of the cases where we make the inputs `Unmeasurable`...

ricardoV94:
The failing `test_profile_variable` is odd, and looks unrelated.

pymc/distributions/distribution.py
```diff
@@ -355,205 +362,58 @@ def dist(
...
```
lucianopaz, line 431: Could clone_replace be dangerous? It might be cloning the input nodes and end up breaking the conditional dependence between input variables and outputs of the cloned graph

ricardoV94: I am just copying what the Aesara rewrite does so I assume it's safe: 

https://github.com/aesara-devs/aesara/blob/d7fb940264a392857781b075e1b8178c603001a2/aesara/compile/builders.py#L932

ricardoV94:
This PR has grown considerably in scope as I attempted to patch issues with `LKJCholeskyCov` and `GaussianRandomWalk`. Those are now `SymbolicRandomVariable`s as well, which makes sense as they are distribution factories (i.e., they must be composed with `init_dist`/`sd_dist`).

The nice thing to come out of this is that now we also have support for arbitrary univariate random walks (support for multivariate will be added later), as that was simpler than just converting the `GaussianRandomWalk`

Most likely some tests will still fail, and I have marked the PR has a draft, since I need to add docstrings and tests for the new `RandomWalk`

ricardoV94:
Tests are finally passing (except RTD).

I still have to add docstrings for `RandomWalk` and some tests as well, but no changes in functionality. So feel free to give another round of review :)

pymc/aesaraf.py
```diff
@@ -875,6 +815,53 @@ def largest_common_dtype(tensors):
     return np.stack([np.ones((), dtype=dtype) for dtype in dtypes]).dtype
 
 
+@_logprob.register(MeasurableJoin)
+def logprob_join_constant_shapes(op, values, axis, *base_vars, **kwargs):
+    """Compute the log-likelihood graph for a `Join`.
+    This overrides the implementation in Aeppl, to constant fold the shapes
+    of the base vars so that RandomVariables do not show up in the logp graph,
+    which is a requirement enforced by `pymc.distributions.logprob.joint_logp`
+    """
```
michaelosthege: This function could use some comments that explain why the steps are about.

For example, I have never seen `at.switch`, so a priori I have no clue what this does..

pymc/distributions/distribution.py
```diff
@@ -355,205 +363,66 @@ def dist(
...
-            transform=transform,
-            initval=initval,
-        )
-        # add in pretty-printing support
-        rv_out.str_repr = types.MethodType(str_for_symbolic_dist, rv_out)
-        rv_out._repr_latex_ = types.MethodType(
-            functools.partial(str_for_symbolic_dist, formatting="latex"), rv_out
-        )
+    ndim_supp: int = None
+    inline_aeppl: bool = False
```
```suggestion
    inline_aeppl: bool = False
    """Specifies whether the logprob function is derived automatically by introspection of the inner graph.
    
    If `False`, a logprob function must be dispatched directly to the subclass.
    """
```

It would help to have such docstrings on the other attributes too..

pymc/distributions/distribution.py
```diff
@@ -355,205 +363,66 @@ def dist(
...
 
-        Parameters
-        ----------
-        dist_params : array-like
-            The inputs to the `RandomVariable` `Op`.
-        shape : int, tuple, Variable, optional
-            A tuple of sizes for each dimension of the new RV.
-        size : int, tuple, Variable, optional
-            For creating the RV like in Aesara/NumPy.
+MeasurableVariable.register(SymbolicRandomVariable)
```
michaelosthege, line 414:
```suggestion
# Let Aeppl know that the SymbolicRandomVariable has a logprob.
MeasurableVariable.register(SymbolicRandomVariable)
```

pymc/distributions/shape_utils.py
```diff
@@ -541,3 +552,101 @@ def find_size(
...
+    raise NotImplementedError(
+        f"Variable {dist} of type {op} has no _change_dist_size implementation."
+    )
+
+
+def change_dist_size(
+    dist: TensorVariable,
+    new_size: PotentialShapeType,
+    expand: bool = False,
+):
```
michaelosthege:
```suggestion
) -> TensorVariable:
```

pymc/distributions/shape_utils.py
```diff
@@ -541,3 +552,101 @@ def find_size(
...
+        new_dist.tag.__dict__.setdefault(k, v)
+
+    if config.compute_test_value != "off":
+        compute_test_value(new_dist)
+
+    return new_dist
+
+
+@_change_dist_size.register(RandomVariable)
+def change_rv_size(op, rv, new_size, expand):
```
michaelosthege:
```suggestion
def change_rv_size(op, rv, new_size, expand) -> TensorVariable:
```

pymc/tests/test_distributions.py
```diff
@@ -3319,9 +3305,9 @@ def test_dist_broadcasted_by_lower_upper(self):
 
 class TestLKJCholeskCov:
     def test_dist(self):
-        sd_dist = pm.Exponential.dist(1, size=(10, 3))
-        x = pm.LKJCholeskyCov.dist(n=3, eta=1, sd_dist=sd_dist, size=10, compute_corr=False)
-        assert x.eval().shape == (10, 6)
+        # sd_dist = pm.Exponential.dist(1, size=(10, 3))
+        # x = pm.LKJCholeskyCov.dist(n=3, eta=1, sd_dist=sd_dist, size=10, compute_corr=False)
+        # assert x.eval().shape == (10, 6)
```
michaelosthege: ToDo

pymc/tests/test_distributions_moments.py
```diff
@@ -98,8 +98,17 @@ def test_all_distributions_have_moments():
         dist_module.simulator.Simulator,
     }
 
+    # Ignore distributions of SymbolicRVs
+    missing_moments -= {
+        dist_module.censored.Censored,
+        dist_module.mixture.Mixture,
+        dist_module.timeseries.RandomWalk,
+        dist_module.timeseries.AR,
+    }
```
michaelosthege: Is it then no longer possible to get the moment of a `RandomWalk`??

And `GaussianRandomWalk` is covered by `RandomWalk`?

pymc/distributions/timeseries.py
```diff
@@ -114,109 +116,131 @@ def get_steps(
...
+        steps = at.as_tensor_variable(intX(steps))
 
-        Returns
-        -------
-        ndarray
-        """
+        if not (
+            isinstance(init_dist, at.TensorVariable)
+            and init_dist.owner is not None
+            and isinstance(init_dist.owner.op, (RandomVariable, SymbolicRandomVariable))
```
michaelosthege, line 158: This means that `init_dist` can be a `Mixture`, right?
That's cool!

michaelosthege: Could `init_dist` be a list of init distributions (with `len(init_dist) == steps`)?

I mean not right now, of course, but in principle?

pymc/distributions/timeseries.py
```diff
@@ -278,57 +294,20 @@ def dist(cls, mu=0.0, sigma=1.0, *, init_dist=None, steps=None, **kwargs) -> at.
                 UserWarning,
             )
             init_dist = Normal.dist(0, 100)
```
michaelosthege, line 298: Shouldn't this be `Normal.dist(mu, sigma)`?

pymc/distributions/timeseries.py
```diff
@@ -114,109 +116,131 @@ def get_steps(
...
-        return np.cumsum(grw, axis=-1)
-
-
-gaussianrandomwalk = GaussianRandomWalkRV()
+            size = at.broadcast_shape(init_dist, innovation_dist[..., 0])
+        innovation_size = tuple(size) + (steps,)
+
+        # Resize input distributions
+        init_dist = change_dist_size(init_dist, size)
+        innovation_dist = change_dist_size(innovation_dist, innovation_size)
```
michaelosthege, line 189: If this line anticipated `innovation_dist` that's already a vector, one could do random walks that have different innovations in every step, right? (e.g. via `at.stack([Normal.dist(), Uniform.dist(), Beta.dist()])`

ricardoV94: This requires more discussion: https://github.com/pymc-devs/pymc/discussions/6090

pymc/aesaraf.py
ricardoV94: I assume you meant `at.split`? It undoes `at.join`, just like Numpy: https://numpy.org/doc/stable/reference/generated/numpy.split.html

ricardoV94: TODO: Test this function more directly

pymc/distributions/distribution.py
ricardoV94: I already included the explanation on the docstrings above. Is there a standard for explaining attributes, like the `parameters` in numpy docstrings?

pymc/distributions/shape_utils.py
ricardoV94: mypy starts complaining somewhere else IIRC

pymc/distributions/timeseries.py
ricardoV94: > `This means that init_dist can be a Mixture, right?`

Yes

> Could init_dist be a list of init distributions (with len(init_dist) == steps)?

`init_dist` should not match the number of steps, but of batched dimensions. You have one per time series, to specify at which point does it start. Maybe you meant `innovation_dist`? If so, it's the same discussion as in #6090

pymc/distributions/timeseries.py
ricardoV94: No. The drift `mu` does not tell us anything about the initial point. You could easily have a timeseries with the same `mu` that starts at `-100` or at `100`. https://github.com/pymc-devs/pymc/issues/5744

pymc/tests/test_distributions.py
ricardoV94: Thanks, it was during debug :D

pymc/distributions/distribution.py
michaelosthege: I'm not sure if I understand the question correctly.

You can put docstrings on attributes by placing a `""" ... """` below them. That's the standard, it's just not well known..

pymc/distributions/timeseries.py
michaelosthege: Yes that was a typo, I meant `innovation_dist`

pymc/tests/test_distributions_moments.py
ricardoV94: It is, but the way this check is done, it checks for `rv_op` which is a `classmethod` in "SymbolicDistributions" and not the type on which `moment` would be dispatched upon. 

Maybe we should distinguish between `rv_op` and `create_rv_op`, which for vanilla distributions is the same. That would mean we get the `logp`/`logcdf`/`moment` dispatching working automatically for now as well...

ricardoV94: Not doing the automatic dispatching, but added a new attribute

pymc/distributions/distribution.py
ricardoV94: Didn't know. Done

pymc/aesaraf.py
ricardoV94: Added a direct test, hopefully it's obvious what it does with it

pymc/distributions/shape_utils.py
ricardoV94: Done.

ricardoV94:
Officially ready for review and merge if people agree

pymc/aesaraf.py
```diff
@@ -1004,13 +970,20 @@ def compile_pymc(
         if isinstance(random_var.owner.op, RandomVariable):
             rng = random_var.owner.inputs[0]
             if not hasattr(rng, "default_update"):
-                rng_updates[rng] = random_var.owner.outputs[0]
+                update_map = {rng: random_var.owner.outputs[0]}
             else:
-                rng_updates[rng] = rng.default_update
+                update_map = {rng: rng.default_update}
```
junpenglao: Nit:
```
            if hasattr(rng, "default_update"):
                update_map = {rng: rng.default_update}
            else:
                update_map = {rng: random_var.owner.outputs[0]}
```

junpenglao: Or if you want to get fancy:
```
if update_val := getattr(rng, "default_update", None):
    update_map = {rng: update_val}
else:
    update_map = {rng: random_var.owner.outputs[0]}
```

pymc/step_methods/metropolis.py
```diff
@@ -1046,12 +1053,14 @@ def sample_except(limit, excluded):
...
-    [logp0], inarray0 = pm.join_nonshared_inputs(point, [logp], vars, shared)
+    [logp0], inarray0 = join_nonshared_inputs(point, [logp], vars, shared)
 
     tensor_type = inarray0.type
     inarray1 = tensor_type("inarray1")
 
-    logp1 = pm.CallableTensor(logp0)(inarray1)
+    logp1 = CallableTensor(logp0)(inarray1)
+    # Replace any potential duplicated RNG nodes
+    (logp1,) = replace_rng_nodes((logp1,))
```
junpenglao, line 1063: Why is this necessary now but not before?

junpenglao (approved):
Excellent work Ricardo! I like that everything is cleaner now.

pymc/aesaraf.py
ricardoV94: I think walrus is not in our oldest supported Python version

pymc/step_methods/metropolis.py
ricardoV94: It was already necessary but we never noticed. The only logp graphs with RVs we allow are those with Simulators so it's a very experimental edge case

pymc/distributions/distribution.py
```diff
@@ -178,11 +185,55 @@ def _make_rv_and_resize_shape_from_dims(
...
+        return {}
+
+
 class Distribution(metaclass=DistributionMeta):
     """Statistical distribution"""
 
-    rv_class = None
-    rv_op: RandomVariable = None
+    rv_op: [RandomVariable, SymbolicRandomVariable] = None
+    rv_type: MetaType = None
```
larryshamalama, line 236: What is the need for `rv_type`? I must be missing something...

Here is my understanding: It seems that non-symbolic distributions still get an `rv_op` but are assigned an `rv_type` when calling `Distribution`, but symbolic distributions need a `rv_op` classmethod to be manually created *without* (? not sure here) `rv_type`.

pymc/distributions/logprob.py
```diff
@@ -320,3 +324,51 @@ def ignore_logprob(rv: TensorVariable) -> TensorVariable:
         return rv
     new_node = assign_custom_measurable_outputs(node, type_prefix=prefix)
     return new_node.outputs[node.outputs.index(rv)]
+
+
+@_logprob.register(MeasurableJoin)
```
larryshamalama, line 329: Do we expect to have more of these AePPL logprob dispatches to be overwritten? Because symbolic distributions don't have a `logp` method, I was getting confused on how the log-probabilities are being retrieved

pymc/distributions/distribution.py
```diff
@@ -178,11 +185,55 @@ def _make_rv_and_resize_shape_from_dims(
...
+    return SymbolicRandomVariable create their these graphs at runtime via the
+    classmethod `cls.rv_op`, taking care to clone and resize random inputs, if needed.
+    """
+
+    ndim_supp: int = None
+    """Number of support dimensions as in RandomVariables
+    (0 for scalar, 1 for vector, ...)
+     """
+
+    inline_aeppl: bool = False
```
larryshamalama, line 207: Are there symbolic rvs that have `inline_aeppl` to be `True` by default?

larryshamalama (commented):
Asking some more questions 😅

pymc/distributions/distribution.py
ricardoV94: The rv_type is not necessary but it was useful for `test_all_distributions_have_moments`. For traditional distributions `rv_type` is added automatically by the metaclass, it's simply the `type(rv_op)`.

This seems useful to know because it's the type on which all methods are dispatched: logp/ logcdf/ moment/ change_dist_size

pymc/distributions/logprob.py
ricardoV94: We should try to avoid as much duplicated work as possible, so I imagine it won't be that frequent.

What do you mean by symbolic distributions don't have logp methods? All except Censored right now have direct logp methods.

pymc/distributions/distribution.py
ricardoV94: Censored does

larryshamalama: I see, thanks!

pymc/distributions/logprob.py
larryshamalama: Sorry, perhaps that wasn't clear or my terminology was wrong. By `logp` method, I meant that they are defined underneath the Distribution class rather than assigned via single dispatch. The second sentence was more of a comment/concern, maybe it doesn't make too much sense...

ricardoV94: That is a shortcut we did sometime ago where the MetaDistribution will do the dispatching automatically. Some context in this discussion: https://github.com/pymc-devs/pymc/issues/5308

pymc/aesaraf.py
junpenglao: 🔜 then

PR #6073, opened 2022-08-27
Fix incorrect formula in NormalMixture docstring

MatthewQuenneville:
The formula for the variance is incorrect in the docstring of the NormalMixture distribution. The correct expression can be found on the Wikipedia page for a [Mixture Distribution](https://en.wikipedia.org/wiki/Mixture_distribution#Moments), or it follows from the moment expressions in section A.2.2 of [The Matrix Cookbook](https://www.math.uwaterloo.ca/~hwolkowi//matrixcookbook.pdf).

The two expressions are not equivalent, as can be verified by checking the case of a mixture of two normal distributions with equal weights, means, and variances. The resulting mixture distribution should be identical to the two individual components, but the expression would give half the variance of the components as written.

...

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Fix the incorrect variance formula in the NormalMixture distribution docstring

### DIFF
--- a/pymc/distributions/mixture.py
+++ b/pymc/distributions/mixture.py
@@ -510,7 +510,7 @@ class NormalMixture:
     ========  =======================================
     Support   :math:`x \in \mathbb{R}`
     Mean      :math:`\sum_{i = 1}^n w_i \mu_i`
-    Variance  :math:`\sum_{i = 1}^n w_i^2 \sigma^2_i`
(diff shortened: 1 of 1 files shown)


maresb:
I'm not an official reviewer, I can confirm that the math checks out. Nice catch!

michaelosthege (approved):
@maresb confirmed that this is correct.

Thanks for the PR !

PR #6080, opened 2022-08-29
Fix formatting in documentation of AR distribution parameters

daniel-saunders-phil:
**What is this PR about?**

Fixed a variety of docstring issues for the pm.AR distribution.

...

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Fix formatting in documentation of AR distribution parameters

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -372,10 +372,10 @@ class AR(SymbolicDistribution):
     constant : bool, default False
         Whether the first element of rho should be used as a constant term in the AR
         process.
-    init_dist : unnamed distribution
+    init_dist : unnamed distribution, optional
         Scalar or vector distribution for initial values. Unnamed refers to distributions
          created with the ``.dist()`` API. Distributions should have shape (*shape[:-1], ar_order).
-        If not, it will be automatically resized.
+        If not, it will be automatically resized. Defaults to pm.Normal.dist(0, 100, shape=...).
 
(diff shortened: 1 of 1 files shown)


reshamas:
#DataUmbrellaPyMCSprint

cluhmann:
This looks good to me.  The only question I would have is about the "unnamed distribution" which seems correct, but there may be some other "docstring-official" way of specifying this.  Not sure.

cluhmann:
And thank you for your contribution @daniel-saunders-phil !

ricardoV94:
> This looks good to me.  The only question I would have is about the "unnamed distribution" which seems correct, but there may be some other "docstring-official" way of specifying this.  Not sure.

We should make one like tensor_like, but there isn't any right now. Do you want to open an issue to track that?

michaelosthege (approved):

OriolAbril:
"unnamed distribution" should be `Distribution` which for now points to the API _section_ with all distributions, we could update it at any point to be a glossary term. The fact it is unnamed is not related to the _type_ but to a property some objects with this type have and that info should be described at the parameter description section (which is already done, so no changes needed here).

cluhmann:
Yeah, I think a glossary term would be good, because I suspect that stating that this argument should be a "Distribution" is going to cause some confusion.  Chime in [here](https://github.com/pymc-devs/pymc/issues/6095).

Issue #5987, opened 2022-07-19
Prediction could be better if shape of observed would not need to be considered


twiecki:
When I have a simple model:

```python
import pymc as pm

with pm.Model() as m:
    x = pm.MutableData("x", input_data)
    y = pm.MutableData("y" , output_data)
    beta = pm.Normal("beta")
    y_est = x * beta
    pm.Normal("y", y_est, observed=y)
```

Now when I want to predict on new data, I have to do:
```python
with m:
    pm.set_data({"x": np.linspace(0, 10, 100), "y": np.zeros(100)})
```

because `x` and `y` need to have the same shape or I get an error. But it seems wrong that I need to set `y` to some dummy-thing just for the shape. So what I *think* it should look like is this:
```python
with m:
    pm.set_data({"x": np.linspace(0, 10, 100)})
```

I'm not sure what the right solution is, perhaps there is a way to link the shapes somehow, so that when I change `x`, `y` automatically also gets changed? Or that when I call `pm.sample_posterior_predictive()` the data in `observed` just gets removed and the shape should automatically propagate anyway?

CC @ricardoV94

michaelosthege:
Every solution other than returning a new model will have nasty side effects on the current model.

But this may actually result in an okay API:

```python
with pmodel.set_data({
    "x" : ...,
}) as postpred_model:
    idatapp = pm.sample_posterior_predictive()
```

twiecki:
So `model.set_data()` would create a new model object?

bherwerth:
Has this been solved in the meantime? The following works for me and the shape change propagates from `x` to `y` when setting the data. I can also sample the posterior. Modified from the original example (I had to rename the observed to `y_obs`, otherwise there is a name clash with `y`):

```python
import numpy as np
import pandas as pd

import pymc as pm

rng = np.random.default_rng(seed=0)
input_data = np.arange(10, dtype=float)
output_data = 0.5*input_data + 10 + rng.random(size=input_data.shape[0])

with pm.Model() as m:
    x = pm.MutableData("x", input_data)
    y_obs = pm.MutableData("y_obs" , output_data)
    beta = pm.Normal("beta")
    y_est = x * beta
    y = pm.Normal("y", y_est, observed=y_obs)
    samples = pm.sample(100, tune=100, cores=1)
print([v.eval().shape for v in [x, y_est, y]]) # All shape (10,)

with m:
    pm.set_data({"x": np.linspace(0, 10, 100)})
[...]

PR #6087, opened 2022-08-31
Update docstrings of pm.set_data and model.Data

bwengals:
Adds docstrings describing how to track changing shape of `MutableData` in the likelihood.  Meant to address issue #5987.   The recommended solution is then to set the `shape=x.shape`, like this:

```python
import pymc as pm                                                                       
with pm.Model() as model:                                                               
    x = pm.MutableData('x', [1., 2., 3.])                                               
    y = pm.MutableData('y', [1., 2., 3.])                                               
    beta = pm.Normal('beta', 0, 1)                                                      
    obs = pm.Normal('obs', x * beta, 1, observed=y, shape=x.shape) # <--- here                      
    idata = pm.sample(1000, tune=1000)   

with model:                                                                             
    pm.set_data({'x': [5., 6., 9., 12., 15.]})                                               
    y_test = pm.sample_posterior_predictive(idata)                                      
y_test.posterior_predictive['obs'].mean(('chain', 'draw'))
```

## Major / Breaking Changes
None

## Bugfixes / New features
None

## Docs / Maintenance
Add recommendation to docstrings relevant to prediction and posterior predictive sampling.

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -592,12 +592,12 @@ def Data(
     :func:`pymc.set_data`.
 
     To set the value of the data container variable, check out
-    :func:`pymc.Model.set_data`.
+    :meth:`pymc.Model.set_data`.
 
     When making predictions or doing posterior predictive sampling, the shape of the
     registered data variable will most likely need to be changed.  If you encounter an
     Aesara shape mismatch error, refer to the documentation for
-    :func:`pymc.model.set_data`.
+    :meth:`pymc.model.set_data`.
 
     For more information, read the notebook :ref:`nb:data_container`.
 
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -594,6 +594,11 @@ def Data(
     To set the value of the data container variable, check out
     :func:`pymc.Model.set_data`.
 
+    When making predictions or doing posterior predictive sampling, the shape of the
+    registered data variable will most likely need to be changed.  If you encounter an
+    Aesara shape mismatch error, refer to the documentation for
+    :func:`pymc.model.set_data`.
+
     For more information, read the notebook :ref:`nb:data_container`.
 
     Parameters
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1847,14 +1847,9 @@ Model._context_class = Model
 
 
 def set_data(new_data, model=None, *, coords=None):
-    """Sets the value of one or more data container variables.
-
-    When doing prediction or posterior predictive sampling, make sure that
-    the random variable acting as the likelihood (`observed` is True) is dynamically
-    resized when the data is updated.  Set the `shape` argument of the random
-    variable acting as the likelihood to the shape of the relevant
-    :class:`pymc.data.MutableData` object, as shown in the example below
-    (the `shape=x.shape` of `obs`).
+    """Sets the value of one or more data container variables.  Note that the shape is also
+    dynamic, it is updated when the value is changed.  See the examples below for two common
+    use-cases that take advantage of this behavior.
 
     Parameters
     ----------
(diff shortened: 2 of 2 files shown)


pymc/model.py
```diff
@@ -1849,6 +1849,13 @@ def point_logps(self, point=None, round_vals=2):
 def set_data(new_data, model=None, *, coords=None):
     """Sets the value of one or more data container variables.
 
+    When doing prediction or posterior predictive sampling, make sure that
```
canyon289: whats the difference between prediction and posterior predictive sampling?

canyon289 (approved):

pymc/model.py
bwengals: Nothing.. wanted to catch the eye of someone skimming with either word.  Maybe better phrasing here?

twiecki: I would just drop the first prediction mention.

"make sure" is also a bit strong, as it's not required, it's more of a convenience. And maybe also say at the end that the benefit of this is that you only have to `set_data()` `x`, but not both.

ricardoV94: I would also add an example of resizing `y` alone first. MutableData is not useful only in posterior predictive sampling with GLMs where this concern for having to resize both variables arises.

It's also useful to fit the same model to different datasets. 

Would be strange for me as user seeing this being discussed in the only example.

bwengals: @twiecki that makes sense, I'll change it.

@ricardoV94 For an example with resizing `y` alone first, would `y` be a `pm.MutableData` object, and observed?  Makes sense RE fitting the same model to different datasets.  In the past I've just remade the model object but you could save on compiling time.

ricardoV94: Yes

junpenglao:
Is this ready to merge?

pymc/data.py
```diff
@@ -594,6 +594,11 @@ def Data(
     To set the value of the data container variable, check out
     :func:`pymc.Model.set_data`.
```
OriolAbril:
```suggestion
    :meth:`pymc.Model.set_data`.
```

pymc/data.py
```diff
@@ -594,6 +594,11 @@ def Data(
...
 
+    When making predictions or doing posterior predictive sampling, the shape of the
+    registered data variable will most likely need to be changed.  If you encounter an
+    Aesara shape mismatch error, refer to the documentation for
+    :func:`pymc.model.set_data`.
```
OriolAbril:
```suggestion
    :func:`pymc.set_data`.
```

The function is imported as `pymc.set_data` and since 4.0 we document objects in the same place they should be imported.

pymc/model.py
```diff
@@ -1849,6 +1849,13 @@ def point_logps(self, point=None, round_vals=2):
...
+    the random variable acting as the likelihood (`observed` is True) is dynamically
+    resized when the data is updated.  Set the `shape` argument of the random
+    variable acting as the likelihood to the shape of the relevant
+    :class:`pymc.data.MutableData` object, as shown in the example below
```
```suggestion
    :class:`pymc.MutableData` object, as shown in the example below
```

bwengals:
Took another crack at it, basically made the docstring more minimal and added an example as per your suggestion @ricardoV94.

Would it be nice to also have the syntax allow `pm.set_data(y=y)` as well as `pm.set_data({'y': y})`?  a la xarray?

ricardoV94 (approved):
Looks great

ricardoV94:
> Would it be nice to also have the syntax allow `pm.set_data(y=y)` as well as `pm.set_data({'y': y})`?  a la xarray? 

I guess so ;)

bwengals:
One last question, if you don't mind, I can't seem to figure out why [pre-commit isn't happy](https://github.com/pymc-devs/pymc/actions/runs/3131146737/jobs/5082180041#step:4:106).  Since this PR just changes docstrings, what's it saying about numpy, positional arguments?

ricardoV94:
That was fixed today. If you rebase from main it should go away

PR #6088, opened 2022-08-31
Update GOVERNANCE.md for PyMCon_2022 planning repo

canyon289:
@OriolAbril let me know if I got this right. If so we can have the other steering committee members review it.

Additionally I can remove all single member additions to the pymcon_2022 repo and instead grant view and commit permissions to all named groups in the pymc-devs organization. This way anyone in those groups can participate in the pymc web series discussion. What do you think of all that?

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -654,7 +654,8 @@ In addition, Council members are given administrative rights to all repositories
 the [pymc-devs](https://github.com/pymc-devs) organization.
 
 ##### Communication Focused Repositroes
-Some repositories on Github may be used primarily for internal knowledge store and communication, rather than as codebases that are packaged and deployed. 
+Some repositories on Github may be used primarily for internal knowledge store and communication, rather than content tha tis curated, published, or released _by the project_ for external users. 
+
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -653,6 +653,9 @@ Team:
 In addition, Council members are given administrative rights to all repositories within
 the [pymc-devs](https://github.com/pymc-devs) organization.
 
+##### Communication Focused Repositroes
+Some repositories on Github may be used primarily for internal knowledge store and communication, rather than as codebases that are packaged and deployd. In this case permissions, such as write permissions, will be granted in a manner more like other communication platforms such as slack.
```
OriolAbril: I would change the `rather than as codebases...` to something like `rather than content that is curated and published/released _by the project_`. 

Then also the next sentence: 

```The permissions of such repositories will be set in order to allow the same participation and access levels we use on private project communication channels like Slack. Therefore, similarly to slack, these repositories will be private and write permissions will be given to all recurrent contributors (that is, anyone with access to slack).```

OriolAbril (approved):

canyon289:
No changes to rendered docs. Merging

Issue #6031, opened 2022-08-06
Move DensityDist docstrings from `__new__` to class


ricardoV94:
Otherwise they don't show up in the docs: https://www.pymc.io/projects/docs/en/v4.1.4/api/distributions/generated/pymc.DensityDist.html

rafaelcharris:
Hi, I am willing to this.

ricardoV94:
> Hi, I am willing to this.
> 

Go ahead!

PR #6097, opened 2022-09-02
Move DensityDist docstrings from __new__ to class

rafaelcharris:
Solving issue #6031

## Docs / Maintenance
- Move the docstrings from the __new__ function to the beginning of the class so it appears in the documentation,


ricardoV94:
@rafaelcharris apologies in PR #6110 that affects the DensityDist I had to tweak the docstrings so I also moved them out of `__new__`. That means this PR will not be needed anymore :/

rafaelcharris:
got it.
No problem

PR #6101, opened 2022-09-05
Make rvs_to_values work with non-RandomVariables

ricardoV94:
**What is this PR about?**
Cleanup some logic that assumed only RandomVariables can be valued

Related to #6072 
Related to https://github.com/aesara-devs/aeppl/issues/174

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- rvs_to_values now works with non-RandomVariables and unvalued RandomVariables

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -11,8 +11,6 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
-import warnings
-
 from typing import (
     Callable,
     Dict,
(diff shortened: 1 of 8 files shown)


pymc/aesaraf.py
```diff
@@ -198,22 +170,40 @@ def extract_obs_data(x: TensorVariable) -> np.ndarray:
     raise TypeError(f"Data cannot be extracted from {x}")
 
 
+def extract_rv_and_value_vars(
+    random_var: TensorVariable,
+) -> Tuple[TensorVariable, TensorVariable]:
```
lucianopaz: The type hint is wrong here. It should say that the `TensorVariable`s are `Optional`.

pymc/aesaraf.py
```diff
@@ -198,22 +170,40 @@ def extract_obs_data(x: TensorVariable) -> np.ndarray:
...
+        A variable corresponding to a ``RandomVariable``.
+
+    Returns
+    =======
+    The first value in the tuple is the ``RandomVariable``, and the second is the
+    measure/log-likelihood value variable that corresponds with the latter.
+
+    """
+    if not random_var.owner:
+        return None, None
```
lucianopaz: It looks like this isn’t tested. Why should it return `None` instead of raising something like a `TypeError`?

lucianopaz (changes requested):
Looks ok @ricardoV94 . I’m entirely sure about how the model graphs are walked and the replacements are being set. But I think it’s just that I’m tired and I’ll have to review it a second time. Anyway, I wanted to ask you to change a type hint that I think is wrong and maybe also explain why you don’t raise an exception instead of returning Nones

pymc/aesaraf.py
ricardoV94: I think this is for input variables that don't have an `owner` like `at.scalar()`. I can however have the check outside. The whole function is probably not needed...

twiecki (approved):

PR #6109, opened 2022-09-07
Update .gitpod.yml

ferrine:
Optimize gitpod initialization as tested in https://github.com/pymc-devs/pymc-experimental/pull/68

image
image


ferrine:
@ricardoV94 seems like we can merge this

ricardoV94 (approved):

Issue #6091, opened 2022-09-01
Flaky `test_profile.py::TestProfile::test_profile_variable `


ricardoV94:
I have seen this one fail repeatedly recently, such as in https://github.com/pymc-devs/pymc/runs/8135110596?check_suite_focus=true

```python
 ================================== FAILURES ===================================
______________________ TestProfile.test_profile_variable ______________________

self = <pymc.tests.test_profile.TestProfile object at 0x0000020C722FB460>

    def test_profile_variable(self):
        rv = self.model.basic_RVs[0]
>       assert self.model.profile(self.model.logp(vars=[rv], sum=False)).fct_call_time
E       assert 0.0
E        +  where 0.0 = <aesara.compile.profiling.ProfileStats object at 0x0000020C729943D0>.fct_call_time
E        +    where <aesara.compile.profiling.ProfileStats object at 0x0000020C729943D0> = <bound method Model.profile of <pymc.model.Model object at 0x0000020C728B4130>>([Elemwise{mul,no_inplace}.0])
E        +      where <bound method Model.profile of <pymc.model.Model object at 0x0000020C728B4130>> = <pymc.model.Model object at 0x0000020C728B4130>.profile
E        +        where <pymc.model.Model object at 0x0000020C728B4130> = <pymc.tests.test_profile.TestProfile object at 0x0000020C722FB460>.model
E        +      and   [Elemwise{mul,no_inplace}.0] = <bound method Model.logp of <pymc.model.Model object at 0x0000020C728B4130>>(vars=[x], sum=False)
E        +        where <bound method Model.logp of <pymc.model.Model object at 0x0000020C728B4130>> = <pymc.model.Model object at 0x0000020C728B4130>.logp
E        +          where <pymc.model.Model object at 0x0000020C728B4130> = <pymc.tests.test_profile.TestProfile object at 0x0000020C722FB460>.model

pymc\tests\test_profile.py:27: AssertionError
```

ricardoV94:
`test_profile_model` also failed in https://github.com/pymc-devs/pymc/actions/runs/3214681730/jobs/5255237075

PR #6112, opened 2022-09-09
Remove support for "dims on the fly"

### LINKED ISSUE #6085: Remove support of "dims on the fly"
### Discussed in https://github.com/pymc-devs/pymc/discussions/6065

<sup>Originally posted by **ricardoV94** August 25, 2022</sup>
Right now we can specify dims on the fly, which are taken from the shape of a variable:

```python
import pymc as pm
with pm.Model() as m:
  x = pm.Normal("x", shape=(3,), dims=("trial",))
  print(m.dim_lengths["trial"].eval())  # 3
```

This is neat, but causes some code complexity, because we also allow dims to inform the shape of a variable.
```python
with pm.Model(coords={"trial": range(3)}) as m:
  x = pm.Normal("x", dims=("trial",))
  print(x.eval().shape)  # (3,)
```

Which forces us to create new variables in a two-step process, first without dim information and then with it (via resizing). Code could be simplified if we don't have to worry about dims being specified on the fly. 

These two behaviors of dims are also potentially confusing when mixed together:
```python
with pm.Model() as m:
  x = pm.Normal("x", shape=5, dims=("trial",))
  y = pm.Normal("y", dims=("trial",))
  print(y.eval().shape)  # (5,)
```
</div>

ricardoV94:
Closes #6085

**Checklist**

## Major / Breaking Changes
- Dims can no longer be implied by the size of a variable or observed, they must always be defined in advance

## Bugfixes / New features
- ...

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -29,7 +29,7 @@ from aeppl.rewriting import logprob_rewrites_db
 from aesara import tensor as at
 from aesara.compile.builders import OpFromGraph
 from aesara.graph import node_rewriter
-from aesara.graph.basic import Node, Variable, clone_replace
+from aesara.graph.basic import Node, clone_replace
 from aesara.graph.rewriting.basic import in2out
 from aesara.graph.utils import MetaType
 from aesara.tensor.basic import as_tensor_variable
@@ -42,9 +42,6 @@ from pymc.aesaraf import convert_observed_data
 from pymc.distributions.shape_utils import (
     Dims,
     Shape,
-    StrongDims,
-    StrongShape,
-    change_dist_size,
     convert_dims,
     convert_shape,
     convert_size,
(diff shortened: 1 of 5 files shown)


twiecki (approved):

ricardoV94:
Marking as a draft because some checks we had in `model.set_data` may no longer make sense?

ricardoV94:
Failing test is unrelated: #6091

pymc/model.py
```diff
@@ -1295,6 +1295,7 @@ def set_data(
                         )
                     else:
                         # The dimension is immutable.
+                        # TODO: Is this still possible?
```
ricardoV94: @michaelosthege  can I get your input here? Do you remember if this condition related only to inferred dims?

michaelosthege: This condition is related to dims that were defined by the `shape` of a non-shared `TensorVariable`.

This was the case for inferred dims, but there are at least unintended ways how to end up here:

```python
with pm.Model() as pmodel:
    x = pm.Normal("x", size=7)

    pmodel.dim_lengths["weekday"] = x.shape[0]       # this definitely
    pmodel.add_coord("weekday", length=x.shape[0])   # maybe also this

    pm.MutableData("y", np.arange(7), dims="weekday")
pmodel.set_data("y", np.arange(10))
```

ricardoV94: Isn't this another form of "dims on the fly"?

ricardoV94: Anyway added your example to cover that branch

Issue #6106, opened 2022-09-07
Remove `reshape_t`


ricardoV94:
It no longer seems to be true that `x.reshape(())` does not work in Aesara.

https://github.com/pymc-devs/pymc/blob/761f77dad29116650ef3ca043645042219654a5c/pymc/aesaraf.py#L618-L623

ricardoV94:
Closed via #6118

PR #6118, opened 2022-09-11
Remove reshape_t

tjburch:
**What is this PR about?**

Addresses #6106, removing the `reshape_t` function and the lone reference to it.

I did run into local issues with tests in `test_smc.py`, but they're all `NotImplementedError`s so it might be external?

## Major / Breaking Changes
- None

## Bugfixes / New features
- None

## Docs / Maintenance
- Cleanup of `reshape_t` which is no longer necessary since `reshape` works in Aesara.


canyon289:
Thanks for the contribution!

canyon289 (approved):

PR #6121, opened 2022-09-12
Add `ZeroSumNormal` distribution

AlexAndorra:
This PR introduces the world famous `ZeroSumNormal` distribution, i.e a Normal distribution where one or several axes are constrained to sum to zero. By default, the last axis is constrained to sum to zero.

The `zerosum_axes` are always in the rightmost position, i.e `zerosum_axes=2` means the two rightmost axes will be constrained to sum to zero (see examples below for more details).

⚠️ **`sigma` has to be a scalar, to ensure the zero-sum constraint**. The ability to specifiy a vector of `sigma` may be added in future versions.

**Checklist**
+ [x] Docstrings
+ [x] Tests
+ [x] Test ZSN variance
+ [x] Test ZSN logp
+ [x] Test `get_support_shape` with 2D

---
Examples:
```python
COORDS = {
    "regions": ["a", "b", "c"],
     "answers": ["yes", "no", "whatever", "don't understand question"],
}
with pm.Model(coords=COORDS) as m:
    # the zero sum axis will be 'answers'
    v = pm.ZeroSumNormal("v", dims=("regions", "answers"))

with pm.Model(coords=COORDS) as m:
    # the zero sum axes will be 'answers' and 'regions'
    v = pm.ZeroSumNormal("v", dims=("regions", "answers"), zerosum_axes=2)

with pm.Model(coords=COORDS) as m:
    # the zero sum axes will be the last two
    v = pm.ZeroSumNormal("v", shape=(3, 4, 5), zerosum_axes=2)
```

## Major / Breaking Changes
- ...

## Bugfixes / New features
- Implemented `ZeroSumNormal` distribution

## Docs / Maintenance
- ...


pymc/distributions/continuous.py
```diff
@@ -585,6 +600,172 @@ def logcdf(value, mu, sigma):
...
+        if zerosum_axes is None:
+            zerosum_axes = [-1]
+
+        sigma = at.as_tensor_variable(floatX(sigma))
+        if sigma.ndim > 0:
+            raise ValueError("sigma has to be a scalar")
+
+        return super().dist([sigma], zerosum_axes=zerosum_axes, **kwargs)
+
+    # TODO: This is if we want ZeroSum constraint on other dists than Normal
```
ricardoV94: @lucianopaz @aseyboldt does any of the math require the summed distribution to be a `Normal` or could it be something else (e.g., `StudentT`)?

pymc/distributions/continuous.py
```diff
@@ -585,6 +600,172 @@ def logcdf(value, mu, sigma):
...
+
+        return ZeroSumNormalRV(
+            inputs=[normal_dist_, sigma_],
+            outputs=[zerosum_rv_],
+            zerosum_axes=zerosum_axes,
+            ndim_supp=0,
+        )(normal_dist, sigma)
+
+
+@_logprob.register(ZeroSumNormalRV)
```
ricardoV94: Just for organization I would put the dispatched functions in this order:
1. change_size
2. moment
3. transform
4. logprob

pymc/distributions/transforms.py
```diff
@@ -266,6 +267,68 @@ def bounds_fn(*rv_inputs):
...
+    fill_val = norm - sum_vals / np.sqrt(n)
+
+    out = at.concatenate([array, fill_val], axis=axis)
+    return out - norm
+
+
+def extend_axis_rev(array, axis):
+    if axis < 0:
+        axis = axis % array.ndim
+    assert axis >= 0 and axis < array.ndim
```
ricardoV94: I think you can just use the numpy `normalize_axis_tuple` thing here instead

AlexAndorra:
Thanks for the first review @ricardoV94 !
I added some tests. How do they look like?

pymc/tests/distributions/test_continuous.py
```diff
@@ -1802,6 +1802,46 @@ class TestTruncatedNormalUpperArray(BaseTestDistributionRandom):
     ]
 
 
+class TestZeroSumNormal(BaseTestDistributionRandom):
```
ricardoV94: That baseclass doesn't make sense for the tests

```suggestion
class TestZeroSumNormal:
```

pymc/tests/distributions/test_continuous.py
```diff
@@ -1802,6 +1802,46 @@ class TestTruncatedNormalUpperArray(BaseTestDistributionRandom):
...
+    COORDS = {
+        "regions": ["a", "b", "c"],
+        "answers": ["yes", "no", "whatever", "don't understand question"],
+    }
+    with pm.Model(coords=COORDS) as m:
```
ricardoV94: You can use parametrize to try different zerosum_axes / type more compactly, instead of having multiple tests / conditions within a test.

pymc/tests/distributions/test_continuous.py
```diff
@@ -1802,6 +1802,46 @@ class TestTruncatedNormalUpperArray(BaseTestDistributionRandom):
...
+        v = pm.ZeroSumNormal("v", dims=("regions", "answers"), zerosum_axes="answers")
+        s = pm.sample(10)
```
ricardoV94: Would be good to try `pm.draw(v)` as well to generate draws from the forward graph and not just from NUTS sampling.

ricardoV94: Make this go a bit faster

```suggestion
        s = pm.sample(10, chains=1, tune=100)
```

pymc/distributions/continuous.py
AlexAndorra: Done

pymc/distributions/transforms.py
AlexAndorra: Ah yeah, good catch! There is a super weird thing happening though: this is breaking static type checks and I have no idea why!

AlexAndorra: Lol, super weird but I had to disable mypy on the import of `normalize_axis_tuple`  because it somehow considers that it doesn't exist 🤨

pymc/tests/distributions/test_continuous.py
AlexAndorra: Done

pymc/distributions/continuous.py
```diff
@@ -585,6 +600,172 @@ def logcdf(value, mu, sigma):
...
+
+@_change_dist_size.register(ZeroSumNormalRV)
+def change_zerosum_size(op, normal_dist, new_size, expand=False):
+    normal_dist, sigma = normal_dist.owner.inputs
+    if expand:
+        new_size = tuple(new_size) + tuple(normal_dist.shape)
+    return ZeroSumNormal.rv_op(sigma=sigma, zerosum_axes=op.zerosum_axes, size=new_size)
+
+
+@_moment.register(ZeroSumNormalRV)
```
ricardoV94: Need a test for this

pymc/distributions/continuous.py
```diff
@@ -585,6 +600,172 @@ def logcdf(value, mu, sigma):
...
+        return ZeroSumNormalRV(
+            inputs=[normal_dist_, sigma_],
+            outputs=[zerosum_rv_],
+            zerosum_axes=zerosum_axes,
+            ndim_supp=0,
+        )(normal_dist, sigma)
+
+
+@_change_dist_size.register(ZeroSumNormalRV)
+def change_zerosum_size(op, normal_dist, new_size, expand=False):
```
ricardoV94: Need a test for this, check tests in test_timeseries for change size

ricardoV94 (commented):
This would make more sense in multivariate even though we are treating it as a scalar as a hack for the time being.

pymc/tests/distributions/test_continuous.py
AlexAndorra: Done

AlexAndorra: Done

pymc/distributions/continuous.py
AlexAndorra: Done

pymc/distributions/continuous.py
AlexAndorra: Done

AlexAndorra:
> This would make more sense in multivariate even though we are treating it as a scalar as a hack for the time being.

Good point. I moved it to `multivariate.py`.

I also added all the tests mentioned and pushed everything. They pass locally. Let's see if they pass here 🤞

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,169 @@ def logp(value, alpha, K):
...
+    rv_type = ZeroSumNormalRV
+
+    def __new__(cls, *args, zerosum_axes=None, dims=None, **kwargs):
+        dims = convert_dims(dims)
+        if zerosum_axes is None:
+            zerosum_axes = [-1]
+        if not isinstance(zerosum_axes, (list, tuple)):
+            zerosum_axes = [zerosum_axes]
+
+        if isinstance(zerosum_axes[0], str):
```
lucianopaz: Can’t you mix dimension names and axis indices in the zero sum axes parameter? Using exclusively one or the other seems unnecessarily restrictive

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,169 @@ def logp(value, alpha, K):
...
+    #     check_dist_not_registered(dist)
+    #     return super().dist([dist, lower, upper], **kwargs)
+
+    @classmethod
+    def rv_op(cls, sigma, zerosum_axes, size=None):
+        if size is None:
+            zerosum_axes_ = np.asarray(zerosum_axes)
+            # just a placeholder size to infer minimum shape
+            size = np.ones(
+                max((max(np.abs(zerosum_axes_) - 1), max(zerosum_axes_))) + 1, dtype=int
```
lucianopaz: Why do you have two max operations here?

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,169 @@ def logp(value, alpha, K):
...
+            zerosum_axes = [zerosum_axes]
+
+        if isinstance(zerosum_axes[0], str):
+            if not dims:
+                raise ValueError("You need to specify dims if zerosum_axes are strings.")
+            else:
+                zerosum_axes_ = []
+                for axis in zerosum_axes:
+                    zerosum_axes_.append(dims.index(axis))
+                zerosum_axes = zerosum_axes_
```
lucianopaz: I might be missing something, but we had a loop over the zero sum axes to convert them all to negative integers, that way they would reference the tensor axis from the last to the first. Was that removed for some reason?

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,169 @@ def logp(value, alpha, K):
...
+            ndim_supp=0,
+        )(normal_dist, sigma)
+
+
+@_change_dist_size.register(ZeroSumNormalRV)
+def change_zerosum_size(op, normal_dist, new_size, expand=False):
+    normal_dist, sigma = normal_dist.owner.inputs
+    if expand:
+        new_size = tuple(new_size) + tuple(normal_dist.shape)
+    return ZeroSumNormal.rv_op(sigma=sigma, zerosum_axes=op.zerosum_axes, size=new_size)
```
lucianopaz: If you are changing size, the safe thing is to always set the zero sum axes to negative integers. If you don’t do that and the size change adds extra dimensions, then your zero sum will be applied to the wrong axis

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,169 @@ def logp(value, alpha, K):
...
+    _degrees_of_freedom = at.prod(_deg_free_shape)
+    zerosums = [
+        at.all(at.isclose(at.mean(value, axis=axis), 0, atol=1e-9)) for axis in op.zerosum_axes
+    ]
+    # out = at.sum(
+    #     pm.logp(dist, value) * _degrees_of_freedom / _full_size,
+    #     axis=op.zerosum_axes,
+    # )
+    # figure out how dimensionality should be handled for logp
+    # for now, we assume ZSN is a scalar distribut, which is not correct
```
lucianopaz: You **do need to sum** across the zero sum axes in the logp. That’s what gives you the formula you are using!
The way in which is logp is derived is by noting that a zerosumNormal is an MvNormal, with a singular covariance matrix. In the old PR about this, I gave a longer explanation on this but the short reason for using the proper logp is that we can use the special MvNormal as a baseline comparison of the logp values

pymc/distributions/transforms.py
```diff
@@ -266,6 +271,66 @@ def bounds_fn(*rv_inputs):
         super().__init__(args_fn=bounds_fn)
 
 
+class ZeroSumTransform(RVTransform):
+    """
+    Constrains the samples of a Normal distribution to sum to zero
```
lucianopaz: The transform isn’t limited to normal distributions. I think that it’s best to take that out of the docstring and just say that it constrain the samples to sum to zero

pymc/distributions/transforms.py
```diff
@@ -266,6 +271,66 @@ def bounds_fn(*rv_inputs):
...
+        for axis in self.zerosum_axes:
+            value = extend_axis(value, axis=axis)
+        return value
+
+    def log_jac_det(self, value, *rv_inputs):
+        return at.constant(0.0)
+
+
+def extend_axis(array, axis):
+    n = array.shape[axis] + 1
```
lucianopaz, line 306: We could maybe add a comment here saying that this is using a householder reflection plus a projection operator to move forward from the constrained space onto the zero sum manifold. I’ll look up our notes and write something here

pymc/tests/distributions/test_multivariate.py
```diff
@@ -1367,6 +1391,100 @@ def test_issue_3706(self):
...
+
+            nonzero_axes = list(set(dims).difference(zerosum_axes))
+            if nonzero_axes:
+                for ax in nonzero_axes:
+                    for samples in [
+                        s.posterior.v.mean(dim=ax),
+                        random_samples.mean(axis=dims.index(ax) + 1),
+                    ]:
+                        assert not np.isclose(
+                            samples, 0
```
lucianopaz: You should draw more samples and compare the standard deviation to the desired value imposed by the distribution

pymc/tests/distributions/test_multivariate.py
```diff
@@ -1367,6 +1391,100 @@ def test_issue_3706(self):
...
+        else:
+            with pytest.raises(AxisError, match="out of bounds"):
+                with pm.Model(coords=COORDS) as m:
+                    _ = pm.ZeroSumNormal("v", dims=dims, zerosum_axes=zerosum_axes)
+
+    def test_zsn_change_dist_size(self):
+        base_dist = pm.ZeroSumNormal.dist(shape=(4, 9))
+
+        new_dist = change_dist_size(base_dist, new_size=(5, 3), expand=False)
+        assert new_dist.eval().shape == (5, 3)
```
lucianopaz: You should also test that once you change size and draw some random numbers from it, your zero sum axes actually sum to zero

pymc/tests/distributions/test_multivariate.py
```diff
@@ -1367,6 +1391,100 @@ def test_issue_3706(self):
...
+
+        new_dist = change_dist_size(base_dist, new_size=(5, 3), expand=True)
+        assert new_dist.eval().shape == (5, 3, 4, 9)
+
+
```
lucianopaz, line 1587: We also need to add a test from the logp against the baseline singular MvNormal

lucianopaz (dismissed):
@AlexAndorra i left a bunch of comments all around the place. I think that I’m missing the part where the RV asserts that the zero sum axis are all negative and non repeating. I also think that you need to add some tests for the logp of the distribution and for the variance of the draws

pymc/distributions/multivariate.py
AlexAndorra: Yeah, the function works with positive integers too now, so we don't need this conversion anymore

pymc/distributions/multivariate.py
AlexAndorra: Ha ha yeah that's super ugly and I don't remember exactly why, but we determined it basically through trial and error. The goal is to pass the minimum size to Aesara (i.e the zerosum axes)

pymc/distributions/multivariate.py
AlexAndorra: Yeah this is still confusing to me too why we don't use the sum here. I deferred to @ricardoV94, who explained to me that, since we're (wrongly) assuming for now that ZSN is a scalar distribution, the results will be correct -- and they indeed seem to be, at least according to the tests we did.
@ricardoV94 you probably wanna chime in?

pymc/distributions/multivariate.py
AlexAndorra: Mmmh that's a good point. I tested this change size function, but I didn't check that the zero sum constraint was still applied to the right axis.
Do you have a test case in mind @lucianopaz ?

pymc/tests/distributions/test_multivariate.py
AlexAndorra: Meaning that the ZSN logp should be equal to the MvNormal's?

pymc/distributions/multivariate.py
lucianopaz: The older versions always could work with positive zero sum axis but we chose to enforce the RV to have negative values so that when changing the size you didn’t have to potentially update the zero sum axes. The distribution itself handled the conversion from dimension names and positive axis to the equivalent negative axis.

pymc/tests/distributions/test_multivariate.py
lucianopaz: It’s equal to a special form of MvNormal. The thing is that the covariance matrix of the MvNormal is singular and you can’t compute its logp using our pymc dist. I’ll pull up the original comment from the old PR where I talked about this

lucianopaz: Here’s the link to the old [thread](https://github.com/pymc-devs/pymc/pull/4776#discussion_r800245250) about the logp

pymc/tests/distributions/test_multivariate.py
AlexAndorra: But the distribution doesn't impose any standard deviation value to ZSN, IIUC: the 1 in `sigma` refers to the underlying, unconstrained Normal distribution

pymc/distributions/multivariate.py
AlexAndorra: Ah ok I see. We definitely need to test that and revert to that old behavior if that's still an issue

pymc/tests/distributions/test_multivariate.py
AlexAndorra: Thanks @lucianopaz ! So this means we should test our ZSN logp against the one you posted in the old thread (`logp(value, sigma)`)?
If yes, then that means we can only test the case where only one axis sums to zero?

lucianopaz: I’ll send you a notebook I have where this is tested against multiple axes

pymc/distributions/multivariate.py
ricardoV94: If you sum the logp across the zerosum axis they should only be allowed to be to the rightmost position. Everything that works with multivariate distributions in Aesara, Aeppl and PyMC expects that at the moment (we will probably relax those constraints in the future)

Also you will need to add an internal variable for the shape dimensions that would correspond to the support dimensions since size cannot include those, like `steps` for timeseries.

Given those constraints I suggested we treat it as a scalar internally temporarily. 

Also because of these API things we have to figure out, I think this makes more sense to go to pymc-experimental first @AlexAndorra

pymc/distributions/multivariate.py
ricardoV94: I disagree. A single type of axis representation seems much more sane, even if it's not particularly difficult to work with mixed types.

pymc/tests/distributions/test_multivariate.py
AlexAndorra: Oh cool, thanks @lucianopaz !

pymc/tests/distributions/test_multivariate.py
AlexAndorra: Just tested this (see updated `test_zsn_change_dist_size`), and it seems to work 🥳

pymc/distributions/multivariate.py
ricardoV94: I am not convinced about the flexibility of non-consecutive zerosum axis. I proposed @AlexAndorra to always zerosum to the right and then the question is not which axes but how many. 

We can obviously add a helper on top of it that Dimshuffles a latent ZSN (or users can do it themselves). This would bypass the issue that ZSN cannot safely be treated as a multivariate because the support dimensions are not to the right.

pymc/distributions/multivariate.py
AlexAndorra: Did that test (see corresponding thread)

pymc/distributions/multivariate.py
AlexAndorra: Doesn't seem to be an issue (see `test_zsn_change_dist_size`)

pymc/distributions/multivariate.py
AlexAndorra: Yeah, I agree with @ricardoV94 on this

pymc/distributions/multivariate.py
AlexAndorra: That's clearer, thanks @ricardoV94.

> Given those constraints I suggested we treat it as a scalar internally temporarily

And to make sure: this doesn't change anything mathematically?

> because of these API things we have to figure out, I think this makes more sense to go to pymc-experimental first

Good point. We should probably wait for Aesara to relax its constraints on multivariate distributions before adding ZSN to PyMC. This would allow to sum the logp and alleviate the need for an internal variable at the same time.
I wonder if I can just just transfer this PR and its history to pymc-experimental 🤔

pymc/distributions/multivariate.py
AlexAndorra: Yep, that's also a good idea. So basically enforcing `zerosum_axes` to be to the rightmost position

pymc/distributions/multivariate.py
lucianopaz: I honestly don’t mind changing our ZSN design to enforce the zero sum axes to be at the right. Users would be able to transpose the results to the order they care about

pymc/tests/distributions/test_multivariate.py
lucianopaz: Yes, but if you set sigma to 2 or something else, the variance across draws will be determined by the sigma value

pymc/tests/distributions/test_multivariate.py
lucianopaz: You need to test after change dist size is called, not before

pymc/tests/distributions/test_multivariate.py
```diff
@@ -1475,14 +1472,39 @@ def test_zsn_fail_axis(self, dims, zerosum_axes):
...
         assert new_dist.eval().shape == (5, 3)
+        random_samples = pm.draw(new_dist, draws=100)
+        self.assert_zerosum_axes(random_samples, zerosum_axes)
 
         new_dist = change_dist_size(base_dist, new_size=(5, 3), expand=True)
         assert new_dist.eval().shape == (5, 3, 4, 9)
+        random_samples = pm.draw(new_dist, draws=100)
+        self.assert_zerosum_axes(random_samples, zerosum_axes)
+
+    def assert_zerosum_axes(self, random_samples, zerosum_axes):
```
lucianopaz: You shouldn’t treat the negative and positive axis differently here. You should just loop across the zero sum axes and try `random_samples.mean(axis=ax)`. If that isn’t close to zero, then there’s a problem. The fact that you had to sum 1 to the positive axes to me means that something fishy is happening

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,171 @@ def logp(value, alpha, K):
...
+        self.zerosum_axes = zerosum_axes
+        super().__init__(*args, **kwargs)
+
+
+class ZeroSumNormal(Distribution):
+    r"""
+    ZeroSumNormal distribution, i.e Normal distribution where one or
+    several axes are constrained to sum to zero.
+    By default, the last axis is constrained to sum to zero.
+    See `zerosum_axes` kwarg for more details.
```
aseyboldt, line 2400: I think we should add a more formal definition, at least for the case of one zerosum axis:

$$
ZSN(\sigma) = N(0, σ^2 (I - \tfrac{1}{n}J)
$$

where $J_{ij} = 1$.

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,171 @@ def logp(value, alpha, K):
...
+
+    Parameters
+    ----------
+    sigma : tensor_like of float
+        Standard deviation (sigma > 0).
```
aseyboldt: `sigma` is not the standard deviation, but a scale parameter.

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,171 @@ def logp(value, alpha, K):
...
+    rv_type = ZeroSumNormalRV
+
+    def __new__(cls, *args, zerosum_axes=None, dims=None, **kwargs):
+        dims = convert_dims(dims)
+        if zerosum_axes is None:
+            zerosum_axes = [-1]
+        if not isinstance(zerosum_axes, (list, tuple)):
+            zerosum_axes = [zerosum_axes]
+
+        if isinstance(zerosum_axes[0], str):
```
aseyboldt: I think we might want to handle the case where `zerosum_axes=[]`?

pymc/distributions/multivariate.py
aseyboldt: Maybe something like this is a bit easier to read?
```python
def find_min_rank(axes):
    if not axes:
        return 0
    
    neg = [-axis for axis in axes if axis < 0]
    pos = [axis + 1 for axis in axes if axis >= 0]
    
    return max(pos + neg)
```

pymc/distributions/multivariate.py
aseyboldt: I might be missing something, but couldn't we treat it as if `ndim_support` is the number of axis to the right of the first zero_sum constraint axis? So something like this:
```
pm.ZeroSumNormal("x", dims=("a", "b", "c"), zerosum_axes=["b"])
```
Would be a RV over `("b", "c")`, ie the elemwise logp would have dim `("a",)`.

pymc/tests/distributions/test_multivariate.py
aseyboldt: In the 1d case the marginal variance should be `sigma ** 2 * (n - 1) / n` (from the covariance in the definition)

pymc/distributions/multivariate.py
AlexAndorra: > I honestly don’t mind changing our ZSN design to enforce the zero sum axes to be at the right

I agree. @ricardoV94 is there any way we enforce this constraint, or it's an assumption we make in the code?

> I might be missing something, but couldn't we treat it as if ndim_support is the number of axis to the right of the first zero_sum constraint axis?

I think it's possible (need @ricardoV94's confirmation). In any case that'd mean the zerosum axes are to the rightmost position

pymc/tests/distributions/test_multivariate.py
AlexAndorra: Ah thanks @aseyboldt , I'll test against this formula

pymc/tests/distributions/test_multivariate.py
AlexAndorra: No I don't think that's a problem here: I had to add 1 because `pm.draw` adds an axis to the left for the draws dimension. So, when the axes are specified with positive integers, you have to account for it. 
Same thing when I test the NUTS samples (I had to add 2 because of `chain` and `draw` added to the left)

pymc/distributions/multivariate.py
AlexAndorra: Isn't that just a Normal distribution?

pymc/distributions/transforms.py
AlexAndorra: Done

pymc/distributions/transforms.py
AlexAndorra: Did you find your notes @lucianopaz ?

pymc/distributions/multivariate.py
AlexAndorra: Easier to read indeed! I think it's doing something slightly different though. If you're doing `pm.ZeroSumNormal.dist(shape=(4, 9), zerosum_axes=(0, 1))`, you'll get `2` with your function, versus `[1, 1]` with the current code

pymc/distributions/multivariate.py
AlexAndorra: Added. Is that better @aseyboldt ?

pymc/distributions/multivariate.py
AlexAndorra: Added. Is that better @aseyboldt ?

AlexAndorra:
There is a weird error on the [ubuntu tests](https://github.com/pymc-devs/pymc/actions/runs/3073048250/jobs/4964929407#step:7:818):
```python
>       self.step_size = step_scale / (size**0.25)
E       ZeroDivisionError: float division by zero
```
It's not even coming from ZSN code, but from NUTS 🤨 
Anybody knows what's that about? Most probably it's a platform issue, because I can't replicate that, and the tests pass for the other platforms.

There is also a [failing test on Windows](https://github.com/pymc-devs/pymc/actions/runs/3073048250/jobs/4964929759#step:7:166), but that's definitely unrelated to this PR:
```python
>       npt.assert_allclose(np.asarray(self.centers), np.sort(Xu), atol=0.1)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0.1
E       
E       Mismatched elements: 1 / 2 (50%)
E       Max absolute difference: 0.10133194
E       Max relative difference: 0.02068561
E        x: array([-5,  5])
E        y: array([-4.898668,  4.96629 ])
```

pymc/distributions/multivariate.py
aseyboldt: It is just computing the rank, so the length of your `size`, you'd get the size as `size = at.zeros(find_min_rank(...), dtype=int)`.
But isn't the rank what you need anyway?

pymc/distributions/multivariate.py
aseyboldt: > In any case that'd mean the zerosum axes are to the rightmost position

Why? It could still be the first axes, only all following axes would be interpreted as core dims, right?

pymc/distributions/multivariate.py
aseyboldt: It is, but if you write more general code that somehow produces the `zerosum_axes` automatically, that could easily happen I think.

pymc/distributions/multivariate.py
ricardoV94: Yeah we could also do that, but I still like the only consecutive zerosum axis +  rightmost position more.

pymc/distributions/multivariate.py
AlexAndorra: Mmmh no, from what I understand you need a placeholder with the right size, not just the size

pymc/distributions/multivariate.py
AlexAndorra: Oh so you mean erroring out in that case, gotcha

aseyboldt: Well, erroring out would be ok, but can't we just handle that case correctly? I can't think of anything that should go wrong in this case. So why not just 
```
if len(zerosum_axes) > 0 and isinstinace(zerosum_axes[0]), str):
```
And maybe add a test that checks if `pm.ZeroSumNormal("y", zerosum_axes=[], shape=(3,))` or so works and gives us a normal dist.

AlexAndorra: Oh ok, I understand what you mean now. I'm curious what @ricardoV94 thinks, but I would prefer not to do that: if people wanna use a Normal they should just use `pm.Normal`, otherwise I find it makes code and models more confusing

pymc/distributions/multivariate.py
AlexAndorra: The rightmost position enforcement is done (see latest updates).
For future reference, when @ricardoV94 says he "likes" a solution, that doesn't mean said solution is _easy_ to implement

pymc/tests/distributions/test_multivariate.py
AlexAndorra: Outdated: `zerosum_axes` are now to the rightmost positions, so what counts is how many there are, not _where_ they are

pymc/distributions/multivariate.py
AlexAndorra: Actually I agree with you now @aseyboldt (mainly because that behavior would be consistent with other PyMC distributions' behavior). 
However, this edge-case needs a bit more work right now, because [`at.stack` in `get_support_shape`](https://github.com/pymc-devs/pymc/pull/6121/files#diff-00de3aa37024b2f780a8b60a86a006fbf3254e37c4aeb604eb3e38aaeb70df51R666) fails with an empty tuple.
As it's not a core feature of ZSN, I'm [marking this as TODO](https://github.com/pymc-devs/pymc/pull/6121/files#diff-ba7699f4e550fe6433c577626dd433a017c2fad177dcbfc83ea8268fe4322761R2491). The code is actually already there, so if someone needs this in the future, there are already some breadcrumbs to get started.

pymc/tests/distributions/test_multivariate.py
AlexAndorra: That works! Added a test 😉

AlexAndorra:
@lucianopaz @ricardoV94 you can re-review. I'll try to add the test against MvNormal logp this weekend, but other than that this PR should be complete 🤞

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,220 @@ def logp(value, alpha, K):
...
+                ndim_supp=zerosum_axes,
+            )
+
+        return super().__new__(
+            cls, *args, zerosum_axes=zerosum_axes, support_shape=support_shape, dims=dims, **kwargs
+        )
+
+    @classmethod
+    def dist(cls, sigma=1, zerosum_axes=None, support_shape=None, **kwargs):
+        if zerosum_axes is None:
```
ricardoV94: This should reuse the cls.check_zerosum_axes

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,220 @@ def logp(value, alpha, K):
...
+
+        assert zerosum_axes == at.get_vector_length(
+            support_shape
+        ), "support_shape has to be as long as zerosum_axes"
+
+        return super().dist(
+            [sigma], zerosum_axes=zerosum_axes, support_shape=support_shape, **kwargs
+        )
+
+    # TODO: This is if we want ZeroSum constraint on other dists than Normal
```
ricardoV94: Let's remove this TODO

pymc/tests/distributions/test_multivariate.py
```diff
@@ -755,7 +755,12 @@ def test_car_logp(self, sparse, size):
 
         # d x d adjacency matrix for a square (d=4) of rook-adjacent sites
         W = np.array(
-            [[0.0, 1.0, 1.0, 0.0], [1.0, 0.0, 0.0, 1.0], [1.0, 0.0, 0.0, 1.0], [0.0, 1.0, 1.0, 0.0]]
+            [
```
ricardoV94: Your pre-commit is changing unrelated stuff

pymc/tests/distributions/test_multivariate.py
```diff
@@ -1027,7 +1045,12 @@ def test_mv_normal_moment(self, mu, cov, size, expected):
     )
     def test_car_moment(self, mu, size, expected):
         W = np.array(
-            [[0.0, 1.0, 1.0, 0.0], [1.0, 0.0, 0.0, 1.0], [1.0, 0.0, 0.0, 1.0], [0.0, 1.0, 1.0, 0.0]]
```
ricardoV94: More unrelated changes

pymc/tests/distributions/test_multivariate.py
```diff
@@ -1368,6 +1391,159 @@ def test_issue_3706(self):
...
+            assert new_dist.eval().shape == (5, 3, 4, 9)
+        random_samples = pm.draw(new_dist, draws=100)
+        self.assert_zerosum_axes(random_samples, zerosum_axes)
+
+        new_dist = change_dist_size(base_dist, new_size=(5, 3), expand=True)
+        assert new_dist.eval().shape == (5, 3, 4, 9)
+        random_samples = pm.draw(new_dist, draws=100)
+        self.assert_zerosum_axes(random_samples, zerosum_axes)
+
+    def assert_zerosum_axes(self, random_samples, axes_to_check, check_zerosum_axes=True):
```
ricardoV94: Can't you reuse this in some of the tests above?

pymc/distributions/multivariate.py
AlexAndorra: Done

pymc/distributions/multivariate.py
AlexAndorra: Done

pymc/tests/distributions/test_multivariate.py
AlexAndorra: I think this is solved

pymc/tests/distributions/test_multivariate.py
AlexAndorra: Added this test and it passes, thanks @lucianopaz !

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,200 @@ def logp(value, alpha, K):
...
+    Examples
+    --------
+    .. code-block:: python
+        COORDS = {
+            "regions": ["a", "b", "c"],
+            "answers": ["yes", "no", "whatever", "don't understand question"],
+        }
+        with pm.Model(coords=COORDS) as m:
+            # the zero sum axis will be 'answers'
+    ...:    v = pm.ZeroSumNormal("v", dims=("regions", "answers"))
```
ricardoV94: I think these shouldn't be here, also below 

```suggestion
            v = pm.ZeroSumNormal("v", dims=("regions", "answers"))
```

pymc/tests/distributions/test_multivariate.py
```diff
@@ -1368,6 +1381,203 @@ def test_issue_3706(self):
         assert prior_pred["X"].shape == (1, N, 2)
 
 
+COORDS = {
+    "regions": ["a", "b", "c"],
+    "answers": ["yes", "no", "whatever", "don't understand question"],
+}
```
ricardoV94: Nitpick: these shouldn't be outside of the TestClass

pymc/distributions/__init__.py
```diff
@@ -116,8 +117,8 @@
     "Uniform",
     "Flat",
     "HalfFlat",
-    "TruncatedNormal",
```
ricardoV94, line 119: Small nitpick: your vscode also changed this without need

AlexAndorra: Ha ha no, _I_ did that because I like alphabetical order 🙈

pymc/distributions/transforms.py
```diff
@@ -39,6 +43,7 @@
     "circular",
     "CholeskyCovPacked",
     "Chain",
+    "ZeroSumTransform",
```
ricardoV94, line 46: Need to add this to the docs as well

pymc/distributions/multivariate.py
```diff
@@ -2380,3 +2383,205 @@ def logp(value, alpha, K):
...
+        For now, ``sigma`` has to be a scalar, to ensure the zero-sum constraint.
+    zerosum_axes: int, defaults to 1
+        Number of axes along which the zero-sum constraint is enforced, starting from the rightmost position.
+        Defaults to 1, i.e the rightmost axis.
+    dims: sequence of strings, optional
+        Dimension names of the distribution. Works the same as for other PyMC distributions.
+        Necessary if ``shape`` is not passed.
+    shape: tuple of integers, optional
+        Shape of the distribution. Works the same as for other PyMC distributions.
+        Necessary if ``dims`` is not passed.
```
```suggestion
        Necessary if ``dims`` or ``observed`` is not passed.
```

ricardoV94 (changes requested):

pymc/distributions/transforms.py
AlexAndorra: Done

pymc/distributions/transforms.py
```diff
@@ -266,6 +271,64 @@ def bounds_fn(*rv_inputs):
...
+    Constrains any random samples to sum to zero along the user-provided ``zerosum_axes``.
+    By default (``zerosum_axes=[-1]``), the sum-to-zero constraint is imposed
+    on the last axis.
+
+    Parameters
+    ----------
+    zerosum_axes : list of ints
+        Must be a list of integers (positive or negative).
+        By default (``zerosum_axes=[-1]``), the sum-to-zero constraint is imposed
+        on the last axis.
```
ricardoV94: This does not seem to be the case
```suggestion
```

ricardoV94 (changes requested):

pymc/distributions/transforms.py
```diff
@@ -266,6 +271,62 @@ def bounds_fn(*rv_inputs):
         super().__init__(args_fn=bounds_fn)
 
 
+class ZeroSumTransform(RVTransform):
+    """
+    Constrains any random samples to sum to zero along the user-provided ``zerosum_axes``.
+    By default (``zerosum_axes=[-1]``), the sum-to-zero constraint is imposed
+    on the last axis.
```
ricardoV94:
```suggestion
```

ricardoV94 (approved):

PR #6125, opened 2022-09-13
Finish restructuring the tests to follow the structure of the code

### LINKED ISSUE #5777: Mirror codebase structure in tests
### Discussed in https://github.com/pymc-devs/pymc/discussions/5579

<sup>Originally posted by **ricardoV94** March 11, 2022</sup>
I think the tests files should mirror the original codebase. So they should be in a file with the same name as where the functions/classes that it tests are defined. E.g.,, timeseries tests should be in tests/distributions/test_timeseries.py. This is how Aesara does it and its much easier to find what tests relate to what code, and also to spot gaps in our tests.

_Originally posted by @ricardoV94 in https://github.com/pymc-devs/pymc/pull/5298#discussion_r822280003_

This would also allow to break up the immensely large and slow `test_distributions.py`, which was already brought up in https://github.com/pymc-devs/pymc/issues/5410#issuecomment-1029759988

Also, https://github.com/pymc-devs/pymc/pull/5577#discussion_r824542298</div>

Armavica:
**What is this PR about?**
This PR restructures the last test files to follow the structure of the code, as discussed in #5777.

I think that this closes #5777. One last unresolved question is whether we want to move `tests` out of `pymc`, like it is done in aesara for example.

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Finish restructuring the tests to follow the structure of the code


Armavica:
These metropolis tests are flaky even though they are supposed to be `SeededTest`s. It is because `_get_seeds_per_chain` creates a `numpy.random.default_rng(None)` which is based on entropy, not on the global numpy random generator: https://github.com/pymc-devs/pymc/blob/4ea8ddee7b55ccaadeec2549bb8c65617e032453/pymc/sampling.py#L287-L290

My suggestion is to use a seed from the global numpy random generator in case number 4 described in this function's docstring: this makes these tests deterministic. Concretely I would add this couple of lines before L287 above:

``` python
if random_state is None:
    random_state = np.random.randint(2**30)
```

Would that be okay?

ricardoV94:
No, we officially don't allow users to control seeding via global seeding anymore (even if we still use global seeding internally), because those have less quality. 

The solution is to explicitly pass `random_seed` to the sampling routines in those tests.

But how often does it fail? I haven't seen it fail before for a long time now.

pymc/tests/helpers.py
```diff
@@ -148,3 +155,66 @@ def assert_random_state_equal(state1, state2):
         (in2out(local_check_parameter_to_ninf_switch), -1)
     )
 )
+
+
+class TestStepMethod:
```
ricardoV94: I think pytest will run these baseclass when it finds it because it starts with `Test`, no?

Armavica:
I see, that makes sense. I will change the name of the class and we will see if that repeats.

But doesn't this mean that `SeededTest` in its current form is not doing much?

ricardoV94:
> I see, that makes sense. I will change the name of the class and we will see if that repeats.
> 
> But doesn't this mean that `SeededTest` in its current form is not doing much?

Yes it's not doing much anymore, other than providing a seed when requested

.github/workflows/tests.yml
```diff
@@ -149,7 +149,7 @@ jobs:
           - pymc/tests/test_variational_inference.py pymc/tests/test_initial_point.py
           - pymc/tests/test_model.py pymc/tests/test_step.py
           - pymc/tests/gp/test_cov.py pymc/tests/gp/test_gp.py pymc/tests/gp/test_mean.py pymc/tests/gp/test_util.py pymc/tests/ode/test_ode.py pymc/tests/ode/test_utils.py pymc/tests/test_smc.py pymc/tests/test_parallel_sampling.py
-          - pymc/tests/test_sampling.py pymc/tests/test_posteriors.py
+          - pymc/tests/test_sampling.py pymc/tests/step_methods/test_metropolis.py pymc/tests/step_methods/test_slicer.py pymc/tests/step_methods/hmc/test_nuts.py
```
ricardoV94, line 152: Isn't it enough to do `-pymc/tests/step_methods/` ?

Armavica: Yes it is, but here I selected only the few files where tests from `test_posteriors` ended up, to not increase too much the runtime of the tests. The rest is already tested in other platforms. I am not sure how much of a difference that makes, though. If we test the whole of `step_methods` I will also check that the `check_all_tests_are_covered.py` script understands what happens.

Armavica:
These two tests (`TestKmeansInducing::test_kmeans` (this one is tested on both Windows and Linux on the github runners but fails only on Windows) and `TestDEMetropolisZ::test_tuning_reset`) have been consistently failing on this PR, even though they pass locally on my computer. I will check if I did something wrong.

ricardoV94:
@Armavica Do you want to try to rebase this from main and see if we can get it merged quickly? One of the flaky tests you mention was fixed recently

ricardoV94:
Otherwise it might be easier to do it one module at a time

Armavica:
The `TestDEMetropolisZ::test_tuning_reset` does seem to fail about 10-20% of the time on my laptop. Should I just add a seed?

ricardoV94:
> The `TestDEMetropolisZ::test_tuning_reset` does seem to fail about 10-20% of the time on my laptop. Should I just add a seed?

Yes

ricardoV94 (approved):

ricardoV94:
Awesome work @Armavica, are we done?

Armavica:
> Awesome work @Armavica, are we done?

Yes, that was the last batch!

PR #6126, opened 2022-09-13
Repair the plot of pm.Interpolated and add an example for pm.Deterministic

### LINKED ISSUE #6124: The docs for deterministic should include an example
Including an example and a better explanation would go a long way in helping people understand this functionality

https://docs.pymc.io/en/latest/api/generated/pymc.Deterministic.html

Armavica:
This PR closes #6124 by adding an example and explanation of `pm.Deterministic`.

It also repairs the plot of `pm.Interpolated` which was not working, and fixes some of the doc makefile commands.

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Fix `make serve`
- Fix `make html`
- Fix `make clean`
- Repair the plot of `pm.Interpolated`
- Add an explanation and example to `pm.Deterministic`

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1937,11 +1937,52 @@ def Point(*args, filter_model_vars=False, **kwargs) -> Dict[str, np.ndarray]:
 
 
 def Deterministic(name, var, model=None, dims=None, auto=False):
-    """Create a named deterministic variable
+    """Create a named deterministic variable.
+
+    Deterministic nodes are only deterministic given all of their inputs, i.e.
+    they don't add randomness to the model.  They are generally used to record
+    an intermediary result.
+
+    Indeed, PyMC allows for arbitrary combinations of random variables, for
+    example in the case of a logistic regression
+
+    .. code:: python
+
+        with pm.Model():
+            alpha = pm.Normal("alpha", 0, 1)
+            intercept = pm.Normal("intercept", 0, 1)
+            p = pm.math.invlogit(alpha * x + intercept)
+            outcome = pm.Bernoulli("outcome", p, observed=outcomes)
+
+
+    but doesn't memorize the fact that the expression ``pm.math.invlogit(alpha *
+    x + intercept)`` has been affected to the variable ``p``.  If the quantity
+    ``p`` is important and one would like to track its value in the sampling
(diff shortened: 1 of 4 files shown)


pymc/model.py
```diff
@@ -1937,11 +1937,45 @@ def Point(*args, filter_model_vars=False, **kwargs) -> Dict[str, np.ndarray]:
...
+    .. code:: python
+
+        with pm.Model():
+            alpha = pm.Normal("alpha", 0, 1)
+            intercept = pm.Normal("intercept", 0, 1)
+            p = pm.math.invlogit(alpha * x + intercept)
+            outcome = pm.Bernoulli("outcome", p, observed=outcomes)
+
+
+    but cannot memorize the fact that the expression ``pm.math.invlogit(alpha *
```
ricardoV94: It can memorize but chooses not to because memory/ speed optimizations can be done when intermediate computations don't have to be returned.

Also worth mentioning that deterministics will only be computed once per step, whereas the [d]logp graph without them may be computed 1000x times during a single NUTS step

Armavica: Thank you for the precision! I will try to improve the text.

> Also worth mentioning that deterministics will only be computed once per step, whereas the [d]logp graph without them may be computed 1000x times during a single NUTS step

I am sorry but I am not sure that I understand what you mean. Does this imply something in terms of sampling efficiency?

ricardoV94: I meant that we never compute Deterministics within a step (where we might evaluate the graph 1000x times), only at the end of a step once

Armavica: Ah yes that makes sense, thank you!

So, can I say that adding `Deterministic`s doesn't affect the sampling efficiency? If I understand correctly, the computation graph can be optimized in the same way with or without a `Deterministic`, which is not on the hot path but just computed once at the end of the step, from the values of the other RVs. Is that correct?

ricardoV94: Yes!

ricardoV94 (approved):

ricardoV94:
My review concerns the changes in docstrings, not the changes to the makerfile

OriolAbril (approved):

PR #6128, opened 2022-09-14
Fix bug in which TruncatedNormal returns -inf for all values if any value is out of bounds

adrn:
**What is this PR about?**

With pymc v4.1.7, I found that evaluating `TruncatedNormal`'s `logp` with an array of values was returning all `-inf` values -- for example:
```python
import numpy as np
import pymc as pm

p = {'mu': 1, 'sigma': 0.1}
with pm.Model() as model:    
    dist1 = pm.TruncatedNormal.dist(**p, lower=0.)
    dist2 = pm.Normal.dist(**p)
    
    x_grid = np.linspace(-5, 10, 1024)
    dist1_logp = pm.Deterministic('dist1', pm.logp(dist1, x_grid))
    dist2_logp = pm.Deterministic('dist2', pm.logp(dist2, x_grid))

func1 = model.compile_fn(dist1_logp, inputs=[])
func2 = model.compile_fn(dist2_logp, inputs=[])

print(func1({}))
print(func2({}))
```
Output:
```
[-inf -inf -inf ... -inf -inf -inf]
[-1798.61635344 -1789.8294493  -1781.06404481 ... -4022.26639085
 -4035.43062232 -4048.61635344]
```
In the output above, the first line should not be `-inf` everywhere, as the grid we evaluate on includes values in the allowed range of values.

With @ricardoV94's help, we tracked this down to the way that `TruncatedNormal.logp` was enforcing the value bounds:
https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/continuous.py#L779

I noticed [this comment](https://github.com/pymc-devs/pymc/blob/7af102d40f7b64184cd0fa013ced8570772fc8eb/pymc/distributions/dist_math.py#L59) in the `check_parameters()` docstring: "Note that check_parameter should not be used to enforce the logic of the logp expression under the normal parameter support as it can be disabled by the user via check_bounds = False in pm.Model()" and indeed the above example works as expected with `check_bounds=False`.

This PR instead follows the implementation in other truncated distributions, for example, [HalfStudentT](https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/continuous.py#L2739) to use a `switch` statement instead. I also added a regression test for the example case above.

See https://discourse.pymc.io/t/truncatednormal-logp-returning-all-inf/10398 for more context.

**Checklist**

## Major / Breaking Changes
- n/a

## Bugfixes / New features
- Fixed a bug in which `TruncatedNormal` would return -inf for all logp values if any input value was outside of the bounds.

## Docs / Maintenance
- n/a

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -777,11 +777,13 @@ class TruncatedNormal(BoundedContinuous):
             norm = 0.0
 
         logp = _logprob(normal, (value,), None, None, None, mu, sigma) - norm
+        logp = at.switch(
+            at.or_(at.le(value, lower), at.ge(value, upper)),
+            -np.inf,
+            logp,
+        )
+
         bounds = []
-        if not unbounded_lower:
-            bounds.append(value >= lower)
-        if not unbounded_upper:
-            bounds.append(value <= upper)
         if not unbounded_lower and not unbounded_upper:
             bounds.append(lower <= upper)
         return check_parameters(logp, *bounds)
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -760,7 +760,9 @@ class TruncatedNormal(BoundedContinuous):
         -------
         TensorVariable
         """
-        is_lower_bounded = not (isinstance(lower, TensorConstant) and np.all(np.isneginf(lower.value)))
+        is_lower_bounded = not (
+            isinstance(lower, TensorConstant) and np.all(np.isneginf(lower.value))
+        )
         is_upper_bounded = not (isinstance(upper, TensorConstant) and np.all(np.isinf(upper.value)))
 
         if is_lower_bounded and is_upper_bounded:
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -778,7 +778,7 @@ class TruncatedNormal(BoundedContinuous):
 
         logp = _logprob(normal, (value,), None, None, None, mu, sigma) - norm
         logp = at.switch(
-            at.or_(at.le(value, lower), at.ge(value, upper)),
+            at.or_(at.lt(value, lower), at.gt(value, upper)),
             -np.inf,
             logp,
         )
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -760,33 +760,38 @@ class TruncatedNormal(BoundedContinuous):
         -------
         TensorVariable
         """
-        unbounded_lower = isinstance(lower, TensorConstant) and np.all(lower.value == -np.inf)
-        unbounded_upper = isinstance(upper, TensorConstant) and np.all(upper.value == np.inf)
+        is_lower_bounded = not (isinstance(lower, TensorConstant) and np.all(lower.value == -np.inf))
+        is_upper_bounded = not (isinstance(upper, TensorConstant) and np.all(upper.value == np.inf))
 
-        if not unbounded_lower and not unbounded_upper:
+        if is_lower_bounded and is_upper_bounded:
             lcdf_a = normal_lcdf(mu, sigma, lower)
             lcdf_b = normal_lcdf(mu, sigma, upper)
             lsf_a = normal_lccdf(mu, sigma, lower)
             lsf_b = normal_lccdf(mu, sigma, upper)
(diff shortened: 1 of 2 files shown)


pymc/distributions/continuous.py
```diff
@@ -777,11 +777,13 @@ def logp(
             norm = 0.0
 
         logp = _logprob(normal, (value,), None, None, None, mu, sigma) - norm
+        logp = at.switch(
+            at.or_(at.le(value, lower), at.ge(value, upper)),
```
ricardoV94: I think the bounds are part of the domain:

```suggestion
            at.or_(at.lt(value, lower), at.gt(value, upper)),
```

pymc/tests/distributions/test_continuous.py
```diff
@@ -108,6 +108,18 @@ def test_lower_bounded(self):
...
+        with pm.Model() as model:
+            dist = pm.TruncatedNormal(
+                'bounded_both', mu=1, sigma=2, lower=0, upper=3
+            )
+            value = pm.Deterministic('value', pm.logp(dist, [-2., 1., 4.]))
+
+        values = model.compile_fn(value)({})
+        assert np.isinf(values[0])
+        assert np.isfinite(values[1])
+        assert np.isinf(values[2])
```
ricardoV94: We don't need a model for this:

```suggestion
        dist = pm.TruncatedNormal.dist(mu=1, sigma=2, lower=0, upper=3)
        logp = pm.logp(dist, [-2., 1., 4.]).eval()
        assert np.isinf(logp[0])
        assert np.isfinite(logp[1])
        assert np.isinf(logp[2])
```

Also can you just put this inside this function: https://github.com/pymc-devs/pymc/blob/c53cd2f5653f46c59206c66acd1fc1bc0ebf9e77/pymc/tests/distributions/test_continuous.py#L828

This class is about the transform more generally, not TruncatedNormal in specific

ricardoV94 (changes requested):

pymc/distributions/continuous.py
adrn: Good catch!

pymc/tests/distributions/test_continuous.py
adrn: Ah, right - moved it!

pymc/distributions/continuous.py
```diff
@@ -777,11 +777,13 @@ def logp(
             norm = 0.0
 
         logp = _logprob(normal, (value,), None, None, None, mu, sigma) - norm
+        logp = at.switch(
```
ricardoV94: Actually to be equivalent to what we had before we should do something like this: 

https://github.com/pymc-devs/pymc/blob/c53cd2f5653f46c59206c66acd1fc1bc0ebf9e77/pymc/distributions/truncated.py#L316-L320

ricardoV94 (changes requested):

pymc/distributions/continuous.py
adrn: Isn't that equivalent to what is implemented here because of the default values for lower and upper?
https://github.com/pymc-devs/pymc/blob/91cbebda4890b36ceae12855ec35e59312f38de7/pymc/distributions/continuous.py#L714-L715

ricardoV94: We retrieve the None case here: https://github.com/pymc-devs/pymc/blob/91cbebda4890b36ceae12855ec35e59312f38de7/pymc/distributions/continuous.py#L763-L764

So in those cases we avoid introducing the useless switch. It's a small optimization but I don't see any reason yo modify it.

adrn: Oh I see. OK

adrn: The last commit makes the implementation here more analogous to the general truncated case you linked above - thanks for that pointer!

adrn: Thanks @ricardoV94 for taking a look and helping out already! Let me know if the implementation in the latest few commits looks ok. Also, it looks like this PR is waiting for approval to run the full test suite.

ricardoV94:
Pre-commit is complaining, otherwise looks good

adrn:
Cool - I forgot to `pre-commit install` so just ran it manually and pushed up the changes. I think the workflows need to be approved to run again?

ricardoV94 (approved):

ricardoV94:
Thanks @adrn !

PR #6131, opened 2022-09-15
Refactor Multivariate RandomWalk distributions

ricardoV94:
Related to https://github.com/pymc-devs/pymc/issues/4642

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- Extend `RandomWalk` to multivariate cases

## Docs / Maintenance
- Refactor `MvGaussianRandomWalk` and `MvStudentTRandomWalk` for V4

### DIFF
--- a/pymc/distributions/shape_utils.py
+++ b/pymc/distributions/shape_utils.py
@@ -20,7 +20,7 @@ samples from probability distributions for stochastic nodes in PyMC.
 import warnings
 
 from functools import singledispatch
-from typing import Any, Optional, Sequence, Tuple, Union
+from typing import Any, Optional, Sequence, Tuple, Union, cast
 
 import numpy as np
 
@@ -671,7 +671,7 @@ def get_support_shape(
     observed: Optional[Any] = None,
     support_shape_offset: Sequence[int] = None,
     ndim_supp: int = 1,
-):
+) -> Optional[TensorVariable]:
     """Extract the support shapes from shape / dims / observed information
 
     Parameters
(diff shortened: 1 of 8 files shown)


pymc/distributions/shape_utils.py
```diff
@@ -702,45 +702,54 @@ def get_support_shape(
...
-            ]
-        )
-
+        inferred_support_shape = [
+            Assert(msg="support_shape does not match respective shape dimension")(
+                inferred, at.eq(inferred, explicit)
+            )
+            for inferred, explicit in zip(inferred_support_shape, support_shape)
+        ]
+    if inferred_support_shape is not None:
```
junpenglao: This condition could be removed as the previous control flow makes sure inferred_support_shape is not None

pymc/distributions/shape_utils.py
```diff
@@ -702,45 +702,54 @@ def get_support_shape(
...
+        if len(dims) < ndim_supp:
+            raise ValueError(f"Number of dims is too small for ndim_supp of {ndim_supp}")
         model = modelcontext(None)
-        inferred_support_shape = at.stack(
-            [
-                model.dim_lengths[dims[i]] - support_shape_offset[i]  # type: ignore
-                for i in np.arange(-ndim_supp, 0)
-            ]
-        )
+        inferred_support_shape = [
```
junpenglao, line 726: IIUC, if user supplies both `dims` and `observed`, we will actually use `observed` for inferred_support_shape and ignore `dims`.

While I think it is fine as is, it might be worth to raise a warning in those cases that `dims` is ignore (or a warning only if there is a mismatch)

pymc/distributions/timeseries.py
```diff
@@ -69,95 +72,156 @@ class RandomWalk(Distribution):
...
+            observed=None,
+        )
+        if steps is None:
+            raise ValueError("Must specify steps or shape parameter")
+        steps = at.as_tensor_variable(intX(steps))
 
         return super().dist([init_dist, innovation_dist, steps], **kwargs)
 
+    @classmethod
+    def get_steps(cls, innovation_dist, steps, shape, dims, observed):
```
junpenglao, line 123: Any particular reason this is a class method instead of just a function?

pymc/distributions/shape_utils.py
ricardoV94: When there are both dims and observed we give preference do dims: https://github.com/pymc-devs/pymc/blob/305cb5f62c8e2a00a6109719d81406f7b1ce141b/pymc/distributions/distribution.py#L284-L288

pymc/distributions/timeseries.py
ricardoV94: Because it seems only relevant for RandomWalk for now

pymc/distributions/shape_utils.py
ricardoV94: Don't know if this comment got stale, but I think this condition is still needed. There are cases where the output of this function is `None`

pymc/distributions/shape_utils.py
junpenglao: Right, after the 1st condition `inferred_support_shape` is no longer `None`.

junpenglao (approved):

pymc/distributions/timeseries.py
```diff
@@ -69,95 +72,156 @@ class RandomWalk(Distribution):
...
             and init_dist.owner is not None
             and isinstance(init_dist.owner.op, (RandomVariable, SymbolicRandomVariable))
-            # TODO: Lift univariate constraint on init_dist
-            and init_dist.owner.op.ndim_supp == 0
         ):
-            raise TypeError("init_dist must be a univariate distribution variable")
+            raise TypeError("init_dist must be a distribution variable")
         check_dist_not_registered(init_dist)
 
         if not (
```
larryshamalama, line 96: Can this check be a helper function? I feel like we often check these conditions

pymc/distributions/timeseries.py
```diff
@@ -174,7 +238,25 @@ def random_walk_logp(op, values, *inputs, **kwargs):
...
+
+    @classmethod
+    def dist(cls, *args, **kwargs) -> at.TensorVariable:
+        init_dist, innovation_dist, kwargs = cls.get_dists(*args, **kwargs)
+        return RandomWalk.dist(init_dist=init_dist, innovation_dist=innovation_dist, **kwargs)
+
+    @classmethod
+    @abc.abstractmethod
+    def get_dists(cls, *args, **kwargs):
+        pass
```
larryshamalama, line 256: raise `NotImplementedError` here maybe?

larryshamalama (commented):
Great PR @ricardoV94, as always! 

It seems like `get_dists` does a lot of the hard work for the `dist` method, but why couldn't it be a part of `RandomWalk`? I am guessing that the answer is inherent to metaclasses but I just can't pinpoint it right now.

pymc/distributions/timeseries.py
ricardoV94: abc.abstractmethod does that ;)

pymc/distributions/timeseries.py
ricardoV94: Yeah I always think about it. It's on the edge of being so short that it doesn't make complete sense. Have to think about it

pymc/distributions/timeseries.py
larryshamalama: ooo

PR #6132, opened 2022-09-15
Add `start_sigma` to ADVI 2

markusschmaus:
Recreation of https://github.com/pymc-devs/pymc/pull/6096 after failing test (fixed in latest commit).

Introduces the `start_sigma` argument which allows to set a starting value for the sigmas of mean field approximation in `ADVI`. I am using the `ordering` property to create the mapping between the variables and the flat 1d array required by `ADVI`. 

See also:
https://discourse.pymc.io/t/quality-of-life-improvements-to-advi/10254

**Checklist**

## Major / Breaking Changes
- None

## Bugfixes / New features
- Allow `start_sigma` being given to `fit`.
- ~~For `ASVGD` forward the key word arguments `start` and `random_seed` to the constructor of the default approximation `FullRank` instead of sending them to `super().__init__` which can't process them~~
- Enable `_iterate_with_loss` to run with `n=0` by giving an appropriate logger message. This allows to check initialization values, which is useful for testing, but could also be useful for debugging a user model

## Docs / Maintenance
- The documentation for the `start` argument was outdated, it claimed type `Point`, but it gets passed through to a function that requires `StartDict`. Since users are probably unfamiliar with this custom type (`Dict[Union[Variable, str], Union[np.ndarray, Variable, str]]`), I decided to also mention the most relevant and most commonly used subtype `dict[str, np.ndarray]`.

### DIFF
--- a/pymc/tests/test_variational_inference.py
+++ b/pymc/tests/test_variational_inference.py
@@ -571,6 +571,28 @@ def test_fit_oo(inference, fit_kwargs, simple_model_data):
     np.testing.assert_allclose(np.std(trace.posterior["mu"]), np.sqrt(1.0 / d), rtol=0.2)
 
 
+def test_fit_start(inference_spec, simple_model):
+    mu_init = 17
+    mu_sigma_init = 13
+
+    with simple_model:
+        if type(inference_spec()) == ADVI:
+            has_start_sigma = True
+        else:
+            has_start_sigma = False
+
+    kw = {"start": {"mu": mu_init}}
+    if has_start_sigma:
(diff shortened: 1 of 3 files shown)


junpenglao (approved):
Thank you!

markusschmaus:
> - For `ASVGD` forward the key word arguments `start` and `random_seed` to the constructor of the default approximation `FullRank` instead of sending them to `super().__init__` which can't process them

For the record, this is what caused the test to fail since it expected `super().__init__` to raise a `TypeError` when it receives a `start` parameter.

PR #6140, opened 2022-09-20
Bump Aesara and Aeppl dependencies

ricardoV94:


twiecki (approved):

twiecki:
Test failures :(.

ricardoV94:
> Test failures :(.

Not super shocked, this one changed quite some fundamentals

ricardoV94:
Bug should be fixed in https://github.com/aesara-devs/aesara/pull/1199

Other bug still unclear: https://github.com/aesara-devs/aesara/issues/1200

ricardoV94:
@Armavica I reverted a commit of yours as it had invalid logic, feel free to open a new PR and I can leave specific feedback there.

pymc/distributions/continuous.py
```diff
@@ -3994,9 +3994,9 @@ def make_node(self, x, h, z):
         x = at.as_tensor_variable(floatX(x))
         h = at.as_tensor_variable(floatX(h))
         z = at.as_tensor_variable(floatX(z))
-        bshape = broadcast_shape(x, h, z)
-        shape = [False] * len(bshape)
-        return Apply(self, [x, h, z], [at.TensorType(aesara.config.floatX, shape=shape)()])
+        shape = broadcast_shape(x, h, z)
+        broadcastable = [] if not shape else [False] * len(shape)
```
ricardoV94, line 3998: broadcastable of `False` corresponds to shape of `None`

pymc/smc/smc.py
```diff
@@ -565,7 +565,7 @@ def _logp_forw(point, out_vars, in_vars, shared):
         new_in_vars = []
         for in_var in in_vars:
             if in_var.dtype in discrete_types:
-                float_var = at.TensorType("floatX", in_var.shape)(in_var.name)
+                float_var = at.TensorType("floatX", in_var.broadcastable)(in_var.name)
```
ricardoV94, line 568: Should have been `n_var.type.shape`

pymc/distributions/continuous.py
Armavica: I understand that the intent is more clearly conveyed with `None`, but isn't this conversion handled [here](https://github.com/aesara-devs/aesara/blob/main/aesara/tensor/type.py#L80-L82)?
I am happy to redo this commit anyway with integers/Nones instead of bools whenever possible, but I just wanted to check with you if I am understanding the logic correctly.

ricardoV94: That's a workaround because types used to be described with True/False flags before shape came around. I think we will eventually deprecate the True/ False, so it's better to do it already.

The other issue was with using var.shape instead of var.type.shape which are not the same usually (one can be a symbolic expression  and the other is always a tuple). In contrast, var.broadcastable and var.type.broadcastable were always the same so it's an understandable mistake.

PR #6145, opened 2022-09-25
Gp sigma parameterize 6094

williambdean:
**What is this PR about?**
Addressing the issue raised: https://github.com/pymc-devs/pymc/issues/6094

Helps migrates toward standardized API for noise in GPs. Changes the `noise` parameter to `sigma` where applicable.  A `FutureWarning` is used if `noise` parameter is still used in the previous locations.

**Checklist**

## Major / Breaking Changes
- NA

## Bugfixes / New features
- Standardize the gp API

## Docs / Maintenance
- GP docstring examples all work and use `sigma` parameter instead of `noise`
- Small changes to the Latent GP docs that mistakenly reference the `noise`

### DIFF
--- a/pymc/gp/gp.py
+++ b/pymc/gp/gp.py
@@ -37,6 +37,27 @@ from pymc.math import cartesian, kron_diag, kron_dot, kron_solve_lower, kron_sol
 __all__ = ["Latent", "Marginal", "TP", "MarginalApprox", "LatentKron", "MarginalKron"]
 
 
+_noise_deprecation_warning = (
+    "The 'noise' parameter has been been changed to 'sigma' "
+    "in order to standardize the gp api and will be "
+    "deprecated in future releases."
+)
+def _handle_sigma_noise_parameters(sigma, noise): 
+    """Helper function for transition of 'noise' parameter to be named 'sigma'."""
+
+    if (sigma is None and noise is None) or (sigma is not None and noise is not None): 
+        raise ValueError("'sigma' argument must be specified.")
+
+    if sigma is None: 
+        warnings.warn(
+            _noise_deprecation_warning, 
+            FutureWarning
+        )
+        return noise
+
+    return sigma 
+
+
 class Base:
     R"""
     Base class.
@@ -218,7 +239,7 @@ class Latent(Base):
         Xnew: array-like
             Function input values.
         given: dict
-            Can optionally take as key value pairs: `X`, `y`, `noise`,
+            Can optionally take as key value pairs: `X`, `y`,
             and `gp`.  See the section in the documentation on additive GP
             models in PyMC for more information.
         jitter: scalar
@@ -359,7 +380,7 @@ class TP(Latent):
         return pm.MvStudentT(name, nu=nu2, mu=mu, cov=cov, **kwargs)
 
 
-@conditioned_vars(["X", "y", "noise"])
+@conditioned_vars(["X", "y", "sigma"])
 class Marginal(Base):
     R"""
     Marginal Gaussian process.
@@ -393,7 +414,7 @@ class Marginal(Base):
 
             # Place a GP prior over the function f.
             sigma = pm.HalfCauchy("sigma", beta=3)
-            y_ = gp.marginal_likelihood("y", X=X, y=y, noise=sigma)
+            y_ = gp.marginal_likelihood("y", X=X, y=y, sigma=sigma)
 
         ...
 
@@ -405,15 +426,15 @@ class Marginal(Base):
             fcond = gp.conditional("fcond", Xnew=Xnew)
     """
 
-    def _build_marginal_likelihood(self, X, noise, jitter):
+    def _build_marginal_likelihood(self, X, sigma, jitter):
         mu = self.mean_func(X)
         Kxx = self.cov_func(X)
-        Knx = noise(X)
+        Knx = sigma(X)
         cov = Kxx + Knx
         return mu, stabilize(cov, jitter)
 
     def marginal_likelihood(
-        self, name, X, y, noise, jitter=JITTER_DEFAULT, is_observed=True, **kwargs
+        self, name, X, y, sigma=None, noise=None, jitter=JITTER_DEFAULT, is_observed=True, **kwargs
     ):
         R"""
         Returns the marginal likelihood distribution, given the input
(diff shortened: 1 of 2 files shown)


williambdean:
Previous tests work on my end with the expect warnings. Working through specific tests soon for warnings

williambdean:
The implementation should be at the point initially described. Namely, 
- switch logic and docs to use `sigma` parameter 
- allow for `noise` parameter until more formally deprecated. Have a warning to indicate

Think the tests are in a fine place as well. Was thinking of having one to compare the model parameters with the `noise` vs `sigma` parameterization but might be overkill way to test handling the parameters. Could just test function if desired. 
Also, put the tests into a separate spot so they can be removed at some point.

Noticed a failing test on one of the windows runs from before. Don't know too much about that at the moment. It doesn't seem related to the changes I've made at first glance. Not sure if anyone has thoughts here.

pymc/gp/gp.py
```diff
@@ -37,6 +37,26 @@
 __all__ = ["Latent", "Marginal", "TP", "MarginalApprox", "LatentKron", "MarginalKron"]
 
 
+_noise_deprecation_warning = (
+    "The 'noise' parameter has been been changed to 'sigma' "
+    "in order to standardize the gp api and will be "
```
bwengals: nit: capitalize acronyms (GP API)

bwengals:
Thanks for the contribution!  I think from a user standpoint, it makes sense to have all functions take `sigma`.  

For some background, the previous reason why one GP implementation was  `noise` and the other was `sigma` was because in `gp.Marginal` you can pass in any covariance function for the noise, white $\sigma^2 \mathbf{I}$ or otherwise.  But `gp.MarginalApprox` only handles white noise, so it's input was specifically $\sigma$ and not `noise` more generally.

That said, while it makes sense to have more simplicity for users to input `sigma` (and say the type can be an arbitrary covariance matrix) it is somewhat incorrect to have things like `Knx = sigma(X)` in the code, since one would reasonably assume `sigma` was a scalar and not an arbitrary covariance function.  What about changing the name to `noise_func` (or similar) to go with `cov_func` internally?

danhphan:
Hi @wd60622 Thank you for the PR, and thanks @bwengals for the comments.

I will have a look in more details at this on tomorrow afternoon. Also, sorry for my late response!

williambdean:
Great. Thank you both for taking a look. 
I like your suggestions @bwengals. Agree with the `noise_func` internal. Good suggestion for clarity. 

I will make the adjustments a bit later today

pymc/gp/gp.py
williambdean: done

williambdean:
@bwengals Changes implemented. Went with `noise_func` alternative. I like that suggestion and added wherever the previous `sigma` was being called. That transformed `noise_func` is still being saved in the `sigma` attribute as that is the argument and the name in the decorator. Tried to stay as true to the previous logic as possible. But feels a little weird to me as it is transformed. Let me know if you have any other comments on this implementation.

bwengals:
This looks great! For the decorator, for using `.conditional` with `given`, the user still specifies it as `sigma` right?  I think as long as the user doesn't see or interact with `noise_func` it works well.

williambdean:
> This looks great! For the decorator, for using `.conditional` with `given`, the user still specifies it as `sigma` right? I think as long as the user doesn't see or interact with `noise_func` it works well.

Yup, correct. Still `sigma` being passed / is the main attribute being used so the decorator is appropriate. 
Just my thought there is that `sigma` will be transformed into `Covariance` (`WhiteNoise`) instance and being passed around under the name `noise_func` even though it is stored in `self.sigma`. 
But a user could pass an instance of that class in the first place so I guess it is not that weird to do. Bit thinking out loud here 😄 

Just waiting on these tests and will decode any fails when they are done. They have been successful in my the local checks.

pymc/gp/gp.py
```diff
@@ -37,6 +37,26 @@
...
+    "in order to standardize the GP API and will be "
+    "deprecated in future releases."
+)
+
+
+def _handle_sigma_noise_parameters(sigma, noise):
+    """Helper function for transition of 'noise' parameter to be named 'sigma'."""
+
+    if (sigma is None and noise is None) or (sigma is not None and noise is not None):
+        raise ValueError("'sigma' argument must be specified.")
```
danhphan, line 51: Hi, just want to understand the logic: If `sigma is not None and noise is not None`, should we raise this error, as we already specified `sigma` in this case?

danhphan: Maybe, the logic here could be:
```
if `sigma is not None`: good
else:
      if `noise is not None`: raise warning
      else: raise error.
```

OR
```
if `sigma is None`:
   if `noise is None`: raise error
   else: raise warning
```

pymc/gp/gp.py
```diff
@@ -218,7 +238,7 @@ def conditional(self, name, Xnew, given=None, jitter=JITTER_DEFAULT, **kwargs):
         Xnew: array-like
             Function input values.
         given: dict
-            Can optionally take as key value pairs: `X`, `y`, `noise`,
+            Can optionally take as key value pairs: `X`, `y`,
```
danhphan, line 241: Do we need to replace `noise` by `sigma` here?

pymc/gp/gp.py
danhphan: If both `sigma` and `noise` are specified, it is not really wrong, right? :) The user just provides more inputs than what is needed.

danhphan (commented):
Hi, thanks @wd60622 

Look good to me. I just add a couple of small comments for clarification.

pymc/gp/gp.py
williambdean: I hear what you are staying but thought it might be fine to nudge them toward the new input. Not likely for both to be used, but thought being explicit about the intended behavior would be best instead of throwing away or checking the two parameters silently.

pymc/gp/gp.py
williambdean: This was for the latent class that doesn't use noise. I believe this was a mistake to begin with. Let me know if I am wrong

bwengals: Yes @wd60622 you are right! thanks to you both for catching this I didn't notice it....

pymc/gp/gp.py
bwengals: If the user passes in sigma=0.1 and noise=0.2, it's not really possible from the outside to know which value is going to be used in the model.  The user must be a bit confused then, and an error message would help them out.  Both your alternatives for doing the logic make sense to me too though @danhphan, but I think it also makes sense how it is.

williambdean: Well said @bwengals That's what I had in mind. 
Was trying to make it easier from the developer's point of view

pymc/gp/gp.py
danhphan: Great, thanks for pointing it out :)

pymc/gp/gp.py
danhphan: Hi, thanks @wd60622 and @bwengals for the clarification. Total agree with "nudge users toward the new sigma" :)

danhphan (approved):
Many thanks @wd60622 for the contribution!

bwengals (approved):
LGTM

PR #6147, opened 2022-09-26
Check shared variable values to determine volatility in posterior predictive sampling

### LINKED ISSUE #6047: Variables with shared inputs are always resampled from the prior in sample_posterior_predictive

Using the model.add_coord method appears to break pm.sample_posterior_predictive. 

Here’s a simple example, estimating the loc and scale of three independent normals:

```python
# Generate data
data = np.random.normal(loc=np.array([3, 5, 8]), scale=np.array([1.1, 6.3, 9.1]), size=(1000, 3))

# Model 1: No coords
with pm.Model() as no_coords_model:
    mu = pm.Normal('mu', mu=0, sigma=10, size=3)
    sigma = pm.HalfNormal('sigma', sigma=10, size=3)
    
    ll = pm.Normal('obs', mu=mu, sigma=sigma, observed=data)
    no_coords_trace = pm.sample()
    no_coords_post = pm.sample_posterior_predictive(no_coords_trace)

# Model 2: Context manager
coords = {'name': ['A', 'B', 'C']}
with pm.Model(coords=coords) as context_model:
    mu = pm.Normal('mu', mu=0, sigma=10, dims=['name'])
    sigma = pm.HalfNormal('sigma', sigma=10, dims=['name'])
    ll = pm.Normal('obs', mu=mu, sigma=sigma, observed=data)
    
    context_trace = pm.sample()
    context_post = pm.sample_posterior_predictive(context_trace)

# Model 3: Within model
with pm.Model() as within_model:
    within_model.add_coord('name', ['A', 'B', 'C'], mutable=True)
    mu = pm.Normal('mu', mu=0, sigma=10, dims=['name'])
    sigma = pm.HalfNormal('sigma', sigma=10, dims=['name'])
    
    ll = pm.Normal('obs', mu=mu, sigma=sigma, observed=data)
    within_trace = pm.sample()
    within_post = pm.sample_posterior_predictive(within_trace)



traces = [no_coords_trace, context_trace, within_trace]
mus = [trace.posterior.mu.values[..., i].mean() for trace in traces for i in range(3)]
sigmas = [trace.posterior.sigma.values[..., i].mean() for trace in traces for i in range(3)]
post_df = pd.DataFrame(np.c_[mus, sigmas], columns=['mu', 'sigma'], index=pd.MultiIndex.from_product([['no_coords', 'context', 'within'], ['A', 'B', 'C']]))
print(post_df.unstack(1).to_string())

                 mu                         sigma                    
                  A         B         C         A         B         C
context    2.977460  4.982624  7.826642  1.081710  6.287514  9.165928
no_coords  2.976785  4.984743  7.827109  1.081657  6.289910  9.174939
within     2.976568  4.990646  7.825051  1.081552  6.286198  9.167916


pps = [no_coords_post, context_post, within_post]
mean_value = [post.posterior_predictive.obs.values[..., i].mean() for post in pps for i in range(3)]
post_df = pd.DataFrame(mean_value, columns=['mean_ppc'], index=pd.MultiIndex.from_product([['no_coords', 'context', 'within'], ['A', 'B', 'C']]))

           mean_ppc                    
                  A         B         C
context    2.977167  4.985852  7.825006
no_coords  2.976837  4.982244  7.818495
within    -0.045788 -0.594845 -0.270400


```

"The dims on within_post are the same as the others, but it seems like totally wrong values are getting sampled. It is telling that the mean of each distribution is not the mean of the means, suggesting it’s not a case of correct values being shuffled."

This is pretty breaking when attempting to do out-of-sample predictions, since coords needs to be set somehow, and it can't (to my knowledge) be re-added to the model context after instantiation. 

* PyMC/PyMC3 Version: 4.1.3
* Aesara/Theano Version: 2.7.7
* Python Version: 3.8
* Operating system: Mac OS
* How did you install PyMC/PyMC3: pip

lucianopaz:
**What is this PR about?**

Closes #6047.

This PR did two things:

1. It makes the mutable dimension `SharedVariable` length have the name of the corresponding dimension.
2. Adds two arguments to `compile_forward_sampling_function`: `constant_data` and `constant_coords`.

These two arguments allow it to determine if a `SharedVariable` has changed after between a call to `sample` and `sample_posterior_predictive`. We have to note that `constant_data` is only knowable if `sample_posterior_predictive` is called supplying an `InferenceData` object, and `constant_coords` are only knowable if `sample_posterior_predictive` is called supplying an `InferenceData` or an `xarray.Dataset` object.

The way `sample_posterior_predictive` is able to determine if a data container changed is by doing the following. It first checks the `InferenceData.constant_data` group to find the values of data containers at inference time, and passes those into the `constant_data` argument of `compile_forward_sampling_function`. When a `SharedVariable` is found while walking the graph, it looks up the entry in `constant_data`, it it finds an entry, it checks whether the values in the dictionary match the values of the `SharedVariable` at run time. If they match, the `SharedVariable` is deemed not volatile, if they don't match, the `SharedVariable` is considered volatile.
To check if a dimension's coordinates changed, `sample_posterior_predictive` compares the model's coordinates to those found in the supplied trace (must be an `InferenceData` or `xarray.Dataset` to have this information). If the coordinates did not change, then the dimension name is added to the `constant_coords` set. Then, when `compile_forward_sampling_function` finds the dimension length's shared variable, it tries to see if its name is in `constant_coords`. If it is, then the dimension is not deemed volatile, if it isn't it is considered volatile.

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- `sample_posterior_predictive`, when supplied with an `InferenceData` object, properly identifies if a `MutableData` or mutable dimension has changed between a call to `pymc.sample` and `pymc.sample_posterior_predictive`. If they have, then the descendant random variables are resampled, if they have not changed, then the descendant random variables are taken from the `InferenceData.posterior`.

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/sampling.py
+++ b/pymc/sampling.py
@@ -1621,6 +1621,8 @@ def compile_forward_sampling_function(
     vars_in_trace: List[Variable],
     basic_rvs: Optional[List[Variable]] = None,
     givens_dict: Optional[Dict[Variable, Any]] = None,
+    constant_data: Optional[Dict[str, np.ndarray]] = None,
+    constant_coords: Optional[Set[str]] = None,
     **kwargs,
 ) -> Tuple[Callable[..., Union[np.ndarray, List[np.ndarray]]], Set[Variable]]:
     """Compile a function to draw samples, conditioned on the values of some variables.
@@ -1634,18 +1636,18 @@ def compile_forward_sampling_function(
     compiled function or after inference has been run. These variables are:
 
     - Variables in the outputs list
-    - ``SharedVariable`` instances that are not ``RandomStateSharedVariable`` or ``RandomGeneratorSharedVariable``
+    - ``SharedVariable`` instances that are not ``RandomStateSharedVariable`` or ``RandomGeneratorSharedVariable``, and whose values changed with respect to what they were at inference time
     - Variables that are in the `basic_rvs` list but not in the ``vars_in_trace`` list
     - Variables that are keys in the ``givens_dict``
     - Variables that have volatile inputs
 
     Concretely, this function can be used to compile a function to sample from the
     posterior predictive distribution of a model that has variables that are conditioned
-    on ``MutableData`` instances. The variables that depend on the mutable data will be
-    considered volatile, and as such, they wont be included as inputs into the compiled function.
-    This means that if they have values stored in the posterior, these values will be ignored
-    and new values will be computed (in the case of deterministics and potentials) or sampled
-    (in the case of random variables).
+    on ``MutableData`` instances. The variables that depend on the mutable data that have changed
+    will be considered volatile, and as such, they wont be included as inputs into the compiled
+    function. This means that if they have values stored in the posterior, these values will be
+    ignored and new values will be computed (in the case of deterministics and potentials) or
+    sampled (in the case of random variables).
 
     This function also enables a way to impute values for any variable in the computational
     graph that produces the desired outputs: the ``givens_dict``. This dictionary can be used
@@ -1672,6 +1674,25 @@ def compile_forward_sampling_function(
         A dictionary that maps tensor variables to the values that should be used to replace them
         in the compiled function. The types of the key and value should match or an error will be
         raised during compilation.
+    constant_data : Optional[Dict[str, numpy.ndarray]]
+        A dictionary that maps the names of ``MutableData`` or ``ConstantData`` instances to their
+        corresponding values at inference time. If a model was created with ``MutableData``, these
+        are stored as ``SharedVariable`` with the name of the data variable and a value equal to
+        the initial data. At inference time, this information is stored in ``InferenceData``
+        objects under the ``constant_data`` group, which allows us to check whether a
+        ``SharedVariable`` instance changed its values after inference or not. If the values have
+        changed, then the ``SharedVariable`` is assumed to be volatile. If it has not changed, then
+        the ``SharedVariable`` is assumed to not be volatile. If a ``SharedVariable`` is not found
+        in either ``constant_data`` or ``constant_coords``, then it is assumed to be volatile.
+        Setting ``constant_data`` to ``None`` is equivalent to passing an empty dictionary.
+    constant_coords : Optional[Set[str]]
+        A set with the names of the mutable coordinates that have not changed their shape after
+        inference. If a model was created with mutable coordinates, these are stored as
+        ``SharedVariable`` with the name of the coordinate and a value equal to the length of said
+        coordinate. This set let's us check if a ``SharedVariable`` is a mutated coordinate, in
+        which case, it is considered volatile. If a ``SharedVariable`` is not found
+        in either ``constant_data`` or ``constant_coords``, then it is assumed to be volatile.
+        Setting ``constant_coords`` to ``None`` is equivalent to passing an empty set.
 
     Returns
     -------
@@ -1686,6 +1707,20 @@ def compile_forward_sampling_function(
 
     if basic_rvs is None:
         basic_rvs = []
+
+    if constant_data is None:
+        constant_data = {}
+    if constant_coords is None:
+        constant_coords = set()
+
+    # We define a helper function to check if shared values match to an array
+    def shared_value_matches(var):
+        try:
+            old_array_value = constant_data[var.name]
+        except KeyError:
+            return var.name in constant_coords
+        current_shared_value = var.get_value(borrow=True)
+        return np.array_equal(old_array_value, current_shared_value)
 
     # We need a function graph to walk the clients and propagate the volatile property
     fg = FunctionGraph(outputs=outputs, clone=False)
@@ -1702,6 +1737,7 @@ def compile_forward_sampling_function(
             or (  # SharedVariables, except RandomState/Generators
                 isinstance(node, SharedVariable)
                 and not isinstance(node, (RandomStateSharedVariable, RandomGeneratorSharedVariable))
+                and not shared_value_matches(node)
             )
             or (  # Basic RVs that are not in the trace
                 node in basic_rvs and node not in vars_in_trace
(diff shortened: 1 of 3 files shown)


pymc/sampling.py
```diff
@@ -1687,6 +1708,20 @@ def compile_forward_sampling_function(
...
+        constant_data = {}
+    if constant_coords is None:
+        constant_coords = set()
+
+    # We define a helper function to check if shared values match to an array
+    def shared_value_matches(var):
+        shared_value = var.get_value()
+        try:
+            array = constant_data[var.name]
+            return shared_value.shape == array.shape and all(shared_value == array)
```
ricardoV94: Does this do the same?
```suggestion
            return np.array_equal(shared_value, array)
```

lucianopaz: Looks like it. I had no idea it existed!

lucianopaz:
I'm not sure why precommit just dies after checking for links.

pymc/tests/test_sampling.py
```diff
@@ -1101,6 +1101,7 @@ def test_logging_sampled_basic_rvs_posterior_mutable(self, caplog):
         # hasn't changed since sampling
         idata = az_from_dict(posterior={"y1": np.zeros(5), "y2": np.zeros(5)})
         with m:
+            m.set_data("x1", 1)
```
ricardoV94: The condition should be changed to not expect resampling (we have one for resampling immediately below), and the comments above updated (they are outdated now as well). 

Or perhaps just remove this test in favor of the new one you added?

pymc/tests/test_sampling.py
```diff
@@ -1939,6 +2062,95 @@ def test_non_random_model_variable(self):
...
+            vars_in_trace=[a, b, mu, sigma],
+            basic_rvs=model.basic_RVs,
+            constant_data={"offsets": offsets.get_value()},
+            constant_coords={"name", "obs"},
+        )
+        assert volatile_rvs == {y}
+        assert {i.name for i in self.get_function_inputs(f)} == {"a", "b", "mu", "sigma"}
+        assert {i.name for i in self.get_function_roots(f)} == {"mu", "sigma", "name", "obs"}
+
+        # When we have both constant_data has different values than the shared variable, then
```
ricardoV94:
```suggestion
        # When we constant_data has different values than the shared variable, then
```

pymc/tests/test_sampling.py
```diff
@@ -1128,6 +1129,128 @@ def test_logging_sampled_basic_rvs_posterior_deterministic(self, caplog):
...
+            b = pm.Normal("b", mu=offsets, sigma=1)
+            mu = pm.Deterministic("mu", a + b[..., None], dims=["obs", "name"])
+            sigma = pm.HalfNormal("sigma", sigma=1, dims=["name"])
+
+            data = pm.MutableData(
+                "y_obs",
+                data,
+                dims=["obs", "name"],
+            )
+            pm.Normal("y", mu=mu, sigma=sigma, observed=data, dims=["obs", "name"])
```
ricardoV94: Can we put the model creation logic in a function to not repeat the code?

pymc/tests/test_sampling.py
```diff
@@ -1128,6 +1129,128 @@ def test_logging_sampled_basic_rvs_posterior_deterministic(self, caplog):
...
+            pm.sample_posterior_predictive(samples)
+        if kind == "MultiTrace":
+            assert caplog.record_tuples == [("pymc", logging.INFO, "Sampling: [a, b, sigma, y]")]
+            caplog.clear()
+        elif kind == "InferenceData":
+            assert caplog.record_tuples == [("pymc", logging.INFO, "Sampling: [y]")]
+            caplog.clear()
+        elif kind == "Dataset":
+            assert caplog.record_tuples == [("pymc", logging.INFO, "Sampling: [b, y]")]
+            caplog.clear()
```
ricardoV94: Can you add a comment explaining what is the "Dataset" condition doing?

pymc/tests/test_sampling.py
```diff
@@ -1128,6 +1129,128 @@ def test_logging_sampled_basic_rvs_posterior_deterministic(self, caplog):
...
+                save_warmup=False,
+                model=model,
+                log_likelihood=False,
+            )
+            if kind == "Dataset":
+                return kind, idata.posterior, model
+            else:
+                return kind, idata, model
+
+    def test_mutable_coords_and_data(self, mock_sample_results, caplog):
```
ricardoV94: It's a bit of a mouthful, but we are only checking the logging so it's more appropriate. If you remove the test mentioned above, I think you can just borrow its name instead: `test_logging_sampled_basic_rvs_posterior_mutable`

```suggestion
    def test_logging_sampled_basic_rvs_posterior_mutable_coords_and_data(self, mock_sample_results, caplog):
```

ricardoV94 (changes requested):

pymc/tests/test_sampling.py
lucianopaz: I'll remove this test in favor of the new one

pymc/tests/test_sampling.py
lucianopaz: I'll do this

pymc/sampling.py
```diff
@@ -1687,6 +1708,20 @@ def compile_forward_sampling_function(
...
+        constant_coords = set()
+
+    # We define a helper function to check if shared values match to an array
+    def shared_value_matches(var):
+        shared_value = var.get_value()
+        try:
+            array = constant_data[var.name]
+            return np.array_equal(shared_value, array)
+        except KeyError:
+            return var.name in constant_coords
```
ricardoV94: Reordering for performance, and also setting `borrow=True`, otherwise it deepcopies the internal value unnecessarily 

```suggestion
        try:
            old_value = constant_data[var.name]
        except KeyError:
            return var.name in constant_coords
        current_value = var.get_value(borrow=True)
        return np.array_equal(old_value, current_value)
```

lucianopaz: Much nicer!

ricardoV94: Btw your last push does not seem to have included any changes

lucianopaz: Oh stupid me!

ricardoV94 (approved):

ricardoV94:
Neat one @lucianopaz!

PR #6149, opened 2022-09-26
Add graphviz support for SymbolicRVs

### LINKED ISSUE #5766: ZeroInflatedPoisson and ZeroInflatedNegativeBinomial show up as Deterministic on graphviz
It's probably because they are not pure RandomVariables. We need to update the `model_graph` logic to accommodate them

larryshamalama:
Closes #5766.

Prior to this PR, symbolic distributions were labelled as "Deterministic" in the graph produced by Graphviz. The recent addition of symbolic RVs in #6072 allows `pymc/model_graph.py` to be adapted more easily. Here is an example of what is obtained using symbolic distributions such as `pm.Mixture`, `pm.Censored`, `pm.ZeroInflatedPoisson`.

```python
with pm.Model() as model:
    psi = pm.Beta("psi", 1., 1.)
    mu_zip = pm.Gamma("mu_zip", 1., 1.)
    zip_var = pm.ZeroInflatedPoisson("zip", psi=psi, mu=mu_zip)
    
    mu_cens = pm.Normal("mu_cens", 0., 5.)
    sigma_cens = pm.HalfCauchy("sigma_cens", 2.5)
    normal_dist = pm.Normal.dist(mu=mu_cens, sigma=sigma_cens)
    censored_normal = pm.Censored("censored_normal", normal_dist, lower=-2., upper=2., observed=[0, 0.5, 1])
    
    mu_grw = pm.Normal("mu_grw", 0., 5.)
    sigma_grw = pm.HalfCauchy("sigma_grw", 2.5)
    pm.GaussianRandomWalk("grw", init_dist=pm.Normal.dist(mu_grw, sigma_grw), steps=10)
    
pm.model_to_graphviz(zip_model)
```

image

Tests in [`pymc/tests/test_model_graph.py`](https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_model_graph.py) checks that the arrows are pointing to and from the right variables and that the overall shape of the Graphviz Digraph is correct. Perhaps an additional check that the distribution names are correct can be added, because this is what this PR is about.

### DIFF
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -21,15 +21,35 @@ from aesara.compile.sharedvalue import SharedVariable
 from aesara.graph import Apply
 from aesara.graph.basic import ancestors, walk
 from aesara.scalar.basic import Cast
+from aesara.tensor.basic import get_scalar_constant_value
 from aesara.tensor.elemwise import Elemwise
 from aesara.tensor.random.op import RandomVariable
 from aesara.tensor.var import TensorConstant, TensorVariable
 
 import pymc as pm
 
+from pymc.distributions import Discrete
+from pymc.distributions.discrete import DiracDelta
 from pymc.util import get_default_varnames, get_var_name
 
 VarName = NewType("VarName", str)
+
+
+def check_zip_graph_from_components(components):
+    """
+    This helper function checks if a mixture sub-graph corresponds to a
+    zero-inflated distribution using its components, a list of length two.
+    """
+    if not any(isinstance(var.owner.op, DiracDelta) for var in components):
+        return False
+
+    dirac_delta_idx = 1 - int(isinstance(components[0].owner.op, DiracDelta))
+    dirac_delta = components[dirac_delta_idx]
+    other_comp = components[1 - dirac_delta_idx]
+
+    return (get_scalar_constant_value(dirac_delta.owner.inputs[3]) == 0) and isinstance(
+        other_comp.owner.op, Discrete
+    )
 
 
 class ModelGraph:
@@ -154,16 +174,40 @@ class ModelGraph:
             shape = "box"
             style = "rounded, filled"
             label = f"{var_name}\n~\nMutableData"
-        elif v.owner and isinstance(v.owner.op, RandomVariable):
+        elif v.owner and (v in self.model.basic_RVs):
             shape = "ellipse"
-            if hasattr(v.tag, "observations"):
+            if v in self.model.observed_RVs:
                 # observed RV
                 style = "filled"
             else:
-                shape = "ellipse"
                 style = None
             symbol = v.owner.op.__class__.__name__
-            if symbol.endswith("RV"):
+            if symbol == "MarginalMixtureRV":
+                components = v.owner.inputs[2:]
+                if len(components) == 2:
+                    component_names = [
+                        var.owner.op.__class__.__name__.replace("Unmeasurable", "")[:-2]
+                        for var in components
+                    ]
+                    if check_zip_graph_from_components(components):
+                        # ZeroInflated distribution
+                        component_names.remove("DiracDelta")
+                        symbol = f"ZeroInflated{component_names[0]}"
+                    else:
+                        # X-Y mixture
+                        symbol = f"{'-'.join(component_names)}Mixture"
+                elif len(components) == 1:
+                    # single component dispatch mixture
+                    symbol = f"{components[0].owner.op.__class__.__name__.replace('Unmeasurable', '')[:-2]}Mixture"
+                else:
+                    symbol = symbol[:-2]  # just MarginalMixture
+            elif symbol == "CensoredRV":
+                censored_dist = v.owner.inputs[0]
+                symbol = symbol[:-2] + censored_dist.owner.op.__class__.__name__[:-2]
+            elif symbol == "TruncatedRV":
+                truncated_dist = v.owner.op.base_rv_op
+                symbol = symbol[:-2] + truncated_dist.__class__.__name__[:-2]
+            elif symbol.endswith("RV"):
                 symbol = symbol[:-2]
             label = f"{var_name}\n~\n{symbol}"
         else:
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -181,14 +181,11 @@ class ModelGraph:
                 style = "filled"
             else:
                 style = None
-            symbol = v.owner.op.__class__.__name__
-            if symbol == "MarginalMixtureRV":
+            symbol = v.owner.op._print_name[0]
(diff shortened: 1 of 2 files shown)


pymc/model_graph.py
```diff
@@ -27,11 +27,26 @@
...
 
+def check_zip_graph_from_components(components):
+    """
+    This helper function checks if a mixture sub-graph corresponds to a
+    zero-inflated distribution using its components
+    """
+    if len(components) != 2:
+        return False
+
+    return any(isinstance(var.owner.op, DiracDelta) for var in components) and all(
```
ricardoV94: To be precise you should also check that the value of the diracdelta is 0. You could have a OneInflatedDistribution as well.

You can use constant fold for that.

pymc/model_graph.py
```diff
@@ -154,7 +169,7 @@ def _make_node(self, var_name, graph, *, nx=False, cluster=False, formatting: st
             shape = "box"
             style = "rounded, filled"
             label = f"{var_name}\n~\nMutableData"
-        elif v.owner and isinstance(v.owner.op, RandomVariable):
+        elif v.owner and isinstance(v.owner.op, (RandomVariable, SymbolicRandomVariable)):
```
ricardoV94: We should just check if a variable is in `model.basic_RVs` instead of forcing a specific type. Similarly, instead of checking if `observations` is in a tag to figure out if it is observed, we should check if it is in `model.observed_RVs`

pymc/model_graph.py
```diff
@@ -163,7 +178,21 @@ def _make_node(self, var_name, graph, *, nx=False, cluster=False, formatting: st
...
+                    component_names = [var.owner.op.__class__.__name__[12:-2] for var in components]
+                    component_names.remove("DiracDelta")
+                    symbol = f"ZeroInflated{component_names[0]}"
+                elif len(components) == 1:
+                    symbol = f"{components[0].owner.op.__class__.__name__[12:-2]}Mixture"
+                else:
+                    symbol = symbol[:-2]  # just MarginalMixture
+            elif symbol == "CensoredRV":
+                censored_dist = v.owner.inputs[0]
+                symbol = symbol[:-2] + censored_dist.owner.op.__class__.__name__[:-2]
```
ricardoV94: You can apply the same logic to "TruncatedRVs"

pymc/model_graph.py
```diff
@@ -163,7 +178,21 @@ def _make_node(self, var_name, graph, *, nx=False, cluster=False, formatting: st
                 shape = "ellipse"
                 style = None
             symbol = v.owner.op.__class__.__name__
-            if symbol.endswith("RV"):
+            if symbol == "MarginalMixtureRV":
+                components = v.owner.inputs[2:]
+                if check_zip_graph_from_components(components):
+                    # [12:-2] because the component name are of the form "Unmeasurable...RV"
+                    component_names = [var.owner.op.__class__.__name__[12:-2] for var in components]
```
ricardoV94: Maybe this is more readable?
```suggestion
                    component_names = [var.owner.op.__class__.__name__.replace("Unmeasurable", "")[:-2], for var in components]
```

Also below

pymc/model_graph.py
```diff
@@ -163,7 +178,21 @@ def _make_node(self, var_name, graph, *, nx=False, cluster=False, formatting: st
...
+                    component_names.remove("DiracDelta")
+                    symbol = f"ZeroInflated{component_names[0]}"
+                elif len(components) == 1:
+                    symbol = f"{components[0].owner.op.__class__.__name__[12:-2]}Mixture"
+                else:
+                    symbol = symbol[:-2]  # just MarginalMixture
```
ricardoV94, line 200: If a mixture has two components that could still be nice to show as `X-YMixture`. What do you think?

pymc/model_graph.py
```diff
@@ -163,7 +178,21 @@ def _make_node(self, var_name, graph, *, nx=False, cluster=False, formatting: st
                 shape = "ellipse"
                 style = None
             symbol = v.owner.op.__class__.__name__
-            if symbol.endswith("RV"):
+            if symbol == "MarginalMixtureRV":
```
ricardoV94: This hard-coded logic makes it impossible to customize the plate name of distributions not implemented directly in PyMC (e.g., from pymc-experimental, or third-party packages). I see two more flexible alternatives:

1. Use a dispatch based system on the `Op`
2. The `Op` can eagerly define it's specialized name when it is created. So the `init` method of `Mixture` could do this logic and override the default `op.name`. We should use that instead of `__class__.__name__` anyway.

I think 2. is fine for this niche application.

larryshamalama: We could use a dispatch system, but I believe that this would entail refactoring some parts of this file, notably `_make_node`. Should we stick with 2. for now?

What do you mean by specialized name and `init` method? I'm not sure to catch exactly what you mean here

pymc/model_graph.py
larryshamalama: I think that this would be a good idea!

pymc/model_graph.py
larryshamalama: Yes!

pymc/model_graph.py
larryshamalama: Could we just check for `var.owner.inputs[2].owner.inputs[3].data`?

ricardoV94: You have to check if it's a constant first. Also sometimes Dimshuffles can crop up in the graphs (and this could change over versions)

pymc/model_graph.py
ricardoV94: It wouldn't require much changes to this file. Basically, we can use the `_print_name` attribute in the `Op` instead of `__class__.__name__[:-2]`.  Those `_print_name`s are static for vanilla `RandomVariable`s but can be customized at creation time by the `SymbolicRandomVariable`s when they are initialized.

The only issue is that the `_print_name` of some RandomVariables is not desirable. For instance `pm.Normal.rv_op._print_name` is `N`. But that just means we don't like that as a `_print_name`, so we can subclass it in PyMC and give it the long name.

This also avoids the whole `Unmeasurable` nonsense, as that is only part of the class name, not the `_print_name`

pymc/model_graph.py
ricardoV94: You can just call `at.get_scalar_constant` on that input, in a try/except. It's not much harder, but you don't have to reinvent that navigation/parsing logic

pymc/model_graph.py
larryshamalama: I can create an issue for creating PyMC RV instances with desired `_print_name`s

larryshamalama:
Closed in favor of a new PR that will use a dispatching approach

PR #6150, opened 2022-09-27
Add Constant fold helper

ricardoV94:
**What is this PR about?**

This PR refactors repeated constant fold logic into a helper function in aesaraf

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- Add `constant_fold` helper

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -81,6 +81,7 @@ __all__ = [
     "set_at_rng",
     "at_rng",
     "convert_observed_data",
+    "compile_pymc",
 ]
 
 
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -823,7 +823,7 @@ def find_rng_nodes(
 
 
 def replace_rng_nodes(outputs: Sequence[TensorVariable]) -> Sequence[TensorVariable]:
-    """Replace any RNG nodes upsteram of outputs by new RNGs of the same type
+    """Replace any RNG nodes upstream of outputs by new RNGs of the same type
 
     This can be used when combining a pre-existing graph with a cloned one, to ensure
     RNGs are unique across the two graphs.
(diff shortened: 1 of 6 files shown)


junpenglao (approved):
Nice.

pymc/aesaraf.py
```diff
@@ -970,3 +975,26 @@ def compile_pymc(
...
+    ------
+    NotConstantValueError:
+        If any of the variables cannot be successfully constant folded
+    """
+    fg = FunctionGraph(outputs=xs, features=[ShapeFeature()], clone=True)
+
+    folded_xs = rewrite_graph(fg, custom_rewrite=topo_constant_folding).outputs
+
+    if not all(isinstance(folded_x, Constant) for folded_x in folded_xs):
+        raise NotConstantValueError
```
junpenglao, line 998: What is the implication if we remove this? `folded_xs` would just be a tensor?
I am thinking that instead of raising error, maybe we can make the function being "best effort constant fold".

ricardoV94: It would be a tensor, but we cloned the graph, so it has nothing to do with the original one (except for shared inputs). If you tried to evaluate with the original inputs you would get `unused input` and `missing input` errors

ricardoV94:
Oops a test was failing... I thought it was just the pre-commit issue that was fixed in the meantime

pymc/aesaraf.py
junpenglao: LOL i was looking at this line and feeling we are being a bit more too aggressive - seems my hunch was justified.

ricardoV94: You mean the follow up PR? If so, yeah I thought I had already pushed it and you were looking at it. My bad :P

PR #6151, opened 2022-09-27
Fix DiscreteUniformRV dropping degenerate dimension

ricardoV94:
**What is this PR about?**

Bug found by @drbenvincent 
Scipy does not respect degenerate dimensions (size=1) implied by the parameters

**Checklist**
+ [x] Explain important implementation details 👆
+ [x] Make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).
+ [x] Link relevant issues (preferably in [nice commit messages](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html))
+ [x] Are the changes covered by tests and docstrings?
+ [x] Fill out the short summary sections 👇

## Major / Breaking Changes
- ...

## Bugfixes / New features
- Fix bug where `DiscreteUniformRV` would drop degenerate dimensions from draws

## Docs / Maintenance
- ...

PR #6154, opened 2022-09-27
Update mypy to version 0.981

lucianopaz:
**What is this PR about?**

Update mypy version to 0.981. According to [this](https://github.com/python/mypy/issues/13627), it should fix the failed CI encountered in both #6151 and #6147 

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/blocking.py
+++ b/pymc/blocking.py
@@ -20,7 +20,7 @@ Classes for working with subsets of parameters.
 import collections
 
 from functools import partial
-from typing import Callable, Dict, Optional, TypeVar
+from typing import Callable, Dict, Generic, Optional, TypeVar
 
 import numpy as np
 
@@ -35,7 +35,7 @@ PointType = Dict[str, np.ndarray]
 RaveledVars = collections.namedtuple("RaveledVars", "data, point_map_info")
 
 
-class Compose:
+class Compose(Generic[T]):
     """
     Compose two functions in a pickleable way
     """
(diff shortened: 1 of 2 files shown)


maresb:
Is the other mypy warning new as of 0.981?

Perhaps the return type should be `Any` instead of `T`?

maresb:
Ah, or should `DictToArrayBijection` be a generic on `T`?

EDIT: I got confused here, this should instead refer to `Compose`.

lucianopaz:
Looks like returning `Any` did the trick.

I ran pre-commit on all the files and that was the only new warning that popped up.

pymc/blocking.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, fa: Callable[[PointType], T], fb: Callable[[RaveledVars], Poi
         self.fa = fa
         self.fb = fb
 
-    def __call__(self, x: RaveledVars) -> T:
+    def __call__(self, x: RaveledVars) -> Any:
```
maresb:
```suggestion
    def __call__(self, x: RaveledVars) -> T:
```

maresb: I think you should change `class Compose:` above to `class Compose(Generic[T]):`. This way, the `T` variable from `__init__` and `__call__` will be linked, and the type variable will actually define a relation.

maresb (changes requested):
I'm thinking that `Compose` should be a generic

lucianopaz:
> I'm thinking that `Compose` should be a generic

Thanks @maresb! The `Generic[T]` class annotation fixed the type hints

maresb (approved):

ricardoV94 (approved):

PR #6159, opened 2022-09-28
Generalization of logp test comparison to facilitate relative and/or absolute comparison

ccaprani:
Following a discussion here https://github.com/pymc-devs/pymc-experimental/pull/84#discussion_r981956590 with @ricardoV94 , there may be cases when it is better to do a relative and/or absolute check on the comparison between the *desired* (SciPy) and *actual* (pymc) values. Currently only an absolute test is done, to `1.5*1e-decimal` [(see here)](https://github.com/numpy/numpy/blob/9c5a2b41cc0ade160a01b7423630f7ee1b8e63dc/numpy/testing/_private/utils.py#L1094) precision.

~~This (small) PR maintains current behaviour (i.e. absolute tolerance), but facilitates future expansion to consider both absolute and/or relative comparison.~~

~~Once/if this PR is accepted, then the `atol` and `rtol` arguments can be propagated out to test functions as a kwarg as `decimal` current is. It could help avoid some of the float32 skipped tests and deal with some other edge cases.~~

- Scope creep means this has been done now!

## Implementation
Change from the not-recommended [assert_almost_equal](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html) to the [assert_allclose](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) from `numpy.testing`. 

Maintain current behaviour by using `atol=decimal` and `rtol=0` in comparing the difference between *actual* and *desired* to `atol + rtol * abs(desired)`.

## Major / Breaking Changes
- None

## Bugfixes / New features
- Facilities a new feature in testing in that future logp could be tested relative and/or absolute.

## Docs / Maintenance
- N/A


ricardoV94:
Do you want to try some breaking change, and suggest values that look reasonable as a default? That would be great for maintenance / future devs

ccaprani:
Changing from using an atol to an rtol gives the same test results on `test_continuous.py` at least:
```python
rtol=10**(-decimal),
atol=0,#1.5 * 10 ** (-decimal),
```
I haven't got `polygamma` installed, so I get:
```
FAILED test_continuous.py::TestMatchesScipy::test_moyal_logcdf - AssertionError: 
FAILED test_continuous.py::TestMoments::test_polyagamma_moment[1.0-0.0-None-0.25] - RuntimeError: polyagamma package is not installed!
FAILED test_continuous.py::TestMoments::test_polyagamma_moment[1.0-z1-None-expected1] - RuntimeError: polyagamma package is not installed!
FAILED test_continuous.py::TestMoments::test_polyagamma_moment[h2-z2-None-expected2] - RuntimeError: polyagamma package is not installed!
FAILED test_continuous.py::TestMoments::test_polyagamma_moment[h3-z3-size3-expected3] - RuntimeError: polyagamma package is not installed!
========================================================================= 5 failed, 268 passed, 2 skipped, 18 warnings in 37.23s ==========================================================================
```

ricardoV94:
Would probably need to tweak the decimal value, no? And also important to check for float32.

ricardoV94:
I suggest you allow the rtol/atol to be configured already so you can unblock your PR on pymc-experimental.

Then we can follow up with a separate PR where we propose changing the default tolerance for the pre-existing tests.

pymc/tests/distributions/util.py
```diff
@@ -317,10 +317,11 @@ def _model_input_dict(model, param_vars, pt):
         pt_d = _model_input_dict(model, param_vars, pt)
         pt_logp = pm.Point(pt_d, model=model)
         pt_ref = pm.Point(pt, filter_model_vars=False, model=model)
-        npt.assert_almost_equal(
```
ricardoV94, line 320: The same changes should be done to `check_logcdf` and `check_selfconsistency_discrete_logcdf`

ccaprani:
@ricardoV94 so an initial refactor is done to now allow both atol and rtol to be specified. 

Given that previous behaviour was an absolute tolerance set at `decimals=6` (just considering float64 for now), then this corresponds to 1e-6 as an absolute precision, so I adopted this as default. However, I moved all `check_logp` and `check_logc` from an absolute to an `rtol` check. I also left a default atol of 1e-10, which is necessary in some cases to get current test behaviour.

I confirmed test behaviour is the same, at least for bound, censored, continuous. Running the suite (and I don't have polyagamma isntalled, so that's 5 fails) gives:
```
50 failed, 2010 passed, 2 skipped, 243 xfailed, 164 warnings in 870.40s (0:14:30)
```

Here are a few observations:
- I moved default tolerances to `select_by_precision` so they can be set in one place only.
- Currently the absolute tolerances for float64 and float32 are set everywhere: suggest this is migrated to avoid duplicating defaults.- 
- I did not adjust tests that did not use `check_logp` or `check_logc` - many functions seem to 'roll their own'
- The float32 checks in many cases were set at *extremely* generous levels (e.g. 1e1!)
- `test_continuous.py::TestMatchesScipy::test_moyal_logcdf` fails, and it's nothing to do with this change.

I think this architecture now allows for a tuning/migration to more flexible and even more stringent test tolerances which can be worked on over time.

ccaprani:
Hi @ricardoV94 just checking in on this as I see with new merges conflicts are arising.

ricardoV94:
@ccaprani let me see if I can get someone else to weight in :) I'll try to revisit again soon as well!

Armavica:
Wouldn't it be more principled in theory to check `logp` and `logcdf` for absolute error, rather than relative?
Unless I am mistaken, MCMC is invariant by addition of any constant to `logp`. Which means that there shouldn't be any particular meaning given to the value 0 of these functions, and an error of `logp` from `0` to `1` is in theory as bad as an error from `1e10` to `1e10 + 1`. For this reason, I think that checking for absolute error is mathematically more appropriate. However, I can also see the technical problem when `logp` has extremely high values such as `1e300`. In this case, a small `rtol` can probably help.

This PR proposes a relatively large `rtol` and a tiny `atol`. But for these reasons I think that I would rather use a fair `atol`, which will determine the testing interval for most of the cases, and a tiny `rtol`, which will still dominate for very large values and overcome float precision issues.

ccaprani:
@Armavica Yes; for some of the models & tests done here it makes sense to have an `rtol` as well as the `atol`. Indeed, if you look through, many tests currently avoid using `check_logp` as I suspect there was insufficient flexibility in setting tolerances for passing tests. This PR would give this flexibility and then a discussion about the most suitable values of `atol` and `rtol` at least has a purpose.

ccaprani:
@michaelosthege thank you. Not sure why [bc111ca](https://github.com/pymc-devs/pymc/pull/6159/commits/bc111cad7380a25379308e1e68316a1f82d1cd56) came along for the ride but [7371d2a](https://github.com/pymc-devs/pymc/pull/6159/commits/7371d2ab0bc37767f1ffdc08a5b03e3c4da03ea8) passes all in `test_discrete` on my Ubuntu 2022 x64 machine with the 'fix'.

However, I'm not sure that we should be adjusting tolerances so casually to get tests to pass. On one hand adjusting until they do makes sense as its backward compatible. On the other, we could introduce some dubious setting. What do you think?

ricardoV94:
These tests are deterministic (in the long run). The only randomness is what subset of parameters are tested each time (the values don't change). You should set `n_samples=-1` to test all combinations. Feel free to override the default once here to confirm all tests pass and then revert that:

https://github.com/pymc-devs/pymc/blob/8b8e49bc22c261b6b97156630ffea5128e8a7a61/pymc/tests/distributions/util.py#L236

As well as in the `logcdf` and `locdf_match_...`

ccaprani:
Hi @ricardoV94 . I see a reference to it in #6850 in v5.8.0. Should we revitalize this? Has its time come?!

ricardoV94:
> Hi @ricardoV94 . I see a reference to it in #6850 in v5.8.0. Should we revitalize this? Has its time come?!

Let's get it done!

ccaprani:
Darn - this was on me to complete. @fonnesbeck do you see value in it now?

fonnesbeck:
If you are motivated to merge it, sure! You will want to make sure it plays nicely with v6. You can either re-open this or create a new PR. The new major version was just a good place for us to start closing out unmerged PRs of a certain vintage.

PR #6173, opened 2022-10-02
Rename SMC files

### LINKED ISSUE #6171: Rename SMC files
I think we should call sample_smc.py, sampling.py as it's already inside the smc module anyway. smc.py would be better named kernels.py

CC @aloctavodia

IMvision12:
Renamed SMC files

Closes : #6171


ricardoV94:
We also need to rename the associated test files and update the tests workflow

PR #6189, opened 2022-10-07
Don't mask inputs to `pm.Data` containers

michaelosthege:
**What is this PR about?**
I ran into a problem where NaN values in my data got masked away when creating `pm.Data` containers.
This was quite unexpected.

We need the masking for `observed`, but that's being taken care of in other places already.

**Checklist**

## Major / Breaking Changes
- NaN values passed to `pm.Data` are no longer masked automatically.

## Bugfixes / New features
- None

## Docs / Maintenance
- None

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -33,8 +33,6 @@ from aesara.tensor.var import TensorConstant, TensorVariable
 
 import pymc as pm
 
-from pymc.aesaraf import convert_observed_data
-
 __all__ = [
     "get_data",
     "GeneratorAdapter",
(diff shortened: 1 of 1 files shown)


michaelosthege:
@ricardoV94 and I discussed this offline and didn't find a good solution that keeps `pm.Data` inputs unmasked while not breaking imputation or `model_to_graphviz` at the same time.

Without a good solution, and keeping in mind that @ricardoV94 wants to refactor how we manage `observed` variables I think we can close this PR.

PR #6192, opened 2022-10-08
Collect sampler warnings only through stats

michaelosthege:
**What is this PR about?**
Deleting code that gave special treatments to sampler warnings.

This PR mostly changes three things:
+ [x] The `SamplerWarning` class and related code, including the code for `run_convergence_checks` is moved from `report.py` to `convergence.py`. Here it is more functionally structure, and no longer coupled to the state of a `SamplerReport`.
+ [x] A new stat called `"warning"` is introduced. This is an `object` stat to which samplers can emit warnings instead of piggy-backing them onto the draws.
+ [x] Handling warnings is removed from the `BaseTrace`, in favor of managing them via the "warning" stat.
+ <s>updating the minimum ArviZ requirement to the latest version that can handle `object`-typed variables (by skipping and warning about them) when saving (related #6194).</s>
+ [x] A `pm.sample(keep_warning_stat={False, True})` setting was added to enable access to sampler warnings without breaking `.to_netcdf()` by default.

The logic in `sampling.py` was modified to stop handling warnings separately.
Instead, the `record_and_warn` function now takes over the task of logging warnings that are coming through via the "warning" stat.

I also modified the corresponding tests to be more targeted.

**Checklist**

## Major / Breaking Changes
- Detailed information about divergent samples is no longer captured.
- Support for handling sampler warnings via `._warnings` attributes on samplers and `BaseTrace` was removed in favor of a "warning" stat.

## Bugfixes / New features
- HMC/NUTS samplers now emit a "warning" stat that ends up in the `InferenceData`.
- With `pm.sample(keep_warning_stat=True)` one can now access detailed info about divergence warnings directly via `idata.sample_stats.warning`, at the cost of not being able to save these `InferenceData` objects.

## Docs / Maintenance
- New `pm.stats.convergence` submodule for functional implementation of convergence diagnostics and warnings.

### DIFF
--- a/pymc/step_methods/hmc/base_hmc.py
+++ b/pymc/step_methods/hmc/base_hmc.py
@@ -17,6 +17,7 @@ import time
 
 from abc import abstractmethod
 from collections import namedtuple
+from typing import Optional
 
 import numpy as np
 
@@ -134,8 +135,6 @@ class BaseHMC(GradientSharedStep):
         self.integrator = integration.CpuLeapfrogIntegrator(self.potential, self._logp_dlogp_func)
 
         self._step_rand = step_rand
-        self._warnings = []
-        self._samples_after_tune = 0
         self._num_divs_sample = 0
 
     @abstractmethod
@@ -173,8 +172,7 @@ class BaseHMC(GradientSharedStep):
                 "critical",
                 self.iter_count,
             )
-            self._warnings.append(warning)
-            raise SamplingError("Bad initial energy")
+            raise SamplingError(f"Bad initial energy: {warning}")
 
         adapt_step = self.tune and self.adapt_step_size
         step_size = self.step_adapt.current(adapt_step)
@@ -190,6 +188,7 @@ class BaseHMC(GradientSharedStep):
 
         self.step_adapt.update(hmc_step.accept_stat, adapt_step)
         self.potential.update(hmc_step.end.q, hmc_step.end.q_grad, self.tune)
+        warning: Optional[SamplerWarning] = None
         if hmc_step.divergence_info:
             info = hmc_step.divergence_info
             point = None
@@ -205,7 +204,7 @@ class BaseHMC(GradientSharedStep):
                     point = DictToArrayBijection.rmap(info.state.q)
 
                 if self._num_divs_sample < 100 and info.state_div is not None:
-                    point = DictToArrayBijection.rmap(info.state_div.q)
+                    point_dest = DictToArrayBijection.rmap(info.state_div.q)
 
                 if self._num_divs_sample < 100:
                     info_store = info
@@ -220,11 +219,7 @@ class BaseHMC(GradientSharedStep):
                 divergence_info=info_store,
             )
 
-            self._warnings.append(warning)
-
         self.iter_count += 1
-        if not self.tune:
-            self._samples_after_tune += 1
 
         stats = {
             "tune": self.tune,
@@ -232,6 +227,7 @@ class BaseHMC(GradientSharedStep):
             "perf_counter_diff": perf_end - perf_start,
             "process_time_diff": process_end - process_start,
             "perf_counter_start": perf_start,
+            "warning": warning,
         }
 
         stats.update(hmc_step.stats)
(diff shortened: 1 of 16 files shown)


pymc/step_methods/hmc/base_hmc.py
```diff
@@ -205,10 +203,11 @@ def astep(self, q0):
                     point = DictToArrayBijection.rmap(info.state.q)
 
                 if self._num_divs_sample < 100 and info.state_div is not None:
-                    point = DictToArrayBijection.rmap(info.state_div.q)
+                    point_dest = DictToArrayBijection.rmap(info.state_div.q)
```
michaelosthege, line 207: @aseyboldt was this a typo? The `point_dest` was always `None`..

aseyboldt: I don't think it is always None. It should be non-None if the destination position contains nans or inf, unless I'm missing something...

aseyboldt:
Hm, I really don't like loosing the extra divergence info. I'm not sure how many people actually use this, but I certainly do :-)
If we want to get rid of the warning system, I think that's fine, but we should then store the information in in the sampler stats instead (or at least have an option to do that).

pymc/step_methods/hmc/base_hmc.py
michaelosthege: In line 196/194 you initialized `point_dest = None` but in the _before_ version of this code it was never assigned to anything.

aseyboldt: Ah I see! That explains some weirdness I've seen recently :-)

Seems that is was broken during the theano->aesara rename:
https://github.com/pymc-devs/pymc/commit/4f5778ea7f1075065de2e6813ff4aefd4d788d67#diff-f95827a70f0ac1a6ce0c09711ddaf46fe6694ef45fda7f88079994865c3fb4beL196

michaelosthege:
> Hm, I really don't like loosing the extra divergence info. I'm not sure how many people actually use this, but I certainly do :-) If we want to get rid of the warning system, I think that's fine, but we should then store the information in in the sampler stats instead (or at least have an option to do that).

Unfortunately it's not that easy, because a stat can't be a dictionary.
In addition it is not clear whether stats can be sparse.

Assuming we want to store them for later it would be good to use the same data structure across HMC implementations (PyMC, nutpie, numpyro). Do you have any idea which structure that could be?

michaelosthege:
I think we can compromise:
Storing `SamplerWarning`-typed items in the `stats` dict and `idata.sample_stats` group actually works.
...until one tries to save it `to_netcdf`.

Coordinating with Oriol, I'll open a PR in ArviZ to filter out `object`-typed variables from the groups before saving. This will make `to_netcdf` also more robust in case there are other variables with invalid dtype.
(We can try converting them, but if this fails log a warning..)

aseyboldt:
I think we should be able to come up with a nice format for those in the sampler stats, without resorting to storing objects.
The most tricky part is dealing with the sparse structure, but we could handle that with a hierarchical index, or alternatively a second variable that maps the divergence to the chain. So something along the lines of this:

```
# Extra dims in the sample_stats:
"divergence" # One key for each divergence across all chains
"unconstrained_parameter"
# Extra vars:
"divergence_start": ("divergence", "unconstrained_parameter")
"divergence_end": ("divergence", "unconstrained_parameter")
"divergence_chain": ("divergence")  # For each divergence the label of the chain where that divergence happend
```

Instead of "divergence_chain" we could also use a hierarchical index for the divergence index. I think there is still some trouble when storing hierarchical indices as netcdf, but maybe after an unstack that would be fine?

michaelosthege:
@aseyboldt a serializable structure would be nice, yes, but I see other issues with higher priority..
With https://github.com/arviz-devs/arviz/pull/2134 ArviZ will automatically drop `object` variables before saving which means that we can safely have the `SamplerWarning` end up `InferenceData`. It just won't be saved, but that wasn't the case before either.

To clean up the sampler stats structure, IMO the next step is #6207 and introducing a coord/naming scheme by which samplers can be identified.

This should also open the door to storing other sampler-wise information such as which variables they were sampling.

In the meantime, this PR will have to wait for the upcoming ArviZ release.

michaelosthege:
The last changes I pushed made thus PR independent of ArviZ updates.

I updated the description, so this is ready to review/merge :)

michaelosthege:
@pymc-devs/core-contributors can I get reviews here?
I'd like to move forward before git conflicts creep in..

OriolAbril:
> I think we should be able to come up with a nice format for those in the sampler stats, without resorting to storing objects.
The most tricky part is dealing with the sparse structure, but we could handle that with a hierarchical index, or alternatively a second variable that maps the divergence to the chain. So something along the lines of this:
>
>     # Extra dims in the sample_stats:
>     "divergence" # One key for each divergence across all chains
>     "unconstrained_parameter"
>     # Extra vars:
>     "divergence_start": ("divergence", "unconstrained_parameter")
>     "divergence_end": ("divergence", "unconstrained_parameter")
>     "divergence_chain": ("divergence")  # For each divergence the label of the chain where that divergence happend

I think it is worth it to move that to an issue. I tried going over the data in the divergences warning but was unable to understand what info was exactly there and how to store it in "regular" arrays. Defining a structure that works in all cases might require a bit of extra work to account for multiple samplers or adding some exceptions for this to work if nuts is the only sampler, but it might be somewhat feasible. At least on xarray side, the dimensions are shared, but variables in a dataset can have any dimensions (there is not need for all variables to have chain and draw dims for example) and a dataset can have as many dimensions and coordinates as desired.

lucianopaz (approved):
Looks good to me, @michaelosthege . I agree with @OriolAbril  that we need to open an issue to change the warning data structure to something that is serializable. It’s a real waste to lose all the warning information when storing the results to netcdf.

michaelosthege:
Thanks Luciano!

As you can see, I created #6252.

I'll merge this before another git conflict creeps in..

PR #6203, opened 2022-10-11
Fix bug in `get_vars_in_point_list` when model does not have variables that exist in the trace

### LINKED ISSUE #6199: `get_vars_in_point_list` should ignore variables that are not in the model

[This thread](https://discourse.pymc.io/t/sample-posterior-predictive-works-fine-in-pymc-3-raises-exception-in-v4/) pointed out a bug in `get_vars_in_point_list`.

```python
with pm.Model() as model:
    a = pm.Normal(“a”,0,1)
    b = pm.Normal(“b”,0,1)

vars_in_trace = pm.sampling.get_vars_in_point_list(
    [{“a”:0, “c”:0}],
    model,
)
```

This should return a list with `a`, but raises a `KeyError` instead because `”c”` is not in `model`.

The solution is simple, just include the tensors that are in the model and in the trace, ignoring the variables in the trace that yield `KeyError`

lucianopaz:
Closes #6199 

**What is this PR about?**
...

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- `get_vars_in_point_list` was implemented in a way that expected all of the variables that were contained in the trace, to also be defined in the model. This fixes that problem and makes the function return a list of the variables that are defined both in the trace and in the model.

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1601,6 +1601,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
             except KeyError:
                 raise e
 
+    def __contains__(self, key):
+        return key in self.named_vars or self.name_for(key) in self.named_vars
+
     def compile_fn(
         self,
         outs: Sequence[Variable],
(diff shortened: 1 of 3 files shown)


ricardoV94 (approved):
Thanks @lucianopaz LGTM

Issue #5160, opened 2021-11-09
v4 <-> InferenceData integration


OriolAbril:
For now a dump of things to take into account regarding integration of PyMC v4 with InferenceData, which can now be improved signifiicantly further now that the converter is part of the PyMC codebase. 

Goals:
* Converting from MultiTrace to InferenceData does not represent loosing of information -> transformed variables, metadata, values in report...
  * feasible
* Storing sampler arguments, initial values, mass matrix... to allow reproducible sampling given a model object and an inferencedata. A bit related to https://github.com/arviz-devs/arviz/issues/220
  * might not be feasible yet

Related issues:
* https://github.com/arviz-devs/arviz/issues/230
* https://github.com/arviz-devs/arviz/issues/420
* Making the converter Dask compatible 
* https://github.com/arviz-devs/arviz/issues/1224
* https://github.com/arviz-devs/arviz/issues/1257
* https://github.com/arviz-devs/arviz/issues/1509
* https://github.com/arviz-devs/arviz/issues/1748

bacalfa:
Please add a way to return trace information from `pm.sample` (https://discourse.pymc.io/t/inferencedata-is-missing-details-about-divergent-transitions/7740).

OriolAbril:
It might make sense to subclass inferencedata at some point to add extra conventions when reading/writing inferencedata files to netcdf or zarr. For example, it might be possible to retrieve the transformations on posterior variables from the names or attributes of posterior and unconstrained_posterior groups (even if not always). I have also not been following the modelbuilder work, but it might also help a bit there?

michaelosthege:
This is obsolete with the upcoming ArviZ 1.0 release terminating `InferenceData` in favor of `xarray.DataTree`: https://python.arviz.org/en/latest/user_guide/migration_guide.html

OriolAbril:
I agree it needs an update and clean up, but I think only my last comment about subclassing the InferenceData class is affected by the move from InferenceData custom class to xarray provided DataTree class. Both classes are a Python representation of hierarchical array files like netcdf or zarr.

That being said, with the zarr backend and many other changes since v4 and the fact this was mostly a list of other issues it might not be worth to keep this issue open

PR #6208, opened 2022-10-11
speed up posterior predictive sampling

OriolAbril:
The goal of this PR is to accelerate the `dataset_to_point_list` function which right now is
often the bottleneck of posterior predictive sampling. Moreover, I would also like to add some
extra flexibility on the dimensions that are considered sample dimensions.

related to #5160 

**Checklist**

## Bugfixes / New features
- Added `sample_dims` argument to `sample_posterior_predictive`.

## Docs / Maintenance
- Improved the performance of `sample_posterior_predictive` when using InferenceData or Dataset as input.

### DIFF
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -431,9 +431,7 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
     def posterior_predictive_to_xarray(self):
         """Convert posterior_predictive samples to xarray."""
         data = self.posterior_predictive
-        dims = {
-            var_name: self.sample_dims + self.dims.get(var_name, []) for var_name in data.keys()
-        }
+        dims = {var_name: self.sample_dims + self.dims.get(var_name, []) for var_name in data}
         return dict_to_dataset(
             data, library=pymc, coords=self.coords, dims=dims, default_dims=self.sample_dims
         )
@@ -442,9 +440,7 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
     def predictions_to_xarray(self):
         """Convert predictions (out of sample predictions) to xarray."""
         data = self.predictions
-        dims = {
-            var_name: self.sample_dims + self.dims.get(var_name, []) for var_name in data.keys()
-        }
+        dims = {var_name: self.sample_dims + self.dims.get(var_name, []) for var_name in data}
         return dict_to_dataset(
             data, library=pymc, coords=self.coords, dims=dims, default_dims=self.sample_dims
         )
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -7,6 +7,7 @@ from typing import (  # pylint: disable=unused-import
     Any,
     Dict,
     Iterable,
+    List,
     Mapping,
     Optional,
     Sequence,
@@ -15,7 +16,6 @@ from typing import (  # pylint: disable=unused-import
 )
 
 import numpy as np
-import xarray as xr
 
 from aesara.graph.basic import Constant
 from aesara.tensor.sharedvar import SharedVariable
@@ -162,6 +162,7 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
         predictions=None,
         coords: Optional[CoordSpec] = None,
         dims: Optional[DimSpec] = None,
+        sample_dims: Optional[List] = None,
         model=None,
         save_warmup: Optional[bool] = None,
     ):
@@ -223,6 +224,9 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
                 for var_name, dims in self.model.RV_dims.items()
             }
             self.dims = {**model_dims, **self.dims}
+        if sample_dims is None:
+            sample_dims = ["chain", "draw"]
+        self.sample_dims = sample_dims
 
         self.observations = find_observations(self.model)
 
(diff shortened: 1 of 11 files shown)


OriolAbril:
Got the proof of concept working with xarray-einstats and einops. Will write a simple xarray reshaper function to avoid the extra dependency. The reshape we need here is the simplest case supported by those.

OriolAbril:
Needs https://github.com/arviz-devs/arviz/pull/2138 to get all tests to pass.

pymc/backends/arviz.py
```diff
@@ -419,36 +423,31 @@ def log_likelihood_to_xarray(self):
...
+            data, library=pymc, coords=self.coords, dims=dims, default_dims=self.sample_dims
         )
 
     @requires(["predictions"])
     def predictions_to_xarray(self):
         """Convert predictions (out of sample predictions) to xarray."""
-        return self.translate_posterior_predictive_dict_to_xarray(self.predictions, "predictions")
+        data = self.predictions
+        dims = {
+            var_name: self.sample_dims + self.dims.get(var_name, []) for var_name in data.keys()
```
lucianopaz:
```suggestion
            var_name: self.sample_dims + self.dims.get(var_name, []) for var_name in data
```

pymc/backends/arviz.py
```diff
@@ -419,36 +423,31 @@ def log_likelihood_to_xarray(self):
...
+        )
 
     @requires(["posterior_predictive"])
     def posterior_predictive_to_xarray(self):
         """Convert posterior_predictive samples to xarray."""
-        return self.translate_posterior_predictive_dict_to_xarray(
-            self.posterior_predictive, "posterior_predictive"
+        data = self.posterior_predictive
+        dims = {
+            var_name: self.sample_dims + self.dims.get(var_name, []) for var_name in data.keys()
```
lucianopaz:
```suggestion
            var_name: self.sample_dims + self.dims.get(var_name, []) for var_name in data
```

pymc/sampling.py
```diff
@@ -1967,23 +1977,30 @@ def sample_posterior_predictive(
...
+        else:
+            ppc_trace[k] = ary.reshape((nchain, len_trace, *ary.shape[1:]))
 
     if not return_inferencedata:
         return ppc_trace
     ikwargs: Dict[str, Any] = dict(model=model, **idata_kwargs)
+    ikwargs.setdefault("sample_dims", sample_dims)
+    if stacked_dims is not None:
+        coords = ikwargs.get("coords", {})
+        ikwargs["coords"] = {**stacked_dims, **coords}
```
lucianopaz, line 1993: If `stacked_dims` is None, unpacking will error out. You should convert it to an empty dict before unpacking, if it were None

pymc/util.py
```diff
@@ -231,38 +230,26 @@ def enhanced(*args, **kwargs):
...
             raise ValueError(f"Variable names must be str, but dataset key {vn} is a {type(vn)}.")
-    # make dicts
-    points: List[Dict[Hashable, np.ndarray]] = []
-    da: "xarray.DataArray"
-    for c in ds.chain:
-        for d in ds.draw:
-            points.append({vn: da.sel(chain=c, draw=d).values for vn, da in ds.items()})
+    num_sample_dims = len(sample_dims)
+    stacked_dims = {dim_name: ds[dim_name] for dim_name in sample_dims}
+    ds = ds.transpose(*sample_dims, ...)
```
lucianopaz, line 243: I’ma bit confused here. Where does the speed up come from? Does it come from not appending to a list?
If yes, then transposing and creating the intermediate dictionaries seems unnecessary to me.

Could you do something like
```
points = [
    {vn: da.values for vn, da in ds.sel(**coords).items()}
    for coords in [
        dict(stacked_dims, coord) for coord in
        itertools.product(stacked_dims.values())
    ]
]
```

I still feel that this could be speed up more if we could avoid the need to create dictionaries and use datasets as mappings. Anyway, if I misunderstood where the speed up is coming from, feel free to ignore this

lucianopaz (dismissed):
Thanks @OriolAbril . Could you tell me where the speed up is coming from?

pymc/sampling.py
ricardoV94: This is in a branch that requires it to not be `None`

pymc/util.py
ricardoV94: IIRC the speedup comes from doing numerical indexing in the inner loop as opposed to dim indexing.

pymc/util.py
```diff
@@ -231,38 +230,26 @@ def enhanced(*args, **kwargs):
...
+    stacked_dict = {
+        vn: da.values.reshape((-1, *da.shape[num_sample_dims:])) for vn, da in ds.items()
+    }
+    points = [
```
ricardoV94, line 247: Perhaps we could yield instead of returning the whole list at once?

pymc/sampling.py
```diff
@@ -1835,36 +1847,34 @@ def sample_posterior_predictive(
         idata_kwargs = {}
     else:
         idata_kwargs = idata_kwargs.copy()
+    if sample_dims is None:
+        sample_dims = ["chain", "draw"]
```
ricardoV94, line 1851: If you use a tuple you can use it directly as the default value in the signature

pymc/util.py
lucianopaz: I agree with using a lazy generator approach unless the whole list is needed at once for some reason

pymc/util.py
lucianopaz: If that’s the case, we could use `isel` instead of `sel` in my suggested snippet to also use numerical indexing

pymc/sampling.py
lucianopaz: Oh, I hadn’t noticed

pymc/util.py
ricardoV94: Maybe, if that's as fast as numpy indexing. Internet tells me xarray operations of this kind have a lot of overhead, but perhaps isel doesn't...

lucianopaz: It looks like xarray's overhead is huge regardless

image

`isel` is ~500x slower than direct indexing of the numpy array
`sel` is ~1000x slower than direct indexing of the numpy array

So, with 4000 chains and draws, xarray would take seconds to finish with this small test `DataArray`, whereas indexing directly with numpy would cut that down to milliseconds.

lucianopaz: I was curious and actually tested what I said above

image

In the end, xarray was faster than I had anticipated, but it's still more than 100x slower than numpy. I'm just amazed at how slow iterating with xarray is actually! I thought that they would have at least tried to stay close to numpy.

ricardoV94: > sel is ~1000x slower than direct indexing of the numpy array

Yeah, Oriol got the same results (1000x vs `.sel`) when he timed it before opening the PR

OriolAbril: Yeah, I think xarray here is more like pandas than numpy, it is explicitly not designed for this kind of thing but to work with objects as a whole or take "array subsets". Both `i/sel(chain=0)` probably check `chain` exists, then find which position it is in the underlying array, generate the whole `0, :` or `:, 0, :, :` to be used with numpy and then call numpy.

As for reshaping, I am under the impression that subsetting over a single axis is marginally faster than over 2 or 3 axis so given how we potentially repeat that same subset a few thousand times tipically it might help a bit once we are in ms range.

pymc/util.py
OriolAbril: If that works later in the code then yes! I only kept the list because the function is called `_to_list`. You should assume I have no idea about the format we need to interface with the aesara random drawing function.

Here would that be using a `()` comprehension or an explicit loop with a yield? Or either?

lucianopaz: I like the `()` comprehensión more

ricardoV94: Me too

ricardoV94:
@OriolAbril can we split the speedup fix of this PR from the flexible `sample_dims`? IIRC we don't depend on Arviz for the first and could get it merged sooner?

OriolAbril:
I plan to release ArviZ on Saturday. In my case the limiting factor is not the ArviZ release but my own time availability. If I have to split the PR it will take longer. If you need that before feel free to split the PR. It doesn't matter if the functionality gets merged in 1 or 2 PRs

ricardoV94:
> I plan to release ArviZ on Saturday. In my case the limiting factor is not the ArviZ release but my own time availability. If I have to split the PR it will take longer. If you need that before feel free to split the PR. It doesn't matter if the functionality gets merged in 1 or 2 PRs

Saturday should be fine. Let me know if there's anything I can help with otherwise.

pymc/util.py
ricardoV94: I noticed the code downstream of this may be incompatible with generators (it asks for len and sometimes to check the first point...)

OriolAbril:
Addressed some of the comments but not all of them. 

I switched the list to generator in `dataset_to_point_list` and got a few errors plus a mypy hell. I can probably update the code to use the generator but I can't do it today, it will need to be toward the end of the week or beginning of the next. I think it can also be a follow-up PR or if someone else has time it can be pushed here directly too.

I also used a tuple for the sample dims, but arviz expects a list so it is either using a list from the start or using a tuple and converting it to a list later on. I don't really care either way but for now left the list from the start.

I ran tests and mypy locally so I expect all tests to pass and the PR to be ready to merge after that.

ricardoV94 (approved):

ricardoV94:
@OriolAbril Can you add a bullet point in the top post under the `Docs / Maintenance` section?

Issue #5339, opened 2022-01-10
Strange Aesara warning and broken pipe error when fitting model with missing values


fonnesbeck:
I'm fitting a manually-constructed GRW model in v4beta that includes a likelihood with a lot of missing values. As the model is fit using `sample`, I get several of these errors.

```
warnings.warn(
/Users/cfonnesbeck/GitHub/pie/.env/lib/python3.8/site-packages/aeppl/joint_logprob.py:161: UserWarning: Found a random variable that was neither among the observations nor the conditioned variables: halfnormal_rv{0, (0, 0), floatX, False}(RandomStateSharedVariable(<RandomState(MT19937) at 0x16E5E6640>), TensorConstant{[]}, TensorConstant{11}, BroadcastTo.0, BroadcastTo.0)
```

The model is below; the node it is referring to is the likelihood standard deviation. If missing value is filled, the model runs normally.

```python
coords = {
    'pitcher': pitcher.values,
    'age': age.values.astype(int)
}
with pm.Model(coords=coords) as age_model:

    z_mu = pm.Normal('z_mu', mu=0, sigma=1, dims='pitcher')
    s_mu = pm.HalfCauchy('s_mu', 3)
    m_mu = pm.Normal('m_mu', mu=92, sigma=5)
    mu = pm.Deterministic('mu', m_mu + s_mu * z_mu)

    # GRW 
    rho = pm.Normal('rho', mu=0, sigma=np.append(0.001, np.ones(len(age)-1)))
    age_curve = pm.Deterministic('age_curve', rho.cumsum(), dims='age')

    theta = mu.dimshuffle(0, 'x') + age_curve

    sigma = pm.HalfNormal('sigma', 10)
    velo = pm.Normal('velo', theta, sigma=sigma, observed=y)
```

When run, the model quickly dies with a broken pipe, apparently related to multiprocessing:

```
/Users/cfonnesbeck/GitHub/pie/.env/src/pymc/pymc/model.py:1301: ImputationWarning: Data in velo contains missing values and will be automatically imputed from the sampling distribution.
  warnings.warn(impute_message, ImputationWarning)
Auto-assigning NUTS sampler...
Initializing NUTS using adapt_full...
/Users/cfonnesbeck/GitHub/pie/.env/src/pymc/pymc/step_methods/hmc/quadpotential.py:611: UserWarning: QuadPotentialFullAdapt is an experimental feature
  warnings.warn("QuadPotentialFullAdapt is an experimental feature")
[...]

PR #3919, opened 2020-05-11
Set start method to forkserver for MacOS

AlexAndorra:
As discussed in #3912 with @lucianopaz and @twiecki, this PR sets the start method to `forkserver` when the platform is MacOS. This is intended to solve #3849.

As this is implemented in `setup.py`, the new start method is modified each time users import PyMC3, but it does it smoothly by using `multiprocessing.get_context` instead of `multiprocessing.set_start_method`. From the [python docs](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods):

> A library which wants to use a particular start method should probably use `get_context()` to avoid interfering with the choice of the library user.

I did some raw tests locally and this does work!

+ [x] No breaking changes (that I know of)
+ [x] Added to RELEASE-NOTES.md

Issue #3849, opened 2020-03-24
macOS: Multiprocessor sampling stops unexpectedly on macOS Catalina if seaborn was imported


ivlis:
Sampling stops with the following error in the console:
```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Couldn't close file
```
Seems to be related with this issue:
https://github.com/matplotlib/matplotlib/issues/15410

I understand that this is an external package issue, but it makes PyMC3 practically unusable. 

**Please provide any additional information below.**

## Versions and main components

* PyMC3 Version: 3.8
* Theano Version: 1.0.4
* Python Version: 3.7
* Operating system: osX Catalina (only Catalina is affected, Mojave is OK)
* How did you install PyMC3: conda

michaelosthege:
Can you provide a minimal example to reproduce the problem?

And how did you reach the conclusion that it could be related to matplotlib? The error message is rather generic. Do you have a trace?

rosgori:
@ivlis Look at this https://discourse.mc-stan.org/t/issues-with-pystan-multiprocessing-and-subsequently-the-pystan3-pre-release-on-mac-os-catalina/12798

ivlis:
@rosgori I tried to manually edit the file in my matplotlib installation, it seems to have no effect.

@michaelosthege 

Here is a minimal example which demonstrates the problem:

```python
import seaborn as sns
import pymc3 as pm
import theano.tensor as tt
import numpy as np
import theano
theano.config.gcc.cxxflags = "-Wno-c++11-narrowing"


features = np.random.rand(4000, 10, 5)
obs_n = [10]*4000
obs_k = [1]*4000

with pm.Model() as model:
    θ = pm.Normal('θ_0', 0, 1, shape=(10, 5))
    a = pm.Deterministic('a', t t.tensordot(θ, features,  axes=[[0,1], [1,2]]))

    y_hat = pm.Binomial('y_hat', p=pm.math.invlogit(a), n=obs_n, observed=obs_k)

with model:
    trace = pm.sample(target_accept=0.95, tune=1000, draws=200, init='adapt_diag')
```

Sampling does not start
```
$ python example.py
Only 200 samples in chain.
Auto-assigning NUTS sampler...
Initializing NUTS using adapt_diag...
Multiprocess sampling (4 chains in 4 jobs)
NUTS: [θ_0]
[...]

PR #6218, opened 2022-10-14
Set start method to fork for MacOs ARM devices

bchen93:
**What is this PR about?**
Fixing an issue with use of parallel sampling on M1 Macs where sampling with more than 1 core results in broken pipe error (#5339). 

Changing multiprocessing context from forkserver to fork resolves this issue and does not seem to trigger issue (#3849) which was previously resolved by (#3919).
 

**Checklist**

## Major / Breaking Changes
- None Noted
## Bugfixes / New features
Setting the default multiprocessing context to "fork" for ARM Macs, and "forkserver" for intel-based Macs.

## Docs / Maintenance
- ...

### DIFF
--- a/docs/source/learn/core_notebooks/pymc_aesara.ipynb
+++ b/docs/source/learn/core_notebooks/pymc_aesara.ipynb
@@ -2,11 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {
-    "pycharm": {
-     "name": "#%% md\n"
-    }
-   },
+   "metadata": {},
    "source": [
     "(pymc_aesara)=\n",
     "\n",
@@ -20,10 +16,7 @@
   {
    "cell_type": "markdown",
    "metadata": {
-    "heading_collapsed": true,
-    "pycharm": {
-     "name": "#%% md\n"
-    }
+    "heading_collapsed": true
    },
    "source": [
     "## Prepare Notebook\n",
@@ -33,49 +26,23 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": 1,
    "metadata": {
-    "hidden": true,
-    "pycharm": {
-     "name": "#%%\n"
-    }
+    "hidden": true
    },
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "\n",
-      "# Aesara version: 2.7.9\n",
-      "# PyMC version: 4.1.4\n",
-      "\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
     "import aesara\n",
     "import aesara.tensor as at\n",
     "import pymc as pm\n",
     "import matplotlib.pyplot as plt\n",
     "import numpy as np\n",
-    "import scipy.stats\n",
-    "\n",
-    "\n",
-    "print(\n",
-    "    f\"\"\"\n",
-    "# Aesara version: {aesara.__version__}\n",
-    "# PyMC version: {pm.__version__}\n",
-    "\"\"\"\n",
-    ")"
+    "import scipy.stats"
    ]
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "pycharm": {
-     "name": "#%% md\n"
-    }
-   },
+   "metadata": {},
    "source": [
     "## Introduction to Aesara\n",
     "\n",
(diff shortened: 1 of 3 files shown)


ricardoV94:
Would it break stuff for older macOS?

bchen93:
> Would it break stuff for older macOS?

I do not think so. I also tested on a 2016 Macbook Pro with i5 intel chip and it works with fork. 

*edit. Not sure about older operating systems, if that's what you were asking.

bchen93:
After doing a little digging. Fork seemed to be the default start method for multiprocessing in MacOs (and still is for Unix) prior to python 3.8, after which they changed it to spawn (which does not work currently in pymc). There are some instances of fork crashing due to MacOs specific security features (after Os High Sierra (10.13). Which is [resolved](https://developer.apple.com/forums/thread/701358) by setting `export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES`. However, I haven't found it necessary to do this after changing the start method to fork in the pymc paralell_sampling.py script. 

The main reason for using spawn over fork is that it is [safer](https://stackoverflow.com/questions/64095876/multiprocessing-fork-vs-spawn), and it is currently the only method available to windows. 

I'm not sure if given the way that pymc uses multi-processing if the [dangers of using fork](https://www.reddit.com/r/programming/comments/vg3be2/fork_without_exec_is_dangerous_in_large_programs/) really apply. 

If not using fork, then I think there's a lot more work that needs to be done to make spawn/forkserver methods working on M1 Macs.

ghost:
This seems like something users should be able to set to their liking. I'd be in favor of having a passable argument for `sample` that defaults to whatever is deemed to be safest or least likely to fail.

bchen93:
Funnily enough, that is already possible in the pm.sample() method. It accepts an arg mp_ctx, which should be a multiprocessing context object. e.g. using the example model in #5339

```
import numpy as np
import pymc as pm
import arviz as az
import multiprocessing as mp

def build_model():
    data = np.loadtxt(
        pm.get_data("efron-morris-75-data.tsv"), delimiter="\t", skiprows=1, usecols=(2, 3)
    )

    atbats = pm.floatX(data[:, 0])
    hits = pm.floatX(data[:, 1])

    N = len(hits)

    with pm.Model() as model:
        phi = pm.Uniform("phi", lower=0.0, upper=1.0)
        pareto_dist = pm.Pareto.dist(alpha=1.0001, m=1.5)
        kappa = pm.Bound("kappa", pareto_dist, lower=1.0)
        thetas = pm.Beta("thetas", alpha=phi * kappa, beta=(1.0 - phi) * kappa, shape=N)
        ys = pm.Binomial("ys", n=atbats, p=thetas, observed=hits)
    return model

# %%
mp_ctx = mp.get_context('fork') # <--- Get fork, spawn, or forkserver context object
model = build_model()
with model:
    trace = pm.sample(1000, target_accept=0.99, mp_ctx = mp_ctx) <---Tell sampler to use this context

```

However, as it stands, anything run on M1 mac will not work without having to define the start method as fork. Would it be better to set the default start method for ARM macs as fork and for intel macs as forkserver as before?

So essentially changing the code in parallel_sampling.py to: 

```
        if mp_ctx is None or isinstance(mp_ctx, str):
            # Closes issue https://github.com/pymc-devs/pymc/issues/3849
            if platform.system() == "Darwin":
                if platform.processor() == 'arm':
                    mp_ctx = 'fork'
                else: 
                    mp_ctx = "forkserver"

```

ricardoV94:
> However, as it stands, anything run on M1 mac will not work without having to define the start method as fork. Would it be better to set the default start method for ARM macs as fork and for intel macs as forkserver as before?
>
> So essentially changing the code in parallel_sampling.py to:

That sounds reasonable, we can always revert if that starts causing problems. Can we get some more M1 users to confirm that the model in the original issue indeed fails for everyone?

twiecki:
@drbenvincent confirmed this fixed the same issue for him on a different model. I think we can merge this.

ricardoV94:
We shouldn't merge this, but the default based on wether processor is arm or not mentioned in the last comment by the Op: https://github.com/pymc-devs/pymc/pull/6218#issuecomment-1283006746

twiecki:
> We shouldn't merge this

Why not?

bchen93:
> > We shouldn't merge this
> 
> Why not?

I believe @ricardoV94 wanted to change the then committed code s.t. the default for ARM Macs was specified as "fork" and and for intel Macs as "forkserver" (as it has been). I just committed the change to the code. So it can be checked/merged now if everyone is in agreement.

ricardoV94:
@bchen93 Can you clean up the changelist in the top comment? It should mention in bugfixes what you did (use different default when on MacOS ARM). You can leave the other sections empty

We use those for the automatically generated release notes.

bchen93:
> @bchen93 Can you clean up the changelist in the top comment? It should mention in bugfixes what you did (use different default when on MacOS ARM). You can leave the other sections empty
> 
> We use those for the automatically generated release notes.

Done!

ricardoV94:
You seem to have caught unrelated changes in your PR

bchen93:
> You seem to have caught unrelated changes in your PR

Sorry about that. I reverted the commits.

twiecki (approved):

twiecki:
Thanks @bchen93!

PR #6219, opened 2022-10-14
Refactoring `_print_name` for certain RVs and specifying `rv_type`s in their distributions

### LINKED ISSUE #6201: Create PyMC-specific RV Ops that overwrite `_print_name`
Currently, some RV's `_print_name`s contain the abbreviated version of the random variable Op rather than its full name, e.g. Normal, Cauchy, etc. When it come to displaying them in Graphviz, we resort to its class name via [`some_op.__class__.__name__`](https://github.com/pymc-devs/pymc/blob/main/pymc/model_graph.py#L165), which, to my understanding, is a bit hackish. A solution would look like the following:

```python
PyMCNormalRV(NormalRV):
    _print_name = ("Normal", ...)

pymc_normal = PyMCNormalRV()

class Normal(Continuous):
    rv_op = pymc_normal
```

In particular, with the newly added `SymbolicRV`s, this could be helpful (see WIP PR #6149).

CC @ricardoV94 @michaelosthege

larryshamalama:
**What is this PR about?**
This PR creates "new" PyMC `RandomVariable`s for those that have an abbreviate `_print_name` such that we don't have to rely on the RV's class name for Latex and Graphviz displays.

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- 
Subclass several random variables to have a more readable `_print_name`.

Closes #6201

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -17,16 +17,16 @@ import aesara.tensor as at
 import numpy as np
 
 from aesara.tensor.random.basic import (
+    GeometricRV,
+    HyperGeometricRV,
+    NegBinomialRV,
+    PoissonRV,
     RandomVariable,
     ScipyRandomVariable,
     bernoulli,
     betabinom,
     binomial,
     categorical,
-    geometric,
-    hypergeometric,
-    nbinom,
-    poisson,
 )
 from scipy import stats
 
@@ -560,6 +560,13 @@ class DiscreteWeibull(Discrete):
         return check_parameters(res, 0 < q, q < 1, 0 < beta, msg="0 < q < 1, beta > 0")
 
 
+class PyMCPoissonRV(PoissonRV):
+    _print_name = ("Poisson", "\\operatorname{Poisson}")
+
+
+pymc_poisson = PyMCPoissonRV()
+
+
 class Poisson(Discrete):
     R"""
     Poisson log-likelihood.
@@ -605,7 +612,8 @@ class Poisson(Discrete):
     The Poisson distribution can be derived as a limiting case of the
     binomial distribution.
     """
-    rv_op = poisson
+    rv_op = pymc_poisson
+    rv_type = PoissonRV
 
     @classmethod
     def dist(cls, mu, *args, **kwargs):
@@ -672,6 +680,13 @@ class Poisson(Discrete):
         )
 
         return check_parameters(res, 0 <= mu, msg="mu >= 0")
+
+
+class PyMCNegativeBinomialRV(NegBinomialRV):
+    _print_name = ("NegBinom", "\\operatorname{NegBinom}")
+
+
+pymc_nbinom = PyMCNegativeBinomialRV()
 
 
 class NegativeBinomial(Discrete):
@@ -746,7 +761,8 @@ class NegativeBinomial(Discrete):
     n : tensor_like of float
         Alternative number of target success trials (n > 0)
     """
-    rv_op = nbinom
+    rv_op = pymc_nbinom
+    rv_type = NegBinomialRV
 
     @classmethod
     def dist(cls, mu=None, alpha=None, p=None, n=None, *args, **kwargs):
@@ -847,6 +863,13 @@ class NegativeBinomial(Discrete):
         )
 
 
+class PyMCGeometricRV(GeometricRV):
+    _print_name = ("Geometric", "\\operatorname{Geometric}")
+
+
+pymc_geometric = PyMCGeometricRV()
+
+
 class Geometric(Discrete):
     R"""
     Geometric log-likelihood.
@@ -886,7 +909,8 @@ class Geometric(Discrete):
         Probability of success on an individual trial (0 < p <= 1).
     """
 
-    rv_op = geometric
+    rv_op = pymc_geometric
+    rv_type = GeometricRV
 
     @classmethod
     def dist(cls, p, *args, **kwargs):
@@ -956,6 +980,13 @@ class Geometric(Discrete):
         )
 
 
+class PyMCHyperGeometricRV(HyperGeometricRV):
+    _print_name = ("HyperGeometric", "\\operatorname{HyperGeometric}")
+
+
+pymc_hypergeometric = PyMCHyperGeometricRV()
+
+
 class HyperGeometric(Discrete):
     R"""
     Discrete hypergeometric distribution.
@@ -1004,7 +1035,8 @@ class HyperGeometric(Discrete):
         Number of samples drawn from the population (0 <= n <= N)
     """
 
-    rv_op = hypergeometric
+    rv_op = pymc_hypergeometric
+    rv_type = HyperGeometricRV
 
     @classmethod
     def dist(cls, N, k, n, *args, **kwargs):
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -32,7 +32,12 @@ from aesara.raise_op import Assert
 from aesara.sparse.basic import sp_sum
 from aesara.tensor import gammaln, sigmoid
 from aesara.tensor.nlinalg import det, eigh, matrix_inverse, trace
-from aesara.tensor.random.basic import dirichlet, multinomial, multivariate_normal
+from aesara.tensor.random.basic import (
+    DirichletRV,
+    MvNormalRV,
+    multinomial,
+    multivariate_normal,
+)
 from aesara.tensor.random.op import RandomVariable, default_supp_shape_from_params
 from aesara.tensor.random.utils import broadcast_params
 from aesara.tensor.slinalg import Cholesky, SolveTriangular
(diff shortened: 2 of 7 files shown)


pymc/distributions/discrete.py
```diff
@@ -674,6 +681,13 @@ def logcdf(value, mu):
         return check_parameters(res, 0 <= mu, msg="mu >= 0")
 
 
+class PyMCNegativeBinomialRV(NegBinomialRV):
+    _print_name = ("NBinom", "\\operatorname{NBinom}")
```
ricardoV94: Why this name?

pymc/tests/distributions/test_shape_utils.py
```diff
@@ -573,7 +574,7 @@ def test_change_rv_size_default_update():
 def test_change_specify_shape_size_univariate():
     with aesara.config.change_flags(mode=Mode("py")):
         s1, s2 = at.iscalars("s1", "s2")
-        x = at.random.normal(size=(s1, s2))
+        x = pymc_normal(size=(s1, s2))
```
ricardoV94: We don't need to change the object used in tests that are not related to the docstrings

pymc/distributions/truncated.py
```diff
@@ -338,7 +341,7 @@ def truncated_logprob(op, values, *inputs, **kwargs):
     return logp
 
 
-@_truncated.register(NormalRV)
+@_truncated.register(PyMCNormalRV)
```
ricardoV94: This should work on the base class (I hope) and we should keep it like that.

pymc/distributions/multivariate.py
```diff
@@ -1868,7 +1884,7 @@ def rng_fn(self, rng, mu, sigma, *covs, size=None):
         if sigma:
             cov = cov + sigma**2 * np.eye(cov.shape[0])
 
-        x = multivariate_normal.rng_fn(rng=rng, mean=mu, cov=cov, size=size)
+        x = pymc_multivariate_normal.rng_fn(rng=rng, mean=mu, cov=cov, size=size)
```
ricardoV94: No need to change this or above. Only thing that matters is the `rv_op` we use in the distribution classes. All other changes should be undone.

ricardoV94 (changes requested):

twiecki:
Can't we just set `_print_name` on the imported aesara rvs?

ricardoV94:
> Can't we just set `_print_name` on the imported aesara rvs?

That would have side-effects on the Aesara side (if it even works)

larryshamalama:
Letting tests run again. Hopefully nothing fails.

Are there any additional tests to add in light of these new `Op`s?

pymc/tests/distributions/test_truncated.py
```diff
@@ -9,15 +9,15 @@
...
+from pymc.distributions.continuous import Exponential, PyMCNormalRV, TruncatedNormalRV
 from pymc.distributions.shape_utils import change_dist_size
 from pymc.distributions.transforms import _default_transform
 from pymc.distributions.truncated import Truncated, TruncatedRV, _truncated
 from pymc.exceptions import TruncationError
 from pymc.tests.distributions.util import assert_moment_is_expected
 
 
-class IcdfNormalRV(NormalRV):
+class IcdfNormalRV(PyMCNormalRV):
```
larryshamalama: In light of this [failing test](https://github.com/pymc-devs/pymc/actions/runs/3274707968/jobs/5388603941), this test RV needs to inherit from `PyMCNormalRV` because the normal distribution's moment is defined under `Normal` which uses the `pymc_normal` `rv_op`.

ricardoV94: In that case let's use `PyMC` versions in all these test classes? Less confusing I think

larryshamalama: I added a comment to clarify why we are doing this then. Let me know if you think that this is good

pymc/tests/distributions/test_truncated.py
```diff
@@ -6,25 +6,32 @@
...
     """Normal RV that has icdf but not truncated dispatching"""
 
 
-class RejectionNormalRV(NormalRV):
+class RejectionNormalRV(PyMCNormalRV):
     """Normal RV that has neither icdf nor truncated dispatching."""
 
 
+"""
```
ricardoV94: I would also do it for the GeometricRV since we also use a PyMCGeometricRV internally. The tests work regardless but I think it's less confusing for future users.

ricardoV94:
I am not super happy with this (even though I suggested it). It's a bit worrisome that our dispatched methods will work for the subclasses but not the original classes, when all we are doing is changing a name. 

We could do something more clever, where we still dispatch on the Aesara classes but return the PyMC ones from Distribution when those are specified. Then both objects will have logp/logcdf/moments/ etc...

That would have avoided the concerns about the tests where we create new classes.

We could use the `rv_type` property, and dispatch on that. So Normal would be

```python
class Normal(Continuous):
  rv_op = pymc_normal
  rv_type = NormalRV  # Aesara
  ...
```
And then tweak the Metaclass of Distribution to use `rv_type` for dispatching when provided explicitly, and otherwise default to the old behavior of obtaining the `rv_type` from `rv_op`. WDYT?

larryshamalama:
> I am not super happy with this (even though I suggested it). It's a bit worrisome that our dispatched methods will work for the subclasses but not the original classes, when all we are doing is changing a name.

I understand. I also had some doubts, but at least we have a clearer picture of how the changes in this PR would affect the codebase.

> We could do something more clever, where we still dispatch on the Aesara classes but return the PyMC ones from Distribution when those are specified. Then both objects will have logp/logcdf/moments/ etc...
> 
> That would have avoided the concerns about the tests where we create new classes.
> 
> We could use the `rv_type` property, and dispatch on that. So Normal would be
> 
> ```python
> class Normal(Continuous):
>   rv_op = pymc_normal
>   rv_type = NormalRV  # Aesara
>   ...
> ```
> 
> And then tweak the Metaclass of Distribution to use `rv_type` for dispatching when provided explicitly, and otherwise default to the old behavior of obtaining the `rv_type` from `rv_op`. WDYT?

On top of my head, both solutions would work, but I'm not sure how I would dispatch the Aesara classes given `rv_op` to be the newly created ones (like `PyMCNormalRV`).

https://github.com/pymc-devs/pymc/blob/13dfeb2d72a001776b0d0408cf28302e826240cd/pymc/distributions/distribution.py#L107-L114

I believe that the dispatch is being done in these lines and perhaps there is a more clever way than checking if `__class__.__name__.startswith("PyMC")`... Is this what you meant?

Again, without trying much, I feel like using `rv_type` would be an easier solution. However, would this have any spillover complications in other areas of the codebase?

ricardoV94:
@larryshamalama what I meant was something like:

```python
        rv_type = clsdict.setdefault("rv_type", None)

        if rv_type is None and isinstance(rv_op, RandomVariable):
            rv_type = type(rv_op)
            clsdict["rv_type"] = rv_type
```

Then when we set:
```python
class Normal(Continuous):
  rv_op = pymc_normal
  rv_type = NormalRV  # Aesara
```
`pm.Normal` will still return a `PyMCNormalRV`, but all the dispatching will be done on the original baseclass. Dispatching works for subclasses but not parent classes (which makes sense).

larryshamalama:
Okay, I can get to it

pymc/distributions/logprob.py
```diff
@@ -145,7 +145,8 @@ def joint_logp(
 
     The input `var` determines which log-likelihood graph is used and
     `rv_value` is that graph's input parameter.  For example, if `var` is
-    the output of a ``NormalRV`` ``Op``, then the output is a graph of the
+    the output of a ``NormalRV`` ``Op``, which has been recently adapted to
```
ricardoV94: We can leave this unchanged

pymc/tests/distributions/test_truncated.py
```diff
@@ -25,6 +25,13 @@ class RejectionNormalRV(NormalRV):
...
+Note: IcdfNormalRV and RejectionNormalRV inherit from PyMCNormalRV rather
```
ricardoV94: Can remove comment

pymc/tests/test_aesaraf.py
```diff
@@ -405,7 +406,8 @@ def test_rvs_to_value_vars_unvalued_rv():
     res_y = res.owner.inputs[1]
     # Graph should have be cloned, and therefore y and res_y should have different ids
     assert res_y is not y
-    assert res_y.owner.op == at.random.normal
+    assert res_y.owner.op == pymc_normal
+    assert isinstance(res_y.owner.op, NormalRV)
```
ricardoV94, line 408: Second assert is enough since this is not a pm.Distribution test

ricardoV94 (approved):

ricardoV94:
@larryshamalama Can you include the default changelist from the PR template? We use those for the auto-generated release notes

larryshamalama:
Done! Thanks for the suggestion, I will start doing this more often

ricardoV94:
Thanks @larryshamalama, now the `print_name_` for SymbolicRVs should be more useful

Issue #6211, opened 2022-10-13
Flaky test: `test_normal_scalar`


michaelosthege:
If failed in https://github.com/pymc-devs/pymc/actions/runs/3240831110/jobs/5311953513

with

```
_______________________ TestSamplePPC.test_normal_scalar _______________________

self = <pymc.tests.test_sampling.TestSamplePPC object at 0x1240cd4c0>

    def test_normal_scalar(self):
        nchains = 2
        ndraws = 500
        with pm.Model() as model:
            mu = pm.Normal("mu", 0.0, 1.0)
            a = pm.Normal("a", mu=mu, sigma=1, observed=0.0)
            trace = pm.sample(
                draws=ndraws,
                chains=nchains,
            )
    
        with model:
            # test list input
            ppc0 = pm.sample_posterior_predictive(
                10 * [model.initial_point()], return_inferencedata=False
            )
            # # deprecated argument is not introduced to fast version [2019/08/20:rpg]
            ppc = pm.sample_posterior_predictive(trace, var_names=["a"], return_inferencedata=False)
            # test empty ppc
            ppc = pm.sample_posterior_predictive(trace, var_names=[], return_inferencedata=False)
            assert len(ppc) == 0
    
            # test keep_size parameter
            ppc = pm.sample_posterior_predictive(trace, return_inferencedata=False)
            assert ppc["a"].shape == (nchains, ndraws)
    
            # test default case
            idata_ppc = pm.sample_posterior_predictive(trace, var_names=["a"])
            ppc = idata_ppc.posterior_predictive
            assert "a" in ppc
            assert ppc["a"].shape == (nchains, ndraws)
            # mu's standard deviation may have changed thanks to a's observed
            _, pval = stats.kstest(
                (ppc["a"] - trace.posterior["mu"]).values.flatten(), stats.norm(loc=0, scale=1).cdf
            )
>           assert pval > 0.001
E           assert 0.0005430565350[245](https://github.com/pymc-devs/pymc/actions/runs/3240831110/jobs/5311953513#step:7:246)692 > 0.001

pymc/tests/test_sampling.py:653: AssertionError
```

mattiadg:
Hi all, I'm new here and would like to work on this.

michaelosthege:
Sure - please investigate why the `stats.kstest` is even in there.
In the end changing the p-value threshold will always work, but if the assert is not checking anything useful you can remove it alltogether..

mattiadg:
[...]

PR #6220, opened 2022-10-16
Seed flaky test `TestSamplePPC.test_normal_scalar`

mattiadg:
This is related to the issue 
* Remove call to function with unused output
* Add assert to test function with list input
* Add comment to refer to the issue

**What is this PR about?**
The issue Flaky test: test_normal_scalar #6211 reported a Flaky test in test_normal_scalar https://github.com/pymc-devs/pymc/blob/main/pymc/tests/test_sampling.py#L617. I could not reproduce it yet, but I found other minor issues in the same test: calls to the tested function that were not calling any assert. I added an assert after the line with an input different from the other calls, and removed the other line that looks redundant. Under the suggestion of @michaelosthege, I added a comment asking to report further failing of this test under the same issue

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Seed flaky test `TestSamplePPC.test_normal_scalar`

### DIFF
--- a/pymc/tests/test_sampling.py
+++ b/pymc/tests/test_sampling.py
@@ -630,8 +630,8 @@ class TestSamplePPC(SeededTest):
             ppc0 = pm.sample_posterior_predictive(
                 10 * [model.initial_point()], return_inferencedata=False
             )
-            # # deprecated argument is not introduced to fast version [2019/08/20:rpg]
-            ppc = pm.sample_posterior_predictive(trace, var_names=["a"], return_inferencedata=False)
+            assert "a" in ppc0
+            assert len(ppc0["a"][0]) == 10
             # test empty ppc
             ppc = pm.sample_posterior_predictive(trace, var_names=[], return_inferencedata=False)
             assert len(ppc) == 0
@@ -646,6 +646,8 @@ class TestSamplePPC(SeededTest):
             assert "a" in ppc
             assert ppc["a"].shape == (nchains, ndraws)
             # mu's standard deviation may have changed thanks to a's observed
+            # If you observe this test failing, please report it in
+            # the github issue https://github.com/pymc-devs/pymc/issues/6211
             _, pval = stats.kstest(
                 (ppc["a"] - trace.posterior["mu"]).values.flatten(), stats.norm(loc=0, scale=1).cdf
             )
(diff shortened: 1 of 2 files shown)


mattiadg:
Apparently, that call was covering some additional lines but no assert was there. I will understand what it was supposed to test and add an assert accordingly

michaelosthege:
> Apparently, that call was covering some additional lines but no assert was there. I will understand what it was supposed to test and add an assert accordingly

Not sure what you mean. For me this looks all good already.

michaelosthege (approved):

michaelosthege:
Relevant comment: https://github.com/pymc-devs/pymc/issues/6210#issuecomment-1280049067

Looks like the `SeededTest` does not affect posterior predictive sampling?!

mattiadg:
Found the problem, we need to pass explicitly the `random_seed`, actually an object of type RandomState to `pm.sample_posterior_predictive`. This way I get the same samples. New commit follows in a while

pymc/tests/test_sampling.py
```diff
@@ -625,23 +625,30 @@ def test_normal_scalar(self):
...
+            ppc = pm.sample_posterior_predictive(
+                trace, return_inferencedata=False, random_seed=random_state
+            )
             assert ppc["a"].shape == (nchains, ndraws)
 
             # test default case
-            idata_ppc = pm.sample_posterior_predictive(trace, var_names=["a"])
+            idata_ppc = pm.sample_posterior_predictive(
+                trace, var_names=["a"], random_seed=random_state
+            )
```
ricardoV94, line 647: Only this test requires a seed, the others are not checking anything that depends on the draws. Although passing a seed doesn't hurt, it makes reading the git history less straightforward (unless it was a commit about adding random_seed to all tests, and not just this one in particular).

ricardoV94 (changes requested):
Looks great, just a request to minimize the PR changes

pymc/tests/test_sampling.py
mattiadg: I see, removed the seed from all the others.

ricardoV94 (approved):

michaelosthege (approved):

ricardoV94:
Thanks @mattiadg!

PR #6222, opened 2022-10-17
Fix flaky `TestMixture.test_component_choice_random`

### LINKED ISSUE #6210: Flaky test: `test_component_choice_random`
It failed in this run https://github.com/pymc-devs/pymc/actions/runs/3240831110/jobs/5311952415

With 

```
___________________ TestMixture.test_component_choice_random ___________________

self = <pymc.tests.distributions.test_mixture.TestMixture object at 0x7fb12dd6a2f0>

    def test_component_choice_random(self):
        """Test that mixture choices change over evaluations"""
        with Model() as m:
            weights = [0.5, 0.5]
            components = [Normal.dist(-10, 0.01), Normal.dist(10, 0.01)]
            mix = Mixture.dist(weights, components)
        draws = draw(mix, draws=10)
        # Probability of coming from same component 10 times is 0.5**10
>       assert np.unique(draws > 0).size == 2
E       assert 1 == 2
E        +  where 1 = array([False]).size
E        +    where array([False]) = <function unique at 0x7fb14babd360>(array([-10.01041454, -10.00804946, -10.02323384, -10.01368128,\n        -9.99331916,  -9.9980023 ,  -9.98098861, -10.00484836,\n        -9.993316  , -10.00568352]) > 0)
E        +      where <function unique at 0x7fb14babd360> = np.unique

pymc/tests/distributions/test_mixture.py:338: AssertionError
```

bherwerth:
**What is this PR about?**
Fixes #6210:
* Use a seed for taking draws
* Increase number of draws to 20

**Checklist**
+ [x] Are the changes covered by tests and docstrings? (does not apply)

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Fix flaky `TestMixture.test_component_choice_random`

### DIFF
--- a/pymc/tests/distributions/test_mixture.py
+++ b/pymc/tests/distributions/test_mixture.py
@@ -329,17 +329,13 @@ class TestMixture(SeededTest):
 
     def test_component_choice_random(self):
         """Test that mixture choices change over evaluations"""
-        p_first_component = 0.5
-        n_draws = 200
         with Model() as m:
-            weights = [p_first_component, 1.0 - p_first_component]
+            weights = [0.5, 0.5]
             components = [Normal.dist(-10, 0.01), Normal.dist(10, 0.01)]
             mix = Mixture.dist(weights, components)
-        draws = draw(mix, draws=n_draws, random_seed=self.get_random_state())
-        k_first_component = np.sum(draws < 0)
-        # The number of draws from the first component should follow a binomial distribution
-        result_binomtest = st.binomtest(k_first_component, n_draws, p_first_component)
-        assert result_binomtest.pvalue > 0.01
+        draws = draw(mix, draws=10, random_seed=self.get_random_state())
+        # Probability of coming from same component 10 times is 0.5**10
+        assert np.unique(draws > 0).size == 2
 
     @pytest.mark.parametrize(
         "comp_dists",
--- a/pymc/tests/distributions/test_mixture.py
+++ b/pymc/tests/distributions/test_mixture.py
@@ -329,13 +329,17 @@ class TestMixture(SeededTest):
 
     def test_component_choice_random(self):
         """Test that mixture choices change over evaluations"""
+        p_first_component = 0.5
+        n_draws = 200
         with Model() as m:
-            weights = [0.5, 0.5]
+            weights = [p_first_component, 1.0 - p_first_component]
             components = [Normal.dist(-10, 0.01), Normal.dist(10, 0.01)]
             mix = Mixture.dist(weights, components)
-        draws = draw(mix, draws=10, random_seed=self.get_random_state())
-        # Probability of coming from same component 10 times is 0.5**10
-        assert np.unique(draws > 0).size == 2
+        draws = draw(mix, draws=n_draws, random_seed=self.get_random_state())
+        k_first_component = np.sum(draws < 0)
+        # The number of draws from the first component should follow a binomial distribution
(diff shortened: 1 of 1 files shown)


pymc/tests/distributions/test_mixture.py
```diff
@@ -329,13 +329,17 @@ def test_list_multivariate_components_deterministic_weights(self, weights, compo
...
+        p_first_component = 0.5
+        n_draws = 200
         with Model() as m:
-            weights = [0.5, 0.5]
+            weights = [p_first_component, 1.0 - p_first_component]
             components = [Normal.dist(-10, 0.01), Normal.dist(10, 0.01)]
             mix = Mixture.dist(weights, components)
-        draws = draw(mix, draws=10)
-        # Probability of coming from same component 10 times is 0.5**10
-        assert np.unique(draws > 0).size == 2
```
ricardoV94, line 338: I would rather keep the old assert, it gives me more confidence that the component actually changed. I don't know if the binomial test would fail if all draws came from one component (and even if it did, it would be less readable), which is what this test wants to confirm.

ricardoV94 (changes requested):
Looks great, but I would rather keep the old check than the new one. What do you think?

pymc/tests/distributions/test_mixture.py
bherwerth: It fails if all draws come from the same component because that's very unlikely to happen under the given binomial:

```python
stats.binomtest(0, 200, 0.5).pvalue  # 1.2446030555722283e-60
```

The assertion is stronger than the original one because it would also fail, say, 50 vs 150 draws from the components:
```python
stats.binomtest(50, 200, 0.5).pvalue # 8.393020875604775e-13
```

However, I fully agree, if the purpose of the test is to only check that both components exist in the draws, then the original version is simpler and to be preferred.

pymc/tests/distributions/test_mixture.py
```diff
@@ -333,8 +333,8 @@ def test_component_choice_random(self):
             weights = [0.5, 0.5]
             components = [Normal.dist(-10, 0.01), Normal.dist(10, 0.01)]
             mix = Mixture.dist(weights, components)
-        draws = draw(mix, draws=10)
-        # Probability of coming from same component 10 times is 0.5**10
+        draws = draw(mix, draws=200, random_seed=self.get_random_state())
```
ricardoV94: This should suffice? Trying to keep the tests lean
```suggestion
        draws = draw(mix, draws=20, random_seed=self.get_random_state())
```

pymc/tests/distributions/test_mixture.py
```diff
@@ -333,8 +333,8 @@ def test_component_choice_random(self):
             weights = [0.5, 0.5]
             components = [Normal.dist(-10, 0.01), Normal.dist(10, 0.01)]
             mix = Mixture.dist(weights, components)
-        draws = draw(mix, draws=10)
-        # Probability of coming from same component 10 times is 0.5**10
+        draws = draw(mix, draws=20, random_seed=self.get_random_state())
+        # Probability of coming from same component 200 times is 0.5**200
```
ricardoV94, line 337:
```suggestion
        # Probability of coming from same component 20 times is 0.5**20
```

ricardoV94 (approved):

ricardoV94:
Thanks @bherwerth!

PR #6227, opened 2022-10-18
Refactor EulerMaruyama to work in v4

### LINKED ISSUE #4642: Port timeseries distributions to V4
```python
import pymc3 as pm
import numpy as np

returns = pd.read_csv(pm.get_data("SP500.csv"), index_col="Date")
returns["change"] = np.log(returns["Close"]).diff()
returns = returns.dropna().astype('float32')
returns.head()

with pm.Model() as model_pymc3:
    step_size = pm.Exponential("step_size", 10.)
    volatility = pm.GaussianRandomWalk("volatility", sigma=step_size, 
                                       size=returns.shape[0], 
                                       init=pm.Normal.dist(0, step_size))
    nu = pm.Exponential("nu", 0.1)
    obs = pm.StudentT(
        "obs", nu=nu, sigma=np.exp(volatility), observed=returns["change"]
    )
```

yields:

```
<ipython-input-11-3f1576c8cf84> in <module>
      1 with pm.Model(check_bounds=False) as model_pymc3:
      2     step_size = pm.Exponential("step_size", 10.)
----> 3     volatility = pm.GaussianRandomWalk("volatility", sigma=step_size, 
      4                                        size=returns.shape[0],
      5                                        init=pm.Normal.dist(0, step_size))

~/projects/pymc/pymc3/distributions/distribution.py in __new__(cls, name, *args, **kwargs)
    161         transform = kwargs.pop("transform", UNSET)
    162 
--> 163         rv_out = cls.dist(*args, rng=rng, **kwargs)
    164 
    165         return model.register_rv(rv_out, name, data, total_size, dims=dims, transform=transform)

TypeError: dist() missing 1 required positional argument: 'dist_params'
```

junpenglao:
Close #4642

## Major / Breaking Changes
- ...

## Bugfixes / New features
- Refactor `EulerMaruyama` distribution

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -912,8 +912,8 @@ class EulerMaruyama(Distribution):
     sde_pars: tuple
         parameters of the SDE, passed as ``*args`` to ``sde_fn``
     init_dist : unnamed distribution, optional
-        Scalar or vector distribution for initial values. Unnamed refers to distributions
-        created with the ``.dist()`` API. Distributions should have shape (*shape[:-1]).
+        Scalar distribution for initial values. Unnamed refers to distributions created with
+        the ``.dist()`` API. Distributions should have shape (*shape[:-1]).
         If not, it will be automatically resized. Defaults to pm.Normal.dist(0, 100, shape=...).
 
         .. warning:: init_dist will be cloned, rendering it independent of the one passed as input.
@@ -953,9 +953,9 @@ class EulerMaruyama(Distribution):
                     f"got {type(init_dist)}"
                 )
             check_dist_not_registered(init_dist)
-            if init_dist.owner.op.ndim_supp > 1:
+            if init_dist.owner.op.ndim_supp > 0:
                 raise ValueError(
-                    "Init distribution must have a scalar or vector support dimension, ",
+                    "Init distribution must have a scalar support dimension, ",
                     f"got ndim_supp={init_dist.owner.op.ndim_supp}.",
                 )
         else:
--- a/pymc/distributions/timeseries.py
+++ b/pymc/distributions/timeseries.py
@@ -27,7 +27,6 @@ from aesara.tensor import TensorVariable
 from aesara.tensor.random.op import RandomVariable
 
 from pymc.aesaraf import constant_fold, floatX, intX
-from pymc.distributions import distribution
 from pymc.distributions.continuous import Normal, get_tau_sigma
 from pymc.distributions.distribution import (
     Distribution,
@@ -461,7 +460,7 @@ class AR(Distribution):
         process.
     init_dist : unnamed distribution, optional
         Scalar or vector distribution for initial values. Unnamed refers to distributions
-         created with the ``.dist()`` API. Distributions should have shape (*shape[:-1], ar_order).
+        created with the ``.dist()`` API. Distributions should have shape (*shape[:-1], ar_order).
         If not, it will be automatically resized. Defaults to pm.Normal.dist(0, 100, shape=...).
 
         .. warning:: init_dist will be cloned, rendering it independent of the one passed as input.
@@ -914,7 +913,7 @@ class EulerMaruyama(Distribution):
         parameters of the SDE, passed as ``*args`` to ``sde_fn``
     init_dist : unnamed distribution, optional
         Scalar or vector distribution for initial values. Unnamed refers to distributions
-         created with the ``.dist()`` API. Distributions should have shape (*shape[:-1], ar_order).
+        created with the ``.dist()`` API. Distributions should have shape (*shape[:-1]).
         If not, it will be automatically resized. Defaults to pm.Normal.dist(0, 100, shape=...).
 
         .. warning:: init_dist will be cloned, rendering it independent of the one passed as input.
@@ -953,7 +952,7 @@ class EulerMaruyama(Distribution):
                     f"Init dist must be a distribution created via the `.dist()` API, "
                     f"got {type(init_dist)}"
                 )
-                check_dist_not_registered(init_dist)
+            check_dist_not_registered(init_dist)
             if init_dist.owner.op.ndim_supp > 1:
                 raise ValueError(
                     "Init distribution must have a scalar or vector support dimension, ",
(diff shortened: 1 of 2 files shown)


pymc/distributions/timeseries.py
```diff
@@ -893,39 +911,149 @@ class EulerMaruyama(distribution.Continuous):
...
-    def _distr_parameters_for_repr(self):
-        return ["dt"]
+    @classmethod
+    def rv_op(cls, init_dist, steps, sde_pars, dt, sde_fn, size=None):
+        # Init dist should have shape (*size,)
+        if size is not None:
+            batch_size = size
+        else:
+            batch_size = at.broadcast_shape(*sde_pars, init_dist)
+        init_dist = change_dist_size(init_dist, batch_size)
```
ricardoV94, line 981: Didn't you allow for `init_dist` to be scalar or vector? In that case the `change_dist_size` must have different inputs depending on the support dimensionality?

pymc/distributions/timeseries.py
```diff
@@ -893,39 +911,149 @@ class EulerMaruyama(distribution.Continuous):
...
+    sde_pars_broadcast = [x[..., None] for x in sde_pars]
+    xtm1 = x[..., :-1]
+    xt = x[..., 1:]
+    f, g = op.sde_fn(xtm1, *sde_pars_broadcast)
+    mu = xtm1 + op.dt * f
+    sigma = at.sqrt(op.dt) * g
+    # Compute and collapse logp across time dimension
+    sde_logp = at.sum(logp(Normal.dist(mu, sigma), xt), axis=-1)
+    init_logp = logp(init_dist, x[..., 0])
+    return init_logp + sde_logp
```
ricardoV94, line 1059: Should we return a concatenation of `init_logp` and `sde_logp`? Also variable support dimensionality of `init_logp` might need to be handled, otherwise one might try to broadcast the other incorrectly.

ricardoV94 (commented):
Looks great @junpenglao, some questions below. 

Also, would it be too much trouble to split the tiny AR fixes from the new distribution refactoring into just 2 commits so that we don't have to squash everything together?

pymc/distributions/timeseries.py
ricardoV94: It seems simpler to allow only for scalar `init_dist`. Note that in the `AR` there is more logic at the resize `init_dist` and `logp` steps to make sure both are properly supported

pymc/tests/distributions/test_timeseries.py
```diff
@@ -830,37 +830,99 @@ def test_change_dist_size(self):
...
-        trace = sample(init="advi+adapt_diag", chains=1)
-
-    ppc = sample_posterior_predictive(trace, model=model)
-
-    p95 = [2.5, 97.5]
-    lo, hi = np.percentile(trace[lamh], p95, axis=0)
-    assert (lo < lam) and (lam < hi)
-    lo, hi = np.percentile(ppc["zh"], p95, axis=0)
-    assert ((lo < z) * (z < hi)).mean() > 0.95
+class TestEulerMaruyama:
```
ricardoV94, line 833: Better to pass `init_dist` everywhere to avoid the warning in the tests. We recently had some large PRs trying to cleanup these sort of warnings from tests

ricardoV94, line 833: Need some tests for multivariate `init_dist` if you want to support them

ricardoV94:
Also, would it be too much work to implement a `moment`?

junpenglao:
Thanks for the comments.

> Also, would it be too much work to implement a moment?

I dont think this possible as user can input arbitrary sde_fn with moment unspecified.

pymc/distributions/timeseries.py
junpenglao: > Should we return a concatenation of init_logp and sde_logp?

Nope because the time dimension is collapsed

> Also variable support dimensionality of init_logp might need to be handled, otherwise one might try to broadcast the other incorrectly.

I think it is fine as is, because:
1. init_logp is scalar and data is (n_batch, t), logp is broadcasted to (n_batch, )
2. init_logp is multivariate and data is (n_batch, t), logp is (n_batch, )

The only potential error is init_logp is multivariate, all the `sde_pars` are scalar and data is (t, ), maybe we can have a test to make sure the rank of `init_dist` is lower or equal to `sde_pars`, WDYT?

ricardoV94: If `init_dist` is multivariate, it must have support_shape `(n_batch[-1],)` (I am assuming this, could be wrong). If `n_batch` is a vector, say `n_batch.shape = (5,)` the `init_logp` will be a scalar, and you will be over counting it 5 times when you add it to the `sde_logp` vector.

ricardoV94: In the `AR` this is okay, because there are multiple initial points per timeseries, depending on the `ar_order`. As mentioned in the docstrings the `init_dist` should have shape `(*shape[:-1], ar_order)`, so we can align the logp whether users pass univariate or multivariate `init_dist` (we just do an extra reduction if it's univariate). 

Does this distribution also have multiple initial points per timeseries?

junpenglao: > Does this distribution also have multiple initial points per timeseries?

It shouldn't. I am only thinking about `n_batch` being a scalar int here.
You are right (2) does not work because the logp will be `()`, maybe we should disable multivariate support for `init_dist` and add that later if needed. WDYT?

ricardoV94: > You are right (2) does not work because the logp will be (), maybe we should disable multivariate support for init_dist and add that later if needed. WDYT?

Yes, I think you can't actually have a multivariate `init_dist` here, otherwise the logp across multiple timeseries is no longer independent, and you would need to collapse those as well (after which you could indeed add the scalar `init_logp` with scalar `sde_logp`)

junpenglao: SG - I will update the doc string etc to make clear it should be (a batch of) scalar time series (and in the future we can make it into multivariate by having something like `sde_logp = at.sum(logp(MvNormal.dist(mu, cov), xt), axis=-1)`

junpenglao:
> Also, would it be too much trouble to split the tiny AR fixes from the new distribution refactoring into just 2 commits so that we don't have to squash everything together?

Sure: https://github.com/pymc-devs/pymc/pull/6240

ricardoV94 (approved):
This is great. We have finally achieved V3 parity xD

ricardoV94:
@junpenglao We use the default PR bullet points for the automatically generated release notes. I'll add it manually ;)

junpenglao:
> @junpenglao We use the default PR bullet points for the automatically generated release notes. I'll add it manually ;)

Thanks! I mindlessly deleted everything when open the PR.

PR #6231, opened 2022-10-19
Fix latex repr of symbolic distributions

### LINKED ISSUE #6202: Improve Latex representation of symbolic distributions
Symbolic distributions often (or *always(?)*) use other distributions as input. As such, when displaying the data-generating model in Latex, it does not look properly rendered, although we can understand what the model is. Here are two examples:

```python
with pm.Model() as censored_model:
    mu = pm.Normal("mu", 0., 5.)
    sigma = pm.HalfCauchy("sigma", 2.5)
    normal_dist = pm.Normal.dist(mu=mu, sigma=sigma)
    censored_normal = pm.Censored("censored_normal", normal_dist, lower=-2., upper=2., observed=[1, 0, 0.5])
```
yields
Screen Shot 2022-10-11 at 9 32 16 AM

and

```python
with pm.Model() as mix_model:
    w = pm.Dirichlet("w", [1, 1])
    mix = pm.Mixture("mix", w=w, comp_dists=[pm.Normal.dist(0., 5.), pm.StudentT.dist(7.)])
```

yields
Screen Shot 2022-10-11 at 9 33 08 AM

mattiadg:
Fix #6202 

* Do not put \operatorname inside \text
* add test

**What is this PR about?**
When a symbolic distribution is defined as parameter of another one, the latex representation is broken.
This PR aims to solve this.

I think that the new failing tests are testing the wrong behavior, this is a draft because I would like someone more experienced with this to check on them. 
What I'm doing is to remove `\operatorname` from inside `\text` because it's wrong.

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- Fix broken latex representation of distributions inside other distributions

## Docs / Maintenance
- ...


mattiadg:
Is anything else needed here?

mattiadg:
I've already solved conflicts twice because people keep adding wrong tests for the latex printing.
That functionality is currently broken as #6202 shows, so the tests must be wrong in order to fail.
Please, have a loot at this PR before merging additional code in test_printing.py.

ricardoV94:
> I've already solved conflicts twice because people keep adding wrong tests for the latex printing. That functionality is currently broken as #6202 shows, so the tests must be wrong in order to fail. Please, have a loot at this PR before merging additional code in test_printing.py.

That sometimes happens, a reviewer may be able to get to another PR and merge it before yours. Sorry for the trouble, but it's also something one must get used to :)

mattiadg:
Yes, I understand that, I don't want to look upset. I'm just saying that
the current tests in that file, and the ones are being added to it, are
(partially) testing the wrong behavior, which is the reason why that issue
exists in the first place.

pymc/printing.py
```diff
@@ -215,9 +217,20 @@ def _expand(x):
...
+    else:
+        return r"f(" + ", ".join([n.strip("$") for n in names]) + ")"
+
+
+def _latex_text_format(text: str) -> str:
+    if "operatorname" in text:
+        return text
     else:
-        return r"f(" + ", ".join(names) + ")"
+        return r"\text{" + text + "}"
```
Spaak, line 233: Many lines are now flagged as not covered by tests. But it seems to me that your `test_model_latex_repr_nested` should have tested these, correct? Perhaps that test is not triggered somehow. Would be good to fix before merging.

pymc/printing.py
```diff
@@ -215,9 +217,20 @@ def _expand(x):
...
+            r"f("
+            + ",~".join([_latex_text_format(_latex_escape(n.strip("$"))) for n in names])
+            + ")"
+        )
+    else:
+        return r"f(" + ", ".join([n.strip("$") for n in names]) + ")"
+
+
+def _latex_text_format(text: str) -> str:
+    if "operatorname" in text:
```
Spaak:
```suggestion
    if "\\operatorname{" in text:
```

Spaak (changes requested):
LGTM in general, though the testing should be fixed.

pymc/printing.py
mattiadg: I pushed this change and fixed the failing test. One reference value was wrong after merging.
Now, I check why the coverage goes down.

Spaak (approved):
LGTM!

Spaak:
Looks good to me, thanks for adding the tests so quickly. I'll leave merging to someone else (eg @ricardoV94) because it's been a while for me.

mattiadg:
> Looks good to me, thanks for adding the tests so quickly. I'll leave merging to someone else (eg @ricardoV94) because it's been a while for me.

That was easy, I added as tests the exact same distributions used in the issue to show the problem. This way it's also possible to reference back to them, should something go wrong in the future.
However, I still need to understand the two branches of Codecov: `patch` and `project`. I am still not sure why `patch` became lower initially

pymc/tests/test_printing.py
```diff
@@ -178,3 +178,43 @@ def test_str_repr(self):
...
+    expected = [
+        "$$",
+        "\\begin{array}{rcl}",
+        "\\text{w} &\\sim & "
+        "\\operatorname{Dirichlet}(\\text{<constant>})\\\\\\text{mix} &\\sim & "
+        "\\operatorname{MarginalMixture}(\\text{w},~\\operatorname{Normal}(0,~5),~\\operatorname{StudentT}(7,~0,~1))",
+        "\\end{array}",
+        "$$",
+    ]
+    assert [line.strip() for line in latex_repr.split("\n")] == expected
```
ricardoV94: Does this check do anything that the corrected pre-existing checks do not?

pymc/printing.py
```diff
@@ -215,9 +217,20 @@ def _expand(x):
...
+        return (
+            r"f("
+            + ",~".join([_latex_text_format(_latex_escape(n.strip("$"))) for n in names])
+            + ")"
+        )
+    else:
+        return r"f(" + ", ".join([n.strip("$") for n in names]) + ")"
+
+
+def _latex_text_format(text: str) -> str:
```
larryshamalama, line 229: Okay for now, but I have a sense that this can be simplified since you're always using it on an output of `_latex_escape`

larryshamalama (approved):
Looks good, thanks for your help! Glad that you were able to solve the other conflicts following the merge of #6242

pymc/tests/test_printing.py
mattiadg: They test additionally `HalfCauchy`, `StudentT` and `Model` itself, but I also added these two tests is to keep under track the models reported in the issue.

pymc/printing.py
mattiadg: I can add the call to `_latex_escape(text.strip("$")` inside `_latex_text_format` and call this also in the other 2 occurrences of `_latex_escape`. Then, we can also remove the naked `\text{` in there.

michaelosthege:
@mattiadg there appears to be one more conflict?

Would be great to push this over the finish line as we have a release coming up!

mattiadg:
Sure, I will do it this evening or tomorrow

mattiadg:
> @mattiadg there appears to be one more conflict?
> 
> Would be great to push this over the finish line as we have a release coming up!

done

michaelosthege (approved):

michaelosthege:
Nice, thank you for fixing these @mattiadg !

PR #6232, opened 2022-10-20
Adding option to include transformed variables in InferenceData

dfm:
This tiny PR adds an optional parameter to the ArviZ `InferenceData` backend to allow the (optional, if requested) inclusion of transformed parameters in the returned `InferenceData` object's `posterior` extension. There is discussion about building out a more formal interface for this in future versions of ArviZ (see: https://github.com/arviz-devs/arviz/issues/230, https://github.com/arviz-devs/arviz/issues/2056, https://github.com/arviz-devs/arviz/pull/2086), but it would be a _huge_ quality of life improvement (for me!) to have this available as an advanced user feature. I regularly use this for debugging sampling issues and for post-sampling plotting of model elements - it is an absolutely crucial component of my workflow! As far as I know, the only way to access these traces currently is to use the `MultiTrace` backend, which is annoying, since I'll always end up wanting to convert it to an `InferenceData` manually anyways.

## Bugfixes / New features
- Adds a optional argument to `to_inference_data` allowing the inclusion of transformed/unconstrained parameters in the returned `InferenceData` object

### DIFF
--- a/pymc/tests/backends/test_arviz.py
+++ b/pymc/tests/backends/test_arviz.py
@@ -622,18 +622,14 @@ class TestDataPyMC:
             pm.Uniform("p", 0, 1)
 
             # First check that the default is to exclude the transformed variables
-            inference_data = pm.sample(
-                500,
-                chains=2,
-                return_inferencedata=True,
-            )
+            sample_kwargs = dict(tune=5, draws=7, chains=2, cores=1)
+            inference_data = pm.sample(**sample_kwargs, step=pm.Metropolis())
             assert "p_interval__" not in inference_data.posterior
 
             # Now check that they are included when requested
             inference_data = pm.sample(
-                500,
-                chains=2,
-                return_inferencedata=True,
+                **sample_kwargs,
+                step=pm.Metropolis(),
                 idata_kwargs={"include_transformed": True},
             )
             assert "p_interval__" in inference_data.posterior
--- a/pymc/tests/backends/test_arviz.py
+++ b/pymc/tests/backends/test_arviz.py
@@ -279,7 +279,7 @@ class TestDataPyMC:
         np.testing.assert_array_equal(idata.observed_data.coords["date"], coords["date"])
         np.testing.assert_array_equal(idata.observed_data.coords["city"], coords["city"])
 
-    def test_ovewrite_model_coords_dims(self):
+    def test_overwrite_model_coords_dims(self):
         """Check coords and dims from model object can be partially overwritten."""
         dim1 = ["a", "b"]
         new_dim1 = ["c", "d"]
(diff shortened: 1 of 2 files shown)


lucianopaz (approved):
LGTM

OriolAbril:
Can you also comment on https://github.com/arviz-devs/arviz/pull/2086 if that would be helpful or not and why? 

Will then that argument be used to indicate if the unconstrained group should be present? Or how will that work?

Also, for tracking purposes, related to #5160

dfm:
Thanks @lucianopaz and @OriolAbril!

> Can you also comment on https://github.com/arviz-devs/arviz/pull/2086 if that would be helpful or not and why?

I think https://github.com/arviz-devs/arviz/pull/2086 is good. For my workflow, it's actually easier if the transformed variables live in the `posterior` group (since, for example, I often want to plot the covariance between some transformed and un-transformed parameters; whichever parameters I'm sampling in), but I could certainly join those two groups. So I'd say that it would be a fine solution for me, but I'd probably always prefer to use something like what is proposed here.

> Will then that argument be used to indicate if the unconstrained group should be present? Or how will that work?

I don't think this would be the right parameter (or interface) to use for the proposal in https://github.com/arviz-devs/arviz/pull/2086. My proposal would be to implement the `kwarg` that I've added here since it's barely part of the public API, but makes something possible that it currently impossible. Then, more thought should probably be given to the best interface when implementing support for the `unconstrained_posterior` group (I expect this probably won't be via the awkward `idata_kwargs` interface!).

pymc/tests/backends/test_arviz.py
```diff
@@ -617,6 +617,27 @@ def test_variable_dimension_name_collision(self):
...
+            )
+            assert "p_interval__" not in inference_data.posterior
+
+            # Now check that they are included when requested
+            inference_data = pm.sample(
+                500,
+                chains=2,
+                return_inferencedata=True,
+                idata_kwargs={"include_transformed": True},
+            )
```
michaelosthege: Just to speed up the test (by a lot)

```suggestion
            sample_kwargs = dict(tune=5, draws=7, chains=2, cores=1)
            inference_data = pm.sample(**sample_kwargs, step=pm.Metropolis())
            assert "p_interval__" not in inference_data.posterior

            # Now check that they are included when requested
            inference_data = pm.sample(
                **sample_kwargs,
                step=pm.Metropolis(),
                idata_kwargs={"include_transformed": True},
            )
```

michaelosthege (commented):
While generally I'm not a fan of the `InferenceDataConverter`, I do understand the motiviation behind your PR, and I think the implementation is fine.

LGTM with one nitpick about test performance

michaelosthege (approved):

ricardoV94:
Thanks @dfm

PR #6235, opened 2022-10-20
Check that sampler vars correspond to value variables in the model

### LINKED ISSUE #6223: Raise informative error when non-value variables are assigned to samplers

In PyMC3, the `pm.inputvars` would (correctly, I think!) identify and the top-level input variables, whereas in PyMC v4 (`main` branch and released), it basically seems to be a no-op. To be more explicit, the following passes in PyMC3 (things get a bit more complicated with transformed variables - that's a different but related conversation!):

```python
with pm.Model():
    x = pm.Normal("x")
    assert pm.inputvars([x]) == [x]

    y = pm.Normal("y")
    assert len(inputvars([x, y])) == 2
    assert x in pm.inputvars([x, y])
    assert y in pm.inputvars([x, y])

    z = x + y
    assert len(pm.inputvars([z])) == 2
    assert x in pm.inputvars([z])
    assert y in pm.inputvars([z])

    w = pm.Deterministic("w", x + y)
    assert len(pm.inputvars([w])) == 2
    assert x in pm.inputvars([w])
    assert y in pm.inputvars([w])
```

but in PyMC (v4) all these asserts fail, since every call to `pm.inputvars` returns an empty list. This can be mitigated to some extent using `Model.rvs_to_values` (see [here](https://github.com/pymc-devs/pymc/blob/9822ce5c7d23ff27734bf1cdebc7efd34e5e6480/pymc/tuning/starting.py#L99-L101), for example), for reasons that I don't totally understand - adding this makes things work correctly for `x` and `y` in the example above, but still returns the empty list for `z` and `w`.

The real issue seems to be that `aesara.graph.basic.graph_inputs` walks up the graph too far:

```python
from aesara.graph.basic import graph_inputs
with pm.Model():
    x = pm.Uniform("x")
    print(list(graph_inputs([x])))

# prints: [RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F4FFBFCC7D0>), TensorConstant{[]}, TensorConstant{11}, TensorConstant{0.0}, TensorConstant{1.0}]
```

I'm opening the issue here instead of over on Aesara, because I think that Aesara is providing the intended behavior, but PyMC is incorrectly using the results.

**What is affected?** I noticed this when using `pm.find_MAP`, but this also leads to strange behavior in `Model.compile_fn` where `inputvars` is used to determine the relevant inputs. This function shows up in a few other spots throughout the codebase where I think issues could pop up.

I tried to look into fixing this, but I don't really follow the core back end details enough to know what the best approach is!

ricardoV94:
Closes #6223 

I decided to not allow the old V3 behavior of passing intermediate variables/ Deterministic because I think that does something strange. Those variables are not the ones being sampled/maximized but instead the underlying input value variables are. 

If a user wants to fit the variables that underlie a given graph, but he doesn't know what those are, they should do something like `pm.find_MAP(vars=pm.inputvars(pm.aesaraf.rvs_to_value_vars(graph)))`

**Update**: It is now allowed but a warning is emitted

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Check that only random variables are assigned to samplers and `find_MAP`

### DIFF
--- a/pymc/util.py
+++ b/pymc/util.py
@@ -21,6 +21,7 @@ import cloudpickle
 import numpy as np
 import xarray
 
+from aesara import Variable
 from aesara.compile import SharedVariable
 from cachetools import LRUCache, cachedmethod
 
@@ -441,3 +442,42 @@ def _get_seeds_per_chain(
         )
 
     return random_state
+
+
+def get_value_vars_from_user_vars(
+    vars: Union[Variable, Sequence[Variable]], model
+) -> List[Variable]:
+    """This function converts user "vars" input into value variables
+
+    More often than not, users will pass random variables, and we will extract the
+    respective value variables, but we also allow for the input to already be value
+    variables, in case the function is called internally or by a "super-user"
+
+    Returns
+    -------
+    value_vars: list of TensorVariable
+        List of model value variables that correspond to the input vars
+
+    Raises
+    ------
+    ValueError:
+        If any of the provided variables do not correspond to any model value variable
+    """
+    if not isinstance(vars, Sequence):
+        # Single var was passed
+        value_vars = [model.rvs_to_values.get(vars, vars)]
+    else:
+        value_vars = [model.rvs_to_values.get(var, var) for var in vars]
+
+    # Check that we only have value vars from the model
+    model_value_vars = model.value_vars
+    notin = [v for v in value_vars if v not in model_value_vars]
+    if notin:
+        notin = list(map(get_var_name, notin))
+        # We mention random variables, even though the input may be a wrong value variable
+        # because most users don't know about that duality
+        raise ValueError(
+            "The following variables are not random variables in the model: " + str(notin)
+        )
+
+    return value_vars
--- a/conda-envs/environment-dev.yml
+++ b/conda-envs/environment-dev.yml
@@ -38,5 +38,5 @@ dependencies:
 - watermark
 - polyagamma
 - sphinx-remove-toctrees
-- mypy
+- mypy=0.982
 - types-cachetools
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -27,5 +27,5 @@ dependencies:
 - pre-commit>=2.8.0
 - pytest-cov>=2.5
 - pytest>=3.0
-- mypy
+- mypy=0.982
 - types-cachetools
--- a/conda-envs/windows-environment-dev.yml
+++ b/conda-envs/windows-environment-dev.yml
@@ -35,5 +35,5 @@ dependencies:
 - sphinx>=1.5
 - watermark
 - sphinx-remove-toctrees
-- mypy
+- mypy=0.982
 - types-cachetools
--- a/conda-envs/windows-environment-test.yml
+++ b/conda-envs/windows-environment-test.yml
@@ -28,5 +28,5 @@ dependencies:
 - pre-commit>=2.8.0
 - pytest-cov>=2.5
 - pytest>=3.0
-- mypy
+- mypy=0.982
 - types-cachetools
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -261,11 +261,11 @@ def replace_rvs_in_graphs(
 
 
 def rvs_to_value_vars(
-    graphs: Iterable[TensorVariable],
+    graphs: Iterable[Variable],
     apply_transforms: bool = True,
-    initial_replacements: Optional[Dict[TensorVariable, TensorVariable]] = None,
+    initial_replacements: Optional[Dict[Variable, Variable]] = None,
     **kwargs,
-) -> List[TensorVariable]:
+) -> List[Variable]:
     """Clone and replace random variables in graphs with their value variables.
 
     This will *not* recompute test values in the resulting graphs.
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -69,6 +69,7 @@ from pymc.util import (
     UNSET,
     WithMemoization,
     get_transformed_name,
+    get_value_vars_from_user_vars,
     get_var_name,
     treedict,
     treelist,
@@ -617,6 +618,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         if grad_vars is None:
             grad_vars = self.continuous_value_vars
         else:
+            grad_vars = get_value_vars_from_user_vars(grad_vars, self)
             for i, var in enumerate(grad_vars):
                 if var.dtype not in continuous_types:
                     raise ValueError(f"Can only compute the gradient of continuous types: {var}")
--- a/pymc/smc/kernels.py
+++ b/pymc/smc/kernels.py
@@ -28,7 +28,6 @@ from scipy.stats import multivariate_normal
 from pymc.aesaraf import (
     compile_pymc,
     floatX,
-    inputvars,
     join_nonshared_inputs,
     make_shared_replacements,
 )
@@ -160,7 +159,7 @@ class SMC_KERNEL(ABC):
         self.rng = np.random.default_rng(seed=random_seed)
 
         self.model = modelcontext(model)
-        self.variables = inputvars(self.model.value_vars)
+        self.variables = self.model.value_vars
 
         self.var_info = {}
         self.tempered_posterior = None
--- a/pymc/step_methods/hmc/base_hmc.py
+++ b/pymc/step_methods/hmc/base_hmc.py
@@ -31,6 +31,7 @@ from pymc.step_methods.arraystep import GradientSharedStep
 from pymc.step_methods.hmc import integration
 from pymc.step_methods.hmc.quadpotential import QuadPotentialDiagAdapt, quad_potential
 from pymc.tuning import guess_scaling
+from pymc.util import get_value_vars_from_user_vars
 
 logger = logging.getLogger("pymc")
 
@@ -91,8 +92,7 @@ class BaseHMC(GradientSharedStep):
         if vars is None:
             vars = self._model.continuous_value_vars
         else:
-            vars = [self._model.rvs_to_values.get(var, var) for var in vars]
-
+            vars = get_value_vars_from_user_vars(vars, self._model)
         super().__init__(vars, blocked=blocked, model=self._model, dtype=dtype, **aesara_kwargs)
 
         self.adapt_step_size = adapt_step_size
--- a/pymc/step_methods/metropolis.py
+++ b/pymc/step_methods/metropolis.py
@@ -56,6 +56,8 @@ __all__ = [
     "MultivariateNormalProposal",
 ]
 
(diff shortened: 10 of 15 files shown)


pymc/util.py
```diff
@@ -366,3 +366,12 @@ def wrapped(**kwargs):
...
 
     return wrapped
+
+
+def check_all_value_vars_in_model(vars, model):
+    notin = [v for v in vars if v not in model.value_vars]
+    if notin:
+        notin = list(map(get_var_name, notin))
+        raise ValueError(
+            "Some variables do not correspond to value variables in the model: " + str(notin)
```
dfm: It might be worth explicitly recommending the use of `aesaraf.rvs_to_value_vars` here. I wasn't able to find that function until you told me about it!!

ricardoV94: Although that's the issue that motivated these changes I am not sure that would be the most common case. I think passing variables from the wrong model is still the most likely reason to end up here, and in that case that advice would not be helpful at all. 

Given our sample size of 1, I would leave as is and reassess if more people complain about it

ricardoV94:
Why did the github bot remove the maintenance label?

pymc/util.py
dfm: lol fine.

dfm: Although I think it's always worth having some useful advice in error messages like this, but you're the boss!

ricardoV94: I agree, but the explicit message only makes sense when we are confident that's the most common cause. For instance, the error message we get when trying to iterate over a TensorVariable is more often that not misplaced. I never face it trying to call `sum` or `max`, but just when passing it into a function that expects an iterable.

```python
import aesara.tensor as at
x = at.vector("x")
# Obvious in this case, but not so much when it's in someone else's function that I don't see
for var in x:
  pass
```
```
TypeError: TensorType does not support iteration. Maybe you are using builtins.sum instead of aesara.tensor.math.sum? (Maybe .max?)
```

Back to the original suggestion. Are you convinced this is a common source of error for PyMC users?

dfm: > that's the most common cause

That's the part that I disagree with! I'm not saying that you should replace this error with _only_ that suggestion. I'm just saying that this change is being made because of a specific issue that actually happens (in at least all models used by an entire field of researchers). This issue has a very simple fix that is deeply buried in a non-public interface. Seems worthwhile documenting that somewhere. Maybe this error message isn't the right place, but the current error message will seem very strange to those of us who _do_ pass in parameters that are a part of the model, just not one that you support!

ricardoV94: I feel the error message is pretty clear now: It tells you that you are passing some variable that is not a model value variable and tells you which one it is. From here it should be straightforward to infer that Deterministics (named or not) are not allowed as inputs to `find_MAP`. We can tweak the docs of `find_MAP`, to be more explicit about that but otherwise I think you are asking for a very specialized error message.

BTW, my word is not final or anything. I'll leave this PR open for others to weight in.

ricardoV94: >  This issue has a very simple fix that is deeply buried in a non-public interface

I agree. I just don't think it's specific to `find_MAP` or even `check_all_value_variables_in_model` (but I could be convinced of the latter). I don't have an immediate alternative suggestion of where to put that information either...

dfm: > I feel the error message is pretty clear now

I'd say this might be related to "value variables" meaning something specific to you. For those of us who are new to the new PyMC-specific terminology, I'd say that it's not super clear as written. I think that updating the `find_MAP` docs to specifically call out this new non-value-variable behavior and suggest `rvs_to_value_vars` as a possible solution, could be a great compromise!

ricardoV94: I'll try to update the docs to be more clear. 

Do note that `rvs_to_value_vars` is no longer sufficient, that gives you a graph built in terms of value variables, but you can still only pass input value variables, not any intermediate variable in the graph (such as a Deterministic). That's a breaking change from V3 that I am proposing to introduce in this PR because I think it's misleading to say we are optimizing with respect to such variables, when we are not.

To backtrack a bit, what is expected of a user when they call `find_MAP` (from now on) is that they specify a subset of free_RVs; not Deterministics, Potentials or other intermediate variables (and obviously not variables that don't belong to the model)

Behind the scenes we convert those free_RVs to the respective value variables. We don't expect most users to ever have to worry about their existence (hence my hesitation to bring it up in a general error message). We do allow value variables to be passed directly, because this function may be called internally by other routines that already converted RVs to their values, but we don't expect users to be doing this conversion themselves.

The current error message makes more sense for developers than for users in that context. I'm happy to change it to be something like: `vars must contain only free_RVs or their respective value variables`.

dfm: Oh I see. I think it's a _huge_ mistake to remove the option of passing variables that aren't `free_RVs` - I'm now sad that I opened the initial issue! But I guess I'll just keep maintaining parallel packages that make PyMC usable in my community :D

ricardoV94: @dfm I added a second commit that adds an informative message about the change. I don't see why do we want to allow users to pass something to `vars` that is not what we are optimizing. 

Does it make sense that both of these calls would do exactly the same thing otherwise (as they did up to this PR)?

```python
import pymc as pm

with pm.Model() as m:

  x = pm.Normal("x")
  y = pm.Deterministic("y", pm.math.exp(x))
  llike = pm.Normal("llike", y, observed=100)

  res1 = pm.find_MAP(vars=pm.aesaraf.rvs_to_value_vars([y]))
  res2 = pm.find_MAP(vars=[x])
```

ricardoV94: By the way, we didn't allow to pass those variables so far in V4 (I think it might have been allowed in V3), that's the "bug" you found where it was trying to optimize wrt to nothing. We were assuming users would pass `free_RVs` already.

The only reason I can see for a user to want to pass anything else is convenience + not knowing what model variables he is working with. In that case the solution as I described above is:

```python
def find_MAP_permissive(graphs):
  # Replace RVs by values in whatever `graphs` is
  value_graphs = pm.aesaraf.rvs_to_values(graphs)
  # Find the input value variables in those graphs
  input_value_graphs = pm.inputvars(value_graphs)
  return find_MAP(vars=input_value_graphs)
```

ricardoV94:
This recent Discourse topic touches on the same issue of trying to pass something that's not an RV to a function that expects RVs: https://discourse.pymc.io/t/trying-to-use-pm-binarymetropolis/10645

With the changes in the PR the new error message would be (condensed example):
```python
import pymc as pm
with pm.Model() as logistic_model_pred:
    p = pm.Beta("p", 1, 1)
    label = pm.Data("label", value=[0, 1, 2], mutable=True)
    observed = pm.Bernoulli("occupancy", p, observed=label)
    step = pm.BinaryMetropolis(vars=label)    
```
```
  File "/home/ricardo/miniconda3/envs/pymc/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3397, in run_code
  File "<ipython-input-2-f13ba6f7e166>", line 6, in <cell line: 2>
    step = pm.BinaryMetropolis(vars=label)
  File "/home/ricardo/Documents/Projects/pymc/pymc/step_methods/metropolis.py", line 395, in __init__
    vars = get_value_vars_from_user_vars(vars, model)
  File "/home/ricardo/Documents/Projects/pymc/pymc/util.py", line 403, in get_value_vars_from_user_vars
    raise ValueError(
ValueError: The following variables are not random variables in the model: ['label']
```

ricardoV94:
@junpenglao One of the EulerMuryama tests just failed here. Maybe it needs some seed/ higher tolerance?

pymc/util.py
junpenglao: While I agree with Ricardo that we are not really optimizing the deterministic, I think it is a reasonable "short cut" of expressing "optimized for all the dependent free_RVs of the deterministic".

How about a compromise that we raise a warning instead of an error if a deterministic from the same graph is inputted?

ricardoV94: Limited to `find_MAP` or also for every step, where we ask users which variables they want to assign to it?

`step = pm.Slice(vars=[deterministic])`...

You could also extend this to `model.logp(vars=[deterministic])`, `model.dlogp(vars=[deterministic])`. I think it makes the API more confusing to be more flexible, but a bit arbitrary to allow for it only in `find_MAP`.

junpenglao: Limited to `find_MAP`, as we cannot assign sampler to the RVs underlying deterministic

ricardoV94: So do you suggest I replace the new error I added in `find_MAP`, by a nested try/except, where we also check if the variable corresponds to a deterministic in the model, and then issue the warning and use the underlying RVs? 

What if it's not a registered Deterministic? Okay to fail then?

ricardoV94: > Limited to find_MAP, as we cannot assign sampler to the RVs underlying deterministic

That exactly the same that's happening here in `find_MAP`. Except it's an optimizator and not a sampler. We also cannot optimize wrt to a deterministic.

junpenglao: Sorry of not being more explicit, what i meant is (in `find_MAP`) something like:

```python
try:
    vars = get_value_vars_from_user_vars(vars, model)
except ValueError as exc:
    vars = pm.aesaraf.rvs_to_value_vars([vars])
    if vars is not Null:
          warnings.warn(
          "Intermediate variable (Deterministics or Potentials) was passed, "
          "find_MAP will optimized its dependent Free_RVs instead."
           )
    else:
        raise ValueError(...) from exc
```

I think this is ok to limit the behavior in find_MAP because find_MAP only works for all continues variable, while in sampler this is not guarantee. 

Let me know if I am missing something.

ricardoV94: Okay that sounds reasonable. Will update

ricardoV94:
Ready for review

pymc/util.py
```diff
@@ -441,3 +442,42 @@ def _get_unique_seeds_per_chain(integers_fn):
...
+    Raises
+    ------
+    ValueError:
+        If any of the provided variables do not correspond to any model value variable
+    """
+    if not isinstance(vars, (tuple, list)):
+        # Single var was passed
+        value_vars = [model.rvs_to_values.get(vars, vars)]
+    else:
+        value_vars = [model.rvs_to_values.get(var, var) for var in vars]
```
junpenglao:
```suggestion
    if not isinstance(vars, Iterable):
        # Single var was passed
        vars = [vars]
    value_vars = [model.rvs_to_values.get(var, var) for var in vars]
```

junpenglao (approved):
LGTM. Minor nit but feel free to ignore.

pymc/util.py
ricardoV94: Python thinks Aesara Variables are iterable, but checking for sequence seems to work

michaelosthege (approved):

ricardoV94:
The MacOS job got stuck in conda-cache. Gonna take a risk and merge regardless

PR #6237, opened 2022-10-20
Deprecate old or unused Model methods

### LINKED ISSUE #5859: Remove `t` suffix from `Model` methods
`model.logpt`, `model.dlogpt` and `model.d2logpt` would be a bit more intuitive without the `t` suffix. Historically that was done to distinguish the `tensor` graphs from the compiled functions, but those are now obtained via `model.compile_logp`, `model.compile_dlogp` and `model.compile_d2logp`.

ricardoV94:
Closes #5859 

## Major / Breaking Changes
- Deprecate several methods from `Model` class

## Bugfixes / New features
- ...

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -12,7 +12,6 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-import collections
 import functools
 import threading
 import types
@@ -65,7 +64,6 @@ from pymc.distributions.logprob import _get_scaling
 from pymc.distributions.transforms import _default_transform
 from pymc.exceptions import ImputationWarning, SamplingError, ShapeError, ShapeWarning
 from pymc.initial_point import make_initial_point_fn
-from pymc.math import flatten_list
 from pymc.util import (
     UNSET,
     WithMemoization,
@@ -85,8 +83,6 @@ __all__ = [
     "Point",
     "compile_fn",
 ]
-
-FlatView = collections.namedtuple("FlatView", "input, replacements")
 
 
 class InstanceMethod:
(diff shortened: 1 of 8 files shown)


pymc/model.py
```diff
@@ -983,6 +907,32 @@ def unobserved_value_vars(self):
...
+
+    @property
+    def discrete_value_vars(self):
+        """All the discrete value variables in the model"""
+        return list(typefilter(self.value_vars, discrete_types))
+
+    @property
+    def cont_vars(self):
+        warnings.warn(
+            "Model.continuous_value_vars has been deprecated. Use Model.continuous_value_vars instead.",
```
twiecki:
```suggestion
            "Model.cont_vars has been deprecated. Use Model.continuous_value_vars instead.",
```

pymc/model.py
```diff
@@ -983,6 +907,32 @@ def unobserved_value_vars(self):
 
         return vars + untransformed_vars + deterministics
 
+    @property
+    def disc_vars(self):
```
twiecki, line 911: Has this been here before?

pymc/model.py
ricardoV94: Good catch

pymc/model.py
ricardoV94: It was not here before, but existed already. I tried to regroup methods to follow a more logical order but it looks like an addition in this specific diff :P

michaelosthege (approved):
Flaky tests being flaky again.

LGTM

twiecki (approved):

PR #6241, opened 2022-10-24
Include `observed_RVs` in `eval_rv_shapes`

ricardoV94:
It seems helpful for debugging.

The givens also seems unnecessary now that we take into consideration the shape of `observations` by default. I assume that when we didn't, it was necessary?

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Include `observed_RVs` in `eval_rv_shapes`


ricardoV94:
Changed my mind :D

PR #6257, opened 2022-11-01
Split `sampling.py` into `sampling.py` and `sampling_forward.py`

michaelosthege:
Closes #6141

**What is this PR about?**
Splitting up the `sampling.py` module such that MCMC and prior/posterior predictive sampling are separated.

This also splits one of the longest-running test files.

I did not change any implementation - not even renamed a function.

For review, please check the `__all__` settings, because they define what is the public API.

**Checklist**
+ [X] Explain important implementation details 👆
+ [x] Make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).
+ [x] Link relevant issues (preferably in [nice commit messages](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html))
+ [x] Are the changes covered by tests and docstrings?
+ [x] Fill out the short summary sections 👇

## Docs / Maintenance
- Split `sampling.py` into `sampling.py` and `sampling_forward.py`

PR #6268, opened 2022-11-05
Move sampling code to a submodule

michaelosthege:
**What is this PR about?**
In #6257 we already wanted to do this, but had no good way to keep the `from pymc import sampling_jax` imports working.

Today I found a simple solution for this, so this PR moves all `*sampling*.py` files into a `sampling` submodule.

I also moved the tests accordingly.

⚠ The commits must not be squashed, otherwise the git history of `jax.py` would get messed-up. ⚠

You can check the diff commit-by-commit to see that I didn't actually change +721-680 lines...

**Checklist**

## Major / Breaking Changes
- None

## Bugfixes / New features
- None

## Docs / Maintenance
- All sampling related modules were moved into a new `pymc.sampling` submodule to facilitate further refactoring.


Armavica (commented):
I only have two comments:
* There needs to be an empty `__init__.py` file inside `tests/sampling`
* There are a few places, mainly in the tests, where you import from `pm.sampling.mcmc` or `pm.sampling.forward` instead of just `pm.sampling`. This is not needed for the few items that are mentioned in `__all__` of these two submodules, because they are reimported in `pm.sampling`. But it works this way too, I didn't know if you actually wanted to be more precise, and I don't have a strong opinion on this question.

michaelosthege:
thanks, @Armavica ! I'll add the init file.
For the import routes that I went with the exact ones because it makes refactoring a little easier.
IMO we should try to eventually arrive at a point where each submodule has an `__all__`-advertised public API and only those functions are used by other components. Once we get there I agree that we should import from the init files.

@ricardoV94 want to add something too before I make the "apply review feedback commit"?

ricardoV94:
> @ricardoV94 want to add something too before I make the "apply review feedback commit"?

LGTM

michaelosthege:
> LGTM, do you want a review already or still need to merge the suggestion?

I already added the `__init__.py` file, this is ready to review. You can even merge if you like it ;)

ricardoV94 (approved):

PR #6269, opened 2022-11-05
Simplify mcmc module

michaelosthege:
**What is this PR about?**

I'm refactoring `mcmc.py` (ex `sampling.py`) to prepare for extraction of population-sampling related code from `mcmc.py`.

Next to the simplification and moving of code, I also removed support for selecting which variables to include in the trace.
I have never seen this applied anywhere, but anybody feel free to object.
However, if we decide to not make this breaking change, we should still refactor the internal functions towards a clear distinction of `trace` vs.  `trace_vars`.

## Major / Breaking Changes
- Removed support for selectively tracking variables via `pm.sample(trace=[...])`.

## Bugfixes / New features
- None

## Docs / Maintenance
- Moved internal functions in `mcmc.py` down into other modules.


ricardoV94:
> Next to the simplification and moving of code, I also removed support for selecting which variables to include in the trace.

Is this similar to the "broken functionality" in this issue: https://github.com/pymc-devs/pymc/issues/6116. If so we could also remove it there and close the issue

michaelosthege:
> Is this similar to the "broken functionality" in this issue: https://github.com/pymc-devs/pymc/issues/6116. If so we could also remove it there and close the issue

This might be a matter of incorrect type annotation: In `pm.sample` the `trace` was advertised as `List[str]` (among other types), but the test case actually fed a `List[TensorVariable]`. I don't know if `List[str]` even worked.

But yes, it sounds like a similar feature.

_If_ we want to support this, I would suggest to reintroduce it with a separate kwarg `trace_vars: Optional[Sequence[str]]` (`str` because we'll only want to support named RVs..).

ricardoV94 (approved):

PR #6272, opened 2022-11-07
Bring back logps

ricardoV94:
This PR brings back univariate continuous logp methods that had been moved to Aeppl

~~I also removed all old logp/logcdf docstrings. They were both wrong and too cumbersome to update.~~ Will do in a separate PR

Closes #5329

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Remove old `logp` and `logcdf` docstrings

Issue #4859, opened 2021-07-13
Update docstrings of distributions `logp` and `logcdf` methods.


ricardoV94:
We should update the docstrings of these methods. They are quite verbose, and the same information could be easily conveyed by type hints. More importantly, they are also outdated (e.g., the methods no longer expect only a value as input)

https://github.com/pymc-devs/pymc3/blob/9c7a92481856e1f86c15868b93a65a5d778b0c22/pymc3/distributions/continuous.py#L2198-L2211

zoj613:
Are you suggesting they be removed entirely?

ricardoV94:
> Are you suggesting they be removed entirely?
> 

Maybe... they are not user facing anymore and all work in the same way. Perhaps it's enough to have the documentation at the parent class level?

What do others suggest?

zoj613:
> > Are you suggesting they be removed entirely?
> 
> Maybe... they are not user facing anymore and all work in the same way. Perhaps it's enough to have the documentation at the parent class level?
> 
> What to others suggest?

I do agree. If they don't ever appear in the docs then maybe they can be removed. This will helping scroll easier since the file is getting bigger.

mjhajharia:
maybe we can do away with the doctoring and have something like `rng_fn` for `logp`, I mean this is enough for explaining the input and output types

```
def rng_fn(
        cls,
        rng: np.random.RandomState,
        loc: np.ndarray,
        scale: np.ndarray,
        size: Tuple[int, ...],
    ) -> np.ndarray:
```

OriolAbril:
Updating the labels as it's not clear how to go about this or even if this should be done at all. 

For some more details see https://github.com/pymc-devs/pymc/issues/5282#issuecomment-1035652666 and https://github.com/pymc-devs/pymc/issues/5308

PR #6280, opened 2022-11-09
Remove wrong type-hints and stale docstrings from distributions

ricardoV94:
Related to #4859

This was originally proposed in #6272, see https://github.com/pymc-devs/pymc/pull/6272#pullrequestreview-1172309680

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Remove stale `logp` and `logcdf` docstrings

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -20,7 +20,7 @@ nodes in PyMC.
 
 import warnings
 
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Union
 
 import aesara
 import aesara.tensor as at
@@ -53,7 +53,7 @@ from aesara.tensor.random.basic import (
     vonmises,
 )
 from aesara.tensor.random.op import RandomVariable
-from aesara.tensor.var import TensorConstant, TensorVariable
+from aesara.tensor.var import TensorConstant
 
 try:
     from polyagamma import polyagamma_cdf, polyagamma_pdf, random_polyagamma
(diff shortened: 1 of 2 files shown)


canyon289:
Thank you Ricardo! I'll get to this in next 96 hours

ricardoV94:
> Not blocking the PR, but trying to think forawrd. How do we expect users to know what parameters they need to pass for logp/logcdf to work? several distributions also have alternative parametrizations

Users should never call these methods themselves. The user-facing API is `pm.logp(rv, value)`

ricardoV94:
I would bet a newbie developer would be more confused than enlightened by these broken docstrings...

pymc/distributions/continuous.py
```diff
@@ -1051,31 +925,7 @@ def get_mu_lam_phi(
...
-        Calculate log-probability of Wald distribution at specified value.
-
-        Parameters
-        ----------
-        value : tensor_like of float
-            Value(s) for which log-probability is calculated. If the log probabilities for multiple
-            values are desired the values must be provided in a numpy array or Aesara tensor.
-        mu : tensor_like of float
-            Mean of the distribution (mu > 0).
-        lam : tensor_like of float
```
canyon289, line 1070: This is a really good example. I wouldnt know what lam or alpha are by memory and from the logp calculation below theres not enough math to tell me. Having even just a basic description of the parameters would be helpful mathematically.

The other is the what is allowed as inputs. If I was to write code that called this logp I would know what to pass, and auto type checking would be useful.

In the scope of things I care way more about 1 than I do 2, though both would be nice

ricardoV94: This is a really good example of why I don't like it. The return type is completely wrong, should be TensorVariable (Type) not RandomVariable (Op). Also we only test and promise to support TensorVariable inputs, not float or np.ndarray (even though many times they will work)

ricardoV94: The parameter information is contained in the main distribution docstring already.

ricardoV94: https://github.com/pymc-devs/pymc/blob/d34b557b5e0a0806842e2d4ec5aa5a842dbf8dfe/pymc/distributions/continuous.py#L861-L868

maresb:
Ok, presumably with the pin in #6282, mypy is now passing. Is this ready to merge? (I have not reviewed.)

OriolAbril (approved):

PR #6281, opened 2022-11-09
Do not rely on tag information for rv and logp conversions

### LINKED ISSUE #5033: Remove value variable information from `RV.tag`

### LINKED ISSUE #6236: Turn `aesaraf.rvs_to_value_vars` into a `Model` method
This function only makes sense in the context of a Model. This should also help with moving away from the `tag.value_var` business as described in #5033

ricardoV94:
This PR will finally clean the interface for accessing meta-information from RVs needed when transforming back and forth between the RV and logp representations of a PyMC model

It does not yet remove that information to not break anything. Some methods relied heavily on this local information and were simply marked for deprecation, with alternatives being provided from scratch.

Accessing these variables (as long as they were introduced by PyMC) via tag will issue a FutureWarning:

```python
with pm.Model() as m:
  x = pm.Normal("x")
x.tag.value_var  # Warning: The tag attribute value_var is deprecated. Use model.rvs_to_values[rv] instead
```

Closes #5033
Closes #6236 

## Major / Breaking Changes
- Deprecated accessing `[value_variable|observations|transform|total_size]` via `var.tag` in favor of `model.rvs_to_[values|transforms|total_sizes]`
- Deprecated `joint_logp` in favor of `model.logp`
- Deprecated `aesaraf.rvs_to_value_vars` in favor of `model.replace_rvs_by_values`

## Bugfixes / New features
- Fix bug when replacing random variables with nested value transforms

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -236,6 +236,7 @@ def replace_rvs_in_graphs(
             new_nodes.extend(replacement_fn(var, replacements))
         return new_nodes
 
+    # This iteration populates the replacements
     for var in walk_model(graphs, expand_fn=expand_replace, **kwargs):
         pass
 
@@ -250,7 +251,15 @@ def replace_rvs_in_graphs(
             clone=False,
         )
 
-        fg.replace_all(replacements.items(), import_missing=True)
+        # replacements have to be done in reverse topological order so that nested
+        # expressions get recursively replaced correctly
+        toposort = fg.toposort()
+        sorted_replacements = sorted(
+            tuple(replacements.items()),
+            key=lambda pair: toposort.index(pair[0].owner),
+            reverse=True,
+        )
+        fg.replace_all(sorted_replacements, import_missing=True)
 
         graphs = list(fg.outputs)
 
--- a/pymc/aesaraf.py
+++ b/pymc/aesaraf.py
@@ -11,6 +11,8 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+import warnings
+
 from typing import (
     Callable,
     Dict,
@@ -31,6 +33,7 @@ import pandas as pd
 import scipy.sparse as sps
 
 from aeppl.logprob import CheckParameterValue
+from aeppl.transforms import RVTransform
 from aesara import scalar
 from aesara.compile.mode import Mode, get_mode
 from aesara.gradient import grad
@@ -205,7 +208,7 @@ def walk_model(
     yield from walk(graphs, expand, bfs=False)
 
 
-def replace_rvs_in_graphs(
+def _replace_rvs_in_graphs(
     graphs: Iterable[TensorVariable],
     replacement_fn: Callable[[TensorVariable], Dict[TensorVariable, TensorVariable]],
     **kwargs,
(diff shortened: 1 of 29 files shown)


ricardoV94:
Could we add a property to the old tag variables, so that if they are accessed it raises a FutureWarning?

```python
with pm.Model() as m:
  x = pm.Normal("x")
x.tag.value_var  # Warning: Use model.rvs_to_values[var] instead
```

michaelosthege:
@ricardoV94 the type hint in `aeppl.transforms.TransformValuesRewrite` iss wrong: https://github.com/aesara-devs/aeppl/blob/961496a17e4291b76a5affc140ce85b42a6527cb/aeppl/transforms.py#L193

It should say `Dict[TensorVariable, Optional[Union[RVTransform, DefaultTransformSentinel]]]`

Best place a `# type: ignore` in the line here for now

pymc/distributions/logprob.py
```diff
@@ -281,6 +263,52 @@ def joint_logp(
...
+    rvs: Sequence[TensorVariable],
+    *,
+    rvs_to_values: Dict[TensorVariable, TensorVariable],
+    rvs_to_transforms: Dict[TensorVariable, RVTransform],
+    jacobian: bool = True,
+    rvs_to_total_sizes: Dict[TensorVariable, TOTAL_SIZE],
+    **kwargs,
+) -> List[TensorVariable]:
+    """Thin wrapper around aeppl.factorized_joint_logprob, extended with PyMC specific
+    concerns such as trnasforms, jacobian, and scaling"""
```
michaelosthege:
```suggestion
    concerns such as transforms, jacobian, and scaling"""
```

pymc/aesaraf.py
```diff
@@ -311,15 +320,72 @@ def populate_replacements(
...
+    rvs_to_values: Dict[TensorVariable, TensorVariable],
+    rvs_to_transforms: Dict[TensorVariable, RVTransform],
+    **kwargs,
+) -> List[TensorVariable]:
+    """Clone and replace random variables in graphs with their value variables.
+
+    This will *not* recompute test values in the resulting graphs.
+
+    Parameters
+    ==========
```
michaelosthege:
```suggestion
    ----------
```

pymc/model.py
```diff
@@ -1582,9 +1579,29 @@ def __getitem__(self, key):
...
+        self,
+        graphs: Sequence[TensorVariable],
+        **kwargs,
+    ) -> List[TensorVariable]:
+        """Clone and replace random variables in graphs with their value variables.
+
+        This will *not* recompute test values in the resulting graphs.
+
+        Parameters
+        ==========
```
```suggestion
        ----------
```

michaelosthege (approved):
Overall looking good; great work, @ricardoV94 👍

I found some minor things in docstrings and I'm slightly worried that the big multiline expressions in some tests might result in hard to read tracebacks if they fail but this is also just a minor concern.

ricardoV94:
>  I'm slightly worried that the big multiline expressions in some tests might result in hard to read tracebacks if they fail but this is also just a minor concern.

Then we will make sure they never fail :D 

Yeah those tests a bit messy, but on the bright sight they are due to be removed when we deprecate the warnings.

PR #6283, opened 2022-11-09
Updated pymc.DiscreteWeibull docstring

hyosubkim:
**What is this PR about?**
...Added parameter descriptions to docstring and other minor edits. 

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- ...

Note: I am not fully confident in nomenclature for q parameter. Some resources call it a 'shape' parameter, so I went with that.

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -462,15 +462,16 @@ discrete_weibull = DiscreteWeibullRV()
 
 
 class DiscreteWeibull(Discrete):
-    R"""Discrete Weibull log-likelihood
+    """Discrete Weibull log-likelihood.
 
     The discrete Weibull distribution is a flexible model of count data that
     can handle both over- and under-dispersion.
     The pmf of this distribution is
 
-    .. math:: f(x \mid q, \beta) = q^{x^{\beta}} - q^{(x + 1)^{\beta}}
+    .. math:: f(x \\mid q, \beta) = q^{x^{\beta}} - q^{(x + 1)^{\beta}}
 
     .. plot::
+        :context: close-figs
 
         import matplotlib.pyplot as plt
         import numpy as np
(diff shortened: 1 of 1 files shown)


pymc/distributions/discrete.py
```diff
@@ -462,15 +462,16 @@ def rng_fn(cls, rng, q, beta, size):
 
 
 class DiscreteWeibull(Discrete):
-    R"""Discrete Weibull log-likelihood
+    """Discrete Weibull log-likelihood.
```
OriolAbril: The `R` is needed to tell python to treat the docstring as a "raw" string. Otherwise, things like `\b` or `\n` are considered single special characthers instead of being one backslash and one letter and that breaks the latex formulas.

pymc/distributions/discrete.py
```diff
@@ -462,15 +462,16 @@ def rng_fn(cls, rng, q, beta, size):
...
 
     The discrete Weibull distribution is a flexible model of count data that
     can handle both over- and under-dispersion.
     The pmf of this distribution is
 
-    .. math:: f(x \mid q, \beta) = q^{x^{\beta}} - q^{(x + 1)^{\beta}}
+    .. math:: f(x \\mid q, \beta) = q^{x^{\beta}} - q^{(x + 1)^{\beta}}
```
OriolAbril: I think this extra `\` will also break latex rendering. The formulas already look right in `latest`: https://www.pymc.io/projects/docs/en/latest/api/distributions/generated/pymc.DiscreteWeibull.html so they should not be modified.

OriolAbril (commented):
It needs a couple changes (see comments below) but otherwise looks good doc-wise.

Can someone familiar with the distribution review the parameter descriptions? Having 2 shape parameters sounds a bit strange but I don't really know what each of them means.

pymc/distributions/discrete.py
OriolAbril: same for the other extra `\` characters in math formulas

pymc/distributions/discrete.py
hyosubkim: Got it.

pymc/distributions/discrete.py
hyosubkim: Strange, as the extra \ doesn't appear in my local copy...

OriolAbril: maybe your ide added that automatically when you removed the R? I think double backlashes are escaped in regular strings, but the previed on readthedocs ended up having latex errors

OriolAbril (approved):

pymc/distributions/discrete.py
hyosubkim: Yeah, it was the missing R. Thanks

michaelosthege (approved):

PR #6287, opened 2022-11-11
Fix Flaky Euler-Maruyama Tests

williambdean:
**What is this PR about?**
This PR addresses the flaky tests with the EulerMaruyama class which were brought up in [here](https://github.com/pymc-devs/pymc/issues/6276) and [here](https://github.com/pymc-devs/pymc/issues/6247). 

This PR includes seeding the current tests and ensuring that the V4 port didn't change the functionality.

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- Seeding the Euler-Maruyama linear_model test
- Adding a space ` : ` in the Parameters for betting render


wd60622:
So I took at look at the differences @ricardoV94... 
The linear_model test seems to be a direct port from the last tagged version of V3. 
[Old test](https://github.com/pymc-devs/pymc/blob/ed74406735b2faf721e7ebfa156cc6828a5ae16e/pymc3/tests/test_distributions_timeseries.py#L145-L171) -> [Latest V4 test](https://github.com/pymc-devs/pymc/blob/d4b5aa86649674bcf6d17344b58eb1aeaf6cb9a0/pymc/tests/distributions/test_timeseries.py#L916-L950)

Also, the implementation of the logp seems to be the same besides the addition of the init_dist's logp. I view this as a bug fix but I haven't dug up that discussion. Not sure if you were involved there.
[Old logp](https://github.com/pymc-devs/pymc/blob/ed74406735b2faf721e7ebfa156cc6828a5ae16e/pymc3/distributions/timeseries.py#L403) -> [Latest V4 logp](https://github.com/pymc-devs/pymc/blob/d4b5aa86649674bcf6d17344b58eb1aeaf6cb9a0/pymc/distributions/timeseries.py#L1059)

Other than that, it seems the port was as intended. I few the random seeding as blocking against the small chance the parameter used to generate is outside of the posterior credible interval

wd60622:
I am also looking into the failure as mentioned in the issue. WIP

wd60622:
Also seeing the previous tests uses `advi+adapt_diag` instead of NUTS

wd60622:
And now another test is failing: `pymc/tests/distributions/test_timeseries.py::TestEulerMaruyama::test_batched_size[False-2]` lol

pymc/tests/distributions/test_timeseries.py
```diff
@@ -853,9 +856,9 @@ def sde_fn(x, k, d, s):
                 "y", dt=0.02, sde_fn=sde_fn, sde_pars=sde_pars, init_dist=init_dist, **kwargs
             )
 
-        y_eval = draw(y, draws=2)
+        y_eval = draw(y, draws=2, random_seed=RANDOM_SEED)
         assert y_eval[0].shape == (batch_size, steps)
-        assert not np.any(np.isclose(y_eval[0], y_eval[1]))
+        assert np.any(~np.isclose(y_eval[0], y_eval[1]))
```
williambdean:
```python 
np.any(~np.isclose([1, 2, 3, 4, 5], [1, 2, 3, 4, 5]))   # -> False since all the same
np.any(~np.isclose([5, 4, 3, 2, 1], [1, 2, 3, 4, 5]))   # -> True since at least one different 
np.any(~np.isclose([1, 2, 3, 4, 5], [6, 7, 8, 9, 10]))  # -> True since all different

**Review by michaelosthege — APPROVED** [review]
Looking good!

One less flaky boi :)

**Comment by wd60622** [comment]
I have to quickly rebase with new main since the `seed` args in test were just changed to `random_seed` (by me)

File: pymc/tests/distributions/test_timeseries.py (hunk 1 above)
Comment (ricardoV94) [reply pymc/tests/distributions/test_timeseries.py]: Do you reckon this check is better, than checking all are different?

File: pymc/tests/distributions/test_timeseries.py [hunk 2]
```diff
@@ -853,9 +856,9 @@ def sde_fn(x, k, d, s):
                 "y", dt=0.02, sde_fn=sde_fn, sde_pars=sde_pars, init_dist=init_dist, **kwargs
             )
 
-        y_eval = draw(y, draws=2)
+        y_eval = draw(y, draws=2, random_seed=RANDOM_SEED)
```
ricardoV94: Suggestion: you can pass the `numpy_rng` as `random_seed`

pymc/tests/distributions/test_timeseries.py
williambdean: Surprisingly, I had it fail on my end where checking that all are different (previous test). Not sure the chances of that happening, but it did. That is, the two arrays were the same at at least 1 spot. In the case of an intersection between the two ys, I saw this better capturing that they are different arrays

wd60622:
Failed test on test_gp.py

michaelosthege:
> Failed test on test_gp.py

That's a known flaky test #6299.

we can just override

pymc/tests/distributions/test_timeseries.py
ricardoV94: The test should now be deterministic no?

williambdean: Yes. There are two changes: 
1. Deterministic because of the random seed
2. Change in the assert which, I thought, better checks that the two draws are different based on the logic described. If you disagree, I will change back

pymc/tests/distributions/test_timeseries.py
williambdean: done

pymc/tests/distributions/test_timeseries.py
williambdean: It still passes under the seed and the previous assert though but so can any test catered to do so

ricardoV94: I don't have a strong preference, but all values differently is certainly stronger evidence that it's being random :D

williambdean: Fair. 😃

ricardoV94 (approved):

PR #6291, opened 2022-11-11
Remove duplicate methods

williambdean:
**What is this PR about?**
Noticed some duplicated methods in the `pymc/model.py` file mentioned in [issue here](https://github.com/pymc-devs/pymc/issues/6290).
Removed the duplication and tried to improve the docstrings

**Checklist**

## Major / Breaking Changes
- Using keyword `seed` for initial point no longer supported

## Bugfixes / New features
- NA

## Docs / Maintenance
- More consistent seeding argument with `random_seed`
- Removed duplication function and method names throughout the package

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -989,9 +989,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
     @property
     def test_point(self) -> Dict[str, np.ndarray]:
-        """Deprecated alias for `Model.initial_point(seed=None)`."""
+        """Deprecated alias for `Model.initial_point(random_seed=None)`."""
         warnings.warn(
-            "`Model.test_point` has been deprecated. Use `Model.initial_point(seed=None)`.",
+            "`Model.test_point` has been deprecated. Use `Model.initial_point(random_seed=None)`.",
             FutureWarning,
         )
         return self.initial_point()
@@ -1002,7 +1002,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         Parameters
         ----------
         random_seed : SeedSequenceSeed, default None
-            Seed(s) for generating initial point from the model. Used in pymc.aesaraf.reseed_rng
+            Seed(s) for generating initial point from the model. Used in pymc.aesaraf.reseed_rngs
 
         Returns
         -------
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1101,15 +1101,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
         self.dim_lengths[name].set_value(new_length)
         return
 
-    @property
-    def test_point(self) -> Dict[str, np.ndarray]:
-        """Deprecated alias for `Model.initial_point(random_seed=None)`."""
-        warnings.warn(
-            "`Model.test_point` has been deprecated. Use `Model.initial_point(random_seed=None)`.",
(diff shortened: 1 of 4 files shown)


pymc/model.py
```diff
@@ -988,23 +989,28 @@ def dim_lengths(self) -> Dict[str, Variable]:
...
         return self.initial_point()
 
-    def initial_point(self, seed=None) -> Dict[str, np.ndarray]:
+    def initial_point(self, random_seed: SeedSequenceSeed = None) -> Dict[str, np.ndarray]:
         """Computes the initial point of the model.
 
+        Parameters
+        ----------
+        random_seed : SeedSequenceSeed, default None
+            Seed(s) for generating initial point from the model. Used in pymc.aesaraf.reseed_rngs
```
williambdean: Not sure if this is the correct indication of this type in the docstring. Kind of long created type. i.e. 
https://github.com/pymc-devs/pymc/blob/5d7283ea581266e05ca88ce2123d71bcba5f7c8e/pymc/aesaraf.py#L1021

Also mentioned the function that this value is fed into. Not sure how that renders atm.

pymc/model.py
```diff
@@ -1259,34 +1265,6 @@ def set_data(
 
         shared_object.set_value(values)
 
-    def initial_point(self, seed=None) -> Dict[str, np.ndarray]:
```
ricardoV94, line 1262: I think I probably caused this when trying to reorder the methods a bit more logically. There is a `shape_from_dims` below the initial point related methods which makes more sense above, close to the other coord/dims related methods.

I would therefore delete the repeated methods above and not these below.

williambdean: Alright. I will reorder

williambdean: I put after the dim/coords methods but before set_data. Thought that made enough sense

pymc/model.py
```diff
@@ -1137,6 +1101,48 @@ def set_dim(self, name: str, new_length: int, coord_values: Optional[Sequence] =
         self.dim_lengths[name].set_value(new_length)
         return
 
+    @property
+    def test_point(self) -> Dict[str, np.ndarray]:
```
ricardoV94: We can get rid of this deprecated property as well

williambdean: Done

pymc/model.py
michaelosthege: Looks about right (from [here](https://pymcio--6291.org.readthedocs.build/projects/docs/en/6291/api/generated/classmethods/pymc.Model.initial_point.html#pymc-model-initial-point))

grafik

michaelosthege (approved):
Looks like this is ready to merge, right @ricardoV94 ?

ricardoV94 (approved):

bmorris3:
Was it intentional/necessary to change the argument `seed` to `random_seed`? This leads to lots of breaks in things downstream.

michaelosthege:
it was an intentional change, but next time we should go for a backwards compatible approach with a deprecation warning..

wd60622:
> Was it intentional/necessary to change the argument `seed` to `random_seed`? This leads to lots of breaks in things downstream.

Can you be specific with "lots" of breaking changes? 

Is there something that isn't covered in the tests? Or was your solutions just a grep replace?

PR #6273, opened 2022-11-07
Use default Aesara RandomStream for VI

ricardoV94:
Closes #4523 

## Major / Breaking Changes
- ...

## Bugfixes / New features
- ...

## Docs / Maintenance
- ...

PR #6304, opened 2022-11-15
Refactor MiniBatch and stop using deprecated MRG sampler

### LINKED ISSUE #4523: VI using deprecated Aesara MRG sampler
The current usage of the mrng sampler in ADVI does not support JAX (https://github.com/pymc-devs/aesara/issues/322). It should be fairly easy to instead make it use the new `RandomVariable` Op https://github.com/pymc-devs/aesara/pull/296 to get JAX support for ADVI.

CC @ferrine

### LINKED ISSUE #6277: Use RandomVariables for Minibatch
Minibatch creates a stochastic view of the underlying dataset using some pretty funky Aesara magic. 

In a straightforward implementation it would do something like:
```python
import pymc as pm
import numpy as np

data = pm.Normal.dist(size=(100, 2)).eval()

with pm.Model() as m:
  data = pm.Data("data", data)

  mb_slice_start, mb_slice_end = pm.Uniform.dist(0, data.shape[0], size=(2)).sort()
  minibatch_data = data[mb_slice_start: mb_slice_end]
  
  x = pm.Normal("x", 0, 1, observed=minibatch_data)
```
And then PyMC also figures out that `x` lopg has to be scaled by the size of the unobserved values in each iteration.

One pressing issues is that it relies on the deprecated MRG sampler #4523, which also has no support for JAX or Numba. It isn't as simple a matter as switching to the default `RandomStream` which returns `RandomVariable`s, because we treat the shared RNG variables used by RandomVariables in a special manner inside `aesaraf.compile_pymc` (the function through which all PyMC functions go through before calling `aesara.function`). Specifically we always set the values of the distinct shared RNG variables to something new, based on what `random_seed` is passed to `compile_pymc`.

This is a problem when you need to synchronize multiple mini-batches of data, say when you have a linear regression model with a minibatch of `x` and `y`. The way it currently works is that you set the same `random_seed` to each (default is hard-coded 42), and then rely on this initial value being the same so that the endpoints of the slice stay synchronized. But if we were to use the standard RandomStream, and pass the graph to `compile_pymc`, the two RNGs associated with the two slices would be overwritten to different values, regardless of the fact that they started with the same value.

Anyway, this trick might also not have been enough, since there was a need to introduce `align_minibatches` in https://github.com/pymc-devs/pymc/pull/2760: 
https://github.com/pymc-devs/pymc/blob/faebc60ff2cb7c9befff5fc1122894ffa5600139/pymc/data.py#L463-L474

https://github.com/pymc-devs/pymc/blob/2296350959e4035b4f1ee13fab88014d4f0fa545/pymc/tests/test_data.py#L754-L773

The big issue here is that we are not representing that the slice endpoints are the same symbolically. The current graph looks something like:

```python
with pm.Model() as m:
  x = pm.Data("x", x)
  y = pm.Data("y", y)

  rng = aesara.shared(np.random.default_rng(42))
  mb_slice_start, mb_slice_end = pm.Uniform.dist(0, data.shape[0], size=(2), rng=rng).sort()
  minibatch_x = x[mb_slice_start: mb_slice_end]

  rng = aesara.shared(np.random.default_rng(42))
  mb_slice_start, mb_slice_end = pm.Uniform.dist(0, data.shape[0], size=(2), rng=rng).sort()
  minibatch_y = y[mb_slice_start: mb_slice_end]
  
  obs = pm.Normal("obs", x, 1, observed=minibatch_y)
```
The two rngs were set to the same initial value, but other than that they are completely different as far as Aesara cares. The correct graph would be:

```python
with pm.Model() as m:
  x = pm.Data("x", x)
  y = pm.Data("y", y)
  
  rng = aesara.shared(np.random.default_rng(42))
  mb_slice_start, mb_slice_end = pm.Uniform.dist(0, data.shape[0], shape=(2,), rng=rng).sort()
  minibatch_x = x[mb_slice_start: mb_slice_end]
  minibatch_y = y[mb_slice_start: mb_slice_end]
  
  obs = pm.Normal("obs", x, 1, observed=minibatch_y)
```

That is, using the same slicing RV for both mini-batches. Regardless of the seed value that is set by `compile_pymc`, the two minibatches will always be compatible.

## The incremental proposal

Refactor Minibatch so that it can accept multiple variables at once, which will share the same random slices. Then start using RandomVariables instead of MRG stuff, and get rid of seed at the `minibatch` level

```python
with pm.Model() as m:
  x, y = pm.Minibatch({x: (10, ...), y: (10)})
```
Maybe even sprinkle some dims if you want

## The radical proposal
Offer the same API, but implement the minibatch view with the more straightforward code from the pseudo-examples in this issue. 

No need to keep a global list of all the RNGs:

https://github.com/pymc-devs/pymc/blob/faebc60ff2cb7c9befff5fc1122894ffa5600139/pymc/data.py#L304

The whole Op view magic shouldn't be needed AFAICT. The only thing is that right now PyMC complains if you pass something to observed other than a shared variable or constant data. The concern here is that the observations should not depend on other model RVs, but that's easy to check with `aesara.graph.ancestors()`

https://github.com/pymc-devs/pymc/blob/faebc60ff2cb7c9befff5fc1122894ffa5600139/pymc/data.py#L339

The other issue is that we don't allow orphan RVs in the logp graph, but we can easily create a subtype like `MinibatchUniformRV`, that is allowed, just like `SimulatorRV`s are allowed.

This should make the Minibatch code much more readable and maintainable, as well as compatible with the non-C backends (I have no idea if either supports `View`)

ferrine:
fix #4523, suspend #6273
closes #6277 

@ricardoV94, I've figured out the root cause of the seeding bug

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -12,7 +12,6 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-import collections
 import io
 import os
 import pkgutil
@@ -20,14 +19,19 @@ import urllib.request
 import warnings
 
 from copy import copy
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
+from typing import Dict, Optional, Sequence, Tuple, Union, cast
 
 import numpy as np
 import pytensor
 import pytensor.tensor as at
 
 from pytensor.compile.sharedvalue import SharedVariable
-from pytensor.graph.basic import Apply
+from pytensor.raise_op import Assert
+from pytensor.scalar import Cast
+from pytensor.tensor.elemwise import Elemwise
+from pytensor.tensor.random import RandomStream
+from pytensor.tensor.random.basic import IntegersRV
+from pytensor.tensor.subtensor import AdvancedSubtensor
 from pytensor.tensor.type import TensorType
 from pytensor.tensor.var import TensorConstant, TensorVariable
 
@@ -39,7 +43,6 @@ __all__ = [
     "get_data",
     "GeneratorAdapter",
     "Minibatch",
-    "align_minibatches",
     "Data",
     "ConstantData",
     "MutableData",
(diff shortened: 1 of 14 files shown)


ricardoV94:
@ferrine this does not fix #4523, VI is still using the MRG RandomStream. This can't be switched yet because of #6277

ricardoV94:
It does fix some lack of seeding, for which I think we don't have an issue open

pymc/variational/opvi.py
```diff
@@ -370,10 +370,11 @@ def step_function(
             more_replacements=more_replacements,
             total_grad_norm_constraint=total_grad_norm_constraint,
         )
+        seed = self.approx.rng.integers(2**30, dtype=np.int64)
         if score:
-            step_fn = compile_pymc([], updates.loss, updates=updates, **fn_kwargs)
+            step_fn = compile_pymc([], updates.loss, updates=updates, random_seed=seed, **fn_kwargs)
```
ricardoV94, line 387: Specifically `random_seed` in `compile_pymc` won't do anything with `MRG` `RNG` state variables. We don't actually want to do that until we sort out #6277, otherwise two distinct minibatch RNGs would always be given different initial seed values.

ferrine: hmm

ferrine: Once minibatching is fixed, it should become valid, right?

ferrine: Does it make sense to fix in one PR and get rid of MRG?

ricardoV94: Yes it does!

pymc/data.py
```diff
@@ -12,34 +12,31 @@
...
 
 from aesara.compile.sharedvalue import SharedVariable
-from aesara.graph.basic import Apply
 from aesara.tensor.type import TensorType
 from aesara.tensor.var import TensorConstant, TensorVariable
 
 import pymc as pm
 
-from pymc.aesaraf import convert_observed_data
+from pymc.aesaraf import at_rng, convert_observed_data
```
ricardoV94: We can also get rid of `at_rng`, as the seeds created by it will be overwritten in `compile_pymc` anyway. Just import `RandomStream` wherever it's needed.

pymc/model.py
```diff
@@ -1303,14 +1303,15 @@ def register_rv(
...
-                and not isinstance(data, (GenTensorVariable, Minibatch))
+                and not isinstance(data, (GenTensorVariable))
                 and data.owner is not None
                 # The only Aesara operation we allow on observed data is type casting
                 # Although we could allow for any graph that does not depend on other RVs
                 and not (
                     isinstance(data.owner.op, Elemwise)
                     and isinstance(data.owner.op.scalar_op, Cast)
                 )
+                and not data.name.startswith("minibatch")
```
ricardoV94: You're not using a custom TensorType anymore? If so you can check against that.

ferrine: no custom tensor type(

pymc/data.py
ferrine: ok

pymc/model.py
ricardoV94: I think `_joint_logp` will complain when it sees the RVs in the logp graph? Have you not seen that?

ricardoV94: If you see that, the easiest workaround would be to subclass the Uniform RV used in the Minibatch Slice and allow that type here:

https://github.com/pymc-devs/pymc/blob/5da26173428cacc6aa5e1d05e4b51ceb69dadfd9/pymc/distributions/logprob.py#L107-L126

`and not isinstance(node.owner.op, (SimulatorRV, MinibatchSliceUniformRV))`

ferrine: I've added tests that only pass after I implement minibatching with a new RV

pymc/tests/test_data.py
```diff
@@ -695,16 +695,13 @@ def test_common_errors(self):
...
 
-    def test_mixed2(self):
         with pm.Model():
-            data = np.random.rand(10, 20, 30, 40, 50)
-            mb = pm.Minibatch(data, [2, None, 20])
-            pm.Normal("n", observed=mb, total_size=(10, None, 30))
+            data = np.random.rand(10, 20)
+            mb = pm.Minibatch(data, batch_size=5)
+            v = pm.Normal("n", observed=mb, total_size=10)
+            assert joint_logp(v) is not None, "Check index is allowed in graph"
```
ricardoV94: `joint_logp` is deprecated. You can test that it does not fail by just calling `m.logp()`

ricardoV94:
I think I saw that kernel crash in my PR, perhaps an out of bounds indexing?

ferrine:
no idea why this windows test fails

ferrine:
hmm, that could be the random stream!

ferrine:
The issue that blocks the PR seems to be in aesara or I'm doing something wrong https://github.com/aesara-devs/aesara/issues/1301

ferrine:
Maybe @brandonwillard or @ricardoV94 have any insights?

michaelosthege:
@ferrine are some of these changes independent and could be merged via a separate PR already?

ricardoV94:
Not really... The change led to some scan issues that need to be addressed otherwise it would lead to a regression

ferrine:
Marked this as draft util we figure out the scan issue

pymc/tests/variational/test_inference.py
```diff
@@ -149,14 +149,13 @@ def fit_kwargs(inference, use_minibatch):
         inference.approx.scale_cost_to_minibatch = False
     else:
         key = "full"
+    if (type(inference), key) in {(SVGD, "mini"), (ASVGD, "mini")}:
+        pytest.skip("Not Implemented Inderence")
```
ricardoV94: Should it be an `xfail`?
```suggestion
        pytest.skip("Not Implemented Inference")
```

pymc/util.py
```diff
@@ -28,6 +28,12 @@
 from cachetools import LRUCache, cachedmethod
 
 
+def as_list(obj):
+    if not isinstance(obj, Sequence):
+        obj = [obj]
+    return list(obj)
+
+
```
ricardoV94: This already exists here: https://github.com/pymc-devs/pymc/blob/12f4cd73bae95c42a28c2f07f6aab60071fd2ca1/pymc/aesaraf.py#L546-L550

Maybe move it to `util.py`?

pymc/variational/approximations.py
```diff
@@ -248,9 +249,7 @@ def randidx(self, size=None):
                 pass
         else:
             size = tuple(np.atleast_1d(size))
-        return self._rng.uniform(
-            size=size, low=pm.floatX(0), high=pm.floatX(self.histogram.shape[0]) - pm.floatX(1e-16)
-        ).astype("int32")
+        return self.srng.integers(size=size, low=0, high=self.histogram.shape[0])
```
ricardoV94: Better to use `at.random.integers` directly. Automatic updates are handled by `compile_pymc` anyway, which is the only thing that RandomStream really does. Not relying on it avoids issues when cloning graphs (the rng.update graph still refers to the original one) because that's invisible to tools like `clone_replace`: https://github.com/Theano/Theano/issues/1467

pymc/distributions/logprob.py
```diff
@@ -126,6 +127,22 @@ def _check_no_rvs(logp_terms: Sequence[TensorVariable]):
...
+    if not isinstance(terms, Sequence):
+        terms = [terms]
+    assert all(issubclass(cls, RandomVariable) for cls in ops)
+    return [
+        node
+        for node in aesara.graph.ancestors(terms)
+        if (node.owner and isinstance(node.owner.op, ops))
+    ]
+
+
```
ricardoV94: This does not seem necessary anymore? Even if so, it's place wouldn't be in this module.

pymc/variational/opvi.py
```diff
@@ -943,9 +960,9 @@ def _new_initial(self, size, deterministic, more_replacements=None):
             if deterministic:
                 return at.ones(shape, dtype) * dist_map
             else:
-                return getattr(self._rng, dist_name)(size=shape)
+                return getattr(self.srng, dist_name)(size=shape)
```
ricardoV94: Same applies here, if you can just do `getattr(at.random, dist_name)`, you don't have the issues with the eager `default_update`

pymc/variational/opvi.py
```diff
@@ -943,9 +960,9 @@ def _new_initial(self, size, deterministic, more_replacements=None):
...
         else:
-            sample = getattr(self._rng, dist_name)(size=shape)
+            sample = getattr(self.srng, dist_name)(size=shape)
```
ricardoV94: Same here

pymc/aesaraf.py
```diff
@@ -49,10 +49,10 @@
 )
 from aesara.graph.fg import FunctionGraph
 from aesara.graph.op import Op
-from aesara.sandbox.rng_mrg import MRG_RandomStream as RandomStream
 from aesara.scalar.basic import Cast
 from aesara.tensor.basic import _as_tensor_variable
 from aesara.tensor.elemwise import Elemwise
+from aesara.tensor.random import RandomStream
```
ricardoV94: Can we remove the whole `at_rng` and `set_rng` related utilities below from this file?

https://github.com/pymc-devs/pymc/blob/12f4cd73bae95c42a28c2f07f6aab60071fd2ca1/pymc/aesaraf.py#L872-L912

ricardoV94 (dismissed):

ricardoV94:
@ferrine What do you think of doing the Minibatch API breaking change for V5, and fixing the MRG later, as that is not user facing anyway.

ferrine:
> @ferrine What do you think of doing the Minibatch API breaking change for V5, and fixing the MRG later, as that is not user facing anyway.

That makes to me, we better get back to it later, than get stuck right now. It is also reasonable to improve minibatch API and VI experience later and reduce to something working now

pymc/variational/approximations.py
ricardoV94: If you need to figure out the updates (e.g, inside a Scan) you can use this utility:

https://github.com/pymc-devs/pymc/blob/4cac135cd21c24ca085cfce4f27416d061c23a5e/pymc/pytensorf.py#L1063

ricardoV94:
Codecov complains that quite some tests are not covered now? https://github.com/pymc-devs/pymc/pull/6304/files

pymc/tests/variational/test_inference.py
ferrine: This is a part of a bigger issue, all references to NotImplementedInference should be xfail

ferrine: Basically, NotImplementedInference is something that was broken because of v3-v4 update and we know about it

pymc/data.py
```diff
@@ -126,351 +126,79 @@ def __hash__(self):
...
 
-    To be more concrete about how we create a minibatch, here is a demo:
-    1. create a shared variable
+minibatch_index = MinibatchIndexRV()
 
-        >>> shared = pytensor.shared(data)
 
-    2. take a random slice of size 10:
+def is_minibatch(v: TensorVariable) -> bool:
+    from pytensor.tensor.subtensor import AdvancedSubtensor
```
michaelosthege: This can be placed at the top

pymc/data.py
```diff
@@ -126,351 +126,79 @@ def __hash__(self):
...
+        and valid_for_minibatch(v.owner.inputs[0])
+    )
 
-    >>> moredata = np.random.rand(10, 20, 30, 40, 50)
 
-    The default ``total_size`` that can be passed to PyMC random node
-    is then ``(10, 20, 30, 40, 50)`` but can be less verbose in some cases
+def valid_for_minibatch(v: TensorVariable) -> bool:
+    from pytensor.scalar import Cast
+    from pytensor.tensor.elemwise import Elemwise
```
michaelosthege: Same here

pymc/tests/test_data.py
```diff
@@ -719,51 +714,30 @@ def test_free_rv(self):
...
-            ([(10, 42), None, (4, 42)], (10, 10, 4, 10, 50)),
-            ([(10, 42), Ellipsis, (4, 42)], (10, 10, 40, 10, 4)),
-            ([(10, 42), None, Ellipsis, (4, 42)], (10, 10, 40, 10, 4)),
-            ([10, None, Ellipsis, (4, 42)], (10, 10, 40, 10, 4)),
-        ],
-    )
-    def test_special_batch_size(self, batch_size, expected):
-        mb = pm.Minibatch(self.data, batch_size)
-        assert mb.eval().shape == expected
+        from pymc.data import is_minibatch
```
michaelosthege: Can be at the top

pymc/tests/test_data.py
```diff
@@ -719,51 +714,30 @@ def test_free_rv(self):
...
-        assert a != b
-        pm.align_minibatches([m, n])
-        a, b = zip(*(f() for _ in range(1000)))
-        assert a == b
+        mb = pm.Minibatch(self.data, batch_size=20)
+        assert is_minibatch(mb)
+        assert mb.eval().shape == (20, 10)
+
+    def test_allowed(self):
+        from pymc.data import is_minibatch
```
michaelosthege: Can be at the top

pymc/tests/test_data.py
```diff
@@ -719,51 +714,30 @@ def test_free_rv(self):
...
+        mb = pm.Minibatch(at.as_tensor(self.data).astype(int), batch_size=20)
+        assert is_minibatch(mb)
+
+    def test_not_allowed(self):
+        with pytest.raises(ValueError):
+            mb = pm.Minibatch(at.as_tensor(self.data) * 2, batch_size=20)
+
+    def test_not_allowed2(self):
+        with pytest.raises(ValueError):
+            mb = pm.Minibatch(self.data, at.as_tensor(self.data) * 2, batch_size=20)
```
michaelosthege:
```suggestion
    def test_not_allowed(self):
        with pytest.raises(ValueError, match="not valid for Minibatch"):
            mb = pm.Minibatch(at.as_tensor(self.data) * 2, batch_size=20)

    def test_not_allowed2(self):
        with pytest.raises(ValueError, match="not valid for Minibatch"):
            mb = pm.Minibatch(self.data, at.as_tensor(self.data) * 2, batch_size=20)
```

pymc/tests/test_data.py
```diff
@@ -719,51 +714,30 @@ def test_free_rv(self):
...
+
+    def test_assert(self):
+        with pytest.raises(AssertionError):
```
michaelosthege: Which `AssertionError` is that? (Please match a part of the message..)

michaelosthege:
I marked the VI `test_profile` tests as `XFAIL` and fixed flakyness of the Moyal logcdf test which is also currently ❌ on main.

pymc/aesaraf.py
michaelosthege: `set_at_rng` is called by `pm.sampling.parallel._Process._start_loop`.

In the interest of the v5 release I think we should take care of that later

michaelosthege:
@ferrine I squashed the commits so we can rebase-merge this.

It looks like you have not set up your `@pymc-devs.org` email address in your GitHub account. That's why your pic doesn't shown on the commits..

pymc/tests/distributions/test_continuous.py
```diff
@@ -778,10 +778,13 @@ def test_moyal_logp(self):
         reason="PyMC underflows earlier than scipy on float32",
     )
     def test_moyal_logcdf(self):
+        # SciPy as precision issues at {mu=-2.1, sigma=0.5, x=2.1}
```
ricardoV94:
```suggestion
        # SciPy has new (?) precision issues at {mu=-2.1, sigma=0.5, x=2.1}
```

michaelosthege: it might also be that PyTensor now has smarter optimizations and therefore no longer runs into the same precision issue?

Our logcdf was [deterministic at -2227.95939](https://github.com/pymc-devs/pymc/actions/runs/3665014447/jobs/6195811961#step:7:780):

```
FAILED pymc/tests/distributions/test_continuous.py::TestMatchesScipy::test_moyal_logcdf - AssertionError: 
Arrays are not almost equal to 6 decimals
{'mu': array(2.1), 'sigma': array(0.5), 'value': array(-2.1)}
x and y -inf location mismatch:
 x: array(-2227.95939)
 y: array(-inf)
```

This value is plausible; at `sigma=0.9` it's around -50.

michaelosthege (approved):
🥳🚀

ricardoV94:
@ferrine can you have one last look and merge it if you agree with the changes?

pymc/aesaraf.py
ricardoV94: We can, but it's only used by VI

pymc/tests/distributions/test_continuous.py
ricardoV94: I don't think we did any stabilization improvements recently. Regardless, part of my suggestion was to also fix the typo in the original comment (as instead of has)

ferrine:
I had one more look at the PR changes and I don't have anything to add to it. I'll merge once tests pass.

michaelosthege:
Just fixuped the edit of the comment into the moyal logcdf commit.

Didn't want to waste CI, so I merged it right away =)

PR #6306, opened 2022-11-15
Create value variable on transformed space

aseyboldt:
I'm not sure if this really is a PR as of yet, or possibly more a discussion:
In the source I can't really find a distinction between a value variable for on the untransformed and on the transformed spaces. Instead, it seems that value variables are reused on both depending on context.
I think this leads to problems however: I don't think there is a reason to expect those to even have the same type. If the shape is statically known, then their (static) shapes would often be different for instance. Or they could even have different ranks.
I ran into this with a development version of aesara, where ZeroSumNormal stopped working, because shape inference improved. The new shape inference was smart enough to figure out that the shape the rv was `(10,)`, but the code then would assume that the transformed value variable would have the same shape, which is incorrect (it should have shape `(9,)`.

The current PR contains some code that fixes this specific issue by creating the value variable with the type of the tranformed variable, but I think this is also not really correct. It seems to me we should have separate value variables for transformed and untransformed spaces.

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1460,14 +1460,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         """
         if value_var is None:
-            value_var = rv_var.type()
-            value_var.name = rv_var.name
-
-        if aesara.config.compute_test_value != "off":
-            value_var.tag.test_value = rv_var.tag.test_value
-
-        _add_future_warning_tag(value_var)
-        rv_var.tag.value_var = value_var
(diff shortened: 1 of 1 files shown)


ricardoV94:
Wanna update the Aesara pin to the latest release to see if anything else broke?

ricardoV94:
About your question, we only definite value variables on the transformed space and don't assume transforms can be changed after the distribution is defined (not my preference)

ricardoV94:
Superseded by #6336

PR #6309, opened 2022-11-17
Model API cleanup

### LINKED ISSUE #5076: Disable sampling of transformed variables from `prior_predictive`

### LINKED ISSUE #6305: Rename `RV_dims` and `initial_values` model mapping variables
For consistency with the changes introduced in #6281, I would rename:
* `model.initial_values` to `rvs_to_initial_values`
* `_RV_dims` to `rvs_to_dims` (and loose the underscore prefix, since we don't have any model function to set those?)

Having clear consistent mapping names is very helpful for model transformations such as in https://github.com/pymc-devs/pymcx/pull/91

Eventually we might add a helper `model.replace_rv` that takes care of these mappings, but this issue is just about normalizing names :)

ricardoV94:
Trying to standardize mapping names a bit more to facilitate model transformations such as in https://github.com/pymc-devs/pymcx/pull/91

Closes #6305 
Closes #5076

## Major / Breaking Changes
- Sampling of transformed variables from `prior_predictive` is no longer allowed

## Bugfixes / New features
- ...

## Docs / Maintenance
- Rename several internal `Model` variables

### DIFF
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -343,7 +343,7 @@ def sample_prior_predictive(
     var_names : Iterable[str]
         A list of names of variables for which to compute the prior predictive
         samples. Defaults to both observed and unobserved RVs. Transformed values
-        are not included unless explicitly defined in var_names.
+        are not allowed.
     random_seed : int, RandomState or Generator, optional
         Seed for the random number generator.
     return_inferencedata : bool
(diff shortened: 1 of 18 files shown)


michaelosthege (approved):
Trusting that you'll fix that one test 👍

pymc/sampling/forward.py
```diff
@@ -343,7 +362,7 @@ def sample_prior_predictive(
     var_names : Iterable[str]
         A list of names of variables for which to compute the prior predictive
         samples. Defaults to both observed and unobserved RVs. Transformed values
-        are not included unless explicitly defined in var_names.
+        are not allowed.
```
pipme, line 365: Any reason why transformed (unconstrained) values are not allowed here? I need to get prior samples in the transformed space, what should I do? Thanks! 

(MCMC sampling by removing observations from the likelihood is not a good option. I guess I could also use `sample_prior_predictive` to get prior parameter samples and then transform back to the unconstrained space but it seems not easy to do it with `Transforms` in an efficient vectorized way, i.e., it seems I have to do a for-loop.)

pipme:
Hi, please see my comment above. Any reason why transformed (unconstrained) values are not allowed in `sample_prior_predictive`? I need to get prior samples in the transformed space, what should I do? Thanks!

pymc/sampling/forward.py
ricardoV94: A helper to go from unconstrained back to constrained draws is needed elsewhere (specially since we don't save them in InferenceData after sampling), so we can add that and will also cover your use case.

I think there is an open issue for that.

ricardoV94: This issue: https://github.com/pymc-devs/pymc/issues/6721

pipme: Thanks for the quick reply! Transforming efficiently between unconstrained and constrained draws would be super useful for our use case. What would be needed for its implementation? In the meantime, what could be a good workaround? (I need to transform a lot of prior samples from constrained to unconstrained. Or alternatively, unconstrained prior samples are even more useful to me than constrained samples. So it would be better if I could get them directly.) Can I modify back `sample_prior_predictive` to include the transformed variables like `sd_log__`?

For the InferenceData, `pm.sample(..., idata_kwargs={"include_transformed": True})` seems to include the `transformed` samples.

ricardoV94:
No we shouldn't change `sample_prior_predictive`, but you can of course change in your local installation if that's an option for you. Or you can do something like this for yourself: https://discourse.pymc.io/t/logp-questions-synthetic-dataset-to-evaluate-modeling/12129/6?u=ricardov94

I'll open a PR soon to add the functionality.

PR #6312, opened 2022-11-18
Add warning about change of sign in hessian functions

### LINKED ISSUE #6310: Hessian (d2logp) has the wrong sign
In PyMC, the Hessian of the a model is [defined as](https://github.com/pymc-devs/pymc/blob/7ccfdaa891ec05c5af2ba69ad548ebefbc6882f3/pymc/aesaraf.py#L520) $-\nabla\nabla\log f$ and not $\nabla\nabla\log f$ which is the generally accepted mathematical definition, as the collection of second-order derivatives.

I accept that there is probably a good reason for computing the negative Hessian rather than the Hessian but, for clarity when reading the code and documentation, could these functions be renamed to indicate that this is the negative Hessian?

Context: For pedagogical/demonstration purposes, I am implementing the [Laplace approximation](https://en.wikipedia.org/wiki/Laplace%27s_approximation) using the `logp`, `dlogp`, and `d2logp` functions and the sign error in `d2logp` took me a long time to figure out, eventually by looking through the code and finding the definition of `aesaraf.hessian`.

aseyboldt:
The hessian function incorrectly returned the negative of the hessian.

This does change the output of `Model.d2logp` however, so we should mention this in the release notes.

## Major / Breaking Changes
Sign of the hessian is fixed, and therefore different than before

Fixes #6310

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -658,6 +658,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         self,
         vars: Variable | Sequence[Variable] | None = None,
         jacobian: bool = True,
+        negate_output=True,
         **compile_kwargs,
     ) -> PointFunc:
         """Compiled log probability density hessian function.
@@ -670,7 +671,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
         jacobian : bool
             Whether to include jacobian terms in logprob graph. Defaults to True.
         """
-        return self.compile_fn(self.d2logp(vars=vars, jacobian=jacobian), **compile_kwargs)
+        return self.model.compile_fn(
+            self.d2logp(vars=vars, jacobian=jacobian, negate_output=negate_output),
+            **compile_kwargs,
+        )
 
     def logp(
         self,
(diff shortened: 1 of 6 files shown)


ricardoV94:
I opted to something like this when we changed the sign of an operation before: https://github.com/pymc-devs/pymc/blob/4acd98ee2bc31ef1fb63accdfb1012ba64fcebd5/pymc/math.py#L308-L331

If this is critical enough, we can do the same for the output of the hessian

maresb:
Without understanding much context, as a geometer I must warn that a common sign convention for the Hessian in geometry and physics is indeed _negative_ nabla-squared.

The reason for this is that it makes the Hessian _positive_ semi-definite on plane-waves of the form $e^{i k x}$. Namely,

$$-\nabla^2 e^{i k x} = +k^2 e^{i k x}.$$

Nerd rant over.  :nerd_face:

ricardoV94:
@maresb I think that keeping it in line with PyTensor is a good reason enough?

@aseyboldt I added a FutureWarning that can be disabled with `negate_output=False`, and kept the default behavior the same. WDYT? Do you mind reviewing my changes? I also changed other places in the codebase to behave the same as before (expect a non-negated output and so on).

maresb (approved):
Given that the typical user of PyMC isn't a differential geometer, I approve of this approach. I like @ricardoV94's deprecation plan

I'm still a bit :exploding_head: that this appears to be an accidental(?) rediscovery of the geometer's convention for the Laplacian. Although perhaps not too surprising, since when considering Bayesian stats as real-valued quantum mechanics, the minus sign that makes -logp positive-ish is the same minus sign that geometers like since it gives a positive spectrum.

PR #6313, opened 2022-11-18
Require all step methods to return stats

### LINKED ISSUE #6270: Require step methods to emit sampler stats
As a follow-up to https://github.com/pymc-devs/pymc/pull/6205 we should require step methods (`astep` in particular) to always return stats.

Doing this will allow us to simplify code in places like these:

https://github.com/pymc-devs/pymc/blob/9c313cb61797fea6d1f04dba0d4797c64a1271df/pymc/sampling.py#L976-L983

https://github.com/pymc-devs/pymc/blob/9c313cb61797fea6d1f04dba0d4797c64a1271df/pymc/parallel_sampling.py#L194-L200

michaelosthege:
The reason for these changes are the resulting simplification of code, including simpler branching and less type ambiguity.

This change may break custom step methods, but whoever had enough skill to implement a custom step method should have no trouble to append a `, []` to the `return` of their `(a)step` method.

Closes #6270

**Checklist**

## Major / Breaking Changes
- Reuires all step methods to return stats from their `step`/`asteo` method.
- The `BlockedStep.generates_stats` attribute was removed.

## Bugfixes / New features
- Removed a branch in `Compound.step` that supported `namedtuple` `stats` which are in violation of the `BlockedStep.step` signature.

## Docs / Maintenance
- Added & fixed some type hints on `step` and `astep` methods. (`arraystep.py` now passing mypy!)

### DIFF
--- a/pymc/step_methods/arraystep.py
+++ b/pymc/step_methods/arraystep.py
@@ -14,21 +14,19 @@
 
 from abc import ABC, abstractmethod
 from enum import IntEnum, unique
-from typing import Dict, List, Tuple, TypeVar, Union
+from typing import Callable, Dict, List, Tuple, Union, cast
 
 import numpy as np
 
 from aesara.graph.basic import Variable
 from numpy.random import uniform
 
-from pymc.blocking import DictToArrayBijection, PointType, RaveledVars
+from pymc.blocking import DictToArrayBijection, PointType, RaveledVars, StatsType
 from pymc.model import modelcontext
 from pymc.step_methods.compound import CompoundStep
 from pymc.util import get_var_name
 
 __all__ = ["ArrayStep", "ArrayStepShared", "metrop_select", "Competence"]
-
-StatsType = TypeVar("StatsType")
 
 
 @unique
@@ -49,7 +47,6 @@ class Competence(IntEnum):
 
 class BlockedStep(ABC):
 
-    generates_stats = False
     stats_dtypes: List[Dict[str, type]] = []
     vars: List[Variable] = []
 
@@ -103,7 +100,7 @@ class BlockedStep(ABC):
         return self.__newargs
 
     @abstractmethod
-    def step(point: PointType, *args, **kwargs) -> Union[PointType, Tuple[PointType, StatsType]]:
+    def step(self, point: PointType) -> Tuple[PointType, StatsType]:
         """Perform a single step of the sampler."""
 
     @staticmethod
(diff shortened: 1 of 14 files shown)


pymc/step_methods/arraystep.py
```diff
@@ -146,35 +143,28 @@ def __init__(self, vars, fs, allvars=False, blocked=True):
...
-            return point_new, stats
-
-        return point_new
+        return point_new, stats
 
     @abstractmethod
-    def astep(
-        self, apoint: RaveledVars, point: PointType, *args
-    ) -> Union[RaveledVars, Tuple[RaveledVars, StatsType]]:
+    def astep(self, apoint: RaveledVars, *args) -> Tuple[RaveledVars, StatsType]:
```
michaelosthege, line 167: The exact composition of the `args` is specified via the constructor parameters 🤐

pymc/step_methods/metropolis.py
```diff
@@ -400,8 +399,8 @@ def __init__(self, vars, scaling=1.0, tune=True, tune_interval=100, model=None):
 
         super().__init__(vars, [model.compile_logp()])
 
-    def astep(self, q0: RaveledVars, logp) -> Tuple[RaveledVars, List[Dict[str, Any]]]:
-
+    def astep(self, q0: RaveledVars, *args) -> Tuple[RaveledVars, StatsType]:
+        logp = args[0]
```
michaelosthege, line 393: `BinaryMetropolis` is not even covered by our tests!

pymc/step_methods/metropolis.py
```diff
@@ -502,8 +501,8 @@ def __init__(self, vars, order="random", transit_p=0.8, model=None):
 
         super().__init__(vars, [model.compile_logp()])
 
-    def astep(self, q0: RaveledVars, logp: Callable[[RaveledVars], np.ndarray]) -> RaveledVars:
-
+    def astep(self, q0: RaveledVars, *args) -> Tuple[RaveledVars, StatsType]:
+        logp: Callable[[RaveledVars], np.ndarray] = args[0]
```
michaelosthege: `BinaryGibbsMetropolis` is also not covered by tests!

pymc/step_methods/metropolis.py
```diff
@@ -67,38 +67,29 @@ def __init__(self, s):
 
 class NormalProposal(Proposal):
     def __call__(self, rng: Optional[np.random.Generator] = None):
-        if rng is None:
-            rng = nr
-        return rng.normal(scale=self.s)
+        return (rng or nr).normal(scale=self.s)
```
michaelosthege, line 71: These changes resolve two `mypy` errors:
1. `rng` is no longer reassigned, and 
2. the `(rng or nr)` has type `Generator | Module("numpy.random")` which resolves `Item "None" of "Optional[Generator]" has no attribute "normal"`

ricardoV94:
We shouldn't remove the trust_input. It allows evaluating the Aesara function much faster because it doesn't check the input types are valid/need to be converted every time it's called

michaelosthege:
> We shouldn't remove the trust_input. It allows evaluating the Aesara function much faster because it doesn't check the input types are valid/need to be converted every time it's called

But we're not using the `trust_input` anywhere in the codebas! Or is it a feature of the Aesara function that is not correctly advertised by the API?
Because mypy complained about it

michaelosthege:
Ah I see, it's some feature of the function... I'll investigate and drop that commit.

Anything else you want to get changed, @ricardoV94 ?

pymc/step_methods/metropolis.py
ricardoV94: Looks much less readable though

pymc/step_methods/metropolis.py
```diff
@@ -1056,7 +1036,7 @@ def delta_logp(
     logp: at.TensorVariable,
     vars: List[at.TensorVariable],
     shared: Dict[at.TensorVariable, at.sharedvar.TensorSharedVariable],
-) -> aesara.compile.Function:
+):
```
ricardoV94: Wasn't this type hint correct?

michaelosthege: yes, it was correct. ~Now it's taken from one level deeper:~ `compile_pymc` had the wrong return type hint

ricardoV94 (approved):

PR #6316, opened 2022-11-20
Removed NoDistribution from docs

stestoll:
Removes a doc reference to NoDistribution, which is no longer implemented.


michaelosthege (approved):

michaelosthege:
well spotted! thanks @stestoll

PR #6331, opened 2022-11-22
Only add `total_size` mapping to observed RVs

ricardoV94:
This fixes inconsistent use of the `rvs_to_total_sizes` mapping. It only assigns observed_RVs to it, including the observed component of an imputed variable (even though that is not allowed at the moment).

It also raises if passed to `free_RVs`

## Docs / Maintenance
- Cleanup `total_size` interface


ricardoV94:
Superseded by #6523

PR #6342, opened 2022-11-28
Fix measurable stack and join with interdependent inputs

ricardoV94:
This fixes measurable stacks and joins that have interdependent variables. 

The main loop of `factorized_joint_logprob` replaces all the measurable variables already seen by their values before calling the logp of a new measurable node, but it does not replace said node by its value variable, as it assumes the returned logprob expression never depends on the original measurable (stochastic) node. 

However this can happen in Stack and Join whenever the base measurable RVs depend on one another. The new tests show this clearly. The solution proposed in this PR is to do the additional replacements needed in the respective logp functions.

## Major / Breaking Changes
- ...

## Bugfixes / New features
- Fix logprob of stack and join when base variables are interdependent

## Docs / Maintenance
- ...

### DIFF
--- a/pymc/logprob/tensor.py
+++ b/pymc/logprob/tensor.py
@@ -129,11 +129,24 @@ MeasurableVariable.register(MeasurableMakeVector)
 
 
 @_logprob.register(MeasurableMakeVector)
-def logprob_make_vector(op, values, *base_vars, **kwargs):
+def logprob_make_vector(op, values, *base_rvs, **kwargs):
     """Compute the log-likelihood graph for a `MeasurableMakeVector`."""
+    # TODO: Sort out this circular dependency issue
+    from pymc.pytensorf import replace_rvs_by_values
+
     (value,) = values
 
-    return at.stack([logprob(base_var, value[i]) for i, base_var in enumerate(base_vars)])
+    base_rvs_to_values = {base_rv: value[i] for i, base_rv in enumerate(base_rvs)}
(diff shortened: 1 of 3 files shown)


pymc/logprob/tensor.py
```diff
@@ -126,11 +126,28 @@ class MeasurableMakeVector(MakeVector):
...
+        base_rv.name = f"base_rv[{i}]"
+        value.name = f"value[{i}]"
+
+    logps = [logprob(base_rv, value) for base_rv, value in base_rvs_to_values.items()]
+
+    # If the stacked variables depend on each other, we have to replace them by the respective values
+    logps = replace_rvs_by_values(
+        logps,
+        rvs_to_values=base_rvs_to_values,
+        rvs_to_transforms={rv: None for rv in base_rvs},
```
michaelosthege: How do we know that there are no transforms?

michaelosthege (approved):
damn, I forgot to submit my review comments

pymc/logprob/tensor.py
ricardoV94: The transforms are applied before the RV-specific logp function is called. The value received is already the one in the constrained space.

ricardoV94: https://github.com/pymc-devs/pymc/blob/ddc6b65b19a2e9f2e6e4a6c2ca62db6cd18d66a2/pymc/logprob/transforms.py#L591-L608

ricardoV94:
CC @tomicapretto

PR #6344, opened 2022-11-28
Add nu parametrization to beta distribution

### LINKED ISSUE #5023: Invalid values being used for testing Beta distribution

This was pointed out in #4887 that the testing for `Beta` distribution used values that might lead to invalid combination of arguments in the following test: 

https://github.com/pymc-devs/pymc3/blob/53e572caa3144f77951e233f6bcca72b63084bd8/pymc3/tests/test_distributions.py#L1164-L1169

The values for `mu` and `sigma` are picked randomly from their respective `Domain`. However in this particular case we only have certain value pairs which are valid. In the current `v4` testing it simply returns `NaN` values for invalid cases. With #4887 we will require a new mechanism/value `Domain`s to be placed over here.

### LINKED ISSUE #6295: DOC: wrong bound for `sigma` in docstring of `Beta`
The docstring of `Beta` says

```
sigma : tensor_like of float, optional
    Alternative standard deviation (1 < ``sigma`` < sqrt(``mu`` * (1 - ``mu``))).
```

while it should be `0 < sigma < ...`.

(Related issue: #5023 says that `sigma` may be dropped in favor of `nu`)

* PyMC/PyMC3 Version: 4.3.0
* Aesara/Theano Version: 2.8.7
* Python Version: 3.10.5
* Operating system: macos 12.6.1 arm
* How did you install PyMC/PyMC3: pip

alekracicot:
**What is this PR about?**
...
This PR addresses closes #5023 and closes #6295 and incorporates the mu, nu parametrization to the Beta distribution as described here https://en.wikipedia.org/wiki/Beta_distribution as well as fixing the error on the sigma bounds in the docstring. 

As for the test, I implemented one, on the logp, and it passed. I assumed that if logp was the same across parametrization, cdf and others would follow. Let me know if this is sufficient. 

**Checklist**

## Major / Breaking Changes
- ...

## Bugfixes / New features
- The `Beta` distribution can now be parametrized by `nu`.

## Docs / Maintenance

- ...


pymc/distributions/continuous.py
```diff
@@ -1081,7 +1081,9 @@ class Beta(UnitContinuous):
     mu : tensor_like of float, optional
         Alternative mean (0 < ``mu`` < 1).
     sigma : tensor_like of float, optional
-        Alternative standard deviation (1 < ``sigma`` < sqrt(``mu`` * (1 - ``mu``))).
+        Alternative standard deviation (0 < ``sigma`` < sqrt(``mu`` * (1 - ``mu``))).
+    nu : tensor_like of float, optional
```
ricardoV94, line 1088: Missing the relationship between the nu and alpha/beta in the math section above

pymc/tests/distributions/test_continuous.py
```diff
@@ -373,6 +375,12 @@ def test_beta_logp(self):
             {"mu": Unit, "sigma": Rplus},
             beta_mu_sigma,
         )
+        check_logp(
```
ricardoV94: No need to test the logp (which is a heavy test). Should be enough to add a test like this: https://github.com/pymc-devs/pymc/blob/3ff4e7af14afbb5f3221572657e04f97e1ae4b44/pymc/tests/distributions/test_continuous.py#L2013-L2020

ricardoV94 (changes requested):
Thanks for opening the PR. I left a couple of suggestion below.

pymc/distributions/continuous.py
alekracicot: I added the relationship in the math section.

pymc/tests/distributions/test_continuous.py
alekracicot: Thank you for the example, I added the equivalent test and removed the logp test.

pymc/distributions/continuous.py
```diff
@@ -1059,10 +1059,11 @@ class Beta(UnitContinuous):
     Support   :math:`x \in (0, 1)`
     Mean      :math:`\dfrac{\alpha}{\alpha + \beta}`
     Variance  :math:`\dfrac{\alpha \beta}{(\alpha+\beta)^2(\alpha+\beta+1)}`
+    Sample size  :math:`\alpha + \beta`
```
ricardoV94: I am not sure about this one. It's not really a property of the distribution, but an interpretation of the parameters, assuming a specific model

ricardoV94 (commented):
Looks good. Just not sure about the "sample size" part in the docstrings.

pymc/distributions/continuous.py
fonnesbeck: You can add this as a comment in the docstring, but I agree that this is not a property per se.

alekracicot: I was hesitating as well, I will remove the reference to "sample size" without the full explanation. Should I add:

'nu : tensor_like of float, optional
        Alternative "sample size" of a Beta distribution **for the Haldane prior probability Beta(0,0)** (``nu`` > 0) .

in the parameter section?

ricardoV94: I think the parameter description is good enough as it is now. 

We can perhaps add a link to the relevant Wikipedia section instead of mentioning the Haldane prior?

https://en.m.wikipedia.org/wiki/Beta_distribution#Mean_and_sample_size

ricardoV94:
Pre-commit is failing, otherwise the PR looks good to me

michaelosthege (approved):

PR #6356, opened 2022-11-29
Fix VI `var_replacement` test

### LINKED ISSUE #6348: VI variable replacement broken with Aesara 2.8.8
### Description

We added an xfail to this [test]() in #6336. The issue is that this VI method is trying to replace an RV by a larger one. This is not allowed when the graph is created with a fixed size RV.

The test case doesn't seem needed at all, one can create a variable sized RV already with a shared variable and then resize it. Would this fulfill all the needs?

ricardoV94:
Closes #6348

### DIFF
--- a/pymc/tests/variational/test_inference.py
+++ b/pymc/tests/variational/test_inference.py
@@ -307,17 +307,19 @@ def test_remove_scan_op():
         buff.close()
 
 
-@pytest.mark.xfail(reason="Broke from static shape handling with Aesara 2.8.8")
 def test_var_replacement():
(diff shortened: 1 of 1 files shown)


ricardoV94:
Failure is unrelated

ferrine (approved):

ferrine:
What is wrong with pymc/tests/distributions/test_timeseries.py::TestGARCH11::test_batched_size ?

ferrine:
is it safe to merge with this test?

ricardoV94:
Yes, test is completely unrelated

PR #6361, opened 2022-12-01
Support logp derivation in `DensityDist`

### LINKED ISSUE #4507: Feature request: Add offset parameter to Exponential class (akin to mu in Laplace class)

### LINKED ISSUE #4530: Consistent API for user friendly distribution customization

ricardoV94:
This PR allows users to return Aesara random graphs in the `random` method of `DensityDist` (now renamed to `CustomDist`), to be used in prior and posterior predictive sampling, as well as for automatic logp derivation using PyMC machinery.

For now, using this functionality will issue an experimental warning.

Closes #4530
Closes #4507

```python
import pymc as pm

def shifted_exponential(lam, shift, size):
    return pm.Exponential.dist(lam, size=size) + shift

with pm.Model() as m:
    lam = pm.HalfNormal("lam")
    shift = -1
    custom_dist = pm.CustomDist("custom_dist", lam, shift, random=shifted_exponential, observed=[-1, -1, 0])

    posterior = pm.sample()
```

Here is an example that address this discourse help request: https://discourse.pymc.io/t/inference-with-aesara-pymc/10945

```python
import pymc as pm

def pert(low, peak, high, lmbda, size):
    range = (high - low)
    s_alpha = 1 + lmbda * (peak - low) / range
    s_beta = 1 + lmbda * (high - peak) / range
    return pm.Beta.dist(s_alpha, s_beta, size=size) * range + low

with pm.Model() as m:
    low = pm.Normal("low", 0, 10)
    peak = pm.Normal("peak", 50, 10)
    high = pm.Normal("high", 100, 10)
    lmbda = 4
    pm.CustomDist("pert", low, peak, high, lmbda, random=pert, observed=[30.0, 35.0, 73.0])

m.point_logps()
```

## Major / Breaking Changes
- ...

## Bugfixes / New features
- `CustomDist` now accepts Aesara random graphs via the `random` kwarg. PyMC can sometimes infer the logp of these distributions without the need for a user to provide a `logp` function. This functionality is experimental!

## Docs / Maintenance
- `DensityDist` is now an alias to `CustomDist`

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -26,18 +26,20 @@ import numpy as np
 from pytensor import tensor as at
 from pytensor.compile.builders import OpFromGraph
 from pytensor.graph import node_rewriter
-from pytensor.graph.basic import Node, clone_replace
+from pytensor.graph.basic import Node, Variable, clone_replace
 from pytensor.graph.rewriting.basic import in2out
 from pytensor.graph.utils import MetaType
 from pytensor.tensor.basic import as_tensor_variable
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.type import RandomType
+from pytensor.tensor.random.utils import normalize_size_param
 from pytensor.tensor.var import TensorVariable
 from typing_extensions import TypeAlias
 
 from pymc.distributions.shape_utils import (
     Dims,
     Shape,
+    _change_dist_size,
     convert_dims,
     convert_shape,
     convert_size,
@@ -53,7 +55,7 @@ from pymc.logprob.abstract import (
 )
 from pymc.logprob.rewriting import logprob_rewrites_db
 from pymc.printing import str_for_dist
-from pymc.pytensorf import convert_observed_data
+from pymc.pytensorf import collect_default_updates, convert_observed_data
 from pymc.util import UNSET, _add_future_warning_tag
 from pymc.vartypes import string_types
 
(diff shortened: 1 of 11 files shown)


pymc/distributions/distribution.py
```diff
@@ -473,123 +475,447 @@ def rng_fn(cls, rng, *args):
...
+        logp: Optional[Callable],
+        logcdf: Optional[Callable],
+        moment: Optional[Callable],
+        size=None,
+        ndim_supp: int,
+    ):
+        size = normalize_size_param(size)
+        dummy_size_param = size.type()
+        dummy_dist_params = [dist_param.type() for dist_param in dist_params]
+        dummy_rv = random(*dummy_dist_params, dummy_size_param)
```
ricardoV94: Can we block access to the `Model` context so that if users try to call `pm.Normal("x")` in the `random` function it errors out?

ricardoV94:
Having some formatting troubles with the code examples, only the last one shows: https://pymcio--6361.org.readthedocs.build/projects/docs/en/6361/api/distributions/generated/pymc.CustomDist.html#pymc.CustomDist

pymc/distributions/distribution.py
ricardoV94: Something like 

```python
with error_on_model_access():
  random(*dummy_dist_params, dummy_size_param)
```

ricardoV94: Implemented one approach in the last commit

pymc/distributions/distribution.py
```diff
@@ -686,29 +557,477 @@ def rv_op(
...
+            f"CustomSymbolicDistRV_{class_name}",
+            (CustomSymbolicDistRV,),
+            # If logp is not provided, we infer it from the random graph
+            dict(
+                inline_logprob=logp is None,
+            ),
+        )
+
+        # Dispatch custom methods
+        if logp is not None:
```
ricardoV94, line 684: else: dispatch the one user provided

zaxtax:
Do you think it might be good to have a stub `DensityDist` so users know they should migrate to `CustomDist`?

ricardoV94:
> Do you think it might be good to have a stub `DensityDist` so users know they should migrate to `CustomDist`?

It's there. Didn't even deprecate it because it's so common.

michaelosthege (approved):

PR #6369, opened 2022-12-05
Do not try to infer the logprob of broadcasted or advanced indexed variables

ricardoV94:
This PR removes an invalid rewrite from the logprob submodule

It also suppresses a NotImplementedError in the Mixture rewrite. Individual rewrites should not judge whether logprob for a graph is implemented or not, because other rewrites might be able to handle it.

### DIFF
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -445,6 +445,10 @@ def logprob_MixtureRV(
             logp_val = at.set_subtensor(logp_val[idx_m_on_axis], logp_m)
 
     else:
+        # FIXME: This logprob implementation does not support mixing across distinct components,
+        # but we sometimes use it, because MixtureRV does not keep information about at which
+        # dimension scalar indexing actually starts
+
         # If the stacking operation expands the component RVs, we have
         # to expand the value and later squeeze the logprob for everything
         # to work correctly
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -58,13 +58,15 @@ from pytensor.tensor.random.rewriting import (
 )
 from pytensor.tensor.shape import shape_tuple
 from pytensor.tensor.subtensor import (
+    AdvancedSubtensor,
+    AdvancedSubtensor1,
     as_index_literal,
     as_nontensor_scalar,
     get_canonical_form_slice,
     is_basic_idx,
 )
 from pytensor.tensor.type import TensorType
-from pytensor.tensor.type_other import NoneConst, NoneTypeT, SliceType
+from pytensor.tensor.type_other import NoneConst, NoneTypeT, SliceConstant, SliceType
 from pytensor.tensor.var import TensorVariable
 
 from pymc.logprob.abstract import (
@@ -308,6 +310,17 @@ def mixture_replace(fgraph, node):
         return None  # pragma: no cover
 
     mixing_indices = node.inputs[1:]
+
+    # TODO: Add check / test case for Advanced Boolean indexing
+    if isinstance(node.op, (AdvancedSubtensor, AdvancedSubtensor1)):
+        # We don't support (non-scalar) integer array indexing as it can pick repeated values,
+        # but the Mixture logprob assumes all mixture values are independent
+        if any(
+            indices.dtype.startswith("int") and sum(1 - b for b in indices.type.broadcastable) > 0
+            for indices in mixing_indices
+            if not isinstance(indices, SliceConstant)
+        ):
+            return None
 
     # We loop through mixture components and collect all the array elements
     # that belong to each one (by way of their indices).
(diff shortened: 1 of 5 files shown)


pymc/tests/logprob/test_transforms.py
```diff
@@ -709,13 +709,13 @@ def test_discrete_rv_multinary_transform_fails():
         joint_logprob({y_rv: y_rv.clone()})
 
 
-@pytest.mark.xfail(reason="Check not implemented yet, see #51")
+@pytest.mark.xfail(reason="Check not implemented yet")
```
ricardoV94, line 781: Add companion test with explicit `broadcast_to` which now fails after this PR

pymc/logprob/mixture.py
```diff
@@ -407,6 +418,8 @@ def logprob_MixtureRV(
 
     assert len(indices) > 0
 
+    # FIXME: This logic does not account for repeated indexes, in which case the Mixture values are not independent.
+    #  It over-counts the logp terms for values that were repeated during indexing
     if len(indices) > 1 or indices[0].ndim > 0:
```
ricardoV94: Make sure we have tests for still supported multiple indexes

```
Normal.dist(size=(2, 3)[0, 0]
```

ricardoV94: Remove or branch (Advanced Indexing)?

zaxtax:
Are we guaranteed to always be able to later rewrite this safely into something that is a proper logprob?

ricardoV94:
> Are we guaranteed to always be able to later rewrite this safely into something that is a proper logprob?

There are two types of logps that are relevant here. An RV that is broadcasted directly and an RV where we are doing advanced indexing (i.e. mixture)

For the former, we would need to perform symbolic unbroadcasting of the value variable, and then evaluate the base RV at that value

```python
import pytensor.tensor as pt

def logp_broadcast_to(value, base_rv, broadcast_shape):
  return logp(base_rv, unbroadcast(value, broadcast_shape))
```

I am not sure what the `unbroadcast` function would look like, but it should do the following (in pseudo-code), when valid, and raise otherwise.
```python
np.testing.assert_array_equal(unbroadcast(broadcast_to(value, shape), shape), value)
```

For the second, we would need to somehow only evaluate the logp at the unique indexes. Something like 

```python
def logp_mixture_advanced_indexing(value, base_rvs, indexes):
  unique_indexes = get_unique_indexes(unique_indexes)
  # And either ignore or assert the value shoud be the same for repeated indexes
  return raveled_mixture_rv(value[unique_indexes], base_rvs, unique_indexes)
```

Neither seems trivial to implement, which is why this PR does not attempt to do so.

zaxtax:
Just to clarify, are there rewrites this allows that weren't allowed before? It seems to be upon closer reading more restrictive about where rewrites are allowed to happen. In either case what we do for Case 1 vs. Case 2 should probably be different.

pymc/logprob/mixture.py
```diff
@@ -300,13 +300,27 @@ def mixture_replace(fgraph, node):
 
     old_mixture_rv = node.default_output()
 
-    mixture_res, join_axis = get_stack_mixture_vars(node)
+    try:
+        mixture_res, join_axis = get_stack_mixture_vars(node)
+    except NotImplementedError:
```
zaxtax: This seems like a code smell. Would it be unreasonable to refactor `get_stack_mixture_vars` to not raise this exception?

pymc/logprob/mixture.py
```diff
@@ -404,6 +418,8 @@ def logprob_MixtureRV(
 
     assert len(indices) > 0
 
+    # FIXME: This logic does not account for repeated indexes, in which case the Mixture values are not independent.
+    #  It over-counts the logp terms for values that were repeated during indexing
```
zaxtax: Should this have a unit test when it's resolved?

pymc/logprob/mixture.py
ricardoV94: I refactored it to avoid raising the exception

pymc/logprob/mixture.py
ricardoV94: I don't know if we will ever resolve it. For now we don't allow creating such MixtureRVs in the first place. The ones that hit this branch don't do AdvancedIndexing

zaxtax:
> > Just to clarify, are there rewrites this allows that weren't allowed before? It seems to be upon closer reading more restrictive about where rewrites are allowed to happen. In either case what we do for Case 1 vs. Case 2 should probably be different.
> 
> Yes this PR restricts the type of rewrites that are possible, and does not enable any new ones. I am not sure we need to bother with Case1, but we can explore it. Case2 should definitely be allowed (at least the case that does not have repeated indexes). Would you mind opening an issue to track that?

I don't think I understand it well-enough to write a good issue for it.

ricardoV94:
Converted to a draft, as this and #6397 touch on similar issues

zaxtax:
LGTM!

PR #6370, opened 2022-12-05
Simplify measurable transform rewrites

### LINKED ISSUE #6353: Order of rewrites in logprob submodule is not respected
### Description

I could be wrong, but when working on https://github.com/pymc-devs/pymc/pull/6072, it seemed that the order of the rewrites in the database is only respected when using the kwarg `position`. Otherwise it seems to be ignored, and only the order of definition matters.

So: 
https://github.com/pymc-devs/pymc/blob/e0d25c899a555d7e17f683f4a42c06b933287783/pymc/logprob/rewriting.py#L263

Should be:
```python
logprob_rewrites_db.register( 
     "measurable_ir_rewrites", measurable_ir_rewrites_db, "basic", position=-10, 
 )
```

We should probably also leave an offset between `pre-canonicalize` and `measurable_ir_rewrites` (both are supposed to be at -10) in case anyone wants to introduce rewrites between those two phases.

ricardoV94:
Just some cleanup of the logprob submodule, removing functionality we don't need or are not using in PyMC

Closes #6353

### DIFF
--- a/pymc/logprob/censoring.py
+++ b/pymc/logprob/censoring.py
@@ -110,7 +110,6 @@ def find_measurable_clips(fgraph: FunctionGraph, node: Node) -> Optional[List[Me
 measurable_ir_rewrites_db.register(
     "find_measurable_clips",
     find_measurable_clips,
-    0,
     "basic",
     "censoring",
 )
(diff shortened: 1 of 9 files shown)


ricardoV94:
I would like to get this one merged before the group hackathon tomorrow. Will make it easier to review the other more-interesting PRs

twiecki (approved):

PR #6371, opened 2022-12-05
Implement logp derivation for division, subtraction and negation

ricardoV94:
This implements measurable rewrites for the following type of graphs:

```python
import pymc as pm

x = pm.Gamma.dist(2, 1)
y = 1 / x
pm.logp(y, 0.5).eval()

```
As well as:
```
y = 5 / x
y = -x
y = 5 - x
```

It works by canonicalizing such operations to the form already understood by `find_measurable_transforms` (only a new condition for `Reciprocal`s was added)

```
(5 / x) -> 5 * reciprocal(x)
(-x) -> x * (-1)
5 - x -> 5 + (x * (-1))
```

These canonicalizations are only applied when MeasurableVariables are involved to minimize disruption of the underlying graph

## Bugfixes / New features
- Automatic logprob derivation for division, subtraction, and negation operations

### DIFF
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -50,7 +50,7 @@ from pytensor.graph.rewriting.basic import GraphRewriter, in2out, node_rewriter
 from pytensor.scalar import Add, Exp, Log, Mul, Reciprocal
 from pytensor.scan.op import Scan
 from pytensor.tensor.exceptions import NotScalarConstantError
-from pytensor.tensor.math import add, exp, log, mul, reciprocal, true_div
+from pytensor.tensor.math import add, exp, log, mul, neg, reciprocal, sub, true_div
 from pytensor.tensor.rewriting.basic import (
     register_specialize,
     register_stabilize,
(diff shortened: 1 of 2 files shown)


pymc/logprob/transforms.py
```diff
@@ -338,6 +370,15 @@ def find_measurable_transforms(fgraph: FunctionGraph, node: Node) -> Optional[Li
     return [transform_out]
 
 
+measurable_ir_rewrites_db.register(
+    "measurable_div_to_reciprocal_product",
+    measurable_div_to_reciprocal_product,
+    -1,
```
ricardoV94: Remove -1

ricardoV94:
CC @tomicapretto

Armavica:
Sorry for the very naive question, but I think I am still missing some pieces of the puzzle. I tried to understand what new models this will allow to write, but it looks like
``` python
with pm.Model():
    x = pm.Normal("x")
    y = pm.Normal("y", 1/x, 1)
```
was already possible before. Why is the `1/x` working here? Am I completely off tracks?

ricardoV94:
> Sorry for the very naive question, but I think I am still missing some pieces of the puzzle. I tried to understand what new models this will allow to write, but it looks like
> 
> ```python
> with pm.Model():
>     x = pm.Normal("x")
>     y = pm.Normal("y", 1/x, 1)
> ```
> 
> was already possible before. Why is the `1/x` working here? Am I completely off tracks?

No, you're not completely off tracks. That's a parameter transformation, not a variable transformation. 

The difference between the two is easier if you just think about the logp you are requestiong:

```python
some_value = 1.0

x = pm.Normal.dist()
y = pm.Normal.dist(1/x)
pm.logp(y, some_value)  # Fine before

z = 1 / x
pm.logp(z, some_value)  # Failed before, now it works!
```

This is more relevant for observed variables, so that you can condition on them directly. After this and #6361 you can do this:

```python
def inverse_normal(mu, sigma, size):
  return 1 / pm.Normal.dist(mu, sigma, size=size)

with pm.Model() as m:
  x = pm.Normal("x")
  y = pm.CustomDist("y", mu, 1, random=inverse_normal, observed=data
```

PyMC will now be able to derive the logp of `y`. Another thing that might be confusing, is that this is also not the same as `pm.Normal("y", mu, 1, observed=1/data)`

michaelosthege (approved):
Elementary school math. I feel confident about this 👍

PR #6373, opened 2022-12-07
Add informative error if user mixes Aesara functions with PyTensor variables

ricardoV94:
To help with the V5 transition. This will happen now:

```python
import aesara.tensor as at
import pymc as pm

at.exp(pm.Normal.dist())
```

```
TypeError: You are calling an Aesara function with PyTensor variables. Did you import aesara.tensor instead of pytensor.tensor?
```

I didn't add a test, to not require Aesara as a dependency.

### DIFF
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -92,6 +92,23 @@ __all__ = [
 ]
 
 
+# Temporarily add an Aesara dispatch for PyTensor TensorVariable to inform users
+# of incorrect mixing of libraries
+try:
+    from aesara.tensor import _as_tensor_variable as _as_tensor_variable_aesara
+
+    @_as_tensor_variable_aesara.register(TensorVariable)
+    def raise_informative_error(*args, **kwargs):
+        raise TypeError(
+            "You are calling an Aesara function with PyTensor variables.\n"
+            "Starting with PyMC 5.0, Aesara was replaced by PyTensor (see https://www.pymc.io/blog/pytensor_announcement.html).\n"
+            "Replace your import of aesara.tensor with pytensor.tensor.",
+        )
+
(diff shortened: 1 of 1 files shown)


michaelosthege (approved):
good idea 👍

pymc/pytensorf.py
```diff
@@ -92,6 +92,22 @@
...
 
+# Temporarily add an Aesara dispatch for PyTensor TensorVariable to inform users
+# of incorrect mixing of libraries
+try:
+    from aesara.tensor import _as_tensor_variable as _as_tensor_variable_aesara
+
+    @_as_tensor_variable_aesara.register(TensorVariable)
+    def raise_informative_error(*args, **kwargs):
+        raise TypeError(
+            "You are calling an Aesara function with PyTensor variables. "
```
twiecki: Should we add more information here, for people who might not have read the announcement?

ricardoV94: Sure, feel free to add a suggestion as review / push to the branch

pymc/pytensorf.py
```diff
@@ -92,6 +92,22 @@
...
+            "Did you import aesara.tensor instead of pytensor.tensor?",
```
twiecki:
```suggestion
            "You are calling an Aesara function with PyTensor variables. "
            "Starting with PyMC 5.0, Aesara was replaced by PyTensor (see https://www.pymc.io/blog/pytensor_announcement.html). Replace your import of aesara.tensor with pytensor.tensor.",
```

ricardoV94: Maybe with a line break between the link and the fix suggestion?

twiecki (approved):

Issue #5282, opened 2021-12-22
Review and update API docs


OriolAbril:
We need to make sure all our public api is documented and that they are documented at the recommended import path. #5266 updated the distributions subsection (but I did not check there were no distributions missing) and can be used as template. 

Note: if there are functions/classes that are not (or should not be) part of the public api but you think it would be helpful (even if only to you) to have a rendered version of their docstring on the website list them here too. I am considering adding a private api subsection within the contributing section

OriolAbril:
#5367 updated import paths _and reviewed everything was present_ for the gp module.

Model, sampling, step functions, backends, logprob, bart, smc, vi... are still missing

OriolAbril:
From https://github.com/pymc-devs/pymc/issues/5308#issuecomment-1008246051. Only the `.dist` method of distributions should be documented, get_moments, logcdf... should not: I think this mostly involves a custom "distribution" template for autosummary, based on https://github.com/pymc-devs/pymc/blob/main/docs/source/_templates/autosummary/class.rst

OriolAbril:
Noticed in #5543 that pymc.draw passes the kwargs to `compile_pymc` and tries to generate a link to it in the kwargs description, but this compile_pymc is not in the `__all__` of aesaraf.py file nor included in the API docs, so no link is generated because the target where it would point to does not exist

OriolAbril:
We can also look into improving the structure, grouping the functions in the math page by type, logical, matrix operations, trigonometric... grouping distributions inside discrete or continuous also in different groups depending on the range for example of their domain, grouping related functions that are in different files (i.e. set_data is in model.py, like Potential or Deterministic and are now in model section, but they could be divided between the data and distribution sections)...

OriolAbril:
All the `logp`, `logcdf`... functions that are now functions at the top pymc namespace instead of methods are missing.

twiecki:
Let's punt this to 4.1.

cluhmann:
[...]

PR #6374, opened 2022-12-07
Add helper to compute `log_likelihood` and stop computing it by default

### LINKED ISSUE #6266: Do not compute model `log_likelihood` by default at the end of sampling
This evaluation can be quite expensive and require too much memory. I think it's safe to say most users don't make use of this information (nor most of our tests where we choose to return inferencedata), making it pretty wasteful.

It's also trivial to provide an API to obtain it after the fact:

```python
with model:
  idata = pm.compute_log_likelihood(idata, extend_inferencedata=True)
```

My suggestion is to add that API, and make the automatic computation at the end of sampling opt-in rather than opt-out as it is now, if a user already knows they will want it.

```python
with model:
  idata = pm.sample(idata_kwargs={"log_likelihood": True})  # Otherwise it's not computed by default
```

I would even consider deprecating that `idata_kwargs` option to keep `pm.sample` a bit more simple, but no strong opinion there

CC @OriolAbril

ricardoV94:
Closes #6266 

Not sure about the function placement...

## Major / Breaking Changes
- `sample` no longer returns the `log_likelihood` group by default. Explicitly set `idata_kwargs={"log_likelihood": True}` or call the helper `compute_log_likelihood(idata)` after sampling. This is needed when using `arviz.compare`


pymc/stats/model_comparison.py
```diff
@@ -0,0 +1,126 @@
...
+import pymc
+
+from pymc.backends.arviz import _DefaultTrace
+from pymc.model import Model, modelcontext
+from pymc.util import dataset_to_point_list
+
+__all__ = ("compute_log_likelihood",)
+
+
+def compute_log_likelihood(
```
ricardoV94: TODO: Show in docs

pymc/stats/model_comparison.py
```diff
@@ -0,0 +1,126 @@
...
+    extend_inferencedata: bool = True,
+    model: Optional[Model] = None,
+    sample_dims: Sequence[str] = ("chain", "draw"),
+    progressbar=True,
+):
+    """Compute elemwise log_likelihood of model given InferenceData with posterior group
+
+    Parameters
+    ----------
+    idata: InferenceData
```
OriolAbril:
```suggestion
    idata : InferenceData
```

colons need a space both before and after

pymc/tests/backends/test_arviz.py
```diff
@@ -63,7 +63,12 @@ def data(self, eight_schools_params, draws, chains):
                 sigma=eight_schools_params["sigma"],
                 observed=eight_schools_params["y"],
             )
-            trace = pm.sample(draws, chains=chains, return_inferencedata=False)
+            trace = pm.sample(
+                draws,
+                chains=chains,
+                return_inferencedata=False,
+                idata_kwargs=dict(log_likelihood=True),
```
OriolAbril: I think this will be ignored here as we use `return_inferencedata=False`

pymc/stats/model_comparison.py
```diff
@@ -0,0 +1,126 @@
...
+
+    idata_loglike = InferenceData(
+        log_likelihood=dict_to_dataset(
+            loglike_trace,
+            library=pymc,
+            dims=posterior.dims,
+            coords=posterior.coords,
+            default_dims=list(sample_dims),
+        )
+    )
```
OriolAbril: Main most important comment: I think this won't preserve the coordinates and dimensions which would be an undesired backward change. I am not even sure it will work, `posterior.dims` is not a dict with variable names as keys and dimensions of those variables as values, and if it were it would not have the variable names in log_likelihood.

---

Extra comment, not sure if it might be helpful. It would be very similar to do:

```suggestion
    log_likelihood=dict_to_dataset(
        loglike_trace,
        library=pymc,
        dims=posterior.dims,
        coords=posterior.coords,
        default_dims=list(sample_dims),
    )
```

Then return the dataset (very close to returning a dict with the pointwise log likelihood values) if extend=False, otherwise use [add_groups](https://python.arviz.org/en/stable/api/generated/arviz.InferenceData.add_groups.html). I am not sure what are the use cases envisioned with `extend=False` so I don't really see pros and cons of each option.

OriolAbril (changes requested):
I would not call the file model comparison. It is not doing any model comparison, it is only an intermediate result that can be used for that, and it is also not specific to model comparison, it is also used in loo_pit which is model checking similar to plot_ppc.

To add to docs, list is somewhere in https://github.com/pymc-devs/pymc/tree/main/docs/source/api. It doesn't really matter if it is "the wrong place" as #5282 is still open and the API docs need significant work.

Most important thing is the comment on dims and coords. If they are indeed preserved, we should add a test for that, if they are not we should fix it and add a test for that.

pymc/stats/model_comparison.py
ricardoV94: Sounds cleaner. I'll check about the dims

pymc/stats/model_comparison.py
ricardoV94: My IDE populates the docstrings automatically without the space before

@MarcoGorelli Any chance we can add this to pre-commit? It's so unintuitive :)

MarcoGorelli: hey 👋 yeah I think pydocstyle might catch this?

ricardoV94: > hey wave yeah I think pydocstyle might catch this?

Thanks! @OriolAbril Do you think it would be feasible?

OriolAbril: Any help on this will be welcome

pymc/stats/model_comparison.py
ricardoV94: Fixed

OriolAbril (approved):
Looks good, thanks!

PR #6387, opened 2022-12-12
Make VI (posterior) mean and std accessible as a structured xarray 

fonnesbeck:
**What is this PR about?**

Moved from #6086 so that merge conflicts could be fixed. See original PR for details.


OriolAbril (approved):

ricardoV94 (changes requested):
The variational tests were moved into tests/variational. I am not sure where this one should go, perhaps @Armavica can weigh in?

Armavica:
I think `test_fit_data` makes sense in `tests/variational/test_inference.py`, as for `test_fit_data_coords` and its fixtures I think it would be fine too…

twiecki:
@fonnesbeck Any update on this?

fonnesbeck:
OK, tests moved over.

fonnesbeck:
@ricardoV94 are you happy with this?

pymc/variational/opvi.py
```diff
@@ -1106,16 +1107,47 @@ def __str__(self):
...
+    def var_to_data(self, shared):
+        """Takes a flat 1-dimensional Aesara variable and maps it to an xarray data set based on the information in
+        `self.ordering`.
+        """
+        # This is somewhat similar to `DictToArrayBijection.rmap`, which doesn't work here since we don't have
+        # `RaveledVars` and need to take the information from `self.ordering` instead
+        shared = shared.eval()
+        result = dict()
+        for name, s, shape, dtype in self.ordering.values():
+            dims = self.model.RV_dims.get(name, None)
```
ricardoV94: We deprecated `Model.RV_dims`, should use `Model.named_vars_to_dims`

ricardoV94 (changes requested):
There is a deprecated method and the use of the `|` operator for dicts which is not supported in our oldest Python dependency (test is failing)

Otherwise LGTM

pymc/variational/opvi.py
```diff
@@ -977,7 +978,7 @@ def symbolic_random(self):
 
     @pytensor.config.change_flags(compute_test_value="off")
     def set_size_and_deterministic(
-        self, node: Variable, s, d: bool, more_replacements: dict | None = None
+        self, node: Variable, s, d: bool, more_replacements: dict = None
```
ricardoV94, line 981:
```suggestion
        self, node: Variable, s, d: bool, more_replacements: Optional[dict] = None
```

pymc/tests/variational/test_inference.py
```diff
@@ -350,3 +350,81 @@ def test_clear_cache():
...
+
+
+@pytest.fixture
+def hierarchical_model_data():
+    group_coords = {
+        "group_d1": np.arange(3),
+        "group_d2": np.arange(7),
+    }
+    group_shape = tuple(len(d) for d in group_coords.values())
+    data_coords = {"data_d": np.arange(11)} | group_coords
```
ricardoV94: Still a `|` between dicts that is making the Windows test fail

pymc/variational/opvi.py
fonnesbeck: Unfortunately, the pre-commit hook reverts this back to `dict | None = None`. A little too agressive!

ricardoV94: You can commit with --no-verify to skip the pre-commit

ricardoV94 (approved):

pymc/variational/opvi.py
```diff
@@ -1106,16 +1107,47 @@ def __str__(self):
...
 
     @node_property
     def cov(self):
+        """Covariance between the latent variables as an unstructured 2-dimensional Aesara variable"""
         raise NotImplementedError
 
     @node_property
     def mean(self):
+        """Mean of the latent variables as an unstructured 1-dimensional Aesara variable"""
         raise NotImplementedError
```
michaelosthege:
```suggestion
    @node_property
    def std(self) -> at.TensorVariable:
        """Standard deviation of the latent variables as an unstructured 1-dimensional tensor variable"""
        raise NotImplementedError()

    @node_property
    def cov(self) -> at.TensorVariable:
        """Covariance between the latent variables as an unstructured 2-dimensional tensor variable"""
        raise NotImplementedError()

    @node_property
    def mean(self) -> at.TensorVariable:
        """Mean of the latent variables as an unstructured 1-dimensional tensor variable"""
        raise NotImplementedError()
```

pymc/variational/opvi.py
```diff
@@ -1106,16 +1107,47 @@ def __str__(self):
...
 
+    def var_to_data(self, shared):
+        """Takes a flat 1-dimensional Aesara variable and maps it to an xarray data set based on the information in
```
michaelosthege:
```suggestion
    def var_to_data(self, shared: at.TensorVariable) -> xarray.Dataset:
        """Takes a flat 1-dimensional tensor variable and maps it to an xarray data set based on the information in
```

pymc/variational/opvi.py
```diff
@@ -1106,16 +1107,47 @@ def __str__(self):
...
+        # `RaveledVars` and need to take the information from `self.ordering` instead
+        shared = shared.eval()
+        result = dict()
+        for name, s, shape, dtype in self.ordering.values():
+            dims = self.model.named_vars_to_dims.get(name, None)
+            if dims is not None:
+                coords = {d: np.array(self.model.coords[d]) for d in dims}
+            else:
+                coords = None
+            values = np.array(shared[s]).reshape(shape).astype(dtype)
```
michaelosthege: mypy doesn't like it when variables are reassigned with another type

Additionally, since `shared.eval()` returns a `np.ndarray` already, there's no need to `np.array(.)` it again
```suggestion
        shared_nda = shared.eval()
        result = dict()
        for name, s, shape, dtype in self.ordering.values():
            dims = self.model.named_vars_to_dims.get(name, None)
            if dims is not None:
                coords = {d: np.array(self.model.coords[d]) for d in dims}
            else:
                coords = None
            values = shared_nda[s].reshape(shape).astype(dtype)
```

pymc/variational/opvi.py
```diff
@@ -1106,16 +1107,47 @@ def __str__(self):
...
+            result[name] = xarray.DataArray(values, coords=coords, dims=dims, name=name)
+        return xarray.Dataset(result)
+
+    @property
+    def mean_data(self):
+        """Mean of the latent variables as an xarray Dataset"""
+        return self.var_to_data(self.mean)
+
+    @property
+    def std_data(self):
```
michaelosthege:
```suggestion
    def mean_data(self) -> xarray.Dataset:
        """Mean of the latent variables as an xarray Dataset"""
        return self.var_to_data(self.mean)

    @property
    def std_data(self) -> xarray.Dataset:
```

PR #6388, opened 2022-12-12
added postprocessing_chunks option to sample_blackjax_nuts and sample…

wnorcbrown:
**What is this PR about?**
Fix for the memory spike that happens at the end of jax sampling see: https://discourse.pymc.io/t/out-of-memory-when-transforming-variables-in-numpyro-jax/9904.

Uses experimental features from jax see: https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.maps.xmap.html.

Tested using `pytest pymc/tests/sampling/test_jax.py::test_transform_samples`
...

## Major / Breaking Changes
- None

## Bugfixes / New features
- Added a kwarg `postprocessing_chunks` to `sample_blackjax_nuts` and `sample_numpyro_nuts` so that the user can specify the chunking they want to use in postprocessing to reduce memory usage

## Docs / Maintenance

### DIFF
--- a/pymc/sampling/jax.py
+++ b/pymc/sampling/jax.py
@@ -338,9 +338,10 @@ def sample_blackjax_nuts(
         "vectorized".
     postprocessing_backend : str, optional
         Specify how postprocessing should be computed. gpu or cpu
-    postprocessing_chunks: int, default 1
+    postprocessing_chunks: Optional[int], default None
         Specify the number of chunks the postprocessing should be computed in. More
-        chunks reduces memory usage at the cost of losing some vectorization
+        chunks reduces memory usage at the cost of losing some vectorization, None
+        uses jax.vmap
     idata_kwargs : dict, optional
         Keyword arguments for :func:`arviz.from_dict`. It also accepts a boolean as
         value for the ``log_likelihood`` key to indicate that the pointwise log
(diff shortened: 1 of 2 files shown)


pymc/sampling/jax.py
```diff
@@ -142,6 +143,22 @@ def _sample_stats_to_xarray(posterior):
     return data
 
 
+def _postprocess_samples(
+    jax_fn: List[TensorVariable],
+    raw_mcmc_samples: List[TensorVariable],
+    postprocessing_backend: str,
+    num_chunks: int = 1,
```
junpenglao: (1) I think we should add a branch here to allow user to use `vmap(vmap(...))`, maybe even with `num_check = None`.
(2) Is `num_chunks = 1` a good default? I have no experience but I see in your snippet on discourse you are using `10`.

wnorcbrown: I can make the argument `Optional` with `None` as the default and use `vmap` in this case?

I think it would be cleaner to only introduce one extra argument, because vmap with a `num_chunks` argument wouldn't make sense

wnorcbrown: I chose 1 as the default argument because its behaviour should be very similar to vmap.

thomasaarholt:
It was a nice experience to have a ram issue, come here, and find that a fix is already proposed!

Can the same solution be used to decrease the ram usage of `_get_log_likelihood`, so that we can also calculate that using jax?

pymc/sampling/jax.py
```diff
@@ -313,6 +338,9 @@ def sample_blackjax_nuts(
         "vectorized".
     postprocessing_backend : str, optional
         Specify how postprocessing should be computed. gpu or cpu
+    postprocessing_chunks: int, default 1
```
junpenglao: Default is now `None`. Please update the docstring that when unspecified use `vmap`.

junpenglao: Same below

junpenglao (approved):

twiecki:
Thanks so much @wnorcbrown!

wnorcbrown:
no problem!

PR #6392, opened 2022-12-13
Update Release Notes template

ricardoV94:
I think this makes the release notes a bit more readable.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -9,16 +9,15 @@ changelog:
     - title: Major Changes 🛠
       labels:
         - major
-    - title: New Features & Bugfixes 🎉
+    - title: New Features 🎉
       labels:
-        - defects
         - enhancements
-        - feature-request
-    - title: Docs & Maintenance 🔧
+        - feature request
+    - title: Bugfixes 🪲
       labels:
+        - bug
+    - title: Documentation 📖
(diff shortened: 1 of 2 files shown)


.github/release.yml
```diff
@@ -11,14 +11,17 @@ changelog:
         - major
     - title: New Features & Bugfixes 🎉
       labels:
-        - defects
+        - bug
         - enhancements
-        - feature-request
-    - title: Docs & Maintenance 🔧
+        - feature request
```
ricardoV94, line 15: The label has a white space, not sure if the `-` replaces that in yml or not?

.github/release.yml
```diff
@@ -11,14 +11,17 @@ changelog:
...
+    - title: Documentation 📖
       labels:
         - docs
+    - title: Maintenance 🔧
+      labels:
         - installation
         - maintenance
         - pre-commit
         - tests
+        - release
```
ricardoV94: Doesn't matter because everything will be captured here, but maybe useful if we ever refactor this?

.github/release.yml
michaelosthege: The space is probably the right way to do it

michaelosthege (approved):

PR #6396, opened 2022-12-14
Remove global RandomStream

ricardoV94:
No longer needed after #6304

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -29,7 +29,6 @@ from pytensor.compile.sharedvalue import SharedVariable
 from pytensor.raise_op import Assert
 from pytensor.scalar import Cast
 from pytensor.tensor.elemwise import Elemwise
-from pytensor.tensor.random import RandomStream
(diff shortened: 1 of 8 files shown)


pymc/data.py
```diff
@@ -132,6 +131,12 @@ def __hash__(self):
 class MinibatchIndexRV(IntegersRV):
     _print_name = ("minibatch_index", r"\operatorname{minibatch\_index}")
 
+    # Work-around for https://github.com/pymc-devs/pytensor/issues/97
+    def make_node(self, rng, *args, **kwargs):
+        if rng is None:
+            rng = pytensor.shared(np.random.default_rng())
+        return super().make_node(rng, *args, **kwargs)
```
ricardoV94, line 138: Can be removed once https://github.com/pymc-devs/pytensor/issues/97 is solved

ferrine (approved):

pymc/data.py
ferrine: sure thing

PR #6397, opened 2022-12-14
Update PyTensor dependency and fix bugs in inferred mixture logprob

ricardoV94:
Will need to wait for the conda-forge release


twiecki (approved):

ricardoV94:
I had to patch quite some things with the inferred mixture logprob. Most of these cases won't be supported after #6369 but we haven't merged that yet

twiecki:
Why don't we?

ricardoV94:
> Why don't we?

Merge? The fixes done here are actually more correct

PR #6400, opened 2022-12-16
Extend logprob inference to power transforms

ricardoV94:
This PR extends logprob inference for power transforms. It replaces the Reciprocal transform, which is just a special case.

```python
import numpy as np
import scipy as sp
import pymc as pm

# A squared normal is equivalent to a chi-square with 1 df
x = pm.Normal.dist(0, 1, shape=(3,)) ** 2

test_val = np.r_[0.5, 1, 2.5]
assert np.allclose(
    pm.logp(x, test_val).eval(),
    sp.stats.chi2(df=1).logpdf(test_val),
)
```

Slightly unsure about the back-transform providing multiple values. It seems this is the correct way to derive the logp. Is it the only one?

## New features
- Extend logprob inference to power transforms

### DIFF
--- a/pymc/tests/logprob/test_transforms.py
+++ b/pymc/tests/logprob/test_transforms.py
@@ -224,7 +224,7 @@ def test_transformed_logprob(at_dist, dist_params, sp_dist, size):
 
     a = at_dist(*dist_params, size=size)
     a.name = "a"
-    a_value_var = at.tensor(a.dtype, shape=(None,) * a.ndim)
+    a_value_var = at.tensor(dtype=a.dtype, shape=(None,) * a.ndim)
     a_value_var.name = "a_value"
 
     b = at.random.normal(a, 1.0)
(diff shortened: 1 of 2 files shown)


pymc/tests/logprob/test_transforms.py
```diff
@@ -807,6 +807,36 @@ def test_reciprocal_rv_transform(numerator):
...
 
+def test_sqr_transform():
+    # The square of a unit normal is a chi-square with 1 df
+    x_rv = at.random.normal(0, 1, size=(3,)) ** 2
+    x_rv.name = "x"
+
+    x_vv = x_rv.clone()
+    x_logp_fn = pytensor.function([x_vv], joint_logprob({x_rv: x_vv}, sum=False))
+
+    x_test_val = np.r_[0.5, 1, 2.5]
```
michaelosthege, line 818: Showing off your NumPy skillz, eh?

pymc/tests/logprob/test_transforms.py
```diff
@@ -807,6 +807,36 @@ def test_reciprocal_rv_transform(numerator):
...
+    assert np.allclose(
+        x_logp_fn(x_test_val),
+        sp.stats.chi2(df=1).logpdf(x_test_val),
+    )
```
michaelosthege, line 822: It's not obvious from the test alone, but the `joint_logprob` here only works because the `measurable_sqrt_sqr_to_power` rewrite and `PowerTransform(RVTransform)` were implemented, right?

Maybe you can add this info in the docstring and assert that the `x_logp_fn` has the rewritten graph structure?

(Same comment applies to `test_sqrt_transform`.)

ricardoV94: It shouldn't matter how the logp was achieved only that we got it and it's correct.

michaelosthege (approved):

Issue #6402, opened 2022-12-18
Infer logprob of absolute operations 


ricardoV94:
### Description

https://github.com/pymc-devs/pymc/pull/6400 extended the logprob inference to 2-to-1 transforms. It's now straightforward to add an Absolute transform to infer the logprob of cases like

```python
import pymc as pm

x = pm.math.abs(pm.Normal.dist())
y = pm.HalfNormal.dist()
assert pm.logp(x, 2.5).eval() == pm.logp(y, 2.5).eval()
```

LukeLB:
I'm taking a look at this, but just wanted to check my understanding...

So similar to what was done in #6400, there needs to be an `AbsoluteTransform()` class created and another if statement added to `find_measurable_transforms()` to check for whether the Elemwise operator is an absolute transform and then set the transform variable to the created `AbsoluteTransform()`. 

Am I missing anything?

ricardoV94:
@LukeLB that's it!

LukeLB:
I'll get to it then!

ricardoV94:
Closed via #6414

PR #6414, opened 2022-12-27
 Infer logprob of absolute operations and fix logprob of powers with negative values

LukeLB:
**What is this PR about?**
Following from #6400 and implements #6402. This PR allows the use of an absolute transform for cases like
````python
import pymc as pm

x = pm.math.abs(pm.Normal.dist())
y = pm.HalfNormal.dist()
assert pm.logp(x, 2.5).eval() == pm.logp(y, 2.5).eval()
````
I have included the above example as a test.

Some notes:
- I wasn't completely sure on how to derive the Jacobian determinant for an absolute function, I think its just the derivative wrt x, so that's what I've included. 
- Should we expect x and y (in the example above) to return the same value if the test_value is negative? e.g.
````python
import pymc as pm

x = pm.math.abs(pm.Normal.dist())
y = pm.HalfNormal.dist()
assert pm.logp(x, -2.5).eval() == pm.logp(y, -2.5).eval()
````
Because in this case, `x` returns `-inf` while `y` returns `nan`

**Checklist**

## New features
- The ability to evaluate the log prob of an absolute transformed random variable

### DIFF
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -374,9 +374,14 @@ def measurable_transform_logprob(op: MeasurableTransform, values, *inputs, **kwa
     else:
         input_logprob = logprob(measurable_input, backward_value)
 
+    if input_logprob.ndim < value.ndim:
+        # Do we just need to sum the jacobian terms across the support dims?
+        raise NotImplementedError("Transform of multivariate RVs not implemented")
+
     jacobian = op.transform_elemwise.log_jac_det(value, *other_inputs)
 
-    return input_logprob + jacobian
+    # The jacobian is used to ensure a value in the supported domain was provided
+    return at.switch(at.isnan(jacobian), -np.inf, input_logprob + jacobian)
 
 
 @node_rewriter([reciprocal])
@@ -711,18 +716,32 @@ class PowerTransform(RVTransform):
         at.power(value, self.power)
 
     def backward(self, value, *inputs):
-        backward_value = at.power(value, (1 / self.power))
+        inv_power = 1 / self.power
+
+        # Powers that don't admit negative values
+        if (np.abs(self.power) < 1) or (self.power % 2 == 0):
+            backward_value = at.switch(value >= 0, at.power(value, inv_power), np.nan)
+        # Powers that admit negative values require special logic, because (-1)**(1/3) returns `nan` in PyTensor
+        else:
+            backward_value = at.power(at.abs(value), inv_power) * at.switch(value >= 0, 1, -1)
 
         # In this case the transform is not 1-to-1
-        if (self.power > 1) and (self.power % 2 == 0):
+        if self.power % 2 == 0:
             return -backward_value, backward_value
         else:
             return backward_value
 
     def log_jac_det(self, value, *inputs):
         inv_power = 1 / self.power
+
         # Note: This fails for value==0
-        return np.log(np.abs(inv_power)) + (inv_power - 1) * at.log(value)
+        res = np.log(np.abs(inv_power)) + (inv_power - 1) * at.log(at.abs(value))
+
+        # Powers that don't admit negative values
+        if (np.abs(self.power) < 1) or (self.power % 2 == 0):
+            res = at.switch(value >= 0, res, np.nan)
+
+        return res
 
 
 class IntervalTransform(RVTransform):
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -49,10 +49,11 @@ from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.op import Op
 from pytensor.graph.replace import clone_replace
 from pytensor.graph.rewriting.basic import GraphRewriter, in2out, node_rewriter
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pymc/logprob/transforms.py [hunk 1]
```diff
@@ -696,6 +699,20 @@ def log_jac_det(self, value, *inputs):
...
+
+    def forward(self, value, *inputs):
+        return at.abs(value)
+
+    def backward(self, value, *inputs):
+        return -value, value
+
+    def log_jac_det(self, value, *inputs):
+        # fails for value == 0
+        return at.log(value / at.abs(value))
```
ricardoV94: This should suffice

```suggestion
        return at.zeros_like(value)
```

ricardoV94:
> Should we expect x and y (in the example above) to return the same value if the test_value is negative? e.g.

Hmm, it should fail for negative values... Not sure what's the best way to encode that. Maybe `value = at.switch(value >= 0, value, np.nan)`?

LukeLB:
> Thanks for opening a PR!
> 
> > Should we expect x and y (in the example above) to return the same value if the test_value is negative? e.g.
> 
> Hmm, it should behave the same for negative values... Not sure what's the best way to encode that.
> 
> What happens with a `sqrt` transform? Is the jacobian what leads to `-inf`?
> 
> In that case the jacobian could perhaps be:
> 
> `return at.switch(value >= 0, 0, -np.inf)`?
> 
> But the logp will be broken in both cases without jacobian.
> 
> Perhaps we need to implement a specific logprob for the absolute and power transforms
> 
> Regardless of the solution, we should add test for those cases, hadn't thought of them

Applogies I have gotten this the wrong way round... in the example:
````python
import pymc as pm
x = pm.math.abs(pm.Normal.dist())
y = pm.HalfNormal.dist() 
z = pm.Normal.dist() ** 0.5
print(pm.logp(x, -2.5).eval(), pm.logp(y, -2.5).eval(), pm.logp(z, -2.5).eval())
````
x = nan, y = -inf, z = nan. So the transforms return nan values.

**Comment by ricardoV94** [comment]
Yup, `-inf` would be preferable

File: pymc/logprob/transforms.py [hunk 2]
```diff
@@ -722,7 +738,9 @@ def backward(self, value, *inputs):
     def log_jac_det(self, value, *inputs):
         inv_power = 1 / self.power
         # Note: This fails for value==0
-        return np.log(np.abs(inv_power)) + (inv_power - 1) * at.log(value)
+        return at.switch(
```
ricardoV94: Have to think about this, odd powers should work with negative numbers, no?

The previous implementation returned -inf as well I think

LukeLB: I see so would adding another check for the power work?
````python
return at.switch(
            (value >= 0) or (self.power % 2 != 0), np.log(np.abs(inv_power)) + (inv_power - 1) * at.log(value), -np.inf
        )
````

ricardoV94: We don't need a switch for the power. We only support constant powers, so you can use a simple python if else, like the backward does.

ricardoV94: Perhaps you need to also use  `at.log(at.abs(value))` instead of `at.log(value)`?

Also, I think you have to use `at.or_`, the Python `or` won't work.

LukeLB:
````python
return at.switch(
    at.or_(value >= 0,  self.power % 2 != 0), np.log(np.abs(inv_power)) + (inv_power - 1) * at.log(at.abs(value)), -np.inf
                 )
````
I'm running into problems with the above and its throwing:
`NotImplementedError: Logprob method not implemented for Elemwise{pow,no_inplace}`
I think its because `value >= 0` evaluates to `Elemwise{ge,no_inplace}.0` while  `self.power % 2 != 0` evaluates to `True`. 

Not sure on what the fix is here... any suggestions?

LukeLB: Figured it out!

LukeLB: Had to change the `backward` function in `PowerTransform` to allow for negative values, but everything seems to work as intended now.

ricardoV94:
Hi @LukeLB 

I wanted to leave a review but ended up having to try many things that it was easier to push my suggestions as a separate commit. 

I realized the issue you found was not specific to power transforms, but any transform that is constrained. For instance the `exp` transform would also return `nan` for a negative value. I tweaked the logprob function to look for `nan` in the jacobian and return a -inf logp in that case. Since the jacobian depends on the values, this should be generally sufficient?

I also tweaked some of your tests and tried to simplify the PowerTransform logic. Let me know what you think, and if you agree with the changes I will clean a bit the commits before merging.

Thanks a lot for the work so far!

ricardoV94:
I'll just clean a bit the commits. I don't think anything else is needed, just wanted your input :)

LukeLB:
Awesome! Thank you for helping me through this PR!

ricardoV94 (approved):
Cleaned up the commits, didn't change any content

ricardoV94:
Thanks @LukeLB! This one required a bit of digging because it wasn't really working from the previous PRs, nice finding!

PR #6422, opened 2023-01-01
Access to external nuts samplers via  `sample`

twiecki:
TODO:

- [x] Tests
- [x] Figure out circular dependency on nutpie test...

### Major
- All arguments except `draws` in `sample` are now keyword-only.

### Enhancements
- Access to external nuts samplers via `sample`

### DIFF
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -221,6 +221,94 @@ def all_continuous(vars):
         return True
 
 
+def _sample_external_nuts(
+    sampler: str,
+    draws: int,
+    tune: int,
+    chains: int,
+    target_accept: float,
+    random_seed: Union[RandomState, None],
+    initvals: Union[StartDict, Sequence[Optional[StartDict]], None],
+    model: Model,
+    progressbar: bool,
+    idata_kwargs: Optional[Dict],
+    **kwargs,
+):
+    warnings.warn("Use of external NUTS sampler is still experimental", UserWarning)
+
+    if sampler == "nutpie":
+        try:
+            import nutpie
+        except ImportError as err:
+            raise ImportError(
+                "nutpie not found. Install it with conda install -c conda-forge nutpie"
+            ) from err
+
+        if initvals is not None:
+            warnings.warn(
+                "`initvals` are currently not passed to nutpie sampler. "
+                "Use `init_mean` kwarg following nutpie specification instead.",
+                UserWarning,
+            )
+
+        if idata_kwargs is not None:
+            warnings.warn(
+                "`idata_kwargs` are currently ignored by the nutpie sampler",
+                UserWarning,
+            )
+
+        compiled_model = nutpie.compile_pymc_model(model)
+        idata = nutpie.sample(
+            compiled_model,
+            draws=draws,
+            tune=tune,
+            chains=chains,
+            target_accept=target_accept,
+            seed=_get_seeds_per_chain(random_seed, 1)[0],
+            progress_bar=progressbar,
+            **kwargs,
+        )
+        return idata
+
+    elif sampler == "numpyro":
+        import pymc.sampling.jax as pymc_jax
+
+        idata = pymc_jax.sample_numpyro_nuts(
+            draws=draws,
+            tune=tune,
+            chains=chains,
+            target_accept=target_accept,
+            random_seed=random_seed,
+            initvals=initvals,
+            model=model,
+            progressbar=progressbar,
+            idata_kwargs=idata_kwargs,
+            **kwargs,
+        )
+        return idata
+
+    elif sampler == "blackjax":
+        import pymc.sampling.jax as pymc_jax
+
+        idata = pymc_jax.sample_blackjax_nuts(
+            draws=draws,
+            tune=tune,
+            chains=chains,
+            target_accept=target_accept,
+            random_seed=random_seed,
+            initvals=initvals,
+            model=model,
+            idata_kwargs=idata_kwargs,
+            **kwargs,
+        )
+        return idata
+
+    else:
+        raise ValueError(
+            f"Sampler {sampler} not found. Choose one of ['nutpie', 'numpyro', 'blackjax', 'pymc']."
+        )
+
+
 def sample(
     draws: int = 1000,
     step=None,
@@ -239,6 +327,7 @@ def sample(
     callback=None,
     jitter_max_retries: int = 10,
     *,
+    nuts_sampler: str = "pymc",
     return_inferencedata: bool = True,
     keep_warning_stat: bool = False,
     idata_kwargs: dict = None,
@@ -257,6 +346,7 @@ def sample(
     init : str
         Initialization method to use for auto-assigned NUTS samplers. See `pm.init_nuts` for a list
         of all options. This argument is ignored when manually passing the NUTS step method.
+        Only applicable to the pymc nuts sampler.
     step : function or iterable of functions
         A step function or collection of functions. If there are variables without step methods,
         step methods for those variables will be assigned automatically. By default the NUTS step
@@ -306,6 +396,10 @@ def sample(
         Maximum number of repeated attempts (per chain) at creating an initial matrix with uniform
         jitter that yields a finite probability. This applies to ``jitter+adapt_diag`` and
         ``jitter+adapt_full`` init methods.
+    nuts_sampler : str
+        Which NUTS implementation to run. One of ["pymc", "nutpie", "blackjax", "numpyro"].
+        This requires the chosen sampler to be installed.
+        All samplers, except "pymc", require the full model to be continuous.
     return_inferencedata : bool
         Whether to return the trace as an :class:`arviz:arviz.InferenceData` (True) object or a
         `MultiTrace` (False). Defaults to `True`.
@@ -401,7 +495,7 @@ def sample(
         if "nuts" in kwargs:
             kwargs["nuts"]["target_accept"] = kwargs.pop("target_accept")
         else:
-            kwargs = {"nuts": {"target_accept": kwargs.pop("target_accept")}}
+            kwargs["nuts"] = {"target_accept": kwargs.pop("target_accept")}
     if isinstance(trace, list):
         raise DeprecationWarning(
             "We have removed support for partial traces because it simplified things."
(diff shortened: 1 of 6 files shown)


pymc/sampling/mcmc.py
```diff
@@ -212,9 +212,81 @@ def all_continuous(vars):
...
+            save_warmup=~discard_tuned_samples,
+        )
+        return idata
+    except ImportError:
+        _log.info("nutpie not found. Install it with mamba install -c conda-forge nutpie")
+    except Exception as e:
+        _log.info(e)
+    try:
+        _log.info("Trying NumPyro JAX sampler...")
+        import pymc.sampling.jax as pymc_jax
```
ricardoV94, line 274: Only the import should be in the try block, sampling exceptions should raise naturally, no?

twiecki: I wanted to hedge a little bit in case the user is e.g. running into numba ops that do not exist yet, we want to fall back to C+Python.

ricardoV94: What about just accepting the sampler as a string and if it fails the user can try the next one?

`nuts_sampler : ["nutpie", "numpyro", "blackjax", "pymc"] = "pymc"`

twiecki: yeah, that was my first implementation. Reconsidering, that might be better.

twiecki: What I don't like about this is that we can't easily switch it on as a default and have a good fallback.

ricardoV94: When we think the time is ready we can start announcing that the default will be nutpie and tell people to explicit set it to "pymc" if they don't want it to change.

I think we will be able to do the switch without to much pain that we don't need to worry about a runtime fallback.

On a separate note, it would be nice to include a sentence about reporting issues to us when attempting to use the nutpie sampler

twiecki: Good point, with conda-forge installation should be fine.

ferrine:
So far nutpie doesn’t provide precompiled Python wheels. So users that pop install pymc are unable to use this sampler. https://github.com/pymc-devs/nutpie/issues/22

ricardoV94:
> So far nutpie doesn’t provide precompiled Python wheels. So users that pip install pymc are unable to use this sampler. https://github.com/pymc-devs/nutpie/issues/22 

That's okay, we're not making the default here. There's a try except import with an informative message

fonnesbeck:
nutpie can currently only be installed via maturin, correct?

pymc/sampling/mcmc.py
fonnesbeck: I like the string argument, particularly since the `fit` function already uses a `method` argument to specify the algorithm.

twiecki:
> nutpie can currently only be installed via maturin, correct?

It's also available via conda-forge.

pymc/sampling/mcmc.py
twiecki: OK, implemented the change.

pymc/sampling/mcmc.py
```diff
@@ -212,9 +212,79 @@ def all_continuous(vars):
...
+def sample_external_nuts(
+    sampler="nutpie",
+    draws: int = 1000,
+    initvals: Optional[Union[StartDict, Sequence[Optional[StartDict]]]] = None,
+    chains: Optional[int] = None,
+    cores: Optional[int] = None,
+    tune: int = 1000,
+    progressbar: bool = True,
+    model=None,
+    random_seed: RandomState = None,
```
ricardoV94: This is wrong, as nutpie only accepts a number IIRC, and from the main sample call your are passing a list of seeds already.

You can do `get_seeds_per_chain(random_seed, 1)[0]` or whatever it's called when passing the seed to nutpie.

pymc/sampling/mcmc.py
```diff
@@ -212,9 +212,79 @@ def all_continuous(vars):
...
+):
+    target_accept = kwargs.get("nuts", {}).get("target_accept", 0.8)
+    # Try different fast samplers in order of their speed
+    if sampler == "nutpie":
+        try:
+            import nutpie
+        except ImportError:
+            _log.info("nutpie not found. Install it with mamba install -c conda-forge nutpie")
+
+        if (draws == 1000) and (tune == 1000):
```
ricardoV94: Is this important? If so you should use None as the default argument for these variables, otherwise you will be ignoring when users explicitly request 1000 draws and tune.

pymc/sampling/mcmc.py
twiecki: Thanks. Can't find `get_seeds_per_chain`.

pymc/sampling/mcmc.py
twiecki: But this is just forwarded for sample() so we'd need to change it there as well. Maybe we'll just drop it for now.

pymc/sampling/mcmc.py
ricardoV94: It's called `_get_seeds_per_chain` and it's already imported in this file

pymc/sampling/mcmc.py
```diff
@@ -212,9 +212,74 @@ def all_continuous(vars):
...
+    discard_tuned_samples: bool = True,
+    **kwargs,
+):
+    target_accept = kwargs.get("nuts", {}).get("target_accept", 0.8)
+    # Try different fast samplers in order of their speed
+    if sampler == "nutpie":
+        try:
+            import nutpie
+        except ImportError:
+            _log.info("nutpie not found. Install it with mamba install -c conda-forge nutpie")
```
ricardoV94: Won't a suggestion to use conda be less confusing than mamba?

pymc/sampling/mcmc.py
```diff
@@ -446,6 +511,21 @@ def sample(
     initial_points = None
     step = assign_step_methods(model, step, methods=pm.STEP_METHODS, step_kwargs=kwargs)
```
ricardoV94, line 550: We should raise if we have a non-nuts stepper and a non-pymc nuts was requested.

Also tests

twiecki:
```
[pymc/sampling/mcmc.py]
pymc/sampling/mcmc.py:253: error: Argument "chains" to "sample_numpyro_nuts" has incompatible type "Optional[int]"; expected "int"
pymc/sampling/mcmc.py:255: error: Argument "random_seed" to "sample_numpyro_nuts" has incompatible type "Union[Union[int, Sequence[int], ndarray[Any, Any], None], RandomState, Generator, None]"; expected "Union[int, Sequence[int], ndarray[Any, Any], None]"
pymc/sampling/mcmc.py:266: error: Argument "chains" to "sample_blackjax_nuts" has incompatible type "Optional[int]"; expected "int"
pymc/sampling/mcmc.py:268: error: Argument "random_seed" to "sample_blackjax_nuts" has incompatible type "Union[Union[int, Sequence[int], ndarray[Any, Any], None], RandomState, Generator, None]"; expected "Union[int, Sequence[int], ndarray[Any, Any], None]"
```

Is this important? Not very familiar with types.

ricardoV94:
Hmm a specific version of pymc is being installed by nutpie... so the test doesn't make much sense...

pymc/sampling/mcmc.py
```diff
@@ -221,9 +221,99 @@ def all_continuous(vars):
...
+    random_seed: Union[RandomState, None],
+    initvals: Union[StartDict, Sequence[Optional[StartDict]], None],
+    model: Model,
+    progressbar: bool,
+    idata_kwargs: Optional[Dict],
+    **kwargs,
+):
+    warnings.warn("Use of external NUTS sampler is still experimental", UserWarning)
+
+    # Try different fast samplers in order of their speed
```
twiecki:
```suggestion
```

twiecki:
@ricardoV94 Any progress? What's the status? Should someone else take over?

twiecki:
OK, lets just not test it then.

pymc/sampling/mcmc.py
twiecki: I'd still remove this ocmment.

twiecki:
Tests pass! Last comment but LGTM otherwise.

ricardoV94:
Tests already  passed before, they were failing for unrelated reasons :)

Still didn't remove the test dependency on nutpie

pymc/sampling/mcmc.py
```diff
@@ -221,9 +221,98 @@ def all_continuous(vars):
...
+    else:
+        raise ValueError(
+            f"Sampler {sampler} not found. Choose one of ['nutpie', 'numpyro', 'blackjax', 'pymc']."
+        )
+
+
 def sample(
     draws: int = 1000,
     step=None,
+    nuts_sampler: str = "pymc",
```
michaelosthege, line 321: we should not change the args. Just place it to be an explicit, optional kwarg

pymc/sampling/mcmc.py
```diff
@@ -503,7 +613,7 @@ def sample(
     )
 
     sample_args = {
-        "draws": draws,
+        "draws": draws + tune,  # FIXME: Why is tune added to draws?
```
michaelosthege, line 618: probably a relic of old signatures

michaelraczycki:
Have you tried using dependency injection? Because if yes, other way would be to use unittest.mock, I've used it in the past but that will depend on how complex the test is.

pymc/sampling/mcmc.py
ricardoV94: I will open an issue after this PR to investigate it

ricardoV94:
Maybe the dependency is not too bad. It first installs nutpie and then the local pymc, so I guess if PyMC depends on a newer PyTensor it will replace by that? We can merge and see what happens when we next bump the PyTensor dependency.

pymc/sampling/mcmc.py
twiecki: Good point.

ricardoV94: Should we go ahead an make `draws` the only non-keyword allowed argument? There is really no way a user can make sense of the ~15 positional arguments that exist now anyway. There is nothing intuitive about their order.

Since `step` is the second argument, I am confident nobody is using anything else without keyword atm

ricardoV94: Added a separate commit that makes everything but draws kwarg only

twiecki:
This looks great! merge?

ricardoV94:
> This looks great! merge?

Never hurts to see what happens next :D

twiecki:
Let's do a minor release!

PR #6443, opened 2023-01-08
Refactoring and addition of helpers to handle flat stats

michaelosthege:
This is one step towards refactoring of the trace backend.

PyMC supports hierarchies of samplers, which means that sampler stats are not a flat list by default.

First I moved `Competence` and `BlockedStep` to the `compound` module to avoid a circular import.

Second, I added a `StatsBijection` which can be used to flatten a list of sampler stats dicts into one dictionary.

Related: https://github.com/pymc-devs/pymc/issues/4602

**Checklist**

## Maintenance
- `Competence` and `BlockedStep` were moved to the `compound` module.
- A non-public `StatsBijection` was added.

### DIFF
--- a/pymc/step_methods/compound.py
+++ b/pymc/step_methods/compound.py
@@ -18,16 +18,15 @@ Created on Mar 7, 2011
 @author: johnsalvatier
 """
 
-
 from abc import ABC, abstractmethod
 from enum import IntEnum, unique
-from typing import Dict, List, Tuple
+from typing import Dict, List, Sequence, Tuple, Union
 
 import numpy as np
 
 from pytensor.graph.basic import Variable
 
-from pymc.blocking import PointType, StatsType
+from pymc.blocking import PointType, StatsDict, StatsType
 from pymc.model import modelcontext
 
 __all__ = ("Competence", "CompoundStep")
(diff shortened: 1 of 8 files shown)


pymc/step_methods/compound.py
```diff
@@ -164,3 +163,43 @@ def reset_tuning(self):
...
+    """Map between a `list` of stats to `dict` of stats."""
+
+    def __init__(self, sampler_stats_dtypes: Sequence[Dict[str, type]]) -> None:
+        # Keep a list of flat vs. original stat names
+        self._stat_groups: List[List[Tuple[str, str]]] = [
+            [(f"sampler_{s}__{statname}", statname) for statname, _ in names_dtypes.items()]
+            for s, names_dtypes in enumerate(sampler_stats_dtypes)
+        ]
+
+    def map(self, stats_list: StatsType) -> StatsDict:
```
Armavica, line 190: Are `map` and `rmap` dedicated or established jargon for this kind of operation? If not, what would you think of using more explicit method names such as `as_dict` (or `to_dict` or something similar) for `map`, and `as_list` for `rmap`?

michaelosthege: I wanted to align with the `DictToArrayBijection`, but otherwise I would agree

Armavica (approved):

PR #6334, opened 2022-11-24
Merge AePPL into logprob submodule

ricardoV94:
There are still some redundancies to clean, namely in `distributions/transforms` and `distributions/logprob` but this should get us 99% there.

TODO
- [x] Figure out if we want an explicit license note for this submodule

## Major / Breaking Changes
- PyMC no longer depends on Aeppl. Relevant dispatch methods can be found in `pymc.logprob.abstract`

PR #6447, opened 2023-01-09
Refactoring printing module with aeppl/printing.py

### LINKED ISSUE #6311: Do not expect model variables to always have `str_repr`
When we create RVs via the the Distribution interface, they get some `str_repr` methods glued to them:

https://github.com/pymc-devs/pymc/blob/4acd98ee2bc31ef1fb63accdfb1012ba64fcebd5/pymc/distributions/distribution.py#L302-L306

The `Model.str_repr` then expects all named variables to have this method, which fails for automatically generated RVs such as those in automatic imputation

```python
import numpy as np
import pymc as pm

with pm.Model() as m:
   pm.Normal("x", observed=[np.nan, 0, 0])
m.str_repr()  # Called automatically in a Jupyter environment for pretty-printing
```
```
AttributeError: 'TensorVariable' object has no attribute 'str_repr'
```

It also which makes it annoying to replace / register variables manually: 

```python
x = pm.Normal.dist()  # No `str_repr` attached
with pm.Model() as m:
  m.register_rv(x, name="x")
m.str_repr()  # Same error
```

## Suggestion
Do not rely on the methods being attached for `model.str_repr`

https://github.com/pymc-devs/pymc/blob/fd5a6ccbcda955197c2c57384407786e245d6f36/pymc/printing.py#L103

Instead just use a dispatch base approach to pretty-print any model variables based on their Op types as discussed elsewhere.

larryshamalama:
Closes #6311 and #6261, supersedes #5793 and part of #6072.

This PR refactors most of our printing module. Most of the contents were taken _verbatim_ from [aeppl/printing.py](https://github.com/aesara-devs/aeppl/blob/main/aeppl/printing.py) as it was not ported in #6334; I was unable to retrieve the Git history from that file and the AePPL license is available in the beginning of the printing module now.

While tensor variables are no longer expected to have a `str_repr` method attached, the main benefits are that 1) printing is done with respect to Ops (instead of seeing `f()`s as per #6261) and 2) there will be a default printing mechanism for newly created distributions due to the dispatching approach, although how good the latter will be remains unclear to me. This PR also makes use of the [printing module](https://github.com/pymc-devs/pytensor/blob/main/pytensor/printing.py) in PyTensor.

I am creating this PR "early on" to receive feedback, especially in how we want to display some newly created symbolic distributions and how the API will work. This refactoring took quite some time, so I also want to check in that this is going in the right direction as there are still many elements to work on. Tests will fail, for now at least. This will be a slow WIP.

To-do's:

- [ ] Create `CensoredPrinter`
- [ ] Adapt `CensoredPrinter` to the distribution being censored (e.g. CensoredNormal)
- [ ] Adapt `MarginalMixturePrinter` when using single component dispatching
- [ ] Refactor tests
- [ ] Tailor docstrings to new printing functionality

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -1249,9 +1249,13 @@ def _zero_inflated_mixture(*, name, nonzero_p, nonzero_dist, **kwargs):
         nonzero_dist,
     ]
     if name is not None:
-        return Mixture(name, weights, comp_dists, **kwargs)
+        out_rv = Mixture(name, weights, comp_dists, **kwargs)
     else:
-        return Mixture.dist(weights, comp_dists, **kwargs)
+        out_rv = Mixture.dist(weights, comp_dists, **kwargs)
+
+    out_rv.is_zero_inflated = True
+
+    return out_rv
 
 
 class ZeroInflatedPoisson:
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -14,7 +14,6 @@
 import contextvars
 import functools
 import sys
-import types
 import warnings
 
 from abc import ABCMeta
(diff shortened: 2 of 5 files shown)


rlouf:
👀

PR #6036, opened 2022-08-07
Add HSGPs, add support for power spectral densities to covariance function objects

bwengals:
**What is this PR about?**
Adding support for power spectral density functions to covariance functions.  Any stationary kernel may define a `psd` method that takes in `omega` as an input.

```python
class ExpQuad(Stationary):
    def __init__(self, ...):
       ...

    def full(self, X, Xs):
        ....

    # allow this method to be defined
    def psd(self, omega):
        ...
```
This will be useful for some kernel approximation algorithms like the [HSGP approximation](https://arxiv.org/pdf/2004.11408.pdf) and [random Fourier features](https://people.eecs.berkeley.edu/~brecht/papers/07.rah.rec.nips.pdf).  It should also make it easier to work with other things like [spectral mixture kernels](https://arxiv.org/pdf/1302.4245.pdf) too.

This also adds the ability for composite kernels to be constructed and their `psd` to be calculated.
```
cov = eta1**2 * pm.gp.cov.ExpQuad(1, ls=ls1) + eta2**2 * pm.gp.cov.Matern52(1, ls=ls2) 
cov.psd(omega) # works
```

It should be OK now for users to define a custom kernel that subclasses `Stationary`, but doesn't define `.full` or `.diag` methods, just `.psd`, which is all some approximations would need.  

**Checklist**
+ [ ] Explain important implementation details 👆
+ [ ] Make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).
+ [ ] Link relevant issues (preferably in [nice commit messages](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)) **No relevant issues, this is a new feature.**
+ [ ] Are the changes covered by tests and docstrings?
+ [ ] Fill out the short summary sections 👇

## Major / Breaking Changes
- Shouldn't break anything

## Bugfixes / New features
- Add `.psd` for `pm.gp.cov.ExpQuad`, `pm.gp.cov.Matern52` and `pm.gp.cov.Matern32`.
- The `.psd` method should work for any dimensional `omega`.  Meaning, 1 or 2+ dimensional processes.

## Docs / Maintenance
- Doesn't add anything major to the docs, just docstrings needed.

PR #6458, opened 2023-01-18
Add HSGP Latent GP approximation

bwengals:
This PR add's the basics required for the HSGP GP approximation.   It replaces #6036, much thanks to @ferrine for rebasing and cleaning that PR up!  This PR also modifies the covariance function classes to allow an additional method `power_spectral_density` which is needed for HSGP (or random fourier features).  Any covariance function that defines a `power_spectral_density` method will work.  Like with the other GP implementations, one can also add covariances, `eta1**2 * pm.gp.cov.ExpQuad(2, ls=ls1) + eta2**2 * pm.gp.cov.Matern52(2, ls=ls2)`.   The implementation also works for any number of input dimensions.

Not part of this PR, but plan on adding in later:
- support for `Periodic` covariance
- warnings for bad choices of `m` and `L` or `c`.
- refactoring the internals to allow advanced users to bypass the GP api and work directly with the basis and coefficients for multi-GP models 
- include more utility functions for checking the accuracy of the approximation

**Checklist**

## Major / Breaking Changes
- None

## New features
- HSGPs

## Bugfixes
- None

## Documentation
- ...

## Maintenance
- ...


ferrine (approved):
I've managed to check this PR in action and it was smooth

pymc/gp/cov.py
```diff
@@ -89,27 +88,18 @@ def __call__(self, X, Xs=None, diag=False):
...
-        return X, Xs
-
     def __add__(self, other):
+        # If it's a scalar, cast as Constant covariance.  This allows validation for power spectral
+        # density calc.
+        if isinstance(other, Number):
+            try:
+                other = Constant(c=_verify_scalar(other))
+            except ValueError:
+                pass
```
michaelosthege: The `isinstance(other, Number)` should already make this go through the corresponding `elif` branch in `_verofy_scalar` such that the `ValueError` is not raised.

pymc/gp/cov.py
```diff
@@ -151,41 +129,127 @@ def __array_wrap__(self, result):
...
+    ----------
+    input_dim: integer
+        The number of input dimensions, or columns of X (or Xs)
+        the kernel will operate on.
+    active_dims: List of integers
+        Indicate which dimension or column of X the covariance
+        function operates on.
+    """
+
+    def __init__(self, input_dim, active_dims=None):
```
michaelosthege:
```suggestion
    def __init__(self, input_dim: int, active_dims: Optional[Sequence[int]]=None):
```

pymc/gp/cov.py
```diff
@@ -194,23 +258,79 @@ def merge_factors(self, X, Xs=None, diag=False):
...
+
         return factor_list
 
 
 class Add(Combination):
     def __call__(self, X, Xs=None, diag=False):
-        return reduce(add, self.merge_factors(X, Xs, diag))
+        return reduce(add, self._merge_factors_cov(X, Xs, diag))
+
+    def psd(self, omega):
```
michaelosthege: What is psd? Can you add type hints and a docstring?

michaelosthege (commented):
Some more type hints would be good. I'm not sure if gp/cov.py is passing mypy already, but if not I'd would be good to check the output of `python scripts/run_mypy.py --verbose`` and fix any errors that might appear in the changed lines

pymc/gp/cov.py
fonnesbeck: Could call the method `power_spectral_density` or `spectral_approx`

pymc/gp/cov.py
```diff
@@ -151,41 +129,127 @@ def __array_wrap__(self, result):
...
+        if active_dims is None:
+            self.active_dims = np.arange(input_dim)
+        else:
+            self.active_dims = np.asarray(active_dims, int)
+
+        if max(self.active_dims) > self.input_dim:
+            raise ValueError("Values in `active_dims` can't be larger than `input_dim`.")
+
+    @property
+    def D(self):
```
fonnesbeck: Might consider a more descriptive name, like `dims`

pymc/gp/cov.py
```diff
@@ -408,6 +526,9 @@ def diag(self, X):
     def full(self, X, Xs=None):
         raise NotImplementedError
 
+    def psd(self, omega):
+        raise NotImplementedError
+
 
 class Periodic(Stationary):
     r"""
```
fonnesbeck, line 518: Should the docstrings for the two different kernels be associated with their respective methods?

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,239 @@
...
+        dims: None
+            Dimension name for the GP random variable.
+        """
+
+        X, _ = self.cov_func._slice(X)
+        self._set_boundary(X)
+        omega, phi, m_star = self._eigendecomposition(X, self.L, self.m, self.D)
+        psd = self.cov_func.psd(omega)
+
+        if self.drop_first:
```
fonnesbeck: Consider simplification to remove mostly redundant `if` statement: set `i = int(self.drop_first==True)` and then use `i` instead of 1 in the subtraction and indices.

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,239 @@
...
+        if getattr(self, "beta", None) is None:
+            raise ValueError(
+                "Prior is not set, can't create a conditional. Call `.prior(name, X)` first."
+            )
+        Xnew, _ = self.cov_func._slice(Xnew)
+        omega, phi, _ = self._eigendecomposition(Xnew, self.L, self.m, self.D)
+        psd = self.cov_func.psd(omega)
+        return self.mean_func(Xnew) + at.squeeze(at.dot(phi, self.beta * psd))
+
+    def conditional(self, name, Xnew, dims=None):
```
fonnesbeck: Add type hints

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,239 @@
...
+        """A helper function which gives the approximate kernel or covariance matrix K. This can be
+        helpful when trying to see how well an approximation may work.
+        """
+        X, _ = self.cov_func._slice(X)
+        self._set_boundary(X)
+        omega, phi, _ = self._eigendecomposition(X, self.L, self.m, self.cov_func.D)
+        psd = self.cov_func.psd(omega)
+        return at.dot(phi * psd, at.transpose(phi))
+
+    def prior(self, name, X, dims=None):
```
fonnesbeck: Add type hints

pymc/gp/cov.py
bwengals: good call, changed to `power_spectral_density`

pymc/gp/cov.py
bwengals: went with `n_dims`, since `dims` is used for other things

pymc/gp/hsgp.py
bwengals: much better... changed to this

bwengals:
Have a type hint question, what's the best type for something like `X`?  It can be a union of np.ndarray, tensorconstant, a tensor of PyMC variables.  Really anything tensorlike should work.  This seems a bit verbose though.  What do you think @michaelosthege or @fonnesbeck?

michaelosthege:
> Have a type hint question, what's the best type for something like `X`? It can be a union of np.ndarray, tensorconstant, a tensor of PyMC variables. Really anything tensorlike should work. This seems a bit verbose though. What do you think @michaelosthege or @fonnesbeck?

It's true that we don't type hint "tensor-like" in most places. We should probably add a convenient type alias somewhere..
`TensorConstant` inherits `TensorVariable`, so `Union[np.ndarray, pt.TensorVariable]` should work here.

Even if you decide to not type hint the tensor-like, type hints for other kwargs in the signature should be added, because IIRC mypy automatically skips functions that don't have any type hints in their signature

pymc/gp/cov.py
bwengals: added

bwengals:
Ah thanks @michaelosthege that makes sense, and explains why I couldn't find that type somewhere else in the codebase.  Going with `Union[np.ndarray, pt.TensorVariable]`.

pymc/gp/cov.py
bwengals: makes sense, moved them down to here

pymc/gp/cov.py
bwengals: I think its OK to keep it how it is, because you only want to try to wrap it as `Constant` if it's `Number`ish.  I put the `isinstance(value, Number)` as a catch all for everything else.  Is that right?  I may not be following what you're saying here let me know.

michaelosthege: When I reviewed there was a branch in `_verify_scalar` that raised the `ValueError`, but it looks like you removed it.
With that, the `except ValueError` here is pretty much unreachable, isn't it? It's not covered by a test either.
If you can cover this line with a test I'll be happy.

...

Looking at https://docs.python.org/3/library/numbers.html it looks like a complex number might trigger this branch, because `numbers.Number` is the most generic supertype.
Should we use `numbers.Real` instead of `numbers.Number`?

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,265 @@
+#   Licensed under the Apache License, Version 2.0 (the "License");
```
michaelosthege:
```suggestion
#   Copyright 2023 The PyMC Developers
#
#   Licensed under the Apache License, Version 2.0 (the "License");
```

pymc/tests/gp/test_cov.py
```diff
@@ -181,6 +181,61 @@ def test_inv_rightprod(self):
...
+
+    def test_copsd_multiply(self):
+        # This could be implemented via convolution
+        L = 10.0
+        omega = np.pi * np.arange(1, 101) / (2 * L)
+        with pm.Model() as model:
+            cov1 = 2 * pm.gp.cov.ExpQuad(1, ls=1)
+            cov2 = pm.gp.cov.ExpQuad(1, ls=1)
+
+        with pytest.raises(NotImplementedError):
```
michaelosthege:
```suggestion
        with pytest.raises(NotImplementedError, match="products of covariance functions"):
```

ideally match messages for the tests below. I could't figure out which exceptions they are testing

michaelosthege:
@bwengals where do you see this PR w.r.t. to the finish 🏁 line?
The three remaining threads should be a low bar to get resolved (or ignore), but from your last commit it looks like you might have more changes in mind?

bwengals:
Pretty close!  I'd thought things were wrapping up, but taking a few iterations to settle on the exact API.  Other than your suggestions (thank you btw), I think I'd like to improve the tests for `HSGP` a bit and that should be pretty much it I think.  Have some thoughts for next steps but will try to save them for a future PR.

michaelosthege:
> Pretty close! I'd thought things were wrapping up, but taking a few iterations to settle on the exact API. Other than your suggestions (thank you btw), I think I'd like to improve the tests for `HSGP` a bit and that should be pretty much it I think. Have some thoughts for next steps but will try to save them for a future PR.

Re API I'd be interested in drawing posterior predictive samples at high resolution `Xnew`, or even better drawing callable like here: https://github.com/pymc-devs/pymc/pull/6475

Maybe you already have one, but a test for this use case would be great

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,265 @@
...
+        X: array-like
+            Function input values.
+        dims: None
+            Dimension name for the GP random variable.
+        """
+
+        phi, sqrt_psd = self.prior_components(X)
+
+        if self.parameterization == "noncentered":
+            self.beta = pm.Normal(f"{name}_hsgp_coeffs_", size=sqrt_psd.size)
```
ferrine: Is it possible to add a possibility for multigp output where size has an additional dimension? So far I would do this manually via exposed self.prior_components method

bwengals: yep that's what `self.prior_components` is for.  There are a few different ways one could do it.  I think it's better to let the user do it because it would depend upstream on the covariance.

ferrine: That makes sense, and is actually enough for doing arbitrary things

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,265 @@
...
+        Parameters
+        ----------
+        name: string
+            Name of the random variable
+        Xnew: array-like
+            Function input values.
+        kwargs: dict-like
+            Optional arguments such as `dims`.
+        """
+        fnew = self._build_conditional(Xnew)
```
ferrine: Should it use the same self.prior_components function?

ferrine: Is this snippet correct?
```python
phi, sqrt_psd = gp.prior_components(Xnew)

if self.parameterization == "centered":
    fnew = phi @ self.beta
else:
    fnew = phi @ (self.beta * sqrt_psd])
```

bwengals: For prediction you only need to compute a new basis phi.  The power spectral density doesn't change, it just depends on the number of basis vectors used.  You use the posterior for beta.  Ive refactored it a bit to hopefully make this more clear.  It also works with `pm.set_data` now, like if it were a linear regression!  I renamed `prior_components` to `prior_linearized` and added an example to the docstring there.  let me know if its clearer

twiecki:
What's the status of this?

pymc/gp/cov.py
bwengals: I think so, I updated this part of the code to just check if it's a `numbers.Real` now.

pymc/tests/gp/test_cov.py
bwengals: gotcha, made those changes

bwengals:
> Re API I'd be interested in drawing posterior predictive samples at high resolution Xnew, or even better drawing callable like here: https://github.com/pymc-devs/pymc/pull/6475

Posterior predictive sampling will be very very fast.

bwengals:
Hopefully closing in on the finish line here!

- Refactored `calc_eigenvectors` and `calc_eigenvalues` and `set_boundaries` out, and allowed them to work on either numpy or PyTensor types to make this code easier to integrate into Bambi, @tomicapretto. 
- Added tests that hopefully check well enough whether the prior and conditional are right.  Using a test statistic that compares two sets of samples from a multivariate distribution.  H0 is that they come from the same distribution.  Description [here](https://torchdrift.org/notebooks/note_on_mmd.html).
- A bit stuck now on mypy errors.  I'm not sure how to fix the remaining ones and would definitely appreciate any clues?
- It's been a while since you commented @michaelosthege, but I tried to address your comments.

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,376 @@
...
+    Parameters
+    ----------
+    Xs: array-like
+        Function input values.  Assumes they have been mean subtracted or centered at zero.
+    """
+
+    if tl.__name__ not in ("numpy", "pytensor.tensor"):
+        raise ValueError("tl must be either numpy or pytensor.tensor.")
+
+    S = tl.max(tl.abs(Xs), axis=0)
```
michaelosthege:
```suggestion
    S = tl.max(tl.abs(Xs), axis=0)
    assert isinstance(S, (np.ndarray, TensorVariable))
```
this should help mypy

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,376 @@
...
+TensorVariable = Union[np.ndarray, pt.TensorVariable]
+TensorConstant = Union[np.ndarray, pt.TensorConstant]
+
+
+def set_boundaries(
+    Xs: TensorVariable,
+    L: Optional[Sequence] = None,
+    c: Optional[Union[float, Sequence]] = None,
+    tl: ModuleType = np,
+) -> Tuple[TensorVariable, TensorVariable, Union[TensorConstant, float]]:
```
michaelosthege:
```suggestion
) -> Tuple[Union[np.ndarray, TensorVariable], TensorVariable, Union[TensorConstant, float]]:
```

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,376 @@
...
+    elif c is None and L is None:
+        raise ValueError("At least one of `c` or `L` must be supplied.")
+    # if both are passed, L takes precedent.
+
+    if tl.__name__ == "numpy":
+        L = tl.asarray(L)
+    elif tl.__name__ == "pytensor.tensor":
+        L = tl.as_tensor_variable(L)
+
+    return S, L, c
```
michaelosthege: mypy complains about `L` return type because you re-used the variable from the functions args.

It's better to create a new local variable with the new type, rather than re-assigning an existing variable with new types.

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+def calc_eigenvalues(L: TensorConstant, m: Sequence[int], tl: ModuleType = np):
+    """R Calculate eigenvalues of the Laplacian."""
+    S = np.meshgrid(*[np.arange(1, 1 + m[d]) for d in range(len(m))])
+    S = np.vstack([s.flatten() for s in S]).T
+    return tl.square((np.pi * S) / (2 * L))
```
michaelosthege:
```suggestion
    S = np.meshgrid(*[np.arange(1, 1 + m[d]) for d in range(len(m))])
    Sarr = np.vstack([s.flatten() for s in S]).T
    return tl.square((np.pi * Sarr) / (2 * L))
```
The first assignment defines the type (here: `list`) and that's why the second assignment gives an error.
```
Incompatible types in assignment (expression has type "ndarray[Any, dtype[Any]]", variable has type "List[ndarray[Any, dtype[Any]]]")
Unsupported operand types for * ("float" and "List[ndarray[Any, dtype[Any]]]")
```
By creating a new variable the type conflicts should be resolved.

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,376 @@
...
+            # and then make predictions for the GP using posterior predictive sampling.
+            with model:
+                ppc = pm.sample_posterior_predictive(idata, var_names=["f"])
+        """
+
+        # Index Xs using input_dim and active_dims of covariance function
+        Xs, _ = self.cov_func._slice(Xs)
+
+        # If not provided, use Xs and c to set L
+        S, self.L, self.c = set_boundaries(Xs, self.L, self.c, tl=pt)
```
michaelosthege: The `set_boundaries` returns a `ndarray` or `TensorVariable` for `L`, but the `__init__` signature defined it as a `Sequence[float]`. This gives errors because `ndarray/TensorVariable` don't qualify as `Sequence[float]`.

I would recommend to make `__init__` take `S: Optional[Union[Sequence[float], np.ndarray, TensorVariable]]` but inside `__init__` (`if S is not None`) apply `as_tensor` or `np.asarray` such that `self.S: Optional[Union[TensorVariable, np.ndarray]]`.

You can make the code more readable by aliasing `TensorLike: TypeAlias = Union[TensorVariable, np.ndarray]` somewhere at the top of the file.

michaelosthege (commented):
> A bit stuck now on mypy errors. I'm not sure how to fix the remaining ones and would definitely appreciate any clues?

I commented a few explainers of the remaining mypy problems. Let me know if you have questions.

pymc/gp/cov.py
```diff
@@ -47,26 +49,24 @@
...
+def _verify_scalar(value):
+    if (
+        isinstance(value, pytensor.compile.SharedVariable)
+        and value.get_value().squeeze().shape == ()
+    ):
+        return at.squeeze(value)
+    elif np.asarray(value).squeeze().shape == ():
+        return np.squeeze(value)
+    elif isinstance(value, numbers.Real):
+        return value
```
michaelosthege: Is the third branch not test-covered because `np.asarray(value).squeeze().shape == ()` applies to numbers already?

pymc/gp/cov.py
```diff
@@ -198,19 +258,73 @@ def merge_factors(self, X, Xs=None, diag=False):
...
                     factor_list.append(factor)
+
+            else:
+                factor_list.append(factor)
+
+        return factor_list
+
+    def _merge_factors_psd(self, omega):
+        """Called to evaluatate spectral densities of combination kernels when possible.  Implements
+        a more restricted set of rules than `_merge_factors_cov` -- just additivity of stationary
```
michaelosthege:
```suggestion
        """Called to evaluatate spectral densities of combination kernels when possible.

        Implements
        a more restricted set of rules than `_merge_factors_cov` -- just additivity of stationary
```

Numpydoc style says that docstrings should start with one sentence, followed by an empty line IIRC.
The docs build warns about these things..

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+    The `gp.HSGP` class is an implementation of the Hilbert Space Gaussian process.  It is a
+    reduced rank GP approximation that uses a fixed set of basis vectors whose coefficients are
+    random functions of a stationary covariance function's power spectral density.  It's usage
+    is largely similar to `gp.Latent`.  Like `gp.Latent`, it does not assume a Gaussian noise model
+    and can be used with any likelihood, or as a component anywhere within a model.  Also like
+    `gp.Latent`, it has `prior` and `conditional` methods.  It supports a limited subset of
+    additive covariances.
+
+    For information on choosing appropriate `m`, `L`, and `c`, refer Ruitort-Mayol et. al. or to
+    the pymc examples that use HSGP.
```
michaelosthege:
```suggestion
    the PyMC examples that use HSGP.
```

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+            # Specify the covariance function.  Three input dimensions, but we only want to use the
+            # last two.
+            cov_func = pm.gp.cov.ExpQuad(3, ls=0.1, active_dims=[1, 2])
+
+            # Specify the HSGP.  Use 25 basis vectors across each active dimension, [1, 2]  for a
+            # total of 25 * 25 = 625.  The value `c = 4` means the boundary of the approximation
+            # lies at four times the half width of the data.  In this example the data lie between
+            # zero and  one, so the boundaries occur at -1.5 and 2.5.  The data, both for training
+            # and prediction should reside well within that boundary..
+            gp = pm.gp.HSGP(m=[25, 25], c=4.0, cov_func=cov_func)
```
michaelosthege:
```suggestion
        with pm.Model() as model:
            # Specify the covariance function.
            # Three input dimensions, but we only want to use the last two.
            cov_func = pm.gp.cov.ExpQuad(3, ls=0.1, active_dims=[1, 2])

            # Specify the HSGP.
            # Use 25 basis vectors across each active dimension for a total of 25 * 25 = 625.
            # The value `c = 4` means the boundary of the approximation
            # lies at four times the half width of the data.
            # In this example the data lie between zero and one,
            # so the boundaries occur at -1.5 and 2.5.  The data, both for
            # training and prediction should reside well within that boundary..
            gp = pm.gp.HSGP(m=[25, 25], c=4.0, cov_func=cov_func)
```
just whitespace changes to remove double-whitespace and unnecessary line breaks within sentences.

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+        with model:
+            fcond = gp.conditional("fcond", Xnew=Xnew)
+
+    References
+    ----------
+    -   Ruitort-Mayol, G., and Anderson, M., and Solin, A., and Vehtari, A. (2022). Practical
+    Hilbert Space Approximate Bayesian Gaussian Processes for Probabilistic Programming
+
+    -   Solin, A., Sarkka, S. (2019) Hilbert Space Methods for Reduced-Rank Gaussian Process
+    Regression.
```
michaelosthege:
```suggestion
    -   Ruitort-Mayol, G., and Anderson, M., and Solin, A., and Vehtari, A. (2022). Practical
        Hilbert Space Approximate Bayesian Gaussian Processes for Probabilistic Programming

    -   Solin, A., Sarkka, S. (2019) Hilbert Space Methods for Reduced-Rank Gaussian Process
        Regression.
```
Looks like automatic line breaks? I'm pretty sure they'd break the enumeration

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+            raise RuntimeError("Boundaries `L` required but still unset.")
+        return self._L
+
+    @L.setter
+    def L(self, value):
+        self._L = pt.as_tensor_variable(value)
+
+    def prior_linearized(self, Xs: TensorVariable):
+        """Returns the linearized version of the HSGP, the Laplace eigenfunctions and the square
+        root of the power spectral density needed to create the GP.  This function allows the user
```
michaelosthege:
```suggestion
        """Linearized version of the HSGP, the Laplace eigenfunctions and the square
        root of the power spectral density needed to create the GP.

        This function allows the user
```

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+        to bypass the GP interface and work directly with the basis and coefficients directly.
+        This format allows the user to create predictions using `pm.set_data` similarly to a
+        linear model.  It also enables computational speed ups in multi-GP models since they may
+        share the same basis.  The return values are the Laplace eigenfunctions `phi`, and the
+        square root of the power spectral density.
+
+        Correct results when using `prior_linearized` in tandem with `pm.set_data` and
+        `pm.MutableData` require two conditions.  First, one must specify `L` instead of `c` when
+        the GP is constructed.  If not, a RuntimeError is raised.  Second, the `Xs` needs to be
+        zero centered, so it's mean must be subtracted.  An example is given below.
```
michaelosthege: Info about the return values should be in its own section

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+        if self._parameterization == "noncentered":
+            return self.mean_func(Xnew) + phi[:, i:] @ (beta * sqrt_psd)
+
+        elif self._parameterization == "centered":
+            return self.mean_func(Xnew) + phi[:, i:] @ beta
+
+    def conditional(self, name: str, Xnew: TensorVariable, *args, **kwargs):
+        R"""
+        Returns the (approximate) conditional distribution evaluated over new input locations
+        `Xnew`.  If using the
```
michaelosthege: * First sentence should be followed by an empty line
* Second sentence `If using the` is incomplete

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+
+        Parameters
+        ----------
+        name: string
```
michaelosthege:
```suggestion
        name
```
Should be `name : str` or since there's already a type hint in the signature IMO it's okay to leave it out here

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+            Name of the random variable
+        Xnew: array-like
+            Function input values.
+        kwargs: dict-like
+            Optional arguments such as `dims`.
```
michaelosthege:
```suggestion
        Xnew : array-like
            Function input values.
        **kwargs : dict, optional
            Optional arguments such as `dims`.
```
See https://numpydoc.readthedocs.io/en/latest/format.html#parameters

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+        phi = calc_eigenvectors(Xnew - X_mean, self.L, eigvals, self._m, tl=pt)
+        i = int(self._drop_first == True)
+
+        if self._parameterization == "noncentered":
+            return self.mean_func(Xnew) + phi[:, i:] @ (beta * sqrt_psd)
+
+        elif self._parameterization == "centered":
+            return self.mean_func(Xnew) + phi[:, i:] @ beta
+
+    def conditional(self, name: str, Xnew: TensorVariable, *args, **kwargs):
```
michaelosthege:
```suggestion
    def conditional(self, name: str, Xnew: TensorVariable, **kwargs):
```
The args are not used within the function!

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+        ----------
+        name: string
+            Name of the random variable
+        Xnew: array-like
+            Function input values.
+        kwargs: dict-like
+            Optional arguments such as `dims`.
+        """
+        fnew = self._build_conditional(Xnew)
+        return pm.Deterministic(name, fnew, dims=kwargs.get("dims"))
```
michaelosthege:
```suggestion
        return pm.Deterministic(name, fnew, **kwargs)
```
This way other kwargs will be forwarded too. If you don't want that, I suggeste having only `dims` in the signature and not a `**kwargs`

pymc/gp/hsgp.py
```diff
@@ -0,0 +1,357 @@
...
+
+        Parameters
+        ----------
+        name: string
+            Name of the random variable
+        X: array-like
+            Function input values.
+        dims: None
+            Dimension name for the GP random variable.
+        """
```
michaelosthege: See similar comments below:
* args/kwargs getting dropped silently
* Parameters section formatting
* Docstring doesn't match signature

michaelosthege (commented):
The HGSP module must be included in `docs/source/api/gp.rst` otherwise it won't be rendered in the docs.

I commented a few (nitpicky) things about docstring formatting.. Let me know if you want me to help out taking care of these (this weekend).

pymc/gp/hsgp.py
bwengals: Right, I added the args/kwargs issues you're pointing to make mypy pass.  The code originally just passed dims like the docstring says.  It's also the only input used.  Without doing this mypy complains that the HSGP function signatures don't match the signatures of the base class.  Is there a third option?  I could have HSGP not be a subclass the base gp class, which seems weird to me because it is a subclass. 

Is it actually a bad thing if args/kwargs get dropped silently?  At least with kwargs isnt that somewhat expected?  The docstring says it just takes dims, so what else should a user expect?

pymc/gp/cov.py
bwengals: yep, looks like it.  I'll remove it

pymc/gp/hsgp.py
michaelosthege: I'd expect it to take any of the kwargs of a typical PyMC distribution, for example `dims`, `initval`...

Can you make the signature of the base class more specific? (Specifying just `dims`, not `*args, **kwargs`..)

Otherwise you can go with a `# type: ignore` comment, but be aware that this violates the Liskov subtitution principle. Best check https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides

pymc/gp/hsgp.py
bwengals: it's saying to get correct results you need to give `L` and zero center `Xs`, I think thats about inputs

pymc/gp/hsgp.py
bwengals: Great, added the ignores.  Maybe in a later PR I can refactor the GP module to not violate the liskov substitution principle.

pymc/gp/hsgp.py
michaelosthege: oh I marked the wrong lines - meant the two last lines of the previous paragraph..

> [...] The return values are the Laplace eigenfunctions `phi`, and the
> square root of the power spectral density.

Regardless, there should be a `Returns` section ;)

bwengals: Would you be willing to approve as is?  This is something that could be fixed in a later pull request.  A separate PR that went through all the methods and added a `Returns` section would be very manageable to write and review.

pymc/gp/cov.py
```diff
@@ -47,26 +49,22 @@
 ]
 
 
-class Covariance:
-    r"""
-    Base class for all kernels/covariance functions.
+def _verify_scalar(value):
+    if (
```
ricardoV94: Why not call `at/np.squeeze` and capture the errors that both emit when the inputs are not allowed to be squeezed?

pymc/gp/cov.py
```diff
@@ -47,26 +49,22 @@
...
+        isinstance(value, pytensor.compile.SharedVariable)
+        and value.get_value().squeeze().shape == ()
+    ):
+        return at.squeeze(value)
```
ricardoV94: This is slightly incorrect. If I pass `pt.shared(np.ones(1))` it will fail and not be captured by your error.

If I pass `pt.shared((1), shape=(1,))` then it will work.

I suggest just calling squeeze directly. Also inputs could be constants (e.g., from pm.ConstantData) and would not meet either branch, no?

ricardoV94 (commented):
LGTM, just a minor remark

pymc/gp/cov.py
bwengals: This functions is only used by exponentiated kernels now (which aren't usable by HSGP).  I tried to use `_verify_scalar` for other HSGP stuff earlier in this PR (that I cant recall) but it got refactored out. 

Thanks for catching this potential regression.  In practice though, I think its extremely unlikely anyone would do anything more sophisticated than
```
cov_func = eta**2 * pm.gp.cov.ExpQuad(1, ls)**2 # or cubed or something
```
`_verify_scalar` is only handling the power, `**2`.  I'm also OK with not checking/supporting type checking here.  Basically usage wise there's really no chance that exponent is anything other than "2" or maybe "3".  I've not see this really in the wild in models, more that its just kinda neat that you can exponentiate kernels (because products of kernels are kernels) so why not support it.   

How about I roll things back to how they are in master for this, which is, 
```python
    def __pow__(self, other):
        if (
            isinstance(other, pytensor.compile.SharedVariable)
            and other.get_value().squeeze().shape == ()
        ):
            other = at.squeeze(other)
            return Exponentiated(self, other)
        elif isinstance(other, Number):
            return Exponentiated(self, other)
        elif np.asarray(other).squeeze().shape == ():
            other = np.squeeze(other)
            return Exponentiated(self, other)

        raise ValueError("A covariance function can only be exponentiated by a scalar value")
```
The errors aren't captured, but users will see what happens when squeeze is attempted.

ricardoV94: No strong preference, but why is this not enough?

```python
def __pow__(self, other):
  other = as_tensor_variable(other).squeeze()
  if not other.ndim == 0:
    raise ValueError(...)
  return Exponentiated(self, other)
```

bwengals: alright that's why they pay you the big bucks, changed it to this

pymc/gp/hsgp.py
bwengals: marking resolved since the mypy stuffs cleaned up now.   Switched to the name `TensorLike` too

pymc/gp/hsgp.py
bwengals: args/kwargs arent getting dropped silently now, and the docstrings match the signature.  I also added a `Returns` text to the docstring for `prior_linearized`.  

I held off on adding `Returns` to the docstrings for the rest of the methods because the rest of the GP module methods don't have a returns sections either.  I think it's good to have of course, but out of scope to add it for everything here.  I also would like to defer refactoring how `Base` is used so it doesn't violate the Liskov sub. principle you pointed out because that also touches the entire gp submodule.  It would be good to fix, but it's been that way for several years now without issue because users don't use `Base`.  But I agree, structurally it could be better.

ricardoV94 (approved):

michaelosthege (approved):
@bwengals I think you can squash-merge 🥳

PR #6462, opened 2023-01-20
Use separate argument in `CustomDist` for functions that return symbolic representations

ricardoV94:
The initial API introduced in https://github.com/pymc-devs/pymc/pull/6361 of overloading the `random` kwarg and branching depending on whether the function returned PyTensor variables or failed whatsoever was pretty bad. For instance it would mask obvious errors like calling `pm.Normal.dist(banana=...)`.

Such errors won't be masked anymore, and the user must instead pass such functions to the `dist` kwarg. I added a temporary check to help users navigate the API change.

It also cleans some stale references to Aesara

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -613,7 +613,7 @@ class _CustomSymbolicDist(Distribution):
         cls,
         *dist_params,
         class_name: str,
-        random: Callable,
+        dist: Callable,
         logp: Optional[Callable] = None,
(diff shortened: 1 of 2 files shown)


michaelosthege (approved):
I wanted to suggest raising a deprecation warning, but old calls will now result in a `TypeError` because `dist` is a required kwarg.

Not great, but I guess there was the experimental warning, so YOLO.

PR #6465, opened 2023-01-21
added SpecShape placeholder class and logprob function

michaelraczycki:
**What is this PR about?**
Work in Progress
Adding logprob function for Assert and SpecifyShape, by creating their corresponding files and classes in logprob folder

## Major / Breaking Changes
- ...

## New features
- added logprob_specify_shape function

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

### DIFF
--- a/pymc/logprob/assert.py
+++ b/pymc/logprob/assert.py
@@ -1,0 +1,58 @@
+#   Copyright 2022- The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+#   MIT License
+#
+#   Copyright (c) 2021-2022 aesara-devs
+#
+#   Permission is hereby granted, free of charge, to any person obtaining a copy
+#   of this software and associated documentation files (the "Software"), to deal
+#   in the Software without restriction, including without limitation the rights
+#   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+#   copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


michaelraczycki:
@ricardoV94, Please take a look at it, if you thing it's what you had in mind I'll create corresponding file for Assert.

ricardoV94:
@michaelraczycki to give some more context (which we should write down as documentation). The idea with the logprob submodule is that it converts graphs of RandomVariables to graphs of log-probabilities. In the case of SpecifyShape, the idea is that we might generate a graph like

```python
import pytensor
import pytensor.tensor as pt
import pymc as pm
import numpy as np

mu = pytensor.shared(np.ones(10,))
x = pm.Normal.dist(mu)
xs = pt.specify_shape(x, (10,))
```
And then we are interested in evaluating the logp of `xs`. If you try it in the main version of PyMC you get an error:
```python
pm.logp(xs, np.zeros((10,)))  # NotImplementedError: PyMC could not infer logp of input variable.
```

We never specified how to infer the logp of a `SpecifyShape` operation. For that we have to do the following:
1. Create a subclass of the original operator `SpecifyShape`. This subclass, will have a logp function, but not the original one. We would call it something like `MeasurableSpecifyShape`
2. Look for graphs where we have a SpecifyShape on another MeasurableVariable. This MeasurableVariable should be unnacounted for (meaning it doesn't have a value variable). If we find such case, we replace the original SpecifyShape by the MeasurableSpecifyShape with the same inputs.

You did part 1) Defined a subclass of SpecifyShape (we should call it MeasurableSpecifyShape) and a logp function
You are missing part 2) where you actually replace SpecifyShape by this new object. Here is where is done for the Cumsum case

https://github.com/pymc-devs/pymc/blob/ae459190ddbeb3d18c9108f789f0177a8b78bceb/pymc/logprob/cumsum.py#L84-L127

ricardoV94:
In terms of files:

We can put the new SpecifyShape stuff in `logprob/shape.py`. We might want to implement other shape related logprobs in the future, and keep them in the same file.

For the Assert, we should actually implement it for the baseclass `CheckAndRaise`: https://github.com/pymc-devs/pytensor/blob/96060bf99920226fd3b5d13cc078e1ad75db0a0c/pytensor/raise_op.py#L27

`Assert` is a specific subclass that raises an AssertionError, other subclasses can raise custom types of Errors. We should put it in `logprob/raise.py` to reflect the original more general Op.

ricardoV94:
Closing in favor of #6538

PR #6470, opened 2023-01-23
Enforce `dims` elements to be strings

### LINKED ISSUE #6379: Bug in conversion of `None` dims to InferenceData
### Description

This snippet hast the effect of shifting the known dims to the "left":

https://github.com/pymc-devs/pymc/blob/6b60804c3a5fdb75f95cc60042061a6e38ec72a3/pymc/backends/arviz.py#L217-L220

```python
import pymc as pm

with pm.Model(coords={"a": range(5)}) as m:
    x = pm.Normal("x", shape=(10, 5), dims=(None, "a"))        
    pm.sample()
```
```
ValueError: conflicting sizes for dimension 'a': length 10 on the data but length 5 on coordinate 'a'
```
Whereas this wrong specification goes unnoticed:
```python
    x = pm.Normal("x", shape=(5, 10), dims=(None, "a"))        
```

Do we need `None` dims? It's annoying if we want to create Xarray datasets in multiple places as in #6374

More generally, it seems we don't do any sort of type check. The following are all allowed at creation time. I don't know if that's by design.

```python
import pymc as pm
with pm.Model(coords={"a": range(5)}) as m:
    x = pm.Normal("x", shape=(10, 5), dims=("a", ..., 5, pm.Normal.dist()))        
    pm.sample()
```

ricardoV94:
I think there was a confusion introduced in https://github.com/pymc-devs/pymc/commit/64e63d8ac83586848c2a997844367f4f61641ebb

Unless I am missing something, it's not possible to have arbitrary missing dims. Only works if missing dims are at the rightmost positions.

Closes #6379

### DIFF
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -214,11 +214,9 @@ class InferenceDataConverter:  # pylint: disable=too-many-instance-attributes
         }
 
         self.dims = {} if dims is None else dims
-        model_dims = {
-            var_name: [dim for dim in dims if dim is not None]
-            for var_name, dims in self.model.named_vars_to_dims.items()
-        }
+        model_dims = {k: list(v) for k, v in self.model.named_vars_to_dims.items()}
         self.dims = {**model_dims, **self.dims}
+
         if sample_dims is None:
             sample_dims = ["chain", "draw"]
         self.sample_dims = sample_dims
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1317,6 +1317,8 @@ class Model(WithMemoization, metaclass=ContextMeta):
         # the length of the corresponding RV dimension.
         if dims is not None:
             for d, dname in enumerate(dims):
+                if not isinstance(dname, str):
+                    raise TypeError(f"Dims must be string. Got {dname} of type {type(dname)}")
                 if dname not in self.dim_lengths:
                     self.add_coord(dname, values=None, length=rv_var.shape[d])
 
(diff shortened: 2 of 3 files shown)


michaelosthege:
I think #6379 was accurate: The problem was only in the `InferenceData` conversion and [the solution Oriol suggested here](https://github.com/pymc-devs/pymc/issues/6379#issuecomment-1366971440) should fix it.
The test you wrote should be able to confirm that

IIRC there are some valid use cases for `None` dims with multivariate RVs such as a `MultivariateNormal`, that's why we (and ArviZ) had this in the first place

ricardoV94:
In that case I would prefer to set the dummy dim names at variable creation time, and not when we convert to InferenceData.

ricardoV94:
There are other issues with `None`. For instance, `model.add_coord` assumes all dims are a string, but we are calling it with `None` dname from `model.register_rv`. It seems like the infrastructure to allow unspecified dims was not really worked out completely.

michaelosthege (approved):

OriolAbril:
looks good.

In case it helps, here is some extra context on arviz and xarray. Neither of them support `None` as dimensions, but they support in some cases not providing dimensions or providing only part of them (in which case, the provided dims will be used to name the first n axis, the n+1 and more axis get default dim names). Moreover, xarray allows any hashable to be used as dimension name, not necessarly strings. ArviZ aims to not break when dims aren't strings (using reprs and str fallbacks if necessary) but it kind of assumes that dimension names are strings.

Issue #6399, opened 2022-12-15
ENH: Add code example to Potential docstrings


ricardoV94:
https://www.pymc.io/projects/docs/en/latest/api/generated/pymc.Potential.html

Note: This is labeled beginner friendly because it is self-contained, it requires modifiying only a single file and not too many lines either. It does require being a PyMC user who knows how to use pymc.Potential. Thus, it is a good choice for someone who is already a PyMC user but has never contributed to the library or to any open source library, while at the same time it is a bad choice for an experienced oss contributor who has never used PyMC.

YASHMANMODE2502:
Hi @ricardoV94 

I can solve this issue. Do we have to just provide a code example of `pymc.Potential()` function?
If yes then where I can update the code?

Thank you

reshamas:
> Hi @ricardoV94
> 
> I can solve this issue. Do we have to just provide a code example of `pymc.Potential()` function? If yes then where I can update the code?
> 
> Thank you

One of the learning steps in contributing is to explore the code base and find which file(s) need to be updated. Want to give that a try? ;)

YASHMANMODE2502:
Hi @reshamas 

Yes I want to give it a try, will you help me so that i will learn how to explore the codebase.

reshamas:
To learn more about the pymc project, check out this playlist:
https://www.youtube.com/playlist?list=PLBKcU7Ik-ir99uTvN0315hIVLuyj4Q1Gt

reshamas:
@ricardoV94 @OriolAbril 
I removed the "beginner friendly" label for this one.

cc:  @SangamSwadiK @BerylKanali

YASHMANMODE2502:
> To learn more about the pymc project, check out this playlist:
> https://www.youtube.com/playlist?list=PLBKcU7Ik-ir99uTvN0315hIVLuyj4Q1Gt

Thank you @reshamas for your help, I will be connected to get some help and give others some help :)

ricardoV94:
> @ricardoV94 @OriolAbril 
> I removed the "beginner friendly" label for this one.
> 
> cc:  @SangamSwadiK @BerylKanali

I think that is setting the bar for beginner friendly issues a bit too low.

OriolAbril:
Added the label again and extended the description

larryshamalama:
Closed by #6559

PR #6489, opened 2023-01-28
Add code example to Potential docstrings

Dhruvanshu-Joshi:
**What is this PR about?**
In this PR , I have provided a solution to close the issue [#6399](https://github.com/pymc-devs/pymc/issues/6399).
The docstring for the Potential function is rather limited. I have made efforts to include some code examples to better understand this. I also focused on the nature of the arguments that are passed to this function and have included an example for the same.

**Checklist**

## Documentation
- Add code example and a short explanation of the code example to Potential docstrings.
- In the first example, we define a constraint such that the probability density that the model produces takes into consideration that the sum of x and y should be lesser than 1.
- In the second example, we emphasis that the var argument can be a PyTensor or Scalar variable. For the same purpose, I use a function that'll return a Scalar variable. Had we not done it , we might have encountered an error.


pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
 def Potential(name, var, model=None):
     """Add an arbitrary factor potential to the model likelihood
 
+      Potential function is used to add arbitrary factors i.e. constraints that we define to the likelihood function of a model to adjust the probablity density of the model.
```
larryshamalama:
```suggestion
      The Potential function is used to add arbitrary factors (such as constraints or other likelihood components) to adjust the probability density of the model. 
```

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+
+        The Potential function here acts as a gatekeeper allowing only samples that satisfy the constraint to contribute to the likelihood. This ensures that the sample generated by the sampler are consistent with the constraint.
+
+        Have a look at the following example:
+       
+     .. code:: python
+
+        with pm.Model() as model:
+          x = pm.Normal('x', mu=0, sigma=1)
+          y = pm.Normal('y',mu=x, sigma=1,observed=data)
```
larryshamalama:
```suggestion
          y = pm.Normal('y', mu=x, sigma=1, observed=data)
```

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+    .. code:: python
+
+        with pm.Model() as my_model:
+          x = pm.Normal('x', mu=0, tau=1)
+          y = pm.Normal('y', mu=x, tau=1)
+
+          constraint = x + y < 1
+          potential = pm.Potential('constraint', constraint, my_model)
+
+        In this example, we define a constraint on the sum of x and y that it can only be lesser than 1. We pass this constraint to the pm.Potential function.
```
larryshamalama:
```suggestion
        In this example, we define a constraint on the sum of x and y to be less than 1. We pass this constraint to the ``pm.Potential`` function.
```

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+        Now the pm.Potential function takes in a expression, a scalar value and the model name as argument.
```
larryshamalama:
```suggestion
```

I would remove this line because the `Parameters` section below provides this information.

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+        Now, the probablity density that the model produces takes into consideration the constraint that the sum of x and y should be lesser than 1.
```
larryshamalama:
```suggestion
        Now, the probability density that the model produces takes into consideration the constraint that the sum of x and y should be less than 1.
```

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+
+        Have a look at the following example:
+       
+     .. code:: python
+
+        with pm.Model() as model:
+          x = pm.Normal('x', mu=0, sigma=1)
+          y = pm.Normal('y',mu=x, sigma=1,observed=data)
+          constraint = x>= 0
+          potential = pm.Potential('x_constraint',-tt.log(tt.switch(constraint, 1, 1e-30)))
```
larryshamalama:
```suggestion
          potential = pm.Potential('x_constraint', -pt.log(pt.switch(constraint, 1, 1e-30)))
```

We recently switched our backend to [PyTensor](https://github.com/pymc-devs/pytensor)!

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+        Now the pm.Potential function takes in a string expression and a scalar value as argument. Hence we pass '-tt.log(tt.switch(constraint, 1, 1e-30))' as second argument which will return a scalar value depending on the constraint and will add to the likelihood function of the model.
+        This is done because PyMC3 uses an energy-based model, which means that the probability of a variable configuration is proportional to the negative of the energy of that configuration. So, the negative logarithm of the switch function ensures that the energy is high (i.e. close to negative infinity) when the constraint is not met, and low (i.e. close to zero) when the constraint is met.
+        The probablity density that this model produces agrees with the constraint that x should be greater than or equal to 0.
+
     Parameters
     ----------
     name: str
-    var: PyTensor variables
+    var: PyTensor or Scalar variables with the same shape as input to the likelihood
+    model: The model object to which the potential function is added. If none is provided, current model is used
```
larryshamalama:
```suggestion
    model: The model object to which the potential function is added. If ``None`` is provided, the current model is used.
```

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+       
+     .. code:: python
+
+        with pm.Model() as model:
+          x = pm.Normal('x', mu=0, sigma=1)
+          y = pm.Normal('y',mu=x, sigma=1,observed=data)
+          constraint = x>= 0
+          potential = pm.Potential('x_constraint',-tt.log(tt.switch(constraint, 1, 1e-30)))
+
+        In this example, we define a constraint on x that it can be either greater than or equal to 0 and pass this constraint to the pm.Potential function.
```
larryshamalama:
```suggestion
        In this example, we define a constraint on x to be greater or equal to 0 via the ``pm.Potential`` function.
```

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+        Now the pm.Potential function takes in a string expression and a scalar value as argument. Hence we pass '-tt.log(tt.switch(constraint, 1, 1e-30))' as second argument which will return a scalar value depending on the constraint and will add to the likelihood function of the model.
```
larryshamalama:
```suggestion
        We pass ``-pt.log(pt.switch(constraint, 1, 1e-30))`` as second argument which will return a scalar value depending on if the constraint is met or not and which will added to the likelihood of the model.
```

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+        This is done because PyMC3 uses an energy-based model, which means that the probability of a variable configuration is proportional to the negative of the energy of that configuration. So, the negative logarithm of the switch function ensures that the energy is high (i.e. close to negative infinity) when the constraint is not met, and low (i.e. close to zero) when the constraint is met.
```
larryshamalama:
```suggestion
The negative logarithm of the switch function ensures that this component of the log-likelihood is high (i.e. close to 30) when the constraint is not met, and low (i.e. close to zero) when the constraint is met.
```

I'm confused with this sentence. What do you mean by "probability of a variable configuration"?

Lastly, personally, I would avoid using the energy concept since it's similar to the likelihood and focus on the log-likelihood since that's what PyMC works with. In the sense that I would avoid introducing too many new concepts as once, although the energy-based analogy is correct (to my understanding and despite my little knowledge of physics 😅). 

However, with that, the sentence would revert the direction of the logic if I'm not mistaken...

larryshamalama (changes requested):
Hi @Dhruvanshu-Joshi, thanks for creating this pull request! Also, thanks for your patience, it seems that our devs have been busy recently... (myself included)

Overall, it looks good! I gave some suggestions to tweak some sentences here and there. Let me know if you would like any assistance with this PR.

Edit: It seems that the pre-commit has not been run successfully; let us know if you would like any pointers regarding that.

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+      Potential function is used to add arbitrary factors i.e. constraints that we define to the likelihood function of a model to adjust the probablity density of the model. 
+
+    .. code:: python
+
+        with pm.Model() as my_model:
+          x = pm.Normal('x', mu=0, tau=1)
+          y = pm.Normal('y', mu=x, tau=1)
+
+          constraint = x + y < 1
+          potential = pm.Potential('constraint', constraint, my_model)
```
ricardoV94: This example does not make sense like this. x + y < 1 is gonna return a variable that evaluates to 0 or 1, which will be added to the model logp. It would make sense to add the `log(constraint)` instead. 

However, I think it's a bit more readable for users to see a constraint with a switch as that's usually how it crops up in examples. Something like `pm.Potential("constraint", pt.log(pt.switch(constraint, 1, 0)))`

But this is very similar to the example below, so just show one of them

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+
+        In this example, we define a constraint on the sum of x and y that it can only be lesser than 1. We pass this constraint to the pm.Potential function.
+        Now the pm.Potential function takes in a expression, a scalar value and the model name as argument.
+        Now, the probablity density that the model produces takes into consideration the constraint that the sum of x and y should be lesser than 1.
+
+        The Potential function here acts as a gatekeeper allowing only samples that satisfy the constraint to contribute to the likelihood. This ensures that the sample generated by the sampler are consistent with the constraint.
```
```suggestion
        The Potential function here acts as a gatekeeper allowing only samples that satisfy the constraint to contribute to the likelihood. This ensures that the samples generated by the sampler are consistent with the constraint.
```

pymc/model.py
```diff
@@ -2035,10 +2035,43 @@ def Deterministic(name, var, model=None, dims=None):
...
+        In this example, we define a constraint on x that it can be either greater than or equal to 0 and pass this constraint to the pm.Potential function.
+        Now the pm.Potential function takes in a string expression and a scalar value as argument. Hence we pass '-tt.log(tt.switch(constraint, 1, 1e-30))' as second argument which will return a scalar value depending on the constraint and will add to the likelihood function of the model.
+        This is done because PyMC3 uses an energy-based model, which means that the probability of a variable configuration is proportional to the negative of the energy of that configuration. So, the negative logarithm of the switch function ensures that the energy is high (i.e. close to negative infinity) when the constraint is not met, and low (i.e. close to zero) when the constraint is met.
+        The probablity density that this model produces agrees with the constraint that x should be greater than or equal to 0.
+
     Parameters
     ----------
     name: str
-    var: PyTensor variables
+    var: PyTensor or Scalar variables with the same shape as input to the likelihood
```
```suggestion
    var: PyTensor expression to be added to the model joint logp 
```

pymc/model.py
ricardoV94: I wouldn't show the `model` argument being used. Most times it's not needed as users add a Potential inside a model context.

pymc/model.py
ricardoV94: I agree with @larryshamalama. Also note that it doesn't have to return a scalar expression at all. Whatever is returned will be summed into a scalar automatically, so a user can return an arbitrary shaped expression

ricardoV94 (changes requested):
Thanks for opening up the PR. This looks nice already. I have three notes:

## First

I think the two examples should be merged into a single one since they both show an example of using Potential to impose a constraint. The first example was incorrect though, and the second one looks better if you make it strict `pt.log(pt.switch(cond, 1, 0))` instead of the arbitrary large or small values.

I would add a second example showing how Potential can be used to obtain an arbitrary prior and not just a binary constraint. Maybe, something like this which corresponds to a power law prior for `max_items` 

```python
with pm.Model():
  # p(max_items) = 1 / max_items
  max_items = pm.Uniform("max_items", lower=1, upper=100)
  pm.Potential('power_prior', pm.math.log(1 / max_items))

  n_items = pm.Uniform('n_items', lower=1, upper=max_items, observed=60)
```

## Second
I would use `pm.math.foo` instead of `pt.foo`.

## Third
We should add a warning box saying that Potentials are only added to the model logp, and as such can only influence logp based sampling, like the one used by `pm.sample`.

They do not affect forward sampling, which is used by `sample_prior_predictive` and `sample_posterior_predictive`. A warning is always raised to alert user of this.

pymc/model.py
Dhruvanshu-Joshi: If we are avoiding to use too many new concepts at once, can we use pm.math for backend computation here instead of explicitly mentioning Theano tensor or PyTensor?
I am quite new to PyTensor and am facing errors when I try to run the suggestion made above.

pymc/model.py
Dhruvanshu-Joshi: Answering the first question,  'probability of variable configuration' refers to the likelihood function. By this I simply  mean to refer the likelihood of a particular set of values for the variables that the model equation uses.

For the second part, no it does not revert the direction of logic... That's bad framing on my part😅

What I meant from "close to  negative infinity" was that the log function/ log-likelihood will be close to negative infinity. The mathematical negative will thus be a large positive value. Energy is the negative logarithmic function that we have defined and thus it will be high when the constraint is not met. 

Can we re-frame this as:
The reciprocal exponential function of the negative logarithm is proportional to the likelihood function. When the constraint is met, the switch function returns 1 making the negative logarithm 0 and hence likelihood function has no penalizing effect. When the constraint is not met the switch function returns 1e-30 and the negative logarithm takes a high value(i.e. 30), reciprocal exponential function of which makes the likelihood function very small and successfully penalizes it.

pymc/model.py
Dhruvanshu-Joshi: I understand how using the logarithm function makes more sense and also notice the repetitiveness in both the examples.
I will add a commit as soon as it is clear if we are moving forward with Pytensor or using pm.math.

ricardoV94: `pm.math` is fine

OriolAbril (changes requested):
Currently on my mobile and can't properly review but want to leave a reminder for myself to review when I gat the chance. It should follow numpydoc: https://numpydoc.readthedocs.io/en/latest/format.html. Examples in their own section below parameters and returns. Warnings also as their own section, the spaces both before and after the parameter name and the parameter info... https://pymc-data-umbrella.xyz/en/latest/sprint/tutorials/docstring_tutorial.html#edit-the-docstring might also be helpful

pymc/model.py
```diff
@@ -2035,10 +2035,49 @@ def Deterministic(name, var, model=None, dims=None):
...
+
+     .. code:: python
+
+        with pm.Model() as model:
+          x = pm.Normal('x', mu=0, sigma=1)
+          y = pm.Normal('y', mu=x, sigma=1, observed=data)
+          constraint = x>= 0
+          potential = pm.Potential('x_constraint', -pm.math.log(pm.math.switch(constraint, 1, 1e-30))
+
+        In this example, we define a constraint on x to be greater or equal to 0 via the ``pm.Potential`` function.
```
ricardoV94: I would put the explanations above the respective examples

pymc/model.py
```diff
@@ -2035,10 +2035,49 @@ def Deterministic(name, var, model=None, dims=None):
...
+
+      Have a look at the following example:
+
+     .. code:: python
+
+        with pm.Model() as model:
+          x = pm.Normal('x', mu=0, sigma=1)
+          y = pm.Normal('y', mu=x, sigma=1, observed=data)
+          constraint = x>= 0
+          potential = pm.Potential('x_constraint', -pm.math.log(pm.math.switch(constraint, 1, 1e-30))
```
ricardoV94: I think the absolute constraint with (1 vs 0) may be more educative than (1 vs 1e-30). I would perhaps show that and mention an alternative something like (1 vs 0.5) would provide a soft constraint instead of a hard one.

I find the 1e-30 odd because it behaves like 0 in practice. The sampler will never accept a negative value for 'x', or perhaps once in a lifetime, which is an odd modelling choice.

We should change the mu of x to be 0.1 perhaps, so that the sampler doesn't start at an invalid point.

ricardoV94 (commented):
Do you think abbreviate a bit the explanations? I don't know if it's just me but they feel a tad too long.

ricardoV94:
Thinking about it again, I like your original idea of adding a joint constraint. Perhaps as a third example?

Something like `Potential(”sum_constraint", -pm.math.log((x + y)**2)`

pymc/model.py
Dhruvanshu-Joshi: I understand that the absolute constraint with (1 vs 0) might be more educative than (1 vs 1e-30), but the problem with it is that it produces error when trying to sample the model because the logarithm of 0 is not defined. If having a working model that can be sampled without encountering errors along with explaining the use of Potential function is not what is required with this code block, a constraint of 1 vs 0 absolutely makes more sense. So to reconfirm, should I change the constraint?

Soft constraints also are very helpful. Should I make a  separate code-block for it or a commented line in the original code-block with an explanation?

ricardoV94: Logarithm of zero is well defined (at least in PyMC land) and leads to -inf logp which is exactly what you want to represent an impossible event in log scale. We use -inf explicitly in most of the logp definitions 9f distributions with constrained support.

About the soft constrained, whichever way you think is better. I have no preference.

Dhruvanshu-Joshi: When I try to sample using potential = pm.Potential('x_constraint',-pm.math.log(pm.math.switch(constraint, 1, 0.0))) I face this error:

[<ipython-input-31-e645e4d4afc7>](https://localhost:8080/#) in <module>
     11   potential = pm.Potential('x_constraint',-pm.math.log(pm.math.switch(constraint, 1, 0)))
     12 
---> 13   trace = pm.sample(1000,tune=1000)

12 frames
[/usr/local/lib/python3.8/dist-packages/aesara/tensor/elemwise.py](https://localhost:8080/#) in transform(r)
    600         def transform(r):
    601             # From a graph of ScalarOps, make a graph of Broadcast ops.
--> 602             if isinstance(r.type, (NullType, DisconnectedType)):
    603                 return r
    604             if r in scalar_inputs:

AttributeError: 'float' object has no attribute 'type'

On defining the Potential function as potential = pm.Potential('x_constraint',-pm.math.log(pm.math.switch(constraint, 1.0, 0.0))) I get this:

Interrupted
[<ipython-input-34-ed86f6db8804>](https://localhost:8080/#) in <module>
     11   potential = pm.Potential('x_constraint',-pm.math.log(pm.math.switch(constraint, 1.0, 0.0)))
     12 
---> 13   trace = pm.sample(1000,tune=1000)

8 frames
[/usr/local/lib/python3.8/dist-packages/pymc/step_methods/hmc/base_hmc.py](https://localhost:8080/#) in astep(self, q0)
    166             message_energy = (
    167                 "Bad initial energy, check any log probabilities that "
--> 168                 "are inf or -inf, nan or very small:\n{}".format(error_logp.to_string())
    169             )
    170             warning = SamplerWarning(

AttributeError: 'numpy.ndarray' object has no attribute 'to_string'

ricardoV94: For it to have a valid initial energy with the hard constraint you have to set the prior of x to not be centered at zero like `x = pm.Normal("x", mu=0.1, sigma=1)` or set an initial value like `x = pm.Normal("x", mu=0, sigma=1, initval=0.1)`

ricardoV94: I don't think it should be negated

```suggestion
          potential = pm.Potential('x_constraint', pm.math.log(pm.math.switch(constraint, 1, 1e-30))
```

pymc/model.py
```diff
@@ -2035,10 +2035,49 @@ def Deterministic(name, var, model=None, dims=None):
...
+        We pass ``-pm.math.log(pm.math.switch(constraint, 1, 1e-30))`` as second argument which will return an expression depending on if the constraint is met or not and which will be added to the likelihood of the model.
+        The reciprocal exponential function of the negative logarithm is proportional to the likelihood function. When the constraint is met,the switch function returns 1 making the negative logarithm 0 and hence likelihood function has no penalizing effect. When the constraint is not met the switch function returns 1e-30 and the negative logarithm takes a high value(i.e. 30) reciprocal exponential function of which makes the likelihood function very small and successfully penalizing it.
+        The probablity density that this model produces agrees with the constraint that x should be greater than or equal to 0.
+
+     .. code:: python
+
+        with pm.Model():
+          # p(max_items) = 1 / max_items
+          max_items = pm.Uniform("max_items", lower=1, upper=100)
+          pm.Potential('power_prior', -pm.math.log(max_items))
```
ricardoV94: They are equivalent, but maybe this one is more readable for users

```suggestion
          pm.Potential('power_prior', pm.math.log(1/max_items))
```

pymc/model.py
larryshamalama: Yes, that's a great idea. @ricardoV94 mentioned it below too

pymc/model.py
Dhruvanshu-Joshi: The negative sign was the problem. I think this is because if we take the negative sign, the Potential function takes in inf as argument instead of -inf.  From my understanding, the potential term should be somewhere close to 0 when the constraint is satisfied and very large when the constraint is not satisfied. What I thought was that both inf and -inf can work as this large value. But when Potential value becomes inf,  for some reason I face the error "'numpy.ndarray' object has no attribute 'to_string'". So, is it wrong if the Potential value becomes inf?

Dhruvanshu-Joshi: The negative sign was the problem. I think this is because if we take the negative sign, the Potential function returns in inf instead of -inf. From my understanding, the potential term should be somewhere close to 0 when the constraint is satisfied and very large when the constraint is not satisfied. What I thought was that both inf and -inf can work as this large value. But when Potential value becomes inf, for some reason I face the error "'numpy.ndarray' object has no attribute 'to_string'". So, is it wrong if the Potential value becomes inf?

ricardoV94: The potential value is just added to the model logp, so it should be -inf (or very negative) when a constraint is violated, so that those draws are rejected. 0 won't have any effect and positive values will make the proposals more likely to be accepted

Dhruvanshu-Joshi: Alright👍 
I shall include this in the docstring too.
Thanks for the help!

pymc/model.py
larryshamalama: Sorry, chipping after much time. I'm not sure what you mean by "reciprocal exponential function of the negative logarithm"... 

However, it looks like you have it figured out in some threads below! Perhaps you can focus on those suggestions rather than on my comments

pymc/model.py
```diff
@@ -2035,10 +2035,74 @@ def Deterministic(name, var, model=None, dims=None):
...
+
+     .. warning::
+
+        Potential functions only influence logp based sampling, like the one used by ``pm.sample``. Potentials, modify the log-probability of the model by adding a contribution to the logp which is used by sampling algorithms which rely on the information about the observed data to generate posterior samples.
+        Potentials are not applicable in the context of forward sampling because they don't affect the prior distribution itself, only the computation of the logp.
+        Forward sampling algorithms generate sample points from the prior distribution of the model, without taking into account the likelihood function. In other words, it does not use the information about the observed data.
+        Hence, Potentials do not affect forward sampling, which is used by ``sample_prior_predictive`` and ``sample_posterior_predictive``. A warning saying "The effect of Potentials on other parameters is ignored during prior predictive sampling" is always raised to alert user of this.
```
OriolAbril: This should use a header following https://numpydoc.readthedocs.io/en/latest/format.html#warnings

pymc/model.py
```diff
@@ -2035,10 +2035,74 @@ def Deterministic(name, var, model=None, dims=None):
...
+      In this case, the larger the deviation, larger will be the negative value (-((x + y)**2)) which the MCMC sampler will attempt to minimize. However, the sampler might generate values for some small deviations but with lower probability hence this is a soft constraint.
+
+     .. code:: python
+
+        with pm.Model() as model:
+          x = pm.Normal('x', mu=0.1, sigma=1)
+          y = pm.Normal('y', mu=x, sigma=1, observed=data)
+          soft_sum_constraint = pm.Potential("soft_sum_constraint", -((x + y)**2))
+
+      The potential value is incorporated into the model log-probability, so it should be -inf (or very negative) when a constraint is violated, so that those draws are rejected. 0 won't have any effect and positive values will make the proposals more likely to be accepted
```
OriolAbril: All this should go at the bottom of the docstring and have a section header too. Ref: https://numpydoc.readthedocs.io/en/latest/format.html#examples

pymc/model.py
```diff
@@ -2035,10 +2035,74 @@ def Deterministic(name, var, model=None, dims=None):
...
+        Forward sampling algorithms generate sample points from the prior distribution of the model, without taking into account the likelihood function. In other words, it does not use the information about the observed data.
+        Hence, Potentials do not affect forward sampling, which is used by ``sample_prior_predictive`` and ``sample_posterior_predictive``. A warning saying "The effect of Potentials on other parameters is ignored during prior predictive sampling" is always raised to alert user of this.
+
+
     Parameters
     ----------
     name: str
-    var: PyTensor variables
+    var: PyTensor expression to be added to the model joint logp
+    model: The model object to which the potential function is added. If ``None`` is provided, the current model is used.
```
```suggestion
    var : PyTensor expression
        Expression to be added to the model joint logp
    model : Model, optional
        The model object to which the potential function is added. If ``None`` is provided, the current model is used.
```

a space is needed both before and after the colon (existing docstrings are generally not a good example to follow), and the first line should contain only `name : type` the description comes in the 2nd+ line and indented.

OriolAbril (changes requested):

Dhruvanshu-Joshi:
Hey @OriolAbril I have implemented all the suggested changes. Is there something else I can do to improve this PR or is this good to merge?

OriolAbril (approved):
Doc preview: https://pymcio--6489.org.readthedocs.build/projects/docs/en/6489/api/generated/pymc.Potential.html looks good, I haven't reviewed the code nor looked at the potential merge/rebase issues

ricardoV94:
@Dhruvanshu-Joshi the changes look good but you got some conflicts (as you can see by the 162 changed files). Usually when this happens the easiest solution is to open a new PR from the up-to-date main branch + your final changes

Dhruvanshu-Joshi:
Hey @OriolAbril @ricardoV94 I have opened a new PR [6559](https://github.com/pymc-devs/pymc/pull/6559) for the same issue and even referenced this PR and the issue there. Hope it solves all the conflicts.

OriolAbril:
closing as it was superseeded by #6559

PR #6475, opened 2023-01-23
Refactoring towards `IBaseTrace` interfaces

michaelosthege:
The goal of this PR is to identify & specify the `BaseTrace` attributes & methods that are necessary from the perspectives of internal sampling functions (`pm.sampling.mcmc.*`) and `MultiTrace`.

By extracting these signatures/interfaces into an abstract class `IBaseTrace` it becomes much easier to write a small adapter class that can wrap other kinds of backends to be used instead of `BaseTrace`. (hint hint 🍔)

**Checklist**
+ [x] Explain important implementation details 👆
+ [x] Make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).
+ [x] Link relevant issues (preferably in [nice commit messages](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html))
+ [x] Are the changes covered by tests and docstrings?
+ [x] Fill out the short summary sections 👇

## Major / Breaking Changes
- `MultiTrace.straces` type changes from `Sequence[BaseTrace]` to `Sequence[IBaseTrace]`

## Maintenance
- Internal backend types and sampling functions are refactored for more substitutability
- Fixes flakyness of the `weibull_logp` test

PR #6493, opened 2023-01-30
Register Op as subclass of Distributions with `rv_type` defined

ricardoV94:


michaelosthege:
I'm not familiar with `__new__` or `clsdict`. What's the purpose of this change?

One of the test failures will be get fixed by #6475. Just sayin' :)

larryshamalama:
> I'm not familiar with `__new__` or `clsdict`. What's the purpose of this change?

My understanding is that symbolic Ops were not subclasses of `Distribution` prior to this PR, but it was the case for RV Ops. Is that correct @ricardoV94 ? However, I forgot why it is beneficial to have such Ops to be subclasses of `Distribution`...

ricardoV94:
The reason for these changes is to have distributions that return RandomVariables and SymbolicrandomVariables behave the same.

The metaclass does 2 things:
1. Dispatching logp/logcdf/moment methods when implemented as methods of the class
2. Register the Op class as a subclass of the PyMC distribution

I was recently writing a test in another library where I asserted we were creating the right variables by saying `isinstance(rv.owner.op, pm.Normal)` but this only works for regular distributions. It doesn't work for Censored for example, in which case I had to import the specific rv_type (CensoresdRV)

There is no reason for this distinction. As long as we keep the Distribution pseudo-classes they should work the same regardless of the type of objects they return.

michaelosthege (approved):

PR #6504, opened 2023-02-05
Fix some type issues related to retrieving stats from traces

michaelosthege:
## New features
- `Statsbijection.rmap` can now take incomplete input dicts.

## Maintenance
- `pm.sample` no longer relies on `MultiTrace.get_sampler_stats` that does some black magic with the fetched array.
- Addition of type hints to stats related signatures.
- Updated some stats related docstrings.

### DIFF
--- a/pymc/backends/base.py
+++ b/pymc/backends/base.py
@@ -58,7 +58,7 @@ class IBaseTrace(ABC, Sized):
     varnames: List[str]
     """Names of tracked variables."""
 
-    sampler_vars: List[Dict[str, type]]
+    sampler_vars: List[Dict[str, Union[type, np.dtype]]]
     """Sampler stats for each sampler."""
 
     def __len__(self):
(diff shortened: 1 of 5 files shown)


pymc/backends/base.py
```diff
@@ -79,23 +79,27 @@ def get_values(self, varname: str, burn=0, thin=1) -> np.ndarray:
...
 
         Returns
         -------
-        If the `sampler_idx` is specified, return the statistic with
-        the given name in a numpy array. If it is not specified and there
-        is more than one sampler that provides this statistic, return
-        a numpy array of shape (m, n), where `m` is the number of
-        such samplers, and `n` is the number of samples.
+        stats : np.ndarray
+            If `sampler_idx` was specified, the shape should be `(draws, samplers)`.
```
Armavica: Isn't it the contrary?

pymc/backends/base.py
```diff
@@ -220,23 +224,31 @@ def __getitem__(self, idx):
...
```
Armavica: Same here

pymc/backends/base.py
```diff
@@ -254,14 +266,16 @@ def get_sampler_stats(self, stat_name, sampler_idx=None, burn=0, thin=1):
 
         if vals.dtype == np.dtype(object):
             try:
-                vals = np.vstack(vals)
+                vals = np.vstack(list(vals))
```
Armavica, line 269: I don't understand what happens here. So, after line 260 `vals` is an array of shape `(draws, len(sampler_idxs))`. At line 266 it is either `(draws, len(samplers_idx))` or `(draws,)`. Now, the `dtype` can be `object` iff `_get_sampler_stats` returns this kind of dtype, say, a `SamplerWarning`. So, after this line, the shape of `vals` is either still `(draws, len(samplers_idx))` or `(draws, 1)`. Did I miss something? So my questions are:
* In what case is the exception raised?
* Why do we remove the last singleton dimension to add it right back?
* What is the `list` doing?

michaelosthege: Yeah.. So this whole thing is a dirty hack to collapse object-arrays containing individual NumPy arrays to one array that maybe has a more informative dtype or shape.

For example:

first create an array containing object arrays:
```
>>> a = np.empty(2, dtype=object)
>>> a
array([None, None], dtype=object)
>>> a[0] = np.ones(2, dtype=object)
>>> a[1] = np.zeros(2, dtype=object)
>>> a
array([array([1, 1], dtype=object), array([0, 0], dtype=object)],
      dtype=object)
```
A `vstack` magically collapses them to one object array:
```
>>> np.vstack(a)
array([[1, 1],
       [0, 0]], dtype=object)
```

Unless they have different shapes, which triggers that `ValueError`.
```
>>> a[0] = np.ones(3, dtype=object)
>>> a
array([array([1, 1, 1], dtype=object), array([0, 0], dtype=object)],
      dtype=object)
>>> np.vstack(a)
  File "<__array_function__ internals>", line 180, in vstack
  File "C:\Users\osthege\Miniconda3\envs\pm500\lib\site-packages\numpy\core\shape_base.py", line 282, in vstack
    return _nx.concatenate(arrs, 0)
  File "<__array_function__ internals>", line 180, in concatenate
ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 3 and the array at index 1 has size 2
>>>
```

The `list(vals)` is just to make mypy happy because `vstack(sequence)` but `isinstance(myndarray, Sequence) == False`.

----------

The whole construct is a massive PITA because it makes the shape of sampler stats very hard to predict.

Armavica: Ok, thank you for the explanation!

Armavica (approved):

PR #6510, opened 2023-02-08
Optional McBackend support

michaelosthege:
This PR implements optional support for McBackend based trace backends.

This builds on top of the refactorings from #6475 that decoupled the `pm.sampling` module from `BaseTrace`.

I split the addition of McBackend support into two commits, the first of which makes the necessary code changes without actually adding McBackend as a dependency for the test environments just to confirm that the optional importing works.

The `mcbackend.RunMeta` object that is created by the `make_runmeta` function contains several important metadata about the current model variables. This includes information such as whether a variable is deterministic or not. Related issues:
* https://github.com/arviz-devs/arviz/issues/420
* https://github.com/arviz-devs/arviz/issues/1748
* https://github.com/pymc-devs/pymc/issues/5160

**Checklist**

## New features
- `pm.sample(trace=...)` now accepts `mcbackend.Backend` instances.

### DIFF
--- a/pymc/backends/mcbackend.py
+++ b/pymc/backends/mcbackend.py
@@ -1,0 +1,286 @@
+#   Copyright 2023 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import base64
+import logging
+import pickle
+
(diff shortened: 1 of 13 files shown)


pymc/backends/mcbackend.py
```diff
@@ -0,0 +1,249 @@
...
+    model : pm.Model
+        The current PyMC model.
+
+    Returns
+    -------
+    adapters
+        Chain recording adapters that wrap McBackend Chains in the PyMC IBaseTrace interface.
+    """
+    meta, point_fn = make_runmeta_and_point_fn(initial_point=initial_point, step=step, model=model)
+    run = backend.init_run(meta)
```
michaelosthege, line 276: For most consistent API we should ideally return the `run` object to the user (`pm.sample(return_mcb=True)`)

michaelosthege: I will do more refactoring of the `pm.sample` (specifically the part where it converts to `InferenceData` and returns the result) to make it more versatile and less 🍝

ricardoV94 (approved):

michaelosthege:
Whoopsie, are you aware that the last commit changed the default for situations where McBackend is installed?
I mean it works for the CI, but it might be a little risky

twiecki:
I don't think we should change the default yet.

PR #6514, opened 2023-02-10
issue #5791, dims & cords inference from xarray

michaelraczycki:
added dim inference from xarray, deprecation warning and unittest for the new feature

**What is this PR about?**
Addresses changes requested in Issue #5791 
+ Inferring dims from named pandas indexes
+ Inferring dims from DataArrays
+ Inferring coords from DataArrays
+ A unit test
+ Backwards compatible with deprecation warning rename of the export_index_as_coords to infer_dims_and_coords.

## New features
- It's now possible to infer dims & coords from xarray variables passed to pm.Data

## Documentation
- inline documentation added to data.py

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -209,8 +209,8 @@ def determine_coords(
     model,
     value: Union[pd.DataFrame, pd.Series, xr.DataArray],
     dims: Optional[Sequence[Optional[str]]] = None,
-    coords: Optional[Dict[str, Sequence]] = None,
-) -> Tuple[Dict[str, Sequence], Sequence[Optional[str]]]:
+    coords: Optional[Dict[str, Union[Sequence, np.ndarray]]] = None,
+) -> Tuple[Dict[str, Union[Sequence, np.ndarray]], Sequence[Optional[str]]]:
     """Determines coordinate values from data or the model (via ``dims``)."""
     if coords is None:
         coords = {}
@@ -240,7 +240,7 @@ def determine_coords(
                 dim_name = dim
                 # because coord is expected to be a sequence, we need to convert xarray
                 # using 'tolist()' function
-                coords[dim_name] = value[dim].to_numpy()
+                coords[str(dim_name)] = value[dim].to_numpy()
 
     if isinstance(value, np.ndarray) and dims is not None:
         if len(dims) != value.ndim:
@@ -266,7 +266,7 @@ def ConstantData(
     value,
     *,
     dims: Optional[Sequence[str]] = None,
-    coords: Optional[Dict[str, Sequence]] = None,
+    coords: Optional[Dict[str, Union[Sequence, np.ndarray]]] = None,
     export_index_as_coords=False,
     infer_dims_and_coords=False,
     **kwargs,
@@ -300,7 +300,7 @@ def MutableData(
     value,
     *,
     dims: Optional[Sequence[str]] = None,
-    coords: Optional[Dict[str, Sequence]] = None,
+    coords: Optional[Dict[str, Union[Sequence, np.ndarray]]] = None,
     export_index_as_coords=False,
     infer_dims_and_coords=False,
     **kwargs,
@@ -334,7 +334,7 @@ def Data(
     value,
     *,
     dims: Optional[Sequence[str]] = None,
-    coords: Optional[Dict[str, Sequence]] = None,
+    coords: Optional[Dict[str, Union[Sequence, np.ndarray]]] = None,
     export_index_as_coords=False,
     infer_dims_and_coords=False,
     mutable: Optional[bool] = None,
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -240,7 +240,7 @@ def determine_coords(
                 dim_name = dim
                 # because coord is expected to be a sequence, we need to convert xarray
                 # using 'tolist()' function
-                coords[str(dim_name)] = value[dim].to_numpy()
+                coords[dim_name] = value[dim].to_numpy()
 
     if isinstance(value, np.ndarray) and dims is not None:
         if len(dims) != value.ndim:
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -22,8 +22,10 @@ from copy import copy
 from typing import Dict, Optional, Sequence, Tuple, Union, cast
 
 import numpy as np
+import pandas as pd
 import pytensor
 import pytensor.tensor as at
+import xarray as xr
 
 from pytensor.compile.sharedvalue import SharedVariable
 from pytensor.raise_op import Assert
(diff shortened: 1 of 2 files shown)


pymc/data.py
```diff
@@ -426,7 +451,15 @@ def Data(
         )
 
     # Optionally infer coords and dims from the input value.
+
     if export_index_as_coords:
+        infer_dims_and_coords = export_index_as_coords
+        warnings.warn(
+            "Deprecation warning: 'export_index_as_coords; is deprecated adn will be removed in future versions. Please use 'infer_dims_and_coords' instead.",
```
twiecki:
```suggestion
            "Deprecation warning: 'export_index_as_coords; is deprecated and will be removed in future versions. Please use 'infer_dims_and_coords' instead.",
```

michaelraczycki:
Do you have any tips on how to make the build pass? It seems really not descriptive for someone who runs it first time, and in the documentation I can't find anything to pin the reason for it failing. I've run the pre-commit checks and it seems fine:
Zrzut ekranu 2023-02-12 o 13 33 53

pymc/data.py
michaelraczycki: I've fixed the typo

twiecki (approved):

twiecki (changes requested):
```
        split_key = key.split(".", 1)
        if len(split_key) != 2:
>           raise KeyError(key)
E           KeyError: 'dim'

../../../miniconda3/envs/pymc-test/lib/python3.9/site-packages/xarray/core/dataset.py:185: KeyError
```

michaelraczycki:
oh, good catch. It seems like I put the quotes around dim in line 241 by mistake

michaelraczycki:
@twiecki , it seems like my expectations on how the coords values should look like aren't good, could you check if I'm going in the right direction? I'm not sure if my initial assumption is correct and there's a bug I need to find, or the assumption is wrong and I should change it

twiecki:
CC @michaelosthege

pymc/tests/test_data.py
```diff
@@ -405,6 +405,18 @@ def test_implicit_coords_dataframe(self):
...
+        xr = pytest.importorskip("xarray")
+        data = xr.DataArray([[1, 2, 3], [4, 5, 6]], dims=("y", "x"))
+        with pm.Model() as pmodel:
+            with pytest.warns(DeprecationWarning):
+                pm.ConstantData("observations", data, dims=("x", "y"), export_index_as_coords=True)
+        assert "x" in pmodel.coords
+        assert "y" in pmodel.coords
+        assert pmodel.named_vars_to_dims == {"observations": ("x", "y")}
+        assert tuple(pmodel.coords["x"]) == (data.coords["x"],)
+        assert tuple(pmodel.coords["y"]) == (data.coords["y"],)
```
michaelosthege: The two lines above fail because the left hand side are tuples of coordinate values, and the right hand side is a length-1 tuple.

pymc/tests/test_data.py
```diff
@@ -405,6 +405,18 @@ def test_implicit_coords_dataframe(self):
...
+        xr = pytest.importorskip("xarray")
+        data = xr.DataArray([[1, 2, 3], [4, 5, 6]], dims=("y", "x"))
+        with pm.Model() as pmodel:
+            with pytest.warns(DeprecationWarning):
+                pm.ConstantData("observations", data, dims=("x", "y"), export_index_as_coords=True)
+        assert "x" in pmodel.coords
+        assert "y" in pmodel.coords
+        assert pmodel.named_vars_to_dims == {"observations": ("x", "y")}
+        assert tuple(pmodel.coords["x"]) == (data.coords["x"])
+        assert tuple(pmodel.coords["y"]) == (data.coords["y"])
```
michaelosthege: I think you'd need this:

```python
        assert tuple(pmodel.coords["x"]) == tuple(data.coords["x"].to_numpy())
        assert tuple(pmodel.coords["y"]) == tuple(data.coords["y"].to_numpy())
```

xarray stores coordinate values as a `DataArray`, so either use the above or `np.testing.assert_array_equal`:

```suggestion
        np.testing.assert_array_equal(pmodel.coords["x"], data.coords["x"])
        np.testing.assert_array_equal(pmodel.coords["y"], data.coords["y"])
```

michaelraczycki: Thanks for the suggestions. One question tho, I've triggered all the tests locally and it actually passed, what's the difference between the suggested solutions and the one I've now uploaded?

michaelosthege: Really? Hm, then I'm a bit confused because unless it's a tuple on the right hand side, it should give a boolean array and the error about its truth value being ambiguous

Did you also run the pre-commit? I'm 80 % positive that black will remove the parentheses on the right hand side

michaelraczycki: Yes, ran both pre-commits and the CI tests. The tests were returning the error about truth value, but only when I left the comas:
        assert tuple(pmodel.coords["x"]) == (data.coords["x"],)
         assert tuple(pmodel.coords["y"]) == (data.coords["y"],)
I think I had something else in the parenthesis at the beginning, and forgot to remove the coma. after getting rid of them, both pre-commits and the CI tests were satisfied

michaelraczycki: you're right about the boolean array, it doesn't cause errors tho.
Here's how does the assertion look like (ofc it's just me trying to reproduce it in notebook)
Zrzut ekranu 2023-02-17 o 15 43 24

michaelosthege: what if you put an `assert ` at the beginning of the last line?

Does it raise if you use `(-1, -2, -3)` as the left hand side?

michaelraczycki: you're right, when I try to assert it then it complains about ambiguity of comparing boolean arrays. Strange that when running the test suite locally it didn't fail there

michaelosthege:
Best run mypy locally by
```
pip install types-cachetools "typing-extensions>=3.7.4" "mypy==0.990"
python scripts/run_mypy.py --verbose
```

twiecki:
@michaelosthege Can we merge this?

twiecki (approved):

pymc/data.py
```diff
@@ -225,14 +227,21 @@ def determine_coords(
...
             dim_name = value.columns.name
         if dim_name is not None:
             coords[dim_name] = value.columns
 
+    if isinstance(value, xr.DataArray):
+        if dims is not None:
+            for dim in dims:
+                dim_name = dim
+                # because coord is expected to be a sequence, we need to convert xarray
+                # using 'tolist()' function
```
michaelosthege:
```suggestion
                # str is applied because dim entries may be None
```

pymc/data.py
```diff
@@ -347,7 +373,8 @@ def Data(
         names.
     coords : dict, optional
         Coordinate values to set for new dimensions introduced by this ``Data`` variable.
-    export_index_as_coords : bool, default=False
+    export_index_as_coords : deprecated, previous version of "infer_dims_and_coords"
```
michaelosthege:
```suggestion
    export_index_as_coords : bool
        Deprecated, previous version of "infer_dims_and_coords"
```

pymc/data.py
```diff
@@ -426,7 +453,15 @@ def Data(
         )
 
     # Optionally infer coords and dims from the input value.
+
```
michaelosthege:
```suggestion
```

michaelosthege (approved):

twiecki:
Thanks @michaelraczycki

PR #6515, opened 2023-02-10
Adding function for mutable coords

Dhruvanshu-Joshi:
**What is this PR about?**
In this PR , I have made attempts to incorporate suggestions made from [6497](https://github.com/pymc-devs/pymc/issues/6497).
Coordinate dimensions by default are assumed to be immutable unless they are explicitly mentioned to be mutable in the ```add_coord``` function. 
In this PR, we aim to register mutable dimension coordinates with the model without having to explicitly set ```mutable = True``` in the ```add_coord``` method for each item in ```coords```.

**Checklist**

## Major / Breaking Changes
- no

## New features
- Added another function called ```add_mutable_coords``` which registers new coordinates that are mutable by iterating in a ```for``` loop and calling ```self.add_coord(..., mutable=True)``` for every iteration.

## Bugfixes
- no

## Documentation
- Documentation has been updated consistent with modifications

## Maintenance
- no

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -588,7 +588,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
             self._dim_lengths = {}
         self.add_coords(coords)
         for name, values in coords_mutable.items():
-            self.add_coord(name, values, mutable=True, length=lengths.get(name, None))
+            self.add_coord(name, values, mutable=True)
 
         from pymc.printing import str_for_model
 
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -51,6 +51,7 @@ from pytensor.tensor.var import TensorConstant, TensorVariable
 
 from pymc.blocking import DictToArrayBijection, RaveledVars
 from pymc.data import GenTensorVariable, is_minibatch
+from pymc.distributions.logprob import _joint_logp
 from pymc.distributions.transforms import _default_transform
 from pymc.exceptions import (
     BlockModelAccessError,
@@ -60,7 +61,6 @@ from pymc.exceptions import (
     ShapeWarning,
 )
 from pymc.initial_point import make_initial_point_fn
-from pymc.logprob.joint_logprob import joint_logp
 from pymc.pytensorf import (
     PointFunc,
     SeedSequenceSeed,
@@ -586,6 +586,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
             self._coords = {}
             self._dim_lengths = {}
         self.add_coords(coords)
+        self.add_mutable_coords(coords)
 
         from pymc.printing import str_for_model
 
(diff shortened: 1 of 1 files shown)


twiecki:
Thanks! We should also change the error message to mention to set mutable=True then.

twiecki:
And I think we should make it mutable by default.

Dhruvanshu-Joshi:
How would you like to make it mutable by default? 
What I was thinking was to create another function called ```add_mutable_coord``` which will be same as ```add_coord``` but will assume ```mutable = True```. We can also update ```add_coord``` to consider only cases when ```mutable = False```. We will also throw error messages when someone tries to add immutable coordinates using ```add_mutable_coord``` and vice-versa. Hence by using ```add_mutable_coords``` which will call ```add_mutable_coord```, one can add mutable coordinates to the model context and by using ```add_coords``` which iteratively calls ```add_coord```, we can add immutable coordinates to the model context.
Should I proceed with this?

michaelosthege:
Let's keep it as simple as possible:
* Introduce `coords_mutable` to the `Model.__init__` signature
* Put a two-line `for` iterator right below the `self.add_coords(coords)` line

There's no need to create another method. Not counting tests this can be a 3-line change!

I'm also against changing defaults [because of possible shape or performance issues](https://github.com/pymc-devs/pymc/issues/6497#issuecomment-1416919611).

pymc/model.py
```diff
@@ -548,6 +548,7 @@ def __init__(
         self,
         name="",
         coords=None,
+        coords_mutable=None,
         check_bounds=True,
         *,
```
michaelosthege: I'll apply this change before triggering the tests. The reason being that only by placing the new parameter behind the `*,` we can avoid a change of signature.

```suggestion
        check_bounds=True,
        *,
        coords_mutable=None,
```

pymc/model.py
```diff
@@ -586,6 +587,8 @@ def __init__(
             self._coords = {}
             self._dim_lengths = {}
         self.add_coords(coords)
+        for name, values in coords_mutable.items():
+            self.add_coord(name, values, mutable=True, length=lengths.get(name, None))
```
```suggestion
            self.add_coord(name, values, mutable=True)
```

Dhruvanshu-Joshi:
Hey @twiecki  @michaelosthege  when we use ```coords_mutable.items```, we are expecting a ```none-type object``` to be a ```dictionary``` and hence all of the tests are failing. I suggest that we include this loop in an ```if``` block which will check  
```if coords_mutable is not None :``` similar to how ```add_coords``` does it to avoid this issue. Should I proceed with this?

michaelosthege:
> Hey @twiecki @michaelosthege when we use `coords_mutable.items`, we are expecting a `none-type object` to be a `dictionary` and hence all of the tests are failing. I suggest that we include this loop in an `if` block which will check `if coords_mutable is not None :` similar to how `add_coords` does it to avoid this issue. Should I proceed with this?

sure, it's your PR!

michaelosthege (approved):

twiecki:
@Dhruvanshu-Joshi Unfortunately there are some conflicts we need to resolve before we can merge this.

Issue #6503, opened 2023-02-05
ENH: Add `BlockedStep.stats_dtypes_shapes` to step signatures


michaelosthege:
### Description

# Motivation

The current `BlockedStep.stats_dtypes` class attribute has some downsides:
- It does not pre-register shape information, which makes it impossible to preallocate memory for stats
- It's a list of dicts, but a dict should be sufficient. This creates a lot of added complexity across the codebase.

# Proposal

We should replace `stats_dtypes` by something that includes shape information:
* `()` for always-present, scalar stats
* `(1, 2, 3)` or the like for always-present, array stats
* `(2, 0)` for always-present stats with dynamic shape
* `None` for unknown/dynamic shape, or sparsely emitted stats

For always-present scalar/array stats we can preallocate memory.
Sparse, unknown or dynamic shape stats will end up in `dtype=object` arrays.
For comparison: https://github.com/michaelosthege/mcbackend/blob/6a75046b1496b1c24d1d8b0b22c4078f1c05592d/mcbackend/core.py#L22-L42

We could introduce a  `BlockedStep.stats_dtypes_shapes: Dict[str, Tuple[StatDtype, StatShape]` property with
* `StatDtype = Union[type, np.dtype]`
* `StatShape = Optional[Sequence[int]]`

Its default implementation as a property `BlockedStep.stats_dtypes_shapes` could inspect the old `BlockedStep.stats_dtypes` while defaulting the shape info to `None`.
Thereby we could refactor the internal code, get memory preallocation for stats and maintain backwards-compat to external/custom samplers.

## The `CompoundStep` case

`CompoundStep` might have been why `stats_dtypes` was defined as a list of dicts in the first place.
However, IMO we should flatten the stats of multiple samplers and rename them in one of the following ways:
* Prefixing with `sampler_#__`, for example `sampler_0__accepted` and `sampler_0__accepted`
* Prefixing with `{samplername}_{number}__`, for example `NUTS_0__accepted`, `Metropolis_0__accepted`, `Metropolis_1__accepted`

# Related issues
* #6207 
* #6194
* #5883

covertg:
I think this is the best place to add this, but apologies if not.

[...]

Issue #6207, opened 2022-10-11
Stop squeezing stats by default


michaelosthege:
Squeezing stats by default removes the `"stepmethod"` dimension (which we don't name explicitly), but this does not always work [for `object`-dtyped stats](https://github.com/pymc-devs/pymc/blob/d47dac0e92da2398206b42d3748bc63ff7ff7103/pymc/backends/base.py#L210-L215).

This is (very often) the case for the "warning" stat introduced by #6192, but also for stats that have varying shape.

We should probably not only stop squeezing, but instead unpack stats into a long form, such that there is not even a `stepmeethod` dimension.
This is how McBackend stores stats, because it works for any (hierarchical) combination of step methods.

ricardoV94:
I think we have already done this

michaelosthege:
No, it's still there: https://github.com/pymc-devs/pymc/blob/904a0eaac216732bc358dd91680cd428d95704f0/pymc/backends/base.py#L533

I was the last one to edit that file (7 months ago)

ricardoV94:
I was thinking of this https://github.com/pymc-devs/pymc/commit/8ea8b3b6c2ceb67b964478783049888bb2ce44c1

PR #6517, opened 2023-02-11
Preregister shapes of sampler stats

michaelosthege:
As described in #6503 this adds a `stats_dtypes_shape` attribute to `BlockedStep` to replace `BlockedStep.stats_dtypes`.

It is implemented in a backwards-compatible manner with a deprecation warning for samplers that are not yet updated.

I also specified all stat shapes I was confident about. If someone with NUTS stat experiments could comment the remaining ones that'd be great!

Related issues
* #6503 
* #6207 depends on this

**Checklist**

## New features
- Step methods can now pre-register the shape of emitted sampler stats using the `stats_dtypes_shapes` class attribute. The `stats_dtypes` attribute is being deprecated.

### DIFF
--- a/pymc/step_methods/compound.py
+++ b/pymc/step_methods/compound.py
@@ -18,15 +18,17 @@ Created on Mar 7, 2011
 @author: johnsalvatier
 """
 
+import warnings
+
 from abc import ABC, abstractmethod
 from enum import IntEnum, unique
-from typing import Any, Dict, List, Mapping, Sequence, Tuple, Union
+from typing import Any, Dict, Iterable, List, Mapping, Sequence, Tuple, Union
 
 import numpy as np
 
 from pytensor.graph.basic import Variable
 
-from pymc.blocking import PointType, StatsDict, StatsType
+from pymc.blocking import PointType, StatDtype, StatsDict, StatShape, StatsType
 from pymc.model import modelcontext
 
 __all__ = ("Competence", "CompoundStep")
@@ -48,9 +50,61 @@ class Competence(IntEnum):
     IDEAL = 3
 
 
+def infer_warn_stats_info(
+    stats_dtypes: List[Dict[str, StatDtype]],
+    sds: Dict[str, Tuple[StatDtype, StatShape]],
+    stepname: str,
+) -> Tuple[List[Dict[str, StatDtype]], Dict[str, Tuple[StatDtype, StatShape]]]:
+    """Helper function to get `stats_dtypes` and `stats_dtypes_shapes` from either of them."""
+    # Avoid side-effects on the original lists/dicts
+    stats_dtypes = [d.copy() for d in stats_dtypes]
+    sds = sds.copy()
+    # Disallow specification of both attributes
+    if stats_dtypes and sds:
+        raise TypeError(
+            "Only one of `stats_dtypes_shapes` or `stats_dtypes` must be specified."
+            f" `{stepname}.stats_dtypes` should be removed."
+        )
+
+    # Infer one from the other
+    if not sds and stats_dtypes:
+        warnings.warn(
+            f"`{stepname}.stats_dtypes` is deprecated."
+            " Please update it to specify `stats_dtypes_shapes` instead.",
+            DeprecationWarning,
+        )
+        if len(stats_dtypes) > 1:
+            raise TypeError(
+                f"`{stepname}.stats_dtypes` must be a list containing at most one dict."
+            )
+        for sd in stats_dtypes:
+            for sname, dtype in sd.items():
+                sds[sname] = (dtype, None)
+    elif sds:
+        stats_dtypes.append({sname: dtype for sname, (dtype, _) in sds.items()})
+    return stats_dtypes, sds
+
+
 class BlockedStep(ABC):
     stats_dtypes: List[Dict[str, type]] = []
+    """A list containing <=1 dictionary that maps stat names to dtypes.
+
+    This attribute is deprecated.
+    Use `stats_dtypes_shapes` instead.
+    """
+
+    stats_dtypes_shapes: Dict[str, Tuple[StatDtype, StatShape]] = {}
+    """Maps stat names to dtypes and shapes.
+
+    Shapes are interpreted in the following ways:
+    - `[]` is a scalar.
+    - `[3,]` is a length-3 vector.
+    - `[4, None]` is a matrix with 4 rows and a dynamic number of columns.
+    - `None` is a sparse stat (i.e. not always present) or a NumPy array with varying `ndim`.
+    """
+
     vars: List[Variable] = []
+    """Variables that the step method is assigned to."""
 
     def __new__(cls, *args, **kwargs):
         blocked = kwargs.get("blocked")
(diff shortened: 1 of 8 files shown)


pymc/step_methods/hmc/hmc.py
```diff
@@ -39,27 +39,25 @@ class HamiltonianMC(BaseHMC):
...
+        "energy_error": (np.float64, None),
+        "energy": (np.float64, None),
+        "path_length": (np.float64, None),
+        "accepted": (bool, []),
+        "model_logp": (np.float64, None),
+        "process_time_diff": (np.float64, None),
+        "perf_counter_diff": (np.float64, None),
+        "perf_counter_start": (np.float64, None),
+        "largest_eigval": (np.float64, None),
+        "smallest_eigval": (np.float64, None),
```
michaelosthege: These should be checked, and the `None`-shaped ones should be updated. Are they all scalar?

If there are stats for which the shape must be determined from the `vars`, we can update their entries from the `__init__`.

michaelosthege:
@covertg want to take a shot a reviewing here?

pymc/step_methods/compound.py
```diff
@@ -77,12 +131,21 @@ def __new__(cls, *args, **kwargs):
         if len(vars) == 0:
             raise ValueError("No free random variables to sample.")
 
+        # Auto-fill stats metadata attributes from whichever was given.
+        stats_dtypes, stats_dtypes_shapes = infer_warn_stats_info(
```
covertg, line 135: Assigning `stats_dtypes` and `stats_dtypes_shapes` here in the constructor means that they could fall out of sync later, if only one were to get modified. Is this a case we should consider? I could imagine this happening if a step method wanted to determine stat shape at initialization, for example perhaps for a stat shape that varies with the number of variables passed to the step method.

I'm not sure if that is a compelling case or not. But if it is — perhap these two attributes would be better exposed via `@property` with getters and setters to ensure they stay in sync?

pymc/step_methods/compound.py
```diff
@@ -48,9 +50,61 @@ class Competence(IntEnum):
...
+    Use `stats_dtypes_shapes` instead.
+    """
+
+    stats_dtypes_shapes: Dict[str, Tuple[StatDtype, StatShape]] = {}
+    """Maps stat names to dtypes and shapes.
+
+    Shapes are interpreted in the following ways:
+    - `[]` is a scalar.
+    - `[3,]` is a length-3 vector.
+    - `[4, 0]` is a matrix with 4 rows and a dynamic number of columns.
```
covertg: It seems that both `0` and `None` can express that the stat shape is dynamic (either one dimension of it, or all/unknown dimensions). Is there a reason to prefer `0` for the first case and `None` for the second case?

One consideration is that shape tuples including 0 mean something different in numpy than what we are meaning here. I.e. `np.ones((..., 0, ...))` will always be empty even if the ellipses are filled with positive values. Perhaps it could be more clear to prefer `None` for both cases? So `[4, None]` is a matrix with 4 rows and a dynamic number of columns.

pymc/step_methods/compound.py
```diff
@@ -126,6 +191,20 @@ def stop_tuning(self):
...
+    steps: Iterable[BlockedStep],
+) -> Dict[str, Tuple[StatDtype, StatShape]]:
+    """Combines stats dtype shape dictionaries from multiple step methods.
+
+    In the resulting stats dict, each sampler stat is prefixed by `sampler_#__`.
+    """
+    result = {}
+    for s, step in enumerate(steps):
+        for sname, (dtype, shape) in step.stats_dtypes_shapes.items():
+            result[f"sampler_{s}__{sname}"] = (dtype, shape)
```
covertg, line 204: Currently these dictionary keys are not passed to arviz when creating InferenceData objects. But if/when they are, we'll probably want a way for the user to map back from `<step name>` to `<list of variables that were sampled by that stepper>`

covertg:
Sure, I'll do my best! First time for pymc so obviously please take it with some grains of salt, but hope this is helpful. I don't have the experience with NUTS to confirm the shapes and dtypes so I won't comment there.

covertg (commented):
Main comment is regarding whether we really want to fix `stats_dtypes` and `stats_dtypes_shapes` in the constructor `__new__`, or whether it would be better to allow the step method to modify them later.

pymc/step_methods/compound.py
covertg: Although this suggestion overloads the use of `None` when specifying shape info to pytensor, it does so in the same spirit as pytensor's behavior

pymc/step_methods/compound.py
michaelosthege: I agree that we should get the flexibility to have samplers initialize stats on instantiation instead of specifying them as class attributes.

I also considered properties, but this wouldn't have worked nicely with the assignment of these fields in the class definition.

I don't think that sync is a problem because we can remove the old attribute.

I will think about how a refactor to definition at initialization will look like.

pymc/step_methods/compound.py
michaelosthege: This is mostly motivated by the way how McBackend encodes shapes into protocol buffers, which is by a `repeated uint64`. I discussed this in chat with @ricardoV94 who proposed to use `-1` for dynamic dimensions and have `0` with the same interpretation as NumPy like you wrote.

pymc/step_methods/compound.py
michaelosthege: Agreed. ArviZ should add such a field, but even before that gets done we can add this to `mcbackend.RunMeta`

pymc/step_methods/compound.py
ricardoV94: Yeah, we should use None and then convert that to 0 or -1 in the backend. Or must we do the conversion ourselves here?

michaelosthege: Conversion must be done before the shape is passed to `mcb.Variable` which is going to be located in `pm.backends.mcbackend`

michaelosthege: I changed the docstring to specify `-1` for dynamic dims as discussed

pymc/step_methods/compound.py
```diff
@@ -48,9 +50,61 @@ class Competence(IntEnum):
...
+    Use `stats_dtypes_shapes` instead.
+    """
+
+    stats_dtypes_shapes: Dict[str, Tuple[StatDtype, StatShape]] = {}
+    """Maps stat names to dtypes and shapes.
+
+    Shapes are interpreted in the following ways:
+    - `[]` is a scalar.
+    - `[3,]` is a length-3 vector.
+    - `[4, -1]` is a matrix with 4 rows and a dynamic number of columns.
```
ricardoV94: I would vote to use `None` at the PyMC level as that maps directly to the convention used in PyTensor. When we use these to create a McBackend trace then we can map `None` to `-1`. 

Other backends can use `None` to specify unknown shape if they want.

```suggestion
    - `[4, None]` is a matrix with 4 rows and a dynamic number of columns.
```

ricardoV94 (commented):
Looks good, except for a nitpick

PR #6522, opened 2023-02-14
Rename `zerosum_axes` to `n_zerosum_axes`

michaelraczycki:
**What is this PR about?**
-added n_zerosum_axes parameter
-added backwards compatibility for previous parameter name

## Major / Breaking Changes
- 

## New features
- 

## Bugfixes
- 

## Documentation
- slightly altered ZeroSumNormal inline documentation

## Maintenance
-

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2394,7 +2394,7 @@ class ZeroSumNormal(Distribution):
     ZeroSumNormal distribution, i.e Normal distribution where one or
     several axes are constrained to sum to zero.
     By default, the last axis is constrained to sum to zero.
-    See `zerosum_axes` kwarg for more details.
+    See `n_zerosum_axes` kwarg for more details.
 
     .. math::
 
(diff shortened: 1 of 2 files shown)


AlexAndorra (approved):
LGTM, thanks @michaelraczycki ! I think it's your very first PR on this repo, so congrats 🍾

AlexAndorra:
The doc fail is expected?

michaelraczycki:
@AlexAndorra I'm not sure. I'm not so familiar with this check, and I can't find a manual way to trigger it locally. From what I can see here in the description, it looks like it hits some kind of timeout. Could you give some advice on how to check it to ensure everything is ok?

ricardoV94:
The ReadTheDocs has been failing with timeout in other places, so don't worry about it here. This failing test however has to be fixed: https://github.com/pymc-devs/pymc/actions/runs/4174655801/jobs/7321891536

michaelraczycki:
It might be that the docs check was failing because the solution had a typo, which resulted in the zerosum_axes param not to be correctly saved into its new version (n_zerosum_axes). That caused the support_shape etc. not to get correct values and the corresponding tests to fail. If readthedocs checks test coverage then indeed it would fail because the typo caused 'new' parameter to appear and the tests were not testing stuff that was actually created. I'm not sure if I understand the check correctly, so in any case any suggestions that would help me get it better are highly appreciated!

michaelraczycki:
second mistake was not adapting unit tests, I noticed that they use the internal functions (ofc!) and I didn't switch the params used there to n_zerosum_axes, where inside the multivariate.py I did. Thus creating the "unexpected kwarg" error

michaelraczycki:
@AlexAndorra  could you approve the checks again? I think it should work now

twiecki (approved):

PR #6523, opened 2023-02-14
Replace rvs_to_total_sizes mapping by MinibatchRandomVariables

### LINKED ISSUE #6061: Variable in `_get_scaling` is never used
PyCharm tells me this variable is never used:

https://github.com/pymc-devs/pymc/blob/4ca3414e2f4e95a9cc8b393e1310de22d3afa473/pymc/distributions/logprob.py#L94

CC: @ferrine and @chritter

### LINKED ISSUE #6332: ENH: Consider representing observed scaled RVs directly in the forward graph
When we create a scaled RV (in relation to MiniBatching) we add an extra piece of information `rvs_to_total_sizes` (potentially renamed in #6331) in the Model that we use when obtaining the logp. It would be neat if we didn't have to do this, and could instead represent the RV itself as being a MiniBatchedRV or whatever (don't know if there is a mathematical concept for what we are trying to do here), which may take the total_size as an extra input.

This way we would obtain the right logp without the over-head of managing this extra mapping at the Model level. The fewer mappings we need to maintain the easier it is to transform Models, as in https://github.com/pymc-devs/pymcx/pull/91

***
Relevant code sections:

https://github.com/pymc-devs/pymc/blob/b4c7eb1b7bcbfaae21847799c42922aedbcbf1d9/pymc/model.py#L750-L756

https://github.com/pymc-devs/pymc/blob/b4c7eb1b7bcbfaae21847799c42922aedbcbf1d9/pymc/distributions/logprob.py#L302-L305

ricardoV94:
This PR refactors the way rescaling of minibatched variables is achieved, so as to not require keeping any meta-information at the model level.

Ideally, the API would follow that of other RV factories with something like `pm.MinibatchRV(dist=dist, total_size=...)` so that we don't have to add any extra complexity at the Model level, but for the sake of backwards compatibility we let `Model` do this for now.

I removed support for scaling of non-observed RVs even though it would be equally easy to support them. **This is a breaking change**

Closes #6061 
Closes #6332

### DIFF
--- a/pymc/variational/minibatch_rv.py
+++ b/pymc/variational/minibatch_rv.py
@@ -1,0 +1,113 @@
+#   Copyright 2023 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+from typing import Any, Sequence, Union, cast
+
+import pytensor.tensor as pt
+
+from pytensor import Variable, config
+from pytensor.graph import Apply, Op
+from pytensor.tensor import NoneConst, TensorVariable, as_tensor_variable
+
+from pymc.logprob.abstract import MeasurableVariable, _get_measurable_outputs, _logprob
+from pymc.logprob.abstract import logprob as logprob_logprob
+from pymc.logprob.utils import ignore_logprob
+
+
+class MinibatchRandomVariable(Op):
+    """RV whose logprob should be rescaled to match total_size"""
+
+    __props__ = ()
+    view_map = {0: [0]}
+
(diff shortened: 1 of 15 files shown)


pymc/variational/minibatch_rv.py
```diff
@@ -0,0 +1,113 @@
...
+
+from pymc.logprob.abstract import MeasurableVariable, _get_measurable_outputs, _logprob
+from pymc.logprob.abstract import logprob as logprob_logprob
+from pymc.logprob.utils import ignore_logprob
+
+
+class MinibatchRandomVariable(Op):
+    """RV whose logprob should be rescaled to match total_size"""
+
+    _props = ()
```
ferrine: Should not it be `__props__`?

pymc/variational/minibatch_rv.py
```diff
@@ -0,0 +1,113 @@
...
+MeasurableVariable.register(MinibatchRandomVariable)
+
+
+@_get_measurable_outputs.register(MinibatchRandomVariable)
+def _get_measurable_outputs_minibatch_random_variable(op, node):
+    return [node.outputs[0]]
+
+
+@_logprob.register(MinibatchRandomVariable)
+def minibatch_rv_logprob(op, values, *inputs, **kwargs):
```
ferrine, line 110: This is much more intuitive now

ferrine (approved):
I like the proposed design change

pymc/variational/minibatch_rv.py
ricardoV94: Oh yeah

Issue #6511, opened 2023-02-09
Raise if step sampler has variables that are not in `model.value_vars`


ricardoV94:
### Description

This showed up in https://discourse.pymc.io/t/incorrect-step-assignments-with-custom-step-function/11360/5

It should be easy to check that step samplers only have variables from `model.value_vars`

michaelraczycki:
I'd like to try that out

michaelraczycki:
From what I found out, sample() calls 'assign_step_methods' to get all the steps initialized. I think that adding something like you can see in line 172-173 can solve this issue of unnoticed incorrect step assignments, but I'm not sure whether we want to raise an error or a warning, and in case of an error - which one?
<img width="671" alt="Zrzut ekranu 2023-02-14 o 20 42 22" src="https://user-images.githubusercontent.com/119355382/218847380-5eb89040-e374-4e21-bded-7d10ec5d7920.png">

ricardoV94:
Probably a ValueError saying the step samplers should only have value variables that belong to the model.

michaelosthege:
Closed by #6524

PR #6524, opened 2023-02-14
Add check for variables in step samplers

michaelraczycki:
**What is this PR about?**
Added safety check to prevent incorrect step assignment with custom step functions.
Nów assign_step_method from mcmc.py before assigning variables from step to
assigned_var, checks if they can be found in model.value_vars.

## Major / Breaking Changes
-

## New features
- step value check for assign_step_methods

## Bugfixes
- 

## Documentation
- 

## Maintenance
-

### DIFF
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -168,6 +168,11 @@ def assign_step_methods(model, step=None, methods=None, step_kwargs=None):
         except TypeError:
             steps.append(step)
         for step in steps:
+            for var in step.vars:
+                if var not in model.value_vars:
+                    raise ValueError(
+                        "Step samplers should only contain value variables that belong to the model."
+                    )
             assigned_vars = assigned_vars.union(set(step.vars))
 
     # Use competence classmethods to select step methods for remaining
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -171,7 +171,7 @@ def assign_step_methods(model, step=None, methods=None, step_kwargs=None):
             for var in step.vars:
                 if var not in model.value_vars:
                     raise ValueError(
-                        "Step samplers should only contain value variables that belong to the model."
+                        f"{var} assigned to {step} sampler is not a value variable in the model. You can use `util.get_value_vars_from_user_vars` to parse user provided variables."
                     )
             assigned_vars = assigned_vars.union(set(step.vars))
 
--- a/tests/sampling/test_mcmc.py
+++ b/tests/sampling/test_mcmc.py
@@ -40,11 +40,11 @@ from pymc.step_methods import (
     NUTS,
     BinaryGibbsMetropolis,
     CategoricalGibbsMetropolis,
+    CompoundStep,
     HamiltonianMC,
     Metropolis,
     Slice,
 )
-from step_methods.compound import CompoundStep
 from tests.helpers import SeededTest, fast_unstable_sampling_mode
 from tests.models import simple_init
 
--- a/tests/sampling/test_mcmc.py
+++ b/tests/sampling/test_mcmc.py
@@ -44,6 +44,7 @@ from pymc.step_methods import (
     Metropolis,
     Slice,
 )
+from step_methods.compound import CompoundStep
 from tests.helpers import SeededTest, fast_unstable_sampling_mode
 from tests.models import simple_init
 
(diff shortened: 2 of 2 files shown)


michaelraczycki:
contains feature requested in Issue #6511

pymc/sampling/mcmc.py
```diff
@@ -168,6 +168,11 @@ def assign_step_methods(model, step=None, methods=None, step_kwargs=None):
         except TypeError:
             steps.append(step)
         for step in steps:
+            for var in step.vars:
+                if var not in model.value_vars:
+                    raise ValueError(
+                        "Step samplers should only contain value variables that belong to the model."
```
ricardoV94: Maybe something more informative?

```suggestion
                        f"{var} assigned to {step} sampler is not a value variable in the model. 
                        You can use `util.get_value_vars_from_user_vars` to parse user provided variables.
```

ricardoV94 (changes requested):
Looks good. I suggest a more informative error message. And we need a test, you can simplify the original issue on discourse and use that as  test case.

pymc/sampling/mcmc.py
michaelraczycki: the phrasing was changed as suggested

ricardoV94:
> @ricardoV94, I've noticed that most of the time the 'docs/readthedocs.org:pymc' test is failing, and the steps executed in it usually have very little to do with the PR's themselves. Have you considered splitting it into smaller steps or maybe increasing the time limit?

I don't think we have control over the time limit, it all happens on the ReadTheDocs side

ricardoV94:
This PR needs a new test to show the check is working as expected

michaelraczycki:
Is there something else that I should do here? If not then I could close the corresponding issue

ricardoV94:
> Is there something else that I should do here? If not then I could close the corresponding issue

Yes, you need to add a test that shows the new check is working as expected. That's always a requirement of pull requests, as it will help not changing the behavior inadvertently in the future.

michaelraczycki:
@ricardoV94 So I have an idea to create the test in this way you can see in the commit, I'm not feeling very good with pytensor operations yet, so in case I'm misunderstanding anything please let me know. I'm not sure if the way I'd like to introduce a variable that's not in the model is correct, the test passes when I execute it locally. Feel free to suggest any changes!

pymc/tests/sampling/test_mcmc.py
```diff
@@ -819,6 +820,22 @@ def test_modify_step_methods(self):
...
+        with pm.Model() as model:
+            c1 = pm.HalfNormal("c1")
+            c2 = pm.HalfNormal("c2")
+
+            with pytensor.config.change_flags(mode=fast_unstable_sampling_mode):
+                step1 = NUTS([c1])
+                step2 = NUTS([c2])
+                step2.vars = [
+                    at.dscalar("x"),
+                ]
```
ricardoV94: I think this is a pretty good way to test the functionality. I would suggest specifically passing the random variable `c2` as that was the mistake the user did in the original issue.

```suggestion
                step2.vars = [c2]
```

pymc/tests/sampling/test_mcmc.py
```diff
@@ -819,6 +820,22 @@ def test_modify_step_methods(self):
...
+                step = CompoundStep([step1, step2])
+                with pytest.raises(ValueError):
```
ricardoV94: Always good the match the exact string that you are testing. You can put a substring in `match=` and pytest will only accept a ValueError whose message contains the substring.

```suggestion
                with pytest.raises(ValueError, match=...):
```

ricardoV94 (commented):
Looks good, some small tweak suggestions

pymc/tests/sampling/test_mcmc.py
michaelraczycki: done!

pymc/tests/sampling/test_mcmc.py
michaelraczycki: also changed

ricardoV94 (approved):

michaelosthege:
@michaelraczycki I rebase on latest main (the tests moved).

It looks like you committed with your `user.email` set to a local account name, which is why your name initially did not show up with the commits after I rebased them.

I edited the commit authors to the autogenerated GitHub user email..

Best update your `git config --global user.email` to an address that is configured under https://github.com/settings/emails

michaelosthege (approved):

PR #6528, opened 2023-02-17
Implement icdf for Univariate distribution

michaelraczycki:
**What is this PR about?**
Adding icdf functions for existing distributions
Creation of tests for added idcf functions

## Major / Breaking Changes
- ...

## New features
* icdf function for Uniform distritubtion

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- extended test_continuous.py with corresponding tests


michaelraczycki:
@ricardoV94 if that's what you had in mind I can continue adding them for other distributions, I wanted to create a PR before writing a lot to make sure we're on the same page

pymc/distributions/continuous.py
```diff
@@ -835,6 +838,9 @@ def logcdf(value, loc, sigma):
             msg="sigma > 0",
         )
 
+    def icdf(value, lower, upper, loc, sigma):
+        return stats.truncnorm.ppf(q=value, a=lower, b=upper, loc=loc, scale=sigma)
```
ricardoV94: The `icdf` has to have the same signature as the `logp` and `logcdf` of the same distribution. 

Also you can't call `scipy` or `numpy` routines, the expression has to be implemented exclusively in terms of pytensor operations

pymc/distributions/continuous.py
```diff
@@ -343,6 +343,9 @@ def logcdf(value, lower, upper):
             msg="lower <= upper",
         )
 
+    def icdf(value, lower, upper):
+        return lower + (upper - lower) * value
```
ricardoV94, line 347: This will need `check_parameters` to make sure `lower <= upper`

ricardoV94 (changes requested):
Looks like a good start! Some issues mentioned in the comments below

pymc/distributions/continuous.py
```diff
@@ -1019,8 +1019,41 @@ def logcdf(value, mu, lam, alpha):
             msg="mu > 0, lam > 0, alpha >= 0",
         )
 
-    def icdf(mu, lam):
-        return stats.invgauss(mu=mu, scale=lam)
+    def icdf(value, mu, lam, alpha=0):
```
ricardoV94: We don't provide defaults in these functions, it's always done through the `dist` classmethod

```suggestion
    def icdf(value, mu, lam, alpha):
```

pymc/distributions/continuous.py
```diff
@@ -1019,8 +1019,41 @@ def logcdf(value, mu, lam, alpha):
...
+
+        Returns
+        -------
+            float: The value x such that P(W <= x) = p, where W is the Wald distribution with given mu and lam.
+
+        Raises
+        ------
+            ValueError: If value is not between 0 and 1.
+            ValueError: If lam is not positive.
+        """
```
ricardoV94: Better not to include docs. Users are not supposed to use these methods directly. Instead they will call `pm.icdf(rv, value)` which will ultimately call this with the right arguments.

The docstrings are also incorrect because inputs and outputs are TensorVariables (not floats), and errors are not really raised until the expression is evaluated. These docstrings are really tricky to write, as people usually come to these with a numpy framework in mind, and they don't match the actual code. That's why we removed all of them quite recently (no docstrings is better than misleading docstrings)

pymc/distributions/continuous.py
```diff
@@ -1019,8 +1019,41 @@ def logcdf(value, mu, lam, alpha):
...
+
+        # Compute Wald ICDF
+        x = loc + std * z / (1 - 0.5 * std * z)
+
+        return check_parameters(
+            x,
+            0 <= value <= 1,
+            lam > 0,
+            alpha >= 0,
+            msg="0<=val<=1, lam > 0, alpha >= 0",
```
ricardoV94: nitpick
```suggestion
            msg="0 <= val <= 1, lam > 0, alpha >= 0",
```

ricardoV94 (changes requested):
Looks good, left some comments below and don't miss the test :)

pymc/distributions/continuous.py
michaelraczycki: adjusted

pymc/distributions/continuous.py
michaelraczycki: ok, I understand. It's removed now

pymc/distributions/continuous.py
michaelraczycki: good catch!

pymc/distributions/continuous.py
michaelraczycki: a trial icdf for Wald dist already commited

michaelraczycki:
@ricardoV94 I've included tests in previous commits, following the lead from test_normal_icdf. Please take a look at it, as I might have missed something

pymc/tests/distributions/test_continuous.py
```diff
@@ -2281,5 +2288,25 @@ def test_normal_icdf(self, dist_params, obs, size):
         dist_params = dict(zip(dist_params_at, dist_params))
 
         x = Normal.dist(*dist_params_at, size=size_at)
-
         scipy_logprob_tester(x, obs, dist_params, test_fn=st.norm.ppf, test="icdf")
+
+    def test_uniform_icdf(self, dist_params, obs, size):
```
ricardoV94: You need to parametrize the test parameters, check the use of `pytest.parametrize` in the normal test above

ricardoV94:
> @ricardoV94 I've included tests in previous commits, following the lead from test_normal_icdf. Please take a look at it, as I might have missed something

Oh sorry I must have missed it, I was looking only at the last commit

michaelraczycki:
I assume that I still need to add params like lower and upper for distribution tests that require them. I'm just not quite sure if it's possible to include conditional arguments in @pytest.mark.parametrize, or I should create another set of test parameters (try to group them up, so they can share the same test datasets)

ricardoV94:
> I assume that I still need to add params like lower and upper for distribution tests that require them. I'm just not quite sure if it's possible to include conditional arguments in @pytest.mark.parametrize, or I should create another set of test parameters (try to group them up, so they can share the same test datasets)

Each row in the parametrize should include one set of parameters that are jointly valid to test the icdf. Adding 2 or 3 rows should be enough.

michaelraczycki:
I have a problem, Uniform distribution test requires some data transformation (uniform.icdf takes in value, lower and upper as parameters, and stats.uniform.ppf expects value, loc, scale= upper - lower) and I can't seem to figure it out. So far I have tried to extract the values given to uniform distribution and compute the needed values for uniform.ppf:
Zrzut ekranu 2023-02-21 o 19 22 24
I must understand something wrong, because the test fails with this message:
Zrzut ekranu 2023-02-21 o 19 23 18

michaelraczycki:
I also tried to play with the idea of adding a new if-else to numpy_res assignment, in which I could maybe alter what goes to it:
Zrzut ekranu 2023-02-21 o 19 47 03
problem in this approach would be that I can't come up with an idea to check if the passed variable is a reference to specific function. Any suggestions on how to handle this are highly appreciated

pymc/distributions/continuous.py
```diff
@@ -650,9 +653,9 @@ def dist(
         cls,
         mu: Optional[DIST_PARAMETER_TYPES] = None,
         sigma: Optional[DIST_PARAMETER_TYPES] = None,
-        tau: Optional[DIST_PARAMETER_TYPES] = None,
         lower: Optional[DIST_PARAMETER_TYPES] = None,
         upper: Optional[DIST_PARAMETER_TYPES] = None,
+        tau: Optional[DIST_PARAMETER_TYPES] = None,
```
ricardoV94: We shouldn't change the order of arguments. We can however make everything after sigma kwarg only, since the position of `tau` is indeed strange.

`def dist(cls, mu, sigma, *, tau, lower ...)`

ricardoV94: Better to do this in a separate PR so that it can be reviewed on its own

ricardoV94:
For the compatibility between PyMC and Scipy parametrizations you can wrap scipy icdf in a lambda function that does the parameter conversion (untested code and parametrization expression):

```python
test_fn=lambda value, lower, upper: stats.uniform.icdf(value, lower, upper-lower)
```

pymc/distributions/continuous.py
michaelraczycki: Ok, I'll take care of creating the new PR and cleaning this up

michaelraczycki:
@ricardoV94 , I've adjusted the uniform icdf to use the new value and parameter checks for icdf, I think we should merge it now so It can be used as a template for contributors with further icdf implementations

ricardoV94:
@michaelraczycki the code still includes the untested icdf for the truncated normal, and it's using the old testing routine, not the one introduced in https://github.com/pymc-devs/pymc/pull/6583

ricardoV94 (approved):
Looks good. Should be squashed merged

PR #6529, opened 2023-02-17
Infer logprob of `Ifelse` graphs

ricardoV94:
```python
import pymc as pm
import numpy as np

from pytensor.ifelse import ifelse

def ifelse_dist(cond, _):
    x = pm.Normal.dist(mu=[-1, 1], shape=(2,))
    y = pm.HalfNormal.dist(sigma=[0.1, 1, 10, 100], shape=(4,))
    return ifelse(cond, x, y)

with pm.Model() as m:
    cond = pm.Bernoulli("cond", p=0.5)
    y = pm.CustomDist("y", 1-cond, dist=ifelse_dist)

logp = m.compile_logp([y], sum=False)
print(logp({"cond": 0, "y": np.ones(2)}))  # [array([-2.91893853, -0.91893853])]
print(logp({"cond": 1, "y": np.ones(4)}))  # [array([-47.92320626,  -0.72579135,  -2.53337645,  -4.83101154])]
```

This kind of graphs can't be handled by switch mixtures, because that requires outputs to have the same shapes. 

This could one day be useful for variable size samplers... for now it's just here for completeness and libraries building on top of PyMC.

### DIFF
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -47,7 +47,7 @@ from pytensor.graph.rewriting.basic import (
     node_rewriter,
     pre_greedy_node_rewriter,
 )
-from pytensor.ifelse import ifelse
+from pytensor.ifelse import IfElse, ifelse
 from pytensor.scalar.basic import Switch
 from pytensor.tensor.basic import Join, MakeVector
 from pytensor.tensor.elemwise import Elemwise
@@ -73,10 +73,11 @@ from pymc.logprob.abstract import MeasurableVariable, _logprob, logprob
 from pymc.logprob.rewriting import (
     local_lift_DiracDelta,
     logprob_rewrites_db,
+    measurable_ir_rewrites_db,
     subtensor_ops,
 )
 from pymc.logprob.tensor import naive_bcast_rv_lift
-from pymc.logprob.utils import ignore_logprob
+from pymc.logprob.utils import ignore_logprob, ignore_logprob_multiple_vars
 
 
 def is_newaxis(x):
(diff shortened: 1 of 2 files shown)


pymc/logprob/mixture.py
```diff
@@ -483,3 +484,71 @@ def logprob_MixtureRV(
...
+    """Compute the log-likelihood graph for an `IfElse`."""
+    from pymc.pytensorf import replace_rvs_by_values
+
+    assert len(values) * 2 == len(base_rvs)
+
+    rvs_to_values_then = {then_rv: value for then_rv, value in zip(base_rvs[: len(values)], values)}
+    rvs_to_values_else = {else_rv: value for else_rv, value in zip(base_rvs[len(values) :], values)}
+
+    logps_then = [
+        logprob(rv_then, value, **kwargs) for rv_then, value in rvs_to_values_then.items()
```
lucianopaz, line 543: What’s the behaviour of logprob when the value variable doesn’t have the expected shape? Will it raise a ValueError? I think that would be bad in this case. Imagine if you have a step method on the condition variable. The stepper might choose to move the condition to an infeasible value and that would kill the sampling process. I would like the condition that doesn’t match shapes to simply return -inf logprob. That way the stepper would discard the proposal and stay in reasonable regions.

pymc/logprob/mixture.py
```diff
@@ -483,3 +484,71 @@ def logprob_MixtureRV(
...
+    logps_else = [
+        logprob(rv_else, value, **kwargs) for rv_else, value in rvs_to_values_else.items()
+    ]
+
+    # If the multiple variables depend on each other, we have to replace them
+    # by the respective values
+    logps_then = replace_rvs_by_values(logps_then, rvs_to_values=rvs_to_values_then)
+    logps_else = replace_rvs_by_values(logps_else, rvs_to_values=rvs_to_values_else)
+
+    return ifelse(if_var, logps_then, logps_else)
```
lucianopaz, line 554: Is the goal only to allow for explicit conditions in the mixture instead of marginalising?

lucianopaz (commented):
Nice approach @ricardoV94. I left a pair of comments. One thing that I found confusing were the first two commits. Are those really central to this PR or could they be PR in itself?

pymc/logprob/mixture.py
ricardoV94: The behavior of the logprob when the value variable doesn't match shape is exactly the same as the logprob of the underlying components. If it doesn't match, it will try to broadcast and fail if it cannot. Other than that we don't use size information in any of the core logprob functions.

`pm.logp(pm.Normal.dist(size=(4,)), np.ones((2,)))` will be happy to return a logp with two values.

One other case where this shows up is in graphs of the form `pt.ones((5,)) + pm.Normal.dist()` which we infer to have an equivalent logp as that of `pm.Normal.dist(shape=(5,))` even though the generative process contains only one true random variable, and not 5.

I think we need a bigger discussion about the role of shape information in the random and logp graphs, so I wouldn't treat `IfElse` differently for now.

pymc/logprob/mixture.py
ricardoV94: The applications may vary. We don't explicitly marginalize anything in the logprob submodule, but something like [MarginalModel](https://github.com/pymc-devs/pymc-experimental/blob/main/pymc_experimental/marginal_model.py) would marginalize this just fine if we can give it the logprob function that this PR offers.

ricardoV94:
> Nice approach @ricardoV94. I left a pair of comments. One thing that I found confusing were the first two commits. Are those really central to this PR or could they be PR in itself?

They could be in a PR by itself, but this one builds directly on that refactoring, so I would wait for it to get merged first if it was separate.

pymc/logprob/mixture.py
ricardoV94: Opened an issue here: https://github.com/pymc-devs/pymc/issues/6530

ricardoV94:
> One thing that I found confusing were the first two commits. Are those really central to this PR or could they be PR in itself?

These have now been merged elsewher

pymc/logprob/mixture.py
```diff
@@ -483,3 +484,71 @@ def logprob_MixtureRV(
...
+    rv_map_feature = getattr(fgraph, "preserve_rv_mappings", None)
+
+    if rv_map_feature is None:
+        return None  # pragma: no cover
+
+    if isinstance(node.op, MeasurableIfElse):
+        return None
+
+    # Check if all components are unvalued measuarable variables
+    if_var, *base_rvs = node.inputs
```
larryshamalama, line 507: Doesn't `base_rvs` always have length 2? Would it be better to write it as `[X_rv, Y_rv]` or something that shows the two components?

larryshamalama, line 507: Would it be of any benefit to check that the `base_rvs` have matching shapes? And/or their shapes are compatible with `if_var`?

Edit: Just saw the related conversation with @lucianopaz below. I'll still leave these comments here

pymc/logprob/mixture.py
larryshamalama, line 554: Does the `ifelse` ever apply within then/else branches? Maybe I'm mixing things up with `Switch`es...

One remark/question is that you seem to be computing the log-probability of each branch, then applying `ifelse`. Is that correct? Would it be possible to just compute the `logp` for _one_ (i.e. the selected) branch?

larryshamalama (commented):
Some more questions :)

pymc/logprob/mixture.py
ricardoV94: No, `base_rvs` can have arbitrarily `n*2` length. IfElse allows multiple outputs in both branches, but they have to be the same number

ricardoV94: `if_var` must always be a scalar (you may be thinking of switch). 

I guess you mean checking if the shape of the value variable matches the shape of the `base_rvs` from the evaluated branch. That would only make sense at evaluation time during the `logp`, as the different branches can have different shapes.

pymc/logprob/mixture.py
ricardoV94: `ifelse` in the C-backend is lazy, so it only evaluates one branch. In JAX and Numba it's not lazy but we will probably rework it to be in the future, as that is a big point of why ifelse exists in the first place.

Properly implemented, `ifelse` is supposed to only evaluate one branch: https://pytensor.readthedocs.io/en/latest/tutorial/conditions.html

pymc/logprob/mixture.py
larryshamalama: Ah! I effectively was getting confused with switch.

larryshamalama (approved):

PR #6531, opened 2023-02-17
Bump PyTensor dependency

ricardoV94:


twiecki (approved):

twiecki:
isn't this also in the precommit?

ricardoV94:
Yeah, I'm investigating it. It has a C implementation...

ricardoV94:
> isn't this also in the precommit?

I don't think it is anymore

ricardoV94:
The external samplers failure is what I feared would happen with the nutpie circular dependency?

ricardoV94:
The warning should be fixed by https://github.com/pymc-devs/pytensor/pull/225

twiecki:
Lots of install conflicts :(.

ricardoV94:
It's because nutpie pins the old pymc's pytensor dependency on conda. We can't test it directly as I suspected. Will remove the nutpie tests for now

tests/sampling/test_mcmc_external.py
```diff
@@ -17,16 +17,15 @@
...
-pytest.importorskip("numpyro")
-
 # turns all warnings into errors for this module
 pytestmark = pytest.mark.filterwarnings("error")
 
 
 @pytest.mark.parametrize("nuts_sampler", ["pymc", "nutpie", "blackjax", "numpyro"])
 def test_external_nuts_sampler(recwarn, nuts_sampler):
+    if nuts_sampler != "pymc":
+        pytest.importorskip(nuts_sampler)
```
ricardoV94, line 27: I checked this works as expected and skips only the parametrization that fails to import, not all of them afterwards. Can be seen in the CI as well: https://github.com/pymc-devs/pymc/actions/runs/4282342784/jobs/7456497043#step:8:79

Issue #6508, opened 2023-02-07
BUG: KaTeX parse error on Jupyter Notebook when using underscore, etc.


fanurs:
### Describe the issue:

_I apologize if this was an intended feature implemented to reflect the demand of the PyMC community. But when I migrate my codebase from PyMC3 to PyMC5, this seems more like a bug to me; it didn't raise an error in PyMC 3.8._

## Issue
When trying to display (not print) a PyMC model on Jupyter Notebook, I received a "KaTeX parse error". Apparently, this is due to the underscore character, `_`, in the name.

### Reproduceable code example:

```python
import pymc as pm
with pm.Model() as model:
    new_a = pm.Normal('new_a')
display(model) # on jupyter notebook
```

### Error message:

```shell
ParseError: KaTeX parse error: Expected 'EOF', got '_' at position 53: … \text{new_̲a} &\sim & \ope…
```

### PyMC version information:

`pip install pymc==5.0.2`

### Context for the issue:

Right now, I could bypass this error by adding a slash, `\_`, e.g. `new\_a` instead of `new_a`. But this is less desirable. For instance, I often have many variables that I would like to define. To do that, I would loop over some dictionary, and declare the variable one by one, while automatically assign them names that match the keys in the dictionary. However, with this "bug", I would either have to avoid using underscores at all or pre-process the names by replacing `_` with `\_`.

### More
Because of this parse error, I thought maybe I could write LaTeX syntax in my name field, or use raw string? But I don't think it works. Below are the things that I have tried:
- `pm.Normal(r'new_a')` - I though raw string would "protect" the underscore from getting parse error, but it still raises the same error.
- `pm.Normal('\alpha')` - Nope.
- `pm.Normal('\\alpha')` - Nope.
- `pm.Normal('$\alpha$')` - Nope.
- `pm.Normal('$\\alpha$')` - Nope.

covertg:
Curious -- what IDE or environment do you use for working in jupyter notebooks?

I've noticed this problem as well, for a while, in VSCode. While Jupyter's environments (notebook or lab) use MathJax for parsing equations, VSCode uses KaTeX.

Turns out there's a difference in how the two backends parse stuff in  `\text{...}`

covertg:
I believe we should be in the clear if we do a replacement as mentioned above--

If we encounter "_" within a \text command, end the \text command, drop in an escaped "_" outside of the \text, then start a new \text with the remaining string.

[...]

PR #6533, opened 2023-02-18
Refactor pretty-printing and fix latex underscores

covertg:
**What is this PR about?**
`printing.py` had developed a decent amount of duplicated code.  I wanted to work on #6508 but found it difficult to parse through the logic. Through the course of understanding how the submodule works, I ended up refactoring it somewhat.

This PR fixes the LaTeX underscore bug as discussed in that issue and also seeks to clean up pretty-printing. In particular,
- Addresses #6508 by escaping \text commands to print underscores
- Unifies `str_for_model_var` and `str_for_potential_or_deterministic`
- Deprecates the unused `include_params` argument
  - The codebase seems to not use this argument at all, and removing it simplifies the logic of some functions in `printing.py` as well as in tests. But I could re-incorporate it if preferred.
- Parses the value of the `formatting` arg a bit more strictly
- Changes the printing of constants to include shape information, e.g. `<constant (1,2)>`
- Updates tests to reflect changes
- Adds a handful of other tests for printing functionality: model names, names with edge case characters (underscore, $, ~), and misc tests for the warning and errors that may be thrown by pretty printing

**Checklist**

## Major / Breaking Changes
- Passing `include_params` to the str_repr function now raises a futurewarning.
- The `formatting` argument is now expected to be exactly either "plain" or "latex".

## New features
- Constant shape information as discussed. (Sorry, issue creep. It's a one-liner now. But I can also leave this for a separate issue if preferred.)
- Tildes (~) in user-specified strings are converted to dashes (-). Practically this is because `str_for_model` uses ~ and `\sim` to align the model str, and it's annoying to typeset a plain tilde character in LaTeX.

## Bugfixes
- #6508
- Also fixed a bug (edge case) where AttributeError was thrown when a Potential received some Variable that didn't have the `.owner` attribute. (*I'm not sure whether this would ever happen in the wild, but model-building allows for this case, so I went ahead and included it.*)

## Documentation
- ...

## Maintenance
- ~~While working on this I noticed that the `_repr_latex_()` methods assigned to Distributions and Models are essentially the same as their `str_repr` functions, just with latex enforced. Nowhere in the pymc codebase uses these methods. Candidate for removal?~~ I've realized that this method is necessary for latex to automagically appear when supported by ipython.
- ~~The inputs to Potentials and Deterministics have been printed as a function-like expression, e.g. `Deterministic(f(arg1, ...))`. In the updated submodule, we maintain this behavior, but it would be easy to drop the `f()` wrapper for this case if wanted.~~ On reflection it seems pretty sensible to keep the f().
- ~~I also noticed that we don't have any tests for printing the model name.~~ Added some tests involving this. This includes the case for nested models (one pm.Model() created within the context of another). I've never done this myself in pymc so could use a sanity check on if we need/want to be testing this.

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -197,7 +197,7 @@ class SymbolicRandomVariable(OpFromGraph):
     """
 
     _print_name: Tuple[str, str] = ("Unknown", "\\operatorname{Unknown}")
-    """Tuple of (name, latex name) used for for pretty-printing variables of this type"""
+    "Tuple of (name, latex name) used for for pretty-printing variables of this type"
 
     def __init__(self, *args, ndim_supp, **kwargs):
         self.ndim_supp = ndim_supp
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -57,7 +57,6 @@ from pymc.logprob.abstract import (
 )
 from pymc.logprob.rewriting import logprob_rewrites_db
 from pymc.model import BlockModelAccess
-from pymc.printing import str_for_dist
 from pymc.pytensorf import collect_default_updates, convert_observed_data
 from pymc.util import UNSET, _add_future_warning_tag
 from pymc.vartypes import string_types
@@ -320,9 +319,12 @@ class Distribution(metaclass=DistributionMeta):
         )
 
         # add in pretty-printing support
-        rv_out.str_repr = types.MethodType(str_for_dist, rv_out)
+        from pymc.printing import str_for_model_var
+
+        rv_out.str_repr = types.MethodType(str_for_model_var, rv_out)
         rv_out._repr_latex_ = types.MethodType(
-            functools.partial(str_for_dist, formatting="latex"), rv_out
+            functools.partial(str_for_model_var, formatting="latex"),
+            rv_out,
         )
 
         rv_out.logp = _make_nice_attr_error("rv.logp(x)", "pm.logp(rv, x)")
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -1116,7 +1116,7 @@ class _LKJCholeskyCovBaseRV(RandomVariable):
     ndim_supp = 1
     ndims_params = [0, 0, 1]
     dtype = "floatX"
-    _print_name = ("_lkjcholeskycovbase", "\\operatorname{_lkjcholeskycovbase}")
+    _print_name = ("_lkjcholeskycovbase", r"\operatorname{\_lkjcholeskycovbase}")
 
     def make_node(self, rng, size, dtype, n, eta, D):
         n = at.as_tensor_variable(n)
@@ -1164,7 +1164,7 @@ _ljk_cholesky_cov_base = _LKJCholeskyCovBaseRV()
 # be safely resized. Because of this, we add the thin SymbolicRandomVariable wrapper
 class _LKJCholeskyCovRV(SymbolicRandomVariable):
     default_output = 1
-    _print_name = ("_lkjcholeskycov", "\\operatorname{_lkjcholeskycov}")
+    _print_name = ("_lkjcholeskycov", r"\operatorname{\_lkjcholeskycov}")
 
     def update(self, node):
         return {node.inputs[0]: node.outputs[0]}
--- a/pymc/printing.py
+++ b/pymc/printing.py
@@ -148,7 +148,7 @@ def _get_varname_distname_args(
         dist_args = _walk_expression_args(var)
     args_str = _str_for_args_list(dist_args, formatting=formatting)
     if _is_potential_or_deterministic(var):
-        args_str = f"f({args_str})"  # TODO do we still want to do this?
+        args_str = f"f({args_str})"
 
     # These three strings are now formatted according to `formatting`. If latex, they
     # just ultimately need to be wrapped in $.
--- a/pymc/printing.py
+++ b/pymc/printing.py
@@ -12,9 +12,11 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+import functools
 import itertools
+import warnings
 
-from typing import Union
+from typing import List, Optional, Tuple, Union
 
 from pytensor.compile import SharedVariable
 from pytensor.graph.basic import Constant, walk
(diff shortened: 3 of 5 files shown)


covertg:
Forgot to mention: As a qualitative test, I also checked that the "monolithic" model in test_printing.py renders correctly in both MathJax and KaTeX.

pymc/printing.py
```diff
@@ -130,148 +93,176 @@ def str_for_model(model: Model, formatting: str = "plain", include_params: bool
...
-        var_type = "shared"
-
-    if len(var_data.shape) == 0:
-        return f"{var_data:.3g}"
-    elif len(var_data.shape) == 1 and var_data.shape[0] == 1:
-        return f"{var_data[0]:.3g}"
-    elif "latex" in formatting:
-        return rf"\text{{<{var_type}>}}"
+        # Recurse
+        return _str_for_input_var(var.owner.inputs[0], formatting=formatting)
```
covertg: I carried this over from the previous version.
When would we receive a DimShuffle here, and can we be sure that only the first input is relevant?

covertg:
I pushed some changes. Mainly, they clarify the exceptions that are thrown, add a number of printing tests, and fix an edge-case bug or two. I've also updated the original post to reflect the latest. I won't be changing this PR any more till review. (though there are a few questions or comments that I will add to the code.)

pymc/printing.py
```diff
@@ -130,148 +102,194 @@ def str_for_model(model: Model, formatting: str = "plain", include_params: bool
...
+            var_type = "constant"
+        else:
+            var_data = var.get_value()
+            var_type = "shared"
+        if var_data.size == 1:
+            return f"{var_data.flatten()[0]:.3g}"
+        else:
+            return f"<{var_type} {var_data.shape}>"
+    elif _has_owner(var):
+        if isinstance(var.owner.op, DimShuffle):
```
covertg, line 176: Still have this question. Is the DimShuffle Op case still one that we need to handle? If so, let's add a test for it.

pymc/tests/test_printing.py
```diff
@@ -27,56 +30,52 @@
...
-        Y = alpha + X.dot(beta) + np.random.randn(size) * sigma
+        Y = alpha0 + X.dot(beta0) + np.random.randn(size) * sigma0
         with Model() as self.model:
             # TODO: some variables commented out here as they're not working properly
             # in v4 yet (9-jul-2021), so doesn't make sense to test str/latex for them
 
             # Priors for unknown model parameters
             alpha = Normal("alpha", mu=0, sigma=10)
-            b = Normal("beta", mu=0, sigma=10, size=(2,), observed=beta)
+            # TODO why is this observed?
```
covertg, line 77: Shouldn't be a huge deal but it seems curious that b/beta has been observed

pymc/tests/test_printing.py
```diff
@@ -27,56 +30,52 @@
...
         size = 100
-
         # Predictor variable
         X = np.random.normal(size=(size, 2)).dot(np.array([[1, 0], [0, 0.2]]))
-
         # Simulate outcome variable
-        Y = alpha + X.dot(beta) + np.random.randn(size) * sigma
+        Y = alpha0 + X.dot(beta0) + np.random.randn(size) * sigma0
         with Model() as self.model:
             # TODO: some variables commented out here as they're not working properly
```
covertg, line 72: Is this still a TODO or is it outdated?

larryshamalama:
Hi @covertg, sorry for the delay in writing back and thanks for your PR :) I'm currently travelling, but I intend to review your PR within the next few days.

pymc/distributions/distribution.py
```diff
@@ -197,7 +197,7 @@ class SymbolicRandomVariable(OpFromGraph):
     """
 
     _print_name: Tuple[str, str] = ("Unknown", "\\operatorname{Unknown}")
-    """Tuple of (name, latex name) used for for pretty-printing variables of this type"""
+    "Tuple of (name, latex name) used for for pretty-printing variables of this type"
```
larryshamalama, line 200: Any specific reason for this change?

pymc/tests/test_printing.py
larryshamalama: Hmm I've never noticed this. I also don't think that we should have `beta` to be observed? It does not seem to affect the test nor its purpose though... Curious indeed

pymc/tests/test_printing.py
```diff
@@ -27,56 +30,52 @@
...
+        assert self.model._repr_latex_() == self.model.str_repr(formatting="latex")
 
 
 class TestMonolith(BaseTestStrAndLatexRepr):
     def setup_class(self):
         # True parameter values
-        alpha, sigma = 1, 1
-        beta = [1, 2.5]
-
+        alpha0, sigma0, beta0 = 1, 1, [1, 2.5]
```
larryshamalama, line 64: maybe `alpha_true, sigma_true, beta_true`? This notation is also okay

pymc/printing.py
```diff
@@ -130,148 +102,194 @@ def str_for_model(model: Model, formatting: str = "plain", include_params: bool
...
+    if command == "text":
+        text = text.replace("_", rf"}}\_\{command}{{")
+    return text
+
+
+def _is_potential_or_deterministic(var: Variable) -> bool:
+    # This is a bit hacky but seems like the best we got
+    if (
+        hasattr(var, "str_repr")
+        and callable(var.str_repr)
```
larryshamalama, line 261: if `hasattr(var, "str_repr")`, it must necessarily be callable, right?

pymc/distributions/multivariate.py
```diff
@@ -1116,7 +1116,7 @@ class _LKJCholeskyCovBaseRV(RandomVariable):
     ndim_supp = 1
     ndims_params = [0, 0, 1]
     dtype = "floatX"
-    _print_name = ("_lkjcholeskycovbase", "\\operatorname{_lkjcholeskycovbase}")
+    _print_name = ("_lkjcholeskycovbase", r"\operatorname{\_lkjcholeskycovbase}")
```
larryshamalama, line 1119: Why the addition of `\` in front of `_`? All the other random variables follow `("rv_name", "\\operatorname{rv_name}")`. Without the `\`, do you obtain a strange looking string?

pymc/distributions/multivariate.py
```diff
@@ -1164,7 +1164,7 @@ def rng_fn(self, rng, n, eta, D, size):
 # be safely resized. Because of this, we add the thin SymbolicRandomVariable wrapper
 class _LKJCholeskyCovRV(SymbolicRandomVariable):
     default_output = 1
-    _print_name = ("_lkjcholeskycov", "\\operatorname{_lkjcholeskycov}")
+    _print_name = ("_lkjcholeskycov", r"\operatorname{\_lkjcholeskycov}")
```
larryshamalama, line 1167: Same comment as above

pymc/printing.py
larryshamalama: I think so. My understanding is that this is to retrieve the original RV that has been resized (or `DimShuffle`d).

pymc/printing.py
```diff
@@ -26,91 +28,61 @@
...
-        dist_name = (
-            dist.owner.op._print_name[0] if hasattr(dist.owner.op, "_print_name") else "Unknown"
+    var_name, dist_name, args_str = _get_varname_distname_args(
+        var, formatting=formatting, dist_name=dist_name
+    )
+    if "include_params" in kwargs:
+        warnings.warn(
+            "The `include_params` argument has been deprecated. Future versions will always include parameter values.",
+            FutureWarning,
+            stacklevel=2,
```
larryshamalama, line 57: Your point of depreciating `include_params` seems valid to me. If IIRC when browsing this file a few weeks or months ago, `include_params` was used to delineate whether they were printed as arguments to other distributions or not. You seem to have addressed this in `_get_varname_distname_args`.

pymc/printing.py
```diff
@@ -130,148 +102,194 @@ def str_for_model(model: Model, formatting: str = "plain", include_params: bool
...
+    # engines such as KaTeX behave differently and expect certain characters to be
+    # escaped to be typeset as desired. We work around this by escaping out of the
+    # command itself, writing the character, then continuing on with the same command.
+    if command == "text":
+        text = text.replace("_", rf"}}\_\{command}{{")
+    return text
+
+
+def _is_potential_or_deterministic(var: Variable) -> bool:
+    # This is a bit hacky but seems like the best we got
```
larryshamalama, line 258: Interesting function 😅. I never thought of checking if a variable is created via Potential or Deterministic this way. If it works, this sounds great but, as you suggest, it does look hackish...

(Previous `_is_potential_or_deterministic` below for reference)

```python
def _is_potential_or_deterministic(var: Variable) -> bool:
      try:
          return var.str_repr.__func__.func is str_for_potential_or_deterministic
      except AttributeError:
          # in case other code overrides str_repr, fallback
          return False
```

larryshamalama (commented):
Thanks again for your PR @covertg and for your patience. It sounds like you've put in a lot of work into this, especially for a heavily utilized functionality of PyMC that is often not a focal point of development.

Overall, this submodule is due for a heavy makeover, as you can tell and have started. I provided some comments and questions but, overall, a dispatch-based approach for printing is probably going to be ultimately the best solution. Evidence for this is that there are a lot of if/else statements going on; a more general solution would be to implement a printing mechanism for each Op. That way, (I don't think that?) we would have to delineate between potentials, deterministics, etc. See some references below. I previously attempted to refactor the whole module in #6447, but I finally decided against it.

The PR is still good. Even if we decide to immediately go for a dispatched-based approach, not all effort will be lost, but I personally think that what you have is a good start. I will need to go over the code again before merging. I'm happy to bounce back some iterations if you have questions or concerns. I can even set aside time if you would like to do some pair programming on this work :)

[PR 6447](https://github.com/pymc-devs/pymc/pull/6447)
[Discussion 6261](https://github.com/pymc-devs/pymc/discussions/6261)
[Issue 6311](https://github.com/pymc-devs/pymc/issues/6311)

pymc/printing.py
```diff
@@ -130,148 +102,194 @@ def str_for_model(model: Model, formatting: str = "plain", include_params: bool
...
+    # in the \text block will be rendered more or less verbatim. Meanwhile, other
+    # engines such as KaTeX behave differently and expect certain characters to be
+    # escaped to be typeset as desired. We work around this by escaping out of the
+    # command itself, writing the character, then continuing on with the same command.
+    if command == "text":
+        text = text.replace("_", rf"}}\_\{command}{{")
+    return text
+
+
+def _is_potential_or_deterministic(var: Variable) -> bool:
```
ricardoV94, line 257: Can we use the model object for this? 

`if var in model.deterministics or var in model.potentials`, 

Just pass the model variable along

pymc/printing.py
```diff
@@ -26,91 +28,61 @@
...
+def str_for_model(model: Model, formatting: str = "plain", **kwargs) -> str:
+    """Make a human-readable string representation of Model including all random
+    variables and their distributions.
+    """
     all_rv = itertools.chain(model.unobserved_RVs, model.observed_RVs, model.potentials)
-
-    rv_reprs = [rv.str_repr(formatting=formatting, include_params=include_params) for rv in all_rv]
-    rv_reprs = [rv_repr for rv_repr in rv_reprs if "TransformedDistribution()" not in rv_repr]
-
+    rv_reprs = [rv.str_repr(formatting=formatting, **kwargs) for rv in all_rv]
```
ricardoV94, line 81: I know it's not the point of this PR but can we already remove the assumption that model RVs have this `str_repr` attached to it? 

Just call the respective function `str_for_dist` or `str_for_deterministic` on the RV

It would fix https://github.com/pymc-devs/pymc/issues/6311

pymc/distributions/multivariate.py
larryshamalama: Oops, thinking about this again... the answer is probably in the title of this PR and related to issue #6508

covertg: Yep exactly!

pymc/printing.py
covertg: Interesting, I hadn't been aware of that issue. I don't think that we can quite yet remove that assumption, essentially because right now the str_repr function (`str_for_model_var` in this PR, or `str_for_dist` and `str_for_deterministic` in mainline) doesn't know how to distinguish between deterministics and potentials.

At the moment that info is inserted using `functools.partial` whenever the Model registers a new deterministic or potential. But, related to the above comment: currently we don't require that a model object is on the context stack in order to pretty-print a model variable. Requiring a model object would simplify this though

pymc/printing.py
covertg: Yep, I know I know....... :sweat_smile: 
I figured that it's not that far removed from the original implementation, since in the big-picture they're both inspecting whether a var's `str_repr` is a functools partial that was made in a certain way.

Anyways, a dispatch-based approach or at least requiring a model object would obviate this type of approach

pymc/printing.py
covertg: I see, cool. What would be an example of a model where the original RV gets resized or dimshuffled? (@ricardoV94 )

pymc/distributions/distribution.py
covertg: Ah, sorry - Refreshing on [pep 257](https://peps.python.org/pep-0257/) I've realized triple quotes are probably preferred!

pymc/printing.py
ricardoV94: Yes it would be fine to require a model to be on the context for the fancier pretty printing.

We can have a much more limited pretty printing for standalone variables (or none at all)

covertg:
@larryshamalama of course, thank you for the review! I understand this PR ballooned a good deal and also recognize that I don't have the full big-picture vision that y'all have. My bad for missing your previous work on the dispatch-based approach.

So thanks for the comments. Did my best to follow up on the specific ones that y'all have left. Moving forward -- I'd be very glad to help work on this in whatever way would best serve pymc in the long run. It'd be fun to work on that together! If I may ask, what led you to decide against the work in #6447?

Also a big-picture question for us. Do we want model *graph*ing to remain separate from model *print*ing? This came to me when considering the question (in comments above) of, do we need to pretty-print pymc variables without a model on the context stack. If the focus is on pretty-printing a full model, it seems like graph-visualizing and model-pretty-printing could possibly be unified. But perhaps that is its own big can of worms.

ricardoV94:
> Also a big-picture question for us. Do we want model _graph_ing to remain separate from model _print_ing? This came to me when considering the question (in comments above) of, do we need to pretty-print pymc variables without a model on the context stack. If the focus is on pretty-printing a full model, it seems like graph-visualizing and model-pretty-printing could possibly be unified. But perhaps that is its own big can of worms.

The big value of the pretty-printing is on printing the whole Model, not individual variables, so I would say that we don't need try to hard to pretty-print individual variables.

ricardoV94:
@covertg Any interest in picking up on this again? I did a small fix on https://github.com/pymc-devs/pymc/pull/6942 but the underscore thing is still broken.

PR #6540, opened 2023-02-21
Tests are no longer part of the package

### LINKED ISSUE #6526: Move the tests directory outside of the code directory
### Discussed in https://github.com/pymc-devs/pymc/discussions/6172

<sup>Originally posted by **Armavica** October  2, 2022</sup>
Hi,
With the reorganization of the tests complete (#5777), a last question remains: where to place the `tests` directory. It currently resides inside the `pymc` code directory:

```
pymc/
├── benchmarks/
├── docs/
├── pymc/
│   ├── __init__.py
│   ├── model.py
│   ├── ...
│   └── tests/
└── scripts/
```

but one could argue that it would make more sense to put it at the same level as `benchmarks/`, `docs/` and `pymc/`:

```
pymc/
├── benchmarks/
├── docs/
├── pymc/
│   ├── __init__.py
│   ├── model.py
│   └── ...
├── scripts/
└── tests/
```

This is what is recommended by the [official Python packaging guide](https://packaging.python.org/en/latest/tutorials/packaging-projects/), by the [Pytest good practices guide](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code), the [Poetry package management tool](https://python-poetry.org/docs/basic-usage/) as well as by [other guides](https://docs.python-guide.org/writing/structure/).
However, I have to note that `numpy`, `scipy`, and `pandas` all have `tests` inside the code directory.

Besides the arguments advanced by the [Pytest good practices guide](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code), I can add that I often `grep` the source code directory to see where or how a function or class is used or defined. With the tests directory where it is currently, a lot of noise is often returned from the tests.

Cost of the transition: probably limited to some `import`s to change in the test files to reflect the move.

What is your opinion on this?</div>

Armavica:
Fixes #6526, as discussed in #6172.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- Move the tests outside of the code directory.


tests/backends/fixtures.py
```diff
@@ -22,7 +22,8 @@
 import pytest
 
 from pymc.backends import base
-from pymc.tests import models
+
+from .. import models
```
ricardoV94: You can still use absolute paths no? like `from tests import models`

That would be more readable.

ricardoV94:
Don't know if we should label this as major, for those people who run tests when they install the package.

tests/backends/fixtures.py
Armavica: I was worried about `tests` conflicting with a Python package, but you are right, that should be possible.

ricardoV94 (approved):

ricardoV94:
A conflict seems to have cropped up, looks good to me otherwise

ricardoV94:
@OriolAbril is the RTD failure new? It's not the usual timeout that was going on for a while

OriolAbril:
The rtd error says it is due to an error when executing tge posterior predictive notebook, specifically in the pm.get_data line

Armavica:
> The rtd error says it is due to an error when executing tge posterior predictive notebook, specifically in the pm.get_data line

Indeed, thank you! I found the mistake.

Issue #6550, opened 2023-02-28
BUG: sample_blackjax_nuts incorrectly reports sampling time


dehorsley:
### Describe the issue:

For larger models, `sample_blackjax_nuts` reports (unfortunately) unrealistically small sampling time and a very large transformation time. This is due to JAX's [Asynchronous dispatch](https://jax.readthedocs.io/en/latest/async_dispatch.html#async-dispatch) model. I'd guess what's currently reported as "sampling time" is really JAX compilation. It should be able to be fixed with a simple `block_until_ready()`. I can send a PR shortly.

Mostly a cosmetic issue, but it caused me some time trying to work out why transformation was so slow 😄.

### Reproduceable code example:

```python
import pymc as pm
from pymc.sampling.jax import sample_blackjax_nuts
import numpy as np

x = np.random.normal(0, 1, 1_000_000)
with pm.Model() as model:
    sigma = pm.HalfCauchy("sigma", 1)
    mu = pm.Normal("mu", 0, 1)
    pm.Normal("x", mu, sigma, observed=x)
    trace_blackjax = sample_blackjax_nuts()
```

### Error message:

```shell
Compiling...
Compilation time =  0:00:02.020721
Sampling...
Sampling time =  0:00:09.755916
Transforming variables...
Transformation time =  0:02:29.152843
```

### PyMC version information:

macOS

```
pytensor            2.9.1
blackjax            0.9.6
pymc                5.0.2+58.g33d641db
jax                 0.4.1
jaxlib              0.4.1

```

fonnesbeck:
Thanks for tracking this down!

ricardoV94:
I'm not sure we want to block for the sake of accurate timings. 

By the way, the same might happen with numpyro, when progressbar = False

PR #6551, opened 2023-02-28
Block `sample_blackjax_nuts` until sampling finished (#6550)

dehorsley:
Due to JAX's asynchronous dispatch model,`sample_blackjax_nuts` sampling time was being underestimated and moved into transformation time. This blocks PR blocks for sampling to complete before calculating the time taken.

## Bugfixes
- `sample_blackjax_nuts` correctly report split between sampling and transformation. (#6550)

### DIFF
--- a/pymc/sampling/jax.py
+++ b/pymc/sampling/jax.py
@@ -438,7 +438,8 @@ def sample_blackjax_nuts(
         )
 
     states, stats = map_fn(get_posterior_samples)(keys, init_params)
-    raw_mcmc_samples = states.position
+    # block to allow accurate sampling time calculation
+    raw_mcmc_samples = jax.block_until_ready(states.position)
     potential_energy = states.potential_energy
     tic3 = datetime.now()
     print("Sampling time = ", tic3 - tic2, file=sys.stdout)


twiecki (approved):

ricardoV94:
I'm not sure we want to block for the sake of accurate timings.

By the way, the same might happen with numpyro, when progressbar = False

dehorsley:
> I'm not sure we want to block for the sake of accurate timings.

In my admittedly simple test, I didn't see any change in the total time taken for sampling+transformation. However, theoretically there could be a slight advantage in the transformation function getting compiled in parallel. 

But perhaps your scepticism is for a different reason? I suppose an alternative is to not make the distinction between sampling and transformation time. 

> By the way, the same might happen with numpyro, when progressbar = False

You're right [MCMC.run](https://num.pyro.ai/en/latest/mcmc.html#numpyro.infer.mcmc.MCMC.run) even has a note in the docs to that effect.

michaelosthege:
@twiecki @ricardoV94 can you make a decision on this?

ricardoV94:
Superseeded by #6963 

We decided to add the block_until_ready. Thanks anyway @dehorsley

PR #6552, opened 2023-03-01
Exclude tests from package discovery

Armavica:
`setuptools` was now picking up the `tests` directory as an installable package and was installing it alongside `pymc`, which could conflict with any other package named `tests`.

This PR configures `setuptools` to exclude `tests` from the automated package discovery.

I took this opportunity to clean up some warning-generating lines in `MANIFEST.in`, and an unused option in `setup.py`.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- Exclude tests from the package discovery

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ if __name__ == "__main__":
         url=URL,
         long_description=LONG_DESCRIPTION,
         long_description_content_type="text/x-rst",
-        packages=find_packages(),
+        packages=find_packages(exclude=["tests*"]),
         # because of an upload-size limit by PyPI, we're temporarily removing docs from the tarball.
         # Also see MANIFEST.in
         # package_data={'docs': ['*']},
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,6 @@ if __name__ == "__main__":
         # because of an upload-size limit by PyPI, we're temporarily removing docs from the tarball.
         # Also see MANIFEST.in
         # package_data={'docs': ['*']},
-        include_package_data=True,
         classifiers=classifiers,
         python_requires=">=3.8",
         install_requires=install_reqs,
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,3 @@
-recursive-include tests/data *
-recursive-include source *
-# because of an upload-size limit by PyPI, we're temporarily removing docs from the tarball:
-recursive-exclude docs *
 include requirements.txt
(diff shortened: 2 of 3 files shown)


ricardoV94:
Can we just move the relevant test utils `TestMatchesScipy` into the pymc package and leave tests outside? Wondering if that could be easier?

Armavica:
> Can we just move the relevant test utils `TestMatchesScipy` into the pymc package like `numpy.testing` module and leave tests outside? Wondering if that could be easier?

Yeah, much easier indeed. I will do that then, thanks :)

Armavica:
I repurposed this PR to only fix the spurious `tests` package installation so that we can re-release quickly.

ricardoV94:
Might have broken ReadTheDocs?

OriolAbril:
Same issue on `test_scores = pd.read_csv(pm.get_data("test_scores.csv"), index_col=0)` line as in the other PR that moved tests

Armavica:
I think I fixed it (pending confirmation). The problem was that `pm.get_data` was first looking inside `tests/data`, which used to contain example files, I suppose. This directory doesn't exist anymore (I don't know since when), so I believe that the fallback (downloading from github `pymc-examples` is always activated. So I just removed the local check.

setup.py
```diff
@@ -65,11 +65,10 @@
         url=URL,
         long_description=LONG_DESCRIPTION,
         long_description_content_type="text/x-rst",
-        packages=find_packages(),
+        packages=find_packages(exclude=["tests*"]),
         # because of an upload-size limit by PyPI, we're temporarily removing docs from the tarball.
         # Also see MANIFEST.in
         # package_data={'docs': ['*']},
-        include_package_data=True,
```
twiecki, line 72: Doesn't this exclude other stuff like example NBs or csv files?

Armavica: There is no more csv files in `pymc` (I think they moved to `pymc-examples`), and if NBs were ever included, I don't think they are not anymore. As far as I can say, building `pymc` with or without this line creates the same wheel.

twiecki (approved):

twiecki:
Thanks for the quick turn-around @Armavica! Should we release 5.1.1 now?

ricardoV94:
@twiecki let's add back the test utilities that are used by other packages first. I'll push a PR now

Issue #6107, opened 2022-09-07
Consider removing old shape handling functions


ricardoV94:
Now that random draws are done by Aesara, we might not need some/all of these anymore:

https://github.com/pymc-devs/pymc/blob/761f77dad29116650ef3ca043645042219654a5c/pymc/distributions/shape_utils.py#L38-L42

First step would be to check which ones are still used and where.

Armavica:
As far as I can say,

* `shapes_broadcasting` is only used in `broadcast_dist_samples_shape` and `get_broadcastable_dist_samples`,
* `broadcast_dist_samples_shape` is only used in `get_broadcastable_dist_samples`,
* `get_broadcastable_dist_samples` is only used in `broadcast_distribution_samples` and `broadcast_dist_samples_to`,
* `broadcast_distribution_samples` is not used anywhere,
* `broadcast_dist_samples_to` is the only of these functions used somewhere else in the code, and only in `MatrixNormalRV`.

ricardoV94:
So we should check if we can get something more immediate to replace `broadcast_dist_samples_to` in `MatrixNormalRV`? I haven't looked at what it is doing

mattiadg:
Is this still an open issue? I am interested in working on it

thomasaarholt:
This issue can be closed, all the functions mentioned in OP are removed.

ricardoV94:
> This issue can be closed, all the functions mentioned in OP are removed.

Thanks!

PR #6556, opened 2023-03-02
Remove shape functions

symeneses:
**What is this PR about?**
It solves #6107 

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- Functions deleted from the API docs.

## Maintenance
- ...


pymc/distributions/multivariate.py
```diff
@@ -1651,7 +1650,9 @@ def rng_fn(cls, rng, mu, rowchol, colchol, size=None):
         output_shape = size + dist_shape
 
         # Broadcasting all parameters
-        (mu,) = broadcast_dist_samples_to(to_shape=output_shape, samples=[mu], size=size)
+        shapes = [mu.shape, output_shape]
+        sp_shapes = [s[len(size) :] if size == s[: min([len(size), len(s)])] else s for s in shapes]
```
michaelosthege: Did you test this by hand?
I'd feel better it this line was covered by tests, or refactored to be easier to comprehend (or both)

michaelosthege (commented):
Wow, this must be a record ratio of lines deleted vs. lines added. Awesome!

pymc/distributions/multivariate.py
symeneses: To test that we are getting the correct broadcastable shape, I think the best option is to keep the function `broadcast_dist_samples_shape` and its tests. `shapes_broadcasting` is replaced by `np.broadcast_shapes`. PR updated.
Also, more tests may be needed, using only what I added here was not correct but all tests were passing.

pymc/distributions/multivariate.py
```diff
@@ -1651,7 +1651,9 @@ def rng_fn(cls, rng, mu, rowchol, colchol, size=None):
         output_shape = size + dist_shape
 
         # Broadcasting all parameters
-        (mu,) = broadcast_dist_samples_to(to_shape=output_shape, samples=[mu], size=size)
+        shapes = [mu.shape, output_shape]
+        broadcastable_shape = broadcast_dist_samples_shape(shapes, size=size)
+        mu = np.broadcast_to(mu, shape=broadcastable_shape)
```
ricardoV94, line 1656: Is `mu = np.broadcast_to(mu, shape=dist_shape)` incorrect?

ricardoV94:
You may be able to use this utility to broadcast everything: https://github.com/pymc-devs/pytensor/blob/63f8d6e732cd19e7f8940e25e4c44231e7765f96/pytensor/tensor/random/utils.py#L24

Although I think this RV was never properly implemented as a RandomVariable. For instance if `size=None`, the size should be given by the batch dimensions of the parameters. Also `dist_shape` should use `rowcol.shape[-1] or [-2]` not `[0]`

michaelosthege (approved):
LGTM

ricardoV94:
This PR may need a rebase before we can merge

ricardoV94:
We can rebase and merge this one as is of it's working and keep cleaning it up in future PRs

twiecki (approved):

ricardoV94:
Opened an issue for the MatrixNormal, which I am not very confident is (or was) working correctly. Would have to check what actual tests there are. Otherwise the changes seem sensible. 

Thanks @symeneses

symeneses:
> Opened an issue for the MatrixNormal, #6066 which I am not very confident is (or was) working correctly. Would have to check what actual tests there are. Otherwise the changes seem sensible.
> 
> Thanks @symeneses

#6066 is a PR not an issue. I could have a look to the tests based on your previous comment and in the new issue.

PR #6573, opened 2023-03-07
Fix bug in transformed scan variables

ricardoV94:
This would happen when scans have different outputs with mixed transforms (e.g., None and LogOddsTransform), as it was using only the transform of the last variable, due to a typo

### DIFF
--- a/tests/logprob/test_transforms.py
+++ b/tests/logprob/test_transforms.py
@@ -914,8 +914,13 @@ def test_scan_transform():
     init = at.random.beta(1, 1, name="init")
     init_vv = init.clone()
 
+    def scan_step(prev_innov):
+        next_innov = at.random.beta(prev_innov * 10, (1 - prev_innov) * 10)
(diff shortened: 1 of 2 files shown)


tests/logprob/test_transforms.py
```diff
@@ -914,8 +914,13 @@ def test_scan_transform():
     init = at.random.beta(1, 1, name="init")
     init_vv = init.clone()
 
+    def scan_step(prev_innov):
+        next_innov = at.random.beta(prev_innov * 10, (1 - prev_innov) * 10)
+        update = {next_innov.owner.inputs[0]: next_innov.owner.outputs[0]}
+        return next_innov, update
```
ricardoV94, line 920: Adding more (non-probability related) outputs to scan showcases the bug.

Updates are a common output in real scan uses, so this is both more realistic and covers the fixed bug

twiecki (approved):

PR #6582, opened 2023-03-08
Extend logprob inference for scans with carried auxiliary states

ricardoV94:
PyMC can now infer graphs like MA(2)

```python
import numpy as np
import pytensor

import pymc as pm

init_eps = pm.DiracDelta.dist(np.zeros(2, dtype="float64"))
rho = [0.3, 0.9]
sigma = 1
n_steps = 100

def ma2_step(eps_tm2, eps_tm1, rho, sigma):
    mu = eps_tm1 * rho[0] + eps_tm2 * rho[1]
    y = pm.Normal.dist(mu, sigma)
    eps = y - mu
    return eps, y

[_, ma2], _ = pytensor.scan(
    fn=ma2_step,
    outputs_info=[{"initial": init_eps, "taps": range(-2, 0)}, None],
    non_sequences=[rho, sigma],
    n_steps=n_steps,
    name="ma2",
)

ma2_test = ma2.eval()
pm.logp(ma2, ma2_test).sum().eval()
```

CC @jessegrabowski

### DIFF
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -24,7 +24,8 @@ import pytest
 from numpy import random as nr
 from numpy import testing as npt
 from pytensor.compile.mode import Mode
-from pytensor.graph.basic import ancestors
+from pytensor.graph.basic import walk
+from pytensor.graph.op import HasInnerGraph
 from pytensor.graph.rewriting.basic import in2out
 from pytensor.tensor import TensorVariable
 from pytensor.tensor.random.op import RandomVariable
@@ -37,7 +38,7 @@ from pymc import Distribution, logcdf, logp
 from pymc.distributions.shape_utils import change_dist_size
 from pymc.initial_point import make_initial_point_fn
 from pymc.logprob import joint_logp
-from pymc.logprob.abstract import icdf
+from pymc.logprob.abstract import MeasurableVariable, icdf
 from pymc.logprob.utils import ParameterValueError
 from pymc.pytensorf import (
     compile_pymc,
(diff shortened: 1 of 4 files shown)


pymc/testing.py
```diff
@@ -958,5 +959,18 @@ def seeded_numpy_distribution_builder(dist_name: str) -> Callable:
...
+        if owner:
+            inputs = list(reversed(owner.inputs))
+
+            if isinstance(owner.op, HasInnerGraph):
+                inputs += owner.op.inner_outputs
+
+            return inputs
+
+    for v in walk([var], expand, False):
+        if v.owner and isinstance(v.owner.op, (RandomVariable, MeasurableVariable)):
```
larryshamalama, line 975: Are `RandomVariable`s not measurable?

pymc/testing.py
```diff
@@ -958,5 +959,18 @@ def seeded_numpy_distribution_builder(dist_name: str) -> Callable:
 
 
 def assert_no_rvs(var):
```
larryshamalama, line 961: IIUC, would the name `assert_no_measurable_rvs` be more fitting?

larryshamalama (commented):
Ouf, `Scan`s a bit beyond my area of comfort. I just have minor comments for this one

pymc/testing.py
ricardoV94: They are, but this makes the code more readable I think. Most people don't know or need to know about MeasurableVariables so I rather make the function name reflect RVs and explicitly include them in the isinstance check

pymc/testing.py
ricardoV94: See reasoning above

twiecki (approved):

twiecki:
Need to add tags so this shows up in release notes.

ricardoV94:
> Need to add tags so this shows up in release notes.

Already had. Without tags it still shows up in the release notes, but in the catch-all "maintenance" bucket. The only case it doesn't show up is if it has the label "no releasenotes"

pymc/testing.py
larryshamalama: Okay, I understand

PR #6583, opened 2023-03-09
Implement `check_icdf` helper to test icdf implementations

ricardoV94:
For use in #6528 

Also cleaned up the `check_logp` like helpers

CC @michaelraczycki

### DIFF
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -37,6 +37,7 @@ from pymc import Distribution, logcdf, logp
 from pymc.distributions.shape_utils import change_dist_size
 from pymc.initial_point import make_initial_point_fn
 from pymc.logprob import joint_logp
+from pymc.logprob.abstract import icdf
 from pymc.logprob.utils import ParameterValueError
 from pymc.pytensorf import (
     compile_pymc,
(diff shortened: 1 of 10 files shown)


pymc/testing.py
```diff
@@ -541,6 +542,110 @@ def check_logcdf(
...
+    ----------
+    pymc_dist: PyMC distribution
+    paramdomains : Dictionary of Parameter : Domain pairs
+        Supported domains of distribution parameters
+    scipy_icdf : Scipy icdf method
+        Scipy icdf (ppp) method of equivalent pymc_dist distribution
+    decimal : Int
+        Level of precision with which pymc_dist and scipy_icdf are compared.
+        Defaults to 6 for float64 and 3 for float32
+    n_samples : Int
```
michaelosthege:
```suggestion
    decimal: Optional[int]=None,
    n_samples: int=100,
):
    """
    Generic test for PyMC icdf methods

    The following tests are performed by default:
        1. Test PyMC icdf and equivalent scipy icdf (ppf) methods give similar
        results for parameters inside the supported edges.
        Edges are excluded by default, but can be artificially included by
        creating a domain with repeated values (e.g., `Domain([0, 0, .5, 1, 1]`)
        2. Test PyMC icdf method raises for invalid parameter values
        outside the supported edges.
        3. Test PyMC icdf method returns np.nan for values below 0 or above 1,
         when using valid parameters.

    Parameters
    ----------
    pymc_dist: PyMC distribution
    paramdomains : Dictionary of Parameter : Domain pairs
        Supported domains of distribution parameters
    scipy_icdf : Scipy icdf method
        Scipy icdf (ppp) method of equivalent pymc_dist distribution
    decimal : int, optional
        Level of precision with which pymc_dist and scipy_icdf are compared.
        Defaults to 6 for float64 and 3 for float32
    n_samples : int
```

michaelosthege (approved):

ricardoV94:
Didn't notice some tests were failing

pymc/testing.py
ricardoV94: Done

michaelosthege (commented):
LGTM

michaelosthege (approved):

Issue #6496, opened 2023-02-02
BUG: Issues creating InferenceData if there is a coord of type Sequence[Sequence]


covertg:
### Describe the issue:

Hi, writing with a niche question. I tried to specify a model where I've flattened multiple coords into a "flattened" single coord, where each value is a particular combination of the original coords. E.g. instead of `{"coord1": ["a", "b"], "coord2": [1, 2, 3]}`, I set `{"c": [("a", 1), ("a", 2), ("a", 3), ("b", 1), ("b", 2), ("b", 3)]`.

Is this supported input? Unfortunately, sampling fails in this case.

### Reproduceable code example:

```python
import pymc as pm
with pm.Model(coords={"c": [("a", 1), ("a", 3), ("b", 2)]}) as m:
    x = pm.Normal("x", mu=0, sigma=1, dims="c")
    # m.coords is ({'c': (('a', 1), ('a', 3), ('b', 2))}
    # x.type is TensorType(float64, (3,)))
    idata = pm.sample_prior_predictive()
    # also fails with,
    # idata = pm.sample()
```

### Error message:
<details>

```shell
ValueError                                Traceback (most recent call last)
Cell In[23], line 2
      1 with m:
----> 2     idata = pm.sample_prior_predictive()

File ...python/site-packages/pymc/sampling/forward.py:440, in sample_prior_predictive(samples, model, var_names, random_seed, return_inferencedata, idata_kwargs, compile_kwargs)
    438 if idata_kwargs:
    439     ikwargs.update(idata_kwargs)
--> 440 return pm.to_inference_data(prior=prior, **ikwargs)

File ...python/site-packages/pymc/backends/arviz.py:496, in to_inference_data(trace, prior, posterior_predictive, log_likelihood, coords, dims, sample_dims, model, save_warmup, include_transformed)
    482 if isinstance(trace, InferenceData):
    483     return trace
    485 return InferenceDataConverter(
    486     trace=trace,
    487     prior=prior,
    488     posterior_predictive=posterior_predictive,
    489     log_likelihood=log_likelihood,
    490     coords=coords,
    491     dims=dims,
    492     sample_dims=sample_dims,
    493     model=model,
    494     save_warmup=save_warmup,
    495     include_transformed=include_transformed,
--> 496 ).to_inference_data()

File ...python/site-packages/pymc/backends/arviz.py:408, in InferenceDataConverter.to_inference_data(self)
    396 def to_inference_data(self):
[...]

Issue #5043, opened 2021-10-05
Passing tuples as coordinate values gives `ValueError` after sampling


michaelosthege:
## Description of your problem

This minimum example uses a tuple for coordinate values. Nothing fancy, right?
```python
with pymc3.Model(coords={
    "city": ("Bonn", "Berlin")
}):
    pymc3.Normal("x", dims="city")
    pymc3.sample(return_inferencedata=True)
```

Kaboom 🧨  _after_ sampling.
```
ValueError: dimensions ('city',) must have the same length as the number of data dimensions, ndim=0
```

<details><summary>Traceback</summary>

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-132-c3a70b936a71> in <module>
      3 }):
      4     pymc3.Normal("x", dims="city")
----> 5     pymc3.sample(return_inferencedata=True)

~\AppData\Local\Continuum\miniconda3\envs\CARenv\lib\site-packages\pymc3\sampling.py in sample(draws, step, init, n_init, start, trace, chain_idx, chains, cores, tune, progressbar, model, random_seed, discard_tuned_samples, compute_convergence_checks, callback, jitter_max_retries, return_inferencedata, idata_kwargs, mp_ctx, pickle_backend, **kwargs)
    652         if idata_kwargs:
    653             ikwargs.update(idata_kwargs)
--> 654         idata = pm.to_inference_data(trace, **ikwargs)
    655 
    656     if compute_convergence_checks:

~\AppData\Local\Continuum\miniconda3\envs\CARenv\lib\site-packages\pymc3\backends\arviz.py in to_inference_data(trace, prior, posterior_predictive, log_likelihood, coords, dims, model, save_warmup, density_dist_obs)
    636         return trace
    637 
--> 638     return InferenceDataConverter(
    639         trace=trace,
    640         prior=prior,

~\AppData\Local\Continuum\miniconda3\envs\CARenv\lib\site-packages\pymc3\backends\arviz.py in to_inference_data(self)
    567         """
    568         id_dict = {
--> 569             "posterior": self.posterior_to_xarray(),
    570             "sample_stats": self.sample_stats_to_xarray(),
    571             "log_likelihood": self.log_likelihood_to_xarray(),

~\AppData\Local\Continuum\miniconda3\envs\CARenv\lib\site-packages\arviz\data\base.py in wrapped(cls, *args, **kwargs)
[...]

PR #6589, opened 2023-03-11
Represent `Model._coords` values as 1-dim numpy arrays

covertg:
**What is this PR about?**
Seeks to address #6496: buggy behavior could occur in InferenceData creation (and perhaps elsewhere, see diff) when a user provides a coordinate value sequence that numpy interprets as `ndim>1`.

We prevent this by changing the internal representation of `Model._coords` to be `Dict[str, Optional[np.ndarray]]` rather than `Dict[str, Optional[Tuple]]`, and we ensure that the internal ndarray is 1dim.

Currently this PR does not change the publicly-named `Model.coords` property. For backwards-compatibility, that property still returns a dict of tuples. Instead we expose a new model property `Model.coords_typed` that gives the model coord values in the ndarray format.

As it turns out, nearly all of the times that pymc library would access `Model.coords`, it would seek to convert the (tupled) coordinate values to numpy array. This PR instead updates the whole library *(not currently tests)* to use the new `coords_typed` property. Ultimately this is how #6496 is resolved.

This PR is currently a draft to request discussion on if this is the best approach. One alternative would be to update the `Model.coords` property itself—this would prevent cluttering Model with another property but would be less backwards-compatible. Another question is if `np.ndarray` is in fact the best way to represent coordinate values. E.g. one existing test case in the library gives an example of passing `pd.MultiIndex` to coords when converting a trace to InferenceData, and allowing pandas indexes has been discussed in #6081.

Once an approach is given the :+1: I'll add some tests (especially regression test).

**Checklist**

## Major / Breaking Changes
- n/a

## New features
- The new `Model.coords_typed` property ensures that coordinate values are 1dim (if not none)

## Bugfixes
- #6496
- Also as discussed, both `coords` properties return coordinates/coordinate values as *copies*

## Documentation
- n/a

## Maintenance
- n/a

### DIFF
--- a/tests/backends/test_arviz.py
+++ b/tests/backends/test_arviz.py
@@ -630,15 +630,30 @@ class TestDataPyMC:
             # We're not automatically converting things other than tuple,
             # so advanced use cases remain supported at the InferenceData level.
             # They just can't be used in the model construction already.
-            converter = InferenceDataConverter(
-                trace=mtrace,
-                coords={
-                    "city": pd.MultiIndex.from_tuples(
-                        [("Bonn", 53111), ("Berlin", 10178)], names=["name", "zipcode"]
-                    )
-                },
-            )
-            assert isinstance(converter.coords["city"], pd.MultiIndex)
+            # TODO: we now ensure that everything passed to arviz is converted to
+            # ndarray, which makes the following test fail.
+            # converter = InferenceDataConverter(
+            #     trace=mtrace,
+            #     coords={
+            #         "city": pd.MultiIndex.from_tuples(
+            #             [("Bonn", 53111), ("Berlin", 10178)], names=["name", "zipcode"]
+            #         )
+            #     },
+            # )
+            # assert isinstance(converter.coords["city"], pd.MultiIndex)
+
+    def test_nested_coords_issue_6496(self):
+        """Regression test to ensure we don't bug out if coordinate values
+        appear "nested" to numpy.
+        """
+        model = pm.Model(coords={"cname": [("a", 1), ("a", 2), ("b", 1)]})
+        idata = to_inference_data(
+            prior={"x": np.zeros((100, 3))}, dims={"x": ["cname"]}, model=model
+        )
+        idata_coord = idata.prior.coords["cname"]
+        assert len(idata_coord) == 3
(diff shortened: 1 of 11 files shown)


covertg:
CI tests found a bug in the original PR; now corrected! I also added an initial regression test for #6496.

tests/backends/test_arviz.py
```diff
@@ -630,15 +630,30 @@ def test_issue_5043_autoconvert_coord_values(self):
...
-            converter = InferenceDataConverter(
-                trace=mtrace,
-                coords={
-                    "city": pd.MultiIndex.from_tuples(
-                        [("Bonn", 53111), ("Berlin", 10178)], names=["name", "zipcode"]
-                    )
-                },
-            )
-            assert isinstance(converter.coords["city"], pd.MultiIndex)
+            # TODO: we now ensure that everything passed to arviz is converted to
```
covertg, line 633: Flagging this for discussion. This PR currently updates InferenceDataConverter to always convert coordinate values to ndarray. (Previously they were only converted if the sequence was a tuple.) Is there a reason we would want to maintain the original behavior...?

This test was originally made for #5043

covertg:
The latest run revealed one other case in the test suite that is incompatible with this PR as it is. One test ([link](https://github.com/pymc-devs/pymc/blob/5dcd1010aa4dde2addeaf953229c0012daf895da/tests/distributions/test_bound.py#L200C9-L215)) for `pm.Bound` uses a model with,
```python
pm.Model(coords={"sample": np.ones((2, 5))})
```
Then when creating a `Bound` variable with `dims="sample"`, it expects that variable to have shape (2,5). This is another case beyond my *confident* assessment of what should be done, but I'm *pretty sure* that we don't want users to do this, right? (i.e. each `dim`/dimension in xarray is definitionally 1-D)

Also looks like the pre-commit run failed due to a different commit touching mcmc.py?

...
Anyways, before changing anything up it would be helpful to have some input on how to best internally represent coordinate values. (cc @michaelosthege @ricardoV94 @aseyboldt ?) I'll be rather unavailable for the next ~7 days but will be glad to help on this however I can after that time!

ricardoV94:
I didn't have time to read on this issue about the internal representation. That Bound test is incorrect, not a design choice, feel free to fix.

PR #6596, opened 2023-03-13
Use more readable ignore logprob

ricardoV94:
This PR replaces uses of `assign_custom_measurable_outputs` by the more readable `ignore_logprob`. 

It also refactors a utility to ignore the logprob of multiple (potentially nested) variables consistently.

This PR is a spinoff of #6529

### DIFF
--- a/pymc/logprob/utils.py
+++ b/pymc/logprob/utils.py
@@ -37,7 +37,17 @@
 import warnings
 
 from copy import copy
-from typing import Callable, Dict, Generator, Iterable, List, Optional, Set, Tuple
+from typing import (
+    Callable,
+    Dict,
+    Generator,
+    Iterable,
+    List,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+)
 
 import numpy as np
 
@@ -262,7 +272,7 @@ def diracdelta_logprob(op, values, *inputs, **kwargs):
 def ignore_logprob(rv: TensorVariable) -> TensorVariable:
     """Return a duplicated variable that is ignored when creating logprob graphs
 
-    This is used in SymbolicDistributions that use other RVs as inputs but account
+    This is used in by MeasurableRVs that use other RVs as inputs but account
     for their logp terms explicitly.
 
     If the variable is already ignored, it is returned directly.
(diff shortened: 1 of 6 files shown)


Armavica (approved):
Is there a reason to not inline `assign_custom_measurable_outputs` inside `ignore_logprob` since I believe it is now only used there?

ricardoV94:
> Is there a reason to not inline `assign_custom_measurable_outputs` inside `ignore_logprob` since I believe it is now only used there?

That function is a bit more powerful. It can be used to toggle some outputs on and off as measurable. 

Although it's not being used anywhere else I would leave it separately until it's proven not ever neeeded

pymc/logprob/utils.py
```diff
@@ -272,6 +282,7 @@ def ignore_logprob(rv: TensorVariable) -> TensorVariable:
     op_type = type(node.op)
     if op_type.__name__.startswith(prefix):
         return rv
```
larryshamalama, line 284: I know that this isn't the point of this PR, but I believe that this check is already done in (unless there's something that is not being caught in `assign_custom_measurable_outputs`)

https://github.com/pymc-devs/pymc/blob/c709abf87967216346d2bb0757334f428e317718/pymc/logprob/abstract.py#L213-L219

pymc/logprob/utils.py
```diff
@@ -294,3 +305,32 @@ def reconsider_logprob(rv: TensorVariable) -> TensorVariable:
...
+
+
+def ignore_logprob_multiple_vars(
+    vars: Sequence[TensorVariable], rvs_to_values: Dict[TensorVariable, TensorVariable]
+) -> List[TensorVariable]:
+    """Return duplicated variables that are ignored when creating logprob graphs.
+
+    This function keeps any interdependencies between variables intact, after
+    making each "unmeasurable", whereas a sequential call to `ignore_logprob`
+    would not do this correctly.
```
larryshamalama, line 317: How would sequential calls to `ignore_logprob` not do this correctly? I see that you make use of [`_replace_rvs_in_graph`](https://github.com/pymc-devs/pymc/blob/main/pymc/pytensorf.py#L208) with `replacement_fn`, perhaps that's the magic but it's a bit unclear to me...

pymc/logprob/utils.py
```diff
@@ -294,3 +305,32 @@ def reconsider_logprob(rv: TensorVariable) -> TensorVariable:
     new_node.op = copy(new_node.op)
     new_node.op.__class__ = original_op_type
     return new_node.outputs[node.outputs.index(rv)]
+
+
+def ignore_logprob_multiple_vars(
```
larryshamalama, line 310: Would it be good to add a test for `ignore_logprob_multiple_vars`?

larryshamalama (commented):
This PR has already been merged, but I still had some questions here and there 😅

pymc/logprob/utils.py
ricardoV94: Yes, the check was done in PyMC before Aeppl did it back in the day

pymc/logprob/utils.py
ricardoV94: If you have:

```python
x = pm.Normal.dist()
y = pm.Normal.dist(x)

new_x = ignore_logprob(x)
new_y = ignore_logprob(y)
assert new_y.owner.inputs[3] is new_x  # Would fail
```

pymc/logprob/utils.py
ricardoV94: The functionality is a core requirement of `stack_logprob`, and it's covered in this test, when it was first introduced:

https://github.com/pymc-devs/pymc/blob/c709abf87967216346d2bb0757334f428e317718/tests/logprob/test_tensor.py#L134-L171

https://github.com/pymc-devs/pymc/pull/6342

This PR just refactored it out, because it proved useful elesewhere #6529 

***

Having said that, we can definitely add a targeted test.

PR #6597, opened 2023-03-13
Allow logcdf and icdf inference

ricardoV94:
This PR allows inference of logcdf and icdf variables (similar to how logp was already possible)

```python
import pymc as pm

dist = pm.Normal.dist().exp()

print(pm.logcdf(dist, 1.0).eval())  # -0.6931471805599453
print(pm.icdf(dist, 0.7).eval())  # 1.6894457434840042
```

## Important
This PR first 3 commits now belong to #6599, only the last one should be reviewed here

PR #6599, opened 2023-03-14
Rename `logprob/joint_logp` to `logprob/basic` and move `logcdf` and `icdf` functions there

ricardoV94:
This PR is a spinoff of #6597, to facilitate review.

Important changes:
* `logprob/joint_logprob` module was renamed to `logprob/basic.py`
   It holds all the user-facing probability constructors
* The dispatch call helpers in abstract are now private-named functions and their behavior is clarified
* `logp` no longer calls the heavier factorized_joint_logprob. It immediately dispatches on the IR


pymc/logprob/basic.py
```diff
@@ -56,23 +56,16 @@
 from pymc.logprob.utils import rvs_to_value_vars
 
 
-def logp(rv: TensorVariable, value) -> TensorVariable:
+def logp(rv: TensorVariable, value: TensorVariable, **kwargs) -> TensorVariable:
     """Return the log-probability graph of a Random Variable"""
 
     value = pt.as_tensor_variable(value, dtype=rv.dtype)
```
Armavica, line 71: If `value` is already a `TensorVariable`, is there a reason to keep this line?

pymc/logprob/basic.py
```diff
@@ -56,23 +56,16 @@
...
     try:
-        return logp_logprob(rv, value)
+        return logp_logprob(rv, value, **kwargs)
     except NotImplementedError:
```
Armavica, line 74: Let me see if I understand this correctly:

We enter this `NonImplementedError` in cases where the `rv.owner.op` has a type that was not registered anywhere with `@_logprob.register(...)`, and so that falls back to the default (non)-implementation in `abstract.py`.

In these cases, why would the `FunctionGraph` approach work?

pymc/logprob/abstract.py
Armavica, line 1: I love this!

pymc/logprob/basic.py
ricardoV94: Actually this allows it to receive tensor-like inputs. Should update the type hints...

pymc/logprob/basic.py
ricardoV94: It falls back to the intermediate representation (IR) which converts stuff like Exp(Beta) into MeasurableExp(Beta) (or whatever is deemed valid), which itself has a logp method dispatched. 

That's basically how inference for non basic RVs is carried out.

pymc/logprob/basic.py
ricardoV94: Updated!

ricardoV94:
Ready for re-review

Armavica (approved):
Ok, I think I understand this well enough. I like the clearer hierarchy of functions and the reduction of user-facing logp-related functions!

PR #6604, opened 2023-03-15
Add definition of B in documentation for Beta distribution

JoKeyser:
**What is this PR about?**

The documentation about the Beta distribution should include a definition of the expression $B$ that is used in the definition.

**Checklist**

## Documentation

- Added the definition of $B$ to the documentation of the Beta distribution.

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1042,6 +1042,9 @@ class Beta(UnitContinuous):
        f(x \mid \alpha, \beta) =
            \frac{x^{\alpha - 1} (1 - x)^{\beta - 1}}{B(\alpha, \beta)}
 
+    where :math:`B(\alpha, \beta) = \frac{\Gamma(\alpha)\Gamma(\beta)}{\Gamma(\alpha + \beta)}`
+    and :math:`\Gamma` is the Gamma function.
+
     .. plot::
         :context: close-figs
 
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1042,8 +1042,9 @@ class Beta(UnitContinuous):
        f(x \mid \alpha, \beta) =
            \frac{x^{\alpha - 1} (1 - x)^{\beta - 1}}{B(\alpha, \beta)}
 
-    where :math:`B(\alpha, \beta) = \frac{\Gamma(\alpha)\Gamma(\beta)}{\Gamma(\alpha + \beta)}`
-    and :math:`\Gamma` is the Gamma function.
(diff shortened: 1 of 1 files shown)


pymc/distributions/continuous.py
```diff
@@ -1042,6 +1042,9 @@ class Beta(UnitContinuous):
        f(x \mid \alpha, \beta) =
            \frac{x^{\alpha - 1} (1 - x)^{\beta - 1}}{B(\alpha, \beta)}
 
+    where :math:`B(\alpha, \beta) = \frac{\Gamma(\alpha)\Gamma(\beta)}{\Gamma(\alpha + \beta)}`
```
ricardoV94: Is it enough to say `where B is the Beta function`?

JoKeyser: Seems OK to me as well.

I'm also wondering if it would be nice to link somewhere to clarify what the Beta function is, or maybe that's too elementary to bother?

ricardoV94: We can add a link to the Wikipedia page for the Beta distribution? From there users should be able to find the formula and specifics

JoKeyser: Linking to Wikipedia makes sense to me, and there are other instances of linking to it [e.g. here](https://www.pymc.io/projects/docs/en/v5.1.2/api/distributions/generated/pymc.WishartBartlett.html), but only very few (according to `grep -R wikipedia pymc`).

Sorry if I'm being pedantic about this, but if we decide to link Wikipedia, it may be more sensible to give a direct link to the distribution instead of just the function?

Here's what I mean, let me know which option you prefer:

1. Option with generic link to the distribution:  
   "where B is the Beta function.  
  
   For more information, see <https://en.wikipedia.org/wiki/Beta_distribution>."
2. Option with specific link to the function:
   "where B is the Beta function. For more information, see <https://en.wikipedia.org/wiki/Beta_function>."

I slightly prefer option 1, it seems more direct, shorter, and generically applicable to similar situations.

ricardoV94: Yes I would link to the Beta distribution, not the function. Thanks for being thoughtful over this!

JoKeyser: And thank you for the useful guidance!
(Just now I realized that you suggested to link to the distribution, not the function in the first place :smile: .)

I've added another commit to change to option 1.
(I guess you could squash them together before applying to `main`?)

twiecki (approved):

ricardoV94:
Thanks @JoKeyser

Issue #6608, opened 2023-03-17
DOC: updates to GP


daniel-saunders-phil:
### Issue with current documentation:

The GP docs haven't been updated to the current style as described in [here](https://pymc-data-umbrella.xyz/en/latest/sprint/tutorials/docstring_tutorial.html#edit-the-docstring). In addition, there are also some small typos and oversights. For example,
- The Student T process [claims](https://www.pymc.io/projects/docs/en/latest/api/gp/generated/classmethods/pymc.gp.TP.conditional.html#pymc.gp.TP.conditional) keyword args will be passed to MvNormal instead of MvStudentT
- gp.Marginal doc string directs readers to methods for `.prior`, when it really should be directing them to `.marginal_likelihood`,  `.predict`.

and a few other small issues like this. I want to take on updating the docs in this area.

### Idea or request for content:

I did want input on a few things before I get too deep into the changes.

- All the GP base classes claim you can pass in a 2d array to the covariance function. While that is true, if you set up the GP with a 2d array, it seems like none of the other methods for the GP work. They all throw up an error saying the 2d array is not callable. It might be helpful to include some indication of when this is useful or remove it from the docs.
- jitter. The default value is 1e-6. The doc string don't explicitly indicate what the default is or that the argument is optional. Should we list the default value of 1e-6 on each docstring or should we say `JITTER_DEFAULT`  or is there some other way of indicating the default?
- In gp.Marginal methods, it clarifies that when X is one-dimensional, it really needs to be a (n, 1) column vector. I imagine we should tell users the same thing about gp.Latent methods and the other classes. I just want to check if there is any reason to keep them separate?

I should have a PR ready in the next couple days so please do jump in.

Issue #5720, opened 2022-04-20
Update Gaussian_Processes.rst for v5


canyon289:
## Description of your problem
As title reads update the docs. Bill and I are on it and will get it done sometime by July. Creating this issue to track

https://github.com/pymc-devs/pymc/issues/5538
https://github.com/pymc-devs/pymc/blob/main/docs/source/guides/Gaussian_Processes.rst

OriolAbril:
Whenever we get to update this page, we have to make sure to add a target `additive_gp` at the section on additive GPs. 

---

With myst, targets are added with:

```markdown
(additive_gp)=
## Section heading
```

with rst it is instead:

```rst
.. _additive_gp:

Section heading
---------------
```

(the underscore is important and part of the syntax, not of the target)

daniel-saunders-phil:
Hi all, I would be interested in picking this up. It would be helpful to get some sense of what needed changes have already been identified. I imagine the main one is just executing the code and making small adjustments to ensure compatibility with version 5. But any other concerns come to mind?

daniel-saunders-phil:
It looks like the change from pymc3 to 4/5 didn't introduce any problems that require the code to adjust. The core functionality of the gp module has stayed pretty consistent across version. The only change I noticed was that `noise` is now called `sigma`. I also spotted and fixed a broken link. Otherwise, I'm not sure what else needs to be done. The content is still very helpful and fills an important gap between the GP examples and the GP documentation.

OriolAbril:
Are there code examples that can be executed? If so I'd update the content to jupyter or myst notebook ([example of myst notebook](https://github.com/arviz-devs/arviz/blob/main/doc/source/user_guide/plots_arguments_guide.md) in case it helps, it is a raw text notebook format) and move it to the [core notebooks](https://github.com/pymc-devs/pymc/tree/main/docs/source/learn/core_notebooks) folder. 

If the code snippets there should not be executed, keep it as rst but move it to the core notebooks folder too.

After that, add it to the toctree in https://github.com/pymc-devs/pymc/blob/main/docs/source/learn/core_notebooks/index.md and it will appear again on the docs

daniel-saunders-phil:
Looks like the code was originally designed to be more schematic and wasn't intended to be executed. I submitted a PR to move the .rst file and also added it to the toctree.

PR #6609, opened 2023-03-17
overhaul gp docstrings

daniel-saunders-phil:
**What is this PR about?** 

#6608 noted some motivation for updating the docs throughout the GP module. Much of the work is just style changes. However a couple of new questions came up as I worked through. I'm marking this as a draft pull request until I can get some input:

1. `is_observed=` argument is used on `MarginalKron.marginal_likelihood` and also `Marginal.marginal_likelihood`. In both cases it is deprecated. But on MarginalKron, there is still a docstring for the argument and on Marginal there isn't. Surely we should standardize but which direction is preferred? I imagine we should eliminate the doc on MarginalKron.
2. The marginal GPs both have a predict method that can take a model argument. But no description is provided. I imagine the right description is something like the way we talk about it around the samplers - `model : Model, optional if in with context`

### DIFF
--- a/pymc/gp/gp.py
+++ b/pymc/gp/gp.py
@@ -111,9 +111,9 @@ class Latent(Base):
 
     Parameters
     ----------
-    mean_func : instance of Mean, default ~pymc.gp.mean.Zero
+    mean_func : Mean, default ~pymc.gp.mean.Zero
         The mean function.
-    cov_func : 2D array-like, or instance of Covariance, default ~pymc.gp.cov.Constant
+    cov_func : 2D array-like, or Covariance, default ~pymc.gp.cov.Constant
         The covariance function.
 
     Examples
@@ -179,7 +179,7 @@ class Latent(Base):
         reparameterize : bool, default True
             Reparameterize the distribution by rotating the random
             variable by the Cholesky factor of the covariance matrix.
-        jitter : scalar, default 1e-6
+        jitter : float, default 1e-6
             A small correction added to the diagonal of positive semi-definite
             covariance matrices to ensure numerical stability.
         **kwargs
@@ -242,9 +242,9 @@ class Latent(Base):
             vector with shape `(n, 1)`.
         given : dict, optional
             Can take as key value pairs: `X`, `y`,
-            and `gp`. See the section in the documentation on additive GP
-            models in PyMC for more information.
-        jitter : scalar, default 1e-6
+            and `gp`. See the :ref:`section <additive_gp>` in the documentation
+            on additive GP models in pymc for more information.
+        jitter : float, default 1e-6
             A small correction added to the diagonal of positive semi-definite
             covariance matrices to ensure numerical stability.
         **kwargs
@@ -264,7 +264,6 @@ class TP(Latent):
     The usage is nearly identical to that of `gp.Latent`.  The differences
     are that it must be initialized with a degrees of freedom parameter, and
     TP is not additive. Given a mean and covariance function, and a degrees of
-    TP is not additive. Given a mean and covariance function, and a degrees of
     freedom parameter, the function :math:`f(x)` is modeled as,
 
     .. math::
@@ -274,11 +273,11 @@ class TP(Latent):
 
     Parameters
     ----------
-    mean_func : instance of Mean, default ~pymc.gp.mean.Zero
+    mean_func : Mean, default ~pymc.gp.mean.Zero
         The mean function.
-    scale_func : 2D array-like, or instance of Covariance, default ~pymc.gp.cov.Constant
+    scale_func : 2D array-like, or Covariance, default ~pymc.gp.cov.Constant
         The covariance function.
-    cov_func : 2D array-like, or instance of Covariance, default None
+    cov_func : 2D array-like, or Covariance, default None
         Deprecated, previous version of "scale_func"
     nu : float
         The degrees of freedom
@@ -334,7 +333,7 @@ class TP(Latent):
         reparameterize : bool, default True
             Reparameterize the distribution by rotating the random
             variable by the Cholesky factor of the covariance matrix.
-        jitter : scalar, default 1e-6
+        jitter : float, default 1e-6
             A small correction added to the diagonal of positive semi-definite
             covariance matrices to ensure numerical stability.
         **kwargs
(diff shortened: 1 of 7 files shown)


daniel-saunders-phil:
@OriolAbril would you be willing to help me with this?

fonnesbeck:
Thanks for taking this on!

reshamas:
@daniel-saunders-phil 
Good to see you here again!
an option: 
- upcoming sprint:  https://www.meetup.com/data-umbrella/events/291768545/

OriolAbril:
If an argument is deprecated it should still be in the docs, with a note in the description about being deprecated. It is once it is eventually removed that both argument and itd docstring are deleted

For model you should use:

```
model : Model, optional
    Description, can be one sentence. It is here where conditions and further explanation om what optional means go. In the first line it is only argument name and type info.
```
Potential description "Model with the GP component for which predictions will be generated. It is optional when inside a with context, otherwise it is required"

pymc/gp/gp.py
```diff
@@ -111,10 +111,10 @@ class Latent(Base):
 
     Parameters
     ----------
-    cov_func: None, 2D array, or instance of Covariance
-        The covariance function.  Defaults to zero.
-    mean_func: None, instance of Mean
-        The mean function.  Defaults to zero.
+    cov_func : 2D array, or instance of Covariance, default Zero
```
OriolAbril:
```suggestion
    cov_func : 2D array_like or Covariance, default ~pymc.gp.cov.Constant
```

Similar comment for mean_func, and I would also switch the order so it matches the order they appear in `__init__`

pymc/gp/gp.py
```diff
@@ -233,15 +234,16 @@ def conditional(self, name, Xnew, given=None, jitter=JITTER_DEFAULT, **kwargs):
...
-            Function input values.
-        given: dict
-            Can optionally take as key value pairs: `X`, `y`,
-            and `gp`.  See the section in the documentation on additive GP
+        Xnew : array-like
+            Function input values. If one-dimensional, must be a column
+            vector with shape `(n, 1)`.
+        given : dict, optional
+            Can take as key value pairs: `X`, `y`,
+            and `gp`. See the section in the documentation on additive GP
```
OriolAbril: This can't be a mention, it should be a cross-reference. Anyone knows what page it means?

pymc/gp/gp.py
```diff
@@ -320,15 +322,19 @@ def prior(self, name, X, reparameterize=True, jitter=JITTER_DEFAULT, **kwargs):
...
+            vector with shape `(n, 1)`.
+        reparameterize : bool, default True
             Reparameterize the distribution by rotating the random
             variable by the Cholesky factor of the covariance matrix.
+        jitter : scalar, default 1e-6
+            A small correction added to the diagonal of positive semi-definite
+            covariance matrices to ensure numerical stability.
         **kwargs
-            Extra keyword arguments that are passed to distribution constructor.
+            Extra keyword arguments that are passed to 'MvStudentT' distribution constructor.
```
OriolAbril:
```suggestion
            Extra keyword arguments that are passed to :class:`~pymc.MvStudentT' distribution constructor.
```

pymc/gp/gp.py
```diff
@@ -589,20 +596,21 @@ def predict(
 
         Parameters
         ----------
-        Xnew: array-like
+        Xnew : array-like
             Function input values.  If one-dimensional, must be a column
             vector with shape `(n, 1)`.
-        point: pymc.model.Point
+        point : pymc.model.Point, optional
```
OriolAbril:
```suggestion
        point : pymc.Point, optional
```

OriolAbril (commented):
Quick review from my phone, not sure when I'll have time for a full review

pymc/gp/gp.py
daniel-saunders-phil: After digging around for a while, it looks like the only illustration of how additive GPs work is https://www.pymc.io/projects/examples/en/latest/gaussian_processes/GP-MaunaLoa.html

It seems like a gap in the examples/docs and might need to be addressed in future work.

OriolAbril: Maybe it's this page still pending update from v3? https://github.com/pymc-devs/pymc/issues/5720 cc @bwengals @canyon289

daniel-saunders-phil: Can we link to unfinished work in a way that the docs will sync up once #5720 is finished or should we leave this be for now?

bwengals: Yeah it might be, it should point here: https://www.pymc.io/projects/docs/en/v3/Gaussian_Processes.html#additive-gps

daniel-saunders-phil:
Okay, I got this pr into a stable shape. It’s ready for review whenever anyone is up for it.

the only unresolved bit is the reference to the additive GP example but given that it is still underconstruction, I’ll leave it be.

OriolAbril (commented):
Some minor general comments:

* I prefer using float/int (depending on which is correct) instead of scalar
* I would also delete the "instance of Class" and leave "Class" only, I think the info conveyed by both is the same.
* For the reference, I would add a 

  ```
  See the :ref:`section <additive_gp>` in the documentation on additive GP models ...
  ```

  We can then comment on the issue to make sure that target is added when the doc is updated

daniel-saunders-phil:
Thanks for taking a look. I adjusted those things and sent in a new commit.

OriolAbril (approved):
not sure what is going on with the extra modified files, so not merging yet. The docstrings look great already

daniel-saunders-phil:
Oh I see what happened with the extra files: a couple weeks ago I messed up some rebasing command and unwittingly added 4 other people's commits to my own. Do you know the best way to proceed? I could open up a new clean PR with only the finished GP file on it. Or is there a way to delete those commits off my branch?

michaelosthege:
> Oh I see what happened with the extra files: a couple weeks ago I messed up some rebasing command and unwittingly added 4 other people's commits to my own. Do you know the best way to proceed? I could open up a new clean PR with only the finished GP file on it. Or is there a way to delete those commits off my branch?

You could try to rebase on latest main and `drop` the commits you want to remove.

If that doesn't work just branch off `main` with a new branch, apply the GP docstring changes, then force-push to the old (this PR) branch.

daniel-saunders-phil:
Thanks Michael for the suggestions. I tried both routes and ran into trouble. After chatting with Oriol, we decided the best route was to just create a new branch and new PR #6652. We successfully merge it a little bit ago. I appreciate the help everyone!

PR #6620, opened 2023-03-24
Improve collect_default_updates

ricardoV94:
This is part of some work to make CustomDist more powerful, specially for timeseries distributions. 

Immediate changes:

* It works with nested RNGs
* It raises error if RNG used in SymbolicRandomVariable is not given an update
* It raises warning if same RNG is used in multiple nodes

### DIFF
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -42,7 +42,6 @@ from pytensor.graph.basic import (
     Variable,
     clone_get_equiv,
     graph_inputs,
-    vars_between,
     walk,
 )
 from pytensor.graph.fg import FunctionGraph
@@ -51,6 +50,7 @@ from pytensor.scalar.basic import Cast
 from pytensor.tensor.basic import _as_tensor_variable
 from pytensor.tensor.elemwise import Elemwise
 from pytensor.tensor.random.op import RandomVariable
+from pytensor.tensor.random.type import RandomType
 from pytensor.tensor.random.var import (
     RandomGeneratorSharedVariable,
     RandomStateSharedVariable,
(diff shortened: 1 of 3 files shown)


pymc/pytensorf.py
```diff
@@ -1000,42 +1000,85 @@ def reseed_rngs(
...
-        and var not in inputs
+    for shared_rng in (
+        inp
+        for inp in graph_inputs(outputs, blockers=inputs)
+        # TODO: Test this directly
+        if (
+            (not must_be_shared or isinstance(inp, SharedVariable))
+            and isinstance(inp.type, RandomType)
+        )
     ):
```
michaelosthege: Extracting this iterator into a local variable would make it easier to read.

Extracting it into a function could make it easier to test too?

michaelosthege (approved):

pymc/pytensorf.py
ricardoV94: Removed the TODO, by adding a direct test. I think this test makes the behavior of the function pretty clear, including the hard to read iterator?

https://github.com/pymc-devs/pymc/blob/d8c57f4af5e272a0d924c76caae625d6f1d0fd1f/tests/test_pytensorf.py#L542-L553

michaelosthege: I still find the iterator hard to read, but proceed on your discretion

ricardoV94: I am still being stubborn, but added a comment and renamed the iterator variable to be a bit more obvious. Will merge when tests pass again :)

https://github.com/pymc-devs/pymc/blob/87c4fe7f6cce8a90d58fc890be63fa36b876e92f/pymc/pytensorf.py#L1062-L1073

Thanks!

PR #6634, opened 2023-03-30
Implement `Model.debug()` helper

### LINKED ISSUE #4205: Improve model debugging
From Prieseman group:
"I don't whether Viola answered this question in the session, I wasn't there, but this is something I can answer as I did a large part of the development. We hadn't had much difficulties with PyMC3, if anything we were surprised I simple the development was. However I missed good debugging tools. It is quite easy to build a model where for some reason nans occur, or the gradient can't be computed. I haven't really found a good way to deal with it. There exist a monitor_mode in theano, but this involves looking through a txt file where the output of each node is printed. It works, but is somewhat cumbersome. I wrote here the different techniques we used: https://covid19-inference.readthedocs.io/en/latest/doc/debugging.html"

ricardoV94:
Example:
```python
import pymc as pm

with pm.Model() as m:
    x = pm.Normal("x", [1, -1, 1])
    y = pm.Normal("y", mu=x, sigma=x * 2)
m.debug()
```

Outputs:
```
point={'x': array([ 1., -1.,  1.]), 'y': array([ 1., -1.,  1.])}

The variable y has the following parameters:
0: x [id A] <TensorType(float64, (3,))>
1: Elemwise{mul,no_inplace} [id B] <TensorType(float64, (3,))>
 |TensorConstant{(1,) of 2.0} [id C] <TensorType(float64, (1,))>
 |x [id A] <TensorType(float64, (3,))>
The parameters evaluate to:
0: [ 1. -1.  1.]
1: [ 2. -2.  2.]
This does not respect one of the following constraints: sigma > 0
```

Another example:
```python
import pymc as pm

with pm.Model() as m:
    theta = pm.Uniform('theta', lower=0, upper=1)
    y = pm.Uniform('y', lower=0, upper=theta, observed=[0.49, 0.27, 0.53, 0.19])
m.debug()    
```
```
point={'theta_interval__': array(0.)}

The variable y has the following parameters:
0: TensorConstant{0.0} [id A] <TensorType(float64, ())>
1: Elemwise{sigmoid,no_inplace} [id B] <TensorType(float64, ())> 'theta'
 |theta_interval__ [id C] <TensorType(float64, ())>
The parameters evaluate to:
0: 0.0
1: 0.5
Some of the observed values of variable y are associated with a non-finite logp:
 value = 0.53 -> logp = -inf
```

Closes #4205

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -13,6 +13,7 @@
 #   limitations under the License.
 
 import functools
+import sys
 import threading
 import types
 import warnings
@@ -24,6 +25,7 @@ from typing import (
     Callable,
     Dict,
     List,
+    Literal,
     Optional,
     Sequence,
     Tuple,
@@ -39,6 +41,7 @@ import pytensor.sparse as sparse
 import pytensor.tensor as pt
 import scipy.sparse as sps
 
+from pytensor.compile import DeepCopyOp, get_mode
 from pytensor.compile.sharedvalue import SharedVariable
 from pytensor.graph.basic import Constant, Variable, graph_inputs
 from pytensor.graph.fg import FunctionGraph
@@ -46,6 +49,7 @@ from pytensor.scalar import Cast
 from pytensor.tensor.elemwise import Elemwise
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.rewriting import local_subtensor_rv_lift
+from pytensor.tensor.random.type import RandomType
 from pytensor.tensor.sharedvar import ScalarSharedVariable
 from pytensor.tensor.var import TensorConstant, TensorVariable
 
@@ -61,6 +65,7 @@ from pymc.exceptions import (
 )
 from pymc.initial_point import make_initial_point_fn
 from pymc.logprob.basic import joint_logp
+from pymc.logprob.utils import ParameterValueError
 from pymc.pytensorf import (
     PointFunc,
     SeedSequenceSeed,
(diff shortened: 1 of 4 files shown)


pymc/model.py
```diff
@@ -1811,6 +1817,137 @@ def point_logps(self, point=None, round_vals=2):
...
+        This function should help to quickly narrow down invalid parametrizations.
+
+        Parameters
+        ----------
+        point: Point
+            Point at which model function should be evaluated
+        fn: str, default "logp"
+            Function to be used for debugging. Can be one of [logp, dlogp, random].
+        verbose: bool, default False
+            Whether to show a more verbose PyTensor output when function cannot be evaluated
```
```suggestion
        point : Point
            Point at which model function should be evaluated
        fn : str, default "logp"
            Function to be used for debugging. Can be one of [logp, dlogp, random].
        verbose : bool, default False
            Whether to show a more verbose PyTensor output when function cannot be evaluated
```

pymc/model.py
```diff
@@ -1811,6 +1817,137 @@ def point_logps(self, point=None, round_vals=2):
...
+                    found_problem = True
+                    debug_parameters(rv)
+                    print_(
+                        f"One of the values of variable {rv} is associated with a non-finite {fn} value: {rv_fn_eval}\n"
+                    )
+
+        if not found_problem:
+            print_("No problems found")
+        else:
+            print_("You can set `verbose=True` for more details")
```
```suggestion
        elif not verbose:
            print_("You can set `verbose=True` for more details")
```

michaelosthege (approved):
Very cool!!

ricardoV94:
I refined it to only show the value-evaluation pairs that lead to non-finite values.

PR #6636, opened 2023-03-30
Make floatX and intX always return TensorVariables

ricardoV94:
Also fixes error when list of TensorVariables was passed as input, encountered by a Discourse user here:

https://discourse.pymc.io/t/sampling-many-dimensions-with-different-pdf-in-each-dimension/11650/12?u=ricardov94

----
:books: Documentation preview :books:: https://pymc--6636.org.readthedocs.build/en/6636/


ricardoV94:
Superseded by #7114

PR #6641, opened 2023-03-31
Added ICDF for the exponential distribution

gokuld:
**What is this PR about?**
Adding the ICDF function for the continuous exponential distribution.
Issue: https://github.com/pymc-devs/pymc/issues/6612
Creation of tests for the added ICDF function.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ICDF function for the continuous exponential distribution.

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- Extended test_continuous.py with corresponding tests.

### DIFF
--- a/tests/distributions/test_continuous.py
+++ b/tests/distributions/test_continuous.py
@@ -443,15 +443,6 @@ class TestMatchesScipy:
             {"lam": Rplus},
             lambda q, lam: st.expon.ppf(q, loc=0, scale=1 / lam),
         )
-        # Custom logp / logcdf / icdf check for invalid parameters
-        invalid_dist = pm.Exponential.dist(lam=-1)
-        with pytensor.config.change_flags(mode=Mode("py")):
-            with pytest.raises(ParameterValueError):
-                logp(invalid_dist, np.array(0.5)).eval()
-            with pytest.raises(ParameterValueError):
-                logcdf(invalid_dist, np.array(0.5)).eval()
-            with pytest.raises(ParameterValueError):
-                icdf(invalid_dist, np.array(0.5)).eval()
 
     def test_laplace(self):
         check_logp(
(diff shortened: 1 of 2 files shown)


tests/distributions/test_continuous.py
```diff
@@ -438,6 +438,20 @@ def test_exponential(self):
             {"lam": Rplus},
             lambda value, lam: st.expon.logcdf(value, 0, 1 / lam),
         )
+        check_icdf(
+            pm.Exponential,
+            {"lam": Rplus},
+            lambda q, lam: st.expon.ppf(q, loc=0, scale=1 / lam),
+        )
+        # Custom logp / logcdf / icdf check for invalid parameters
```
ricardoV94: This shouldn't be needed, the `check*` functions already do this. 

The reason we needed special logic for uniform functions was that the helper functions can't automatically figure out what is an invalid parameter, because it depends on both lower and higher. 

For these simpler cases the helper functions try `domain.lower-1` if you check the source code

ricardoV94 (changes requested):
Looks good, just a comment below about the test

ricardoV94:
I suggest you rebase from main instead of merging so that the commits don't show up on the PR (see changed files)

ricardoV94:
https://stackoverflow.com/questions/71074242/github-old-commits-shows-up-in-new-pull-request

tests/distributions/test_continuous.py
gokuld: Yes this makes sense, thanks.
I've deleted the redundant checks and pushed a new commit.

gokuld:
Thanks, done!

ricardoV94:
We will squash merge this, but for future reference when you rebase from main it's also a good opportunity to squash together commits that are not important on their own (e.g., fixing a typo introduced in a previous commit or, in this case, reverting the new tests you added in your first commit).

If you do rebase -i like in the link, you can replace pick by squash prefix so that a commit gets squashed with the previous one.

ricardoV94 (approved):

ricardoV94:
thanks @gokuld!

PR #6642, opened 2023-03-31
Add ICDF for the Kumaraswamy distribution

gokuld:
**What is this PR about?**
Adding the ICDF function for the Kumaraswamy distribution.
Issue: https://github.com/pymc-devs/pymc/issues/6612
Creation of tests for the added ICDF function.

Reference: https://en.wikipedia.org/wiki/Kumaraswamy_distribution
SciPy does not have an implementation for the Kumaraswamy distribution so this ICDF was coded explicitly.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ICDF function for the Kumaraswamy distribution.

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
-  Extended test_continuous.py with corresponding tests.

----
:books: Documentation preview :books:: https://pymc--6642.org.readthedocs.build/en/6642/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1323,6 +1323,16 @@ class Kumaraswamy(UnitContinuous):
             msg="a > 0, b > 0",
         )
 
+    def icdf(value, a, b):
+        res = pt.exp(pt.reciprocal(a) * pt.log1mexp(pt.reciprocal(b) * pt.log1p(-value)))
+        res = check_icdf_value(res, value)
+        return check_icdf_parameters(
+            res,
+            a > 0,
+            b > 0,
+            msg="a > 0, b > 0",
+        )
+
 
 class Exponential(PositiveContinuous):
     r"""
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -668,6 +668,82 @@ def check_selfconsistency_discrete_logcdf(
             )
 
 
+def check_selfconsistency_continuous_icdf(
+    distribution: Distribution,
+    paramdomains: Dict[str, Domain],
+    decimal: Optional[int] = None,
+    n_samples: int = 100,
+) -> None:
+    """
+    Check that the icdf and logcdf functions of the distribution are consistent for a sample of probability values.
+    """
+    if decimal is None:
+        decimal = select_by_precision(float64=6, float32=3)
+
+    dist = create_dist_from_paramdomains(distribution, paramdomains)
+    value = dist.type()
+    value.name = "value"
+
+    dist_icdf = icdf(dist, value)
+    dist_icdf_fn = pytensor.function(list(inputvars(dist_icdf)), dist_icdf)
+
+    dist_logcdf = logcdf(dist, value)
+    dist_logcdf_fn = compile_pymc(list(inputvars(dist_logcdf)), dist_logcdf)
+
+    domains = paramdomains.copy()
+    domains["value"] = Domain(np.linspace(0, 1, 10))
+
+    for point in product(domains, n_samples=n_samples):
+        point = dict(point)
+        value = point.pop("value")
+
+        with pytensor.config.change_flags(mode=Mode("py")):
+            npt.assert_almost_equal(
+                value,
+                np.exp(dist_logcdf_fn(**point, value=dist_icdf_fn(**point, value=value))),
+                decimal=decimal,
+                err_msg=f"point: {point}, value: {value}",
+            )
+
+
+def check_selfconsistency_discrete_icdf(
+    distribution: Distribution,
+    domain: Domain,
+    paramdomains: Dict[str, Domain],
+    n_samples: int = 100,
+) -> None:
+    """
+    Check that the icdf and logcdf functions of the distribution are
+    consistent for a sample of values in the domain of the
+    distribution.
+    """
+    dist = create_dist_from_paramdomains(distribution, paramdomains)
+
+    value = pt.TensorType(dtype="float64", shape=[])("value")
+
+    dist_icdf = icdf(dist, value)
+    dist_icdf_fn = pytensor.function(list(inputvars(dist_icdf)), dist_icdf)
+
+    dist_logcdf = logcdf(dist, value)
+    dist_logcdf_fn = compile_pymc(list(inputvars(dist_logcdf)), dist_logcdf)
+
+    domains = paramdomains.copy()
+    domains["value"] = domain
+
+    for point in product(domains, n_samples=n_samples):
+        point = dict(point)
+        value = point.pop("value")
+
+        with pytensor.config.change_flags(mode=Mode("py")):
+            expected_value = value
+            computed_value = dist_icdf_fn(
+                **point, value=np.exp(dist_logcdf_fn(**point, value=value))
+            )
+            assert (
+                expected_value == computed_value
+            ), f"expected_value = {expected_value}, computed_value = {computed_value}, {point}"
+
+
 def assert_support_point_is_expected(model, expected, check_finite_logp=True):
     fn = make_initial_point_fn(
         model=model,
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -710,6 +710,7 @@ def check_selfconsistency_discrete_icdf(
     distribution: Distribution,
     domain: Domain,
     paramdomains: Dict[str, Domain],
+    decimal: Optional[int] = None,
     n_samples: int = 100,
 ) -> None:
     """
@@ -717,6 +718,13 @@ def check_selfconsistency_discrete_icdf(
     consistent for a sample of values in the domain of the
     distribution.
     """
+
+    def ftrunc(values, decimal=0):
+        return np.trunc(values * 10**decimal) / (10**decimal)
+
+    if decimal is None:
+        decimal = select_by_precision(float64=6, float32=3)
+
     dist = create_dist_from_paramdomains(distribution, paramdomains)
 
     value = pt.TensorType(dtype="float64", shape=[])("value")
@@ -737,7 +745,7 @@ def check_selfconsistency_discrete_icdf(
         with pytensor.config.change_flags(mode=Mode("py")):
             expected_value = value
             computed_value = dist_icdf_fn(
-                **point, value=np.exp(dist_logcdf_fn(**point, value=value))
+                **point, value=ftrunc(np.exp(dist_logcdf_fn(**point, value=value)), decimal=decimal)
             )
             npt.assert_almost_equal(
                 expected_value,
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -668,27 +668,32 @@ def check_selfconsistency_discrete_logcdf(
             )
 
 
-def check_selfconsistency_continuous_icdf(
+def check_selfconsistency_icdf(
     distribution: Distribution,
-    paramdomains: Dict[str, Domain],
-    decimal: Optional[int] = None,
+    paramdomains: dict[str, Domain],
+    *,
+    decimal: int | None = None,
     n_samples: int = 100,
 ) -> None:
-    """
-    Check that the icdf and logcdf functions of the distribution are consistent for a sample of probability values.
+    """Check that the icdf and logcdf functions of the distribution are consistent.
+
+    Only works with continuous distributions.
     """
     if decimal is None:
         decimal = select_by_precision(float64=6, float32=3)
 
     dist = create_dist_from_paramdomains(distribution, paramdomains)
-    value = dist.type()
-    value.name = "value"
-
+    if dist.type.dtype.startswith("int"):
+        raise NotImplementedError(
+            "check_selfconsistency_icdf is not robust against discrete distributions."
+        )
+    value = dist.astype("float64").type("value")
     dist_icdf = icdf(dist, value)
-    dist_icdf_fn = pytensor.function(list(inputvars(dist_icdf)), dist_icdf)
+    dist_cdf = pt.exp(logcdf(dist, value))
 
-    dist_logcdf = logcdf(dist, value)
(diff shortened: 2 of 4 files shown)


tests/distributions/test_continuous.py
```diff
@@ -424,6 +424,11 @@ def scipy_log_cdf(value, a, b):
             {"a": Rplus, "b": Rplus},
             scipy_log_cdf,
         )
+        check_icdf(
+            pm.Kumaraswamy,
+            {"a": Rplus, "b": Rplus},
+            lambda q, a, b: (1 - (1 - q) ** (1 / b)) ** (1 / a),
```
ricardoV94: Doesn't scipy have a ppf function? The point of this test is more to compare with an external reference to be sure we implemented it correctly.

ricardoV94: I see you mention it doesn't. In that case an alternative would be to check that for some values `exp(logcdf(dist, icdf(dist, value))) == value` 

Or something along those lines, basically testing that the inverse property holds.

You could add that as a util test function in `testing.py`. 

We do something similar for the consistency between discrete logp and logcdf helper.

It's a bit more work but it will be very helpful for testing other distributions whose ppf is missing in scipy.

gokuld: No, SciPy doesn't have a ppf function for the Kumaraswamy distribution.

Sure, I will look into adding the util test function for testing when scipy doesn't have an implementation. I will get back on this if I need help, or if not, with the next commit.

gokuld: @ricardoV94 I added a util function for testing the inverse of the icdf matches `exp logcdf`; 
However I found it is failing tests for small values of distribution parameter $b$ due to numerical issues.

Example test case where this is failing:

$a=1$, $b=0.01$, $value=0.4$
The icdf expression is:
$(1 - (1 - value) ^ \frac{1}{b}) ^ \frac{1}{a} $

For the above values for $a$ and $b$, the inner term:
$t = (1 - value) ^ \frac{1}{b}) = 0.6^{100} = 6.533186235000685e-23$

For such a small value of $t$, $1 - t$ is now represented as 1, due to numerical issues with adding 1 to such a small number. Thus the result of the ICDF expression, $(1 - t)^\frac{1}{a}$ is also 1.

This fails the self consistency test (between icdf and cdf) for small values of $b$.

Do you have any idea to make this numerically stable? Working with the logarithms of numbers doesn't make sense as there is addition happening in the ICDF expression which also causes the numerical stability issue.

The other alternative is to simply skip the tests for small values of $b$. I would like to avoid that if another way is possible.

**Edit**: I am now thinking on the lines of using `scipy.special.powm1` or `scipy.special.log1p`.

ricardoV94: How are you testing? We don't want to test equality directly, but using something like `np.testing.assert_almost_equal` or something that allows for some wiggle room. 

It may make sense to compare que log of the icdf value if that's more stable.

But yes, we don't need to test values very close to 0 or 1. It would be good to check if the distributions that already have icdf pass as well.

gokuld: I am already using `np.testing.assert_almost_equal`. This still doesn't work because the probability returned by the cdf of the value returned by the icdf, is $1$, instead of a value 0.4, because of the numerical issue. 

> It may make sense to compare que log of the icdf value if that's more stable.

log of the icdf does make sense, but we have to implement a function like logicdf everywhere. log(icdf) will not work as icdf already returns the less precise value, so it has to be implemented directly as logicdf.

> But yes, we don't need to test values very close to 0 or 1.

Okay, in that case I will change the domain of testing to exclude small values of $b$.

>  It would be good to check if the distributions that already have icdf pass as well.

The other continuous distributions (uniform and normal) with icdf are passing the icdf self consistency test, even with small values of the distribution parameters.
For the discrete case, the test passed for the discrete uniform distribution, but failed for the geometric distribution for small values of the distribution parameter ($0.001$).

ricardoV94: We don't need to implement log-icdf, that's just the log of icdf which we already have.

My proposal was to compare `logcdf(dist, icdf(dist, value)) = log(value)`, if that makes sense. Instead of exponentiating the logcdf

gokuld: > My proposal was to compare logcdf(dist, icdf(dist, value)) = log(value)

No that doesn't work, I already tried it before I narrowed down to the root cause.

The numerical error starts at the output of icdf itself for very small parameters for distributions involving power terms with this parameter. This the log outside does not help.

I will proceed now by excluding very small values of the distribution parameters for the self consistency tests for icdf, if that sounds good.

gokuld: Well actually I see that for some of the domains, domains excluding small parameters is already defined.
For example, using `Rplusbig` instead of `Rplus` for the a and b distribution parameters for the Kumaraswamy distribution already passed the test.

I could simply use these domains for the tests. Does this sound good?

gokuld: Oh while this inverse icdf -cdf self test works for continuous distributions, it doesn't work in the same way for discrete distributions.
This is because there is a band of probability for which the icdf has the same value and this this need not match the probability from the cdf.

I will now implement another util test function for discrete distributions which checks that `value = icdf(dist, exp(logcdf(dist, value)))` for a sample of integer values in the domain of the discrete distribution.

pymc/distributions/continuous.py
```diff
@@ -1287,6 +1287,16 @@ def logcdf(value, a, b):
             msg="a > 0, b > 0",
         )
 
+    def icdf(value, a, b):
+        res = (1 - (1 - value) ** pt.reciprocal(b)) ** pt.reciprocal(a)
```
ricardoV94: Looking at this, would it help to compute this icdf on the log scale (to get rid of these exponents) and only exponentiate the final result?


```suggestion
        res = exp(log(1 - (1 - value) * pt.reciprocal(b) * pt.reciprocal(a))
```
Or something along those lines (might have made a mistake)

pymc/testing.py
```diff
@@ -212,6 +212,7 @@ def RandomPdMatrix(n):
 Rplusbig = Domain([0, 0.5, 0.9, 0.99, 1, 1.5, 2, 20, np.inf])
 Rminusbig = Domain([-np.inf, -2, -1.5, -1, -0.99, -0.9, -0.5, -0.01, 0])
 Unit = Domain([0, 0.001, 0.1, 0.5, 0.75, 0.99, 1])
+Unitbig = Domain([0.1, 0.5, 0.75, 0.99, 1])
```
ricardoV94: The edges aren't included, do to test `0.1` you need to add something before
```suggestion
Unitbig = Domain([0, 0.1, 0.5, 0.75, 0.99, 1])
```

tests/distributions/test_continuous.py
```diff
@@ -284,6 +289,10 @@ def test_normal(self):
             {"mu": R, "sigma": Rplus},
             lambda q, mu, sigma: st.norm.ppf(q, mu, sigma),
         )
+        check_selfconsistency_icdf(
```
ricardoV94: We don't need to test this for scipy distributions that have ppf, I meant just for you to confirm it's working well

ricardoV94 (commented):
Shaping up really well!

Some suggestions below

pymc/distributions/continuous.py
gokuld: Done.

pymc/testing.py
gokuld: Unitbig isn't used now, so I am discarding this.

It was added to test the self-consistency of ICDF-logcdf for the uniform distribution, but we are not using this as they have scipy ppf functions available.

tests/distributions/test_continuous.py
gokuld: Done.
I've removed these tests where scipy ppf functions are available.

pymc/distributions/continuous.py
```diff
@@ -1287,6 +1287,16 @@ def logcdf(value, a, b):
             msg="a > 0, b > 0",
         )
 
+    def icdf(value, a, b):
+        res = pt.exp(pt.reciprocal(a) * pt.log1mexp(pt.reciprocal(b) * pt.log1p(-value)))
```
ricardoV94: Did this seem more stable?

pymc/testing.py
```diff
@@ -657,6 +657,44 @@ def check_selfconsistency_discrete_logcdf(
...
+    value.name = "value"
+
+    dist_icdf = icdf(dist, value)
+    dist_icdf_fn = pytensor.function(list(inputvars(dist_icdf)), dist_icdf)
+
+    dist_logcdf = logcdf(dist, value)
+    dist_logcdf_fn = compile_pymc(list(inputvars(dist_logcdf)), dist_logcdf)
+
+    domains = paramdomains.copy()
+    domains["value"] = Domain(np.linspace(0.1, 1, 100))
```
ricardoV94: Probably enough?
```suggestion
    domains["value"] = Domain(np.linspace(0, 1, 10))
```

ricardoV94 (commented):
Sorry, I forgot about this one. 

Is #6669  a blocker for this PR?

pymc/testing.py
gokuld: Sure.

ricardoV94:
@gokuld How do you prefer to proceed? Get this done or wait and also get the discrete cases sorted out?

I would suggest we wait, but up to you :)

gokuld:
We can wait.

gokuld:
@ricardoV94 I've added self consistency tests for discrete ICDF, now that the blocker https://github.com/pymc-devs/pymc/pull/6671 is closed.

Though SciPy PPFs are available for the discrete uniform and discrete geometric distributions and used with `check_icdf`, I have still added the `icdf-logcdf self consistency` check for these two distributions as it caught a bug in the implementation of discrete uniform ICDF that was not caught by `check_icdf`.

I have also fixed the bug and pushed the commit here.

gokuld:
`pytest -v tests/distributions/test_discrete.py -k 'test_geometric ` passes on my machine.  Looking into this.

ricardoV94:
Try setting `n_samples=-1` in check_logp and alike helpers to make sure all parameter combinations are tested locally

gokuld:
> Try setting `n_samples=-1` in check_logp and alike helpers to make sure all parameter combinations are tested locally

It is passing locally even with `n_samples=-1`. 

The failure upstream is happening in `check_selfconsistency_discrete_icdf` for some reason.

ricardoV94:
Does this branch already include the fixes from the other PR?

gokuld:
> Does this branch already include the fixes from the other PR?

Forgot to sync my forked repo. :/ I have synced it now. How do I trigger the tests again? Force push?

ricardoV94:
You might want to rebase from main and then force push so only the new commits show up (if that's the problem). Make sure to checkout a backup branch first in case something goes wrong.

pymc/testing.py
```diff
@@ -657,6 +657,82 @@ def check_selfconsistency_discrete_logcdf(
...
+    for point in product(domains, n_samples=n_samples):
+        point = dict(point)
+        value = point.pop("value")
+
+        with pytensor.config.change_flags(mode=Mode("py")):
+            expected_value = value
+            computed_value = dist_icdf_fn(
+                **point, value=np.exp(dist_logcdf_fn(**point, value=value))
+            )
+            assert (
```
ricardoV94: `np.testing.assert_almost_equal` will give you a nice printing when it fails out of the box

gokuld:
@ricardoV94 This passed the tests in `test_discrete.py` now. Do you think we should proceed with the truncation approach, within `check_selfconsistency_discrete_icdf` ?

ricardoV94:
> At the moment I check that ICDF(CDF(value) == value, where value is an integer for discrete distributions. It is possible to replace this test condition (note the swap) by CDF(ICDF(value) == value, here value being a real number 0 <= value <=1. However in this case, value is a real number and not an integer and has to be compared using np.testing.assert_almost_equal which effectively truncates the value to k places and checks. There seems to simple be no way to totally avoid the truncation during testing.

What is the disadvantage of the current approach `ICDF(CDF)`? If it works, let's keep it?

Does this PR include changes that were already committed elsewhere?

gokuld:
> What is the disadvantage of the current approach `ICDF(CDF)`? If it works, let's keep it?

One issue is that after computing the CDF, if the result has the tiniest numerical error, it can create a huge error in the ICDF (exactly a 1 integer size error). This is because ICDF is integer valued but is very sensitive at the places in the CDF graph where there is a vertical jump. (The CDF function has a staircase  like appearance with multiple vertical jumps.) This means, `ICDF(CDF)` is not enough, but we need to truncate it this way: `ICDF(truncate(CDF))`. 

And while truncating we need to decide the number of decimal places which is another magic number / parameter, makes it less elegant.

> Does this PR include changes that were already committed elsewhere?

No these changes are made only in this PR.

ricardoV94:
Then let's do it the other way around. `cdf(icdf(value))`? You're saying that will be less discontinuous? 

For your other question, its' fine to check with `allclose` and a specific `rtol` and `atol`. That's always the case when comparing floats.

gokuld:
> Then let's do it the other way around. `cdf(icdf(value))`? You're saying that will be less discontinuous?

An issue with this other way round  `cdf(icdf(value))` is that the CDF of discrete distributions takes a set of discrete values. These values are dependent on the distribution itself (and it's parameters). When using a generic discrete set of CDF values for testing, `cdf(icdf(value))` would 'snap' the `value` to the set of discrete possible values for the particular distribution, and `value` may not match `cdf(icdf(value))`, though the test should pass.

ricardoV94:
@gokuld I'm trying to revive this PR. Given the complexity I decided to remove the discrete self_consistency helper and stay only with the continuous one. The helper will raise if a user tries to call it with a discrete dist.

gokuld:
@ricardoV94 Thank you, sounds good. Glad to see this PR closed.

PR #6645, opened 2023-04-02
Allow passing masked arrays with missing values to pm.Data()

kamicollo:
This PR implements the changes required to support passing masked arrays to pm.Data() as discussed in issue https://github.com/pymc-devs/pymc/issues/6626.

(rest of PR description TBD)
...

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6645.org.readthedocs.build/en/6645/

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -454,6 +454,33 @@ def test_get_data():
     assert type(data) == io.BytesIO
 
 
+def test_masked_data_mutable():
+    with pm.Model():
+        data = np.ma.MaskedArray([1.0, 2.0, 3], [0, 0, 1])
+        expected = np.array([1, 2, np.nan])
+        with pytest.warns(UserWarning, match="masked arrays"):
+            result = pm.MutableData("test", data).get_value()
+        np.testing.assert_array_equal(result, expected)
+
+
+def test_masked_data_constant():
+    with pm.Model():
+        data = np.ma.MaskedArray([1.0, 2.0, 3], [0, 0, 1])
+        expected = np.array([1, 2, np.nan])
+        with pytest.warns(UserWarning, match="masked arrays"):
+            result = pm.ConstantData("test", data).data
+        np.testing.assert_array_equal(result, expected)
+
+
+def test_masked_integer_data():
+    with pm.Model():
+        data = np.ma.MaskedArray([1, 2, 3], [0, 0, 1])
+        with pytest.raises(TypeError, match="Masked integer"):
+            pm.ConstantData("test", data)
+        with pytest.raises(TypeError, match="Masked integer"):
+            pm.MutableData("test", data)
+
+
 class _DataSampler:
     """
     Not for users
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,7 +39,7 @@ repos:
     - id: black
     - id: black-jupyter
 - repo: https://github.com/PyCQA/pylint
-  rev: v2.17.0
+  rev: v2.17.1
   hooks:
     - id: pylint
       args: [--rcfile=.pylintrc]
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -36,7 +36,7 @@ from pytensor.tensor.var import TensorConstant, TensorVariable
 
 import pymc as pm
 
-from pymc.pytensorf import convert_observed_data
+from pymc.pytensorf import convert_observed_data, unmask_masked_data
 
 __all__ = [
     "get_data",
@@ -419,9 +419,19 @@ def Data(
         )
     name = model.name_for(name)
 
+    if isinstance(value, np.ma.MaskedArray):
+        warnings.warn(
+            "If possible, masked arrays will be converted to standard numpy arrays with np.nan values for compatibility with PyTensor."
+        )
+
     # `convert_observed_data` takes care of parameter `value` and
     # transforms it to something digestible for PyTensor.
     arr = convert_observed_data(value)
+
+    # because converted_observed_data() is also used outside pyTensor, we need an extra step to ensure that any masked arrays
+    # produced by it are converted back to np.ndarray() with np.nan value.
+    # This is not very efficient and will not be necessary once pyTensor implements MaskedArray support
+    arr = unmask_masked_data(arr)
 
     if mutable is None:
         warnings.warn(
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -75,6 +75,7 @@ from pymc.pytensorf import (
     hessian,
     inputvars,
     replace_rvs_by_values,
+    unmask_masked_data,
 )
 from pymc.util import (
     UNSET,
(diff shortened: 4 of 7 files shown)


tests/test_data.py
```diff
@@ -454,6 +454,33 @@ def test_get_data():
...
+        data = np.ma.MaskedArray([1.0, 2.0, 3], [0, 0, 1])
+        expected = np.array([1, 2, np.nan])
+        with pytest.warns(UserWarning, match="masked arrays"):
+            result = pm.ConstantData("test", data).data
+        np.testing.assert_array_equal(result, expected)
+
+
+def test_masked_integer_data():
+    with pm.Model():
+        data = np.ma.MaskedArray([1, 2, 3], [0, 0, 1])
```
ricardoV94, line 477: Integers should be fine, otherwise we can't input discrete variables?

ricardoV94 (commented):
Thanks for opening the PR.

We need to test if automatic imputation is working as that is the point of passing nan to observed in PyMC.

We also need to decide what to do if a user passes nan to a Mutabledata that is used as observed, only after defining the observed variable, as that won't trigger the automatic imputation routine. Maybe raise in that case?

I am also afraid MutableData+Imputation won't ever be useful during posterior predictive so maybe we shouldn't allow nan in Mutabledata at all? That avoids the problem above.

kamicollo:
Thanks for the comments. 

Agreed, the PR still needs to include the part that makes Mutable/Constant data to be used in the automatic imputation process (I just opened a draft PR as per guidelines to indicate I started working on it). I planned to extend the code that does automatic imputation to work directly with `nan` (instead of masks) when the observed variable is a Mutable/Constant data one. After all, that was the overall goal - to enable usage of `pm.Data()` with automatic imputation.

I'm not 100% sure what will happen with posterior predictive, but I think it should be the same as when passing a masked array to `observed` directly. I would have assumed that works - could you explain why you think it will be an issue with MutableData? (I can also test myself later today).

tests/test_data.py
kamicollo: Unfortunately, we can't - you cannot have an integer NumPy array with `nan` values, ie. this throws an error: `np.array([1,2,3, np.nan[. dtype=int)`. That's because `nan` is strictly a float concept. So yes, we would not be able to allow users pass an integer masked array into `pm.Data()`. If they want to benefit from automatic imputation, they can today (and will be after this PR) pass a masked integer array directly into `observed` parameter of an RV. I have an error message that explains the options in the code: 

https://github.com/pymc-devs/pymc/pull/6645/commits/28d15f809ddfa8d5d490e30107194d2d1b7d7228#diff-823b37f218229d363550b4cc387cfffa180c5c6e0e5ad0e174f2f0be7aa4692aR102

```
if isinstance(data, np.ma.MaskedArray):
        if "int" in str(data.dtype):
            raise TypeError(
                "Masked integer arrays (integer type datasets with missing values) are not supported by pm.Data() / pm.Model.set_data() at this time. \n"
                "Consider if using a float type fits your use case. \n"
                "Alternatively, if you want to benefit from automatic imputation in pyMC, pass a masked array directly to `observed=` parameter when defining a distribution."
            )
        else:
            ret = data.filled(fill_value=np.nan)
```

ricardoV94:
The issue is what happens when you change Mutabledata between sampling and posterior predictive. PyMC will resample any variable that depends on a MutableData variable that has changed. This would include the imputed variables.

If that makes sense than it's fine of course. I am just not sure what people need when they use MutableData for both imputation and prediction.

ricardoV94:
And apologies, I didn't see you marked the PR as draft. No worries anyway :)

kamicollo:
No problem at all :)

I have added the rest of the PR which enables automatic imputation support for `pm.MutableData()` / `pm.ConstantData()` variables. Testing-wise, I added extra tests to `make_obs_var` function (which performs automatic imputation, and went through all existing automatic imputation tests and parameterized them to run 3 times (original numpy masked/nan array, MutableData variable, ConstantData variable). It would be great if you could let me know if you think any other tests are needed.

Regarding posterior predictive - good question. I looked at how pyMC does it right now (when a masked numpy array is passed to observed) for a simple model where a covariate is missing, and I can see that:

 * `sample_prior_predictive` yields: `X_missing`  and `X` in `prior` trace, and `X_observed` in `prior_predictive` trace.
 * `sample` yields `X_missing` and `X` in `posterior` trace.
 * `sample_posterior_predictive` yields: `X_observed` in `observed_data` trace; `X` and `X_observed` in `posterior_predictive` trace.

The `X` in `posterior` trace is the same as the `X` in the `posterior_predictive` trace for all indexes corresponding to missing data. It is different for non-missing values. That, by the way, makes me wonder if predicted values for `Y` are based on a mix of imputed values for `X` and observed values for `X` (which seems right) or predicted values of `X` (which seems wrong)...

So in a simple case, we don't seem to re-sample imputed values. However, if a user provides a new set of (missing) data, I would intuitively say it would make sense to resample the imputations, too. The missing entries in the new data are not the same as the old ones, after all.

Having said that - my current PR does not seem to work with `sample_posterior_predictive` and passing new data to a variable defined with `pm.MutableData()`:
 * If I run the initial model with missing data, and then pass new missing data via `model.set_data()`, it seems it gets ignored in the posterior predictive sampling
 * If I run the initial model without missing data and then pass new missing data, the posterior predictive sampling errors out.

I suspect it's because the automatic imputation logic only gets triggered when the model is created, and so in the second case pps fails because of nan values. I am unsure why the first case doesn't work - possibly because `set_data()` doesn't propagate to the underlying sub-tensors behind pm.MutableData()? Would you have immediate thoughts for possible solutions?

tests/test_data.py
ricardoV94: I wasn't clear. Is any error raised if user pass floats observed to discrete variables? I think it works just fine.

jonsedar:
Stumbling into this 18 months later, I'd like to be able to support the usual workflow, and allow for data to be missing in specified features:
1. Sample posterior (and posterior predictive, lets live a little) on in-sample dataset
2. Replace data with out-of-sample dataset and make forecast predictions

The auto-impute works great for the in-sample dataset, but when I want to replace data in the `Data` containers, there doesn't seem to be a way to replace the masked array.

E.g.  inside model context I want to replace `dfx_train` with `dfx_holdout`, but there seems no straightforward way to do this

```
...
xk_ma = np.ma.masked_array(dfx_train[fts_xk].values, mask=np.isnan(dfx_train[fts_xk].values))
xk_mu = pm.Normal('xk_mu', mu=0.0, sigma=1, dims='xk_nm')
xk = pm.Normal('xk', mu=xk_mu, sigma=1.0, observed=xk_ma, dims=('oid', 'xk_nm'))
...
```

Any ideas of alternative methods?

PR #6650, opened 2023-04-04
Remove `joint_logprob` function from `tests.logprob.utils`

### LINKED ISSUE #6619: Remove `joint_logprob` function from `tests.logprob.util` submodule
### Description

This function is not very helpful and only used in tests. We should be able to remove it, and use either `logp` or `factorized_joint_logprob` in the tests.

https://github.com/pymc-devs/pymc/blob/83cd926e572e710dc659ed4481de9e556b31bcf8/tests/logprob/utils.py#L50-L74

shreyas3156:
**What is this PR about?**
Closes #6619 

The `joint_logprob()` function in `tests.logprob.utils` uses `factorized_joint_logprob()` and only checks for summing over the dimensions of the logp calculated. It can therefore be replaced with `logp()` or `factorized_joint_logprob()` depending on the number of RVs to be calculated, with an additional statement to sum over the dimensions in the test cases.

**Checklist**

## Major / Breaking Changes
- None

## Maintenance
- Removing the method should somewhat decrease the functional overhead.

----
:books: Documentation preview :books:: https://pymc--6650.org.readthedocs.build/en/6650/

### DIFF
--- a/tests/logprob/test_basic.py
+++ b/tests/logprob/test_basic.py
@@ -60,19 +60,19 @@ from pymc.logprob.transforms import LogTransform
 from pymc.logprob.utils import rvs_to_value_vars, walk_model
 from pymc.pytensorf import replace_rvs_by_values
 from pymc.testing import assert_no_rvs
-from tests.logprob.utils import joint_logprob
 
 
-def test_joint_logprob_basic():
-    # A simple check for when `joint_logprob` is the same as `logprob`
+def test_factorized_joint_logprob_basic():
+    # A simple check for when `factorized_joint_logprob` is the same as `logprob`
     a = pt.random.uniform(0.0, 1.0)
     a.name = "a"
     a_value_var = a.clone()
 
-    a_logp = joint_logprob({a: a_value_var}, sum=False)
+    a_logp = factorized_joint_logprob({a: a_value_var})
+    a_logp_comb = tuple(a_logp.values())[0]
     a_logp_exp = logp(a, a_value_var)
 
-    assert equal_computations([a_logp], [a_logp_exp])
+    assert equal_computations([a_logp_comb], [a_logp_exp])
 
     # Let's try a hierarchical model
     sigma = pt.random.invgamma(0.5, 0.5)
@@ -81,7 +81,8 @@ def test_joint_logprob_basic():
     sigma_value_var = sigma.clone()
     y_value_var = Y.clone()
 
-    total_ll = joint_logprob({Y: y_value_var, sigma: sigma_value_var}, sum=False)
+    total_ll = factorized_joint_logprob({Y: y_value_var, sigma: sigma_value_var})
+    total_ll_combined = pt.add(*total_ll.values())
 
     # We need to replace the reference to `sigma` in `Y` with its value
     # variable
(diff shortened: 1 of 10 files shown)


tests/logprob/test_basic.py
```diff
@@ -58,16 +58,15 @@
 from pymc.logprob.basic import factorized_joint_logprob, icdf, joint_logp, logcdf, logp
 from pymc.logprob.utils import rvs_to_value_vars, walk_model
 from pymc.testing import assert_no_rvs
-from tests.logprob.utils import joint_logprob
 
 
 def test_joint_logprob_basic():
```
ricardoV94:
```suggestion
def test_factorized_joint_logprob_basic():
```

tests/logprob/test_basic.py
```diff
@@ -121,21 +122,25 @@ def test_joint_logprob_multi_obs():
...
+    logp_res = factorized_joint_logprob({x: x_val, y: y_val})
+    exp_logp = factorized_joint_logprob({x: x_val, y: y_val})
+    logp_res_comb = pt.sum([pt.sum(factor) for factor in logp_res.values()])
+    exp_logp_comb = pt.sum([pt.sum(factor) for factor in exp_logp.values()])
 
-    assert equal_computations([logp_res], [exp_logp])
+    assert equal_computations([logp_res_comb], [exp_logp_comb])
 
 
 def test_joint_logprob_diff_dims():
```
ricardoV94:
```suggestion
def test_factorized_joint_logprob_diff_dims():
```

tests/logprob/test_basic.py
```diff
@@ -205,9 +211,10 @@ def test_joint_logprob_subtensor():
     I_value_var = I_rv.type()
     I_value_var.name = "I_value"
 
-    A_idx_logp = joint_logprob({A_idx: A_idx_value_var, I_rv: I_value_var}, sum=False)
+    A_idx_logp = factorized_joint_logprob({A_idx: A_idx_value_var, I_rv: I_value_var})
```
ricardoV94: This test should be moved to `logprob/mixture.py` or removed if there's already an equivalent one

ricardoV94 (dismissed):
Looks good, just spotted some test name changes and one test that should be moved or removed.

tests/logprob/test_basic.py
```diff
@@ -58,16 +58,15 @@
...
 
 def test_joint_logprob_basic():
-    # A simple check for when `joint_logprob` is the same as `logprob`
+    # A simple check for when `factorized_joint_logprob` is the same as `logprob`
     a = pt.random.uniform(0.0, 1.0)
     a.name = "a"
     a_value_var = a.clone()
 
-    a_logp = joint_logprob({a: a_value_var}, sum=False)
+    a_logp = logp(a, a_value_var)
```
Armavica: Shouldn't this be `factorized_joint_logprob`?

ricardoV94: For single variables it's fine to use logp directly

ricardoV94: Ah it's suppose to be a test of factorized, then yes

Armavica: Yes but then doesn't the test compare two equal things?

tests/logprob/test_basic.py
shreyas3156: Some tests in `test_mixture.py` use indexing but none explicitly check for a valid indexing operation. So I think it is good to simply move it there.

ricardoV94 (approved):

PR #6651, opened 2023-04-05
Issue warning if RVs are present in derived probability graphs

ricardoV94:
This is problematic when using Potentials with derived quantities expressions, such as in:

```python
with pm.Model() as m:
  x = pm.Normal("x")
  y = pm.Potential("y", pm.logp(pm.Normal.dist(x).exp(), 1.0))  # raises new warning
  m.logp()  # contains a random variable (clone of x)
```

Where `x` is not be replaced by the value variable in the Potential expression of `y`, leading to the wrong model logp.

I couldn't think of a clean solution, so we just give an informative warning about the problem and suggestion on how to fix it.

Note that this is not a problem in `CustomDist` because the logp receives the value variables already instead of the random inputs.

### DIFF
--- a/pymc/logprob/basic.py
+++ b/pymc/logprob/basic.py
@@ -65,7 +65,33 @@ from pymc.logprob.utils import rvs_to_value_vars
 TensorLike: TypeAlias = Union[Variable, float, np.ndarray]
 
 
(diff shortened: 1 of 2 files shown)


michaelosthege (approved):
LGTM

PR #6658, opened 2023-04-07
Fix bug in random function of HalfStudent

Dhruvanshu-Joshi:
**What is this PR about?**
In this PR, we aim to solve the issue [6657](https://github.com/pymc-devs/pymc/issues/6657). Originally what was passed as ```sigma``` is now passed by keyword as ```scale```.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- Fixes the bug as found in [Discourse_link](https://discourse.pymc.io/t/pm-halfstudentt-dist-failing-to-give-correct-shape/11821)

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6658.org.readthedocs.build/en/6658/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2501,8 +2501,8 @@ class HalfStudentTRV(RandomVariable):
     _print_name = ("HalfStudentT", "\\operatorname{HalfStudentT}")
 
     @classmethod
-    def rng_fn(cls, rng, nu, sigma, size=None) -> np.ndarray:
-        return np.asarray(np.abs(stats.t.rvs(nu, sigma, size=size, random_state=rng)))
+    def rng_fn(cls, rng, nu, scale=None, size=None) -> np.ndarray:
+        return np.asarray(np.abs(stats.t.rvs(nu, scale, size=size, random_state=rng)))
 
 
 halfstudentt = HalfStudentTRV()
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2501,8 +2501,8 @@ class HalfStudentTRV(RandomVariable):
     _print_name = ("HalfStudentT", "\\operatorname{HalfStudentT}")
 
     @classmethod
-    def rng_fn(cls, rng, nu, scale=None, size=None) -> np.ndarray:
-        return np.asarray(np.abs(stats.t.rvs(nu, scale, size=size, random_state=rng)))
+    def rng_fn(cls, rng, nu, sigma, size=None) -> np.ndarray:
+        return np.asarray(np.abs(stats.t.rvs(nu, scale=sigma, size=size, random_state=rng)))
 
 
 halfstudentt = HalfStudentTRV()
--- a/tests/distributions/test_continuous.py
+++ b/tests/distributions/test_continuous.py
@@ -1749,6 +1749,22 @@ class TestStudentT(BaseTestDistributionRandom):
     ]
 
 
+class TestHalfStudentT(BaseTestDistributionRandom):
+    def halfstudentt_rng_fn(self, df, loc, scale, size, rng):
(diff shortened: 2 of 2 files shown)


ricardoV94:
Oops #6654 seems to not be working

pymc/distributions/continuous.py
```diff
@@ -2501,8 +2501,8 @@ class HalfStudentTRV(RandomVariable):
     _print_name = ("HalfStudentT", "\\operatorname{HalfStudentT}")
 
     @classmethod
-    def rng_fn(cls, rng, nu, sigma, size=None) -> np.ndarray:
-        return np.asarray(np.abs(stats.t.rvs(nu, sigma, size=size, random_state=rng)))
+    def rng_fn(cls, rng, nu, scale=None, size=None) -> np.ndarray:
+        return np.asarray(np.abs(stats.t.rvs(nu, scale, size=size, random_state=rng)))
```
ricardoV94: Should be something like this.

```suggestion
    def rng_fn(cls, rng, nu, sigma, size=None) -> np.ndarray:
        return np.abs(stats.t.rvs(nu, scale=sigma, size=size, random_state=rng))
```

Hopefully one of our existing tests now fails and will need to be updated. 

Otherwise it means this functionality was not covered by our existing tests and we need a new test or improve the existing ones.

ricardoV94 (changes requested):

pymc/distributions/continuous.py
Dhruvanshu-Joshi: Hey @ricardoV94 I have made changes according to the suggestions. Hopefully the tests pass but if they don't, I'll work on adding the tests. Also from what I understand, the changes here might not be able to eliminate the errors raised in the discussion [here](https://discourse.pymc.io/t/pm-halfstudentt-dist-failing-to-give-correct-shape/11821/5)  and [here](https://discourse.pymc.io/t/pm-halfstudentt-dist-failing-to-give-correct-shape/11821/6) where we want to pass keyword ```scale``` instead of ```sigma``` to the ```pm.HalfStudentT```.

ricardoV94: No, the idea is not to pass scale. The problem was sigma was being passed as loc (which is the second unnamed argument) to the scipy function by mistake. It should have been passed as scale, as this PR does.

ricardoV94:
The tests not being triggered should be fixed once https://github.com/pymc-devs/pymc/pull/6659 is merged

ricardoV94:
@Dhruvanshu-Joshi If you rebase from main and push again the tests should run

ricardoV94:
No test failed which means we need to tweak the HalfStudentT random test to cover the change (it should fail if the change hadn't been done or reverted)

ricardoV94:
Since this is blocking a release I went ahead and added a test. 

Apparently the HalfSTudentT random method was not being tested at all!

Dhruvanshu-Joshi:
Hey @ricardoV94 . I was working on the test but was trying a different approach wherein I tried to generate the sample using pm.HalfStudentT rv and compare the mean and std of the sample with the theoretical mean and std. I am still pretty naive at generating tests and will go through the codebase again to have a better grasp and be quick to complete the task so the it does not block any future release. Thank You for your help.

ricardoV94:
@Dhruvanshu-Joshi There was absolutely nothing wrong with the time you were taking! 

In the end the test solution was easier: just use the framework we have in place to test the random methods. I assumed this was already being used for the current RV so I didn't give you the proper advice.

If anything it was my fault :)

ricardoV94 (approved):

Dhruvanshu-Joshi:
Thanks @ricardoV94 . I'll go through the framework to better improve my understanding🙂

Issue #6631, opened 2023-03-29
Implement more types of measurable elemwise transformations


ricardoV94:
We already have support for
- [x] add (and subtraction via addition)
- [x] mul (and division and neg via multiplication)
- [x] pow (inludes reciprocal, sqrt, sqr, etc...)
- [x] exp
- [x] log
- [x] abs

Other operations we could support
- [x] cosh, sinh and tanh (maybe other?)
- [x] erf, erfc, erfcx?
- [x] arcsinh, arctanh (see https://discourse.pymc.io/t/the-sinh-arcsinh-normal-distribution/12136)
- [x] arccosh is multivalued, or complex? Didn't have time to dig

https://github.com/pymc-devs/pymc/blob/f3ce16f2606f523137c27466069f1ab737626f21/pymc/logprob/transforms.py#L339

Other operations we can support via rewrites to already supported forms
- [x] exp2, log2, log10
- [x] log1p
- [x] expm1
- [x] log1mexp
- [x] log1pexp (softplus)
- [x] sigmoid

Example of such rewrites: https://github.com/pymc-devs/pymc/blob/f3ce16f2606f523137c27466069f1ab737626f21/pymc/logprob/transforms.py#L467-L483

LukeLB:
@ricardoV94 is this just a continuation of the work done here https://github.com/pymc-devs/pymc/pull/6414/files? If so I can follow what I did there and knock a few of these off

ricardoV94:
Yes, 👍

LukeLB:
I've created Transform classes for cosh, sinh, tanh, erf, erfc however there doesn't appear to be an inverse function available for erfcx in pytensor or `pymc.math`. Any suggestions on what to do here? It feels like writing a new function in pytensor for erfcxinv may be overkill for this issue...

ricardoV94:
You can check if it exists in `pytensor.tensor`. `pymc.math` is just a restricted subset of those.

If not you can play with the equivalence between erfcx and erfc

LukeLB:
So if I rewrite it into that form I can get the foward function, but I don't understand how to get the backward function from that? This discussion on the inverse of erfcx in Matlab seems to suggest writing a specific functon https://www.mathworks.com/matlabcentral/answers/302915-inverse-of-erfcx-scaled-complementary-error-function

LukeLB:
Not to worry, it did have me confused for a bit there! 

[...]

PR #6664, opened 2023-04-10
Derive logprob for hyperbolic and error transformations

LukeLB:
**What is this PR about?**
I have implemented additional Elemwise transformations as suggested in issue #6631. Specifically, this pull request adds cosh, sinh, tanh, erf, erfc, and erfcx functions. I plan to address the other suggested transformations in a separate pull request, as they require a more significant rewrite of existing functions. However, if it is preferred to include them all in one pull request, I'm happy to do so.

Please note that this is still a work in progress, and I have not yet written any tests for the new Transforms. I would appreciate some guidance on how to design these tests as its not clear to me what I should be testing them against.

Also for the erfcx transform it would be great double check my math is correct, for the backward I have rewrote a matlab function and for the log jacobian determinant I used wolfram alpha to get the derivative of erfcx. 
...

**Checklist**

## Major / Breaking Changes
- New elemwise transforms 
- Cleaned up the if block in `find_measureable_transforms()` as it was getting quite large

## New features
Transforms for:
- cosh
- sinh
- tanh
- erf
- erfc
- erfcx 

## Bugfixes
- NA

## Documentation
- Haven't added any in this PR 

## Maintenance
- NA 

----
:books: Documentation preview :books:: https://pymc--6664.org.readthedocs.build/en/6664/

### DIFF
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -42,6 +42,7 @@ from typing import Callable, Dict, List, Optional, Sequence, Tuple, Union
 import numpy as np
 import pytensor.tensor as pt
 
+from pytensor import scan
 from pytensor.gradient import DisconnectedType, jacobian
 from pytensor.graph.basic import Apply, Node, Variable
 from pytensor.graph.features import AlreadyThere, Feature
@@ -791,9 +792,9 @@ class ErfcxTransform(RVTransform):
                 2 * prior_result * pt.erfcx(prior_result) - 2 / pt.sqrt(np.pi)
             )
 
-        result, updates = pytensor.scan(
+        result, updates = scan(
             fn=calc_delta_x,
-            outputs_info=at.ones_like(x),
+            outputs_info=pt.ones_like(x),
             sequences=value,
             non_sequences=x,
             n_steps=k,
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -419,7 +419,7 @@ def measurable_transform_logprob(op: MeasurableTransform, values, *inputs, **kwa
         jacobian = jacobian.sum(axis=tuple(range(-ndim_supp, 0)))
 
     # The jacobian is used to ensure a value in the supported domain was provided
-    return pt.switch(pt.isnan(input_logprob + jacobian), -np.inf, input_logprob + jacobian)
+    return pt.switch(pt.isnan(jacobian), -np.inf, input_logprob + jacobian)
 
 
 @_logcdf.register(MeasurableTransform)
@@ -737,7 +737,6 @@ class CoshTransform(RVTransform):
 
     def backward(self, value, *inputs):
         return pt.arccosh(value)
-
 
 
 class TanhTransform(RVTransform):
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -719,6 +719,7 @@ measurable_ir_rewrites_db.register(
 
 class SinhTransform(RVTransform):
     name = "sinh"
+    ndim_supp = 0
 
     def forward(self, value, *inputs):
         return pt.sinh(value)
@@ -726,12 +727,10 @@ class SinhTransform(RVTransform):
     def backward(self, value, *inputs):
         return pt.arcsinh(value)
 
-    def log_jac_det(self, value, *inputs):
-        return pt.log(pt.cosh(value))
-
 
 class CoshTransform(RVTransform):
     name = "cosh"
+    ndim_supp = 0
 
     def forward(self, value, *inputs):
         return pt.cosh(value)
@@ -739,21 +738,19 @@ class CoshTransform(RVTransform):
     def backward(self, value, *inputs):
         return pt.arccosh(value)
 
-    def log_jac_det(self, value, *inputs):
-        return pt.log(pt.sinh(value))
+    # def log_jac_det(self, value, *inputs):
+    #     return pt.log(pt.sinh(value))
 
 
 class TanhTransform(RVTransform):
     name = "tanh"
+    ndim_supp = 0
 
     def forward(self, value, *inputs):
         return pt.tanh(value)
 
     def backward(self, value, *inputs):
         return pt.arctanh(value)
-
-    def log_jac_det(self, value, *inputs):
-        return pt.log(1 / pt.cosh(value))
 
 
 class ErfTransform(RVTransform):
@@ -769,6 +766,7 @@ class ErfTransform(RVTransform):
 
 class ErfcTransform(RVTransform):
     name = "erfc"
+    ndim_supp = 0
 
     def forward(self, value, *inputs):
         return pt.erfc(value)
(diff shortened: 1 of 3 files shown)


pymc/logprob/transforms.py
```diff
@@ -682,6 +710,94 @@ def find_measurable_transforms(fgraph: FunctionGraph, node: Node) -> Optional[Li
...
+
+    def forward(self, value, *inputs):
+        return pt.erfcx(value)
+
+    def backward(self, value, tol=1e-10, max_iter=100):
+        # computes the inverse of erfcx, this was adapted from
+        # https://tinyurl.com/4mxfd3cz
+        x = pt.switch(value <= 1, 1.0 / (value * pt.sqrt(np.pi)), -pt.sqrt(pt.log(value)))
+        iter_count = 0
+        while iter_count < max_iter:
```
ricardoV94: You will have to write this using a pytensor scan. We can't use vanilla python for loops or if statements with Pytensor variables:

https://pytensor.readthedocs.io/en/latest/library/scan.html

LukeLB: Ah my bad I thought that was only when accessing the tensors directly with the loop. I'll rewrite.

LukeLB:
@ricardoV94 I've been working on 2. and the test now runs however its throwing an assertion error. The test is:
```python
@pytest.mark.parametrize("transform", [ErfTransform])
def test_erf_logp(transform):
    base_rv = pt.random.normal(0.5, 1, name="base_rv")  # Something not centered around 0 is usually better
    rv = pt.erf(base_rv)
    vv = rv.clone()
    rv_logp = joint_logprob({rv: vv})

    transform = transform()
    expected_logp = joint_logprob({rv: transform.backward(vv)}) + transform.log_jac_det(vv)

    vv_test = np.array(0.25)  # Arbitrary test value
    np.testing.assert_almost_equal(
    rv_logp.eval({vv: vv_test}),
    expected_logp.eval({vv: vv_test}),
    )
```
This gives the assertion error:
```
E       AssertionError: 
E       Arrays are not almost equal to 7 decimals
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 0.06346299
E       Max relative difference: 0.07601085
E        x: array(-0.898383)
E        y: array(-0.83492)
```
They're close but still quite a difference. I'm not sure if this is the way I've written the logp in the test or that the internal transform functions are wrong. Any ideas?

ricardoV94:
Oh my example was wrong. You want to compare with the `base_rv + jacobian`, not `rv + jacobian`:

This passes locally:
```python
import numpy as np
import pytensor.tensor as pt
from pymc.logprob.basic import logp
from pymc.logprob.transforms import ErfTransform

base_rv = pt.random.normal(0.5, 1, name="base_rv")  # Something not centered around 0 is usually better
rv = pt.erf(base_rv)

vv = rv.clone()
rv_logp = logp(rv, vv)

transform = ErfTransform()
expected_logp = logp(base_rv, transform.backward(vv)) + transform.log_jac_det(vv)

vv_test = np.array(0.25)  # Arbitrary test value
np.testing.assert_almost_equal(
    rv_logp.eval({vv: vv_test}),
    expected_logp.eval({vv: vv_test}),
)
```

LukeLB:
Cheers will make the change!

LukeLB:
> Oh my example was wrong. You want to compare with the `base_rv + jacobian`, not `rv + jacobian`:
> 
> This passes locally:
> 
> ```python
> import numpy as np
> import pytensor.tensor as pt
> from pymc.logprob.basic import logp
> from pymc.logprob.transforms import ErfTransform
> 
> base_rv = pt.random.normal(0.5, 1, name="base_rv")  # Something not centered around 0 is usually better
> rv = pt.erf(base_rv)
> 
> vv = rv.clone()
> rv_logp = logp(rv, vv)
> 
> transform = ErfTransform()
> expected_logp = logp(base_rv, transform.backward(vv)) + transform.log_jac_det(vv)
> 
> vv_test = np.array(0.25)  # Arbitrary test value
> np.testing.assert_almost_equal(
>     rv_logp.eval({vv: vv_test}),
>     expected_logp.eval({vv: vv_test}),
> )
> ```

LukeLB:
Woops clicked the wrong button didn't mean to close! 

So test 2. now works for all Transforms however I'm having an issue with the `check_jacobian_det` which gives an assertion error for all transforms. Does this suggest that the math is wrong?

Note for test 2. I had to make changes to the test by adding a switch statement and editing the switch statement on line 416 in transforms.py to take input_logprob AND jacobian because of the descrepency of returning nans vs. -infs as if input_logprob is nan then this also returns nan and not -inf.

ricardoV94:
> Woops clicked the wrong button didn't mean to close!

No problem :)

> So test 2. now works for all Transforms however I'm having an issue with the `check_jacobian_det` which gives an assertion error for all transforms. Does this suggest that the math is wrong?

I think there must have been an error in your `log_jac_det` expression. I tweaked the default implementation in `RVTransform` so that it works for both elemwise and vector transforms, and (after allowing the test to accept `nan`) it passes, whereas with your hand-written implementation it did not.

I think it's fine to use the default implementation (for the cases where it works).

LukeLB:
Great, I'll take a look at what you did and see if I can try and implement it with the other transforms.

LukeLB:
OK all done, all tests pass.

pymc/logprob/transforms.py
```diff
@@ -682,6 +717,90 @@ def find_measurable_transforms(fgraph: FunctionGraph, node: Node) -> Optional[Li
...
+    ndim_supp = 0
+
+    def forward(self, value, *inputs):
+        return pt.cosh(value)
+
+    def backward(self, value, *inputs):
+        return pt.arccosh(value)
+
+    # def log_jac_det(self, value, *inputs):
+    #     return pt.log(pt.sinh(value))
```
ricardoV94:
```suggestion
```

pymc/logprob/transforms.py
```diff
@@ -391,7 +419,7 @@ def measurable_transform_logprob(op: MeasurableTransform, values, *inputs, **kwa
         jacobian = jacobian.sum(axis=tuple(range(-ndim_supp, 0)))
 
     # The jacobian is used to ensure a value in the supported domain was provided
-    return pt.switch(pt.isnan(jacobian), -np.inf, input_logprob + jacobian)
+    return pt.switch(pt.isnan(input_logprob + jacobian), -np.inf, input_logprob + jacobian)
```
ricardoV94: Can we revert this change?

ricardoV94 (commented):
Looks good. I don't know why the coverage shows some of the new transforms not being covered, just a fluke?

I just have a question about a change below.

pymc/logprob/transforms.py
LukeLB: Potentially let me check. The reason I did that was because it meant that we return `-np.inf` consistently when `input_logprob = nan`, which is the case for some of the transforms.

LukeLB: Okay seems that isn't the case anymore and tests pass with the reverted change

ricardoV94 (approved):

ricardoV94:
Awesome work @LukeLB! Looking forward to your next PR :)

Issue #244, opened 2013-05-08
attributeError in Binomial variables


fonnesbeck:
Using a Binomial in as a non-observed stochastic results in an AttributeError being raised. For example:

```
quitters_pred = Binomial('quitters_pred', n=n, p=p_quit)
```

yields:

```
AttributeError: <pymc.quickclass.Binomial object at 0x10755f650> does not have a value for any of: ['mode']
```

whereas an observed variable does not:

```
quitters = Binomial('quitters', n=n, p=p_quit, observed=y) 
```

jsalvatier:
That is weird, can you check to see if you haven't reassigned Binomial or
something? Binomial does have tests that don't use observed:
https://github.com/pymc-devs/pymc/blob/pymc3/pymc/tests/test_distributions.py

fonnesbeck:
I figured it out. Some of the discrete variables were missing mode attributes. Should be fixed in the commit above.

PR #6665, opened 2023-04-11
Bump Pytensor dependency

ricardoV94:
Trying to use a minimum version range, so we don't need to update for bugfixes... WDYT?

Marking this as major as the Latex/Str representation of the variables now use the long name due/thanks to #244

----
:books: Documentation preview :books:: https://pymc--6665.org.readthedocs.build/en/6665/


twiecki:
`E       FutureWarning: sgn is deprecated and will stop working in the future, use sign instead.`

twiecki (approved):

michaelosthege (approved):

PR #6668, opened 2023-04-13
`CustomDist` and `Simulator` no longer require `class_name` when creating a `dist`

ricardoV94:
Apparently cloudpickle is happy with classes with the same name, so we don't need to request them from users...

```python
import cloudpickle

class A:
    def __init__(self, a):
        self.a = a
        
a = A(5)

class A:
    def __init__(self, b):
        self.b = b
        
b = A(3)

a_, b_ = cloudpickle.loads(cloudpickle.dumps((a, b)))

print(a_.a, b_.b)  # 5, 3
print(type(a_) is type(b_))  # False
print(type(a))  # <class '__main__.A'>
print(type(b))  # <class '__main__.A'>
print(type(a_) is A)  # False
print(type(b_) is A)  # True
```

----
:books: Documentation preview :books:: https://pymc--6668.org.readthedocs.build/en/6668/

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -684,9 +684,11 @@ class _CustomSymbolicDist(Distribution):
             def custom_dist_logp(op, values, size, *params, **kwargs):
                 return logp(values[0], *params[: len(dist_params)])
 
-        @_logcdf.register(rv_type)
-        def custom_dist_logcdf(op, value, size, *params, **kwargs):
-            return logcdf(value, *params[: len(dist_params)])
+        if logcdf is not None:
+
+            @_logcdf.register(rv_type)
+            def custom_dist_logcdf(op, value, size, *params, **kwargs):
+                return logcdf(value, *params[: len(dist_params)])
 
         @_moment.register(rv_type)
         def custom_dist_get_moment(op, rv, size, *params):
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -488,7 +488,6 @@ class _CustomDist(Distribution):
     def dist(
         cls,
         *dist_params,
-        class_name: str,
         logp: Optional[Callable] = None,
         logcdf: Optional[Callable] = None,
         random: Optional[Callable] = None,
@@ -496,6 +495,7 @@ class _CustomDist(Distribution):
         ndim_supp: int = 0,
         ndims_params: Optional[Sequence[int]] = None,
         dtype: str = "floatX",
+        class_name: str = "CustomDist",
         **kwargs,
     ):
         dist_params = [as_tensor_variable(param) for param in dist_params]
@@ -523,7 +523,6 @@ class _CustomDist(Distribution):
 
         return super().dist(
             dist_params,
-            class_name=class_name,
             logp=logp,
             logcdf=logcdf,
             random=random,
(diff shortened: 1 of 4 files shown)


lucianopaz:
@ricardoV94, is this linked to an open issue? Or is the change just to simplify the API? Anyway, I will need to read up on what is happening under the hood to understand what the problematic API was, how this is an improvement, and why `cloudpickle` makes this work.

ricardoV94:
For more context,

I think I borrowed this constraint from Simulator design. We started implementing it prior to the switch to cloudpickle and I think we were having troubles implementing the dynamic class in a way that was serializable, and one of the things I tried (without remembering now if it even mattered in the end) was to request a unique `class_name`

lucianopaz:
So I've dug a bit into `cloudpickle` and how it manages to handle this case. They basically opt to pickle the objects by value, instead of by reference. That means that `a.__class__` and `b.__class__` values are pickled too. They do this in a very cool way that I haven't spent enough time to understand fully. They seem to leverage the information in the `__code__` objects to serialize and rebuild everything:

```python
>>> import dis
>>> dis.disco(cloudpickle.loads(cloudpickle.dumps(a)).__class__.__init__.__code__)
3           0 LOAD_FAST                1 (a)
            2 LOAD_FAST                0 (self)
            4 STORE_ATTR               0 (a)
            6 LOAD_CONST               0 (None)
            8 RETURN_VALUE
>>> dis.disco(cloudpickle.loads(cloudpickle.dumps(b)).__class__.__init__.__code__)
9           0 LOAD_FAST                1 (b)
            2 LOAD_FAST                0 (self)
            4 STORE_ATTR               0 (b)
            6 LOAD_CONST               0 (None)
            8 RETURN_VALUE
```

The class name is indeed the same for `A` and `B`, but the underlying class object isn't garbage collected until `a` is out of scope. That makes `pickle` raise an informative error when you try to pickle `a`:

```python
>>> import pickle
>>> pickle.loads(pickle.dumps(a))
PicklingError                             Traceback (most recent call last)
Cell In[26], line 1
----> 1 pickle.loads(pickle.dumps(a))

PicklingError: Can't pickle <class '__main__.A'>: it's not the same object as __main__.A
```

pymc/distributions/distribution.py
```diff
@@ -979,36 +968,36 @@ def __new__(
         dist_params = cls.parse_dist_params(dist_params)
         cls.check_valid_dist_random(dist, random, dist_params)
         if dist is not None:
+            kwargs.setdefault("class_name", f"CustomSymbolicDist_{name}")
             return _CustomSymbolicDist(
                 name,
                 *dist_params,
-                class_name=name,
```
lucianopaz, line 985: Wait, was old behavior to use the RV name as the class name?

lucianopaz (approved):
This looks good to me

pymc/distributions/distribution.py
ricardoV94: We were doing `"CustomDist_{class_name}", so yes, if a user created one in a model context it was `CustomDIst_mu` for example. Pretty meh

ricardoV94: Did you have a concern in mind?

lucianopaz: No concern, just checking my understanding

ricardoV94:
Thanks for the review 🙏 @lucianopaz

PR #6671, opened 2023-04-13
Improve numerical precision of discrete uniform and geometric ICDFs

### LINKED ISSUE #6670: BUG: ICDF implementation for the discrete geometric distribution fails some tests.
### Describe the issue:

The ICDF function for the discrete geometric distribution involves the ceiling function applied on a ratio of two values which is a floating point number. If this ratio is close to an integer, a small noise added to it even in the smallest digit of precision can sway the result of the ceiling function to the other side resulting in a wrong ICDF value.

The example code below is for an example of a Geometric distribution with `p=0.99`, and CDF value `0.9999` whose ICDF is expected to be `2`. Within the ICDF function, defining  the ratio `a = pt.log1p(-value) / pt.log1p(-p)`
```
    Ratio a before ceil: 2.000000000000025
    Ratio a after ceil: 3.0
    Expected value after ceil: 2
```
The tiny numerical error `000000000000025` contributes to the wrong value after ceil.

I also have a solution to this, which involves truncating the floating point value to a certain number of digits before using the ceiling function on it. This gets rid of the sensitivity to the smallest of noise. I will push a commit with this fix shortly.

```python
import pytensor as pt
import pymc as pm
from pymc.logprob.basic import icdf
from pymc.pytensorf import inputvars
import numpy as np

value = 0.9999 # CDF value for n=2
p = 0.99 # Parameter p of the discrete Geometric distribution.

# The below ratio 'a' is used inside the ICDF function:
a = np.log1p(-value) / np.log1p(-p)

print(
    f"""
    Ratio before ceil: {a}
    Ratio after ceil: {np.ceil(a)}
    Expected value after ceil: 2
    """
)

dist = pm.Geometric.dist(p=p)

dist_icdf = icdf(dist, pt.tensor.type.TensorType(dtype='float64', shape=[])('value'))
dist_icdf_fn = pt.function(list(inputvars(dist_icdf)), dist_icdf)

assert dist_icdf_fn(value) == 2
```

### Error message:

```shell
AssertionError
```

### PyMC version information:

pytensor: 2.10.1
Python: 3.10.10
PyMC: latest commit 2a324bc04d65637f427858c519f2eb9c247ebc0b

### Context for the issue:

_No response_

gokuld:
**What is this PR about?**
Improve numerical precision issues for the ICDF function for the discrete geometric distribution.

Actually, this is an issue in general, wherever the ceiling function is used on a floating point number to get the ICDF value. Thus this fix for the precision issue applies to most discrete ICDF functions, and not only the Geometric distribution.

Details are outlined in this issue https://github.com/pymc-devs/pymc/issues/6670 .

Closes #6670

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- Fix numerical precision issues for the discrete ICDF functions.

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6671.org.readthedocs.build/en/6671/

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -48,7 +48,7 @@ from pymc.distributions.dist_math import (
 from pymc.distributions.distribution import Discrete
 from pymc.distributions.mixture import Mixture
 from pymc.distributions.shape_utils import rv_size_is_none
-from pymc.logprob.basic import logp
+from pymc.logprob.basic import logcdf, logp
 from pymc.math import sigmoid
 from pymc.pytensorf import floatX, intX
 from pymc.vartypes import continuous_types
@@ -823,6 +823,10 @@ class Geometric(Discrete):
 
     def icdf(value, p):
         res = pt.ceil(pt.log1p(-value) / pt.log1p(-p)).astype("int64")
+        res_1m = pt.maximum(res - 1, 0)
+        dist = pm.Geometric.dist(p=p)
+        value_1m = pt.exp(logcdf(dist, res_1m))
+        res = pt.switch(value_1m >= value, res_1m, res)
         res = check_icdf_value(res, value)
         return check_icdf_parameters(
             res,
@@ -1060,6 +1064,11 @@ class DiscreteUniform(Discrete):
 
     def icdf(value, lower, upper):
         res = pt.ceil(value * (upper - lower + 1)).astype("int64") + lower - 1
+        res_1m = pt.maximum(res - 1, lower)
+        dist = pm.DiscreteUniform.dist(lower=lower, upper=upper)
+        value_1m = pt.exp(logcdf(dist, res_1m))
+        res = pt.switch(value_1m >= value, res_1m, res)
+
         res = check_icdf_value(res, value)
         return check_icdf_parameters(
             res,


gokuld:
`test_geometric` and `test_discrete_unif` passed locally on my machine. 
Let me check what error I have made and get back on this.

pymc/distributions/discrete.py
```diff
@@ -823,6 +824,10 @@ def logcdf(value, p):
 
     def icdf(value, p):
         res = pt.ceil(pt.log1p(-value) / pt.log1p(-p)).astype("int64")
+        res_1m = pt.maximum(res - 1, 0)
+        dist = pm.Geometric.dist(p=p)
+        value_1m = pt.exp(logcdf(dist, res_1m))
+        res = ifelse(value_1m >= value, res_1m, res)
```
ricardoV94: You have to use a `switch` here, `ifelse` will only work for scalar cases

ricardoV94 (changes requested):

ricardoV94:
Make sure you pass `n_samples=-1` so all combinations are tried and not just a random subset

gokuld:
Thanks, replaced `ifelse` with `switch`.

Even with `n_samples=-1`. `pytest -v tests/distributions/test_discrete.py -k 'test_geometric or test_discrete_unif'` worked locally on my machine. Perhaps it has to do something with switch / ifelse

pymc/testing.py
```diff
@@ -574,7 +574,7 @@ def check_icdf(
     # Test pymc and scipy distributions match for values and parameters
     # within the supported domain edges (excluding edges)
     domains = paramdomains.copy()
-    domain = Domain([0, 0.1, 0.5, 0.75, 0.95, 0.99, 1])  # Values we test the icdf at
+    domain = Domain([0, 0.1, 0.5, 0.75, 0.95, 0.99, 0.9999, 1])  # Values we test the icdf at
```
ricardoV94: Do the old implementations fail with just 3 9's? I am wary of testing such an extreme value by default because it puts a high burden on future icdfs on being very accurate for extreme values. We know these things must always break at some point, specially for unbounded distributions.

```suggestion
    domain = Domain([0, 0.1, 0.5, 0.75, 0.95, 0.99, 0.999, 1])  # Values we test the icdf at
```

ricardoV94: If you want to cover the changes in this PR, you can add a small test where you just evaluate the icdf at the values that used to fail for the two distributions

ricardoV94:
> Thanks, replaced `ifelse` with `switch`.
> 
> Even with `n_samples=-1`. `pytest -v tests/distributions/test_discrete.py -k 'test_geometric or test_discrete_unif'` worked locally on my machine. Perhaps it has to do something with `switch` / `ifelse`

Perfect

pymc/testing.py
gokuld: > If you want to cover the changes in this PR, you can add a small test where you just evaluate the icdf at the values that used to fail for the two distributions

To clarify my understanding, you mean I add the specific test for those distributions that fail and not something general that applies to all future distributions right?

gokuld: > Do the old implementations fail with just 3 9's? I am wary of testing such an extreme value by default because it puts a high burden on future icdfs on being very accurate for extreme values. We know these things must always break at some point, specially for unbounded distributions.

I added `0.9999`, since as outlined in https://github.com/pymc-devs/pymc/issues/6670,
for a Geometric distribution with `p=0.99`, and CDF value `0.9999` whose ICDF is expected to be `2` but it gives `3` as the result. It works fine where the ICDF of `0.99` is `1`. To summarize this in a table:

`p=0.99`
| Value | CDF | expected ICDF of the CDF | returned ICDF of the CDF | Bug |
| --- | --- | --- | --- | --- |
| 1 | 0.99 | 1 | 1| No |
| 2 | 0.9999 | 2 | 3 | Yes |

gokuld: I am guessing it will pass the test for `0.999`, whose ICDF is still `1`. Instead of this I think I can remove this value altogether, and instead add the other type of test you were suggesting, that is `ICDF(CDF(k)) = k` for discrete distributions as I have added for the continuous cases.

ricardoV94: > To clarify my understanding, you mean I add the specific test for those distributions that fail and not something general that applies to all future distributions right?

Yes that's what I suggest. Also note that instead of tweaking the icdf value we can tweak the domain of the distribution parameters that is tested. Probably testing more extreme Geometric `p` values would also have shown the error?

Regardless, a separate test is also fine

gokuld: Thanks @ricardoV94 .

I have decided to do this, let me know if it makes sense:

With my latest commit, I have removed the value `0.9999` in the test for ICDF to not burden all future ICDF code regarding high accuracy.

`p=0.99` is already used, as defined by `Unit` in `testing.py`.

The ICDF tests will instead be covered by the self-consistency check for ICDF of discrete distributions, which I will add in this PR https://github.com/pymc-devs/pymc/pull/6642 . (Note that the self-consistency check for ICDF of _continuous_ distributions is already added in https://github.com/pymc-devs/pymc/pull/6642, and the _discrete_ self consistency check is blocked on the current PR because of this accuracy issue. Thus I will add it to https://github.com/pymc-devs/pymc/pull/6642 after this current PR is closed.)

ricardoV94: Sounds good

ricardoV94:
Is this ready to merge then?

ricardoV94 (approved):

gokuld:
> Is this ready to merge then?

Yes, in my opinion.

ricardoV94:
Awesome, great work!

gokuld:
Thank you! :)

PR #6672, opened 2023-04-13
Fix typos in PolyaGamma's docstring

aleicazatti:
Fix typos and change a detail in PolyaGamma's docstring.

* A backslash and a backtick were missing in the math code block for the mean and the deployment looked like this:
Captura desde 2023-04-13 20-03-09

* The example plot wasn't deployed and it was the only distribution with `plt.style.use('seaborn-darkgrid')`. This commit imports arviz and uses `plt.style.use('arviz-darkgrid')` instead.

...

**Checklist**

----
:books: Documentation preview :books:: https://pymc--6672.org.readthedocs.build/en/6672/

### DIFF
--- a/conda-envs/environment-docs.yml
+++ b/conda-envs/environment-docs.yml
@@ -22,6 +22,7 @@ dependencies:
 - jupyter-sphinx
 - myst-nb
 - numpydoc
+- polyagamma
 - pre-commit>=2.8.0
 - pymc-sphinx-theme==0.13
 - sphinx-copybutton
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -3771,7 +3771,8 @@ class PolyaGamma(PositiveContinuous):
         import matplotlib.pyplot as plt
         import numpy as np
         from polyagamma import polyagamma_pdf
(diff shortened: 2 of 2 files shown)


ricardoV94:
The plot is still not visible in the preview: https://pymcio--6672.org.readthedocs.build/projects/docs/en/6672/api/distributions/generated/pymc.PolyaGamma.html

CC @OriolAbril

OriolAbril:
The rtd log has `ModuleNotFoundError: No module named 'polyagamma'`. The code snippet uses the polyagamma library but I guess it is not part of the docs conda env, so it can't actually generate the plot

ricardoV94:
> The rtd log has `ModuleNotFoundError: No module named 'polyagamma'`. The code snippet uses the polyagamma library but I guess it is not part of the docs conda env, so it can't actually generate the plot

Right, it's an optional dependency. Shall we merge as is, or is it a problem?

OriolAbril:
I wouldn't merge as is. Either add the dependency to the doc build env, or remove the code to generate the plot (the code isn't shown, so the only value of it being in the docstring is the plot it generates)

OriolAbril (approved):

Issue #6673, opened 2023-04-14
BUG: Combination covariance errors with WhiteNoise


dehorsley:
### Describe the issue:

Post v5.2.0, `gp.cov.Combination` covariance has extra checks (at [cov.py:189](https://github.com/pymc-devs/pymc/blame/aae97a2abf9496c9d8348c185a9ffcee294bef09/pymc/gp/cov.py#L189)) that fail with a `WhiteNoise` covariance.

Currently, the check for `input_dim` equality fails there is also a lurking problem with calculating `active_dims`. 

### Reproduceable code example:

```python
import pm
pm.gp.cov.Exponential(2, ls=1) + pm.gp.cov.WhiteNoise(sigma=1)
```

### Error message:
<details>

```
AttributeError                            Traceback (most recent call last)

<ipython-input-1-7257c4d70473> in <cell line: 3>()
      1 import pymc as pm
      2 print(pm.__version__)
----> 3 pm.gp.cov.Exponential(2, ls=1) + pm.gp.cov.WhiteNoise(sigma=1)

2 frames

/usr/local/lib/python3.9/dist-packages/pymc/gp/cov.py in __add__(self, other)
     83         if isinstance(other, numbers.Real):
     84             other = Constant(c=other)
---> 85         return Add([self, other])
     86 
     87     def __mul__(self, other):

/usr/local/lib/python3.9/dist-packages/pymc/gp/cov.py in __init__(self, factor_list)
    187 
    188         # Check if all input_dim are the same in factor_list
--> 189         input_dims = {factor.input_dim for factor in factor_list if isinstance(factor, Covariance)}
    190 
    191         if len(input_dims) != 1:

/usr/local/lib/python3.9/dist-packages/pymc/gp/cov.py in <setcomp>(.0)
    187 
    188         # Check if all input_dim are the same in factor_list
--> 189         input_dims = {factor.input_dim for factor in factor_list if isinstance(factor, Covariance)}
    190 
    191         if len(input_dims) != 1:

AttributeError: 'WhiteNoise' object has no attribute 'input_dim'
```
</details>

### PyMC version information:

<details>
pymc==5.2.0

installed via pip

OS: *
</details>

### Context for the issue:

Affects any use of latent GPs with noise.

dehorsley:
Likely also an issue with `Constant` covariance kernel

ricardoV94:
Closed via #6674

PR #6674, opened 2023-04-14
Fix `WhiteNoise` Covariance bug

dehorsley:
Since #6458, Covariance is now the base class for kernels/covariance functions with input_dim and active_dims, which does not include WhiteNoise and Constant kernels. While #6458 introduced the `BaseCovariance` class to make this distinction, it looks like `WhiteNoise` was inadvertently not updated to use the new base class.

Combination kernels, which use instance of `Covariance` to test for the presents of input_dim and active_dims, fail with a `WhiteNoise` kernel input.

## Bugfixes
- fix WhiteNoise use in Combination kernels (#6673)

----
:books: Documentation preview :books:: https://pymc--6674.org.readthedocs.build/en/6674/

### DIFF
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -386,7 +386,7 @@ class Constant(BaseCovariance):
             return pt.alloc(self.c, X.shape[0], Xs.shape[0])
 
 
-class WhiteNoise(Covariance):
+class WhiteNoise(BaseCovariance):
     r"""
     White noise covariance function.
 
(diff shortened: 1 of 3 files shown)


bwengals:
Hey @dehorsley, looks like tests are passing.  Ready for review+merge, or is there more you were planning on doing?

dehorsley:
Hi @bwengals, yup ready to merge!

bwengals (approved):
Looks great, thank you for this fix!

PR #6677, opened 2023-04-15
Added scale parameterization to Exponential

### LINKED ISSUE #6637: Add alternative scale parametrization to Exponential
### Description

scale = 1 / lam
PyMC currently uses `lam`, and others (e.g., Numpy) use `scale`

This should be relatively simple, here is an existing example of a distribution with multiple parametrizations:

https://github.com/pymc-devs/pymc/blob/380e68cebcab94d9266fd28a3daeae4a9201de2b/pymc/distributions/discrete.py#L126-L132

Test example: https://github.com/pymc-devs/pymc/blob/9a3f5040b46da694de9abbc92da6cf71566f1001/tests/distributions/test_discrete.py#L962-L966

See also docstrings examples of how multiple parametrizations are described.

manulpatel:
**What is this PR about?**
Fixes #6637 
Adds alternative **scale** parametrization to exponential function. scale = 1 / lam

**Checklist**

## Major / Breaking Changes
- ...
## New features
- Extended functionality of specifing alternative scale parameter in exponential.

 
## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- Relavant test cases added in test_continuous.py.

----
:books: Documentation preview :books:: https://pymc--6677.org.readthedocs.build/en/6677/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1347,13 +1347,22 @@ class Exponential(PositiveContinuous):
     ----------
     lam : tensor_like of float
         Rate or inverse scale (``lam`` > 0).
+    scale: tensor_like of float
+        Alternative parameter (scale = 1/lam).
     """
     rv_op = exponential
 
     @classmethod
-    def dist(cls, lam: DIST_PARAMETER_TYPES, *args, **kwargs):
-        lam = pt.as_tensor_variable(floatX(lam))
+    def dist(cls, lam=None, scale=None, *args, **kwargs):
+        if lam is not None and scale is not None:
+            raise ValueError("Incompatible parametrization. Can't specify both lam and scale.")
+        elif lam is None and scale is None:
+            raise ValueError("Incompatible parametrization. Must specify either lam or scale.")
 
+        if scale is not None:
+            lam = pt.reciprocal(scale)
+
+        lam = pt.as_tensor_variable(floatX(lam))
         # PyTensor exponential op is parametrized in terms of mu (1/lam)
         return super().dist([pt.reciprocal(lam)], **kwargs)
 
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1359,12 +1359,12 @@ class Exponential(PositiveContinuous):
         elif lam is None and scale is None:
             raise ValueError("Incompatible parametrization. Must specify either lam or scale.")
 
-        if scale is not None:
-            lam = pt.reciprocal(scale)
+        if scale is None:
+            scale = pt.reciprocal(lam)
 
-        lam = pt.as_tensor_variable(floatX(lam))
+        scale = pt.as_tensor_variable(floatX(scale))
         # PyTensor exponential op is parametrized in terms of mu (1/lam)
-        return super().dist([pt.reciprocal(lam)], **kwargs)
+        return super().dist([scale], **kwargs)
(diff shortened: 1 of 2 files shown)


pymc/distributions/continuous.py
```diff
@@ -1347,13 +1347,22 @@ class Exponential(PositiveContinuous):
...
     @classmethod
-    def dist(cls, lam: DIST_PARAMETER_TYPES, *args, **kwargs):
-        lam = pt.as_tensor_variable(floatX(lam))
+    def dist(cls, lam=None, scale=None, *args, **kwargs):
+        if lam is not None and scale is not None:
+            raise ValueError("Incompatible parametrization. Can't specify both lam and scale.")
+        elif lam is None and scale is None:
+            raise ValueError("Incompatible parametrization. Must specify either lam or scale.")
+
+        if scale is not None:
```
ricardoV94: We were already using scale internally (see super.dist() below) without naming it so. 

So lets flip it: `if scale is None: scale=reciprocal(lam)` instead. Otherwise we are doing two useless `reciprocal` when a user passes scale

tests/distributions/test_continuous.py
```diff
@@ -432,18 +432,43 @@ def test_exponential(self):
             {"lam": Rplus},
             lambda value, lam: st.expon.logpdf(value, 0, 1 / lam),
         )
+        check_logp(
```
ricardoV94: No need to check logp and logcdf, the new random test below is enough.

ricardoV94 (changes requested):
Thanks for opening the PR. I left some comments below

pymc/distributions/continuous.py
manulpatel: Yes I noticed that. But we if implement this, then we need to change **return** to: `return super().dist([scale], **kwargs)`

tests/distributions/test_continuous.py
manulpatel: Okay got it! Will remove them in next commit.

pymc/distributions/continuous.py
ricardoV94: Yeah we should do that, not a problem

manulpatel: Okay! So now we would be formally using scale and returning it.

tests/distributions/test_continuous.py
```diff
@@ -444,6 +444,19 @@ def test_exponential(self):
             lambda q, lam: st.expon.ppf(q, loc=0, scale=1 / lam),
         )
 
+    def test_exponential_wrong_arguments(self):
+        m = pm.Model()
```
ricardoV94: You don't need to use a PyMC model. Just create the variable with `pm.Exponential.dist`

manulpatel: Means `m=pm.Exponential.dist` ?

ricardoV94: You don't need `m` or a `Model` at all.

Just do:

```python
with pytest.raises(ValueError, match=msg):
   pm.Exponential.dist(lam=0.5, scale=5)
```

manulpatel: Got it. Noted.

ricardoV94 (approved):

ricardoV94:
Thanks @manulpatel, will merge once tests pass

PR #6678, opened 2023-04-17
Make metropolis elemwise updates independent of each other

ricardoV94:
Not updating q0 after each elemwise update rendered subsequent proposals dependent on the previous ones.

This issue was revealed when testing the GeneralizedPoisson introduced in https://github.com/pymc-devs/pymc-experimental/pull/143

----
:books: Documentation preview :books:: https://pymc--6678.org.readthedocs.build/en/6678/

### DIFF
--- a/pymc/step_methods/metropolis.py
+++ b/pymc/step_methods/metropolis.py
@@ -257,6 +257,7 @@ class Metropolis(ArrayStepShared):
             q = floatX(q0d + delta)
 
         if self.elemwise_update:
+            q0d = q0d.copy()
             q_temp = q0d.copy()
             # Shuffle order of updates (probably we don't need to do this in every step)
             np.random.shuffle(self.enum_dims)
@@ -264,7 +265,7 @@ class Metropolis(ArrayStepShared):
                 q_temp[i] = q[i]
                 accept_rate_i = self.delta_logp(q_temp, q0d)
                 q_temp_, accepted_i = metrop_select(accept_rate_i, q_temp, q0d)
-                q_temp[i] = q_temp_[i]
+                q_temp[i] = q0d[i] = q_temp_[i]
                 self.accept_rate_iter[i] = accept_rate_i
                 self.accepted_iter[i] = accepted_i
                 self.accepted_sum[i] += accepted_i
--- a/tests/step_methods/test_metropolis.py
+++ b/tests/step_methods/test_metropolis.py
@@ -120,6 +120,16 @@ class TestMetropolis:
         assert az.rhat(trace).max()["batched_dist"].values < 1.1
         assert az.ess(trace).min()["batched_dist"].values > 50
 
+    def test_elemwise_update_different_scales(self):
+        mu = [1, 2, 3, 4, 5, 100, 1_000, 10_000]
+        with pm.Model() as m:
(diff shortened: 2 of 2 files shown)


tests/step_methods/test_metropolis.py
```diff
@@ -120,6 +120,16 @@ def test_elemwise_update(self, batched_dist):
...
 
+    def test_elemwise_update_different_scales(self):
+        mu = [1, 2, 3, 4, 5, 100, 1_000, 10_000]
+        with pm.Model() as m:
+            x = pm.Poisson("x", mu=mu)
+            step = pm.Metropolis([x])
+            trace = pm.sample(draws=1000, chains=2, step=step, random_seed=128).posterior
+
+        np.testing.assert_allclose(trace["x"].mean(("draw", "chain")), mu, rtol=0.1)
+        np.testing.assert_allclose(trace["x"].var(("draw", "chain")), mu, rtol=0.2)
```
ricardoV94, line 131: The relative error was ~2x worse before the changes

bwengals:
Maybe NHST works, I guess this should happen 1/100 times. 

Yes absolutely, I should have done that.

pymc/step_methods/metropolis.py
```diff
@@ -257,14 +257,15 @@ def astep(self, q0: RaveledVars) -> Tuple[RaveledVars, StatsType]:
...
+            q0d = q0d.copy()
             q_temp = q0d.copy()
             # Shuffle order of updates (probably we don't need to do this in every step)
             np.random.shuffle(self.enum_dims)
             for i in self.enum_dims:
                 q_temp[i] = q[i]
                 accept_rate_i = self.delta_logp(q_temp, q0d)
                 q_temp_, accepted_i = metrop_select(accept_rate_i, q_temp, q0d)
-                q_temp[i] = q_temp_[i]
+                q_temp[i] = q0d[i] = q_temp_[i]
```
lucianopaz, line 268: If I understand correctly, this basically means that you are performing Gibbs updates at each batch dimension iteratively. So if you have a vector variable $\vec{v}$ and a proposal vector $\vec{u}$, you will first see if you accept or reject the vector $(u_{0}, v_{1}, ..., v_{n})$ versus $\vec{v}$. If you accept this proposal, the next proposal will be $(u_{0}, u_{1}, v_{2}, ..., v_{n})$ versus $(u_{0}, v_{1}, ..., v_{n})$, and so on. If I understand correctly, the past behavior was doing the iterative element wise updates but always compared against $\vec{v}$.
If I my understanding is correct, then this looks ok to me

lucianopaz (approved):

pymc/step_methods/metropolis.py
ricardoV94: Yes that's exactly it!

Only thing is we don't distinguish between core and batched dimensions, everything is unrolled as one long vector. 

We don't perform `elemwise_update` if there is any multivariate discrete variable in the sampler, as those are usually impossible to update one value at a time (e.g., Multinomial). 

For multivariate continuous, we implicitly assume they have a transform that makes each core dim more or less independent (or it's naturally so without a transform).

bwengals:
Ah ok thanks @ricardoV94, I commented there before seeing this.  I'll try running it locally and see

PR #6679, opened 2023-04-18
Seed HSGP bootstrap tests

ricardoV94:
Failed in #6678

----
:books: Documentation preview :books:: https://pymc--6679.org.readthedocs.build/en/6679/

### DIFF
--- a/tests/gp/test_hsgp_approx.py
+++ b/tests/gp/test_hsgp_approx.py
@@ -51,7 +51,7 @@ def build_mmd_func(sample1, sample2):
     return calc_mmd
 
 
-def two_sample_test(sample1, sample2, n_sims=1000, alpha=0.05):
+def two_sample_test(sample1, sample2, rng: np.random.Generator, n_sims=1000, alpha=0.05):
     """Calculate test whose null hypothesis is that two sets of samples were drawn from
     the same distribution.
 
@@ -67,7 +67,7 @@ def two_sample_test(sample1, sample2, n_sims=1000, alpha=0.05):
 
     h0 = []
     for i in range(n_sims):
-        np.random.shuffle(ix)
+        rng.shuffle(ix)
         X = X[ix, :]
         h0.append(calc_mmd(X[:half_N, :], X[half_N:, :]))
     h0 = np.asarray(h0)
@@ -171,6 +171,9 @@ class TestHSGP:
         prior and compare them using MMD two sample test.  Tests both centered and non-centered
         parameterizations.
         """
+        seeds = np.arange(5) + 174  # 5 possible seeds
+        rng = np.random.default_rng(np.random.choice(seeds))
+
         with model:
             hsgp = pm.gp.HSGP(m=[200], c=2.0, parameterization=parameterization, cov_func=cov_func)
             f1 = hsgp.prior("f1", X=X1)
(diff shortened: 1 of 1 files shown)


bwengals:
weird that a different seed it still fails.  maybe very unlucky...?  I couldnt get it to fail in the original PR.  Does it fail for like, any seed?

ricardoV94:
My last commit tries 5 seeds, it fails in one of them. The failure rate seems higher than 0.01. Even with 5 tests should only fail 5% of the time

bwengals:
yup somethings wrong, will pull and give it a try

tests/gp/test_hsgp_approx.py
```diff
@@ -194,17 +197,20 @@ def test_conditional(self, model, cov_func, X1, parameterization):
         prior and compare them using MMD two sample test.  Tests both centered and non-centered
         parameterizations.  The conditional should match the prior when no data is observed.
         """
+        seeds = np.arange(5) + 197  # 5 possible seeds
+        rng = np.random.default_rng(np.random.choice(seeds))
+
```
bwengals, line 202: It seems cleaner to use `@parameterize` like for `test_prior`?

tests/gp/test_hsgp_approx.py
```diff
@@ -166,25 +166,28 @@ def test_parametrization_drop_first(self, model, cov_func, X1, drop_first):
...
+    @pytest.mark.parametrize("seed", list(np.arange(5) + 174))
+    def test_prior(self, model, cov_func, X1, parameterization, seed):
         """Compare HSGP prior to unapproximated GP prior, pm.gp.Latent.  Draw samples from the
         prior and compare them using MMD two sample test.  Tests both centered and non-centered
         parameterizations.
         """
+        rng = np.random.default_rng(seed)
+
         with model:
             hsgp = pm.gp.HSGP(m=[200], c=2.0, parameterization=parameterization, cov_func=cov_func)
```
bwengals, line 178: try increasing `m` here to 500

tests/gp/test_hsgp_approx.py
```diff
@@ -166,25 +166,28 @@ def test_parametrization_drop_first(self, model, cov_func, X1, drop_first):
...
             f1 = hsgp.prior("f1", X=X1)
 
             gp = pm.gp.Latent(cov_func=cov_func)
             f2 = gp.prior("f2", X=X1)
 
-            idata = pm.sample_prior_predictive(samples=1000)
+            idata = pm.sample_prior_predictive(samples=1000, random_seed=rng)
```
bwengals, line 184: I think its ok to decrease `samples` here to 500, will run faster which will be nice for trying the different random seeds

tests/gp/test_hsgp_approx.py
```diff
@@ -194,17 +197,20 @@ def test_conditional(self, model, cov_func, X1, parameterization):
         prior and compare them using MMD two sample test.  Tests both centered and non-centered
         parameterizations.  The conditional should match the prior when no data is observed.
         """
+        seeds = np.arange(5) + 197  # 5 possible seeds
+        rng = np.random.default_rng(np.random.choice(seeds))
+
         with model:
             hsgp = pm.gp.HSGP(m=[100], c=2.0, parameterization=parameterization, cov_func=cov_func)
```
bwengals, line 204: try increasing `m` here to 500 too.  This got rid of the random failures for me locally

bwengals (changes requested):
Increasing `m` seems to have fixed things for me locally.  Then, decreasing the number of samples in `pm.sample_posterior_predictive` helps it go a bit faster.  I think that should fix it

tests/gp/test_hsgp_approx.py
ricardoV94: The @parametrize was just to show what seed caused the test to fail. The idea was to revert back to choosing one of 5 possible seeds once the root cause was found.

I imagine we don't want to run each test 5 times, but instead allow some variation across tests.

Issue #6633, opened 2023-03-29
Implement logprob inference for binary operations


ricardoV94:
### Description

Using CDFs it should be simple to derive the logp of graphs of the form:

```python
import pymc as pm

x = pm.Normal.dist()
y = x < 0.5
pm.logp(y, value=1).eval()  # same as pm.logcdf(x, 0.5)
pm.logp(y, value=0).eval()  # same as log(1 - exp(pm.logcdf(x, 0.5)))
```

https://github.com/pymc-devs/pymc/blob/main/pymc/logprob/censoring.py includes rewrites of other operations that rely on CDF such as `clip`, and `round`. 

More challenging, but also fun, would be to support `all` and `any`, whose logp should be the sum of the logp that all binary variables evaluate to True or False: `any = 1 - all(x==False)`

For ordering, and min/max see https://github.com/pymc-devs/pymc/issues/6350

shreyas3156:
@ricardoV94 I would like to work on this issue starting with the logical comparison Ops (gt, lt, ge, le), but I needed some clarification about the value of logp for y.

> `pm.logp(y, 0).eval()  # same as pm.logcdf(y, 0.5)`

Since y is a bool type, how exactly is `pm.logcdf` for y defined? And consequently, what should `pm.logcdf(y, 0.5)` evaluate to?

ricardoV94:
@shreyas3156 My code comments where wrong, `pm.logp(y, True) == pm.logp(x < 0.5, True) == pm logcdf(x, 0.5)`.

Does that make sense? Just saying the probability of a simple constant inequality is the same as some CDF expression on the same variable.

shreyas3156:
Yes, that makes perfect sense now. Thanks!

ricardoV94:
Next interesting binary Ops would be `and`, `or`, `all` and `any`

PR #6682, opened 2023-04-20
Make logprob inference for binary ops independent of order of inputs

shreyas3156:
**What is this PR about?**
This PR makes the logprob derivation of binary comparison ops (`<=, >=, <, >`) independent of the order of the inputs. e.g. `pt.le(x, 0.4)` and `pt.le(0.4, x)` would have the same logp. 

Addresses #6633.

**Checklist**

----
:books: Documentation preview :books:: https://pymc--6682.org.readthedocs.build/en/6682/

### DIFF
--- a/pymc/logprob/binary.py
+++ b/pymc/logprob/binary.py
@@ -62,6 +62,7 @@ def find_measurable_comparisons(
     if len(measurable_inputs) != 1:
         return None
 
+    # Make the measurable base_var always be the first input to the MeasurableComparison node
     base_var: TensorVariable = measurable_inputs[0][0]
 
     # Check that the other input is not potentially measurable, in which case this rewrite
@@ -79,6 +80,7 @@ def find_measurable_comparisons(
 
     node_scalar_op = node.op.scalar_op
 
+    # Change the Op if the base_var is the second input in node.inputs. e.g. pt.lt(const, dist) -> pt.gt(dist, const)
     if measurable_inputs[0][1] == 1:
         if isinstance(node_scalar_op, LT):
             node_scalar_op = GT()
--- a/pymc/logprob/binary.py
+++ b/pymc/logprob/binary.py
@@ -20,6 +20,7 @@ from pytensor.graph.basic import Node
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.rewriting.basic import node_rewriter
 from pytensor.scalar.basic import GE, GT, LE, LT
+from pytensor.tensor import TensorVariable
 from pytensor.tensor.math import ge, gt, le, lt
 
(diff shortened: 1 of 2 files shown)


pymc/logprob/binary.py
```diff
@@ -50,26 +51,35 @@ def find_measurable_comparisons(
...
+    if not check_potential_measurability(const, rv_map_feature):
         return None
 
+    const = const[0]
+
     # Make base_var unmeasurable
     unmeasurable_base_var = ignore_logprob(base_var)
 
     compared_op = MeasurableComparison(node.op.scalar_op)
     compared_rv = compared_op.make_node(unmeasurable_base_var, const).default_output()
```
ricardoV94, line 95: Won't you have to change the `scalar_op` used depending on the order? 

If you have `gt(const, dist)` you don't want to return `measurable_gt(dist, const)`, or you changed the meaning. Did I miss something?

shreyas3156: Oh you're absolutely right! I missed it while thinking about the structure of the rewrite. Will make the change right away.

pymc/logprob/binary.py
```diff
@@ -50,26 +51,47 @@ def find_measurable_comparisons(
...
 
+    const = const[0]
+
     # Make base_var unmeasurable
     unmeasurable_base_var = ignore_logprob(base_var)
 
-    compared_op = MeasurableComparison(node.op.scalar_op)
+    node_scalar_op = node.op.scalar_op
+
+    if measurable_inputs[0][1] == 1:
```
ricardoV94, line 84: I think a comment here explaining what's happening would be very useful to increase readability

pymc/logprob/binary.py
```diff
@@ -50,26 +51,47 @@ def find_measurable_comparisons(
...
+        if isinstance(node_scalar_op, LT):
```
ricardoV94, line 85: I think you want to replace `LT -> GE` and not `LT -> GT`, as well as `LE -> GT` instead of `LE -> GE`

ricardoV94 (changes requested):

pymc/logprob/binary.py
shreyas3156: If the original expression is `y = pt.lt(0.5, x)`, then wouldn't this be equivalent to `y = pt.gt(x, 0.5)`? (considering we're changing the order of the inputs in the IR graph)

ricardoV94: Yeah, you're right. My confusion.

ricardoV94 (approved):

shreyas3156:
@ricardoV94 Once this is merged, I can create a PR for the not-operation logprob inference.

PR #6685, opened 2023-04-22
Passed meta-information for MeasurableOps

Dhruvanshu-Joshi:
**What is this PR about?**
This PR aims to solve issue [6360](https://github.com/pymc-devs/pymc/issues/6360) by incorporating RV meta information in intermediate MeasurableVariables. The Measurable ops covered are MeasurableComparison, MeasurableClip, MeasurableRound, MeasurableSpecifyShape, MeasurableCheckAndRaise, MeasurableIfElse, MeasurableScan, MeasurableMakeVector, MeasurableJoin, MeasurableDimShuffle, MeasurableTransforms and DiracDelta.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6685.org.readthedocs.build/en/6685/

### DIFF
--- a/pymc/logprob/abstract.py
+++ b/pymc/logprob/abstract.py
@@ -131,6 +131,11 @@ def _icdf_helper(rv, value, **kwargs):
 class MeasurableVariable(abc.ABC):
     """A variable that can be assigned a measure/log-probability"""
 
+    def __init__(self, ndim_supp, support_axis, d_type):
+        self.ndim_supp = ndim_supp
+        self.support_axis = support_axis
+        self.d_type = d_type
+
 
 MeasurableVariable.register(RandomVariable)
 
@@ -251,3 +256,30 @@ class MeasurableElemwise(Elemwise):
 
 
 MeasurableVariable.register(MeasurableElemwise)
+
+from enum import Enum, auto
+
+
+class MeasureType(Enum):
+    Discrete = auto()
+    Continuous = auto()
+    Mixed = auto()
+
+
+def get_default_measurable_metainfo(base_op: Op) -> Tuple[int, Tuple[int], MeasureType]:
+    if not isinstance(base_op, MeasurableVariable):
+        raise TypeError("base_op must be a RandomVariable or MeasurableVariable")
+
+    ndim_supp = base_op.ndim_supp
+
+    supp_axes = getattr(base_op, "supp_axes", None)
+    if supp_axes is None:
+        supp_axes = tuple(range(-base_op.ndim_supp, 0))
+
+    measure_type = getattr(base_op, "measure_type", None)
+    if measure_type is None:
+        measure_type = (
+            MeasureType.Discrete if base_op.dtype.startswith("int") else MeasureType.Continuous
+        )
+
+    return ndim_supp, supp_axes, measure_type
--- a/pymc/logprob/abstract.py
+++ b/pymc/logprob/abstract.py
@@ -38,7 +38,7 @@ import abc
 
 from copy import copy
 from functools import singledispatch
-from typing import Callable, List, Sequence, Tuple
+from typing import Any, Callable, List, Sequence, Tuple, Union
 
 from pytensor.graph.basic import Apply, Variable
 from pytensor.graph.op import Op
@@ -131,10 +131,10 @@ def _icdf_helper(rv, value, **kwargs):
 class MeasurableVariable(abc.ABC):
     """A variable that can be assigned a measure/log-probability"""
 
-    def __init__(self, ndim_supp, support_axis, d_type):
+    def __init__(self, ndim_supp, supp_axes, measure_type):
         self.ndim_supp = ndim_supp
-        self.support_axis = support_axis
-        self.d_type = d_type
+        self.supp_axes = supp_axes
+        self.measure_type = measure_type
 
 
 MeasurableVariable.register(RandomVariable)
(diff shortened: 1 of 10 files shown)


pymc/logprob/binary.py
```diff
@@ -38,6 +38,12 @@ class MeasurableComparison(MeasurableElemwise):
 
     valid_scalar_types = (GT, LT, GE, LE)
 
+    def __init__(self, scalar, ndim_supp=0, support_axis=None, d_type="mixed"):
+        super().__init__(scalar)
+        self.ndim_supp = ndim_supp
+        self.support_axis = support_axis
+        self.d_type = d_type
```
ricardoV94: Instead of adding these attributes to each subclass, we can add it to the base MeasurableVariable class. Also we shouldn't give defaults, unless we are sure they always apply. Since most rewrites work with both discrete and continuous RVs, that's not the case.

The rewrites that introduce the MeasurableOps should always define them. We can have a helper that works for the default cases where `ndim_supp`, `support_axes` and `measure_type` are the same as the base RV class. 

Also, for `measure_dtype` let's use Enum to avoid typing bugs.

It might look something like:

```python
from enum import Enum, auto

class MeasureType(Enum):
  Discrete = auto()
  Continuous = auto()
  Mixed = auto()


def get_default_measurable_metainfo(base_op: Op) -> Tuple[int, Tuple[int], MeasureType]:
  if not isinstance(base_op, MeasurableVariable):
      raise TypeError("base_op must be a RandomVariable or MeasurableVariable")
  
  ndim_supp = base_op.ndim_supp

  supp_axes = getattr(base_op, "supp_axes", None)
  if supp_axes is None:
    supp_axes = tuple(range(-base_op.ndim_supp, 0))

  measure_type = getattr(base_op, "measure_type", None):
  if measure_type is None:
    measure_type = MeasureType.Discrete if base_op.dtype.startswith("int") else MeasureType.Continuous
  
  return ndim_supp, supp_axes, measure_type
```

ricardoV94 (changes requested):

pymc/logprob/abstract.py
```diff
@@ -131,6 +131,11 @@ def _icdf_helper(rv, value, **kwargs):
 class MeasurableVariable(abc.ABC):
```
ricardoV94, line 131: Oh we are not subclassing from it directly so its not being used (almost all of our logprob rewrites should have failed).

We usually subclass from the Ops... so maybe we should add a class decorator to add these properties to init and pass the remaining ones to the original Op class when it's initialized.

Dhruvanshu-Joshi: So from what I understand, we should add meta-information as parameters to the init using decorators for the pytensor ops which we will be using namely: Elemwise, CumOp, CheckAndRaise, SpecifyShape, IfElse, Scan , Join, MakeVector and Dimshuffle. We can then use the ```get_default_measurable_metainfo``` on the node.op which is an object of the pytensor op in the ```find_measurable_operation``` function to get the default values wherever required. Am I correct?

ricardoV94: The decorator would be applied on the Measurable subclasses of those Ops, e.g.

```python
@measurable_op
class MeasurableElemwise(Elemwise):
  pass
```

So that we can kind of subclass without multiple inheritance. The decorator would override `__init__` to capture the new parameters we need, and call the original `__init__` with the rest.

ricardoV94: But maybe Multiple Inheritance is the best here. We might need to think/try both to figure it out.

EDIT: I am more convinced of this one now :/

Dhruvanshu-Joshi: I tried the decorator method and it works fine. There is one problem that whenever the ```MeasurableOp``` class is called in the ```find_measurable_op``` method, in many cases(almost all), it requires some default values for all the parameters including the metainfo parameters and the original init parameters. The ```get_default_measurable_metainfo``` does not work on ```node.op``` as ```node.op``` is an object of the ```Op``` and not ```MeasurableOp``` where the parameters are added. 
For eg:
https://github.com/pymc-devs/pymc/blob/a617bf29393200b9b998ae354863eda4d0df670c/pymc/logprob/binary.py#L44-L97
Replacing line 93 and 94 as 
```python
ndim_supp, supp_axes, measure_type = get_default_measurable_metainfo(node.op)
compared_op = MeasurableComparison(ndim_supp, supp_axes, measure_type, node_scalar_op)
```
produces an attribute error 
```
AttributeError: 'Elemwise' object has no attribute 'ndim_supp'
```
I guess multiple Inheritance will also lead to this error as the metainfo are passed to our ```MeasurableOps``` and not the original pytensor ```Ops```.

ricardoV94: I think that's exactly why multiple inheritance would work better, as those attributes would now be part of the Op through the MeasurableClass path.

Dhruvanshu-Joshi: I have made attributes part of the Op through the MeasurableClass path by including them in the ```init``` a done according to the first commit which , I believe is the approach that multiple inheritance follows. The attributes are accessible through an object of the MeasurableClass and not the Op class. However, I am still confused on how to proceed with default values. As mentioned earlier, ```node.op``` is derived from the MeasurableClass itself and not the Op. Hence in case of MeasurableElemwise and sub-class which derive from it, ```scalar_op``` being an attribute of the op itself gives no error for ```node.op.scalar_op``` but ```node.op.ndim_supp``` does.

Dhruvanshu-Joshi:
continued in [#6754](https://github.com/pymc-devs/pymc/pull/6754)

PR #6688, opened 2023-04-25
Add Hurdle distributions

tomicapretto:
**What is this PR about?**

Implements four hurdle distributions. These are like zero-inflated distributions, but the process that generates the zeroes is independent of the process that generates the non-zero values (in zero-inflated distributions the zeros can come from the two distinct processes).

These distributions are added in `distributions/mixture.py`. To do so, I had to move a couple of things around

* `DiracDelta` is part of `distributions/distribution.py`
* Zero-inflated distributions are now part of `distributions/mixture.py`
* These new distributions go into `distributions/mixture.py` as well
* Moved tests for `DiracDelta` and zero-inflated distributions to their respective places

There's a `hurdle.ipynb` notebook. My goal is not to merge it into the repository (it's in a very bad place). But I think it can be useful if you want to see how these distributions work. 

My end goal is to add hurdle families to Bambi and I think it would be better to have these distributions implemented and tested in PyMC rather than as something that only lives within Bambi.

**Checklist**

## New features
- Added hurdle distributions (`HurdlePoisson`, `HurdleNegativeBinomial`, `HurdleGamma`, and `HurdleLogNormal`).

## Maintenance
- Move `DiracDelta` to `distributions/distribution.py`
- Move zero-inflated families to `distributions/mixture.py`
- Remove `pm.Constant`

**Edit**: @ricardoV94


pymc/distributions/distribution.py
```diff
@@ -1116,3 +1119,81 @@ def default_moment(rv, size, *rv_inputs, rv_name=None, has_fallback=False, ndim_
...
+        )
+        return DiracDelta(*args, **kwargs)
+
+    @classmethod
+    def dist(cls, *args, **kwargs):
+        warnings.warn(
+            "pm.Constant has been deprecated. Use pm.DiracDelta instead.",
+            FutureWarning,
+        )
+        return DiracDelta.dist(*args, **kwargs)
```
ricardoV94: Good time as any to remove it

ricardoV94 (commented):
Looks good functionality-wise. 

In terms of organization, since there are two things going on, can I ask you to clean the git history and separate the `move stuff around` and `implement hurdle models` into two distinct commits?

pymc/distributions/distribution.py
tomicapretto: Done :)

ricardoV94 (changes requested):
We should remove the notebooks (right?)

There seem to be some conflict lines `<<<<` in the intermediate commits. It probably makes things easier for you to 1) remove Constant, 2) Move things around, 3) Implement the new stuff, in that order.

Let me know if you want me to do it instead. 

Everything else :ok_hand:

tomicapretto:
> There seem to be some conflict lines `<<<<` in the intermediate commits

I thought they were completely removed. I expected them to be caught by the linting tools. It seems I was wrong. 

> Let me know if you want me to do it instead.

I would appreciate it a lot :D

ricardoV94:
And its correct to remove the notebooks?

tomicapretto:
> And its correct to remove the notebooks?

Yes, my bad. It was something I created to see how it worked. But it's not needed.

ricardoV94 (approved):

PR #6680, opened 2023-04-19
Derive logprob of `>=` and `<=` operations

shreyas3156:
**What is this PR about?**
This PR implements the logprob inference for binary comparison Ops `>=` and `<=` in continuation to #6662.
The derivation for discrete distributions has been implemented separately from continuous to club together the similarities in logprob of `<=` and `>` ops.

Addresses #6633. 

**Checklist**
+ [x] Explain important implementation details 👆
+ [x] Make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).
+ [x] Link relevant issues (preferably in [nice commit messages](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html))
+ [x] Are the changes covered by tests and docstrings?
+ [x] Fill out the short summary sections 👇

## New features
- Logprob derivation for GE and LE ops.

----
:books: Documentation preview :books:: https://pymc--6680.org.readthedocs.build/en/6680/

PR #6689, opened 2023-04-26
Add logprob inference for `not` operations

shreyas3156:
**What is this PR about?**
Implementation of logprob derivation for bitwise 'not' operations as discussed in #6680, using the following logic:
`not(gt(x, const), True) -> (gt(x, const), not(True))`.

**Checklist**

## New features
- Logprob inference for bitwise 'not' Ops.

----
:books: Documentation preview :books:: https://pymc--6689.org.readthedocs.build/en/6689/

### DIFF
--- a/pymc/logprob/binary.py
+++ b/pymc/logprob/binary.py
@@ -161,6 +161,9 @@ def find_measurable_bitwise(fgraph: FunctionGraph, node: Node) -> Optional[List[
     ):
         return None
 
+    if not base_var.dtype.startswith("bool"):
+        raise None
+
     # Make base_var unmeasurable
     unmeasurable_base_var = ignore_logprob(base_var)
 
@@ -184,6 +187,6 @@ measurable_ir_rewrites_db.register(
 def bitwise_not_logprob(op, values, base_rv, **kwargs):
     (value,) = values
 
-    logprob = _logprob_helper(base_rv, np.bitwise_not(value), **kwargs)
+    logprob = _logprob_helper(base_rv, invert(value), **kwargs)
 
     return logprob
--- a/pymc/logprob/binary.py
+++ b/pymc/logprob/binary.py
@@ -19,9 +19,9 @@ import pytensor.tensor as pt
 from pytensor.graph.basic import Node
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.rewriting.basic import node_rewriter
-from pytensor.scalar.basic import GE, GT, LE, LT
+from pytensor.scalar.basic import GE, GT, LE, LT, Invert
 from pytensor.tensor import TensorVariable
-from pytensor.tensor.math import ge, gt, le, lt
+from pytensor.tensor.math import ge, gt, invert, le, lt
 
 from pymc.logprob.abstract import (
     MeasurableElemwise,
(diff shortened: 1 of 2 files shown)


pymc/logprob/binary.py
```diff
@@ -136,3 +136,54 @@ def comparison_logprob(op, values, base_rv, operand, **kwargs):
...
+    "basic",
+    "bitwise",
+)
+
+
+@_logprob.register(MeasurableBitwise)
+def bitwise_not_logprob(op, values, base_rv, **kwargs):
+    (value,) = values
+
+    logprob = _logprob_helper(base_rv, np.bitwise_not(value), **kwargs)
```
ricardoV94: Better to use pytensor operators directly.

```suggestion
    logprob = _logprob_helper(base_rv, invert(value), **kwargs)
```

pymc/logprob/binary.py
```diff
@@ -136,3 +136,54 @@ def comparison_logprob(op, values, base_rv, operand, **kwargs):
...
+    base_var = node.inputs[0]
+    if not (
+        base_var.owner
+        and isinstance(base_var.owner.op, MeasurableVariable)
+        and base_var not in rv_map_feature.rv_values
+    ):
+        return None
+
+    # Make base_var unmeasurable
+    unmeasurable_base_var = ignore_logprob(base_var)
```
ricardoV94, line 168: We should make sure the base_var is of boolean dtype. Because `np.bitwise_not(1)` (as well as pytensor invert) does something funny :)

ricardoV94 (changes requested):
Looks good, some small tweak suggestion below

pymc/logprob/binary.py
```diff
@@ -136,3 +136,55 @@ def comparison_logprob(op, values, base_rv, operand, **kwargs):
...
+
+    if isinstance(node.op, MeasurableBitwise):
+        return None  # pragma: no cover
+
+    base_var = node.inputs[0]
+    if not (
+        base_var.owner
+        and isinstance(base_var.owner.op, MeasurableVariable)
+        and base_var not in rv_map_feature.rv_values
+        and base_var.dtype.startswith("bool")
```
ricardoV94: Nitpick: What about separating this check? 

The other checks are basically done by every rewrite and this one is specific to the invert. When parsing what this rewrite is doing it would help to see it in a separate if block.

shreyas3156: Definitely, that makes more sense.

pymc/logprob/binary.py
```diff
@@ -136,3 +136,57 @@ def comparison_logprob(op, values, base_rv, operand, **kwargs):
...
+    base_var = node.inputs[0]
+    if not (
+        base_var.owner
+        and isinstance(base_var.owner.op, MeasurableVariable)
+        and base_var not in rv_map_feature.rv_values
+    ):
+        return None
+
+    if not base_var.dtype.startswith("bool"):
+        raise TypeError("Base RV should be of type boolean")
```
ricardoV94:
```suggestion
        raise TypeError("Base RV should be of type boolean")
```
We shouldn't raise errors, because another rewrite (including a user custom one) could implement a transformation for this type of variable.

One thing we might want to start doing sooner or later is to add `log.debug` statements so that users can figure out why didn't pymc figure out the logprob for their expression (where here we could say that `bitwise` for integer variables is not supported by this rewrite).

ricardoV94: Opened an issue https://github.com/pymc-devs/pymc/issues/6691

Here we should just return `None`, which means the current rewrite does not apply

shreyas3156: > We shouldn't raise errors, because another rewrite (including a user custom one) could implement a transformation for this type of variable.

Understood. I'll replace it to return None.

> One thing we might want to start doing sooner or later is to add log.debug statements so that users can figure out why didn't pymc figure out the logprob for their expression (where here we could say that bitwise for integer variables is not supported by this rewrite).

I have a question about this. Since the logprob rewrites are in the IR graph, should the user be exposed to the internal errors? (e.g.: If the value variable is potentially measurable).

ricardoV94: > Since the logprob rewrites are in the IR graph, should the user be exposed to the internal errors? (e.g.: If the value variable is potentially measurable).

What do you mean by internal errors?

ricardoV94 (approved):

pymc/logprob/binary.py
shreyas3156: I was thinking of a scenario when we check the potential measurability of the value variables and whether the user would be concerned with the error. But I figured that it has to be the expression defined by the user that would cause the error, so the answer should be yes. It may have been an oversight on my part since I'm still getting acquainted with the use cases of PyMC.

ricardoV94: The point I was trying to make is that these rewrites are local. So what looks like something invalid to them, may be valid for another rewrite (ours or provided by the user). So we shouldn't raise any errors ourselves, but simply return `None`, which states: this rewrite does not apply to this graph.

What I suggested in the linking issue, is that we can issue a log message saying why this local rewrite was not applied, as it can be very useful for the user to understand why the logprob was not derived for their graph.

shreyas3156: Okay, I understand now. Thank you so much for clarifying!

PR #6696, opened 2023-04-28
Support Scans in `CustomDist`

ricardoV94:
Here is how one can now write an AR with StudentT innovations:

```python
import numpy as np
import pymc as pm
import pytensor.tensor as pt
from pytensor.scan import scan

def ar_t(nu, sigma, rho, size):
    
    def step(xtm1, nu, sigma, rho):
        x = xtm1 * rho + pm.StudentT.dist(nu=nu, sigma=sigma)
        return x, collect_default_updates([x])
    
    xs, _ = scan(
        fn=step,
        outputs_info=[pt.zeros(())],
        non_sequences=[nu, sigma, rho],
        n_steps=size[0],
    )
    
    return xs

with pm.Model() as m:
    nu = 4
    rho = pm.Uniform("rho")
    steps=100
    
    pm.CustomDist("ar_t", nu, sigma, rho, dist=ar_t, observed=np.random.randn(steps))
    
    posterior = pm.sample()    
```

The part that was not supported was that we would not be able to collect the update for the scan (used in prior/posterior predictive), and the user had no way of specifying it manually either.

----
:books: Documentation preview :books:: https://pymc--6696.org.readthedocs.build/en/6696/

### DIFF
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -47,6 +47,7 @@ from pytensor.graph.basic import (
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.op import Op
 from pytensor.scalar.basic import Cast
+from pytensor.scan.op import Scan
 from pytensor.tensor.basic import _as_tensor_variable
 from pytensor.tensor.elemwise import Elemwise
 from pytensor.tensor.random.op import RandomVariable
@@ -1004,16 +1005,49 @@ def reseed_rngs(
 
 
 def collect_default_updates(
-    inputs: Sequence[Variable],
     outputs: Sequence[Variable],
+    *,
+    inputs: Optional[Sequence[Variable]] = None,
     must_be_shared: bool = True,
 ) -> Dict[Variable, Variable]:
     """Collect default update expression for shared-variable RNGs used by RVs between inputs and outputs.
 
-    If `must_be_shared` is False, update expressions will also be returned for non-shared input RNGs.
-    This can be useful to obtain the symbolic update expressions from inner graphs.
-    """
+    Parameters
+    ----------
+    outputs: list of PyTensor variables
+        List of variables in which graphs default updates will be collected.
+    inputs: list of PyTensor variables, optional
+        Input nodes above which default updates should not be collected.
+        When not provided, search will include top level inputs (roots).
+    must_be_shared: bool, default True
+        Used internally by PyMC. Whether updates should be collected for non-shared
+        RNG input variables. This is used to collect update expressions for inner graphs.
 
+    Examples
+    --------
+    .. code:: python
+        import pymc as pm
+        from pytensor.scan import scan
+        from pymc.pytensorf import collect_default_updates
+
+        def scan_step(xtm1):
+            x = xtm1 + pm.Normal.dist()
+            x_update = collect_default_updates([x])
+            return x, x_update
+
+        x0 = pm.Normal.dist()
+
+        xs, updates = scan(
+            fn=scan_step,
+            outputs_info=[x0],
+            n_steps=10,
+        )
+
+        # PyMC makes use of the updates to seed xs properly.
+        # Without updates, it would raise an error.
+        xs_draws = pm.draw(xs, draws=10)
+
+    """
     # Avoid circular import
     from pymc.distributions.distribution import SymbolicRandomVariable
 
@@ -1048,15 +1082,30 @@ def collect_default_updates(
             next_rng = client.op.update(client).get(rng)
             if next_rng is None:
                 raise ValueError(
-                    f"No update mapping found for RNG used in SymbolicRandomVariable Op {client.op}"
+                    f"No update found for at least one RNG used in SymbolicRandomVariable Op {client.op}"
+                )
+        elif isinstance(client.op, Scan):
+            # Check if any shared output corresponds to the RNG
+            rng_idx = client.inputs.index(rng)
+            io_map = client.op.get_oinp_iinp_iout_oout_mappings()["outer_out_from_outer_inp"]
+            out_idx = io_map.get(rng_idx, -1)
+            if out_idx != -1:
+                next_rng = client.outputs[out_idx]
+            else:  # No break
+                raise ValueError(
+                    f"No update found for at least one RNG used in Scan Op {client.op}.\n"
+                    "You can use `pytensorf.collect_default_updates` inside the Scan function to return updates automatically."
                 )
         else:
-            # We don't know how this RNG should be updated (e.g., Scan).
+            # We don't know how this RNG should be updated (e.g., OpFromGraph).
             # The user should provide an update manually
             return None
 
         # Recurse until we find final update for RNG
         return find_default_update(clients, next_rng)
+
+    if inputs is None:
+        inputs = []
 
     outputs = makeiter(outputs)
     fg = FunctionGraph(outputs=outputs, clone=False)
@@ -1129,7 +1178,7 @@ def compile_pymc(
     """
     # Create an update mapping of RandomVariable's RNG so that it is automatically
     # updated after every function call
-    rng_updates = collect_default_updates(inputs, outputs)
+    rng_updates = collect_default_updates(inputs=inputs, outputs=outputs)
 
     # We always reseed random variables as this provides RNGs with no chances of collision
     if rng_updates:
--- a/tests/distributions/test_distribution.py
+++ b/tests/distributions/test_distribution.py
@@ -23,6 +23,7 @@ import pytensor.tensor as pt
 import pytest
 import scipy.stats as st
 
+from pytensor import scan
 from pytensor.tensor import TensorVariable
 
 import pymc as pm
@@ -51,6 +52,7 @@ from pymc.exceptions import BlockModelAccessError
 from pymc.logprob.abstract import get_measurable_outputs
 from pymc.logprob.basic import logcdf, logp
 from pymc.model import Deterministic, Model
+from pymc.pytensorf import collect_default_updates
 from pymc.sampling import draw, sample
 from pymc.testing import (
     BaseTestDistributionRandom,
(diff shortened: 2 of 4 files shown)


tests/distributions/test_distribution.py
```diff
@@ -516,6 +518,48 @@ def old_random(size):
...
+        nu = 4
+        sigma = 0.7
+        steps = 99
+        batch_size = 3
+        x = CustomDist.dist(nu, sigma, steps, dist=trw, size=batch_size)
+
+        x_draw = pm.draw(x, random_seed=1)
+        assert x_draw.shape == (steps, batch_size)
+        np.testing.assert_allclose(pm.draw(x, random_seed=1), x_draw)
+        assert not np.any(pm.draw(x, random_seed=2) == x_draw)
```
michaelosthege, line 555:
```suggestion
        # Check that seeding works as expected
        x_draw = pm.draw(x, random_seed=1)
        assert x_draw.shape == (steps, batch_size)
        np.testing.assert_allclose(pm.draw(x, random_seed=1), x_draw)
        assert not np.any(pm.draw(x, random_seed=2) == x_draw)
```

tests/distributions/test_distribution.py
```diff
@@ -516,6 +518,48 @@ def old_random(size):
...
+            innovation_dist=pm.StudentT.dist(nu=nu, sigma=sigma),
+            steps=steps,
+            size=(batch_size,),
+        )
+        ref_val = pt.concatenate([np.zeros((1, batch_size)), x_draw]).T
+
+        np.testing.assert_allclose(
+            pm.logp(x, x_draw).eval().sum(0),
+            pm.logp(ref_dist, ref_val).eval(),
+        )
```
michaelosthege, line 568: What are you doing here?
Please add a comment what this `ref_val` concatenation is good for.

With the `assert_allclose` you're testing that the logp derived via both distributions is the same? At least that's my high level understanding, why do you need zeroes for one, and `sum(0)` the first but not the second?

tests/test_pytensorf.py
```diff
@@ -538,19 +539,56 @@ def test_nested_updates(self):
...
+            fn=step_with_update,
+            outputs_info=[pt.zeros(())],
+            non_sequences=[rng],
+            n_steps=10,
+        )
+
+        assert collect_default_updates([ys]) == {rng: tuple(next_rng.values())[0]}
+
+        fn = compile_pymc([], ys, random_seed=1)
+        assert not (set(fn()) & set(fn()))
```
michaelosthege, line 592:
```suggestion
        # Confirm thet step functions using RNGs result in scan
        # with updates for that RNG.
        ys, next_rng = scan(
            fn=step_with_update,
            outputs_info=[pt.zeros(())],
            non_sequences=[rng],
            n_steps=10,
        )

        assert collect_default_updates([ys]) == {rng: tuple(next_rng.values())[0]}

        # This should result in no overlaps between random drwas
        fn = compile_pymc([], ys, random_seed=1)
        assert not (set(fn()) & set(fn()))
```

michaelosthege (commented):
When reviewing I slowed down at `find_default_update` inside `collect_default_update`. This entire local function can be extracted which would make it a lot easier to review & test.

Overall it looks good, but I'd recommend to add some comments to make it comprehensible without knowing the context of this PR.

ricardoV94:
> This entire local function can be extracted which would make it a lot easier to review & test.

It's the inner recursive function, I don't think it makes sense to be outside of the initial caller function.

tests/distributions/test_distribution.py
```diff
@@ -516,6 +518,48 @@ def old_random(size):
...
+        steps = 99
+        batch_size = 3
+        x = CustomDist.dist(nu, sigma, steps, dist=trw, size=batch_size)
+
+        x_draw = pm.draw(x, random_seed=1)
+        assert x_draw.shape == (steps, batch_size)
+        np.testing.assert_allclose(pm.draw(x, random_seed=1), x_draw)
+        assert not np.any(pm.draw(x, random_seed=2) == x_draw)
+
+        ref_dist = pm.RandomWalk.dist(
```
ricardoV94, line 557:
```suggestion
        # We compare it with an equivalent RandomWalk
        # We have to add one entry for the initial state which is part of the RandomWalk. 
        # We set `init_dist` to Flat so it doesn't change the logp
        ref_dist = pm.RandomWalk.dist(
```

tests/distributions/test_distribution.py
```diff
@@ -516,6 +518,48 @@ def old_random(size):
...
+            init_dist=pm.Flat.dist(),
+            innovation_dist=pm.StudentT.dist(nu=nu, sigma=sigma),
+            steps=steps,
+            size=(batch_size,),
+        )
+        ref_val = pt.concatenate([np.zeros((1, batch_size)), x_draw]).T
+
+        np.testing.assert_allclose(
```
ricardoV94, line 565:
```suggestion
        # The RandomWalk automatically sums the logp across the time dimension
        np.testing.assert_allclose(
```

tests/distributions/test_distribution.py
ricardoV94: Added some comments answering those questions

tests/test_pytensorf.py
```diff
@@ -538,19 +539,59 @@ def test_nested_updates(self):
...
+            name="test_scan",
+        )
+
+        assert not next_rng
+
+        with pytest.raises(
+            ValueError,
+            match=r"No update mapping found for RNG used in Scan Op for\{cpu,test_scan\}",
+        ):
+            collect_default_updates([xs])
```
jessegrabowski, line 580: This case seems a bit artificial, given that (to my knowledge) `collect_default_updates` is always called by a user inside the inner function. Is this error possible to trigger in that context, or is there any reason a user would want to structure code this way?

pymc/pytensorf.py
```diff
@@ -1000,16 +1001,16 @@ def reseed_rngs(
 
 
 def collect_default_updates(
-    inputs: Sequence[Variable],
     outputs: Sequence[Variable],
+    *,
+    inputs: Optional[Sequence[Variable]] = None,
```
jessegrabowski, line 1010: I guess inputs are optional precisely for the scan case? Is there any advantage/disadvantage to providing or not providing them (maybe it shortens the worst-case number of recursions on this function?)

pymc/pytensorf.py
```diff
@@ -1046,14 +1047,26 @@ def find_default_update(clients, rng: Variable) -> Union[None, Variable]:
                 raise ValueError(
                     f"No update mapping found for RNG used in SymbolicRandomVariable Op {client.op}"
                 )
+        elif isinstance(client.op, Scan):
```
jessegrabowski, line 1087: Futureproofing question: does the new scalar scan from pymc-devs/pytensor#283 subclass from `Scan` (so it will be caught here?)

jessegrabowski (commented):
I think I understand what's going on here. The key change is just to add a check for `Scan` `Op`s and, if so, reach into the `outer_out_from_outer_inp` mapping and grab the `rng` object, raising an error if it's not found. 

I also found the function inside recursive function a bit unusual. [This SO thread](https://stackoverflow.com/questions/14055843/what-is-the-performance-overhead-of-nested-functions) claims there is some overhead associated with doing things this way, so I'm wondering if that would matter in a worst-case recursion (very big graph? multiple scans?). 

Otherwise looks good to me. I would second a bit of documentation in the functions -- a general complaint I have about the Theano codebase is a lack of useful docstrings.

pymc/pytensorf.py
ricardoV94: No it doesn't, but ScalarScan does now allows RNGs in the inner graph

pymc/pytensorf.py
ricardoV94: It has no performance reasons. It is just for user convenience so they don't have to pass inputs. You either pass all inputs or None and often users don't even realize what are all the root inputs of their variables

tests/test_pytensorf.py
ricardoV94: This function is not only called by users, it's also always called by `compile_pymc` to automatically collect updates for the compiled function so the user doesn't have to.

What seems artificial?

ricardoV94:
> I think I understand what's going on here. The key change is just to add a check for `Scan` `Op`s and, if so, reach into the `outer_out_from_outer_inp` mapping and grab the `rng` object, raising an error if it's not found. 

Yup 


> I also found the function inside recursive function a bit unusual. [This SO thread](https://stackoverflow.com/questions/14055843/what-is-the-performance-overhead-of-nested-functions) claims there is some overhead associated with doing things this way, so I'm wondering if that would matter in a worst-case recursion (very big graph? multiple scans?). 


I think that's only meaningful when you are calling the outer function many times, in this case the outer function is called once and then the inner function is called multiple times. Also this is not performance critical code at all since it's done once before compiling a pytensor function. The compilation is orders of magnitude more expensive than this one time graph recursion. And this is itself order of magnitude slower than the presumed heavy use of the final compiled function.

> Otherwise looks good to me. I would second a bit of documentation in the functions -- a general complaint I have about the Theano codebase is a lack of useful docstrings.

Sure

tests/test_pytensorf.py
jessegrabowski: That we're first running the whole scan, then calling `collect_default_updates` on the result, rather than writing `return x, collect_default_updates(x)` inside the inner function. 

We still have to include the `collect_default_updates` as the 2nd return of a scan's inner function after this PR, no? (seems so, looking at the `trw` test). So I didn't understand how we could get to `compile_pymc` calling `collect_default_updates` on the outer output (as in this test case) without it having already been called on all the inner outputs. So I assumed this error would ultimately have to be triggered on inner outputs in a "normal" case.

EDIT: I think I'm an idiot, because this tests *EXACTLY* the case where a user doesn't include `collect` as a return from the inner function. If I'm right about that, maybe something about this in the error to make it more clear what is going wrong? Like, "Please verify the inner function provided to Scan returns an update mapping for RNG"?

pymc/pytensorf.py
jessegrabowski: Plus it's a lot less verbose!

tests/test_pytensorf.py
ricardoV94: Ah I see what you were thinking. Yes I can tweak the error message to be a bit more clear

ricardoV94: Done

ricardoV94:
I'm weary of expanding the docstrings of `collect_default_updates` too much because the concept of updates is itself hard to explain and I rather not write a small blogpost in the docstrings of the function.

Suggestions welcome

ricardoV94:
Updated docstrings and added one code example

michaelosthege (approved):

Issue #6660, opened 2023-04-08
Upload to codecov only once all_tests are run


ricardoV94:
### Description

Following #6654 we can now wait until all tests have run before uploading the coverage report to codecov.

This will avoid the premature bot message saying coverage has dropped that happens all the time as the initial partial coverage files are uploaded by each individual job.

PyTensor can probably be used as a template: https://github.com/pymc-devs/pytensor/blob/98de246293855def134b86b5af6599e74a44835b/.github/workflows/test.yml#L235-L262

ricardoV94:
Hopefully this will also fix this strangr behavior noted by @bwengals https://github.com/pymc-devs/pymc/pull/6801#pullrequestreview-1525338260

PR #6705, opened 2023-05-04
codecov coverage uploaded after tests succeed

michaelraczycki:
**What is this PR about?**
Adaptations of tests.yaml requested in #6660 

**Implementation details:
Removed partial codecov upload that was triggered after each subtest, instead added one that is made only In case of all tests passing

----
:books: Documentation preview :books:: https://pymc--6705.org.readthedocs.build/en/6705/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -493,10 +493,10 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: Install coverage
-        run: conda install coverage -y
       - name: Merge coverage reports
         run: |
+          # Activate conda environment
+          conda activate pymc-test
           # Find all coverage reports and merge them into a single file
           python -m coverage combine --append coverage-*.xml
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -173,7 +173,12 @@ jobs:
       - name: Run tests
         run: |
           conda activate pymc-test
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -488,3 +485,54 @@ jobs:
...
+          # Activate conda environment
+          conda activate pymc-test
+          # Find all coverage reports and merge them into a single file
+          python -m coverage combine --append coverage-*.xml
+
+      - name: Save combined coverage report
+        uses: actions/upload-artifact@v2
+        with:
+          name: coverage
+          path: coverage
```
ricardoV94: Why do we need this extra job? PyTensor doesn't seem to have an equivalent one.

michaelraczycki: pytensor has just one job ('test_ubuntu') that generates the codecov report. in pymc there are many more, that generate partial reports. Adding step that combines them into one final report after all report-generating jobs finish seemed like an easier task

ricardoV94: Okay, but it can still be done in a single job no? Just merge the two steps

michaelraczycki: by two steps do you mean merge-reports and upload coverage?

ricardoV94: Yes, this way there's a bit less clutter in the PR checks

michaelraczycki: sure, should be possible

PR #6709, opened 2023-05-08
Type overloading for `return_inferencedata` in `sample`

thomasaarholt:
**What is this PR about?**
This PR adds typing support which lets a type checker (like mypy or vscode's pylance) infer that when when passing `pm.sample(return_inferencedata=True)`, then the return type is `InferenceData`, and when `return_inferencedata=False`, it is `MultiTrace`. This helps later on when referring to methods or properties that only exist on one of the two types.

**Checklist**

## Maintenance
Added typing support to let a type checker (like mypy or vscode's pylance) infer that when passing `pm.sample(return_inferencedata=True)`, then the return type is `InferenceData`, and when `return_inferencedata=False`, it is `MultiTrace`.

----
:books: Documentation preview :books:: https://pymc--6709.org.readthedocs.build/en/6709/


thomasaarholt:
Mypy is failing with
```[pymc/sampling/mcmc.py]
pymc/sampling/mcmc.py:333: error: Overloaded function signatures 1 and 2 overlap with incompatible return types
```

Pyright (`pyright pymc/sampling/mcmc.py`) does not raise an error, and I'm inclined to believe that it is more accurate than mypy for this. If pymc-devs disagree, I'm happy to take any advice.

michaelosthege (approved):
I never knew how to do this, so thanks for showing!

PR #6711, opened 2023-05-08
Speedup Slice sampler

ricardoV94:
Inherits from StepArrayShared to avoid the many mappings between raveled and non-raveled arrays, which greatly slowdown the sampler. This simple model samples 3 times as fast in my machine after the changes:

```python
import arviz as az
import pymc as pm

with pm.Model() as m:
    x = pm.Normal("x", shape=(200,))
    idata = pm.sample(step=pm.Slice([x]), random_seed=1)
    
az.summary(idata)
```

----
:books: Documentation preview :books:: https://pymc--6711.org.readthedocs.build/en/6711/

### DIFF
--- a/pymc/step_methods/slicer.py
+++ b/pymc/step_methods/slicer.py
@@ -21,7 +21,8 @@ import numpy.random as nr
 
 from pymc.blocking import RaveledVars, StatsType
 from pymc.model import modelcontext
-from pymc.step_methods.arraystep import ArrayStep
+from pymc.pytensorf import compile_pymc, join_nonshared_inputs, make_shared_replacements
+from pymc.step_methods.arraystep import ArrayStepShared
 from pymc.step_methods.compound import Competence
 from pymc.util import get_value_vars_from_user_vars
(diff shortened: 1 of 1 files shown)


pymc/step_methods/slicer.py
```diff
@@ -51,61 +52,66 @@ class Slice(ArrayStep):
...
+        "tune": (bool, []),
         "nstep_out": (int, []),
         "nstep_in": (int, []),
     }
 
     def __init__(self, vars=None, w=1.0, tune=True, model=None, iter_limit=np.inf, **kwargs):
-        self.model = modelcontext(model)
-        self.w = w
+        model = modelcontext(model)
+        self.w = np.asarray(w).copy()
```
fonnesbeck, line 62: Does `asarray` not make a copy, or is that `array`?

fonnesbeck (approved):

pymc/step_methods/slicer.py
ricardoV94: asarray returns the same object if it's already a numpy array

Issue #6717, opened 2023-05-15
BUG: MvNormal logp recomputes Cholesky factorization


dehorsley:
### Describe the issue:

When given a Cholesky factor, L, an observed `MvNormal`, eg

```python
pm.MvNormal('y', mu=np.zeros(2), chol=chol, observed=np.zeros(2))
```

the logp function unnecessarily computes the matrix product of L.LT, then recomputes the Cholesky factor. 

This is expensive for large matrices. 

This originates in the changes for pymc 4, where both Cholesky and precision matrix parameterizations were modified to transform into a covariance matrix parameterization. I'm guessing there are some performance improvements to be had with the precision matrix as well.

### Reproduceable code example:

```python
import numpy as np
import pymc as pm
import pytensor

with pm.Model() as m:
   chol, corr, sigmas = pm.LKJCholeskyCov('cov', n=2, eta=1, sd_dist=pm.HalfNormal.dist())
   pm.MvNormal('y', mu=np.zeros(2), chol=chol, observed=np.zeros(2))

pytensor.dprint(m.logp())
```

### Error message:

Region of logp dprint interest highlighted here, full logp below. The function computes L.LT then recomputes cholesky(L.LT).

```
       | |     |   |   | |     | |Cholesky{lower=True, destructive=False, on_error='nan'} [id GL]
       | |     |   |   | |     |   |dot [id GM]
       | |     |   |   | |     |     |AdvancedIncSubtensor{inplace=False,  set_instead_of_inc=True} [id GN]
       | |     |   |   | |     |     | |Alloc [id GO]
       | |     |   |   | |     |     | | |TensorConstant{0.0} [id GP]
       | |     |   |   | |     |     | | |TensorConstant{2} [id GQ]
       | |     |   |   | |     |     | | |TensorConstant{2} [id GQ]
       | |     |   |   | |     |     | |TransformedVariable [id CB] 'cov_cholesky-cov-packed___cholesky-cov-packed'
       | |     |   |   | |     |     | |TensorConstant{[0 1 1]} [id GR]
       | |     |   |   | |     |     | |TensorConstant{[0 0 1]} [id GS]
       | |     |   |   | |     |     |InplaceDimShuffle{1,0} [id GT]
       | |     |   |   | |     |       |AdvancedIncSubtensor{inplace=False,

```

<details>

```
Sum{acc_dtype=float64} [id A] '__logp'
 |MakeVector{dtype='float64'} [id B]
[...]

PR #6736, opened 2023-05-25
Rewrite logp graph before taking the gradient

dehorsley:
With pymc-devs/pytensor#303 merged, `cholesky(L.dot(L.T))` will be rewritten to `L` if `L.tag.lower_triangular` is set. This change adds these where appropriate. This is important for #6717, however more work is likely required to improve the gradient in such cases.

Some rough benchmarks of computing logp and its grad on the initial point of following model.

<details>

```
n=1000
with pm.Model() as m:
    chol, corr, sigmas = pm.LKJCholeskyCov('cov', n=n, eta=1, sd_dist=pm.HalfNormal.dist())
    pm.MvNormal('y', mu=np.zeros(n), chol=chol, observed=np.ones((1000, n)))

```

</details>

|  | C backend |  JAX backend |
|-|---|-----|
|before | 292 ms ± 28.6 ms | 285 ms ± 28.9 ms  | 
|after | 260 ms ± 19.9 ms | 107 ms ± 969 µs |

The major difference in the "after" between C and JAX backends is that JAX is computing the grad after the rewrite is applied. As mentioned in #6717, this is probably a good motivator for performing some kind of rewrite before computing the gradient.

## Maintenance
- improve performance of MvNormal logp with Cholesky factor.

----
:books: Documentation preview :books:: https://pymc--6736.org.readthedocs.build/en/6736/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -132,6 +132,9 @@ def quaddist_matrix(cov=None, chol=None, tau=None, lower=True, *args, **kwargs):
         chol = pt.as_tensor_variable(chol)
         if chol.ndim != 2:
             raise ValueError("chol must be two dimensional.")
+
+        # tag as lower triangular to enable pytensor rewrites of chol(l.l') -> l
+        chol.tag.lower_triangular = True
         cov = chol.dot(chol.T)
 
     return cov
--- a/pymc/math.py
+++ b/pymc/math.py
@@ -443,11 +443,17 @@ def expand_packed_triangular(n, packed, lower=True, diagonal_only=False):
     elif lower:
         out = pt.zeros((n, n), dtype=pytensor.config.floatX)
         idxs = np.tril_indices(n)
-        return pt.set_subtensor(out[idxs], packed)
+        # tag as lower triangular to enable pytensor rewrites
+        out = pt.set_subtensor(out[idxs], packed)
+        out.tag.lower_triangular = True
+        return out
     elif not lower:
         out = pt.zeros((n, n), dtype=pytensor.config.floatX)
         idxs = np.triu_indices(n)
-        return pt.set_subtensor(out[idxs], packed)
+        # tag as upper triangular to enable pytensor rewrites
+        out = pt.set_subtensor(out[idxs], packed)
+        out.tag.upper_triangular = True
+        return out
 
 
 class BatchedDiag(Op):
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -75,6 +75,7 @@ from pymc.pytensorf import (
     hessian,
     inputvars,
     replace_rvs_by_values,
+    rewrite_pregrad,
 )
 from pymc.util import (
     UNSET,
@@ -380,6 +381,8 @@ class ValueGradFunction:
             )
             self._extra_vars_shared[var.name] = shared
             givens.append((var, shared))
+
+        cost = rewrite_pregrad(cost)
 
         if compute_grads:
             grads = pytensor.grad(cost, grad_vars, disconnected_inputs="ignore")
@@ -824,6 +827,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
                     )
 
         cost = self.logp(jacobian=jacobian)
+        cost = rewrite_pregrad(cost)
         return gradient(cost, value_vars)
 
     def d2logp(
@@ -862,6 +866,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
                     )
 
         cost = self.logp(jacobian=jacobian)
+        cost = rewrite_pregrad(cost)
         return hessian(cost, value_vars)
 
     @property
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -1228,3 +1228,10 @@ def constant_fold(
     return tuple(
         folded_x.data if isinstance(folded_x, Constant) else folded_x for folded_x in folded_xs
     )
+
+
+def rewrite_pregrad(graph):
+    """Apply simplifying or stabilizing rewrites to graph that are safe to use
+    pre-grad.
+    """
+    return rewrite_graph(graph, include=("canonicalize", "stabilize"))
--- a/tests/test_pytensorf.py
+++ b/tests/test_pytensorf.py
@@ -30,6 +30,7 @@ from pytensor.graph.basic import Variable, equal_computations
 from pytensor.tensor.random.basic import normal, uniform
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.var import RandomStateSharedVariable
+from pytensor.tensor.slinalg import Cholesky
 from pytensor.tensor.subtensor import AdvancedIncSubtensor, AdvancedIncSubtensor1
 from pytensor.tensor.var import TensorVariable
 
@@ -878,3 +879,34 @@ class TestReplaceRVsByValues:
         [new_x], _ = _replace_vars_in_graphs([x], replacement_fn=replacement_fn)
 
         assert new_x.eval() > 50
+
+
+def test_mvnormal_no_cholesky_op():
+    """
+    Test MvNormal likelihood when using Cholesky factor parameterization does not unnecessarily
+    recompute the cholesky factorization
+    Reversion test of #6717
+    """
+    with pm.Model() as m:
+        n = 3
+        sd_dist = pm.HalfNormal.dist(shape=n)
+        chol, corr, sigmas = pm.LKJCholeskyCov("cov", n=n, eta=1, sd_dist=sd_dist)
+        mu = np.zeros(n)
+        data = np.ones((10, n))
+        pm.MvNormal("y", mu=mu, chol=chol, observed=data)
+
+    contains_cholesky_op = lambda fgraph: any(
+        isinstance(node.op, Cholesky) for node in fgraph.apply_nodes
+    )
+
+    logp = m.compile_logp()
+    assert not contains_cholesky_op(logp.f.maker.fgraph)
+
+    dlogp = m.compile_dlogp()
+    assert not contains_cholesky_op(dlogp.f.maker.fgraph)
+
+    d2logp = m.compile_d2logp()
+    assert not contains_cholesky_op(d2logp.f.maker.fgraph)
+
+    logp_dlogp = m.logp_dlogp_function()
+    assert not contains_cholesky_op(logp_dlogp._pytensor_function.maker.fgraph)


pymc/distributions/multivariate.py
```diff
@@ -1456,6 +1459,8 @@ def dist(cls, eta, n, sd_dist, *, compute_corr=True, **kwargs):
     @classmethod
     def helper_deterministics(cls, n, packed_chol):
         chol = pm.expand_packed_triangular(n, packed_chol, lower=True)
+        # tag as lower triangular to enable pytensor rewries
+        chol.tag.lower_triangular = True
```
ricardoV94: Can this be done in the `expand_packed_triangular`?

dehorsley: good idea!

ricardoV94:
This is a bit brute-force but can you benchmark after replacing these lines:
https://github.com/pymc-devs/pymc/blob/8c93bb54aca83fd915ca4b16a3a08feb7d432814/pymc/logprob/basic.py#L386-L387

By:
```python
    logp_terms = list(logp_terms.values())
    _check_no_rvs(logp_terms)

    from pytensor.compile import optdb
    from pytensor.graph import FunctionGraph
    
    rewrite = optdb.query("+canonicalize")
    fg = FunctionGraph(outputs=logp_terms, clone=False)
    rewrite.rewrite(fg)
```

That should remove the useless cholesky before the gradient is generated.

dehorsley:
Updated benchmarks:


|  | C backend |  JAX backend |
|-|---|-----|
|before |  245 ms ± 4.88 ms | 265 ms ± 7.38 ms | 
|after | 231 ms ± 9.16 ms | 106 ms ± 3.15 ms |
|after +  pre `grad` rewrite | 90.3 ms ± 2.15 ms | 109 ms ± 4.06 ms


@ricardoV94, do you think it's worth making the rewite optional via `Model.logp`? I can imagine it'd useful to see the "raw" logp graph while debugging

ricardoV94:
Awesome results. About the rewrites, we shouldn't introduce them now. Users should always look at the compiled function when they want to investigate the graph for performance concerns.

We should introduce the cholesky rewrite (and a couple others like `log1mexp` and other `useless`) when calling `model.dlogp`

ricardoV94:
By the way can you share the benchmark script so I can replicate locally?

We should add a test for the logp/dlogp to make sure the useless cholesky is removed (and avoid a regression in the future)

pymc/distributions/multivariate.py
ricardoV94: We can remove it from here right?

dehorsley:
@ricardoV94, is this the kind of thing you were thinking? I've made a new rewrite db if we want to be a bit more precise about what we want to perform. Not sure if that's the best place to put the rewrites.  I'll add tests in the next few days.

This is the script I was using for benchmarks (ipynb but I can't attach here):
<details>

```
# %%
import pymc as pm
import numpy as np
import pytensor
from pymc.blocking import DictToArrayBijection

# %%
print(pm.__version__)
print(pytensor.__version__)

# %%
n=1000
with pm.Model() as m:
    chol, corr, sigmas = pm.LKJCholeskyCov('cov', n=n, eta=1, sd_dist=pm.HalfNormal.dist())
    pm.MvNormal('y', mu=np.zeros(n), chol=chol, observed=np.ones((1000, n)))


# %%
logp = m.logp_dlogp_function()
logp.set_extra_values([])
pt = DictToArrayBijection.map(m.initial_point())
logp(pt)

# %%
%timeit -n 10 logp(pt)

# %%
pytensor.dprint(logp._pytensor_function.maker.fgraph.outputs[0])

# %%
pytensor.config.mode = pytensor.compile.get_default_mode().excluding("cholesky_ldotlt")

# %%
logp = m.logp_dlogp_function()
logp.set_extra_values([])
pt = DictToArrayBijection.map(m.initial_point())
logp(pt)

# %%
pytensor.dprint(logp._pytensor_function.maker.fgraph.outputs[0])

# %%
%timeit -n 10 logp(pt)

# %%
import jax 
import jax.numpy as jnp
from pymc.sampling.jax import get_jaxified_logp

# %%
logp_jax = get_jaxified_logp(m)
rvs = [rv.name for rv in m.value_vars]
init_position_dict = m.initial_point()
init_position = [jnp.array(init_position_dict[rv]) for rv in rvs]
grad_and_logp_jax = jax.jit(jax.value_and_grad(logp_jax))
grad_and_logp_jax(init_position)

# %%
%timeit -n 10 grad_and_logp_jax(init_position)

# %%
pytensor.compile.mode.JAX = pytensor.compile.mode.JAX.excluding("cholesky_ldotlt")

# %%
logp_jax = get_jaxified_logp(m)
rvs = [rv.name for rv in m.value_vars]
init_position_dict = m.initial_point()
init_position = [jnp.array(init_position_dict[rv]) for rv in rvs]
grad_and_logp_jax = jax.jit(jax.value_and_grad(logp_jax))
grad_and_logp_jax(init_position)

# %%
%timeit -n 10 grad_and_logp_jax(init_position)

# %%
m.optdb.print_summary()

```

</details>

pymc/logprob/rewriting.py
```diff
@@ -351,3 +351,18 @@ def construct_ir_fgraph(
...
     return fgraph, rv_values, memo
+
+
+pregrad_logprob_rewrites_db = SequenceDB()
+pregrad_logprob_rewrites_db.name = "pregrad_logprob_rewrites_db"
+pregrad_logprob_rewrites_db.register("pre-canonicalize", optdb.query("+canonicalize"), "basic")
+pregrad_logprob_rewrites_db.register("stabilize", optdb.query("+stabilize"), "basic")
+
+
+def rewrite_pregrad(graph, rewriter=None):
```
ricardoV94: I would maybe put this in `pytensorf`, we already have some custom rewrite logic going on there

pymc/model.py
```diff
@@ -308,6 +309,8 @@ class ValueGradFunction:
         back from the array dtype to the variable dtype.
     compute_grads: bool, default=True
         If False, return only the logp, not the gradient.
+    pregrad_rewrite: bool, default=True
```
ricardoV94: I would be opinionated and always run it (and not give the flag to users)?

pymc/logprob/rewriting.py
```diff
@@ -351,3 +351,18 @@ def construct_ir_fgraph(
         fgraph.replace_all(new_to_old)
 
     return fgraph, rv_values, memo
+
+
+pregrad_logprob_rewrites_db = SequenceDB()
+pregrad_logprob_rewrites_db.name = "pregrad_logprob_rewrites_db"
+pregrad_logprob_rewrites_db.register("pre-canonicalize", optdb.query("+canonicalize"), "basic")
+pregrad_logprob_rewrites_db.register("stabilize", optdb.query("+stabilize"), "basic")
```
ricardoV94: I wonder if this might be an overkill, should we start with only cherry picked rewrites?

ricardoV94 (commented):
Looks pretty good, just wondering whether we should be more targeted in the rewrites we include

pymc/logprob/rewriting.py
dehorsley: @ricardoV94 I'm not sure what we should do here. What kind of inclusion criteria are you thinking? I imagine anything that simplifies or improves the stability of the logp graph would be nice?

dehorsley:
New test will fail until pytensor version is bumped.

pymc/logprob/rewriting.py
ricardoV94: I was thinking of starting with just the things we know about. The reason for this is that we might want to use this as the "base" for pre-grad rewrites in PyTensor if/when we decide to make them lazy. This would be a good way to start learning what rewrites seem useful for gradients...  Probably everything in `canonicalize` makes sense as that makes the graphs "homogenous", but everything in `stabilize` might be an overkill or even counterproductive. `stabilizing` the base graph may make the gradient one more complex.

On another angle, I worry that the extra rewrites could make compilation much slower.

Could you perhaps benchmark how long does it take to obtain `dlogp` and, separately, to compile it for 2-3 different models?

The compilation should be checked with and without having already cached results. You can erase the pytensor cache with `pytensor-cache purge` or something similar.

pymc/distributions/multivariate.py
```diff
@@ -132,6 +132,9 @@ def quaddist_matrix(cov=None, chol=None, tau=None, lower=True, *args, **kwargs):
         chol = pt.as_tensor_variable(chol)
         if chol.ndim != 2:
             raise ValueError("chol must be two dimensional.")
+
+        # tag as lower triangular to enable pytensor rewriets of chol(l.l') -> l
```
```suggestion
        # tag as lower triangular to enable pytensor rewrites of chol(l.l') -> l
```

pymc/math.py
```diff
@@ -443,11 +443,17 @@ def expand_packed_triangular(n, packed, lower=True, diagonal_only=False):
     elif lower:
         out = pt.zeros((n, n), dtype=pytensor.config.floatX)
         idxs = np.tril_indices(n)
-        return pt.set_subtensor(out[idxs], packed)
+        # tag as lower triangular to enable pytensor rewries
```
ricardoV94:
```suggestion
        # tag as lower triangular to enable pytensor rewrites
```

pymc/math.py
```diff
@@ -443,11 +443,17 @@ def expand_packed_triangular(n, packed, lower=True, diagonal_only=False):
...
+        out = pt.set_subtensor(out[idxs], packed)
+        out.tag.lower_triangular = True
+        return out
     elif not lower:
         out = pt.zeros((n, n), dtype=pytensor.config.floatX)
         idxs = np.triu_indices(n)
-        return pt.set_subtensor(out[idxs], packed)
+        # tag as upper triangular to enable pytensor rewries
```
ricardoV94:
```suggestion
        # tag as upper triangular to enable pytensor rewrites
```

tests/test_pytensorf.py
```diff
@@ -878,3 +879,31 @@ def replacement_fn(var, replacements):
         [new_x], _ = _replace_vars_in_graphs([x], replacement_fn=replacement_fn)
 
         assert new_x.eval() > 50
+
+
+def test_mvnormal_no_cholesky_op():
+    """
+    Test MvNormal likelyhood when using Cholesky factor parameterization does not unnecessarily
```
ricardoV94:
```suggestion
    Test MvNormal likelihood when using Cholesky factor parameterization does not unnecessarily
```

ricardoV94 (commented):
Some small typos.

Does the last PyTensor already include the rewrites?

pymc/logprob/rewriting.py
ricardoV94: Yeah, I think this would be the easiest. If the rewrite time is not too affected by including all of `stabilize` vs not (just to test more easily), then I am okay with the current approach.

dehorsley:
Some rough benchmarks for compiling `dlogp`, with the C backend:

|       | pregrad rewites (pr  #6736) |        | dev (7b08fc1) |        |
|-------|-----------------------------|--------|---------------|--------|
| **model** | **uncached**                   | **cached** | **uncached**      | **cached** |
| Large MvNormal (m1)     | 135 s                       | 1.0 s  | 151 s         | 1.2 s  |
| Large no. normal vars (m2)     | 47 s                        | 2.4 s  | 56 s          | 3.0 s  |
| 8 schools, smallish every day model (m3)     | 33 s                        | 0.3 s  | 33 s          | 0.3 s  |

Generally, including the pregrad rewrites is faster overall, if not the same. I'm guessing the time saved by compiling the simpler dlogp is making up for the extra time spend optimising. Admittedly these are not very interesting models. Let me know if you have some others to test! 


## Methodology

<details>

For each of the following, I executed `pytensor-cache purge`, created the model, then ran 

```
%%time 
m.compile_dlogp()
```

For uncached, then for cached

```
%timeit m.compile_dlogp()
```
Models:

```python
n=1000
with pm.Model() as m1:
    chol, corr, sigmas = pm.LKJCholeskyCov('cov', n=n, eta=1, sd_dist=pm.HalfNormal.dist())
    pm.MvNormal('y', mu=np.zeros(n), chol=chol, observed=np.ones((1000, n)))

with pm.Model() as m2:
    sigma = pm.HalfCauchy('sigma', 1)
    for i in range(1,30):
        pm.Normal(f'm{i}', 0, i*sigma)


y = np.array([28, 8, -3, 7, -1, 1, 18, 12])
sigma = np.array([15, 10, 16, 11, 9, 11, 10, 18])
J = len(y)

with pm.Model() as m3:
    eta = pm.Normal("eta", 0, 1, shape=J)
    # Hierarchical mean and SD
    mu = pm.Normal("mu", 0, sigma=10)
    tau = pm.HalfNormal("tau", 10)

    # Non-centered parameterization of random effect
    theta = pm.Deterministic("theta", mu + tau * eta)

    obs = pm.Normal("obs", theta, sigma=sigma, observed=y)
```
I had to decrease the number of variables in model 2 as it was hitting the compilers nesting level max depth, but only in the case without the pregrad rewrites. (Of course there are other ways to get around this, and this is an intentionally pathological way to write this model!)

</details>

dehorsley:
> Some small typos.

Thanks, fixed.

> Does the last PyTensor already include the rewrites?

Yes, included since [2.12.2](https://github.com/pymc-devs/pytensor/releases/tag/rel-2.12.2). I'm not sure why tests are failing, looks like the test environment has v2.12.3.

ricardoV94:
Benchmarks look good. I'll mark this PR as requiring a major release and I'll ask other devs to keep an eye to see if their runtimes are significantly affected.

pymc/pytensorf.py
```diff
@@ -1228,3 +1230,18 @@ def constant_fold(
...
     )
+
+
+pregrad_logprob_rewrites_db = SequenceDB()
+pregrad_logprob_rewrites_db.name = "pregrad_logprob_rewrites_db"
+pregrad_logprob_rewrites_db.register("pre-canonicalize", optdb.query("+canonicalize"), "basic")
+pregrad_logprob_rewrites_db.register("stabilize", optdb.query("+stabilize"), "basic")
+
+
+def rewrite_pregrad(graph, rewriter=None):
```
ricardoV94: Can we use this helper? 

https://github.com/pymc-devs/pytensor/blob/60bc3688f67b2bcb0fa4f833d2d7e3f4b2671bb7/pytensor/graph/rewriting/utils.py#L20

Also, any reason why you created a custom rewrite db, instead of just doing a query on the global `optdb`? I think the second should be fine? That's what using the helper would do, by just specifying `include`

ricardoV94 (changes requested):
Just a last-minute doubt about the helper. WDYT?

pymc/pytensorf.py
dehorsley: I created a custom rewrite db, as I was expecting to be more precise about what we wanted and followed the logprob module. Going with the helper looks better to me! I fix this up in the next day or so.

ricardoV94: I see, the logprob submodule is special in that those rewrites are not general, but exclusive to the rv-logp conversion. Here we're thinking about general rewrites so the `optdb` database makes sense (at least initially).

dehorsley:
@ricardoV94, should be good to go assuming tests all pass

ricardoV94 (approved):

ricardoV94:
Thanks a lot @dehorsley!

PR #6740, opened 2023-05-29
GP Covariance Function Type Hints

jahall:
Adding type-hints to `cov.py`. Also moving `Periodic` to appear below the other stationary kernels.

## New features
- Type hints in `cov.py`

----
:books: Documentation preview :books:: https://pymc--6740.org.readthedocs.build/en/6740/


pymc/gp/cov.py
```diff
@@ -978,13 +1058,13 @@ def full(self, X, Xs=None):
...
 
-    def diag(self, X):
+    def diag(self, X: TensorLike) -> TensorVariable:
         X, _ = self._slice(X, None)
         index = pt.cast(X, "int32")
         return pt.diag(self.B)[index.ravel()]
 
 
-def handle_args(func, args):
+def handle_args(func: Callable) -> Callable:
```
bwengals, line 1075: args got lost here

pymc/gp/cov.py
```diff
@@ -842,7 +908,7 @@ def __init__(self, input_dim, lengthscale_func, args=None, active_dims=None):
                 raise NotImplementedError(("Higher dimensional inputs ", "are untested"))
         if not callable(lengthscale_func):
             raise TypeError("lengthscale_func must be callable")
-        self.lfunc = handle_args(lengthscale_func, args)
+        self.lfunc = handle_args(lengthscale_func)
```
bwengals, line 919: should be `self.lfunc = handle_args(lengthscale_func, args)`

pymc/gp/cov.py
```diff
@@ -559,11 +560,12 @@ class ExpQuad(Stationary):
...
 
-    def full(self, X, Xs=None):
+    def full(self, X: TensorLike, Xs: Optional[TensorLike] = None) -> TensorVariable:
         X, Xs = self._slice(X, Xs)
-        return pt.exp(-0.5 * self.square_dist(X, Xs))
+        r2 = self.square_dist(X, Xs)
+        return pt.exp(-0.5 * r2)
 
-    def power_spectral_density(self, omega):
+    def power_spectral_density(self, omega) -> TensorVariable:
```
bwengals: omega is `TensorLike`

pymc/gp/cov.py
```diff
@@ -71,29 +78,29 @@ def __call__(self, X, Xs=None, diag=False):
...
-    def diag(self, X):
+    def diag(self, X: TensorLike) -> TensorVariable:
         raise NotImplementedError
 
-    def full(self, X, Xs=None):
+    def full(self, X: TensorLike, Xs: Optional[TensorLike] = None) -> TensorVariable:
         raise NotImplementedError
 
-    def __add__(self, other):
+    def __add__(self, other) -> "Add":
```
bwengals, line 88: might be an ignorant question, but why is the return type a string instead of `Add`?

.pre-commit-config.yaml
```diff
@@ -49,6 +49,7 @@ repos:
   rev: 6.3.0
   hooks:
     - id: pydocstyle
+      additional_dependencies: [tomli]
```
bwengals: what's this for?

pymc/gp/cov.py
jahall: It's actually not necessary to pass them when constructing the callable. They're only required when calling the returned func.

pymc/gp/cov.py
jahall: Done

pymc/gp/cov.py
jahall: Basically because the `Add` class has not yet been defined (as it is a subclass of this one), therefore, as per [PEP 484](https://peps.python.org/pep-0484/#forward-references) you can define it as a string instead and mypy will work accordingly. Though I *think* as of 3.10 you should be able to specify without the string.

.pre-commit-config.yaml
jahall: When I ran the pre-commit hooks the pydocstyle was failing for me, complaining that `tomli` was not installed...

pymc/gp/cov.py
bwengals: oh yeah youre right, ok let's keep

.pre-commit-config.yaml
bwengals: Might be better to leave out of this particular PR then... Maybe it would be better raised as an issue?  I just don't know much about precommit and these things and so would rather not slip this in with a GP PR.

jahall: Will remove

bwengals:
Hey @jahall sorry for the delay!  I (finally) enabled the test suite for you.  Looks like `test_handleargs` needs a small update to reflect your change, and there are a couple mypy issues.  There's a script in the repo to run the same mypy checks locally that the test runner does,  `python scripts/run_mypy.py --verbose`.  By the way, thanks for your contributions here and for the new periodic kernel.  Looking forward to using it!

jahall:
@bwengals No problem! That should be mypy/tests fixed now. I'll wait till we merge this PR then will update the WrappedPeriodic PR to reflect these changes.

jahall:
@bwengals ...bit confused on the test failures ...

ricardoV94:
They showed up in another PR, I think it's unrelated, probably something that changed with the latest PyTensor release

ricardoV94:
Should be fixed by https://github.com/pymc-devs/pymc/pull/6782

jahall:
@bwengals @ricardoV94 This should be good to go now after merging Ricardo's fix

jahall:
@bwengals @ricardoV94 Should be good to go now after merging latest main.

jahall:
@bwengals @ricardoV94 Thoughts?

pymc/gp/cov.py
```diff
@@ -129,6 +141,10 @@ def __array_wrap__(self, result):
                 "Known types are `Add` or `Prod`."
             )
 
+    @staticmethod
+    def _alloc(X, *shape: int) -> TensorVariable:
+        return pt.alloc(X, *shape)  # type: ignore
+
```
bwengals, line 148: Is this factored out for mypys sake?

bwengals (approved):
thank you @jahall !

pymc/gp/cov.py
jahall: Ah, yes, this was for mypys sake...I think because `pt.alloc` can technically return things other than `TensorVariable`...

jahall:
@ricardoV94 Can we merge this?

pymc/gp/cov.py
ricardoV94: It cannot, but mypy doesn't understand that. If you would like ta take a look at the PyTensor level to see if it can be fixed that would be great!

jahall: Sure - let me take a look

PR #6741, opened 2023-05-29
Ignore named variables that are not traceable in `get_vars_in_point_list`

ricardoV94:
This showed up when trying to do posterior predictive on https://github.com/pymc-devs/pymc-experimental/pull/168

It would consider a named_var that is now just a `ConstantData` or `MutableData` as having a trace equivalent. The `compile_forward_function` would then fail because we can't pass constants or shared variables as explicit inputs to a PyTensor function

----
:books: Documentation preview :books:: https://pymc--6741.org.readthedocs.build/en/6741/

### DIFF
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -89,7 +89,8 @@ def get_vars_in_point_list(trace, model):
         names_in_trace = list(trace[0])
     else:
         names_in_trace = trace.varnames
-    vars_in_trace = [model[v] for v in names_in_trace if v in model]
+    traceable_varnames = {var.name for var in (model.free_RVs + model.deterministics)}
+    vars_in_trace = [model[v] for v in names_in_trace if v in traceable_varnames]
     return vars_in_trace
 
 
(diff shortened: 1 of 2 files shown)


pymc/sampling/forward.py
```diff
@@ -89,7 +89,10 @@ def get_vars_in_point_list(trace, model):
         names_in_trace = list(trace[0])
     else:
         names_in_trace = trace.varnames
-    vars_in_trace = [model[v] for v in names_in_trace if v in model]
+    traceable_varnames = {
+        var.name for var in (model.free_RVs + model.observed_RVs + model.deterministics)
```
lucianopaz: Why should we include `observed_RVs` here? Will they ever take their values from the trace at this point?

lucianopaz (changes requested):
Looks good but why do you include `observedRVs`? Wont that cause problems for `sample_posterior_predictive`? Maybe I'm missing something here.

pymc/sampling/forward.py
ricardoV94: Yeah I think you're right

ricardoV94: Fixed

lucianopaz (approved):
Lgtm

PR #6698, opened 2023-04-29
Rectification in MeasurableCheckAndRiase and tests

Dhruvanshu-Joshi:
**What is this PR about?**
This PR aims to provide details of the approaches that I tried to solve the issue [#6684](https://github.com/pymc-devs/pymc/issues/6684). There were some errors in ```pymc/logprob/checks.py``` and ```tests/logprob/test_checks.py``` which this PR aims to solve.

**Checklist**
+ [x] Explain important implementation details 👆
+ [x] Make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).
+ [x] Link relevant issues (preferably in [nice commit messages](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html))
+ [ ] Are the changes covered by tests and docstrings?
+ [ ] Fill out the short summary sections 👇

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6698.org.readthedocs.build/en/6698/

PR #6746, opened 2023-05-31
Allow CustomDist with inferred logp in Mixture

### LINKED ISSUE #6728: Mixture with CustomDist with inferred logp fails
### Description

Reported in https://discourse.pymc.io/t/nonetype-error-on-custom-distribution/12176/7?u=ricardov94

```python
import numpy as np
import pymc as pm

def dist_beta(a, b, loc, scale, size):
    return pm.Beta.dist(a, b, size=size) * scale + loc

with pm.Model() as m:

    left_beta = pm.CustomDist.dist(
        1,
        1,
        -0.5,
        2,
        dist=dist_beta,
    )

    right_beta = pm.CustomDist.dist(
        1,
        1,
        0.5,
        2,
        dist=dist_beta,
    )

    w = pm.Dirichlet('w', np.ones(2))
    resp = pm.Mixture('resp', w=w, comp_dists=[left_beta, right_beta])
    
m.logp()  # NotImplementedError: Logprob method not implemented for Elemwise{add,no_inplace}
```

ricardoV94:
This PR changes the logic used for logprob inference. Instead of eager bottom-up conversion to measurable variables in the IR rewrites, we only convert nodes whose outputs were marked as "needs_measuring". This is achieved with the new `PreserveRVMappings.request_measurable` method.

This strategy obviates the need to undo unnecessary conversions. It also obviates a subtle need for graph cloning via the `ignore_logprob` helper, which prevented intermediate measurable rewrites from being reversed when they were needed to derive the logprob of valued variables, but were not directly valued. This indirect role of `ignore_logprob` is now done more explicitly and efficiently via the `request_measurable` method.

All other uses of `ignore_logprob` (and `reconsider_logprob`) were removed from the codebase

The `get_measurable_outputs` dispatching was also abandoned in favor of only considering outputs associated with value variables.

A new MergeOptimizerRewrite was written to further target local rewrites to only those nodes whose variables have been marked as `needs_measuring`. This should be rather more efficient

Closes #6728 
Supersedes #6698 

CC @shreyas3156 @Dhruvanshu-Joshi 

----
:books: Documentation preview :books:: https://pymc--6746.org.readthedocs.build/en/6746/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,6 +105,7 @@ jobs:
             tests/logprob/test_abstract.py
             tests/logprob/test_basic.py
             tests/logprob/test_binary.py
+            tests/logprob/test_checks.py
             tests/logprob/test_censoring.py
             tests/logprob/test_composite_logprob.py
             tests/logprob/test_cumsum.py
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -36,7 +36,6 @@ from pytensor.tensor.var import TensorConstant, TensorVariable
 
 import pymc as pm
 
-from pymc.logprob.abstract import _get_measurable_outputs
 from pymc.pytensorf import convert_observed_data
 
 __all__ = [
@@ -133,11 +132,6 @@ class MinibatchIndexRV(IntegersRV):
         if rng is None:
             rng = pytensor.shared(np.random.default_rng())
         return super().make_node(rng, *args, **kwargs)
-
-
-@_get_measurable_outputs.register(MinibatchIndexRV)
-def minibatch_index_rv_measuarable_outputs(op, node):
-    return []
 
 
 minibatch_index = MinibatchIndexRV()
--- a/pymc/distributions/bound.py
+++ b/pymc/distributions/bound.py
@@ -26,7 +26,6 @@ from pymc.distributions.distribution import Continuous, Discrete
 from pymc.distributions.shape_utils import to_tuple
 from pymc.distributions.transforms import _default_transform
 from pymc.logprob.basic import logp
-from pymc.logprob.utils import ignore_logprob
 from pymc.model import modelcontext
 from pymc.pytensorf import floatX, intX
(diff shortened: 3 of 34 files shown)


.github/workflows/tests.yml
```diff
@@ -105,6 +105,7 @@ jobs:
             tests/logprob/test_abstract.py
             tests/logprob/test_basic.py
             tests/logprob/test_binary.py
+            tests/logprob/test_checks.py
```
ricardoV94, line 108: This feels very dangerous, there is no fallback for tests that are not included explicitly?

These tests were completely broken for a while

Dhruvanshu-Joshi: Previously, the asserts were not being defined in this test effectively and hence the desired exceptions were not raised. I think the current changes should be sufficient to rectify this.

ricardoV94: The problem I am referring to is not the underlying bugs, but the fact that we were not running the tests for this file at all, which is a more broad issue.

twiecki:
image
🔥

.github/workflows/tests.yml
michaelosthege: We have a script under `./scripts` that should have caught this. Why didn't it?

ricardoV94: No idea, can you take a look?

michaelosthege: Log indicates that the check was skipped.
Maybe because this line doesn't match the filename of the `tests.yml` workflow: https://github.com/pymc-devs/pymc/blob/4d2b2efc2b8ad7b5060d3f4a642534408b5551eb/.pre-commit-config.yaml#L67


I'm just on my phone in the train, so I can't open a PR. Can you try it on this branch?

ricardoV94: Opened PR in #6756

twiecki (approved):

PR #6765, opened 2023-06-09
Fix bug in switch mixture logp

ricardoV94:
----
:books: Documentation preview :books:: https://pymc--6765.org.readthedocs.build/en/6765/


twiecki (approved):

larryshamalama (commented):
Well spotted! Thanks @ricardoV94

IIRC, I also think that `find_measurable_switch_mixture` does not broadcast well (the `NoneConst` should be more general)

Issue #6612, opened 2023-03-18
Add icdf functions for distributions


michaelraczycki:
### Description

We are looking for help to implement inverse cumulative distribution (ICDF) functions for our distributions!

### How to help?

This PR should give a template on how to implement and test new icdf functions for distributions: https://github.com/pymc-devs/pymc/pull/6528

ICDF functions allow users to get the value associated with a specific cumulative probability. 

So far we've added 2 examples for continuous distribution 
* Uniform:
https://github.com/pymc-devs/pymc/blob/2fcce433548a411f695d684ffb2d001a82a35f20/pymc/distributions/continuous.py#L348-L351
* Normal: 
https://github.com/pymc-devs/pymc/blob/2fcce433548a411f695d684ffb2d001a82a35f20/pymc/distributions/continuous.py#L541-L548

And an example for a discrete distribution:
* Geometric:
https://github.com/pymc-devs/pymc/blob/2fcce433548a411f695d684ffb2d001a82a35f20/pymc/distributions/discrete.py#L824-L832

Multiple sources describing the icdf function for any specific distribution can be found, you're free to choose which one is working for you. To start with I recommend checking:

* https://help.imsl.com/c/2016/html/cnlstat/index.html#page/CNL%2520Stat%2Fcsch11.14.01.html%23
* Wikipedia,
E.g.: https://en.wikipedia.org/wiki/Normal_distribution
 ![Zrzut ekranu 2023-03-18 o 15 49 21](https://user-images.githubusercontent.com/119355382/226107058-645d110a-07ad-445d-b540-b8b73b6346ef.png)
 It should be called "Quantile" as on the screenshot above.

New tests have to be added in test_continuous.py for continuous distributions, and test_discrete.py for discrete ones. You can use existing tests as a template:

https://github.com/pymc-devs/pymc/blob/2fcce433548a411f695d684ffb2d001a82a35f20/tests/distributions/test_continuous.py#L282-L286

Don't hesitate to ask any questions. You can grab as many distributions to implement moments as you want. Just make sure to write in this issue so that we can keep track of it.

Profit with your new open source KARMA!

The following distributions don't have an icdf method implemented:

- [x]  Beta #6845
- [x]  Kumaraswamy  #6642 
- [x]  Exponential #6641
- [x]  Laplace #6707 
- [x]  StudentT #6845 
- [x]  Cauchy  #6747
- [x]  HalfCauchy
- [x]  Gamma #6845 
- [x]  HalfNormal
[...]

PR #6766, opened 2023-06-09
Add icdf functions for Lognormal, Half Cauchy and Half Normal distributions

amyoshino:
**What is this PR about?**
Adds ICDF functions to Lognormal Distribution

Issue https://github.com/pymc-devs/pymc/issues/6612
Comment on issues found while working on Issue https://github.com/pymc-devs/pymc/issues/6747

**References:**
Lognormal: 
- https://en.wikipedia.org/wiki/Log-normal_distribution
- https://help.imsl.com/c/2016/html/cnlstat/index.html#page/CNL%2520Stat%2Fcsch11.14.40.html%23

HalfCauchy: matches formulas used in either SciPy and R extraDistr
- [SciPy implementation](https://github.com/scipy/scipy/blob/d604a5462dadf799f8af2c177179a01875b40796/scipy/stats/_continuous_distns.py#L4167)
- [R extraDistr implementation](https://github.com/twolodzko/extraDistr/blob/master/src/half-cauchy-distribution.cpp#L58)

HalfNormal: matches formulas used in either SciPy and R extraDistr
- [SciPy implementation](https://github.com/scipy/scipy/blob/d604a5462dadf799f8af2c177179a01875b40796/scipy/stats/_continuous_distns.py#L4288)
- [R extraDistr implementation](https://github.com/twolodzko/extraDistr/blob/master/src/half-normal-distribution.cpp#L54)

**Checklist**

## Major / Breaking Changes
- ...

## New features
- LogNormal icdf function

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6766.org.readthedocs.build/en/6766/


amyoshino:
I am having an issue when developing the test for this function. 
The implementation of the inverse CDF for the lognormal should be very straightforward, from Wikipedia we can see that the lognormal quantile function is just the exponential of the normal quantile function:
Normal: 
image
Lognormal: 
image

So, I just got the already implemented icdf function for the normal and added np.exp() to it.

But it looks like the implemented function for the normal distribution icdf has a tiny difference if compared to the scipy implementation :
For mu = 2.1, sigma = 20.0, p = 0.99
np.exp(mu + sigma * -np.sqrt(2.0) * pt.erfcinv(2 * value)) - st.norm.ppf(value, mu, sigma) =  7.105427357601002e-15 
(passes the test of 6 digits precision)

But when this is passed to the exponential function, the difference becomes larger than the tolerance of 6 decimals and the test fail. 

image


Using the values of the failed test above, I get the same values of error locally, and the result for the icdf differs both from scipy and R stats (`qlnorm` function) starting on the 8th digit:
image

Investigating it further, the Scipy implementation uses a different way to compute the icdf by approximating it with a routine described by this script: https://github.com/scipy/scipy/blob/2f3831503aff159994eafa75745c9537f8db060f/scipy/special/cephes/ndtri.c#L1

Which is also different from the R stats qnorm function:
https://github.com/wch/r-source/blob/a6d764783f5010268a33e3610189983e9bb778db/src/nmath/qnorm.c#L23
That is used to get qlnorm by exponentiating it: https://github.com/wch/r-source/blob/a6d764783f5010268a33e3610189983e9bb778db/src/nmath/qlnorm.c#L29

Hence the difference in the results. Any ideas on how to handle this divergences on implementations during the tests?

ricardoV94:
Small numerical differences are fine. You can get around them by either tuning `decimal` or choosing different set of `parameters` so that more extreme combinations don't show up in the tests.

For testing locally, don't forget to set `n_samples=-1` so that all combinations are tested.

We have been thinking about changing the precision criteria because the one used now is pretty dumb: https://github.com/pymc-devs/pymc/pull/6159

amyoshino:
@ricardoV94, thank you for the comments of handling the precision issues. I have successfully added tests for the lognormal icdf function. 
Some tests are failing, and I'm not sure why because it looks like things I worked on are not related to the failures. 
I think I have updated the branch right before committing, but I might be wrong, so if it is the case let me know.

ricardoV94:
> I also added the icdf functions for the HalfCauchy and HalfNormal distributions but didn't manage to use the function you pointed out in https://github.com/pymc-devs/pymc/issues/6612.

Can you share what did you try exactly?

pymc/distributions/continuous.py
```diff
@@ -856,6 +856,15 @@ def logcdf(value, loc, sigma):
             msg="sigma > 0",
         )
 
+    def icdf(value, loc, sigma):
+        res = Normal.icdf((value + 1.0) / 2.0, loc, sigma)
```
ricardoV94: Should use `pm.icdf(Normal.dist(loc, sigma), value)` instead. Same in other places where you used `.icdf` directly

pymc/distributions/continuous.py
```diff
@@ -856,6 +856,15 @@ def logcdf(value, loc, sigma):
...
+        res = check_icdf_value(res, value)
+        return check_icdf_parameters(
+            res,
+            sigma > 0,
+            msg="sigma > 0",
+        )
```
ricardoV94: None of these checks are needed, because the Normal icdf already does it

pymc/distributions/continuous.py
```diff
@@ -1714,12 +1723,22 @@ def logcdf(value, mu, sigma):
...
         )
 
+    def icdf(value, mu, sigma):
+        res = pt.exp(Normal.icdf(value, mu, sigma))
+        res = check_icdf_value(res, value)
+        return check_icdf_parameters(
+            res,
+            sigma > 0,
+            msg="sigma > 0",
+        )
```
ricardoV94: Checks aren't needed because the icdf of the Normal already does it

ricardoV94 (changes requested):
Looks great, only some small changes needed

ricardoV94:
> @ricardoV94, thank you for the comments on handling the precision issues. I have successfully added tests for the lognormal icdf function. I also added the icdf functions for the HalfCauchy and HalfNormal distributions but didn't manage to use the function you pointed out in #6612.
> 
> ```
> def icdf(value, *args)
>   return icdf(Full.dist(*args), value)
> ```
> 
> Instead I used the formulas consistent with the implementations found in [SciPy](https://github.com/scipy/scipy/blob/d604a5462dadf799f8af2c177179a01875b40796/scipy/stats/_continuous_distns.py#L4288) and R [extraDistr](https://github.com/twolodzko/extraDistr/blob/master/src/half-normal-distribution.cpp#L54) packages.

You're right, thanks for checking. Opened related PR to fail explicitly for the icdf

pymc/distributions/continuous.py
ricardoV94: This is a bit annoying because we don't allow users to create a `HalfNormal` with `loc != 0` but in theory they could call `icdf` or reach this function with a HalfNormal that has a custom non-zero `loc`. 

My question is, will this expression work for non-zero `loc`?

This is also a question for the `HalfCauchy`. 

I wonder if the best solution is to reimplement these RandomVariables directly in PyMC in a way that they don't accept a `loc` argument (just like the `HalfStudentTRV` in this file). This way we don't have to worry about `loc` in the logp/logcdf/icdf/moment functions.

If we go down that path, we can remove the current `HalfNormal` and `HalfCauchy` RandomVariables from PyTensor as they aren't really needed there.

pymc/distributions/continuous.py
amyoshino: When removing the `res = check_icdf_value(res, value)`, it fails the tests for `q < 0` or `q > 1`  since we are getting the icdf through `Normal.dist()` and `pymc.logprob.basic.icdf()` and both don't check for it automatically. 

For example, in:

```
from pymc.logprob.basic import icdf

def icdf(value, loc, sigma):
    res = icdf(Normal.dist(loc, sigma), (value + 1.0) / 2.0)
    #res = check_icdf_value(res, value)
    return res
```

I get the following test failure:

image

With `res = check_icdf_value(res, value)` uncommented the test pass successfully. 

Should I keep it or am I missing anything?

pymc/distributions/continuous.py
amyoshino: Alright!! Thanks for the revision, changed that for the next commit!

amyoshino: The implemented formula work well even when `loc != 0` and is consistent with SciPy results:

Testing HalfNormal:
image

Testing HalfCauchy:
image

Allow me some more time to bring some screenshots calling the implemented functions with `pm.HalfNormal.icdf()` and `pm.HalfCauchy.icdf()`

If you think the best solution is still to reimplement RandomVariables directly in PyMC in a way that they don't accept a loc argument, let me know.

ricardoV94: We don't need to do it in this PR, but the fact that it isn't being tested in our suite (and not just the new icdf) is already a good argument to drop it

ricardoV94: Good news that it works though!

pymc/distributions/continuous.py
ricardoV94: Oh because we are transforming the value that goes into the Normal icdf. Yeah let's keep that. The sigma shouldn't be needed still

pymc/distributions/continuous.py
ricardoV94: I guess you didn't push the changes yet?

amyoshino: Just did it! I was waiting for the other threads to resolve and push it all in one commit.

pymc/distributions/continuous.py
amyoshino: Sounds good!! Thanks

pymc/distributions/continuous.py
amyoshino: Same as above.

pymc/distributions/continuous.py
```diff
@@ -1714,12 +1720,18 @@ def logcdf(value, mu, sigma):
...
+
         return check_parameters(
             res,
             sigma > 0,
             msg="sigma > 0",
         )
 
+    def icdf(value, mu, sigma):
+        res = pt.exp(icdf(Normal.dist(mu, sigma), value))
+        res = check_icdf_value(res, value)
```
ricardoV94: This one isn't needed, since the same value is used in the normal icdf and has the check already

amyoshino: I got confused with this part being necessary in the HalfNormal and kept it here for the LogNormal as well. Sorry about that! This is fixed now.

ricardoV94: No problem

ricardoV94 (approved):

ricardoV94:
Thanks @amyoshino!

Issue #6350, opened 2022-11-29
Implement logprob derivation for order statistics


ricardoV94:
### Description

The idea would be to derive the logprob for the following type of graphs

```python
import aesara.tensor as at
import pymc as pm

y_rv = at.max(at.random.uniform(0, 1, size=3))   # or min
y_rv = at.sort(at.random.uniform(0, 1, size=3))
y_rv = at.sort(at.random.uniform(0, 1, size=3))[idx]  # max / min correspond to idx==-1 or idx==0

y_vv = y_rv.clone()
pm.logprob.factorized_joint_logprob({y_rv: y_vv})
```

https://en.wikipedia.org/wiki/Order_statistic#Probabilistic_analysis

This might be a bit far-fetched / difficult to find a good general solution that goes beyond a few simple cases (e.g, order statistics with non-i.i.d RVs):

```python
y_rv = at.max(at.random.uniform([0, 1], [2, 3]))
y_rv = at.max(at.stack([at.random.uniform(0, 1), at.random.normal(0, 1)]))
```

Probability for order statistics of IID variables are pretty straightforward to obtain, requiring expressions that depend on the CDFs. For non IID, things grow quickly in complexity: https://en.wikipedia.org/wiki/Bapat%E2%80%93Beg_theorem

ricardoV94:
Superseded by #7121

Issue #334, opened 2013-09-08
[pymc3] Gamma(1,1) MAP issue


kforeman:
I've found that when I run find_MAP() on a Gamma(1,1) distribution it ends up with negative value (and thus -Inf logp).

Here's an illustration of the issue: http://nbviewer.ipython.org/6488376

If I change it to Gamma(10,1) it works fine...

kforeman:
Did some testing and found that the error occurs for any value of alpha <= 1.0. beta has no effect.

Notebook: http://nbviewer.ipython.org/6490621

jsalvatier:
This is because the mode is 0 for alpha < 1
http://en.wikipedia.org/wiki/Gamma_distribution

It would be nice if the scipy optimization functions would actually find
this value, but they don't seem to. Any suggestions on what kind of
error/behavior would have been more informative?

twiecki:
Seems like it might be related to https://github.com/pymc-devs/pymc/issues/289.

For some reason the optimization is not good at backtracking once it walked out of the support.

kforeman:
Ah, interesting. I think Gamma(1,1) might come up relatively often, since it corresponds to Gelman's Inv-Gamma(1,1) uninformative prior on variances for hierarchical models. 

In my case I'd just substitute alpha=1.0001, but that of course gets dicey as a general rule. 

I'd vote for an error message along the lines of "Optimized value for tau (-0.00048828) is outside of its support", as it wasn't clear that's what was going on. Especially in models with lots of parameters, you have to backtrack to figure out which array entries for max/logp correspond to which parameters.

jsalvatier:
Yeah, 'outside support' is a good phrase to have in that error message.

jsalvatier:
Right now we're using `scipy.fmin_bfgs` as the default optimizer, and as you guys have both seen, it doesn't deal with boundaries very well at all. I'd love to switch to something more robust, even at the expense of some speed (though it has to have decent asymptotic behavior), but I don't know what to switch to. 

I've brought this up to Scipy before, but without much result: https://github.com/scipy/scipy/issues/2169

kforeman:
`scipy.optimize.fmin_tnc` works for alpha<1 and is significantly faster. I think that might be a good solution.

`scipy.optimize.fmin_l_bfgs_b` could also work by adding in the bounds=[] arguments.

kforeman:
[...]

PR #6769, opened 2023-06-11
Logprob derivation for Max of continuous IID variables

Dhruvanshu-Joshi:
**What is this PR about?**
This PR aims to provide a solution for implementing the Max operation from order statistics to solve the issue [#6350] (https://github.com/pymc-devs/pymc/issues/6350) and issue [#6773](https://github.com/pymc-devs/pymc/issues/6773). More tests are to be added.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- Logprob derivation for ```pt.max```

## Bugfixes
- ...
## To-do:
+ [ ] Derive logprob for Max
+ [ ] Add tests to verify the logprob generated for max operation
+ [ ] Derive logprob for Min
+ [ ] Add tests to verify the logprob generated for min operation
+ [ ] Add support for different types of Distributions
## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6769.org.readthedocs.build/en/6769/


pymc/logprob/order.py
```diff
@@ -0,0 +1,114 @@
...
+    if rv_map_feature is None:
+        return None  # pragma: no cover
+
+    if isinstance(node.op, MeasurableMax):
+        return None  # pragma: no cover
+
+    base_var = node.inputs[0]
+
+    if base_var.owner.inputs[3].type.ndim != 0:
+        return None
```
ricardoV94: This is implicitly assuming the input to the Max is a variable with a `RandomVariable` `Op`, in which case we should check explicitly. Also we should check all the distribution parameters, which would be all `inputs[3:]`

pymc/logprob/order.py
```diff
@@ -0,0 +1,114 @@
...
+
+    if base_rv_op.name:
+        logprob.name = f"{base_rv_op}_logprob"
+        logcdf.name = f"{base_rv_op}_logcdf"
+
+    size_var = base_rv.owner.inputs[1]
+    string_size = str(size_var)
+    for b in (0, len(string_size) - 1):
+        if string_size[b] == "}":
+            a = string_size[b - 1]
```
ricardoV94: This should just use `value.size`. In general you don't need to worry about having constants, it's fine for a model to have a `size` that's dynamic and perfectly valid to create PyTensor logprob expressions using non-constant size variables.

tests/logprob/test_order.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_max():
+    x = pt.random.normal(0, 1, size=(3,))
+    x_name = "x"
+    x_max = pt.max_and_argmax(x, axis=-1)
+    x_max_value = pt.vector("x_max_value")
+    x_max_logprob = logp(x_max[0], x_max_value)
+
+    assert_no_rvs(x_max_logprob)
```
ricardoV94: We should test a couple of things more:
1. This fails when we ask for the logprob of argmax
2. This fails when an input is not a RandomVariable (just because it's a limitation for now)
3. This fails when the variable is not IID
4. This gives the expected result when evaluated

pymc/logprob/order.py
```diff
@@ -0,0 +1,114 @@
...
+from pytensor.graph.fg import FunctionGraph
+from pytensor.graph.rewriting.basic import node_rewriter
+from pytensor.tensor.math import MaxAndArgmax
+
+from pymc.logprob.abstract import MeasurableVariable, _logcdf, _logprob
+from pymc.logprob.rewriting import measurable_ir_rewrites_db
+
+
+class MeasurableMax(MaxAndArgmax):
+    """A placeholder used to specify a log-likelihood for a clipped RV sub-graph."""
```
ricardoV94:
```suggestion
    """A placeholder used to specify a log-likelihood for a max sub-graph."""
```

pymc/logprob/order.py
```diff
@@ -0,0 +1,114 @@
...
+
+    base_var = node.inputs[0]
+
+    if base_var.owner.inputs[3].type.ndim != 0:
+        return None
+
+    if not rv_map_feature.request_measurable(node.inputs):
+        return None
+
+    axis = node.op.axis
```
ricardoV94: We have to handle axis either here (e.g., not supporting anything other than `max` across all axis, or in the logp (supporting different axis). 

As with the other limitations, we can start more strict and gradually expand the functionality

pymc/logprob/order.py
```diff
@@ -0,0 +1,114 @@
...
+from pytensor.graph.basic import Node
+from pytensor.graph.fg import FunctionGraph
+from pytensor.graph.rewriting.basic import node_rewriter
+from pytensor.tensor.math import MaxAndArgmax
+
+from pymc.logprob.abstract import MeasurableVariable, _logcdf, _logprob
+from pymc.logprob.rewriting import measurable_ir_rewrites_db
+
+
+class MeasurableMax(MaxAndArgmax):
```
ricardoV94: Actually there seems to be a pure `Max` operator in PyTensor. So to make our lives easier, we can have a rewrite that replaces `MaxAndArgmax` by an equivalent `Max`, and then have this rewrite only have to worry about `Max`ed variables.

The Op is defined here: https://github.com/pymc-devs/pytensor/blob/ec6a3153fbf5aaa023368f36aac58212c9bedbd4/pytensor/tensor/math.py#L622-L630

There is also one for `Argmax`: https://github.com/pymc-devs/pytensor/blob/ec6a3153fbf5aaa023368f36aac58212c9bedbd4/pytensor/tensor/math.py#L331

This kind of intermediate rewrites that don't return a MeasurableVariable look like this: https://github.com/pymc-devs/pymc/blob/765fdef219ea821dcc8ddbbd4291dc42e739341b/pymc/logprob/transforms.py#L490-L494

ricardoV94 (commented):
Looks like a great start, left some comments below.

pymc/logprob/order.py
ricardoV94: Actually the rewrite already exists here: https://github.com/pymc-devs/pytensor/blob/ec6a3153fbf5aaa023368f36aac58212c9bedbd4/pytensor/tensor/rewriting/uncanonicalize.py#L45C1-L60

We can simply register it in our pre-canonicalization phase: https://github.com/pymc-devs/pymc/blob/765fdef219ea821dcc8ddbbd4291dc42e739341b/pymc/logprob/rewriting.py#L360

I don't know the syntax from memory. Maybe `optdb.query("+canonicalize").including(name_of_extra_rewrite)`?

pymc/logprob/order.py
```diff
@@ -0,0 +1,114 @@
...
+    "find_measurable_max",
+    find_measurable_max,
+    "basic",
+    "max",
+)
+
+
+@_logprob.register(MeasurableMax)
+def max_logprob(op, values, base_rv, **kwargs):
+    (value,) = values
```
ricardoV94, line 118: We haven't been doing this consistently so far but we should add some docstrings that explain what the `logprop` implementation does and why it's correct.

tests/logprob/test_order.py
Dhruvanshu-Joshi: These tests make a lot of sense. I have some questions regarding them. In case of argmax, the tests currently pass because ```pt.argmax``` creates a ```MaxAndArgmax``` op which is expected by the rewrite.

pymc/logprob/order.py
Dhruvanshu-Joshi: What do you mean by handling axis? The operation that we are asking to be perform is ```max``` so how does anything other than ```max``` come into the picture?

pymc/logprob/order.py
Dhruvanshu-Joshi: I first tried using the pure ```Max``` argument in the rewrite itself. But the problem is that ```pt.max``` creates the Op ```MaxAndArgmax``` instead of the expected ```Max``` and hence the rewrite is not able to find the measurable op corresponding to ```MaxAndArgmax```. I think this will not be an issue once [#334](https://github.com/pymc-devs/pytensor/issues/334) is solved.
Do you mean to write a rewrite that converts ```MaxAndArgmax``` op to ```Max``` so that ```find_measurable_max``` works with ```Max``` instead of ```MaxAndArgmax```? If so what should be done when someone uses ```pt.argmax```? Both ```pt.max``` and ```pt.argmax``` create a ```MaxAndArgmax``` ```op ``` and if we do this, we will also convert ```MaxAndArgmax``` op to ```Max`` when ```pt.argmax``` is called.

tests/logprob/test_order.py
ricardoV94: That happens because the rewrite is not currently distinguishing which of the outputs is trying to be measured. You could have checked by seeing which of the two node outputs is actually present in `preserve_rv_mappings.needs_measuring`, and fail when it's the second (argmax). Regardless, the best I think is to apply that rewrite to transform `MaxAndArgmax` into simple `Max` or `Argmax` first and then this measurable rewrite only has to worry about `Max`.

We will probably remove `MaxAndArgmax` in the future in PyTensor anyway:  https://github.com/pymc-devs/pytensor/issues/334

pymc/logprob/order.py
ricardoV94: `max` accepts `axis` just like numpy: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.max.html

In PyTensor, it's contained as a property of the `node.op`

So you have to decide which `axis` you support right now. The one where all the dimensions are collapsed is the case where `axis` contains all the elements in `range(var.ndim)`, meaning no axis remains in the output.

If you want to only support the case where all axis are reduced, you can make sure the condition is met in the rewrite where you add `MeasurableMax`, or return `None` otherwise

pymc/logprob/order.py
ricardoV94: > Do you mean to write a rewrite that converts MaxAndArgmax op to Max so that find_measurable_max works with Max instead of MaxAndArgmax? 

Yes

> If so what should be done when someone uses pt.argmax? Both pt.max and pt.argmax create a MaxAndArgmax op and if we do this, we will also convert MaxAndArgmax op to Max`` when pt.argmax``` is called

No, the rewrite that converts `Argmax` to `Max` only does so when only the first output `max` is being used and not the second `argmax`. If a user asks for a `pt.argmax`, the produced `MaxAndArgmax` will never be rewritten to a `Max`, as that would be wrong. If anything it would be replaced by a pure `Argmax` which is not a problem either.

This is checked in this line: https://github.com/pymc-devs/pytensor/blob/ec6a3153fbf5aaa023368f36aac58212c9bedbd4/pytensor/tensor/rewriting/uncanonicalize.py#L52

The number of clients is how many nodes depend on that variable. In this case you see we only replace by `Max` when the second output `Argmax` has zero clients, meaning it is never used by the user.

Dhruvanshu-Joshi: Makes sense. Thanks for the clarification 😃

pymc/logprob/order.py
```diff
@@ -0,0 +1,114 @@
...
+
+    base_rv_op = base_rv.owner.op
+    base_rv_inputs = base_rv.owner.inputs
+
+    logprob = _logprob(base_rv_op, (value,), *base_rv_inputs, **kwargs)
```
larryshamalama: I have a slight preference of calling this `logpdf` (or something along those lines) since the submodule is already called `logprob`. Not a huge deal

tests/logprob/test_order.py
```diff
@@ -0,0 +1,50 @@
...
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+#   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+#   SOFTWARE.
+
+import pytensor.tensor as pt
+
+from pymc import logp
+from pymc.testing import assert_no_rvs
+
+
```
larryshamalama, line 49: A good check would be to use the fact that the order statistics of i.i.d. uniform RVs ~ Beta:
$$U_1, \dots, U_n \stackrel{\text{i.i.d.}}{\sim} \text{Uniform}(0, 1) \Rightarrow U_{(k)} \sim \text{Beta}(k, n + 1- k)$$
for all $1 \leq k \leq n$.

You can write a test along the lines of generate i.i.d. uniform RV values and asserting the `logp`s obtained from your rewrite and the Beta distribution from the identity above correspond.

https://en.wikipedia.org/wiki/Order_statistic#Order_statistics_sampled_from_a_uniform_distribution

This check would even be applicable when working with general order statistics :)

larryshamalama (commented):
Great progress @Dhruvanshu-Joshi! Some comments to complement @ricardoV94's review

tests/logprob/test_order.py
```diff
@@ -43,8 +51,63 @@
...
     x_max_value = pt.vector("x_max_value")
-    x_max_logprob = logp(x_max[0], x_max_value)
+    x_max_logprob = logp(x_max, x_max_value)
 
     assert_no_rvs(x_max_logprob)
+
+
+def test_argmax():
+    x = pt.random.normal(0, 1, size=(3,))
+    x_name = "x"
```
larryshamalama: do you mean `x.name = "x"`? I believe that `x_name` is not assigned to any variable in your tests

tests/logprob/test_order.py
```diff
@@ -43,8 +51,63 @@
...
+    x = pm.Categorical.dist([1, 1, 1, 1], shape=(5,))
+    x_name = "x"
+    x_max = pt.max(x, axis=-1)
+    x_max_value = pt.vector("x_max_value")
+    # REASON for this failing is lines 71 - 74 and not the expected 89-91 which would pass?
+    with pytest.raises(RuntimeError, match=re.escape("Logprob method not implemented")):
+        x_max_logprob = logp(x_max, x_max_value)
+
+
+def test_max_logprob():
```
larryshamalama: With your current setup, we're looking for something along the lines of:

```python
test_value = 0.8 # or a vector of your choice

beta_rv = pt.random.beta(3, 1, name="beta")
beta_vv = beta_rv.clone()
beta_rv_logprob = logp(beta_rv, beta_vv)

assert np.isclose(
    beta_rv_logprob.eval({beta_vv: test_value}),
    x_max_logprob.eval({x_max_value: test_value})
)
```
Mathematically, the pdf of an order statistics from uniform i.i.d. RVs is equivalent to a Beta distribution, with alpha = k and beta = n - k + 1 (respectively 3 and 1 in the max with n = 3).

The code above may not be 100% right, but I think that the idea is there: the logprob of whatever test value you use should be the same whether you directly use a beta log-pdf or via your rewrite.

After this, you can perhaps extend this test to testing out multiple value of `test_value`, different values of the size argument, which you can do nicely via `pytest.mark.parametrize`, etc.

larryshamalama: In short, this is a nice test because, with uniform i.i.d. RVs, you can verify your derivation of the log-probability with a nice mathematical property.

tests/logprob/test_order.py
Dhruvanshu-Joshi: Yeah. This is a repetitive mistake. I'll rectify this.

pymc/logprob/order.py
```diff
@@ -0,0 +1,130 @@
...
+            if op.type.ndim != 0:
+                return None
+
+    if not rv_map_feature.request_measurable(node.inputs):
+        return None
+
+    axis = set(node.op.axis)
+    base_var_dims = set(range(base_var.ndim))
+    if not axis.issubset(base_var_dims):
+        return None
```
ricardoV94: You want strict equality here no? Also should check what happens with the default `axis=None` which is the same as `range(base_var.ndim)` and should be supported

pymc/logprob/order.py
```diff
@@ -0,0 +1,130 @@
...
+
+    if base_var.owner is None:
+        return None
+
+    if not isinstance(base_var.owner.op, RandomVariable):
+        return None
+
+    # univariate iid test which also rules out other distributions
+    if isinstance(base_var.owner.op, RandomVariable):
+        for op in base_var.owner.inputs[3:]:
```
ricardoV94:
```suggestion
        for param in base_var.owner.inputs[3:]:
```

And below

pymc/logprob/order.py
```diff
@@ -0,0 +1,130 @@
...
+        return None
+
+    axis = set(node.op.axis)
+    base_var_dims = set(range(base_var.ndim))
+    if not axis.issubset(base_var_dims):
+        return None
+
+    for ndim_param, param in zip(base_var.owner.op.ndims_params, base_var.owner.inputs[3:]):
+        if param.type.ndim != ndim_param:
+            return None
```
ricardoV94: As we discussed, this can be removed if you intend to use the more restrictive ndim=0 check above

pymc/logprob/order.py
```diff
@@ -0,0 +1,130 @@
...
+)
+
+
+@_logprob.register(MeasurableMax)
+def max_logprob(op, values, base_rv, **kwargs):
+    r"""Compute the log-likelihood graph for the `Max` operation.
+
+    The formula that we use here is :
+        \ln(f_{(n)}(x)) = \ln(n) + (n-1) \ln(F(x)) + \ln(f(x))
+    where f(x) represents the p.d.f and F(x) represents the c.d.f of the distrivution respectively.
```
ricardoV94:
```suggestion
    where f(x) represents the p.d.f and F(x) represents the c.d.f of the distribution respectively.
```

tests/logprob/test_order.py
```diff
@@ -0,0 +1,112 @@
...
+    x = pm.Normal.dist([0, 1, 2, 3, 4], 1, shape=(5,))
+    x.name = "x"
+    x_max = pt.max(x, axis=-1)
+    x_max_value = pt.vector("x_max_value")
+    with pytest.raises(RuntimeError, match=re.escape("Logprob method not implemented")):
+        x_max_logprob = logp(x_max, x_max_value)
+
+
+def test_max_non_rv_fails():
+    x = pt.exp(pt.random.normal(0, 1, size=(3,)))
```
ricardoV94: Better to use something that has no closed form like `exp(beta)`, because in the future we may rewrite `exp(normal)` as `lognormal` and then the rewrite would work

tests/logprob/test_order.py
```diff
@@ -0,0 +1,112 @@
...
+import pymc as pm
+
+from pymc import logp
+from pymc.testing import assert_no_rvs
+
+
+def test_max():
+    x = pt.random.normal(0, 1, size=(3,))
+    x.name = "x"
+    x_max = pt.max(x, axis=-1)
```
ricardoV94: We should test non-supported axis cases like `max(matrix, axis=-1))`. We should check `axis=None` works

pymc/logprob/order.py
```diff
@@ -0,0 +1,128 @@
...
+
+class MeasurableMax(Max):
+    """A placeholder used to specify a log-likelihood for a cmax sub-graph."""
+
+
+MeasurableVariable.register(MeasurableMax)
+
+
+@node_rewriter([Max])
+def find_measurable_max(fgraph: FunctionGraph, node: Node) -> Optional[List[MeasurableMax]]:
```
ricardoV94: The Return type hint is wrong I think. `MeasurableMax` is an `Op`, but what is returned are `TensorVariables` whose node has that `Op`

Dhruvanshu-Joshi: Many of the other ops also don't follow this. Should this be rectified for other ops too?

ricardoV94: Yes, but better to do that in a separate PR

larryshamalama: @Dhruvanshu-Joshi Would you like to open an issue about this to keep track and address at a later date?

docs/source/api/order_stats.rst
```diff
@@ -0,0 +1,41 @@
+================
+Order_Statistics
+================
+
+------------
+Introduction
+------------
+
+Users can derive the nth Order Statistic using PyMC for their custom DIstributions and the logarithmic probablity related to them.
```
larryshamalama:
```suggestion
Users can derive the nth Order Statistic using PyMC for their custom Distributions and the logarithmic probability related to them.
```
Will have to think about how to rephrase this...

pymc/logprob/order.py
```diff
@@ -0,0 +1,216 @@
...
+    MeasurableVariable,
+    _logcdf_helper,
+    _logprob,
+    _logprob_helper,
+)
+from pymc.logprob.rewriting import measurable_ir_rewrites_db
+
+
+class MeasurableMax(Max):
+    """A placeholder used to specify a log-likelihood for a cmax sub-graph."""
```
larryshamalama:
```suggestion
    """A placeholder used to specify a log-likelihood for a max sub-graph."""
```
What's cmax?

pymc/logprob/order.py
```diff
@@ -0,0 +1,216 @@
...
+    values : tensor_like
+    rv : TensorVariable
+
+    Returns
+    -------
+    logprob : TensorVariable
+
+    Examples
+    --------
+    It is often desirable to find the Maximum from the distribution of random variables.
```
larryshamalama:
```suggestion
    It is often desirable to find the log-probability of the maximum of i.i.d. random variables.
```
I think that the delineation between finding the maximum vs. finding the log-probability of the maximum of random variables are different tasks, the latter of which is your focus (and the more challenging one of the two)!

pymc/logprob/order.py
```diff
@@ -0,0 +1,216 @@
...
+        import pytensor.tensor as pt
+        from pymc import logp
+
+        x = pt.exp(pt.random.beta(0, 1, size=(3,)))
+        x.name = "x"
+        x_max = pt.max(x, axis=-1)
+        x_max_value = pt.vector("x_max_value")
+        x_max_logprob = logp(x_max, x_max_value)
+
+    The above code gives a Runtime error stating logprob method was not implemented as x in this case is a Non random variable distribution.
```
larryshamalama:
```suggestion
    The above code gives a Runtime error stating logprob method was not implemented as x in this case is not a pure random variable.
```

larryshamalama: Maybe we would need a small description of what is a "pure" random variable?

pymc/logprob/order.py
```diff
@@ -0,0 +1,216 @@
...
+
+    We only consider independent and identically distributed random variables.
```
larryshamalama:
```suggestion
    We only consider independent and identically distributed random variables, for now.
```

pymc/logprob/order.py
```diff
@@ -0,0 +1,216 @@
...
+    In probability theory and statistics, a collection of random variables is independent and identically distributed if each random variable has the same probability distribution as the others and all are mutually independent.
+    Hence the logp method fails for non-ids.
```
larryshamalama:
```suggestion
```
Redundant from the two previous sentences?

pymc/logprob/order.py
```diff
@@ -0,0 +1,216 @@
...
+
+        x = pm.Normal.dist([0, 1, 2, 3, 4], 1, shape=(5,))
+        x.name = "x"
+        x_max = pt.max(x, axis=-1)
+        x_max_value = pt.vector("x_max_value")
+        x_max_logprob = logp(x_max, x_max_value)
+
+    The above code gives a Runtime error stating logprob method was not implemented as x in this case is a Non-iid distribution.
+
+    Note: We assume a very fluid definition of iid here.We assume only univariate distributions to be iids which rejects any multivariate distribution even though it might be iid by definition.
```
larryshamalama: I would phrase this differently. 

Regarding independence, we can say that RVs do not have different stochastic ancestors. (Is this correct @ricardoV94 ?) This can also be added as a check in the rewrite

Regarding univariate vs. multivariate, we don't assume univariate RVs for computational convenience but rather the concept of ordering for multivariate variables is ambiguous since a "depth function" is required (some method to allow ranking of multivariate RVs).

larryshamalama (commented):
Great progress @Dhruvanshu-Joshi! Slight comments about the documentation on my side

pymc/logprob/order.py
Dhruvanshu-Joshi: A typing error 😅

pymc/logprob/order.py
Dhruvanshu-Joshi: Yes I completely agree. What I intended to do here is first give a gist of what a maximum operation is and how users can use this themselves before introducing them to finding logp of the maximum of random variables. Maybe I can rephrase this in a better way.

tests/logprob/test_order.py
```diff
@@ -0,0 +1,149 @@
...
+    x_max_value = pt.vector("x_max_value")
+    x_max_logprob = logp(x_max, x_max_value)
+
+    assert_no_rvs(x_max_logprob)
+
+
+def test_axis_max():
```
ricardoV94: I think it's cleaner to use `pytest.mark.parametrize` on the test above to run the same test with both axis=None and axis=-1.

tests/logprob/test_order.py
```diff
@@ -0,0 +1,149 @@
...
+        x_max_logprob = logp(x_max, x_max_value)
+
+
+def test_non_supp_axis_max():
+    """Test whether the logprob for ```pt.max``` for unsupported axis is rejected correctly"""
+    x = pt.random.normal(0, 1, size=(3, 3))
+    x.name = "x"
+    x_max_none = pt.max(x, axis=None)
+    x_max_value = pt.vector("x_max_value")
+    x_max_logprob = logp(x_max_none, x_max_value)
```
ricardoV94: This is not needed as we already test it above

docs/source/api/logprob.rst
```diff
@@ -21,3 +21,11 @@ Conditional probability
 
    conditional_logp
    transformed_conditional_logp
+
+Logarithmic probability
+-----------------------
```
ricardoV94:
```suggestion
Derived probability functions
------------------------------------------
```

docs/source/api/order_stats.rst
```diff
@@ -0,0 +1,42 @@
+================
+Order_Statistics
+================
+
+------------
+Introduction
+------------
+
+In statistics, the kth order statistic of a statistical sample is equal to its kth-smallest value.
+In this section, we'll tackle how users can find the Logarithmic probability corresponding to the nth order statistic (maximum value) using PyMC for their own Custom distributions.
```
ricardoV94: I think we should use the term log-probability

```suggestion
In this section, we'll tackle how users can find the log-probability corresponding to the nth order statistic (maximum value) using PyMC for their own Custom distributions.
```

docs/source/api/order_stats.rst
```diff
@@ -0,0 +1,42 @@
+================
+Order_Statistics
+================
+
+------------
+Introduction
+------------
+
+In statistics, the kth order statistic of a statistical sample is equal to its kth-smallest value.
```
ricardoV94: Maybe something less technical? "The kth order statistic refers to the kth value in a sorted sample.

And give some concrete examples (min, median, max) with 3 element arrays

docs/source/api/order_stats.rst
```diff
@@ -0,0 +1,42 @@
...
+In this section, we'll tackle how users can find the Logarithmic probability corresponding to the nth order statistic (maximum value) using PyMC for their own Custom distributions.
+
+In PyMC users can derive their own custom distributions. Custom distribution refers to the ability to define and use probability distributions that are not included in the standard set of distributions provided.
+While PyMC provides a wide range of common probability distributions (e.g., Normal, Bernoulli, etc.), there may be cases where you need to use a distribution that is not available by default. In such cases, you can create your own custom distribution using the pm.DensityDist class provided by PyMC.
+Simplest way to define a Custom Distribution can be better understood from the following example:
```
ricardoV94: This should be in the logprob.rst section, as it's general for all supported inferenc.

Should reference `CustomDist` directly in the text and link to the respective docs

docs/source/api/order_stats.rst
```diff
@@ -0,0 +1,42 @@
...
+
+.. code-block:: python
+
+    import numpy as np
+    import pymc as pm
+    from pytensor.tensor import TensorVariable
+
+    def logp(value: TensorVariable, mu: TensorVariable) -> TensorVariable:
+        return -(value - mu)**2
```
ricardoV94: Use an example with `dist` which is the relevant thing here. You can pick one from the `CustomDist` code snippets

docs/source/api/order_stats.rst
```diff
@@ -0,0 +1,42 @@
...
+        mu = pm.Normal('mu',0,1)
+        pm.CustomDist(
+            'custom_dist',
+            mu,
+            logp=logp,
+            observed=np.random.randn(100),
+        )
+        idata = pm.sample(100)
+
+Here, we create a CustomDist that wraps a black-box logp function. This variable cannot be used in prior or posterior predictive sampling because no random function was provided.
```
ricardoV94: This is not relevant when using `dist` instead of providing a `logp`, which is what we want users to have in mind anyway

docs/source/api/order_stats.rst
```diff
@@ -0,0 +1,42 @@
...
+
+------------------------
+`Max`
+------------------------
+Using PyMC and Pytensor, users can extract the maximum of a distribution and derive the log-probablity corresponding to this operation.
+
+.. autofunction:: pymc.logprob.order.max_logprob
```
ricardoV94: Include a `logp` example (just like in the tests) and a `CustomDist` example using `max`.

The second is something like 

```
import pymc as pm

data = ...

def max_dist(mu, sigma, size):
  return pm.math.max(pm.Normal.dist(mu, sigma, size=size))

with pm.Model() as m:
  mu = pm.Normal("mu")
  obs = pm.CustomDist("obs", mu, sigma, dist=max_dist, observed=data)
```

ricardoV94 (changes requested):
Looks good. Let's separate the docs into their own PR.

You can use an already supported case like `cumsum` instead of the `max` in that PR

pymc/logprob/order.py
```diff
@@ -0,0 +1,124 @@
...
+
+    if base_var.owner is None:
+        return None
+
+    # NonRVS must be rejected
+    if not isinstance(base_var.owner.op, RandomVariable):
+        return None
+
+    # univariate i.i.d. test which also rules out other distributions
+    if isinstance(base_var.owner.op, RandomVariable):
```
larryshamalama: perhaps just `else` here? The condition is identical to the line above. The comment can go inside the `else` statement:
```python
if not isinstance(base_var.owner.op, RandomVariable):
        # Non-RVS must be rejected
        return None
else:
       # univariate i.i.d. test which also rules out other distributions
       for params in base_var.owner.inputs[3:]:
            if params.type.ndim != 0:
                return None

File: pymc/logprob/order.py [hunk 36]
```diff
@@ -0,0 +1,124 @@
...
+
+    if isinstance(node.op, MeasurableMax):
+        return None  # pragma: no cover
+
+    base_var = node.inputs[0]
+
+    if base_var.owner is None:
+        return None
+
+    # NonRVS must be rejected
```
larryshamalama:
```suggestion
    # Non-RVs must be rejected
```

pymc/logprob/order.py
```diff
@@ -0,0 +1,124 @@
...
+        return None
+
+    # univariate i.i.d. test which also rules out other distributions
+    if isinstance(base_var.owner.op, RandomVariable):
+        for params in base_var.owner.inputs[3:]:
+            if params.type.ndim != 0:
+                return None
+
+    if not rv_map_feature.request_measurable(node.inputs):
+        return None
```
larryshamalama, line 87: Can we add a test for this?

Along the same lines: at this point, `base_var.owner.op` is necessarily a univariate `RandomVariable`. What situations would we return `None` here?

tests/logprob/test_order.py
```diff
@@ -0,0 +1,148 @@
...
+    x = pm.Normal.dist([0, 1, 2, 3, 4], 1, shape=(5,))
+    x.name = "x"
+    x_max = pt.max(x, axis=-1)
+    x_max_value = pt.vector("x_max_value")
+    with pytest.raises(RuntimeError, match=re.escape("Logprob method not implemented")):
+        x_max_logprob = logp(x_max, x_max_value)
+
+
+def test_max_non_rv_fails():
+    """Test whether the logprob for ```pt.max``` for non RVs is rejected correctly"""
```
larryshamalama:
```suggestion
    """Test whether the logprob for ```pt.max``` for non-RVs is correctly rejected"""
```

tests/logprob/test_order.py
```diff
@@ -0,0 +1,148 @@
...
+        U_1, \\dots, U_n \\stackrel{\text{i.i.d.}}{\\sim} \text{Uniform}(0, 1) \\Rightarrow U_{(k)} \\sim \text{Beta}(k, n + 1- k)
+    for all 1<=k<=n
+    """
+    x = pt.random.uniform(0, 1, size=(3,))
+    x.name = "x"
+    x_max = pt.max(x, axis=-1)
+    x_max_value = pt.scalar("x_max_value")
+    x_max_logprob = logp(x_max, x_max_value)
+
+    test_value = 0.85  # or a vector of your choice
```
larryshamalama:
```suggestion
    test_value = 0.85
```

tests/logprob/test_order.py
```diff
@@ -0,0 +1,148 @@
...
+    """Test whether the logprob for ```pt.max``` for unsupported axis is rejected correctly"""
+    x = pt.random.normal(0, 1, size=(3, 3))
+    x.name = "x"
+    x_max = pt.max(x, axis=-1)
+    x_max_value = pt.vector("x_max_value")
+    with pytest.raises(RuntimeError, match=re.escape("Logprob method not implemented")):
+        x_max_logprob = logp(x_max, x_max_value)
+
+
+def test_max_logprob():
```
larryshamalama: - Perhaps parametrize this for several `test_value`s? Something like 0.01, 0.2, 0.5, 0.85, 0.9. It wouldn't hurt to check for more values.
- You can also parametrize the `n` here (which has a value of 3). This would be more important than the comment above.

tests/logprob/test_order.py
```diff
@@ -0,0 +1,148 @@
...
+import pytensor.tensor as pt
+import pytest
+
+import pymc as pm
+
+from pymc import logp
+from pymc.testing import assert_no_rvs
+
+
+def test_max():
```
larryshamalama: I think this test is redundant with `test_max_logprob` below, is it? If so, we can slide `assert_no_rvs(x_max_logprob)` into the last test

tests/logprob/test_order.py
```diff
@@ -0,0 +1,148 @@
...
+    x.name = "x"
+    x_max = pt.argmax(x, axis=-1)
+    x_max_value = pt.vector("x_max_value")
+
+    with pytest.raises(RuntimeError, match=re.escape("Logprob method not implemented for Argmax")):
+        x_max_logprob = logp(x_max, x_max_value)
+
+
+def test_max_non_iid_fails():
+    """Test whether the logprob for ```pt.max``` for non i.i.d is rejected correctly"""
```
larryshamalama:
```suggestion
    """Test whether the logprob for ```pt.max``` for non i.i.d is correctly rejected"""
```

tests/logprob/test_order.py
```diff
@@ -0,0 +1,148 @@
...
+    x = pt.exp(pt.random.beta(0, 1, size=(3,)))
+    x.name = "x"
+    x_max = pt.max(x, axis=-1)
+    x_max_value = pt.vector("x_max_value")
+    with pytest.raises(RuntimeError, match=re.escape("Logprob method not implemented")):
+        x_max_logprob = logp(x_max, x_max_value)
+
+
+def test_max_categorical():
+    """Test whether the logprob for ```pt.max``` for unsupported distributions is rejected correctly"""
```
larryshamalama:
```suggestion
    """Test whether the logprob for ```pt.max``` for unsupported distributions is correctly rejected"""
```

tests/logprob/test_order.py
```diff
@@ -0,0 +1,148 @@
...
+    """Test whether the rewrite takes into account ```None``` axis"""
+    x_max = pt.max(x, axis)
+    x_max_value = pt.vector("x_max_value")
+    x_max_logprob = logp(x_max, x_max_value)
+
+    assert_no_rvs(x_max_logprob)
+
+
+def test_argmax():
+    """Test whether the logprob for ```pt.argmax``` is rejected correctly"""
```
larryshamalama:
```suggestion
    """Test whether the logprob for ```pt.argmax``` is correctly rejected"""
```

tests/logprob/test_order.py
```diff
@@ -0,0 +1,148 @@
...
+    x = pm.Categorical.dist([1, 1, 1, 1], shape=(5,))
+    x.name = "x"
+    x_max = pt.max(x, axis=-1)
+    x_max_value = pt.vector("x_max_value")
+    with pytest.raises(RuntimeError, match=re.escape("Logprob method not implemented")):
+        x_max_logprob = logp(x_max, x_max_value)
+
+
+def test_non_supp_axis_max():
+    """Test whether the logprob for ```pt.max``` for unsupported axis is rejected correctly"""
```
larryshamalama:
```suggestion
    """Test whether the logprob for ```pt.max``` for unsupported axis is correctly rejected"""
```

larryshamalama (commented):
Some comments in preparation for tomorrow's meeting :) This PR is shaping up very nicely, we're close to merging!

pymc/logprob/order.py
larryshamalama: Or just remove the second condition altogether since returning `None` would exit this function:
```python
if not isinstance(base_var.owner.op, RandomVariable):
        # Non-RVS must be rejected
        return None

 # univariate i.i.d. test which also rules out other distributions
 for params in base_var.owner.inputs[3:]:
      if params.type.ndim != 0:
          return None
```
This now becomes a Python style question...

tests/logprob/test_order.py
Dhruvanshu-Joshi: Yeah that makes sense. This test was mostly used during the initial phase to test if the logprob was succesfully implemented or not.

pymc/logprob/order.py
```diff
@@ -0,0 +1,127 @@
...
+    base_var = node.inputs[0]
+
+    if base_var.owner is None:
+        return None
+
+    if not rv_map_feature.request_measurable(node.inputs):
+        return None
+
+    # Non-RVS must be rejected
+    if not isinstance(base_var.owner.op, RandomVariable):
```
ricardoV94:
```suggestion
    # Non-univariate non-RVs must be rejected
    if not (isinstance(base_var.owner.op, RandomVariable) and base_var.owner.op.ndim_supp==0):
```

ricardoV94 (commented):
Everything looks pretty good, just a small check missing I think.

pymc/logprob/order.py
```diff
@@ -0,0 +1,130 @@
...
+
+    if base_var.owner is None:
+        return None
+
+    if not rv_map_feature.request_measurable(node.inputs):
+        return None
+
+    # Non-univariate non-RVs must be rejected
+    if (
+        not isinstance(base_var.owner.op, RandomVariable)
```
ricardoV94: The not should cover both cases of the `and`

pymc/logprob/order.py
```diff
@@ -81,12 +81,12 @@ def find_measurable_max(fgraph: FunctionGraph, node: Node) -> Optional[List[Tens
     # Non-univariate non-RVs must be rejected
     if (
         not isinstance(base_var.owner.op, RandomVariable)
-        and base_var.owner.inputs[0].owner.op.ndim_supp == 0
+        and base_var.owner.inputs[0].owner.op.ndim_supp != 0
```
Dhruvanshu-Joshi: On changing both to not, our test fails to reject non-rvs such as ``` x = pt.exp(pt.random.beta(0, 1, size=(3,))) ``` as the ndim_supp in this case comes out to be equal to zero, hence the None statement is never encountered.

ricardoV94: I don't remember my original suggestion but the idea was `if not (rv and ndim_supp==0): return None`

Dhruvanshu-Joshi: Originally I think we planned to reject any Non-RVs.
By this logic, ```x = pt.exp(pt.random.beta(0, 1, size=(3,)))``` does not fail as though exp (being elemwise) is a non rv, a beta distribution with size = (3, ) does not meet the above condition. Should I update the test to something like ```x = pt.exp(pt.random.beta(0, 1, size=(3, 3)))``` so that it meets the condition?

ricardoV94: Oh I was not clear, the point was to **also** reject multivariate RVs which I think the rewrite doesn't reject. For instance StickBreakingWeights is a multivariate RV with scalar inputs.

`if not (rv and univariate): reject`

ricardoV94 (dismissed):

larryshamalama:
Remind me what's pending for merging this PR?

ricardoV94:
I think it just needs to be rebased so the tests pass. There was an unrelated failure that crop up elsewhere

ricardoV94 (approved):

PR #6772, opened 2023-06-12
Simplify Potential docstrings and examples

ricardoV94:
This PR simplifies the examples and explanations used in the Potential docstrings

----
:books: Documentation preview :books:: https://pymc--6772.org.readthedocs.build/en/6772/

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -2199,15 +2199,14 @@ def Deterministic(name, var, model=None, dims=None):
     return var
 
 
-def Potential(name, var, model=None, dims=None):
-    """
-    Add an arbitrary factor potential to the model likelihood
-
-    The Potential function is used to add arbitrary factors (such as constraints or other likelihood components) to adjust the probability density of the model.
+def Potential(name, var: TensorVariable, model=None, dims=None) -> TensorVariable:
+    """Add an arbitrary term to the model log-probability.
 
     Warnings
     --------
-    Potential functions only influence logp-based sampling. Therefore, they are applicable for sampling with ``pm.sample`` but not ``pm.sample_prior_predictive`` or ``pm.sample_posterior_predictive``.
+    Potential terms only influence probability-based sampling, such as ``pm.sample``, but not forward sampling like
+    ``pm.sample_prior_predictive`` or ``pm.sample_posterior_predictive``. A warning is raised when doing forward
+    sampling with models containing Potential terms.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


daniel-saunders-phil:
Hey Ricardo, is the first example is supposed to be executable?

```
data = stats.norm(0.05,1).rvs(50)

with pm.Model() as m8:
    x = pm.Normal("x", mu=0, sigma=1)
    y = pm.Normal("y", mu=x, sigma=1, observed=data)
    constraint = x >= 0
    potential = pm.Potential("x_constraint", pm.math.log(pm.math.switch(constraint, 1, 0)))
    
    t8 = pm.sample()

```

I get `AttributeError: 'float' object has no attribute 'type'` . Full error trace back: https://gist.github.com/daniel-saunders-phil/b6656f51b0fae351274711dcfb2641f2 . The other examples under Potential work fine. Also, if I change the switch from a hard constraint to a very firm constraint (`pm.math.switch(constraint, 1, 0.1)`), the example runs (with poor sampler performance). So it looks like the problem is unique to adding log of 0 to the potential.

ricardoV94:
I think this was fixed in the very last release of PyTensor (1 or 2 days old) and I was probably already on it.

Can you update and check again?

ricardoV94:
By the way I think you're looking at the old example (not that it matters for the error your stumbled upon)

Also these model are not expected to sample well, it's just for illustration. Discontinuities like this are pretty bad for NUTS

twiecki (approved):

daniel-saunders-phil:
> I think this was fixed in the very last release of PyTensor (1 or 2 days old) and I was probably already on it.
> 
> Can you update and check again?

Ahh okay you're on top of it. Doesn't look the latest version is available through conda-forge at the moment. I can only get 2.12.1. The new examples are also much nicer to read btw

pymc/model.py
```diff
@@ -2228,62 +2227,80 @@ def Potential(name, var, model=None, dims=None):
...
-        with pm.Model():
+        import pymc as pm
+
+        with pm.Model() as model:
+            x = pm.Normal("x", mu=0, sigma=10)
+            y = pm.Normal("y", mu=0, sigma=10)
+            soft_sum_constraint = pm.Potential("soft_sum_constraint", -((x + y)**2))
+
+    A Potential can be used to define a specific prior term.
+    The following example imposes a prior law prior on `max_items`, under the form ``log(1/max_items)``,
```
bwengals, line 2272: prior law prior -> power law prior

bwengals:
```suggestion
    The following example imposes a power law prior on `max_items`, under the form ``log(1/max_items)``,
```

PR #6793, opened 2023-06-23
NotImplementedError for icdf of non-injective MeasurableTransforms

ricardoV94:
We were already doing it for `logcdf` but not `icdf`. Also added direct tests.

----
:books: Documentation preview :books:: https://pymc--6793.org.readthedocs.build/en/6793/

### DIFF
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -448,7 +448,8 @@ def measurable_transform_logcdf(op: MeasurableTransform, value, *inputs, **kwarg
 
     backward_value = op.transform_elemwise.backward(value, *other_inputs)
 
-    # Some transformations, like squaring may produce multiple backward values
+    # Fail if transformation is not injective
+    # A TensorVariable is returned in 1-to-1 inversions, and a tuple in 1-to-many
     if isinstance(backward_value, tuple):
         raise NotImplementedError
 
@@ -468,6 +469,11 @@ def measurable_transform_icdf(op: MeasurableTransform, value, *inputs, **kwargs)
 
     input_icdf = _icdf_helper(measurable_input, value)
     icdf = op.transform_elemwise.forward(input_icdf, *other_inputs)
+
+    # Fail if transformation is not injective
+    # A TensorVariable is returned in 1-to-1 inversions, and a tuple in 1-to-many
+    if isinstance(op.transform_elemwise.backward(icdf, *other_inputs), tuple):
+        raise NotImplementedError
 
     return icdf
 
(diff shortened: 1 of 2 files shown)


tests/logprob/test_transforms.py
```diff
@@ -1080,3 +1080,37 @@ def test_check_jac_det(transform):
...
+    x = pt.exp(pt.random.uniform(0, 1))
+    value = x.type()
+    icdf_fn = pytensor.function([value], icdf(x, value))
+
+    np.testing.assert_almost_equal(icdf_fn(1e-16), 1)
+    np.testing.assert_almost_equal(icdf_fn(0.5), np.exp(0.5))
+    np.testing.assert_almost_equal(icdf_fn(1 - 1e-16), np.e)
+
+
+def test_icdf_measurable_non_injective_fails():
```
larryshamalama, line 1112: Can this test be combined with `test_logcdf_measurable_non_injective_fails`?

pymc/logprob/transforms.py
```diff
@@ -469,6 +469,10 @@ def measurable_transform_icdf(op: MeasurableTransform, value, *inputs, **kwargs)
     input_icdf = _icdf_helper(measurable_input, value)
     icdf = op.transform_elemwise.forward(input_icdf, *other_inputs)
 
+    # Fail if transformation is not injective
```
larryshamalama, line 473: Curious: why does `tuple` imply an injective transform here?

larryshamalama (approved):
Looks great, just some quick questions :)

Preemptively approving this because I can take a while to respond

tests/logprob/test_transforms.py
ricardoV94: We could, but I think this flows more nicely with the supported logcdf/icdf check done before

pymc/logprob/transforms.py
ricardoV94: It was just a choice when I first implemented it. It returns a TensorVariable when it's a 1-to-1 inversion and a tuple with several TensorVariables when it's a 1-to-many case. Seemed like a natural way of going about it.

tests/logprob/test_transforms.py
larryshamalama: Okay, I trust your judgment more here :)

pymc/logprob/transforms.py
larryshamalama: Can you add this as a comment? Perhaps something like:
```suggestion
    # Fail if transformation is not injective. A TensorVariable is returned in a 1-to-1 inversion,
    # whereas a tuple is returned in a 1-to-many case.
```

ricardoV94: Sounds good

ricardoV94: Done

PR #6796, opened 2023-06-26
Allow non-scalar measurable switch mixtures

ricardoV94:
This PR expands the supported Mixture graphs beyond the scalar case. 

It comes at the cost of not creating a lazy evaluation logp graph that the old MixtureRV approach did, but benefits from allowing arbitrary measurable variables and not just pure RVs.

In a future PR I intend to introduce a lazy Mixture RV that brings back that functionality, but that's right now blocked by https://github.com/pymc-devs/pytensor/pull/329

CC @shreyas3156 

----
:books: Documentation preview :books:: https://pymc--6796.org.readthedocs.build/en/6796/

### DIFF
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -44,6 +44,8 @@ from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.op import Op, compute_test_value
 from pytensor.graph.rewriting.basic import node_rewriter, pre_greedy_node_rewriter
 from pytensor.ifelse import IfElse, ifelse
+from pytensor.scalar import Switch
+from pytensor.scalar import switch as scalar_switch
 from pytensor.tensor.basic import Join, MakeVector, switch
 from pytensor.tensor.random.rewriting import (
     local_dimshuffle_rv_lift,
@@ -55,7 +57,6 @@ from pytensor.tensor.subtensor import (
     AdvancedSubtensor,
     AdvancedSubtensor1,
     as_index_literal,
-    as_nontensor_scalar,
     get_canonical_form_slice,
     is_basic_idx,
 )
@@ -63,7 +64,12 @@ from pytensor.tensor.type import TensorType
 from pytensor.tensor.type_other import NoneConst, NoneTypeT, SliceConstant, SliceType
 from pytensor.tensor.var import TensorVariable
 
-from pymc.logprob.abstract import MeasurableVariable, _logprob, _logprob_helper
+from pymc.logprob.abstract import (
+    MeasurableElemwise,
+    MeasurableVariable,
+    _logprob,
+    _logprob_helper,
+)
 from pymc.logprob.rewriting import (
     PreserveRVMappings,
     assume_measured_ir_outputs,
@@ -325,37 +331,6 @@ def find_measurable_index_mixture(fgraph, node):
     return [new_mixture_rv]
 
 
-@node_rewriter([switch])
-def find_measurable_switch_mixture(fgraph, node):
-    rv_map_feature: Optional[PreserveRVMappings] = getattr(fgraph, "preserve_rv_mappings", None)
-
-    if rv_map_feature is None:
-        return None  # pragma: no cover
-
-    old_mixture_rv = node.default_output()
-    idx, *components = node.inputs
-
-    if rv_map_feature.request_measurable(components) != components:
-        return None
-
-    mix_op = MixtureRV(
-        2,
-        old_mixture_rv.dtype,
-        old_mixture_rv.broadcastable,
-    )
-    new_mixture_rv = mix_op.make_node(
-        *([NoneConst, as_nontensor_scalar(node.inputs[0])] + components[::-1])
-    ).default_output()
-
-    if pytensor.config.compute_test_value != "off":
-        if not hasattr(old_mixture_rv.tag, "test_value"):
-            compute_test_value(node)
-
-        new_mixture_rv.tag.test_value = old_mixture_rv.tag.test_value
-
-    return [new_mixture_rv]
-
-
 @_logprob.register(MixtureRV)
 def logprob_MixtureRV(
     op, values, *inputs: Optional[Union[TensorVariable, slice]], name=None, **kwargs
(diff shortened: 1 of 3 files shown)


pymc/logprob/mixture.py
```diff
@@ -433,6 +408,53 @@ def logprob_MixtureRV(
...
+@node_rewriter([switch])
+def find_measurable_switch_mixture(fgraph, node):
+    rv_map_feature: Optional[PreserveRVMappings] = getattr(fgraph, "preserve_rv_mappings", None)
+
+    if rv_map_feature is None:
+        return None  # pragma: no cover
+
+    switch_cond, *components = node.inputs
+
+    # We don't support broadcasting of components (as those values are not independent).
```
larryshamalama: Do you mind expanding on this comment? Do you mean that broadcasted components yields dependent (i.e. identical) values within components?

pymc/logprob/mixture.py
```diff
@@ -433,6 +408,53 @@ def logprob_MixtureRV(
...
+    # Broadcasting of the switch condition is fine
+    out_bcast = node.outputs[0].type.broadcastable
+    if any(comp.type.broadcastable != out_bcast for comp in components):
+        return None
+
+    # Check that `switch_cond` is not potentially measurable
+    valued_rvs = rv_map_feature.rv_values.keys()
+    if check_potential_measurability([switch_cond], valued_rvs):
+        return None
```
larryshamalama, line 437: Shall we add a quick test for the `if` block above?

tests/logprob/test_mixture.py
```diff
@@ -927,40 +929,73 @@ def test_switch_mixture():
...
+        np.where(
+            test_switch_cond,
+            logp(true_branch, test_switch_value).eval(),
+            logp(false_branch, test_switch_value).eval(),
+        ),
+    )
+
+
+def test_switch_mixture_invalid_bcast():
+    """Test that we don't mark switches where components are broadcsated as measurable"""
```
larryshamalama:
```suggestion
    """Test that we don't mark switches where components are broadcasted as measurable"""
```

tests/logprob/test_mixture.py
```diff
@@ -927,40 +929,73 @@ def test_switch_mixture():
...
+    valid_switch_cond = pt.vector("switch_cond", dtype=bool)
+    invalid_switch_cond = pt.matrix("switch_cond", dtype=bool)
+
+    valid_true_branch = pt.exp(pt.random.normal(size=(4,)))
+    valid_false_branch = pt.abs(pt.random.normal(size=(4,)))
+    invalid_false_branch = pt.abs(pt.random.normal(size=()))
+
+    valid_mix = pt.switch(valid_switch_cond, valid_true_branch, valid_false_branch)
+    fgraph, _, _ = construct_ir_fgraph({valid_mix: valid_mix.type()})
+    assert isinstance(fgraph.outputs[0].owner.op, MeasurableVariable)
```
larryshamalama, line 1004: More precisely a `MeasurableSwitchMixture`? What is the benefit of checking the broader category of `MeasurableVariable` here?

tests/logprob/test_mixture.py
```diff
@@ -927,40 +929,73 @@ def test_switch_mixture():
...
+
+    invalid_mix = pt.switch(invalid_switch_cond, valid_true_branch, valid_false_branch)
+    fgraph, _, _ = construct_ir_fgraph({invalid_mix: invalid_mix.type()})
+    assert not isinstance(fgraph.outputs[0].owner.op, MeasurableVariable)
+
+    invalid_mix = pt.switch(valid_switch_cond, valid_true_branch, invalid_false_branch)
+    fgraph, _, _ = construct_ir_fgraph({invalid_mix: invalid_mix.type()})
+    assert not isinstance(fgraph.outputs[0].owner.op, MeasurableVariable)
```
larryshamalama, line 1013: Would it be more clear to rename the `fgraph`s and `invalid_mix`es in this test?

larryshamalama (commented):
As always, looks great @ricardoV94 😅 As usual, some questions from my side

pymc/logprob/mixture.py
ricardoV94: Yes. `switch(idx, normal(), normal(size=(5,))` broadcasts (copies) the entries of the scalar branch in the forward grah. So the correct logp would need to not overcount those / return 0 when they are different (since that's not possible). Because the logp does not do that, we fail the rewrite. I'll make it more explicit in the comment

Related to https://github.com/pymc-devs/pymc/issues/6398

pymc/logprob/mixture.py
ricardoV94: Yeah, totally!

tests/logprob/test_mixture.py
ricardoV94: Yeah, that makes sense

tests/logprob/test_mixture.py
ricardoV94: sure!

pymc/logprob/mixture.py
ricardoV94: Expanded comment

pymc/logprob/mixture.py
ricardoV94: Added

tests/logprob/test_mixture.py
ricardoV94: On a second thought, I think this makes sense as it's showing the opposite behavior when you have invalid mixtures.

tests/logprob/test_mixture.py
ricardoV94: On a second thought... rename to what? I feel the names are pretty reasonable?

tests/logprob/test_mixture.py
larryshamalama: Okay I see. I was thinking of checking valid mixtures with `MeasurableSwitchMixture` and invalid with what you have below. Not a big deal at all, was just wondering if we could be more precise

tests/logprob/test_mixture.py
larryshamalama: Ah, I was just alluding to the redundancy in names (three presences of `fgraph`, two of `invalid_mix`).

larryshamalama (approved):

pymc/logprob/mixture.py
```diff
@@ -433,6 +408,54 @@ def logprob_MixtureRV(
     return logp_val
 
 
+class MeasurableSwitchMixture(MeasurableElemwise):
+    valid_scalar_types = (Switch,)
+
+
+MeasurableVariable.register(MeasurableSwitchMixture)
```
ricardoV94:
```suggestion
```
Not needed?

tests/logprob/test_mixture.py
ricardoV94: I check both now

Issue #6626, opened 2023-03-29
BUG: automatic data imputation does not work when observed=pm.Data() tensors


kamicollo:
### Describe the issue:

Automatic imputation fails silently in pyMC if a user passes partially observed data held in `pm.ConstantData()` or `pm.MutableData()`  to `observed` parameter of any distribution. In simple models, the user won't be able to sample (as loglik will evaluate to nan), but I have also been able to run more complex (GP) models that sampled - likely producing wrong results. (see https://discourse.pymc.io/t/issue-imputing-data-for-gaussian-process-model/11626/3 for detail).

Based on my initial review of the source code, it seems the culprit is `Model.make_obs_var()` method, where the check whether passed data is performed with  `mask = getattr(data, "mask", None)`, which always returns None for tensors.

In case of  `pm.ConstantData()`, the fix appears to be quite simple (need to retrieve masked values by `mask = getattr(data.value, "mask", None)` instead. In case of `pm.MutableData()`, however, the issue seems to be that `pytensor.shared()` does not maintain masked values. That is very problematic on its own if masked values are represented by actual numbers and not `np.nan`. I'll file an issue under pytensor project about this, too.

I'd be happy to contribute a PR for `pm.ConstantData()` fix + possibly a `NotImplemented` error for `pm.MutableData()` if this indeed cannot be solved in other ways. I'm new to pyMC code base and may be missing the big picture!

### Reproduceable code example:

```python
import pymc as pm
import numpy as np

real_X = np.random.default_rng().normal(size=1000)
Y = np.random.default_rng().normal(loc=3 * real_X, scale=0.1)
X = real_X.copy()
X[0:10] = np.nan
masked_X = np.ma.masked_where(np.isnan(X), X)

with pm.Model() as m:

    β = pm.Normal("β", 0, 1)
    σ = pm.Exponential("σ", 1)

    # This works
    # X = pm.Normal("X", 0, 1, observed = masked_X, dims='test')   
    
    #T his even fails to sample 
    X = pm.Normal("X", 0, 1, observed = pm.ConstantData("masked_X", masked_X))    
    
    pm.Normal("Y", pm.math.dot(X, β), σ, observed=Y) 

    pm.sample()
```

### PyMC version information:

pymc 5.1.2
pytensor 2.10.1

### Context for the issue:

The fact that it fails silently on some models is particularly concerning - it means some users may be using pyMC and getting wrong inference results.

ricardoV94:
[...]

Issue #5255, opened 2021-12-12
Compute correct log_likelihood for models with partially observed variables


ricardoV94:
Masked observations are no longer associated with the observed RV, but to a separate free RV. 

```python
import numpy as np
import pymc as pm

with pm.Model() as m:
  x = pm.Normal('x', observed=[np.nan, 1, 2, 3])

print(m['x_observed'].tag.observations)  # TensorConstant{[1.0 2.0 3.0]}
```

As such I am not sure whether we are computing the correcting model `log_likelihood` for partially observed models, **assuming the imputed observations should appear in the final log_likelihood**:

Running all tests in `test_idata_conversion.py` with coverage, confirmed that these lines lines in `InferenceDataConverter.log_likelihood-vals_point` are not being triggered:

https://github.com/pymc-devs/pymc/blob/0c90e8204fcf4040edca62a490e2f5aee982f96a/pymc/backends/arviz.py#L248-L257

This came up in #5245

ricardoV94:
Anyone has a clear source (or knowledge) of whether imputed variables should be considered during model comparison (and as such be part of the InferenceData log-likelihood field)? 

In V3 they were, now they are not.

twiecki:
CC @junpenglao @fonnesbeck @ColCarroll

OriolAbril:
My hunch is that neither option is correct, but with loo it's relatively easy to check it out by comparing psis-loo results (via az.loo) to "brute force" cross validation (which can be partly automated with az.reloo).

I lean towards neither because both waic and loo are designed for model comparison based on predictive accuracy but when performing inputation, part of the prediction task is "included" in the model and affects the logp. IMO using waic/loo for model *-comparison with missing data should be fit the model on the observed data, then loo/waic are an estimation of how well the model should be able to predict the missing data and you can use that to to compare models. The first and most important question when using loo/waic is defining what is actually the "predictive task".

Disclaimer: I might be a bit biased against imputation. From what I said above, one could also argue that the model with data imputation is taking into account some extra information (as if one used better priors) and therefore both could be correct depending on the goal. 

What would make some more sense to me would be to exclude the missing/imputed data from the pointwise log likelihood data (but not from the fit) and then compare the model that excludes the missing data from the fit with the model with imputed data. Now both would be using the same data to estimate their predictive accuracy and can be compared. 

[...]

PR #6797, opened 2023-06-27
Support automatic imputation for multivariate and symbolic distributions

### LINKED ISSUE #5260: Partially observed multivariate distributions not implemented in V4
I added an explicit `NotImplementedError` and future test in #5245 - https://github.com/pymc-devs/pymc/pull/5245/commits/022cefa3e6b64f73a8e91f927c4db3d7454c4418

Partial observed variables now rely on Aesara's `local_subtensor_rv_lift`, to separate the observed and unobserved portions of a RV into two separate RVs. 

https://github.com/aesara-devs/aesara/blob/4e1077210721deb007b4f556f08702cec5a74dfb/aesara/tensor/random/opt.py#L259

This rewrite needs to be updated to work for Multivariate distributions, there are a couple of comments in its code suggesting how to start approaching such task. If it does not already exist, we should open an issue in Aesara as those code changes would go there.

Importantly, even after this rewrite, we won't have the same flexibility we had in V3, where we might be missing some portions of a the base multidimensional value. Let's call these "sub-values", such as in `pm.MvNormal("x", np.ones(2), np.eye(2), observed=np.array([[1, 1], [1, np.nan]])`, because the missing portion of the distribution does not correspond to a `MvNormal` distribution. In contrast, `pm.MvNormal("x", np.ones(2), np.eye(2), observed=np.array([[1, np.nan], [1, np.nan]])` should work just fine, as we can split that into two `MvNormal`s. 

On the bright side, we would retain the ability to transform imputed values which was not the case with V3, and which would not be trivial if we were to somehow allow for "sub-value" imputation, because we don't know what are the constraints on those "sub-values".

One different issue, concerns partial observations for aeppl "derived" distributions such as in #5169. These won't work out of the box, because `local_subtensor_rv_lift` is meant to work only with pure `RandomVariables`. In theory we could make such rewrite (or a copycat one) work with these "derived distributions", as long as we can provide the value and parameters support/mapping information that the rewrite requires, as well as the ability to re-create a node on the spot. In #5169 I had to implement most of these features anyway, but kept them accessible only via the `cls` object. These can easily be made more accessible via class properties or dispatching.

CC @brandonwillard

ricardoV94:
Closes #5260 
Related to #6626
Related to #5255
Related to #6645

This PR creates a new Op: `PartialObservedRV` that splits the sample space according to a boolean mask and allows separate variables/values for these. This enables automatic imputation for cases not supported before. For multivariate cases it is not always possible to "attribute" the logp to one variable or the other, so they are all associated with the observed variable. This means the values will show up exclusively in the model `log_likelihood`, even if some (or all) entries were imputed. Directly related to: https://github.com/pymc-devs/pymc/issues/5255

There is some logic to avoid the use of the `PartialObservedRV` for pure Multivariate variables when it's safe to do so (i.e., there is no mixed indexing across the support dims). This has the benefit that automatic transforms will still apply. It also avoids the logp issue mentioned above. This behavior is the same that existed until now for univariate pure RandomVariables, which keep behaving as before.

The new `PartialObservedRV` allows for symbolic (constant or mutable) mask, but this is not accessible to users using the current API based on `MaskedArray` or `nan` entries. Similarly, one can't use `ConstantData`/`MutableData` because those try to convert such arrays to `MaskedArray` which aren't supported in PyTensor: https://github.com/pymc-devs/pytensor/issues/259. I added an early error for that.

An alternative would be to provide a separate API for users, something like `pm.Imputed(name, dist, obs_mask, obs_data)` where both data and the mask could be tensors wrapped in `Data`.

----
:books: Documentation preview :books:: https://pymc--6797.org.readthedocs.build/en/6797/

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -423,6 +423,11 @@ def Data(
     # `convert_observed_data` takes care of parameter `value` and
     # transforms it to something digestible for PyTensor.
     arr = convert_observed_data(value)
+    if isinstance(arr, np.ma.MaskedArray):
+        raise NotImplementedError(
+            "Masked arrays or arrays with `nan` entries are not supported. "
+            "Pass them directly to `observed` if you want to trigger auto-imputation"
+        )
 
     if mutable is None:
         warnings.warn(
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -25,13 +25,14 @@ import numpy as np
 
 from pytensor import tensor as pt
 from pytensor.compile.builders import OpFromGraph
-from pytensor.graph import node_rewriter
+from pytensor.graph import FunctionGraph, node_rewriter
 from pytensor.graph.basic import Node, Variable
 from pytensor.graph.replace import clone_replace
 from pytensor.graph.rewriting.basic import in2out
 from pytensor.graph.utils import MetaType
 from pytensor.tensor.basic import as_tensor_variable
 from pytensor.tensor.random.op import RandomVariable
+from pytensor.tensor.random.rewriting import local_subtensor_rv_lift
 from pytensor.tensor.random.utils import normalize_size_param
 from pytensor.tensor.var import TensorVariable
 from typing_extensions import TypeAlias
@@ -49,6 +50,7 @@ from pymc.distributions.shape_utils import (
 )
 from pymc.exceptions import BlockModelAccessError
 from pymc.logprob.abstract import MeasurableVariable, _icdf, _logcdf, _logprob
+from pymc.logprob.basic import logp
 from pymc.logprob.rewriting import logprob_rewrites_db
 from pymc.model import BlockModelAccess
 from pymc.printing import str_for_dist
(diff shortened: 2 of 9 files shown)


pymc/data.py
```diff
@@ -423,6 +423,11 @@ def Data(
     # `convert_observed_data` takes care of parameter `value` and
     # transforms it to something digestible for PyTensor.
     arr = convert_observed_data(value)
+    if isinstance(arr, np.ma.MaskedArray):
```
ricardoV94, line 426: This would have failed in the call to `as_tensor_variable` or `shared` anyway

pymc/distributions/distribution.py
```diff
@@ -1148,3 +1150,145 @@ def logcdf(value, c):
...
+        Constant or variable boolean mask. True entries correspond to components of the variable that are not observed.
+
+    Returns
+    -------
+    observed_rv and mask : Tuple of TensorVariable
+        The observed component of the RV and respective indexing mask
+    unobserved_rv and mask: Tuple of TensorVariable
+        The unobserved component of the RV and respective indexing mask
+    joined_rv:
+        The symbolic join of the observed and unobserved components.
```
```suggestion
    unobserved_rv and mask : Tuple of TensorVariable
        The unobserved component of the RV and respective indexing mask
    joined_rv
        The symbolic join of the observed and unobserved components.
```

pymc/model.py
```diff
@@ -1409,67 +1407,27 @@ def make_obs_var(
...
-            )
+            # Register ObservedRV corresponding to observed component
+            observed_rv.name = f"{name}_observed"
+            self.create_value_var(observed_rv, transform=None, value_var=observed_data)
+            self.add_named_variable(observed_rv)
+            self.observed_RVs.append(observed_rv)
 
-            (missing_rv_var,) = local_subtensor_rv_lift.transform(fgraph, fgraph.outputs[0].owner)
+            # Register FreeRV corresponding to unobserved components
+            self.register_rv(unobserved_rv, f"{name}_missing", transform=transform)
```
```suggestion
            self.register_rv(unobserved_rv, f"{name}_unobserved", transform=transform)
```
this would match the `_observed` for the other and maybe avoid confusion?

tests/backends/test_arviz.py
```diff
@@ -361,19 +360,25 @@ def test_mv_missing_data_model(self):
...
             # pylint: disable=unpacking-non-sequence
-            chol, *_ = pm.LKJCholeskyCov("chol_cov", n=2, eta=1, sd_dist=sd_dist, compute_corr=True)
+            chol, *_ = pm.LKJCholeskyCov("chol_cov", n=2, eta=1, sd_dist=sd_dist)
             # pylint: enable=unpacking-non-sequence
             with pytest.warns(ImputationWarning):
                 y = pm.MvNormal("y", mu=mu, chol=chol, observed=data)
-            inference_data = pm.sample(100, chains=2, return_inferencedata=True)
+            inference_data = pm.sample(
+                tune=100,
+                draws=100,
```
michaelosthege:
```suggestion
                tune=5,
                draws=7,
```

tests/distributions/test_distribution.py
```diff
@@ -700,3 +702,217 @@ def test_dtype(self, floatX):
...
+
+    @pytest.mark.parametrize(
+        "mask",
+        [
+            pt.constant(np.array([[True, True, True, True]])),
+            pt.constant(np.array([[False, False, False, False]])),
+        ],
+    )
+    def test_multivariate_constant_mask_separable(self, mask):
+        obs_component_selected = not mask.data[0, 0]
```
michaelosthege:
```suggestion
    @pytest.mark.parametrize("obs_component_selected", [True, False])
    def test_multivariate_constant_mask_separable(self, obs_component_selected):
        mask = pt.constant(np.array([[obs_component_selected] * 4], dtype=bool))
```
simpler?

tests/distributions/test_distribution.py
```diff
@@ -700,3 +702,217 @@ def test_dtype(self, floatX):
...
+            }
+        )
+        logp_fn = pytensor.function([], list(logp.values()))
+        obs_logp, unobs_logp = logp_fn()
+        expected_logp = pm.logp(rv, unobs_data).eval()
+        np.testing.assert_almost_equal(obs_logp, [])
+        np.testing.assert_array_equal(unobs_logp, expected_logp)
+
+        # Test that we can update a shared mask
+        mask.set_value(np.array([False]))
```
michaelosthege, line 848: 🤯

michaelosthege (approved):
cool stuff!!

pymc/model.py
ricardoV94: This could count as a breaking change, that's why I didn't change it. Worth the trouble?

tests/distributions/test_distribution.py
ricardoV94: Yup

pymc/model.py
ricardoV94: I do think unobserved is a better name

michaelosthege: I'd say it's okay to break. Imputation isn't that popular, and the breaking change isn't dangerously silent or sth like that.

ricardoV94: It changes the variable name, that's all

ricardoV94: Updated

PR #6799, opened 2023-06-28
Remove `SeededTest` class

### LINKED ISSUE #6234: Remove `SeededTest` class
This class is not that relevant now that our sampling routines do not rely on global seeding. It causes some confusions because it seems like the tests within should be seeded by default (they used to)

The only few uses that I think are still relevant is when we use it to obtain a seed or create fake data with legacy numpy random routines (i.e., not Generators). I think we can just refactor it to not rely on a base Test class.

aerubanov:
This PR remove `SeededTest` class and close #6234 

----
:books: Documentation preview :books:: https://pymc--6799.org.readthedocs.build/en/6799/

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -44,8 +44,7 @@ def strict_float32():
 
 @pytest.fixture(scope="function", autouse=False)
 def seeded_test():
-    # TODO: use this instead of SeededTest
-    np.random.seed(42)
+    np.random.seed(20160911)
 
 
 @pytest.fixture
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -775,24 +775,7 @@ def discrete_random_tester(
         assert p > alpha, str(point)
 
 
-class SeededTest:
-    random_seed = 20160911
-    random_state = None
-
-    @classmethod
-    def setup_class(cls):
-        nr.seed(cls.random_seed)
-
-    def setup_method(self):
-        nr.seed(self.random_seed)
-
-    def get_random_state(self, reset=False):
-        if self.random_state is None or reset:
-            self.random_state = nr.RandomState(self.random_seed)
-        return self.random_state
-
-
-class BaseTestDistributionRandom(SeededTest):
+class BaseTestDistributionRandom:
     """
     Base class for tests that new RandomVariables are correctly
     implemented, and that the mapping of parameters between the PyMC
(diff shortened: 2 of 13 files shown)


aerubanov:
Some tests fails after removing `SeededTest` and I`m working on fixing it.

aerubanov:
@ricardoV94 , @twiecki Could you please review this PR when you will have some time?

tests/conftest.py
```diff
@@ -44,4 +44,4 @@ def strict_float32():
 @pytest.fixture(scope="function", autouse=False)
 def seeded_test():
     # TODO: use this instead of SeededTest
-    np.random.seed(42)
+    np.random.seed(20160911)
```
ricardoV94, line 47: Any reason for this change?

tests/conftest.py
```diff
@@ -44,4 +44,4 @@ def strict_float32():
 @pytest.fixture(scope="function", autouse=False)
 def seeded_test():
     # TODO: use this instead of SeededTest
```
ricardoV94:
```suggestion
```

ricardoV94 (commented):
Looks pretty good. Small question and suggestion below

tests/conftest.py
aerubanov: I found some test that fails with random seed 42 - the result differs from expected just a little bit more than the selected threshold value. That is why I decided to use here value from the `SeededTest` class for random seed.

ricardoV94 (approved):

ferrine:
Why not using a fixture with autouse?

ricardoV94:
> Why not using a fixture with autouse?

For what? The point in that global seeding doesn't do anything in PyMC anymore so each test should create a generator/seed to pass to the functions that need seeding.

Do you mean a fixture to create such a seed?

aerubanov:
@ricardoV94 I think @ferrine ask about pytest fixture [autouse](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request) argument to apply `seeded_test` fixture to all tests.

ricardoV94:
> @ricardoV94 I think @ferrine ask about pytest fixture [autouse](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request) argument to apply `seeded_test` fixture to all tests.

I don't think we should do that. If anything we should move away from using any numpy RandomState stuff, so the fixture would be useless.

ricardoV94:
There's a conflict with main now. Otherwise I think it's good to merge

ricardoV94:
Thanks @aerubanov!

Issue #6792, opened 2023-06-23
CAR exhibits inconsistent behavior at the alpha parameter boundary


daniel-saunders-phil:
# Motivating examples

Currently, the CAR distribution is set up to accept values >= -1 and <= 1. When alpha = 1, the distribution starts behaving in difficult to understand ways, giving what looks like inconsistent answers on very similar problems. Here's two examples that illustrate the problem. First, suppose the adjacency structure is a line. If you try to evaluate the logp of some example values, it returns `nan`.
```python
import pymc as pm
import numpy as np

N = 4
adj_matrix = np.array([[0,1,0,0],
                      [1,0,1,0],
                      [0,1,0,1],
                      [0,0,1,0]])

values = np.array([0.5,0.1,0.2,0.6])

phi = pm.CAR.dist(mu=np.zeros(N), tau=1, alpha=1, W=adj_matrix)

pm.logp(phi,values).eval()
```
However, if we adjust the adjacency structure slightly so it's a square lattice instead, it returns a value, in this case -17.8452531.
```python
N = 4
adj_matrix = np.array([[0,1,0,1],
                      [1,0,1,0],
                      [0,1,0,1],
                      [1,0,1,0]])

values = np.array([0.5,0.1,0.2,0.6])

phi = pm.CAR.dist(mu=np.zeros(N), tau=1, alpha=1, W=adj_matrix)

pm.logp(phi,values).eval()
```
The pattern of results holds up across a bunch of adjustments. If you pass it a fully connected 4x4 adjacency matrix, `nan`. If you pass a 4x4 adjacency matrix where everyone is connected except one pair, `logp` gives you a value. Similarly things apply for matrices of different sizes. I tried both types of 3x3 adjacency matrices and got one nan and one result. `pm.draw` produces results that follow a similar pattern. If the logp would return nan, then draw throws an error and says the matrix is not positive semi-definite. Finally, none of these issues arise if alpha is any positive number less than 1.

# Why is this happening

The big picture here is that when alpha = 1, the resulting precision matrix, $Q$, contains linear dependencies and cannot be inverted to get the covariance matrix:

$$\phi \sim MvN(0,Q^{-1})$$

PyMC's implementation doesn't take the matrix inverse of $Q$ directly but uses an eigenvalue strategy. I think this strategy is allowing PyMC carry out matrix operations that would otherwise be undefined. I dug around in the source code until I could isolate the cause to numerical precision errors. The logp finds eigenvalues with:
```python
[...]

PR #6801, opened 2023-06-28
Restrict domain on alpha in the CAR distribution

daniel-saunders-phil:
#6792

Hi, I think I need some help writing the checks correctly. The intended change is supposed to be fairly trivial. the alpha parameter should be restricted to be (-1,1). I tried making the changes that I thought would enforce them but they are not working. The `make_node()` check is not triggering when I create the random variable with `.dist` or when I use `pm.draw()`. The `lopg()` check is triggering all the time, even when alpha is in the acceptable range.

Maybe something we can discuss @fonnesbeck @bwengals today?

----
:books: Documentation preview :books:: https://pymc--6801.org.readthedocs.build/en/6801/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2063,8 +2063,6 @@ class CARRV(RandomVariable):
         tau = pt.as_tensor_variable(floatX(tau))
 
         alpha = pt.as_tensor_variable(floatX(alpha))
-        msg = "the domain of alpha is: -1 < alpha < 1."
-        alpha = Assert(msg)(alpha, pt.lt(alpha, 1) and pt.gt(alpha, -1))
 
         return super().make_node(rng, size, dtype, mu, W, alpha, tau)
 
@@ -2080,6 +2078,9 @@ class CARRV(RandomVariable):
         Journal of the Royal Statistical Society Series B, Royal Statistical Society,
         vol. 63(2), pages 325-338. DOI: 10.1111/1467-9868.00288
         """
+        if np.all(alpha >= 1) or np.all(alpha <= -1):
+            raise ValueError("the domain of alpha is: -1 < alpha < 1")
+
         if not scipy.sparse.issparse(W):
             W = scipy.sparse.csr_matrix(W)
         s = np.asarray(W.sum(axis=0))[0]
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2078,7 +2078,7 @@ class CARRV(RandomVariable):
         Journal of the Royal Statistical Society Series B, Royal Statistical Society,
         vol. 63(2), pages 325-338. DOI: 10.1111/1467-9868.00288
         """
-        if np.all(alpha >= 1) or np.all(alpha <= -1):
+        if np.any(alpha >= 1) or np.any(alpha <= -1):
             raise ValueError("the domain of alpha is: -1 < alpha < 1")
 
         if not scipy.sparse.issparse(W):
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2061,7 +2061,10 @@ class CARRV(RandomVariable):
             W = Assert(msg)(W, pt.allclose(W, W.T))
 
         tau = pt.as_tensor_variable(floatX(tau))
+
         alpha = pt.as_tensor_variable(floatX(alpha))
+        msg = "the domain of alpha is: -1 < alpha < 1."
+        alpha = Assert(msg)(alpha, pt.lt(alpha, 1) and pt.gt(alpha, -1))
 
         return super().make_node(rng, size, dtype, mu, W, alpha, tau)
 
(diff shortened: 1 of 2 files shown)


pymc/distributions/multivariate.py
```diff
@@ -2061,7 +2061,10 @@ def make_node(self, rng, size, dtype, mu, W, alpha, tau):
             W = Assert(msg)(W, pt.allclose(W, W.T))
 
         tau = pt.as_tensor_variable(floatX(tau))
+
         alpha = pt.as_tensor_variable(floatX(alpha))
+        msg = "the domain of alpha is: -1 < alpha < 1."
+        alpha = Assert(msg)(alpha, pt.lt(alpha, 1) and pt.gt(alpha, -1))
```
ricardoV94: You shouldn't put this assert here. Instead raise an error in the `rng_fn` if the alpha is actually problematic for random draws.

Otherwise Asserts will complicate the graph needlessly.

Anyway the reason you don't see the exception is probably that you need to use `pt.and_` instead of python `and`.

daniel-saunders-phil:
> > The logp() check is triggering all the time, even when alpha is in the acceptable range.
> 
> What do you mean?

    RV = CAR.dist(W=np.array([[1,0],[0,1]]),alpha=.95,mu=[0,0],tau=1)
    pm.logp(RV,np.array([1,1.5]))

returns `Check{-1 < alpha < 1 and tau > 0}.0` instead of the logp value.

ricardoV94:
`logp` returns a symbolic graph. You need to actually compile it (or for debugging) call `.eval`. During compilation the check will be removed if it's not needed (i.e. alpha is constant and in the required range).

ricardoV94:
This may be helpful to understand: https://www.pymc.io/projects/docs/en/stable/learn/core_notebooks/pymc_pytensor.html#enough-with-random-variables-i-want-to-see-some-log-probabilities

daniel-saunders-phil:
> `logp` returns a symbolic graph. You need to actually compile it in a `pytensor.function` and call it (or for debugging call `.eval`) to get a number our of it.
> 
> During compilation the check will be removed if it's not needed (i.e. alpha is constant and in the required range).

Oh right of course! yeah the logp is behaving properly too.

pymc/distributions/multivariate.py
daniel-saunders-phil: Yup the `pt.and_` was what I was missing. So instead of using assert in the rng function, I just raise a ValueError? Is that right?

    if alpha >= 1 or alpha <= -1:
                raise ValueError("the domain of alpha is: -1 < alpha < 1")

ricardoV94: Yes, but alpha could be a vector so you need a `np.all(...)`

daniel-saunders-phil: gotcha - okay sent in a new commit with these changes.

daniel-saunders-phil:
I'm happy with the PR at this point - just having trouble with codecov. It looks like coverage increases on the file I changed but I'm getting a lot of indirect coverage issues that are causing it to fail. No clue what I need to fix. Does anyone know why these might be coming up?

pymc/distributions/multivariate.py
```diff
@@ -2077,6 +2078,9 @@ def rng_fn(cls, rng: np.random.RandomState, mu, W, alpha, tau, size):
         Journal of the Royal Statistical Society Series B, Royal Statistical Society,
         vol. 63(2), pages 325-338. DOI: 10.1111/1467-9868.00288
         """
+        if np.all(alpha >= 1) or np.all(alpha <= -1):
```
ricardoV94: I assume you mean this?

```suggestion
        if np.any(alpha >= 1) or np.any(alpha <= -1):
```

tests/distributions/test_multivariate.py
```diff
@@ -835,6 +835,23 @@ def test_car_matrix_check(sparse):
...
+
+    W = np.array([[0, 1, 0], [1, 0, 1], [0, 1, 0]])
+
+    tau = 1
+    mu = np.array([0, 0, 0])
+
+    car_dist = pm.CAR.dist(W=W, alpha=alpha, mu=mu, tau=tau)
+
+    with pytest.raises(ValueError, match="the domain of alpha is: -1 < alpha < 1"):
+        pm.draw(car_dist)
```
ricardoV94, line 853: Maybe worth to test the logp as well?

ricardoV94:
The codecov here on GitHub is a bit flaky, don't worry about it

tests/distributions/test_multivariate.py
```diff
@@ -835,6 +835,23 @@ def test_car_matrix_check(sparse):
             car_dist = pm.CAR.dist(mu, W, alpha, tau)
 
 
+@pytest.mark.parametrize("alpha", [1, -1])
+def test_car_alpha(alpha):
```
```suggestion
def test_car_alpha_bounds(alpha):
```

tests/distributions/test_multivariate.py
daniel-saunders-phil: Yeah I think so too. Is it okay if I write it within the same test or do we need two separate ones? I went with putting them in the same test with the latest commit but happy to reorganize it.

bwengals: I think either is OK, it's clear what you're testing either way

bwengals (approved):
This looks good to me!  It looks like codecov is comparing your branch to 7b08fc1, which was from a couple weeks ago, not current main, which is causing those indirect coverage changes.  Am I correct in guessing that this is why it's a non-issue then for merging @ricardoV94?

ricardoV94:
> This looks good to me!  It looks like codecov is comparing your branch to 7b08fc1, which was from a couple weeks ago, not current main, which is causing those indirect coverage changes.  Am I correct in guessing that this is why it's a non-issue then for merging @ricardoV94?

I never investigated why codecov bot is so flaky so it's nice to have an hypothesis. Many times I had to ignore because ot was clearly wrong (as is here)

Issue #6398, opened 2022-12-14
Implement logprob derivation for some forms of AdvancedIndexing mixtures


ricardoV94:
> Are we guaranteed to always be able to later rewrite this safely into something that is a proper logprob?

There are two types of logps that are relevant here. 1) An RV that is broadcasted directly and 2) an RV where we are doing advanced indexing (i.e. mixture)

For the former, we would need to perform symbolic unbroadcasting of the value variable, and then evaluate the base RV at that value

```python
def logp_broadcast_to(value, base_rv, broadcast_shape):
  return logp(base_rv, unbroadcast(value, broadcast_shape))
```

I am not sure what the `unbroadcast` function would look like, but it should do the following (in pseudo-code), when valid, and raise otherwise.
```python
np.testing.assert_array_equal(unbroadcast(broadcast_to(value, shape), shape).eval(), value)
```

For the second, we would need to somehow only evaluate the logp at the unique indexes. Something like 

```python
def logp_mixture_advanced_integer_indexing(value, base_rvs, indexes):
  unique_indexes = [pt.unique(index) for index in indexes]
  # And either ignore or assert the value should be the same for repeated indexes
  return logp_basic_mixture_rv(value[unique_indexes], base_rvs, unique_indexes)
```

This requires some thought to implement, which is why this PR does not attempt to do so.

We could at least implement the logp for the cases where indexes are unique (either because the user promised us so, or we added an assert)

```python
def logp_mixture_advanced_integer_indexing(value, base_rvs, indexes):
  indexes_with_assert = [assert_op(index, pt.unique(index).size == index.size) for index in indexes]
  return logp_basic_mixture_rv(value, base_rvs, indexes_with_assert)
```

_Originally posted by @ricardoV94 in https://github.com/pymc-devs/pymc/issues/6369#issuecomment-1339693790_

Om-Doiphode:
I would like to work on this issue. Can you tell me on how to get started with it?

Issue #6360, opened 2022-11-30
Incorporate RV meta information in intermediate MeasurableVariables


ricardoV94:
### Description

This pertains to the `logprob` submodule. During logprob derivation of an expression like

```python
import numpy as np
import pymc as pm

x_raw = pm.Normal.dist(np.arange(5), shape=(2, 5))
x = pm.math.clip(x_raw, -1, 1)  # Censored normal

pm.logp(x, np.zeros((2, 5)))
```
We create a `MeasurableClip` that replaces `x`, when we identify we can derive the logprob as a simple censored pdf. This MeasurableClip however does not retain any of the meta-information about the type of RV that it encapsulates (ndim_supp, dtype, support axis).

https://github.com/pymc-devs/pymc/blob/a0d6ba079eac2f044ed40cc5747f1079d99f9f16/pymc/logprob/censoring.py#L61-L67

If we wanted to further compose the graph, we would find issues when some operation needs to know this information

```python
x_raw = pm.Normal.dist(np.arange(5), shape=(2, 5))
x = pm.math.clip(x_raw, -1, 1)  # Censored normal<
x = x.T

pm.logp(x, np.zeros((5, 2)))  # NotImplementedError: PyMC could not infer logp of input variable.
```

This happens because to infer the logprob of a transposed (dimshuffled) variable, we need to know the original support dimensionality and support axis (which is always the rightmost for pure distributions):

https://github.com/pymc-devs/pymc/blob/a0d6ba079eac2f044ed40cc5747f1079d99f9f16/pymc/logprob/tensor.py#L285-L298

If we propagated that information to the MeasurableClip (ndim_supp=0, support_axis=None, dtype="mixed"), the `Dimshuffle` rewrite could be safely used and we could derive the logp for the second example. This is also useful for other rewrites...

More context in https://github.com/aesara-devs/aeppl/discussions/183

Dhruvanshu-Joshi:
Hey @ricardoV94 . I would like to work on this issue. From what I understand, we will need to pass meta-information all Measurable ops like MeasurableClip, MeasurableROund, MeasurableDimShuffle, etc using an ```init``` function. If you think my understanding and approach are the right direction, I'll proceed with implementing this.

ricardoV94:
Yes. And the meta-info we want (at least for now) is: `ndim_supp`, `supp_axis`, and something new like `type` which can be discrete, continuous, or mixed.

ricardoV94:
[...]

PR #6808, opened 2023-06-30
Derive probability for broadcasting operations

ricardoV94:
Related to #6398

TODO: 
- [ ] Cover `Second`/`Alloc` which are other froms of broadcasting

----
:books: Documentation preview :books:: https://pymc--6808.org.readthedocs.build/en/6808/

CC @shreyas3156

### DIFF
--- a/pymc/logprob/shape.py
+++ b/pymc/logprob/shape.py
@@ -1,0 +1,121 @@
+#   Copyright 2023 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+from typing import Optional
+
+import numpy as np
+import pytensor.tensor as pt
+
+from pytensor.graph import node_rewriter
+from pytensor.tensor.extra_ops import BroadcastTo
+
+from pymc.logprob.abstract import MeasurableVariable, _logprob, _logprob_helper
+from pymc.logprob.rewriting import PreserveRVMappings, measurable_ir_rewrites_db
+
+
+class MeasurableBroadcast(BroadcastTo):
+    pass
+
+
+MeasurableVariable.register(MeasurableBroadcast)
+
+
+measurable_broadcast = MeasurableBroadcast()
+
+
+@_logprob.register(MeasurableBroadcast)
+def broadcast_logprob(op, values, rv, *shape, **kwargs):
+    """Log-probability expression for (statically-)broadcasted RV
+
+    The probability is the same as the base RV, if no broadcasting had happened:
+
+    ``logp(broadcast_to(normal(size=(3, 1)), (2, 3, 4)), zeros((2, 3, 4))) == logp(normal(size=(3, 1)), zeros((3, 1)))``
+
+    And zero if the value couldn't have possibly originated via broadcasting:
+
+    ``logp(broadcast_to(normal(size=(1,)), (3,)), [1, 2, 3]) == [-np.inf]``
+
+    """
+    [value] = values
+
+    n_new_dims = len(shape) - rv.ndim
+    assert n_new_dims >= 0
+
+    # Enumerate broadcasted dims
+    expanded_dims = tuple(range(n_new_dims))
+    broadcast_dims = tuple(
+        i + n_new_dims
+        for i, (v_bcast, rv_bcast) in enumerate(
+            zip(value.broadcastable[n_new_dims:], rv.broadcastable)
+        )
+        if (not v_bcast) and rv_bcast
+    )
+
+    # "Unbroadcast" value via indexing.
+    # All entries in the broadcasted dimensions should be the same, so we simply select the first of each.
+    indices = []
+    for i in range(value.ndim):
+        # Remove expanded dims
+        if i in expanded_dims:
+            indices.append(0)
+        # Keep first entry of broadcasted (but not expanded) dims
+        elif i in broadcast_dims:
+            indices.append(slice(0, 1))
+        else:
+            indices.append(slice(None))
+
+    unbroadcast_value = value[tuple(indices)]
+    logp = _logprob_helper(rv, unbroadcast_value)
+
+    # Check that dependent values were indeed identical, by comparing with a re-broadcasted value
+    valid_value = pt.broadcast_to(unbroadcast_value, shape)
+    # Note: This could fail due to float-precision issues.
+    # If that proves to be a problem we should switch to `pt.allclose`
+    check = pt.all(pt.eq(value, valid_value))
+    logp = pt.switch(check, logp, -np.inf)
+
+    # Reintroduce expanded_dims in the returned logp
+    if n_new_dims > 0:
+        logp = pt.shape_padleft(logp, n_new_dims)
+
+    return logp
+
+
+@node_rewriter([BroadcastTo])
+def find_measurable_broadcast(fgraph, node):
+    r"""Finds `BroadcastTo`\s for which a `logprob` can be computed."""
+
+    if isinstance(node.op, MeasurableBroadcast):
+        return None  # pragma: no cover
+
+    rv_map_feature: Optional[PreserveRVMappings] = getattr(fgraph, "preserve_rv_mappings", None)
+
+    if rv_map_feature is None:
+        return None  # pragma: no cover
+
+    base_rv, *shape = node.inputs
+
+    if not rv_map_feature.request_measurable([base_rv]):
+        return None
+
+    new_rv = measurable_broadcast.make_node(base_rv, *shape).default_output()
+
+    return [new_rv]
+
+
+measurable_ir_rewrites_db.register(
+    "find_measurable_broadcast",
+    find_measurable_broadcast,
+    "basic",
+    "shape",
+)
--- a/pymc/logprob/shape.py
+++ b/pymc/logprob/shape.py
@@ -1,0 +1,137 @@
+#   Copyright 2024 - present The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


larryshamalama:
I will get to this tomorrow morning, sorry about the delay!

pymc/logprob/shape.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+MeasurableVariable.register(MeasurableBroadcast)
+
+
+measurable_broadcast = MeasurableBroadcast()
+
+
+@_logprob.register(MeasurableBroadcast)
+def broadcast_logprob(op, values, rv, *shape, **kwargs):
```
larryshamalama, line 34: Thinking out loud: could this possibly result in inconsistencies elsewhere? For instance, having Mixture components that have been broadcasted which would render them dependent, if that would be an issue

pymc/logprob/shape.py
```diff
@@ -0,0 +1,121 @@
...
+
+    ``logp(broadcast_to(normal(size=(1,)), (3,)), [1, 2, 3]) == [-np.inf]``
+
+    """
+    [value] = values
+
+    n_new_dims = len(shape) - rv.ndim
+    assert n_new_dims >= 0
+
+    # Enumerate broadcasted dims
```
larryshamalama, line 55: Trying to follow along here, this comment is more for "mental scribbles".
```python
rv = pt.random.normal(size=(3, 1))
x = pt.broadcast_to(rv, (5, 2, 3, 4)) # a bit more than your example above
# rv.broadcastable = (False, False, False, False)

n_new_dims = 2 # 4 - 2
expanded_dims = (0, 1)

value.broadcastable[n_new_dims:] = (False, False) # (3, 4)
rv.broadcastable = (False, True) # (3, 1)

# condition is True only: if (not v_bcast) and rv_bcast = if (not False) and True
# condition is True only if v_bast is False and rv_bcast is True
broadcast_dims = (3,) # (0 + 2, 1 + 2) but conditions are (False, True)?
```

pymc/logprob/shape.py
```diff
@@ -0,0 +1,121 @@
...
+    expanded_dims = tuple(range(n_new_dims))
+    broadcast_dims = tuple(
+        i + n_new_dims
+        for i, (v_bcast, rv_bcast) in enumerate(
+            zip(value.broadcastable[n_new_dims:], rv.broadcastable)
+        )
+        if (not v_bcast) and rv_bcast
+    )
```
larryshamalama, line 63: What is `broadcast_dims` in words? It's a bit hard to follow 😅

larryshamalama: AH, is it the axis/es which is/are being broadcasted?

pymc/logprob/shape.py
```diff
@@ -0,0 +1,121 @@
...
+        return None  # pragma: no cover
+
+    rv_map_feature: Optional[PreserveRVMappings] = getattr(fgraph, "preserve_rv_mappings", None)
+
+    if rv_map_feature is None:
+        return None  # pragma: no cover
+
+    base_rv, *shape = node.inputs
+
+    if not rv_map_feature.request_measurable([base_rv]):
```
larryshamalama: PR #6746 has been a while, but I just want to clarify `request_measurable`. If `base_rv` is unmeasurable, we exit? This method is the new `assign_custom_measurable_outputs`, is that right?

pymc/logprob/shape.py
```diff
@@ -0,0 +1,121 @@
...
+            indices.append(slice(None))
+
+    unbroadcast_value = value[tuple(indices)]
+    logp = _logprob_helper(rv, unbroadcast_value)
+
+    # Check that dependent values were indeed identical, by comparing with a re-broadcasted value
+    valid_value = pt.broadcast_to(unbroadcast_value, shape)
+    # Note: This could fail due to float-precision issues.
+    # If that proves to be a problem we should switch to `pt.allclose`
+    check = pt.all(pt.eq(value, valid_value))
```
larryshamalama: It is my understanding that this segment (and some lines above) are to ensure that we have:
```python
logp(broadcast_to(normal(size=(1,)), (3,)), [1, 2, 3]) == [-np.inf]
```
Can you briefly explain the process of unbroadcasting and re-broadcasting to perform this check?

tests/logprob/test_transforms.py
```diff
@@ -807,16 +807,36 @@ def test_discrete_rv_multinary_transform_fails():
         conditional_logp({y_rv: y_rv.clone()})
 
 
-@pytest.mark.xfail(reason="Check not implemented yet")
-def test_invalid_broadcasted_transform_rv_fails():
+@pytest.mark.filterwarnings("error")  # Fail if unexpected warning is issued
+@pytest.mark.parametrize("implicit_broadcast", (True, False))
+def test_broadcasted_transform_rv(implicit_broadcast):
```
larryshamalama: Correct me if I'm wrong here: logprob inference for broadcasted variables work only if they are 1) pure RVs and 2) affine-transformed pure RVs?

larryshamalama (commented):
Great work @ricardoV94! :) A lot of nice abstractions, which, together, are why I have many questions

pymc/logprob/shape.py
ricardoV94: The index mixture only works for basic RVs still so that's fine.

The switch mixture could actually wrongly broadcast the logp. In fact we should also check for invalid switches that mix support dimensions. The current implementation is only correct for `ndim_supp==0`!

This is another example of why it's so important to have the meta-info for all the MeasurableOps (#6360).

Once we have the meta-info, the Mixture will unambiguously know what kind of measurable variable it is dealing with. In the case of MeasurableBroadcasting, for example, the `ndim_supp` will have to be at least as large as the number of broadcasted dims (which means we should collapse that logp dimension instead of leaving it as we were doing now!). 

We will also know where those support dims are, so that Mixture can know whether we are sub-selecting across core dims.

Without the meta-info, the only way of knowing `ndim_supp` is by checking the dimensionality of the value vs the logp. We use this logic in some places already:

https://github.com/pymc-devs/pymc/blob/f67ff8bc22fcc72ead96eeeaef04173ed53650cb/pymc/logprob/transforms.py#L432-L437

https://github.com/pymc-devs/pymc/blob/f67ff8bc22fcc72ead96eeeaef04173ed53650cb/pymc/logprob/tensor.py#L185-L189

Which makes me worry whether the probability of a transformed broadcasted variable may be invalid because the "Jacobian" term is going to be counted multiple times?

pymc/logprob/shape.py
ricardoV94: Yes

pymc/logprob/shape.py
ricardoV94: No, this method is what guides the search for measurable variables. Quoting myself:

> Instead of eager bottom-up conversion to measurable variables in the IR rewrites, we only convert nodes whose outputs were marked as "needs_measuring". This is achieved with the new `PreserveRVMappings.request_measurable method`.

The `assign_custom_measurable_outputs` is completely gone.

pymc/logprob/shape.py
ricardoV94: We check if we can recreate the original broadcasted value by selecting the first entries and rebroadcasting. If we cannot, the graph couldn't have arise from broadcasting.

In the simplest case:
```python
import numpy as np

a = np.array([0, 0, 0, 0, 0])
assert np.all(a == np.broadcast_to(a[0], a.shape))  # Checks out

a = np.array([0, 1, 2, 3, 4])
assert np.all(a == np.broadcast_to(a[0], a.shape))  # Raises
```

tests/logprob/test_transforms.py
ricardoV94: No. It should work for anything that is explicitly under `BroadcastTo` or implicitly in one of the supported binary Elemwise (Add/ Mul / Pow...)

pymc/logprob/shape.py
larryshamalama: I understand, I'm still getting used to these newer changes

pymc/logprob/shape.py
larryshamalama: Thanks for clarifying, the example here helps

pymc/logprob/shape.py
ricardoV94: You raised a very good point, which makes me wonder to what extent https://github.com/pymc-devs/pymc/pull/6797 is correct in general? 

For instance, if you scale a `3-vector Dirichlet` you shouldn't count the Jacobian 3 times, because one of the entries is redundant.

Do we need to propagate information about over-determined elements in multi-dimensional RVs?

ricardoV94: The first part of this answer suggests you count it 3 times indeed: https://stats.stackexchange.com/a/487538

I'm surprised :D

Edit: As seen below, that answer is wrong

ricardoV94: This I think says something else and correct? https://upcommons.upc.edu/bitstream/handle/2117/366723/p20-CoDaWork2011.pdf?sequence=1&isAllowed=y

I think these should match:
```python
import pymc as pm
import numpy as np

x = 0.75
print(
    pm.logp(pm.Beta.dist(5, 9), x).eval(),
    pm.logp(pm.Dirichlet.dist([5, 9]), [x, 1-x]).eval(),
)  # -3.471576058736023 -3.471576058736023

print(
    pm.logp(2 * pm.Beta.dist(5, 9), 2 * x).eval(),
    pm.logp(2 * pm.Dirichlet.dist([5, 9]), 2*np.array([x, 1-x])).eval(),
)  # -4.164723239295968 -4.857870419855914

print(
    pm.logp(2 * pm.Beta.dist(5, 9), 2 * x).eval(),
    (pm.logp(pm.Dirichlet.dist([5, 9]), ([x, 1-x])) - np.log(2)).eval(),
)  # -4.164723239295968 -4.164723239295968
```

larryshamalama: > Once we have the meta-info, the Mixture will unambiguously know what kind of measurable variable it is dealing with. In the case of MeasurableBroadcasting, for example, the `ndim_supp` will have to be at least as large as the number of broadcasted dims (which means we should collapse that logp dimension instead of leaving it as we were doing now!).

This makes sense! Would you say that it's better to wait for #6360?

> The first part of this answer suggests you count it 3 times indeed: https://stats.stackexchange.com/a/487538
> 
> I'm surprised :D

I'm not sure if I fully follow 😅 Nonetheless, I'm glad that this question raised some interesting concerns

jessegrabowski (approved):

PR #6809, opened 2023-07-02
Prevent unbound trace due to type hints

thomasaarholt:
There are currently two type errors in the Model class, preventing smooth usage with type checkers (like in VSCode):
This one became present in recent versions of pyright:
```python
# script.py
import pymc as pm
import arviz as az

with pm.Model() as m: # <-- (...) does not implement __enter__ 
    pm.Normal("a", mu=0, sigma=1)
    trace = pm.sample()

az.plot_posterior(trace)
```
and before that one, this one has been there for a long time:
```python
# script.py
import pymc as pm
import arviz as az

with pm.Model() as m:
    pm.Normal("a", mu=0, sigma=1)
    trace = pm.sample()

az.plot_posterior(trace) # <-- 9:19 - error: "trace" is possibly unbound (reportUnboundVariable)
```

This PR implements two small changes:
1. Sets the type of the `instance` variable in the ContextMeta `__call__` method to "Model". This fixes the first error.
2. Sets the return type of the context manager used by `with pm.Model()` to return `None`, which means that it won't be catching errors.

In the first fix, `instance` should be a resultant type that the `ContextMeta` is a metaclass for. I'm not sure how if a type hint that says "returns the child of a metaclass" exists. But I think this is safe, since `ContextMeta` isn't used as a metaclass anywhere else, and I'm not going for perfect type coverage, just to make it easy for people to code with autocompletion.

**Checklist**

## Bugfixes
- Fixed pm.Model not being interpreted as a context manager by pyright
- Fixed unbound type errors due to objects being created inside the pm.Model context

----
:books: Documentation preview :books:: https://pymc--6809.org.readthedocs.build/en/6809/

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -50,6 +50,7 @@ from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.type import RandomType
 from pytensor.tensor.sharedvar import ScalarSharedVariable
 from pytensor.tensor.var import TensorConstant, TensorVariable
+from typing_extensions import Self
 
 from pymc.blocking import DictToArrayBijection, RaveledVars
 from pymc.data import GenTensorVariable, is_minibatch
@@ -478,10 +479,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
     if TYPE_CHECKING:
 
-        def __enter__(self: "Model") -> "Model":
+        def __enter__(self: Self) -> Self:
             ...
 
-        def __exit__(self: "Model", *exc: Any) -> bool:
+        def __exit__(self, exc_type: None, exc_val: None, exc_tb: None) -> None:
             ...
 
     def __new__(cls, *args, **kwargs):
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -107,6 +107,14 @@ class ContextMeta(type):
     the `with` statement.
     """
 
+    if TYPE_CHECKING:
+
+        def __enter__(self: Self) -> Self:
+            ...
+
+        def __exit__(self, exc_type: None, exc_val: None, exc_tb: None) -> None:
+            ...
+
     def __new__(cls, name, bases, dct, **kwargs):  # pylint: disable=unused-argument
         """Add __enter__ and __exit__ methods to the class."""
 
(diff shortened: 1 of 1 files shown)


thomasaarholt:
Annoyingly, I just updated pyright to the latest version, and pyright now says that `pm.Model` can't be used as a context manager, despite the presence of the `if TYPE_CHECKING` code block within it.

```
script.py:5:6 - error: Object of type "type[Model]" cannot be used with "with" because it does not implement __enter__ (reportGeneralTypeIssues)
script.py:5:6 - error: Object of type "type[Model]" cannot be used with "with" because it does not implement __exit__ (reportGeneralTypeIssues)
```

It *should* work, but my guess is that this is due to the complexity / metaclass stuff :/ 

I still think this PR should be merged though :)

thomasaarholt:
Hah! I figured it out. Pyright was hung up on the [`instance`](https://github.com/pymc-devs/pymc/blob/f67ff8bc22fcc72ead96eeeaef04173ed53650cb/pymc/model.py#L216) type returned by `__new__` being incorrect. `instance` should be a resultant type that the ContextMeta is a metaclass for, so overriding this to `"Model"` fixes the type checking, and the old `if TYPE_CHECKING` works as expected.

thomasaarholt:
Tagging @michaelosthege on this, since he reviewed my other typing PR. I've also edited the original post in this PR and deleted my now superfluous comments.

I realise it's summer holidays for many, and I'm in no rush to get it merged, but don't want this to get lost.

ricardoV94:
Thanks @thomasaarholt!

pymc/model.py
```diff
@@ -478,10 +481,10 @@ def __init__(self, mean=0, sigma=1, name=''):
 
     if TYPE_CHECKING:
```
michaelosthege, line 482: is this even necessary?

michaelosthege (approved):

pymc/model.py
thomasaarholt: Yes - I tried removing it at first, but pyright is unable to infer that `Model`, which is a child of the metaclass `ContextMeta`, implements `__enter__`.

PR #6478, opened 2023-01-24
The Gumbel Distribution is incorrectly described

helios1014:
The version currently in PYTensor is the Gumbel Right Skew which gives the extreme maximum. Anyone using this to find the extreme minimum--in survival analysis for example, is using the wrong distribution. A sign change on the x and y inputs is an effective solution if the user is notified before hand.

**What is this PR about?**
...

**Checklist**
+ [x] Explain important implementation details 👆
+ [ ] Make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).
+ [ ] Link relevant issues (preferably in [nice commit messages](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html))
+ [ ] Are the changes covered by tests and docstrings?
+ [ ] Fill out the short summary sections 👇

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- Muy only chnage is to the documentation for the distribution so that it reads as floows:
    Univariate Gumbel right Skew log-likelihood. This parameterization will provide the extrem maximum value. 
    Those looking to find the extrem minimum provided by the left skew Gumbel should 
    invert the sign to all X and Y values (positive to negative and negative to positive).

## Maintenance
- ...

PR #6810, opened 2023-07-02
Adding description to Gumbel Distribution

amyoshino:
**What is this PR about?**
Closes PR "The Gumbel Distribution is incorrectly described" #6478

It looks like it is only missing the corrections pointed out in the PR #6478. 
I have modified helios1014's PR to address comments and revisions made by michaelosthege and OriolAbril.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6810.org.readthedocs.build/en/6810/


pymc/distributions/continuous.py
```diff
@@ -3121,7 +3121,11 @@ def triangular_default_transform(op, rv):
 
 class Gumbel(Continuous):
     r"""
-    Univariate Gumbel log-likelihood.
+    Univariate Gumbel right Skew log-likelihood.
```
larryshamalama:
```suggestion
    Univariate right skewed Gumbel log-likelihood.
```
Or do we prefer "right-skewed" (hyphenated)?

pymc/distributions/continuous.py
```diff
@@ -3121,7 +3121,11 @@ def triangular_default_transform(op, rv):
...
+
+    This parameterization will provide the extreme maximum value.
```
larryshamalama: What does this mean? I'm not sure to understand... Maybe "This distribution is typically used for modeling maximum (or extreme) values."

pymc/distributions/continuous.py
```diff
@@ -3121,7 +3121,11 @@ def triangular_default_transform(op, rv):
...
+    Those looking to find the extreme minimum provided by the left skewed Gumbel should
+    invert the sign to all X and Y values (positive to negative and negative to positive).
```
larryshamalama:
```suggestion
    invert the sign to all X and Y values.
```

larryshamalama: As per above, Gumbel is a univariate distribution, what do we mean by X and Y values here? I understand using a negative to model the minimum or retrieve the left-skewed Gumbel, maybe we can phrase this better... I have no particular suggestion in mind on top of my head 😅

larryshamalama (commented):
Thanks for opening this PR :) Some comments on clarifying the description

pymc/distributions/continuous.py
amyoshino: Will add it in the next commit.

EDIT: Added in the last commit!

pymc/distributions/continuous.py
amyoshino: The original PR https://github.com/pymc-devs/pymc/pull/6478 referred to this [SciPy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gumbel_r.html) reference. There, the `y` is equivalent of the `z` used in [our documentation](https://www.pymc.io/projects/docs/en/latest/api/distributions/generated/pymc.Gumbel.html).

However, I believe it would be better to modify the text to point changes in the sign of all `x` and `mu`, (which is equivalent to inverting the sign of  all `x` and `z`) because we have control over the `mu` as an arguments of the function, but not `z` (I believe). 

I have tested this and our implementation matches [scipy.stats.gumbel_r ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gumbel_r.html)(for Max) and [scipy.stats.gumbel_l](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gumbel_l.html) (for Min, with negative x and mu passed to our function).

image

I have corrected the text to be "Those looking to find the extreme minimum provided by the left-skewed Gumbel should invert the sign to all x and mu values."

Let me know if this make sense or we prefer to refer `z` instead of mu since it is defined in our documentation below the pdf formula.

pymc/distributions/continuous.py
amyoshino: I like your "right-skewed" (hyphenated) suggestion!

amyoshino:
I have made all changes I believe we needed to address your review. Thanks again for the guidance and let me know if we are missing anything.

pymc/distributions/continuous.py
```diff
@@ -3121,7 +3121,11 @@ def triangular_default_transform(op, rv):
 
 class Gumbel(Continuous):
     r"""
-    Univariate Gumbel log-likelihood.
+    Univariate right-skewed Gumbel log-likelihood.
+
+    This distribution is typically used for modeling maximum (or extreme) values.
+    Those looking to find the extreme minimum provided by the left-skewed Gumbel should
+    invert the sign to all x and mu values.
```
larryshamalama:
```suggestion
    invert the sign of all x and mu values.
```

larryshamalama (approved):
Looks good, thanks @amyoshino for this PR and thanks for your patience 😅

amyoshino:
@larryshamalama thank you again for your guidance!! I am happy to be able to contribute 😄

PR #6821, opened 2023-07-09
Fix mathematical notation in conditional_logprob docstrings

### LINKED ISSUE #6798: DOC: Fix mathematical notation in `conditional_logprob` docstrings
### Description

Nitpick issue here

- $\Sigma$ is conventionally used for covariance matrices, so I prefer using $\sigma$ since the example is univariate.

https://github.com/pymc-devs/pymc/blob/7b08fc160bc07adcfcd10e135e727c681ffb1b77/pymc/logprob/basic.py#L433-L434

- For continuous-valued random variables, I would prefer not writing $P(Y = y)$ because, in principle, it is always zero.
https://github.com/pymc-devs/pymc/blob/7b08fc160bc07adcfcd10e135e727c681ffb1b77/pymc/logprob/basic.py#L448-L449

What I would recommend is perhaps to use the random variables as indices to the log-density and valued variables as arguments to delineate their differences while addressing the point above: $\log p_{Y, \sigma^2}(y, s^2) = \log p_{Y \mid \sigma^2}(y \mid \sigma^2 = s^2) + \log p_{\sigma^2}(s^2)$. To me, this is better, but maybe not for everyone...

The last instance where change may be needed is: https://github.com/pymc-devs/pymc/blob/7b08fc160bc07adcfcd10e135e727c681ffb1b77/pymc/logprob/basic.py#L438

amyoshino:
**What is this PR about?**
Closes #6798

This PR Fix mathematical notation in `conditional_logprob` docstrings

The changes are pointed out by @larryshamalama in #6798. The notation changes are:
- Given that $\Sigma$ is commonly used for covariance matrices, it is replaced by $\sigma$ which is a more common notation.
- $P(Y = y)$ is replaced by $P_{Y}(y)$ to avoid incorrect notation in the case of continuous-valued random variables.

image

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6821.org.readthedocs.build/en/6821/

### DIFF
--- a/pymc/logprob/basic.py
+++ b/pymc/logprob/basic.py
@@ -430,12 +430,12 @@ def conditional_logp(
 
     .. math::
 
-        \Sigma^2 \sim& \operatorname{InvGamma}(0.5, 0.5) \\
-        Y \sim& \operatorname{N}(0, \Sigma)
+        s^2 \sim& \operatorname{InvGamma}(0.5, 0.5) \\
+        Y \sim& \operatorname{N}(0, s)
 
     If we create a value variable for ``Y_rv``, i.e. ``y_vv = pt.scalar("y")``,
     the graph of ``conditional_logp({Y_rv: y_vv})`` is equivalent to the
-    conditional log-probability :math:`\log p(Y = y \mid \Sigma^2)`, with a stochastic
+    conditional log-probability :math:`\log p_{Y \mid s^2}(y \mid s^2)`, with a stochastic
     ``sigma2_rv``.
 
     If we specify a value variable for ``sigma2_rv``, i.e.
@@ -445,8 +445,8 @@ def conditional_logp(
 
     .. math::
 
-        \log p(Y = y, \Sigma^2 = \sigma^2) =
-            \log p(Y = y \mid \Sigma^2 = \sigma^2) + \log p(\Sigma^2 = \sigma^2)
+        \log p_{Y,s^2}(y, \sigma^2) =
+            \log p_{Y \mid s^2}(y \mid \sigma^2) + \log p_{s^2}(\sigma^2)
 
 
     Parameters
--- a/pymc/logprob/basic.py
+++ b/pymc/logprob/basic.py
@@ -430,7 +430,7 @@ def conditional_logp(
 
     .. math::
 
-        s^2 \sim& \operatorname{InvGamma}(0.5, 0.5) \\
+        \sigma^2 \sim& \operatorname{InvGamma}(0.5, 0.5) \\
         Y \sim& \operatorname{N}(0, s)
 
     If we create a value variable for ``Y_rv``, i.e. ``y_vv = pt.scalar("y")``,
--- a/pymc/logprob/basic.py
+++ b/pymc/logprob/basic.py
@@ -431,7 +431,7 @@ def conditional_logp(
     .. math::
 
         \sigma^2 \sim& \operatorname{InvGamma}(0.5, 0.5) \\
-        Y \sim& \operatorname{N}(0, s)
+        Y \sim& \operatorname{N}(0, \sigma^2)
 
     If we create a value variable for ``Y_rv``, i.e. ``y_vv = pt.scalar("y")``,
     the graph of ``conditional_logp({Y_rv: y_vv})`` is equivalent to the
--- a/pymc/logprob/basic.py
+++ b/pymc/logprob/basic.py
@@ -435,7 +435,7 @@ def conditional_logp(
 
     If we create a value variable for ``Y_rv``, i.e. ``y_vv = pt.scalar("y")``,
     the graph of ``conditional_logp({Y_rv: y_vv})`` is equivalent to the
-    conditional log-probability :math:`\log p_{Y \mid s^2}(y \mid s^2)`, with a stochastic
+    conditional log-probability :math:`\log p_{Y \mid \sigma^2}(y \mid s^2)`, with a stochastic
     ``sigma2_rv``.
 
     If we specify a value variable for ``sigma2_rv``, i.e.
@@ -445,8 +445,8 @@ def conditional_logp(
 
     .. math::
 
-        \log p_{Y,s^2}(y, \sigma^2) =
-            \log p_{Y \mid s^2}(y \mid \sigma^2) + \log p_{s^2}(\sigma^2)
+        \log p_{Y, \sigma^2}(y, s^2) =
+            \log p_{Y \mid \sigma^2}(y \mid s^2) + \log p_{\sigma^2}(s^2)
 
 
     Parameters
--- a/pymc/logprob/basic.py
+++ b/pymc/logprob/basic.py
@@ -439,7 +439,7 @@ def conditional_logp(
     ``sigma2_rv``.
 
     If we specify a value variable for ``sigma2_rv``, i.e.
-    ``s_vv = pt.scalar("s2")``, then ``conditional_logp({Y_rv: y_vv, sigma2_rv: s_vv})``
+    ``s2_vv = pt.scalar("s2")``, then ``conditional_logp({Y_rv: y_vv, sigma2_rv: s2_vv})``
     yields the conditional log-probabilities of the two variables.
     The sum of the two terms gives their joint log-probability.
 


amyoshino:
@larryshamalama I have followed your suggestions in the issue #6798 but I am not sure if the representation for the line L438 is correct: $\log p(Y = y \mid \Sigma^2)$ --> $\log p_{Y \mid s^2}(y \mid s^2)$. It looks odd to me because I am repeating the $s$ in the subscript and in the conditional side. Let me know if this is not the best way to represent it.

In the lines L448-L449 I shifted the $\Sigma$ with $s$ in your suggestion because I believe this is how the original version is  described after replacing $\Sigma$ (subscript) with $s$ (value). Let me know if I am mistaken here as well. 
https://www.pymc.io/projects/docs/en/stable/api/generated/pymc.logprob.conditional_logp.html#pymc.logprob.conditional_logp

pymc/logprob/basic.py
```diff
@@ -430,12 +430,12 @@ def conditional_logp(
...
+        s^2 \sim& \operatorname{InvGamma}(0.5, 0.5) \\
+        Y \sim& \operatorname{N}(0, s)
 
     If we create a value variable for ``Y_rv``, i.e. ``y_vv = pt.scalar("y")``,
     the graph of ``conditional_logp({Y_rv: y_vv})`` is equivalent to the
-    conditional log-probability :math:`\log p(Y = y \mid \Sigma^2)`, with a stochastic
+    conditional log-probability :math:`\log p_{Y \mid s^2}(y \mid s^2)`, with a stochastic
     ``sigma2_rv``.
 
     If we specify a value variable for ``sigma2_rv``, i.e.
```
larryshamalama, line 441: Line below: 
```python
 s2_vv = pt.scalar("s2")``, then ``conditional_logp({Y_rv: y_vv, sigma2_rv: s2_vv})
```

pymc/logprob/basic.py
```diff
@@ -430,12 +430,12 @@ def conditional_logp(
 
     .. math::
 
-        \Sigma^2 \sim& \operatorname{InvGamma}(0.5, 0.5) \\
-        Y \sim& \operatorname{N}(0, \Sigma)
+        s^2 \sim& \operatorname{InvGamma}(0.5, 0.5) \\
```
larryshamalama:
```suggestion
        \sigma^2 \sim& \operatorname{InvGamma}(0.5, 0.5) \\
```

pymc/logprob/basic.py
```diff
@@ -430,12 +430,12 @@ def conditional_logp(
...
+        Y \sim& \operatorname{N}(0, s)
```
larryshamalama:
```suggestion
        Y \sim& \operatorname{N}(0, \sigma^2)
```

pymc/logprob/basic.py
```diff
@@ -430,12 +430,12 @@ def conditional_logp(
...
 
     If we create a value variable for ``Y_rv``, i.e. ``y_vv = pt.scalar("y")``,
     the graph of ``conditional_logp({Y_rv: y_vv})`` is equivalent to the
-    conditional log-probability :math:`\log p(Y = y \mid \Sigma^2)`, with a stochastic
+    conditional log-probability :math:`\log p_{Y \mid s^2}(y \mid s^2)`, with a stochastic
```
larryshamalama:
```suggestion
    conditional log-probability :math:`\log p_{Y \mid \sigma^2}(y \mid s^2)`, with a stochastic
```
Note here that the arguments and indices should be different

pymc/logprob/basic.py
```diff
@@ -445,8 +445,8 @@ def conditional_logp(
 
     .. math::
 
-        \log p(Y = y, \Sigma^2 = \sigma^2) =
-            \log p(Y = y \mid \Sigma^2 = \sigma^2) + \log p(\Sigma^2 = \sigma^2)
+        \log p_{Y,s^2}(y, \sigma^2) =
```
larryshamalama:
```suggestion
        \log p_{Y, \sigma^2}(y, s^2) =
```
For this line in particular, your original line can be correct. To keep consistent with the code and other comments, I would prefer what I just suggested

pymc/logprob/basic.py
```diff
@@ -445,8 +445,8 @@ def conditional_logp(
...
+            \log p_{Y \mid s^2}(y \mid \sigma^2) + \log p_{s^2}(\sigma^2)
```
larryshamalama:
```suggestion
            \log p_{Y \mid \sigma^2}(y \mid s^2) + \log p_{\sigma^2}(s^2)
```
Idem as the comment immediately above

larryshamalama (commented):
Thanks @amyoshino for opening this PR, you're on a roll!

Apologies if the initial issue was confusing. I wrote a long text ahead, none of this is major but rather a clarification of what I previously meant. I can get quite picky with mathematical notation...

> @larryshamalama I have followed your suggestions in the issue #6798 but I am not sure if the way I represented the line L438 is correct: log⁡p(Y=y∣Σ2) --> log⁡pY∣s2(y∣s2). It looks odd to me because I am repeating the s in the subscript and in the conditional side. Let me know if this is not the best way to represent it.

Your instinct is correct, we shouldn't have s as a subscript (random variable) and as an argument (value variable). In the hierarchical model definition, we just want random variables. In the mathematical function representing the log-density, arguments are the "value variables". I hope that this clarifies things.

We ideally want to adopt a notation that is more "standard" and easier to understand. That's mostly subjective... 1) I prefer using $\sigma^2$ to denote the variance. 2) Often, using both random variables as indices and arguments to denote what densities we are referring to can be too much... Here, I think that it's important, because we would like to delineate random variables from value variables in PyMC/PyTensor.

It seems that you use $\sigma^2$ as an argument and $s^2$ as the stochastic variable. (\*) I personally would change the order, especially that we define `sigma2_rv` to be the PyTensor random variable and `s_vv` to be the valued variable. (\*) We should probably even write `s2_vv` instead of `s_vv`. If you have any preferences, you can add your suggestion. Otherwise, I provided mine in this review.

In short, your instinct correctly addresses an error in mathematical notation, but the solution can be entirely subjective. Actionable items are presented by a (\*) for clarity :)

pymc/logprob/basic.py
larryshamalama, line 441: Line below
```python
s2_vv = pt.scalar("s2")``, then ``conditional_logp({Y_rv: y_vv, sigma2_rv: s2_vv})
```

pymc/logprob/basic.py
larryshamalama:
```suggestion
    conditional log-probability :math:`\log p_{Y \mid \sigma^2}(y \mid s^2)`, with a stochastic
```
(I messed up something in the original review comment)

pymc/logprob/basic.py
amyoshino: Just pushed in the last commit "[s_vv to s2_vv](https://github.com/pymc-devs/pymc/pull/6821/commits/c44a5d2f6f6b17ce28081664840fb8ee9d40d696)"

amyoshino:
@larryshamalama thank you for the comments and review! I think it makes way more sense to me now, my apologies for not making sense of your initial comments and changing the use of $\sigma$ and $s$. I believe it is now very clear to the readers. 

I have commited your suggestions and added a last commit to cover the changes from `s_vv` to `s2_vv`. Sorry the number of commits as well, I started accepting one by one instead of accepting all in one batch, lesson learned for the next PR :)

I think everything is ok now, but let me know if I forgot to address anything or if you have any other suggestions or comments.

larryshamalama:
> @larryshamalama thank you for the comments and review! I think it makes way more sense to me now, my apologies for not making sense of your initial comments and changing the use of σ and s. I believe it is now very clear to the readers.

Ah please don't be sorry, I'm very happy to help :)

> I have commited your suggestions and added a last commit to cover the changes from `s_vv` to `s2_vv`. Sorry the number of commits as well, I started accepting one by one instead of accepting all in one batch, lesson learned for the next PR :)

Don't worry about the number of commits! We often squash merge, but there is a certain "art" to writing nice commits. I'm still learning myself, @ricardoV94 writes great commits if you want to have some inspiration. Around the beginning of my open-source involvement, I was referred to this website, which I found somewhat helpful: https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes

larryshamalama (approved):
Looks good, thanks @amyoshino!

amyoshino:
@larryshamalama Thank you for your guidance!

PR #6822, opened 2023-07-10
Allow creating CustomDist inside another CustomDist

ricardoV94:
Fixes issue describe in this discourse thread: https://discourse.pymc.io/t/dynamic-shaping-round-function-jax-and-a-few-more-questions/12264/17?u=ricardov94

----
:books: Documentation preview :books:: https://pymc--6822.org.readthedocs.build/en/6822/


larryshamalama (approved):
Looks good! A more general question from my side: when do we want to use `BlockModelAccess`, i.e. prevent user access to Model contexts? Can we specify this in the `BlockModelAccess` docstring?

PR #6775, opened 2023-06-12
Probabilty inference for arc transformations

LukeLB:
Users have asked for the ability to use aditional transformations #6631, also see https://discourse.pymc.io/t/the-sinh-arcsinh-normal-distribution/12136. This ended up being quite fast as the logic is very similar to many of the previous transforms I've implemented.

I've added these transforms to the existing tests, which pass. However, it seems there are a few tests that aren't passing in tests/logprob/test_transform.py. I think they may have been failing prior to me, as they fail on main as well....

...

**Checklist**
+ [x] Explain important implementation details 👆
+ [x] Make sure that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html).
+ [x] Link relevant issues (preferably in [nice commit messages](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html))
+ [x] Are the changes covered by tests and docstrings?
+ [x] Fill out the short summary sections 👇

## New features
- Arcsinh transform
- Arccosh transform
- Arctanh transform

----
:books: Documentation preview :books:: https://pymc--6775.org.readthedocs.build/en/6775/

PR #6826, opened 2023-07-13
Derive logprob for exp2, log2, log10, log1p, expm1, log1mexp, log1pexp (softplus), and sigmoid transformations

LukeLB:
This builds upon the previous pull requests, #6664 and #6775, and completes the work of #6631.

I have attempted to rewrite the logp graph only for log1p, expm1, and log1pexp (softplus). My reasoning is that in these cases, the inputs are transformed directly without affecting the backward or log_jac_det transforms. For all other cases, I have made changes to the existing transform classes.

Although some tests are still failing (see below), I'm uncertain about the reasons for the first two failures. However, it seems that the final failure is due to a floating point error, and I'm unsure how to resolve it.
```
FAILED test_transforms.py::test_transformed_logprob[at_dist6-dist_params6-<lambda>-size6] - AssertionError: 
E           AssertionError: 
E           Arrays are not almost equal to 4 decimals
E           
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference: 0.03384969
E           Max relative difference: 0.00578904
E            x: array(-5.8811)
E            y: array(-5.8472)
FAILED test_transforms.py::test_mixture_transform - assert False
E       assert False
E        +  where False = equal_computations([Sum{axes=None}.0], [Sum{axes=None}.0])
FAILED test_transforms.py::test_check_jac_det[transform9] - AssertionError: 
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=0
E           
E           Mismatched elements: 1 / 2 (50%)
E           Max absolute difference: 6.59892563e-09
E           Max relative difference: 1.69410928e-07
E            x: array([ 1.05966 , -0.038952])
E            y: array([ 1.05966 , -0.038952])
```

Currently, there are no specific tests written for the transforms log1p, log1mexp, log1pexp (softplus), and sigmoid. I would appreciate assistance in developing tests for these transforms.
...

**Checklist**

## Major / Breaking Changes
- Changes to existing Exp and Log Transform classes 

## New features
- The ability to derive log prob for the following transforms exp2, log2, log10, log1p, expm1, log1mexp, log1pexp (softplus), and sigmoid.

----
:books: Documentation preview :books:: https://pymc--6826.org.readthedocs.build/en/6826/


pymc/logprob/transforms.py
```diff
@@ -377,7 +393,15 @@ class MeasurableTransform(MeasurableElemwise):
 
     valid_scalar_types = (
         Exp,
+        Exp2,
+        Expm1,
+        Sigmoid,
         Log,
+        Log2,
+        Log10,
+        Log1p,
```
ricardoV94: The cases that don't require a new transform aren't needed here, since we convert them to previously supported cases

pymc/logprob/transforms.py
```diff
@@ -563,7 +621,12 @@ def find_measurable_transforms(fgraph: FunctionGraph, node: Node) -> Optional[Li
 
     transform_dict = {
         Exp: ExpTransform(),
+        Exp2: ExpTransform(base=2),
+        Expm1: ExpTransform(m=True),
         Log: LogTransform(),
+        Log2: LogTransform(base=2),
```
ricardoV94: I was thinking of converting stuff like `log2(x) -> log(x) / log(2)` . This way we don't need to implement the new transforms

ricardoV94: I think it can also be done for exponents?

pymc/logprob/transforms.py
```diff
@@ -940,6 +1030,20 @@ def log_jac_det(self, value, *inputs):
             return value
 
 
+class SigmoidTransform(RVTransform):
```
ricardoV94: For the sigmoid we can use the `1 / 1 + exp(-x)` equivalence, which PyMC should be able to then derive as well. PyTensor has many rewrites to stabilize sigmoid forms so no need to worry about it at this point

ricardoV94 (commented):
Thanks a lot for picking this up. My suggestions are all about being a bit lazier (less code to maintain and test), but the idea is totally right!

ricardoV94:
> My reasoning is that in these cases, the inputs are transformed directly without affecting the backward or log_jac_det transforms. For all other cases, I have made changes to the existing transform classes.

You don't need to worry about jacobian when rewriting equivalent forms like `log2(c) -> log(c) / log(2)`. PyMC will have to introduce them anyway when it derives the probability of the second graph.

pymc/logprob/transforms.py
LukeLB: I see! Just to check my understanding then, I can add to this function
```python 
@node_rewriter([log1p, softplus, log1mexp])
def measurable_special_log_to_log(fgraph, node):
    """Convert log1p, log1mexp, softplus of `MeasurableVariable`s to log form."""
    [inp] = node.inputs

    if isinstance(node.op.scalar_op, Log1p):
        return [pt.log(1 + inp)]
    if isinstance(node.op.scalar_op, Softplus):
        return [pt.log(1 + pt.exp(inp))]
    if isinstance(node.op.scalar_op, Log1mexp):
        return [pt.log(1 - pt.exp(pt.neg(inp)))]
    if isinstance(node.op.scalar_op, Log2):
        return pt.log(inp)/pt.log(2)
```

pymc/logprob/transforms.py
LukeLB: Yeah that makes sense. Would it also make sense to tie in the `LogOddsTransform` into similar a rewrite, `pt.log(inp/1-inp)`?

ricardoV94: That one requires a special logic because inp shows up twice in the graph. PyMC should actually refuse to convert that graph (I am curious if it actually does). 

It should require its own specialized measurable rewrite.

pymc/logprob/transforms.py
ricardoV94: Yes 👍

pymc/logprob/transforms.py
```diff
@@ -523,8 +539,54 @@ def measurable_sub_to_neg(fgraph, node):
...
+
+@node_rewriter([exp2, expm1, sigmoid])
+def measurable_special_exp_to_exp(fgraph, node):
+    """Convert log1p, log1mexp, softplus of `MeasurableVariable`s to log form."""
+    [inp] = node.inputs
+
+    if isinstance(node.op.scalar_op, Exp2):
+        return [pt.power(2, inp)]
+    if isinstance(node.op.scalar_op, Expm1):
+        return [pt.exp(inp + 1)]
```
ricardoV94: I think this is the correct equivalency?

```suggestion
        return [pt.exp(inp) - 1]
```

pymc/logprob/transforms.py
```diff
@@ -523,8 +539,54 @@ def measurable_sub_to_neg(fgraph, node):
...
+        return [pt.log(inp) / pt.log(10)]
+
+
+@node_rewriter([exp2, expm1, sigmoid])
+def measurable_special_exp_to_exp(fgraph, node):
+    """Convert log1p, log1mexp, softplus of `MeasurableVariable`s to log form."""
+    [inp] = node.inputs
+
+    if isinstance(node.op.scalar_op, Exp2):
+        return [pt.power(2, inp)]
```
ricardoV94: I don't think we support this one, only powers with fixed exponent and variable base

ricardoV94: You can convert to `exp(ln(2)*x)` instead, which PyMC will know how to handle

ricardoV94: We should actually use that for any `power(const, x) -> exp(log(const) * x)` which we currently don't support. But maybe that's better left for another PR? 

It requires checking we are interested in x and not const.

LukeLB: Ah yeah I had a feeling that one may be an issue, I'll make the change. I'll add a new function to generalise this power(const, x) -> exp(log(const) * x) functionality.

LukeLB: So I've just pushed this new functionality. Do we care that it won't work for const <= 0?

pymc/logprob/transforms.py
```diff
@@ -523,8 +539,64 @@ def measurable_sub_to_neg(fgraph, node):
...
+@node_rewriter([exp2, pow])
+def measurable_general_exp_to_exp(fgraph, node):
+    """Convert exp2 and any const^x of `MeasurableVariable`s to exp form."""
+    if len(node.inputs) > 1:
+        [const, inp] = node.inputs
+    else:
+        [inp] = node.inputs
+    if isinstance(node.op.scalar_op, Exp2):
+        return [pt.exp(pt.log(2) * inp)]
+    if isinstance(node.op.scalar_op, Pow) and isinstance(inp, pt.TensorVariable):
```
ricardoV94: Have to think about this carefully because stuff like x^2 needs the old special logic and for that the power must be fixed. Also this will fail with negative exponents

ricardoV94: I suggest we leave the new power stuff to a distinct PR

LukeLB: Unless I'm mistaken this conserves the old logic so x^n still works as normal, because if x^2 then `inp` will be a scalar. But I agree it fails for const <= 0 which is more challenging. I'll revert back to just dealing exp2 and we can deal with the general power stuff another time.

ricardoV94: Sounds good. Also I was handwavy with "const". I meant a conditionally independent variable (either a constant or a variable that's valued).

LukeLB: Ah I see, that makes it harder

LukeLB:
Woops, sorry about that I will fix

LukeLB:
It seems that the code coverage test is failing because if the new node rewriting functions. Should we be testing them? Or are they fine because they rely on existing functionality thats already tested?

ricardoV94:
> It seems that the code coverage test is failing because if the new node rewriting functions. Should we be testing them? Or are they fine because they rely on existing functionality thats already tested?

We should test them. We can test that the output `logp` is the same for the new cases we support and what we think they are equivalent to, using the `equal_computations` helper. Something like

```python
def test():
  base_rv = pt.random.normal(name="base_rv")
  vv = pt.scalar("vv")
  
  logp_test = logp(pt.log1p(base_rv), vv)
  logp_ref = logp(pt.log(1 + base), vv)
  assert equal_computations([logp_test], [logp_ref])
```

pymc/logprob/transforms.py
ricardoV94: Let's open a separate issue / PR for this. I thought it over a bit and I think we can definitely support a couple of cases.

```
power(const, x), for any const > 0 and any x
power(const, x), for any const and discrete x  (we can play with `log(abs(neg_const))` and x's parity)
```
The first case we don't have to constrain ourselves to actual "constants", we can add a symbolic assert that `const > 0`.

The second requires us to implement transforms for discrete variables, which would probably need #6360 first, so we can focus on the first case, which is also probably more useful anyway.

We just have to make sure not to rewrite stuff like `power(x, const)` accidentally as those are implemented via our `PowerTransform`. This can be done by checking which of the inputs has a path to unvalued random variables.

ricardoV94:
@LukeLB shall we push the PR to the finish line? Let me know if you don't have the time right now

LukeLB:
Hey really sorry about the lack of communication, I've been on holiday. I'm going to have a look at this this week.

LukeLB:
Ahhh done the same thing as before and somehow merged unrelated commits! Will fix on my side

LukeLB:
While I fix that I am having a problems with two of the tests. For the test
```python
TRANSFORMATIONS = {
    "log1p": (pt.log1p, lambda x: pt.log(1 + x)),
    "softplus": (pt.softplus, lambda x: pt.log(1 + pt.exp(x))),
    "log1mexp": (pt.log1mexp, lambda x: pt.log(1 - pt.exp(pt.neg(x)))),
    "log2": (pt.log2, lambda x: pt.log(x) / pt.log(2)),
    "log10": (pt.log10, lambda x: pt.log(x) / pt.log(10)),
    "exp2": (pt.exp2, lambda x: pt.exp(pt.log(2) * x)),
    "expm1": (pt.expm1, lambda x: pt.exp(x) - 1),
    "sigmoid": (pt.sigmoid, lambda x: 1 / (1 + pt.exp(-x))),
}


@pytest.mark.parametrize("transform", TRANSFORMATIONS.keys())
def test_special_log_exp_transforms(transform):
    base_rv = pt.random.normal(name="base_rv")
    vv = pt.scalar("vv")

    transform_func, ref_func = TRANSFORMATIONS[transform]
    transformed_rv = transform_func(base_rv)
    ref_transformed_rv = ref_func(base_rv)

    logp_test = logp(transformed_rv, vv)
    logp_ref = logp(ref_transformed_rv, vv)

    assert equal_computations([logp_test], [logp_ref])
```
when transform is log2 or log10 then test fails for equal computation, I'm not sure what is causing that...

LukeLB:
@ricardoV94 OK I think I've got to the bottom of what is causing the failure. Looks like a floating point in precision in one of the nodes of the graph between the test case and the reference case.
image
However I'm not sure I can control that in the original lambda function that I've written. Any ideas?

ricardoV94:
>  However I'm not sure I can control that in the original lambda function that I've written. Any ideas?

Hmm. Let's do a logp evaluation for that one (separate test) and check for output closeness?

ricardoV94:
File changes is showing some accidental overwriting previous changes? https://github.com/pymc-devs/pymc/pull/6826/files

tests/logprob/test_transforms.py
```diff
@@ -1118,14 +1117,73 @@ def test_cosh_rv_transform():
...
+)
+def test_check_jac_det(transform):
+    check_jacobian_det(
+        transform,
+        Vector(Rplusbig, 2),
+        pt.dvector,
+        [0.1, 0.1],
+        elemwise=True,
+        rv_var=pt.random.normal(0.5, 1, name="base_rv"),
+    )
```
ricardoV94: You probably missed this during the rebase, but these tests were moved into this TestClass https://github.com/pymc-devs/pymc/blob/346f085bee900e992d99fd26eedc904299ad44e6/tests/logprob/test_transforms.py#L751-L759

pymc/logprob/transforms.py
```diff
@@ -569,8 +585,53 @@ def measurable_sub_to_neg(fgraph, node):
...
+def measurable_special_log_to_log(fgraph, node):
+    """Convert log1p, log1mexp, softplus, log2, log10 of `MeasurableVariable`s to log form."""
+    [inp] = node.inputs
+
+    if isinstance(node.op.scalar_op, Log1p):
+        return [pt.log(1 + inp)]
+    if isinstance(node.op.scalar_op, Softplus):
+        return [pt.log(1 + pt.exp(inp))]
+    if isinstance(node.op.scalar_op, Log1mexp):
+        return [pt.log(1 - pt.exp(pt.neg(inp)))]
```
ricardoV94: Are you sure the input should be negated? I think not

```suggestion
        return [pt.log(1 - pt.exp(pt.neg(inp)))]
```

tests/logprob/test_transforms.py
LukeLB: My bad, I will fix

pymc/logprob/transforms.py
LukeLB: > Are you sure the input should be negated? I think not

Unless I'm missing something, it appears to be negated in other documentation https://www.rdocumentation.org/packages/VGAM/versions/1.1-8/topics/log1mexp.

ricardoV94: We are not using that implementation. The `Op` is defined here: https://github.com/pymc-devs/pytensor/blob/30e08e23c1ab0995eb8c34cc3914a9917228f2d0/pytensor/scalar/math.py#L1277

tests/logprob/test_transforms.py
```diff
@@ -1052,6 +1052,32 @@ def test_negative_value_frac_power_transform_logp(self, power):
...
+            CoshTransform(),
+            TanhTransform(),
+            ArcsinhTransform(),
+            ArccoshTransform(),
+            ArctanhTransform(),
+            LogTransform(),
+            ExpTransform(),
+        ],
+    )
+    def test_check_jac_det(self, transform):
```
ricardoV94: This is still duplicated. The original one is in the class `TestRVTransform`

ricardoV94 (approved):

ricardoV94:
Awesome work @LukeLB. Do you want to pursue the power exponent next?

LukeLB:
Thanks @ricardoV94, I feel like I learnt a lot on this one! 

> Do you want to pursue the power exponent next?

Yep I'll start looking into that, if I run into any trouble I'll drop you a mesage on slack.

twiecki:
Congrats @LukeLB! This is a major and non-trivial contribution.

PR #6828, opened 2023-07-14
Improve support for `dims` in `LKJCholeskyCov`

jessegrabowski:
**What is this PR about?**

There are currently several pain points when using  labeled dims with`LKJCholeskyCov`:

1. The distribution samples 2d matrices, but passing a pair of dimensions results in an error, because internally the distribution is represented in packed lower-triangular form.
2. Any dimensions given to `LKJCholeskyCov` are not propagated to internally generated deterministics, `{name}_std` and `{name}_corr`. If a user wants these to be labeled, he has to pass a long dictionary to `idata_kwargs` 
3. After sampling, the 1's on the diagonal of all samples drawn from `{name}_corr` causes an error in `arviz` when computing within-chain variance. 

This PR tries to correct all three of these. Here is an example model under this PR:

```python
from string import ascii_uppercase

n = 3
n_obs = 100
mean = np.zeros(n)
L = np.random.normal(size=(n, n))
cov = L @ L.T
data = np.random.multivariate_normal(mean=mean, cov=cov, size=(n_obs, ))

with pm.Model(coords={'dim':ascii_uppercase[:n], 
                      'dim_aux':ascii_uppercase[:n]},
              coords_mutable={'obs_idx':np.arange(n_obs, dtype='int')}) as mod:
    sd_dist = pm.Exponential.dist(1)
    chol, *_ = LKJCholeskyCov('chol', n=n, sd_dist=sd_dist, eta=1, dims=['dim', 'dim_aux'])
    obs = pm.MvNormal('obs', mu=0, chol=chol, observed=data, dims=['obs_idx', 'dim'])
    
    idata = pm.sample()
```

First, I pass two dimensions to `LKJCholeskyCov` -- one for the columns, and one for the rows. This corresponds to the expectation that I am drawing from a matrix-valued random variable.

Internally, I take the Cartesian product between these two dims, and use the lower triangle of the resulting matrix to make and register a new coordinate: `packed_tril_{name}`. This is then set as the dims on `packed_chol`. 

Next, only the upper triangle (excluding the diagonal) of the correlation matrix is stored in a deterministic. Another new coordinate is registered: `corr_{name}`.

Finally, the first dim is used to add a labeled dimension to `{name}_std`.

This results in the following graph:
image

Here is the result plotted with `az.plot_trace`:
image

This PR still needs a bit of work, including:

1. Unit tests for the new functionality
2. Documentation
3. The generated dimensions are added in a "hacky" way, I hope this can be improved
4. The names on the "packed" dimensions of chol and cov are not great. It would be nice if a MultiIndex could be specified here, but I don't think it's currently possible without some changes (see [here](https://discourse.pymc.io/t/pymc-arviz-how-to-make-the-most-of-labeled-coords-and-dims-in-pymc-4-0/9581/9?u=jessegrabowski), but maybe this is out of date?)  

It's also possible that the problem (3) is a problem on the arviz side of things, and should be fixed there instead of here. But in that case, it would still be nice to propagate the matrix dims to the full square correlation matrix.

Because of these points, I'm marking this as a draft PR. But I would still like feedback on the idea of automatically generating coords, or at least on how dim handling can be improved in `LKJCholeskyCov`  

**Checklist**

## Major / Breaking Changes
- None

## New features
- Allow `LKJCholeskyCov` to generate and register new model `coords` corresponding to the distributions it internally registers

## Bugfixes
- Allows plotting of generated correlation matrix in `az.plot_trace`, but that might not be something that should be fixed on the PyMC side. See discussion above.

## Documentation
-None

## Maintenance
-None

----
:books: Documentation preview :books:: https://pymc--6828.org.readthedocs.build/en/6828/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -1413,14 +1413,32 @@ class LKJCholeskyCov:
     """
 
     def __new__(cls, name, eta, n, sd_dist, *, compute_corr=True, store_in_trace=True, **kwargs):
+        dims = kwargs.pop("dims", None)
+
+        if dims is not None:
+            # TODO: Add check for 2d dims?
+            packed_dim_name, packed_dim_value = cls._make_packed_coord_from_dims(
+                n, dims, "packed_tril"
+            )
+            cls._register_new_coords_with_model(packed_dim_name, packed_dim_value)
+            kwargs["dims"] = [packed_dim_name]
+
         packed_chol = _LKJCholeskyCov(name, eta=eta, n=n, sd_dist=sd_dist, **kwargs)
+
         if not compute_corr:
             return packed_chol
         else:
             chol, corr, stds = cls.helper_deterministics(n, packed_chol)
             if store_in_trace:
-                corr = pm.Deterministic(f"{name}_corr", corr)
-                stds = pm.Deterministic(f"{name}_stds", stds)
+                corr_triu = corr[pt.triu_indices_from(corr, k=1)]
+                corr_triu_dim_name, corr_triu_dim_value = cls._make_packed_coord_from_dims(
+                    n, dims, "corr", lower=False, k=1
+                )
+                cls._register_new_coords_with_model(corr_triu_dim_name, corr_triu_dim_value)
+
+                corr_tril = pm.Deterministic(f"{name}_corr", corr_triu, dims=corr_triu_dim_name)
+                stds = pm.Deterministic(f"{name}_stds", stds, dims=dims[0])
+
             return chol, corr, stds
 
     @classmethod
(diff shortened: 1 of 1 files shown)


pymc/distributions/multivariate.py
```diff
@@ -1463,6 +1482,28 @@ def helper_deterministics(cls, n, packed_chol):
...
+        tri_coords = coord_product[flat_tri_idx].tolist()
+
+        packed_dim_name = f"{name_prefix}_{dims[0]}"
+        return packed_dim_name, tri_coords
+
+    @classmethod
+    def _register_new_coords_with_model(cls, name, value):
+        mod = pm.modelcontext(None)
+        mod.coords[name] = value
+        mod.dim_lengths[name] = pt.TensorConstant(pt.lscalar, np.array(len(value)))
```
ricardoV94: Should probably use `model.add_coord/s` and check name doesn't clash with existing one

pymc/distributions/multivariate.py
```diff
@@ -1463,6 +1482,28 @@ def helper_deterministics(cls, n, packed_chol):
...
+    def _make_packed_coord_from_dims(cls, n, dims, name_prefix, lower=True, k=0):
+        mod = pm.modelcontext(None)
+        chol_dims = [mod.coords[dim] for dim in dims]
+        if lower:
+            f_idx = np.tril_indices
+        else:
+            f_idx = np.triu_indices
+
+        flat_tri_idx = np.arange(n**2, dtype=int).reshape(n, n)[f_idx(n, k=k)]
+        coord_product = np.fromiter([f"{x}" for x in product(*chol_dims)], dtype="object")
```
aseyboldt: I'm a bit scared that this could break things when we serialize traces. Can we at least store those in netcdf and zarr?
I wouldn't mind if this just had integer coords either...

pymc/distributions/multivariate.py
```diff
@@ -1463,6 +1482,28 @@ def helper_deterministics(cls, n, packed_chol):
...
+        tri_coords = coord_product[flat_tri_idx].tolist()
+
+        packed_dim_name = f"{name_prefix}_{dims[0]}"
```
aseyboldt: I would much prefer a postfix instead of a prefix, especially in this case, because we already use that for the deterministics.

pymc/distributions/multivariate.py
```diff
@@ -1443,6 +1461,28 @@ def helper_deterministics(cls, n, packed_chol):
...
+        return packed_dim_name, tri_coords
+
+    @classmethod
+    def _register_new_coords_with_model(cls, name, value):
+        mod = pm.modelcontext(None)
+        mod.coords[name] = value
+        mod.dim_lengths[name] = pt.TensorConstant(pt.lscalar, np.array(len(value)))
```
ricardoV94, line 1484: This must be a method already in Model no?

pymc/distributions/multivariate.py
```diff
@@ -1413,14 +1413,32 @@ class LKJCholeskyCov:
     """
 
     def __new__(cls, name, eta, n, sd_dist, *, compute_corr=True, store_in_trace=True, **kwargs):
+        dims = kwargs.pop("dims", None)
+
+        if dims is not None:
```
ricardoV94, line 1418: Not sure this should be done by default. If users want to resize the model they will have to know there is a special dims they also need to update

PR #6830, opened 2023-07-17
Bump PyTensor dependency

ricardoV94:
Bumps PyTensor dependency to 2.14.1

This PyTensor release reverts dynamic broadcasting in Elemwise. Code that started to rely on it will fail, hence why it is marked as a major PyMC release as well.

We also keep shape dependencies in the original graph, so some logic that expected eager Constants had to be tweaked.

Finally, no more automatic support shape inference for Multivariate RVs. The thing we had going on wasn't robust and made errors cryptic in the cases where it didn't apply. This includes dropping support for Multivariate CustomDist not implemented with via `dist`

----
:books: Documentation preview :books:: https://pymc--6830.org.readthedocs.build/en/6830/

### DIFF
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -87,7 +87,6 @@ __all__ = [
     "generator",
     "convert_observed_data",
     "compile_pymc",
-    "constant_fold",
 ]
 
 
--- a/conda-envs/environment-dev.yml
+++ b/conda-envs/environment-dev.yml
@@ -14,7 +14,7 @@ dependencies:
 - numpy>=1.15.0
 - pandas>=0.24.0
 - pip
-- pytensor>=2.12.0,<2.13
+- pytensor>=2.14.1,<2.15
 - python-graphviz
 - networkx
 - scipy>=1.4.1
--- a/conda-envs/environment-docs.yml
+++ b/conda-envs/environment-docs.yml
@@ -12,7 +12,7 @@ dependencies:
 - numpy>=1.15.0
 - pandas>=0.24.0
 - pip
-- pytensor>=2.12.0,<2.13
+- pytensor>=2.14.1,<2.15
 - python-graphviz
 - scipy>=1.4.1
 - typing-extensions>=3.7.4
--- a/conda-envs/environment-test.yml
+++ b/conda-envs/environment-test.yml
@@ -17,7 +17,7 @@ dependencies:
 - numpy>=1.15.0
 - pandas>=0.24.0
 - pip
-- pytensor>=2.12.0,<2.13
+- pytensor>=2.14.1,<2.15
 - python-graphviz
 - networkx
 - scipy>=1.4.1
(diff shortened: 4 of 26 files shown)


ricardoV94:
Had to pin arviz from above, as it was installing on the python 3.8 and failing

setup.py
OriolAbril, line 1: To drop python 3.9 support you need to change the `python_requires` key at line 73/74

twiecki:
Let's just drop 3.9 too.

ricardoV94:
> Let's just drop 3.9 too.

3.9 is supported by Numpy until next January. I think Oriol meant 3.8 in the comment he left, if that's what led you to say that?

OriolAbril:
I did mean 3.8, sorry

ricardoV94:
Last PyTensor release broke something on VI @ferrine.

My guess it has to do with one of these: 
* https://github.com/pymc-devs/pytensor/pull/280
* https://github.com/pymc-devs/pytensor/pull/345

Looking at the `step_fn` I see one extra `NormalRV` out of nowhere. And the first evaluation doesn't seem deterministic anymore either, probably because the `replacements` are introducing a new RV that didn't use to be introduced (and which is not properly seeded by the logic in `compile_pymc`?

Any chance you can take a look?

pymc/pytensorf.py
```diff
@@ -87,7 +87,6 @@
     "generator",
     "convert_observed_data",
     "compile_pymc",
-    "constant_fold",
```
ferrine, line 90: I saw it is imported in some other modules but is removed from `__all__`

ferrine: What it is used for?

ricardoV94: It's used to contant fold a graph, usually shape graphs (PyMC refuses to have RVs in logp graphs, even if they are just used for shapes). I import it manually, I don't think it should be available at the `pymc` root level, which everything in `__all__` is

ricardoV94:
Took 3 releases, but CI is passing! :)

ferrine:
I'm good on the vi side, anything else?

ferrine (approved):

ricardoV94:
Thanks @ferrine! Also with the help debugging!

I requested another review for the non-VI changes. Will wait a bit for feedback before merging.

michaelosthege:
major as in PyMC v6 ?

I'm just on the phone until Sunday, so I can't really check the diff. (Don't wait for mel

However, I'd like to request an example of a line of code that will fail and how to fix it to be included in the release notes.

ricardoV94:
No, major as in 

> major as in PyMC v6 ?

No, major as in PyMC v5.X.0

We are not doing "major, major" releases unless the PyMC API / dependencies change drastically. 
I don't have a better name for these, so I've been calling them major. Apologies for the confusion

> However, I'd like to request an example of a line of code that will fail and how to fix it to be included in the release notes.

We are reintroducing static broadcasting-only again with the PyTensor bump, so for example broadcasting with `MutableData("x", np.ones(1, 10))` will fail, unless you write `MutableData("x", np.ones(1, 10), shape=(1, None))`.

All other needed changes were internally-only and shouldn't affect users. Mostly, I imagine it will break custom defined MultivariateRVs that relied on the old automatic shape inference..

Issue #388, opened 2013-11-15
Theano requirement for python 3


kyleam:
The theano 0.6.0rc3 requirement is only accurate for python 2 because the theano release in PyPI doesn't support python 3. If the latest release from PyPI is installed under python 3, it results in the import error below. This can be corrected by installing from the theano git repo. Adding `-e git+https://github.com/Theano/Theano.git#egg=Theano` to requirements.txt would fix this (when using `pip -r`), but it would break reading the requirements.txt in as a list to setup.py `install_requires`. With pymc3 in alpha, I don't imagine this is much of an issue at the moment, but, if you all think it's worthwhile, I'll have a shot at getting it running by tweaking setup.py to use the `dependency_links` argument.

```
In [1]: import pymc as pm
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-290dd01c5432> in <module>()
----> 1 import pymc as pm

/home/kyle/src/python/pymc/pymc/__init__.py in <module>()
      1 __version__ = "3.0"
      2
----> 3 from .core import *
      4 from .distributions import *
      5 from .math import *

[...]

/usr/lib/python3.3/site-packages/theano/gof/op.py in <module>()
     23 from theano.misc.windows import call_subprocess_Popen
     24
---> 25 from . import cc
     26 from . import graph
     27 from . import utils

ImportError: cannot import name cc
```

twiecki:
Thanks for pointing that out. Definitely an issue that should be resolved. If you wanted to take a stab at it that'd be very much appreciated.

It also wouldn't be a big deal to just duplicate requirements.txt and setup.py deps.

kyleam:
What is the current use of requirements.txt? Considering it doesn't contain any stricter specifications than`install_requires`, perhaps there isn't much point to having it? I'm thinking along the lines of [this](https://caremad.io/blog/setup-vs-requirement/).

twiecki:
Agreed. I did not know about this so that seems like the way to go. Can you do a PR applying this change?

kyleam:
Sure. The pull request now has this change.

PR #6831, opened 2023-07-17
add ICARRV and ICAR

daniel-saunders-phil:
This PR adds support for intrinsic conditional autoregressive (ICAR) models to PyMC. The ICAR is a special case of the CAR distribution that arises when the $\alpha$ parameter equals $1$ and we add a zero-sum constraint. There are two big advantages of ICAR over CAR. 

First, computing the logp likelihood is much faster, especially on large datasets. The CAR distribution involves either inverting a large precision matrix or a variety of eigenvalue strategies. As the dataset gets larger, those operations become quite expensive, even with the sparsity tricks PyMC currently employs. In some benchmarking I was doing I found it takes nearly 60 seconds to find the logp of a dataset with 6000 observations. ICAR exploits a pairwise difference calculation which grows basically linearly with the size of the dataset.

image

Second, the zero-sum constraint can also help improve the identifiability of autoregressive models. Suppose you're working on a spatial statistics problem with the 50 US states. You might want a global intercept in your model plus 50 spatial random effects. With CAR, this leaves you overparameterized. If the sampler increases the mean value of the 50 spatial effects, then it can decrease the value of the global intercept to get the same overall intercept. The vector of random effects in an ICAR distribution is forced to sum to zero, which allows the sampler to independently adjust the global intercept.

Finally, ICAR often serves as a sub-component of larger models, like the BYM model. Providing a convenient API can make easy to build up to the more complex models.

More information on the design of this ICAR distribution can be found in a [stan case study](https://mc-stan.org/users/documentation/case-studies/icar_stan.html)


bwengals:
RE checking if W contains ints.  If W doesn't contain ints and the code runs but gives wrong results it'd def be good to check.  If the code errors and says something informative then there's no need to check I think

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2218,174 @@ def logp(value, mu, W, alpha, tau):
...
 
+class ICARRV(RandomVariable):
+    name = "icar"
+    ndim_supp = 1
+    ndims_params = [2, 1, 1, 0, 0, 0]
+    dtype = "floatX"
+    _print_name = ("ICAR", "\\operatorname{ICAR}")
+
+    def __call__(self, W, node1, node2, N, sigma, centering_strength, size=None, **kwargs):
+        return super().__call__(W, node1, node2, N, sigma, centering_strength, size=size, **kwargs)
```
bwengals: I think it'd be better to use a different word than "centering", since that's usually more about centered vs non-centered parameterizations.  Maybe "mean_zero_strength" or something similar instead?

daniel-saunders-phil:
> RE checking if W contains ints. If W doesn't contain ints and the code runs but gives wrong results it'd def be good to check. If the code errors and says something informative then there's no need to check I think

Okay that sounds good. Looks like the code fails silently when they pass an adjacency matrix with a non-zero, non-one in it. People might be inclined to do this, some literature in the CAR area discusses using edge weights for more flexibility. But the pairwise difference formula cannot handle edge weights. I'll add a check that requires every entry to be 0 or 1.

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+    """
+
+    rv_op = icar
+
+    @classmethod
+    def dist(cls, W, sigma=1, zero_sum_strength=0.001, **kwargs):
+        # check that adjacency matrix is two dimensional,
+        # square,
+        # symmetrical
+        # and composed of 1s or 0s.
```
bwengals: this block of comments doesn't add too much to the code doing the checks below, dont think its necessary

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+        if not W.shape[0] == W.shape[1]:
+            raise ValueError("W must be a square matrix")
+
+        if not np.allclose(W.T, W):
+            raise ValueError("W must be a symmetric matrix")
+
+        if np.any((W != 0) & (W != 1)):
+            raise ValueError("W must be composed of only 1s and 0s")
+
+        # convert adjacency matrix to edgelist representation
```
bwengals, line 2387: here though it might be good to expand on what an edgelist representation is?

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+
+        node1 = pt.as_tensor_variable(node1, dtype=int)
+        node2 = pt.as_tensor_variable(node2, dtype=int)
+
+        W = pt.as_tensor_variable(W, dtype=int)
+
+        N = pt.shape(W)[0]
+        N = pt.as_tensor_variable(N)
+
+        # check on sigma
```
bwengals: this comment and the comment check on centering strength are out of date now

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+        return super().dist([W, node1, node2, N, sigma, zero_sum_strength], **kwargs)
+
+    def moment(rv, size, W, node1, node2, N, sigma, zero_sum_strength):
+        return pt.zeros(N)
+
+    def logp(value, W, node1, node2, N, sigma, zero_sum_strength):
+        pairwise_difference = (-1 / (2 * sigma**2)) * pt.sum(
+            pt.square(value[node1] - value[node2])
+        )
+        soft_center = (
```
bwengals: same as before RE wording, "center" vs zero sum.  I think the latter is better and is more consistent with `zero_sum_strength`

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+    covariance between neighboring areas on large datasets. It is a special case
+    of the :class:`~pymc.CAR` distribution where alpha is set to 1.
+
+    The log probability density function is
+
+    .. math::
+        f(\\phi| W,\\sigma) =
+          -\frac{1}{2\\sigma^{2}} \\sum_{i\\sim j} (\\phi_{i} - \\phi_{j})^2 -
+          \frac{1}{2}*\frac{\\sum_{i}{\\phi_{i}}}{0.001N}^{2} - \\ln{\\sqrt{2\\pi}} -
+          \\ln{0.001N}
```
bwengals: why are some backslashes twice and others once?

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+        raise NotImplementedError("Cannot sample from ICAR prior")
+
+
+icar = ICARRV()
+
+
+class ICAR(Continuous):
+    r"""
+    The intrinsic conditional autoregressive prior. It is primarily used to model
+    covariance between neighboring areas on large datasets. It is a special case
```
bwengals: > on large datasets

Or small datasets!

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+        f(\\phi| W,\\sigma) =
+          -\frac{1}{2\\sigma^{2}} \\sum_{i\\sim j} (\\phi_{i} - \\phi_{j})^2 -
+          \frac{1}{2}*\frac{\\sum_{i}{\\phi_{i}}}{0.001N}^{2} - \\ln{\\sqrt{2\\pi}} -
+          \\ln{0.001N}
+
+    The first term represents the spatial covariance component. Each $\\phi_{i}$ is penalized
+    based on the square distance from each of its neighbors. The notation $i\\sim j$
+    indicates a sum over all the neighbors of $\\phi_{i}$. The last three terms are the
+    Normal log density function where the mean is zero and the standard deviation is
+    $N * 0.001$ (where N is the length of the vector $\\phi$). This component imposed the zero-sum
```
bwengals:
```suggestion
    $N * 0.001$ (where N is the length of the vector $\\phi$). This component imposes a zero-sum
```

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+    constraint by finding the sum of the vector $\\phi$ and penalizing based on its
+    distance from zero.
+
+    Parameters
+    ----------
+    W : ndarray of int
+        Symmetric adjacency matrix of 1s and 0s indicating adjacency between elements.
+        Must pass either W or both node1 and node2.
```
bwengals:
```suggestion
```

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+            sigma = pm.Exponential('sigma',1)
+            phi = pm.ICAR('phi',W=W,sigma=sigma)
+
+            mu = phi
+
+        # non-centered parameterization
+
+        with pm.Model():
+            sigma = pm.Exponential('sigma',1)
+            phi = pm.ICAR('phi',W=W)
```
bwengals:
```suggestion
        W = np.array([
            [0, 1, 0, 1],
            [1, 0, 1, 0],
            [0, 1, 0, 1],
            [1, 0, 1, 0]
        ])

        # centered parameterization
        with pm.Model():
            sigma = pm.Exponential('sigma', 1)
            phi = pm.ICAR('phi', W=W, sigma=sigma)
            mu = phi

        # non-centered parameterization
        with pm.Model():
            sigma = pm.Exponential('sigma', 1)
            phi = pm.ICAR('phi', W=W)
            mu = sigma * phi
```

adjusted some formatting here, mostly putting a space after comma

pymc/distributions/multivariate.py
```diff
@@ -2218,6 +2219,169 @@ def logp(value, mu, W, alpha, tau):
...
+
+    sigma : scalar, default 1
+        Standard deviation of the vector of phi's. Putting a prior on sigma
+        will result in a centered parameterization. In most cases, it is
+        preferable to use a non-centered parameterization by using the default
+        value and multiplying the resulting phi's by sigma. See the example below.
+
+    zero_sum_strength : scalar, default 0.001
+        Controls how strongly to enforce the zero-sum constraint. It sets the
+        standard deviation of a normal density function with mean zero.
```
bwengals: Could add a bit of detail here, like, "It puts an additional normal prior on the sum of the phi, such that the sum is normally distributed with mean zero and a small standard deviation, whose value is `zero_sum_strength`.  Maybe `zero_sum_stdev` is a clearer name then?

bwengals (changes requested):
Looks good!  mostly cosmetic stuff

pymc/distributions/multivariate.py
daniel-saunders-phil: Not sure, I wrote them with single slashes. Maybe one of the pre-commit checks added them by mistake?

daniel-saunders-phil:
Hi @ricardoV94 , I've been working on adding a shape inference function since the change in #388 . I thought `supp_shape_from_ref_param_shape` would be my best bet. In the ICAR, the output shape is always either the length or width of `W`, the adjacency matrix. I thought it was similar to the multivariate normal case where the length or width of the covariance matrix gives the output shape.

Continuous integration failed though because it couldn't import `supp_shape_from_ref_param_shape`. I also couldn't get the test to run locally for the same reason.

Is this related to releases? I assumed the continuous integration would run the latest pytensor but now I'm not so sure. Otherwise, do you have suggestions for how I should handle shape inference?

bwengals (approved):

bwengals:
Looks great!

fonnesbeck:
Thanks, @daniel-saunders-phil !

PR #6834, opened 2023-07-19
Logprob derivation for switch-encoding graphs

shreyas3156:
**What is this PR about?**
This draft PR consists of the headway made in inferring the log-probability of switch-encoding expressions for the following cases:

1) when both branches have an encoding constant,
```python
x = pt.random.normal(0.3)
y = pt.switch(x > 0.5, 1, 2)
```

2) when one of the two branches is measurable.
```python
x = pt.random.normal(0.3)
y = pt.switch(x > 0.5, 1, x)
```

The pending challenges with the second case above include the following:

1. Add calculation for the logprob when one of the branches is measurable,
2. Get the `pt.invert` to go through the IR rewrite and convert to its Measurable counterpart.

@ricardoV94 @larryshamalama

### DIFF
--- a/pymc/logprob/censoring.py
+++ b/pymc/logprob/censoring.py
@@ -248,11 +248,12 @@ class MeasurableSwitchEncoding(MeasurableElemwise):
     """A placeholder used to specify the log-likelihood for a encoded RV sub-graph."""
 
     valid_scalar_types = (Switch,)
-    # number of measurable branches to facilitate correct logprob calculation
-    measurable_branches = 0
 
-
-measurable_switch_encoding = MeasurableSwitchEncoding(scalar_switch)
+    def __init__(self, measurable_branches):
+        super().__init__(scalar_switch)
+        self.__props__ = super().__props__ + ("measurable_branches",)
+        self.measurable_branches = measurable_branches
+        # number of measurable branches to facilitate correct logprob calculation
 
 
 @node_rewriter(tracks=[switch])
@@ -285,6 +286,9 @@ def find_measurable_switch_encoding(
     [base_var] = rv_map_feature.request_measurable([switch_condn.owner.inputs[0]])
     if base_var.dtype.startswith("int"):
         return None
+
+    # default number of measurable branches is zero
+    measurable_switch_encoding = MeasurableSwitchEncoding(measurable_branches=0)
 
     # Maximum one branch allowed to be measurable
     if len(measurable_comp_list) > 1:
@@ -339,7 +343,7 @@ def switch_encoding_logprob(op, values, *inputs, **kwargs):
             ),
         )
     else:
-        base_var = components[1]  # there needs to be a better way to obtain the base variable.
+        base_var = components[1]
 
         logp_first_branch = _logprob_helper(switch_condn, pt.as_tensor(np.array(True)), **kwargs)
 
--- a/pymc/logprob/censoring.py
+++ b/pymc/logprob/censoring.py
@@ -279,6 +279,10 @@ def find_measurable_switch_encoding(
     if rv_map_feature.request_measurable([switch_condn]) != [switch_condn]:
         return None
 
+    [base_var] = rv_map_feature.request_measurable([switch_condn.owner.inputs[0]])
+    if base_var.dtype.startswith("int"):
+        return None
+
     # Maximum one branch allowed to be measurable
     if len(measurable_comp_list) > 1:
         return None
--- a/pymc/logprob/censoring.py
+++ b/pymc/logprob/censoring.py
@@ -265,37 +265,34 @@ def find_measurable_switch_encoding(
     switch_condn, *components = node.inputs
 
     # broadcasting of switch condition is not supported
-    if switch_condn.ndim != 0:
-        if any(switch_condn.type.broadcastable):
-            return None
+    if switch_condn.type.broadcastable != node.outputs[0].type.broadcastable:
+        return None
 
     if rv_map_feature.request_measurable([switch_condn]) != [switch_condn]:
         return None
-        # this automatically checks the measurability of the switch condition and converts switch to MeasurableSwitch
+    # this automatically checks the measurability of the switch condition and converts switch to MeasurableSwitch
 
-    measurable_comp_idx = next(
-        (
-            idx
-            for idx, component in enumerate(components)
-            if check_potential_measurability([component], valued_rvs)
-        ),
-        -1,
-    )
+    measurable_comp_list = [
+        idx
+        for idx, component in enumerate(components)
+        if check_potential_measurability([component], valued_rvs)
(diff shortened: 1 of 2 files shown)


pymc/logprob/censoring.py
```diff
@@ -237,3 +240,149 @@ def round_logprob(op, values, base_rv, **kwargs):
...
+        return None  # pragma: no cover
+
+    valued_rvs = rv_map_feature.rv_values.keys()
+
+    switch_condn, *components = node.inputs
+
+    # broadcasting of switch condition is not supported
+    if switch_condn.ndim != 0:
+        if any(switch_condn.type.broadcastable):
+            return None
```
ricardoV94: This check is faulty. To check for broadcasting you can do `switch_cond.type.broadcastable != node.outputs[0].type.broadcastable` instead

pymc/logprob/censoring.py
```diff
@@ -237,3 +240,149 @@ def round_logprob(op, values, base_rv, **kwargs):
...
+        # this automatically checks the measurability of the switch condition and converts switch to MeasurableSwitch
+
+    measurable_comp_idx = next(
+        (
+            idx
+            for idx, component in enumerate(components)
+            if check_potential_measurability([component], valued_rvs)
+        ),
+        -1,
+    )
```
ricardoV94: Not sure what's going on here, why not just a list comprehension?

pymc/logprob/censoring.py
```diff
@@ -237,3 +240,149 @@ def round_logprob(op, values, base_rv, **kwargs):
...
+
+    # If at least one of the branches is measurable
+    if measurable_comp_idx != -1:
+        measurable_component = components[measurable_comp_idx]
+
+        # broadcasting of the measurable component is not supported
+        if measurable_component.ndim != 0 and any(measurable_component.type.broadcastable):
+            return None
```
ricardoV94: Same as with the switch condition

pymc/logprob/censoring.py
```diff
@@ -237,3 +240,149 @@ def round_logprob(op, values, base_rv, **kwargs):
...
+
+        if not compare_measurability_source([switch_condn, measurable_component], valued_rvs):
+            return None
+
+        measurable_inputs = rv_map_feature.request_measurable(components)
+        # Maximum one branch allowed to be measurable
+        if len(measurable_inputs) > 1:
+            return None
```
ricardoV94: This has to be checked some other way than through `request_measurable`, as one input could become measurable before the other, and the check `>1` would not be triggered accidentally

pymc/logprob/censoring.py
```diff
@@ -237,3 +240,149 @@ def round_logprob(op, values, base_rv, **kwargs):
...
+
+        if measurable_comp_idx == 0:
+            # changing the first branch of switch to always be the encoding
+            encoded_rv = measurable_switch_encoding.make_node(
+                pt.invert(switch_condn), *components[::-1]
+            ).default_output()
+            # FIXME: For graphs like y = pt.switch(x > 0.5, x, 0.3), they should be rewritten
+            #  to pt.switch(x <= 0.5, 0.3, x).
+            #  But the invert Op does not get converted to its Measurable counterpart.
```
ricardoV94: You have to create a `MeasurableInvert` directly. Otherwise you never told PyMC that this new invert has to be measurable.

pymc/logprob/censoring.py
shreyas3156: Using `next()` would help find the first index for which the check_potential_measurability() function returns True with the default value of -1.

But a list comprehension does make more sense because we can then check for the number of measurable components without calling `request_measurable()` on them.

pymc/logprob/censoring.py
```diff
@@ -237,3 +241,147 @@ def round_logprob(op, values, base_rv, **kwargs):
...
+        idx
+        for idx, component in enumerate(components)
+        if check_potential_measurability([component], valued_rvs)
+    ]
+
+    # Maximum one branch allowed to be measurable
+    if len(measurable_comp_list) > 1:
+        return None
+
+    # If at least one of the branches is measurable
```
larryshamalama, line 297: Any chance that there can be no measurable branches following the previous checks? Then `len(measurable_comp_list) == 1` necessarily?

pymc/logprob/censoring.py
```diff
@@ -237,3 +241,147 @@ def round_logprob(op, values, base_rv, **kwargs):
...
+)
+
+
+def compare_measurability_source(
+    inputs: Tuple[TensorVariable], valued_rvs: Container[TensorVariable]
+) -> bool:
+    ancestor_var_set = set()
+
+    # retrieve the source of measurability for all elements in 'inputs' separately.
+    for inp in inputs:
```
larryshamalama, line 375: Nice that you figured out a solution here :)

pymc/logprob/censoring.py
```diff
@@ -237,3 +241,147 @@ def round_logprob(op, values, base_rv, **kwargs):
...
+                ancestor_var.owner
+                and isinstance(ancestor_var.owner.op, RandomVariable)
+                and ancestor_var not in valued_rvs
+            ):
+                ancestor_var_set.add(ancestor_var)
+
+    return len(ancestor_var_set) == 1
+
+
+def walk_model(
```
larryshamalama, line 391: Probably okay to have this here temporarily. We could modify the other `walk_model` in the codebase such that we can use it to find `RandomVariable`s like in your use case.

IIRC, we have already two `walk_model` functions... not sure why

larryshamalama (commented):
Great start @shreyas3156 :) Some minor comments for now as you continue to make progress. Keep up the good work!

pymc/logprob/censoring.py
```diff
@@ -247,6 +248,8 @@ class MeasurableSwitchEncoding(MeasurableElemwise):
     """A placeholder used to specify the log-likelihood for a encoded RV sub-graph."""
 
     valid_scalar_types = (Switch,)
+    # number of measurable branches to facilitate correct logprob calculation
+    measurable_branches = 0
```
ricardoV94: This should be a property added during `__init__`. You must also add it as `__props__` so that PyTensor doesn't consider two Ops with the same inputs but different measurable_branches equivalent and tries to merge them.

Something like `self.__props__ = super().__props__ + ("measurable_branches",)`

ricardoV94:
Hey @shreyas3156 what's the status of the PR? Ready for review? There's some conflicts with main

shreyas3156:
> Hey @shreyas3156 what's the status of the PR? Ready for review? There's some conflicts with main

This was the first draft PR which was supposed to be a draft for the main [PR #6949 ](https://github.com/pymc-devs/pymc/pull/6949), which is ready for review. We can close this one.

PR #6835, opened 2023-07-20
Fixes for the McBackend adapter

michaelosthege:
**What is this PR about?**
This fixes two bugs that were reported by @thelogicalgrammar in https://github.com/michaelosthege/mcbackend/issues/93.

The first was a leftover of a previous refactoring of sampler stats, and the way how the `"tune"` stat is handled in PyMC.

**For the record:** Our current approach of passing information about tuning/not-tuning via sample stats is brittle. It's a per-sampler thing, but in practice it's actually a per-iteration thing from the PyMC, ArviZ and McBackend perspective. IMO we should consider taking it out of the step method attributes and stats, and instead pass `tune` as a parameter to the `astep` method.

However, this is out of scope for this PR, so I aligned with the current design and added `"tune"` to two step methods that didn't have it, much like Ricardo did for `Slice` in fee9a02c55763dea9dfe497ed509a3c05ad04972.

The second bug was just a small dtype thing for pickled stat objects.
I tested it locally with a ClickHouse server:

```python
def test_issue_93_b():
    seconds = np.linspace(0, 5)
    observations = np.random.normal(0.5 + np.random.uniform(size=3)[:, None] * seconds[None, :])
    with pm.Model(
        coords={
            "condition": ["A", "B", "C"],
        }
    ) as pmodel:
        x = pm.ConstantData("seconds", seconds, dims="time")
        a = pm.Normal("scalar")
        b = pm.Uniform("vector", dims="condition")
        pm.Deterministic("matrix", a + b[:, None] * x[None, :], dims=("condition", "time"))
        obs = pm.MutableData("obs", observations, dims=("condition", "time"))
        pm.Normal("L", pmodel["matrix"], observed=obs, dims=("condition", "time"))

    ch_client = clickhouse_driver.Client("localhost")
    backend = ClickHouseBackend(ch_client)
    with pmodel:
        pm.sample(trace=backend, tune=5, draws=7, discard_tuned_samples=False)
    assert idata.warmup_posterior.sizes["draw"] == 5
    assert idata.posterior.sizes["draw"] == 7
    pass
```

**Checklist**

## Major / Breaking Changes
- None

## New features
- None

## Bugfixes
- Two bugs in McBackend adapter were fixed

## Documentation
- None

## Maintenance
- All of PyMC's step methods now emit a `"tune"` stat

----
:books: Documentation preview :books:: https://pymc--6835.org.readthedocs.build/en/6835/

### DIFF
--- a/pymc/backends/mcbackend.py
+++ b/pymc/backends/mcbackend.py
@@ -221,9 +221,13 @@ def make_runmeta_and_point_fn(
                 (-1 if s is None else s)
                 for s in (shape or [])
             ]
+            dt = np.dtype(dtype).name
+            # Object types will be pickled by the ChainRecordAdapter!
+            if dt == "object":
+                dt = "str"
             svar = mcb.Variable(
                 name=sname,
-                dtype=np.dtype(dtype).name,
+                dtype=dt,
                 shape=sshape,
                 undefined_ndim=shape is None,
             )
--- a/pymc/backends/mcbackend.py
+++ b/pymc/backends/mcbackend.py
@@ -33,6 +33,7 @@ from pymc.step_methods.compound import (
     BlockedStep,
     CompoundStep,
     StatsBijection,
+    check_step_emits_tune,
     flat_statname,
     flatten_steps,
 )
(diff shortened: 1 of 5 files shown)


ricardoV94 (approved):
Not sure what's going on with the `"object"` vs `"str"` dtype. Otherwise LGTM

michaelosthege:
> Not sure what's going on with the `"object"` vs `"str"` dtype. Otherwise LGTM

thanks for the review!

To clarify this part: Sampler warnings are `SamplerWarning` objects, but as such they can't be stored in all backends. Therefore, the `ChainRecordAdapter` pickles them to a string, which can be stored in HDF5/ClickHouse/whatever.

PR #6845, opened 2023-08-04
Add icdf functions for Beta, Gamma, Chisquared and StudentT distributions

amyoshino:
**What is this PR about?**
Adds ICDF functions to Beta, Gamma, Chisquared and StudentT distributions.

Issue https://github.com/pymc-devs/pymc/issues/6612

This is a work in progress. The quantile functions depend on two main functions: 
- Inverse of the regularized incomplete beta function: `betaincinv`
- Inverse to the regularized lower incomplete gamma function: `gammaincinv`

Those are implemented in SciPy but not in Pytensor yet. When importing the two functions from `scipy.special` and plugging them in the desired formula, I get a `TypeError` like this below: 

`"TypeError: ufunc 'betaincinv' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''"`

image

It looks like the most straightforward way to make these icdf functions work is to implement them in Pytensor, in the same way the [`pt.betainc` has been implemented](https://github.com/pymc-devs/pytensor/blob/0b558d8e65ee6603b80be566d36a036b68a97ee0/pytensor/scalar/math.py#L1335)

Any comments about if this is the right path? If so I am happy to open an issue in Pytensor as a feature request, and possibly helping on its development if nobody else wants to take it. But if there is any ideas on how to make it simpler than that, let me know. 

References:
- SciPy betaincinv: https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.betaincinv.html
- SciPy gammaincinv: https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.gammaincinv.html
- ChiSquared: https://www.wolframalpha.com/input?i=chi+squared+distribution+quantile+function
- Beta: https://www.wolframalpha.com/input?i=beta+distribution+quantile+function
- Gamma: https://www.wolframalpha.com/input?i=gamma+distribution+quantile+function
- StudentT: https://www.wolframalpha.com/input?i=student+distribution+quantile+function

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6845.org.readthedocs.build/en/6845/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2088,7 +2088,7 @@ class Cauchy(Continuous):
     def icdf(value, alpha, beta):
         res = alpha + beta * pt.tan(np.pi * (value - 0.5))
         res = check_icdf_value(res, value)
-        return check_parameters(
+        return check_icdf_parameters(
             res,
             beta > 0,
             msg="beta > 0",
@@ -2172,7 +2172,7 @@ class HalfCauchy(PositiveContinuous):
     def icdf(value, loc, beta):
         res = loc + beta * pt.tan(np.pi * (value) / 2.0)
         res = check_icdf_value(res, value)
-        return check_parameters(
+        return check_icdf_parameters(
             res,
             beta > 0,
             msg="beta > 0",
@@ -2591,7 +2591,7 @@ class Weibull(PositiveContinuous):
     def icdf(value, alpha, beta):
         res = beta * (-pt.log(1 - value)) ** (1 / alpha)
         res = check_icdf_value(res, value)
-        return check_parameters(
+        return check_icdf_parameters(
             res,
             alpha > 0,
             beta > 0,
@@ -3151,7 +3151,7 @@ class Triangular(BoundedContinuous):
             upper - np.sqrt((upper - lower) * (upper - c) * (1 - value)),
         )
         res = check_icdf_value(res, value)
-        return check_parameters(
+        return check_icdf_parameters(
             res,
             lower <= c,
             c <= upper,
@@ -3254,7 +3254,7 @@ class Gumbel(Continuous):
     def icdf(value, mu, beta):
         res = mu - beta * pt.log(-pt.log(value))
         res = check_icdf_value(res, value)
-        return check_parameters(
+        return check_icdf_parameters(
             res,
             beta > 0,
             msg="beta > 0",
@@ -3472,7 +3472,7 @@ class Logistic(Continuous):
     def icdf(value, mu, s):
         res = mu + s * pt.log(value / (1 - value))
         res = check_icdf_value(res, value)
-        return check_parameters(
+        return check_icdf_parameters(
             res,
             s > 0,
             msg="s > 0",
@@ -3822,7 +3822,7 @@ class Moyal(Continuous):
     def icdf(value, mu, sigma):
         res = sigma * -pt.log(2.0 * pt.erfcinv(value) ** 2) + mu
         res = check_icdf_value(res, value)
-        return check_parameters(
+        return check_icdf_parameters(
             res,
             sigma > 0,
             msg="sigma > 0",
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -33,7 +33,7 @@ from pytensor.graph.op import Op
 from pytensor.raise_op import Assert
 from pytensor.tensor import gammaln
 from pytensor.tensor.extra_ops import broadcast_shape
-from pytensor.tensor.math import tanh
+from pytensor.tensor.math import betaincinv, gammaincinv, tanh
 from pytensor.tensor.random.basic import (
     BetaRV,
     _gamma,
@@ -1227,6 +1227,16 @@ class Beta(UnitContinuous):
             msg="alpha > 0, beta > 0",
         )
 
+    def icdf(value, alpha, beta):
+        res = betaincinv(alpha, beta, value)
+        res = check_icdf_value(res, value)
+        return check_icdf_parameters(
+            res,
+            alpha > 0,
+            beta > 0,
+            msg="alpha > 0, beta > 0",
+        )
+
 
 class KumaraswamyRV(RandomVariable):
     name = "kumaraswamy"
@@ -1872,6 +1882,21 @@ class StudentT(Continuous):
             msg="nu > 0, sigma > 0",
         )
 
+    def icdf(value, nu, mu, sigma):
+        res = pt.switch(
+            pt.lt(value, 0.5),
+            -pt.sqrt(nu) * pt.sqrt((1.0 / betaincinv(nu * 0.5, 0.5, 2.0 * value)) - 1.0),
+            pt.sqrt(nu) * pt.sqrt((1.0 / betaincinv(nu * 0.5, 0.5, 2.0 * (1 - value))) - 1.0),
+        )
+        res = mu + res * sigma
+        res = check_icdf_value(res, value)
+        return check_icdf_parameters(
+            res,
+            nu > 0,
+            sigma > 0,
+            msg="nu > 0, sigma > 0",
+        )
+
 
 class Pareto(BoundedContinuous):
     r"""
@@ -2271,6 +2296,16 @@ class Gamma(PositiveContinuous):
             pt.log(pt.gammainc(alpha, beta * value)),
         )
         return check_parameters(res, 0 < alpha, 0 < beta, msg="alpha > 0, beta > 0")
+
+    def icdf(value, alpha, scale):
+        res = scale * gammaincinv(alpha, value)
+        res = check_icdf_value(res, value)
+        return check_icdf_parameters(
+            res,
+            alpha > 0,
+            scale > 0,
+            msg="alpha > 0, beta > 0",
+        )
 
 
 class InverseGamma(PositiveContinuous):
--- a/tests/distributions/test_continuous.py
+++ b/tests/distributions/test_continuous.py
@@ -411,6 +411,13 @@ class TestMatchesScipy:
             lambda value, alpha, beta: st.beta.logcdf(value, alpha, beta),
         )
 
+    def test_beta_icdf(self):
+        check_icdf(
+            pm.Beta,
+            {"alpha": Rplus, "beta": Rplus},
+            lambda q, alpha, beta: st.beta.ppf(q, alpha, beta),
+        )
+
     def test_kumaraswamy(self):
         # Scipy does not have a built-in Kumaraswamy
         def scipy_log_pdf(value, a, b):
@@ -557,6 +564,13 @@ class TestMatchesScipy:
             lambda value, nu, mu, sigma: st.t.logcdf(value, nu, mu, sigma),
         )
 
+    def test_studentt_icdf(self):
+        check_icdf(
+            pm.StudentT,
+            {"nu": Rplusbig, "mu": R, "sigma": Rplusbig},
+            lambda q, nu, mu, sigma: st.t.ppf(q, nu, mu, sigma),
+        )
+
     def test_cauchy(self):
         check_logp(
             pm.Cauchy,
@@ -621,6 +635,13 @@ class TestMatchesScipy:
             Rplus,
             {"alpha": Rplusbig, "beta": Rplusbig},
             lambda value, alpha, beta: st.gamma.logcdf(value, alpha, scale=1.0 / beta),
+        )
+
+    def test_gamma_icdf(self):
+        check_icdf(
+            pm.Gamma,
+            {"alpha": Rplusbig, "beta": Rplusbig},
+            lambda q, alpha, beta: st.gamma.ppf(q, alpha, scale=1.0 / beta),
         )
 
     def test_inverse_gamma_logp(self):


amyoshino:
@ricardoV94 now that https://github.com/pymc-devs/pytensor/pull/502 is merged, we are able to complete this PR.

The tests are failing because `betaincinv` and `gammaincinv` cannot be imported from `pytensor.tensor.math`, I guess it is matter of time to pytensor to be updated. I will restart the checks when it happens. 

Let me know what you think.

ricardoV94:
They should be imported directly from pytensor.tensor.

pymc.math has a subset of hand-picked functions for users

ricardoV94:
You can see which PyTensor version is being installed in the CI.

pymc/distributions/continuous.py
```diff
@@ -1235,6 +1235,16 @@ def logcdf(value, alpha, beta):
             msg="alpha > 0, beta > 0",
         )
 
+    def icdf(value, alpha, beta):
+        res = betaincinv(alpha, beta, value)
+        res = check_icdf_value(res, value)
+        return check_parameters(
```
ricardoV94: Should use https://github.com/pymc-devs/pymc/blob/main/pymc%2Fdistributions%2Fdist_math.py#L76

And in the other functions 

This is because the default `check_parameters` was developed for logp in mind and pymc replaces it by -inf sometimes, which doesn't make sense for icdf expressions

ricardoV94 (changes requested):

pymc/distributions/continuous.py
```diff
@@ -1875,6 +1885,22 @@ def logcdf(value, nu, mu, sigma):
             msg="nu > 0, sigma > 0",
         )
 
+    def icdf(value, nu, mu, sigma):
+        _, sigma = get_tau_sigma(sigma=sigma)
```
ricardoV94: This is useless?

pymc/distributions/continuous.py
```diff
@@ -2271,6 +2297,17 @@ def logcdf(value, alpha, scale):
         )
         return check_parameters(res, 0 < alpha, 0 < beta, msg="alpha > 0, beta > 0")
 
+    def icdf(value, alpha, scale):
+        beta = pt.reciprocal(scale)
```
ricardoV94: If beta is just used for the assert, could instead assert scale and avoid the reciprocal

pymc/distributions/continuous.py
```diff
@@ -2439,6 +2476,10 @@ def __new__(cls, name, nu, **kwargs):
     def dist(cls, nu, **kwargs):
         return Gamma.dist(alpha=nu / 2, beta=1 / 2, **kwargs)
 
+    def icdf(value, nu):
```
ricardoV94: Not needed. This Class just returns a Gamma variable so the icdf will work already like this.

tests/distributions/test_continuous.py
```diff
@@ -332,6 +332,13 @@ def test_chisquared_logcdf(self):
             lambda value, nu: st.chi2.logcdf(value, df=nu),
         )
 
+    def test_chisquared_icdf(self):
```
ricardoV94: No need to test, the Gamma is enough

tests/distributions/test_continuous.py
```diff
@@ -562,6 +576,18 @@ def test_studentt_logcdf(self):
             lambda value, nu, mu, sigma: st.t.logcdf(value, nu, mu, sigma),
         )
 
+    def test_studentt_icdf(self):
+        check_icdf(
+            pm.StudentT,
+            {"nu": Rplusbig, "mu": R, "lam": Rplus},
+            lambda q, nu, mu, lam: st.t.ppf(q, nu, mu, lam**-0.5),
+        )
+        check_icdf(
```
ricardoV94: No need for both parametrization tests. That's something we used to do for logp but makes no sense really, unless the expression we used depended on the parametrization

ricardoV94 (commented):
Couple more comments. Looks great otherwise

ricardoV94:
Yeah it's strange it's still installing 2.18.4, even though 2.18.5 was released already some days ago. We may need to be just a bit more patient?

pymc/distributions/continuous.py
amyoshino: Ops, my bad! Actually I just realized that some of the other icdf functions I have implemented were using `check_parameters` instead of `check_icdf_parameters`. I've changed them in the commit https://github.com/pymc-devs/pymc/pull/6845/commits/be24e627a698659037db8cb8affb9c82ad24d2ee, but if you think those are out of topic, I can revert it and create a new PR only for these small changes.

pymc/distributions/continuous.py
amyoshino: It looks like so, I kept it earlier because the logcdf function above it also uses it in the same way. 
Because we are testing from tau as well in the test suite, I thought it would be part of the tau/sigma conversion, or that it initialize sigma = 1 whenever neither sigma and tau are not passed as arguments. 
However, when I take it out, the tests pass anyways even for non initialized arguments, so I think it is useless.
Removed in the commit https://github.com/pymc-devs/pymc/pull/6845/commits/be24e627a698659037db8cb8affb9c82ad24d2ee

pymc/distributions/continuous.py
amyoshino: Done! Even though checking for scale, I kept beta on the user facing message, because this is the parameterization used in the documentation. Fixed in https://github.com/pymc-devs/pymc/pull/6845/commits/be24e627a698659037db8cb8affb9c82ad24d2ee

pymc/distributions/continuous.py
amyoshino: Very interesting!! Removed that! Thanks for pointing this out. 
Fixed in https://github.com/pymc-devs/pymc/pull/6845/commits/be24e627a698659037db8cb8affb9c82ad24d2ee

tests/distributions/test_continuous.py
amyoshino: Forgot to change it in https://github.com/pymc-devs/pymc/commit/be24e627a698659037db8cb8affb9c82ad24d2ee. Fixed in https://github.com/pymc-devs/pymc/pull/6845/commits/fdfe4a1de8dc59293e8ab1e935780f5f7257f2e3.

tests/distributions/test_continuous.py
amyoshino: Pardon my lack of attention, I forgot to add this as well in previous commits. Fixed in https://github.com/pymc-devs/pymc/pull/6845/commits/e045748f94a51bf1d7905ff7526bbe82f03577e9

amyoshino:
> Yeah it's strange it's still installing 2.18.4, even though 2.18.5 was released already some days ago. We may need to be just a bit more patient?

Right! That's what I thought, I saw the pytenor release 4 days ago and thought it was a good experience to try it now and see if the update immediately applies. No rush on it, it will work soon.

pymc/distributions/continuous.py
ricardoV94: It's fine to do it here. Thanks for the fix

pymc/distributions/continuous.py
ricardoV94: Sounds good

pymc/distributions/continuous.py
ricardoV94: Can you just squash the commits related to the new functionality into a single commit and have the fix for the previous implementations as a separate commit?

You will have to force-push after that. Also if it's your first time backup your branch first!

amyoshino: I will try that! It will be an interesting experience.

amyoshino: After failing quite a bit, it is done!
Icdf implementations + tests in https://github.com/pymc-devs/pymc/pull/6845/commits/babc47923fd8272014be2c2f948ad917cfc60609
Fixing `check_parameters` --> `check_icdf_parameters` in https://github.com/pymc-devs/pymc/pull/6845/commits/c54cd1d05cfe9fabd913bcf1479dc6c730cd53b6

ricardoV94: Awesome. I see however there's a 3rd commit: "merge main". usually when keeping a clean git history in your PR you would update via pull rebase instead of the default pull merge.

I think this won't be a problem and the commit will drop when we rebase and merge but something I thought you might want to know

ricardoV94:
@maresb any idea why CI is not yet installing the last PyTensor? Was there an issue on the conda side?

pymc/distributions/continuous.py
amyoshino: Thanks for guiding me through this. I always used the "update branch" in the github UI at the top of the repository, whenever I start working or before the revision. 
The previous "merge main" was kind of a hassle when I tried to rebase to squash commits. It was a good learning lesson, and from now on I will shift to pull rebase.

ricardoV94: I'm still learning myself. The recommended source, if you want to learn it less randomly, is usually https://git-scm.com/book/en/v2

Some chapters are more useful than others if you already know you are working specifically with git + github

amyoshino: Nice!! Thank you for the resource! I definitely want to level up in git skills.

maresb:
It uploaded successfully on anaconda.org, so I restarted the tests.

amyoshino:
Interesting way to restart the tests! I searched for and the recommendation I found was to push an empty commit, but I think this is nicer.

ricardoV94:
You can also just ask a dev woth write permissions. We can restart them from GitHub interface

amyoshino:
Apologies, I have re-requested review by mistake.

amyoshino:
@ricardoV94 can you restart the tests? I don't think it is failing due to the changes we've made in this PR. The failed ones below seems to be unrelated with the icdf functions, let me know if I am mistaken. 

tests/distributions/test_mixture.py::TestHurdleMixtures::test_hurdle_zero_draws[dist-0.05-non_psi_args8] FAILED [ 97%]
tests/distributions/test_mixture.py::TestHurdleMixtures::test_hurdle_zero_draws[dist-0.3-non_psi_args9] FAILED [ 97%]
tests/distributions/test_mixture.py::TestHurdleMixtures::test_hurdle_zero_draws[dist-0.4-non_psi_args10] FAILED [ 97%]
tests/distributions/test_mixture.py::TestHurdleMixtures::test_hurdle_zero_draws[dist-0.9-non_psi_args11] FAILED [ 98%]

ricardoV94:
They could actually be related. The hurdle mixture uses truncated rvs for the non-zero part. And with these icdfs several truncated variables will switch from rejection sampling to inverse cdf sampling

amyoshino:
Got it, thanks for the pointers. I am still investigating what is the problem. 
I have spent few hours trying to debug but I still can't find what is going on. 
So far, drawing from `Gamma` works well, but inside the `Truncated` it does not. The same do not happen with other functions that I tried (Normal, Beta, HalfCauchy, InverseGamma - picked here some that have icdf functions and some that does not).

It looks like the error points to this numpy function, and now I am in the quest to find how to access `high` and `low` for the Truncated Gamma: https://github.com/numpy/numpy/blob/448e4dabb5e7d593af7d164a335dc7e44e17ec95/numpy/random/_generator.pyx#L1081

Here is what I got until now, and I will keep you posted about progress. But If by looking at it you have any clues, let me know. 

Drawing from Gamma, Beta, etc works as expected (I added some prints for debug purposes):
image

Drawing from Truncated only fails for Gamma:
image

amyoshino:
I have an update but it is still a mystery for me that it raises the `high - low overflow error` when passing only lower but not when passing lower and upper. And that it only happens with `Truncated(Gamma)` but not with `Truncated(Beta)` for example.

image

amyoshino:
I just narrowed the problem down to the following:
As stated by @ricardoV94, the hurdle mixture uses truncated rvs for the non-zero part. The failing tests are breaking when passing Gamma.dist to the Truncated.dist. 
- When I try to simulate the part it breaks the test with `Truncated.dist(Gamma.dist(alpha = 5, beta = 2), lower = 0.01).eval()` it  fails in the middle when trying to sample from a uniform distribution with `UniformRV.rng_fn`. It does so because the inputs for this function (rng_fn) in `tensor.random.op.py` is something like [array(7.64934096e-07), array(nan)] and when trying to generate a number with upper or lower as NaN, it gives the Overflow error we are observing (see [this link line 1081](https://github.com/numpy/numpy/blob/448e4dabb5e7d593af7d164a335dc7e44e17ec95/numpy/random/_generator.pyx#L1081))
- When trying another distribution with icdf implemented, like Beta.distr, and run `Truncated.dist(Beta.dist(alpha = 5, beta = 2), lower = 0.01).eval()` it works. In this case the arguments for the rng_fn function is something like [array(1.835008e-05), array(1.)], the number is then sampled from a uniform distribution and the function finish running. 
- Another experiment that went successful is when I pass lower and upper to `Truncated.dist(Gamma.dist(alpha = 5, beta = 2), lower = 0.01, upper = 1000).eval()` it finishes without any problems. 

I am clueless about why in Gamma the Op does not create the second argument in the case of Truncated(Gamma) but does for other Truncated distributions I have tested. I am stuck in the part where it fails when calling the function`getattr()` inside `pytensor.tensor.random.op.rng_fn()`, and running the debug mode in this goes to so many places that makes me lost. In summary, I did my best in trying to find where (and if) I can solve this problem within the scope of this PR, and can't proceed here without help with this.

ricardoV94:
Is the nan coming from `pm.logcdf(pm.Gamma.dist(alpha=2, beta=5), np.inf)?

ricardoV94:
Sounds like PyTensor gammainc C code does not have a special case for np.inf:

https://github.com/pymc-devs/pytensor/blob/19e1a982e5a2e9ba2bd9d708133abf102c6fb835/pytensor/scalar/c_code/gamma.c#L217-L223

In contrast this scipy has a special case? 
https://github.com/scipy/scipy/blob/4ecf5638c36c3a87b473056f5fffecfeddffd368/scipy/special/cephes/igam.c#L149-L151

amyoshino:
> `pm.logcdf(pm.Gamma.dist(alpha=2, beta=5), np.inf)`

Indeed, this code returns nan. I tried to fix it writing the special case for np.inf in the `gamma.c` function but it didn't work out yet. Maybe there are other places I need to change in order to make it work. 


`DEVICE double GammaP (double n, double x)
{                               /* --- regularized Gamma function P */
  if ((n <= 0) || (x < 0)) return NPY_NAN;  /* check the function arguments */
  if (x <=  0) return 0;        /* treat x = 0 as a special case */
  if (x < n+1) return _series(n, x) *exp(n *log(x) -x -logGamma(n));
  if (isinf(n)) {
    if (isinf(x)) return NPY_NAN;
    return 0;
  } 
  if (isinf(x)) return 1;
  return 1 -_cfrac(n, x) *exp(n *log(x) -x -logGamma(n));
}  /* GammaP() */

/*--------------------------------------------------------------------*/

DEVICE double GammaQ (double n, double x)
{                               /* --- regularized Gamma function Q */
  if ((n <= 0) || (x < 0)) return NPY_NAN;  /* check the function arguments */
  if (x <=  0) return 1;        /* treat x = 0 as a special case */
  if (x < n+1) return 1 -_series(n, x) *exp(n *log(x) -x -logGamma(n));
  if (isinf(n)) {
    if (isinf(x)) return NPY_NAN;
    return 1;
  } 
  if (isinf(x)) return 0;
  return _cfrac(n, x) *exp(n *log(x) -x -logGamma(n));
}  /* GammaQ() */`

Even with this, the `pm.logcdf(pm.Gamma.dist(alpha=2, beta=5), np.inf)` still returns nan.

ricardoV94:
You need to delete pytensor cache or bump the Op c_cache_version number. Try `pytensor-cache clear` or `pytensor cache-clear` I never remember. If clear doesn't work use purge

ricardoV94:
That PR should probably do a similar fix for the complementary gammainc function (gammaincc?)

ricardoV94:
@amyoshino PyTensor new release should be in conda soon. We will need to update the pin on PyTensor version here, as it was a "major" release (or in a separate PR if there are other things that need fix) 

https://github.com/pymc-devs/pytensor/releases/tag/rel-2.19.0

amyoshino:
> @amyoshino PyTensor new release should be in conda soon. We will need to update the pin on PyTensor version here, as it was a "major" release (or in a separate PR if there are other things that need fix)
> 
> https://github.com/pymc-devs/pytensor/releases/tag/rel-2.19.0

That's awesome!! Looking forward to it to be updated! 
Let me know if I need to do anything with respect to this PR as part of the process.

ricardoV94:
You need to bump the PyTensor dependency constraints on the PyMC side

amyoshino:
Ah, I see! Alright, let me try that then! Thanks for the pointer.

amyoshino:
Just pushed the bump dependency modifications in here. Seems like there is one test failing though, which wasn't failing when I ran the test suite locally.

ricardoV94:
@amyoshino I'm taking core of the Pytensor bump in https://github.com/pymc-devs/pymc/pull/7203

ricardoV94:
> Just pushed the bump dependency modifications in here. Seems like there is one test failing though, which wasn't failing when I ran the test suite locally.

May have been an unlucky seed, rerunning it

amyoshino:
> @amyoshino I'm taking core of the Pytensor bump in #7203

In much better hands now 😄 


> > Just pushed the bump dependency modifications in here. Seems like there is one test failing though, which wasn't failing when I ran the test suite locally.
> 
> May have been an unlucky seed, rerunning it

It was indeed a bad luck, the only missing part was the mypy, I guess because of the very outdated repo. I rebased and left only the previous commits here, so it is failing now because PyTensor is still 2.18. I thin the way then is to wait for your PR to be merged and try again, right?

ricardoV94:
> I think the way to go then is to wait for your PR to be merged and try again, right?

Right. I was hoping it was an easier bump, but it also unpins a Python major version :D

amyoshino:
Cool! It looks like the bump is merged! Looking forward to restart the checks when you think is appropriate.

amyoshino:
> > Cool! It looks like the bump is merged! Looking forward to restart the checks when you think is appropriate.
> 
> You can rebase anytime to incorporate the bump

Ah, true! I got so excited that the bump was merged and this potentially can be merged after so long that I forgot that 😅  My apologies haha

ricardoV94 (approved):
Amazing work @amyoshino !!

twiecki:
Congrats, this was a tricky one!

amyoshino:
That was fun! 😄 
Thank you for the support through this journey!

PR #6848, opened 2023-08-07
Do not use `seeded_test` fixture in exported `BaseTestDistributionRandom`

ricardoV94:
This caused pymc-experimental to fail as the test is used there but there is no access to the fixture. 
The fixture was added after the changes in #6799 
I don't think we need it?

CC @aerubanov 

----
:books: Documentation preview :books:: https://pymc--6848.org.readthedocs.build/en/6848/

### DIFF
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -848,7 +848,7 @@ class BaseTestDistributionRandom:
     repeated_params_shape = 5
     random_state = None
 
-    def test_distribution(self, seeded_test):
+    def test_distribution(self):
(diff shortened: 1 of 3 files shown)


aerubanov:
@ricardoV94 Yeah, looks like `seeded_test` fixture is actually is not needed here.

jessegrabowski (approved):
LGTM. The tests even seem like they were written to be like this in the first place (judging from the `partial` functions that pass in the random state)

PR #6850, opened 2023-08-08
Fix logcdf and icdf derivations for non-monotonically increasing transformations

ricardoV94:
https://github.com/pymc-devs/pymc/pull/6597 was only valid for strictly monotonically increasing transformations.

This PR fixes the derivations for monotonically decreasing transformations and adds a special case for logcdf of negative powers for distributions that allow negative values.

Also fixes the derivations for cosh transformed variables as that is not injective (CC @LukeLB)

Refactored a bit the test module with Test classes. It was growing a tad too unmanageable. 
The first two commits are just refactoring and bugfixes come after

----
:books: Documentation preview :books:: https://pymc--6850.org.readthedocs.build/en/6850/

### DIFF
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -440,6 +440,10 @@ def measurable_transform_logprob(op: MeasurableTransform, values, *inputs, **kwa
     return pt.switch(pt.isnan(jacobian), -np.inf, input_logprob + jacobian)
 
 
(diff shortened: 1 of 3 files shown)


larryshamalama (approved):
Apologies for the delay... Looks great @ricardoV94 :) No major questions from me this time. Admittedly, I didn't look _too_ carefully at the tests, I have full confidence in them 😅

PR #6853, opened 2023-08-11
Add JAX implementation fol `MatrixIsPositiveDefinite` `Op`

### LINKED ISSUE #6849: ENH: JAX OP `MatrixIsPositiveDefinite`
Hi! I tried to replicate the Correlated priors with `LKJCorr` from https://tomicapretto.github.io/posts/2022-06-12_lkj-prior/#model-4-correlated-priors-with-lkjcorr.-replicate-rstanarm-prior using the `numpyro` sampler and got the error:

### Before

```python
`NotImplementedError: No JAX conversion for the given `Op`: MatrixIsPositiveDefinite`. 
```

So it would be great to add this operator. 

Is this something a newbie like me could try?

### After

```python
We should be able to run the model above wit the JAX sampler from `numpyro`.
```

juanitorduz:
Closes: https://github.com/pymc-devs/pymc/issues/6849

Following https://pytensor.readthedocs.io/en/latest/extending/creating_a_numba_jax_op.html

----
:books: Documentation preview :books:: https://pymc--6853.org.readthedocs.build/en/6853/


tests/sampling/test_jax.py
```diff
@@ -49,6 +52,14 @@ def test_old_import_route():
     assert set(new_sj.__all__) <= set(dir(old_sj))
 
 
+def test_jax_PosDefMatrix():
+    x = pt.tensor()
+    matrix_pos_def = PosDefMatrix()
+    x_is_pos_def = matrix_pos_def(x)
+    out_fg = FunctionGraph(outputs=[x_is_pos_def])
+    compare_jax_and_py(fgraph=out_fg, test_inputs=pt.eye(3))
```
juanitorduz: @ricardoV94 I can not use `from tests.link.jax.test_basic import compare_jax_and_py` from `pymc` (because it belongs to the tests) ...shall we add a copy?

pymc/sampling/jax.py
```diff
@@ -74,6 +74,15 @@ def assert_fn(value, *inps):
     return assert_fn
 
 
+@jax_funcify.register(PosDefMatrix)
+def jax_funcify_PosDefMatrix(op, **kwargs):
+    def posdefmatrix_fn(value, *inps):
+        no_pos_def = jnp.any(jnp.isnan(jnp.linalg.cholesky(value)))
+        return (no_pos_def == jnp.array(False, dtype="bool")).astype(jnp.int_)
```
juanitorduz: Do we have a better way to negate a statement and convert it to `int`? This smells bad ...

tests/sampling/test_jax.py
ricardoV94: I don't think there's anything special about that. I would just compile a function to JAX and confirm it returns the expected results

juanitorduz: I want to get the test working with a simple case, and then I will add parametrize it with edge cases.

ricardoV94: We should compile only once and test with multiple values (pos and not pos). Parametrize is not ideal because we don't want to recompile/jit just to test a new value

pymc/sampling/jax.py
ricardoV94: The PyTensor Op doesn't return a bool? It seems like it should

juanitorduz: It returns an integer:

```python
import pytensor.tensor as pt
from pymc.distributions.multivariate import PosDefMatrix

matrix_pos_def = PosDefMatrix()

matrix_pos_def(pt.eye(2)).eval()
>> array(1, dtype=int8)
```

ricardoV94: Let's check if we can change it to bool. Should need tweaking of `make_node` and `perform` of the original Op

juanitorduz: Tried to do it in https://github.com/pymc-devs/pymc/pull/6853/commits/013ead3c8b00b7331429f7702a90092ea733e1a7

juanitorduz: (and tests https://github.com/pymc-devs/pymc/pull/6853/commits/75ca38bdc7f1bf4acee903cfec8b79de3bfa5f10)

ricardoV94: Then the jax one can/shoud return the boolean variable `not any(...)`

juanitorduz: In https://github.com/pymc-devs/pymc/pull/6853/commits/b9a2ef8007bfcce9c166858addf909b3c3bce878 I made the output boolean (as in the tests)

pymc/sampling/jax.py
```diff
@@ -74,6 +74,15 @@ def assert_fn(value, *inps):
     return assert_fn
 
 
+@jax_funcify.register(PosDefMatrix)
+def jax_funcify_PosDefMatrix(op, **kwargs):
+    def posdefmatrix_fn(value, *inps):
+        no_pos_def = jnp.any(jnp.isnan(jnp.linalg.cholesky(value)))
+        return no_pos_def == jnp.array(False, dtype="bool")
```
ricardoV94: Isn't this enough?

```suggestion
        return not no_pos_def
```

ricardoV94: I would make it a one liner because the double negation with the name is confusing

ricardoV94:
We have to check that HSGP failing test

juanitorduz:
> We have to check that HSGP failing test

It is passing for me locally 🤔 
```python
tests/gp/test_hsgp_approx.py .. 
```

tests/sampling/test_jax.py
```diff
@@ -49,6 +51,26 @@ def test_old_import_route():
...
+    argvalues=[
+        (jnp.eye(3), True),
+        (jnp.zeros(shape=(10, 10)), False),
+        (jnp.array([[1, -1.5], [0, 1.2]], dtype="float32"), True),
+        (-1 * jnp.array([[1, -1.5], [0, 1.2]], dtype="float32"), False),
+        (jnp.array([[1, -1.5], [0, -1.2]], dtype="float32"), False),
+        (jnp.array([[1, 0.99, 1], [0.99, 1, 0.999], [1, 0.999, 1]], dtype="float32"), False),
+    ],
+)
+def test_jax_PosDefMatrix(input, expected):
```
ricardoV94: Let's not parametrize, it's quite wasteful to compile a function for each different value we want to test.

juanitorduz: So should l loop over the cases inside the test to test all the cases? (after the compilation)

juanitorduz:
> We have to check that HSGP failing test

The tests pass locally 🤔

tests/sampling/test_jax.py
ricardoV94: Yup

ricardoV94 (approved):

juanitorduz:
Some random (unrelated) tests failed 🤷

ricardoV94:
I am not sure is random, doesn't it use the Op we modified?

juanitorduz:
Well, before the last commit where I simply changed the test parameterization all the tests passed. Can you maybe rerun the failed jobs? I can double check from my side.

ricardoV94:
I don't know if the test is deterministic, so passing once isn't a conclusive thing. I can have a look on Monday

ricardoV94:
There was a bug in the MatrixNormal test, it was supposed to try a different draw in the k2_samp test when it failed (10 times), but it was reusing the same draws everytime.

ricardoV94:
I fixed the test and cleaned the PR git history so we can merge without squash. Let me know if it looks correct (and if you need chages be carefull with fetching the remote branch)

juanitorduz:
Great, thank you! Looks great!

tests/gp/test_hsgp_approx.py
```diff
@@ -178,7 +178,7 @@ def test_prior(self, model, cov_func, X1, parameterization):
             gp = pm.gp.Latent(cov_func=cov_func)
             f2 = gp.prior("f2", X=X1)
 
-            idata = pm.sample_prior_predictive(samples=1000)
+            idata = pm.sample_prior_predictive(samples=1000, random_seed=rng)
```
juanitorduz, line 181: 🙏

ricardoV94: One day we will get this 5 line PR merged. Just you wait :D

PR #6865, opened 2023-08-17
Add `Model.to_graphviz` shortcut

### LINKED ISSUE #6794: Add `graphviz` method to `Model`
### Description

I think this is used so often that it's worthwhile offering as a method from the Model class. WDYT?

juanitorduz:
Closes https://github.com/pymc-devs/pymc/issues/6794

----
:books: Documentation preview :books:: https://pymc--6865.org.readthedocs.build/en/6865/

### DIFF
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1922,7 +1922,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
                 obs = Normal("obs", theta, sigma=sigma, observed=y)
 
-            model_to_graphviz(schools)
+            schools.graphviz()
         """
         return model_to_graphviz(model=self, var_names=var_names, formatting=formatting)
 
@@ -1969,7 +1969,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
                 obs = Normal("obs", theta, sigma=sigma, observed=y)
 
-            model_to_networkx(schools)
+            schools.networkx()
         """
         return model_to_networkx(model=self, var_names=var_names, formatting=formatting)
 
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -1924,7 +1924,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
                 obs = Normal("obs", theta, sigma=sigma, observed=y)
 
-            schools.graphviz()
+            schools.to_graphviz()
         """
         return model_to_graphviz(model=self, var_names=var_names, formatting=formatting)
 
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -66,7 +66,7 @@ from pymc.exceptions import (
 from pymc.initial_point import make_initial_point_fn
 from pymc.logprob.basic import transformed_conditional_logp
 from pymc.logprob.utils import ParameterValueError
-from pymc.model_graph import VarName, model_to_graphviz, model_to_networkx
+from pymc.model_graph import VarName, model_to_graphviz
 from pymc.pytensorf import (
     PointFunc,
     SeedSequenceSeed,
@@ -1925,53 +1925,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
             schools.graphviz()
         """
         return model_to_graphviz(model=self, var_names=var_names, formatting=formatting)
-
-    def networkx(self, *, var_names: Optional[Iterable[VarName]] = None, formatting: str = "plain"):
-        """Produce a networkx Digraph from a PyMC model.
-
-        Requires networkx, which may be installed most easily with::
-
-            conda install networkx
-
-        Alternatively, you may install using pip with::
-
-            pip install networkx
-
-        See https://networkx.org/documentation/stable/ for more information.
-
-        Parameters
-        ----------
-        var_names : iterable of str, optional
-            Subset of variables to be plotted that identify a subgraph with respect to the entire model graph
-        formatting : str, optional
-            one of { "plain" }
-
-        Examples
-        --------
-        How to plot the graph of the model.
-
-        .. code-block:: python
-
-            import numpy as np
-            from pymc import HalfCauchy, Model, Normal, model_to_networkx
-
-            J = 8
-            y = np.array([28, 8, -3, 7, -1, 1, 18, 12])
-            sigma = np.array([15, 10, 16, 11, 9, 11, 10, 18])
-
-            with Model() as schools:
-
-                eta = Normal("eta", 0, 1, shape=J)
-                mu = Normal("mu", 0, sigma=1e6)
-                tau = HalfCauchy("tau", 25)
-
-                theta = mu + tau * eta
-
-                obs = Normal("obs", theta, sigma=sigma, observed=y)
-
-            schools.networkx()
-        """
-        return model_to_networkx(model=self, var_names=var_names, formatting=formatting)
 
 
 # this is really disgusting, but it breaks a self-loop: I can't pass Model
--- a/pymc/model.py
+++ b/pymc/model.py
@@ -24,6 +24,7 @@ from typing import (
     Any,
     Callable,
     Dict,
+    Iterable,
     List,
     Literal,
     Optional,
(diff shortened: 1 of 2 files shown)


pymc/model.py
```diff
@@ -1879,6 +1881,98 @@ def debug_parameters(rv):
         elif not verbose:
             print_("You can set `verbose=True` for more details")
 
+    def graphviz(self, *, var_names: Optional[Iterable[VarName]] = None, formatting: str = "plain"):
+        """Produce a graphviz Digraph from a PyMC model.
```
juanitorduz, line 1887: I took the docstrings from original methods. Is this enough?

tests/test_model.py
```diff
@@ -1653,3 +1654,26 @@ def test_model_logp_fast_compile():
...
+            obs = pm.Normal("obs", theta, sigma=sigma, observed=y)
+        return schools
+
+    def test_graphviz(self) -> None:
+        model = self.school_model()
+        assert model.graphviz()
+
+    def test_networkx(self) -> None:
+        model = self.school_model()
+        model_to_networkx(model=model) == model.networkx()
```
juanitorduz: What should I test so that it is not redundant with respect to the `model_graph` tests?

pymc/model.py
```diff
@@ -1879,6 +1881,98 @@ def debug_parameters(rv):
...
+
+                eta = Normal("eta", 0, 1, shape=J)
+                mu = Normal("mu", 0, sigma=1e6)
+                tau = HalfCauchy("tau", 25)
+
+                theta = mu + tau * eta
+
+                obs = Normal("obs", theta, sigma=sigma, observed=y)
+
+            model_to_graphviz(schools)
```
ricardoV94: Probably better to show the use as a method?

pymc/model.py
```diff
@@ -1879,6 +1881,98 @@ def debug_parameters(rv):
...
+        """
+        return model_to_graphviz(model=self, var_names=var_names, formatting=formatting)
+
+    def networkx(self, *, var_names: Optional[Iterable[VarName]] = None, formatting: str = "plain"):
```
ricardoV94: Maybe this one isn't used enough to justify the shortcut method?

juanitorduz: Just for the sake of completeness? (I do not have a strong opinion).

pymc/model.py
juanitorduz: Indeed! Fixed in https://github.com/pymc-devs/pymc/pull/6865/commits/576d916eb0c28cce768ca376e8ddae5c8100489d

pymc/model.py
ricardoV94: Too many methods hurt discoverability

juanitorduz: True. ok ! I will remove it

juanitorduz: Removed in https://github.com/pymc-devs/pymc/pull/6865/commits/0457a3480573913b78969cc7f44f503e5ee4bcba

tests/test_model.py
```diff
@@ -1653,3 +1654,22 @@ def test_model_logp_fast_compile():
...
+            eta = pm.Normal("eta", 0, 1, shape=J)
+            mu = pm.Normal("mu", 0, sigma=1e6)
+            tau = pm.HalfCauchy("tau", 25)
+            theta = mu + tau * eta
+            obs = pm.Normal("obs", theta, sigma=sigma, observed=y)
+        return schools
+
+    def test_graphviz(self) -> None:
+        model = self.school_model()
+        assert model.graphviz()
```
ricardoV94: Can the output be compared by equality to what is returned when calling `model_graphviz(model)`?

juanitorduz: I tried the simple `assert pm.model_graph_model_to_graphviz(model) == model.graphviz()` but this failed because probably there is no default how to compare the output graphs. I will investigate.

juanitorduz: Addressed in https://github.com/pymc-devs/pymc/pull/6865/commits/3c252d02520949546b5ef4793a258cc9a2fb54ff

tests/test_model.py
```diff
@@ -1653,3 +1654,32 @@ def test_model_logp_fast_compile():
...
+            eta = pm.Normal("eta", 0, 1, dims="school")
+            mu = pm.Normal("mu", 0, sigma=1e6)
+            tau = pm.HalfCauchy("tau", 25)
+            theta = mu + tau * eta
+            pm.Normal("obs", theta, sigma=sigma, observed=y, dims="school")
+        return schools
+
+    def test_graphviz(self) -> None:
+        model = self.school_model(J=8)
+        g_actual = model.graphviz()
```
michaelosthege: Technically, one could use patch/mock and only assert that everything was forwarded to the underlying `model_to_graphviz` function

michaelosthege (commented):
I would prefer the `to_graphviz` prefix - not only to align with the original functions, but also to make it more discoverable with other (future) conversion methods. Thinking of `to_networkx`..

tests/test_model.py
juanitorduz: I added such mock in https://github.com/pymc-devs/pymc/pull/6865/commits/ec02a583a97c7b0fff1c74a6452e5d9ffc682df8 . Open for feedback regarding the tests we cactually need and which ones we can delete

tests/test_model.py
```diff
@@ -1653,3 +1656,43 @@ def test_model_logp_fast_compile():
...
+        assert g_actual.body == g_expected.body
+        assert g_actual.source == g_expected.source
+
+    def test_graphviz_different_models(self) -> None:
+        model_1 = self.school_model(J=8)
+        model_2 = self.school_model(J=5)
+        g_1 = model_1.to_graphviz()
+        g_2 = model_2.to_graphviz()
+        assert g_1.body != g_2.body
+        assert g_1.source != g_2.source
```
michaelosthege:
```suggestion
            )
```

IMO the mocked test is sufficient. For the other tests to catch something that's not already caught by existing tests one would have to actively sabotage the new method

michaelosthege (approved):

tests/test_model.py
juanitorduz: ok! I'll do so thanks!

pymc/model.py
```diff
@@ -1879,6 +1881,53 @@ def debug_parameters(rv):
         elif not verbose:
             print_("You can set `verbose=True` for more details")
 
+    def to_graphviz(
```
twiecki, line 1884: should we call this graphviz? who cares what tool is being used under the hood?

twiecki: Maybe `plot_graph()`?

juanitorduz: We do have the function `model_to_networkx` so `graphviz` is not the only option ¯\\_(ツ)_/¯

twiecki:
We still need the string so that this shows up in the release notes.

pymc/model.py
```diff
@@ -1879,6 +1881,53 @@ def debug_parameters(rv):
...
+
+                eta = Normal("eta", 0, 1, shape=J)
+                mu = Normal("mu", 0, sigma=1e6)
+                tau = HalfCauchy("tau", 25)
+
+                theta = mu + tau * eta
+
+                obs = Normal("obs", theta, sigma=sigma, observed=y)
+
+            schools.graphviz()
```
```suggestion
            schools.to_graphviz()
```

ricardoV94 (changes requested):

ricardoV94:
> We still need the string so that this shows up in the release notes.

What do you mean? The PR title is all that's needed. Then the label determines in which section it shows up.

Armavica:
I was using this today, I forgot how it worked and all the explanations assume that people use notebooks: would it make sense to add a return type (`str`) to the function, and explain somewhere that outside of a notebook, you would need to write this string to a file and call `dot` on it?

ricardoV94 (approved):

ricardoV94:
> I was using this today, I forgot how it worked and all the explanations assume that people use notebooks: would it make sense to add a return type (`str`) to the function, and explain somewhere that outside of a notebook, you would need to write this string to a file and call `dot` on it?

Do you mind opening an issue for that? I don't think it should block this issue

PR #6866, opened 2023-08-19
Allow Multinomial to accept unnormalized dists

Armavica:
I moved a sum from the `check_parameters` arguments to the logp, in order to allow `Multinomial` to accept unnormalized distributions.
This simplifies the function graph, because from two sums over `p` (the user's, and the sum in the `check_parameters` arguments), now we can have only a single one (in the `logp` term). And if the user's symbolic distribution is somehow already normalized, then it is the same number of sums.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- The `Multinomial` distribution now allows probability arrays that don't sum to 1.

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--6866.org.readthedocs.build/en/6866/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -563,10 +563,7 @@ class Multinomial(Discrete):
                     "You can rescale them directly to get rid of this warning.",
                     UserWarning,
                 )
-                p_ = p_ / pt.sum(p_, axis=-1, keepdims=True)
-                p = pt.as_tensor_variable(p_)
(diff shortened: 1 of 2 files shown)


ricardoV94:
It was an explicit decision not to normalize probabilities for the user in Multinomial/Categorical: https://github.com/pymc-devs/pymc/discussions/5246

Armavica:
Ok, never mind then. It looks like the extra loop can be optimized away anyway.

PR #6872, opened 2023-08-21
Fix wrong ZeroSumNormal logp expression

### LINKED ISSUE #6871: BUG: `ZeroSumNormal` has an incorrect `logp`
### Describe the issue:

The `ZeroSumNormal` distribution has the wrong `logp` expression (more details after the code example below). A way to see this is to compare draws from the prior using the `rng` method and draws via HMC. The draws from the `rng` method have the correct variance but the draws from HMC don't.

```python
import numpy as np
import pymc as pm


with pm.Model():
    rv = pm.ZeroSumNormal("zsn", sigma=2, shape=2)
    idata = pm.sample()
    idata.extend(pm.sample_prior_predictive(1000))
print(idata.prior.zsn.var(["chain", "draw"]).data)  # prints something close to [2, 2]
print(idata.posterior.zsn.var(["chain", "draw"]).data)  # prints something close to [4, 4]
```

### Error message:

_No response_

### PyMC version information:

From pip

### Context for the issue:

This will be a rather long explanation of what is actually wrong with the `logp` implementation.

A `ZeroSumNormal` distributed random variable is a special kind of `MvNormal` that comes from a reparametrization of a vector of I.I.D normals.

$$
\epsilon_{i} \sim Normal(0, \sigma) \quad 0\leq i \leq N
$$

$$
\tilde{\epsilon_{i}} = \epsilon_{i} - \frac{\sum_{i}\epsilon_{i}}{N}
$$

$\vec{\tilde{\epsilon}}$ follows a `ZeroSumNormal` with `sigma` given by the scale of $\vec{\epsilon}$. The `MvNormal` distribution that corresponds to $\vec{\tilde{\epsilon}}$ is one that has a zero mean vector and a covariance matrix that is the result of the following operation:

$$
\tilde{\Sigma} = C\Sigma C
$$

where $C$ is called the [centering matrix](https://en.wikipedia.org/wiki/Centering_matrix) or operator, and $\Sigma$ is the diagonal matrix with every entry equal to $\sigma$. The resulting covariance matrix of the zero sum normal distribution is degenerate, it has one eigenvalue that is exactly equal to zero (the corresponding eigenvector has 1's at every entry, i.e. `np.ones`).

Following the notation from the wikipedia article for the degenerate multivariate normal, the probability density function of $\vec{\tilde{\epsilon}}$ is

$$
f(\vec{\tilde{\epsilon}}) = \frac{1}{\sqrt{Det^{*}\left(2\pi\tilde{\Sigma}\right)}}\exp\left(\vec{\tilde{\epsilon}}^{T} . \tilde{\Sigma}^{+} . \vec{\tilde{\epsilon}}\right)
$$

where $Det^{*}$ mean pseudo-determinant and $\tilde{\Sigma}^{+}$ is the pseudo-inverse of $\tilde{\Sigma}$ (as a side-note, this is the density if the value $\vec{\tilde{\epsilon}}$ is one the zero-sum manifold, if it isn't, it has 0 probability). Looking at the exponent, we can work out a simpler expression for the density using the definition of $\tilde{\Sigma}$ and the fact that $C$ is idempotent and an identity operator for vectors on the zero-sum manifold:

$$
\tilde{\Sigma}^{+} = C^{+}\Sigma^{-1} C^{+}
$$

$$
C\vec{\tilde{\epsilon}} = \vec{\tilde{\epsilon}}
$$

So the density can be re-written as

$$
f(\vec{\tilde{\epsilon}}) = \frac{1}{\sqrt{Det^{*}\left(2\pi\tilde{\Sigma}\right)}}\exp\left(\vec{\tilde{\epsilon}}^{T} . \Sigma^{-1} . \vec{\tilde{\epsilon}}\right)
$$

The final simplification is that the pseudo-determinant is equal to the product of the non-zero eigenvalues of the matrix that is being considered. Since we are working with the case where $\Sigma$ is proportional to an identity matrix, $\tilde{\Sigma}$ has all of its eigenvalues equal to $\sigma^{2}$ except for a single eigenvalue equal to zero, so the pseudo-determinant is equal to $(2\pi\sigma^{2})^{N-1}$, which produces a very simple probability density:

$$
f(\vec{\tilde{\epsilon}}) = \left(2\pi\sigma^{2}\right)^{-\frac{N-1}{2}}\exp\left(\vec{\tilde{\epsilon}}^{T} . \Sigma^{-1} . \vec{\tilde{\epsilon}}\right)
$$

The resulting logp is:

$$
\log(f(\vec{\tilde{\epsilon}})) = -\frac{N-1}{2}\log\left(2\pi\sigma^{2}\right) - \left(\vec{\tilde{\epsilon}}^{T} . \Sigma^{-1} . \vec{\tilde{\epsilon}}\right)
$$

which is very similar to the `Normal` logp, except for the fact that the normalization constant changes according to the "degree of freedom" of the zero sum normal, instead of the raw length of the vector.

This issue is caused by the fact that [`ZeroSumNormal.logp`](https://github.com/pymc-devs/pymc/blob/main/pymc/distributions/multivariate.py#L2795) is rescaling both the exponent and the normalization constant of the `logp`. To fix this issue, the logp should be rewritten as something like

```python
pt.sum(
    -0.5 * pt.pow(value / sigma, 2)
    - (
        pt.log(pt.sqrt(2.0 * np.pi)) + pt.log(sigma)
    ) * _degrees_of_freedom / _full_size,
    axis=-1,
)
```

lucianopaz:
Closes #6871

**What is this PR about?**

This PR fixes the `ZeroSumNormal` `logp` expression. It also changes the `logp` test to to compare the `logp` to the reference value for points drawn at random in the zero sum manifold and not only at the mode.

**Checklist**

## Bugfixes
- Fix the `logp` expression for the `ZeroSumNormal` distribution

----
:books: Documentation preview :books:: https://pymc--6872.org.readthedocs.build/en/6872/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2781,6 +2781,7 @@ def zerosumnormal_logp(op, values, normal_dist, sigma, support_shape, **kwargs):
     (value,) = values
     shape = value.shape
     n_zerosum_axes = op.ndim_supp
+    *_, sigma = normal_dist.owner.inputs
 
     _deg_free_support_shape = pt.inc_subtensor(shape[-n_zerosum_axes:], -1)
     _full_size = pt.prod(shape)
@@ -2792,7 +2793,8 @@ def zerosumnormal_logp(op, values, normal_dist, sigma, support_shape, **kwargs):
     ]
 
     out = pt.sum(
-        pm.logp(normal_dist, value) * _degrees_of_freedom / _full_size,
+        -0.5 * pt.pow(value / sigma, 2)
+        - (pt.log(pt.sqrt(2.0 * np.pi)) + pt.log(sigma)) * _degrees_of_freedom / _full_size,
         axis=tuple(np.arange(-n_zerosum_axes, 0)),
     )
 
--- a/tests/distributions/test_multivariate.py
+++ b/tests/distributions/test_multivariate.py
@@ -1596,7 +1596,7 @@ class TestZeroSumNormal:
             (5, 3, None, [-1]),
             (2, 6, None, [-1]),
             (5, (7, 3), None, [-1]),
-            (5, (2, 7, 3), 2, [1, 2]),
+            (5, (2, 7, 3), 2, [-2, -1]),
         ],
     )
     def test_zsn_logp(self, sigma, shape, n_zerosum_axes, mvn_axes):
@@ -1629,8 +1629,9 @@ class TestZeroSumNormal:
             return np.where(inds, np.sum(-psdet - exp, axis=-1), -np.inf)
 
         zsn_dist = pm.ZeroSumNormal.dist(sigma=sigma, shape=shape, n_zerosum_axes=n_zerosum_axes)
-        zsn_logp = pm.logp(zsn_dist, value=np.zeros(shape)).eval()
-        mvn_logp = logp_norm(value=np.zeros(shape), sigma=sigma, axes=mvn_axes)
+        zsn_draws = pm.draw(zsn_dist, 100)
+        zsn_logp = pm.logp(zsn_dist, value=zsn_draws).eval()
+        mvn_logp = logp_norm(value=zsn_draws, sigma=sigma, axes=mvn_axes)
 
         np.testing.assert_allclose(zsn_logp, mvn_logp)
 


ricardoV94 (approved):

lucianopaz:
I'll go ahead and merge

PR #6873, opened 2023-08-22
Default `moment` for `CustomDist` provided with a `dist` function

### LINKED ISSUE #6804: Implement default `moment` function for `CustomDist` provided with a `dist` function
### Description

```python
import pymc as pm

def dist(mu, sigma, size):
  return pm.math.exp(pm.Normal.dist(mu, sigma size=size))`

with pm.Model() as m:
  ...
  x = pm.CustomDist("x", mu, sigma, dist=dist)

m.initial_point()
```

Since the `dist` is a symbolic forward graph, we could implement the `moment` automatically as
```python
def moment_dist(mu, sigma, size):
  return pm.math.exp(moment(pm.Normal.dist(mu, sigma, size))
```

Would just need to replace the RVs in the `dist` graph, by their moments.

aerubanov:
Add default implementation of `moment` for `CustomDist` with a `dist` function and close #6804

----
:books: Documentation preview :books:: https://pymc--6873.org.readthedocs.build/en/6873/

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -95,7 +95,7 @@ class MomentRewrite(GraphRewriter):
                 if (
                     hasattr(inp, "owner")
                     and hasattr(inp.owner, "op")
-                    and isinstance(inp.owner.op, Distribution)
+                    and isinstance(inp.owner.op, RandomVariable)
                 ):
                     fgraph.replace(node.inputs[i], moment(node.inputs[i]))
 
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -34,7 +34,7 @@ from pytensor.scan.op import Scan
 from pytensor.tensor.basic import as_tensor_variable
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.rewriting import local_subtensor_rv_lift
-from pytensor.tensor.random.type import RandomType
+from pytensor.tensor.random.type import RandomGeneratorType, RandomType
 from pytensor.tensor.random.utils import normalize_size_param
 from pytensor.tensor.rewriting.shape import ShapeFeature
 from pytensor.tensor.variable import TensorVariable
@@ -86,7 +86,7 @@ PLATFORM = sys.platform
 
 
 def filter_RNGs(params):
-    return [p for p in params if not isinstance(p.type, RandomType)]
+    return [p for p in params if not isinstance(p.type, (RandomType, RandomGeneratorType))]
 
 
 class MomentRewrite(GraphRewriter):
@@ -661,13 +661,6 @@ class CustomSymbolicDistRV(SymbolicRandomVariable):
         return updates
 
 
-def get_rv_fgraph(dist_fn, dist_params, size):
-    rv = dist_fn(*dist_params, size=size)
-    outputs = [rv]
-    fgraph = FunctionGraph(outputs=outputs, clone=True)
-    return fgraph
-
-
 class _CustomSymbolicDist(Distribution):
     rv_type = CustomSymbolicDistRV
 
@@ -690,9 +683,32 @@ class _CustomSymbolicDist(Distribution):
             logcdf = default_not_implemented(class_name, "logcdf")
 
         def dist_moment(rv, size, *dist_params):
-            fgraph = get_rv_fgraph(dist, dist_params, size)
+            size = normalize_size_param(size)
+            dummy_size_param = size.type()
+            dummy_dist_params = [dist_param.type() for dist_param in dist_params]
+            dummy_rv = dist(*dummy_dist_params, dummy_size_param)
+            # dummy_updates_dict = collect_default_updates(inputs=dummy_dist_params, outputs=(dummy_rv,))
+            dummy_params = [dummy_size_param] + dummy_dist_params
+            # dummy_updates_dict = collect_default_updates(inputs=dummy_params, outputs=(dummy_rv,))
+            rv_type = type(
+                class_name,
+                (CustomSymbolicDistRV,),
+                # If logp is not provided, we try to infer it from the dist graph
+                dict(
+                    inline_logprob=logp is None,
+                ),
+            )
+            rv_op = rv_type(
+                inputs=dummy_params,
+                # inputs=dummy_dist_params,
+                outputs=[dummy_rv],
+                ndim_supp=ndim_supp,
+            )
+            fgraph = rv_op.fgraph.clone()
             replace_moments = MomentRewrite()
             replace_moments.rewrite(fgraph)
+            for i, par in enumerate([size] + list(dist_params)):
+                fgraph.replace(fgraph.inputs[i], par)
             [moment] = fgraph.outputs
             return moment
 
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -127,12 +127,12 @@ class MomentRewrite(GraphRewriter):
 
     def apply(self, fgraph):
         for node in fgraph.toposort():
-            if isinstance(node.op, Scan):
+            if isinstance(node.op, (RandomVariable, SymbolicRandomVariable)):
+                fgraph.replace(node.out, moment(node.out))
+            elif isinstance(node.op, Scan):
                 new_node = rewrite_moment_scan_node(node)
                 for out1, out2 in zip(node.outputs, new_node.outputs):
                     fgraph.replace(out1, out2)
-            elif isinstance(node.op, (RandomVariable, SymbolicRandomVariable)):
-                fgraph.replace(node.out, moment(node.out))
 
 
 class _Unpickling:
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -98,17 +98,6 @@ class MomentRewrite(GraphRewriter):
                 ):
                     fgraph.replace(node.inputs[i], moment(node.inputs[i]))
 
-    # def transform(self, fgraph, node):
-    #     flag = False
-    #     for i, inp in enumerate(node.inputs):
-    #         if hasattr(inp, "owner") and hasattr(inp.owner, "op") and isinstance(inp.owner.op, Distribution):
-    #             #node.inputs[i] = moment(inp)
-    #             node.inputs[i] = inp
-    #             flag = True
-    #     if flag:
-    #         return [node]
-    #     return False
-
 
 class _Unpickling:
     pass
@@ -657,12 +646,6 @@ class _CustomSymbolicDist(Distribution):
             return moment
 
         if moment is None:
-            # moment = functools.partial(
-            #    default_moment,
-            #    rv_name=class_name,
-            #    has_fallback=True,
-            #    ndim_supp=ndim_supp,
-            # )
             moment = dist_moment
 
         return super().dist(
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -85,10 +85,6 @@ vectorized_ppc: contextvars.ContextVar[Optional[Callable]] = contextvars.Context
 PLATFORM = sys.platform
 
 
-def filter_RNGs(params):
-    return [p for p in params if not isinstance(p.type, (RandomType, RandomGeneratorType))]
-
-
 class MomentRewrite(GraphRewriter):
     def rewrite_moment_scan_node(self, node):
         if not isinstance(node.op, Scan):
@@ -695,8 +691,8 @@ class _CustomSymbolicDist(Distribution):
         if logcdf is None:
             logcdf = default_not_implemented(class_name, "logcdf")
 
-        # if moment is None:
-        moment = dist_moment
+        if moment is None:
+            moment = dist_moment
 
         return super().dist(
             dist_params,
@@ -757,7 +753,15 @@ class _CustomSymbolicDist(Distribution):
 
             @_moment.register(rv_type)
             def custom_dist_get_moment(op, rv, size, *params):
-                return moment(rv, size, *filter_RNGs(params))
+                return moment(
+                    rv,
+                    size,
+                    *[
+                        p
+                        for p in params
+                        if not isinstance(p.type, (RandomType, RandomGeneratorType))
+                    ],
+                )
 
         @_change_dist_size.register(rv_type)
         def change_custom_symbolic_dist_size(op, rv, new_size, expand):
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -85,22 +85,6 @@ vectorized_ppc: contextvars.ContextVar[Optional[Callable]] = contextvars.Context
 PLATFORM = sys.platform
 
 
-class InnerMomentRewrite(GraphRewriter):
-    def add_requirements(self, fgraph):
-        fgraph.attach_feature(ReplaceValidate())
-
-    def apply(self, fgraph):
-        # breakpoint()
-        for node in fgraph.toposort():
-            for i, inp in enumerate(node.inputs):
-                if (
-                    hasattr(inp, "owner")
-                    and hasattr(inp.owner, "op")
-                    and isinstance(inp.owner.op, (RandomVariable, SymbolicRandomVariable))
-                ):
-                    fgraph.replace(node.inputs[i], moment(node.inputs[i]))
-
-
 class MomentRewrite(GraphRewriter):
     def add_requirements(self, fgraph):
         fgraph.attach_feature(ReplaceValidate())
@@ -108,8 +92,7 @@ class MomentRewrite(GraphRewriter):
     def apply(self, fgraph):
         for node in fgraph.toposort():
             if hasattr(node.op, "fgraph"):
-                moment_replace = InnerMomentRewrite()
-                moment_replace.rewrite(node.op.fgraph)
+                self.rewrite(node.op.fgraph)
             elif isinstance(node.op, (RandomVariable, SymbolicRandomVariable)):
                 fgraph.replace(node.out, moment(node.out))
 
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -95,7 +95,7 @@ class MomentRewrite(GraphRewriter):
                 if (
                     hasattr(inp, "owner")
                     and hasattr(inp.owner, "op")
-                    and isinstance(inp.owner.op, RandomVariable)
+                    and isinstance(inp.owner.op, (RandomVariable, SymbolicRandomVariable))
                 ):
                     fgraph.replace(node.inputs[i], moment(node.inputs[i]))
 
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -112,6 +112,8 @@ class MomentRewrite(GraphRewriter):
         if len(replace_with_moment) > 0:
             for item in replace_with_moment:
                 givens[item] = moment(item)
+        else:
+            return
         op_outs = clone_replace(node_outputs, replace=givens)
 
         nwScan = Scan(
@@ -135,9 +137,10 @@ class MomentRewrite(GraphRewriter):
             if isinstance(node.op, (RandomVariable, SymbolicRandomVariable)):
                 fgraph.replace(node.out, moment(node.out))
             elif isinstance(node.op, Scan):
-                new_node = rewrite_moment_scan_node(node)
-                for out1, out2 in zip(node.outputs, new_node.outputs):
-                    fgraph.replace(out1, out2)
+                new_node = self.rewrite_moment_scan_node(node)
+                if new_node is not None:
+                    for out1, out2 in zip(node.outputs, new_node.outputs):
+                        fgraph.replace(out1, out2)
 
 
 class _Unpickling:
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -23,7 +23,6 @@ from typing import Callable, Optional, Sequence, Tuple, Union
 
 import numpy as np
 
-from pytensor import shared
 from pytensor import tensor as pt
 from pytensor.compile.builders import OpFromGraph
 from pytensor.graph import FunctionGraph, clone_replace, node_rewriter
@@ -662,6 +661,20 @@ class CustomSymbolicDistRV(SymbolicRandomVariable):
         return updates
 
 
+def dist_moment(rv, *args):
+    node = rv.owner
+    op = node.op
+    rv_out_idx = node.outputs.index(rv)
+
+    fgraph = op.fgraph.clone()
+    replace_moments = MomentRewrite()
+    replace_moments.rewrite(fgraph)
+    # Replace dummy inner inputs by outer inputs
+    fgraph.replace_all(tuple(zip(op.inner_inputs, node.inputs)), import_missing=True)
+    moment = fgraph.outputs[rv_out_idx]
+    return moment
+
+
 class _CustomSymbolicDist(Distribution):
     rv_type = CustomSymbolicDistRV
 
@@ -682,26 +695,6 @@ class _CustomSymbolicDist(Distribution):
 
         if logcdf is None:
             logcdf = default_not_implemented(class_name, "logcdf")
-
-        def dist_moment(rv, size, *dist_params):
-            fgraph = rv.owner.op.fgraph.clone()
-            replace_moments = MomentRewrite()
-            replace_moments.rewrite(fgraph)
-            # we need to replace dummy variable by actual dist params
-            for i, par in enumerate([size] + list(dist_params)):
-                fgraph.replace(fgraph.inputs[i], par)
-            # we need to replace dymmy random generators in Scan node inputs
-            for node in fgraph.toposort():
-                if isinstance(node.op, Scan):
-                    for inp in node.inputs:
-                        if isinstance(inp.type, RandomGeneratorType):
-                            fgraph.replace(
-                                inp,
-                                shared(np.random.Generator(np.random.PCG64())),
-                                import_missing=True,
-                            )
-            moment = fgraph.outputs[-1]
-            return moment
 
         if moment is None:
             moment = dist_moment
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -85,11 +85,12 @@ vectorized_ppc: contextvars.ContextVar[Optional[Callable]] = contextvars.Context
 PLATFORM = sys.platform
 
 
-class MomentRewrite(GraphRewriter):
+class InnerMomentRewrite(GraphRewriter):
     def add_requirements(self, fgraph):
         fgraph.attach_feature(ReplaceValidate())
 
     def apply(self, fgraph):
+        # breakpoint()
         for node in fgraph.toposort():
             for i, inp in enumerate(node.inputs):
                 if (
@@ -98,6 +99,19 @@ class MomentRewrite(GraphRewriter):
                     and isinstance(inp.owner.op, (RandomVariable, SymbolicRandomVariable))
                 ):
                     fgraph.replace(node.inputs[i], moment(node.inputs[i]))
+
+
+class MomentRewrite(GraphRewriter):
+    def add_requirements(self, fgraph):
+        fgraph.attach_feature(ReplaceValidate())
+
+    def apply(self, fgraph):
+        for node in fgraph.toposort():
+            if hasattr(node.op, "fgraph"):
+                moment_replace = InnerMomentRewrite()
+                moment_replace.rewrite(node.op.fgraph)
+            elif isinstance(node.op, (RandomVariable, SymbolicRandomVariable)):
+                fgraph.replace(node.out, moment(node.out))
 
 
 class _Unpickling:
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -25,12 +25,12 @@ import numpy as np
 
 from pytensor import tensor as pt
 from pytensor.compile.builders import OpFromGraph
-from pytensor.graph import FunctionGraph, node_rewriter
-from pytensor.graph.basic import Node, Variable
+from pytensor.graph import FunctionGraph, clone_replace, node_rewriter
+from pytensor.graph.basic import Node, Variable, io_toposort
 from pytensor.graph.features import ReplaceValidate
-from pytensor.graph.replace import clone_replace
 from pytensor.graph.rewriting.basic import GraphRewriter, in2out
 from pytensor.graph.utils import MetaType
+from pytensor.scan.op import Scan
 from pytensor.tensor.basic import as_tensor_variable
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.rewriting import local_subtensor_rv_lift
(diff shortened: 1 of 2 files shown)


aerubanov:
@ricardoV94 I added implementation which use graph rewriting to replace distributions by corresponding moments. Could you please take a look?

pymc/distributions/distribution.py
```diff
@@ -83,6 +85,21 @@
...
+    def add_requirements(self, fgraph):
+        fgraph.attach_feature(ReplaceValidate())
+
+    def apply(self, fgraph):
+        for node in fgraph.toposort():
+            for i, inp in enumerate(node.inputs):
+                if (
+                    hasattr(inp, "owner")
+                    and hasattr(inp.owner, "op")
+                    and isinstance(inp.owner.op, Distribution)
```
ricardoV94: Maybe we should check if it's a RandomVariable instead of Distribution.

We also need to check if an Op had an InnerGraph (like Scan) and do the replacements inside

ricardoV94 (commented):
Neat approach! 

Just need to think about Ops with InnerGraphs

pymc/distributions/distribution.py
aerubanov: Thank you @ricardoV94, will do it.

pymc/distributions/distribution.py
```diff
@@ -83,6 +85,21 @@
...
+    def add_requirements(self, fgraph):
+        fgraph.attach_feature(ReplaceValidate())
+
+    def apply(self, fgraph):
+        for node in fgraph.toposort():
+            for i, inp in enumerate(node.inputs):
+                if (
+                    hasattr(inp, "owner")
+                    and hasattr(inp.owner, "op")
+                    and isinstance(inp.owner.op, RandomVariable)
```
ricardoV94: Ah could also be a symbolic one!

```suggestion
                    and isinstance(inp.owner.op, (RandomVariable, SymbolicRandomVariable))
```

aerubanov: Thank you!

aerubanov:
@ricardoV94 Do you know som simple example of symbolic `dist` with `Scan` op? I want to add test case with `Scan`, but I do not used it before, so when I try this:
```python
def test_custom_dist_custom_moment_inner_graph(self):
        def dist(mu, size):
            ys, _ = pytensor.scan(
                fn=lambda x: pt.exp(pm.Normal.dist(x, 1, size=size)),
                sequences=[mu],
                outputs_info=[None], 
                name="ys",
                )
            return pt.sum(ys)
        
        with Model() as model:
            CustomDist("x", pt.ones(2), dist=dist)
        assert_moment_is_expected(model, 2)
```
I get  `ValueError: No update found for at least one RNG used in Scan Op Scan{ys, while_loop=False, inplace=none}.` I guess I just use `scan` wrong

ricardoV94:
You need to return random updates from the scan. Check the utility `collect_default_updates` and the code example in the docstrings:

https://github.com/pymc-devs/pymc/blob/main/pymc/pytensorf.py#L1000

aerubanov:
@ricardoV94 yeah, it work, thank you!

pymc/distributions/distribution.py
```diff
@@ -83,6 +85,18 @@
...
 
 
+class MomentRewrite(GraphRewriter):
+    def add_requirements(self, fgraph):
+        fgraph.attach_feature(ReplaceValidate())
+
+    def apply(self, fgraph):
+        for node in fgraph.toposort():
+            if hasattr(node.op, "fgraph"):
+                self.rewrite(node.op.fgraph)
```
ricardoV94: We have to double check this isn't mutating the fgraph of the original Ops (I am not sure they are cloned by default in FunctionGraph).

More generally I am not sure this is the best approach. The fgraph is supposed to be just a property of the Op, the real underlying variables are Op.outputs I think? So we could create a FunctionGraph from those, rewrite them and then create a new node with a new op that uses those new outputs and replace it in the outer graph...

However these Ops don't have a shared construction API IIRC, so we may need special logic for each one. They are Scan, OpFromGraph, Composite, and ScalarLoop. 

I think we can focus on Scan alone since the others aren't user facing or don't make sense inside a CustomDist (just raise NotImplementedError in those cases)

aerubanov: @ricardoV94 thank you for review! I will try to replace `Scan` op in outer graph by new node with moment replaced inner graph.

aerubanov: @ricardoV94 If I understand correctly, new `Scan` op in outer graph is already created in my implementation when I clone graph  from dist function. So we need only replace inner graph, or do I miss something?

aerubanov: I mean something like
```python
if hasattr(node.op, "fgraph"):
    inner_graph = node.op.fgraph.clone()
    self.rewrite(inner_graph)
    node.op.fgraph = inner_graph
```
should be enough?

ricardoV94: Unfortunately we can't just modify the fgraph, as that's just a dynamic property of the Op. In some places it may be used directly, but that's not guaranteed. The Op inner inputs / outputs are the real properties of the Op, which means we need to create a new Scan op that uses the new outputs. Recreating a Scan is pretty annoying, you can check pytensor scan rewrites to see how it's approached there.

tests/distributions/test_distribution.py
```diff
@@ -473,25 +473,22 @@ def test_custom_dist_default_moment(self, dist_params, size, expected, dist_fn):
...
-
             xs, updates = scan(
                 fn=scan_step,
-                outputs_info=[x0],
-                n_steps=2,
+                sequences=[pt.ones(2)],
                 name="xs",
             )
-            return xs[-1]
+            return pt.sum(xs)
```
ricardoV94: `pt.sum` gives a variable for which PyMC doesn't know the logp. You can return the whole `xs` that should be fine. Would be good to use something like a `Uniform(-2, -1)` inside the Scan, for which the old `CustomDist` moment of zeros would fail. 

And then you can remove the `check_finite_logp=False`

pymc/distributions/distribution.py
```diff
@@ -622,13 +679,16 @@ def dist(
         if logcdf is None:
             logcdf = default_not_implemented(class_name, "logcdf")
 
+        def dist_moment(rv, size, *dist_params):
+            # TODO: add check for other op with inner graph (not Scan)
+            fgraph = FunctionGraph(outputs=[dist(*dist_params, size=size)], clone=True)
```
aerubanov: @ricardoV94 do we need some check here for` OpFromGraph`, `Composite`, and `ScalarLoop` Ops in `fgraph`? I mean to use default moment like we did before in this case.

ricardoV94: Yes, let's raise `NotImplementedError` for those, put the new logic in a `try/except` and default back to the zeros like before when it raises. 

I think we should put this `dist_moment` helper inside the old `default_moment` function?

ricardoV94: Wait only `OpFromGraph` is problematic. The other remaining cases can't have `RandomVariables` inside so we can just replace their inputs outside without concern for their inner graph.

Let's only raise when it has InnerGraph and it's not Scan/Composite/ScalarLoop, which are the cases we can safely handle.

ricardoV94: Then we can add a github issue to also implemnet the OpFromGraph case which shouldn't be too hard (similar to scan). Or do you want to do that in this PR after all? That would neatly conclude all cases.

aerubanov: @ricardoV94 Well, I think I want to implement  the OpFromGraph case in this PR. Do you know some example on how to use OpFromGraph? I need to create some test case for it

ricardoV94: I'll write something tomorrow. Thanks for the initiative in advance!

aerubanov:
@ricardoV94 just friendly reminder about example for `OpFromGraph` )

aerubanov:
@aerubanov Yeah, I need to add test case for `pm.Truncated` because I have not it now. And may be will need to add some changes in my implementation, because I did not take into account this case when was worked on my implementation. So thank you for pointing out on this!

pymc/distributions/distribution.py
```diff
@@ -83,6 +85,61 @@
...
+    def apply(self, fgraph):
+        for node in fgraph.toposort():
+            if isinstance(node.op, Scan):
+                # inner_graph = node.op.fgraph.clone()
+                # self.rewrite(inner_graph)
+                # node.op.fgraph = inner_graph
+                new_node = rewrite_moment_scan_node(node)
+                for out1, out2 in zip(node.outputs, new_node.outputs):
+                    fgraph.replace(out1, out2)
+            elif isinstance(node.op, (RandomVariable, SymbolicRandomVariable)):
```
ricardoV94: I think this branch should come first

pymc/distributions/distribution.py
```diff
@@ -687,6 +747,7 @@ def custom_dist_logcdf(op, value, size, *params, **kwargs):
 
         @_moment.register(rv_type)
         def custom_dist_get_moment(op, rv, size, *params):
+            params = [i for i in params if not isinstance(i, RandomGeneratorSharedVariable)]
```
ricardoV94: Why is this needed?

pymc/distributions/distribution.py
ricardoV94: Should we specify the FunctionGraph inputs? I guess something like:

```python
rv = dist(*dis_params, size=size) 
# Filter shared variables / constants that I think cannot be inputs to the FunctionGraph
filtered_inputs = [inp for inp in rv.owner.inputs if inp...]
fgraph = FunctionGraph(inputs=filtered_inputs, outputs=rv, clone=True)
```

pymc/distributions/distribution.py
aerubanov: `params` consists `RandomGeneratorSharedVariable`s, which are not match with `dist` function signature. So I filter out it here, but may be there is a better way to do it

pymc/distributions/distribution.py
aerubanov: I experimented with this approach a little bit and it is not working. We need a whole graph, but `rv.owner.inputs` get us only last node and I got `MissingInputError`. For example for  `pt.exp(pm.Normal.dist(mu, sigma, size=size))` dist function `rv.owner.inputs` is just `[normal_rv{0, (0, 0), floatX, False}.out]`

pymc/distributions/distribution.py
ricardoV94: Ah, I think below we do something like `params[:len(dist_params)]`? Does that work? Would be nice to add a comment to say we are excluding the shared RNGs

pymc/distributions/distribution.py
ricardoV94: Hmm, that doesn't sound right. `dist` in a CustomDist should return an `OpFromGraph`, whose inputs would be `size` `mu`, `sigma`, and the shared RNGs.

```python
import pymc as pm

def dist_fn(mu, sigma, size):
    return pm.math.exp(pm.Normal.dist(mu, sigma, shape=size))

rv = pm.CustomDist.dist(0, 1, dist=dist_fn)
rv.owner.inputs
# [TensorConstant(TensorType(int64, shape=(0,)), data=array([], ... ype=int64)),
#  TensorConstant(TensorType(int8, shape=()), data=array(0, dtype=int8)),
#  TensorConstant(TensorType(int8, shape=()), data=array(1, dtype=int8)),
#  RandomGeneratorSharedVariable(<Generator(PCG64) at 0x7F61326D0820>)]
```

The point is that we don't want to rewrite the graph above the CustomDist.

ricardoV94: Actually we can just apply the moment in the `rv.owner.op.fgraph.outputs` (after cloning it) and return that? More specifically on the output that corresponds to `rv`?

aerubanov: @ricardoV94  now we take output of `dist_fn` (not `CustomDist.dist`), build fgraph  and apply moment to it nodes. Do you think we need to take `CustomDist.dist` instead of `dist_fn`?

pymc/distributions/distribution.py
aerubanov: I think `params[:len(dist_params)]` should work here, but do we have access to `dist_params` from this function? Looks like no.

ricardoV94: In that case let's keep like you did, but perhaps use a helper function with a readable name? Could that helper be used below?

pymc/distributions/distribution.py
aerubanov: Will add comment

ricardoV94: > @ricardoV94  now we take output of `dist_fn` (not `CustomDist.dist`), build fgraph  and apply moment to it nodes. Do you think we need to take `CustomDist.dist` instead of `dist_fn`? 

Yes that's better, because it already takes care of encapsulating the graph that corresponds to the CustomDist

aerubanov: I think we can not use `CustomDist.dist` because we need to build `moment` inside `dist` method

pymc/distributions/distribution.py
aerubanov: I added helper function to filter out shared RNGs

pymc/distributions/distribution.py
ricardoV94: Can we extract the code that creates the fgraph from user provided fn and inputs into a helper outside of `CustomDist.dist`? That was the only reason I was suggesting we call it again.

aerubanov: @ricardoV94 I found that `FunctionGraph` can extract and filter graph inputs on its own:
 ```
         if inputs is None:
            inputs = [
                i for i in graph_inputs(outputs) if not isinstance(i, AtomicVariable)
            ]
```
 https://github.com/pymc-devs/pytensor/blob/7bb18f3a3590d47132245b7868b3a4a6587a4667/pytensor/graph/fg.py#L105
 so I think we need not to specify inputs manually. But I moved graph construction in helper function anyway.

File: pymc/distributions/distribution.py [hunk 8]
```diff
@@ -83,6 +85,60 @@
 PLATFORM = sys.platform
 
 
+def filter_shared_RNGs(params):
+    return [p for p in params if not isinstance(p, RandomGeneratorSharedVariable)]
```
ricardoV94: This should be more general and still apply to your use case.
```suggestion
def filter_RNGs(params):
    return [p for p in params if not isinstance(p.type, RandomType)]
```

pymc/distributions/distribution.py
```diff
@@ -83,6 +85,60 @@
...
+
+
+def rewrite_moment_scan_node(node):
```
ricardoV94: This can be a method of `MomentRewriter` just to keep things more tidy

pymc/distributions/distribution.py
```diff
@@ -622,13 +685,15 @@ def dist(
         if logcdf is None:
             logcdf = default_not_implemented(class_name, "logcdf")
 
+        def dist_moment(rv, size, *dist_params):
+            fgraph = get_rv_fgraph(dist, dist_params, size)
+            replace_moments = MomentRewrite()
```
ricardoV94: I am still worried that this will rewrite `dist_params` as well, and not just the CustomDist graph between `dist` and `dist_params`. 

Would the following work? Clone the inner fgraph from `rv.owner.op.fgraph`. This graph doesn't have `dist_params` directly but dummy placeholders called `NominalVariable`s. Apply the rewrite on this inner graph, and once you are done, replace the `NominalVariable`s by the respective `dist_params`?

ricardoV94 (commented):
`CustomDist` is getting more complex, we might want to move it into its own file later down the road.

pymc/distributions/distribution.py
ricardoV94: Actually what I am suggesting is pretty similar to what you did with `Scan`, so you could perhaps reuse some of the same logic?

pymc/distributions/distribution.py
```diff
@@ -83,6 +85,60 @@
...
+        if nd not in to_replace_set and isinstance(nd.op, (RandomVariable, SymbolicRandomVariable)):
+            replace_with_moment.append(nd.out)
+            to_replace_set.add(nd)
+    givens = {}
+    if len(replace_with_moment) > 0:
+        for item in replace_with_moment:
+            givens[item] = moment(item)
+    op_outs = clone_replace(node_outputs, replace=givens)
+
+    nwScan = Scan(
```
ricardoV94: Is there a way to check if there was actually any replacement done? If not, there's no point in creating a new Scan

aerubanov: Yeah, I think we can add this check

pymc/distributions/distribution.py
aerubanov: @ricardoV94  When I trying to create `FunctionGraph` with `NominalVariable`s I get error about missing input values - I think `FunctionGraph` constructor do not support dummy placeholders as input. But I`m try another approach:
```python
def dist_moment(rv, size, *dist_params, dist):
    rv = dist(*dist_params, size=size)
    inputs, outputs = dist_params, [rv.owner.out]
    fgraph_topo = io_toposort(inputs, outputs)
    replace_with_moment = []
    to_replace_set = set()

    for nd in fgraph_topo:
          if nd not in to_replace_set and isinstance(
               nd.op, (RandomVariable, SymbolicRandomVariable)
          ):
                replace_with_moment.append(nd.out)
                to_replace_set.add(nd)
     givens = {}
     for item in replace_with_moment:
            givens[item] = moment(item)
     [out] = clone_replace(outputs, replace=givens)
     return out
```
Looks like it work but do not support `Scan` for now (but I can add it). Do you think such approach will be better?

ricardoV94: You don't have to recreate the `FunctionGraph`, you can just do `fgraph.clone()`?

ricardoV94:
BTW @aerubanov although this is a tricky one it seems to be getting really good! Great work so far

pymc/distributions/distribution.py
aerubanov: @ricardoV94 I try this approach, but I need to create `CustomDist` op first because just output of `dist` function do not have `fgraph` attribute.
```python
def dist_moment(rv, size, *dist_params):
            size = normalize_size_param(size)
            dummy_size_param = size.type()
            dummy_dist_params = [dist_param.type() for dist_param in dist_params]
            dummy_rv = dist(*dummy_dist_params, dummy_size_param)
            dummy_params = [dummy_size_param] + dummy_dist_params
            rv_type = type(
                class_name,
                (CustomSymbolicDistRV,),
                # If logp is not provided, we try to infer it from the dist graph
                dict(
                    inline_logprob=logp is None,
                ),
            )
            rv_op = rv_type(
                inputs=dummy_params,
                outputs=[dummy_rv],
                ndim_supp=ndim_supp,
            )
            fgraph = rv_op.fgraph.clone()
            replace_moments = MomentRewrite()
            replace_moments.rewrite(fgraph)
            for i, par in enumerate([size] + list(dist_params)):
                fgraph.replace(fgraph.inputs[i], par)
            [moment] = fgraph.outputs
            return moment
```
Do you think this way better? It is  also do not work with `Scan` yet. I need to figure out why and I will move graph creation logic to helper function if we will decide to keep this way.

ricardoV94: You have fgraph in the `rv.owner.op` from the `rv` that's passed as input.

About Scan what exactly is the problem? The approach you had written already seemed fine.  Was it failing?

aerubanov: For scan I got Assertion Error:
```
        # allocate storage for intermediate computation
        for node in order:
            for r in node.inputs:
                if r not in storage_map:
>                   assert isinstance(r, Constant)
E                   AssertionError

../../miniconda3/envs/pymc-dev/lib/python3.11/site-packages/pytensor/link/utils.py:135: AssertionError
```
It is hapens when I run `TestCustomSymbolicDist::test_custom_dist_default_moment_inner_graph` test case

aerubanov: @ricardoV94 If I create rv like this it`s do not have fgraph atribute:
```
rv = dist(*dist_params, size=size)
>>> rv
Exp.0
>>> rv.owner
Exp(normal_rv{0, (0, 0), floatX, False}.out)
>>> rv.owner.op
Elemwise(scalar_op=exp,inplace_pattern=<frozendict {}>)
>>> rv.owner.op.fgraph
*** AttributeError: 'Elemwise' object has no attribute 'fgraph'
```
I use `lambda mu, sigma, size: pt.exp(pm.Normal.dist(mu, sigma, size=size))` for `dist` function  here.

ricardoV94: No, use the `rv` that is passed to the moment function directly (first argument), don't recreate it

aerubanov: Ohm yeah, my bad. Will try it

aerubanov: @ricardoV94 No, I can not use `rv` that is passes as parameter:
```
>>> rv
*** Not yet returned!
```
So I tried to re-create it

pymc/distributions/distribution.py
```diff
@@ -622,13 +682,38 @@ def dist(
         if logcdf is None:
             logcdf = default_not_implemented(class_name, "logcdf")
 
-        if moment is None:
-            moment = functools.partial(
-                default_moment,
-                rv_name=class_name,
-                has_fallback=True,
+        def dist_moment(rv, size, *dist_params):
```
ricardoV94: Why can't you use the fgraph of the `rv` here? I don't follow

ricardoV94: Here is what I have in mind:

```python
def dist_moment(rv, *rv_inputs):
  rv_op = rv.owner.op
  fgraph = rv_op.fgraph.clone()
  replace_moments = MomentRewrite()
  replace_moments.rewrite(fgraph)
  fgraph.replace_all(tuple(zip(rv_op.inner_inputs, rv_inputs)))
  [moment] = fgraph.outputs
  return moment
```

aerubanov: @ricardoV94 because I get
```
>>> rv
*** Not yet returned!
```
Not sure why actually

ricardoV94: I don't know what Not yet returned means?

aerubanov: @ricardoV94 I just used pdb wrong, we do not need re-create rv here.

aerubanov:
@ricardoV94 Could you please take a look on recent changes? Just friendly reminder)

pymc/distributions/distribution.py
```diff
@@ -622,13 +683,28 @@ def dist(
         if logcdf is None:
             logcdf = default_not_implemented(class_name, "logcdf")
 
+        def dist_moment(rv, size, *dist_params):
+            fgraph = rv.owner.op.fgraph.clone()
```
ricardoV94: This method can now be used for any OpFromGraph not just CustomDist ones, so I would move it to the MomentRewriter

pymc/distributions/distribution.py
```diff
@@ -622,13 +683,28 @@ def dist(
...
+            replace_moments = MomentRewrite()
+            replace_moments.rewrite(fgraph)
+            # we need to replace dummy variable by actual dist params
+            for i, par in enumerate([size] + list(dist_params)):
+                fgraph.replace(fgraph.inputs[i], par)
```
ricardoV94: You can use `fgraph.replace_all` to avoid the loop

pymc/distributions/distribution.py
```diff
@@ -622,13 +683,28 @@ def dist(
...
+            # we need to replace dymmy random generators in Scan node inputs
+            for node in fgraph.toposort():
+                if isinstance(node.op, Scan):
```
ricardoV94: Why do we need a special logic for Scan inside an OpFromGraph?

aerubanov: @ricardoV94 After cloning fgraph we have dummy placeholders in `Scan` inner graph inputs for `RandomGenerator`, and we need to replace it by real random generator to avoid error later when we will try to calculate moment. Do we have another way to do it?

ricardoV94: I see, so the problem is that the scan moment rewrite does not remove the useless updates. That's probably okay. The issue is that we filtered out the RNGs!!! If we had them the replacement would have worked fine. I was able to simplify the function to this and it passes the test that has a scan:

```python
        def dist_moment(rv, *args):
            node = rv.owner
            op = node.op
            rv_out_idx = node.outputs.index(rv)
            
            fgraph = op.fgraph.clone()
            replace_moments = MomentRewrite()
            replace_moments.rewrite(fgraph)
            # Replace dummy inner inputs by outer inputs
            fgraph.replace_all(tuple(zip(op.inner_inputs, node.inputs)), import_missing=True)
            moment = fgraph.outputs[rv_out_idx]
            return moment
```

Note that I basically got a handle of the filtered RNGs by doing node.inputs. We don't need to do this if we don't filter them out.

To tidy up the code, we should dispatch this moment function on `CustomSymbolicDistRV` outside of the `_CustomSymbolicDist` class. Then we only dispatch a user provided moment when that's given like we do now with `if logp is not None:`

You can remove the moment related stuff that now exists in `_CustomSymbolicDist.dist`

aerubanov: @ricardoV94 Great, thank you! Will do

tests/distributions/test_distribution.py
```diff
@@ -430,6 +430,85 @@ def custom_dist(mu, sigma, size):
...
+                lambda mu1, sigma, size: pm.Normal.dist(mu1, sigma, size=size),
+            ),
+        ],
+    )
+    def test_custom_dist_default_moment(self, dist_params, size, expected, dist_fn):
+        with Model() as model:
+            CustomDist("x", *dist_params, dist=dist_fn, size=size)
+        assert_moment_is_expected(model, expected)
+
+    def test_custom_dist_default_moment_inner_graph(self):
```
ricardoV94:
```suggestion
    def test_custom_dist_default_moment_scan(self):
```

tests/distributions/test_distribution.py
```diff
@@ -430,6 +430,85 @@ def custom_dist(mu, sigma, size):
...
+        def scan_step(left, right):
```
ricardoV94, line 476: Can we add a second test with recurring inputs like `step = lambda xtm1: pm.Normal.dist(xtm1)` (plus the updates of course)?

pymc/distributions/distribution.py
```diff
@@ -83,6 +85,64 @@
...
+
+    def apply(self, fgraph):
+        for node in fgraph.toposort():
+            if isinstance(node.op, (RandomVariable, SymbolicRandomVariable)):
+                fgraph.replace(node.out, moment(node.out))
+            elif isinstance(node.op, Scan):
+                new_node = self.rewrite_moment_scan_node(node)
+                if new_node is not None:
+                    for out1, out2 in zip(node.outputs, new_node.outputs):
+                        fgraph.replace(out1, out2)
```
ricardoV94: Use replace_all?

pymc/distributions/distribution.py
```diff
@@ -687,7 +756,7 @@ def custom_dist_logcdf(op, value, size, *params, **kwargs):
 
         @_moment.register(rv_type)
         def custom_dist_get_moment(op, rv, size, *params):
-            return moment(rv, size, *params[: len(params)])
+            return moment(rv, size, *filter_RNGs(params))
```
ricardoV94: Only register this when `moment is not None`. And the default case you can register in the base class `CustomSymbolicDistRV` outside

pymc/distributions/distribution.py
```diff
@@ -623,12 +697,7 @@ def dist(
             logcdf = default_not_implemented(class_name, "logcdf")
 
         if moment is None:
-            moment = functools.partial(
-                default_moment,
-                rv_name=class_name,
-                has_fallback=True,
-                ndim_supp=ndim_supp,
-            )
+            moment = dist_moment
```
ricardoV94: Remove this if statement (see other comment about default)

tests/distributions/test_distribution.py
```diff
@@ -430,6 +430,104 @@ def custom_dist(mu, sigma, size):
...
+            )
+            return xs
+
+        with Model() as model:
+            CustomDist("x", dist=dist)
+        assert_moment_is_expected(model, np.array([-3, -2]))
+
+    def test_custom_dist_default_moment_scan_recurring(self):
+        def scan_step(xtm1):
+            x = pm.Normal.dist(xtm1)
```
ricardoV94: Let's try something that requires the recursion to provide a correct output
```suggestion
            x = pm.Normal.dist(xtm1 + 1)
```

ricardoV94 (commented):
This looks really great. Just some minor cleaning up suggestions left

pymc/distributions/distribution.py
```diff
@@ -83,6 +85,64 @@
 PLATFORM = sys.platform
 
 
+def filter_RNGs(params):
+    return [p for p in params if not isinstance(p.type, (RandomType, RandomGeneratorType))]
```
ricardoV94: Since this ended up not being needed let's just inline it in the custom moment function

pymc/distributions/distribution.py
```diff
@@ -601,6 +661,20 @@ def update(self, node: Node):
...
+def dist_moment(rv, *args):
+    node = rv.owner
+    op = node.op
+    rv_out_idx = node.outputs.index(rv)
+
+    fgraph = op.fgraph.clone()
+    replace_moments = MomentRewrite()
+    replace_moments.rewrite(fgraph)
+    # Replace dummy inner inputs by outer inputs
+    fgraph.replace_all(tuple(zip(op.inner_inputs, node.inputs)), import_missing=True)
```
ricardoV94: Can use `args` once you dispatch this to the default case, as we won't be filtering the RVs anymore.
I suggest calling it `*inputs`

```suggestion
    fgraph.replace_all(tuple(zip(op.inner_inputs, args)), import_missing=True)
```

pymc/distributions/distribution.py
aerubanov: @ricardoV94  Hm, I think need this condition to avoid overriding moment provided by user, how can we avoid it without condition?

aerubanov: More specifically, If I try to remove  if statement `TestCustomSymbolicDist::test_custom_methods` fails

ricardoV94: There are two levels. We dispatch the general moment to the parent base class. Whenever a new subclass is created here and the user provided a moment we register it on the subclass (so it has preference). If the user didn't provide anything, we don't register and the parent class one will be used.

pymc/distributions/distribution.py
```diff
@@ -696,8 +695,8 @@ def dist(
         if logcdf is None:
             logcdf = default_not_implemented(class_name, "logcdf")
 
-        if moment is None:
-            moment = dist_moment
+        # if moment is None:
```
ricardoV94: Also remove `moment = dist_moment`

pymc/distributions/distribution.py
```diff
@@ -601,6 +656,20 @@ def update(self, node: Node):
         return updates
 
 
+def dist_moment(rv, *args):
```
ricardoV94: In case it was confusing, this is how you would register it in the base class. Then you don't need to do it everytime a `CustomDist` is called, only when the user wants to provide a non-default moment

```suggestion
@_moment.register(CustomSymbolicDistRV)
def dist_moment(op, rv, *args):
```

pymc/distributions/distribution.py
```diff
@@ -601,6 +656,22 @@ def update(self, node: Node):
         return updates
 
 
+@_moment.register(CustomSymbolicDistRV)
+def dist_moment(rv, *args):
+    x = args[0]
```
aerubanov: @ricardoV94 I not very like this `args[0]` here and `args[1]` bellow, I mean magic numbers for index. Do you think we can avoid it somehow?

pymc/distributions/distribution.py
```diff
@@ -601,6 +656,22 @@ def update(self, node: Node):
         return updates
 
 
+@_moment.register(CustomSymbolicDistRV)
+def dist_moment(rv, *args):
```
ricardoV94: First argument in now op when you register. It also means you don't need to recover it from the node inside

```suggestion
def dist_moment(op, rv, *args):
```

aerubanov: Yeah, thank you!

pymc/distributions/distribution.py
```diff
@@ -601,6 +656,21 @@ def update(self, node: Node):
         return updates
 
 
+@_moment.register(CustomSymbolicDistRV)
+def dist_moment(op, rv, *args):
+    node = rv.owner
+    op = node.op
```
ricardoV94: Last nitpick, we already have op from the signature
```suggestion
```

ricardoV94 (approved):

ricardoV94:
Great work @aerubanov!

PR #6897, opened 2023-09-08
Allow batched parameters in MvNormal and MvStudentT distributions

ricardoV94:
Also provides default zero mu

Related to #5383

----
:books: Documentation preview :books:: https://pymc--6897.org.readthedocs.build/en/6897/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -115,40 +115,37 @@ nan_lower_cholesky = partial(cholesky, lower=True, on_error="nan")
 
 
 def quaddist_matrix(cov=None, chol=None, tau=None, lower=True, *args, **kwargs):
-    if chol is not None and not lower:
-        chol = chol.T
-
     if len([i for i in [tau, cov, chol] if i is not None]) != 1:
         raise ValueError("Incompatible parameterization. Specify exactly one of tau, cov, or chol.")
 
     if cov is not None:
         cov = pt.as_tensor_variable(cov)
-        if cov.ndim != 2:
-            raise ValueError("cov must be two dimensional.")
+        if cov.ndim < 2:
+            raise ValueError("cov must be at least two dimensional.")
     elif tau is not None:
         tau = pt.as_tensor_variable(tau)
-        if tau.ndim != 2:
-            raise ValueError("tau must be two dimensional.")
-        # TODO: What's the correct order/approach (in the non-square case)?
-        # `pytensor.tensor.nlinalg.tensorinv`?
+        if tau.ndim < 2:
+            raise ValueError("tau must be at least two dimensional.")
         cov = matrix_inverse(tau)
     else:
-        # TODO: What's the correct order/approach (in the non-square case)?
         chol = pt.as_tensor_variable(chol)
-        if chol.ndim != 2:
-            raise ValueError("chol must be two dimensional.")
+        if chol.ndim < 2:
+            raise ValueError("chol must be at least two dimensional.")
+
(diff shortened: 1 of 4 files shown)


twiecki:
Is this enabled by the new `pt.vectorize`?

ricardoV94:
> Is this enabled by the new `pt.vectorize`?

Yes

pymc/distributions/multivariate.py
```diff
@@ -391,7 +357,7 @@ class MvStudentT(Continuous):
     rv_op = mv_studentt
 
     @classmethod
-    def dist(cls, nu, Sigma=None, mu=None, scale=None, tau=None, chol=None, lower=True, **kwargs):
+    def dist(cls, nu, *, Sigma=None, mu, scale=None, tau=None, chol=None, lower=True, **kwargs):
```
fonnesbeck: Sigma is optional but mu is not?

ricardoV94:
The failing tests are fixed by #6903 and latest patch release of PyTensor

pymc/distributions/multivariate.py
ricardoV94: You need to provide one of the several alternatives (chol, tau, etc...), it raises later if none is given

aseyboldt: I think it would be nice if an unspecified (so None) mu would be equivalent to passing mu=zeros(n). And the same for the MvNormal?
Seems this wasn't happening before though...

ricardoV94:
Need to sort out https://github.com/pymc-devs/pytensor/issues/451 first

tests/distributions/test_multivariate.py
```diff
@@ -330,6 +330,41 @@ def test_mvnormal_init_fail(self):
             with pytest.raises(ValueError):
                 x = pm.MvNormal("x", mu=np.zeros(3), cov=np.eye(3), tau=np.eye(3), size=3)
 
+    @pytest.mark.parametrize("batch_mu", (False, True))
+    @pytest.mark.parametrize("batch_cov", (False, True))
+    @pytest.mark.parametrize("cov_form", ("cov", "tau", "chol_lower", "chol_upper"))
```
ricardoV94, line 337: @fonnesbeck now testing `chol` parametrizations as well. That was an oversight on my end :)

pymc/distributions/multivariate.py
ricardoV94: Mu now defaults to zero in both MvNormal and MvStudentT

ricardoV94:
This now depends on #6934

pymc/distributions/multivariate.py
```diff
@@ -333,14 +294,21 @@ def _supp_shape_from_params(self, dist_params, param_shapes=None):
...
     @classmethod
     def rng_fn(cls, rng, nu, mu, cov, size):
+        if size is None:
+            # When size is implicit, we need to broadcast parameters correctly,
+            # so that the MvNormal draws and the chisquare draws have the same number of batch dimensions.
+            # nu broadcasts mu and cov
+            if np.ndim(nu) > max(mu.ndim - 1, cov.ndim - 2):
+                _, mu, cov = broadcast_params((nu, mu, cov), ndims_params=cls.ndims_params)
+            # nu is broadcasted by either mu or cov
+            elif np.ndim(nu) < max(mu.ndim - 1, cov.ndim - 2):
```
fonnesbeck, line 304: Is there no equivalence condition?

fonnesbeck (approved):
LGTM

Issue #5674, opened 2022-03-31
Implement `default_transform` and `transform` argument for distributions


ricardoV94:
### Discussed in https://github.com/pymc-devs/pymc/discussions/5404

Most of the times when users apply a custom transform, they do it because they want to "distort" the base distribution (i.e., to impose ordering, sum to a constant, centered parametrization). This however, overrides default transforms that are set for constrained distributions, leading to inefficient sampling. 

The correct thing to do in this case would be the Chain the default transform with the new one, so that the "distorted" distribution is still stable during sampling. However, this is pretty verbose and most users are not even aware of the default transform behavior.

The proposed solution is to implement a new transform keyword `default_transform`, that behaves like `transform` now does. The `transform` keyword, if set, would then be chained together with whatever the `default_transform` is (or used as is when a distribution does not have a default transform).

If users want to disable any transform whatsoever, they will now have to set `default_transform` to `None`. This is usually only done for pedagogical reasons. We can detect if users are trying to do this the old way, when they set `transform=None` (the default is a flag variable `UNSET`),  and raise an error that explains the changed behavior. Setting `transform=None` would never have an effect because that is always the default behavior.

***

In addition we can use this information raise a warning when users do prior predictive, which are generally invalid for "distorted" distributions. This is very similar to what we already do for models containing `Potentials` for the very same reasons

https://github.com/pymc-devs/pymc/blob/80b5e86da5d0f55bee59a7b2b8dc9ecee31af2dd/pymc/sampling.py#L2014-L2020

PR #6903, opened 2023-09-12
Fix bug in univariate Ordered and SumTo1 transform logp

ricardoV94:
The logic introduced in #6255 was wrong because the jacobian of the multivariate transform would broadcast with the univariate logp (counting it repeatedly!).

When one applies a transform that combines information across batch dimensions, those can no longer be considered independent. An ordered uniform is in fact a multivariate RV. Accordingly, after this PR the logp of the base RVs is collapsed before adding the respective jacobian term.

With the fix it's no longer needed to distinguish between `univariate` and `multivariate` cases.

CC @TimOliverMaier 

***

Note: It would be nice to use `SymbolicRandomVariable`s for these kind of transforms, so that PyMC logprob inference has accurate information about the ndim support of the variables. This would also allow us to return "correct" forward draws, although I am not sure what those correspond to (rejection sampling for Ordered and something like what the ZeroSumNormal does for the SumTo1)? This is related to #6360 and #5674 

----
:books: Documentation preview :books:: https://pymc--6903.org.readthedocs.build/en/6903/

### DIFF
--- a/pymc/distributions/transforms.py
+++ b/pymc/distributions/transforms.py
@@ -96,14 +96,14 @@ class Ordered(RVTransform):
 
     def backward(self, value, *inputs):
         x = pt.zeros(value.shape)
-        x = pt.inc_subtensor(x[..., 0], value[..., 0])
-        x = pt.inc_subtensor(x[..., 1:], pt.exp(value[..., 1:]))
+        x = pt.set_subtensor(x[..., 0], value[..., 0])
+        x = pt.set_subtensor(x[..., 1:], pt.exp(value[..., 1:]))
         return pt.cumsum(x, axis=-1)
 
     def forward(self, value, *inputs):
         y = pt.zeros(value.shape)
-        y = pt.inc_subtensor(y[..., 0], value[..., 0])
-        y = pt.inc_subtensor(y[..., 1:], pt.log(value[..., 1:] - value[..., :-1]))
+        y = pt.set_subtensor(y[..., 0], value[..., 0])
+        y = pt.set_subtensor(y[..., 1:], pt.log(value[..., 1:] - value[..., :-1]))
         return y
 
     def log_jac_det(self, value, *inputs):
--- a/pymc/distributions/transforms.py
+++ b/pymc/distributions/transforms.py
@@ -11,6 +11,8 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+import warnings
+
 from functools import singledispatch
 
 import numpy as np
@@ -39,17 +41,26 @@ __all__ = [
     "logodds",
     "Interval",
     "log_exp_m1",
-    "univariate_ordered",
-    "multivariate_ordered",
+    "ordered",
     "log",
     "sum_to_1",
-    "univariate_sum_to_1",
-    "multivariate_sum_to_1",
     "circular",
     "CholeskyCovPacked",
     "Chain",
     "ZeroSumTransform",
 ]
+
+
+def __getattr__(name):
+    if name in ("univariate_ordered", "multivariate_ordered"):
+        warnings.warn(f"{name} has been deprecated, use ordered instead.", FutureWarning)
+        return ordered
+
+    if name in ("univariate_sum_to_1, multivariate_sum_to_1"):
+        warnings.warn(f"{name} has been deprecated, use sum_to_1 instead.", FutureWarning)
+        return sum_to_1
+
+    raise AttributeError(f"module {__name__} has no attribute {name}")
 
 
 @singledispatch
@@ -79,13 +90,9 @@ class LogExpM1(RVTransform):
 class Ordered(RVTransform):
     name = "ordered"
 
-    def __init__(self, ndim_supp=0):
-        if ndim_supp > 1:
-            raise ValueError(
-                f"For Ordered transformation number of core dimensions"
-                f"(ndim_supp) must not exceed 1 but is {ndim_supp}"
-            )
-        self.ndim_supp = ndim_supp
+    def __init__(self, ndim_supp=None):
+        if ndim_supp is not None:
+            warnings.warn("ndim_supp argument is deprecated and has no effect", FutureWarning)
 
     def backward(self, value, *inputs):
         x = pt.zeros(value.shape)
@@ -100,10 +107,7 @@ class Ordered(RVTransform):
         return y
 
     def log_jac_det(self, value, *inputs):
-        if self.ndim_supp == 0:
-            return pt.sum(value[..., 1:], axis=-1, keepdims=True)
-        else:
-            return pt.sum(value[..., 1:], axis=-1)
+        return pt.sum(value[..., 1:], axis=-1)
 
 
 class SumTo1(RVTransform):
@@ -114,13 +118,9 @@ class SumTo1(RVTransform):
 
     name = "sumto1"
 
-    def __init__(self, ndim_supp=0):
-        if ndim_supp > 1:
-            raise ValueError(
-                f"For SumTo1 transformation number of core dimensions"
-                f"(ndim_supp) must not exceed 1 but is {ndim_supp}"
-            )
-        self.ndim_supp = ndim_supp
+    def __init__(self, ndim_supp=None):
+        if ndim_supp is not None:
+            warnings.warn("ndim_supp argument is deprecated and has no effect", FutureWarning)
 
     def backward(self, value, *inputs):
         remaining = 1 - pt.sum(value[..., :], axis=-1, keepdims=True)
@@ -131,10 +131,7 @@ class SumTo1(RVTransform):
 
     def log_jac_det(self, value, *inputs):
         y = pt.zeros(value.shape)
-        if self.ndim_supp == 0:
-            return pt.sum(y, axis=-1, keepdims=True)
-        else:
-            return pt.sum(y, axis=-1)
+        return pt.sum(y, axis=-1)
 
 
 class CholeskyCovPacked(RVTransform):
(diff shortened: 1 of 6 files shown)


tests/distributions/test_transform.py
```diff
@@ -345,56 +345,13 @@ def check_transform_elementwise_logp(self, model):
             .sum()
             .eval({x_val_untransf: test_array_untransf})
         )
-        close_to(v1, v2, tol)
+        log_jac_det_eval = jacob_det.sum().eval({x_val_transf: test_array_transf})
```
ricardoV94: Testing the jacobian provides a regression check

pymc/distributions/transforms.py
```diff
@@ -79,13 +82,9 @@ def log_jac_det(self, value, *inputs):
...
-    def __init__(self, ndim_supp=0):
-        if ndim_supp > 1:
-            raise ValueError(
-                f"For Ordered transformation number of core dimensions"
-                f"(ndim_supp) must not exceed 1 but is {ndim_supp}"
-            )
-        self.ndim_supp = ndim_supp
+    def __init__(self, ndim_supp=None):
+        if ndim_supp is not None:
+            warnings.warn("ndim_supp argument is deprecated", FutureWarning)
```
lucianopaz:
```suggestion
            warnings.warn("ndim_supp argument is deprecated and its value will be ignored", FutureWarning)
```

pymc/distributions/transforms.py
```diff
@@ -79,13 +82,9 @@ def log_jac_det(self, value, *inputs):
...
 
     def backward(self, value, *inputs):
         x = pt.zeros(value.shape)
```
lucianopaz, line 98: Why does `backward` and `forward` use `inc_subtensor` instead of `set_subtensor`? Is there something that I'm missing? The former incurs in some useless computation.

pymc/distributions/transforms.py
```diff
@@ -114,13 +110,9 @@ class SumTo1(RVTransform):
...
-    def __init__(self, ndim_supp=0):
-        if ndim_supp > 1:
-            raise ValueError(
-                f"For SumTo1 transformation number of core dimensions"
-                f"(ndim_supp) must not exceed 1 but is {ndim_supp}"
-            )
-        self.ndim_supp = ndim_supp
+    def __init__(self, ndim_supp=None):
+        if ndim_supp is not None:
+            warnings.warn("ndim_supp argument is deprecated", FutureWarning)
```
lucianopaz:
```suggestion
            warnings.warn("ndim_supp argument is deprecated and its value will be ignored", FutureWarning)
```

pymc/distributions/transforms.py
```diff
@@ -359,12 +348,20 @@ def extend_axis_rev(array, axis):
...
 
-univariate_ordered = Ordered(ndim_supp=0)
+# Deprecated
+ordered = Ordered()
+ordered.__doc__ = """
+Instantiation of :class:`pymc.distributions.transforms.Ordered`
+for use in the ``transform`` argument of a random variable."""
+
+# Deprecated
+univariate_ordered = Ordered()
```
lucianopaz: You should do some `deprecation` of this. Maybe add a `__getattr__` to the module ([as in this answer](https://stackoverflow.com/a/48242860/7253166)) that issues a deprecation warning for the `univariate_*` and `multivariate_*` transforms. Another thing I'd do is to state that these attributes will be removed in version 5.some_number.

pymc/logprob/transforms.py
```diff
@@ -1229,6 +1232,10 @@ def transformed_logprob(op, values, *inputs, use_jacobian=True, **kwargs):
                 assert isinstance(value.owner.op, TransformedVariable)
                 original_forward_value = value.owner.inputs[1]
                 jacobian = transform.log_jac_det(original_forward_value, *inputs).copy()
+                # Check if jacobian dimensions align with logp
+                if jacobian.ndim < logp.ndim:
+                    diff_ndims = logp.ndim - jacobian.ndim
```
lucianopaz: I would love if you only had to test whether the `jacobian` and `logp` `ndims` were the same or slightly different. As far as I understand this is the key piece of the PR so I want to be extra sure that we're covering every angle. Could you tell me if it makes sense to take a defensive approach and test the following circumstances?

- `jacobian.ndim > logp.ndim`: an example would be a `Log` transform on an `MvNormal`. It doesn't really make sense to me, but it _could_ happen under some circumstances that I don't really know at the moment. What would be the approach to take there? To sum the across the jacobian extra dimensions?
- `jacobian.ndim <= logp.ndim` and `jacobian.shape != logp.shape[:jacobian.ndim]`: This would mean that the "batch" dimensions of the jacobian don't match with the "batch" dimensions of the `logp`. To me this sounds like an error and it should at some point. The question is if it makes sense to add an assertion to the `logp` graph or not. It might be a very picky thing to do, and completely wasteful.
- `jacobian.ndim > logp.ndim` and `jacobian.shape[:logp.ndim] != logp.shape`: to me, this is just like the case listed above. Something was wrong in the `jacobian` or the `logp` and an error should be raised. The question is whether it makes sense to do so at each `logp` evaluation.

If you can argue why you don't need to test the above cases, then I think that this change is good enough

tests/distributions/test_transform.py
```diff
@@ -149,14 +149,14 @@ def test_simplex_accuracy():
 
 
 def test_sum_to_1():
-    check_vector_transform(tr.univariate_sum_to_1, Simplex(2))
-    check_vector_transform(tr.univariate_sum_to_1, Simplex(4))
+    check_vector_transform(tr.sum_to_1, Simplex(2))
+    check_vector_transform(tr.sum_to_1, Simplex(4))
 
-    with pytest.raises(ValueError, match=r"\(ndim_supp\) must not exceed 1"):
+    with pytest.warns(FutureWarning, match="ndim_supp argument is deprecated"):
```
lucianopaz, line 156: If you accept my above suggestion, you might want to change the message

tests/distributions/test_transform.py
```diff
@@ -270,36 +270,33 @@ def test_circular():
 
 
 def test_ordered():
-    check_vector_transform(tr.univariate_ordered, SortedVector(6))
+    check_vector_transform(tr.ordered, SortedVector(6))
 
-    with pytest.raises(ValueError, match=r"\(ndim_supp\) must not exceed 1"):
-        tr.Ordered(2)
+    with pytest.warns(FutureWarning, match="ndim_supp argument is deprecated"):
```
lucianopaz, line 276: If you accept my above suggestion, you might want to change the message

tests/distributions/test_transform.py
```diff
@@ -345,56 +345,13 @@ def check_transform_elementwise_logp(self, model):
             .sum()
```
lucianopaz: Are you sure that you want to `sum` here? You might want to actually compare shapes.

pymc/logprob/transforms.py
```diff
@@ -1229,6 +1232,10 @@ def transformed_logprob(op, values, *inputs, use_jacobian=True, **kwargs):
...
+                    logp = logp.sum(axis=np.arange(-diff_ndims, 0))
```
lucianopaz: This line only happens if `use_jacobian` is `True`. If it's false, but the transform is still applied, should you still `sum` across the extra dimensions in the `logp`? If you don't then you'll be returning an inconsistently shaped tensor depending on this flag.

tests/distributions/test_transform.py
lucianopaz: I think that you might be missing some edge cases if you `sum` the `jacob_det` and the other `logp`s. You might want to be pedantic with the test for the batch shapes and core shapes.

lucianopaz (commented):
Looks good @ricardoV94. I left one important comment that needs to be addressed before merging this. The rest are minor nitpicks

tests/distributions/test_transform.py
ricardoV94: The batch shapes would raise in the reference case with jacobian=True and on the call to `model.logp(sum=False)` above. In did that's what started to fail in #6897 where there was a simplex transform applied to univariate uniforms (and jacobian.shape=(2,) did not align with base_logp.shape=(3,)). The shape issue would always crop up by static shape inference or if the function was evaluated (which it was not before). 

The broadcasting issue (e.g, if you do a simplex on a uniform with shape=(2,)) is now captured as well by summing the jacobian and base_logp separately, and only then adding them. I think the purpose of this test helper is fine as it is.

pymc/distributions/transforms.py
ricardoV94: I think this was just how it was written originally, I'll change it

pymc/logprob/transforms.py
ricardoV94: Good points. 

> jacobian.ndim > logp.ndim: an example would be a Log transform on an MvNormal. It doesn't really make sense to me, but it could happen under some circumstances that I don't really know at the moment. What would be the approach to take there? To sum the across the jacobian extra dimensions?

That touches on https://github.com/pymc-devs/pymc/issues/6360#issuecomment-1621358657, and in generally the jacobian can't be simply summed if the core entries are somewhat "redundant". For instance, we shouldn't count the jacobian 3 times for a scale transformed dirichlet([1, 1, 1]), but we can do it for a multivariate normal. I will add an error for this.

> jacobian.ndim <= logp.ndim and jacobian.shape != logp.shape[:jacobian.ndim]: This would mean that the "batch" dimensions of the jacobian don't match with the "batch" dimensions of the logp. To me this sounds like an error and it should at some point. The question is if it makes sense to add an assertion to the logp graph or not. It might be a very picky thing to do, and completely wasteful.

The nice thing is that the static shape information is capturing more and more of these errors without any extra symbolic operations. Indeed this was what caused test failures when we bumped the PyTensor dependency in another PR and let me to realize our approach was completely wrong (even in the case it doesn't fail because it broadcasts). A correct jacobian determinant should never affect the batch dimensions, and shouldn't have "lingering" core dimensions. I think it's fine to assume they will be correctly implemented going into the future, and leave edge bugs to be captured by static shape info and making sure there is no sneaky broadcasting going on (which the changed test now does)

ricardoV94: Ah we can check the broadcasting flags of the jacobian (and summed logp) match. This will also clearly raise for invalid broadcasting (which should never happen between jacobian and logp). Since we don't allow "dynamic broadcasting" anymore, even if something ends up with a dim length of 1 at runtime, and we didn't know about it, it will still fail.

pymc/logprob/transforms.py
ricardoV94: Yeah, let's do it. The jacobian flag is a bit hackish, but I think it's better to return the same shapes.

ricardoV94:
In general, I really don't like these multivariate transforms being applied to univariate RVs. I understand they were a very convenient way of implementing fancy constraints in V3, but they pose a couple of issues:
1. We no longer know the true support dimension of variables when doing logp inference. This isn't too bad because the transforms come at the "end" of the graph, and derived variables can never depend on valued variables
2. Forward draws can't obviously respect these constraints, which is surprising for many users.

I think in the long term we would be better of restricting ourselves to transforms that match the "ndim_supp" of the base RV, and don't change the meaning of the variable (A simplex-transformed dirichlet is still the same dirichlet, but an ordered transform one is in general not). The ZeroSumNormal is a good example of a more consistent approach, and I think we can do similar stuff for these transforms with distribution factories (just like Censored and Truncated do).

pymc/distributions/transforms.py
ricardoV94: The value is already ignored

pymc/distributions/transforms.py
ricardoV94: Done

tests/distributions/test_transform.py
ricardoV94: Shapes must match for the `transform_logp_eval` to work in the first place. Notice we checked the ndim is correct already.

ricardoV94: I also added an explicit test for shapes

pymc/distributions/transforms.py
ricardoV94: Changed the message, WDYT?

lucianopaz (approved):
Looks good to me. Thanks @ricardoV94!

PR #6918, opened 2023-09-18
Introduce Valued Variables in logprob IR

### LINKED ISSUE #6917: Relationship between valued RVs lost during logp inference
### Description

Found out by @tvwenger in https://github.com/pymc-devs/pymc/discussions/6905#discussioncomment-7027204

```python
import pymc as pm
from pymc.logprob import conditional_logp

a_base = pm.Normal.dist()
a = a_base * 5
b = pm.Normal.dist(a * 8)

a_value = a.type()
b_value = b.type()
conditional_logp({a: a_value, b: b_value})  #  UserWarning: Random variables detected in the logp graph:
```
In this case the value variable of `a` is not replaced in the graph of `b`, because `b` is rewritten as `pm.Normal.dist(a_base * 40)` which doesn't have a value variable! 

The immediate issue here, is that we apply canonicalization rewrites that don't care about whether a variable is valued or not.
 
The bigger issue is that we don't have a means of allowing symbolic value transforms in our logprob inference. The correct thing would be to allow variables to depend on invertible transformations of valued variables. In this case `a_base * 40 = (a_value / 5) * 40`

ricardoV94:
This simplifies the rewrite logic quite a bit and fixes #6917 

----
:books: Documentation preview :books:: https://pymc--6918.org.readthedocs.build/en/6918/


ricardoV94:
Closing in favor of #7491

PR #6923, opened 2023-09-22
Fix rejection-based truncation of scalar variables

ricardoV94:
Reported in https://github.com/pymc-devs/pytensor/issues/442

----
:books: Documentation preview :books:: https://pymc--6923.org.readthedocs.build/en/6923/

### DIFF
--- a/pymc/distributions/truncated.py
+++ b/pymc/distributions/truncated.py
@@ -211,10 +211,14 @@ class Truncated(Distribution):
         # Fallback to rejection sampling
         def loop_fn(truncated_rv, reject_draws, lower, upper, rng, *rv_inputs):
             next_rng, new_truncated_rv = dist.owner.op.make_node(rng, *rv_inputs).outputs
-            truncated_rv = pt.set_subtensor(
-                truncated_rv[reject_draws],
-                new_truncated_rv[reject_draws],
-            )
+            # Avoid scalar boolean indexing
+            if truncated_rv.type.ndim == 0:
+                truncated_rv = new_truncated_rv
+            else:
+                truncated_rv = pt.set_subtensor(
+                    truncated_rv[reject_draws],
+                    new_truncated_rv[reject_draws],
+                )
             reject_draws = pt.or_((truncated_rv < lower), (truncated_rv > upper))
 
             return (
--- a/tests/distributions/test_truncated.py
+++ b/tests/distributions/test_truncated.py
@@ -101,11 +101,12 @@ def test_truncation_specialized_op(shape_info):
 
 @pytest.mark.parametrize("lower, upper", [(-1, np.inf), (-1, 1.5), (-np.inf, 1.5)])
 @pytest.mark.parametrize("op_type", ["icdf", "rejection"])
-def test_truncation_continuous_random(op_type, lower, upper):
+@pytest.mark.parametrize("scalar", [True, False])
+def test_truncation_continuous_random(op_type, lower, upper, scalar):
     loc = 0.15
     scale = 10
     normal_op = icdf_normal if op_type == "icdf" else rejection_normal
-    x = normal_op(loc, scale, name="x", size=100)
+    x = normal_op(loc, scale, name="x", size=() if scalar else (100,))
 
     xt = Truncated.dist(x, lower=lower, upper=upper)
     assert isinstance(xt.owner.op, TruncatedRV)
@@ -134,7 +135,7 @@ def test_truncation_continuous_random(op_type, lower, upper):
         assert np.unique(xt_draws).size == xt_draws.size
     else:
         with pytest.raises(TruncationError, match="^Truncation did not converge"):
-            draw(xt)
+            draw(xt, draws=100 if scalar else 1)
 
 
 @pytest.mark.parametrize("lower, upper", [(-1, np.inf), (-1, 1.5), (-np.inf, 1.5)])


pymc/distributions/distribution.py
```diff
@@ -208,6 +208,7 @@ def __init__(self, *args, ndim_supp, **kwargs):
         self.ndim_supp = ndim_supp
         kwargs.setdefault("inline", True)
         super().__init__(*args, **kwargs)
+        self.kwargs["ndim_supp"] = self.ndim_supp
```
lucianopaz: Is there a reason this has to happen after `super().__init__` and not before?

pymc/distributions/truncated.py
```diff
@@ -171,22 +174,36 @@ def rv_op(cls, dist, lower, upper, max_n_steps, size=None):
...
         dist = change_dist_size(dist, new_size=size)
+        dist_out_idx = dist.owner.outputs.index(dist)
+
+        rv_rng_inputs = [inp for inp in dist.owner.inputs if isinstance(inp.type, RandomType)]
+        rv_non_rng_inputs = [
+            inp for inp in dist.owner.inputs if not isinstance(inp.type, RandomType)
+        ]
+
+        if len(rv_rng_inputs) > 1:
+            raise NotImplementedError("Truncation of variable with multiple RNGs not supported")
```
lucianopaz: Why? Is it because you wouldn't know which of the multiple rng inputs needs to be truncated? So if you got a mixture distribution, you would not be able to distinguish between the mixing indexes and the mixture components?

pymc/distributions/truncated.py
```diff
@@ -195,11 +212,12 @@ def rv_op(cls, dist, lower, upper, max_n_steps, size=None):
                 cdf_lower_,
                 cdf_upper_,
                 rng=rng,
-                size=rv_inputs_[0],
+                size=rv_.shape,
             )
-            truncated_rv_ = icdf(rv_, uniform_)
+            truncated_rv_ = icdf(rv_, uniform_, warn_rvs=False)
```
lucianopaz: What does the `warn_rvs=False` do here?

pymc/distributions/truncated.py
```diff
@@ -209,17 +227,21 @@ def rv_op(cls, dist, lower, upper, max_n_steps, size=None):
...
+            new_truncated_rv = dist.owner.clone_with_new_inputs(rv_inputs).outputs[dist_out_idx]
+            # Avoid scalar boolean indexing
+            if truncated_rv.type.ndim == 0:
+                truncated_rv = new_truncated_rv
+            else:
+                truncated_rv = pt.set_subtensor(
+                    truncated_rv[reject_draws],
+                    new_truncated_rv[reject_draws],
+                )
             reject_draws = pt.or_((truncated_rv < lower), (truncated_rv > upper))
```
lucianopaz, line 222: Shouldn't this also have some kind of `pt.and_(not(reject_draws), ...` so that the draws that were already accepted don't get resampled? I fail to see where that is happening.

pymc/distributions/truncated.py
```diff
@@ -209,17 +227,21 @@ def rv_op(cls, dist, lower, upper, max_n_steps, size=None):
...
 
             return (
                 (truncated_rv, reject_draws),
-                [(rng, next_rng)],
+                collect_default_updates([new_truncated_rv]),
                 until(~pt.any(reject_draws)),
```
lucianopaz: Will this `until` cut the `scan` short of the `max_n_steps` if the condition is met sooner?

lucianopaz (commented):
This looks very nice @ricardoV94. I might need you to guide me a bit through the logic here because I haven't grasped the subtleties of trying to truncate `SymbolicDistribution`s yet.

pymc/distributions/distribution.py
ricardoV94: I don't remember

pymc/distributions/truncated.py
ricardoV94: It's much more mundane than that, multiple RNGs would complicate the logic. The biggest challenge of a Truncated Variable is that in the fallback random case I need to be able to do rejection sampling of a variable in a Scan, for that I need to be able to:
1. Create an equivalent RV inside the Scan
2. Provide the update expressions for any RNG in this RV

As you can see a big part of this PR is just figuring out which position does a RNG input have to go to recreate the base RV. By not supporting cases with multiple RNGs, I only have to worry about one number.

I can try to be less lazy in this PR

ricardoV94: More importantly, you made me realize Truncated still doesn't work for any SymbolicRandomVariables but only CustomDist ones. I reintroduced the check and added a message to guide users to try and wrap their variables inside a CustomDist.

This is a bit ugly, but the problem is that the graph of stuff like `pm.Mixture` is not all wrapped inside an OpFromGraph, so I don't know what the true "inputs" are to be able to recreate the variable inside the rejection sample scan... Gotta think about this one better.

pymc/distributions/truncated.py
ricardoV94: It emits a warning if there are random variables in the icdf graph. In this case there is one and it is supposed to be there (the uniform) since the icdf is part of the forward graph itself.

pymc/distributions/truncated.py
ricardoV94: The draws that were already accepted will always be within upper and lower. The set_subtensor only changes the indexes that were not already valid.

pymc/distributions/truncated.py
ricardoV94: Yes, that's the whole point of the `until` (Also you can never define one without `max_n_steps` if that helps)

ricardoV94:
@lucianopaz, I realized I needed a bigger refactor, mostly because https://github.com/pymc-devs/pytensor/issues/473 makes it hard to box other SymbolicRVs safely.

This PR is now just fixing the bug with the scalar case, and I'll open another one later with then new functionality

lucianopaz (approved):

PR #6933, opened 2023-09-27
Improve multinomial moment

### LINKED ISSUE #6233: Multinomial moment produces invalid values for some parameters
Reported in https://discourse.pymc.io/t/pm-multinomial-gives-sampling-error/10639

```python
import pymc as pm

with pm.Model() as m:
  p = [0.1441,0.1363,0.1385,0.1348,0.1521,0.1500,0.1442]
  x = pm.Multinomial("x", n=4, p=p)

m.initial_point()  # {'x': array([-2,  1,  1,  1,  1,  1,  1])}
```

The function that generates initial values for the Multinomial is here: https://github.com/pymc-devs/pymc/blob/b5b63f5d6c8d5891382cce1af9bd5ea22de795cf/pymc/distributions/multivariate.py#L560-L569

aerubanov:
Fix for bug in multinomial dist moment #6233. When we have some `diff` between sum(mode) and n we need to change diff largest element of mode by one instead of changing one element by diff.

----
:books: Documentation preview :books:: https://pymc--6933.org.readthedocs.build/en/6933/


pymc/distributions/multivariate.py
```diff
@@ -572,8 +572,15 @@ def moment(rv, size, n, p):
         n = pt.shape_padright(n)
         mode = pt.round(n * p)
         diff = n - pt.sum(mode, axis=-1, keepdims=True)
-        inc_bool_arr = pt.abs(diff) > 0
-        mode = pt.inc_subtensor(mode[inc_bool_arr.nonzero()], diff[inc_bool_arr.nonzero()])
+        if len(mode.shape.eval()) == 1:
```
ricardoV94: We shouldn't have evals in intermediate graphs. 

I think a proper solution would need an iterative algorithm with Scan and I am not sure that's worth the complexity for an edge case of a distribution that won't be properly sampled by our samplers anyway.

Users can always pass an initial value manually or set `initval="prior"`

ricardoV94: Does argsort work with axis?

pymc/distributions/multivariate.py
```diff
@@ -572,8 +572,15 @@ def moment(rv, size, n, p):
...
+            mode = mode.reshape((1, -1))
+        for i in range(mode.shape.eval()[0]):
+            elem_to_change = pt.abs(diff[i].astype(int))
```
ricardoV94: Instead of trying to distribute the difference among the first n largest elements, we could try to assign it all to the largest element of each mode.

That is still bound to fail now and then but probably less often than the even dumber old approach of distributing it to the first element of each mode.

That way the algorithm would still work in a vectorized manner.

Even your algorithm I think is not bullet proof, e.g. if the only solution is to move the largest entry in mode by more than one unit.

aerubanov: @ricardoV94 I like the idea  to assign all difference to the largest element of each mode, will implement it.

pymc/distributions/multivariate.py
```diff
@@ -572,15 +572,11 @@ def moment(rv, size, n, p):
...
-            mode = mode.reshape((1, -1))
-        for i in range(mode.shape.eval()[0]):
-            elem_to_change = pt.abs(diff[i].astype(int))
-            ind = pt.argsort(mode[i, :])[-elem_to_change:]
-            mode = pt.inc_subtensor(
-                mode[i, ind], diff[i] / pt.abs(diff[i]) * pt.ones_like(mode[i, ind])
-            )
-        mode = mode.squeeze()
+        inc_bool_arr = (pt.abs(diff) > 0).nonzero()
+        max_elem = pt.argmax(mode, axis=-1).reshape((-1, 1))
```
ricardoV94, line 576:
```suggestion
        max_elem = pt.argmax(mode, axis=-1, keepdims=True)
```

pymc/distributions/multivariate.py
```diff
@@ -572,15 +572,11 @@ def moment(rv, size, n, p):
...
+        for i in range(len(inc_bool_arr)):
+            pt.set_subtensor(inc_bool_arr[i][-1], max_elem[i][0])
```
ricardoV94: Can you give an intuition what this is doing?

tests/distributions/test_multivariate.py
```diff
@@ -938,38 +938,32 @@ class TestMoments:
...
+                np.array([[1, 0, 0, 0], [2, 3, 3, 2]]),
             ),
             (
                 np.array([[0.25, 0.25, 0.25, 0.25], [0.26, 0.26, 0.26, 0.22]]),
                 np.array([1, 10]),
                 (3, 2),
-                np.full((3, 2, 4), [[0, 0, 0, 1], [3, 3, 2, 2]]),
-            ),
-            (
-                np.array([0.1441, 0.1363, 0.1385, 0.1348, 0.1521, 0.1500, 0.1442]),
```
ricardoV94: You can keep this as a known xfail with pytest.params

pymc/distributions/multivariate.py
aerubanov: I added comment

pymc/distributions/multivariate.py
```diff
@@ -572,8 +572,16 @@ def moment(rv, size, n, p):
...
-        mode = pt.inc_subtensor(mode[inc_bool_arr.nonzero()], diff[inc_bool_arr.nonzero()])
+        inc_bool_arr = (pt.abs(diff) > 0).nonzero()
+        # inc_bool_arr is a tuple of mode elemeents that we need to change to get 0 diff.
+        # Each element is a vector of 1 or 2 elements. The first element (if present) is
+        # a row of elements that we want to change, and the second element is the
+        # corresponding column. Columm is always a zero, but we want to change it to
+        # maximum element in the row to awoid a risk of getting a negative value
+        max_elem = pt.argmax(mode, axis=-1).reshape((-1, 1))
+        for i in range(len(inc_bool_arr)):
+            pt.set_subtensor(inc_bool_arr[i][-1], max_elem[i][0])
```
ricardoV94: This doesn't do anything if you don't save the result of `set_subtensor`

tests/distributions/test_multivariate.py
```diff
@@ -965,6 +965,13 @@ class TestMoments:
                 (3, 2),
                 np.full((3, 2, 4), [[1, 0, 0, 0], [2, 3, 3, 2]]),
             ),
+            pytest.param(
+                np.array([0.1441, 0.1363, 0.1385, 0.1348, 0.1521, 0.1500, 0.1442]),
+                4,
+                None,
+                np.array([1, 1, 1, 1, 0, 0, 0]),
+                marks=pytest.mark.xfail(reason="First element is negative"),
```
```suggestion
                marks=pytest.mark.xfail(reason="Known failure in mode approximation "),
```

ricardoV94 (changes requested):
@aerubanov I found some issues with the current version. For example the following case fails:

```python
import pymc as pm

from pymc.distributions.distribution import moment

x = pm.Multinomial.dist(n=[5, 3], p=[0., 0.5, 0.5])
moment(x).eval()
```

However, I think I found a simpler solution. The code in mode would look something like:

```python
    def moment(rv, size, n, p):
        n = pt.shape_padright(n)
        mode = pt.round(n * p)
        # Add correction term between n and approximation on largest entry 
        # to reduce chances of negative values.
        # This can cause negative entries, but is computationally simple.
        diff = n - pt.sum(mode, axis=-1, keepdims=True)
        max_elem_idx = pt.argmax(mode, axis=-1, keepdims=True)
        mode = pt.inc_subtensor(
            pt.take_along_axis(mode, max_elem_idx, axis=-1),
            diff,
        )
        if not rv_size_is_none(size):
            ...
```

**Edit: Better solution below**

ricardoV94:
We could also add a PyTensor assert, checking known failure scenarios:
1. Any entry in the final mode is negative 
2. There is a non-zero entry for a zero-probability p.

This could tell the user that the moment function for Multinomial has known limitations, and the user should provide an initial value manually. An example of the second failure case:

```python
import pymc as pm
from pymc.distributions.distribution import moment

x = pm.Multinomial.dist(n=[1], p=[0., 0.25, 0.25, 0.25, 0.25])
moment(x).eval()  # [1, 0, 0, 0, 0]
```

ricardoV94:
>     2. There is a non-zero entry for a zero-probability p.

Hmm, if we do argmax on the mean and not mode, I think we can always avoid this failure case, so the only remaining one is when you need to subtract more entries than there are in the largest one:

```python
    def moment(rv, size, n, p):
        n = pt.shape_padright(n)
        mean = n * p
        mode = pt.round(mean)
        # Add correction term between n and approximation on highest expected entry
        # to reduce chances of negative values.
        # This can cause negative entries, but is computationally simple.
        diff = n - pt.sum(mode, axis=-1, keepdims=True)
        max_elem_idx = pt.argmax(mean, axis=-1, keepdims=True)
        mode = pt.inc_subtensor(
            pt.take_along_axis(mode, max_elem_idx, axis=-1),
            diff,
        )
        if not rv_size_is_none(size):
            output_size = pt.concatenate([size, [p.shape[-1]]])
            mode = pt.full(output_size, mode)
        return mode
```

aerubanov:
@ricardoV94 I added your implementation of moment, new test case and changed old to match new implementation

ricardoV94:
> @ricardoV94 I added your implementation of moment, new test case and changed old to match new implementation

Neat! Does it work for all cases? 

I think we should add a Pytensor assert in the final mode that raises if any entry is negative, with an informative message.

aerubanov:
> Neat! Does it work for all previous cases?
Yes it`s work, I will add assertion and fix test for `DirichletMultinomial` moment since it use multinomial moment too (with modification of `p` value).

pymc/distributions/multivariate.py
```diff
@@ -570,14 +570,26 @@ def dist(cls, n, p, *args, **kwargs):
...
+        )
         if not rv_size_is_none(size):
             output_size = pt.concatenate([size, [p.shape[-1]]])
             mode = pt.full(output_size, mode)
-        return mode
+        return Assert(
+            """
+        Negative value in moment found, it`s a result of
+        a known limitation of Multinomial distribution.
+        Please provide initial value manualy.
```
```suggestion
        Negative value in computed moment of Multinomial. 
        It is a known limitation that can arise when the expected largest count is small. 
        Please provide an initial value manually.
```

ricardoV94: This will have indentation in the error message, because of the triple quotes, no? If so, we should instead do:
```python
(
   "Error line 1. "
   "Error line 2. "
   "Error line 3."
 )
 ```

Comment (ricardoV94) [reply pymc/distributions/multivariate.py]: Just the nitpick above about not using triple quotes for the error message

File: pymc/distributions/multivariate.py [hunk 9]
```diff
@@ -584,11 +584,9 @@ def moment(rv, size, n, p):
...
             mode = pt.full(output_size, mode)
         return Assert(
-            """
-        Negative value in computed moment of Multinomial. 
-        It is a known limitation that can arise when the expected largest count is small. 
-        Please provide an initial value manually.
-        """
+            "Negative value in computed moment of Multinomial."
+            "It is a known limitation that can arise when the expected largest count is small."
+            "Please provide an initial value manually."
```
ricardoV94, line 559:
```suggestion
            "Negative value in computed moment of Multinomial. "
            "It is a known limitation that can arise when the expected largest count is small. "
            "Please provide an initial value manually."
```

ricardoV94 (approved):
Awesome, thanks!

aerubanov:
@ricardoV94 after rebase from main tests are passed, only test codecov is failed

ricardoV94:
> @ricardoV94 after rebase from main tests are passed, only codecov is failed

That test doesn't update properly so we can ignore

PR #6937, opened 2023-09-30
Add type hints to dist() for discrete distributions

johanbog:
**What is this PR about?**
Add type hints to the dist() class methods in the discrete distributions.
Issue https://github.com/pymc-devs/pymc/issues/5358

Discussion points:
- I am not sure if the type hints are correct for the `cutpoints` parameter in OrderedProbit and OrderedLogistic
- I believe that the type hints should also include list[Union[int, float]] (or list[int], in the cases where ints are expected), but that would require changing the implementation of typing for the continuous distributions as well, so I can do that in a separate PR if it should be done.

**Checklist**

## Major / Breaking Changes
N/A

## New features
N/A

## Bugfixes
N/A

## Documentation

N/A
## Maintenance
N/A

----
:books: Documentation preview :books:: https://pymc--6937.org.readthedocs.build/en/6937/


thomasaarholt:
I found this PR because I intended to do the same, starting with `pm.Normal`. 

My intention is to first add type hints on the `dist` methods, and then later (if deemed reasonable) copy the `dist` argument types over to a new `__init__` method, which itself just returns `None`. This should let us have the correct type signature (no more `*args`, a huge benefit) with hinting (a big benefit) with a small cost of some code duplication.

thomasaarholt:
Let me provide some evidence that the above method will work (here in VSCode). 
## Behaviour before code change
This is the current behaviour of the main branch:
image

## Code change
Here I make only the following code change to pymc's `pm.Normal` class:
image


## Result
On hover of the `Normal` class, we see a much nicer function definition:

Screenshot 2024-01-30 at 10 59 31

thomasaarholt:
I originally made some comments on this PR, which I then turned into a feature request that I am willing to work on, over at #7122.

pymc/distributions/discrete.py
```diff
@@ -878,7 +913,14 @@ class HyperGeometric(Discrete):
     rv_op = hypergeometric
 
     @classmethod
-    def dist(cls, N, k, n, *args, **kwargs):
+    def dist(
+        cls,
+        N: Optional[DISCRETE_DIST_PARAMETER_TYPES],
```
thomasaarholt: I don't think this and the next two variables should be Optional. They are required by the distribution.

thomasaarholt (approved):
Flyby review here, not a maintainer (though keen on pymc and type checking).
With the exception of one bit, this looks good to me.

thomasaarholt:
The mypy error is going to be a pain to fix. Here is a minimum repro of the issue.

```python
class Dist:
    def dist(self, *args: int):
        return 1

class Foo(Dist):
    def dist(self, foo: int, bar: int):
        return 1
```
```
run.py:6: error: Signature of "dist" incompatible with supertype "Dist"  [override]
run.py:6: note:      Superclass:
run.py:6: note:          def dist(self, *args: int) -> Any
run.py:6: note:      Subclass:
run.py:6: note:          def dist(self, foo: int, bar: int) -> Any
```

The issue is that in the Distribution class we have:
```python
    @classmethod
    def dist(
        cls,
        dist_params, # <-- main issue is here
        *,
        shape: Shape | None = None,
        **kwargs,
    ) -> TensorVariable:
```
For the Binomial class we have multiple parameters where the superclass (Distribution) has one, and that one isn't even an `*args`. But that wouldn't work anyway.
```
    @classmethod
    def dist(
        cls,
        n: DISCRETE_DIST_PARAMETER_TYPES,
        p: DIST_PARAMETER_TYPES | None = None,
        logit_p: DIST_PARAMETER_TYPES | None = None,
        *args,
        **kwargs,
    ):
```

I am tempted to suggest that this PR won't be merged and that we focus on replacing the class-based system with a function-based one as per #5308.

PR #6942, opened 2023-10-05
Make `Model.str_repr` robust to variables without monkey-patch

### LINKED ISSUE #6311: Do not expect model variables to always have `str_repr`

ricardoV94:
Closes #6311 which became even more annoying after the `do`/`observe`

----
:books: Documentation preview :books:: https://pymc--6942.org.readthedocs.build/en/6942/

### DIFF
--- a/pymc/printing.py
+++ b/pymc/printing.py
@@ -12,8 +12,6 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-import itertools
-
 from typing import Union
 
 from pytensor.compile import SharedVariable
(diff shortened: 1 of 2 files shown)


larryshamalama (approved):
Looks good! 👍

twiecki (approved):

PR #6943, opened 2023-10-06
Add constant and observed data to nutpie idata

Y0dler:
Temporary work-around for https://github.com/pymc-devs/nutpie/issues/74.  
With gracious assistance from Michael Osthege.

**What is this PR about?**
When sampling with the external NUTS sampler "nutpie", idata was not updated with neither attributes nor constant or observed data. Using pre-existing methods, this was fixed.

**Checklist**

## Bugfixes
- Observed and constant data are now included in the inference data object resulting from nutpie sampling.

----
:books: Documentation preview :books:: https://pymc--6943.org.readthedocs.build/en/6943/

### DIFF
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -39,7 +39,8 @@ from typing import (
 import numpy as np
 import pytensor.gradient as tg
 
-from arviz import InferenceData
+from arviz import InferenceData, dict_to_dataset
+from arviz.data.base import make_attrs
 from fastprogress.fastprogress import progress_bar
 from pytensor.graph.basic import Variable
 from typing_extensions import Protocol, TypeAlias
@@ -47,6 +48,11 @@ from typing_extensions import Protocol, TypeAlias
 import pymc as pm
 
 from pymc.backends import RunType, TraceOrBackend, init_traces
+from pymc.backends.arviz import (
+    coords_and_dims_for_inferencedata,
+    find_constants,
+    find_observations,
+)
 from pymc.backends.base import IBaseTrace, MultiTrace, _choose_chains
 from pymc.blocking import DictToArrayBijection
 from pymc.exceptions import SamplingError
@@ -293,8 +299,24 @@ def _sample_external_nuts(
                 "`idata_kwargs` are currently ignored by the nutpie sampler",
                 UserWarning,
             )
-
+        # gather observed and constant data as nutpie.sample() has no access to the PyMC model
+        coords, dims = coords_and_dims_for_inferencedata(model)
+        constant_data = dict_to_dataset(
+            find_constants(model),
+            library=pm,
+            coords=coords,
+            dims=dims,
+            default_dims=[],
+        )
+        observed_data = dict_to_dataset(
+            find_observations(model),
+            library=pm,
+            coords=coords,
+            dims=dims,
+            default_dims=[],
+        )
         compiled_model = nutpie.compile_pymc_model(model)
+        t_start = time.time()
         idata = nutpie.sample(
             compiled_model,
             draws=draws,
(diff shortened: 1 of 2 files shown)


ricardoV94:
@Y0dler this is probably better done on the nutpie side

michaelosthege:
> @Y0dler this is probably better done on the nutpie side

I disagree, because we'd introduce several dependencies onto internal PyMC functions to the nutpie codebase.
Additionally, we'd have to change the signature of the `nupie.sample` function to take a `pm.Model`, where right now it's totally PyMC-independent.

ricardoV94:
nutpie wants to return a full InferenceData regardless of pymc so this would be rendered useless pretty soon

michaelosthege:
> nutpie wants to return a full InferenceData regardless of pymc so this would be rendered useless pretty soon

At that point nutpie will probably require `constant_data` and `observed_data` as additional kwargs, right?
Then most of the code from this PR is still needed at the PyMC side - only the `idata.add_groups` and `attrs` would move to `nutpie`.

ricardoV94:
Calling `nutpie.sample` wants to return `constant_data`, that's all. Regardless of whether it was given a PyMC or Stan model I guess, although I'm not too worried about the Stan side. CC @aseyboldt

aseyboldt:
Thank you @Y0dler, I think this looks good.
I also think the proper place would be nutpie however. It has to access someone internal pymc structures anyway, because it needs to build the cfunc to call from rust, and needs to handle shared data etc.

The nutpie API itself also directly takes a model as input as it is:

```python
import nutpie
compiled = nutpie.compiled_pymc_model(model)
nutpie.sample(compiled)
```

And if used like this, it also should store the constant data (even if it was changed using the `with_data` method on `compiled`).

michaelosthege:
> And if used like this, it also should store the constant data (even if it was changed using the with_data method on compiled).

I don't understand what you're suggesting. I checked [`compile_pymc.py`](https://github.com/pymc-devs/nutpie/blob/2938d5a0f04a8797792d1f7746f8a24de250db82/python/nutpie/compile_pymc.py#L42) but neither `CompiledPyMCModel` nor `compile_pymc_model` make any attempt at taking care of `constant_data` or `observed_data`.
One could pass `compile_pymc_model(pmodel, constant_data=..., observed_data=...)` for the `nupie.sample` code to pass them to `arviz.from_dict(...)`, but this would break if someone uses the `.with_data()` method.

The solution proposed in **this PR fixes the bug for the vast majority of PyMC users**, without introducing complexity of coupling between the packages. And **it can easily be adapted/removed once the `nutpie` code has been refactored** to gather `constant_data` and `observed_data` from arbitrarily updated compiled models.

ricardoV94:
> One could pass compile_pymc_model(pmodel, constant_data=..., observed_data=...)

Why would one have to do that? All the information about constant/observed data is contained in the pymc model passed to `compile_pymc_model`. By that token one would have to call `compile_pymc_model(pmdoel, rvs=..., value_vars=..., transforms=...)` which is obviously not true.

michaelosthege:
Oh nutpie could totally do it! But the current control flow, and its internal design makes is very non-obvious how. At least I don't see how it could/should be done.

From the PyMC perspective this PR is a totally fine fix until it's taken care of on the nupie side. And to be honest, the time @Y0dler and me spent on this PR cost our employer about 400 bucks, so there won't be any more heavy lifting on this issue from our side.

ricardoV94:
> And to be honest, the time @Y0dler and me spent on this PR cost our employer about 400 bucks, so there won't be any more heavy lifting on this issue from our side.

That doesn't sound like a fair argument to influence design choices. The issue was moved to the nutpie repo and the PR was opened without any consultation. 

Anyway I am fine with the PR as a quick patch, with a comment to revert once the original issue is addressed.

pymc/sampling/mcmc.py
```diff
@@ -293,8 +299,24 @@ def _sample_external_nuts(
                 "`idata_kwargs` are currently ignored by the nutpie sampler",
                 UserWarning,
             )
-
+        # gather observed and constant data as nutpie.sample() has no access to the PyMC model
+        coords, dims = coords_and_dims_for_inferencedata(model)
```
ricardoV94, line 317: This can be done after sampling (if sampling fails, there's no point in doing this work in advance). 

Please add a comment along the lines of `# Temporary work-around. Revert once https://github.com/pymc-devs/nutpie/issues/74 is fixed`

michaelosthege: It's a very cheap function call: https://github.com/pymc-devs/pymc/blob/6f4a04017e2d32d64fcb85a2390965c9aa73f6a4/pymc/backends/arviz.py#L103-L112

And sometimes this step catches issues with coordinate value types, that's why I recommended to put it before sampling (just like gathering constant and observed data)

michaelosthege (approved):

ricardoV94 (approved):

PR #6961, opened 2023-10-18
 Use numpy testing utilities instead of custom close_to*

### LINKED ISSUE #6748: Use numpy testing utilities instead of custom `close_to*`
https://github.com/pymc-devs/pymc/blob/e1d36cacb2cfb946f56872ee03e8678dc6ccf7f4/tests/checks.py#L18-L27

Numpy testing routines already take care of asserting / printing failures nicely. We should just use them: https://numpy.org/doc/stable/reference/routines.testing.html

erik-werner:
Closes #6748 

**What is this PR about?**
This PR replaces our custom close_to* functions with numpy testing asserts with nice error displays and more careful array shape checking.

**Checklist**

## Major / Breaking Changes
- ...

## New features
- ...

## Bugfixes
- Fix test_simplex_accuracy() which was previously broken

## Documentation
- ...

## Maintenance
- Replace custom close_to functions with numpy equivalents

----
:books: Documentation preview :books:: https://pymc--6961.org.readthedocs.build/en/6961/


tests/distributions/test_transform.py
```diff
@@ -145,7 +145,7 @@ def test_simplex_accuracy():
     val = floatX(np.array([-30]))
     x = pt.vector("x")
     x.tag.test_value = val
-    identity_f = pytensor.function([x], tr.simplex.forward(x, tr.simplex.backward(x, x)))
+    identity_f = pytensor.function([x], tr.simplex.forward(tr.simplex.backward(x)))
```
erik-werner, line 148: I'm not very familiar with the transforms, so please check that I'm not doing anything stupid.

ricardoV94: Shouldn't have to change any code of that kind

erik-werner: As mentioned in the commit message, the previous test was broken without us noticing. The identity_f function returned an empty array, but our testing function missed this. Without this commit the test fails:
```
___________________________________________________________________________________________________________________________________ test_simplex_accuracy ___________________________________________________________________________________________________________________________________

    def test_simplex_accuracy():
        val = floatX(np.array([-30]))
        x = pt.vector("x")
        x.tag.test_value = val
        identity_f = pytensor.function([x], tr.simplex.forward(x, tr.simplex.backward(x, x)))
>       assert_allclose(val, identity_f(val), tol)

tests/distributions/test_transform.py:149: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (<function assert_allclose.<locals>.compare at 0x7fedd6730b80>, array([-30.]), array([], dtype=float64)), kwds = {'equal_nan': True, 'err_msg': '', 'header': 'Not equal to tolerance rtol=1e-07, atol=0', 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=0
E           
E           (shapes (1,), (0,) mismatch)
E            x: array([-30.])
E            y: array([], dtype=float64)
```

ricardoV94: Ugh, interesting. Will have to check

ricardoV94: Your change is correct, good catch!

erik-werner:
@ricardoV94  Any idea about what happened here? The measured code coverage has gone down by 20%.

ricardoV94:
The codecov is stupid, it doesn't rerun after the first time, and if that failed that's what stays: https://github.com/pymc-devs/pymc/issues/6660

No need to worry about it

tests/helpers.py
```diff
@@ -118,11 +118,11 @@ def setup_class(self):
     def teardown_class(self):
         shutil.rmtree(self.temp_dir)
 
-    def check_stat(self, check, idata, name):
+    def check_stat(self, check, idata):
         group = idata.posterior
         for var, stat, value, bound in check:
             s = stat(group[var].sel(chain=0), axis=0)
-            close_to(s, value, bound, name)
+            assert_array_less(np.abs(s.values - value), bound)
```
ricardoV94, line 125: Can we use `assert_allclose` here?

erik-werner: No, because ``bound`` is an array.

ricardoV94: `assert_allclose` works with arrays as well

erik-werner: It didn't work for me. And according to the numpy docs, ``rtol`` and ``atol`` have to be floats. https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html

ricardoV94: Ahh the tolerance is the array? Is that ever used, or are we passing scalars (perhaps replicated) all the time?

erik-werner: It's used. The test is used for multivariate arrays, and the tolerance is set to match the expected error from MC samples.

ricardoV94: My bad again then! Sorry for the trouble

ricardoV94 (approved):

ricardoV94:
Thanks a ton @erik-werner

tests/helpers.py
erik-werner: No trouble at all, thanks for checking!

PR #6962, opened 2023-10-19
Support logp derivation of `power(base, rv)`

### LINKED ISSUE #6896: Support logp derivation of power exponent
```
power(const, x), for any const > 0 and any x
power(const, x), for any const and discrete x  (we can play with `log(abs(neg_const))` and x's parity)
```
The first case we don't have to constrain ourselves to actual "constants", we can add a symbolic assert that `const > 0`.

The second requires us to implement transforms for discrete variables, which would probably need #6360 first, so we can focus on the first case, which is also probably more useful anyway.

We just have to make sure not to rewrite stuff like `power(x, const)` accidentally as those are implemented via our `PowerTransform`. This can be done by checking which of the inputs has a path to unvalued random variables.

_Originally posted by @ricardoV94 in https://github.com/pymc-devs/pymc/pull/6826#discussion_r1285793140_

LukeLB:
Addresses the first case in #6896, allowing users to evaluate the log prob of generalised bases e.g, power(base, x). This constricts base to be either a constant or a continuous random variable as long as its vaue is greater or equal to 0. 

I've added a node rewrite function to deal with the power(base, x) case. To make sure that power(x, exponent) is unaffected by this rewrite I have made changes to `find_measurable_transform`. This is based upon the ordering of the node inputs which is assumed to always be node.inputs = [base, exponent], where the new rewrite takes the case where index 1 is measurable and PowerTransform takes the case where index 0 is measurable.

Closes #6896 
...

**Checklist**

## New features
- Allows users to evaluate the logp of a generalised power(base, x) or base**x

----
:books: Documentation preview :books:: https://pymc--6962.org.readthedocs.build/en/6962/

### DIFF
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -617,6 +617,20 @@ def measurable_special_exp_to_exp(fgraph, node):
         return [1 / (1 + pt.exp(-inp))]
 
 
+@node_rewriter([pow])
+def measurable_power_expotent_to_exp(fgraph, node):
+    exponent, inp = node.inputs
+    try:
+        scalar_exponent = pt.get_underlying_scalar_constant_value(exponent).item()
+    except:
+        return None
+
+    if inp.type.dtype.startswith("int"):
+        return None
+    if scalar_exponent > 0:
+        return [pt.exp(pt.log(scalar_exponent) * inp)]
+
+
 @node_rewriter(
     [
         exp,
@@ -691,8 +705,7 @@ def find_measurable_transforms(fgraph: FunctionGraph, node: Node) -> Optional[Li
         if measurable_input_idx != 0:
             return None
         try:
-            (power,) = other_inputs
-            power = pt.get_underlying_scalar_constant_value(power).item()
+            power = pt.get_underlying_scalar_constant_value(node.inputs[1]).item()
         # Power needs to be a constant
         except NotScalarConstantError:
             return None
@@ -769,6 +782,12 @@ measurable_ir_rewrites_db.register(
     "transform",
 )
 
+measurable_ir_rewrites_db.register(
+    "measurable_power_expotent_to_exp",
+    measurable_power_expotent_to_exp,
+    "basic",
+    "transform",
+)
 
 measurable_ir_rewrites_db.register(
     "find_measurable_transforms",
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -127,7 +127,7 @@ from pymc.logprob.rewriting import (
     cleanup_ir_rewrites_db,
     measurable_ir_rewrites_db,
 )
-from pymc.logprob.utils import check_potential_measurability
+from pymc.logprob.utils import CheckParameterValue, check_potential_measurability
 
 
 class TransformedVariable(Op):
@@ -619,12 +619,14 @@ def measurable_special_exp_to_exp(fgraph, node):
 
 @node_rewriter([pow])
 def measurable_power_expotent_to_exp(fgraph, node):
-    exponent, inp = node.inputs
+    base, inp_exponent = node.inputs
+    base = CheckParameterValue("base > 0")(base, pt.all(pt.ge(base, 0.0)))
 
-    if inp.type.dtype.startswith("int"):
+    # check whether inp is discrete
+    if inp_exponent.type.dtype.startswith("int"):
         return None
-    # if scalar_exponent > 0:
-    return [pt.exp(pt.log(exponent) * inp)]
+
+    return [pt.exp(pt.log(base) * inp_exponent)]
 
 
 @node_rewriter(
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -618,7 +618,7 @@ def measurable_special_exp_to_exp(fgraph, node):
 
 
 @node_rewriter([pow])
-def measurable_power_expotent_to_exp(fgraph, node):
+def measurable_power_exponent_to_exp(fgraph, node):
     """Convert power(base, x) of `MeasurableVariable`s to exp form."""
     base, inp_exponent = node.inputs
 
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -619,7 +619,7 @@ def measurable_special_exp_to_exp(fgraph, node):
 
 @node_rewriter([pow])
 def measurable_power_exponent_to_exp(fgraph, node):
-    """Convert power(base, x) of `MeasurableVariable`s to exp form."""
+    """Convert power(base, rv) of `MeasurableVariable`s to exp(log(base) * rv) form."""
     base, inp_exponent = node.inputs
 
     # When the base is measurable we have `power(rv, exponent)`, which should be handled by `PowerTransform` and needs no further rewrite.
@@ -628,10 +628,6 @@ def measurable_power_exponent_to_exp(fgraph, node):
         return None
 
     base = CheckParameterValue("base >= 0")(base, pt.all(pt.ge(base, 0.0)))
-
-    # check whether inp_exponent is discrete
-    if inp_exponent.type.dtype.startswith("int"):
-        return None
 
     return [pt.exp(pt.log(base) * inp_exponent)]
 
@@ -710,8 +706,9 @@ def find_measurable_transforms(fgraph: FunctionGraph, node: Node) -> Optional[Li
         if measurable_input_idx != 0:
             return None
         try:
-            power = pt.get_underlying_scalar_constant_value(node.inputs[1]).item()
-        # Power needs to be a constant, if not then proceed to the other case power(const, x)
+            (power,) = other_inputs
+            power = pt.get_underlying_scalar_constant_value(power).item()
+        # Power needs to be a constant, if not then proceed to the other case power(base, rv)
         except NotScalarConstantError:
             return None
         transform_inputs = (measurable_input, power)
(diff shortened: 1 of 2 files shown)


pymc/logprob/transforms.py
```diff
@@ -617,6 +617,25 @@ def measurable_special_exp_to_exp(fgraph, node):
         return [1 / (1 + pt.exp(-inp))]
 
 
+@node_rewriter([pow])
+def measurable_power_expotent_to_exp(fgraph, node):
```
ricardoV94:
```suggestion
def measurable_power_exponent_to_exp(fgraph, node):
```

pymc/logprob/transforms.py
```diff
@@ -617,6 +617,25 @@ def measurable_special_exp_to_exp(fgraph, node):
...
+    """Convert power(base, x) of `MeasurableVariable`s to exp form."""
+    base, inp_exponent = node.inputs
+
+    # if base is measurable then it is power(x, exponent) and not power(base, x) as expected
+    # and should use PowerTransform
```
ricardoV94: Trying to clarify a bit more:
```suggestion
    # When the base is measurable we have `power(rv, exponent)`, which should be handled by `PowerTransform` and needs no further rewrite.
    # Here we change only the cases where exponent is measurable `power(base, rv)` which is not supported by the `PowerTransform`
```

pymc/logprob/transforms.py
```diff
@@ -617,6 +617,25 @@ def measurable_special_exp_to_exp(fgraph, node):
         return [1 / (1 + pt.exp(-inp))]
 
 
+@node_rewriter([pow])
+def measurable_power_expotent_to_exp(fgraph, node):
+    """Convert power(base, x) of `MeasurableVariable`s to exp form."""
```
ricardoV94: Maybe include how the exp form looks like just for clarity

pymc/logprob/transforms.py
```diff
@@ -617,6 +617,25 @@ def measurable_special_exp_to_exp(fgraph, node):
...
+    # if base is measurable then it is power(x, exponent) and not power(base, x) as expected
+    # and should use PowerTransform
+    if check_potential_measurability([base], fgraph.preserve_rv_mappings.rv_values.keys()):
+        return None
+
+    base = CheckParameterValue("base >= 0")(base, pt.all(pt.ge(base, 0.0)))
+
+    # check whether inp_exponent is discrete
+    if inp_exponent.type.dtype.startswith("int"):
+        return None
```
ricardoV94: Why can't it be discrete?

IIRC correctly the discrete case was even more flexible because we could potentially support negative values in the base (don't need to implement that now), but should work equally fine for the positive support

pymc/logprob/transforms.py
```diff
@@ -691,9 +710,8 @@ def find_measurable_transforms(fgraph: FunctionGraph, node: Node) -> Optional[Li
         if measurable_input_idx != 0:
             return None
         try:
-            (power,) = other_inputs
-            power = pt.get_underlying_scalar_constant_value(power).item()
-        # Power needs to be a constant
+            power = pt.get_underlying_scalar_constant_value(node.inputs[1]).item()
```
ricardoV94: I find it more readable in the old version, where we see we take it from the other inputs

tests/logprob/test_transforms.py
```diff
@@ -1159,6 +1160,73 @@ def test_special_log_exp_transforms(transform):
         assert equal_computations([logp_test], [logp_ref])
 
 
+def test_power_const_exponent_output():
```
ricardoV94:
```suggestion
def test_measurable_power_exponent_with_constant_base():
```

tests/logprob/test_transforms.py
```diff
@@ -1159,6 +1160,73 @@ def test_special_log_exp_transforms(transform):
...
+    x_vv_pow = x_rv_pow.clone()
+    x_vv_exp2 = x_rv_exp2.clone()
+
+    x_logp_fn_pow = pytensor.function([x_vv_pow], pt.sum(logp(x_rv_pow, x_vv_pow)))
+    x_logp_fn_exp2 = pytensor.function([x_vv_exp2], pt.sum(logp(x_rv_exp2, x_vv_exp2)))
+
+    np.testing.assert_allclose(x_logp_fn_pow(0.1), x_logp_fn_exp2(0.1))
+
+
+def test_power_const_exponent_negative_rv_error():
```
ricardoV94:
```suggestion
def def test_measurable_power_exponent_with_variable_base():
```

tests/logprob/test_transforms.py
```diff
@@ -1159,6 +1160,73 @@ def test_special_log_exp_transforms(transform):
...
+
+    with pytest.raises(ParameterValueError, match="base >= 0"):
+        logp_vals_fn(np.array([-2]), np.array([2]))
+
+
+def test_power_const_exponent_negative_const_error():
+    with pytest.raises(ParameterValueError, match="base >= 0"):
+        x_rv = pt.pow(-2, pt.random.normal())
+        x_vv = x_rv.clone()
+        logp(x_rv, x_vv)
```
ricardoV94: You can test this in the constant test, no need to rederive the logp in a separate test (should be a bit faster, and it's conceptually related anyway)

tests/logprob/test_transforms.py
```diff
@@ -1159,6 +1160,73 @@ def test_special_log_exp_transforms(transform):
...
+
+
+def test_power_const_exponent_rv():
```
ricardoV94: What does this test cover that is not covered before? I would perhaps just add a test that when both the base and power are not measurable, the logp inference fails?

ricardoV94 (changes requested):
This looks amazing!!!

Just some small suggestions for comments / tests, but nothing about the fundamentals

pymc/logprob/transforms.py
LukeLB: I've misread your orignal comment as saying no discrete variables at all, but I see now it's only for the second case. I'll remove.

tests/logprob/test_transforms.py
LukeLB: My thinking was to test that we get the expected behaviour from using `power(base, rv)`, against a ground truth. But maybe thats also covered by `test_power_const_exponent_output`. I'm happy to remove if you think it's redundant.

LukeLB: I'll add the test you suggested as well.

tests/logprob/test_transforms.py
LukeLB: Not sure I understand. I've changed the test to be,
```python
def test_measurable_power_exponent_with_constant_base():
    x_rv_pow = pt.pow(2, pt.random.normal())
    x_rv_exp2 = pt.exp2(pt.random.normal())

    x_vv_pow = x_rv_pow.clone()
    x_vv_exp2 = x_rv_exp2.clone()

    x_logp_fn_pow = pytensor.function([x_vv_pow], pt.sum(logp(x_rv_pow, x_vv_pow)))
    x_logp_fn_exp2 = pytensor.function([x_vv_exp2], pt.sum(logp(x_rv_exp2, x_vv_exp2)))

    np.testing.assert_allclose(x_logp_fn_pow(0.1), x_logp_fn_exp2(0.1))

    with pytest.raises(ParameterValueError, match="base >= 0"):
        x_rv_neg = pt.pow(-2, pt.random.normal())
        x_vv_neg = x_rv_neg.clone()
        logp(x_rv_neg, x_vv_neg)
```

But I don't think thats what you want, as I'm still rederiving the logp.

tests/logprob/test_transforms.py
```diff
@@ -1159,6 +1160,52 @@ def test_special_log_exp_transforms(transform):
...
+    x_rv = pt.power(base_rv, x_raw_rv)
+
+    x_rv.name = "x"
+    base_rv.name = "base"
+    base_vv = base_rv.clone()
+    x_vv = x_rv.clone()
+
+    res = conditional_logp({base_rv: base_vv, x_rv: x_vv})
+    factors = [factor for factor in res.values()]
+    logp_vals_fn = pytensor.function([base_vv, x_vv], factors[1])
```
ricardoV94:
```suggestion
    x_logp = res[x_vv]
    logp_vals_fn = pytensor.function([base_vv, x_vv], x_logp)
```

tests/logprob/test_transforms.py
```diff
@@ -1159,6 +1160,52 @@ def test_special_log_exp_transforms(transform):
...
+
+    with pytest.raises(ParameterValueError, match="base >= 0"):
+        logp_vals_fn(np.array([-2]), np.array([2]))
+
+
+def test_base_exponent_non_measurable():
```
ricardoV94, line 1202: This test doesn't make much sense. What I would imagine is a test like this:

```python
    base_rv = pt.random.normal([2])
    x_raw_rv = pt.random.normal()
    x_rv = pt.power(base_rv, x_raw_rv)
    x_rv.name = "x"
    
    x_vv = x.clone()
    with pytest.raises(...):
        conditional_logp({x_rv: x_vv})
 ``` 
Which will fail because there are two potential sources of measurability for `x_rv`

tests/logprob/test_transforms.py
ricardoV94: Hmm so the Assert is actually triggered before the `logp` is ever evaluated? Okay, I guess we can let it be.

tests/logprob/test_transforms.py
LukeLB: I see, so without conditioning on the base we get failure because of the dual measurability. Thanks for clearing that up.

ricardoV94 (approved):

ricardoV94:
Thanks @LukeLB, sounds line we nailed all these Elemwise transforms?

LukeLB:
Cheers! Yep all done💥

PR #6965, opened 2023-10-20
Allow Stationary covariance functions to use user defined distance functions

bwengals:
This PR adds an optional argument to any covariance function that inherits from `Stationary` called `square_dist`.  New feature and existing model code shouldn't be affected.  

**Checklist**

## Major / Breaking Changes
- None

## New features
- Allow user to pass different distance functions

## Bugfixes
- None

## Documentation
- None

## Maintenance
- None

----
:books: Documentation preview :books:: https://pymc--6965.org.readthedocs.build/en/6965/

### DIFF
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -510,8 +510,8 @@ class Stationary(Covariance):
     ls: Lengthscale.  If input_dim > 1, a list or array of scalars or PyMC random
         variables.  If input_dim == 1, a scalar or PyMC random variable.
     ls_inv: Inverse lengthscale.  1 / ls.  One of ls or ls_inv must be provided.
-    square_dist: An optional (squared) distance function.  If None is supplied, the 
-        default is the square of the Euclidean distance.  The signature of this 
+    square_dist: An optional (squared) distance function.  If None is supplied, the
+        default is the square of the Euclidean distance.  The signature of this
         function is `square_dist(X: TensorLike, Xs: Tensorlike, ls: Tensorlike)`.
     """
 
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -533,8 +533,13 @@ class Stationary(Covariance):
                 ls = 1.0 / ls_inv
         self.ls = pt.as_tensor_variable(ls)
 
+        if square_dist is None:
+            self.square_dist = self.default_square_dist
+        else:
+            self.square_dist = square_dist
+
     @staticmethod
-    def square_dist(X, Xs, ls):
+    def default_square_dist(X, Xs, ls):
         X = pt.mul(X, 1.0 / ls)
         X2 = pt.sum(pt.square(X), 1)
         Xs = pt.mul(Xs, 1.0 / ls)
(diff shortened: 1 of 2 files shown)


tests/gp/test_cov.py
```diff
@@ -408,10 +408,36 @@ def test_stable(self):
...
+            )
+            return pt.clip(sqd, 0.0, np.inf)
+
+        X = np.linspace(-5, 5, 100)[:, None]
+        with pm.Model() as model:
+            cov1 = pm.gp.cov.Matern32(1, ls=1)
+            cov2 = pm.gp.cov.Matern32(1, ls=1, square_dist=square_dist)
+        K1 = cov1(X).eval()
+        K2 = cov2(X, X).eval()
+        npt.assert_allclose(K1, K2, atol=1e-5)
```
ricardoV94, line 438: maybe also test you can get something different from the default?

bwengals: good call

pymc/gp/cov.py
```diff
@@ -518,6 +521,7 @@ def __init__(
         ls=None,
         ls_inv=None,
         active_dims: Optional[IntSequence] = None,
+        square_dist: Optional[Callable] = None,
```
lucianopaz, line 524: I can't see where the `square_dist` argument is being used later. Do you want to attach it to the covariance instance at the end of `__init__`? As it stands now, it looks like this isn't doing anything. Furthermore, even if you attach it to the object, would that be called instead of the `__class__` `square_dist` accessor?

pymc/gp/cov.py
```diff
@@ -529,23 +533,19 @@ def __init__(
...
-            sqd = -2.0 * pt.dot(X, pt.transpose(X)) + (
-                pt.reshape(X2, (-1, 1)) + pt.reshape(X2, (1, -1))
-            )
-        else:
-            Xs = pt.mul(Xs, 1.0 / self.ls)
-            Xs2 = pt.sum(pt.square(Xs), 1)
-            sqd = -2.0 * pt.dot(X, pt.transpose(Xs)) + (
-                pt.reshape(X2, (-1, 1)) + pt.reshape(Xs2, (1, -1))
-            )
+        Xs = pt.mul(Xs, 1.0 / ls)
```
lucianopaz, line 545: Why did you take out the `is None` case? Can't you just do `Xs = X if Xs is None else Xs` and have a single branch?

tests/gp/test_cov.py
lucianopaz: I agree with this

lucianopaz (changes requested):
Thanks @bwengals. I'm not sure if the PR, as it stands, actually customizes the `square_dist` callable? I might have missed something, but I didn't see where it got attached to the object.

There is one theoretical thing that we need to be aware of and we could add into the docstring of `square_dist`: **not all distance metrics yield valid covariance kernels**. There's a short summary of this [here](https://arxiv.org/pdf/2202.10613.pdf) in section 3.1.3. I fell into this trap once when I tried to implement a covariance kernel using the Haversine distance on the surface of a sphere, until I later found out that was wrong.

bwengals:
> I'm not sure if the PR, as it stands, actually customizes the square_dist callable?

No.. Sorry for the sloppiness on my end!  The test you and @ricardoV94 mentioned would have caught this issue.

pymc/gp/cov.py
bwengals: I moved it down to line 564 (and your one-liner is nicer).   I moved it because I thought it'd be easier for someone making a different distance function if the inputs didn't have defaults.

PR #6978, opened 2023-11-01
Recognize alternative form of sigmoid in logprob inference

ricardoV94:
----
:books: Documentation preview :books:: https://pymc--6978.org.readthedocs.build/en/6978/

### DIFF
--- a/tests/logprob/test_transforms.py
+++ b/tests/logprob/test_transforms.py
@@ -1127,31 +1127,31 @@ def test_cosh_rv_transform():
     )
 
 
-TRANSFORMATIONS = {
-    "log1p": (pt.log1p, lambda x: pt.log(1 + x)),
-    "softplus": (pt.softplus, lambda x: pt.log(1 + pt.exp(x))),
-    "log1mexp": (pt.log1mexp, lambda x: pt.log(1 - pt.exp(x))),
-    "log2": (pt.log2, lambda x: pt.log(x) / pt.log(2)),
-    "log10": (pt.log10, lambda x: pt.log(x) / pt.log(10)),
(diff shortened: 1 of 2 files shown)


tests/logprob/test_transforms.py
```diff
@@ -1127,31 +1127,34 @@ def test_cosh_rv_transform():
...
-    "log10": (pt.log10, lambda x: pt.log(x) / pt.log(10)),
-    "exp2": (pt.exp2, lambda x: pt.exp(pt.log(2) * x)),
-    "expm1": (pt.expm1, lambda x: pt.exp(x) - 1),
-    "sigmoid": (pt.sigmoid, lambda x: 1 / (1 + pt.exp(-x))),
-}
-
-
-@pytest.mark.parametrize("transform", TRANSFORMATIONS.keys())
-def test_special_log_exp_transforms(transform):
+TRANSFORMATIONS = {}
```
LukeLB: I don't think this dictionary is needed anymore

tests/logprob/test_transforms.py
```diff
@@ -1127,31 +1127,34 @@ def test_cosh_rv_transform():
...
+
+
+@pytest.mark.parametrize(
```
LukeLB, line 1130: The parameritisation is much cleaner I don't know why I didn't do it before

LukeLB (changes requested):
All looks good to me! Just need to remove that empty dictionary

tests/logprob/test_transforms.py
ricardoV94: Thanks for the catch

PR #6982, opened 2023-11-01
Fix compute_test_value error when creating observed variables

### LINKED ISSUE #6981: BUG: `TypeError: Variables do not support boolean operations` with `compute_test_value = "raise"` for observed random variable
### Describe the issue:

Hi!

When running the tests for dfm/celerite2, with `pytensor.config.compute_test_value = "raise"`, for PyMC v5 and v4, I encountered the error below. I'm adding a simple example to reproduce.

The error seems to come from the use of `np.broadcast_to()` on `rv_var.shape` [here](https://github.com/pymc-devs/pymc/blob/58217cc821209d6a541ae6cfae9541c776d6b6cd/pymc/model/core.py#L1355).
I believe replacing it `rv_var.type.shape` fixes the issue.

Should I submit a PR with the change? If I write a test with the example below, where is the best place to put it?

Thank you!

```python
import numpy as np
import pytensor
import pytensor.tensor as pt

import pymc as pm

pytensor.config.compute_test_value = "raise"

data = rng.standard_normal(100)
with pm.Model():
    # NOTE: pt -> np yields same error
    obs = pm.Normal("obs", mu=pt.zeros_like(data), sigma=pt.ones_like(data), observed=data)
```

### Error message:

<details>
```python
  File "/home/vandal/repos/astro/pymc/repro_error.py", line 13, in <module>
    obs = pm.Normal("obs", mu=pt.zeros_like(data), sigma=pt.ones_like(data), observed=data)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vandal/repos/astro/pymc/pymc/distributions/distribution.py", line 316, in __new__
    rv_out = model.register_rv(
             ^^^^^^^^^^^^^^^^^^
  File "/home/vandal/repos/astro/pymc/pymc/model/core.py", line 1315, in register_rv
    rv_var = self.make_obs_var(rv_var, observed, dims, transform, total_size)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vandal/repos/astro/pymc/pymc/model/core.py", line 1356, in make_obs_var
    rv_var.tag.test_value = np.broadcast_to(test_value, rv_var.shape)
  File "/home/vandal/repos/astro/exoplanet-dev/celerite2-pymc5/venv/lib/python3.11/site-packages/numpy/lib/stride_tricks.py", line 413, in broadcast_to
    return _broadcast_to(array, shape, subok=subok, readonly=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vandal/repos/astro/exoplanet-dev/celerite2-pymc5/venv/lib/python3.11/site-packages/numpy/lib/stride_tricks.py", line 345, in _broadcast_to
    if any(size < 0 for size in shape):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vandal/repos/astro/exoplanet-dev/celerite2-pymc5/venv/lib/python3.11/site-packages/pytensor/tensor/variable.py", line 72, in __bool__
    raise TypeError("Variables do not support boolean operations.")
TypeError: Variables do not support boolean operations.
```
</details>

### PyMC version information:

<details>
Output of `pymc.__version__`: 5.9.1+1.g58217cc8.dirty (I played with the code to debug the issue, but it's basically the main branch)
</details>

### Context for the issue:

As mentioned above, the error affects dfm/celerite2 when running the test suite. Here is the output of that:

<details>
```python
python/test/pymc/test_pymc_celerite2.py:115: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
python/celerite2/pymc/celerite2.py:96: in marginal
    return CeleriteNormal(
../../pymc/pymc/distributions/distribution.py:316: in __new__
    rv_out = model.register_rv(
../../pymc/pymc/model/core.py:1315: in register_rv
    rv_var = self.make_obs_var(rv_var, observed, dims, transform, total_size)
../../pymc/pymc/model/core.py:1356: in make_obs_var
    rv_var.tag.test_value = np.broadcast_to(test_value, rv_var.shape)
venv/lib/python3.11/site-packages/numpy/lib/stride_tricks.py:413: in broadcast_to
    return _broadcast_to(array, shape, subok=subok, readonly=True)
venv/lib/python3.11/site-packages/numpy/lib/stride_tricks.py:345: in _broadcast_to
    if any(size < 0 for size in shape):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Lt.0

    def __bool__(self):
        # This is meant to prohibit stuff like a < b < c, which is internally
        # implemented as (a < b) and (b < c). The trouble with this is the
        # side-effect that checking for a non-NULL a by typing "if a: ..."
        # uses the same __nonzero__ method.  We want these both to work, but
        # it seems impossible.  Currently, all vars evaluate to nonzero except
        # the return values of comparison operators, which raise this
        # exception.  If you can think of a better solution, go for it!
        #
        # __bool__ is Python 3.x data model. __nonzero__ is Python 2.x.
        if self._is_nonzero:
            return True
        else:
>           raise TypeError("Variables do not support boolean operations.")
E           TypeError: Variables do not support boolean operations.

venv/lib/python3.11/site-packages/pytensor/tensor/variable.py:72: TypeError
```
</details>

vandalt:
**What is this PR about?**
This fixes #6981. The issue was about observed variables causing a `TypeError` when a model is created with `pytensor.config.compute_test_value != "off"` It adds a small test that reproduces the issue and replaces `rv_var.shape` by `rv_var.type.shape` in the `np.broadcast_to()` call.

Let me know if I should move the test elsewhere or if something else is needed!

Thanks!

**Checklist**

## Bugfixes
- Fixes #6981.

----
:books: Documentation preview :books:: https://pymc--6982.org.readthedocs.build/en/6982/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1352,7 +1352,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
             if test_value is not None:
                 # We try to reuse the old test value
-                rv_var.tag.test_value = np.broadcast_to(test_value, rv_var.shape)
+                rv_var.tag.test_value = np.broadcast_to(test_value, rv_var.type.shape)
             else:
                 rv_var.tag.test_value = data
 
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1350,10 +1350,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         if pytensor.config.compute_test_value != "off":
             test_value = getattr(rv_var.tag, "test_value", None)
 
-            if test_value is not None:
-                # We try to reuse the old test value
-                rv_var.tag.test_value = np.broadcast_to(test_value, rv_var.type.shape)
-            else:
+            if test_value is None:
                 rv_var.tag.test_value = data
 
         mask = getattr(data, "mask", None)
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1347,12 +1347,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
                 "Dimensionality of data and RV don't match.", actual=data.ndim, expected=rv_var.ndim
             )
 
-        if pytensor.config.compute_test_value != "off":
-            test_value = getattr(rv_var.tag, "test_value", None)
-
-            if test_value is None:
-                rv_var.tag.test_value = data
-
         mask = getattr(data, "mask", None)
         if mask is not None:
             impute_message = (
--- a/tests/model/test_core.py
+++ b/tests/model/test_core.py
@@ -223,16 +223,13 @@ class TestObserved:
         assert x1.type.dtype == X.type.dtype
         assert x2.type.dtype == X.type.dtype
 
+    @pytensor.config.change_flags(compute_test_value="raise")
     def test_observed_compute_test_value(self):
-        with pytensor.config.change_flags(compute_test_value="raise"):
-            rng = np.random.default_rng(123)
-            data = rng.standard_normal(100)
-            with pm.Model():
-                obs = pm.Normal(
-                    "obs", mu=pt.zeros_like(data), sigma=pt.ones_like(data), observed=data
-                )
-                assert obs.tag.test_value.shape == data.shape
-                assert obs.tag.test_value.dtype == data.dtype
+        data = np.zeros(100)
+        with pm.Model():
+            obs = pm.Normal("obs", mu=pt.zeros_like(data), sigma=1, observed=data)
+        assert obs.tag.test_value.shape == data.shape
+        assert obs.tag.test_value.dtype == data.dtype
 
 
 def test_duplicate_vars():
--- a/tests/model/test_core.py
+++ b/tests/model/test_core.py
@@ -223,6 +223,20 @@ class TestObserved:
         assert x1.type.dtype == X.type.dtype
         assert x2.type.dtype == X.type.dtype
 
+    def test_observed_compute_test_value(self):
+        try:
+            with pytensor.config.change_flags(compute_test_value="raise"):
+                rng = np.random.default_rng(123)
+                data = rng.standard_normal(100)
+                with pm.Model():
+                    obs = pm.Normal(
+                        "obs", mu=pt.zeros_like(data), sigma=pt.ones_like(data), observed=data
+                    )
+        except TypeError:
+            pytest.fail(
+                "TypeError raised when trying to add an observed RV with `compute_test_value='false'`"
+            )
+
 
 def test_duplicate_vars():
     with pytest.raises(ValueError) as err:
--- a/tests/model/test_core.py
+++ b/tests/model/test_core.py
@@ -228,7 +228,11 @@ class TestObserved:
             rng = np.random.default_rng(123)
             data = rng.standard_normal(100)
             with pm.Model():
-                pm.Normal("obs", mu=pt.zeros_like(data), sigma=pt.ones_like(data), observed=data)
+                obs = pm.Normal(
+                    "obs", mu=pt.zeros_like(data), sigma=pt.ones_like(data), observed=data
+                )
+                assert obs.tag.test_value.shape == data.shape
+                assert obs.tag.test_value.dtype == data.dtype
 
 
 def test_duplicate_vars():
--- a/tests/model/test_core.py
+++ b/tests/model/test_core.py
@@ -224,18 +224,11 @@ class TestObserved:
         assert x2.type.dtype == X.type.dtype
 
     def test_observed_compute_test_value(self):
-        try:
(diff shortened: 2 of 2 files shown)


pymc/model/core.py
```diff
@@ -1352,7 +1352,7 @@ def make_obs_var(
 
             if test_value is not None:
                 # We try to reuse the old test value
-                rv_var.tag.test_value = np.broadcast_to(test_value, rv_var.shape)
+                rv_var.tag.test_value = np.broadcast_to(test_value, rv_var.type.shape)
             else:
                 rv_var.tag.test_value = data
```
ricardoV94: There could be None in `type.shape`. Broadcasting shouldn't be needed because the RV should always have the right shape.

I suggest simplifying the logic to:
```
if test_value is None:
    rv_var.tag.test_value = data
```
And ignore the whole broadcast logic

tests/model/test_core.py
```diff
@@ -223,6 +223,20 @@ def test_observed_type(self):
         assert x1.type.dtype == X.type.dtype
         assert x2.type.dtype == X.type.dtype
 
+    def test_observed_compute_test_value(self):
+        try:
```
ricardoV94: No need for the try except

pymc/model/core.py
ricardoV94: I can't think of a scenario where this could be None though because the RV is created by the time this method is called. Can we just remove this section al together?

tests/model/test_core.py
```diff
@@ -223,6 +223,20 @@ def test_observed_type(self):
...
+            with pytensor.config.change_flags(compute_test_value="raise"):
+                rng = np.random.default_rng(123)
+                data = rng.standard_normal(100)
+                with pm.Model():
+                    obs = pm.Normal(
```
ricardoV94: should test something about the tag.test_value of obs

ricardoV94 (commented):
Thanks for opening a PR.

This is related to some functionality that is not much needed in PyMC anymore and we may remove this PyTensor flag in the near future. Anyway, I've left some comments below

pymc/model/core.py
vandalt: I implemented the first suggestion.

I can't see a case where `test_value` is None either, but you're much more familiar with this code than I am so I'll implement the option you prefer! :) Should I just remove the `if pytensor.config.compute_test_value != "off":` block completely?

ricardoV94: Yes, let's do that and see if any tests fails

tests/model/test_core.py
vandalt: I added a test that compares `shape` and `dtype` with `data`. Let me know if you had something else in mind!

pymc/model/core.py
vandalt: Done! I ran tests in `test_core.py` locally, but not the whole suite as I'm guessing it will run here.

ricardoV94 (approved):

ricardoV94:
I pushed a commit that simplifies the test, after confirming it still failed in the main branch.

Thanks for the PR. This is the best kind of fix :D

PR #6988, opened 2023-11-03
Refactor `get_tau_sigma` and support lists of variables

### LINKED ISSUE #6987: BUG: `get_tau_sigma` does not support list of variables
### Describe the issue:

The function `get_tau_sigma` in `distributions/continuous.py` does not properly handle lists of `Variable`s. When a list is passed, the function attempts to convert it to a `numpy` array before checking for positivity (N.B. the positivity check does not occur if a `Variable` is passed).

```python
import pymc as pm
print(pm.__version__)
import pytensor.tensor as pt
import numpy as np

data = np.random.normal(size=100)

with pm.Model() as model:
    mu_1 = pm.Normal('mu_1', mu=0.0, sigma=1.0)
    mu_2 = pm.Normal('mu_2', mu=0.0, sigma=1.0)
    mu = [mu_1, mu_2]
    
    sigma_1 = pm.HalfNormal('sigma_1', sigma=1.0)
    sigma_2 = pm.HalfNormal('sigma_2', sigma=1.0)
    sigma = [sigma_1, sigma_2]
    # the following works:
    # sigma = pt.as_tensor_variable([sigma_1, sigma_2])
    
    w = pm.Dirichlet('w', a=np.ones(2))
    _ = pm.NormalMixture('data', w=w, mu=mu, sigma=sigma, observed=data)
```

### Error message:

```shell
5.9.1+3.gec4407d3
Cell In[6], line 18
     15 sigma = [sigma_1, sigma_2]
     17 w = pm.Dirichlet('w', a=np.ones(2))
---> 18 _ = pm.NormalMixture('data', w=w, mu=mu, sigma=sigma, observed=data)

File ~/science/python/pymc/pymc/distributions/mixture.py:552, in NormalMixture.__new__(cls, name, w, mu, sigma, tau, comp_shape, **kwargs)
    551 def __new__(cls, name, w, mu, sigma=None, tau=None, comp_shape=(), **kwargs):
--> 552     _, sigma = get_tau_sigma(tau=tau, sigma=sigma)
    554     return Mixture(name, w, Normal.dist(mu, sigma=sigma, size=comp_shape), **kwargs)

File ~/science/python/pymc/pymc/distributions/continuous.py:242, in get_tau_sigma(tau, sigma)
    240 else:
    241     sigma_ = np.asarray(sigma)
--> 242     if np.any(sigma_ <= 0):
    243         raise ValueError("sigma must be positive")
    244     tau = sigma_**-2.0

File ~/miniconda3/envs/pymc/lib/python3.11/site-packages/pytensor/tensor/variable.py:72, in _tensor_py_operators.__bool__(self)
     70     return True
     71 else:
---> 72     raise TypeError("Variables do not support boolean operations.")

TypeError: Variables do not support boolean operations.
```

### PyMC version information:

5.9.1+3.gec4407d3

### Context for the issue:

Since the distributions that use `get_tau_sigma` (e.g., `NormalMixture`) state that `sigma` and `tau` should be `tensor_like`, and since `tensor_like` includes lists and other iterables, this is a bug.

tvwenger:
**What is this PR about?**
This PR refactors the `get_tau_sigma` logic in order to support lists of `Variable`s and fixes https://github.com/pymc-devs/pymc/issues/6987.

N.B. The original code had a positivity check for non-`Variable` arguments, but not for `Variable` arguments. Since this PR casts the arguments to `Variable`, I've removed the positivity check for non-`Variable` arguments.

**Checklist**

## Major / Breaking Changes
- `get_tau_sigma` now always returns a `Variable`, even when scalars are passed

## New features
- N/A

## Bugfixes
- https://github.com/pymc-devs/pymc/issues/6987

## Documentation
- N/A

## Maintenance
- Updated tests

----
:books: Documentation preview :books:: https://pymc--6988.org.readthedocs.build/en/6988/

### DIFF
--- a/pymc/distributions/shape_utils.py
+++ b/pymc/distributions/shape_utils.py
@@ -109,6 +109,7 @@ def broadcast_dist_samples_shape(shapes, size=None):
     Examples
     --------
     .. code-block:: python
+
         size = 100
         shape0 = (size,)
         shape1 = (size, 5)
@@ -116,7 +117,9 @@ def broadcast_dist_samples_shape(shapes, size=None):
         out = broadcast_dist_samples_shape([shape0, shape1, shape2],
                                            size=size)
         assert out == (size, 4, 5)
+
     .. code-block:: python
+
         size = 100
         shape0 = (size,)
         shape1 = (5,)
@@ -124,7 +127,9 @@ def broadcast_dist_samples_shape(shapes, size=None):
         out = broadcast_dist_samples_shape([shape0, shape1, shape2],
                                            size=size)
         assert out == (size, 4, 5)
+
     .. code-block:: python
+
         size = 100
         shape0 = (1,)
         shape1 = (5,)
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -808,6 +808,7 @@ def collect_default_updates(
     Examples
     --------
     .. code:: python
+
         import pymc as pm
         from pytensor.scan import scan
         from pymc.pytensorf import collect_default_updates


ricardoV94:
@tvwenger would you like to update this PR? Would only need to add some regression tests for the original issue

tvwenger:
@ricardoV94 Thanks for the reminder and sorry for the delay. I added some regression tests, so this should be good to go once the checks are finished.

tvwenger:
@ricardoV94 This PR is ready for review

pymc/distributions/continuous.py
```diff
@@ -229,32 +229,18 @@ def get_tau_sigma(tau=None, sigma=None):
...
-                tau = sigma_**-2.0
+    if tau is not None and sigma is not None:
+        raise ValueError("Can't pass both tau and sigma")
 
+    if tau is None and sigma is None:
+        sigma = pt.as_tensor_variable(1.0)
+        tau = pt.as_tensor_variable(1.0)
+    elif tau is None:
+        sigma = pt.as_tensor_variable(sigma)
+        tau = (sigma**-2.0) * pt.sign(sigma)
```
ricardoV94: I would keep the old comment

pymc/distributions/continuous.py
```diff
@@ -229,32 +229,18 @@ def get_tau_sigma(tau=None, sigma=None):
...
-            if isinstance(tau, Variable):
-                # Keep sigma negative, if tau was negative, so that it will fail when used
-                sigma = pt.abs(tau) ** (-0.5) * pt.sign(tau)
-            else:
-                tau_ = np.asarray(tau)
-                if np.any(tau_ <= 0):
-                    raise ValueError("tau must be positive")
-                sigma = tau_**-0.5
+        tau = pt.as_tensor_variable(tau)
+        sigma = pt.abs(tau) ** (-0.5) * pt.sign(tau)
```
ricardoV94: This actually doesn't need the sign trick, because it will `nan` if tau is negative which is fine enough

ricardoV94 (approved):
This looks good, just two nitpicks, one of which is not related to your changes. Let me know if you want me to merge or if you want to address them.

pymc/distributions/continuous.py
tvwenger: If we're going to keep the behavior that the returned tau is negative if the passed sigma is negative, then I think we should ensure that the same is true when the user passes negative tau: we should return a negative sigma and it will fail when used.

tvwenger:
@ricardoV94 Ready for review when checks are done

pymc/distributions/continuous.py
ricardoV94: The reason for the abs was just not to mask invalid parameters. The -2 power would make a negative float positive without any complaints. 

That is not a concern with the -0.5 power, which will be nan.

ricardoV94: The reason this matters is that we get a simpler graph still without silent erros

ricardoV94: If anything depended on this feel free to revert the changes and we merge your previous iteration

PR #6991, opened 2023-11-06
Fix memory leak in logp of transformed variables

### LINKED ISSUE #6990: BUG: logp of transformed variables leaks memory
### Describe the issue:

I am working with relatively large glm-s right now, and was running into very severe memory issues. 
So I broke the model into smaller sub-models and running them in sequence. It helped, but the memory footprint is still increasing on the order of gigabytes per model run, and by the fourth sub-piece still maxing out my 30gb of ram. 
If I save the intermediate result, restart the kernel and then run the last model, memory footprint is only ~6gb.

So I tried looking into it myself. First with tracemalloc (which made things very slow) and later with objgraph (which is zero overhead) and they both tell the same story: the issue is with the abstract type system (abc) as all of the .register calls leave a weakref that is never collected, so they just keep accumulating, and as best I could tell with an hour of googling, there is no easy way to reset or release them. 

Tested with all three samplers (regular, jax_numpyro, jax_blackjax) and they all exhibit the same behavior.

It would be quite difficult for me to share the models I am using, as they are generated using a set of rather complex functions that I would all have to share to provide a working example, leading to a very long and convoluted setup to what is a one-line problem. I believe the problem should be easy to reproduce with any decent-sized model and I hope everyone reading this has one lying around. If not, I can add a simple model description here.

```python
# Use any decent-sized pymc model as 'model'

import objgraph, gc # Lowest overhead debug tool to show the effect

objgraph.show_growth()
for _ in range(10):
    pm.sampling_jax.sample_numpyro_nuts(model)
    # or pm.sample(model), or pm.sampling_jax.sample_blackjax_nuts(model)
    gc.collect() # Force collection to prove accumulation
    objgraph.show_growth()

# Note counts of weakref and builtin_function_or_method increase with every run
```
In the case of my models, last two steps look like this:

```
weakref                    65217987 +33382139
builtin_function_or_method 65166223 +33371309
dict                         480797   +114889
Var                          260439    +80020
list                         319239    +71068
tuple                        193843    +31638
```

```
weakref                    82420114 +17202127
builtin_function_or_method 82366482 +17200259
dict                         529400    +48603
Var                          294436    +33997
list                         349182    +29943
tuple                        210387    +16544
```

### Error message:

_No response_

### PyMC version information:

PyMC 5.9.1

### Context for the issue:

Normally, you would not think this is an issue, but in my case, one single sub-model (out of 5) generated more than 3 x 10^9 of such references according to tracemalloc. objgraph shows more modest but still large counts at 3*10^7. In practical matters, however, my 30 GB of memory currently only manages 4 out of the 5 submodels in a row before filling up and forcing me to restart the kernel. 

I am trying to build a pipeline that automatically processes survey data we collect, and needing a python restart for it to fully run makes it really awkward to set up on the devops side.

ricardoV94:
Closes #6990 

First commit is just refactoring, second commit fixes the linked issue

----
:books: Documentation preview :books:: https://pymc--6991.org.readthedocs.build/en/6991/


ricardoV94:
Weird failure on Windows, seems related to the progress_bar. Something akin to: https://github.com/fastai/fastprogress/issues/37

tests/smc/test_smc.py
```diff
@@ -28,6 +29,8 @@
 from pymc.smc.kernels import IMH, systematic_resampling
 from tests.helpers import assert_random_state_equal
 
+_IS_WINDOWS = platform.system() == "Windows"
```
ricardoV94, line 32: @aloctavodia I had to disable the progressbar on windows. Tests were failing with a cryptic error that I feel is unrelated?

https://github.com/pymc-devs/pymc/pull/6991#issuecomment-1795066441

I couldn't get on a Windows machine to reproduce.

twiecki (approved):

pymc/logprob/transform_value.py
```diff
@@ -0,0 +1,370 @@
...
+
+
+MeasurableVariable.register(TransformedValueRV)
+
+
+@_logprob.register(TransformedValueRV)
+def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs):
+    """Compute the log-probability graph for a `TransformedRV`.
+
+    This is introduced by
```
```suggestion
    This is introduced by the `TransformValuesRewrite`
```

pymc/logprob/transform_value.py
```diff
@@ -0,0 +1,370 @@
...
+from pytensor.scan.op import Scan
+from pytensor.tensor.variable import TensorVariable
+
+from pymc.logprob.abstract import MeasurableVariable, _logprob
+from pymc.logprob.rewriting import PreserveRVMappings, cleanup_ir_rewrites_db
+from pymc.logprob.transforms import RVTransform
+
+
+class TransformedValue(Op):
+    """A no-op that identifies a transform and its un-transformed input.
```
```suggestion
    """A no-op that pairs the original value with its transformed version.
```

Issue #6697, opened 2023-04-28
CheckParameterValues not replaced in inner graphs


ricardoV94:
We include these rewrites when compiling logp graphs, but this won't replace inner logp graphs (e.g, inside Scan or not-inlined OpFromGraph)

https://github.com/pymc-devs/pymc/blob/371472dcf1c44f99a9ad4e7317606404b2a47a0e/pymc/pytensorf.py#L1142-L1144

The example in #6696 suffers from this. Sampling is quite faster when the rewrite is manually added

aerubanov:
@ricardoV94 Looks like we can check inner graph for nodes that we need to replace, and if any found - create new `Scan` or `OpFromGraph` node with changed fgraph and replace old one. In you opinion will this approach work?

ricardoV94:
Actually for `Scan` there might be a shortcut, as `Scan` has it's own `mode` object. We can simple replace the old scan by a clone whose mode includes the rewrite we want to apply.

Eventually we want to get rid of this inner mode thing, but hopefully by that time we will have made Scan simpler so that's easier to manipulate anyway.

ricardoV94:
But we can do either way. The priority is more with Scan

PR #6976, opened 2023-10-30
Merge redundant code across `logprob`, `pytensorf` and `distributions/transform`

ricardoV94:
Helps with #6908 
Closes #6859 

Also gets rid of some annoying circular imports

----
:books: Documentation preview :books:: https://pymc--6976.org.readthedocs.build/en/6976/

PR #6996, opened 2023-11-07
Default rewrites in inner graph

aerubanov:
This PR will add implementation of default rewrites for nodes with inner `fgraph` and related with issue #6697. For now I just added new test cases

----
:books: Documentation preview :books:: https://pymc--6996.org.readthedocs.build/en/6996/


ricardoV94:
The tests are great!

aerubanov:
Logic for removing `CheckParameterValues` from `Scan` is very similar to one we recently added in #6873 for moment calculation in `Scan`, so may be I will move it to some helper function later

ricardoV94:
There is some refactoring going on in #6976 that will impact the location of these changes.

One thing we need to think about is nested Scans/OpFromGraph, I think this approach will only work for the first level.

aerubanov:
I added recursive check for nested `Scans`

aerubanov:
@ricardoV94 I added logic for replacement `CheckParameterValues` by ninf in `Scan` and rebased my branch. Now I am going to add support for `OpFromGraph` and want to extract common logic for rewrites in inner fgraphs - probably some helper function. I think we can add some general solution for inner fgraphs rewriting.

aerubanov:
I added support for replacements in OpFromGraph

aerubanov:
I move some common logic into `InnerGraphRewrite` class

aerubanov:
@ricardoV94 Could you please take a look when you will have some time?

pymc/logprob/utils.py
```diff
@@ -201,18 +203,120 @@ def __str__(self):
         return f"Check{{{self.msg}}}"
 
 
-@node_rewriter(tracks=[CheckParameterValue])
+class InnerGraphRewriter:
```
ricardoV94, line 206: This may be overdoing it, but what do you think of implementing this as something like WalkingGraphRewriter: https://github.com/ricardoV94/pytensor/blob/9cf2d181f07dc99bbd2e7c9e2b4a3e1b0aeff034/pytensor/graph/rewriting/basic.py#L1998

A `WalkingNestedGraphRewriter` which applies the same `node_rewriter` to both the outer graph and inner graphs. The idea is you would pass the previous rewrite which doesn't distinguish between the core case or a `Scan` `OpFromGraph`. It would be the `WalkingNestedGraphRewriter` that would apply that logic regardless of which `NodeRewriter` it's given?

ricardoV94, line 206: This class should be implemented in `pytensorf`. The specific rewrites that use it can be implemented here though

tests/test_pytensorf.py
```diff
@@ -342,6 +342,140 @@ def test_check_parameters_can_be_replaced_by_ninf(self):
...
+
+        m.check_bounds = False
+        with m:
+            fn = compile_pymc([], xs)
+            assert np.all(fn() == 0)
+
+        m.check_bounds = True
+        with m:
+            fn = compile_pymc([], xs)
+            assert np.all(fn() == -np.inf)
```
ricardoV94, line 374: Instead of messing with models and `compile_pymc`, let's just use `pytensor.function` directly and pass the rewrites we want.

```python
pytensor.function(..., mode=get_mode().including("rewrite_name"))
```

These tests should be in `logprob/test_utils.py`

pymc/logprob/utils.py
ricardoV94: If not, this should at least inherit from `NodeRewriter`, which already takes care of enforcing a `self.transform` is implemented, so you don't have to.

pymc/logprob/utils.py
```diff
@@ -201,18 +203,120 @@ def __str__(self):
...
+    def transform(self, node: Apply) -> Union[Variable, str, None]:
+        """
+        Return: "remove" if node should be removed
+                new_node if node should be replaced by new_node
+                None if node should not be replaced
+        """
+        raise NotImplementedError
+
+    def rewrite(self, node: Union[Scan, OpFromGraph]):
```
ricardoV94: Just to keep with the API used by other PyTensor rewrites

```suggestion
    def apply(self, fgraph: FunctionGraph, node: Union[Scan, OpFromGraph]):
```

ricardoV94 (commented):
This is great. Functionality wise I think it's complete. I just have some suggestions to make it more similar to other PyTensor rewriters, and some test organization. Feel free to push back on my suggestions.

pymc/logprob/utils.py
aerubanov: Hmm, sounds great. I've been thinking about something like that, not sure how to best implement it.

tests/test_pytensorf.py
aerubanov: I will change and move tests

aerubanov:
I will re-implement this with `WalkingNestedGraphRewriter` after merging https://github.com/pymc-devs/pytensor/pull/556

PR #6997, opened 2023-11-07
Add Hilbert Space Student T Process

### LINKED ISSUE #6920: HSTP
### Description

After some research, seems like if you replace `Normal` distribution with `StudentT` distribution and get HSTP

Here is a related paper as well
https://proceedings.mlr.press/v202/sellier23a.html

https://github.com/pymc-devs/pymc/blob/bc4b51dcaea4e0e9fdf8268672bb996a18628944/pymc/gp/hsgp_approx.py#L322-L331

Turning into

```python
        if self._parameterization == "noncentered":
            self._beta = pm.StudentT(
                f"{name}_hsgp_coeffs_", self.nu, size=self._m_star - int(self._drop_first)
            )
            self._sqrt_psd = sqrt_psd
            f = self.mean_func(X) + phi @ (self._beta * self._sqrt_psd)


        elif self._parameterization == "centered":
            self._beta = pm.StudentT(f"{name}_hsgp_coeffs_", self.nu, sigma=sqrt_psd)
            f = self.mean_func(X) + phi @ self._beta
```

Alternatively, you can use Fisher transfrom or any other inverse normalizing transformations to "pull out" `nu`

https://www.jstor.org/stable/2334301

ferrine:
Closes #6920 

----
:books: Documentation preview :books:: https://pymc--6997.org.readthedocs.build/en/6997/

### DIFF
--- a/pymc/gp/hsgp_approx.py
+++ b/pymc/gp/hsgp_approx.py
@@ -15,6 +15,7 @@
 import numbers
 import warnings
 
+from abc import ABC, abstractmethod
 from types import ModuleType
 from typing import Optional, Sequence, Union
 
@@ -28,6 +29,8 @@ from pymc.gp.gp import Base
 from pymc.gp.mean import Mean, Zero
 
 TensorLike = Union[np.ndarray, pt.TensorVariable]
+
+__all__ = ["HSGP", "HSTP"]
 
 
 def set_boundary(Xs: TensorLike, c: Union[numbers.Real, TensorLike]) -> TensorLike:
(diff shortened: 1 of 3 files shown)


bwengals:
After seeing this there's a couple things that make me prefer the option of passing `nu` as an optional kwarg to `HSGP.__init__`.   Mostly am concerned that Periodic HSGP https://github.com/pymc-devs/pymc/pull/6877 won't fit within this pattern, which is a bit difficult because it's structured differently itself (an approximate spectral density and a different basis).  Also it's a bit unfortunate that the docstring is split, so now it ends up being mostly boilerplate for the t-process version.  

What about `nu` as a kwarg for `HSGP`, with a default of `None`?  And use the `_base_dist` idea that you have to set it.  The docstring of `HSGP` can explain the `nu` argument.  Then we could add a function to be an alias for `HSTP`, like

```python
def HSTP(mean_func, cov_func, nu):
    """ Hilbert Space Student-t process
    (Specifi
    Reference: ...
    """
    return HSGP(...)
```

Then we don't need to think about an inheritance structure or refactor again so it works with HSGP Periodic, especially because it'd be great to have HSTP Periodic.

RE _base_dist, could it be _anything_?  I would expect it'd only be `Normal` or `StudentT`.  Might be nice to be explicit about that.

But I'm super excited to use this, I think HSTP will be used as a default before HSGP in a lot of cases.

pymc/gp/hsgp_approx.py
```diff
@@ -372,3 +300,205 @@ def conditional(self, name: str, Xnew: TensorLike, dims: Optional[str] = None):
...
+        super().__init__(
+            m,
+            L,
+            c,
+            drop_first,
+            parameterization,
+            mean_func=mean_func,
+            cov_func=scale_func,
+        )
+        self.nu = pt.as_tensor_variable(nu)
```
ricardoV94, line 501: Referencing variables in the scope of the class is what's messing up with https://github.com/pymc-devs/pymc/issues/6883

Is there another way?

PS: I know this is just the tip of the iceberg but would be great to figure it out sooner rather than later.

ferrine:
Let's wait the https://github.com/pymc-devs/pymc/issues/6883 and then continue, so it will be one round of refactoring

bwengals:
> Let's wait the https://github.com/pymc-devs/pymc/issues/6883 and then continue, so it will be one round of refactoring

Are you sure?  Pytensorifying the GP and covariance function library is a huge refactor.  I think that it'd be great to get HSTP's in ASAP, and a refactor isn't necessary to do so (see my comment from Nov 7 above here).

ricardoV94:
I agree with @bwengals, it might take a while to do the other changes.

pymc/gp/hsgp_approx.py
bwengals: > Is there another way?

Yeah, probably for the GP classes it's possible, but I think for the covariance functions is where it will be a lot harder.

ferrine:
Still not sure what should be the design choice, and should this eve stop me from taking any branch I find more appealing. We'll anyway refactor everything as far as I get it

bwengals:
What about just keeping the `_base_dist` idea like you have, and have `nu` be a optional kwarg to `HSGP`?   Then `_base_dist` can either be `StudentT` if `nu` is not None or `Normal` if `nu` is None.  That would keep it all within `HSGP`, and you could just modify the docstring once there.  I think it'd make HSTPs easier for users to discover.  Then if you wanted to make an quick alias for `HSTP` you could.  wdyt?

ferrine:
After all the refactorings in HSGP, it makes sense to create the PR from scratch

PR #7007, opened 2023-11-11
ChiSquared now returns a Gamma random variable

### LINKED ISSUE #6994: Remove ChiSquared Distribution in favor of helper
### Description

It's just a `Gamma(alpha=nu/2, beta=1/2)`. No point in having the extra code:

https://github.com/pymc-devs/pymc/blob/ec4407d3db2c8ebe77e62c65f2eda3f5703e6279/pymc/distributions/continuous.py#L2378-L2434

```python
class Chisquare:
  """docs"""

  def __new__(cls, name, nu, **kwargs):
    return Gamma(name, nu/2, 1/2, **kwargs)
  
  @classmethod
  def dist(cls, nu, **kwargs):
    return Gamma.dist(nu/2, 1/2, **kwargs)
```

Likewise, the `ChiSquareRV` in PyTensor can be removed in favor of a helper that just returns the right gamma:

https://github.com/pymc-devs/pytensor/blob/3169197c54e2cab7ba27c647c760b93e12abca5e/pytensor/tensor/random/basic.py#L490-L540

williambdean:
**What is this PR about?**
Closes https://github.com/pymc-devs/pymc/issues/6994

**Checklist**

## Major / Breaking Changes
- moment automatically from custom dist

## New features
- N/A

## Bugfixes
- N/A

## Documentation
- N/A

## Maintenance
- ChiSquared used Gamma relationship now

----
:books: Documentation preview :books:: https://pymc--7007.org.readthedocs.build/en/7007/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2378,7 +2378,7 @@ class ChiSquared:
     :math:`\chi^2` log-likelihood.
 
     This is the distribution from the sum of the squares of :math:`\nu` independent standard normal random variables or a special
-    case of the gamma distribution.
+    case of the gamma distribution with :math:`\alpha = \nu/2` and :math:`\beta = 1/2`.
 
     The pdf of this distribution is
 
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -55,7 +55,6 @@ from pytensor.tensor.random.basic import (
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.variable import TensorConstant
 
-from pymc.distributions.distribution import CustomDist
 from pymc.logprob.abstract import _logprob_helper
 from pymc.logprob.basic import icdf
 
@@ -2378,11 +2377,16 @@ class ChiSquared:
     r"""
     :math:`\chi^2` log-likelihood.
 
+    This is the distribution from the sum of the squares of :math:`\nu` independent standard normal random variables or a special
+    case of the gamma distribution.
+
     The pdf of this distribution is
 
     .. math::
 
        f(x \mid \nu) = \frac{x^{(\nu-2)/2}e^{-x/2}}{2^{\nu/2}\Gamma(\nu/2)}
+
+    Read more about the :math:`\chi^2` distribution at https://en.wikipedia.org/wiki/Chi-squared_distribution
 
     .. plot::
         :context: close-figs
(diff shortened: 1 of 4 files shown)


pymc/distributions/continuous.py
```diff
@@ -2413,24 +2413,27 @@ class ChiSquared(PositiveContinuous):
...
+            dist=cls.chisquared_dist,
+            class_name="ChiSquared",
+            **kwargs,
+        )
 
-    def logcdf(value, nu):
-        return _logcdf_helper(Gamma.dist(alpha=nu / 2, beta=0.5), value)
+    @classmethod
+    def dist(cls, nu, **kwargs):
+        return CustomDist.dist(
```
ricardoV94: We don't need a CustomDist, we can return Gamma directly since it's just a parametrization not a variable transformation.

That mean we need a `__new__` method as well

ricardoV94 (changes requested):

pymc/distributions/continuous.py
williambdean: I went with this route as the suggested makes the `__repr__` says its just a Gamma distribution. I thought itd be good to display a different name. I can go with your suggestion if you disagree

ricardoV94: We shouldn't create a new object, indeed it's just a gamma. We can even state so in the docstrings

williambdean: Where in the docs? I'm just seeing the pdf https://www.pymc.io/projects/docs/en/stable/api/distributions/generated/pymc.ChiSquared.html

I will add this relationship in the docstring under the class

ricardoV94 (commented):
Wr should also remove the RV from PyTensor and use only a helper there

wd60622:
> Wr should also remove the RV from PyTensor and use only a helper there

sounds good. I will take a look

wd60622:
New PyMC docs
Screenshot 2023-11-12 at 14 40 19

wd60622:
I'm interpreting the tests as ones that needs to be removed. For instance, only the `check_pymc_draws_match_reference` check passes still. One depends on rv_op attribute

The other is for the rewrite which wouldn't be supported anymore through that chisquare

ricardoV94 (approved):

ricardoV94:
> The other is for the rewrite which wouldn't be supported anymore through that chisquare

Just make it use `pm.Chisquare.dist` instead of `pt.random.chisquare` as it was before

wd60622:
All of the `pt.random.chisquare` are replaced with `ChiSquare`. Have all tests passing locally

PR #7008, opened 2023-11-13
fix: adding py.typed file as marker for mypy

Drezzerd:
**What is this PR about?**
This PR add a new file `py.typed` at the root of the module to fix this [issue](https://discourse.pymc.io/t/missing-library-stubs-or-py-typed-marker-are-we-missing-something/13268)
The goal is to allow mypy analyze

**Checklist**

## Major / Breaking Changes
- None

## New features
- Enable mypy type analysis
- Fixes mypy error message

## Bugfixes
- `my_file.py:1: error: Skipping analyzing “pymc”: module is installed, but missing library stubs or py.typed marker [import-untyped]`

## Documentation
- [PEP 561](https://peps.python.org/pep-0561/) 
- [mypy docs](https://mypy.readthedocs.io/en/stable/running_mypy.html#fix-missing-imports)

## Maintenance
- None

----
:books: Documentation preview :books:: https://pymc--7008.org.readthedocs.build/en/7008/


thomasaarholt:
The [pyright docs](https://github.com/microsoft/pyright/blob/419ae4212f38feda0f4194203f9ce3e5fa7a634c/docs/type-stubs.md) also make it clear that a package that does type hinting (it doesn't need to be _fully_ typed, or even _correctly_ typed), should have a py.typed file in the root package directory.

This PR makes sense.

maresb:
`py.typed` should go in the package root, not the project root. (Just pushed a commit to fix.)

maresb:
Our typing seems very rudimentary to me, so I wonder if doing this will lead to more chaos than it solves.

Out of curiosity I checked the situation with scipy, and they still don't have a `py.typed` file yet. They added one in https://github.com/scipy/scipy/pull/11946 and removed in in https://github.com/scipy/scipy/pull/12210 shortly after. However, the removal doesn't seem associated with any particular complaint.

maresb:
@michaelosthege, could you explain what this is for? It seems to be unused.

https://github.com/pymc-devs/pymc/blob/05b557f6460a10c29c3db33690ee535f5b1ecde0/scripts/run_mypy.py#L53-L61

lucianopaz:
I'm confused with the [PEP](https://peps.python.org/pep-0561/#packaging-type-information). It says that we need to include the py.typed file as a package data in pyproject.toml, but maybe this is not necessary thanks to [this](https://github.com/pypa/setuptools/pull/4021)?

maresb:
@lucianopaz I think it's included due to setuptools-scm. I verified that it's included in both sdist and wheels.

michaelosthege:
The `enforce_pep561` was a hack to add a `py.typed` to Aesara back in the day when the Aesara devs categorically refused to add `py.typed` until type hints were flawless.
We used this to already start working on improving PyMC type hints.

Meanwhile, PyTensor [has the `py.typed` marker](https://github.com/pymc-devs/pytensor/blob/863efc014503318c76b53f599646a631be592e50/pyproject.toml#L88) so the hack can safely be removed 👍

maresb:
Thanks a lot @michaelosthege for the context! Seems like you already have some experience with `py.typed`. Do you see any potential disadvantages to enabling?

I'm assuming that users who are using pymc as a library may see a bunch of new type errors, but the main userbase is probably people writing notebooks who just want their IDEs to give type hints, so the benefits will greatly outweigh the downsides. Does this seem like an accurate assessment?

michaelosthege:
> Thanks a lot @michaelosthege for the context! Seems like you already have some experience with `py.typed`. Do you see any potential disadvantages to enabling?
> 
> I'm assuming that users who are using pymc as a library may see a bunch of new type errors, but the main userbase is probably people writing notebooks who just want their IDEs to give type hints, so the benefits will greatly outweigh the downsides. Does this seem like an accurate assessment?

I'll have to try this myself in one project with a rather involved model. If you check the mypy logs, we still have a bunch (>100?) errors that get ignore and many of those are due to function dispatching and PyMCs abuse of `__new__` in the distribution API.

95 % of user interaction with the PyMC API goes through the distribution API (which totally breaks mypy and IDEs) and another 4 % is via `pm.sample*` which is trivial w.r.t. type hints.

@thomasaarholt did you try a real-world project with `py.typed` enabled?

twiecki (approved):

PR #7011, opened 2023-11-14
Added ReduceLROnPlateau callback for VI.

alvaropp:
This has been discussed in the [forums](https://discourse.pymc.io/t/learning-rate-scheduling-for-variational-inference/13075/25) and in this [issue](https://github.com/pymc-devs/pymc/issues/7010).

There’s big parallels between PyMC’s variational inference and training neural networks, where the choice of optimiser and learning rate have a huge impact on training quality and speed. A common technique for training neural networks is using learning rate schedulers which reduce the learning rate on a schedule, to get faster convergence by starting high and reducing it in successive epochs where you want to be more precise.

Currently, in PyMC, you need to specify a suitable learning rate that is used for fitting the model. Too large and it won't converge, too small and it will be too slow. Training once with a large-ish learning rate and then taking the results of that training round as a starting point for another training round with a smaller learning rate is not trivial and not very elegant.

To address this issue, I've implemented a new callback for pymc.variational, following Keras' [ReduceLROnPlateau](https://github.com/keras-team/keras/blob/v2.14.0/keras/callbacks.py). Basically it monitors the model loss at every iteration of VI.fit() and it reduces the learning rate by a user-specified amount if the loss doesn't improve after a user-specified number of iterations.

## Major / Breaking Changes
None (I think!)

## New features
Added a ReduceLROnPlateau callback.

## Bugfixes
None.

## Documentation
None so far, but it would be nice to add an example.

## Maintenance
None.

----
:books: Documentation preview :books:: https://pymc--7011.org.readthedocs.build/en/7011/

### DIFF
--- a/pymc/variational/callbacks.py
+++ b/pymc/variational/callbacks.py
@@ -13,7 +13,6 @@
 #   limitations under the License.
 
 import collections
-
 from typing import Callable, Dict
 
 import numpy as np
@@ -101,8 +100,8 @@ class ReduceLROnPlateau(Callback):
 
     Parameters
     ----------
-    optimiser: callable
-        PyMC optimiser
+    optimizer: callable
+        PyMC optimizer
     factor: float
         factor by which the learning rate will be reduced: `new_lr = lr * factor`
     patience: int
@@ -117,13 +116,13 @@ class ReduceLROnPlateau(Callback):
 
     def __init__(
         self,
-        optimiser,
+        optimizer,
         factor=0.1,
         patience=10,
         min_lr=1e-6,
         cooldown=0,
     ):
-        self.optimiser = optimiser
+        self.optimizer = optimizer
         self.factor = factor
         self.patience = patience
         self.min_lr = min_lr
@@ -156,10 +155,10 @@ class ReduceLROnPlateau(Callback):
                 self.wait = 0
 
     def reduce_lr(self):
-        old_lr = float(self.optimiser.keywords["learning_rate"])
+        old_lr = float(self.optimizer.keywords["learning_rate"])
         if old_lr > self.min_lr:
             new_lr = max(old_lr * self.factor, self.min_lr)
-            self.optimiser.keywords["learning_rate"] = new_lr
+            self.optimizer.keywords["learning_rate"] = new_lr
 
     def in_cooldown(self):
         return self.cooldown_counter > 0
--- a/pymc/variational/callbacks.py
+++ b/pymc/variational/callbacks.py
@@ -13,6 +13,7 @@
 #   limitations under the License.
 
 import collections
+
 from typing import Callable, Dict
 
 import numpy as np
--- a/pymc/variational/callbacks.py
+++ b/pymc/variational/callbacks.py
@@ -18,7 +18,13 @@ from typing import Callable, Dict
 
 import numpy as np
 
-__all__ = ["Callback", "CheckParametersConvergence", "ReduceLROnPlateau", "Tracker"]
+__all__ = [
+    "Callback",
+    "CheckParametersConvergence",
+    "ExponentialDecay",
+    "ReduceLROnPlateau",
+    "Tracker",
+]
 
 
 class Callback:
--- a/pymc/variational/callbacks.py
+++ b/pymc/variational/callbacks.py
@@ -18,7 +18,9 @@ from typing import Callable, Dict
 
 import numpy as np
 
-__all__ = ["Callback", "CheckParametersConvergence", "Tracker"]
+from pytensor import shared
+
+__all__ = ["Callback", "CheckParametersConvergence", "ReduceLROnPlateau", "Tracker"]
 
 
 class Callback:
@@ -91,6 +93,78 @@ class CheckParametersConvergence(Callback):
     @staticmethod
     def flatten_shared(shared_list):
         return np.concatenate([sh.get_value().flatten() for sh in shared_list])
+
+
+class ReduceLROnPlateau(Callback):
+    """Reduce learning rate when the loss has stopped improving.
+
+    This is inspired by Keras' homonymous callback:
+    https://github.com/keras-team/keras/blob/v2.14.0/keras/callbacks.py
+
+    Parameters
+    ----------
+    learning_rate: shared
+        shared variable containing the learning rate
+    factor: float
+        factor by which the learning rate will be reduced: `new_lr = lr * factor`
+    patience: int
+        number of epochs with no improvement after which learning rate will be reduced
+    min_lr: float
+        lower bound on the learning rate
+    cooldown: int
+        number of iterations to wait before resuming normal operation after lr has been reduced
+    verbose: bool
+        false: quiet, true: update messages
+    """
+
+    def __init__(
+        self,
+        initial_learning_rate: shared,
+        factor=0.1,
+        patience=10,
+        min_lr=1e-6,
+        cooldown=0,
+    ):
+        self.learning_rate = initial_learning_rate
+        self.factor = factor
+        self.patience = patience
+        self.min_lr = min_lr
+        self.cooldown = cooldown
+
+        self.cooldown_counter = 0
+        self.wait = 0
+        self.best = float("inf")
+        self.old_lr = None
+
+    def __call__(self, approx, loss_hist, i):
+        current = loss_hist[-1]
+
+        if np.isinf(current):
+            return
+
+        if self.in_cooldown():
+            self.cooldown_counter -= 1
+            self.wait = 0
+            return
+
+        if current < self.best:
+            self.best = current
+            self.wait = 0
+        elif not np.isinf(self.best):
+            self.wait += 1
+            if self.wait >= self.patience:
+                self.reduce_lr()
+                self.cooldown_counter = self.cooldown
+                self.wait = 0
+
+    def reduce_lr(self):
+        old_lr = float(self.learning_rate.get_value())
+        if old_lr > self.min_lr:
+            new_lr = max(old_lr * self.factor, self.min_lr)
+            self.learning_rate.set_value(new_lr)
+
+    def in_cooldown(self):
+        return self.cooldown_counter > 0
 
 
 class Tracker(Callback):
--- a/pymc/variational/callbacks.py
+++ b/pymc/variational/callbacks.py
@@ -106,7 +106,7 @@ class LearningRateScheduler(Callback):
         self.optimizer.keywords["learning_rate"] = new_lr
 
 
-class ExponentialScheduler(LearningRateScheduler):
+class ExponentialDecay(LearningRateScheduler):
     """
     Exponentially decays the learning rate.
 
--- a/pymc/variational/callbacks.py
+++ b/pymc/variational/callbacks.py
@@ -93,8 +93,56 @@ class CheckParametersConvergence(Callback):
         return np.concatenate([sh.get_value().flatten() for sh in shared_list])
 
 
-class ReduceLROnPlateau(Callback):
-    """Reduce learning rate when the loss has stopped improving.
+class LearningRateScheduler(Callback):
+    """Baseclass for learning rate schedulers."""
+
+    def __init__(self, optimizer):
+        self.optimizer = optimizer
+
+    def __call__(self, approx, loss_hist, i):
+        raise NotImplementedError("Must be implemented in subclass.")
+
+    def _set_new_lr(self, new_lr):
+        self.optimizer.keywords["learning_rate"] = new_lr
+
+
+class ExponentialScheduler(LearningRateScheduler):
+    """
+    Exponentially decays the learning rate.
+
+    Parameters
+    ----------
+    decay_steps : int
+        Number of steps at which the learning rate decay happens.
+    decay_rate : float
+        Rate of decay.
+    min_lr: float
+        lower bound on the learning rate
+    staircase : bool
+        If True, decay the learning rate at discrete intervals.
+    """
+
+    def __init__(self, optimizer, decay_steps, decay_rate, min_lr=1e-6, staircase=False):
+        super().__init__(optimizer)
+        self.decay_steps = decay_steps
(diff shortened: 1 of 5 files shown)


alvaropp:
@jessegrabowski raised a good point that it's not ideal to have the user need to define a shared variable, so I've reworked my callback so that it takes a vanilla PyMC optimiser instance and it modifies its learning rate directly:

```python
with pm.Model() as model:
    [...]

    optimiser = pm.adam(learning_rate=0.1)
    tracker = Tracker(lr=lambda: optimiser.keywords["learning_rate"])
    fit = pm.fit(
        obj_optimizer=optimiser,
        callbacks=[
            tracker,
            CheckParametersConvergence(tolerance=0.01),
            ReduceLROnPlateau(
                optimiser=optimiser,
            ),
        ],
    )
```

pymc/variational/callbacks.py
```diff
@@ -93,6 +93,78 @@ def flatten_shared(shared_list):
...
+        lower bound on the learning rate
+    cooldown: int
+        number of iterations to wait before resuming normal operation after lr has been reduced
+    verbose: bool
+        false: quiet, true: update messages
+    """
+
+    def __init__(
+        self,
+        optimiser,
```
twiecki: -> `optimizer`

twiecki (approved):

alvaropp:
Is it possible to add an example somewhere? maybe here: https://www.pymc.io/projects/examples/en/latest/gallery.html#variational-inference

pymc/variational/callbacks.py
```diff
@@ -93,6 +93,77 @@ def flatten_shared(shared_list):
...
+        lower bound on the learning rate
+    cooldown: int
+        number of iterations to wait before resuming normal operation after lr has been reduced
+    verbose: bool
+        false: quiet, true: update messages
+    """
+
+    def __init__(
+        self,
+        optimizer,
```
jessegrabowski, line 175: Does the user have to provide this? Can it instead be inferred somehow from the host VI object? It's ugly to have to pass the optimizer twice (once for the VI itself, then again in the callback)

tests/variational/test_callbacks.py
```diff
@@ -52,3 +52,33 @@ def test_tracker_callback():
     tracker = pm.callbacks.Tracker(bad=lambda t: t)  # bad signature
     with pytest.raises(TypeError):
         tracker(None, None, 1)
+
+
+def test_reducelronplateau_callback():
+    optimizer = pm.adam(learning_rate=0.1)
```
jessegrabowski: Use a pytest.mark.parameterize here to test all the optimizers and make sure there's no unexpected behaviors between them

pymc/variational/callbacks.py
```diff
@@ -93,6 +93,77 @@ def flatten_shared(shared_list):
...
+            self.best = current
+            self.wait = 0
+        elif not np.isinf(self.best):
+            self.wait += 1
+            if self.wait >= self.patience:
+                self.reduce_lr()
+                self.cooldown_counter = self.cooldown
+                self.wait = 0
+
+    def reduce_lr(self):
```
jessegrabowski: I would still prefer that this was done symbolically with shared variables, because it will allow for composition between learning rate annealing strategies.

jessegrabowski (changes requested):
I still prefer the Torch-style API to the callback API, but this PR is great so I'll let it go. The tests need to be a bit more robust, though. I also want to keep advocating for a slightly more general solution that will allow for 1) different learning rate adjustment strategies, and 2) composition of learning rate strategies. 

I'm not saying this PR needs to implement every single learning rate scheduler, but it would be nice to have LearningRateStrategy base class that could be extended in the future, then subclass ReduceLROnPlateau from that

pymc/variational/callbacks.py
alvaropp: Well, this would be great, but I haven't figured out whether it's possible. Probably one for someone more familiar with the codebase :)

alvaropp:
> I still prefer the Torch-style API to the callback API, but this PR is great so I'll let it go. The tests need to be a bit more robust, though. I also want to keep advocating for a slightly more general solution that will allow for 1) different learning rate adjustment strategies, and 2) composition of learning rate strategies.
> 
> I'm not saying this PR needs to implement every single learning rate scheduler, but it would be nice to have LearningRateStrategy base class that could be extended in the future, then subclass ReduceLROnPlateau from that

I've implemented an ExponentialDecay callback and refactored things a bit.
plateau
expdecay
expdecay_staircase

alvaropp:
Hi @jessegrabowski, any concrete ideas for taking this PR forward?

As you said, it would be nice to infer the optimiser from the code rather than pass it to the callback, but I don't know how to go about that.

```python
with pm.Model() as model:
    [...]

    optimiser = pm.adam(learning_rate=0.1)
    fit = pm.fit(
        obj_optimizer=optimiser,
        callbacks=[
            ReduceLROnPlateau(optimiser=optimiser)
        ],
    )
```

jessegrabowski:
@alvaropp sorry for letting this get stale, I am going to do a careful review now/this weekend so you can get back on track and merged in ASAP

jessegrabowski:
I'm going over the code carefully, and I don't think the callbacks as written do anything. There is a misunderstanding how about compiled pytensor functions work.

The implementation is very nice, and uses the following logic: 
1. There is a `learning_rate` parameter in every optimizer object. 
2. The optimizer object is called iteratively inside the `_iterate_with_loss` method of the `Inference` class.
3. Therefore, we instantiate the optimizer class once and share it between a callback and the `Inference` class
4. At each step, the callback will update the `learning_rate` of the optimizer
5. Which will be propagated to the `Inference` class (since it's the same object).

Unfortunately, this is not at all how the compiled pytensor function works. First, the implementation of the optimizers is deceptive. They are not classes, they are partial functions, whose only role is to return an `updates` dictionary.

This updates dictionary is important, because after a pytensor function is compiled, **changes to variables outside the function will have no effect**. Consider the following graph:

```
x = pt.dscalar('x')
y = 1.0
z = x + y
f = pytensor.function([x], z)

print(f(4)) # 5.0
print(f(5)) # 6.0

y = 2

print(f(4)) # 5.0
```

What happened? After running `pytensor.function` to compile the function, the variable `y` in the python namespace no longer has anything to do with the computation. If we try to naively update it, nothing will happen. This is exactly the situation we are in when running `pm.fit`. If you check the `fit` method in the `Inference` class, the line that actually does all the computation is [this innocuous one](https://github.com/pymc-devs/pymc/blob/02cbac6592506a547117b2daa3f05847d16fe063/pymc/variational/inference.py#L204). What, then, is step_func? It's created [here](https://github.com/pymc-devs/pymc/blob/02cbac6592506a547117b2daa3f05847d16fe063/pymc/variational/opvi.py#L324), and as you can see, it's a compiled pytensor function (compiled using `compile_pymc` rather than `pytensor.function`, but if you keep digging you'll see `pytensor.function` is called by `compile_pymc`). So we're in exactly the case of my code snippet above.

To illustrate what's going on, I made a dummy "optimizer" that always adds the learning rate to the parameters:

```
from pymc.variational.updates import _get_call_kwargs, get_or_compute_grads
from functools import partial
from collections import OrderedDict

def dummy_optimizer(loss_or_grads=None, params=None, learning_rate=1e-3):
    if loss_or_grads is None and params is None:
        return partial(dummy_optimizer, **_get_call_kwargs(locals()))
    elif loss_or_grads is None or params is None:
        raise ValueError("Please provide both `loss_or_grads` and `params` to get updates")
    # grads = get_or_compute_grads(loss_or_grads, params)
    updates = OrderedDict()

    for param in params:
        new_param = param - learning_rate
        updates[param] = new_param

    return updates
```

We can use this optimizer to check the effect of the current schedulers. Helper functions:

```
def make_regression():
    X = np.random.normal(size=(100,))
    beta = np.random.normal()
    alpha = np.random.normal()
    noise = np.random.normal(size=(100,))
    y = alpha + X * beta + noise

    return X, y

def make_model():
    X, y = make_regression()
    with pm.Model() as mod:
        b = pm.Normal('b')
        a = pm.Normal('a')
        # sigma = pm.HalfNormal('sigma')
    
        mu = a + X * b
        y_hat = pm.Normal('y_hat', mu, 1, observed=y)

    return mod
```

Here's a base case with no scheduling callback. I do use a tracker so we can see how the dummy works:

```
learning_rate = 1
optimizer = dummy_optimizer(learning_rate=learning_rate)

with make_model():
    advi = pm.ADVI()
    tracker = pm.callbacks.Tracker(mean=advi.approx.mean.eval)
    approx = advi.fit(100, callbacks=[tracker], obj_optimizer=optimizer)
```

Check that the parameters are deterministically updated by the learning rate at every step:

```
mean_history = np.r_[tracker['mean']]
np.all(np.diff(mean_history, axis=0) == learning_rate) # True
```

Now run it again with the ExponentialDecay schedule callback:

```
learning_rate = 1
optimizer = dummy_optimizer(learning_rate=learning_rate)
scheduler = pm.callbacks.ExponentialDecay(optimizer=optimizer, decay_steps=10, decay_rate=0.5)

with make_model():
    advi = pm.ADVI()
    tracker = pm.callbacks.Tracker(mean=advi.approx.mean.eval)
    approx = advi.fit(100, callbacks=[tracker, scheduler], obj_optimizer=optimizer)
```

Check the mean history:

```
mean_history = np.r_[tracker['mean']]
np.all(np.diff(mean_history, axis=0) == learning_rate) # True
```

As you can see, the learning rate has not changed. We can check that the scheduler updated as expected:

```
scheduler.optimizer.keywords['learning_rate'] == 1 * 0.5 ** 10 # True
```

So what is the solution? To interact with a compiled function, you need to use shared variables. These are a special type of pytensor object that can be adjusted between function calls. Returning to the code snippet, let's use a shared variable:

```
x = pt.dscalar('x')
y = pytensor.shared(1.0, name='y')
z = x + y
f = pytensor.function([x], z)

print(f(4)) # 5.0
print(f(5)) # 6.0

y.set_value(2)

print(f(4)) # 6.0
```

After using a shared variable, we can adjust the value of `y` to adjust the output of `f`, even though it's already compiled. 

The way that shared variables can be **automatically** updated from run-to-run of a function is through the use of an update dictionary. An update dictionary is a mapping from old shared values to new shared values. This is exactly the dictionary that is returned by the optimizer, as noted above. Let's take a look at the SGD function from [here](https://github.com/pymc-devs/pymc/blob/02cbac6592506a547117b2daa3f05847d16fe063/pymc/variational/updates.py#L186):

```
    grads = get_or_compute_grads(loss_or_grads, params)
    updates = OrderedDict()

    for param, grad in zip(params, grads):
        updates[param] = param - learning_rate * grad

    return updates
```

You can see that it computes the gradients of the loss function, then computes SGD update rule for each parameter in turn. It stores these in an `OrderedDict`, where the keys are the old parameters, and the values are the updated parameters. It then returns the dictionary.

As a final demonstration, let's adjust the dummy optimizer to have an exponential decay on the learning rate:

```
def dummy_optimizer_w_decay(loss_or_grads=None, params=None, learning_rate=1e-3):
    if loss_or_grads is None and params is None:
        return partial(dummy_optimizer, **_get_call_kwargs(locals()))
    elif loss_or_grads is None or params is None:
        raise ValueError("Please provide both `loss_or_grads` and `params` to get updates")
    # grads = get_or_compute_grads(loss_or_grads, params)
    updates = OrderedDict()
    if not isinstance(learning_rate, pt.sharedvar.SharedVariable):
        learning_rate = pytensor.shared(pm.floatX(learning_rate), name='learning_rate')

    for param in params:
        new_param = param + learning_rate
        updates[param] = new_param

    updates[learning_rate] = learning_rate * 0.9
    return updates
```

And see how this looks in a model:

```
learning_rate = 1
optimizer = dummy_optimizer(learning_rate=learning_rate)

with make_model():
    advi = pm.ADVI()
    tracker = pm.callbacks.Tracker(mean=advi.approx.mean.eval)
    approx = advi.fit(100, callbacks=[tracker], obj_optimizer=optimizer)
```

Here's a plot of the tracked parameters. As you can see, we now get the desired updating in the learning rate.

Untitled

jessegrabowski:
So can the callback approach be saved? Unfortunately, I don't think so. By the time the callbacks are invoked, in the `fit` loop,  the `step_func` function is already compiled. At this stage it's too late to adjust the function without some pretty impressive gymnastics. This is going to have to go back to the drawing board using something like the wrapper approach.

alvaropp:
D'oh, good spot!

fonnesbeck:
It seems like a good idea to use schedulers universally, with a `Constant` scheduler as the default. This may be a cleaner API and allow us to easily infer the optimizer, etc.

jessegrabowski:
My proposed changes follow the pytorch API that looks like this:

```python
optimizer = torch.optim.SGD(model.parameters(), lr=0.1, momentum=0.9)
scheduler = ReduceLROnPlateau(optimizer, 'min')
```

So the user has to pass the optimizer to the scheduler explicitly in the code. It's also written so that users can compose schedulers by chaining them together. 

From a clean code perspective I like the idea of a "ConstantOptimizer" as a base class from which others inherit from. A universal default scheduler makes sense if it's a keyword for example, but not as much this way?

sandstromviktor:
What is the status of this PR? I think this feature is highly needed.

sandstromviktor:
For anyone needing this sort of feature in their own work, here is a simple way to change learning rate during training.
This is not particularly nice, but it works and is simple. 
```python

base_lr = 0.1

# Create a shared tensorvariable
lr = pytensor.shared(base_lr)

# Define your optimizer
optimizer = pm.adam(learning_rate=lr)

# Create some function that changes the learning rate
def update_learning_rate(approx, loss, iteration):
    # Change learning rate after 1000 iterations
    if iteration > 1000:
       lr.set_value(0.01)

advi = advi.fit(
    10000,
    obj_optimizer=optimizer,
    callbacks=[update_learning_rate],
)
```

PR #7014, opened 2023-11-17
Add examples explaining advanced applications of `sample_posterior_predictive`

ricardoV94:
This is still a draft. I want to add two more examples first about the default ppc and predictions use-cases

Related to #7069 

## TODO
- [x] Add example about common use for in-sample posterior_predictive
- [x] Add example about common use for predictions
- [x] Add examples on using new models 
- [x] Change conclusion about non-dependence on deterministics or fix behavior
- [x] ~Explain volatility when MutableData changes~ I think this is itself a mine-field and I would rather change the behavior first

Link to function docs preview: https://pymcio--7014.org.readthedocs.build/projects/docs/en/7014/api/generated/pymc.sampling.forward.sample_posterior_predictive.html

----
:books: Documentation preview :books:: https://pymc--7014.org.readthedocs.build/en/7014/

### DIFF
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -448,35 +448,48 @@ def sample_posterior_predictive(
     idata_kwargs: Optional[dict] = None,
     compile_kwargs: Optional[dict] = None,
 ) -> Union[InferenceData, dict[str, np.ndarray]]:
-    """Generate posterior predictive samples from a model given a trace.
+    """Generate forward samples for `var_names`, conditioned on the posterior samples of variables found in the `trace`.
+
+    This method can be used to perform different kinds of model predictions, including posterior predictive checks.
+
+    The matching of unobserved model variables, and posterior samples in the `trace` is made based on the variable
+    names. Therefore, a different model than the one used for posterior sampling may be used for posterior predictive
+    sampling, as long as the variables whose posterior we want to condition on have the same name, and compatible shape
+    and coordinates.
+
 
     Parameters
     ----------
     trace : backend, list, xarray.Dataset, arviz.InferenceData, or MultiTrace
-        Trace generated from MCMC sampling, or a list of dicts (eg. points or from find_MAP()),
-        or xarray.Dataset (eg. InferenceData.posterior or InferenceData.prior)
+        Trace generated from MCMC sampling, or a list of dicts (eg. points or from :func:`~pymc.find_MAP`),
+        or :class:`xarray.Dataset` (eg. InferenceData.posterior or InferenceData.prior)
     model : Model (optional if in ``with`` context)
         Model to be used to generate the posterior predictive samples. It will
-        generally be the model used to generate the ``trace``, but it doesn't need to be.
-    var_names : Iterable[str]
+        generally be the model used to generate the `trace`, but it doesn't need to be.
+    var_names : Iterable[str], optional
         Names of variables for which to compute the posterior predictive samples.
+        By default, only observed variables are sampled.
+        See the example below for what happens when this argument is customized.
     sample_dims : list of str, optional
         Dimensions over which to loop and generate posterior predictive samples.
-        When `sample_dims` is ``None`` (default) both "chain" and "draw" are considered sample
+        When ``sample_dims`` is ``None`` (default) both "chain" and "draw" are considered sample
         dimensions. Only taken into account when `trace` is InferenceData or Dataset.
     random_seed : int, RandomState or Generator, optional
         Seed for the random number generator.
     progressbar : bool
-        Whether or not to display a progress bar in the command line. The bar shows the percentage
+        Whether to display a progress bar in the command line. The bar shows the percentage
         of completion, the sampling speed in samples per second (SPS), and the estimated remaining
         time until completion ("expected time of arrival"; ETA).
     return_inferencedata : bool, default True
         Whether to return an :class:`arviz:arviz.InferenceData` (True) object or a dictionary (False).
     extend_inferencedata : bool, default False
         Whether to automatically use :meth:`arviz.InferenceData.extend` to add the posterior predictive samples to
-        ``trace`` or not. If True, ``trace`` is modified inplace but still returned.
+        `trace` or not. If True, `trace` is modified inplace but still returned.
     predictions : bool, default False
-        Flag used to set the location of posterior predictive samples within the returned ``arviz.InferenceData`` object. If False, assumes samples are generated based on the fitting data to be used for posterior predictive checks, and samples are stored in the ``posterior_predictive``. If True, assumes samples are generated based on out-of-sample data as predictions, and samples are stored in the ``predictions`` group.
+        Flag used to set the location of posterior predictive samples within the returned ``arviz.InferenceData`` object.
+        If False, assumes samples are generated based on the fitting data to be used for posterior predictive checks,
+        and samples are stored in the ``posterior_predictive``. If True, assumes samples are generated based on
+        out-of-sample data as predictions, and samples are stored in the ``predictions`` group.
     idata_kwargs : dict, optional
         Keyword arguments for :func:`pymc.to_inference_data` if ``predictions=False`` or to
         :func:`pymc.predictions_to_inference_data` otherwise.
@@ -489,24 +502,221 @@ def sample_posterior_predictive(
         An ArviZ ``InferenceData`` object containing the posterior predictive samples (default), or
         a dictionary with variable names as keys, and samples as numpy arrays.
 
+
     Examples
     --------
-    Thin a sampled inferencedata by keeping 1 out of every 5 draws
-    before passing it to sample_posterior_predictive
+    Posterior predictive checks and predictions
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    The most common use of `sample_posterior_predictive` is to perform posterior predictive checks (in-sample predictions)
+    and new model predictions (out-of-sample predictions).
+
+    .. code:: python
+
+        import pymc as pm
+
+        with pm.Model(coords_mutable={"trial": [0, 1, 2]}) as model:
+            x = pm.MutableData("x", [-1, 0, 1], dims=["trial"])
+            beta = pm.Normal("beta")
+            noise = pm.HalfNormal("noise")
+            y = pm.Normal("y", mu=x * beta, sigma=noise, observed=[-2, 0, 3], dims=["trial"])
+
+            idata = pm.sample()
+            # in-sample predictions
+            posterior_predictive = pm.sample_posterior_predictive(idata).posterior_predictive
+
+        with model:
+            pm.set_data({"x": [-2, 2]}, coords={"trial": [3, 4]})
+            # out-of-sample predictions
+            predictions = pm.sample_posterior_predictive(idata, predictions=True).predictions
+
+
+    Using different models
+    ^^^^^^^^^^^^^^^^^^^^^^
+
+    It's common to use the same model for posterior and posterior predictive sampling, but this is not required.
+    The matching between unobserved model variables and posterior samples is based on the name alone.
+
+    For the last example we could have created a new predictions model. Note that we have to specify
+    `var_names` explicitly, because the newly defined `y` was not given any observations:
+
+    .. code:: python
+
+        with pm.Model(coords_mutable={"trial": [3, 4]}) as predictions_model:
+            x = pm.MutableData("x", [-2, 2], dims=["trial"])
+            beta = pm.Normal("beta")
+            noise = pm.HalfNormal("noise")
+            y = pm.Normal("y", mu=x * beta, sigma=noise, dims=["trial"])
+
+            predictions = pm.sample_posterior_predictive(idata, var_names=["y"], predictions=True).predictions
+
+
+    The new model may even have a different structure and unobserved variables that don't exist in the trace.
+    These variables will also be forward sampled. In the following example we added a new ``extra_noise``
+    variable between the inferred posterior ``noise`` and the new StudentT observational distribution  ``y``:
+
+    .. code:: python
+
+        with pm.Model(coords_mutable={"trial": [3, 4]}) as distinct_predictions_model:
+            x = pm.MutableData("x", [-2, 2], dims=["trial"])
+            beta = pm.Normal("beta")
+            noise = pm.HalfNormal("noise")
+            extra_noise = pm.HalfNormal("extra_noise", sigma=noise)
+            y = pm.StudentT("y", nu=4, mu=x * beta, sigma=extra_noise, dims=["trial"])
+
+            predictions = pm.sample_posterior_predictive(idata, var_names=["y"], predictions=True).predictions
+
+
+    For more about out-of-model predictions, see this `blog post <https://www.pymc-labs.com/blog-posts/out-of-model-predictions-with-pymc/>`_.
+
+    The behavior of `var_names`
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    The function returns forward samples for any variable included in `var_names`,
+    conditioned on the values of other random variables found in the trace.
+
+    To ensure the samples are internally consistent, any random variable that depends
+    on another random variable that is being sampled is itself sampled, even if
+    this variable is present in the trace and was not included in `var_names`.
+    The final list of variables being sampled is shown in the log output.
+
+    Note that if a random variable has no dependency on other random variables,
+    these forward samples are equivalent to their prior samples.
+    Likewise, if all random variables are being sampled, the behavior of this function
+    is equivalent to that of :func:`~pymc.sample_prior_predictive`.
+
+    .. warning:: A random variable included in `var_names` will never be copied from the posterior group. It will always be sampled as described above. If you want, you can copy manually via ``idata.posterior_predictive["var_name"] = idata.posterior["var_name"]``.
+
+
+    The following code block explores how the behavior changes with different `var_names`:
+
+    .. code:: python
+
+        from logging import getLogger
+        import pymc as pm
+
+        # Some environments like google colab suppress
+        # the default logging output of PyMC
+        getLogger("pymc").setLevel("INFO")
+
+        kwargs = {"progressbar": False, "random_seed": 0}
+
+        with pm.Model() as model:
+            x = pm.Normal("x")
+            y = pm.Normal("y")
+            z = pm.Normal("z", x + y**2)
+            det = pm.Deterministic("det", pm.math.exp(z))
+            obs = pm.Normal("obs", det, 1, observed=[20])
+
+            idata = pm.sample(tune=10, draws=10, chains=2, **kwargs)
+
+    Default behavior. Generate samples of ``obs``, conditioned on the posterior samples of ``z`` found in the trace.
+    These are often referred to as posterior predictive samples in the literature:
+
+    .. code:: python
+
+        with model:
+            pm.sample_posterior_predictive(idata, var_names=["obs"], **kwargs)
+            # Sampling: [obs]
+
+    Re-compute the deterministic variable ``det``, conditioned on the posterior samples of ``z`` found in the trace:
+
+    .. code :: python
+
+          pm.sample_posterior_predictive(idata, var_names=["det"], **kwargs)
+          # Sampling: []
+
+    Generate samples of ``z`` and ``det``, conditioned on the posterior samples of ``x`` and ``y`` found in the trace.
+
+    .. code :: python
+
+        with model:
+            pm.sample_posterior_predictive(idata, var_names=["z", "det"], **kwargs)
+            # Sampling: [z]
+
+
+    Generate samples of ``y``, ``z`` and ``det``, conditioned on the posterior samples of ``x`` found in the trace.
+
+    Note: The samples of ``y`` are equivalent to its prior, since it does not depend on any other variables.
+    In contrast, the samples of ``z`` and ``det`` depend on the new samples of ``y`` and the posterior samples of
+    ``x`` found in the trace.
+
+    .. code :: python
+
+        with model:
+            pm.sample_posterior_predictive(idata, var_names=["y", "z", "det"], **kwargs)
+            # Sampling: [y, z]
+
+
+    Same as before, except ``z`` is not stored in the returned trace.
+    For computing ``det`` we still have to sample ``z`` as it depends on ``y``, which is also being sampled.
+
+    .. code :: python
+
+        with model:
+            pm.sample_posterior_predictive(idata, var_names=["y", "det"], **kwargs)
+            # Sampling: [y, z]
+
+    Every random variable is sampled. This is equivalent to calling :func:`~pymc.sample_prior_predictive`
+
+    .. code :: python
+
+        with model:
+            pm.sample_posterior_predictive(idata, var_names=["x", "y", "z", "det", "obs"], **kwargs)
+            # Sampling: [x, y, z, obs]
+
+
+    Note that "sampling" a :func:`~pymc.Deterministic` does not force random variables
+    that depend on this quantity to be sampled too. In the following example ``z`` will not
+    be resampled even though it depends on ``det_xy``:
+
+    .. code :: python
+
+        with pm.Model() as model:
+          x = pm.Normal("x")
+          y = pm.Normal("y")
+          det_xy = pm.Deterministic("det_xy", x + y**2)
+          z = pm.Normal("z", det_xy)
+          det_z = pm.Deterministic("det_z", pm.math.exp(z))
+          obs = pm.Normal("obs", det_z, 1, observed=[20])
+
+          idata = pm.sample(tune=10, draws=10, chains=2, **kwargs)
+
+          pm.sample_posterior_predictive(idata, var_names=["det_xy", "det_z"], **kwargs)
+          # Sampling: []
+
+
+    Controlling the number of samples
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    You can manipulate the InferenceData to control the number of samples
+
+    .. code:: python
+
+        import pymc as pm
+
+        with pm.Model() as model:
+            ...
+            idata = pm.sample()
+
+    Generate 1 posterior predictive sample for every 5 posterior samples.
 
     .. code:: python
 
         thinned_idata = idata.sel(draw=slice(None, None, 5))
         with model:
-            idata.extend(pymc.sample_posterior_predictive(thinned_idata))
+            idata.extend(pm.sample_posterior_predictive(thinned_idata))
 
-    Generate 5 posterior predictive samples per posterior sample.
+
+    Generate 5 posterior predictive samples for every posterior sample.
 
     .. code:: python
 
         expanded_data = idata.posterior.expand_dims(pred_id=5)
         with model:
-            idata.extend(pymc.sample_posterior_predictive(expanded_data))
+            idata.extend(pm.sample_posterior_predictive(expanded_data))
+
+
     """
 
     _trace: Union[MultiTrace, PointList]
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -666,9 +666,8 @@ def sample_posterior_predictive(
             # Sampling: [x, y, z, obs]
 
 
-    Note that "sampling" a :func:`~pymc.Deterministic` does not force random variables
-    that depend on this quantity to be sampled too. In the following example ``z`` will not
-    be resampled even though it depends on ``det_xy``:
+    .. danger:: Including a :func:`~pymc.Deterministic` in `var_names` may incorrectly force a random variable to be resampled, as happens with ``z`` in the following example:
+
 
     .. code :: python
 
@@ -683,7 +682,7 @@ def sample_posterior_predictive(
           idata = pm.sample(tune=10, draws=10, chains=2, **kwargs)
 
           pm.sample_posterior_predictive(idata, var_names=["det_xy", "det_z"], **kwargs)
-          # Sampling: []
+          # Sampling: [z]
 
 
     Controlling the number of samples


pymc/sampling/forward.py
```diff
@@ -497,22 +499,85 @@ def sample_posterior_predictive(
 
     Examples
     --------
-    Thin a sampled inferencedata by keeping 1 out of every 5 draws
-    before passing it to sample_posterior_predictive
+    Sample posterior predictive samples any variable included in `var_names`,
```
jessegrabowski: typo: for any variable

pymc/sampling/forward.py
```diff
@@ -497,22 +499,85 @@ def sample_posterior_predictive(
...
+    To ensure the samples are internally consistent, any random variable that depends
+    on another random variable that is being sampled is also sampled.
+    The final list of variables being sampled is shown in the log output.
+
+    Notice that if a random variable has no dependency on other random variables,
+    posterior predictive sampling is equivalent to prior sampling.
+    Likewise, if all random variables are being sampled, the behavior is equivalent to
+    that of sample_prior_predictive.
+
+    Notice how the behavior of the function changes with different `var_names` below:
```
jessegrabowski: suggested rephrasing: The following code block explores how the behavior changes with different `var_names`:

pymc/sampling/forward.py
```diff
@@ -497,22 +499,85 @@ def sample_posterior_predictive(
...
     --------
-    Thin a sampled inferencedata by keeping 1 out of every 5 draws
-    before passing it to sample_posterior_predictive
+    Sample posterior predictive samples any variable included in `var_names`,
+    conditioned on the values of other random variables found in the trace.
+    To ensure the samples are internally consistent, any random variable that depends
+    on another random variable that is being sampled is also sampled.
+    The final list of variables being sampled is shown in the log output.
+
+    Notice that if a random variable has no dependency on other random variables,
```
jessegrabowski: I'd put this in a :: warning box for the Sphinx api docs, and strengthen the language. Suggested:

**Important**: If a random variable does not depend on any other random variables, it will simply be sampled from itself -- that is, from the prior distributed defined in the model. **This function will _never_ result in a copy of a random variable's posterior distribution being returned**

pymc/sampling/forward.py
```diff
@@ -497,22 +499,85 @@ def sample_posterior_predictive(
...
+          x = pm.Normal("x")
+          y = pm.Normal("y")
+          z = pm.Normal("z", x + y**2)
+          det = pm.Deterministic("det", pm.math.exp(z))
+          obs = pm.Normal("obs", det, 1, observed=[20])
+
+          idata = pm.sample(tune=10, draws=10, chains=2, **kwargs)
+
+          # Default behavior
+          # Posterior predictive draws of `obs`, conditioned on the posterior draws of `z`
```
jessegrabowski: suggested rephrase: ...conditioned on the posterior draws of `z` found in `idata`

Same for other similar comments below

pymc/sampling/forward.py
```diff
@@ -497,22 +499,85 @@ def sample_posterior_predictive(
...
+          pm.sample_posterior_predictive(idata, var_names=["obs"], **kwargs)
+          # Sampling: [obs]
 
-    Generate 5 posterior predictive samples per posterior sample.
+          # Deterministic re-computation of `det` given the posterior draws of `z`
```
jessegrabowski: Suggested: Re-compute the deterministic variable `det` given...

pymc/sampling/forward.py
```diff
@@ -497,22 +499,85 @@ def sample_posterior_predictive(
...
+          pm.sample_posterior_predictive(idata, var_names=["det"], **kwargs)
+          # Sampling: []
+
+          # Posterior predictive draws of `z` and `det`, conditioned on the posterior draws of `x` and `y`
+          # This corresponds to the default behavior if `z` were an observed variable
```
jessegrabowski: This isn't the default behavior, because it includes `det`.

If you remove `det`, I'd suggest this to be: Posterior predictive draws of `z`. Because `z` depends on `x`, we obtain a valid posterior predictive. If `z` had been observed ,this would be the default behavior.

pymc/sampling/forward.py
```diff
@@ -497,22 +499,85 @@ def sample_posterior_predictive(
...
+          pm.sample_posterior_predictive(idata, var_names=["z", "det"], **kwargs)
+          # Sampling: [z]
+
+          # Posterior predictive draws of `y`, `z` and `det`, conditioned on the posterior draws of `x`
+          # Note that `y` samples are equivalent to the prior, since no other variables influence it.
+          pm.sample_posterior_predictive(idata, var_names=["y", "z", "det"], **kwargs)
```
jessegrabowski: I think it's important to have an example with `y` alone, and the note about priors. Including `z` and `det` here distracts from the point

pymc/sampling/forward.py
```diff
@@ -497,22 +499,100 @@ def sample_posterior_predictive(
...
+    You can manipulate the InferenceData to control the number of samples
 
     .. code:: python
+        import pymc as pm
+
+        with pm.Model() as model:
+            ...
+            idata = pm.sample()
+
+        # Generate 1 posterior predictive sample for every 5 posterior samples.
```
jessegrabowski: TBH this example could just be removed. It has nothing to do with the behavior of this function, and we don't recommend thinning anyway.

pymc/sampling/forward.py
```diff
@@ -497,22 +499,100 @@ def sample_posterior_predictive(
...
+          y = pm.Normal("y")
+          det_xy = pm.Deterministic("det_xy", x + y**2)
+          z = pm.Normal("z", x + y**2)
+          det_z = pm.Deterministic("det_z", pm.math.exp(z))
+          obs = pm.Normal("obs", det, 1, observed=[20])
+
+          idata = pm.sample(tune=10, draws=10, chains=2, **kwargs)
+
+          pm.sample_posterior_predictive(idata, var_names=["det_xy", det_z"], **kwargs)
+          Sampling: [z]
```
jessegrabowski: You said it doesn't force resample, but it's resampling?

pymc/sampling/forward.py
ricardoV94: This was an pre-existing example in the documentation, it's a different use case and hence in a different block. Users were indeed asking how to reduce the number of draws in posterior_predictive when we took an argument away that allowed to do it. 

The thinning recommendation is for posterior sampling diagnostics, it's still fine to thin for doing predictions, specially if you have memory concerns

pymc/sampling/forward.py
```diff
@@ -497,22 +499,100 @@ def sample_posterior_predictive(
...
+          y = pm.Normal("y")
+          det_xy = pm.Deterministic("det_xy", x + y**2)
+          z = pm.Normal("z", x + y**2)
+          det_z = pm.Deterministic("det_z", pm.math.exp(z))
+          obs = pm.Normal("obs", det, 1, observed=[20])
+
+          idata = pm.sample(tune=10, draws=10, chains=2, **kwargs)
+
+          pm.sample_posterior_predictive(idata, var_names=["det_xy", det_z"], **kwargs)
+          # Sampling: []
```
ricardoV94: This is not actually how it is behaving. It's sampling `z` in this case!

This is actually useful now as a work-around to this bug: https://github.com/pymc-devs/pymc/issues/6977, but I don't think it should be the final behavior once we sort it out?

 CC @lucianopaz

pymc/sampling/forward.py
ricardoV94: Not sold on it, but change my view

pymc/sampling/forward.py
ricardoV94: Yeah, I removed that sentence already, Just said that it doesn't matter that `z` is unobserved

pymc/sampling/forward.py
ricardoV94: See my comment below! I was expecting this to be the behavior but it's not.

pymc/sampling/forward.py
jessegrabowski: The interesting point about this case is that it results in draws from the prior of `y`. Issues related to the interdependence between `z` and `det` are treated in detail below. 

Basically there are two huge gotchas with this function:

1. You sometimes get back priors
2. You have to think about (potentially complex) inter-dependencies between RVs.

(2) is well treated in the examples that follow, but (1) is only given a passing mention. I think it deserves a "full spotlight" example, which could be this one.

ricardoV94: I am worried about too many examples. I don't think the fact that a posterior predictive sample of a "root" variable corresponds to the prior is particularly relevant. It's just a fun coincidence. I would rather focus on what happens to z, if you also sample x, y, both or neither. If you understand that logic, the behavior of y should kind of be obvious.

I don't like that when people toy with var_names and get confused, they conclude that all draws are from the prior. That's only so when you are sampling everything that comes before. Hence my focus on an intermediate variable `z` here.

As a counterpoint, I do mention the prior equivalence three times. Before the example, here as a comment and in the last example where I say it's equivalent to `sample_prior_predictive`

pymc/sampling/forward.py
ricardoV94: That's a great idea!

pymc/sampling/forward.py
jessegrabowski: OK that sells me on this example as-is. But I don't think anything about this should be assumed obvious ;)

pymc/sampling/forward.py
ricardoV94: I added below, instead of replacing the text. WDYT?

pymc/sampling/forward.py
ricardoV94: I said conditionally obvious, not obvious!

pymc/sampling/forward.py
jessegrabowski: Yes I like how you did it

jessegrabowski:
Last suggestion, getting into the weeds of personal preferences -- I think the API docs would look nicer if you added a title heading to each of the two examples. 

I also think the code in the new section would be easier to follow if you split the model definition plus each function call into a separate code block and wrote your commentary as text instead of code comments.

ricardoV94:
> Last suggestion, getting into the weeds of personal preferences -- I think the API docs would look nicer if you added a title (in bold or italics or w/e) to each of the two examples.
> 
> I also think the code in the new section would be easier to follow if you split the model definition plus each function call into a separate code block and wrote your commentary as text instead of code comments.

I agree, I just don't know if that's supported :D

jessegrabowski:
I think you can use restructuredtext subsubsection headings, like:

Example 1: Exploring the effect of `var_names` on outputs
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

See [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections). It's pretty hideous for viewing in an IDE, though.

ricardoV94:
That worked @jessegrabowski

pymc/sampling/forward.py
```diff
@@ -459,21 +459,23 @@ def sample_posterior_predictive(
...
+    var_names : Iterable[str], optional
         Names of variables for which to compute the posterior predictive samples.
+        By default, only observed variables are sampled.
+        See the example below for what happens when this argument is customized.
     sample_dims : list of str, optional
         Dimensions over which to loop and generate posterior predictive samples.
-        When `sample_dims` is ``None`` (default) both "chain" and "draw" are considered sample
-        dimensions. Only taken into account when `trace` is InferenceData or Dataset.
+        When ``sample_dims`` is ``None`` (default) both "chain" and "draw" are considered sample
+        dimensions. Only taken into account when ``trace`` is InferenceData or Dataset.
```
OriolAbril: references to arguments should have only one backtick, everything else we want formatted as code two according to numpydoc (and it is planned to use that to link/highlight between arguments in the future)

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+          obs = pm.Normal("obs", det, 1, observed=[20])
+
+          idata = pm.sample(tune=10, draws=10, chains=2, **kwargs)
+
+    Default behavior. Posterior predictive draws of ``obs``,
+    conditioned on the posterior draws of ``z`` found in the trace.
+
+    .. code:: python
+
+          pm.sample_posterior_predictive(idata, var_names=["obs"], **kwargs)
```
OriolAbril: Given this is called default behaviour, I would make it `pm.sample_posterior_predictive(idata)` nothing else.

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+          # Sampling: [obs]
+
+    Re-compute the deterministic variable ``det``,
+    conditioned on the posterior draws of ``z`` found in the trace.
+
+    .. code :: python
+
+          pm.sample_posterior_predictive(idata, var_names=["det"], **kwargs)
+          # Sampling: []
```
OriolAbril: with the warning above it's not clear what is happening here. It might be good to do `post_pred = pm.sample_posterior...` then add a 2nd comment line indicating which variables will be there

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+
+    Posterior predictive draws of ``z`` and ``det``,
```
OriolAbril: I would not call those posterior predictive draws. The most similar (and "sensible"?) case that comes to mind is predicting **parameters** for unseen groups in hierarchical models, which to me would be posterior samples. i.e. the alphas in https://mc-stan.org/docs/stan-users-guide/mixed-replication.html which don't have a name there either.

Happy to try and find/come up with a name for these kind of quantities but would avoid calling them posterior predictive (even if generated by `sample_posterior_predictive`) as this is one of the key points of all the discussion imo, `sample_posterior_predictive` doesn't necessarly generate posterior predictive samples

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+    conditioned on the posterior draws of ``x``` and ``y``` found in the trace.
+
+    Note: It does not matter that ``z``` was not an observed variable.
```
OriolAbril:
```suggestion
    Note: It does not matter that ``z`` was not an observed variable.
```

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+
+    Same as before, except ``z`` is not stored in the returned trace.
+    For computing ``det`` we still have to sample ``z`` as it depends on ``y``, which is also being sampled.
+
+    .. code :: python
+
+          pm.sample_posterior_predictive(idata, var_names=["y", "det"], **kwargs)
+          # Sampling: [y, z]
+
+    Every random variable is sampled. This is equivalent to calling :func:~`pymc.sample_prior_predictive`
```
OriolAbril:
```suggestion
    Every random variable is sampled. This is equivalent to calling :func:`~pymc.sample_prior_predictive`
```

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+
+    Note: The samples of ``y`` are equivalent to its prior, since it does not depend on any other variables.
+
+    .. code :: python
+
+          pm.sample_posterior_predictive(idata, var_names=["y", "z", "det"], **kwargs)
+          # Sampling: [y, z]
+
+
+    Same as before, except ``z`` is not stored in the returned trace.
```
OriolAbril, line 651: I was not aware of that! I would have guessed `z` is also in the trace. Does it require less memory or is it the same?

pymc/sampling/forward.py
ricardoV94: Afaict there is no name for this. I still don't buy the distinction, it's a prediction and it's based on the posterior. Whether you could ever observe this variable or not sounds a bit too philosophical.

OriolAbril: I started in astrophysics, where what we really care about are the parameters, the observations are little more than a means to an end; the end goal was finding the values of the parameters in the model and their uncertainty. To me calling these predictions is too similar to calling the posterior itself predictions which imo would be more confusing than anything. We can do a lot of things, observe some variables or some others, but even assuming any variable/parameter could be observed, our main tool is the model, which is based on considering some quantities parameters and other observations.

It might be a bit philosophical, but loosing sight of this would render the model useless; we can't always treat anything as a "variable", you need to treat some as parameters and others as observations. I'd even argue than in most cases we rely directly on this distinction: sampling needs the logp which needs to evaluate the different contributions, some as priors others as likelihood; model comparison either with elpd or bayes factors also relies on the distinction even if in different ways. The philosophical "parts" are what can tell us if what we are doing is right/wrong/what we really want to do.

Given this is documentation I believe we have to be extra careful, and try to get somewhere that allows power users to bend the model/rules as much as possible while still not being confusing to beginners. Otherwise I wouldn't really care

ricardoV94: > sampling needs the logp which needs to evaluate the different contributions, some as priors others as likelihood

The distinction doesn't matter for mcmc :)

I am happy to use another name if one can be suggested, but I think this really touches on the issue of behavior vs intention. I am more interested in explaining the behavior, because the intention will always depend on the user.

It's like if we worried (too much) about not confusing users in the docstrings of `pm.sample` when we show models without observed variables. Are they posterior or prior (+- predictive adjective) samples? But then we can philosophize that a prior is a posterior when data is empty.

pymc/sampling/forward.py
ricardoV94: It was not selected in `var_names` so it's not requested as an output?

pymc/sampling/forward.py
OriolAbril: > docstrings of pm.sample when we show models without observed variables. Are they posterior or prior (+- predictive adjective) samples?

I wasn't aware of this, but it can be fine if the focus is on behaviour and the result is not called "posterior".

We can rewrite the text here to avoid using "posterior predictive" maybe even give that quantity a name, and focusing on behaviour is helpful (and in my opinion also prevents us from using posterior predictive here as posterior predictive can't be disentangled with intent). We could use something like:

```
Generate draws from the distributions of ``z`` and ``det``,
conditioned on the draws of ``x``` and ``y``` found in `trace`.
```

I have also removed "posterior" and "the trace" to reference only `` `trace` `` aka the first argument of the function

pymc/sampling/forward.py
OriolAbril: I wasn't completely aware of the behaviour mix between select/sample of `var_names`. I took it to mean that "variables being sampled" were being sampled+returned. No need to change anything on the function regarding behaviour or docs on this though.

pymc/sampling/forward.py
ricardoV94: Sounds good

pymc/sampling/forward.py
AlexAndorra: det isn't observed either. Maybe this point should be raised above too

pymc/sampling/forward.py
AlexAndorra: same, wasn't aware of that

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+    .. code :: python
+
+          pm.sample_posterior_predictive(idata, var_names=["z", "det"], **kwargs)
+          # Sampling: [z]
+
+
+    Posterior predictive draws of ``y``, ``z`` and ``det``,
+    conditioned on the posterior draws of ``x`` found in the trace.
+
+    Note: The samples of ``y`` are equivalent to its prior, since it does not depend on any other variables.
```
AlexAndorra, line 640: that means the samples of z and det too are eqivalent to the prior, aren't they? because they depend on y

pymc/sampling/forward.py
AlexAndorra: agreed

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+
+    .. code :: python
+
+          pm.sample_posterior_predictive(idata, var_names=["y", "z", "det"], **kwargs)
+          # Sampling: [y, z]
+
+
+    Same as before, except ``z`` is not stored in the returned trace.
+    For computing ``det`` we still have to sample ``z`` as it depends on ``y``, which is also being sampled.
```
AlexAndorra, line 652: Same here, right: all the samples will be from the prior?

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+
+    Every random variable is sampled. This is equivalent to calling :func:~`pymc.sample_prior_predictive`
+
+    .. code :: python
+
+          pm.sample_posterior_predictive(idata, var_names=["x", "y", "z", "det", "obs"], **kwargs)
+          # Sampling: [x, y, z, obs]
+
+
+    Note that "sampling" a dependent deterministic does not force random variables to be sampled.
```
AlexAndorra: What do you mean? Is this referring to the example above or below?

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+
+    .. code :: python
+
+        with pm.Model() as model:
+          x = pm.Normal("x")
+          y = pm.Normal("y")
+          det_xy = pm.Deterministic("det_xy", x + y**2)
+          z = pm.Normal("z", x + y**2)
+          det_z = pm.Deterministic("det_z", pm.math.exp(z))
+          obs = pm.Normal("obs", det, 1, observed=[20])
```
AlexAndorra:
```suggestion
          obs = pm.Normal("obs", det_z, 1, observed=[20])
```

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+          idata = pm.sample(tune=10, draws=10, chains=2, **kwargs)
+
+          pm.sample_posterior_predictive(idata, var_names=["det_xy", "det_z"], **kwargs)
+          # Sampling: []
+
+
+    Controlling the number of draws
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    You can manipulate the InferenceData to control the number of samples
```
AlexAndorra, line 691: Really helpful examples! I didn't know how to do that since the dispearance of `n_draws`

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+    The function returns posterior predictive samples of any variable included in ``var_names``,
+    conditioned on the values of other random variables found in the trace.
+    To ensure the samples are internally consistent, any random variable that depends
+    on another random variable that is being sampled is itself sampled.
+    The final list of variables being sampled is shown in the log output.
+
+    Note that if a random variable has no dependency on other random variables,
+    posterior predictive sampling is equivalent to prior sampling.
+    Likewise, if all random variables are being sampled, the behavior is equivalent to
+    that of :func:`~pymc.sample_prior_predictive`.
```
AlexAndorra: Can we create a `UserWarning` to warn users when a a variable was sampled from the prior? I feel like this would help tremendously in disambiguating what `sample_posterior_predictive` is doing. Because right now it's doing it implicitly

AlexAndorra:
Thanks @ricardoV94 , these clarifications are very much welcome!

> I want to add two more examples first about the default ppc and predictions use-cases

Seems like default ppc is done and you only need predictions use-case?

pymc/sampling/forward.py
ricardoV94: I don't think users ever want (as in 99.9% of the time) an unobserved variable to be resampled regardless of whether that's equivalent to the prior or slightly different because it still depends on some non volatile variable.

I would rather go with my suggestion in https://github.com/pymc-devs/pymc/issues/7069#issuecomment-1878388607 which fails explicitly unless you manually add the variable to `sample_vars`

pymc/sampling/forward.py
ricardoV94: I think this helps understanding how the behavior changes when you start playing with it the kwarg

pymc/sampling/forward.py
ricardoV94: deterministics are never observed so I don't follow

pymc/sampling/forward.py
ricardoV94: No, they are different from the prior, because they are still conditioned on the posterior of x. It's their posterior predictive distribution ;)

Only for y is the posterior predictive distribution equivalent to the prior

pymc/sampling/forward.py
ricardoV94: No

pymc/sampling/forward.py
ricardoV94: Below, will make more clear

pymc/sampling/forward.py
AlexAndorra: > I would rather go with my suggestion in https://github.com/pymc-devs/pymc/issues/7069#issuecomment-1878388607 which fails explicitly unless you manually add the variable to sample_vars

That's fine by me too. My point is that this behavior should be explicit, to avoid surprising hidden behavior, so making it fail unless the user asks for it sounds great to me!

Bonus question though: what do you mean by "(non) volatile variable"? I've seen it mentioned several time on this PR, but I haven't seen it defined.

pymc/sampling/forward.py
AlexAndorra: Why do you precise that it doesn't matter that z was not observed?

pymc/sampling/forward.py
AlexAndorra: But z and det are a function of x and y. So that means z is a mixture of posterior samples of x and prior samples of y ?

pymc/sampling/forward.py
ricardoV94: Because you are calling a function whose name and default behavior suggests it only acts on observed variables. I was emphasizing that it doesn't matter if you are sampling observed or unobserved variables? It behaves the same

pymc/sampling/forward.py
ricardoV94: Yes. It's a mixture of posterior draws of x and posterior predictive draws of y. The fact posterior predictive draws of y are equivalent to prior is unimportant.

I emphasize this difference because people don't see the point of the function behavior because they focus too much on the fact sometimes this is the same as the prior. Which raises the question why would this function even have this functionality... seems useless.

But it's not. If you want to ask what z would like for new draws of y this function gives you that. In a PyMC model you may be interested in resampling a subsect of the variables for your forecasting problem, conditioned on everything else you learned.

pymc/sampling/forward.py
ricardoV94: > That's fine by me too. My point is that this behavior should be explicit, to avoid surprising hidden behavior, so making it fail unless the user asks for it sounds great to me!

Sure but I'm not tackling this in this PR because it will probably be a breaking change and require consensus. This PR only documents the existing behavior and as such can't be really rejected (besides whether the examples are correct/clear).

I also think this will help with the discussion because right now most people don't even know how it works

ricardoV94: > Bonus question though: what do you mean by "(non) volatile variable"? I've seen it mentioned several time on this PR, but I haven't seen it defined.

I forgot to define it. Volatile variables are variables that depend on variables being resampled (or coords/mutable data that have changed) and as such must also be resampled even if they existed in the trace and where not included in `var_names`, to ensure all returned draws are jointly consistent

AlexAndorra: > Sure but I'm not tackling this in this PR because it will probably be a breaking change and require consensus. This PR only documents the existing behavior and as such can't be really rejected (besides whether the examples are correct/clear).
> 
> I also think this will help with the discussion because right now most people don't even know how it works

💯 on this. I think we should merge this PR ASAP, to 1) make things clearer ASAP for users, 2) open that discussion for the breaking change and move towards solving it.
Do you need any help on this? Beside reviewing consistently.

Thanks for the definition of volatile variables. Much clearer now 👌

pymc/sampling/forward.py
AlexAndorra: > Yes. It's a mixture of posterior draws of x and posterior predictive draws of y.

Ok, then now that makes sense to me.

> If you want to ask what z would like for new draws of y this function gives you that. In a PyMC model you may be interested in resampling a subsect of the variables for your forecasting problem, conditioned on everything else you learned.

I agree with that. I think what's potentially confusing is it sounds like we're throwing away what we learned about `y` by drawing from its prior again. Making the behavior clearer, as we discussed above, will go a long way towards solving these confusions.

AlexAndorra, line 640:
```suggestion
    Note: The samples of ``y`` are equivalent to its prior, since it does not depend on any other variables. That also means that the samples of ``z`` and ``det`` are a mixture of posterior samples of ``x`` and prior samples of ``y``, since ``z`` and ``det`` are a function of ``x`` and ``y``.
```

pymc/sampling/forward.py
AlexAndorra: Yes, so my point is that comment still holds for `det`, since it's not observed either, so I suggest to also put that comment in the previous example. Will add a suggestion commit

pymc/sampling/forward.py
```diff
@@ -497,22 +499,148 @@ def sample_posterior_predictive(
...
+    Default behavior. Posterior predictive draws of ``obs``,
+    conditioned on the posterior draws of ``z`` found in the trace.
+
+    .. code:: python
+
+          pm.sample_posterior_predictive(idata, var_names=["obs"], **kwargs)
+          # Sampling: [obs]
+
+    Re-compute the deterministic variable ``det``,
+    conditioned on the posterior draws of ``z`` found in the trace.
```
AlexAndorra:
```suggestion
    conditioned on the posterior draws of ``z`` found in the trace.
    
    Note: It does not matter that ``det`` was not an observed variable.
```

pymc/sampling/forward.py
ricardoV94: I removed the note

pymc/sampling/forward.py
ricardoV94: Ah this is why it was failing, so it's actually working as expected

pymc/sampling/forward.py
ricardoV94: This was just a typo, it was referring to a `det` in the previous example

ricardoV94:
I have addressed several review comments:
* Do not call new draws posterior predictive draws, but add a note for the first case where it is equivalent
* Fix the Deterministics example

I also expanded with two new example sections:
* Posterior predictive checks and predictions: Shows the common use case
* Using new models: Shows how you can use a new model with new variables

pymc/sampling/forward.py
```diff
@@ -490,24 +503,215 @@ def sample_posterior_predictive(
...
+        import pymc as pm
+
+        with pm.Model(coords_mutable={"trial": [0, 1, 2]}) as model:
+            x = pm.MutableData("x", [-1, 0, 1], dims=["trial"])
+            beta = pm.Normal("beta")
+            noise = pm.HalfNormal("noise")
+            y = pm.Normal("y", mu=x * beta, sigma=noise, observed=[-2, 0, 3], dims=["trial"])
+
+            idata = pm.sample()
+            posterior_predictive = pm.sample_posterior_predictive(idata).posterior_predictive
```
AlexAndorra, line 526:
```suggestion
            # in-sample predictions
            posterior_predictive = pm.sample_posterior_predictive(idata).posterior_predictive
```

pymc/sampling/forward.py
```diff
@@ -490,24 +503,215 @@ def sample_posterior_predictive(
...
+
```
AlexAndorra, line 527:
```suggestion
# out-of-sample predictions
```

pymc/sampling/forward.py
```diff
@@ -490,24 +503,215 @@ def sample_posterior_predictive(
...
+        with pm.Model(coords_mutable={"trial": [3, 4]}) as distinct_predictions_model:
+            x = pm.MutableData("x", [-2, 2], dims=["trial"])
+            beta = pm.Normal("beta")
+            noise = pm.HalfNormal("noise")
+            extra_noise = pm.HalfNormal("extra_noise", sigma=noise)
+            y = pm.StudentT("y", nu=4, mu=x * beta, sigma=extra_noise, dims=["trial"])
+
+            predictions = pm.sample_posterior_predictive(idata, var_names=["y"], predictions=True).predictions
+
+
```
AlexAndorra, line 569: Really cool examples!
```suggestion
For more about out-of-model predictions, see this [blog post](https://www.pymc-labs.com/blog-posts/out-of-model-predictions-with-pymc/).

```

pymc/sampling/forward.py
```diff
@@ -490,24 +503,215 @@ def sample_posterior_predictive(
...
+    Every random variable is sampled. This is equivalent to calling :func:`~pymc.sample_prior_predictive`
+
+    .. code :: python
+
+        with model:
+            pm.sample_posterior_predictive(idata, var_names=["x", "y", "z", "det", "obs"], **kwargs)
+            # Sampling: [x, y, z, obs]
+
+
+    Note that "sampling" a :func:`~pymc.Deterministic` does not force dependent random variables to be sampled.
```
AlexAndorra:
```suggestion
    Note that "sampling" a :func:`~pymc.Deterministic` does not force random variables depending on this :func:`~pymc.Deterministic` to be sampled too.
```
Ambiguous otherwise. Also specify which RVs you'd expect to be sampled too if `det_xy` and `det_z` weren't `Deterministic`

AlexAndorra (changes requested):
Love it @ricardoV94 ! Just suggested a few changes, then good to merge

pymc/sampling/forward.py
ricardoV94: Done

ricardoV94:
@AlexAndorra thanks for the review and suggestions. I've addressed them

pymc/sampling/forward.py
```diff
@@ -490,24 +503,221 @@ def sample_posterior_predictive(
...
+            x = pm.MutableData("x", [-2, 2], dims=["trial"])
+            beta = pm.Normal("beta")
+            noise = pm.HalfNormal("noise")
+            extra_noise = pm.HalfNormal("extra_noise", sigma=noise)
+            y = pm.StudentT("y", nu=4, mu=x * beta, sigma=extra_noise, dims=["trial"])
+
+            predictions = pm.sample_posterior_predictive(idata, var_names=["y"], predictions=True).predictions
+
+
+    For more about out-of-model predictions, see this [blog post](https://www.pymc-labs.com/blog-posts/out-of-model-predictions-with-pymc/).
```
AlexAndorra: The link is not rendering properly for some reason, but I can't find how to add external links in Myst 🤔 
@OriolAbril ?

pymc/sampling/forward.py
```diff
@@ -490,24 +503,221 @@ def sample_posterior_predictive(
...
+    .. code :: python
+
+        with model:
+            pm.sample_posterior_predictive(idata, var_names=["x", "y", "z", "det", "obs"], **kwargs)
+            # Sampling: [x, y, z, obs]
+
+
+    Note that "sampling" a :func:`~pymc.Deterministic` does not force random variables
+    that depend on this quantity to be sampled too. In the following example ``z`` will not
+    be resampled even though it depends on ``det_xy``:
```
AlexAndorra: Now that we have to wait for Oriol's answer on the extrernal link before merging, let me add a nit 🙈 
I don't think it'll be super clear to all users that `z` depends on `det_xy`, because it looks like it only depends on `x` and `y `.
If you did this: `z = pm.Normal("z", det_xy)`, that'd probably be clearer. Not a big deal, just my 2 cents

AlexAndorra (approved):
All good, thanks @ricardoV94 🥳 
Just waiting for an answer to solve link formatting -- and took the opportunity to add a nit 😜

pymc/sampling/forward.py
ricardoV94: That's what I actually wanted to do! Wonder if the bug comes bug up... xD

AlexAndorra: Ah ok! Then that makes more sense

pymc/sampling/forward.py
OriolAbril: docstrings still have to be written in rst, not markdown. rst link syntax: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages

ricardoV94: Ah the worst aspect of rst syntax :D

That underscore at the end is the cherry on top of the cake xD

pymc/sampling/forward.py
ricardoV94: It shows the bug again! I added a separate commit that includes a danger section

ricardoV94: Opened an issue: https://github.com/pymc-devs/pymc/issues/7183

ricardoV94:
Things seem to be working. @AlexAndorra wanna do one last pass?

AlexAndorra (approved):
All good now @ricardoV94 🍾

PR #7023, opened 2023-11-21
Fix error with LKJCorr default transform

### LINKED ISSUE #7002: BUG: LKJCorr default transform raises error
### Describe the issue:

I am trying to run the model 4 from https://tomicapretto.github.io/posts/2022-06-12_lkj-prior/#model-4-correlated-priors-with-lkjcorr.-replicate-rstanarm-prior and I am getting the following error

```python
NotImplementedError: Univariate transform Interval cannot be applied to multivariate lkjcorr_rv{1, (0, 0), floatX, False}
```

```python
import arviz as az
import pytensor.tensor as pt
import matplotlib as mpl
import numpy as np
import pandas as pd
import pymc as pm

url = "https://raw.githubusercontent.com/vincentarelbundock/Rdatasets/master/csv/lme4/sleepstudy.csv"
data = pd.read_csv(url, index_col = 0)

# Subjects and subjects index
subjects, subjects_idx = np.unique(data["Subject"], return_inverse=True)

# Coordinates to handle dimensions of PyMC distributions and use better labels
coords = {"subject": subjects}

# Response mean -- Used in the prior for the intercept
y_mean = data["Reaction"].mean()

# Days variable
days = data["Days"].values

coords.update({"effect": ["intercept", "slope"]})

J = 2 # Order of covariance matrix

with pm.Model(coords=coords) as model_lkj_corr_2:
    # Common part
    β0 = pm.Normal("β0", mu=y_mean, sigma=50)
    β1 = pm.Normal("β1", mu=0, sigma=10)
    
    # Residual SD 
    σ = pm.HalfStudentT("σ", nu=4, sigma=50)
    
    # Group-specific part
    # Begin of rstanarm approach ----------------------------------
    τ = pm.Gamma("τ", alpha=1, beta=1)
    Σ_trace = J * τ ** 2
    π = pm.Dirichlet("π", a=np.ones(J), dims="effect")
    σ_u = pm.Deterministic("b_σ", σ * π * (Σ_trace) ** 0.5)
    # End of rstanarm approach ------------------------------------
    
    # Triangular upper part of the correlation matrix
    Ω_triu = pm.LKJCorr("Ω_triu", eta=1, n=2)
     
    # Correlation matrix
    Ω = pm.Deterministic(
        "Ω", pt.fill_diagonal(Ω_triu[np.zeros((2, 2), dtype=np.int64)], 1.)
    )

    # Construct diagonal matrix of standard deviation
    σ_u_diagonal = pm.Deterministic("σ_u_diagonal", pt.eye(2) * σ_u)
    
    # Covariance matrix
    Σ = pt.nlinalg.matrix_dot(σ_u_diagonal, Ω, σ_u_diagonal)
    
    # Cholesky decomposition, lower triangular matrix.
    L = pm.Deterministic("L", pt.slinalg.cholesky(Σ))
    u_raw = pm.Normal("u_raw", mu=0, sigma=1, dims=("effect", "subject")) 
    
    u = pm.Deterministic("u", pt.dot(L, u_raw).T, dims=("subject", "effect"))
                         
    u0 = pm.Deterministic("u0", u[:, 0], dims="subject")
    σ_u0 = pm.Deterministic("σ_u0", σ_u[0])
    
    u1 = pm.Deterministic("u1", u[:, 1], dims="subject")
    σ_u1 = pm.Deterministic("σ_u1", σ_u[1])
    
    # Correlation
    ρ_u = pm.Deterministic("ρ_u", Ω[0, 1])
         
    # Construct intercept and slope
    intercept = pm.Deterministic("intercept", β0 + u0[subjects_idx]) 
    slope = pm.Deterministic("slope", (β1 + u1[subjects_idx]) * days) 
    
    # Conditional mean
    μ = pm.Deterministic("μ", intercept + slope)
       
    y = pm.Normal("y", mu=μ, sigma=σ, observed=data["Reaction"])
```

### Error message:

```shell
NotImplementedError                       Traceback (most recent call last)
/Users/juanitorduz/Documents/website_projects/Python/multilevel_elasticities_single_sku_2.ipynb Cell 9 line 2
      1 with model_cov:
----> 2     idata_cov = pm.sample(
      3         target_accept=0.9,
      4         draws=4_000,
      5         chains=4,
      6         random_seed=rng,
      7     )
      8     posterior_predictive_cov = pm.sample_posterior_predictive(
      9         trace=idata_cov, random_seed=rng
     10     )

File ~/.local/share/virtualenvs/website_projects-1IZj_WTw/lib/python3.11/site-packages/pymc/sampling/mcmc.py:689, in sample(draws, tune, chains, cores, random_seed, progressbar, step, nuts_sampler, initvals, init, jitter_max_retries, n_init, trace, discard_tuned_samples, compute_convergence_checks, keep_warning_stat, return_inferencedata, idata_kwargs, nuts_sampler_kwargs, callback, mp_ctx, model, **kwargs)
    686         auto_nuts_init = False
    688 initial_points = None
--> 689 step = assign_step_methods(model, step, methods=pm.STEP_METHODS, step_kwargs=kwargs)
    691 if nuts_sampler != "pymc":
    692     if not isinstance(step, NUTS):

File ~/.local/share/virtualenvs/website_projects-1IZj_WTw/lib/python3.11/site-packages/pymc/sampling/mcmc.py:217, in assign_step_methods(model, step, methods, step_kwargs)
    215 methods_list: List[Type[BlockedStep]] = list(methods or pm.STEP_METHODS)
    216 selected_steps: Dict[Type[BlockedStep], List] = {}
--> 217 model_logp = model.logp()
    219 for var in model.value_vars:
    220     if var not in assigned_vars:
    221         # determine if a gradient can be computed

File ~/.local/share/virtualenvs/website_projects-1IZj_WTw/lib/python3.11/site-packages/pymc/model/core.py:721, in Model.logp(self, vars, jacobian, sum)
    719 rv_logps: List[TensorVariable] = []
    720 if rvs:
--> 721     rv_logps = transformed_conditional_logp(
    722         rvs=rvs,
    723         rvs_to_values=self.rvs_to_values,
    724         rvs_to_transforms=self.rvs_to_transforms,
    725         jacobian=jacobian,
    726     )
    727     assert isinstance(rv_logps, list)
    729 # Replace random variables by their value variables in potential terms

File ~/.local/share/virtualenvs/website_projects-1IZj_WTw/lib/python3.11/site-packages/pymc/logprob/basic.py:612, in transformed_conditional_logp(rvs, rvs_to_values, rvs_to_transforms, jacobian, **kwargs)
    609     transform_rewrite = TransformValuesRewrite(values_to_transforms)  # type: ignore
    611 kwargs.setdefault("warn_rvs", False)
--> 612 temp_logp_terms = conditional_logp(
...
   1269 else:
   1270     # Check there is no broadcasting between logp and jacobian
   1271     if logp.type.broadcastable != log_jac_det.type.broadcastable:

NotImplementedError: Univariate transform Interval cannot be applied to multivariate lkjcorr_rv{1, (0, 0), floatX, False}
```

### PyMC version information:

```
Last updated: Wed Nov 08 2023

Python implementation: CPython
Python version       : 3.11.5
IPython version      : 8.17.2

pytensor: 2.17.3

pandas    : 2.1.2
pytensor  : 2.17.3
numpy     : 1.23.5
matplotlib: 3.8.1
arviz     : 0.16.1
pymc      : 5.9.1

Watermark: 2.4.3
```

### Context for the issue:

_No response_

juanitorduz:
Closes https://github.com/pymc-devs/pymc/issues/7002

----
:books: Documentation preview :books:: https://pymc--7023.org.readthedocs.build/en/7023/

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,8 +9,8 @@
 ## Related Issue
 <!--- It is good practice to first open an issue explaining the bug / new feature that is addressed by this PR -->
 <!--- Please type an `x` in one of the boxes below and provide the issue number after the # sign: -->
-- [ ] Closes issue: #
-- [ ] Related issue (not closed by this PR): #
+- [ ] Closes #
+- [ ] Related to #
 
 ## Checklist
 <!--- Make sure you have completed the following steps before submitting your PR -->


juanitorduz:
@ricardoV94 I want to kick start this PR. In [242be23](https://github.com/pymc-devs/pymc/pull/7023/commits/242be2364c3f731ecdbfd941764d15230ae6d898) I simply removed the condition. I am still unsure where (which line) and how to sum the jabian. Should it be something like the statement above?

```python
diff_ndims = log_jac_det.ndim - logp.ndim
log_jac_det = log_jac_detsum(axis=np.arange(-diff_ndims, 0))
```
?

Or am I totally lost 🙈 ?

ricardoV94:
You're correct, and that should be inside the old elif statement, so don't remove it, just put that inside instead of raising. 

We should add a comment explaining when does that branch get triggered: univariate transform applied to a multivariate RV

pymc/logprob/transform_value.py
```diff
@@ -119,12 +119,6 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
         if log_jac_det.ndim < logp.ndim:
             diff_ndims = logp.ndim - log_jac_det.ndim
             logp = logp.sum(axis=np.arange(-diff_ndims, 0))
-        # This case is sometimes, but not always, trivial to accomodate depending on the "space rank" of the
-        # multivariate distribution. See https://proceedings.mlr.press/v130/radul21a.html
-        elif log_jac_det.ndim > logp.ndim:
-            raise NotImplementedError(
-                f"Univariate transform {transform} cannot be applied to multivariate {rv_op}"
-            )
         else:
```
ricardoV94: We should actually always run this last else check

juanitorduz: Fixed in https://github.com/pymc-devs/pymc/pull/7023/commits/0084e030bac14cc81cd6c7fa34956ca869d97bae

pymc/logprob/transform_value.py
```diff
@@ -119,19 +119,20 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
         if log_jac_det.ndim < logp.ndim:
             diff_ndims = logp.ndim - log_jac_det.ndim
             logp = logp.sum(axis=np.arange(-diff_ndims, 0))
+
+        # The following case gets triggered when univariate transform applied to a multivariate RV.
```
```suggestion
        # The following case gets triggered when a univariate transform is applied to a multivariate RV.
```

ricardoV94 (commented):
Looks good. We should add a test for the LKJCorr.

We should check that `model.compile_logp(sum=False)(model.initial_point())` has the same shape with and without the default transform. Before we introduced this restriction it didn't.

tests/distributions/test_transform.py
```diff
@@ -631,6 +631,12 @@ def test_univariate_transform_multivariate_dist_raises():
             m.logp(jacobian=jacobian)
 
 
+def test_log_jac_det_ndim_geq_logp_ndim():
```
juanitorduz: @ricardoV94 I must confess I am a bit lost. I pushed this dummy test (does nothing yet). What do you mean the "default transform" ? (i guess bounds between -1 and 1?). Also, how do I pass the identity (or None) to test the non-transformed?

juanitorduz: I also see the test `test_univariate_transform_multivariate_dist_raises` is failing because it is not raising enaything because of the change. Should I delete it or how should I modify it? Thanks 🙏

ricardoV94: You can change the old test and confirm the logp is the expected one, and has the right shape (before it would wrongly be a vector of length 3 instead of a scalar

ricardoV94: > Also, how do I pass the identity (or None) to test the non-transformed?

You can set `transform=None` when defining the LKJ to disable transform.

To test the default transform you can just not define `transform` or set it to `pymc.util.UNSET` which is the default value.

To make sure the model is what you expect it to be you can assert `m.rvs_to_transforms[rv]` is something or None depending on the case.

I would call this test something like LKJCorr default transform

juanitorduz: ok! Here is another attempt: https://github.com/pymc-devs/pymc/pull/7023/commits/c49bf6d4fd97997e8d66f2d71eda0625ac83ed0f

BTW: other small changes are due to Black 🙈  to make the pre-commit happy

juanitorduz:
@ricardoV94 all tests passed 🙌 !

tests/distributions/test_transform.py
```diff
@@ -522,7 +533,8 @@ def transform_params(*inputs):
         self.check_vectortransform_elementwise_logp(model)
 
     @pytest.mark.parametrize(
-        "mu,kappa,size", [(0.0, 1.0, (2,)), (floatX(np.zeros(3)), floatX(np.ones(3)), (4, 3))]
```
ricardoV94: You are introducing a lot of unrelated changes by adding a comma at the end of a function call that then makes black split across multiple lines

ricardoV94 (changes requested):
Should revert the unrelated changes. Your editor or pre-commit is probably more opinionated than the PyMC one

tests/distributions/test_transform.py
```diff
@@ -619,16 +631,26 @@ def test_transform_univariate_dist_logp_shape():
...
-            m.logp(jacobian=jacobian)
+
+@pytest.mark.parametrize(
+    argnames="jacobian", argvalues=[True, False], ids=["jacobian", "no_jacobian"]
+)
+@pytest.mark.parametrize(argnames="n", argvalues=[2, 3], ids=["n=2", "n=3"])
+def test_lkjcorr_default_transform(jacobian, n):
+    with pm.Model() as m:
+        pm.LKJCorr("Ω_triu", eta=1, n=n, transform=None)
+    assert m.compile_logp(jacobian=jacobian, sum=False)(m.initial_point())[0] == m.compile_logp(
```
ricardoV94: Not sure about the test.

Perhaps it's enough to test the default transform is Interval and that calling logp doesn't fail and has the right shape. This can be done in `tests/distributions/test_multivariate.py` as a specific regression test for the default transform of this Distribution (if this test had existed we would have had to address it when we did the breaking changes before)

tests/distributions/test_transform.py
```diff
@@ -619,16 +631,26 @@ def test_transform_univariate_dist_logp_shape():
...
 
 
-def test_univariate_transform_multivariate_dist_raises():
+@pytest.mark.parametrize(
+    argnames="jacobian", argvalues=[True, False], ids=["jacobian", "no_jacobian"]
+)
+def test_univariate_transform_multivariate_dist(jacobian):
     with pm.Model() as m:
         pm.Dirichlet("x", [1, 1, 1], transform=tr.log)
+    assert m.logp(jacobian=jacobian).type.shape == ()
```
ricardoV94: Let's add batch dims and test the logp is the expected one.

Something like `logp(dirichlet_dist, pt.log(test_value)) + LogTransform.log_jac_det(test_value).sum(-1)`

Also make sure it's not just -inf.

juanitorduz:
Dam! I erased the changes trying to revert 🤦 . I will try to bring them back

juanitorduz:
ok! seems we are back 😅

tests/distributions/test_transform.py
juanitorduz: yeah! I rever them (and created a mess in between but I think we are ok now)

tests/distributions/test_transform.py
juanitorduz: I am a bit lost. What would be the assert statement ?
I know ` assert np.all(pm.logp(x, pt.log(test_value)).eval() < np.inf)` ✅ , but what is the expected relationship with `tr.LogTransform().log_jac_det(test_value).sum(-1)` ?

tests/distributions/test_transform.py
juanitorduz: Hey! Thanks! Some questions:

> Perhaps it's enough to test the default transform is Interval 

Do you mean `m.compile_logp(jacobian=jacobian, sum=False)(m.initial_point())` is in certain interval? which one?

> calling logp doesn't fail and has the right shape.

Don't we have this already?

> This can be done in tests/distributions/test_multivariate.py as a specific regression test for the default transform of this Distribution

So you mean we move this test to a new one in `tests/distributions/test_multivariate.py`

PR #7026, opened 2023-11-23
Better float32 sampling support for TruncatedNormal

JasonTam:
When using float32 (via `PYTENSOR_FLAGS='floatX=float32'`), the current usage of `stats.truncnorm.rvs` in `rng_fn` can return `nan` for `TruncatedNormalRV` (and probably other functions that use a scipy flavored rvs). Reason is described in the issue linked below:

An issue was raised here (but will be closed without fix):
https://github.com/scipy/scipy/issues/19554 (and at a higher level, https://github.com/scipy/scipy/issues/15928 )

This PR is meant to just be a draft based on the suggested workaround and a discussion of how to best proceed. For now, I have personally just been using a custom patched distribution class with the changes in the PR.

----
:books: Documentation preview :books:: https://pymc--7026.org.readthedocs.build/en/7026/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -28,6 +28,7 @@ import numpy as np
 import pytensor
 import pytensor.tensor as pt
 
+from pytensor.configdefaults import config
 from pytensor.graph.basic import Apply, Variable
 from pytensor.graph.op import Op
 from pytensor.raise_op import Assert
@@ -569,6 +570,8 @@ class TruncatedNormalRV(RandomVariable):
     ) -> np.ndarray:
         # Upcast to float64 and then downcast if necessary
         #   (Work-around for https://github.com/scipy/scipy/issues/15928)
+        if cls.dtype == "floatX":
+            dtype = config.floatX
         return stats.truncnorm.rvs(
             a=((lower - mu) / sigma).astype("float64"),
             b=((upper - mu) / sigma).astype("float64"),
@@ -576,7 +579,7 @@ class TruncatedNormalRV(RandomVariable):
             scale=(sigma).astype("float64"),
             size=size,
             random_state=rng,
-        ).astype(mu.dtype)
+        ).astype(dtype)
 
 
 truncated_normal = TruncatedNormalRV()
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -567,15 +567,14 @@ class TruncatedNormalRV(RandomVariable):
         upper: Union[np.ndarray, float],
         size: Optional[Union[List[int], int]],
     ) -> np.ndarray:
-        a = (lower - mu) / sigma
-        b = (upper - mu) / sigma
-
-        dist = stats.truncnorm(a=a, b=b, loc=mu, scale=sigma)
-        # Underlying uniform should be of the same dtype as other inputs (`mu` for now)
-        ps = rng.random(size=size, dtype=mu.dtype)
-        ret = dist.ppf(ps)
-        # Enforce bounds b/c precision can violate
-        return ret.clip(lower, upper)
+        return stats.truncnorm.rvs(
+            a=((lower - mu) / sigma).astype("float64"),
+            b=((upper - mu) / sigma).astype("float64"),
+            loc=(mu).astype("float64"),
+            scale=(sigma).astype("float64"),
+            size=size,
+            random_state=rng,
+        ).astype(mu.dtype)
 
 
 truncated_normal = TruncatedNormalRV()
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -567,14 +567,15 @@ class TruncatedNormalRV(RandomVariable):
         upper: Union[np.ndarray, float],
         size: Optional[Union[List[int], int]],
     ) -> np.ndarray:
-        return stats.truncnorm.rvs(
-            a=(lower - mu) / sigma,
-            b=(upper - mu) / sigma,
-            loc=mu,
-            scale=sigma,
-            size=size,
-            random_state=rng,
-        )
+        a = (lower - mu) / sigma
+        b = (upper - mu) / sigma
+
+        dist = stats.truncnorm(a=a, b=b, loc=mu, scale=sigma)
+        # Underlying uniform should be of the same dtype as other inputs (`mu` for now)
+        ps = rng.random(size=size, dtype=mu.dtype)
+        ret = dist.ppf(ps)
+        # Enforce bounds b/c precision can violate
+        return ret.clip(lower, upper)
 
 
 truncated_normal = TruncatedNormalRV()
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -567,6 +567,8 @@ class TruncatedNormalRV(RandomVariable):
         upper: Union[np.ndarray, float],
         size: Optional[Union[List[int], int]],
     ) -> np.ndarray:
+        # Upcast to float64 and then downcast if necessary
+        #   (Work-around for https://github.com/scipy/scipy/issues/15928)
         return stats.truncnorm.rvs(
             a=((lower - mu) / sigma).astype("float64"),
             b=((upper - mu) / sigma).astype("float64"),
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -568,12 +568,8 @@ class TruncatedNormalRV(RandomVariable):
         upper: Union[np.ndarray, float],
         size: Optional[Union[List[int], int]],
     ) -> np.ndarray:
-        # Upcast to float64 and then downcast if necessary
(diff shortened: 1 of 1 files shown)


ricardoV94:
@JasonTam thanks for the investigation and write-up. I think the other suggestion in the linked thread is easier to maintain: upcast the parameters to float64 and downcast to the dtype of the RV afterwards.

JasonTam:
@ricardoV94 I've updated the PR to showcase the upcast/downcast method
Based on some quick tests, performance seems comparable between the methods. Casting seems to be a bit better
```python
with pm.Model() as model_cmp:
    x_control = TruncatedNormal("x_control", mu=1, sigma=3, lower=0, upper=10, size=100_000)
    x_manual_inv = CustomTruncatedNormal1("x_manual_inv", mu=1, sigma=3, lower=0, upper=10, size=100_000)
    x_cast_cast = CustomTruncatedNormal2("x_cast_cast", mu=1, sigma=3, lower=0, upper= 10, size=100_000)
---
%%timeit
x_control_samp = x_control.eval()
8.99 ms ± 61.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
---
%%timeit
x_manual_inv_samp = x_manual_inv.eval()
9.74 ms ± 83.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
_samp
---
%%timeit
x_cast_cast_samp = x_cast_cast.eval()
9.18 ms ± 130 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I'm just curious in terms of the pymc library, is this something that is worth fixing for all scipy RV's, or just mentioned in comments/docs, or maybe just having this PR here will be enough for posterity while https://github.com/scipy/scipy/issues/15928 is worked on.

ricardoV94:
No we shouldn't do anything special for all functions. This is a Scipy bug, but also seems to depend on how exactly the RV is being generated. If it was more widespread, it would be worth trying to press Scipy to fix it sooner.

ricardoV94:
calling `eval` is not a good way to benchmark, most of the time is consumed by compiling the function than actually evaluating it. You could simply time the python code directly. But RVs are not so time-sensitive like other aspects of the library. Otherwise we would not be using simple python code for them to begin with.

JasonTam:
Oh, I see

I tried timing the `rng_fn` themselves and saw very similar results anyway
```python
def rng_fn_control(
    rng: np.random.RandomState,
    mu: Union[np.ndarray, float],
    sigma: Union[np.ndarray, float],
    lower: Union[np.ndarray, float],
    upper: Union[np.ndarray, float],
    size: Optional[Union[List[int], int]],
) -> np.ndarray:
    return stats.truncnorm.rvs(
        a=(lower - mu) / sigma,
        b=(upper - mu) / sigma,
        loc=mu,
        scale=sigma,
        size=size,
        random_state=rng,
    )
    
def rng_fn_manual_inv(
    rng: np.random.RandomState,
    mu: Union[np.ndarray, float],
    sigma: Union[np.ndarray, float],
    lower: Union[np.ndarray, float],
    upper: Union[np.ndarray, float],
    size: Optional[Union[List[int], int]],
) -> np.ndarray:
    a = (lower - mu) / sigma
    b = (upper - mu) / sigma
​
    dist = stats.truncnorm(a=a, b=b, loc=mu, scale=sigma)
    # Underlying uniform should be of the same dtype as other inputs (`mu` for now)
    ps = rng.random(size=size, dtype=mu.dtype)
    return dist.ppf(ps).clip(lower, upper)
​
​
def rng_fn_cast_cast(
    rng: np.random.RandomState,
    mu: Union[np.ndarray, float],
    sigma: Union[np.ndarray, float],
    lower: Union[np.ndarray, float],
    upper: Union[np.ndarray, float],
    size: Optional[Union[List[int], int]],
) -> np.ndarray:
    return stats.truncnorm.rvs(
        a=((lower - mu) / sigma).astype('float64'),
        b=((upper - mu) / sigma).astype('float64'),
        loc=(mu).astype('float64'),
        scale=(sigma).astype('float64'),
        size=size,
        random_state=rng,
    ).astype(mu.dtype)
rng = np.random.default_rng(0)
mu, sigma, lower, upper = np.asarray([1, 3, 0, 10], dtype=np.float32)
size = 100_000
---
%%timeit
x_control_samp = rng_fn(rng, mu, sigma, lower, upper, size)
9.08 ms ± 48.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
---
%%timeit
x_manual_inv_samp = rng_fn_manual_inv(rng, mu, sigma, lower, upper, size)
9.87 ms ± 65.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
---
%%timeit
x_cast_cast_samp = rng_fn_cast_cast(rng, mu, sigma, lower, upper, size)
9.24 ms ± 98 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

pymc/distributions/continuous.py
```diff
@@ -568,13 +568,13 @@ def rng_fn(
...
-            loc=mu,
-            scale=sigma,
+            a=((lower - mu) / sigma).astype("float64"),
+            b=((upper - mu) / sigma).astype("float64"),
+            loc=(mu).astype("float64"),
+            scale=(sigma).astype("float64"),
             size=size,
             random_state=rng,
-        )
+        ).astype(mu.dtype)
```
ricardoV94:
```suggestion
        ).astype(self.dtype)
```

pymc/distributions/continuous.py
```diff
@@ -568,13 +568,13 @@ def rng_fn(
         size: Optional[Union[List[int], int]],
     ) -> np.ndarray:
         return stats.truncnorm.rvs(
-            a=(lower - mu) / sigma,
-            b=(upper - mu) / sigma,
-            loc=mu,
-            scale=sigma,
+            a=((lower - mu) / sigma).astype("float64"),
```
ricardoV94, line 573: Can you add an inline comment explaining why we do this and linking to the scipy issue?

pymc/distributions/continuous.py
JasonTam: Seems like `rng_fn` is a classmethod. I'm not sure if this would work(?). And `cls.dtype` would give the string literal `"floatX"`. Is it worth putting this conversion in here?
```python
from pytensor.configdefaults import config
...
if cls.dtype == "floatX":
    dtype = config.floatX
```

I admit `mu.dtype` is a bit ugly

pymc/distributions/continuous.py
JasonTam: Done in https://github.com/pymc-devs/pymc/pull/7026/commits/b1dabb856b43b6841f9a03398633a78ec502e3bc
lmk if it could be clearer / formatted differently

pymc/distributions/continuous.py
JasonTam: Implemented the mapped dtype in https://github.com/pymc-devs/pymc/pull/7026/commits/82228a903fc11965698ba8c62192b5bbed0ceba8 + https://github.com/pymc-devs/pymc/pull/7026/commits/e5fd34582baa4645bfd124afdc06b65182bd38b5

ricardoV94 (approved):

pymc/distributions/continuous.py
```diff
@@ -567,14 +568,20 @@ def rng_fn(
         upper: Union[np.ndarray, float],
         size: Optional[Union[List[int], int]],
     ) -> np.ndarray:
+        # Upcast to float64 and then downcast if necessary
+        #   (Work-around for https://github.com/scipy/scipy/issues/15928)
+        if cls.dtype == "floatX":
+            dtype = config.floatX
+        else:
```
ricardoV94: Hmm doesn't the perform method which calls this downcast already based on self.dtype? In that case we don't need this "hack" here. Specially because config.floatX could change after the RV is defined

ricardoV94 (changes requested):

pymc/distributions/continuous.py
JasonTam: I'm not sure I follow :confused: . I just see that `rng_fn` is a classmethod here. It will only have the `cls` attribute (?)

ricardoV94: This classmethod is always called by the baseclass perform method which already casts to the right dtype if needed: https://github.com/pymc-devs/pytensor/blob/7ecb9f8c6b6a2eac940947bac955a10785240667/pytensor/tensor/random/op.py#L342-346

So we shouldn't have to worry about it here, just upcast the inputs to float64

JasonTam: I see! I've removed the downcasting here https://github.com/pymc-devs/pymc/pull/7026/commits/93f9f3bd0e3320ab2162ec8b4ae833d931d58f69

ricardoV94:
Looks good to me, just need to run pre-commit. It's complaining about the unused import

ricardoV94 (approved):

twiecki (approved):

PR #7033, opened 2023-11-24
Remove several deprecated model properties and deprecate new ones

ricardoV94:
Sometimes less is more...

The `self.model` seems to have been introduced with ancient GLM / ModelBuilder ideas: https://github.com/pymc-devs/pymc/pull/1525

@lucianopaz  and I were surprised when we found it today. We thought we were getting a wrong object that wrapped model, but no, model does in fact have the most useless property ever :)

The deprecation of `pytensor_config` was discussed in https://github.com/pymc-devs/pymc/pull/5915 and I think it's as good time as any to deprecate it. 

@ferrine was the proponent of this functionality. This only seems relevant if you want different modes for different models, but I say power-users can just use `pytensor.config.change_flags` manually or build a helper that achieves the same functionality. The goal here is less magic. Note that this wouldn't work for nested models nor after cloning from fgraph as `do` and `observe` do.

Removed a bunch of methods that were deprecated sometime ago

----
:books: Documentation preview :books:: https://pymc--7033.org.readthedocs.build/en/7033/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -496,7 +496,13 @@ class Model(WithMemoization, metaclass=ContextMeta):
             instance._parent = kwargs.get("model")
         else:
             instance._parent = cls.get_context(error_if_none=False)
-        instance._pytensor_config = kwargs.get("pytensor_config", {})
+        pytensor_config = kwargs.get("pytensor_config", {})
+        if pytensor_config:
+            warnings.warn(
+                "pytensor_config is deprecated. Use pytensor.config or pytensor.config.change_flags context manager instead.",
+                FutureWarning,
+            )
+        instance._pytensor_config = pytensor_config
         return instance
 
     @staticmethod
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -559,6 +559,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
     @property
     def model(self):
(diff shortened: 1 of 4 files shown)


pymc/model/core.py
```diff
@@ -559,6 +565,7 @@ def __init__(
 
     @property
     def model(self):
+        warnings.warn("Model.model property is deprecated. Just use Model.", FutureWarning)
```
twiecki, line 568: 🤣

twiecki (approved):

michaelosthege (approved):

pymc/model/core.py
```diff
@@ -646,7 +653,7 @@ def compile_dlogp(
         jacobian:
             Whether to include jacobian terms in logprob graph. Defaults to True.
         """
-        return self.model.compile_fn(self.dlogp(vars=vars, jacobian=jacobian))
+        return self.compile_fn(self.dlogp(vars=vars, jacobian=jacobian))
```
ferrine, line 656: Should not it be root model instead?

ricardoV94: This is the previous behavior. I don't know what should be happening with compiled function of nested models. 

I think nested models are a mistake. The functionality we want to provide with them (plating) is much more narrow than what nested models do. We should probably use something like `OpFromGraph` for nested models, but that's another story.

PR #7034, opened 2023-11-26
Implement logcdf for TruncatedNormal

### LINKED ISSUE #7003: Implement logcdf for TruncatedNormal
### Description

We implemented the logcdf for the general Truncated cases in https://github.com/pymc-devs/pymc/pull/6690, but TruncatedNormal is its own distribution and doesn't have a logcdf currently implemented.

We should refactor the logcdf function added in that PR so that we can call it from the logcdf of TruncatedNormal as well as the general Truncated (no point in duplicating the code)

Reported in https://discourse.pymc.io/t/censoring-a-truncated-distribution/13261

LukeLB:
Closes #7003 first highlighted by a user on discord (https://discourse.pymc.io/t/censoring-a-truncated-distribution/13261) when trying to use the following model,
```python
with pm.Model() as m:
    mu = pm.Normal("mu", 0, 1)
    sigma = pm.Normal("sigma", 0, 1)
    normal_ = pm.Normal.dist(mu, sigma)
    trunc_normal = pm.Truncated.dist(normal_, lower=0)
    censored = pm.Censored("censored", trunc_normal, lower=None, upper=upper_bounds, observed=cy)
```
By refactoring the truncated `truncated_logcdf` to take a `base_rv_op` we can now call it from `TruncatedNormal`. I've also moved the `truncated_logcdf_from_base_rv` out of truncated.py and into dist_math.py as I was having curcular import errors.

**Checklist**

## New features
-TruncatedNormal logcdf

----
:books: Documentation preview :books:: https://pymc--7034.org.readthedocs.build/en/7034/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -85,6 +85,7 @@ from pymc.distributions.dist_math import (
     check_parameters,
     clipped_beta_rvs,
     i0e,
+    log_diff_normal_cdf,
     log_normal,
     logpow,
     normal_lccdf,
@@ -742,6 +743,31 @@ class TruncatedNormal(BoundedContinuous):
             )
 
         return logp
+
+    def logcdf(value, mu, sigma, lower, upper):
+        logcdf = log_diff_normal_cdf(mu, sigma, value, lower) - log_diff_normal_cdf(
+            mu, sigma, upper, lower
+        )
+
+        is_lower_bounded = not (
+            isinstance(lower, TensorConstant) and np.all(np.isneginf(lower.value))
+        )
+        is_upper_bounded = not (isinstance(upper, TensorConstant) and np.all(np.isinf(upper.value)))
+
+        if is_lower_bounded:
+            logcdf = pt.switch(value < lower, -np.inf, logcdf)
+
+        if is_upper_bounded:
+            logcdf = pt.switch(value <= upper, logcdf, 0.0)
+
+        if is_lower_bounded and is_upper_bounded:
+            logcdf = check_parameters(
+                logcdf,
+                pt.le(lower, upper),
+                msg="lower_bound <= upper_bound",
+            )
+
+        return logcdf
 
 
 @_default_transform.register(TruncatedNormal)
--- a/tests/distributions/test_continuous.py
+++ b/tests/distributions/test_continuous.py
@@ -37,6 +37,7 @@ from pymc.testing import (
     Circ,
     Domain,
     R,
+    Rminusbig,
     Rplus,
     Rplusbig,
     Rplusunif,
@@ -934,6 +935,11 @@ class TestMatchesScipy:
                 value, (lower - mu) / sigma, (upper - mu) / sigma, loc=mu, scale=sigma
             )
 
+        def scipy_logcdf(value, mu, sigma, lower, upper):
+            return st.truncnorm.logcdf(
+                value, (lower - mu) / sigma, (upper - mu) / sigma, loc=mu, scale=sigma
+            )
+
         check_logp(
             pm.TruncatedNormal,
             R,
@@ -957,6 +963,33 @@ class TestMatchesScipy:
             R,
             {"mu": R, "sigma": Rplusbig, "lower": -Rplusbig},
             ft.partial(scipy_logp, upper=np.inf),
+            decimal=select_by_precision(float64=6, float32=1),
+            skip_paramdomain_outside_edge_test=True,
+        )
+
+        check_logcdf(
+            pm.TruncatedNormal,
+            R,
+            {"mu": R, "sigma": Rplusbig, "lower": -Rplusbig, "upper": Rplusbig},
+            scipy_logcdf,
+            decimal=select_by_precision(float64=6, float32=1),
+            skip_paramdomain_outside_edge_test=True,
+        )
+
+        check_logcdf(
+            pm.TruncatedNormal,
+            R,
+            {"mu": R, "sigma": Rplusbig, "upper": Rplusbig},
+            ft.partial(scipy_logcdf, lower=-np.inf),
+            decimal=select_by_precision(float64=6, float32=1),
+            skip_paramdomain_outside_edge_test=True,
+        )
+
+        check_logcdf(
+            pm.TruncatedNormal,
+            R,
+            {"mu": R, "sigma": Rplusbig, "lower": -Rplusbig},
+            ft.partial(scipy_logcdf, upper=np.inf),
             decimal=select_by_precision(float64=6, float32=1),
             skip_paramdomain_outside_edge_test=True,
         )


pymc/distributions/dist_math.py
```diff
@@ -115,6 +117,45 @@ def std_cdf(x):
     return 0.5 + 0.5 * pt.erf(x / pt.sqrt(2.0))
 
 
+def truncated_logcdf_from_base_rv(base_rv_op, value, lower, upper, *rv_inputs, **kwargs):
```
ricardoV94: I think we don't need kwargs

```suggestion
def truncated_logcdf_from_base_rv(base_rv_op, value, lower, upper, *rv_inputs):
```

tests/distributions/test_truncated.py
```diff
@@ -400,6 +404,28 @@ def test_truncated_inference():
...
+        sigma = Uniform("sigma", 1, 10)
+        normal_ = Normal.dist(mu, sigma)
+        trunc_normal = Truncated.dist(normal_, lower=0)
+        censored = Censored("censored", trunc_normal, lower=None, upper=upper, observed=obs)
+
+    ip = m.initial_point()
+    norm_logp, trunc_logp, cens_logp = m.compile_logp(sum=False)(ip)
+    assert np.all(np.isfinite(norm_logp))
+    assert np.all(np.isfinite(trunc_logp))
+    assert np.all(np.isfinite(cens_logp))
```
ricardoV94: This test is a bit too indirect. We should just check the logcdf in the `TruncatedNormal` like we test other distributions logcdf in `tests/distributions/continuous.py`

ricardoV94 (commented):
Thanks! Just suggesting a more direct test

tests/distributions/test_truncated.py
LukeLB: Hey so I've tried this test instead, 
```python
from pymc.testing import (
    R,
    Rplus,
    Rminusbig,
    check_logcdf,
    select_by_precision,
)
import pymc as pm
def test_trunc_norm_logcdf():
        check_logcdf(
            pm.TruncatedNormal,
            R,
            {"mu": R, "sigma": Rplus, 'lower': Rminusbig, 'upper':Rplus},
            lambda value, mu, sigma, lower, upper: scipy.stats.truncnorm((lower - mu) / sigma, (upper - mu) / sigma, mu, sigma,).logcdf(value),
            decimal=select_by_precision(float64=5, float32=1),
        )
```
But I'm getting an assertion error:
`
AssertionError: 
Arrays are not almost equal to 5 decimals
{'mu': array(-2.1), 'sigma': array(0.01), 'lower': array(-1.), 'upper': array(1.5), 'value': array(0.01)}
x and y nan location mismatch:
 x: array(nan)
 y: array(-0.)
`
I'm not sure if this is just a factor of the test conditions I've set or whether there is actually something wrong/different with the logcdf implementation vs. scipy....

ricardoV94: That may be too extreme, mu is -2.1, and sigma is only 0.01, and we are testing logcdf at value 0.01.

Which implementation is returning nan, PyMC or Scipy?

LukeLB: PyMC is returning `nan` and scipy `-0` I believe. I had a look in `test_continuous` and realised there were a similar set of tests for `TruncatedNormal` logp which set `skip_paramdomain_outside_edge_test = True` and use different domains, so I have used the same params and the logcdf tests pass. I've pushed this so let me kow if you're happy with the new tests.

ricardoV94: Can you check where the nan is coming from by breaking apart the pytensor expression? I imagine there is a -inf - -inf or something, but not clear where from? Asking in case there is a specific edge case we can patch to avoid nan

LukeLB: I found it and fixed the bug, the test with the domains mentioned in the comment above now pass.

pymc/distributions/dist_math.py
```diff
@@ -115,6 +117,47 @@ def std_cdf(x):
...
+    elif is_upper_bounded:
+        lognorm = upper_logcdf
+
+    logcdf_numerator = logdiffexp(logcdf, lower_logcdf) if is_lower_bounded else logcdf
+    logcdf_trunc = pt.switch(
+        pt.isinf(logcdf_numerator) & pt.isinf(lognorm), 0.0, logcdf_numerator - lognorm
+    )
+
+    if is_lower_bounded:
+        logcdf_trunc = pt.switch(value < lower, -np.inf, logcdf_trunc)
```
LukeLB: This line is causing the test to fail, 
```
AssertionError: 
E           Arrays are not almost equal to 6 decimals
E           {'mu': array(-2.1), 'sigma': array(0.01), 'lower': array(-1.), 'upper': array(0.01), 'value': array(-1.)}
E           x and y -inf location mismatch:
E            x: array(0.)
E            y: array(-inf)
```
as `value` == `lower`. If I change to `<=` then the tests pass. Is there a reason why it is `<` and not `<=`?

ricardoV94: Perhaps for discrete truncated variables?

ricardoV94: What do you mean this line is causing it to fail? Sounds like you mean that that line could make it work, that's different. If you commented it out it would still fail no?

ricardoV94: Is the problem in the `logcdf_numerator` / `lognorm` already? I see you did something with those

LukeLB: Applogies I haven't been very clear, there seems to be two problems. With this test,
```python
check_logcdf(
        pm.TruncatedNormal,
        R,
        {"mu": R, "sigma": Rplus, 'lower': Rminusbig, 'upper':Rplus},
        lambda value, mu, sigma, lower, upper: scipy.stats.truncnorm((lower - mu) / sigma, (upper - mu) / sigma, mu, sigma,).logcdf(value),
        decimal=select_by_precision(float64=5, float32=1),
 )
```
which had the error,
```
AssertionError:  
Arrays are not almost equal to 5 decimals
 {'mu': array(-2.1), 'sigma': array(0.01), 'lower': array(-1.), 'upper': array(1.5), 'value': array(0.01)} x and y nan location mismatch:
x: array(nan) 
y: array(-0.)
```
This was caused because `logcdf_numerator` and `lognorm` were both `-inf` on the line,
```python
logcdf_trunc = logcdf_numerator - lognorm
```
So I changed to,
```python
  logcdf_trunc = pt.switch(
        pt.isinf(logcdf_numerator) & pt.isinf(lognorm), 0.0, logcdf_numerator - lognorm
    )
```

What I then found was that this test was also failing,
```python
check_logcdf(
            pm.TruncatedNormal,
            R,
            {"mu": R, "sigma": Rplus, "lower": Rminusbig, "upper": Rplus},
            ft.partial(scipy_logcdf, lower=-np.inf),
            decimal=select_by_precision(float64=6, float32=1),
            skip_paramdomain_outside_edge_test=True,
        )
```
which had the error
```
AssertionError: 
E           Arrays are not almost equal to 6 decimals
E           {'mu': array(-2.1), 'sigma': array(0.01), 'lower': array(-1.), 'upper': array(0.01), 'value': array(-1.)}
E           x and y -inf location mismatch:
E            x: array(0.)
E            y: array(-inf)
```
this led to my thinking in the comment above.

ricardoV94: Does it make sense that that `lognorm` was -inf?

ricardoV94: It does, it's a pretty extreme parametrization for the Truncated. What does scipy do exactly that returns 0?

Also in your check do we want to restrict to -inf? `pt.isinf` will work with both positive and negative infinity IIRC.

ricardoV94: scipy does some fancy tricks: https://github.com/scipy/scipy/blob/f16c3b1665d961bcbcc9e533ed8f45b54a6bf256/scipy/stats/_continuous_distns.py#L9708

Since we have a specialized TruncatedNormal and this one is probably the most used one, it may make sense to implement a specialized logcdf and leave the Truncated one alone. For instance I see we have this function, which may be just what we need to get some extra precision? https://github.com/pymc-devs/pymc/blob/02cbac6592506a547117b2daa3f05847d16fe063/pymc/distributions/dist_math.py#L137

LukeLB: OK sure that makes sense, I'll revert the truncated_logcdf function and write a new one specifically for the TruncatedNormal case.

LukeLB:
I've rewritten the function and it returns the correct answers but not at the precision we would like, it is correct to 3 decimal places but past that the tests fail. Any idea why this might be occuring?

ricardoV94:
> I've rewritten the function and it returns the correct answers but not at the precision we would like, it is correct to 3 decimal places but past that the tests fail. Any idea why this might be occuring?

We are still probably testing too extreme values, did it fix the nan thing at least?

LukeLB:
Yeah the nan problem is sorted, shall I just change the domains in the tests then? The domains in the `check_logp` tests seem to work fine, e.g.,
```python
check_logp(
    pm.TruncatedNormal,
    R,
    {"mu": R, "sigma": Rplusbig, "lower": -Rplusbig, "upper": Rplusbig},
    scipy_logp,
    decimal=select_by_precision(float64=6, float32=1),
    skip_paramdomain_outside_edge_test=True,
        )
```

ricardoV94:
Yup, sounds good

ricardoV94 (approved):
Looks great

ricardoV94:
It was a PyTensor issue that we already fixed, should work fine if we rerun them (I'm restarting them)

PR #7037, opened 2023-11-28
Adding add_coords function

zaxtax:
This adds the `add_coords` function so that coords can be updated without needing to explicitly mentioning a model

**Checklist**

## New features
- Add a new function

----
:books: Documentation preview :books:: https://pymc--7037.org.readthedocs.build/en/7037/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -95,6 +95,7 @@ __all__ = [
     "modelcontext",
     "Deterministic",
     "Potential",
+    "add_coords",
     "set_data",
     "Point",
     "compile_fn",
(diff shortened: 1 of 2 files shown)


ricardoV94:
Would something that can also update existing coords be more useful? Perhaps named `set_coords`?

zaxtax:
> Would something that can also update existing coords be more useful? Perhaps named `set_coords`?

Maybe, but updating coords likely would require something like the tricky shape-checking logic in `set_data`. Is there a use-case you have in mind for `set_coords`?

twiecki (approved):

PR #7047, opened 2023-12-04
Make coords and data always mutable

### LINKED ISSUE #6972: Reassess `MutableData` and `coords_mutable`
We could make all coords mutable by default and replace mutable lengths by constants before doing any operations on a Model object. Same for MutableData. We could have a helper function `freeze_graph` that does that.

Note we already do something similar for jax based sampling since shared variables aren't really supported in that backend

ricardoV94:
Closes #6972 

This PR provides a new model transform that freezes RV dims that depend on coords as well as mutable data, for those worried about performance issues or incompatibilities with JAX dynamic shape limitations. I expect most users won't need this. The default C-backend doesn't really exploit static shapes. I believe the simpler API for users is a net win.

Note that JAX dynamic shape stuff is not relevant when using JAX samplers because we already replace any shared variables by constants anyways. It's only relevant when compiling PyTensor functions with mode="JAX"

The big picture here is that we define the most general model first, and later specialize if needed. Going from a model with constant shapes to another with different constant shapes is generally not possible because PyTensor eagerly computes static shape outputs for intermediate nodes, and rebuilding with different constant types is not always supported.

Starting with more general models could be quite helpful for producing predictive models automatically.

**Note:** If there's resistance, this PR can be narrowed down in scope to just remove the distinction between coords_mutable and coords, but still leave MutableData vs ConstantData


twiecki:
Love it.

twiecki (approved):

twiecki:
We need to adapt the pymc-examples (and maybe the NBs here too?).

ricardoV94:
> We need to adapt the pymc-examples (and maybe the NBs here too?).

Possibly

ricardoV94:
It shouldn't fail immediately, just issue a warning.

Only when we remove the kwarg will it fail hard

lhelleckes:
Hi @ricardoV94!
@michaelosthege and I found out that the latest changes create issues when using `pm.ConstantData` (or `pm.Data`) and setting a dtype explicitly. We don't understand why because `pytensor.shared` has no problem with the dtype argument.

Here is an example:

```python
with pm.Model():
    pm.Data("b", [True, False], dtype=bool)
```
```
  File "<stdin>", line 2, in <module>
  File "C:\Users\osthege\AppData\Local\mambaforge\envs\dibecs_6.13.0\lib\site-packages\pymc\data.py", line 420, in Data
    x = pytensor.shared(arr, name, **kwargs)
  File "C:\Users\osthege\AppData\Local\mambaforge\envs\dibecs_6.13.0\lib\site-packages\pytensor\compile\sharedvalue.py", line 202, in shared
    var = shared_constructor(
  File "C:\Users\osthege\AppData\Local\mambaforge\envs\dibecs_6.13.0\lib\functools.py", line 889, in wrapper
TypeError: tensor_constructor() got an unexpected keyword argument 'dtype'
```

Do you have any idea why this is happening? Thanks in advance for your help!

PR #7058, opened 2023-12-09
Add helper functions to calculate required M for hsgp

ferrine:
It is always a trouble to calculate required M for HSGP, adding helper functions to simplify that

**What is this PR about?**
...

**Checklist**

## Major / Breaking Changes
- ...

## New features
- Calculating HSGP `m` parameter for common kernels

## Bugfixes
- ...

## Documentation
- ...

## Maintenance
- ...

----
:books: Documentation preview :books:: https://pymc--7058.org.readthedocs.build/en/7058/

### DIFF
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -17,6 +17,7 @@ import warnings
 
 from collections import Counter
 from functools import reduce
+from math import ceil
 from operator import add, mul
 from typing import Any, Callable, List, Optional, Sequence, Union
 
@@ -564,6 +565,10 @@ class Stationary(Covariance):
     def power_spectral_density(self, omega: TensorLike) -> TensorVariable:
         raise NotImplementedError
 
+    @staticmethod
+    def required_num_eigenvectors(L: float, ls: float) -> int:
+        raise NotImplementedError
+
 
 class ExpQuad(Stationary):
     r"""
@@ -594,6 +599,28 @@ class ExpQuad(Stationary):
         c = pt.power(pt.sqrt(2.0 * np.pi), self.n_dims)
         exp = pt.exp(-0.5 * pt.dot(pt.square(omega), pt.square(ls)))
         return c * pt.prod(ls) * exp
+
+    @staticmethod
+    def required_num_eigenvectors(L: float, ls: float) -> int:
+        r"""Number of eigenvectors in Hilbert space that approximate the GP well.
+
+        .. math::
+
+            m \ge 1.75  \frac{L}{ls}
+
+        Parameters
+        ----------
+        L : float
+            Approximation bound
+        ls : float
+            lengthscale
+
+        Returns
+        -------
+        int
+            Number of eigenvectors
+        """
+        return ceil(1.75 * L / ls)
 
 
 class RatQuad(Stationary):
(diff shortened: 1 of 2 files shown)


pymc/gp/cov.py
```diff
@@ -595,6 +600,28 @@ def power_spectral_density(self, omega: TensorLike) -> TensorVariable:
...
+            Approximation bound
+        ls : float
+            lengthscale
+
+        Returns
+        -------
+        int
+            Number of eigenvectors
+        """
+        return ceil(1.75 * L / ls)
```
michaelosthege, line 623: Can you link the source of these numbers in the docstring?

michaelosthege (changes requested):

pymc/gp/cov.py
michaelosthege: Also, you can rebase on `main` to fix these ❌ tests. I merged Ricardos #7057

pymc/gp/cov.py
```diff
@@ -663,6 +690,28 @@ def power_spectral_density(self, omega: TensorLike) -> TensorVariable:
         pow = pt.power(5.0 + pt.dot(pt.square(omega), pt.square(ls)), -1 * D52)
         return (num / den) * pt.prod(ls) * pow
 
+    @staticmethod
+    def required_num_eigenvectors(L: float, ls: float) -> int:
+        r"""Number of eigenvectors in Hilbert space that approximate the GP well.
```
bwengals, line 695: Probably should add some more detail in the docstrings here?  Would be kind of mysterious to see this if you're not using HSGPs, or know how to choose L.  

Another idea (take it or leave it) would be to have this put into a single function in `hsgp_approx.py` that does this all in one go, something like

```python
def required_num_eigenvectors(cov_func, L, ls):
    if cov_func.__class__.__name__ == "ExpQuad":
        return ceil(1.75 * L / ls)
    elif ...:
        ...
```
which maybe is better because the docstring can be in one place and the logic and purpose is isolated in one spot instead of spread out.  I wonder if the recommendations may evolve too, since in the paper they didn't try for higher than 1D inputs.

fonnesbeck: Even more convenient would be to have this function called automatically if `m` is not passed to HSGP (along with a healthy warning message informing users what is going on).

bwengals: and `c` too (but maybe not in this PR)

pymc/gp/cov.py
```diff
@@ -595,6 +600,28 @@ def power_spectral_density(self, omega: TensorLike) -> TensorVariable:
         exp = pt.exp(-0.5 * pt.dot(pt.square(omega), pt.square(ls)))
         return c * pt.prod(ls) * exp
 
+    @staticmethod
+    def required_num_eigenvectors(L: float, ls: float) -> int:
+        r"""Number of eigenvectors in Hilbert space that approximate the GP well.
```
juanitorduz, line 605: Could we add the paper (and maybe an equation reference) to understand where to learn more about the factors (for us, the newbies?)

pymc/gp/cov.py
ferrine: Yeah, I'll do this

ricardoV94:
Marked as a draft given all requests for changes in the review. Would be great to get this one over the finish line!

PR #7062, opened 2023-12-11
Add KL Divergence helper

ferrine:
## Related Issue

- [ ] Closes issue: #
- [ ] Related issue (not closed by this PR): #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
:books: Documentation preview :books:: https://pymc--7062.org.readthedocs.build/en/7062/


lucianopaz (changes requested):
This looks very good @ferrine! One thing that you definitely need to do before you can merge this is to add a page about this to the documentation. Maybe make a folder for `logprob` and include a subfolder for the KL divergence, since it will start to grow as more divergences get added.

pymc/logprob/basic.py
```diff
@@ -632,6 +633,27 @@ def transformed_conditional_logp(
     return logp_terms_list
 
 
+def kl_div(
```
ricardoV94, line 636: What about being a bit more verbose? I don't think it would hurt more as I don't imagine it will be invoked many times in a workflow?

`kl_divergence`?

pymc/logprob/kl_divergence.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@_kl_div.register(pytensor.tensor.random.basic.NormalRV, pytensor.tensor.random.basic.NormalRV)
+def _normal_normal_kl(
+    q_dist: pytensor.tensor.random.basic.NormalRV,
+    p_dist: pytensor.tensor.random.basic.NormalRV,
+    q_inputs: List[TensorVariable],
+    p_inputs: List[TensorVariable],
+):
+    _, _, _, q_mu, q_sigma = q_inputs
```
ricardoV94: Should probably consider size like moment does?

pymc/logprob/basic.py
ricardoV94: Ah scipy uses this name. Then it's better like this

ferrine:
> This looks very good @ferrine! One thing that you definitely need to do before you can merge this is to add a page about this to the documentation. Maybe make a folder for `logprob` and include a subfolder for the KL divergence, since it will start to grow as more divergences get added.

I found that our `pm.Distribution` like `pm.Normal` can pass `isinstance(rv.owner.op, pm.Normal)` checks, I wonder if I can further rely on such functionality

ferrine:
I am also confused about the design choice since `moment` implementation is scattered across the codebase...

ricardoV94:
The base functionality should be in logprob, but the specific implementations should be in `Distributions`

> I am also confused about the design choice, since moment implementation is scattered across the codebase...

`moment` was implemented at a different time for a different purpose. It's not even really `moment` but `finite_logp_point`. It's never used for logprob stuff, just samplers.

> I found that our pm.Distribution like pm.Normal can pass isinstance(rv.owner.op, pm.Normal) checks, I wonder if I can further rely on such functionality

You can but it won't work for things that look like Distributions but are just helpers to create distributions like `pm.ZeroInflatedBinomial`, or `LKJCholeskyCov`. What should always be safe is the `rv_op` that's actually returned.

pymc/distributions/continuous.py
```diff
@@ -551,6 +551,23 @@ def icdf(value, mu, sigma):
         )
 
 
+@_kl_div.register(Normal, Normal)
```
ricardoV94: I would still put it in a new folder in `distributions` though

ferrine:
Moved kl into a private pymc.distribution._stats because these are functions that will be never used by anyone

ricardoV94:
I don't like the underscore, why note just `distributions/kl_div`?

ferrine:
@ricardoV94 can you please reiterate on the review? Did I miss something?

ricardoV94:
Tests are failing with import issue

ricardoV94:
How many pairs of distributions do we expect to actually be able to support?

ferrine:
> How many pairs of distributions do we expect to actually be able to support?

Many of them https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/kl_divergence

pymc/distributions/stats/kl_divergence.py
```diff
@@ -0,0 +1,35 @@
...
+def _normal_normal_kl(
+    q_dist: Normal,
+    p_dist: Normal,
+    q_inputs: list[pt.TensorVariable],
+    p_inputs: list[pt.TensorVariable],
+):
+    _, _, _, q_mu, q_sigma = q_inputs
+    _, _, _, p_mu, p_sigma = p_inputs
+    diff_log_scale = pt.log(q_sigma) - pt.log(p_sigma)
+    return (
```
ricardoV94, line 32: May want to broadcast to size, like we do with `moment`, if someone does `kl_div(pm.Normal.dist(shape=5), pm.Normal.dist(mu=1))`

ferrine:
Just did the rebase, anything we can add or change on top of that?

pymc/distributions/stats/kl_divergence.py
ricardoV94: This is still relevant, you're ignoring batch dimensions encoded in the size parameter

pymc/logprob/abstract.py
```diff
@@ -153,3 +155,29 @@ def __init__(self, scalar_op, *args, **kwargs):
...
 
 MeasurableVariable.register(MeasurableElemwise)
+
+
+@multipledispatch(Op, Op)
+def _kl_div(
+    q_dist: Op,
+    p_dist: Op,
+    q_inputs: List[TensorVariable],
+    p_inputs: List[TensorVariable],
```
ricardoV94, line 165: This is wrong, RVs have non tensor inputs as well
```suggestion
    q_inputs: List[Variable],
    p_inputs: List[Variable],
```

pymc/logprob/abstract.py
```diff
@@ -153,3 +155,29 @@ def __init__(self, scalar_op, *args, **kwargs):
...
+
+
+def _kl_div_helper(
+    q_rv: TensorVariable,
+    p_rv: TensorVariable,
+) -> TensorVariable:
+    kl = _kl_div(
+        q_rv.owner.op,
+        p_rv.owner.op,
+        q_inputs=q_rv.owner.inputs,
```
ricardoV94, line 179: Perhaps pass the node instead of inputs. Allows stuff like op.dist_params(node) and op.size_param(node) inside the dispatch functions. Not sure though

PR #7063, opened 2023-12-12
Allow batched scalar sigma in ZeroSumNormal

ricardoV94:
## Description

The check that existed before was meant to prevent non-scalar sigma across zero-sum axes, but there is no reason to prevent it across batch dimensions.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7063.org.readthedocs.build/en/7063/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2591,8 +2591,7 @@ class ZeroSumNormal(Distribution):
     sigma : tensor_like of float
         Scale parameter (sigma > 0).
         It's actually the standard deviation of the underlying, unconstrained Normal distribution.
-        Defaults to 1 if not specified.
-        For now, ``sigma`` has to be a scalar, to ensure the zero-sum constraint.
+        Defaults to 1 if not specified. ``sigma`` cannot have length > 1 across the zero-sum axes.
     n_zerosum_axes: int, defaults to 1
         Number of axes along which the zero-sum constraint is enforced, starting from the rightmost position.
         Defaults to 1, i.e the rightmost axis.
@@ -2606,8 +2605,7 @@ class ZeroSumNormal(Distribution):
 
     Warnings
     --------
-    ``sigma`` has to be a scalar, to ensure the zero-sum constraint.
-    The ability to specify a vector of ``sigma`` may be added in future versions.
+    Currently, ``sigma``cannot have length > 1 across the zero-sum axes to ensure the zero-sum constraint.
 
     ``n_zerosum_axes`` has to be > 0. If you want the behavior of ``n_zerosum_axes = 0``,
     just use ``pm.Normal``.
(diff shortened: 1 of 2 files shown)


pymc/distributions/multivariate.py
```diff
@@ -2669,8 +2667,8 @@ def dist(cls, sigma=1, n_zerosum_axes=None, support_shape=None, **kwargs):
         n_zerosum_axes = cls.check_zerosum_axes(n_zerosum_axes)
 
         sigma = pt.as_tensor_variable(floatX(sigma))
-        if sigma.ndim > 0:
-            raise ValueError("sigma has to be a scalar")
+        if not all(sigma.type.broadcastable[-n_zerosum_axes:]):
+            raise NotImplementedError("sigma must have length one across the zerosumaxes")
```
lucianopaz: I would still raise a `ValueError` here instead of a `NotImplementedError`

pymc/distributions/multivariate.py
```diff
@@ -2706,7 +2702,12 @@ def check_zerosum_axes(cls, n_zerosum_axes: Optional[int]) -> int:
 
     @classmethod
     def rv_op(cls, sigma, n_zerosum_axes, support_shape, size=None):
-        shape = to_tuple(size) + tuple(support_shape)
+        if size is not None:
+            shape = tuple(size) + tuple(support_shape)
```
lucianopaz, line 2706: Do we need to check that `size` and `sigma.shape` are compatible or that done somewhere else before this call?

tests/distributions/test_multivariate.py
```diff
@@ -1704,6 +1704,44 @@ def test_does_not_upcast_to_float64(self):
...
+        )
+        assert pm.draw(batch_zsn).shape == (5, 3, 2)
+        np.testing.assert_allclose(
+            pm.logp(batch_zsn, batch_test_value).eval(),
+            expected_logp,
+        )
+
+        # Test NotImplementedError
+        with pytest.raises(
+            NotImplementedError, match="sigma must have length one across the zerosumaxes"
```
lucianopaz: If you accept my suggestion of using `ValueError`, you'll have to change the type here as well.

lucianopaz (approved):
Looks good to me. I just left one nitpick and one question in the code

pymc/distributions/multivariate.py
ricardoV94: We shouldn't have to check. Shape issues will arise when creating `Normal` below or at evaluation time when broadcasting fails between size and sigma

pymc/distributions/multivariate.py
ricardoV94: I just did this because it's a promised feature in the docstrings with the "currently"

ricardoV94: I like `ValueError` more though. Changed.

PR #7091, opened 2024-01-08
Ruff linter

### LINKED ISSUE #7084: Ruff linter?
### Description

Are we interested in adopting [`ruff`](https://github.com/astral-sh/ruff) ? There is some work on PyTenor already: https://github.com/pymc-devs/pytensor/pull/539

juanitorduz:
Closes https://github.com/pymc-devs/pymc/issues/7084

I tried adding minimal changes to keep up the default ruff configuration (see https://github.com/astral-sh/ruff?tab=readme-ov-file#configuration). Please review carefully 🤓 !

## Type of change

- [x] Maintenance

----
📚 Documentation preview 📚: https://pymc--7091.org.readthedocs.build/en/7091/


pymc/backends/arviz.py
```diff
@@ -240,7 +241,9 @@ def __init__(
 
         self.observations = find_observations(self.model)
 
-    def split_trace(self) -> Tuple[Union[None, "MultiTrace"], Union[None, "MultiTrace"]]:
+    def split_trace(
```
ricardoV94: Was the line length limit changed, which causes this and other lines to now split? Can that be set to the old value?

pymc/model/core.py
```diff
@@ -1740,7 +1738,7 @@ def debug_parameters(rv):
             done = {}
             used_ids = {}
             for i, out in enumerate(rv_inputs.maker.fgraph.outputs):
-                print_(f"{i}: ", end=""),
+                (print_(f"{i}: ", end=""),)
```
ricardoV94: I think this was a bug

```suggestion
                print_(f"{i}: ", end="")
```

pymc/backends/arviz.py
juanitorduz: I did keep `line-length = 100` and it is not the `ruff` but the `ruff formater`. I am following the recommended configuration. i will investgate.

ricardoV94: Then may be fine. Changes otherwise look good!

ricardoV94 (approved):

pymc/backends/arviz.py
juanitorduz: Let's see if https://github.com/pymc-devs/pymc/pull/7091/commits/bf843a4dfef3df67fce106b59bdf4bc3cf26850b works 🤞

juanitorduz: It worked! Do you want me to revert any other line?

ricardoV94: > It worked! Do you want me to revert any other line?

Yeah that would be great

juanitorduz:
@ricardoV94 Here are some additional improvements after your first review:
- Revert line-length changes ✅ 
- Remove all `pylint` mentions and comments (same for `isort`) ✅ 
- Stricter rules trying to avoid global ignore (we just have one now). "Solved" many via `#noqa` ✅ 
- Moved some parts of `setup.cfg` into `pyproject.toml`. I wanted to remove it all but I need to take care of `versioneeer` (see https://github.com/pymc-devs/pymc/pull/7091/commits/3130413219bcbf44213b28050f7c405d88dc8a2f). I will tackle this in another PR. 

The pre-commit job goes from 55s -> 23s (on average) after the `ruff` change. Also, the project dependencies decreased :)

pymc/logprob/__init__.py
```diff
@@ -42,7 +42,6 @@
     transformed_conditional_logp,
 )
 
-# isort: off
```
ricardoV94, line 45: How is import sorting controlled now?

juanitorduz: I think they are ignored now because of the lines in https://github.com/pymc-devs/pymc/pull/7091/commits/e6d3240c770cff1ca6610c068f3c1540b0c48492 (I just added the explanation).

Otherwise, I would need to fix and ignore one-by-one. Any preferences?

ricardoV94: Is there any isort at all left? That's pretty handy

juanitorduz: Yeah, you are right! Let me try to solve this so that we do not ignore that many rules on the _init_.py files 🙏

juanitorduz: Fixed! I did not include isort in the config! Now https://github.com/pymc-devs/pymc/pull/7091/commits/6819090f9d2e65d80b6a257d63ea8bdea4900689 failed on purpose and https://github.com/pymc-devs/pymc/pull/7091/commits/c58501bed6eec60ded25671e28e627897d6bdc14 fixes it!

juanitorduz:
i had forgotten to include `isort` in the config! Now is included!

.pre-commit-config.yaml
```diff
@@ -23,28 +23,18 @@ repos:
...
-- repo: https://github.com/PyCQA/isort
-  rev: 5.13.2
-  hooks:
-    - id: isort
-      name: isort
 - repo: https://github.com/asottile/pyupgrade
   rev: v3.15.0
   hooks:
     - id: pyupgrade
       args: [--py37-plus]
```
maresb:
```suggestion
```

pyproject.toml
```diff
@@ -1,21 +1,43 @@
...
-]
+xfail_strict = true
+addopts = ["--color=yes"]
 
-[tool.black]
+[tool.ruff]
 line-length = 100
 
+[tool.ruff.lint]
+select = ["E4", "E7", "E9", "F", "I"]
```
maresb:
```suggestion
select = ["E4", "E7", "E9", "F", "I", "UP"]
```

pyproject.toml
```diff
@@ -1,21 +1,43 @@
...
-xfail_strict=true
-addopts = [
-    "--color=yes",
-]
+xfail_strict = true
+addopts = ["--color=yes"]
 
-[tool.black]
+[tool.ruff]
 line-length = 100
```
maresb, line 9: This should be `"py39"`, but I don't think we want to tackle that in this PR.

```suggestion
target-version = "py37"

```

maresb (changes requested):
This is pretty exciting. We should be able to get rid of `pyupgrade` too.

maresb:
Similarly for `pydocstyle` with select `"D"`.

maresb:
The default rules in `ruff` are extremely minimalistic. I'd try selecting all of `E` and also `W` and check if you can make things stricter (possibly with exceptions) without blowing everything up.

It can be disappointing to have a previously-passing check that doesn't run, and then by the time you discover and fix it, it requires changing substantial parts of the codebase. (I've seen that happen with a malfunctioning line-length check, and then it's a battle to get that under control again.)

juanitorduz:
> The default rules in `ruff` are extremely minimalistic. I'd try selecting all of `E` and also `W` and check if you can make things stricter (possibly with exceptions) without blowing everything up.
> 
> It can be disappointing to have a previously-passing check that doesn't run, and then by the time you discover and fix it, it requires changing substantial parts of the codebase. (I've seen that happen with a malfunctioning line-length check, and then it's a battle to get that under control again.)

When I add `E` I get 252 errors of because of long single docstrings" (`E501`). Can we tackle this in another PR? Otherwise, this initial change can take ages 😅 .

juanitorduz:
@maresb Thanks for the feedback! I think addressed all your comments. Please let me know if I missed something 🙏

maresb (approved):
Looks excellent, great work!!!

ricardoV94:
Thanks a ton guys!

juanitorduz:
Thanks for your help! Please report any issue and I will try to solve it 💪

ricardoV94:
@juanitorduz I hear you are doing PyTensor next :D?

juanitorduz:
There is a PR open so I would not want to interfere 😅. I'll he happy to support though !

ricardoV94:
I think that one is stale, should be fine to take over (you can ask)

PR #7094, opened 2024-01-09
Report divergences from JAX samplers

### LINKED ISSUE #7041: Report divergences from JAX samplers
### Description

PyMC users are used to getting an immediate report when there are divergences in the standard NUTS sampler. I suggest we print (log) this information when we call the JAX samplers on behalf of the users.

This can be specially puzzling when there are nan in the gradient and the whole sampling finishes in an instant without a single warning, even though there were 100% divergences.

Recent user question: https://discourse.pymc.io/t/sampling-with-blackjax-no-divergence-report/13394

mohammed052:
## Description

PyMC users are used to getting an immediate report when there are divergences in the standard NUTS sampler. So I added a warning to be printed in case of divergence.

## Related Issue

- [x] Closes #7041

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7094.org.readthedocs.build/en/7094/

### DIFF
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -342,6 +342,8 @@ def _sample_external_nuts(
             coords=coords,
             dims=dims,
         )
+        warns = run_convergence_checks(idata, model)
+        log_warnings(warns)
         return idata
 
     elif sampler == "numpyro":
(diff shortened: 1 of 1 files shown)


ricardoV94:
@mohammed052 it seems like this issue was already being worked out in https://github.com/pymc-devs/pymc/pull/7051

In this case it should be fine for you to overtake it, but in general you don't want to do that as we will likely give priority to the open PRs. In any case the same comment I did here, applies to your changes: https://github.com/pymc-devs/pymc/pull/7051#issuecomment-1845537979

PR #7100, opened 2024-01-11
Add a flag to LKJCorr to return the unpacked correlation matrix

### LINKED ISSUE #7074: ENH: LKJCorr should return matrix rather than vector
### Before

`LKJCorr` returns an array corresponding to the off-diagonal elements of an upper-triangular matrix. 

```python
> pm.LKJCorr.dist(eta=1, n=3).eval()
array([ 0.06383647, -0.609944  , -0.13745336])
```

### After

`LKJCorr` should return a matrix with ones on the diagonal and the off-diagonal elements in their appropriate positions.

```python
> pm.LKJCorr.dist(eta=1, n=3).eval()
array([[ 1.        ,  0.06383647, -0.609944  ],
       [ 0.        ,  1.        , -0.13745336],
       [ 0.        ,  0.        ,  1.        ]])
```

### Context for the issue:

In the current implementations, the user is left to arrange this into a matrix themselves, which is cumbersome when there are more than one off-diagonal element. The pytensor `fill_diagonal` function only allows for scalar assignment, not array assignment as is the case with NumPy, so it must be filled by hand. Since almost every application will ultimately require the matrix, it should be returned to the user by default. This is also more consistent with `LKJCholeskyCov`, which returns an actual matrix and not just a flat array of values.

velochy:
## Description
Provide a wrapper around previous LKJCorr that reshapes the output to a full correllation matrix instead of the somewhat awkward vector of upper-triangular values. 
This is based on the discussion with @ricardoV94 and @jessegrabowski  in https://discourse.pymc.io/t/using-lkjcorr-together-with-mvnormal/13606/27 .

## Related Issue
- [x] Closes #7074

## Checklist

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7100.org.readthedocs.build/en/7100/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -1597,7 +1597,6 @@ def lkjcorr_default_transform(op, rv):
     return MultivariateIntervalTransform(floatX(-1.0), floatX(1.0))
 
 
-# Thin wrapper around _LKJCorr
 class LKJCorr:
     r"""
     The LKJ (Lewandowski, Kurowicka and Joe) log-likelihood.
@@ -1619,9 +1618,9 @@ class LKJCorr:
         implies a uniform distribution of the correlation matrices;
         larger values put more weight on matrices with few correlations.
     return_matrix : bool, default=False
-        If True, returns the full correllation matrix.
+        If True, returns the full correlation matrix.
         False only returns the values of the upper triangular matrix excluding
-        diagonal in a single vector of length n(n-1)/2 for backwards compatibility
+        diagonal in a single vector of length n(n-1)/2 for memory efficiency
 
     Notes
     -----
@@ -1641,13 +1640,17 @@ class LKJCorr:
                 'corr', eta=4, n=10, return_matrix=True
             )
 
-            # Define a new MvNormal with the given correllation matrix
+            # Define a new MvNormal with the given correlation matrix
             vals = sds*pm.MvNormal('vals', mu=np.zeros(10), cov=corr, shape=10)
 
             # Or transform an uncorrelated normal distribution:
             vals_raw = pm.Normal('vals_raw', shape=10)
             chol = pt.linalg.cholesky(corr)
             vals = sds*pt.dot(chol,vals_raw)
+
+            # The matrix is internally still sampled as a upper triangular vector
+            # If you want access to it in matrix form in the trace, add
+            pm.Deterministic('corr_mat', corr)
 
 
     References
@@ -1659,15 +1662,14 @@ class LKJCorr:
     """
 
     def __new__(cls, name, n, eta, *, return_matrix=False, **kwargs):
+        c_vec = _LKJCorr(name, eta=eta, n=n, **kwargs)
         if not return_matrix:
-            return _LKJCorr(name, eta=eta, n=n, **kwargs)
+            return c_vec
         else:
-            c_vec = _LKJCorr(name + "_raw", eta=eta, n=n, **kwargs)
-            return pm.Deterministic(name, cls.vec_to_corr_mat(c_vec, n))
+            return cls.vec_to_corr_mat(c_vec, n)
 
     @classmethod
-    def dist(cls, n, eta, *, return_matrix=True, **kwargs):
-        # compute Cholesky decomposition
+    def dist(cls, n, eta, *, return_matrix=False, **kwargs):
         c_vec = _LKJCorr.dist(eta=eta, n=n, **kwargs)
         if not return_matrix:
             return c_vec
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -1505,6 +1505,7 @@ class LKJCorrRV(RandomVariable):
     def _random_corr_matrix(cls, rng, n, eta, flat_size):
         # original implementation in R see:
         # https://github.com/rmcelreath/rethinking/blob/master/R/distributions.r
+
         beta = eta - 1.0 + n / 2.0
         r12 = 2.0 * stats.beta.rvs(a=beta, b=beta, size=flat_size, random_state=rng) - 1.0
         P = np.full((flat_size, n, n), np.eye(n))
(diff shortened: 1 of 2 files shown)


tests/distributions/test_multivariate.py
```diff
@@ -1469,6 +1470,20 @@ def test_with_cov_rv(
...
 
+    def test_with_lkjcorr_matrix(
+        self,
+    ):
+        with pm.Model() as model:
+            mu = pm.Normal("mu", 0.0, 1.0, shape=3)
+            sd = pm.Exponential("sd", 1.0, shape=3)
+
+            corr = pm.LKJCorr("corr", n=3, eta=2, return_matrix=True)
+            # pylint: enable=unpacking-non-sequence
```
```suggestion
```

tests/distributions/test_multivariate.py
```diff
@@ -1469,6 +1470,20 @@ def test_with_cov_rv(
...
+            mv = pm.MvNormal("mv", mu, cov=sd * (sd * corr).T, size=4)
+            prior = pm.sample_prior_predictive(samples=10, return_inferencedata=False)
+
+        assert prior["mv"].shape == (10, 4, 3)
```
ricardoV94: I would like to directly assert the values of `corr` are correct

pymc/distributions/multivariate.py
```diff
@@ -1631,11 +1592,95 @@ def logp(value, n, eta):
...
+        extended onion method." Journal of multivariate analysis,
+        100(9), pp.1989-2001.
+    """
+
+    def __new__(cls, name, n, eta, *, return_matrix=False, **kwargs):
+        if not return_matrix:
+            return _LKJCorr(name, eta=eta, n=n, **kwargs)
+        else:
+            c_vec = _LKJCorr(name + "_raw", eta=eta, n=n, **kwargs)
+            return pm.Deterministic(name, cls.vec_to_corr_mat(c_vec, n))
```
ricardoV94: Not sure about the name change and wrapping in a Deterministic by default, but not completely against it either

pymc/distributions/multivariate.py
```diff
@@ -1631,11 +1592,95 @@ def logp(value, n, eta):
         )
 
 
-@_default_transform.register(LKJCorr)
+@_default_transform.register(_LKJCorr)
 def lkjcorr_default_transform(op, rv):
     return MultivariateIntervalTransform(floatX(-1.0), floatX(1.0))
 
 
+# Thin wrapper around _LKJCorr
```
ricardoV94: Comments like this have a tendency to get left behind. I would exclude or add it inside the class definition
```suggestion
```

pymc/distributions/multivariate.py
ricardoV94: The argument for having the packed form is that it saves memory. This actually more than doubles it because we save both the packed form and the dense one. We could leave to the users to wrap in a Deterministic if they need to have it after sampling.

OTOH this is similar to what LKJ does by default now.

ricardoV94:
Thanks for picking this up @velochy!

pymc/distributions/multivariate.py
jessegrabowski: Maybe check the discussion #6828 about setting internal deterministics/coords?

Reminds me I need to finish that PR

pymc/distributions/multivariate.py
```diff
@@ -1631,11 +1592,98 @@ def logp(value, n, eta):
...
+    Parameters
+    ----------
+    n : tensor_like of int
+        Dimension of the covariance matrix (n > 1).
+    eta : tensor_like of float
+        The shape parameter (eta > 0) of the LKJ distribution. eta = 1
+        implies a uniform distribution of the correlation matrices;
+        larger values put more weight on matrices with few correlations.
+    return_matrix : bool, default=False
+        If True, returns the full correllation matrix.
```
twiecki:
```suggestion
        If True, returns the full correlation matrix.
```

velochy:
Made changes corresponding to all the comments so far.

pymc/distributions/multivariate.py
```diff
@@ -1631,11 +1592,98 @@ def logp(value, n, eta):
...
+    def __new__(cls, name, n, eta, *, return_matrix=False, **kwargs):
+        if not return_matrix:
+            return _LKJCorr(name, eta=eta, n=n, **kwargs)
+        else:
+            c_vec = _LKJCorr(name, eta=eta, n=n, **kwargs)
+            return cls.vec_to_corr_mat(c_vec, n)
+
+    @classmethod
+    def dist(cls, n, eta, *, return_matrix=True, **kwargs):
+        # compute Cholesky decomposition
```
```suggestion
```

pymc/distributions/multivariate.py
```diff
@@ -1631,11 +1592,98 @@ def logp(value, n, eta):
...
+    ----------
+    .. [LKJ2009] Lewandowski, D., Kurowicka, D. and Joe, H. (2009).
+        "Generating random correlation matrices based on vines and
+        extended onion method." Journal of multivariate analysis,
+        100(9), pp.1989-2001.
+    """
+
+    def __new__(cls, name, n, eta, *, return_matrix=False, **kwargs):
+        if not return_matrix:
+            return _LKJCorr(name, eta=eta, n=n, **kwargs)
```
ricardoV94: Can just call once and then branch in the return now that they're the same

pymc/distributions/multivariate.py
```diff
@@ -1631,11 +1592,98 @@ def logp(value, n, eta):
...
+    n : tensor_like of int
+        Dimension of the covariance matrix (n > 1).
+    eta : tensor_like of float
+        The shape parameter (eta > 0) of the LKJ distribution. eta = 1
+        implies a uniform distribution of the correlation matrices;
+        larger values put more weight on matrices with few correlations.
+    return_matrix : bool, default=False
+        If True, returns the full correlation matrix.
+        False only returns the values of the upper triangular matrix excluding
+        diagonal in a single vector of length n(n-1)/2 for backwards compatibility
```
```suggestion
        diagonal in a single vector of length n(n-1)/2 for memory efficiency
```

ricardoV94 (commented):
Looks pretty good! Left some minor suggestions

ricardoV94 (approved):
Thanks, this looks like a great improvement in UX

PR #7103, opened 2024-01-15
Make VI compatible with JAX backend

### LINKED ISSUE #7104: BUG: VI can't be used with Jax
### Describe the issue:

Fails due to missing jax support for VI

```python
pm.fit(
    fn_kwargs=dict(mode="JAX"),
)
```
```


### Error message:

_No response_

### PyMC version information:

5.10.3

### Context for the issue:

_No response_

### PR DESCRIPTION
**Pull request by ferrine**
## Related Issue

- [x] Closes #7104
- [x] Related to https://github.com/pymc-devs/pytensor/issues/595

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7103.org.readthedocs.build/en/7103/

### DIFF
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -22,7 +22,6 @@ import pytensor
 import pytensor.tensor as pt
 import scipy.sparse as sps
 
-from pytensor import scalar
 from pytensor.compile import Function, Mode, get_mode
 from pytensor.compile.builders import OpFromGraph
 from pytensor.gradient import grad
@@ -38,6 +37,7 @@ from pytensor.graph.basic import (
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.op import Op
 from pytensor.scalar.basic import Cast
+from pytensor.scalar.basic import identity as scalar_identity
 from pytensor.scan.op import Scan
 from pytensor.tensor.basic import _as_tensor_variable
 from pytensor.tensor.elemwise import Elemwise
@@ -412,28 +412,6 @@ def hessian_diag(f, vars=None, negate_output=True):
         return empty_gradient
 
 
-class IdentityOp(scalar.UnaryScalarOp):
-    @staticmethod
-    def st_impl(x):
-        return x
-
-    def impl(self, x):
-        return x
-
-    def grad(self, inp, grads):
-        return grads
-
-    def c_code(self, node, name, inp, out, sub):
-        return f"{out[0]} = {inp[0]};"
-
-    def __eq__(self, other):
-        return isinstance(self, type(other))
-
-    def __hash__(self):
-        return hash(type(self))
-
-
-scalar_identity = IdentityOp(scalar.upgrade_to_float, name="scalar_identity")
 identity = Elemwise(scalar_identity, name="identity")
 
 
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -37,9 +37,8 @@ from pytensor.graph.basic import (
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.op import Op
 from pytensor.scalar.basic import Cast
-from pytensor.scalar.basic import identity as scalar_identity
 from pytensor.scan.op import Scan
-from pytensor.tensor.basic import _as_tensor_variable
+from pytensor.tensor.basic import _as_tensor_variable, tensor_copy
 from pytensor.tensor.elemwise import Elemwise
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.type import RandomType
@@ -412,7 +411,7 @@ def hessian_diag(f, vars=None, negate_output=True):
         return empty_gradient
 
 
-identity = Elemwise(scalar_identity, name="identity")
+identity = tensor_copy
 
 
 def make_shared_replacements(point, vars, model):
--- a/pymc/sampling/jax.py
+++ b/pymc/sampling/jax.py
@@ -46,7 +46,6 @@ from pymc.backends.arviz import (
 from pymc.distributions.multivariate import PosDefMatrix
 from pymc.initial_point import StartDict
 from pymc.logprob.utils import CheckParameterValue
-from pymc.pytensorf import IdentityOp
 from pymc.sampling.mcmc import _init_jitter
 from pymc.stats.convergence import log_warnings, run_convergence_checks
 from pymc.util import (
@@ -68,14 +67,6 @@ __all__ = (
     "sample_blackjax_nuts",
     "sample_numpyro_nuts",
 )
-
-
-@jax_funcify.register(IdentityOp)
-def jax_funcify_Identity(op, **kwargs):
-    def identity_fn(value):
-        return value
-
-    return identity_fn
 
 
 @jax_funcify.register(Assert)
--- a/pymc/sampling/jax.py
+++ b/pymc/sampling/jax.py
@@ -46,6 +46,7 @@ from pymc.backends.arviz import (
 from pymc.distributions.multivariate import PosDefMatrix
 from pymc.initial_point import StartDict
 from pymc.logprob.utils import CheckParameterValue
+from pymc.pytensorf import IdentityOp
 from pymc.sampling.mcmc import _init_jitter
 from pymc.stats.convergence import log_warnings, run_convergence_checks
 from pymc.util import (
@@ -67,6 +68,14 @@ __all__ = (
     "sample_blackjax_nuts",
     "sample_numpyro_nuts",
 )
+
+
+@jax_funcify.register(IdentityOp)
+def jax_funcify_Identity(op, **kwargs):
+    def identity_fn(value):
+        return value
+
+    return identity_fn
 
 
 @jax_funcify.register(Assert)
--- a/tests/sampling/test_jax.py
+++ b/tests/sampling/test_jax.py
@@ -86,12 +86,17 @@ def test_jax_PosDefMatrix():
     [
         pytest.param(1),
         pytest.param(
-            2, marks=pytest.mark.skipif(len(jax.devices()) < 2, reason="not enough devices")
+            2,
+            marks=pytest.mark.skipif(
+                len(jax.devices()) < 2, reason="not enough devices"
+            ),
         ),
     ],
 )
 @pytest.mark.parametrize("postprocessing_vectorize", ["scan", "vmap"])
-def test_transform_samples(sampler, postprocessing_backend, chains, postprocessing_vectorize):
+def test_transform_samples(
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: pymc/sampling/jax.py [hunk 1]
```diff
@@ -67,6 +68,14 @@
 )
 
 
+@jax_funcify.register(IdentityOp)
```
ricardoV94: PyTensor has an identity Op already that gets automatically removed during compilation (so a dispatch wouldn't be needed). Can we use that instead and remove the PyMC one?

ricardoV94: https://github.com/pymc-devs/pytensor/blob/c4ae6e342facd1d4a1bb6b066f8a3b7efac1d3bc/pytensor/tensor/basic.py#L1921

Name is not great, so you can just import with an alias.

ferrine: yes, would be much better

pymc/pytensorf.py
```diff
@@ -47,6 +46,7 @@
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.op import Op
 from pytensor.scalar.basic import Cast
+from pytensor.scalar.basic import identity as scalar_identity
```
ricardoV94: You don't need to create a new Elemwise, there's already one defined in `tensor.math` (or basic), just called `tensor_copy`

pymc/pytensorf.py
```diff
@@ -387,7 +386,7 @@ def hessian_diag(f, vars=None):
         return empty_gradient
 
 
-identity = Elemwise(scalar_identity, name="identity")
+identity = tensor_copy
```
ricardoV94, line 414: Nitpick just import it directly in the VI module, no need to define it in pytensorf?

ferrine: It might be used by someone else I assume

ricardoV94: I don't think so, but even if we keep we should add a deprecatation warning

ferrine:
Windows tests seem to be very weird and can't reproduce it on a Linux machine, is shape inference platform dependent?

ricardoV94:
windows behaves differently with regard to integers. Default type is int32, which sometimes causes problems due to some rewrite or check that doesn't expect that (shape in PyTensor is supposed to be int64)

Just a guess from previous experiences. I can have a look on my windows machine next week

ferrine:
I see one of the issues got resolved with sort op recently. Any updates for Windows?

ricardoV94:
> Any updates for Windows?

I don't think anyone investigated the problem yet

ferrine:
How about marking these tests as fail then?

ricardoV94:
> How about marking these tests as xfail then?

Let me or someone investigate on a Windows machine. Seems like an important failure on Windows. In the meantime you can rebase and pin PyMC to the next PyTensor version to see if the current xfail can be removed?

ferrine:
@ricardoV94 updated the dependency on pytensor and commented on one of the xfails in the tests. Hope windows tests get  resolved with newer pytensor

ferrine:
In addition, mypy started to complain about pytensor

```
[pymc/sampling/forward.py]
pymc/sampling/forward.py:201: error: No overload variant of "general_toposort" matches argument types "list[Variable[Any, Any]]", "Callable[[Any], Any]"
pymc/sampling/forward.py:201: note: Possible overload variants:
pymc/sampling/forward.py:201: note:     def [T <: Node] general_toposort(outputs: Iterable[T], deps: None, compute_deps_cache: Callable[[T], Union[OrderedSet, list[T], None]], deps_cache: Optional[dict[T, list[T]]], clients: Optional[dict[T, list[T]]]) -> list[T]
pymc/sampling/forward.py:201: note:     def [T <: Node] general_toposort(outputs: Iterable[T], deps: Callable[[T], Union[OrderedSet, list[T]]], compute_deps_cache: None, deps_cache: None, clients: Optional[dict[T, list[T]]]) -> list[T]
```

ricardoV94:
@ferrine feel free to rebase, we have already bumped the dependency on main

tests/sampling/test_jax.py
```diff
@@ -503,3 +509,31 @@ def test_convergence_warnings(caplog, nuts_sampler):
 
     [record] = caplog.records
     assert re.match(r"There were \d+ divergences after tuning", record.message)
+
+
+@pytest.mark.parametrize("method", ["advi", "fullrank_advi"])
+def test_vi_sampling_jax(method):
+    with pm.Model() as model:
+        x = pm.Normal("x")
+        pm.fit(10, method=method, fn_kwargs=dict(mode="JAX"))
```
fonnesbeck, line 551: To be consistent with `pm.sample` and the `nuts_sampler=` arg, should we have a dedicated argument for the VI backend instead of kwargs?

twiecki: I vote yes, this API looks super weird.

ricardoV94: What looks weird? This is the compilation mode, would be exactly the same if you wanted to use Numba or JAX for the PyMC nuts sampler or for prior/posterior predictive.

The only thing I would change is the name of fn_kwargs, which is called compile_kwargs I think in those other functions

twiecki: Wouldn't this be what the user would have to do if they wanted to run VI on JAX?

ricardoV94: I don't understand the question, this PR is just doing minor tweaks so the PyMC VI module can compile to JAX. It's not linking to specific JAX VI libraries.

twiecki: You call it "VI module can compile to JAX", I call it "This PR adds an exciting feature that allows users to run VI on the much faster JAX backend wich also supports GPUs. Simply call `pm.fit(backend="jax")` to speed up your VI."

ricardoV94: I'm just saying this is not specific to VI nor JAX. Any function that compiles a Pytensor function allows JAX or Numba mode: draw, sample, prior/posterior/predictive, compile_logp...

I don't see why it should be a VI only argument. And my comment above is this is not equivalent to the `nuts_sampler` kwarg in `sample`.

If we want to add a backend kwarg as a shortcut for mode we would add everywhere for consistency.

twiecki: > If we want to add a backend kwarg as a shortcut for mode we would add everywhere for consistency.

Where else would you add it?

ricardoV94: All the functions mentioned above that already have a `compile_kwargs` equivalent to this `fn_kwargs` in VI. Neither kwarg is new.

Literally any function that compiles a Pytensor function should have the same kwargs API

ricardoV94: https://github.com/pymc-devs/pymc/blob/d8816c711df65974fe9785a877af1fccef2a1e14/pymc/sampling/forward.py#L368

twiecki: I'm fine adding it anywhere for consistency, but I mainly care about having it where a non-expert user would actually use it, like here.

The one other relevant case I can think of it would actually almost be detrimental: `pm.sample(backend="JAX")` where as a user would expect it just run everything on JAX (like with VI), which it wouldn't do because the sampler is still in Python (unless we made it call `nuts_sampler="numpyro"` under the hood).

I don't think the users should need to know about these finer points, they should just get a "make it fast with jax" button in the right place.

ricardoV94: We used this for sample_posterior_predictive for projects just last week, as we were sampling new variables that had heavy matmuls, went down from hours to minutes.

pm.sample is still useful as you can sample discrete variables with JAX this way.

twiecki: > We used this for sample_posterior_predictive for projects just last week, as we were sampling new variables that had heavy matmuls, went down from hours to minutes.

Great idea, should definitely add it there too.

> pm.sample is still useful as you can sample discrete variables with JAX this way.

That makes sense, I'm not opposed to adding it there. Maybe we can add a warning that the sampler is still running Python and they likely will want to use `nuts_sampler`.

ricardoV94: This is still doing python loops, it's exactly the same argument you need for pm.sample.

It's different than linking to a JAX VI library, which is what would be equivalent to the `nuts_sampler` kwarg that Chris mentioned in the first comment

twiecki: > This is still doing python loops, it's exactly the same argument you need for pm.sample.

Oh, I somehow assumed that VI was implemented mostly in PyTensor?

ferrine:
> @ferrine feel free to rebase, we have already bumped the dependency on main

Just rebased, let's see how it goes

tests/sampling/test_jax.py
ferrine: As for this, I'd prefer to focus this PR on backend compatibility and later address possible API changes in a new issue + PR. Agreed that there is inconsistency, we need to resolve that, but this will only defer the push to main with at least some working solution which went through many issues already.

ricardoV94: Agreed @ferrine. My only suggestion is to switch `fn_kwargs` to `compile_kwargs` which we use in the other sample methods

ferrine:
rebased the old PR to see if any issues got resolved

pymc/pytensorf.py
```diff
@@ -412,29 +411,7 @@ def hessian_diag(f, vars=None, negate_output=True):
...
-    def __eq__(self, other):
-        return isinstance(self, type(other))
-
-    def __hash__(self):
-        return hash(type(self))
-
-
-scalar_identity = IdentityOp(scalar.upgrade_to_float, name="scalar_identity")
-identity = Elemwise(scalar_identity, name="identity")
+identity = tensor_copy
```
ricardoV94, line 414: Just do `from pytensor... import tensor_copy as identity`?

tests/sampling/test_jax.py
```diff
@@ -536,3 +542,31 @@ def test_dirichlet_multinomial_dims():
...
+
+
+@pytest.mark.xfail(
+    reason="""
+During equilibrium rewriter this error happens. Probably one of the routines in SVGD is problematic.
+
+TypeError: The broadcast pattern of the output of scan
+(Matrix(float64, shape=(?, 1))) is inconsistent with the one provided in `output_info`
+(Vector(float64, shape=(?,))). The output on axis 0 is `True`, but it is `False` on axis
```
ricardoV94, line 560: This is actually something wrong, if the number of dimensions of a recurring output is different from the initial state. The difference between None and 1 is more annoying but this one looks like an error.

PR #7129, opened 2024-02-02
Upgrade to ruff 0.2.0 and fix RUF017

Armavica:
* --show-source -> --output-format=full
* --line-length removed from .pre-commit because it is in pyproject
* ruff.extend-per-file-ignores -> ruff.lint.extend-per-file-ignores
* fixed three instances of quadratic list summations

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7129.org.readthedocs.build/en/7129/


Armavica:
It looks like this test failure is unrelated to these changes, as it already occurs in `main`.

.pre-commit-config.yaml
```diff
@@ -24,12 +24,11 @@ repos:
...
         - --exclude=versioneer.py
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.1.14
+  rev: v0.2.0
   hooks:
     - id: ruff
-      args: ["--fix", "--show-source"]
+      args: ["--fix", "--output-format=full"]
     - id: ruff-format
-      args: ["--line-length=100"]
```
maresb, line 32: Note to self: this is redundant with the setting already in `pyproject.toml`.

Note to self 2: check the commit notes! :joy:

requirements-dev.txt
```diff
@@ -30,4 +30,4 @@ sphinx>=1.5
 sphinxext-rediraffe
 types-cachetools
 typing-extensions>=3.7.4
-watermark
\ No newline at end of file
+watermark
```
maresb: Note: Fixed missing line ending

maresb (approved):
All good! No changes required, just a few notes to myself below. Thanks @Armavica!

It's nice to fix those sums being used to concatenate lists (!!)

maresb:
Hmm...

```
FAILED tests/variational/test_inference.py::test_fit_start[SVGD-mini] - Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.
 Emitted warnings: [].
```

pymc/backends/report.py
```diff
@@ -34,7 +34,7 @@ def __init__(self) -> None:
 
     @property
     def _warnings(self):
-        chains = sum(self._chain_warnings.values(), [])
+        chains = [w for warnings in self._chain_warnings.values() for w in warnings]
```
ricardoV94: I don't find the two level list comprehension more readable than the sum concatenation tbh. Is this a default code style rule or personal preference?

ricardoV94:
> Hmm...
> 
> ```
> FAILED tests/variational/test_inference.py::test_fit_start[SVGD-mini] - Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.
>  Emitted warnings: [].
> ```

it's failing everywhere I'll check Monday

pymc/backends/report.py
ricardoV94: I prefer chain.from_iterable

Armavica: RUF017 suggests using `ft.reduce`, but the comprehension is faster. But it is indeed less readable, and `it.chain.from_iterable` looks like the best option all around: https://stackoverflow.com/questions/952914/how-do-i-make-a-flat-list-out-of-a-list-of-lists/953097#953097
I'll change everything.

pymc/logprob/scan.py
```diff
@@ -395,19 +395,14 @@ def find_measurable_scans(fgraph, node):
...
-                    # because they only ever have one output
-                    (cl.default_output(), i)
-                    for cl, _ in fgraph.get_clients(out)
-                    if isinstance(cl.op, Subtensor)
-                ]
-                for i, out in enumerate(node.outputs)
-            ],
-            [],
-        )
+        output_clients: list[tuple[Variable, int]] = [
```
Armavica, line 398: @ricardoV94 I changed the other two, but would you like me to change this one too? It's a bit different

pymc/backends/report.py
maresb: Is it really the double-comprehension that makes this code difficult to read?

Does using itertools actually make this easier to read?

I personally would have used something like: 

```python
warning_groups = self._chain_warnings.values()
chained_warnings = [warning for group in warning_groups for warning in group]
return chained_warnings + self._global_warnings
```

But it's all just style preference.

ricardoV94: double nested comprehension just to flatten is really odd for me. You are not filtering or doing anything with the items, which is the first thing I would be looking for.

The purpose of the operation is much more obvious with list(chain.from_iterable(iterable))

pymc/backends/report.py
```diff
@@ -13,6 +13,7 @@
 #   limitations under the License.
 
 import dataclasses
+import itertools as it
```
ricardoV94: just import chain, no need for the it alias?

pymc/logprob/scan.py
ricardoV94: This one seems fine. The sum was never needed if I'm decoding it correctly?

ricardoV94 (approved):

pymc/backends/report.py
Armavica: I will do it if you have a strong opinion about it, but I am a bit against extracting functions from their namespaces in general, I think it makes it more difficult to understand where they come from. Especially in the context of PyMC where "chain" can have a whole different meaning…

maresb: I think the bigger objection here is the alias. How about just `import itertools`?

I think the similarity is giving @ricardoV94 flashbacks to `at` :joy:

ricardoV94: Yes my problem is the alias

Armavica: I thought that was a standard alias. I'll remove it.

PR #7136, opened 2024-02-05
Fix VI test not skipping in Windows

ricardoV94:
## Description

The test `test_fit_start[SVGD-mini]` stopped raising a NotImplementedInferenceError in Windows recently. 

https://github.com/pymc-devs/pymc/blob/f8b19a52c19dcc77aaabd6dc7464bee9304ebc58/tests/variational/test_inference.py#L206-L210

It should come from https://github.com/pymc-devs/pymc/blob/2051d0b422a10ef0477cff0b3994ede4ec7c26ad/pymc/variational/operators.py#L145-L148

It still works correctly on Linux?

Link to failing test:
https://github.com/pymc-devs/pymc/actions/runs/7783900052/job/21223348962?pr=7136

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

- [ ] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Type of change

- [ ] New feature / enhancement
- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7136.org.readthedocs.build/en/7136/

PR #7144, opened 2024-02-07
Fix failing VI test due to pytest change

aerubanov:
## Description
In pytest 7 missing warning were ignored if some exception was raised, but pytest 8 checking them and thats why `test_fit_start[SVGD-mini]` fails. I replaced `pytest.warn()` by manual check for warning raised. 

## Related Issue

- [x] Closes #7136

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7144.org.readthedocs.build/en/7144/

### DIFF
--- a/tests/variational/test_inference.py
+++ b/tests/variational/test_inference.py
@@ -210,7 +210,7 @@ def test_fit_start(inference_spec, simple_model):
             pytest.skip(str(e))
 
     if expected_warning:
-        assert len(record) > 1
+        assert len(record) > 0
         for item in record:
             assert issubclass(item.category, UserWarning)
             assert "Could not extract data from symbolic observation" in str(item.message)
--- a/tests/variational/test_inference.py
+++ b/tests/variational/test_inference.py
@@ -197,13 +197,9 @@ def test_fit_start(inference_spec, simple_model):
 
     # Minibatch data can't be extracted into the `observed_data` group in the final InferenceData
     [observed_value] = [simple_model.rvs_to_values[obs] for obs in simple_model.observed_RVs]
-    if observed_value.name.startswith("minibatch"):
-        warn_ctxt = pytest.warns(
-            UserWarning, match="Could not extract data from symbolic observation"
-        )
-    else:
-        warn_ctxt = nullcontext()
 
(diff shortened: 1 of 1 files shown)


ferrine:
Many thanks investigating that!

tests/variational/test_inference.py
```diff
@@ -196,18 +197,27 @@ def test_fit_start(inference_spec, simple_model):
 
     # Minibatch data can't be extracted into the `observed_data` group in the final InferenceData
     [observed_value] = [simple_model.rvs_to_values[obs] for obs in simple_model.observed_RVs]
-    if observed_value.name.startswith("minibatch"):
```
ricardoV94, line 199: What about something like

```python
expected_warning: bool = ...

with warnings.catch_warns(record=True) as record:
  ...
  try:
    ...
  except NotImplementedInference as err:
    pytest.skip(...)
    
if expected_warning:
  assert len(record) == 1
  assert ... # Check expected warning
else:
  assert not record
```

We still want to raise for unexpected warnings

ricardoV94 (changes requested):

tests/variational/test_inference.py
aerubanov: @ricardoV94 I added check for unexpected warnings

tests/variational/test_inference.py
```diff
@@ -196,18 +197,26 @@ def test_fit_start(inference_spec, simple_model):
...
+        warnings.simplefilter("always")
+        try:
             trace = inference.fit(n=0).sample(10000)
-    except NotImplementedInference as e:
-        pytest.skip(str(e))
+        except NotImplementedInference as e:
+            pytest.skip(str(e))
+
+    if expected_warning:
+        assert len(record) > 1
```
ricardoV94: Why is it bigger than 1 and not just bigger than zero?

aerubanov: Ups, just my stupid mistake

ricardoV94 (approved):

ricardoV94:
Thanks a ton @aerubanov!

PR #7149, opened 2024-02-12
Implement `compute_log_prior` utility

aloctavodia:
Add `compute_log_prior` function. the log_prior values are needed for prior sensitivity checks as described in this paper https://arxiv.org/abs/2107.14054. Part of the functionality described in that paper is already implemented in ArviZ.

----
📚 Documentation preview 📚: https://pymc--7149.org.readthedocs.build/en/7149/


ricardoV94 (commented):
Can't this all be implemented in the same file?

aloctavodia:
sure,  we just need to rename `log_likelihood.py` to `log_density` or something similar and fix a few imports.

pymc/stats/log_density.py
```diff
@@ -54,20 +55,92 @@ def compute_log_likelihood(
...
+):
+    """
+    Compute elemwise log_likelihood or log_prior of model given InferenceData with posterior group
     """
 
     posterior = idata["posterior"]
 
     model = modelcontext(model)
 
+    if kind not in ("likelihood", "prior"):
```
ricardoV94, line 127: This utility will ignore the transforms+jacobians, as if the model had no transformed variables. Is that what you want?

aloctavodia: Yes. This is what I need.

ricardoV94 (approved):

PR #7153, opened 2024-02-13
Fix copying of shared variables in `fgraph_from_model`

### LINKED ISSUE #7105: `fgraph_from_model ` fails when cloning shared variables with specialized static shape
### Description

```python
import pymc as pm
from pymc.model.fgraph import fgraph_from_model

with pm.Model() as m:
    x = pm.MutableData("x", [0, 0, 0], shape=(3,))

fgraph_from_model(m)
```
```
AssertionError                            Traceback (most recent call last)
Cell In[19], line 8
      4 with pm.Model() as m:
      5     x = pm.MutableData("x", [0, 0, 0], shape=(3,))
----> 8 fgraph_from_model(m)

File ~/miniconda3/envs/colgate-shelf-sow2/lib/python3.11/site-packages/pymc/model/fgraph.py:204, in fgraph_from_model(model, inlined_views)
    201 else:
    202     new_var = shared(var.get_value(borrow=False))
--> 204 assert new_var.type == var.type
    205 new_var.name = var.name
    206 new_var.tag = copy(var.tag)

AssertionError: 
```

We need to pass the right kwargs (like shape), when recreating the shared variable. We should check if there's a general way to achieve this. Maybe a method should be added to Shared Variables in PyTensor

ricardoV94:
Closes #7105 

Also removes use of deprecated ScalarSharedVariable class https://github.com/pymc-devs/pytensor/pull/629

----
📚 Documentation preview 📚: https://pymc--7153.org.readthedocs.build/en/7153/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -44,7 +44,6 @@ from pytensor.scalar import Cast
 from pytensor.tensor.elemwise import Elemwise
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.type import RandomType
-from pytensor.tensor.sharedvar import ScalarSharedVariable
 from pytensor.tensor.variable import TensorConstant, TensorVariable
 from typing_extensions import Self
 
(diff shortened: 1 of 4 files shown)


AlexAndorra (approved):
Looks all good, thanks @ricardoV94 . Really interesting to read this code, helps to understand pytensor a bit more.
I'll test it on my usecase now

PR #7158, opened 2024-02-16
Add option to save model graph to an image

aloctavodia:
----
📚 Documentation preview 📚: https://pymc--7158.org.readthedocs.build/en/7158/

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -270,7 +270,7 @@ class TestData:
         error.match("The variable `beta` must be a `SharedVariable`")
 
     @pytest.mark.xfail(reason="Depends on ModelGraph")
-    def test_model_to_graphviz_for_model_with_data_container(self):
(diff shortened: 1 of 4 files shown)


tests/test_data.py
```diff
@@ -307,6 +309,11 @@ def test_model_to_graphviz_for_model_with_data_container(self):
             for expected in expected_substrings:
                 assert expected in g.source
 
+        pm.model_to_graphviz(model, save="model.png")
+        assert path.exists("model.png")
```
ricardoV94: Better not to create permanent files during tests.

ricardoV94 (changes requested):

tests/test_data.py
ricardoV94: https://docs.pytest.org/en/6.2.x/tmpdir.html

ricardoV94 (approved):

PR #7159, opened 2024-02-16
Add gufunc signature to SymbolicRandomVariables

ricardoV94:
## Description

This meta-info is necessary to reason about batch dims of SymbolicRandomVariables in the context of https://github.com/pymc-devs/pymc-experimental/pull/300

This is probably what we should use for RandomVariables, instead of defining `ndims_params` and `ndim_supp`. Those can be properties derived from the gufunc signature.

There is however a limitation with gufunc signatures, which has to do with inputs and outputs that are not tensors, such as RNGs and the size vector which obviously cannot have batch dimensions. For now I am treating those as scalars so in the signature they show up as `()`, but perhaps it makes sense to deviate a bit from numpy and use `[]` or `None`? 

For vanilla RandomVariables like Normal. the signature would be `None,None,None,(),()->None,()`, for the inputs: rng, size, dtype, mu, sigma and outputs: next_rng, draws.

## Related Issue

- [ ] Closes #
- [x] Related to https://github.com/pymc-devs/pymc-experimental/pull/300

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7159.org.readthedocs.build/en/7159/

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -33,11 +33,13 @@ from pytensor.graph.rewriting.basic import GraphRewriter, in2out
 from pytensor.graph.utils import MetaType
 from pytensor.scan.op import Scan
 from pytensor.tensor.basic import as_tensor_variable
+from pytensor.tensor.blockwise import safe_signature
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.rewriting import local_subtensor_rv_lift
 from pytensor.tensor.random.type import RandomGeneratorType, RandomType
 from pytensor.tensor.random.utils import normalize_size_param
 from pytensor.tensor.rewriting.shape import ShapeFeature
+from pytensor.tensor.utils import _parse_gufunc_signature
 from pytensor.tensor.variable import TensorVariable
 from typing_extensions import TypeAlias
 
@@ -261,6 +263,12 @@ class SymbolicRandomVariable(OpFromGraph):
     (0 for scalar, 1 for vector, ...)
      """
 
+    ndims_params: Optional[Sequence[int]] = None
+    """Number of core dimensions of the distribution's parameters."""
+
+    signature: str = None
+    """Numpy-like vectorized signature of the distribution."""
+
     inline_logprob: bool = False
     """Specifies whether the logprob function is derived automatically by introspection
     of the inner graph.
@@ -271,9 +279,25 @@ class SymbolicRandomVariable(OpFromGraph):
     _print_name: tuple[str, str] = ("Unknown", "\\operatorname{Unknown}")
     """Tuple of (name, latex name) used for for pretty-printing variables of this type"""
 
-    def __init__(self, *args, ndim_supp, **kwargs):
-        """Initialitze a SymbolicRandomVariable class."""
-        self.ndim_supp = ndim_supp
+    def __init__(
+        self,
+        *args,
+        **kwargs,
+    ):
+        """Initialize a SymbolicRandomVariable class."""
+        if self.signature is None:
+            self.signature = kwargs.get("signature", None)
+
+        if self.signature is not None:
+            inputs_sig, outputs_sig = _parse_gufunc_signature(self.signature)
+            self.ndims_params = [len(sig) for sig in inputs_sig]
+            self.ndim_supp = max(len(out_sig) for out_sig in outputs_sig)
+
+        if self.ndim_supp is None:
+            self.ndim_supp = kwargs.get("ndim_supp", None)
+            if self.ndim_supp is None:
+                raise ValueError("ndim_supp or gufunc_signature must be provided")
+
         kwargs.setdefault("inline", True)
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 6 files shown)


pymc/distributions/distribution.py
```diff
@@ -259,9 +267,25 @@ class SymbolicRandomVariable(OpFromGraph):
...
+        """Initialize a SymbolicRandomVariable class."""
+        if self.signature is None:
+            self.signature = kwargs.get("signature", None)
+
+        if self.signature is not None:
+            inputs_sig, outputs_sig = _parse_gufunc_signature(self.signature)
+            self.ndims_params = [len(sig) for sig in inputs_sig]
+            self.ndim_supp = max(len(out_sig) for out_sig in outputs_sig)
+
+        if self.ndim_supp is None:
```
zaxtax, line 296: Is `ndim_supp` going to be deprecated? If so should we include a warning?

pymc/distributions/distribution.py
```diff
@@ -594,7 +629,7 @@ def rv_op(
         random: Optional[Callable],
         moment: Optional[Callable],
         ndim_supp: int,
-        ndims_params: Optional[Sequence[int]],
+        ndims_params: Sequence[int],
```
zaxtax, line 652: Why is this no longer `Optional` while everywhere else ndim becomes optional

zaxtax (commented):
I have some questions, but understand it enough

pymc/distributions/distribution.py
ricardoV94: May be deprecated but I didn't want to decide yet on that. Won't be deprecated without a warning beforehand

pymc/distributions/distribution.py
ricardoV94: Users don't call this method, the user facing one `dist` and `__new__` always defines ndim from the optional arguments and then passes it here

pymc/distributions/distribution.py
```diff
@@ -725,6 +772,24 @@ def rv_op(
         dummy_params = [dummy_size_param, *dummy_dist_params]
         dummy_updates_dict = collect_default_updates(inputs=dummy_params, outputs=(dummy_rv,))
 
+        # Add size and updates in gufunc signature if they are missing
```
jessegrabowski: This should be split into a helper function imo

pymc/distributions/distribution.py
```diff
@@ -874,14 +939,18 @@ class CustomDist:
...
+        Inferred from signature, if provided. Defaults to assuming
+        a scalar distribution, i.e. ``ndim_supp = 0``
     ndims_params : Optional[Sequence[int]]
         The list of number of dimensions in the support of each of the distribution's
-        parameters. If ``None``, it is assumed that all parameters are scalars, hence
-        the number of dimensions of their support will be 0. This is not needed if an
-        PyTensor dist function is provided.
+        parameters. Inferred from signature, if provided. Defaults to assuming
+        all parameters are scalars, i.e. ``ndims_params=[0, ...]``.
+    signature : Optional[str]
```
jessegrabowski, line 976: The docstring probably needs a code example using the signature, because it's not going to be super clear how to use it right away. Knowing how many empty parenthesis should go on the LHS isn't intuitive if you've never seen it before.

tests/distributions/test_distribution.py
```diff
@@ -728,6 +728,34 @@ def inner_dist(size=None):
...
+        assert out.owner.op.ndim_supp == 0
+        assert out.owner.op.ndims_params == [0, 1]
+
+        # A safe signature can be inferred from ndim_supp and ndims_params
+        out = CustomDist.dist([0.25, 0.75], dist=dist, ndim_supp=0, ndims_params=[0, 1])
+        assert out.owner.op.signature == "(),(i10)->(),()"
+        assert out.owner.op.ndim_supp == 0
+        assert out.owner.op.ndims_params == [0, 1]
+
+        # Otherwise be default we assume everything is scalar, even though it's wrong in this case
```
jessegrabowski, line 769: Will having a wrong signature cause an error somewhere down the pipe we can check with an assert_raises here?

I also prefer to split these tests cases into individual functions to make it more obvious what isn't working when a test fails.

jessegrabowski (approved):
lgtm, added nitpicks but no blockers. Maybe the code example in the `pm.CustomDist` (and/or `SymbolicRandomVariable` for future contributors) would be the closest. 

In general these are great though. Much easier to grok than ndims_support, and potentially more useful.

tests/distributions/test_distribution.py
ricardoV94: Pytest shows which line failed, which is more readable than parametrize imo. 

Parametrize is more readable when you have a loop in the test because the line doesn't tell in which iteration of the loop it failed.

pymc/distributions/distribution.py
ricardoV94: Yeah, I was too lazy. Will add better explanation/ example in the docstrings

pymc/distributions/distribution.py
ricardoV94: Done

pymc/distributions/distribution.py
ricardoV94: Done, although no example has an interesting no scalar sig

tests/distributions/test_distribution.py
ricardoV94: Resolving without changes

PR #7166, opened 2024-02-21
Rename `moment` to `support_point`

### LINKED ISSUE #6974: Rename `moment` to `finite_logp_point`
### Description

The name `moment` has raised confusion a couple of times, recently in #6804 

What that function is supposed to do is to return a point at which the distribution logp is finite (if any point exists).
This is often the mean or mode, but need not always be. We should rename it accordingly

aerubanov:
## Description
Rename `moment` method to avoid confusion with distribution mean. 

## Related Issue
- [x] Closes #6974 

## Checklist
- [x] Rename `moment` method in code (and other related variable names)
- [X] Update documentation

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7166.org.readthedocs.build/en/7166/

### DIFF
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -653,34 +653,34 @@ def check_selfconsistency_discrete_logcdf(
             )
 
 
-def assert_finite_logp_point_is_expected(model, expected, check_finite_logp=True):
+def assert_support_point_is_expected(model, expected, check_finite_logp=True):
     fn = make_initial_point_fn(
         model=model,
         return_transformed=False,
-        default_strategy="finite_logp_point",
+        default_strategy="support_point",
     )
-    finite_logp_point = fn(0)["x"]
+    support_point = fn(0)["x"]
     expected = np.asarray(expected)
     try:
         random_draw = model["x"].eval()
     except NotImplementedError:
-        random_draw = finite_logp_point
+        random_draw = support_point
 
-    assert finite_logp_point.shape == expected.shape
+    assert support_point.shape == expected.shape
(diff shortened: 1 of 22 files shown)


aerubanov:
I rename `moment` to `finite_logp_point` in code. Will change docs accordingly next step.

aerubanov:
@ricardoV94  could you please take a look?

ricardoV94:
@aerubanov great work. I have some suggestions, let me know what you think:
1. Instead of `finite_logp_point`, call it `support_point`? A bit less verbose and still precise
2. Don't break code for people defining Distributions with `moment` methods or using the old `_moment` dispatch or the user-helper `moment`. Instead issue an informative FutureWarning about the name change.

aerubanov:
@ricardoV94 your suggestions sounds good, will implement it

pymc/testing.py
```diff
@@ -653,34 +653,34 @@ def check_selfconsistency_discrete_logcdf(
             )
 
 
-def assert_moment_is_expected(model, expected, check_finite_logp=True):
+def assert_support_point_is_expected(model, expected, check_finite_logp=True):
```
ricardoV94, line 665: This is also user facing (in that we use this utility in other libraries), so should also add a wrapper with a deprecation warning.

aerubanov: I added option for using old name

ricardoV94 (approved):

PR #7188, opened 2024-03-09
Implement specialized transformed logp dispatch

ricardoV94:
## Description

Adds a specialized dispatch for transformed logps

### TODO
- [ ] Add a direct test
- [x] Implement case for ZeroSumNormal

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7188.org.readthedocs.build/en/7188/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -69,7 +69,7 @@ from pymc.distributions.shape_utils import (
     to_tuple,
 )
 from pymc.distributions.transforms import Interval, ZeroSumTransform, _default_transform
-from pymc.logprob.abstract import _logprob
+from pymc.logprob.abstract import _logprob, _transformed_logprob
 from pymc.math import kron_diag, kron_dot
 from pymc.pytensorf import intX
 from pymc.util import check_dist_not_registered
@@ -2818,8 +2818,7 @@ def zerosumnormal_support_point(op, rv, *rv_inputs):
 
 @_default_transform.register(ZeroSumNormalRV)
 def zerosum_default_transform(op, rv):
-    n_zerosum_axes = tuple(np.arange(-op.ndim_supp, 0))
-    return ZeroSumTransform(n_zerosum_axes)
+    return ZeroSumTransform(n_zerosum_axes=op.ndim_supp)
 
 
 @_logprob.register(ZeroSumNormalRV)
@@ -2845,3 +2844,12 @@ def zerosumnormal_logp(op, values, normal_dist, sigma, support_shape, **kwargs):
     )
 
     return check_parameters(out, *zerosums, msg="mean(value, axis=n_zerosum_axes) = 0")
+
+
+@_transformed_logprob.register(ZeroSumNormalRV, ZeroSumTransform)
+def transformed_zerosumnormal_logp(op, transform, unconstrained_value, rv_inputs):
+    _, sigma, _ = rv_inputs
+    zerosum_axes = transform.zerosum_axes
+    if len(zerosum_axes) != op.ndim_supp:
+        raise NotImplementedError
+    return pm.logp(Normal.dist(0, sigma), unconstrained_value).sum(zerosum_axes)
--- a/pymc/logprob/abstract.py
+++ b/pymc/logprob/abstract.py
@@ -38,10 +38,15 @@ import abc
 
 from collections.abc import Sequence
 from functools import singledispatch
+from typing import Union
 
+import multipledispatch
+import pytensor.tensor as pt
+
+from pytensor.gradient import jacobian
 from pytensor.graph.op import Op
 from pytensor.graph.utils import MetaType
-from pytensor.tensor import TensorVariable
+from pytensor.tensor import TensorVariable, Variable
 from pytensor.tensor.elemwise import Elemwise
 from pytensor.tensor.random.op import RandomVariable
 
@@ -153,3 +158,52 @@ class MeasurableElemwise(Elemwise):
 
 
 MeasurableVariable.register(MeasurableElemwise)
+
+
+class Transform(abc.ABC):
+    ndim_supp = None
+
+    @abc.abstractmethod
+    def forward(self, value: TensorVariable, *inputs: Variable) -> TensorVariable:
+        """Apply the transformation."""
+
+    @abc.abstractmethod
+    def backward(
+        self, value: TensorVariable, *inputs: Variable
+    ) -> Union[TensorVariable, tuple[TensorVariable, ...]]:
+        """Invert the transformation. Multiple values may be returned when the
+        transformation is not 1-to-1"""
+
+    def log_jac_det(self, value: TensorVariable, *inputs) -> TensorVariable:
+        """Construct the log of the absolute value of the Jacobian determinant."""
+        if self.ndim_supp not in (0, 1):
+            raise NotImplementedError(
+                f"RVTransform default log_jac_det only implemented for ndim_supp in (0, 1), got {self.ndim_supp=}"
+            )
+        if self.ndim_supp == 0:
+            jac = pt.reshape(pt.grad(pt.sum(self.backward(value, *inputs)), [value]), value.shape)
+            return pt.log(pt.abs(jac))
+        else:
+            phi_inv = self.backward(value, *inputs)
+            return pt.log(pt.abs(pt.nlinalg.det(pt.atleast_2d(jacobian(phi_inv, [value])[0]))))
+
+    def __str__(self):
+        return f"{self.__class__.__name__}"
+
+
+@multipledispatch.dispatch(Op, Transform)
+def _transformed_logprob(
+    op: Op,
+    transform: Transform,
+    unconstrained_value: TensorVariable,
+    rv_inputs: Sequence[TensorVariable],
+):
+    """Create a graph for the log-density/mass of a transformed ``RandomVariable``.
+
+    This function dispatches on the type of ``op``, which should be a subclass
+    of ``RandomVariable`` and ``transform``, which should be a subclass of ``Transform``.
+
+    """
+    raise NotImplementedError(
+        f"Transformed logprob method not implemented for {op} with transform {transform}"
+    )
--- a/pymc/logprob/transform_value.py
+++ b/pymc/logprob/transform_value.py
@@ -23,12 +23,12 @@ from pytensor.graph.features import AlreadyThere, Feature
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.replace import clone_replace
 from pytensor.graph.rewriting.basic import GraphRewriter, in2out, node_rewriter
+from pytensor.graph.rewriting.db import RewriteDatabaseQuery, SequenceDB
 from pytensor.scan.op import Scan
 from pytensor.tensor.variable import TensorVariable
 
-from pymc.logprob.abstract import MeasurableVariable, _logprob
+from pymc.logprob.abstract import MeasurableVariable, Transform, _logprob, _transformed_logprob
 from pymc.logprob.rewriting import PreserveRVMappings, cleanup_ir_rewrites_db
-from pymc.logprob.transforms import Transform
 
 
 class TransformedValue(Op):
@@ -97,7 +97,26 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
     This is introduced by the `TransformValuesRewrite`
     """
     rv_op = rv_outs[0].owner.op
+    transforms = op.transforms
     rv_inputs = rv_outs[0].owner.inputs
+
+    if use_jacobian and len(values) == 1 and len(transforms) == 1:
+        # Check if there's a specialized transform logp implemented
+        [value] = values
+        assert isinstance(value.owner.op, TransformedValue)
+        unconstrained_value = value.owner.inputs[1]
+        [transform] = transforms
+        try:
+            return _transformed_logprob(
+                rv_op,
+                transform,
+                unconstrained_value=unconstrained_value,
+                rv_inputs=rv_inputs,
+                **kwargs,
+            )
+        except NotImplementedError:
+            pass
+
     logprobs = _logprob(rv_op, values, *rv_inputs, **kwargs)
 
     if not isinstance(logprobs, Sequence):
@@ -112,8 +131,8 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
             continue
 
         assert isinstance(value.owner.op, TransformedValue)
-        original_forward_value = value.owner.inputs[1]
-        log_jac_det = transform.log_jac_det(original_forward_value, *rv_inputs).copy()
+        unconstrained_value = value.owner.inputs[1]
+        log_jac_det = transform.log_jac_det(unconstrained_value, *rv_inputs).copy()
         # The jacobian determinant has less dims than the logp
         # when a multivariate transform (like Simplex or Ordered) is applied to univariate distributions.
         # In this case we have to reduce the last logp dimensions, as they are no longer independent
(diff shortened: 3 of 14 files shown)


pymc/logprob/abstract.py
```diff
@@ -153,3 +158,52 @@ def __init__(self, scalar_op, *args, **kwargs):
 
 
 MeasurableVariable.register(MeasurableElemwise)
+
+
+class Transform(abc.ABC):
```
ricardoV94: Moved to abstract, without any changes

ricardoV94:
Is there any expected advantage to this specialization besides avoiding constraints checks (which the user can already do via the model `check_bounds` flag

aseyboldt:
Nice :D

In the case of the ZeroSumNormal probably not really much, once we figure out the expression in the untransformed space it doesn't change too much, and I guess it is nice to have for instance so that the logp can be computed. Just being able to do it on the transformed space would have saved us doing some math...

 I think it is quite common of the logp expressions to be a bit cleaner and possibly numerically more stable on the transformed space.

* For instance for the beta distribution we have this here in the code currently (before the bounds check):
  ```python
        res = (
            pt.switch(pt.eq(alpha, 1.0), 0.0, (alpha - 1.0) * pt.log(value))
            + pt.switch(pt.eq(beta, 1.0), 0.0, (beta - 1.0) * pt.log1p(-value))
            - (pt.gammaln(alpha) + pt.gammaln(beta) - pt.gammaln(alpha + beta))
        )
        res = pt.switch(pt.bitwise_and(pt.ge(value, 0.0), pt.le(value, 1.0)), res, -np.inf)
  ```
  on the (logit) transformed space it could simply be
  ```python
  normalizing_factor = pt.gammaln(alpha + beta) - pt.gammaln(alpha) - pt.gammaln(beta) 
  density = - pt.log1p(pt.exp(-logit_x)) * (alpha + beta) - logit_x * beta + normalizing_factor
  ```
  All the switches are gone, the graph is much shorter (in the original `value` is actually expit(x), so we actually have things like log(expit(x)) and log(1- expit(x)) in the graph).

* Or for instance for the gamma with log transform:
  Now, we have 
  ```python
        beta = pt.reciprocal(scale)
        res = -pt.gammaln(alpha) + logpow(beta, alpha) - beta * value + logpow(value, alpha - 1)
        res = pt.switch(pt.ge(value, 0.0), res, -np.inf)
  ```
  Which could be
  ```python
  beta = pt.reciprocal(scale)
  norm = - beta * pt.log(alpha) - pt.gammaln(beta)
  density = beta * log_x - exp(log_x) / alpha
  ```
  (I guess we could also pass in the transformed point into the logp function, so that the `exp(log_x)` could just be x. The original had things like logpow(exp(x), alpha - 1) in the graph, the new one doesn't. Not a big deal I think, but it is cleaner.

* For  the lognormal what we have to do right now is even a bit silly:

  ```python
        res = pt.switch(
            pt.or_(pt.le(value, 0), pt.ge(value, 1)),
            -np.inf,
            (
                -0.5 * tau * (logit(value) - mu) ** 2
                + 0.5 * pt.log(tau / (2.0 * np.pi))
                - pt.log(value * (1 - value))
            ),
        )
  ```

  This contains `logit(value)`. So we first compute `value = expit(logit_x)` and then just undo it. We also have the ` - pt.log(value * (1 - value))`, which is just the jacobian det of the transform. So we are taking a point in the untransformed space, compute the transformed value and the jac det, then we (manually) untransform the value again, and (manually) subtract the logdet again...

I guess things like this might also be happining in the dirichlet dists, but looking through that is a bit more work...
Nothing here really is a game-changer, but I think it might make quite a few graphs a bit cleaner and a bit more stable.

---

There is also an additional thing we could do with this that right now we can't do with transformations at all. I'm not 100% sure if we should do this though:

Right now all our transformations are injective, because the whole trick with the jacobian determinant doesn't work otherwise. But if we can compute the logp on the transformed space, we could get rid of that requirement.
So for instance we could have a Horseshoe distribution with a transformed space that contains the lambda *and* the x values, and the tranformation just multiplies those.

Or we could have a transformation that maps a point in 2d space to an angel, and use that in for instance the VanMises distribution to avoid the topology problems we have there right now.

pymc/logprob/transform_value.py
```diff
@@ -97,7 +97,26 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
...
+        # Check if there's a specialized transform logp implemented
+        [value] = values
+        assert isinstance(value.owner.op, TransformedValue)
+        unconstrained_value = value.owner.inputs[1]
+        [transform] = transforms
+        try:
+            return _transformed_logprob(
+                rv_op,
+                transform,
+                unconstrained_value=unconstrained_value,
```
aseyboldt, line 113: It would be nice if we could also pass in the transformed value. That way we can avoid computing it twice in a graph if the logp can use that value too. I guess pytensor might get rid of that duplication anyway, but I don't know how reliable that is if the transformation is doing something more complicated.

ricardoV94: We can pass it too. 

You raise a point. Maybe transforms should be encapsulated in an OpFromGraph so that we can easily reverse symbolically and not worry whether they will be simplified in their raw form or not.

ricardoV94:
Re: switches/bound checks, most of those could probably be removed with some domain analysis like exp(x) -> positive so any ge(x, 0) are true, that could be useful beyond PyMC. This could be coupled with user provided type hints (which we are kind of doing with transformed values here) to seed initial info that is not extractable from the graph.

Re: examples, many of those sound like interesting rewrites that PyTensor already does or could do. That need no be a blocker to allow users to work directly on transformed space if they want, but may be a good source of ideas for future rewrites.

The best argument for the functionality in my perspective is the "avoiding hard math", which is more than a nice to have, and could speedup development or allow models that are otherwise impossible to write without a hacky Potential?. 

For our code-base I think we still want to try and provide constrained space logps as much as possible, since not all we do (or may want to do) is logp-based sampling.

twiecki:
What does this do?

aseyboldt:
I just had a look at the generated graphs of each of those.
Let's say we want the logp of the model:

```python
with pm.Model(check_bounds=False) as model:
    pm.ZeroSumNormal("x", shape=(10, 10), n_zerosum_axes=2)
```

The logp + grad with the PR:
```
>>> func1 = model.logp_dlogp_function(mode="NUMBA")
>>> pytensor.dprint(func1._pytensor_function)
Sum{axes=None} [id A] 1
 └─ Composite{((-0.5 * sqr(i0)) - 0.9189385332046727)} [id B] 'x_zerosum___logprob' 0
    └─ x_zerosum__ [id C]
Neg [id D] 'x_zerosum___grad' 2
 └─ x_zerosum__ [id C]

Inner graphs:

Composite{((-0.5 * sqr(i0)) - 0.9189385332046727)} [id B]
 ← sub [id E] 'o0'
    ├─ mul [id F]
    │  ├─ -0.5 [id G]
    │  └─ sqr [id H]
    │     └─ i0 [id I]
    └─ 0.9189385332046727 [id J]
```

And without:

<details>

```
Switch [id A] 'mean(value, axis=n_zerosum_axes) = 0' 31
 ├─ All{axes=None} [id B] 27
 │  └─ MakeVector{dtype='bool'} [id C] 24
 │     ├─ All{axes=None} [id D] 20
 │     │  └─ Composite{and(le((0.1 * abs(i0)), 1e-09), invert(or(isnan((0.1 * i0)), isinf((0.1 * i0)))))} [id E] 16
 │     │     └─ Sum{axis=1} [id F] 11
 │     │        └─ Composite{...}.0 [id G] 9
 │     │           ├─ Join [id H] 8
 │     │           │  ├─ 1 [id I]
 │     │           │  ├─ Sub [id J] 4
 │     │           │  │  ├─ Join [id K] 3
 │     │           │  │  │  ├─ 0 [id L]
 │     │           │  │  │  ├─ x_zerosum__ [id M]
 │     │           │  │  │  └─ Composite{...}.1 [id N] 2
 │     │           │  │  │     └─ ExpandDims{axis=0} [id O] 1
 │     │           │  │  │        └─ Sum{axis=0} [id P] 0
 │     │           │  │  │           └─ x_zerosum__ [id M]
 │     │           │  │  └─ Composite{...}.0 [id N] 2
 │     │           │  │     └─ ···
 │     │           │  └─ Composite{...}.1 [id Q] 7
 │     │           │     └─ ExpandDims{axis=1} [id R] 6
 │     │           │        └─ Sum{axis=1} [id S] 5
 │     │           │           └─ Sub [id J] 4
 │     │           │              └─ ···
 │     │           └─ Composite{...}.0 [id Q] 7
 │     │              └─ ···
 │     └─ All{axes=None} [id T] 22
 │        └─ Composite{and(le((0.1 * abs(i0)), 1e-09), invert(or(isnan((0.1 * i0)), isinf((0.1 * i0)))))} [id U] 18
 │           └─ Sum{axis=0} [id V] 12
 │              └─ Composite{...}.0 [id G] 9
 │                 └─ ···
 ├─ Sum{axes=None} [id W] 13
 │  └─ Composite{...}.2 [id G] 9
 │     └─ ···
 └─ -inf [id X]
Add [id Y] 'x_zerosum___grad' 34
 ├─ SpecifyShape [id Z] 30
 │  ├─ Split{2}.0 [id BA] 26
 │  │  ├─ Add [id BB] 23
 │  │  │  ├─ SpecifyShape [id BC] 15
 │  │  │  │  ├─ Split{2}.0 [id BD] 10
 │  │  │  │  │  ├─ Composite{...}.1 [id G] 9
 │  │  │  │  │  │  └─ ···
 │  │  │  │  │  ├─ 1 [id I]
 │  │  │  │  │  └─ [9 1] [id BE]
 │  │  │  │  ├─ 10 [id BF]
 │  │  │  │  └─ 9 [id BG]
 │  │  │  ├─ Composite{(0.07597469266479578 * (i0 + i1))} [id BH] 21
 │  │  │  │  ├─ SpecifyShape [id BI] 14
 │  │  │  │  │  ├─ Split{2}.1 [id BD] 10
 │  │  │  │  │  │  └─ ···
 │  │  │  │  │  ├─ NoneConst{None} [id BJ]
 │  │  │  │  │  └─ 1 [id I]
 │  │  │  │  └─ ExpandDims{axis=1} [id BK] 17
 │  │  │  │     └─ Sum{axis=1} [id F] 11
 │  │  │  │        └─ ···
 │  │  │  └─ Mul [id BL] 19
 │  │  │     ├─ [[-0.31622777]] [id BM]
 │  │  │     └─ SpecifyShape [id BI] 14
 │  │  │        └─ ···
 │  │  ├─ 0 [id L]
 │  │  └─ [9 1] [id BE]
 │  ├─ 9 [id BG]
 │  └─ 9 [id BG]
 ├─ Composite{(0.07597469266479578 * (i0 - i1))} [id BN] 32
 │  ├─ SpecifyShape [id BO] 29
 │  │  ├─ Split{2}.1 [id BA] 26
 │  │  │  └─ ···
 │  │  ├─ 1 [id I]
 │  │  └─ NoneConst{None} [id BJ]
 │  └─ ExpandDims{axis=0} [id BP] 28
 │     └─ Sum{axis=0} [id BQ] 25
 │        └─ Add [id BB] 23
 │           └─ ···
 └─ Mul [id BR] 33
    ├─ [[-0.31622777]] [id BM]
    └─ SpecifyShape [id BO] 29
       └─ ···

Inner graphs:

Composite{and(le((0.1 * abs(i0)), 1e-09), invert(or(isnan((0.1 * i0)), isinf((0.1 * i0)))))} [id E]
 ← AND [id BS] 'o0'
    ├─ LE [id BT]
    │  ├─ mul [id BU]
    │  │  ├─ t0{0.1} [id BV]
    │  │  └─ Abs [id BW]
    │  │     └─ i0 [id BX]
    │  └─ 1e-09 [id BY]
    └─ Invert [id BZ]
       └─ OR [id CA]
          ├─ IsNan [id CB]
          │  └─ mul [id CC] 't9'
          │     ├─ t0{0.1} [id BV]
          │     └─ i0 [id BX]
          └─ IsInf [id CD]
             └─ mul [id CC] 't9'
                └─ ···

Composite{...} [id G]
 ← sub [id CE] 'o0'
    ├─ i0 [id CF]
    └─ i1 [id CG]
 ← neg [id CH] 'o1'
    └─ sub [id CE] 'o0'
       └─ ···
 ← sub [id CI] 'o2'
    ├─ mul [id CJ]
    │  ├─ -0.5 [id CK]
    │  └─ sqr [id CL]
    │     └─ sub [id CE] 'o0'
    │        └─ ···
    └─ 0.7443402118957849 [id CM]

Composite{...} [id N]
 ← mul [id CN] 'o0'
    ├─ 0.07597469266479578 [id CO]
    └─ i0 [id CP]
 ← sub [id CQ] 'o1'
    ├─ mul [id CN] 'o0'
    │  └─ ···
    └─ mul [id CR]
       ├─ 0.31622776601683794 [id CS]
       └─ i0 [id CP]

Composite{...} [id Q]
 ← mul [id CT] 'o0'
    ├─ 0.07597469266479578 [id CU]
    └─ i0 [id CV]
 ← sub [id CW] 'o1'
    ├─ mul [id CT] 'o0'
    │  └─ ···
    └─ mul [id CX]
       ├─ 0.31622776601683794 [id CY]
       └─ i0 [id CV]

Composite{and(le((0.1 * abs(i0)), 1e-09), invert(or(isnan((0.1 * i0)), isinf((0.1 * i0)))))} [id U]
 ← AND [id CZ] 'o0'
    ├─ LE [id DA]
    │  ├─ mul [id DB]
    │  │  ├─ t7{0.1} [id DC]
    │  │  └─ Abs [id DD]
    │  │     └─ i0 [id DE]
    │  └─ 1e-09 [id DF]
    └─ Invert [id DG]
       └─ OR [id DH]
          ├─ IsNan [id DI]
          │  └─ mul [id DJ] 't11'
          │     ├─ t7{0.1} [id DC]
          │     └─ i0 [id DE]
          └─ IsInf [id DK]
             └─ mul [id DJ] 't11'
                └─ ···

Composite{(0.07597469266479578 * (i0 + i1))} [id BH]
 ← mul [id DL] 'o0'
    ├─ 0.07597469266479578 [id DM]
    └─ add [id DN]
       ├─ i0 [id DO]
       └─ i1 [id DP]

Composite{(0.07597469266479578 * (i0 - i1))} [id BN]
 ← mul [id DQ] 'o0'
    ├─ 0.07597469266479578 [id DR]
    └─ sub [id DS]
       ├─ i0 [id DT]
       └─ i1 [id DU]
```

</details>

I think I like this PR :D

Maybe we can also add an implementation for the beta and the logit-normal dists:

```python
@pm.logprob.abstract._transformed_logprob.register(pt.random.basic.BetaRV, pm.distributions.transforms.LogOddsTransform)
def transformed_beta_logp(op, transform, unconstrained_value, rv_inputs):
    *_, alpha, beta = rv_inputs
    logit_x = unconstrained_value

    normalizing_factor = pt.gammaln(alpha + beta) - pt.gammaln(alpha) - pt.gammaln(beta) 
    logp = normalizing_factor - pt.log1p(pt.exp(-logit_x)) * (alpha + beta) - logit_x * beta
    return pm.distributions.multivariate.check_parameters(
        logp,
        alpha > 0,
        beta > 0,
        "Alpha and beta parameters must be positive in Beta distribution",
    )

@pm.logprob.abstract._transformed_logprob.register(pm.distributions.continuous.LogitNormalRV, pm.distributions.transforms.LogOddsTransform)
def transformed_logit_normal_logp(op, transform, unconstrained_value, rv_inputs):
    *_, mu, sigma = rv_inputs

    return pm.logp(pm.Normal.dist(mu, sigma), unconstrained_value)
```

pymc/distributions/multivariate.py
```diff
@@ -2845,3 +2844,12 @@ def zerosumnormal_logp(op, values, normal_dist, sigma, support_shape, **kwargs):
     )
 
     return check_parameters(out, *zerosums, msg="mean(value, axis=n_zerosum_axes) = 0")
+
+
+@_transformed_logprob.register(ZeroSumNormalRV, ZeroSumTransform)
+def transformed_zerosumnormal_logp(op, transform, unconstrained_value, rv_inputs):
+    _, sigma, _ = rv_inputs
+    zerosum_axes = transform.zerosum_axes
+    if len(zerosum_axes) != op.ndim_supp:
```
aseyboldt, line 2853: I don't think this check is necessary?

ricardoV94: It was a sanity check. A user could have defined an invalid transform manually. I don't really care either way

aseyboldt:
You are right, it seems check_bounds=False didn't do anything, because I wasn't compiling the pytensor functions in a model context. I had no idea I had to do that...

pymc/logprob/transform_value.py
ricardoV94: Actually we have the constrained value as well, its `value.owner.inputs[0]`

PR #7203, opened 2024-03-20
Bump PyTensor dependency and support Python 3.12

### LINKED ISSUE #7202: Extend support to Python 3.12
### Description

Now that there's a new `pytensor` [release](https://github.com/pymc-devs/pytensor/issues/472) supporting Python 3.12, would it be possible to do the same for `pymc`?
For reference, here's the last PR where `pytensor` was bumped: https://github.com/pymc-devs/pymc/pull/7020. I'm not sure if the blockers mentioned there still hold.

ricardoV94:
## Related Issue

- [x] Closes #7202 

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7203.org.readthedocs.build/en/7203/

### DIFF
--- a/tests/backends/test_arviz.py
+++ b/tests/backends/test_arviz.py
@@ -32,7 +32,11 @@ from pymc.backends.arviz import (
 from pymc.exceptions import ImputationWarning
 
 # Turn all warnings into errors for this module
-pytestmark = pytest.mark.filterwarnings("error")
+pytestmark = pytest.mark.filterwarnings(
+    "error",
+    # Related to https://github.com/arviz-devs/arviz/issues/2327
+    "ignore:datetime.datetime.utcnow():DeprecationWarning",
+)
 
 
 @pytest.fixture(scope="module")
@@ -672,13 +676,17 @@ class TestDataPyMC:
             )
             assert "p_interval__" in inference_data.posterior
 
+    @pytest.mark.filterwarnings(
+        "error",
+        # Related to https://github.com/arviz-devs/arviz/issues/2327
+        "ignore:datetime.datetime.utcnow():DeprecationWarning",
+    )
     @pytest.mark.parametrize("chains", (1, 2))
     def test_single_chain(self, chains):
         # Test that no UserWarning is raised when sampling with NUTS defaults
 
         # When this test was added, a `UserWarning: More chains (500) than draws (1)` used to be issued
         # when sampling with a single chain
-        warnings.simplefilter("error")
         with pm.Model():
             pm.Normal("x")
             pm.sample(chains=chains, return_inferencedata=True)
--- a/tests/distributions/test_timeseries.py
+++ b/tests/distributions/test_timeseries.py
@@ -47,10 +47,8 @@ from pymc.sampling.mcmc import sample
 from pymc.testing import assert_support_point_is_expected, select_by_precision
 
 # Turn all warnings into errors for this module
-# Ignoring NumPy deprecation warning tracked in https://github.com/pymc-devs/pytensor/issues/146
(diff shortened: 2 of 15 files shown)


ricardoV94:
Temporarily running the whole suite on 3.12 only

ricardoV94:
All tests passed on 3.12: https://github.com/pymc-devs/pymc/actions/runs/8379574806/job/22948777361?pr=7203

Going to revert so we still run some of the tests on 3.9

tests/distributions/test_timeseries.py
```diff
@@ -48,7 +48,12 @@
 
 # Turn all warnings into errors for this module
 # Ignoring NumPy deprecation warning tracked in https://github.com/pymc-devs/pytensor/issues/146
-pytestmark = pytest.mark.filterwarnings("error", "ignore: NumPy will stop allowing conversion")
+pytestmark = pytest.mark.filterwarnings(
+    "error",
+    "ignore: NumPy will stop allowing conversion",
```
maresb: Probably out-of-scope for this PR, but is this ignore still necessary? The corresponding issue is closed.

Also minor nit, it'd be nice to copy the style for the last comment/argument and inline those two preceding comments.

maresb (approved):
I didn't realize that we only test Python 3.9 with Windows. That should be fine 99% of the time, but for those crazy edge cases maybe we want some sort of pre-release workflow that runs the whole matrix?

tests/distributions/test_timeseries.py
ricardoV94: Removed, thanks for pointing it out

ricardoV94:
> That should be fine 99% of the time, but for those crazy edge cases maybe we want some sort of pre-release workflow that runs the whole matrix?

I'm inclined not to :)

tests/backends/test_arviz.py
```diff
@@ -32,7 +32,11 @@
 from pymc.exceptions import ImputationWarning
 
 # Turn all warnings into errors for this module
```
maresb, line 34: My suggestion was to also inline this. (Turn all warnings into errors for this module)

maresb: But now I'm getting to the nittiest of nits. :)

ricardoV94: But that's why the filterwarnings is there in the first place? Without the temporary filters it looks like
```python
# Turn all warnings ...
pytest.mark.filterwarnings("error")
```

maresb: Do I misunderstand how `filterwarnings` works? The arguments get stacked, right?

```python
pytestmark = pytest.mark.filterwarnings(
    # Turn all warnings into errors for this module
    "error",
    # except those related to https://github.com/arviz-devs/arviz/issues/2327
    "ignore:datetime.datetime.utcnow():DeprecationWarning",
)
```

ricardoV94: My point is that, without the temporary ignores, it makes more sense to have the comment above the call. That also explains why the call is there in the first place.

```python
# Turn all warnings ...
pytest.mark.filterwarnings("error")
```
Makes more sense than
```python
pytest.mark.filterwarnings(
  # Turn all warnings ...
  "error",
)
```

ricardoV94: I'm gonna take the nitpick hint and merge as is

maresb: https://en.wikipedia.org/wiki/Law_of_triviality

PR #7207, opened 2024-03-21
Implement `default_transform` and `transform` argument for distributions

### LINKED ISSUE #5674: Implement `default_transform` and `transform` argument for distributions

aerubanov:
## Related Issue

- [x] Closes #5674 
- [ ] Related to #

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7207.org.readthedocs.build/en/7207/

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -415,8 +415,8 @@ class Distribution(metaclass=DistributionMeta):
         rv_out = model.register_rv(
             rv_out,
             name,
-            observed,
-            total_size,
+            observed=observed,
+            total_size=total_size,
             dims=dims,
             transform=transform,
             default_transform=default_transform,
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -333,6 +333,7 @@ class Distribution(metaclass=DistributionMeta):
         observed=None,
         total_size=None,
         transform=UNSET,
+        default_transform=UNSET,
         **kwargs,
     ) -> TensorVariable:
         """Adds a tensor variable corresponding to a PyMC distribution to the current model.
@@ -418,6 +419,7 @@ class Distribution(metaclass=DistributionMeta):
             total_size,
             dims=dims,
             transform=transform,
+            default_transform=default_transform,
             initval=initval,
         )
 
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -398,6 +398,16 @@ class Distribution(metaclass=DistributionMeta):
         if not isinstance(name, string_types):
             raise TypeError(f"Name needs to be a string but got: {name}")
 
+        if transform is None:
+            transform = UNSET
+            default_transform = None
+            warnings.warn(
+                "To disable default transform, please use default_transform=None"
+                " instead of transform=None. Setting transform to None will"
+                " not take effect in future.",
+                UserWarning,
+            )
+
         dims = convert_dims(dims)
         if observed is not None:
             observed = convert_observed_data(observed)
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -398,15 +398,6 @@ class Distribution(metaclass=DistributionMeta):
         if not isinstance(name, string_types):
             raise TypeError(f"Name needs to be a string but got: {name}")
 
-        if transform is None and default_transform is UNSET:
-            default_transform = None
-            warnings.warn(
-                "To disable default transform, please use default_transform=None"
-                " instead of transform=None. Setting transform to None will"
-                " not have any effect in future.",
-                UserWarning,
-            )
-
         dims = convert_dims(dims)
         if observed is not None:
             observed = convert_observed_data(observed)
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -398,13 +398,12 @@ class Distribution(metaclass=DistributionMeta):
         if not isinstance(name, string_types):
             raise TypeError(f"Name needs to be a string but got: {name}")
 
-        if transform is None:
-            transform = UNSET
+        if transform is None and default_transform is UNSET:
             default_transform = None
             warnings.warn(
                 "To disable default transform, please use default_transform=None"
                 " instead of transform=None. Setting transform to None will"
-                " not take effect in future.",
+                " not have any effect in future.",
                 UserWarning,
             )
 
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1239,7 +1239,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         dims : tuple
             Dimension names for the variable.
         transform
-            Additianal transform which may be applied after default transform.
+            Additional transform which may be applied after default transform.
         default_transform
             A default transform for the random variable in log-likelihood space.
         initval
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -22,7 +22,6 @@ from collections.abc import Iterable, Sequence
 from sys import modules
 from typing import (
     TYPE_CHECKING,
-    Any,
     Literal,
     Optional,
     TypeVar,
@@ -58,6 +57,7 @@ from pymc.exceptions import (
 )
 from pymc.initial_point import make_initial_point_fn
 from pymc.logprob.basic import transformed_conditional_logp
+from pymc.logprob.transforms import Transform
 from pymc.logprob.utils import ParameterValueError, replace_rvs_by_values
 from pymc.model_graph import model_to_graphviz
 from pymc.pytensorf import (
@@ -1239,7 +1239,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
         dims : tuple
             Dimension names for the variable.
         transform
-            A transform for the random variable in log-likelihood space.
+            Additianal transform which may be applied after default transform.
+        default_transform
+            A default transform for the random variable in log-likelihood space.
         initval
             The initial value of the random variable.
 
@@ -1298,8 +1300,8 @@ class Model(WithMemoization, metaclass=ContextMeta):
         rv_var: TensorVariable,
         data: np.ndarray,
         dims,
-        transform: Any | None,
-        default_transform: Any | None,
+        transform: Transform | None,
+        default_transform: Transform | None,
         total_size: int | None,
     ) -> TensorVariable:
         """Create a `TensorVariable` for an observed random variable.
@@ -1313,7 +1315,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
             The observed data.
         dims : tuple
             Dimension names for the variable.
-        transform : int, optional
+        transform
+            Additianal transform which may be applied after default transform.
+        default_transform
             A transform for the random variable in log-likelihood space.
 
         Returns
@@ -1394,8 +1398,8 @@ class Model(WithMemoization, metaclass=ContextMeta):
         self,
         rv_var: TensorVariable,
         *,
-        transform: Any,
-        default_transform: Any,
+        transform: Transform,
+        default_transform: Transform,
         value_var: Variable | None = None,
     ) -> TensorVariable:
         """Create a ``TensorVariable`` that will be used as the random
@@ -1411,7 +1415,11 @@ class Model(WithMemoization, metaclass=ContextMeta):
         ----------
         rv_var : TensorVariable
 
-        transform : Any
+        transform: Transform
+            Additianal transform which may be applied after default transform.
+
+        default_transform: Transform
+            A transform for the random variable in log-likelihood space.
 
         value_var : Variable, optional
 
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1221,8 +1221,8 @@ class Model(WithMemoization, metaclass=ContextMeta):
         observed=None,
         total_size=None,
         dims=None,
-        transform=UNSET,
         default_transform=UNSET,
+        transform=UNSET,
         initval=None,
     ):
         """Register an (un)observed random variable with the model.
@@ -1238,10 +1238,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
             upscales logp of variable with ``coef = total_size/var.shape[0]``
         dims : tuple
             Dimension names for the variable.
-        transform
-            Additional transform which may be applied after default transform.
         default_transform
             A default transform for the random variable in log-likelihood space.
+        transform
+            Additional transform which may be applied after default transform.
         initval
             The initial value of the random variable.
 
@@ -1290,7 +1290,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
             # `rv_var` is potentially changed by `make_obs_var`,
             # for example into a new graph for imputation of missing data.
             rv_var = self.make_obs_var(
-                rv_var, observed, dims, transform, default_transform, total_size
+                rv_var, observed, dims, default_transform, transform, total_size
             )
 
         return rv_var
@@ -1300,8 +1300,8 @@ class Model(WithMemoization, metaclass=ContextMeta):
         rv_var: TensorVariable,
         data: np.ndarray,
         dims,
-        transform: Transform | None,
         default_transform: Transform | None,
+        transform: Transform | None,
         total_size: int | None,
     ) -> TensorVariable:
         """Create a `TensorVariable` for an observed random variable.
@@ -1315,10 +1315,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
             The observed data.
         dims : tuple
             Dimension names for the variable.
-        transform
-            Additianal transform which may be applied after default transform.
         default_transform
             A transform for the random variable in log-likelihood space.
+        transform
+            Additional transform which may be applied after default transform.
 
         Returns
         -------
@@ -1398,8 +1398,8 @@ class Model(WithMemoization, metaclass=ContextMeta):
         self,
         rv_var: TensorVariable,
         *,
-        transform: Transform,
         default_transform: Transform,
+        transform: Transform,
         value_var: Variable | None = None,
     ) -> TensorVariable:
         """Create a ``TensorVariable`` that will be used as the random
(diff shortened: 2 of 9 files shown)


pymc/model/core.py
```diff
@@ -1205,7 +1205,15 @@ def set_data(
         shared_object.set_value(values)
 
     def register_rv(
-        self, rv_var, name, observed=None, total_size=None, dims=None, transform=UNSET, initval=None
+        self,
+        rv_var,
+        name,
+        observed=None,
```
ricardoV94, line 1221:
```suggestion
        *,
        observed=None,
```

pymc/model/core.py
```diff
@@ -1361,7 +1377,11 @@ def make_obs_var(
         return rv_var
 
     def create_value_var(
-        self, rv_var: TensorVariable, transform: Any, value_var: Optional[Variable] = None
+        self,
+        rv_var: TensorVariable,
+        transform: Any,
```
```suggestion
        *,
        transform: Any,
```

pymc/model/core.py
```diff
@@ -1387,11 +1407,17 @@ def create_value_var(
...
+
+        if default_transform is UNSET:
             if rv_var.owner is None:
-                transform = None
+                default_transform = None
             else:
-                transform = _default_transform(rv_var.owner.op, rv_var)
+                default_transform = _default_transform(rv_var.owner.op, rv_var)
+
+        if transform is UNSET:
```
ricardoV94, line 1448: We should add a transition branch, in the case the user set `transform=None` but left `default_transform=UNSET`, which is the code someone would write if they wanted to disable the default transforms in the old API.

If they do that we can, for now, set `default_transform=None`, and issue a UserWarning about the new behavior, explaining that if they want to disable default transforming they will have to set `default_transform=None` in the future instead.

pymc/model/core.py
```diff
@@ -1387,11 +1407,17 @@ def create_value_var(
...
+            transform = default_transform
+        else:
+            transform = ChainedTransform([default_transform, transform])
```
ricardoV94, line 1451: I'm always confused about the order. If we have something like `default_transform=logodds, transform=ordered`, we want `transform.backward` to be ordered but still in logoods. I think your order is correct, but a test will be useful :)

ricardoV94 (commented):
Thanks a lot, left some small user-quality of life suggestions

pymc/model/core.py
aerubanov: Yeah, it`s confused me too. I will add tests for this PR

pymc/distributions/distribution.py
```diff
@@ -396,6 +397,16 @@ def __new__(
...
 
+        if transform is None:
+            transform = UNSET
+            default_transform = None
+            warnings.warn(
+                "To disable default transform, please use default_transform=None"
+                " instead of transform=None. Setting transform to None will"
+                " not take effect in future.",
+                UserWarning,
+            )
```
ricardoV94: We can be a tiny bit more precise here

```suggestion
        if transform is None and default_transform is UNSET:
            default_transform = None
            warnings.warn(
                "To disable default transform, please use default_transform=None"
                " instead of transform=None. Setting transform to None will"
                " not have any effect in future.",
                UserWarning,
            )
```

If the user specified `default_transform`, then we can assume they already know the new API and the warning/magic behavior is not needed. Also I think there's nothing to be gained by setting `transform=UNSET`?

ricardoV94:
> Looks like I fixed all failed test cases. Going add some new tests and changes to documentation as a next steps.

@mkusnetsov took the documentation initiative in #7232 so we should be good to go just with docstrings and tests

aerubanov:
@ricardoV94 I added some test cases to check warning it `transform=None` and transform order if both `default_transform` and `transform` are using

pymc/distributions/distribution.py
```diff
@@ -397,6 +398,15 @@ def __new__(
         if not isinstance(name, string_types):
             raise TypeError(f"Name needs to be a string but got: {name}")
 
+        if transform is None and default_transform is UNSET:
```
ricardoV94: This warning should be in the relevant section in `pm.Model` instead of `Distribution`

pymc/model/core.py
```diff
@@ -1396,11 +1422,16 @@ def create_value_var(
...
             if rv_var.owner is None:
-                transform = None
+                default_transform = None
             else:
-                transform = _default_transform(rv_var.owner.op, rv_var)
+                default_transform = _default_transform(rv_var.owner.op, rv_var)
+
+        if transform is UNSET:
+            transform = default_transform
+        elif transform and default_transform:
```
ricardoV94: Be explicit about the possible cases at this point?
```suggestion
        elif transform is not None and default_transform is not None:
```

tests/distributions/test_mixture.py
```diff
@@ -1359,17 +1359,25 @@ def test_warning(self):
 
             with warnings.catch_warnings():
                 warnings.simplefilter("error")
-                Mixture("mix4", w=[0.5, 0.5], comp_dists=comp_dists, transform=None)
+                Mixture("mix4", w=[0.5, 0.5], comp_dists=comp_dists, default_transform=None)
+
+            with pytest.warns(
```
ricardoV94: This functionality shouldn't be tested here, since it's not specific to Mixture. Probably in model/test_core.py there should be related stuff already? I see you already have it there, so is this test needed?

tests/distributions/test_transform.py
```diff
@@ -619,7 +619,7 @@ def test_transform_univariate_dist_logp_shape():
 
 def test_univariate_transform_multivariate_dist_raises():
     with pm.Model() as m:
-        pm.Dirichlet("x", [1, 1, 1], transform=tr.log)
+        pm.Dirichlet("x", [1, 1, 1], transform=tr.log, default_transform=None)
```
ricardoV94: I think this is more in the spirit of the original test?
```suggestion
        pm.Dirichlet("x", [1, 1, 1], default_transform=tr.log)
```

tests/distributions/test_transform.py
```diff
@@ -645,7 +645,7 @@ def log_jac_det(self, value, *inputs):
     buggy_transform = BuggyTransform()
 
     with pm.Model() as m:
-        pm.Uniform("x", shape=(4, 3), transform=buggy_transform)
+        pm.Uniform("x", shape=(4, 3), transform=buggy_transform, default_transform=None)
```
ricardoV94:
```suggestion
        pm.Uniform("x", shape=(4, 3), default_transform=buggy_transform)
```

tests/logprob/test_utils.py
```diff
@@ -218,11 +218,15 @@ def test_interdependent_transformed_rvs(self, reversed):
...
             # Operation between the variables provides a regression test for #7054
-            y = pm.Uniform("y", lower=0, upper=pt.exp(x), transform=transform)
-            z = pm.Uniform("z", lower=0, upper=y, transform=transform)
-            w = pm.Uniform("w", lower=0, upper=pt.square(z), transform=transform)
+            y = pm.Uniform(
+                "y", lower=0, upper=pt.exp(x), transform=transform, default_transform=None
+            )
+            z = pm.Uniform("z", lower=0, upper=y, transform=transform, default_transform=None)
+            w = pm.Uniform(
+                "w", lower=0, upper=pt.square(z), transform=transform, default_transform=None
```
ricardoV94: Pass transform to `default_transform`

tests/model/test_core.py
```diff
@@ -527,6 +527,42 @@ def test_model_var_maps():
...
+                nonlocal transform_order
+                transform_order.append(self.marker)
+                return value
+
+            def backward(self, value, *inputs):
+                return value
+
+        with pm.Model() as model:
+            x = pm.Normal("x", transform=DummyTransform(2), default_transform=DummyTransform(1))
+        assert transform_order == [1, 2]
```
ricardoV94: You can use regular transforms, and simply assert the obtained transform is a Chain, which as a property like `transform_list` that includes the transforms, and you can assert those are also the expected ones. The transform is available in `models.rvs_to_transforms[x]`

Also, would be nice to include a numerical example that would have led to `nan` or `-inf` probability before the change, like an `ordered mixture of LogNormals` evaluated at -1

ricardoV94 (changes requested):

pymc/model/core.py
```diff
@@ -1214,7 +1214,16 @@ def set_data(
...
-        self, rv_var, name, observed=None, total_size=None, dims=None, transform=UNSET, initval=None
+        self,
+        rv_var,
+        name,
+        *,
+        observed=None,
+        total_size=None,
+        dims=None,
+        transform=UNSET,
+        default_transform=UNSET,
```
ricardoV94, line 1224: Missing description in the docstrings

pymc/model/core.py
```diff
@@ -1288,6 +1299,7 @@ def make_obs_var(
         data: np.ndarray,
         dims,
         transform: Any | None,
+        default_transform: Any | None,
```
ricardoV94: Missing description in the docstrings.

ricardoV94: Also type hint isn't great, should be `RVTransform | None` (or something like that, don't remember the exact class now)

pymc/model/core.py
```diff
@@ -1363,14 +1382,21 @@ def make_obs_var(
...
 
         return rv_var
 
     def create_value_var(
-        self, rv_var: TensorVariable, transform: Any, value_var: Variable | None = None
+        self,
+        rv_var: TensorVariable,
+        *,
+        transform: Any,
+        default_transform: Any,
```
ricardoV94: Update type hints and docstring

ricardoV94 (commented):
Besides making the `test_transform_order` simpler, just need some tweaks to the docstrings. They don't describe the parameter. We can also give better type hint than `Any`. If mypy complaints just revert to `Any`

aerubanov:
@ricardoV94 Is there any way to get `logp` value after transform applied? I`m trying something like   
```python
    with pm.Model() as model:
        x1 = pm.LogNormal("x1", 0.0, 1.0)
        x2 = pm.LogNormal("x2", 0.0, 1.0, default_transform=LogTransform())
        assert pm.logp(x1, 1.0).eval() != pm.logp(x2, 1.0).eval()
```
and got assertion error

ricardoV94:
> @ricardoV94 Is there any way to get `logp` value after transform applied? I`m trying something like
> 
> ```python
>     with pm.Model() as model:
>         x1 = pm.LogNormal("x1", 0.0, 1.0)
>         x2 = pm.LogNormal("x2", 0.0, 1.0, default_transform=LogTransform())
>         assert pm.logp(x1, 1.0).eval() != pm.logp(x2, 1.0).eval()
> ```
> 
> and got assertion error

You want to use `model.compile_logp()`, and you'll have to create two models, one with transform and one without

aerubanov:
@ricardoV94 I added new test case with numerical example for transform args. Please let me know what do you thunk about it.

tests/model/test_core.py
```diff
@@ -527,6 +527,33 @@ def test_model_var_maps():
...
+                " please use default_transform=None"
+                " instead of transform=None. Setting transform to"
+                " None will not have any effect in future.",
+            ):
+                a = pm.Normal("a", transform=None)
+
+    def test_transform_order(self):
+        with pm.Model() as model:
+            x = pm.Normal("x", transform=Interval(0, 1), default_transform=log)
+        assert isinstance(model.rvs_to_transforms[x].transform_list[0], LogTransform)
```
ricardoV94: Just for clarity assert the transform is a ChainTransform before asserting the contents?

tests/model/test_core.py
```diff
@@ -527,6 +527,33 @@ def test_model_var_maps():
...
+        assert isinstance(model.rvs_to_transforms[x].transform_list[1], Interval)
+
+    def test_default_transform_is_applied(self):
+        with pm.Model() as model1:
+            x1 = pm.LogNormal("x1", 0, 1, transform=Interval(-2, 2), default_transform=None)
+        with pm.Model() as model3:
+            x2 = pm.LogNormal("x2", 0, 1, transform=Interval(-2, 2), default_transform=log)
```
ricardoV94: This is fine but is not very realistic. What about ordered transform + log which is the example that motivated this PR?

```suggestion
            x2 = pm.LogNormal("x2", 0, 1, transform=Interval(-2, 2), default_transform=log)
```

pymc/model/core.py
```diff
@@ -1230,7 +1239,9 @@ def register_rv(
         dims : tuple
             Dimension names for the variable.
         transform
-            A transform for the random variable in log-likelihood space.
+            Additianal transform which may be applied after default transform.
```
ricardoV94: Typo:
```suggestion
            Additional transform which may be applied after default transform.
```

pymc/model/core.py
```diff
@@ -1230,7 +1239,9 @@ def register_rv(
...
+        default_transform
```
ricardoV94, line 1241: Nitpick: show default_transform before transform (also in the signature)?

pymc/model/core.py
```diff
@@ -1301,7 +1315,9 @@ def make_obs_var(
             The observed data.
         dims : tuple
             Dimension names for the variable.
-        transform : int, optional
+        transform
+            Additianal transform which may be applied after default transform.
+        default_transform
```
ricardoV94: Same as above

pymc/model/core.py
```diff
@@ -1385,7 +1415,11 @@ def create_value_var(
         ----------
         rv_var : TensorVariable
 
-        transform : Any
+        transform: Transform
+            Additianal transform which may be applied after default transform.
+
+        default_transform: Transform
+            A transform for the random variable in log-likelihood space.
```
ricardoV94: Same as above

tests/model/test_core.py
ricardoV94: The rest of the logic of the test is spot-on!

ricardoV94 (commented):
Thanks @aerubanov 

I left some minor comments, I think this is the last round!

aerubanov:
@ricardoV94 please review again

tests/model/test_core.py
```diff
@@ -527,6 +534,34 @@ def test_model_var_maps():
...
+                a = pm.Normal("a", transform=None)
+
+    def test_transform_order(self):
+        with pm.Model() as model:
+            x = pm.Normal("x", transform=Interval(0, 1), default_transform=log)
+        assert isinstance(model.rvs_to_transforms[x], ChainedTransform)
+        assert isinstance(model.rvs_to_transforms[x].transform_list[0], LogTransform)
+        assert isinstance(model.rvs_to_transforms[x].transform_list[1], Interval)
+
+    def test_default_transform_is_applied(self):
```
ricardoV94, line 557: The previous test can be removed, in favor of this one. Just include the Chain / ordering checks here?

tests/model/test_core.py
```diff
@@ -527,6 +534,34 @@ def test_model_var_maps():
...
+        with pm.Model() as model1:
+            x1 = pm.LogNormal("x1", [0, 0], [1, 1], transform=ordered, default_transform=None)
+        with pm.Model() as model3:
```
ricardoV94: Super nitpick, feel free to ignore:
```suggestion
        with pm.Model() as model2:
```

tests/model/test_core.py
```diff
@@ -527,6 +534,34 @@ def test_model_var_maps():
...
+            x2 = pm.LogNormal("x2", [0, 0], [1, 1], transform=ordered, default_transform=log)
```
ricardoV94: I would test this one like this (the way the user would write it):

```suggestion
            x2 = pm.LogNormal("x2", [0, 0], [1, 1], transform=ordered)
```

tests/model/test_core.py
ricardoV94: Never mind, I see what you did

tests/model/test_core.py
```diff
@@ -527,6 +534,34 @@ def test_model_var_maps():
...
+                " please use default_transform=None"
+                " instead of transform=None. Setting transform to"
+                " None will not have any effect in future.",
+            ):
+                a = pm.Normal("a", transform=None)
+
+    def test_transform_order(self):
+        with pm.Model() as model:
+            x = pm.Normal("x", transform=Interval(0, 1), default_transform=log)
+        assert isinstance(model.rvs_to_transforms[x], ChainedTransform)
```
ricardoV94: Nitpick (feel free to ignore): Save the transform in a separate variable so you don't need to write 3 times `model.rvs_to_transforms[x]`

ricardoV94 (approved):
Thanks, looks great!

ricardoV94:
We should follow up with an issue to provide informative warnings when doing prior/posterior predictive sampling of variables with custom non-default transforms

PR #7211, opened 2024-03-22
List more math function in API docs

### LINKED ISSUE #7130: ENH: Expansion of pymc.math to include more numpy/pytensor functions
### Before

```python
Due to pymc.math missing a number of common numpy/pytensor operations, this is primarily the inverse trig functions but it would be worth combing through the full list to see if there are any other notable absences. However since these functions can still be found through pytensor like so:


import pymc as pm
import pytensor.tensor as pt

test=pt.scalar(name="test")
res=pm.math.arccos(test) #Fails
res=pt.arccos(test) #Succeeds
```
```


### After

```python
This should work essentially the same as the other pymc.math functions so along the lines of:


import pymc as pm
import pytensor.tensor as pt

test=pt.scalar(name="test")
res=pm.math.arccos(test) #Succeeds
```
```


### Context for the issue:

This issue was spawned from this post from the PyMC discourse (https://discourse.pymc.io/t/no-inverse-trig-in-pm-math-functions/13681/2) and the slightly later discussion post (https://github.com/pymc-devs/pymc/discussions/7124). The issue is mainly a matter of convenience so as to have a complete set of numpy/pytensor operations all accessible through pm.math, and the pymc.math documentation on the website should hopefully reflect this as well.

### PR DESCRIPTION
**Pull request by brandonhorsley**
## Description

This commit is a solution to the problems presented in PyMC github issue #7130. The first part of the issue was that certain functions like inverse trigonometric functions didn't seem to work, however on a clean install for this commit and pull request, that issue disappeared and the functions work as intended. These functions were missing among others from the website documentation which is the second part of the github issue and is what this particular git commit concerns. Note I have included tround although other codes reflect this as being deprecated but still able to be used. I have also rearranged the ordering to better reflect similar functions being bunched together.
## Related Issue

- [x] Closes #7130
- [] Related to # 

## Checklist

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7211.org.readthedocs.build/en/7211/

### DIFF
--- a/pymc/math.py
+++ b/pymc/math.py
@@ -335,7 +335,7 @@ def log1mexp_numpy(x, *, negative_input=False):
     https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf
     """
     warnings.warn(
-        "pymc.math.log1mexp_numpy is being deprecated. Use log1mexp instead.",
+        "pymc.math.log1mexp_numpy is being deprecated.",
         FutureWarning,
         stacklevel=2,
     )
--- a/pymc/math.py
+++ b/pymc/math.py
@@ -282,7 +282,7 @@ def logdiffexp(a, b):
 def logdiffexp_numpy(a, b):
     """log(exp(a) - exp(b))"""
     warnings.warn(
-        "pymc.math.logdiffexp_numpy is being deprecated. Use logdiffexp instead.",
+        "pymc.math.logdiffexp_numpy is being deprecated.",
         FutureWarning,
         stacklevel=2,
     )
--- a/pymc/math.py
+++ b/pymc/math.py
@@ -73,6 +73,7 @@ from pytensor.tensor import (
     ones_like,
     or_,
     prod,
+    round,
     sgn,
     sigmoid,
     sin,
@@ -178,6 +179,7 @@ __all__ = [
     "expand_packed_triangular",
     "batched_diag",
     "block_diagonal",
+    "round",
 ]
 
 
@@ -272,20 +274,6 @@ def kron_diag(*diags):
     return reduce(flat_outer, diags)
 
 
-def round(*args, **kwargs):
-    """
-    Temporary function to silence round warning in PyTensor. Please remove
-    when the warning disappears.
-    """
-    kwargs["mode"] = "half_to_even"
-    return pt.round(*args, **kwargs)
-
-
-def tround(*args, **kwargs):
-    warnings.warn("tround is deprecated. Use round instead.")
-    return round(*args, **kwargs)
-
-
 def logdiffexp(a, b):
     """log(exp(a) - exp(b))"""
     return a + pt.log1mexp(b - a)
@@ -293,6 +281,11 @@ def logdiffexp(a, b):
 
 def logdiffexp_numpy(a, b):
     """log(exp(a) - exp(b))"""
+    warnings.warn(
+        "pymc.math.logdiffexp_numpy is being deprecated. Use logdiffexp instead.",
+        FutureWarning,
+        stacklevel=2,
+    )
     return a + log1mexp_numpy(b - a, negative_input=True)
 
 
@@ -341,6 +334,11 @@ def log1mexp_numpy(x, *, negative_input=False):
     For details, see
     https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf
     """
+    warnings.warn(
+        "pymc.math.log1mexp_numpy is being deprecated. Use log1mexp instead.",
+        FutureWarning,
+        stacklevel=2,
+    )
     x = np.asarray(x, dtype="float")
 
     if not negative_input:
--- a/tests/distributions/test_continuous.py
+++ b/tests/distributions/test_continuous.py
@@ -418,6 +418,7 @@ class TestMatchesScipy:
             return np.log(a) + np.log(b) + (a - 1) * np.log(value) + (b - 1) * np.log(1 - value**a)
 
         def scipy_log_cdf(value, a, b):
+            warnings.warn("pymc.math.log1mexp_numpy is being deprecated.", FutureWarning)
             return pm.math.log1mexp_numpy(b * np.log1p(-(value**a)), negative_input=True)
 
         check_logp(
--- a/tests/distributions/test_continuous.py
+++ b/tests/distributions/test_continuous.py
@@ -417,15 +417,11 @@ class TestMatchesScipy:
         def scipy_log_pdf(value, a, b):
             return np.log(a) + np.log(b) + (a - 1) * np.log(value) + (b - 1) * np.log(1 - value**a)
 
+        def log1mexp(x):
+            return np.log1p(-np.exp(x)) if x < np.log(0.5) else np.log(-np.expm1(x))
+
         def scipy_log_cdf(value, a, b):
-            x = b * np.log1p(-(value**a))
-            x = np.asarray(x, dtype="float")
-            out = np.empty_like(x)
-            mask = x < -0.6931471805599453  # log(1/2)
-            out[mask] = np.log1p(-np.exp(x[mask]))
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by brandonhorsley** [comment]
Doc build worked as expected which is good news, does read a bit long so maybe further polish could be doing further category subdivisions like 'Trig functions', 'tensor functions'...etc but this would probably warrant discussion on what good subcategories could be

**Review by OriolAbril — APPROVED** [review]
Everything looks good, I agree with sorting them into multiple topic related subsections but can be done in a follow up PR

**Comment by jessegrabowski** [comment]
Do all of these functions need to be exposed? I don't think the `_numpy` functions need to be there, but also the linear algebra routines.

**Comment by OriolAbril** [comment]
> Do all of these functions need to be exposed? I don't think the _numpy functions need to be there, but also the linear algebra routines.

If we expect them to be used they should be somewhere. I am under the impression it is the case, but if it were preferred to use them from pytensor then they can be removed from here and listed on pytensor only.

**Comment by jessegrabowski** [comment]
My only strong objection is to the numpy ones, because we shouldn't be even suggesting that users use numpy functions on pytensor objects. But I guess this PR is just updating the docs to reflect what functions are already there, so maybe this concern should go in a different thread.

**Comment by OriolAbril** [comment]
There is the underlying issue that it isn't really clear what is PyMC's public API, even to team members. But I think this is a good place to discuss that for these functions. Any functions we don't think are public should NOT go into the docs

**Comment by brandonhorsley** [comment]
Yeah I mean with the pull request I have put forward I just covered all the functions presented in the math.py file just out of sheer completeness since I didn't want to make that executive decision without consultation. I believe I also included 'tround' in the math.rst file for this pull request which is listed as deprecated in math.py but does still work so that one would be another one to possibly put onto the cutting room floor since all it does is raise a deprecation warning and call 'round' anyway

**Comment by ricardoV94** [comment]
Let's just show the PyTensor functions and not list the numpy ones. Further refining can be done later.

**Comment by brandonhorsley** [comment]
Is that something to do on my end? If so is it worth omitting the deprecated 'tround' from the list as well or just the numpy ones?

**Comment by ricardoV94** [comment]
You can actually remove the deprecated tround and yes omit the numpy ones

**Review by ricardoV94 — APPROVED** [review]

**Comment by ricardoV94** [comment]
I actually meant removing the actual tround code as well

**Comment by brandonhorsley** [comment]
Ahh I'll do that now, remove the numpy code from math.py too?

**Comment by ricardoV94** [comment]
I don't know if the numpy code is being used.

Regarding tround, you can remove that code. There's also a round wrapper that can be removed and the module changed to provide access to pytensor.tensor round directly

**Comment by ricardoV94** [comment]
If you want (totally optional), you can deprecate the `_numpy` functions for now with a `FutureWarning`. I see that one of them is being used in a test, but we can just define it there then, no need to be part of our public modules

**Review by jessegrabowski — APPROVED** [review]

File: pymc/math.py [hunk 1]
```diff
@@ -272,27 +274,18 @@ def kron_diag(*diags):
...
-
 def logdiffexp(a, b):
     """log(exp(a) - exp(b))"""
     return a + pt.log1mexp(b - a)
 
 
 def logdiffexp_numpy(a, b):
     """log(exp(a) - exp(b))"""
+    warnings.warn(
+        "pymc.math.logdiffexp_numpy is being deprecated. Use logdiffexp instead.",
```
ricardoV94: They are not equivalent. We don't need to offer an alternative

```suggestion
        "pymc.math.logdiffexp_numpy is being deprecated.",
```

pymc/math.py
```diff
@@ -341,6 +334,11 @@ def log1mexp_numpy(x, *, negative_input=False):
     For details, see
     https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf
     """
+    warnings.warn(
+        "pymc.math.log1mexp_numpy is being deprecated. Use log1mexp instead.",
```
ricardoV94:
```suggestion
        "pymc.math.log1mexp_numpy is being deprecated.",
```

ricardoV94 (changes requested):

pymc/math.py
ricardoV94: Also need to implement it locally here, so that it can be safely removed later:

https://github.com/pymc-devs/pymc/blob/f302882da263d8569af1f55b3d758c614ecb3799/tests/distributions/test_continuous.py#L427

brandonhorsley: OK, will do, I'll check for all cases where those numpy functions show up, is that just to have the comment visually written there so it can be identified from looking at the file?

ricardoV94: Sorry, what do you mean?

brandonhorsley: As in since the warning is applied to the function definition 'log1mexp_numpy' the warning would be raised when those codes are run, so if I'm understanding correctly I am to write another warning on that test_continuous.py line so when you talk about safely removing it, I assume that means that by writing the warning code there that you know that it is to be removed/altered there as well by just looking at the line of code itself without needing to run the script to get the warning If that makes sense?

I also see there is another file where the function comes up as well so I will edit that too and I will do a search for cases where logdiffexp_numpy pops up and another search in case tround happens to show up anywhere else too

ricardoV94: I mean no other part of the codebase should be using a function marked for deprecation (other than code that is testing that function directly). That ensures then when the time comes all we need to do is remove the deprecated code.

tests/distributions/test_continuous.py
```diff
@@ -418,6 +418,7 @@ def scipy_log_pdf(value, a, b):
             return np.log(a) + np.log(b) + (a - 1) * np.log(value) + (b - 1) * np.log(1 - value**a)
 
         def scipy_log_cdf(value, a, b):
+            warnings.warn("pymc.math.log1mexp_numpy is being deprecated.", FutureWarning)
             return pm.math.log1mexp_numpy(b * np.log1p(-(value**a)), negative_input=True)
```
ricardoV94: The test function doesn't need to issue a warning. Instead it should not use `math.log1mex_numpy` but a non user facing helper that achieves the same

tests/test_math.py
```diff
@@ -148,6 +148,8 @@ def test_log1mexp():
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", "divide by zero encountered in log", RuntimeWarning)
         warnings.filterwarnings("ignore", "invalid value encountered in log", RuntimeWarning)
+
+        warnings.warn("pymc.math.log1mexp_numpy is being deprecated.", FutureWarning)
```
ricardoV94: No need to issue the warning in the tests. Test code is not user facing. Same for all other warnings below.

tests/distributions/test_continuous.py
brandonhorsley: Ahh OK

tests/test_math.py
brandonhorsley: OK so I can take out those warnings and I presume then it is the same as the test_continuous case of just substituting in alternative code that achieves the same or is it that i should just undo my edits in test_math.py since they are test codes?

ricardoV94: This is different from the `test_continuous`. This is testing the functions directly, so can stay as is. This code will be removed when the respective functions are removed.

The `test_continuous` is the only one that needs a separate version of the function. As it will still need it when we remove the deprecated code.

ricardoV94:
@brandonhorsley I pushed a commit that refactors a bit the test code to be more readable. Otherwise I think all is good. Tests are running now

ricardoV94 (approved):

ricardoV94:
Thanks @brandonhorsley

brandonhorsley:
Thanks for your help as well @ricardoV94 et al., you've been a big help for getting this first pull request under my belt!

ricardoV94:
You're welcome. Looking for your second if that happens!

PR #7231, opened 2024-03-29
Make default STEP_METHODS a list that can be modified

ricardoV94:
## Description

Having as a tuple is a bit cumbersome and makes it seem like it should not be modified. However, until we have a more proper system (disptach based probably), this is the API for other packages (including BART, pymc-experimental) to define default samplers.

This PR makes it a list (and over-engineers a fixture for the test :D, happy to remove if you think it's silly)

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/step_methods/__init__.py
+++ b/pymc/step_methods/__init__.py
@@ -12,7 +12,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-from pymc.step_methods.compound import CompoundStep
+from pymc.step_methods.compound import BlockedStep, CompoundStep
 from pymc.step_methods.hmc import NUTS, HamiltonianMC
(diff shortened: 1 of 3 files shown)


twiecki (approved):

jessegrabowski (approved):
Looks nice and simple. Trying to think about corner cases where a simple list wouldn't be good enough for holding the step methods. Does it matter if a user appends the same sampler to the list multiple times? Or is the ordering of the list important anywhere downstream?

ricardoV94:
The order is not important but repeated entries wouldn't be either

Issue #7009, opened 2023-11-14
Deprecate SumTo1 Transform


ricardoV94:
### Description

This seems to be an old transform that was replaced by the Simplex transform.

twiecki:
@PulsatingGenius Great, just open a PR (can be draft too).

nyeshlur:
Hello if this issue is not currently being worked on I would like to work on it. I was wondering what needs to be included in a draft pull request?

ricardoV94:
For now just a FutureWarning when users access the transform to let them know it's deprecated and will be removed in the future 

Plus a test that verified the warning is working but you can still access it

nyeshlur:
Hi so I have created a FutureWarning in pymc/distributions/transforms.py and I have written a test in tests/distributions/test_transform.py. I'm not sure if this is what is needed and I'm also unsure of how to run the test. Any help is appreciated.

ricardoV94:
Thanks, I'll take a look but in the meantime this may give you some hints: https://www.pymc.io/projects/docs/en/stable/contributing/running_the_test_suite.html

Diwakar-Gupta:
Hi @ricardoV94 this is resolved and can be closed.
PR: https://github.com/pymc-devs/pymc/pull/7232

ricardoV94:
It still exists, it should be removed after enough time passed

Jiya873:
Hi @ricardoV94, I checked `pymc/distributions/transforms.py` and found that `SumTo1` is still there and currently lacks a deprecation warning. I want to ask you whether I should add a `FutureWarning` or directly remove the entire class.

ricardoV94:
We can warn first

andreruizloera:
I've opened #8356 implementing the soft deprecation discussed here: a `FutureWarning` on `SumTo1()` instantiation and on `transforms.sum_to_1` access, with `sum_to_1` served lazily via a module `__getattr__` so `import pymc` stays warning-free. The transform remains fully functional, and there's a test covering both access paths. Happy to adjust the approach or messaging.

Issue #6975, opened 2023-10-29
Ordered transform incompatible with constrained space transforms (ZeroSum, Simplex, etc.)


michaelosthege:
### Describe the issue:

When applying the ordered transform to the ZSN the model suffers a `-inf` logp.

### Reproduceable code example:

```python
with pm.Model() as pmodel:
    pm.ZeroSumNormal("zsn", shape=2, transform=pm.distributions.transforms.ordered)
pmodel.debug()
```

### Error message:

```shell
RuntimeWarning: divide by zero encountered in log
  variables = ufunc(*ufunc_args, **ufunc_kwargs)
point={'zsn_ordered__': array([  0., -inf])}

The variable zsn has the following parameters:
0: normal_rv{0, (0, 0), floatX, False}.1 [id A] <Vector(float64, shape=(2,))>
 ├─ RandomGeneratorSharedVariable(<Generator(PCG64) at 0x1E355016500>) [id B] <RandomGeneratorType>
 ├─ [2] [id C] <Vector(int64, shape=(1,))>
 ├─ 11 [id D] <Scalar(int64, shape=())>
 ├─ 0 [id E] <Scalar(int8, shape=())>
 └─ 1.0 [id F] <Scalar(float64, shape=())>
1: 1.0 [id F] <Scalar(float64, shape=())>
2: [2] [id G] <Vector(int32, shape=(1,))>
The parameters evaluate to:
0: [1.40383536 0.612628  ]
1: 1.0
2: [2]
Some of the values of variable zsn are associated with a non-finite logp:
 value = [  0. -inf] -> logp = -inf
```

### PyMC version information:

5.9.1 at 419af0688353292c0a356cddbb9271737e89a723

michaelosthege:
I looked into it, but my understanding of the ZSN and Ordered transform is limited.

The Ordered transform works by doing a set-subtensor on a diff, and I guess this is what clashes with how the ZSN is also using `set_subtensor` in its logp?

ricardoV94:
You need to chain the OrderedTransform with the default transform ZSN has

ricardoV94:
Related to https://github.com/pymc-devs/pymc/issues/5674

michaelosthege:
Okay, makes sense.
I tried this, but it doesn't work, so obviously I'm doing something wrong:

```python
with pm.Model() as pmodel:
    rv = pm.ZeroSumNormal.dist(shape=2)
    trafos = [
[...]

PR #7232, opened 2024-03-29
Add introduction section to Transformations API page to document role of Transforms

### LINKED ISSUE #7040: Document the role of transforms
### Describe the issue:

If a transformation is done with an observed variable, it seems to flat out be ignored. 

In the example below,  the observations are out of the interval given to pm.Normal, but it still fits as if the transformation was not there. Sampling the posterior predictive also ignores the transform, giving values way out of the range.

Transformations are currently not very well documented so I might be misunderstanding something, but having an interval transform keep values within the interval even on observations seems like a sensible expectation so I'm filing this as a bug.

```python
import pymc as pm
import numpy as np

with pm.Model() as model:
    
    obs = pm.MutableData("obs", np.array([2]*10 + [-2]*10))                
    
    sd = pm.HalfNormal('sd', sigma=2)
    val = pm.Normal('val', mu=0, sigma=sd,    
                     transform=pm.distributions.transforms.Interval(-1,1),
                     observed=obs)
    
    idata = pm.sample()
    pp = pm.sample_posterior_predictive(idata)
    
    pp.posterior_predictive.val.min(), pp.posterior_predictive.val.max()
```

### Error message:

_No response_

### PyMC version information:

PyMC 5.9.2

### Context for the issue:

_No response_

mkusnetsov:
## Description

This PR adds an introduction section to the Transformations API page to:
- explain the principal purpose of transforms
- warn that most users don't need to play with these and that most attempts to do so don't do what users tend to think
- provide some guidance on when tweaking transforms may be appropriate and what other options user have when wanting to deal with transformed random variables

In addition this PR:
- removes a seemingly incorrect deprecation comment next to the `ordered` transform (I think it was accidentally left over from when `univariate_ordered` was deprecated)
- removes mention of `SumsTo1`, which is being deprecated
- organises instance and class names in the autosummary to 
  1. alphebtize them, and
  2. add missing class names for the included instances

This PR assumes that the issue #5674 is closed (via PR #7207) to avoid extensive (and somewhat non-intuitive to casual user) guidance that will shortly be outdated

## Related Issue

- [x] Closes #7040
- [x] Related to #6975, #5674, #7009

## Checklist

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7232.org.readthedocs.build/en/7232/

### DIFF
--- a/docs/source/api/distributions/transforms.rst
+++ b/docs/source/api/distributions/transforms.rst
@@ -4,6 +4,36 @@ Transformations
 
 .. currentmodule:: pymc.distributions.transforms
 
+While many distributions are defined on constrained spaces (e.g. intervals), MCMC samplers typically perform best when sampling on the unconstrained real line; this is especially true of HMC samplers. PyMC balances this through the use of transforms. A transform instance can be passed to the constructor of a random variable to tell the sampler how to move between the underlying unconstrained space where the samples are actually drawn and the transformed space constituting the support of the random variable. Transforms are not currently implemented for discrete random variables.
+
+All transforms have three core methods:
+
+* ``forward``: The map from a constrained space to the unconstrained space.
+* ``backward``: The inverse map from the unconstrained space to a constrained space.
+* ``log_jac_det``: The log of the determinant of the Jacobian of the ``backward`` map. This is used to account for the transformed random variable correctly in the posterior log-probability.
+
+.. note::
+   Transforms are principally intended for internal use and in most cases users do not need to change them. In particular, all continuous distributions on a constrained domain that are implemented in PyMC have a ``default_transform`` that will automatically transform the random variables as required without needing any extra work from the user.
+
+The main use-cases for setting custom transforms include the following:
+
+#. The ``default_transform`` may need to be replaced with an alternative transform on the same constained space. For example, the ``default_transform`` for positive-valued random variables is the :class:`log` transform but in some cases it may be advantageous to use the :class:`log_exp_m1` transform instead.
+#. The ``default_transform`` may be removed entirely in some cases when using non-HMC samplers.
+#. Exceptionally, transforms can be used to add constraints to the model specification. However this should not be viewed as a default use-case and, in practice, this is mostly limited to using :class:`ordered` in mixture models.
+
+   * NB: :class:`ordered` is not guaranteed to work correctly when used in combination with other transforms, such as :class:`simplex` and :class:`ZeroSumTransform`.
+
+.. warning::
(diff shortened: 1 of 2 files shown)


ricardoV94 (approved):
This explanation is absolutely incredible!

ricardoV94 (changes requested):
Just a small suggestion, as you mentioned we will have both `transform` and `default_transform` after: https://github.com/pymc-devs/pymc/pull/7207

All you said about `default_transform` is correct. When users want to add other constraints but not modify the default one, they should probably use the `transform` kwarg. Under the hood, we will build a ChainTransform, that applies the user provided transform on top of the default one.

mkusnetsov:
Thanks, @ricardoV94! That's a good point - I pushed some changes to clarify `transform` kwarg as distinct from `default_transform`.

But I might be doing something wrong? Although I can see the commits and the changes, the preview documentation is still showing the original version, not the revised one... (I know it takes a while to re-build but it seems to be saying "Read the Docs build succeeded!")

ricardoV94:
> But I might be doing something wrong? Although I can see the commits and the changes, the preview documentation is still showing the original version, not the revised one... (I know it takes a while to re-build but it seems to be saying "Read the Docs build succeeded!")

It takes some time for the preview to update. I think it's correct now? https://pymcio--7232.org.readthedocs.build/projects/docs/en/7232/api/distributions/transforms.html

ricardoV94 (approved):
I think this is perfect as is. I would just wait for the related PR to be merged, so that our docs are not "ahead of time"

mkusnetsov:
Thank you for the kind words! It is correct now, yes. Will know that it takes a while, in the future. And yes - it absolutely makes sense to wait for the other PR.

I left some [open questions](https://github.com/pymc-devs/pymc/issues/7040#issuecomment-2027815599) on the underlying issue. Are any of them worth pursuing either in a separate issue or a discussion? Just checking because I assume once that issue is closed they might get lost.

OriolAbril (approved):
I review docs related PRs quite sparsely if I have time and after seeing this one I wanted to add my praise too at use of rST, both at syntax level and conceptual level 😄 📚 

Also, in case someone wonders, the fact that references to pymc.sample, pymc.sample_prior_predictive... are not (yet) rendered as links in the rtd preview is an issue with how those are documented, once that is fixed all cross-references here will work

ricardoV94:
Thanks a ton @mkusnetsov, don't hesitate in opening issues/PRs for other areas you find documentation to be lacking

PR #7234, opened 2024-03-31
Error in idata.posterior variable

logongas:
Run the example in the line:
`idata.posterior["y_model"] = idata.posterior["Intercept"] + idata.posterior["x"] * xr.DataArray(x) `
show de error:

> KeyError: "No variable named 'x'. Variables on the dataset include ['chain', 'draw', 'Intercept', 'slope', 'sigma', 'y_model']"

## Description
change variable "x" by "slope"

## Type of change
- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7234.org.readthedocs.build/en/7234/

### DIFF
--- a/docs/source/learn/core_notebooks/GLM_linear.ipynb
+++ b/docs/source/learn/core_notebooks/GLM_linear.ipynb
@@ -416,7 +416,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "idata.posterior[\"y_model\"] = idata.posterior[\"Intercept\"] + idata.posterior[\"x\"] * xr.DataArray(x)"
+    "idata.posterior[\"y_model\"] = idata.posterior[\"Intercept\"] + idata.posterior[\"slope\"] * xr.DataArray(x)"
    ]
   },
   {


twiecki (approved):

twiecki (changes requested):
`[0;31mKeyError[0m: "No variable named 'slope'. Variables on the dataset include ['chain', 'draw', 'Intercept', 'x', 'y_sigma']"`

OriolAbril:
Hi @logongas, thanks for the PR but sadly it is not needed. I'll try to explain why.

All the notebooks in https://www.pymc.io/projects/docs/en/stable/learn/core_notebooks/index.html are executed with every commit made to pymc to make sure they run with each change. All builds have been able to execute the notebook correctly.

What I think is happening to you is that you are using the initial pymc model, the one defined as:

```python
with Model() as model:  # model specifications in PyMC are wrapped in a with-statement
    # Define priors
    sigma = HalfCauchy("sigma", beta=10)
    intercept = Normal("Intercept", 0, sigma=20)
    slope = Normal("slope", 0, sigma=20)

    # Define likelihood
    likelihood = Normal("y", mu=intercept + slope * x, sigma=sigma, observed=y)

    # Inference!
    # draw 3000 posterior samples using NUTS sampling
    idata = sample(3000)
```

in order to run the plots at the bottom. However, if you run the notebook sequentially you'll see that before getting to the plots, a roughly equivalent model is defined in bambi:

```python
model = bmb.Model("y ~ x", data)
idata = model.fit(draws=3000)
```

These two lines do basically the same as the code cell above, but not all the variable names are the same in both models (even if meaning the same conceptually), nor all priors are the same (bambi sets defaults for them), nor the samples will be exactly the same either even if the observed data is the same because MCMC has a random component.

If you do want to use the first model you need to change how `y_model` is defined, as you have done, but the notebook executes the plots on the bambi model. Don't hesitate to let us know if you have any more doubts.

logongas:
@OriolAbril Thank you very much for the reply. Your explanation is clear

PR #7239, opened 2024-04-05
Implement several RandomVariables as SymbolicRandomVariables

ricardoV94:
## Description

Implement several RandomVariables as SymbolicRandomVariables

This allows sampling from multiple backends without having to dispatch for each one

Also:
* Allow signature to handle rng and size arguments explicitly.
* Move rv_op method to the SymbolicRandomVariable class and get rid of dummy inputs logic (it was needed in previous versions of PyTensor)
* Allow dispatch methods without filtering of inputs for SymbolicRandomVariable distributions

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7239.org.readthedocs.build/en/7239/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -52,10 +52,12 @@ from pytensor.tensor.random.basic import (
     vonmises,
 )
 from pytensor.tensor.random.op import RandomVariable
+from pytensor.tensor.random.utils import normalize_size_param
 from pytensor.tensor.variable import TensorConstant
 
 from pymc.logprob.abstract import _logprob_helper
 from pymc.logprob.basic import icdf
+from pymc.pytensorf import normalize_rng_param
 
 try:
     from polyagamma import polyagamma_cdf, polyagamma_pdf, random_polyagamma
@@ -73,7 +75,6 @@ except ImportError:  # pragma: no cover
 
 from scipy import stats
 from scipy.interpolate import InterpolatedUnivariateSpline
-from scipy.special import expit
 
 from pymc.distributions import transforms
 from pymc.distributions.dist_math import (
@@ -90,8 +91,8 @@ from pymc.distributions.dist_math import (
     normal_lcdf,
     zvalue,
 )
-from pymc.distributions.distribution import DIST_PARAMETER_TYPES, Continuous
-from pymc.distributions.shape_utils import rv_size_is_none
+from pymc.distributions.distribution import DIST_PARAMETER_TYPES, Continuous, SymbolicRandomVariable
+from pymc.distributions.shape_utils import implicit_size_from_params, rv_size_is_none
 from pymc.distributions.transforms import _default_transform
 from pymc.math import invlogit, logdiffexp, logit
 
@@ -1236,20 +1237,28 @@ class Beta(UnitContinuous):
         )
 
 
-class KumaraswamyRV(RandomVariable):
+class KumaraswamyRV(SymbolicRandomVariable):
     name = "kumaraswamy"
-    ndim_supp = 0
-    ndims_params = [0, 0]
-    dtype = "floatX"
+    signature = "[rng],[size],(),()->[rng],()"
     _print_name = ("Kumaraswamy", "\\operatorname{Kumaraswamy}")
 
     @classmethod
-    def rng_fn(cls, rng, a, b, size) -> np.ndarray:
-        u = rng.uniform(size=size)
-        return np.asarray((1 - (1 - u) ** (1 / b)) ** (1 / a))
+    def rv_op(cls, a, b, *, size=None, rng=None):
+        a = pt.as_tensor(a)
+        b = pt.as_tensor(b)
+        rng = normalize_rng_param(rng)
+        size = normalize_size_param(size)
 
+        if rv_size_is_none(size):
+            size = implicit_size_from_params(a, b, ndims_params=cls.ndims_params)
 
-kumaraswamy = KumaraswamyRV()
+        next_rng, u = uniform(size=size, rng=rng).owner.outputs
+        draws = (1 - (1 - u) ** (1 / b)) ** (1 / a)
+
+        return cls(
+            inputs=[rng, size, a, b],
+            outputs=[next_rng, draws],
+        )(rng, size, a, b)
 
 
 class Kumaraswamy(UnitContinuous):
@@ -1296,13 +1305,11 @@ class Kumaraswamy(UnitContinuous):
         b > 0.
     """
 
-    rv_op = kumaraswamy
+    rv_type = KumaraswamyRV
+    rv_op = KumaraswamyRV.rv_op
 
     @classmethod
     def dist(cls, a: DIST_PARAMETER_TYPES, b: DIST_PARAMETER_TYPES, *args, **kwargs):
-        a = pt.as_tensor_variable(a)
-        b = pt.as_tensor_variable(b)
-
         return super().dist([a, b], *args, **kwargs)
 
     def support_point(rv, size, a, b):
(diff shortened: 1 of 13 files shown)


ricardoV94:
Marked as a draft because it includes commits from #6947

pymc/distributions/distribution.py
```diff
@@ -886,10 +963,10 @@ def _infer_final_signature(signature: str, n_inputs, n_updates) -> str:
             inputs, outputs = _parse_gufunc_signature(signature)
         if len(inputs) == n_inputs - 1:
```
ricardoV94: Should also add rngs to input signature

pymc/distributions/continuous.py
```diff
@@ -1236,20 +1235,33 @@ def icdf(value, alpha, beta):
...
+
+        if rng is None:
+            rng = pytensor.shared(np.random.default_rng())
 
+        if size is not None:
+            size = pt.as_tensor(size, dtype="int64", ndim=1)
 
-kumaraswamy = KumaraswamyRV()
+        if rv_size_is_none(size):
+            size = pt.as_tensor(pt.broadcast_shape(a, b))
```
jessegrabowski: This should be a helper function -- `_setup_rng(rng, size, *params)` ?

pymc/distributions/distribution.py
```diff
@@ -185,30 +185,56 @@ def _random(*args, **kwargs):
...
             if class_logp:
 
                 @_logprob.register(rv_type)
                 def logp(op, values, *dist_params, **kwargs):
-                    dist_params = dist_params[3:]
-                    (value,) = values
+                    if isinstance(op, RandomVariable):
+                        rng, size, dtype, *dist_params = dist_params
+                    elif params_idxs:
+                        dist_params = [dist_params[i] for i in params_idxs]
```
jessegrabowski, line 211: Same comment; duplicated code

pymc/distributions/distribution.py
```diff
@@ -267,7 +298,10 @@ class SymbolicRandomVariable(OpFromGraph):
     """Number of core dimensions of the distribution's parameters."""
 
     signature: str = None
-    """Numpy-like vectorized signature of the distribution."""
+    """Numpy-like vectorized signature of the distribution.
+
+    It allows tokens [rng], [size] to identify the special inputs.
```
jessegrabowski, line 307: Maybe some examples of what a signature should look like?

pymc/distributions/timeseries.py
```diff
@@ -967,55 +998,6 @@ def dist(cls, dt, sde_fn, sde_pars, *, init_dist=None, steps=None, **kwargs):
...
-            batch_size = size
-        else:
-            batch_size = pt.broadcast_shape(*sde_pars, init_dist)
-        init_dist = change_dist_size(init_dist, batch_size)
-
-        # Create OpFromGraph representing random draws from SDE process
-        # Variables with underscore suffix are dummy inputs into the OpFromGraph
-        init_ = init_dist.type()
-        sde_pars_ = [x.type() for x in sde_pars]
-        steps_ = steps.type()
```
jessegrabowski, line 983: Is this `type` thing not required anymore? I remember having to do that when I wrote distributions.

pymc/distributions/truncated.py
```diff
@@ -71,6 +76,138 @@ def __init__(
...
+                truncated_rv = pt.set_subtensor(
+                    truncated_rv[reject_draws],
+                    new_truncated_rv[reject_draws],
+                )
+            reject_draws = pt.or_((truncated_rv < lower), (truncated_rv > upper))
+
+            return (
+                (truncated_rv, reject_draws),
+                collect_default_updates(new_truncated_rv),
+                until(~pt.any(reject_draws)),
```
jessegrabowski, line 156: Does `~` work as expected on tensor variables? I always thought you had to use `pt.bitwise_not`

jessegrabowski (approved):
Seems like this is mostly shuffling around and simplifying stuff that already exists, so if tests pass it looks great. I'm a bit unclear on how the signatures for RVs work, was that changed in this PR or another one? The square brackets in particular throw me off. There's also some opportunities for refactoring duplicated code.

pymc/distributions/timeseries.py
ricardoV94: That was all a side effect of not being able to use shared variables and constants as inputs. No longer needed

pymc/distributions/truncated.py
ricardoV94: I think only eq and neq don't work

ricardoV94:
> Seems like this is mostly shuffling around and simplifying stuff that already exists, so if tests pass it looks great.

There are two commits, in this PR, the first one shuffles and refactors things so that the second commit (converting existing RandomVariables to SymbolicRandomVariables is done mostly without hassle)

> I'm a bit unclear on how the signatures for RVs work, was that changed in this PR or another one? The square brackets in particular throw me off. 

I think this is the kind of signature we should use in PyTensor, so I was testing it here. The numpy vectorize signature doesn't really handle stuff like rng, size, axis, so I think we should take the initative here. In a previous PR I pretend those were scalars, so stuff looked like `(),(),(),()->(),()`, which is both fake and I think less readable than `[rng],[size],(),()->[rng],()`. Wdyt? This would ultimately replace the `ndim_supp` and `ndims_params` which is very strict and only works because we assume all RVs have the same non-numerical signature. This is not useful for SymbolicRandomVariables that can have multiple (or zero) RNGs and only optionally size.

> There's also some opportunities for refactoring duplicated code.

I'm weary of some helpers, but I can optimize for a bit less DRY

ricardoV94:
@jessegrabowski I simplified the logic needed to build standard SymbolicRandomVariables, what do you think?

pymc/distributions/distribution.py
ricardoV94: I'm okay with this repetition

pymc/distributions/distribution.py
```diff
@@ -279,24 +308,70 @@ class SymbolicRandomVariable(OpFromGraph):
     _print_name: tuple[str, str] = ("Unknown", "\\operatorname{Unknown}")
     """Tuple of (name, latex name) used for for pretty-printing variables of this type"""
 
+    @classmethod
+    def _params_signature(cls):
+        inputs_signature, outputs_signature = cls.signature.split("->")
+        inputs_params_signature = [
+            inp_sig for inp_sig in inputs_signature.split(",") if inp_sig not in ("[rng]", "[size]")
```
jessegrabowski: Maybe a global `SPECIAL_TOKENS = ["[rng]", "[size]"]` in case you want to add/remove/modify these later?

jessegrabowski: I think this whole section is just to remove the special tokens before passing it to `_parse_gufunc_signature` yeah? Couldn't this be done more transparently with a regex, like `core_signature = re.sub(special_token_pattern, '', cls.signature)`?

ricardoV94: Cross that bridge when/if we get there?

pymc/distributions/distribution.py
```diff
@@ -244,6 +275,21 @@ def fn(*args, **kwargs):
...
+    which may be provided either as a class attribute or as an instance attribute.
+
+    Adapted from https://stackoverflow.com/a/28238047
+    """
+
+    def __get__(self, instance, type_):
+        descr_get = super().__get__ if instance is None else self.__func__.__get__
+        return descr_get(instance, type_)
+
+
```
ricardoV94: Some bit of blackmagic to allow dynamic properties from the signature, to work both when signatures are class attributes or only instance attributes. Let me know if you see a better/ less magical solution.

pymc/distributions/distribution.py
jessegrabowski: sure

pymc/distributions/distribution.py
jessegrabowski: Why is it needed though?

ricardoV94: Sometimes we have the signature fixed in the class (DM), sometimes only known at runtime (Mixture). This allows us to parse ndims_params, ndim_supp, default_outputs from the signature regardless of when it's defined.

pymc/distributions/distribution.py
ricardoV94: > I think this whole section is just to remove the special tokens before passing it to _parse_gufunc_signature yeah? Couldn't this be done more transparently with a regex, like core_signature = re.sub(special_token_pattern, '', cls.signature)?

Needs to be able to get rid of the optional/useless commas, but otherwise I'm not against regex, just not my zone.

ricardoV94: Done the regex way

pymc/distributions/distribution.py
jessegrabowski: Would it be better to have a class_signature (always defined) and a instance_signature (default None), and drop back to the class_signature if instance_signature is None?

ricardoV94: A class signature cannot always be defined. The point is I want to extract `ndim_supp` from either a class signature (without having to initiate said class) or, if it cannot be known in advance, from an instance signature. I need a static class method or instance method that can read either the class or instance signature, this achieves exactly that.

ricardoV94: Do I really need this? Not so much. Alternatives are:
1. Hard code ndim_supp/ndims_params/default_output when they can be known in advance. It's just redundant/error prone because all info exists in the signature.
2. Don't provide any of these until the Op is defined. Drawback? I can't use these lines as straightfoward:

We have to create dummy Ops with empty size just just to find the ndim_supp:
https://github.com/pymc-devs/pymc/blob/5f95fc292adebbeb31b6de825b7b407042e382fd/pymc/distributions/distribution.py#L486-L489

I can't write classmethods of the new SymbolicRandomVariables like this:
```python
        if rv_size_is_none(size):
            size = implicit_size_from_params(b, kappa, mu, ndims_params=cls.ndims_params)
```
Because cls.ndims_params doesn't exist until we create the `Op` below. 

I could obviously parse ndims_params/ndim_supp on the spot from the signature. It's just a small convenience. But OTOH it's also just 5 lines of Python

ricardoV94: @jessegrabowski My last commit reverts the magic solution for the cleanest alternative I can think of. Which one do you prefer?

ricardoV94: I removed the non-magic commit, for reference it's here: https://github.com/pymc-devs/pymc/commit/20e6e3ba23b27020ce99a603394d8b67b8fef64a

PR #7241, opened 2024-04-06
Register the overloads added by CustomDist in worker processes

### LINKED ISSUE #7224: BUG: `pymc.sample_smc` fails with `pymc.CustomDist`
### Describe the issue:

`pymc.sample_smc` raises a `NotImplementedError` due to a missing logp method if a `pymc.CustomDist` is used in a model without `dist` argument. In addition to using `dist`, switching to `pm.Potential` works.

```python
import pymc as pm
import numpy as np


def _logp(value, mu, sigma):
    dist = pm.Normal.dist(mu=mu, sigma=sigma)

    return pm.logp(dist, value)


def _random(mu, sigma, rng, size):
    if rng is None:
        rng = np.random.default_rng()
    sample = rng.normal(loc=mu, scale=sigma, size=size)

    return sample


def _logcdf(value, mu, sigma):
    dist = pm.Normal.dist(mu=mu, sigma=sigma)

    return pm.logcdf(dist, value)


def _dist(mu, sigma, size):
    return pm.Normal.dist(mu, sigma, size=size)


def main():
    data = np.random.default_rng().normal(5, 2, 1000)

    with pm.Model():
        mu = pm.Normal("mu", mu=0, sigma=10)
        sigma = pm.HalfNormal("sigma", sigma=10)
        pm.CustomDist(
            "y",
            mu,
            sigma,
            logp=_logp,
            random=_random,
            logcdf=_logcdf,
            observed=data,
        )
        sample = pm.sample_smc()  # NotImplementedError

    with pm.Model():
        pm.CustomDist(
            "y",
            2,
            10,
            logp=_logp,
            random=_random,
            logcdf=_logcdf,
        )
        sample = pm.sample_smc()  # NotImplementedError

    with pm.Model():
        mu = pm.Normal("mu", mu=0, sigma=10)
        sigma = pm.HalfNormal("sigma", sigma=10)
        pm.CustomDist(
            "y",
            mu,
            sigma,
            dist=_dist,
            observed=data,
        )
        sample = pm.sample_smc()  # Works

    with pm.Model():
        mu = pm.Normal("mu", mu=0, sigma=10)
        sigma = pm.HalfNormal("sigma", sigma=10)
        pm.Potential(
            "y",
            _logp(data, mu, sigma),
        )
        sample = pm.sample_smc()  # Works


if __name__ == "__main__":
    main()
```

### Error message:

<details>

```shell
multiprocessing.pool.RemoteTraceback: 
"""
  File "\envs\pymc\Lib\multiprocessing\pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\multiprocessing\pool.py", line 51, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
  File "\envs\pymc\Lib\site-packages\pymc\smc\sampling.py", line 421, in _apply_args_and_kwargs
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\site-packages\pymc\smc\sampling.py", line 344, in _sample_smc_int
    smc._initialize_kernel()
  File "\envs\pymc\Lib\site-packages\pymc\smc\kernels.py", line 239, in _initialize_kernel
    initial_point, [self.model.varlogp], self.variables, shared
                    ^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\site-packages\pymc\model\core.py", line 832, in varlogp
    return self.logp(vars=self.free_RVs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\site-packages\pymc\model\core.py", line 717, in logp
    rv_logps = transformed_conditional_logp(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\site-packages\pymc\logprob\basic.py", line 612, in transformed_conditional_logp
    temp_logp_terms = conditional_logp(
                      ^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\site-packages\pymc\logprob\basic.py", line 542, in conditional_logp
    q_logprob_vars = _logprob(
                     ^^^^^^^^^
  File "\envs\pymc\Lib\functools.py", line 909, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\site-packages\pymc\logprob\abstract.py", line 63, in _logprob
    raise NotImplementedError(f"Logprob method not implemented for {op}")
NotImplementedError: Logprob method not implemented for CustomDist_y_rv{0, (0, 0), floatX, False}
"""

The above exception was the direct cause of the following exception:

  File "issue.py", line 79, in <module>
    main()
  File "issue.py", line 44, in main
    sample = pm.sample_smc()  # Exception has occurred: NotImplementedError
             ^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\site-packages\pymc\smc\sampling.py", line 213, in sample_smc
    results = run_chains_parallel(
              ^^^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\site-packages\pymc\smc\sampling.py", line 390, in run_chains_parallel
    results = _starmap_with_kwargs(
              ^^^^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\site-packages\pymc\smc\sampling.py", line 417, in _starmap_with_kwargs
    return pool.starmap(_apply_args_and_kwargs, args_for_starmap)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\multiprocessing\pool.py", line 375, in starmap
    return self._map_async(func, iterable, starmapstar, chunksize).get()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\envs\pymc\Lib\multiprocessing\pool.py", line 774, in get
    raise self._value
NotImplementedError: Logprob method not implemented for CustomDist_y_rv{0, (0, 0), floatX, False}
```

</details>

### PyMC version information:

<details>

```
Python 3.11.7
pymc 5.10.0
pytensor 2.18.6
Win 10
Environment set up via conda but updated pymc and pytensor with pip
```
Also fails with these environments
[conda_env.txt](https://github.com/pymc-devs/pymc/files/15465575/conda_env.txt)
[conda_env_dev.txt](https://github.com/pymc-devs/pymc/files/15465576/conda_env_dev.txt)

</details>

### Context for the issue:

I'm testing a model which suffers from slow sampling, possibly due to expensive gradient calculations. I tested SMC as a possible solution [as suggested on the forums](https://discourse.pymc.io/t/implementing-rounding-by-manual-integration-more-efficiently/13807/10) but got this error message.

Using the `dist` argument could work in most cases, but there's cases when the distributions provided by `pymc` are not enough. Using `pm.Potential` could help with sampling but that would in turn make forward sampling less straightforward.

EliasRas:
## Description
Currently `sample_smc` can fail due to a `NotImplementedError` if it's used with a model defined using`CustomDist`. If a `CustomDist` is used without `dist` parameter, the overloads for `_logprob`, `_logcdf` and `_support_point` are registered only in the main process.

This PR adds an initializer which registers the overloads in the worker processes of the pool used in `sample_smc`.

## Related Issue
- [x] Closes #7224
- [ ] Related to #

## Checklist

## Type of change
- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7241.org.readthedocs.build/en/7241/


pymc/smc/sampling.py
```diff
@@ -406,3 +415,25 @@ def run_chains(chains, progressbar, params, random_seed, kernel_kwargs, cores):
                         progress.update(status=f"Stage: {stage} Beta: {beta:.3f}", task_id=task_id)
 
         return tuple(cloudpickle.loads(r.result()) for r in futures)
+
+
+def _find_custom_methods(model):
+    custom_methods = {}
+    for rv in model.free_RVs + model.observed_RVs:
+        cls = rv.owner.op.__class__
+        if hasattr(cls, "_random_fn"):
```
EliasRas: The way I'm checking if the rv is from `CustomDist` feels kind dirty. Is there a better way? I couldn't come up with anything better after `_print_name` was changed in aa679f3

ricardoV94:
Hi @EliasRas, I'll need some time to review this one properly. Thanks for taking the initiative

EliasRas:
Looks like I messed up by rebasing instead of merging and introduced plenty of unnecessary commits to this feature. Does it need to be fixed?

twiecki:
Yes, that needs to be fixed, happens to everyone. One approach is to start clean and cherry-pick your commits.

pymc/smc/sampling.py
ricardoV94: You can check that `rv.owner.op` is an instance of `CustomDistRV` (or whatever it is called)

tests/smc/test_smc.py
```diff
@@ -133,6 +133,18 @@ def test_unobserved_categorical(self):
...
 
+    def test_parallel_custom(self):
+        def _logp(value, mu):
+            return -((value - mu) ** 2)
+
+        def _random(mu, rng=None, size=None):
+            return rng.normal(loc=mu, scale=1, size=size)
+
+        with pm.Model():
+            mu = pm.CustomDist("mu", 0, logp=_logp, random=_random)
```
ricardoV94: There are two kinds of CustomDist, if instead of `random` you pass `dist` you get a different Op type, that I guess would still fail after this PR

Edit: I see you mentioned this in your top message

pymc/smc/sampling.py
```diff
@@ -414,3 +423,25 @@ def run_chains(chains, progressbar, params, random_seed, kernel_kwargs, cores):
                         )
 
         return tuple(cloudpickle.loads(r.result()) for r in futures)
+
+
+def _find_custom_methods(model):
+    custom_methods = {}
+    for rv in model.free_RVs + model.observed_RVs:
```
ricardoV94:
```suggestion
    for rv in model.basic_RVs:
```

ricardoV94:
I think it's more complicated than this. The following example has specific dispatch, but no RV that shows up in the graph:

```python
import pymc as pm

def _logp(value, mu):
    return -((value - mu) ** 2)

def _dist(mu, size=None):
    return pm.Normal.dist(mu, 1, size=size)

with pm.Model():
    mu = pm.Normal("mu", 0)
    pm.Potential("term", pm.logp(pm.CustomDist.dist(mu, logp=_logp, dist=_dist), [1, 2]))
    pm.sample_smc(draws=6, cores=1)        
 ```

File: tests/smc/test_smc.py (hunk 2 above)
Comment (ricardoV94) [reply tests/smc/test_smc.py]: Why does it fail with `random` but not `dist`. Testing locally it seems to fail with both for me?

Comment (EliasRas) [reply tests/smc/test_smc.py]: I just tested both the code I used in the linked issue and your example below. I still get no errors using `pm.Potential` or `dist` argument. Originally I was using `pymc==5.10.0` but now I tested with `pymc==5.13.0`. Maybe there are differences between Windows and Linux if you're using one?

I'm not completely sure why using `dist` works for me but, based on some quick testing, `DistributionMeta.__new__` is called when e.g. `Normal` is defined and the overloads for builtin distributions are registered there. I'm not well versed in multiprocessing or the way that Python does importing but my hunch is that the worker processes automatically import stuff from `pymc` and the overloads get registered as a side effect. For user-defined `logprob` etc. this is not the case since the registration isn't done during importing.

**Comment by EliasRas** [comment]
> It also fails even with a single core

22e8f0bb4a02d874856438065efa7b3ef2645e13 did refactoring for `sample_smc` and I think that errors should now pop up even with single core since the sampling is always done in another process. Previously this was the case only when `cores>1` since there were separate `run_chains_parallel` and `run_chains_sequential`.

**Comment by ricardoV94** [comment]
Somehow I am getting `ConnectionResetError: [Errno 104] Connection reset by peer` even for unrelated models without any sort of CustomDist

File: tests/smc/test_smc.py (hunk 2 above)
Comment (ricardoV94) [inline tests/smc/test_smc.py]: This way we test both kinds of CustomDist with customized dispatched methods. You can use the `_dist` from my example

```suggestion
            mu = pm.CustomDist("mu", 0, logp=_logp, dist=_dist)
```

ricardoV94:
> Somehow, in main, I am getting `ConnectionResetError: [Errno 104] Connection reset by peer` even for unrelated models without any sort of CustomDist

Okay it's something about the new progressbal and pycharm interactive python console. If I use from ipython/terminal it works. But also works in main for me?

ricardoV94:
I cannot reproduce a failure with your test locally (after avoding the pycharm issue) nor in a Colab environment: https://colab.research.google.com/drive/1I1n6c9IlmXknIfhxC5s7sAQghv0vfRSY?usp=sharing

Can you share more details about your environment/setup?

EliasRas:
> Can you share more details about your environment/setup?

I added the output of `conda list` to "PyMC version information" section of #7224. I'm running the code using VSCode if that matters. Do you need anything else?

Basically I followed the install instructions and [the pull request tutorial](https://www.pymc.io/projects/docs/en/stable/contributing/pr_tutorial.html#pr-tutorial) when installing. Might have also `pip install`ed a couple of extra packages here and there.

ricardoV94:
> I added the output of conda list to "PyMC version information" section of https://github.com/pymc-devs/pymc/issues/7224. I'm running the code using VSCode if that matters. Do you need anything else?

We should have at least one person reproduce the problem because I cannot. It may be a VSCode environment issue. Ideally we wouldn't have to change the codebase

EliasRas:
The test does fail without the changes when I run it from miniforge prompt though.

EliasRas:
Is there anything that needs to be done here besides running the tests?

twiecki:
> Is there anything that needs to be done here besides running the tests?

Sorry for the delay, just kicked off tests.

lucianopaz:
Thanks @EliasRas, I haven't been able to reproduce this yet but that's just because I'm in the middle of switching workstations and haven't gotten everything setup yet.
Your fix looks fine to me and I understand what you identified as the cause of the issue: the dispatching mechanism isn't registering the logp and other methods to the dynamically created class. I think that this highlights a caveat in `pymc`'s and `pytensor`'s design: **spawned processes may not have all the registered dispatch signatures as the main process**. I imagine that this is mostly a problem on Windows, where multiprocessing can only spawn new processes whereas linux based systems will default to forks which in principle should copy over the memory contents of the main processes. I'm not sure what will happen under MacOS because I think that they cannot use fork multiprocessing for some reason either.
With this design caveat in hand, I'm not sure if it's better to have a package level utility function that serves as a sort of book-keeper or something that can handle communicating the extra dispatch registration needed to ensure that child processes will use the correct dispatching functions. I'm curious to know what @ricardoV94 thinks about this. I don't think that this PR should have to tackle this kind of work, but I think that we can discuss if it's necessary here, and maybe later open an issue and a separate PR (also maybe in `pytensor` where dispatching is used for transpilation/compilation and maybe at some point for lazy gradients?).

lucianopaz:
> I guess the underlying reason for the failure is that pickling of `DensityDist` doesn't work out of the box?

I don’t think the problem is about pickling. The `DensityDist` end up returning an op that can be cloudpickled. If I recall correctly it can’t be pickled because the op class is created on the fly. In the process of creating the op, the dispatchers get populated with the callables that are supplied as inputs to the distribution class. As far as I understand, those functions are detached from the rv op and that’s why they never get populated on a spawned process.

aseyboldt:
I don't mean that the pickling itself throws an error (it doesn't), but that it would be the responsibility of the `DensityDist` object to ensure that the set-up it needs (ie registering the logp) is done when it is unpickled.

For instance the following fails with the `NotImplementedError`, and has nothing to do with smc, so I guess the solution shouldn't be specific to smc?

```python
import pymc as pm
import cloudpickle
import multiprocessing


def use_logp_func(pickled_model):
    model = cloudpickle.loads(pickled_model)
    logp = model.logp()
    func = pm.pytensorf.compile_pymc(model.value_vars, logp)
    print(func(1.0))


if __name__ == "__main__":
    with pm.Model() as model:

        def logp(value):
            return -(value**2)

        pm.DensityDist("x", logp=logp)

    logp = model.logp()
    func = pm.pytensorf.compile_pymc(model.value_vars, logp)
    pickled_model = cloudpickle.dumps(model)

    ctx = multiprocessing.get_context("spawn")
    process = ctx.Process(target=use_logp_func, args=(pickled_model,))
    process.start()
    process.join()
```

lucianopaz:
I completely agree that this problem isn’t unique to smc and is a design caveat that needs to be addressed more comprehensively.
I think that we can kind of patch some things:
1. Make `Model` objects `__setstate__` and `__getstate__` repopulate the dispatch registries
2. Get `CustomDist` rv ops to have these methods defined somehow (maybe clojures) that repopulates the dispatch registries.

I’m not sure if these two methods can cover all use patterns though.

ricardoV94:
Alternatively we could pass the functions needed to each process which is more like what `pm.sample ` does. 

This also avoids recompiling the same function multiple times?

EliasRas:
@lucianopaz 
Point 1. is pretty straightforward but could you explain what you meant by 2.? How would it be different from overriding `__getstate__` and `__setstate__`?

EliasRas:
I started work on `__getstate__` and `__setstate__` but I realized that I can't just copy the current implementation due to circular imports. Would either of these sound like a good idea?

1. Create `utils` module (or something similar) in `pymc.model` that handles the custom methods and registering them.
2. Add a method for RVs that registers the custom methods if necessary. `CustomDistRV` and `CustomSymbolicDistRV` would have the registrations and `RandomVariable` and `SymbolicRandomVariable` would just pass.

ricardoV94:
I think we should explore an alternative where we compile the functions SMC needs and fork afterwards like pm.sample does. 

This approach seems more brittle?

EliasRas:
I agree on the re-compiling part but shouldn't this still be fixed? It feels like an arbitrary decision to "disallow" using multiprocessing this way only on Windows.

ricardoV94:
Using the class that's being dispatched to register the dispatches during pickling seems at odds with the point of dispatching. The class shouldn't have to know what's being dispatched upon.

For instance we also have icdf methods, what if someone dispatched on it from the outside, does pickling work for it? Or would the setstate/getstate need to know about icdf (as well any other dispatch that may not even be part of PyMC)?

EliasRas:
Thanks for the explanation. I'll start working on the compilation approach.

ricardoV94:
However since this fixes existing behavior I think we can go ahead and merge it as a temporary patch?

pymc/smc/sampling.py
```diff
@@ -383,11 +386,18 @@ def run_chains(chains, progressbar, params, random_seed, kernel_kwargs, cores):
             # main process and our worker functions
             _progress = manager.dict()
 
+            # check if model contains CustomDistributions defined without dist argument
+            custom_methods = _find_custom_methods(params[3])
```
ricardoV94:
```suggestion
            custom_dist_dispatch_methods = _find_custom_dist_dispatch_methods(params[3])
```

pymc/smc/sampling.py
```diff
@@ -420,3 +430,29 @@ def run_chains(chains, progressbar, params, random_seed, kernel_kwargs, cores):
                         )
 
         return tuple(cloudpickle.loads(r.result()) for r in done)
+
+
+def _find_custom_methods(model):
+    custom_methods = {}
+    for rv in model.basic_RVs:
+        rv_type = rv.owner.op
+        cls = rv_type.__class__
```
ricardoV94: Does this also work for the registry key?
```suggestion
        cls = type(rv_type)
```

If not, what's class exactly?

pymc/smc/sampling.py
```diff
@@ -420,3 +430,29 @@ def run_chains(chains, progressbar, params, random_seed, kernel_kwargs, cores):
...
+        if isinstance(rv_type, CustomDistRV | CustomSymbolicDistRV):
+            custom_methods[cloudpickle.dumps(cls)] = (
+                cloudpickle.dumps(_logprob.registry.get(cls, None)),
+                cloudpickle.dumps(_logcdf.registry.get(cls, None)),
+                cloudpickle.dumps(_support_point.registry.get(cls, None)),
```
ricardoV94, line 445: Include icdf just in case?

ricardoV94 (approved):
Can you open a follow up issue for the compiling once and then forking?

EliasRas:
> I started working on compiling the functions in the main process. Should I close this or is there any chance of this getting merged before the newer PR?

@ricardoV94 @lucianopaz @aseyboldt @twiecki Just following up on this. Should I close the PR?

ricardoV94:
@EliasRas I had approved this PR but didn't merge. Apologies, did it just now

PR #7248, opened 2024-04-12
Allow freezing only subset of data and dims

ricardoV94:
## Description

This improves the optimization introduced on #7047, to allow freezing a subset of dims and data
It also fixes a bug where the dim_lengths were not returned

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7248.org.readthedocs.build/en/7248/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1050,7 +1050,14 @@ class Model(WithMemoization, metaclass=ContextMeta):
                     expected=new_length,
                 )
             self._coords[name] = tuple(coord_values)
-        self.dim_lengths[name].set_value(new_length)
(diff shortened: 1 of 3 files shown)


ricardoV94:
One day mypy will be useful, today is not that day

zaxtax (commented):
We should have a test that the original model can still be changed. Otherwise looks fine to me!

ricardoV94:
> We should have a test that the original model can still be changed. Otherwise looks fine to me!

Added that check to the first test

zaxtax (approved):
LGTM

PR #7252, opened 2024-04-13
Skewed Student-T distribution

fonnesbeck:
## Description

Added skewed Student T distribution (Jones and Faddy implementation). 

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7252.org.readthedocs.build/en/7252/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1981,11 +1981,7 @@ class SkewStudentT(Continuous):
 
         a_ = (a + 0.5) * pt.log(1 + x / pt.sqrt(a + b + x**2))
         b_ = (b + 0.5) * pt.log(1 - x / pt.sqrt(a + b + x**2))
-        c = (
-            (a + b - 1) * pt.log(2)
-            + (pt.gammaln(a) + pt.gammaln(b) - pt.gammaln(a + b))
-            + 0.5 * pt.log(a + b)
-        )
+        c = (a + b - 1) * pt.log(2) + pt.special.betaln(a, b) + 0.5 * pt.log(a + b)
 
         res = a_ + b_ - c
 
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1981,7 +1981,11 @@ class SkewStudentT(Continuous):
 
         a_ = (a + 0.5) * pt.log(1 + x / pt.sqrt(a + b + x**2))
         b_ = (b + 0.5) * pt.log(1 - x / pt.sqrt(a + b + x**2))
-        c = (a + b - 1) * pt.log(2) + pt.log(pt.betainc(a, b)) + 0.5 * pt.log(a + b)
+        c = (
+            (a + b - 1) * pt.log(2)
+            + (pt.gammaln(a) + pt.gammaln(b) - pt.gammaln(a + b))
+            + 0.5 * pt.log(a + b)
+        )
 
         res = a_ + b_ - c
 
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -1995,7 +1995,7 @@ class SkewStudentT(Continuous):
         b_ = (b + 0.5) * pt.log(1 - x / pt.sqrt(a + b + x**2))
         c = (a + b - 1) * pt.log(2) + pt.special.betaln(a, b) + 0.5 * pt.log(a + b)
 
-        res = a_ + b_ - c
+        res = a_ + b_ - c - pt.log(sigma)
 
         return check_parameters(
             res,
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -29,6 +29,7 @@ import pytensor.tensor as pt
 from pytensor.graph.basic import Apply, Variable
 from pytensor.graph.op import Op
 from pytensor.raise_op import Assert
+from pytensor.tensor import gamma as gammafn
 from pytensor.tensor import gammaln
 from pytensor.tensor.extra_ops import broadcast_shape
 from pytensor.tensor.math import betaincinv, gammaincinv, tanh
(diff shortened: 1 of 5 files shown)


ricardoV94:
Add in pymc-experimental first?

jessegrabowski:
If something like this goes in experimental, there should be a clear roadmap to go from there back to the main repo. This is a straight-forward distribution that (I guess?) can't be easily implemented as a generative graph using only existing distributions. If it's in pmx, what's the criteria to move it over? It's not something like statespace or marginalmodel that proposes a huge new functionality, or something like r2d2m2 that implements a distribution that doesn't fit neatly into a typical pymc model as we present them in books/tutorials.

ricardoV94:
I don't think this one has a straightforward generative graph

ricardoV94:
I assumed something "according to x and y" was a bit experimental, hence my suggestion

tests/distributions/test_continuous.py
```diff
@@ -549,6 +549,20 @@ def test_studentt_logp(self):
...
         )
 
+    def test_skewstudentt_logp(self):
+        check_logp(
+            pm.SkewStudentT,
+            R,
+            {"a": Rplus, "b": Rplus, "mu": R, "lam": Rplus},
+            lambda value, a, b, mu, lam: st.t.logpdf(value, a, b, mu, lam**-0.5),
+        )
+        check_logp(
```
ricardoV94: No need to test logp/logcdf twice just for the sigma/tau parametrization. These are the most expensive tests 

The helper to test RVs directly (which is missing) is better for just making sure tau gets converted into the expected sigma

Also missing the support_point test

fonnesbeck:
The "Following Jones and Faddy 2003" was to distinguish it from the range of available skewed Student T implementations. I chose this one because it is available in SciPy and relatively straightforward to implement.

fonnesbeck:
Jax failure does not seem to be related to this PR

ricardoV94:
But this one is: https://github.com/pymc-devs/pymc/actions/runs/8682226852/job/23806440256?pr=7252#step:7:616

You should try running those logp/logcdf/icdf with `n_samples=-1` or whatever it is, to test all combinations locally instead of only a random subset

ricardoV94:
You're missing the tests for the RV itself. That's also where we cover alternative parametrizations. Something like these

https://github.com/pymc-devs/pymc/blob/34c2d31484143b7b5c6130927ed4efde93a27a0f/tests/distributions/test_continuous.py#L2179-L2200

pymc/distributions/continuous.py
```diff
@@ -1896,6 +1898,140 @@ def icdf(value, nu, mu, sigma):
...
+    def logp(value, a, b, mu, sigma):
+        _, sigma = get_tau_sigma(sigma=sigma)
+
+        x = (value - mu) / sigma
+
+        a_ = (a + 0.5) * pt.log(1 + x / pt.sqrt(a + b + x**2))
+        b_ = (b + 0.5) * pt.log(1 - x / pt.sqrt(a + b + x**2))
+        c = (
+            (a + b - 1) * pt.log(2)
+            + (pt.gammaln(a) + pt.gammaln(b) - pt.gammaln(a + b))
```
aloctavodia: nitpick use the betaln function.

fonnesbeck: There is currently no `betaln` in `pytensor.tensor` AFAIK

aloctavodia: It is under the special module. `pt.special.betaln`

pymc/distributions/continuous.py
```diff
@@ -1908,6 +1910,136 @@ def icdf(value, nu, mu, sigma):
...
+class SkewStudentTRV(RandomVariable):
+    name = "skewstudentt"
+    ndim_supp = 0
+    ndims_params = [0, 0, 0, 0]
+    dtype = "floatX"
+    _print_name = ("SkewStudentT", "\\operatorname{SkewStudentT}")
+
+    @classmethod
+    def rng_fn(cls, rng, a, b, mu, sigma, size=None) -> np.ndarray:
+        return np.asarray(stats.jf_skew_t.rvs(a=a, b=b, size=size, random_state=rng)) * sigma + mu
```
ricardoV94: Can't sigma and mu be passed as arguments to the rvs? 

The way you wrote it has a similar bug that was addressed by this recent bugfix, when there are batched sigma and mu and size is None: https://github.com/pymc-devs/pymc/pull/7288

ricardoV94 (changes requested):

pymc/distributions/continuous.py
ricardoV94: You marked as resolved but I still see the old code?

fonnesbeck: Sorry, I had forgotten to push.

ricardoV94 (approved):

michaelosthege:
Looks like `test_skewstudentt_logp` is failing on main, or is at least flaky?

ricardoV94:
These tests run a subset of combinations. We should set `n_samples=-1` locally and see if every combination passes (both float64 and float32), I guess it does not

michaelosthege:
Running float64 this gives:

```
E           AssertionError: 
E           Arrays are not almost equal to 6 decimals
E           {'a': array(0.01), 'b': array(100.), 'mu': array(-2.1), 'sigma': array(0.01), 'value': array(-0.01)}
E           x and y -inf location mismatch:
E            x: array(-751.339449)
E            y: array(-inf)
```

ricardoV94:
Which underflows? PyMC or Scipy?

michaelosthege:
> Which underflows? PyMC or Scipy?

SciPy produces the `-inf`.

What's the fix? I have it open and can push a branch real quick (if it's simple)

ricardoV94:
> > Which underflows? PyMC or Scipy?
> 
> SciPy produces the `-inf`.
> 
> What's the fix? I have it open and can push a branch real quick (if it's simple)

Try slightly less extreme parameter domains

PR #7261, opened 2024-04-16
Fix gradient bug in models with max operation

### LINKED ISSUE #7251: `pt.max` is not differentiable in PyMC models
### Description

The following `model.dlogp()` raises a `NotImplemented` error:

```
with pm.Model() as model:
    x = pm.Normal('x', shape=(2,))
    mu = x.max()
    pm.Normal('obs', mu, observed=np.random.uniform())
```

This should be differentiable (`pt.max` has gradients implemented), so it seems like something is going wrong in rewrites, either with respect to logp rewrites or with respect to `MaxAndArgmax`

ricardoV94:
## Related Issue

- [x] Closes #7251 

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7261.org.readthedocs.build/en/7261/

### DIFF
--- a/pymc/logprob/rewriting.py
+++ b/pymc/logprob/rewriting.py
@@ -365,8 +365,6 @@ logprob_rewrites_db.register(
     "local_exp_over_1_plus_exp", out2in(local_exp_over_1_plus_exp), "basic"
 )
 logprob_rewrites_db.register("pre-canonicalize", optdb.query("+canonicalize"), "basic")
-# Split max_and_argmax
-logprob_rewrites_db.register("local_max_and_argmax", out2in(local_max_and_argmax), "basic")
 
 # These rewrites convert un-measurable variables into their measurable forms,
 # but they need to be reapplied, because some of the measurable forms require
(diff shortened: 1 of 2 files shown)


pymc/logprob/rewriting.py
```diff
@@ -376,6 +374,12 @@ def incsubtensor_rv_replace(fgraph, node):
 
 logprob_rewrites_db.register("measurable_ir_rewrites", measurable_ir_rewrites_db, "basic")
 
+# Split max_and_argmax
+# We only register this in the measurable IR db because max does not have a grad implemented
```
jessegrabowski, line 378: What do you mean by "max doesn't have a grad implemented"? `pt.grad(pt.max(x).sum(), x)` works fine. Is it because internally that graph is using `MaxAndArgmax`, not `Max` itself?

tests/logprob/test_order.py
```diff
@@ -293,3 +294,18 @@ def test_min_max_bernoulli():
     min_logp_fn = pytensor.function([value], pm.logp(pt.min(x), value))
     np.testing.assert_allclose(min_logp_fn(1), np.log(p**n))
     np.testing.assert_allclose(min_logp_fn(0), np.log(1 - p**n))
+
+
+def test_non_measurable_max_grad():
+    # Regression test for https://github.com/pymc-devs/pytensor/issues/711
+    x = pt.random.normal(0, 1, size=(3,))
+    max_x = x.max()
```
jessegrabowski, line 302: Is there any value-add to including a test that uses the `axis` kwarg?

jessegrabowski (approved):

pymc/logprob/rewriting.py
ricardoV94: Yes

tests/logprob/test_order.py
ricardoV94: That looks irrelevant for what caused the original issue

PR #7265, opened 2024-04-17
Fix need for dummy data in `sample_posterior_predictive`

ricardoV94:
## Description

Fix arviz backend find_constants

This also fixes a bug when doing sample_posterior_predictive with resized X data, where the old unused Y data would be retrieved and a dimension mismatch would crop up, unless a dummy Y was set prior to sampling.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7265.org.readthedocs.build/en/7265/

### DIFF
--- a/tests/backends/test_arviz.py
+++ b/tests/backends/test_arviz.py
@@ -454,13 +454,37 @@ class TestDataPyMC:
         test_dict = {
             "posterior": ["beta"],
             "observed_data": ["obs"],
-            "constant_data": ["x", "y", "beta_sigma"],
+            "constant_data": ["x", "beta_sigma"],
         }
         fails = check_multiple_attrs(test_dict, inference_data)
         assert not fails
         assert inference_data.log_likelihood["obs"].shape == (2, 100, 3)
         # test that scalars are dimensionless in constant_data (issue #6755)
         assert inference_data.constant_data["beta_sigma"].ndim == 0
+
+    @pytest.mark.parametrize("constant_in_generative_graph", [True, False])
+    def test_observed_data_also_constant(self, constant_in_generative_graph):
+        """Test that wen the same variable is used as constant data and observed data, it shows up in both groups."""
+        with pm.Model(coords={"trial": [0, 1, 2]}) as model:
+            x = pm.Data("x", [1.0, 2.0, 3.0], dims=["trial"])
+            sigma = pm.HalfNormal("sigma", 1)
+            mu = x - 1 if constant_in_generative_graph else 0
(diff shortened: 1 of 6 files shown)


tests/backends/test_arviz.py
```diff
@@ -454,7 +454,7 @@ def test_constant_data(self, use_context):
         test_dict = {
             "posterior": ["beta"],
             "observed_data": ["obs"],
-            "constant_data": ["x", "y", "beta_sigma"],
+            "constant_data": ["x", "beta_sigma"],
```
ricardoV94, line 457: This `y` was in fact observed data... Now nothing stops the user from using `y` both as observed data and constant data, so I am not sure we want to truly remove it.

We can check if `y` shows up in the generative graph.

ricardoV94: doing it now

zaxtax (approved):
LGTM

tests/sampling/test_forward.py
```diff
@@ -1017,6 +1017,55 @@ def test_logging_sampled_basic_rvs_posterior_mutable(self, mock_sample_results,
...
+    def test_observed_data_needed_in_pp(self):
+        # Model where y_data is not part of the generative graph.
+        # It shouldn't be needed to set a dummy value for posterior predictive sampling
+
+        with pm.Model(coords={"trial": range(5), "feature": range(3)}) as m:
+            x_data = pm.Data("x_data", np.random.normal(size=(5, 3)), dims=("trial", "feat"))
+            y_data = pm.Data("y_data", np.random.normal(size=(5,)), dims=("trial",))
+            sigma = pm.HalfNormal("sigma")
+            mu = x_data.sum(-1)
+            pm.Normal("y", mu=mu, sigma=sigma, observed=y_data, shape=mu.shape, dims=("trial",))
```
jessegrabowski, line 1029: I thought you wanted to do away with `shape=mu.shape` in this PR?

ricardoV94: No, that's needed so the shape doesn't depend on the dim. What I'm doing away with is having to set the dim if it's not part of the generative graph

jessegrabowski: Oh right right, I understand what you're fixing now. I got confused because of the discourse thread that triggered this PR :)

jessegrabowski:
Let's get this merged? The CI failure looks unrelated.

PR #7270, opened 2024-04-21
Reenable some compiler errors

maresb:
Ideally we fix issues instead of suppressing them

## Description

I'm hoping that https://github.com/pymc-devs/pytensor/pull/725 will make this flag unnecessary. Looking forward to seeing what happens.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7270.org.readthedocs.build/en/7270/

### DIFF
--- a/pymc/__init__.py
+++ b/pymc/__init__.py
@@ -29,14 +29,13 @@ def __set_compiler_flags():
     import pytensor
 
     current = pytensor.config.gcc__cxxflags
-    augmented = f"{current} -Wno-c++11-narrowing"
 
     # Work around compiler bug in GCC < 8.4 related to structured exception
     # handling registers on Windows.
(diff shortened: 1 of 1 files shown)


maresb:
@ricardoV94, it seems to me like we're missing these upstream macos failures since the PyTensor test suite doesn't run on macos. Any particular reason it's absent? Seems like it would be prudent to add.

ricardoV94:
> @ricardoV94, it seems to me like we're missing these upstream macos failures since the PyTensor test suite doesn't run on macos. Any particular reason it's absent? Seems like it would be prudent to add.

We have refrained from duplicating the whole test suite across different OSes because it's pretty wasteful. But we can add other OS jobs with targeted tests. The hard part is just picking what to select

PR #7288, opened 2024-04-28
Fix draws of Weibull when alpha and beta implicitly define size

### LINKED ISSUE #7220: BUG: Generation of random draws from Weibull returns constant values
### Describe the issue:

The generation of random draws from a Weibull distribution is returning the same values when the input is constant along some axis. This does not happen with other distributions such as the Gamma. See the example below.

```python
import numpy as np
import scipy.special as sp
import pymc as pm

# Simulate mean from an only-intercept model. 2 chains, 100 draws, 5 observations.
# So 'mu' is the same for all the observations (because it's intercept-only)
mu_draws = np.abs(150 + np.dstack([rng.normal(size=(2, 100, 1))] * 5))

# Simulate some alpha values
alpha_draws = np.abs(rng.normal(size=(2, 100, 1)))

# With 'mu' and 'alpha' get 'beta', which is what pm.Weibull needs
beta_draws = mu_draws / sp.gamma(1 + 1 / alpha_draws)

# See the draws, for a given chain and draw, they look all the same!
weibull_draws = pm.draw(pm.Weibull.dist(alpha=alpha_draws, beta=beta_draws))
weibull_draws

print((weibull_draws == weibull_draws[:, :, 0][..., None]).all())
# True --> they're in fact all the same
```

You can see this is not the case for the gamma distribution

```python
gamma_draws = pm.draw(pm.Gamma.dist(alpha=alpha_draws, beta=beta_draws))
(gamma_draws == gamma_draws[:, :, 0][..., None]).all() # False
```

### Error message:

_No response_

### PyMC version information:

5.11.0

### Context for the issue:

Someone reported this in Bambi after they saw a warning with `az.plot_ppc()`
https://github.com/bambinos/bambi/discussions/788

tomicapretto:
## Description

This PR incorporates the broadcasting of the shapes of `alpha` and `beta` in the random value generation method from the Weibull distribution to provide a fix to the problem shown in #7220 

## Related Issue

Closes #7220

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7288.org.readthedocs.build/en/7288/

### DIFF
--- a/tests/distributions/test_continuous.py
+++ b/tests/distributions/test_continuous.py
@@ -2369,6 +2369,25 @@ class TestWeibull(BaseTestDistributionRandom):
         "check_rv_size",
     ]
 
+    # See issue #7220
+    def test_rng_different_shapes(self):
+        rng = np.random.default_rng(123)
+
+        # Simulate mean from an only-intercept model. 2 chains, 100 draws, 5 observations.
+        # So 'mu' is the same for all the observations (because it's intercept-only)
+        mu_draws = np.abs(150 + np.dstack([rng.normal(size=(2, 100, 1))] * 5))
+
+        # Simulate some alpha values
+        alpha_draws = np.abs(rng.normal(size=(2, 100, 1)))
+
+        # With 'mu' and 'alpha' get 'beta', which is what pm.Weibull needs
+        beta_draws = mu_draws / sp.gamma(1 + 1 / alpha_draws)
+
+        # See the draws, for a given chain and draw, they look all the same!
+        weibull_draws = pm.draw(pm.Weibull.dist(alpha=alpha_draws, beta=beta_draws))
+
+        assert not (weibull_draws == weibull_draws[:, :, 0][..., None]).all()
+
 
 @pytest.mark.skipif(
     condition=_polyagamma_not_installed,
--- a/tests/distributions/test_continuous.py
+++ b/tests/distributions/test_continuous.py
@@ -2369,24 +2369,13 @@ class TestWeibull(BaseTestDistributionRandom):
         "check_rv_size",
     ]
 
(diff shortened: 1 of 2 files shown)


ricardoV94 (approved):

ricardoV94:
Thanks @tomicapretto, great test although a bit unfair that it only exists for this RV.

That means I'm okay with no regression test here, but I am definitely not against it either

tests/distributions/test_continuous.py
```diff
@@ -2369,6 +2369,25 @@ def seeded_weibul_rng_fn(self):
         "check_rv_size",
     ]
 
+    # See issue #7220
```
ricardoV94: Nitpick, comment inside the test function is less likely to be lost/out of place in the future

tests/distributions/test_continuous.py
```diff
@@ -2369,6 +2369,25 @@ def seeded_weibul_rng_fn(self):
...
+    def test_rng_different_shapes(self):
+        rng = np.random.default_rng(123)
```
ricardoV94, line 2374: Could this test simply batch one of the arguments and check len(np.unique(draws)) == draws.size?

ricardoV94: Something like

```python
alpha = rng.normal(size=5)  # or whatever domain is valid
beta = rng.normal(size=(3, 1))
draws = pm.draw(pm.Weibull.dist(alpha, beta), random_seed=rng)
assert len(np.unique(draws)) == draws.size  # or np.nunique if that exists
```

ricardoV94:
Thanks @tomicapretto !

PR #7290, opened 2024-04-28
Implement CustomProgress that does not output empty divs when disabled

### LINKED ISSUE #7264: BUG: new progressar adds empty spaces when progressbar=False
### Describe the issue:

As I need to sample the posterior on a very large sample of 1M rows, I have batched the operation and run it in groups of 10k
To monitor the progress, I put a tqdm progress bar on the bach loop. 

This used to behave as expected, but since 5.13.1 this produces the following:
image

```python
import pymc as pm

# Define and sample any model you like into idata

from tqdm import tqdm
with model:   
    for i in tqdm(list(range(10))):
        pm.sample_posterior_predictive(
                    idata,
                    progressbar=False,
                )
```

### Error message:

_No response_

### PyMC version information:

5.13.1

### Context for the issue:

_No response_

tomicapretto:
## Description

This PR adds a new class `CustomProgress` that inherits from `rich.progress.Progress`. I overloaded the necessary methods to make it not output anything when we disable the progress bar. **I feel it's a bit hacky** so I would like to know what others think about this modification.

This problem is noticeable especially in VS Code. It's not such a big issue in Jupyter Notebook. Some screenshots:

VS Code `progressbar=True` (before and after we get the same result, I cannot get rid of those empty spaces)

capture1

VS Code `progressbar=False` before the changes

capture2

VS Code `progressbar=False` after the changes

image

Jupyter Notebook after the changes

capture3

**Update**

In the terminal, after the change

image

## Related Issue

-  Closes #7264

----
📚 Documentation preview 📚: https://pymc--7290.org.readthedocs.build/en/7290/

### DIFF
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -32,7 +32,7 @@ from arviz import InferenceData, concat, rcParams
 from arviz.data.base import CoordSpec, DimSpec, dict_to_dataset, requires
 from pytensor.graph import ancestors
 from pytensor.tensor.sharedvar import SharedVariable
-from rich.progress import Console, Progress
+from rich.progress import Console
 from rich.theme import Theme
 from xarray import Dataset
 
@@ -40,7 +40,7 @@ import pymc
 
 from pymc.model import Model, modelcontext
 from pymc.pytensorf import PointFunc, extract_obs_data
-from pymc.util import default_progress_theme, get_default_varnames
+from pymc.util import CustomProgress, default_progress_theme, get_default_varnames
 
 if TYPE_CHECKING:
     from pymc.backends.base import MultiTrace
@@ -649,8 +649,10 @@ def apply_function_over_dataset(
     out_dict = _DefaultTrace(n_pts)
     indices = range(n_pts)
 
-    with Progress(console=Console(theme=progressbar_theme), disable=not progressbar) as progress:
-        task = progress.add_task("Computing ...", total=n_pts, visible=progressbar)
+    with CustomProgress(
+        console=Console(theme=progressbar_theme), disable=not progressbar
+    ) as progress:
+        task = progress.add_task("Computing ...", total=n_pts)
         for idx in indices:
             out = fn(posterior_pts[idx])
             fn.f.trust_input = True  # If we arrive here the dtypes are valid
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -44,6 +44,7 @@ from pytensor.tensor.random.var import (
 )
 from pytensor.tensor.sharedvar import SharedVariable
 from rich.console import Console
+from rich.progress import BarColumn, TextColumn, TimeElapsedColumn, TimeRemainingColumn
 from rich.theme import Theme
 
 import pymc as pm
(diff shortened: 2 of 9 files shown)


tomicapretto:
Some tests are failing, it's because there's something I missed in the child class I created.

tomicapretto:
@ricardoV94 this is ready to be reviewed

ricardoV94:
@tomicapretto you seem to have picked commits from main, can you rebase? Hard to review otherwise

tomicapretto:
> @tomicapretto you seem to have picked commits from main, can you rebase? Hard to review otherwise

I think I did not the cleanest thing but I think it does not show anything from main now

tomicapretto:
Converted to a draft because I found other points of improvement.

* I think the `refresh` argument is not needed and it causes the sampler to be slower
* There should be some kind of final update when computing predictions, see the bar is not full:

image

fonnesbeck:
The `refresh` argument was added because (1) progress bars did not always end at 100% after sampling was complete and (2) the divergences text did not update without it. Can you confirm that your changes don't result in a regression for these behaviours?

fonnesbeck:
I see you noticed (1) in your comment above. What about adding a single `refresh` update at the end of sampling? Perhaps that would ensure the last update is complete(?)

tomicapretto:
@fonnesbeck I'll test some ideas / combinations and update on my findings

drbenvincent:
FYI: This has also affected doctests in CausalPy, see https://github.com/pymc-labs/CausalPy/issues/323 Would be very pleased to see a fix :)

tomicapretto:
Before the last commit, it wouldn't update to 100% using one chain (because it doesn't use multiprocess sampling)

completed_not_full

Now it works well

image

I can confirm divergences are updated and it goes up to 100%
divergences

The progressbar for posterior predictive sampling was also broken:
posterior_predictive_not

Now it works fine:
posterior_predictive_ok

tomicapretto:
I'm now trying to see if I can get those two extra divs not added. It does not take any relevant space when using jupyter notebook, but it appears on VS Code and it's really bothering. 

image

tomicapretto:
Ok I have some findings. It has to do with how `rich` handles the output in jupyter notebooks. `Console` has a `force_jupyter` parameter which is `None` by default (and with notebooks it's set to `True` internally). See what happens when we change it.


In VS Code, it removes the extra space that gets added, but notice it changes the color and the description:
force_jupyter_vs

In VS Code, it doesn't do any harm, but it stillchanges the color and the description:
force_jupyter_jpt


I'm not sure if this is something we can take care of. It's really bothering for the ones who use VS Code, but I don't see any simple solution. I'll do a bit more research but I'm not going to invest a lot of time on it.

ricardoV94:
Not to be a bummer, but should we revert to the old progessbar? @fonnesbeck @zaxtax ?

fonnesbeck:
I can dig into it a bit later today. If it comes down to color differences, I'm fine with it. If its deeper than that then we probably should seek alternatives (including reverting).

aseyboldt:
As I noticed in nutpie, it is also surprisingly easy to build a custom progress bar. We could also adapt something like the nutpie one into pymc if that makes sense.
It won't work (without modifications) in the terminal however.

fonnesbeck:
@aseyboldt you are using `fastprogress` in nutpie, correct?

aseyboldt:
Not anymore. This included an update progressbar: https://github.com/pymc-devs/nutpie/pull/105
image

It's not released yet though.
The code is around here: https://github.com/pymc-devs/nutpie/blob/main/python/nutpie/sample.py#L308

fonnesbeck:
I'm inclined to blame VS Code here TBH. The Python interactive pane is very buggy, not just here but in a variety of settings. Buttons on the toolbar often do not work, the input field occasionally gets resized so you can't read the text in it, and so on.

fonnesbeck:
In the meantime, should we at least go ahead and merge some version of this PR? It clearly fixes things for a lot of people.

tomicapretto:
> In the meantime, should we at least go ahead and merge some version of this PR? It clearly fixes things for a lot of people.

I agree. I've seen the speed being a lot worse just because of the progressbar. So this is already something.

fonnesbeck:
The outstanding failure appears to be related to Jax versioning(?) Its complaining about using `max` as an argument to `clip` instead of `a_max`.

ricardoV94:
> The outstanding failure appears to be related to Jax versioning(?) Its complaining about using `max` as an argument to `clip` instead of `a_max`.

It's blackjax fault: https://github.com/pymc-devs/pymc/pull/7317

fonnesbeck:
xfail here too, then?

fonnesbeck:
@tomicapretto can you rebase this to get the fix in #7320 ?

tomicapretto:
> @tomicapretto can you rebase this to get the fix in #7320 ?

@fonnesbeck I think rebased correctly, but I'm not 100% as I don't find myself doing it often.

fonnesbeck (approved):
LGTM

PR #7293, opened 2024-04-29
Handle NaNs in the logp in SMC

### LINKED ISSUE #7292: BUG: NaNs in logp mess up resampling step in SMC
### Describe the issue:

The model I am working with sometimes returns NaN for the logp (it is an ODE model, and for certain nasty parameter combinations the solver gives up). When using SMC as a sampler, this leads to unexpected behavior in the resampling step: All the weights will be NaN, and as a result the resampling indices will all be zero.

Arguably, the logp should not be NaN, and it is unclear (at least to me) what behavior would be correct in this situation. But the current behavior will lead to a collapse of the samples without any warning or explanation of what is happening. My pragmatic solution would be to ignore any samples with a NaN logp during resampling by setting their logp to -inf.

I don't know how to create a complete example model to reproduce this. But the following code demonstrates the behavior of the current SMC implementation:

```python
from scipy.special import logsumexp
from pymc.smc.kernels import systematic_resampling
import numpy as np

# Create random logp values
probs = np.random.rand(20)
probs[3] = np.NaN
log_weights_un = np.log(probs)
# Set NaNs to -inf to resolve the problem
# log_weights_un = np.where(np.isnan(log_weights_un), -np.inf, log_weights_un)

# Compute the weights as in the SMC kernel
log_weights = log_weights_un - logsumexp(log_weights_un)
weights = np.exp(log_weights)

# Get the resampling indices
rng = np.random.default_rng(seed=5)
systematic_resampling(weights, rng=rng)
```

### Error message:

_No response_

### PyMC version information:

PyMC version: 5.13.1 (local installation of main branch)

### Context for the issue:

_No response_

astoeriko:
## Description
The changes deal with the case that the logp is NaN in the SMC sampler.
With the changes, samples with a NaN-logp will be simply discarded during the resampling step by assigning them a logp of -inf.

I know that this is not an ideal solution to the problem, but rather a pragmatic workaround. Maybe it would be wise to add a warning that there were NaN values?

## Related Issue

- [x] Closes #7292
- [ ] Related to #

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7293.org.readthedocs.build/en/7293/

### DIFF
--- a/pymc/smc/kernels.py
+++ b/pymc/smc/kernels.py
@@ -269,6 +269,7 @@ class SMC_KERNEL(ABC):
         while up_beta - low_beta > 1e-6:
             new_beta = (low_beta + up_beta) / 2.0
             log_weights_un = (new_beta - old_beta) * self.likelihood_logp
+            log_weights_un = np.where(np.isnan(log_weights_un), -np.inf, log_weights_un)
             log_weights = log_weights_un - logsumexp(log_weights_un)
             ESS = int(np.exp(-logsumexp(log_weights * 2)))
             if ESS == rN:
@@ -280,6 +281,7 @@ class SMC_KERNEL(ABC):
         if new_beta >= 1:
             new_beta = 1
             log_weights_un = (new_beta - old_beta) * self.likelihood_logp
+            log_weights_un = np.where(np.isnan(log_weights_un), -np.inf, log_weights_un)
             log_weights = log_weights_un - logsumexp(log_weights_un)
 
         self.beta = new_beta


aseyboldt:
Thank you for the PR!

It would be nice if we can come up with a test that this is in fact enough.
Maybe we can arbitrarily introduce some nan values into a model, and sample it? So for instance

```
with pm.Model():
    x = pm.Normal("x")
    pm.Normal("y", mu=x, sigma=0.1, observed=1)
    # Return nan in 50% of the prior draws
    pm.Potential("make_nan", pt.where(pt.geq(x, 0), 0, np.nan))
```

astoeriko:
Testing this with a simple example would be great! I was wondering if there is a way to artificially introduce NaN values in a model. So thanks for providing an example, this will help me verify with a simpler model if the problems I am seeing when sampling with SMC are indeed related to not handling NaN values.
I am still not very clear about what the test should test in the end. That we do not end up with a single sample per chain?

aseyboldt:
I think checking that all samples are positive and the posterior variance is reasonable should be enough.
The true posterior standard deviation should be $\sqrt{(1 + 1/100)^{-1}} \approx 0.1$, so maybe we just check that it's between 0.05 and 0.2 or so? We could also be more thorough and do a ks test against the true posterior, but I think for our purposes here that shouldn't be necessary.

PR #7296, opened 2024-05-01
Fix bug when `freezing_rv_and_dims` after a model transformation

ricardoV94:
## Description

Also adds optimization to avoid cloning the fgraph when it will be discarded anyway. 
Benchmarked with long models and can cut `clone_model` (the simplest workflow) by 2x

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7296.org.readthedocs.build/en/7296/

### DIFF
--- a/pymc/model/fgraph.py
+++ b/pymc/model/fgraph.py
@@ -296,11 +296,20 @@ def model_from_fgraph(fgraph: FunctionGraph) -> Model:
     model = Model()
     if model.parent is not None:
         raise RuntimeError("model_to_fgraph cannot be called inside a PyMC model context")
-    model._coords = getattr(fgraph, "_coords", {})
-    model._dim_lengths = getattr(fgraph, "_dim_lengths", {})
+
+    _coords = getattr(fgraph, "_coords", {})
+    _dim_lengths = getattr(fgraph, "_dim_lengths", {})
+
+    fgraph, memo = fgraph.clone_get_equiv(check_integrity=False, attach_feature=False)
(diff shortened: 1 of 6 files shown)


ricardoV94:
I think I broke something in the fgraph stuff, need to investigate

pymc/model/fgraph.py
```diff
@@ -278,12 +278,18 @@ def fgraph_from_model(
     return fgraph, memo
 
 
-def model_from_fgraph(fgraph: FunctionGraph) -> Model:
+def model_from_fgraph(fgraph: FunctionGraph, mutate_fgraph: bool = False) -> Model:
```
jessegrabowski, line 281: Every time you call the function you use `mutate_fgraph = True`, should that be the default instead? Or are the instances in this PR special cases (and there are many others)

jessegrabowski (approved):
Looks good. The bug was just related to the shapes of transformed models not being correctly frozen?

pymc/model/fgraph.py
ricardoV94: I prefer to ask for consent

ricardoV94:
> Looks good. The bug was just related to the shapes of transformed models not being correctly frozen?

Yes, because the shared dim length was not the exact same variable used in the size of the RVs due to an extra cloning of the later. They would still be resized correctly if you updated the dim length but not frozen when you called the associated utility

PR #7298, opened 2024-05-03
Fixes & improvements on HSGP

### LINKED ISSUE #7240: BUG: HSGP making wrong out-of-sample predictions when X domain changes
### Describe the issue:

While working on the HSGP tutorial with @bwengals, we discovered out-of-sample predictions are wrong when the range of the X domain changes, because `L`, the parameter on the boundary condition is computed again the new values of X... but shouldn't.
I'll fix that in a PR coming very soon. I'll also make sure the mean-centering of data, necessary to compute `L`, is done under the hood, instead of requiring the user to do it -- which is the current API but is a sneaky source of error

```python
x = np.linspace(-5, 5, 100)

with pm.Model() as model:
    # Set mutable data
    X = pm.MutableData("X", x[:, None])

    m = [5]
    c = 5.0

    X_mu = np.mean(x, axis=0)
    Xs = X - X_mu
    
    ### WITH EVAL, NO BUG.  WITHOUT EVAL, BUG
    L = pm.gp.hsgp_approx.set_boundary(Xs, c).eval()
    
    eigvals = pm.gp.hsgp_approx.calc_eigenvalues(L, m, tl=pt)
    phi = pm.gp.hsgp_approx.calc_eigenvectors(Xs, L, eigvals, m, tl=pt)

phi1 = phi.eval()

x_new = np.linspace(-5, 10, 200)

with model:
    pm.set_data({"X": x_new[:, None]})
    
plt.plot(x, phi1, 'k');
plt.plot(x_new, phi.eval(), 'b', alpha=0.5);
```

### Error message:

_No response_

### PyMC version information:

5.12.0

### Context for the issue:

_No response_

AlexAndorra:
Closes #7240 

Working on this in concert with @bwengals for [this example](https://github.com/pymc-devs/pymc-examples/pull/

- [x] Fix the [bug in the computation](https://github.com/pymc-devs/pymc/issues/7240) of `L`
- [x] Do mean-centering of `X` values under the hood, instead of requiring it from users
- [x] Enable setting `prior_linearized` with `m` and `c` instead of only `m` and `L`
- [x] `parametrization` is not documented
- [x] Add option to pass in nicely named `coords` to HSGP for the `beta` coefficients
- [x] Make `._m_star` more accessible / user friendly (make a property with no leading underscore so existing code doesn't break), document it, and fix example in `prior_linearized` docstring.
- [x] Get rid of numpy option, `tl=np`.  It's subtly wrong when calculating the eigen stuff and its not used anywhere.
- [x] Add the `approx_params` function to `hsgp_approx` file

## Type of change
- New feature / enhancement
- Bug fix

----
📚 Documentation preview 📚: https://pymc--7298.org.readthedocs.build/en/7298/

### DIFF
--- a/pymc/gp/hsgp_approx.py
+++ b/pymc/gp/hsgp_approx.py
@@ -32,10 +32,10 @@ TensorLike = np.ndarray | pt.TensorVariable
 
 def set_boundary(Xs: TensorLike, c: numbers.Real | TensorLike) -> TensorLike:
     """Set the boundary using the mean-subtracted `Xs` and `c`.  `c` is usually a scalar
-    multiplyer greater than 1.0, but it may be one value per dimension or column of `Xs`.
+    multiplier greater than 1.0, but it may be one value per dimension or column of `Xs`.
     """
     S = pt.max(Xs, axis=0)
-    L = (c * S).eval()  # to make sure L is not updated with out-of-sample preds
+    L = (c * S).eval()  # eval() makes sure L is not changed with out-of-sample preds
     return L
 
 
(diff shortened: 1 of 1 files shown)


AlexAndorra:
Closing in favor #7335 , for clean git history

Issue #7277, opened 2024-04-22
BUG: pm.Data does not accept dtype argument


ricardoV94:
@michaelosthege and I found out that the latest changes create issues when using `pm.ConstantData` (or `pm.Data`) and setting a dtype explicitly. We don't understand why because `pytensor.shared` has no problem with the dtype argument.

Here is an example:

```python
with pm.Model():
    pm.Data("b", [True, False], dtype=bool)
```
```
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "C:\Users\osthege\AppData\Local\mambaforge\envs\dibecs_6.13.0\lib\site-packages\pymc\data.py", line 420, in Data
    x = pytensor.shared(arr, name, **kwargs)
  File "C:\Users\osthege\AppData\Local\mambaforge\envs\dibecs_6.13.0\lib\site-packages\pytensor\compile\sharedvalue.py", line 202, in shared
    var = shared_constructor(
  File "C:\Users\osthege\AppData\Local\mambaforge\envs\dibecs_6.13.0\lib\functools.py", line 889, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
TypeError: tensor_constructor() got an unexpected keyword argument 'dtype'
```

Do you have any idea why this is happening? Thanks in advance for your help!

_Originally posted by @lhelleckes in https://github.com/pymc-devs/pymc/issues/7047#issuecomment-2070305967_

michaelosthege:
[This conversion](https://github.com/pymc-devs/pymc/blob/34c2d31484143b7b5c6130927ed4efde93a27a0f/pymc/model/core.py#L1312) given in #7257 is a good example for the solution that @lhelleckes and I just discussed:

Adding a `convert_observed_data(value, dtype: Optional=None)` option which gets priority would allow the line linked above to enforce the type and thereby avoid the casting.

PR #7299, opened 2024-05-06
Refactor convert_observed_data

lhelleckes:
## Description
In order to improve type hints in the `convert_observed_data` function and to ultimately resolve issue #7277, the generator part of the code was separated in a statement with return. This will make it easier to apply dtypes to the other data structures in the next step.

## Related Issue
- [ ] Closes #
- [x] Related to #7277

## Checklist

- [x] Included tests that prove the fix is effective or that the new feature works (not applicable since refactoring)

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7299.org.readthedocs.build/en/7299/

### DIFF
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -78,8 +78,11 @@ __all__ = [
 ]
 
 
-def convert_observed_data(data):
+def convert_observed_data(data) -> np.ndarray | Variable:
     """Convert user provided dataset to accepted formats."""
+
+    if isgenerator(data):
+        return floatX(generator(data))
 
     if hasattr(data, "to_numpy") and hasattr(data, "isnull"):
         # typically, but not limited to pandas objects
@@ -116,8 +119,6 @@ def convert_observed_data(data):
         ret = data
     elif sps.issparse(data):
         ret = data
-    elif isgenerator(data):
-        ret = generator(data)
     else:
         ret = np.asarray(data)
 


michaelosthege:
Good news is that all tests except `test_skewstudentt_logp` worked.

Flaky tests are annoying because we can't XFAIL them (I forgot).

michaelosthege:
@lhelleckes you can rebase now :)

pymc/pytensorf.py
```diff
@@ -78,9 +78,12 @@
 ]
 
 
-def convert_observed_data(data):
+def convert_observed_data(data) -> np.ndarray | Variable:
     """Convert user provided dataset to accepted formats."""
 
+    if isgenerator(data):
+        return floatX(generator(data))
```
ricardoV94, line 85: I think this was on purpose, for something fancy in VI. @ferrine can you confirm whether it's true and whether we still need it?

michaelosthege: Yes, the refactor only moves the code, but if we can delete it that'd be even better!

ricardoV94: Doesn't wrapping the generator in floatX consume it immediately?

ricardoV94: nvm it was also done before

michaelosthege: User-provided `observed` data comes through this, therefore I'm pretty certain that this generator branch is still relevant for VI.

One next step after this PR could be to split the function into two overloads - one of which applies to `generator` type `data`, and the other to all the rest.

The more important next step should be the introduction of a `dtype` kwarg so we can get #7277 fixed.

michaelosthege (approved):

pymc/pytensorf.py
aseyboldt: Just a guess, but that might be for stochastic optimization? If so, that generator could even be infinite.
Why not `(floatX(value) for value in generator)`?

aseyboldt:
Couldn't the observed value be supposed to be an integer?

pymc/pytensorf.py
ricardoV94: If the output can be a generator, the type signature is wrong

michaelosthege: No, even with a generator, the returned value is a `TensorVariable`:

grafik

michaelosthege:
> Couldn't the observed value be supposed to be an integer?

Yes, in that case it will result in a NumPy array too:

```
>>> type(pm.floatX(5))
<class 'numpy.ndarray'>
```

aseyboldt:
I meant that it might be an array (pytensor or numpy) with integer dtype. Unless I'm missing some context we can't just convert that to a float type.

pymc/pytensorf.py
ricardoV94: I see some black magic.

michaelosthege:
> I meant that it might be an array (pytensor or numpy) with integer dtype. Unless I'm missing some context we can't just convert that to a float type.

If you follow the branching, you'll find that we made that conversion all the time already.

In my opinion we should merge this and continue adding a `dtype` kwarg to fix #7277.

-------------

The whole "preparing generator data for VI" should be refactored. I would probably even give it it's own `pm.GeneratorData` container.
This is not the scope of this PR though.

If y'all agree I can take the first step towards putting `generator` data into a `pm.GeneratorData` container. (`to_graphviz` style, deprecation warning, ...)

aseyboldt:
> If you follow the branching, you'll find that we made that conversion all the time already.

Sorry, I don't know what you mean. Can you point me to an example? I don't think we are converting data that a users specified as an int type to a float type automatically, do we?

michaelosthege:
> > If you follow the branching, you'll find that we made that conversion all the time already.
> 
> Sorry, I don't know what you mean. Can you point me to an example? I don't think we are converting data that a users specified as an int type to a float type automatically, do we?

main branch: https://github.com/pymc-devs/pymc/blob/fd11cf012895a9981351097df420b7fbbfb693a4/pymc/pytensorf.py#L119-L133

When `data` is a generator, the `if isgenerator(data)` is the first case that evaluates `True`.
Then `if hasattr(data, "dtype") → False` and `return floatX(ret)`.

If we _should_ do that is a separate, VI-specific question which is IMO best dealt with by separating the generator case away.

Issue #6716, opened 2023-05-12
Allow more minimal graphviz representation


drbenvincent:
In Bayesian Network and Causal Graph land, you will often see quite minimal networks which focus on the core variables of interest and their statistical and/or causal relations. Here is an example.
![Screenshot 2023-05-12 at 15 45 10](https://github.com/pymc-devs/pymc/assets/6765047/4e81a2b7-2e9f-416c-9e07-765bed775d9d)

If you code up such a model in PyMC (e.g. with linear relationships between the nodes) you get a 'full' representation of the graph that PyMC builds, like this...

![Screenshot 2023-05-12 at 15 49 06](https://github.com/pymc-devs/pymc/assets/6765047/ef3bc582-486e-4c0c-b30b-7bbbd46d101b)

But it would be extremely useful to be able to produce more minimal representations, like the first figure.

So this is a feature request for that. 

In terms of implementation, it could be done using a `minimal=True` kwarg to `pm.model_to_graphviz`. Alternatively it could be a new function, such as `pm.model_to_simple_graphviz`. My initial preference would be for the latter because you could imaging wanting to have a number of other kwargs to control the visualisation which you might not want in the existing function.

This issue is not entirely unrelated to #5794, but it is not the same

ricardoV94:
I suspect that the `minimal=True` is a bit ambiguous and different users may want different "minimal" versions of the graph. I think supporting `var_names` would be the most flexible way and hopefully not terrible to work with.

drbenvincent:
Though having to manually specify could be far from the quick and easy process of visualising that we have at the moment.

drbenvincent:
Perhaps `style = "causal_dag"`. That would be future proof and allow for multiple different graphviz DAG styles.

fonnesbeck:
Perhaps something like `include_dist=False`, which would exclude the tilde and distribution type from the graph?

drbenvincent:
> Perhaps something like `include_dist=False`, which would exclude the tilde and distribution type from the graph?

That would be pretty good. Presumably that alone would still leave the RV's and deterministics in the graph? But I like the idea in general of allowing users to configure the graphviz output. I think this is a really strong aspect of PyMC and that there are things we can do to make the outputs more suited to different contexts.

larryshamalama:
[...]

PR #7302, opened 2024-05-07
Allow customizing style of model_graph nodes

williambdean:
## Description

Allow users to override the default graphviz by passing mapping from node type to function that creates the node kwargs. 

The default behavior is the same but now the user can override based on the node type defined: 

- Data
- Free Random Variable
- Observed Random Variable
- Deterministic
- Potential

User needs to define function from node variable to kwargs passed to graphviz / networkx

```python
def node_formatter(var: TensorVariable) -> dict[str, Any]:
    return {"label": var.name}
```

then pass a mapping from node type to node_formatter.

```python
node_mapping = {"Data": node_formatter}
pm.model_to_graphviz(model, node_formatters=node_mapping)
```

## Example

```python
import pymc as pm 

with pm.Model() as model: 
    a = pm.Normal("a")
    b = pm.Normal("b", mu=a)
    c = pm.Normal("c", mu=a)
    d = pm.Normal("d", mu=b + c, observed=0)

default = pm.model_to_graphviz(model)

simple_random_variable_formatter = {
    "Free Random Variable": lambda var: {"shape": "circle", "label": var.name}, 
}
pm.model_to_graphviz(model, node_formatters=simple_random_variable_formatter)

fancy_formatter = {
    "Free Random Variable": lambda var: {"shape": "polygon", "sides": "7", "label": var.name, "style": "dashed"}, 
    "Observed Random Variable": lambda var: {"shape": "circle", "label": var.name, "style": "solid"},
}
pm.model_to_graphviz(model, node_formatters=fancy_formatter)
```
Default: 
default

Simple: 
simple

Fancy: 

polygon

## Related Issue

- [ ] Closes #
- [ ] Related to #6716 

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7302.org.readthedocs.build/en/7302/

### DIFF
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -47,7 +47,7 @@ class NodeType(str, Enum):
     """Enum for the types of nodes in the graph."""
 
     POTENTIAL = "Potential"
-    BASIC_RV = "Basic Random Variable"
+    FREE_RV = "Basic Random Variable"
     OBSERVED_RV = "Observed Random Variable"
     DETERMINISTIC = "Deterministic"
     DATA = "Data"
@@ -121,10 +121,10 @@ def get_node_type(var_name: VarName, model) -> NodeType:
 
     if v in model.potentials:
         return NodeType.POTENTIAL
-    elif v in model.basic_RVs and v in model.observed_RVs:
+    elif v in model.observed_RVs:
         return NodeType.OBSERVED_RV
-    elif v in model.basic_RVs:
-        return NodeType.BASIC_RV
+    elif v in model.Free_RVs:
+        return NodeType.Free_RV
     elif v in model.deterministics:
         return NodeType.DETERMINISTIC
     else:
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -142,7 +142,7 @@ DEFAULT_NODE_FORMATTERS: NodeTypeFormatterMapping = {
 }
 
 
-def update_node_formatters(node_formatters: NodeTypeFormatterMapping):
+def update_node_formatters(node_formatters: NodeTypeFormatterMapping) -> NodeTypeFormatterMapping:
     node_formatters = {**DEFAULT_NODE_FORMATTERS, **node_formatters}
 
     unknown_keys = set(node_formatters.keys()) - set(NodeType)
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -47,7 +47,7 @@ class NodeType(str, Enum):
     """Enum for the types of nodes in the graph."""
 
     POTENTIAL = "Potential"
-    FREE_RV = "Basic Random Variable"
+    FREE_RV = "Free Random Variable"
     OBSERVED_RV = "Observed Random Variable"
     DETERMINISTIC = "Deterministic"
     DATA = "Data"
@@ -58,7 +58,7 @@ NodeFormatter = Callable[[TensorVariable], GraphvizNodeKwargs]
 
 
 def default_potential(var: TensorVariable) -> GraphvizNodeKwargs:
-    """Default data for the node in the graph."""
+    """Default data for potential in the graph."""
     return {
         "shape": "octagon",
         "style": "filled",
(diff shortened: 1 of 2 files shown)


wd60622:
@drbenvincent might be interested in this example

```python
with pm.Model() as model: 
    c = pm.Normal("c")
    z = pm.Normal("z", mu=c)
    y = pm.Normal("y", mu=c + z)

node_formatters = {
    "Basic Random Variable": lambda var: {"shape": "circle", "label": var.name},
}
model_graph.model_to_graphviz(model, node_formatters=node_formatters)
```
simple

pymc/model_graph.py
```diff
@@ -41,6 +43,118 @@ def fast_eval(var):
     return function([], var, mode="FAST_COMPILE")()
 
 
+class NodeType(str, Enum):
+    """Enum for the types of nodes in the graph."""
+
+    POTENTIAL = "Potential"
+    BASIC_RV = "Basic Random Variable"
```
ricardoV94: model.basic_RVs contains both observed and free RVs

```suggestion
    FREE_RV = "Basic Random Variable"
```

pymc/model_graph.py
```diff
@@ -41,6 +43,118 @@ def fast_eval(var):
...
+
+def get_node_type(var_name: VarName, model) -> NodeType:
+    v = model[var_name]
+
+    if v in model.potentials:
+        return NodeType.POTENTIAL
+    elif v in model.basic_RVs and v in model.observed_RVs:
+        return NodeType.OBSERVED_RV
+    elif v in model.basic_RVs:
+        return NodeType.BASIC_RV
```
ricardoV94:
```suggestion
    elif v in model.observed_RVs:
        return NodeType.OBSERVED_RV
    elif v in model.Free_RVs:
        return NodeType.Free_RV
```

pymc/model_graph.py
```diff
@@ -41,6 +43,118 @@ def fast_eval(var):
...
+        "shape": "box",
+        "style": "rounded, filled",
+        "label": f"{var.name}\n~\nData",
+    }
+
+
+def get_node_type(var_name: VarName, model) -> NodeType:
+    v = model[var_name]
+
+    if v in model.potentials:
```
ricardoV94: Small optimization, most models don't have potentials, and most have many free_RVs, and deterministics, so I would suggest changing the order of this to be something like `if deterministics elif free rvs elif observed rvs elif data elif potentials` instead

ricardoV94 (commented):
Looks sweet. Left some minor suggestions.

pymc/model_graph.py
williambdean: Should I change the name of the string as well?

williambdean: Slash do all of the string names make sense?

ricardoV94: > Should I change the name of the string as well?

Yeah

And yes the others made sense

wd60622:
Thank you for the review @ricardoV94 
Will add these changes and add some tests

tests/test_model_graph.py
```diff
@@ -421,3 +421,47 @@ def test_model_graph_with_intermediate_named_variables():
...
+def test_custom_node_formatting_graphviz(simple_model):
+    node_formatters = {
+        "Free Random Variable": lambda var: {
+            "label": var.name,
+        },
+    }
+
+    G = model_to_graphviz(simple_model, node_formatters=node_formatters)
+    assert G.source == (
+        "digraph {\n\ta [label=a]\n\tb [label=b]" "\n\tc [label=c]\n\ta -> b\n\tb -> c\n}\n"
```
williambdean: Any thoughts on to tests while accounting for graphviz putting the nodes in random order in the string? @ricardoV94

williambdean: Think it might have something to do with our side actual and the set ordering. [graphviz python source code for reference](https://github.com/xflr6/graphviz/blob/b820426c566f6ed6d874cbcb4118bb0492062dd0/graphviz/dot.py#L254)

williambdean: but not 100% sure about that

wd60622:
Some random / unrelated tests seem to be failing

ricardoV94 (approved):

PR #7305, opened 2024-05-09
Use more numerically stable function in TruncatedNormal area computation

velochy:
## Description
Replace the complex logic for area computation in TruncateNormal with a single, more numerically stable function meant for exactly that computation. 

## Related Issue
This is mainly code cleanup, although I did run into a few cases where it actually improves convergence.

## Checklist

## Type of change

- [X] Maintenance

----
📚 Documentation preview 📚: https://pymc--7305.org.readthedocs.build/en/7305/


ricardoV94:
can you add a test that checks the improved stability (and would fail before the changes)?

velochy:
I tried. Turns out to not be easy to condense down to a simple example. 
Isn't the code cleanup aspect here enough?

ricardoV94 (approved):

ricardoV94:
Thanks @velochy

PR #7311, opened 2024-05-13
Reduce JAX post-processing memory usage

andrewdipper:
-Change blackjax sampling to only retain the relevant sampling info - reduces sampling memory requirements
-Change `_postprocess_samples` to reuse the input arrays - reduces postprocessing memory requirements

## Description
By default the current pymc blackjax sampler accumulates all the info provided by blackjax only to subsequently delete it.  This results in memory usage several times what is expected (some info is `num_samples * num_vars` in size).  The change only stores what is used resulting in memory scaling similar to that of the numpyro jax sampler.  It's worth noting that `blackjax.window_adaptation` also has excessive memory usage but that needs to be fixed from blackjax https://github.com/blackjax-devs/blackjax/issues/667.  As such if `tune` is not set sufficiently small memory usage will still be excessive.

Changes the `"vmap"` mode of `_postprocess_samples` to donate the input device arrays resulting in (for my rough tests / models) constant additional memory usage.  This should make the `"scan"` mode unnecessary.  However, I left it in to not break anything.

## Related Issue

This should resolve by using `"vmap"` mode: https://github.com/pymc-devs/pymc/issues/6744
This should be unnecessary given the reduction in `"vmap"` memory usage: https://github.com/pymc-devs/pymc/pull/7116

## Checklist

- [ ] Included tests that prove the fix is effective or that the new feature works - prior coverage should be sufficient

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7311.org.readthedocs.build/en/7311/


andrewdipper:
Looks like test failures are due to an older version of jax and a recent blackJAX PR to fix an argument deprecation to jnp.clip: https://github.com/blackjax-devs/blackjax/pull/664
https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.clip.html

andrewdipper:
Also added in a bugfix for pm.sample not respecting compute_convergence_checks with numpyro/blackjax sampler

twiecki (approved):

twiecki:
Thanks @andrewdipper!

twiecki:
Probably need to wait for https://github.com/pymc-devs/pymc/pull/7317 to merge.

ricardoV94:
>  This should make the "scan" mode unnecessary. However, I left it in to not break anything.

Let's start deprecating it! Can you add a FutureWarning about this argument/ value being removed in the future when the user manually specifies it?

ricardoV94:
Rebasing from main should fix the test failures and allow us to confirm nothing got broken

pymc/sampling/jax.py
```diff
@@ -515,7 +496,7 @@ def sample_jax_nuts(
     keep_untransformed: bool = False,
     chain_method: str = "parallel",
     postprocessing_backend: Literal["cpu", "gpu"] | None = None,
-    postprocessing_vectorize: Literal["vmap", "scan"] = "scan",
+    postprocessing_vectorize: Literal["vmap", "scan"] = "vmap",
```
ricardoV94: We should add None as an option and make that the default, then we can raise the warning if the user specifies it manually (regardless of which) so that we can safely remove the kwarg altogether in the future

pymc/sampling/jax.py
```diff
@@ -181,7 +185,8 @@ def _postprocess_samples(
     jax_fn: Callable,
     raw_mcmc_samples: list[TensorVariable],
     postprocessing_backend: Literal["cpu", "gpu"] | None = None,
-    postprocessing_vectorize: Literal["vmap", "scan"] = "scan",
+    postprocessing_vectorize: Literal["vmap", "scan"] = "vmap",
+    donate_samples: bool = True,
```
ricardoV94: Let's set it to False by default and call with True internally?

pymc/sampling/jax.py
andrewdipper: makes sense, done

pymc/sampling/jax.py
andrewdipper: done

andrewdipper:
I added the fix for reducing memory from the blackjax window_adaptation.  But that depends on https://github.com/blackjax-devs/blackjax/pull/674 which was just merged earlier today.  So blackjax would have to be up to date - not sure how that's best handled.

ricardoV94:
Unfortunately we can't test earlier versions of blackjax because of https://github.com/conda-forge/jaxlib-feedstock/issues/249

andrewdipper:
I made some more changes for my own experiments to offload on lighter hardware as follows:

- Enable sampling in chunks to make memory independent of number of samples (if chunks > 1 then save on host memory)
- Allow pmap of vmap with shard map - so sampling of 4 chains can be done on 2 devices with 2 chains per device etc.
- Postprocessing / likelihood is done on a per chunk basis instead of all at the end (should eliminate the need for choosing the device for postprocessing)

I planned on later putting it up as a proposal if I liked it but since this change is stalled I figured it might be worth doing all together if you think the modifications are worthwhile.  It's been helpful so far for me - let me know

andrewdipper:
Just removed the blackjax window adaptation memory fix that is stalled by no conda jaxlib updates.  It's now covered by https://github.com/pymc-devs/pymc/pull/7407.  The remaining changes help with the sampling memory and should avoid repeat investigations into it

ricardoV94 (approved):

ricardoV94:
Thanks @andrewdipper looks like a great improvement

PR #7318, opened 2024-05-15
Add blas_cores argument to pm.sample

aseyboldt:
## Description

We currently do not configure blas in any way. This can lead to very bad behavior if we sample in several threads:
Many blas implementations default to using one worker thread per hardware thread in the machine. But if we sample in parallel with multiprocessing, each chain will use an independent thread pool, so we end up starting `chains*hardware_chains` worker threads. Combined with some spinnlocking that some blas implementations seem to do, this can lead to terrible performance.

This PR adds a `blas_cores` argument to `pm.sample()`, and then uses `threadpoolctl` to control how many worker threads we start.

If it is set to `None`, we don't do anything, and keep the current behavior of just using whatever the blas implementation uses as default. If set to `auto` (the default) use the cores argument to guess a decent number of blas worker threads. If it is set to an integer, we use that number of total blas worker.

See for instance [here](https://discourse.pymc.io/t/slow-sampling-speed-with-newer-versions-of-pymc/14421/18?u=aseyboldt) for a model that shows bad behavior without this PR.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7318.org.readthedocs.build/en/7318/

### DIFF
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -14,13 +14,14 @@
 
 """Functions for MCMC sampling."""
 
+import contextlib
 import logging
 import pickle
 import sys
 import time
 import warnings
 
-from collections.abc import Iterator, Mapping, Sequence
+from collections.abc import Callable, Iterator, Mapping, Sequence
 from typing import (
     Any,
     Literal,
@@ -37,6 +38,7 @@ from pytensor.graph.basic import Variable
 from rich.console import Console
 from rich.progress import Progress
 from rich.theme import Theme
+from threadpoolctl import threadpool_limits
 from typing_extensions import Protocol
 
 import pymc as pm
@@ -396,6 +398,7 @@ def sample(
     nuts_sampler_kwargs: dict[str, Any] | None = None,
     callback=None,
     mp_ctx=None,
+    blas_cores: int | None | Literal["auto"] = "auto",
     **kwargs,
 ) -> InferenceData: ...
 
@@ -427,6 +430,7 @@ def sample(
     callback=None,
     mp_ctx=None,
     model: Model | None = None,
+    blas_cores: int | None | Literal["auto"] = "auto",
     **kwargs,
 ) -> MultiTrace: ...
 
@@ -456,6 +460,7 @@ def sample(
     nuts_sampler_kwargs: dict[str, Any] | None = None,
     callback=None,
     mp_ctx=None,
+    blas_cores: int | None | Literal["auto"] = "auto",
     model: Model | None = None,
     **kwargs,
 ) -> InferenceData | MultiTrace:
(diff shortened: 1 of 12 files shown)


pymc/sampling/mcmc.py
```diff
@@ -499,6 +504,13 @@ def sample(
         Which NUTS implementation to run. One of ["pymc", "nutpie", "blackjax", "numpyro"].
         This requires the chosen sampler to be installed.
         All samplers, except "pymc", require the full model to be continuous.
+    blas_cores: int or "auto" or None, default = "auto"
+        The total number of threads blas and openmp functions should use during sampling. If set to None,
```
ricardoV94: Explain the default first?

pymc/sampling/mcmc.py
```diff
@@ -644,6 +656,37 @@ def sample(
...
+    cores = min(cores, chains)
+
+    if cores < 1:
+        raise ValueError("`cores` must be larger or equal to one")
+
+    if chains < 1:
+        raise ValueError("`chains` must be larger or equal to one")
+
+    if blas_cores is not None and blas_cores < 1:
+        raise ValueError("`blas_cores` must be larger or equal to one")
```
ricardoV94: Remove it for the sake of less code? I don't believe anybody was ever hurt by this and couldn't figure out the problem?

tests/sampling/test_mcmc.py
```diff
@@ -507,6 +507,14 @@ def test_empty_model():
         error.match("any free variables")
 
 
+def test_blas_cores():
+    with pm.Model():
+        pm.Normal("a")
+        pm.sample(blas_cores="auto", tune=10, cores=2, draws=10)
+        pm.sample(blas_cores=None, tune=10, cores=2, draws=10)
+        pm.sample(blas_cores=2, tune=10, cores=2, draws=10)
```
ricardoV94, line 515: Can we actually test anything? I guess not in a reasonable amount of logic

pymc/sampling/mcmc.py
aseyboldt: +1
Do you think we should already default to "auto", or first release something where the default is None so that this can be tested a bit more?

tests/sampling/test_mcmc.py
aseyboldt: I couldn't think of any way to check if it really does what it's supposed to do.
I checked locally that it uses the correct number of threads, but otherwise I don't really know anything.

pymc/sampling/mcmc.py
ricardoV94: I think the new default makes much more sense. This often shows up in MvNormal models and it's very tricky for beginners to debug

tests/sampling/test_mcmc.py
ricardoV94: Fair enough

ricardoV94 (approved):

ricardoV94:
pre commit failed

PR #7330, opened 2024-05-22
Get rid of most intX usages

ricardoV94:
## Description

Removed all but one uses of intX, which we started to do in https://github.com/pymc-devs/pymc/pull/7114
There is a tricky use in `convert_observed_data` that I will defer until a later PR, as it has more implications than the removals we did.

Also got rid of `floatX_array` which was just used in tests.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7330.org.readthedocs.build/en/7330/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -72,7 +72,7 @@ from pymc.distributions.shape_utils import (
 from pymc.distributions.transforms import Interval, ZeroSumTransform, _default_transform
 from pymc.logprob.abstract import _logprob
 from pymc.math import kron_diag, kron_dot
-from pymc.pytensorf import intX, normalize_rng_param
+from pymc.pytensorf import normalize_rng_param
 from pymc.util import check_dist_not_registered
 
 __all__ = [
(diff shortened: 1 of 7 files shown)


twiecki (approved):

pymc/distributions/multivariate.py
```diff
@@ -929,7 +929,7 @@ class Wishart(Continuous):
 
     @classmethod
     def dist(cls, nu, V, *args, **kwargs):
-        nu = pt.as_tensor_variable(intX(nu))
+        nu = pt.as_tensor_variable(nu, dtype=int)
```
lucianopaz, line 932: What will this default to on the `pytensor` side?

lucianopaz (commented):
LGTM. My only question is what will `as_tensor_variable(..., dtype=int)` default to on the `pytensor` side. `int` is pure python - arbitrary precision integer, it's not numpy int32 or numpy int64. Once we have that out of the way, I think you're good to merge.

pymc/distributions/multivariate.py
ricardoV94: The same that numpy does, i.e., int64.

Issue #7331, opened 2024-05-22
Problematic casting of integers in `convert_observed_data`


ricardoV94:
### Description

When we use `pytensor.config.floatX=="float32"`, integer data is downcast to `"int16"` which has a pretty limited range of 32k. For count-based likelihoods this is way too narrow. I am not sure we should be doing anything with integers to begin with. Why are PyTensor casting rules (and customization flags) not sufficient?

PR #7334, opened 2024-05-23
Split convert observed data

michaelosthege:
## Description
`convert_observed_data` is a messy do-it-all conversion function that was recently refactored in #7299.

This PR takes the next step and splits it such that generator data is treated separately.

## Related Issue
- Makes it easier to work on #7277.

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Maintenance
- [x] Other (please specify):
  - Generator data is no longer supported with `pm.Data`.
  - Generator data of integer type is no longer converted to floatX, but only to intX.

----
📚 Documentation preview 📚: https://pymc--7334.org.readthedocs.build/en/7334/

### DIFF
--- a/pymc/pytensorf.py
+++ b/pymc/pytensorf.py
@@ -74,6 +74,8 @@ __all__ = [
     "join_nonshared_inputs",
     "make_shared_replacements",
     "generator",
+    "convert_data",
+    "convert_generator_data",
     "convert_observed_data",
     "compile_pymc",
 ]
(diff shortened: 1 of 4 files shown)


ricardoV94:
I'm still of the opinion we should drop generators altogether. I don't find any documentation on how to use it, and I can't really think of all the implications this Op means inside PyTensor, which is supposed to be side-effects free unless told otherwise (it's not being told so here). It's also only applicable in the default backend.

michaelosthege:
I'll add a depreaction warning to the conversion of generator data, asking people to get in touch if this is relevant.

pymc/pytensorf.py
```diff
@@ -68,22 +68,38 @@
     "floatX",
     "intX",
     "smartfloatX",
+    "smarttypeX",
```
ricardoV94: I would rather not add these new methods to all as I prefer we to get rid of them in the future, so the less discoverable they are the better.

For reference the whole intX is pretty questionable: #7331

ricardoV94 (approved):

PR #7335, opened 2024-05-26
HSGP improvements

### LINKED ISSUE #7240: BUG: HSGP making wrong out-of-sample predictions when X domain changes

AlexAndorra:
Closes #7240 

Working on this in concert with @bwengals for [this example](https://github.com/pymc-devs/pymc-examples/pull/647)

- [x] Fix the [bug in the computation](https://github.com/pymc-devs/pymc/issues/7240) of `L`
- [x] Do mean-centering of `X` values under the hood, instead of requiring it from users
- [x] Enable setting `prior_linearized` with `m` and `c` instead of only `m` and `L`
- [x] `parametrization` is not documented
- [x] Add option to pass in nicely named `coords` to HSGP for the `beta` coefficients
- [x] Make `._m_star` more accessible / user friendly (make a property with no leading underscore so existing code doesn't break), document it, and fix example in `prior_linearized` docstring.
- [x] Get rid of numpy option, `tl=np`.  It's subtly wrong when calculating the eigen stuff and its not used anywhere.
- [x] Add the `approx_params` function to `hsgp_approx` file

## Type of change
- New feature / enhancement
- Bug fix

----
📚 Documentation preview 📚: https://pymc--7335.org.readthedocs.build/en/7335/

### DIFF
--- a/pymc/gp/hsgp_approx.py
+++ b/pymc/gp/hsgp_approx.py
@@ -30,12 +30,12 @@ from pymc.gp.mean import Mean, Zero
 TensorLike = np.ndarray | pt.TensorVariable
 
 
-def set_boundary(Xs: TensorLike, c: numbers.Real | TensorLike) -> TensorLike:
+def set_boundary(Xs: TensorLike, c: numbers.Real | TensorLike) -> np.ndarray:
     """Set the boundary using the mean-subtracted `Xs` and `c`.  `c` is usually a scalar
-    multiplyer greater than 1.0, but it may be one value per dimension or column of `Xs`.
+    multiplier greater than 1.0, but it may be one value per dimension or column of `Xs`.
     """
-    S = pt.max(pt.abs(Xs), axis=0)
-    L = c * S
+    S = pt.max(Xs, axis=0)
+    L = (c * S).eval()  # eval() makes sure L is not changed with out-of-sample preds
     return L
 
 
@@ -221,7 +221,7 @@ class HSGP(Base):
         self._m_star = int(np.prod(self._m))
         self._L: pt.TensorVariable | None = None
         if L is not None:
-            self._L = pt.as_tensor(L)
+            self._L = pt.as_tensor(L).eval()  # make sure L cannot be changed
         self._c = c
         self._parameterization = parameterization
 
@@ -287,7 +287,7 @@ class HSGP(Base):
 
                 # Order is important.
                 # First calculate the mean, then make X a shared variable, then subtract the mean.
-                #  When X is mutated later, the correct mean will be subtracted.
+                # When X is mutated later, the correct mean will be subtracted.
                 X_mean = np.mean(X, axis=0)
                 X = pm.Data("X", X)
                 Xs = X - X_mean
@@ -355,8 +355,8 @@ class HSGP(Base):
             Dimension name for the GP random variable.
         """
         self._X_mean = pt.mean(X, axis=0)
-
         phi, sqrt_psd = self.prior_linearized(X - self._X_mean)
+
         if self._parameterization == "noncentered":
             self._beta = pm.Normal(
                 f"{name}_hsgp_coeffs_", size=self._m_star - int(self._drop_first)
--- a/pymc/gp/hsgp_approx.py
+++ b/pymc/gp/hsgp_approx.py
@@ -108,9 +108,9 @@ def approx_hsgp_hyperparams(
 
     Parameters
     ----------
-    x_range : List[float]
-        The range of the input variable on which the GP is going to be evaluated.
-        Should be a list with two elements [x_min, x_max].
+    x : np.ndarray
+        The input variable on which the GP is going to be evaluated.
+        Careful: should be the X values you want to predict over, not *only* the training X.
     lengthscale_range : List[float]
         The range of the lengthscales. Should be a list with two elements [lengthscale_min, lengthscale_max].
     cov_func : str
--- a/pymc/gp/hsgp_approx.py
+++ b/pymc/gp/hsgp_approx.py
@@ -39,11 +39,11 @@ def set_boundary(Xs: TensorLike, c: numbers.Real | TensorLike) -> np.ndarray:
     return L
 
 
-def calc_eigenvalues(L: TensorLike, m: Sequence[int], tl: ModuleType = np):
+def calc_eigenvalues(L: TensorLike, m: Sequence[int]):
     """Calculate eigenvalues of the Laplacian."""
     S = np.meshgrid(*[np.arange(1, 1 + m[d]) for d in range(len(m))])
     S_arr = np.vstack([s.flatten() for s in S]).T
-    return tl.square((np.pi * S_arr) / (2 * L))
+    return np.square((np.pi * S_arr) / (2 * L))
 
 
 def calc_eigenvectors(
@@ -51,18 +51,19 @@ def calc_eigenvectors(
     L: TensorLike,
     eigvals: TensorLike,
     m: Sequence[int],
-    tl: ModuleType = np,
 ):
     """Calculate eigenvectors of the Laplacian. These are used as basis vectors in the HSGP
     approximation.
     """
     m_star = int(np.prod(m))
-    phi = tl.ones((Xs.shape[0], m_star))
+
+    phi = pt.ones((Xs.shape[0], m_star))
     for d in range(len(m)):
-        c = 1.0 / tl.sqrt(L[d])
-        term1 = tl.sqrt(eigvals[:, d])
-        term2 = tl.tile(Xs[:, d][:, None], m_star) + L[d]
-        phi *= c * tl.sin(term1 * term2)
+        c = 1.0 / pt.sqrt(L[d])
+        term1 = pt.sqrt(eigvals[:, d])
+        term2 = pt.tile(Xs[:, d][:, None], m_star) + L[d]
+        phi *= c * pt.sin(term1 * term2)
+
     return phi
 
 
@@ -332,8 +333,8 @@ class HSGP(Base):
         else:
             self.L = self._L
 
-        eigvals = calc_eigenvalues(self.L, self._m, tl=pt)
-        phi = calc_eigenvectors(Xs, self.L, eigvals, self._m, tl=pt)
+        eigvals = calc_eigenvalues(self.L, self._m)
+        phi = calc_eigenvectors(Xs, self.L, eigvals, self._m)
         omega = pt.sqrt(eigvals)
         psd = self.cov_func.power_spectral_density(omega)
 
@@ -385,8 +386,8 @@ class HSGP(Base):
 
         Xnew, _ = self.cov_func._slice(Xnew)
 
-        eigvals = calc_eigenvalues(self.L, self._m, tl=pt)
-        phi = calc_eigenvectors(Xnew - X_mean, self.L, eigvals, self._m, tl=pt)
+        eigvals = calc_eigenvalues(self.L, self._m)
+        phi = calc_eigenvectors(Xnew - X_mean, self.L, eigvals, self._m)
         i = int(self._drop_first is True)
 
         if self._parameterization == "noncentered":
--- a/pymc/gp/hsgp_approx.py
+++ b/pymc/gp/hsgp_approx.py
@@ -123,8 +123,8 @@ class HSGP(Base):
         Default `False`. Sometimes the first basis vector is quite "flat" and very similar to
         the intercept term.  When there is an intercept in the model, ignoring the first basis
         vector may improve sampling. This argument will be deprecated in future versions.
-    parameterization: str
-        Whether to use the `centered` or `noncentered` parameterization when multiplying the
+    parametrization: str
+        Whether to use the `centered` or `noncentered` parametrization when multiplying the
         basis by the coefficients.
     cov_func: Covariance function, must be an instance of `Stationary` and implement a
         `power_spectral_density` method.
@@ -179,7 +179,7 @@ class HSGP(Base):
         L: Sequence[float] | None = None,
         c: numbers.Real | None = None,
         drop_first: bool = False,
-        parameterization: str | None = "noncentered",
+        parametrization: str | None = "noncentered",
         *,
         mean_func: Mean = Zero(),
         cov_func: Covariance,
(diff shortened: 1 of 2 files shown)


AlexAndorra:
The failing test on Windows isn't related to these changes 🤷‍♂️ 
All the tests are passing

pymc/gp/hsgp_approx.py
```diff
@@ -30,39 +30,42 @@
...
 
-def set_boundary(Xs: TensorLike, c: numbers.Real | TensorLike) -> TensorLike:
+def set_boundary(Xs: TensorLike, c: numbers.Real | TensorLike) -> np.ndarray:
     """Set the boundary using the mean-subtracted `Xs` and `c`.  `c` is usually a scalar
-    multiplyer greater than 1.0, but it may be one value per dimension or column of `Xs`.
+    multiplier greater than 1.0, but it may be one value per dimension or column of `Xs`.
     """
-    S = pt.max(pt.abs(Xs), axis=0)
-    L = c * S
+    S = pt.max(Xs, axis=0)
```
juanitorduz: What would happen if `Xs` is always negative? say `[-1, -0.8, -0.6]`? If `c` is positive then `L` will be negative. Shall we keep the `abs`?

pymc/gp/hsgp_approx.py
```diff
@@ -85,6 +88,67 @@ def calc_basis_periodic(
...
+
+    S = (x_range[1] - x_range[0]) / 2
+
+    if cov_func.lower() == "expquad":
+        a1, a2 = 3.2, 1.75
+
+    elif cov_func.lower() == "matern52":
+        a1, a2 = 4.1, 2.65
+
+    elif cov_func.lower() == "matern32":
```
juanitorduz, line 154: Shall we raise an error with a friendly error message if `cov_function` does not match the expected ones?

pymc/gp/hsgp_approx.py
```diff
@@ -85,6 +88,67 @@ def calc_basis_periodic(
...
+        a1, a2 = 4.5, 3.42
+
+    c = max(a1 * (lengthscale_range[1] / S), 1.2)
+    m = int(a2 * c / (lengthscale_range[0] / S))
+
+    return m, c, S
```
juanitorduz: Just a suggestion: what about returning a [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) so that we do not depend on the special order of the parameters? something like

```python
HSGPParam = namedtuple('HSGPParam ', ['m', 'c', 'S'])
```

pymc/gp/hsgp_approx.py
```diff
@@ -252,8 +317,8 @@ def prior_linearized(self, Xs: TensorLike):
 
         Correct results when using `prior_linearized` in tandem with `pm.set_data` and
         `pm.Data` require two conditions.  First, one must specify `L` instead of `c` when
-        the GP is constructed.  If not, a RuntimeError is raised.  Second, the `Xs` needs to be
-        zero-centered, so its mean must be subtracted.  An example is given below.
+        the GP is constructed.  If not, a RuntimeError is raised.  Second, the `Xs` needs to be  ### CHECH WITH c INSTEAD!!
```
juanitorduz: Do you want to remove `### CHECH WITH c INSTEAD!!` ?

juanitorduz: (same below)

tests/gp/test_hsgp_approx.py
```diff
@@ -108,12 +108,12 @@ def model(self):
 class TestHSGP(_BaseFixtures):
     def test_set_boundaries_1d(self, X1):
         X1s = X1 - np.mean(X1, axis=0)
-        L = pm.gp.hsgp_approx.set_boundary(X1s, c=2).eval()
+        L = pm.gp.hsgp_approx.set_boundary(X1s, c=2)
```
juanitorduz: Can we parametrize the test to make sure is cover the cases when `Xs` is always negative and when it contains negative values and positive ones?

juanitorduz (commented):
This looks great! Just some minor suggestions :) 

Maybe I missed but I missed, but is there a test where we verify the mean sustraction works as expected in the out-of-sample case?

pymc/gp/hsgp_approx.py
AlexAndorra: Oooooh, you're right! Adding it back

pymc/gp/hsgp_approx.py
AlexAndorra: Good point. Done

pymc/gp/hsgp_approx.py
AlexAndorra: Good idea. Done

pymc/gp/hsgp_approx.py
AlexAndorra: Good catch

tests/gp/test_hsgp_approx.py
AlexAndorra: Done

AlexAndorra:
Thanks @juanitorduz ! I just pushed all the changes.

> is there a test where we verify the mean sustraction works as expected in the out-of-sample case?

Good point. Added it

pymc/gp/hsgp_approx.py
```diff
@@ -218,12 +296,13 @@ def __init__(
 
         self._drop_first = drop_first
         self._m = m
-        self._m_star = int(np.prod(self._m))
+        self._m_star = self.n_basis_vectors = int(np.prod(self._m))
```
juanitorduz, line 301: is this double `=` intentional?

```python
  self._m_star = self.n_basis_vectors = int(np.prod(self._m))
```

juanitorduz (approved):
Looks great! Just left a comment just to make sure its intentional and not a typo ;)

pymc/gp/hsgp_approx.py
AlexAndorra: Yes, it's a double assignment (if I remember the technical term correctly)

juanitorduz: ok! just double checking ;)

pymc/gp/hsgp_approx.py
bwengals: Personally think this is over-engineering, because introducing a new `HSGPParam` type serves no purpose other than to compensate for a perceived defect with the Python language.  I think it was clearer before.

pymc/gp/hsgp_approx.py
```diff
@@ -315,12 +383,17 @@ def prior_linearized(self, Xs: TensorLike):
...
+                pm.set_data({"X": x_new[:, None]})
 
             # and then make predictions for the GP using posterior predictive sampling.
             with model:
                 ppc = pm.sample_posterior_predictive(idata, var_names=["f"])
         """
+        # Important: fix the computation of the mean. If X is mutated later,
+        # the training mean will be subtracted, not the testing mean.
+        if self._X_mean is None:
+            self._X_mean = pt.mean(Xs, axis=0).eval()
```
bwengals: Hate to delay, and OK if it waits till a later PR, but it'd be better to use the midpoint of X here instead of the mean.  So, 
```python
self._X_center = (pt.max(Xs, axis=0) - pt.min(Xs, axis=0)) / 2
```

pymc/gp/hsgp_approx.py
```diff
@@ -385,14 +469,14 @@ def _build_conditional(self, Xnew):
...
         Xnew, _ = self.cov_func._slice(Xnew)
 
-        eigvals = calc_eigenvalues(self.L, self._m, tl=pt)
-        phi = calc_eigenvectors(Xnew - X_mean, self.L, eigvals, self._m, tl=pt)
+        eigvals = calc_eigenvalues(self.L, self._m)
+        phi = calc_eigenvectors(Xnew - X_mean, self.L, eigvals, self._m)
         i = int(self._drop_first is True)
 
-        if self._parameterization == "noncentered":
+        if self._parametrization == "noncentered":
```
fonnesbeck, line 476: Need to undo this spelling error, and the one below.

fonnesbeck, line 478: FWIW, either spelling is correct.

pymc/gp/hsgp_approx.py
AlexAndorra: On the contrary, it's great to have your input @bwengals !
Why is that though? For `S` I see why, but for the eigen vectors, all the examples before were using the mean, not the center if I'm not mistaken 🤔

pymc/gp/hsgp_approx.py
AlexAndorra: Is it?? Damn, all these changes for nothing then 😅

pymc/gp/hsgp_approx.py
bwengals: yah they were, was a mistake I think.  I think for the same reason as S, so those line up.  Then the accuracy of the approximation doesn't depend on the density of the x locations.

pymc/gp/hsgp_approx.py
juanitorduz: No strong opinion here. I just mess up indices so often that I prefer going safe with names instead 🙈

pymc/gp/hsgp_approx.py
juanitorduz: Wow ... this is an interesting point! Imagine X consists of many points around zero and a single point at 100. If we want the box dimension `L` to be as small as possible (that is the whole aim right?) then we should not take the mean but the mid-point indeed 🤔

AlexAndorra: Mind blown @bwengals 🤯 But that makes total sense now! Even better that we're doing the transformation under the hood now, otherwise we'd have to change a lot of examples...
Just pushed the changes. Feel free to merge if it looks good now

bwengals: > Even better that we're doing the transformation under the hood now, otherwise we'd have to change a lot of examples... 

Definitely, awesome that you got that working now

bwengals (approved):

bwengals:
Looks like there's one failing test mentioning `_X_mean`,

```python
         assert np.allclose(
            gp._X_mean, original_mean
>       ), "gp._X_mean should not change after updating data for out-of-sample predictions."
E       AttributeError: 'HSGP' object has no attribute '_X_mean'
```

pymc/gp/hsgp_approx.py
bwengals: yah but I would just expect people to unpack like, `m, c, S = ...`.  

Which makes me think it might make sense (in this or a later PR) to slightly refactor this so that there's a different function,
```python
S = calc_X_radius(X) # or calc_halfwidth or something
```
that could be used later where `_X_mean` was replaced.  Then this function only needs to return `m` and `c` and not also `S`, since that's handled now.

pymc/gp/hsgp_approx.py
```diff
@@ -30,39 +31,42 @@
...
 
 
-def set_boundary(Xs: TensorLike, c: numbers.Real | TensorLike) -> TensorLike:
-    """Set the boundary using the mean-subtracted `Xs` and `c`.  `c` is usually a scalar
-    multiplyer greater than 1.0, but it may be one value per dimension or column of `Xs`.
+def set_boundary(Xs: TensorLike, c: numbers.Real | TensorLike) -> np.ndarray:
+    """Set the boundary using the mean-subtracted `Xs` and `c`. `c` is usually a scalar
+    multiplier greater than 1.0, but it may be one value per dimension or column of `Xs`.
     """
     S = pt.max(pt.abs(Xs), axis=0)
```
AlexAndorra: Actually, shouldn't this also be using the midpoint @bwengals  (i.e `self._X_center`):
```python
S = (pt.max(Xs, axis=0) - pt.min(Xs, axis=0)) / 2
```
The current formula seems to compute the _largest magnitude_ of X, _not_ its half range, does it?

pymc/gp/hsgp_approx.py
```diff
@@ -85,6 +89,82 @@ def calc_basis_periodic(
...
+    ----------
+    - Ruitort-Mayol, G., Anderson, M., Solin, A., Vehtari, A. (2022).
+    Practical Hilbert Space Approximate Bayesian Gaussian Processes for Probabilistic Programming
+    """
+    if lengthscale_range[0] >= lengthscale_range[1]:
+        raise ValueError("One of the boundaries out of order")
+
+    X_center = (np.max(x, axis=0) - np.min(x, axis=0)) / 2
+    Xs = x - X_center
+    S = np.max(np.abs(Xs), axis=0)
```
AlexAndorra, line 146: After talking to @bwengals , I changed the computation here, to make sure it matches the definition of `S` in `set_boundary`.
The previous version was only doing;
```python
S = (x_range[1] - x_range[0]) / 2
```
If that looks good, I think we're good to merge 🥳

bwengals: yah good change.  Hit the merge button!

bwengals (approved):

pymc/gp/hsgp_approx.py
```diff
@@ -85,6 +89,82 @@ def calc_basis_periodic(
...
+    based on recommendations from Ruitort-Mayol et. al.
+
+    In practice, you need to choose `c` large enough to handle the largest lengthscales,
+    and `m` large enough to accommodate the smallest lengthscales.
+
+    NB: These recommendations are based on a one-dimensional GP.
+
+    Parameters
+    ----------
+    x_range : List[float]
```
bwengals: just noticed that this bit of the docstring is out of date.  looks like the function takes x instead of x_range now

bwengals: Since it's x now, it'd be good to remind people in the docstring that this `x` should be the x values you'll want to predict over, not just the training x.

AlexAndorra: Good catch @bwengals ! Just pushed the change. Will merge once it's all green

PR #7342, opened 2024-06-01
HSGP misc fixes

bwengals:
## Description
- Fix centering calculation (replace - with +) (nice catch @ulfaslak!)
- Use `x_range` instead of `x` in `approx_hsgp_hyperparams`.  I don't _think_ the full `x` is needed anymore, could be missing something though.
- Add more detail in docstring of `approx_hsgp_hyperparams` about usage.
- Renamed `Xs` to `X`.  The "s" stood for "scaled", for when the user had to center or mean subtract the incoming `X`.  Now this is done automatically, so it can just be called `X`.  Though minor, **it is a breaking change**.  Would it be good to add a deprecation warning and keep the `Xs` syntax for a while?

Also, didn't make this change, but what do people think about not returning `S` from `approx_hsgp_hyperparams`, so it would just return `m` and `c`?  The idea was that returning S here would be helpful for the user when centering things, but that's not really necessary anymore.

----
📚 Documentation preview 📚: https://pymc--7342.org.readthedocs.build/en/7342/

### DIFF
--- a/pymc/gp/hsgp_approx.py
+++ b/pymc/gp/hsgp_approx.py
@@ -96,23 +96,31 @@ class HSGPParams(NamedTuple):
 
 
 def approx_hsgp_hyperparams(
-    x: np.ndarray, lengthscale_range: list[float], cov_func: str
+    x_range: list[float], lengthscale_range: list[float], cov_func: str
 ) -> HSGPParams:
     """Utility function that uses heuristics to recommend minimum `m` and `c` values,
     based on recommendations from Ruitort-Mayol et. al.
 
     In practice, you need to choose `c` large enough to handle the largest lengthscales,
-    and `m` large enough to accommodate the smallest lengthscales.
+    and `m` large enough to accommodate the smallest lengthscales.  Use your prior on the 
+    lengthscale as guidance for setting the prior range.  For example, if you believe 
+    that 95% of the prior mass of the lengthscale is between 1 and 5, set the
+    `lengthscale_range` to be [1, 5], or maybe a touch wider.
+
+    Also, be sure to pass in an `x` that is exemplary of the domain not just of your 
+    training data, but also where you intend to make predictions.  For instance, if your
+    training x values are from [0, 10], and you intend to predict from [7, 15], you can 
+    pass in `x_range = [0, 15]`.
 
     NB: These recommendations are based on a one-dimensional GP.
 
     Parameters
     ----------
-    x : np.ndarray
-        The input variable on which the GP is going to be evaluated.
-        Careful: should be the X values you want to predict over, not *only* the training X.
+    x_range : list[float]
+        The range of the x values you intend to both train and predict over.  Should be a list with
+        two elements, [x_min, x_max].
     lengthscale_range : List[float]
-        The range of the lengthscales. Should be a list with two elements [lengthscale_min, lengthscale_max].
+        The range of the lengthscales. Should be a list with two elements, [lengthscale_min, lengthscale_max].
     cov_func : str
         The covariance function to use. Supported options are "expquad", "matern52", and "matern32".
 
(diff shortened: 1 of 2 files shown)


AlexAndorra:
Thannks @bwengals !

> Renamed Xs to X. [...] Would it be good to add a deprecation warning and keep the Xs syntax for a while?

I think it's fine because it's gonna get used more now that we have examples and tutorials up, so better do the change now than later.

> What do people think about not returning S from approx_hsgp_hyperparams, so it would just return m and c?

Agreed, we can get rid of that, as it makes the API simpler. Then we can probably get rid of the NamedTuple?
Note that all those changes need to be added to the [tutorial part 1](https://github.com/pymc-devs/pymc-examples/blob/main/examples/gaussian_processes/HSGP-Basic.ipynb) that we just published.

pymc/gp/hsgp_approx.py
```diff
@@ -96,23 +96,31 @@ class HSGPParams(NamedTuple):
...
     based on recommendations from Ruitort-Mayol et. al.
 
     In practice, you need to choose `c` large enough to handle the largest lengthscales,
-    and `m` large enough to accommodate the smallest lengthscales.
+    and `m` large enough to accommodate the smallest lengthscales.  Use your prior on the
+    lengthscale as guidance for setting the prior range.  For example, if you believe
+    that 95% of the prior mass of the lengthscale is between 1 and 5, set the
+    `lengthscale_range` to be [1, 5], or maybe a touch wider.
+
+    Also, be sure to pass in an `x` that is exemplary of the domain not just of your
```
AlexAndorra:
```suggestion
    Also, be sure to pass in an `x_range` that is exemplary of the domain not just of your
```

AlexAndorra (changes requested):
Almost all good @bwengals !
Just one change I highlighted, plus getting rid of returning `S` in the approx function (as well as updating the NB on pymc-examples)

pymc/gp/hsgp_approx.py
```diff
@@ -394,8 +396,8 @@ def prior_linearized(self, Xs: TensorLike):
         # Important: fix the computation of the midpoint of X.
         # If X is mutated later, the training midpoint will be subtracted, not the testing one.
         if self._X_center is None:
-            self._X_center = (pt.max(Xs, axis=0) - pt.min(Xs, axis=0)).eval() / 2
-        Xs = Xs - self._X_center  # center for accurate computation
+            self._X_center = (pt.max(X, axis=0) + pt.min(X, axis=0)).eval() / 2
+        Xs = X - self._X_center  # center for accurate computation
```
juanitorduz, line 400: **Small suggestion:** Could a unit test be added to check this so that we are sure future changes won't break the signs? We could just wrap this little logic into a small axillary function so that we can easily test it.

juanitorduz:
This looks great! Thanks! I agree with Alex's comments. I just left a small suggestion above :)

pymc/gp/hsgp_approx.py
```diff
@@ -32,10 +31,13 @@
...
-    """Set the boundary using the `Xs` centered around 0 and `c`. `c` is usually a scalar
-    multiplier greater than 1.0, but it may be one value per dimension or column of `Xs`.
+    """Set the boundary using the `Xs` and `c`.  `Xs` must be centered around zero, and `c`
+    is usually a scalar multiplier greater than 1.0, but it may also be one value per dimension
+    or column of `Xs`.
     """
-    S = pt.max(pt.abs(Xs), axis=0)  # important: the Xs should be centered around 0
+    S = pt.max(
+        pt.abs(Xs), axis=0
+    )  # important: the Xs should have previously been centered around 0
```
AlexAndorra: And if we used the formula for computing the radius instead? It's more generic, so it'll work and give the same results, whether the Xs have been centered or not. That's one possibly confusing place less.
`S = (pt.max(X, axis=0) - pt.min(X, axis=0)) / 2.0`

AlexAndorra (changes requested):
Just have a last question @bwengals

pymc/gp/hsgp_approx.py
AlexAndorra: What do you think @bwengals ? After that, I think we're good to merge

bwengals:
Sounds good, I can make these changes on Friday

pymc/gp/hsgp_approx.py
AlexAndorra: Just implemented this

AlexAndorra (approved):

PR #7345, opened 2024-06-03
Implement specialized MvNormal density based on precision matrix

ricardoV94:
## Description

This PR is exploring a specialized logp for a MvNormal (and possible MvStudentT) parametrized directly in terms of tau. According to  common model implementation looks like:

```python
import pymc as pm
import numpy as np

A = np.array([
    [0, 1, 1],
    [1, 0, 1], 
    [1, 1, 0]
])
D = A.sum(axis=-1)
np.testing.assert_allclose(A, A.T), "should be symmetric"

with pm.Model() as m:
    tau = pm.InverseGamma("tau", 1, 1)
    alpha = pm.Beta("alpha", 10, 10)
    Q = tau * (D - alpha * A)
    y = pm.MvNormal("y", mu=np.zeros(3), tau=Q)
```

## TODO (some are optional for this PR)
- [x] Benchmark to confirm it's more performant than old form where there's a matrix inverse and a cholesky decomposition (now we have a slogdet that does LU factorization under the hood)
- [x] Benchmark gradients as well
- [x] Check whether CAR is redundant with this?
- [x] Add tests
- [ ] Consider extending to MvStudentT
- [x] Explore whether something based on Cholesky decomposition still makes (more) sense here. CC @aseyboldt 
- [x] ~~Sparse implementation? May need some ideas like: https://stackoverflow.com/questions/19107617/how-to-compute-scipy-sparse-matrix-determinant-without-turning-it-to-dense~~ Investigate in a follow up PR

## Related Issue

- [ ] Related to https://github.com/pymc-devs/pymc-experimental/issues/340

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

CC @theorashid @elizavetasemenova

----
📚 Documentation preview 📚: https://pymc--7345.org.readthedocs.build/en/7345/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -24,6 +24,7 @@ import pytensor
 import pytensor.tensor as pt
 import scipy
 
+from pytensor.graph import node_rewriter
 from pytensor.graph.basic import Apply, Variable
 from pytensor.graph.op import Op
 from pytensor.raise_op import Assert
@@ -39,7 +40,7 @@ from pytensor.tensor.elemwise import DimShuffle
 from pytensor.tensor.exceptions import NotScalarConstantError
 from pytensor.tensor.linalg import cholesky, det, eigh, solve_triangular, trace
 from pytensor.tensor.linalg import inv as matrix_inverse
-from pytensor.tensor.random.basic import dirichlet, multinomial, multivariate_normal
+from pytensor.tensor.random.basic import MvNormalRV, dirichlet, multinomial, multivariate_normal
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.utils import (
     broadcast_params,
@@ -77,6 +78,9 @@ from pymc.distributions.shape_utils import (
 )
 from pymc.distributions.transforms import Interval, ZeroSumTransform, _default_transform
 from pymc.logprob.abstract import _logprob
+from pymc.logprob.rewriting import (
+    specialization_ir_rewrites_db,
+)
 from pymc.math import kron_diag, kron_dot
 from pymc.pytensorf import normalize_rng_param
 from pymc.util import check_dist_not_registered
@@ -157,6 +161,13 @@ def quaddist_matrix(cov=None, chol=None, tau=None, lower=True, *args, **kwargs):
     return cov
 
 
+def _logdet_from_cholesky(chol: TensorVariable) -> tuple[TensorVariable, TensorVariable]:
+    diag = pt.diagonal(chol, axis1=-2, axis2=-1)
+    logdet = pt.log(diag).sum(axis=-1)
+    posdef = pt.all(diag > 0, axis=-1)
+    return logdet, posdef
+
+
 def quaddist_chol(value, mu, cov):
     """Compute (x - mu).T @ Sigma^-1 @ (x - mu) and the logdet of Sigma."""
     if value.ndim == 0:
@@ -167,23 +178,23 @@ def quaddist_chol(value, mu, cov):
     else:
         onedim = False
 
-    delta = value - mu
     chol_cov = nan_lower_cholesky(cov)
+    logdet, posdef = _logdet_from_cholesky(chol_cov)
 
-    diag = pt.diagonal(chol_cov, axis1=-2, axis2=-1)
-    # Check if the covariance matrix is positive definite.
-    ok = pt.all(diag > 0, axis=-1)
-    # If not, replace the diagonal. We return -inf later, but
-    # need to prevent solve_lower from throwing an exception.
-    chol_cov = pt.switch(ok[..., None, None], chol_cov, 1)
+    # solve_triangular will raise if there are nans
+    # (which happens if the cholesky fails)
+    chol_cov.dprint(print_type=True, depth=1)
+    posdef.dprint(print_type=True, depth=1)
+    chol_cov = pt.switch(posdef[..., None, None], chol_cov, 1)
+
+    delta = value - mu
     delta_trans = solve_lower(chol_cov, delta, b_ndim=1)
     quaddist = (delta_trans**2).sum(axis=-1)
-    logdet = pt.log(diag).sum(axis=-1)
 
     if onedim:
-        return quaddist[0], logdet, ok
+        return quaddist[0], logdet, posdef
     else:
-        return quaddist, logdet, ok
+        return quaddist, logdet, posdef
 
 
 class MvNormal(Continuous):
@@ -283,14 +294,78 @@ class MvNormal(Continuous):
         -------
         TensorVariable
         """
-        quaddist, logdet, ok = quaddist_chol(value, mu, cov)
+        quaddist, logdet, posdef = quaddist_chol(value, mu, cov)
         k = value.shape[-1].astype("floatX")
         norm = -0.5 * k * np.log(2 * np.pi)
         return check_parameters(
             norm - 0.5 * quaddist - logdet,
-            ok,
-            msg="posdef",
+            posdef,
+            msg="posdef covariance",
         )
+
+
+class PrecisionMvNormalRV(SymbolicRandomVariable):
+    r"""A specialized multivariate normal random variable defined in terms of precision.
+
+    This class is introduced during specialization logprob rewrites, and not meant to be used directly.
+    """
+
+    name = "precision_multivariate_normal"
+    extended_signature = "[rng],[size],(n),(n,n)->(n)"
+    _print_name = ("PrecisionMultivariateNormal", "\\operatorname{PrecisionMultivariateNormal}")
+
+    @classmethod
+    def rv_op(cls, mean, tau, *, rng=None, size=None):
+        rng = normalize_rng_param(rng)
+        size = normalize_size_param(size)
+        cov = pt.linalg.inv(tau)
+        next_rng, draws = multivariate_normal(mean, cov, size=size, rng=rng).owner.outputs
+        return cls(
+            inputs=[rng, size, mean, tau],
+            outputs=[next_rng, draws],
+        )(rng, size, mean, tau)
+
+
+@_logprob.register
+def precision_mv_normal_logp(op: PrecisionMvNormalRV, value, rng, size, mean, tau, **kwargs):
+    [value] = value
+    k = value.shape[-1].astype("floatX")
+
+    delta = value - mean
+    quadratic_form = delta.T @ tau @ delta
+    logdet, posdef = _logdet_from_cholesky(nan_lower_cholesky(tau))
+    logp = -0.5 * (k * pt.log(2 * np.pi) + quadratic_form) + logdet
+
+    return check_parameters(
+        logp,
+        posdef,
+        msg="posdef precision",
+    )
+
+
+@node_rewriter(tracks=[MvNormalRV])
+def mv_normal_to_precision_mv_normal(fgraph, node):
+    """Replaces MvNormal(mu, inv(tau)) -> PrecisionMvNormal(mu, tau)
+
+    This is introduced in logprob rewrites to provide a more efficient logp for a MvNormal
+    that is defined by a precision matrix.
+
+    Note: This won't be introduced when calling `pm.logp` as that will dispatch directly
+    without triggering the logprob rewrites.
+    """
+
+    rng, size, mu, cov = node.inputs
+    if cov.owner and cov.owner.op == matrix_inverse:
+        tau = cov.owner.inputs[0]
+        return PrecisionMvNormalRV.rv_op(mu, tau, size=size, rng=rng).owner.outputs
+    return None
+
+
+specialization_ir_rewrites_db.register(
+    mv_normal_to_precision_mv_normal.__name__,
+    mv_normal_to_precision_mv_normal,
+    "basic",
+)
 
 
 class MvStudentTRV(RandomVariable):
--- a/pymc/logprob/rewriting.py
+++ b/pymc/logprob/rewriting.py
@@ -60,6 +60,7 @@ from pytensor.graph.rewriting.basic import (
     out2in,
 )
 from pytensor.graph.rewriting.db import (
+    EquilibriumDB,
     LocalGroupDB,
     RewriteDatabase,
     RewriteDatabaseQuery,
@@ -378,6 +379,14 @@ logprob_rewrites_db.register("measurable_ir_rewrites", measurable_ir_rewrites_db
 # "up" through the random/measurable variables and into their inputs.
 measurable_ir_rewrites_db.register("subtensor_lift", local_subtensor_rv_lift, "basic")
 measurable_ir_rewrites_db.register("incsubtensor_lift", incsubtensor_rv_replace, "basic")
+
+# These rewrites are used to introduce specalized operations with better logprob graphs
+specialization_ir_rewrites_db = EquilibriumDB()
+specialization_ir_rewrites_db.name = "specialization_ir_rewrites_db"
+logprob_rewrites_db.register(
+    "specialization_ir_rewrites_db", specialization_ir_rewrites_db, "basic"
+)
+
 
 logprob_rewrites_db.register("post-canonicalize", optdb.query("+canonicalize"), "basic")
 
--- a/tests/distributions/test_multivariate.py
+++ b/tests/distributions/test_multivariate.py
@@ -26,11 +26,13 @@ import scipy.stats as st
 from pytensor import tensor as pt
 from pytensor.tensor import TensorVariable
 from pytensor.tensor.blockwise import Blockwise
+from pytensor.tensor.nlinalg import MatrixInverse
 from pytensor.tensor.random.utils import broadcast_params
 from pytensor.tensor.slinalg import Cholesky
 
 import pymc as pm
 
+from pymc import Model
 from pymc.distributions.multivariate import (
     MultivariateIntervalTransform,
     _LKJCholeskyCov,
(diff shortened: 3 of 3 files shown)


ricardoV94:
Implementation checks may fail until https://github.com/pymc-devs/pytensor/issues/799 is fixed

ricardoV94:
Benchmark code

```python
import pymc as pm
import numpy as np
import pytensor
import pytensor.tensor as pt

n = 100
L = rng.uniform(low=0.1, high=1.0, size=(n,n))
Sigma = L @ L.T
Q_test = np.linalg.inv(Sigma)
x_test = rng.normal(size=n)

with pm.Model(check_bounds=False) as m:
    Q = pm.Data("Q", Q_test)
    x = pm.MvNormal("x", mu=pt.zeros(n), tau=Q)

logp_fn = m.compile_logp().f
logp_fn.trust_input=True
print("logp")
pytensor.dprint(logp_fn)


dlogp_fn = m.compile_dlogp().f
dlogp_fn.trust_input=True
print("dlogp")
pytensor.dprint(dlogp_fn)

np.testing.assert_allclose(logp_fn(x_test), np.array(-1789.93662205))

np.testing.assert_allclose(np.sum(dlogp_fn(x_test) ** 2), np.array(18445204.8755109), rtol=1e-6)

# Before: 2.66 ms
# After: 1.31 ms
%timeit -n 1000 logp_fn(x_test)

# Before: 2.45 ms
# After: 72 µs
%timeit -n 1000 dlogp_fn(x_test)
```

ricardoV94:
Final question is just whether we want / can do a similar thing for the MvStudentT. Otherwise it's ready to merge on my end

pymc/distributions/multivariate.py
```diff
@@ -289,10 +293,72 @@ def logp(value, mu, cov):
...
+            outputs=[next_rng, draws],
+        )(rng, size, mean, tau)
+
+
+@_logprob.register
+def precision_mv_normal_logp(op: PrecisionMvNormalRV, value, rng, size, mean, tau, **kwargs):
+    [value] = value
+    k = value.shape[-1]
+    delta = value - mean
+    det_sign, logdet_tau = pt.nlinalg.slogdet(tau)
```
aseyboldt: I'm not sure if the slogdet is necessarily a good idea here. Internally at least in numpy this does a lu factorization, which is I think takes theoretically about twice as long as the cholesky, and should usually be less stable. (But I think the performance can differ a lot based on number of threads and blas). So for a matrix that is not constant this might be slower than the usual MvNormal right now.

So I think it is better to use the cholesky decomposition here and use that to get the logdet.

ricardoV94: In the example above the matrix is not constant, so you can use it to benchmark

ricardoV94: Here: https://github.com/pymc-devs/pymc/pull/7345#issuecomment-2182930220

ricardoV94: And it twas 2x faster logp and 100x faster dlogp on my crappy PC. Are you skeptical of those numbers, or you think we can do even better?

ricardoV94: Are you suggesting something like this https://math.stackexchange.com/questions/2001041/logarithm-of-the-determinant-of-a-positive-definite-matrix ?

aseyboldt: Yes, `logdet = -2*pt.log(pt.diag(cholesky(inv_cov))).sum()`.

The benchmark code uses a constant tau though, or am I missing something?

ricardoV94: No it is a shared variable (pm.Data) which is not constant for PyTensor.

I set the logp to:
```python
    [value] = value
    k = value.shape[-1]
    delta = value - mean
    logdet_tau = 2 * pt.trace((pt.log(pt.linalg.cholesky(tau))), axis1=-2, axis2=-1)
    # det_sign, logdet_tau = pt.nlinalg.slogdet(tau)
    quadratic_form = delta.T @ tau @ delta
    logp = -0.5 * (k * pt.log(2 * pt.pi) - logdet_tau + quadratic_form)
    return check_parameters(
        logp,
        # det_sign > 0,
        logdet_tau > 0,
        msg="posdef precision",
    )
```

And then ran this:
```python
%env OMP_NUM_THREADS=1

import pymc as pm
import numpy as np
import pytensor
import pytensor.tensor as pt

n = 100
L = rng.uniform(low=0.1, high=1.0, size=(n,n))
Sigma = L @ L.T
Q_test = np.linalg.inv(Sigma)
x_test = rng.normal(size=n)

with pm.Model(check_bounds=False) as m:
    Q = pm.Data("Q", Q_test)
    x = pm.MvNormal("x", mu=pt.zeros(n), tau=Q)

logp_fn = m.compile_logp().f
logp_fn.trust_input=True
print("logp")
pytensor.dprint(logp_fn)

dlogp_fn = m.compile_dlogp().f
dlogp_fn.trust_input=True
print("dlogp")
# pytensor.dprint(dlogp_fn)

np.testing.assert_allclose(logp_fn(x_test), np.array(-1789.93662205))

np.testing.assert_allclose(np.sum(dlogp_fn(x_test) ** 2), np.array(18445204.8755109), rtol=1e-6)

# With slogdet: 318 µs ± 42.5 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# With cholesky logdet: 310 µs ± 35.7 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
%timeit -n 10000 logp_fn(x_test)

# Before: 2.45 ms
# With slogdet: # 31 µs ± 2.43 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# With cholesky logdet: 32.5 µs ± 4.51 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
%timeit -n 10000 dlogp_fn(x_test)
```
At least for this example it's pretty mach a draw on my machine. Do you get different results?

ricardoV94: the logdet > 0 thing is silly haha, but anyway we are running without check_bounds so it doesn't show in the graph

aseyboldt: Here would be some code with a non-constant matrix, but I get a NotImplementederror for the grad of Blockwise?

```
import pymc as pm
import numpy as np
import pytensor
import pytensor.tensor as pt

n = 100
L = rng.uniform(low=0.1, high=1.0, size=(n,n))
Sigma = L @ L.T
Q_test = np.linalg.inv(Sigma)
x_test = rng.normal(size=n)
v_test = rng.normal(size=n)

with pm.Model(check_bounds=False) as m:
    Q = pm.Data("Q", Q_test)
    v = pm.Normal("v", shape=n)
    x = pm.MvNormal("x", mu=pt.zeros(n), tau=Q + v[None, :] * v[:, None])

logp_fn = m.compile_logp().f
logp_fn.trust_input=True
print("logp")
#pytensor.dprint(logp_fn)


dlogp_fn = m.compile_dlogp().f
dlogp_fn.trust_input=True
print("dlogp")
#pytensor.dprint(dlogp_fn)

#np.testing.assert_allclose(logp_fn(x_test, v_test), np.array(-1789.93662205))

#np.testing.assert_allclose(np.sum(dlogp_fn(x_test, v_test) ** 2), np.array(18445204.8755109), rtol=1e-6)
with threadpoolctl.threadpool_limits(1):

    # Before: 2.66 ms
    # After: 1.31 ms
    %timeit logp_fn(x_test, v_test)
    
    # Before: 2.45 ms
    # After: 72 µs
    %timeit dlogp_fn(x_test, v_test)
```

ricardoV94: Your expression moves the log above the diagonal, which makes sense, but I think the minus in `-2 * ` is wrong?

aseyboldt: The performance of the original code looks I think pretty bad by the way:
```
#353 μs ± 4.53 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
#1.61 ms ± 32.5 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
If everything is implemented well I don't think there is a good reason why the gradient should be much slower than just the logp. Factoring the matrx onces should be enough.
I guess this is because we compute the pullback of the cholesky or so, even though the pullback of the logdet is actually pretty easy if you already have a factorization...
There might be a nice usecase for an OpFromGraph and overwriting the forward values hiding here somewhere...

ricardoV94: Seems to be 40us (1.2x faster) with the cholesky now that I only do log on the diagonal. Funny enough it's fusing the log and the sum, which is our only elemwise + reduce fusion we have :)

```python
logdet_tau = 2 * pt.log(pt.diagonal(pt.linalg.cholesky(tau), axis1=-2, axis2=-1)).sum()
```

```python
%env OMP_NUM_THREADS=1

import pymc as pm
import numpy as np
import pytensor
import pytensor.tensor as pt

n = 100
L = rng.uniform(low=0.1, high=1.0, size=(n,n))
Sigma = L @ L.T
Q_test = np.linalg.inv(Sigma)
x_test = rng.normal(size=n)

with pm.Model(check_bounds=False) as m:
    Q = pm.Data("Q", Q_test)
    x = pm.MvNormal("x", mu=pt.zeros(n), tau=Q)

logp_fn = m.compile_logp().f
logp_fn.trust_input=True
print("logp")
pytensor.dprint(logp_fn)

dlogp_fn = m.compile_dlogp().f
dlogp_fn.trust_input=True
print("dlogp")
pytensor.dprint(dlogp_fn)

np.testing.assert_allclose(logp_fn(x_test), np.array(-1789.93662205))

np.testing.assert_allclose(np.sum(dlogp_fn(x_test) ** 2), np.array(18445204.8755109), rtol=1e-6)

# With slogdet: 236 µs ± 31.7 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# With cholesky logdet: 192 µs ± 25.6 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
%timeit -n 10000 logp_fn(x_test)

# With slogdet: 29.8 µs ± 3.19 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
# With cholesky logdet: 32.5 µs ± 4.51 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
%timeit -n 10000 dlogp_fn(x_test)
```

ricardoV94: Pushed a commit with the cholesky factorization

pymc/distributions/multivariate.py
```diff
@@ -289,10 +293,76 @@ def logp(value, mu, cov):
...
+    # Equivalent, but possibly faster / more stable than slogdet, which relies on LU factorization
+    # det_sign, logdet_tau = pt.nlinalg.slogdet(tau)
+    cholesky_diagonal = pt.diagonal(pt.linalg.cholesky(tau), axis1=-2, axis2=-1)
+    logdet = 2 * pt.log(cholesky_diagonal).sum()
+
+    quadratic_form = delta.T @ tau @ delta
+    logp = -0.5 * (k * pt.log(2 * pt.pi) - logdet + quadratic_form)
+    return check_parameters(
+        logp,
+        (cholesky_diagonal > 0).all(-1),
```
ricardoV94: I assume this is the right check for posdef-ness? @aseyboldt

aseyboldt: The cholesky will simply fail (throw an exception) if the matrix is not posdef. I think we check for that in the perform method and return nan, but I don't know what for instance jax will do.

ricardoV94: Okay so we could check for nan? I'll see what JAX does

ricardoV94: I guess all(nan) will also be false so this ends up being the same thing? Or does it evaluate to nan...?

I'll check

ricardoV94:
Last benchmarks, running the following script:

```python
%env OMP_NUM_THREADS=1

USE_TAU = True

import pymc as pm
import numpy as np
import pytensor
import pytensor.tensor as pt

n = 100
L = rng.uniform(low=0.1, high=1.0, size=(n,n))
Sigma = L @ L.T
Q_test = np.linalg.inv(Sigma)
x_test = rng.normal(size=n)

with pm.Model(check_bounds=False) as m:
    Q = pm.Data("Q", Q_test)
    if USE_TAU:
        x = pm.MvNormal("x", mu=pt.zeros(n), tau=Q)
    else:
        x = pm.MvNormal("x", mu=pt.zeros(n), cov=Q)

logp_fn = m.compile_logp().f
logp_fn.trust_input=True
print("logp")
pytensor.dprint(logp_fn)

dlogp_fn = m.compile_dlogp().f
dlogp_fn.trust_input=True
print("dlogp")
pytensor.dprint(dlogp_fn)

%timeit -n 10000 logp_fn(x_test)
%timeit -n 10000 dlogp_fn(x_test)
```

## USE_TAU = TRUE, befare the PR:
```
logp
Composite{((i2 - (i0 * i1)) - i3)} [id A] 'x_logprob' 8
 ├─ 0.5 [id B]
 ├─ DropDims{axis=0} [id C] 7
 │  └─ CAReduce{Composite{(i0 + sqr(i1))}, axis=1} [id D] 6
 │     └─ Transpose{axes=[1, 0]} [id E] 4
 │        └─ SolveTriangular{trans=0, unit_diagonal=False, lower=True, check_finite=True, b_ndim=2} [id F] 2
 │           ├─ Cholesky{lower=True, destructive=False, on_error='nan'} [id G] 1
 │           │  └─ Q [id H]
 │           └─ ExpandDims{axis=1} [id I] 0
 │              └─ x [id J]
 ├─ -91.89385332046727 [id K]
 └─ CAReduce{Composite{(i0 + log(i1))}, axes=None} [id L] 5
    └─ ExtractDiag{offset=0, axis1=0, axis2=1, view=False} [id M] 3
       └─ Cholesky{lower=True, destructive=False, on_error='nan'} [id G] 1
          └─ ···

dlogp
DropDims{axis=1} [id A] '(dx_logprob/dx)' 6
 └─ SolveTriangular{trans=0, unit_diagonal=False, lower=False, check_finite=True, b_ndim=2} [id B] 5
    ├─ Transpose{axes=[1, 0]} [id C] 4
    │  └─ Cholesky{lower=True, destructive=False, on_error='nan'} [id D] 1
    │     └─ Q [id E]
    └─ Neg [id F] 3
       └─ SolveTriangular{trans=0, unit_diagonal=False, lower=True, check_finite=True, b_ndim=2} [id G] 2
          ├─ Cholesky{lower=True, destructive=False, on_error='nan'} [id D] 1
          │  └─ ···
          └─ ExpandDims{axis=1} [id H] 0
             └─ x [id I]
262 µs ± 57.8 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
277 µs ± 54.2 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

## USE_TAU = True after the PR

```
logp
Composite{(i4 * ((i2 - (i0 * i1)) + i3))} [id A] 'x_logprob' 11
 ├─ 2.0 [id B]
 ├─ CAReduce{Composite{(i0 + log(i1))}, axes=None} [id C] 10
 │  └─ ExtractDiag{offset=0, axis1=0, axis2=1, view=False} [id D] 9
 │     └─ Cholesky{lower=True, destructive=False, on_error='raise'} [id E] 8
 │        └─ Q [id F]
 ├─ 183.78770664093454 [id G]
 ├─ DropDims{axis=0} [id H] 7
 │  └─ CGemv{inplace} [id I] 6
 │     ├─ AllocEmpty{dtype='float64'} [id J] 5
 │     │  └─ 1 [id K]
 │     ├─ 1.0 [id L]
 │     ├─ ExpandDims{axis=0} [id M] 4
 │     │  └─ x [id N]
 │     ├─ CGemv{inplace} [id O] 3
 │     │  ├─ AllocEmpty{dtype='float64'} [id P] 2
 │     │  │  └─ Shape_i{1} [id Q] 1
 │     │  │     └─ Q [id F]
 │     │  ├─ 1.0 [id L]
 │     │  ├─ Transpose{axes=[1, 0]} [id R] 'Q.T' 0
 │     │  │  └─ Q [id F]
 │     │  ├─ x [id N]
 │     │  └─ 0.0 [id S]
 │     └─ 0.0 [id S]
 └─ -0.5 [id T]

dlogp
CGemv{inplace} [id A] '(dx_logprob/dx)' 5
 ├─ CGemv{inplace} [id B] 4
 │  ├─ AllocEmpty{dtype='float64'} [id C] 3
 │  │  └─ Shape_i{0} [id D] 2
 │  │     └─ Q [id E]
 │  ├─ 1.0 [id F]
 │  ├─ Q [id E]
 │  ├─ Mul [id G] 1
 │  │  ├─ [-0.5] [id H]
 │  │  └─ x [id I]
 │  └─ 0.0 [id J]
 ├─ -0.5 [id K]
 ├─ Transpose{axes=[1, 0]} [id L] 'Q.T' 0
 │  └─ Q [id E]
 ├─ x [id I]
 └─ 1.0 [id F]
160 µs ± 34.1 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
15.9 µs ± 2.29 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

## For reference: USE_TAU = False before and after (unchanged)
Before:
```
...
260 µs ± 13 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
275 µs ± 19.5 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

After:
```
...
259 µs ± 46.3 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
275 µs ± 30.6 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

ricardoV94:
`numpyro` tests are failing probably because it now requires the more recent versions of jax. should be fixed by #7407

ricardoV94:
@aseyboldt any thing that should block this PR?

aseyboldt:
Looks good. I think it is possible that we could further improve the MvNormal in both parametrizations, but this is definetly an improvement as it is.
Most of all I think we should do the same for the CholeskyMvNormal. Looks like we are just computing the cov just to re-compute the cholesky again? At some point we did make use of the cholesky directly, but I guess that got lost in a refactor wtih the pytensor RVs?

aseyboldt (approved):

ricardoV94:
> Looks good. I think it is possible that we could further improve the MvNormal in both parametrizations, but this is definetly an improvement as it is.
> Most of all I think we should do the same for the CholeskyMvNormal. Looks like we are just computing the cov just to re-compute the cholesky again? At some point we did make use of the cholesky directly, but I guess that got lost in a refactor wtih the pytensor RVs?

We don't recompute the Cholesky, we have rewrites to remove it and even a specific test for it: https://github.com/pymc-devs/pymc/blob/b407c01accc65a68d17803b6208d6dfdb0e40877/tests/distributions/test_multivariate.py#L2368

PR #7349, opened 2024-06-05
Fix bug with `compute_p` for partially observed OrderedLogistic and OrderedProbit variables

ricardoV94:
This fixes a bug when auto-imputation takes place, in which case the logic to retrieve `p` wasn't valid. This will also show `p` as an input of the underlying Categorical in the graphviz.

## Description

This PR fixes a bug when setting `compute_p=True` for partially observed OrderedLogistic/Probit. The logic of retrieving `p` from the third input does not work in this case, because when there is imputation a different kind of variable is returned (the joined vector of imputed and observed variables)

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7349.org.readthedocs.build/en/7349/

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -1197,30 +1197,9 @@ class _OrderedLogistic(Categorical):
     See docs for the OrderedLogistic wrapper class for more details on how to use it in models.
     """
 
-    rv_op = categorical
-
-    @classmethod
-    def dist(cls, eta, cutpoints, *args, **kwargs):
-        eta = pt.as_tensor_variable(eta)
-        cutpoints = pt.as_tensor_variable(cutpoints)
-
-        pa = sigmoid(cutpoints - pt.shape_padright(eta))
-        p_cum = pt.concatenate(
-            [
-                pt.zeros_like(pt.shape_padright(pa[..., 0])),
-                pa,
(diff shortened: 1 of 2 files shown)


pymc/distributions/discrete.py
```diff
@@ -1287,50 +1266,39 @@ class OrderedLogistic:
         plt.hist(posterior["cutpoints"][1], 80, alpha=0.2, color='k');
     """
 
-    def __new__(cls, name, *args, compute_p=True, **kwargs):
-        out_rv = _OrderedLogistic(name, *args, **kwargs)
+    def __new__(cls, name, eta, cutpoints, compute_p=True, **kwargs):
+        p = cls.compute_p(eta, cutpoints)
         if compute_p:
-            pm.Deterministic(f"{name}_probs", out_rv.owner.inputs[3], dims=kwargs.get("dims"))
+            p = pm.Deterministic(f"{name}_probs", p, dims=kwargs.get("dims"))
```
ricardoV94, line 1272: The changes have the nice side-effect that `p` will now be an arrow going into the underlying Categorical

AlexAndorra (approved):
Good catch @ricardoV94 !

Issue #7348, opened 2024-06-04
ENH: Support foward sampling via JAX


jessegrabowski:
### Before

```python
import pymc as pm
from pymc.model.transform.optimization import freeze_dims_and_data

with pm.Model() as m:
    ...

with freeze_dims_and_data(m):
    idata_prior = pm.sample_prior_predictive(compile_kwargs={"mode":"JAX"})

with m:
    idata = pm.sample()

with freeze_dims_and_data(m):
    idata = pm.sample_posterior_predictive(idata, extend_inferencedata=True, compile_kwargs = {"mode":"JAX"})
```

### After

```python
import pymc as pm

with pm.Model() as m:
   ...
   idata_prior = pm.sample_prior_predictive(mode="JAX")
   idata = pm.sample()
   idata = pm.sample_posterior_predictive(idata, mode="JAX", extend_inferencedata=True)
```

### Context for the issue:

For models involving scan or multivariate-normal distributions, you get big speedups by passing `compile_kwargs={'mode':'JAX'}` to `pm.sample_prior_predictive` and `pm.sample_posterior_predictive`. This has already proven useful in statespace modeling (in pymc-experimental, https://github.com/pymc-devs/pymc-experimental/pull/346) and instrumental variable modeling (in casualpy, https://github.com/pymc-labs/CausalPy/pull/345). In each of these cases using the JAX backend offers *significant* speedups, and is a highly desirable feature.

This was technically never a supported feature, but it could be made to work by consciously specifying the whole model to be static (e.g. using `pm.ConstantData` and avoiding `mutable_kwargs`). After #7047 this is obviously no longer possible. The work-around is to use `freeze_dims_and_data`, but this is somewhat cumbersome, especially with prior predictive sampling, where a typical workflow calls `pm.sample_prior_predictive` in the model block at construction time. I have also come up with cases where `freeze_dims_and_data` fails. A trivial example is in predictive modeling using `pm.Flat` dummies -- this adds non-`None` entries to `model.rvs_to_initial_values`, causing `model_to_fgraph` to fail. 

My proposal would be to simply add a "freezing" step to `compile_forward_sampling_function`. This would alleviate the need for users to be aware of the `freeze_dims_and_data` helper function, allow JAX forward sampling without breaking out of a single model context, and also support any future backend that requires all shape information to be known.

[...]

Issue #7268, opened 2024-04-19
Allow extracting logp and random functions directly from fgraph representation


ricardoV94:
### Description

The PyMC Model omnipresent context makes it hard to transform models under the hood, because you can't create a new model inside another model without registering it's variables in the outer model.

Issues like #7263 could be avoided by using the recommended helper without users awareness, except that helper can't be called inside a PyMC model context as it requires recreating the model again if we want to extract the logp from it. Instead if we extracted the logp from the fgraph representation directly we wouldn't need to create a new model and so it could be done safely inside the outer context.

PR #7352, opened 2024-06-10
Allow opting out of model nesting by setting `model=None`

### LINKED ISSUE #6715: DOC: Provide a working example of inheriting from pymc.Model

The documentation for [`pymc.Model`](https://www.pymc.io/projects/docs/en/stable/api/generated/pymc.Model.html) seems to give some conceptual gist of how it works, but the example given in the documentation [does not work](https://stackoverflow.com/questions/76224256/what-is-the-model-argument-in-the-super-init-in-the-example-subclassing/76224316?noredirect=1#comment134420040_76224316).

There are no imports, and I am not sure where `pt` is coming from in `pt.constant(1)`.  I have thought maybe `pt` was for "PyTensor" but there is no `constant` in the namespace of `pt` resulting from `import pytensor as pt`. There is a `pymc.math.constant`, but I am unsure if that is the same thing.

But it is also unclear if the imports are the only issue. Here is some code where I have attempted to address the above namespace issues.

```python
from pymc import *

class CustomModel(Model):
    # 1) override init
    def __init__(self, mean=0, sigma=1, name=''):
        # 2) call super's init first, passing model and name
        # to it name will be prefix for all variables here if
        # no name specified for model there will be no prefix
        super().__init__(name, model)
        # now you are in the context of instance,
        # `modelcontext` will return self you can define
        # variables in several ways note, that all variables
        # will get model's name prefix

        # 3) you can create variables with the register_rv method
        self.register_rv(Normal.dist(mu=mean, sigma=sigma), 'v1', initval=1)
        # this will create variable named like '{name::}v1'
        # and assign attribute 'v1' to instance created
        # variable can be accessed with self.v1 or self['v1']

        # 4) this syntax will also work as we are in the
        # context of instance itself, names are given as usual
        Normal('v2', mu=mean, sigma=sigma)

        # something more complex is allowed, too
        half_cauchy = HalfCauchy('sigma', beta=10, initval=1.)
        Normal('v3', mu=mean, sigma=half_cauchy)

        # Deterministic variables can be used in usual way
        Deterministic('v3_sq', self.v3 ** 2)

        # Potentials too
        Potential('p1', pt.constant(1))

# After defining a class CustomModel you can use it in several
# ways

# I:
#   state the model within a context
with Model() as model:
    CustomModel()
    # arbitrary actions

# II:
#   use new class as entering point in context
with CustomModel() as model:
    Normal('new_normal_var', mu=1, sigma=0)

# III:
#   just get model instance with all that was defined in it
model = CustomModel()

# IV:
#   use many custom models within one context
with Model() as model:
    CustomModel(mean=1, name='first')
    CustomModel(mean=2, name='second')

# variables inside both scopes will be named like `first::*`, `second::*`
```

Running the above I get an `AttributeError` that I don't think is related because it occurs at `super().__init__(name, model)`.

```python
  File "/usr/lib/python3.10/idlelib/run.py", line 578, in runcode
    exec(code, self.locals)
  File "/home/galen/pymc_inheritance_question.py", line 42, in <module>
    CustomModel()
  File "/home/galen/.local/lib/python3.10/site-packages/pymc/model.py", line 264, in __call__
    instance.__init__(*args, **kwargs)
  File "/home/galen/pymc_inheritance_question.py", line 10, in __init__
    super().__init__(name, model)
  File "/home/galen/.local/lib/python3.10/site-packages/pymc/model.py", line 588, in __init__
    self.add_coords(coords)
  File "/home/galen/.local/lib/python3.10/site-packages/pymc/model.py", line 1080, in add_coords
    for name, values in coords.items():
AttributeError: 'Model' object has no attribute 'items'
```

Please provide a minimum working example.

ricardoV94:
## Description

This PR allows setting `model=None` to explicitly disable model nesting. This is needed for model transformations `do, observe, clone, remove_value_transforms, freeze_rv_and_dims` to be possible under a model context without side-effects on the original model.

I still think model nesting is an overkill, but a lot of code out there depends on it so I am not touching it.

I also updated the docstrings with more relevant (imo) examples. The previous example was broken as per #6715 and I am not sure we really want to incentivize people to subclass PyMC models. That certainly shouldn't be the focus of the docstrings?

Finally, removed deprecated functionality.

## Related Issue

- [x] Closes #6715 
- [x] Related to #7268 
- [x] Related to https://github.com/pymc-devs/pymc-experimental/pull/345#discussion_r1627311093
- [x] Related to #7348

## Checklist

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7352.org.readthedocs.build/en/7352/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -107,18 +107,10 @@ class ContextMeta(type):
 
         def __enter__(self):
             self.__class__.context_class.get_contexts().append(self)
-            # self._pytensor_config is set in Model.__new__
-            self._config_context = None
-            if hasattr(self, "_pytensor_config"):
-                self._config_context = pytensor.config.change_flags(**self._pytensor_config)
-                self._config_context.__enter__()
             return self
 
         def __exit__(self, typ, value, traceback):
             self.__class__.context_class.get_contexts().pop()
-            # self._pytensor_config is set in Model.__new__
-            if self._config_context:
-                self._config_context.__exit__(typ, value, traceback)
 
         dct[__enter__.__name__] = __enter__
         dct[__exit__.__name__] = __exit__
@@ -485,13 +477,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
             instance._parent = kwargs.get("model")
         else:
             instance._parent = cls.get_context(error_if_none=False)
-        pytensor_config = kwargs.get("pytensor_config", {})
-        if pytensor_config:
-            warnings.warn(
-                "pytensor_config is deprecated. Use pytensor.config or pytensor.config.change_flags context manager instead.",
-                FutureWarning,
-            )
-        instance._pytensor_config = pytensor_config
         return instance
 
     @staticmethod
@@ -507,10 +492,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
         check_bounds=True,
         *,
         coords_mutable=None,
-        pytensor_config=None,
         model=None,
     ):
-        del pytensor_config, model  # used in __new__
+        del model  # used in __new__
         self.name = self._validate_name(name)
         self.check_bounds = check_bounds
 
@@ -559,11 +543,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
         self._repr_latex_ = types.MethodType(
             functools.partial(str_for_model, formatting="latex"), self
         )
-
-    @property
-    def model(self):
-        warnings.warn("Model.model property is deprecated. Just use Model.", FutureWarning)
-        return self
 
     @property
     def parent(self):
@@ -671,7 +650,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         jacobian : bool
             Whether to include jacobian terms in logprob graph. Defaults to True.
         """
-        return self.model.compile_fn(
+        return self.compile_fn(
             self.d2logp(vars=vars, jacobian=jacobian, negate_output=negate_output),
             **compile_kwargs,
         )
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -392,76 +392,103 @@ class Model(WithMemoization, metaclass=ContextMeta):
     name : str
         name that will be used as prefix for names of all random
         variables defined within model
+    coords : dict
+        Xarray-like coordinate keys and values. These coordinates can be used
+        to specify the shape of random variables and to label (but not specify)
+        the shape of Determinsitic, Potential and Data objects.
+        Other than specifying the shape of random variables, coordinates have no
+        effect on the model. They can't be used for label-based broadcasting or indexing.
+        You must use numpy-like operations for those behaviors.
     check_bounds : bool
         Ensure that input parameters to distributions are in a valid
         range. If your model is built in a way where you know your
         parameters can only take on valid values you can set this to
         False for increased speed. This should not be used if your model
         contains discrete variables.
+    model : PyMC model, optional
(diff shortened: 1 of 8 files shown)


pymc/sampling/deterministic.py
```diff
@@ -83,7 +83,7 @@ def compute_deterministics(
     model = modelcontext(model)
 
     if var_names is None:
-        deterministics = model.deterministics
+        deterministics = list(model.deterministics)
```
ricardoV94, line 86: To appease mypy

pymc/stats/log_density.py
```diff
@@ -126,46 +128,35 @@ def compute_log_density(
     if kind not in ("likelihood", "prior"):
         raise ValueError("kind must be either 'likelihood' or 'prior'")
 
+    # We need to disable transforms, because the InferenceData only keeps the untransformed values
+    umodel = remove_value_transforms(model)
```
ricardoV94, line 132: This kind of change is why this PR was done. A similar thing was taking place in https://github.com/pymc-devs/pymc-experimental/pull/345#discussion_r1627311093

pymc/model/core.py
```diff
@@ -400,76 +393,99 @@ class Model(WithMemoization, metaclass=ContextMeta):
     name : str
         name that will be used as prefix for names of all random
         variables defined within model
+    coords: dict
```
OriolAbril:
```suggestion
    coords : dict
```

pymc/model/core.py
```diff
@@ -400,76 +393,99 @@ class Model(WithMemoization, metaclass=ContextMeta):
...
+        is created inside another model's context, the parent model will be set to that model.
+        If `None` the model will not have a parent.
 
     Examples
     --------
-    How to define a custom model
+    Use context manager to define model and respective variables
 
     .. code-block:: python
+        import pymc as pm
```
OriolAbril: I think you need a blank line for this to render correctly. The preview is missing these code blocks: https://pymcio--7352.org.readthedocs.build/projects/docs/en/7352/api/model/generated/pymc.model.core.Model.html

pymc/model/core.py
```diff
@@ -400,76 +393,99 @@ class Model(WithMemoization, metaclass=ContextMeta):
...
+
+        model = pm.Model()
+        x = pm.Normal("x", model=model)
+
+
+    Use coords for defining the shape of random variables and labeling other model variables
+
+    .. code-block:: python
+    import pymc as pm
```
OriolAbril: Here in addition to the black line the code itself must be indented

pymc/model/core.py
```diff
@@ -400,76 +393,99 @@ class Model(WithMemoization, metaclass=ContextMeta):
...
+    import numpy as np
+
+    coords = {"feature", ["A", "B", "C"], "trial", [1, 2, 3, 4, 5]}
```
michaelosthege:
```suggestion
    coords = {
        "feature", ["A", "B", "C"],
        "trial", [1, 2, 3, 4, 5],
    }
```
a little easier to read

michaelosthege (approved):
I wonder how `.to_graphviz` looks for hierarchical models 🤔

Please apply the docstring fixes that Oriol suggested 👍

ricardoV94:
> I wonder how .to_graphviz looks for hierarchical models 🤔

Doesn't do anything different for the outer model. You can just call `.to_graphviz` on submodels and will only show the variables that belong to it IIRC

PR #7357, opened 2024-06-15
Fix `import pymc.testing`

### LINKED ISSUE #7218: `pymc.testing` attempts to import `pytest` but `pytest` is not in `pymc` install requirements
### Describe the issue:

a41d5244b8ec78372751924ba0484ed02d57b2cb included the `testing.py` module at the root level of `pymc`. [setup.py](https://github.com/pymc-devs/pymc/blob/main/setup.py#L75) says that the install requires are taken from [requirements.txt](https://github.com/pymc-devs/pymc/blob/main/requirements.txt), and that file does not include `pytest`.

This makes `from pymc.testing import anything` fail unless you already have installed `pytest`. Should `pymc` raise some kind of informative message instead of saying `ModuleNotFoundError: No module named 'pytest'`?

```python
from pymc import testing
```

### Error message:

```shell
ModuleNotFoundError: No module named 'pytest'
```

### PyMC version information:

pymc version: 5.11.0

### Context for the issue:

_No response_

bomtall:
## Description

- Move `import pytest` to be local so it does not break `import pymc.testing`
- It is hard to write a test for this because the test will have `pytest` already

## Related Issue

- [x ] Closes #7218 
- [ ] Related to #

## Checklist

- [x ] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ x] Added necessary documentation (docstrings and/or example notebooks)

## Type of change

- [x ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7357.org.readthedocs.build/en/7357/

### DIFF
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -21,7 +21,6 @@ from typing import Any
 import numpy as np
 import pytensor
 import pytensor.tensor as pt
-import pytest
 
 from numpy import random as nr
 from numpy import testing as npt
@@ -342,6 +341,8 @@ def check_logp(
     scipy_args : Dictionary with extra arguments needed to call scipy logp method
         Usually the same as extra_args
     """
+    import pytest
+
     if decimal is None:
         decimal = select_by_precision(float64=6, float32=3)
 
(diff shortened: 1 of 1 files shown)


twiecki:
Thanks!

PR #7363, opened 2024-06-15
Move CustomDist logic to dedicated module and docs section

### LINKED ISSUE #7247: Move CustomDist logic to its own file
### Description

The logic of CustomDist is getting pretty large. After #7239 I would refactor it into it's own file (and same for tests).

### LINKED ISSUE #7373: Add CustomDist as second level header below Distributions in the API
### Description

Right now is tucked away inside `Distribution utilities`: 

image

We have been pushing for it to be a more common tool, so I would show it immediately one level above. May want to wait for #

markgreene74:
@Hemant19870601 and I worked on this PR during the PyMC Hackathon at PyData London 2024.

## Description
With this PR `CustomDist` is moved to a separate file (`custom_distribution.py`) along with related classes (`CustomDistRV`, `CustomDistRV`,  `CustomSymbolicDistRV`, `_CustomSymbolicDist`) and helper functions (`default_not_implemented`, `default_support_point`).

The imports in `distribution.py` and `__init__.py` have been updated to reflect the changes.

`DensityDist` has been moved to the new file as well.

Note that in `distribution.py` the import of `CustomSymbolicDistRV` has been added close to where the class is used to avoid circular import. A `noqa` has been added on the same line to avoid a `ruff` error for the misplaced import (`E402`).

Similarly, the `CustomDist` tests have been moved to their own file (`test_custom_distribution.py`) and are passing.

Test results (before/after):
| file | before (`main` branch)  | after |
| ----- | ----- | ----- |
| `test_distribution.py` | 71 pass/14 ignored/85 total  | 33 pass/0 ignored/33 total  |
| `test_custom_distribution.py` | N/A  | 38 pass/14 ignored/52 total  |
|   | 71 pass/14 ignored/85 total | 71 pass/14 ignored/85 total  |
## Related Issue
- [X] Closes #7247
- [x] Closes #7373
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [x] Documentation
- [X] Maintenance
- [X] Other (please specify): Refactor

----
📚 Documentation preview 📚: https://pymc--7363.org.readthedocs.build/en/7363/


tests/distributions/test_custom_distribution.py
```diff
@@ -0,0 +1,280 @@
...
+
+import cloudpickle
+import numpy as np
+import numpy.random as npr
+import pytensor
+import pytensor.tensor as pt
+import pytest
+import scipy.stats as st
+
+import pymc as pm
```
ricardoV94: Ideally we don't import pymc as pm, but import whatever functions are needed instead

tests/distributions/test_distribution.py
```diff
@@ -166,253 +164,6 @@ def test_all_distributions_have_support_points():
...
-        x = cloudpickle.loads(cloudpickle.dumps(x))
-
-        test_value = pm.draw(x, random_seed=1)
-        assert np.all(test_value == pm.draw(x, random_seed=1))
-
-        x_logp = pm.logp(x, test_value)
-        assert np.allclose(x_logp.eval(), st.norm(1).logpdf(test_value))
-
-
 class TestCustomSymbolicDist:
```
ricardoV94: These should also be moved

tests/distributions/test_custom_distribution.py
markgreene74: Hi @ricardoV94, had a look at this.

What we did was just copy over the imports needed from `test_distribution.py` to `test_custom_distribution.py` without altering them.

It seems to me that the import was done in this way in the original `test_distribution.py` to keep, for example, `pm.Normal` distinct from `pymc.distributions.Normal` (imported in [L38 here](https://github.com/pymc-devs/pymc/blob/main/tests/distributions/test_distribution.py#L38)).

In this regard see for example [L107](https://github.com/pymc-devs/pymc/blob/main/tests/distributions/test_distribution.py#L38) (uses `pm.Normal`) vs [L173](https://github.com/pymc-devs/pymc/blob/main/tests/distributions/test_distribution.py#L173) (uses `pymc.distribution.Normal` just as `Normal`).

I found a few other cases to which the above applies (`logp`, `Model`).

Thoughts? Shall I still go ahead?

tests/distributions/test_distribution.py
markgreene74: Done in 34e41e5

I will update the test results in the PR description with the new figures:
```
33/0/33 (total)
38/14/52 (total)
```

tests/distributions/test_custom_distribution.py
ricardoV94: Yeah that's a lazy and discouraged pattern, since we're moving things around it's a good time as any to clean it up and import things properly

markgreene74: I will add another commit shortly

tests/distributions/test_custom_distribution.py
```diff
@@ -0,0 +1,662 @@
...
+import numpy as np
+import numpy.random as npr
+import pytensor
+import pytensor.tensor as pt
+import pytest
+import scipy.stats as st
+
+from pytensor import scan
+
+from pymc import Bernoulli as pm_Bernoulli
```
ricardoV94: Why the alias? Should just be `from pymc.distributions import Bernoulli, Beta, ...`

ricardoV94 (dismissed):

tests/distributions/test_custom_distribution.py
markgreene74: I just realised that after pushing the commit, it's fixed in e365f4c where the alias is only applied to those imports that would conflict

markgreene74:
Hi @ricardoV94 thank you for all the comments :) I think this PR should be ready for review now, happy for me to change the status?

tests/distributions/test_custom_distribution.py
```diff
@@ -0,0 +1,664 @@
...
+import numpy as np
+import numpy.random as npr
+import pytensor
+import pytensor.tensor as pt
+import pytest
+import scipy.stats as st
+
+from pytensor import scan
+
+from pymc import (
```
ricardoV94: It's better to import more specifically like this.
```suggestion
from pymc.distributions import (
```

This will be merged with the existing 3 imports from distributions already happening below by the pre-commit

It doesn't apply to the last import in this tuple, so don't just apply the suggested changes

ricardoV94:
I rebased the PR on top of main to solve the conflicts with #7370

tests/distributions/test_custom_distribution.py
ricardoV94: I addressed that after the rebase, thanks!

twiecki (approved):

PR #7375, opened 2024-06-19
Convert ndarray to scalar to avoid deprecation warning

### LINKED ISSUE #7369: Numpy deprecation warning issued from within `test_metropolis.py`
### Description

```python
 tests/step_methods/test_metropolis.py: 15950 warnings
  D:\a\pymc\pymc\pymc\backends\ndarray.py:118: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    data[key][self.draw_idx] = val
```

prathamhole14:
## Description
This PR fixes a Numpy deprecation warning by converting ndarray values to scalars in the record method.

Changes-
Updated record method to convert ndarray to scalar using val.item() before assignment.
Reason-
Prevents future errors due to deprecated implicit array-to-scalar conversions in Numpy.

## Related Issue

- [x] Closes #7369
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7375.org.readthedocs.build/en/7375/


pymc/backends/ndarray.py
```diff
@@ -115,6 +115,8 @@ def record(self, point, sampler_stats=None) -> None:
         if sampler_stats is not None:
             for data, vars in zip(self._stats, sampler_stats):
                 for key, val in vars.items():
+                    if isinstance(val, np.ndarray) and val.ndim > 0:
```
ricardoV94, line 118: Is the val always a single number?

prathamhole14: Right, currently `val` is converted to a scalar only when it is a single-element `ndarray`
I think the correct implementation would be 
```
if isinstance(val, np.ndarray):
    # Extract the first element using indexing (preserves order)
    val = val.flat[0]  
data[key][self.draw_idx] = val
```

ricardoV94: I'm not sure these records are supposed to always be single elements. If not, discarding the rest doesn't make sense

PR #7380, opened 2024-06-21
Implement unconstraining transform for LKJCorr

### LINKED ISSUE #7101: BUG: LKJCorr breaks when used as covariance with MvNormal
### Describe the issue:

LKJCorr backwards sampling is currently broken and does not properly constrain values to a positive semidefinite matrix early enough. This leads to compilation randomly failing when providing it as a covariance matrix for MvNormal and other similar multivariate distributions. The failure is inherently random, but it gets increasingly likely with larger n values and fails almost always with n>=10 and can in theory be seen with n as low as 3.

Error message is informative, as it is easy to verify the matrix reported in error has negative eigenvalues, proving it is not PSD.

This only affects the backwards sampling (i.e. logp computations), with .eval() and predictive sampling working as they should. 

```python
import pymc as pm
import pytensor.tensor as pt
import numpy as np

n, eta = 10, 1

# Turn LKJCorr output to a proper square corr matrix
def corr_to_mat(corr_values, n):
    corr = pt.zeros((n, n))
    corr = pt.set_subtensor(corr[np.triu_indices(n, k=1)], corr_values)
    return corr + corr.T + pt.identity_like(corr)

# Generate Data
corr = corr_to_mat(pm.LKJCorr.dist(n=n, eta=eta),n)
y = pm.draw(pm.MvNormal.dist(mu=0, cov=corr), 100)

# PyMC Model
with pm.Model() as m:
    corr_values = pm.LKJCorr('corr_values', n=n, eta=eta)
    corr = corr_to_mat(corr_values,n)
    y_hat = pm.MvNormal('y_hat', mu=0, cov=corr, observed=y)
    idata = pm.sample()
```

### Error message:

```shell
  File "/home/velochy/pymc/test3.py", line 22, in <module>
    idata = pm.sample()
            ^^^^^^^^^^^
  File "/home/velochy/pymc/pymc/sampling/mcmc.py", line 744, in sample
    model.check_start_vals(ip)
  File "/home/velochy/pymc/pymc/model/core.py", line 1660, in check_start_vals
    raise SamplingError(
pymc.exceptions.SamplingError: Initial evaluation of model at starting point failed!
Starting values:
{'corr_values_interval__': array([ 0.96546568,  0.55627918, -0.52092415, -0.79551073, -0.04416735,
       -0.61208579, -0.04889871, -0.62181685,  0.13621748,  0.32908589,
       -0.38471896, -0.89088229, -0.1326561 , -0.18828583,  0.26747631,
       -0.47067211,  0.51480876, -0.56323033, -0.3513053 ,  0.30480191,
       -0.36865109,  0.1479414 , -0.66890036,  0.98484518, -0.85993115,
        0.48094929, -0.04547197, -0.07743477, -0.3381161 ,  0.12565052,
       -0.06450642, -0.69698417,  0.80246982, -0.67958155, -0.29175309,
       -0.02445876, -0.7858127 ,  0.17738997, -0.87779485, -0.27196011,
       -0.52231265,  0.46271978,  0.0810375 ,  0.84458683, -0.72285369])}

Logp initial evaluation results:
{'corr_values': -inf, 'y_hat': -inf}
You can call `model.debug()` for more details.
```

### PyMC version information:

pymc 5.10.3

### Context for the issue:

This arose in discussion with @ricardoV94 and @jessegrabowski in https://discourse.pymc.io/t/using-lkjcorr-together-with-mvnormal/13606/29 . Thread has more useful context

johncant:
I've ported [this bijector](https://github.com/tensorflow/probability/blob/94f592af363e13391858b48f785eb4c250912904/tensorflow_probability/python/bijectors/correlation_cholesky.py#L194) from `tensorflow` and added to `LKJCorr`. This ensures that initial samples drawn from `LKJCorr` are positive definite, which fixes #7101 . Sampling now completes successfully with no divergences.

There are several parts I'm not comfortable with:
  - https://github.com/johncant/pymc/blob/fix_lkjcorr_positive_definiteness/pymc/distributions/multivariate.py#L1583 - it seems like a bad idea to run a pytensor graph here. Is there any way to get the LKJCorr `n` parameter from `op` or `rv` without `eval`ing any pytensors?
  - https://github.com/johncant/pymc/blob/fix_lkjcorr_positive_definiteness/pymc/distributions/transforms.py#L176-L177 - not sure whether or not this is the right way to create a constant pytensor tensor here.

@fonnesbeck @twiecki @jessegrabowski @velochy - please could you take a look? I would like to make sure that this fix makes sense before adding tests and making the linters pass.

Notes:
  - Tests not yet written, linters not yet ran
  - The original tensorflow bijector is defined in the opposite sense to pymc transforms, i.e. `forward` in `tensorflow_probability` is `backward` in `pymc`
  - The original tensorflow bijector produces cholesky factors, not actual correlation matrices, so in this implementation, we have to do a cholesky decomposition in the forward transform.
  - In the tensorflow bijector, the triagonal elements of a matrix are filled in a clockwise spiral, as opposed to numpy which defines indices in a row-major order.

### Backward method
1. Start with identity matrix and fill lower triangular elements with unconstrained real numbers.
2. Normalize each row so the L-2 norm is 1
3. This is now a Cholesky factor that will always result in positive definite correlation matrices

### Forward method

1. Reconstruct the correlation matrix from its upper triangular elements
2. Perform cholesky decomposition to obtain L
3. The diagonal elements of L are multipliers we used to normalize the other elements.
4. Extract those diagonal elements and divide to undo the backward method

### log_jac_det

This was quite complicated to implement, so I used the symbolic jacobian.

## Related Issue
- [ ] Closes #7101 

## Checklist

## Type of change
- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7380.org.readthedocs.build/en/7380/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -43,7 +43,6 @@ from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.utils import (
     normalize_size_param,
 )
-from pytensor.tensor.type import TensorType
 from scipy import stats
 
 import pymc as pm
@@ -73,7 +72,11 @@ from pymc.distributions.shape_utils import (
     rv_size_is_none,
     to_tuple,
 )
-from pymc.distributions.transforms import Interval, ZeroSumTransform, _default_transform
+from pymc.distributions.transforms import (
+    CholeskyCorrTransform,
+    ZeroSumTransform,
+    _default_transform,
+)
 from pymc.logprob.abstract import _logprob
 from pymc.logprob.rewriting import (
     specialization_ir_rewrites_db,
@@ -1099,31 +1102,30 @@ def WishartBartlett(name, S, nu, is_cholesky=False, return_cholesky=False, initv
 
 
 def _lkj_normalizing_constant(eta, n):
-    # TODO: This is mixing python branching with the potentially symbolic n and eta variables
-    if not isinstance(eta, int | float):
-        raise NotImplementedError("eta must be an int or float")
-    if not isinstance(n, int):
-        raise NotImplementedError("n must be an integer")
-    if eta == 1:
-        result = gammaln(2.0 * pt.arange(1, int((n - 1) / 2) + 1)).sum()
-        if n % 2 == 1:
-            result += (
+    result_1 = gammaln(2.0 * pt.arange(1, ((n - 1) / 2) + 1)).sum()
+    result_2 = -(n - 1) * gammaln(eta + 0.5 * (n - 1))
+    k = pt.arange(1, n)
+
+    return pt.switch(
+        pt.eq(eta, 1.0),
+        pt.switch(
+            pt.eq(n % 2, 1.0),
+            result_1
+            + (
                 0.25 * (n**2 - 1) * pt.log(np.pi)
                 - 0.25 * (n - 1) ** 2 * pt.log(2.0)
-                - (n - 1) * gammaln(int((n + 1) / 2))
-            )
-        else:
-            result += (
+                - (n - 1) * gammaln((n + 1) / 2)
+            ),
+            result_1
+            + (
                 0.25 * n * (n - 2) * pt.log(np.pi)
                 + 0.25 * (3 * n**2 - 4 * n) * pt.log(2.0)
                 + n * gammaln(n / 2)
                 - (n - 1) * gammaln(n)
-            )
-    else:
-        result = -(n - 1) * gammaln(eta + 0.5 * (n - 1))
-        k = pt.arange(1, n)
-        result += (0.5 * k * pt.log(np.pi) + gammaln(eta + 0.5 * (n - 1 - k))).sum()
-    return result
+            ),
+        ),
+        result_2 + (0.5 * k * pt.log(np.pi) + gammaln(eta + 0.5 * (n - 1 - k))).sum(),
+    )
 
 
 # _LKJCholeskyCovBaseRV requires a properly shaped `D`, which means the variable can't
@@ -1133,11 +1135,11 @@ class _LKJCholeskyCovRV(SymbolicRandomVariable):
     _print_name = ("_lkjcholeskycov", "\\operatorname{_lkjcholeskycov}")
 
     @classmethod
-    def rv_op(cls, n, eta, sd_dist, *, size=None):
+    def rv_op(cls, n, eta, sd_dist, *, size=None, rng=None):
         # We don't allow passing `rng` because we don't fully control the rng of the components!
         n = pt.as_tensor(n, dtype="int64", ndim=0)
         eta = pt.as_tensor_variable(eta, ndim=0)
-        rng = pytensor.shared(np.random.default_rng())
+        rng = normalize_rng_param(rng)
         size = normalize_size_param(size)
 
         # We resize the sd_dist automatically so that it has (size x n) independent
@@ -1161,15 +1163,11 @@ class _LKJCholeskyCovRV(SymbolicRandomVariable):
         D = sd_dist.type(name="D")  # Make sd_dist opaque to OpFromGraph
         size = D.shape[:-1]
 
-        # We flatten the size to make operations easier, and then rebuild it
-        flat_size = pt.prod(size, dtype="int64")
-
-        next_rng, C = LKJCorrRV._random_corr_matrix(rng=rng, n=n, eta=eta, flat_size=flat_size)
-        D_matrix = D.reshape((flat_size, n))
-        C *= D_matrix[..., :, None] * D_matrix[..., None, :]
+        final_rng, C = LKJCorrRV._random_corr_matrix(n=n, eta=eta, size=size, rng=rng)
+        C = D[..., :, None] * C
 
         tril_idx = pt.tril_indices(n, k=0)
-        samples = pt.linalg.cholesky(C)[..., tril_idx[0], tril_idx[1]]
+        samples = C[..., tril_idx[0], tril_idx[1]]
 
         if rv_size_is_none(size):
             samples = samples[0]
@@ -1179,7 +1177,7 @@ class _LKJCholeskyCovRV(SymbolicRandomVariable):
 
         return _LKJCholeskyCovRV(
             inputs=[rng, n, eta, D],
-            outputs=[next_rng, samples],
+            outputs=[final_rng, samples],
         )(rng, n, eta, sd_dist)
 
     def update(self, node):
@@ -1211,7 +1209,7 @@ class _LKJCholeskyCov(Distribution):
 
 @_change_dist_size.register(_LKJCholeskyCovRV)
 def change_LKJCholeksyCovRV_size(op, dist, new_size, expand=False):
-    n, eta, sd_dist = dist.owner.inputs[1:]
+    _, n, eta, sd_dist = dist.owner.inputs
 
     if expand:
         old_size = sd_dist.shape[:-1]
@@ -1452,7 +1450,7 @@ class LKJCholeskyCov:
 
 class LKJCorrRV(SymbolicRandomVariable):
     name = "lkjcorr"
-    extended_signature = "[rng],[size],(),()->[rng],(n)"
+    extended_signature = "[rng],[size],(),()->[rng],(n,n)"
     _print_name = ("LKJCorrRV", "\\operatorname{LKJCorrRV}")
 
     def make_node(self, rng, size, n, eta):
@@ -1466,142 +1464,72 @@ class LKJCorrRV(SymbolicRandomVariable):
 
         return super().make_node(rng, size, n, eta)
 
+    def update(self, node):
+        return {node.inputs[0]: node.outputs[0]}
+
     @classmethod
-    def rv_op(cls, n: int, eta, *, rng=None, size=None):
-        # We flatten the size to make operations easier, and then rebuild it
+    def rv_op(cls, n: int, eta, *, size=None, rng=None):
         n = pt.as_tensor(n, ndim=0, dtype=int)
         eta = pt.as_tensor(eta, ndim=0)
         rng = normalize_rng_param(rng)
+
         size = normalize_size_param(size)
+        final_rng, C = cls._random_corr_matrix(rng=rng, n=n, eta=eta, size=size)
 
-        if rv_size_is_none(size):
-            flat_size = 1
-        else:
-            flat_size = pt.prod(size, dtype="int64")
-
-        next_rng, C = cls._random_corr_matrix(rng=rng, n=n, eta=eta, flat_size=flat_size)
-
-        triu_idx = pt.triu_indices(n, k=1)
-        samples = C[..., triu_idx[0], triu_idx[1]]
-
-        if rv_size_is_none(size):
-            samples = samples[0]
-        else:
-            dist_shape = (n * (n - 1)) // 2
-            samples = pt.reshape(samples, (*size, dist_shape))
-
-        return cls(
-            inputs=[rng, size, n, eta],
-            outputs=[next_rng, samples],
-        )(rng, size, n, eta)
-
-        return samples
+        return cls(inputs=[rng, size, n, eta], outputs=[final_rng, C])(rng, size, n, eta)
 
     @classmethod
     def _random_corr_matrix(
-        cls, rng: Variable, n: int, eta: TensorVariable, flat_size: TensorVariable
+        cls, n: int, eta: TensorVariable, size: TensorVariable, rng
     ) -> tuple[Variable, TensorVariable]:
-        # original implementation in R see:
-        # https://github.com/rmcelreath/rethinking/blob/master/R/distributions.r
+        size = () if rv_size_is_none(size) else size
 
-        beta = eta - 1.0 + n / 2.0
-        next_rng, beta_rvs = pt.random.beta(
-            alpha=beta, beta=beta, size=flat_size, rng=rng
-        ).owner.outputs
-        r12 = 2.0 * beta_rvs - 1.0
-        P = pt.full((flat_size, n, n), pt.eye(n))
-        P = P[..., 0, 1].set(r12)
-        P = P[..., 1, 1].set(pt.sqrt(1.0 - r12**2))
-        n = get_underlying_scalar_constant_value(n)
-        for mp1 in range(2, n):
-            beta -= 0.5
-            next_rng, y = pt.random.beta(
-                alpha=mp1 / 2.0, beta=beta, size=flat_size, rng=next_rng
+        beta0 = eta - 1.0 + n / 2.0
+
+        next_rng, y0 = pt.random.beta(alpha=beta0, beta=beta0, size=size, rng=rng).owner.outputs
+
+        r12 = 2.0 * y0 - 1.0
+
+        P0 = pt.full((*size, n, n), pt.eye(n))
+        P0 = P0[..., 0, 1].set(r12)
+        P0 = P0[..., 1, 1].set(pt.sqrt(1.0 - r12**2))
+
+        def step(mp1, beta, P, prev_rng):
+            beta_next = beta - 0.5
+
+            middle_rng, y = pt.random.beta(
+                alpha=mp1 / 2.0, beta=beta, size=size, rng=prev_rng
             ).owner.outputs
-            next_rng, z = pt.random.normal(
-                loc=0, scale=1, size=(flat_size, mp1), rng=next_rng
+
+            final_rng, z = pt.random.normal(
+                loc=0, scale=1, size=(*size, mp1), rng=middle_rng
             ).owner.outputs
-            z = z / pt.sqrt(pt.einsum("ij,ij->i", z, z.copy()))[..., np.newaxis]
-            P = P[..., 0:mp1, mp1].set(pt.sqrt(y[..., np.newaxis]) * z)
+
+            ein_sig_z = "i, i->" if z.ndim == 1 else "...ij, ...ij->...i"
+
+            z = z / pt.sqrt(pt.einsum(ein_sig_z, z, z.copy()))[..., None]
+            P = P[..., 0:mp1, mp1].set(pt.sqrt(y[..., None]) * z)
             P = P[..., mp1, mp1].set(pt.sqrt(1.0 - y))
-        C = pt.einsum("...ji,...jk->...ik", P, P.copy())
-        return next_rng, C
 
+            return beta_next, P, final_rng
 
-class MultivariateIntervalTransform(Interval):
-    name = "interval"
-
-    def log_jac_det(self, *args):
-        return super().log_jac_det(*args).sum(-1)
-
-
-# Returns list of upper triangular values
-class _LKJCorr(BoundedContinuous):
-    rv_type = LKJCorrRV
-    rv_op = LKJCorrRV.rv_op
-
-    @classmethod
-    def dist(cls, n, eta, **kwargs):
-        n = pt.as_tensor_variable(n).astype(int)
-        eta = pt.as_tensor_variable(eta)
-        return super().dist([n, eta], **kwargs)
-
-    def support_point(rv, *args):
-        return pt.zeros_like(rv)
-
-    def logp(value, n, eta):
-        """
-        Calculate logp of LKJ distribution at specified value.
-
-        Parameters
-        ----------
-        value: numeric
-            Value for which log-probability is calculated.
-
-        Returns
-        -------
-        TensorVariable
-        """
-        if value.ndim > 1:
-            raise NotImplementedError("LKJCorr logp is only implemented for vector values (ndim=1)")
-
-        # TODO: PyTensor does not have a `triu_indices`, so we can only work with constant
-        #  n (or else find a different expression)
-        try:
-            n = int(get_underlying_scalar_constant_value(n))
-        except NotScalarConstantError:
-            raise NotImplementedError("logp only implemented for constant `n`")
-
-        shape = n * (n - 1) // 2
-        tri_index = np.zeros((n, n), dtype="int32")
-        tri_index[np.triu_indices(n, k=1)] = np.arange(shape)
-        tri_index[np.triu_indices(n, k=1)[::-1]] = np.arange(shape)
-
-        value = pt.take(value, tri_index)
-        value = pt.fill_diagonal(value, 1)
-
-        # TODO: _lkj_normalizing_constant currently requires `eta` and `n` to be constants
-        try:
-            eta = float(get_underlying_scalar_constant_value(eta))
-        except NotScalarConstantError:
-            raise NotImplementedError("logp only implemented for constant `eta`")
-        result = _lkj_normalizing_constant(eta, n)
-        result += (eta - 1.0) * pt.log(det(value))
-        return check_parameters(
-            result,
-            value >= -1,
-            value <= 1,
-            matrix_pos_def(value),
-            eta > 0,
+        _, P_seq, final_rng = pytensor.scan(
+            fn=step,
+            sequences=[pt.arange(2, n)],
+            outputs_info=[beta0, P0, next_rng],
+            strict=True,
+            return_updates=False,
         )
 
+        # This is the concatenation of P_seq and the initial P0 (scan slices it away by default to be "helpful")
+        # Select the final element along the scan axis (this is the result)
+        P = P_seq.owner.inputs[0][-1]
+        # C = pt.einsum("...ji,...jk->...ik", P, P.copy())
 
-@_default_transform.register(_LKJCorr)
-def lkjcorr_default_transform(op, rv):
-    return MultivariateIntervalTransform(-1.0, 1.0)
+        return final_rng, P.mT
 
 
-class LKJCorr:
+class LKJCorr(BoundedContinuous):
     r"""
     The LKJ (Lewandowski, Kurowicka and Joe) distribution.
 
@@ -1621,10 +1549,6 @@ class LKJCorr:
         The shape parameter (eta > 0) of the LKJ distribution. eta = 1
         implies a uniform distribution of the correlation matrices;
         larger values put more weight on matrices with few correlations.
-    return_matrix : bool, default=False
-        If True, returns the full correlation matrix.
-        False only returns the values of the upper triangular matrix excluding
-        diagonal in a single vector of length n(n-1)/2 for memory efficiency
 
     Notes
     -----
@@ -1639,7 +1563,7 @@ class LKJCorr:
             # Define the vector of fixed standard deviations
             sds = 3 * np.ones(10)
 
-            corr = pm.LKJCorr("corr", eta=4, n=10, return_matrix=True)
+            corr = pm.LKJCorr("corr", eta=4, n=10)
 
             # Define a new MvNormal with the given correlation matrix
             vals = sds * pm.MvNormal("vals", mu=np.zeros(10), cov=corr, shape=10)
@@ -1648,10 +1572,6 @@ class LKJCorr:
             vals_raw = pm.Normal("vals_raw", shape=10)
             chol = pt.linalg.cholesky(corr)
             vals = sds * pt.dot(chol, vals_raw)
-
-            # The matrix is internally still sampled as a upper triangular vector
-            # If you want access to it in matrix form in the trace, add
-            pm.Deterministic("corr_mat", corr)
 
 
     References
@@ -1662,26 +1582,57 @@ class LKJCorr:
         100(9), pp.1989-2001.
     """
 
-    def __new__(cls, name, n, eta, *, return_matrix=False, **kwargs):
-        c_vec = _LKJCorr(name, eta=eta, n=n, **kwargs)
-        if not return_matrix:
-            return c_vec
-        else:
-            return cls.vec_to_corr_mat(c_vec, n)
+    rv_type = LKJCorrRV
+    rv_op = LKJCorrRV.rv_op
 
     @classmethod
-    def dist(cls, n, eta, *, return_matrix=False, **kwargs):
-        c_vec = _LKJCorr.dist(eta=eta, n=n, **kwargs)
-        if not return_matrix:
-            return c_vec
-        else:
-            return cls.vec_to_corr_mat(c_vec, n)
+    def dist(cls, n, eta, **kwargs):
+        n = pt.as_tensor_variable(n).astype(int)
+        eta = pt.as_tensor_variable(eta)
+        return super().dist([n, eta], **kwargs)
 
-    @classmethod
-    def vec_to_corr_mat(cls, vec, n):
-        tri = pt.zeros(pt.concatenate([vec.shape[:-1], (n, n)]))
-        tri = pt.subtensor.set_subtensor(tri[(..., *np.triu_indices(n, 1))], vec)
-        return tri + pt.moveaxis(tri, -2, -1) + pt.diag(pt.ones(n))
+    @staticmethod
+    def support_point(rv: TensorVariable, *args):
+        return pt.broadcast_to(pt.eye(rv.shape[-1]), rv.shape)
+
+    @staticmethod
+    def logp(value: TensorVariable, n, eta):
+        """
+        Calculate the logp of a correlation matrix under an LKJ distribution.
+
+        Parameters
+        ----------
+        value: numeric
+            Value for which log-probability is calculated.
+
+        Returns
+        -------
+        TensorVariable
+        """
+        # n has to be a constant, otherwise the shape of the RV would not be fixed between draws.
+        try:
+            n = int(get_underlying_scalar_constant_value(n))
+        except NotScalarConstantError:
+            raise NotImplementedError("logp only implemented for constant `n`")
+
+        result = _lkj_normalizing_constant(eta, n) + (eta - 1.0) * 2 * pt.diagonal(
+            value, axis1=-2, axis2=-1
+        ).log().sum(axis=-1)
+
+        row_norms = pt.sum(value**2, axis=-1)
+
+        return check_parameters(
+            result,
+            eta > 0,
+            pt.isclose(row_norms, 1.0),
+            msg="Invalid values passed to LKJCorr logp: value is not a valid correlationm matrix or eta <= 0.",
+        )
+
+
+@_default_transform.register(LKJCorr)
+def lkjcorr_default_transform(op, rv):
+    rng, shape, n, eta, *_ = rv.owner.inputs
+    return CholeskyCorrTransform(n=n, upper=False)
 
 
 class MatrixNormalRV(RandomVariable):
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -22,8 +22,7 @@ import pytensor.tensor as pt
 import scipy
 
 from pytensor.graph import node_rewriter
-from pytensor.graph.basic import Apply, Variable
-from pytensor.graph.op import Op
+from pytensor.graph.basic import Variable
 from pytensor.raise_op import Assert
 from pytensor.sparse.basic import DenseFromSparse
 from pytensor.sparse.math import sp_sum
@@ -897,50 +896,10 @@ class OrderedMultinomial:
         return _OrderedMultinomial.dist(*args, **kwargs)
 
 
-def posdef(AA):
-    try:
-        scipy.linalg.cholesky(AA)
-        return True
-    except scipy.linalg.LinAlgError:
-        return False
-
-
-class PosDefMatrix(Op):
-    """Check if input is positive definite. Input should be a square matrix."""
-
-    # Properties attribute
-    __props__ = ()
-
-    # Compulsory if itypes and otypes are not defined
-
-    def make_node(self, x):
-        x = pt.as_tensor_variable(x)
-        assert x.ndim == 2
-        o = TensorType(dtype="bool", shape=[])()
-        return Apply(self, [x], [o])
-
-    # Python implementation:
-    def perform(self, node, inputs, outputs):
-        (x,) = inputs
-        (z,) = outputs
-        try:
-            z[0] = np.array(posdef(x), dtype="bool")
-        except Exception:
-            pm._log.exception("Failed to check if %s positive definite", x)
-            raise
-
-    def infer_shape(self, fgraph, node, shapes):
-        return [[]]
-
-    def grad(self, inp, grads):
-        (x,) = inp
-        return [x.zeros_like(pytensor.config.floatX)]
-
-    def __str__(self):
-        return "MatrixIsPositiveDefinite"
-
-
-matrix_pos_def = PosDefMatrix()
+def matrix_pos_def(X, tol=1e-8):
+    L = pt.linalg.cholesky(X)
+    diag = pt.diagonal(L, axis1=-2, axis2=-1)
+    return pt.all(~pt.isnan(diag)) & pt.all(diag > tol)
 
 
 class WishartRV(RandomVariable):
--- a/pymc/distributions/transforms.py
+++ b/pymc/distributions/transforms.py
@@ -19,7 +19,7 @@ import pytensor.tensor as pt
 
 from numpy.lib.array_utils import normalize_axis_tuple
 from pytensor.graph import Op
-from pytensor.tensor import TensorVariable
+from pytensor.tensor import TensorLike, TensorVariable
 
 from pymc.logprob.transforms import (
     ChainedTransform,
@@ -33,10 +33,15 @@ from pymc.logprob.transforms import (
 
 __all__ = [
     "Chain",
+    "Chain",
+    "CholeskyCorrTransform",
+    "CholeskyCovPacked",
     "CholeskyCovPacked",
     "Interval",
     "Transform",
     "ZeroSumTransform",
+    "ZeroSumTransform",
+    "circular",
     "circular",
     "log",
     "log_exp_m1",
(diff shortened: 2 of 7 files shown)


pymc/distributions/transforms.py
```diff
@@ -138,6 +140,115 @@ def log_jac_det(self, value, *inputs):
...
+            pt.zeros((self.n, self.n))[self.triu_r_idxs, self.triu_c_idxs],
+            value
+        )
+        corr = corr + corr.T + pt.eye(self.n)
+
+        chol = pt.linalg.cholesky(corr)
+
+        # Are the diagonals always guaranteed to be positive?
+        # I don't know, so we'll use abs
+        row_norms = 1/pt.abs(pt.diag(chol))
```
jessegrabowski: Yep, always positive. You don't need abs here

pymc/distributions/transforms.py
```diff
@@ -138,6 +140,115 @@ def log_jac_det(self, value, *inputs):
...
+
+    def _generate_triu_indices(self):
+        row_indices, col_indices = np.triu_indices(self.n, 1)
+        return (
+            pytensor.shared(row_indices),
+            pytensor.shared(col_indices)
+        )
+
+    def _jacobian(self, value, *inputs):
+        return pt.jacobian(
```
jessegrabowski: `pt.jacobian` can be quite expensive, because it requires us to loop over every input and compute the associated symbolic gradients. There's a closed form solution for the log-det jacobian in the TFP code, so you can eliminate this method and implement the closed form log-det jac:

```
    n = ps.shape(y)[-1]
    return -tf.reduce_sum(
        tf.range(2, n + 2, dtype=y.dtype) * tf.math.log(tf.linalg.diag_part(y)),
        axis=-1)
```
`diag_part` would just be `pt.diagonal(y, axis1=-2, axis2=-1)`. That will account for potential batching on `y`. So something like:

```
n = y.shape[-1]
return -(pt.arange(2, n+2, dtype=y.dtype) * pt.log(pt.diagonal(y, axis1=-2, axis2=-1))).sum(axis=-1)
```

pymc/distributions/transforms.py
```diff
@@ -138,6 +140,115 @@ def log_jac_det(self, value, *inputs):
...
+        self.n = n
+        self.m = int(n*(n-1)/2) # number of off-diagonal elements
+        self.tril_r_idxs, self.tril_c_idxs = self._generate_tril_indices()
+        self.triu_r_idxs, self.triu_c_idxs = self._generate_triu_indices()
+
+    def _generate_tril_indices(self):
+        row_indices, col_indices = np.tril_indices(self.n, -1)
+        return (
+            pytensor.shared(row_indices),
+            pytensor.shared(col_indices)
```
jessegrabowski: There's no need to save these as shared variables, you can use the numpy indices directly. Making the numpy indices is pretty cheap, I'm not sure its worth it to cache them

pymc/distributions/transforms.py
```diff
@@ -138,6 +140,115 @@ def log_jac_det(self, value, *inputs):
...
+        # Are the diagonals always guaranteed to be positive?
+        # I don't know, so we'll use abs
+        row_norms = 1/pt.abs(pt.diag(chol))
+
+        # Multiply by the row norms to undo the normalization
+        unconstrained = chol*row_norms[:, pt.newaxis]
+
+        return unconstrained[self.tril_r_idxs, self.tril_c_idxs]
+
+    def backward(self, value, *inputs, foo=False):
```
jessegrabowski: You need to check that these functions match the expected outputs from TFP. I used the test case from the tfp docs and got the wrong values -- `array([0.89442719, 0.81649658, 0.91287093])` vs the reference solution

```
array([[1.        , 0.        , 0.        ],
       [0.70710678, 0.70710678, 0.        ],
       [0.66666667, 0.66666667, 0.33333333]])
```

You did some extra research so I might be missing something?

jessegrabowski: Something like this matches tfp:

```
        """
        Convert unconstrained real numbers to the off-diagonal elements of the
        cholesky decomposition of a correlation matrix.
        """
        def unpack_upper_tril_with_eye_diag(x, core_shape):
            """1D allocation case"""
            return pt.set_subtensor(pt.eye(core_shape)[*np.tril_indices(core_shape, k=-1)], x[::-1])
        
        value = pt.as_tensor_variable(value)
        core_shape = value.type.shape[-1]
        
        # Vectorize the 1D case to handle potential batch dimensions
        out = pt.vectorize(partial(unpack_upper_tril_with_eye_diag, core_shape=core_shape), '(n)->(n,n)')(value)
        
        # Vector L2 norm without .real call to speed things up a bit
        norm = pt.sqrt(pt.sum((out ** 2), axis=-1, keepdims=True))
        return out / norm
```

pymc/distributions/transforms.py
```diff
@@ -138,6 +140,115 @@ def log_jac_det(self, value, *inputs):
...
+
+        Parameters
+        ----------
+        n: int
+            Size of correlation matrix
+        """
+        self.n = n
+        self.m = int(n*(n-1)/2) # number of off-diagonal elements
+        self.tril_r_idxs, self.tril_c_idxs = self._generate_tril_indices()
+        self.triu_r_idxs, self.triu_c_idxs = self._generate_triu_indices()
```
jessegrabowski: See below, not sure we need to cache these. `__init__` is probably unnecessary

pymc/distributions/transforms.py
```diff
@@ -138,6 +140,115 @@ def log_jac_det(self, value, *inputs):
...
+        """
+        Compute log of the determinant of the jacobian.
+
+        There are no clever tricks here - we literally compute the jacobian
+        then compute its determinant then take log.
+        """
+        jac = self._jacobian(value)
+        return pt.log(pt.linalg.det(jac))
+
+    def forward(self, value, *inputs):
```
jessegrabowski: See below. I'm pretty sure this needs to go from matrix to vector (to match the tfp case) @junpenglao might know for sure.

jessegrabowski (changes requested):

pymc/distributions/transforms.py
johncant: Thanks for that code above, there's some great stuff I can reuse.

Need to address this comment first, since actually working in the first place is really fundamental. Here's a notebook that demonstrates that this implementation does replicate the original reference implementation from TFP: https://colab.research.google.com/drive/1BBNNfBUNJPGT_7MxVboTqvRegJ-TUamc?usp=sharing

Here's why it didn't work for you:
  - PyMC implementation needs to output the upper triangular elements of the correlation matrix, whereas the TFP implementation outputs a Cholesky factor.
  - Differences in indexing off-diagonal elements. TFP actually fills in off-diagonal elements in a clockwise spiral, whereas np.triu_indices is row major. I notice you reverse the elements in the code above which is correct for 3x3 but not for larger matrices.

jessegrabowski: I got access denied to the colab :)

I'm not surprised my code doesn't work, but I'm glad you know why. In the general case, tensorflow concatenates a reflected copy of the array to itself then reshapes and masks out the lower/upper triangle -- see [here ](https://www.tensorflow.org/probability/api_docs/python/tfp/math/fill_triangular)if you haven't already. There's no reason why we couldn't just do that.

I'm not sure that we need to copy their output 1:1 -- after all, the important thing is that we can go from unconstrained samples to a valid cholesky decomposed correlation matrix. Is the order we put the numbers into the matrix relevant? I'm not sure, but my instinct is no. On the other hand, if we copy 1:1 we can be sure it's right.

jessegrabowski: > PyMC implementation needs to output the upper triangular elements of the correlation matrix, where the TFP implementation outputs a Cholesky factor.

Are you sure? I thought the upper triangular elements are the cholesky factorized correlation matrix. If you're right though we just need to add a matmul to the end right?

johncant: Apologies - Link is now open with Viewer permission and I've made you an Editor.

I don't think the order we insert the off-diagonal elements into an array is very important, but it is needed in order to compare results between this implementation and the one in TFP. I would suggest sticking with `np.triu_indices` here.

johncant: > Are you sure? I thought the upper triangular elements are the cholesky factorized correlation matrix. If you're right though we just need to add a matmul to the end right?

Yes, you can see this by looking at the implementation of LKJCorr. I originally thought the same thing, implemented the transform accordingly, then was surprised that non-posdef matrices were generated. :facepalm:

pymc/distributions/multivariate.py
```diff
@@ -1579,7 +1579,9 @@ def logp(value, n, eta):
 
 @_default_transform.register(_LKJCorr)
 def lkjcorr_default_transform(op, rv):
-    return MultivariateIntervalTransform(-1.0, 1.0)
```
ricardoV94, line 1642: Can you delete this transform class as well? It was a (wrong) patch to the problem you're solving

pymc/distributions/transforms.py
```diff
@@ -138,6 +140,115 @@ def log_jac_det(self, value, *inputs):
...
+        n: int
+            Size of correlation matrix
+        """
+        self.n = n
+        self.m = int(n*(n-1)/2) # number of off-diagonal elements
+        self.tril_r_idxs, self.tril_c_idxs = self._generate_tril_indices()
+        self.triu_r_idxs, self.triu_c_idxs = self._generate_triu_indices()
+
+    def _generate_tril_indices(self):
+        row_indices, col_indices = np.tril_indices(self.n, -1)
```
ricardoV94: Not sure if it matters but there is a `pt.tril_indices` and `pt.triu_indices` so no need to eval `n`. If it's already restricted to be constant elsewhere (like the logp), then it's fine either way

pymc/distributions/transforms.py
jessegrabowski: Great! Sounds like you have a good handle on things. I think what would be a really important next step would be to add a test that your implementation correctly makes a round trip from $\mathbb R \to \Omega \to \mathbb R$, where $\Omega$ is the set of correlation matrices.

pymc/distributions/transforms.py
jessegrabowski: I think it's good practice to use the `pt` version, even if `n` is fixed

pymc/distributions/transforms.py
johncant: Awesome - I can do that

pymc/distributions/multivariate.py
johncant: Can do. Just to confirm, you don't consider MultivariateIntervalTransform to be part of pymc's public API?

pymc/distributions/transforms.py
johncant: I originally tried to use the `pt` version, but one of the function calls required constant values. However, I've made so many changes, that might no longer be the case. I'll try the `pt` version again and see if I can get it to work.

pymc/distributions/multivariate.py
ricardoV94: Nope, can be removed without worries

johncant: Ok - great

pymc/distributions/transforms.py
junpenglao: +1 it is better for the unbounded being a vector.

johncant: Sorry, I am a bit confused and don't understand what you mean.

Specifically, do you mean this function needs to work along the last axis for arrays of arbitrary number of dimensions, and that the current iteration assumes that `value` will only have dimension 1?

pymc/distributions/transforms.py
johncant: Can you point me to some info on how that's derived? Going to need to modify it to work with the correlation matrix.

johncant: Never mind, found in the comments in TFP _inverse_log_det_jacobian

johncant:
Hi, It's unlikely I'm going to have any time to work on this for the next 6 months. The hardest part is coming up with a closed form solution for log_det_jac, which I don't think I'm very close to doing.

twiecki:
Thanks for the update @johncant and for pushing this as far as you did.

tests/distributions/test_transform.py
```diff
@@ -673,3 +674,142 @@ def test_deprecated_ndim_supp_transforms():
...
+    x_tensor = pt.as_tensor_variable(x)
+
+    # Perform forward transform to obtain Cholesky factors
+    y = transform.forward(x_tensor).eval()
+
+    # Compute the log determinant using the transform's method
+    computed_log_jac_det = transform.log_jac_det(y).eval()
+
+    # Expected log determinant: 0 (since row norms are 1)
+    expected_log_jac_det = 0.0
```
ricardoV94: Weak test. Tell it to compare with pytensor jacobian machinery with a non-trivial input. And to reuse test code that already exists to do that

ricardoV94 (changes requested):

jessegrabowski:
This is updated to a fully working implementation of the transformer, with non-trivial tests.

It is currently blocked by #7910, because it uses `pt.flip` with negative indices, which are bugged in the current main branch. Tests pass locally on the current pytensor main.

There's some jank -- I'm not sure we need to pass `n` to the constructor. We can easily infer `n` from all the inputs. I guess the benefit is to be able to check that shapes haven't changed somewhere since the transformer was initialized? But these are so internal to PyMC that users shouldn't be able to make that happen anyway (and if one can, he's a super user so who cares).

If we got rid of the `n` property, it would also do away with the `unit_diag` arguments on the triangular_pack helpers.

On that note, those two triangular pack helpers could themselves be a separate transformation, since they define a Bijection with forward and backward methods. That might be conceptually better, but on the other hand I don't think these would be used anywhere else, so it also makes sense to keep them as part of this class.

ricardoV94:
BTW once this is merged we should still explore the new transform by that Stan dev, it should have better sampling properties IIRC

jessegrabowski:
Do you have a link to the actual implementation? I was rooting around in the stan/stan-math repo and couldn't find it

johncant:
Wow - congratulations @jessegrabowski !

jessegrabowski:
> Wow - congratulations @jessegrabowski !

You had it 90% of the way there, we were just missing this weird spiral triangular construction that tfp was doing. I have no idea why they do it this way though, I just blindly copied xD

pymc/distributions/multivariate.py
```diff
@@ -1517,74 +1515,57 @@ def make_node(self, rng, size, n, eta):
 
     @classmethod
     def rv_op(cls, n: int, eta, *, rng=None, size=None):
-        # We flatten the size to make operations easier, and then rebuild it
+        # HACK: normalize_size_param doesn't handle size=() properly
```
ricardoV94: It doesn't? Size=(), has a specific mean. It means no batch dimensions

jessegrabowski: I forget why I wrote this but I believe my past self.

ricardoV94: I use / wrote the helper so I'm skeptical

jessegrabowski:
I switch the LKJCorr implementation to use `scan` to generate samples. This means:

1. Both `eta` and `n` can be symbolic
2. Larger matrices sample much faster.
3. But small matrices are somewhat slower

We could potentially try to check `n` and unroll the loop if it's sufficiently small. It also made me think an `scan_to_unrolled_loop` rewrite/tool might be interesting/useful in general.

Here is the benchmark I ran:

```
@pytest.mark.parametrize("n", [2, 10, 100])
def test_lkjcorr_sampling_benchmark(n, benchmark):
    d = pm.LKJCorr.dist(n=n, eta=2.0)
    sample_fn = pm.compile(inputs=[], outputs=[d], mode='FAST_RUN')
    sample_fn()

    benchmark(sample_fn)
```


Before (unrolled loop):
```
----------------------------------------------------------------------------------------------------- benchmark: 3 tests -----------------------------------------------------------------------------------------------------                                                                                                                                              
Name (time in us)                             Min                    Max                  Mean                StdDev                Median                 IQR            Outliers           OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------                                                                                                                                              
test_lkjcorr_sampling_benchmark[2]         3.1670 (1.0)          38.4999 (1.0)          3.6752 (1.0)          0.5252 (1.0)          3.7500 (1.0)        0.3759 (1.0)       435;361  272,093.1823 (1.0)       17765           1                                                                                                                                              
test_lkjcorr_sampling_benchmark[10]       59.1250 (18.67)       250.4170 (6.50)        68.1735 (18.55)        4.7520 (9.05)        69.0001 (18.40)      4.0832 (10.86)    2420;568   14,668.4544 (0.05)      12006           1
test_lkjcorr_sampling_benchmark[100]     831.8330 (262.66)   21,313.4581 (553.60)   1,588.8558 (432.32)   1,978.1464 (>1000.0)  1,050.2914 (280.08)   179.7080 (478.07)     63;117      629.3837 (0.00)        862           1                                                                                                                                              
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------                                                                                                                                              
```


After (scan):
```
----------------------------------------------------------------------------------------------------- benchmark: 3 tests ----------------------------------------------------------------------------------------------------                                                                                                                                               
Name (time in us)                               Min                   Max                  Mean              StdDev                Median                 IQR            Outliers           OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------                                                                                                                                               
test_lkjcorr_sampling_benchmark[2]           7.9999 (1.0)         77.4580 (1.0)          9.2234 (1.0)        2.3224 (1.0)          8.5421 (1.0)        0.9161 (1.0)       691;911  108,419.6575 (1.0)        8120           1                                                                                                                                               
test_lkjcorr_sampling_benchmark[10]        229.1250 (28.64)      581.1669 (7.50)       253.4111 (27.47)     21.7310 (9.36)       249.2500 (29.18)     19.0000 (20.74)     283;121    3,946.1577 (0.04)       3424           1
test_lkjcorr_sampling_benchmark[100]     3,908.5420 (488.57)   6,156.4590 (79.48)    4,609.0084 (499.71)   446.9420 (192.45)   4,542.3959 (531.77)   660.9370 (721.49)       70;3      216.9664 (0.00)        208           1                                                                                                                                               
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------                                                                                                                                               

```

jessegrabowski:
We can support estimation of `eta` now. We should be able to add broadcasting on these parameters. but I'll leave that for a future PR.

ricardoV94:
Those timings look good enough I wouldn't worry. It will hopefully fall out from our goal of speeding up Scan. Can you try with Numba mode out of curiosity?

pymc/distributions/multivariate.py
```diff
@@ -1513,78 +1517,83 @@ def make_node(self, rng, size, n, eta):
...
+            fn=step,
+            outputs_info=[beta0, P0],
+            sequences=[pt.arange(2, n)],
+            non_sequences=[scan_rng],
+            strict=True,
+        )
 
-    def log_jac_det(self, *args):
-        return super().log_jac_det(*args).sum(-1)
+        P = pytensor.ifelse(n < 3, P0, P_seq[-1])
```
jessegrabowski: Testing numba actually revealed an error in the RV here, curious if you have a better suggestion of how to proceed. Obviously the easy way out is to drop back to disallowing symbolic `n`, but I'm trying to avoid that.

ricardoV94: What is the error?

jessegrabowski:
Numba timings. With scan:


```
----------------------------------------------------------------------------------------------------- benchmark: 3 tests -----------------------------------------------------------------------------------------------------                                                                                                                                              
Name (time in us)                               Min                    Max                  Mean              StdDev                Median                 IQR            Outliers           OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------                                                                                                                                              
test_lkjcorr_sampling_benchmark[2]           8.5830 (1.0)          93.3340 (1.0)          9.6742 (1.0)        1.1092 (1.0)          9.8330 (1.0)        0.8750 (1.0)       215;186  103,367.4023 (1.0)       11911           1                                                                                                                                              
test_lkjcorr_sampling_benchmark[10]        199.0420 (23.19)    40,315.1250 (431.94)     272.4426 (28.16)    823.9673 (742.82)     225.4579 (22.93)     24.5001 (28.00)      23;326    3,670.4975 (0.04)       4113           1
test_lkjcorr_sampling_benchmark[100]     3,578.3750 (416.92)    5,509.3330 (59.03)    4,043.9366 (418.01)   256.9228 (231.62)   4,027.7915 (409.62)   316.2490 (361.44)       71;4      247.2838 (0.00)        238           1                                                                                                                                              
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------                                                                                                                                              

```

Unrolled loop (old implementation):

```
-------------------------------------------------------------------------------------------------- benchmark: 3 tests --------------------------------------------------------------------------------------------------                                                                                                                                                    
Name (time in us)                             Min                   Max                Mean              StdDev              Median                 IQR             Outliers  OPS (Kops/s)            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------                                                                                                                                                    
test_lkjcorr_sampling_benchmark[2]         4.7911 (1.0)         49.6670 (1.0)        5.5282 (1.0)        0.7295 (1.0)        5.5420 (1.0)        0.1250 (1.0)       174;2127      180.8892 (1.0)       10846           1                                                                                                                                                    
test_lkjcorr_sampling_benchmark[10]        7.5829 (1.58)        54.9171 (1.11)       8.7685 (1.59)       0.8032 (1.10)       8.8330 (1.59)       0.2501 (2.00)     4005;7133      114.0448 (0.63)      34385           1
test_lkjcorr_sampling_benchmark[100]     167.4580 (34.95)    2,400.6250 (48.33)    317.5335 (57.44)    235.1336 (322.34)   250.0410 (45.12)    105.5831 (844.46)     162;312        3.1493 (0.02)       2211           1                                                                                                                                                    
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------                                                                                                                                                    
```

Pretty bad!

jessegrabowski:
JAX also doesn't work with the new scan implementation, I guess because `n` is symbolic so `pt.arange(2, n)`(the length of the scan) is unknown at compile time

pymc/distributions/multivariate.py
jessegrabowski: The scan is defined over `arange(2, n)`, so if `n < 3` the output of the scan is 0d, and `P = P_seq[-1]` errors out.

I changed it to use concat instead.

ricardoV94:
length shouldn't matter if only last state is used? can you write index as a recurring out?

I don't recall where arange shows up

ricardoV94:
timings are pretty nice, not worried about the small case

pymc/distributions/multivariate.py
jessegrabowski: rolled this back

jessegrabowski:
> timings are pretty nice, not worried about the small case

No, it's bad. I updated the original timings with a compiled function instead of `pm.draw` and the scan is significantly worse across the board. I can't reproduce the original timings, so I might have been picking up first-run compile times or something.

ricardoV94:
In the compiled version did you set updates so the rng is not copied?

jessegrabowski:
I used `pm.compile`, doesn't that handle it for you?

ricardoV94:
Ah so you were seeing longer compile times that made it look like scan was doing better? Unrolled 100 things or more is gonna be pretty ugly though

ricardoV94:
Can you show the final numba graph?

pymc/distributions/multivariate.py
```diff
@@ -918,18 +923,15 @@ def posdef(AA):
 class PosDefMatrix(Op):
```
ricardoV94: Get rid of this Op? Requires python mode and it's only calling cholesky to see if it raises

pymc/distributions/multivariate.py
```diff
@@ -1513,78 +1517,83 @@ def make_node(self, rng, size, n, eta):
...
+            fn=step,
+            outputs_info=[beta0, P0],
+            sequences=[pt.arange(2, n)],
+            non_sequences=[scan_rng],
+            strict=True,
+        )
 
-    def log_jac_det(self, *args):
-        return super().log_jac_det(*args).sum(-1)
+        P = pt.concatenate([P0[None], P_seq], axis=0)[-1]
```
ricardoV94: Whay do you concatenate and then take the last value? Sometimes P_seq is empty?

pymc/testing.py
```diff
@@ -735,7 +735,13 @@ def continuous_random_tester(
         while p <= alpha and f > 0:
             s0 = pymc_rand()
             s1 = floatX(ref_rand(size=size, **point))
-            _, p = st.ks_2samp(np.atleast_1d(s0).flatten(), np.atleast_1d(s1).flatten())
+
+            # If a distribution has non-stochastic elements in the output (e.g. LKJCorr putting 1's on the diagonal),
+            # it will mess up the KS test. So we filter those out here.
+            stacked_samples = np.c_[np.atleast_1d(s0).flatten(), np.atleast_1d(s1).flatten()]
+            samples = stacked_samples[~np.isclose(stacked_samples[..., 0], stacked_samples[..., 1])]
```
ricardoV94, line 953: I don't like this change. Use some other routine in the test instead?

pymc/distributions/multivariate.py
jessegrabowski: Yes, when `n=2`. I had an `ifelse` for that but it still breaks in non-vm backends, so this was my other idea. Final fallback is to go back to non-symbolic `n`, but that's a bit of a bummer

pymc/testing.py
jessegrabowski: Sure. I think it's a pretty questionable test in the first place.

tests/distributions/test_multivariate.py
```diff
@@ -2153,59 +2160,58 @@ class TestLKJCorr(BaseTestDistributionRandom):
...
-            ),
-        ),
-    ],
-)
+            return corr
+
+        # n can be symbolic, but only n=2 is tested for two reasons:
+        #   1) if n > 2, the ref_rand function is wrong. We don't have a good reference for sampling LKJ
+        #   2) Although n can be symbolic, the inner scan graph needs to be rebuilt after it changes. The approach
+        #      implemented in this tester does not rebuild the inner function graph, causing an error.
```
ricardoV94: Comment doesn't make sense to me. Why / how could the Scan be rebuilt when n changes if n can be symbolic?

pymc/distributions/multivariate.py
jessegrabowski: Sure. What we really need anyway is a complete typing system for matrices :)

pymc/distributions/multivariate.py
```diff
@@ -1647,7 +1654,9 @@ def logp(value, n, eta):
 
 @_default_transform.register(_LKJCorr)
 def lkjcorr_default_transform(op, rv):
-    return MultivariateIntervalTransform(-1.0, 1.0)
+    rng, scan_rng, shape, n, eta, *_ = rv.owner.inputs
+    n = pt.get_scalar_constant_value(n)  # Safely extract scalar value without eval
```
ricardoV94: if n can be arbitrarily symbolic there is no way to safely extract it :)

pymc/distributions/multivariate.py
```diff
@@ -1593,12 +1602,26 @@ class _LKJCorr(BoundedContinuous):
...
+        # In general, RVs are expected to take an "rng" argument. We allow it here to prevent API break, but
+        # ignore it. This can be changed in the future if we relax the requirement that rng be a shared variable
+        # in a scan.
+        rng = kwargs.pop("rng", None)
+        if rng is not None:
+            warnings.warn(
+                "You passed a random generator to LKJCorr via the `rng` keyword argument, but it is not "
+                "used. To seed LKJCorr, pass two random generators via the `outer_rng` and `scan_rng` "
+                "keyword arguments.",
+            )
```
ricardoV94: This can be removed once we pin to the next PyTensor release

pymc/distributions/multivariate.py
ricardoV94: Scan over `pt.arange(2, pt.maximum(n, 3))` and then `P = pt.where(n > 2, P_seq[-1], P0)`?

jessegrabowski: That will still compute the `P_seq[-1]` branch in all cases though?

jessegrabowski: oh you want to do an extra scan step for safety. Seems about the same, both allocate an extra `n,n` in memory, but yours also does the extra computation

tests/distributions/test_multivariate.py
jessegrabowski: i don't remember, it was something weird about how these tests work.

pymc/distributions/multivariate.py
jessegrabowski: It's not required to be non-symbolic anyway

tests/distributions/test_multivariate.py
jessegrabowski: I recommend you just try adding a 2nd value to the domain of `n` and see. You get an error like this.

These tests are quite over-engineered in my opinion. It's extremely hard to follow what is going on.

<details>

<summary> Traceback! </summary>

```
pymc/testing.py:869 (TestLKJCorr.test_distribution)
>   ???

pytensor/scan/scan_perform.pyx:418: 
../pytensor/pytensor/graph/op.py:545: in rval
    r = p(n, [x[0] for x in i], o)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
../pytensor/pytensor/tensor/random/op.py:428: in perform
    self.rng_fn(rng, *args, None if size is None else tuple(size)),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

self = BetaRV(name=beta,signature=(),()->(),dtype=float64,inplace=False)
rng = None, args = (array([1.]), array([2.]), (np.int64(1000),)), kwargs = {}

    def rng_fn(self, rng, *args, **kwargs) -> int | float | np.ndarray:
        """Sample a numeric random variate."""
>       return getattr(rng, self.name)(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: 'NoneType' object has no attribute 'beta'

../pytensor/pytensor/tensor/random/op.py:194: AttributeError

During handling of the above exception, another exception occurred:

node = Scan{scan_fn, while_loop=False, inplace=all}(Composite{...}.1, Mul.0, ARange{dtype='int64'}.0, SetSubtensor{:stop}.0, SetSubtensor{:stop}.0, RNG(<None>))
inputs = [array(2), array([[1. ],
       [1.5]]), array([2, 3]), array([ 2.00000000e+00, -8.44352946e-69]), array([[[[ 1.      ...  ,  0.        ],
         [ 0.        ,  0.        ,  0.        ,  0.        ]]]],
      shape=(3, 1000, 4, 4)), None]
outputs = [[array([ 2.00000000e+00, -8.44352946e-69])], [array([[[[ 1.        , -0.24330985,  0.        ,  0.        ],
        ...  0.        ],
         [ 0.        ,  0.        ,  0.        ,  0.        ]]]],
      shape=(3, 1000, 4, 4))], [None]]

    def p(node, inputs, outputs):
        t0_call = time.perf_counter()
    
        try:
>           t_fn, n_steps = scan_perform_ext.perform(
                self.info.n_shared_outs,
                self.info.n_mit_mot_outs,
                self.info.n_seqs,
                self.info.n_mit_mot,
                self.info.n_mit_sot,
                self.info.n_sit_sot,
                self.info.n_nit_sot,
                self.info.as_while,
                cython_mintaps,
                cython_pos,
                cython_store_steps,
                self.info.mit_mot_in_slices
                + self.info.mit_sot_in_slices
                + self.info.sit_sot_in_slices,
                tap_array_len,
                cython_vector_seqs,
                cython_vector_outs,
                self.info.mit_mot_out_slices,
                cython_mitmots_preallocated,
                mit_mot_out_to_tap_idx,
                cython_outs_is_tensor,
                inner_input_storage,
                inner_output_storage,
                cython_destroy_map,
                inputs,
                outputs,
                outer_output_dtypes,
                outer_output_ndims,
                self.fn.vm,
            )

../pytensor/pytensor/scan/op.py:1587: 

>   ???
E   pytensor.scan.utils.InnerFunctionError: (AttributeError("'NoneType' object has no attribute 'beta'"), <traceback object at 0x3314aee80>)

pytensor/scan/scan_perform.pyx:420: InnerFunctionError

During handling of the above exception, another exception occurred:

self = <pytensor.compile.function.types.Function object at 0x332383410>
output_subset = None, args = (), kwargs = {}, trust_input = False
vm = <pytensor.link.c.cvm.CVM object at 0x331876190>, profile = None
arg_container = <array(1.)>

    def __call__(self, *args, output_subset=None, **kwargs):
        """
        Evaluates value of a function on given arguments.
    
        Parameters
        ----------
        args : list
            List of inputs to the function. All inputs are required, even when
            some of them are not necessary to calculate requested subset of
            outputs.
    
        kwargs : dict
            The function inputs can be passed as keyword argument. For this, use
            the name of the input or the input instance as the key.
    
            Keyword argument ``output_subset`` is a list of either indices of the
            function's outputs or the keys belonging to the `output_keys` dict
            and represent outputs that are requested to be calculated. Regardless
            of the presence of ``output_subset``, the updates are always calculated
            and processed. To disable the updates, you should use the ``copy``
            method with ``delete_updates=True``.
    
        Returns
        -------
        list
            List of outputs on indices/keys from ``output_subset`` or all of them,
            if ``output_subset`` is not passed.
        """
        trust_input = self.trust_input
        input_storage = self.input_storage
        vm = self.vm
        profile = self.profile
    
        if profile:
            t0 = time.perf_counter()
    
        if output_subset is not None:
            warnings.warn("output_subset is deprecated.", FutureWarning)
            if self.output_keys is not None:
                output_subset = [self.output_keys.index(key) for key in output_subset]
    
        # Reinitialize each container's 'provided' counter
        if trust_input:
            # zip strict not specified because we are in a hot loop
            for arg_container, arg in zip(input_storage, args):
                arg_container.storage[0] = arg
        else:
            for arg_container in input_storage:
                arg_container.provided = 0
    
            if len(args) + len(kwargs) > len(input_storage):
                raise TypeError("Too many parameter passed to pytensor function")
    
            # Set positional arguments
            # zip strict not specified because we are in a hot loop
            for arg_container, arg in zip(input_storage, args):
                # See discussion about None as input
                # https://groups.google.com/group/theano-dev/browse_thread/thread/920a5e904e8a8525/4f1b311a28fc27e5
                if arg is None:
                    arg_container.storage[0] = arg
                else:
                    try:
                        arg_container.storage[0] = arg_container.type.filter(
                            arg,
                            strict=arg_container.strict,
                            allow_downcast=arg_container.allow_downcast,
                        )
    
                    except Exception as e:
                        i = input_storage.index(arg_container)
                        function_name = "pytensor function"
                        argument_name = "argument"
                        if self.name:
                            function_name += ' with name "' + self.name + '"'
                        if hasattr(arg, "name") and arg.name:
                            argument_name += ' with name "' + arg.name + '"'
                        where = get_variable_trace_string(self.maker.inputs[i].variable)
                        if len(e.args) == 1:
                            e.args = (
                                "Bad input "
                                + argument_name
                                + " to "
                                + function_name
                                + f" at index {int(i)} (0-based). {where}"
                                + e.args[0],
                            )
                        else:
                            e.args = (
                                "Bad input "
                                + argument_name
                                + " to "
                                + function_name
                                + f" at index {int(i)} (0-based). {where}"
                            ) + e.args
                        self._restore_defaults()
                        raise
                arg_container.provided += 1
    
        # Set keyword arguments
        if kwargs:  # for speed, skip the items for empty kwargs
            for k, arg in kwargs.items():
                self[k] = arg
    
        if not trust_input:
            # Collect aliased inputs among the storage space
            for potential_group in self._potential_aliased_input_groups:
                args_share_memory: list[list[int]] = []
                for i in potential_group:
                    i_type = self.maker.inputs[i].variable.type
                    i_val = input_storage[i].storage[0]
    
                    # Check if value is aliased with any of the values in one of the groups
                    for j_group in args_share_memory:
                        if any(
                            i_type.may_share_memory(input_storage[j].storage[0], i_val)
                            for j in j_group
                        ):
                            j_group.append(i)
                            break
                    else:  # no break
                        # Create a new group
                        args_share_memory.append([i])
    
                # Check for groups of more than one argument that share memory
                for group in args_share_memory:
                    if len(group) > 1:
                        # copy all but the first
                        for i in group[1:]:
                            input_storage[i].storage[0] = copy.copy(
                                input_storage[i].storage[0]
                            )
    
            # Check if inputs are missing, or if inputs were set more than once, or
            # if we tried to provide inputs that are supposed to be implicit.
            for arg_container in input_storage:
                if arg_container.required and not arg_container.provided:
                    self._restore_defaults()
                    raise TypeError(
                        f"Missing required input: {getattr(self.inv_finder[arg_container], 'variable', self.inv_finder[arg_container])}"
                    )
                if arg_container.provided > 1:
                    self._restore_defaults()
                    raise TypeError(
                        f"Multiple values for input: {getattr(self.inv_finder[arg_container], 'variable', self.inv_finder[arg_container])}"
                    )
                if arg_container.implicit and arg_container.provided > 0:
                    self._restore_defaults()
                    raise TypeError(
                        f"Tried to provide value for implicit input: {getattr(self.inv_finder[arg_container], 'variable', self.inv_finder[arg_container])}"
                    )
    
        # Do the actual work
        if profile:
            t0_fn = time.perf_counter()
        try:
>           outputs = vm() if output_subset is None else vm(output_subset=output_subset)
                      ^^^^

../pytensor/pytensor/compile/function/types.py:1038: 
../pytensor/pytensor/scan/op.py:1659: in rval
    r = p(n, [x[0] for x in i], o)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
../pytensor/pytensor/scan/op.py:1626: in p
    link_utils.raise_with_op(
../pytensor/pytensor/link/utils.py:526: in raise_with_op
pytensor/scan/scan_perform.pyx:418: in pytensor.scan.scan_perform.perform
    ???
../pytensor/pytensor/graph/op.py:545: in rval
    r = p(n, [x[0] for x in i], o)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
../pytensor/pytensor/tensor/random/op.py:428: in perform
    self.rng_fn(rng, *args, None if size is None else tuple(size)),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

</details>

pymc/distributions/multivariate.py
ricardoV94: Otherwise Scan actually produces the full vector with the initial state in it, just slices away the initial state for the user. You can retrieve it with `P_seq.owner.inputs[0]`

jessegrabowski: I already have it, it's `P0`. The issue isn't how to get the initial state, it's how to give it back when (at runtime) n=2.

ricardoV94: I'm telling you you already have this concatenated vector because it's the true output of Scan Op. You're just being returned `seq[1:]` the the scan helper. If you navigate the graph one step you get `seq` which for 0 steps will just have P0 in it

jessegrabowski: Oh! Mind blown.

jessegrabowski: `P = P_seq.owner.inputs[0][-1]` still fails on the `n=2` case. I'm sticking with the concatenate method unless you spoon feed me a better way

ricardoV94:
I have some asks, test-wise

- [x] Test the transform directly: forward-backward is identity and log-jac-det is correct, like we do for other transforms
- [x] Test that the forward of a random vector is indeed the cholesky factor of a valid correlation matrix (we test that indirectly via the new finite logp check, but I want to test the transform independently, because the logp could be wrong / or changed in a way that doesn't check that, and we would lose coverage)

And follow-up, we should open issues for:
- [ ] Fix the double scan in numba mode
- [ ] Compare against the new stan impl: https://discourse.mc-stan.org/t/updated-cholesky-corr-parameterization-testing/38827

jessegrabowski:
right now there's some confusion in the code about what `LKJCorr` should return:

- the transform works with cholesky decomposed matrices, so values will be a flat vector and the forward method return `L` such that `C = L @ L.T`
- But the forward draws return`C`, the correlation matrix itself, directly.

Tests are failing as a result. Working with `L` is easier internally, but users will expect `C`, so I guess there needs to be a deterministic helper and a flag for which to give back?

ricardoV94:
Aren't you mixing `LKJCorr` and `_LKJCorr`? `LKJCorr` is the user facing one that returns the correlation matrix (optionally?), under the hood it works with `_LKJCorr`, which works on whatever space it wants. That's the one that needs an unconstraining transform.

tests/distributions/test_transform.py
```diff
@@ -666,3 +666,90 @@ def log_jac_det(self, value, *inputs):
             match="are not allowed to broadcast together. There is a bug in the implementation of either one",
         ):
             m.logp(jacobian=jacobian_val)
+
+
+class TestLJKCholeskyCorrTransform:
```
jessegrabowski, line 663: >  Test the transform directly: forward-backward is identity and log-jac-det is correct, like we do for other transforms

I think we already have that here, let me know if you want more.

ricardoV94: yeah perfect

jessegrabowski:
> Aren't you mixing LKJCorr and _LKJCorr? LKJCorr is the user facing one that returns the correlation matrix (optionally?), under the hood it works with _LKJCorr, which works on whatever space it wants. That's the one that needs an unconstraining transform.

I'm not sure we need _LKJCorr at all anymore

ricardoV94:
> > Aren't you mixing LKJCorr and _LKJCorr? LKJCorr is the user facing one that returns the correlation matrix (optionally?), under the hood it works with _LKJCorr, which works on whatever space it wants. That's the one that needs an unconstraining transform.
> 
> I'm not sure we need _LKJCorr at all anymore

What changed? I didn't see us changing anything in the RV side yet

jessegrabowski:
I added the other requested test.

I am hitting a problem now where the RNG returned from my scan in `_random_corr_matrix` is None in the `TestLKJCorr:check_draws_match_expected` test. I think it is something to do with how the scan Op is being rebuilt?

jessegrabowski:
> What changed? I didn't see us changing anything in the RV side yet

`_LKJCorr` previously returned the packed lower-triangle of the cholesky decomposed matrix. The logp function also acted on vectors, and we sampled vectors.

Now everything works on matrices natively, so there's no need for helpers to do all the packing/unpacking. The transform returns matrix, the logp evaluates matrices, and the rv_op returns random matrices.

ricardoV94:
So is the underlying distribution no longer working on the non-zero-triangular entries now?

ricardoV94:
I don't know if it was worth it, but I understand that was done for memory efficiency. To avoid storing all those useless entries. 

Also, computationally could the logp more expensive now?

And what are we working on, the correlation matrix or still the cholesky factor of it?

jessegrabowski:
We're working on the cholesky decomposition of the correlation matrix. This should be computationally cheaper across the board, but yes it will require more memory for random draws. But users can't do anything with the low memory draws anyway. If you actually want to compute with a correlation matrix, you are going to have to materialize it at some point.

The logp in this PR should be cheaper than what's on main. Here's the current code:

```py
        shape = n * (n - 1) // 2
        tri_index = np.zeros((n, n), dtype="int32")
        tri_index[np.triu_indices(n, k=1)] = np.arange(shape)
        tri_index[np.triu_indices(n, k=1)[::-1]] = np.arange(shape)

        value = pt.take(value, tri_index)
        value = pt.fill_diagonal(value, 1)

        # TODO: _lkj_normalizing_constant currently requires `eta` and `n` to be constants
        try:
            eta = float(get_underlying_scalar_constant_value(eta))
        except NotScalarConstantError:
            raise NotImplementedError("logp only implemented for constant `eta`")
        result = _lkj_normalizing_constant(eta, n)
        result += (eta - 1.0) * pt.log(det(value))
        return check_parameters(
            result,
            value >= -1,
            value <= 1,
            matrix_pos_def(value),
            eta > 0,
        )
```

And here's the new logp:

```py
        result = _lkj_normalizing_constant(eta, n) + (eta - 1.0) * 2 * pt.diagonal(
            value, axis1=-2, axis2=-1
        ).log().sum(axis=-1)

        row_norms = pt.sum(value**2, axis=-1)

        return check_parameters(
            result,
            eta > 0,
            pt.isclose(row_norms, 1.0),
            msg="Invalid values passed to LKJCorr logp: value is not a valid correlationm matrix or eta <= 0.",
        )
```

You can see previously they did all this reshaping business to get the packed vector into a full correlation matrix then computed the logp. 

Now we just exploit the cholesky structure to get cheap logdet. Our rewrites definitely wouldn't have been able to infer any structure from what were previously doing.

jessegrabowski:
> But users can't do anything with the low memory draws anyway

I guess the point is that one can avoid storing them in the trace? That's an advantage I didn't think of. Personally I like storing the whole thing so I can have square dimensions on it and easily look at results. Working with the packed lower triangle requires multi-indices, which arviz supports only barely.

jessegrabowski:
>  Fix the double scan in numba mode

Just to keep things as mysterious as possible, I don't see the double scan anymore in the current state of the PR.

ricardoV94:
You were doing something with the rng thing wrongly? Would still be nice to go back to the broken state for debugging, because the issue was real. Obviously not a blocker for this PR

jessegrabowski:
> You were doing something with the rng thing wrongly? Would still be nice to go back to the broken state for debugging, because the issue was real. Obviously not a blocker for this PR

I'll see what I can do. My guess is something related to rng. There's still a different rng problem in the PR im struggling to debug.

jessegrabowski:
I tracked the issue to the use of a shared variable in `TestLKJCorr.check_draws_match_expected`. Here's an example:


```py
n = 2
eta = 1
pm.draw(pm.LKJCorr.dist(n=n, eta=eta))

array([[ 1.        ,  0.        ],
       [-0.52803988,  0.84921957]])

n = pytensor.shared(2)
pm.draw(pm.LKJCorr.dist(n=n, eta=eta))

array([[0., 0.],
       [0., 0.]])

```

ricardoV94:
Sus

jessegrabowski:
Actually the switch fix didn't work. If we compile the function and call it twice, it errors on the 2nd call, even without symbolic inputs:

```
import pymc as pm 
n = 2
eta = 1
fn = pm.compile([], pm.LKJCorr.dist(n=n, eta=eta))

fn() # works
fn() # raises
```

ricardoV94:
> Actually the switch fix didn't work. If we compile the function and call it twice, it errors on the 2nd call, even without symbolic inputs:
> 
> ```
> import pymc as pm 
> n = 2
> eta = 1
> fn = pm.compile([], pm.LKJCorr.dist(n=n, eta=eta))
> 
> fn() # works
> fn() # raises
> ```

That's a separate error no?

jessegrabowski:
I think it's the same thing. My hypothesis is that the RNG update is wrong. In the shared case I think it was picking the the other shared variable and getting a bad state. In this case it's picking up None, and when it tries to update it breaks.

ricardoV94:
So Scan in python/cython just sets untraced_sit_sot outputs to None when n_steps=0, bug. I think we should be pragmatic here and allow n_steps to be constant like before. I'll fix upstream, but still don't see the benefit of not constant-folding n

jessegrabowski:
>  I think we should be pragmatic here and allow n_steps to be constant like before. 

allow or require? I changed it to be maximalist, but I am 1000000% ready to back off and just mandate that it be static again. We still get a win by allowing eta to be symbolic.

ricardoV94:
> > I think we should be pragmatic here and allow n_steps to be constant like before.
> 
> allow or require? I changed it to be maximalist, but I am 1000000% ready to back off and just mandate that it be static again. We still get a win by allowing eta to be symbolic.

require

ricardoV94:
Bad numba perf likely due to https://github.com/pymc-devs/pytensor/issues/2002

ricardoV94:
These are the benchmarks with the changes from pytensor v3:

```
Unrolled loop C
---------------------------------------------------------------------------------------------------- benchmark: 3 tests ---------------------------------------------------------------------------------------------------
Name (time in us)                               Min                   Max                  Mean             StdDev                Median                IQR            Outliers           OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_lkjcorr_sampling_benchmark[2]           3.1760 (1.0)         73.0370 (1.0)          4.1251 (1.0)       1.3111 (1.0)          4.1480 (1.0)       0.8310 (1.0)       354;334  242,419.0523 (1.0)        9375           1
test_lkjcorr_sampling_benchmark[10]         82.8160 (26.08)      294.8320 (4.04)       104.6812 (25.38)    21.2259 (16.19)      105.4880 (25.43)    22.7497 (27.38)     819;287    9,552.8150 (0.04)       5903           1
test_lkjcorr_sampling_benchmark[100]     1,122.2130 (353.34)   2,687.6270 (36.80)    1,163.2385 (281.99)   87.6914 (66.89)    1,159.4930 (279.53)   28.7140 (34.55)        4;11      859.6689 (0.00)        639           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Scan C
---------------------------------------------------------------------------------------------------- benchmark: 3 tests ----------------------------------------------------------------------------------------------------
Name (time in us)                               Min                   Max                  Mean              StdDev                Median                IQR            Outliers           OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_lkjcorr_sampling_benchmark[2]           7.4940 (1.0)         53.8910 (1.0)          8.7679 (1.0)        2.0809 (1.0)          7.8950 (1.0)       1.3620 (1.0)       576;496  114,051.8640 (1.0)        5057           1
test_lkjcorr_sampling_benchmark[10]        110.2360 (14.71)    1,374.6460 (25.51)      118.4666 (13.51)     22.1089 (10.62)      117.0400 (14.82)     4.8200 (3.54)       87;521    8,441.1956 (0.07)       6850           1
test_lkjcorr_sampling_benchmark[100]     1,692.9130 (225.90)   3,854.1230 (71.52)    1,758.3064 (200.54)   142.9172 (68.68)    1,736.1940 (219.91)   66.6097 (48.91)         5;9      568.7291 (0.00)        543           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Unrolled loop numba
------------------------------------------------------------------------------------------------ benchmark: 3 tests ------------------------------------------------------------------------------------------------
Name (time in us)                             Min                   Max                Mean             StdDev              Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_lkjcorr_sampling_benchmark[2]         8.7060 (1.0)        668.1320 (1.28)       9.7113 (1.0)      11.7710 (1.0)        9.0970 (1.0)      0.1710 (1.0)        23;425      102.9729 (1.0)        3229           1
test_lkjcorr_sampling_benchmark[10]       12.4230 (1.43)     6,908.2670 (13.21)     14.0546 (1.45)     77.0932 (6.55)      12.9440 (1.42)     0.2090 (1.22)        2;583       71.1511 (0.69)       8003           1
test_lkjcorr_sampling_benchmark[100]     114.4040 (13.14)      523.1110 (1.0)      123.6932 (12.74)    12.6183 (1.07)     120.6760 (13.27)    5.0825 (29.72)     218;374        8.0845 (0.08)       5477           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Scan numba
------------------------------------------------------------------------------------------------ benchmark: 3 tests -----------------------------------------------------------------------------------------------
Name (time in us)                             Min                 Max                Mean             StdDev              Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_lkjcorr_sampling_benchmark[2]         8.8860 (1.0)       34.8860 (1.0)        9.3758 (1.0)       1.0074 (1.0)        9.2470 (1.0)       0.1400 (1.0)       191;279      106.6577 (1.0)        6147           1
test_lkjcorr_sampling_benchmark[10]       13.7360 (1.55)     125.6460 (3.60)      14.6854 (1.57)      1.9309 (1.92)      14.3570 (1.55)      0.2510 (1.79)     704;1137       68.0948 (0.64)      14582           1
test_lkjcorr_sampling_benchmark[100]     448.5810 (50.48)    754.8650 (21.64)    466.5757 (49.76)    24.2944 (24.12)    461.0440 (49.86)    14.5220 (103.73)     100;93        2.1433 (0.02)       1909           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Still doing something bad on the numba scan, I suspect https://github.com/pymc-devs/pytensor/issues/1680 is the next low level hanging fruit. Regardless this works and benefits from shorter compile time. I'm also not sure about iterating over arange instead of computing them in the loop, so practices that were done to accomodate c-backend scan may be worse in numba-land

ricardoV94:
I locally patched numba scan to not copy sit_sot in every iteration when there is only one of it anyway (basically behave like an untraced_sit_sot): 

```
Scan, no stupid copy in every iteration
------------------------------------------------------------------------------------------------ benchmark: 3 tests ------------------------------------------------------------------------------------------------
Name (time in us)                             Min                   Max                Mean             StdDev              Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_lkjcorr_sampling_benchmark[2]         9.1670 (1.0)        156.1520 (1.0)        9.9854 (1.0)       4.3959 (1.0)        9.5680 (1.0)      0.1900 (1.0)       100;636      100.1466 (1.0)        6831           1
test_lkjcorr_sampling_benchmark[10]       13.4150 (1.46)       210.6060 (1.35)      14.4569 (1.45)      5.2457 (1.19)      13.9970 (1.46)     0.3010 (1.58)     208;1033       69.1711 (0.69)      13545           1
test_lkjcorr_sampling_benchmark[100]     111.5800 (12.17)    1,135.1690 (7.27)     119.2550 (11.94)    42.8868 (9.76)     115.0160 (12.02)    4.1387 (21.78)      47;342        8.3854 (0.08)       5475           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
No more slowdown from the scan as N grows, but still constant compile time

jessegrabowski:
> I locally patched numba scan to not copy sit_sot in every iteration when there is only one of it anyway (basically behave like an untraced_sit_sot):

Can we do something with rewrites in numba mode for this case? Or is it just a purely a bug

ricardoV94:
> > I locally patched numba scan to not copy sit_sot in every iteration when there is only one of it anyway (basically behave like an untraced_sit_sot):
> 
> Can we do something with rewrites in numba mode for this case? Or is it just a purely a bug

That's what I'm doing, it needed a bugfix on numba scan, because it didn't think tensor variables may be untraced_sit_sot. I'm also doing another cleanup that showed in this graph with a useless inc_subtensor that overwrites the whole buffer. That gives another small speedup and puts scan ahead of unrolled

ricardoV94:
After https://github.com/pymc-devs/pytensor/pull/2005

```
----------------------------------------------------------------------------------------------- benchmark: 3 tests ----------------------------------------------------------------------------------------------
Name (time in us)                             Min                 Max                Mean            StdDev              Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_lkjcorr_sampling_benchmark[2]         8.8370 (1.0)       49.8130 (1.0)        9.4646 (1.0)      1.0519 (1.0)        9.2870 (1.0)      0.1800 (1.0)       277;438      105.6571 (1.0)        6342           1
test_lkjcorr_sampling_benchmark[10]       12.7930 (1.45)      53.1800 (1.07)      13.5296 (1.43)     1.1595 (1.10)      13.3150 (1.43)     0.2500 (1.39)     494;1010       73.9119 (0.70)      13087           1
test_lkjcorr_sampling_benchmark[100]     103.4930 (11.71)    138.4690 (2.78)     107.4807 (11.36)    3.9867 (3.79)     106.0980 (11.42)    1.8040 (10.02)     637;876        9.3040 (0.09)       6256           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

ricardoV94 (approved):

jessegrabowski (approved):

PR #7381, opened 2024-06-21
Do not consider dims without coords volatile if length has not changed

### LINKED ISSUE #7376: BUG: Using a length-only coord causes it to be volatile
### Describe the issue:

When a variable is defined with a coord, it's considered volatile. Then `sample_posterior_predictive` does not transfer the node over.
I believe the volatility condition that's met is https://github.com/pymc-devs/pymc/blob/9eaf92ab40ff95cf7f5cb2763377e06bb0d7be0f/pymc/sampling/forward.py#L208-L212

The coord is intended to be immutable, but after https://github.com/pymc-devs/pymc/pull/7047, coords always mutable

Somewhat related:
- https://github.com/pymc-devs/pymc/issues/7069
- https://github.com/pymc-devs/pymc/issues/6876
- https://github.com/pymc-devs/pymc/issues/7183
(lmk if this issue is considered a duplicate of anything)

```python
import pymc as pm

y_obs = [0] * 3
with pm.Model() as m:
    m.add_coord("coord0", length=1)
    x = pm.Data("x", [0] * 3)
    b = pm.Normal("b", dims="coord0")  # Causes `b` to be volatile
    # b = pm.Normal("b", shape=1)      # Works as expected (`b` is transferred)
    y = pm.Normal("y", b[0] * x, observed=y_obs)
    idata = pm.sample()
    
    pm.sample_posterior_predictive(trace=idata, var_names='y')
```

### Error message:

```shell
Sampling: [b, y]
```

### PyMC version information:

5.15.1

### Context for the issue:

Cannot use `sample_posterior_predictive` as intended

JasonTam:
## Description

Allow coords defined by length only (no values) to be considered `constant_coords` if the new length matches.

## Related Issue

- [x] Closes #7376 
- [ ] Related to #

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7381.org.readthedocs.build/en/7381/


JasonTam:
The main part I feel uneasy about is having to call `current_length.eval()` since the length stored in `model.dim_lengths` is a pt shared variable

pymc/sampling/forward.py
```diff
@@ -795,10 +795,17 @@ def sample_posterior_predictive(
     constant_coords = set()
     for dim, coord in trace_coords.items():
         current_coord = model.coords.get(dim, None)
+        current_length = model.dim_lengths.get(dim, None)
+        if hasattr(current_length, 'eval'):
+            current_length = current_length.eval()
```
ricardoV94: It can only be a shared variable in which case you can retrieve with `get_value()`, or a Constant in which case you can retrieve with `.data` That should be much more efficient than `eval` unless `eval` does something clever under the hood

ricardoV94:
Also we need a test @JasonTam

JasonTam:
@ricardoV94 
I've pulled out the constant coord check logic into its own method in order to test in isolation. As I wasn't sure how to test this against the full `sample_posterior_predictive` method (or if that's even practical). Also, that method seems to have grown quite large.

ricardoV94:
@JasonTam here is where we test this volatility logic: https://github.com/pymc-devs/pymc/blob/cbf1591295e030058a0a76108fef14ec5bc5e819/tests/sampling/test_forward.py#L113

Can you add something there instead. No need to do actual sampling, just make a scenario where a variable depends on a dim without coords

pymc/sampling/forward.py
```diff
@@ -76,6 +76,28 @@
 _log = logging.getLogger(__name__)
 
 
+def get_constant_coords(trace_coords: dict[str, np.ndarray], model: Model) -> set:
+    """Get the set of coords that have remained constant between the trace and model"""
+    constant_coords = set()
+    for dim, coord in trace_coords.items():
+        current_coord = model.coords.get(dim, None)
+        current_length = model.dim_lengths.get(dim, None)
+        if hasattr(current_length, "get_value"):
```
ricardoV94: I prefer to be more explicit. `if isinstance(current_length, TensorSharedVariable) : ....get_value() elif isinstance(current_length, TensorConstant): ....data`

JasonTam:
@ricardoV94 I'm not sure how to write the test without actually sampling. In the similar tests under `TestCompileForwardSampler`, volatile vars are checked from the output of `compile_forward_sampling_function`. But that function already takes in `constant_coords` as a parameter.

The new logic of determining `constant_coords` can either be tested via the new low level function `get_constant_coords` (as currently written in the PR), or in the method which calls it, `sample_posterior_predictive`.

Maybe I'm misunderstanding what you mean.

pymc/sampling/forward.py
JasonTam: Implemented in https://github.com/pymc-devs/pymc/pull/7381/commits/694660c5b1fb12ebd91ae6e9d323f5433fda7deb

ricardoV94:
You're right @JasonTam, I was confused by the fact that the changes in constant data are checked inside the compile_forward_function but the coords are not. I guess this is the case because coords are not part of the graph while SharedVariables are.

ricardoV94 (dismissed):

tests/sampling/test_forward.py
```diff
@@ -1669,6 +1670,20 @@ def test_Triangular(
         assert prior["target"].shape == (prior_samples, *shape)
 
 
+def test_get_constant_coords():
+    with pm.Model() as model:
```
ricardoV94, line 1714: Can we add a more integration-like test like?

Spit-balling, perhaps something like:

```python
with pm.Model() as m:
  m.add_coord("trial", length=3)
  x = pm.Normal("x", shape="trial")
  y = pm.Deterministic("y", x.mean())

# pass dims as well  
idata = az.from_dict("x": [[np.pi, np.pi, np.pi]])

with m:
  pp1 = pm.sample_posterior_predictive(idata, var_names=["y"]).posterior_predictive

assert pp1["y"] == np.pi

with m:
  # change coord length 
  pp2 = pm.sample_posterior_predictive(idata, var_names=["y"]).posterior_predictive

assert pp2["y"] != np.pi

```

JasonTam: Does this look ok? https://github.com/pymc-devs/pymc/pull/7381/commits/d40fffec21077e7b4906b320f0605c5f4e67ddcb

tests/sampling/test_forward.py
```diff
@@ -427,6 +428,36 @@ def test_mutable_coords_volatile(self):
...
+            coords={"trial": range(3)},
+            dims={"x": ["trial"]},
+        )
+        with model:
+            pp_same_len = pm.sample_posterior_predictive(
+                trace_same_len, var_names=["y"]
+            ).posterior_predictive
+        assert pp_same_len["y"] == np.pi
+
+        # Coord length changed -- `x` is volatile
```
ricardoV94, line 450: I prefer to change the dim length in the pymc model and pass exactly the same idata to sample_posterior_predictive as was done in the first call.

ricardoV94: The way to update an existing dim is with `Model.set_dim()`

JasonTam: Sorry, just got to this now. Implemented in https://github.com/pymc-devs/pymc/pull/7381/commits/49bbf8c68f44bf25454598af41096dabbcd001ad -- hope I understood you correctly

tests/sampling/test_forward.py
```diff
@@ -428,6 +429,45 @@ def test_mutable_coords_volatile(self):
...
+            dims={"x": ["trial"]},
+        )
+        with model:
+            pp_diff_len = pm.sample_posterior_predictive(
+                trace_diff_len, var_names=["y"]
+            ).posterior_predictive
+        assert pp_diff_len["y"] != np.pi
+
+        # Changing the dim length on the model itself
+        # -- `x` is volatile even though trace has same len
```
ricardoV94:
```suggestion
        # -- `x` is volatile because trace has same len as original model
```

tests/sampling/test_forward.py
ricardoV94: Yup you got it!

ricardoV94 (approved):

ricardoV94:
Thanks @JasonTam

Issue #7384, opened 2024-06-24
RTD is broken


twiecki:
### Description

```
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc/conda/7383/lib/python3.11/site-packages/jupyter_cache/executors/utils.py", line 58, in single_nb_execution
    executenb(
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc/conda/7383/lib/python3.11/site-packages/nbclient/client.py", line 1314, in execute
    return NotebookClient(nb=nb, resources=resources, km=km, **kwargs).execute()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc/conda/7383/lib/python3.11/site-packages/jupyter_core/utils/__init__.py", line 165, in wrapped
    return loop.run_until_complete(inner)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc/conda/7383/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc/conda/7383/lib/python3.11/site-packages/nbclient/client.py", line 709, in async_execute
    await self.async_execute_cell(
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc/conda/7383/lib/python3.11/site-packages/nbclient/client.py", line 1062, in async_execute_cell
    await self._check_raise_for_error(cell, cell_index, exec_reply)
  File "/home/docs/checkouts/readthedocs.org/user_builds/pymc/conda/7383/lib/python3.11/site-packages/nbclient/client.py", line 918, in _check_raise_for_error
    raise CellExecutionError.from_cell_and_msg(cell, exec_reply_content)
nbclient.exceptions.CellExecutionError: An error occurred while executing the following cell:
------------------
with pm.Model() as model_1:
    a = pm.Normal("a", 0.0, 10.0)
    b = pm.Normal("b", 0.0, 10.0)

    mu = a + b * predictor_scaled
    sigma = pm.Exponential("sigma", 1.0)

[...]

PR #7389, opened 2024-06-25
Run Notebooks locally

ricardoV94:
## Related Issue

- [ ] Related to #7384 

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7389.org.readthedocs.build/en/7389/


docs/source/conf.py
```diff
@@ -98,7 +99,8 @@
 
 # General information about the project.
 project = "PyMC"
-copyright = "2021, The PyMC Development Team"
+year = date.today().strftime("%Y")
+copyright = f"{year}, The PyMC Development Team"
```
OriolAbril: This would mean we renounce the copyright of the content created before that date. If you want to be more technically correct we can change it to "2021-present, PyMC development team", probably even an earlier year. I have updated many dates in the past but have since learned that is actually wrong

OriolAbril (changes requested):

docs/source/conf.py
ricardoV94: Thanks, fixed

OriolAbril (approved):

PR #7390, opened 2024-06-25
Assert ndim and number of dims match

ricardoV94:
Fixes bug where invalid number of dims was given to Deterministic introduced by OrderedLogistic/Probit

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7390.org.readthedocs.build/en/7390/

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -1185,13 +1185,6 @@ class Categorical(Discrete):
         )
 
 
-class _OrderedLogistic(Categorical):
-    r"""
-    Underlying class for ordered logistic distributions.
-    See docs for the OrderedLogistic wrapper class for more details on how to use it in models.
-    """
-
-
 class OrderedLogistic:
     R"""Ordered Logistic distribution.
 
@@ -1263,7 +1256,7 @@ class OrderedLogistic:
     def __new__(cls, name, eta, cutpoints, compute_p=True, **kwargs):
         p = cls.compute_p(eta, cutpoints)
         if compute_p:
-            p = pm.Deterministic(f"{name}_probs", p, dims=kwargs.get("dims"))
+            p = pm.Deterministic(f"{name}_probs", p)
         out_rv = Categorical(name, p=p, **kwargs)
         return out_rv
 
@@ -1367,7 +1360,7 @@ class OrderedProbit:
     def __new__(cls, name, eta, cutpoints, sigma=1, compute_p=True, **kwargs):
         p = cls.compute_p(eta, cutpoints, sigma)
         if compute_p:
-            p = pm.Deterministic(f"{name}_probs", p, dims=kwargs.get("dims"))
+            p = pm.Deterministic(f"{name}_probs", p)
         out_rv = Categorical(name, p=p, **kwargs)
         return out_rv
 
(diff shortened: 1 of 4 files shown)


tests/model/test_core.py
```diff
@@ -890,7 +890,20 @@ def test_add_named_variable_checks_dim_name(self):
...
+        with pm.Model(coords={"bad": range(6)}) as m:
+            with pytest.raises(ValueError, match=match):
+                m.add_named_variable(pt.random.normal(size=(6, 6, 6), name="a"), dims=("bad",))
+
+            with pytest.raises(ValueError, match=match):
+                m.add_named_variable(pt.random.normal(size=(6, 6, 6), name="b"), dims="bad")
+
+            # For variables without ndim we can't check
+            m.add_named_variable(pytensor.as_symbolic(None, name="c"), dims=("bad",))
+            assert m.named_vars_to_dims == {"c": ("bad",)}
```
ricardoV94, line 904: Alternatively we could forbid these variables from having dims

pymc/distributions/discrete.py
```diff
@@ -1263,7 +1256,7 @@ class OrderedLogistic:
     def __new__(cls, name, eta, cutpoints, compute_p=True, **kwargs):
         p = cls.compute_p(eta, cutpoints)
         if compute_p:
-            p = pm.Deterministic(f"{name}_probs", p, dims=kwargs.get("dims"))
+            p = pm.Deterministic(f"{name}_probs", p)
```
ricardoV94, line 1259: This is not trivial. p may have more or less dims than the respective Categorical variable, and these may or not be broadcasting.  I decided it was not worth the trouble for the time being. In general we still need to figure out an approach for automatic dims (see #6828)

jessegrabowski: I think partial dim assignment is currently supported, so you could do something like `dims=(None,) * dim_diff + ('categories', )`, no? The last one at least is always known iiuc

ricardoV94: No this is not sufficient either because p could have a broadcastable dim on the same batch dimensions of the underlying categorical variable. `pm.Categorical(..., p=[[0.5, 0.5]], dims="obs”)`. You cannot say p has dims ("obs", None) because it has shape (1, 2) not (2, 2)

ricardoV94: Or you're suggestion adding a "categories dim" regardless of the rest. That has another problem, what if the user creates 2 variables with different number of categories? Then again you have two dims that are incompatible (same name, different shape). 

That can also be solved (by using the variable name) but I don't want to do it here, I just want to not break stuff for now. 

Besides that was not what the code was doing, it seems like the biggest focus was on reusing the user provided dims, not auto naming the extra dim

jessegrabowski: I was assuming that "categories" already was provided somewhere, not that we generate or auto-name anything.

But yes I understand this is a rabbit hole, focusing on not breaking things is fine.

ricardoV94: In that case the PR is ready for review :)

jessegrabowski (approved):

Issue #6335, opened 2022-11-24
BUG: Deterministic variables with `dims` containing `None` break `model_to_graphviz`


michaelosthege:
### Describe the issue:

When a `Deterministic` variable is created, one may pass a `dims=(None, "time")` which is a valid `dims` tuple.

However, when attempting a `pm.model_to_graphviz` later, a `KeyError: None` is encountered when accessing `Model.dim_lengths`.

### Reproduceable code example:

```python
import numpy as np
import pymc as pm

with pm.Model(coords=dict(time=np.arange(5))) as pmodel:
    data = at.as_tensor(np.ones((3, 5)))
    # This works:
    # pm.Normal("n", data, dims=(None, "time"))

    # This does not:
    pm.Deterministic("n", data, dims=(None, "time"))
    
pm.model_to_graphviz(pmodel)
```

### Error message:

```shell
File pymc\model_graph.py:198, in <genexpr>(.0)
    195 v = self.model[var_name]
    196 if var_name in self.model.RV_dims:
    197     plate_label = " x ".join(
--> 198         f"{d} ({self._eval(self.model.dim_lengths[d])})"
    199         for d in self.model.RV_dims[var_name]
    200     )
    201 else:
    202     plate_label = " x ".join(map(str, self._eval(v.shape)))

KeyError: None
```

### PyMC version information:

4.1.7

michaelosthege:
Using `pm.ConstantData` instead of `pm.Deterministic` breaks in a different way:

```python
File pymc\data.py:716, in Data(name, value, dims, coords, export_index_as_coords, mutable, **kwargs)
    706         if not dname in model.dim_lengths:
    707             model.add_coord(
    708                 name=dname,
    709                 # Note: Coordinate values can't be taken from
   (...)
    713                 length=xshape[d],
    714             )
--> 716 model.add_random_variable(x, dims=dims)
    718 return x

File pymc\model.py:1552, in Model.add_random_variable(self, var, dims)
   1550             raise ValueError(f"Dimension {dim} is not specified in `coords`.")
   1551     if any(var.name == dim for dim in dims):
-> 1552         raise ValueError(f"Variable `{var.name}` has the same name as its dimension label.")
   1553     self._RV_dims[var.name] = dims
   1555 self.named_vars[var.name] = var

ValueError: Variable `None` has the same name as its dimension label.
```

michaelosthege:
I think this might be a rather simple fix: Checking `is None` and then printing more useful information if available.

Issue #7048, opened 2023-12-04
Reasses dims without coords


ricardoV94:
### Description

It is currently possible to create dims without coords via `pm.Data`:

```python
import pymc as pm

with pm.Model() as m:
  x = pm.MutableData("x", [0, 1, 2], dims=("dim_without_coord",))
  y = pm.Normal("y", x.sum(), dims=("dim_without_coord",))
```

We used to allow this also from RVs in the past but reverted in https://github.com/pymc-devs/pymc/pull/6112
I don't see a fundamental difference between the two cases, and in that PR I just missed this. 

I would suggest we remove this functionality because it adds quite some complexity in the Model class, including fancy checks to see if a Data variable defined from dims without coords of another Data variable is being resized before the original one (but after is fine?): https://github.com/pymc-devs/pymc/blob/02cbac6592506a547117b2daa3f05847d16fe063/tests/model/test_core.py#L923-L937

(The unawareness) of this option has also led to bugs that were only patched recently: https://github.com/pymc-devs/pymc/pull/6882

ricardoV94:
I think we should reassess, it's still a common source of bugs and confusion:
#7392 
#7376

ricardoV94:
We should also reassess allowing `None` dims for the time being: as in `pm.Deterministic("x", ..., dims=(None, "trial")`. This was discussed in #6485 

Since the code is mostly broken anyway why not be strict and reassess when the need becomes obvious?

Issue #6485, opened 2023-01-27
BUG: Setting None coord is problematic


ferrine:
### Describe the issue:

When setting a coord None. Something goes broken in pymc 5.0.2.

### Reproduceable code example:

```python
# This is good
import pymc as pm
import numpy as np
with pm.Model(coords=dict(d1=range(2), d2=range(6))) as model:
    pm.Data("a", np.random.randn(2, 6), dims=("d1", "d2"), mutable=True)
    pm.Normal("b", 10)
    t = pm.sample(1, tune=1)

# this is broken
import pymc as pm
import numpy as np
with pm.Model(coords=dict(d1=range(2), d2=range(6))) as model:
    pm.Data("a", np.random.randn(2, 6), dims=(None, "d2"), mutable=True)
    pm.Normal("b", 10)
    t = pm.sample(1, tune=1)
```

### Error message:
<details>

```shell
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[14], line 6
      4 pm.Data("a", np.random.randn(2, 6), dims=(None, "d2"), mutable=True)
      5 pm.Normal("b", 10)
----> 6 t = pm.sample(1, tune=1)

File ~/micromamba/envs/gp_mmm/lib/python3.9/site-packages/pymc/sampling/mcmc.py:612, in sample(draws, step, init, n_init, initvals, trace, chains, cores, tune, progressbar, model, random_seed, discard_tuned_samples, compute_convergence_checks, callback, jitter_max_retries, return_inferencedata, keep_warning_stat, idata_kwargs, mp_ctx, **kwargs)
    610 if idata_kwargs:
    611     ikwargs.update(idata_kwargs)
--> 612 idata = pm.to_inference_data(mtrace, **ikwargs)
    614 if compute_convergence_checks:
    615     if draws - tune < 100:

File ~/micromamba/envs/gp_mmm/lib/python3.9/site-packages/pymc/backends/arviz.py:485, in to_inference_data(trace, prior, posterior_predictive, log_likelihood, coords, dims, sample_dims, model, save_warmup, include_transformed)
    482 if isinstance(trace, InferenceData):
    483     return trace
--> 485 return InferenceDataConverter(
    486     trace=trace,
    487     prior=prior,
    488     posterior_predictive=posterior_predictive,
    489     log_likelihood=log_likelihood,
    490     coords=coords,
    491     dims=dims,
    492     sample_dims=sample_dims,
    493     model=model,
[...]

PR #7392, opened 2024-06-26
Refactor model graph and allow suppressing dim lengths

### LINKED ISSUE #7319: ENH: Customizable plate labels
### Before

```python
# Currently would have to create graphviz instance
graph = pm.model_to_graphviz(model)
# Save output to .dot file and manually modify subgraph labels
print(graph)
# Resave as pdf using dot cli 
```

### After

```python
# Might be similar to that of where the entry point to the change would be in `pm.model_to_graphviz`

def named_vars_plate_label(
    var: TensorVariable, 
    named_vars_to_dims: dict[str, Sequence[str]], 
    dim_lengths: dict[str, int],
) -> str: 
    dims = named_vars_to_dims[var.name]
    return " x ".join(str(dim) for dim in dims)
    
plate_formatters = {
    "Named Vars": named_vars_plate_label,
}
pm.model_to_graphviz(model, plate_formatters=plate_formatters)
```

### Context for the issue:

I am interested in making graphviz diagraphs with a bit more customization of the plate labels than currently supported. Specifically, I would like to allow for labels without the shape size to provide a more abstract representation of the model. My specific example might look like this:
example-graph

I am willing to take up this issue. Wondering if there are any preferences on the implementation details? My idea would involve extracting this logic [here](https://github.com/pymc-devs/pymc/blob/main/pymc/model_graph.py#L311-L327) and providing some logic based on: 
- var : TensorVariable
- named_vars_to_dims : var.name mapping to coord names
- dim_lengths : coord names mapping to either TensorVariable or pre-evaluated shapes (addressing [this TODO](https://github.com/pymc-devs/pymc/blob/43b40de8ee84d7ad9813c83824a850b4dce8f924/pymc/model_graph.py#L305-L307))

Related to #6716

williambdean:
## Description

Pulled any graph related information into two methods: 

1. `get_plates`: Get plate meta information and the nodes that are associated with each plate
2. `edges`: Edges between nodes as a list[tuple[VarName, VarName]]

The `get_plates` methods returns a list of `Plate` objects which store all the variable information. That data include:

1. `DimInfo` with stores the dim names and lengths
1. `NodeInfo` which stores the model variable and it's NodeType in the graph (introduced in #7302)
1. `Plate` which is a collection of the `DimInfo` and `list[NodeInfo]`

With `list[tuple[VarName, VarName]]` and `list[Plate]`, a user can now make use of the exposed `make_graph` and `make_networkx` functions to create customized graphviz or networkx graphs.

The previous behavior of `model_to_graphviz` and `model_to_networkx` is still maintained. However, there is a new `include_dim_lengths` parameter that can be used to include the dim lengths in the plate labels.

The previous issue #6335 behavior has changed to now include all the variables on a plate with dlen instead of var_name_dim{d}. (See examples below)

## Related Issue

- [x] Closes #7319
- [ ] Related to 

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7392.org.readthedocs.build/en/7392/

### DIFF
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -43,17 +43,17 @@ __all__ = (
 @dataclass
 class DimInfo:
     names: tuple[str | None]
-    sizes: tuple[int]
+    lengths: tuple[int]
 
     def __post_init__(self) -> None:
-        if len(self.names) != len(self.sizes):
-            raise ValueError("The number of names and sizes must be equal.")
+        if len(self.names) != len(self.lengths):
+            raise ValueError("The number of names and lengths must be equal.")
 
     def __hash__(self):
-        return hash((self.names, self.sizes))
+        return hash((self.names, self.lengths))
 
     def __bool__(self) -> bool:
-        return len(self.sizes) > 0 or len(self.names) > 0
+        return len(self.lengths) > 0 or len(self.names) > 0
 
 
 def create_plate_label(
@@ -72,7 +72,7 @@ def create_plate_label(
 
         return label
 
-    values = enumerate(zip(dim_info.names, dim_info.sizes))
+    values = enumerate(zip(dim_info.names, dim_info.lengths))
     return " x ".join(create_label(d, dname, dlen) for d, (dname, dlen) in values)
 
 
@@ -365,21 +365,21 @@ class ModelGraph:
             if var_name in self.model.named_vars_to_dims:
                 # The RV is associated with `dims` information.
                 names = []
-                sizes = []
+                lengths = []
                 for d, dname in enumerate(self.model.named_vars_to_dims[var_name]):
                     names.append(dname)
-                    sizes.append(dim_lengths.get(dname, shape[d]))
+                    lengths.append(dim_lengths.get(dname, shape[d]))
 
                 dim_info = DimInfo(
                     names=tuple(names),
-                    sizes=tuple(sizes),
+                    lengths=tuple(lengths),
                 )
             else:
                 # The RV has no `dims` information.
                 dim_size = len(shape)
                 dim_info = DimInfo(
                     names=tuple([None] * dim_size),
-                    sizes=tuple(shape),
+                    lengths=tuple(shape),
                 )
 
             v = self.model[var_name]
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -42,7 +42,7 @@ __all__ = (
 
 
 @dataclass
-class PlateMeta:
+class DimInfo:
     names: tuple[str | None]
     sizes: tuple[int]
 
@@ -55,7 +55,7 @@ class PlateMeta:
 
 def create_plate_label(
     var_name: str,
-    plate_meta: PlateMeta,
+    dim_info: DimInfo,
     include_size: bool = True,
 ) -> str:
     def create_label(d: int, dname: str, dlen: int):
@@ -70,7 +70,7 @@ def create_plate_label(
         return label
 
     values = enumerate(
-        zip_longest(plate_meta.names, plate_meta.sizes, fillvalue=None),
+        zip_longest(dim_info.names, dim_info.sizes, fillvalue=None),
     )
     return " x ".join(create_label(d, dname, dlen) for d, (dname, dlen) in values)
 
@@ -90,7 +90,7 @@ class NodeType(str, Enum):
 
 
 @dataclass
-class NodeMeta:
+class NodeInfo:
     var: TensorVariable
     node_type: NodeType
 
@@ -100,8 +100,8 @@ class NodeMeta:
 
 @dataclass
 class Plate:
-    meta: PlateMeta | None
-    variables: list[NodeMeta]
+    dim_info: DimInfo | None
+    variables: list[NodeInfo]
 
 
 GraphvizNodeKwargs = dict[str, Any]
@@ -208,7 +208,7 @@ def update_node_formatters(node_formatters: NodeTypeFormatterMapping) -> NodeTyp
 
 
 def _make_node(
-    node: NodeMeta,
+    node: NodeInfo,
     *,
     node_formatters: NodeTypeFormatterMapping,
     add_node: Callable[[str, ...], None],
@@ -369,25 +369,29 @@ class ModelGraph:
                     names.append(dname)
                     sizes.append(dim_lengths.get(dname, shape[d]))
 
-                plate_meta = PlateMeta(
+                dim_info = DimInfo(
                     names=tuple(names),
                     sizes=tuple(sizes),
                 )
             else:
                 # The RV has no `dims` information.
-                plate_meta = PlateMeta(
-                    names=(),
+                dim_size = len(shape)
+                dim_info = DimInfo(
+                    names=tuple([None] * dim_size),
                     sizes=tuple(shape),
                 )
 
             v = self.model[var_name]
             node_type = get_node_type(var_name, self.model)
-            var = NodeMeta(var=v, node_type=node_type)
-            plates[plate_meta].add(var)
+            var = NodeInfo(var=v, node_type=node_type)
+            plates[dim_info].add(var)
 
         return [
-            Plate(meta=plate_meta if plate_meta else None, variables=list(variables))
-            for plate_meta, variables in plates.items()
+            Plate(
+                dim_info=dim_info if dim_info else None,
+                variables=list(variables),
+            )
+            for dim_info, variables in plates.items()
         ]
 
     def edges(
@@ -445,11 +449,11 @@ def make_graph(
 
     graph = graphviz.Digraph(name)
     for plate in plates:
-        if plate.meta:
+        if plate.dim_info:
             # must be preceded by 'cluster' to get a box around it
             plate_label = create_plate_label(
                 plate.variables[0].var.name,
-                plate.meta,
+                plate.dim_info,
                 include_size=include_dim_lengths,
             )
             with graph.subgraph(name="cluster" + plate_label) as sub:
@@ -517,12 +521,12 @@ def make_networkx(
 
     graphnetwork = networkx.DiGraph(name=name)
     for plate in plates:
-        if plate.meta:
+        if plate.dim_info:
             # # must be preceded by 'cluster' to get a box around it
 
             plate_label = create_plate_label(
                 plate.variables[0].var.name,
-                plate.meta,
+                plate.dim_info,
                 include_size=include_dim_lengths,
             )
             subgraphnetwork = networkx.DiGraph(name="cluster" + plate_label, label=plate_label)
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -105,6 +105,12 @@ class Plate:
     dim_info: DimInfo
     variables: list[NodeInfo]
 
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, Plate):
+            return False
+
+        return self.dim_info == other.dim_info and set(self.variables) == set(other.variables)
+
 
 GraphvizNodeKwargs = dict[str, Any]
 NodeFormatter = Callable[[TensorVariable], GraphvizNodeKwargs]
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -49,6 +49,9 @@ class PlateMeta:
     def __hash__(self):
         return hash((self.names, self.sizes))
 
+    def __bool__(self) -> bool:
+        return len(self.sizes) > 0 or len(self.names) > 0
+
 
 def create_plate_label(
     var_name: str,
@@ -97,7 +100,7 @@ class NodeMeta:
 
 @dataclass
 class Plate:
-    meta: PlateMeta
+    meta: PlateMeta | None
     variables: list[NodeMeta]
 
 
@@ -202,6 +205,24 @@ def update_node_formatters(node_formatters: NodeTypeFormatterMapping) -> NodeTyp
         )
 
     return node_formatters
+
+
+def _make_node(
+    node: NodeMeta,
+    *,
+    node_formatters: NodeTypeFormatterMapping,
+    add_node: Callable[[str, ...], None],
+    cluster: bool = False,
+    formatting: str = "plain",
+):
+    """Attaches the given variable to a graphviz or networkx Digraph"""
+    node_formatter = node_formatters[node.node_type]
+    kwargs = node_formatter(node.var)
+
+    if cluster:
+        kwargs["cluster"] = cluster
+
+    add_node(node.var.name.replace(":", "&"), **kwargs)
 
 
 class ModelGraph:
@@ -311,24 +332,6 @@ class ModelGraph:
 
         return input_map
 
-    def _make_node(
-        self,
-        node: NodeMeta,
-        *,
-        node_formatters: NodeTypeFormatterMapping,
-        add_node: Callable[[str, ...], None],
-        cluster: bool = False,
-        formatting: str = "plain",
-    ):
-        """Attaches the given variable to a graphviz or networkx Digraph"""
-        node_formatter = node_formatters[node.node_type]
-        kwargs = node_formatter(node.var)
-
-        if cluster:
-            kwargs["cluster"] = cluster
-
-        add_node(node.var.name.replace(":", "&"), **kwargs)
-
     def get_plates(
         self,
         var_names: Iterable[VarName] | None = None,
@@ -380,7 +383,7 @@ class ModelGraph:
             plates[plate_meta].add(var)
 
         return [
-            Plate(meta=plate_meta, variables=list(variables))
+            Plate(meta=plate_meta if plate_meta else None, variables=list(variables))
             for plate_meta, variables in plates.items()
         ]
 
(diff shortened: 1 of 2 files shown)


pymc/model_graph.py
```diff
@@ -302,33 +335,50 @@ def get_plates(self, var_names: Iterable[VarName] | None = None) -> dict[str, se
...
+                )
+
+            v = self.model[var_name]
+            node_type = get_node_type(var_name, self.model)
+            var = NodeMeta(var=v, node_type=node_type)
+            plates[plate_meta].add(var)
 
         return dict(plates)
 
+    def edges(self, var_names: Iterable[VarName] | None = None):
```
ricardoV94: Any reason to make this a generator?

pymc/model_graph.py
```diff
@@ -302,33 +335,50 @@ def get_plates(self, var_names: Iterable[VarName] | None = None) -> dict[str, se
...
+        for child, parents in self.make_compute_graph(var_names=var_names).items():
+            # parents is a set of rv names that precede child rv nodes
+            for parent in parents:
+                yield child.replace(":", "&"), parent.replace(":", "&")
+
     def make_graph(
```
ricardoV94: Should `make_graph` and `make_networkx` now be functions that take plates and edges as inputs?

pymc/model_graph.py
williambdean: Would you prefer returning a list of tuples? or questioning pulling this out of the two level loop?

pymc/model_graph.py
williambdean: That would just remove calling `get_plates` and `edges` methods. Don't have much of a preference

pymc/model_graph.py
```diff
@@ -357,26 +409,31 @@ def make_graph(
         node_formatters = update_node_formatters(node_formatters)
 
         graph = graphviz.Digraph(self.model.name)
-        for plate_label, all_var_names in self.get_plates(var_names).items():
-            if plate_label:
+        for plate_meta, all_vars in self.get_plates(var_names).items():
+            if plate_meta:
                 # must be preceded by 'cluster' to get a box around it
+                plate_label = create_plate_label(plate_meta, include_size=include_shape_size)
```
williambdean: Noticing that the plate_label actually depends on var_name in case of previous "{var_name}_dim{d}". However, the plate_label is required before the looping of `all_var_names`. i.e. `all_var_names` is assumed to be one element? Maybe that should be an explicit case?

pymc/model_graph.py
ricardoV94: Yeah I was talking about returning the evaluated generator. No opinion on the two level looping

pymc/model_graph.py
ricardoV94: It would make it more modular, in that if you find a way to create your own plates and edges, you can just pass it to the functions that then display it?

pymc/model_graph.py
ricardoV94: Didn't manage to follow, can you explain again?

pymc/model_graph.py
williambdean: Yeah, sure. I think that makes sense then. 

The dictionary of {PlateMeta : set[NodeMeta]} is a bit weird and hard to work with. i.e. set is not subscritable and looking up by PlateMeta key is a bit tricky.

I was thinking of having another object, `Plate` which would be:

```python
@dataclass 
class Plate: 
    plate_meta: PlateMeta
    nodes: list[NodeMeta]
```
and that would be in the input to `make_graph` and `make_networkx` instead. Making the signature: (plates: list[Plate], edges: list[tuple[str, str]], ...)

pymc/model_graph.py
wd60622: The `graph.subgraph` name of \"cluster\" + plate_label is dependent on the var_name which used to be constructed in the `get_plates` method (the previous keys).

However, after the subgraph is constructed, the `all_var_names` is looped over. This is assuming that `all_var_names` is only one element since the plate_label is used in the subgraph name.

wd60622: The `graph.subgraph` name of \"cluster\" + plate_label is dependent on the var_name which used to be constructed in the `get_plates` method (the previous keys).

However, after the subgraph is constructed, the `all_var_names` is looped over. This is assuming that `all_var_names` is only one element since the plate_label is used in the subgraph name.

wd60622: The `graph.subgraph` name of \"cluster\" + plate_label is dependent on the var_name which used to be constructed in the `get_plates` method (the previous keys).

However, after the subgraph is constructed, the `all_var_names` is looped over. This is assuming that `all_var_names` is only one element since the plate_label is used in the subgraph name.

williambdean: The `graph.subgraph` name of \"cluster\" + plate_label is dependent on the var_name which used to be constructed in the `get_plates` method (the previous keys of dictionary where the plate_label).

However, after the subgraph is constructed, the `all_var_names` is looped over. This is assuming that `all_var_names` is only one element since the plate_label is used in the subgraph name.

pymc/model_graph.py
williambdean: Also, does it make sense as a method still? Do you see model_to_graphviz taking this input as well?

pymc/model_graph.py
```diff
@@ -357,26 +436,35 @@ def make_graph(
         node_formatters = update_node_formatters(node_formatters)
 
         graph = graphviz.Digraph(self.model.name)
-        for plate_label, all_var_names in self.get_plates(var_names).items():
-            if plate_label:
+        for plate in self.get_plates(var_names):
+            plate_meta = plate.meta
+            all_vars = plate.variables
+            if plate_meta.names or plate_meta.sizes:
```
ricardoV94: Can we simplify? Could plate_meta be None for the scalar variables?

pymc/model_graph.py
ricardoV94: Lost track of the specific methods we're discussing. My low resolution guess was that once we have the plates / edges we can just pass them to a function that uses those to render graphviz or networkx graphs. Let me know if you were asking about something else or see a problem (or no point) with that

pymc/model_graph.py
williambdean: This logic would still be needed somewhere. Likely in `get_plates` then. 
How about having the `__bool__` method for `Plate` class that does this logic. 
Then would act like None and read like: 
```python
if plate_meta: # Truthy if sizes or names
    # plate_meta has sizes or names that are not empty tuples
```

pymc/model_graph.py
williambdean: Sounds good. Let me push something up and you can give feedback

pymc/model_graph.py
ricardoV94: You have that information when you defined the plate.meta no? Can't you do it immediately?

pymc/model_graph.py
```diff
@@ -49,6 +49,9 @@ class PlateMeta:
     def __hash__(self):
         return hash((self.names, self.sizes))
 
+    def __bool__(self) -> bool:
+        return len(self.sizes) > 0 or len(self.names) > 0
```
ricardoV94: What is a plate without names?

pymc/model_graph.py
williambdean: Just pushed. 
If user has arbitrary list[Plate] and list[tuple[VarName, VarName]] then they can use `make_graph` or `make_networkx` in order to make the graphviz or networkx, respectively. 
`pm.model_to_graphviz` and `pm.model_to_networkx` are still wrappers. 
`ModelGraph` class can be used to create the plates and edges in the previous manner if desired with the `get_plates` and `edges` methods

pymc/model_graph.py
```diff
@@ -302,164 +348,215 @@ def get_plates(self, var_names: Iterable[VarName] | None = None) -> dict[str, se
...
+            shape: tuple[int, ...] = tuple(fast_eval(v.shape))
             if var_name in self.model.named_vars_to_dims:
                 # The RV is associated with `dims` information.
+                names = []
+                sizes = []
                 for d, dname in enumerate(self.model.named_vars_to_dims[var_name]):
-                    if dname is None:
-                        # Unnamed dimension in a `dims` tuple!
-                        dlen = shape[d]
-                        dname = f"{var_name}_dim{d}"
```
williambdean, line 319: @ricardoV94 do you know a graph where this would be used?

pymc/model_graph.py
```diff
@@ -302,164 +348,215 @@ def get_plates(self, var_names: Iterable[VarName] | None = None) -> dict[str, se
...
-                        dlen = fast_eval(self.model.dim_lengths[dname])
-                    dim_labels.append(f"{dname} ({dlen})")
-                plate_label = " x ".join(dim_labels)
+                    names.append(dname)
+                    sizes.append(dim_lengths.get(dname, shape[d]))
+
+                plate_meta = PlateMeta(
+                    names=tuple(names),
+                    sizes=tuple(sizes),
+                )
```
ricardoV94: I don't understand this tbh. Are we creating one plate per variable? But a plate can contain multiple variables?

Also names is ambiguous, it is dim_names? We should name it like that to distinguish from var_names?
Also sizes -> dim_lengths

pymc/model_graph.py
williambdean:
```python
with pm.Model(): 
    pm.Normal("y", shape=3)
```
y has sizes but no dim names. 
Currently creates Plate(meta=PlateMeta(names=(), sizes=(5, )), variables=[NodeMeta(var=y, node_type=...)])

Think there should be some cases to test now that this logic is exposed. Will be much easier to confirm

pymc/model_graph.py
ricardoV94: Ah plates are hashable... so you mutate the same thing...

williambdean: Just working with what was there. The historical { str: set[VarName] } is created with loop which I changed to { PlateMeta : set[NodeMeta] } 
But switched to list[Plate] ultimately. 
Ideally, there could be more straight-foward path to list[Plate]

pymc/model_graph.py
williambdean: Yes. I changed to have it happen in the `get_plates` methods. Scalars will have Plate(meta=None, variables=[...])

pymc/model_graph.py
williambdean: names here are dim names

pymc/model_graph.py
ricardoV94: I think it's enough to check for `sizes`? It is not possible for a plate to have names, but not sizes?

We should rename those to `dim_names, and dim_lengths`. And perhaps use `None` for dim_lengths for which we don't know the name?

pymc/model_graph.py
ricardoV94: What happens for a Deterministic with `dims=("test_dim", None)`? Apparently we still allow None dims for things that are not RVs

pymc/model_graph.py
ricardoV94: I mentioned a case here: https://github.com/pymc-devs/pymc/pull/7392#discussion_r1656914362

wd60622:
The name `Plate` and `PlateMeta` come from the historical `get_plates` method of `ModelGraph`. However, get_plates also get scalars which were "" before and now `Plate(meta=None, variables=[...])`

Is `Plate` still a good name? It is a collection of variables all with the same dims. Plate in my mind is Bayesian graphical model and might deviate with the scalars.
`PlateMeta` might be more suited as `DimsMeta` since the names and sizes are the dims of the variables

Any thoughts here on terminology?

pymc/model_graph.py
ricardoV94: This logic feels rather convoluted to be honest. Maybe we can take a step back and see what is actually needed. 
Step1: Collect the dim names and dim lengths of every variable we want to plot. This seems simple enough, and we can do in a loop
Step2: Merge variables that have identical dim_names and dim_lengths into "plates". The hashable Plate thing may be a good trick to achieve that, or just a defaultdict with keys: `tuple[dim_names, dim_lengths]`

Would the code be more readable if we didn't try to do both things at once?

ricardoV94:
I'm okay with Plate or Cluster. Why the Meta in it?

pymc/model_graph.py
```diff
@@ -302,164 +348,215 @@ def get_plates(self, var_names: Iterable[VarName] | None = None) -> dict[str, se
...
-                    if dname is None:
-                        # Unnamed dimension in a `dims` tuple!
-                        dlen = shape[d]
-                        dname = f"{var_name}_dim{d}"
-                    else:
-                        dlen = fast_eval(self.model.dim_lengths[dname])
-                    dim_labels.append(f"{dname} ({dlen})")
-                plate_label = " x ".join(dim_labels)
+                    names.append(dname)
+                    sizes.append(dim_lengths.get(dname, shape[d]))
```
ricardoV94: This is weird, why would sizes have dim_names in it? I think sizes should be tuple[int, ...], and names should be tuple[str | None, ...]?

pymc/model_graph.py
ricardoV94: Edit: Updated comment above

pymc/model_graph.py
ricardoV94: So IIUC, scalars should belong to a "Plate" with dim_names = (), and dim_lengths = ()?

ricardoV94: And now I understand your approach and I think it was better like you did. The `__bool__` sounds fine as well!

Sorry I got confused by the names of the things

pymc/model_graph.py
ricardoV94: That `y` should be `names=(None,)` ?

pymc/model_graph.py
williambdean: dim_lengths is mapping { dname : dlen } so it does a look up
sizes will always be tuple[int, ...]

wd60622:
> I'm okay with Plate or Cluster. Why the Meta in it?

Meta would be information about the variables / plate to construct a plate label. Previously it was always " x ".join([f"{dname} ({dlen})" for ...]
Meta just provides the information to customize based on what was there before

pymc/model_graph.py
ricardoV94: Okay I see it now

ricardoV94, line 367: dim_lengths could be a symbolic variable, we don't handle that here?

ricardoV94:
I don't love the word meta, it's too abstract. `Plate.dim_names, Plate.dim_lengths, Plate.vars`?

wd60622:
> I don't love the word meta, it's too abstract. `Plate.dim_names, Plate.dim_lengths, Plate.vars`? or `Plate.var_names` if that's what we are storing

I think itd be nice to keep the names and sizes together since they are related. How about DimInfo

ricardoV94:
Is the question whether we represent a data structure that looks like (in terms of access): `((dims_names, dim_lengths), var_names)` vs `(dim_names, dim_lengths, var_names)`? Seems like a tiny detail. I have a slight preference for having it flat but up to you

ricardoV94:
This PR refreshed my mind that #6485 and #7048 exist.

To summarize: We can have variables that have entries in `named_vars_to_dims` of type `tuple[str | None, ...]`. We can also have variables that don't show up in `named_vars_to_dims` at all? Which is odd, since we already allow None to represent unknown dims, so all variables could conceivable have entries (or we would not allow `None`).

Then dims can have coords or not, but always have dim_lengths, which always work when we do the fast_eval for `dim_lengths`, so that's not a problem that shows up here. I think that doesn't matter here for us. Just mentioning in case I brought it up by mistake in my comments.

pymc/model_graph.py
ricardoV94: I'm thinking of `pm.Deterministic("x", np.zeros((3, 3)), dims=("hello", None))` and `pm.Deterministic("y", np.zeros((3, 3)), dims=(None, "hello")`. We don't want to put those in the same plate because dims can't be repeated, so they are definitely different things?

Can we add a test for that?

pymc/model_graph.py
```diff
@@ -302,164 +348,215 @@ def get_plates(self, var_names: Iterable[VarName] | None = None) -> dict[str, se
...
+
+    graph = graphviz.Digraph(name)
+    for plate in plates:
+        if plate.meta:
+            # must be preceded by 'cluster' to get a box around it
+            plate_label = create_plate_label(
+                plate.variables[0].var.name,
+                plate.meta,
+                include_size=include_shape_size,
+            )
```
ricardoV94: Should `create_plate_label` now take `plate_formatters` that among other things decides on whether to include_size?

pymc/model_graph.py
```diff
@@ -302,164 +348,215 @@ def get_plates(self, var_names: Iterable[VarName] | None = None) -> dict[str, se
...
+def make_graph(
+    name: str,
+    plates: list[Plate],
+    edges: list[tuple[VarName, VarName]],
+    formatting: str = "plain",
+    save=None,
+    figsize=None,
+    dpi=300,
+    node_formatters: NodeTypeFormatterMapping | None = None,
+    include_shape_size: bool = True,
```
ricardoV94: Better name?
```suggestion
    include_dim_lengths: bool = True,
```

williambdean: done

pymc/model_graph.py
williambdean: Yeah, I think that is fair. Where do you view that being exposed?

wd60622:
> Is the question whether we represent a data structure that looks like (in terms of access): `((dims_names, dim_lengths), var_names)` vs `(dim_names, dim_lengths, var_names)`? Seems like a tiny detail. I have a slight preference for having it flat but up to you

There is also the NodeType which is why I went for the small dataclass wrapper that contains TensorVariable and the preprocessed label. I think have a small data structure isn't the end of the world but also helps structure the problem a bit more. The user can clearly see what is part of the new data structures in my mind

pymc/model_graph.py
williambdean: Added a test. I had to wrap the data in as_tensor_variable or I'd get an error saying the data needs name attribute

pymc/model_graph.py
williambdean: defaultdict with dims_names and dim_lengths is same as what is currently happening. But there is a wrapper class around it. Personally, I find the class helpful and more user friendly. But I could be wrong

For instance, 
```python
Plate(
    DimInfo(names=("obs", "covariate"), sizes=(10, 5)), 
    variables=[
        NodeInfo(X, node_type=DATA), 
        NodeInfo(X_transform, node_type=DETERMINISTIC), 
        NodeInfo(tvp, node_type=FREE_RV),
    ]
)
```
over 
```python
(("obs", "covariate"), (10, 5), (X, X_transform, tvp), (DATA, DETERMINSTIC, FREE_RV))
```
lines up a bit better in my mind that the first two are related objects and the last two are related objects as well

ricardoV94: Sure, just thinking of how easy we make it for users to define their custom stuff. Either way seems manageable

tests/test_model_graph.py
```diff
@@ -473,3 +481,56 @@ def test_custom_node_formatting_graphviz(simple_model):
...
+
+def test_scalars_have_no_dim_info() -> None:
+    with pm.Model() as model:
+        pm.Normal("x")
+
+    graph = ModelGraph(model)
+
+    assert graph.get_plates() == [
+        Plate(
+            dim_info=None,
```
ricardoV94: I prefer your first approach first, where it's a DimInfo with empty names and sizes (which can evaluate to False as well)
```suggestion
            dim_info=None,
```

pymc/model_graph.py
```diff
@@ -53,6 +89,21 @@ class NodeType(str, Enum):
     DATA = "Data"
 
 
+@dataclass
+class NodeInfo:
+    var: TensorVariable
```
ricardoV94, line 94: Are we using the `var` info for anything, or is the name sufficient?

pymc/model_graph.py
```diff
@@ -39,6 +41,40 @@
 )
 
 
+@dataclass
+class DimInfo:
+    names: tuple[str | None]
+    sizes: tuple[int]
```
ricardoV94: size is usually the product of a shape (or equivalent the dimension lengths that make a shape)
```suggestion
    lengths: tuple[int]
```

tests/test_model_graph.py
```diff
@@ -473,3 +481,56 @@ def test_custom_node_formatting_graphviz(simple_model):
...
+def test_none_dim_in_plate() -> None:
+    coords = {
+        "obs": range(5),
+    }
+    with pm.Model(coords=coords) as model:
+        data = pt.as_tensor_variable(
+            np.ones((5, 3)),
+            name="data",
+        )
+        pm.Deterministic("C", data, dims=("obs", None))
```
ricardoV94, line 587: Can you test together with the transposed case, with dims(None, "obs") to confirm there are two plates? Use a square shape to make it specially tricky

williambdean: Sounds good. Is the non-square case still important then?

wd60622:
Need to 
- [ ] cover the {var_name}_dim{d} case still #6335. Unless the naming should changed?
- [ ] Fix the previous tests

The 6335 comes up with this example: 
```python
coords = {
    "obs": range(5),
}
    data = pt.as_tensor_variable(
        np.ones((5, 3)),
        name="data",
    )
    pm.Deterministic("C", data, dims=("obs", None))
    pm.Deterministic("D", data, dims=("obs", None))
    pm.Deterministic("E", data, dims=("obs", None))

pm.model_to_graphviz(model)
```
Result: 
previous-with-none

wd60622:
I did just catch this bug: It comes from the `make_compute_graph` which causes a self loop

```python
from pymc.model_graph import ModelGraph

coords = {
    "obs": range(5),
}
    data = pt.as_tensor_variable(
        np.ones((5, 3)),
        name="C",
    )
    pm.Deterministic("C", data, dims=("obs", None))

error_compute_graph = ModelGraph(model).make_compute_graph() # defaultdict(set, {"C": {"C"}})
# Visualize error:
pm.model_to_graphviz(model)
```
Result: 



compute-graph-bug

Shall I make a separate issue?

ricardoV94:
I think they should be in the same plate, because in the absense of dims, the shape is used to cluster RVs?

ricardoV94:
Self loop is beautiful :)

wd60622:
> I think they should be in the same plate, because in the absense of dims, the shape is used to cluster RVs?

How should the {var_name}_dim{d} be handled then to put them on the same plate?

Just dim{d}?

ricardoV94:
Just the length, how does it look like for a plate without any dims?

wd60622:
> Just the length? how does a plate without any dims look like?
> 
> I imagine the mix would be 50 x trial(30) or however the trial dim is usually displayed.
> 
> WDYT?

This mixing of dlen and "{dname} ({dlen})" is what I had in mind. That is the current behavior.

Here are some examples: 
```python
import numpy as np
import pymc as pm
import pytensor.tensor as pt

coords = {
    "obs": range(5),
}
    data = pt.as_tensor_variable(
        np.ones((5, 3)),
        name="data",
    )
    C = pm.Deterministic("C", data, dims=("obs", None))
    D = pm.Deterministic("D", data, dims=("obs", None))
    E = pm.Deterministic("E", data, dims=("obs", None))

pm.model_to_graphviz(model)
```
same-plate
```python
# Same as above
pm.model_to_graphviz(model, include_dim_lengths=False)
```
same-plate-without

And larger example with various items:

```python
import numpy as np
import pymc as pm
import pytensor.tensor as pt

coords = {
    "obs": range(5),
    "covariates": ["X1", "X2", "X3"],
}
    data1 = pt.as_tensor_variable(
        np.ones((5, 3)),
        name="data1",
    )
    data2 = pt.as_tensor_variable(
        np.ones((5, 3)),
        name="data2",
    )
    C = pm.Deterministic("C", data1, dims=("obs", None))
    CT = pm.Deterministic("CT", C.T, dims=(None, "obs"))
    D = pm.Deterministic("D", C @ CT, dims=("obs", "obs"))

    E = pm.Deterministic("E", data2, dims=("obs", None))
    beta = pm.Normal("beta", dims="covariates")
    pm.Deterministic("product", E[:, None, :] * beta[:, None], dims=("obs", None, "covariates"))

pm.model_to_graphviz(model)
```
larger-example

pymc/model_graph.py
williambdean: I exposed the create_plate_label in both `make_graph` and `make_networkx`. However, left it out in the `model_to_graphviz` and `model_to_networkx` functions.

If a user defines `Callable[[DimInfo], str]` function, then that can be used in the first two, more general functions

pymc/model_graph.py
```diff
@@ -39,6 +40,47 @@
 )
 
 
+@dataclass
+class DimInfo:
+    names: tuple[str | None, ...]
+    lengths: tuple[int, ...]
+
+    def __post_init__(self) -> None:
```
ricardoV94, line 48: TIL

pymc/model_graph.py
```diff
@@ -39,6 +40,47 @@
...
+
+
+PlateLabelFunc = Callable[[DimInfo], str]
+
+
+def create_plate_label_without_dim_length(
+    dim_info: DimInfo,
+) -> str:
+    def create_label(dname: str | None, dlen: int):
+        return f"{dname}" if dname else f"{dlen}"
```
ricardoV94: Inline this function in the generator below?

pymc/model_graph.py
```diff
@@ -156,6 +213,29 @@ def update_node_formatters(node_formatters: NodeTypeFormatterMapping) -> NodeTyp
     return node_formatters
 
 
+class AddNode(Protocol):
```
ricardoV94: Why this?

pymc/model_graph.py
```diff
@@ -39,6 +40,47 @@
...
+
+    return " x ".join(
+        create_label(dname, dlen) for (dname, dlen) in zip(dim_info.names, dim_info.lengths)
+    )
+
+
+def create_plate_label_with_dim_length(
+    dim_info: DimInfo,
+) -> str:
+    def create_label(dname: str | None, dlen: int):
```
ricardoV94: Same here, inline?

pymc/model_graph.py
williambdean: mypy to allow callable with first arg str then any amount of kwargs. 
Do you have an alternative?

ricardoV94: `# type: ignore`

pymc/model_graph.py
williambdean: nice in `pydantic.dataclasses.dataclass` as well

ricardoV94 (approved):

ricardoV94:
Thanks @wd60622

PR #7395, opened 2024-06-30
Fix typing in `logprob.tensor`

michaelosthege:
## Description

Movin' in on the all-green mypy check...

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7395.org.readthedocs.build/en/7395/

### DIFF
--- a/pymc/logprob/rewriting.py
+++ b/pymc/logprob/rewriting.py
@@ -36,7 +36,7 @@
 import warnings
 
 from collections import deque
-from collections.abc import Sequence
+from collections.abc import Collection, Sequence
 
 import pytensor.tensor as pt
 
@@ -484,7 +484,7 @@ def cleanup_ir(vars: Sequence[Variable]) -> None:
 
 
 def assume_measured_ir_outputs(
-    inputs: Sequence[TensorVariable], outputs: Sequence[TensorVariable]
+    inputs: Collection[TensorVariable], outputs: Sequence[TensorVariable]
 ) -> Sequence[TensorVariable]:
     """Run IR rewrite assuming each output is measured.
 
--- a/pymc/logprob/utils.py
+++ b/pymc/logprob/utils.py
@@ -36,7 +36,7 @@
 import typing
 import warnings
 
-from collections.abc import Container, Sequence
+from collections.abc import Container, Iterable, Sequence
 
 import numpy as np
 import pytensor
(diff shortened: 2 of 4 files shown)


pymc/logprob/rewriting.py
```diff
@@ -484,7 +484,7 @@ def cleanup_ir(vars: Sequence[Variable]) -> None:
 
 
 def assume_measured_ir_outputs(
-    inputs: Sequence[TensorVariable], outputs: Sequence[TensorVariable]
+    inputs: Collection[TensorVariable], outputs: Sequence[TensorVariable]
```
ricardoV94, line 487: Why does it prefer collection?

pymc/logprob/utils.py
```diff
@@ -173,7 +173,7 @@ def indices_from_subtensor(idx_list, indices):
 
 
 def check_potential_measurability(
-    inputs: tuple[TensorVariable], valued_rvs: Container[TensorVariable]
+    inputs: Reversible[TensorVariable], valued_rvs: Container[TensorVariable]
```
ricardoV94: Why does it prefer Reversible? I would never have come up with such a hint

pymc/logprob/rewriting.py
michaelosthege: It was called with a `dict[TensorVariable, something]`, which is a `Collection[TensorVariable]` (but not a sequence).

The code here didn't need `inputs` to be in some order.

pymc/logprob/utils.py
michaelosthege: The function was called with a `list[TensorVariable]`, which is not a `tuple`.

This is a classical case of unnecessary restrictive, invariant input type.

The `inputs` are used in a `reversed(inputs)` call, which is where I got the `Reversible` from.

ricardoV94: Isn't sequence enough?

michaelosthege: Actually, now that I look at it again, this only needs to be a `Collection[TensorVariable]`, because `inputs` is just passed to `walk`.

ricardoV94 (approved):

twiecki (approved):

PR #7408, opened 2024-07-09
Fix bug with multiple minibatch variables

ricardoV94:
## Description

Reported in https://discourse.pymc.io/t/verifying-that-minibatch-is-actually-randomly-sampling/14308

The bug occurred due to separate calls to model.logp (from model.datalogp and model.varlogp) that create distinct clones of the RandomIntegersRV underlying minibatch slicing. `compile_pymc` would not set any updates in this case

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7408.org.readthedocs.build/en/7408/

### DIFF
--- a/tests/variational/test_inference.py
+++ b/tests/variational/test_inference.py
@@ -472,3 +472,25 @@ def test_fit_data_coords(hierarchical_model, hierarchical_model_data):
             hierarchical_model_data["group_coords"].keys()
         )
         assert data["mu"].shape == tuple()
+
+
+def test_multiple_minibatch_variables():
+    """Regression test for bug reported in
+    https://discourse.pymc.io/t/verifying-that-minibatch-is-actually-randomly-sampling/14308
+    """
+    true_weights = np.array([-5, 5] * 5)
+    feature = np.repeat(np.eye(10), 10_000, axis=0)
+    y = feature @ true_weights
+
+    with pm.Model() as model:
+        minibatch_feature, minibatch_y = pm.Minibatch(feature, y, batch_size=1)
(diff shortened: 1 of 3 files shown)


tests/variational/test_inference.py
```diff
@@ -472,3 +472,25 @@ def test_fit_data_coords(hierarchical_model, hierarchical_model_data):
...
+        weights = pm.Normal("weights", 0, 10, shape=10)
+        pm.Normal(
+            "y",
+            mu=minibatch_feature @ weights,
+            sigma=0.01,
+            observed=minibatch_y,
+            total_size=len(y),
+        )
+        mean_field = pm.fit(10_000, obj_optimizer=pm.adam(learning_rate=0.01), progressbar=False)
+    np.testing.assert_allclose(mean_field.mean.get_value(), true_weights, rtol=1e-1)
```
jessegrabowski, line 496: Does this test need to run the whole model and check parameter recovery? It should be enough to compile the function and check that `minibatch_feature` and `minibatch_y` change after each loss function execution right?

jessegrabowski (approved):

tests/variational/test_inference.py
ricardoV94: This should run pretty fast, with minibatch of 1. I think it's a useful integration test, we didn't have any linear regression minibatch test in the codebase.

Besides VI has a very complex logic leading to building the function that I rather treat as a black box

PR #7233, opened 2024-03-30
Replace fastprogress progress bars with rich

fonnesbeck:
## Description

This PR aims to improve the robustness of PyMC's progress bars by implementing them with the `rich` library in place of `fastprogress`

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Type of change

- [x] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7233.org.readthedocs.build/en/7233/

PR #7410, opened 2024-07-10
Log SMC progress via Python logger when progress bar deactivated

### LINKED ISSUE #7409: BUG: SMC progress bars do not display correctly when called from RStudio
### Describe the issue:

When calling pyMC SMC sampling from RStudio with multiple chains, the new Rich progress bars fill the Console and the Rmarkdown output with `chains` lines every 0.1 seconds, eventually flooding the console and slowing down RStudio to a crawl.

This is because RStudio console output does not implement cursor movement ANSI control sequences used by Rich multi-progress bars. 

See https://github.com/rstudio/reticulate/issues/1632 and https://github.com/rstudio/rstudio/issues/14942 

```python
import pymc as pm

# Run this via a Python chunk in Rstudio with reticulate

def main():

    with pm.Model() as model:
        x = pm.Normal("x", mu=0, sigma=1)
        y = pm.Normal("y", mu=x, sigma=0.001, observed=2)

    with model:
        trace = pm.sample_smc(draws=1000, chains=12, cores=1, progressbar=False)


if __name__ == "__main__":
    main()
```

### Error message:

```shell
No error message. See the screenshot.
```

image

### PyMC version information:

<details>
- PyMC 5.16.1, installed via Conda
- Pytensor 2.23.0 (but not important)
- macOS 14.5 Catalina, and also reproduced on Windows Server 2022
- rich 13.7.1 (relevant!)

### Context for the issue:

This ruins every call to PyMC SMC sampler with progress bars from RStudio, and hence the use of SMC Sampler by R-heavy biostatisticians.

Turning off the progressbar would work, but then would give no visibility as to how the algorithm is progressing.

I'm submitting a PR in a minute to fix this :)

jucor:
## Description

Output terminals that do not support cursor movement ANSI control sequences behave very badly with Rich ProgressBars, and thus need progressbar=false.  See discussion in https://github.com/pymc-devs/pymc/pull/7233#issuecomment-2214524984

However, even in those cases it is still very worthy to know how the SMC tasks are progressing.

Therefore, in this PR, when progress bars are deactivated, any change to the state of the chains is logged to the default logger, using the exact same state dictionary as used for the progressbars. The fact we only log when a change happens means we do not flood the log.

The use of the logging facility also allows the caller to add timestamps in the logging handlers for a sense of elapsing time, or to filter out these progress updates.

## Related Issue

- [x] Closes #7409
- [x] Related to #7233

## Checklist

- [x] ~Included tests that prove the fix is effective or that the new feature works~ (no test for logging)
- [x] ~Added necessary documentation (docstrings and/or example notebooks)~ (no documentation necessary)

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7410.org.readthedocs.build/en/7410/

### DIFF
--- a/pymc/smc/sampling.py
+++ b/pymc/smc/sampling.py
@@ -345,7 +345,8 @@ def _sample_smc_int(
     while smc.beta < 1:
         smc.update_beta_and_weights()
 
-        progress_dict[task_id] = {"stage": stage, "beta": smc.beta}
+        # Index by chain because task_id is None if no progressbar is present
+        progress_dict[chain] = {"stage": stage, "beta": smc.beta, "task_id": task_id}
 
         smc.resample()
         smc.tune()
@@ -378,6 +379,7 @@ def run_chains(chains, progressbar, params, random_seed, kernel_kwargs, cores):
         disable=not progressbar,
     ) as progress:
         futures = []  # keep track of the jobs
+        _log = logging.getLogger(__name__)
         with multiprocessing.Manager() as manager:
             # this is the key - we share some state between our
             # main process and our worker functions
@@ -391,6 +393,8 @@ def run_chains(chains, progressbar, params, random_seed, kernel_kwargs, cores):
                 for c in range(chains):  # iterate over the jobs we need to run
                     # set visible false so we don't have a lot of bars all at once:
                     task_id = progress.add_task(f"Chain {c}", status="Stage: 0 Beta: 0")
+                    if not progressbar:
+                        _log.info(f"Queueing Chain {c} Stage: 0 Beta: 0")
                     futures.append(
                         executor.submit(
                             _sample_smc_int,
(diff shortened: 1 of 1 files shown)


jucor:
@fonnesbeck That might be of interest to you :)

jucor:
@zaxtax Would you please be able to review, given how you were the reviewer for #7233? Thank you :)

ricardoV94:
I'm still not convinced the change to rich progressbar was a net positive :)

aloctavodia (approved):
LGTM!

jucor:
@aloctavodia @ricardoV94 @aseyboldt Do you think this could be merged, please?

zaxtax:
That sounds reasonable!

jucor:
👋  Any help merging this please? I believe @ricardoV94 requested a review by @aseyboldt on July 12th, but I imagine Adrian is possibly swamped with less trivial PRs :) Any chance to merge this please?

ricardoV94:
Is PyMC log info showing by default? Usually progressbar is disabled when people don't want any output.

jucor:
Good point, thanks for raising it. I believe the `progressbar` argument means just that: "do you want a progressbar?", and does not say anything about any other way to log.
I've checked the codebase, and found several examples confirming this:
- In SMC initialization: https://github.com/pymc-devs/pymc/blob/28560627ad75729ac0da865584fdea2344e63a50/pymc/smc/sampling.py#L202 outputs to the logger regardless of `progressbar`.
- In the tuning phase of MCMC:  https://github.com/pymc-devs/pymc/blob/28560627ad75729ac0da865584fdea2344e63a50/pymc/tuning/starting.py#L160 outputs to the logger regardless of the value of `progressbar`.

Seems logical: besides the pretty display of a progressbar, any other information is just part of the use of the standard logging facility, and can be muted or filtered using the core library's logging module.

So if anything, we could argue that I should maybe modify this PR to log to the logger **even if** `progressbar` is `True`. However I do not think that such a "logging + progressbar" is warranted in this case,  because:
- the progressbar provides more information than the logging
- and the logging occurs during the inner loop and thus, if printed, would possibly mess up the rendering of the progressbar (I do not know how gracefully `rich` handles text that is printed by non-rich calls between two updates to its progressbar)

PR #7439, opened 2024-07-29
Enforce custom `initval` in SMC

tvwenger:
## Description
Per [the discussion here](https://discourse.pymc.io/t/modeling-a-spectrum-of-gaussians-with-smc/14854), it is apparent that SMC does not respect custom `initval`. This is a problem when a custom `initval` is required in order to support a transformation, such as `Ordered`.

## Related Issue

- [X] Closes https://github.com/pymc-devs/pymc/issues/7438

## Checklist

## Type of change

- [X] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7439.org.readthedocs.build/en/7439/


tests/smc/test_smc.py
```diff
@@ -79,7 +80,9 @@ def test_sample(self):
         initial_rng_state = np.random.get_state()
         with self.SMC_test:
             mtrace = pm.sample_smc(
-                draws=self.samples, return_inferencedata=False, progressbar=not _IS_WINDOWS
+                draws=self.samples,
+                return_inferencedata=False,
+                progressbar=not _IS_WINDOWS,
```
ricardoV94: your PR is introducing several unrelated formatting changes by adding the trailing comma after the last argument. Can you revert those?

pymc/smc/kernels.py
```diff
@@ -193,10 +193,16 @@ def initialize_population(self) -> dict[str, np.ndarray]:
             )
 
             model = self.model
+
+            # Respect custom initvals (issue #7438)
+            initval_strats = {
+                **model.rvs_to_initial_values,
```
ricardoV94: Won't this override the default behavior of sampling from the prior, to using the moment?

pymc/smc/kernels.py
```diff
@@ -193,10 +193,16 @@ def initialize_population(self) -> dict[str, np.ndarray]:
             )
 
             model = self.model
+
+            # Respect custom initvals (issue #7438)
```
ricardoV94: No need for comment of intermediate variable, just define directly in the call

tvwenger:
@ricardoV94 thanks for the review! Before I revert the formatting changes (thanks VSCode!), I want to address the comment

> Won't this override the default behavior of sampling from the prior, to using the moment?

I guess this is true, and we'll need a different strategy for this fix! If you have any ideas, please let me know.

ricardoV94:
What is model.rvs_to_initvals when those are not specified though? I'm not sure the default is moment, my comment was a question not a statement

tvwenger:
> What is model.rvs_to_initvals when those are not specified though? I'm not sure the default is moment, my comment was a question not a statement

Thanks for the clarification @ricardoV94 ! Indeed, the default value for `model.rvs_to_initvals` is `None` unless `initval` is set. Thus, sampling from the prior distribution will be respected unless `initval` is set. I've updated the documentation to make this explicit!

I've also reverted those pesky formatting changes.

aloctavodia (approved):

aloctavodia:
Thanks @tvwenger!

tvwenger:
@aloctavodia Actually, I think this "fix" causes other issues. When sampling a model with `sample_smc` and this change, I get errors like:

"""
  File "/home/twenger/miniconda3/envs/bayes_spec-dev/lib/python3.12/concurrent/futures/process.py", line 263, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/twenger/science/python/pymc/pymc/smc/sampling.py", line 352, in _sample_smc_int
    smc.mutate()
  File "/home/twenger/science/python/pymc/pymc/smc/kernels.py", line 428, in mutate
    pearson_r = corr.get(self.tempered_posterior)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/twenger/science/python/pymc/pymc/smc/kernels.py", line 466, in get
    return np.abs(ab / (self.aa * bb))
"""

Presumably, these errors are due to the fact that all of the starting positions are equal, thus there is no variance in the starting particle positions and the denominator of this `pearson_r` derivation is zero. Yikes!

@jessegrabowski [had another idea here](https://discourse.pymc.io/t/modeling-a-spectrum-of-gaussians-with-smc/14854/4?u=tvwenger) that I will try to implement...

aloctavodia:
Ohh, I see. Thanks for keep trying to find a good solution

ricardoV94:
@aloctavodia the PR should have been squashed+merge, since it had useless / badly titled commits.

PR #7443, opened 2024-08-01
Cleanup logprob module

ricardoV94:
## Description

Some non-functional cleanup of the logprob module that spinoff from other PRs

----
📚 Documentation preview 📚: https://pymc--7443.org.readthedocs.build/en/7443/

### DIFF
--- a/pymc/logprob/abstract.py
+++ b/pymc/logprob/abstract.py
@@ -35,6 +35,7 @@
 #   SOFTWARE.
 
 import abc
+import warnings
 
 from collections.abc import Sequence
 from functools import singledispatch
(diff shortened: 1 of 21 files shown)


pymc/logprob/abstract.py
```diff
@@ -131,14 +143,21 @@ def _icdf_helper(rv, value, **kwargs):
...
+MeasurableOp.register(RandomVariable)
 
-MeasurableVariable.register(RandomVariable)
 
+class MeasurableOpMixin(MeasurableOp):
+    """MeasurableOp Mixin with a distinctive string representation"""
 
-class MeasurableElemwise(Elemwise):
+    def __str__(self):
+        return f"Measurable{super().__str__()}"
```
ricardoV94, line 157: This allows seeing when a Measurable replacement was actually performed in the IR graph. Most times we use a subclass of the pre-existing Op, and it would not be distinguishable in the string representation

twiecki (approved):
Why can we get rid of the tests?

ricardoV94:
> Why can we get rid of the tests?

We removed a hacky functionality that is properly implemented by the auto-imputation machinery

PR #7463, opened 2024-08-15
Add support for polars dataframes and series

### LINKED ISSUE #7462: ENH: Replace pandas dependence/use with narwhals
### Before

_No response_

### After

_No response_

### Context for the issue:

With the rise in popularity of packages such as polars, arrow, and others PyMC's dependence on pandas is looking less and less universally useful.  [Narwhals](https://narwhals-dev.github.io/narwhals/) was developed for  developers of python libraries that consumes dataframes who wish wishing to make their libraries completely dataframe-agnostic.  So maybe we consider using that instead of pandas per se? Narwhals has no dependencies and has negligible overhead, so it seems relatively lightweight.

It will require some refactoring as it relies on (as subset of) the polars API.

fonnesbeck:
## Description

Mostly superficial changes to recognize polars data structures.

## Related Issue

- [x] Closes #7462

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7463.org.readthedocs.build/en/7463/

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -22,6 +22,7 @@ from typing import cast
 
 import numpy as np
 import pandas as pd
+import polars as pl
 import pytensor
 import pytensor.tensor as pt
 import xarray as xr
@@ -204,7 +205,7 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size: 
 
 def determine_coords(
     model,
-    value: pd.DataFrame | pd.Series | xr.DataArray,
+    value: pd.DataFrame | pd.Series | pl.DataFrame | pl.Series | xr.DataArray,
     dims: Sequence[str | None] | None = None,
     coords: dict[str, Sequence | np.ndarray] | None = None,
 ) -> tuple[dict[str, Sequence | np.ndarray], Sequence[str | None]]:
@@ -348,7 +349,7 @@ def Data(
     ----------
     name : str
         The name for this variable.
-    value : array_like or pandas.Series, pandas.Dataframe
+    value : array_like or pandas.Series, pandas.Dataframe, polars.DataFrame, polars.Series, xarray.DataArray
         A value to associate with this variable.
     dims : str or tuple of str, optional
         Dimension names of the random variables (as opposed to the shapes of these
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -21,8 +21,7 @@ from copy import copy
 from typing import cast
 
 import numpy as np
-import pandas as pd
-import polars as pl
+from numpy.typing import ArrayLike
 import pytensor
 import pytensor.tensor as pt
 import xarray as xr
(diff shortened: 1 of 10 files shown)


pymc/data.py
```diff
@@ -22,6 +22,7 @@
 
 import numpy as np
 import pandas as pd
+import polars as pl
```
ricardoV94: This should be optional dependency (pandas also but that is futile because arviz requires it)

pymc/pytensorf.py
```diff
@@ -96,12 +97,16 @@ def convert_generator_data(data) -> TensorVariable:
 
 def convert_data(data) -> np.ndarray | Variable:
     ret: np.ndarray | Variable
-    if hasattr(data, "to_numpy") and hasattr(data, "isnull"):
```
ricardoV94: Should we keep the old logic, but now with an or hasattr(data, is_null) for the second condition?

ricardoV94 (changes requested):

ricardoV94:
polars should be an optional dependency. For the dispatch can be done with a try except import

pymc/pytensorf.py
```diff
@@ -141,11 +157,18 @@ def convert_data(data) -> np.ndarray | Variable:
...
+    @_as_tensor_variable.register(pd.DataFrame)
+    @_as_tensor_variable.register(pl.DataFrame)
+    @_as_tensor_variable.register(pl.Series)
+    def dataframe_to_tensor_variable(df: pd.DataFrame | pl.DataFrame, *args, **kwargs) -> TensorVariable:
+        return pt.as_tensor_variable(df.to_numpy(), *args, **kwargs)
+else:
+    @_as_tensor_variable.register(pd.Series)
+    @_as_tensor_variable.register(pd.DataFrame)
+    def dataframe_to_tensor_variable(df: pd.DataFrame, *args, **kwargs) -> TensorVariable:
+        return pt.as_tensor_variable(df.to_numpy(), *args, **kwargs)
```
ricardoV94, line 171: This is more succinct. Also type hint of df was wrong, so I just removed it.

```suggestion
@_as_tensor_variable.register(pd.Series)
@_as_tensor_variable.register(pd.DataFrame)
def dataframe_to_tensor_variable(df, *args, **kwargs) -> TensorVariable:
    return pt.as_tensor_variable(df.to_numpy(), *args, **kwargs)

if pl is not None:
    @_as_tensor_variable.register(pl.DataFrame)
    @_as_tensor_variable.register(pl.Series)
    def polars_dataframe_to_tensor_variable(df, *args, **kwargs) -> TensorVariable:
        return pt.as_tensor_variable(df.to_numpy(), *args, **kwargs)

```

pymc/pytensorf.py
```diff
@@ -111,6 +115,18 @@ def convert_data(data) -> np.ndarray | Variable:
             ret = np.ma.MaskedArray(vals, mask)
         else:
             ret = vals
+    elif hasattr(data, "to_numpy") and hasattr(data, "is_null"):
```
ricardoV94, line 118:
```suggestion
    elif hasattr(data, "to_numpy") and hasattr(data, "is_null"):
        # Probably polars object
```

lorentzenchr: Why not a bit more explicit:
```suggestion
    elif pl is not None and isinstance(data, (pl.DataFrame, pl.Series)):
```
The polars namespace is used anyway (in the except clause).

PR #7473, opened 2024-08-23
Simplify Model __new__ and metaclass

thomasaarholt:
## Description
This PR is a refactor that improves the code that creates models and keeps track of the model contexts that are active at a given moment.

I have previously done some work (#6809) on type hinting the Model class. It was a bit difficult to understand what was going on, and the code was lacking documentation that would help further development. In particular, the code creating the Model class and instance, and keeping track of active contexts was rather hacky.

I am on parental leave, which (maybe ironically) gives me the time to take a look behind the scenes and hopefully make some improvements that give back to the community. I do recognize that this PR comes a bit "unannounced", and since it doesn't fix any major issues or introduce hot features, I don't have any expectations for it to be reviewed soon (or even at all).

Each commit is a logical change with a reasonably descriptive text. 

This passes all tests in `tests/model`. 

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7473.org.readthedocs.build/en/7473/

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -221,9 +221,9 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size: 
 def determine_coords(
     model,
     value: pd.DataFrame | pd.Series | xr.DataArray,
-    dims: Sequence[str | None] | None = None,
+    dims: Sequence[str] | None = None,
     coords: dict[str, Sequence | np.ndarray] | None = None,
-) -> tuple[dict[str, Sequence | np.ndarray], Sequence[str | None]]:
+) -> tuple[dict[str, Sequence | np.ndarray], Sequence[str] | Sequence[None]]:
     """Determine coordinate values from data or the model (via ``dims``)."""
     if coords is None:
         coords = {}
@@ -268,9 +268,10 @@ def determine_coords(
 
     if dims is None:
         # TODO: Also determine dim names from the index
-        dims = [None] * np.ndim(value)
-
-    return coords, dims
+        new_dims: Sequence[str] | Sequence[None] = [None] * np.ndim(value)
+    else:
+        new_dims = dims
+    return coords, new_dims
 
 
 def ConstantData(
@@ -366,7 +367,7 @@ def Data(
         The name for this variable.
     value : array_like or pandas.Series, pandas.Dataframe
         A value to associate with this variable.
-    dims : str or tuple of str, optional
+    dims : str, tuple of str or tuple of None, optional
         Dimension names of the random variables (as opposed to the shapes of these
         random variables). Use this when ``value`` is a pandas Series or DataFrame. The
         ``dims`` will then be the name of the Series / DataFrame's columns. See ArviZ
@@ -451,13 +452,16 @@ def Data(
             expected=x.ndim,
         )
 
+    new_dims: Sequence[str] | Sequence[None] | None
     if infer_dims_and_coords:
-        coords, dims = determine_coords(model, value, dims)
+        coords, new_dims = determine_coords(model, value, dims)
+    else:
+        new_dims = dims
 
-    if dims:
+    if new_dims:
         xshape = x.shape
         # Register new dimension lengths
-        for d, dname in enumerate(dims):
+        for d, dname in enumerate(new_dims):
             if dname not in model.dim_lengths:
                 model.add_coord(
                     name=dname,
@@ -467,6 +471,6 @@ def Data(
                     length=xshape[d],
                 )
 
-    model.register_data_var(x, dims=dims)
+    model.register_data_var(x, dims=new_dims)
 
     return x
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -462,7 +462,7 @@ def Data(
         xshape = x.shape
         # Register new dimension lengths
         for d, dname in enumerate(new_dims):
-            if dname not in model.dim_lengths:
+            if dname not in model.dim_lengths and dname is not None:
                 model.add_coord(
                     name=dname,
                     # Note: Coordinate values can't be taken from
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -24,7 +24,6 @@ from typing import (
     TYPE_CHECKING,
     Literal,
     Optional,
-    TypeVar,
     Union,
     cast,
     overload,
@@ -92,9 +91,6 @@ __all__ = [
 ]
 
 
-T = TypeVar("T", bound="ContextMeta")
-
-
 class ContextMeta(type):
     """Functionality for objects that put themselves in a context manager."""
 
@@ -125,13 +121,13 @@ class ContextMeta(type):
             cls._context_class = context_class
         super().__init__(name, bases, nmspc)
 
-    def get_context(cls, error_if_none=True, allow_block_model_access=False) -> T | None:
+    def get_context(cls, error_if_none=True, allow_block_model_access=False) -> "Model | None":
         """Return the most recently pushed context object of type ``cls`` on the stack, or ``None``.
 
         If ``error_if_none`` is True (default), raise a ``TypeError`` instead of returning ``None``.
         """
         try:
-            candidate: T | None = cls.get_contexts()[-1]
+            candidate: Model | None = cls.get_contexts()[-1]
         except IndexError:
             # Calling code expects to get a TypeError if the entity
             # is unfound, and there's too much to fix.
@@ -142,7 +138,7 @@ class ContextMeta(type):
             raise BlockModelAccessError(candidate.error_msg_on_access)
         return candidate
 
-    def get_contexts(cls) -> list[T]:
+    def get_contexts(cls) -> list["Model"]:
         """Return a stack of context instances for the ``context_class`` of ``cls``."""
         # This lazily creates the context class's contexts
         # thread-local object, as needed. This seems inelegant to me,
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -129,29 +129,6 @@ class ContextMeta(type):
             cls._context_class = context_class
         super().__init__(name, bases, nmspc)
 
-    def get_contexts(cls) -> list[Model]:
-        """Return a stack of context instances for the ``context_class`` of ``cls``."""
-        # This lazily creates the context class's contexts
-        # thread-local object, as needed. This seems inelegant to me,
-        # but since the context class is not guaranteed to exist when
-        # the metaclass is being instantiated, I couldn't figure out a
-        # better way. [2019/10/11:rpg]
-
-        # no race-condition here, contexts is a thread-local object
-        # be sure not to override contexts in a subclass however!
-        context_class = cls.context_class
-        assert isinstance(
-            context_class, type
-        ), f"Name of context class, {context_class} was not resolvable to a class"
-        if not hasattr(context_class, "contexts"):
-            context_class.contexts = threading.local()
-
-        contexts = context_class.contexts
-
-        if not hasattr(contexts, "stack"):
-            contexts.stack = []
-        return contexts.stack
-
     # the following complex property accessor is necessary because the
     # context_class may not have been created at the point it is
     # specified, so the context_class may be a class *name* rather
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -2003,11 +2003,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
         )
 
 
-# this is really disgusting, but it breaks a self-loop: I can't pass Model
-# itself as context class init arg.
-Model._context_class = Model
-
-
 class BlockModelAccess(Model):
     """Can be used to prevent user access to Model contexts."""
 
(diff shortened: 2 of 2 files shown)


thomasaarholt:
The mypy failure is because my code change allows mypy to infer the type of [`model` in `pymc/data.py:446`](https://github.com/thomasaarholt/pymc/blob/d203e0f1b117a17802235035ac04372658e8b7b5/pymc/data.py#L446) as `Model` rather than `Unknown`, which in turns reveals that `values` isn't of type `Unknown` on line 450. So I've just exposed something that was wasn't typed correctly.

The test failures seem strange to me. I am tempted to attribute them to external factors, like an update to pytensor.

twiecki:
Thanks @thomasaarholt, there are still mypy failures, do you have a handle on them or would you like some help?

thomasaarholt:
@twiecki Thanks! The remaining one is one that I'm not sure on how best to handle:
From the mypy github action:
```
pymc/data.py:439: error: Incompatible types in assignment (expression has type "Sequence[str | None]", variable has type "Sequence[str] | None")
```
This stems from the `determine_coords` function having a return signature for `dim` that is `Sequence[str | None]`. But the `dim` type hint from higher up in the `def Data` constructor only allows a sequence of `Sequence[str]`. The sequence can be of None inner type [here](https://github.com/pymc-devs/pymc/blob/d596afbe266ae1bdafec6735d73a82885e06dea1/pymc/data.py#L255).

Help much appreciated!

thomasaarholt:
There! @twiecki wanna take a look? I went through the logic related to the above failing mypy error and think I found a good resolution. See commit description of the last two commits.

The failing test is unrelated.

twiecki:
Any idea about that failing test?
```
================================== FAILURES ===================================
_________________ TestMarginalVsLatent.testLatentMultioutput __________________

self = <tests.gp.test_gp.TestMarginalVsLatent object at 0x0000029EDDB84820>

    def testLatentMultioutput(self):
        n_outputs = 2
        X = np.random.randn(20, 3)
        y = np.random.randn(n_outputs, 20)
        Xnew = np.random.randn(30, 3)
        pnew = np.random.randn(n_outputs, 30)
    
        with pm.Model() as latent_model:
            cov_func = pm.gp.cov.ExpQuad(3, [0.1, 0.2, 0.3])
            mean_func = pm.gp.mean.Constant(0.5)
            latent_gp = pm.gp.Latent(mean_func=mean_func, cov_func=cov_func)
            latent_f = latent_gp.prior("f", X, n_outputs=n_outputs, reparameterize=True)
            latent_p = latent_gp.conditional("p", Xnew)
    
        with pm.Model() as marginal_model:
            cov_func = pm.gp.cov.ExpQuad(3, [0.1, 0.2, 0.3])
            mean_func = pm.gp.mean.Constant(0.5)
            marginal_gp = pm.gp.Marginal(mean_func=mean_func, cov_func=cov_func)
            marginal_f = marginal_gp.marginal_likelihood("f", X, y, sigma=0.0)
            marginal_p = marginal_gp.conditional("p", Xnew)
    
        assert tuple(latent_f.shape.eval()) == tuple(marginal_f.shape.eval()) == y.shape
        assert tuple(latent_p.shape.eval()) == tuple(marginal_p.shape.eval()) == pnew.shape
    
        chol = np.linalg.cholesky(cov_func(X).eval())
        v = np.linalg.solve(chol, (y - 0.5).T)
        A = np.linalg.solve(chol, cov_func(X, Xnew).eval()).T
        mu_cond = mean_func(Xnew).eval() + (A @ v).T
        cov_cond = cov_func(Xnew, Xnew).eval() - A @ A.T
    
        with pm.Model() as numpy_model:
            numpy_p = pm.MvNormal.dist(mu=pt.as_tensor(mu_cond), cov=pt.as_tensor(cov_cond))
    
        latent_rv_logp = pm.logp(latent_p, pnew)
        marginal_rv_logp = pm.logp(marginal_p, pnew)
        numpy_rv_logp = pm.logp(numpy_p, pnew)
    
        assert (
            latent_rv_logp.shape.eval()
            == marginal_rv_logp.shape.eval()
            == numpy_rv_logp.shape.eval()
        )
    
>       npt.assert_allclose(latent_rv_logp.eval(), marginal_rv_logp.eval(), atol=5)

tests\gp\test_gp.py:412: 

args = (<function assert_allclose.<locals>.compare at 0x0000029EE75B9AB0>, array([-41.92778875, -45.52605[201](https://github.com/pymc-devs/pymc/actions/runs/11177316239/job/31072575657?pr=7473#step:7:202)]), array([-40.07459855, -65.84149075]))
kwds = {'equal_nan': True, 'err_msg': '', 'header': 'Not equal to tolerance rtol=1e-07, atol=5', 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=5
E           
E           Mismatched elements: 1 / 2 (50%)
E           Max absolute difference: 20.31543874
E           Max relative difference: 0.30855071
E            x: array([-41.927789, -45.526052])
E            y: array([-40.074599, -65.841491])

C:\Miniconda3\envs\pymc-test\lib\contextlib.py:79: AssertionError
```

thomasaarholt:
I assume that if we reran it (maybe you can trigger that?) it would pass. Seeing as it fails only on windows and not Linux/mac, I assumed it was a result of variance due to random sampling.

ricardoV94:
I restarted it

thomasaarholt:
Ran passing everything. Ready for review!

pymc/model/core.py
```diff
@@ -1243,16 +1170,16 @@ def set_data(
 
     def register_rv(
         self,
-        rv_var,
-        name,
+        rv_var: TensorVariable,
```
ricardoV94: Technically, RVs only need to be Variable, they need not be TensorVariable
```suggestion
        rv_var: Variable,
```

ricardoV94:
@thomasaarholt the PR looks good, but since it touches on such as fundamental functionality I've asked for further reviews.

You have some conflicts that need to be resolved as well.

Thanks for the work so far!

pymc/data.py
```diff
@@ -218,9 +218,9 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size:
 def determine_coords(
     model,
     value: pd.DataFrame | pd.Series | xr.DataArray,
-    dims: Sequence[str | None] | None = None,
```
lucianopaz, line 224: Did this ever change? I thought that some variables could have `dims=[None, "some_dim_name"]`. If this is still the case, then the type hint needs to left as it was before.

pymc/data.py
```diff
@@ -218,9 +218,9 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size:
...
+    dims: Sequence[str] | None = None,
     coords: dict[str, Sequence | np.ndarray] | None = None,
-) -> tuple[dict[str, Sequence | np.ndarray], Sequence[str | None]]:
```
lucianopaz, line 226: Same as above. If variables can still have `dims=[None, "some_dim_name"]` then the old type hint is correct and the new one is wrong.

pymc/model/core.py
```diff
@@ -577,6 +495,17 @@ def __init__(
...
+    def get_context(
+        cls, error_if_none: bool = True, allow_block_model_access: bool = False
+    ) -> Model | None:
+        model = MODEL_MANAGER.current_context
+        if isinstance(model, BlockModelAccess) and not allow_block_model_access:
+            raise BlockModelAccessError(model.error_msg_on_access)
+        if model is None and error_if_none:
+            raise TypeError("No model on context stack")
+        return model
+
```
lucianopaz, line 508: I'm a bit at a loss here. Why do we even need this class method? Can't it be replaced by a simple call to `modelcontext(None)`?

pymc/model/core.py
```diff
@@ -92,118 +88,36 @@
...
-        with instance:  # appends context
-            instance.__init__(*args, **kwargs)
-        return instance
-
-
-def modelcontext(model: Optional["Model"]) -> "Model":
+def modelcontext(model: Model | None) -> Model:
     """Return the given model or, if None was supplied, try to find one in the context stack."""
     if model is None:
         model = Model.get_context(error_if_none=False)
```
lucianopaz, line 123: Can't this be replaced by `MODEL_MANAGER.current_context`? That way, you could drop the kind of useless `get_context` class method.

lucianopaz (commented):
Nice work @thomasaarholt. The original code was very hard to read, and yours is much better now. I think that your refactor could be extended to a few other places. I left comments in the relevant parts.

pymc/data.py
thomasaarholt: I think you're right. I guess someone could pass a data cube with some "batch" or otherwise uninteresting dimension? I don't really see why you would ever do that, but I'm not gonna stop them. I'll change it back.

pymc/model/core.py
thomasaarholt: I think I was trying to keep things as close to the old code as possible. I'll try to get rid of it as you suggest.

pymc/data.py
thomasaarholt: Actually, do you mind giving an example? I observe that in main branch, both `pm.Data` and `pm.ConstantData` take `dims: Sequence[str] | None`. Are these incorrect and `Sequence[str]` should be `Sequence[str | None]`?

pymc/model/core.py
thomasaarholt: Can you expand on what you are suggesting? 

In main branch, `modelcontext` always returns a model (or raises an error). I can modify `modelcontext` to take `error_if_none: bool = True, allow_block_model_access: bool = False` as arguments and include the logic you highlight above, as well as change the return type from `Model` to `Model | None`. 

Then I'll need to add `@overload`s for those two arguments to maintain correct typing, since there are over 50 locations `modelcontext` is used. I'm happy to do this, but I want to be sure this is what you're intending.

thomasaarholt: These are relevant instances/test where get_context are currently used:

https://github.com/pymc-devs/pymc/blob/938aff4edb0d92fdb0b5d690a1f9d0ebc488e7e7/tests/model/test_core.py#L1212

https://github.com/pymc-devs/pymc/blob/938aff4edb0d92fdb0b5d690a1f9d0ebc488e7e7/pymc/model/core.py#L2091

pymc/data.py
ricardoV94: I don't think we allow that anymore

pymc/data.py
ricardoV94: I don't think we allow mixed None / defined dims anymore

pymc/model/core.py
thomasaarholt: Done!

thomasaarholt:
@lucianopaz you tagged the wrong person ;)

Please see my reply about the `get_context` classmethod. Happy to get rid of more stuff (there are a few things more that are unused in model/core.py), but I'll prefer to do that in an other PR.

pymc/model/core.py
thomasaarholt: Actually, I did make this change, but that threw an error due to the internals of `get_context` not applying (yet) to `modelcontext`. Let's resolve the classmethod feedback first, and I'll get back to this.

pymc/data.py
lucianopaz: I know that it used to be possible in the past, but at some point @michaelosthege and @ricardoV94 discussed about simplifying the code base and removing part of this flexibility. I don't recall the details, but with what Ricardo commented, I think it's safe to assume that the flexibility here is gone and that your type hints are correct.

pymc/data.py
lucianopaz: Based on [this](https://github.com/pymc-devs/pymc/pull/7473#discussion_r1793425912), the new type hint that you suggest is correct.

pymc/model/core.py
lucianopaz: Ok, I see now that the `allow_block_model_access` is what's preventing us from using a `modelcontext` call. I think that it's ok to keep this class method then.

lucianopaz:
> @lucianopaz you tagged the wrong person ;)

Oops, how embarrassing! Sorry about that 😬
 
> Please see my reply about the `get_context` classmethod. Happy to get rid of more stuff (there are a few things more that are unused in model/core.py), but I'll prefer to do that in an other PR.

Everything looks fine now. Thanks for the great work!

lucianopaz (approved):

lucianopaz:
@ricardoV94, I remember seeing the same test error on one of your earlier PRs. Was it fixed now? @thomasaarholt, maybe if you rebase on top of the latest main, the test will pass.

pymc/model/core.py
```diff
@@ -1233,16 +1162,16 @@ def set_data(
 
     def register_rv(
         self,
-        rv_var,
-        name,
+        rv_var: RandomVariable,
```
ricardoV94, line 1165: This is wrong, RandomVariable is not a variable type, it's an Op type

thomasaarholt: Hello there 😊

PR #7480, opened 2024-08-27
Allow Minibatch of derived RVs and deprecate generators as data

ricardoV94:
## Description

This PR fixes issues related to Minibatch indexing reported in https://discourse.pymc.io/t/warning-using-minibatch-and-censored-together-rng-variable-has-shared-clients/14943 and extends the MinibatchRV functionality for derived RVs.

Minibatch value variables are uniquely tricky because they are random graphs, that can share RNG with other variables in the forward / logp graph. As such we need to make sure they are not mutated for the default updates to work. We tried some tricks in the past but as revealed in the discourse issue that was not enough. This PR solves the problem by encapsulating the random graph in an OpFromGraph so that the inner graph will not be touched by PyMC logprob derivation routines. It will still be inlined in the final compiled functions to avoid overhead.

I also decided to deprecate Generators as data, which showed up in some of the refactoring. The GeneratorOp is not a true Op, which should not have any side-effects. It is also not compatible with non default backends like Numba and JAX that we are moving towards to. If needed, the logic should be handled by the sampler by consuming the generator and setting the values before subsequent function calls.

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -26,18 +26,19 @@ import pytensor
 import pytensor.tensor as pt
 import xarray as xr
 
+from pytensor.compile.builders import OpFromGraph
 from pytensor.compile.sharedvalue import SharedVariable
+from pytensor.graph.basic import Variable
 from pytensor.raise_op import Assert
 from pytensor.scalar import Cast
 from pytensor.tensor.elemwise import Elemwise
 from pytensor.tensor.random.basic import IntegersRV
-from pytensor.tensor.subtensor import AdvancedSubtensor
 from pytensor.tensor.type import TensorType
 from pytensor.tensor.variable import TensorConstant, TensorVariable
 
 import pymc as pm
 
-from pymc.pytensorf import convert_data, smarttypeX
+from pymc.pytensorf import GeneratorOp, convert_data, smarttypeX
 from pymc.vartypes import isgenerator
 
 __all__ = [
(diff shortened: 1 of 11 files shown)


tests/test_data.py
```diff
@@ -593,44 +592,34 @@ class TestMinibatch:
...
         A = np.arange(1000)
-        B = np.arange(1000)
+        B = -np.arange(1000)
         mA, mB = pm.Minibatch(A, B, batch_size=10)
 
         [draw_mA, draw_mB] = pm.draw([mA, mB])
         assert draw_mA.shape == (10,)
-        np.testing.assert_allclose(draw_mA, draw_mB)
-
-        # Check invalid dims
```
ricardoV94, line 627: This was already checked in the test above

pymc/data.py
```diff
@@ -188,18 +190,29 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size:
...
+        upper = Assert(
+            "All variables shape[0] in Minibatch should be equal, check your Minibatch(data1, data2, ...) code"
+        )(upper, pt.all([pt.eq(upper, other_tensor.shape[0]) for other_tensor in tensors[1:]]))
+
+    rng = pytensor.shared(np.random.default_rng())
+    rng_update, mb_indices = minibatch_index(0, upper, size=batch_size, rng=rng).owner.outputs
+    mb_tensors = [tensor[mb_indices] for tensor in tensors]
+
+    # Wrap graph in OFG so it's easily identifiable and not rewritten accidentally
+    *mb_tensors, _ = MinibatchOp([*tensors, rng], [*mb_tensors, rng_update])(*tensors, rng)
```
ferrine, line 211: nice trick, did not know that

pymc/pytensorf.py
```diff
@@ -666,6 +672,9 @@ class GeneratorOp(Op):
     __props__ = ("generator",)
 
     def __init__(self, gen, default=None):
+        warnings.warn(
```
ferrine, line 675: lgtm

tests/test_data.py
ferrine: ok

tests/variational/test_inference.py
```diff
@@ -162,22 +159,7 @@ def fit_kwargs(inference, use_minibatch):
 
 
 def test_fit_oo(inference, fit_kwargs, simple_model_data):
-    # Minibatch data can't be extracted into the `observed_data` group in the final InferenceData
```
ferrine, line 165: no more issues there?

tests/variational/test_minibatch_rv.py
```diff
@@ -165,10 +166,7 @@ def test_minibatch_parameter_and_value(self):
         with pm.Model(check_bounds=False) as m:
             AD = pm.Data("AD", np.arange(total_size, dtype="float64"))
             TD = pm.Data("TD", np.arange(total_size, dtype="float64"))
-
-            minibatch_idx = minibatch_index(0, 10, size=(9,))
-            AD_mt = AD[minibatch_idx]
-            TD_mt = TD[minibatch_idx]
+            AD_mt, TD_mt = Minibatch(AD, TD, batch_size=9)
```
ferrine, line 169: thisis much cleaner

ferrine (approved):

ferrine:
I've created an issue to continue this work and improve scalability of minibatches https://github.com/pymc-devs/pymc/issues/7496

tests/variational/test_inference.py
ricardoV94: Nope I allow to extract the data in the idata, it extracts the whole data

PR #7482, opened 2024-08-28
Add ability to set mode in `check_start_vals`

### LINKED ISSUE #7481: ENH: Add a way to specify `mode` for `check_start_vals`
### Before

_No response_

### After

```python
with pm.Model() as m:
    ...
m.check_start_vals(..., mode="JAX")
```

### Context for the issue:

Sometimes we deploy `pymc` in environments were we actually only care about GPU sampling via JAX, so there shouldn't be a need for a C compiler. This is all fine and good because we can `sample_prior_predictive`, `sample` and `sample_posterior_predictive` using JAX or any other mode that doesn't need C. The problem comes when `sample` calls `model.check_start_vals`, this will use the default mode and in the absence of a compiler, the python code might take very long to run. It would be great to have a way to set the compilation mode for `check_start_vals` as well.

lucianopaz:
## Description
This just adds a some piping of keyword arguments from `model.check_start_vals` in order to be able to pass a mode to `compile_pymc`.

## Related Issue

- [X] Closes #7481

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7482.org.readthedocs.build/en/7482/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1747,7 +1747,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
         )
         return {name: tuple(shape) for name, shape in zip(names, f())}
 
-    def check_start_vals(self, start):
+    def check_start_vals(self, start, **kwargs):
         r"""Check that the starting values for MCMC do not cause the relevant log probability
         to evaluate to something invalid (e.g. Inf or NaN)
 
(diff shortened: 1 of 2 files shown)


ricardoV94 (approved):

ricardoV94:
I changed the label to maintenance to de-emphasize in the release notes. Feel free to revert if you disagree

PR #7484, opened 2024-08-29
Fix some type hinting to help with migrating `Distribution`

thomasaarholt:
## Description
This is a small PR. I've added and fixed type hinting to a few lines of code. The idea is to make it easier to migrate the `Distribution` class to a function-based approach as discussed in the related issue.

## Related Issue
- [x] Related to #6083

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7484.org.readthedocs.build/en/7484/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -55,10 +55,10 @@ from pytensor.tensor.random.basic import (
 )
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.utils import normalize_size_param
-from pytensor.tensor.variable import TensorConstant
+from pytensor.tensor.variable import TensorConstant, TensorVariable
 
 from pymc.logprob.abstract import _logprob_helper
-from pymc.logprob.basic import icdf
+from pymc.logprob.basic import TensorLike, icdf
 from pymc.pytensorf import normalize_rng_param
 
 try:
@@ -214,7 +214,9 @@ def assert_negative_support(var, label, distname, value=-1e-6):
     return Assert(msg)(var, pt.all(pt.ge(var, 0.0)))
 
 
-def get_tau_sigma(tau=None, sigma=None):
+def get_tau_sigma(
+    tau: TensorLike | None = None, sigma: TensorLike | None = None
+) -> tuple[TensorVariable, TensorVariable]:
     r"""
     Find precision and standard deviation. The link between the two
     parameterizations is given by the inverse relationship:
@@ -241,13 +243,14 @@ def get_tau_sigma(tau=None, sigma=None):
         sigma = pt.as_tensor_variable(1.0)
         tau = pt.as_tensor_variable(1.0)
     elif tau is None:
+        assert sigma is not None  # Just for type checker
         sigma = pt.as_tensor_variable(sigma)
         # Keep tau negative, if sigma was negative, so that it will
         # fail when used
         tau = (sigma**-2.0) * pt.sign(sigma)
     else:
         tau = pt.as_tensor_variable(tau)
-        # Keep tau negative, if sigma was negative, so that it will
+        # Keep sigma negative, if tau was negative, so that it will
         # fail when used
         sigma = pt.abs(tau) ** -0.5 * pt.sign(tau)
 
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -152,7 +152,7 @@ class BoundedContinuous(Continuous):
     """Base class for bounded continuous distributions"""
 
     # Indices of the arguments that define the lower and upper bounds of the distribution
-    bound_args_indices: list[int] | None = None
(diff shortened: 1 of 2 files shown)


ricardoV94 (approved):

ricardoV94:
mypy is not happy with the changes? https://github.com/pymc-devs/pymc/actions/runs/10613016885/job/29415940076?pr=7484

thomasaarholt:
**Trying a solution, see below.**

Yes. That's [this commit](https://github.com/pymc-devs/pymc/pull/7484/commits/505effa5dc110d4f1cfc77136f3be59b7126a4dd), where I fixed the incorrect type hint syntax. 

All three reported errors are the same:
```
Incompatible types in assignment 
(expression has type "Callable[[Any, VarArg(Any), DefaultNamedArg(Any, 'size')], Any]", 
base class "Distribution" defined the type as "RandomVariable | SymbolicRandomVariable")
```

and are due to `Mixture`, `Censored` and `Discrete` distributions having `rv_op = <a method>`, while e.g. `Normal` has `rv_op = <a class>`. 

```python
class Censored(Distribution):
    rv_type = CensoredRV
    rv_op = CensoredRV.rv_op # <-- a method

class Normal(Distribution):
    rv_op = NormalRV() # <-- an instance of NormalRV, which is a subclass of RandomVariable
```

**Solution**
Typing the above, I realize that since we are really wanting to type this *by behaviour* rather than *state*. So I've tried typing it as Callable[..., TensorVariable] instead, which is how it always behaves **(I think)**.

ricardoV94:
sounds correct

thomasaarholt:
I am not going to worry about the error in `func_utils.py`. It's in `find_constrained_prior`, which is a) deprecated and b) not typed correctly (first argument should be `type[Distribution]`).

I am currently testing if I can forego the `= None` on the `rv_op` and `rv_type` completely. I don't think these are necessary.

ricardoV94:
> I am not going to worry about the error in `func_utils.py`. It's in `find_constrained_prior`, which is a) deprecated and b) not typed correctly (first argument should be `type[Distribution]`).
> 
> I am currently testing if I can forego the `= None` on the `rv_op` and `rv_type` completely. I don't think these are necessary.
> 

Then can you remove the type hints, we can't merge with a known failing CI or it will be a source of confusion for everyone else

thomasaarholt:
I can remove the type hint for class attribute `rv_op`, but then I either need to leave it `= None` (which the passing tests just established is unnecessary), replace it (with `Any`, probably, and a comment explaining what I've learnt about it), or remove the entire attribute, since the following is not legal Python:

```python
class Foo:
    bar
```

thomasaarholt:
@ricardoV94 mind taking a look?

pymc/distributions/distribution.py
```diff
@@ -122,6 +122,8 @@ def __new__(cls, name, bases, clsdict):
                     )
 
             class_change_dist_size = clsdict.get("change_dist_size")
+            if class_change_dist_size is not None:
+                raise ValueError("HAHAHA")
```
ricardoV94: Fun debug statement left :)
```suggestion
```

ricardoV94 (changes requested):

pymc/distributions/continuous.py
```diff
@@ -152,7 +152,7 @@ class BoundedContinuous(Continuous):
     """Base class for bounded continuous distributions"""
 
     # Indices of the arguments that define the lower and upper bounds of the distribution
-    bound_args_indices: list[int] | None = None
+    bound_args_indices: tuple[int, int] | None = None
```
ricardoV94: Aren't there distributions with only one of the bounds (lower, upper)?

ricardoV94 (approved):

pymc/distributions/continuous.py
thomasaarholt: You're right, [there are](https://github.com/thomasaarholt/pymc/blob/157c29a60edbe670e245b1208e3e97749669da8d/pymc/distributions/continuous.py#L2084)! I was wondering why mypy didn't flag it, but of course it was originally an error as `list[int]` doesn't allow `(3, None)`! Fixing this

pymc/distributions/distribution.py
thomasaarholt: Hehe! That's what I get for not adding my changes line by line for this particular commit!

thomasaarholt:
The failing tests stem from [this commit](https://github.com/pymc-devs/pymc/commit/465d8ac1675f65f28051423bd7fd84dc5cd2a715), not this PR.

ricardoV94:
@thomasaarholt should be fixed by #7529

twiecki (approved):

ricardoV94:
Thanks @thomasaarholt

Issue #6909, opened 2023-09-13
Scan logprob fails when unvalued stochastic outputs are returned


ricardoV94:
### Description

Found by @lucianopaz 

```python
import numpy as np
import pymc as pm
import pytensor
import pytensor.tensor as pt

from pymc.pytensorf import collect_default_updates

steps = 4

def ar_dist1(rho, sigma, size):
    def ar_step(x_tm1, rho, sigma):
        eps_t = pm.Normal.dist(sigma=sigma)
        mu = x_tm1 * rho
        x = mu + eps_t
        return x, collect_default_updates([x])

    ar_innov, _ = pytensor.scan(
        fn=ar_step,
        outputs_info=[{"initial": pt.zeros(()), "taps": [-1]}],
        non_sequences=[rho, sigma],
        n_steps=steps,
        strict=True,
    )

    return ar_innov


def ar_dist2(rho, sigma, size):
    def ar_step(x_tm1, rho, sigma):
        eps_t = pm.Normal.dist(sigma=sigma)
        mu = x_tm1 * rho
        x = mu + eps_t
        return [x, eps_t], collect_default_updates([x])

    [ar_innov, _], _ = pytensor.scan(
        fn=ar_step,
        outputs_info=[{"initial": pt.zeros(()), "taps": [-1]}, None],
        non_sequences=[rho, sigma],
        n_steps=steps,
        strict=True,
    )

    return ar_innov


with pm.Model() as m:
    rho = 0.1
    sigma = 0.1
    observed = np.arange(steps)

    pm.CustomDist(
        "ar_dist1",
        rho,
        sigma,
        dist=ar_dist1,
        observed=observed,
    )

    pm.CustomDist(
        "ar_dist2",
        rho,
        sigma,
        dist=ar_dist2,
        observed=observed,
    )

logp1, logp2 = m.compile_logp(sum=False)({})
np.testing.assert_allclose(logp1, logp2)
"""
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0
Mismatched elements: 2 / 4 (50%)
Max absolute difference: 58.
Max relative difference: 0.12928641
 x: array([   1.383647,  -48.616353, -179.116353, -390.616353])
 y: array([   1.383647,  -48.616353, -198.616353, -448.616353])
"""
```

PR #7491, opened 2024-09-04
Introduce value variables in logprob IR

### LINKED ISSUE #691: Specify mean for GaussianRandomWalk innovations
Currently there is no way to specify mean for GaussianRandomWalk innovations (just hard coded to zero in the code).

### LINKED ISSUE #6917: Relationship between valued RVs lost during logp inference

ricardoV94:
## Description

This supersedes stale #6918, it introduces ValuedRV nodes in the IR so rewrites can transparently reason about the conditioning points. The main purpose is to simplify the IR rewrite logic.

It also prevents default PyTensor rewrites from breaking dependency on valued RVs (which was behind the bug in #6917)

It also fixes some limitations in derived Scans, and makes it more strict. For instance, #6909 now fails explicitly instead of returning wrong result silently.

## Related Issue

- [x] Closes #6917

----
📚 Documentation preview 📚: https://pymc--7491.org.readthedocs.build/en/7491/

### DIFF
--- a/pymc/logprob/abstract.py
+++ b/pymc/logprob/abstract.py
@@ -53,7 +53,7 @@ def __getattr__(name):
             f"{name} has been deprecated in favor of MeasurableOp. Importing will fail in a future release.",
             FutureWarning,
         )
-        return MeasurableOpMixin
+        return MeasurableOp
 
     raise AttributeError(f"module {__name__} has no attribute {name}")
 
@@ -150,14 +150,7 @@ class MeasurableOp(abc.ABC):
 MeasurableOp.register(RandomVariable)
 
 
-class MeasurableOpMixin(MeasurableOp):
-    """MeasurableOp Mixin with a distinctive string representation"""
-
-    def __str__(self):
-        return f"Measurable{super().__str__()}"
-
-
-class MeasurableElemwise(MeasurableOpMixin, Elemwise):
+class MeasurableElemwise(MeasurableOp, Elemwise):
     """Base class for Measurable Elemwise variables"""
 
     valid_scalar_types: tuple[MetaType, ...] = ()
@@ -169,3 +162,6 @@ class MeasurableElemwise(MeasurableOpMixin, Elemwise):
                 f"Acceptable types are {self.valid_scalar_types}"
             )
         super().__init__(scalar_op, *args, **kwargs)
+
+    def __str__(self):
+        return f"Measurable{super().__str__()}"
--- a/pymc/logprob/abstract.py
+++ b/pymc/logprob/abstract.py
@@ -40,7 +40,7 @@ import warnings
 from collections.abc import Sequence
 from functools import singledispatch
 
-from pytensor.graph.op import Op
+from pytensor.graph import Apply, Op, Variable
 from pytensor.graph.utils import MetaType
 from pytensor.tensor import TensorVariable
 from pytensor.tensor.elemwise import Elemwise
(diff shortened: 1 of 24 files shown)


ricardoV94:
The failing test should pass after the changes in #7480

pymc/logprob/abstract.py
```diff
@@ -169,3 +162,49 @@ def __init__(self, scalar_op, *args, **kwargs):
...
+
+    def __str__(self):
+        return f"Measurable{super().__str__()}"
+
+
+class ValuedRV(Op):
+    r"""Represents the association of a measurable variable and its value.
+
+    A `ValuedVariable` node represents the pair :math:`(Y, y)`, where
+    :math:`Y` is a random variable and :math:`y \sim Y`.
```
larryshamalama:
```suggestion
    `y` the value at which :math:`Y`'s density or probability mass function is evaluated.
```

pymc/logprob/abstract.py
```diff
@@ -169,3 +162,49 @@ def __init__(self, scalar_op, *args, **kwargs):
...
+
+    Log-probability (densities) are functions over these pairs, which makes
```
larryshamalama:
```suggestion
    The log-probability function takes such pairs as input, which makes
```
IIRC densities are only for continuous random variables. This is kind of nitpicky but you are the judge :)

pymc/logprob/abstract.py
```diff
@@ -169,3 +162,49 @@ def __init__(self, scalar_op, *args, **kwargs):
...
+
+    def infer_shape(self, fgraph, node, input_shapes):
+        return [input_shapes[0]]
+
+
+valued_rv = ValuedRV()
+
+
+class PromisedValuedRV(Op):
+    r"""Marks a variable as being promised a valued variable in the logprob method."""
```
larryshamalama: Why can't we directly provide a corresponding `ValuedRV` during the rewrite? Just asking to clarify. I see that this (only?) is used for `Join`s and `MakeVector`s

twiecki (approved):

twiecki:
This is pretty substantial.

tests/logprob/test_basic.py
```diff
@@ -413,3 +414,22 @@ def test_icdf_discrete():
...
+    # `b` to be rewritten as `pm.Normal.dist(a_base * 40)`, as it would then be disconnected from the valued `a`
+    a_base = pm.Normal.dist()
+    a = a_base * 5
+    b = pm.Normal.dist(a * 8)
+
+    a_value = a.type()
+    b_value = b.type()
+    logp_b = conditional_logp({a: a_value, b: b_value})[b_value]
+
+    assert_no_rvs(logp_b)
```
larryshamalama, line 434: Brainstorming to see if I understand here. From #6917, IIUC `assert_no_rvs(logp_b)` would throw a warning (or an error?).

When would we want `b` to be rewritten as `pm.Normal.dist(a_base * 40)`? When, say, `a_base_value = a_base.type()` is provided in `conditional_logp`? My understanding is that, while both situations are mathematically equivalent, PyMC's log-prob inference does not work well when value variables (`a_value`) of deterministic transformations of `RandomVariable`s (`a`) are provided in lieu of their value variables (`a_base.type()`; not provided). Is this correct?

pymc/logprob/abstract.py
larryshamalama: Revisiting this comment after skimming a bit more

I now see that there are a set of newly defined early measurable ir rewrites and I imagine that these promised value RVs are temporarily substituted into that intermediate graph before applying the other logprob rewrites. Why does this approach prevent breaking interdependencies?

larryshamalama (commented):
Hi @ricardoV94, great PR, as always, especially with so much extension in abstraction :)

A lot of my comments are to clarify some parts of the code and me thinking out loud as I went over the PR

Are you able to add a test showing an IR graph after applying rewrites from `early_measurable_ir_rewrites_db` but before `measurable_ir_rewrites_db`? Perhaps this could highlight where and why exactly `PromisedValuedRV`s are needed

tests/logprob/test_basic.py
ricardoV94: Yes, the general problem is that a second variable cannot depend on a transformation of the value variable. We don't have the machinery to do that sort of inversion (other than the ad-hoc transform_value rewrite). For example ,the following is not supported:

```python
x = pt.random.normal()
x_exp = pt.exp(x)
y = pt.random.normal(loc=x)

x_exp_vv = pt.scalar("x_exp_vv")
y_vv = pt.scalar("y_vv")
conditional_logprob({x_exp: x_exp_vv, y: y_vv})
```

There is nothing wrong in principle, but we don't have the machinery to find out that the density of `y` should depend on a (log) transform of `x_exp_vv`. Would be nice to have, but I haven't come across an elegant solution. The changes in this PR prevent our rewrites from introducing such indirections by accident by changing the IR of  this test example to something like.

```python
 a_base = pm.Normal.dist()
 a = valued_rv(a_base * 5, a_value)
 b = valued_rv(pm.Normal.dist(a * 8), b_value)
```
Since there are no default PyTensor rewrites that know what to do with a `valued_rv`, there is no risk of "mixing" information before and after the conditioning points (in this case, constant_folding `5 * 8 = 40` in the graph of `b`)

pymc/logprob/abstract.py
ricardoV94: We're trying to fix the same problem of not breaking dependencies. The funny thing with Join/MakeVector is that they combine multiple RVs, potentially interdependent, into a single composite valued node. Only in the logp function is this value split and sent to each component, but we still want to prevent the same outer-problem that motivated this PR. If you have:

```python
x1 = pt.random.normal() * 5
x2 = pt.random.normal(x1 * 8)
xs = pt.stack([x1, x2])
xs_vv = pt.vector("xs_vv", shape=(2,))
```

By the time you get to the logp of `xs_vv`, you want to basically do the same thing that the new test does, plus concatenate the logp terms. To prevent the conditioning from breaking during the IR rewrites we would want to introduce the ValuedRV nodes. We have to do this as an early rewrite, before anything comes up and breaks it. In normal situations we also do it before any rewrites in the `construct_ir_fgraph` code, but for join/stack this already accounts as some sort of inference.

Now why Promised and not just vanilla Valued? Just for convenience, because we still want a function from `xs_vv` to `stack([logp(x1), logp(x2)])` in the end, and if I split the values, I wouldn't know how to stack them later in the loop in `conditional_logp` (or if this was a single `logp` call), or even know that I needed to. So we just "promise" there will be values to avoid rewrites from breaking the dependency that will be required in the logp function.

Another way of thinking is that we are basically trying to truncate the graphs between valued nodes and do manipulations only within these subgraphs, not across them. We could have literally split the graph, done IR in each and collect logp expressions. What we are doing is identical. Either way, we would still need to do a further split for the RVs within Join/MakeVector.

ricardoV94:
@larryshamalama thanks for the thoughtful questions. I ended up cleaning up the answers and adding them in the docstrings of `ValuedRV` and `PromisedValuedRV`!

Issue #6883, opened 2023-08-30
Do not reference model variables in GP internals


ricardoV94:
### Description

The model fgraph functionality (used by `do` and `observe`) doesn't work with GP variables because those are not represented as pure PyTensor objects and make use of variables from the old model in the instance scope.

ferrine:
Can we refactor the code in the way it works correctly?

ricardoV94:
It's not easy we need to figure out how to represent GPs as PyTensor Ops

ricardoV94:
It's not easy, we need to figure out how to represent GPs as PyTensor variables/Ops.

Some very rough (and likely bad) ideas in https://github.com/pymc-devs/pymc-experimental/blob/2084e167486a42b313abcec77958d3a24e001ae3/pymc_experimental/gp/GPs%20from%20scratch.ipynb

ricardoV94:
There's actually no GP object saved in the model, so no good way to raise an error.

The problem is that GP objects hold references to model variables, so if you tried to do a gp.conditional after cloning a model (which happens whenever you do/observe) you would be re-introducing the old model variables, instead of referring to cloned ones. 

```python
import pymc as pm
import numpy as np
from pytensor.graph import ancestors
from pymc.model.fgraph import clone_model

with pm.Model() as model:
    ell = pm.Gamma("ell", alpha=2, beta=1)
    cov = 2 * pm.gp.cov.ExpQuad(1, ell)
    gp = pm.gp.Latent(cov_func=cov)
    f = gp.prior("f", X=np.arange(10)[:, None])
    pm.Normal("y", f * 2)

assert ell in ancestors(f)

cloned_model = clone_model(model)
with cloned_model:
    f_cloned = gp.prior("f_cloned", X=np.arange(10)[:, None])
    pm.Normal("y_cloned", f_cloned * 2)

assert model["ell"] in ancestors(f_cloned)  # bad
assert cloned_model["ell"] not in ancestors(f_cloned)  # bad
cloned_model.logp()  # ValueError: Random variables detected in the logp graph: {ell}.
```

I see two ways of fixing this:
1. Use an approach more like the `statespace` module in pymc-experimental, which retrieves all variables by name from the model context when those are needed (is that correct @jessegrabowski?)
2. Try to define GPs as PyTensor objects/operations (this is not trivial, but could be interesting in the long term). See link above.

jessegrabowski:
[...]

PR #7492, opened 2024-09-05
Allow copy and deepcopy of PYMC models

### LINKED ISSUE #6985: Implement `__copy__` and `__deepcopy__` model methods
### Description

A couple of users on discourse got some nasty bugs when trying to do `copy.deepcopy(model)`. Whatever the default is doing, it ain't great.

We have a `clone_model` in `model/fgraph` that we can use for both copy and deepcopy!

It should work in all cases except for the known issue with GP variables https://github.com/pymc-devs/pymc/issues/6883

Dekermanjian:
## Description
I added  `__copy__` and `__deepcopy__` methods to the Model class. Both will use the `clone_model()` method from `pymc.model.fgraph` and if any guassian process variables are detected in the model an exception is raised. I also created 2 unit tests to test both that the methods work and that the exception is raised upon detecting a gaussian process variable

## Related Issue

- [x] Closes #6985 
- [x] Related to #6883 

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7492.org.readthedocs.build/en/7492/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1579,8 +1579,9 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
     def copy(self):
         """
-        Clone a pymc model by overiding the python copy method using the clone_model method from fgraph.
-        Constants are not cloned and if guassian process variables are detected then a warning will be triggered.
+        Clone the model
+
+        To access variables in the cloned model use `cloned_model["var_name"]`.
 
         Examples
         --------
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1570,6 +1570,22 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
     def __contains__(self, key):
         return key in self.named_vars or self.name_for(key) in self.named_vars
+    
+    def __copy__(self):
+        from pymc.model.fgraph import clone_model
+        check_for_gp_vars = [k for x in ['_rotated_', '_hsgp_coeffs_'] for k in self.named_vars.keys() if x in k]
+        if len(check_for_gp_vars) > 0:
+            raise Exception("Unable to clone Gaussian Process Variables")
+    
+        return clone_model(self)
+    
+    def __deepcopy__(self, _):
+        from pymc.model.fgraph import clone_model
+        check_for_gp_vars = [k for x in ['_rotated_', '_hsgp_coeffs_'] for k in self.named_vars.keys() if x in k]
+        if len(check_for_gp_vars) > 0:
+            raise Exception("Unable to clone Gaussian Process Variables")
+        
+        return clone_model(self)
 
     def replace_rvs_by_values(
         self,
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1570,21 +1570,35 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
     def __contains__(self, key):
         return key in self.named_vars or self.name_for(key) in self.named_vars
-    
+
     def __copy__(self):
+        """
+        Clone a pymc model by overiding the python copy method using the clone_model method from fgraph.
+        if guassian process variables are detected then an exception will be raised.
+        """
         from pymc.model.fgraph import clone_model
-        check_for_gp_vars = [k for x in ['_rotated_', '_hsgp_coeffs_'] for k in self.named_vars.keys() if x in k]
+
+        check_for_gp_vars = [
+            k for x in ["_rotated_", "_hsgp_coeffs_"] for k in self.named_vars.keys() if x in k
+        ]
         if len(check_for_gp_vars) > 0:
             raise Exception("Unable to clone Gaussian Process Variables")
-    
+
         return clone_model(self)
-    
+
     def __deepcopy__(self, _):
+        """
+        Clone a pymc model by overiding the python copy method using the clone_model method from fgraph.
+        if guassian process variables are detected then an exception will be raised.
+        """
         from pymc.model.fgraph import clone_model
-        check_for_gp_vars = [k for x in ['_rotated_', '_hsgp_coeffs_'] for k in self.named_vars.keys() if x in k]
+
+        check_for_gp_vars = [
+            k for x in ["_rotated_", "_hsgp_coeffs_"] for k in self.named_vars.keys() if x in k
+        ]
         if len(check_for_gp_vars) > 0:
             raise Exception("Unable to clone Gaussian Process Variables")
-        
+
         return clone_model(self)
 
     def replace_rvs_by_values(
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1574,30 +1574,56 @@ class Model(WithMemoization, metaclass=ContextMeta):
     def __copy__(self):
         """
         Clone a pymc model by overiding the python copy method using the clone_model method from fgraph.
-        if guassian process variables are detected then an exception will be raised.
+        Constants are not cloned and if guassian process variables are detected then a warning will be triggered.
+
+        Examples
+        --------
+        .. code-block:: python
+
+            import pymc as pm
+            import copy
+
+            with pm.Model() as m:
+                p = pm.Beta("p", 1, 1)
+                x = pm.Bernoulli("x", p=p, shape=(3,))
+
+            clone_m = copy.copy(m)
+
+            # Access cloned variables by name
+            clone_x = clone_m["x"]
+
+            # z will be part of clone_m but not m
+            z = pm.Deterministic("z", clone_x + 1)
         """
         from pymc.model.fgraph import clone_model
-
-        check_for_gp_vars = [
-            k for x in ["_rotated_", "_hsgp_coeffs_"] for k in self.named_vars.keys() if x in k
-        ]
-        if len(check_for_gp_vars) > 0:
-            raise Exception("Unable to clone Gaussian Process Variables")
 
         return clone_model(self)
 
     def __deepcopy__(self, _):
         """
         Clone a pymc model by overiding the python copy method using the clone_model method from fgraph.
-        if guassian process variables are detected then an exception will be raised.
+        Constants are not cloned and if guassian process variables are detected then a warning will be triggered.
+
+        Examples
+        --------
+        .. code-block:: python
+
+            import pymc as pm
+            import copy
+
+            with pm.Model() as m:
+                p = pm.Beta("p", 1, 1)
+                x = pm.Bernoulli("x", p=p, shape=(3,))
+
+            clone_m = copy.deepcopy(m)
+
+            # Access cloned variables by name
+            clone_x = clone_m["x"]
+
+            # z will be part of clone_m but not m
+            z = pm.Deterministic("z", clone_x + 1)
         """
         from pymc.model.fgraph import clone_model
-
-        check_for_gp_vars = [
-            k for x in ["_rotated_", "_hsgp_coeffs_"] for k in self.named_vars.keys() if x in k
-        ]
-        if len(check_for_gp_vars) > 0:
-            raise Exception("Unable to clone Gaussian Process Variables")
 
         return clone_model(self)
 
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -1572,6 +1572,12 @@ class Model(WithMemoization, metaclass=ContextMeta):
         return key in self.named_vars or self.name_for(key) in self.named_vars
 
     def __copy__(self):
+        return self.copy()
+
+    def __deepcopy__(self, _):
+        return self.copy()
+
+    def copy(self):
         """
         Clone a pymc model by overiding the python copy method using the clone_model method from fgraph.
         Constants are not cloned and if guassian process variables are detected then a warning will be triggered.
@@ -1588,34 +1594,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
                 x = pm.Bernoulli("x", p=p, shape=(3,))
 
             clone_m = copy.copy(m)
-
-            # Access cloned variables by name
-            clone_x = clone_m["x"]
-
-            # z will be part of clone_m but not m
-            z = pm.Deterministic("z", clone_x + 1)
-        """
-        from pymc.model.fgraph import clone_model
-
-        return clone_model(self)
-
-    def __deepcopy__(self, _):
-        """
-        Clone a pymc model by overiding the python copy method using the clone_model method from fgraph.
-        Constants are not cloned and if guassian process variables are detected then a warning will be triggered.
-
-        Examples
-        --------
-        .. code-block:: python
-
-            import pymc as pm
-            import copy
-
-            with pm.Model() as m:
-                p = pm.Beta("p", 1, 1)
-                x = pm.Bernoulli("x", p=p, shape=(3,))
-
-            clone_m = copy.deepcopy(m)
 
             # Access cloned variables by name
             clone_x = clone_m["x"]
--- a/pymc/model/fgraph.py
+++ b/pymc/model/fgraph.py
@@ -160,6 +160,12 @@ def fgraph_from_model(
             "Nested sub-models cannot be converted to fgraph. Convert the parent model instead"
         )
 
+    check_for_gp_vars = [
+        k for x in ["_rotated_", "_hsgp_coeffs_"] for k in model.named_vars.keys() if x in k
+    ]
+    if len(check_for_gp_vars) > 0:
+        warnings.warn("Unable to clone Gaussian Process Variables", UserWarning)
+
     # Collect PyTensor variables
     rvs_to_values = model.rvs_to_values
     rvs = list(rvs_to_values.keys())
@@ -393,11 +399,6 @@ def clone_model(model: Model) -> Model:
             z = pm.Deterministic("z", clone_x + 1)
 
     """
-    check_for_gp_vars = [
-        k for x in ["_rotated_", "_hsgp_coeffs_"] for k in model.named_vars.keys() if x in k
-    ]
-    if len(check_for_gp_vars) > 0:
-        warnings.warn("Unable to clone Gaussian Process Variables", UserWarning)
     return model_from_fgraph(fgraph_from_model(model)[0], mutate_fgraph=True)
 
 
--- a/pymc/model/fgraph.py
+++ b/pymc/model/fgraph.py
@@ -160,11 +160,13 @@ def fgraph_from_model(
             "Nested sub-models cannot be converted to fgraph. Convert the parent model instead"
         )
 
-    check_for_gp_vars = [
-        k for x in ["_rotated_", "_hsgp_coeffs_"] for k in model.named_vars.keys() if x in k
-    ]
-    if len(check_for_gp_vars) > 0:
-        warnings.warn("Unable to clone Gaussian Process Variables", UserWarning)
+    if any(
+        ("_rotated_" in var_name or "_hsgp_coeffs_" in var_name) for var_name in model.named_vars
+    ):
+        warnings.warn(
+            "Detected variables likely created by GP objects. Further use of these old GP objects should be avoided as it may reintroduce variables from the old model. See issue: https://github.com/pymc-devs/pymc/issues/6883",
+            UserWarning,
+        )
 
     # Collect PyTensor variables
     rvs_to_values = model.rvs_to_values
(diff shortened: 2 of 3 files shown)


Dekermanjian:
Hey @ricardoV94, sorry if this is a silly question. Do I need to do anything to get the `all_tests` and `mypy` to start running?

ricardoV94:
CI has to be manually approved for first time contributors, doing it now

Dekermanjian:
Okay! Thank you very much!!

pymc/model/core.py
```diff
@@ -1584,6 +1584,36 @@ def __getitem__(self, key):
...
+        Clone a pymc model by overiding the python copy method using the clone_model method from fgraph.
+        if guassian process variables are detected then an exception will be raised.
+        """
+        from pymc.model.fgraph import clone_model
+
+        check_for_gp_vars = [
+            k for x in ["_rotated_", "_hsgp_coeffs_"] for k in self.named_vars.keys() if x in k
+        ]
+        if len(check_for_gp_vars) > 0:
+            raise Exception("Unable to clone Gaussian Process Variables")
```
ricardoV94: This should be in the low-level utility used by `clone_model`, `fgraph_from_model`? Also I think it should be a warning, because we are not sure something with `_rotated_` is a GP or just a user variable name for a vanilla variable.

Also @bwengals any other names we could look for to detect GPs in the model? This is not perfect, but avoiding as much surprise as possible would be great.

tests/model/test_core.py
```diff
@@ -1761,3 +1762,52 @@ def test_graphviz_call_function(self, var_names, filenames) -> None:
                 figsize=None,
                 dpi=300,
             )
+
+
+class TestModelCopy(unittest.TestCase):
```
ricardoV94: We use pytest, any reason we need to subclass `unittest`?

pymc/model/core.py
```diff
@@ -1584,6 +1584,36 @@ def __getitem__(self, key):
     def __contains__(self, key):
         return key in self.named_vars or self.name_for(key) in self.named_vars
 
+    def __copy__(self):
```
ricardoV94, line 1574: Mention `clone_model`  docstrings / add code example on how to access variables after cloning (i.e, `new_model["var_name"]`) since the variables are copied and the original ones are not part of the cloned model. 

I suggest this because it's something unique about working with model copies, and not many people know how to retrieve variables by name

ricardoV94:
@Dekermanjian looks great! I left some suggestions, let me know if something does not make sense

pymc/model/core.py
Dekermanjian: @ricardoV94 Yes, you are absolutely correct. This should be a warning and yes, I can move it to the lower-level utility used by `clone_model`. Do you know if there is a way that I can check model variables to discern the type of the variable it is? Something like `type(model.named_vars)` and it would say it was a GP

tests/model/test_core.py
Dekermanjian: Oh, I must have made a mistake. I am sorry I am new to using pytest. I will fix that as well!

pymc/model/core.py
Dekermanjian: Okay, I will mention the `clone_model` documentation and I will add a code example for accessing the variables after cloning!

Dekermanjian:
Thank you @ricardoV94 for your feedback! I appreciate it!

pymc/model/core.py
ricardoV94: Nope, no way to know it was a GP, that's the unfortunate bit. GPs produce vanilla MvNormals or deterministics

pymc/model/fgraph.py
```diff
@@ -391,6 +393,11 @@ def clone_model(model: Model) -> Model:
             z = pm.Deterministic("z", clone_x + 1)
 
     """
+    check_for_gp_vars = [
```
ricardoV94: should be in the even lower fgraph_from_model

Dekermanjian: Oh! my bad, I will fix that now. Thank you @ricardoV94!

Dekermanjian:
Hey @ricardoV94, is there anything you'd like me to modify/add to this pull request?

pymc/model/core.py
bwengals: Sorry for the very delayed response.  No, there isn't.  What I've done to smuggle GPs out with the model in the past is to do something like:

```python
with pm.Model() as model:
    # model code here
    gp1 = pm.gp.Latent(...)
    gp2  = pm.gp.HSGP(...)

model.gps = {"gp1": gp1, "gp2": gp2, ...}
```

If GPs added themselves to the model context automatically they could be tracked.  They don't have names though, so need a little thought for how to key that dictionary, although maybe putting them in a list is OK.

Dekermanjian: Hey @bwengals, if I understand you correctly, you are saying that we can add GPs to the Model class as an attribute and that would happen in `core.py` in the section after line 539 and then we can pull those variables just like the rest of the variables in `fgraph_from_model()` in the section after line 169 inside of the `fgraph.py` file. 
(sorry for the line number references I couldn't figure out how to highlight the code in the comment. I recently just started contributing to open source)

pymc/model/fgraph.py
```diff
@@ -158,6 +160,12 @@ def fgraph_from_model(
             "Nested sub-models cannot be converted to fgraph. Convert the parent model instead"
         )
 
+    check_for_gp_vars = [
+        k for x in ["_rotated_", "_hsgp_coeffs_"] for k in model.named_vars.keys() if x in k
+    ]
+    if len(check_for_gp_vars) > 0:
+        warnings.warn("Unable to clone Gaussian Process Variables", UserWarning)
```
ricardoV94: This can be simplified

```suggestion
    if any(name in ("_rotated_", "_hsgp_coeffs_") for name in model.named_vars):
        warnings.warn("Unable to clone Gaussian Process Variables", UserWarning)
```

pymc/model/fgraph.py
```diff
@@ -369,7 +377,7 @@ def clone_model(model: Model) -> Model:
 
     Recreates a PyMC model with clones of the original variables.
     Shared variables will point to the same container but be otherwise different objects.
-    Constants are not cloned.
+    Constants are not cloned and if guassian process variables are detected then a warning will be triggered.
```
ricardoV94: Let's not mention it, this is just a temporary thing we want to fix

```suggestion
    Constants are not cloned.
```

pymc/model/fgraph.py
Dekermanjian: Yeah, that looks better. I will make that change. Thank you!

pymc/model/fgraph.py
Dekermanjian: Okay, I will make that change as well!

pymc/model/fgraph.py
Dekermanjian: Oh, hey @ricardoV94. Since we moved that check to the `fgraph_from_model()` function maybe the warning should say something different instead of mentioning cloning? Is this function only used for cloning?

Dekermanjian: Hey, one more thing. I wasn't able to get the above suggestion to work. I think that we need to check if the string subset ("_rotated_" & "_hsgp_coeffs_") are in the name of the variable. I was able to get it to work by making a little change:

```
if any(gp_name in var_name for gp_name in ["_rotated_", "_hsgp_coeffs_"] for var_name in model.named_vars):
        warnings.warn("Unable to clone Gaussian Process Variables", UserWarning)
```
Is that okay with you?

ricardoV94: Oh sorry of course. Since there are only two options I would do an `or` statement then: `if '_rotated_' in var_name or '_hsgp_coeffs_' in var_name for var_name in ...`

Dekermanjian: Hey @ricardoV94, I couldn't figure out how to get it to work by only looping through model.named_vars one time like you have in your example. I had to do a loop in each one of the `or` cases like this: 

```
if any("_rotated_" in var_name for var_name in model.named_vars) or any("_hsgp_coeffs_" in var_name for var_name in model.named_vars):
        warnings.warn("Unable to clone Gaussian Process Variables", UserWarning)
```

Am I missing something?

ricardoV94: `if any(("_rotated_" in var_name or "_hsgp_coeffs_" in var_name) for var_name in model.named_vars)`

ricardoV94: Also let's give a more informative warning. Something like: `"Detected variables likely created by GP objects. Further use of these old GP objects should be avoided as it may reintroduce variables from the old model. See issue: https://github.com/pymc-devs/pymc/issues/6883"`.

Dekermanjian: Thanks @ricardoV94, I have made the suggestions and pushed the changes.

Dekermanjian:
Thank you @ricardoV94 for your help with this and for your patience. I really appreciate it.

tests/model/test_core.py
```diff
@@ -1761,3 +1762,62 @@ def test_graphviz_call_function(self, var_names, filenames) -> None:
...
+    def gp_model() -> pm.Model:
+        with pm.Model() as gp_model:
+            ell = pm.Gamma("ell", alpha=2, beta=1)
+            cov = 2 * pm.gp.cov.ExpQuad(1, ell)
+            gp = pm.gp.Latent(cov_func=cov)
+            f = gp.prior("f", X=np.arange(10)[:, None])
+            pm.Normal("y", f * 2)
+        return gp_model
+
+    def test_copy_model(self) -> None:
```
ricardoV94: These tests have a lot of duplicated code that we can avoid if you use `pytest.mark.parametrize("copy_method", (copy.copy, copy.deepcopy))`

tests/model/test_core.py
```diff
@@ -1761,3 +1762,62 @@ def test_graphviz_call_function(self, var_names, filenames) -> None:
                 figsize=None,
                 dpi=300,
             )
+
+
+class TestModelCopy:
+    @staticmethod
+    def simple_model() -> pm.Model:
```
ricardoV94: No need to use staticmethods as these aren't used in more than one single test?

tests/model/test_core.py
```diff
@@ -1761,3 +1762,62 @@ def test_graphviz_call_function(self, var_names, filenames) -> None:
...
+            pm.Normal("y", f * 2)
+        return gp_model
+
+    def test_copy_model(self) -> None:
+        simple_model = self.simple_model()
+        copy_simple_model = copy.copy(simple_model)
+        deepcopy_simple_model = copy.deepcopy(simple_model)
+
+        with simple_model:
+            simple_model_prior_predictive = pm.sample_prior_predictive(random_seed=42)
```
ricardoV94: Taking a single draw should be enough.

I would also test that adding a deterministic to the copy model does not introduce one in the original model (basically the example you had in the docstrings)

pymc/model/core.py
```diff
@@ -1571,6 +1571,62 @@ def __getitem__(self, key):
...
```
ricardoV94, line 1574: Since these are dunder methods I guess the docstrings won't be seen? Can we add a third method `.copy()` that actually has the docstrings? 

Also `__copy__` and `__deepcopy__` can just call `self.copy()` to avoid the extra lines for importing clone_model

tests/model/test_core.py
```diff
@@ -1761,3 +1762,62 @@ def test_graphviz_call_function(self, var_names, filenames) -> None:
...
+        assert np.isclose(
+            simple_model_prior_predictive_mean, copy_simple_model_prior_predictive_mean
+        )
+        assert np.isclose(
+            simple_model_prior_predictive_mean, deepcopy_simple_model_prior_predictive_mean
+        )
+
+    def test_guassian_process_copy_failure(self) -> None:
+        gaussian_process_model = self.gp_model()
+        with pytest.warns(UserWarning):
```
ricardoV94: Add a match kwarg to check the UserWarning is actually the one we care about 
```suggestion
        with pytest.warns(UserWarning match=...):
```

ricardoV94 (changes requested):
Hi, I had another pass and I have a couple minor suggestions. I think this should be it!

Thanks for your patience as well 🙏

Dekermanjian:
Hey Ricardo, I made the suggested changes. The tests look a lot better now. Thank you for reviewing it!

pymc/model/core.py
```diff
@@ -1571,6 +1571,40 @@ def __getitem__(self, key):
...
+        return self.copy()
+
+    def __deepcopy__(self, _):
+        return self.copy()
+
+    def copy(self):
+        """
+        Clone a pymc model by overiding the python copy method using the clone_model method from fgraph.
+        Constants are not cloned and if guassian process variables are detected then a warning will be triggered.
```
```suggestion
        Clone the model.
        
        To access variables in the cloned model use `cloned_model["var_name"]`.
```

tests/model/test_core.py
```diff
@@ -1761,3 +1762,63 @@ def test_graphviz_call_function(self, var_names, filenames) -> None:
...
+            copy_simple_model_prior_predictive = pm.sample_prior_predictive(
+                samples=1, random_seed=42
+            )
+
+        simple_model_prior_predictive_mean = simple_model_prior_predictive["prior"]["y"].mean(
+            ("chain", "draw")
+        )
+        copy_simple_model_prior_predictive_mean = copy_simple_model_prior_predictive["prior"][
+            "y"
+        ].mean(("chain", "draw"))
```
ricardoV94: No need to take them mean, now that it's a single value. Just retrieve it with `simple_model_prior_preictive.prior["y"].values`

tests/model/test_core.py
```diff
@@ -1761,3 +1762,63 @@ def test_graphviz_call_function(self, var_names, filenames) -> None:
...
+            pm.Normal("y", f * 2)
+
+        with pytest.warns(
+            UserWarning,
+            match="Detected variables likely created by GP objects. Further use of these old GP objects should be avoided as it may reintroduce variables from the old model. See issue: https://github.com/pymc-devs/pymc/issues/6883",
+        ):
+            copy_method(gaussian_process_model)
+
+    @pytest.mark.parametrize("copy_method", (copy.copy, copy.deepcopy))
+    def test_adding_deterministics_to_clone(self, copy_method) -> None:
```
ricardoV94: This check can be done in the first test. That way you can also confirm the prior_predictive["z"] draw is what you expect (and only exists in the cloned model)

tests/model/test_core.py
```diff
@@ -1761,3 +1762,63 @@ def test_graphviz_call_function(self, var_names, filenames) -> None:
...
+            )
+
+        simple_model_prior_predictive_mean = simple_model_prior_predictive["prior"]["y"].mean(
+            ("chain", "draw")
+        )
+        copy_simple_model_prior_predictive_mean = copy_simple_model_prior_predictive["prior"][
+            "y"
+        ].mean(("chain", "draw"))
+
+        assert np.isclose(
```
ricardoV94: You can check exact equality, since the draws are exactly the same

tests/model/test_core.py
```diff
@@ -1786,15 +1782,24 @@ def test_copy_model(self, copy_method) -> None:
...
 
-        assert np.isclose(
-            simple_model_prior_predictive_mean, copy_simple_model_prior_predictive_mean
+        assert simple_model_prior_predictive_val == copy_simple_model_prior_predictive_val
+
+        with copy_simple_model:
+            z = pm.Deterministic("z", copy_simple_model["alpha"] + 1)
+            copy_simple_model_prior_predictive = pm.sample_prior_predictive(
+                samples=1, random_seed=42
+            )
```
ricardoV94: You can do this above, and call `sample_prior_predictive` only once for the `copy_simple_model`

tests/model/test_core.py
```diff
@@ -1786,15 +1782,24 @@ def test_copy_model(self, copy_method) -> None:
                 samples=1, random_seed=42
             )
 
-        simple_model_prior_predictive_mean = simple_model_prior_predictive["prior"]["y"].mean(
-            ("chain", "draw")
-        )
-        copy_simple_model_prior_predictive_mean = copy_simple_model_prior_predictive["prior"][
+        simple_model_prior_predictive_val = simple_model_prior_predictive["prior"]["y"].values
+        copy_simple_model_prior_predictive_val = copy_simple_model_prior_predictive["prior"][
```
ricardoV94: Just compare directly, no need to assign to separate variables, that are almost as verbose as the way they are accessed

ricardoV94 (commented):
Very minor nitpicks to reduce test verbosity.

ricardoV94:
It seems I keep trying to make the tests smaller. Very sorry! Let me know if it feels too much :)

ricardoV94:
> So I have to sample prior predictives twice one without the deterministic and another time with the deterministic. Anyway that can be simplified?

Ah... I didn't expect that, it shouldn't be the case in theory. Let me have a look, but we can leave as is if true

Dekermanjian:
> It seems I keep trying to make the tests smaller. Very sorry! Let me know if it feels too much :)

Not a problem at all. It is for the best to remove any redundancies.

Dekermanjian:
> > So I have to sample prior predictives twice one without the deterministic and another time with the deterministic. Anyway that can be simplified?
> 
> Ah... I didn't expect that, it shouldn't be the case in theory. Let me have a look, but we can leave as is if true

I tried this again. When I sample prior predictive with the copy model that includes the deterministic the variable 'y' value no longer matches the value from the original model and the assert statement doesn't pass.

ricardoV94:
I get the same values locally:

```python
import pymc as pm
from pymc.model.fgraph import clone_model
with pm.Model() as m:
    x = pm.Normal("x")
    print(pm.sample_prior_predictive(draws=1, random_seed=1).prior["x"].values)     
    
with clone_model(m) as new_m:
    y = pm.Deterministic("y", new_m["x"] + 1)
    print(pm.sample_prior_predictive(draws=1, random_seed=1).prior["x"].values)            
    
Sampling: [x]
[[-0.64031853]]
Sampling: [x]
[[-0.64031853]]
```

Dekermanjian:
> ```python
> import pymc as pm
> from pymc.model.fgraph import clone_model
> with pm.Model() as m:
>     x = pm.Normal("x")
>     print(pm.sample_prior_predictive(draws=1, random_seed=1).prior["x"].values)     
>     
> with clone_model(m) as new_m:
>     y = pm.Deterministic("y", new_m["x"] + 1)
>     print(pm.sample_prior_predictive(draws=1, random_seed=1).prior["x"].values)  
> ```

Hmm, I can get your example to work locally but it still gives me a different value for my example:
image

Dekermanjian:
I also notice that if my deterministic is a function of alpha or error it will change the value of the prior predictive. However, if my deterministic is a function of the likelihood (like in your example) then the prior values are exactly the same.

ricardoV94:
Yes, I said so above, but maybe it was not obvious. With multiple variables, the deterministic ends up changing which seed goes to which variable (after they are split internally). But your test does not need the extra variables right? The simpler, single variable should be enough (it doesn't matter if it's prior or likelihood, that's all the same for prior_predictive sampling)

ricardoV94 (approved):

ricardoV94:
Thanks a ton @Dekermanjian 😊

Dekermanjian:
@ricardoV94 Happy to help! :)

Ricardo, I have a quick question for you. I am trying to learn the lower level code for PyTensor so that I can contribute in a more meaningful way. I have walked through the PyTensor Dev tutorials but the tutorials are a little bit terse. I am wondering if there are more resources to learning the low level code. Maybe an example somewhere that shows how a simple PYMC model is constructed without using the high level abstractions. So defining the graph, the logp, and the sampling algorithm from scratch. Is something like that available?

ricardoV94:
@Dekermanjian have you seen this guide? https://www.pymc.io/projects/docs/en/stable/learn/core_notebooks/pymc_pytensor.html

I have actually been thinking about doing a series on implementing the core of PyMC in a live stream, which you might have liked if it already existed ? :)

Dekermanjian:
> https://www.pymc.io/projects/docs/en/stable/learn/core_notebooks/pymc_pytensor.html

@ricardoV94 I have not seen that one before! Thank you this will be very helpful. 

>  have actually been thinking about doing a series on implementing the core of PyMC in a live stream, which you might have > liked if it already existed ? :)

Yes, that would be fantastic. PYMC is wonderful for users because of the abstractions but for those who are interested in contributing development efforts it is difficult to connect the pieces just by going through the code base.

Issue #7495, opened 2024-09-06
Latex`_` in variable names is not properly escaped


ferrine:
https://github.com/pymc-devs/pymc/blob/d596afbe266ae1bdafec6735d73a82885e06dea1/pymc/printing.py#L261

Having this issue for years but did not pay much attention to it

<img width="783" alt="image" src="https://github.com/user-attachments/assets/3a9e42c8-f33b-42e8-839a-d21cfe0a928a">

ricardoV94:
There's an old open PR that was fixing it

ricardoV94:
#6533

Dekermanjian:
Hey, I believe that this is now fixed in #7501

PR #7501, opened 2024-09-11
Fix latex rendering of variables with underscore in name

### LINKED ISSUE #6508: BUG: KaTeX parse error on Jupyter Notebook when using underscore, etc.

Dekermanjian:
## Description
I added a simple fix to LaTeX Repr that escapes variables that have underscores in them and I created a unit test to test the fix.

## Related Issue

- [x] Closes #6508 
- [ ] Related to #7495 and #6533

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7501.org.readthedocs.build/en/7501/

### DIFF
--- a/pymc/printing.py
+++ b/pymc/printing.py
@@ -58,6 +58,7 @@ def str_for_dist(
     if "latex" in formatting:
         if print_name is not None:
             print_name = r"\text{" + _latex_escape(print_name.strip("$")) + "}"
+            print_name = _format_underscore(print_name)
 
         op_name = (
             dist.owner.op._print_name[1]
@@ -307,8 +308,15 @@ def _format_underscore(variable: str) -> str:
     """
     if "_" not in variable:
         return variable
+
     inds = [i for i, ltr in enumerate(variable) if ltr == "_"]
-    for i, ind in enumerate(inds):
-        ind = ind + i
-        variable = variable[:ind] + "\\" + variable[ind:]
+    var_len_original = len(variable)
+    var_len = None
+    for ind in inds:
+        if var_len:
+            if var_len != var_len_original:
+                ind = ind + (var_len - var_len_original)
+        if variable[ind - 1 : ind] != "\\":
+            variable = variable[:ind] + "\\" + variable[ind:]
+            var_len = len(variable)
     return variable
--- a/pymc/printing.py
+++ b/pymc/printing.py
@@ -114,6 +114,7 @@ def str_for_model(model: Model, formatting: str = "plain", include_params: bool 
     if not var_reprs:
         return ""
     if "latex" in formatting:
+        var_reprs = [_format_underscore(x) for x in var_reprs]
         var_reprs = [
             var_repr.replace(r"\sim", r"&\sim &").strip("$")
             for var_repr in var_reprs
@@ -295,3 +296,19 @@ try:
 except (ModuleNotFoundError, AttributeError):
     # no ipython shell
     pass
+
+
+def _format_underscore(variable: str) -> str:
+    """
+    formats variables with underscores in its name by prefixing underscores by '\\'
+    ---
+    Params:
+        variable: The string representation of the variable in the model
+    """
+    if "_" not in variable:
+        return variable
+    inds = [i for i, ltr in enumerate(variable) if ltr == "_"]
+    for i, ind in enumerate(inds):
+        ind = ind + i
+        variable = variable[:ind] + "\\" + variable[ind:]
+    return variable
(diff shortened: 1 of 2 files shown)


Dekermanjian:
Hey @ricardoV94, I added a simplistic fix to address the underscores breaking the KaTeX output in VSCode and Jupyter. I tested in VSCode and Jupyter Lab. Are there any other environments I should test it out in?

twiecki (approved):

Dekermanjian:
Hey @twiecki, I think some of the tests failed because they expect the LaTeX representation of "_"'s to be unescaped. Would it be okay, to modify those tests so that they expect "_"'s for LaTeX to be escaped?

twiecki:
> Hey @twiecki, I think some of the tests failed because they expect the LaTeX representation of underscores to be unescaped. Would it be okay, to modify those tests so that they expect underscores for LaTeX to be escaped?

Yes.

Dekermanjian:
Hey @twiecki, I applied the underscore escaping function to also work at the distribution level within a model. So both calling objects holding distributions or calling the entire model will escape underscore variables for LaTeX representation. I also changed the expected string objects in the testing script to expect that underscores in LaTeX be escaped.

pymc/printing.py
```diff
@@ -295,3 +297,26 @@ def _default_repr_pretty(obj: TensorVariable | Model, p, cycle):
 except (ModuleNotFoundError, AttributeError):
     # no ipython shell
     pass
+
+
+def _format_underscore(variable: str) -> str:
```
twiecki, line 304: Does this work equivalently?
```
import re

def _format_underscore(variable: str) -> str:
    """
    Escapes all unescaped underscores in the variable name for LaTeX representation.
    """
    return re.sub(r'(?<!\\)_', r'\\_', variable)
```

Dekermanjian: Oh, that looks a lot better. Let me test it out.

tests/test_printing.py
```diff
@@ -316,3 +318,25 @@ def random(rng, mu, size):
...
+class TestLatexRepr:
+    @staticmethod
+    def simple_model() -> Model:
+        with Model() as simple_model:
+            error = HalfNormal("error", 0.5)
+            alpha_a = Normal("alpha_a", 0, 1)
+            Normal("y", alpha_a, error)
+        return simple_model
+
+    def test_latex_escaped_underscore(self):
```
twiecki, line 331: Instead of iterating through matches, directly assert that the LaTeX string contains only escaped underscores:
```
assert "\\_" in model_str
assert "_" not in model_str.replace("\\_", "")
```

pymc/printing.py
Dekermanjian: Yup, that seems to work equivalently! I will replace the long function with this.

tests/test_printing.py
Dekermanjian: Yes, I agree. That is a much cleaner way of doing it. I will make the change now.

twiecki (approved):

PR #7506, opened 2024-09-17
Add Truncated normal dispatches

### LINKED ISSUE #7489: ENH: Add dispatches of `TruncatedNormal` distribution for forward sampling
### Before

```python
with pm.Model() as m:
    a = pm.TruncatedNormal("a", 0, 1, lower=-1, upper=1)

pm.draw(a, mode="JAX") # Fails
pm.draw(a, mode="NUMBA") # Fails
```

### After

```python
with pm.Model() as m:
    a = pm.TruncatedNormal("a", 0, 1, lower=-1, upper=1)

pm.draw(a, mode="JAX") # Works
pm.draw(a, mode="NUMBA") # Works
```

### Context for the issue:

The `TruncatedNormal` distribution creates a `TruncatedNormalRV` `Op` that doesn't have dispatch rules for either JAX or NUMBA. The `Truncated` class itself seems to work though. In general, it would be nice to know where to write dispatches for these special pymc Ops.

HarshvirSandhu:
## Description
Add jax dispatch for truncated normal distribution

## Related Issue

- [x] Closes #7489

## Checklist

## Type of change

----
📚 Documentation preview 📚: https://pymc--7506.org.readthedocs.build/en/7506/

### DIFF
--- a/pymc/dispatch/dispatch_jax.py
+++ b/pymc/dispatch/dispatch_jax.py
@@ -12,6 +12,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 import jax
+import jax.numpy as jnp
 
 from pytensor.link.jax.dispatch import jax_funcify
 
@@ -25,8 +26,19 @@ def jax_funcify_TruncatedNormalRV(op, **kwargs):
         rng_key, sampling_key = jax.random.split(rng_key, 2)
         key["jax_state"] = rng_key
 
-        truncnorm = jax.nn.initializers.truncated_normal(sigma, lower=lower, upper=upper)
+        if lower is None:
+            lower = -jnp.inf
+        if upper is None:
+            upper = jnp.inf
+        else:
+            new_lower, new_upper = (lower - mu) / sigma, (upper - mu) / sigma
 
-        return key, truncnorm(key["jax_state"], size) + mu
+        if size is None:
+            size = jnp.broadcast_arrays(jnp.array(mu), jnp.array(sigma))[0].shape
+
+        res = jax.random.truncated_normal(key["jax_state"], new_lower, new_upper, shape=size)
+        res = res * sigma + mu
+
+        return key, res
 
     return trunc_normal_fn
--- a/pymc/dispatch/dispatch_jax.py
+++ b/pymc/dispatch/dispatch_jax.py
@@ -15,14 +15,18 @@ import jax
 
 from pytensor.link.jax.dispatch import jax_funcify
 
-from pymc.distributions.continous import TruncatedNormalRV
+from pymc.distributions.continuous import TruncatedNormalRV
 
 
 @jax_funcify.register(TruncatedNormalRV)
 def jax_funcify_TruncatedNormalRV(op, **kwargs):
     def trunc_normal_fn(key, size, mu, sigma, lower, upper):
-        return None, jax.random.truncated_normal(
-            key["jax_state"], lower=lower, upper=upper, shape=size
-        )
+        rng_key = key["jax_state"]
+        rng_key, sampling_key = jax.random.split(rng_key, 2)
+        key["jax_state"] = rng_key
+
+        truncnorm = jax.nn.initializers.truncated_normal(sigma, lower=lower, upper=upper)
+
+        return key, truncnorm(key["jax_state"], size) + mu
 
     return trunc_normal_fn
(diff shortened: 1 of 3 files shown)


tests/dispatch/test_jax.py
```diff
@@ -0,0 +1,36 @@
...
+            [],
+            [pm.TruncatedNormal("a", 0, 1, lower=-1, upper=2, rng=np.random.default_rng(seed=123))],
+            mode="JAX",
+        )
+        f_py = function(
+            [],
+            [pm.TruncatedNormal("b", 0, 1, lower=-1, upper=2, rng=np.random.default_rng(seed=123))],
+        )
+
+    assert jax.numpy.array_equal(a1=f_py(), a2=f_jax())
```
HarshvirSandhu: This fails with a `NotImplementedError: No JAX implementation for the given distribution: truncated_normal`

ricardoV94: The dispatch file needs to be imported when pymc is imported in order to be registered

pymc/dispatch/dispatch_jax.py
```diff
@@ -0,0 +1,28 @@
...
+
+from pytensor.link.jax.dispatch import jax_funcify
+
+from pymc.distributions.continous import TruncatedNormalRV
+
+
+@jax_funcify.register(TruncatedNormalRV)
+def jax_funcify_TruncatedNormalRV(op, **kwargs):
+    def trunc_normal_fn(key, size, mu, sigma, lower, upper):
+        return None, jax.random.truncated_normal(
```
ricardoV94: Mu and sigma missing and the split rng should be returned, not None. 

Check some of the dispatches in PyTensor for a template

ricardoV94 (changes requested):

pymc/dispatch/dispatch_jax.py
HarshvirSandhu: I am using `jax.nn.initializers.truncated_normal` now, but the tests still fail. Not sure if  I have used the `rng` parameter correctly in tests

pymc/dispatch/dispatch_jax.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@jax_funcify.register(TruncatedNormalRV)
+def jax_funcify_TruncatedNormalRV(op, **kwargs):
+    def trunc_normal_fn(key, size, mu, sigma, lower, upper):
+        rng_key = key["jax_state"]
+        rng_key, sampling_key = jax.random.split(rng_key, 2)
+        key["jax_state"] = rng_key
+
+        truncnorm = jax.nn.initializers.truncated_normal(sigma, lower=lower, upper=upper)
```
ricardoV94: Why not this? https://jax.readthedocs.io/en/latest/_autosummary/jax.random.truncated_normal.html

tests/dispatch/test_jax.py
```diff
@@ -0,0 +1,38 @@
...
```
ricardoV94: The two are not expected to match in values, because JAX uses a different implementation than numpy. You can make a TruncatedNormal with a large sigma, and confirm it does not go beyond the bounds as a check

pymc/dispatch/dispatch_jax.py
```diff
@@ -0,0 +1,32 @@
...
+
+        return key, truncnorm(key["jax_state"], size) + mu
```
ricardoV94: Adding mu like this is potentially wrong, because when size is None, mu could be larger and we end up with repeated values

ricardoV94: Also you should dispatch on the more specific `jax_sample_fn`. For the issue with broadcasting, check how we do it here for Normal for example: https://github.com/pymc-devs/pytensor/blob/5d4b0c4b9a1e478dda48e912ee708a9e557e9343/pytensor/link/jax/dispatch/random.py#L147-L173

pymc/dispatch/dispatch_jax.py
HarshvirSandhu: We can't pass sigma or mu as parameters in `jax.random.truncated_normal`

HarshvirSandhu: I was able to use `jax.random.truncated_normal`, had to transform `lower` and `upper`

Issue #3661, opened 2019-10-18
Resuming a multi-process/chain run


yanirs:
Looking through #143 and #292, it seems like `iter_sample()` is the proposed way to resume sampling from an existing trace. However, it looks like `iter_sample()` doesn't support multi-processes, so it only samples from a single chain. Can it be extended to sample from multiple chains in parallel, as done by `sample()`?

mathDR:
are there plans on the roadmap for this request?  Or is this something that hasn't been scoped yet?

fonnesbeck:
There are no plans for it at the moment. It is a feature that we get requests for from time to time, but has never been taken up by anyone AFAIK.

Issue #292, opened 2013-06-13
Ability to work with chains in progress


aarchiba:
One rarely knows ahead of time how long one should run one's chains. So it would be really nice if one could start a chain running, perhaps on a cluster node, with no fixed end time. In this ideal world, one could then load the progress-so-far into another process and use pymc's usual tools to examine the chain. 

This appears to be almost possible: you can write a loop that runs the chain for a modest number of steps, then commit()s the database to disk, then starts a new chain. The new chain begins at the place the old chain left off, so one doesn't have to restart the burn-in process. It's not clear to me, from either the documentation or the code, whether the various tuning processes retain their state across this interruption. If they do, then this is almost a solution. I say "almost" because it's also not clear whether .commit() is supposed to produce a consistent database (in my tests with the txt backend it seems that it sometimes doesn't).

Alternatively, if the database reliably saves the step method's state, one could close the database then reopen it and start a new chain.

But wouldn't it be nice if the sampling process had a checkpointing function? Something like isample's pause function, that saved the database in a consistent state I could copy elsewhere and use.

fonnesbeck:
We have had requests similar to this before (e.g. #143), though chains without a pre-defined sampling length is a new one. I think it is clear that PyMC 3 ought to be able to continue sampling from the point where the last chain left off, and be able to save and restore the state of the sampler.

One idea that we might consider is to monitor the Gelman-Rubin statistic at regular intervals during sampling (of multiple chains) and provide some notification when all stochastics are within some distance of Rhat=1. From this it would be easy to implement something that continues to sample until convergence criteria are met.

Note also that the Hamiltonian MC step methods in PyMC 3 should help deal with convergence issues in a lot of large problems. I encourage you to try them out as we get closer to a beta release.

twiecki:
Implemented via `iter_sample()`

Issue #143, opened 2013-01-06
Resuming a MCMC run


kyleabeauchamp:
I noticed that various discussion threads have requested the ability to continue a MCMC run.  By continue, I mean that after calling sample(), one could call sample again and append the results to generate a longer chain.  

I realize that there is the "interactive" sampler, but to me it would be nice for all sampling to have a hot restart / append capabilities.  

Feel free to close this if this has already been discussed and deemed undesirable.

fonnesbeck:
Hey Kyle,

You cannot resume a chain with `sample()`, but you can sample additional chains that will be started where the last chain left off when using `sample()` serially, then simply concatenating the chains together. PyMC3 will run multiple chains in parallel by default, which can then be treated as a single sample if desired for inference. 

I will add the enhancement tag to this issue, so that restarting is universal for PyMC3.

jsalvatier:
we have this now

PR #7508, opened 2024-09-18
Add step method state and make step results deterministic with respect to it

### LINKED ISSUE #5797: Refactor step methods to use their own random stream

This issue is a spin off from a discussion in #5787. At the moment of that PR, our step methods rely on the global numpy random state to generate random numbers. It would be great to modify the step methods to allow them to have their own random generator that could then be used to generate all of their random number proposals.

Surprisingly, [metropolis](https://github.com/pymc-devs/pymc/blob/main/pymc/step_methods/metropolis.py) has some proposal distributions that could in principle take a random stream, the problem is that the actual steppers don't expose an easy to access random stream generator argument for users (or `pm.sample`) to set. HMC on the other hand is plagued with calls to `numpy.random.some_method`

### LINKED ISSUE #7503: ENH: Add checkpoints during sampling
### Before

_No response_

### After

```python
with pm.Model():
    ...
    pm.sample(..., checkpoint_file=some_path, checkpoint_freq=10)
```

### Context for the issue:

If one has models that take very long to sample, it would be great to have a way to store the information of the `steppers` in a checkpoint file so that if something happens and sampling stops, we could pick up from where we left off. This is a very old feature request that is related to #292, #143 and #3661.

Those issues talk about `iter_sample` that works as a generator that one could simply pause and resume later. The problem with that is that there is no access to the stepper's state. I think that we need two things to get the samplers warm started:

1. The trace that was collected so far
2. The step method's state

Currently, most samplers and step methods provide some ways to get 1 but we never have access to 2. The current pymc samplers have a bunch of `KeyboardInterrupt` catches ([here](https://github.com/pymc-devs/pymc/blob/main/pymc/sampling/mcmc.py#L1177), [here](https://github.com/pymc-devs/pymc/blob/main/pymc/sampling/population.py#L438), [here](https://github.com/pymc-devs/pymc/blob/main/pymc/sampling/parallel.py#L169), and [here](https://github.com/pymc-devs/pymc/blob/main/pymc/sampling/parallel.py#L189)). We could add a handling call there to also store the step method's state. `nutpie` has the [non-blocking sampling](https://github.com/pymc-devs/nutpie/blob/main/python/nutpie/sample.py#L466) with an `abort` function call when `KeyboardInterrupt` gets hit. We could maybe add a similar state recording thing there. `blackjax` has [its progress bar conditional steps](https://github.com/blackjax-devs/blackjax/blob/main/blackjax/progress_bar.py) which we could try to mimic to get the same effect. `numpyro` has a similar thing going with the [progress bar](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/util.py#L368) but it looks like it's way deeper than with `blackjax`.

All of this to say that I think that we need to define some kind of standard way for the samplers to provide their state information. The specific samplers would then have to conform to the standard using whatever internal things they need. For `pymc` samplers it would be some way to recreate the step methods (maybe using some kind of `__setstate__` and `__getstate__`), for `nutpie` it would have to be some new datatype that could be sent into ruff, for `blackjax` it could be the kernel and random keys. I think that the important thing is to get the standard approach to which samplers should conform to, and once we have those, we could build support for checkpoints and restarting sampling from them later.

lucianopaz:
## Description

The original intent of this PR was to fully address #7503. That proved to be a very long task, so the current PR focuses only on closing #5797 and to add `sampling_state` to all pymc step methods. I'll leave the past description further down.

## Related Issue

- [X] Closes #5797
- [X] Related to #143, #292, #3661, and #7503

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7508.org.readthedocs.build/en/7508/

## Original intent (**OUTDATED**)

This will be a long PR, but I want to open the discussion of the design in its early phases. The overall goal is to provide the ability to pause and later resume sampling that is based on pymc step methods. Once this PR is finished, I hope that we'll get into the problem of adding this ability when sampling with `blackjax`, `nutpie` and `numpyro`.

There will be 4 subgoals in this PR. I'll write them down and list the tasks that I'll do in each:

1. Write something that can dump or load the step method's sampling state.
  - [X] Create a class that represents the sampling state
  - [X] Get to dump or load the state for metropolis step methods
  - [x] Get to dump or load the state for compound step methods
  - [x] Get to dump or load the state for slice step methods
  - [x] Get to dump or load the state for HMC step methods
  - [x] Get to dump or load the state for NUTS step methods
  - [x] Ensure sampling state completely determines the step method's results
2. Write something that can dump or load the trace in its intermediate stages.
  - [ ] Get to dump or load the trace for NDArray backend
  - [ ] Get to dump or load the trace for McBackend (this might be done already?)
  - [ ] Get to dump or load the state for Arviz backend.
3. Add a way to restart sampling using previous sampling states and traces
  - [ ] Single chain MCMC
  - [ ] Parallel MCMC
  - [ ] Population sampling
4. Add async sampling methods (like nutpie's non blocking)
  - [ ] Single chain MCMC
  - [ ] Parallel MCMC
  - [ ] Population sampling


lucianopaz:
I think that this PR now has a good coverage for the first subgoal that I talked about in the description. All of pymc step methods now expose a `sampling_state` property that fully determines the following draws that the sampler will get. To do this, and also to test it, I had to do one **major change: I had to make all of the step methods become detached from numpy's global random state**. This doesn't bear significant consequences for an average user, but it opens up many possibilities:

- samplers could run concurrently if they use different step methods
- to ensure reproducibility, one doesn't need to affect the global state, which might have consequences on other packages that are out of pymc's scope
- the `init_nuts` step method needs to be adapted down the line to use random generators instead of a list of random seeds.

I'll ping some people to get some feedback on the current situation, because I've invested quite a lot of time into getting it right (and also have a clean git history), and I would like to get reviews before moving on. Pinging @ricardoV94, @michaelosthege, @Armavica, @ColCarroll and @junpenglao.

pymc/step_methods/arraystep.py
```diff
@@ -41,11 +39,12 @@ class ArrayStep(BlockedStep):
     blocked: Boolean (default True)
     """
 
-    def __init__(self, vars, fs, allvars=False, blocked=True):
+    def __init__(self, vars, fs, allvars=False, blocked=True, rng: RandomGenerator = None):
```
ricardoV94, line 46:
```suggestion
    def __init__(self, vars, fs, allvars=False, blocked=True, rng: RandomGenerator | None = None):
```
Also in other places

pymc/step_methods/hmc/base_hmc.py
```diff
@@ -52,12 +58,26 @@ class HMCStepData(NamedTuple):
...
+    _num_divs_sample: int
+
+
+class BaseHMC(GradientSharedStep, WithSamplingState):
     """Superclass to implement Hamiltonian/hybrid monte carlo."""
 
     integrator: integration.CpuLeapfrogIntegrator
     default_blocked = True
 
+    _state_class = BaseHMCState
```
ricardoV94, line 80: Why is it an underscore property?

pymc/step_methods/hmc/base_hmc.py
```diff
@@ -52,12 +58,26 @@ class HMCStepData(NamedTuple):
...
+    iter_count: int
+    step_size: np.ndarray
+    step_adapt: StepSizeState
+    target_accept: float
+    tune: bool
+    potential: PotentialState
+    _num_divs_sample: int
+
+
+class BaseHMC(GradientSharedStep, WithSamplingState):
```
ricardoV94: Why do we need a separate class? Why not be part of the baseclass?

pymc/step_methods/state.py
```diff
@@ -0,0 +1,98 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+from copy import deepcopy
+from dataclasses import dataclass, fields
+from typing import Any
+
+import numpy as np
+
+
+class MetaDataClassState(type):
```
ricardoV94: What is this doing?

pymc/step_methods/state.py
```diff
@@ -0,0 +1,98 @@
...
+    def __new__(cls, name, bases, namespace):
+        return dataclass(eq=False, kw_only=True)(super().__new__(cls, name, bases, namespace))
+
+
+class DataClassState(metaclass=MetaDataClassState):
```
ricardoV94: Why is this needed? Can we do simpler than this? 

Can we just use a vanilla dataclass?

pymc/step_methods/state.py
```diff
@@ -0,0 +1,98 @@
...
+        ), f"Encountered invalid state class '{state.__class__}'. State must be '{state_class}'"
+        for field in fields(state_class):
+            state_val = deepcopy(getattr(state, field.name))
+            self_val = getattr(self, field.name)
+            is_inmutable = field.metadata.get("inmutable", False)
+            if is_inmutable:
+                if not state.compare_values(state_val, self_val):
+                    raise ValueError(
+                        "The received sampling state must have the same values for the "
+                        f"inmutable fields. Field {field.name!r} has different values. "
```
ricardoV94: Rename inmutable to static/frozen?

pymc/step_methods/state.py
```diff
@@ -0,0 +1,98 @@
...
+        for field in fields(state_class):
+            val = getattr(self, field.name)
+            if isinstance(val, WithSamplingState):
+                _val = val.sampling_state
+            else:
+                _val = val
+            kwargs[field.name] = deepcopy(_val)
+        return state_class(**kwargs)
+
+    @sampling_state.setter
```
ricardoV94, line 76: When is the setter used?

pymc/util.py
```diff
@@ -578,3 +580,9 @@ def update(
                 **fields,
             )
         return None
+
+
+def get_random_generator(seed: RandomGenerator = None, copy=True) -> np.random.Generator:
+    if copy:
+        seed = deepcopy(seed)
```
ricardoV94, line 635: When/why is copying the seed needed?

ricardoV94 (commented):
Hi Luciano, had a quick glance. Looks alright but I wonder if we can make it a tad simpler by using dataclasses instead of reinventing something that seems similar?

I left other questions as comments.

Thanks for the initiative, making step methods less opaque will be great for customizing/controlling sampling

ricardoV94:
> to ensure reproducibility, one doesn't need to affect the global state, which might have consequences on other packages that are out of pymc's scope

This is a desired feature not a problem. There's an old issue that you can link to/ close if we get rid of the global RNG with this PR

> samplers could run concurrently if they use different step methods

How? Don't they always require interleaving steps (ie conditoned on a valid state?)

lucianopaz:
> Hi Luciano, had a quick glance. Looks alright but I wonder if we can make it a tad simpler by using dataclasses instead of reinventing something that seems similar?
> 
> I left other questions as comments.
> 
> Thanks for the initiative, making step methods less opaque will be great for customizing/controlling sampling

Thanks for the review @ricardoV94! I am using `dataclass`. The original approach was to add the `@dataclass` decorator to every state class. The problem there was that `__eq__` didn't work well with numpy arrays or with random number generators. That's why my second approach was to have a base class (`DataClassState`) that was a `dataclass` with an `__eq__` method. The problem with that approach was that I also had to add the `@dataclass(eq=False)` decorator to every state class that inherited from that common base. While working on this, I frequently forgot to add either the decorator or the `eq=False`, leading to failures down the line. That's why, in the final and current approach, I have a metaclass that will call `dataclass(eq=False)(cls)` on any class that inherits from the `DataClassState` base class. That way, I can ensure that all child classes are handled as dataclasses, and also use the base class `__eq__` that handles the weird array and random generator types. One last thing was that I also had to add `kw_only` to the dataclass, to enable safe inheritance between the classes. If I didn't, some classes that had default values for some fields could not be used as ancestors because the auto generated `__init__` would put those before other mandatory positional arguments.

pymc/step_methods/arraystep.py
lucianopaz: The `RandomGenerator` type already includes `None`

pymc/step_methods/hmc/base_hmc.py
lucianopaz: I honestly have no answer for this. Some people have the convention of using underscore to represent protected class members, meaning that they are not intended to be part of the public API. The idea being that the public API is supposed to be stable and usable by other devs without fear of having it changed without much notice. I just wanted to convey that `_state_class` is something that people shouldn't touch or try to modify in their own code.

pymc/step_methods/hmc/base_hmc.py
lucianopaz: Which baseclass? `WithSamplingState`? Or `BaseHMCState`? If you mean the latter, it's because the `BaseHMC` step method has properties that are different from other step methods, so I need to represent its state differently. If you mean the `WithSamplingState`, that's because the `WithSamplingState` provides the `sampling_state` property accessors to the step method. Let me know if you meant something else.

pymc/step_methods/state.py
lucianopaz: I already said it somewhere else but I'll add it here too. I wanted to make the step method states simple `dataclass` wrapped classes. I ran into problems revolving around `__eq__` and then also around positional arguments being defined after arguments with default values. To avoid these problems, I had to add some arguments to the `dataclass` decorator and I would sometimes forget them, leading to errors down the line. The simplest solution that didn't involve having to always write down boilerplate code with every `State` definition was to add this metaclass. It simply creates the subclass type and then wraps that using the `dataclass(eq=False, kw_only=True)` decorator. That way, you just need to inherit from `DataClassState` and you'll get the guarantee of working with a `dataclass` that has the specially crafted `__eq__` method.

pymc/step_methods/state.py
lucianopaz: I already answered this above. This solution avoids having to add boilerplate code all around the `State` definitions. The main problem is that if you have a class that inherits from another class that uses the `@dataclass` decorator, the subclass wont be a "`dataclass`". With the metaclass approach, the subclasses will also be "`dataclass`" types and they will use the nice `__eq__` method that I had to write here.

pymc/step_methods/state.py
lucianopaz: I don't know about this. Don't you think inmutable is understandable?

pymc/step_methods/state.py
lucianopaz: At the moment, it's only used in the tests. Calling `step_method.sampling_state = state` will set the step method's attributes to the proper values represented by the provided `state`. The goal is that we'll use it when we want to jump start the sampler to some past sampling state. The workflow I have in mind is this:
1. Have the model already built
2. Have the samplers already built
3. If someone provides a past sampling state from which to start, set the sampler's state to that
4. Start sampling using the sampler's set state and collect the results

pymc/util.py
lucianopaz: When `seed` is a `numpy.random.Generator`. If you don't do that, `numpy.random.default_rng(seed)` will return `seed`. If you provide a `BitGenerator`, the `Generator` object itself will be new, but its `BitGenerator` will be the same object that you passed in, making it potentially shared with another `Generator`. To ensure that those two scenarios wont happen, I `deepcopy` the `seed` by default.

pymc/step_methods/hmc/base_hmc.py
ricardoV94: I mean all step methods should have whatever WithSamplingState implements in the base class

pymc/step_methods/state.py
ricardoV94: I think the other terms I suggested are more common in the python ecosystem. Or use the non-negative `mutable` as the base case

pymc/step_methods/state.py
ricardoV94: I prefer an explicit method for that functionality? Something like `step_method.set_state(state)`?

It's 100% subjective opinion though

pymc/util.py
ricardoV94: Add an explicit instance check for Generator? And/or comment?

pymc/step_methods/hmc/base_hmc.py
lucianopaz: Oh, I think you’re right. It must be a left over from a previous state of my commits

pymc/util.py
lucianopaz: I’ll add a comment but I found that numpy does all of the hard type checking work and it felt like a waste to repeat

lucianopaz:
> Why do we need eq?

For convenience. If we need to assert equality, it’s much better to have this method

pymc/step_methods/state.py
lucianopaz: Ok, I’ll switch it frozen

ricardoV94:
> > Why do we need eq?
> 
> For convenience. If we need to assert equality, it’s much better to have this method 

Why not wait until we see a need then?

pymc/step_methods/state.py
lucianopaz: I agree to disagree on this one. The `set_state` method looks like any other method call and it’s not explicitly saying that it won’t have any return value. The attribute assignment syntax on the other hand is much more explicit on its intent.

lucianopaz:
> > > Why do we need eq?
> > 
> > 
> > For convenience. If we need to assert equality, it’s much better to have this method
> 
> Why not wait until we see a need then?

I did need it in all of the tests I wrote

ricardoV94:
> > > > Why do we need eq?
> > > 
> > > 
> > > For convenience. If we need to assert equality, it’s much better to have this method
> > 
> > Why not wait until we see a need then?
> 
> I did need it in all of the tests I wrote

That's more an argument for a test utility than code we need to strictly maintain

pymc/step_methods/hmc/base_hmc.py
michaelosthege: I would consider a mixin to be a cleaner design, because it doesn't make third party step methods inherit `WithSamplingState` even if they aren't compatible. It should also make things easier to test by not introducing cross-dependencies.

tests/step_methods/test_metropolis.py
```diff
@@ -292,7 +295,7 @@ def test_step_discrete(self):
         unc = np.diag(C) ** 0.5
         check = (("x", np.mean, mu, unc / 10.0), ("x", np.std, unc, unc / 10.0))
         with model:
-            step = Metropolis(S=C, proposal_dist=MultivariateNormalProposal)
+            step = Metropolis(S=C, proposal_dist=MultivariateNormalProposal, rng=123456)
```
michaelosthege, line 309: Why is this seed different?

tests/step_methods/test_metropolis.py
```diff
@@ -36,6 +36,8 @@
 from tests.helpers import RVsAssignmentStepsTester, StepMethodTester
 from tests.models import mv_simple, mv_simple_discrete, simple_categorical
 
+SEED = sum(ord(c) for c in "test_metropolis")
```
michaelosthege, line 49: but why 😂

pymc/step_methods/state.py
```diff
@@ -0,0 +1,98 @@
...
+    def __new__(cls, name, bases, namespace):
+        return dataclass(eq=False, kw_only=True)(super().__new__(cls, name, bases, namespace))
+
+
+class DataClassState(metaclass=MetaDataClassState):
+    def __eq__(self, other: Any) -> bool:
+        if self.__class__ != other.__class__:
+            return False
+        this_fields = set([f.name for f in fields(self)])
+        other_fields = set([f.name for f in fields(other)])
```
michaelosthege: inner list comprehension is unnecessary

pymc/step_methods/state.py
```diff
@@ -0,0 +1,98 @@
...
+            return all(DataClassState.compare_values(v1[k], v2[k]) for k in v1)
+        elif isinstance(v1, np.ndarray):
+            return bool(np.array_equal(v1, v2, equal_nan=True))
+        elif isinstance(v1, np.random.Generator):
+            return DataClassState.compare_values(v1.bit_generator.state, v2.bit_generator.state)
+        else:
+            return v1 == v2
+
+
+class WithSamplingState:
```
michaelosthege, line 52: Can you add docstrings for the three new classes to explain how they fit together?

For `WithSamplingState` I understand that it's a mixin adding a `sampling_state` property which, upon access, returns a new container object of a `DataClassState` subtype. This container holds copies of field values of the `WithSamplingState` object. (?)

michaelosthege (commented):
Great work! The refactoring of sampler RNGs could be extracted into its own PR and merged first?

Item 4. of your description sounds unrelated.

If I understand correctly, your approach with the mixin class has the nice benefit of not adding overhead to every iteration!

Regarding item 2. (where to dump traces) I previously pointed at the `stats`, because they already contain some/many state fields. In McBackend the stats can be sparse>, so one could emit a `sampler_state` state every 100 iterations or so.
* Last time I checked ArviZ/xarray didn't support saving sparse arrays to disk. But that's for sure workaroundable at the ArviZ level.
* `ClickHouseBackend` can persist sparse stats already.

pymc/step_methods/hmc/base_hmc.py
ricardoV94: I don't see any compatibility issues it a step sampler spent make use of the functionality in the base class.

ricardoV94:
> If I understand correctly, your approach with the mixin class has the nice benefit of not adding overhead to every iteration!

What overhead? There's nothing computed every iteration

pymc/step_methods/hmc/base_hmc.py
lucianopaz: Oh, I see what @michaelosthege is saying. The current design is more inline with what @ricardoV94 said, I already have a `BlockedStep` as a subclass of `WithSamplingState`, so including it in `BaseHMC` ancestors explicitly is pointless. The base `StepMethodState` only tries to access the `rng` property. I could provide a default value factory for that property, and that way there won't be any problem for third party libraries that define their own step methods. They would simply get a useless random number `Generator` object. They could use it if they wanted to though.

tests/step_methods/test_metropolis.py
lucianopaz: Seeding these tests was a PITA. I kept running into sporadic errors and flakiness while I was polishing the step method detachment from the global random state, and I some rng's were left with the weird intermediate seeds.

tests/step_methods/test_metropolis.py
lucianopaz: Same as above. I can only answer with an argentinian meme

imagen

pymc/step_methods/state.py
lucianopaz: Why not? The `fields` function will return `Field` objects, that have a bunch of extra dataclass specific attributes (e.g. `type`, `default`, `default_factory`). I just want to check that the names are the same and use those names later for `getattr`.

lucianopaz: Oh! I think I understand what you're saying. A `set` of `Field` objects should already be enough to test if `this_fields == other_fields` because all of the other attributes should also match.

pymc/step_methods/state.py
lucianopaz: That's exactly right. I'll add comments and docstrings.

lucianopaz:
> Great work! The refactoring of sampler RNGs could be extracted into its own PR and merged first?

Thanks @michaelosthege! I think that it could be extracted. I would need to add docstring entries for `rng` first though.

> Regarding item 2. (where to dump traces) I previously pointed at the `stats`, because they already contain some/many state fields. In McBackend the stats can be sparse>, so one could emit a `sampler_state` state every 100 iterations or so.
> 
>     * Last time I checked ArviZ/xarray didn't support saving sparse arrays to disk. But that's for sure workaroundable at the ArviZ level.
> 
>     * `ClickHouseBackend` can persist sparse stats already.

Thanks for the pointers. I'll try to use that once I arrive at point 2

lucianopaz:
> > If I understand correctly, your approach with the mixin class has the nice benefit of not adding overhead to every iteration!
> 
> What overhead? There's nothing computed every iteration

I think that Michael means that I'm not building a `StepMethodState` at each step. The state is a property that can be built on demand, but there is no extra compute involved in regular sampling. The step method just chugs along until we eventually call `step.sampling_state` at some point during sampling. On the other hand, `stats` objects are built at each step, so they produce some overhead when compared to a situation in which only the samples got collected.

michaelosthege:
> On the other hand, `stats` objects are built at each step, so they produce some overhead when compared to a situation in which only the samples got collected.

The stats are just `dict`s and as far as I can tell you didn't change anything about how they are collected (alongside `draws`) in every iteration.

ricardoV94:
What's the problem with ndarray backend for draws?

And if we need something else why do we need distinct stuff for draws and step sampler info?

Possibly related to https://github.com/pymc-devs/pymc/issues/7300

pymc/step_methods/state.py
michaelosthege: actually that was not my point (but you might be right about it)

`set(generator)` aka `set(a for a in "ABCD")` works. You can leave out creating the inner list (and then iterating it again when creating the set)

michaelosthege:
> @michaelosthege, do you think mcbackend could serve as a suitable intermediate representation format for draws always? Something that then gets converted into NDArray or InferenceData backends?

> What's the problem with ndarray backend for draws?

[The NumPy backend in McBackend](https://github.com/pymc-devs/mcbackend/blob/main/mcbackend/backends/numpy.py) is essentially a modernized copy of PyMC's `NDArray` backend. Except with a cleaner initialization and interchangable with other backends.

> Another task that I haven't though of yet is how to dump the step method states. [...] do you think that Arrow could be a good format to move that state information around? Or should I have a look at protocol buffers?

We can also implement a PyArrow backend in McBackend. That's probably an afternoon of pair-programming for me + somebody with PyArrow prior.

I didn't propose to make McBackend's NumPyBackend the default in PyMC yet, because I didn't wanto to introduce a dependency on protocol buffers without y'all agreeing.
If we want a serialization of sampler state (rather than duping all state variables as sparse `sampler_stats`) I would propose to go for protocol buffers so PyArrow can remain optional. (but I'm biased, of course)

ricardoV94:
Sounds good to me, can you add a more informative PR title?

ricardoV94:
BTW I still favor going with dataclasses instead of the custom new classes and have the equality as a detached test utility. There is no functionality that depends on equality right now or in the foreseeable future, unless I missed something.

Complexity for testing purposes seems backwards to me.

ricardoV94:
BTW I still favor going with dataclasses instead of the custom new classes and have the equality as a detached test utility. There is no functionality that depends on equality right now or in the foreseeable future, unless I missed something.

Complexity for testing purposes seems backwards to me.

pymc/step_methods/state.py
```diff
@@ -0,0 +1,114 @@
...
+                return False
+        return this_fields == other_fields
+
+    @staticmethod
+    def compare_values(v1: Any, v2: Any) -> bool:
+        if v1.__class__ != v2.__class__:
+            return False
+        if isinstance(v1, (list, tuple)):  # noqa: UP038
+            return len(v1) == len(v2) and all(
+                DataClassState.compare_values(v1i, v2i) for v1i, v2i in zip(v1, v2)
```
ricardoV94: You could use zip(..., strict=True) and avoid explicitly comparing the length

pymc/step_methods/state.py
ricardoV94: So why is there a set_rng? The same argument you used against set_state would apply? Also, in my experience properties have always be a PITA because at some point we figure out we would like kwargs/customization and there's no way to refactor a property into a method with back_compat.

lucianopaz:
> BTW I still favor going with dataclasses instead of the custom new classes and have the equality as a detached test utility. There is no functionality that depends on equality right now or in the foreseeable future, unless I missed something.
> 
> Complexity for testing purposes seems backwards to me.

I think that I can have a workaround for part of this, but it really depends on what you mean by dataclasses. Is the problem that I'm using a metaclass? Is it that I'm defining an `__eq__`? Is the problem that I'm using inheritance between `DataClassState` and some of its subclasses?

If your problem is that I'm relying on metaclasses, I think that I can do something differently to avoid them. If the problem is the `__eq__`, I think that I can move that to the tests as a function. What I don't want to change is the inheritance from `DataClassState`. That is important for static type analysis and also helped with reducing the amount of duplicate code.

ricardoV94:
Yes my problem is the implementation of `__eq__` and `metaclass`. I don't see why we need this for what is basically a namedtuple. We hold RNG / np.ndarray in many kinds of objects and we don't usually go about implementing equality if we don't have to (we do for TensorConstants in PyTensor for example)

lucianopaz:
> Yes my problem is the implementation of `__eq__` and `metaclass`. I don't see why we need this for what is basically a namedtuple. We hold RNG / np.ndarray in many kinds of objects and we don't usually go about implementing equality if we don't have to (we do for TensorConstants in PyTensor for example)

@ricardoV94, I changed the code to avoid the `metaclass` and detached the `__eq__` from the code. I did had to leave a comparison utility function in the main codebase because I want to be able to compare `frozen` fields. Let me know if you think that's good enough. If it is, I'll clean up the last commit and we can merge. I chose to leave it dirty for now to make it easier to undo the change if we don't like it.

pymc/step_methods/state.py
lucianopaz: This actually wasn’t good because `strict=True` raises a `ValueError`, and I just want it to return `False`. I’ll keep the length comparison as it was

pymc/step_methods/state.py
lucianopaz: I added a `set_rng` method because the subclasses are supposed to overload it. The HMC step methods have their own `rng` and their `potential` also have a spawned `rng`. Setting the `rng` had to work differently than with the rest of the step methods, so I decided to make it a method instead of a `property`. Anyway, I still prefer the `sampling_state` as a property mixin. If we eventually realize that we don't want to, we can always add a new `set_state` method and issue a `DeprecationWarning` or `RuntimeError` in the property setter.

ricardoV94: Let me try one last argument. Setting a property (which looks like an attribute) does not make me intuitively think that it will actually affect the sampler. Specially in this case, where the property (sampler.sample_sate) is actually a read-only copy of the internal state, not the state itself. 

It's like if you have a PyMC model, which has the attribute `model.datalogp`. I wouldn't expect  `model.datalogp = 0` to be a valid way of overriding the model logp. Of course if I read the source code I'll be able to figure out, but from just reading use code I wouldn't think it would do what I want it to do.

I find `sampler.set_rng()` must more obvious that will actually affect the rng used. And `sampler.set_state()` that it will actually affect the state used.

Final argument I found, but not necessarily care about has to do with inheritance. Calling `super()` on a property is clumsy, if you want to combine the effects of the base class method and some tweaking in the inherited class.

ricardoV94:
@lucianopaz sounds good to me. Besides my personal preference for methods vs setter, I have one last suggestion and one question.

Rename `compare_dataclass_values`, compare_states` to `equal_dataclass_values`, `equal_states` (or whatever it was called, point being `compare` -> `equal`.

What's the deal with frozen fields? Why do we need them / to worry about them?

To be clear, I'm happy with the state and I am not blocking the merge after the rebase.

lucianopaz:
> Rename `compare_dataclass_values` and `compare_states` to `equal_dataclass_values`, `equal_states` (or whatever it was called, point being `compare` -> `equal`.

Good point, I'll do that.
 
> What's the deal with frozen fields? Why do we need them / to worry about them?

The step methods have a bunch of extra information in them that gets set when they are created. My very first idea was to try to have some kind of `pickle.dump` approach where the entire step method would be stored. The problem with that was that the step methods have references to model variables and compiled functions. Serializing the step method would then force us to also serialize the whole model instance along with some compiled functions. This extra burden made me think that it would be better to only store and set some small bits of information from the steppers. Part of this information changes as samples are drawn, and other parts don't.

The long term goal was to be able to set the step method to a state where it could continue sampling as it had been doing before. Since I wouldn't be able to rebuild the full step method from what I save to disk, I needed to add some way to determine that the stored state was compatible with the step method that was being read. That's why I decided to include some step information that doesn't change during sampling as `frozen` fields. If the saved state doesn't match with the stepper's frozen fields, that means that the state is not valid for the step method and an error should be raised.

ricardoV94 (approved):

PR #7514, opened 2024-09-25
Fix for issue 7369

### LINKED ISSUE #7369: Numpy deprecation warning issued from within `test_metropolis.py`

westford14:
Fix for numpy deprecation warning (#7369) 

## Description

Small PR that addresses the numpy deprecation warnings that are seen when running tests:

```
tests/step_methods/test_metropolis.py: 15950 warnings
  REDACTED/pymc/pymc/backends/ndarray.py:118: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    data[key][self.draw_idx] = val
```

Change looks at the type of the `val` and if it is a `np.ndarray` and has `.shape[0]` == 1, then we pop the item out to prevent the deprecation warning.

## Related Issue

- [X] Closes #7369 
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [X] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7514.org.readthedocs.build/en/7514/

### DIFF
--- a/pymc/backends/ndarray.py
+++ b/pymc/backends/ndarray.py
@@ -114,9 +114,13 @@ class NDArray(base.BaseTrace):
             raise ValueError("Unknown sampler_stats")
         if sampler_stats is not None:
             for data, vars in zip(self._stats, sampler_stats):
-                for key, val in vars.items():
-                    if isinstance(val, np.ndarray) and val.shape[0] == 1:
-                        val = val.item()
+                compressed_vars = {}
+                for k, v in vars.items():
+                    if isinstance(v, np.ndarray) and v.shape[0] == 1:
+                        compressed_vars[k] = v.item()
+                    else:
+                        compressed_vars[k] = v
+                for key, val in compressed_vars.items():
                     data[key][self.draw_idx] = val
         self.draw_idx += 1
 
--- a/pymc/backends/ndarray.py
+++ b/pymc/backends/ndarray.py
@@ -116,8 +116,8 @@ class NDArray(base.BaseTrace):
             for data, vars in zip(self._stats, sampler_stats):
                 compressed_vars = {}
                 for k, v in vars.items():
-                    if isinstance(v, np.ndarray) and v.shape[0] == 1:
-                        compressed_vars[k] = v.item()
+                    if isinstance(v, np.ndarray) and v.shape[0] == 1 and len(v.shape) == 1:
+                        compressed_vars[k] = v.reshape(1, 1)
                     else:
                         compressed_vars[k] = v
                 for key, val in compressed_vars.items():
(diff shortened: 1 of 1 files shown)


pymc/backends/ndarray.py
```diff
@@ -115,6 +115,8 @@ def record(self, point, sampler_stats=None) -> None:
         if sampler_stats is not None:
             for data, vars in zip(self._stats, sampler_stats):
                 for key, val in vars.items():
+                    if isinstance(val, np.ndarray) and val.shape[0] == 1:
```
ricardoV94: Can val have more than one dim? More importantly it would be good if we didn't need this check in every iteration/ for every stat. 

Can we solve it from another angle?

westford14: none of the tests passed in any values of greater than one dimension -- but noted on the point about trying to solve this outside the loop!

ricardoV94:
What happens in val has `shape==(1, 2)`?

Also the problem here seems to be the buffer was allocated with the wrong shape? If a stat has shape (1,) in every draw, the buffer should have (draws, 1) shape, not (draws,) and no coercing to scalar would be needed.

Am I off track here?

westford14:
> What happens if val has `shape==(1, 2)`? Your code will try to cerce into a scalar and fail right?
> 
> Taking a step back, the problem here seems to be the buffer was allocated with the wrong shape? If a stat has shape (1,) in every draw, the buffer should have (draws, 1) shape, not (draws,) and no coercing to scalar would be needed.
> 
> Am I off track here?

okay yea that makes sense -- i'm not sure if there's a better way to coerce here than trying to find arrays of shape (1,) because i don't think it makes sense to traverse the `sampler_stats` prior to this inner loop

ricardoV94:
We should look at the code where the stats arrays are allocated to see how it's deciding on the shape

ricardoV94:
The assumption that stats are scalars happens here: https://github.com/pymc-devs/pymc/blob/45069a966d3e123bf98efbc4f28008e31b243df8/pymc/backends/ndarray.py#L90

If that's a requirement, than the problem is in Metropolis not respecting. If that's not a requirement, the problem is in assuming the stats will be scalars. Which statistic from Metropolis step sampler is not respecting the scalar shape?

PR #7523, opened 2024-10-03
Remove release test

ricardoV94:
Failed last time probably because of python3.7 (should we bump the build python as well).

Removed as @maresb thinks it's useless.

----
📚 Documentation preview 📚: https://pymc--7523.org.readthedocs.build/en/7523/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,14 +16,6 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: 3.7
-    # TODO: ideally, this pipeline should run parallelized tests in upstream jobs..
-    #- name: Install test tooling
-    #  run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -16,14 +16,6 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: 3.7
-    # TODO: ideally, this pipeline should run parallelized tests in upstream jobs..
```
ricardoV94, line 19: I have no idea what this comment is about, to rerun the tests before uploading?

lucianopaz (approved):
This looks good to me. I'd never actually looked at this workflow but it is awful! It's using a very old python and it's also using twine?!? I think that pytensor has a much more modern approach to the upload process. Might be worth opening an issue

Issue #7528, opened 2024-10-08
Deprecate initval at the distribution level


ricardoV94:
### Description

initvals are not a real property of the model, but an aspect of the sampler. We already have a `initvals` kwarg in `pm.sample` that can be used to customize initial values when needed.

We can also allow extra kwargs to be passed from `model.initial_point` to `make_initial_point_fn` to allow the equivalent behavior of `initval=x` used in conjunction with `model.initial_point()`

Removing them would make our distribution and model objects simpler, and provide one consistent way of defining initvals.

On purpose, I did not make initvals part of the fgraph representation of models, which raises NotImplementedError when custom ones are set. This forces an error in utilities that make use of it: `model.copy()`, `pm.observe`, `pm.do` , etc...

If people really object to not removing them from the model, we have to go and make it part of the fgraph representation.
If we accept it, we should deprecate them for a while since they are an old API and used many times in conjunction with ordered transforms (which are mostly used because we don't have #7297 yet)

PR #7529, opened 2024-10-08
Do not use initval in test model

ricardoV94:
PRs https://github.com/pymc-devs/pymc/pull/7508 and https://github.com/pymc-devs/pymc/pull/7492 introduced incompatible changes but were not tested simultaneously.

Deepcopying the steps in the tests leads to deepcopying the model which uses `clone_model`, which in turn does not support initvals.

See #7528 why this is currently not implemented.

@lucianopaz irregardless of the issues with copying models with initvals, is there any concern for the step functionality that model.copy uses `clone_model` these days? AFAICT this was only used in the tests so a first pass I would guess no?

----
📚 Documentation preview 📚: https://pymc--7529.org.readthedocs.build/en/7529/

### DIFF
--- a/tests/models.py
+++ b/tests/models.py
@@ -18,7 +18,6 @@ import numpy as np
 import pytensor
 import pytensor.tensor as pt
 
-from pytensor import config
 from pytensor.compile.ops import as_op
 
 import pymc as pm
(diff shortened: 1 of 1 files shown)


twiecki (approved):

lucianopaz:
> @lucianopaz irregardless of the issues with copying models with initvals, is there any concern for the step functionality that model.copy, uses `clone_model` these days? AFAICT this was only used in the tests so a first pass I would guess no?

No, there should be no issue with that. The step methods `sampling_state` is independent from the model instance and its variables, so as long as copying the model doesn't break the relationship with the step method's `var`, then nothing should break.

PR #7532, opened 2024-10-10
Improve type hints

Armavica:
## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7532.org.readthedocs.build/en/7532/

### DIFF
--- a/pymc/_version.py
+++ b/pymc/_version.py
@@ -29,11 +29,11 @@ import os
 import re
 import subprocess
 import sys
-from typing import Any, Callable, Dict, List, Optional, Tuple
+from typing import Any, Callable
 import functools
 
 
-def get_keywords() -> Dict[str, str]:
+def get_keywords() -> dict[str, str]:
     """Get the keywords needed to look up the version information."""
     # these strings will be replaced by git during git-archive.
     # setup.py/versioneer.py will grep for the variable names, so they must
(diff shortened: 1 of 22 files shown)


pymc/distributions/custom.py
```diff
@@ -604,8 +602,8 @@ def logp(value: TensorVariable, mu: TensorVariable) -> TensorVariable:
 
         def random(
             mu: np.ndarray | float,
-            rng: Optional[np.random.Generator] = None,
-            size: Optional[Tuple[int]] = None,
+            rng: np.random.Generator | None = None,
```
ricardoV94, line 605: For the example: mu is always np.ndarray, rng is always provided never None

pymc/initial_point.py
```diff
@@ -93,21 +107,18 @@ def make_initial_point_fns_per_chain(
...
-                return_transformed=True,
-            )
-            for chain_overrides in overrides
-        ]
-    else:
-        raise ValueError(
-            f"Number of initval dicts ({len(overrides)}) does not match the number of chains ({chains})."
+        return ipfns
+
+    assert isinstance(overrides, Sequence) and len(overrides) == chains
```
ricardoV94, line 112: Already checked above

Armavica: It's also related to the meaning of `assert`. I see it less as a control flow, more as an executable comment. Converting this to a comment would leave the possibility that the comment goes out of date.
Is this function typically hot?
Also, asserts are disabled by `python -O`, perhaps this is something that we could do?

ricardoV94: I don't think anyone ever disables asserts tbh

pymc/_version.py
```diff
@@ -268,7 +268,7 @@ def git_pieces_from_vcs(
     root: str,
     verbose: bool,
     runner: Callable = run_command
```
0xs1d, line 270: we can callables type annotated too for greater clarity

PR #7540, opened 2024-10-16
Add ZarrTrace

### LINKED ISSUE #7612: Sampling with generators as seeding no longer deterministic
### Description

This was revealed in https://github.com/pymc-labs/pymc-marketing/pull/1269

```python
import numpy as np
import pymc as pm

with pm.Model() as m:
    x = pm.Normal("x")

    post1 = pm.sample(tune=10, draws=10, random_seed=np.random.default_rng(42)).posterior
    post2 = pm.sample(tune=10, draws=10, random_seed=np.random.default_rng(42)).posterior
assert post1.equals(post2), (post1["x"].mean().item(), post2["x"].mean().item())
# AssertionError: (0.22006495904628473, -0.31090965213192406)
```

lucianopaz:
## Description

This PR is related to #7503. It specifically focuses on having a way to store intermediate trace results and the step methods sampling state somewhere (See task 2 of #7508).

To be honest, the current situation of the `MultiTrace` and `NDArray` backends is terrible. These backend classes have inconsistent signatures across subclasses, and it's very awkward to write new backends that adhere to them. `McBackend` was an attempt to make things sane again. As far as I understand, `McBackend` does support ways to dump samples to disk instead of holding them in memory using the `ClickHouse` database. However, I found the backend a bit detached from arviz and xarray, and it seemed to be tightly linked to protocol buffers, which made it harder for me to see how I could customize stuff.

These considerations brought me to the approach I'm pursuing in this PR: **add a backend that uses [zarr](https://zarr.readthedocs.io/en/latest/getting_started.html)**. Using zarr has the following benefits:

1. `xarray` can read zarr stores directly making it possible to write `InferenceData` objects to disk directly almost without even having to call a converter.
2. zarr works with hierarchically structured data. It's possible to store arrays for each variable inside of a group (e.g. posterior, observed_data) directly.
3. zarr arrays handle numpy arrays nicely. Fixed sized binary data can be fit into zarr arrays seemlessly.
4. zarr arrays also have the possibility of storing `object` dtyped arrays using the [`numcodec`](https://numcodecs.readthedocs.io/en/stable/index.html) package. This makes it possible use the same store to hold sample stats warning objects and step methods `sampling_state` in the same place as the actual samples from the posterior.
5. zarr hierarchies can use many different kinds of [storage](https://zarr.readthedocs.io/en/latest/_autoapi/zarr/storage/index.html): they can be held in memory, saved as a directory structure, inside of a zip file, or even remotely on s3 buckets.
6. It's also possible to write to the same zarr object [concurrently](https://zarr.readthedocs.io/en/latest/tutorial.html#parallel-computing-and-synchronization) from different processes or threads, as long as a synchronization object is provided.
7. zarr also stores the data using a [compressed binary representation](https://zarr.readthedocs.io/en/stable/tutorial.html#compressors). The actual compressor can be customized.
8. zarr arrays are chunked. This means that they don't need to be loaded entirely onto memory, making it possible to leave a smaller memory footprint while sampling. Another benefit of chunking is that write operations on different chunks should be completely independent from each other.

Having stated all of these considerations I intend to:

- [x] Build a zarr trace backend
- [X] Have `ZarrTrace` integrate well with `pymc.sample`
- [X] WONT DO NOW ~~Replace the `MultiTrace` and `NDArray` backend defaults with their Zarr counterparts~~
- [X] Handle sampling state information in the zarr backend
- [x] Document `ZarrTrace`
- [x] Buffer write operations in `ZarrChain.record`
- [x] Record sampling state information periodically during sampling
- [X] NEED TO DO IN SEPARATE PR ~~Make it possible to load the zarr trace backend and resume sampling from it.~~

## Related Issue

- [X] Closes #7612 
- [X] Related to #7503

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7540.org.readthedocs.build/en/7540/

### DIFF
--- a/pymc/backends/zarr.py
+++ b/pymc/backends/zarr.py
@@ -11,7 +11,7 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
-from collections.abc import Mapping, MutableMapping, Sequence
+from collections.abc import Callable, Mapping, MutableMapping, Sequence
 from typing import Any
 
 import arviz as az
@@ -91,10 +91,11 @@ class ZarrChain(BaseTrace):
         vars: Sequence[TensorVariable] | None = None,
         test_point: dict[str, np.ndarray] | None = None,
         draws_per_chunk: int = 1,
+        fn: Callable | None = None,
     ):
         if not _zarr_available:
             raise RuntimeError("You must install zarr to be able to create ZarrChain instances")
-        super().__init__(name="zarr", model=model, vars=vars, test_point=test_point)
+        super().__init__(name="zarr", model=model, vars=vars, test_point=test_point, fn=fn)
         self._step_method: BlockedStep | CompoundStep | None = None
         self.unconstrained_variables = {
             var.name for var in self.vars if is_transformed_name(var.name)
@@ -168,7 +169,7 @@ class ZarrChain(BaseTrace):
         :meth:`~ZarrChain.flush`
         """
         unconstrained_variables = self.unconstrained_variables
-        for var_name, var_value in zip(self.varnames, self.fn(draw)):
+        for var_name, var_value in zip(self.varnames, self.fn(**draw)):
             if var_name in unconstrained_variables:
                 self.buffer(group="unconstrained_posterior", var_name=var_name, value=var_value)
             else:
@@ -452,13 +453,18 @@ class ZarrTrace:
         )
         self.vars = [var for var in vars if var.name in self.varnames]
 
-        self.fn = model.compile_fn(self.vars, inputs=model.value_vars, on_unused_input="ignore")
+        self.fn = model.compile_fn(
+            self.vars,
+            inputs=model.value_vars,
+            on_unused_input="ignore",
+            point_fn=False,
+        )
 
         # Get variable shapes. Most backends will need this
         # information.
         if test_point is None:
             test_point = model.initial_point()
-        var_values = list(zip(self.varnames, self.fn(test_point)))
+        var_values = list(zip(self.varnames, self.fn(**test_point)))
         self.var_dtype_shapes = {
             var: (value.dtype, value.shape)
             for var, value in var_values
@@ -528,6 +534,7 @@ class ZarrTrace:
                 test_point=test_point,
                 stats_bijection=StatsBijection(step.stats_dtypes),
                 draws_per_chunk=self.draws_per_chunk,
+                fn=self.fn,
             )
             for _ in range(chains)
         ]
--- a/pymc/backends/zarr.py
+++ b/pymc/backends/zarr.py
@@ -1,0 +1,846 @@
+#   Copyright 2024 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+from collections.abc import Mapping, MutableMapping, Sequence
+from typing import Any
+
+import arviz as az
+import numcodecs
+import numpy as np
+import xarray as xr
+import zarr
+
+from arviz.data.base import make_attrs
+from arviz.data.inference_data import WARMUP_TAG
+from numcodecs.abc import Codec
+from pytensor.tensor.variable import TensorVariable
+
+import pymc
+
+from pymc.backends.arviz import (
+    coords_and_dims_for_inferencedata,
+    find_constants,
+    find_observations,
+)
+from pymc.backends.base import BaseTrace
+from pymc.blocking import StatDtype, StatShape
+from pymc.model.core import Model, modelcontext
+from pymc.step_methods.compound import (
+    BlockedStep,
+    CompoundStep,
+    StatsBijection,
+    get_stats_dtypes_shapes_from_steps,
+)
+from pymc.util import UNSET, _UnsetType, get_default_varnames, is_transformed_name
+
+try:
+    from zarr.storage import BaseStore, default_compressor
+    from zarr.sync import Synchronizer
+
+    _zarr_available = True
+except ImportError:
+    _zarr_available = False
+
+
+class ZarrChain(BaseTrace):
+    """Interface object to interact with a single chain in a :class:`~.ZarrTrace`.
+
+    Parameters
+    ----------
+    store : zarr.storage.BaseStore | collections.abc.MutableMapping
+        The store object where the zarr groups and arrays will be stored and read from.
+        This store must exist before creating a ``ZarrChain`` object. ``ZarrChain`` are
+        only intended to be used as interfaces to the individual chains of
+        :class:`~.ZarrTrace` objects. This means that the :class:`~.ZarrTrace` should
+        be the one that creates the store that is then provided to a ``ZarrChain``.
+    stats_bijection : pymc.step_methods.compound.StatsBijection
+        An object that maps between a list of step method stats and a dictionary of
+        said stats with the accompanying stepper index.
+    synchronizer : zarr.sync.Synchronizer | None
+        The synchronizer to use for the underlying zarr arrays.
+    model : Model
+        If None, the model is taken from the `with` context.
+    vars : Sequence[TensorVariable] | None
+        Sampling values will be stored for these variables. If None,
+        `model.unobserved_RVs` is used.
+    test_point : dict[str, numpy.ndarray] | None
+        This is not used and is inherited from the signature of :class:`~.BaseTrace`,
+        which uses it to determine the shape and dtype of `vars`.
+    draws_per_chunk : int
+        The number of draws that make up a chunk in the variable's posterior array.
+        The interface only writes the samples to the store once a chunk is completely
+        filled.
+    """
+
+    def __init__(
+        self,
+        store: BaseStore | MutableMapping,
+        stats_bijection: StatsBijection,
+        synchronizer: Synchronizer | None = None,
+        model: Model | None = None,
+        vars: Sequence[TensorVariable] | None = None,
+        test_point: dict[str, np.ndarray] | None = None,
+        draws_per_chunk: int = 1,
+    ):
+        if not _zarr_available:
+            raise RuntimeError("You must install zarr to be able to create ZarrChain instances")
+        super().__init__(name="zarr", model=model, vars=vars, test_point=test_point)
+        self._step_method: BlockedStep | CompoundStep | None = None
+        self.unconstrained_variables = {
+            var.name for var in self.vars if is_transformed_name(var.name)
+        }
+        self.draw_idx = 0
+        self._buffers: dict[str, dict[str, list]] = {
+            "posterior": {},
+            "sample_stats": {},
+        }
+        self._buffered_draws = 0
+        self.draws_per_chunk = int(draws_per_chunk)
+        assert self.draws_per_chunk > 0
+        self._posterior = zarr.open_group(
+            store, synchronizer=synchronizer, path="posterior", mode="a"
+        )
+        if self.unconstrained_variables:
+            self._unconstrained_posterior = zarr.open_group(
+                store, synchronizer=synchronizer, path="unconstrained_posterior", mode="a"
+            )
+            self._buffers["unconstrained_posterior"] = {}
+        self._sample_stats = zarr.open_group(
+            store, synchronizer=synchronizer, path="sample_stats", mode="a"
+        )
+        self._sampling_state = zarr.open_group(
+            store, synchronizer=synchronizer, path="_sampling_state", mode="a"
+        )
+        self.stats_bijection = stats_bijection
+
+    def link_stepper(self, step_method: BlockedStep | CompoundStep):
+        """Provide a reference to the step method used during sampling.
+
+        This reference can be used to facilite writing the stepper's sampling state
+        each time the samples are flushed into the storage.
+        """
+        self._step_method = step_method
+
+    def setup(self, draws: int, chain: int, sampler_vars: Sequence[dict] | None):  # type: ignore[override]
+        self.chain = chain
+        self.total_draws = draws
+        self.draws_until_flush = min([self.draws_per_chunk, draws - self.draw_idx])
+        self.clear_buffers()
+
+    def clear_buffers(self):
+        for group in self._buffers:
+            self._buffers[group] = {}
+        self._buffered_draws = 0
+
+    def buffer(self, group, var_name, value):
+        buffer = self._buffers[group]
+        if var_name not in buffer:
+            buffer[var_name] = []
+        buffer[var_name].append(value)
+
+    def record(
+        self, draw: Mapping[str, np.ndarray], stats: Sequence[Mapping[str, Any]]
+    ) -> bool | None:
+        """Record the step method's returned draw and stats.
+
+        The draws and stats are first stored in an internal buffer. Once the buffer is
+        filled, the samples and stats are written (flushed) onto the desired zarr store.
+
+        Returns
+        -------
+        flushed : bool | None
+            Returns ``True`` only if the data was written onto the desired zarr store.
+            Any other time that the recorded draw and stats are written into the
+            internal buffer, ``None`` is returned.
+
+        See Also
+        --------
+        :meth:`~ZarrChain.flush`
+        """
+        unconstrained_variables = self.unconstrained_variables
+        for var_name, var_value in zip(self.varnames, self.fn(draw)):
+            if var_name in unconstrained_variables:
+                self.buffer(group="unconstrained_posterior", var_name=var_name, value=var_value)
+            else:
+                self.buffer(group="posterior", var_name=var_name, value=var_value)
+        for var_name, var_value in self.stats_bijection.map(stats).items():
+            self.buffer(group="sample_stats", var_name=var_name, value=var_value)
+        self._buffered_draws += 1
+        if self._buffered_draws == self.draws_until_flush:
+            self.flush()
+            return True
+        return None
+
+    def record_sampling_state(self, step: BlockedStep | CompoundStep | None = None):
+        """Record the sampling state information to the store's ``_sampling_state`` group.
+
+        The sampling state includes the number of draws taken so far (``draw_idx``) and
+        the step method's ``sampling_state``.
+
+        Parameters
+        ----------
+        step : BlockedStep | CompoundStep | None
+            The step method from which to take the ``sampling_state``. If ``None``,
+            the ``step`` is taken to be the step method that was linked to the
+            ``ZarrChain`` when calling :meth:`~ZarrChain.link_stepper`. If this method was never
+            called, no step method ``sampling_state`` information is stored in the
+            chain.
+        """
+        if step is None:
+            step = self._step_method
+        if step is not None:
+            self.store_sampling_state(step.sampling_state)
+        self._sampling_state.draw_idx.set_coordinate_selection(self.chain, self.draw_idx)
+
+    def store_sampling_state(self, sampling_state):
+        self._sampling_state.sampling_state.set_coordinate_selection(
+            self.chain, np.array([sampling_state], dtype="object")
+        )
+
+    def flush(self):
+        """Write the data stored in the internal buffer to the desired zarr store.
+
+        After writing the draws and stats returned by each step of the step method,
+        the :meth:`~ZarrChain.record_sampling_state` is called, the internal buffer is cleared and
+        the number of steps until the next flush is determined.
+        """
+        chain = self.chain
+        draw_slice = slice(self.draw_idx, self.draw_idx + self.draws_until_flush)
+        for group_name, buffer in self._buffers.items():
+            group = getattr(self, f"_{group_name}")
+            for var_name, var_value in buffer.items():
+                group[var_name].set_orthogonal_selection(
+                    (chain, draw_slice),
+                    np.stack(var_value),
+                )
+        self.draw_idx += self.draws_until_flush
+        self.record_sampling_state()
+        self.clear_buffers()
+        self.draws_until_flush = min([self.draws_per_chunk, self.total_draws - self.draw_idx])
+
+
+FILL_VALUE_TYPE = float | int | bool | str | np.datetime64 | np.timedelta64 | None
+DEFAULT_FILL_VALUES: dict[Any, FILL_VALUE_TYPE] = {
+    np.floating: np.nan,
+    np.integer: 0,
+    np.bool_: False,
+    np.str_: "",
+    np.datetime64: np.datetime64(0, "Y"),
+    np.timedelta64: np.timedelta64(0, "Y"),
+}
+
+
+def get_initial_fill_value_and_codec(
+    dtype: Any,
+) -> tuple[FILL_VALUE_TYPE, np.typing.DTypeLike, numcodecs.abc.Codec | None]:
+    _dtype = np.dtype(dtype)
+    fill_value: FILL_VALUE_TYPE = None
+    codec = None
+    try:
+        fill_value = DEFAULT_FILL_VALUES[_dtype]
+    except KeyError:
+        for key in DEFAULT_FILL_VALUES:
+            if np.issubdtype(_dtype, key):
+                fill_value = DEFAULT_FILL_VALUES[key]
+                break
+        else:
+            codec = numcodecs.Pickle()
+    return fill_value, _dtype, codec
+
+
+class ZarrTrace:
+    """Object that stores and enables access to MCMC draws stored in a :class:`zarr.hierarchy.Group` objects.
+
+    This class creats a zarr hierarchy to represent the sampling information which is
+    intended to mimic :class:`arviz.InferenceData`. The hierarchy looks like this:
+
+    | root
+    | |--> constant_data
+    | |--> observed_data
+    | |--> posterior
+    | |--> unconstrained_posterior
+    | |--> sample_stats
+    | |--> warmup_posterior
+    | |--> warmup_unconstrained_posterior
+    | |--> warmup_sample_stats
+    | |--> _sampling_state
+
+    The root group is created when the ``ZarrTrace`` object is initialized. The rest of
+    the groups are created once :meth:`~ZarrChain.init_trace` is called with a few exceptions:
+    unconstrained_posterior is only created if ``include_transformed = True``, and the
+    groups prefixed with ``warmup_`` are created only after calling
+    :meth:`~ZarrTrace.split_warmup_groups`.
+
+    Since ``ZarrTrace`` objects are intended to be as close to
+    :class:`arviz.InferenceData` objects as possible, the groups store the dimension
+    and coordinate information following the `xarray zarr standard <https://xarray.pydata.org/en/v2023.11.0/internals/zarr-encoding-spec.html>`_.
+
+    Parameters
+    ----------
+    store : zarr.storage.BaseStore | collections.abc.MutableMapping | None
+        The store object where the zarr groups and arrays will be stored and read from.
+        Any zarr compatible storage object works. Keep in mind that if ``None`` is
+        provided, a :class:`zarr.storage.MemoryStore` will be used, which means that
+        information won't be visible to other processes and won't persist after the
+        ``ZarrTrace`` life-cycle ends. If you want to have persistent storage, please
+        use one of the multiple disk backed zarr storage options, e.g.
+        :class:`~zarr.storage.DirectoryStore` or :class:`~zarr.storage.ZipStore`.
+    synchronizer : zarr.sync.Synchronizer | None
+        The synchronizer to use for the underlying zarr arrays.
+    compressor : numcodec.abc.Codec | None | pymc.util.UNSET
+        The compressor to use for the underlying zarr arrays. If ``None``, no compressor
+        is used. If ``UNSET``, zarr's default compressor is used.
+    draws_per_chunk : int
+        The number of draws that make up a chunk in the variable's posterior array.
+        Each variable's array shape is set to ``(n_chains, n_draws, *rv_shape)``, but
+        the chunks are set to ``(1, draws_per_chunk, *rv_shape)``. This means that each
+        chain will have it's own chunk to read or write to, allowing for concurrent
+        write operations of different chains not to interfere with each other, and that
+        multiple draws can belong to the same chunk. The variable's core dimension
+        however, will never be split across different chunks.
+    include_transformed : bool
+        If ``True``, the transformed, unconstrained value variables are included in the
+        storage group.
+
+    Notes
+    -----
+    ``ZarrTrace`` objects represent the storage information. If the underlying store
+    persists on disk or over the network (e.g. with a :class:`zarr.storage.FSStore`)
+    multiple process will be able to concurrently access the same storage and read or
+    write to it.
+
+    The intended division of labour is for ``ZarrTrace`` to handle the creation and
+    management of the zarr group and storage objects and arrays, and for individual
+    :class:`~.ZarrChain` objects to handle recording MCMC samples to the trace. This
+    division was chosen to stay close to the existing `pymc.backends.base.MultiTrace`
+    and `pymc.backends.ndarray.NDArray` way of working with the existing samplers.
+
+    One extra feature of ``ZarrTrace`` is that it enables direct access to any array's
+    metadata. ``ZarrTrace`` takes advantage of this to tag arrays as ``deterministic``
+    or ``freeRV`` depending on what kind of variable they were in the defining model.
+
+    See Also
+    --------
+    :class:`~pymc.backends.zarr.ZarrChain`
+    """
+
+    def __init__(
+        self,
+        store: BaseStore | MutableMapping | None = None,
+        synchronizer: Synchronizer | None = None,
+        compressor: Codec | None | _UnsetType = UNSET,
+        draws_per_chunk: int = 1,
+        include_transformed: bool = False,
+    ):
+        if not _zarr_available:
+            raise RuntimeError("You must install zarr to be able to create ZarrTrace instances")
+        self.synchronizer = synchronizer
+        if compressor is UNSET:
+            compressor = default_compressor
+        self.compressor = compressor
+        self.root = zarr.group(
+            store=store,
+            overwrite=True,
+            synchronizer=synchronizer,
+        )
+
+        self.draws_per_chunk = int(draws_per_chunk)
+        assert self.draws_per_chunk >= 1
+
+        self.include_transformed = include_transformed
+
+        self._is_base_setup = False
+
+    def groups(self) -> list[str]:
+        return [str(group_name) for group_name, _ in self.root.groups()]
+
+    @property
+    def posterior(self) -> zarr.Group:
+        return self.root.posterior
+
+    @property
+    def unconstrained_posterior(self) -> zarr.Group:
+        return self.root.unconstrained_posterior
+
+    @property
+    def sample_stats(self) -> zarr.Group:
+        return self.root.sample_stats
+
+    @property
+    def constant_data(self) -> zarr.Group:
+        return self.root.constant_data
+
+    @property
+    def observed_data(self) -> zarr.Group:
+        return self.root.observed_data
+
+    @property
+    def _sampling_state(self) -> zarr.Group:
+        return self.root._sampling_state
+
+    def init_trace(
+        self,
+        chains: int,
+        draws: int,
+        tune: int,
+        step: BlockedStep | CompoundStep,
+        model: Model | None = None,
+        vars: Sequence[TensorVariable] | None = None,
+        test_point: dict[str, np.ndarray] | None = None,
+    ):
+        """Initialize the trace groups and arrays.
+
+        This function creates and fills with default values the groups below the
+        ``ZarrTrace.root`` group. It creates the ``constant_data``, ``observed_data``,
+        ``posterior``, ``unconstrained_posterior`` (if ``include_transformed = True``),
+        ``sample_stats``, and ``_sampling_state`` zarr groups, and all of the relevant
+        arrays that must be stored there.
+
+        Every array in the posterior and sample stats groups will have the
+        (chains, tune + draws) batch dimensions to the left of the core dimensions of
+        the model's random variable or the step method's stat shape. The warmup (tuning
+        draws) and the posterior samples are split at a later stage, once
+        :meth:`~ZarrTrace.split_warmup_groups` is called.
+
+        After the creation if the zarr hierarchies, it initializes the list of
+        :class:`~pymc.backends.zarr.Zarrchain` instances (one for each chain) under the
+        ``straces`` attribute. These objects serve as the interface to record draws and
+        samples generated by the step methods for each chain.
+
+        Parameters
+        ----------
+        chains : int
+            The number of chains to use to initialize the arrays.
+        draws : int
+            The number of posterior draws to use to initialize the arrays.
+        tune : int
+            The number of tuning steps to use to initialize the arrays.
+        step : pymc.step_methods.compound.BlockedStep | pymc.step_methods.compound.CompoundStep
+            The step method that will be used to generate the draws and stats.
+        model : pymc.model.core.Model | None
+            If None, the model is taken from the ``with`` context.
+        vars : Sequence[TensorVariable] | None
+            Sampling values will be stored for these variables. If ``None``,
+            ``model.unobserved_RVs`` is used.
+        test_point : dict[str, numpy.ndarray] | None
+            This is not used and is a product of the inheritance of :class:`ZarrChain`
+            from :class:`~.BaseTrace`, which uses it to determine the shape and dtype
+            of `vars`.
+        """
+        if self._is_base_setup:
+            raise RuntimeError("The ZarrTrace has already been initialized")  # pragma: no cover
+        model = modelcontext(model)
+        self.model = model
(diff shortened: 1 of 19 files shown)


pymc/backends/zarr.py
```diff
@@ -0,0 +1,301 @@
...
+    elif np.issubdtype(_dtype, "bool"):
+        return (False, _dtype, None)
+    elif np.issubdtype(_dtype, "str"):
+        return ("", _dtype, None)
+    elif np.issubdtype(_dtype, "datetime64"):
+        return (np.datetime64(0, "Y"), _dtype, None)
+    elif np.issubdtype(_dtype, "timedelta64"):
+        return (np.timedelta64(0, "Y"), _dtype, None)
+    else:
+        return (None, _dtype, numcodecs.Pickle())
```
maresb: Question from my own ignorance, since I don't understand so much how fill values are implemented. Are we just hoping that these fill values don't actually occur in the data?

If so, this seems especially perilous for `bool` :sweat_smile:

lucianopaz: No, they are supposed to be the initialisation values for the entries. When the sampler completes its run, all entries will be filled with the correct value. Zarr just needs you to tell it what value to give to unwritten places. In the storage, these entries are never actually written, they are produced when you ask for the concrete values in the array.
The dangerous part is that xarray is interpreting `fill_value` as an indicator of whether the actual value should be masked to nan. This seems to be because of the netcdf standard treats fill_value as something completely different.
To keep things as clean as possible, I’ll store the draw_idx of each chain in a separate group that should never be converted to xarray.

maresb: Ah, that makes a lot more sense now, thanks for the explanation!

In case it's non-obvious to more than me, maybe it would be helpful to try to make this more self-evident. Perhaps by calling the function `get_initial_fill_value_and_codec`, or make some little comment that the fill value is used for initialization?

michaelosthege:
> the current situation of the `MultiTrace` and `NDArray` backends is terrible. These backend classes have inconsistent signatures across subclasses, and it's very awkward to write new backends that adhere to them.

Yes, therefore I would recommend not to use them for any new implementation.

> McBackend was an attempt to make things sane again. As far as I understand, McBackend does support ways to dump samples to disk instead of holding them in memory using the ClickHouse database.

Just to clarify:
* `NumPyBackend` is the go-to for _in memory_ situations
* `ClickHouseBackend` is for storing on disk (in a database that may even sit on a different machine!)

It should be quite simple to implement a `ZarrBackend` with McBackend!
I would recommend to do that first, because McBackend's test suite already covers all (?) of the nasty edge cases.

> However, I found the backend a bit detached from arviz and xarray, and it seemed to be tightly linked to protocol buffers, which made it harder for me to see how I could customize stuff.

Yes and No. I would say ArviZ is a first-class citizen, because [`Run.to_inferencedata()`](https://github.com/pymc-devs/mcbackend/blob/be11eb84dce17480473d76369069687174641845/mcbackend/core.py#L194) is in the base class.
There are two things which McBackend does not integrate tightly:
1. `xarray` because it doesn't/didn't support sparse arrays (needed for sparse stats or variables with varying shape)
2. `InferenceData` groups other than `.posterior`

I consider 2. the primary weakness, and it's the only reason why I don't use McBackend by default.
I see that you added "the other" groups as properties to the `ZarrTrace`. Maybe this is something we should do on a more abstract level? Have `McBackend` define the signature of `InferenceData` without requiring a specific implementation for it?

First we must find answers to:
* How do prior/posterior/log_likelihood data points arrive?
* Do they arrive in some kind of "sampling" process that may get parallelized or doesn't fit into memory?
* If yes, should the storage backend even make a difference between MCMC and forward sampling?

__Protocol buffers__

I used them because they are convenient for specifying a data structure and not having to write serialization/deserialization code for it.

And they are only for the constant metadata `{constant_data, observed_data, coords, names dtypes, ...}` because this this needs to be serializable from a semi-clean data structure supporting all the weird data types that users may put into their `coords` (timestamps anybody?).

From the Python perspective this could also be done with `zarr` or `xarray` (they can serialize to binary), but can you serialize/deserialize that in another language?
The protobufs can be compiled to C++ or Rust to easily read/write run metadata from those languages too!

Is that tight integration?

The important design decision is not which implementation is used to serialize/deserialize metadata, but rather to freeze and detach these (meta)data from chains, draws and stats:
1. Determine it _before_ starting the costly MCMC
2. Serialize it to its own "blob" of data. Think of `{constant_data, observed_data, coords, names dtypes, ...}` as the "header" section of a trace.

lucianopaz:
Thanks @michaelosthege for the feedback!

> > McBackend was an attempt to make things sane again. As far as I understand, McBackend does support ways to dump samples to disk instead of holding them in memory using the ClickHouse database.
> 
> Just to clarify:
> 
>     * `NumPyBackend` is the go-to for _in memory_ situations
> 
>     * `ClickHouseBackend` is for storing on disk (in a database that may even sit on a different machine!)
> 
> 
> It should be quite simple to implement a `ZarrBackend` with McBackend! I would recommend to do that first, because McBackend's test suite already covers all (?) of the nasty edge cases.

I understand what the two backends for McBackend offer and that McBackend already has a test suite. Despite this, I'll try to argue in favor of writing something that's detached from McBackend.

> > However, I found the backend a bit detached from arviz and xarray, and it seemed to be tightly linked to protocol buffers, which made it harder for me to see how I could customize stuff.
> 
> Yes and No. I would say ArviZ is a first-class citizen, because [`Run.to_inferencedata()`](https://github.com/pymc-devs/mcbackend/blob/be11eb84dce17480473d76369069687174641845/mcbackend/core.py#L194) is in the base class.

The way I see this is that McBackend offers a signature to convert from a kind of storage (like `MultiTrace`) into another one (`arviz.InferenceData`). I understand that with this method, you guarantee that there should always be a method to go from an McBackend `Run` to `arviz.InferenceData`, but you have to handle a lot of transformation logic in this conversion (just like the extra conversion logic that's already in `pymc.backends.arviz`). In my opinion, this isn't tight integration. Having the data stored in native zarr makes it possible to generate `xarray.Dataset` objects with a simple `xr.open_zarr(store, group)` calls, and then these can be wrapped into an `InferenceData` object with a simple `InferenceData(posterior=zarr_posterior, ...)` (and potentially even into the future `DataTree` objects, since zarr hierarchies are already very much tree-like).

> There are two things which McBackend does not integrate tightly:
> 
>     1. `xarray` because it doesn't/didn't support sparse arrays (needed for sparse stats or variables with varying shape)

`xarray` does not support sparse arrays. At the moment, the posterior samples are initialized as "empty" zarr arrays (in practice, filled arrays with a `fill_value`). The nice thing about zarr arrays is that these filled, uninitialized places, don't take up almost any space because they aren't actually stored. If queried, their value gets set from the `fill_value` attribute. `xarray` still needs to figure out pydata/xarray#5475 though.
 
>     2. `InferenceData` groups other than `.posterior`
> 
> 
> I consider 2. the primary weakness, and it's the only reason why I don't use McBackend by default. I see that you added "the other" groups as properties to the `ZarrTrace`.

The key thing is that I added these groups to the zarr hierarchy, having them as `ZarrTrace` properties is not necessary. By having them in a single shared zarr entity, they are stored almost like an `InferenceData` from zarr. I need to actually check if `arviz` has a `from_zarr` method, because that would be the direct conversion method from a `ZarrTrace` to an `InferenceData` object without having to add any extra conversion code.

> Maybe this is something we should do on a more abstract level? Have `McBackend` define the signature of `InferenceData` without requiring a specific implementation for it?
> 
> First we must find answers to:
> 
>     * How do prior/posterior/log_likelihood data points arrive?
> 
>     * Do they arrive in some kind of "sampling" process that may get parallelized or doesn't fit into memory?
> 
>     * If yes, should the storage backend even make a difference between MCMC and forward sampling?

I decided to only focus on MCMC for now, and I'm trying to make `ZarrTrace` handle concurrent writes to the zarr store from multiple processes during sampling. Having said that, it's almost effortless to add other groups to a zarr hierarchy, and the same store could house `prior`, `prior_predictive`, `posterior_predictive` and `predictions` as well without having to handle almost any extra logic.

> **Protocol buffers**
> 
> I used them because they are convenient for specifying a data structure and not having to write serialization/deserialization code for it.
> 
> And they are only for the constant metadata `{constant_data, observed_data, coords, names dtypes, ...}` because this this needs to be serializable from a semi-clean data structure supporting all the weird data types that users may put into their `coords` (timestamps anybody?).
> 
> From the Python perspective this could also be done with `zarr` or `xarray` (they can serialize to binary), but can you serialize/deserialize that in another language? The protobufs can be compiled to C++ or Rust to easily read/write run metadata from those languages too!

Yes, you can deserialize almost all of the contents into C++ or Rust. [zarr](https://zarr.dev/) can be readable from python, Julia, C++, rust, javascript and Java. The only content that would not be readable in other languages would come from arrays with `object` dtype. At the moment, this is limited to two things:

1. [`SamplerWarning`](https://github.com/pymc-devs/pymc/blob/main/pymc/stats/convergence.py#L52)
2. [`StepMethodState`](https://github.com/pymc-devs/pymc/blob/main/pymc/step_methods/compound.py#L93)

The latter isn't a problem in my opinion because it is related exclusively to the python pymc step methods, and I detached it to its own private group in the zarr hierarchy. The former might be more problematic, but since `SamplerWarning` is a `dataclass`, it could potentially be converted into a dictionary and then represented as a json object, which [zarr can serialize without problems](https://zarr.readthedocs.io/en/stable/tutorial.html#object-arrays).

Having said that, there are other benefits that we would get if we were to rely on zarr directly, such as:

- Offloading the maintenance cost of the storage backend code
- Growing set of features that will become available to us as time goes by
- Seamless compression of the arrays to save storage space
- Integration with [multiple on disk storage options](https://zarr.readthedocs.io/en/stable/tutorial.html#storage-alternatives) that range from directory structure, zipfiles and multiple SQL and no-SQL databases
- Integration with [distributed or cloud storage](https://zarr.readthedocs.io/en/stable/tutorial.html#distributed-cloud-storage) like S3, Hadoop, Google Cloud Storage and Azure storage blob, and also [fsspec](https://zarr.readthedocs.io/en/stable/tutorial.html#io-with-fsspec).

I think that these added benefits plus the drop in maintenance costs in the long run warrant using zarr directly and not through a new backend for McBackend.

maresb:
@lucianopaz, have you done some benchmarks with this yet (in particular with S3)? I'm a bit concerned that with (1, 1, ...) chunk size that I/O will be a bottleneck.

pymc/backends/zarr.py
```diff
@@ -0,0 +1,432 @@
...
+        try:
+            global_attrs = {
+                "tuning_steps": self.tuning_steps,
+                "sampling_time": self.sampling_time,
+            }
+        except AttributeError:
+            global_attrs = {}  # pragma: no cover
+        for name, _ in self.root.groups():
+            if name not in SUPPORTED_GROUPS_ALL:
+                continue
```
OriolAbril: I would instead exclude whatever groups are considered special/private by pymc. The supported groups concept will be removed as it goes against the general idea with inferencedata (if things with a specific name are available they will be used and interpreted in the way defined in the schema, but storing arbitrary things that help you in there is also perfectly valid).

The warning was useful at first when actively developing converters and for developers to easily catch having named a group `poterior` instead of `posterior` but now that inferencedata uses are more advanced, the warning is very annoying.

OriolAbril (commented):
I love this direction. I left a comment on ArviZ integration.

I also have more ideas of things that can be done to integrate better the sampling outputs with inferencedata but it might be better to address them in follow up PRs. Not having to go into the current ArviZ converter might help get this things off the ground. Many of these are around since https://github.com/pymc-devs/pymc/issues/5160

Also, anything on ArviZ side that can help with this let me know

---

Better sample_stats. [sample_stats](https://python.arviz.org/en/stable/schema/schema.html#sample-stats) doesn't necessarily need to restrict itself to having `chain, draw` dimensions in all its variables. the mass matrix could also go in there and a `divergence_id` even (with extra coordinate values or a multiindex to store the start and end points of divergences) which would complement the boolean `diverging` variable with `chain, draw` dimension.

samples in the [unconstrained space](https://python.arviz.org/en/stable/schema/schema.html#unconstrained-groups). related to https://github.com/pymc-devs/pymc/issues/6721 and to a lesser extent https://github.com/arviz-devs/arviz-base/issues/8

pymc/backends/zarr.py
```diff
@@ -0,0 +1,433 @@
...
+        )
+        self._sampling_state = zarr.open_group(
+            store, synchronizer=synchronizer, path="_sampling_state", mode="a"
+        )
+        self.stats_bijection = stats_bijection
+
+    def setup(self, draws: int, chain: int, sampler_vars: Sequence[dict] | None):  # type: ignore[override]
+        self.chain = chain
+
+    def record(self, draw: Mapping[str, np.ndarray], stats: Sequence[Mapping[str, Any]]):
```
aseyboldt: I did not check the source, but I think zarr will write the whole chunk each time we set a draw here, even if that chunk is not full yet. If that is indeed the case, we should be able to speed this up a lot if draws_per_chunk is >1 if we buffer `draws_per_chunk` draws, and the set values in one go.

aseyboldt (commented):
I absolutely love this! :D

pymc/backends/zarr.py
lucianopaz: Thanks for letting me know. I'll go back to my original approach which was to only store groups that didn't start with an underscore.

pymc/backends/zarr.py
lucianopaz: Oh, it would be a shame that zarr itself doesn't buffer the write until the chunk is filled.

aseyboldt:
I just did a little experiment with strace, to see what zarr does when we write a value to a chunk.

I wrote a little script that creates the array, writes two chunks and then closes the file. With `strace python the-script.py` we can see all the syscalls it uses in between

```python
import zarr
import numpy as np
import sys

store = zarr.DirectoryStore("zarr-test.zarr")
#store = zarr.LRUStoreCache(store, max_size=2**28)
data = zarr.open(store)
foo = data.array("foo", np.zeros((0, 0, 0)), chunks=(10, 10, 1))
foo.resize((1000, 10, 1))

# Mark the position in the code to make it easier to find the correct part
print("start--", flush=True, file=sys.stderr)

foo[0, 0, 0] = 1.0
print("mid--", flush=True, file=sys.stderr)
foo[1, 0, 0] = 2.0

print("done--", flush=True, file=sys.stderr)
```


The first write triggers this:

```strace
write(2, "start--\n", 8start--
)                = 8
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0", 0x7ffd56fa6d50) = -1 ENOENT (No such file or directory)
getpid()                                = 414823
futex(0x7b79166b0a88, FUTEX_WAKE_PRIVATE, 2147483647) = 0
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0", 0x7ffd56fa6cd0) = -1 ENOENT (No such file or directory)
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo", {st_mode=S_IFDIR|0755, st_size=14, ...}) = 0
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo", {st_mode=S_IFDIR|0755, st_size=14, ...}) = 0
fstat(3, {st_mode=S_IFCHR|0666, st_rdev=makedev(0x1, 0x9), ...}) = 0
read(3, "\231;\177Y\rp\267\323W?\16\212n\372\346\372", 16) = 16
openat(AT_FDCWD, "/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0.993b7f590d7047d3973f0e8a6efae6fa.partial", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 4
fstat(4, {st_mode=S_IFREG|0644, st_size=0, ...}) = 0
ioctl(4, TCGETS, 0x7ffd56fa6c10)        = -1 ENOTTY (Inappropriate ioctl for device)
lseek(4, 0, SEEK_CUR)                   = 0
write(4, "\2\0011\10 \3\0\0 \3\0\0/\0\0\0\24\0\0\0\27\0\0\0\37\0\1\0\377\377F\37"..., 47) = 47
close(4)                                = 0
rename("/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0.993b7f590d7047d3973f0e8a6efae6fa.partial", "/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0") = 0
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0.993b7f590d7047d3973f0e8a6efae6fa.partial", 0x7ffd56fa6d30) = -1 ENOENT (No such file or directory)
```

The second write triggers this:

```
write(2, "mid--\n", 6mid--
)                  = 6
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0", {st_mode=S_IFREG|0644, st_size=47, ...}) = 0
openat(AT_FDCWD, "/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0", O_RDONLY|O_CLOEXEC) = 4
fstat(4, {st_mode=S_IFREG|0644, st_size=47, ...}) = 0
ioctl(4, TCGETS, 0x7ffd56fa6c30)        = -1 ENOTTY (Inappropriate ioctl for device)
lseek(4, 0, SEEK_CUR)                   = 0
lseek(4, 0, SEEK_CUR)                   = 0
fstat(4, {st_mode=S_IFREG|0644, st_size=47, ...}) = 0
read(4, "\2\0011\10 \3\0\0 \3\0\0/\0\0\0\24\0\0\0\27\0\0\0\37\0\1\0\377\377F\37"..., 48) = 47
read(4, "", 1)                          = 0
close(4)                                = 0
getpid()                                = 414823
getpid()                                = 414823
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0", {st_mode=S_IFREG|0644, st_size=47, ...}) = 0
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo", {st_mode=S_IFDIR|0755, st_size=24, ...}) = 0
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo", {st_mode=S_IFDIR|0755, st_size=24, ...}) = 0
fstat(3, {st_mode=S_IFCHR|0666, st_rdev=makedev(0x1, 0x9), ...}) = 0
read(3, "q\313\267\231t\364\276E\235\223J4\354}\372\240", 16) = 16
openat(AT_FDCWD, "/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0.71cbb79974f44e459d934a34ec7dfaa0.partial", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 4
fstat(4, {st_mode=S_IFREG|0644, st_size=0, ...}) = 0
ioctl(4, TCGETS, 0x7ffd56fa6c10)        = -1 ENOTTY (Inappropriate ioctl for device)
lseek(4, 0, SEEK_CUR)                   = 0
write(4, "\2\0011\10 \3\0\0 \3\0\0006\0\0\0\24\0\0\0\36\0\0\0\37\0\1\0\377\377F\37"..., 54) = 54
close(4)                                = 0
rename("/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0.71cbb79974f44e459d934a34ec7dfaa0.partial", "/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0") = 0
stat("/home/adr/git/flow-experiment/zarr-test.zarr/foo/0.0.0.71cbb79974f44e459d934a34ec7dfaa0.partial", 0x7ffd56fa6d30) = -1 ENOENT (No such file or directory)
write(2, "done--\n", 7done--
)                 = 7
```

So there's a lot going on for each write to the array. If I read this correctly, for the second store it actually reads the chunk from the disc, then modifies the chunk with the indexing update, writes the new chunk to a temporary file and then replaces that with the original chunk file.

For the first write it skips reading in the chunk data, because there still is nothing there to read.

So I think if we want to get good performance from this, we should try to combine writes, by buffering `draws_per_chunk` items, and then writing them in one go.

requirements.txt
```diff
@@ -8,3 +8,4 @@ rich>=13.7.1
 scipy>=1.4.1
 threadpoolctl>=3.1.0,<4.0.0
 typing-extensions>=3.7.4
+zarr>=2.5.0,<3
```
ricardoV94: Shouldn't be part of `requirements`, just dev ones?

lucianopaz: Yeah, we could make it optional without much hassle, I'd just have to do a bunch of conditional imports. The thing is that I'd like to have it included and used by default eventually.

lucianopaz:
Thanks everyone for your comments. I think that this PR is almost ready. The things that are still missing are:

- [ ] Documentation
- [ ] Store sampling information periodically
- [ ] Buffer write operations depending on `draws_per_chunk` (as @aseyboldt pointed out)
- [ ] Warm start sampling from a `ZarrTrace`

Maybe the last step is a bit too much for this PR actually and could be addressed later next month (I really have to transition into other projects now).

lucianopaz:
@aseyboldt, I wanted to ask if you could check out the latest changes I added. I did two big things:

1. Buffer write operations in `ZarrChain`
2. Store the draws in the subprocesses themselves when the strace is a `ZarrChain`. I would like to always record the sampling state at the end of a successful run, or also grab it upon abortion, but I might need help with the multiprocess communication that has to go on to get that to work.

lucianopaz:
If the tests pass, the only steps left for this PR to be ready to merge are:

- [ ] Documentation
- [ ] Start sampling from a previous run's sampling state

@ricardoV94, to ensure that the previous sampling state is consistent with the model and step methods that are fed to `pymc.sample`, I might need to grab [these hash utilities](https://github.com/pymc-devs/pymc-experimental/pull/277). Since I can't build the model nor the step methods from the sampling state, I need to add a check for consistency (e.g. if data changed, if the step methods are not applied to the same variables, if the variable's computation graph are different, an informative error should be thrown). I guess I would like to ask if you think that those hash utilities are in a good enough stage to go into one of the main repos?

lucianopaz:
> Why do you need to know whether the model changed for this specific backend?

It's not for the backend. The backend works with what I've written so far. It stores all of the inference data information, and it also stores the step method's sampling state. Maybe I can leave this PR with that functionality only. What I also wanted to include was the ability for `pm.sample` to grab a `ZarrTrace` and set the step method's sampling state to whatever value was stored. That would allow:

1. the `ZarrTrace` to act as a checkpoint file during sampling (if the machine that is running `pm.sample` crashes, it might be possible to resume sampling from the partially complete `ZarrTrace` file)
2. to run the tuning phase somewhere and then re-use those tuning results to start drawing samples later (I don't think this is good practice, but I've heard people ask for it).

I only need the hash utilities for starting sampling from a previously stored state.

ricardoV94:
> * o act as a checkpoint file during sampling (if the machine that is running `pm.sample` crashes, it might be possible to resume sampling from the partially complete `ZarrTrace` file)
> 
> * to run the tuning phase somewhere and then re-use those tuning results to start drawing samples later (I don't think this is good practice, but I've heard people ask for it).

Why do we need to check for model consistency? That should be the responsibility of the user. I'm sure functional PPLs like blackjax don't try to guess if the density function changed during evals?

lucianopaz:
> > * o act as a checkpoint file during sampling (if the machine that is running `pm.sample` crashes, it might be possible to resume sampling from the partially complete `ZarrTrace` file)
> > * to run the tuning phase somewhere and then re-use those tuning results to start drawing samples later (I don't think this is good practice, but I've heard people ask for it).
> 
> The use case sounds great, but why do we need to check for model consistency? That should be the responsibility of the user. I'm sure functional PPLs like blackjax don't try to guess if the density function changed during evals?

The step methods have a reference to the tensor variable they are supposed to step. I wanted to make sure that if we're trying to set the state of a NUTS stepper, it should at least have been defined for the same variable. The problem is that the stored value won't be the exact same object, but a copy of it. That's why I want to check if it's hash representation matches.

ricardoV94:
I don't think we should check, the step samplers should be a step function, given a logp/dlogp function and a previous state they propose the next value/state.

They shouldn't need to know anything about a pymc model or pytensor variables? In theory they could work with densities provided by the user that doesn't even involve pymc, like nutpie does.

The whole shared variable setting across samplers we have going on now is pretty lame

ricardoV94:
Matching should be done by variable names like posterior predictive does.

ricardoV94:
Can we keep this PR about adding Zarr trace only?

ricardoV94:
Re flaky test, @bwengals was surprised by the magnitude of the error and there is an open issue for it.

If you're confident about there being no issue please close the associated issue. Perhaps as a separate PR to unblock other contributions depending on how long this takes to get merged

bwengals:
Put a fix in for the flaky GP test here #7567

lucianopaz:
> Can we keep this PR about adding Zarr trace only?
> 
> And not sampling state/ resuming

Yes, the only thing that I still need to figure out is what needs to be changed in the current `ZarrTrace` and `ZarrChain` classes to make sure that the sampling state/resuming will be workable without a big refactor of these two classes.

lucianopaz:
> Put a fix in for the flaky GP test here #7567

Your fix looks great. I was just trying out stuff to see if the CI would build. I'll drop the commit with the test patch.

lucianopaz:
@ricardoV94, @OriolAbril, @aseyboldt, @maresb and @michaelosthege. I think that this is ready for a proper review. I'll try to handle the mambaforge sunset error in a different PR

lucianopaz:
As I said before, I think that this PR is no longer a draft, but I think that there are some choices that I made might be better to discuss here:

1. Add zarr as a mandatory requirement. In all honesty, I don't think that this is a bad thing, and I chose to leave it like this for simplicity and because I would love to eventually deprecate `MultiTrace` et al. But as @ricardoV94 pointed out, I could make zarr an optional requirement and then add a bunch of conditional imports.
2. Record the warmup draws and posterior draws in the same array only to split them at the end of sampling. I chose to do it like this because the other backends operated this way, but I think that it might be a waste. I think that it's possible and maybe even preferable to create the warmup and normal groups and have the `ZarrChain.record` determine where to put the draw. It would involve a bit more logic, but I think it's doable.
3. How to handle `report` objects. I decided to ignore the `report` object that gets attached to the `MultiTrace`, but I could actually store that information inside a private group of the `ZarrTrace`. Something like `ZarrTrace.root._sampling_state`. Do you guys think that it's worth the effort?
4. Add `sampling_time` and `tuning_steps` as arrays in `_sampling_state` instead of adding them as `attrs` of the groups. The latter option is how `arviz` stores this info, but I decided to make them full fledged arrays in the `_sampling_state` group. I don't think that it's worth the effort to duplicate the same timing and step information across multiple groups, and having it only in the `_sampling_state` makes it easier to keep track of it.

pymc/sampling/parallel.py
```diff
@@ -96,13 +102,12 @@ def __init__(
         shared_point,
         draws: int,
         tune: int,
-        rng: np.random.Generator,
-        seed_seq: np.random.SeedSequence,
+        rng_state: RandomGeneratorState,
         blas_cores,
     ):
         # For some strange reason, spawn multiprocessing doesn't copy the rng
         # seed sequence, so we have to rebuild it from scratch
```
ricardoV94: Comment can be removed/updated? It's the bug you found for numpy<2.0?

pymc/backends/zarr.py
```diff
@@ -0,0 +1,836 @@
...
+            vars = model.unobserved_value_vars
+
+        unnamed_vars = {var for var in vars if var.name is None}
+        assert not unnamed_vars, f"Can't trace unnamed variables: {unnamed_vars}"
+        self.varnames = get_default_varnames(
+            [var.name for var in vars], include_transformed=self.include_transformed
+        )
+        self.vars = [var for var in vars if var.name in self.varnames]
+
+        self.fn = model.compile_fn(self.vars, inputs=model.value_vars, on_unused_input="ignore")
```
ricardoV94: Allow passing the compiled function already and use `pytensor.In`/`pytensor.Out` and trust_input for the default as was done in https://github.com/pymc-devs/pymc/pull/7578/commits/b589ce8562fe9fb23be420f4208333bd2db8357e

May be worth to refactor into a helper

pymc/sampling/mcmc.py
```diff
@@ -914,18 +926,69 @@ def _sample_return(
...
+        # Compute number of actual draws per chain
+        total_draws_per_chain = traces._sampling_state.draw_idx[:]
+        n_chains = len(traces.straces)
+        desired_tune = traces.tuning_steps
+        desired_draw = len(traces.posterior.draw)
+        tuning_steps_per_chain = np.clip(total_draws_per_chain, 0, desired_tune)
+        draws_per_chain = total_draws_per_chain - tuning_steps_per_chain
+
+        total_n_tune = tuning_steps_per_chain.sum()
+        total_draws = draws_per_chain.sum()
```
ricardoV94, line 996: How is this dealing with uneven draws, when sampling is Interrupted? MultiTrace was discarding draws from longer chains so everything is square

pymc/sampling/mcmc.py
```diff
@@ -1232,6 +1234,8 @@ def _iter_sample(
     step.set_rng(rng)
 
     point = start
+    if isinstance(trace, ZarrChain):
+        trace.link_stepper(step)
```
ricardoV94, line 1281: This could also be added to `NDarray`?

pymc/sampling/mcmc.py
```diff
@@ -1256,12 +1260,18 @@ def _iter_sample(
 
             yield diverging
     except KeyboardInterrupt:
+        if isinstance(trace, ZarrChain):
+            trace.record_sampling_state(step=step)
```
ricardoV94: Should we make this part of `trace.close`?

ricardoV94:
@lucianopaz PR looks good. Left some small comments

pymc/sampling/parallel.py
lucianopaz: Yes, you’re right. I’ll update the comment

pymc/sampling/mcmc.py
lucianopaz: When the `ZarrTrace` groups are created, they all get a fill value and the final shape for each array. The fill value is not actually stored in the `zarr.store`, so only the actual draws take up any memory. That being said, when you open and load an unfinished run, the arrays in the trace will have the shape as if they had been completely sampled, but the unsampled draws will be set to the fill value.

pymc/sampling/mcmc.py
lucianopaz: I don't want to change the signature of `close` and I don't want to make `trace` necessarily have a reference to the `step`. We could add a `record_sampling_state` method to `NDArray` though

requirements.txt
lucianopaz: @ricardoV94, what do you think about this in the end? Do we make it optional?

pymc/sampling/mcmc.py
ricardoV94: What is the default value? Would it make sense to trim it like MultiTrace does?

requirements.txt
ricardoV94: I would suggest making it optional until it becomes well-tested / good candidate for default?

pymc/sampling/mcmc.py
ricardoV94: Sounds good, and doesn't have to be done in this PR. Just please open an issue?

lucianopaz: Opened #7621 to track this.

pymc/sampling/mcmc.py
lucianopaz: Also tracked in #7621

pymc/sampling/mcmc.py
lucianopaz: The default values are [here](https://github.com/lucianopaz/pymc/blob/zarr/pymc/backends/zarr.py#L225-L232)

pymc/backends/zarr.py
lucianopaz: Should be addressed by 596a27c. I didn't refactor the precompilation into a helper, but I did add a flag to `model.compile_fn`.

lucianopaz:
@ricardoV94, I addressed the last couple of comments. Can you check if you think this is now ready to merge? Once this is merged, we could check whether #7612 is still an issue or not.

pymc/model/core.py
```diff
@@ -1611,6 +1611,7 @@ def compile_fn(
         inputs: Sequence[Variable] | None = None,
         mode=None,
         point_fn: bool = True,
+        borrow_vars: bool = False,
```
ricardoV94: Why an option and with this default? 

It shouldn't need to be an option. The point is that if input=output there's no need for a deepcopy. Step samplers always copy the point before mutating it, it's not supposed to be mutated inplace

Also, no reason to treat any differently than NDarray backend

pymc/model/core.py
```diff
@@ -1645,6 +1650,8 @@ def compile_fn(
                 mode=mode,
                 **kwargs,
             )
+        if borrow_vars:
+            fn.trust_input = True
```
ricardoV94: This is not related to borrow_vars. Also, no reason to treat any differently than NDarray backend

pymc/sampling/mcmc.py
ricardoV94: I don't like this. I think what MultiTrace is doing is more sensible. You won't get junk draws if you interrupt earlier. Here you'll get zeros and false for integers and booleans, for chains that are incomplete.

requirements.txt
ricardoV94: Did you make it optional?

ricardoV94 (changes requested):

ricardoV94:
@lucianopaz can you add a regression test for #7612 ?

pymc/model/core.py
ricardoV94: Ah this is happening inside `Model`. I didn't see the file header...

pymc/model/core.py
ricardoV94: Still not related to borrow_vars

pymc/model/core.py
ricardoV94: We don't need these changes, as the `BaseTrace` class is the one compiling the function and it doesn't go through here. So no code duplication. Was there one in a previous version when I left my comment?

I would revert these changes in model core.py. They are not being used. The BaseTrace sidesteps Model altogether to create the point function

pymc/sampling/mcmc.py
```diff
@@ -1232,12 +1301,18 @@ def _iter_sample(
...
         trace.close()
         raise
     except BaseException:
+        if isinstance(trace, ZarrChain):
+            trace.record_sampling_state(step=step)
         trace.close()
```
ricardoV94: Why the repeated except instead of just `except (KeyboardInterrupt, BaseExcetion)`?

ricardoV94: I know it was already present before

pymc/sampling/parallel.py
```diff
@@ -96,16 +104,29 @@ def __init__(
...
         self._step_method = step_method
         self._step_method_is_pickled = step_method_is_pickled
+        self.chain = chain
+        self._zarr_recording = False
+        self._zarr_chain: ZarrChain | None = None
+        if zarr_chains_is_pickled:
+            self._zarr_chain = cloudpickle.loads(zarr_chains)[self.chain]
+        elif zarr_chains is not None:
+            self._zarr_chain = cast(list[ZarrChain], zarr_chains)[self.chain]
+        self._zarr_recording = self._zarr_chain is not None
```
ricardoV94, line 128: This and other changes in the sampling module suggest we don't have a good abstraction for backends, if we actually need to change the code to handle a new one. Can we add methods to the base class that handle the new required setup code and are no-ops for the existing ones when not needed?

pymc/sampling/mcmc.py
lucianopaz: I don't know, that was the way it was before and I didn't touch it.

pymc/model/core.py
lucianopaz: Is this wrong then? I left it as `False` by default to keep backwards compatibility.

pymc/model/core.py
lucianopaz: Should I drop this then?

pymc/sampling/mcmc.py
ricardoV94: Feel free to clean it up. Otherwise nobody will :)

pymc/model/core.py
ricardoV94: Yes, I don't think you need any of the changes in `model/core.py` right? The baseclass of the trace is handling all the function creation logic

pymc/sampling/mcmc.py
lucianopaz: Yes, you'll get NaNs but you'll also have the information of what was the last draw that was taken in the trace, so it will be easy to start from where you left off. Xarray isn't great for ragged arrays. What NDArray does is to cut down the chains until you shortest one you had, but we can't do that without losing the sampling state information that is needed for a re-start.

pymc/backends/zarr.py
```diff
@@ -452,7 +453,12 @@ def init_trace(
         )
         self.vars = [var for var in vars if var.name in self.varnames]
 
-        self.fn = model.compile_fn(self.vars, inputs=model.value_vars, on_unused_input="ignore")
+        self.fn = model.compile_fn(
+            self.vars,
+            inputs=model.value_vars,
+            on_unused_input="ignore",
+            borrow_vars=True,
+        )
```
ricardoV94: Did I mess it up. Doesn't `BaseTrace.__init__` already create the fn you need?

pymc/sampling/parallel.py
lucianopaz: I agree, we don't have a good abstraction. The `ZarrTrace` is essentially different than the `BaseTrace`. The `BaseTrace` has a single chain in mind, the `ZarrTrace` has all of them. The workers then just get views or chunks onto which they are allowed to write draws. In the other backends, the `BaseTrace` works on a single chain, handles all of the creation logic for that, and `MultiTrace` is just a thin wrapper around many chains.
In this PR, I tried to stay as close to the `BaseTrace` as possible and kept the core signatures in place.

requirements.txt
lucianopaz: Now it's optional

pymc/backends/zarr.py
ricardoV94: All my comments about borrow and trust_input were based on this assumption. Note the base class does not call `model.compile_fn`, it sidesteps it completely, hence why I was surprised you were changing it previously

lucianopaz: No, the `BaseTrace` for each chain (`ZarrChain`) is initialized in `ZarrTrace`. Now I'm sending off the `fn` that is compiled in `ZarrTrace` over to every chain instance to use it without having to recompile it.

pymc/sampling/mcmc.py
ricardoV94: Hmm, that's a good point. Then feel free to leave as is or issue a user warning about pre-filled values when sampling is aborted if you think that could avoid troubles to users.

pymc/sampling/parallel.py
ricardoV94: Do you think we could (fine if in a future PR) refactor MultiTrace to work more like ZarrTrace in regards to multiple chains so that the logic here can be shared?

pymc/backends/zarr.py
ricardoV94: We should do that for MultiTrace as well. I had that in mind as a follow up in #7578 where a lot of the wall time was just avoiding needless function compilations. 

My early comment still stands you shouldn't call `model.compile_fn` and you should (unless you don't want the speedup for some reason) have trust_input and borrow inputs and outputs to avoid deep copies needlessly. AFAICT the same logic should be applicable to the new and old traces.

ricardoV94: However, the reason I didn't do it for multiprocessing yet is that we have to be careful about point functions with RNGs (say with stochastic Deterministics). We actually need to copy the function and provide new shared RNGs for each function. If you're sharing `self.fn` across multiple chains (and they are called in different processes) the same concern may apply here?

ricardoV94: Here is a related issue: https://github.com/pymc-devs/pymc/issues/7588

ricardoV94: If you don't copy the function and do the RNG swap it's better not to share for now, as it may mess up the RNG updates and you get crappy random draws.

ricardoV94: Here is the kind of logic needed for this function copy and RNG swap: https://github.com/pymc-devs/pymc-extras/blob/c1809e8149fc89ac6eadf4bf73050ea6fe82955c/pymc_experimental/sampling/optimizations/conjugate_sampler.py#L84-L98

pymc/sampling/parallel.py
lucianopaz: I hope so, yes. But I think that it should be done in a different PR

pymc/backends/zarr.py
lucianopaz: Is the problem that the random generators inside `self.fn` are shared across chains? What I mean is that, is the problem that when different chains call the same `fn`, the random generator state will advance mixing up the different chains? I can see that this could happen in the context of a single process that runs multiple chains, either in threads or sequentially. But I don't see this happening in different processes.
Or is the problem that we don't manage to control the random generator state that gets called in `fn` as we do with step methods? If this is the real problem, we could address it using the `RandomGeneratorState`, and `random_generator_from_state` and `get_state_from_generator` to create copies of random generators that adhere to whatever state was inputed.

lucianopaz: Anyway, I think that #7588 really captures the core problem with what might be going on, so I don't think that it's worth opening a different issue. This PR doesn't close the issue though, but I don't think that it's intention should be to do so. I have an idea to try and tackle 7588 using the `RandomGeneratorState` stuff that I developed here.

pymc/model/core.py
ricardoV94: You seem to have reintroduced the changes in `model/core.py`, I thought you had agreed to reverse them?

lucianopaz:
@ricardoV94, I think that I fixed the merge conflicts with upstream main and I've removed the code I had added to `model.core`. If this is good to merge, let's get it in

ricardoV94 (approved):

ricardoV94:
This has big potential! Thanks for it

fonnesbeck:
We def need a notebook tutorial for this!

PR #7542, opened 2024-10-18
Derive logprob of matmul

ricardoV94:
This can be pretty useful for timeseries models as well as normalizing flows. 

Here is a simple example:

```python
import numpy as np
import pytensor.tensor as pt

import pymc as pm

rng = np.random.default_rng(37)
x = pm.MvNormal.dist(cov=np.eye(2), size=(128,))

n_layers = 3
A = pt.tensor("A", shape=(n_layers, 2, 2))
b = pt.tensor("b", shape=(n_layers, 2,))

# Repeated layers of Affine transform -> Tanh
for i in range(n_layers):
    y = A[i] @ x + b[i]
    # parametrized leaky-relu would be nicer: https://github.com/pymc-devs/pymc/issues/7543
    # y = pt.switch(y > 0, y, c[i] * y)
    y = pt.tanh(y)

A_test = rng.normal(size=A.type.shape)
b_test = rng.normal(size=b.type.shape)
y_test = rng.uniform(-1, 1, size=y.type.shape)
pm.logp(y, y_test).sum().eval({A: A_test, b: b_test})  # array(-3.54498234)
```

### DIFF
--- a/pymc/logprob/linalg.py
+++ b/pymc/logprob/linalg.py
@@ -1,0 +1,102 @@
+#   Copyright 2024 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+import pytensor.tensor as pt
+
+from pytensor.graph.rewriting.basic import node_rewriter
+from pytensor.tensor.math import _matrix_matrix_matmul
+
+from pymc.logprob.abstract import MeasurableBlockwise, MeasurableOp, _logprob, _logprob_helper
+from pymc.logprob.rewriting import measurable_ir_rewrites_db
+from pymc.logprob.utils import check_potential_measurability, filter_measurable_variables
+
+
+class MeasurableMatMul(MeasurableBlockwise):
+    """Measurable matrix multiplication operation."""
+
+    right_measurable: bool
+
+    def __init__(self, measurable_right: bool, **kwargs):
+        self.right_measurable = measurable_right
+        super().__init__(**kwargs)
+
+
+@_logprob.register(MeasurableMatMul)
+def logprob_measurable_matmul(op, values, l, r):  # noqa: E741
+    [y_value] = values
+    if op.right_measurable:
+        A, x = l, r
+        x_value = pt.linalg.solve(A, y_value)
+    else:
+        x, A = l, r
+        x_value = pt.linalg.solve(A.mT, y_value.mT).mT
+
+    x_logp = _logprob_helper(x, x_value)
+
+    # The operation has a support dimensionality of 2
+    # We need to reduce it if it's still present in the base logp
+    if x_logp.type.ndim == x_value.type.ndim:
+        x_logp = pt.sum(x_logp, axis=(-1, -2))
+    elif x_logp.type.ndim == x_value.type.ndim - 1:
+        x_logp = pt.sum(x_logp, axis=-1)
+
+    _, log_abs_jac_det = pt.linalg.slogdet(A)
+
+    return x_logp - log_abs_jac_det
+
+
+@node_rewriter(tracks=[_matrix_matrix_matmul])
+def find_measurable_matmul(fgraph, node):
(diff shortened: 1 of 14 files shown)


pymc/logprob/linalg.py
```diff
@@ -0,0 +1,102 @@
...
+    x_logp = _logprob_helper(x, x_value)
+
+    # The operation has a support dimensionality of 2
+    # We need to reduce it if it's still present in the base logp
+    if x_logp.type.ndim == x_value.type.ndim:
+        x_logp = pt.sum(x_logp, axis=(-1, -2))
+    elif x_logp.type.ndim == x_value.type.ndim - 1:
+        x_logp = pt.sum(x_logp, axis=-1)
+
+    _, log_abs_jac_det = pt.linalg.slogdet(A)
```
ricardoV94, line 53: The kind of PyTensor rewrites that @tanish1729 has been working on in PyTensor are critical here.

We should look at whether we are doing stuff with slogdet and not only det.
And would be nice to try to optimize for special matrices, which is basically what all the normalizing flows tricks boil down to

pymc/logprob/utils.py
```diff
@@ -320,7 +320,7 @@ def find_negated_var(var):
     return None
 
 
-def get_related_valued_nodes(node: Apply, fgraph: FunctionGraph) -> list[Apply]:
+def get_related_valued_nodes(fgraph: FunctionGraph, node: Apply) -> list[Apply]:
```
ricardoV94, line 323: This is much more natural order used in all sorts of pytensor utilities that require a node/variable and its' fgraph

pymc/logprob/tensor.py
```diff
@@ -215,29 +225,66 @@ def logprob_dimshuffle(op: MeasurableDimShuffle, values, base_var, **kwargs):
...
-    base_var = node.inputs[0]
+    # In cases where DimShuffle transposes dimensions, we only apply this rewrite when only Elemwise
+    # operations separate it from the valued node. Further transformations likely need to know where
+    # the support axes are for a correct implementation (and thus assume they are the rightmost axes).
+    # TODO: When we include the support axis as meta information in each  intermediate MeasurableVariable,
+    #  we can lift this restriction (see https://github.com/pymc-devs/pymc/issues/6360)
+    if tuple(node.op.shuffle) != tuple(sorted(node.op.shuffle)) and not _elemwise_univariate_chain(
+        fgraph, node
+    ):
+        return None
```
ricardoV94, line 285: These DimShuffle changes were needed to naturally accommodate `A @ x` when x is a vector, which looks like:
```python
import pytensor.tensor as pt

A = pt.matrix("A")
x = pt.vector("x")
y = A @ x
y.dprint()
# DropDims{axis=1} [id A]
#  └─ Blockwise{dot, (m,k),(k,n)->(m,n)} [id B]
#     ├─ A [id C]
#     └─ ExpandDims{axis=1} [id D]
#        └─ x [id E]
```

It's also more strict / correct than the limitation we had before, because the concerns are much more about what's after the DimShuffle not so much before.

pymc/logprob/linalg.py
jessegrabowski: We're didn't do anything with slogdet yet because that op is in need of some attention first. Notably, it doesn't have gradients implemented.

ricardoV94: Interesting, and `det` has? If so one more reason to only introduce it during specialization, as brought up in https://github.com/pymc-devs/pytensor/issues/1039

pymc/logprob/linalg.py
```diff
@@ -0,0 +1,102 @@
...
+
+@_logprob.register(MeasurableMatMul)
+def logprob_measurable_matmul(op, values, l, r):  # noqa: E741
+    [y_value] = values
+    if op.right_measurable:
+        A, x = l, r
+        x_value = pt.linalg.solve(A, y_value)
+    else:
+        x, A = l, r
+        x_value = y_value @ pt.linalg.inv(A)
```
jessegrabowski: You can do `pt.linalg.solve(A.T, y.T).T`  to solve `xA = y` 

~~But we can't guarantee that the incoming matrix will be invertible -- what happens if `A` isn't square?~~

I see you're checking for that below :)

ricardoV94: If A isn't square we fail at runtime when trying to invert. When we know already from the static shape it isn't we don't even try.

ricardoV94: Thanks for the solve, didn't know! One more for the list of linalg rewrites?

jessegrabowski: We should definitely rewrite graphs that do invert then matmul to solve

ricardoV94: It seems like we already have the rewrite :) just not with batch dims: https://github.com/pymc-devs/pytensor/issues/945

jessegrabowski (approved):
lgtm : )

PR #7553, opened 2024-10-29
Derive probability of set_subtensor operations

ricardoV94:
The logprob for the following kind of graphs can now be derived:

```python
import pytensor.tensor as pt
import pymc as pm

x = pt.zeros((3, 3))
y = x[[0, 1, 2], [0, 1, 2]].set(pm.Normal.dist(shape=(3,)))
z = y[[0, 1, 2], [2, 1, 0]].set(pm.StudentT.dist(nu=2, shape=(3,)))

pm.logp(z, pt.zeros((3, 3))).eval().round(2)
# array([[-0.92,  0.  , -1.04],
#        [ 0.  , -1.04,  0.  ],
#        [-1.04,  0.  , -0.92]])
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -121,6 +121,7 @@ jobs:
             tests/logprob/test_order.py
             tests/logprob/test_rewriting.py
             tests/logprob/test_scan.py
+            tests/logprob/test_set_subtensor.py
             tests/logprob/test_tensor.py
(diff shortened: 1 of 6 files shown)


ricardoV94:
Should restrict to when updates do not depend on previous writes to the buffer. Otherwise logp is not as straightforward/always deriverable

PR #7569, opened 2024-11-11
Allow interdependent initial points from same OpFromGraph node

ricardoV94:
This is needed to allow proper initival support of Marginalized variables in https://github.com/pymc-devs/pymc-experimental/pull/388, where multiple interdependent variables may become collapsed in a single node. We want initial values replacements to still propagate across these dependencies.

To achieve that the initial point machinery now makes use of `toposort_replace` (instead of assuming model variables are provided in a topological order) and the utility itself is expanded to untie variables coming from the same OpFromGraph node

----
📚 Documentation preview 📚: https://pymc--7569.org.readthedocs.build/en/7569/

### DIFF
--- a/tests/test_initial_point.py
+++ b/tests/test_initial_point.py
@@ -47,12 +47,17 @@ class TestInitvalEvaluation:
             )
         pass
 
-    def test_dependent_initvals(self):
+    @pytest.mark.parametrize("reverse_rvs", [False, True])
+    def test_dependent_initvals(self, reverse_rvs):
         with pm.Model() as pmodel:
             L = pm.Uniform("L", 0, 1, initval=0.5)
             U = pm.Uniform("U", lower=9, upper=10, initval=9.5)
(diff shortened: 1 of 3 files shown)


tests/test_initial_point.py
```diff
@@ -187,6 +193,40 @@ def test_string_overrides_work(self):
         assert np.isclose(iv["B_log__"], 0)
         assert iv["C_log__"] == 0
 
+    @pytest.mark.parametrize("reverse_rvs", [False, True])
+    def test_dependent_initval_from_OFG(self, reverse_rvs):
```
zaxtax, line 197: Is this test indicative of why this features is need for the other PR?

zaxtax (approved):
LGTM

tests/test_initial_point.py
ricardoV94: Yup

Issue #7539, opened 2024-10-16
Consider disabling PyTensor GC in sampling functions


ricardoV94:
### Description

This could avoid many reallocation between function calls in the step samplers and improve throughput.

We just have to remember to call it manually when sampler finishes or is aborted.

Before we decide to go down that route we should benchmark to see if it actually helps

PR #7578, opened 2024-11-19
Speedup `sample` and allow specifying `compile_kwargs` (several major changes related to step samplers)

ricardoV94:
## Major changes
1. internal uses of logp_dlogp_function now work with raveled inputs. External use will issue a warning unless `ravel_inputs` is specified explicitly. Eventually it will only be possible to use `ravel_inputs=True`.
1. Step samplers arguments besides vars must be passed by keyword
1. RaveledVars point_map_info is now a 4-n tuple, with size introduced.
1. `assign_step_method` does not call `instantiate_steppers`, but returns arguments needed for the latter.
1. Allow passing `compile_kwargs` to `pm.sample` which is then forwarded to the step samplers functions

## Enhancement
This PR speedups `NUTS` (and other step samplers), by:
  1. Avoiding many variable unravel and copies, by doing it inside PyTensor
  3. Avoiding copies when setting shared variables (borrow=True)
  6. Setting `trust_input=True` which can have a large overhead.
  7. Disabling GC collection for the C-backend function (related to #7539)
  8. Using slots for faster attribute access in the Tree class (and smaller footprint)
  9. Inlining some functions and being more lazy when possible

This PR speedups sample by:
 1. Avoiding way too many pytensor function compilations (model.initial_point() and very silly trace.fn after slicing at the end. It's also silly to compile the same function for every trace. We should just copy it.
 3. Avoid initializing NUTS step sampler just for most of the times then immediately discard it and using the one inside `init_nuts`. This will also reduce the path towards external samplers with nutpie/numpyro as it avoids the costly and useless compilation of the logp_dlogp_function
 4. Using `trust_input` and avoiding deepcopies in the trace function by using `pytensor.In(borrow=True)` and `pytensor.Out(borrow=True)`.

Further speedups should come for free from https://github.com/pymc-devs/pymc/issues/7539, specially for the Numba backend.

## Benchmark 

In the example below, sampling time is now only 7x slower than nutpie (5s vs 0.7s), compared to 13.5x slower (9.45s vs 0.7s) before. This assuming the same number of logp evals, in fact nutpie tuning allows us to get out with half the evals! We can hopefully bring it over.

Full time until from `pm.sample` to getting a trace is roughly halved as well (7.5s vs 14.4s), although this gain is not proportional to the number of draws.

With `compile_kwargs=(mode="NUMBA")`, sampling time is only  3x slower (2.3s).

```python
import time
import pymc as pm
import numpy as np
import nutpie
import pandas as pd

# Load the radon dataset
data = pd.read_csv(pm.get_data("radon.csv"))
data["log_radon"] = data["log_radon"].astype(np.float64)
county_idx, counties = pd.factorize(data.county)
coords = {"county": counties, "obs_id": np.arange(len(county_idx))}

# Create a simple hierarchical model for the radon dataset
with pm.Model(coords=coords, check_bounds=False) as model:
    intercept = pm.Normal("intercept", sigma=10)

    # County effects
    raw = pm.ZeroSumNormal("county_raw", dims="county")
    sd = pm.HalfNormal("county_sd")
    county_effect = pm.Deterministic("county_effect", raw * sd, dims="county")

    # Global floor effect
    floor_effect = pm.Normal("floor_effect", sigma=2)

    # County:floor interaction
    raw = pm.ZeroSumNormal("county_floor_raw", dims="county")
    sd = pm.HalfNormal("county_floor_sd")
    county_floor_effect = pm.Deterministic(
        "county_floor_effect", raw * sd, dims="county"
    )

    mu = (
        intercept
        + county_effect[county_idx]
        + floor_effect * data.floor.values
        + county_floor_effect[county_idx] * data.floor.values
    )

    sigma = pm.HalfNormal("sigma", sigma=1.5)
    pm.Normal(
        "log_radon", mu=mu, sigma=sigma, observed=data.log_radon.values, dims="obs_id"
    )

from pymc.model.transform.optimization import freeze_dims_and_data
model = freeze_dims_and_data(model)
```
```python
compiled_model = nutpie.compile_pymc_model(model)

start = time.perf_counter()
# More draws to make up for the fact that nutpie tunes better
trace_pymc = nutpie.sample(compiled_model, chains=1, tune=500, draws=1500, progress_bar=False)
end = time.perf_counter()
print(end - start)
```
```python
idata = pm.sample(
    model=model, 
    chains=1,
    tune=500, 
    draws=500, 
    progressbar=False, 
    compute_convergence_checks=False, 
    return_inferencedata=False,
    # compile_kwargs=dict(mode="NUMBA")
)
print(idata._report.t_sampling)
```

----
📚 Documentation preview 📚: https://pymc--7578.org.readthedocs.build/en/7578/

### DIFF
--- a/pymc/backends/base.py
+++ b/pymc/backends/base.py
@@ -30,9 +30,11 @@ from typing import (
 )
 
 import numpy as np
+import pytensor
 
 from pymc.backends.report import SamplerReport
 from pymc.model import modelcontext
+from pymc.pytensorf import compile_pymc
 from pymc.util import get_var_name
 
 logger = logging.getLogger(__name__)
@@ -168,7 +170,13 @@ class BaseTrace(IBaseTrace):
             raise Exception(f"Can't trace unnamed variables: {unnamed_vars}")
 
         if fn is None:
-            fn = model.compile_fn(vars, inputs=model.value_vars, on_unused_input="ignore")
+            # borrow=True avoids deepcopy when inputs=output which is the case for untransformed value variables
+            fn = compile_pymc(
+                inputs=[pytensor.In(v, borrow=True) for v in model.value_vars],
+                outputs=[pytensor.Out(v, borrow=True) for v in vars],
+                on_unused_input="ignore",
+            )
+            fn.trust_input = True
 
         # Get variable shapes. Most backends will need this
         # information.
--- a/pymc/backends/base.py
+++ b/pymc/backends/base.py
@@ -147,32 +147,45 @@ class BaseTrace(IBaseTrace):
         use different test point that might be with changed variables shapes
     """
 
-    def __init__(self, name, model=None, vars=None, test_point=None):
-        self.name = name
-
+    def __init__(
+        self,
+        name=None,
+        model=None,
+        vars=None,
+        test_point=None,
+        *,
+        fn=None,
+        var_shapes=None,
+        var_dtypes=None,
+    ):
         model = modelcontext(model)
-        self.model = model
+
         if vars is None:
             vars = model.unobserved_value_vars
 
         unnamed_vars = {var for var in vars if var.name is None}
         if unnamed_vars:
             raise Exception(f"Can't trace unnamed variables: {unnamed_vars}")
-        self.vars = vars
-        self.varnames = [var.name for var in vars]
-        self.fn = model.compile_fn(vars, inputs=model.value_vars, on_unused_input="ignore")
+
+        if fn is None:
+            fn = model.compile_fn(vars, inputs=model.value_vars, on_unused_input="ignore")
 
         # Get variable shapes. Most backends will need this
         # information.
-        if test_point is None:
-            test_point = model.initial_point()
-        else:
-            test_point_ = model.initial_point().copy()
-            test_point_.update(test_point)
-            test_point = test_point_
-        var_values = list(zip(self.varnames, self.fn(test_point)))
-        self.var_shapes = {var: value.shape for var, value in var_values}
-        self.var_dtypes = {var: value.dtype for var, value in var_values}
+        if var_shapes is None or var_dtypes is None:
+            if test_point is None:
+                test_point = model.initial_point()
+            var_values = tuple(zip(vars, fn(**test_point)))
+            var_shapes = {var.name: value.shape for var, value in var_values}
+            var_dtypes = {var.name: value.dtype for var, value in var_values}
+
+        self.name = name
+        self.model = model
+        self.fn = fn
+        self.vars = vars
+        self.varnames = [var.name for var in vars]
+        self.var_shapes = var_shapes
+        self.var_dtypes = var_dtypes
         self.chain = None
         self._is_base_setup = False
         self.sampler_vars = None
--- a/pymc/blocking.py
+++ b/pymc/blocking.py
@@ -39,11 +39,11 @@ StatDtype: TypeAlias = type | np.dtype
 StatShape: TypeAlias = Sequence[int | None] | None
 
 
-# `point_map_info` is a tuple of tuples containing `(name, shape, dtype)` for
+# `point_map_info` is a tuple of tuples containing `(name, shape, size, dtype)` for
 # each of the raveled variables.
 class RaveledVars(NamedTuple):
     data: np.ndarray
-    point_map_info: tuple[tuple[str, tuple[int, ...], np.dtype], ...]
+    point_map_info: tuple[tuple[str, tuple[int, ...], int, np.dtype], ...]
 
 
 class Compose(Generic[T]):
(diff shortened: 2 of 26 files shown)


pymc/model/core.py
```diff
@@ -231,13 +232,22 @@ def __init__(
             grads = pytensor.grad(cost, grad_vars, disconnected_inputs="ignore")
             for grad_wrt, var in zip(grads, grad_vars):
                 grad_wrt.name = f"{var.name}_grad"
-            outputs = [cost, *grads]
+            grads = pt.join(0, *[pt.atleast_1d(grad.ravel()) for grad in grads])
```
jessegrabowski, line 236: Why the zero?

ricardoV94: That's the axis

ricardoV94:
10 minutes seem to be saved in pytest CI time compared to previous runs

pymc/blocking.py
```diff
@@ -91,19 +90,14 @@ def rmap(
...
-
         last_idx = 0
-        for name, shape, dtype in array.point_map_info:
-            arr_len = np.prod(shape, dtype=int)
-            var = array.data[last_idx : last_idx + arr_len].reshape(shape).astype(dtype)
-            result[name] = var
-            last_idx += arr_len
+        for name, shape, size, dtype in array.point_map_info:
+            result[name] = array.data[last_idx : last_idx + size].reshape(shape).astype(dtype)
+            last_idx += size
```
ricardoV94: don't compute end twice

pymc/backends/base.py
```diff
@@ -147,32 +147,45 @@ class BaseTrace(IBaseTrace):
...
 
         unnamed_vars = {var for var in vars if var.name is None}
         if unnamed_vars:
             raise Exception(f"Can't trace unnamed variables: {unnamed_vars}")
-        self.vars = vars
-        self.varnames = [var.name for var in vars]
-        self.fn = model.compile_fn(vars, inputs=model.value_vars, on_unused_input="ignore")
+
+        if fn is None:
+            self.fn = model.compile_fn(vars, inputs=model.value_vars, on_unused_input="ignore")
```
jessegrabowski:
```py
fn = model.compile_fn(vars, inputs=model.value_vars, on_unused_input="ignore")
```

Otherwise this will be overwritten by `None` below, when `self.fn = fn`?

pymc/backends/base.py
```diff
@@ -168,7 +170,12 @@ def __init__(
             raise Exception(f"Can't trace unnamed variables: {unnamed_vars}")
 
         if fn is None:
-            self.fn = model.compile_fn(vars, inputs=model.value_vars, on_unused_input="ignore")
+            fn = compile_pymc(
+                inputs=[pytensor.In(v, borrow=True) for v in model.value_vars],
```
jessegrabowski, line 175: What is the advantage of using `pytensor.In` and `Out` here, compared to just passing the variables directly?

pymc/step_methods/arraystep.py
```diff
@@ -185,17 +185,17 @@ def __init__(
...
-            func = logp_dlogp_func
-
-        self._logp_dlogp_func = func
+            logp_dlogp_func = model.logp_dlogp_function(
+                vars,
+                dtype=dtype,
+                ravel_inputs=True,
+                **pytensor_kwargs,
+            )
+            logp_dlogp_func.trust_input = True
```
jessegrabowski, line 201: Is it worth also pulling out the raw jit function here if it exists with something like `if hasattr(logp_dlogp_fun.vm, 'jit_fn'): logp_dlogp_fun = logp_dlogp_fun.vm.jit_fn`  ? I use that in my code to get a bit of speedup. Or has this overhead been addressed on the pytensor side?

pymc/step_methods/hmc/nuts.py
```diff
@@ -296,7 +318,7 @@ def extend(self, direction):
         """
         if direction > 0:
             tree, diverging, turning = self._build_subtree(
-                self.right, self.depth, floatX(np.asarray(self.step_size))
+                self.right, self.depth, np.asarray(self.step_size, dtype=self.floatX)
```
jessegrabowski, line 321: iirc if you do `np.asarray` on an array, numpy will copy and re-create the array. Is it necessary to do this at every step? Or can `step_size` be set just once once?

jessegrabowski (approved):
lgtm, left a few small comments

pymc/step_methods/hmc/nuts.py
ricardoV94: np.asarray doesn't copy if it's already an array. np.array does

pymc/step_methods/arraystep.py
ricardoV94: There's a PR to reduce overhead on function and we still need it to handle the shared variables + updates if there happen to be any

pymc/backends/base.py
ricardoV94: It will not copy variables when input=output (due to the borrow=True)

pymc/step_methods/arraystep.py
ricardoV94: The overhead will be on the order of 2us after: https://github.com/pymc-devs/pytensor/pull/1101 which I think is reasonable enough to not mess with the internal API. It accounts for .06s of the total sampling time. It will go further down when we remove deprecated function stuff like `output_subset` and dict returns

pymc/backends/base.py
ricardoV94: Is this on an intermediate commit? I don't see it in the final changes

jessegrabowski: Yes, i reviewed commit by commit. I guess this one was changed, so the comment isn't relevant

ricardoV94: You're right though. I'll fix that commit

Issue #7069, opened 2023-12-15
Make `sample_posterior_predictive` API less surprising


ricardoV94:
## Description

The current behavior of `sample_posterior_predictive` regarding "volatility" and `var_names` is very opaque and error-prone. 

If you include a model variable in `var_names` it will always be sampled, even if it was an unobserved variable with values in the trace. Furthermore, any variables that depend on variables being sampled ("volatile") will also be sampled even if they are not in `var_names`, and become "volatile" themselves. With this, it's easy for users to end up resampling the whole model from the prior by accident when they only wanted to do predictions and copy some variables from the posterior to the predictions group.

**Most of the times I've seen users including unobserved variables in `var_names`, they seemed to only want to copy those over from the posterior to the posterior_predictive, to be together with the newly sampled observed variables.** The volatile logic was a bug from their perspective.

Furthermore if an unobserved variable depends on MutableData/Coords that have changed that will also be resampled, and the whole cascade of volatility propagation happens the same.

If this doesn't make sense, I tried to explain the current behavior in #7014 

### Why did we ever think this behavior was a good idea?

Both behaviors are sensible in that they provide internally consistent draws, and avoid reusing posteriors in case they are likely to make little theoretical sense (if you have an intercept for the cities ["Paris", "Lisbon"], and now changed the coords to be ["London", "Madrid"], there is no reason to think the old posteriors should apply to the new cities), or computational sense (the length of the intercepts may have changed, and reusing the old ones would likely cause shape errors or silent bugs). 

*Aside: Maybe in that example users set the new coords to be ["Paris, "Lisbon", "London"], or they changed the shape from 2 to 3, expecting the new dim would be sampled, but the original two reused. That would be nice but hard to do, and on the verge of mind reading if only the shape changed. Such behavior should not be the competence of `sample_posterior_predictive`, but helper model transformation function. `sample_posterior_predictive` should only have to worry whether the variables with the same name as those in the posterior trace can be used as is, or must be resampled.*

The current flexibility is not only a burden! It allows to do many interesting things like https://www.pymc-labs.io/blog-posts/out-of-model-predictions-with-pymc/

Anyway, the original motivation for these changes was because of Deterministics, which should sometimes be resampled (because they depended on MutableData that has changed) and other times reused to avoid useless re-computations in `sample_posterior_predictive`.
* https://github.com/pymc-devs/pymc/issues/5302
* https://github.com/pymc-devs/pymc/issues/5512

[...]

Issue #6876, opened 2023-08-24
`sample_posterior_predictive` does not consider changes in "ConstantData" and "Constant coords"


ricardoV94:
### Description

The existing checks assume the same model is used between sample and sample_posterior_predictive, in which case the only variable that can be changed are mutable ones.

However if one wants to sample in a new model (as described in https://www.pymc-labs.io/blog-posts/out-of-model-predictions-with-pymc/), they have to artificially use mutable variables even if they never intend to change them after definition:

```python
import pymc as pm

with pm.Model() as m:
    x = pm.ConstantData("x", [0, 0, 0, 0, 0, 0])
    b = pm.Normal("b", x)
    y = pm.Normal("y", b.sum(), observed=0)
    idata = pm.sample()

with pm.Model() as pred_m:
    x = pm.ConstantData("x", [0, 0, 0])
    # x = pm.MutableData("x", [0, 0, 0])  # This works fine
    b = pm.Normal("b", x)
    y = pm.Normal("y", b.sum(), observed=0)
    idata = pm.sample_posterior_predictive(idata, predictions=True,)  # Fails because b isn't resampled
```

A similar thing happens for variables whose `dims` are changed. If they were defined in `coords_mutable` it will resample correctly, but not if they were defined as vanilla "constant" coords

ricardoV94:
Related to #6977

ricardoV94:
Nothing changed

PR #7596, opened 2024-11-29
Make do interventions shared variables by default

### LINKED ISSUE #6977: Posterior predictive doesn't resample intermediate Deterministics of intervened variables
### Describe the issue:

When sampling with do operator, Deterministics are injected from the trace and do operation gives no effect

```python
with pm.Model() as model:
    a = pm.Normal("a")
    b = pm.Deterministic("b", a + 10)
    c = pm.Deterministic("c", b * 3)
    d = pm.Normal("d", c, observed=0)
with model:
    trace = pm.sample()
with pm.do(model, {"a": -100}):
    trace1 = pm.sample_posterior_predictive(trace, var_names=["c"])
# should raise an error since a is -100
np.testing.assert_allclose(trace1.posterior_predictive["c"], trace.posterior["c"])
with pm.do(model, {"a": -100}):
    trace1 = pm.sample_posterior_predictive(trace, var_names=["c", "b"])
# raises an error since a is -100
np.testing.assert_allclose(trace1.posterior_predictive["c"], trace.posterior["c"])
```

### Error message:

```shell
...
```

### PyMC version information:

5.9.1

### Context for the issue:

_No response_

ricardoV94:
This behavior should be more intuitive, since do interventions create `Data` containers, and users expect those to be mutable by default. This also alleviates some issues related to `sample_posterior_predictive` that currently does not look for sources of mutability on constants:

Closes #6977 (I tested the MWE locally)
Related to #7069 

We could consider something similar for `pm.observe` but that doesn't create data containers at all right now, so it's a tad less relevant and would require more changes.

----
📚 Documentation preview 📚: https://pymc--7596.org.readthedocs.build/en/7596/

### DIFF
--- a/pymc/model/transform/conditioning.py
+++ b/pymc/model/transform/conditioning.py
@@ -16,7 +16,9 @@ import warnings
 from collections.abc import Mapping, Sequence
 from typing import Any, Union
 
-from pytensor.graph import ancestors
+import pytensor
+
+from pytensor.graph import Constant, ancestors
 from pytensor.tensor import TensorVariable
 
 from pymc.logprob.transforms import Transform
@@ -126,7 +128,9 @@ def observe(
 def do(
     model: Model,
     vars_to_interventions: Mapping[Union["str", TensorVariable], Any],
-    prune_vars=False,
+    *,
+    make_interventions_shared: bool = True,
+    prune_vars: bool = False,
 ) -> Model:
     """Replace model variables by intervention variables.
 
@@ -140,6 +144,8 @@ def do(
         Dictionary that maps model variables (or names) to intervention expressions.
         Intervention expressions must have a shape and data type that is compatible
         with the original model variable.
+    make_interventions_shared: bool, defaults to True,
+        Whether to make constant interventions shared variables.
     prune_vars: bool, defaults to False
         Whether to prune model variables that are not connected to any observed variables,
         after the interventions.
@@ -170,11 +176,14 @@ def do(
 
     """
     do_mapping = {}
-    for var, obs in vars_to_interventions.items():
+    for var, intervention in vars_to_interventions.items():
         if isinstance(var, str):
             var = model[var]
         try:
-            do_mapping[var] = var.type.filter_variable(obs)
+            intervention = var.type.filter_variable(intervention)
+            if make_interventions_shared and isinstance(intervention, Constant):
+                intervention = pytensor.shared(intervention.data, name=var.name)
+            do_mapping[var] = intervention
         except TypeError as err:
             raise TypeError(
                 "Incompatible replacement type. Make sure the shape and datatype of the interventions match the original variables"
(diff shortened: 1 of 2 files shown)


pymc/model/transform/conditioning.py
```diff
@@ -167,11 +173,14 @@ def do(
...
     """
     do_mapping = {}
-    for var, obs in vars_to_interventions.items():
+    for var, intervention in vars_to_interventions.items():
         if isinstance(var, str):
             var = model[var]
         try:
-            do_mapping[var] = var.type.filter_variable(obs)
+            intervention = var.type.filter_variable(intervention)
+            if make_interventions_shared and isinstance(intervention, Constant):
```
lucianopaz, line 184: To me `make_interventions_shared` is easier than the past version, but I think that it should be more stringent. I think that if `make_interventions_shared` is `True`, it should wrap the `intervention` with a `shared` regardless of the type of data provided (be it a shared, a constant or a numpy array). Furthermore, if `make_interventions_shared` is `False` I think that the interventions should be set to `Contant`s.

ricardoV94: I'll have to review the code, but the thing is intervetions may not be constants, they could be pytensor graphs depending on other variables or input variables. In which case they can't be made Shared nor Constant. This check for `isinstance(intervention, Constant)` was just handling that.

Basically the same as calling `as_tensor` and checking if it comes back as a Constant or not. If not it's a more complex graph.

lucianopaz: Right! Ok, so then this is fine with me. But could you also handle the `make_interventions_shared=False` situation similarly? Something that checks if `intervention` is a `SharedVariable` instance and converts it to a `Constant`?

ricardoV94: Do we need that? Something being shared is never a problem. If the user wanted them constant he could do it themselves?

drbenvincent (commented):
Might not be strictly required for this issue, but would solve one of my recurring concerns... but could potentially add a test for something like

```python
with pm.Model() as model:
    a = pm.Normal("a")
    b = pm.Deterministic("b", a*2)
    c = pm.Normal("c", b/2)
```

then intervene on `a` with `pm.do`, resample with `pm.sample_posterior_predictive`, and check that the values of `c` have changed. Just an idea, might be not fully formed.

lucianopaz (approved):
LGTM

ricardoV94:
@drbenvincent I am a bit conflicted. I do want to test that as a user, but not as part of the codebase as it's covering something that's orthogonal to it: #6876 

This PR is not a solution to that problem, but sidesteps it in this specific case. Even if we didn't have that problem we would still want this PR so you can use `set_data` if the users wished to.

drbenvincent:
Sure - totally fine for you to ignore my suggestion. There are a bunch of PR's which are all closely related and I don't have enough knowledge to know which exact PR would be relevant for a user-facing test like this. Feel free to ignore. Can possibly add a test along these lines to one of the other PR's potentially.

drbenvincent (commented):
new test looks cool

PR #7603, opened 2024-12-03
Implement Ordered distribution factory

### LINKED ISSUE #7297: Implement `Ordered` distribution factory
### Description

For univariate IID, adding a transform=`ordered` is equivalent to sorting the raw draws (forward pass). The logp is proportional to the density of the original draws + ordered transform jacobian. We would just need to figure out the normalization constant so it defines a proper multivariate variable that integrates to 1.

With this users would have a generative graph for ordered variables, that they can also do prior_predictive sampling from. 
Also default initvals would work out of the box. Right now users always need to provide them.

This would also pretty much also obliviate the need for `default_transform` and `transform` that we added in #5674 and simplify the API.

The normalization constant is probably just `size!`: https://en.wikipedia.org/wiki/Order_statistic#The_joint_distribution_of_the_order_statistics_of_an_absolutely_continuous_distribution

The API could look something like:
```python
with pm.Model() as m:
  x = pm.Ordered("x", pm.Normal.dist(), shape=(3,))
```

ricardoV94:
Closes #7297 

----
📚 Documentation preview 📚: https://pymc--7603.org.readthedocs.build/en/7603/

### DIFF
--- a/pymc/distributions/ordered.py
+++ b/pymc/distributions/ordered.py
@@ -1,0 +1,133 @@
+#   Copyright 2024 The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+import pytensor.tensor as pt
+
+from pytensor.tensor.random.op import RandomVariable
+from pytensor.tensor.random.utils import normalize_size_param
+from pytensor.tensor.variable import TensorVariable
+
+from pymc.distributions.distribution import (
+    Distribution,
+    SymbolicRandomVariable,
+    _support_point,
+)
+from pymc.distributions.shape_utils import change_dist_size, get_support_shape_1d, rv_size_is_none
+from pymc.distributions.transforms import _default_transform, ordered
+
+
+class OrderedRV(SymbolicRandomVariable):
+    inline_logprob = True
+    extended_signature = "(x)->(x)"
+    _print_name = ("Ordered", "\\operatorname{Ordered}")
+
+    @classmethod
+    def rv_op(cls, dist, *, size=None):
+        # We don't allow passing `rng` because we don't fully control the rng of the components!
+
+        size = normalize_size_param(size)
+
+        if not rv_size_is_none(size):
+            core_shape = tuple(dist.shape)[-1]
+            shape = (*tuple(size), core_shape)
+            dist = change_dist_size(dist, shape)
+
+        sorted_rv = pt.sort(dist, axis=-1)
+
+        return OrderedRV(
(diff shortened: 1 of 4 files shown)


pymc/distributions/ordered.py
```diff
@@ -0,0 +1,133 @@
...
+        return ordered
+    else:
+        return None
+
+
+@_support_point.register(OrderedRV)
+def support_point_ordered(op, rv, dist):
+    # FIXME: This does not work with the default ordered transform
+    #  It leads to a -inf log_jac_det
+    # return support_point(dist)
```
ricardoV94: @aseyboldt @lucianopaz any idea if we could easily modify the OrderedTransform to work nicely with equal values in the constrained space? The current one maps a `[0, 0, 0]` vector to `[0, -inf, -inf]`. My guess is not. 

But then I don't have a nice way to define a valid support_point. I could try to add an increasing jitter but that will sooner or later fail for bounded distributions.

pymc/logprob/order.py
```diff
@@ -158,3 +183,54 @@ def max_logprob_discrete(op, values, base_rv, **kwargs):
...
+@_logprob.register(MeasurableSort)
+def sort_logprob(op, values, base_rv, axis, **kwargs):
+    r"""Compute the log-likelihood graph for the `Sort` operation."""
+    (value,) = values
+
+    logprob = _logprob_helper(base_rv, value).sum(axis=-1)
+
+    base_rv_shape = constant_fold(tuple(base_rv.shape), raise_not_constant=False)
+    n = pt.prod(base_rv_shape, axis=-1)
+    sorted_logp = pt.log(n) + logprob
```
lucianopaz: This doesn't seem to be the correct logp. I looked it up in the wikipedia page you linked [here](https://en.wikipedia.org/wiki/Order_statistic#Cumulative_distribution_function_of_order_statistics) and it's much more complicated than that.

To get it, we need the `base_rv` logcdf too:
imagen

lucianopaz: So logprob would be something like

```python
r = pt.arange(n)
sorted_logprob = (
    gamma_ln(n + 1) - gamma_ln(r) - gamma_ln(n - r)
    + logprob
    + (r - 1) * logcdf
    + (n - r) * logcdf
)
```

lucianopaz: Using that same link, you could also register a logcdf expression using this:
imagen

It would look something like
```python
r = pt.arange(n)
sorted_logcdf = logsumexp(
    gamma_ln(n + 1) - gamma_ln(r + 1) - gamma_ln(n - r)
    + r * logcdf
    + (n - r) * logcdf
)
```

lucianopaz: Oh no. I gave this some more thought and realized that the expressions I pasted above were for a single draw, not for the joint distribution. And the part that you posted about the joint says

imagen

So your expression is almost correct, except for the log(n) instead of log(n!)

```
sorted_logp = pt.gamma_ln(n + 1) + logprob
```

ricardoV94: Yeah the key is I'm constraining to the IID case and observing all the values. 

Yes, it was supposed to be a factorial, thanks for spotting it

ricardoV94: Yes, for (incomplete) order stats you need cdfs, you chan check the max logp (same file) uses it.

You would also need them for non-IID stuff, but it gets super crazy fast

ricardoV94: fixed

PR #7604, opened 2024-12-04
Exponential scale default to 1.0

ricardoV94:
----
📚 Documentation preview 📚: https://pymc--7604.org.readthedocs.build/en/7604/


ricardoV94:
> Any opinion on putting both `lam` and `scale` behind the `*`, to force users to specify which one they want? I'm sure there must be silent parameterization errors because of that

That would be a major change, historically only lam existed and could be passed by argument only. There was also a discussion with Normal, and people didn't want to force them to be kwarg only

jessegrabowski (approved):

PR #7606, opened 2024-12-05
Rename `compile_pymc` to `compile`

ricardoV94:
This function is becoming increasingly more important, so it deserves a less silly name

----
📚 Documentation preview 📚: https://pymc--7606.org.readthedocs.build/en/7606/

### DIFF
--- a/docs/source/api/pytensorf.rst
+++ b/docs/source/api/pytensorf.rst
@@ -6,7 +6,7 @@ PyTensor utils
 .. autosummary::
    :toctree: generated/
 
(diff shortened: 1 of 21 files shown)


aloctavodia (approved):
LGTM! This is also becoming central for Kulprit and relevant for some functionality in PreliZ. So glad to have a better name.

aloctavodia:
I just realized that `compile` is one of the built-in functions in Python https://docs.python.org/3/library/functions.html#compile

PR #7607, opened 2024-12-05
Bump numpy version due to use of `Generator.spawn` only available in `>=1.25`

### LINKED ISSUE #7605: BUG: Sampling fails because 'numpy.random._generator.Generator' object has no attribute 'spawn'
### Describe the issue:

PyMC now relies on the `spawn()` method from `numpy.random._generator.Generator`. But this is a relatively recent addition to numpy ([added in v1.25](https://numpy.org/doc/2.0/reference/random/generated/numpy.random.Generator.spawn.html)). So we need to set a lower limit on the numpy dependency.

```python
import pymc as pm

with pm.Model() as model:
    a = pm.Normal("a")
    obs = pm.Normal("obs", a, 1, observed=[1,2])
    idata = pm.sample()
```

### Error message:

<details>
```shell
Input In [7], in <cell line: 6>()
      7     a = pm.Normal("a")
      8     obs = pm.Normal("obs", a, 1, observed=[1,2])
----> 9     idata = pm.sample()
     11 print(az.summary(idata))
File ~/mambaforge/envs/pymc/lib/python3.10/site-packages/pymc/sampling/mcmc.py:733, in sample(draws, tune, chains, cores, random
_seed, progressbar, progressbar_theme, step, var_names, nuts_sampler, initvals, init, jitter_max_retries, n_init, trace, discard
_tuned_samples, compute_convergence_checks, keep_warning_stat, return_inferencedata, idata_kwargs, nuts_sampler_kwargs, callback
, mp_ctx, blas_cores, model, compile_kwargs, **kwargs)
    731 if random_seed == -1:
    732     random_seed = None
--> 733 rngs = get_random_generator(random_seed).spawn(chains)
    734 random_seed_list = [rng.integers(2**30) for rng in rngs]
    736 if not discard_tuned_samples and not return_inferencedata:
AttributeError: 'numpy.random._generator.Generator' object has no attribute 'spawn'
```
</details>

### PyMC version information:

<details>
# packages in environment at /home/xian/mambaforge/envs/pymc:
#
# Name                    Version                   Build  Channel
numpy                     1.24.2          py310h8deb116_0    conda-forge
pymc                      5.19.0               hd8ed1ab_0    conda-forge
pymc-base                 5.19.0             pyhd8ed1ab_0    conda-forge
pytensor                  2.26.4          py310ha549d7f_0    conda-forge
pytensor-base             2.26.4          py310h89e8f5a_0    conda-forge
</details>

ricardoV94:
Introduced in https://github.com/pymc-devs/pymc/pull/7508 (specifically https://github.com/pymc-devs/pymc/commit/4ea1406331923444838c7c8dfe9fc411934ed5a0)

Closes #7605 

Actually #7508 had a subtle breaking change, in that it used to be possible to seed each chain separately by passing a list of integers/rngs. Now there's no such control. Hopefully nobody needs that behavior these days...

----
📚 Documentation preview 📚: https://pymc--7607.org.readthedocs.build/en/7607/

### DIFF
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -727,6 +727,10 @@ def sample(
         )
 
     if random_seed == -1:
+        warnings.warn(
+            "Setting random_seed = -1 is deprecated. Pass `None` to not specify a seed.",
+            FutureWarning,
+        )
         random_seed = None
     elif isinstance(random_seed, tuple | list):
         warnings.warn(
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -528,10 +528,8 @@ def sample(
     random_seed : int, array-like of int, or Generator, optional
         Random seed(s) used by the sampling steps. Each step will create its own
         :py:class:`~numpy.random.Generator` object to make its random draws in a way that is
-        indepedent from all other steppers and all other chains. If a list, tuple or array of ints
-        is passed, each entry will be used to seed the creation of ``Generator`` objects.
-        A ``ValueError`` will be raised if the length does not match the number of chains.
-        A ``TypeError`` will be raised if a :py:class:`~numpy.random.RandomState` object is passed.
+        indepedent from all other steppers and all other chains.
+        A ``TypeError`` will be raised if a legacy :py:class:`~numpy.random.RandomState` object is passed.
         We no longer support ``RandomState`` objects because their seeding mechanism does not allow
         easy spawning of new independent random streams that are needed by the step methods.
     progressbar : bool, optional default=True
@@ -730,6 +728,12 @@ def sample(
 
     if random_seed == -1:
         random_seed = None
+    elif isinstance(random_seed, tuple | list):
+        warnings.warn(
+            "A list or tuple of random_seed no longer specifies the specific random_seed of each chain. "
+            "Use a single seed instead.",
+            UserWarning,
+        )
     rngs = get_random_generator(random_seed).spawn(chains)
     random_seed_list = [rng.integers(2**30) for rng in rngs]
 
--- a/conda-envs/environment-dev.yml
+++ b/conda-envs/environment-dev.yml
@@ -10,7 +10,7 @@ dependencies:
 - cachetools>=4.2.1
 - cloudpickle
 - h5py>=2.7
-- numpy>=1.15.0
+- numpy>=1.25.0
 - pandas>=0.24.0
 - pip
 - pytensor>=2.26.2,<2.27
--- a/conda-envs/environment-docs.yml
+++ b/conda-envs/environment-docs.yml
@@ -8,7 +8,7 @@ dependencies:
 - arviz>=0.13.0
 - cachetools>=4.2.1
 - cloudpickle
-- numpy>=1.15.0
+- numpy>=1.25.0
 - pandas>=0.24.0
 - pip
 - pytensor>=2.26.2,<2.27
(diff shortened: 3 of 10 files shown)


pymc/step_methods/hmc/base_hmc.py
```diff
@@ -166,7 +166,7 @@ def __init__(
         if scaling is None and potential is None:
             mean = floatX(np.zeros(size))
             var = floatX(np.ones(size))
-            potential = QuadPotentialDiagAdapt(size, mean, var, 10, rng=self.rng.spawn(1)[0])
+            potential = QuadPotentialDiagAdapt(size, mean, var, 10, rng=self.rng)
```
ricardoV94: Why was this spawning a new rng?

pymc/sampling/mcmc.py
```diff
@@ -729,8 +727,18 @@ def joined_blas_limiter():
...
+        )
         random_seed = None
-    rngs = get_random_generator(random_seed).spawn(chains)
+    elif isinstance(random_seed, tuple | list):
+        warnings.warn(
+            "A list or tuple of random_seed no longer specifies the specific random_seed of each chain. "
+            "Use a single seed instead.",
+            UserWarning,
+        )
+    rngs = np.random.SeedSequence(get_random_generator(random_seed)).spawn(chains)
```
lucianopaz: This needs to change a bit. `np.random.SeedSequence` will create a new `SeedSequence`, not a `Generator`. I suppose that you can access the generator's seed sequence as `get_random_generator(random_seed).bit_generator.seed_seq`, and then `spawn` from there.

From the numpy source, it looks rather easy to be able to [spawn new `Generator`s](https://github.com/numpy/numpy/blob/main/numpy/random/_generator.pyx#L301) from `BitGenerator`s. But the spawn method for those was also added on version 1.25, so we'll have to also look at how [`BitGenerator.spawn`](https://github.com/numpy/numpy/blob/main/numpy/random/bit_generator.pyx#L632) works off of seed sequences.

I think that the way to go would look something like:

```python
bit_gen = rng.bit_generator
seed_seq = bit_gen.seed_seq
rngs = [type(rng)(type(bit_gen)(seed)) for seed in seed_seq.spawn(chains)]
```

lucianopaz: This could be unpacked into a utility function and used on the other similar calls

ricardoV94: I see, so maybe lets just bumpy numpy min version?

ricardoV94: Okay numpy is dropping support for 1.24 on december 18, so maybe it's just fine: https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule

lucianopaz: I’m in favour of bumping

pymc/step_methods/hmc/base_hmc.py
lucianopaz: Because the potential and the step method have to own independent generators. If they were to share the generator, the sampling state will lose that information, and the step method you’d get from the sampling state would produce different draws than the original step method with a shared generator

ricardoV94: There is one potential per step, so makes sense to have a single source of randomness. The state of the potential should be a subset of the the state of the step.

lucianopaz (approved):
Lgtm. I’m not sure if there were any other files that needed the numpy version bump. I suppose that the tests should pick that up.

pymc/step_methods/hmc/base_hmc.py
lucianopaz: And I think that you get that with the spawning. Both streams will be independent, but providing the rng to the step method will determine the stream that is produced by the potential

PR #7613, opened 2024-12-13
Use static shape in join_nonshared_inputs

ricardoV94:
The built graph requires static shapes anyway, as they are used in the reshape after the slicing. Should generate slightly more readable graphs

----
📚 Documentation preview 📚: https://pymc--7613.org.readthedocs.build/en/7613/


ricardoV94:
Closed in favor of #7745

PR #7622, opened 2024-12-22
Enable trusted publishing (OIDC)

maresb:
## Description

Increase the security of our release process by enabling OIDC.

This mirrors https://github.com/pymc-devs/pytensor/pull/1135 and follows similar efforts in https://github.com/pymc-labs/pymc-marketing/pull/975 and https://github.com/pymc-labs/CausalPy/commit/f38030bdf658cec2b43c074e8b48779e46972dca which AFAIK have been working seamlessly.

While this is a draft release, the only blocker is my lack of permissions on PyPI and GitHub. TODO:

1. In GitHub under Settings > Environments create a new environment called `release` with required reviewers. Limit the branches and tags to `v*.*.*` to ensure correct formatting of the tags.
2. In PyPI, enable trusted publishing from this `release` environment

I'm not sure who has GitHub admin, but going from PyPI, I need help from one of @ColCarroll, @fonnesbeck, @michaelosthege, @twiecki. I'm happy to hop on a call with any of you so that we can bring this over the finish line (if you also have GitHub permissions). 

This requires a new release to properly test. (We could also mostly-test this if we set up a test-pypi step, but let's do that separately if people want.)

The only change is that when publishing, one of the "required reviewers" for releases as defined above needs to approve the "Upload release to PyPI" job before it executes. This helps to limit the possibilities for anyone trying to publish a fake release to PyPI.

Once there has been a release so that we confirm that this workflow is working properly, then we should delete the PyPI token. (Otherwise there is a lingering risk that someone somehow steals the token and gains privileges over the PyPI project.)

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7622.org.readthedocs.build/en/7622/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,8 @@ jobs:
       PYPI_TOKEN: ${{ secrets.PYPI_TOKEN_PYMC }}
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      with:
+        persist-credentials: false
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,57 +1,42 @@
 name: release-pipeline
 
 on:
+  push:
+    branches:
+      - main
   release:
     types:
-      - created
+      - published
 
 jobs:
-  release-job:
+  build-package:
     runs-on: ubuntu-latest
-    env:
-      PYPI_TOKEN: ${{ secrets.PYPI_TOKEN_PYMC }}
+    permissions:
+      # write attestations and id-token are necessary for attest-build-provenance-github
+      attestations: write
+      id-token: write
     steps:
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
-      with:
-        persist-credentials: false
-    - name: Set up Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: 3.7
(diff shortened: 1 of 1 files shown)


maresb:
Note that 61f59c372520af0cbab6724a43ac6d04ff288812 is irrelevant except to avoid merge conflicts with #7624. (It is cherry-picked from there.)

michaelosthege:
I created this 👇 environment, but I don't have management permissions on the PyPI package.

image

lucianopaz (approved):
Looks good to me. I had to google around quite a lot to understand the different actions and their configuration arguments, but they all make sense now and seem standard. Thanks @maresb!

maresb:
Thanks @michaelosthege for creating the `release` environment! Note that now I still need someone with management permissions on PyPI.

The deployment branches and tags look not quite correct. Since publishing should only take place for tagged releases, it's important to change the ref type to "Tag" instead of "Branch" when adding the pattern for branches and tags:
image

So `main` should not be allowed, and the `v*.*.*` pattern should match a whole lot of tags (instead of 0).

I think we can get by with unchecking "Allow administrators to bypass configured protection rules", but I'm not entirely sure.

Thanks @lucianopaz for checking everything so thoroughly! I felt bad that you had to google around, so I added more comments to hopefully make the context more clear.

maresb:
Thanks a lot to @fonnesbeck for getting PyPI and GitHub configured. This should be ready to go.

PR #7629, opened 2024-12-26
Add way to set backend `fn` random generators

### LINKED ISSUE #7588: trace fn does not respect random seed for stochastic "Deterministics"
### Description

```python
import pymc as pm

with pm.Model() as m:
    x = pm.Bernoulli("x", p=0.5) * 0 # Force it to be zero
    pm.Deterministic("y", x + pm.Normal.dist())

    idata1 = pm.sample(tune=0, draws=1, random_seed=1)
    idata2 = pm.sample(tune=0, draws=1, random_seed=1)
    
print(
    idata1.posterior["y"].sel(draw=0).values,
    idata2.posterior["y"].sel(draw=0).values,
    sep="\n"
)
# [-0.39110838 -0.10602868 -0.64066713  0.59270806]
# [ 1.30907711 -0.1391542   0.52679751  0.47189673]
```

We could pass the rng to the inner `compile_pymc`. But it's still a bit silly to compile the same function per chain. We could compile the function once and pass it to the trace. After #7578 this should be allowed. We just need to make sure we `copy` the function with new shared RNGs.

Alternatively, we could defer to `compute_deterministics`, and move the deterministics computations out of the sampling loop? This would come at a higher memory cost when users choose not to store certain free variables in the trace.

lucianopaz:
This PR depends on #7540. Do not merge before that one has been merged.

## Description
This PR adds the `set_function_rngs` function that takes a compiled pytensor function, looks for any random generators in it, and makes a copy that sets all generators to spawned versions of a supplied generator. The implementation is taken from @ricardoV94's [function here](https://github.com/pymc-devs/pymc-extras/blob/c1809e8149fc89ac6eadf4bf73050ea6fe82955c/pymc_experimental/sampling/optimizations/conjugate_sampler.py#L84-L98) (and that's why I listed him as coauthor of this commit). This function is then used in `init_traces` and base trace initialization to make the backend's `fn` have a properly seeded random generator. Thanks to this change, we can reproducible sampling results even when the function has `Deterministic` that depend on raw pytensor random variables.

## Related Issue

- [X] Closes #7588
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):


ricardoV94:
Need to raise an error if the linker is JAXLinker as `function.get_shared()` doesn't retrieve the shared RNGs that are actually used by JAX functions (they are incompatible with numpy Generators): https://github.com/pymc-devs/pymc-extras/blob/c1809e8149fc89ac6eadf4bf73050ea6fe82955c/pymc_experimental/sampling/optimizations/conjugate_sampler.py#L79-L82

ricardoV94:
Perhaps a more explicit name like `copy_function_with_new_rngs`? That gives the mental model that the old function is not affected?

lucianopaz:
> Need to raise an error if the linker is JAXLinker as `function.get_shared()` doesn't retrieve the shared RNGs that are actually used by JAX functions (they are incompatible with numpy Generators): https://github.com/pymc-devs/pymc-extras/blob/c1809e8149fc89ac6eadf4bf73050ea6fe82955c/pymc_experimental/sampling/optimizations/conjugate_sampler.py#L79-L82

I don't like errorring out here. Maybe a warning will be good for now, until we figure out how to control jax's random keys.

PR #7634, opened 2025-01-06
Show one progress bar per chain when sampling

jessegrabowski:
## Description

I really like what nutpie gives you while sampling, so I tried to make something using rich that copies it. Example:

https://github.com/user-attachments/assets/9c634f3e-933f-4d26-a954-a5da3d5e4098

Features are:

1. One progress bar per chain
2. Sampling statistics per chain. I copied nutpie, but we can haggle over what these should be (or give the user more control)
3. Color change based on status. Blue when sampling, turns red after a divergence. Finished bar is either green (no divergences) or purple (with divergences).

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7634.org.readthedocs.build/en/7634/

### DIFF
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -276,7 +276,7 @@ def _print_step_hierarchy(s: Step, level: int = 0) -> None:
     else:
         varnames = ", ".join(
             [
-                get_untransformed_name(v.name) if is_transformed_name(v.name) else v.name
+                get_untransformed_name(v.name) if is_transformed_name(v.name) else v.name  # type: ignore[misc]
                 for v in s.vars
             ]
         )
@@ -1238,7 +1238,7 @@ def _iter_sample(
     rng: np.random.Generator,
     model: Model | None = None,
     callback: SamplingIteratorCallback | None = None,
-) -> Iterator[bool]:
+) -> Iterator[list[dict[str, Any]]]:
     """Sample one chain with a generator (singleprocess).
 
     Parameters
@@ -1261,8 +1261,8 @@ def _iter_sample(
 
     Yields
     ------
-    diverging : bool
-        Indicates if the draw is divergent. Only available with some samplers.
+    stats : list of dict
+        Dictionary of statistics returned by step sampler
     """
     draws = int(draws)
 
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -302,7 +302,7 @@ def _sample_external_nuts(
     initvals: StartDict | Sequence[StartDict | None] | None,
     model: Model,
     var_names: Sequence[str] | None,
-    progressbar: bool | ProgressType,
+    progressbar: bool,
     idata_kwargs: dict | None,
     compute_convergence_checks: bool,
     nuts_sampler_kwargs: dict | None,
@@ -401,7 +401,7 @@ def _sample_external_nuts(
             initvals=initvals,
             model=model,
             var_names=var_names,
-            progressbar=True if progressbar else False,
+            progressbar=progressbar,
             nuts_sampler=sampler,
             idata_kwargs=idata_kwargs,
             compute_convergence_checks=compute_convergence_checks,
@@ -423,7 +423,7 @@ def sample(
     chains: int | None = None,
     cores: int | None = None,
     random_seed: RandomState = None,
-    progressbar: bool = True,
+    progressbar: bool | ProgressType = True,
     progressbar_theme: Theme | None = default_progress_theme,
     step=None,
     var_names: Sequence[str] | None = None,
@@ -455,7 +455,7 @@ def sample(
     chains: int | None = None,
     cores: int | None = None,
     random_seed: RandomState = None,
-    progressbar: bool = True,
+    progressbar: bool | ProgressType = True,
     progressbar_theme: Theme | None = default_progress_theme,
     step=None,
     var_names: Sequence[str] | None = None,
@@ -540,17 +540,16 @@ def sample(
         easy spawning of new independent random streams that are needed by the step methods.
     progressbar: bool or ProgressType, optional
             How and whether to display the progress bar. If False, no progress bar is displayed. Otherwise, you can ask
-            for either:
-            - "combined": A single progress bar that displays the progress of all chains combined.
-            - "chain": A separate progress bar for each chain.
+            for one of the following:
+            - "combined": A single progress bar that displays the total progress across all chains. Only timing
+                information is shown.
+            - "split": A separate progress bar for each chain. Only timing information is shown.
+            - "combined+stats" or "stats+combined": A single progress bar displaying the total progress across all
+                chains. Aggregate sample statistics are also displayed.
+            - "split+stats" or "stats+split": A separate progress bar for each chain. Sample statistics for each chain
+                are also displayed.
 
-            You can also combine the above options with:
-            - "simple": A simple progress bar that displays only timing information alongside the progress bar.
-            - "full": A progress bar that displays all available statistics.
-
-            These can be combined with a "+" delimiter, for example: "combined+full" or "chain+simple".
-
-            If True, the default is "chain+full".
+            If True, the default is "split+stats" is used.
     step : function or iterable of functions
         A step function or collection of functions. If there are variables without step methods,
         step methods for those variables will be assigned automatically. By default the NUTS step
@@ -716,6 +715,10 @@ def sample(
     if isinstance(trace, list):
         raise ValueError("Please use `var_names` keyword argument for partial traces.")
 
+    # progressbar might be a string, which is used by the ProgressManager in the pymc samplers. External samplers and
+    # ADVI initialization expect just a bool.
+    progress_bool = True if progressbar else False
+
     model = modelcontext(model)
     if not model.free_RVs:
         raise SamplingError(
@@ -812,7 +815,7 @@ def sample(
                 initvals=initvals,
                 model=model,
                 var_names=var_names,
-                progressbar=progressbar,
+                progressbar=progress_bool,
                 idata_kwargs=idata_kwargs,
                 compute_convergence_checks=compute_convergence_checks,
                 nuts_sampler_kwargs=nuts_sampler_kwargs,
@@ -831,9 +834,7 @@ def sample(
                 n_init=n_init,
                 model=model,
                 random_seed=random_seed_list,
-                progressbar=True
-                if progressbar
-                else False,  # ADVI doesn't use the ProgressManager; pass a bool only
+                progressbar=progress_bool,
                 jitter_max_retries=jitter_max_retries,
                 tune=tune,
                 initvals=initvals,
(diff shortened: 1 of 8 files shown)


pymc/util.py
```diff
@@ -1,4 +1,4 @@
-#   Copyright 2024 The PyMC Developers
+#   Copyright 2025 The PyMC Developers
```
ricardoV94: According to Oriol this should just be 20xx-present

ricardoV94 (commented):
How does this look like when 1) you have another step sampler in the mix and 2) there's no nuts at all or 3) there are more than one NUTS step samplers?

jessegrabowski:
https://github.com/user-attachments/assets/80d860c6-f593-4bdd-a628-ecadcb9ca123

Here's a comparison between NUTS and non-nuts sampler. 

Ideally we'd add a method to the step samplers themselves that would return the rich columns that sampler wants to use, then we just gather them and display. In that case you could even different sampler stats from different steps in the same run. Maybe it's worth doing. The actual code for this PR is pretty gnarly.

jessegrabowski:
I moved the responsibility for setting up the progressbars and updating stats to the step samplers. This means each step method can choose what stats are to be shown on the progress bars, and we can also combine them. Example vid attached.


https://github.com/user-attachments/assets/6f7dfaee-1e84-413c-9400-4c1b9d12c03f

This is a pretty big scope creep for this PR, so I'm not against reverting these changes and going with something more basic. If we like it though I can lean into it.

I will say it's broken right now because when you have e.g. multiple metropolis steps (one per variable) the only stats that get reported are the last one. It needs some logic on how to aggregate the stats across samplers with the same stats.

ricardoV94:
The step sampler specifics looks amazing 😍 Gonna give it a try today.

I'll test it but I assume things behave gracefully if the step samplers don't specify the display columns info?

jessegrabowski:
No it will break. I need to put in a default for the base class.

ricardoV94:
I would still like to see the global runtime and /eta like we had  before. Is that feasible or too ugly?

Re: repeated samplers, show the mean? Or maybe only display specialized info when a single step sampler is being used?

jessegrabowski:
I added the base impl, so things will go gracefully if there's no implementation. This would only show the NUTS stats for example, because there's no implementation for BinaryGibbsMetropolis:

```py
import pymc as pm

with pm.Model() as m:
    x = pm.Bernoulli('x', p=0.5)
    y = pm.Normal('y', mu=pm.math.switch(x, -3, 3), sigma=10, shape=(10,))
    
    idata = pm.sample(step=[pm.BinaryGibbsMetropolis(x), pm.NUTS(y)], tune=2000, draws=2000, chains=8, cores=8, compile_kwargs={'mode':'NUMBA'})
```

Re: global, yes we can keep it. But we can't have it as a single long bar that breaks the columns, because there's no colspan operator for rich tables (see https://github.com/Textualize/rich/issues/164).

We could make a separate table though. It just won't be as pretty as nutpie.

ricardoV94:
Sequential sampling (cores=1) still has the old approach. It has one bar per chain but not the stats

ricardoV94:
> Re: global, yes we can keep it. But we can't have it as a single long bar that breaks the columns, because there's no colspan operator for rich tables (see https://github.com/Textualize/rich/issues/164).

What if we show as a column per chain then? `elapsed/left`?

jessegrabowski:
New version with timing info:


https://github.com/user-attachments/assets/bd7f6af0-09e4-43e0-a7b6-fc44b628c214

ricardoV94:
> Sequential sampling (cores=1) still has the old approach. It has one bar per chain but not the stats

Did you address this?

jessegrabowski:
Not yet, but it will be an easy fix.

ricardoV94:
Some failing tests as well. Otherwise I'm happy with the changes. I'll paste in the discord to see if anybody has big complaints

aloctavodia:
This is looks really nice and modern and its very informative, but do we have an option for a single progress bar with less information.

jessegrabowski:
We can do that painlessly yeah

twiecki:
This looks great. I assume blue means tuning and red means post-tuning? If so, I wonder if red is the best color choice as it suggests something gone wrong. Maybe replace red with green? Or make tuning red and sampling blue?

ricardoV94:
> This looks great. I assume blue means tuning and red means post-tuning? If so, I wonder if red is the best color choice as it suggests something gone wrong. Maybe replace red with green? Or make tuning red and sampling blue?

It turns red if there's any divergence

jessegrabowski:
It turns red once the chain hits a divergence, otherwise it stays blue and turns green when its done

twiecki:
I see, maybe then green post-tuning without divergences? Or maybe a non-colorblind color.

aloctavodia:
If you want some colorblind-friendly palletes https://github.com/arviz-devs/arviz-plots/tree/main/src/arviz_plots/styles

fonnesbeck:
I like it a lot.

I don't know if we need different colors for pre-/post-tuning. 

Can we get red for any warning (so that it makes users read the warning)?

Definitely no green if we are using red. I like blue/red for clean/warning.

ricardoV94:
I think 2 colors is enough. It will be clear when you use compared to a gif. Otherwise a single color.

@fonnesbeck what sort of warnings are you thinking about? Are they emmited during sampling or only at the end?

fonnesbeck:
Things like max tree depth, missing the target acceptance rate and r-hat. Those that are  calculated at the end could flip the line to red upon completion.

Not  a big deal, though. What's here is a big improvement!

jessegrabowski:
https://github.com/user-attachments/assets/835d8c97-b2ec-4144-bdea-0741fdc1ad93

Here's how it looks after the last round of feedback/requests. I don't love the string parsing on the `progressbar` argument, but I also don't love more keyword arguments on `pm.sample`.

pymc/util.py
```diff
@@ -607,6 +622,257 @@ def update(
...
+    def __init__(self, step_method, chains, draws, tune, progressbar, progressbar_theme):
+        mode = "chain"
+        stats = "full"
+
+        if isinstance(progressbar, bool):
+            show_progress = progressbar
+        else:
+            show_progress = True
+
+            if "+" in progressbar:
```
ricardoV94: Perhaps more readable / less bug-prone to if/elif all the valid cases?

jessegrabowski: Yeah definitely. Or replace `progress_theme` with `progress_kwargs` that takes `compact: bool, show_fewer_stats: bool, progress_theme: whatever` ?

ricardoV94: I think that's harder to use

pymc/util.py
```diff
@@ -704,28 +704,38 @@ def callbacks(self, task: "Task"):
...
-
-        if mode not in ["chain", "combined"]:
-            raise ValueError('Invalid mode. Valid values are "chain" and "combined"')
-        if stats not in ["full", "simple"]:
-            raise ValueError('Invalid stats. Valid values are "full" and "simple"')
+        self.combined_progress = False
+        self.full_stats = True
+        show_progress = True
+
+        match progressbar:
```
ricardoV94, line 771: first use of match in pymc I think!

ricardoV94 (approved):

tomicapretto:
I think we need to update the docstring here 

https://github.com/jessegrabowski/pymc/blob/a629a9703be2023217de108ea201a139124c2387/pymc/sampling/mcmc.py#L540-L544

tomicapretto:
Have you noticed **Draws** stops at something smaller than the total number of draws? 

image

You can also see it in your last video

image

ricardoV94:
Usually needs one more call at the end, weren't we doing that before already?

jessegrabowski:
It only happens in the collapsed mode, I have to fix the logic there.

juanitorduz:
This looks great 🚀 !

jessegrabowski:
I want to rename the options for the progress bar to be less random, I was thinking about:

- `one_bar` for the combined mode
- `per_chain` for the one bar per chain mode
- `full_stats` for show all the stats possible
- `fewer_stats` for show timing only

Any thoughts?

ricardoV94:
combined/split+stats:
*combined
*split
*combined+stats
*split+stats

pymc/sampling/mcmc.py
```diff
@@ -709,6 +715,10 @@ def sample(
     if isinstance(trace, list):
         raise ValueError("Please use `var_names` keyword argument for partial traces.")
 
+    # progressbar might be a string, which is used by the ProgressManager in the pymc samplers. External samplers and
+    # ADVI initialization expect just a bool.
+    progress_bool = True if progressbar else False
```
ricardoV94: Nit

```suggestion
    progress_bool = bool(progressbar)
```

pymc/util.py
```diff
@@ -30,11 +30,35 @@
...
+from rich.table import Column, Table
 from rich.theme import Theme
 
 from pymc.exceptions import BlockModelAccessError
 
+if TYPE_CHECKING:
+    from pymc.step_methods.compound import BlockedStep, CompoundStep
+
+
+ProgressType = Literal[
```
ricardoV94:
```suggestion
ProgressBarType = Literal[
```

pymc/util.py
```diff
@@ -607,6 +635,314 @@ def update(
...
+        divergences = task.fields.get("divergences", 0)
+        if isinstance(divergences, float | int) and divergences > 0:
+            self.complete_style = Style.parse("rgb({},{},{})".format(*self.diverging_rgb))
+            self.finished_style = Style.parse("rgb({},{},{})".format(*self.diverging_rgb))
+        else:
+            self.complete_style = self.non_diverging_style
+            self.finished_style = self.non_diverging_finished_style
+
+
+class ProgressManager:
```
ricardoV94:
```suggestion
class ProgressBarManager:
```

pymc/util.py
```diff
@@ -607,6 +635,314 @@ def update(
...
+        match progressbar:
+            case True:
+                show_progress = True
+            case False:
+                show_progress = False
+            case "combined":
+                self.combined_progress = True
+                self.full_stats = False
+            case "split":
+                self.combined_progress = False
```
ricardoV94, line 785: Nit: It's a bit arbitrary that in some match cases you redefine defaults and others not

fonnesbeck:
> I want to rename the options for the progress bar to be less random, I was thinking about:
> 
> * `one_bar` for the combined mode
> * `per_chain` for the one bar per chain mode
> * `full_stats` for show all the stats possible
> * `fewer_stats` for show timing only
> 
> Any thoughts?

Maybe instead of `full_stats` vs `fewer_stats` just have a `verbosity` and have 0,1,2 as options?

pymc/util.py
jessegrabowski: only the boolean cases don't redefine. Would you prefer everything is defined in every case? No objection, I was just trying to save some lines

ricardoV94: Either be verbose or save all you can then, don't redefine any "defaults"

jessegrabowski: You'd make a poor King Solomon

jessegrabowski:
> Maybe instead of `full_stats` vs `fewer_stats` just have a `verbosity` and have 0,1,2 as options?

I wanted to avoid having additional keyword arguments to `pm.sample`, because I agree with @cluhmann in general that's its way too bloated.

ricardoV94 (approved):

tomicapretto:
Congrats @jessegrabowski!

PR #7641, opened 2025-01-12
Check for observed variables in the trace

zaxtax:
## Description
This introduces the enhancement discussed in https://github.com/pymc-devs/pymc/discussions/7225

This makes it easier to use `sample_posterior_predictive` in model factory workflows

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7641.org.readthedocs.build/en/7641/

### DIFF
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -820,7 +820,7 @@ def sample_posterior_predictive(
     else:
         vars_ = model.observed_RVs + observed_dependent_deterministics(model)
         if observed_data is not None:
-            vars_ += [model[x] for x in observed_data if x in model]
+            vars_ += [model[x] for x in observed_data if x in model and x not in vars_]
 
     vars_to_sample = list(get_default_varnames(vars_, include_transformed=False))
 
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -775,7 +775,7 @@ def sample_posterior_predictive(
             trace_coords.update({str(k): v.data for k, v in _constant_data.coords.items()})
             constant_data.update({str(k): v.data for k, v in _constant_data.items()})
         idata = trace
-        observed_data = trace["observed_data"]
+        observed_data = trace.get("observed_data", None)
         trace = trace["posterior"]
     if isinstance(trace, xarray.Dataset):
         trace_coords.update({str(k): v.data for k, v in trace.coords.items()})
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -821,10 +821,12 @@ def sample_posterior_predictive(
     if var_names is not None:
         vars_ = [model[x] for x in var_names]
     else:
-        vars_ = model.observed_RVs + observed_dependent_deterministics(model)
+        observed_vars = model.observed_RVs
         if observed_data is not None:
-            vars_ += [model[x] for x in observed_data if x in model and x not in vars_]
-            vars_ += observed_dependent_deterministics(model, vars_)
+            observed_vars += [
+                model[x] for x in observed_data if x in model and x not in observed_vars
+            ]
+        vars_ = observed_vars + observed_dependent_deterministics(model, observed_vars)
 
     vars_to_sample = list(get_default_varnames(vars_, include_transformed=False))
 
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -767,6 +767,7 @@ def sample_posterior_predictive(
     if "coords" not in idata_kwargs:
         idata_kwargs["coords"] = {}
     idata: InferenceData | None = None
+    observed_data = None
     stacked_dims = None
     if isinstance(trace, InferenceData):
         _constant_data = getattr(trace, "constant_data", None)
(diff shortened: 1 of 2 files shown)


tests/sampling/test_forward.py
```diff
@@ -481,6 +481,16 @@ def test_normal_scalar(self):
...
 
+        # test that trace is used in ppc
+        with pm.Model() as model_ppc:
+            mu = pm.Normal("mu", 0.0, 1.0)
+            a = pm.Normal("a", mu=mu, sigma=1)
+
+        ppc = pm.sample_posterior_predictive(
+            trace=trace, model=model_ppc, return_inferencedata=False
+        )
+        assert "a" in ppc
```
ricardoV94: Can you put this in its own test?

pymc/sampling/forward.py
```diff
@@ -817,6 +819,8 @@ def sample_posterior_predictive(
         vars_ = [model[x] for x in var_names]
     else:
         vars_ = model.observed_RVs + observed_dependent_deterministics(model)
+        if observed_data is not None:
```
ricardoV94, line 825: BTW, the the `observed_dependent_deterministics` above is not going to work if these variables are not observed in the model.

That happens with auto-imputation models, which I assume the as_model wrapper won't handle correctly either because the models are different depending on whether you pass data or not.

Just something to keep in mind

zaxtax:
Can you give an example of an auto-imputation model?

pymc/sampling/forward.py
lucianopaz: I agree with this. You'll have to adapt `observed_dependent_deterministics` to also accept a list of extra variables that will depend on your `observed_data`

pymc/sampling/forward.py
```diff
@@ -817,6 +819,8 @@ def sample_posterior_predictive(
...
+            vars_ += [model[x] for x in observed_data if x in model]
```
lucianopaz: maybe also add `and if x not in vars_` to the list comprehension.

tests/sampling/test_forward.py
lucianopaz: I would also make the test more stringent. Test that only the variables that you want are actually included in the trace. Also add a case where one node is conditionally dependent on the `trace.observed_data` so that you see that the auto added variables include conditional nodes.

lucianopaz (changes requested):
I second Ricardo with his comments. Just a few changes and I think this will be ok to merge.

zaxtax:
Thanks! Can you give an example model where a node is conditionally
dependent?

ricardoV94:
That helper was created specifically for the deterministic created by automatic imputation that joins the observed and unobserved components, so I'm not too worried about it.

But if you wanted just add a deterministic that's like y + 1, where y is the variable that was observed during sampling

zaxtax:
It's not fully clear the best way to make sure the correct conditional nodes are added when the model that produced the trace isn't readily available. So I have written a test we expect to fail in the meanwhile.

pymc/sampling/forward.py
```diff
@@ -821,6 +824,7 @@ def sample_posterior_predictive(
         vars_ = model.observed_RVs + observed_dependent_deterministics(model)
         if observed_data is not None:
             vars_ += [model[x] for x in observed_data if x in model and x not in vars_]
+            vars_ += observed_dependent_deterministics(model, vars_)
```
ricardoV94: This is going to duplicate deterministics of an observed variable, in case there's a mix of observed model variables and implied observed variables from the idata. the `+ observed_dependent_determininstics)` should be called only once after the if branch?

zaxtax:
@lucianopaz does this address your concerns?

tests/sampling/test_forward.py
```diff
@@ -540,6 +540,50 @@ def test_normal_scalar_idata(self):
             ppc = pm.sample_posterior_predictive(idata, return_inferencedata=False)
             assert ppc["a"].shape == (nchains, ndraws)
 
+    def test_external_trace(self):
```
ricardoV94: Remove this test? The one with `det` is strictly more comprehensive than this?

tests/sampling/test_forward.py
```diff
@@ -540,6 +540,50 @@ def test_normal_scalar_idata(self):
...
+            a = pm.Normal("a", mu=mu, sigma=1)
+
+        ppc = pm.sample_posterior_predictive(
+            trace=trace, model=model_ppc, return_inferencedata=False
+        )
+        assert list(ppc.keys()) == ["a"]
+
+    def test_external_trace_det(self):
+        nchains = 2
+        ndraws = 500
```
ricardoV94, line 566: We don't care about the actual draws

```suggestion
        tune=50
        nchains = 1
        ndraws = 50
```

tests/sampling/test_forward.py
```diff
@@ -540,6 +540,50 @@ def test_normal_scalar_idata(self):
...
+        with pm.Model() as model:
+            mu = pm.Normal("mu", 0.0, 1.0)
+            a = pm.Normal("a", mu=mu, sigma=1, observed=0.0)
+            b = pm.Deterministic("b", a + 1)
+            trace = pm.sample(
+                draws=ndraws,
+                chains=nchains,
+            )
```
ricardoV94: Since we are not using the draws for anything

```suggestion
        with pm.Model() as model:
            mu = pm.Normal("mu", 0.0, 1.0)
            a = pm.Normal("a", mu=mu, sigma=1, observed=0.0)
            b = pm.Deterministic("b", a + 1)
            trace = pm.sample(
               tune=50,
                draws=50,
                chains=1,
                compute_convergence_checks=False
            )
```

ricardoV94 (approved):

PR #7647, opened 2025-01-15
Use updated `head_of_apache` pre-commit step

lucianopaz:
This PR re-enables the pre commit step that applies the Apache v2 license header to all files. The updated version now always renders a year range, and the end year is set to the current year (as long as it has not been manually set to present)

----
📚 Documentation preview 📚: https://pymc--7647.org.readthedocs.build/en/7647/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,16 +37,17 @@ repos:
   hooks:
     - id: sphinx-lint
       args: ["."]
-#- repo: https://github.com/lucianopaz/head_of_apache
-#  rev: "0.0.3"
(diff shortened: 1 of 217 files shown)


lucianopaz:
> I thought the ideal was start-present, like 2022-present? CC @OriolAbril

I set it to the actual year on purpose. The hook will update the end year to the current year automatically. If you prefer, I can add an argument to use the `start - present` format instead.

ricardoV94:
It also has the advantage we don't need a commit on every file every year

ricardoV94 (approved):

PR #7652, opened 2025-01-20
fix: deep copy nuts_sampler_kwarg to prevent `.pop` side effects

### LINKED ISSUE #7632: Use of `dict.pop` in `pm.sample` leads to side effects
### Description

I commonly create a `sample_kwargs` variable in notebooks and re-use it in multiple models. Such a dictionary ends up being modified in-place by `pm.sample` when entries are popped, for example [here ](https://github.com/pymc-devs/pymc/blob/ae4a6292aa48d5722dba6ab422e1a3d895ca7bf7/pymc/sampling/mcmc.py#L336) and [here](https://github.com/pymc-devs/pymc/blob/ae4a6292aa48d5722dba6ab422e1a3d895ca7bf7/pymc/sampling/mcmc.py#L689), leading to models beyond the first silently using defaults settings.

I guess the easiest fix would just be to `deepcopy` `kwargs`, so that there are no side effects on user inputs. Using `pop` here is a bit overkill (why not just `,get`?) but `deepcopy` is a 1-line change.

inclinedadarsh:
## Description

Deep copy the `nuts_sampler_kwargs` to prevent the function (`.pop` method) mutating the original passed dictonary.

## Related Issue

- [x] Closes #7632 

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7652.org.readthedocs.build/en/7652/

### DIFF
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -309,7 +309,6 @@ def _sample_external_nuts(
     nuts_sampler_kwargs: dict | None,
     **kwargs,
 ):
-    nuts_sampler_kwargs = nuts_sampler_kwargs.copy()
     if nuts_sampler_kwargs is None:
         nuts_sampler_kwargs = {}
 
@@ -339,6 +338,7 @@ def _sample_external_nuts(
                 UserWarning,
             )
         compile_kwargs = {}
+        nuts_sampler_kwargs = nuts_sampler_kwargs.copy()
         for kwarg in ("backend", "gradient_backend"):
             if kwarg in nuts_sampler_kwargs:
                 compile_kwargs[kwarg] = nuts_sampler_kwargs.pop(kwarg)
@@ -687,7 +687,6 @@ def sample(
             mean     sd  hdi_3%  hdi_97%
         p  0.609  0.047   0.528    0.699
     """
-    nuts_sampler_kwargs = nuts_sampler_kwargs.copy()
     if "start" in kwargs:
         if initvals is not None:
             raise ValueError("Passing both `start` and `initvals` is not supported.")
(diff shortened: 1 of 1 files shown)


pymc/sampling/mcmc.py
```diff
@@ -339,8 +342,8 @@ def _sample_external_nuts(
             )
         compile_kwargs = {}
         for kwarg in ("backend", "gradient_backend"):
-            if kwarg in nuts_sampler_kwargs:
-                compile_kwargs[kwarg] = nuts_sampler_kwargs.pop(kwarg)
+            if kwarg in nuts_sampler_kwargs_copy:
```
ricardoV94: Shallow copy seems enough and you can simply override the original variable. `nuts_sampler_kwargs = nuts_sampler_kwargs.copy()`.

This should allow a much more concise fix. 

It's also only needed on the branch where we pop

ricardoV94 (changes requested):

pymc/sampling/mcmc.py
inclinedadarsh: Hey,
Thank you for the comment. I have just pushed a commit resolving this.

ricardoV94:
The kwargs.pop in `sample` is fine, because it comes from `**kwargs`, it doesn't need any copy. 

The only one that needs is the `nuts_sample_kwargs`, and only in the `nutpie` branch before we start popping it.
Just add the `nuts_sample_kwargs = nuts_sample_kwargs.copy()` line here: https://github.com/pymc-devs/pymc/blob/e6c5b1a20996e4ba5ac5cd54305dea8b5a8ee6ea/pymc/sampling/mcmc.py#L341

inclinedadarsh:
Thank you for correcting me, @ricardoV94; it means a lot to me.
I now understand why we just need the copy in the nutpie branch.

Thank you!

ricardoV94 (approved):

ricardoV94:
Thanks @inclinedadarsh

Once the tests pass we can merge

PR #7656, opened 2025-01-24
Support PyTensor deterministic operations as observations

### LINKED ISSUE #7649: BUG: data as observed in RV
### Describe the issue:

Example:

https://github.com/pymc-labs/pymc-marketing/issues/1383#issuecomment-2595717439

```python
import numpy as np
import pymc as pm

seed = sum(map(ord, "Scaling the likelihood depended variables doesn't work in PyMC"))
rng = np.random.default_rng(seed)

true_mu = 100
true_sigma = 30

n_obs = 10
coords = {
    "date": np.arange(n_obs),
}

dist = pm.Normal.dist(mu=true_mu, sigma=true_sigma, shape=n_obs)
data = pm.draw(dist, random_seed=rng)

scaling = data.max()

    mu = pm.Normal("mu")

    target = pm.Data("target", data, dims="date")
    scaled_target = target / scaling

    pm.Normal("observed", mu=mu, sigma=sigma, observed=scaled_target)
```

### Error message

```
            if not is_valid_observed(observed):
>               raise TypeError(
                    "Variables that depend on other nodes cannot be used for observed data."
                    f"The data variable was: {observed}"
                )
E               TypeError: Variables that depend on other nodes cannot be used for observed data.The data variable was: True_div.0
```

### PyMC version information:

5.20.0

### Context for the issue:

Would like to define my scaling within the model definition

williambdean:
## Description

Will need some help on this implementation and how to best test this.

## Related Issue

- [x] Closes #7649 
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7656.org.readthedocs.build/en/7656/

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -190,7 +190,7 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size: 
     for i, v in enumerate(tensors):
         if not is_valid_observed(v):
             raise ValueError(
-                f"{i}: {v} is not valid for Minibatch, only constants or constants.astype(dtype) are allowed"
+                f"{i}: {v} is not valid for Minibatch, only non-random variables are allowed"
             )
 
     upper = tensors[0].shape[0]
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -33,6 +33,7 @@ from pytensor.raise_op import Assert
 from pytensor.scalar import Cast
 from pytensor.tensor.elemwise import Elemwise
 from pytensor.tensor.random.basic import IntegersRV
+from pytensor.tensor.random.var import RandomGeneratorSharedVariable
 from pytensor.tensor.type import TensorType
 from pytensor.tensor.variable import TensorConstant, TensorVariable
 
@@ -148,6 +149,19 @@ class MinibatchOp(OpFromGraph):
         return "Minibatch"
 
 
+def first_inputs(r):
+    if not r.owner:
+        return
+
+    first_input = r.owner.inputs[0]
+    yield first_input
+    yield from first_inputs(first_input)
+
+
+def has_random_ancestor(r):
+    return any(isinstance(i, RandomGeneratorSharedVariable) for i in first_inputs(r))
+
+
 def is_valid_observed(v) -> bool:
     if not isinstance(v, Variable):
         # Non-symbolic constant
@@ -165,6 +179,7 @@ def is_valid_observed(v) -> bool:
             and isinstance(v.owner.op.scalar_op, Cast)
             and is_valid_observed(v.owner.inputs[0])
         )
+        or not has_random_ancestor(v)
         # Or Minibatch
         or (
             isinstance(v.owner.op, MinibatchOp)
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -30,15 +30,13 @@ from pytensor.compile.builders import OpFromGraph
 from pytensor.compile.sharedvalue import SharedVariable
 from pytensor.graph.basic import Variable
 from pytensor.raise_op import Assert
-from pytensor.scalar import Cast
-from pytensor.tensor.elemwise import Elemwise
 from pytensor.tensor.random.basic import IntegersRV
-from pytensor.tensor.random.var import RandomGeneratorSharedVariable
 from pytensor.tensor.type import TensorType
 from pytensor.tensor.variable import TensorConstant, TensorVariable
 
 import pymc as pm
 
+from pymc.logprob.utils import rvs_in_graph
 from pymc.pytensorf import GeneratorOp, convert_data, smarttypeX
 from pymc.vartypes import isgenerator
 
@@ -149,22 +147,8 @@ class MinibatchOp(OpFromGraph):
         return "Minibatch"
 
 
-def first_inputs(r):
-    if not r.owner:
-        return
-
-    inputs = r.owner.inputs
-
-    if not inputs:
-        return
-
-    first_input = inputs[0]
-    yield first_input
-    yield from first_inputs(first_input)
-
-
 def has_random_ancestor(r):
-    return any(isinstance(i, RandomGeneratorSharedVariable) for i in first_inputs(r))
+    return rvs_in_graph(r) != set()
 
 
 def is_valid_observed(v) -> bool:
@@ -177,14 +161,7 @@ def is_valid_observed(v) -> bool:
         return True
 
     return (
-        # The only PyTensor operation we allow on observed data is type casting
-        # Although we could allow for any graph that does not depend on other RVs
-        (
-            isinstance(v.owner.op, Elemwise)
-            and isinstance(v.owner.op.scalar_op, Cast)
-            and is_valid_observed(v.owner.inputs[0])
-        )
-        or not has_random_ancestor(v)
+        not has_random_ancestor(v)
         # Or Minibatch
         or (
             isinstance(v.owner.op, MinibatchOp)
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -153,7 +153,12 @@ def first_inputs(r):
     if not r.owner:
         return
 
-    first_input = r.owner.inputs[0]
+    inputs = r.owner.inputs
+
+    if not inputs:
+        return
+
+    first_input = inputs[0]
     yield first_input
     yield from first_inputs(first_input)
 
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -147,10 +147,6 @@ class MinibatchOp(OpFromGraph):
         return "Minibatch"
 
 
-def has_random_ancestor(r):
-    return rvs_in_graph(r) != set()
-
-
 def is_valid_observed(v) -> bool:
     if not isinstance(v, Variable):
         # Non-symbolic constant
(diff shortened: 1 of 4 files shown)


pymc/data.py
```diff
@@ -148,6 +149,24 @@ def __str__(self):
...
+    if not inputs:
+        return
+
+    first_input = inputs[0]
+    yield first_input
+    yield from first_inputs(first_input)
+
+
+def has_random_ancestor(r):
+    return any(isinstance(i, RandomGeneratorSharedVariable) for i in first_inputs(r))
```
ricardoV94: Check: https://github.com/pymc-devs/pymc/blob/fa43eba8d682fbf4039bbd1c228c940c872e2d5d/pymc/logprob/utils.py#L133-L150

pymc/data.py
```diff
@@ -165,6 +184,7 @@ def is_valid_observed(v) -> bool:
             and isinstance(v.owner.op.scalar_op, Cast)
             and is_valid_observed(v.owner.inputs[0])
         )
+        or not has_random_ancestor(v)
```
ricardoV94: The previous branch is now a subset of the new more general behavior

ricardoV94:
We need to tell InferenceData converter how to get the data values to put in `constant_data`, and `observed_data`. It had some hardcode logic assuming you could only due Casts

ricardoV94:
I don't consider this a bugfix, it's behavior that was explicitly forbidden for conservative reasons (more like a NotImplementedError)

pymc/data.py
williambdean: Thanks for pointing out. I've changed to use this

pymc/data.py
williambdean: Removed

wd60622:
> We need to tell InferenceData converter how to get the data values to put in `constant_data`, and `observed_data`. It had some hardcode logic assuming you could only due Casts

Where does this happen in the code?

wd60622:
I think the MiniBatch tests will still fail. Any thoughts on how that should behave?

wd60622:
Also, feel free to choose a better title! I couldn't express it too well

tests/test_data.py
```diff
@@ -623,3 +623,14 @@ def test_multiple_vars(self):
...
+
+def test_scaling_data_works_in_likelihood() -> None:
+    data = np.array([10, 11, 12, 13, 14, 15])
+
+    with pm.Model() as model:
+        target = pm.Data("target", data)
+        scale = 12
+        scaled_target = target / scale
+        mu = pm.Normal("mu", mu=0, sigma=1)
+        pm.Normal("x", mu=mu, sigma=1, observed=scaled_target)
```
williambdean, line 549: Should I sample this to check that it has the correct data in the InferenceData?

ricardoV94: No, we have more direct ways of testing it

ricardoV94:
> > We need to tell InferenceData converter how to get the data values to put in `constant_data`, and `observed_data`. It had some hardcode logic assuming you could only due Casts
> 
> Where does this happen in the code?

https://github.com/pymc-devs/pymc/blob/fa43eba8d682fbf4039bbd1c228c940c872e2d5d/pymc/backends/arviz.py#L63

https://github.com/pymc-devs/pymc/blob/fa43eba8d682fbf4039bbd1c228c940c872e2d5d/pymc/pytensorf.py#L152

wd60622:
> We should be able to just use constant_fold (also in pytensor) for it

I have an implementation before this suggestion. I need a little help understanding this. Feel free to put suggestion

ricardoV94:
>  > We should be able to just use constant_fold (also in pytensor) for it
> 
> I have an implementation before this suggestion. I need a little help understanding this. Feel free to put suggestion

What do you mean? My suggestion is to simply call constant_fold on the observed variable

wd60622:
I am not able to make this work with either `constant_fold(x)` or `constant_fold([x])`. Both return errors.

ricardoV94:
> I am not able to make this work with either `constant_fold(x)` or `constant_fold([x])`. Both return errors. Is the constant_fold from pymc.pytensorf or from pytensor?

From pymc.pytensorf. There's a `raise_if_not_constant` flag you can set, but what cases is it failing?

ricardoV94:
```python
from pytensor.compile.mode import Mode
import pymc as pm

with pm.Model(coords={"date": [0, 1, 2]}) as m:
    data = pm.Data("data", [0, 1, 2], dims="date")
    x = pm.Normal("x")
    y = pm.Normal("y", x, observed=data, dims="date")

cheap_eval_mode = Mode(linker="py", optimizer=None)
m.rvs_to_values[y].eval(mode=cheap_eval_mode)
```

wd60622:
Thanks for the suggestion. I've used mode on the random variable directly. Is it different to get the rvs_to_values from the model? it currently isn't accessed from the function at the moment.

wd60622:
I've just added to associated pytensorf test suite

pymc/data.py
```diff
@@ -148,6 +147,10 @@ def __str__(self):
         return "Minibatch"
 
 
+def has_random_ancestor(r):
+    return rvs_in_graph(r) != set()
```
ricardoV94:
```suggestion
    return bool(rvs_in_graph(r))
```
But then you might as well skip the one liner and just do `not rvs_in_graph(r)` below?

pymc/pytensorf.py
```diff
@@ -177,6 +177,12 @@ def extract_obs_data(x: TensorVariable) -> np.ndarray:
             mask[mask_idx] = 1
             return np.ma.MaskedArray(array_data, mask)
 
+    from pymc.data import has_random_ancestor
+
+    if not has_random_ancestor(x):
+        cheap_eval_mode = Mode(linker="py", optimizer=None)
+        return x.eval(mode=cheap_eval_mode)
```
ricardoV94, line 171: Perhaps also check that `pm.pytensorf.inputvars(x)` is empty. A counter-example would be an explicit input in the graph, say `pt.scalar("x")`, that will fail with `MissingInput` when you try this eval.

tests/test_data.py
ricardoV94: Well maybe. Just make sure to do a cheap sampling, since we don't care about draws at all?

pymc/pytensorf.py
williambdean: Good sugguestion. I've added that and a test for it

pymc/data.py
williambdean: Done

tests/test_data.py
williambdean: I've tested the "extract_..." function directly

pymc/pytensorf.py
ricardoV94: Can you check how much time the creation of the cheap eval mode takes? Wondering if we should define it only once outside or if it's cheap enough that it doesn't matter

wd60622:
There are some failing tests related to the Mini batch also using the is_valid_observed function as well. How should those be handled. Are those test tests no longer valid?

tests/test_data.py
ricardoV94: Yeah that's a more direct unit test, this would be a CI, we don't care how it's done just want to be sure the data is there in the end?

ricardoV94:
> There are some failing tests related to the Mini batch also using the is_valid_observed function as well. How should those be handled. Are those test tests no longer valid?

Yeah sounds like the test should be changed to the more strict kind of stuff that we still don't allow, like having a pt.random.normal in it.

tests/test_data.py
```diff
@@ -602,11 +602,11 @@ def test_allowed(self):
         mb = pm.Minibatch(pt.as_tensor(self.data).astype(int), batch_size=20)
         assert isinstance(mb.owner.op, MinibatchOp)
 
-        with pytest.raises(ValueError, match="not valid for Minibatch"):
-            pm.Minibatch(pt.as_tensor(self.data) * 2, batch_size=20)
+        mb = pm.Minibatch(pt.as_tensor(self.data) * 2, batch_size=20)
```
ricardoV94, line 512: We should leave a case that fails

wd60622:
Looks like a random test failure?

Will address rest of feedback when I have chance

ricardoV94:
> Looks like a random test failure?

Yeah that one is flaky, we should probably try/except on that error and just skip

cetagostini:
Are we far from merge here?

tests/test_data.py
williambdean: I've add a case that tries to pass pt.random.normal data

wd60622:
Hi @ricardoV94 
Any modifications or additional tests needed for this?

tests/test_data.py
ricardoV94: I still want some sampling here to check the outputs? This test is not testing anything explicitly at the moment

williambdean: I will add sampling to this test when I can

williambdean: done

williambdean: Any changes needed here?

ricardoV94 (approved):

ricardoV94:
Thanks @wd60622 and apologies for the delay

PR #7666, opened 2025-01-30
Allow debug evaling IR logp graphs

ricardoV94:
A user on discourse recently raised the question of why we can't eval logp graphs during their construction. We have some tagging Ops that raise on the perform method, that prevent this. 

https://discourse.pymc.io/t/how-to-implement-bivariate-poisson-logbp/16400/7?u=ricardov94

This PR allows evaluating them, but issues a warning. I've felt the need to get values in interactive debug many times.

----
📚 Documentation preview 📚: https://pymc--7666.org.readthedocs.build/en/7666/

### DIFF
--- a/pymc/logprob/abstract.py
+++ b/pymc/logprob/abstract.py
@@ -236,13 +236,16 @@ class ValuedRV(Op):
     and breaking the dependency of `b` on `a`. The new nodes isolate the graphs between conditioning points.
     """
 
+    view_map = {0: [0]}
+
     def make_node(self, rv, value):
         assert isinstance(rv, Variable)
         assert isinstance(value, Variable)
         return Apply(self, [rv, value], [rv.type(name=rv.name)])
 
     def perform(self, node, inputs, out):
-        raise NotImplementedError("ValuedVar should not be present in the final graph!")
+        warnings.warn("ValuedVar should not be present in the final graph!")
+        out[0][0] = inputs[0]
 
     def infer_shape(self, fgraph, node, input_shapes):
         return [input_shapes[0]]
--- a/pymc/logprob/transform_value.py
+++ b/pymc/logprob/transform_value.py
@@ -11,6 +11,7 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+import warnings
 
 from collections.abc import Sequence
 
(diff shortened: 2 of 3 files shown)


pymc/logprob/abstract.py
```diff
@@ -236,13 +236,16 @@ class ValuedRV(Op):
     and breaking the dependency of `b` on `a`. The new nodes isolate the graphs between conditioning points.
     """
 
+    view_map = {0: [0]}
```
ricardoV94, line 239: Needs view_map so PyTensor knows not to do inplace operations on this. The other Op already had the view_map defined

pymc/logprob/abstract.py
```diff
@@ -236,13 +236,16 @@ class ValuedRV(Op):
...
+
     def make_node(self, rv, value):
         assert isinstance(rv, Variable)
         assert isinstance(value, Variable)
         return Apply(self, [rv, value], [rv.type(name=rv.name)])
 
     def perform(self, node, inputs, out):
-        raise NotImplementedError("ValuedVar should not be present in the final graph!")
+        warnings.warn("ValuedVar should not be present in the final graph!")
```
jessegrabowski, line 247: should we be logging instead of warning so people can more easily turn it off?

pymc/logprob/abstract.py
jessegrabowski: Interesting that you have to use view_map to do that. Somewhat hidden functionality? I thought pytensor only does inplace on things that proactively define a destroy_map.

jessegrabowski (approved):
lgtm, just some questions

pymc/logprob/abstract.py
ricardoV94: No, I want this to be loud, because it shouldn't be in the final graph.

pymc/logprob/abstract.py
ricardoV94: If I don't say this output returns a view of the input, and this output is not itself a graph output, PyTensor will assume it can be safely destroyed as it is just an intermediate operation.

ricardoV94:
Your messages are coming in weird place so I can't reply. 

You are reading it a bit backwards. By default an Op must never return a view or mutate an input in place. If you have an Op that does that, that's the "special case", and you have to tell PyTensor about it. That's the contract.

Otherwise every Op would have to be explicit about view_map, and most times would be `view_map={}`. That's fine but would be a bit of a pain in the ass to deprecate / refactor.

PR #7680, opened 2025-02-13
remove find_contrained_prior

aloctavodia:
`find_contrained_prior` was deprecated in v5.17.0 (https://github.com/pymc-devs/pymc/pull/7458), released on Oct 3, 2024. This PR removes it completely. 

----
📚 Documentation preview 📚: https://pymc--7680.org.readthedocs.build/en/7680/


twiecki:
Can you link the issue about the deprecation?

ricardoV94:
How long ago was v5.17 released? Time is more important than version numbers

aloctavodia:
Comment updated

twiecki:
I now I should have commented on the original issue but can't we add the max entropy condition to `find_constrained_prior`? I'm not even sure what the alternative workflow is to do this with `preliz` but it seems to require it's own, different set of distributions.

aloctavodia:
PreliZ's distributions have a `to_pymc` method

twiecki:
How difficult would it be to port max_ent to find_constrained_prior?
Can you share example code of how this would look like in preliz?

aloctavodia:
If you want to rewrite `find_constrained_prior`  to behave like maxent, then entropy needs to be added to distributions and then I would recommend to also use an initialization routine like the one used in maxent so users don't have to guess initial values themselves. 

This will compute the maxentropy Gamma distribution that has 90% of the mass between 1 and 10, and it will return it as a PyMC distribution, that you can then use as usual.  
 
```python
pz.maxent(pz.Gamma(), 1, 10, 0.9).to_pymc()
```
 
 You can also do other stuff like fix some of the parameters (mu in this example)
 
 ```python
pz.maxent(pz.Gamma(mu=4), 1, 10, 0.9)
```

Or stats like median or mode, 

 ```python
pz.maxent(pz.Beta(), 0.1, 0.7, 0.94, fixed_stat=("mode", 0.3))
 ```
 
 More examples here. https://preliz.readthedocs.io/en/latest/examples/gallery/direct_elicitation_1D.html

twiecki:
Would it be too hacky to replace `find_constrained_prior` with a call to preliz? I assume since you have code that translates preliz distribution to pymc you could have a function to tranlsate a pymc distribution into preliz.

aloctavodia:
We did that with arviz for a long time, so as hacky as that.
There is no `from_pymc` method, but I guess it should not be that hard.

PR #7610, opened 2024-12-11
Use jaxified logp for initial point evaluation when sampling via Jax

nataziel:
Use jaxified logp for initial point evaluation when sampling via Jax

## Description
- get jaxified logp function in `sample_jax_nuts`
  - uses different parameters to get jaxified function depending on which `nuts_sampler` is specified
- pass jaxified logp function to `_get_batched_jittered_initial_points`
  - added `logp_fn` parameter to function signature
  - wrap passed function to conform to how `_init_jitter` will call it
- pass wrapped function to `_init_jitter`
  - added `logp_fn` parameter to function signature
- added logic in `_init_jitter` to decide which function to use to evaluate the generated points
- added a bunch of type annotations

## Related Issue
- [x] Closes #7608

## Checklist
- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [x] Included tests that prove the fix is effective or that the new feature works
- [x] Added necessary documentation (docstrings and/or example notebooks) - have a question about the `sample_blackjax_nuts` function docstring, will put in comment below
- [x] If you are a pro: each commit corresponds to a [relevant logical change]

## Type of change
- [x] New feature / enhancement
- [x] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7610.org.readthedocs.build/en/7610/

PR #7681, opened 2025-02-14
Reuse jaxified logp when sampling via jax

nataziel:
reuse jaxified logp when sampling via jax

## Description

#7610 added logic to handle passing a pre-jaxified logp function into the blackjax/numpyro samplers, but missed actually passing the jaxified logp that is computed in `sample_jax_nuts`

## Checklist
- [x] If you are a pro: each commit corresponds to a [relevant logical change]

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7681.org.readthedocs.build/en/7681/

### DIFF
--- a/pymc/sampling/jax.py
+++ b/pymc/sampling/jax.py
@@ -669,6 +669,7 @@ def sample_jax_nuts(
         random_seed=random_seed,
         initial_points=initial_points,
         nuts_kwargs=nuts_kwargs,
+        logp_fn=logp_fn,
     )
     tic2 = datetime.now()
 


nataziel:
@ricardoV94 not sure if you've seen this, but it's a super tiny change that we should have included with #7610 that I just missed

twiecki:
Thanks for this PR! I see this is a small but important maintenance fix that adds the  parameter to the sampler function call, which was computed earlier but not passed through. This ensures we reuse the already jaxified logp function when sampling via JAX instead of recomputing it unnecessarily.\n\nThis change looks good and should improve efficiency by avoiding duplicate computation of the logp function. Appears to be a small oversight in PR #7610 that this properly fixes.

ricardoV94:
> Claude Code: Thanks for this PR! I see this is a small but important maintenance fix that adds the parameter to the sampler function call, which was computed earlier but not passed through. This ensures we reuse the already jaxified logp function when sampling via JAX instead of recomputing it unnecessarily.\n\nThis change looks good and should improve efficiency by avoiding duplicate computation of the logp function. Appears to be a small oversight in PR #7610 that this properly fixes.

Useless

ricardoV94 (approved):

ricardoV94:
Thanks @nataziel

PR #7684, opened 2025-02-18
Increase tolerance in TestGARCH11::test_batched_size to reduce flakiness

### LINKED ISSUE #7674: Flaky tests/distributions/test_timeseries.py::TestGARCH11::test_batched_size
### Description

Failed last time in https://github.com/pymc-devs/pymc/actions/runs/13036553157/job/36368381129?pr=7664#step:5:646 but I've seen it multiple times before.

We should fix it, either increase tolerance or make smaller sequences

JigyasuRajput:
## Description

Increased the absolute (`atol=1e-3`) and relative (`rtol=1e-2`) tolerances in `np.testing.assert_allclose` within `TestGARCH11::test_batched_size`.  
This should help reduce test flakiness caused by minor numerical variations.

## Related Issue

- [x] Closes #7674

## Checklist

- [x] Included tests that prove the fix is effective or that the new feature works (ran `pytest tests/distributions/test_timeseries.py`)

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7684.org.readthedocs.build/en/7684/

### DIFF
--- a/tests/distributions/test_timeseries.py
+++ b/tests/distributions/test_timeseries.py
@@ -818,6 +818,8 @@ class TestGARCH11:
         np.testing.assert_allclose(
             t0.compile_logp()(t0.initial_point()),
             t1.compile_logp()(t1.initial_point()),
+            atol=1e-3,
+            rtol=1e-2,
         )
 
     @pytest.mark.parametrize(


ricardoV94:
Hmm looks like that test should be deterministic, so the failures may indicate an underlying bug. Can you check if the results come different in different runs of the test?

nataziel:
The test isn't deterministic because of the `random.randn` on L788 I think

ricardoV94:
Oh that would explain it. Can you replace it by a seedes np.random.default_rng() ?

> The test isn't deterministic because of the `random.randn` on L788 I think

That would definitely not be deterministic. We can replace with a seeded generator.

Also not sure whether the initial points is deterministic or samples from the prior for this RV. That would also explain errors

PR #7685, opened 2025-02-19
Allow draws from `Weibull`, `MvStudentT`, `LKJCorr` and `LKJCholeskyCovRV` in alternative backends 

ricardoV94:
This PR converts more RandomVariable Ops to SymbolicRandomVariable (built on top of other RV Ops), which makes them automatically compatible with alternative backends.

Depends on: 
https://github.com/pymc-devs/pytensor/pull/1222
https://github.com/pymc-devs/pytensor/pull/1223

----
📚 Documentation preview 📚: https://pymc--7685.org.readthedocs.build/en/7685/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -302,7 +302,19 @@ class MvNormal(Continuous):
         )
 
 
-class PrecisionMvNormalRV(SymbolicRandomVariable):
+class SymbolicMVNormalUsedInternally(SymbolicRandomVariable):
+    """Helper subclass that handles the forwarding / caching of method to `MvNormal` used internally."""
+
+    def __init__(self, *args, method: str, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.method = method
+
+    def rebuild_rv(self, *args, **kwargs):
+        # rv_op is a classmethod, so it doesn't have access to the instance method
+        return self.rv_op(*args, method=self.method, **kwargs)
+
+
+class PrecisionMvNormalRV(SymbolicMVNormalUsedInternally):
     r"""A specialized multivariate normal random variable defined in terms of precision.
 
     This class is introduced during specialization logprob rewrites, and not meant to be used directly.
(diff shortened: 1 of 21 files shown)


ricardoV94:
I added an option to forward the decomposition method to the underlying MvNormal. This will require the next PyTensor release

twiecki (approved):
Skimmed.

pymc/distributions/multivariate.py
```diff
@@ -1525,55 +1515,66 @@ def make_node(self, rng, size, n, eta):
...
+            next_rng, y = pt.random.beta(
+                alpha=mp1 / 2.0, beta=beta, size=flat_size, rng=next_rng
+            ).owner.outputs
+            next_rng, z = pt.random.normal(
+                loc=0, scale=1, size=(flat_size, mp1), rng=next_rng
+            ).owner.outputs
+            z = z / pt.sqrt(pt.einsum("ij,ij->i", z, z.copy()))[..., np.newaxis]
+            P = P[..., 0:mp1, mp1].set(pt.sqrt(y[..., np.newaxis]) * z)
+            P = P[..., mp1, mp1].set(pt.sqrt(1.0 - y))
+        C = pt.einsum("...ji,...jk->...ik", P, P.copy())
```
jessegrabowski, line 1576: Would `P.mT @ P` just work here? We know `P` will always be at least 2d.

tests/distributions/test_random_alternative_backends.py
```diff
@@ -0,0 +1,70 @@
...
+import numpy as np
+import pytest
+
+import pymc as pm
+
+from pymc import DirichletMultinomial, MvStudentT
+from pymc.model.transform.optimization import freeze_dims_and_data
+
+
+@pytest.fixture(params=["FAST_RUN", "JAX", "NUMBA"])
```
jessegrabowski, line 25: unrelated to this PR, but do tests pass in torch mode?

tests/sampling/test_mcmc_external.py
```diff
@@ -81,7 +81,8 @@ def test_step_args():
             nuts_sampler="numpyro",
             target_accept=0.5,
             nuts={"max_treedepth": 10},
-            random_seed=1410,
+            random_seed=1411,
```
jessegrabowski, line 84: What's the story with the seeds? Is the test fragile on purpose?

jessegrabowski (approved):

tests/sampling/test_mcmc_external.py
ricardoV94: It started failing on the CI, perhaps numpyro changed something in the adaptation and doesn't try to respect target accept as much. It passed locally for me no matter what versions I tried to install. It's definitely unrelated to this PR so I'm not worried

tests/distributions/test_random_alternative_backends.py
ricardoV94: We don't have RVs in the torch backend yet

pymc/distributions/multivariate.py
ricardoV94: Yes

ricardoV94: Einsum should arrive at the same thing though

jessegrabowski: Sure, and I love einsum more than the average person, and actually find it more readable. Just curious what you thought about the alternative

ricardoV94: The .mt @ is a bit more direct (avoids the OFG that is inlined later), this makes it easier to review the changes :)

PR #7687, opened 2025-02-21
Allow for pymc native samplers to resume sampling from `ZarrTrace`

### LINKED ISSUE #7503: ENH: Add checkpoints during sampling

lucianopaz:
## Description
Big PR approaching! This finishes adding the ability of pymc native step methods to resume sampling from an existing trace (as long as it's a `ZarrTrace`!). This means that you can now continue tuning or sampling from a pre-existing sample run. For example

```python
with model:
    # First tuning run
    pm.sample(tune=400, draws=0, trace=trace)

    # Do whatever to decide if you want to continue tuning   
    pm.sample(tune=800, draws=0, trace=trace)

    # Switch to sampling
    pm.sample(tune=800, draws=1000, trace=trace)
```

Another thing is that the `chunks_per_draw` from `ZarrTrace` along with its persistent [storage backends](https://zarr.readthedocs.io/en/v2.18.4/tutorial.html#storage-alternatives) (like `ZipStore` or `DirectoryStore`) makes the sampling store the results and final sampling state periodically, so in case of a crash during sampling, you can use the existing store to load the trace using `ZarrTrace.from_store` and then resume sampling from there.

The only thing that I haven't tested for yet is to add an `Op` that makes `pm.sample` crash to see if I can reload the partial results from the store and resume sampling. @ricardoV94 gave me some pointers to that, but I won't be working on this for the rest of the month and I thought it best to open a draft PR to kick off any discussion you have or collect feedback

## Related Issue

- [X] Closes #7503 
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7687.org.readthedocs.build/en/7687/


pymc/backends/__init__.py
```diff
@@ -132,15 +132,41 @@ def init_traces(
...
+            backend.init_trace(
+                chains=chains,
+                draws=expected_length - tune,
+                tune=tune,
+                step=step,
+                model=model,
+                vars=trace_vars,
+                test_point=initial_point,
+            )
+        except TraceAlreadyInitialized:
```
fonnesbeck, line 145: Maybe just `InitializedTrace`? Seems a little verbose!

ricardoV94: Sounds fine to me, it's an internal thing

pymc/sampling/mcmc.py
```diff
@@ -1160,13 +1157,24 @@ def _sample_many(
...
+            trace = traces[i]
+            if isinstance(trace, ZarrChain):
+                progress_manager.set_initial_state(*trace.completed_draws_and_divergences())
+                progress_manager._progress.update(
+                    progress_manager.tasks[i],
+                    draws=progress_manager.completed_draws
+                    if progress_manager.combined_progress
+                    else progress_manager.draws,
+                    divergences=progress_manager.divergences,
+                    refresh=True,
```
ricardoV94: I still don't like this abstraction leaking elsewhere, just provide a default to the Ndarray backend that makes it work for either method. In that case I suppose start everything at zero

pymc/sampling/mcmc.py
```diff
@@ -1285,14 +1293,33 @@ def _iter_sample(
     step.set_rng(rng)
 
     point = start
+    initial_draw_idx = 0
+    step.tune = bool(tune)
+    if hasattr(step, "reset_tuning"):
+        step.reset_tuning()
     if isinstance(trace, ZarrChain):
         trace.link_stepper(step)
+        stored_draw_idx = trace._sampling_state.draw_idx[chain]
```
ricardoV94: Same here all this logic including the old `link_stepper` can have a sensible default in the base trace class so you don't need to worry about what kind of trace you have here. Just make `link_stepper` a no op and `stored_draw_idx` to be zero by default?

pymc/sampling/parallel.py
```diff
@@ -194,6 +194,24 @@ def _start_loop(self):
...
+                if stored_sampling_state is not None:
+                    self._step_method.sampling_state = stored_sampling_state
+                else:
+                    raise RuntimeError(
+                        "Cannot use the supplied ZarrTrace to restart sampling because "
+                        "it has no sampling_state information stored. You will have to "
+                        "resample from scratch."
+                    )
+                draw = stored_draw_idx
+                self._write_point(trace.get_mcmc_point())
```
ricardoV94, line 211: Duplicated logic, should be a property of the backend object?

pymc/sampling/parallel.py
```diff
@@ -491,6 +509,10 @@ def __init__(
             progressbar=progressbar,
             progressbar_theme=progressbar_theme,
         )
+        if self.zarr_recording:
```
ricardoV94: abstraction leaking

ricardoV94 (commented):
I like the new functionality, I am deeply against all the `if isinstance(..., ZarrTrace)` in the codebase. Either our code is supposed to allow different trace backends or it is not, this suggests you want to drop the Ndarray altogether, which fine if you do.

Otherwise all these cases seem like they could be handled by the BaseTrace having sensible default for these methods. We used to have continuation of traces in the past with Ndarray, I don't see anything that fundamentally needs ZarrTrace other than dev interest in it? So just make it raise `NotImplementedErrors` or make them no-ops and adjust the external code appropriately

I stopped half-way so it was not an extensive review. I think this is a bigger design point that needs decision before settling on the details of the PR.

pymc/sampling/mcmc.py
lucianopaz: I tried to improve the abstraction to prevent most of the custom `ZarrChain` checks

pymc/backends/mcbackend.py
```diff
@@ -201,6 +201,42 @@ def _slice(self, idx: slice) -> "IBaseTrace":
     def point(self, idx: int) -> dict[str, np.ndarray]:
         return self._chain.get_draws_at(idx, [v.name for v in self._chain.variables.values()])
 
+    def completed_draws_and_divergences(self, chain_specific: bool = True) -> tuple[int, int]:
```
ricardoV94, line 204: question why do we have special handling for divergences, instead of being sampler agnostic?

pymc/sampling/mcmc.py
```diff
@@ -1403,7 +1423,9 @@ def _mp_sample(
         progressbar_theme=progressbar_theme,
         blas_cores=blas_cores,
         mp_ctx=mp_ctx,
-        zarr_chains=zarr_chains,
+        # We only need to pass the traces when zarr_recording is happening because
+        # it's the only backend that can resume sampling
+        traces=traces if zarr_recording else None,
```
ricardoV94, line 1423: Someone could implement a different trace outside of PyMC so why assume this?

pymc/sampling/parallel.py
```diff
@@ -194,6 +195,24 @@ def _start_loop(self):
 
         draw = 0
         tuning = True
+        if self._zarr_recording:
```
ricardoV94, line 197: This should be a method of the trace, still zarr specific

pymc/sampling/parallel.py
```diff
@@ -271,15 +291,15 @@ def __init__(
...
 
-        zarr_chains_send: list[ZarrChain] | bytes | None = None
-        if zarr_chains_pickled is not None:
-            zarr_chains_send = zarr_chains_pickled
-        elif zarr_chains is not None:
+        traces_send: Sequence[IBaseTrace] | bytes | None = None
+        if traces_pickled is not None:
+            traces_send = traces_pickled
+        elif traces is not None:
             if mp_ctx.get_start_method() == "spawn":
```
ricardoV94, line 297: Note this will be the default soon in linux as well (fork is being deprecated). In case we are not automating something we should be automating

pymc/step_methods/state.py
```diff
@@ -26,9 +30,78 @@
 class DataClassState:
     __dataclass_fields__: ClassVar[dict[str, Field[Any]]] = {}
 
+    def is_compatible(self, other: Any) -> bool:
+        return compatible_dataclass_values(self, other)
+
+
+def resolve_typehint(hint: Any, anchor: object = None) -> Any:
```
ricardoV94, line 37: These new functions seems like a big code smell. Can we approach it differently?

tests/step_methods/test_state.py
```diff
@@ -130,6 +131,11 @@ def test_sampling_state():
     with pytest.raises(AssertionError, match="Encountered invalid state class"):
         b1.sampling_state = b1_state.state1
 
+    with pytest.raises(
+        AssertionError, match="The supplied state is incompatible with the current sampling state."
+    ):
+        b1.sampling_state = b5.sampling_state
```
ricardoV94, line 137: Nit, I still dislike magic property. A method call is much more obvious that some checks may be done, and you're not just doing a random assignment. It also gives you a hatch to disable them if you're confident it's correct and want to avoid costly checks, because you're allowed to have kwargs then

ricardoV94: And you can't refactor properties into methods with back-compat. We had stuff with `model.logp` and it was a pita for saving two parenthesis back in the day

ricardoV94 (commented):
Changes look okay, I still have some reservations on code modularity and complexity. I don't like the extra work we're doing to validate states are compatible. If it's that hard just don't do it.

Some places still treat Zarr specially without much reason for it. 

I'm okay with approving this next iteration / if you disagree, so this doesn't hang for ever

requirements.txt
```diff
@@ -7,4 +7,5 @@ pytensor>=2.30.2,<2.31
 rich>=13.7.1
 scipy>=1.4.1
 threadpoolctl>=3.1.0,<4.0.0
+typeguard>=4.4.2,<5.0.0
```
williambdean, line 10: needs to be added to conda-envs as well

schlich:
Hello friends, I will be working on closing out this issue/PR today until it is done.  I have questions for y'alls preferred review workflow:

1.  Should I open a PR against this branch?  Or a separate one against main?  Or other?
1.  I'm going to start with refactoring these tests to be a little bit more modular and structured.  If i push these refactorings to the PR branch as I go, would  you like to review them as they are pushed?  Or would you like me to save all pushes until i'm done with the refactor and/or rework?f
1.   Assuming it's in scope for this PR, i'd like to complete the zarr v3 upgrade as it is completely blocking me from this feature for my setup.  I plan to also incorporate appropriate error handling etc to accommodate v2, which should not be a problem since there are clear corollaries between the Store objects in each version for our purposes.

In the meantime I can push my changes to my own fork regardless, will share a link after first commit.  Feel free to examine if needed.

Thanks in advance!

lucianopaz:
Thanks @schlich! I think that maybe I can rebase to fix the conflicts, merge the PR and you can take it from there with a fresh new PR on top of main. I don't think that there was much left to do to get this merged and I wont be able to make significant improvements before the end of August. zarr v3 support would be a great addition. Be sure to mention #7752 when you open your PR

schlich:
@lucianopaz thanks for getting that patched up!  I have it built from git in my local project, haven't tested it closely but so far so good 👍

avm19:
#8104 seems to be addressed by this contribution, which is another reason to prioritise and expedite this PR.

@lucianopaz I tried your branch and it appears to solve my problem. Just to double check, you are not calling `Zarr.Array.__getitem()__` anywhere any more, right? 
Also, I haven't look at the new code closely, but you are working with Zarr via xarray, right? Is this the recommended way? This one I am asking for myself. Thanks!

lucianopaz:
> @lucianopaz I tried your branch and it appears to solve my problem. Just to double check, you are not calling `Zarr.Array.__getitem()__` anywhere any more, right?

I don't remember, I'll have to check and get back to you.

> Also, I haven't look at the new code closely, but you are working with Zarr via xarray, right? Is this the recommended way? This one I am asking for myself. Thanks!

No, this works with zarr directly. It only adds the extra xarray specific attributes to zarr groups and arrays to code the dimension names.

pymc/backends/zarr.py
```diff
@@ -286,10 +379,9 @@ class ZarrTrace:
     | |--> _sampling_state
 
     The root group is created when the ``ZarrTrace`` object is initialized. The rest of
-    the groups are created once :meth:`~ZarrChain.init_trace` is called with a few exceptions:
-    unconstrained_posterior is only created if ``include_transformed = True``, and the
-    groups prefixed with ``warmup_`` are created only after calling
-    :meth:`~ZarrTrace.split_warmup_groups`.
+    the groups are created once :meth:`~ZarrChain.init_trace` is called with the exception
```
avm19, line 382: Probably ZarrTrace.init_trace(), not ZarrChain.init_trace()

pymc/backends/zarr.py
```diff
@@ -411,16 +549,17 @@ def init_trace(
...
         Every array in the posterior and sample stats groups will have the
-        (chains, tune + draws) batch dimensions to the left of the core dimensions of
+        (chains, draws) batch dimensions to the left of the core dimensions of
         the model's random variable or the step method's stat shape. The warmup (tuning
-        draws) and the posterior samples are split at a later stage, once
-        :meth:`~ZarrTrace.split_warmup_groups` is called.
+        draws) posterior and sample stats will have (chains, tune) batch dimensions
+        instead.
 
         After the creation if the zarr hierarchies, it initializes the list of
```
avm19, line 564: typo: if -> of

pymc/backends/zarr.py
avm19, line 585: ~~test_point may be not used in ZarrChain. But you are definitely using it here, in ZarrTrace.init_trace, so the docstring should say this.~~
Actually, I think test_point argument is used in both ZarrChain and ZarrTrace. So the docstrings should be corrected in both places. As far as I could see, ZarrTrace assigns a value to test_point if not provided, and ZarrTrace passes it via super() to BaseTrace, which uses it to get shapes and dtypes for input and output variables. In the ideal world, ZarrChain and ZarrTrace should simply repeat the test_point's docstring entry from BaseTrace. But BaseTrace docstring is not very informative, so why not come up with something better.

pymc/backends/zarr.py
```diff
@@ -101,50 +109,58 @@ def __init__(
...
         self.chain = chain
-        self.total_draws = draws
-        self.draws_until_flush = min([self.draws_per_chunk, draws - self.draw_idx])
+        self.draws = draws
+        self.tune = tune
+        self.draw_idx = self._sampling_state.draw_idx[chain] or 0
+        self.update_draws_until_flush()
         self.clear_buffers()
 
     def clear_buffers(self):
```
avm19, line 166: It looks like ZarrChain methods `clear_buffers`, `buffer`, `flush`, `update_draws_until_flush` should be private.

On a more general note...
To make the code more understandable and maintainable by other than the original author, I would suggest to consider the following practice (not sure what standards PyMC devs follow, but this sounds like common sense to me):
- private methods must be explicitly marked as such (i.e. start with a `_`)
- private methods may or may not have a proper docstring, but it is encouraged where appropriate
- public methods _must_ have a proper docstring (which could be inherited from a parent class. Then should a comment be added in the subclass method implementation that it is being overridden/extended?)
- a method should probably be private if its signature or semantics is likely to change, or if its use is almost certainly internal
- a method should be public if a user is expected to call it in reasonable workflows

It was difficult for me to read current stable PyMC docs as well as this PR, because I couldn't tell a significant and reliable abstraction from a volatile and discretionary implementation detail. A clear delineation of private vs public would have helped me. Speaking of docstrings, since ZarrChain is a backend implementation of BaseTrace, maybe improving BaseTrace and IBaseTrace docstrings is a good idea.

pymc/backends/zarr.py
avm19, line 88: `model : Model | None`

avm19, line 94: test_point *is* used. See a comment to test_point in ZarrTrace docstring.

pymc/backends/zarr.py
```diff
@@ -101,50 +109,58 @@ def __init__(
         test_point: dict[str, np.ndarray] | None = None,
         draws_per_chunk: int = 1,
```
avm19, line 110: The default value is missing from the docstring.

pymc/backends/zarr.py
```diff
@@ -860,3 +1093,195 @@ def to_inferencedata(self, save_warmup: bool = False) -> az.InferenceData:
...
+        zarr_observed_data = (
+            [name for name in self.observed_data.array_keys() if name not in zarr_coords]
+            if "observed_data" in zarr_groups
+            else []
+        )
+        autogenerated_dims = {dim for dim in zarr_coords if re.search(r"_dim_\d+$", dim)}
+
+        # Check deterministics, free RVs and transformed RVs
+        _, var_names = self.parse_varnames(model, vars)
+        assert set(var_names) == set(zarr_free_vars + zarr_deterministics), (
```
avm19, line 1219: Kind of nitpicking, but consider replacing `assert`s with explicit `if ...: raise ...`. Asserts can be disabled and ignored in some production environments, and you can also be more expressive in choosing an appropriate Exception subclass (or defaulting to ValueError).

pymc/backends/zarr.py
avm19, line 405: Add a note with something along these lines:
`draws_per_chunk` determines chunks in the underlying Zarr Arrays and may have a huge impact on performance. As the optimal value depends on many factors (e.g. model computational complexity, sampler, type of storage, core dimensions of sampled variables), the choice is left to the user.

avm19:
> > @lucianopaz I tried your branch and it appears to solve my problem. Just to double check, you are not calling `Zarr.Array.__getitem()__` anywhere any more, right?
> 
> I don't remember, I'll have to check and get back to you.

@lucianopaz I traced execution for a toy example. Get methods of `Zarr.Array` are called mostly on arrays in `self._sampling_state` group and such. Large data arrays are only written to, when pm.sample() is called with
```python
return_inferencedata=False, discard_tuned_samples=False, compute_convergence_checks=False
```

So https://github.com/pymc-devs/pymc/issues/8104 is indeed resolved by this.

avm19 (commented):
I am currently using [lucianopaz:zarr_continue](https://github.com/lucianopaz/pymc/tree/zarr_continue) branch in my work, and here are the thoughts I had while reading and tracing the code. The greatest difficulty I've had so far is understanding the relation between `ZarrTrace`, `ZarrChain`, `BaseTrace`, `MultiTrace`, `Backend` and other classes. Clarifications about the separation of labour between `ZarrTrace` and `ZarrChain` helped a lot, but the bigger picture (how they interface with `pm.sample`) is still somewhat opaque, as I mostly looked into `zarr.py`. Note that I am not a PyMC developer. I hope my comments will help you merge this sooner.

pymc/backends/zarr.py
```diff
@@ -286,10 +379,9 @@ class ZarrTrace:
     | |--> _sampling_state
 
     The root group is created when the ``ZarrTrace`` object is initialized. The rest of
```
avm19, line 381: ... overwriting any existing data.

pymc/backends/zarr.py
```diff
@@ -860,3 +1093,195 @@ def to_inferencedata(self, save_warmup: bool = False) -> az.InferenceData:
...
+        self.compressor = default_compressor
+
+        groups = set(self.root.group_keys())
+        assert groups >= {
+            "posterior",
+            "sample_stats",
+            "warmup_posterior",
+            "warmup_sample_stats",
+            "constant_data",
+            "observed_data",
```
avm19, line 1121: init_trace() docstring says:
> ``observed_data`` (only if the model has observed),

So the edge case with no observed data fails here. Same for constant_data.

P.S. Is there a good reason for not creating empty groups? I don't think saving a dozen of I/O is worth the trouble.

avm19:
~~Direct use of `ZarrTrace` results in the creation of (1,1)-chunked arrays, where each file stores a single value. Any ideas or plans on how to improve this?~~

~~Any suggestions as to mitigation while this is being developed? The storage overhead can be huge. If stored as `Zarr.DirectoryStore`, then the efficiency is mere 5% (expected over 100%), not taking into account file system cluster waste.~~ Update. Parameter `draws_per_chunk` should help, still might be a huge disparity in data sizes between `posterior` and `sample_stats`. Inspecting micro-chunked `sample_stats` is much slower than it should be. To improve this in my application, I manually re-chunk arrays in `[warmup_]sample_stats` group after sampling is completed.


I also tried `Zarr.ZipStore`, but it raises errors when `cores` > 1 and "Duplicate name" warnings when `pm.sample(..., cores=1)`.

P.S. For `sample_stats/sampler_0__energy` float64 there is 33% efficiency byte-wise, which becomes ~2% for a 4KB-block file system. On average, each sample creates 20 files and wastes 80KB in disk space via `sample_stats`.

<details>

<summary> Example:  </summary>

 `trace['sample_stats']['sampler_0__warning'].info`: 

Name | /sample_stats/sampler_0__warning
-- | --
zarr.core.Array
object
(2, 6000)
(1, 1)
C
False
Pickle(protocol=5)
Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
zarr.storage.DirectoryStore
96000 (93.8K)
1980513 (1.9M)
0.0
12000/12000
</details>

PR #7695, opened 2025-02-26
Fix bug when reusing jax logp for initial point generation

ricardoV94:
Fixes issue described in https://discourse.pymc.io/t/attributeerror-numpy-ndarray-object-has-no-attribute-at-when-sampling-lkj-cholesky-covariance-priors-for-multivariate-normal-models-example-with-numpyro-or-blackjax/16598/3

Caused by #7681 
CC @nataziel 

----
📚 Documentation preview 📚: https://pymc--7695.org.readthedocs.build/en/7695/


jessegrabowski (approved):

nataziel:
Just for anyone investigating this later I think this was directly caused by #7610, not #7681

PR #7701, opened 2025-03-01
Added type hints to discrete distributions in pymc/distributions/discrete.pyAdd type hints to discrete distributions

deveshbervar:
----
📚 Documentation preview 📚: https://pymc--7701.org.readthedocs.build/en/7701/

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -13,6 +13,8 @@
 #   limitations under the License.
 import warnings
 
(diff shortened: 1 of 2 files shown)


OriolAbril (changes requested):
I don't know what the `DIST_PARAMETER_TYPES` is supposed to be nor I can find it defined anywhere. Would you care to elaborate on the goal of the PR? Otherwise we'll have to close it

PR #7707, opened 2025-03-03
Expand logging test cases for sample_prior_predictive and add return type overloads

### LINKED ISSUE #7703: BUG: `pm.sample_prior_predictive` reports it will sample all volatile basic RVs even when `var_names` is supplied
### Describe the issue:

When you pass a list of variable names to `pm.sample_prior_predictive` a message is logged that suggests a much larger list of variables will be sampled than what you might provide.

The function appears to correctly constrain the list of output variables passed to `compile_forward_sampling_function`, and the returned data seems to correctly only contain the variables you request, but the logged message suggests it will sample all volatile basic RVs.

Seems to be caused by this line:
https://github.com/pymc-devs/pymc/blob/d1aff0b766eb1cc217c1c15a8c3bab52f0517fdc/pymc/sampling/forward.py#L452

```python
import pymc as pm
import numpy as np


def main():
    data = [5.1, 5.2, 4.9, 4.8]
    with pm.Model() as model:
        target_value = pm.Data(name="target_y", value=data, dims=("x"))
        a = pm.Normal("a")

        y_hat = pm.Deterministic("y_hat", var=a + 5)

        y_like = pm.Normal("y_like", mu=y_hat, observed=target_value)

        my_model = model

    prior_trace = pm.sample_prior_predictive(model=my_model, var_names=["y_like"])

    print(prior_trace)


if __name__ == "__main__":
    main()
```

### Error message:

```shell
Sampling: [a, y_like]
Inference data with groups:
        > prior_predictive
        > observed_data
        > constant_data
```

### PyMC version information:

Currently running PyMC 5.21

### Context for the issue:

low priority, but just wanted to make it known (or be told that my understanding is incorrect)

nataziel:
expand logging test cases for sample_prior_predictive & add return type overloads

## Description

Expanded logging test cases to clarify expected behaviour/purpose of logged message in `sample_prior_predictive`

Added return type overloads while I'm in there for useability (and to remove a bunch of instances of `assert isinstance(trace, az.InferenceData)` from my own code :joy:)

## Related Issue
- [x] Closes #7703 

## Checklist
- [x] If you are a pro: each commit corresponds to a [relevant logical change]

## Type of change
- [x] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7707.org.readthedocs.build/en/7707/

### DIFF
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -473,9 +473,7 @@ def sample_prior_predictive(
     )
 
     # All model variables have a name, but mypy does not know this
-    _log.info(
-        f"Sampling: {sorted(volatile_basic_rvs.intersection(vars_to_sample), key=lambda var: var.name)}"  # type: ignore[arg-type, return-value]
-    )
+    _log.info(f"Sampling: {sorted(volatile_basic_rvs, key=lambda var: var.name)}")  # type: ignore[arg-type, return-value]
     values = zip(*(sampler_fn() for i in range(draws)))
 
     data = {k: np.stack(v) for k, v in zip(names, values)}
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -22,6 +22,7 @@ from typing import (
     Any,
     TypeAlias,
     cast,
+    overload,
 )
 
 import numpy as np
@@ -360,6 +361,28 @@ def observed_dependent_deterministics(model: Model, extra_observeds=None):
     ]
 
 
+@overload
+def sample_prior_predictive(
+    draws: int = 500,
+    model: Model | None = None,
+    var_names: Iterable[str] | None = None,
+    random_seed: RandomState = None,
+    return_inferencedata: bool = True,
+    idata_kwargs: dict | None = None,
+    compile_kwargs: dict | None = None,
+    samples: int | None = None,
+) -> InferenceData: ...
+@overload
+def sample_prior_predictive(
+    draws: int = 500,
+    model: Model | None = None,
+    var_names: Iterable[str] | None = None,
+    random_seed: RandomState = None,
+    return_inferencedata: bool = False,
+    idata_kwargs: dict | None = None,
+    compile_kwargs: dict | None = None,
+    samples: int | None = None,
+) -> dict[str, np.ndarray]: ...
 def sample_prior_predictive(
     draws: int = 500,
     model: Model | None = None,
@@ -449,7 +472,7 @@ def sample_prior_predictive(
     )
 
     # All model variables have a name, but mypy does not know this
-    _log.info(f"Sampling: {sorted(volatile_basic_rvs, key=lambda var: var.name)}")  # type: ignore[arg-type, return-value]
+    _log.info(f"Sampling: {sorted(vars_to_sample, key=lambda var: var.name)}")  # type: ignore[arg-type, return-value]
     values = zip(*(sampler_fn() for i in range(draws)))
 
     data = {k: np.stack(v) for k, v in zip(names, values)}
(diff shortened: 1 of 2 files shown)


nataziel:
Happy to split the overload addition to a separate PR if it's preferrable

pymc/sampling/forward.py
```diff
@@ -449,7 +473,9 @@ def sample_prior_predictive(
     )
 
     # All model variables have a name, but mypy does not know this
-    _log.info(f"Sampling: {sorted(volatile_basic_rvs, key=lambda var: var.name)}")  # type: ignore[arg-type, return-value]
+    _log.info(
+        f"Sampling: {sorted(volatile_basic_rvs.intersection(vars_to_sample), key=lambda var: var.name)}"  # type: ignore[arg-type, return-value]
```
ricardoV94: This is changing too much, we want to show RVs that must be sampled in order to get the ones the user requested, this won't show them?

https://github.com/pymc-devs/pymc/issues/5973

Are you trying to just remove the deterministics?

tests/sampling/test_forward.py
```diff
@@ -862,11 +862,18 @@ def test_logging_sampled_basic_rvs_prior(self, caplog):
...
+        # RV only
         with m:
             pm.sample_prior_predictive(draws=1, var_names=["x"])
         assert caplog.record_tuples == [("pymc.sampling.forward", logging.INFO, "Sampling: [x]")]
         caplog.clear()
 
+        # observed only
+        with m:
+            pm.sample_prior_predictive(draws=1, var_names=["z"])
+        assert caplog.record_tuples == [("pymc.sampling.forward", logging.INFO, "Sampling: [z]")]
```
ricardoV94: We want this to show [x, z], because x is being sampled as well in order to get z

ricardoV94: The new test is actually useful to cement what the behavior should be

tests/sampling/test_forward.py
```diff
@@ -862,11 +862,18 @@ def test_logging_sampled_basic_rvs_prior(self, caplog):
         assert caplog.record_tuples == [("pymc.sampling.forward", logging.INFO, "Sampling: [x, z]")]
         caplog.clear()
 
+        # RV only
         with m:
             pm.sample_prior_predictive(draws=1, var_names=["x"])
         assert caplog.record_tuples == [("pymc.sampling.forward", logging.INFO, "Sampling: [x]")]
         caplog.clear()
 
+        # observed only
```
ricardoV94: Nitpick, the distinction is not RV vs observed, is free_RV vs observed_RV, but either way I don't think the comment helps.

pymc/sampling/forward.py
ricardoV94: I guessed deterministics from the commit messages, but it sounds you are just trying to remove RVs that don't show up in the trace, that's not what we want

nataziel: If the desired behaviour is that the logged message lists all the random variables that will be sampled in order for the output trace to contain the requested vars from `var_names` then my changes are unnecessary and I can just convert this PR to implement the type return overload stuff (or make a new branch/PR).

Might just be a misunderstanding from my end, please clarify 😅

ricardoV94: Yes that's the desired behavior. We can leave your new test to make it more clear for devs in the future. Just remove the comments I nitpicked about and/or add a more informative comment about why the new test excepts the two RVs in the log

ricardoV94 (approved):

ricardoV94:
Thanks @nataziel

PR #7712, opened 2025-03-05
Do not monkey-patch Ipython pretty representation on model variables

ricardoV94:
Only model objects will have a default Ipython/Jupyter representation.

This simplifies the codebase and saves half a second of import time.

Before:
```python
import pymc as pm

with pm.Model() as m:
    x = pm.Normal("x")
    a = pm.Deterministic("a", x + 1)
    y = pm.Normal("y", a)

# Before this would print x ~ Nomal(0, 1)
# Now it just prints x (the name)
x
# x
```

```python
# Still prints the same as before
m
# x ~ Normal(0, 1)
# y ~ Normal(f(x), 1)
# a ~ Deterministic(f(x))
```

----
📚 Documentation preview 📚: https://pymc--7712.org.readthedocs.build/en/7712/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -16,7 +16,6 @@ from __future__ import annotations
 import functools
 import sys
 import threading
-import types
 import warnings
 
 from collections.abc import Iterable, Sequence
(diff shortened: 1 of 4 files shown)


pymc/model/core.py
```diff
@@ -2014,6 +2006,19 @@ def to_graphviz(
             dpi=dpi,
         )
 
+    def _repr_pretty_(self, p, cycle):
```
williambdean, line 2009: cycle isn't used. Is this some requirement from IPython? 

Also what are the types here?

ricardoV94: I think so, and I don't know

Issue #7713, opened 2025-03-09
Add `rng_fn` to CAR/ICAR


jessegrabowski:
### Description

I was talking to @theorashid who linked me to [this case study](https://mc-stan.org/learn-stan/case-studies/icar_stan.html) of CAR priors. It seems like they're just MvNormals, but with degenerate covariance matrices. We can now sample from such distributions use the new `method="eig"` or `method="svd"` argument. So a potential `rng_fn` would just make an appropriate MvNormal with the method argument set, then return it's rng_fn. 

Looking at the `logp` method for these distributions, it seems like it's just using the eig method; so we might be able to simplify these to wrappers around `MvNormalRV` that just constructs the mean/covariance and sets the appropriate method, but that's a step beyond what this PR is asking for.

ricardoV94:
Plus some special logic for sparse covariances that would be nice to support in MvNormal as well?

jessegrabowski:
A hidden internal `SparseMvNormal`, in the vein of `PrecisionMvNormal` would be nice. We could rewrite to it when we see the covariance is sparse.

Muhammad-Rebaal:
Hi @jessegrabowski !
Hope you are fine !
If no one is working on this issue , could you assign this to me I'd like to solve this.
Thank You !

ricardoV94:
We don't assign issues, you can just open a PR

Muhammad-Rebaal:
> We don't assign issues, you can just open a PR

Ok sure thing I'll open a draft one.

asifzubair:
Hi @jessegrabowski , @ricardoV94 , I started this PR ( #7879 ) to address this issue. Please note I've only implemented ICARRV ( and corresponding test ) for now. I was hoping to get your reviews and then the implementation for CARRV would be easier. Please let me know your thoughts. Thank you 🙏

ricardoV94:
Just because we can, should we? Are these draws valid in any meaningful sense? Are they useful?

jessegrabowski:
It would enable prior predictive checks for ICAR models, which would be nice.

ricardoV94:
But are they valid/meaningful? This is an improper prior IIUC

ricardoV94:
Is it something like that? https://stats.stackexchange.com/a/159322

And if valid, should we use svd specifically, instead of eigh?

jessegrabowski:
Yeah I might have missed some details.

[...]

PR #7723, opened 2025-03-15
Implemented `rng_fn` to CAR/ICAR

Muhammad-Rebaal:
Hi @jessegrabowski , @ricardoV94 !
I've implemented the `rng_fn` for the designated classes `CAR/ICAR` in the issue [#7713](https://github.com/pymc-devs/pymc/issues/7713) , so could you please review and let me know if there is anything needs to be changed . 
Thank You !

----
📚 Documentation preview 📚: https://pymc--7723.org.readthedocs.build/en/7723/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2156,56 +2156,56 @@ class CARRV(RandomVariable):
         return super().make_node(rng, size, mu, W, alpha, tau, W_is_valid)
 
     @classmethod
-    def rng_fn(cls, rng: np.random.RandomState, mu, W, alpha, tau, W_is_valid, size):
-        """Sample a numeric random variate.
-
-        Implementation of algorithm from paper
-        Havard Rue, 2001. "Fast sampling of Gaussian Markov random fields,"
-        Journal of the Royal Statistical Society Series B, Royal Statistical Society,
-        vol. 63(2), pages 325-338. DOI: 10.1111/1467-9868.00288.
+    def rng_fn(cls, rng, mu, W, alpha, tau, W_is_valid, size=None):
+        """Sample from the CAR distribution.
+        
+        Parameters
+        ----------
+        rng : numpy.random.Generator
+            Random number generator
+        mu : ndarray
+            Mean vector
+        W : ndarray
+            Symmetric adjacency matrix
+        alpha : float
+            Autoregression parameter (-1 < alpha < 1)
+        tau : float
+            Precision parameter (tau > 0)
+        W_is_valid : bool
+            Flag indicating whether W is a valid adjacency matrix
+        size : tuple, optional
+            Size of the samples to generate
+        
+        Returns
+        -------
+        ndarray
+            Samples from the CAR distribution
         """
-        if not W_is_valid.all():
-            raise ValueError("W must be a valid adjacency matrix")
+        if not W_is_valid:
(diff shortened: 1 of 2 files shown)


jessegrabowski (changes requested):
Instead of doing all the decomposition here, just pass the precision matrix and set `method = "eig"` to `multivariate_normal`. 

You will also need to add a test.

ricardoV94:
Unless there's some subtle complexity I would suggest to implement as a SymbolicRV like in: https://github.com/pymc-devs/pymc/pull/7685

Muhammad-Rebaal:
> Instead of doing all the decomposition here, just pass the precision matrix and set `method = "eig"` to `multivariate_normal`.
> 
> You will also need to add a test.

Yeah sure I'll do that

jessegrabowski:
Go into `tests/distributions/test_multivariate.py::TestICAR::test_icar_rng_fn` and update the test to no longer error. Check the other `test_rng_fn` tests for how to validate the output

Muhammad-Rebaal:
> Go into `tests/distributions/test_multivariate.py::TestICAR::test_icar_rng_fn` and update the test to no longer expect an error. Check the other `test_rng_fn` tests for how to validate the output.
> 
> It looks like you also have to run pre-commit

Thanks @jessegrabowski !

I've updated the rng_fn and added the test as per your request. Let me know if anything needs to change ?

Once again Thank You !

jessegrabowski:
Did you check the PR that Ricardo linked? This would be the preferred way to go.

Muhammad-Rebaal:
> Did you check the PR that Ricardo linked? This would be the preferred way to go.

You're referring that to write the `rng_fn` like `SymbolicRV` or something else?

Muhammad-Rebaal:
> You would convert the entire distribution into a `SymbolicMVNormalUsedInternally` subclass, like how `MvStudentTRV` was changed in the linked PR (check [here](https://github.com/pymc-devs/pymc/pull/7685/files#diff-ba7699f4e550fe6433c577626dd433a017c2fad177dcbfc83ea8268fe4322761R385) specifically). Then no rng_fn is required at all. The `rv_op` signature will take all the inputs (`mu, W, alpha, tau`), and the `extended_signature` will be `"[rng],[size],(n),(n,n),(),()->[rng],(n)"`, because `mu` is a vector, `W` is a matrix,`and`alpha`and`tau` are scalars, and it returns a vector.
> 
> Then do the same logic you already did in the `rng_fn` to compute `draws`, but symbolically in the `rv_op`. It will return:
> 
> ```python
>         return cls(
>             inputs=[rng, size, mu, W, alpha, tau],
>             outputs=[next_rng, draws],
>             method=method,
>         )(rng, size, nu, mean, scale)
> ```
> 
> The `method` argument should be `eig` by default.

Hey @jessegrabowski  !
Hope you are fine !
 
I've implemented that part but struggling with test part . Can you help me out with and tell me that what I'm missing ? Here's the link of the [code file & error](https://drive.google.com/file/d/1OfJTBagVWpR71nfWrxBQ1O_Q5igmFmZA/view?usp=sharing)

Thank You !

jessegrabowski:
You can push the code you have so far and we can look at it here

asifzubair:
Hi @Muhammad-Rebaal , are you still working on this ? If not, do you mind if I take a look at it ? Thank you 🙏

PR #7732, opened 2025-03-21
Fix domain specification in OrderedLogistic and OrderedProbit documentation

coderpheonix:
This PR corrects the domain specification in the documentation of the OrderedLogistic and OrderedProbit distributions. It ensures that the indexing is clearly mentioned as 0-based to avoid errors.

----
📚 Documentation preview 📚: https://pymc--7732.org.readthedocs.build/en/7732/

### DIFF
--- a/pymc/distributions/discrete.py
+++ b/pymc/distributions/discrete.py
@@ -1189,7 +1189,7 @@ class OrderedLogistic:
     R"""Ordered Logistic distribution.
 
     Useful for regression on ordinal data values whose values range
-    from 1 to K as a function of some predictor, :math:`\eta`. The
+    from 0 to K-1 as a function of some predictor, :math:`\eta`
     cutpoints, :math:`c`, separate which ranges of :math:`\eta` are
     mapped to which of the K observed dependent variables. The number
     of cutpoints is K - 1. It is recommended that the cutpoints are


pymc/distributions/discrete.py
```diff
@@ -1189,7 +1189,7 @@ class OrderedLogistic:
     R"""Ordered Logistic distribution.
 
     Useful for regression on ordinal data values whose values range
-    from 1 to K as a function of some predictor, :math:`\eta`. The
+    from 0 to K-1 as a function of some predictor, :math:`\eta`
```
OriolAbril, line 1192: It looks like you also removed the ". The" part which makes the [docstring](https://pymcio--7732.org.readthedocs.build/projects/docs/en/7732/api/distributions/generated/pymc.OrderedLogistic.html) confusing: 

imatge

Seems there are eta cutpoints with this updated wording which is not the case.

OriolAbril (changes requested):
Thanks! 

Would you mind also updating the parameters section?

* `cutpoints : tensor_like of array` -> `cutpoints : tensor_like of float`
* `compute_p: boolean, default True` needs a space before the colon as numpydoc requires spaces both before and after the colon that separates name and type of parameters

PR #7741, opened 2025-03-30
Allow passing graph level attributes to graphviz

zaxtax:
This add the ability to pass in graph level attributes to the `model_to_graphviz` function. This makes it easier to tweak the layout of the graph.

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7741.org.readthedocs.build/en/7741/

### DIFF
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -463,10 +463,7 @@ def make_graph(
     node_formatters = node_formatters or {}
     node_formatters = update_node_formatters(node_formatters)
 
-    graph = graphviz.Digraph(name)
-    if graph_attrs is not None:
-        graph.attr(**graph_attrs)
-
+    graph = graphviz.Digraph(name, graph_attr=graph_attrs)
     for plate in plates:
         if plate.dim_info:
             # must be preceded by 'cluster' to get a box around it
--- a/pymc/model_graph.py
+++ b/pymc/model_graph.py
@@ -430,9 +430,6 @@ class ModelGraph:
         ]
 
 
-GraphAttrMapping = dict[Any, Any]
-
-
 def make_graph(
     name: str,
     plates: list[Plate],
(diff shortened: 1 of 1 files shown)


pymc/model_graph.py
```diff
@@ -430,6 +430,9 @@ def edges(
         ]
 
 
+GraphAttrMapping = dict[Any, Any]
```
jessegrabowski:
```suggestion
GraphAttrMapping = dict[str, Any]
```

I'm just guessing, but this is probably more correct?

pymc/model_graph.py
```diff
@@ -439,6 +442,7 @@ def make_graph(
     figsize=None,
     dpi=300,
     node_formatters: NodeTypeFormatterMapping | None = None,
+    graph_attrs: GraphAttrMapping | None = None,
```
jessegrabowski: Not sure I like the custom type, it makes it seem more complex than it really is. It's not much shorter than `dict[str, Any]`

pymc/model_graph.py
```diff
@@ -460,6 +464,9 @@ def make_graph(
     node_formatters = update_node_formatters(node_formatters)
 
     graph = graphviz.Digraph(name)
+    if graph_attrs is not None:
+        graph.attr(**graph_attrs)
+
```
jessegrabowski:
```suggestion
    graph = graphviz.Digraph(name, graph_attr=graph_attrs)
```

Maybe a bit more direct? From a quick look at the docs, it seems that `graph_attr` can take None, so there's no checking necessary.

jessegrabowski (changes requested):

pymc/model_graph.py
zaxtax: I have no strong feelings about this other than it was to be more consistent with the rest of the file

jessegrabowski (approved):

pymc/model_graph.py
jessegrabowski: up to you; not a blocker

ricardoV94 (approved):

PR #7742, opened 2025-03-30
Add warning if a Minibatched variable is used without total_size

zaxtax:
## Description

This should help catch a well-known Minibatch footgun where a user forgets to include the `total_size` argument

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7742.org.readthedocs.build/en/7742/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -40,7 +40,7 @@ from pytensor.tensor.random.type import RandomType
 from pytensor.tensor.variable import TensorConstant, TensorVariable
 
 from pymc.blocking import DictToArrayBijection, RaveledVars
-from pymc.data import is_valid_observed
+from pymc.data import is_valid_observed, MinibatchOp
 from pymc.exceptions import (
     BlockModelAccessError,
     ImputationWarning,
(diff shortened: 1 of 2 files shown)


jessegrabowski (approved):
LGTM. My only suggestion is that we have some way for a user who knows he doesn't want to set `total_size` to somehow avoid the warning. I can't think of a case where anyone would want to do this, but something like `total_size=False` to silence the warning might be nice? 

Might also be too convoluted.

pymc/model/core.py
```diff
@@ -1241,6 +1241,15 @@ def register_rv(
             self.add_named_variable(rv_var, dims)
             self.set_initval(rv_var, initval)
         else:
+            if (
```
ricardoV94, line 1244: this should have checked if minibatch is anywhere in the ancestors of value not just immediately

zaxtax:
How do I do that? I'll make a fix

pymc/model/core.py
zaxtax: Do we need to do that? This check is only for observed random variables. I thought for those we don't allow the value to depend on another node

PR #7759, opened 2025-04-21
Implement reversed and strictly positive ordered transform

velochy:
## Description
Generalize Ordered transform to take two optional parameters:
positive: (default: False) also ensure values are positive. This has better geometry than chaining order with log-transform as it avoids double-exponentiation
ascending: (default: True) allow for both ascending and descending orders. Latter is achieved simply by reversing the vector in both forward and backward transforms.

This is motivated by a [draft PR into pymc-extras](https://github.com/pymc-devs/pymc-extras/pull/465) where I need a default transform for a positive and descending-ordered RV (NormalSingularValue)

## Related Issue
NA

## Checklist

## Type of change

### DIFF
--- a/pymc/distributions/transforms.py
+++ b/pymc/distributions/transforms.py
@@ -89,26 +89,48 @@ class LogExpM1(Transform):
 
 
 class Ordered(Transform):
+    """
+    Transforms a vector of values into a vector of ordered values.
+
+    Parameters
+    ----------
+    positive: If True, all values are positive. This has better geometry than just chaining with a log transform.
+    ascending: If True, the values are in ascending order (default). If False, the values are in descending order.
+    """
+
     name = "ordered"
 
-    def __init__(self, ndim_supp=None):
+    def __init__(self, ndim_supp=None, positive=False, ascending=True):
         if ndim_supp is not None:
             warnings.warn("ndim_supp argument is deprecated and has no effect", FutureWarning)
+        self.positive = positive
+        self.ascending = ascending
 
     def backward(self, value, *inputs):
-        x = pt.zeros(value.shape)
-        x = pt.set_subtensor(x[..., 0], value[..., 0])
-        x = pt.set_subtensor(x[..., 1:], pt.exp(value[..., 1:]))
-        return pt.cumsum(x, axis=-1)
+        if self.positive:  # Transform both initial value and deltas to be positive
+            x = pt.exp(value)
+        else:  # Transform only deltas to be positive
+            x = pt.empty(value.shape)
+            x = pt.set_subtensor(x[..., 0], value[..., 0])
+            x = pt.set_subtensor(x[..., 1:], pt.exp(value[..., 1:]))
+        x = pt.cumsum(x, axis=-1)  # Add deltas cumulatively to initial value
+        if not self.ascending:
+            x = x[..., ::-1]
+        return x
 
     def forward(self, value, *inputs):
-        y = pt.zeros(value.shape)
-        y = pt.set_subtensor(y[..., 0], value[..., 0])
+        if not self.ascending:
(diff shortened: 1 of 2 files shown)


pymc/distributions/transforms.py
```diff
@@ -89,26 +89,48 @@ def log_jac_det(self, value, *inputs):
...
+        self.ascending = ascending
 
     def backward(self, value, *inputs):
-        x = pt.zeros(value.shape)
-        x = pt.set_subtensor(x[..., 0], value[..., 0])
-        x = pt.set_subtensor(x[..., 1:], pt.exp(value[..., 1:]))
-        return pt.cumsum(x, axis=-1)
+        if self.positive:
+            x = pt.exp(value)
+        else:  # Everything except the first element is positive
```
ricardoV94: Comment is a but imprecise. The deltas are positive but the values may still be negative after the cumsum

pymc/distributions/transforms.py
```diff
@@ -89,26 +89,48 @@ def log_jac_det(self, value, *inputs):
...
+            x = pt.zeros(value.shape)
```
ricardoV94: We should use empty, also below

tests/distributions/test_transform.py
```diff
@@ -281,6 +281,18 @@ def test_ordered():
     vals = get_values(tr.ordered, Vector(R, 3), pt.vector, floatX(np.zeros(3)))
     assert_array_equal(np.diff(vals) >= 0, True)
 
+    # Check that positive=True creates positive and still ordered values
```
ricardoV94, line 284: Do we have a test that checks that forward/backward are inverses + jacobian that we could parametrize with the new parameters?

ricardoV94 (commented):
Looks great, I left some small comments and request for testing

tests/distributions/test_transform.py
velochy: Added them. Had to adjust the check_jacobian_det to take the absolute value and to have an absolute tolerance along a relative one, because ascending=False permutes values and this leads to a negative determinant and a slightly less clean logdet computation in general as it can't just multiply the main diagonal.

pymc/distributions/transforms.py
velochy: Fixed. But just out of curiosity - how and why does it matter?

pymc/distributions/transforms.py
velochy: Changed it to hopefully be more understandable

pymc/distributions/transforms.py
ricardoV94: It's not a fix just an optimization. Theres was another use in the other method.

It's faster because the allocated array doesn't have to be filled with zeros.

pymc/distributions/transforms.py
```diff
@@ -89,26 +89,48 @@ def log_jac_det(self, value, *inputs):
...
-        x = pt.set_subtensor(x[..., 0], value[..., 0])
-        x = pt.set_subtensor(x[..., 1:], pt.exp(value[..., 1:]))
-        return pt.cumsum(x, axis=-1)
+        if self.positive:  # Transform both initial value and deltas to be positive
+            x = pt.exp(value)
+        else:  # Transform only deltas to be positive
+            x = pt.empty(value.shape)
+            x = pt.set_subtensor(x[..., 0], value[..., 0])
+            x = pt.set_subtensor(x[..., 1:], pt.exp(value[..., 1:]))
+        x = pt.cumsum(x, axis=-1)  # Add deltas cumulativelyto initial value
```
ricardoV94: typo

ricardoV94 (approved):
Looks good, small typo

ricardoV94 (approved):

PR #7761, opened 2025-04-22
Add public testing function to mock `sample`

williambdean:
## Description

Related to https://github.com/pymc-devs/pymc-extras/issues/464

CC: @jessegrabowski

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -1028,9 +1028,9 @@ def mock_sample(draws: int = 10, **kwargs):
         coords={"chain": np.arange(n_chains)},
     )
     idata.add_groups(
-        posterior=(idata.prior.mean("chain") * expanded_chains).transpose("chain", "draw", ...),
+        posterior=(idata["prior"].mean("chain") * expanded_chains).transpose("chain", "draw", ...),
     )
-    del idata.prior
+    del idata["prior"]
     if "prior_predictive" in idata:
-        del idata.prior_predictive
+        del idata["prior_predictive"]
     return idata
--- a/pymc/testing.py
+++ b/pymc/testing.py
@@ -1036,7 +1036,7 @@ def mock_sample(draws: int = 10, **kwargs):
     return idata
 
 
-def mock_sample_setup_and_breakdown():
+def mock_sample_setup_and_teardown():
     """Set up and tear down mocking of PyMC sampling functions for testing.
 
     This function is designed to be used with pytest fixtures to temporarily replace
@@ -1055,10 +1055,10 @@ def mock_sample_setup_and_breakdown():
 
         import pytest
         import pymc as pm
-        from pymc.testing import mock_sample_setup_and_breakdown
+        from pymc.testing import mock_sample_setup_and_teardown
 
         # Register as a pytest fixture
-        mock_pymc_sample = pytest.fixture(scope="function")(mock_sample_setup_and_breakdown)
+        mock_pymc_sample = pytest.fixture(scope="function")(mock_sample_setup_and_teardown)
 
 
         # Use in a test function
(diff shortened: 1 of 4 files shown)


williambdean:
Maybe `mock_sample` isn't the best name for this function. Open to suggestions here

pymc/testing.py
```diff
@@ -982,3 +984,53 @@ def assert_no_rvs(vars: Sequence[Variable]) -> None:
...
+        random_seed=random_seed,
+        draws=draws,
+    )
+
+    expanded_chains = DataArray(
+        np.ones(n_chains),
+        coords={"chain": np.arange(n_chains)},
+    )
+    idata.add_groups(
+        posterior=(idata.prior.mean("chain") * expanded_chains).transpose("chain", "draw", ...)
```
jessegrabowski: We should also add an `observed_data` group (or maybe prior_predictive gives it to you already? I don't remember :|)

pymc/testing.py
```diff
@@ -982,3 +984,53 @@ def assert_no_rvs(vars: Sequence[Variable]) -> None:
...
+
+    .. code-block:: python
+
+        import pytest
+
+        import pymc as pm
+        from pymc.testing import mock_sample
+
+
+        @pytest.fixture(scope="module")
```
jessegrabowski, line 1009: IMO we should just provide this fixture, since this is how it will be used in most cases

jessegrabowski:
The name is OK for me, not sure what would be better

pymc/testing.py
williambdean: I've run into some issue with the scope. I forget the exact rules but it seems like there is some clash when scope is different. That is my only worry so we don't need to define all the scopes. Will explore

pymc/testing.py
williambdean: Yes, it does. Good call

EDIT : already along for the ride

williambdean:
I guess mock sounds like a verb and adjective

jessegrabowski:
It's clear from the context which is being used

williambdean:
Another subtle item here is that Flat and HalfFlat will not be mocked with this function. But that is seems a bit separate


That was needed for this:

jessegrabowski:
We could rewrite Flat and HalfFlat to Uniform and HalfUniform with big bounds (like -1e9 to 1e9) for the purposes of the tests? Is that too overkill?

williambdean:
> We could rewrite Flat and HalfFlat to Uniform and HalfUniform with big bounds (like -1e9 to 1e9) for the purposes of the tests? Is that too overkill?

I think that would be in the actual fixture since that provides the setup and breakdown. This function by itself doesn't

jessegrabowski:
Yeah for sure. But a flat-to-uniform model transformation might be in-scope for this PR.

pymc/testing.py
```diff
@@ -982,3 +984,104 @@ def assert_no_rvs(vars: Sequence[Variable]) -> None:
...
+        random_seed=random_seed,
+        draws=draws,
+    )
+
+    expanded_chains = DataArray(
+        np.ones(n_chains),
+        coords={"chain": np.arange(n_chains)},
+    )
+    idata.add_groups(
+        posterior=(idata["prior"].mean("chain") * expanded_chains).transpose("chain", "draw", ...),
```
ricardoV94: What is this line doing? Also may consider using `.expand_dims` to broadcast across fake chains

ricardoV94:
We should add an entry to the docs

williambdean:
Done

pymc/testing.py
williambdean: Yeah, good idea

jessegrabowski (approved):
lgtm

michaelosthege (approved):

jessegrabowski:
@fonnesbeck can we do a release so we can use this tool in extras?

PR #7764, opened 2025-04-27
chore: adding NoneConst and None check 

Hashcode-Ankit:
## Description

Fixes issue: https://github.com/pymc-devs/pymc/issues/7762
## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7764.org.readthedocs.build/en/7764/

### DIFF
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -291,10 +291,9 @@ def find_measurable_index_mixture(fgraph, node):
         if any(
             (
                 isinstance(indices, (type(NoneConst) | type(None)))
-                or
-                (
-                    indices.dtype.startswith("int") and 
-                    sum(1 - b for b in indices.type.broadcastable) > 0
+                or (
+                    indices.dtype.startswith("int")
+                    and sum(1 - b for b in indices.type.broadcastable) > 0
                 )
             )
             for indices in mixing_indices
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -289,7 +289,14 @@ def find_measurable_index_mixture(fgraph, node):
         # We don't support (non-scalar) integer array indexing as it can pick repeated values,
         # but the Mixture logprob assumes all mixture values are independent
         if any(
-            indices.dtype.startswith("int") and sum(1 - b for b in indices.type.broadcastable) > 0
+            (
+                isinstance(indices, (type(NoneConst), type(None)))
+                or
+                (
+                    indices.dtype.startswith("int") and 
+                    sum(1 - b for b in indices.type.broadcastable) > 0
+                )
+            )
             for indices in mixing_indices
             if not isinstance(indices, SliceConstant)
         ):
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -290,7 +290,7 @@ def find_measurable_index_mixture(fgraph, node):
         # but the Mixture logprob assumes all mixture values are independent
         if any(
             (
-                isinstance(indices, (type(NoneConst), type(None)))
+                isinstance(indices, (type(NoneConst) | type(None)))
                 or
                 (
                     indices.dtype.startswith("int") and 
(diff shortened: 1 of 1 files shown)


pymc/logprob/mixture.py
```diff
@@ -289,7 +289,13 @@ def find_measurable_index_mixture(fgraph, node):
         # We don't support (non-scalar) integer array indexing as it can pick repeated values,
         # but the Mixture logprob assumes all mixture values are independent
         if any(
-            indices.dtype.startswith("int") and sum(1 - b for b in indices.type.broadcastable) > 0
+            (
+                isinstance(indices, (type(NoneConst) | type(None)))
```
ricardoV94, line 293: I think this is what you need more directly: `isinstance(indices.type, NoneTypeT)`

pymc/logprob/mixture.py
```diff
@@ -289,7 +289,13 @@ def find_measurable_index_mixture(fgraph, node):
         # We don't support (non-scalar) integer array indexing as it can pick repeated values,
         # but the Mixture logprob assumes all mixture values are independent
         if any(
```
ricardoV94, line 291: This whole check is a bit of a mess we should think a bit more. It could also be a non-constant slice (see `if not`)

Hashcode-Ankit: @ricardoV94  by this line `it could also be a non-constant slice` do you mean to say that dynamic slices (`non-constant slice`) are also need to be skipped?

Also what is your view if we check for tensor variable and only in that case graph node get modified? (solution for mess)

ricardoV94: I think this check should be just something like

```python
if any(
            (
                    isinstance(indices, TensorVariable)
                    and indices.dtype.startswith("int")
                    and any(not b for b in indices.type.broadcastable)
                )
            )
            for indices in mixing_indices
        ):
```
Basically rulling out potentially repeated integer indices.

hash-data: Yes by tensorvariable i was saying same, the thing is in this case only we need to modify graph else return none.

Hashcode-Ankit: @ricardoV94 are you saying to modify graph node in above condition only,
 if it is true (which i understand) then we have to use all not any:
```
if all( // all not any
            (
                    isinstance(indices, TensorVariable)
                    and indices.dtype.startswith("int")
                    and any(not b for b in indices.type.broadcastable)
                )
            )
            for indices in mixing_indices
        ):
        // modify the nodes 
        
 else :
     None
```

ricardoV94: No, that `if any` I suggested is the one leading to `return None`. No need for a new branch

Hashcode-Ankit: then in None and NoneConst cases it is going to modify node which is not recommended in issue description :


I am running below code
```
from pytensor.tensor.type_other import NoneConst, NoneTypeT, SliceConstant, SliceType
from pytensor.tensor import TensorVariable
mixing_indices = [None, NoneConst]

if any(
   (
      isinstance(indices, TensorVariable)
      and indices.dtype.startswith("int")
      and any(not b for b in indices.type.broadcastable)
   )
   for indices in mixing_indices
  ):
  print("return none")
else:
  print("modify node")

```

ricardoV94: The only thing we should have to worry about is repeated indices, `None`/`NoneConst` don't introduce repeated indices, just a dummy dimension

Hashcode-Ankit: in issue:
```
The [line that I quoted above](https://github.com/pymc-devs/pymc/blob/main/pymc/logprob/mixture.py#L294) needs to also check if the indices are None or [NoneConst](https://github.com/pymc-devs/pytensor/blob/main/pytensor/tensor/type_other.py#L132). That way, the rewrite will return None when it has a mixture of integer indexes, and slices or new axis.
```

Hashcode-Ankit: Hi @ricardoV94 i have done the changes, I am new to codebase (hope you understand), that is why having that much of questions.
Thanks

PR #7783, opened 2025-05-15
Compute Metropolis average accept probability with `logsumexp`

jessegrabowski:
In the Metropolis step stats that we report on the progressbar, we have a probability of acceptance. This is currently computed in a very direct way, just `np.mean(np.exp(log_accept_prob))`. We can do better with `np.exp(logsumexp(log_accept_prob) - log(len(log_accept_prob)))`.

In really gnarly models, I've encountered overflows in this step. This might also be because there's a max(delta_logp, 1) missing from the acceptance probabilities. But even if we add that, the computation method in this PR is more numerically stable and accurate.

----
📚 Documentation preview 📚: https://pymc--7783.org.readthedocs.build/en/7783/

### DIFF
--- a/pymc/step_methods/metropolis.py
+++ b/pymc/step_methods/metropolis.py
@@ -308,13 +308,12 @@ class Metropolis(ArrayStepShared):
         else:
             accept_rate = self.delta_logp(q, q0d)
             q, accepted = metrop_select(accept_rate, q, q0d, rng=self.rng)
-            self.accept_rate_iter = accept_rate
+            self.accept_rate_iter[:] = accept_rate
             self.accepted_iter[0] = accepted
             self.accepted_sum += accepted
 
         self.steps_until_tune -= 1
 
-        self.accept_rate_iter = np.atleast_1d(self.accept_rate_iter)
         log_N = np.log(self.accept_rate_iter.shape[0])
 
         stats = {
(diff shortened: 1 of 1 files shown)


pymc/step_methods/metropolis.py
```diff
@@ -308,16 +308,18 @@ def astep(self, q0: RaveledVars) -> tuple[RaveledVars, StatsType]:
         else:
             accept_rate = self.delta_logp(q, q0d)
             q, accepted = metrop_select(accept_rate, q, q0d, rng=self.rng)
-            self.accept_rate_iter = accept_rate
+            self.accept_rate_iter[:] = accept_rate
```
ricardoV94: Nitpick, looks like the line below
```suggestion
            self.accept_rate_iter[0] = accept_rate
```

ricardoV94 (approved):

jessegrabowski:
@ricardoV94 any thoughts about just clipping the acceptance probability to be in 0,1 ?

ricardoV94:
> @ricardoV94 any thoughts about just clipping the acceptance probability to be in 0,1 ?

Is that why you were getting nans after the exp? Shouldn't hurt, as long as it's all cheap stuff

jessegrabowski:
Yes I assume it was overflowing. In the progress bar I was seeing large values for acceptance

ricardoV94:
> Yes I assume it was overflowing. In the progress bar I was seeing large values for acceptance

Then feel free to clip the exp from above or the log to be at most 0 for the returned stat

PR #7785, opened 2025-05-15
Add `model_to_minibatch` transformation to convert all `pm.Data` to `pm.Minibatch`

jessegrabowski:
## Description
A pain point for me when testing different algorithms (e.g. MCMC vs VI) is that I don't want to write a 2nd version of the model with `pm.Minibatch` on the data. 

This PR adds a model transformation that does that for the user. It's the reverse of the `remove_minibatched_nodes` transformer that @zaxtax implemented recently.

This is a WIP, it doesn't actually work now, because I can't figure out how to rebuild the observed variable with the `total_size` set correctly. Help wanted.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7785.org.readthedocs.build/en/7785/

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -92,6 +92,9 @@ minibatch_index = MinibatchIndexRV()
 class MinibatchOp(OpFromGraph):
     """Encapsulate Minibatch random draws in an opaque OFG."""
 
+    # FIXME: __props__ should not be empty
+    __props__ = ()
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs, inline=True)
 
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -92,9 +92,6 @@ minibatch_index = MinibatchIndexRV()
 class MinibatchOp(OpFromGraph):
     """Encapsulate Minibatch random draws in an opaque OFG."""
 
-    # FIXME: __props__ should not be empty
-    __props__ = ()
-
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs, inline=True)
 
--- a/pymc/model/transform/basic.py
+++ b/pymc/model/transform/basic.py
@@ -13,11 +13,9 @@
 #   limitations under the License.
 from collections.abc import Sequence
 
-from pytensor import Variable, clone_replace
+from pytensor import Variable
 from pytensor.graph import ancestors
-from pytensor.graph.fg import FunctionGraph
 
-from pymc.data import Minibatch, MinibatchOp
 from pymc.model.core import Model
 from pymc.model.fgraph import (
     ModelObservedRV,
@@ -25,7 +23,6 @@ from pymc.model.fgraph import (
     fgraph_from_model,
     model_from_fgraph,
 )
-from pymc.pytensorf import toposort_replace
 
 ModelVariable = Variable | str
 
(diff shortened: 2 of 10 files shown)


ricardoV94:
> This is a WIP, it doesn't actually work now, because I can't figure out how to rebuild the observed variable with the total_size set correctly. Help wanted.

You can use the lower level utility: https://github.com/pymc-devs/pymc/blob/ef26ae88e87c2120c2700d062e404ed1f777d358/pymc/variational/minibatch_rv.py#L53

Then make that a vanilla observed RV

ricardoV94:
Ah you already did that, so your question is how to get total size? Grab the batch shape of the variable and constant fold it without raising if it can't be fully folded

jessegrabowski:
My real issue was not understanding what needs to be the key and value in the replacements, between:

1. The model variable
2. The memo variable
3. The fgraph variable

ricardoV94:
the best is usual to replace the whole `ModelObservedRV` by a new one. You probably have to discard any dims on the batch dimension which is an input to that op

jessegrabowski:
I don't really understand what that answer means

ricardoV94:
dprint the fgraph and it will perhaps be more obvious what I am mumbling

jessegrabowski:
The problem i was running into was that I ended up with two `beta` RVs after doing the replace. Beta was the only RV implicated in the `ModelObservedRV` sub-graph

zaxtax:
Because Minibatch assumes the data variables have the same length, it might
make sense to take a variables argument. Or have some way to group data
variables of the same size (same dim name maybe?)

pymc/model/transform/basic.py
```diff
@@ -62,6 +66,47 @@ def parse_vars(model: Model, vars: ModelVariable | Sequence[ModelVariable]) -> l
...
+    data_vars = [
+        memo[datum].owner.inputs[0]
+        for datum in (model.named_vars[datum_name] for datum_name in model.named_vars)
+        if isinstance(datum, SharedVariable)
+    ]
+
+    minibatch_vars = Minibatch(*data_vars, batch_size=batch_size)
+    replacements = {datum: minibatch_vars[i] for i, datum in enumerate(data_vars)}
+    assert 0
+    # Add total_size to all observed RVs
```
ricardoV94: Should only add to those that depend on the minibatch data no?

pymc/model/transform/basic.py
```diff
@@ -62,6 +66,47 @@ def parse_vars(model: Model, vars: ModelVariable | Sequence[ModelVariable]) -> l
...
+    total_size = data_vars[0].get_value().shape[0]
```
ricardoV94: total size can be symbolic I think?

pymc/model/transform/basic.py
```diff
@@ -62,6 +66,47 @@ def parse_vars(model: Model, vars: ModelVariable | Sequence[ModelVariable]) -> l
...
+    """Replace all Data containers with pm.Minibatch, and add total_size to all observed RVs."""
+    from pymc.variational.minibatch_rv import create_minibatch_rv
+
+    fgraph, memo = fgraph_from_model(model, inlined_views=True)
+
+    # obs_rvs, data_vars = model.rvs_to_values.items()
+
+    data_vars = [
+        memo[datum].owner.inputs[0]
+        for datum in (model.named_vars[datum_name] for datum_name in model.named_vars)
```
ricardoV94: There's a `model.data_vars`. You should however allow users to specify which data vars to be minibatched (default to all is fine). Alternatively we could restrict this to models with dims, and the user has to tell us which dim is being minibatched?

That makes the graph analysis easier

zaxtax:
Yep, I have reworked this code and need to push my changes!

pymc/model/transform/basic.py
ricardoV94: The correct thing would be a dim analysis like we do for MarginaModel  to confirm the first dim of the data maps to the first dim of the observed rvs, which is when the rewrite is valid. We may not want to do that, but we should be clear about the assumptions in the docstrings. 

Example where minibatch rewrite will fail / do the wrong thing, is if you tranpose the data before you used it in the observations.

pymc/data.py
```diff
@@ -92,6 +92,9 @@ class MinibatchIndexRV(IntegersRV):
 class MinibatchOp(OpFromGraph):
     """Encapsulate Minibatch random draws in an opaque OFG."""
 
+    # FIXME: __props__ should not be empty
+    __props__ = ()
```
ricardoV94: The underlying issue is that OpFromGraph doesn't have equality implemented: https://github.com/pymc-devs/pytensor/issues/1606

jessegrabowski: You know this but for future readers, the reason these lines were added in this PR was to let the `assert_model_equality` check to pass on models with `MinibatchRV`, which is an `OpFromGraph`

ricardoV94: Yeah but we should remove this from the PR and test differently

jessegrabowski: sure

ricardoV94:
I pushed intermediate changes that I'll clean better, probably broken atm

ricardoV94:
I  pulled changes from https://github.com/pymc-devs/pymc-extras/pull/211

In general this rewrite is type-unsafe. If the variable you're trying to minibatch has static shape you can't apply the rewrite. Instead of doing a lame failure I added the functionality from that draft PR to rebuild the graph when types change. This is a very useful functionality that together with `toposort_replace` we probably want to upstream to PyTensor later.

This can be used in the `remove_minibatch` pre-existing transform, which was doing clone_replace. This rewrite wasn't complete, because it didn't remove minibatch RVs which it should as well. To do that properly we need the new implementation because we also need to substitute the non-minibatch variables in the minibatch RV graph.

I'll clean everything. One small thing still failing in the new code

Issue #7771, opened 2025-04-29
Integrate notebook comparing Stan and PyMC


ricardoV94:
### Description

I wrote a notebook comparing how models are defined in Stan and PyMC that other devs found useful and suggested we add to our core docs. If someone has the time to push that, it would be great: https://gist.github.com/ricardoV94/c421ddacb3ba7a19ac46efa253ea466c

Cocoden7:
Hey, can you attribute it to me? Thanks

vjhanani:
hello, is this issue still available?

PR #7786, opened 2025-05-16
Add the rosetta stone for PyMC and Stan notebook in docs/source/learn…

Cocoden7:
## Description

Add the `rosetta_stone_pymc_stan.ipynb` (written by @ricardoV94) that discusses how models can be translated between PyMC and Stan in `core_notebooks` documentation's folder.

## Related Issue

- [ ] Closes #
- [x] Related to #7771 

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7786.org.readthedocs.build/en/7786/

### DIFF
--- a/docs/source/learn/core_notebooks/rosetta_stone_pymc_stan.ipynb
+++ b/docs/source/learn/core_notebooks/rosetta_stone_pymc_stan.ipynb
@@ -1,27 +1,10 @@
 {
-  "nbformat": 4,
-  "nbformat_minor": 0,
-  "metadata": {
-    "colab": {
-      "provenance": [],
-      "toc_visible": true,
-      "authorship_tag": "ABX9TyOq6bvFhXaSXJ3+9UdQv7Ip",
-      "include_colab_link": true
-    },
-    "kernelspec": {
-      "name": "python3",
-      "display_name": "Python 3"
-    },
-    "language_info": {
-      "name": "python"
-    }
-  },
   "cells": [
     {
       "cell_type": "markdown",
       "metadata": {
-        "id": "view-in-github",
-        "colab_type": "text"
+        "colab_type": "text",
+        "id": "view-in-github"
       },
       "source": [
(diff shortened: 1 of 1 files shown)


zaxtax:
I would add an authorship cell as well as a watermark cell to conform with how the other notebooks are written

Cocoden7:
> I would add an authorship cell as well as a watermark cell to conform with how the other notebooks are written

I have added the authorship in the first cell. Thanks for your comment

twiecki (approved):

OriolAbril:
I had not seen that yet, I think this should be an example notebook, not a core one. But I am happy to be convinced otherwise.

---

Full context:

At some point all notebooks were here in the main repo. This was problematic because they would only get updated with the library releases even though most notebooks are not related at all to the pymc releases, in fact, many didn't even run. After several failed attempts to try and keep notebooks re-executed as part of the release process plus the complains of people who added notebooks that worked with the latest release but were not available until the following release we decided to have example notebooks and library on different release/publishing schedules.

However, due to fast development, very close in time major versions and other reasons, we ended up with notebooks breaking very often, even key/core/vital/... notebooks like the getting started, glm overview, prior/posterior predictive... so to compensate a bit we chose to move back a handful of notebooks covering the core functionality of the library **and have them be executed as part of the doc building process**. This ensured those notebooks were always synced with the library releases; and as they were only a handful of fast-ish notebooks and not the whole over 100 examples gallery it was feasible.

Second however, at some point, automatic execution broke. It hasn't been fixed yet but it probably has an issue, and afaik, the idea it to fix that so core notebooks go back to sync with the library releases.

I personally think conceptually it is not a very good fit for the "core notebooks" concept, but I think this is arguable and as I said, I am happy to be convinced otherwise. If we merge that notebook though, fixing notebook execution would mean getting this to run on every commit and PR preview, so we'd need a working env on readthedocs with both pymc and stan properly installed and correctly available through the same ipython kernel. I would prefer not doing that, and if we decide that we don't plan to fix automatic execution then all notebooks (including this one) can go back to pymc-examples (and the ones we deem "core" can still be listed somewhere else in the "Learn" section).

ricardoV94:
On the location, I think it was @fonnesbeck who suggested it as a core notebook. I'm okay either way

AdishreeB:
I would be happy to move this notebook to the examples gallery and adjust the docs links accordingly, if that aligns better with the current documentation strategy.

Issue #7788, opened 2025-05-17
BUG: Missing check for collision between coord name and data variable name


drbenvincent:
### Describe the issue:

There is a bug where it is possible to define a coord name which is also the same name as a data variable. So we are missing a check for name collisions. This leads to easy to miss downstream errors.

### Reproduceable code example:

```python
import pymc as pm


with pm.Model(coords={"a": [0, 1]}) as m:
    a = pm.Data("a", [5, 10])
    b = pm.Normal("b", dims="a")
```

### Error message:

```shell
Code snippet above should raise an exception, but does not.
```

### PyMC version information:

pymc version: 5.22.0

### Context for the issue:

The issue first showed up as a) correct sampling of random variables with `pm.sample`, b) re-sampling of random variables when calling `pm.sample_posterior_predictive`. So what was being returned was actually prior predictive samples, not posterior predictive samples. @ricardoV94 helped me track down that it was caused by a common label for a coord and a data node.

Chirag3841:
@drbenvincent I would like to solve this problem .
Here you can add collision check between Model.coords and Model.named_vars and raise ValueError immediately.

drbenvincent:
Looks like it's worth checking #7794

PR #7794, opened 2025-05-21
Fix: Prevent name collision between coords and data variables

twiecki:
Fix: Prevent name collision between coords and data variables

Adds checks to prevent you from defining a coordinate with the same name as a data variable, or vice-versa.

This addresses issue #7788, where such name collisions could lead to downstream errors, particularly with `sample_posterior_predictive` returning prior predictive samples instead of posterior predictive samples.

The following changes were made:
- Modified `Model.add_named_variable` to check if the proposed variable name already exists as a coordinate name.
- Modified `Model.add_coord` to check if the proposed coordinate name already exists as a variable name.
- Added a test case to `tests/model/test_core.py` to verify these checks.

----
📚 Documentation preview 📚: https://pymc--7794.org.readthedocs.build/en/7794/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -948,6 +948,11 @@ class Model(WithMemoization, metaclass=ContextMeta):
                 FutureWarning,
             )
 
+        if name in self.named_vars:
+            raise ValueError(
+                f"Name '{name}' already exists as a variable name in the model. Please choose a different name for the coordinate."
+            )
+
         if name in {"draw", "chain", "__sample__"}:
             raise ValueError(
                 "Dimensions can not be named `draw`, `chain` or `__sample__`, "
@@ -1463,6 +1468,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
         """
         if var.name is None:
             raise ValueError("Variable is unnamed.")
+        if var.name in self.coords:
+            raise ValueError(
+                f"Name '{var.name}' already exists as a coordinate name in the model. Please choose a different name for the variable."
+            )
         if self.named_vars.tree_contains(var.name):
             raise ValueError(f"Variable name {var.name} already exists.")
 
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -950,7 +950,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         if name in self.named_vars:
             raise ValueError(
-                f"Name '{name}' already exists as a variable name in the model. Please choose a different name for the coordinate."
+                f"Name '{name}' already exists as a variable name in the model. Please choose a different name for the dimension."
             )
 
         if name in {"draw", "chain", "__sample__"}:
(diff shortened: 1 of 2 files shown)


pymc/model/core.py
```diff
@@ -948,6 +948,11 @@ def add_coord(
                 FutureWarning,
             )
 
+        if name in self.named_vars:
+            raise ValueError(
+                f"Name '{name}' already exists as a variable name in the model. Please choose a different name for the coordinate."
```
```suggestion
                f"Name '{name}' already exists as a variable name in the model. Please choose a different name for the dimension."
```

pymc/model/core.py
```diff
@@ -1463,6 +1468,10 @@ def add_named_variable(self, var, dims: tuple[str | None, ...] | None = None):
         """
         if var.name is None:
             raise ValueError("Variable is unnamed.")
+        if var.name in self.coords:
+            raise ValueError(
+                f"Name '{var.name}' already exists as a coordinate name in the model. Please choose a different name for the variable."
```
ricardoV94, line 1473:
```suggestion
                f"Name '{var.name}' already exists as a dimension name in the model. Please choose a different name for the variable."
```

ricardoV94 (changes requested):
Looks good, just a small change in the error message and agree with copilot to test with a distribution as well as data

tests/model/test_core.py
```diff
@@ -99,6 +99,16 @@ def test_setattr_properly_works(self):
                 assert len(submodel.value_vars) == 2
             assert len(model.value_vars) == 3
 
+    def test_name_conflict_variable_and_coord(self):
```
ricardoV94, line 102: wrong test class to put this test in, this is for nested models tests

drbenvincent:
## Update: Duplicate PR #8093 and what to take from it

@twiecki @ricardoV94 — heads up that PR #8093 by @Chirag3841 addresses the same issue (#7788) with a very similar approach. I've done a detailed comparison of both PRs and posted it on #8093.

The preliminary recommendation is to proceed with this PR (#7794) since it has the correct check ordering (validation before state mutations) and already has Ricardo's review. However, there is one improvement from #8093 worth adopting:

### Cherry-pick: `tree_contains` in `add_coord`

This PR currently uses `name in self.named_vars` in `add_coord`, which only checks the current model's variables. PR #8093 uses `self.named_vars.tree_contains(name)`, which traverses parent models in nested model scenarios. The latter is more correct and consistent with how `add_named_variable` already uses `tree_contains` for the existing duplicate-variable check.

### Remaining items from Ricardo's original review

1. Error message wording: "dimension" instead of "coordinate" (partially addressed)
2. Move test to a proper class (not `TestNested`)
3. Add a distribution-based test (not just `pm.Data`)
4. Investigate the pre-existing failing test that Ricardo flagged — existing code may partially handle this already

We're waiting for input from @twiecki and @ricardoV94 before making a final call on closing #8093.

PR #7810, opened 2025-06-04
Implement specialized Hurdle distribution

ricardoV94:
It indirectly addresses the issue reported in in https://github.com/pymc-devs/nutpie/issues/163

The new objects have a logp that handles the discrete + continuous process correctly, without requiring the arbitrary truncation of the latter at epsilon. This provides a cheaper and more stable logp / logcdf.
For discrete variables we keep using a truncation

~~Also added special logic to truncate a Hurdle distribution which solves https://github.com/bambinos/bambi/issues/768~~, this is not the desired behavior, reverted it

CC @zwelitunyiswa

----
📚 Documentation preview 📚: https://pymc--7810.org.readthedocs.build/en/7810/

### DIFF
--- a/pymc/distributions/mixture.py
+++ b/pymc/distributions/mixture.py
@@ -56,10 +56,8 @@ __all__ = [
 ]
 
 
-class MarginalMixtureRV(SymbolicRandomVariable):
-    """A placeholder used to specify a distribution for a mixture sub-graph."""
-
-    _print_name = ("MarginalMixture", "\\operatorname{MarginalMixture}")
+class _BaseMixtureRV(SymbolicRandomVariable):
+    """Base class SymbolicRandomVariable for Mixture and Hurdle RVs."""
 
     @classmethod
     def rv_op(cls, weights, *components, size=None):
@@ -139,7 +137,7 @@ class MarginalMixtureRV(SymbolicRandomVariable):
             comps_s = ",".join(f"({s})" for _ in components)
             extended_signature = f"[rng],(w),{comps_s}->[rng],({s})"
 
-        return MarginalMixtureRV(
+        return cls(
             inputs=[mix_indexes_rng, weights, *components],
             outputs=[mix_indexes_rng_next, mix_out],
             extended_signature=extended_signature,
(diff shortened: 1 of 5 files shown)


tests/distributions/test_distribution.py
```diff
@@ -92,8 +92,14 @@ def test_all_distributions_have_support_points():
 
     dists = (getattr(dist_module, dist) for dist in dist_module.__all__)
     dists = (dist for dist in dists if isinstance(dist, DistributionMeta))
+    generic_func = _support_point.dispatch(object)
     missing_support_points = {
-        dist for dist in dists if getattr(dist, "rv_type", None) not in _support_point.registry
+        dist
+        for dist in dists
+        if (
+            getattr(dist, "rv_type", None) is not None
```
ricardoV94, line 100: This was too restrictive, a subclass also inherits the dispatch function, and need not be in the registry explicitly

pymc/distributions/truncated.py
```diff
@@ -472,3 +476,32 @@ def _truncated_normal(op, lower, upper, size, rng, old_size, mu, sigma):
...
+        1,
+    )
+    adjusted_weights = pt.stack([1 - adjusted_psi, adjusted_psi], axis=-1)
+
+    # The only remaining step is to truncate the other distribution
+    truncated_dist = Truncated.dist(dist, lower=lower, upper=upper, max_n_steps=max_n_steps)
+
+    # Creating a hurdle with the adjusted weights and the truncated distribution
+    # Should be equivalent to truncating the original hurdle distribution
+    return op.rv_op(adjusted_weights, zero_dist, truncated_dist, size=size)
```
ricardoV94: Actually this is not quite correct, one must also adjust psi by the amount of mass truncated in the other dist...

Also users may be more interested in a Hurdle with a truncated dist, instead of a truncated hurdle, as it won't distort the meaning of the original psi

This is something that could be added at the bambi level perhaps @tomicapretto?

zwelitunyiswa:
> It indirectly addresses the issue reported in in [pymc-devs/nutpie#163](https://github.com/pymc-devs/nutpie/issues/163)
> 
> The new objects have a logp that handles the discrete + continuous process correctly, without requiring the arbitrary truncation of the latter at epsilon. This provides a cheaper and more stable logp / logcdf. For discrete variables we keep using a truncation
> 
> Also added special logic to truncate a Hurdle distribution which solves [bambinos/bambi#768](https://github.com/bambinos/bambi/issues/768)
> 
> CC @zwelitunyiswa
> 
> 📚 Documentation preview 📚: https://pymc--7810.org.readthedocs.build/en/7810/

@ricardoV94 This is amazing. Thank you so much for this!

pymc/distributions/truncated.py
zwelitunyiswa: @ricardoV94  In my use case I am interested in constraining the gamma distribution of the hurdle-gamma. The reason is that there are natural upper limits as it relates to body parts (i.e. the injury area cannot exceed the size of the body part or some proportion thereof). Therefore, any predictions are above that threshold would not make clinical sense.

ricardoV94: It's a subtle difference, a `Truncated(Hurdle(psi, Gamma))` vs `Hurdle(psi, Truncated(Gamma))`

The generative process of the first is that you resample the whole hurdle process everytime you get a draw out of the truncation bounds, which necessarily inflates the zeros above (1-psi).

In the second is the usual hurdle process, you either get zeros with (1-psi) or you get non zeros from a truncated gamma.

I guess your is more like the second?

zwelitunyiswa: @ricardoV94 It is the Hurdle(psi, Truncated(Gamma)).

ricardoV94: Then I suspect some functionality needs to go into bambi. You can create that distribution already in PyMC, using the same strategy with which HurdleGamma is defined, but using a TruncatedGamma as the dist.

Does Bambi allow to use arbitrary PyMC distributions or only a subset that it already understands? 

Because otherwise the constrained thing in bambi would do the `Truncated(Hurdle(Gamma))` not the one you want.

zwelitunyiswa: Bambi does allow arbitrary non-mixture continuous PYMC distributions (the way @tomicapretto implemented it is very intuitive and flexible). Ah, so constrained would do Truncated(Hurdle(psi, Gamma)) and inflate 0 counts. Good to know, especially that is very important in what I study. Bambi allows you to apply truncated priors, but erroneous data could overwhelm that prior ... I should actually test that.

ricardoV94: I removed the commit to truncate hurdles. Instead what you want seems to be this:

```python
import pymc as pm
from pymc.distributions import Gamma, Truncated
from pymc.distributions.mixture import _Hurdle

class HurdleTruncatedGamma:
    def __new__(cls, name, psi, alpha=None, beta=None, mu=None, sigma=None, upper=None, **kwargs):
        return _Hurdle._create(
            name=name,
            nonzero_p=psi,
            nonzero_dist=Truncated.dist(Gamma.dist(alpha=alpha, beta=beta, mu=mu, sigma=sigma), upper=upper),
            **kwargs,
        )

    @classmethod
    def dist(cls, psi, alpha=None, beta=None, mu=None, sigma=None, upper=None, **kwargs):
        return _Hurdle._create(
            name=None,
            nonzero_p=psi,
            nonzero_dist=Truncated.dist(Gamma.dist(alpha=alpha, beta=beta, mu=mu, sigma=sigma), upper=upper),
            **kwargs,
        )
    
dist = HurdleTruncatedGamma.dist(psi=0.5, mu=20, sigma=10, upper=30, size=(1000,))
draws = pm.draw(dist, random_seed=48)
print((draws == 0).mean(), draws.max())  # 0.5 29.921184342002615
```

That means we may still need the scalarloop implemented in numba, since I suspect the special operation will show up in the gradient of the Truncated Gamma. Anyway, this PR already makes hurdles less expensive by default, just not this special Hurdle of a Truncated variable.

zaxtax (approved):
LGTM! We should probably add a test similar to one that motivate this PR in the first class

ricardoV94:
> LGTM! We should probably add a test similar to one that motivate this PR in the first class

We have the pre-existing hurdlesl tests, in a sense this is just a refactor/optimization. Can't think of anything ~~reasonable~~ obvious to test here?

zaxtax:
> > LGTM! We should probably add a test similar to one that motivate this PR in the first class
> 
> We have the pre-existing hurdlesl tests, in a sense this is just a refactor/optimization. Can't think of anything ~reasonable~ obvious to test here?

What caused the error originally reported here? https://github.com/pymc-devs/nutpie/issues/163 Does that have a test already?

ricardoV94:
> > > LGTM! We should probably add a test similar to one that motivate this PR in the first class
> > 
> > We have the pre-existing hurdlesl tests, in a sense this is just a refactor/optimization. Can't think of anything ~reasonable~ obvious to test here?
> 
> What caused the error originally reported here? https://github.com/pymc-devs/nutpie/issues/163 Does that have a test already?

That was fixed sometime ago in PyTensor:  https://github.com/pymc-devs/pytensor/pull/1137

The performance question when in numba is addressed by https://github.com/pymc-devs/pytensor/pull/1445

Neither is PyMC specific

zaxtax:
Sounds good to me!

pymc/distributions/mixture.py
```diff
@@ -56,8 +56,8 @@
 ]
 
 
-class MarginalMixtureRV(SymbolicRandomVariable):
-    """A placeholder used to specify a distribution for a mixture sub-graph."""
+class _BaseMixtureRV(SymbolicRandomVariable):
```
lucianopaz, line 59: Should you also change the `_print_name`?

zaxtax:
> > > > LGTM! We should probably add a test similar to one that motivate this PR in the first class
> > > 
> > > 
> > > We have the pre-existing hurdlesl tests, in a sense this is just a refactor/optimization. Can't think of anything ~reasonable~ obvious to test here?
> > 
> > 
> > What caused the error originally reported here? [pymc-devs/nutpie#163](https://github.com/pymc-devs/nutpie/issues/163) Does that have a test already?
> 
> That was fixed sometime ago in PyTensor: [pymc-devs/pytensor#1137](https://github.com/pymc-devs/pytensor/pull/1137)
> 
> The performance question when in numba is addressed by [pymc-devs/pytensor#1445](https://github.com/pymc-devs/pytensor/pull/1445)
> 
> Neither is PyMC specific

Feel free to merge whenever you feel comfortable! I think it's good to go

pymc/distributions/mixture.py
```diff
@@ -333,39 +222,32 @@ def change_marginal_mixture_size(op, dist, new_size, expand=False):
...
-    mix_logp = check_parameters(
-        mix_logp,
-        0 <= weights,
-        weights <= 1,
-        pt.isclose(pt.sum(weights, axis=-1), 1),
-        msg="0 <= weights <= 1, sum(weights) == 1",
-    )
-
-    return mix_logp
+    mix_support_point = pt.sum(weights * support_point_components, axis=mix_axis)
```
lucianopaz, line 241: Why not use the logsumexp and have log scale weights here? Is it because the weights are already in the 0-1 range and taking the log won’t help with precision?

lucianopaz (approved):
Looks good to me. I just left two very minor comments

pymc/distributions/mixture.py
ricardoV94: We're not computing any log quantities nor starting with any log quantities so I don't think it would help. Also the initial point is not so critical?

pymc/distributions/truncated.py
tomicapretto: @ricardoV94 @zwelitunyiswa yes, that is something we can do at the bambi level :)

pymc/distributions/mixture.py
ricardoV94: We should get rid of those, we don't use them for anything, but will fix

PR #7814, opened 2025-06-09
Advanced sub tensor with none and integers fix

### LINKED ISSUE #7762: BUG: `AdvancedSubTensor` with `None` and integer indices raises a `logprob` error instead of silently failing
### Describe the issue:

I just ran into a logprob rewrite error with an `AdvancedSubTensor` op that mixed `None` entries and `int32` indices together. This wasn't actually a mixture model but logprob found the op and tried to apply its rewrite rules and raised an error instead of just failing silently. The problem seems to be from [this line](https://github.com/pymc-devs/pymc/blob/main/pymc/logprob/mixture.py#L294) that doesn't include a guard against a `None` constant as well as a `slice` constant.

```python
import numpy as np
import pymc as pm


obs = np.random.default_rng().normal(size=(7, 4))
with pm.Model():
   inds = np.arange(obs.shape[1])
   a = pm.Normal("a", shape=10)
   b = pm.Deterministic("b", a[None, inds])
   c = pm.Normal("c", mu=b, sigma=1, observed=obs)
   pm.sample()
```

### Error message:

```python
ERROR (pytensor.graph.rewriting.basic): Rewrite failure due to: find_measurable_index_mixture
ERROR (pytensor.graph.rewriting.basic): node: AdvancedSubtensor(a, NoneConst{None}, [0 1 2 3])
ERROR (pytensor.graph.rewriting.basic): TRACEBACK:
ERROR (pytensor.graph.rewriting.basic): Traceback (most recent call last):
  File "pytensor/graph/rewriting/basic.py", line 1913, in process_node
    replacements = node_rewriter.transform(fgraph, node)
  File "pytensor/graph/rewriting/basic.py", line 1085, in transform
    return self.fn(fgraph, node)
  File "pymc/logprob/mixture.py", line 291, in find_measurable_index_mixture
    if any(
  File "pymc/logprob/mixture.py", line 292, in <genexpr>
    indices.dtype.startswith("int") and sum(1 - b for b in indices.type.broadcastable) > 0
AttributeError: 'Constant' object has no attribute 'dtype'. Did you mean: 'type'?
```
but sampling works fine because the rewrite was actually supposed to fail and return `None`.

### PyMC version information:

Github main

### Context for the issue:

This doesn't really affect anything. It just confuses regular users that see the error traceback from rewriting and get alarmed. It would be more elegant to handle this extra indexer type just like with slice constants.

golocule:
## Description

Fixed the AdvancedSubTensor with None and Integer values bug of having a logprob error instead of gracefully crashing through including a check for NoneConst types in mixture.py. None is returned if None values are found in a logprob. Added a test in test_mixture.py called test_advanced_subtensor_none_and_integer that validates the bug is fixed through checking there is a runtime error instead of an internal error when a logprob dictionary of integers and None values is attempted to be created.

## Related Issue

- [x] Closes #7762
- [ ] Related to #

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7814.org.readthedocs.build/en/7814/

### DIFF
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -291,7 +291,9 @@ def find_measurable_index_mixture(fgraph, node):
         if any(
             indices.dtype.startswith("int") and sum(1 - b for b in indices.type.broadcastable) > 0
             for indices in mixing_indices
+            # this is the line with the bug
             if not isinstance(indices, SliceConstant)
+            if not isinstance(indices, type(NoneConst))
         ):
             return None
 
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -291,7 +291,6 @@ def find_measurable_index_mixture(fgraph, node):
         if any(
             indices.dtype.startswith("int") and sum(1 - b for b in indices.type.broadcastable) > 0
             for indices in mixing_indices
-            # this is the line with the bug
             if not isinstance(indices, SliceConstant)
             if not isinstance(indices, type(NoneConst))
         ):
--- a/tests/logprob/test_mixture.py
+++ b/tests/logprob/test_mixture.py
@@ -1156,3 +1156,26 @@ def test_nested_ifelse():
     np.testing.assert_almost_equal(mix_logp_fn(0, test_value), sp.norm.logpdf(test_value, -5, 1))
     np.testing.assert_almost_equal(mix_logp_fn(1, test_value), sp.norm.logpdf(test_value, 0, 1))
     np.testing.assert_almost_equal(mix_logp_fn(2, test_value), sp.norm.logpdf(test_value, 5, 1))
+
+
+def test_advanced_subtensor_none_and_integer():
+    a = pt.random.normal(0, 1, size=(10,), name="a")
+    inds = np.array([0, 1, 2, 3], dtype="int32")
+    b = a[None, inds]
+
+    b_val = b.type()
+    b_val.name = "b_val"
+    a_val = a.type()
+    a_val.name = "a_val"
+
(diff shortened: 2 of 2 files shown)


tests/logprob/test_mixture.py
```diff
@@ -1156,3 +1156,26 @@ def test_nested_ifelse():
...
+    a = pt.random.normal(0, 1, size=(10,), name="a")
+    inds = np.array([0, 1, 2, 3], dtype="int32")
+    b = a[None, inds]
+
+    b_val = b.type()
+    b_val.name = "b_val"
+    a_val = a.type()
+    a_val.name = "a_val"
+
+    try:
```
ricardoV94, line 1171: Use pytest.raises

pymc/logprob/mixture.py
```diff
@@ -292,6 +292,7 @@ def find_measurable_index_mixture(fgraph, node):
             indices.dtype.startswith("int") and sum(1 - b for b in indices.type.broadcastable) > 0
             for indices in mixing_indices
             if not isinstance(indices, SliceConstant)
```
ricardoV94, line 294:
```suggestion
             if not isinstance(indices.type, SliceType | NoneTypeT)
```

ricardoV94 (changes requested):

asifzubair:
Hi @golocule , are you still working on this ? If not, would you mind if I take it over ? Thank you 🙏

ricardoV94:
Fixed by #7877

Issue #15, opened 2011-02-26
pymc.Matplot.plot sometimes ignore path argument


kforeman:
If you initialize a model using vars() and then plot it, the "compound" objects (e.g. MVN) don't use the specified path when saving.

a = pymc.MvNormalCov('a',[10,10],[[1,0],[0,1]])
b = pymc.Normal('b',10,1)

m = pymc.MCMC(vars())
m.sample(10)

pymc.Matplot.plot(m.a, path='C:/tmp', format='pdf')
pymc.Matplot.plot(m.b, path='C:/tmp', format='pdf')

In this case b.pdf will be saved in C:/tmp, but a.pdf will be placed in the working directory. Tested in both 2.1alpha and latest dev version.

pymc-devs:
Fixed in 6c1954f55ce5255dd954d634211e1b137b0aaf27. Thanks.

PR #7820, opened 2025-06-17
Add experimental `dims` module with objects that follow dim-based semantics (like xarray without coordinates)

ricardoV94:
This builds on top of PyTensor [xtensor module](https://pytensor.readthedocs.io/en/latest/library/xtensor/index.html), to introduce distributions and model objects that follow xarray-like semantics (without coordinates). Example model:

```python
import numpy as np
import pymc as pm
import pymc.dims as pmd

# Very realistic looking data!
observed_response_np = np.ones((5, 20), dtype=int)
coords = coords = {
    "participant": range(5),
    "trial": range(20),
    "item": range(3),
}
with pm.Model(coords=coords) as dmodel:
    observed_response = pmd.Data(
        "observed_response", observed_response_np, dims=("participant", "trial")
    )

    # Participant constant preferences for each item
    participant_preference = pmd.ZeroSumNormal(
        "participant_preference", 
        core_dims="item", 
        dims=("participant", "item"),
    )

    # Shared time effects across all participants
    time_effects = pmd.Normal("time_effects", dims=("item", "trial"))

    trial_preference = pmd.Deterministic(
        "trial_pereference",
        participant_preference + time_effects,
        dims=(...,),  # No need to specify, PyMC knows them
    )

    response = pmd.Categorical(
        "response",
        p=pmd.math.softmax(trial_preference, dim="item"),
        core_dims="item",
        observed=observed_response,
        dims=(...,), # No need to specify, PyMC knows them
    )
```

Equivalently, with the traditional API:
```python
    observed_response = pm.Data(
        "observed_response", observed_response_np, dims=("participant", "trial")
    )

    # Participant constant preferences for each item
    participant_preference = pm.ZeroSumNormal(
        "participant_preference", 
        n_zerosum_axes=1,
       dims=("participant", "item"),
    )

    # Shared time effects across all participants
    time_effects = pm.Normal("time_effects", dims=("trial", "item"))

    trial_preference = pm.Deterministic(
        "trial_preference",
        participant_preference[:, None, :] + time_effects[None, :, :],
        dims=("participant", "trial", "item"),
    )

    response = pm.Categorical(
        "response",
        p=pm.math.softmax(trial_preference, axis=-1),
        observed=observed_response,
        dims=("participant", "trial"),
    )
```

More details in the new [core notebook](https://www.pymc.io/projects/docs/en/stable/learn/core_notebooks/dims_module.html) and [WIP docs](https://www.pymc.io/projects/docs/en/stable/api/dims.html)

<details><summary>Long dev discussion</summary>

## Points of contention
People have expressed dissatisfaction with some aspects of the current approach. I'll try to summarize it here as fair as possible, and feel free to correct me in the comments

### Dims being strings

Current design: dims are just strings that label the axis of a tensor.

There are 3 concerns with this simplicity:
1. Typos can introduce bugs. Say you have time and item it may be easy to mistype between the two. I don't think this is inherent to strings and the user could always assign the string `item` or `time` to something less confusing like `product_dim = "item"`, or just use other labels.
2. Dims can be accidentally repeated if one is not careful. See the contortions one may need to go through to avoid this in the current implementation of MvNormal:
```python
chol = cls._as_xtensor(chol)
# chol @ chol.T in xarray semantics requires a rename
safe_name = "_"
if "_" in chol.type.dims:
    safe_name *= max(map(len, chol.type.dims)) + 1
cov = chol.dot(chol.rename({d0: safe_name}), dim=d1).rename({safe_name: d1})
```
Obviously we can have `safe_name = get_discardable_new_dim_name(chol.type.dims)`, but the point still stands that it can be awkward to do abstract work on dims.

3. Dims don't provide enough structure / error-checking in them. Having `x[1:]` come out with the same dims as `x` is error-prone because you may then accidentally add them back and get an error only at runtime. I agree partially with this concern.

Note: It's not that dims must be strings (although they do now). Dims could be made to be any hashable / pickleable object, not just strings without much work in PyTensor. When we talk of richer dims, we mean something like: https://github.com/pymc-devs/pytensor/pull/407

Counterpoints:
1. I think the shape problem is mostly mitigated by using static shapes in PyTensor. XTensorVariables still have the  `type.shape` attribute and if you try to add `x + x[1:]`, if x had a static shape it would raise immediately. However we don't use static shapes in PyMC by default, all dims start as being mutable and can later be made constant with helpers like `freeze_data_and_dims`, which the jax / nutpie samplers do by default, but not the pymc ones.

2. It's much easier to write the actual model. You can say `concat([x, y, x], dim="time")`, whereas with specialized dims this may not be possible or actively forbidden. Alternative APIs may look like some variant of `concat([x, y, x], dim=(x.dim[0], y.dim[1], x.dim[0])`, which imo defeats the main purpose of having dims (not having to worry about dimension position in the underlying memory). 
 
3. Alternatively, richer structured dims introduced by PyTensor automatically (say when you do `x[::-1]`) could be made somehow comparable / matcheable to the original dims the user provided. So an operator like concat may be able to treat the dims of `x[::-1].dims` as identical to the original `x.dims` for the purpose of deciding which axis to align. If this can be achieved, it should require no user-facing changes in the current code and so the existing code could be merged as is.

### No coordinates
As you can see from the comment below by @twiecki (and it's indeed the first thing everyone asks), people wish to be able to do coordinate based operations, mainly selection like `x.sel(time="yesterday")`. That's not possible in the current implementation (a useful error message is raised if you try to use that method).

Also it creates a hard problems for us. This is discussed in the new notebook, but in summary, the current PyMC API uses coords to actually define the model *AND* propagates these to xarray objects after sampling, whenever the user tells us that variable has the same dims. With the new dims model we always know the dims of every variable, so we should always propagate?

Not so easy. The original coordinates are not trivially valid for intermediate operations. They may have the wrong length or order. So what options do we have?
1. Don't propagate dims by default, because of the 1% of the time it can create problems. I think this is fundamentally wrong, because if a user then goes and tries to do this, they get orthogonal broadcast:

```python
# Note this is regular PyMC code
with pm.Model(coords={"time": range(2)}) as m:
  x = pm.Normal("x", dims=("time",))
  x_reversed = pm.Deterministic("x_reversed", x[::-1])

  prior = pm.sample_prior_predictive(draws=1).prior.squeeze(("chain", "draw"))
  assert (prior["x"] + prior["x_reversed"]).shape == (2, 2)
```

2. Don't propagate coords by default. Then the behavior is just like the one in the PyMC model. This is obviously infuriating for users, why do you ask for coords if you don't use them besides getting their size? Totally agree. 

**Also note that xarray Datasets always need coords, even if individual DataArrays do not**. We have to decide what to do with coords one way or another :)

This PR takes the most dumb pragmatic approach. Keep coords unless you detect a shape inconsistency. In that case, refuse to propagate them with a warning. Options for users are to use custom coords for intermediate operations (in which case they also need a custom dim, since PyMC doesn't have a concept of coords per variable (https://github.com/pymc-devs/pymc/issues/7852) or to use the old `pm.Deterministic` without dims (instead of pm.dims.Deterministic) and get the behavior mentioned above (no custom coords nor dims). 

If users care about coords they have to mend them manually after the fact. This is discussed at the end of the new notebook (still have to add example of mending).

#### So why don't we have coords in the graph? 

I don't think this can be realistically done in a way that is both performant, simple and doesn't restrict us to static shapes.

Just give up on static shapes? Some people really doubt that we need this flexibility in PyTensor (JAX does just fine without it right?). I **strongly** disagree. Just to give one example: It rules out any sort of symbolic iterative algorithms unless you pad everything to the worst case scenarios. Saying this example is not worthwhile is not saying we never need runtime shapes! It's just one example where it's needed, there are more.

Besides getting rid of runtime shapes doesn't make it trivially easy to implement and work with coords abstractly. A lot of PyMC magic is transforming user code to go from graph A -> graph B. Adding coordinates increases code complexity.

### These are related concerns
Talking about concerns for the lack of coords circles back to the concerns about dims listed above. coords in a sense supercharge dims in xarray. Look at how the behavior of `x[:-idx] + x[idx:]` changes when you have arrays with coordinates vs without.

```python
import xarray as xr

x = xr.DataArray([0, 1, 2, 3], dims=("day",))
y = xr.DataArray([0, 1, 2, 3], dims=("day",), coords={"day": ["Mon", "Tue", "Wed", "Thu"]})

print((x[1:] + x[:-1]).values)  # [1 3 5]
print((y[1:] + y[:-1]).values)  # [2 4]

print((x - x[::-1]).values)  # [-3 -1  1  3]
print((y - y[::-1]).values)  # [0 0 0 0]
```

If we know the coordinates of variables, it's no longer confusing that things may have different shapes. There are clear rules how to solve the discrepancy (which are controlled by global variables in xarray..., which may affect how much you like it.)

**Reminder: The current approach works the same as xarrays without coords.** Some proposals for richer dims without coords would still have to decide what `x[idx:] + x[:-idx]` should do. Options and my opinionated downsides are
1. It fails at write time (worry: writing models becomes cumbersome)
6. It warns at write time with a helpful suggestion of how to do it correctly (worry: mypy PTSD flashbacks)
7. Dims broadcast orthogonally (worry: it's never what the user would expect)
8. Work like it does now (in which case it need to block us merging the current code)

## Final note
Perfection is the enemy of progress. We have had 5 solid years with deficient use of coords/dims. People seem to like it, and I don't think anyone is doing anything better out there. I believe this is a solid incremental step that brings dims to the model for real, while still handling coords deficiently.

I'm not against coords, I am against bad performance, code complexity and static-shape handicaps (in that order). Maybe these problems can be solved, but they won't be solved by me. 

I'm also not against more structured dims at the PyTensor level. Happy to help out there. 

I also don't think we should wait. The XTensor stuff, simple as it is, took a while. I opened a draft PR two summers ago, and there is still a lot to do. I would like to share it and start iterating with user feedback, instead of just theoretical concerns. I'm optimistic that this would both inform a more advanced approach and allow us to transition from a less "beta" stage than the one we are at now.

I'm also optimistic the community will accept the experimental / exploratory nature of the new approach, and that we may drop it for something else down the road, that we believe is even better.

</details>


pymc/dims/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+    # Register lower_xtensor rewrites in logprob_rewrites_db
+    logprob_rewrites_db.register("pre_lower_xtensor", optdb.query("+lower_xtensor"), "basic")
+    _XRV.append(XRV)
+    INITIAL_POINT_REWRITES.append("lower_xtensor")
+
+    # TODO: Better model of probability of bugs
+    day_of_conception = datetime.date(2025, 6, 17)
+    day_of_last_bug = datetime.date(2025, 6, 17)
+    today = datetime.date.today()
+    days_with_bugs = (day_of_last_bug - day_of_conception).days
```
twiecki, line 53: wtf 😆

ricardoV94: This has two purposes: distract reviewers so they don't focus on the critical changes, and prove that OSS libraries can't be fun.

twiecki:
Can this index using labels? `x["a"]`

ricardoV94:
> Can this index using labels? `x["a"]`

I don't know what `x["a"]` means :). 

Is `"a"` a coordinate? `x.loc["a"]` would be the xarray syntax? You can't do that.

Like in xarray, you can do `x.isel(dim=idxs)` or `x[{dim: idxs}]`. 

You cannot do `x.sel(dim=coords)` or `x.loc[coords]`

The new PyTensor objects have dims but not coords. It's not trivial to encode coord based operations in our backends.

twiecki:
We should make this the 6.0 release.

ricardoV94:
> We should make this the 6.0 release.

I agree, but would perhaps wait until we beta-tested it to the point it no longer feels too experimental

docs/source/learn/core_notebooks/dims_module.ipynb
OriolAbril, line 1: Inside the notebook you have to use myst syntax, so you'll have to replace the ``:reftype:`reftarget` `` with  ``{reftype}`reftarget` ``

OriolAbril (commented):
adding one extra comment about myst syntax but outside reviewnb because it always messes up rendering of backticks.

I will try to play around and build/port some models one of these days, and look at the code itself while I do that

docs/source/learn/core_notebooks/dims_module.ipynb
```diff
@@ -0,0 +1,1666 @@
+{
```
OriolAbril, line 1: <p>I think this is too high up in the notebook as I envision this as something beginner users should not do, but I think it would be nice to show how to get array values for an XTensorVariable into their corresponding DataArray, the <code>xr.DataArray(pm.draw(...), dims=outer_addition.type.dims)</code> . Maybe this could go into a dims FAQ? It could be started at the bottom of this page and if we see it grows extend it to their own page.</p><p> </p><p>Or start a global FAQ page with whichever questions/answers in the discord post are still relevant and add a <code>pymc.dims</code> section. Feel free to make an issue for this to keep the PR scope clear. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/7820/discussion/#comment-2179899760'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='23'/>

OriolAbril, line 1: <p>minor nits, it currently says <code>isel</code> won't be supported instead of <code>sel</code>. for the link to the dist method to work once we add api docs for the module it needs to be <code>:meth:</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/7820/discussion/#comment-2179899767'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='31'/>

OriolAbril, line 1: <p>Add a tilde before <code>pytensor.xtensor...</code> (still inside the tildes) so it is rendered as <code>values</code> only instead of the full name. We should also double check the attribute is documented explicitly in the pytensor docs, otherwise this won't be rendered as a cross-reference  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/7820/discussion/#comment-2179899769'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='33'/>

OriolAbril, line 1: <p>Given we have just looked at the number of nodes as a measure of model complexity/efficiency. I think we should show that to for the <code>dims_splines_model</code> to show it is roughly the same as the vectorized model, but much easier to read, and (arguably?) easier to code too</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/7820/discussion/#comment-2179899773'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='57'/>

OriolAbril, line 1: <p>I am skipping this last part for now to see if we converge on the custom object dim approach on pytensor which should make this a non-issue</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/7820/discussion/#comment-2179899775'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='63'/>

ricardoV94: Good idea, I think we can sneak it in at the end where I mention coordinates. Add a "why are the variables not producing DataArrays directly" with this recipe

docs/source/learn/core_notebooks/dims_module.ipynb
```diff
@@ -0,0 +1,2910 @@
...
```
tomicapretto, line 1: <p>Is <code>pymc.dims.as_xtensor</code>correct? I find it surprising to use <code>as_xtensor</code> to get a regular non-dimmed variable.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/7820/discussion/#comment-2200857928'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='36'/>

tomicapretto, line 1: <p>Is it worth adding a note explaining you use <code>.observed_RVs</code> first and <code>.basic_RVs</code> second?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/7820/discussion/#comment-2200857933'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='59'/>

tomicapretto, line 1: <pre class="ql-syntax" spellcheck="false">Line #15.        beta = pm.Deterministic("beta", pmd.concat([beta0, spline_slopes.diff("knot")], dim="knot"))
</pre><p>Does pm.Deterministic support both <code>dim</code> and <code>dims</code>? Is it on purpose?   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/7820/discussion/#comment-2200857935'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='63'/>

tomicapretto, line 1: <p>Is it a bug or a feature that no dims are seen for <code>beta</code>?  <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/7820/discussion/#comment-2200857938'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='65'/>

ricardoV94: <p>Oh sorry, that should be <code>tensor_from_xtensor</code></p>

ricardoV94: <p>No, but yes that was a mistake, it should both have been <code>.basic_RVs</code></p>

ricardoV94: <p>No, the <code><code>dim</code></code> you're seeing is the one for <code>concat</code> </p>

ricardoV94: <p>It's because I didn't use <code>pm.dims.Deterministic, but the regular pm.Deterministic` .Good catch </code></p>

docs/source/learn/core_notebooks/dims_module.ipynb
ricardoV94: I don't think we can make this a non-issue, I didn't know xarray Dataset requires coords all the time back then. The most clean solution I see is https://github.com/pymc-devs/pymc/issues/7852 or to put coords in the graph (hard)

docs/source/learn/core_notebooks/dims_module.ipynb
tomicapretto: :')

williambdean:
Pasting the documentation link again
https://pymcio--7820.org.readthedocs.build/projects/docs/en/7820/api.html

The new notebook doesn't show up:
https://pymcio--7820.org.readthedocs.build/projects/docs/en/7820/learn/core_notebooks/index.html

williambdean:
How about adding here too: https://pymcio--7820.org.readthedocs.build/projects/docs/en/7820/api.html

ricardoV94:
Thanks @williambdean it should now be listed. I also tried to add some stuff in the API but it's likely broken. Help appreciated @OriolAbril (last commit)

williambdean:
The perfect PR doesnt exi...

docs/source/api/dims.rst
```diff
@@ -0,0 +1,18 @@
+.. _api_dims:
+
+Dims
+====
+
+This submodule contains functions for defining distributions and operations that use explicit dimensions.
+
+The module is presented in :doc:`dims_module`.
```
OriolAbril: should be a `ref` type cross-reference: `` :ref:`dims_module` ``

docs/source/api/dims.rst
```diff
@@ -0,0 +1,18 @@
...
+
+.. currentmodule:: pymc.dims
+
+.. autosummary::
+   :toctree: generated/
```
OriolAbril: this should be a toctree directive only as (for now) it is listing other doc pages, not yet pymc objects

docs/source/api/dims.rst
OriolAbril, line 1: Current preview of this page: https://pymcio--7820.org.readthedocs.build/projects/docs/en/7820/api/dims.html. I had to enter the url manually because it is not part of the toctree, it should be added to https://github.com/pymc-devs/pymc/blob/main/docs/source/api.rst?plain=1#L7-L25 (adding it here will also make the page have a sidebar like the other api pages)

docs/source/api/dims.rst
```diff
@@ -0,0 +1,18 @@
+.. _api_dims:
+
+Dims
+====
+
+This submodule contains functions for defining distributions and operations that use explicit dimensions.
```
OriolAbril, line 8: I would probably add a warning about the module here too

pymc/dims/distributions/scalar.py
```diff
@@ -0,0 +1,191 @@
...
+
+    return ptx.math.reciprocal(ptx.math.sqrt(tau))
+
+
+class Flat(DimDistribution):
+    xrv_op = pxr.as_xrv(flat)
+
+    @classmethod
+    def dist(cls, **kwargs):
+        return super().dist([], **kwargs)
```
OriolAbril, line 47: These currently end up having an empty docstring: https://pymcio--7820.org.readthedocs.build/projects/docs/en/7820/api/dims/generated/pymc.dims.Flat.html. Is it possible to dynamically copy the docstring from the "regular" distribution?

For vector distributions or transforms we already have specific docstrings (and I think we want that there) but scalar ones can probably share the docstring after a search/replace of `tensor_like` -> `xtensor_like`

docs/source/api/dims.rst
ricardoV94: Can you propose the changes with github comment thing? I'm not sure if I should take autosummary or something else?

OriolAbril: I tried to and didn't work, now it does allow me:

```suggestion
.. toctree::
```

pymc/dims/distributions/scalar.py
ricardoV94: I would leave this to a follow up PR

docs/source/api/dims.rst
ricardoV94: sure

pymc/dims/distributions/scalar.py
ricardoV94: https://github.com/pymc-devs/pymc/issues/7861

OriolAbril (approved):
not sure if my approval is enough but let's get the ball rolling

maresb:
Hey, how bad is it to use this PR with PyTensor v2.31.2?

The [feedstock](https://github.com/conda-forge/pymc-feedstock/pull/148) wasn't updated so probably I have to mark that build as broken. :sob:

ricardoV94:
> Hey, how bad is it to use this PR with PyTensor v2.31.2?

Won't work with it

maresb:
Ok, I have a proposed fix in https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/1056

In the meantime be aware for potentially broken environments from users who upgrade only PyMC.

maresb:
Indeed:

```bash
$ conda install -c conda-forge pymc "pytensor==2.31.2"
$ python -c "import pymc"
  File "<string>", line 1, in <module>
    import pymc
  File "/opt/conda/lib/python3.13/site-packages/pymc/__init__.py", line 50, in <module>
    from pymc import _version, gp, ode, sampling
  File "/opt/conda/lib/python3.13/site-packages/pymc/gp/__init__.py", line 17, in <module>
    from pymc.gp import cov, mean, util
  File "/opt/conda/lib/python3.13/site-packages/pymc/gp/cov.py", line 52, in <module>
    from pymc.pytensorf import constant_fold
  File "/opt/conda/lib/python3.13/site-packages/pymc/pytensorf.py", line 44, in <module>
    from pytensor.tensor.random.op import RandomVariable, RNGConsumerOp
ImportError: cannot import name 'RNGConsumerOp' from 'pytensor.tensor.random.op' (/opt/conda/lib/python3.13/site-packages/pytensor/tensor/random/op.py)
```

PR #7836, opened 2025-06-26
PyMC implementation of WALNUTS sampler

fonnesbeck:
## Description

First pass at implementing WALNUTS, partly as a learning opportunity. It runs with `pm.sample` but does not yet produce a good sample. Still lots of work to do.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7836.org.readthedocs.build/en/7836/

### DIFF
--- a/pymc/step_methods/__init__.py
+++ b/pymc/step_methods/__init__.py
@@ -15,7 +15,7 @@
 """Step methods."""
 
 from pymc.step_methods.compound import BlockedStep, CompoundStep
-from pymc.step_methods.hmc import NUTS, HamiltonianMC
+from pymc.step_methods.hmc import NUTS, WALNUTS, HamiltonianMC
 from pymc.step_methods.metropolis import (
     BinaryGibbsMetropolis,
     BinaryMetropolis,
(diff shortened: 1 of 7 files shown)


twiecki:
Why not hook up to the walnuts implrementation Bob et al provide?

fonnesbeck:
This borrows directly and heavily from Bob's repo. Its really not a lot of code.

twiecki:
> This borrows directly and heavily from Bob's repo. Its really not a lot of code.

Sure, but they'll keep iterating on it so we'd always need to keep up with changes. Anyway, I'm not against it, we can always revert back to their implementation if we find it's a great sampler. Or pymc-extras might also be a place.

fonnesbeck:
Yeah, this should probably be in `pymc-extras`. I will move it.

I think that if it turns out to be a performant algorithm we will want our own implementation of it (doesn't have to be this one, of course).

fonnesbeck:
Moved to https://github.com/pymc-devs/pymc-extras/pull/530

PR #7841, opened 2025-07-03
Add `vectorize_over_posterior` to `pymc.sampling.forward`

lucianopaz:
This PR adds the function `pymc.sampling.forward.vectorize_over_posterior`.

This function basically vectorizes the computation of a list of output variables over possible values of random variables or deterministics that are stored in an `xarray.Dataset`. Any extra random variable remaining in the graph will be vectorized as well, either by changing its size explicitly to match the batch of `chain * draw` dimensions, or implicitly given its inputs.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7841.org.readthedocs.build/en/7841/

### DIFF
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -28,9 +28,11 @@ from typing import (
 
 import numpy as np
 import xarray
+import xarray as xr
 
 from arviz import InferenceData
 from pytensor import tensor as pt
+from pytensor.graph import vectorize_graph
 from pytensor.graph.basic import (
     Apply,
     Constant,
@@ -42,7 +44,7 @@ from pytensor.graph.basic import (
 from pytensor.graph.fg import FunctionGraph
 from pytensor.tensor.random.var import RandomGeneratorSharedVariable
 from pytensor.tensor.sharedvar import SharedVariable, TensorSharedVariable
-from pytensor.tensor.variable import TensorConstant
+from pytensor.tensor.variable import TensorConstant, TensorVariable
 from rich.console import Console
 from rich.progress import BarColumn, TextColumn, TimeElapsedColumn, TimeRemainingColumn
 from rich.theme import Theme
@@ -52,6 +54,8 @@ import pymc as pm
 from pymc.backends.arviz import _DefaultTrace, dataset_to_point_list
 from pymc.backends.base import MultiTrace
 from pymc.blocking import PointType
+from pymc.distributions.shape_utils import change_dist_size
+from pymc.logprob.utils import rvs_in_graph
 from pymc.model import Model, modelcontext
 from pymc.pytensorf import compile
 from pymc.util import (
@@ -68,6 +72,7 @@ __all__ = (
     "draw",
     "sample_posterior_predictive",
     "sample_prior_predictive",
+    "vectorize_over_posterior",
 )
 
 ArrayLike: TypeAlias = np.ndarray | list[float]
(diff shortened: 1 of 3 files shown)


tests/sampling/test_forward.py
```diff
@@ -1801,3 +1806,114 @@ def test_sample_prior_predictive_samples_deprecated_warns() -> None:
...
+    if request.param == "rv_from_posterior":
+        if has_nested_random_variables:
+            return ["x_parent", "x"]
+        else:
+            return ["x"]
+    else:
+        return []
+
+
+def test_vectorize_over_posterior(
```
ricardoV94, line 1865: Nitpick, I really don't like fixtures used like this, makes reading the test behavior very complex. Why not just have if elses in the test function based on the parametrization. Also define the model in the test function?

ricardoV94, line 1865: I would like a test that actually tries to evaluate a vectorized function and assert something about the results. Can be a simpler graph.

pymc/sampling/forward.py
```diff
@@ -984,3 +989,99 @@ def sample_posterior_predictive(
...
+
+    # Replace placeholders with actual posterior samples
+    nsamples = len(posterior.coords["chain"]) * len(posterior.coords["draw"])
+    replace_dict: dict[Variable, Variable] = {}
+    for rv in needed_rvs:
+        posterior_samples = posterior[rv.name].data
+        shape = posterior_samples.shape
+
+        replace_dict[rv] = pt.constant(
+            posterior_samples.reshape((nsamples, *shape[2:])).astype(rv.dtype),
```
ricardoV94: Suggestion, don't flatten the sample dimensions. I would make this function take an optional argument `sample_dims` that defaults to `("chain", "draw")`. If a user wants the batch dimensions flatten they can pass `posterior=dataset.stack(sample=("chain", "draw")), sample_dims=("sample",))`? Vectorize has no problem with multiple batch dimensions

tests/sampling/test_forward.py
```diff
@@ -1801,3 +1806,114 @@ def test_sample_prior_predictive_samples_deprecated_warns() -> None:
...
+        d = pm.Data("d", np.array([1, 2, 3]))
+        obs = pm.Data("obs", np.ones_like(d.get_value()))
+        y = pm.Deterministic("y", x * d)
+        z = pm.Gamma("z", mu=pt.exp(y), sigma=pt.exp(y) * 0.1, observed=obs)
+        pm.Deterministic("z_downstream", z * 2)
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore")
+        with model:
+            idata = pm.sample(100, tune=100, chains=1, cores=1)
```
ricardoV94: Use prior_predictive and pretend it is a posterior?

ricardoV94 (commented):
Looks good, some minor suggestions and complaints. This PR reminds me that I don't think we have `vectorize_node` implemented for SymbolicRandomVariables. Shoud be easy to implement, at least for those we have a signature (it's how we know how to resize them automatically)

https://github.com/pymc-devs/pymc/blob/3ae509505f830c990b4f67cc08d37cd42d538d65/pymc/distributions/distribution.py#L395-L414

ricardoV94:
> Looks good, some minor suggestions and complaints. This PR reminds me that I don't think we have `vectorize_node` implemented for SymbolicRandomVariables. Shoud be easy to implement, at least for those we have a signature (it's how we know how to resize them automatically)

I'm not suggesting we should do it in this PR, just reminded me

tests/sampling/test_forward.py
lucianopaz: Agree to disagree here. Having the fixture makes it very easy to add a new orthogonal condition that then gets automatically combined with the existing ones. I understand that the test might be harder to follow, but it makes it easier to write extend the test coverage to new scenarios while ensuring that everything else still works fine.

ricardoV94: I don't see how you can extend and still not have the source of truth (whatever you want to assert) on the test itself, which means you still need to know what all the parametrization and their combinations do. You just make it harder to read because information is spread all over the place.

To be clear I'm a big fan of parametrize, just no via fixtures. But it's personal preference.

lucianopaz: I'm not sure if the test I added is satisfying. We can't ensure that the samples from `sample_posterior_predictive` will be exactly the same as the samples drawn from a vectorized graph because of the order in which each rng is called to generate the samples is different.

tests/sampling/test_forward.py
```diff
@@ -1801,3 +1806,144 @@ def test_sample_prior_predictive_samples_deprecated_warns() -> None:
...
+def test_vectorize_over_posterior_matches_sample():
+    rng = np.random.default_rng(1234)
+    with pm.Model() as model:
+        x = pm.Normal("x")
+        sigma = 0.1
+        obs = pm.Normal("obs", x, sigma, observed=rng.normal(size=10))
+        det = pm.Deterministic("det", obs + 1)
+
+    with model:
+        idata = pm.sample_prior_predictive(10000)
```
ricardoV94: what about constructing a simple fake posterior with x of your liking? Maybe something centered around -10 in chain 1, and 10 in chain 2? They can all have the same values, so the expectation is simpler

pymc/sampling/forward.py
```diff
@@ -68,6 +72,7 @@
     "draw",
     "sample_posterior_predictive",
     "sample_prior_predictive",
+    "vectorize_over_posterior",
```
ricardoV94, line 75: can you check if it shows anywhere in the docs?

tests/sampling/test_forward.py
ricardoV94: I think a statistical check like you did is fine. I suggested a form that should be stronger to false positives

lucianopaz: Done

pymc/sampling/forward.py
lucianopaz: It didn't show. Now I've added it. The sphinx-lint pre-commit hook was failing for me locally. I was a bit surprised because it said that most of the doc fails had problems. I tried to push without verification to see if pre-commit complains online.

tests/sampling/test_forward.py
lucianopaz: I understand. I personally like it with fixtures.

ricardoV94 (approved):

ricardoV94:
Looks great, failing test is unrelated I've seen it fail sporadically

PR #7844, opened 2025-07-03
Fix issues with model graph 

### LINKED ISSUE #7397: BUG: `make_compute_graph` creates self loop
I did just catch this bug: It comes from the `make_compute_graph` which causes a self loop

```python
from pymc.model_graph import ModelGraph

coords = {
    "obs": range(5),
}
    data = pt.as_tensor_variable(
        np.ones((5, 3)),
        name="C",
    )
    pm.Deterministic("C", data, dims=("obs", None))

error_compute_graph = ModelGraph(model).make_compute_graph() # defaultdict(set, {"C": {"C"}})
# Visualize error:
pm.model_to_graphviz(model)
```
Result: 

compute-graph-bug

_Originally posted by @wd60622 in https://github.com/pymc-devs/pymc/issues/7392#issuecomment-2196586849_

ricardoV94:
Closes https://github.com/pymc-devs/pymc/issues/7397

Also when working on #7820 I noticed our detection of observed variables was based on old behavior where only Cast operations were allowed in the graph. This PR fixes that as well

----
📚 Documentation preview 📚: https://pymc--7844.org.readthedocs.build/en/7844/

### DIFF
--- a/tests/test_model_graph.py
+++ b/tests/test_model_graph.py
@@ -541,6 +541,16 @@ def test_model_graph_with_intermediate_named_variables():
     assert ModelGraph(m3).make_compute_graph() == {"C": set(), "D": set(), "E": {"C"}}
 
 
+def test_model_graph_complex_observed_dependency():
+    with pm.Model() as model:
+        x = pm.Data("x", [0])
+        y = pm.Data("y", [0])
+        observed = pt.exp(x) + pt.log(y)
+        pm.Normal("obs", mu=0, observed=observed)
+
+    assert ModelGraph(model).make_compute_graph() == {"obs": set(), "x": {"obs"}, "y": {"obs"}}
+
+
 @pytest.fixture
 def simple_model() -> pm.Model:
     with pm.Model() as model:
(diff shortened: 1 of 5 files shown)


ricardoV94:
@williambdean do you want to review?

tests/test_model_graph.py
```diff
@@ -470,7 +470,7 @@ class TestVariableSelection:
         [
             (["c"], ["a", "b", "c"], {"c": {"a", "b"}, "a": set(), "b": set()}),
             (
-                ["L"],
+                ["L", "obs"],
```
ricardoV94: You need to specify `obs` if you want it in the graph now. I don't see a reason to treat it differently from other Data variables?

jessegrabowski: Agreed that it should always show by default. Very weird behavior otherwise?

jessegrabowski (approved):
lgtm

tests/test_model_graph.py
ricardoV94: It's not that, when you specify a subset of variables it would show the observed data variables if you selected the observed variable,  but not say X data variables.

Now if you specify a subset it won't show it unless you specify it as well

jessegrabowski: Oh, sure. If it's for user-specified subsets, you should get back what you ask for and nothing else.

ricardoV94: actually I think it should behave as before now .. should recheck it.

williambdean (approved):
Thank you, you OSS legend

tests/test_model_graph.py
ricardoV94: Yeah it works as before, reverted the test changes

PR #7848, opened 2025-07-06
DOC: Update sponsors list

### LINKED ISSUE #7846: DOC: Update sponsors list

The sponsors list here is outdated.
Should also have a section for "Past Sponsors".

https://github.com/pymc-devs/pymc/blob/main/README.rst

### References
References https://github.com/pymc-devs/pymc.io/pull/107#issue-3201174763

_No response_

cc: @purna135

purna135:
## Description
Updated the README to include a new sponsor logo and changed the layout of the sponsor section to a 1x3 grid view using reStructuredText tables for better visual presentation.

## Related Issue

- [x] Closes #7846
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7848.org.readthedocs.build/en/7848/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -281,13 +281,24 @@ You can get professional consulting support from `PyMC Labs <https://www.pymc-la
 Sponsors
 ========
 
-|NumFOCUS|
+.. list-table::
+   :widths: 33 33 33
+   :align: center
 
-|PyMCLabs|
+   * - |NumFOCUS|
+     - |PyMCLabs|
+     - |OWR|
 
-|Mistplay|
+Past Sponsors
+=============
 
-|ODSC|
+.. list-table::
+   :widths: 33 33 33
+   :align: center
+
+   * - |Mistplay|
+     - |ODSC|
+     - |AdiaLab|
 
 Thanks to our contributors
 ==========================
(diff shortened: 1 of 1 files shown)


reshamas (approved):

OriolAbril (commented):
I haven't been following this sorry. I see this is doing the same that was done on the website so the comment is way late. I hardly see any difference between treatment of current sponsors/institutional partners and past sponsors and I think current sponsors should be much more prominently showcased. Otherwise I see little to no incentive for sponsors to continue being sponsors if when they stop sponsoring everything so close to when they are active sponsors.

I think it would also be nice to mention the sponsorship relation. Most of these are sponsors of pymcon webseries, which also had multiple tiers, and in general I think project sponsors or even better institutional partners are preferred (but harder to get)

While I go on about complaining and making requests. We also had sponsors from the first PyMCon edition: https://pymc-devs.github.io/pymcon/#sponsors and project sponsors from pymc3 times: https://github.com/pymc-devs/pymc/tree/c1e272c6454a1766523ee5ab5b7083fb100a1002?tab=readme-ov-file#sponsors (images don't seem to work anymore but the list had numfoucs, quantopian and odsc)

---

All being said, none of this are blockers the syntax is correct and renders as intended. So I will defer to the steering council to merge this as is, decide how past sponsorship should be recognized and request changes or open a project discussion (if we already did and I missed it just merge and ignore my comments).

purna135:
What do you think, @reshamas? 
Should we show just the names of past sponsors instead of the logos? like arviz website: https://www.arviz.org/en/latest/sponsors_partners.html#historical-funding

or a detailed list with logo like Scikit-learn keeps: https://scikit-learn.org/stable/about.html#funding

OriolAbril:
In my opinion the key difference I am commenting on is not the style of the past sponsors list but its placement.

Note that the two pages you linked for both scikit-learn and arviz are not their homepage but a different page with info about the project and/or its sponsors only. Where the relation of each sponsor with the project is explained in more detail.

Both projects do show active sponsors on their homepage (right at the bottom, and only as logos but I think the logo or card thing is secondary) https://scikit-learn.org/stable/ and https://www.arviz.org/en/latest/ but no trace of past sponsors in either homepage

On the other hand the pymc homepage has much more detail as well as past sponsors https://www.pymc.io/, and the readme (arguably a landing page for developers) would too with this PR.

PR #7855, opened 2025-07-17
Replace VarName with built-in str

### LINKED ISSUE #7843: Remove `VarName` from codebase
### Description

Why are we defining a new string type? An alias is sufficient if we even need it, but I think `str` is just fine. Also all PyTensor variables have a `name` attribute we don't need to call `getattr(variable, "name")` it may just be None

https://github.com/pymc-devs/pymc/blob/f6bfdfd0cbbc371259536e47092ff83b0fb6f742/pymc/util.py#L61

https://github.com/pymc-devs/pymc/blob/f6bfdfd0cbbc371259536e47092ff83b0fb6f742/pymc/util.py#L253-L255

osyuksel:
## Description
This change:

* Replaces VarName type with built-in `str`
* Updates `get_var_name` to have it work without getattr

## Related Issue

- [x] Closes #7843
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7855.org.readthedocs.build/en/7855/

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -67,7 +67,6 @@ from pymc.pytensorf import (
 )
 from pymc.util import (
     UNSET,
-    VarName,
     WithMemoization,
     _UnsetType,
     get_transformed_name,
@@ -1968,7 +1967,7 @@ class Model(WithMemoization, metaclass=ContextMeta):
     def to_graphviz(
         self,
         *,
-        var_names: Iterable[VarName] | None = None,
+        var_names: Iterable[str] | None = None,
         formatting: str = "plain",
         save: str | None = None,
         figsize: tuple[int, int] | None = None,
@@ -2172,7 +2171,7 @@ def compile_fn(
     )
 
 
-def Point(*args, filter_model_vars=False, **kwargs) -> dict[VarName, np.ndarray]:
+def Point(*args, filter_model_vars=False, **kwargs) -> dict[str, np.ndarray]:
     """Build a point.
 
     Uses same args as dict() does.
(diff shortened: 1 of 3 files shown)


osyuksel:
I also attempted to address the second point re. `getattr(variable, "name")` in the [issue](https://github.com/pymc-devs/pymc/issues/7843) with this [commit](https://github.com/pymc-devs/pymc/pull/7855/commits/981e20042892a7b54c43c469f98e10cf3c10aa40), however, making an explicit attribute reference (i.e. calling var.name instead of getattr) causes the tests to fail, so I assume there must be some variables where the attribute is not present.



```
==================================== ERRORS ====================================
__________________ ERROR collecting tests/test_model_graph.py __________________
tests/test_model_graph.py:467: in <module>
    class TestVariableSelection:
tests/test_model_graph.py:499: in TestVariableSelection
    ModelGraph(model_with_different_descendants()).vars_to_plot(),
pymc/model_graph.py:238: in __init__
    self._all_var_names = get_default_varnames(self.model.named_vars, include_transformed=False)
pymc/util.py:212: in get_default_varnames
    return [var for var in var_iterator if not is_transformed_name(get_var_name(var))]
pymc/util.py:217: in get_var_name
    return str(var.name if var.name is not None else var)
E   AttributeError: 'str' object has no attribute 'name'
```

ricardoV94:
That should check explicitly if the input is a string, instead of trying to call var.name on it

osyuksel:
Makes sense, missed that var could be str. Updated PR.

ricardoV94 (approved):

ricardoV94:
This got forgotten but thanks @osyuksel

PR #7856, opened 2025-07-17
Do not fail with zero-sized arrays in `dataset_to_point_list`

ricardoV94:
Numpy does not support reshape(-1, ...) when size is zero

----
📚 Documentation preview 📚: https://pymc--7856.org.readthedocs.build/en/7856/

### DIFF
--- a/pymc/backends/arviz.py
+++ b/pymc/backends/arviz.py
@@ -618,11 +618,13 @@ def dataset_to_point_list(
     for vn in var_names:
         if not isinstance(vn, str):
             raise ValueError(f"Variable names must be str, but dataset key {vn} is a {type(vn)}.")
+
     num_sample_dims = len(sample_dims)
     stacked_dims = {dim_name: ds[var_names[0]][dim_name] for dim_name in sample_dims}
     transposed_dict = {vn: da.transpose(*sample_dims, ...) for vn, da in ds.items()}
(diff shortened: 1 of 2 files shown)


pymc/backends/arviz.py
```diff
@@ -618,11 +618,13 @@ def dataset_to_point_list(
     for vn in var_names:
         if not isinstance(vn, str):
             raise ValueError(f"Variable names must be str, but dataset key {vn} is a {type(vn)}.")
+
     num_sample_dims = len(sample_dims)
     stacked_dims = {dim_name: ds[var_names[0]][dim_name] for dim_name in sample_dims}
     transposed_dict = {vn: da.transpose(*sample_dims, ...) for vn, da in ds.items()}
+    stacked_size = np.prod(transposed_dict[var_names[0]].shape[:num_sample_dims], dtype=int)
```
OriolAbril, line 625: should be the same as 

```
stacked_size = np.prod([ds.sizes[dim] for dim in sample_dims], dtype=int)
```

feel free to choose whichever you prefer

OriolAbril (approved):

pymc/backends/arviz.py
ricardoV94: ds may be a dict

PR #7870, opened 2025-07-24
Show more digits of step size in progress_bar

ricardoV94:
The two digits was copied from nutpie according to @jessegrabowski. I noticed that we always have a 0.0x, so it's not very useful. @aseyboldt explained that something something pymc nuts considers small eigenvalues of the matrix, while nutpie ignores it, which means our step-sizes tend to be smaller. 

This PR adds one extra digit for me to look at.

----
📚 Documentation preview 📚: https://pymc--7870.org.readthedocs.build/en/7870/


jessegrabowski (approved):

jessegrabowski:
We might want to consider increasing the width of the Rich console, so we have a bit more real estate to show all these digits.

ricardoV94:
The width isn't dynamic?

PR #7875, opened 2025-07-27
Derive logprob for Split operation

ricardoV94:
----
📚 Documentation preview 📚: https://pymc--7875.org.readthedocs.build/en/7875/

### DIFF
--- a/pymc/logprob/tensor.py
+++ b/pymc/logprob/tensor.py
@@ -33,20 +33,23 @@
 #   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 #   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 #   SOFTWARE.
+import typing
 
 from pathlib import Path
 
 from pytensor import tensor as pt
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.rewriting.basic import node_rewriter
+from pytensor.npy_2_compat import normalize_axis_index
 from pytensor.tensor import TensorVariable
-from pytensor.tensor.basic import Join, MakeVector
+from pytensor.tensor.basic import Join, MakeVector, Split
 from pytensor.tensor.elemwise import DimShuffle, Elemwise
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.rewriting import (
     local_dimshuffle_rv_lift,
 )
 
+from pymc.exceptions import NotConstantValueError
 from pymc.logprob.abstract import (
     MeasurableOp,
     ValuedRV,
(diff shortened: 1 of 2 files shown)


ricardoV94:
Classical mypy, feel free to review ignoring that, I'll fix it

pymc/logprob/tensor.py
```diff
@@ -183,6 +186,64 @@ def find_measurable_stacks(fgraph, node) -> list[TensorVariable] | None:
...
+            constant_axis = constant_fold([axis])
+        except NotConstantValueError:
+            raise NotImplementedError("Cannot split multivariate logp with non-constant axis")
+
+        constant_axis = normalize_axis_index(constant_axis, join_value.ndim)  # type: ignore[arg-type, assignment]
+        if constant_axis >= join_logp.ndim:
+            # If the axis is over a dimension that was reduced in the logp (multivariate logp),
+            # We cannot split it into distinct entries. The mapping between values-densities breaks.
+            # We return the weighted logp by the split sizes. This is a good solution as any?
+            split_weights = splits / pt.sum(splits)
```
jessegrabowski, line 232: Is this legit?

tests/logprob/test_tensor.py
```diff
@@ -372,3 +373,155 @@ def test_unmeargeable_dimshuffles():
...
+        # (3, 5) Dirichlet
+        rng = np.random.default_rng(426)
+        rng_pt = random_generator_type("rng")
+        alpha = np.linspace(1, 10, 5) * np.array([1, 10, 100])[:, None]
+        x = pt.random.dirichlet(alpha, rng=rng_pt)
+
+        # axis=-2 (i.e., 0, - batch dimension)
+        x_parts = pt.split(x, splits_size=[2, 1], n_splits=2, axis=-2)
+        x_parts_vv = [x_part.clone() for x_part in x_parts]
+        logp_parts = list(conditional_logp(dict(zip(x_parts, x_parts_vv))).values())
```
jessegrabowski, line 439: Do i understand this correctly that each part is conditioned on the values of all other parts?

Thinking about e.g. the MVN case, where if you split the vector and condition each split on the other, you get two new MVN distributions

pymc/logprob/tensor.py
ricardoV94: I think so? In MarginalMixture we decided to set the whole logp on the first entry, and zero for others, I like this approach more

tests/logprob/test_tensor.py
ricardoV94: There's no marginalization going on, you can't evaluate the logp of only one part without providing the remaining ones. The only thing we do is join the value, get the logp, and split it again. We could argue that we don't want to do this for multivariate variables split along the core dimension, since there's no way you can split the logp (I did the weighing, but we can revert and raise NotImplemented)

jessegrabowski (approved):
We should merge this

PR #7877, opened 2025-07-27
Fix bug in mixture logprob inference with `None` indices

### LINKED ISSUE #7762: BUG: `AdvancedSubTensor` with `None` and integer indices raises a `logprob` error instead of silently failing

asifzubair:
## Description

Previously, calculating the log-probability (`logp`) of an `AdvancedSubTensor` that included a `None` index would cause the system to crash with an unhandled internal `AttributeError`.

This change resolves the bug by adding a guard within the `logp` rewriting logic in `mixture.py`. Instead of crashing, the system now correctly identifies the ambiguous operation and raises a controlled `RuntimeError`.

A new test, `test_advanced_subtensor_none_and_integer`, has been added to `test_mixture.py` to validate this fix and prevent regressions.

Reproducible example from the ticket now executes:

```python
>>> import numpy as np
... import pymc as pm
... 
... 
... obs = np.random.default_rng().normal(size=(7, 4))
... with pm.Model():
...    inds = np.arange(obs.shape[1])
...    a = pm.Normal("a", shape=10)
...    b = pm.Deterministic("b", a[None, inds])
...    c = pm.Normal("c", mu=b, sigma=1, observed=obs)
...    pm.sample()
...    
Initializing NUTS using jitter+adapt_diag...
Multiprocess sampling (3 chains in 3 jobs)
NUTS: [a]
                                                                                                                                  
  Progress                                   Draws   Divergences   Step size   Grad evals   Sampling Speed   Elapsed   Remaining  
 ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── 
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   2000    0             0.871       3            792.47 draws/s   0:00:02   0:00:00    
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   2000    0             0.819       3            807.38 draws/s   0:00:02   0:00:00    
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   2000    0             0.951       3            783.45 draws/s   0:00:02   0:00:00    
                                                                                                                                  
Sampling 3 chains for 1_000 tune and 1_000 draw iterations (3_000 + 3_000 draws total) took 3 seconds.
We recommend running at least 4 chains for robust computation of convergence diagnostics
Inference data with groups:
	> posterior
	> sample_stats
	> observed_data
``` 

## Related Issue

- [x] Closes #7762 
- [ ] Related to #

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7877.org.readthedocs.build/en/7877/

### DIFF
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -62,7 +62,7 @@ from pytensor.tensor.subtensor import (
     is_basic_idx,
 )
 from pytensor.tensor.type import TensorType
-from pytensor.tensor.type_other import NoneConst, NoneTypeT, SliceConstant, SliceType
+from pytensor.tensor.type_other import NoneConst, NoneTypeT, SliceType
 from pytensor.tensor.variable import TensorVariable
 
 from pymc.logprob.abstract import (
@@ -289,9 +289,10 @@ def find_measurable_index_mixture(fgraph, node):
         # We don't support (non-scalar) integer array indexing as it can pick repeated values,
         # but the Mixture logprob assumes all mixture values are independent
         if any(
-            indices.dtype.startswith("int") and sum(1 - b for b in indices.type.broadcastable) > 0
+            hasattr(indices, "dtype")
+            and indices.dtype.startswith("int")
+            and sum(1 - b for b in indices.type.broadcastable) > 0
             for indices in mixing_indices
-            if not isinstance(indices, SliceConstant)
         ):
             return None
 
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -289,7 +289,7 @@ def find_measurable_index_mixture(fgraph, node):
         # We don't support (non-scalar) integer array indexing as it can pick repeated values,
         # but the Mixture logprob assumes all mixture values are independent
         if any(
-            hasattr(indices, "dtype")
+            isinstance(indices, TensorVariable)
             and indices.dtype.startswith("int")
             and not all(indices.type.broadcastable)
             for indices in mixing_indices
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -291,7 +291,7 @@ def find_measurable_index_mixture(fgraph, node):
         if any(
             hasattr(indices, "dtype")
             and indices.dtype.startswith("int")
-            and sum(1 - b for b in indices.type.broadcastable) > 0
+            and not all(indices.type.broadcastable)
             for indices in mixing_indices
         ):
             return None
--- a/tests/logprob/test_mixture.py
+++ b/tests/logprob/test_mixture.py
@@ -1178,5 +1178,8 @@ def test_advanced_subtensor_none_and_integer():
     a_val = a.type()
     a_val.name = "a_val"
 
-    with pytest.raises(RuntimeError, match="logprob terms of the following value variables could not be derived: {b_val}"):
+    with pytest.raises(
+        RuntimeError,
+        match="logprob terms of the following value variables could not be derived: {b_val}",
+    ):
         conditional_logp({b: b_val, a: a_val})
--- a/tests/logprob/test_mixture.py
+++ b/tests/logprob/test_mixture.py
@@ -1178,9 +1178,5 @@ def test_advanced_subtensor_none_and_integer():
     a_val = a.type()
     a_val.name = "a_val"
 
-    with pytest.raises(RuntimeError) as e:
+    with pytest.raises(RuntimeError, match="logprob terms of the following value variables could not be derived: {b_val}"):
         conditional_logp({b: b_val, a: a_val})
-
-    # Assert that the error message does NOT contain "AttributeError",
-    # which would indicate the presence of the original bug.
-    assert "AttributeError" not in str(e.value)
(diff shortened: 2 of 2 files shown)


pymc/logprob/mixture.py
```diff
@@ -289,9 +289,10 @@ def find_measurable_index_mixture(fgraph, node):
         # We don't support (non-scalar) integer array indexing as it can pick repeated values,
         # but the Mixture logprob assumes all mixture values are independent
         if any(
-            indices.dtype.startswith("int") and sum(1 - b for b in indices.type.broadcastable) > 0
+            hasattr(indices, "dtype")
```
ricardoV94: You can check `isinstance(indices, TensorVariable)` and then it's guaranteed to have a dtype and a broadcastable

pymc/logprob/mixture.py
```diff
@@ -289,9 +289,10 @@ def find_measurable_index_mixture(fgraph, node):
...
+            and indices.dtype.startswith("int")
+            and sum(1 - b for b in indices.type.broadcastable) > 0
```
ricardoV94: While we are here, a bit more readable:
```suggestion
            and not all(indices.type.broadcastable)
```

asifzubair:
Hi @ricardoV94 , Thank you for your comments. I've addressed them now

~However, I did a rebase and push (as recommended) but it seems the rebase has added new commits on my PR ? Would you have a suggestion on how to handle this ? I can start a new PR in the worst case.~ Thank you 🙏 

EDIT: I had to troubleshoot a bit, but now the PR looks to be in good state. Sorry for the alarm. :)

tests/logprob/test_mixture.py
```diff
@@ -1156,3 +1156,31 @@ def test_nested_ifelse():
...
+    a = pt.random.normal(0, 1, size=(10,), name="a")
+    inds = np.array([0, 1, 2, 3], dtype="int32")
+    b = a[None, inds]
+
+    b_val = b.type()
+    b_val.name = "b_val"
+    a_val = a.type()
+    a_val.name = "a_val"
+
+    with pytest.raises(RuntimeError) as e:
```
ricardoV94: Match on the error message to be more robust as a test

```suggestion
    with pytest.raises(RuntimeError, match=) as e:
```

tests/logprob/test_mixture.py
```diff
@@ -1156,3 +1156,31 @@ def test_nested_ifelse():
...
+        conditional_logp({b: b_val, a: a_val})
+
+    # Assert that the error message does NOT contain "AttributeError",
+    # which would indicate the presence of the original bug.
+    assert "AttributeError" not in str(e.value)
```
ricardoV94: I don't think we need this, if there was an AttributeError it would have raised, pytest only catches the specific error, even more with a match

ricardoV94 (commented):
Looks good just a small cleanup suggestion on the test

ricardoV94 (approved):

ricardoV94:
Thanks @asifzubair

PR #7883, opened 2025-07-31
Add marimo progress bar for pymc sampler

williambdean:
## Description

The rich progress bars don't work well in marimo notebooks. Here is a basic implementation using their progress bar component to 
track the `pymc` sampler. 

An implementation note: tasks being attached to the progress bar and progress bar manager is a bit confusing, but I tried to keep as lean as possible to get to work.

https://github.com/user-attachments/assets/04607b16-70a0-4e1b-9692-3bda981934ab

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7883.org.readthedocs.build/en/7883/

### DIFF
--- a/pymc/progress_bar.py
+++ b/pymc/progress_bar.py
@@ -12,7 +12,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Literal, Protocol, cast
 
 from rich.box import SIMPLE_HEAD
 from rich.console import Console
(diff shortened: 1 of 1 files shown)


ricardoV94:
Don't love this. Is there any plan for rich to work with marimo?

williambdean:
> Don't love this. Is there any plan for rich to work with marimo?

rich shut it down
https://github.com/Textualize/rich/issues/3750

ricardoV94:
Also we use `CustomProgress` directly in other places (.e.g, pm.compute_log_likelihood, sample_posterior_predictive)

pymc/progress_bar.py
```diff
@@ -192,6 +192,143 @@ def callbacks(self, task: "Task"):
...
+    if speed > 1 or speed == 0:
+        unit = "draws/s"
+    else:
+        unit = "s/draws"
+        speed = 1 / speed
+
+    return speed, unit
+
+
+def create_rich_progress_bar(full_stats, step_columns, progressbar, progressbar_theme):
```
ricardoV94, line 236: Any reason why these aren't methods anymore? It's not like they are used outside of ProgressBarManager (compute_draw_speed was static method anyway)

williambdean:
> Also we use `CustomProgress` directly in other places (.e.g, pm.compute_log_likelihood, sample_posterior_predictive)

CustomProgress isn't changed and would still work. The ProgressBarManager now excepts other ProgressBar protocols

pymc/progress_bar.py
```diff
@@ -192,6 +192,143 @@ def callbacks(self, task: "Task"):
...
+        self.divergences[task_id.chain_idx] = kwargs.get("divergences", 0)
+
+        total_divergences = sum(self.divergences.values())
+
+        update_kwargs = {}
+        if total_divergences > 0:
+            word = "draws" if total_divergences > 1 else "draw"
+            update_kwargs["subtitle"] = f"{total_divergences} diverging {word}"
+
+        self.bar.progress.update(**update_kwargs)
```
ricardoV94, line 320: We changed the logic to not be obsessed with NUTS. 

There is now a `failing` variable that the step methods report back, and that's passed to `update` to indicate sampling is failing. Also each step decides what is a meaningful stat to report.

This reverts back to over-specializing for nuts

pymc/progress_bar.py
williambdean: ProgressBarManager seemed to be doing too much. The second one was specific to the rich implementation

ricardoV94:
> > Also we use `CustomProgress` directly in other places (.e.g, pm.compute_log_likelihood, sample_posterior_predictive)
> 
> CustomProgress isn't changed and would still work. The ProgressBarManager now excepts other ProgressBar protocols

Other methods don't use ProgressBarManager, but CustomProgress, which is a rich object

pymc/progress_bar.py
williambdean: Does failing report the number of divergences? I can remove this logic and just not have a subtitle which would be specific to NUTS

ricardoV94:
Such as https://github.com/pymc-devs/pymc/blob/011fb355c8c932143632bf7760d10ea969779e71/pymc/backends/arviz.py#L686-L688

pymc/progress_bar.py
ricardoV94: Failing is just binary, we use for coloring. The relevant summary stats are reported as part of `all_step_stats`, we don't treat them specially.

https://github.com/pymc-devs/pymc/blob/886b584d2dae258d7ae4902ddd3080b986596167/pymc/progress_bar.py#L517

williambdean:
> Such as
> 
> https://github.com/pymc-devs/pymc/blob/011fb355c8c932143632bf7760d10ea969779e71/pymc/backends/arviz.py#L686-L688

That is using rich (inherited) `CustomProgress` directly... completely coupled with `rich` and not the `ProgressBarManager` absraction. 

I was focusing on the NUTS sampler progress bar specifically but can broaden to the where `CustomProgress` is used if that's your concern. However, those just won't work in marimo as they use rich.

ricardoV94:
Furthermore, is there a way to have multiple progress_bars, color them manually? Fine if not, but keeping things similar between backends may help with code complexity

pymc/progress_bar.py
ricardoV94: This still seems like a function that would ever only be called by ProgressBarManager, the inputs are way too specific

ricardoV94: And the object that is created is also called very specifically by ProgressBarManager

williambdean:
> Furthermore, is there a way to have multiple progress_bars, color them manually? Fine if not, but keeping things similar between backends may help with code complexity

You can, but it takes but a bit of real estate so I opted for the single one to be simple. I'd think that'd be determined by the `add_tasks` method from the `ProgressBar` protocol

pymc/progress_bar.py
williambdean: ideally it would be passed in as many of the init kwargs of the ProgressBarManager are specific to this type of ProgressBar

ricardoV94: What would be passed in? I was just arguing against this function being moved out of the class, because it seems tightly coupled to it anyway

ricardoV94:
> I was focusing on the NUTS sampler progress bar specifically but can broaden to the where CustomProgress is used if that's your concern. However, those just won't work in marimo as they use rich.

Yeah and that's a valid strategy. I'm raising it in case that also solving the CustomProgress fixes more cases / provides a solution that is different than the one you took here specifically for the ProgressBarManager

ricardoV94:
@williambdean bump, people seem to be finding this issue

ericmjl (approved):
I'm happy to see this merged!

pymc/progress_bar.py
ricardoV94: I'll still nit about this floating function that won't ever be used elsewhere

pymc/progress_bar.py
```diff
@@ -275,11 +412,16 @@ def __init__(
 
         progress_columns, progress_stats = step_method._progressbar_config(chains)
 
-        self._progress = self.create_progress_bar(
-            progress_columns,
-            progressbar=progressbar,
-            progressbar_theme=progressbar_theme,
-        )
+        if in_marimo_notebook():
```
ricardoV94, line 415: cache this result, imports are a tad slow

williambdean: Thought imports were alright cached 🤔

ricardoV94: Python always has to check some stuff again (no idea what), even though it doesnt re-import modules. 
It's particularly bad when it's not installed, as it will search for it everywhere again and again.

```python
import numpy as np

def foo(x):
    return np.exp(x)

def bar(x):
    import numpy as np
    return np.exp(x)

def qux(x):
    try:
        import not_installed
    except ImportError:
        pass
    return np.exp(x)

%timeit foo(1)  # 592 ns ± 4.57 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
%timeit bar(1)  # 915 ns ± 3.59 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
%timeit qux(1)  # 57 μs ± 292 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

ricardoV94: Plus I don't know how much the `mo.running_in_marimo()` takes, but that's not my concern as a non-marimo user :)

koaning: One observation on this front, [pep 810 has been accepted](https://peps.python.org/pep-0810/), so with the advent of lazy imports being a thing in the future this may be less of a concern.

ricardoV94: Yeah but that will take a while until our oldest compatible python version is covered

fonnesbeck:
This doesn't seem to work with the `progressbar=` string arguments(?)

image

williambdean:
> This doesn't seem to work with the `progressbar=` string arguments(?)
> 
> 
> 
> image
> 
> 

Exactly

williambdean:
Marimo didnt have a stacked progress bar that I found. Thats where custom html like nutpie would be good. 

We need BYOPB 😁

fonnesbeck:
@williambdean  Have a peek at #8055 and see what you think

PR #7885, opened 2025-08-01
Add a function to create a tensor that represents the loop over the posterior

lucianopaz:
## Description
This PR adds the `loop_over_posterior` function. It complements `vectorize_over_posterior` and can be used whenever `vectorize_graph` is not implemented for some nodes in the computational graph (i.e. some kinds of `AdvancedSetSubtensor` raise `NotImplementedError` when trying to vectorize them).

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7885.org.readthedocs.build/en/7885/

### DIFF
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -42,6 +42,7 @@ from pytensor.graph.basic import (
     walk,
 )
 from pytensor.graph.fg import FunctionGraph
+from pytensor.scan.basic import scan
 from pytensor.tensor.random.var import RandomGeneratorSharedVariable
 from pytensor.tensor.sharedvar import SharedVariable, TensorSharedVariable
 from pytensor.tensor.variable import TensorConstant, TensorVariable
@@ -57,7 +58,12 @@ from pymc.blocking import PointType
 from pymc.distributions.shape_utils import change_dist_size
 from pymc.model import Model, modelcontext
 from pymc.progress_bar import CustomProgress, default_progress_theme
-from pymc.pytensorf import compile, rvs_in_graph
+from pymc.pytensorf import (
+    clone_while_sharing_some_variables,
+    collect_default_updates,
+    compile,
+    rvs_in_graph,
+)
 from pymc.util import (
     RandomState,
     _get_seeds_per_chain,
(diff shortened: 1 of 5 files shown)


ricardoV94:
I would rather work on pytensor to refactor AdvancedSubtensor. The problem is described in, https://github.com/pymc-devs/pytensor/issues/541

And/or allow `vectorize_graph` to use scan, but keep the same API

pymc/sampling/forward.py
```diff
@@ -1083,3 +1090,122 @@ def vectorize_over_posterior(
...
+        input_values = args[: len(args) - len(input_tensors) - len(independent_rvs)]
+        non_sequences = args[len(args) - len(input_tensors) - len(independent_rvs) :]
+
+        # Compute output sample value for input sample values
+        replace = {
+            **dict(zip(rvs_from_posterior, input_values, strict=True)),
+        }
+        samples = clone_while_sharing_some_variables(
+            outputs, replace=replace, kept_variables=non_sequences
+        )
```
ricardoV94, line 1169: I know VI does something like this, either by passing things that shouldn't be replaced as keys:values, or by using `graph_replace`, can you check if the same approach works?

ricardoV94: https://github.com/pymc-devs/pymc/blob/127ac30e5b41d55ab73265e1c08b71c25fac8789/pymc/variational/opvi.py#L1062C9-L1067C10

pymc/sampling/forward.py
```diff
@@ -1083,3 +1090,122 @@ def vectorize_over_posterior(
                 f"The following random variables found in the extracted graph: {remaining_rvs}"
             )
     return vectorized_outputs
+
+
+def loop_over_posterior(
```
ricardoV94, line 1095: instead of a separate function have an argument: `use_scan: bool = False` in the previous function? Most logic should be the same. 

This flatten batch -> scan -> reshape logic should go in PyTensor, it's needed in general

Issue #7894, opened 2025-08-11
Refactor: Make precision_mv_normal_logp a general-purpose function


Michal-Novomestsky:
### Description

Quality-of-life refactor to facilitate https://github.com/pymc-devs/pymc-extras/issues/532

PR #7895, opened 2025-08-11
refactor: move _precition_mv_normal_logp into a seperate function

Michal-Novomestsky:
Addresses #7894 as part of https://github.com/pymc-devs/pymc-extras/issues/532

----
📚 Documentation preview 📚: https://pymc--7895.org.readthedocs.build/en/7895/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -338,16 +338,21 @@ class PrecisionMvNormalRV(SymbolicMVNormalUsedInternally):
             method=method,
         )(rng, size, mean, tau)
 
-
-@_logprob.register
-def precision_mv_normal_logp(op: PrecisionMvNormalRV, value, rng, size, mean, tau, **kwargs):
-    [value] = value
+def _precision_mv_normal_logp(value, mean, tau):
     k = value.shape[-1].astype("floatX")
 
     delta = value - mean
     quadratic_form = delta.T @ tau @ delta
     logdet, posdef = _logdet_from_cholesky(nan_lower_cholesky(tau))
     logp = -0.5 * (k * pt.log(2 * np.pi) + quadratic_form) + logdet
+
+    return logp, posdef
+
+@_logprob.register
+def precision_mv_normal_logp(op: PrecisionMvNormalRV, value, rng, size, mean, tau, **kwargs):
+    [value] = value
+
+    logp, posdef = _precision_mv_normal_logp(value, mean, tau)
 
     return check_parameters(
         logp,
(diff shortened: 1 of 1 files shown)


ricardoV94:
Let's duplicate it in pymc-extras first to see if we really need it refactored. In theory you should be able to introduce a `PrecisionMVNormal` RV and call `logp` on it to get this implementation. Anyway it will be more obvious if we need the refactor once the dependent work on pymc-extras is ready to be merged.

Michal-Novomestsky:
@ricardoV94 Thoughts on this having seen the main PR now? Also I'm a little unsure about creating a `PrecisionalMVNormal` RV inside the `logp` function - I tried doing that with `MvNormal(tau=...)`, but then it complained that the RV already exists (I believe it happens because the `logp` gets called repeatedly and ig the garbage collector doesn't wipe the RV in time. Please correct me if I'm wrong, but here I feel that simply calling `precision_mv_normal_logp` is much more lightweight, especially given speed is a concern.

Michal-Novomestsky:
pre-commit.ci autofix

Michal-Novomestsky:
> @ricardoV94 Thoughts on this having seen the main PR now? Also I'm a little unsure about creating a `PrecisionalMVNormal` RV inside the `logp` function - I tried doing that with `MvNormal(tau=...)`, but then it complained that the RV already exists (I believe it happens because the `logp` gets called repeatedly and ig the garbage collector doesn't wipe the RV in time. Please correct me if I'm wrong, but here I feel that simply calling `precision_mv_normal_logp` is much more lightweight, especially given speed is a concern.

@ricardoV94 Just making sure you're ok with the above?

ricardoV94:
What I'm saying is that you can already get the logp function wherever it is implemented by following the logp API which is:

```python
import pymc as pm
import pytensor.tensor as pt
from pymc.distributions.multivariate import PrecisionMvNormalRV

# There's a bug here, it should be possible to do
# rv = PrecisionMvNormalRV.rv_op(mean=pt.ones(3), tau=pt.eye(3))
# but we wrote the signature wrong

_, rv = PrecisionMvNormalRV.rv_op(mean=pt.ones(3), tau=pt.eye(3)).owner.outputs
logp = pm.logp(rv, pt.zeros(3))
logp.eval()
```

This is the API to access the logp function. It should work just fine for you inside a logp expression. We do it here as well: https://github.com/pymc-devs/pymc/blob/09603234f416866c9c9e32d1954fb691bd1580cd/pymc/distributions/discrete.py#L734

ricardoV94:
In either case this PR is fine, I just want to do it after the work on the pymc-extras is ready. Until then just copy the logp function there

Michal-Novomestsky:
>In either case this PR is fine, I just want to do it after the work on the pymc-extras is ready. Until then just copy the logp function there

Sounds good - should I just raise an issue for now in that case?

ricardoV94:
> Sounds good - should I just raise an issue for now in that case?

Just add the code in the same PR where you need it

Michal-Novomestsky:
>Just add the code in the same PR where you need it

Oh yes I see, just done now.

PR #7905, opened 2025-09-06
use preliz under the hood of `find_constrained_prior`

aloctavodia:
In https://github.com/pymc-devs/pymc/pull/7680#issuecomment-2709633765 it was suggested that instead of deprecating `find_constrained_prior`, we should call PreliZ's maxent function.

----
📚 Documentation preview 📚: https://pymc--7905.org.readthedocs.build/en/7905/

### DIFF
--- a/pymc/func_utils.py
+++ b/pymc/func_utils.py
@@ -14,16 +14,6 @@
 
 import warnings
 
-from sys import modules
-
-try:
-    from preliz.unidimensional import maxent
-except ImportError as e:
-    raise ImportError(
-        "The `find_constrained_prior` function requires the `preliz` package. "
-        "You can install it via `pip install preliz` or `conda install -c conda-forge preliz`."
-    ) from e
-
 import pymc as pm
 
 __all__ = ["find_constrained_prior"]
@@ -112,6 +102,15 @@ def find_constrained_prior(
             mass=0.9,
         )
     """
+    try:
+        from preliz import distributions
+        from preliz.unidimensional import maxent
+    except ImportError as e:
+        raise ImportError(
+            "The `find_constrained_prior` function requires the `preliz` package. "
+            "You can install it via `pip install preliz` or `conda install -c conda-forge preliz`."
+        ) from e
+
     if fixed_params is None:
         fixed_params = {}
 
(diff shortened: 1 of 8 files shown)


pymc/func_utils.py
```diff
@@ -11,15 +11,18 @@
...
+import warnings
 
-import numpy as np
-import pytensor.tensor as pt
+from sys import modules
 
-from pytensor.gradient import NullTypeGradError
-from scipy import optimize
+try:
+    from preliz.unidimensional import maxent
```
ricardoV94: this should only happen when someone tries to call `find_constrained_prior` (including the import, to not slowdown pymc import in general)

Also: why do you import modules from sys?

ricardoV94 (changes requested):

ricardoV94 (approved):
I still think we should revisit a pymc-based implementation later on, because it wouldn't be restricted to the offers /parametrizations in preliz and you could even have symbolic inputs / fixed parameters so you can apply it to intermediate variables, now that we have symbolic MinimizeOp in PyTensor.

ricardoV94:
Adding preliz to the env seems to have broken CI (or if not, something unrelated to this PR. We may be testing on a too old version of Python... (didn't check)

aloctavodia:
Makes sense. I think there is no unidimensional distribution/parameterization in pymc absent in preliz. But having symbolic inputs could be interesting. Maybe something worth to try is to replace variance with entropy. While entropy is more general and the correct thing to do. Maximizing the variance should work in most cases (or some other measure of spread).

aloctavodia:
Preliz use to have some functions that were removed and replaced for function already implemented in arviz-stats, which requieres python > 3.10

ricardoV94:
If python 3.10 is no longer supported by the numpy NEP we can drop it here as well, but maybe worth doing in a separate PR. CC'ing @maresb in case he wants to do it and you don't have the capacity @aloctavodia ?

ricardoV94:
Sidenote, I guess we don't have mode in Beta in PyMC? Or did I miss the feature being introduced at some point?

aloctavodia:
I don't remember the exact date to drop 3.10, but it was in the first semester. 
I also don't remember if the mode is available on PyMC, but that should not be a problem for this function to work, as the mode restraint is applied to a preliz distribution, and if the distribution does not have a mode, the user is notified.

maresb:
I have no strong feelings. I am pretty low availability this month. I'll take care of the conda-forge updates for everything though. (Ping me if needed.)

PR #7910, opened 2025-09-23
Update PyTensor dependency (drop numpy <2.0, and Python 3.10.)

ricardoV94:
Some major import changes in the latest release

----
📚 Documentation preview 📚: https://pymc--7910.org.readthedocs.build/en/7910/

### DIFF
--- a/tests/sampling/test_jax.py
+++ b/tests/sampling/test_jax.py
@@ -27,7 +27,6 @@ import pytensor
 import pytensor.tensor as pt
 import pytest
 
-from numpyro.infer import MCMC
 from pytensor.compile import SharedVariable
 from pytensor.graph import graph_inputs
 
@@ -44,6 +43,8 @@ from pymc.sampling.jax import (
     sample_blackjax_nuts,
     sample_numpyro_nuts,
 )
+
+MCMC = pytest.importorskip("numpyro.infer.MCMC")
 
 
 def test_jax_PosDefMatrix():
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           environment-file: conda-envs/environment-test.yml
           create-args: >-
-            python=3.10
+            python=3.11
           environment-name: pymc-test
           init-shell: bash
           cache-environment: true
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -146,7 +146,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       TEST_SUBSET: ${{ matrix.test-subset }}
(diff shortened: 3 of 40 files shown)


ricardoV94:
ugh loads of broken stuff

ricardoV94:
Failures had to do with functions that guessed the inputs for the user. The order changed since the last release and now gradients / forward were not aligned.

This was not a great idea to begin with, so I now raise anytime such method is called in a way that the input order is ambiguous.

ricardoV94:
CI broken by https://github.com/pymc-devs/pytensor/pull/1660 which means we'll need another release (which is also a major release as it drops old numpy / python) :)

ricardoV94:
getting close. @maresb @lucianopaz any luck with the windows blas issue on conda?

ricardoV94:
The verbose PR title is just to remind to add a note in the release notes

ricardoV94:
I'm not testing on Windows because the release is held at conda-forge. I'll open an issue to revert the relevant commit once https://github.com/conda-forge/pytensor-suite-feedstock/pull/177 is fixed

maresb:
Ya, sorry, I'm working on it, but not sure yet what the problem is. Will try to follow Luciano's advice

maresb:
> Ya, sorry, I'm working on it, but not sure yet what the problem is. Will try to follow Luciano's advice

Solved! :tada:

ricardoV94:
reverted the commit so we test on windows as usual

ricardoV94:
Uhuh tests are passing

lucianopaz (approved):
I think the changes look OK, but maybe someone else should have a look too.

tests/sampling/test_jax.py
```diff
@@ -45,6 +44,8 @@
     sample_numpyro_nuts,
 )
 
+MCMC = pytest.importorskip("numpyro.infer.MCMC")
```
Armavica, line 47: What is the behaviour of `importorskip` outside a test function? Does it skip all the rest of the file?

Armavica (approved):

tests/sampling/test_jax.py
ricardoV94: The whole module

ricardoV94: Should change, good catch

jessegrabowski: This whole file depends on numpyro/blackjax, so we should be doing importorskip at the file level. Otherwise we have to do it on a test-by-test basis, which will be a lot more verbose

ricardoV94: okay then

maresb:
Oops! We forgot to update python_requires:

https://github.com/pymc-devs/pymc/blob/f80e896a94f4fa6252e1a9991320de976a369109/setup.py#L73

PR #7908, opened 2025-09-22
Use `dict_to_dataset_drop_incompatible_coords` everywhere

ricardoV94:
Closes #7891 

The example in that issue:

```python
import pymc as pm

with pm.Model(coords={'dim_1': ['a', 'b', 'c'], 'other': [1, 2, 3, 4]}) as m:
    x = pm.Normal('x', mu=0, sigma=1, dims=['dim_1'])
    mu = pm.Deterministic('mu', 1 + x, dims=['other'])
    sigma = pm.HalfNormal('sigma', 1)

    y = pm.Normal('y', mu=mu, sigma=sigma, dims=['dim_1'])
    idata = pm.sample()
    
```
Outputs

```
UserWarning: Incompatible coordinate length of 4 for dimension 'other' of variable 'mu'.
The original coordinates for this dim will not be included in the returned dataset for any of the variables. Instead they will default to `np.arange`, possibly right-padded with nan.
To make this warning into an error set `pymc.backends.arviz.RAISE_ON_INCOMPATIBLE_COORD_LENGTHS` to `True`
```

----
📚 Documentation preview 📚: https://pymc--7908.org.readthedocs.build/en/7908/

PR #7912, opened 2025-09-24
Graceful incompatible coords handling

### LINKED ISSUE #7891: Check idata dims/coords for consistency before sampling begins
### Description

This is a painful gotcha I get hit by once in a while. If you have inconsistencies between the model coords and variable shapes, PyMC will still happy sample (often for hours, ask me how I know), then error out when it gets to the final step of building the idata.

I don't think there needs to be complex logic to validate static shapes etc, I'm thinking we could do a smoke test on the model initial point (or a single prior draw) to make sure it can be converted to an idata, and raise *before* sampling if not.

SelimSBE:
## Description
This PR updates the internal conversion logic to use `dict_to_dataset_drop_incompatible_coords` across all relevant backends. 

## Related Issue
- [x] Related to #7908
- [x] Closes #7891

## Checklist

## Type of change
- [x] Maintenance

----
📚 Documentation preview 📚: https://pymc--7912.org.readthedocs.build/en/7912/

### DIFF
--- a/tests/backends/test_arviz.py
+++ b/tests/backends/test_arviz.py
@@ -851,7 +851,16 @@ class TestDatasetToPointList:
         assert pl[0]["x"].dtype == np.float64
 
 
-def test_incompatible_coordinate_lengths():
+@pytest.mark.parametrize(
+    "sampling_method",
+    (
+        lambda: pm.sample_prior_predictive(draws=1).prior,
+        lambda: pm.sample(
+            chains=1, draws=1, tune=0, compute_convergence_checks=False, progressbar=False
+        ).posterior,
+    ),
+)
+def test_incompatible_coordinate_lengths(sampling_method):
     with pm.Model(coords={"a": [-1, -2, -3]}) as m:
         x = pm.Normal("x", dims="a")
         y = pm.Deterministic("y", x[1:], dims=("a",))
@@ -862,7 +871,7 @@ def test_incompatible_coordinate_lengths():
                 "Incompatible coordinate length of 3 for dimension 'a' of variable 'y'"
             ),
         ):
-            prior = pm.sample_prior_predictive(draws=1).prior.squeeze(("chain", "draw"))
+            prior = sampling_method().squeeze(("chain", "draw"))
         assert prior.x.dims == prior.y.dims == ("a",)
         assert prior.x.shape == prior.y.shape == (3,)
         assert np.isnan(prior.y.values[-1])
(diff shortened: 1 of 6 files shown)


SelimSBE:
@jessegrabowski as requested, I’ve just opened this PR for ticket #7891 , it’s ready for review. Thanks!

tests/backends/test_arviz.py
```diff
@@ -213,15 +213,16 @@ def test_predictions_to_idata_new(self, data, eight_schools_params):
...
-            posterior_predictive = pm.sample_posterior_predictive(
-                data.obj, return_inferencedata=False
-            )
-            inference_data = to_inference_data(
-                trace=data.obj,
-                posterior_predictive=posterior_predictive,
-                coords={"school": np.arange(eight_schools_params["J"])},
-                dims={"theta": ["school"], "eta": ["school"]},
-            )
+            with pytest.warns(UserWarning, match="Incompatible coordinate length"):
```
ricardoV94, line 216: Why were the coords invalid? I don't think they are supposed to be in this test, so maybe the test was faulty?

tests/backends/test_arviz.py
```diff
@@ -236,7 +237,8 @@ def test_posterior_predictive_thinned(self, data):
                 warnings.filterwarnings("ignore", ".*number of samples.*", UserWarning)
                 idata = pm.sample(tune=5, draws=draws, chains=2, return_inferencedata=True)
             thinned_idata = idata.sel(draw=slice(None, None, thin_by))
-            idata.extend(pm.sample_posterior_predictive(thinned_idata))
+            with pytest.warns(UserWarning, match="Incompatible coordinate length"):
```
ricardoV94, line 240: This should only affect the draw dimension, which is not part of the model?

tests/distributions/test_timeseries.py
```diff
@@ -951,7 +951,8 @@ def _gen_sde_path(sde, pars, dt, n, x0):
         with model:
             trace = sample(chains=1, random_seed=numpy_rng)
 
-        ppc = sample_posterior_predictive(trace, model=model, random_seed=numpy_rng)
+        with pytest.warns(UserWarning, match="Incompatible coordinate length"):
```
ricardoV94, line 954: What is the invalid coord here? We should fix the model if there's one

tests/backends/test_arviz.py
OriolAbril: I am also confused by this. Wouldn't these have errored out before the behaviour change breaking the test if they are now emiting this warning?

Maybe the condition for the warning is now too general?

Issue #7907, opened 2025-09-21
Add `transform_coords` method to `Transform` method to automatically label unconstrained draws


jessegrabowski:
### Description

Certain transformations change the shape of a variable, which causes us to lose dimension information for the transformed variable. An example is `SimplexTransform`, which removes the last value of the core dimension. This means that when we return inference data for transformed values, we have to fall back to assigning new, unknown dims/labels, even though we actually know what these should be.

My proposal is to add a method to each transformation that knows what to do with labels. Then if we want to return the unconstrained samples for whatever reason, we can consult the relevant transformations and give the user back label dimensions as expected. Many of the most common transformations will just end up being the identity function.

This would add an `abstractmethod` to the `Transform` base class, something like:

```py
class Transform(abc.ABC):
    ...

    @abc.abstractmethod
    def transform_labels(self, labels: Sequence[str]) -> Sequence[str]:
        """Mutate user-provided coordinates associated with the variable to label transformed values returned by this class"""
```

ricardoV94:
I guess you would want a method for forward and backward, although in one direction you would have to make up labels :D

jessegrabowski:
We could embed and ship with an llm to generate them :) 

I didn't propose a method for going the other way because it would have to make something up. I envision the main usecase to just be for the `unconstrained_parameters` idata group.

ricardoV94:
Right

AndrewKillion:
Are `SimplexTransform` and `ChainedTransform` the only ones where `transform_labels` don't end up being identity functions?

For `SimplexTransform` i think this would do it right?
```Python
    def transform_labels(self, coords):
        if len(coords) == 0:
            return coords
        return coords[:-1]
```

jessegrabowski:
That looks right for `SimplexTransform`, yes. I'm not sure if there are other transformation that mutate the coords. If we ever finished #7700 that would be one.

ricardoV94:
ZeroSumNormal? SumTo1?

faraz-shamim:
hi @ricardoV94 @jessegrabowski ! please review. Thankyou!

PR #7913, opened 2025-09-26
Add transform_coords abstract method. SimplexTransform drops last coo…

AndrewKillion:
## Description

   - Add abstract transform_coords method to Transform base class
   - Implement method in all concrete Transform classes  
   - SimplexTransform and SumTo1 remove last coordinate
   - Most other transforms preserve coordinates
   - Add comprehensive test suite

## Related Issue

- [ ] Closes #
- [X] Related to #7907

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7913.org.readthedocs.build/en/7913/

### DIFF
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -35,7 +35,7 @@
 #   SOFTWARE.
 import abc
 
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 
 import numpy as np
 import pytensor.tensor as pt
@@ -140,6 +140,10 @@ class Transform(abc.ABC):
 
         Multiple values may be returned when the transformation is not 1-to-1.
         """
+
+    @abc.abstractmethod
+    def transform_coords(self, coords: Sequence[str]) -> Sequence[str]:
+        """Mutate user-provided coordinates associated with the variable to label transformed values returned by this class."""
 
     def log_jac_det(self, value: TensorVariable, *inputs) -> TensorVariable:
         """Construct the log of the absolute value of the Jacobian determinant."""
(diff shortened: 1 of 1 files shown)


jessegrabowski:
Thanks for opening this!

Docs are failing because you missed adding `transform_coords` to a few transformers. My suggestion would actually be to use multiple inheritance in this case, we can have a `OneToOneTransformerMixin` class that just defines `transform_coords(coords): return coords`, and you can use that everywhere.

ricardoV94:
> OneToOneTransformerMixin class that just defines transform_coords(coords): return coords, and you can use that everywhere.

Have a subclass of Transform that behaves like that, and use that? Why mixin?

ricardoV94:
Actually all these with `ndim_supp=0`, are those cases. Scalar transforms that don't modify coords

jessegrabowski:
> Have a subclass of Transform that behaves like that, and use that? Why mixin?

Mixins are cool and hip. `sklearn` does mixins!

Anyway a subclass would accomplish the same thing as well, and I have no strong preference for one over the other.

ricardoV94:
I just don't see what it accomplishes, seems more complex than it needs to be

jessegrabowski:
I guess it's meant to give classes a more "modular" feel? It's not that a transfomer that doesn't mutate coords isn't a transformer (it is), it just also a coordinate transformation behavior  defined by a `OneToOneCoordMixin`. A `OneToOneTransfomer` subclass would accomplish the same thing, but it would feel conceptually somewhat more removed from the `Transformer` base class, which it isn't.

ricardoV94:
Let's do it when we have a reason for needing that modularity. These are not really things we're composing or intending to in the foreseeable future

jessegrabowski:
Like I said, I don't object to a subclass. I'm just making the case for a mixin.

ricardoV94:
I object to a mixin and I'm not making a case but I can

PR #7917, opened 2025-09-26
Add icdf functions for Inverse Gamma distribution

lucaseckes:
Adds ICDF functions to Inverse Gamma distribution

## Description
The quantile function depends on the quantile function of Gamma distribution: https://github.com/pymc-devs/pymc/pull/6845

## Related Issue

- [ ] Closes #
- [x] Related to https://github.com/pymc-devs/pymc/issues/6612

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7917.org.readthedocs.build/en/7917/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2530,6 +2530,16 @@ class InverseGamma(PositiveContinuous):
             beta > 0,
             msg="alpha > 0, beta > 0",
         )
+    
+    def icdf(value, alpha, beta):
+        res = 1 / Gamma.icdf(value=1-value, alpha=alpha, scale=1 / beta)
+        res = check_icdf_value(res, value)
+        return check_icdf_parameters(
+            res,
+            alpha > 0,
+            beta > 0,
+            msg="alpha > 0, beta > 0",
+        )
 
 
 class ChiSquared:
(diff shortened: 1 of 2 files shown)


jessegrabowski (approved):
Looks good to me! @ricardoV94 do you want to check it before I snap merge?

pymc/distributions/continuous.py
```diff
@@ -2531,6 +2531,16 @@ def logcdf(value, alpha, beta):
             msg="alpha > 0, beta > 0",
         )
 
+    def icdf(value, alpha, beta):
+        res = 1 / Gamma.icdf(value=1 - value, alpha=alpha, scale=1 / beta)
```
ricardoV94, line 2535: Better to use the stable API, and then you don't need to hack with the scale thing either

```suggestion
        res = 1 / icdf(Gamma.dist(alpha=alhpa, beta=beta), value=1 - value)
```

ricardoV94, line 2535: For curiosity only, any intuition why the icdf looks like it does, specially the flipped value. It's interesting

ricardoV94:
I suggest a more general approach in #7956

PR #7933, opened 2025-10-19
Fix/missing init jitter function

AhmadYasser1:
## Summary
Fixes a critical bug where the `_init_jitter` function was being called but didn't exist, causing a `NameError` at runtime.

## Problem
The code in `pymc/sampling/mcmc.py` was calling `_init_jitter()` function on line 1598, but this function no longer exists in the codebase. This would cause a `NameError` at runtime when trying to initialize MCMC sampling with jitter.

## Solution
- Replaced the call to `_init_jitter` with the modern `make_initial_point_fns_per_chain` API
- Added proper error handling for jitter failures with fallback to no-jitter mode
- Updated the `test_init_jitter` test to use the new initial point generation API
- Removed the `@pytest.mark.skip` decorator from the test

## Changes
1. **pymc/sampling/mcmc.py**: Replaced `_init_jitter` call with `make_initial_point_fns_per_chain`
2. **tests/sampling/test_mcmc.py**: Updated test to use new API and removed skip decorator

## Impact
- Fixes a critical runtime error that would prevent MCMC sampling with jitter
- Enables the previously skipped test to run
- Improves code reliability and maintainability

## Testing
The fix maintains backward compatibility while using the modern initial point generation API. The test now runs without being skipped.

----
📚 Documentation preview 📚: https://pymc--7933.org.readthedocs.build/en/7933/


docs/source/installation.md
```diff
@@ -1,7 +1,7 @@
 (installation)=
 # Installation
 
-We recommend using [Anaconda](https://www.anaconda.com/) (or [Miniforge](https://github.com/conda-forge/miniforge)) to install Python on your local machine, which allows for packages to be installed using its `conda` utility.
+PyMC requires Python 3.11 or greater. We recommend using [Anaconda](https://www.anaconda.com/) (or [Miniforge](https://github.com/conda-forge/miniforge)) to install Python on your local machine, which allows for packages to be installed using its `conda` utility.
```
twiecki, line 4: None of these or previous changes should be in this PR.

PR #7935, opened 2025-10-23
Fix print name of `LogitNormal` and `DiscreteWeibull`

aloctavodia:
`dist.owner.op._print_name[0]` was returning incorrect names for:
* DiscreteWibull and LogitNormal, as they did not follow the conventions used for the rest of the distributions
* ~ChiSquared was returning "Gamma", because it's implemented as a special case of Gamma.~

~Not sure if there is a simpler way to fix the issue for ChiSquared.~

----
📚 Documentation preview 📚: https://pymc--7935.org.readthedocs.build/en/7935/

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -2577,20 +2577,12 @@ class ChiSquared:
         Degrees of freedom (nu > 0).
     """
 
-    name = "chi_squared"
-    extended_signature = "[rng],[size],(),()->[rng],()"
-    _print_name = ("ChiSquared", "\\operatorname{ChiSquared}")
-
     def __new__(cls, name, nu, **kwargs):
-        obj = Gamma(name, alpha=nu / 2, beta=1 / 2, **kwargs)
-        obj.owner.op._print_name = cls._print_name
-        return obj
+        return Gamma(name, alpha=nu / 2, beta=1 / 2, **kwargs)
 
     @classmethod
     def dist(cls, nu, **kwargs):
-        obj = Gamma.dist(alpha=nu / 2, beta=1 / 2, **kwargs)
-        obj.owner.op._print_name = cls._print_name
-        return obj
+        return Gamma.dist(alpha=nu / 2, beta=1 / 2, **kwargs)
 
 
 class WeibullBetaRV(SymbolicRandomVariable):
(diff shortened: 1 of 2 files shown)


pymc/distributions/continuous.py
```diff
@@ -2532,7 +2532,22 @@ def logcdf(value, alpha, beta):
         )
 
 
-class ChiSquared:
+class ChiSquaredRV(RandomVariable):
```
ricardoV94: We shouldn't add more Ops just for naming. Less Ops means we can sample in other backends / do rewrites with fewer building blocks

aloctavodia: In the last commit ok?

ricardoV94: Nope, that's just a temporary hack. As soon as you do something like `model.copy()`, or any other model transform it's going to be lost. Why do you need it to show up as ChiSquared in the graphviz? 

It's just a parametrization. Similarly we don't write UnitNormal if it happens to have loc=0, scale=1, even if we offered a helper to create them?

aloctavodia: I was creating a pymc_to_preliz function with the inmediate goal of being able to call pz.plot() for both preliz and pymc distributions. I found this and assumed it was of broader interest to show proper names. But if not, I can try to solve it inside that function.

ricardoV94: I don't think naming should come at the cost of more code complexity/redundancy. This is just a helper that creates a specific Gamma and seems sensible to write it as such. Can't you call pz.plot on the gamma object that's returned?

PS: We can emphasize more in the docstrings that this is just a helper that return a regular Gamma distribution

aloctavodia: Yes, the call to pz.plot works fine. Because the ChiSquared is correctly converted into a Gamma PreliZ distribution, and then everything works. The only "issue" is that the result is a Gamma and not a ChiSquared. So this was more about avoiding potential confusion on the user side than fixing something not working.

ricardoV94: Do you want to add a comment to reduce confusion for future devs/users?

aloctavodia: Added. I will also add a note on the PreliZ side.

ricardoV94 (approved):

PR #7948, opened 2025-11-03
Implement ICAR prior as thin MvNormal wrapper with eigen-based pseudo-inverse covariance

premshaw04:
## Related Issue

- [ ] Closes #
- [x] Related to #

## Checklist

## Type of change

- [*] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7948.org.readthedocs.build/en/7948/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -67,6 +67,7 @@ from pymc.distributions.distribution import (
 )
 from pymc.distributions.shape_utils import (
     _change_dist_size,
+    build_icar_covariance,
     change_dist_size,
     get_support_shape,
     implicit_size_from_params,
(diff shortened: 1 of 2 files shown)


pymc/distributions/multivariate.py
```diff
@@ -2452,6 +2453,15 @@ class ICAR(Continuous):
     rv_op = icar
 
     @classmethod
+    def ICAR(name, W, tau=1.0, mu=None, **kwargs):
```
williambdean, line 2456: Doesn't seem like you want the classmethod decorator here. It should stay on `dist` method .

ricardoV94:
There was discussion going on in the original PR/Issue as to whether we want to do this at all. No point in resuming the PR without a consensus first

pymc/distributions/multivariate.py
premshaw04: ok sir i will fix it.

PR #7951, opened 2025-11-04
Don't use deprecated batched_dot

### LINKED ISSUE #7878: KroneckerNormal using deprecated batched_dot
https://github.com/pymc-devs/pymc/blob/dc7cfeeaba98330f2881d627ee988ac267f18df2/pymc/distributions/multivariate.py#L2128

I'm not sure if this is supposed to be a matmul or vecdot. I think the later. If someone is familiar with the distribution please weigh in.

CC @bwengals or perhaps @jessegrabowski (in case we can simplify this with PyTensor magic)

atheendre130505:
## Description
This PR replaces the deprecated `pt.batched_dot` function with the preferred `pt.sum` operation in the KroneckerNormal distribution's logprob calculation, addressing issue #7878.

## Problem
The current implementation uses `batched_dot`, which is deprecated in PyTensor and triggers warnings. Deprecated functions may lead to future breakage and lower performance.

## Solution
Refactored the `KroneckerNormal` logprob code to use `pt.sum` with explicit axis parameters, achieving the same functionality without relying on deprecated APIs.

## Tests
- All existing tests related to KroneckerNormal pass successfully.
- No deprecation warnings are shown when running the updated code.
- Validated numerical equivalence with the previous implementation.

## Related Issue
Fixes #7878

## Checklist
- [x] Follows PyMC contributing guidelines
- [x] All tests pass locally
- [x] No API-breaking changes

----
📚 Documentation preview 📚: https://pymc--7951.org.readthedocs.build/en/7951/


atheendre130505:
pre-commit.ci autofix

pymc/distributions/multivariate.py
```diff
@@ -2124,8 +2124,8 @@ def logp(value, rng, size, mu, sigma, *covs):
         sqrt_quad = sqrt_quad / pt.sqrt(eigs[:, None])
         logdet = pt.sum(pt.log(eigs))
 
-        # Square each sample
-        quad = pt.batched_dot(sqrt_quad.T, sqrt_quad.T)
+        # Square each sample - compute squared norm for each sample
+        quad = pt.sum(sqrt_quad.T**2, axis=-1)
```
jessegrabowski, line 2129:
```suggestion
        quad = pt.sum(sqrt_quad**2, axis=-2)
```
More explicit

jessegrabowski (changes requested):

pymc/distributions/multivariate.py
jessegrabowski: The comment is also incorrect, it's not a "squared norm".

atheendre130505:
@jessegrabowski @ricardoV94 @zaxtax

PR #7955, opened 2025-11-05
Premshaw04/#7713/ICAR wrapper should return MvNormal.dist instead of constructing the RV directly

### LINKED ISSUE #7713: Add `rng_fn` to CAR/ICAR

premshaw04:
## Related Issue

- [*] Closes #7713 
- [ ] Related to #

## Checklist

- [*] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [*] Included tests that prove the fix is effective or that the new feature works

## Type of change

- [*] New feature / enhancement
- [ ] Bug fix
- [] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7955.org.readthedocs.build/en/7955/

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -67,6 +67,7 @@ from pymc.distributions.distribution import (
 )
 from pymc.distributions.shape_utils import (
     _change_dist_size,
+    build_icar_covariance,
     change_dist_size,
     get_support_shape,
     implicit_size_from_params,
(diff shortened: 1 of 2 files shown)


premshaw04:
please pre-commit.ci autofix

PR #7916, opened 2025-09-26
Add icdf for `LogitNormal`

jnetzel1:
**What is this PR about?**
Adds ICDF functions to Logit Normal Distribution

Issue https://github.com/pymc-devs/pymc/issues/6612

**References:**
Logit Normal:
https://en.wikipedia.org/wiki/Logit-normal_distribution

SciPy Implementations:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.expit.html
https://github.com/scipy/scipy/blob/d604a5462dadf799f8af2c177179a01875b40796/scipy/stats/_continuous_distns.py#L380

## Checklist
- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [x] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change]

## Type of change
- [x] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7916.org.readthedocs.build/en/7916/

PR #7619, opened 2024-12-20
implement invgamma icdf + test

berkalpay:
## Description
Implements the ICDF of the InverseGamma distribution, for #6612, with a test comparing it to the scipy implementation.

----
📚 Documentation preview 📚: https://pymc--7619.org.readthedocs.build/en/7619/

PR #7956, opened 2025-11-08
Automatically derive icdf for negative power transforms

ricardoV94:
(* non even powers)

This way we get icdf easily for InverseGamma and LogitNormal. I removed the latter in favor of a CustomDist. I didn't do it for InverseGamma because that's a pretty popular distribution and don't want to risk messing with precision

closes #7619
closes #7917
closes #7916

### DIFF
--- a/pymc/distributions/continuous.py
+++ b/pymc/distributions/continuous.py
@@ -53,6 +53,7 @@ from pytensor.tensor.random.op import RandomVariable
 from pytensor.tensor.random.utils import normalize_size_param
 from pytensor.tensor.variable import TensorConstant, TensorVariable
 
+from pymc.distributions.custom import CustomDist
 from pymc.logprob.abstract import _logprob_helper
 from pymc.logprob.basic import TensorLike, icdf
 from pymc.pytensorf import normalize_rng_param
@@ -92,7 +93,7 @@ from pymc.distributions.dist_math import (
 from pymc.distributions.distribution import DIST_PARAMETER_TYPES, Continuous, SymbolicRandomVariable
 from pymc.distributions.shape_utils import implicit_size_from_params, rv_size_is_none
 from pymc.distributions.transforms import _default_transform
-from pymc.math import invlogit, logdiffexp, logit
+from pymc.math import invlogit, logdiffexp
 
 __all__ = [
     "AsymmetricLaplace",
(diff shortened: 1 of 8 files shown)


jessegrabowski:
Tests are failing because you can't pass rng to the new signatures

pymc/distributions/continuous.py
```diff
@@ -2531,6 +2532,9 @@ def logcdf(value, alpha, beta):
             msg="alpha > 0, beta > 0",
         )
 
+    def icdf(value, alpha, beta):
```
larryshamalama, line 2535: I don't visualize the math on top of my head. I presume this is what `test_inverse_gamma_icdf` below checks?

tests/distributions/test_continuous.py
```diff
@@ -687,6 +687,13 @@ def test_inverse_gamma_logcdf(self):
             lambda value, alpha, beta: st.invgamma.logcdf(value, alpha, scale=beta),
         )
 
+    def test_inverse_gamma_icdf(self):
```
larryshamalama, line 690: Some distributions have all functionality tests (logp, logcdf, icdf) under a single method (`test_half_cauchy`), whereas other distributions have one method per functionality. Is there any particular reason for this?

larryshamalama (commented):
Hi @ricardoV94 , great work as always :) I haven't followed much of recent codebase progress, as per @jessegrabowski 's comment other changes seem necessary w.r.t. test failures. I just added minor questions that came to mind as per usual as I read this PR

pymc/distributions/continuous.py
ricardoV94: yes

tests/distributions/test_continuous.py
ricardoV94: No grand plan. I prefer separate but if a function was already testing multiple for that distribution I will just append to it

ricardoV94:
> Tests are failing because you can't pass rng to the new signatures

CustomDist didn't allow passing explicit rng. I tweaked to allow it. This will be useful as we transition to more explicit rng handling internally in the future. Otherwise I could just test the RV differently, since this is such a trivial RV

jessegrabowski (approved):

PR #7958, opened 2025-11-10
Infer logp and logcdf of abs of discrete variables

ricardoV94:
I was going to work on general icdf(abs(x)) which requires an optimization routine, but found some small crumbs along the way.

* Allow logp(abs(x)) and logcdf(abs(x)) for discrete x. 
* Fix logcdf of DiscreteUniform for x == lower
* Handle special case in logdiffexp(-inf, -inf) so it can be safely used in more contexts

### DIFF
--- a/pymc/logprob/transforms.py
+++ b/pymc/logprob/transforms.py
@@ -40,7 +40,7 @@ from collections.abc import Callable
 import numpy as np
 import pytensor.tensor as pt
 
-from pytensor import scan
+from pytensor import graph_replace, scan
 from pytensor.gradient import jacobian
 from pytensor.graph.basic import Apply, Variable
 from pytensor.graph.fg import FunctionGraph
@@ -163,6 +163,8 @@ class Transform(abc.ABC):
 class MeasurableTransform(MeasurableElemwise):
     """A placeholder used to specify a log-likelihood for a transformed measurable variable."""
 
+    __props__ = ("scalar_op", "inplace_pattern", "is_discrete")
+
     valid_scalar_types = (
         Exp,
         Log,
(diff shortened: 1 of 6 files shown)


pymc/logprob/transforms.py
```diff
@@ -224,8 +271,11 @@ def measurable_transform_logprob(op: MeasurableTransform, values, *inputs, **kwa
         ndim_supp = value.ndim - input_logprob.ndim
         jacobian = jacobian.sum(axis=tuple(range(-ndim_supp, 0)))
 
+    # Discrete transformations do not need the jacobian adjustment
```
ricardoV94, line 274: We don't really support any discrete cases with det(jacobian) != 1 but I prefer to have this branch already

tests/logprob/test_transforms.py
```diff
@@ -501,21 +539,33 @@ def test_negative_value_frac_power_transform_logp(self, power):
...
-    y_rv = pt.random.halfnormal()
+@pytest.mark.parametrize("continuous", (True, False))
+def test_absolute_rv_transform(continuous):
+    if continuous:
+        x_rv = pt.abs(pt.random.normal(size=(5,)))
+        ref_rv = pt.random.halfnormal(size=(5,))
+    else:
+        x_rv = pt.abs(DiscreteUniform.dist(-4, 4, size=(5,)))
+        # |x_rv| = DiscreteUniform(0,4) with P(X=0) halved relative to other values
+        # We can use a Categorical to representh this
```
jessegrabowski, line 550:
```suggestion
        # We can use a Categorical to represent this
```

jessegrabowski (approved):
I found a typo! I'm helping!

PR #7963, opened 2025-11-16
Tweak run_mypy.py to be more user-friendly

jessegrabowski:
Very small changes to reflect how this script is used by developers:

- `--verbose`  is now the default
- By default, "expected" failures are filtered. You can see all failures by passing a `--show-expected` flag.

Justifications:

1. I have never run this command without `--verbose`, so why not?
2. The way this script is used is to fix mypy problems with a specific PR. This requires fixing only the new failures. Unless we're actively working on cleaning up the code base, we're only interested in the new failures.

### DIFF
--- a/scripts/run_mypy.py
+++ b/scripts/run_mypy.py
@@ -96,7 +96,7 @@ def mypy_to_pandas(input_lines: Iterator[str]) -> pandas.DataFrame:
     return pandas.DataFrame(data=data).set_index(["file", "line"])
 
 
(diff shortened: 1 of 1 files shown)


jessegrabowski:
It was pointed out [here](https://github.com/pymc-devs/pytensor/pull/1598#issuecomment-3446673403) that we should just be using `--output json` instead of manually parsing the mypy output, so I've also done that.

michaelosthege (approved):
Looks good to me 👍

Armavica:
Ohh thank you for this change!

PR #7964, opened 2025-11-17
Use `narwhals` as dataframe-agnostic backend 

### LINKED ISSUE #7462: ENH: Replace pandas dependence/use with narwhals

jessegrabowski:
## Description
Continues/completes [#7462](https://github.com/pymc-devs/pymc/pull/7918). I didn't have permission to push into that PR, so I'm opening this one. 

The purpose of this PR is to use narwhals as a one-stop shop dataframe backend. Currently, we use pandas in `data.py` and `pytensorf.py` to allow users to pass dataframe objects into `pm.Data` and `pt.as_tensor`, respectively. I add a narwhals compatibility layer between the input and the pymc model to allow the user to bring his data in any form that narwhals supports, provided we register the libraries. 

(If we could eliminate registration that would also be great, but I wasn't clever enough to figure out the multiple dispatch using only narwhals as a dependency. Maybe @MarcoGorelli could help 👉 👈 )

 Some notes:

1. Since generalized dataframes don't have a notion of an index, we don't look at the index to find the labels for the left-most dimension provided to `pm.Data`. Instead, we look for a column matching that dimension name. If it is found, it is used as labels, and excluded from the values. 
2. Narwhals has a lazy API via `nw.LazyFrame` and `nw.LazySeries`. I don't think we can do anything with those at the modeling level (maybe in the future with minibatching?). For now, I'm just calling `.collect()` on them to make them eager. 
3. As mentioned above, we don't get all of narwhals for free yet -- the PR as it currently stands forces us to register each backend library we want to support. I don't think this is so bad, because it forces us to write tests for say DuckDB if someone comes along and really wants that. But it's a bit ugly. I added a `dask.dataframe` backend as an example of how we could extend things.

As of this PR, `pandas` could be made an optional or dev-only dependency for us. I didn't do it right away because I wanted to take people's temperature on the idea.

## Related Issue

- [ ] Closes #7462 Closes #7463
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/data.py
+++ b/pymc/data.py
@@ -18,20 +18,17 @@ import urllib.request
 
 from collections.abc import Sequence
 from copy import copy
-from typing import Union, cast
 
 import numpy as np
-import pandas as pd
 import pytensor
 import pytensor.tensor as pt
-import xarray as xr
 
 from pytensor.compile.builders import OpFromGraph
 from pytensor.compile.sharedvalue import SharedVariable
 from pytensor.graph.basic import Variable
 from pytensor.raise_op import Assert
 from pytensor.tensor.random.basic import IntegersRV
-from pytensor.tensor.variable import TensorConstant, TensorVariable
+from pytensor.tensor.variable import TensorVariable
 
 from pymc.exceptions import ShapeError
 from pymc.pytensorf import convert_data, rvs_in_graph
@@ -163,76 +160,17 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size: 
     return mb_tensors if len(variables) else mb_tensors[0]
 
 
-def determine_coords(
-    model,
-    value: pd.DataFrame | pd.Series | xr.DataArray,
-    dims: Sequence[str] | None = None,
-    coords: dict[str, Sequence | np.ndarray] | None = None,
-) -> tuple[dict[str, Sequence | np.ndarray], Sequence[str] | Sequence[None]]:
-    """Determine coordinate values from data or the model (via ``dims``)."""
-    if coords is None:
-        coords = {}
-
-    dim_name = None
-    # If value is a df or a series, we interpret the index as coords:
-    if hasattr(value, "index"):
-        if dims is not None:
-            dim_name = dims[0]
-        if dim_name is None and value.index.name is not None:
-            dim_name = value.index.name
-        if dim_name is not None:
-            coords[dim_name] = value.index
-
-    # If value is a df, we also interpret the columns as coords:
-    if hasattr(value, "columns"):
-        if dims is not None:
-            dim_name = dims[1]
-        if dim_name is None and value.columns.name is not None:
-            dim_name = value.columns.name
-        if dim_name is not None:
-            coords[dim_name] = value.columns
-
-    if isinstance(value, xr.DataArray):
-        if dims is not None:
-            for dim in dims:
-                dim_name = dim
-                # str is applied because dim entries may be None
-                coords[str(dim_name)] = cast(xr.DataArray, value[dim]).to_numpy()
-
-    if isinstance(value, np.ndarray) and dims is not None:
-        if len(dims) != value.ndim:
-            raise ShapeError(
-                "Invalid data shape. The rank of the dataset must match the length of `dims`.",
-                actual=value.shape,
-                expected=value.ndim,
-            )
-        for size, dim in zip(value.shape, dims):
-            coord = model.coords.get(dim, None)
-            if coord is None and dim is not None:
-                coords[dim] = range(size)
-
-    if dims is None:
-        # TODO: Also determine dim names from the index
-        new_dims: Sequence[str] | Sequence[None] = [None] * np.ndim(value)
-    else:
-        new_dims = dims
-    return coords, new_dims
-
-
 def Data(
     name: str,
     value,
     *,
     dims: Sequence[str] | None = None,
-    coords: dict[str, Sequence | np.ndarray] | None = None,
-    infer_dims_and_coords=False,
-    model: Union["Model", None] = None,
+    model: "Model | None" = None,
     **kwargs,
-) -> SharedVariable | TensorConstant:
+) -> SharedVariable:
     """Create a data container that registers a data variable with the model.
 
-    Depending on the ``mutable`` setting (default: True), the variable
-    is registered as a :class:`~pytensor.compile.sharedvalue.SharedVariable`,
+    The variable is registered as a :class:`~pytensor.compile.sharedvalue.SharedVariable`,
     enabling it to be altered in value and shape, but NOT in dimensionality using
     :func:`pymc.set_data`.
 
@@ -250,23 +188,19 @@ def Data(
     ----------
     name : str
         The name for this variable.
-    value : array_like or pandas.Series, pandas.Dataframe
-        A value to associate with this variable.
-    dims : str, tuple of str or tuple of None, optional
-        Dimension names of the random variables (as opposed to the shapes of these
-        random variables). Use this when ``value`` is a pandas Series or DataFrame. The
-        ``dims`` will then be the name of the Series / DataFrame's columns. See ArviZ
-        documentation for more information about dimensions and coordinates:
-        :ref:`arviz:quickstart`.
-        If this parameter is not specified, the random variables will not have dimension
-        names.
-    coords : dict, optional
-        Coordinate values to set for new dimensions introduced by this ``Data`` variable.
-    export_index_as_coords : bool
-        Deprecated, previous version of "infer_dims_and_coords"
-    infer_dims_and_coords : bool, default=False
-        If True, the ``Data`` container will try to infer what the coordinates
-        and dimension names should be if there is an index in ``value``.
+    value : array-like, DataFrame, or Series
+        A value to associate with this variable. Accepts numpy arrays, lists,
+        scipy sparse matrices, xarray DataArrays, and any DataFrame or Series
+        supported by `narwhals <https://narwhals-dev.github.io/narwhals/>`_
+        (pandas, polars, dask, pyarrow, modin, cudf, ibis, ...). The value is
+        converted to a numpy array.
+    dims : str or tuple of str, optional
+        Dimension names of the data variable. See ArviZ documentation for more
+        information about dimensions and coordinates: :ref:`arviz:quickstart`.
+        If not specified, the data variable will not have dimension names.
+    model : pymc.Model, optional
+        Model to which to add the data variable. If not specified, the data
+        variable will be added to the model on the context stack.
     **kwargs : dict, optional
         Extra arguments passed to :func:`pytensor.shared`.
 
@@ -293,13 +227,6 @@ def Data(
     """
     from pymc.model.core import modelcontext
 
-    if coords is None:
-        coords = {}
-
-    if isinstance(value, list):
-        value = np.array(value)
-
-    # Add data container to the named variables of the model.
     try:
         model = modelcontext(model)
     except TypeError:
@@ -309,14 +236,18 @@ def Data(
         )
     name = model.name_for(name)
 
-    # Transform `value` it to something digestible for PyTensor.
     if isgenerator(value):
         raise NotImplementedError(
             "Generator type data is no longer supported with pm.Data.",
             # It messes up InferenceData and can't be the input to a SharedVariable.
         )
-    else:
-        arr = convert_data(value)
+
+    if isinstance(value, list | tuple):
+        # Promote here so the inferred dtype (e.g. int64 for an index list) survives
+        # `convert_data`, instead of being floatX'd by the "no dtype" fallback.
+        value = np.asarray(value)
+
+    arr = convert_data(value)
 
     if isinstance(arr, np.ma.MaskedArray):
         raise NotImplementedError(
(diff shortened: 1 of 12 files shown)


pymc/data.py
```diff
@@ -161,65 +163,164 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size:
...
-    return coords, new_dims
+        raise ValueError(
+            f"Dimension '{index_dim}' not found in DataFrame columns or model coordinates. Cannot infer "
+            "index coordinates."
+        )
+
+    if len(dims) > 1:
+        column_dim = dims[1]
+        if column_dim is not None:
+            coords[column_dim] = value.select(nw.exclude(index_dim)).columns
```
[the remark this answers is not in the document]
jessegrabowski: `nw.exclude(None)` is valid and returns all columns

pymc/data.py
```diff
@@ -161,65 +163,164 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size:
...
+@determine_coords.register(np.ndarray)
+def determine_array_coords(
+    value: np.ndarray,
+    model: "Model",
     dims: Sequence[str] | None = None,
     coords: dict[str, Sequence | np.ndarray] | None = None,
+):
+    if coords is None:
+        coords = {}
+
```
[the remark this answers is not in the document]
jessegrabowski: The robot is confused on this one, the function early exists when dims is None, and never checks the length.

juanitorduz:
I love this! I will be reviewing this soon :)

pymc/data.py
```diff
@@ -161,65 +163,178 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size:
...
+    for dim in dims:
+        dim_name = dim
+        # str is applied because dim entries may be None
+        coords[str(dim_name)] = cast(xr.DataArray, value[dim]).to_numpy()
+
+    return coords, _handle_none_dims(dims, value.ndim)
+
+
+def _dataframe_agnostic_coords(
```
juanitorduz: Maybe we can add a simple test for this function?

juanitorduz (commented):
LGTM. Just a small comment / request

pymc/data.py
jessegrabowski: I added a test for this. The new test revealed that my code for inferring an "index column" in the dataframe based on the left-most dim was broken.

I was trying to do was to allow for an "index column" in backends with no index (Polars). So if your dims were `(date, country)` and you passed in a polars dataframe with columns [USA, Chile, Mozambique, date], it would drop the "date" column, use it as a coord, then pass in the rest as the data to `pm.Data`.  But this ended up being something of an API break, and I didn't like that we ended up mutating the user inputs silently in the background. I removed it, so only a pandas dataframe will give you index coords when `infer_coords_and_dims=True`

Frankly though I kind of hate this feature, it's too magical. Just force people to write out their coords.

ricardoV94:
Suggestion: Stop supporting inferring coords from `pm.Data`. One and only place to define them is in `Model(coords=...)`. Then `Data` and this PR can be simplified quite a lot (I suspect).

Would be major change, I'm okay with.

Before - choosing awkward example to make a point:
```python
some_coords = {"trial": [0, 1, 2]}
with pm.Model(coords=some_coords):
  x = pm.Data("x", some_dataframe, dims="day", infer_dims_and_coords=True)
  # Model coords changed, now has day and trial
  beta = pm.Normal("beta")
  mu = x * beta
  sigma = pm.HalfNormal(dims="trial")
  y = pm.Normal(mu[:, None], sigma, dims=("day", "trial")
```

After - only option:
```python
all_coords = {"trial": [0, 1, 2], "day": df.index}
with pm.Model(coords=all_coords):
  x = pm.Data("x", df, dims="day")
  beta = pm.Normal("beta")
  mu = x * beta
  sigma = pm.HalfNormal(dims="trial")
  y = pm.Normal(mu[:, None], sigma, dims=("day", "trial")
```

pymc/data.py
```diff
@@ -161,72 +176,109 @@ def Minibatch(variable: TensorVariable, *variables: TensorVariable, batch_size:
...
-            coord = model.coords.get(dim, None)
-            if coord is None and dim is not None:
-                coords[dim] = range(size)
+@singledispatch
+def prepare_user_data(data):
+    try:
+        # We can't explicitly dispatch on all possible DataFrame/Series types supported by narwhals without making them
+        # package dependencies. So in the generic function we can just try to convert using narwhals, then
+        # dispatch based on the native Narwhals types if it works.
+        df = nw.from_native(data, allow_series=True)
```
fonnesbeck: Does this support tuples?

pymc/pytensorf.py
```diff
@@ -75,64 +76,56 @@
...
+            return pt.as_tensor_variable(df.to_numpy(), *args, **kwargs)
+
+    except ImportError:
+        # Data backends are optional. Take no action if not installed.
+        pass
+
+
+_register_dataframe_backend("pandas")
+_register_dataframe_backend("polars")
+_register_dataframe_backend("dask.dataframe")
```
fonnesbeck: Do we want pyarrow too, or no?

pymc/data.py
jessegrabowski: I should have a dispatch for list | tuple, if I don't it's a mistake

pymc/pytensorf.py
jessegrabowski: yeah we can do as many as we want. This is just for `pt.as_tensor_variable` to be clear. We could also refuse to support that to make this PR simpler, and force the user to convert to numpy first.

pymc/pytensorf.py
```diff
@@ -48,6 +48,7 @@
 from pytensor.tensor.sharedvar import SharedVariable
 from pytensor.tensor.subtensor import AdvancedIncSubtensor, AdvancedIncSubtensor1
 from pytensor.tensor.variable import TensorVariable
+from pytensor.xtensor.type import XTensorVariable
```
ricardoV94: This is considered experimental, and there's a warning if you import it

pymc/pytensorf.py
```diff
@@ -75,64 +76,56 @@
...
 
-@_as_tensor_variable.register(pd.Series)
-@_as_tensor_variable.register(pd.DataFrame)
-def dataframe_to_tensor_variable(df: pd.DataFrame, *args, **kwargs) -> TensorVariable:
-    return pt.as_tensor_variable(df.to_numpy(), *args, **kwargs)
-
+# Optional registrations for DataFrame packages
+def _register_dataframe_backend(library_name: str):
+    try:
+        library = importlib.import_module(library_name)
```
ricardoV94: only this line should be in the try/except

pymc/pytensorf.py
ricardoV94: We should implement a lazy  dispatch system. 

Try `as_tensor_variable`, if that fails, try to import/register pandas, if that fails... try the next one, and so on and so on. So we don't pay the cost of importing pandas/polars/dask when we launch pymc

pymc/pytensorf.py
jessegrabowski: Yeah I discovered that

pymc/pytensorf.py
jessegrabowski: Agreed, and there should be a better way to dispatch it to Narwhals completely without ever having to interact with the underlying libraries. I improved it in the other file and something similar can be done here.

jessegrabowski:
@ricardoV94 @juanitorduz @fonnesbeck I took a 2nd cut at this with some help from a certain special sequence of matrix multiplications. It is *greatly* simplified. Gives all the benefits with none of the multiple dispatch stuff I was doing in the old version.

pymc/pytensorf.py
```diff
@@ -129,10 +126,23 @@ def convert_data(data) -> np.ndarray | Variable:
     return smarttypeX(ret)
 
 
-@_as_tensor_variable.register(pd.Series)
-@_as_tensor_variable.register(pd.DataFrame)
-def dataframe_to_tensor_variable(df: pd.DataFrame, *args, **kwargs) -> TensorVariable:
-    return pt.as_tensor_variable(df.to_numpy(), *args, **kwargs)
+@_as_tensor_variable.register(object)
```
ricardoV94, line 129: The issue with it is that it is not cooperative. What if someone in another package wants to add a fuzzy dispatch for as_tensor_variable for another base type? You overwrote object, only one can take it.

I've thought about doing a numba-like fallback dispatch that iterates over a list of registered functions until one accepts it. This acceptance could actually cause a runtime level dispatch on the types so it only goes there once. Something like (apologies for stupid sniffer terminology the bot came up with):

```python
# ─── pytensor/tensor/__init__.py ──────────────────────────────────────────                                                                                                                                 
  from functools import singledispatch                                                                                                                                                                         
                                                                                                                                                                                                               
  _fallback_sniffers: list = []                                                                                                                                                                                
                                                                                                                                                                                                               
                                                                                                                                                                                                               
  def register_as_tensor_fallback(sniffer):                                                                                                                                                                    
      """Append a cooperative fallback for as_tensor_variable.
                                                                                                                                                                                                               
      Sniffer signature: ``(x, name, ndim, **kwargs) -> dispatch | NotImplemented``.
      Return ``NotImplemented`` if you don't recognize ``x``. Otherwise return a                                                                                                                               
      dispatch callable with the same signature as ``_as_tensor_variable`` that                                                                                                                                
      performs the actual conversion — it will be registered on ``type(x)`` so                                                                                                                                 
      later calls with that type skip the sniff loop entirely.                                                                                                                                                 
                                                                                                                                                                                                               
      Sniffers are tried in registration order; register greedier ones                                                                                                                                         
      (e.g. narwhals ``from_native``) last so more specific handlers win.                                                                                                                                      
      """                                                                                                                                                                                                      
      _fallback_sniffers.append(sniffer)
      return sniffer                                                                                                                                                                                           
                                                                    

  @singledispatch
  def _as_tensor_variable(x, name, ndim, **kwargs):
      for sniff in _fallback_sniffers:                                                                                                                                                                         
          dispatch = sniff(x, name, ndim, **kwargs)
          if dispatch is NotImplemented:                                                                                                                                                                       
              continue                                                                                                                                                                                         
          _as_tensor_variable.register(type(x), dispatch)
          return dispatch(x, name, ndim, **kwargs)                                                                                                                                                             
      raise NotImplementedError(f"Cannot convert {x!r} to a tensor variable.")                                                                                                                                 
                                                                                                                                                                                                               
```
Then here:

```python                                                                                                                                                                                                                    
  # ─── pymc/pytensorf.py ────────────────────────────────────────────────────
  import narwhals as nw                                                                                                                                                                                        
  import pytensor.tensor as pt                                      
  from pytensor.tensor import register_as_tensor_fallback
                                                                                                                                                                                                               
   
  @register_as_tensor_fallback                                                                                                                                                                                 
  def _narwhals_sniffer(x, name, ndim, **kwargs):                   
      nw_x = nw.from_native(x, allow_series=True, pass_through=True)
      if nw_x is x:                                                                                                                                                                                            
          return NotImplemented                                                                                                                                                                                
                                                                                                                                                                                                               
      def dispatch(x, name, ndim, **kwargs):                                                                                                                                                                   
          nw_x = nw.from_native(x, allow_series=True)               
          if isinstance(nw_x, nw.LazyFrame):                                                                                                                                                                   
              nw_x = nw_x.collect()
          return pt.as_tensor_variable(nw_x.to_numpy(), name=name, ndim=ndim, **kwargs)                                                                                                                        
                                                                                                                                                                                                               
      return dispatch     
```

And we can get lazy imports for like scipy sparse (which slows down import time quite a lot and most users don't make use of):

```python
                                                                                                                                                                                                               
  # ─── pytensor/tensor/_sparse_fallback.py ──────────────────────────────────                                                                                                                                 
  from pytensor.tensor import register_as_tensor_fallback                                                                                                                                                      
                                                                                                                                                                                                               
   
  @register_as_tensor_fallback                                                                                                                                                                                 
  def _sparse_sniffer(x, name, ndim, **kwargs):                     
      import scipy.sparse as sps

      if not sps.issparse(x):                                                                                                                                                                                  
          return NotImplemented            

     from pytensor.sparse.foo import _build_from_sparse                         

      def dispatch(x, name, ndim, **kwargs):
          return _build_from_sparse(x, sps, name=name, ndim=ndim, **kwargs)                                                                                                                                    
   
      return dispatch                                                                                                                                                                                          
                                     
```

jessegrabowski: ~~i don't get it, isn't this just multiple dispatch at home?~~

I read the lazy thing more carefully, I get it. But it's not something that could be fixed in this PR?

Also how do I decide to register as a fallback or a dispatch

jessegrabowski: The try/except registration system I had worked fine, I could go back to it. It was just a lot of lines of code, and requires that we explicitly add whichever dataframe backends we want to support.

ricardoV94: > Also how do I decide to register as a fallback or a dispatch

Wdym? there's a register function that's adds it to the list of fallbacks that are attempted. If you return a hit (which is a function) that's the function that gets registered as the dispatch. So first time a narwhals compatible type showsnup here you return the dispatch, then PyTensor registers that function as the usual dispatch. 

The point is you don't gobble up object and (why I had been thinking about it) it can be done lazily like for sparse types.

Code wise needs changes in PyTensor yes. The same strategy would apply to Shared variables where we have the registration happening.

I did not fully get your concerns though

ricardoV94: My thing doesn't need lot's of try except either

jessegrabowski: > Wdym? there's a register function that's adds it to the list of fallbacks that are attempted

I mean as a 3rd party library maintainer, how do I know if I want to decorate with `@_as_tensor_variable.register` or `register_as_tensor_fallback`. I guess a more clear name would be helpful, like `register_lazy_tensor_variable` or something, to emphasize that this is for cases where you don't want to eagerly import something for typing?

ricardoV94: name is terrible agree, as a third party library we know here we want lazy. So I think it kind of answers? This doesn't even need to be public - presented API or anything the like

PR #7967, opened 2025-11-18
display variables in marimo

williambdean:
## Description

This uses the `model_to_mermaid` function for generating graph for a specific variable in the model.

Method bound to: 

- Distribution
- Deterministic
- Potential

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/distributions/distribution.py
+++ b/pymc/distributions/distribution.py
@@ -60,7 +60,7 @@ from pymc.pytensorf import (
     convert_observed_data,
     floatX,
 )
-from pymc.util import UNSET
+from pymc.util import UNSET, safe_display
 from pymc.vartypes import continuous_types, string_types
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


williambdean:
Display a single variable ...

Screenshot 2025-11-18 at 12 54 38 PM

... from a larger model ... 

Screenshot 2025-11-18 at 12 54 20 PM

ericmjl (approved):
LGTM! Do you reckon we'll need to start testing the marimo-related features?

pymc/distributions/distribution.py
```diff
@@ -544,6 +544,10 @@ def __new__(
         rv_out._repr_latex_ = types.MethodType(
             functools.partial(str_for_dist, formatting="latex"), rv_out
         )
+
+        # https://docs.marimo.io/guides/integrating_with_marimo/displaying_objects/#option-1-implement-a-_display_-method
+        rv_out._display_ = types.MethodType(functools.partial(safe_display, model=model), rv_out)
```
ricardoV94, line 549: I think it's a good time as any to remove pretty print from the variables and leave it to the model graphviz only.

There was an issue/ PR to remove those others for ipython. This is never going to be exhaustive because there are many other ways of creating variables for models

ricardoV94:
This goes in the opposite direction I suggest we should take in https://github.com/pymc-devs/pymc/pull/7712

Also graph representation instead of textual by default seems very verbose

pymc/distributions/distribution.py
williambdean: So does that mean, removing other pretty print methods as well? I'm not following your comment fully.

ricardoV94: Yes, only the model remains with pretty print

PR #7973, opened 2025-11-22
Add power_spectral_density to RatQuad covariance kernel for HSGP support

fonnesbeck:
## Description

PSD method and test added. A simple HSGP fits successfully with the resulting RatQuad kernel.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -651,8 +651,12 @@ class RatQuad(Stationary):
 
         z = pt.sqrt(2 * alpha) * pt.sqrt(pt.dot(pt.square(omega), pt.square(ls)))
         coeff = 2.0 * pt.power(2.0 * np.pi * alpha, D / 2.0) * pt.prod(ls) / pt.gamma(alpha)
-        term_z = pt.power(z / 2.0, nu) * pt.kv(nu, z)
-
+        
+        # Handle singularity at z=0
+        safe_z = pt.switch(pt.eq(z, 0), 1.0, z)
+        term_z = pt.power(safe_z / 2.0, nu) * pt.kv(nu, safe_z)
+        term_z = pt.switch(pt.eq(z, 0), pt.gamma(nu) / 2.0, term_z)
+        
         return coeff * term_z
 
 
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -614,6 +614,27 @@ class RatQuad(Stationary):
             -1.0 * self.alpha,
         )
 
+    def power_spectral_density(self, omega: TensorLike) -> TensorVariable:
+        r"""
+        Power spectral density for the Rational Quadratic kernel.
+
+        .. math::
+           S(\boldsymbol\omega) = \frac{2 (2\pi\alpha)^{D/2} \prod_{i=1}^D \ell_i}{\Gamma(\alpha)}
+                                  \left(\frac{z}{2}\right)^{\nu}
+                                  K_{\nu}(z)
+        where :math:`z = \sqrt{2\alpha} \sqrt{\sum \ell_i^2 \omega_i^2}` and :math:`\nu = \alpha - D/2`.
+        """
+        ls = pt.ones(self.n_dims) * self.ls
+        alpha = self.alpha
+        D = self.n_dims
+        nu = alpha - D / 2.0
+
+        z = pt.sqrt(2 * alpha) * pt.sqrt(pt.dot(pt.square(omega), pt.square(ls)))
+        coeff = 2.0 * pt.power(2.0 * np.pi * alpha, D / 2.0) * pt.prod(ls) / pt.gamma(alpha)
+        term_z = pt.power(z / 2.0, nu) * pt.kv(nu, z)
+
+        return coeff * term_z
+
 
 class Matern52(Stationary):
     r"""
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -651,12 +651,10 @@ class RatQuad(Stationary):
 
         z = pt.sqrt(2 * alpha) * pt.sqrt(pt.dot(pt.square(omega), pt.square(ls)))
         coeff = 2.0 * pt.power(2.0 * np.pi * alpha, D / 2.0) * pt.prod(ls) / pt.gamma(alpha)
-        
+
         # Handle singularity at z=0
-    term_z = pt.switch(pt.eq(z, 0), 
-                       pt.gamma(nu) / 2.0, 
-                       pt.power(z / 2.0, nu) * pt.kv(nu, z))
-        
+        term_z = pt.switch(pt.eq(z, 0), pt.gamma(nu) / 2.0, pt.power(z / 2.0, nu) * pt.kv(nu, z))
+
         return coeff * term_z
 
 
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -653,9 +653,9 @@ class RatQuad(Stationary):
         coeff = 2.0 * pt.power(2.0 * np.pi * alpha, D / 2.0) * pt.prod(ls) / pt.gamma(alpha)
         
         # Handle singularity at z=0
-        safe_z = pt.switch(pt.eq(z, 0), 1.0, z)
-        term_z = pt.power(safe_z / 2.0, nu) * pt.kv(nu, safe_z)
-        term_z = pt.switch(pt.eq(z, 0), pt.gamma(nu) / 2.0, term_z)
+    term_z = pt.switch(pt.eq(z, 0), 
+                       pt.gamma(nu) / 2.0, 
+                       pt.power(z / 2.0, nu) * pt.kv(nu, z))
         
         return coeff * term_z
 
--- a/pymc/gp/cov.py
+++ b/pymc/gp/cov.py
@@ -623,6 +623,26 @@ class RatQuad(Stationary):
                                   \left(\frac{z}{2}\right)^{\nu}
                                   K_{\nu}(z)
         where :math:`z = \sqrt{2\alpha} \sqrt{\sum \ell_i^2 \omega_i^2}` and :math:`\nu = \alpha - D/2`.
+
+        Derivation
+        ----------
+        The Rational Quadratic kernel can be expressed as a scale mixture of Squared Exponential kernels:
+
+        .. math::
+            k_{RQ}(r) = \int_0^\infty k_{SE}(r; \lambda) p(\lambda) d\lambda
+
+        where :math:`k_{SE}(r; \lambda) = \exp\left(-\frac{\lambda r^2}{2}\right)` and the mixing distribution
+        on the precision parameter :math:`\lambda` is :math:`\lambda \sim \text{Gamma}(\alpha, \beta)`
+        with rate parameter :math:`\beta = \alpha \ell^2`.
+
+        By the linearity of the Fourier transform, the PSD of the Rational Quadratic kernel is the expectation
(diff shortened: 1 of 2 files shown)


tests/gp/test_cov.py
```diff
@@ -533,6 +533,26 @@ def test_1d(self):
         Kd = cov(X, diag=True).eval()
         npt.assert_allclose(np.diag(K), Kd, atol=1e-5)
 
+    def test_psd(self):
```
jessegrabowski: This test is just test that numpy matches pytensor, since its the same formulas twice. Is there some other way to get a more meaningful test?

pymc/gp/cov.py
```diff
@@ -614,6 +614,27 @@ def full_from_distance(self, dist: TensorLike, squared: bool = False) -> TensorV
             -1.0 * self.alpha,
         )
 
+    def power_spectral_density(self, omega: TensorLike) -> TensorVariable:
+        r"""
+        Power spectral density for the Rational Quadratic kernel.
```
jessegrabowski, line 619: Did you derive this formula or is there a citation we can add?

tests/gp/test_cov.py
fonnesbeck: Open to suggestions here -- I was just following the approach used by the other PSD tests. Do we want to check that you can successfully fit a model with it?

pymc/gp/cov.py
fonnesbeck: Derived, with guidance from Solloch and Williams 2005. Since its a scale mixture of expquads, we can just integrated over the expquad PSD. I will add a little more detail to the docstring.

fonnesbeck:
BTW, one limitation here is that `pytensor.kv` does not appear to support gradient calculation so NUTS can't be fully supported. There is an alternative parameterization that may not need Bessel functions, unless said support is coming.

tests/gp/test_cov.py
jessegrabowski: My understanding isn't super complete, but the PSD is related to the eigenvalules of the full covariance matrix? Can we use that fact (if true) as a check?

fonnesbeck: Yeah, that would work. I assume we'd want to do that for all of the stationary kernels then. Should I do that here, or in a separate PR?

jessegrabowski: Do it here for this test, then we can open a new PR for the others

jessegrabowski:
> BTW, one limitation here is that pytensor.kv does not appear to support gradient calculation so NUTS can't be fully supported. There is an alternative parameterization that may not need Bessel functions, unless said support is coming.

Does jax or tfp have the gradients we could use as the basis for an implementation?

fonnesbeck:
Apparently not. 

image

Looks like a numerical approximation (via finite differences) is as close as you can get.

jessegrabowski:
They have an expression for the 2nd input: https://github.com/tensorflow/probability/blob/732e7d16235a38b639a3b2cd00e76a4add45118c/tensorflow_probability/python/math/bessel.py#L1057

ricardoV94:
In the past I've implemented a couple gradients using stan math / wolfram as reference. Never had a case where tfp was ahead of the curve for these special functions

ricardoV94:
Gradient wrt second input is the trivial one in these functions, we also have it

fonnesbeck:
There has been some useful discussion on Jax: https://github.com/jax-ml/jax/issues/12402

fonnesbeck:
FWIW, it does work without a fully differentiable function -- it just uses a slice sampler for alpha.

[ratquad_hsgp.html](https://github.com/user-attachments/files/23733399/ratquad_hsgp.html)

tests/gp/test_cov.py
bwengals: Yeah that's a much better test idea.  Blame me for the lame tests that numpy match pytensor

tests/gp/test_cov.py
```diff
@@ -533,6 +534,56 @@ def test_1d(self):
...
+        test_1d_psd = (
+            pm.gp.cov.RatQuad(1, ls=ell, alpha=alpha)
+            .power_spectral_density(omega[:, None])
+            .flatten()
+            .eval()
+        )
+        npt.assert_allclose(true_1d_psd, test_1d_psd, atol=1e-5)
+
+    def test_psd_eigenvalues(self):
+        # Test PSD implementation using Szegő’s Theorem
```
bwengals: Will this work for a 2 or higher dimensional X?

tests/gp/test_cov.py
```diff
@@ -533,6 +534,56 @@ def test_1d(self):
...
+        L = 50.0
+        dx = L / N
+        X = np.linspace(0, L, N)[:, None]
+
+        with pm.Model():
+            cov = pm.gp.cov.RatQuad(1, alpha=alpha, ls=ls)
+
+        K = cov(X).eval()
+
+        evals = scipy.linalg.eigvalsh(K)
```
jessegrabowski:
```suggestion
        eigs = scipy.linalg.eigvalsh(K)
```
or `eig_vals` -- I thought that this variable was samples (like "evaluations") and I was wondering why you were doing sampling

tests/gp/test_cov.py
fonnesbeck: [It does](https://github.com/user-attachments/files/23736136/ratquad_hsgp_2d.html)

pymc/gp/cov.py
```diff
@@ -614,6 +614,51 @@ def full_from_distance(self, dist: TensorLike, squared: bool = False) -> TensorV
...
+        D = self.n_dims
+        nu = alpha - D / 2.0
+
+        z = pt.sqrt(2 * alpha) * pt.sqrt(pt.dot(pt.square(omega), pt.square(ls)))
+        coeff = 2.0 * pt.power(2.0 * np.pi * alpha, D / 2.0) * pt.prod(ls) / pt.gamma(alpha)
+        
+        # Handle singularity at z=0
+        safe_z = pt.switch(pt.eq(z, 0), 1.0, z)
+        term_z = pt.power(safe_z / 2.0, nu) * pt.kv(nu, safe_z)
+        term_z = pt.switch(pt.eq(z, 0), pt.gamma(nu) / 2.0, term_z)
```
jessegrabowski:
```suggestion
    term_z = pt.switch(pt.eq(z, 0), 
                       pt.gamma(nu) / 2.0, 
                       pt.power(z / 2.0, nu) * pt.kv(nu, z))
```

Switch is elementwise, so it's safe to just have a single check.

tests/gp/test_cov.py
```diff
@@ -533,6 +534,56 @@ def test_1d(self):
...
+
+        test_1d_psd = (
+            pm.gp.cov.RatQuad(1, ls=ell, alpha=alpha)
+            .power_spectral_density(omega[:, None])
+            .flatten()
+            .eval()
+        )
+        npt.assert_allclose(true_1d_psd, test_1d_psd, atol=1e-5)
+
+    def test_psd_eigenvalues(self):
```
jessegrabowski, line 536: I was trying to cook up something more accurate, and gippity came up with Rayleigh quotients. It seems to be more accurate than using `linalg.eigh`. I also found that a smaller length scale increased accuracy, I guess because we can lay down more gridpoints more densely in a small region. Cranking up N (or decreasing dx) did not seem to help.

```py
import numpy as np
from scipy import linalg

alpha = 1.5
ls = 0.1
L = 50
dx = 0.05
N = int(L / dx)
X = (np.arange(N) * dx)[:, None]

with pm.Model():
    cov = pm.gp.cov.RatQuad(1, alpha=alpha, ls=ls)

K = cov(X).eval()

theta = 2 * np.pi * np.fft.fftfreq(N, d=1)

omegas = theta / dx
psd = cov.power_spectral_density(omegas[:, None]).eval()
psd_scaled = psd.ravel() / dx

j = np.arange(N)
modes = np.exp(1j * np.outer(theta, j))
num = modes.conj() @ (K @ modes.T)
den = np.sum(np.abs(modes)**2, axis=1)
rayleigh_quotient = np.real(np.diag(num) / den)


np.testing.assert_allclose(psd_scaled, rayleigh_quotient, atol=0.05)
```

pymc/gp/cov.py
```diff
@@ -614,6 +614,51 @@ def full_from_distance(self, dist: TensorLike, squared: bool = False) -> TensorV
...
+
+        Substituting the known PSD of the Squared Exponential kernel and evaluating the integral yields
+        the expression involving the modified Bessel function of the second kind, :math:`K_{\nu}(z)`.
+        """
+        ls = pt.ones(self.n_dims) * self.ls
+        alpha = self.alpha
+        D = self.n_dims
+        nu = alpha - D / 2.0
+
+        z = pt.sqrt(2 * alpha) * pt.sqrt(pt.dot(pt.square(omega), pt.square(ls)))
```
jessegrabowski, line 652:
```suggestion
        z = pt.sqrt(2 * alpha) * pt.sqrt(pt.outer(pt.square(omega), pt.square(ls)))
```
A bit more readable?

tests/gp/test_cov.py
jessegrabowski: I would just remove this test now

jessegrabowski (changes requested):

tests/gp/test_cov.py
jessegrabowski: Actually, plotting the approximation, it seems like all of the error comes from the boundaries. So it might make more sense to do a test like:

```py

trim = 100
np.testing.assert_allclose(psd_scaled[trim:-trim], rayleigh_quotient[trim:-trim], atol=1e-2)

```

Plot:

```py
fig, ax = plt.subplots()
order = np.argsort(theta)

ax.plot(theta[order], psd_scaled[order], label="Closed-Form")
ax.plot(theta[order], Rq[order], "--", label="Rayleigh quotient")
ax.set(xlabel = "θ", ylabel='spectral density', yscale='log')

plt.legend()
```

image

pymc/gp/cov.py
fonnesbeck: Replacing an inner product with an outer product?

jessegrabowski: I'm known for my big brain takes

fonnesbeck: OK good, I had to read it about 7 times to make sure I wasn't misreading something! :sweat_smile:

jessegrabowski: Yeah sorry. I only tested the 1d guess where they're the same.

tests/gp/test_cov.py
fonnesbeck: We can do this. Did you come across a reference for this to add to the docstring?

jessegrabowski: I think we can just cite the wiki for the formula: https://en.wikipedia.org/wiki/Rayleigh_quotient (or use one of the citations therein)

tests/gp/test_cov.py
jessegrabowski: I still think this test has no value and should be removed.

jessegrabowski (approved):
look good, but zizmor is failing (never seen that before) and I'd still recommend removing the first test, since it is just a test of pytensor math implementations, not the code in this PR.

jessegrabowski (approved):
Looks great. Thanks for improving the test. Could you make an issue to improve the others if you haven't already

@maresb do you know why zizmor is failing?

PR #7974, opened 2025-11-22
Add core notebook demonstrating vectorized/shaped variables (Fixes #4007)

### LINKED ISSUE #4007: Documentation: Create a tutorial demonstrating vector variables
If you have questions about a specific use case, or you are not sure whether this is a bug or not, please post it to our discourse channel: https://discourse.pymc.io  

Hi all, with some frequency I have multiple groups of data that I want to model at the same time.

Here is a small example:

```python
import numpy as np
import pymc3 as pm
from scipy.stats import norm

num_groups = 5
group_size = 200
sigma = 1
data = np.concatenate([norm.rvs(loc=mu, scale=sigma, size=group_size) for mu in range(num_groups)])
data_labels = np.concatenate([np.ones(group_size) * group_id for group_id in range(num_groups)]).astype(int)

with pm.Model() as model:
    mu = pm.HalfNormal("mu", sigma=10, shape=num_groups)
    sigma = pm.HalfNormal("sigma", sigma=20, shape=num_groups)

    likelihood = pm.Normal("data", mu=mu[data_labels], sigma=sigma[data_labels], observed=data)
```

Every time I do this (assuming it is the right thing to do), I forget how to setup the observed data and slicing the RVs. I'll eventually find an old example of mine, but can't readily find PyMC3 documentation to help me out. Shaped/vectored RVs are mentioned in the [Getting Started Tutorial](https://docs.pymc.io/notebooks/getting_started.html#Model-Specification), but not to the extent I've shown here. In the past, I think I've learned some from the [Rugby Example](https://docs.pymc.io/notebooks/rugby_analytics.html) or Discourse etc.

If it makes sense to others, I'd like to see some tutorial/documentation that explicitly details this usage. It may already exist, but I haven't been able to search for it effectively, so some change could be made to make it easier to find. This may also be due to the fact that I'm not sure if these would be called vector variables, shaped variables etc.

Abhishek6122008:
## Description
This PR adds a new core notebook `vectorized_variables.ipynb` under  
`docs/source/learn/core_notebooks/`.

The notebook provides a clear and concise tutorial demonstrating how to work
with vectorized and shaped random variables in PyMC. It addresses the request
in Issue #4007 by showing:

- How to define RVs with shapes / dims / coords
- How to slice and index vector-valued parameters (e.g., `mu[data_labels]`)
- How broadcasting works in PyMC models
- How to model multiple groups simultaneously using vectorized priors
- A small end-to-end example model that mirrors the use case described in #4007

The goal is to give users a direct, accessible reference for one of the most
common modeling patterns in PyMC: handling multiple groups with vector-valued
parameters.

## Related Issue
- Closes #4007

## Checklist
- [x] Added new example notebook
- [x] Confirmed the notebook renders correctly in the docs
- [x] Checked pre-commit linting
- [ ] Tests not required for documentation notebooks

## Type of change
- [x] Documentation (new example notebook)


Ramji-Purwar:
can i do this

PR #7979, opened 2025-11-30
Fixed UserWarning when converting sample_stats to idata

### LINKED ISSUE #7821: sample_smc conversion of sample stats to idata can result in warning
### Description

Sometimes we have less sample stats than chains with SMC (the concept of `draws` is hacked a bit), and this line here will trigger an arviz warning: https://github.com/pymc-devs/pymc/blob/cc902124ed08fe8c3007699451cc0c16f38520d2/pymc/smc/sampling.py#L266-L270

Reported by an user in https://discourse.pymc.io/t/incorrect-number-of-draws-in-smc-example/17119/2

We should suppress that warning, (perhaps in all internal uses of `dict_do_dataset` by PyMC)?

CC @aloctavodia @OriolAbril

MSK-005:
## Description
This PR aims to close [#7821](https://github.com/pymc-devs/pymc/issues/7821). Previously, in some cases, running `pm.sample_smc()` resulted in a UserWarning being generated: 

> UserWarning: More chains (5) than draws (1). Passed array should have shape (chains, draws, *shape)   warnings.warn(

What I did was update the `dict_to_dataset()` function call(s) in `pymc/backends/arviz.py` and `pymc/smc/sampling.py`. In `pymc/backends/arviz.py`, I changed the import statement so that `dict_to_dataset()` was imported from `arviz_base` rather than `arviz.data.base`, and the rest of the changes just involved updating the function call. Another important detail is that in `pymc/smc/sampling/py`, I added another argument when calling `dict_to_dataset`, which is `sample_dims=["chain"]`, since the variable `sample_stats_dict` was a 1-D dict.

## Related Issue

- [x] Closes [#7821](https://github.com/pymc-devs/pymc/issues/7821)
- [ ] Related to #

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):


MSK-005:
Can anyone explain how to fix these checks? `pre-commit` passes all, but here, one test is failing.

williambdean:
Both of the requirements.txt files should not be manually edited (see comment at top of the file). If you need to add dependencies, change conda-envs/environment-dev.yml

ricardoV94:
I'm not sure what's the status of arviz-base and if it's stable enough to switch as a dependency. This would probably require a major PyMC version bump as well?

CC @OriolAbril and @aloctavodia

aloctavodia:
In general we could begin a migration to "New ArviZ", but to reduce friction, we should do it in a coordinated manner and change arviz imports everywhere, not just here.  It may be a good idea to have a dedicated PyMC release to work on the transition.

In particular, about this PR, `from_dict` returns a DataTree object, not an InferenceData object. DataTree is a native xarray object that effectively replaces InferenceData in ArviZ. Many functions (all that make sense?) in ArviZ still understand InferenceData objects, but only for backward compatibility.

OriolAbril:
I will take a closer look in the coming days but it looks good. We can have both arviz and arviz-base installed together without problem  and I think doing that doesn't require a major version because for the most part, users won't really notice any change.

In this case it is only a switch between the two `dict_to_dataset` whose only difference are a couple function arguments, that users can't change anyway.

The main things to comment are trying to switch as many imports as possible to arviz-base instead of combining arviz-base and arviz.data imports in the same file. And requiring arviz-base newer than 0.7

MSK-005:
I'll wait for Mr Oriol's comments in the coming days. If you guys feel like it's too early to shift to `arviz-base`, I can revert the changes and wrap the function call in an ignore-warnings context. Thank you, all of you, for your time.

MSK-005:
Hi, @OriolAbril. Before I begin making further changes, I'd really like your thoughts on the way this PR is going.

Thanks a lot!

MSK-005:
@ricardoV94, I understand that `arviz_base` works with `DataTree` and therefore, I've only changed some of the imports so that no functionality is broken. I've run the tests locally, and all are passing. The original `UserWarning` has been fixed, but with the import of `arviz_base`, there is another `UserWarning`. I'm not fixing that as it relates to `InferenceData` objects, which will become irrelevant when we shift to the refactored `arviz` packages, for which a separate PR must be opened.

CC: @OriolAbril

OriolAbril (approved):
LGTM, this only affects `dict_to_dataset` so outputs continue to be datasets. Later on we can change InferenceData to DataTree as they are collections of datasets

MSK-005:
Hi, @ricardoV94. Could you please approve the the running of the core CI tests? I'm a first-time contributor, so the CI tests don't run unless approved by a core maintainer. Apologies for the inconvenience.

I ran `test_arviz`, `mypy`, and `test_smc` locally and they all pass.

ricardoV94:
The PR needs to be rebased from main as it is picking commits that were already merged and makes it hard to see the changes that are specific to it.

Also how does this compare with #8019 ?

MSK-005:
@ricardoV94, done rebasing.

As for the mentioned PR #8019, I'm using the individual ArviZ library, arviz-base, but ArviZ 1.0 serves as an umbrella for these individual libraries (arviz-base, arviz-plots, and arviz-stats) anyway, as can be seen in their code. I'd say the better move would be to close this PR (don't merge it) and move ahead with their work, since their PR has a greater scope and uses a single public dependency library. I've seen that PR's code, and it has already implemented the changes that my PR does.

ricardoV94:
@MSK-005 you still say we should close this in favor of #8019 ?

MSK-005:
@ricardoV94, yes. His PR already covers all the code I wrote and only uses arviz as a dependency package, whereas I've used 3 (arviz-plots, arviz-base, and arviz-stats).

ricardoV94:
Thanks for the initiative and checking the other PR

PR #7987, opened 2025-12-03
Consolidate typing for coords with Coords and StrongCoords aliases (fixes #7972)

### LINKED ISSUE #7972: Consolidate typing for `coords`
coords have a mix of types in the codebase. Some places: 

https://github.com/pymc-devs/pymc/blob/4ad7fa8f88525fae096886fb7a7b19247a7c5176/pymc/model/core.py?plain=1#L458

The conversion type: 

https://github.com/pymc-devs/pymc/blob/4ad7fa8f88525fae096886fb7a7b19247a7c5176/pymc/model/core.py?plain=1#L911-L914

https://github.com/pymc-devs/pymc/blob/4ad7fa8f88525fae096886fb7a7b19247a7c5176/pymc/model/core.py?plain=1#L1065

https://github.com/pymc-devs/pymc/blob/4ad7fa8f88525fae096886fb7a7b19247a7c5176/pymc/backends/arviz.py?plain=1#L126

Seems to be a mix of `Sequence` and `Sequence | np.ndarray` so some
consolidation would be nice. For example, there is already a various amount of
typing for other objects like `dims` and `shape`:

https://github.com/pymc-devs/pymc/blob/4ad7fa8f88525fae096886fb7a7b19247a7c5176/pymc/distributions/shape_utils.py?plain=1#L88-L98

My main concern is not within `pymc` package directly but in the downstream
usage where using `dict[str, Any]` is probably the most common way to type
`coords`. So exposing would have some typing benefit.

aman-coder03:
this PR addresses issue #7972 by consolidating the typing of `coords` across the PyMC codebase.

Previously, `coords` appeared with a mix of types such as:
- `Sequence`
- `Sequence | np.ndarray`
- `dict[str, Any]`

this PR introduces explicit and consistent type aliases:
- `CoordValue`
- `Coords`
- `StrongCoordValue`
- `StrongCoords`

(similar to the existing `Shape`, `Dims`, and `StrongDims` aliases in `shape_utils`)

### Key changes
- standardizes `Model.coords` to return `StrongCoords`
- updates ArviZ conversion to return `StrongCoords` instead of `dict[str, Any}`
- removes the mixed use of `Sequence`, `Sequence | np.ndarray`, and untyped dictionaries
- improves static typing and downstream library compatibility

### Notes
while testing this change, a runtime circular import involving:
`pymc.model.core`
`pymc.distributions.shape_utils`
`from pymc.model import modelcontext`
`pymc.model` (partially initialized)
was encountered.

Fixes #7972


pymc/distributions/shape_utils.py
```diff
@@ -164,21 +170,27 @@ def convert_size(size: Size) -> StrongSize | None:
...
-    dims : tuple of (str or None)
-        Names or None for all RV dimensions.
+    shape : tuple
+        Shape inferred from model dimension lengths.
     """
+    # Lazy import to break circular dependency
+    if model is None:
+        from pymc.model.core import modelcontext
+
+        model = modelcontext(None)
```
ricardoV94: Why this functional change? Doesn't seem necessary or a good idea

pymc/distributions/shape_utils.py
```diff
@@ -97,6 +97,12 @@ def _check_shape_type(shape):
 StrongDimsWithEllipsis: TypeAlias = Sequence[str | EllipsisType]
 StrongSize: TypeAlias = TensorVariable | tuple[int | Variable, ...]
 
+CoordValue: TypeAlias = Sequence[Hashable] | np.ndarray | None
+Coords: TypeAlias = Mapping[str, CoordValue]
+
+StrongCoordValue: TypeAlias = tuple[Hashable, ...] | None
+StrongCoords: TypeAlias = Mapping[str, StrongCoordValue]
```
ricardoV94: Put these in a more independent place to avoid the new circular import issues. Coordinates are not really distribution specific anyway

aman-coder03: i can move `CoordValue`, `Coords`, `StrongCoordValue`, and `StrongCoords `to a more independent location and update downstream imports accordingly. Let me know if you have a preferred target module for these.

pymc/distributions/shape_utils.py
aman-coder03: i introduced the lazy import only to work around the circular import that surfaced during the typing consolidation (`shape_utils -> model -> distributions -> shape_utils`). I agree that changing runtime behavior here is not ideal
if you prefer, i can revert this functional change and instead resolve the circular dependency purely by relocating the typing aliases to a more independent module

aman-coder03:
@ricardoV94 it would be really helpful if you could review the changes!

pymc/typing.py
```diff
@@ -0,0 +1,33 @@
...
+from __future__ import annotations
+
+from collections.abc import Hashable, Mapping, Sequence
+from typing import TypeAlias
+
+import numpy as np
+
+# -------------------------
+# Coordinate typing helpers
+# -------------------------
```
ricardoV94: Remove comment

ricardoV94:
I'll ping some folks who are bigger fans of type-hints than me for review. 

A whole new file seems a bit cluttering, although it could be useful to move more things in the future. My gut feeling would be to just throw this into one of the `utils.py` files for now.

aman-coder03:
@ricardoV94 Thanks for the feedback!! I have implemented the changes!!

williambdean:
Thoughts on a `pymc.typing` module? 

```python
from pymc.typing import Coords
import pymc as pm


coords: Coords = {"covariate": ["A", "B", "C"]}
model = pm.Model(coords=coords)
```

Or directly from `pm`

Seems out of place in the `utils` module.

ricardoV94:
What would go in it? If it's just a handful doesn't seem worth it

aman-coder03:
Thanks for the feedback! @ricardoV94 @williambdean 

Right now it’s just the coord/shape aliases (Coords, CoordValue, StrongCoords, etc.).
So I agree it might feel too small to justify a standalone pymc.typing module at the moment.

Reason for grouping them was:
They do not really fit semantically into `util.py`
They are used across several modules (`model`, `backends`, `shape_utils`)
A dedicated namespace keeps things cleaner as more typing-related additions come in later

But I'm also happy to keep everything inside `pymc.util`
Let me know which direction you would like, and I will update the PR accordingly.

aman-coder03:
@ricardoV94 @williambdean i have implemented some changes, it would be really helpful if you take a look!!

pymc/distributions/shape_utils.py
```diff
@@ -6,20 +6,22 @@
 #
 #       http://www.apache.org/licenses/LICENSE-2.0
 #
-#   Unless required by applicable law or agreed to in writing, software
+#   Unless required by applicable law or deemed in writing, software
```
michaelosthege: no unrelated changes, please
```suggestion
#   Unless required by applicable law or agreed to in writing, software
```

pymc/distributions/shape_utils.py
```diff
@@ -164,7 +164,7 @@ def convert_size(size: Size) -> StrongSize | None:
         )
 
 
-def shape_from_dims(dims: StrongDims, model) -> StrongShape:
+def shape_from_dims(dims: StrongDims, model: Model) -> StrongShape:
```
michaelosthege: This is the only function taking `model: Model` in this file, and it's causing the circular import.

In my opinion this indicates that the function should not be defined here.
It also already defined as a model method:
[`Model.shape_from_dims`](https://github.com/pymc-devs/pymc/blob/05c93329db000dfe5b5f9bd6b9ebea6bbf5b01f9/pymc/model/core.py#L922)

The one place where it is used (see [search results](https://github.com/search?q=repo%3Apymc-devs%2Fpymc%20shape_from_dims&type=code)) can be changed to `model.shape_from_dims(...)`.

Therefore I suggest to:
* not type hint `Model` here
* and instead delete the function or make it `raise DeprecationWarning`

pymc/util.py
```diff
@@ -16,21 +16,51 @@
...
+
+# User-provided dims that may include ellipsis (...)
+DimsWithEllipsis: TypeAlias = str | EllipsisType | Sequence[str | None | EllipsisType]
+
+# User-provided size before processing
+Size: TypeAlias = int | TensorVariable | Sequence[int | Variable]
+
+# Strong / normalized versions used internally
+StrongDimsWithEllipsis: TypeAlias = Sequence[str | EllipsisType]
+StrongSize: TypeAlias = TensorVariable | tuple[int | Variable, ...]
```
michaelosthege: A `"""` string below the alias definitions might be useful to explain the subtle differences.

michaelosthege (commented):
I'm in favor of a `pymc.typing` module just like `numpy.typing`.

It may seem small at first, but `pymc.util` is already quite big.

aman-coder03:
can you please guide me on the basis of the latest commit? @michaelosthege

pymc/distributions/distribution.py
```diff
@@ -522,7 +521,7 @@ def __new__(
         # finally, observed, to determine the shape of the variable.
         if kwargs.get("size") is None and kwargs.get("shape") is None:
             if dims is not None:
-                kwargs["shape"] = shape_from_dims(dims, model)
+                kwargs["shape"] = model.shape_from_dims(dims)
```
michaelosthege: Looks like `model.shape_from_dims` returns the non-symbolic lengths of entries in the `Model.coords` whereas `pm.distributions.shape_utils.shape_from_dims` returns the symbolic dim lengths originating from `model.dim_lengths`.

Unfortunate naming..

Nevertheless I would define `shape_from_dims` on the `Model`, but with a more informative name. Maybe `symbolic_shape_from_dims`?

pymc/util.py
```diff
@@ -16,21 +16,62 @@
...
+Type alias groups:
+
+- User-facing types (Coords, CoordValue, Shape, Dims, Size) represent the flexible inputs
+  accepted from users when defining models.
+
+- Strong/internal types (StrongCoords, StrongDims, StrongShape, StrongSize, etc.) represent
+  normalized, fully-validated forms used internally after processing.
+
+These distinctions allow looser user input while keeping strict internal consistency.
+"""
```
michaelosthege: I meant individual docstrings for each type alias.

They should be only one line each and capture exactly what the type is about.
__Please do not use AI for this.__ It just gives bullshit results.

aman-coder03: sorry for the confusion!! @michaelosthege

pymc/util.py
```diff
@@ -16,21 +16,59 @@
...
+from pytensor.graph.basic import Variable
+from pytensor.tensor.variable import TensorVariable
 
 from pymc.exceptions import BlockModelAccessError
 
+CoordValue: TypeAlias = Sequence[Hashable] | np.ndarray | None
+# User-provided values for a single coordinate dimension.
+
+Coords: TypeAlias = Mapping[str, CoordValue]
+# Mapping from dimension name to its coordinate values.
```
OriolAbril: I am definitely not a type expert and I always struggle because I often see a trade off between clarity, ensuring the type hint doesn't allow clearly wrong types as input and ensuring all possible valid types are covered by the type hints.

For coords, I see we are using `Coords` type alias as the type hint for the `coords` argument in the `Model` class. However, `None` is not valid through `pm.Model(coords=...)`, it is only valid if when using `model.add_coords()` we pass something other than `None` as the `lengths` argument. When initializing a model, `coords` are passed to `.add_coords()` with no argument, so using `None` as a coord value will always raise an error.

This one is way more obscure and might not have many (or any) actual use. But it is fine to have non hashable values inside the coordvalue sequence as long as numpy doesn't struggle to construct an `object` dtype array out of the sequence.

That is `pm.Model(coords={"name": [["a"], "b"]})` errors out when converting to inferencedata because we convert the coord values to an array which fails with `ValueError: setting an array element with a sequence.`

But `pm.Model(coords={"name": [{"a": "c"}, "b"]})` works even though `{"a": "c"}` is not hashable. Moreover, if passed as `pm.Model(coords={"name": np.array([{"a": "c"}, "b"])})` it would match the `CoordValue` type from the ndarray side, so it is a bit confusing for me.

pymc/distributions/shape_utils.py
```diff
@@ -33,18 +35,29 @@
...
+Shape = int | TensorVariable | Sequence[int | Variable]
+Dims = str | Sequence[str | None]
+DimsWithEllipsis = str | EllipsisType | Sequence[str | None | EllipsisType]
+Size = int | TensorVariable | Sequence[int | Variable]
+
+# Strong (validated) types from util
+
+# Additional strong types needed inside this file
+StrongDimsWithEllipsis = Sequence[str | EllipsisType]
+StrongSize = TensorVariable | tuple[int | Variable, ...]
```
OriolAbril: why are we duplicating all of these instead of importing from `util.py`?

pymc/util.py
aman-coder03: Thanks @OriolAbril , this makes sense. I agree that `CoordValue `is inherently a compromise: `None `is only valid in the `add_coords `pathway, and hashability isn’t the real constraint but NumPy array construction is. 
My goal here was to document the intended usage rather than fully encode all runtime constraints, but I’m happy to adjust the aliases or their docstrings to better reflect these nuances if you think that would improve clarity for future contributors.

pymc/distributions/shape_utils.py
aman-coder03: Good point @OriolAbril  this was left over from when the aliases lived in `pymc.typing` I’ll switch `shape_utils.py` to import the shared aliases from `pymc.util` and remove the local redefinitions so there is a single source of truth.

pymc/util.py
```diff
@@ -16,21 +16,59 @@
...
 from cachetools import LRUCache, cachedmethod
-from pytensor import Variable
 from pytensor.compile import SharedVariable
+from pytensor.graph.basic import Variable
+from pytensor.tensor.variable import TensorVariable
 
 from pymc.exceptions import BlockModelAccessError
 
+CoordValue: TypeAlias = Sequence[Hashable] | np.ndarray | None
+# User-provided values for a single coordinate dimension.
```
michaelosthege: Also on the others, this is a docstring for a variable:
```suggestion
"""User-provided values for a single coordinate dimension."""
```

aman-coder03: done the changes!!! @michaelosthege

aman-coder03:
@ricardoV94 @williambdean can we merge the PR, if there are no additional changes?

OriolAbril:
one of my comments is still unresolved, but for me it is not a blocker. I am by no means a type hint expert (or even very good user of type hints) and it is probably fine as it is even if there are a couple known edge cases of the type hint saying a type will work and it doesn't or viceversa.

michaelosthege:
And we require all tests to pass

PR #7988, opened 2025-12-05
Clarify ZeroSumNormal covariance documentation

### LINKED ISSUE #7904: DOC: Covariance matrix for ZeroSumNormal incorrect?

The [docs](https://github.com/pymc-devs/pymc/blob/9f653a6ebcc4a205b1f506e2f3e513753d718b28/pymc/distributions/multivariate.py#L2706C1-L2712C21) state that `ZSN(sigma) = N(0, sigma^2 (I - (1/n)J)` where `n` is the nbr of zero-sum axes".

It had been my understanding that the marginal variance of any one dimension of a ZSN variable would be `sigma^2 (1 - 1/K)`. And the covariance between any two (different) dimensions would be, I think, `-sigma^2/K`. (This is supposing K dimensions and just 1 zero-sum axis.)

I could be wrong, but to the best of my understanding I think that this indicates that 1/K should be the scaling term on J, not 1/n. (And then the notation will get messier if expressing multiple zero-sum axes.)

I'd be glad to submit a PR on the matter, but wanted to confirm the issue first, as I'm not sure I 100% understand the statistical machinery going on for pymc's ZSN implementation.

_No response_

swarnaliiiiii:
Fixes: #7904

This PR clarifies the documentation for `ZeroSumNormal` to use K (size of the constrained axis)
instead of n (number of zero-sum axes), as discussed in the issue.

It updates the docstring math expression:

    ZSN(σ) = N(0, σ² (I_K - 1/K J_K))

No functional code changes — documentation only.

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -2705,31 +2705,31 @@ class ZeroSumNormal(Distribution):
 
     .. math::
 
-    \begin{align*}
-    ZSN(\sigma) = N \Big( 0, \sigma^2 (I_K - \tfrac{1}{K}J_K) \Big) \\
-    \text{where} \ ~ J_{ij} = 1 \ ~ \text{and} \\
-    K = \text{size (length) of the constrained axis}
-    \end{align*}
+        \begin{align*}
+            ZSN(\sigma) = N \Big( 0, \sigma^2 (I_K - \tfrac{1}{K}J_K) \Big) \\
+            \text{where} \ ~ J_{ij} = 1 \ ~ \text{and} \\
+            K = \text{size (length) of the constrained axis}
+        \end{align*}
 
     Parameters
     ----------
     sigma : tensor_like of float
-    Scale parameter (sigma > 0).
-    It's actually the standard deviation of the underlying, unconstrained Normal distribution.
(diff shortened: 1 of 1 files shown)


pymc/distributions/multivariate.py
```diff
@@ -2700,60 +2700,34 @@ class ZeroSumNormal(Distribution):
...
+    - :math:`J_K` is the :math:`K \times K` matrix of ones,
+    - :math:`K` is the size of the constrained axis.
+
+    Using :math:`K` avoids confusion with ``n_zerosum_axes``, which counts how many
+    axes are constrained, not their length.
 
     Parameters
     ----------
     sigma : tensor_like of float
-        Scale parameter (sigma > 0).
```
ricardoV94, line 2717: Should not remove any of this previous stuff

swarnaliiiiii:
Updated the docstring to use K (size of the constrained axis) instead of n in the covariance formula. This corrects the notation as discussed in issue #7904. No other text was removed or changed.

ricardoV94:
There are clearly other changes, which you can confirm by clicking on 
image

swarnaliiiiii:
The math block has been updated to use K (size of the constrained axis) instead of n, with all previous documentation left intact. Only a part of the formula was replaced.

lucianopaz (approved):
Lgtm

PR #7991, opened 2025-12-10
Add background sampling handle to pm.sample

### LINKED ISSUE #7929: Non-blocking sampling
### Description

Nutpie allow sampling in a non-blocking thread, which is great for interactive workflows. You can start playing with plotting - summary code while the model is sampling. We should try to enable that for PyMC samplers as well, running it in a distinct thread from main?

CC @lucianopaz @aseyboldt

XPE-7:
## Description

This adds an opt-in background mode to `pm.sample` so users can start sampling and continue working while it runs in a background thread. When `background=True`, `pm.sample` returns a handle with `done()`, `result()`, and `exception()` helpers, and the progress bar is suppressed to keep output clean. Background mode is currently limited to the built-in `nuts_sampler="pymc"`; other samplers raise `NotImplementedError` to make the scope clear. The model is resolved eagerly so the background thread always has a valid context.

Docstring notes describe how to use `background=True` and the current limitations. Tests cover both a successful background run and error propagation (`tests/sampling/test_background_sampling.py`).

Summary of changes:
- Add `background` flag to `pm.sample` and return a `BackgroundSampleHandle` when enabled.
- Suppress progress bars in background mode.
- Resolve the model before launching the background thread.
- Document background mode in the `pm.sample` docstring.
- Add tests for background sampling (happy path and error propagation).

## Related Issue
- [x] Closes #7929

## Checklist
- [ ] Checked that pre-commit linting/style checks pass
- [ ] If you are a pro: each commit corresponds to a relevant logical change

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,6 +85,7 @@ jobs:
             tests/sampling/test_deterministic.py
             tests/sampling/test_forward.py
             tests/sampling/test_population.py
+            tests/sampling/test_background_sampling.py
             tests/stats/test_convergence.py
             tests/stats/test_log_density.py
             tests/distributions/test_distribution.py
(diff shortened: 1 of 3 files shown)


XPE-7:
Hi! This PR adds the opt-in background sampling handle to `pm.sample` (see #7929). Pre-commit and docs are green. Could a maintainer please approve the pending workflows (`all_tests`, `mypy`) so they can run, and review when you have a moment. Thanks!

ricardoV94:
No point in running the tests before checking the approach. This issue probably needed some more planning before jumping in

XPE-7:
Thanks @ricardoV94 for the feedback, that actually makes sense, especially since this change affects a core part of `pm.sample`.
My intention was to start with a very small and conservative step: an opt-in, thread-based background mode that doesn’t change existing behavior, avoids multiprocessing, and is limited to `nuts_sampler="pymc"` for now. The idea was to keep the scope narrow and the impact contained while addressing the original use case.
I agree this kind of change would benefit from a design discussion first. Before going further, I’d really appreciate guidance on whether this general direction (background execution with a handle-based API) fits PyMC’s expectations, or if there is any different approach you would prefer to be explored.

PR #7995, opened 2025-12-15
Add logprob support for leaky-ReLU switch transforms

### LINKED ISSUE #7543: Derive logprob of Leaky ReLU transform
### Description

```python
import pymc

a = 0.5
x = pm.Normal.dist()
y = pm.math.switch(x > 0, x, a * x)
pm.logp(y, 2.3).eval()  # NotImplementedError
```

We already have a logprob derivation for mixture switches where the condition is constant, but not if it depends on the same measurable variable that is in the branches. This is not a mixture but an invertible transform.

We could support an arbitrary functions on both branches as long as the domains retain the same sign after the transformation (so that it's easy to invert). To respect this, the leaky ReLU actually requires `a` to be positive, so a runtime check may be needed.

We could actually support arbitrary cutoff points, but it becomes increasingly tricky to figure out which branch to go down when inverting the graph.

In any case, because it is such a common transformation, it would great to at least support the special case of the leaky ReLu.

eclipse1605:
## Description
added log-probability support for leaky-ReLU graphs constructed as

    y = switch(x > 0, x, a * x)

where `x` is a single continuous measurable variable.

### notes
- only supports a single continuous measurable variable.
- the slope `a` must be non-measurable and strictly positive.
- behavior at `y == 0` follows the `y <= 0` branch (measure-zero set).

## Related Issue
- Closes [#7543](https://github.com/pymc-devs/pymc/issues/7543) 

## Checklist

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -47,7 +47,10 @@ from pytensor.graph.traversal import ancestors
 from pytensor.ifelse import IfElse, ifelse
 from pytensor.scalar import Switch
 from pytensor.scalar import switch as scalar_switch
+from pytensor.scalar.basic import GE, GT, LE, LT, Mul
 from pytensor.tensor.basic import Join, MakeVector, switch
+from pytensor.tensor.elemwise import Elemwise
+from pytensor.tensor.exceptions import NotScalarConstantError
 from pytensor.tensor.random.rewriting import (
     local_dimshuffle_rv_lift,
     local_rv_size_lift,
@@ -80,7 +83,9 @@ from pymc.logprob.rewriting import (
     measurable_ir_rewrites_db,
     subtensor_ops,
 )
+from pymc.logprob.transforms import MeasurableTransform
 from pymc.logprob.utils import (
+    CheckParameterValue,
     check_potential_measurability,
     filter_measurable_variables,
     get_related_valued_nodes,
@@ -407,6 +412,80 @@ class MeasurableSwitchMixture(MeasurableElemwise):
 measurable_switch_mixture = MeasurableSwitchMixture(scalar_switch)
 
 
+class MeasurableLeakyReLUSwitch(MeasurableElemwise):
+    """A placeholder for leaky-ReLU graphs built via `switch(x > 0, x, a * x)`.
+
+    this is an invertible, piecewise-linear transform of a single continuous measurable variable.
+    """
+
+    valid_scalar_types = (Switch,)
+
+
+measurable_leaky_relu_switch = MeasurableLeakyReLUSwitch(scalar_switch)
+
+
+def _is_x_positive_condition(cond: TensorVariable, x: TensorVariable) -> bool:
+    if cond.owner is None:
+        return False
+    if not isinstance(cond.owner.op, Elemwise):
+        return False
+    scalar_op = cond.owner.op.scalar_op
+    if not isinstance(scalar_op, GT | GE | LT | LE):
+        return False
+
+    left, right = cond.owner.inputs
+
+    def _is_zero(v: TensorVariable) -> bool:
+        try:
+            return pt.get_underlying_scalar_constant_value(v) == 0
+        except NotScalarConstantError:
+            return False
+
+    # x > 0 or x >= 0
+    if left is x and _is_zero(right) and isinstance(scalar_op, GT | GE):
+        return True
+    # 0 < x or 0 <= x
+    if right is x and _is_zero(left) and isinstance(scalar_op, LT | LE):
+        return True
+    return False
+
+
+def _extract_leaky_relu_slope(
+    neg_branch: TensorVariable, x: TensorVariable
+) -> TensorVariable | None:
+    """Extract slope `a` from `neg_branch` assuming it represents `a * x`.
+
+    supports both plain `Elemwise(Mul)` and `MeasurableTransform` scale rewrites.
+    """
+    if neg_branch is x:
+        return pt.constant(1.0)
+
+    if neg_branch.owner is None:
+        return None
+
+    # handle case where `a * x` was already rewritten into a measurable scale transform
+    if isinstance(neg_branch.owner.op, MeasurableTransform):
+        op = neg_branch.owner.op
+        if not isinstance(op.scalar_op, Mul):
+            return None
+        # MeasurableTransform takes (measurable_input, scale)
+        if len(neg_branch.owner.inputs) != 2:
+            return None
+        if neg_branch.owner.inputs[op.measurable_input_idx] is not x:
+            return None
+        scale = neg_branch.owner.inputs[1 - op.measurable_input_idx]
+        return cast(TensorVariable, scale)
+
+    # plain multiplication
+    if isinstance(neg_branch.owner.op, Elemwise) and isinstance(neg_branch.owner.op.scalar_op, Mul):
+        left, right = neg_branch.owner.inputs
+        if left is x:
+            return cast(TensorVariable, right)
+        if right is x:
+            return cast(TensorVariable, left)
+    return None
+
+
 @node_rewriter([switch])
 def find_measurable_switch_mixture(fgraph, node):
     if isinstance(node.op, MeasurableOp):
@@ -431,6 +510,46 @@ def find_measurable_switch_mixture(fgraph, node):
     return [measurable_switch_mixture(switch_cond, *components)]
 
 
+@node_rewriter([switch])
+def find_measurable_leaky_relu_switch(fgraph, node):
+    """Detect `switch(x > 0, x, a * x)` and replace it by a measurable op.
+
+    This enables a change-of-variables logprob derivation instead of treating it as a mixture.
+    """
+    if isinstance(node.op, MeasurableOp):
+        return None
+
+    cond, pos_branch, neg_branch = node.inputs
+
+    if not filter_measurable_variables([pos_branch]):
+        return None
+    x = cast(TensorVariable, pos_branch)
+
+    if x.type.dtype.startswith("int"):
+        return None
+
+    if x.type.broadcastable != node.outputs[0].type.broadcastable:
+        return None
+
+    if not _is_x_positive_condition(cast(TensorVariable, cond), x):
+        return None
+
+    a = _extract_leaky_relu_slope(cast(TensorVariable, neg_branch), x)
+    if a is None:
+        return None
+
+    if check_potential_measurability([a]):
+        return None
+
+    return [
+        measurable_leaky_relu_switch(
+            cast(TensorVariable, cond),
+            x,
+            cast(TensorVariable, neg_branch),
+        )
+    ]
+
+
 @_logprob.register(MeasurableSwitchMixture)
 def logprob_switch_mixture(op, values, switch_cond, component_true, component_false, **kwargs):
     [value] = values
@@ -440,6 +559,32 @@ def logprob_switch_mixture(op, values, switch_cond, component_true, component_fa
         _logprob_helper(component_true, value),
         _logprob_helper(component_false, value),
     )
+
+
+@_logprob.register(MeasurableLeakyReLUSwitch)
+def logprob_leaky_relu_switch(op, values, cond, x, neg_branch, **kwargs):
+    (value,) = values
+
+    a = _extract_leaky_relu_slope(cast(TensorVariable, neg_branch), cast(TensorVariable, x))
+    if a is None:
+        raise NotImplementedError("Could not extract leaky-ReLU slope")
+
+    # enforce a > 0 at runtime to ensure invertibility and valid Jacobian
+    a = CheckParameterValue("leaky_relu slope > 0")(a, pt.all(pt.gt(a, 0)))
+
+    # inverse mapping x(y) = y if y>0 else y/a
+    x_inv = pt.switch(pt.gt(value, 0), value, value / a)
+
+    base_logp = _logprob_helper(x, x_inv, **kwargs)
+
+    # jacobian term: 0 for y>0, -log(a) for y<=0
+    jacobian = pt.switch(pt.gt(value, 0), pt.zeros_like(value), -pt.log(a))
+
+    if base_logp.ndim < value.ndim:
+        ndim_supp = value.ndim - base_logp.ndim
+        jacobian = jacobian.sum(axis=tuple(range(-ndim_supp, 0)))
+
+    return base_logp + jacobian
 
 
 measurable_ir_rewrites_db.register(
@@ -454,6 +599,13 @@ measurable_ir_rewrites_db.register(
     find_measurable_switch_mixture,
     "basic",
     "mixture",
+)
+
+measurable_ir_rewrites_db.register(
+    "find_measurable_leaky_relu_switch",
+    find_measurable_leaky_relu_switch,
+    "basic",
+    "transform",
 )
 
 
--- a/pymc/logprob/mixture.py
+++ b/pymc/logprob/mixture.py
@@ -47,10 +47,7 @@ from pytensor.graph.traversal import ancestors
 from pytensor.ifelse import IfElse, ifelse
 from pytensor.scalar import Switch
 from pytensor.scalar import switch as scalar_switch
-from pytensor.scalar.basic import GE, GT, LE, LT, Mul
 from pytensor.tensor.basic import Join, MakeVector, switch
-from pytensor.tensor.elemwise import Elemwise
-from pytensor.tensor.exceptions import NotScalarConstantError
 from pytensor.tensor.random.rewriting import (
     local_dimshuffle_rv_lift,
     local_rv_size_lift,
@@ -83,9 +80,7 @@ from pymc.logprob.rewriting import (
     measurable_ir_rewrites_db,
     subtensor_ops,
 )
-from pymc.logprob.transforms import MeasurableTransform
 from pymc.logprob.utils import (
-    CheckParameterValue,
     check_potential_measurability,
     filter_measurable_variables,
     get_related_valued_nodes,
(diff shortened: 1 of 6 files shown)


ricardoV94:
The more general requirement which we could try to see of we can cover is a switch where the two branches don't overlap, so when you invert you know for sure which branch it came from.

eclipse1605:
that makes sense. correct me if im wrong, but the key property we need is that the two branches map to non overlapping regions in `y`, so that when we observe a `y` we can tell which branch it came from and apply the correct inverse + Jacobian. 

if so, we can extend the current pattern matcher to `y = switch(x > k, f1(x), f2(x))`.  then we can extract `(m1, b1)` and `(m2, b2)` for the two branches. check that each branch is monotone (`m != 0`). check the images don’t overlap once you restrict to the domains `x ? k`. if the check passes, build the inverse per branch `x = (y - b) / m` and add the Jacobian term `-log|m|`, using a switch on `y` with the boundary induced by `k`.

eclipse1605:
@ricardoV94  if my understanding is right would you prefer that i extend this PR with that generalization, or should I open a follow-up PR to keep the changes scoped?

ricardoV94:
> that makes sense. correct me if im wrong, but the key property we need is that the two branches map to non overlapping regions in `y`, so that when we observe a `y` we can tell which branch it came from and apply the correct inverse + Jacobian.
> 
> if so, we can extend the current pattern matcher to `y = switch(x > k, f1(x), f2(x))`. then we can extract `(m1, b1)` and `(m2, b2)` for the two branches. check that each branch is monotone (`m != 0`). check the images don’t overlap once you restrict to the domains `x ? k`. if the check passes, build the inverse per branch `x = (y - b) / m` and add the Jacobian term `-log|m|`, using a switch on `y` with the boundary induced by `k`.

Yeah exactly! But you don't need to implement the logic to invert the branches, or the jacobian. If both branches of the switch are measurable, it means PyMC figured out the logp and you can just evaluate it at the final value (that will take care of those details). The question posed by the switch is which of them you need? 

Logp for this switch might look something like (pseudocode):

```python
def conditional_switch_logp(value, true_branch_rv, false_branch_rv):
  value_implies_true_branch = f(value)
  # Note the logp is evaluated at the value, the switch just gates which one is selected
  logp = switch(value_implies_true_branch, logp(true_branch1_rv, value), logp(true_branch2_rv, value))
  return logp
```

I think (need to confirm) that the way things are setup, the rewrite that marks the switch as being measurable only has to worry about whether we meet the constraints you mentioned.

Current code can already figure out the logp(Normal, value), or logp(Normal * a, value), for you.

The strategy may look something like this sequence of checks:
1. You have a switch
2. It's not yet Measurable
3. Both true and false branches are measurable (that means PyMC already known how to get their logp)
4. We have a simple condition expression, x > k or something along those lines. This is where you find what `x` even is.
5. Both branches are related to x
  5.1 If none is connected, this is already handled by the switch mixture machinery  switch(cond, x, y), do nothing
  5.2 If only one is connected, and the other is a constant, it's actually a censored process something like switch(x > 0, k, x). But this will never be the case given requirement 3. 
  5.3 if only one is connected, and the other is a measurable variable it's also fine, but the checks for invertibility may be trickier? Didn't want to think about this right now. If it seems simple to you let me know.
6. The setup passes the constraints for invertibility (from what we can infer)

Examples:
`y = switch(x > 0, x, x * a), a > 0 -> true_branch = y > 0`
`y = switch(x > 0, x * a, x * b), a,b > 0 -> true_branch = y > 0`
`y = switch(x > 0, x * a, x * b), a, b < 0 -> true_branch = y < 0`

But also:
`y = switch(x > 1, x ** 3, x) -> true_branch = y > 1`
`y = switch(x > 0, exp(x) - 1, x) -> true_branch = y > 0`
`y = switch(x > -1, x, exp(x+1) - 2) -> true_branch = y > -1`
`y = switch(x > 1, exp(x - 1), log(x) + 1) -> true_branch = y > 1`

(y is what becomes value in the logp function)

Restricting to the original monotonically increasing leaky RELU case is fine, but I would like to structure the code so it's ready to extend to more cases in the future.

If once you figure it out, you want to extend that's awesome and welcome but not a blocker.

How does that sound?

eclipse1605:
@ricardoV94 thanks, that makes sense, i refactored the implementation to follow that approach.

- rewrote `find_measurable_leaky_relu_switch` so it now only tags the switch as measurable when both branches are already measurable, so we can delegate all inversion/Jacobian details to existing logprob rules for each branch.
- the `_logprob` for `MeasurableLeakyReLUSwitch` now just gates between branch logps evaluated at the observed `value`:
  `switch(value > 0, _logprob_helper(x, value), _logprob_helper(neg_branch, value))`
- kept the runtime `CheckParameterValue("leaky_relu slope > 0")` guard to ensure the “value implies branch” predicate is valid, and attached it to the returned expression so it can’t get optimized away.

this is currently scoped to the leaky ReLU pattern, but the structure is such that extending to other non overlapping switch patterns should be straightforward (separate predicate + constraints logic). if approved ill go ahead trying to implement the general “non-overlapping images” framework.

eclipse1605:
@ricardoV94 should i make any more changes in this PR?

pymc/logprob/mixture.py
```diff
@@ -431,6 +510,51 @@ def find_measurable_switch_mixture(fgraph, node):
...
+        return None
+
+    cond, pos_branch, neg_branch = node.inputs
+
+    # we only mark the switch measurable once both branches are already measurable.
+    # so, the switch logprob can simply gate between branch logps (delegating inversion/Jacobian details to each branch).
+    if set(filter_measurable_variables([pos_branch, neg_branch])) != {pos_branch, neg_branch}:
+        return None
+
+    if not filter_measurable_variables([pos_branch]):
```
ricardoV94: Isn't this check redundant with the previous one?

pymc/logprob/mixture.py
```diff
@@ -431,6 +510,51 @@ def find_measurable_switch_mixture(fgraph, node):
...
+
+    if not _is_x_positive_condition(cast(TensorVariable, cond), x):
+        return None
+
+    a = _extract_leaky_relu_slope(cast(TensorVariable, neg_branch), x)
+    if a is None:
+        return None
+
+    if check_potential_measurability([a]):
+        return None
```
ricardoV94: why this check?

pymc/logprob/mixture.py
```diff
@@ -431,6 +510,51 @@ def find_measurable_switch_mixture(fgraph, node):
     return [measurable_switch_mixture(switch_cond, *components)]
 
 
+@node_rewriter([switch])
+def find_measurable_leaky_relu_switch(fgraph, node):
+    """Detect `switch(x > 0, x, a * x)` and replace it by a measurable op.
+
+    This enables a change-of-variables logprob derivation instead of treating it as a mixture.
```
ricardoV94: This second sentence is confusing, I would just remove it

pymc/logprob/mixture.py
```diff
@@ -407,6 +412,80 @@ class MeasurableSwitchMixture(MeasurableElemwise):
 measurable_switch_mixture = MeasurableSwitchMixture(scalar_switch)
 
 
+class MeasurableLeakyReLUSwitch(MeasurableElemwise):
```
ricardoV94: Let's give it a more abstract name like MeasurableSwitchStepChange (better suggestions welcome)?, to pave the road to more cases than LeakyRelu? 

Also, this shouldn't be in `mixture.py` because it's not a mixture

pymc/logprob/mixture.py
```diff
@@ -407,6 +412,80 @@ class MeasurableSwitchMixture(MeasurableElemwise):
...
+    supports both plain `Elemwise(Mul)` and `MeasurableTransform` scale rewrites.
+    """
+    if neg_branch is x:
+        return pt.constant(1.0)
+
+    if neg_branch.owner is None:
+        return None
+
+    # handle case where `a * x` was already rewritten into a measurable scale transform
+    if isinstance(neg_branch.owner.op, MeasurableTransform):
```
ricardoV94: I think this is the only case you need to handle, because you have a check in the rewrite that both branches are Measurable?

ricardoV94 (commented):
Sorry for the delay.

This is looking great, but I do have some comments / questions.

Let me know if my comments are off.

pymc/logprob/mixture.py
eclipse1605: oh thats right, my bad

pymc/logprob/mixture.py
eclipse1605: to reject the cases where the scale could be an RV because this handler assumes a deterministic step change parameter

pymc/logprob/mixture.py
eclipse1605: > This second sentence is confusing, I would just remove it

sure

pymc/logprob/mixture.py
eclipse1605: what about something like `MeasurableSwitchNonOverlapping`? also should I just move this measurable switch transform logic into a new module like `switch.py` or something? this isn’t a mixture and having it in a separate module would make it easier to extend later as you mentioned

pymc/logprob/mixture.py
eclipse1605: what if rewrite ordering (or a future refactor) means the `a * x`  branch is still a plain `Elemwise(Mul)` at the point we match the switch, even though the branch is “measurable” overall

pymc/logprob/mixture.py
ricardoV94: Both suggestions sound great!

pymc/logprob/mixture.py
ricardoV94: A plain Elemwise(Mul) is not measurable (even if its inputs are already measurable). It becomes measurable when a rewrite converts it into one, at which points it ceases to be a plain Elemwise(Mul)

eclipse1605: makes sense, ill drop the plain `Elemwise(Mul)` as only to match the measurable scale-transform case.

pymc/logprob/mixture.py
eclipse1605: ill make those changes

eclipse1605:
how about this @ricardoV94?

pymc/logprob/switch.py
```diff
@@ -0,0 +1,202 @@
...
+    cond, pos_branch, neg_branch = node.inputs
+
+    # Only mark the switch measurable once both branches are already measurable.
+    # Then the logprob can simply gate between branch logps evaluated at `value`.
+    if set(filter_measurable_variables([pos_branch, neg_branch])) != {pos_branch, neg_branch}:
+        return None
+
+    x = cast(TensorVariable, pos_branch)
+
+    if x.type.dtype.startswith("int"):
```
ricardoV94: To be sure, work only with floats (not unsigned integers, booleans or complex)

```suggestion
    if x.type.numpy_dtype.kind != "f":
```

pymc/logprob/switch.py
```diff
@@ -0,0 +1,202 @@
...
+    Currently matches leaky-ReLU graphs of the form `switch(x > 0, x, a * x)`.
+    """
+
+    valid_scalar_types = (Switch,)
+
+
+measurable_switch_non_overlapping = MeasurableSwitchNonOverlapping(scalar_switch)
+
+
+def _is_x_threshold_condition(cond: TensorVariable, x: TensorVariable) -> bool:
```
ricardoV94: better name for this helper to emphasize zero? perhaps `_is_zero_x_threshold`

pymc/logprob/switch.py
```diff
@@ -0,0 +1,202 @@
...
+    if neg_branch.owner.inputs[op.measurable_input_idx] is not x:
+        return None
+
+    scale = neg_branch.owner.inputs[1 - op.measurable_input_idx]
+    return cast(TensorVariable, scale)
+
+
+@node_rewriter([tensor_switch])
+def find_measurable_switch_non_overlapping(fgraph, node):
+    """Detect `switch(x > 0, x, a * x)` and replace it by a measurable op."""
```
ricardoV94, line 142: Can you open an issue for follow-up extensions? From the top of my mind:
1. We should try and support the equivalent switch written the other way around `switch(x <= 0, a * x, x)`. The user shouldn't have to guess the specific format that we support

2. Allow scaling factors on both branches, still with the constraint that they must have the same sign (or if we want to remain more restrictive for now, that they are both non-negative). Just because it isn't too much harder than what we have now.

3. The more general cases discussed above with monotonic functions of `x` that don't overlap an the two branches

pymc/logprob/switch.py
```diff
@@ -0,0 +1,202 @@
...
+        cast(TensorVariable, neg_branch), cast(TensorVariable, x)
+    )
+    if a is None:
+        raise NotImplementedError("Could not extract non-overlapping scale")
+
+    a_is_positive = pt.all(pt.gt(a, 0))
+
+    # For `a > 0`, `switch(x > 0, x, a * x)` maps to disjoint regions in `value`:
+    # true branch -> value > 0, false branch -> value <= 0.
+    value_implies_true_branch = pt.gt(value, 0)
```
ricardoV94: Mathematically it should be the same with a continuous function, but the logp may be more stable? If we want to be pedantic we could check if the original cond had the exact zero in the true or neg branch.

```suggestion
    value_implies_true_branch = pt.ge(value, 0)
```

ricardoV94 (commented):
This is pretty great. I only have minor requests/questions left

tests/logprob/test_transforms.py
```diff
@@ -229,6 +232,57 @@ def test_exp_transform_rv():
...
 
+def test_leaky_relu_switch_logp_scalar():
+    a = 0.5
+    x = pm.Normal.dist(mu=0, sigma=1)
+    y = pm.math.switch(x > 0, x, a * x)
+
+    v_pos = 1.2
+    np.testing.assert_allclose(
+        pm.logp(y, v_pos, warn_rvs=False).eval(),
+        pm.logp(x, v_pos, warn_rvs=False).eval(),
```
ricardoV94:
```suggestion
        pm.logp(y, v_pos).eval(),
        pm.logp(x, v_pos).eval(),
```

tests/logprob/test_transforms.py
```diff
@@ -229,6 +232,57 @@ def test_exp_transform_rv():
...
+    )
+
+    v_neg = -2.0
+    np.testing.assert_allclose(
+        pm.logp(y, v_neg, warn_rvs=False).eval(),
```
ricardoV94: If you're testing with two values, define the logp variable once (or compile a function with the logp as output once), and reuse it. That will avoid duplicated logp inference calls.

tests/logprob/test_transforms.py
```diff
@@ -229,6 +232,57 @@ def test_exp_transform_rv():
...
+        pm.logp(x, v_neg / a, warn_rvs=False).eval() - np.log(a),
+    )
+
+    # boundary point (measure-zero for continuous RVs): should still produce a finite logp
+    assert np.isfinite(pm.logp(y, 0.0, warn_rvs=False).eval())
+
+
+def test_leaky_relu_switch_logp_vectorized():
```
ricardoV94: Use vectorized in the first test, and remove this. It's not conceputally that different

tests/logprob/test_transforms.py
```diff
@@ -229,6 +232,57 @@ def test_exp_transform_rv():
...
+    y = pm.math.switch(x > 0, x, a * x)
+
+    v = np.array([-2.0, 0.0, 1.5])
+    expected = pm.logp(x, np.where(v > 0, v, v / a), warn_rvs=False).eval() + np.where(
+        v > 0, 0.0, -np.log(a)
+    )
+    np.testing.assert_allclose(pm.logp(y, v, warn_rvs=False).eval(), expected)
+
+
+def test_leaky_relu_switch_logp_symbolic_slope_checks_positive():
```
ricardoV94: use symbolic a in the first test and test the error there. Still a pretty straightforward test

tests/logprob/test_transforms.py
```diff
@@ -229,6 +232,57 @@ def test_exp_transform_rv():
     )
 
 
+def test_leaky_relu_switch_logp_scalar():
```
ricardoV94: tests should be moved to a `test_switch.py` file

ricardoV94: Also add a separate test that shows the failure if x is broadcast by cond or a, or if it's discrete.

ricardoV94 (changes requested):

tests/logprob/test_transforms.py
eclipse1605: > tests should be moved to a `test_switch.py` file

oh right, that's my bad

eclipse1605: > Also add a separate test that shows the failure if x is broadcast by cond or a, or if it's discrete.

sure

pymc/logprob/switch.py
eclipse1605: sure, that makes sense

eclipse1605:
@ricardoV94 now?

pymc/logprob/switch.py
eclipse1605: > 1. We should try and support the equivalent switch written the other way around `switch(x <= 0, a * x, x)`. The user shouldn't have to guess the specific format that we support

I have raised #8049 for this, please check it and let me know if I should raise the rest of the issues as well, or if you’d prefer a different structure/scope.

I was thinking to raise issues for the following extensions:

- support equivalent `switch` spellings (e.g. `switch(x <= 0, a*x, x)` / swapped comparisons) so users don’t need to match a single canonical form.
- support scaling on both branches (`switch(x > 0, a_pos*x, a_neg*x)`) with constraints that guarantee non-overlapping images (start with both scales strictly positive or same-sign + non-zero).
- track a broader "non-overlapping switch transform" framework for monotone branch functions of `x` where the observed value implies the branch, while still gating between existing branch logps.
- support non-zero thresholds `(x ? k)` once the above is stable, since determining the `value -> branch` predicate becomes more subtle.

ricardoV94: Yeah that's a solid plan. Feel free to describe all that in the issue or open separate ones

ricardoV94: Yeah that's a solid plan

pymc/logprob/switch.py
```diff
@@ -0,0 +1,225 @@
...
+def logprob_switch_non_overlapping(op, values, cond, x, neg_branch, **kwargs):
+    (value,) = values
+
+    a = _extract_scale_from_measurable_mul(
+        cast(TensorVariable, neg_branch), cast(TensorVariable, x)
+    )
+    if a is None:
+        raise NotImplementedError("Could not extract non-overlapping scale")
+
+    a_is_positive = pt.all(pt.gt(a, 0))
```
ricardoV94, line 200: Does it need to be positive or non-negative is enough?

tests/logprob/test_switch.py
```diff
@@ -0,0 +1,103 @@
...
+import pymc as pm
+
+from pymc.logprob.basic import logp
+from pymc.logprob.utils import ParameterValueError
+
+
+def test_switch_non_overlapping_logp_matches_change_of_variables():
+    scale = pt.scalar("scale")
+    x = pm.Normal.dist(mu=0, sigma=1, size=(3,))
+    y = cast(TensorVariable, pt.switch(x > 0, x, scale * x))
```
ricardoV94: No need for casting in the tests, we don't run mypy here and it clutters the code

tests/logprob/test_switch.py
```diff
@@ -0,0 +1,103 @@
...
+    np.testing.assert_allclose(logp(y_s, v_pos).eval({scale: 0.5}), logp(x_s, v_pos).eval())
+
+    v_neg = -2.0
+    np.testing.assert_allclose(
+        logp(y_s, v_neg).eval({scale: 0.5}),
+        logp(x_s, v_neg / 0.5).eval() - np.log(0.5),
+    )
+
+    # boundary point (measure-zero for continuous RVs): should still produce a finite logp
+    assert np.isfinite(logp(y_s, 0.0, warn_rvs=False).eval({scale: 0.5}))
```
ricardoV94: Why are these needed? Can't you test everything with the first functions?

tests/logprob/test_switch.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_switch_non_overlapping_does_not_rewrite_if_x_replicated_by_condition():
+    scale = pt.scalar("scale")
+    x = pm.Normal.dist(mu=0, sigma=1, size=(3,))
+    cond = (x[None, :] > 0) & pt.ones((2, 1), dtype="bool")
+    y = cast(TensorVariable, pt.switch(cond, x, scale * x))
+
+    with pytest.raises(NotImplementedError, match="Logprob method not implemented for Switch"):
+        logp(y, np.zeros((2, 3)), warn_rvs=False).eval({scale: 0.5})
```
ricardoV94: No point in eval as it should already fail in the logp call

tests/logprob/test_switch.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_switch_non_overlapping_does_not_rewrite_if_scale_broadcasts_x():
+    x = pm.Normal.dist(mu=0, sigma=1)
+    scale = pt.vector("scale")
+    y = cast(TensorVariable, pt.switch(x > 0, x, scale * x))
+
+    with pytest.raises(NotImplementedError, match="Logprob method not implemented for Switch"):
+        logp(y, np.zeros((3,)), warn_rvs=False).eval({scale: np.array([0.5, 0.5, 0.5])})
```
ricardoV94: same

ricardoV94: Also I would merge with previous test, conceptually similar

ricardoV94: Also I would merge with previous test, conceptually similar

tests/logprob/test_switch.py
```diff
@@ -0,0 +1,103 @@
...
+    # boundary point (measure-zero for continuous RVs): should still produce a finite logp
+    assert np.isfinite(logp(y_s, 0.0, warn_rvs=False).eval({scale: 0.5}))
+
+
+def test_switch_non_overlapping_requires_positive_scale():
+    scale = pt.scalar("scale")
+    x = pm.Normal.dist(mu=0, sigma=1)
+    y = cast(TensorVariable, pt.switch(x > 0, x, scale * x))
+
+    with pytest.raises(ParameterValueError, match="switch non-overlapping scale > 0"):
```
ricardoV94, line 43: you can use the function from the first test to check the negative scale raising, no need for separate test

ricardoV94 (changes requested):
Couple more nits, just about tests now.

I don't understand why you keep passing `warn_rvs=False`, there shouldn't be any and the default is more strict

pymc/logprob/switch.py
eclipse1605: it needs to be positive because if `a == 0`, the negative branch becomes `0 * x = 0`, so a whole half-line of `x` collapses to a single value which is not invertible. ill add a comment about this in the file

tests/logprob/test_switch.py
eclipse1605: oh sure

tests/logprob/test_switch.py
eclipse1605: yeah they aren’t needed.

pymc/logprob/switch.py
ricardoV94: oh yeah that makes sense

ricardoV94: OTOH If it goes on the `0 * x` branch logp, that will already return `nan` (or inf if the value is zero):

```python
pm.logp(pm.Normal.dist() * pt.scalar("a"), 1).eval({"a": 0})  # np.nan 
```

Probably better to leave as is though.

ricardoV94 (approved):
Awesome

PR #8003, opened 2025-12-16
Fix numpydoc spacing in `get_tau_sigma` docstring

Jiya873:
related to #5459


OriolAbril (approved):
LGTM, not sure if it is worth running the tests for this though. Is it possible to merge without them running or do we have non-bypassable rules against it?

PR #8014, opened 2025-12-20
Removed tune stat

Jiya873:
related to #7997

PR #7776, opened 2025-05-05
Fix progressbar with nested compound step samplers

ricardoV94:
Alternative to #7730 that fixes #7721 and fixes #7724

PR #7730, opened 2025-03-20
Fix progress bar error when nested `CompoundStep` samplers are assigned

jessegrabowski:
#7721 reports an error in the presence of nested `CompoundStep`. Here's a prettier version of what pymc gives for the example in that issue:

```
CompoundStep
├─CompoundStep
│   ├─ Metropolis: [a]
│   ├─ Metropolis: [b]
│   └─Metropolis: [c]
└─NUTS: [d]
```

So there are 4 steps, but there's a compound step on the outside and on the inside. At each step, we get a flat list of 4 dictionaries holding statistics for each step. Currently, the logic for updating the progress bars makes the assumption that the list of step statistics returned at each step matches the list of step samplers. It was assumed that, if there is a compound step, there should only be one, so it can zip over the steps. Here is the display stat update for `CompoundStep`:

```
            for step_stat, update_fn in zip(step_stats, update_fns):
                displayed_stats = update_fn(displayed_stats, step_stat, chain_idx)
```

The problem is that if there's a nested structure, one of the `udpate_fns` will do this loop again. Since `step_stats` does *not* have the same nested structure, it ends up iterating over dictionary keys and raising the error.

Open to suggestions on how to proceed, because the solution isn't obvious. Opened this as a draft PR to address the problem ASAP, since its been reported 3 times now.

## Related Issue

- [ ] Closes #7721 #7724 #
- [ ] Related to https://github.com/pymc-devs/pymc-extras/issues/331

## Checklist

- [ ] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Type of change

- [ ] New feature / enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7730.org.readthedocs.build/en/7730/

Issue #7724, opened 2025-03-16
BUG: Indexing TypeError when sampling with missing observations


fonnesbeck:
### Describe the issue:

Sampling with observations containing missing values is failing in 5.20. `Metropolis` raises a `TypeError` when updating the sampling metadata. This does not occur when `Metropolis` is used for the entire model.

It appears that `stats['tune']` ends up being a string in this scenario instead of an array.

### Reproduceable code example:

```python
import pymc as pm
import numpy as np

disasters_missing = np.array([ 4, 5, 4, 0, 1, 4, 3, 4, 0, 6, 3, 3, 4, 0, 2, 6,
3, 3, 5, 4, 5, 3, 1, 4, 4, 1, 5, 5, 3, 4, 2, 5,
2, 2, 3, 4, 2, 1, 3, np.nan, 2, 1, 1, 1, 1, 3, 0, 0,
1, 0, 1, 1, 0, 0, 3, 1, 0, 3, 2, 2, 0, 1, 1, 1,
0, 1, 0, 1, 0, 0, 0, 2, 1, 0, 0, 0, 1, 1, 0, 2,
3, 3, 1, np.nan, 2, 1, 1, 1, 1, 2, 4, 2, 0, 0, 1, 4,
0, 0, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 1, 0, 1])

N = len(disasters_missing)

with pm.Model() as missing_data_model:

    # Prior for distribution of switchpoint location
    switchpoint = pm.DiscreteUniform('switchpoint', lower=0, upper=N)
    # Priors for pre- and post-switch mean number of disasters
    early_mean = pm.Exponential('early_mean', lam=1.)
    late_mean = pm.Exponential('late_mean', lam=1.)

    # Allocate appropriate Poisson rates to years before and after current
    # switchpoint location
    idx = np.arange(N)
    rate = pm.math.switch(switchpoint >= idx, early_mean, late_mean)

    # Data likelihood
    disasters = pm.Poisson('disasters', rate, observed=disasters_missing)

    trace_missing = pm.sample()
```

### Error message:

```shell
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[60], line 2
      1 with missing_data_model:
----> 2     trace_missing = pm.sample()

[...]

PR #8015, opened 2025-12-20
Remove `tune` stat from steps and fix non-discarding of tuning draws from trace

### LINKED ISSUE #7997: BUG: CategoricalGibbsMetropolis doesn't respect the tune parameter
### Describe the issue:

I think that [this PR](https://github.com/pymc-devs/pymc/pull/7776) changed the behaviour of the `CategoricalGibbsMetropolis` sampler which now treats all `tune` and `draws` as `draws`. By comparison, the behaviour of `BinaryGibbsMetropolis` was unchanged, so I am guessing that this change was unintended.

If intended I find this behaviour confusing and we should probably emit a warning message to clarify?

```python
import pymc as pm
with pm.Model():
    pm.Categorical("cat", [0.02, 0.08, 0.9])
    idata = pm.sample(draws=20, tune=20)
# Sampling 4 chains for 0 tune and 40 draw iterations (0 + 160 draws total)
# Returns an idata object with 40 draws
```

### Error message:

```shell

```

### PyMC version information:

PyMC Version: 5.24 to current

### Context for the issue:

_No response_

eclipse1605:
## Description

I tried to make “tuning vs draws” a driver owned concept again. Right now, parts of sampling/postprocessing infer warmup length from a per-step `"tune"` sampler stat, which can get out of sync (e.g. a step method returning `"tune": False` everywhere makes PyMC think `n_tune == 0`, so warmup isn’t discarded and the logs look wrong).

## Related Issues

Fixes: [#7997](https://github.com/pymc-devs/pymc/issues/7997) 
Context: [#7776](https://github.com/pymc-devs/pymc/pull/7776) (progressbar/stat refactor that exposed the mismatch)
Related discussion/attempts: [#7730](https://github.com/pymc-devs/pymc/pull/7730), [#7721](https://github.com/pymc-devs/pymc/issues/7721), [#7724](https://github.com/pymc-devs/pymc/issues/7724), [#8014](https://github.com/pymc-devs/pymc/pull/8014)

### DIFF
--- a/pymc/backends/base.py
+++ b/pymc/backends/base.py
@@ -17,12 +17,14 @@
 See the docstring for pymc.backends for more information
 """
 
+import inspect
 import itertools as itl
 import logging
 import warnings
 
 from abc import ABC
 from collections.abc import Mapping, Sequence, Sized
+from functools import cache
 from typing import (
     Any,
     TypeVar,
@@ -38,6 +40,58 @@ from pymc.pytensorf import compile
 from pymc.util import get_var_name
 
 logger = logging.getLogger(__name__)
+
+
+@cache
+def _record_supports_in_warmup(trace_type: type) -> str:
+    """Return how to call `trace.record` for this backend type.
+
+    Returns
+    -------
+    mode : {"kw", "no", "try"}
+        - "kw": safe to pass `in_warmup=` (parameter present or **kwargs supported)
+        - "no": do not pass `in_warmup=`
+        - "try": signature introspection failed; try `in_warmup=` and fallback on
+          unexpected-keyword errors.
+    """
+    try:
+        sig = inspect.signature(trace_type.record)
+    except (TypeError, ValueError):
+        return "try"
+
+    parameters = sig.parameters
+    if "in_warmup" in parameters:
+        return "kw"
+    if any(p.kind == inspect.Parameter.VAR_KEYWORD for p in parameters.values()):
+        return "kw"
+    return "no"
+
+
+def _record_with_in_warmup(
+    trace: "IBaseTrace",
+    draw: Mapping[str, np.ndarray],
+    stats: Sequence[Mapping[str, Any]],
+    *,
+    in_warmup: bool,
+):
+    """Record a draw, passing `in_warmup` when the backend supports it.
+
+    this will keep compatibility with custom traces that predate the `in_warmup` kwarg.
+    """
+    mode = _record_supports_in_warmup(type(trace))
+    if mode == "kw":
+        return trace.record(draw, stats, in_warmup=in_warmup)
+    if mode == "no":
+        return trace.record(draw, stats)
+
+    # fallback for backends we can't introspect reliably.
+    try:
+        return trace.record(draw, stats, in_warmup=in_warmup)
+    except TypeError as err:
+        message = str(err)
+        if "unexpected keyword argument" in message and "in_warmup" in message:
+            return trace.record(draw, stats)
+        raise
 
 
 class BackendError(Exception):
@@ -118,7 +172,7 @@ class IBaseTrace(ABC, Sized):
         draw: Mapping[str, np.ndarray],
         stats: Sequence[Mapping[str, Any]],
         *,
-        tune: bool | None = None,
+        in_warmup: bool,
     ):
         """Record results of a sampling iteration.
 
@@ -128,8 +182,8 @@ class IBaseTrace(ABC, Sized):
             Values mapped to variable names
         stats: list of dicts
             The diagnostic values for each sampler
-        tune: bool | None
-            Whether this draw belongs to the tuning/warmup phase. This is a driver-owned
+        in_warmup: bool
+            Whether this draw belongs to the warmup phase. This is a driver-owned
             concept and is intended for storage/backends to persist warmup information.
         """
         raise NotImplementedError()
--- a/pymc/backends/base.py
+++ b/pymc/backends/base.py
@@ -17,14 +17,12 @@
 See the docstring for pymc.backends for more information
 """
 
-import inspect
 import itertools as itl
 import logging
 import warnings
 
 from abc import ABC
 from collections.abc import Mapping, Sequence, Sized
-from functools import cache
 from typing import (
     Any,
     TypeVar,
@@ -40,58 +38,6 @@ from pymc.pytensorf import compile
 from pymc.util import get_var_name
 
 logger = logging.getLogger(__name__)
-
-
-@cache
-def _record_supports_in_warmup(trace_type: type) -> str:
-    """Return how to call `trace.record` for this backend type.
-
-    Returns
-    -------
-    mode : {"kw", "no", "try"}
-        - "kw": safe to pass `in_warmup=` (parameter present or **kwargs supported)
-        - "no": do not pass `in_warmup=`
-        - "try": signature introspection failed; try `in_warmup=` and fallback on
-          unexpected-keyword errors.
-    """
-    try:
-        sig = inspect.signature(trace_type.record)
-    except (TypeError, ValueError):
-        return "try"
-
-    parameters = sig.parameters
-    if "in_warmup" in parameters:
-        return "kw"
-    if any(p.kind == inspect.Parameter.VAR_KEYWORD for p in parameters.values()):
-        return "kw"
-    return "no"
-
-
-def _record_with_in_warmup(
-    trace: "IBaseTrace",
-    draw: Mapping[str, np.ndarray],
-    stats: Sequence[Mapping[str, Any]],
-    *,
-    in_warmup: bool,
-):
-    """Record a draw, passing `in_warmup` when the backend supports it.
-
-    this will keep compatibility with custom traces that predate the `in_warmup` kwarg.
-    """
-    mode = _record_supports_in_warmup(type(trace))
-    if mode == "kw":
-        return trace.record(draw, stats, in_warmup=in_warmup)
-    if mode == "no":
-        return trace.record(draw, stats)
-
-    # fallback for backends we can't introspect reliably.
-    try:
-        return trace.record(draw, stats, in_warmup=in_warmup)
-    except TypeError as err:
-        message = str(err)
-        if "unexpected keyword argument" in message and "in_warmup" in message:
-            return trace.record(draw, stats)
-        raise
 
 
 class BackendError(Exception):
--- a/pymc/backends/base.py
+++ b/pymc/backends/base.py
@@ -113,7 +113,13 @@ class IBaseTrace(ABC, Sized):
         """
         raise NotImplementedError()
 
-    def record(self, draw: Mapping[str, np.ndarray], stats: Sequence[Mapping[str, Any]]):
+    def record(
+        self,
+        draw: Mapping[str, np.ndarray],
+        stats: Sequence[Mapping[str, Any]],
+        *,
+        tune: bool | None = None,
+    ):
         """Record results of a sampling iteration.
 
         Parameters
@@ -122,6 +128,9 @@ class IBaseTrace(ABC, Sized):
             Values mapped to variable names
         stats: list of dicts
             The diagnostic values for each sampler
+        tune: bool | None
+            Whether this draw belongs to the tuning/warmup phase. This is a driver-owned
+            concept and is intended for storage/backends to persist warmup information.
         """
         raise NotImplementedError()
 
--- a/pymc/backends/mcbackend.py
+++ b/pymc/backends/mcbackend.py
@@ -105,9 +105,9 @@ class ChainRecordAdapter(IBaseTrace):
             {sname: stats_dtypes[fname] for fname, sname, is_obj in sstats}
             for sstats in stats_bijection._stat_groups
         ]
-        if "tune" in stats_dtypes and self.sampler_vars:
-            # expose driver-owned warmup marker via the sampler-stats API.
-            self.sampler_vars[0].setdefault("tune", stats_dtypes["tune"])
+        if "in_warmup" in stats_dtypes and self.sampler_vars:
+            # Expose driver-owned warmup marker via the sampler-stats API.
+            self.sampler_vars[0].setdefault("in_warmup", stats_dtypes["in_warmup"])
 
         self._chain = chain
         self._point_fn = point_fn
@@ -119,12 +119,12 @@ class ChainRecordAdapter(IBaseTrace):
         draw: Mapping[str, np.ndarray],
         stats: Sequence[Mapping[str, Any]],
         *,
-        tune: bool | None = None,
+        in_warmup: bool,
     ):
         values = self._point_fn(draw)
         value_dict = dict(zip(self.varnames, values))
         stats_dict = self._statsbj.map(stats)
-        stats_dict["tune"] = bool(tune)
+        stats_dict["in_warmup"] = bool(in_warmup)
         # Apply pickling to objects stats
         for fname in self._statsbj.object_stats.keys():
             val_bytes = pickle.dumps(stats_dict[fname])
@@ -157,8 +157,9 @@ class ChainRecordAdapter(IBaseTrace):
         self, stat_name: str, sampler_idx: int | None = None, burn=0, thin=1
     ) -> np.ndarray:
         slc = slice(burn, None, thin)
-        if stat_name == "tune":
-            return self._get_stats("tune", slc)
+        if stat_name in {"in_warmup", "tune"}:
+            # Backwards-friendly alias for users that might try "tune".
+            return self._get_stats("in_warmup", slc)
         # When there's just one sampler, default to remove the sampler dimension
         if sampler_idx is None and self._statsbj.n_samplers == 1:
             sampler_idx = 0
@@ -247,7 +248,7 @@ def make_runmeta_and_point_fn(
     # driver owned warmup marker. stored once per draw.
     sample_stats.append(
         mcb.Variable(
-            name="tune",
+            name="in_warmup",
             dtype=np.dtype(bool).name,
             shape=[],
             undefined_ndim=False,
--- a/pymc/backends/mcbackend.py
+++ b/pymc/backends/mcbackend.py
@@ -34,7 +34,6 @@ from pymc.step_methods.compound import (
     BlockedStep,
     CompoundStep,
     StatsBijection,
-    check_step_emits_tune,
     flat_statname,
     flatten_steps,
 )
@@ -209,8 +208,6 @@ def make_runmeta_and_point_fn(
     model: Model,
 ) -> tuple[mcb.RunMeta, PointFunc]:
     variables, point_fn = get_variables_and_point_fn(model, initial_point)
-
-    check_step_emits_tune(step)
 
     # In PyMC the sampler stats are grouped by the sampler.
     sample_stats = []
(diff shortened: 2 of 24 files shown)


ricardoV94:
@OriolAbril / @aloctavodia does any part of Arviz require the step samples to have a tune flag? Is it enough that we have warmup / posterior distinction, each with their number of draws?

pymc/step_methods/compound.py
```diff
@@ -352,12 +356,6 @@ def flatten_steps(step: BlockedStep | CompoundStep) -> list[BlockedStep]:
 
 
 def check_step_emits_tune(step: CompoundStep | BlockedStep):
```
ricardoV94: If this ends up working without tune, remove this function as well

pymc/backends/mcbackend.py
```diff
@@ -210,8 +209,6 @@ def make_runmeta_and_point_fn(
 ) -> tuple[mcb.RunMeta, PointFunc]:
     variables, point_fn = get_variables_and_point_fn(model, initial_point)
 
-    check_step_emits_tune(step)
```
ricardoV94, line 213: @michaelosthege what's the requirement here? Can we change things on mcbackend to not expect tune info attached to the steps?

pymc/step_methods/metropolis.py
```diff
@@ -331,15 +329,13 @@ def competence(var, has_grad):
     @staticmethod
     def _progressbar_config(n_chains=1):
         columns = [
-            TextColumn("{task.fields[tune]}", table_column=Column("Tuning", ratio=1)),
```
ricardoV94, line 334: We'll need a work-around for this, as we still want to show whether we're in tuning or not in the progressbar

ricardoV94:
Taking a step back, would it make sense for a `tune=None` mode where the sampler(s) decide how much tune they need? In that case it would make sense for the individual steps to report back whether they're tuning or not. 

Even if that's the case, I think it still makes sense to remove this currently useless stat and reintroduce in a separate PR (provided nobody finds a reason why it is actually useful/needed).

CC @aloctavodia, @lucianopaz @aseyboldt

pymc/backends/mcbackend.py
michaelosthege: No we can not change that tune info must be attached to a step. See my comment in https://github.com/pymc-devs/pymc/issues/7997#issuecomment-3679093688.

michaelosthege:
> Taking a step back, would it make sense for a `tune=None` mode where the sampler(s) decide how much tune they need? In that case it would make sense for the individual steps to report back whether they're tuning or not.

Automatically stopping the warmup early would be nice. I think we should agree on cleanly separated definitions of warmup, burn-in and tuning. Samplers not needing to tune parameters doesn't mean that there's no need for a warmup phase of burn-in iterations (however one might call it).

Our current implementation is bad because it doesn't separate the concepts.

aloctavodia:
ArviZ does not require or use a "tune" stats anywhere.

eclipse1605:
[#7997 (comment)](https://github.com/pymc-devs/pymc/issues/7997#issuecomment-3698649409)

pymc/backends/base.py
```diff
@@ -113,7 +113,13 @@ def point(self, idx: int) -> dict[str, np.ndarray]:
         """
         raise NotImplementedError()
 
-    def record(self, draw: Mapping[str, np.ndarray], stats: Sequence[Mapping[str, Any]]):
+    def record(
+        self,
+        draw: Mapping[str, np.ndarray],
+        stats: Sequence[Mapping[str, Any]],
+        *,
+        tune: bool | None = None,
```
michaelosthege: I would suggest to call it `in_warmup` to align with the commonly understood PyMC/ArviZ terminology of warmup vs. posterior phase.

Additionally I think we should not set a default value.

The sampling iterators in `sampling/mcmc.py` (what you call "driver") should always pass the new parameter and that breaks any custom trace that does not expect it.

pymc/backends/mcbackend.py
```diff
@@ -106,16 +105,26 @@ def __init__(
...
+        self,
+        draw: Mapping[str, np.ndarray],
+        stats: Sequence[Mapping[str, Any]],
+        *,
+        tune: bool | None = None,
+    ):
         values = self._point_fn(draw)
         value_dict = dict(zip(self.varnames, values))
         stats_dict = self._statsbj.map(stats)
+        stats_dict["tune"] = bool(tune)
```
michaelosthege: If we call it `in_warmup` I can change [this line in McBackend](https://github.com/pymc-devs/mcbackend/blob/366ab2d316c53170fc0d5663b2b20c8d718a4eb6/mcbackend/core.py#L154) to look for it.

This has the advantage of not colliding with sampler-emitted `"tune"` stats which thereby become an optional, sampler-owned sampler-stat.

pymc/sampling/mcmc.py
```diff
@@ -1221,7 +1213,7 @@ def _sample(
     try:
         for it, stats in enumerate(sampling_gen):
             progress_manager.update(
-                chain_idx=chain, is_last=False, draw=it, stats=stats, tuning=it > tune
+                chain_idx=chain, is_last=False, draw=it, stats=stats, tuning=it < tune
```
michaelosthege, line 1299: is this an unrelated bugfix?

(if yes keep it, and make sure to mention it in the description)

michaelosthege (commented):
I like where this is going!

Using a slightly different naming I think we can simplify a bit more.

pymc/backends/mcbackend.py
eclipse1605: makes sense, ill make the change across drivers/backends/progress bar/tests.

pymc/sampling/mcmc.py
eclipse1605: > is this an unrelated bugfix?
> 
> (if yes keep it, and make sure to mention it in the description)

in the previous update def, the progress bar would flip the meaning and show “tuning” during posterior, and “not tuning” during warmup, plus it would be off by 1 around the boundary `it == tune` so I flipped it.

eclipse1605:
@michaelosthege does this make sense?

pymc/sampling/mcmc.py
```diff
@@ -1284,7 +1284,7 @@ def _iter_sample(
                 step.stop_tuning()
 
             point, stats = step.step(point)
-            trace.record(point, stats, tune=i < tune)
+            _record_with_in_warmup(trace, point, stats, in_warmup=i < tune)
```
michaelosthege: Inspecting the signature thousands of times creates a lot of unnecessary overhead.

I would prefer the breaking change of requiring custom trace backends to adapt immediately. (If there are any custom trace backends in existence at all.)

eclipse1605: true, but for what its worth the signature inspection isn’t happening per draw; it’s cached per trace-backend class. still, i agree the hot path shouldn’t be doing any of this, so i can drop the compatibility shim and treat this as a breaking change. as you said, custom trace backends will need to update their `record(..., *, in_warmup: bool)` signature. does that sound ok?

michaelosthege: Yes let's do it. Less code to maintain.

eclipse1605:
@michaelosthege check this out

pymc/step_methods/compound.py
michaelosthege: the whole function can be removed

pymc/step_methods/metropolis.py
```diff
@@ -351,7 +347,7 @@ def _make_progressbar_update_functions():
         def update_stats(step_stats):
             return {
                 "accept_rate" if key == "accept" else key: step_stats[key]
-                for key in ("tune", "accept", "scaling")
+                for key in ("accept", "scaling")
             }
```
michaelosthege, line 351: This part can now use the `in_warmup` stat?

pymc/step_methods/slicer.py
```diff
@@ -202,18 +200,17 @@ def competence(var, has_grad):
...
-        stats = {"tune": [True] * n_chains, "nstep_out": [0] * n_chains, "nstep_in": [0] * n_chains}
+        stats = {"nstep_out": [0] * n_chains, "nstep_in": [0] * n_chains}
 
         return columns, stats
 
     @staticmethod
     def _make_progressbar_update_functions():
         def update_stats(step_stats):
-            return {key: step_stats[key] for key in {"tune", "nstep_out", "nstep_in"}}
+            return {key: step_stats[key] for key in {"nstep_out", "nstep_in"}}
```
michaelosthege, line 214: same here?

tests/backends/test_mcbackend.py
```diff
@@ -293,13 +293,11 @@ def test_return_multitrace(self, simple_model, discard_warmup):
...
         assert isinstance(mtrace, pm.backends.base.MultiTrace)
-        tune = mtrace._straces[0].get_sampler_stats("tune")
-        assert isinstance(tune, np.ndarray)
+        # warmup is tracked by the sampling driver
         if discard_warmup:
-            assert tune.shape == (7, 3)
+            assert len(mtrace) == 7
         else:
-            assert tune.shape == (12, 3)
-        pass
```
michaelosthege, line 302: can this test remain as before, but using the `in_warmup` stat instead?

michaelosthege (commented):
I'm not familiar with how the progress bar gets updated. Possibly my two comments on that matter are invalid, but please check them.

I'll also trigger the CI tests

eclipse1605:
> I'm not familiar with how the progress bar gets updated. Possibly my two comments on that matter are invalid, but please check them.


hey, sorry for the delay but i think they're valid because warmup bookkeeping is now explicitly driver owned

eclipse1605:
@michaelosthege  ive made the tests consistent with the changes, running the ci tests again will mostly pass now

michaelosthege (approved):
Looks good to me!

Thanks @eclipse1605 for your endurance with this!

eclipse1605:
> Looks good to me!
> 
> Thanks @eclipse1605 for your endurance with this!

thanks a ton for the reviews and guidance @michaelosthege and @ricardoV94, really appreciate the patience since im still getting my bearings here :)

pymc/step_methods/metropolis.py
```diff
@@ -594,8 +586,6 @@ def __init__(
     ):
         model = pm.modelcontext(model)
 
-        # Doesn't actually tune, but it's required to emit a sampler stat
-        # that indicates whether a draw was done in a tuning phase.
         self.tune = True
```
ricardoV94: remove self.tune

pymc/step_methods/slicer.py
```diff
@@ -202,18 +200,17 @@ def competence(var, has_grad):
     @staticmethod
     def _progressbar_config(n_chains=1):
         columns = [
-            TextColumn("{task.fields[tune]}", table_column=Column("Tuning", ratio=1)),
```
ricardoV94, line 205: We still want to show this information in the task bar, where is this defined now?

tests/backends/test_arviz.py
```diff
@@ -799,7 +799,7 @@ def test_save_warmup_issue_1208_after_3_9(self):
                 idata = to_inference_data(trace[-30:], save_warmup=True)
             test_dict = {
                 "posterior": ["u1", "n1"],
-                "sample_stats": ["~tune", "accept"],
+                "sample_stats": ["~in_warmup", "accept"],
```
ricardoV94, line 803: I'm not sure about changing the output variable name, this seems like a breaking change for users?

The specific line I pointed to may not be relevant. The general question is whether we changed anything in MultiTrace/InferenceData output with this PR other than the `tune` flag not existing per step.

ricardoV94 (changes requested):

pymc/step_methods/slicer.py
eclipse1605: > We still want to show this information in the task bar, where is this defined now?

warmup now comes straight from `ProgressBarManager`, each task is initialised and updated with the driver’s `in_warmup` flag, so the column still renders. you can see the diff of `progress_bar.py`

tests/backends/test_arviz.py
eclipse1605: it now writes the warmup flag once as `in_warmup`, but for users, nothing new shows up. when we persist sampler stats (e.g. in mcbackend) we store that boolean and keep `trace.get_sampler_stats("tune")` working by aliasing to the new field. the default NDArray backend still omits both names, just like before. and `to_inference_data` continues to drop whichever warmup marker exists, so the resulting InferenceData matches main; the test only switches the "absent" check to the new internal name. no other MultiTrace/InferenceData variables changed.

ricardoV94 (approved):
This looks sleek, I just want to do a manual integration test locally before merging

eclipse1605:
> This looks sleek, I just want to do a manual integration test locally before merging

sounds good!

eclipse1605:
hey @ricardoV94 i tried to understand the failed test but didn't really get very far with it. is it failing because jax spits out NaNs when the dirichlet concentration is super skewed, so the multinomial never sees a clean prob vector?

ricardoV94:
That one fails now and then, don't worry about it

eclipse1605:
> That one fails now and then, don't worry about it

ok sure :)

ricardoV94:
Did I mislead you? Tune/Warmup was not shown in the progressbar before?

Top: Main branch
Bottom: This PR

image

Should revert then

eclipse1605:
> Did I mislead you? Tune/Warmup was not shown in the progressbar before?

on main only some steps injected a `Tuning` column via their `_progressbar_config`; NUTS never did. when we dropped the sampler emitted `tune` stat  those columns disappeared so I added the `Warmup` column to preserve that signal

ricardoV94:
Let's take it out. Don't want to make the bar too verbose

eclipse1605:
i have removed it @ricardoV94, we just wanted to remove it from the rendered table right?

ricardoV94:
There's still `in_warmup` logic in a couple of places in the progress bar. I assume that is no longer needed either

eclipse1605:
> There's still `in_warmup` logic in a couple of places in the progress bar. I assume that is no longer needed either

right, ill remove it from the internal progress bar tasks

ricardoV94 (approved):

tests/backends/test_mcbackend.py
ricardoV94: @eclipse1605 this comment still sounds relevant though

eclipse1605: as in what exactly? do we want to keep a tune alias test?

ricardoV94:
@eclipse1605 did you see the missed comment above about doing more minimal changes to the pre-existing test?

eclipse1605:
> @eclipse1605 did you see the missed comment above about doing more minimal changes to the pre-existing test?

do you mean [this](https://github.com/pymc-devs/pymc/pull/8015#discussion_r2686965719)

ricardoV94:
https://github.com/pymc-devs/pymc/pull/8015#discussion_r2657839967

eclipse1605:
i saw that, but i wanted a clarification as to whether we want to add an explicit tune alias assertion to preserve that compatibility

tests/backends/test_mcbackend.py
```diff
@@ -293,13 +294,20 @@ def test_return_multitrace(self, simple_model, discard_warmup):
...
+        in_warmup = mtrace.get_sampler_stats("in_warmup", combine=False, squeeze=False)
+        assert len(in_warmup) == 3
+        assert all(s.dtype == np.dtype(bool) for s in in_warmup)
+
+        # Warmup is tracked by the sampling driver and persisted via `in_warmup`.
         if discard_warmup:
-            assert tune.shape == (7, 3)
+            assert len(mtrace) == 7
+            assert all(len(s) == 7 for s in in_warmup)
+            assert all(not np.any(s) for s in in_warmup)
```
ricardoV94, line 305: What is this `in_warmup` object we're seeing here? From the test alone I have a hard time figuring out. Is it a numpy array?

pymc/backends/ndarray.py
```diff
@@ -97,7 +97,7 @@ def setup(self, draws, chain, sampler_vars=None) -> None:
                     new = np.zeros(draws, dtype=dtype)
                     data[varname] = np.concatenate([old, new])
 
-    def record(self, point, sampler_stats=None) -> None:
+    def record(self, point, sampler_stats=None, *, in_warmup: bool) -> None:
```
ricardoV94, line 100: Are we just ignoring the `in_warmup` in this function?

tests/backends/test_mcbackend.py
ricardoV94: It's unclear to me why this changed, it seemed like we just moved the source of `tune/warmup`, not the final stored contents?

pymc/backends/ndarray.py
michaelosthege: it's not used/needed here because this backend is not capable of separating between warmup/posterior.

Traditionally that was only done after the sampling loop - by splicing the multitrace.

tests/backends/test_mcbackend.py
eclipse1605: its an array returned by get_sampler_stats("in_warmup", combine=False, squeeze=False), so it’s a list of NumPy boolean arrays (one per chain). the contents are the same as the old tune stat, only the source/name changed because warmup is now driver owned.

we can still add a check like
```py
assert all(s.shape == (7,) for s in in_warmup)
```

pymc/backends/ndarray.py
eclipse1605: yeah it ignores  in_warmup because it cant separate warmup/posterior internally like mcbackend/zarr. we still split after sampling by splicing the MultiTrace.

eclipse1605:
@ricardoV94 any changes required in this?

tests/backends/test_mcbackend.py
ricardoV94: @eclipse1605 ^ comment

ricardoV94:
I have a question about why the test changed, thought the output would still be the same. Also we merged another PR so this one now has conflicts that need to be solved. Let me know if you need help

eclipse1605:
as i said above, the test changed because the tune sampler stat has been removed right. so the warmup tracking is now handled by the sampling driver, and the backend no longer stores tune. earlier in the test, tune was retrieved using get_sampler_stats("tune"), and its shape was checked to verify the number of warmup and posterior samples. 

the test asserted that tune was a NumPy array and checked its shape:
- if discard_warmup was True, the shape was (7, 3).
- if discard_warmup was False, the shape was (12, 3).

because warmup tracking is now managed directly by the sampling process, tune doesn't need to be stored in the backend, so the test cannot retrieve it. so instead of checking the shape of tune, it checks the length of the MultiTrace object (len(mtrace)) to determine the number of posterior samples. 

let me know if that helps clarify things.

eclipse1605:
also, if we move on to merging this, ill likely need help fixing the merge conflicts

jessegrabowski:
> also, if we move on to merging this, ill likely need help fixing the merge conflicts

I'm happy to help but I'd want to let #8047 go in first because it will change things around again.

eclipse1605:
> > also, if we move on to merging this, ill likely need help fixing the merge conflicts
> 
> I'm happy to help but I'd want to let #8047 go in first because it will change things around again.

sure, makes sense. let me know when we want to merge this given there are no more changes required :)

ricardoV94:
Thanks @michaelosthege. The failing tests are a known issue. I'll try to fix them after, but need not block this PR.

michaelosthege:
> Thanks @michaelosthege. The failing tests are a known issue. I'll try to fix them after, but need not block this PR.

😅 I just fixed them. Will do another PR then ;)

PR #8016, opened 2025-12-21
fix derived scan logprob when observed provides more broadcastable information

### LINKED ISSUE #7892: Derived scan logprob fails when observed data provides more broadcastable information than generative graph
### Description

Reported by @jessegrabowski 

```python
import numpy as np
import pytensor
import pymc as pm
from pymc.distributions.shape_utils import rv_size_is_none

def GRW(y_init, size=None):
    def grw_step(y_tm1):
        y = pm.Normal.dist(mu=y_tm1)
        return y, pm.pytensorf.collect_default_updates([y])
    
    if rv_size_is_none(size):
        n_steps = 10
    else:
        n_steps = size[0]

    y_hat, updates = pytensor.scan(fn=grw_step, outputs_info=[y_init], n_steps=n_steps)
    return y_hat

coords = {
    'date': range(10),
    'item': [1],
}
with pm.Model(coords=coords) as m:
    y0 = pm.Normal('y0', 0, 0.1, dims=['item'])
    y_hat = pm.CustomDist(
        'y_hat',
          y0,
          dist=GRW,
          dims=['date', 'item'],
          observed=np.ones((10, 1)),
    )

m.logp()  # TypeError: The broadcast pattern of the output of scan (Matrix(float64, shape=(?, 1))) is inconsistent with the one provided in `output_info` (Vector(float64, shape=(?,))). The output on axis 0 is `True`, but it is `False` on axis 1 in `output_info`. This can happen if one of the dimension is fixed to 1 in the input, while it is still variable in the output, or vice-verca. You have to make them consistent, e.g. using pytensor.tensor.specify_broadcastable.
```

We need to somehow introduce the correct broadcastable information during the logprob inference...

eclipse1605:
## Description

fixed a failure in derived scan logprob construction when the observed/value tensor provides more static broadcastability information than the generative scan graph (e.g. observed has a size-1 axis like `(date, 1)` while the scan state was inferred as non broadcastable on that axis).

in this, `model.logp()` could fail during the measurable scan rewrite with a scan `outputs_info` broadcast pattern mismatch (scan output inferred as matrix like vs. `outputs_info` expecting vector-like).

- the scan logprob rewrite propagate/align broadcastability metadata introduced by observed/value variables so the rewritten scan is internally consistent.
- this happens without turning scan inner placeholders into non nominal `Apply` nodes (so scan reconstruction remains valid).

### note
I think the same idea can be generalized by treating static broadcastability metadata as part of the measurable scan rewrite contract:
- identify the “outer” value/observed variables that participate in the measurable rewrite and extract their broadcastable axes (ignoring the time axis for sequences).
- when creating inner scan placeholders / `outputs_info` proxies for the logprob rewritten scan, ensure their `TensorType.shape` reflects any size-1/broadcastable axes implied by the outer variables.
- apply the same normalization to tapped init buffers (`pt.join` / `outputs_info`) so that init and scan outputs agree on broadcastability, without inserting broadcast `Apply` nodes into the inner graph (placeholders must remain nominal vars).

## Related Issue

- Closes [#7892](https://github.com/pymc-devs/pymc/issues/7892)

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/logprob/scan.py
+++ b/pymc/logprob/scan.py
@@ -41,6 +41,7 @@ from typing import cast
 import numpy as np
 import pytensor.tensor as pt
 
+from pytensor import clone_replace
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.rewriting.basic import node_rewriter
 from pytensor.scan.op import Scan
@@ -103,6 +104,38 @@ def convert_outer_out_to_in(
     """
     output_scan_args = copy(input_scan_args)
     inner_outs_to_new_inner_ins = {}
+    inner_var_replacements: dict[TensorVariable, TensorVariable] = {}
+
+    def ensure_inner_matches_outer_broadcastable(
+        inner_var: TensorVariable, outer_in_var: TensorVariable
+    ) -> TensorVariable:
+        """Return a nominal inner variable matching `outer_in_var` broadcastability.
+
+        `Scan.make_node` checks that outer sequence inputs and inner sequence
+        placeholders agree on static broadcastability (excluding the time axis).
+        Observed/value variables can introduce extra broadcastable axes (size-1
+        dimensions), so we may need to upgrade the inner placeholders.
+
+        We must not use `pt.specify_broadcastable` here, because that creates an
+        Apply node and the resulting variable is no longer a valid nominal inner
+        input (leading to MissingInputError when Scan builds its nominal graph).
+        """
+        inner_var = inner_var_replacements.get(inner_var, inner_var)
+        if outer_in_var.type.ndim != inner_var.type.ndim + 1:
+            return inner_var
+        new_shape = list(inner_var.type.shape)
+        changed = False
+        for axis in range(inner_var.type.ndim):
+            if outer_in_var.type.broadcastable[axis + 1] and not inner_var.type.broadcastable[axis]:
+                new_shape[axis] = 1
+                changed = True
+        if not changed:
+            return inner_var
+        new_type = inner_var.type.clone(shape=tuple(new_shape))
+        new_inner = new_type()
+        new_inner.name = inner_var.name
+        inner_var_replacements[inner_var] = new_inner
+        return new_inner
 
     # Map inner-outputs to outer-outputs
     old_inner_outs_to_outer_outs = {}
@@ -141,9 +174,12 @@ def convert_outer_out_to_in(
         var_idx = inner_out_info.index
 
         # The old inner-output variable becomes the a new inner-input
-        new_inner_in_var = old_inner_out_var.clone()
+        new_inner_in_var = old_inner_out_var.type()
         if new_inner_in_var.name:
             new_inner_in_var.name = f"{new_inner_in_var.name}_vv"
+
+        outer_in_var = new_outer_input_vars[oo_var]
+        new_inner_in_var = ensure_inner_matches_outer_broadcastable(new_inner_in_var, outer_in_var)
 
         inner_outs_to_new_inner_ins[old_inner_out_var] = new_inner_in_var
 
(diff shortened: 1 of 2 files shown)


ricardoV94:
I'm inclined to fix this on the Scan side in PyTensor. Basically if `inp.type.filter_variable(out.type)` passes we should be good to go. This would mean that `shape=(None,) -> (5,)` or `(None,)->(1,)` is fair game, just not `(5,)->(2,)` or something that changes number of dims. 

Furthemore `shape=(5)->(None,)` would end up with a specify_shape in the output and therefore `shape=(5,)->(5,)`. This is the same rule we use for rewrites.

Otherwise it's too tricky to work with scans programatically, and we end up with failures/awkward work-arounds as demonstrated in this PR

eclipse1605:
sorry for the slow reply, am i getting this right, implementation wise? we’d update PyTensor so that:  
- in [`check_broadcast`](https://github.com/pymc-devs/pytensor/blob/975ca888d457a1f5fb6346dfd2afc2afb7d1074c/pytensor/scan/op.py#L156), we only raise if the inner expects broadcastable axes that the outer doesn’t have, but we allow the outer to be more broadcastable/specific than the inner. 
- in [`Scan.validate_inner_graph`](https://github.com/pymc-devs/pytensor/blob/975ca888d457a1f5fb6346dfd2afc2afb7d1074c/pytensor/scan/op.py#L694), we consider input/output types compatible if the input type is a supertype (same dtype/ndim, output can be more specific), rather than requiring exact broadcastable equality.

eclipse1605:
@ricardoV94 does this sound right?

ricardoV94:
The inner should have as much information as the outer, the outre can have less. But yes. Let's open a PR and see how it goes

eclipse1605:
so in [`pytensor/pytensor/scan/op.py`](https://github.com/pymc-devs/pytensor/blob/main/pytensor/scan/op.py)

we change [`if b1 != b2`](https://github.com/pymc-devs/pytensor/blob/9834e96cee86d9f18436be5fa9e31c692bf6326f/pytensor/scan/op.py#L183) to `if b1 and not b2`

 and in [`validate_inner_graph`](https://github.com/pymc-devs/pytensor/blob/9834e96cee86d9f18436be5fa9e31c692bf6326f/pytensor/scan/op.py#L694-L727) instead of 

```py
if (
    type_input.dtype != type_output.dtype
    or type_input.broadcastable != type_output.broadcastable
):
```
we replace the broadcastable eq check with a compatibility check using `TensorType.is_super()`

```py
if type_input.dtype != type_output.dtype:
    # dtype error
elif isinstance(type_input, TensorType) and isinstance(type_output, TensorType):
    if not type_input.is_super(type_output):
        # compatibility error using is_super
else:
    if type_input != type_output:
        # fallback error
```

ricardoV94:
I think you can use `is_super` always, but let's try

eclipse1605:
@ricardoV94 i have openned a PR (https://github.com/pymc-devs/pytensor/pull/1861)

eclipse1605:
@ricardoV94 should i close this PR?

ricardoV94:
No, but it should be changed to just have a regression test for the original issue to make sure it's working once we fix it in PyTensor

eclipse1605:
yeah makes sense

eclipse1605:
@ricardoV94 once we fix it in pytensor do we just add the snippet used to repro the bug in https://github.com/pymc-devs/pymc/issues/7892 as a test?

ricardoV94:
> @ricardoV94 once we fix it in pytensor do we just add the snippet used to repro the bug in #7892 as a test?

Yes, plus we test the result is non-sensical when evaluated

PR #8019, opened 2025-12-23
Use arviz 1.0

### LINKED ISSUE #7821: sample_smc conversion of sample stats to idata can result in warning

aloctavodia:
## Description
This makes the changes to use arviz 1.0. Not all tests are passing yet. 

Closes #7821 

## Checklist

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):


aloctavodia:
Tests are passing locally, except for `test_mcbackend`. mcbackend needs to be updated to use `DataTree` instead of  `InferenceData`.

conda-envs/environment-alternative-backends.yml
```diff
@@ -5,7 +5,7 @@ channels:
 - defaults
 dependencies:
 # Base dependencies
-- arviz>=0.13.0
+- arviz>=1.0.0rc0
```
OriolAbril: if we need to run the test suite before we make the stable 1.0 release we can add arviz-plots here in the conda section and then this `arviz` indicator with the release candidate on the pip section. If we use the `rc` part in the version pin for pip it should automatically take pre-releases into account.

docs/source/conf.py
```diff
@@ -66,7 +66,7 @@
     "ndarray": ":class:`~numpy.ndarray`",
     "Covariance": ":mod:`Covariance <pymc.gp.cov>`",
     "Mean": ":mod:`Mean <pymc.gp.mean>`",
-    "InferenceData": ":class:`~arviz.InferenceData`",
+    "DataTree": ":class:`~xarray.DataTree`",
```
OriolAbril, line 69: This is also something I am unsure about for arviz docs but we might want to keep the alias around for `InferenceData` and maybe point to the schema page or a page about the schema implementation in python arviz. All "InferenceData" objects are compatible with `DataTree` but not the other way around so I think there is still value in using "InferenceData" in docs as in addition to saying the python type is `xarray.DataTree` it also says it will follow the schema (i.e. if a `posterior` group is present it has a specific meaning, `log_likelihood` group has that other meaning, `fit_info` group could also be there but has no convention agreed-on meaning)

pymc/backends/__init__.py
```diff
@@ -57,7 +57,7 @@
 Loading a saved backend
 -----------------------
 
-Saved backends can be loaded using `arviz.from_netcdf`
+Saved backends can be loaded using `arviz.convert_to_datatree`
```
OriolAbril: not sure if we have an issue to try and keep `from_netcdf` around as an alias to `xarray.open_datatree`. If we eventually didn't add the aliases I think it would be better to point users directly to `xarray.open_datatree` instead of `convert_to_datatree` (I am also not even sure this docstring ends up on the public docs but still better to be extra safe)

pymc/backends/arviz.py
```diff
@@ -82,7 +85,9 @@ def dict_to_dataset_drop_incompatible_coords(vars_dict, *args, dims, coords, **k
                     coords_lengths.pop(dim)
 
     # FIXME: Would be better to drop coordinates altogether, but arviz defaults to `np.arange(var_length)`
-    return dict_to_dataset(vars_dict, *args, dims=dims, coords=safe_coords, **kwargs)
+    return dict_to_dataset(
+        vars_dict, *args, dims=dims, coords=safe_coords, sample_dims=sample_dims, **kwargs
+    )
```
OriolAbril, line 90: I think this fixme is related to https://github.com/arviz-devs/arviz-base/issues/79 in case someone wants to lend a hand.

pymc/backends/arviz.py
```diff
@@ -285,89 +290,103 @@ def split_trace(self) -> tuple[Union[None, "MultiTrace"], Union[None, "MultiTrac
             trace_posterior = self.trace[self.ntune :]
         return trace_posterior, trace_warmup
 
+    @requires("trace")
     @requires("trace")
```
OriolAbril: the requires decorator ended up duplicated

pymc/backends/zarr.py
```diff
@@ -61,6 +61,9 @@
     _zarr_available = False
 
 
+WARMUP_TAG = "warmup_"
```
OriolAbril, line 63: should we keep this variable somewhere in arviz-base even if we delete the custom InferenceData class?

pymc/sampling/deterministic.py
```diff
@@ -99,9 +99,13 @@ def compute_deterministics(
 
     coords, dims = coords_and_dims_for_inferencedata(model)
 
+    is_datatree = hasattr(dataset, "to_dataset")
+    input_dataset = dataset.to_dataset() if is_datatree else dataset
```
OriolAbril: for the initial boolean variable maybe using isisnstance DataTree is clearer. For the second line, using `.dataset` would be better. Unless I misread the code, we are not modifying `input_dataset` so using a datasetlike view of the DataTree should be enough, which is what `.dataset` does. `.to_dataset()` creates an actual dataset with the same data. It doesn't copy the numpy array's themselves so `.to_dataset()` will generally be very fast so using one or the other won't make any real difference in this case, but better to stick to best practices when possible

pymc/sampling/forward.py
```diff
@@ -420,8 +419,8 @@ def sample_prior_predictive(
 
     Returns
     -------
-    arviz.InferenceData or Dict
-        An ArviZ ``InferenceData`` object containing the prior and prior predictive samples (default),
+    arviz.xarray.DataTree or Dict
```
OriolAbril:
```suggestion
    xarray.DataTree or dict
```

pymc/sampling/forward.py
```diff
@@ -828,17 +827,20 @@ def sample_posterior_predictive(
...
         if _constant_data is not None:
             trace_coords.update({str(k): v.data for k, v in _constant_data.coords.items()})
             constant_data.update({str(k): v.data for k, v in _constant_data.items()})
         idata = trace
         observed_data = trace.get("observed_data", None)
-        trace = trace["posterior"]
+        if "posterior" in trace.children:
+            trace = trace["posterior"].to_dataset()
+        else:
+            trace = trace.to_dataset()
```
OriolAbril: here I think should also be `.dataset`

pymc/sampling/forward.py
```diff
@@ -986,7 +988,7 @@ def sample_posterior_predictive(
     idata_pp = pm.to_inference_data(posterior_predictive=ppc_trace, **ikwargs)
 
     if extend_inferencedata and idata is not None:
-        idata.extend(idata_pp)
+        idata.update(idata_pp)
```
OriolAbril, line 978: I think `update` has the opposite behaviour `extend` used to have. The default in `extend` was a "left join". That is, whenever a group was present in both inputs the one in `idata` was kept, ignoring the one in `idata_pp`. `update` will take all groups in `idata_pp` and add them to `idata`, overwriting already present groups in `idata`. If this happens for `observed_data` which will generally be the conflicting one it doesn't matter in most cases but it might be worth to think about how we want things to behave while doing the update.

OriolAbril (commented):
I haven't finished to go over all the files, will try to finish soon

docs/source/learn/core_notebooks/dims_module.ipynb
```diff
@@ -19,9 +19,9 @@
     "\n",
```
OriolAbril, line 19: <p>The behaviour has changed here and I am not completely sure how or why.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/8019/discussion/#comment-2799254344'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='79'/>

docs/source/learn/core_notebooks/model_comparison.ipynb
```diff
@@ -9,7 +9,7 @@
      "name": "stdout",
```
OriolAbril, line 9: <p>this will need rerunning to show the right output but nothing to worry about</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/8019/discussion/#comment-2799254496'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='7'/>

OriolAbril, line 9: <p>extremely dumb but while we are here we should update the dictionary given to <code>az.compare</code> to use <code>hierarchical_loo</code> and <code>pooled_loo</code>.We have already computed those and stored the result, no need to recompute from the whole trace again.    </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/8019/discussion/#comment-2799254502'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='18'/>

pymc/backends/arviz.py
```diff
@@ -187,8 +192,8 @@ def insert(self, k: str, v, idx: int):
             self.trace_dict[k][idx, :] = v
 
 
-class InferenceDataConverter:
-    """Encapsulate InferenceData specific logic."""
+class DataTreeConverter:
+    """Encapsulate DataTree specific logic."""
```
OriolAbril: maybe it would be clearer to update this to something like: "Encapsulate conventions of InferenceData schema in DataTree conversion."

pymc/backends/arviz.py
```diff
@@ -464,7 +489,14 @@ def to_inference_data(self):
...
             id_dict["constant_data"] = self.constant_data_to_xarray()
-        idata = InferenceData(save_warmup=self.save_warmup, **id_dict)
+        id_dict = {k: v for k, v in id_dict.items() if v is not None}
+        idata = from_dict(
+            id_dict,
+            save_warmup=self.save_warmup,
+            coords=self.coords,
+            dims=self.dims,
+            sample_dims=self.sample_dims,
+        )
```
OriolAbril: Using `arviz_base.from_dict` means the warmup conventions are automatically enforced, but we are also repeating all the `dict_to_dataset` conversion from scratch (which has already happened in every `xyz_to_xarray` method.

I think it would be better to handle the filtering of warmup groups more manually and then call `DataTree.from_dict`

pymc/backends/arviz.py
```diff
@@ -632,12 +664,10 @@ def predictions_to_inference_data(
...
         return new_idata
     elif inplace:
-        concat([idata_orig, new_idata], dim=None, inplace=True)
+        idata_orig.update(new_idata)
         return idata_orig
     else:
-        # if we are not returning in place, then merge the old groups into the new inference
-        # data and return that.
-        concat([new_idata, idata_orig], dim=None, copy=True, inplace=True)
+        new_idata.update(idata_orig)
```
OriolAbril, line 684: the swapped order will take care of the copy/not copy, but it also means if `idata_orig` already has a `predictions` group for example it will overwrite the new one. I am not completely sure but I think with `concat` that would raise an error. In any case, now is the time for better behaviour.

IMO, anything other than `observed_data` should not be silently ignored/overwritten. How about adding explicit checks for group differences between `new_idata` and `idata_orig`?

I have no idea if this is used a bit or at all in the wild, so feel free to ignore or mark as extremely low priority.

pymc/backends/zarr.py
```diff
@@ -815,27 +818,30 @@ def split_warmup(self, group_name: str, error_if_already_split: bool = True):
...
-        ``InferenceData`` object. The only exception is that ``_sampling_state`` is
+        ``DataTree`` object. The only exception is that ``_sampling_state`` is
         excluded.
 
         Parameters
         ----------
         save_warmup : bool
             If ``True``, all of the warmup groups are stored in the inference data
             object.
+        eager : bool
```
OriolAbril, line 839: should we change this to `load` instead? It was the rcParams key and the xarray method being called.

pymc/backends/zarr.py
```diff
@@ -857,6 +863,8 @@ def to_inferencedata(self, save_warmup: bool = False) -> az.InferenceData:
...
             data = xr.open_zarr(store, group=name, mask_and_scale=False)
             attrs = {**data.attrs, **global_attrs}
-            data.attrs = make_attrs(attrs=attrs, library=pymc)
-            groups[name] = data.load() if az.rcParams["data.load"] == "eager" else data
-        return az.InferenceData(**groups)
+            data.attrs = make_attrs(attrs=attrs, inference_library=pymc)
+            groups[name] = data.load() if eager else data
+        return az.from_dict(
+            groups, save_warmup=save_warmup, coords=self.coords, dims=self.vars_to_dims
+        )
```
OriolAbril: Here I think it is also better suited for `DataTree.from_dict(groups)`, and it looks like we have already filtered out the warmup groups from being added to this dict.

pymc/sampling/forward.py
```diff
@@ -559,20 +558,20 @@ def sample_posterior_predictive(
...
     progressbar : bool
         Whether to display a progress bar in the command line. The bar shows the percentage
         of completion, the sampling speed in samples per second (SPS), and the estimated remaining
         time until completion ("expected time of arrival"; ETA).
     return_inferencedata : bool, default True
-        Whether to return an :class:`arviz:arviz.InferenceData` (True) object or a dictionary (False).
+        Whether to return an :class:`xarray:xarray.DataTree` (True) object or a dictionary (False).
     extend_inferencedata : bool, default False
-        Whether to automatically use :meth:`arviz.InferenceData.extend` to add the posterior predictive samples to
+        Whether to automatically use :meth:`xarray.DataTree.extend` to add the posterior predictive samples to
```
OriolAbril: this links to the method, it should either be removed completely or updated to point to `DataTree.update` instead.

pymc/sampling/forward.py
```diff
@@ -824,17 +823,20 @@ def sample_posterior_predictive(
...
+    if isinstance(trace, xarray.DataTree):
         _constant_data = getattr(trace, "constant_data", None)
         if _constant_data is not None:
             trace_coords.update({str(k): v.data for k, v in _constant_data.coords.items()})
             constant_data.update({str(k): v.data for k, v in _constant_data.items()})
         idata = trace
         observed_data = trace.get("observed_data", None)
-        trace = trace["posterior"]
+        if "posterior" in trace.children:
+            trace = trace["posterior"].to_dataset()
```
OriolAbril:
```suggestion
            trace = trace["posterior"].dataset
```

the other if branch uses it and I think we only need a view here so better to use `.dataset` only

pymc/sampling/mcmc.py
OriolAbril, line 419: there might be no test checking this case, but `.update` can't take `coords` and `dims` as keyword arguments, it only take a 2nd DataTree as single argument.

It looks like we have already used `dict_to_dataset` for `observed_data` and `constant_data` variables, so this should be:


```suggestion
        idata.update(
            xr.DataTree.from_dict({"constant_data": constant_data, "observed_data": observed_data})
        )
```

pymc/smc/sampling.py
```diff
@@ -267,14 +268,15 @@ def _save_sample_stats(
         sample_stats = dict_to_dataset(
             sample_stats_dict,
             attrs=sample_settings_dict,
-            library=pymc,
+            inference_library=pymc,
+            sample_dims=[],
```
OriolAbril: what are the contents of `sample_stats_dict` here? I assume this is where we ended up with the more chains than draws warning, but we weren't using `default_dims` before so I assume we do want the leading dims labeled as `chain` and `draw`, what we don't want is the warning. If that were the case it is better to use [check_conventions=False](https://python.arviz.org/projects/base/en/latest/api/generated/arviz_base.dict_to_dataset.html) instead which is a new argument.

If we do keep `sample_dims=[]` we might want to add some coords to get more sensible names in the output. This might be a good scenario for https://github.com/arviz-devs/arviz-base/issues/113 too

pymc/smc/sampling.py
```diff
@@ -267,14 +268,15 @@ def _save_sample_stats(
...
         )
 
         ikwargs: dict[str, Any] = {"model": model}
         if idata_kwargs is not None:
             ikwargs.update(idata_kwargs)
         idata = to_inference_data(trace, **ikwargs)
-        idata = InferenceData(**idata, sample_stats=sample_stats)  # type: ignore[arg-type]
+        idata = from_dict(idata)
```
OriolAbril: also better suited for `DataTree.from_dict` instead. Also, is the `sample_stats` group being lost now?

pymc/stats/log_density.py
```diff
@@ -201,7 +204,7 @@ def compute_log_density(
     )
 
     if extend_inferencedata:
-        idata.add_groups({f"log_{kind}": logdens_dataset})
+        idata.update({f"log_{kind}": DataTree(logdens_dataset)})
```
OriolAbril:
```suggestion
        idata[f"log_{kind}"] = logdens_dataset
```

given we are only ever adding a single group here I think this is clearer (and setitem in datatrees also works out of the box with datasets too)

pymc/testing.py
```diff
@@ -1108,33 +1107,34 @@ def mock_tree_depth(size: tuple[int, int]) -> NDArray:
...
-            .transpose("chain", "draw", ...)
-        )
+    posterior_ds = (
+        idata["prior"]
+        .to_dataset()
+        .isel(chain=0)
+        .expand_dims({"chain": range(n_chains)})
+        .transpose("chain", "draw", ...)
     )
+    idata.update({"posterior": xr.DataTree(posterior_ds)})
```
OriolAbril:
```suggestion
    idata["posterior"] = (
        idata["prior"]
        .to_dataset()
        .isel(chain=0)
        .expand_dims({"chain": range(n_chains)})
        .transpose("chain", "draw", ...)
)
```

pymc/testing.py
```diff
@@ -1108,33 +1107,34 @@ def mock_tree_depth(size: tuple[int, int]) -> NDArray:
...
+        posterior_ds = idata["posterior"].to_dataset()
+        sizes = posterior_ds.sizes
         size = (sizes["chain"], sizes["draw"])
         sample_stats_ds = xr.Dataset(
             {name: (("chain", "draw"), creator(size)) for name, creator in sample_stats.items()},
-            coords=idata["posterior"].coords,
+            coords=posterior_ds.coords,
         )
-        idata.add_groups(sample_stats=sample_stats_ds)
+        idata.update({"sample_stats": xr.DataTree(sample_stats_ds)})
```
OriolAbril:
```suggestion
        idata["sample_stats"] = sample_stats_ds
```

pymc/util.py
```diff
@@ -241,35 +240,43 @@ def enhanced(*args, **kwargs):
...
+    This function should be applied to an ``DataTree`` object obtained with
     ``pm.sample(keep_warning_stat=True)`` before trying to ``.to_netcdf()`` or ``.to_zarr()`` it.
     """
-    nidata = arviz.InferenceData(attrs=idata.attrs)
-    for gname, group in idata.items():
-        if "sample_stat" in gname:
+    tree_dict = {}
+
+    for gname, group in dt.items():
+        ds = group.ds if hasattr(group, "ds") else group
```
OriolAbril: in general it is better to check `isinstance(DataTree)` (which is what all the other updated files to). I am also not completely sure about the `.ds` status but I do know `.dataset` is recommended since DataTree moved to main xarray library.

tests/model/test_core.py
```diff
@@ -229,8 +229,8 @@ def test_nested_model_to_netcdf(self, tmp_path):
         with pm.Model("scope") as model:
             b = pm.Normal("var")
             trace = pm.sample(100, tune=0)
-        az.to_netcdf(trace, tmp_path / "trace.nc")
-        trace1 = az.from_netcdf(tmp_path / "trace.nc")
+        trace.to_netcdf(tmp_path / "trace.nc")
+        trace1 = az.convert_to_datatree(str(tmp_path / "trace.nc"))
```
OriolAbril: this is only on main, didn't make the release candidate but this can go back to `az.from_netcdf`

tests/sampling/test_forward.py
```diff
@@ -441,27 +440,26 @@ def test_length_coords_volatile(self):
...
+            {"posterior": {"x": np.array([[[np.pi] * 3]])}},
             coords={"trial": range(3)},
             dims={"x": ["trial"]},
         )
         with model:
             pp_same_len = pm.sample_posterior_predictive(
                 trace_same_len, var_names=["y"]
             ).posterior_predictive
-        assert pp_same_len["y"] == np.pi
+        assert pp_same_len["y"].values.item() == np.pi
```
OriolAbril:
```suggestion
        assert pp_same_len["y"].item() == np.pi
```

`.item()` on a DataArray already ends up as a scalar or error so no need for the intermediate `.values`

tests/sampling/test_forward.py
```diff
@@ -441,27 +440,26 @@ def test_length_coords_volatile(self):
...
+            {"posterior": {"x": np.array([[[np.pi] * 2]])}},
             coords={"trial": range(2)},
-            dims={"x": ["trial"]},
         )
         with model:
             pp_diff_len = pm.sample_posterior_predictive(
                 trace_diff_len, var_names=["y"]
             ).posterior_predictive
-        assert pp_diff_len["y"] != np.pi
+        assert pp_diff_len["y"].values.item() != np.pi
```
OriolAbril:
```suggestion
        assert pp_diff_len["y"].item() != np.pi
```

tests/sampling/test_forward.py
```diff
@@ -470,7 +468,7 @@ def test_length_coords_volatile(self):
             pp_diff_len_model_set = pm.sample_posterior_predictive(
                 trace_same_len, var_names=["y"]
             ).posterior_predictive
-        assert pp_diff_len_model_set["y"] != np.pi
+        assert pp_diff_len_model_set["y"].values.item() != np.pi
```
OriolAbril:
```suggestion
        assert pp_diff_len_model_set["y"].item() != np.pi
```

tests/backends/test_zarr.py
```diff
@@ -305,23 +303,26 @@ def test_record(model, model_step, include_transformed, draws_per_chunk):
...
     }
     if include_transformed:
-        expected_groups.add("unconstrained_posterior")
-        expected_groups.add("warmup_unconstrained_posterior")
-    assert set(idata.groups()) == expected_groups
-    for group in idata.groups():
+        expected_groups.add("/unconstrained_posterior")
+        expected_groups.add("/warmup_unconstrained_posterior")
+    assert set(idata.groups) == expected_groups
+    for group in idata.groups:
```
OriolAbril: this could be updated to `set(idata.children)` and `for group in idata.children` to keep the original group names, no leading / and no root "/" group either.

It feels very pedantic to point this out as it will do exactly the same, but I think patterns in pymc tests extend quite easily to other related libraries and would prefer not seeing this become the standard way of interacting with groups on DataTree objects.

tests/test_util.py
```diff
@@ -180,7 +182,7 @@ def test_drop_warning_stat():
     assert new.attrs.get("version") == "0.1.2"
 
     for gname in ["sample_stats", "warmup_sample_stats"]:
-        ss = new.get(gname)
+        ss = new[gname].ds
```
OriolAbril:
```suggestion
        ss = new[gname].dataset
```

aloctavodia:
<p>good catch!</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/8019/discussion/#comment-3891329615'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
<p>Mention the output is a datatree following the InferenceData schema and link to <a href="https://python.arviz.org/en/stable/schema/schema.html" target="_blank">https://python.arviz.org/en/stable/schema/schema.html</a> (I think will be <code>arviz:schema</code>)</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc/pull/8019/discussion/#comment-3891329519'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

pymc/sampling/forward.py
OriolAbril: Left to decide. What do we want the behaviour to be when `idata` already has a `posterior_predictive` or `predictions` group? I think either warning+overwrite or error, but not sure which.

The <6 behaviour was no warning, no error and no overwrite, the original `poisterior_predictive` group in `idata` was preserved, and the one in `idata_pp` was ignored. With these changes in >=6 it would be overwritten so the one in `idata` is silently ignored and the one in `idata_pp` is kept.

pymc/backends/arviz.py
OriolAbril: we have the same question above

ricardoV94:
@aloctavodia this can be directed to the v6 branch now and we can merge as soon as it's ready

williambdean:
whats the plans with various `return_inferencedata`, `extend_inferencedata`, `idata_kwargs`, etc? Is the term "InferenceData" sticking around as just a DataTree that take a specific form?

aloctavodia:
I think that's something up for discussion. While the object is now a `DataTree`, in the context of PyMC/ArviZ, we still follow the [InferenData schema](https://python.arviz.org/en/latest/schema/schema.html), so we may want to keep the InferenceData name.

conda-envs/windows-environment-test.yml
```diff
@@ -5,12 +5,13 @@ channels:
 - nodefaults
 dependencies:
 # Base dependencies (see install guide for Windows)
-- arviz>=0.13.0,<1.0
+- arviz>=1.0.0
```
ricardoV94: Everywhere:
```suggestion
- arviz>=1.0.0,2
```

pymc/sampling/jax.py
```diff
@@ -668,10 +660,8 @@ def sample_jax_nuts(
     )
     tic2 = datetime.now()
 
-    if idata_kwargs is None:
-        idata_kwargs = {}
-    else:
-        idata_kwargs = idata_kwargs.copy()
+    # TODO: Add a better way to do idata_kwargs that is better combatible with Arviz 1.0
+    idata_kwargs = {}
```
ricardoV94: is this doing anything?

pymc/sampling/mcmc.py
```diff
@@ -421,6 +422,12 @@ def _sample_external_nuts(
     elif sampler in ("numpyro", "blackjax"):
         import pymc.sampling.jax as pymc_jax
 
+        if idata_kwargs is not None:
+            warnings.warn(
+                "`idata_kwargs` are currently ignored by the nutpie sampler",
```
ricardoV94: wrong message, this is the jax samplers branch

pymc/sampling/jax.py
zaxtax: It's there so I don't have to totally rip out the log_likelihood code for when we put back this functionality. Happy to remove all the dead-code if we want to leave idata_kwargs for a future PR

zaxtax: Especially if we want to be nudging people towards `compute_log_likelihood`

ricardoV94: I think there's too much muscle memory with compute_log_likelihood that we want to not break

zaxtax: But do we want to keep supporting `idata_kwargs={"log_likelihood": False}` as an argument `sample_jax_nuts` ?

With the way the PR is now, log_likelihood is not added to the idata and you have to explicitly call compute_log_likelihood later to get that group. What's the desired behaviour here?

ricardoV94: retain the old behavior with a message of how to proceed if we want to deprecate. These changes don't make the behavior harder to implement, it's as awkward as before?

aloctavodia:
This LGTM! Thanks @zaxtax for finishing it!!!

docs/source/conf.py
ricardoV94: Yeah I think DataTree is too generic for us. Need not be a blocker to merge right now, but internally we may want to keep talking about inferencedata, even if it's a different structure now

pymc/backends/arviz.py
ricardoV94: Tracking in https://github.com/pymc-devs/pymc/issues/8233

pymc/sampling/forward.py
ricardoV94: I think overwrite+warning or just overwrite, dropping out the new data made no sense

pymc/sampling/jax.py
```diff
@@ -674,6 +671,12 @@ def sample_jax_nuts(
         idata_kwargs = idata_kwargs.copy()
 
     if idata_kwargs.pop("log_likelihood", False):
+        warnings.warn(
+            "`passing log_likelihood` is deprecated and will be removed in future versions. Use "
+            ":func:`pymc.compute_log_likelihood` instead.",
+            DeprecationWarning,
```
ricardoV94: DerprecationWarning are not visible by default

```suggestion
            FutureWarning,
```

ricardoV94: also stacklevel should be at least 2

ricardoV94: Also I need some context why is this path affected but not the same in `pm.sample`?

ricardoV94 (changes requested):
Left two comments above (in the same place), mainly use FutureWarning, and a question about why the log_likelihood changes only touched the jax sampler, while regular pm.sample also has(had?) that option

pymc/sampling/jax.py
zaxtax: @ricardoV94 pm.sample has this logic in `_sample_return` which the external nuts samplers bypass. Would you like a `FutureWarning` there as well?

ricardoV94: it should be the same approach, regardless of where it's handled from

ricardoV94 (approved):
Test failing but otherwise LGTM

PR #8020, opened 2025-12-23
Pre-sampling validation for shape mismatches in PyMC

### LINKED ISSUE #7891: Check idata dims/coords for consistency before sampling begins

arec1b0:
# Pre-sampling validation for shape mismatches in PyMC

## Description
This PR implements pre-sampling validation to catch shape mismatches between model coordinates and variable dimensions **before** sampling begins, preventing potentially hours of wasted computation time.

Previously, PyMC would happily sample for extended periods only to fail during the final `InferenceData` conversion step when shape inconsistencies existed between model `coords` and variable `dims`.  
This PR adds a validation function that runs a smoke test **before** sampling starts, evaluating variable shapes and comparing them against declared dimensions to catch errors early.

---

### 1. Added `_validate_idata_conversion()` function in `mcmc.py`
- Uses PyTensor’s shape compilation to evaluate **actual** variable shapes  
- Compares declared `dims` against coordinate lengths  
- Provides **clear, actionable** error messages indicating exactly which variable has the problem and how to fix it  
- **Minimal overhead**: single shape evaluation, no sampling  

### 2. Integrated validation into `sample()` function
- Automatically runs when `return_inferencedata=True` (the default)  
- Executes **after** initial points are created but **before** any sampling begins  
- Gracefully handles errors without breaking existing workflows  

### 3. Comprehensive test suite in `test_issue_7891.py`
- Tests the exact scenario from the issue (deterministic with wrong `dims`)  
- Validates that **correct** models still work normally  
- Covers edge cases (scalar variables, multi-dimensional arrays)  
- Ensures **helpful** error messages are produced  

---

## Benefits
✅ **Saves computation time**: Catches errors immediately instead of after hours of sampling  
✅ **Better UX**: Clear error messages tell users exactly what’s wrong and how to fix it  
✅ **Minimal overhead**: Fast shape evaluation without actual sampling  
✅ **Backwards compatible**: Only runs when creating `InferenceData`; doesn’t affect existing code  
✅ **Comprehensive**: Works with scalars, multi-dimensional variables, and deterministics  

---

## Example

**Before this PR**, the following model would sample for hours then fail:

```python
with pm.Model(coords={"obs": range(10)}) as m:
    μ = pm.Normal("μ")
    y = pm.Deterministic("y", μ, dims="obs")  # shape mismatch!
    trace = pm.sample()
```

**After this PR**, it **fails immediately** with a helpful error:

```
ValueError: Variable 'y' has shape () but its dimension 'obs' implies length 10.
Verify that the variable's shape matches the coordinate length.
```

---

## Related Issue
Closes [#7891](https://github.com/pymc-devs/pymc/issues/7891)

---

## Checklist
- [x] Added necessary documentation (docstrings in the validation function)  
- [x] Each commit corresponds to a relevant logical change  

---

## Type of change
- New feature / enhancement  
- Bug fix (prevents wasted computation from preventable errors)

### DIFF
--- a/pymc/sampling/mcmc.py
+++ b/pymc/sampling/mcmc.py
@@ -857,6 +857,11 @@ def sample(
         if isinstance(step, list):
             step = CompoundStep(step)
 
+    # Validate that InferenceData conversion will work before sampling
+    # This catches shape/dimension mismatches early to avoid wasting compute time
+    if return_inferencedata:
+        _validate_idata_conversion(model, initial_points[0], idata_kwargs or {})
+
     if var_names is not None:
         trace_vars = [v for v in model.unobserved_RVs if v.name in var_names]
         trace_vars = model.replace_rvs_by_values(trace_vars)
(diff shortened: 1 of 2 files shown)


ricardoV94:
As linked and discussed to in the original issue the alternative to just fail gracefully is being investigated, specifically https://github.com/pymc-devs/pymc/pull/7912

The check proposed in this PR seems rather inefficient and will incur an overhead that would at least have to be benchmarked.

Also the LLM "summary" with 100 lines of code is not useful at all. Please summarize better if the LLM can't do it itself.

PR #8031, opened 2026-01-03
test: document empty input behavior in get_value_vars_from_user_vars

jyotihotwani25:
## Description
This PR adds a test documenting that `get_value_vars_from_user_vars([], model)` returns an empty list when given empty input.

The test reflects the current behavior and improves coverage for an edge case without changing any functionality.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -240,6 +240,9 @@ def test_get_value_vars_from_user_vars():
         x2 = pm.Normal("x2", mu=0, sigma=1)
         y2 = pm.Normal("y2", mu=0, sigma=1)
         det2 = pm.Deterministic("det2", x2 + y2)
+        
+    with pm.Model() as model:
+        assert get_value_vars_from_user_vars([], model) == []
 
     prefix = "The following variables are not random variables in the model:"
     with pytest.raises(ValueError, match=rf"{prefix} \['x2', 'y2'\]"):
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -240,7 +240,7 @@ def test_get_value_vars_from_user_vars():
         x2 = pm.Normal("x2", mu=0, sigma=1)
         y2 = pm.Normal("y2", mu=0, sigma=1)
         det2 = pm.Deterministic("det2", x2 + y2)
-        
+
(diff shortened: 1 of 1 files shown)


ricardoV94:
Why do you feel this is needed?

tests/test_util.py
```diff
@@ -241,6 +241,9 @@ def test_get_value_vars_from_user_vars():
         y2 = pm.Normal("y2", mu=0, sigma=1)
         det2 = pm.Deterministic("det2", x2 + y2)
 
+    with pm.Model() as model:
+        assert get_value_vars_from_user_vars([], model) == []
```
ricardoV94: You can add it in the first block of checks from the first model. It doesn't need to be tested with an empty model?

jyotihotwani25: The reason for adding this check is to explicitly document the current contract that empty input is a valid case and returns an empty list, rather than failing implicitly or raising an error. Without an explicit test, this behavior could
easily change unintentionally during refactors without being caught.

I agree that the behavior itself does not depend on model contents, and placing the assertion alongside the existing `model1` checks makes the intent clearer. I’ve moved it there.

I’m still getting familiar with the PyMC codebase, if there are particular parts you’d recommend focusing on early when exploring the codebase, I’d appreciate any pointers.
Thanks for the review.

ricardoV94 (approved):

ricardoV94:
I'm approving this one @jyotihotwani25 , but I feel it's a code-change looking for a reason rather than the other way around. Python is loosely typed and any function that accepts/returns containers will have empty edge case. I don't believe we should test it all just "because". I suggest looking for code changes that are meaningful to the users for your next steps (if you so wish to proceed), or actual bug-fixes.

Rule-of-thumb: if there's no issue open for it, it's probably not a good first issue

jyotihotwani25:
Thanks for the approval and the feedback. I’ll focus future contributions on more clearly user-facing and impactful changes.
I Appreciate the guidance.

Issue #7861, opened 2025-07-22
Create automatic docstrings for dims module from original distributions


ricardoV94:
> These currently end up having an empty docstring: https://pymcio--7820.org.readthedocs.build/projects/docs/en/7820/api/dims/generated/pymc.dims.Flat.html. Is it possible to dynamically copy the docstring from the "regular" distribution?
> 
> For vector distributions or transforms we already have specific docstrings (and I think we want that there) but scalar ones can probably share the docstring after a search/replace of `tensor_like` -> `xtensor_like`

_Originally posted by @OriolAbril in https://github.com/pymc-devs/pymc/pull/7820#discussion_r2204912560_

aman-coder03:
Hi @ricardoV94  
I am interested in working on this.

MSK-005:
@ricardoV94 and @OriolAbril. 
Looking at `dims/distributions/scalar.py`, we could import the classes from `distributions/continuous.py`, and then use a simple loop at the end of `scalar.py` to copy the docstrings using the `__doc__` attribute, and of course, replace `tensor_like` with `xtensor_like`.

However, for that, we will need to import many classes from `distributions/continuous.py`, which might make for messy code and violate the purpose of the file. May we create another file in the `dims/distributions` folder with the name of `docstrings.py`?

Lastly, although many classes can have the same docstrings copied to them, I think one or two might require a manual rewrite to adjust for the parameters for the dims classes.

I'd like to discuss this before I open a PR. Thanks.

ricardoV94:
I think it's possible to import a module as an alias and later delete the alias, something like: `import pymc.distributions as _regular_dists -> ... do stuff with it -> del _regular_dists`

MSK-005:
@ricardoV94, ready for review: [PR #8041](https://github.com/pymc-devs/pymc/pull/8041)

PR #8041, opened 2026-01-05
Add Docstrings for pymc.dims.distributions.scalar

MSK-005:
## Description

This PR (for [#7861](https://github.com/pymc-devs/pymc/issues/7861)) adds a function `create_scalar_dims_docstrings()` that copies the docstrings from all relevant classes in `pymc.distributions.continuous` to `pymc.dims.distributions.scalar`, replacing "tensor_like" with "xtensor_like" where applicable.
## Related Issue

- [x] Closes [#7861](https://github.com/pymc-devs/pymc/issues/7861)
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/dims/distributions/scalar.py
+++ b/pymc/dims/distributions/scalar.py
@@ -194,3 +194,35 @@ class InverseGamma(PositiveDimDistribution):
                 "Incompatible parameterization. Either use alpha and (optionally) beta, or mu and sigma"
             )
         return super().dist([alpha, beta], **kwargs)
+
+
+def create_scalar_dims_docstrings():
+    import inspect
+    import sys
+
+    import pymc.distributions.continuous as _regular_dists
+
+    from pymc.distributions import Continuous
+
+    # Get all subclasses of Continuous class
+    imported_dists = {
+        name: cls
+        for name, cls in inspect.getmembers(_regular_dists, inspect.isclass)
+        if issubclass(cls, Continuous) and cls is not Continuous
+    }
+
+    # Get all classes declared in this file
+    dims_dists = {
+        name: cls
+        for name, cls in inspect.getmembers(sys.modules[__name__], inspect.isclass)
+        if issubclass(cls, DimDistribution) and cls.__module__ == __name__
+    }
+
+    # Copy docstring from regular distribution to dims distribution
+    for dist_class_name in dims_dists:
+        imported_cls = imported_dists.get(dist_class_name)
+        dims_cls = dims_dists.get(dist_class_name)
+        if imported_cls and imported_cls.__doc__ and dims_cls.__doc__ is None:
+            dims_cls.__doc__ = imported_cls.__doc__.replace("tensor_like", "xtensor_like")
+
+    del _regular_dists, Continuous, inspect, sys
--- a/pymc/dims/distributions/scalar.py
+++ b/pymc/dims/distributions/scalar.py
@@ -22,19 +22,15 @@ from pymc.dims.distributions.core import (
     DimDistribution,
     PositiveDimDistribution,
     UnitDimDistribution,
+    copy_docstring,
 )
 from pymc.distributions.continuous import Beta as RegularBeta
 from pymc.distributions.continuous import Gamma as RegularGamma
 from pymc.distributions.continuous import HalfCauchyRV, HalfStudentTRV, flat, halfflat
 
 
-def _copy_docstring(regular_cls):
-    def get_regular_docstring(dims_cls):
-        if regular_cls and regular_cls.__doc__ and dims_cls.__doc__ is None:
-            dims_cls.__doc__ = regular_cls.__doc__.replace("tensor_like", "xtensor_like")
-        return dims_cls
-
-    return get_regular_docstring
+def _change_docstring_words(docstring):
+    return docstring.replace("tensor_like", "xtensor_like")
 
 
 def _get_sigma_from_either_sigma_or_tau(*, sigma, tau):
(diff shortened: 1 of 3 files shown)


MSK-005:
@ricardoV94, looking forward to your code review.

pymc/dims/distributions/scalar.py
```diff
@@ -194,3 +194,35 @@ def dist(cls, alpha=None, beta=None, *, mu=None, sigma=None, **kwargs):
...
+    import pymc.distributions.continuous as _regular_dists
+
+    from pymc.distributions import Continuous
+
+    # Get all subclasses of Continuous class
+    imported_dists = {
+        name: cls
+        for name, cls in inspect.getmembers(_regular_dists, inspect.isclass)
+        if issubclass(cls, Continuous) and cls is not Continuous
+    }
```
ricardoV94: Small performance idea: I guess we can get `dims_dists` first and do a more targetted filter (matches name -> is a subclass of `Distribution`)?

ricardoV94:
Seems to work for the docs: https://pymcio--8041.org.readthedocs.build/projects/docs/en/8041/api/dims/generated/pymc.dims.HalfCauchy.html

pymc/dims/distributions/scalar.py
MSK-005: Done.

pymc/dims/distributions/scalar.py
```diff
@@ -194,3 +194,34 @@ def dist(cls, alpha=None, beta=None, *, mu=None, sigma=None, **kwargs):
                 "Incompatible parameterization. Either use alpha and (optionally) beta, or mu and sigma"
             )
         return super().dist([alpha, beta], **kwargs)
+
+
+def create_scalar_dims_docstrings():
```
ricardoV94: Sorry to rethink again.

What if instead we use a decorator on the specific classes or manually do `Univariate.__doc__ = foo(regular_dists.Univariate.__doc__)`? So we don't need to loop and try to guess what goes with what. This also avoids accidentally overriding custom docstrings if we add them in the future.

MSK-005: Went forward with the decorator approach. Also, [pm.dims.Univariate.dist](https://pymcio--8041.org.readthedocs.build/projects/docs/en/8041/api/dims/generated/classmethods/pymc.dims.Normal.dist.html) has no documentation. What do we do about that?

pymc/dims/distributions/scalar.py
```diff
@@ -26,6 +28,15 @@
 from pymc.distributions.continuous import HalfCauchyRV, HalfStudentTRV, flat, halfflat
 
 
+def _copy_docstring(regular_cls):
```
ricardoV94: Makes more sense to have this helper in `pymc/dims/distributions/core.py`?

pymc/dims/distributions/scalar.py
ricardoV94: The dist method having no documentation? It's fine? They don't have in regular distributions either if I recall correctly

pymc/dims/distributions/scalar.py
MSK-005: Done. And I added a `change_doc` parameter that will replace "tensor_like" with "xtensor_like". The purpose was to provide flexibility if you want to modify docstrings before you copy them.

pymc/dims/distributions/scalar.py
MSK-005: Here's the dist() page for pm.Normal.dist: [link](https://www.pymc.io/projects/docs/en/stable/api/distributions/generated/classmethods/pymc.Normal.dist.html)

Here's the dist() page for pm.dims.Normal.dist: [link](https://www.pymc.io/projects/docs/en/latest/api/dims/generated/classmethods/pymc.dims.Normal.dist.html)

ricardoV94: I think it's just showing the docstrings of the base class, nothing specific about the Normal dist

pymc/dims/distributions/scalar.py
```diff
@@ -16,16 +16,23 @@
...
     UnitDimDistribution,
+    copy_docstring,
 )
 from pymc.distributions.continuous import Beta as RegularBeta
 from pymc.distributions.continuous import Gamma as RegularGamma
 from pymc.distributions.continuous import HalfCauchyRV, HalfStudentTRV, flat, halfflat
 
 
+def _change_docstring_words(docstring):
+    return docstring.replace("tensor_like", "xtensor_like")
```
ricardoV94: still think it should be in `core.py`. Also let's leave the argument flexibility for when we actually need it (which may be never). It's an internal function so we don't need to worry about changing API

MSK-005: Done. Sorry about that.

ricardoV94: There's nothing to apologize about

ricardoV94 (approved):

ricardoV94:
This is pretty sweet, thanks @MSK-005

MSK-005:
@ricardoV94, thank you for your guidance. Truly a blessing. All tests pass. Also a blessing.
Will appear in more PRs, hopefully.

OriolAbril (approved):
Looks great!

Two nits which can perfectly be tackled in a future PR would be adding `xtensor_like` to the glossary like [tensor_like](https://pymcio--8041.org.readthedocs.build/projects/docs/en/8041/glossary.html#term-tensor_like) and to the type aliases.

Sources for these two things: https://github.com/pymc-devs/pymc/blob/main/docs/source/glossary.md and variable `numpydoc_xref_aliases` of https://github.com/pymc-devs/pymc/blob/main/docs/source/conf.py#L63

Issue #8039, opened 2026-01-05
Remove `model.check_bounds` in favor of graph introspection


ricardoV94:
### Description

In the following graph:

```python
import pymc as pm

with pm.Model(check_bounds=True):
  sigma = pm.HalfNormal("sigma")
  y = pm.Normal("y", sigma=sigma)
m.compile_logp().fn.dprint()
```

There ends up being a `switch(exp(i0) > 0, something, -inf)` that disappears with `check_bounds=False` (one other stays regardless).

```
    ├─ Switch [id E]
    │  ├─ GE [id F]
    │  │  ├─ exp [id G] 't1'
    │  │  │  └─ i0 [id H]
    │  │  └─ 0.0 [id I]
    │  ├─ add [id J]
    │  │  ├─ -0.22579135264472738 [id K]
    │  │  └─ mul [id L]
    │  │     ├─ -0.5 [id M]
    │  │     └─ sqr [id N]
    │  │        └─ exp [id G] 't1'
    │  │           └─ ···
    │  └─ -inf [id O]
```

We could rewrite `exp(x) >= 0) -> True` in PyTensor. The one case it's not correct is if it's a `nan` originally. We could tag such rewrite as `nan_unsafe`, or we could promise pytensor that we are not going to propose `nan` for x. I prefer the first approach.

But some of the switches are more complex, such as IntervalTransform(0, 1), for unit variables. We could however check that `rvs_to_transforms` has the right transform that ensures the constraint, and remove the check like `check_bounds` does. It may be easier to do that, than to verify the computational graph implies the constraints.

These are a bit myopic points. The general idea is to make `check_bounds` redundant when a graph is well defined, and remove that flag from the model. Streamlining the UX

Utkarshkarki:
I am interested in this problem and would give my insights or solution soon.

ricardoV94:
This is not a beginner friendly or even fully defined issue. I suggest not picking this up

Utkarshkarki:
I see two paths:
 (1) implement smart check removal in PyMC using  rvs_to_transforms
 (2) Contribute mathematical rewrites to PyTensor first. 
Which would you prefer? Or should we do a hybrid approach—start with PyMC and contribute to PyTensor in parallel?  

If you think my way of understanding is correct, we can move ahead with this one issue.

Utkarshkarki:
but writing rewrites would be little tedious process

ricardoV94:
[...]

Issue #8038, opened 2026-01-05
`check_bounds` rewrite should be implemented in the model not in `compile`


ricardoV94:
### Description

This is mixing behavior. `check_bounds` is mostly a rewrite to avoid `ParameterCheckBounds` in the logp-related functions obtained from the model. Therefore it should be the one handling it.

https://github.com/pymc-devs/pymc/blob/056e80c716c35d30a37fe51de6e254a7041ad5b2/pymc/pytensorf.py#L948-L958

We may however decide to remove the functionality altogether. See ...

Chirag3841:
@ricardoV94 Here we can  make check_bounds explicit input to check_bounds() and never read modelcontext inside.
Now  compile function will directly override the behaviour.
Previously, it always depended on the model context.

ricardoV94:
I don't have full picture on how to proceed on this tbh. Better to leave the issue dormant

PR #8046, opened 2026-01-10
Feature/transform aware bounds checking

Utkarshkarki:
This PR makes bounds checking in PyMC **transform-aware**, allowing redundant runtime checks
(`CheckParameterValue → Switch(condition, logp, -inf)`) to be safely skipped when the associated
constraint is already enforced by a variable transform (e.g., `HalfNormal` with `LogTransform`).

The change introduces optional metadata on `CheckParameterValue` and a transform-aware rewrite
used during `compile()`. Checks are removed **per parameter**, never globally, and only when a
transform provably guarantees the constraint. Existing behavior is preserved when metadata
is absent.

Distribution `logp` implementations are intentionally left unchanged; guidance is requested on
the preferred way to attach RV context in a follow-up.

## Related Issue
- [ ] Closes #
- [x] Related to #8039

## Checklist
- [x] Included tests covering transformed and untransformed parameters
- [ ] Added necessary documentation (not required for infrastructure change)
- [x] Each commit corresponds to a logical change

## Type of change
- [x] Maintenance
- [ ] New feature/enhancement
- [ ] Bug fix
- [ ] Documentation
- [ ] Other (please specify):

### DIFF
--- a/pymc/logprob/utils.py
+++ b/pymc/logprob/utils.py
@@ -183,7 +183,7 @@ class CheckParameterValue(CheckAndRaise):
     """Implements a parameter value check in a logprob graph.
 
     Raises `ParameterValueError` if the check is not True.
-    
+
     Parameters
     ----------
     msg : str
@@ -198,21 +198,46 @@ class CheckParameterValue(CheckAndRaise):
         Type of constraint: "positive", "unit_interval", "interval", etc.
     """
 
-    __props__ = ("msg", "exc_type", "can_be_replaced_by_ninf", "rv", "param_idx", "constraint_type")
+    __props__ = ("msg", "exc_type", "can_be_replaced_by_ninf", "param_idx", "constraint_type")
 
     def __init__(
-        self, 
-        msg: str = "", 
+        self,
+        msg: str = "",
         can_be_replaced_by_ninf: bool = False,
-        rv: typing.Any = None,
         param_idx: int | None = None,
         constraint_type: str | None = None,
     ):
         super().__init__(ParameterValueError, msg)
         self.can_be_replaced_by_ninf = can_be_replaced_by_ninf
-        self.rv = rv
         self.param_idx = param_idx
         self.constraint_type = constraint_type
+
+    def make_node(self, value, cond, rv=None):
+        """Create a CheckParameterValue node.
+
+        The first input is the value to return, the second is the condition,
+        and the third (optional) is the RV metadata.
+        """
+        from pytensor.scalar.basic import as_scalar
+
+        value = pt.as_tensor_variable(value)
+        cond = as_scalar(cond)
+        if cond.dtype != "bool":
+            cond = cond.astype("bool")
+
+        inputs = [value, cond]
+        if rv is not None:
+            rv = pt.as_tensor_variable(rv)
+            inputs.append(rv)
+
+        return Apply(self, inputs, [value.type()])
+
+    def perform(self, node, inputs, outputs):
+        (out,) = outputs
+        val, cond, *remainder = inputs
+        out[0] = val
+        if not cond:
+            raise self.exc_type(self.msg)
 
     def __str__(self):
         """Return a string representation of the object."""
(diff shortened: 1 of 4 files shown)


pymc/logprob/utils.py
```diff
@@ -183,13 +183,36 @@ class CheckParameterValue(CheckAndRaise):
...
+        self, 
+        msg: str = "", 
+        can_be_replaced_by_ninf: bool = False,
+        rv: typing.Any = None,
+        param_idx: int | None = None,
+        constraint_type: str | None = None,
+    ):
         super().__init__(ParameterValueError, msg)
         self.can_be_replaced_by_ninf = can_be_replaced_by_ninf
+        self.rv = rv
```
ricardoV94: self.rv is not a good strategy. We don't want state at the Op level like this, if it's needed it should be in the graph as an extra input.

Example if we do `model.clone()` it will now reference the old RV, not the cloned one

pymc/pytensorf.py
```diff
@@ -946,16 +946,26 @@ def compile(
...
+        rvs_to_transforms = model.rvs_to_transforms
     except TypeError:
         check_bounds = True
-    check_parameter_opt = (
-        "local_check_parameter_to_ninf_switch" if check_bounds else "local_remove_check_parameter"
-    )
+        rvs_to_transforms = {}
+    
+    # Use transform-aware rewrite if we have transforms and check_bounds is enabled
+    if check_bounds and rvs_to_transforms:
```
ricardoV94: logic on transforms / bounds should be moved away from `pytensorf.compile` and `model.logp` or `model.fn` (#8038)

tests/distributions/test_simulator.py
```diff
@@ -98,15 +96,13 @@ def test_one_gaussian(self, seeded_test):
...
-                    and not isinstance(get_default_mode().linker, NumbaLinker),
-                    reason="Needs investigation",
+                    condition=sys.version_info.minor == 14, reason="Needs investigation"
                 ),
             ),
             "float64",
         ],
     )
-    def test_custom_dist_sum_stat(self, seeded_test, floatX):
+    def test_custom_dist_sum_stat(self, floatX):
```
ricardoV94: This branch is picking/ starting from unrelated changes to this PR

Utkarshkarki: I've rebased this PR on the latest upstream/main to remove unrelated commits. 
The code changes remain the same—only the commit history is now clean.

Utkarshkarki (commented):
Hi @ricardoV94,

Implemented transform-aware bounds checking (#8039) with your #8038 feedback:

- ✅ Moved check_bounds logic from `pytensorf.compile()` to `Model.compile_fn()`
- ✅ [rv](cci:1://file:///c:/Users/utkar/Desktop/PYMC/pymc/pymc/model/core.py:1194:4-1269:21) passed as graph input (compatible with `model.copy()`)
- ✅ Added [create_transform_aware_check_rewrite()](cci:1://file:///c:/Users/utkar/Desktop/PYMC/pymc/pymc/logprob/utils.py:320:0-367:34) using `rvs_to_transforms`

 Distribution [logp](cci:1://file:///c:/Users/utkar/Desktop/PYMC/pymc/pymc/model/core.py:658:4-737:26) methods are staticmethods without RV access. What's the preferred approach to pass RV context?

Thanks!

Utkarshkarki:
@ricardoV94, Can you please review my changes and tell me the next steps? I

ricardoV94:
@Utkarshkarki I don't have capacity to guide in this PR, as it is rather technical. I suggest you try another issue if it's not obvious for you how to proceed alone (it is not obvious for me either until I think a bit more about it). 

An easier starting point is to add rewrites like `exp(x) > 0 -> True` in PyTensor (with a `nan_unsafe` tag, that specifies this rewrite ignores `nan` edge cases). This sort of stuff will cover some of the useless checks in the logp for properly transformed variables, moving us closer towards the `check_bounds` being useless, without any changes in PyMC. We may find this sort of approach covers enough cases that we don't need to do anything in PyMC. Even if it doesn't, it's still widely useful.

Utkarshkarki:
i will look into it

PR #8047, opened 2026-01-10
SMC Multiprocessing and Progress Bar Refactor

### LINKED ISSUE #5609: pickle error in pm.samples for cores > 1 with black box likelihood
I have  a model which depends on a block box likelihood function, which calls into [AMICI](https://github.com/AMICI-dev/AMICI).
This model samples fine when using `cores=1`. But when switching to anything higher I get the following error:

```python

  File "/miniconda3/envs/rjmcmc/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/miniconda3/envs/rjmcmc/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/.vscode-oss/extensions/ms-python.python-2022.2.1924087327/pythonFiles/lib/python/debugpy/__main__.py", line 45, in <module>
    cli.main()
  File "/.vscode-oss/extensions/ms-python.python-2022.2.1924087327/pythonFiles/lib/python/debugpy/../debugpy/server/cli.py", line 444, in main
    run()
  File "/.vscode-oss/extensions/ms-python.python-2022.2.1924087327/pythonFiles/lib/python/debugpy/../debugpy/server/cli.py", line 331, in run_module
    run_module_as_main(target_as_str, alter_argv=True)
  File "/miniconda3/envs/rjmcmc/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/miniconda3/envs/rjmcmc/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/sketches/blasi_model_selection.py", line 264, in <module>
    trace = pm.sample(ndraws, tune = nburn, step = step, discard_tuned_samples = True, cores=2, chains=nchains)
  File "/miniconda3/envs/rjmcmc/lib/python3.9/site-packages/pymc/sampling.py", line 569, in sample
    mtrace = _mp_sample(**sample_args, **parallel_args)
  File "/miniconda3/envs/rjmcmc/lib/python3.9/site-packages/pymc/sampling.py", line 1493, in _mp_sample
    sampler = ps.ParallelSampler(
  File "/miniconda3/envs/rjmcmc/lib/python3.9/site-packages/pymc/parallel_sampling.py", line 411, in __init__
    step_method_pickled = cloudpickle.dumps(step_method, protocol=-1)
  File "/miniconda3/envs/rjmcmc/lib/python3.9/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
    cp.dump(obj)
  File "/miniconda3/envs/rjmcmc/lib/python3.9/site-packages/cloudpickle/cloudpickle_fast.py", line 602, in dump
    return Pickler.dump(self, obj)
TypeError: cannot pickle 'SwigPyObject' object

```

This error actually only started appearing after switching from pymc3 to the new 4.0.0bx. but I need some of the new features so I can't go back.

* PyMC/PyMC3 Version: 4.0.0b3
* Aesara/Theano Version: 2.4.0
* Python Version: 3.9.10
* Operating system: macOS 11.6.4
* How did you install PyMC/PyMC3: pip within a conda environment

### LINKED ISSUE #7078: BUG: as_op not pickled, making parallel SMC crash
### Describe the issue:

As it stands, SMC sampler cannot be parallelized with custom ops.

When using SMC sampler with more than one core (i.e. parallel sampling) and an `as_op` custom op, the op is not pickled properly in the "manual" pickling at https://github.com/pymc-devs/pymc/blob/118be0f23782945dc03c5fb36d58d6ce4a1f619f/pymc/smc/sampling.py#L385 , thus causing the run to fail.

```python
import pymc as pm
import pytensor.tensor as pt

from pytensor.compile.ops import as_op

@as_op(itypes=[pt.dvector], otypes=[pt.dvector])
def twice(x):
    return 2*x

with pm.Model() as model:
    x = pm.Normal('x', mu=[0, 0], sigma=1)
    y = twice(x)
    z = pm.Normal(name='z', mu=y, observed=[1, 1])

    # Using cores=1 would work, but cores=2 throws an error
    trace = pm.sample_smc(10,cores=2)
```

### Error message:

```shell
<details>
{
	"name": "AttributeError",
	"message": "module '__main__' has no attribute 'twice'",
	"stack": "---------------------------------------------------------------------------
RemoteTraceback                           Traceback (most recent call last)
RemoteTraceback: 
\"\"\"
  File \"/opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/multiprocessing/pool.py\", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File \"/opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/multiprocessing/pool.py\", line 51, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
  File \"/opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/site-packages/pymc/smc/sampling.py\", line 419, in _apply_args_and_kwargs
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File \"/opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/site-packages/pymc/smc/sampling.py\", line 320, in _sample_smc_int
    (draws, kernel, start, model) = map(
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File \"/opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/site-packages/pytensor/compile/ops.py\", line 221, in load_back
    obj = getattr(module, name)
          ^^^^^^^^^^^^^^^^^^^^^
AttributeError: module '__main__' has no attribute 'twice'
\"\"\"

The above exception was the direct cause of the following exception:

Cell In[14], line 2
      1 with model:
----> 2     trace = pm.sample_smc(10,cores=2)

File /opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/site-packages/pymc/smc/sampling.py:213, in sample_smc(draws, kernel, start, model, random_seed, chains, cores, compute_convergence_checks, return_inferencedata, idata_kwargs, progressbar, **kernel_kwargs)
    210 t1 = time.time()
    212 if cores > 1:
--> 213     results = run_chains_parallel(
    214         chains, progressbar, _sample_smc_int, params, random_seed, kernel_kwargs, cores
    215     )
    216 else:
    217     results = run_chains_sequential(
    218         chains, progressbar, _sample_smc_int, params, random_seed, kernel_kwargs
    219     )

File /opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/site-packages/pymc/smc/sampling.py:388, in run_chains_parallel(chains, progressbar, to_run, params, random_seed, kernel_kwargs, cores)
    386 params = tuple(cloudpickle.dumps(p) for p in params)
    387 kernel_kwargs = {key: cloudpickle.dumps(value) for key, value in kernel_kwargs.items()}
--> 388 results = _starmap_with_kwargs(
    389     pool,
    390     to_run,
    391     [(*params, random_seed[chain], chain, pbars[chain]) for chain in range(chains)],
    392     repeat(kernel_kwargs),
    393 )
    394 results = tuple(cloudpickle.loads(r) for r in results)
    395 pool.close()

File /opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/site-packages/pymc/smc/sampling.py:415, in _starmap_with_kwargs(pool, fn, args_iter, kwargs_iter)
    411 def _starmap_with_kwargs(pool, fn, args_iter, kwargs_iter):
    412     # Helper function to allow kwargs with Pool.starmap
    413     # Copied from https://stackoverflow.com/a/53173433/13311693
    414     args_for_starmap = zip(repeat(fn), args_iter, kwargs_iter)
--> 415     return pool.starmap(_apply_args_and_kwargs, args_for_starmap)

File /opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/multiprocessing/pool.py:375, in Pool.starmap(self, func, iterable, chunksize)
    369 def starmap(self, func, iterable, chunksize=None):
    370     '''
    371     Like `map()` method but the elements of the `iterable` are expected to
    372     be iterables as well and will be unpacked as arguments. Hence
    373     `func` and (a, b) becomes func(a, b).
    374     '''
--> 375     return self._map_async(func, iterable, starmapstar, chunksize).get()

File /opt/homebrew/Caskroom/mambaforge/base/envs/pymc_env/lib/python3.11/multiprocessing/pool.py:774, in ApplyResult.get(self, timeout)
    772     return self._value
    773 else:
--> 774     raise self._value

AttributeError: module '__main__' has no attribute 'twice'"
}
</details>
```

### PyMC version information:

pymc: 5.10.3
pytensor: 2.18.4
python: 3.11.7

Installed in a fresh conda environment with 
`conda create -c conda-forge -n pymc_env "pymc>=5"`

### Context for the issue:

As it stands, SMC sampler cannot run the official PyMC example from https://www.pymc.io/projects/examples/en/latest/ode_models/ODE_Lotka_Volterra_multiple_ways.html 
Any simple ODE where sunode is overkill will crash similarly, as it requires a custom op, that is not pickled.

The workaround of using a single core makes the method much slower than needed.

Is there a way to serialize the custom operation please?

### LINKED ISSUE #7413: ENH: Compile the functions needed by SMC before the worker processes are started
### Before

_No response_

### After

_No response_

### Context for the issue:

As explained in #7224, `sample_smc` can fail on Windows if the model is defined using `CustomDist`. Everything works on Linux because the worker processes are _forked_ by default whereas they're _spawned_ on Windows. Spawning leads to missing dispatches.

#7241 addressed the issue by manually registering the dispatches in the worker processes. A better way to solve the problem would be to compile the necessary functions before starting the worker processes (as is done in `sample`).

### LINKED ISSUE #8022: BUG: `sample_smc` can lead to compilation halting
### Describe the issue:

I upgraded to `pymc 5.27.0` and noticed that my models do not sample with `sample_smc` nearly as efficiently. I just confirmed this on a fresh install and running the model from https://www.pymc.io/projects/examples/en/latest/samplers/SMC2_gaussians.html

Using `pymc=5.23.0`, each chain starts up right away and completes sampling within 10 seconds. The output is:
```
Chain 0  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 1  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 2  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 3  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 4  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 5  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 6  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 7  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 8  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 9  ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 10 ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
Chain 11 ⠸ -:--:-- / 0:00:09 Stage: 0 Beta: 1.000
```

Using `pymc=5.27.0`, chains take almost 1 minute to start, and they do not all start at the same time, and they take longer to finish. The output is
```
Chain 0  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 1  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 2  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 3  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 4  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 5  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 6  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 7  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 8  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 9  ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 10 ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
Chain 11 ⠦ -:--:-- / 0:01:39 Stage: 0 Beta: 1.000
```

```python
https://www.pymc.io/projects/examples/en/latest/samplers/SMC2_gaussians.html
```

### Error message:

```shell

```

### PyMC version information:

5.27.0

### Context for the issue:

There is a 10x runtime increase in `sample_smc` between `5.23.0` and `5.27.0`.

### LINKED ISSUE #8030: BUG: `sample_smc` stalls on final stage
### Describe the issue:

Sometimes, randomly as far as I can tell, `sample_smc` hangs when it reaches the final stage. This is a hard one to reproduce -- sometimes I can sample 100s of models before one hangs.

This has been discussed here: https://discourse.pymc.io/t/sample-smc-stalls-at-final-stage

@ricardoV94 suggested downgrading to `5.12.0` (before the new progress bars). This solves the issue -- I do not run into the hanging after 1000s of models sampled.

```python
import pymc as pm
...
pm.sample_smc()
```

### Error message:

```shell

```

### PyMC version information:

5.27.0

### Context for the issue:

_No response_

jessegrabowski:
## Description
I always like SMC as a gradient-free option for my big silly models with few parameters, but it always gave me trouble because of the API break between it and `pm.sample`. This PR aims to harmonize the two by bringing over a bunch of functionality from `pm.sample` to `pm.sample_smc`. 

This PR is intended to be reviewed commit by commit. I verified that the test suite runs in all intermediate forms. Here is a summary of each commit:

- `Use multiprocessing for SMC sampling`: The `multiprocessing` library is now used to handle parallel SMC sampling. This commit was heavily Claude-assisted, so it should receive special scrutiny. The objective was to make SMC multiprocessing look exactly like MCMC multiprocessing. It also exposes an `mp_ctx` argument to `pm.sample_smc`, which can allow compiling with e.g. JAX (using `mp_ctx ='forkserver'`).
-  `Sample SMC sequentially when cores=1` adds separate logic for sequential sampling on one core. Again, this copies the relevant MCMC functions.
- `Initialize SMC Kernels on main process` is a major performance change, intended to address e.g.  https://github.com/pymc-devs/pymc/issues/8030. Pytensor compilation is not thread-safe, so we shouldn't be doing it on the workers. In this PR, the kernel is compiled once on the main process, then serialized and sent to the workers. This matches what we do with step functions in MCMC. Importantly, this commit eliminates the need for serialization of many auxiliary objects, including the pymc model itself, and some special logic for custom distributions. To do this, a couple of ancillary changes had to be made -- for example, transformation of the chains from numpy to NDArray objects happens on the main process now, after all sampling is done.
- `Add blas_cores argument to sample_smc` again, this copies over multiprocessing machinery from `pm.sample` to `pm.sample_smc` by adding a `blas_cores` argument to `pm.sample_smc`, for the same reasons it exists over there.
- `Add custom progress bar for SMC` adds a progress bar style to `sample_smc` that matches that of `pm.sample`. The bars fill from 0-1 following the value of `beta`, and we provide an estimated time to completion by measuring the speed per step. It looks like this:

```
 Progress                                   Stage   Beta     Stage Speed    Elapsed   Remaining 
────────────────────────────────────────────────────────────────────────────────────────────────
 ━━╺━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   3       0.0620   4.66 s/stage   0:00:17   0:03:46   
 ━━╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   3       0.0634   5.42 s/stage   0:00:17   0:04:18   
 ━╺━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   2       0.0269   6.15 s/stage   0:00:17   0:09:29   
 ━━╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   3       0.0646   5.38 s/stage   0:00:17   0:04:10   
 ━━╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   3       0.0658   5.16 s/stage   0:00:17   0:03:55   
 ━━╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   3       0.0658   5.55 s/stage   0:00:17   0:04:12   
 ━━╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   3       0.0639   4.83 s/stage   0:00:17   0:03:46   
```

I observed big speed gains using `sample_smc` after this PR. I timed this simple hierarchical model:

```py
import pymc as pm

with pm.Model() as m:
    idxs = pm.draw(pm.Categorical.dist(p=[1] * 10, shape=(100,)))
    effect_loc = pm.Normal('mu_loc', 0, 1)
    effect_scale = pm.HalfNormal('mu_scale', 1)
    effect = pm.Normal('mu', mu=effect_loc, sigma=effect_scale, shape=(10,))
    
    X = pm.draw(pm.Normal.dist(0, 1, shape=(100, 5)))
    beta = pm.Normal('beta', 0, 1, shape=(5,))
    
    mu = effect[idxs] + X @ beta
    sigma = pm.Exponential('sigma', 1)
    
    obs = pm.Normal('obs', mu=mu, sigma=sigma)

draw = prior.prior.obs.sel(chain=0, draw=123).values
m2 = pm.observe(m, {obs:draw})

%%time 
with m2:
    idata = pm.sample_smc()

```

Timings went from 6.1 s to 1.44 s using the C backend, and 1.46 s to 1.09 s using Numba mode (with cache). Running `test_smc.py` locally goes from 1m4s  to 6.264 seconds.

I could run more formal benchmarks if someone asks, but I don't really want to.

## Related Issue

- [ ] Closes #8030
- [ ] Closes #8022
- [ ] Closes #7413 
- [ ] Closes #7078 
- [ ] Closes #5609

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):


jessegrabowski:
@tvwenger it would be nice if you could try your SMC models that have been giving you trouble on this PR branch and report back, since you've been the one doing the heavy lifting bug-hunting on SMC lately.

pymc/smc/parallel.py
```diff
@@ -0,0 +1,431 @@
...
+    def join(self, timeout=None):
+        """Join the worker process."""
+        self._process.join(timeout)
+
+    def terminate(self):
+        """Terminate the worker process."""
+        self._process.terminate()
+
+    @staticmethod
+    def recv_message(processes, timeout=3600):
```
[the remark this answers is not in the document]
jessegrabowski: Amusing robot misfire.

pymc/smc/parallel.py
```diff
@@ -0,0 +1,431 @@
...
+            self._start = cloudpickle.loads(self._start)
+
+    def run(self):
+        with threadpool_limits(limits=self._blas_cores):
+            try:
+                self._unpickle_objects()
+                self._run_smc()
+            except KeyboardInterrupt:
+                pass
+            except BaseException as e:
```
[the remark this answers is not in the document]
jessegrabowski: Maybe this should change but it's following the pattern set in `sampling/parallel.py`, so I don't want to change it.

If I do change it, it will be to make the multiprocessing code more method agnostic so we can share all this setup/teardown code between MCMC, SMC, and others (minibatch ADVI, for example). But I don't want to do that in this PR.

pymc/smc/sampling.py
```diff
@@ -175,29 +187,132 @@ def sample_smc(
...
+                # We need to convert it to dict of lists (one list per stat)
+                stats_dict: dict[str, list] = {}
+                for stage_stats in final_result.sample_stats:
+                    for stat, value in stage_stats.items():
+                        if stat not in stats_dict:
+                            stats_dict[stat] = []
+                        stats_dict[stat].append(value)
+
+                sample_stats.append(stats_dict)
+                sample_settings.append(final_result.sample_settings)
```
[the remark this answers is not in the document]
jessegrabowski: I adjusted this to defensively fail if a chain is empty. I have no idea how this could happen live but now it's covered I guess.

ricardoV94:
Check https://github.com/pymc-devs/pymc/pull/8044

pymc/progress_bar.py
```diff
@@ -423,3 +423,174 @@ def create_progress_bar(self, step_columns, progressbar, progressbar_theme):
             disable=not progressbar,
             include_headers=True,
         )
+
+
+class SMCProgressBarManager(ProgressBarManager):
```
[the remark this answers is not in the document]
jessegrabowski: Not needed.

jessegrabowski:
> Check https://github.com/pymc-devs/pymc/pull/8044

I addressed this in the [Extract and share Parallel setup code between MCMC and SMC](https://github.com/pymc-devs/pymc/pull/8047/commits/8f6f2adc7f5e813edc119b4d18cf32f7475fff03) commit, but now the two PRs are overlapping.

pymc/smc/parallel.py
```diff
@@ -0,0 +1,431 @@
...
+        self._blas_cores = blas_cores
+
+    def _unpickle_objects(self):
+        """Unpickle kernel and start point."""
+        self._kernel = cloudpickle.loads(self._kernel)
+        if self._start is not None:
+            self._start = cloudpickle.loads(self._start)
+
+    def run(self):
+        with threadpool_limits(limits=self._blas_cores):
```
ricardoV94: Using threadpool_limits regardless of context?

ricardoV94:
So there's an edge case with the pickle function -> send to process approach. If the pickled functions have random number generators these need to de changed so as to have independent streams.

Usually this isn't a problem in mcmc because we never wanted to use functions with randomness in it in our step samplers, but this is not the case for SMC, and specially SMC-ABC with Simulator, which definitely supposed to be random.

When you call `model.logp()` in a model with a Simulator the logp as RandomGenerator variables in it. These are unique everytime you called `model.logp()`, so the compile in each process didn't have the duplication issue (entropy properties may not have been great, but that's a different matter). Here they'll be exactly the same.

The approach may require something like the `set_rng` code added for the conjugate samplers in this PR: https://github.com/ricardoV94/pymc-extras/blob/35530daf55a8ae7f5cbcf5ae00760da32abf560a/pymc_extras/sampling/optimizations/conjugate_sampler.py#L84-L98, and make sure it's called at the beginning of sampling in each thread

pymc/smc/sampling.py
```diff
@@ -175,29 +186,115 @@ def sample_smc(
...
 
-    results = run_chains(chains, progressbar, params, random_seed, kernel_kwargs, cores)
+    rngs = [np.random.default_rng(seed) for seed in random_seed]
 
-    (
-        traces,
-        sample_stats,
-        sample_settings,
-    ) = zip(*results)
+    with model:
```
ricardoV94, line 200: why `with_model`?

pymc/smc/sampling.py
```diff
@@ -279,157 +375,194 @@ def _save_sample_stats(
...
+            if var.dtype in discrete_types:
+                var_samples = np.round(var_samples).astype(var.dtype)
+            value.append(var_samples.reshape(shape))
+            size += new_size
+        strace.record(point=dict(zip(varnames, value)))
+
+    return strace
+
+
+def _sample_smc_many(
```
ricardoV94: The function calling seems as complex as the inner code, so inline it in `sample_smc`?

pymc/smc/parallel.py
```diff
@@ -0,0 +1,431 @@
...
+        while smc.beta < 1:
+            smc.update_beta_and_weights()
+
+            self._msg_pipe.send(("progress", stage, smc.beta))
+
+            msg = self._recv_msg()
+            if msg[0] == "abort":
+                raise KeyboardInterrupt()
+            elif msg[0] != "continue":
+                raise ValueError("Unknown message " + msg[0])
```
ricardoV94: Why is this done in between two stages of the sampler?

pymc/smc/parallel.py
```diff
@@ -0,0 +1,431 @@
...
+
+            smc.resample()
+            smc.tune()
+            smc.mutate()
```
ricardoV94: Can we avoid duplication of the steps between sequential and parallel sampling? Perhaps we could reuse/abstract the mcmc parallel/sequential logic and implement the SMC kernel like one of the step samplers that that logic uses?

pymc/smc/parallel.py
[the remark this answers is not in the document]
ricardoV94: Open issue?

pymc/smc/parallel.py
jessegrabowski: Yeah I hated this too.

jessegrabowski:
> The approach may require something like the set_rng code added for the conjugate samplers in this PR: https://github.com/ricardoV94/pymc-extras/blob/35530daf55a8ae7f5cbcf5ae00760da32abf560a/pymc_extras/sampling/optimizations/conjugate_sampler.py#L84-L98, and make sure it's called at the beginning of sampling in each thread

Could we just make the rng an explicit input to the function we pickle up and send out, to avoid the copy?

ricardoV94:
> Could we just make the rng an explicit input to the function we pickle up and send out, to avoid the copy?

Not without much more changes in the codebase

jessegrabowski:
OK, but I only need to copy once per worker right? After the RNG is split and set we're good?

ricardoV94:
> OK, but I only need to copy once per worker right? After the RNG is split and set we're good?

I think so

pymc/progress_bar.py
```diff
@@ -332,13 +332,13 @@ def _initialize_tasks(self):
             ]
 
     @staticmethod
-    def compute_draw_speed(elapsed, draws):
-        speed = draws / max(elapsed, 1e-6)
+    def compute_draw_speed(elapsed, actions, action_name="draw", time_unit="s"):
```
zaxtax: This change is more confusing. Should be reverted

pymc/progress_bar.py
```diff
@@ -423,3 +423,174 @@ def create_progress_bar(self, step_columns, progressbar, progressbar_theme):
...
+    def _initialize_tasks(self):
+        """Initialize progress bar tasks for each chain."""
+        self.tasks = [
+            self._progress.add_task(
+                f"Chain {chain_idx}",
+                completed=0.0,
+                total=self.total_beta,
+                chain_idx=chain_idx,
+                sampling_speed=0,
+                speed_unit="stages/s",
```
zaxtax: stages/s seems like a weird unit. What's wrong with draws/s?

pymc/sampling/mcmc.py
```diff
@@ -769,27 +770,10 @@ def sample(
...
-
-        def joined_blas_limiter():
-            return threadpool_limits(limits=blas_cores)
-
-        num_blas_cores_per_chain = blas_cores // cores
-    else:
-        raise ValueError(
-            f"Invalid argument `blas_cores`, must be int, 'auto' or None: {blas_cores}"
-        )
+    mp_ctx = _initialize_multiprocessing_context(mp_ctx, quiet=quiet)
```
zaxtax, line 776: This needs a unit test

pymc/smc/kernels.py
```diff
@@ -336,17 +336,60 @@ def sample_settings(self) -> SMCSettings:
...
-    def _posterior_to_trace(self, chain=0) -> NDArray:
+    @staticmethod
+    def _progressbar_config(n_chains=1):
+        """Configure progress bar columns for SMC sampling.
+
+        Returns columns to display and initial stats values.
+        """
+        columns = [
+            TextColumn("{task.fields[stage]}", table_column=Column("Stage", ratio=1)),
+            TextColumn("{task.fields[beta]:.4f}", table_column=Column("Beta", ratio=1)),
```
zaxtax, line 423: What do Stage and Beta mean? How do they relate to SMC?

pymc/smc/kernels.py
```diff
@@ -336,17 +336,60 @@ def sample_settings(self) -> SMCSettings:
...
+        return columns, stats
+
+    @staticmethod
+    def _make_progressbar_update_functions():
+        """Create functions to update progress bar statistics."""
+
+        def update_stats(stats):
+            return {
+                "stage": stats.get("stage", 0),
+                "beta": stats.get("beta", 0.0),
```
zaxtax, line 438: What do stage and beta mean?

pymc/smc/sampling.py
```diff
@@ -241,7 +338,6 @@ def _save_sample_stats(
     sample_stats_dict = sample_stats[0]
 
     if chains > 1:
-        # Collect the stat values from each chain in a single list
```
zaxtax, line 244: I'd keep the comment

zaxtax (commented):
Overall looks good. Some of the code could still be simplified

tvwenger:
@jessegrabowski ping me again when this is stable and I'll test it with my models and workflow.

jessegrabowski:
> @jessegrabowski ping me again when this is stable and I'll test it with my models and workflow.

It should work as-is, what I'll do from here is mostly refactors and cleanup. I'm mostly curious if you have a chonky model you're fitting using SMC currently, if you can use this branch and do `pm.sample_smc(compile_kwargs={'mode':'NUMBA'})` to get a free speedup (and verify that everything works)

pymc/progress_bar.py
jessegrabowski: Draws aren't the unit of iteration in SMC

pymc/smc/kernels.py
jessegrabowski: Beta is the temperature parameter, controlling the annealing mixture. Each time the algorithm decides to increase the temperature, it enters a new stage.

pymc/smc/parallel.py
jessegrabowski: This is the same behavior as `sampling/parallel.py` [here](https://github.com/pymc-devs/pymc/blob/689f73632a535868bed119db3745606770d9e937/pymc/sampling/parallel.py#L146)

pymc/smc/sampling.py
jessegrabowski: NDArray looks for it. I can pass it as a kwarg instead, maybe that's more clear.

pymc/smc/parallel.py
jessegrabowski: I followed the example [here](https://github.com/pymc-devs/pymc/blob/689f73632a535868bed119db3745606770d9e937/pymc/sampling/parallel.py#L218).

Stages are the looping concept, so I don't know when else we could check for keyboard interrupt.

pymc/progress_bar.py
jessegrabowski: I'd rather have a flexible method than two identical ones with a different string. What is confusing about it?

pymc/smc/sampling.py
jessegrabowski: I added the stat_dict_dtype metadata to each kernel, so all of this logic about converting from lists of dicts to dicts of lists is now gone.

pymc/smc/parallel.py
jessegrabowski: I paired it down a bit. There's still some code duplication but at this point the alternative would be to move more stuff into the kernels and I think they're already quite heavy.

pymc/progress_bar.py
zaxtax: I thought it was only used once in the code.

jessegrabowski: It's shared by mcmc (which loops over draws) and smc (which loops over stages)

In the near future i'll redo the ADVI progress bars, which loop over steps or epochs. So it's handy to be able to re-use.

pymc/sampling/mcmc.py
jessegrabowski: It was existing code, i just put it into a little wrapper

pymc/smc/parallel.py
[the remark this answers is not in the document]
jessegrabowski: I've been thinking about this, and I think BaseException is correct here. Basically we're at the boundary between worker and main process, and we never want to have the worker die unexpectedly. We're also re-raising.

pymc/smc/parallel.py
```diff
@@ -0,0 +1,430 @@
...
+        )
+        self._msg_pipe.send(("done", result))
+
+
+def _run_smc_process(*args):
+    """Entry point for SMC process."""
+    _SMCProcess(*args).run()
+
+
+class SMCProcessAdapter:
```
zaxtax: Multiprocess code is tricky. We should have unit test coverage for this.

pymc/sampling/mcmc.py
```diff
@@ -1025,6 +1009,45 @@ def joined_blas_limiter():
...
+    elif isinstance(blas_cores, int):
+        if mp_ctx.get_start_method() == "fork":
+            # https://github.com/pymc-devs/pymc/issues/7354
+            joined_blas_limiter = contextlib.nullcontext
+        else:
+
+            def joined_blas_limiter():
+                return threadpool_limits(limits=blas_cores)
+
+        num_blas_cores_per_chain = blas_cores // cores
```
zaxtax: Should cores be chains here?

jessegrabowski: Maybe, this was pre-existing code. I don't want to introduce changes to MCMC in this PR

pymc/smc/parallel.py
jessegrabowski: What do you suggest for MP specifically? All existing SMC tests pass through this code.

pymc/progress_bar/utils.py
```diff
@@ -14,14 +14,16 @@
 """Shared utilities for progress bar rendering."""
 
 
-def compute_draw_speed(elapsed: float, draws: int) -> tuple[float, str]:
+def compute_draw_speed(
```
ricardoV94: inline for the smc progressbar and revert the changes

or inline for everyone... hate this helper

pymc/sampling/mcmc.py
ricardoV94: cores here is already <= to chains, I think this is right

ricardoV94: correct name would be `num_blas_cores_per_worker` or smth (chains will wait while pool is exhausted)

pymc/smc/kernels.py
```diff
@@ -127,6 +129,12 @@ class SMC_KERNEL(ABC):
 
     """
 
+    stats_dtypes_shapes: dict[str, tuple[type, list]] = {
+        "log_marginal_likelihood": (float, []),
+        "beta": (float, []),
+    }
+    """Maps stat names to (dtype, shape) tuples."""
```
ricardoV94:
```suggestion
```

pymc/smc/kernels.py
```diff
@@ -169,86 +177,115 @@ def __init__(
...
-                },
-                default_strategy="prior",
-                return_transformed=True,
-            )
-            prior_values = draw(prior_expression, draws=self.draws, random_seed=self.rng)
-
-            names = [model.rvs_to_values[rv].name for rv in model.free_RVs]
-            dict_prior = {k: np.stack(v) for k, v in zip(names, prior_values)}
+            prior_values = draw(self._prior_expression, draws=self.draws, random_seed=self.rng)
+            dict_prior = {k: np.stack(v) for k, v in zip(self._prior_var_names, prior_values)}
```
ricardoV94: I think this `np.stack` is a (costly) no-op, and it will fail for scalars. Try to remove?

pymc/smc/parallel.py
```diff
@@ -0,0 +1,430 @@
...
+            self._start = cloudpickle.loads(self._start)
+
+    def run(self):
+        with (
+            nullcontext()
+            if self._mp_start_method == "fork"
+            else threadpool_limits(limits=self._blas_cores)
+        ):
+            try:
+                self._unpickle_objects()
```
ricardoV94: I think _unpickle_objcts need not be inside this context

pymc/smc/kernels.py
jessegrabowski: nice catch

pymc/smc/parallel.py
jessegrabowski: Done, but im curious why you flagged this, just because it's not going to be the reason for failure? Or is there a performance reason

jessegrabowski:
There were some major conflicts with the recent round of progress bar PRs, so I went ahead and did a bit of refactoring:

- ProgressBarManager is now an ABC that holds common progress bar logic
- MCMCProgressManager handles MCMC 
- SMCProgressManager handles SMC

I renamed everything in the rich/marimo progress bar backends to have more agnostic argument names (chains -> n_bars, total_draws -> total, draw -> completed, etc). 

The marimo backend won't work with SMC because it has a ton of hard-coded logic for MCMC and I don't care enough to handle it. If a future dev cares both about SMC and Marimo, he can handle it in the future.

pymc/smc/parallel.py
ricardoV94: It's common pattern to only try/except the code you actually expect to fail. It's easier to read the code, and less likely you accidentally catch errors that you were not supposed to catch (that is hide bugs)

ricardoV94:
@jessegrabowski is this ready for final review?

jessegrabowski:
yep

pymc/progress_bar/marimo_progress.py
```diff
@@ -131,58 +132,63 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
...
                 {
-                    "draws": 0,
-                    "total": self.total_draws * self.chains,
+                    "completed": 0,
+                    "total": self.total * self.n_bars,
                     "failing": False,
                     "stats": {},
                 }
             ]
             self._start_times = [perf_counter()]
```
ricardoV94, line 145: Perhaps initialize only when the task / bar advances for the first time. So that sequential sampling is not measuring speed relative to the start of the first chain (otherwise it seems each chain is slower than the previous one...). Does not need to be done in this PR

pymc/smc/kernels.py
```diff
@@ -169,86 +176,115 @@ def __init__(
...
+        so we replace the shared RNGs in the compiled functions with new ones created from the provided `rng`.
+
+        This method copies the functions, so it is expensive and should only be called once per worker!
+        """
+
+        def make_rng_swaps(fn, rng):
+            shared_rngs = [
+                var for var in fn.get_shared() if isinstance(var.type, RandomGeneratorType)
+            ]
+            n_shared_rngs = len(shared_rngs)
```
ricardoV94, line 244: raise NotImplementedError if there are `shared_rngs` and isinstance(fn.maker.linker, JAXLinker)`, as the rngs have different format after compile and no longer are retrieved by `fn.get_shared()` either.

pymc/smc/kernels.py
```diff
@@ -169,86 +176,115 @@ def __init__(
...
+                    shared_rngs, rng.spawn(n_shared_rngs), strict=True
+                )
+            }
+
+        self.prior_logp_func = self.prior_logp_func.copy(
+            swap=make_rng_swaps(self.prior_logp_func, rng)
+        )
+        self.likelihood_logp_func = self.likelihood_logp_func.copy(
+            swap=make_rng_swaps(self.likelihood_logp_func, rng)
+        )
```
ricardoV94, line 265: For safety, even if not used, do `self.rng = rng`

pymc/smc/sampling.py
```diff
@@ -279,157 +366,178 @@ def _save_sample_stats(
...
+                model=model,
+                traces=traces,
+                sample_stats=sample_stats,
+                sample_settings=sample_settings,
+                progress_manager=progress_manager,
             )
 
-    return custom_methods
 
+def _sample_smc(
```
ricardoV94: inline this in `_sample_smc_many`?

tests/smc/test_smc.py
```diff
@@ -149,6 +150,27 @@ def _dist(mu, size=None):
             pm.CustomDist("y", mu, logp=_logp, class_name="", random=_random, observed=[1, 2])
             pm.sample_smc(draws=6, cores=2)
 
+    @pytest.mark.parametrize("chains", [1, 2], ids=["1_chain", "2_chains"])
+    def test_sequential(self, chains, caplog):
```
ricardoV94: parametrize the first sample_test to conver both sequential and parallel.

tests/smc/test_smc.py
```diff
@@ -149,6 +150,27 @@ def _dist(mu, size=None):
...
+        """Test sequential SMC sampling by setting cores=1"""
+        with self.fast_model:
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", ".*number of samples.*", UserWarning)
+                trace = pm.sample_smc(
+                    draws=50,
+                    chains=chains,
+                    cores=1,
+                    return_inferencedata=False,
+                    progressbar=not _IS_WINDOWS,
```
ricardoV94: We shouldn't need this `not _IS_WINDOWS` logic anywhere anymore with the new progressbar (it existed in previous tests)

ricardoV94 (commented):
My last requests (97% confidence)

pymc/progress_bar/marimo_progress.py
jessegrabowski: I looked into this a little bit and couldn't find a trivial solution in rich. So I'll open an issue and leave it for later.

jessegrabowski:
@ricardoV94 should be good to go. I collected everything into 3 logical commits.

ricardoV94 (approved):
This change is too large and technical to be confident we didn't break anything from code review and CI passing alone. So let's keep an eye after we merge...

The idea is great!

PR #8056, opened 2026-01-15
Add rewrites with RV reparametrization tricks

lucianopaz:
This was born from pymc-devs/pytensor#1424. I started to write down the reparametrization tricks I could find or figure out for most of pytensor's basic `RandomVariable` `Op`s

### DIFF
--- a/pymc/distributions/rewrites/reparametrization.py
+++ b/pymc/distributions/rewrites/reparametrization.py
@@ -132,7 +132,11 @@ def inv_gamma_reparametrization(fgraph, node):
 
 def gamma_reparametrization_impl(rng, size, shape, scale):
     # We'll implement the Marsaglia-Tsang boosted algorithm to sample
-    # We follow the implementation used in jax
+    # https://dl.acm.org/doi/epdf/10.1145/358407.358414
+    # We follow the algorithm from section 3 without squeezing.
+    # The squeeze algorithm from section 4 is more efficient if we can avoid
+    # computing all branches of the if-else clauses, which is not possible
+    # when using pytensor switches.
     # For context, shape is equal to alpha in all of the following math.
     # Sampling for alpha >= 1 is done with a rejection algorithm that finishes in constant time.
     # For alpha < 1, we need to boost the samples using:
@@ -141,9 +145,6 @@ def gamma_reparametrization_impl(rng, size, shape, scale):
     # log(Gamma(alpha, 1)) = log(Gamma(alpha + 1, 1)) + log(Uniform(0, 1)) / alpha
     # We can note that log(Uniform(0, 1)) = -Exponential(1)
     # and we have to guard agains the case where the exponential sample is equal to 0.
-    #
-    # Jax's implementation of the rejection algorithm does not match with what's
-    # written in wikipedia.
     assert size == (), (
         "Gamma reparametrization requires that you first apply the local_rv_size_lift "
         "rewrite in order to have size equal to an empty tuple."
@@ -164,40 +165,21 @@ def gamma_reparametrization_impl(rng, size, shape, scale):
             ones_like(alpha),
             rng=rng,
         ).owner.outputs
-
-        def inner_step(x, v, chosen, rng, c):
-            next_rng, X = NormalRV()(
-                zeros_like(c),
-                ones_like(c),
-                rng=rng,
-            ).owner.outputs
-            v = 1 + c * X
-            indicators = and_(pt.ge(v, 0), ~chosen)
-            chosen = switch(indicators, ones_like(chosen, dtype="bool"), chosen)
-            return (X, v, chosen, next_rng), until(pt.all(chosen))
-
-        xs, vs, _, next_rng = scan(
-            inner_step,
-            outputs_info=[
-                zeros_like(alpha),  # X
-                -ones_like(alpha),  # V
-                zeros_like(alpha, dtype="bool"),  # Chosen
-                next_rng,
-            ],
-            non_sequences=[c],
-            n_steps=10_000,
-            return_updates=False,
-        )
-
-        x = xs[-1]
-        v = vs[-1]
+        next_rng, x = NormalRV()(
+            zeros_like(c),
+            ones_like(c),
+            rng=next_rng,
+        ).owner.outputs
+        V = (1 + c * x) ** 3
 
         X = x * x
-        V = v**3
 
         indicators = and_(
             ~chosen,
-            ~and_(pt.gt(U, 1 - 0.0331 * X * X), pt.gt(log(U), X / 2 + d * (1 - V + log(V)))),
+            and_(
+                V > 0,
+                pt.lt(log(U), X / 2 + d * (1 - V + log(V))),
+            ),
         )
         chosen = switch(indicators, ones_like(chosen, dtype="bool"), chosen)
         output = switch(indicators, V, output)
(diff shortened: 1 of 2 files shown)


pymc/distributions/rewrites/reparametrization.py
```diff
@@ -0,0 +1,206 @@
...
+        GumbelRV,
+        HalfCauchyRV,
+        HalfNormalRV,
+        LaplaceRV,
+        LogisticRV,
+        NormalRV,
+    ]
+)
+def loc_scale_reparametrization(fgraph, node):
+    rng, size, loc, scale = node.inputs
```
jessegrabowski, line 130: The `Half` distributions won't have a loc input

jessegrabowski: nevermind it does. Why don't we expose that in pymc? It's useful.

pymc/distributions/rewrites/reparametrization.py
```diff
@@ -0,0 +1,206 @@
...
+def scale_reparametrization(fgraph, node):
+    rng, size, scale = node.inputs
+    return scale * node.op(1.0, rng=rng, size=size)
+
+
+@register_random_reparametrization
+@node_rewriter([GammaRV, InvGammaRV])
+def gamma_reparametrization(fgraph, node):
+    rng, size, shape, scale = node.inputs
+    return scale * node.op(shape, 1.0, rng=rng, size=size)
```
jessegrabowski: We can't pull the shape parameter out of the RV?

jessegrabowski: The actual reparameterization trick in this case requires some effort ([Marsaglia-Tsang method](https://en.wikipedia.org/wiki/Gamma_distribution#Random_variate_generation)), jax implementation here:

https://github.com/jax-ml/jax/blob/c6568036b83b39d556c68d647a99196e63062612/jax/_src/random.py#L1298

pymc/distributions/rewrites/reparametrization.py
```diff
@@ -0,0 +1,206 @@
...
+        name = kwargs.pop("name", None) or node_rewriter.__name__
+        reparametrization_trick_db.register(name, node_rewriter, *tags, **kwargs)
+        return node_rewriter
+
+
+@register_random_reparametrization
+@node_rewriter([BernoulliRV])
+def bernoulli_reparametrization(fgraph, node):
+    rng, size, p = node.inputs
+    return switch(
```
ricardoV94, line 81: Commenting here but in general you may be making non equivalent graphs, by using the size argument. Remember it may be None, meaning it will be implied by the parameters shape. In that case you may be rewriting something like `normal([0, 1, 2], size=None)` as `[0, 1, 2] + normal(size=None)`

There's a non default PyTensor rewrite that makes size explicit that you may want to use prior to these rewrites. Then make sure these rewrites fail if size is None `isinstance(size.type, NoneTypeT)`. If you do have size it will be all you need

pymc/distributions/rewrites/reparametrization.py
ricardoV94: I don't think it's useful, it's just an accident from copying scipy api that always likes to have loc parameter. We don't expose it to users from pm.HalfNormal and we don't have a default transform that considers it either.

Weird shifts can be attained with customdist

jessegrabowski: Yeah I guess you can also just use a deterministic (1 + x) instead of x ~ HN(loc=1) and it's a lot more clear.

pymc/distributions/rewrites/reparametrization.py
lucianopaz: I'll write the implementation for that then. Thanks @jessegrabowski

pymc/distributions/rewrites/reparametrization.py
lucianopaz: Hmm. I'll have to ask you how to add that rewrite into the database and make it go before any reparametrization rewrites

pymc/distributions/rewrites/reparametrization.py
```diff
@@ -0,0 +1,206 @@
...
+    ParetoRV,
+    StudentTRV,
+    TriangularRV,
+    UniformRV,
+    WaldRV,
+    WeibullRV,
+)
+from pytensor.tensor.slinalg import cholesky
+
+reparametrization_trick_db = RewriteDatabaseQuery(include=["random_reparametrization_trick"])
```
ricardoV94, line 59: This works but it's a bit expensive every time you extend it. You should start with an actual database like SequenceDB, then you can register rewrite phases in it. `pymc/logprob/rewriting.py` may be a good example.

I'm even surprise that the RewriteDatabaseQuery has a register argument. Feels like bad design

pymc/distributions/rewrites/reparametrization.py
lucianopaz: @jessegrabowski, I wrote the generator code. I still need to test that the samples it returns are distributed like the reference Gamma, but maybe you could have a look and comment on the scans

lucianopaz: I've fixed the gamma reparametrization and it now matches the generation we get from `Gamma.dist`. I can't take gradients through the reparametrized version because of pymc-devs/pytensor#555

pymc/distributions/rewrites/reparametrization.py
```diff
@@ -0,0 +1,310 @@
...
+)
+def loc_scale_reparametrization(fgraph, node):
+    rng, size, loc, scale = node.inputs
+    return loc + scale * node.op(zeros_like(loc), ones_like(scale), rng=rng, size=size)
+
+
+@register_random_reparametrization
+@node_rewriter([DirichletRV])
+def dirichlet_reparametrization(fgraph, node):
+    raise NotImplementedError("DirichletRV is not reparametrizable")
```
ricardoV94: I doesn't need to block this PR but I know tfp has something for the dirichlet that is one time differentiable

jessegrabowski: It comes from the Gamma, so we need to sort that out first

ricardoV94: They might have changed and they just define dirichlet with the normalized batch of gamma variates, which would be differentiable if the gamma also are?

jessegrabowski: jax impl: https://github.com/jax-ml/jax/blob/c6568036b83b39d556c68d647a99196e63062612/jax/_src/random.py#L1240

lucianopaz: Thanks, I'll have a look at that

pymc/distributions/rewrites/reparametrization.py
```diff
@@ -0,0 +1,292 @@
...
+        next_rng, U = UniformRV()(
+            zeros_like(alpha),
+            ones_like(alpha),
+            rng=rng,
+        ).owner.outputs
+        next_rng, x = NormalRV()(
+            zeros_like(c),
+            ones_like(c),
+            rng=next_rng,
+        ).owner.outputs
```
jessegrabowski, line 230: It's wasteful, but to avoid the current grad limitations you can sample all of these once outside of the scan then pass them in as sequences

jessegrabowski: Because nothing about U or x actually depends on the state of the scan

jessegrabowski:
@lucianopaz idk if you saw but i did some work on this a few weeks ago, I think this is actually pretty close?

PR #8057, opened 2026-01-15
Fix NaN gradients in Hurdle distributions for continuous components

### LINKED ISSUE #8053: dlogp fails for `HurdleGamma` observation model
### Description

Someone reported a problem in Bambi when they used a `HurdleGamma` observation model (https://github.com/bambinos/bambi/issues/951).
I figured out the dlogp was failing.

Here's an example in PyMC:

```python
import pytensor

import numpy as np
import pymc as pm

n = 50
psi_true = 0.3
alpha_true = 2.0
beta_true = 1.5

dist = pm.HurdleGamma.dist(psi=psi_true, alpha=alpha_true, beta=beta_true)
y = pm.draw(dist, draws=n, random_seed=1)

with pm.Model() as model:
    alpha = pm.HalfNormal("alpha", sigma=2.0)
    beta = pm.HalfNormal("beta", sigma=2.0)
    psi = pm.Beta("psi", alpha=2.0, beta=2.0)
    y_obs = pm.HurdleGamma("y_obs", psi=psi, alpha=alpha, beta=beta, observed=y)

logp_fn = model.compile_logp()
dlogp_fn = model.compile_dlogp()
ip = model.initial_point()

print(ip)
print(logp_fn(ip))
print(dlogp_fn(ip))

print("\n")
ip2 = {'alpha_log__': np.array(.0), 'beta_log__': np.array(.0), 'psi_logodds__': np.array(.5)}
print(logp_fn(ip2))
print(dlogp_fn(ip2))
```

And this is the output

```python
{'alpha_log__': array(0.69314718), 'beta_log__': array(0.69314718), 'psi_logodds__': array(0.)}
-51.17773249816925
[         nan  -1.25171776 -11.        ]


-59.52213360592764
[         nan   0.12414112 -17.61280388]
```

I'm using:
```shell
Python: 3.13.11
PyMC: 5.27.0
PyTensor: 2.36.3 
```

Were you aware of recent changes that may have affected this?

pirzada-ahmadfaraz:
## Summary

Fixes the `dlogp` function returning NaN for `HurdleGamma` and `HurdleLogNormal` when observed data contains zeros.

## Problem

The gradient computation (`dlogp`) was failing because `pt.where` evaluates both branches before selecting one. When computing `logp(dist, 0)` for continuous distributions like Gamma, this produces `-inf`, which results in NaN gradients - even though that branch is not selected for zero values.

## Solution

Replace zero values with a safe dummy value (1.0) before computing the logp of the continuous distribution:

```python
safe_value = pt.switch(pt.eq(value, 0), 1.0, value)
hurdle_logp = pt.where(
    pt.eq(value, 0),
    pt.log(1 - psi),
    pt.log(psi) + logp(dist, safe_value),  # Use safe_value instead of value
)
```

Since `pt.where` still selects the correct branch based on the condition, the final logp result is unchanged, but the gradient computation is now well-defined.

## Testing

Added regression tests for both `HurdleGamma` and `HurdleLogNormal` that verify `dlogp` does not return NaN values.

Fixes #8053

### DIFF
--- a/tests/distributions/test_mixture.py
+++ b/tests/distributions/test_mixture.py
@@ -1717,3 +1717,49 @@ class TestHurdleDistributions:
                 return np.log(psi) + st.lognorm.logpdf(value, sigma, 0, np.exp(mu))
 
         check_logp(HurdleLogNormal, Rplus, {"psi": Unit, "mu": R, "sigma": Rplusbig}, logp_fn)
+
+    def test_hurdle_gamma_dlogp_no_nan(self):
+        """Test that dlogp does not return NaN for HurdleGamma.
+
+        Regression test for issue #8053. The gradient of pt.where evaluates both
+        branches, so logp(dist, 0) would produce invalid gradients for continuous
+        distributions like Gamma. The fix uses a safe value for the logp computation.
+        """
+        psi_true, alpha_true, beta_true = 0.3, 2.0, 1.5
+
+        dist = HurdleGamma.dist(psi=psi_true, alpha=alpha_true, beta=beta_true)
+        y = draw(dist, draws=50, random_seed=1)
+
+        with pm.Model() as model:
+            alpha = pm.HalfNormal("alpha", sigma=2.0)
+            beta = pm.HalfNormal("beta", sigma=2.0)
+            psi = pm.Beta("psi", alpha=2.0, beta=2.0)
+            HurdleGamma("y_obs", psi=psi, alpha=alpha, beta=beta, observed=y)
+
+        dlogp_fn = model.compile_dlogp()
+        ip = model.initial_point()
+        dlogp_val = dlogp_fn(ip)
+
+        assert not np.any(np.isnan(dlogp_val)), f"dlogp contains NaN: {dlogp_val}"
+
+    def test_hurdle_lognormal_dlogp_no_nan(self):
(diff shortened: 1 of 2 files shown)


ricardoV94 (approved):

ricardoV94:
This is also fixed by https://github.com/pymc-devs/pytensor/pull/1850 but this implementation is more robust

ricardoV94:
Pre-commit is failing

tests/distributions/test_mixture.py
```diff
@@ -1717,3 +1717,49 @@ def logp_fn(value, psi, mu, sigma):
                 return np.log(psi) + st.lognorm.logpdf(value, sigma, 0, np.exp(mu))
 
         check_logp(HurdleLogNormal, Rplus, {"psi": Unit, "mu": R, "sigma": Rplusbig}, logp_fn)
+
+    def test_hurdle_gamma_dlogp_no_nan(self):
```
ricardoV94: Can you pytest.mark.parametrize the test on `HurdleGamma | HurdleLogNormal` ? The logic is nearly the same

pirzada-ahmadfaraz:
@ricardoV94 ping

ricardoV94:
Thanks @pirzada-ahmadfaraz

PR #8058, opened 2026-01-17
logprob for non-overlapping switch accepts equivalent spellings  

### LINKED ISSUE #8049: ENH: extending equivalent non-overlapping switch logp forms
Extend the current measurable rewrite/logp to accept logically equivalent spellings such as `switch(x <= 0, a * x, x)` and swapped comparisons `(0 < x, 0 <= x)`. Users shouldn’t need to guess a single canonical syntax to get logp support. 

Related to: #7543

eclipse1605:
Closes #8049

### DIFF
--- a/pymc/logprob/switch.py
+++ b/pymc/logprob/switch.py
@@ -173,6 +173,7 @@ def _extract_scale_from_measurable_mul(
 
     return neg_branch.owner.inputs[1 - op.measurable_input_idx]
 
+
 @node_rewriter([tensor_switch])
 def find_measurable_switch_non_overlapping(fgraph, node):
     """Detect a non-overlapping scaled-branch switch and replace it by a measurable op.
--- a/pymc/logprob/switch.py
+++ b/pymc/logprob/switch.py
@@ -63,7 +63,8 @@ from pymc.logprob.utils import (
 class MeasurableSwitchNonOverlapping(MeasurableElemwise):
     """Placeholder for switch transforms whose branch images do not overlap.
 
-    Currently matches leaky-ReLU graphs of the form `switch(x > 0, x, a * x)`.
+    Currently matches graphs with branches `(x, a * x)` in either order and a zero-threshold
+    condition on `x` (e.g. `x > 0`, `x <= 0`, `0 < x`).
     """
 
     valid_scalar_types = (Switch,)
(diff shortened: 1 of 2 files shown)


pymc/logprob/switch.py
```diff
@@ -99,12 +104,59 @@ def _is_zero(v: TensorVariable) -> bool:
...
+                return (True, True)
+            if isinstance(scalar_op, GT):
+                # 0 > x  <=>  x < 0
+                return (False, False)
+            if isinstance(scalar_op, GE):
+                # 0 >= x  <=>  x <= 0
+                return (True, False)
+        return None
+
+    return _direct_form() or _swapped_form()
```
ricardoV94: Can we simplify this code?

tests/logprob/test_switch.py
```diff
@@ -23,16 +23,90 @@
...
+    ],
+)
+def test_switch_non_overlapping_logp_matches_change_of_variables(
+    cond_variant, true_branch_is_scaled
+):
     scale = pt.scalar("scale")
     x = pm.Normal.dist(mu=0, sigma=1, size=(3,))
-    y = pt.switch(x > 0, x, scale * x)
+
+    if cond_variant == "x_gt_0":
```
ricardoV94: Why not parametrize already with the objects you gonna need?

pymc/logprob/switch.py
```diff
@@ -179,41 +234,46 @@ def find_measurable_switch_non_overlapping(fgraph, node):
...
 
 
 @_logprob.register(MeasurableSwitchNonOverlapping)
-def logprob_switch_non_overlapping(op, values, cond, x, neg_branch, **kwargs):
+def logprob_switch_non_overlapping(op, values, cond, true_branch, false_branch, **kwargs):
     (value,) = values
 
-    a = _extract_scale_from_measurable_mul(
-        cast(TensorVariable, neg_branch), cast(TensorVariable, x)
+    match = _match_scaled_switch_branches(
```
ricardoV94: Instead of having to guess, you could normalize the switch so that it's always switch`(cond, x, neg_branch)`. You can write `switch(c, t, f) -> switch(~c, f, t)`.

Maybe this allows you to simplify more logic elsewhere?

ricardoV94 (commented):
I'm hoping we can simplify the code a bit further. Also let's get rid of all these `typing.cast`. It's in almost every line. Just mark the file as failing mypy if it requires so much work.

eclipse1605:
> I'm hoping we can simplify the code a bit further. Also let's get rid of all these `typing.cast`. It's in almost every line. Just mark the file as failing mypy if it requires so much work.

sure, is this a general convention we follow throughout the codebase?

pymc/logprob/switch.py
eclipse1605: ya makes sense, though we still can’t use cond directly inside the logprob, because cond depends on the latent x

ricardoV94:
> > I'm hoping we can simplify the code a bit further. Also let's get rid of all these `typing.cast`. It's in almost every line. Just mark the file as failing mypy if it requires so much work.
> 
> sure, is this a general convention we follow throughout the codebase?

No, but in general I favor not making our code a mess because of mypy

eclipse1605:
anyways i misunderstood mypy failing locally in the earlier commit, it wasn't required

pymc/logprob/switch.py
ricardoV94: in the logprob you should be able to assume you already have a canonical form, because that's how you made it in the rewrite

eclipse1605: actually i think we can do this and it should simplify the logprob. we can canonicalize in the rewrite so downstream logprob can assume one shape.

maybe something like this (pseudocode):
```py
# in measurable rewrite:
# detect either ordering + equivalent conditions
#   switch(cond(x ? 0),  x,   a*x)
#   switch(cond(x ? 0), a*x,  x)
# and also swapped comparisons like 0 < x, 0 >= x, ...

(x, ax, sem) = match_branches_and_condition(node)

# normalize to always: switch(cond_canon, x, ax)
if true_branch_is(ax):
    sem = negate(sem)          # because switch(c, ax, x) == switch(~c, x, ax)
cond_canon = cond_from_semantics(x, sem) 

return measurable_switch_non_overlapping(cond_canon, x, ax)

# then in logprob:
# assume inputs are already canonical: switch(cond_canon, x, a*x)
sem = parse_canonical_cond(cond_canon, x)
gate = value_based_gate(value, sem) 
return switch(gate, logp(x, value), logp(a*x, value)) + check(a > 0)
```

ricardoV94: yeah, let's see if it works

eclipse1605:
@ricardoV94 the failing test is flaky right?

eclipse1605:
@ricardoV94 anything left in this?

PR #8061, opened 2026-01-19
Progress bar follow up

ricardoV94:
Follow-up to #8055


pymc/progress_bar/backend.py
```diff
@@ -0,0 +1,13 @@
+#   Copyright 2026 - present The PyMC Developers
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
```
ricardoV94: This file shouldn't be here

pymc/progress_bar/progress.py
```diff
@@ -242,32 +223,41 @@ def update(self, chain_idx: int, is_last: bool, draw: int, tuning: bool, stats)
         )
 
 
+class SimpleProgressBackend(Protocol):
```
ricardoV94, line 225: This complexity that comes from rich API should be hidden away. We should only need a description, counter and total for initialization and an advance function for the main loop

pymc/progress_bar/progress.py
```diff
@@ -81,6 +66,10 @@ def update(
     ) -> None: ...
 
 
+class NullProgressBackend(nullcontext):
+    def update(self, *args, **kwargs): ...
+
+
 class ProgressBarManager:
```
ricardoV94, line 71: We should rename this to reflect to very specific mcmc with multiple chains usage

jessegrabowski: Or try to generalize it more? I'd like if we could do away with the Simple class and use the column-headed table look for everything.

ricardoV94:
@jessegrabowski we need this before we cut a release or we may have install issues because of the css file

ricardoV94:
@jessegrabowski we need this before we cut a release or we may have install issues because of the css file

jessegrabowski:
is it ready for review?

jessegrabowski (approved):

PR #8065, opened 2026-01-21
Implement Uniform and TruncatedNormal dims distributions

ricardoV94:
With the constraint that for free rvs the bounds must be constant. This touches on a limitation we have on transforms that depend on parameters. The previous strategy was to pass the rv inputs to the transform, but this doesn't work for derived variables like `XTensorFromTensor` and `DimShuffled` RVs, since the inputs of those are not the ones needed for the transform.

I'll be thinking about how to better handle this, so we can lift the restriction.

Also fixes a bug in `eval_rv_shapes` with `XTensorVariables` for which `.shape` returns tuples instead of a tensor

### DIFF
--- a/docs/source/api/dims/distributions.rst
+++ b/docs/source/api/dims/distributions.rst
@@ -12,8 +12,10 @@ Scalar distributions
 
    Flat
    HalfFlat
+   Uniform,
    Normal
    HalfNormal
+   TruncatedNormal
    LogNormal
    StudentT
    HalfStudentT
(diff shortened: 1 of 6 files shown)


ricardoV94:
> Also fixes a bug in eval_rv_shapes with XTensorVariables for which .shape returns tuples instead of a tensor

If you wonder why, I think that's what `.shape` should return so I decided to take that approach with XTensorVariables which are fresh anyway. Can reconsider.

This wouldn't have been a problem if pytensor function supported arbitrary tuple/list input/output format like any modern jit library does... We should open an issue for that

OriolAbril (commented):
everything looks and sounds good, I'll try to join the hackathons starting next week to go back to more active contributions on label related things

ricardoV94:
@OriolAbril was there a reason you didn't approve the PR? Did you want someone else to review?

jessegrabowski (approved):

PR #8068, opened 2026-01-23
Update links and remove reference to pymc 4

aloctavodia:
@OriolAbril The link [Exploratory Analysis of Bayesian Models](https://arviz-devs.github.io/EABM/) is giving trouble; you may have already told me how to fix it, but I can't remember.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -89,7 +89,7 @@ repos:
                contributing/versioning_schemes_explanation.md|
                learn/examples.md)
       entry: >
-          (?x)(arviz-devs.github.io|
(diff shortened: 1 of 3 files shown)


OriolAbril:
you have to update the pre-commit check to be more specific. It currently checks for [anything inside](arviz-devs.github.io https://github.com/pymc-devs/pymc/blob/main/.pre-commit-config.yaml#L92) to ensure it is referenced with cross-references instead. It should be updated to `arviz-devs.github.io/arviz` (old docs) like we did in pymc-examples

OriolAbril (approved):

PR #8072, opened 2026-01-25
Derive Categorical from Argmin/Argmax

ricardoV94:
Couldn't resist the temptation...

Needed some import shuffling so that I'm allowed to import a SymbolicRandomVariable in logprob rewrites

### DIFF
--- a/pymc/logprob/order.py
+++ b/pymc/logprob/order.py
@@ -35,14 +35,25 @@
 #   SOFTWARE.
 from typing import cast
 
+import numpy as np
 import pytensor.tensor as pt
 
+from numpy.lib._array_utils_impl import normalize_axis_tuple
+from pytensor.graph import ancestors
 from pytensor.graph.basic import Apply
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.rewriting.basic import node_rewriter
-from pytensor.tensor.math import Max
+from pytensor.scalar import Add, Mul
+from pytensor.tensor import get_underlying_scalar_constant_value
+from pytensor.tensor.elemwise import Elemwise
+from pytensor.tensor.exceptions import NotScalarConstantError
+from pytensor.tensor.math import Argmax, Max, variadic_add, variadic_mul
+from pytensor.tensor.random.basic import ExponentialRV, GumbelRV
+from pytensor.tensor.rewriting.basic import broadcasted_by
+from pytensor.tensor.type_other import NoneTypeT
 from pytensor.tensor.variable import TensorVariable
 
+from pymc.distributions.continuous import WeibullBetaRV
 from pymc.logprob.abstract import (
     MeasurableElemwise,
     MeasurableOp,
@@ -122,6 +133,7 @@ measurable_ir_rewrites_db.register(
     "find_measurable_max",
     find_measurable_max,
     "basic",
+    "order",
     "max",
 )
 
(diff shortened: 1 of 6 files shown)


ricardoV94:
CC @eclipse1605 if you want to review / ask questions

pymc/logprob/order.py
```diff
@@ -35,14 +35,25 @@
 #   SOFTWARE.
 from typing import cast
 
+import numpy as np
 import pytensor.tensor as pt
 
+from numpy.lib._array_utils_impl import normalize_axis_tuple
```
eclipse1605: why are we importing it from a priv numpy module instead of using `numpy.lib.array_utils.normalize_axis_tuple`?

ricardoV94: My editor chose that import automatically, you're right

pymc/logprob/order.py
```diff
@@ -158,3 +170,179 @@ def max_logprob_discrete(op, values, base_rv, **kwargs):
...
+    elif (
+        len(base_node.inputs) == 2
+        and isinstance(base_node.op, Elemwise)
+        and isinstance(base_node.op.scalar_op, Mul)
+        and (
+            (right_neg := is_minus_1(base_node.inputs[1]))
+            or (right_neg := is_minus_1(base_node.inputs[0]))
+        )
+    ):
+        base_var = base_node.inputs[0 if right_neg else 1]
```
eclipse1605: got a notif from this pr so checked it again

when -1 is on the left (-1 * rv), the flag gets set to true anyway, so we keep the constant instead of the rv and the rewrite bails out no?

ricardoV94: Thanks, that was dumb

ricardoV94: It wouldn't be caught because -1 is alway son the left from canonicalization.

pymc/logprob/order.py
```diff
@@ -158,3 +170,176 @@ def max_logprob_discrete(op, values, base_rv, **kwargs):
...
+        and isinstance(base_node.op.scalar_op, Mul)
+        and ((right_neg := is_minus_1(base_node.inputs[1])) or is_minus_1(base_node.inputs[0]))
+    ):
+        base_var = base_node.inputs[0 if right_neg else 1]
+        base_node = base_var.owner
+        if base_node is None:
+            return None
+
+        if isinstance(base_node.op, ExponentialRV):
+            # argmin(exponential(rate)) -> categorical(rate)
```
eclipse1605, line 306: wont this break on plain `Neg` or fused `exp * -scale` unless lift_loc_scale gets there first? idk...

ricardoV94: Neither is a problem. Neg gets canonicalized to mul by -1, and for the second, bailing is exactly the right thing. The rewrites are recalled until saturation (when a pass changes nothing). So if this comes before the lift_loc_scale, it bails, then lift_loc_scale does its thing, and then the next call is a hit. You can see them happening if you enable `config.optimizer_verbose`:

```python
import numpy as np
import pytensor
import pytensor.tensor as pt
from pymc.logprob import logp

pytensor.config.optimizer_verbose = True

probs = np.array([0.1, 0.3, 0.6])
x = pt.argmin(pt.random.exponential(scale=1 / probs), axis=-1)

logp(x, x.type())
# rewriting: rewrite local_neg_to_mul replaces Neg.0 of Neg(exponential_rv.out) with Mul.0 of Mul(ExpandDims{axis=0}.0, exponential_rv.out)
# rewriting: rewrite find_measurable_transforms replaces Mul.0 of Mul([-1.], exponential_rv.out) with MeasurableMul.0 of MeasurableMul(exponential_rv.out, [-1.])
# rewriting: rewrite categorical_from_argmax replaces argmax of Argmax{axis=(0,)}(MeasurableMul.0) with categorical_rv.out ...

x = pt.argmax(pt.random.exponential(size=3) * (-1 / probs), axis=-1)

logp(x, x.type())
# rewriting: rewrite lift_loc_scale replaces MeasurableMul.0 of MeasurableMul(exponential_rv.out, [-10. ... .66666667]) with Mul.0 of Mul(exponential_rv.out, ExpandDims{axis=0}.0)   # <- lifted rv * -1
# rewriting: rewrite find_measurable_transforms replaces Mul.0 ... with MeasurableMul.0 ...
# rewriting: rewrite categorical_from_argmax replaces argmax of Argmax{axis=(0,)}(MeasurableMul.0) with categorical_rv.out ...
```

eclipse1605: on yeah fair my bad

ricardoV94: no need to apologize, they were good questions

eclipse1605: thanks :) but for a bit more clarity, the saturation thing makes sense when lift_loc_scale can actually run

for symbolic outer scale i still dont think that rescue happens because lift_loc_scale bails every pass on constant_fold so graph never changes and hits saturation with neither rewrite matching

```py
import numpy as np
import pytensor
import pytensor.tensor as pt
from pymc.logprob import logp
pytensor.config.optimizer_verbose = True

scale = pt.vector("scale")
x = pt.argmin(pt.random.exponential(scale=1, size=3) * scale, axis=-1)
logp(x, x.type())
# rewriting: rewrite local_neg_to_mul replaces Neg.0 of Neg(Mul.0) with Mul.0 of Mul(ExpandDims{axis=0}.0, Mul.0)
# rewriting: rewrite local_mul_canonizer replaces Mul.0 of Mul(ExpandDims{axis=0}.0, Mul.0) with Mul.0 of Mul(ExpandDims{axis=0}.0, exponential_rv{"()->()"}.out, scale)
# rewriting: rewrite find_measurable_transforms replaces Mul.0 of Mul([-1.], exponential_rv{"()->()"}.out, scale) with MeasurableMul.0 of MeasurableMul(exponential_rv{"()->()"}.out, Mul.0)
# >>>  NotImplementedError: Logprob method not implemented for Argmax{axis=(0,)}

```
constant outer scale does hit lift + categorical

idk if this makes sense, maybe im missing something

ricardoV94: We reject that because scale could be negative and we can't lift negative scales inside the positive distributions. The inline comment says we could do more work say if scale is abs/exp of a root variable then we'd be safe again

eclipse1605: gotit makes sense thanks :)

jessegrabowski (approved):

tests/logprob/test_order.py
```diff
@@ -280,3 +281,171 @@ def test_non_measurable_max_grad(self):
...
+        raises=NotImplementedError,
+        reason="Implicitly depends on https://github.com/pymc-devs/pytensor/issues/1851",
+    )
+    def test_derived_frechet(self):
+        probs = np.array([0.1, 0.2, 0.7])
+        # Frechet is simply an exponentiated gumbel
+        # Since exponentiation is monotonic it doesn't change the result of argmax,
+        # and the probability is therefore the same as that of the argmin(gumbel)
+        s = pt.exp(pt.random.gumbel(loc=pt.log(probs)))
+        x = pt.argmax(s)
```
ricardoV94, line 443: if exp is useless, why is the distribution derived like this?

PR #8083, opened 2026-01-30
Support logcdf, logccdf and icdf for dim RVs

ricardoV94:
Only `_logprob` had a dispatch before

### DIFF
--- a/pymc/dims/distributions/core.py
+++ b/pymc/dims/distributions/core.py
@@ -29,7 +29,7 @@ from pymc import SymbolicRandomVariable, modelcontext
 from pymc.dims.distributions.transforms import DimTransform, log_odds_transform, log_transform
 from pymc.distributions.distribution import _support_point, support_point
 from pymc.distributions.shape_utils import DimsWithEllipsis, convert_dims_with_ellipsis
-from pymc.logprob.abstract import MeasurableOp, _logprob
+from pymc.logprob.abstract import MeasurableOp, _icdf, _logccdf, _logcdf, _logprob
 from pymc.logprob.rewriting import measurable_ir_rewrites_db
 from pymc.logprob.tensor import MeasurableDimShuffle
 from pymc.logprob.utils import filter_measurable_variables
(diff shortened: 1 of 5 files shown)


pymc/dims/distributions/core.py
```diff
@@ -103,16 +103,38 @@ def find_measurable_xtensor_from_tensor(fgraph, node) -> list[XTensorVariable] |
     return [cast(XTensorVariable, new_out)]
 
 
-@_logprob.register(MeasurableXTensorFromTensor)
-def measurable_xtensor_from_tensor(op, values, rv, **kwargs):
-    rv_logp = _logprob(rv.owner.op, tuple(v.values for v in values), *rv.owner.inputs, **kwargs)
+def _to_xtensor(expr, op: MeasurableXTensorFromTensor):
```
jessegrabowski: nit: expr is a vague variable name. I associated it with polars expressions, definitely would not have guessed that we want to pass one of _logprob, _logcdf or _icdf here

tests/dims/distributions/test_core.py
```diff
@@ -192,3 +198,45 @@ def test_multivariate_distribution_dims():
...
                 "x_bad", mu, chol=chol_xr, core_dims=("core1", "core2"), dims=("core2", ...)
             )
+
+
+def test_density_helpers():
+    mean = np.array([1, 2, 3])
+    x = Normal.dist(
+        mu=as_xtensor(mean, dims=("city",)),
+        sigma=1,
+        dim_lengths={"time": 5},
```
jessegrabowski, line 208: There's no time dim declared, is this a non-public API for making a dims-aware RV?

jessegrabowski (approved):

jessegrabowski:
Approved but there's a failing test, i guess unrelated.

tests/dims/distributions/test_core.py
ricardoV94: This is how you specify shape in pymc.dims/xtensor.random variables. There's no size/shape argument, only this. When you are in a model context with a named variable, it forwards this information from `model.dim_lengths`, but if you are working with a `.dist` you work with this manually.

pymc/dims/distributions/core.py
ricardoV94: renamed to `var`

PR #8084, opened 2026-02-02
Add variances dispatch for distributions

jessegrabowski:
## Description
This PR adds a `variance` function that returns the (symbolic) variance of a distribution. I added it because I thought it would be nice to have for the AutoGuide stuff; we can use the `mean` and `variance` of priors to initialize the parameters.
 
This is a 90%+ LLM PR. I went through and tried to clean things up. I mostly copied the patterns from `mean.py`.

## Related Issue

- [ ] Closes #
- [ ] Related to #7530 

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/distributions/moments/means.py
+++ b/pymc/distributions/moments/means.py
@@ -78,9 +78,10 @@ from pymc.distributions.multivariate import (
     MatrixNormalRV,
     MvStudentTRV,
     StickBreakingWeightsRV,
+    WishartRV,
     _LKJCholeskyCovRV,
 )
-from pymc.distributions.shape_utils import rv_size_is_none, maybe_resize
+from pymc.distributions.shape_utils import maybe_resize, rv_size_is_none
 from pymc.exceptions import UndefinedMomentException
 
 __all__ = ["mean"]
@@ -244,7 +245,7 @@ def invgamma_mean(op, rv, rng, size, alpha, beta):
 
 
 @_mean.register(KroneckerNormalRV)
-def kronecker_normal_mean(op, rv, rng, size, mu, covs, chols, evds):
+def kronecker_normal_mean(op, rv, rng, size, mu, sigma, *covs):
     mean = mu
     if not rv_size_is_none(size):
         mean_size = pt.concatenate([size, mu.shape])
@@ -370,7 +371,9 @@ def polya_gamma_mean(op, rv, rng, size, h, z):
 
 
 @_mean.register(RiceRV)
-def rice_mean(op, rv, rng, size, nu, sigma):
+def rice_mean(op, rv, rng, size, b, sigma):
+    # b is the shape parameter, nu = b * sigma is the noncentrality parameter
+    nu = b * sigma
     nu_sigma_ratio = -(nu**2) / (2 * sigma**2)
     return maybe_resize(
         sigma
@@ -448,3 +451,12 @@ def wald_mean(op, rv, rng, size, mu, lam, alpha):
 @_mean.register(WeibullBetaRV)
 def weibull_mean(op, rv, rng, size, alpha, beta):
     return maybe_resize(beta * pt.gamma(1 + 1 / alpha), size)
+
+
+@_mean.register(WishartRV)
+def wishart_mean(op, rv, rng, size, nu, V):
+    mean = nu * V
+    if not rv_size_is_none(size):
+        mean_size = pt.concatenate([size, V.shape[-2:]])
+        mean = pt.full(mean_size, mean)
+    return mean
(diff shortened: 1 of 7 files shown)


pymc/distributions/moments/variances.py
```diff
@@ -0,0 +1,481 @@
...
+def variance(rv: TensorVariable) -> TensorVariable:
+    """Compute the variance of a random variable.
+
+    The only parameter to this function is the RandomVariable
+    for which the value is to be derived.
+    """
+    return _variance(rv.owner.op, rv, *rv.owner.inputs)
+
+
+def maybe_resize(a: TensorVariable, size) -> TensorVariable:
```
lucianopaz: Maybe this and the one in means should be moved to some common utility module?

lucianopaz (approved):
LGTM. My only suggestion is that we move the `maybe_resize` utility from `means` and `variances` to `utils.py`, but I'll approve anyway and you can move it or not.

pymc/distributions/moments/variances.py
jessegrabowski: Let me see if I can just import the one from means and use it. Ricardo made a case against utility modules a while back that I liked, so I want to avoid that.

ricardoV94: May fit fine in existing pytensorf or distributions files

ricardoV94: May fit fine in existing pytensorf or distributions files

jessegrabowski:
I move the `maybe_reize` helper to `distributions/shape_utils.py`. 

I also added variance for some missing distributions (CAR, Wishart, and KroneckerNormal). 

Wishart was missing from `means.py` so I added it. The signature of KroneckerNormal was wrong in means, so I fixed it. And the formula for the mean of the rice distribution was wrong too, so I fixed that as well.

pymc/distributions/moments/variances.py
```diff
@@ -142,7 +147,7 @@ def categorical_variance(op, rv, *args):
 
 @_variance.register(CauchyRV)
 def cauchy_variance(op, rv, rng, size, alpha, beta):
-    raise UndefinedMomentException("The variance of the Cauchy distribution is undefined")
+    return maybe_resize(pt.full_like(alpha, np.inf), size)
```
lucianopaz: I don't know what to say about this. I suppose that it's correct to raise or to return inf, but I'm a bit more in favor of the former. Someone else should do the tie break.

pymc/distributions/moments/means.py
```diff
@@ -370,7 +371,9 @@ def polya_gamma_mean(op, rv, rng, size, h, z):
 
 
 @_mean.register(RiceRV)
-def rice_mean(op, rv, rng, size, nu, sigma):
+def rice_mean(op, rv, rng, size, b, sigma):
+    # b is the shape parameter, nu = b * sigma is the noncentrality parameter
+    nu = b * sigma
```
lucianopaz, line 376: I'm glad that you caught this! I didn't know what I had done wrong to get the distribution moments not to match with the reference.

lucianopaz (commented):
Thanks for the fixes and improvements. I'm a bit unsure whether the variances should return inf or raise an error for Cauchy, Flat et al. I'm a bit more in favor of raising an error, but someone else should tie break.

pymc/distributions/moments/variances.py
jessegrabowski: just to make my case, I was thinking about linalg, where it's maddening when a program fails to execute instead of returning inf/nan.

I recognize this situation is different because we deterministically know which distribution we have, and that the variance doesn't exist -- we won't have an issuer where user sampling will unexpectedly fail due to random draws. 

Still, "is not defined" is the meaning of `nan`.

pymc/distributions/moments/means.py
jessegrabowski: If you assume scipy always uses the most horrible convoluted parameterization possible, you'll always get the right answer

pymc/distributions/moments/variances.py
lucianopaz: Why don't we make it return NaN instead then?

ricardoV94: Raising is fine, it's not much different than not having a dispatch implemented which is always a possibility. 

I'm not throwing a vote though, let me think a bit more

jessegrabowski: I brought back the errors, I can't think of a case where you'd be programmatically trying different distributions and would benefit from not getting runtime errors.

PR #8093, opened 2026-02-06
Validate coordinate and variables don't have the same names

### LINKED ISSUE #7788: BUG: Missing check for collision between coord name and data variable name

Chirag3841:
## Related Issue

- [x] Closes #7788
- [x] Related to #7788

## Checklist

- [x ] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [x ] Included tests that prove the fix is effective or that the new feature works

## Type of change

- [ x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

Description

It adds explicit validation checks in:
Model.add_coord: raises if a coord name already exists in Model.named_vars
Model.add_named_variable: raises if a variable name already exists in Model.coords
This ensures invalid models fail early and avoids silent downstream issues .

### DIFF
--- a/tests/model/test_core.py
+++ b/tests/model/test_core.py
@@ -942,7 +942,7 @@ class TestSetUpdateCoords:
             pmodel.add_coord("nomnom", [1, 2])
 
             # No name collisions
-            with pytest.raises(ValueError, match="same name as"):
+            with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
                 pmodel.add_named_variable(rv, dims="nomnom")
 
             # This should work (regression test against #6335)
@@ -1806,3 +1806,32 @@ class TestModelCopy:
             match="Detected variables likely created by GP objects. Further use of these old GP objects should be avoided as it may reintroduce variables from the old model. See issue: https://github.com/pymc-devs/pymc/issues/6883",
         ):
             copy_method(gaussian_process_model)
+
+
+class TestCoordVariableCollision:
+    def test_variable_name_conflicts_with_existing_coord(self):
+        with pm.Model(coords={"a": [0, 1]}):
+            with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
+                pm.Data("a", [5, 10])
+
+            with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
+                pm.Normal("a", dims="a")
+
+            with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
+                pm.Deterministic("a", pt.ones(2))
+
+            with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
+                pm.Potential("a", pt.ones(2))
+
+    def test_add_coord_conflicts_with_existing_variable_name(self):
+        with pm.Model() as m:
+            pm.Data("a", [5, 10])
+
+            with pytest.raises(ValueError, match="conflicts with an existing model variable name"):
+                m.add_coord("a", [0, 1])
+
+    def test_register_rv_with_dim_matching_name(self):
+        with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
+            with pm.Model() as pmodel:
+                var = pt.as_tensor([1, 2, 3])
+                pmodel.register_rv(var, name="time", dims=("time",))
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -963,6 +963,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
                 "Dimensions can not be named `draw`, `chain` or `__sample__`, "
                 "as those are reserved for use in `InferenceData`."
             )
+        if self.named_vars.tree_contains(name):
+            raise ValueError(
+                f"Dimension name '{name}' conflicts with an existing model variable name."
+            )
         if values is None and length is None:
             raise ValueError(
                 f"Either `values` or `length` must be specified for the '{name}' dimension."
@@ -1466,6 +1470,10 @@ class Model(WithMemoization, metaclass=ContextMeta):
             raise ValueError("Variable is unnamed.")
         if self.named_vars.tree_contains(var.name):
             raise ValueError(f"Variable name {var.name} already exists.")
+        if var.name in self.coords:
+            raise ValueError(
+                f"Variable name '{var.name}' conflicts with an existing dimension name."
+            )
 
         if dims is not None:
             if isinstance(dims, str):
@@ -1473,8 +1481,6 @@ class Model(WithMemoization, metaclass=ContextMeta):
             for dim in dims:
                 if dim not in self.coords and dim is not None:
                     raise ValueError(f"Dimension {dim} is not specified in `coords`.")
-            if any(var.name == dim for dim in dims if dim is not None):
-                raise ValueError(f"Variable `{var.name}` has the same name as its dimension label.")
             # This check implicitly states that only vars with .ndim attribute can have dims
             if var.ndim != len(dims):
                 raise ValueError(
--- a/tests/backends/test_arviz.py
+++ b/tests/backends/test_arviz.py
@@ -15,7 +15,6 @@ import re
 import warnings
 
 import numpy as np
-import pytensor.tensor as pt
 import pytest
 import xarray
 
@@ -702,12 +701,6 @@ class TestDataPyMC:
                 },
             )
             assert isinstance(converter.coords["city"], pd.MultiIndex)
-
-    def test_variable_dimension_name_collision(self):
-        with pytest.raises(ValueError, match="same name as its dimension"):
-            with pm.Model() as pmodel:
-                var = pt.as_tensor([1, 2, 3])
-                pmodel.register_rv(var, name="time", dims=("time",))
 
     def test_include_transformed(self):
         with pm.Model():


Chirag3841:
@drbenvincent 
I have implemented the changes and added tests too. Please have a look.

Chirag3841:
@drbenvincent  @ricardoV94  Any update regarding this PR ?

Chirag3841:
@drbenvincent  Any update on this pr (still needed for implementation or not)?

drbenvincent:
Hi @Chirag3841. Sorry for the delayed response. I don't have permissions to set the workflows running. Review coming momentarily...

drbenvincent (changes requested):
First, thank you for taking on this fix. This addresses a real and subtle bug, and the direction of the change is exactly right: failing early on coord/variable name collisions will prevent hard-to-debug downstream behavior.

The implementation in `Model.add_coord` and `Model.add_named_variable` is well targeted, and the new tests are a good start. I appreciate the clear intent and the added coverage.

## Requested changes before merge

To get this over the line, a couple of small improvements are needed:

1. Improve one error message for clarity and polish
   - Current message in `add_named_variable` reads:
     - `"Variable name '{var.name}' conflicts with an existing coord (same name as)."`
   - The phrase `"(same name as)."` appears incomplete/awkward.
   - Please update it to a clean, direct message (for example: `"Variable name '{var.name}' conflicts with an existing coord name."`).

2. Make tests assert more specific error text
   - Current tests match only `"conflicts"`, which is broad and can pass even if a different branch raises.
   - Please tighten assertions so each test validates the specific collision path and message intent (coord-vs-variable and variable-vs-coord).

3. (Optional but helpful) Rename one test for clearer intent
   - `test_variable_name_conflicts_with_coord_name` currently exercises the case where the variable is created first and then `add_coord` is called.
   - A name reflecting operation order would improve readability for future maintainers.

## Merge readiness

This is very close. With the error-message polish and tighter test assertions, it will be in good shape to merge.

Thanks again for the contribution and for addressing an issue that can produce especially confusing model behavior.

Chirag3841:
> First, thank you for taking on this fix. This addresses a real and subtle bug, and the direction of the change is exactly right: failing early on coord/variable name collisions will prevent hard-to-debug downstream behavior.
> 
> The implementation in `Model.add_coord` and `Model.add_named_variable` is well targeted, and the new tests are a good start. I appreciate the clear intent and the added coverage.
> 
> ## Requested changes before merge
> 
> To get this over the line, a couple of small improvements are needed:
> 
> 1. Improve one error message for clarity and polish
>    - Current message in `add_named_variable` reads:
>      - `"Variable name '{var.name}' conflicts with an existing coord (same name as)."`
>    - The phrase `"(same name as)."` appears incomplete/awkward.
>    - Please update it to a clean, direct message (for example: `"Variable name '{var.name}' conflicts with an existing coord name."`).
> 
> 2. Make tests assert more specific error text
>    - Current tests match only `"conflicts"`, which is broad and can pass even if a different branch raises.
>    - Please tighten assertions so each test validates the specific collision path and message intent (coord-vs-variable and variable-vs-coord).
> 
> 3. (Optional but helpful) Rename one test for clearer intent
>    - `test_variable_name_conflicts_with_coord_name` currently exercises the case where the variable is created first and then `add_coord` is called.
>    - A name reflecting operation order would improve readability for future maintainers.
> 
> ## Merge readiness
> 
> This is very close. With the error-message polish and tighter test assertions, it will be in good shape to merge.
> 
> Thanks again for the contribution and for addressing an issue that can produce especially confusing model behavior.
> 
@drbenvincent Thanks for the clarification and reviewing this PR. I will address the following changes you requested.
I reordered the checks in `add_named_variable` so that existing validations (like dims and dimension-name conflicts) run before the new coord–variable name check. This keeps the previous behavior intact while still catching the new collision case.
Changed test functions name to enhance readability and improve error message for polish and clarity.

image
all test cases passed locally as per the requirement of this PR.

Chirag3841:
Any update ?

drbenvincent:
Hi. I'm day 3 of a fever. I'll be catching up over the week ahead

Chirag3841:
> Hi. I'm day 3 of a fever. I'll be catching up over the week ahead

Thanks for the clarification.
Waiting for your further response.

Chirag3841:
> Hi. I'm day 3 of a fever. I'll be catching up over the week ahead

Hi! Hope you're feeling better 
Just a gentle ping on this PR whenever you have time. All requested changes have been addressed. Happy to make any further updates if needed. Thanks!

Chirag3841:
@drbenvincent @ricardoV94 Any update regarding this PR?

drbenvincent:
Sorry for the delay, I know that it's frustrating, but work has been busy.

This issue is a duplicate of #7794 by @twiecki, reviewed by @ricardoV94.

drbenvincent (commented):
## Review: Detailed assessment

Thanks again for the contribution, Chirag. The core fix is correct — both collision directions are now guarded, and the exact reproducer from #7788 would raise a clear `ValueError`. The scoping also works across nested models since `_coords` is a shared dict reference and `named_vars` uses `tree_contains` to traverse the parent chain.

There is one bug and a few test coverage gaps to address before this is merge-ready.

---

### Bug: Check ordering in `add_named_variable` — state mutation before validation

The new coord-collision check is placed **after** the dims processing block, which includes the state mutation `self.named_vars_to_dims[var.name] = dims`:

```python
# dims block runs first...
            self.named_vars_to_dims[var.name] = dims    # ← state mutation

# ...then the new check
        if var.name in self.coords:                      # ← validation after mutation
            raise ValueError(...)
```

If a variable has dims AND its name collides with a coord, `named_vars_to_dims` gets populated before the `ValueError` is raised. This leaves an orphaned entry — a dims mapping for a variable that was never actually registered in `named_vars`. Since `named_vars_to_dims` is a `treedict`, the mutation propagates to parent models too.

**Fix:** Move the `if var.name in self.coords` check to immediately after the existing `tree_contains` check, before the dims block. All validation should precede all state mutations.

---

### Tests placed in wrong class

The two new test methods are appended at the end of the file, which places them inside `TestModelCopy`. These are collision tests, not copy tests. Please move them into their own class (e.g. `TestCoordVariableCollision`).

---

### Test coverage gaps

The current tests only cover two `pm.Data` scenarios. Several important cases are missing:

1. **RV colliding with a coord** — this is the more dangerous case since it's what caused the original silent `sample_posterior_predictive` bug from #7788:

   ```python
   def test_rv_name_conflicts_with_existing_coord(self):
       with pytest.raises(ValueError, match="conflicts with an existing coord name"):
           with pm.Model(coords={"a": [0, 1]}):
               pm.Normal("a", dims="a")
   ```

2. **Deterministic colliding with a coord** — `Deterministic` and `Potential` also go through `add_named_variable`:

   ```python
   def test_deterministic_name_conflicts_with_existing_coord(self):
       with pytest.raises(ValueError, match="conflicts with an existing coord name"):
           with pm.Model(coords={"a": [0, 1]}):
               pm.Deterministic("a", pt.ones(2))
   ```

3. **Auto-coord creation colliding with existing variable** — `register_rv` auto-creates coords from unknown dims (line 1241 in `core.py`), which hits the `add_coord` check:

   ```python
   def test_auto_coord_from_dims_conflicts_with_existing_variable(self):
       with pytest.raises(ValueError, match="conflicts with an existing model variable name"):
           with pm.Model() as m:
               pm.Data("x", [5])
               pm.Normal("b", dims="x")  # auto-creates coord "x" → collides
   ```

4. **Happy-path regression guard** — confirm that distinct coord and variable names still work together without raising:

   ```python
   def test_different_coord_and_variable_names_coexist(self):
       with pm.Model(coords={"a": [0, 1]}) as m:
           pm.Data("b", [5, 10], dims="a")
       assert "b" in m.named_vars
       assert "a" in m.coords
   ```

---

### Housekeeping

PR #7794 is still open and addresses the same issue. One of the two should be closed to avoid duplicate effort.

---

### Summary

| Item | Severity |
|---|---|
| Check ordering (state mutation before validation) | Medium — bug |
| Tests in wrong class (`TestModelCopy`) | Low |
| Missing RV / Deterministic collision tests | Medium |
| Missing auto-coord collision test | Medium |
| Missing happy-path regression test | Low |
| Duplicate PR #7794 still open | Housekeeping |

The direction is right and the fix addresses the reported issue. With the check reordering and broader test coverage, this will be in good shape to merge.

drbenvincent:
## Duplicate PR assessment: #7794 vs #8093

@twiecki @ricardoV94 — flagging this for your input before we make a final call.

PR #7794 (opened May 2025 by @twiecki) and this PR (#8093, opened Feb 2026 by @Chirag3841) both address #7788 with the same approach: adding collision checks in `add_coord` and `add_named_variable`. Here's a head-to-head comparison:

| Aspect | #7794 (twiecki) | #8093 (Chirag) |
|---|---|---|
| Check ordering in `add_named_variable` | **Correct** — coord check before state mutations | **Bug** — coord check placed after `named_vars_to_dims` mutation |
| `add_coord` nested-model scoping | `name in self.named_vars` (current model only) | **Better** — `self.named_vars.tree_contains(name)` (traverses parents) |
| Reviewer | @ricardoV94 (core) | @drbenvincent |
| Error message terminology | Ricardo suggested "dimension" (PyMC convention) | Uses "coord name" |
| Test class placement | Wrong (`TestNested`) | Wrong (`TestModelCopy`) |
| Test breadth | `pm.Data` only | `pm.Data` only |
| Pre-existing test insight | Ricardo flagged existing partial handling that needs investigation | Not investigated |
| Activity | Stale ~11 months | Active |
| CI coverage | Full suite ran | Minimal (labeler + pre-commit only) |

### Preliminary recommendation

Pursue #7794 and close this PR, for these reasons:

1. **Correct check ordering.** #7794 places the coord validation before all state mutations. #8093 has a real (if minor) bug where `named_vars_to_dims` gets an orphaned entry if the coord check fails.

2. **Maintainer review already done.** Ricardo's review on #7794 is specific and actionable, including the observation about a pre-existing failing test that hints at existing partial handling — that investigation should happen on the PR where it was raised.

3. **Error message convention.** Ricardo asked for "dimension" rather than "coordinate", consistent with PyMC terminology.

4. **One thing to cherry-pick from this PR:** The `add_coord` check here correctly uses `self.named_vars.tree_contains(name)` instead of `name in self.named_vars`, which properly handles nested models. This improvement should be adopted into #7794.

@Chirag3841 — thank you for the contribution and for your patience through the review process. The `tree_contains` improvement you made in `add_coord` is genuinely better than the corresponding code in #7794, and we'd want to carry that forward. We're waiting for input from @twiecki and @ricardoV94 before making a final call on which PR to proceed with.

Chirag3841:
Thanks for the detailed review!
Regarding #7794, I understand the preference. However, since that PR seems inactive for quite some time.
Please let me know how you'd like to proceed before closing this PR #8093.
Here, we can create new class for placing the following test files as per the requirement for solving the issue which passed all test cases and changing coord -> dimension in both cases in core.py too as well as in test_core.py .
And  While implementing the validation ordering fix, I noticed an interaction with an existing test (test_add_named_variable_checks_dim_name).
In the case where var.name matches a dimension label (via dims), the expected behavior is to raise the "same name as" error. If the coord-collision check is performed first, it raises a different error ("conflicts with an existing dimension name"), causing that test to fail.
To preserve the intended behavior, I ensured that the dims-related validation (including the "same name as" check) runs before the coord-collision check, while still keeping all validation before any state mutation.

```
class TestCoordVariableCollision:
    def test_variable_name_conflicts_with_existing_coord(self):
        with pytest.raises(
            ValueError,
            match="Variable name 'a' conflicts with an existing dimension name.",
        ):
            with pm.Model(coords={"a": [0, 1]}):
                pm.Data("a", [5, 10])

    def test_add_coord_conflicts_with_existing_variable_name(self):
        with pytest.raises(
            ValueError,
            match="Dimension name 'a' conflicts with an existing model variable name.",
        ):
            with pm.Model() as m:
                pm.Data("a", [5, 10])
                m.add_coord("a", [0, 1])

    def test_rv_name_conflicts_with_existing_coord(self):
        with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
            with pm.Model(coords={"a": [0, 1]}):
                pm.Normal("a", dims="a")

    def test_deterministic_name_conflicts_with_existing_coord(self):
        with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
            with pm.Model(coords={"a": [0, 1]}):
                pm.Deterministic("a", pt.ones(2))

    def test_auto_coord_conflicts_with_existing_variable(self):
        with pytest.raises(
            ValueError, match="conflicts with an existing model variable name"
        ):
            with pm.Model() as m:
                pm.Data("x", [5])
                m.add_coord("x", [0])

    def test_different_coord_and_variable_names_coexist(self):
        with pm.Model(coords={"a": [0, 1]}) as m:
            pm.Data("b", [5, 10], dims="a")

        assert "b" in m.named_vars
        assert "a" in m.coords
    ```

**Comment by drbenvincent** [comment]
Yes, I won't auto close this. The other pr has been inactive and you've been active 🚀. So maybe we'll close the other one. Just waiting to get input from those involved before we do anything

File: tests/model/test_core.py [hunk 1]
```diff
@@ -1809,3 +1809,45 @@ def test_guassian_process_copy_failure(self, copy_method) -> None:
...
+            with pm.Model() as m:
+                pm.Data("x", [5])
+                m.add_coord("x", [0])
+
+    def test_different_coord_and_variable_names_coexist(self):
+        with pm.Model(coords={"a": [0, 1]}) as m:
+            pm.Data("b", [5, 10], dims="a")
+
+        assert "b" in m.named_vars
+        assert "a" in m.coords
```
ricardoV94: We don't need this test, it's certainly tested implicitly anywhere we use dims

tests/model/test_core.py
```diff
@@ -1809,3 +1809,45 @@ def test_guassian_process_copy_failure(self, copy_method) -> None:
...
+
+    def test_rv_name_conflicts_with_existing_coord(self):
+        with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
+            with pm.Model(coords={"a": [0, 1]}):
+                pm.Normal("a", dims="a")
+
+    def test_deterministic_name_conflicts_with_existing_coord(self):
+        with pytest.raises(ValueError, match="conflicts with an existing dimension name"):
+            with pm.Model(coords={"a": [0, 1]}):
+                pm.Deterministic("a", pt.ones(2))
```
ricardoV94: These can be merged with the previous test, just call each one `pm.Data`, `pm.Normal`, `pm.Deterministic` (you can add pm.Potential) while we are at it, in succession, each one inside their pytest.raises (which should have been just around the line that triggers, not pm.Model())

tests/model/test_core.py
```diff
@@ -1809,3 +1809,45 @@ def test_guassian_process_copy_failure(self, copy_method) -> None:
             match="Detected variables likely created by GP objects. Further use of these old GP objects should be avoided as it may reintroduce variables from the old model. See issue: https://github.com/pymc-devs/pymc/issues/6883",
         ):
             copy_method(gaussian_process_model)
+
+
+class TestCoordVariableCollision:
+    def test_variable_name_conflicts_with_existing_coord(self):
+        with pytest.raises(
```
ricardoV94: Put pytest.raises aound pm.Data, (the line that actually raises)

ricardoV94 (changes requested):
Just some test cleanup

ricardoV94:
@drbenvincent tell the bot to cut those reports to 2-3 lines, essay isn't needed.

Chirag3841:
@ricardoV94 Changes addressed as per the requirement in test_core.py.
Could you please suggest how to deal with this issue:-
```
FAILED tests/model/test_core.py::TestSetUpdateCoords::test_add_named_variable_checks_dim_name - AssertionError: Regex pattern did not match.
  Expected regex: 'same name as'
  Actual message: "Variable name 'nomnom' conflicts with an existing dimension name."
```

Chirag3841:
@ricardoV94 Any Update regarding this Pr?

Chirag3841:
@ricardoV94 Sorry for the ping just asking if there’s any update regarding this PR?

ricardoV94:
@Chirag3841 can you move the target branch to pymc-devs:v6? main is now frozen while we work on the next main release.

On github, editing the title gives you this option (weird UI)

Chirag3841:
@ricardoV94 Done as per the requirement. Getting same error as I mentioned in above comment. Could you please provide me any suggestion to resolve that issue?

Chirag3841:
@ricardoV94 Any update?

ricardoV94:
> @ricardoV94 Done as per the requirement. Getting same error as I mentioned in above comment. Could you please provide me any suggestion to resolve that issue?

You changed the error message that was triggered (check where it was triggered from) and the pre-existing test now fails because it expected the old error message. There may also be redundancy between your new tests and the old one.

ricardoV94 (approved):

ricardoV94:
failing CI is not your fault now, we renamed v3 -> main. I'll pick this up once we transition. Changes look good otherwise

Chirag3841:
> failing CI is not your fault now, we renamed v3 -> main. I'll pick this up once we transition. Changes look good otherwise

Thanks for you review .

ricardoV94:
Thanks @Chirag3841

Issue #7780, opened 2025-05-09
Replace `assert` checks in `logprob.scan` with real error handling


jessegrabowski:
### Description

There are two annoying assert checks that I commonly run up against when writing time series models [here](https://github.com/pymc-devs/pymc/blob/ef26ae88e87c2120c2700d062e404ed1f777d358/pymc/logprob/scan.py#L115) and [here](https://github.com/pymc-devs/pymc/blob/ef26ae88e87c2120c2700d062e404ed1f777d358/pymc/logprob/scan.py#L116). Other users are hitting them too, see for example [here](https://discourse.pymc.io/t/smooth-local-linear-trend-with-customdist/16968/6).

It would be much nicer to have a real error message that explains what has gone wrong and why, or at least gives some information about which parts of the graph are going wrong. 

I'm not 100% sure how one ends up hitting these asserts, so I tagged this with needs info. It seems to have to do with having multiple outputs from the scan inner function in a custom dist.

ricardoV94:
In general we should address https://github.com/pymc-devs/pymc/issues/6691 to tell users why logprob inference failed. The assert failing is a bug though, when we can't derive the logprob we should not even reach that line, and fallback to the default uninformative `RuntimeError: Logprob could not be derived`

PR #8095, opened 2026-02-06
Replace assert checks in logprob.scan with explicit error handling

swarnaliiiiii:
## Summary

This PR replaces internal `assert` checks in `logprob.scan` with explicit, user-facing errors and adds early validation to prevent reaching those assertions.

## Motivation

Users working with time series models that rely on `Scan` (e.g. via `CustomDist`) have reported hitting internal assertions during log-probability computation, resulting in uninformative `AssertionError`s. This makes it difficult to understand why log-probability derivation failed.

A concrete example is discussed in the PyMC Discourse thread:
https://discourse.pymc.io/t/smooth-local-linear-trend-with-customdist/16968

## What this PR does

- Adds an early guard in `logprob_scan` to explicitly reject scans with multiple random outputs, which are currently unsupported.
- Replaces internal `assert` statements in `convert_outer_out_to_in` with explicit `RuntimeError`s that explain why log-probability derivation failed.

## What this PR does NOT do

- This PR does **not** attempt to support multi-output scans.
- It does **not** change Scan semantics or PyTensor internals.
- It does **not** alter any supported behavior; it only improves failure modes.

## Related issue

addresses #7780

### DIFF
--- a/pymc/logprob/scan.py
+++ b/pymc/logprob/scan.py
@@ -112,8 +112,18 @@ def convert_outer_out_to_in(
             oo_var, field_filter=lambda x: x.startswith("outer_out")
         )
 
-        assert var_info is not None
-        assert oo_var in new_outer_input_vars
+        if var_info is None:
+            raise RuntimeError(
+                "Failed to compute log-probability for Scan: "
+                "could not match an outer output to an inner output. "
+                "This usually indicates an unsupported Scan structure."
+            )
+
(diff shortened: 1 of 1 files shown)


vkverma9534:
@swarnaliiiiii I think the failing pre-commit can be auto-fixed locally as well refer to https://pandas.pydata.org/docs/dev/development/contributing_codebase.html#pre-commit you can follow it up locally

PR #8096, opened 2026-02-08
`dims.Data` creates `XTensorSharedVariables`

ricardoV94:
Depends on https://github.com/pymc-devs/pytensor/pull/1882

We were registering regular TensorSharedVariables in `pymc.dims.Data`, and then wrapping those in `as_xtensor`. The problem with that is that when you then try to act on `m["x"]` you were not acting on the same variable that `pymc.dims.Data` returned to you, instead  you were acting on the input of `as_xtensor`...

This made it hard to adopt it in pymc-marketing

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -32,7 +32,7 @@ import pytensor.sparse as sparse
 import pytensor.tensor as pt
 import scipy.sparse as sps
 
-from pytensor.compile import DeepCopyOp, Function, ProfileStats, get_mode
+from pytensor.compile import DeepCopyOp, Function, ProfileStats, get_mode, view_op
 from pytensor.compile.sharedvalue import SharedVariable
 from pytensor.graph.basic import Constant, Variable
 from pytensor.graph.traversal import ancestors, explicit_graph_inputs, graph_inputs
@@ -2263,7 +2263,7 @@ def Deterministic(name, var, model=None, dims=None):
     random variables.
     """
     model = modelcontext(model)
-    var = var.copy(model.name_for(name))
+    var = view_op(var, name=model.name_for(name))
     model.deterministics.append(var)
     model.add_named_variable(var, dims)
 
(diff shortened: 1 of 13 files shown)


pymc/model/core.py
```diff
@@ -2263,7 +2263,7 @@ def Deterministic(name, var, model=None, dims=None):
     random variables.
     """
     model = modelcontext(model)
-    var = var.copy(model.name_for(name))
+    var = view_op(var, name=model.name_for(name))
```
jessegrabowski, line 2266: What's the difference? Maybe a better question - why were we copying in the first place?

pymc/model/fgraph.py
```diff
@@ -280,7 +279,7 @@ def fgraph_from_model(
         fgraph.remove_output(first_idx_to_remove)
 
     # Now that we have Deterministic dummy Ops, we remove the noisy `Identity`s from the graph
-    remove_identity_rewrite.apply(fgraph)
+    remove_view_rewrite.apply(fgraph)
```
jessegrabowski, line 282: update comment since we're removing View now

tests/dims/test_model.py
```diff
@@ -106,7 +117,7 @@ def test_complex_model():
 
     coords = {
         "group": range(3),
-        "knots": range(N_knots),
+        "knot": range(N_knots),
```
jessegrabowski, line 120: singular dims gang rise up!

jessegrabowski (approved):
approved with some stupid questions and a nitpick about a comment, typical PR review smh

pymc/model/core.py
ricardoV94: You need an op sometimes because you can't have x be both a free rv and a deterministic, but x and view(x) is fine. Copy was another way of achieving this but it's Elemwise or XElemwise Op depending on the type. View instead works with any input type (and achieves the same). 

Even variables without a .copy method would work in the future.

The immediate motivation for changing though was to keep pretending xtensorvariables don't exist as much as possible in the rest of the codebase

PR #8101, opened 2026-02-09
Refactor transformed_conditional_logp to remove IR dummy Ops (issue #8100)

### LINKED ISSUE #8100: Refactor value transform logic in logprob to avoid dummy Ops and complex rewrites
Refactor value transform logic in `logprob` to avoid dummy Ops and complex rewrites

The current implementation of value transforms in `pymc/logprob/transform_value.py` is overly convoluted. It relies on dummy Ops (`TransformedValue`, `TransformedValueRV`) and a complex `TransformValuesRewrite` that is injected into `conditional_logp` via `extra_rewrites`.

This complexity is unnecessary because we can achieve the same result by simple graph manipulation in two stages:

1.  **Stage 1: Derive Logp with back-transformed values.**
    Instead of using a rewriter to inject transforms during the `conditional_logp` walk, we can pre-process the value variables. For each Random Variable `rv` with an associated `transform`, we can use its back-transformed version `val_constrained = transform.backward(val_unconstrained, *rv.owner.inputs)` as the value variable for `conditional_logp`.
    
    `conditional_logp` will then:
    *   Calculate the log-probability of `rv` evaluated at `val_constrained`.
    *   Correcty replace any occurrences of `rv` in *other* log-probability terms with `val_constrained`.

2.  **Stage 2: Apply Jacobian correction.**
    After `conditional_logp` returns the log-probability terms, we can add the Jacobian correction `transform.log_jac_det(val_unconstrained, *rv.owner.inputs)` to the term associated with `rv`. We must ensure that any Random Variables remaining in the Jacobian expression are also replaced by their respective value variables (using `replace_rvs_by_values`).

### Benefits
*   Eliminates `TransformedValue`, `TransformedValueRV`, and `TransformValuesRewrite`.
*   Removes the need for a specialized `_logprob` implementation for `TransformedValueRV`.
*   Makes the `logprob` derivation process more transparent and easier to debug.
*   Simplifies the internal IR by removing dummy Ops that "should not be present in the final graph".

### Proposed Implementation Sketch

```python
def transformed_conditional_logp(
    rvs, 
    rvs_to_values, 
    rvs_to_transforms, 
    jacobian=True, 
    **kwargs
):
    # 1. Prepare value variables (constrained for those with transforms)
    logp_rv_values = {}
    for rv, val in rvs_to_values.items():
        transform = rvs_to_transforms.get(rv)
        if transform:
            logp_rv_values[rv] = transform.backward(val, *rv.owner.inputs)
        else:
            logp_rv_values[rv] = val
    
    # 2. Derive logp terms
    logp_terms = conditional_logp(logp_rv_values, **kwargs)
    
    # 3. Apply Jacobian and map back to unconstrained value variables
    final_terms = []
    for rv in rvs:
        val_unconstrained = rvs_to_values[rv]
        val_constrained = logp_rv_values[rv]
        logp = logp_terms[val_constrained]
        
        transform = rvs_to_transforms.get(rv)
        if transform and jacobian:
            jac = transform.log_jac_det(val_unconstrained, *rv.owner.inputs)
            # Replace RVs in Jacobian
            [jac] = replace_rvs_by_values([jac], rvs_to_values=rvs_to_values, rvs_to_transforms=rvs_to_transforms)
            
            # Handle potential dimension mismatch (logic from current transformed_value_logprob)
            if jac.ndim < logp.ndim:
                logp = logp.sum(axis=np.arange(jac.ndim - logp.ndim, 0))
            
            logp += jac
            
        final_terms.append(logp)
        
    return final_terms
```

This refactoring would significantly clean up `pymc/logprob/transform_value.py` and improve the overall maintainability of the `logprob` module.

mikjkd:
## Description
Replace the TransformValuesRewrite mechanism with a simpler two-stage approach as suggested by Ricardo Vieira in the issue #8100 : apply transform.backward() before conditional_logp, then add Jacobian correction afterward. This eliminates TransformedValue and TransformedValueRV dummy Ops from the intermediate representation.

## Related Issue
- [x] Closes #8100 
- [ ] Related to #

## Checklist

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/logprob/basic.py
+++ b/pymc/logprob/basic.py
@@ -58,9 +58,8 @@ from pymc.logprob.abstract import (
     _logprob_helper,
 )
 from pymc.logprob.rewriting import cleanup_ir, construct_ir_fgraph
-from pymc.logprob.transform_value import TransformValuesRewrite
 from pymc.logprob.transforms import Transform
-from pymc.logprob.utils import get_related_valued_nodes
+from pymc.logprob.utils import get_related_valued_nodes, replace_rvs_by_values
 from pymc.pytensorf import expand_inner_graph, replace_vars_in_graphs
 
 TensorLike: TypeAlias = Variable | float | np.ndarray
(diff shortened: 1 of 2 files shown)


pymc/logprob/basic.py
```diff
@@ -623,45 +622,111 @@ def transformed_conditional_logp(
...
+    final_terms = []
     for rv in rvs:
-        value_var = rvs_to_values[rv]
-        logp_terms[value_var] = temp_logp_terms[value_var]
+        val_unconstrained = rvs_to_values[rv]
+        val_constrained = logp_rv_values[rv]
+        logp = logp_terms[val_constrained]
+
+        transform = rvs_to_transforms.get(rv)
+        if transform is not None:
```
eclipse1605, line 688: @ricardoV94 here the jacobian is computed unconditionally whenever a transform exists and the jacobian flag is only checked at addition time. wont this lead to extra graph construction and replacement work even when caller disables jacobian. also transform.log_jac_det might fail on paths that should have been skipped when jacobian=False.

PR #8103, opened 2026-02-10
Scalar Resolution with Regression Test

### LINKED ISSUE #7429: Debug message for scalar values fails
https://github.com/pymc-devs/pymc/blob/e988bc5416d8df6d32e4e984173d0f91c9644823/pymc/model/core.py#L1966

raises an error when variable is scalar

```
IndexError                                Traceback (most recent call last)
Cell In[23], line 1
----> 1 model.debug(fn="dlogp")

File [/dev/pymc/pymc/model/core.py#line=1965), in Model.debug(self, point, fn, verbose)
   1962 print_(
   1963     f"Some of the{observed}values of variable {rv} are associated with a non-finite {fn}:"
   1964 )
   1965 mask = ~np.isfinite(rv_fn_eval)
-> 1966 for value, fn_eval in zip(values[mask], rv_fn_eval[mask]):
   1967     print_(f" value = {value} -> {fn} = {fn_eval}")
   1968 print_()

IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
```

AdishreeB:
## Description
This PR fixes an IndexError that occurs when using model.debug() on scalar variables. 

The issue was that the debug function attempted to index scalar variables using values[mask], which fails because scalars are 0-dimensional and cannot be indexed. 

The fix adds a dimension check (values.ndim == 0) before indexing: 
- For scalar variables: handles the mask as a single boolean value 
- For array variables: uses the original indexing logic 

This ensures the debug function works correctly for both scalar and array random variables.

A regression test is included to prevent this bug from reappearing.

## Related Issue
- [x] Closes #7429
- [ ] Related to #

## Checklist
- [ ] If you are a pro: each commit corresponds to a relevant logical change

## Type of change
- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):


tests/model/test_core.py
```diff
@@ -1809,3 +1804,14 @@ def test_guassian_process_copy_failure(self, copy_method) -> None:
...
         ):
             copy_method(gaussian_process_model)
+
+
+def test_model_debug_scalar_resolution():
+    with pm.Model() as model:
+        pm.Normal("x", 0, 1)
+
+    try:
+        model.debug(fn="dlogp")
```
ricardoV94: test the output directly, instead of just testing it doesn't raise. You can use capsys to capture it.

Also would be nice to test a case that has an error and one that doesn't. Both presumably failed before, but we are still not testing one of them.

You just need to give a negative sigma to the Normal to get a case that fails (at least the logp, not sure about the dlgop).

ricardoV94 (approved):

ricardoV94:
pre-commit is failing

AdishreeB:
i dont understand why the pre-commit is failing

AdishreeB:
Also if it is okay to note - The 2 failing checks are unrelated to this PR. The `all_tests` failure is just a cascading summary job, and `test_deterministics_out_of_idata[True]` is a pre-existing failure in `test_forward.py` and neither file was touched for change in this PR.

PR #8105, opened 2026-02-11
Broadcast probability functions with value and size

ricardoV94:
Since our syntax is `foo(rv, value)`, we should respect the shape of the rv. We were only respecting the one implied by the parameters, but not size. 

`pm.logp(pm.Normal.dist(size=(10, 3)), 0))` would return a single item tensor

This PR also tests we are allowed to further broadcast with value.

`pm.logp(pm.Normal.dist(size=(10, 3)), zeros((3, 10, 3))`

Impl-wise I think it's cleaner to not include size in the definitions themselves as they are our construct. 

This should also be compatible out of the box when we replace ours by those in pytensor-distributions.

### DIFF
--- a/conda-envs/environment-alternative-backends.yml
+++ b/conda-envs/environment-alternative-backends.yml
@@ -7,7 +7,7 @@ dependencies:
 # Base dependencies
 - arviz>=0.13.0
 - blas
-- cachetools>=4.2.1
+- cachetools>=4.2.1,<7
 - cloudpickle
 - zarr>=2.5.0,<3
 - numba
--- a/pymc/dims/distributions/core.py
+++ b/pymc/dims/distributions/core.py
@@ -19,6 +19,8 @@ import numpy as np
 
 from pytensor.graph import node_rewriter
 from pytensor.graph.basic import Variable
+from pytensor.tensor import TensorVariable
+from pytensor.tensor import expand_dims as pt_expand_dims
 from pytensor.tensor.elemwise import DimShuffle
 from pytensor.tensor.random.op import RandomVariable
 from pytensor.xtensor import as_xtensor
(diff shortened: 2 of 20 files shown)


tests/dims/distributions/test_core.py
```diff
@@ -29,6 +29,7 @@
 pytestmark = pytest.mark.filterwarnings(
     "error",
     "ignore::numba.core.errors.NumbaPerformanceWarning",
+    "ignore:create_index_for_new_dim:UserWarning",
```
ricardoV94, line 32: already fixed upstream, but we haven't released yet

pymc/logprob/transforms.py
```diff
@@ -269,7 +270,7 @@ def measurable_transform_logcdf(op: MeasurableTransform, value, *inputs, **kwarg
             logcdf_zero = _logcdf_helper(measurable_input, 0)
             logcdf = pt.switch(
                 pt.lt(backward_value, 0),
-                pt.log(pt.exp(logcdf_zero) - pt.exp(logcdf)),
+                logdiffexp(logcdf_zero, logcdf),
```
ricardoV94, line 273: With the broadcasted logp, this started triggering https://github.com/pymc-devs/pytensor/pull/1883

So I fixed it in our math helper to go around the bug, and avoid having to wait for a pytensor release

conda-envs/environment-alternative-backends.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
 # Base dependencies
 - arviz>=0.13.0
 - blas
-- cachetools>=4.2.1
+- cachetools>=4.2.1,<7
```
ricardoV94, line 10: unrelated issue: https://github.com/pymc-devs/pymc/issues/8106

pymc/logprob/transforms.py
jessegrabowski: add a TODO that links to the pytensor issue and says to remove this once its fixed upstream

pymc/distributions/distribution.py
```diff
@@ -132,45 +133,61 @@ def change_dist_size(op, rv, new_size, expand):
...
                     if isinstance(op, RandomVariable):
                         rng, size, *dist_params = dist_params
-                    elif params_idxs:
-                        dist_params = [dist_params[i] for i in params_idxs]
-                    return class_logcdf(value, *dist_params)
+                    else:
+                        if size_idx is not None:
+                            size = dist_params[size_idx]
+                        if params_idxs is not None:
+                            dist_params = [dist_params[i] for i in params_idxs]
```
jessegrabowski: Refactor this out to a helper, it's repeated 4x

tests/distributions/test_censored.py
```diff
@@ -114,14 +114,14 @@ def test_dist_broadcasted_by_lower_upper(self):
         assert tuple(x.owner.inputs[0].shape.eval()) == (3, 4, 2)
 
     def test_censored_categorical(self):
-        cat = pm.Categorical.dist([0.1, 0.2, 0.2, 0.3, 0.2], shape=(5,))
+        cat = pm.Categorical.dist([0.1, 0.2, 0.2, 0.3, 0.2], shape=())
```
jessegrabowski, line 117: Why change the shapes in these tests?

jessegrabowski: Oh I think I get it, because previously you had to say 5 to match the incoming parameter shape?

pymc/dims/distributions/core.py
```diff
@@ -103,38 +105,67 @@ def find_measurable_xtensor_from_tensor(fgraph, node) -> list[XTensorVariable] |
     return [cast(XTensorVariable, new_out)]
 
 
-def _to_xtensor(var, op: MeasurableXTensorFromTensor):
+def _to_tensor(op: MeasurableXTensorFromTensor, value: XTensorVariable) -> TensorVariable:
+    # Align dims thate are shared between value and op to the right
```
jessegrabowski: typo: that (can't suggest im in commit-by-commit mode)

pymc/dims/distributions/core.py
```diff
@@ -103,38 +105,67 @@ def find_measurable_xtensor_from_tensor(fgraph, node) -> list[XTensorVariable] |
...
+    value_dims_set = set(value.dims)
+    shared_dims = [dim for dim in op.dims if dim in value_dims_set]
+    value = value.transpose(..., *shared_dims)
+    # Add dummy broadcastable dimensions for dimensions present in the op but missing in the value
+    n_value_unique_dims = len(value_dims_set) - len(shared_dims)
+    missing_axis = [
+        i for i, dim in enumerate(op.dims, start=n_value_unique_dims) if dim not in value_dims_set
+    ]
+    return pt_expand_dims(value.values, axis=missing_axis)
```
jessegrabowski, line 118: It seems lke we ought to have a helper that does this on the pytensor side

pymc/dims/distributions/core.py
```diff
@@ -103,38 +105,67 @@ def find_measurable_xtensor_from_tensor(fgraph, node) -> list[XTensorVariable] |
...
+
+
+def _to_xtensor(
+    op: MeasurableXTensorFromTensor, value: XTensorVariable, var: TensorVariable
+) -> XTensorVariable:
+    value_unique_dims = [dim for dim in value.dims if dim not in op.dims]
```
jessegrabowski: Name could be better, because dims are always unique by construction. value_only_dims?

pymc/logprob/transforms.py
ricardoV94: We have this helper in pymc.math anyway, so no need to revert anything? It just starts already with the optimized form

jessegrabowski (approved):

tests/distributions/test_censored.py
ricardoV94: No the previous was just wrong but because it was ignored nothing happened.

pymc/dims/distributions/core.py
ricardoV94: `extra_value_dims` ?

PR #8111, opened 2026-02-18
Use nutpie by default

Jayant-kernel:
# Closes 8079


Jayant-kernel:
@ricardoV94 
I've opened a draft PR for this here:

It implements the auto-selection logic: 
if nutpie  is installed and the model is suitable (all continuous variables), it will be used by default when nuts_sampler=None. It falls back to pymc
 otherwise.

Let me know what you think!

pymc/sampling/mcmc.py
```diff
@@ -84,6 +83,9 @@
 except ImportError:
     MemoryStore = type("MemoryStore", (), {})
 
+def _nutpie_is_installed() -> bool:
```
ricardoV94: nutpie will be a default dependency once we do these changes, no need to check if it's installed

pymc/sampling/mcmc.py
```diff
@@ -822,6 +844,16 @@ def sample(
         )
     )
 
+    if nuts_sampler is None:
+        if (
+            exclusive_nuts
+            and _nutpie_is_installed()
+            and (compile_kwargs is None or not compile_kwargs)
```
ricardoV94: compile kwargs is not enough to rule out nutpie. Only if it implies a pytensor compilation mode that is not numba or jax backend. Something like `nutpie_compatibile_mode = isinstance(get_mode(compile_kwargs.get("mode")).linker, JAXLinker| NumbaLinker)`

tests/sampling/test_nutpie_selection.py
```diff
@@ -0,0 +1,91 @@
+#   Copyright 2024 - present The PyMC Developers
```
ricardoV94: no reason for a new test file, it should be in test_mcmc, or whatever file tests `sample` functionality now

pymc/sampling/mcmc.py
ricardoV94: We also need to propagate this information into the `backend` kwarg for nutpie

Jayant-kernel:
@ricardoV94 T
hanks again for the patience and the detailed review! I've now addressed all the feedback and fixed the CI issues:

1. **Removed `_nutpie_is_installed()`** — `nutpie` is now treated as a default dependency.
2. **Proper Linker Detection** — Implemented the check using `isinstance(get_mode(...).linker, JAXLinker | NumbaLinker)` (using union syntax to satisfy Ruff UP038).
3. **Backend Propagation** — Added logic to set `backend="jax"` or `backend="numba"` for nutpie based on the PyTensor linker.
4. **Moved Tests** — Deleted `test_nutpie_selection.py` and moved all tests into [tests/sampling/test_mcmc.py](cci:7://file:///c:/Users/jayan/Desktop/pymc/tests/sampling/test_mcmc.py:0:0-0:0).
5. **Fixed Pre-commit** — Fixed all trailing whitespace, import sorting, and Ruff formatting issues (v0.11.13).

All CI checks (Docs, Pre-commit, etc.) are now passing (green). Ready for re-review!

twiecki:
@ricardoV94 wdyt, would be nice to get this default on the v6 branch.

ricardoV94:
Closing in favor of #8248

PR #8113, opened 2026-02-18
GP Doc improvements

elc45:
## Description

The GP `Base` class declares several methods to be implemented in the child classes, e.g. `gp.Latent`, `gp.Marginal`, but some classes do not implement these methods, leading to misleading docs. For instance, doc for `gp.Latent` says it has `predict` and `marginal_likelihood` methods, when it does not. I remove these base class instantiations, since they do not add anything as far as I can tell, but just confuse the docs. I also clean up some docstrings, adding hyperlinks to make the module more navigable for users. 

## Checklist

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/gp/gp.py
+++ b/pymc/gp/gp.py
@@ -54,18 +54,6 @@ class Base:
         cov_total = self.cov_func + other.cov_func
         return self.__class__(mean_func=mean_total, cov_func=cov_total)
 
-    def prior(self, name, X, *args, **kwargs):
-        raise NotImplementedError
-
-    def marginal_likelihood(self, name, X, *args, **kwargs):
-        raise NotImplementedError
-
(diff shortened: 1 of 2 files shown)


ricardoV94:
It's not uncommon for a base class to define methods that subclasses can implement optionally. If the issue is with the docs that should be addressed directly. I don't think we should remove them from the base class. Among other things it establishes the API and signature that implementations should follow

elc45:
Sorry for the late response. My first attempt was to address this via Sphinx, but I couldn't get it to work any way I tried. This included adding :meta private: to docstrings and whatnot, but nothing worked. Any tips would be appreciated.

Issue #8106, opened 2026-02-11
Deprecation warning from cachetools v7


ricardoV94:
### Description

Started failing in the CI, I pinned it #8105 (https://github.com/pymc-devs/pymc/commit/c32715af3b430cf7b9272b93370ca19d15024c29)

```
================================== FAILURES ===================================
__________________________ test_fit_with_nans[True] ___________________________

score = True

    @pytest.mark.parametrize("score", [True, False])
    def test_fit_with_nans(score):
        X_mean = pm.floatX(np.linspace(0, 10, 10))
        y = pm.floatX(np.random.normal(X_mean * 4, 0.05))
        with pm.Model():
            inp = pm.Normal("X", X_mean, size=X_mean.shape)
            coef = pm.Normal("b", 4.0)
            mean = inp * coef
            pm.Normal("y", mean, 0.1, observed=y)
            with pytest.raises(FloatingPointError) as e:
>               advi = pm.fit(100, score=score, obj_optimizer=pm.adam(learning_rate=float("nan")))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests\variational\test_inference.py:44: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pymc\variational\inference.py:775: in fit
    return inference.fit(n, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
pymc\variational\inference.py:158: in fit
    step_func = self.objective.step_function(score=score, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
C:\Users\runneradmin\micromamba\envs\pymc-test\Lib\site-packages\pytensor\configparser.py:44: in res
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
pymc\variational\opvi.py:393: in step_function
    updates = self.updates(
pymc\variational\opvi.py:268: in updates
    self.add_obj_updates(
pymc\variational\opvi.py:313: in add_obj_updates
    obj_target = self(
C:\Users\runneradmin\micromamba\envs\pymc-test\Lib\site-packages\pytensor\configparser.py:44: in res
[...]

PR #8115, opened 2026-02-19
cachetools library dependency removal

Dekermanjian:
## Description
This PR addresses the dependency that PyMC has on the `cachetools` library. We use this library very minimally in some VI and testing utilities. I replaced the functionality that was leveraging the `cachetools` library with equivalents built from python standard libraries. 

## Related Issue

- [ ] Closes #
- [x] Related to #8106 

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/conda-envs/environment-alternative-backends.yml
+++ b/conda-envs/environment-alternative-backends.yml
@@ -7,7 +7,6 @@ dependencies:
 # Base dependencies
 - arviz>=0.13.0,<1.0
 - blas
-- cachetools>=4.2.1,<7
 - cloudpickle
 - zarr>=2.5.0,<3
 - numba
--- a/conda-envs/environment-dev.yml
+++ b/conda-envs/environment-dev.yml
@@ -7,7 +7,6 @@ dependencies:
 # Base dependencies
 - arviz>=0.13.0,<1.0
 - blas
-- cachetools>=4.2.1,<7
 - cloudpickle
 - numpy>=1.25.0
 - pandas>=0.24.0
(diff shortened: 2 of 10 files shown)


ricardoV94:
I would try something more bold. Do we need the caching machinery that was being used in VI at all?

Dekermanjian:
> I would try something more bold. Do we need the caching machinery that was being used in VI at all?

I assumed it was helping out with performance. I will run some side by side tests with and without caching and report the results back here. You might be right in that caching is not making much of a difference.

ricardoV94:
Like are these cached methods being used in a hot loop or it's only about saving recompilation time across multiple fits? If they're used in a hot loop, certainly there is a better design that relying on cached methods, but that would be out of scope for a simple PR. 

Note we have a VI redesign in the working in pymc-extras so refactor/compatibilty work may not be worth it

Dekermanjian:
@maahn  @ricardoV94 I looked into removing the caching logic from the variational module, but it’s deeply integrated into the VI architecture, and removing it would introduce errors. I asked @jessegrabowski for help yesterday, but given the new VI architecture, refactoring this part may not be advisable.

jessegrabowski:
I think you wanted to tag @ferrine 

My suggestion is to just switch to the basic caching in this PR so we can clear the dependency error, then open a new issue to explore what to do about the caching itself.

Dekermanjian:
Whoops! Sorry to tag you out of the blue @maahn.

PR #8122, opened 2026-02-26
Fix compute_log_prior argument handling

vkverma9534:
compute_log_prior unintentionally accepted positional arguments, unlike
compute_log_likelihood, which enforces keyword-only usage. This could lead to
silent argument mis-binding and confusing errors when arguments are passed
positionally.

This pr makes compute_log_prior keyword-only for consistency and safety, and
adds a regression test that locks the function’s public API contract.

No behavior change for valid usage.

And I added a Test as well.

### DIFF
--- a/pymc/stats/log_density.py
+++ b/pymc/stats/log_density.py
@@ -74,6 +74,7 @@ def compute_log_likelihood(
 
 def compute_log_prior(
     idata: InferenceData,
+    *,
     var_names: Sequence[str] | None = None,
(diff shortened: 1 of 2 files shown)


ricardoV94:
PR was fine, but no need for a test

vkverma9534:
I apologise @ricardoV94 for pulling this too early because i forgot to use local pre-commit autofix i will delete this branch and raise a pr again sorry again

PR #8136, opened 2026-02-28
Fix logp shape of transformed ZeroSumNormal and dims.ZeroSumNormal with batch sigma

ricardoV94:
Two unrelated issues found together during testing

### DIFF
--- a/pymc/distributions/transforms.py
+++ b/pymc/distributions/transforms.py
@@ -309,7 +309,7 @@ class ZeroSumTransform(Transform):
         return value
 
     def log_jac_det(self, value, *rv_inputs):
-        return pt.constant(0.0)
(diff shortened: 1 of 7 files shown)


pymc/distributions/transforms.py
```diff
@@ -309,7 +309,7 @@ def backward(self, value, *rv_inputs):
         return value
 
     def log_jac_det(self, value, *rv_inputs):
-        return pt.constant(0.0)
+        return value.sum(self.zerosum_axes).zeros_like()
```
ricardoV94, line 312: The logp of a variable gets reduced to the shape of the jacobian to avoid broadcasting / double counting the latter.

juanitorduz (approved):

PR #8154, opened 2026-03-02
ADD: Show remaining time in marimo progress bar

vkverma9534:
While using marimo notebooks, it was hard to tell how much longer a sampling run would take. This change adds a remaining-time (ETA) column to the marimo progress bar to make that clearer. The change is purely visual and does not affect sampling behavior, and a focused test is included to lock in the new output.

### DIFF
--- a/pymc/progress_bar/marimo_progress.py
+++ b/pymc/progress_bar/marimo_progress.py
@@ -16,6 +16,7 @@ from time import perf_counter
 from typing import Any, Self
 
 from pymc.progress_bar.marimo_progress_css import DEFAULT_CSS
+from pymc.progress_bar.utils import compute_remaining_time
 
 
 def format_time(seconds: float) -> str:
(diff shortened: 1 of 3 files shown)


vkverma9534:
@ricardoV94 can you help with any suggestions for the failing ci check

ricardoV94 (commented):
If we want this, it shouldn't be marimo specific.

Global rate isn't the best estimate, since tuning can be slower than after tuning but it contributes to the total elapsed time.

Also for sequential sampling we'll be considering the total elapsed time but per chain draws (I suspect given that rate is itself broken like this)

vkverma9534:
@ricardoV94 Thanks for the feedback I agree that's a good point. I'll refactor the ETA calculation into the shared progress logic so it isn't marimo-specific, and I'll also update the estimation formula and add better test coverage.

vkverma9534:
@ricardoV94 I pushed an update based on my understanding of your suggestion to make the ETA logic less marimo-specific. I added a small utils.py helper, though I'm not sure if introducing a new file was the intended direction happy to adjust if you'd prefer it somewhere else.

PR #8118, opened 2026-02-24
fix update_node_formatters

mullachv:
Fix node_formatters to accept string keys matching NodeType.value in model_to_graphviz and model_to_networkx

## Description
`update_node_formatters` in `pymc/model_graph.py` previously validated `node_formatters` keys only against `NodeType` enum members. The docstrings for `model_to_graphviz` and `model_to_networkx` show examples using string keys such as `"Free Random Variable"` and `"Observed Random Variable"` (i.e. `NodeType.value`), which then raised a `ValueError`.

This change normalizes user-provided keys so that both are accepted:
- String keys that match a `NodeType.value` (e.g. `"Free Random Variable"`) are mapped to the corresponding `NodeType` before merging with defaults and validation.
- `NodeType` enum keys continue to work as before.

Invalid keys still raise a clear `ValueError` listing the invalid keys and the allowed values. The error message was updated to mention that keys may be either `NodeType` or a valid `NodeType` value string.

## Related Issue

- [x] Closes #8091
- [ ] Related to #

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [x] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation - no new docs; behavior was fixed to match existing doc examples
- [ ] If you are a pro: each commit corresponds to a [relevant logical change]

## Type of change
- [ ] New feature / enhancement
- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

PR #8160, opened 2026-03-04
Fix mypy in zarr.py: handle StatShape int|None in np.arange

mullachv:
## Description
Fixes mypy CI failure in `pymc/backends/zarr.py` (line 686). 

## Related Issue
- [ ] Closes #
- [x] Related to #8118

## Checklist

## Type of change
- [ ] Bug fix
- [ ] Documentation
- [x] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/backends/zarr.py
+++ b/pymc/backends/zarr.py
@@ -683,7 +683,8 @@ class ZarrTrace:
                 for i, shape_i in enumerate(shape):
                     dim = f"{name}_dim_{i}"
                     dims.append(dim)
-                    group_coords[dim] = np.arange(shape_i, dtype="int")
+                    size = shape_i if shape_i is not None else 0
+                    group_coords[dim] = np.arange(size, dtype="int")
             dims = ("chain", "draw", *dims)
             attrs = extra_var_attrs[name] if extra_var_attrs is not None else {}
             attrs.update({"_ARRAY_DIMENSIONS": dims})


pymc/backends/zarr.py
```diff
@@ -683,7 +683,8 @@ def init_group_with_empty(
                 for i, shape_i in enumerate(shape):
                     dim = f"{name}_dim_{i}"
                     dims.append(dim)
-                    group_coords[dim] = np.arange(shape_i, dtype="int")
+                    size = shape_i if shape_i is not None else 0
```
ricardoV94, line 686: why zero?

avm19: Is it true that `shape_i` can never be None for sampled variables in "posterior" and "unconstrained_posterior" groups, and it can only be None in "sample_stats" (depending on sampler steps used)? In which case it denotes a variable-size dimension?

P.S. Afaiu, neither Zarr nor Xarray can handle non-rectangular data. In addition, Xarray.open_zarr() verifies that dimensions match lengths of corresponding coordinate arrays, so there is no point in providing a zero-length coordinate. I believe it would be safe to raise an informative error here if `shape_i is None`. Neither of tests in tests/backends/test_zarr.py cover this possibility.

ricardoV94: Sounds good (deleted my previous comment)

ricardoV94:
Fixed in https://github.com/pymc-devs/pymc/pull/8015

PR #8170, opened 2026-03-05
Better `str_repr` for models with dims variables

ricardoV94:
Depends on https://github.com/pymc-devs/pytensor/pull/1937


ricardoV94:
Failing is known and unrelated

jessegrabowski (approved):

PR #8174, opened 2026-03-06
Fix broadcast check on log_jac_det

### LINKED ISSUE #8173: BUG: freeze_dims_and_data breaks HalfStudentT logp when a dimension has size 1
### Describe the issue:

When freeze_dims_and_data is applied to a model containing a HalfStudentT RV with a size-1 dims coordinate, the subsequent `model.logp()` call raises a `ValueError`.

```python
import pymc as pm
from pymc.model.transform.optimization import freeze_dims_and_data

with pm.Model(coords={'x_dim': ['only_one']}) as model:
    x = pm.HalfStudentT('x', nu=7, sigma=1, dims='x_dim')

fmodel = freeze_dims_and_data(model)
fmodel.logp()
```

### Error message:

```shell
  File "/home/velochy/salk/sandbox/debug/Rohetiiger3_elekter_clean_citizen_imputed_package/tmp_fa_test_pp_nps/test_st.py", line 8, in <module>
    fmodel.logp()  # raises ValueError
    ^^^^^^^^^^^^^
  File "/home/velochy/miniconda3/envs/salk/lib/python3.12/site-packages/pymc/model/core.py", line 711, in logp
    rv_logps = transformed_conditional_logp(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/velochy/miniconda3/envs/salk/lib/python3.12/site-packages/pymc/logprob/basic.py", line 642, in transformed_conditional_logp
    temp_logp_terms = conditional_logp(
                      ^^^^^^^^^^^^^^^^^
  File "/home/velochy/miniconda3/envs/salk/lib/python3.12/site-packages/pymc/logprob/basic.py", line 572, in conditional_logp
    node_logprobs = _logprob(
                    ^^^^^^^^^
  File "/home/velochy/miniconda3/envs/salk/lib/python3.12/functools.py", line 909, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/velochy/miniconda3/envs/salk/lib/python3.12/site-packages/pymc/logprob/transform_value.py", line 116, in transformed_value_logprob
    raise ValueError(
ValueError: The logp of halfstudentt{inline=True} and log_jac_det of LogTransform are not allowed to broadcast together. There is a bug in the implementation of either one.
```

### PyMC version information:

PyMC 5.28.0
pytensor 2.38.1
Python 3.12

### Context for the issue:

After freezing, the size-1 dimension is replaced by a constant that PyTensor marks as broadcastable. This makes the logp of HalfStudentT return a tensor with broadcastable=(True,), while LogTransform.log_jac_det (computed from the forward value, not the frozen size) retains broadcastable=(False,). The mismatch triggers the validation check in transform_value.py.

The same pattern does not affect HalfNormal or Exponential (which use the same LogTransform), nor does it affect HalfStudentT with 2+ categories. It also does not occur on the unfrozen model — only after freeze_dims_and_data.

velochy:
## Description
Fix for https://github.com/pymc-devs/pymc/issues/8173

## Related Issue

- [X] Closes #8173

## Checklist

## Type of change

- [X] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/logprob/transform_value.py
+++ b/pymc/logprob/transform_value.py
@@ -16,6 +16,7 @@ import warnings
 from collections.abc import Sequence
 
 import numpy as np
+import pytensor.tensor as pt
 
 from pytensor.graph import Apply, Op
 from pytensor.graph.features import AlreadyThere, Feature
@@ -113,10 +114,18 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
             )
         # Check there is no broadcasting between logp and jacobian
         if logp.type.broadcastable != log_jac_det.type.broadcastable:
-            raise ValueError(
-                f"The logp of {rv_op} and log_jac_det of {transform} are not allowed to broadcast together. "
-                "There is a bug in the implementation of either one."
-            )
+            lb, jb = logp.type.broadcastable, log_jac_det.type.broadcastable
+            broadcastable_axes = [
+                i for i, (ai, bi) in enumerate(zip(lb, jb, strict=True)) if ai or bi
+            ]
+            try:
+                logp = pt.specify_broadcastable(logp, *broadcastable_axes)
+                log_jac_det = pt.specify_broadcastable(log_jac_det, *broadcastable_axes)
+            except ValueError as err:
+                raise ValueError(
+                    f"The logp of {rv_op} and log_jac_det of {transform} are not allowed to broadcast together. "
+                    "There is a bug in the implementation of either one."
+                ) from err
 
         if use_jacobian:
             if value.name:
--- a/tests/distributions/test_transform.py
+++ b/tests/distributions/test_transform.py
@@ -657,12 +657,12 @@ def test_invalid_jacobian_broadcast_raises():
 
     buggy_transform = BuggyTransform()
 
-    with pm.Model() as m:
-        pm.Uniform("x", shape=(4, 3), default_transform=buggy_transform)
+    import numpy as np
 
     for jacobian_val in (True, False):
-        with pytest.raises(
-            ValueError,
-            match="are not allowed to broadcast together. There is a bug in the implementation of either one",
-        ):
-            m.logp(jacobian=jacobian_val)
+        with pm.Model() as m:
+            pm.Uniform("x", shape=(4, 3), default_transform=buggy_transform)
+
+        logp_fn = m.compile_logp(jacobian=jacobian_val)
+        with pytest.raises(AssertionError, match="SpecifyShape"):
+            logp_fn({"x_buggy__": np.zeros((4, 3))})
--- a/tests/logprob/test_transform_value.py
+++ b/tests/logprob/test_transform_value.py
@@ -578,6 +578,20 @@ def test_scan_transform():
     np.testing.assert_allclose(logp_fn(**test_point), ref_logp_fn(**test_point))
 
 
+def test_halfstudent_t_with_frozen_dims():
+    """Regression test: log_jac_det had mismatched broadcastable dims vs logp when
+    dims were frozen to a single-element coordinate, causing a ValueError.
+    """
+    from pymc.model.transform.optimization import freeze_dims_and_data
+
+    with pm.Model(coords={"x_dim": ["only_one"]}) as model:
+        pm.HalfStudentT("x", nu=7, sigma=1, dims="x_dim")
+
+    fmodel = freeze_dims_and_data(model)
+    [x_logp] = fmodel.logp(sum=False)
+    assert x_logp.type.shape == (1,)
+
+
 def test_weakref_leak():
     """Check that the rewrite does not have a growing memory footprint.
 


pymc/logprob/transform_value.py
```diff
@@ -111,7 +112,22 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
             raise NotImplementedError(
                 f"Univariate transform {transform} cannot be applied to multivariate {rv_op}"
             )
-        # Check there is no broadcasting between logp and jacobian
+        # Check there is no broadcasting between logp and jacobian.
```
ricardoV94: comment is too verbose and specific

pymc/logprob/transform_value.py
```diff
@@ -111,7 +112,22 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
...
+        # If logp has broadcastable=True on axes where log_jac_det has False, those axes
+        # must be size-1 at runtime (e.g. after freeze_dims_and_data on a SymbolicRandomVariable
+        # whose scalar params don't inherit the frozen broadcastable size annotation).
+        # Annotate log_jac_det to match so the addition is valid.
+        if logp.type.broadcastable != log_jac_det.type.broadcastable:
```
ricardoV94, line 116: do we want to apply the specify both directions? If log_jac_det.type.shape==(None, 1) and log.shape.type.shape==(1, None), then both should have shape=(1, 1) to be valid?

The call can be wrapped in a try/except to raise the old more informative error in the case it is known to broadcast like (5, 1), vs (1, 5), which specify_broadcastable will raise immediately (IIRC)

pymc/logprob/transform_value.py
velochy: My bad. AIs tend to be verbose with these...

pymc/logprob/transform_value.py
velochy: Makes sense. Changed it to specify on both, and added the try-catch

velochy:
@ricardoV94 failing tests look pre-existing. Should I rebase on a branch other than main?

pymc/logprob/transform_value.py
```diff
@@ -111,17 +112,28 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
...
+        # Axes where one side is broadcastable and the other isn't must be size-1
+        # at runtime (e.g. after freeze_dims_and_data on a SymbolicRandomVariable).
+        # Apply specify_broadcastable in both directions so the addition is valid.
         if logp.type.broadcastable != log_jac_det.type.broadcastable:
-            raise ValueError(
-                f"The logp of {rv_op} and log_jac_det of {transform} are not allowed to broadcast together. "
-                "There is a bug in the implementation of either one."
-            )
+
+            def broadcastable_axes(a, b):
```
ricardoV94: can be simplified. 
`broadcastable_axes = [i for i, (ai, bi) in enumerate(zip(...)) if ai or bi]`

And then use in both.

There's no extra cost of specifying a broadcastable_axes that is already known to be broadcastable.

Only thing to make sure is that ndim matches

pymc/logprob/transform_value.py
```diff
@@ -111,17 +112,28 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
...
+
+            lb, jb = logp.type.broadcastable, log_jac_det.type.broadcastable
+            logp = pt.specify_broadcastable(logp, *broadcastable_axes(jb, lb))
+            log_jac_det = pt.specify_broadcastable(log_jac_det, *broadcastable_axes(lb, jb))
 
         if use_jacobian:
             if value.name:
                 log_jac_det.name = f"{value.name}_jacobian"
-            logprobs_jac.append(logp + log_jac_det)
+            try:
```
ricardoV94, line 121: the try/except should be in the `specify_broadcastable`

ricardoV94:
> @ricardoV94 failing tests look pre-existing. Should I rebase on a branch other than main?

I think main is still failing, it's addressed by a PR that's still open. We can ignore

pymc/logprob/transform_value.py
velochy: not sure I understand the ndim question here. Is it possible they have different numbers of dimensions? If so, how should they be matched?

ricardoV94: We may already have logic above that handles different ndim. 

If not we should fail, it's a guaranteed form of broadcasting. But the current checks may miss on it

velochy: So I just add a length equality check and throw an exception if there is a mismatch?
Or no action needed here?

pymc/logprob/transform_value.py
```diff
@@ -111,12 +112,20 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
...
+            lb, jb = logp.type.broadcastable, log_jac_det.type.broadcastable
+            broadcastable_axes = [i for i, (ai, bi) in enumerate(zip(lb, jb)) if ai or bi]
+            try:
+                logp = pt.specify_broadcastable(logp, *broadcastable_axes)
+                log_jac_det = pt.specify_broadcastable(log_jac_det, *broadcastable_axes)
+            except ValueError:
+                raise ValueError(
+                    f"The logp of {rv_op} and log_jac_det of {transform} are not allowed to broadcast together. "
+                    "There is a bug in the implementation of either one."
+                )
```
```suggestion
            except ValueError as err:
                raise ValueError(
                    f"The logp of {rv_op} and log_jac_det of {transform} are not allowed to broadcast together. "
                    "There is a bug in the implementation of either one."
                ) from err
```

pymc/logprob/transform_value.py
```diff
@@ -111,12 +112,20 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
             raise NotImplementedError(
                 f"Univariate transform {transform} cannot be applied to multivariate {rv_op}"
             )
-        # Check there is no broadcasting between logp and jacobian
+        # Axes where one side is broadcastable and the other isn't must be size-1
```
ricardoV94: Still prefer the original comment

pymc/logprob/transform_value.py
ricardoV94: Nothing to do, it's handled above with if log_jac_det.ndim < logp.ndim and elif branches above. Just put a `strict=True` in the zip in case we modify those and forget to handle it here

pymc/logprob/transform_value.py
```diff
@@ -111,12 +112,20 @@ def transformed_value_logprob(op, values, *rv_outs, use_jacobian=True, **kwargs)
             raise NotImplementedError(
                 f"Univariate transform {transform} cannot be applied to multivariate {rv_op}"
             )
-        # Check there is no broadcasting between logp and jacobian
+        # Check there is no broadcasting difference between logp and jacobian
```
ricardoV94: I don't want to hold us on this comment, but broadcasting is an action between two variables, not a property of each variable variables (hence why it's called type.broadcastable). We are checking/enforcing that the variables won't broadcast together. 

So original comment is still more accurate than current one

ricardoV94 (commented):
looks good, other than nit about comment.

More importantly, can you add a regression test? Your thing with frozen length 1 dim is fine, just calling model.logp(sum=False) and checking it comes back with the right static shape (and implicitly that it didn't raise)

pymc/logprob/transform_value.py
velochy: noted. Kept the original comment unchanged now.

ricardoV94 (approved):

ricardoV94:
Looks like this test needa updating FAILED tests/distributions/test_transform.py::test_invalid_jacobian_broadcast_raises - Failed: DID NOT RAISE <class 'ValueError'>

velochy:
Error moved from compile time to run time, so changed the test to compile and then call the logp instead. Hope that works for you @ricardoV94

ricardoV94:
Thanks @velochy. There's another PR I want to get merged and then I'll cut a release

velochy:
Much appreciated, and thank you again @ricardoV94

PR #8172, opened 2026-03-06
Fix DEMetropolis stats type mismatch in population sampler output

SanketMeghale:
## Description
This PR fixes a **stats contract violation** in `DEMetropolis` that can break population sampling output handling and trigger CI failures like:

- `ValueError: setting an array element with a sequence`

Root cause:
- `DEMetropolis.stats_dtypes_shapes` declares scalar stats (`tune: bool`, `scaling: float64`),
- but `DEMetropolis.astep` was emitting:
  - `tune` as `str | None`
  - `scaling` as a NumPy array

This mismatch can propagate into downstream stats packing/trace handling and fail tests in `tests/sampling/test_population.py`.

### Fix
In `DEMetropolis.astep` stats payload:
- `"tune": self.tune is not None` (bool)
- `"scaling": float(np.mean(self.scaling))` (scalar float)

This restores consistency with `stats_dtypes_shapes` and aligns behavior with `DEMetropolisZ` (which already reports scalar `scaling`).

Fixes #8169

## Testing
Targeted population sampler checks:

```powershell
$env:PYTENSOR_FLAGS='cxx='
.\.venv\Scripts\python.exe -m pytest tests/sampling/test_population.py -k "checks_population_size or demcmc_warning_on_small_populations or nonparallelized_chains_are_random or parallelized_chains_are_random" -q
```

Observed result relevant to this PR:
- The original `ValueError: setting an array element with a sequence` is no longer reproduced.
- `test_checks_population_size`: pass
- `test_demcmc_warning_on_small_populations`: pass

Issue #8169, opened 2026-03-05
`TestPopulationSamplers` started failing in the CI


ricardoV94:
### Description

```
FAILED tests/sampling/test_population.py::TestPopulationSamplers::test_checks_population_size - ValueError: setting an array element with a sequence.
FAILED tests/sampling/test_population.py::TestPopulationSamplers::test_demcmc_warning_on_small_populations - ValueError: setting an array element with a sequence.
FAILED tests/sampling/test_population.py::TestPopulationSamplers::test_nonparallelized_chains_are_random - ValueError: setting an array element with a sequence.
FAILED tests/sampling/test_population.py::TestPopulationSamplers::test_parallelized_chains_are_random - ValueError: setting an array element with a sequence.
```

SanketMeghale:
I can work on this. I’ll submit a focused PR for the DEMetropolis stats type mismatch causing these TestPopulationSamplers failures.

aniketpandey05:
@SanketMeghale  Thanks for the detailed write-up! The root cause analysis and fix look correct. One observation: the test command uses -k to filter all 4 failing tests from issue #8169, but the observed results only mention 2: could you update the description to confirm that all 4 pass

PR #7871, opened 2025-07-24
Fix `numpy` deprecation warning when running tests in `test_metropolis.py`

asifzubair:
## Description
This PR is an attempt to address the `numpy` deprecation warning (#7369). Previous attempts to make the fix (ex., #7514) focussed on checking for scalars in `pymc/backends/ndarray.py`. Following this [hint](https://github.com/pymc-devs/pymc/pull/7514#issuecomment-2382409866), the present PR instead checks which offending stat variable is not respecting the scalar requirement and modifies that.

After the fix, we can see that no warnings are seen:
```sh
(pymc-dev) jovyan@75886b6f8645:~/work$ python -m pytest tests/step_methods/test_metropolis.py 
=== test session starts ===
 ...

tests/step_methods/test_metropolis.py .............................................                                               [100%]

==== 45 passed in 259.02s (0:04:19) ===
```

## Related Issue
- [ ] Closes #7369 
- [ ] Related to #

## Checklist

- [x] Checked that [the pre-commit linting/style checks pass](https://docs.pymc.io/en/latest/contributing/python_style.html)
- [ ] Included tests that prove the fix is effective or that the new feature works
- [ ] Added necessary documentation (docstrings and/or example notebooks)
- [ ] If you are a pro: each commit corresponds to a [relevant logical change](https://wiki.openstack.org/wiki/GitCommitMessages#Structural_split_of_changes)

## Type of change

- [ ] New feature / enhancement
- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

----
📚 Documentation preview 📚: https://pymc--7871.org.readthedocs.build/en/7871/

PR #8178, opened 2026-03-08
Emit scalar `DEMetropolis` "scaling" stat

michaelosthege:
Replaces #8172 by doing the same fix already done in #7871, closing #8169.

~Follow-up to https://github.com/pymc-devs/pymc/pull/8015 that tracks warmup information independent of stats on `IBaseTrace`/`NDarray` backends:~

* ~`IBaseTrace.warmup_count()` for counting the total number of warmup steps~
* ~`_sample_return` switches  to using `warmup_count()` instead of relying on `tune` parameter. This works also when chains do not have the same number of warmup iterations.~

not worth it I guess

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/step_methods/metropolis.py
+++ b/pymc/step_methods/metropolis.py
@@ -993,7 +993,7 @@ class DEMetropolis(PopulationArrayStepShared):
         self.steps_until_tune -= 1
 
         stats = {
-            "scaling": self.scaling,
+            "scaling": np.mean(self.scaling),
             "lambda": self.lamb,
             "accept": np.exp(accept),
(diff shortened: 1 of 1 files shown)


ricardoV94:
@michaelosthege we have two branches going on now, v6 (default) and main (bugfixes). For this one let's target main still and then I'll cherry pick it to v6. It will also get rid of the extra commits it's picking

ricardoV94 (approved):

PR #8183, opened 2026-03-10
fix webbrowser path for `make view`

### LINKED ISSUE #8182: DOC: `make view` opens blank page

When running `make view` after `make html`, the default browser (Chrome, in my case) opens to a blank page instead of the correct homepage. Opening a one-line PR for this momentarily.

_No response_

elc45:
## Related Issue

- [x] Closes #8182 

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):


docs/source/api/dims/transforms.rst
```diff
@@ -2,7 +2,7 @@
 Distribution Transforms
 ***********************
 
-.. currentmodule:: pymc.dims.transforms
+.. currentmodule:: pymc.dims.distributions.transforms
```
OriolAbril, line 5: this is probably more for @ricardoV94. Do we want users to import dim aware transforms from `pymc.dims.distributions.transforms`? or from `pymc.dims.transforms`? If we want the second option and it doesn't currently work we'd need to fix the respective `__init__.py`, not change this.

docs/source/conf.py
```diff
@@ -296,6 +296,8 @@
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = False
 
+suppress_warnings = ["toc.not_included"]
```
OriolAbril: I think supressing this warning is a very bad idea. Pages that trigger this warning are not reachable through the nabvar+left sidebar so the only way to get to them is through cross-references from other pages. It is ok to do this on purpose for selected pages in which case it needs to be indicated with the [`:orphan:` frontmatter](https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html#special-metadata-fields) which will exclude that page from this toctree check.

docs/source/conf.py
OriolAbril:
```suggestion
    "pmx": ("https://www.pymc.io/projects/extras/en/latest", None),
```

docs/source/learn.md
```diff
@@ -28,6 +28,6 @@ glossary
   - Book: [Bayesian Modeling and Computation in Python](https://bayesiancomputationbook.com/welcome.html)
 
 ### Advanced
-  - {octicon}`plug;1em;sd-text-info` Experimental and cutting edge functionality: {doc}`pmx:index` library
+  - {octicon}`plug;1em;sd-text-info` Experimental and cutting edge functionality: [pymc-experimental](https://github.com/pymc-devs/pymc-experimental) library
```
OriolAbril: I added a comment on conf.py to fix the intersphinx mapping for pymc-extras (it was renamed from experimental to extras quite a while ago). This change should be reverted.

docs/source/conf.py
```diff
@@ -296,6 +296,8 @@
...
+
 # intersphinx configuration to ease linking arviz docs
 intersphinx_mapping = {
     "arviz": ("https://python.arviz.org/en/latest/", None),
```
OriolAbril:
```suggestion
    "arviz": ("https://python.arviz.org/", "https://python.arviz.org/en/stable/objects.inv"),
```

I forgot to do this with the 1.0 release.

pymc/backends/arviz.py
```diff
@@ -509,7 +509,7 @@ def to_inference_data(
     All three of them are optional arguments, but at least one of ``trace``,
     ``prior`` and ``posterior_predictive`` must be present.
     For a usage example read the
-    :ref:`Creating InferenceData section on from_pymc <creating_InferenceData>`
+    `InferenceData schema specification <https://python.arviz.org/en/latest/schema/schema.html>`_
```
OriolAbril: I could continue using a cross reference: `` :ref:`InferenceData schema specification <arviz:schema>` ``

OriolAbril (commented):
this goes way beyond the PR title but most of the changes are positive even if smallish. I have added some comments that need to be addressed. Some are a couple changes I think are negative and a few others are things I realized while reviewing and either directly or partially related

docs/source/api/dims/transforms.rst
ricardoV94: The more verbose should be fine

OriolAbril: let's leave it as is then

OriolAbril (approved):

elc45:
@ricardoV94 any chance you could rerun the random failing tests?

ricardoV94:
docs can't possible affect these tests so don't worry. the prv should change the target branch to v6 though

PR #8184, opened 2026-03-10
Clean up marimo progress bar

fonnesbeck:
## Description

Fix progbar typo: extra "s" was being added to "draws"

## Checklist

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/progress_bar/rich_progress.py
+++ b/pymc/progress_bar/rich_progress.py
@@ -201,7 +201,7 @@ class RichProgressBackend:
         columns += [
             TextColumn(
                 "{task.fields[sampling_speed]:0.2f} {task.fields[speed_unit]}",
-                table_column=Column("Sampling Speed", ratio=1),
+                table_column=Column("Speed", ratio=1),
             ),
             TimeElapsedColumn(table_column=Column("Elapsed", ratio=1)),
             TimeRemainingColumn(table_column=Column("Remaining", ratio=1)),
--- a/pymc/progress_bar/rich_progress.py
+++ b/pymc/progress_bar/rich_progress.py
@@ -186,14 +186,12 @@ class RichProgressBackend:
         progress_columns: list,
         theme: Theme,
     ) -> CustomProgress:
-        columns: list[ProgressColumn] = []
-        if self.step_name.lower() in self.progress_stats:
-            columns.append(
-                TextColumn(
-                    "{" + f"task.fields[{self.step_name.lower()}]" + "}",
-                    table_column=Column(self.step_name.title(), ratio=1),
-                )
+        columns: list[ProgressColumn] = [
+            TextColumn(
+                "{" + f"task.fields[{self.step_name.lower()}]" + "}",
+                table_column=Column(self.step_name.title(), ratio=1),
             )
+        ]
 
         if self.full_stats:
             columns += progress_columns
--- a/pymc/progress_bar/rich_progress.py
+++ b/pymc/progress_bar/rich_progress.py
@@ -186,12 +186,14 @@ class RichProgressBackend:
         progress_columns: list,
         theme: Theme,
     ) -> CustomProgress:
-        columns: list[ProgressColumn] = [
-            TextColumn(
-                "{" + f"task.fields[{self.step_name.lower()}]" + "}",
-                table_column=Column(self.step_name.title(), ratio=1),
+        columns: list[ProgressColumn] = []
+        if self.step_name.lower() in self.progress_stats:
+            columns.append(
+                TextColumn(
+                    "{" + f"task.fields[{self.step_name.lower()}]" + "}",
+                    table_column=Column(self.step_name.title(), ratio=1),
+                )
             )
-        ]
 
         if self.full_stats:
             columns += progress_columns
--- a/pymc/progress_bar/rich_progress.py
+++ b/pymc/progress_bar/rich_progress.py
@@ -302,9 +302,9 @@ class RichProgressBackend:
         action = self.step_name.lower()
         speed = completed / max(elapsed, 1e-6)
         if speed > 1 or speed == 0:
-            unit = f"{action}s/s"
+            unit = f"{action}/s"
         else:
-            unit = f"s/{action}"
+            unit = f"s/{action.rstrip('s')}"
             speed = 1 / speed
 
         self._progress.update(
--- a/pymc/progress_bar/rich_progress.py
+++ b/pymc/progress_bar/rich_progress.py
@@ -302,9 +302,9 @@ class RichProgressBackend:
         action = self.step_name.lower()
         speed = completed / max(elapsed, 1e-6)
         if speed > 1 or speed == 0:
-            unit = f"{action}/s"
+            unit = f"{action}s/s"
         else:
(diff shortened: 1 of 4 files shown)


jessegrabowski:
I think the actual problem is that the code assumes `self.step_name` will be singular, but [here](https://github.com/pymc-devs/pymc/blob/d3ec5c7ef107d49f88c85f9ea8b7bb395d5a9949/pymc/progress_bar/progress.py#L217) it is defined as plural. What if we just remove the s there?

pymc/progress_bar/rich_progress.py
```diff
@@ -290,9 +290,9 @@ def update(
         action = self.step_name.lower()
         speed = completed / max(elapsed, 1e-6)
         if speed > 1 or speed == 0:
-            unit = f"{action}s/s"
+            unit = f"{action}/s"
         else:
-            unit = f"s/{action}"
+            unit = f"s/{action.rstrip('s')}"
```
ricardoV94: why the strip?

fonnesbeck: If there's an "s" there it will remove it. I don't think we pass any actions ending in "s" that are not plural.

fonnesbeck:
@jessegrabowski as long as we are comfortable with the header in the sampler stats table using "draw" instead of "draws" as well.

jessegrabowski:
> @jessegrabowski as long as we are comfortable with the header in the sampler stats table using "draw" instead of "draws" as well.

We could have a standard for that as well? step_name stores the stem, so we add +ing for the name, +s when needed, or whatever else?

ricardoV94:
`tuneing` and `tunes`? I think it's too cute. 

If there's a need for two names, just specify the two? We have just two cases now anyway (mcmc/smc) so not any meaningful amount of work

fonnesbeck:
I think "draw" and "stage" are currently the only values used for `action`. Is there anything other than MCMC and SMC that uses this particular progbar?

ricardoV94:
> I think "draw" and "stage" are currently the only values used for `action`. Is there anything other than MCMC and SMC that uses this particular progbar?

No

fonnesbeck:
Shape errors in DEMetropolis can't be related to these changes.

ricardoV94:
Have to cherry pick some patches from main to v6, give me a sec

ricardoV94:
@fonnesbeck do you want this to go to main for a patch-release? If so change the target of the branch

fonnesbeck:
Probably should, yes.

ricardoV94:
@fonnesbeck I changed the branch and rebased

fonnesbeck:
Anything else we need to do here?

ricardoV94:
Can you share a print screen of before and after of both marimo and rich progress bar (or at least after)

fonnesbeck:
Pixi before:

Screenshot From 2026-03-15 09-39-51

Pixi after:

Screenshot From 2026-03-15 09-26-46

Jupyter before:

Screenshot From 2026-03-15 09-40-56

Jupyter after:

Screenshot From 2026-03-15 09-27-26

ricardoV94:
did you swap jupyter before and after?

fonnesbeck:
Fixed

fonnesbeck:
Should try and get rid of the n-1 artefact in Jupyter though.

fonnesbeck:
Rebased against main after #8201 merge

ricardoV94:
failing test is fixed in v6 but not here, nothing to worry about. Can you post the before/after screenshots?

fonnesbeck:
After:
Screenshot From 2026-03-19 16-25-18
Screenshot From 2026-03-19 16-24-34

jessegrabowski:
Is it weird that the Accept column in Marimo has 0.000 and 1.000?

fonnesbeck:
Before:
Screenshot From 2026-03-19 16-50-09
Screenshot From 2026-03-19 16-50-58

jessegrabowski:
I guess i was asking what the accept column meant

ricardoV94:
Seems like you have more columns than before, was that intended?

ricardoV94:
Also sampling speed erased at the end in rich? would be nice to leave

fonnesbeck:
Yes, the accept is odd. Not a fair comparison, I guess because I've showed different samplers before and after.

fonnesbeck:
Updated afters:
Screenshot From 2026-03-19 21-57-16
Screenshot From 2026-03-19 21-55-32

ricardoV94:
elapsed is zeroed out?

ricardoV94:
Also maybe out of scope, but sampling speed -> speed in rich progressbars?

fonnesbeck:
> elapsed is zeroed out?

It is not. The sampling took less than a second. Sampling speed was over 3k/second for only a few iterations. It does appear to be rounding, however.

fonnesbeck:
After:
Screenshot From 2026-03-20 07-41-13
Screenshot From 2026-03-20 07-43-14

I ran these longer so that the elapsed time could be seen.

fonnesbeck:
Do we want to have perfect correspondence between the progress bars? i.e., Div vs. Divergences, Step vs. Step size, Tree vs. Grad evals. I like the compactness of Div and Tree.

jessegrabowski:
Yes I think so. My vision for the design was that the steps owned the column names, so the progress backend we put behind them shouldn't matter. The fact that these are out of sync means it's not the case.

I don't have a strong view on which name to use. The shorter ones are nice because they're short, but new users might be confused about what they mean. I like something about gradient evals over tree or depth.

ricardoV94:
Is it still showing grad evals?

Otherwise looks good

fonnesbeck:
OK, columns correspond now.
image
Screenshot From 2026-03-20 07-43-14

pymc/progress_bar/rich_progress.py
```diff
@@ -186,20 +186,22 @@ def _create_progress_bar(
         progress_columns: list,
         theme: Theme,
     ) -> CustomProgress:
-        columns: list[ProgressColumn] = [
-            TextColumn(
-                "{" + f"task.fields[{self.step_name.lower()}]" + "}",
-                table_column=Column(self.step_name.title(), ratio=1),
+        columns: list[ProgressColumn] = []
+        if self.step_name.lower() in self.progress_stats:
```
ricardoV94: what does this new if fix?

fonnesbeck: It filters out columns from the progress_stats table when the backend does not have data for it. Otherwise Rich would unconditionally create a `TextColumn` referencing task.fields[foobar] that does not exist, which would raise a KeyError.

ricardoV94: Why does that mismatch happen and how comes it didn't raise before?

fonnesbeck: I don't think it will happen currently. Just prevents failure if a key is missing. Ditch it?

ricardoV94: yes, unless you have reason to believe it could happen

ricardoV94 (approved):

PR #8185, opened 2026-03-10
DOC: Clarify jacobian parameter and overall docstring for Model.logp and similar

avm19:
## Description

Improve docstrings for `Model.logp()` and related methods (`dlogp()`, `d2logp()`, `compile_*()`) and clarify what parameter `jacobian` does.

## Related Issue

- [x] Closes #8114 

## Checklist

## Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/docs/source/api/distributions/transforms.rst
+++ b/docs/source/api/distributions/transforms.rst
@@ -2,7 +2,7 @@
 Transformations
 ***************
 
-.. currentmodule:: pymc.distributions.transforms
+.. module:: pymc.distributions.transforms
 
 While many distributions are defined on constrained spaces (e.g. intervals), MCMC samplers typically perform best when sampling on the unconstrained real line; this is especially true of HMC samplers. PyMC balances this through the use of transforms. A transform instance can be passed to the constructor of a random variable to tell the sampler how to move between the underlying unconstrained space where the samples are actually drawn and the transformed space constituting the support of the random variable. Transforms are not currently implemented for discrete random variables.
 
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -608,11 +608,11 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         See Also
         --------
-        :py:meth:`~pymc.model.core.Model.logp` :
+        :py:meth:`logp` :
             log-probability density as a Variable (in a symbolic form).
-        :py:meth:`~pymc.model.core.Model.compile_dlogp` :
+        :py:meth:`compile_dlogp` :
             gradient of log-probability density as a compiled function.
-        :py:meth:`~pymc.model.core.Model.compile_d2logp` :
+        :py:meth:`compile_d2logp` :
             Hessian of log-probability density as a compiled function.
         """
         compile_kwargs.setdefault("on_unused_input", "ignore")
@@ -650,11 +650,11 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         See Also
         --------
-        :py:meth:`~pymc.model.core.Model.dlogp` :
+        :py:meth:`dlogp` :
             gradient of log-probability density as a Variable (in a symbolic form).
-        :py:meth:`~pymc.model.core.Model.compile_logp` :
+        :py:meth:`compile_logp` :
             log-probability density as a compiled function.
-        :py:meth:`~pymc.model.core.Model.compile_d2logp` :
+        :py:meth:`compile_d2logp` :
             Hessian of log-probability density as a compiled function.
         """
         compile_kwargs.setdefault("on_unused_input", "ignore")
@@ -695,11 +695,11 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         See Also
         --------
-        :py:meth:`~pymc.model.core.Model.d2logp` :
+        :py:meth:`d2logp` :
             Hessian of log-probability density as a Variable (in a symbolic form).
-        :py:meth:`~pymc.model.core.Model.compile_logp` :
+        :py:meth:`compile_logp` :
             log-probability density as a compiled function.
-        :py:meth:`~pymc.model.core.Model.compile_dlogp` :
+        :py:meth:`compile_dlogp` :
             gradient of log-probability density as a compiled function.
         """
         compile_kwargs.setdefault("on_unused_input", "ignore")
@@ -736,24 +736,24 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         See Also
         --------
-        :py:meth:`~pymc.model.core.Model.compile_logp` :
+        :py:meth:`compile_logp` :
             log-probability density as a compiled function.
-        :py:meth:`~pymc.model.core.Model.dlogp` :
+        :py:meth:`dlogp` :
             gradient of log-probability density as a Variable (in a symbolic form).
-        :py:meth:`~pymc.model.core.Model.d2logp` :
+        :py:meth:`d2logp` :
             Hessian of log-probability density as a Variable (in a symbolic form).
-        :py:meth:`~pymc.model.core.Model.logp_dlogp_function` :
+        :py:meth:`logp_dlogp_function` :
             compile logp and its gradient as a single function.
-        :py:attr:`~pymc.model.core.Model.varlogp` :
+        :py:attr:`varlogp` :
             convenience property for logp of all free (unobserved) RVs.
-        :py:attr:`~pymc.model.core.Model.varlogp_nojac` :
+        :py:attr:`varlogp_nojac` :
             convenience property for logp of all free (unobserved) RVs without transformation
             corrections.
-        :py:attr:`~pymc.model.core.Model.observedlogp` :
+        :py:attr:`observedlogp` :
             convenience property for logp of all observed RVs.
-        :py:attr:`~pymc.model.core.Model.potentiallogp`.
+        :py:attr:`potentiallogp`.
             convenience property for all additional logp terms (potentials).
-        :py:attr:`~pymc.model.core.Model.point_logps` :
+        :py:attr:`point_logps` :
             convenience property for numerical evaluation of local logps at a point.
         """
         varlist: list[TensorVariable]
@@ -837,11 +837,11 @@ class Model(WithMemoization, metaclass=ContextMeta):
 
         See Also
         --------
-        :py:meth:`~pymc.model.core.Model.compile_dlogp` :
+        :py:meth:`compile_dlogp` :
             gradient of log-probability density as a compiled function.
-        :py:meth:`~pymc.model.core.Model.logp` :
+        :py:meth:`logp` :
             log-probability density as a Variable (in a symbolic form).
-        :py:meth:`~pymc.model.core.Model.d2logp` :
+        :py:meth:`d2logp` :
             Hessian of log-probability density as a Variable (in a symbolic form).
         """
         if vars is None:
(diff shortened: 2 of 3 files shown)


docs/source/api/distributions/transforms.rst
```diff
@@ -2,7 +2,7 @@
 Transformations
 ***************
 
-.. currentmodule:: pymc.distributions.transforms
+.. module:: pymc.distributions.transforms
```
ricardoV94, line 5: cc @OriolAbril

pymc/model/core.py
```diff
@@ -582,20 +582,37 @@ def compile_logp(
...
-        jacobian : bool
-            Whether to include jacobian terms in logprob graph. Defaults to True.
-        sum : bool
-            Whether to sum all logp terms or return elemwise logp for each variable.
-            Defaults to True.
+        vars : Variable, sequence of Variable or None, default None
+            Random variables or potential terms whose contribution to logp is to be included.
+            If None, use all basic (free or observed) variables and potentials defined in the model.
+        jacobian : bool, default True
+            If True, add Jacobian contributions associated with automatic variable transformations,
```
ricardoV94, line 593: Needs some working.

pymc/model/core.py
```diff
@@ -742,19 +818,30 @@ def dlogp(
         vars: Variable | Sequence[Variable] | None = None,
         jacobian: bool = True,
     ) -> Variable:
-        """Gradient of the models log-probability w.r.t. ``vars``.
+        """Gradient of the joint log-probability density of the model.
```
ricardoV94, line 821: density is a strong word, model may be discrete or a mix

docs/source/api/distributions/transforms.rst
avm19: Just to clarify. `:module::` *must* be specified at least once (and probably only once) for Sphinx to create an "index entry". Without it, references `:currentmodule::...` and ` :mod:...` would lead to nowhere. It so happens that `module:: pymc.distributions.transforms` was absent, even though `currentmodule` was present, and Sphinx does not check this.

pymc/model/core.py
avm19: Anything specific?

pymc/model/core.py
avm19: Technically, a probability mass function is a probability density (i.e. a Radon-Nikodym derivative) with respect to the counting measure, in contrast to a "regular" PDF of an absolutely continuous variable, which is wrt the standard Lebesgue-Borel measure; and so the mixed case should also be a density wrt a certain product measure.

I agree that people might find this confusing, and some may even assume that discrete variables are marginalised out or not allowed at all in this context. On the other hand, I used the word "density" because Jacobians arise only in (absolutely) continuous variables; if it were not a density, there would be no Jacobians.

docs/source/api/distributions/transforms.rst
OriolAbril: Using the module directive here is probably fine, I am not sure how relevant it is though. Is `pymc.distributions.transforms` as a module something relevant to users we'll want to reference? If so we need to switch, otherwise using `currentmodule` is perfectly fine but using `module` won't hurt either.

---

Full context:

`.. currentmodule::` is basically syntactic sugar, its only role is defining the module name to prepend all the autodoc/autosummary entries in that file. That is, we can use `circular` or `ordered` in the autosummary directive below instead of needing to specify `pymc.distributions.transforms.circular`...

`.. module::` does the same but also generates a _target_ that can be referenced from other places using the `` :mod:`pymc.distributions.transforms` `` role; `.. currentmodule::` does not reference the `module` directive with the same name nor needs it to exist to work properly though.

The `module` directive can also take a `:synopsis:` to manually add the module docstring. In our case however, using this optional argument wouldn't make much sense, it would be much better to use `.. automodule::` instead which defines a `module` directive pulling the docstring directly from the module in question (like we do for classes or functions).

Sphinx does not check this because `module`/`automodule` should indeed be used at most once, but it isn't really necessary to use it at least once, that depends on how the package maintainers decide to define and document the public API.

avm19: > Using the module directive here is probably fine, I am not sure how relevant it is though.

Without it references did not work, which is why I made the change.

P.S. I don't remember if Sphinx threw an error or just no hyperlink without the `module`. I don't know much about this, but I guess you were not using `automodule` because this module is accompanied by a lot of text, which was placed in `transform.rst` (better have it there, than a huge module-level docstring). Some other entries, e.g. `shape_utils.rst`, also contain text in addition to Sphinx directives.

> Is `pymc.distributions.transforms` as a module something relevant to users we'll want to reference?

Yes. The module page explains how PyMC deals with transformations, and I reference it as a good source for further details in a method's docstring.

pymc/model/core.py
ricardoV94: We do constant use the term log-probability though. Yes jacobian corrections don't exist for discrete variables, I don't think that's going to be what makes it confusing to understand jacobian kwarg here

pymc/model/core.py
ricardoV94: `Include jacobian correction term for transformed variables` ?

ricardoV94: I don't think we should explain why, specially because the why is context specific. Maybe you don't want it because you want to do optimization on the constrained space, using unconstrained variables.

avm19: It is the goal of this PR to give a concise and clear description of what `jacobian=True` does, so this bit is central. I am trying to share my perspective of an outsider to make the docstrings friendlier to users who are not used to PyMC parlance. 

>I don't think we should explain why

Here I am explaining not so much "why", as "what". The phrase "`Jacobian correction term`" might sound precise and unquestionable to people working on PyMC or specialising in Bayesian modelling, but to someone coming from a slightly different field it is very ambiguous and uninformative. Which Jacobian? Why correction? There are so many Jacobians, it is not clear which one it is about. Any gradient, such as `Model.dlogp`, is also a Jacobian (neglect the differential geometry subtleties of co- and contra-variance), and one would naturally think that the Jacobian refers to `dlogp`, because `d2logp` is called a Hessian in the same file. Further, what is meant by "correction"? Who made the error that needs to be corrected? Or is it like the prediction-correction optimisation methods used in engineering? They use first and second derivatives, so sounds about right... Frankly, I never saw the change-of-variables Jacobian determinant being referred to as "correction", so it is the last connection I would make.

I am trying to explain what `jacobian` does in a way that does not allow misinterpretations. I am not try trying to convince the reader _why_ they should use it.

I thought about using "which" here:

> jacobian : bool, default True
>   If True, add Jacobian contributions associated with automatic variable transformations,
>   ~so that~ **which make** the result ~is~ the true density of transformed random variables.
>   See :py:mod:`pymc.distributions.transforms` for details.

but then it is not clear if "which" refers to "transformations" (incorrect) or to "contributions" (correct). This is why I chose "so that" as a conjunction -- it refers to "add", i.e. the purpose of the parameter `jacobian`.

ricardoV94: I don't like the `true` qualifier. What about:

```
  jacobian : bool, default True
      Include log-det-Jacobian terms for change of variables                                                                                                                                                                                 
      introduced by transformed random variables.        
```
Should be clear which jacobian(s) and where it comes from?

avm19: How about this:

> jacobian : bool, default True
>   Add Jacobian terms related to automatic variable transformations,
>   so that the result is the true log density of transformed random variables.
>   See :py:mod:`pymc.distributions.transforms` for details.

I think "`automatic`" is important, because the user might not realise that transformations may take place behind the scenes without their express request. 

I also want to keep "`true density of transformed random variables`". The message here is that the resulting expression is like $\ln p_{X_t, Y_t, Z_t, ...}(x_t,y_t,z_t,...)$, as opposed to $\ln p_{X,Y,Z,...}(x_t,y_t,z_t,...)$ or $\ln p_{X,Y,Z,...}(x(x_t),y(y_t),z(z_t),...)$ or other possibilities. It is easy to get lost in transformed and untransformed variables when the former can be expressed through the latter and vice versa. This short clarification is intended to assure the user if they are on the right track or correct them if not. 

And the reference to `transforms` module is there for those users who have no idea what this is about; the module page has a nice explanation. Alternatively, one could reference [guides/Probability_Distributions.html#auto-transformation](https://www.pymc.io/projects/docs/en/stable/guides/Probability_Distributions.html#auto-transformation), but I like the module page better.

ricardoV94: They may not be automatic, the user can override behavior with `transform=x` when defining a variable.

I'm not sure about the true thing still, doesn't sound quite right.

pymc/model/core.py
avm19: Actually, the word "density" was there before me in `compile_logp`, `compile_dlogp`, `compile_d2logp` only. I only put it everywhere else for consistency. 

https://github.com/pymc-devs/pymc/blob/8a1896bad8a12fecc14bb59e71f60c01475581bb/pymc/model/core.py#L613

Should I make it "log-probability" (hyphenated, noun) and remove "density" from everywhere (except `jacobian` parameter where "density" is relevant)?

pymc/model/core.py
avm19: > They may not be automatic, the user can override behavior with `transform=x` when defining a variable.

They are still automatic in the sense that you don't have to change variables in the model definition. Maybe it would be correct to say that all these transformations are automatic, which includes both the default transformations and the user-specified transformations? Here is a relevant quote:
https://github.com/pymc-devs/pymc/blob/8a1896bad8a12fecc14bb59e71f60c01475581bb/docs/source/api/distributions/transforms.rst?plain=1#L82-L85


> true thing still, doesn't sound quite right.

Oh, I misunderstood your previous message, I thought you were referring to somewhat redundant "`If True ...`". Let me think if there are ways to re-phrase "`true density`".

avm19: > I'm not sure about the true thing still, doesn't sound quite right

You are right, "`true density`" doesn't sound right, neither does "proper", and I haven't found anything better. How about:

> jacobian : bool, default True
>   If True, add Jacobian contributions associated with automatic variable transformations,
>   so that the result is the log-probability density of transformed random variables.
>   See :py:mod:`pymc.distributions.transforms` for details.

That is, if I convinced you that user-specified and default transforms are all "automatic". Otherwise, remove the word "automatic".

jessegrabowski: Agreed that "true" is not right, it makes a subjective judgement where none exists. You can see an example [here](https://github.com/pymc-devs/pymc-extras/pull/641) of a case where applying it leads to the "false" logp. It applies a change of variables correction implied by the model's transformation, if one exists. I would write a docstring that says something like that:

```
jacobian: bool, default True
     If True, the change-of-variable correction implied by each random variable's transformation, if applicable. This correction accounts for the distortion of probability mass induced by the application of a non-linear function to a random variable. See :py:mod:pymc.distributions.transforms for details. 
```

avm19: > You can see an example [here](https://github.com/pymc-devs/pymc-extras/pull/641) of a case where applying it leads to the "false" logp.

After a cursory reading, I tend to think that the confusion in there was between  $\ln p_{X_t, Y_t, Z_t, ...}(x_t,y_t,z_t,...)$ and $\ln p_{X,Y,Z,...}(x(x_t),y(y_t),z(z_t),...)$. Both are functions of transformed variables, both are "true" logps. But the former is the log density of transformed RVs, and the latter is the log density of the original RVs. To sample transformed variables one would use $p_{X_t, Y_t, Z_t}(...)$, and to optimise $p_{X,Y,Z}(...)$ one would use $p_{X,Y,Z}(...)$. I alluded to this in the conversation and in my suggested version. The "`true`" was a bad choice (it may also suggest ground truth vs approximation). But I think `"density of transformed random variables"` is a good part, which pushes the reader in the right direction. 

Maybe we should also add that with `jacobian=False` the result is "density of ***un***transformed variables"? (I am like 95% sure, but I can check the maths and get a toy example)

ricardoV94: (I tend to think of transformed/untransformed as unconstrained/constrained, but that's giving an interpretation of why the transform was applied)

>  Maybe we should also add that with jacobian=False the result is "density of untransformed variables"? (I am like 95% sure, but I can check the maths and get a toy example)

I would maybe try to emphasize this function always takes a point in transformed space and "untransforms" it to evaluate the logprob function of each random variable in it's "natural"/original space. What's optional is whether we add the terms corresponding to this transformation.

This is not a specific suggestion, just sharing how I think mechanistically about it.

PR #8199, opened 2026-03-16
Adapt to PyTensor v3 major changes

ricardoV94:
So far only needed to handle the AdvancedSubtensor refactor (and a JAX test that now passes)


ricardoV94:
fantastic widows is crashing in every job...

ricardoV94:
Ah because we're not using conda... ofc(?)

jessegrabowski (approved):

PR #8201, opened 2026-03-17
Fix off-by-one progress bar bugs

jessegrabowski:
Progress bar is currently off by one when it completes, and @ricardoV94 is bullying me about it. The issue was we added a `-1` factor to the total at some point when we shouldn't have. This caused the samplers to stop at `n-1 / n` (for which I was bullied). In the case of SMC sampling, it caused the progressbar maximum to be 0, so progress immediately stopped and no timing info is given.

I also re-wrote the test_manager tests. They weren't testing anything, just type checks and deterministic code paths. I added some tests that would have caught these off-by-one bugs.

I also simplified the `is_last` logic. Previously it was trying to calculate the remaining steps then advancing by that much, then doing a cleanup step. Now we just directly set the total completed to the expected total. This should be 100% bulletproof against embarrassing off-by-one bugs.

### DIFF
--- a/pymc/progress_bar/marimo_progress.py
+++ b/pymc/progress_bar/marimo_progress.py
@@ -183,12 +183,7 @@ class MarimoProgressBackend:
         self._task_state[task_id]["stats"] = stats
 
         if is_last:
-            # Ensure bar is fully filled on completion
-            total = self._task_state[task_id]["total"]
-            completed = self._task_state[task_id]["completed"]
-            remaining = total - completed
-            if remaining > 0:
-                self._task_state[task_id]["completed"] = total
+            self._task_state[task_id]["completed"] = self._task_state[task_id]["total"]
 
         self._render()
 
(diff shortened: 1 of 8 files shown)


jessegrabowski:
Failures look unrelated?

fonnesbeck (approved):
LGTM. Let's merge this first and I can rebase #8184 from it.

I was getting the same failures for #8184 and they magically disappeared one day.

ricardoV94:
The failures was due to patches on main that are not in V3 yet. I changed the branch of both PRs to main (@fonnesbeck 's a few days ago, which is how they "magically" went away)

fonnesbeck:
Magic!

ricardoV94:
It's hanging in the `Check for changes` job

ricardoV94:
@fonnesbeck merged

jessegrabowski:
> Magic!

Any sufficiently advanced Ricardo PR is indistinguishable from magic

PR #8203, opened 2026-03-17
add optional TypedDict type annotations for pm.Model

williambdean:
## Description

Adds optional TypedDict type annotations to `pm.Model`. Works both pm and pmd distributions.

Example: 

```python
from typing import TypedDict, NotRequired

from pytensor.tensor.variable import TensorVariable
import pymc as pm

class ModelVars(TypedDict)
    mu: TensorVariable
    sigma: TensorVariable
    nu: NotRequired[TensorVariable]


model: pm.Model[ModelVars]
```

This really stands a documentation and structure to specific models. No runtime checks here. Some IDEs might get better autocomplete.

## Related Issue

- [ ] Closes #
- [ ] Related to #

## Checklist

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Maintenance
- [ ] Other (please specify):

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -21,7 +21,9 @@ import warnings
 
 from collections.abc import Iterable, Sequence
 from typing import (
+    Generic,
     Literal,
+    TypeVar,
     cast,
     overload,
 )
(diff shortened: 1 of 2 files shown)


ricardoV94:
This should have been discussed before. I'm against it.

That sounds more like something for ModelBuilder or other Model wrapper classes. It's also just one way we could think of hint-parametrize Model, why should it be this one?

williambdean:
Open to having it placed somewhere else. However, even objects like ModelBuilder still
just rely directly on pm.Model and az.InferenceData (and xr.Dataset) (also very arbitrary) 

How might these alternative solutions look?

PR #8205, opened 2026-03-18
Richer textual representation: show Data variables, fold constants, and use semantic separators

### LINKED ISSUE #7536: Textual representation of the model does not show Data variables or dependency on them
### Description

```python
import pymc as pm
from pymc.printing import str_for_model

with pm.Model() as m:
    x = pm.Data("x", 0)
    y = pm.Normal("y")
    f = pm.Deterministic("f", x + y)
    
str_for_model(m)
# y ~ Normal(0, 1)
# f ~ Deterministic(f(y))
```

### LINKED ISSUE #7538: Richer textual representation of parameter relationships
### Description

One easy gain is to get rid of useless `f()` that depend only on contants.

```python
import pymc as pm
from pymc.printing import str_for_model

with pm.Model() as m:
    x = pm.Exponential("x", lam=2)

str_for_model(m)
# x ~ Exponential(f())
```

Instead if we rewrite the graph first:
```python
import pymc as pm
from pymc.printing import str_for_model

from pytensor.graph.rewriting.utils import rewrite_graph
from pymc.model.fgraph import fgraph_from_model, model_from_fgraph

with pm.Model() as m:
    x = pm.Exponential("x", lam=2)

fgraph, _ = fgraph_from_model(m)
new_m = model_from_fgraph(rewrite_graph(fgraph))
str_for_model(new_m)
# x ~ Exponential(0.5)
```

We may also want to use pytensor pretty print to show the content of some of the functions:
```python
import pytensor.tensor as pt
from pytensor.printing import pprint

x = pt.matrix("x")
y = pt.matrix("y")
mu = pm.math.dot(x, y.T) ** 2
    
pprint(mu)
# ((x \dot y.T) ** 2)
```

drbenvincent:
## Summary

Improve model text representations by addressing two open issues and one semantic cleanup:

- **#7536** — `pm.Data` variables now appear in `str_for_model` output and in Deterministic/Potential dependency lists.
- **#7538** — Constant-only parameters no longer render as opaque `f()`; they now render as `<constant>`.
- **Semantic separators** — Data and Deterministic variables use `=` instead of `~`. Potential keeps `~`.

## Closes #7536 — Data variables now visible

For:

```python
with pm.Model() as m:
    x = pm.Data("x", 0)
    y = pm.Normal("y")
    f = pm.Deterministic("f", x + y)
```

`str_for_model(m)` now includes both:

- `x = Data(0)`
- `f = Deterministic(f(y, x))`

## Closes #7538 — Constant-only `f()` eliminated

For:

```python
with pm.Model() as m:
    x = pm.Exponential("x", lam=2)
```

`str_for_model(m)` now shows:

- `x ~ Exponential(<constant>)`

So the empty `f()` placeholder is gone, without adding constant-fold computation in this print path.

## Semantic separators

- Data: `x = Data(...)`
- Deterministic: `mu = Deterministic(...)`
- Potential: `pot ~ Potential(...)`
- Random variables: unchanged (`~`)

### `pymc/printing.py`

- Added `str_for_data_var` for Data variable rendering.
- Included `model.data_vars` in `str_for_model` output.
- Updated alignment logic to support both `~` and `=` in plain and LaTeX output.
- Deterministic uses `=` and Potential uses `~` in `str_for_potential_or_deterministic`.
- Threaded model-derived `named_vars` through internals for dependency naming.
- Removed heuristic fallback checks in `_str_for_input_var`.
- Kept constant-only expression rendering as `<constant>` (no print-time constant folding pass).

### `tests/test_printing.py`

- Updated expected outputs for:
  - semantic separators (`Deterministic` now `=`)
  - constant-only parameter rendering (`f()` -> `<constant>`)
- Added targeted tests:
  - `test_data_vars_in_model_repr`
  - `test_constant_folding_in_repr`
  - `test_data_var_repr_no_params`
  - `test_data_var_latex_underscore_escaping`
  - `test_function_of_named_var_in_repr`
- Simplified repr tests to focus on model repr assertions per review feedback.

## Test plan

- [x] `pytest tests/test_printing.py -v`
- [x] `pre-commit run --all-files`

### DIFF
--- a/pymc/printing.py
+++ b/pymc/printing.py
@@ -111,14 +111,14 @@ def str_for_data_var(
         latex_name = r"\text{" + _latex_escape(print_name.strip("$")) + "}"
         latex_name = _format_underscore(latex_name)
         if value_str is not None:
-            return rf"${latex_name} \sim \operatorname{{Data}}({value_str.strip('$')})$"
+            return rf"${latex_name} = \operatorname{{Data}}({value_str.strip('$')})$"
         else:
-            return rf"${latex_name} \sim \operatorname{{Data}}$"
+            return rf"${latex_name} = \operatorname{{Data}}$"
     else:
         if value_str is not None:
-            return rf"{print_name} ~ Data({value_str})"
+            return rf"{print_name} = Data({value_str})"
         else:
-            return rf"{print_name} ~ Data"
+            return rf"{print_name} = Data"
 
 
 def str_for_model(model: Model, formatting: str = "plain", include_params: bool = True) -> str:
@@ -146,25 +146,34 @@ def str_for_model(model: Model, formatting: str = "plain", include_params: bool 
         return ""
     if "latex" in formatting:
         var_reprs = [_format_underscore(x) for x in var_reprs]
-        var_reprs = [
-            var_repr.replace(r"\sim", r"&\sim &").strip("$")
-            for var_repr in var_reprs
-            if var_repr is not None
-        ]
+        formatted = []
+        for var_repr in var_reprs:
+            if var_repr is None:
+                continue
+            s = var_repr.strip("$")
+            if r"\sim" in s:
+                s = s.replace(r"\sim", r"&\sim &", 1)
+            else:
+                s = s.replace(" = ", " &= &", 1)
+            formatted.append(s)
         return r"""$$
             \begin{{array}}{{rcl}}
             {}
             \end{{array}}
-            $$""".format("\\\\".join(var_reprs))
+            $$""".format("\\\\".join(formatted))
     else:
-        # align vars on their ~
-        names = [s[: s.index("~") - 1] for s in var_reprs]
-        distrs = [s[s.index("~") + 2 :] for s in var_reprs]
-        maxlen = str(max(len(x) for x in names))
-        var_reprs = [
-            ("{name:>" + maxlen + "} ~ {distr}").format(name=n, distr=d)
-            for n, d in zip(names, distrs)
-        ]
+        sep_pattern = re.compile(r" ([~=]) ")
+        names = []
+        seps = []
+        distrs = []
+        for s in var_reprs:
+            m = sep_pattern.search(s)
+            assert m is not None
+            names.append(s[: m.start()])
+            seps.append(m.group(1))
+            distrs.append(s[m.end() :])
+        maxlen = max(len(n) for n in names)
+        var_reprs = [f"{n:>{maxlen}} {sep} {d}" for n, sep, d in zip(names, seps, distrs)]
         return "\n".join(var_reprs)
 
 
@@ -180,17 +189,19 @@ def str_for_potential_or_deterministic(
     values included.
     """
     print_name = var.name if var.name is not None else "<unnamed>"
+    sep_plain = "~" if dist_name == "Potential" else "="
+    sep_latex = r"\sim" if dist_name == "Potential" else "="
     if "latex" in formatting:
         print_name = r"\text{" + _latex_escape(print_name.strip("$")) + "}"
         if include_params:
-            return rf"${print_name} \sim \operatorname{{{dist_name}}}({_str_for_expression(var, formatting=formatting)})$"
+            return rf"${print_name} {sep_latex} \operatorname{{{dist_name}}}({_str_for_expression(var, formatting=formatting)})$"
         else:
-            return rf"${print_name} \sim \operatorname{{{dist_name}}}$"
+            return rf"${print_name} {sep_latex} \operatorname{{{dist_name}}}$"
     else:  # plain
         if include_params:
-            return rf"{print_name} ~ {dist_name}({_str_for_expression(var, formatting=formatting)})"
+            return rf"{print_name} {sep_plain} {dist_name}({_str_for_expression(var, formatting=formatting)})"
         else:
-            return rf"{print_name} ~ {dist_name}"
+            return rf"{print_name} {sep_plain} {dist_name}"
 
 
 def _str_for_input_var(var: Variable, formatting: str) -> str:
--- a/pymc/printing.py
+++ b/pymc/printing.py
@@ -96,9 +96,7 @@ def str_for_dist(dist: Variable, formatting: str = "plain", include_params: bool
                 return dist_name
 
 
-def str_for_data_var(
-    var: Variable, formatting: str = "plain", include_params: bool = True
-) -> str:
+def str_for_data_var(var: Variable, formatting: str = "plain", include_params: bool = True) -> str:
     """Make a human-readable string representation of a Data variable in a model."""
     print_name = var.name if var.name is not None else "<unnamed>"
 
@@ -278,11 +276,7 @@ def _str_for_expression(var: Variable, formatting: str) -> str:
                     xname = rf"<{opname}>"
             assert xname is not None
             names.append(xname)
-        elif (
-            isinstance(x, SharedVariable)
-            and x.name
-            and not isinstance(x.type, RandomType)
-        ):
+        elif isinstance(x, SharedVariable) and x.name and not isinstance(x.type, RandomType):
             parents.append(x)
             names.append(x.name)
 
--- a/pymc/printing.py
+++ b/pymc/printing.py
@@ -96,7 +96,9 @@ def str_for_dist(dist: Variable, formatting: str = "plain", include_params: bool
                 return dist_name
 
 
-def str_for_data_var(var: Variable, formatting: str = "plain", include_params: bool = True) -> str:
+def str_for_data_var(
+    var: Constant | SharedVariable, formatting: str = "plain", include_params: bool = True
+) -> str:
     """Make a human-readable string representation of a Data variable in a model."""
     print_name = var.name if var.name is not None else "<unnamed>"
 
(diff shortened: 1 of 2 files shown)


drbenvincent:
The two failing CI checks (`alternative_backends` cvm and numba) are unrelated to this PR. Both fail on the same test:

```
test_dirichlet_multinomial_dims[JAX] — Failed: DID NOT RAISE <class 'TypeError'>
```

This test expects a `TypeError` when sampling `DirichletMultinomial` with JAX, but the error is no longer raised — likely due to an upstream JAX or PyTensor fix. This PR only touches `pymc/printing.py` and `tests/test_printing.py` and has no interaction with alternative backends or random sampling.

pymc/printing.py
```diff
@@ -178,13 +219,21 @@ def _is_potential_or_deterministic(var: Variable) -> bool:
...
+            return _str_for_expression(var, formatting)
+        try:
+            from pymc.exceptions import NotConstantValueError
+            from pymc.pytensorf import constant_fold
+
+            [folded] = constant_fold([var], raise_not_constant=True)
+            assert isinstance(folded, np.ndarray)
+            return _str_for_constant_value(folded, formatting)
+        except NotConstantValueError:
+            return _str_for_expression(var, formatting)
```
ricardoV94: This is too expensive for a print method

pymc/printing.py
```diff
@@ -233,12 +288,13 @@ def _expand(x):
             parents.append(x)
             xname = x.name
             if xname is None:
-                # If the variable is unnamed, we show the op's name as we do
-                # with constants
                 if (opname := getattr(x.owner.op, "name", None)) is not None:
                     xname = rf"<{opname}>"
             assert xname is not None
             names.append(xname)
+        elif isinstance(x, SharedVariable) and x.name and not isinstance(x.type, RandomType):
```
ricardoV94: I already left a comment elsewhere in the codebase. We should only pretty print models, and we know all variables categorizations (deterministic, potential, data_var, etc.). Just use that knowledge instead of this approach where we try to guess what is what as if we didn't know it

drbenvincent:
Thanks for the review @ricardoV94! Both points addressed in the latest push:

**1. `constant_fold` removed** — pure-constant expressions now show `<constant>` instead of computing values. Zero compilation overhead.

**2. `named_vars` threaded from model** — `str_for_model` builds a set of all known model variables (`data_vars`, `free_RVs`, `observed_RVs`, `deterministics`, `potentials`) and passes it through to `_str_for_input_var` and `_str_for_expression`. Ancestor identification now uses model knowledge instead of isinstance guessing. The `SharedVariable` isinstance hack has been removed from `_str_for_expression`.

**One question for your guidance:** I kept `_is_potential_or_deterministic` (the existing monkey-patch check with the FIXME) as a fallback for standalone calls where `named_vars` is `None` — e.g. when a user calls `var.str_repr()` on an individual variable outside of `str_for_model`. When `named_vars` is provided (the `str_for_model` path), it's completely bypassed.

Should I remove `_is_potential_or_deterministic` entirely? That would mean standalone `str_repr` for potentials/deterministics would expand through named intermediaries (e.g. a Potential depending on a Deterministic `mu` would show `f(alpha, beta)` instead of `f(mu)`). Happy to go either way.

pymc/printing.py
```diff
@@ -143,48 +201,54 @@ def str_for_potential_or_deterministic(
...
         else:
-            return rf"{print_name} ~ {dist_name}"
+            return rf"{print_name} {sep_plain} {dist_name}"
 
 
-def _str_for_input_var(var: Variable, formatting: str) -> str:
+def _str_for_input_var(
+    var: Variable, formatting: str, named_vars: set[Variable] | None = None
+) -> str:
     def _is_potential_or_deterministic(var: Variable) -> bool:
```
ricardoV94: get rid of this fallback

pymc/printing.py
```diff
@@ -143,48 +201,54 @@ def str_for_potential_or_deterministic(
...
             return False
 
     if isinstance(var, Constant | SharedVariable):
         return _str_for_constant(var, formatting)
-    elif isinstance(var.owner.op, MeasurableOp) or _is_potential_or_deterministic(var):
-        # show the names for RandomVariables, Deterministics, and Potentials, rather
-        # than the full expression
+    elif (
+        (named_vars is not None and var in named_vars)
+        or isinstance(var.owner.op, MeasurableOp)
```
ricardoV94: get rid of these checks and rely on named_vars always existing

ricardoV94 (commented):
Left some comments, go more aggressive with ripping off these fallbacks

drbenvincent:
Thanks @ricardoV94 — I went more aggressive on removing fallbacks.

Changes in this push:
- Removed the monkey-patch fallback path in `_str_for_input_var` (`_is_potential_or_deterministic`).
- Removed optional fallback branching in internal traversal helpers and made them rely on the model-driven `named_vars` path.
- Kept standalone behavior explicit (no heuristic inference of named intermediates), and added a focused test:
  - `test_standalone_potential_repr_does_not_use_named_var_fallback`

I also updated `TestMonolith` expectations to separate standalone variable repr from model repr where they now intentionally diverge after fallback removal.

Validation run:
- `conda run -n pymc-dev pytest tests/test_printing.py -v`
- Result: `18 passed`

tests/test_printing.py
```diff
@@ -46,19 +46,21 @@ def test__repr_latex_(self):
...
+        model_expected = getattr(self, "model_expected", self.expected)
 
         # make sure each variable is in the model
-        for tex in self.expected[("latex", True)]:
+        for tex in model_expected[("latex", True)]:
             for segment in self._latex_segments(tex):
                 assert segment in model_tex
 
     def test_str_repr(self):
+        model_expected = getattr(self, "model_expected", self.expected)
```
ricardoV94: why is there a self.model_expected and self.expected now?

tests/test_printing.py
```diff
@@ -459,6 +468,23 @@ def test_constant_fold_fallback():
     assert "b ~ Normal(f(a), 1)" in text
 
 
+def test_standalone_potential_repr_does_not_use_named_var_fallback():
```
ricardoV94: This makes it sound like it's a behavior we try hard to enforce. imo we don't need this test at all

tests/test_printing.py
```diff
@@ -370,6 +388,103 @@ def test_latex_repr(self):
...
+    latex_no_params = str_for_data_var(my_data, formatting="latex", include_params=False)
+    assert r"my\_data" in latex_no_params
+    assert r"\operatorname{Data}" in latex_no_params
+
+    # Via model repr
+    model_latex = model.str_repr(formatting="latex")
+    assert r"my\_data" in model_latex
+
+
+def test_constant_fold_fallback():
```
ricardoV94: this test name is outdated now that we're not doing constant fold? Better name is test_function_of_named_var?

ricardoV94:
> I also updated TestMonolith expectations to separate standalone variable repr from model repr where they now intentionally diverge after fallback removal.

Just remove the tests of direct variable repr

drbenvincent:
Thanks @ricardoV94 — I simplified the tests in line with your latest feedback.

Changes made:
- Removed direct variable repr assertions from the shared test flow (tests are now model-repr-focused).
- Removed the `self.model_expected` vs `self.expected` indirection.
- Removed `test_standalone_potential_repr_does_not_use_named_var_fallback`.
- Renamed `test_constant_fold_fallback` to `test_function_of_named_var_in_repr` to match current semantics.

Validation:
- `conda run -n pymc-dev pytest tests/test_printing.py -v` -> 17 passed
- `conda run -n pymc-dev pre-commit run --all-files` -> passed

drbenvincent:
@ricardoV94 I pushed a cleanup follow-up based on your latest review feedback:

- removed the `self.model_expected` split and the direct variable-level repr assertions
- removed the standalone fallback-focused test
- renamed the outdated fallback test to `test_function_of_named_var_in_repr`
- reran `pytest tests/test_printing.py -v` and `pre-commit run --all-files` (both pass)

I also updated the PR description so #7538 accurately reflects current behavior (`f()` -> `<constant>`, not numeric folding).

Could you please take another look when you have a moment?

tests/test_printing.py
```diff
@@ -370,6 +371,86 @@ def test_latex_repr(self):
                 assert text in model_text
 
 
+def test_data_vars_in_model_repr():
+    """Data variables appear in model repr and in Deterministic dependency lists (issue #7536)."""
+    import pymc as pm
```
ricardoV94: remove local import. Don't use pm. but explicit imports like other tests

tests/test_printing.py
```diff
@@ -370,6 +371,86 @@ def test_latex_repr(self):
...
+    assert "x = Data(0)" in text
+    assert "y ~ Normal(0, 1)" in text
+    assert "f = Deterministic(f(y, x))" in text
+
+    latex = model.str_repr(formatting="latex")
+    assert r"&= &\operatorname{Data}(0)" in latex
+    assert r"\text{x}" in latex
+
+
+def test_constant_folding_in_repr():
```
ricardoV94: Don't mention "folding" in test title

ricardoV94 (commented):
Two final nits, looks great

tests/test_printing.py
drbenvincent: Done in b2498afd1 — removed local  usage from the new tests and switched to explicit imports/symbols to match the rest of this test file.

tests/test_printing.py
drbenvincent: Done in b2498afd1 — renamed  to  so the title no longer mentions folding.

tests/test_printing.py
drbenvincent: Clarification: removed local `import pymc as pm` usage from the new tests and switched to explicit imports/symbols to match file conventions.

tests/test_printing.py
drbenvincent: Clarification: renamed `test_constant_folding_in_repr` to `test_constant_only_expression_in_repr` so the title matches current semantics.

drbenvincent:
@ricardoV94: addressed your two final nits in b2498afd1 (explicit imports in new tests and renamed test title), reran pytest on `tests/test_printing.py` plus pre-commit successfully, and resolved both threads. Could you give this a final pass?

ricardoV94 (approved):

ricardoV94:
Thanks @drbenvincent

PR #8209, opened 2026-03-21
Change `sample_posterior_predictive` API wrt to volatility

### LINKED ISSUE #6614: Remove `givens_dict` argument in sample posterior predictive internal function
https://github.com/pymc-devs/pymc/blob/473c9528e56fc291e323e30082c8f4295dbd6149/pymc/sampling/forward.py#L618

If I pass variables to cmopile forward function, would it be a correct do-inference?

### LINKED ISSUE #6876: `sample_posterior_predictive` does not consider changes in "ConstantData" and "Constant coords"

### LINKED ISSUE #7069: Make `sample_posterior_predictive` API less surprising

ricardoV94:
Keep in mind this is like our second most important function in PyMC, and it was easily a footgun whenever var_names was involved or there were unobserved RVs that depended on modified data/coordinates (see #7069).

After this RV variables in the trace are never resampled by default, even if it may mean the graph is invalid (shape or semantically). A warning is issued, directing users to explicitly add variables "deemed volatile" into `sample_vars` (to resample them) or `freeze_vars` (to confirm this was intended).

Regular uses that don't touch `var_names` should stay the same as before. After these I think `var_names` becomes very intuitive (and less important as well)

**Main knobs**
- `sample_vars` — variables to force sampling. RVs present in the trace will never be resampled unless they show up here. Deterministics are still recomputed automatically based on volatility, but can be included here in case the volatility analysis wouldn't trigger them (there's a pm.do use case in the tests/docstrings). Variables that don't exist in the trace (including observeds) are always sampled, but can be added here for completeness.
- `freeze_vars` — variables to explicitly reuse from the trace. Should also be used for RVs in the trace that seem volatile but are intended to not be resampled. Also works for deterministics — e.g. the HSGP-style case of a training-time mean-centering that shouldn't be rederived from prediction data.                                                                                                
- `ImplicitFreezeWarning` in `pymc.exceptions` — emitted when a trace RV is deemed volatile but not explicitly in `sample_vars` or `freeze_vars`. This is the knob to alert users they need to make a decision, reducing silent "footguns".
- `var_names` is strictly output-gating now. It just determines what variables appear in the returned trace.

**Other changes**

- `var_names` / `sample_vars` / `freeze_vars`, and everything after `model`, are now keyword-only.
- A `pm.Data` replaced by a `TensorConstant` with a different value (e.g. via `pm.do(…, make_interventions_shared=False)` or `freeze_dims_and_data`) now participates in volatility detection (closes #6876).
- Docstring rewritten: common uses first (standard PPC, `set_data` predictions, freezing/forcing a deterministic, out-of-model predictions), detailed volatility rules only after.
- Some (hopefully) sane default for default var_names, when `sample_vars` is specified.
- Removed givens argument in the inner function (closes #6614)

Closes #7069 
Closes #6876 
Closes #6614

TODO:
- [x] clean code
- [x] clean tests

### DIFF
--- a/pymc/sampling/forward.py
+++ b/pymc/sampling/forward.py
@@ -31,15 +31,13 @@ import xarray as xr
 from pytensor import tensor as pt
 from pytensor.graph import vectorize_graph
 from pytensor.graph.basic import (
-    Apply,
     Constant,
     Variable,
 )
 from pytensor.graph.fg import FunctionGraph
 from pytensor.graph.traversal import ancestors, general_toposort, walk
-from pytensor.tensor.random.variable import RandomGeneratorSharedVariable
+from pytensor.tensor.random.type import RandomType
 from pytensor.tensor.sharedvar import SharedVariable, TensorSharedVariable
-from pytensor.tensor.variable import TensorConstant
 from rich.theme import Theme
 
 import pymc as pm
@@ -48,6 +46,7 @@ from pymc.backends.arviz import _DefaultTrace, dataset_to_point_list
 from pymc.backends.base import MultiTrace
 from pymc.blocking import PointType
 from pymc.distributions.shape_utils import change_dist_size
+from pymc.exceptions import ImplicitFreezeWarning
 from pymc.model import Model, modelcontext
 from pymc.progress_bar import create_simple_progress, default_progress_theme
 from pymc.pytensorf import compile, rvs_in_graph
@@ -72,26 +71,41 @@ PointList: TypeAlias = list[PointType]
 _log = logging.getLogger(__name__)
 
 
-def get_constant_coords(trace_coords: dict[str, np.ndarray], model: Model) -> set:
-    """Get the set of coords that have remained constant between the trace and model."""
-    constant_coords = set()
-    for dim, coord in trace_coords.items():
-        current_coord = model.coords.get(dim, None)
-        current_length = model.dim_lengths.get(dim, None)
-        if isinstance(current_length, TensorSharedVariable):
-            current_length = current_length.get_value()
-        elif isinstance(current_length, TensorConstant):
-            current_length = current_length.data
-        if (
-            current_coord is not None
-            and len(coord) == len(current_coord)
-            and np.all(coord == current_coord)
-        ) or (
-            # Coord was defined without values (only length)
-            current_coord is None and len(coord) == current_length
-        ):
-            constant_coords.add(dim)
-    return constant_coords
+def _build_constant_data(
+    trace_constant_data: dict[str, np.ndarray],
+    trace_coords: dict[str, np.ndarray],
+    model: Model,
+) -> dict[Variable, Any]:
+    """Collect trace-time values for data-like nodes in the model.
+
+    Returns a mapping from the model's registered data Variables (``pm.Data``
+    SharedVariables plus named Constants introduced by ``pm.do`` /
+    ``freeze_dims_and_data``) and its shared dim-length vars to the value each
+    held at inference time, as recovered from the trace.
+
+    Dim-length shareds are only included when the trace's coord labels still
+    match the model's (or when neither has explicit labels), so a same-length
+    coord with different labels still counts as volatile.
+    """
+    constant_data: dict[Variable, Any] = {}
+    for dv in model.data_vars:
+        if dv.name in trace_constant_data:
+            constant_data[dv] = trace_constant_data[dv.name]
+    for dim, length_var in model.dim_lengths.items():
+        if not isinstance(length_var, SharedVariable) or dim not in trace_coords:
+            continue
+        trace_coord = trace_coords[dim]
+        current_coord = model.coords.get(dim)
+        if current_coord is not None:
+            unchanged = len(current_coord) == len(trace_coord) and bool(
+                np.all(np.asarray(current_coord) == trace_coord)
+            )
+        else:
+            # Coord declared with a length only; match against the current length.
+            unchanged = int(length_var.get_value()) == len(trace_coord)
+        if unchanged:
+            constant_data[length_var] = len(trace_coord)
+    return constant_data
 
 
 def get_vars_in_point_list(trace, model):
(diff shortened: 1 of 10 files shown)


pymc/sampling/forward.py
```diff
@@ -687,79 +753,162 @@ def sample_posterior_predictive(
...
-    Re-compute the deterministic variable ``det``, conditioned on the posterior samples of ``z`` found in the trace:
+    Selecting a variable already in the trace just copies its values — no resampling
+    happens for ``z``, but ``obs`` is still resampled because it is observed:
 
     .. code :: python
 
-          pm.sample_posterior_predictive(idata, var_names=["det"], **kwargs)
-          # Sampling: []
+        with model:
+            pm.sample_posterior_predictive(idata, var_names=["z"], **kwargs)
```
ricardoV94: this is dumb, no point in sampling obs if it isn't needed for getting `var_names`

pymc/sampling/forward.py
```diff
@@ -105,13 +119,154 @@ def get_vars_in_point_list(trace, model):
...
+            var in volatile_vars
+            or (var in rvs and var not in vars_in_trace)
+            or (var.owner is None and _data_var_is_volatile(var, constant_data))
+            or (var.owner is not None and any(inp in volatile_closure for inp in var.owner.inputs))
+        ):
+            volatile_closure.add(var)
+    return volatile_closure
+
+
+def _maybe_warn_implicit_freeze(
```
ricardoV94, line 190: this one is a bit complex, but I think it really delivers UX for users

ricardoV94:
There is one test failing due to https://github.com/pymc-devs/pytensor/pull/2051

Verified that the patch fixes it.

Our step samplers don't eval before multiprocess splitting so numba compiles in each process... wasteful and shows the bug here.

We'll just wait but we may want to tweak the steps to eval once after compile.

Regular nuts pipeline does eval before split (in init nuts jitter)

jessegrabowski (commented):
General comment: the process of walking the fgraph and classifying nodes as volatile or not is exactly the function of a `Feature`, is there a reason why you didn't use one here?

pymc/sampling/forward.py
```diff
@@ -105,13 +119,154 @@ def get_vars_in_point_list(trace, model):
...
+    }
+    non_data_sources = {
+        var
+        for var in volatile_closure
+        if var in volatile_vars or (var in rv_set and var not in vars_in_trace_set)
+    }
+
+    warn_reasons: dict[str, list[str]] = {}
+    for rv in auto_frozen:
+        ancs = set(ancestors([rv])) - {rv}
```
jessegrabowski, line 218: ancestors returns the node itself?

pymc/sampling/forward.py
```diff
@@ -105,13 +119,154 @@ def get_vars_in_point_list(trace, model):
...
+    for dim_name, dim_length_var in model.dim_lengths.items():
+        if dim_name not in trace_coords or not isinstance(dim_length_var, TensorSharedVariable):
+            continue
+        new_length = int(dim_length_var.get_value())
+        old_length = len(trace_coords[dim_name])
+        if new_length != old_length:
+            length_changes.append((dim_name, old_length, new_length))
+
+    details = "; ".join(
+        f"{name!r} ({'; '.join(reasons)})" for name, reasons in sorted(warn_reasons.items())
```
jessegrabowski, line 245: Maybe a tree structure for the output of this, like dprint but with reasons  on the 2nd level and variables on the 1st level (maybe there are nested relationships, like A is volatile because (nest B is volatile because (nest Data changed)))

pymc/sampling/forward.py
```diff
@@ -592,16 +732,74 @@ def sample_posterior_predictive(
...
+            pm.Normal("obs", m["det"], 1, observed=[0.0])
+            idata = pm.sample(**kwargs)
+
+        do_m = pm.do(m, {m["det"]: m["x"] ** 3})
+
+        with do_m:
+            # Default copies old `x**2` values from the trace.
+            pm.sample_posterior_predictive(idata, var_names=["det"], **kwargs)
+            # Force recomputation using the new `x**3` graph.
+            pm.sample_posterior_predictive(idata, var_names=["det"], sample_vars=["det"], **kwargs)
```
jessegrabowski: use `sample_vars` without `var_names` in this example?

pymc/sampling/forward.py
```diff
@@ -687,80 +894,74 @@ def sample_posterior_predictive(
...
 
     .. code-block:: python
 
         with model:
-            pm.sample_posterior_predictive(idata, var_names=["obs"], **kwargs)
+            pm.sample_posterior_predictive(idata, **kwargs)
             # Sampling: [obs]
 
-    Re-compute the deterministic variable ``det``, conditioned on the posterior samples of ``z`` found in the trace:
+    Copy the trace values for ``z`` and ``det``. Nothing is resampled without explicit `sample_vars`:
```
jessegrabowski, line 911: This behavior seems completely useless, just a confusing footgun

pymc/sampling/forward.py
```diff
@@ -687,80 +894,74 @@ def sample_posterior_predictive(
...
             # Sampling: [z]
 
+    Generate samples of y, z and det, conditioned on the posterior samples of x found in the trace.
 
-    Generate samples of ``y``, ``z`` and ``det``, conditioned on the posterior samples of ``x`` found in the trace.
-
-    Note: The samples of ``y`` are equivalent to its prior, since it does not depend on any other variables.
-    In contrast, the samples of ``z`` and ``det`` depend on the new samples of ``y`` and the posterior samples of
-    ``x`` found in the trace.
+    Note: The samples of y are equivalent to its prior, since it does not depend on any other variables.
```
jessegrabowski: We might want this to be in a `.. warning::` directive

pymc/sampling/forward.py
```diff
@@ -861,28 +1062,126 @@ def sample_posterior_predictive(
...
+    # (upstream Data/coords changed, or an ancestor is being resampled).
+    auto_frozen_vars = [
+        rv for rv in trace_rvs if rv.name not in sample_var_names and rv not in frozen_vars
+    ]
+    frozen_vars.update(auto_frozen_vars)
+
+    # Run the full volatility analysis once, over the combined graph of
+    # output_vars (so we can classify deterministics in var_names)
+    # and the auto-frozen trace RVs (so the warning below can check their ancestors).
+    volatility_fg = FunctionGraph(outputs=list(output_vars) + auto_frozen_vars, clone=False)
```
jessegrabowski, line 1158: if this was a model fgraph instead would it make it easier to do some of the reasoning about data, deterministics, coords, etc?

tests/model/transform/test_conditioning.py
```diff
@@ -189,7 +189,7 @@ def test_do_posterior_predictive():
     # Replace `y` by a constant `100.0`
     m_do = do(m, {y: 100.0})
     with m_do:
-        idata_do = pm.sample_posterior_predictive(idata_m, var_names="z")
+        idata_do = pm.sample_posterior_predictive(idata_m, sample_vars=["z"])
```
jessegrabowski, line 192: are we dropping support for single-string inputs? I would prefer we don't, I like when I can just type one name if I only want one thing (vs a list of one item)

jessegrabowski (commented):
Some comments.

With this refactor I think we should really think about renaming this function to something like `conditional_resample` then implement `sample_posterior_predictive` as a special case that only generates the posterior predictive. My logic is that the functionality we now offer goes well beyond just sampling the posterior predictive distribution, which is a well-defined object in the literature.

pymc/sampling/forward.py
ricardoV94: Yes

pymc/sampling/forward.py
ricardoV94: It's the logical conclusion of the functionality. It would require EXTRA logic to not support it, and I don't think we should do that. Like we shouldn't error when we call `pm.sample` and there are no free rvs to sample, just return an empty posterior.

pymc/sampling/forward.py
ricardoV94: I think this is easier to follow, I'll add a note saying var_names wasn't needed as it would defaulted to the same

pymc/sampling/forward.py
ricardoV94: okay

pymc/sampling/forward.py
ricardoV94: maybe yes, OTOH that's much expensive (deep cloning of data variables and all that)

tests/model/transform/test_conditioning.py
ricardoV94: will readd

ricardoV94:
> General comment: the process of walking the fgraph and classifying nodes as volatile or not is exactly the function of a `Feature`, is there a reason why you didn't use one here?

A Feature is specifically about adding/keep meta information while a graph is being rewritten, we are not doing any graph rewrite here, just a snapshot analysis. I can't think of any angle where it would help

ricardoV94:
> With this refactor I think we should really think about renaming this function to something like conditional_resample then implement sample_posterior_predictive as a special case that only generates the posterior predictive. My logic is that the functionality we now offer goes well beyond just sampling the posterior predictive distribution, which is a well-defined object in the literature.

This was discussed in the original issue (and I think more internally) and rejected, Back then I was advocating for it, although I didn't have as good a name as yours (except the re- in resample, as we are not always "re-sampling"). 

But now I don't agree so much. What is the strict criteria for distinguishing whether this is a `sample_posterior_predictive` vs not? The freeze_vars is not imo. Missing vars is not (think of extra group in hierarchical models). Discarding trace variables? Again depends, maybe you are discarding because you are sampling new coordinates only, and you are just telling the function that.

So... I'm inclined to think there is no obvious set of arguments/restrictions that would constrict this function to only do "sample posterior predictive". A bit like "sample" may very well be used to get prior / prior_predictive draws (or compute deterministics), the meaning of the operation is not something you can pin-point from the computation being run.

If you agree with this, the argument becomes why we should have function with another name that has exactly the same signature/behavior? Since for most users the function does what they want, let the power users make their own alias if they want, and keep official naming simple

pymc/sampling/forward.py
jessegrabowski: how incestuous

jessegrabowski:
> This was discussed in the original issue (and I think more internally) and rejected, Back then I was advocating for it, although I didn't have as good a name as yours (except the re- in resample, as we are not always "re-sampling").

I don't agree that the ppd is hard to pointpoint though. It's defined by the fact that we're sampling the observed distribution, conditioned on the posterior samples -- it's the base case. The advantage would be to hide complexity from the median user. That said, it is likely sufficient that `pm.sample_posterior_predictive(idata)` does the right thing.

If this was already discussed let's move on instead of re-litigating it ad nauseam. I accept the judgement of the collective.

pymc/sampling/forward.py
jessegrabowski: ok, but does that point to a flaw in model fgraph implementation? Do we need model frozen fgraph xD

ricardoV94:
> It's defined by the fact that we're sampling the observed distribution, conditioned on the posterior samples -- it's the base case.

As long as you use even one sample from the posterior it fits the mold. I don't think it must use all samples from the posterior (trivial case: dead path) or can't introduce new variables (e.g.,  marginalized variables that are being recovered, you are using a stochastic approximation to the posterior, conjugate posterior with closed-form solution)?

Also maybe you didn't mark your variable as observed from PyMC's standpoint (but you know this is your observation). I don't think the meaning can be decided in general.

pymc/sampling/forward.py
ricardoV94: OTOH the lower level function here doesn't need to presume any model, just the minimum set of rv variables needed to infer volatility -> define the pytensor function.

pymc/sampling/forward.py
ricardoV94: can we leave to follow up PR?

pymc/sampling/forward.py
ricardoV94: sorry I thought this was on the section on volatility. Agree. Also removed the first example. The nudge being users shouldn't think about var_names (or at least not first). but sample_vars.

pymc/sampling/forward.py
ricardoV94: done, but it's weird with the In contrast part after ... ?

tests/model/transform/test_conditioning.py
ricardoV94: done

pymc/sampling/forward.py
jessegrabowski: sure

jessegrabowski (approved):

PR #8215, opened 2026-03-23
Fix expected covariance in test

### LINKED ISSUE #8214: Failing `tests/step_methods/hmc/test_quadpotential.py::test_full_adapt_sample_p`
### Description

Getting some failures in the CI recently

ricardoV94:
Closes #8214

I think the value was wrong, and after 4ea140633 it stopped being implicitly seeded by the global seed.

My understanding is that the test should have a failure rate of >5sigma, but was failing like 1.5% of the time with random seed. I think they just missed a square factor on the diagonal terms.

The bots agree...

### DIFF
--- a/tests/step_methods/hmc/test_quadpotential.py
+++ b/tests/step_methods/hmc/test_quadpotential.py
@@ -195,16 +195,17 @@ def test_weighted_covariance(ndim=10, seed=5432):
     assert np.allclose(cov_est2, cov_est0)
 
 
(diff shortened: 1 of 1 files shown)


jessegrabowski (approved):
i do like spite

ricardoV94:
> i do like spite

Sounding like a cuckoo out of context

jessegrabowski:
that's the ha-ha point

PR #8228, opened 2026-03-28
Clarify Wishart warning messaging

yejiming:
## Summary
- clarify the `pm.Wishart` docstring to explain that PyMC can evaluate its log-probability but it should not be used as a covariance prior in models
- align the runtime warning with that wording and keep the LKJ alternatives explicit
- update Wishart warning assertions in the distribution tests and shared test helpers

## Testing
- `HOME=/tmp XDG_CACHE_HOME=/tmp MPLCONFIGDIR=/tmp .venv/bin/python -m pytest tests/distributions/test_multivariate.py -k Wishart -q`

### DIFF
--- a/pymc/distributions/multivariate.py
+++ b/pymc/distributions/multivariate.py
@@ -998,8 +998,9 @@ class Wishart(Continuous):
 
     Notes
     -----
-    This distribution is unusable in a PyMC model. You should instead
-    use LKJCholeskyCov or LKJCorr.
+    PyMC can evaluate the log-probability of a Wishart distribution, but it
+    should not be used as a prior for covariance matrices in a PyMC model.
+    Use LKJCholeskyCov or LKJCorr instead.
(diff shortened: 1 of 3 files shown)


twiecki:
Well technically it should not be used as a prior, period, not just for covariance matrices, no?

ricardoV94:
> Well technically it should not be used as a prior, period, not just for covariance matrices, no?

Depends what you are going to sample it with. If it's forward only it's fine, if you have a custom sampler it's fine, if you have a custom transform it's fine.

Maybe you are not even going to sample it, just evaluate a model logp for whatever purposes you want. Maybe you just want a pretty grahpviz...

This warning is too opinionated.

ricardoV94:
Ah I jumped too fast, yes. In that context, it shouldn't be used as a prior period, doesn't matter what that prior is used for next

PR #8235, opened 2026-03-31
docs: add state-space example to Potential docstring (#6399)

### LINKED ISSUE #6399: ENH: Add code example to Potential docstrings

Surjeet4146:
Hi team,

Working on this for GSoC 2026 (Scalable Online Bayesian State Space Models — see my post in the Discourse thread).

Added a short example showing how `pm.Potential` can be used for custom terms in state-space models (smoothness penalty between consecutive states).

Closes #6399.

### DIFF
--- a/pymc/model/core.py
+++ b/pymc/model/core.py
@@ -2381,7 +2381,17 @@ def Potential(name, var: TensorVariable, model=None, dims=None) -> TensorVariabl
 
             data = [0.1, 0.5, 0.9]
             llike = pm.Potential("llike", normal_logp(data, mu, sigma))
+    A Potential can be used to implement custom transition or observation terms in state-space models.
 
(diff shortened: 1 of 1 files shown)


ricardoV94:
that issue is already closed this isn't conceptually very direct from the first example (add vs sub). We can't add every conceivable application of potential to the docs

PR #8236, opened 2026-03-31
Add ADVI float32 test to test_vi_float_precision (#5678)

### LINKED ISSUE #5678: Add VI to float32 test suit
VI is sensitive to float32 and might break. Currently it it is [missing](https://github.com/pymc-devs/pymc/blob/741d45540016ab52a15ad5bff521614b6ae1d835/.github/workflows/tests.yml#L350) from the float32 test suit

anandraj095:
This PR adds a parameterized test for Variational Inference (ADVI) under float32 precision.
It ensures VI does not break when using float32, complementing the existing float64 tests.

- Parameterized test using `pytest.mark.parametrize` for float32 and float64
- Overrides `pytensor.config.floatX` within the test
- Asserts that posterior samples (`mu` and `sigma`) have the correct dtype

Closes #5678.

### DIFF
--- a/tests/variational/test_vi_float.py
+++ b/tests/variational/test_vi_float.py
@@ -1,0 +1,47 @@
+import numpy as np
+import pymc as pm
+import pytest
+import pytensor
+
+
+@pytest.mark.parametrize("dtype_str", ["float32", "float64"])
+def test_vi_float_precision(dtype_str):
+    """
+    Test that Variational Inference (ADVI) works under float32 and float64 precision.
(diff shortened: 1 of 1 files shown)


anandraj095:
Hi, CI is currently failing because `test_vi_float.py` appears to be completely ignored.

It seems this file might not be included in the CI test matrix.
Would you prefer that I add it to `tests.yml`, or should these tests be integrated into an existing file?

Thanks!

PR #8243, opened 2026-04-08
Handle new PyTensor v3 breaking changes

jessegrabowski:


ricardoV94:
Truncation test failing. Bot came up with this MWE, I didn't have time to check, suspect the issue is in PyMC not PyTensor.

```python
"""Pytensor scan + until + OpFromGraph (with rng inputs) drops rng updates.

Without ``until`` the same scan body propagates the rng updates correctly
across iterations.  With ``until`` the inner OpFromGraph's rng inputs are
stuck on their initial values and every iteration produces identical draws.
"""
import warnings
warnings.simplefilter("ignore")

import numpy as np
import pytensor
import pytensor.tensor as pt
from pytensor.scan import scan, until

from pymc import HalfNormal, LogNormal, Mixture, CustomDist


def mix_dist_fn(size):
    return Mixture.dist(
        w=[0.3, 0.7],
        comp_dists=[HalfNormal.dist(), LogNormal.dist()],
        shape=size,
    )


x = CustomDist.dist(dist=mix_dist_fn, shape=(5,))


def loop_fn_no_until(prev, *rngs):
    new = x.owner.op.make_node(*rngs).default_output()
    return new, collect_default_updates(new)


def loop_fn_until(prev, *rngs):
    new = x.owner.op.make_node(*rngs).default_output()
    return new, collect_default_updates(new), until(pt.all(new < -100))


print("=== no until ===")
out, updates = scan(
    loop_fn_no_until,
    outputs_info=[pt.zeros((5,))],
    non_sequences=list(x.owner.inputs),
    n_steps=3,
    strict=True,
)
print(pytensor.function([], out, updates=updates)())

print("=== with until ===")
out, updates = scan(
    loop_fn_until,
    outputs_info=[pt.zeros((5,))],
    non_sequences=list(x.owner.inputs),
    n_steps=3,
    strict=True,
)
print(pytensor.function([], out, updates=updates)())
```

ricardoV94:
This is blocked on https://github.com/pymc-devs/pytensor/pull/2039

ricardoV94 (approved):

ricardoV94:
Probably need to change some linalg imports now?

ricardoV94:
@jessegrabowski you should review my changes on this PR. Mostly handling rng/scan API  change

jessegrabowski:
I'll look for linalg stuff as well. I *think* I did a cleanup pass on this once already, so everything was using `pt.linalg`, but doesn't hurt to check.

ricardoV94:
We seem to have broken the slinalg/nlinalg shims (from the failing tests)

pymc/distributions/timeseries.py
```diff
@@ -412,7 +411,7 @@ def __init__(self, *args, ar_order, constant_term, **kwargs):
     @classmethod
     def rv_op(cls, rhos, sigma, init_dist, steps, ar_order, constant_term, size=None):
         # We don't allow passing `rng` because we don't fully control the rng of the components!
```
jessegrabowski, line 413: is this still true?

ricardoV94: yes, change_dist_size doesn't take a rng as input and so those always create a fresh RNG.

Even if it did it would be tricky to know if we didn't forget to plug somewhere and it still ended up using fresh rngs. 

Something to think about later down the road

tests/distributions/test_truncated.py
```diff
@@ -537,6 +538,22 @@ def mix_dist_fn(size):
     next_rngs = [out for out in x_trunc.owner.outputs if isinstance(out.type, RandomType)]
     assert len(set(rngs)) == len(set(next_rngs)) == 3
 
+    # Verify that all RNG inputs inside the compiled scan are destroyed
+    # (i.e., mutated in-place by a consumer). A regression could turn an
+    # RNG update into a ViewOp or DeepCopy, leaving the original untouched
+    # and causing stuck draws across scan iterations.
+    from pymc.pytensorf import compile as pymc_compile
```
jessegrabowski: imports to the top

ricardoV94: fixed

PR #8245, opened 2026-04-08
Fix latex error in documentation

alipatti:
The [introductory overview](https://www.pymc.io/projects/docs/en/latest/learn/core_notebooks/pymc_overview.html) page of the docs currently has a broken MathJax env caused by an `align` environment being used where there should be an `aligned`.

image

Pretty sure this will fix the issue, but I can't get the development environment working on my machine at the moment to re-render the docs and confirm.

Thanks for a great project!


ricardoV94:
We'll have to wait until after #8199 for the read the docs preview to render. The fix does seem right!

Thanks a lot in advance

